PR #1011, opened 2019-04-05
receive: Add block shipping

brancz:
## Changes

This adds block shipping to the `receive` component.

## Verification

Ran the quickstart script and watched blocks being uploaded, and stopped the `receive` component and could still query the uploaded data. Ideally this would have an e2e test, but I'm not sure how to realize that, and I couldn't find e2e tests for this for the sidecar component either.

@domgreen @bwplotka @FUSAKLA @GiedriusS @povilasv

cc @metalmatze @squat


cmd/thanos/receive.go
```diff
@@ -65,10 +74,26 @@ func runReceive(
...
 ) error {
 	logger = log.With(logger, "component", "receive")
 	level.Warn(logger).Log("msg", "setting up receive; the Thanos receive component is EXPERIMENTAL, it may break significantly without notice")
 
+	externalLabels := labels.Labels{}
+	labelStrings := strings.Split(externalLabelsString, ",")
+	for _, labelString := range labelStrings {
+		keyValueStrings := strings.Split(labelString, "=")
+		if len(keyValueStrings) != 2 {
+			return errors.New("external labels are malformed")
```
povilasv: consider doing a bit better error message here, i.e. which label was malormed.

cmd/thanos/receive.go
```diff
@@ -225,6 +250,12 @@ func runReceive(
 			},
 		)
 	}
+
+	err := addShipper(logger, reg, g, nil, metadata.ReceiveSource, &promMetadata{labels: externalLabels}, dataDir, objStoreConfig, false)
+	if err != nil {
+		return err
```
povilasv: maybe wrap the error here?

cmd/thanos/sidecar.go
```diff
@@ -229,6 +229,16 @@ func runSidecar(
 		})
 	}
 
+	err := addShipper(logger, reg, g, peer, metadata.SidecarSource, m, dataDir, objStoreConfig, uploadCompacted)
+	if err != nil {
+		return err
```
povilasv: could we wrap the error here

cmd/thanos/sidecar.go
```diff
@@ -281,10 +291,13 @@ func runSidecar(
...
-				} else {
-					m.UpdateTimestamps(minTime, math.MaxInt64)
+					return nil
+				}
+
+				m.UpdateTimestamps(minTime, math.MaxInt64)
+				mint, maxt := m.Timestamps()
 
-					mint, maxt := m.Timestamps()
+				if peer != nil {
```
povilasv: I think we don't need this as we have noop peer in the gossip, and this will be removed in https://github.com/improbable-eng/thanos/pull/1008 so less merge issues for me :D

povilasv:
@brancz re e2e tests, we definetely need to work on this.

brancz:
Addressed comments. I'd say let's do upload tests for sidecar and receive as a follow up @povilasv ?

brancz:
Looks like something was odd with Circle CI, re-triggering with force push.

cmd/thanos/receive.go
```diff
@@ -65,10 +74,26 @@ func runReceive(
...
 ) error {
 	logger = log.With(logger, "component", "receive")
 	level.Warn(logger).Log("msg", "setting up receive; the Thanos receive component is EXPERIMENTAL, it may break significantly without notice")
 
+	externalLabels := labels.Labels{}
+	labelStrings := strings.Split(externalLabelsString, ",")
+	for _, labelString := range labelStrings {
+		keyValueStrings := strings.Split(labelString, "=")
+		if len(keyValueStrings) != 2 {
+			return errors.New("external labels are malformed: a list of labels passed to the --external-labels flag must be comma separated and labels in key=value form")
```
povilasv:
```suggestion
			return errors.Errorsf("external labels are malformed: a list of labels passed to the --external-labels flag must be comma separated and labels in key=value form. Got a label: %s",labelString)
```

cmd/thanos/receive.go
```diff
@@ -36,6 +39,10 @@ func registerReceive(m map[string]setupFunc, app *kingpin.Application, name stri
 	dataDir := cmd.Flag("tsdb.path", "Data directory of TSDB.").
 		Default("./data").String()
 
+	externalLabels := cmd.Flag("external-labels", "External labels to announce. Comma separated list of key=value pairs. This flag will be removed in the future when handling multiple tsdb instances is added.").PlaceHolder("key1=value1,key2=value2").String()
```
bwplotka: So we have just `--labels` naming in ruler. We should be consistent proably and rename this or ruler ones?

```
This flag will be removed in the future when handling multiple tsdb instances is added.
```
Any view how it will look like then?

cmd/thanos/receive.go
```diff
@@ -36,6 +39,10 @@ func registerReceive(m map[string]setupFunc, app *kingpin.Application, name stri
...
+
+	objStoreConfig := regCommonObjStoreFlags(cmd, "", false)
```
bwplotka: Isn't this required? Basing of fact that retention is not configurable?

cmd/thanos/receive.go
```diff
@@ -65,10 +74,26 @@ func runReceive(
...
 	remoteWriteAddress string,
 	dataDir string,
+	objStoreConfig *pathOrContent,
+	externalLabelsString string,
 ) error {
 	logger = log.With(logger, "component", "receive")
 	level.Warn(logger).Log("msg", "setting up receive; the Thanos receive component is EXPERIMENTAL, it may break significantly without notice")
 
+	externalLabels := labels.Labels{}
+	labelStrings := strings.Split(externalLabelsString, ",")
```
bwplotka: There is `parseLabels` function that gives you that and is used in more than one place I believe, can we reuse it? https://github.com/improbable-eng/thanos/blob/59634d99e4b7aa74d53814ed9b2401f6068e17cf/cmd/thanos/rule.go#L718

cmd/thanos/receive.go
bwplotka: Not a blocker,

cmd/thanos/receive.go
```diff
@@ -225,6 +250,12 @@ func runReceive(
 			},
 		)
 	}
+
+	err := addShipper(logger, reg, g, metadata.ReceiveSource, &promMetadata{labels: externalLabels}, dataDir, objStoreConfig, false)
```
bwplotka: We used to do `if err := ... ; err != nil {` paradigm in Thanos (:

cmd/thanos/receive.go
```diff
@@ -225,6 +250,12 @@ func runReceive(
...
+	if err != nil {
+		return errors.Wrap(err, "failed to add block shipper to run group")
```
bwplotka: Minor nit, when you wrap error for Thanos, let's avoid `failed to` prefixes. It's already known it failed. Wrap is only to figure out where it fail and why NOT to check IF it failed (: 
```suggestion
		return errors.Wrap(err, "add block shipper to run group")
```

cmd/thanos/sidecar.go
```diff
@@ -229,6 +229,16 @@ func runSidecar(
 		})
 	}
 
+	err := addShipper(logger, reg, g, metadata.SidecarSource, m, dataDir, objStoreConfig, uploadCompacted)
```
bwplotka: ditto with `if err := ... ` paradigm

cmd/thanos/sidecar.go
```diff
@@ -229,6 +229,16 @@ func runSidecar(
...
+	if err != nil {
+		return errors.Wrap(err, "failed to add block shipper to run group")
```
bwplotka: ditto with error wrap

cmd/thanos/sidecar.go
```diff
@@ -229,6 +229,16 @@ func runSidecar(
...
+	}
+
+	level.Info(logger).Log("msg", "starting sidecar", "peer", peer.Name())
+	return nil
+}
+
+func addShipper(logger log.Logger, reg prometheus.Registerer, g *run.Group, sourceType metadata.SourceType, m *promMetadata, dataDir string, objStoreConfig *pathOrContent, uploadCompacted bool) error {
 	confContentYaml, err := objStoreConfig.Content()
```
bwplotka: I am not a fan of this. You save a line but `addShipper` means to me your are actually adding this NOT adding only IF something happens, can we rearrange? Sorry for being picky ): I don't like methods that hides and suprises.

cmd/thanos/sidecar.go
```diff
@@ -229,6 +229,16 @@ func runSidecar(
...
+	err := addShipper(logger, reg, g, metadata.SidecarSource, m, dataDir, objStoreConfig, uploadCompacted)
+	if err != nil {
+		return errors.Wrap(err, "failed to add block shipper to run group")
+	}
+
+	level.Info(logger).Log("msg", "starting sidecar", "peer", peer.Name())
+	return nil
+}
+
+func addShipper(logger log.Logger, reg prometheus.Registerer, g *run.Group, sourceType metadata.SourceType, m *promMetadata, dataDir string, objStoreConfig *pathOrContent, uploadCompacted bool) error {
```
bwplotka: What about rule.go shipper?

cmd/thanos/receive.go
brancz: It'll be a whitelist of label-names, and set on the fly based on incoming data, as receivers have to handle multiple. As the proposal mentioned, we'll need to add the ability to publish multiple sets of labels that a receiver can respond to.

Agreed regarding consistency, will adapt.

cmd/thanos/receive.go
brancz: I could see people just caring about a single instance of this with just local storage. I can see scenarios where it would not have to be set.

cmd/thanos/receive.go
brancz: Wonderful :+1:

cmd/thanos/sidecar.go
brancz: Agreed.

cmd/thanos/sidecar.go
brancz: Not sure exactly what you mean here.

brancz: Ah, I need to look at that, I'll take care of it tomorrow!

brancz: I looked at it a bit, and I feel like the best decision around this for now is to not generalize this in the `addShipper` function, but once gossip removal lands, this is nicely generalizable for receive/sidecar/rule.

brancz: Actually reducing this to what's necessary in receive, I think I might even prefer the slight duplication, as the code necessary is _a lot_ smaller than the `addShipper` function.

cmd/thanos/receive.go
brancz: As a side note, retention should be configurable. People should be able to make a choice how they want local storage to be treated when running receive I think

PR #1016, opened 2019-04-09
dns: Added miekgdns resolver as a hidden option to query and ruler.

bwplotka:
Fixes: https://github.com/improbable-eng/thanos/issues/1015

### DIFF
--- a/cmd/thanos/query.go
+++ b/cmd/thanos/query.go
@@ -82,8 +82,12 @@ func registerQuery(m map[string]setupFunc, app *kingpin.Application, name string
 	fileSDInterval := modelDuration(cmd.Flag("store.sd-interval", "Refresh interval to re-read file SD files. It is used as a resync fallback.").
 		Default("5m"))
 
+	// TODO(bwplotka): Grab this from TTL at some point.
 	dnsSDInterval := modelDuration(cmd.Flag("store.sd-dns-interval", "Interval between DNS resolutions.").
 		Default("30s"))
+
+	dnsSDResolver := cmd.Flag("store.sd-dns-resolver", fmt.Sprintf("Resolver to use. Possible options: [%s, %s]", dns.GolangResolverType, dns.MiekgdnsResolverType)).
+		Default(string(dns.GolangResolverType)).Hidden().String()
 
 	unhealthyStoreTimeout := modelDuration(cmd.Flag("store.unhealthy-timeout", "Timeout before an unhealthy store is cleaned from the store UI page.").Default("5m"))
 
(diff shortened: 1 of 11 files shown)


cmd/thanos/query.go
```diff
@@ -82,9 +82,13 @@ func registerQuery(m map[string]setupFunc, app *kingpin.Application, name string
 	fileSDInterval := modelDuration(cmd.Flag("store.sd-interval", "Refresh interval to re-read file SD files. It is used as a resync fallback.").
 		Default("5m"))
 
+	// TODO(bwplotka): Grab this from TTL at some point.
 	dnsSDInterval := modelDuration(cmd.Flag("store.sd-dns-interval", "Interval between DNS resolutions.").
 		Default("30s"))
 
+	dnsSDResolver := cmd.Flag("store.sd-dns-resolver", fmt.Sprintf("Resolver to use. Possible options: [%s, %s]", dns.GolangResolverType, dns.MiekgdnsResolverType)).
```
ivan-valkov: can you link to the issue so that people know why they would use one over the other? Should this one also be hidden like in ruler?

pkg/discovery/dns/provider.go
```diff
@@ -22,19 +24,39 @@ type Provider struct {
...
+	MiekgdnsResolverType ResolverType = "miekgdns"
+)
+
+func (t ResolverType) ToResolver(logger log.Logger) ipLookupResolver {
+	var r ipLookupResolver
+	switch t {
+	case MiekgdnsResolverType:
+		r = &miekgdns.Resolver{ResolvConf: miekgdns.DefaultResolvConfPath}
+	default:
+		level.Warn(logger).Log("msg", "no such resolver type, defaulting to golang", "type", t)
```
ivan-valkov: you will warn even if I select `golang`. and another case

bwplotka:
@ivan-valkov I documented on CHANGELOG, can you see this?

cmd/thanos/query.go
bwplotka: I think we should unhide maybe (or hide) in all of them, my mistake. For reference.. not sure. Do you think that\s the proper place for it?

ivan-valkov:
Yeah, maybe that is good enough, especially if it stays as a hidden flag.

cmd/thanos/query.go
ivan-valkov: if we don't hide the flag, I think this is the place where it will be nice to see what they actually do. As a user I have no idea what miekgdns is.

PR #1021, opened 2019-04-11
feature: support POST method for series endpoint

eahydra:
Signed-off-by: Joseph Lee <joseph.t.lee@outlook.com>

## Changes

Prometheus v2.9.0 support POST method for Query API `/api/v1/series` endpoint
We need do same change in Thanos that compatible with Prometheus 

## Verification

Have run in our daily and product environment.


povilasv:
Could you add tests that verify this behaviour.

Especially the test with stuff passed as a form instead of query params.

>You can URL-encode these parameters directly in the request body by using the POST method and Content-Type: application/x-www-form-urlencoded header. This is useful when specifying a large or dynamic number of series selectors that may breach server-side URL character limits.

eahydra:
Hi @povilasv , I have add the unit tests, but circleci failed that caused by other tests.  I don't know how to rerun the test job, can you help me?

bwplotka:
Happy to merge, if @povilasv is happy too.

PR #1026, opened 2019-04-11
dns+ruler: Improve tests, block scheme for now.

bwplotka:

### DIFF
--- a/cmd/thanos/rule.go
+++ b/cmd/thanos/rule.go
@@ -92,7 +92,7 @@ func registerRule(m map[string]setupFunc, app *kingpin.Application, name string)
 
 	objStoreConfig := regCommonObjStoreFlags(cmd, "", false)
 
-	queries := cmd.Flag("query", "Addresses of statically configured query API servers (repeatable). The scheme may be prefixed with 'dns+' or 'dnssrv+' to detect query API servers through respective DNS lookups.").
+	queries := cmd.Flag("query", "Addresses of statically configured query API servers (repeatable). The scheme may be prefixed with 'dns+' or 'dnssrv+' to detect query API servers through respective DNS lookups. No scheme prefix allowed (https://github.com/improbable-eng/thanos/issues/1025)").
 		PlaceHolder("<query>").Strings()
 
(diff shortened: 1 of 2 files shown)


cmd/thanos/rule.go
```diff
@@ -92,7 +92,7 @@ func registerRule(m map[string]setupFunc, app *kingpin.Application, name string)
 
 	objStoreConfig := regCommonObjStoreFlags(cmd, "", false)
 
-	queries := cmd.Flag("query", "Addresses of statically configured query API servers (repeatable). The scheme may be prefixed with 'dns+' or 'dnssrv+' to detect query API servers through respective DNS lookups.").
+	queries := cmd.Flag("query", "Addresses of statically configured query API servers (repeatable). The scheme may be prefixed with 'dns+' or 'dnssrv+' to detect query API servers through respective DNS lookups. No scheme prefix allowed (https://github.com/improbable-eng/thanos/issues/1025)").
```
povilasv: I think the comment here doesn't make sense. As we have a sentence `The scheme may be prefixed with 'dns+' or 'dnssrv+'` 
and after that a sentence `No scheme prefix allowed (https://github.com/improbable-eng/thanos/issues/1025)")`

cmd/thanos/rule.go
```diff
@@ -130,6 +130,13 @@ func registerRule(m map[string]setupFunc, app *kingpin.Application, name string)
 
 		lookupQueries := map[string]struct{}{}
 		for _, q := range *queries {
+			if q == "" {
+				return errors.New("Static querier address cannot be empty")
+			}
+			if strings.Contains(q, "://") {
```
povilasv: So we won't support `dns://` urls anymore?

cmd/thanos/rule.go
bwplotka: yes, We need to reword ;p

It's tricky as we can do `dns+abc:port`, `abc:port` etc, but we cannot put `dns+http://abc:port` or http://abc:port` due to implementation limitations

cmd/thanos/rule.go
bwplotka: We never was suporting those. Only `dns+` URLS and others with `+`

PR #1029, opened 2019-04-12
compact:  group concurrency improvements

mjd95:
Follow on from https://github.com/improbable-eng/thanos/pull/1010

https://github.com/improbable-eng/thanos/pull/1010 added group-level concurrency to compaction.  It used an `errGroup` to manage the concurrency, which errors out all the concurrent tasks as soon as one fails, which can mean wasted resources and partial uploads.

This PR changes the behaviour so that once an error is encountered, we wait before all the active goroutines finish before erroring out the compaction.

### DIFF
--- a/pkg/compact/compact.go
+++ b/pkg/compact/compact.go
@@ -7,6 +7,7 @@ import (
 	"os"
 	"path/filepath"
 	"sort"
+	"strings"
 	"sync"
 	"time"
 
@@ -22,7 +23,6 @@ import (
 	"github.com/prometheus/client_golang/prometheus"
 	"github.com/prometheus/tsdb"
 	"github.com/prometheus/tsdb/labels"
-	"golang.org/x/sync/errgroup"
 )
 
 type ResolutionLevel int64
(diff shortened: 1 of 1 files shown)


pkg/compact/compact.go
```diff
@@ -966,16 +971,28 @@ func (c *BucketCompactor) Compact(ctx context.Context) error {
 		}
 
 		// Send all groups found during this pass to the compaction workers.
+	groupLoop:
 		for _, g := range groups {
 			select {
-			case <-errGroupCtx.Done():
-				break
+			case err = <-errChan:
+				if err != nil {
```
bwplotka: minor inconsistency. You are checking if err != nil here but not on line 992. I would say, don't check it our code couple of lines above never pushes nil? However if channel is closed you would get nil here, but reader is responsible for closing

pkg/compact/compact.go
```diff
@@ -966,16 +971,28 @@ func (c *BucketCompactor) Compact(ctx context.Context) error {
...
 			case groupChan <- g:
 			}
 		}
 		close(groupChan)
-		if err := errGroup.Wait(); err != nil {
-			return err
+		wg.Wait()
+
+		close(errChan)
+		workCtxCancel()
```
bwplotka: Why you cancel if all go routines already left? just to release resources right?

pkg/compact/compact.go
```diff
@@ -966,16 +971,28 @@ func (c *BucketCompactor) Compact(ctx context.Context) error {
...
+		if err != nil {
+			errMsgs := []string{err.Error()}
+			// Collect any other errors reported by the workers.
+			for e := range errChan {
```
bwplotka: :+1:

pkg/compact/compact.go
mjd95: Yeah good point.  I've removed the extra check here as I don't think it's needed, as you said the reader is doing the closing.

pkg/compact/compact.go
mjd95: Yeah just to release the resources

pkg/compact/compact.go
```diff
@@ -921,17 +921,22 @@ func (c *BucketCompactor) Compact(ctx context.Context) error {
...
+			finishedAllGroups      = true
+			mtx                    sync.Mutex
 		)
 
 		// Set up workers who will compact the groups when the groups are ready.
+		// They will compact available groups until they encounter an error, after which they will stop.
 		for i := 0; i < c.concurrency; i++ {
-			errGroup.Go(func() error {
+			go func() {
+				wg.Add(1)
```
povilasv: Could we move this to `main` goroutine as per docs:

> Typically this means the calls to Add should execute before the statement creating the goroutine or other event to be waited for.

https://golang.org/pkg/sync/#WaitGroup.Add

bwplotka: Yes! That's a major one. Good spotting :+1: 

Otherwise let's merge.

mjd95: Good catch!  Fixed now

PR #1035, opened 2019-04-16
Change Proxy to also be Store Client

povilasv:
## Changes

For `pkg/querier/querier.go` StoreClient interface is much nicer as `Series` call takes a context and all the calls take `grpc.CallOptions`. This allows us to set Deadlines/Timeouts/ Grpc specific options per grpc call and per grpc client .

In future I plan to makes this configurable when we get rid of gossip.

Further more HTTP2 supports streaming. So in theory we could stream back the data to Query clients as soon as we have some series computed and merged.

This also simplifies our context passing as now Series call takes a context in an argument:

> Series(ctx context.Context, in *SeriesRequest, opts ...grpc.CallOption)

### Server Interface:
```
	// / Info returns meta information about a store e.g labels that makes that store unique as well as time range that is
	// / available.
	Info(context.Context, *InfoRequest) (*InfoResponse, error)
	// / Series streams each Series (Labels and chunk/downsampling chunk) for given label matchers and time range.
	Series(*SeriesRequest, Store_SeriesServer) error
	// / LabelNames returns all label names that is available.
	// / Currently unimplemented in all Thanos implementations, because Query API does not implement this either.
	LabelNames(context.Context, *LabelNamesRequest) (*LabelNamesResponse, error)
	// / LabelValues returns all label values for given label name.
	LabelValues(context.Context, *LabelValuesRequest) (*LabelValuesResponse, error)

```

## Client interface 

```
type StoreClient interface {
	// / Info returns meta information about a store e.g labels that makes that store unique as well as time range that is
	// / available.
	Info(ctx context.Context, in *InfoRequest, opts ...grpc.CallOption) (*InfoResponse, error)
	// / Series streams each Series (Labels and chunk/downsampling chunk) for given label matchers and time range.
	Series(ctx context.Context, in *SeriesRequest, opts ...grpc.CallOption) (Store_SeriesClient, error)
	// / LabelNames returns all label names that is available.
	// / Currently unimplemented in all Thanos implementations, because Query API does not implement this either.
	LabelNames(ctx context.Context, in *LabelNamesRequest, opts ...grpc.CallOption) (*LabelNamesResponse, error)
	// / LabelValues returns all label values for given label name.
	LabelValues(ctx context.Context, in *LabelValuesRequest, opts ...grpc.CallOption) (*LabelValuesResponse, error)
}
```

REF: https://github.com/improbable-eng/thanos/pull/895

### DIFF
--- a/pkg/query/querier.go
+++ b/pkg/query/querier.go
@@ -29,7 +29,7 @@ type WarningReporter func(error)
 type QueryableCreator func(deduplicate bool, maxSourceResolution time.Duration, partialResponse bool, r WarningReporter) storage.Queryable
 
 // NewQueryableCreator creates QueryableCreator.
-func NewQueryableCreator(logger log.Logger, proxy storepb.StoreServer, replicaLabel string) QueryableCreator {
+func NewQueryableCreator(logger log.Logger, proxy storepb.StoreClient, replicaLabel string) QueryableCreator {
 	return func(deduplicate bool, maxSourceResolution time.Duration, partialResponse bool, r WarningReporter) storage.Queryable {
 		return &queryable{
 			logger:              logger,
@@ -46,7 +46,7 @@ func NewQueryableCreator(logger log.Logger, proxy storepb.StoreServer, replicaLa
 type queryable struct {
 	logger              log.Logger
 	replicaLabel        string
-	proxy               storepb.StoreServer
+	proxy               storepb.StoreClient
 	deduplicate         bool
 	maxSourceResolution time.Duration
 	partialResponse     bool
@@ -64,7 +64,7 @@ type querier struct {
 	cancel              func()
 	mint, maxt          int64
 	replicaLabel        string
-	proxy               storepb.StoreServer
+	proxy               storepb.StoreClient
 	deduplicate         bool
 	maxSourceResolution int64
 	partialResponse     bool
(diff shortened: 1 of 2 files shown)


povilasv:
Thinking about some more I have doubts about the usefullness of it.

A part from passing context to `Series` call (which is really nice) it doesn't really have more benefits?

As you can do per store config in proxy.go
Also the current interface allows to stream back the data?

povilasv:
> I think you are missing essential thing, that:
> 
> `Series(*SeriesRequest, Store_SeriesServer) error`
> 
> `Store_SeriesServer` has `Context()` inside. And we set it in `proxy.go` :
> 
> [thanos/pkg/query/querier.go](https://github.com/improbable-eng/thanos/blob/1cd9ddd14999d6b074f34a4328e03f7ac3b7c26a/pkg/query/querier.go#L183)
> 
> Line 183 in [1cd9ddd](/improbable-eng/thanos/commit/1cd9ddd14999d6b074f34a4328e03f7ac3b7c26a)
> 
>  resp := &seriesServer{ctx: ctx}

Yeah so I know this.

I'm just interested in what you think as I really found Thanos Query calling proxy which is `GRPC Server` confusing as hell.

IMO Thanos Query should use a `MergeSeriesSetClient` to connect to the backends. Thus the proxy would be a GRPC client. In this case we can pass grpc call options, context as first argument directly from the querier.

But this is just for exploration of cleaner code, rather than for any actual issue fix.

PR #1041, opened 2019-04-16
ruler: Fixed Chunk's samples number overflowing Max Uint16.

bwplotka:
Fixes https://github.com/improbable-eng/thanos/issues/1038

CC @davtsur

### DIFF
--- a/pkg/store/tsdb.go
+++ b/pkg/store/tsdb.go
@@ -76,10 +76,6 @@ func (s *TSDBStore) Series(r *storepb.SeriesRequest, srv storepb.Store_SeriesSer
 		return status.Error(codes.InvalidArgument, err.Error())
 	}
 
-	// TODO(fabxc): An improvement over this trivial approach would be to directly
-	// use the chunks provided by TSDB in the response.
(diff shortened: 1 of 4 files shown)


pkg/store/tsdb.go
```diff
@@ -111,31 +114,46 @@ func (s *TSDBStore) Series(r *storepb.SeriesRequest, srv storepb.Store_SeriesSer
...
-	}
-	var mint int64
+		app.Append(it.At())
+		chkMaxt, _ := it.At()
 
-	for i := 0; it.Next(); i++ {
-		if i == 0 {
-			mint, _ = it.At()
+		isNext = it.Next()
+		if isNext && chk.NumSamples() < maxSamplesPerChunk {
```
davtsur: Won't this be a problem if  there's a single chunk that doesn't reach maxSamplesPerChunk ?

bwplotka: That's why "isNext" check here. We check if it reaches max AND there is next item. 

So if the iterator ends, we close the chunk even though it does not have max chunk items.

PR #1046, opened 2019-04-18
receive: Add retention flag for local tsdb storage

brancz:
## Changes

Add flag to configure retention on receive component.

## Verification

Modified min/max block size, to produce small blocks and watched blocks out of retention to be deleted. Same as #1011, this should have e2e tests. I'll look into that as a follow up. I've thought about adding min/max block duration as hidden flags for testing purposes, but I think I'd prefer to just produce old blocks for upload and retention scenarios in e2e tests.

@bwplotka @domgreen @GiedriusS @povilasv


bwplotka:
That was quick :heart:

PR #1053, opened 2019-04-18
Compactor: remove malformed blocks after delay

mjd95:
## Changes
Introduce a consistency delay in the compactor.  If the compactor sees a block that does not have a `meta.json` in its initial sync, and the block is older than `consistencyDelay`, then the compactor removes it.

Related issues https://github.com/improbable-eng/thanos/issues/621 https://github.com/improbable-eng/thanos/issues/976

## Verification
Manually deleted the `meta.json` files from some blocks in our testing bucket, ran this compactor against them and checked that it deleted the blocks in its inital sync then continued with the rest of the compaction.

### DIFF
--- a/cmd/thanos/compact.go
+++ b/cmd/thanos/compact.go
@@ -82,7 +82,7 @@ func registerCompact(m map[string]setupFunc, app *kingpin.Application, name stri
 
 	objStoreConfig := regCommonObjStoreFlags(cmd, "", true)
 
-	syncDelay := modelDuration(cmd.Flag("sync-delay", "Minimum age of fresh (non-compacted) blocks before they are being processed.").
+	syncDelay := modelDuration(cmd.Flag("sync-delay", "Minimum age of fresh (non-compacted) blocks before they are being processed. Blocks older than this which are malformed will be removed.").
 		Default("30m"))
 
 	retentionRaw := modelDuration(cmd.Flag("retention.resolution-raw", "How long to retain raw samples in bucket. 0d - disables this retention").Default("0d"))
--- a/cmd/thanos/compact.go
+++ b/cmd/thanos/compact.go
@@ -82,7 +82,7 @@ func registerCompact(m map[string]setupFunc, app *kingpin.Application, name stri
 
 	objStoreConfig := regCommonObjStoreFlags(cmd, "", true)
 
-	consistencyDelay := modelDuration(cmd.Flag("consistency-delay", fmt.Sprintf("Minimum age of fresh (non-compacted) blocks before they are being processed. Malformed blocks older than the maximum of consistency-delay and %s will be removed.", compact.MinimumConsistencyDelay)).
+	consistencyDelay := modelDuration(cmd.Flag("consistency-delay", fmt.Sprintf("Minimum age of fresh (non-compacted) blocks before they are being processed. Malformed blocks older than the maximum of consistency-delay and %s will be removed.", compact.MinimumAgeForRemoval)).
 		Default("30m"))
 
 	retentionRaw := modelDuration(cmd.Flag("retention.resolution-raw", "How long to retain raw samples in bucket. 0d - disables this retention").Default("0d"))
--- a/cmd/thanos/compact.go
+++ b/cmd/thanos/compact.go
@@ -82,7 +82,7 @@ func registerCompact(m map[string]setupFunc, app *kingpin.Application, name stri
 
 	objStoreConfig := regCommonObjStoreFlags(cmd, "", true)
 
-	consistencyDelay := modelDuration(cmd.Flag("consistency-delay", fmt.Sprintf("Minimum age of fresh (non-compacted) blocks before they are being processed. Blocks older than this which are malformed will be removed.  The smallest possible value of this flag is %s.", compact.MinimumConsistencyDelay)).
+	consistencyDelay := modelDuration(cmd.Flag("consistency-delay", fmt.Sprintf("Minimum age of fresh (non-compacted) blocks before they are being processed. Malformed blocks older than the maximum of consistency-delay and %s will be removed.", compact.MinimumConsistencyDelay)).
 		Default("30m"))
 
 	retentionRaw := modelDuration(cmd.Flag("retention.resolution-raw", "How long to retain raw samples in bucket. 0d - disables this retention").Default("0d"))
--- a/cmd/thanos/compact.go
+++ b/cmd/thanos/compact.go
@@ -82,7 +82,7 @@ func registerCompact(m map[string]setupFunc, app *kingpin.Application, name stri
 
 	objStoreConfig := regCommonObjStoreFlags(cmd, "", true)
 
-	consistencyDelay := modelDuration(cmd.Flag("consistency-delay", "Minimum age of fresh (non-compacted) blocks before they are being processed. Blocks older than this which are malformed will be removed.").
+	consistencyDelay := modelDuration(cmd.Flag("consistency-delay", fmt.Sprintf("Minimum age of fresh (non-compacted) blocks before they are being processed. Blocks older than this which are malformed will be removed.  The smallest possible value of this flag is %s.", compact.MinimumConsistencyDelay)).
 		Default("30m"))
 
 	retentionRaw := modelDuration(cmd.Flag("retention.resolution-raw", "How long to retain raw samples in bucket. 0d - disables this retention").Default("0d"))
--- a/cmd/thanos/compact.go
+++ b/cmd/thanos/compact.go
@@ -82,7 +82,7 @@ func registerCompact(m map[string]setupFunc, app *kingpin.Application, name stri
 
 	objStoreConfig := regCommonObjStoreFlags(cmd, "", true)
 
-	syncDelay := modelDuration(cmd.Flag("sync-delay", "Minimum age of fresh (non-compacted) blocks before they are being processed. Blocks older than this which are malformed will be removed.").
+	consistencyDelay := modelDuration(cmd.Flag("consistency-delay", "Minimum age of fresh (non-compacted) blocks before they are being processed. Blocks older than this which are malformed will be removed.").
 		Default("30m"))
 
 	retentionRaw := modelDuration(cmd.Flag("retention.resolution-raw", "How long to retain raw samples in bucket. 0d - disables this retention").Default("0d"))
@@ -114,7 +114,7 @@ func registerCompact(m map[string]setupFunc, app *kingpin.Application, name stri
 			*httpAddr,
 			*dataDir,
 			objStoreConfig,
-			time.Duration(*syncDelay),
+			time.Duration(*consistencyDelay),
 			*haltOnError,
 			*acceptMalformedIndex,
 			*wait,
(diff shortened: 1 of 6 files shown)


pkg/compact/compact.go
```diff
@@ -31,6 +31,8 @@ const (
 	ResolutionLevelRaw = ResolutionLevel(downsample.ResLevel0)
 	ResolutionLevel5m  = ResolutionLevel(downsample.ResLevel1)
 	ResolutionLevel1h  = ResolutionLevel(downsample.ResLevel2)
+
+	consistencyDelay = time.Duration(30 * time.Minute)
```
bwplotka: so I feel we have `syncTime`. I like `consistencyTime` name better, so can we just reuse and rename variable?

pkg/compact/compact.go
```diff
@@ -194,6 +197,9 @@ func (c *Syncer) syncMetas(ctx context.Context) error {
 					continue
 				}
 				if err != nil {
+					if removed := c.removeIfMalformed(workCtx, id); removed {
```
bwplotka: all of this is great but look 2 lines above. We already have `blockTooFreshSentinelError` and stuff that checks if is fresh enough! Let's reuse exactly the same thing if we can!

mjd95: I've changed it up so that in either case where the block is too fresh this is reported by `DownloadMeta`.  The logic is slightly different based on whether or not we were able to get a meta for that block, but it's nicer to have only one place doing the freshness check.

mjd95:
Have changed around to re-use `syncDelay`.  It has a default value of 30 minutes which I think is good.  The only danger I guess would be if people have been running the compactor with a very low value?  I wonder if we should still keep a safety check in `removeIfMalformed` anyway?

Discussed offline and won't add anything to the store gateway right now, as it's good to keep the store as read only.

pkg/block/block.go
```diff
@@ -145,3 +145,12 @@ func IsBlockDir(path string) (id ulid.ULID, ok bool) {
 	id, err := ulid.Parse(filepath.Base(path))
 	return id, err == nil
 }
+
+// MetaExists checks whether the meta file exists for the provided block.
+func MetaExists(ctx context.Context, logger log.Logger, bkt objstore.Bucket, id ulid.ULID) (bool, error) {
```
bwplotka: Is it really that useful method?  I would just drop it. It's too shallow

cmd/thanos/compact.go
```diff
@@ -82,7 +82,7 @@ func registerCompact(m map[string]setupFunc, app *kingpin.Application, name stri
 
 	objStoreConfig := regCommonObjStoreFlags(cmd, "", true)
 
-	syncDelay := modelDuration(cmd.Flag("sync-delay", "Minimum age of fresh (non-compacted) blocks before they are being processed.").
+	syncDelay := modelDuration(cmd.Flag("sync-delay", "Minimum age of fresh (non-compacted) blocks before they are being processed. Blocks older than this which are malformed will be removed.").
```
bwplotka: let's rename variable here as well. Wonder if flag name as well? I would do that.

pkg/compact/compact.go
```diff
@@ -271,6 +279,27 @@ func (c *Syncer) downloadMeta(ctx context.Context, id ulid.ULID) (*metadata.Meta
 	return &meta, nil
 }
 
+// removeIfMalformed removes a block from the bucket if that block does not have a meta file.
+// It is the responsibility of the caller to ensure that enough time has passed for the block to become consistent
```
bwplotka:
```suggestion
// It is the responsibility of the caller to ensure that enough time has passed for the block to become consistent.
```

pkg/compact/compact.go
```diff
@@ -271,6 +279,27 @@ func (c *Syncer) downloadMeta(ctx context.Context, id ulid.ULID) (*metadata.Meta
...
+func (c *Syncer) removeIfMalformed(ctx context.Context, id ulid.ULID) bool {
```
bwplotka: let's maybe be more accurate in naming:
```suggestion
func (c *Syncer) removeIfMetaMalformed(ctx context.Context, id ulid.ULID) bool {
```

pkg/compact/compact.go
```diff
@@ -271,6 +279,27 @@ func (c *Syncer) downloadMeta(ctx context.Context, id ulid.ULID) (*metadata.Meta
...
+		level.Warn(c.logger).Log("msg", "failed to check if block is malformed", "block", id)
+		return false
+	}
+	if exists {
+		// Meta exists, block is not malformed.
+		return false
+	}
+
+	if err := block.Delete(ctx, c.bkt, id); err != nil {
+		level.Warn(c.logger).Log("msg", "failed to delete malformed block", "block", id)
```
bwplotka: return false here?

bwplotka: can we put error as well?

pkg/compact/compact.go
```diff
@@ -271,6 +279,27 @@ func (c *Syncer) downloadMeta(ctx context.Context, id ulid.ULID) (*metadata.Meta
 	return &meta, nil
 }
 
+// removeIfMalformed removes a block from the bucket if that block does not have a meta file.
+// It is the responsibility of the caller to ensure that enough time has passed for the block to become consistent
+func (c *Syncer) removeIfMalformed(ctx context.Context, id ulid.ULID) bool {
+	exists, err := block.MetaExists(ctx, c.logger, c.bkt, id)
+	if err != nil {
+		level.Warn(c.logger).Log("msg", "failed to check if block is malformed", "block", id)
```
bwplotka: can we print error here?

pkg/compact/compact.go
```diff
@@ -271,6 +280,32 @@ func (c *Syncer) downloadMeta(ctx context.Context, id ulid.ULID) (*metadata.Meta
...
 
+// removeIfMalformed removes a block from the bucket if that block does not have a meta file.
+// It is the responsibility of the caller to ensure that enough time has passed for the block to become consistent.
+func (c *Syncer) removeIfMetaMalformed(ctx context.Context, id ulid.ULID) bool {
+	metaExists, err := c.bkt.Exists(ctx, path.Join(id.String(), block.MetaFilename))
+	if err != nil {
+		level.Warn(c.logger).Log("msg", "failed to check meta exists for block", "block", id, "err", err)
+		return false
+	}
+	if err != nil {
```
bwplotka: hm this is weird.

cmd/thanos/compact.go
bwplotka: ignore me, see my comment below

bwplotka:
> Have changed around to re-use syncDelay. It has a default value of 30 minutes which I think is good. The only danger I guess would be if people have been running the compactor with a very low value? I wonder if we should still keep a safety check in removeIfMalformed anyway?

Good point. I think we need to define some minimumDelay, Let's imagine you run strongly consistent objstorage like GCS. Even in this case `consistencyDelay=0s` is wrong as sidecars can upload blocks slowly, so it may take some time where the block is partially uploaded and we don't want compactor to remove that, right?

pkg/compact/compact.go
mjd95: oops :) . fixed

mjd95:
Yeah @bwplotka that's the sort of thing I was worried about.  Now that the `--sync-delay` flag is also controlling whether the block is old enough to be deleted or not, we need to be more careful if the flag is set very low.  

I think we should probably have some sort of extra safety check in `removeIfMetaMalformed`.  Either we could force the `--sync-delay` flag value to be `>10m`, or we could just add in an additional check `removeIfMetaMalformed` for some delay Thanos itself sets.  I think I prefer the latter. 

(I changed the flag name BTW.  If we land it like this, it's annoying that the name change is a breaking changed, but it does help to show that the behaviour of the flag has changed.)

bwplotka:
Yes, I think the flag name here is reasonable and it's not heavy used I think. Not sure for the the constraint. I think syncDelay delay minimum value makes sense more

bwplotka:
Okey, let's not fit this into release although I am super sad ): I think this is critial piece. If you have 0 consistency delay it means you will kill your sidecar uploads ): Not fun. No way to separate those from bucket perspecitve as all is in meta. So this means some sort of consistency delay is must have, even for tests. 

Alternatives: 
* Push meta.json first and have another special file for locking (we would rather have native block support)
* Separate blocks to tell which ones are from compactor? Maybe compacted block should have some indication in dir name? or special file?

pkg/compact/compact.go
```diff
@@ -194,6 +198,9 @@ func (c *Syncer) syncMetas(ctx context.Context) error {
 					continue
 				}
 				if err != nil {
+					if removedOrIgnored := c.removeIfMetaMalformed(workCtx, id); removedOrIgnored {
```
bwplotka: Well technically you can do just `if c.removeIfMetaMalformed(workCtx, id)`  here (:

xjewer:
why it removes data by default, can we add a flag to explicitly do removal?

midnightconman:
> why it removes data by default, can we add a flag to explicitly do removal?

Like a debug or dry-run flag?

xjewer:
>Like a debug or dry-run flag?

the opposite, the flag which explicitly tells to remove these blocks, like `--compact.remove-malformed-blocks`

bwplotka:
Very good point @xjewer, we could do that for safety, but we cannot see any current case when we remove important block, not duplicated in other places (even with eventually consistent storage), so would you not remove automatically? If you are worried I am happy to add such flag.

xjewer:
while it's happening https://github.com/improbable-eng/thanos/issues/318#issuecomment-482611807, i'd prefer removing manually instead of losing data, e.g. if I a meta.json is missing, I can get it at least from debug directory

bwplotka:
I cannot see the use case but I am fine with manual option. I would argue if manual should be default though.

Even the issue you mentioned - I doubt it is causing partial uploads as we remove the partial uploaded one immdiately after failure (see: https://github.com/improbable-eng/thanos/blob/e6d5b49bf2d2281f8f2f2d0e853158174a2e541f/pkg/block/block.go#L113) and we do not proceed with removal of old compacted blocks.

xjewer:
> I doubt it is causing partial uploads as we remove the partial uploaded one immdiately after failure

for some reason, there's a 400 error on proxy, but minio (seems to be multipart upload ¯\\\_(ツ)_/¯) doesn't treat it as an error, thus it uploads block with no error and as a consequence there's no cleanUp.

>we do not proceed with removal of old compacted blocks.

wait, I see the opposite 

`Malformed blocks older than the maximum of consistency-delay and %s will be removed`

https://github.com/improbable-eng/thanos/pull/1053/files#diff-921a1164068d3a04217b2d85714b16adR298

bwplotka:
I meant "we do not proceed with removal of old compacted blocks." if the upload was failed (:

bwplotka:
@xjewer let's jump on slack I am confused (: 

> for some reason, there's a 400 error on proxy, but minio (seems to be multipart upload ¯\_(ツ)_/¯) doesn't treat it as an error, thus it uploads block with no error and as a consequence there's no cleanUp.

You mean file was NOT uploaded but we get 200 and success from minio Upload request?

PR #1055, opened 2019-04-18
Disabled gossip by default.

bwplotka:
Since 0.4.0 is last release with gossip available, let's disable it so people will be aware of this.

+ changed small label inconsistency.


drax68:
What should be the value of that flag to reenable gossip? 
false
0
False
F
FALSE
seems to be non valid.

forkbomber:
@drax68  use `--no-cluster.disable`

PR #1056, opened 2019-04-18
Make SafeDelete faster for large blocks; TSDB block repair case

jjneely:
This makes SafeDelete faster and less error prone when dealing with large blocks.  This prevents SafeDelete from re-downloading a block that was just downloaded.

This PR also addresses a TSDB block repair case when duplicate series are detected.

## Changes

* SafeDelete takes another argument to use a pre-existing temporary directory possibly containing a previsouly downloaded TSDB block.
* Duplicate series in a TSDB block are now handled when rewriting a block.

## Verification

I run these patches in production, and have repaired quite a few hundred gig of blocks with these.


pkg/block/index.go
```diff
@@ -495,12 +495,16 @@ func Repair(logger log.Logger, dir string, id ulid.ULID, source metadata.SourceT
...
 	resmeta.Thanos.Source = source    // update source
 
-	if err := rewrite(indexr, chunkr, indexw, chunkw, &resmeta, ignoreChkFns); err != nil {
+	if err := rewrite(logger, indexr, chunkr, indexw, chunkw, &resmeta, ignoreChkFns); err != nil {
 		return resid, errors.Wrap(err, "rewrite block")
 	}
 	if err := metadata.Write(logger, resdir, &resmeta); err != nil {
 		return resid, err
 	}
+	// TSDB may rewrite metadata in bdir
```
jjneely: The TSDB code handily rewrites the block metadata even though we don't write or change read block here.  This nukes the extra metadata Thanos uses.  Write it back.

pkg/block/index.go
```diff
@@ -665,8 +670,16 @@ func rewrite(
 		return labels.Compare(series[i].lset, series[j].lset) < 0
 	})
 
+	lastSet := labels.Labels{}
 	// Build a new TSDB block.
 	for _, s := range series {
+		if labels.Compare(lastSet, s.lset) == 0 {
```
jjneely: The series are sorted based on their labelSet.  What happens if two labelSets are identical?  For one they will be consecutive in the slice.  So if the labelSet we have now is the same as the previous labelSet we have a duplicate series which the TSDB code will produce an error on.  So we just drop that series.

pkg/verifier/safe_delete.go
```diff
@@ -31,23 +35,33 @@ func SafeDelete(ctx context.Context, logger log.Logger, bkt objstore.Bucket, bac
 		return errors.Errorf("%s dir seems to exists in backup bucket. Remove this block manually if you are sure it is safe to do", id)
 	}
 
-	tempdir, err := ioutil.TempDir("", "safe-delete")
-	if err != nil {
-		return err
+	if tempdir == "" {
```
jjneely: If we already have the block, don't download it again.  Blocks can be large.  30GiB or close to it in my environment.

jjneely:
I've tested this several times on my own TSDB blocks.  I've been wondering how to get better tests integrated into the code as unit tests, but I cannot make these TSDB blocks of mine public.  But I've repaired several hundred GiB of data with this.

pkg/verifier/safe_delete.go
```diff
@@ -31,23 +35,33 @@ func SafeDelete(ctx context.Context, logger log.Logger, bkt objstore.Bucket, bac
...
-	if err != nil {
-		return err
-	}
-	defer func() {
-		if err := os.RemoveAll(tempdir); err != nil {
-			level.Warn(logger).Log("msg", "failed to delete dir", "dir", tempdir, "err", err)
+
+	if _, err := os.Stat(dir); err != nil {
+		// TSDB block not already present, download it
+		err = os.Mkdir(dir, 0755)
```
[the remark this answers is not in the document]
jjneely: Fixed, thanks!

pkg/block/index.go
jjneely: Done.  Thank you!

pkg/verifier/safe_delete.go
```diff
@@ -31,23 +35,33 @@ func SafeDelete(ctx context.Context, logger log.Logger, bkt objstore.Bucket, bac
...
+			}
+		}()
 
-	if err := block.Download(ctx, logger, bkt, id, dir); err != nil {
-		return errors.Wrap(err, "download from source")
+		if err := block.Download(ctx, logger, bkt, id, dir); err != nil {
+			return errors.Wrap(err, "download from source")
+		}
+	} else {
+		level.Info(logger).Log("msg", "using previously downloaded tsdb block",
```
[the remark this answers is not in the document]
jjneely: My blocks (and I'm sure others) easily grow above 30GiB when compacted.  So if I need to repair block A, Thanos will download block A, create a repaired block B, upload block B.  So at this point, we've created, successfully, a new block from A and that block B now passes the `VerifyIndex()` test.

At this point we call `SafeDelete()` which downloads block A, uploads block A to the backup bucket, then deletes block A from the main bucket.

This process can be time consuming (and perhaps expensive).  So if blocks are immutable, and the repair process verifies that A is a legitimate block (if malformed) we should be able to skip the redundant download of A and backup the existing downloaded copy.

pkg/verifier/safe_delete.go
```diff
@@ -14,10 +15,13 @@ import (
...
 
-// SafeDelete moves block to backup bucket and if succeeded, removes it from source bucket.
-// It returns error if block dir already exists in backup bucket (blocks should be immutable) or any
-// of the operation fails.
-func SafeDelete(ctx context.Context, logger log.Logger, bkt objstore.Bucket, backupBkt objstore.Bucket, id ulid.ULID) error {
+// SafeDelete moves block to backup bucket and if succeeded, removes it from
+// source bucket.  It returns error if block dir already exists in backup
+// bucket (blocks should be immutable) or any of the operation fails.  If
+// the block has already been downloaded it must exist in tempdir, else it
+// will be downloaded to be copied to the backup bucket.  tempdir set to the
```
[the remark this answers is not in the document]
jjneely: `SafeDelete()` doesn't know the temporary directory where a previous function may have downloaded the block, so we need to pass in a path somehow.  Perhaps rename this argument something like `useExisting`?  Which would mean we use the existing block if found under that path, otherwise a new temporary directory is created and the block is downloaded.

pkg/block/index.go
brancz: Sorry one more thing. As this shouldn't generally happen, let's add a metric so we can monitor when it does happen.

jjneely: The repair process batch job only which makes metrics less useful.  I had some similar comments in https://github.com/improbable-eng/thanos/pull/964

Perhaps I should leave a comment if that were to change?

brancz: Yes let's in that case put a TODO in there

pkg/verifier/safe_delete.go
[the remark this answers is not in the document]
brancz: sorry I meant let's pass two parameters to explicitly describe the intention

pkg/verifier/safe_delete.go
```diff
@@ -31,23 +35,33 @@ func SafeDelete(ctx context.Context, logger log.Logger, bkt objstore.Bucket, bac
...
 	dir := filepath.Join(tempdir, id.String())
-	err = os.Mkdir(dir, 0755)
-	if err != nil {
-		return err
-	}
-	defer func() {
-		if err := os.RemoveAll(tempdir); err != nil {
-			level.Warn(logger).Log("msg", "failed to delete dir", "dir", tempdir, "err", err)
+
+	if _, err := os.Stat(dir); os.IsNotExist(err) {
```
brancz: we're swallowing any error that is not "IsNotExist" here

GiedriusS: I think we should just do here `os.Mkdir` and ignore the error if it is saying that the directory already exists. Otherwise, we will have racy behaviour here (of course, there's barely any chance that someone will run two of these at the same time but still)

brancz: Yeah, good suggestions!

jjneely: I added some additional logic here to handle any other possible error from `os.Stat()`.  

The pre-existence of the directory triggers the code path where we can avoid having to re-download a possibly large TSDB block again.  Races with the `os.Mkdir()` are not anticipated as if the temporary directory isn't supplied `ioutil.TempDir()`is used to generate a value.  When we do supply `tempdir` in `pkg/verifier/index_issue.go` is it also generated with `ioutil.Tempdir()`:

```
tmpdir, err := ioutil.TempDir("", fmt.Sprintf("index-issue-block-%s-", id))
```

pkg/block/index.go
jjneely: TODO added.

pkg/verifier/safe_delete.go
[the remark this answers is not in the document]
jjneely: WIP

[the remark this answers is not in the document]
jjneely: Ok, committed a patch for makes this explicit.  I've tested the repair process with my collection of broken blocks I've started keeping and had the behavior expected.  (No repeated downloads.)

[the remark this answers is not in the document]
jjneely: Had thanos-compact cut a 450GiB TSDB block today which I referenced in https://github.com/improbable-eng/thanos/issues/1152 which is part of the motivation for this.

pkg/verifier/safe_delete.go
```diff
@@ -14,10 +15,14 @@ import (
...
-// of the operation fails.
-func SafeDelete(ctx context.Context, logger log.Logger, bkt objstore.Bucket, backupBkt objstore.Bucket, id ulid.ULID) error {
+// SafeDelete moves a TSDB block to a backup bucket and, on success, removes
+// it from the source bucket.  It returns error if block dir already exists in
+// the backup bucket (blocks should be immutable) or if any of the operations
+// fail.  When useExisting is true do not download the block but use the
+// previously downloaded block found in tempdir to avoid repeated downloads.
+// If tempdir is an empty string a unique temporary directory is created for
+// scratch space, otherwise the given directory is used.
+func SafeDelete(ctx context.Context, logger log.Logger, bkt objstore.Bucket, backupBkt objstore.Bucket, id ulid.ULID, useExisting bool, tempdir string) error {
```
brancz: sorry for nitpicking on this, but can we name this `useExistingTempdir`? I realize that's long, but we'll understand in 3 months what this means that way :)

pkg/verifier/safe_delete.go
```diff
@@ -31,23 +36,45 @@ func SafeDelete(ctx context.Context, logger log.Logger, bkt objstore.Bucket, bac
...
-	defer func() {
-		if err := os.RemoveAll(tempdir); err != nil {
-			level.Warn(logger).Log("msg", "failed to delete dir", "dir", tempdir, "err", err)
-		}
-	}()
 
-	if err := block.Download(ctx, logger, bkt, id, dir); err != nil {
-		return errors.Wrap(err, "download from source")
+	if useExisting {
+		if _, err := os.Stat(filepath.Join(dir, "meta.json")); err != nil {
```
brancz: Is this check really sufficient? Shouldn't we check that all the downloaded content is equal to what is in object storage, ideally by their hash?

pkg/verifier/safe_delete.go
```diff
@@ -31,23 +36,45 @@ func SafeDelete(ctx context.Context, logger log.Logger, bkt objstore.Bucket, bac
 		return errors.Errorf("%s dir seems to exists in backup bucket. Remove this block manually if you are sure it is safe to do", id)
 	}
 
-	tempdir, err := ioutil.TempDir("", "safe-delete")
-	if err != nil {
-		return err
+	if tempdir == "" {
```
brancz: Maybe I wasn't clear, I would expect that when `useExisting` is set, that `tempDir` must be set, so I would expect an error from this function if that is not the case, and this if statement to be combined with the parts below, I think that'll also make the code a lot more easy to understand.

jjneely: Ok, refactored this a bit.  It will return an error if you ask it to use an existing tempdir and do not provide one, and the logic is a little simpler.

pkg/verifier/safe_delete.go
jjneely: There don't appear to be tools for this in the existing code base.  So this seems like a bit of scope creep.  Glad to add additional checks here, but these features sound like another GitHub Issue/PR.

In fact, why isn't there a checksum in meta.json?  It doesn't have to be cryptographically secure, but it would serve to verify that the block was intact.  As we move from "metric data is ephemeral" to "metric data is important long term" that would be quite handy.

pkg/verifier/safe_delete.go
jjneely: Not a problem.  Fixed.

jjneely:
cluster_test.go:129 fails on me again...

brancz:
gossip has been removed, if you rebase I think those flakes should go away :)

brancz:
Sorry for the silence, I'm working off the backlog that accumulated over kubecon, once rebased and CI is green I'm happy to give this another pass

brancz:
The test that kept flaking is gone due to the rebase, that's all :)

jjneely:
Updates on getting this merged?

pkg/block/index.go
```diff
@@ -665,8 +670,22 @@ func rewrite(
...
+		// The TSDB library will throw an error if we add a series with
+		// identical labels as the last series.  This means that we have
+		// discovered a duplicate time series in the old block.  We drop
+		// all duplicate series preserving the first one.
+		// TODO: Add metric to count dropped series if repair becomes a daemon
+		// rather than a batch job.
+		if labels.Compare(lastSet, s.lset) == 0 {
+			level.Warn(logger).Log("msg",
+				"dropping duplicate series in tsdb block found",
+				"labelset", fmt.Sprintf("%s", s.lset),
```
GiedriusS: You should use `s.lset.String()` here. The new meta linter will complain about this :smile:

pkg/block/index.go
GiedriusS: So this is also a bug that you are trying to fix? Perhaps we should add it to the changelog and/or the PR messsage.

pkg/verifier/safe_delete.go
```diff
@@ -14,10 +15,14 @@ import (
...
-// of the operation fails.
-func SafeDelete(ctx context.Context, logger log.Logger, bkt objstore.Bucket, backupBkt objstore.Bucket, id ulid.ULID) error {
+// SafeDelete moves a TSDB block to a backup bucket and, on success, removes
+// it from the source bucket.  It returns error if block dir already exists in
+// the backup bucket (blocks should be immutable) or if any of the operations
+// fail.  When useExistingTempdir is true do not download the block but use the
+// previously downloaded block found in tempdir to avoid repeated downloads.
+// If tempdir is an empty string a unique temporary directory is created for
+// scratch space, otherwise the given directory is used.
+func SafeDelete(ctx context.Context, logger log.Logger, bkt objstore.Bucket, backupBkt objstore.Bucket, id ulid.ULID, useExistingTempdir bool, tempdir string) error {
```
GiedriusS: Couldn't we just get rid of `useExistingTempdir` here and use `tempdir` to control how this function works? It seems like we are making this unnecessarily complex. Perhaps I am missing something.

pkg/verifier/safe_delete.go
```diff
@@ -31,23 +36,48 @@ func SafeDelete(ctx context.Context, logger log.Logger, bkt objstore.Bucket, bac
...
 
-	tempdir, err := ioutil.TempDir("", "safe-delete")
-	if err != nil {
-		return err
-	}
-	dir := filepath.Join(tempdir, id.String())
-	err = os.Mkdir(dir, 0755)
-	if err != nil {
-		return err
+	if useExistingTempdir && tempdir == "" {
```
GiedriusS: It seems like they **both** can be set or unset which proves my point.

pkg/verifier/safe_delete.go
brancz: I felt like giving a particular value of `tempdir` special meaning was not obvious enough. I think I prefer this over tempdir being empty. I do agree that below the if doesn't need the `tempdir == ""` case.

pkg/block/index.go
jjneely: Not a bug, this is part of the initial patches to be able to reuse downloaded blocks.  The TSDB library isn't aware of the additional meta data Thanos provides and was removing it.  Therefore, when a previously downloaded block is reused it would have incomplete metadata without this change.

pkg/verifier/safe_delete.go
brancz: If you feel strongly about this @GiedriusS I'm ok with the other behavior, just feel I will have forgotten the implicit behavior of the empty string in 3 months, hence I wanted it to be more explicit.

pkg/block/index.go
jjneely: Fixed.

pkg/verifier/safe_delete.go
jjneely: I touched up the comments for this function again so that they better describe the behavior of the current code.  No code changes though.

jjneely:
Updates?

pkg/block/index.go
bwplotka: So this is being changed now: https://github.com/prometheus/tsdb/pull/637 so this won't be needed TBH (once we use latest TSDB).

 Also this is true what @GiedriusS said - splitting PRs into focused changes helps a lot. However this change relies on this fix. I would be more explicit in description. This will help us to understand this early on (:

bwplotka: Also missing trailing period in comment.

pkg/block/index.go
```diff
@@ -665,8 +670,22 @@ func rewrite(
 		return labels.Compare(series[i].lset, series[j].lset) < 0
 	})
 
+	lastSet := labels.Labels{}
 	// Build a new TSDB block.
 	for _, s := range series {
+		// The TSDB library will throw an error if we add a series with
```
bwplotka: Can we remove double spaces in this comment?

pkg/block/index.go
bwplotka: Let's add TODO to update TSDB and remove this as it will be not needed in newest TSDB. For now we need this, thanks for fixing this :+1:

pkg/verifier/safe_delete.go
```diff
@@ -14,10 +15,13 @@ import (
 	"github.com/pkg/errors"
 )
 
-// SafeDelete moves block to backup bucket and if succeeded, removes it from source bucket.
-// It returns error if block dir already exists in backup bucket (blocks should be immutable) or any
-// of the operation fails.
-func SafeDelete(ctx context.Context, logger log.Logger, bkt objstore.Bucket, backupBkt objstore.Bucket, id ulid.ULID) error {
+// SafeDelete moves a TSDB block to a backup bucket and, on success, removes
```
bwplotka:
```suggestion
// SafeDelete moves a TSDB block to a backup bucket and on success removes
```

pkg/verifier/safe_delete.go
```diff
@@ -14,10 +15,13 @@ import (
...
+// it from the source bucket.  It returns error if block dir already exists in
```
bwplotka:
```suggestion
// it from the source bucket. It returns error if block dir already exists in
```

pkg/verifier/safe_delete.go
```diff
@@ -14,10 +15,13 @@ import (
...
+// the backup bucket (blocks should be immutable) or if any of the operations
+// fail.  When useExistingTempdir is true do not download the block but use the
```
bwplotka:
```suggestion
// fail. When useExistingTempdir is true do not download the block but use the
```

pkg/verifier/safe_delete.go
bwplotka: Hm, I feel strongly about this - boolean in random place of the functions IMO smells bad. Same with 2 flags controlling same thing in very strict way.Essentially it does not make sense to set `useExistingTempdir = true` and `tempdir = ""` so we need to validate that etc. Interface is broader then needed.

Also if userExistingTempdir = true and tempdir is specified you literally don't use `bkt` so it's even worse in terms of readability and surprises ):

I would propose: 
*  `BackupAndDelete(ctx context.Context, logger log.Logger, bkt objstore.Bucket, backupBkt objstore.Bucket, id ulid.ULID)` as it was.
* `BackupAndDeleteFromLocal(ctx context.Context, logger log.Logger, bdir string, backupBkt objstore.Bucket)`

This will ensure separation of concerns AND in fact `BackupAndDelete` can use `BackupAndDeleteFromLocal` internally. Right now what I can see is 2 function baked in one function (:

pkg/verifier/safe_delete.go
bwplotka: I proposed different approach - splitting this into 2 function which will solve this issue (:

bwplotka:
And sorry for massive delay in review!

pkg/block/index.go
jjneely: Todo added, thanks!

jjneely: Done.

pkg/block/index.go
jjneely: Old habits die hard...fixed.

pkg/block/index.go
```diff
@@ -495,12 +495,17 @@ func Repair(logger log.Logger, dir string, id ulid.ULID, source metadata.SourceT
...
 
-	if err := rewrite(indexr, chunkr, indexw, chunkw, &resmeta, ignoreChkFns); err != nil {
+	if err := rewrite(logger, indexr, chunkr, indexw, chunkw, &resmeta, ignoreChkFns); err != nil {
 		return resid, errors.Wrap(err, "rewrite block")
 	}
 	if err := metadata.Write(logger, resdir, &resmeta); err != nil {
 		return resid, err
 	}
+	// TSDB may rewrite metadata in bdir.
+	// TODO: This is not needed in newer TSDB code.
```
bwplotka: PR number would be nice for easier trackngm but not a blocker (:

pkg/block/index.go
bwplotka: Can I ask where is this habit from? Sorry for my ignorance - super curious (:

pkg/verifier/safe_delete.go
brancz: Yeah I like @bwplotka's suggestion best as well.

pkg/block/index.go
jjneely: Lol...  Search google for "double space after period" you'll learn more than you bargained for.  Like, my age.  :-)

pkg/verifier/safe_delete.go
jjneely: This is now committed.  The function signatures didn't quite work out as we would have liked, but I totally aggree, this feels much, much nicer.

pkg/block/index.go
jjneely: Added

pkg/block/index.go
FUSAKLA: TIL never heard of this either. But actually the rule could apply here since code formatting is using monospaced fonts mostly :)

pkg/verifier/safe_delete.go
```diff
@@ -46,15 +53,56 @@ func SafeDelete(ctx context.Context, logger log.Logger, bkt objstore.Bucket, bac
...
+	// Does this TSDB block exist in backupBkt already?
+	found, err := TSDBBlockExistsInBucket(ctx, backupBkt, id)
+	if err != nil {
+		return err
+	}
+	if found {
+		return errors.Errorf("%s dir seems to exists in backup bucket. Remove this block manually if you are sure it is safe to do", id)
+	}
+
+	// Now we call backupAndDeleteDownloaded() to handle the work.
```
bwplotka: I think this comment is not needed.

pkg/verifier/safe_delete.go
```diff
@@ -46,15 +53,56 @@ func SafeDelete(ctx context.Context, logger log.Logger, bkt objstore.Bucket, bac
...
+	return backupAndDeleteDownloaded(ctx, logger, bdir, bkt, backupBkt, id)
+}
+
+// backupAndDeleteDownloaded is a helper function that uploads a TSDB block
+// found on disk to the backup bucket and, on success, removes the TSDB block
+// from the source bucket.
+func backupAndDeleteDownloaded(ctx context.Context, logger log.Logger, bdir string, bkt, backupBkt objstore.Bucket, id ulid.ULID) error {
```
bwplotka: Sorry for being picky but there is some improvement we can make here.

I like `BackupAndDelete` vs `BackupAndDeleteDownloaded`, but it's not clear what's the difference between `backupAndDeleteDownloaded`  and `BackupAndDeleteDownloaded` is. Sure they are heavily commented, but comment should have less priority than naming. I think here the proper naming and small possible signature of function would do the work.

So what about:
```suggestion
func backupDownloaded(ctx context.Context, logger log.Logger, bdir string, bktToRemove, backupBkt objstore.Bucket, id ulid.ULID) error {
```
and then if no error do on the caller in both `BackupAndDeleteDownloaded` and `BackupAndDelete` just.
```
// Block uploaded, so we are ok to remove from src bucket.
if err := block.Delete(ctx, bkt, id); err != nil {
	return errors.Wrap(err, "delete from source")
}
```
This will make the signature of  `BackupAndDeleteDownloaded`  and `backupDowloaded` smaller (thus reducing complexity) as you can drop `bkt` argument. 

What do you think? (:

jjneely: Adjusted as requested, and went through another round of testing.

bwplotka:
I think netlify issue is because of not rebased PR - nothing changed for netlify here so merging.

jjneely:
Awesome, thanks!

Issue #1034, opened 2019-04-15
store: Selector labels from store


bwplotka:
AC:
 * Store gateway has selector labels similar as Querier has (`--selector)` which allows to specify some labels that will part of Announced labels (through Info) but will be manually specified.
 * Document this feature better.

## Why?

This will make Store GW smart sharding better. So we can have kind of `external labels` querier can use for initial filtering. Problem right now is that it's not easy for store to deduce those labels from the storage as it changes frequently. Same issue is for federated Querier, although for Querier adding such logic (propagate all announced labels from underlying StoreAPIs might be bit easier)  

cc @devnev @domgreen

bwplotka:
https://github.com/thanos-io/thanos/pull/1501

bwplotka:
We can now shard Store Gateway. The stable version of this should be released with v0.9.0

PR #1059, opened 2019-04-19
Store: selector labels from store

jojohappy:
Signed-off-by: jojohappy <sarahdj0917@gmail.com>

## Changes
Implements: #1034 

Add `--selector` labels for store gateway.

## Verification
To see the store page to verify the announced labels field.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -19,7 +19,7 @@ We use *breaking* word for marking changes that are not backward compatible (rel
 - [#1358](https://github.com/thanos-io/thanos/pull/1358) Added `part_size` configuration option for HTTP multipart requests minimum part size for S3 storage type
 - [#1363](https://github.com/thanos-io/thanos/pull/1363) Thanos Receive now exposes `thanos_receive_hashring_nodes` and `thanos_receive_hashring_tenants` metrics to monitor status of hash-rings
 - [#1395](https://github.com/thanos-io/thanos/pull/1395) Added `/-/ready` and `/-/healthy` endpoints to Thanos sidecar.
-- [#1059](https://github.com/improbable-eng/thanos/pull/1059) Store: selector labels from store.
+- [#1059](https://github.com/improbable-eng/thanos/pull/1059) Store: Added label selecting. This allows store gateway to serve only certain TSDB blocks from object store.
 
 ### Changed
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -19,6 +19,7 @@ We use *breaking* word for marking changes that are not backward compatible (rel
 - [#1358](https://github.com/thanos-io/thanos/pull/1358) Added `part_size` configuration option for HTTP multipart requests minimum part size for S3 storage type
 - [#1363](https://github.com/thanos-io/thanos/pull/1363) Thanos Receive now exposes `thanos_receive_hashring_nodes` and `thanos_receive_hashring_tenants` metrics to monitor status of hash-rings
 - [#1395](https://github.com/thanos-io/thanos/pull/1395) Added `/-/ready` and `/-/healthy` endpoints to Thanos sidecar.
+- [#1059](https://github.com/improbable-eng/thanos/pull/1059) Store: selector labels from store.
 
 ### Changed
 
--- a/cmd/thanos/store.go
+++ b/cmd/thanos/store.go
@@ -11,7 +11,7 @@ import (
 	opentracing "github.com/opentracing/opentracing-go"
 	"github.com/pkg/errors"
 	"github.com/prometheus/client_golang/prometheus"
-	"github.com/prometheus/tsdb/labels"
+	"github.com/prometheus/prometheus/tsdb/labels"
 	"github.com/thanos-io/thanos/pkg/model"
 	"github.com/thanos-io/thanos/pkg/objstore/client"
 	"github.com/thanos-io/thanos/pkg/runutil"
@@ -57,7 +57,7 @@ func registerStore(m map[string]setupFunc, app *kingpin.Application, name string
 	maxTime := model.TimeOrDuration(cmd.Flag("max-time", "End of time range limit to serve. Thanos Store serves only blocks, which happened eariler than this value. Option can be a constant time in RFC3339 format or time duration relative to current time, such as -1d or 2h45m. Valid duration units are ms, s, m, h, d, w, y.").
 		Default("9999-12-31T23:59:59Z"))
 
-	selectorLabels := cmd.Flag("selector-label", "Store Gateway selector labels that will be exposed in info endpoint (repeated).").
+	selectorLabels := cmd.Flag("selector-label", "Query selector labels that will be exposed in info endpoint (repeated).").
 		PlaceHolder("<name>=\"<value>\"").Strings()
 
 	m[name] = func(g *run.Group, logger log.Logger, reg *prometheus.Registry, tracer opentracing.Tracer, debugLogging bool) error {
--- a/cmd/thanos/store.go
+++ b/cmd/thanos/store.go
@@ -66,6 +66,11 @@ func registerStore(m map[string]setupFunc, app *kingpin.Application, name string
 				minTime, maxTime)
 		}
 
+		selectorLset, err := parseFlagLabels(*selectorLabels)
+		if err != nil {
+			return errors.Wrap(err, "parse selector labels")
+		}
+
 		return runStore(g,
 			logger,
 			reg,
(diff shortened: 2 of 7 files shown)


pkg/store/prometheus.go
```diff
@@ -290,6 +290,32 @@ func labelsMatches(lset labels.Labels, ms []storepb.LabelMatcher) (bool, []store
...
+		v := lset.Get(tm.Name())
+		if v == "" {
+			continue
+		}
+
+		if !tm.Matches(v) {
+			return false, nil, nil
+		}
+
+		tms = append(tms[:i], tms[i+1:]...)
```
jojohappy: I have a question here. Why do we remove the match label from request `LabelMatcher`? I follow the implemete of the `labelsMatches` method above.

bwplotka:
@jojohappy should we get back to this? Sorry for not reviewing this for some time - but make it non-draft - less chances the review will be postponed.

I think we all agree to this so let's implement this. Thanks for helping!

cmd/thanos/store.go
```diff
@@ -50,11 +51,20 @@ func registerStore(m map[string]setupFunc, app *kingpin.Application, name string
 	blockSyncConcurrency := cmd.Flag("block-sync-concurrency", "Number of goroutines to use when syncing blocks from object storage.").
 		Default("20").Int()
 
+	selectorLabels := cmd.Flag("selector-label", "Store Gateway selector labels that will be exposed in info endpoint (repeated).").
```
bwplotka: Let's make sure this is the same, consistent for `receive` and maybe `rule`? Querier either dropped it already or will drop as we aggregate labels there from storeAPIs.

pkg/store/prometheus.go
```diff
@@ -298,6 +298,32 @@ func labelsMatches(lset labels.Labels, ms []storepb.LabelMatcher) (bool, []store
 	return true, newMatcher, nil
 }
 
+func labelsMatchesAndTranslate(lset labels.Labels, ms []storepb.LabelMatcher) (bool, []labels.Matcher, error) {
```
bwplotka: This might work after few nits, but let's focus on *reusing* stuff. We do matching almost on every component. We for sure can figure out some helper/abstraction for this somehow critical path. (: More code reused = potentially fewer bugs and less code to read. What do you think?

pkg/store/prometheus.go
bwplotka: Sorry what do you mean remove? You mean `continue` part on line 315? This feels wrong. It's bit different to what we have here: https://github.com/thanos-io/thanos/blob/d02488d9be253223393515a0f9c1fce5318ca3af/pkg/store/prometheus.go#L311

That's why.. can we have just one function for matching + selector labels? (:

pkg/store/prometheus.go
jojohappy: Totally agree with you. I have delete the method and reuse the `matchesExternalLabels` to check the labels matcher.

pkg/store/prometheus.go
jojohappy: Re-reading the code, I'm sorry that I misunderstood the `external_labels` that those should not be in the `storepb.LabelMatcher`, because those would be added to any series. Here the `selector-labels` are as same as the `external_labels` used by store gateway sharding selector. We assume that all of series in the bucket queried by `store` component have the `selector-labels` pairs.

I have reused the `matchesExternalLabels` function to check `selector-labels`, and do `translateMatcher` again to get the new `labels.Matcher`.

jojohappy:
@bwplotka I have updated. PTAL! After that, I will try to add some testings.

CHANGELOG.md
```diff
@@ -15,6 +15,7 @@ We use *breaking* word for marking changes that are not backward compatible (rel
 
 - [#1358](https://github.com/thanos-io/thanos/pull/1358) Added `part_size` configuration option for HTTP multipart requests minimum part size for S3 storage type
 - [#1363](https://github.com/thanos-io/thanos/pull/1363) Thanos Receive now exposes `thanos_receive_hashring_nodes` and `thanos_receive_hashring_tenants` metrics to monitor status of hash-rings
+- [#1059](https://github.com/improbable-eng/thanos/pull/1059) Store: selector labels from store.
```
bwplotka:
```suggestion
- [#1059](https://github.com/improbable-eng/thanos/pull/1059) Store: Added label selecting. This allows store gateway to serve only certain TSDB blocks from object store. 
```

cmd/thanos/store.go
```diff
@@ -49,7 +50,15 @@ func registerStore(m map[string]setupFunc, app *kingpin.Application, name string
 	blockSyncConcurrency := cmd.Flag("block-sync-concurrency", "Number of goroutines to use when syncing blocks from object storage.").
 		Default("20").Int()
 
+	selectorLabels := cmd.Flag("store.selector-label", "Store Gateway selector labels that will be exposed in info endpoint (repeated).").
```
bwplotka: hm we have 
```
```

Maybe worth to be consistent with naming and remove `store` prefix?  essentially it has not much to do with store API for example.

pkg/store/bucket.go
```diff
@@ -474,12 +477,19 @@ func (s *BucketStore) TimeRange() (mint, maxt int64) {
...
+	infoResp := &storepb.InfoResponse{
 		StoreType: component.Store.ToProto(),
 		MinTime:   mint,
 		MaxTime:   maxt,
-	}, nil
+		Labels:    make([]storepb.Label, 0, len(s.selectorLabels)),
+	}
+
+	for _, lset := range s.selectorLabels {
+		infoResp.Labels = append(infoResp.Labels, storepb.Label{Name: lset.Name, Value: lset.Value})
```
bwplotka: Before this change we need to adjust Querier logic: https://github.com/thanos-io/thanos/blob/865d5ec710b5651ead08a7dc73864a6f5dbeeaa9/pkg/query/storeset.go#L248

Otherwise, sidecar + store gateway will be assumed a "duplicate" in Querier. Let's think about how we can fix this.

The storeAPI duplication detection logic is embedded in Querier to make sure user doesn't put many sidecars that upload blocks with the same external labels (common bug). I think we might want to change `storeset` behavior to look on StoreAPI  [`InfoResponse.StoreType`](https://github.com/thanos-io/thanos/blob/a460fa613d79877ee0b927650c9314cfb62fb010/pkg/store/storepb/rpc.pb.go#L34)  

I only don't like the fact that for Querier all of this is no longer `some StoreAPI` but actually `Sidecar, Store...` 

I might think we actually would need to change StoreType to `Producer, Browser` as that would be a good abstraction over different implementations. 

Thoughts @povilasv @GiedriusS @brancz ?

brancz: Am I missing something or shouldn't this just be the set of all labelsets seen in all blocks?

bwplotka: Can you elaborate?

Or better: Let's maybe quickly write up some proposal as we want consistent thing everywhere we do selectors + maybe selecting prefixes as well.

cc @jojohappy

bwplotka:
I think we need this, but we need something we discussed here: https://github.com/thanos-io/thanos/pull/1245#issuecomment-501669550 

I think we need short proposal maybe (?)

bwplotka:
@jojohappy are you happy to close this in the meantime to stop confusion? I will start some PR with the design it would be also awesome if you would join us in making this if you want (:

jojohappy:
> are you happy to close this in the meantime to stop confusion?

That's OK.

>  I will start some PR with the design it would be also awesome if you would join us in making this if you want (:

Yes, I'd like to.

bwplotka:
https://github.com/thanos-io/thanos/pull/1501

cmd/thanos/store.go
```diff
@@ -56,12 +57,20 @@ func registerStore(m map[string]setupFunc, app *kingpin.Application, name string
 	maxTime := model.TimeOrDuration(cmd.Flag("max-time", "End of time range limit to serve. Thanos Store serves only blocks, which happened eariler than this value. Option can be a constant time in RFC3339 format or time duration relative to current time, such as -1d or 2h45m. Valid duration units are ms, s, m, h, d, w, y.").
 		Default("9999-12-31T23:59:59Z"))
 
+	selectorLabels := cmd.Flag("selector-label", "Query selector labels that will be exposed in info endpoint (repeated).").
```
bwplotka: On each component that works on the object storage (e.g Store GW and Compactor), add `--selector.relabel-config` (and corresponding `--selector.relabel-config-file`) as in proposal

bwplotka: On top of this we want to make sure Info() endpoint has exact labels of blocks that we shard.

Issue #1065, opened 2019-04-22
compactor: downsampler dont generate meta.json


sasah:
level=info ts=2019-04-22T12:40:23.651280397Z caller=downsample.go:239 msg="downsampled block" from=01D9284J4BDB9SV9KV74ADMP32 to=01D92GS3GDTH2Q4WRJ46T22HMA duration=3m33.154079955s
level=error ts=2019-04-22T12:40:35.549001957Z caller=main.go:182 msg="running command failed" err="downsampling failed: downsampling to 5 min: upload downsampled block 01D92GS3GDTH2Q4WRJ46T22HMA: read meta: open data/01D92GS3GDTH2Q4WRJ46T22HMA/meta.json: no such file or directory"

sh-4.2$ cd data
sh-4.2$ ls -laR
.:
total 0
drwxr-xr-x. 4 1000620000 root 74 Apr 22 15:36 .
drwxrwxrwt. 1 root       root 18 Apr 22 15:36 ..
drwxr-xr-x. 3 1000620000 root 74 Apr 22 15:36 01D9284J4BDB9SV9KV74ADMP32
drwxr-xr-x. 3 1000620000 root 33 Apr 22 15:36 01D92GS3GDTH2Q4WRJ46T22HMA

./01D9284J4BDB9SV9KV74ADMP32:
total 1484752
drwxr-xr-x. 3 1000620000 root         74 Apr 22 15:36 .
drwxr-xr-x. 4 1000620000 root         74 Apr 22 15:36 ..
drwxr-xr-x. 2 1000620000 root         34 Apr 22 15:36 chunks
-rw-r--r--. 1 1000620000 root 1408701585 Apr 22 15:36 index
-rw-r--r--. 1 1000620000 root  111669428 Apr 22 15:36 index.cache.json
-rw-r--r--. 1 1000620000 root       4140 Apr 22 15:36 meta.json

./01D9284J4BDB9SV9KV74ADMP32/chunks:
total 590160
drwxr-xr-x. 2 1000620000 root        34 Apr 22 15:36 .
drwxr-xr-x. 3 1000620000 root        74 Apr 22 15:36 ..
-rw-r--r--. 1 1000620000 root 536866757 Apr 22 15:36 000001
-rw-r--r--. 1 1000620000 root  67455797 Apr 22 15:36 000002

./01D92GS3GDTH2Q4WRJ46T22HMA:
total 1351092
[...]

PR #1068, opened 2019-04-23
Test that metadata is written and readable after downsampling #1065

shivas:
Additional check to downsample test, checking that meta.json file is readable in a new block directory

## Changes

Trying to read meta.json file after downsample complete

## Verification

This test will fail until issue #1065 is fixed


bwplotka:
Awesome, thanks, will use this.

GiedriusS:
Added in https://github.com/improbable-eng/thanos/pull/1070. Thanks for this! :+1:

PR #1070, opened 2019-04-23
 Fixed Downsampling process; Fixed `runutil.CloseAndCaptureErr`

bwplotka:
Fixes https://github.com/improbable-eng/thanos/issues/1065

Root cause:
* runutil defered capture error function was not returning error properly so unit tests were passing, even though there was bug
* streamed block write index cache requires index file which was not closed (saved) properly yet. Closers need to be closed to perform this.

PTAL @xjewer @shivas

@shivas included your PR https://github.com/improbable-eng/thanos/pull/1068/files here.

### DIFF
--- a/pkg/compact/downsample/streamed_block_writer.go
+++ b/pkg/compact/downsample/streamed_block_writer.go
@@ -167,26 +167,20 @@ func (w *streamedBlockWriter) Close() error {
 	if w.finalized {
 		return nil
 	}
-
-	var merr tsdb.MultiError
 	w.finalized = true
 
-	// Finalise data block only if there wasn't any internal errors.
-	if !w.ignoreFinalize {
-		merr.Add(w.finalize())
+	merr := tsdb.MultiError{}
+
+	if w.ignoreFinalize {
+		// Close open file descriptors anyway.
+		for _, cl := range w.closers {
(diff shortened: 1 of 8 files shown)


pkg/compact/downsample/streamed_block_writer.go
```diff
@@ -207,8 +209,14 @@ func (w *streamedBlockWriter) finalize() error {
 		return errors.Wrap(err, "sync blockDir")
 	}
 
+	if err := merr.Err(); err != nil {
```
xjewer: it's better to put it before [196](https://github.com/improbable-eng/thanos/pull/1070/files#diff-059ebb3a930d53936a164ef53a79d7f0R196) line to avoid wasting time/resources on writing index cache

xjewer:
>Closers need to be closed to perform this.

right, missed this dependency 😞

pkg/compact/downsample/streamed_block_writer.go
bwplotka: You are right. Do you have time for small PR on this? (:

bwplotka:
@xjewer No worries, the main problem is our capture function for closers was not propagating error properly so our unit tests missed this. Now we should be able to catch those error early :+1:

PR #1073, opened 2019-04-23
Ensured index cache is best effort, refactored tests, validated edge cases.

bwplotka:
Fixes https://github.com/improbable-eng/thanos/issues/651

cc @mjd95


pkg/store/bucket.go
```diff
@@ -225,10 +226,18 @@ func NewBucketStore(
 		return nil, errors.Errorf("max concurrency value cannot be lower than 0 (got %v)", maxConcurrent)
 	}
 
-	indexCache, err := newIndexCache(reg, indexCacheSizeBytes)
+	// TODO(bwplotka): Add as a flag?
+	maxItemSizeBytes := uint64(float64(indexCacheSizeBytes) * 0.5)
```
bwplotka: Assumption: Max item size should be not bigget than 50% of Overall size. I think it's generous already.

pkg/store/bucket.go
```diff
@@ -225,10 +226,18 @@ func NewBucketStore(
...
+
+	indexCache, err := storecache.NewIndexCache(logger, reg, storecache.Opts{
+		SetTimeout:       300 * time.Millisecond, // TODO(bwplotka): Add as a flag?
```
bwplotka: arbitrary number, should be enough for typical setup?

pkg/store/cache/cache.go
```diff
@@ -0,0 +1,335 @@
...
+		reg.MustRegister(prometheus.NewGaugeFunc(prometheus.GaugeOpts{
+			Name: "thanos_store_index_cache_max_item_size_bytes",
+			Help: "Maximum number of bytes for single entry to be held in the index cache.",
+		}, func() float64 {
+			return float64(c.maxItemSizeBytes)
+		}))
+		reg.MustRegister(c.requests, c.hits, c.added, c.evicted, c.current, c.currentSize, c.overflow, c.cancelled)
+	}
+
+	if opts.MaxSizeBytes > 1e12 {
```
devnev: Isn't the `1e12` the maximum number of _items_ in the LRU, unrelated to their size?

bwplotka: You are right, changing the comment (and removing this check)

pkg/store/bucket.go
devnev: `maxItemsSizeBytes = indexCacheSizeBytes / 2`

bwplotka: lol

bwplotka:
PTAL @devnev

pkg/store/cache/cache.go
```diff
@@ -0,0 +1,325 @@
...
+		return cacheTypeSeries
+	}
+	return "<unknown>"
+}
+
+func (c cacheItem) size() uint64 {
+	switch k := c.key.(type) {
+	case cacheKeyPostings:
+		// ULID + 2 slice headers + number of chars in value and name.
+		return sliceHeaderSize + 2*16 + uint64(len(k.Value)+len(k.Name))
```
devnev: Comment and code don't quite match up ("ulid + 2 slice headers" vs `sliceHeaderSize + 2*16`).

pkg/store/cache/cache.go
```diff
@@ -0,0 +1,325 @@
...
+type cacheKeySeries uint64
+
+type IndexCache struct {
+	mtx sync.Mutex
+
+	logger           log.Logger
+	lru              *lru.LRU
+	maxSizeBytes     uint64
+	maxItemSizeBytes uint64
+	setTimeout       time.Duration
```
devnev: `setTimeout` unused, remove

pkg/store/cache/cache.go
```diff
@@ -0,0 +1,325 @@
...
+		"maxSizeBytes", c.maxSizeBytes,
+		"setTimeout", c.setTimeout.String(),
+		"maxItems", "math.MaxInt64",
+	)
+	return c, nil
+}
+
+func (c *IndexCache) onEvict(key, val interface{}) {
+	k := key.(cacheItem).keyType()
+	entrySize := sliceHeaderSize + uint64(len(val.([]byte)))
```
devnev: Duplicating the entrySize calculation everywhere seems dubious. It's already inconsistent, sometimes using a literal `16` and sometimes `sliceHeaderSize`.

pkg/store/cache/cache.go
```diff
@@ -0,0 +1,325 @@
...
+)
+
+const (
+	cacheTypePostings string = "Postings"
+	cacheTypeSeries   string = "Series"
+
+	sliceHeaderSize = 16
+)
+
+type cacheItem struct {
```
devnev: Reading the code is made somewhat more confusing by this being called `cacheItem` when it is used as the cache key. WDYT about renaming to `cacheKey`?

pkg/store/cache/cache.go
bwplotka: sorry, might miss some places ):

pkg/store/cache/cache.go
bwplotka: yes!

pkg/store/cache/cache.go
bwplotka: what do you mean duplicating though?

devnev: There are three locations where you calculate `entrySize`, all of which must do the same calculation for the cache to work correctly. It might be worth factoring this out, although I guess it might be good enough as is.

Issue #448, opened 2018-07-27
store: Store gateway consuming lots of memory / OOMing


felipejfc:
**Thanos, Prometheus and Golang version used**
thanos v0.1.0rc2

**What happened**
Thanos-store is consuming 50gb of memory during startup

**What you expected to happen**
Thanos-store does not consume so much memory for starting up

**Full logs to relevant components**
store:
```
level=debug ts=2018-07-27T15:51:21.415788856Z caller=cluster.go:132 component=cluster msg="resolved peers to following addresses" peers=100.96.232.51:10900,100.99.70.149:10900,100.110.182.241:10900,100.126.12.148:10900
level=debug ts=2018-07-27T15:51:21.416254389Z caller=store.go:112 msg="initializing bucket store"
level=warn ts=2018-07-27T15:52:05.28837034Z caller=bucket.go:240 msg="loading block failed" id=01CKE41VDSJMSAJMN6N6K8SABE err="new bucket block: load index cache: download index file: copy object to file: write /var/thanos/store/01CKE41VDSJMSAJMN6N6K8SABE/index: cannot allocate memory"
level=warn ts=2018-07-27T15:52:05.293692332Z caller=bucket.go:240 msg="loading block failed" id=01CKE41VE4XXTN9N55YPCJSPP2 err="new bucket block: load index cache: download index file: copy object to file: write /var/thanos/store/01CKE41VE4XXTN9N55YPCJSPP2/index: cannot allocate memory"
```

**Anything else we need to know**
Some time after initialization the ram usage goes down to normal levels, something around 8Gb

Another thing that's happening is that my thanos-compactor consumer way too much ram memory as well, the last time it ran, it used up to 60Gb of memory.

**I run store with this args:**

```
      containers:
      - args:
        - store
        - --log.level=debug
        - --tsdb.path=/var/thanos/store
        - --s3.endpoint=s3.amazonaws.com
        - --s3.access-key=xxx
        - --s3.bucket=xxx
        - --cluster.peers=thanos-peers.monitoring.svc.cluster.local:10900
        - --index-cache-size=2GB
        - --chunk-pool-size=8GB
```
**Environment**:
- OS (e.g. from /etc/os-release): kubernetes running on debian

felipejfc:
hi @Bplotka 
I do actually, it also uses tons of memory (~60Gb) in the last run, is this normal?

bwplotka:
> it also uses tons of memory (~60Gb) in the last run, is this normal?
[...]

PR #1077, opened 2019-04-24
Store: Add time & duration based partitioning

povilasv:
## Changes

V2 of https://github.com/improbable-eng/thanos/pull/957/files

Doesn't involve any deletions & adds support for durations

https://github.com/improbable-eng/thanos/issues/814

## Verification

Multiple tests added.

### Validated mintime > maxtime behaviour

```
./thanos store --log.level=debug --min-block-start-time=2020-01-01T00:00:00Z --max-block-start-time=2019-01-01T23:59:59Z --objstore.config-file=$UBUNTU_HOME/thanos/dev-telecom.yaml
level=info ts=2019-05-24T12:07:19.750115246Z caller=flags.go:87 msg="gossip is disabled"
store command failed: invalid argument: --min-block-start-time '2020-01-01 00:00:00 +0000 UTC' can't be greater than --max-block-start-time  '2019-01-01 23:59:59 +0000 UTC'
```

```
./thanos store --log.level=debug --min-block-start-time=-10d --max-block-start-time=-15d --objstore.config-file=$UBUNTU_HOME/thanos/dev-telecom.yaml
level=info ts=2019-05-24T12:08:18.891954647Z caller=flags.go:87 msg="gossip is disabled"
store command failed: invalid argument: --min-block-start-time '-10d' can't be greater than --max-block-start-time  '-15d'
```

#### validated blocks loaded

```
./thanos store --log.level=debug --min-block-start-time=-10d --max-block-start-time=-5d --objstore.config-file=$UBUNTU_HOME/thanos/dev-telecom.yaml
```

Checked the metrics:

```
# HELP thanos_bucket_store_block_loads_total Total number of remote block loading attempts.
# TYPE thanos_bucket_store_block_loads_total counter
thanos_bucket_store_block_loads_total 14.0
# HELP thanos_bucket_store_blocks_loaded Number of currently loaded blocks.
# TYPE thanos_bucket_store_blocks_loaded gauge
thanos_bucket_store_blocks_loaded 7.0
```

Actually ran this in dev.


pkg/store/bucket.go
```diff
@@ -314,6 +321,17 @@ func (s *BucketStore) SyncBlocks(ctx context.Context) error {
 		if err != nil {
 			return nil
 		}
+
+		inRange, err := s.isBlockInMinMaxRange(ctx, id)
```
brancz: I can see us adding further methods of selection in the future, what do you think of wrapping this in a function like "keepBlock"?

povilasv: Sure SGTM, but maybe let's do this after we add more filters?

povilasv: I think we would probably end up with some kind of config file for store for specifying external labels and block time ranges. So this would be refactored to something more generic

brancz: that works for me as well

cmd/thanos/store.go
```diff
@@ -50,11 +50,23 @@ func registerStore(m map[string]setupFunc, app *kingpin.Application, name string
 	blockSyncConcurrency := cmd.Flag("block-sync-concurrency", "Number of goroutines to use when syncing blocks from object storage.").
 		Default("20").Int()
 
+	minTime := store.TimeOrDuration(cmd.Flag("min-time", "Start of time range limit to serve. Option can be a constant time in RFC3339 format or time duration relative to current time, such as -1.5d or 2h45m. Valid duration units are ms, s, m, h, d, w, y.").
+		Default("0000-01-01T00:00:00Z"))
+
+	maxTime := store.TimeOrDuration(cmd.Flag("max-time", "End of time range limit to serve. Option can be a constant time in RFC3339 format or time duration relative to current time, such as -1.5d or 2h45m. Valid duration units are ms, s, m, h, d, w, y.").
```
brancz: This should be more explicit whether the exact timestamp would be included or not. Maybe even an example in the docs of what a sane setup would be.

povilasv: :+1:

bwplotka: Agree, some short docs would be nice.

cmd/thanos/store.go
```diff
@@ -50,11 +50,23 @@ func registerStore(m map[string]setupFunc, app *kingpin.Application, name string
...
+		Default("9999-12-31T23:59:59Z"))
+
 	m[name] = func(g *run.Group, logger log.Logger, reg *prometheus.Registry, tracer opentracing.Tracer, debugLogging bool) error {
 		peer, err := newPeerFn(logger, reg, false, "", false)
 		if err != nil {
 			return errors.Wrap(err, "new cluster peer")
 		}
+
+		// Sanity check minTime
```
bwplotka: missing trailing period

cmd/thanos/store.go
```diff
@@ -50,11 +50,23 @@ func registerStore(m map[string]setupFunc, app *kingpin.Application, name string
...
+		if minTime.PrometheusTimestamp() > maxTime.PrometheusTimestamp() {
+			return errors.Errorf("error min-time '%s' can't be greater than max-time '%s'", minTime, maxTime)
```
bwplotka:
```suggestion
			return errors.Errorf("invalid argument: min-time '%s' can't be greater than max-time '%s'", minTime, maxTime)
```

cmd/thanos/store.go
```diff
@@ -101,6 +115,8 @@ func runStore(
 	verbose bool,
 	syncInterval time.Duration,
 	blockSyncConcurrency int,
+	minTime *store.TimeOrDurationValue,
```
bwplotka: Hmmm I let's maybe put this `TimeOrDurationValue` to separate package? Can be useul for other components like compactor, querier etc

pkg/store/bucket.go
```diff
@@ -382,6 +400,27 @@ func (s *BucketStore) numBlocks() int {
...
+func (s *BucketStore) isBlockInMinMaxRange(ctx context.Context, id ulid.ULID) (bool, error) {
+	dir := filepath.Join(s.dir, id.String())
+
+	b := &bucketBlock{
+		logger: s.logger,
+		bucket: s.bucket,
+		id:     id,
+		dir:    dir,
+	}
+	if err := b.loadMeta(ctx, id); err != nil {
```
bwplotka: Can we load meta once? I think we do it everythere in small functions ):

bwplotka: maybe we should cache those on disk, not sure

pkg/store/bucket.go
```diff
@@ -382,6 +400,27 @@ func (s *BucketStore) numBlocks() int {
...
+		return false, err
+	}
+
+	// We check for blocks in configured minTime, maxTime range
```
bwplotka: missing trailing period

cmd/thanos/store.go
```diff
@@ -50,11 +50,23 @@ func registerStore(m map[string]setupFunc, app *kingpin.Application, name string
 	blockSyncConcurrency := cmd.Flag("block-sync-concurrency", "Number of goroutines to use when syncing blocks from object storage.").
 		Default("20").Int()
 
+	minTime := store.TimeOrDuration(cmd.Flag("min-time", "Start of time range limit to serve. Option can be a constant time in RFC3339 format or time duration relative to current time, such as -1.5d or 2h45m. Valid duration units are ms, s, m, h, d, w, y.").
+		Default("0000-01-01T00:00:00Z"))
```
bwplotka: hm.. I think the earliest date might be `-9999-12-31T23:59:59Z` but not 100% sure ;p to be strict

pkg/store/flag.go
```diff
@@ -0,0 +1,74 @@
...
+// Only one will be set.
+type TimeOrDurationValue struct {
+	t   *time.Time
+	dur *model.Duration
+}
+
+// Set converts string to TimeOrDurationValue
+func (tdv *TimeOrDurationValue) Set(s string) error {
+	t, err := time.Parse(time.RFC3339, s)
+	if err != nil {
```
bwplotka: please keep bigger code blocks with smaller indent. This means filtering err == nil first and then handling the err case

pkg/store/flag_test.go
```diff
@@ -0,0 +1,36 @@
...
+	"time"
+
+	"github.com/improbable-eng/thanos/pkg/store"
+	"github.com/improbable-eng/thanos/pkg/testutil"
+	"github.com/prometheus/prometheus/pkg/timestamp"
+	"gopkg.in/alecthomas/kingpin.v2"
+)
+
+func TestTimeOrDurationValue(t *testing.T) {
+	cmd := kingpin.New("test", "test")
```
bwplotka: You don't need kingpin for this test IMO

cmd/thanos/store.go
bwplotka: Plust @brancz mentioned use cases applies

pkg/store/bucket.go
```diff
@@ -382,6 +400,27 @@ func (s *BucketStore) numBlocks() int {
...
+	if b.meta.MinTime < s.minTime.PrometheusTimestamp() || b.meta.MaxTime > s.maxTime.PrometheusTimestamp() {
```
claytono: I think you want to compare against a blocks mintime, or maxtime, but not both.  The scenario I have in mind is something like this:

- You have a block that covers 01:00:00 to 03:00:00.
- You have a `thanos-store` process covering 00:00:00 to 02:00:00
- You have a second `thanos-store` process covering 02:00:01 to 04:00:00

I think with the logic above, *neither* thanos-store process will load the block because it's mintime is inside the first `thanos-store` process' range, but the maxtime is outside of its range.  For the second `thanos-store` you have the opposite situation.  Since neither of them own the *entire* block, neither of them will serve it.

bwplotka: Yea I think essentially we need this:
```suggestion
	if b.meta.MaxTime < s.minTime.PrometheusTimestamp() || b.meta.MinTime > s.maxTime.PrometheusTimestamp() {
```

 So:
```
Store range:                         |min --------------max|
Blocks:                  | ------ A--------| | -------- B --------|
```

Thanks for spotting this @claytono 

Also can we have tests that would catch this?

claytono: @bwplotka I think in that scenario, we will end up with the same block served by two stores.  For example:

```
Store A range: |min --- max|
Store B range              |min --- max|
Store C range                          |min --- max|
Blocks:        |----- A -----|----- B -----|----- C -----|
```

In this scenario, I think we want blocks A, B and C to be served by a single store.  With my previous PR, I picked minTime as the determining factor, so with that PR the assignments would be like this:

- Block A -> Store A 
- Block B -> Store A
- Block C -> Store B

I'm not sure there is a "right" answer for whether or not we should use minTime or maxTime for assigning blocks to stores, but I think in the interest of efficiency, if store time range are contiguous, then each block should only be owned by a single store.

pkg/store/flag.go
povilasv: nice catch :heart:

pkg/store/flag_test.go
povilasv: @bwplotka but I really like it :P 

I think it's and e2e test for flag parsing and as the package imports kingpin and has a special function for it, IMO it makes sense.

cmd/thanos/store.go
povilasv: Moved to model package

pkg/store/bucket.go
povilasv: @bwplotka @claytono 

IMO it's fine in current version. i.e. users just need to be careful then setting right parameters.

So you set `|minTime - maxTime|` and store only will respond with blocks that fit that range. Making thanos store returning blocks outside the range is I think broken behavior, as it will just confuse people, why `| mnTime - maxTime|` isn't working.


I'll add docs for that

povilasv: BTW there are tests for current behaviour. If we decide to change them we can refactor those.

pkg/store/bucket.go
povilasv: `b.loadMeta` already does this:

```
	if _, err := os.Stat(b.dir); os.IsNotExist(err) {
		if err := os.MkdirAll(b.dir, 0777); err != nil {
		}
		src := path.Join(id.String(), block.MetaFilename)

		level.Debug(b.logger).Log("msg", "download file", "bucket", b.bucket, "src", src, "dir", b.dir)
		if err := objstore.DownloadFile(ctx, b.logger, b.bucket, src, b.dir); err != nil {
			return errors.Wrap(err, "download meta.json")
		}
	} else if err != nil {
		return err
	}
	meta, err := metadata.Read(b.dir)
	if err != nil {
		return errors.Wrap(err, "read meta.json")
	}

```

cmd/thanos/store.go
povilasv: it doesn't time.Parse:

```
unexpected error: parsing time "-9999-12-31T23:59:59Z" as "2006-01-02T15:04:05Z07:00": cannot parse "-9999-12-31T23:59:59Z" as "2006"
```

```
func TestTime(t *testing.T) {
	ti, err := time.Parse(time.RFC3339, "-9999-12-31T23:59:59Z")
	testutil.Ok(t, err)

	fmt.Printf("time.Unix: %v, prom.timestamp: %v\n", ti.Unix(), timestamp.FromTime(ti))
}

```

docs/components/store.md
```diff
@@ -140,5 +140,28 @@ Flags:
...
+                                 m, h, d, w, y.
 
 ```
+
+## Time & Duration based partioning
+
+By default Thanos Store Gateway looks at all the data in Object Store and returns it based on query's time range.
+You can shard Thanos Store gateway based on time or duration relative to current time.
+
+For example setting: `--min-time=-6w` & `--max-time=-2w` will make Thanos Store Gateway look at blocks that fall within `now - 6 weeks` up to `now - 2 weeks`.
```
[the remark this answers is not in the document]
povilasv: now being  time of the `sync`, we run syncs every `3m`. I'll update docs.

povilasv:
@bwplotka @claytono @brancz 

Maybe we should actually do filtering based on just block's start time, for `duration` based approach it makes sense:
Lets's say 

Thanos Store1 `--min-time=0` to `-max-time=2w`
Thanos Store2 `--min-time=2w` to `-max-time=1m`

blocks would just constantly flow from Store1 to Store2, without gaps

pkg/store/bucket.go
claytono: I think it depends on what your use case is.   If the implementation requires that blocks must be entirely within the time range for the store, it may be impossible to pick time ranges where all blocks will be handled by a store, without having overlapping time ranges, which has other efficiency issues. 

Our primary desire for time based ranges is to be able to partition by time ranges, so the behavior as implemented now doesn't work well for us.  Ss the compactor generates larger and larger blocks, overlap between blocks with different labels makes it more and more likely that you're going to be serving very large blocks from multiple stores.  This means you'll be loading large blocks into memory in multiple places and wasting memory.  For us, memory usage of the store processes is the primary cost for Thanos.

Perhaps this should be configurable if different people want different behavior.

claytono:
@povilasv Sorry, I didn't see this newer reply before replying to the older one.  Filtering on just block start time would work well for us.

docs/components/store.md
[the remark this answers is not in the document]
povilasv: I've updated the docs and changed the implementation

povilasv:
I think looking at block's start time makes most sense for time based partitioning.

I've changed the implementation and updated docs & tests.

@brancz and @bwplotka PTAL

cmd/thanos/store.go
bwplotka: ;p

go.mod
```diff
@@ -50,3 +50,5 @@ require (
 	gopkg.in/fsnotify/fsnotify.v1 v1.4.7 // indirect
 	gopkg.in/yaml.v2 v2.2.2
 )
+
+go 1.13
```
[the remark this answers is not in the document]
povilasv: oops, I'm running go dev version so it adds this from time to time :D

cmd/thanos/store.go
```diff
@@ -51,11 +52,23 @@ func registerStore(m map[string]setupFunc, app *kingpin.Application, name string
 	blockSyncConcurrency := cmd.Flag("block-sync-concurrency", "Number of goroutines to use when syncing blocks from object storage.").
 		Default("20").Int()
 
+	minTime := model.TimeOrDuration(cmd.Flag("min-time", "Start of time range limit to serve. Thanos Store serves only blocks, which have start time greater than this value. Option can be a constant time in RFC3339 format or time duration relative to current time, such as -1.5d or 2h45m. Valid duration units are ms, s, m, h, d, w, y.").
```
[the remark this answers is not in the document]
povilasv: Re 1) `block-min-time` I can name it this way and add `block-max-time`, not a problem.

> (all blocks that overlaps are synced and only exact time range is announced and returned).
> If both max and min are inclusive or both exclusive you use case will not work in all cases. How you will compose chain of store gateways then?

Overlaps are not a problem for us, Thanos Querier will merge responses, so you just make overlaping chain of `--block-min-time` & `--block-max-time`.

> Now if you want to spread blocks evenly among store gateway this gets quite tricky.

This is PR is not about spreading the blocks evenly. I don't really care about that and I don't really see the usecase of spreading blocks evenly.

> How you will partition time when compactor is compacting blocks a lot? When those block time ranges are super dynamic?

Not an issue the sync job already adds new blocks which fit the time range and removes the old ones. So some times it will go into one Thanos Store, sometimes into another.

>I find it extremely not intuitive to work on startTime only. 

We can add 4 flags then: `--block-start-time-min-time`, `--block-start-time-max-time`
, `--block-end-time-min-time`, `--block-end-time-max-time`

> I think @claytono can achieve block paritioning by pure time based partitioning as well -> just make sure time range is not overlapping with not wanted blocks AFAIK

I don't follow, why do we actually care that time range is not overlapping?

The idea is that we need to scale Thanos Store, it's impossible to keep low response times if we don't partition Thanos Store gateway. I.e. what happens to our LRU cache if someone queries for short time period <1w and then a few people on long >3mo, etc. When you shard you can optimize for that type of a deal. Also in general it is more approachable instead of having one big store gateway we have many little ones, which have advantage of easier scheduling (don't need a seperate node for store gw anymore), reducing query latency (every store performs smaller amount of work), etc. 

IMO block juggling is fine. Maybe I need to improve docs more about overlapping case, so that people don't try to do perfect ranges.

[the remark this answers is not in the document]
claytono: I agree with @povilasv.  For us this is about horizontal scaling.  We're already partitioning by time by having multiple buckets and creating new ones as needed.  This PR will allow us to reduce a lot of manual work involved in that and automate the rest.

While there may be better solutions to https://github.com/improbable-eng/thanos/issues/448, I think we're going to always have a need for horizontal scaling and we don't have a solution to #448 today.  I definitely want to see the memory efficiency improved, but horizontally scaling has allowed us to ignore the problem for the time being.  This PR gives us breathing room while that's worked out.

povilasv:
I've renamed the flags and test it in dev environment. There is ` quay.io/utilitywarehouse/thanos:time-partioning-2019-05-24-00cf917a` with docker image.

This is how it looks:
Old:
sidecar


New with additional Thanos Store:
sidecar
short-store
long-store

New Config:
2x stores with:
```
      containers:
      - args:
        - store
        - --log.level=info
        - --data-dir=/var/thanos/store
        - --objstore.config-file=/etc/thanos/config.yaml
        - --index-cache-size=12GB
        - --chunk-pool-size=6GB
        - --max-block-start-time=-40h # This will skip blocks, which start time < now-40h, as they are served by sidecar
        - --min-block-start-time=-5w # This will serve blocks up to now-5w
```

2 Store's with:
```
      containers:
      - name: thanos-store
        image: quay.io/utilitywarehouse/thanos:time-partioning-2019-05-24-00cf917a
        args:
        - store
        - --log.level=info
        - --data-dir=/var/thanos/store
        - --objstore.config-file=/etc/thanos/config.yaml
        - --index-cache-size=12GB
        - --chunk-pool-size=6GB
        - --max-block-start-time=-5w
```
The idea behind this config is that:

thanos sidecar responds data with < 40h
thanos sidecar + thanos store responds to data with >40h && <= 48h
thanos store repsonds with >48h < -5w
another thanos store responds with >5w

As stated before overlaping blocks are fine, Thanos Querier merges them. 
In current default config we are constantly merging responses from Sidecar and Thanos Store Gateway.

povilasv:
I did some stupid simple "benchmarking", opened Thanos Store grafana dashboard loaded and did 3 cases :
last 2d, interval 30min
Last 7d interval 30min
Last 30d interval 30min

Current default config:

2d

old 2d 30min interval

7d

old 7d 30min interval

30d

old 30d 30min interval

New Config 2 Thanos Store's with short term data, 2 Thanos Store's with long term data, only Sidecar's serve < 40h.

2d
new2d 30min

7d
new 7d 30min

30d

new 30d 30min

povilasv:
Lastly, IMO time-sharding is a must. Imagine if you have 10 years of data or 100 years of data, we will need mainframes to handle 100year type of queries. Sharding does solve this.

cmd/thanos/store.go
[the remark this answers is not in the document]
bwplotka: > and many more clients againts Object store, which as experience shows is faster than having a single client against bucket.

Where you get that from? (: 

Ok @claytono @povilasv so maybe from different position. Let's say we have just `min-time` `max-time` ranges? Nothing block based, but purely time wise. Why is that insufficient? 

My main concern is that tieing some user expierience API (store gw flag) to some hidden format that is **eventually** synced will cause only pain. I would like to avoid this? Having simple time range flags would be both simple, understandable and solve you use case, isn't it?

claytono:
@bwplotka I'd like to see this merged because it solves an actual problem we have.  I'm fine with other approaches if there is a PR you'd prefer over this one.  Unfortunately the operational burden of manually partitioning into multiple buckets is becoming too much.   

If there isn't a mainline approach for horizontal scaling, we're going to have to carry a patch internally for this sort of functionality.  That's a burden we'd really like to avoid if possible.

bwplotka:
Waiting for @povilasv on this. If you feel blocked you can indeed apply patch to your fork, but we would love to unblock you. 

However I would love to make it right and non confusing from beginning. IMO `min-time` and `max-time` will help you.


Also we are considering adding something like this to store gateway as well: https://github.com/improbable-eng/thanos/pull/1245 so you can shard functionally.

cmd/thanos/store.go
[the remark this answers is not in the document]
povilasv: >Where you get that from? (:

For example https://care.qumulo.com/hc/en-us/articles/360015703713-Performance-Results-of-MinIO-with-Qumulo

Take a look at any object store benchmarks, I haven't seen a benchmark which tells that if you use 1 client it's going to be have faster reads then multi client. But you can prove me wrong.

Minio case singe client was:  643 MB/s
Multi-Stream Read: 1.1-1.7GB/s

S3:
Single: 390.8MB/sec
Multi-Stream Read: 1-1.5GB/s

>  Nothing block based, but purely time wise. Why is that insufficient?

Again time duration based is just a simple way to shard thanos store for large deployments.

> My main concern is that tieing some user expierience API (store gw flag) to some hidden format that is eventually synced will cause only pain.

Yes that's the only drawback of this solution, is that people will need to be educated on doing overlapping time ranges, or not doing it at all.

S3 has eventual consistency anyway so in any solution we would have to take similar measures.

povilasv:
@bwplotka 

> paritionining that is agnostic to storage format would be more both more flexible and 5 times less confusing then this. Happy to discuss though.

I don't follow. So this just introduces time sharding based on time ranges. In any case the data will be time lined so the flags are agnostic as far as I know. Could you elaborate on thjis?

> 
> > Lastly, IMO time-sharding is a must. Imagine if you have 10 years of data or 100 years of data, we will need mainframes to handle 100year type of queries. Sharding does solve this.
> 
> I am not opposed to time sharding. I am only saying that block based partitioning in unneccesary tying user to block formats that can be fully dynamic and eventually consistent or might even disappear totally in next iterations of Thanos and TSDB.

Again in this case, we can agree on flags, still land this implementation and warn users about overlapping blocks. If / when we have something better, we can just warn users about implementation change and keep the same flags.

If you don't like the flag names, suggestions are welcome :)

bwplotka:
> If / when we have something better, we can just warn users about implementation change and keep the same flags

Warn users? And who will spent time to explain how to use it? And how to explain if we don't have any good answer to dynamic blocks.

It's not about flag names, it's about the user expierience. IMO we need `min-time` and `max-time` that would in naive way map to underlying blocks based on what min and max time they have in meta.json. Easy to use, low cognitive load and expected behaviour. Plus it should solve requirements mentioned here. 

Any argument why we should introduce this massive complexity?

brancz:
I agree with @bwplotka here. Two flags are enough, and whether that resolves to the full block being handled or not is up to the implementation detail of the store. Something we need to be careful here though, especially with relative times is clock synchronization, meaning we probably intentionally want overlap in time ranges to account for slight clock synchronization issues. Basically we want as much overlap as we want to tolerate clock skew.

bwplotka:
I think we had agreement offline on this @povilasv.

Do you have time to finish this? (: It would be nice to have.

pkg/model/timeduration.go
```diff
@@ -0,0 +1,75 @@
...
+	}
+	tdv.Dur = &dur
+	return nil
+}
+
+// String returns either tume or duration.
+func (tdv *TimeOrDurationValue) String() string {
+	switch {
+	case tdv.Time != nil:
+		return tdv.Time.String()
```
devnev: Would it make sense for this to format back to RFC3339?

cmd/thanos/store.go
```diff
@@ -51,11 +52,37 @@ func registerStore(m map[string]setupFunc, app *kingpin.Application, name string
 	blockSyncConcurrency := cmd.Flag("block-sync-concurrency", "Number of goroutines to use when syncing blocks from object storage.").
 		Default("20").Int()
 
+	minBlockStartTime := model.TimeOrDuration(cmd.Flag("min-block-start-time", "Start of time range limit to serve. Thanos Store serves only blocks, which have start time greater than this value. Option can be a constant time in RFC3339 format or time duration relative to current time, such as -1.5d or 2h45m. Valid duration units are ms, s, m, h, d, w, y.").
+		Default("0000-01-01T00:00:00Z"))
+
+	maxBlockStartTime := model.TimeOrDuration(cmd.Flag("max-block-start-time", "End of time range limit to serve. Thanos Store serves only blocks, which have start time is less than this value. Option can be a constant time in RFC3339 format or time duration relative to current time, such as -1.5d or 2h45m. Valid duration units are ms, s, m, h, d, w, y.").
```
devnev: The documentation for the duration variant read like its parsing using `time.ParseDuration` but the implementation uses `model.ParseDuration` which IIUC only allows a single value and unit and no dot. Am I missing something?

antonio:
> I think we had agreement offline on this @povilasv.

Is there a gist that you can share?

> Do you have time to finish this? (: It would be nice to have.

:+1: Please let us know if we can help in any way.

povilasv:
I'm continuing work on a fresh PR https://github.com/thanos-io/thanos/pull/1408

PR #1078, opened 2019-04-24
website: Moved to netlify.

bwplotka:
cc @adrien-f 

Also updated hugo version.


bwplotka:
PTAL

PR #1082, opened 2019-04-25
store/proxy: properly check if context has ended

GiedriusS:
How the code was before it could happen that we might receive some
series from the stream however by the time we'd send them back to the
reader, it would not read it any more since the deadline would have been
exceeded.

Properly use a `select` here to get out of the goroutine if the deadline
has been exceeded.

### DIFF
--- a/pkg/store/proxy.go
+++ b/pkg/store/proxy.go
@@ -294,10 +294,6 @@ func startStreamSeriesSet(
 				return
 			}
 
-			if ctx.Err() != nil {
-				return
-			}
-
 			if err != nil {
 				wrapErr := errors.Wrapf(err, "receive series from %s", s.name)
 				if partialResponse {
@@ -315,7 +311,14 @@ func startStreamSeriesSet(
 				s.warnCh.send(storepb.NewWarnSeriesResponse(errors.New(w)))
 				continue
 			}
-			s.recvCh <- r.GetSeries()
+
+			select {
+			case s.recvCh <- r.GetSeries():
+				continue
+			case <-ctx.Done():
+				return
+			}
+
 		}
 	}()
 	return s


pkg/store/proxy.go
```diff
@@ -315,7 +311,14 @@ func startStreamSeriesSet(
 				s.warnCh.send(storepb.NewWarnSeriesResponse(errors.New(w)))
 				continue
 			}
-			s.recvCh <- r.GetSeries()
+
+			select {
+			case s.recvCh <- r.GetSeries():
+				continue
+			case <-ctx.Done():
+				return
```
povilasv: I think this should also take care of partial responses.

In the case of context cancel we want:
```
	case <-ctx.Done():

	wrapErr := errors.Wrapf(ctx.Err(), "receive series from %s", s.name)
	if partialResponse {
			s.warnCh.send(storepb.NewWarnSeriesResponse(wrapErr))					
			return				
	}			
	
	s.errMtx.Lock()					s.errMtx.Lock()
	s.err = wrapErr					s.err = wrapErr
	s.errMtx.Unlock()					s.errMtx.Unlock()
	return
```

povilasv:
@GiedriusS so your fix will timeout the case once the buffered channel is full, which I don't think will actually help you, but it's still a reasonable PR.

povilasv:
And please add the test to that test suite. basically fill the buffered channel, don't have anything consume from it and then look how timeouts behave.

povilasv:
FYI nothing in goroutines.txt indicate that we we leak goroutines.

GiedriusS:
Will do, removed that stuff from the PR. Most likely it is unrelated, sorry for the confusion.

GiedriusS:
Small update:
Without this, the pattern of the metric `prometheus_engine_queries` on Thanos Query was this:
Screenshot from 2019-04-26 11-55-46
(restarted the node at the end)

With this the pattern is back to normal:
Screenshot from 2019-04-26 11-50-21

Yellow line shows `prometheus_engine_queries` with this change, green is without. Both nodes are behind a RR LB. Either way, let me make a test case for this :smile:

pkg/store/proxy.go
bwplotka: Hmm, we did not do it previously though... Let me look on the flow. We definitely want to show partial response OR fail whole thing in this case.

bwplotka: yes! @GiedriusS Double checked and we need to:

Return error if partial response is disabled or send err to warn Ch (with who canelled as well somehow). (: I know your PR did not add the regression, but let's fix it?

GiedriusS:
... but with that added extra message we would get warning messages like this:
```
proxy_test.go:550:got [warning failed to receive any data in 4s from test: context deadline exceeded warning receive series from test: context canceled]                                                             
                                                                                                                                                                                                                     
        exp: 2                                                                                                                                                                                                       
                                                                                                                                                                                                                     
        got: 4
```
perhaps it would be better to leave it as-is?

GiedriusS:
Decided to leave it as-is because `Next()` already catches the errors on this context (it is the same one). However, the test suite still fails:
```
proxy_test.go:550:got [warning failed to receive any data in 4s from test: context deadline exceeded warning]                                                                                                       

        exp: 2

        got: 3
```
I'm not sure if I'll have enough time to fix this in time if you want to do the RC today :P

pkg/store/proxy.go
bwplotka: discussed offline all good.

PR #323, opened 2018-05-06
Use custom roundtripper in the minio client

alvaroaleman:
This allows us to set a `ResponseHeaderTimeout` which avoids
cases where the client blocks forever when the S3 endpoint
takes connections but doesn't return anything.

Fixes #315

Issue #318, opened 2018-05-02
Retry on network failures (e.g uploads)


bwplotka:
Not critical since compactor just restarted and continued just fine, but can be annoying.

```
level=error name=thanos-compactor ts=2018-04-28T10:32:12.73383864Z caller=main.go:147 msg="running command failed" err="first pass of downsampling failed: retrieve bucket block metas: get meta for block 01C6XZ1256S7VFNQP9D36XJ4F4: Get https://storage.googleapis.com/thanos-alpha/01C6XZ1256S7VFNQP9D36XJ4F4/meta.json: dial tcp [xxx]:443: connect: network is unreachable"
```

Bplotka:
Especially funny is the single error during sync metas that causes compactor to retry WHOLE sync.

```
level=debug ts=2018-05-31T00:14:19.617242203Z caller=compact.go:165 msg="download meta" block=01C8146Z5B7AX5HYTV6S15G044
level=error ts=2018-05-31T00:14:34.617896977Z caller=compact.go:205 msg="retriable error" err="sync: retrieve bucket block metas: downloading meta.json for 01C8146Z5B7AX5HYTV6S15G044: decode meta.json for block 01C8146Z5B7AX5HYTV6S15G044: Get https://prod-int-spaces01.nyc3.internal.digitalocean.com/pandora-lts-ams2-hvs/01C8146Z5B7AX5HYTV6S15G044/meta.json: net/http: timeout awaiting response headers"
level=info ts=2018-05-31T00:15:57.63331754Z caller=compact.go:126 msg="start sync of metas"
```

Can see for @TimSimmons logs that it happens quite often. We should retry just the problematic thing.

asbjxrn:
This happens both for downloads and uploads (of compacted blocks) for me. And I also see the same timeouts when uploading from the sidecars, so I think this issue applies to all components that communicates with the block store.

With a large enough number of thanos sidecars this issue can be quite bad as once you fall behind the number of files up/downloaded gets large which means higher chance of hitting the issue which may put you even further behind and so on.

Bplotka:
yup, exactly.

bwplotka:
Ok this is interesting as s3 client really have retries: https://sourcegraph.com/github.com/minio/minio-go@master/-/blob/api.go#L524:17

Maybe it's worth to reach them?

bwplotka:
We double check and retries are already implemented in minio and GCS client. For each client we need to double check and add if missing (per client).

[...]

PR #1094, opened 2019-04-29
objstore: Increase the response header timeout in the S3 provider to 2 minutes and make it configurable

antonio:
## Changes

This PR increases the response header timeout in the S3 provider from 15 seconds (introduced in #323) to 2 minutes, an arbitrary number that I've found works well with large blocks as a result of debugging #318. This has solved my recurring problems with the `net/http: timeout awaiting response headers` error.

I'm not sure whether a fixed number like this or having a configuration flag is better: while the latter is more flexible I believe it introduces complexity and does not provide much value. I'm happy to update this if that's the preferred solution, though.

## Verification

I have been running three different compactor processes against a few buckets that would always fail to be compacted with the aforementioned error, and haven't seen a single error in 4 days.


midnightconman:
This seems very similar to https://github.com/improbable-eng/thanos/pull/567, can you explain the difference between the two options?

antonio:
@povilasv @bwplotka thank you for your feedback, I'll make the rest of the HTTP options configurable in the same way that it's done in the PR that @midnightconman linked to.

antonio:
@povilasv @bwplotka I think this is ready for another :eyes:

I've decided to only make that one option configurable in the end as opposed to make every `HTTPTransport` timeout configurable to not add extra complexity, but could do the latter if you feel different.

I've refactored the tests to make it a little bit more clear what the default values are for the HTTP client, and reworded some parts of the documentation while I dropped a line about this new option.

GiedriusS:
:tada: thank you :heart:

PR #1097, opened 2019-04-30
feat: added thanos check rules command

### LINKED ISSUE #1096: Rule: Add linting command for prometheus rules
Thanos rule has additional config `partial_response_strategy` for rule groups in the rules file.

Prometheus unfortunately uses strict yaml unmarshaling so the `promtool check rules ...` fails with:
```
FAILED:
yaml: unmarshal errors:
  line 3: field partial_response_strategy not found in type rulefmt.RuleGroup
```
Possible solution could be adding own rule lint command to the thanos like `thanos check rules` which would check the syntax valid for thanos-rule.

FUSAKLA:
Resolves #1096 
This adds `thanos check rules` command to check Prometheus rules.

We cannot use the `promtool check rules` since Thanos-rule allows the additional field `partial_response_strategy`

## Changes

- [x] added the check rules cmd
- [x] changelog entry TBD
- [x] docs update?

## Verification

Tested it on different types of invalid rule files even valid. Works fine.

### DIFF
--- a/cmd/thanos/check.go
+++ b/cmd/thanos/check.go
@@ -55,15 +55,14 @@ func checkRulesFiles(logger log.Logger, files *[]string) error {
 	return nil
 }
 
-type ThanosRuleGroup struct{
-    PartialResponseStrategy string `yaml:"partial_response_strategy"`
-    rulefmt.RuleGroup `yaml:",inline"`
+type ThanosRuleGroup struct {
+	PartialResponseStrategy string `yaml:"partial_response_strategy"`
+	rulefmt.RuleGroup       `yaml:",inline"`
 }
 
 type ThanosRuleGroups struct {
 	Groups []ThanosRuleGroup `yaml:"groups"`
 }
-
 
 func checkRules(logger log.Logger, filename string) (int, tsdb.MultiError) {
 	level.Info(logger).Log("msg", "checking", "filename", filename)
--- a/cmd/thanos/check.go
+++ b/cmd/thanos/check.go
@@ -9,7 +9,7 @@ import (
 	"github.com/opentracing/opentracing-go"
 	"github.com/prometheus/client_golang/prometheus"
 	"github.com/prometheus/prometheus/pkg/rulefmt"
-	"github.com/prometheus/tsdb"
+	"github.com/prometheus/tsdb/errors"
 	"gopkg.in/alecthomas/kingpin.v2"
 	"gopkg.in/yaml.v2"
 )
@@ -34,7 +34,7 @@ func registerCheckRules(m map[string]setupFunc, root *kingpin.CmdClause, name st
 }
 
 func checkRulesFiles(logger log.Logger, files *[]string) error {
-	failed := tsdb.MultiError{}
+	failed := errors.MultiError{}
 
 	for _, f := range *files {
 		n, errs := checkRules(logger, f)
@@ -64,9 +64,9 @@ type ThanosRuleGroups struct {
 	Groups []ThanosRuleGroup `yaml:"groups"`
 }
 
-func checkRules(logger log.Logger, filename string) (int, tsdb.MultiError) {
+func checkRules(logger log.Logger, filename string) (int, errors.MultiError) {
 	level.Info(logger).Log("msg", "checking", "filename", filename)
-	checkErrors := tsdb.MultiError{}
+	checkErrors := errors.MultiError{}
 
 	b, err := ioutil.ReadFile(filename)
 	if err != nil {
--- a/cmd/thanos/check.go
+++ b/cmd/thanos/check.go
@@ -5,7 +5,6 @@ import (
 
 	"github.com/go-kit/kit/log"
 	"github.com/go-kit/kit/log/level"
-	thanosrule "github.com/improbable-eng/thanos/pkg/rule"
 	"github.com/oklog/run"
 	"github.com/opentracing/opentracing-go"
 	"github.com/prometheus/client_golang/prometheus"
@@ -56,6 +55,16 @@ func checkRulesFiles(logger log.Logger, files *[]string) error {
 	return nil
 }
 
+type ThanosRuleGroup struct{
+    PartialResponseStrategy string `yaml:"partial_response_strategy"`
+    rulefmt.RuleGroup `yaml:",inline"`
+}
+
+type ThanosRuleGroups struct {
+	Groups []ThanosRuleGroup `yaml:"groups"`
+}
+
+
 func checkRules(logger log.Logger, filename string) (int, tsdb.MultiError) {
 	level.Info(logger).Log("msg", "checking", "filename", filename)
 	checkErrors := tsdb.MultiError{}
@@ -66,8 +75,8 @@ func checkRules(logger log.Logger, filename string) (int, tsdb.MultiError) {
 		return 0, checkErrors
 	}
 
-	var rgs thanosrule.RuleGroups
-	if err := yaml.Unmarshal(b, &rgs); err != nil {
+	var rgs ThanosRuleGroups
+	if err := yaml.UnmarshalStrict(b, &rgs); err != nil {
 		checkErrors.Add(err)
 		return 0, checkErrors
 	}
@@ -89,7 +98,7 @@ func checkRules(logger log.Logger, filename string) (int, tsdb.MultiError) {
 	return numRules, checkErrors
 }
 
-func thanosRuleGroupsToPromRuleGroups(ruleGroups thanosrule.RuleGroups) rulefmt.RuleGroups {
+func thanosRuleGroupsToPromRuleGroups(ruleGroups ThanosRuleGroups) rulefmt.RuleGroups {
 	promRuleGroups := rulefmt.RuleGroups{Groups: []rulefmt.RuleGroup{}}
 	for _, g := range ruleGroups.Groups {
 		group := rulefmt.RuleGroup{
--- a/cmd/thanos/check.go
+++ b/cmd/thanos/check.go
@@ -1,0 +1,119 @@
+package main
+
+import (
+	"io/ioutil"
+
+	"github.com/go-kit/kit/log"
+	"github.com/go-kit/kit/log/level"
+	thanosrule "github.com/improbable-eng/thanos/pkg/rule"
+	"github.com/oklog/run"
+	"github.com/opentracing/opentracing-go"
+	"github.com/prometheus/client_golang/prometheus"
+	"github.com/prometheus/prometheus/pkg/rulefmt"
+	"github.com/prometheus/tsdb"
+	"gopkg.in/alecthomas/kingpin.v2"
+	"gopkg.in/yaml.v2"
+)
+
+func registerChecks(m map[string]setupFunc, app *kingpin.Application, name string) {
+	cmd := app.Command(name, "Linting tools for Thanos")
+	registerCheckRules(m, cmd, name)
+}
+
+func registerCheckRules(m map[string]setupFunc, root *kingpin.CmdClause, name string) {
+	checkRulesCmd := root.Command("rules", "Check if the rule files are valid or not.")
+	ruleFiles := checkRulesCmd.Arg(
+		"rule-files",
+		"The rule files to check.",
+	).Required().ExistingFiles()
+
+	m[name+" rules"] = func(g *run.Group, logger log.Logger, reg *prometheus.Registry, _ opentracing.Tracer, _ bool) error {
+		// Dummy actor to immediately kill the group after the run function returns.
+		g.Add(func() error { return nil }, func(error) {})
+		return checkRulesFiles(logger, ruleFiles)
+	}
+}
+
+func checkRulesFiles(logger log.Logger, files *[]string) error {
+	failed := tsdb.MultiError{}
+
+	for _, f := range *files {
+		n, errs := checkRules(logger, f)
+		if errs.Err() != nil {
+			level.Error(logger).Log("result", "FAILED")
+			for _, e := range errs {
+				level.Error(logger).Log("error", e.Error())
+				failed.Add(e)
+			}
+			level.Info(logger).Log()
+			continue
+		}
+		level.Info(logger).Log("result", "SUCCESS", "rules found", n)
+	}
+	if failed.Err() != nil {
+		return failed
+	}
+	return nil
+}
+
+func checkRules(logger log.Logger, filename string) (int, tsdb.MultiError) {
+	level.Info(logger).Log("msg", "checking", "filename", filename)
+	checkErrors := tsdb.MultiError{}
+
+	b, err := ioutil.ReadFile(filename)
+	if err != nil {
+		checkErrors.Add(err)
+		return 0, checkErrors
+	}
+
+	var rgs thanosrule.RuleGroups
+	if err := yaml.Unmarshal(b, &rgs); err != nil {
+		checkErrors.Add(err)
+		return 0, checkErrors
+	}
+
+	// We need to convert Thanos rules to Prometheus rules so we can use their validation.
+	promRgs := thanosRuleGroupsToPromRuleGroups(rgs)
+	if errs := promRgs.Validate(); errs != nil {
+		for _, e := range errs {
+			checkErrors.Add(e)
+		}
+		return 0, checkErrors
+	}
+
+	numRules := 0
+	for _, rg := range rgs.Groups {
+		numRules += len(rg.Rules)
+	}
+
+	return numRules, checkErrors
+}
+
+func thanosRuleGroupsToPromRuleGroups(ruleGroups thanosrule.RuleGroups) rulefmt.RuleGroups {
+	promRuleGroups := rulefmt.RuleGroups{Groups: []rulefmt.RuleGroup{}}
+	for _, g := range ruleGroups.Groups {
+		group := rulefmt.RuleGroup{
+			Name:     g.Name,
+			Interval: g.Interval,
+			Rules:    []rulefmt.Rule{},
+		}
+		for _, r := range g.Rules {
+			group.Rules = append(
+				group.Rules,
+				rulefmt.Rule{
+					Record:      r.Record,
+					Alert:       r.Alert,
+					Expr:        r.Expr,
+					For:         r.For,
+					Labels:      r.Labels,
+					Annotations: r.Annotations,
+				},
+			)
+		}
+		promRuleGroups.Groups = append(
+			promRuleGroups.Groups,
+			group,
+		)
+	}
+	return promRuleGroups
+}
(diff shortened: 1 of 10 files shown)


cmd/thanos/check.go
```diff
@@ -0,0 +1,113 @@
...
+	).Required().ExistingFiles()
+
+	m[name+" rules"] = func(g *run.Group, logger log.Logger, reg *prometheus.Registry, _ opentracing.Tracer, _ bool) error {
+		// Dummy actor to immediately kill the group after the run function returns.
+		g.Add(func() error { return nil }, func(error) {})
+		return checkRulesFiles(ruleFiles)
+	}
+}
+
+func checkRulesFiles(files *[]string) error {
```
povilasv: could you add tests for this.

cmd/thanos/check.go
```diff
@@ -0,0 +1,113 @@
...
+	var failed error = nil
```
povilasv:
```suggestion
	var failed error
```

cmd/thanos/check.go
```diff
@@ -0,0 +1,113 @@
...
+
+	for _, f := range *files {
+		if n, errs := checkRules(f); errs != nil {
+			fmt.Fprintln(os.Stderr, "  FAILED:")
+			for _, e := range errs {
+				fmt.Fprintln(os.Stderr, e.Error())
+			}
+			failed = errs[0]
+		} else {
```
povilasv: remove else statement, and add `continue` to `errs != nil` case.

as per https://github.com/golang/go/wiki/CodeReviewComments#indent-error-flow

cmd/thanos/check.go
```diff
@@ -0,0 +1,113 @@
...
+func checkRulesFiles(files *[]string) error {
+	var failed error = nil
+
+	for _, f := range *files {
+		if n, errs := checkRules(f); errs != nil {
+			fmt.Fprintln(os.Stderr, "  FAILED:")
+			for _, e := range errs {
+				fmt.Fprintln(os.Stderr, e.Error())
+			}
+			failed = errs[0]
```
povilasv: Maybe we should just join together the errors, instead of returining first error?

FUSAKLA: Yes, I didn't want to over-complicate it since those errors are all printed out nevertheless but in case of reuse of the function in the future it would be better. Will merge those thanks for pointing out.

FUSAKLA:
Just one general thought. Should we run this check in the thanos-rule also? 
Would it be more safe to validate the rules before startup to provide better output and and avoid crashing on invalid data? Possibly as a new PR.

cmd/thanos/check.go
FUSAKLA: hmm this was just copied from the `promtool` but there is still the `fmt.Println()` which is common for both flows. I'd need to duplicate it not sure if that is better.

cmd/thanos/check.go
```diff
@@ -0,0 +1,113 @@
...
+		fmt.Println()
+	}
+	if failed != nil {
+		return failed
+	}
+	return nil
+}
+
+func checkRules(filename string) (int, []error) {
+	fmt.Println("Checking", filename)
```
bwplotka: some logger would be nice?

cmd/thanos/check.go
```diff
@@ -0,0 +1,113 @@
...
+	if err != nil {
+		return 0, []error{err}
+	}
+
+	var rgs thanosrule.RuleGroups
+	if err := yaml.Unmarshal(b, &rgs); err != nil {
+		return 0, []error{err}
+	}
+
+	// We need to convert Thanos rules to Prometheus rules so we can use their validation
```
bwplotka: missing trailing comma (:

cmd/thanos/check.go
```diff
@@ -0,0 +1,113 @@
...
+
+	for _, f := range *files {
+		if n, errs := checkRules(f); errs != nil {
+			fmt.Fprintln(os.Stderr, "  FAILED:")
+			for _, e := range errs {
+				fmt.Fprintln(os.Stderr, e.Error())
+			}
+			failed = fmt.Errorf("Errors: %v", errs)
+		} else {
+			fmt.Printf("  SUCCESS: %d rules found\n", n)
```
bwplotka: can we use logger?

cmd/thanos/check.go
```diff
@@ -0,0 +1,113 @@
...
+		}
+		fmt.Println()
+	}
+	if failed != nil {
+		return failed
+	}
+	return nil
+}
+
+func checkRules(filename string) (int, []error) {
```
bwplotka: worth to just return `error` and use  `tsdb.Multierrror`?

cmd/thanos/check.go
FUSAKLA: refactored, should be ok now

FUSAKLA:
@povilasv Added tests and resolved all your comments PTAL if that's ok with you.

cmd/thanos/check.go
FUSAKLA: Great, didn't know about the tsdb multierror. Will change thanks.

FUSAKLA: done, switched to multierror (just a heads up in new version of TSDB multierror moves from generic tsdb package to tsdb/errors package. Imported if according to godoc and then gomod completely broke the dependencies since it upgraded tsdb to new version since the old did not had the package :D )

cmd/thanos/check.go
FUSAKLA: Switched to logger everywhere in the check cmd

cmd/thanos/check.go
FUSAKLA: thanks, added :)

FUSAKLA:
So I managed to update even the docs and the genflags script.
Please check those also.

Also the output is now using logger so it looks like this:
```bash
$ ./thanos check rules cmd/thanos/testdata/rules-files/*.yaml ; echo "$?"
level=info ts=2019-05-02T05:33:04.821854781Z caller=check.go:61 msg=Checking filename=cmd/thanos/testdata/rules-files/invalid-rules-data.yaml
level=error ts=2019-05-02T05:33:04.82280852Z caller=check.go:44 result=FAILED
level=error ts=2019-05-02T05:33:04.822879215Z caller=check.go:46 Groupnameshouldnotbeempty=(MISSING)
level=info ts=2019-05-02T05:33:04.822932943Z caller=check.go:48
level=info ts=2019-05-02T05:33:04.822970322Z caller=check.go:61 msg=Checking filename=cmd/thanos/testdata/rules-files/invalid-yaml-format.yaml
level=error ts=2019-05-02T05:33:04.823160151Z caller=check.go:44 result=FAILED
level=error ts=2019-05-02T05:33:04.823198963Z caller=check.go:46 yaml:line4:couldnotfindexpected':'=(MISSING)
level=info ts=2019-05-02T05:33:04.823230325Z caller=check.go:48
level=info ts=2019-05-02T05:33:04.823263265Z caller=check.go:61 msg=Checking filename=cmd/thanos/testdata/rules-files/valid.yaml
level=info ts=2019-05-02T05:33:04.823943932Z caller=check.go:52 result=SUCCESS rulesfound=2
check rules command failed: yaml: line 4: could not find expected ':'
1
```

FUSAKLA:
The netify preview fails with
```
7:28:17 AM: Error: Error building site: failed to render pages: render of "page" failed: execute of template failed: html/template:proposal/single.html:5:28: no such template "_default/sidemenu.html"
``` 

Is that me or some flaky behavior?

docs/components/check.md
```diff
@@ -0,0 +1,82 @@
...
+
+
+### Verify
+
+`check rules` checks the Prometheus rules, used by the Thanos rule node, if they are valid.
+The check should be equivalent for the `promtool check rules` but that cannot be used because
+Thanos rule has extended rules file syntax for example with `partial_response_strategy`
+filed which `promtool` does not allow.
+
+If the check fails the command ends with error code `1` otherwise `0`.
```
povilasv:
```suggestion
If the check fails the command fails with exit code `1`, otherwise `0`.
```

cmd/thanos/check.go
```diff
@@ -0,0 +1,121 @@
...
+				level.Error(logger).Log("error", e.Error())
+				failed.Add(e)
+			}
+			level.Info(logger).Log()
+			continue
+		}
+		level.Info(logger).Log("result", "SUCCESS", "rules found", n)
+	}
+	if failed.Err() != nil {
+		return errors.New(failed.Error())
```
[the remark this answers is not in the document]
FUSAKLA: oh it implements the error interface, great thanks

FUSAKLA:
Thanks @povilasv ! all applied.

cmd/thanos/check.go
povilasv:
```suggestion
		return failed
```

I think you forgot to do this one :P 
Or do you disagree?

povilasv:
@FUSAKLA you just forced pushed which removed the fixes :P could reapply again?

FUSAKLA:
Oh my.. sorry it's been a long day. Thanks for noticing!

facepalm

docs/components/check.md
FUSAKLA: sorry should be there again

cmd/thanos/check.go
FUSAKLA: sorry should be there again

cmd/thanos/check.go
```diff
@@ -0,0 +1,121 @@
...
+		level.Info(logger).Log("result", "SUCCESS", "rules found", n)
+	}
+	if failed.Err() != nil {
+		return failed
+	}
+	return nil
+}
+
+func checkRules(logger log.Logger, filename string) (int, tsdb.MultiError) {
+	level.Info(logger).Log("msg", "Checking", "filename", filename)
```
GiedriusS: I'm not sure but I think the convention is to start the log messages with a lower-case letter. Could anyone else chime in on this?

cmd/thanos/check.go
```diff
@@ -0,0 +1,121 @@
...
+	"gopkg.in/yaml.v2"
+)
+
+func registerChecks(m map[string]setupFunc, app *kingpin.Application, name string) {
+	cmd := app.Command(name, "Linting tools for Thanos")
+	registerRulesCheck(m, cmd, name)
+	return
+}
+
+func registerRulesCheck(m map[string]setupFunc, root *kingpin.CmdClause, name string) {
```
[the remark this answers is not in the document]
FUSAKLA: sounds good, renamed

cmd/thanos/check.go
FUSAKLA: You are right, I believe it was because of error wrapping maybe? 
Changed to lowercase.

cmd/thanos/check.go
```diff
@@ -0,0 +1,121 @@
...
+	"github.com/prometheus/prometheus/pkg/rulefmt"
+	"github.com/prometheus/tsdb"
+	"gopkg.in/alecthomas/kingpin.v2"
+	"gopkg.in/yaml.v2"
+)
+
+func registerChecks(m map[string]setupFunc, app *kingpin.Application, name string) {
+	cmd := app.Command(name, "Linting tools for Thanos")
+	registerRulesCheck(m, cmd, name)
+	return
```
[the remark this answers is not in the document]
FUSAKLA: removed thanks :)

FUSAKLA:
Thanks @GiedriusS, should be fixed and rebased on master now

FUSAKLA:
Thanks! Should be up to date now.

cmd/thanos/check.go
```diff
@@ -0,0 +1,119 @@
...
+	checkErrors := tsdb.MultiError{}
+
+	b, err := ioutil.ReadFile(filename)
+	if err != nil {
+		checkErrors.Add(err)
+		return 0, checkErrors
+	}
+
+	var rgs thanosrule.RuleGroups
+	if err := yaml.Unmarshal(b, &rgs); err != nil {
```
GiedriusS: One more small thing to think about: maybe it would be worth to be strict here too like in `promtool` and use `UnmarshalStrict`? That way we would ensure that no unknown fields are used.

GiedriusS: @FUSAKLA @bwplotka @povilasv what is your opinion on this?

FUSAKLA: Hm, the thing is Thanos rule uses also Unmarshal only. 
Promtool uses strict because Prometheus uses strict. 

Strict can help you with typos for example instead using some defaults and ignoring the misspelled field but I also like the consistency. 

Maybe in next PR make it strict in both cases and use the validation even on the Thanos rule start?

povilasv: IMO, let's do strict here and do a PR to change Thanos RULE to use strict. Nice catch :100:

FUSAKLA:
Ok sounds good to me.
Will switch it to the strict mode.

cmd/thanos/check.go
FUSAKLA: Ok sounds good to me.
Will switch it to the strict mode.

FUSAKLA: Taking it back.. currently this cannot be done because the unmarshaling is done in two steps. 
First unmarshals without strict only the struct with partial response and in second step unmarshals the imported prometheus rule group also with unstrict unmarshall so it passes the additional partial_response_strategy field.

I'll take a closer look but now it fails with the strict.

```go
rs := struct {
		String string `yaml:"partial_response_strategy"`
	}{}

	errMsg := fmt.Sprintf("failed to unmarshal 'partial_response_strategy'. Possible values are %s", strings.Join(storepb.PartialResponseStrategyValues, ","))
	if err := unmarshal(&rs); err != nil {
		return errors.Wrapf(err, errMsg)
	}

	rg := rulefmt.RuleGroup{}
	if err := unmarshal(&rg); err != nil {
		return errors.Wrapf(err, errMsg)
	}
```

povilasv: Ah I see. Maybe we can add our own RuleGroup struct which embedds prometheus one?

```
type RuleGroup struct{
    PartialResponseStrategy string `yaml:"partial_response_strategy"`
    rulefmt.RuleGroup
}
```

This would simplify code, and then we could do unmarshal in one pass + use strict. WDYT?

FUSAKLA: Uh.. looking into it more it's not that simple. `PartialResponseStrategy` is defined in protobuffer and is `int32` the `RuleGroup` defined in `rule.go` has overridden the `UnmarshalYAML` where the mapping from `WARN` to `0` is done. Also setting the default value if the `partial_response_strategy` is missing at all. 

Not sure if it is possible to add own `UnmarshalYAML` to the protobuf type? This could possibly simplify it? 

Maybe @bwplotka could add more info to this I remember we discussed this a bit when he was adding the partial response strategy.

FUSAKLA: Sorry, I didn't notice you changed the type to the `string`. 
This way it can be done but there will be two new structs instead using the Thanos rule ones.

FUSAKLA: Is this what you meant? https://github.com/improbable-eng/thanos/pull/1097/commits/864a74cafb1f301143f11bf5c88cd6333ce1413b

TBH I don't like making the lint and actual loading rules in rule more different but if we eventually unify it in some next PR than I'm OK with it I suppose.

povilasv: Yes, thats what I meant but I think you could do `var rgs []ThanosRuleGroup` instead of `var rgs ThanosRuleGroups`, no ? so then it's only 1 struct :)

GiedriusS:
Could you rebase it one more time? Then I'll give it one last review and I think we can merge this. Thank you for this work and the persistence, it's not easy

cmd/thanos/check.go
FUSAKLA: Great, I thing this cannot be done because I need to unmarshal  the `groups: ` key to the slice.
I can't unmarshal it to slice directly. Or maybe I misunderstood?

FUSAKLA:
Sorry for the inactivity, I'm back. 

@GiedriusS rebased so PTAL if you find a time :pray:, thanks!

FUSAKLA:
@GiedriusS not sure if this is a bug. It looks like the cmds return error and if the whole process returns any it's printed out by the and this is not using the logger.

For example if you run:
```
$ ./thanos rule --log.format=json
{"caller":"flags.go:87","level":"info","msg":"gossip is disabled","ts":"2019-05-27T21:12:00.055497911Z"}
rule command failed: Gossip is disabled and no --query parameter was given.
```
Yep, it's here I believe?
https://github.com/improbable-eng/thanos/blob/34ad98eda29fc4977462e3e4beaa023186fecdf3/cmd/thanos/main.go#L167

I can print it out using the logger if you want but that would affect all other thanos components but it works as expected so it up to you.
```
$ ./thanos rule --log.format=json
{"caller":"flags.go:87","level":"info","msg":"gossip is disabled","ts":"2019-05-27T21:21:19.432487894Z"}
{"caller":"main.go:168","err":"rule command failed: Gossip is disabled and no --query parameter was given.","level":"error","ts":"2019-05-27T21:21:19.432716717Z"}
```

GiedriusS:
> @GiedriusS not sure if this is a bug. It looks like the cmds return error and if the whole process returns any it's printed out by the and this is not using the logger.
> 
> For example if you run:
> 
> ```
> $ ./thanos rule --log.format=json
> {"caller":"flags.go:87","level":"info","msg":"gossip is disabled","ts":"2019-05-27T21:12:00.055497911Z"}
> rule command failed: Gossip is disabled and no --query parameter was given.
> ```
> 
> Yep, it's here I believe?
> 
> https://github.com/improbable-eng/thanos/blob/34ad98eda29fc4977462e3e4beaa023186fecdf3/cmd/thanos/main.go#L167
> 
> I can print it out using the logger if you want but that would affect all other thanos components but it works as expected so it up to you.
> 
> ```
> $ ./thanos rule --log.format=json
> {"caller":"flags.go:87","level":"info","msg":"gossip is disabled","ts":"2019-05-27T21:21:19.432487894Z"}
> {"caller":"main.go:168","err":"rule command failed: Gossip is disabled and no --query parameter was given.","level":"error","ts":"2019-05-27T21:21:19.432716717Z"}
> ```

Sorry, I haven't looked into this completely when commenting originally (didn't see that you haven't introduced this yourself) but yes, I think that should be removed and the proper logger should be used. I mean the user asks for a JSON formatted log so it should be printed like that, no? :smile:

cmd/thanos/check.go
povilasv: makes sense. all good :+1:

FUSAKLA:
@GiedriusS sorry for the delay. I changed the log and rebased on the recent master if that is ok with you now? I agree this way it's more consistent :)

FUSAKLA:
I rebased it once again, fixed issues with MultiError from TSDB which moved meantime to different package and updated the docs cmd flags which changed also.  CI is green :tada:

PR #1098, opened 2019-05-01
store: Compose indexCache properly allowing injection for testing purposes.

bwplotka:


pkg/store/bucket_e2e_test.go
```diff
@@ -9,6 +9,8 @@ import (
 	"testing"
 	"time"
 
+	"github.com/oklog/ulid"
+
```
povilasv: remove the spacey

bwplotka: will do in next PR.

PR #1104, opened 2019-05-02
Store: Add min resolution to be specific what downsampling levels to use on query

mjd95:
## Changes

Refactor the series interface to take a list of allowed resolutions instead of a max resolution.  This is to allowed viewing e.g. only 5m downsampled data, which is a useful debugging tool.  (c.f. https://github.com/improbable-eng/thanos/issues/922)

WIP as I'm currently having issues generating the protos via the Makefile (probably something about my setup).  Also wondering whether we should have stronger typing on the new fields (e.g., a `repeated enum` rather than a `repeated int64`).

## Verification

Only updated the unit tests to pass so far, happy to add more.


mjd95:
Let me know what you think of this approach @bwplotka

mjd95:
Thanks for the comments @povilasv @bwplotka 

Min source resolution sounds like a good idea.  It covers all the use cases (just debugging basically) I had for `allowedResolution` and it a bit more flexible as well.

GiedriusS:
I completely agree with @bwplotka but I just had one idea while working through getFor's code: we should probably prioritize the higher resolution data in general. For example, the user specifies that they want max 1h resolution and raw resolution as the minimum. In such a case, we should probably opt to first fill the requested time range with the raw resolution where possible and only then fill it up with the downsampled one. It would probably make more sense from the user's perspective if they are using a small range vector in their query. Such a change would give more results to the users compared to selecting the data with the worst resolution first.

Thoughts?

mjd95:
Changed to use the min resolution suggestion, which is much nicer :)  I also rebased on top of https://github.com/improbable-eng/thanos/pull/1132 for building the protos, but the main changes for this PR are still the ones to the store API.

@GiedriusS that's an interesting question whether we should prefer raw data or downsampled data.  Preferring the downsampled like we do now does mean we're sending less data overall which is good, but on the other hand having the raw data is more useful for the user.  I think what we're doing now is reasonable but would have to think about it more, there's definitely a trade-off here.

pkg/store/bucket.go
```diff
@@ -996,44 +996,43 @@ func int64index(s []int64, x int64) int {
...
 
 // getFor returns a time-ordered list of blocks that cover date between mint and maxt.
-// Blocks with the lowest resolution possible but not lower than the given resolution are returned.
-func (s *bucketBlockSet) getFor(mint, maxt, minResolution int64) (bs []*bucketBlock) {
-	if mint == maxt {
-		return nil
-	}
-
+// The covering is greedy, preferring blocks of the lowest allowed resolutions.
+func (s *bucketBlockSet) getFor(minTime, maxTime, minResolution, maxResolution int64) (bs []*bucketBlock) {
```
bwplotka: We change a big for this part, so wonder how it will look like after rebase, but generally it looks like something we want :+1:

pkg/store/bucket.go
```diff
@@ -996,44 +996,43 @@ func int64index(s []int64, x int64) int {
...
 
-	// Find first matching resolution.
-	i := 0
-	for ; i < len(s.resolutions) && s.resolutions[i] > minResolution; i++ {
+	for i := 0; i < len(s.resolutions); i++ {
+		if s.resolutions[i] >= minResolution && s.resolutions[i] <= maxResolution {
+			if len(bs) == 0 {
+				// Nothing covered so far, cover as much as possible of whole range with current resolution.
+				bs = coverInterval(minTime, maxTime, s.blocks[i])
+			} else {
```
bwplotka: As general tip: I would change `else` to just continue on above case, bit cleaner.

pkg/store/bucket.go
mjd95: 👍  My changes are very small now actually.  Previously I had refactored this function to not be recursive, but I don't think that it helpful any more (the recursion does a nice job of filling in the gaps).

mjd95:
Sorry for the delay, PTAL @bwplotka

mjd95:
Closing as the only current use case for adding this was debugging https://github.com/improbable-eng/thanos/issues/922, but there is a plan to add a UI for `thanos bucket web` which seems like it will cover this debugging case and more.

Discussed whether there might be any other use cases.  Potentially, but we don't have any FRs yet it's easy enough to add if we get one :)

bwplotka:
Thanks for this @mjd95 

We agreed offline that the debug use case that this is solving is solved in better way by this: https://github.com/improbable-eng/thanos/issues/1246 (cc @jaseemabid), so postponing this until more use cases appears. (:

PR #1107, opened 2019-05-02
Cut release v0.4.0

bwplotka:


CHANGELOG.md
```diff
@@ -20,6 +20,9 @@ See [this](docs/proposals/approved/201809_gossip-removal.md) for more details.
 
 ### Added
 
+- thanos.io website & automation :tada: 
+- [#1053](https://github.com/improbable-eng/thanos/pull/1053) Finally, compactor is able to handle restart in very ugly moment (partial upload). It will remove those blocks after safe 30 min time.
```
brancz: I understand the sentiment, but I think we should re-phrase this to something like this for the compactor: "Allow compactor to handle restarts in case of partial uploads. Left over blocks will be safely deleted 30 minutes afterwards, implicitly limiting the amount of time an upload can take to 30 minutes as well."

Also could you elaborate how this affects the store gateway more precisely?

CHANGELOG.md
```diff
@@ -20,6 +20,9 @@ See [this](docs/proposals/approved/201809_gossip-removal.md) for more details.
 
 ### Added
 
+- thanos.io website & automation :tada:
```
domgreen: `thanos.io` > `[thanos.io](https://thanos.io)`

CHANGELOG.md
domgreen: agree

PR #1118, opened 2019-05-06
add cross domain scoping

auhlig:
As described in https://github.com/improbable-eng/thanos/issues/1117

### DIFF
--- a/pkg/objstore/swift/swift.go
+++ b/pkg/objstore/swift/swift.go
@@ -11,14 +11,13 @@ import (
 	"testing"
 	"time"
 
-	"github.com/improbable-eng/thanos/pkg/objstore"
-
 	"github.com/go-kit/kit/log"
 	"github.com/gophercloud/gophercloud"
 	"github.com/gophercloud/gophercloud/openstack"
 	"github.com/gophercloud/gophercloud/openstack/objectstorage/v1/containers"
 	"github.com/gophercloud/gophercloud/openstack/objectstorage/v1/objects"
 	"github.com/gophercloud/gophercloud/pagination"
+	"github.com/improbable-eng/thanos/pkg/objstore"
 	"github.com/pkg/errors"
 	yaml "gopkg.in/yaml.v2"
 )
@@ -27,16 +26,20 @@ import (
 const DirDelim = "/"
 
 type SwiftConfig struct {
-	AuthUrl       string `yaml:"auth_url"`
-	Username      string `yaml:"username"`
-	UserId        string `yaml:"user_id"`
-	Password      string `yaml:"password"`
-	DomainId      string `yaml:"domain_id"`
-	DomainName    string `yaml:"domain_name"`
-	TenantID      string `yaml:"tenant_id"`
-	TenantName    string `yaml:"tenant_name"`
-	RegionName    string `yaml:"region_name"`
-	ContainerName string `yaml:"container_name"`
+	AuthUrl           string `yaml:"auth_url"`
+	Username          string `yaml:"username"`
+	UserDomainName    string `yaml:"user_domain_name"`
+	UserDomainID      string `yaml:"user_domain_id"`
+	UserId            string `yaml:"user_id"`
+	Password          string `yaml:"password"`
+	DomainId          string `yaml:"domain_id"`
+	DomainName        string `yaml:"domain_name"`
+	ProjectID         string `yaml:"project_id"`
+	ProjectName       string `yaml:"project_name"`
+	ProjectDomainID   string `yaml:"project_domain_id"`
+	ProjectDomainName string `yaml:"project_domain_name"`
+	RegionName        string `yaml:"region_name"`
+	ContainerName     string `yaml:"container_name"`
 }
 
 type Container struct {
@@ -46,23 +49,14 @@ type Container struct {
 }
 
 func NewContainer(logger log.Logger, conf []byte) (*Container, error) {
-	var sc SwiftConfig
-	if err := yaml.Unmarshal(conf, &sc); err != nil {
+	sc, err := parseConfig(conf)
+	if err != nil {
 		return nil, err
 	}
 
-	authOpts := gophercloud.AuthOptions{
-		IdentityEndpoint: sc.AuthUrl,
-		Username:         sc.Username,
-		UserID:           sc.UserId,
-		Password:         sc.Password,
-		DomainID:         sc.DomainId,
-		DomainName:       sc.DomainName,
-		TenantID:         sc.TenantID,
-		TenantName:       sc.TenantName,
-
-		// Allow Gophercloud to re-authenticate automatically.
-		AllowReauth: true,
+	authOpts, err := authOptsFromConfig(sc)
+	if err != nil {
+		return nil, err
 	}
 
 	provider, err := openstack.AuthenticatedClient(authOpts)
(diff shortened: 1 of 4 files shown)


pkg/objstore/swift/swift.go
```diff
@@ -12,6 +12,7 @@ import (
 	"time"
 
 	"github.com/improbable-eng/thanos/pkg/objstore"
+	yaml "gopkg.in/yaml.v2"
```
povilasv: please remove the blank line after this one

pkg/objstore/swift/swift.go
```diff
@@ -65,6 +76,36 @@ func NewContainer(logger log.Logger, conf []byte) (*Container, error) {
...
+		authOpts.TenantID = sc.ProjectID
+	}
+	if sc.ProjectName != "" {
+		authOpts.TenantName = sc.ProjectName
+	}
+
+	// Support for cross domain scoping (user in different domain than project)
+	// If a userDomainName or userDomainID is given, a user is scoped to this domain
+	// and the tenant (aka project) is expected to be in the domain given by domainName or domainID.
+	if sc.UserDomainName != "" {
```
povilasv: could we use switch here instead

```
switch {
case sc.UserDomaiName != "": 
...
case sc.UserDomainID != "":

}
```

pkg/objstore/swift/swift.go
```diff
@@ -65,6 +76,36 @@ func NewContainer(logger log.Logger, conf []byte) (*Container, error) {
...
+			DomainName: sc.DomainName,
+		}
+	} else if sc.UserDomainID != "" {
+		authOpts.DomainID = sc.UserDomainID
+		authOpts.Scope = &gophercloud.AuthScope{
+			DomainID: sc.DomainId,
+		}
+	}
+	if authOpts.Scope != nil {
+		if sc.TenantName != "" {
```
povilasv: could we use switch here.

pkg/objstore/swift/swift.go
auhlig: ran `goimports` again

pkg/objstore/swift/swift.go
auhlig: right. done

pkg/objstore/swift/swift.go
auhlig: done

pkg/objstore/swift/swift.go
povilasv: goimports doesn't solve the blank line issue, please remove the blank line after this

povilasv:
@auhlig you just commented, but didn't actually push the changes

pkg/objstore/swift/swift.go
auhlig: i meant removed blank line and did the goimports again

auhlig:
@povilasv github was down this morning -.-

pkg/objstore/swift/swift.go
```diff
@@ -58,13 +69,57 @@ func NewContainer(logger log.Logger, conf []byte) (*Container, error) {
...
+	// In Identity v3 the term `project` is used instead.
+	switch {
+	case sc.TenantName != "":
+		level.Warn(logger).Log("msg", "usage of tenant_name is deprecated. use project_name instead")
+		authOpts.TenantName = sc.TenantName
+	case sc.TenantID != "":
+		level.Warn(logger).Log("msg", "usage of tenant_id is deprecated. use project_id instead")
+		authOpts.TenantID = sc.TenantID
+	case sc.ProjectName != "" || sc.ProjectID != "":
+		level.Info(logger).Log("msg", "using project_id or project_name")
```
povilasv: I think we should remove this, it isn't a good message and IMO doesn't add any value

pkg/objstore/swift/swift.go
```diff
@@ -58,13 +69,57 @@ func NewContainer(logger log.Logger, conf []byte) (*Container, error) {
...
+		authOpts.Scope = &gophercloud.AuthScope{
+			DomainName: sc.ProjectDomainName,
+		}
+	case sc.ProjectDomainID != "":
+		authOpts.Scope = &gophercloud.AuthScope{
+			DomainID: sc.ProjectDomainID,
+		}
+	}
+	if authOpts.Scope != nil {
+		level.Info(logger).Log("msg", "scoping to domain, project")
```
povilasv: same here let's not log this message.

pkg/objstore/swift/swift.go
```diff
@@ -58,13 +69,57 @@ func NewContainer(logger log.Logger, conf []byte) (*Container, error) {
...
+		TenantID:         sc.ProjectID,
+		TenantName:       sc.ProjectName,
 
 		// Allow Gophercloud to re-authenticate automatically.
 		AllowReauth: true,
 	}
 
+	// The term `tenant` is used in the deprecated OpenStack Identity v2.
+	// In Identity v3 the term `project` is used instead.
+	switch {
```
povilasv: could you also add a couple of examples how or when to configure project id vs project name in the `docs/storage.md` file

pkg/objstore/swift/swift.go
auhlig: right. removed

pkg/objstore/swift/swift.go
auhlig: removed

pkg/objstore/swift/swift.go
auhlig: updated docs. added link to official documentation with examples

auhlig:
LGTY @povilasv?

povilasv:
@auhlig Sort of yes. Did you look at the tests? Would be amazing if we could add some proper testing here, so that when we change code we don't break things.

pkg/objstore/swift/swift.go
```diff
@@ -11,32 +11,43 @@ import (
...
+	DomainId          string `yaml:"domain_id"`
+	DomainName        string `yaml:"domain_name"`
+	ProjectID         string `yaml:"project_id"`
+	ProjectName       string `yaml:"project_name"`
+	ProjectDomainID   string `yaml:"project_domain_id"`
+	ProjectDomainName string `yaml:"project_domain_name"`
+	RegionName        string `yaml:"region_name"`
+	ContainerName     string `yaml:"container_name"`
+
+	// Deprecated: Please use `project_id` instead.
```
bwplotka: I would maybe kill this and use `yaml.UmarshalStrict` instead? (: We already documented the change so let's fail fast on umarshalling! (:

pkg/objstore/swift/swift.go
```diff
@@ -58,13 +69,54 @@ func NewContainer(logger log.Logger, conf []byte) (*Container, error) {
...
+		TenantName:       sc.ProjectName,
 
 		// Allow Gophercloud to re-authenticate automatically.
 		AllowReauth: true,
 	}
 
+	// The term `tenant` is used in the deprecated OpenStack Identity v2.
+	// In Identity v3 the term `project` is used instead.
+	switch {
+	case sc.TenantName != "":
```
bwplotka: let's use `yaml.UnmarshalStrict` instead?

pkg/objstore/swift/swift.go
```diff
@@ -180,13 +232,19 @@ func (c *Container) deleteContainer(name string) error {
...
+		Password:          os.Getenv("OS_PASSWORD"),
+		RegionName:        os.Getenv("OS_REGION_NAME"),
+		ContainerName:     os.Getenv("OS_CONTAINER_NAME"),
+		ProjectID:         os.Getenv("OS_PROJECT_ID"),
+		ProjectName:       os.Getenv("OS_PROJECT_NAME"),
+		UserDomainID:      os.Getenv("OS_USER_DOMAIN_ID"),
+		UserDomainName:    os.Getenv("OS_USER_DOMAIN_NAME"),
+		ProjectDomainID:   os.Getenv("OS_PROJET_DOMAIN_ID"),
+		ProjectDomainName: os.Getenv("OS_PROJECT_DOMAIN_NAME"),
+		TenantID:          os.Getenv("OS_TENANT_ID"),
```
bwplotka: Kill this

auhlig:
Made this a breaking change by dropping `tenant_id`, `tenant_name`. 
Extracted logic to `parseConfig`, `authOptsFromConfig` for easier unit testing.
Added unit tests.

LGTY @povilasv @bwplotka

auhlig:
Netlify doesn't seem to be in the mood to run the tests. CircleCI passed however.

pkg/objstore/swift/swift.go
auhlig: killed

pkg/objstore/swift/swift.go
auhlig: done

pkg/objstore/swift/swift.go
auhlig: sounds good. done

auhlig:
squashed into single commit and force pushed.
giving netlify another chance

auhlig:
Updated the changelog after some merge conflicts. Can this be merged @povilasv @bwplotka?

GiedriusS:
@auhlig so essentially with this PR we are getting rid of the support of the V2 API? I'm a bit anxious about this since, for example, some companies might still be running old versions of that API (*cough* not gonna point any fingers). Would it be hard to add some kind of backward compatibility?

auhlig:
@GiedriusS This PR does **not** drop Identity v2 support. It's still possible to authenticate using `project_id` or `project_name` instead of the deprecated terms `tenant_id`, `tenant_name`.
We only fix the terminology ( tenant became project with Identity v3) and add support for for authentication if a  user is in a different domain than the project.
In summary, the only breaking change is the terminology.

povilasv:
Ok so this does include a breaking change
`tenantID`, `tenantName` -> `projectID`, `projectName`.

I think in order not to break users, just fill tenantID to projectID and tenantName to projectName.

We try to avoid breaking changes when possible and in this case it's definitely doable

bwplotka:
@povilasv I think breaking here is better that having inconsistent configuration -> users will be immdiately notified by verbose error message.

bwplotka:
Waiting for @povilasv final LGTM to merge (:

auhlig:
Thank you guys 🎉

PR #1122, opened 2019-05-07
feat docs: added seznam.cz sponsors logo

FUSAKLA:
Add Seznam.cz logo to `used by` section.

We use Thanos in production for almost a year now running in Kubernetes using Minio as long term storage. 

More info about the company:
- https://o.seznam.cz/en/about-us/ 
- https://en.wikipedia.org/wiki/Seznam.cz

I know our company is locally based so there is not much of an importance from the world wide community view so feel free to remove the logo if more recognized companies kick in in case you accept it :)  

image


bwplotka:
No, this is fine (: What matters is your contributions and expierience while using Thanos (: especially with minio storage which is unusual.

PR #1132, opened 2019-05-10
fix proto generation

mjd95:
## Changes

Fix https://github.com/improbable-eng/thanos/issues/1131

### DIFF
--- a/scripts/genproto.sh
+++ b/scripts/genproto.sh
@@ -21,8 +21,7 @@ fi
 echo "installing gogofast"
 GO111MODULE=on go install "github.com/gogo/protobuf/protoc-gen-gogofast"
 
-THANOS_ROOT=$(pwd)
-PROM_PATH="${THANOS_ROOT}/pkg/store/prompb"
+PROM_PATH="$(pwd)/pkg/store/prompb"
 GOGOPROTO_ROOT="$(GO111MODULE=on go list -f '{{ .Dir }}' -m github.com/gogo/protobuf)"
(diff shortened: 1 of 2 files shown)


adrien-f:
Hi :wave: !

Thanks for the changes, I'm trying to test them here on OSX and first, it won't work if I don't `go get -u github.com/gogo/protobuf/version` before. And then, like you said, `cp` from BSD won't work.

Could you maybe look into what the Prometheus team did for this similar issue here: https://github.com/prometheus/prometheus/blob/master/scripts/genproto.sh I think it's better to move that to our own script :smile: !

mjd95:
Thanks @adrien-f, I agree.  I've updated our `genprotos.sh` to fetch the dependencies in the same way as the Prometheus one does and removed the hack from the `Makefile`.

scripts/genproto.sh
```diff
@@ -18,13 +18,17 @@ if ! [[ $(${PROTOC_BIN} --version) =~ "3.4.0" ]]; then
 	exit 255
 fi
 
+echo "installing gogofast"
+GO111MODULE=on go install "github.com/gogo/protobuf/protoc-gen-gogofast"
+
 THANOS_ROOT=$(pwd)
-PROM_PATH="${THANOS_ROOT}/pkg/store/storepb"
-GOGOPROTO_ROOT="${THANOS_ROOT}/vendor/github.com/gogo/protobuf"
+PROM_PATH="${THANOS_ROOT}/pkg/store/prompb"
```
domgreen: nit - inline `THANOS_ROOT`?

PR #1139, opened 2019-05-13
website: update padding around sponsors

adrien-f:
## Changes

website: update padding around sponsors


brancz:
Nice, yes I agree this looks better!

PR #1140, opened 2019-05-13
store/sidecar: use minio-go version that support ap-east-1

cjohnston-fuze:
## Changes

Update minio-go to pseudo-version that support new ap-east-1 region.

## Verification

Applied to v0.4.0 tag and confirmed its now connecting to ap-east-1 dualstack endpoint instead of the default us-east-1.


bwplotka:
have you run `go mod tidy` or `go mod verify` after change?

cjohnston-fuze:
Sorry I'm not very familiar with go. I do see the changes to go.sum are automatic and have made that commit.

GiedriusS:
If it's not a problem then maybe we could wait until the next official release? Then we'd be able to also add error-only tracing to our S3 client which is amazing :heart_eyes:

bwplotka:
Up to you - I am happy to brough this unless you are concerned @GiedriusS

xjewer:
>Then we'd be able to also add error-only tracing to our S3 client which is amazing 😍

and this change https://github.com/minio/minio-go/pull/1104

PR #1142, opened 2019-05-14
store/cache: drop updates to existing items and dedupe get/set methods

### LINKED ISSUE #651: store: sometimes pkg/store.(*indexCache).ensureFits spins forever, blocking other goroutines
Thanos components: improbable/thanos:v0.1.0
Prometheus: prom/prometheus:v2.4.2

**What happened**
image
Number of goroutines in Thanos Store instances started growing inexorably. It never came back down and Thanos Store was essentially unresponsive to all queries during this. 

Goroutine number to stay around the same number i.e. have a low standard deviation, and Thanos Store to be responsive.

To be done. since this is hard to reproduce and depends on the data.

Thanos Query started timing out:
```
level=error ts=2018-11-29T13:20:41.368751553Z caller=proxy.go:117 err="fetch series for [{monitor prtoss} {replica a}]: rpc error: code = Canceled desc = context canceled"
level=error ts=2018-11-29T13:20:41.368755808Z caller=proxy.go:117 err="fetch series for [{monitor prtoss} {replica b}]: rpc error: code = Canceled desc = context canceled"
level=error ts=2018-11-29T13:20:41.368808725Z caller=proxy.go:117 err="fetch series for []: rpc error: code = Canceled desc = context canceled"
level=warn ts=2018-11-29T13:20:41.368849077Z caller=proxy.go:126 err="No store matched for this query"
level=error ts=2018-11-29T13:20:41.368848275Z caller=proxy.go:117 err="fetch series for []: rpc error: code = Canceled desc = context canceled"
level=warn ts=2018-11-29T13:20:41.368909531Z caller=proxy.go:126 err="No store matched for this query"
level=error ts=2018-11-29T13:20:41.369833973Z caller=engine.go:499 msg="error expanding series set" err="context canceled"
```

No more `series query processed` messages in Thanos Store whatsoever, only `Blocks source resolution`. Only this one got through somehow in a period of about an hour:
```
Lap 29 14:07:42 foobar docker[21891]: level=debug ts=2018-11-29T13:07:42.784986336Z caller=bucket.go:802 msg="series query processed" stats="&{blocksQueried:0 postingsTouched:0 postingsTouchedSizeSum:0 postingsFetched:0 postingsFetchedSizeSum:0 postingsFetchCount:0 postingsFetchDurationSum:0 seriesTouched:0 seriesTouchedSizeSum:0 seriesFetched:0 seriesFetchedSizeSum:0 seriesFetchCount:0 seriesFetchDurationSum:0 chunksTouched:0 chunksTouchedSizeSum:0 chunksFetched:0 chunksFetchedSizeSum:0 chunksFetchCount:0 chunksFetchDurationSum:0 getAllDuration:877 mergedSeriesCount:0 mergedChunksCount:0 mergeDuration:1273}"
```
When something happened but exactly zero data was pulled down so maybe this is a *separate* issue?

The goroutine dump of Thanos Store at the moment of this freeze:
[goroutine_dmp.zip](https://github.com/improbable-eng/thanos/files/2632509/goroutine_dmp.zip)

Notice that exactly one goroutine is stuck in `ensureFits` and it locks the lock, thus preventing other goroutines from performing their work. 

**Workaround**: increased the index cache significantly and the problem went away. Perhaps this has to do with something that some index cache files are really large:

```
$ du -t 250M -hs /var/lib/thanos/*/index.cache.json
279M    /var/lib/thanos/01CXD021G0JHNRQWJKVKCHD3QR/index.cache.json
1,6G    /var/lib/thanos/01CXF4WQE8H27NZK0W4VEQB3XP/index.cache.json
1,5G    /var/lib/thanos/01CXFJT3R8J59MD4YC9AC0TPE5/index.cache.json
...
```

Environment:
* OS: CentOS Linux 7 (Core)
* Kernel: 3.10.0-862.11.6.el7.x86_64
* Docker: 18.03.1-ce, build 9ee9f40
* Memory: 256 GB

abursavich:
1. When an item is successfully updated, metrics and cache size tracking added the new version of the item but did not remove the old version of the item.

    For example, with a fresh cache, if you set the same item `m` times with each version having entry size `n`, the cache would think it had `m` items with current size `m*n` instead of `1` item with current size `n`.

    If you then emptied the cache by evicting every item (only one), the cache would think it had `m-1` items with current size `(m-1)*n`, instead of `0` items with current size `0`.

    This would allow the cache to get stuck in an eviction loop which persists even after the cache has been emptied.

2. When an item is unsuccessfully updated in the cache because the new version is too big, the old version was left in the cache.

## Changes

If an item exists in the cache, remove it and then re-add it.  It would be possible to update the metrics in place, skipping what is counted as an eviction, but the code would be more complicated and brittle.

## Verification

Added tests which previously failed and now pass.

Fixes #651

### DIFF
--- a/pkg/store/cache/cache.go
+++ b/pkg/store/cache/cache.go
@@ -190,6 +190,48 @@ func (c *IndexCache) onEvict(key, val interface{}) {
 	c.curSize -= entrySize
 }
 
+func (c *IndexCache) get(typ string, key cacheKey) ([]byte, bool) {
+	c.requests.WithLabelValues(typ).Inc()
+
+	c.mtx.Lock()
+	defer c.mtx.Unlock()
+
+	v, ok := c.lru.Get(key)
+	if !ok {
+		return nil, false
+	}
+	c.hits.WithLabelValues(typ).Inc()
+	return v.([]byte), true
+}
+
+func (c *IndexCache) set(typ string, key cacheKey, val []byte) {
+	var size = sliceHeaderSize + uint64(len(val))
+
+	c.mtx.Lock()
+	defer c.mtx.Unlock()
+
+	if _, ok := c.lru.Get(key); ok {
+		return
+	}
+
+	if !c.ensureFits(size, typ) {
+		c.overflow.WithLabelValues(typ).Inc()
+		return
+	}
+
+	// The caller may be passing in a sub-slice of a huge array. Copy the data
+	// to ensure we don't waste huge amounts of space for something small.
+	v := make([]byte, len(val))
+	copy(v, val)
+	c.lru.Add(key, v)
+
+	c.added.WithLabelValues(typ).Inc()
+	c.currentSize.WithLabelValues(typ).Add(float64(size))
+	c.totalCurrentSize.WithLabelValues(typ).Add(float64(size + key.size()))
+	c.current.WithLabelValues(typ).Inc()
+	c.curSize += size
+}
+
 // ensureFits tries to make sure that the passed slice will fit into the LRU cache.
 // Returns true if it will fit.
 func (c *IndexCache) ensureFits(size uint64, typ string) bool {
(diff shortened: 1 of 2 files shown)


GiedriusS:
Hmm, makes sense that we could do better when modifying an item but looking at the code and logically this should never happen AFAICT. It seems to me that the problem is that at first we check https://github.com/improbable-eng/thanos/blob/master/pkg/store/bucket.go#L1421 if there is an entry in the cache for that, and *only then* we add it https://github.com/improbable-eng/thanos/blob/master/pkg/store/bucket.go#L1493 to the cache. What do you think? (cc @bwplotka)

abursavich:
AFAICT, the "check" and "set" is not atomic so there may be many concurrent "checks" that miss the same items in the cache and then independently load and "set" them.

GiedriusS:
Yes so that's my point - there's no reason to perform all of this work of removing and adding the same item again if we can put some locking around that part to make it atomic. What is your opinion on that?

bwplotka:
Wow, have you found the actual fix?? Awesome, cannot wait for some time to dive into it and review! :heart: Thanks!

abursavich:
I think the cache should support setting the same key multiple times. Since it goes out of its way to support concurrent access, I'm less inclined to accept an argument in favor of pushing the problem onto the caller.

The lookup to remove the existing item is O(1), so it's not so bad. You can always add a [singleflight](https://godoc.org/golang.org/x/sync/singleflight) in the caller later if you want to prevent duplicated work.

bwplotka:
Awesome spotting of this edge case. As you said, Get & Set are not atomic , but we missed that during implementation and testing. 

For the fix. You current approach of always removing makes sense. It is simple and effective, but not that efficient as we are removing from LRU something was added milisecond before. Let's see if that's the best what we can do. 

**Let's agree on one invariant for this: I think this cache items are immutable. So there is nothing like old version of cache or new. Updating is not needed for the sake of it.**

I like the @GiedriusS idea of making it atomic, or use `singleflight`. We check for hit in cache then if not we agreggate + paritition + fetch + split + put in to cache. Quite many things in between. However making it atomic/singleflight might be very beneficial if 100 requests are touching the same postings. However on what thing to lock/singleflight on? Block? Particular posting (too fine grained)? We should look on it create a ticket and design, but let's fix this issue in simpler form like proposed here.

What about simple alternative to this - just try to GetPosting again just before Set in lock. O(1) as well. And this will ensure proper behaviour, but will still fetch postigns unncecessary. Your solution will also help if our invariant is wrong and cache is malformed.

Thoughts @GiedriusS @abursavich ? Happy to discuss offline on slack as well (:

abursavich:
I changed it to make cache items immutable by aborting the set if the key already exists (instead of replacing the existing value).

I also refactored the get and set methods to reduce duplication.

bwplotka:
Forgot about change log :man_facepalming:

PR #1144, opened 2019-05-15
query/api: properly pass downsampling param

GiedriusS:
As compact.ResolutionLevel{Raw,5m,1h} says the resolution levels are
expressed in milliseconds. Currently, parseDownsamplingParam() calls
parseDuration() and friends which express this value in nanoseconds.
Thus, we need to divide the value by 1000*1000 to have them in
milliseconds.

Add test cases to check all of this.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -11,6 +11,9 @@ We use *breaking* word for marking changes that are not backward compatible (rel
 
 ## Unreleased
 
+### Fixed
+- [#1144](https://github.com/improbable-eng/thanos/pull/1144) Query/API: properly pass the downsampling parameter. Before this, wrong max resolution of the metrics data might have been selected.
+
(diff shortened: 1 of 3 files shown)


brancz:
this lgtm, but we should note this change in the changelog

bwplotka:
Wow

bwplotka:
I think we should close like at least 3 tickets mentioning that choosing downsampling does not always work ;p

bwplotka:
Great job @GiedriusS

bwplotka:
This is wrong change IMO - the time is being divided to match miliseconds further down in stack ): Also I did not realize that we are trying to somehow adjust unit in vatiable with unit agnostic type - time.Duration .... :man_facepalming: ((: 

Hopefully we are on it here: https://github.com/improbable-eng/thanos/pull/1146 

Consequence is that downsampling is technically never working now.

Should we revert this maybe? Unless I am wrong @GiedriusS (:

PR #1145, opened 2019-05-15
Add Qonto logo

AlexisSellier:
Add Qonto as a Thanos user !


povilasv:
@AlexisSellier it is live https://thanos.io/ :1st_place_medal:

AlexisSellier:
Thanks a lot @povilasv

bwplotka:
Can you share optionally @AlexisSellier anything about setup size, deployment model, after and cloud provider if any? (:

AlexisSellier:
In summary, everything run on AWS. We have 6 Kubernetes cluster (not EKS but kubesprayed) with one prometheus per cluster. The sidecar is installed using the prometheus operator. 

I was asking myself if it was worth to write an article on our Blog to describe all the architecture (I do not know if there is a lot a material about Thanos on the web and do not want to be redundant). I could give more information that way. Tell me if it is worth for you in term of visibility.

PR #1146, opened 2019-05-15
store/bucket: make getFor() work with interleaved resolutions

GiedriusS:
Add interleaved resolutions test for getFor(). Ordinary users could run into such a case when, for example, Thanos Compact would downsample metrics but also the metrics with the better resolution would still be available. Fix this by making getFor() look for blocks "in the middle" as well.

Fix the getDownsamplingParam() interface (and rename it to getDownsamplingParamMillis()) so that it would return a value which will be ready to use by the querier code.

Add unit tests for all of that.

Verification: `go test` and Grafana dashboards actually started showing all of the data.


pkg/store/bucket.go
```diff
@@ -1030,6 +1032,19 @@ func (s *bucketBlockSet) getFor(mint, maxt, minResolution int64) (bs []*bucketBl
 	if len(bs) == 0 {
 		return s.getFor(mint, maxt, s.resolutions[i])
 	}
+
+	until := len(bs) - 1
+	for bsi := 0; bsi < until; bsi++ {
```
domgreen: `bsi`?

domgreen: `bsi` > `j`? imo easier to recognize that it's just a loop counter rather than a cryptic TLA ;)

GiedriusS: I agree, I don't remember why I chose such name. I think I had some kind of different version before and in that version I had a variable named `j` already so I chose a different one. Fixed now :smile:

GiedriusS:
Since this touches a critical part of the code I think it would be best to wait for @bwplotka to review this (and, of course, anyone else who wants :P)

pkg/query/querier.go
```diff
@@ -55,7 +55,7 @@ type queryable struct {
 
 // Querier returns a new storage querier against the underlying proxy store API.
 func (q *queryable) Querier(ctx context.Context, mint, maxt int64) (storage.Querier, error) {
-	return newQuerier(ctx, q.logger, mint, maxt, q.replicaLabel, q.proxy, q.deduplicate, int64(q.maxSourceResolution/time.Millisecond), q.partialResponse, q.warningReporter), nil
+	return newQuerier(ctx, q.logger, mint, maxt, q.replicaLabel, q.proxy, q.deduplicate, int64(q.maxSourceResolution), q.partialResponse, q.warningReporter), nil
```
bwplotka: Quite confusing - hard to tell for me if this makes sense (: 

This has the same unit as mint, maxt int64 right? Maybe making the underlying variable `maxSourceResolutionMiliseonconds` or keeping it time.Duration as long as possible makes sense?

bwplotka: Hm.. am I right that is PR was actually wrong? https://github.com/improbable-eng/thanos/pull/1144 And we moved to miliseconds there but we are doing exactly the same here and we missed it? :thinking: 

If yes, let's remove https://github.com/improbable-eng/thanos/blob/2c4f64b1b96907295a7f8e99d8fd64697f0eb12a/pkg/query/api/v1.go#L227 I think it's wrong. All is in time.Duration so moving to miliseconds there does not make sense. It's here that we transition from duration to integer having arbitrary unit (miliseconds) so I think this should stay and last PR should be reverted.

bwplotka: And yea, master is broken now in terms of downsampling if I am not wrong?

pkg/store/bucket.go
```diff
@@ -675,7 +675,7 @@ func populateChunk(out *storepb.AggrChunk, in chunkenc.Chunk, aggrs []storepb.Ag
 // labels and resolution. This is important because we allow mixed resolution results, so it is quite crucial
 // to be aware what exactly resolution we see on query.
 // TODO(bplotka): Consider adding resolution label to all results to propagate that info to UI and Query API.
-func debugFoundBlockSetOverview(logger log.Logger, mint, maxt int64, lset labels.Labels, bs []*bucketBlock) {
+func debugFoundBlockSetOverview(logger log.Logger, mint, maxt, maxresolution int64, lset labels.Labels, bs []*bucketBlock) {
```
bwplotka: is `maxresolution` really conforming Golang style guide? Can we fix it? (:

pkg/query/querier.go
bwplotka: Revert PR:https://github.com/improbable-eng/thanos/pull/1154

pkg/store/bucket.go
```diff
@@ -1030,6 +1032,19 @@ func (s *bucketBlockSet) getFor(mint, maxt, minResolution int64) (bs []*bucketBl
 	if len(bs) == 0 {
 		return s.getFor(mint, maxt, s.resolutions[i])
 	}
+
+	until := len(bs) - 1
+	for j := 0; j < until; j++ {
```
bwplotka: It's tricky, need to dive more but generally makes sense. Explaining what we are seeing now. Thanks for tests catching this as well! 

All of this assumes that blocks within different resolutions are aligned ideally right? Can we comment this somehow?

Also be aware of this: https://github.com/improbable-eng/thanos/pull/1104 Hope you algorthim would be nicely extendable I guess.
 cc @mjd95

GiedriusS: They always should be aligned ideally or on the same timestamps because in other cases we would get a problem known as "overlapping blocks".

This change should be easy to integrate into that PR.

pkg/store/bucket.go
GiedriusS: What name do you suggest?

pkg/query/querier.go
GiedriusS: Perhaps we should move the whole function to return `int64` and in a form appropriate to the caller since as we can see it is hard to make a mental connection between such distant places in the code? Because that's what parsing should do IMHO - we shouldn't do more parsing down the road.

GiedriusS: > And yea, master is broken now in terms of downsampling if I am not wrong?

I think that's a bit of an overstatement if it has never worked properly 😄

pkg/store/bucket.go
GiedriusS: Perhaps `maxResolution`?

pkg/query/querier.go
bwplotka: Sure but now it does not work even more, so there is regression (:  

Anyway, I just want to avoid confusion, sorry for being quite strict here ): 

Now once we had no regression we can actually make that work and nice. Yes, I am fine with:
* sticking to `maxResolutionMilisec int64` name everywhere.
* parseResolution returning int64 straight away.

Thoughts? (:

GiedriusS: Indeed that's what I've done. Hopefully it is much more clearer now about what kind of units are being used and that no one in the future will run into this problem like me when I was trying to hunt down one problem but accidentally got caught up in this.

pkg/store/bucket_test.go
```diff
@@ -16,6 +16,149 @@ import (
...
+func TestBucketBlockSet_Duplicated(t *testing.T) {
+	defer leaktest.CheckTimeout(t, 10*time.Second)()
+
+	set := newBucketBlockSet(labels.Labels{})
+
+	type resBlock struct {
+		mint, maxt int64
+		window     int64
+	}
+	input := []resBlock{
```
bwplotka: What do you think about joinging those together in form of table test? essentiall add those as test cases to `TestBucketBlockSet_addGet`?  might be more consistent and easier to read just "cases" as test itself feels similar TBH?

pkg/store/bucket.go
```diff
@@ -997,7 +997,7 @@ func int64index(s []int64, x int64) int {
 
 // getFor returns a time-ordered list of blocks that cover date between mint and maxt.
 // Blocks with the lowest resolution possible but not lower than the given resolution are returned.
```
bwplotka:
```suggestion
// Blocks with the biggest resolution possible but not bigger than the given max resolution are returned.
```

pkg/store/bucket.go
bwplotka: OK, It makes total sense BUT to me `left := s.getFor(mint, bs[0].meta.MinTime, s.resolutions[i])` and `right` are just `between` 0 and first min and lastMax and maxt, right?

Can we remove this left and right and treat is properly in this loop as `between`? I think it would simplify flow and algorithm.

GiedriusS: The idea sounds and I played around with it a bit but I doubt that this function can be made more succinct because we have three distinct cases of the `append` call: at the beginning of the array, between two elements, and at the end. Please correct me if I'm wrong.

pkg/store/bucket_test.go
GiedriusS: These ones are a bit special because we are doing a property-based tested instead of an ad-hoc one. I have removed the other test cases that I've added before and only left this one in - this already covers the "interleaved resolutions" case.

pkg/store/bucket.go
GiedriusS: :+1: fixed this comment, good catch

pkg/store/bucket.go
bwplotka: Hm, is it really too complex? I will propose something to your PR

pkg/store/bucket_test.go
bwplotka: :+1:

pkg/store/bucket.go
bwplotka: Proposed: https://github.com/GiedriusS/thanos/pull/1

IMO it's way simpler, what do you think? @GiedriusS 

```go
// getFor returns a time-ordered list of blocks that cover date between mint and maxt.
// Blocks with the biggest resolution possible but not bigger than the given max resolution are returned.
func (s *bucketBlockSet) getFor(mint, maxt, maxResolutionMillis int64) (bs []*bucketBlock) {
	if mint == maxt {
		return nil
	}

	s.mtx.RLock()
	defer s.mtx.RUnlock()

	// Find first matching resolution.
	i := 0
	for ; i < len(s.resolutions) && s.resolutions[i] > maxResolutionMillis; i++ {
	}

	// Fill the given interval with the blocks for the current resolution.
	// Our current resolution might not cover all data, so recursively fill the gaps with higher resolution blocks if there is any.
	start := mint
	for _, b := range s.blocks[i] {
		if b.meta.MaxTime <= mint {
			continue
		}
		if b.meta.MinTime >= maxt {
			break
		}

		if i+1 < len(s.resolutions) {
			bs = append(bs, s.getFor(start, b.meta.MinTime, s.resolutions[i+1])...)
		}
		bs = append(bs, b)
		start = b.meta.MaxTime
	}

	if i+1 < len(s.resolutions) {
		bs = append(bs, s.getFor(start, maxt, s.resolutions[i+1])...)
	}
	return bs
}
```

PR #1154, opened 2019-05-16
Revert "query/api: properly pass downsampling param (#1144)"

bwplotka:
If I am not wrong downsampling on master is broken for 2days.

This reverts commit 7a767eff3c7ecf55acbdc9d15a505207b2a0a7a2.

See: https://github.com/improbable-eng/thanos/pull/1144#issuecomment-493263066

Reason: 
* It has to be in in milisecond but we already do that underneath in Querier: https://github.com/improbable-eng/thanos/blob/3b9afb435215e761ecd34b3ccfd69d9946f6f725/pkg/query/querier.go#L58
* Conversion does not make sense here as time.Duration is unit agnostic type, so we should do conversion in the moment of transformation to `int64` only.

Tests are nice but they are assuming no-one converts resolution in further stack. Let's revert this and fix the getFor as it's done here: https://github.com/improbable-eng/thanos/pull/1146

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -30,10 +30,6 @@ We use *breaking* word for marking changes that are not backward compatible (rel
     - [ENHANCEMENT] Improve rule views by wrapping lines [PR #4702](https://github.com/prometheus/prometheus/pull/4702)
     - [ENHANCEMENT] Show rule evaluation errors on rules page [PR #4457](https://github.com/prometheus/prometheus/pull/4457)
     
-### Fixed
-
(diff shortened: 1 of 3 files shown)


brancz:
Just curious, wouldn't it make sense to rather remove the conversions lower in the stack?

GiedriusS:
That's what I was talking about in the other PR. Because otherwise IMHO we would just be kicking the can further down the road until someone in the future again will get confused about what actual unit is being used. So I propose just using `int64` and everywhere else consistently.

bwplotka:
Still, can we revert first and then talk about improvements? (:

PR #1156, opened 2019-05-17
Moved CI to build Thanos with golang 1.12.5

bwplotka:
This is to mitigate potential memory allocation latency bugs: https://github.com/prometheus/prometheus/issues/5524

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -4,7 +4,7 @@ jobs:
   test:
     docker:
       # Available from https://hub.docker.com/r/circleci/golang/
-      - image: circleci/golang:1.12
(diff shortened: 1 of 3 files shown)


bwplotka:
cc @sylr

sylr:
Is it really needed ? I suppose by now all Go1.12 images already include 1.12.5. I think we just need to rebuild.

bwplotka:
Can we confirm that is the case? 

Also I would be 10x more comfortable if we would pin to single version. Otherwise we have 0 record and control of what the Golang version was when building image. That's unacceptable IMO.

brancz:
I’d prefer exact versions as well.

PR #1157, opened 2019-05-17
query: Check duplicate external labels for all stores include existed store

godliness:
When we change the external labels after store has already joined into storesets that will makes the external labels duplicated.

## Changes

Check duplicate external labels before checking whether the store is exist or not.

## Verification

Run unit test which has passed.

/assign @bwplotka 

/cc @GiedriusS


godliness:
Does it could be merged? @bwplotka @GiedriusS

bwplotka:
Thanks!

PR #1162, opened 2019-05-17
Upgrade promu and get rid of VERSION file

sylr:
Let's use git tag and discard the VERSION file.

```
sylvain@ubuntu-1604-dev:~/gopath/src/github.com/improbable-eng/thanos[promu]$ ./thanos --version
thanos, version v0.4.0-12-gc642113-dirty (branch: promu, revision: c64211352a26c2dbbee9d598321cc42818c143c7)
  build user:       sylvain@ubuntu-1604-dev
  build date:       20190517-13:18:15
  go version:       go1.12.5
```


bwplotka:
Docs has to be updated (spotted by CI)

bwplotka:
Do you know how the promu build will look then now? 

So for if commit has some tag it's trivial - it will take tag. But what if we build with prom from master - from commit which does not have a tag? It will take last known tag? (kind of not what we want)

Also if we agree probably change in docs/release-proces.md is required as well.

brancz:
@simonpasquier might know :)

bwplotka:
Our point for changing some release e.g `v0.4.0` to `v0.4.0-master` right after release is to distinguish the release exact version with "something that is newer than release". It's useful for humans if someone reports issue. I guess it has always revision so maybe we can live without it (:

bwplotka:
BTW wonder if Thanos can reuse something from potential Circle CI orb @simonpasquier, so we can join forces.

simonpasquier:
> So if commit has some tag it's trivial - it will take tag. But what if we build with prom from master - from commit which does not have a tag? It will take last known tag? (kind of not what we want)

@sylr surely knows as he implemented the `promu` change :smirk:
If I'm not wrong, the version will be `<LAST TAG>@g<SHORT COMMIT ID>{-dirty}`.

Definitely up to share some of the CI burden :)

sylr:
Any commit on master that is not a tag will look like this:

`v<LAST_TAG>-<NUMBER_OF_COMMITS_SINCE_LAST_TAG>-g<COMMIT_ID>[-dirty]`

brancz:
Whops sorry about that, I didn’t mean to imply anything, I didn’t know that. That’s awesome!

bwplotka:
Thanks @sylr

PR #1165, opened 2019-05-17
Use go proxy to fasten building at first build (useful for docker & ci)

sylr:
```
sylvain@ubuntu-1604-dev:~/gopath/src/github.com/improbable-eng/thanos[master]$ time make docker-multi-stage DOCKER_IMAGE_NAME=quay.io/sylr/thanos:test
...
...

real    6m58.813s
user    0m0.372s
sys     0m0.352s
```

```
sylvain@ubuntu-1604-dev:~/gopath/src/github.com/improbable-eng/thanos[goproxy]$ time make docker-multi-stage DOCKER_IMAGE_NAME=quay.io/sylr/thanos:test
...
...

real    2m10.127s
user    0m0.376s
sys     0m0.268s
```

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -8,6 +8,8 @@ TMP_GOPATH        ?= /tmp/thanos-go
 GOBIN             ?= ${GOPATH}/bin
 GO111MODULE       ?= on
 export GO111MODULE
+GOPROXY           ?= https://proxy.golang.org
+export GOPROXY
 
 # Tools.
 EMBEDMD           ?= $(GOBIN)/embedmd-$(EMBEDMD_VERSION)


sylr:
I believe it’s hosted by google so as far as I am concerned it’s as “safe” as GitHub.

sylr:
```
Go Module Mirror, Index, and global go.sum

The Go team is providing the following services run by Google: a module mirror for accelerating Go module downloads, an index for discovering new modules, and a global go.sum database for authenticating module content.

Our privacy policy explains how we collect and use your information. The privacy policy for all of these services is proxy.golang.org/privacy.
```

GiedriusS:
The hashes are saved in our `go.sum` so should be safe - the only difference is that we will be downloading these files from a different place.

PR #1167, opened 2019-05-21
Add Jetstack to the users section

simonswine:
Spoke to @bwplotka today. We here at Jetstack are happy users of Thanos. 

Thanks for the hard work on it :+1:


bwplotka:
Thanks!

PR #1169, opened 2019-05-21
Add Meltwater to the users on the website

dezmodue:
## Changes
Added Meltwater to the users of Thanos in the website - as discussed with @bwplotka at kubecon


bwplotka:
Thanks!

Issue #1160, opened 2019-05-17
Use of json.Decoder


miekg:
We (also) seeing increased memory use (and still some leaks) in our thanos-store (v 0.4).

json.Decoder jumps out:
~~~
Showing top 10 nodes out of 18
      flat  flat%   sum%        cum   cum%
  700.75MB 30.94% 30.94%  2263.38MB 99.93%  github.com/improbable-eng/thanos/pkg/block.ReadIndexCache
  453.84MB 20.04% 50.98%   453.84MB 20.04%  encoding/json.(*Decoder).refill
  439.02MB 19.38% 70.36%   439.02MB 19.38%  encoding/json.(*decodeState).literalStore
  355.28MB 15.69% 86.05%   355.28MB 15.69%  reflect.mapassign
  225.06MB  9.94% 95.99%   225.06MB  9.94%  reflect.unsafe_NewArray
   71.43MB  3.15% 99.14%    71.43MB  3.15%  github.com/improbable-eng/thanos/pkg/block.ReadIndexCache.func1
       8MB  0.35% 99.49%  1037.36MB 45.80%  encoding/json.(*decodeState).object
    0.50MB 0.022% 99.51%   225.56MB  9.96%  reflect.MakeSlice
         0     0% 99.51%  1491.21MB 65.84%  encoding/json.(*Decoder).Decode
         0     0% 99.51%   453.84MB 20.04%  encoding/json.(*Decoder).readValue 
~~~

Reading this was illuminating: https://ahmet.im/blog/golang-json-decoder-pitfalls/ (although targets older version of Go [1.7]).

We're planning to rip out `json.Decoder` and move everything to `json.Unmarshal` and see how that fares.

bwplotka:
Yea, so there is overall quite large of improvements we can make to Thanos Store https://github.com/improbable-eng/thanos/issues/448

I fully agree here with Decoder not being designed for this :+1:  But I think all those 3 flaws are either fixed (not draining connections) in Go1.12 or not affecting us really (ignoring errs, "it's not for non stream"), am I wrong?

brancz:
If we do consider optimizing json decoding we should probably consider [json-iterator](https://github.com/json-iterator/go).

GiedriusS:
[...]

PR #1173, opened 2019-05-23
store,compact,sidecar: replace all usage of json.Decoder with json.Unmarshal

tmhdgsn:
fixes issue #1160 

## Changes

replaced usages of json.NewDecoder Decode, with ioutil.ReadAll/File and json.Unmarshal

## Verification

been running in production for a few days, memory graphs appear marginally better fixes

### DIFF
--- a/cmd/thanos/compact.go
+++ b/cmd/thanos/compact.go
@@ -364,7 +364,7 @@ func genMissingIndexCacheFiles(ctx context.Context, logger log.Logger, bkt objst
 			return errors.Wrap(err, "read meta")
 		}
 
-		if err = json.Unmarshal(obj, meta); err != nil {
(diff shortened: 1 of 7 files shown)


GiedriusS:
On my cluster the change is inconsequential:

Screenshot from 2019-05-23 17-09-10

The difference is only ~50MiB which you can attribute to the different queries which come into my cluster since they are load-balanced. My cluster has ~2TiB data so it's not a small one :( 

What's the difference on your cluster with this change? Do you have any graphs?

GiedriusS:
It is definitely answering to queries hence the sharp-tooth pattern that you see (: that is Thanos Store, by the way. Either way, technically it is a good change.

PR #1182, opened 2019-05-28
Track promql query errors with logs, tracing spans and metrics

jaseemabid:
Occasionally the memory usage across thanos components spikes up a lot and we've
usually had a hard time finding the offending query that lead to it. This is to
help debug and find bad queries easily. All queries are logged with debug level,
and `qry.Exec()` errors with an additional error label.

A new metric `thanos_query_api_failures_total` keeps track of the query error
count. Along with `instantQueryDuration + rangeQueryDuration` this should allow
us to compute the percentage of promql queries that fail.

Example log line:

```
level=debug ts=2019-05-28T16:46:36.10666776Z caller=v1.go:307 query="topk(10, count by (__name__)({__name__=~\".+\"}))" error="query was canceled in expression evaluation" duration=1.400047035
```

### DIFF
--- a/pkg/query/api/v1.go
+++ b/pkg/query/api/v1.go
@@ -29,6 +29,7 @@ import (
 	"github.com/NYTimes/gziphandler"
 
 	"github.com/go-kit/kit/log"
+	"github.com/go-kit/kit/log/level"
 	"github.com/improbable-eng/thanos/pkg/query"
 	"github.com/improbable-eng/thanos/pkg/runutil"
 	"github.com/improbable-eng/thanos/pkg/tracing"
@@ -58,6 +59,7 @@ const (
 	errorCanceled           = "canceled"
 	errorExec               = "execution"
 	errorBadData            = "bad_data"
+	errorTooMany            = "too_many_samples"
 	ErrorInternal           = "internal"
 )
 
@@ -101,8 +103,10 @@ type API struct {
 	queryableCreate query.QueryableCreator
 	queryEngine     *promql.Engine
 
-	instantQueryDuration   prometheus.Histogram
-	rangeQueryDuration     prometheus.Histogram
+	instantQueryDuration prometheus.Histogram
+	rangeQueryDuration   prometheus.Histogram
+	// Total number of query failures grouped by type of error.
+	queryFailures          *prometheus.CounterVec
 	enableAutodownsampling bool
 	enablePartialResponse  bool
 	now                    func() time.Time
@@ -131,6 +135,10 @@ func NewAPI(
 			0.05, 0.1, 0.25, 0.6, 1, 2, 3.5, 5, 7.5, 10, 15, 20,
 		},
 	})
+	queryFailures := prometheus.NewCounterVec(prometheus.CounterOpts{
+		Name: "thanos_query_api_failures_total",
+		Help: "Total number of failed query operations.",
+	}, []string{"error"})
 
 	reg.MustRegister(
 		instantQueryDuration,
@@ -142,6 +150,7 @@ func NewAPI(
 		queryableCreate:        c,
 		instantQueryDuration:   instantQueryDuration,
 		rangeQueryDuration:     rangeQueryDuration,
+		queryFailures:          queryFailures,
 		enableAutodownsampling: enableAutodownsampling,
 		enablePartialResponse:  enablePartialResponse,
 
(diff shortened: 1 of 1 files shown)


brancz:
How about we instead add the query as a tag to the trace span?

jaseemabid:
@brancz That's a good idea! I've updated the patch. We are not tracing thanos in our env yet, so I've not tested that part. 

The logs seems to be working alright, but I'm yet to see it being useful because I'vent had a OOM kill after this deployment.

brancz:
If that's the problem, then probably adding a debug log line that always logs every query before actually starting to execute it sounds like a more useful place.

jaseemabid:
@brancz Updated. All queries are logged with debug level, ones that failed are additionally logged in warning level as well.

pkg/query/api/v1.go
```diff
@@ -288,18 +290,24 @@ func (api *API) query(r *http.Request) (interface{}, []error, *ApiError) {
...
 	// We are starting promQL tracing span here, because we have no control over promQL code.
 	span, ctx := tracing.StartSpan(r.Context(), "promql_instant_query")
+	span.SetTag("query", raw)
 	defer span.Finish()
 
 	begin := api.now()
-	qry, err := api.queryEngine.NewInstantQuery(api.queryableCreate(enableDedup, 0, enablePartialResponse, warningReporter), r.FormValue("query"), ts)
+	qry, err := api.queryEngine.NewInstantQuery(api.queryableCreate(enableDedup, 0, enablePartialResponse, warningReporter), raw, ts)
 	if err != nil {
+		level.Warn(api.logger).Log("query", raw, "error", errorBadData)
```
brancz: I'm not so sure about this one. This will fail when someone sends an invalid query, which will happen a lot and this will pollute the logs.

pkg/query/api/v1.go
```diff
@@ -288,18 +290,24 @@ func (api *API) query(r *http.Request) (interface{}, []error, *ApiError) {
...
 		return nil, nil, &ApiError{errorBadData, err}
 	}
 
 	res := qry.Exec(ctx)
 	if res.Err != nil {
+		level.Warn(api.logger).Log("query", raw, "error", res.Err.Error(), "duration", time.Since(begin).Seconds())
```
brancz: I think even this should be debug level.

jaseemabid:
@brancz Changed both to debug.

pkg/query/api/v1.go
```diff
@@ -26,6 +26,8 @@ import (
 	"sync"
 	"time"
 
+	"github.com/go-kit/kit/log/level"
+
```
bwplotka: lets' kill this new line and rerun `make format` (:

pkg/query/api/v1.go
```diff
@@ -26,6 +26,8 @@ import (
...
 	"github.com/NYTimes/gziphandler"
```
bwplotka: and this

pkg/query/api/v1.go
```diff
@@ -288,18 +290,24 @@ func (api *API) query(r *http.Request) (interface{}, []error, *ApiError) {
...
-	qry, err := api.queryEngine.NewInstantQuery(api.queryableCreate(enableDedup, 0, enablePartialResponse, warningReporter), r.FormValue("query"), ts)
+	qry, err := api.queryEngine.NewInstantQuery(api.queryableCreate(enableDedup, 0, enablePartialResponse, warningReporter), raw, ts)
 	if err != nil {
+		level.Debug(api.logger).Log("query", raw, "error", errorBadData)
 		return nil, nil, &ApiError{errorBadData, err}
 	}
 
 	res := qry.Exec(ctx)
 	if res.Err != nil {
+		level.Debug(api.logger).Log("query", raw, "error", res.Err.Error(), "duration", time.Since(begin).Seconds())
```
bwplotka: You already have the query printed above, so maybe those log lines on errors (debug) are enough?

jaseemabid:
@bwplotka @brancz Just to make sure we are on the same page, 

1. log every query in debug level; this will help us find bad queries before memory spikes
2. Ignore `errorBadData`; this is just bad queries which should just be ignored as this wont cause prometheus/thanos issues.  
3. Log all `qry.Exec()` errors as debug with an additional `error` label; so this can be grepped and analysed. 

Is that OK? 👍

jaseemabid:
Adding a new type `errorTooMany`, because it was bundled with exec error and this might be the cause for memory spikes in my environment. 

Refactored the type cast into a small function, because its used in a couple of places. 

I think `return nil, nil, &ApiError{errorExec, res.Err}` in the original code might be dead code, because the case match just before covered all cases? I've explicitly commented this in the helper function.

jaseemabid:
Sample metrics: 

```
thanos_query_api_failures_total{app="thanos-query",error="canceled",instance="thanos-query-68878f4f75-psb9t",job="monitoring/thanos-query",namespace="monitoring"}
thanos_query_api_failures_total{app="thanos-query",error="timeout",instance="thanos-query-68878f4f75-psb9t",job="monitoring/thanos-query",namespace="monitoring"}
```

pkg/query/api/v1.go
```diff
@@ -101,8 +103,10 @@ type API struct {
 	queryableCreate query.QueryableCreator
 	queryEngine     *promql.Engine
 
-	instantQueryDuration   prometheus.Histogram
-	rangeQueryDuration     prometheus.Histogram
+	instantQueryDuration prometheus.Histogram
+	rangeQueryDuration   prometheus.Histogram
+	// Total number of query failures grouped by type of error.
+	queryFailures          *prometheus.CounterVec
```
bwplotka: This is a bit of duplicate to the gRPC server metrics, so not sure. I guess GRPC will not show us per error type.

pkg/query/api/v1.go
```diff
@@ -288,27 +297,30 @@ func (api *API) query(r *http.Request) (interface{}, []error, *ApiError) {
 		warnmtx.Unlock()
 	}
 
+	raw := r.FormValue("query")
+	level.Debug(api.logger).Log("query", raw)
```
bwplotka: I don't really like the duplication here. We log twice in error case. Fix fo this would be to have this log line at the end on success and keep those on error case on error, can we do this?

The question is do we really want to log query on every (even sucessful) request, not sure, I don't see much of benefit but we can start with that.

pkg/query/api/v1.go
```diff
@@ -288,27 +297,30 @@ func (api *API) query(r *http.Request) (interface{}, []error, *ApiError) {
...
-		switch res.Err.(type) {
-		case promql.ErrQueryCanceled:
-			return nil, nil, &ApiError{errorCanceled, res.Err}
-		case promql.ErrQueryTimeout:
-			return nil, nil, &ApiError{errorTimeout, res.Err}
-		case promql.ErrStorage:
-			return nil, nil, &ApiError{ErrorInternal, res.Err}
-		}
-		return nil, nil, &ApiError{errorExec, res.Err}
+		// Cast Prometheus errors to Thanos error constants
```
bwplotka: I think this comment is not needed

pkg/query/api/v1.go
bwplotka: Also what about having separate to `range` and `instant` to be consistent with duration historgrams above?

pkg/query/api/v1.go
```diff
@@ -288,27 +297,30 @@ func (api *API) query(r *http.Request) (interface{}, []error, *ApiError) {
...
+		e := toThanosError(res.Err)
+
+		level.Debug(api.logger).Log("query", raw, "error", res.Err.Error(), "duration", time.Since(begin).Seconds())
+		api.queryFailures.With(prometheus.Labels{"error": string(e)}).Inc()
+
+		return nil, nil, &ApiError{e, res.Err}
```
bwplotka: This looks like wrong error handling to me as it's easy to make mistake and log/use metric on caller side, but I guess it's good enough for now.

pkg/query/api/v1.go
```diff
@@ -649,3 +661,20 @@ func (api *API) labelNames(r *http.Request) (interface{}, []error, *ApiError) {
 
 	return names, warnings, nil
 }
+
+// toThanosError converts a Prometheus execute error into one of the Thanos error constants and returns a generic
+// `errorExec` for unknown types.
+func toThanosError(e error) ErrorType {
```
bwplotka: let's be explicit maybe?
```suggestion
func promqlToThanosErr(e error) ErrorType {
```

jaseemabid:
We were hoping that extensively logging queries will help us find the ones that lead to memory spikes but in practice it didn't.  Closing this off.

PR #1183, opened 2019-05-28
Propagate retriable/haltable errors from compactor

mattrco:
## Changes

We observed thanos compactor was exiting on upload errors. Previously, a new 
error was created from all worker errors so the custom type was not propagated.
This introduces `go-multierror` to preserve the error types.

If all errors returned are retriable, it is safe to retry the compaction.

Since the compactor may still return a single error this logic is preserved.

(Happy to refactor out the duplicated logic with individual error checks.)

## Verification

Tests are passing and I'll run this change with real workloads shortly.

### DIFF
--- a/cmd/thanos/compact.go
+++ b/cmd/thanos/compact.go
@@ -285,9 +285,9 @@ func runCompact(
 			if err == nil {
 				return nil
 			}
+
 			// The HaltError type signals that we hit a critical bug and should block
-			// for investigation.
-			// You should alert on this being halted.
+			// for investigation. You should alert on this being halted.
 			if compact.IsHaltError(err) {
 				if haltOnError {
 					level.Error(logger).Log("msg", "critical error detected; halting", "err", err)
@@ -299,7 +299,7 @@ func runCompact(
 			}
 
 			// The RetryError signals that we hit an retriable error (transient error, no connection).
-			// You should alert on this being triggered to frequently.
+			// You should alert on this being triggered too frequently.
 			if compact.IsRetryError(err) {
 				level.Error(logger).Log("msg", "retriable error", "err", err)
 				retried.Inc()
--- a/pkg/compact/compact.go
+++ b/pkg/compact/compact.go
@@ -8,15 +8,13 @@ import (
 	"path"
 	"path/filepath"
 	"sort"
-	"strings"
 	"sync"
 	"time"
-
-	"github.com/improbable-eng/thanos/pkg/block/metadata"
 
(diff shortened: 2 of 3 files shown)


mattrco:
Also, although this will fix compact exiting, the work it the compactor has written to disk is thrown away when the compaction is retried. I'm thinking about a good way to avoid this - the simplest thing would be to just retry the bucket requests more times before returning an error.

GiedriusS:
`prometheus/tsdb` already has a struct for this `tsdb.MultiError` that we use in some places already. Would it be possible to reuse it here?

mattrco:
@GiedriusS sure, I can use that instead 👍

mattrco:
Updated, I'll squash the commits if we're happy to proceed.

cmd/thanos/compact.go
```diff
@@ -284,9 +284,35 @@ func runCompact(
 			if err == nil {
 				return nil
 			}
+
+			if multiErr, ok := err.(tsdb.MultiError); ok {
```
bwplotka: So this to me looks like a bit of duplication. Can we make sure `IsHaltError` checks if error is multierror and then checks if any has halt? (: And same for retry?

cmd/thanos/compact.go
```diff
@@ -284,9 +284,35 @@ func runCompact(
...
+				retriable := true
+
+				for _, err := range multiErr {
+					if compact.IsHaltError(err) {
+						if haltOnError {
+							level.Error(logger).Log("msg", "critical error detected; halting", "err", err)
+							halted.Set(1)
+							select {}
```
povilasv: I'm not sure if this is the best approach? Do we really want to leave pod hanging on the error?

cmd/thanos/compact.go
```diff
@@ -284,9 +284,35 @@ func runCompact(
...
+						} else {
```
povilasv: we can remove this else

cmd/thanos/compact.go
mattrco: @bwplotka great idea, I've updated the error checking functions to handle multierrors 👍 Hopefully this is good to go now.

cmd/thanos/compact.go
mattrco: @povilasv appreciate the review - this is duplicated code that I've removed now 👍

mattrco:
@bwplotka @povilasv if you have a moment to take another look at this, that'd be much appreciated 🙇

PR #1187, opened 2019-05-28
.*: initial staticcheck implementation

GiedriusS:

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -24,6 +24,7 @@ jobs:
             fi
       - run: make check-go-mod
       - run: make errcheck
+      - run: make staticcheck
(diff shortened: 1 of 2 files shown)


brancz:
As suggested here https://github.com/improbable-eng/thanos/pull/1184#issuecomment-496614283, I think we should just use [golangci](https://golangci.com/), as we will inevitably add all of them.

GiedriusS:
Yes, it looks good from the onset however I have never used it. Could we discuss a bit about it in #thanos-dev when you'll have time?

GiedriusS:
Oh, that definitely looks good 👍 we should really probably migrate to it

PR #1194, opened 2019-05-30
correct service-discovery.md bullet points

swtch1:
## Changes
The site is rendering the bullet points incorrectly.  Adding a space before them should correct the issue.

## Verification
I am unable to verify this directly, but it appears you are using Hugo, and this is a known issue for Hugo.


bwplotka:
> I am unable to verify this directly, but it appears you are using Hugo, and this is a known issue for Hugo.

We use netlify for preview, available here: https://deploy-preview-1194--thanos-io.netlify.com/

PR #1200, opened 2019-05-31
Fix Thanos Query hang

povilasv:
## Changes

https://github.com/improbable-eng/thanos/issues/705

defer kicks in  and waits forever for other store to complete

https://github.com/improbable-eng/thanos/blob/master/pkg/store/proxy.go#L166

But actually what we would like is to cancel the queries to other stores.

So I've added a query level context, which is parent for all calls to Store APIs, once parent context is canceled this should cancel all children contexts (individual calls to Store APIs) and return response.

## Verification

TODO

### DIFF
--- a/pkg/store/proxy.go
+++ b/pkg/store/proxy.go
@@ -162,7 +162,9 @@ func (s *ProxyStore) Series(r *storepb.SeriesRequest, srv storepb.Store_SeriesSe
 			wg = &sync.WaitGroup{}
 		)
 
+		queryCtx, cancelQuery := context.WithCancel(gctx)
 		defer func() {
+			cancelQuery()
 			wg.Wait()
 			closeFn()
 		}()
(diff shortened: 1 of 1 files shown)


pkg/store/proxy.go
```diff
@@ -178,10 +180,10 @@ func (s *ProxyStore) Series(r *storepb.SeriesRequest, srv storepb.Store_SeriesSe
 			storeDebugMsgs = append(storeDebugMsgs, fmt.Sprintf("store %s queried", st))
 
 			// This is used to cancel this stream when one operations takes too long.
-			seriesCtx, closeSeries := context.WithCancel(gctx)
-			defer closeSeries()
+			storeCtx, cancelStore := context.WithCancel(queryCtx)
+			defer cancelStore()
```
bwplotka: We already defer this which should cancel all so wg.Wait should not block forever unless there is a bug inside streamSeriesSet

povilasv:
NVM I don't think this will actually fix it. As I was running old version, which didn't had:
```
s.recvCh <- r.GetSeries():
```
instead of
```
			select {
			case s.recvCh <- r.GetSeries():
				continue
			case <-ctx.Done():
				return
			}

```

PR #1201, opened 2019-05-31
 go mod tidy.

bwplotka:
cc @sylr 


sylr:
`go mod tidy` should also apply changes to go.sum I believe.

bwplotka:
Good point, done. I thought we have CI verifying this, looks it does not catch enough.

PR #1203, opened 2019-05-31
Fix Thanos Query hang

povilasv:

### DIFF
--- a/pkg/store/proxy.go
+++ b/pkg/store/proxy.go
@@ -220,9 +220,18 @@ func (s *ProxyStore) Series(r *storepb.SeriesRequest, srv storepb.Store_SeriesSe
 		return mergedSet.Err()
 	})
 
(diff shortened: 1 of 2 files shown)


pkg/store/proxy.go
```diff
@@ -220,9 +220,18 @@ func (s *ProxyStore) Series(r *storepb.SeriesRequest, srv storepb.Store_SeriesSe
 		return mergedSet.Err()
 	})
 
-	for resp := range respRecv {
-		if err := srv.Send(resp); err != nil {
-			return status.Error(codes.Unknown, errors.Wrap(err, "send series response").Error())
+OuterLoop:
+	for {
+		select {
+		case resp, ok := <-respRecv:
```
bwplotka: We could do so, but are you doing that "just in case" or actually there is some case that requires it? Becasue I cannot see any ):

PR #1208, opened 2019-05-31
Use prometheus v2.10.0

sylr:
Signed-off-by: Sylvain Rabot <s.rabot@lectra.com>


brancz:
Not against this at all, but just trying to understand why we're making this change. Is it just so we are actually testing against this version? (I'd be happy with that, just trying to understand)

sylr:
I think the amount of changes needed in thanos if we keep track of prometheus is smaller and easier to validate.

If we let prometheus in the same version and decide to bump it to let's 5~10 versions then the amount of work to incorporate the new version grows consequently and is prone to mistake.

pkg/util/labels.go
```diff
@@ -0,0 +1,16 @@
+package util
```
sylr: pkg/util/labels.go what maybe not the good choice.

What about a new pkg/utils tree where we would move all xxxutil packages, ie.:

- pkg/utils/labelutil/*.go
- pkg/utils/runutil/*.go
- pkg/utils/strutil/*.go
...

What do you think @bwplotka ?

brancz:
> If we let prometheus in the same version and decide to bump it to let's 5~10 versions then the amount of work to incorporate the new version grows consequently and is prone to mistakes.

Perfectly happy with that, and totally agree!

pkg/util/labels.go
brancz: just flattening this to 

* pkg/labelutil/*.go
* pkg/runutil/*.go
* pkg/strutil/*.go

sounds good to me as well.

bwplotka:
Yea, testing is not the case here, only to use latest fixes and be up to date. Fully agree with @sylr that it's painful and more riskt to do it rarely.

Thanks for doing it!

PR #1216, opened 2019-06-02
Remove command-line flags page from query UI

### LINKED ISSUE #248: Command-line flags page for query instances is not populated
The 'command-line flags' page is visible in the UI for query instances, but is not populated.

If the UI is only enabled on the query instances, perhaps it's not worth including this in the UI and would be better as an API endpoint?

It's inconsistent to expose this information for query instances but not other instance types, so maybe we should remove it entirely.

AlexisSellier:
This PR closes #248 

## Changes

I removed the command line flags panel from the query UI. Since there is now only one section to the **Status** drop-down, I removed it to only keep the **Runtime & Build Information** panel in the global menu


bwplotka:
Awesome, thanks for this! I think it makes sense but I would vote for not changing the `Status` page more.
The reason is that we copy the UI from Prometheus and it easier if we do as small number of our changes as possible.

Can we just remove command line flags? (:

AlexisSellier:
Ok i will leave the status page as it is in master

AlexisSellier:
@bwplotka status dropdown is back

PR #1217, opened 2019-06-03
pkg/receive: basic hashring support

squat:
## Changes

This PR adds a basic hashring type to the receive package. The hashring does not implement consistent hashing because today we do not care about needing to shift data around between scale-up/scale-down. The current plan is that at every scale event we flush all data, write it to block storage, and start again.

Going forward, a PR will embed a Hashring into the Handler struct so that when handling requests, the Handler can decide if it needs to handle the request itself or forward it to another host.
Forwarding functionality will be added in a follow up.

## Verification

Unit testing of hashring.

cc @metalmatze @brancz

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -5,6 +5,7 @@ require (
 	github.com/Azure/azure-storage-blob-go v0.0.0-20181022225951-5152f14ace1c
 	github.com/NYTimes/gziphandler v1.1.1
 	github.com/armon/go-metrics v0.0.0-20180917152333-f0300d1749da
+	github.com/cespare/xxhash v1.1.0
 	github.com/fatih/structtag v1.0.0
 	github.com/fortytw2/leaktest v1.3.0
 	github.com/fsnotify/fsnotify v1.4.7
--- a/pkg/receive/hashring.go
+++ b/pkg/receive/hashring.go
@@ -1,0 +1,114 @@
+package receive
+
+import (
+	"errors"
+	"sort"
+
+	"github.com/improbable-eng/thanos/pkg/store/prompb"
+
+	"github.com/cespare/xxhash"
+	"github.com/prometheus/common/model"
+	"github.com/prometheus/prometheus/discovery/targetgroup"
+)
+
+const sep = '\xff'
+
+// Hashring finds the correct host to handle a given time series
+// for a specified tenant.
+// It returns the hostname and any error encountered.
+type Hashring interface {
+	GetHost(tenant string, timeSeries *prompb.TimeSeries) (string, error)
+}
+
+// Matcher determines whether or tenant matches a hashring.
+type Matcher interface {
+	Match(tenant string, hashring string) bool
+}
+
+// MultiMatcher is a list of Matchers that implements the Matcher interface.
+type MultiMatcher []Matcher
+
+// Match implements the Matcher interface.
+func (m MultiMatcher) Match(tenant, hashring string) bool {
+	for i := range m {
+		if !m[i].Match(tenant, hashring) {
+			return false
+		}
+	}
+	return true
+}
+
+// MatcherFunc is a shim to use a func as a Matcher.
+type MatcherFunc func(string, string) bool
+
+// Match implements the Matcher interface.
+func (m MatcherFunc) Match(tenant, hashring string) bool {
+	return m(tenant, hashring)
+}
+
+// ExactMatcher is a matcher that checks if the tenant exactly matches the hashring name.
+var ExactMatcher = MatcherFunc(func(tenant, hashring string) bool { return tenant == hashring })
+
+// hash returns a hash for the given tenant and time series.
+func hash(tenant string, ts *prompb.TimeSeries) uint64 {
+	// Sort labelset to ensure a stable hash.
+	sort.Slice(ts.Labels, func(i, j int) bool { return ts.Labels[i].Name < ts.Labels[j].Name })
+
+	b := make([]byte, 0, 1024)
+	b = append(b, []byte(tenant)...)
+	b = append(b, sep)
+	for _, v := range ts.Labels {
+		b = append(b, v.Name...)
+		b = append(b, sep)
+		b = append(b, v.Value...)
+		b = append(b, sep)
+	}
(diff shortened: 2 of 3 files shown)


metalmatze:
Very nice. I like that it is so simple. Should be a good start!

/cc @bwplotka

pkg/receive/hashring.go
```diff
@@ -0,0 +1,68 @@
...
+	"github.com/cespare/xxhash"
+)
+
+const sep = '\xff'
+
+// Hashring finds the correct host to handle a given time series
+// for a specified tenant.
+// It returns the hostname and any error encountered.
+type Hashring interface {
+	GetHost(string, *prompb.TimeSeries) (string, error)
```
brancz: could we name these parameters here already, just so it's clear what the input and outputs are for this interface

pkg/receive/hashring.go
```diff
@@ -0,0 +1,68 @@
...
+}
+
+// hash returns a hash for the given tenant and time series.
+func hash(tenant string, ts *prompb.TimeSeries) uint64 {
```
brancz: nothing to do, just a side note to think about for the future: we probably want to at some point cache this, as sorting + hashing seems rather expensive, plus we can probably re-use the byte slices created for these hashes, but let's leave these optimizations for later until we actually measured that they are necessary

pkg/receive/hashring.go
```diff
@@ -0,0 +1,68 @@
...
+	// Sort labelset to ensure a stable hash.
+	sort.Slice(ts.Labels, func(i, j int) bool { return ts.Labels[i].Name < ts.Labels[j].Name })
+
+	b := make([]byte, 0, 1024)
+	b = append(b, []byte(tenant)...)
```
brancz: this needs a separator at the end, otherwise with clever label names collisions could be engineered

pkg/receive/hashring.go
```diff
@@ -0,0 +1,68 @@
...
+	}
+	if h, ok := m[""]; ok {
+		return h.GetHost(tenant, ts)
+	}
+	return "", errors.New("no matching hosts to handle tenant")
+}
+
+// NewHashring creates a multi-tenant hashring for a given
+// configuration that maps tenants to groups of hosts.
+func NewHashring(tenants map[string][]string) Hashring {
```
brancz: let's architect this in a way right now where we instead of a direct map, have a list of filters that if all passed decides for that particular hashring, and for now just implement an exact-match-filter

go.mod
```diff
@@ -5,6 +5,7 @@ require (
 	github.com/Azure/azure-storage-blob-go v0.0.0-20181022225951-5152f14ace1c
 	github.com/NYTimes/gziphandler v1.1.1
 	github.com/armon/go-metrics v0.0.0-20180917152333-f0300d1749da
+	github.com/cespare/xxhash v1.1.0
```
brancz: perfect choice :+1:

let's document that we're using this hash in the design doc/proposal

pkg/receive/hashring.go
```diff
@@ -0,0 +1,74 @@
...
+	}
+	if h, ok := m[""]; ok {
+		return h.GetHost(tenant, ts)
+	}
+	return "", errors.New("no matching hosts to handle tenant")
+}
+
+// NewHashring creates a multi-tenant hashring for a given
+// configuration that maps tenants to groups of hosts.
+func NewHashring(groups []*targetgroup.Group) Hashring {
```
brancz: This can technically change at runtime, no? I'm ok with skipping this aspect for now, but we either need to make clear that only what is discovered at startup is what is discovered over the lifetime of the process.

pkg/receive/hashring.go
```diff
@@ -0,0 +1,72 @@
...
+}
+
+// GetHost returns a hostname to handle the given tenant and time series.
+func (s *simpleHashring) GetHost(tenant string, ts *prompb.TimeSeries) (string, error) {
+	// Always return nil here to implement the Hashring interface.
+	return string(s.Targets[hash(tenant, ts)%uint64(len(s.Targets))][model.AddressLabel]), nil
+}
+
+// multiTenantHashring represents a set of Hashrings for different tenants.
+type multiTenantHashring map[string]Hashring
```
brancz: Instead of a direct map, let's use a list of objects that contain a list of filters, that if all filters are passed, the hashring is chosen. If necessary we can cache this at runtime into a data-structure like this one, as this data is perfectly cachable.

pkg/receive/hashring.go
squat: good catch here

pkg/receive/hashring.go
squat: I actually wrote another hashring implementation called `discoveryHashring`, which updates the hashring as new peers are discovered, but decided to throw it out as we need need to shut down the receiver *anyways* when new peers are discovered, so there was little benefit in having a dynamic hashring right now.

The other choice is to make it updateable on the hashring-side of things and still cleanly shutdown the receiver anyways when the hashring changes. WDYT?

brancz: Yeah, I was more saying that if we ever detect a change of these groups at runtime, we should log that that has happened.

pkg/receive/hashring.go
squat: yes, we'll probably want to make a buffer pool to keep allocations low. caching may be expensive as there can easily tens of millions of time series + tenant combinations

go.mod
brancz: this is still to be addressed

pkg/receive/hashring.go
brancz: yeah optimizing here now would be premature, once it becomes a problem we can try out the different strategies and prove with numbers what's best, nothing to do for now

go.mod
squat: I looked back, and it was actually already included it in the proposal: https://github.com/improbable-eng/thanos/commit/64e7a2607392289bec5d39b9c40fb6c261eed9ba#diff-4413dcb353f516890e28f066b0e56b90R96

PR #1218, opened 2019-06-03
Create jsonnet based kubernetes manifests

metalmatze:
This is an extensible approach to configuration management. You start
off with a querier only. Then you can add stores and PVCs as you like.

I've also added a generate YAML file containing everything to deploy:
`jsonnet/thanos/kubernetes.yaml`. This should probably be removed before we want to merge.

TODOs:
* [ ] Make the headless service for Thanos Store actually work
* [ ] Maybe add example Prometheus so that we could see data

/cc @bwplotka @brancz @gouthamve @aditya-konarde @squat

### DIFF
--- a/jsonnet/.gitignore
+++ b/jsonnet/.gitignore
@@ -1,0 +1,1 @@
+thanos/vendor/
--- a/jsonnet/thanos/jsonnetfile.json
+++ b/jsonnet/thanos/jsonnetfile.json
@@ -1,0 +1,14 @@
+{
+    "dependencies": [
+        {
+            "name": "ksonnet",
+            "source": {
+                "git": {
+                    "remote": "https://github.com/ksonnet/ksonnet-lib",
+                    "subdir": ""
+                }
+            },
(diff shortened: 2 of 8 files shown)


metalmatze:
Thinking about the fact, that @bwplotka doesn't really want this in this repository, I am going to create a kube-thanos repository. I'm happy to discuss where to put this in the future. I'd like to move it as close to Thanos itself, as possible.

metalmatze:
I've created https://github.com/metalmatze/kube-thanos for all Kubernetes specific things.

@brancz and I discussed that we still think the alerts, recording rules and dashboards based on jsonnet should go into this repository in the future, as we can move that along with the exposed metrics in the code.
We'll open an new PR / issue for that, once we're there.

povilasv:
@metalmatze we also have https://github.com/utilitywarehouse/thanos-manifests which is kustomize template, which I plan to propose to merge them here. 

Any thoughts? :)

PR #1227, opened 2019-06-05
.*: replace linters with golangci-lint

GiedriusS:
Replace all of our current linters with a meta linter which includes all
of the linters that we've used before and more.

Fixed detected problems:
* `api/v1/query{,range}` now properly adds timeout on top of the context
* Added missing error checking
* `pkg/pool` now uses pointers to slices. AIUI this means that we are now holding references to slices for a smaller amount of time and the deallocation happens faster (it works more efficiently) thus the `alloc` counter doesn't make much sense any more. I have removed it. 

These two simple programs show that indeed it is more efficient to store ptrs-to-slices in a `sync.Pool`:
https://play.golang.org/p/nWaPErUmKM9
https://play.golang.org/p/7tSpZYMhjNa

Plus: http://staticcheck.io/docs/checks#SA6002

* Other minor cleanups of dead code and now we follow best practices

Verification: `go test`, played around with Thanos Query w/ these changes - everything seems to work like expected.


Makefile
```diff
@@ -151,20 +150,6 @@ check-docs: $(EMBEDMD) $(LICHE) build
...
-# errcheck performs static analysis and returns error if any of the errors is not checked.
-.PHONY: errcheck
-errcheck: $(ERRCHECK)
-	@echo ">> errchecking the code"
-	$(ERRCHECK) -verbose -exclude .errcheck_excludes.txt ./cmd/... ./pkg/... ./test/...
-
-# format formats the code (including imports format).
-# NOTE: format requires deps to not remove imports that are used, just not resolved.
-# This is not encoded, because it is often used in IDE onSave logic.
-.PHONY: format
```
bwplotka: This is bit different... Why we are removing this? It's nice to have a makefile for formating a file (this also allows format single file)

We use it in Goland as well (:

bwplotka:
We need some fixes, but I think we also need to make sure there is exclude list for `errcheck` things. We used to have a list in https://github.com/improbable-eng/thanos/blob/b4d233e6fe904c52a22948afea3c3002d894e5df/.errcheck_excludes.txt#L1 but seems like now it is ignored by this change?

Makefile
GiedriusS: It's part of the `golangci-lint` as well I think, need to double check

GiedriusS:
> We need some fixes, but I think we also need to make sure there is exclude list for `errcheck` things. We used to have a list in
> 
> https://github.com/improbable-eng/thanos/blob/b4d233e6fe904c52a22948afea3c3002d894e5df/.errcheck_excludes.txt#L1
> but seems like now it is ignored by this change?

Yes, will do. Will bring it back. This is just a very initial version for testing how it works in general.

Makefile
GiedriusS: Brought it back but now it uses `golint-ci`

GiedriusS:
I will add `InstrumentHandler()` deprecation warnings into an ignore file for now and then I'll make an issue for it - we need to definitely migrate off of them because they have been completely removed recently. Rationale: I don't want to make this PR too big.

bwplotka:
@adrien-f is it similar to your work? @GiedriusS is it rdy for review?

GiedriusS:
Not yet, got a bit too trigger happy. Still need to fix out why the tests keep failing.

cmd/thanos/rule.go
```diff
@@ -769,7 +769,11 @@ func queryFunc(
 				// TODO(bwplotka): Propagate those to UI, probably requires changing rule manager code ):
 				level.Warn(logger).Log("warnings", strings.Join(warns, ", "), "query", q)
 			}
-			return v, err
+
+			if err == nil {
```
povilasv:
```
if err != nil {
return nil,err
}
```
please  :)

pkg/compact/compact.go
```diff
@@ -63,9 +63,6 @@ type syncerMetrics struct {
 	garbageCollectionDuration prometheus.Histogram
 	compactions               *prometheus.CounterVec
 	compactionFailures        *prometheus.CounterVec
-	indexCacheBlocks          prometheus.Counter
```
povilasv: No more metrics?

pkg/pool/pool.go
```diff
@@ -42,8 +42,9 @@ func NewBytesPool(minSize, maxSize int, factor float64, maxTotal uint64) (*Bytes
 		buckets:  make([]sync.Pool, len(sizes)),
 		sizes:    sizes,
 		maxTotal: maxTotal,
-		new: func(sz int) []byte {
-			return make([]byte, 0, sz)
+		new: func(sz int) *[]byte {
```
povilasv: Are we sure about pointer to bite slice?

slice is a pointer interally, right?

pkg/store/bucket.go
```diff
@@ -1199,12 +1195,12 @@ func (b *bucketBlock) readIndexRange(ctx context.Context, off, length int64) ([]
 	return c, nil
 }
 
-func (b *bucketBlock) readChunkRange(ctx context.Context, seq int, off, length int64) ([]byte, error) {
+func (b *bucketBlock) readChunkRange(ctx context.Context, seq int, off, length int64) (*[]byte, error) {
```
povilasv: ditto with pointer

pkg/store/proxy.go
```diff
@@ -3,13 +3,14 @@ package store
...
 	"fmt"
-	"github.com/improbable-eng/thanos/pkg/tracing"
 	"io"
 	"math"
 	"strings"
 	"sync"
 	"time"
 
+	"github.com/improbable-eng/thanos/pkg/tracing"
+
```
povilasv: remove blink line please :)

cmd/thanos/main.go
```diff
@@ -144,6 +144,10 @@ func main() {
 		var closer io.Closer
 		var confContentYaml []byte
 		confContentYaml, err = tracingConfig.Content()
+		if err != nil {
```
bwplotka: :scream:

cmd/thanos/rule.go
bwplotka: :+1:

pkg/compact/compact.go
```diff
@@ -1000,25 +997,29 @@ func (c *BucketCompactor) Compact(ctx context.Context) error {
 
 		// Clean up the compaction temporary directory at the beginning of every compaction loop.
 		if err := os.RemoveAll(c.compactDir); err != nil {
+			workCtxCancel()
```
bwplotka: Defer wouldn't be better? I think you can cancel twice just fine.

pkg/component/component.go
```diff
@@ -8,22 +8,18 @@ import (
...
 // Source is a Thanos component that produce blocks of metrics.
 type Source interface {
-	producesBlocks()
 	String() string
 }
 
 // SourceStoreAPI is a component that implements Thanos' gRPC StoreAPI
 // and produce blocks of metrics.
 type SourceStoreAPI interface {
-	implementsStoreAPI()
```
bwplotka: I think it might be needed to ensure we have type safety (so someone cannot put StoreAPI as Source.

Having lowercase interface method should be fine, is it marked as wrong?

pkg/pool/pool.go
bwplotka: Well it's a pointer to array but not to slice header (len, cap). But yea, interested to know more why that change. Looks wrong especialy that we allocate new slice so no point in having pointer.

GiedriusS: @bwplotka @povilasv Yes, this is to expected because we want to avoid storing the length and capacity of the slice so we should use pointers to slice as per this recommendation here http://staticcheck.io/docs/checks#SA6002. Do you disagree with it?

pkg/compact/compact.go
GiedriusS: They were never there. I'm not sure what this is supposed to even show. Either way, no one has missed this so I have removed it 😅

pkg/pool/pool.go
GiedriusS: Anyway, made two simple tests so that we could put this matter to rest:
https://play.golang.org/p/nWaPErUmKM9
https://play.golang.org/p/7tSpZYMhjNa

You can see that with ptrs the bytes of allocated heap objects is a bit lower.

pkg/component/component.go
GiedriusS: I saw this trick recently used in Prometheus code - good observation. Will leave it there!

.golangci.yml
```diff
@@ -0,0 +1,21 @@
+linters-settings:
+    errcheck:
+        exclude: ./.errcheck_excludes.txt
+
+issues:
+    exclude-rules:
+        # TODO: move away from deprecated prometheus.InstrumentHandler{,Func}
```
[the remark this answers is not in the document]
GiedriusS: https://github.com/improbable-eng/thanos/issues/1304 :+1: good catch

bwplotka:
I think @povilasv had quesitons around byte slice. I think if it helps allocationwise - I don't see harm. I would merge this.

povilasv:
> I think @povilasv had quesitons around byte slice. I think if it helps allocationwise - I don't see harm. I would merge this.

Not really, so the slice header is sort of:
```
type sliceHeader struct {
ptr *Elem
len int
cap int
}
```

copying 2 ints is a trivial op, so I still don't like the pointers to slices.

GiedriusS:
It sounds to me like that is just your personal preference when it brings tangible benefits especially considering how in those code paths a lot of allocations/deallocations are happening. I don't see anything wrong with that code: we just simply want to reuse the slice. Plus, IMHO it is probably not without reason that even Go developers themselves recommend this since using ptrs to slices avoids pointless allocations which is the point of `sync.Pool` to begin with.

brancz:
I agree with @GiedriusS, the code paths using `sync.Pool` are intentionally optimizing allocations, so I don't think this is over-optimizing, it's exactly what should be done in these code paths in my opinion.

povilasv:
If it's an optimization, then there should be some benchmarks, as right now we are going blind.

AFAIK non pointer stuff is easier on escape analysis and inclining as there are no references to the thing.

https://github.com/golang/go/wiki/CompilerOptimizations#escape-analysis-and-inlining

So I'm really not sure whether we get less allocations here.

GiedriusS:
There is a small benchmark - I have pasted the links in the comments and in the original ticket. Do you think it's wrong?

povilasv:
Nope as the go wiki specifically states if you do:
> For example, anything assigned to any kind of indirection (*p = ...) is considered escaped.

the variables has escaped, meaning no inlining.

> There is a small benchmark - I have pasted the links

Doesn't consider our code base properly, we would need proper benchmarks like https://dave.cheney.net/2013/06/30/how-to-write-benchmarks-in-go

which shows allocs.

PR #1229, opened 2019-06-06
[WIP] Change folders structure.

krasi-georgiev:
For quick code change/test cycles I want to be able to just `go run cmd/thanos/main.go --flags` and for this we need to keep a single file in that dir.

What do you think about moving some files around to allow this?

obviously the PR is just as a proposal and I will tidy up the whole thing and fix many lint errors if you think this change is a good idea.

### DIFF
--- a/cmd/thanos/main.go
+++ b/cmd/thanos/main.go
@@ -14,7 +14,7 @@ import (
 	gprom "github.com/armon/go-metrics/prometheus"
 	"github.com/go-kit/kit/log"
 	"github.com/go-kit/kit/log/level"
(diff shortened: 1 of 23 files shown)


bwplotka:
We could do that but if you want quick code fun/test changes that's not good enough argument to do this change. (:

Just copy this and paste or use from Linux history:
`go run ./cmd/thanos/!(*_test).go`

Also after one day you will text enough and outscale single component so starting few by hand will be painful anyway (: you will fallback to script or kubernetes.

krasi-georgiev:
ok I will give it a try and will let you know.

bwplotka:
Closing then for now (:

PR #1231, opened 2019-06-06
pkg/receive: remove deadlock on interrupt

squat:
## Changes

Currently, the receive component blocks indefinitely when the Thanos
process is interrupted because the given context is not actually used
to stop the server.

* This change adjusts the receive Handler struct to quit on context done.
* Remove unused quit chan

## Verification

* start the receive component;
* send SIGINT
* verify that before the change the process does not exit
* verify that after fix process does exit

cc @brancz @bwplotka

### DIFF
--- a/cmd/thanos/receive.go
+++ b/cmd/thanos/receive.go
@@ -233,16 +233,12 @@ func runReceive(
 
 	level.Debug(logger).Log("msg", "setting up receive http handler")
 	{
-		ctx, cancel := context.WithCancel(context.Background())
 		g.Add(
 			func() error {
-				if err := webHandler.Run(ctx); err != nil {
-					return fmt.Errorf("error starting web server: %s", err)
-				}
-				return nil
+				return errors.Wrap(webHandler.Run(), "error starting web server")
 			},
 			func(err error) {
-				cancel()
+				webHandler.Close()
 			},
 		)
 	}
--- a/pkg/receive/handler.go
+++ b/pkg/receive/handler.go
@@ -1,7 +1,6 @@
 package receive
 
 import (
-	"context"
 	"fmt"
 	"io/ioutil"
 	stdlog "log"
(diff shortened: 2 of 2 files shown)


pkg/receive/handler.go
```diff
@@ -159,14 +158,12 @@ func (h *Handler) Run(ctx context.Context) error {
 		ErrorLog: errlog,
 	}
 
-	var g run.Group
-	g.Add(func() error {
-		return httpSrv.Serve(listener)
-	}, func(error) {
+	go func() {
+		<-ctx.Done()
 		runutil.CloseWithLogOnErr(h.logger, listener, "receive HTTP listener")
```
bwplotka: Why not having `handler.Close` method and closing listener there? And having `Close()` method invoked in root `oklog/run`? Would be more explicit and consistent? (:

squat: yes absolutely sounds good, I assumed we were using context for consistency :)

bwplotka: Hard to say about any consistency here.. but you are right we mix things everywhere in this space. (: We can do both, but this sounds like extra hidden go routine (:

cmd/thanos/receive.go
```diff
@@ -233,16 +233,15 @@ func runReceive(
 
 	level.Debug(logger).Log("msg", "setting up receive http handler")
 	{
-		ctx, cancel := context.WithCancel(context.Background())
 		g.Add(
 			func() error {
-				if err := webHandler.Run(ctx); err != nil {
+				if err := webHandler.Run(); err != nil {
 					return fmt.Errorf("error starting web server: %s", err)
```
bwplotka: `errors.Wrap` instead, also I am 90% sure `Run` will never return nil error (: Not a blocker.

pkg/receive/handler.go
```diff
@@ -122,9 +119,9 @@ func (h *Handler) testReady(f http.HandlerFunc) http.HandlerFunc {
 	}
 }
 
-// Quit returns the receive-only quit channel.
-func (h *Handler) Quit() <-chan struct{} {
-	return h.quitCh
+// Close stops the Handler.
+func (h *Handler) Close() {
```
bwplotka: Much more cleaner :heart:

brancz:
Thanks! lgtm :+1:

PR #1242, opened 2019-06-10
feat: added prober package to generalize readiness probes

FUSAKLA:
This PR adds prober package/component which will unify readiness and health checks of all thanos components.

For more info please refer to the PR https://github.com/improbable-eng/thanos/pull/656

### DIFF
--- a/pkg/component/component.go
+++ b/pkg/component/component.go
@@ -6,6 +6,11 @@ import (
 	"github.com/improbable-eng/thanos/pkg/store/storepb"
 )
 
+// Component is a generic component interface
+type Component interface {
+	String() string
+}
+
 // StoreAPI is a component that implements Thanos' gRPC StoreAPI.
 type StoreAPI interface {
 	implementsStoreAPI()
--- a/pkg/component/component.go
+++ b/pkg/component/component.go
@@ -6,7 +6,7 @@ import (
 	"github.com/improbable-eng/thanos/pkg/store/storepb"
 )
 
-// Component is a generic component interface
+// Component is a generic component interface.
 type Component interface {
 	String() string
 }
--- a/pkg/prober/prober.go
+++ b/pkg/prober/prober.go
@@ -17,8 +17,6 @@ import (
 const (
 	healthyEndpointPath  = "/-/healthy"
 	readyEndpointPath    = "/-/ready"
-	okProbeFmt           = "thanos %v is %v"
-	errorProbeFmt        = "thanos %v is not %v. Reason: %v"
 	probeErrorHTTPStatus = 503
 	initialErrorFmt      = "thanos %s is initializing"
 )
@@ -46,12 +44,14 @@ func NewProber(component component.Component, logger log.Logger, reg prometheus.
 		healthiness: initialErr,
 		readiness:   initialErr,
 		readyStateMetric: prometheus.NewGauge(prometheus.GaugeOpts{
-			Name: fmt.Sprintf("thanos_%s_ready", component),
-			Help: "Represents readiness status of the thanos component.",
+			Name:        "prober_ready",
+			Help:        "Represents readiness status of the component Prober.",
+			ConstLabels: map[string]string{"component": component.String()},
 		}),
 		healthyStateMetric: prometheus.NewGauge(prometheus.GaugeOpts{
-			Name: fmt.Sprintf("thanos_%s_healthy", component),
-			Help: "Represents healthiness status of the thanos component.",
+			Name:        "prober_healthy",
+			Help:        "Represents health status of the component Prober.",
+			ConstLabels: map[string]string{"component": component.String()},
 		}),
 	}
 	if reg != nil {
@@ -72,13 +72,12 @@ func (p *Prober) RegisterInMux(mux *http.ServeMux) {
 	mux.HandleFunc(readyEndpointPath, p.probeHandlerFunc(p.IsReady, "ready"))
 }
 
-func (p *Prober) writeResponse(w http.ResponseWriter, probeFunc func() error, probeType string) {
-	err := probeFunc()
-	if err != nil {
-		http.Error(w, fmt.Sprintf(errorProbeFmt, p.component, probeType, err), probeErrorHTTPStatus)
+func (p *Prober) writeResponse(w http.ResponseWriter, probeFn func() error, probeType string) {
+	if err := probeFn(); err != nil {
+		http.Error(w, fmt.Sprintf("thanos %v is not %v. Reason: %v", p.component, probeType, err), probeErrorHTTPStatus)
 		return
 	}
-	if _, e := io.WriteString(w, fmt.Sprintf(okProbeFmt, p.component, probeType)); e == nil {
+	if _, err := io.WriteString(w, fmt.Sprintf("thanos %v is %v", p.component, probeType)); err == nil {
 		level.Error(p.logger).Log("msg", "failed to write probe response", "probe type", probeType, "err", err)
 	}
 }
--- a/pkg/prober/prober.go
+++ b/pkg/prober/prober.go
@@ -6,6 +6,8 @@ import (
 	"net/http"
 	"sync"
 
+	"github.com/prometheus/client_golang/prometheus"
+
 	"github.com/go-kit/kit/log"
 	"github.com/go-kit/kit/log/level"
 	"github.com/improbable-eng/thanos/pkg/component"
@@ -23,37 +25,39 @@ const (
 
 // Prober represents health and readiness status of given component.
 type Prober struct {
-	logger       log.Logger
-	componentMtx sync.RWMutex
-	component    component.Component
-	readyMtx     sync.RWMutex
-	readiness    error
-	healthyMtx   sync.RWMutex
-	healthiness  error
-}
-
-// SetComponent sets component name of the Prober displayed in responses.
-func (p *Prober) SetComponent(component component.Component) {
-	p.componentMtx.Lock()
-	defer p.componentMtx.Unlock()
-	p.component = component
-}
-
-func (p *Prober) getComponent() component.Component {
-	p.componentMtx.RLock()
-	defer p.componentMtx.RUnlock()
-	return p.component
+	logger             log.Logger
+	componentMtx       sync.RWMutex
+	component          component.Component
+	readyMtx           sync.RWMutex
+	readiness          error
+	healthyMtx         sync.RWMutex
+	healthiness        error
+	readyStateMetric   prometheus.Gauge
+	healthyStateMetric prometheus.Gauge
 }
 
 // NewProber returns Prober representing readiness and healthiness of given component.
-func NewProber(component component.Component, logger log.Logger) *Prober {
+func NewProber(component component.Component, logger log.Logger, reg prometheus.Registerer) *Prober {
 	initialErr := fmt.Errorf(initialErrorFmt, component)
-	return &Prober{
+
+	p := &Prober{
 		component:   component,
 		logger:      logger,
 		healthiness: initialErr,
 		readiness:   initialErr,
+		readyStateMetric: prometheus.NewGauge(prometheus.GaugeOpts{
+			Name: fmt.Sprintf("thanos_%s_ready", component),
+			Help: "Represents readiness status of the thanos component.",
+		}),
+		healthyStateMetric: prometheus.NewGauge(prometheus.GaugeOpts{
+			Name: fmt.Sprintf("thanos_%s_healthy", component),
+			Help: "Represents healthiness status of the thanos component.",
+		}),
 	}
+	if reg != nil {
+		reg.MustRegister(p.readyStateMetric, p.healthyStateMetric)
+	}
+	return p
 }
 
 // RegisterInRouter registers readiness and liveness probes to router.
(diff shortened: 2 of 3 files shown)


pkg/component/component.go
```diff
@@ -6,6 +6,11 @@ import (
 	"github.com/improbable-eng/thanos/pkg/store/storepb"
 )
 
+// Component is a generic component interface
```
bwplotka: missing trailing period

pkg/prober/prober.go
```diff
@@ -0,0 +1,158 @@
...
+	"github.com/go-kit/kit/log"
+	"github.com/go-kit/kit/log/level"
+	"github.com/improbable-eng/thanos/pkg/component"
+	"github.com/prometheus/common/route"
+)
+
+const (
+	healthyEndpointPath  = "/-/healthy"
+	readyEndpointPath    = "/-/ready"
+	okProbeText          = "thanos %v is %v"
```
bwplotka:
```suggestion
	okProbeFmt          = "thanos %v is %v"
```
To show it is templated

bwplotka: and below (:

pkg/prober/prober.go
```diff
@@ -0,0 +1,158 @@
...
+	errorProbeText       = "thanos %v is not %v. Reason: %v"
+	probeErrorHTTPStatus = 503
+	initialErrorText     = "thanos %s is initializing"
+)
+
+// Prober represents health and readriness status of given compoent.
```
bwplotka:
```suggestion
// Prober represents health and readiness status of given component.
```

pkg/prober/prober.go
```diff
@@ -0,0 +1,158 @@
...
+	componentMtx sync.Mutex
+	component    component.Component
+	readyMtx     sync.Mutex
+	readiness    error
+	healthyMtx   sync.Mutex
+	healthiness  error
+}
+
+// SetLogger sets logger used by the Prober.
+func (p *Prober) SetLogger(logger log.Logger) {
```
bwplotka: Is it needed?

pkg/prober/prober.go
```diff
@@ -0,0 +1,158 @@
...
+	p.logger = logger
+}
+
+func (p *Prober) getLogger() log.Logger {
```
bwplotka: same here, do we need this?

pkg/prober/prober.go
```diff
@@ -0,0 +1,158 @@
...
+	p.componentMtx.Lock()
+	defer p.componentMtx.Unlock()
+	return p.component
+}
+
+// NewProber returns Prober representing readiness and healthiness of given component.
+func NewProber(component component.Component, logger log.Logger) *Prober {
+	initialErr := fmt.Errorf(initialErrorText, component)
+	prober := &Prober{}
+	prober.SetComponent(component)
```
bwplotka: Prober should not use public method TBH (:

bwplotka: Can we just create prober with filled struct?

pkg/prober/prober.go
FUSAKLA: Yeah, it was needed formerly when there was mutex for logger as well but you suggested to drop it and I just didn't notice there is no need for the getter/setter now without it. 
Thanks for pointing out, will remove.

pkg/prober/prober.go
FUSAKLA: Definitely, just out of curiosity why should be this avoided? Is that generic thing in Golang not to use own public methods?

FUSAKLA: I removed also calls to `SetNotReady` and `SetNotHealthy` which caused that the mutex locking is gone but since it's the instantiation of the struct it should be ok I hope?

pkg/prober/prober.go
FUSAKLA: removed

pkg/prober/prober.go
FUSAKLA: removed, more info above

pkg/component/component.go
FUSAKLA: thanks, should be added

pkg/prober/prober.go
FUSAKLA: all should be renamed

FUSAKLA:
@bwplotka thanks fot the comments! Should be all resolved now I hope.

pkg/prober/prober.go
```diff
@@ -0,0 +1,150 @@
...
+func (p *Prober) getComponent() component.Component {
+	p.componentMtx.Lock()
+	defer p.componentMtx.Unlock()
+	return p.component
+}
+
+// NewProber returns Prober representing readiness and healthiness of given component.
+func NewProber(component component.Component, logger log.Logger) *Prober {
+	initialErr := fmt.Errorf(initialErrorFmt, component)
+	prober := &Prober{
```
bwplotka: you don't need this variable, IMO just `return &Prober`

pkg/prober/prober.go
```diff
@@ -0,0 +1,150 @@
...
+		component:   component,
+		logger:      logger,
+		healthiness: initialErr,
+		readiness:   initialErr,
+	}
+	return prober
+}
+
+// RegisterInRouter registers readiness and liveness probes to mux.
```
bwplotka:
```suggestion
// RegisterInRouter registers readiness and liveness probes to router.
```

pkg/prober/prober.go
```diff
@@ -0,0 +1,150 @@
...
+
+func (p *Prober) probeHandlerFunc(probeFunc func() error, probeType string) func(http.ResponseWriter, *http.Request) {
+	return func(w http.ResponseWriter, _ *http.Request) {
+		p.writeResponse(w, probeFunc, probeType)
+	}
+}
+
+// IsReady returns error if component is not ready and nil if it is.
+func (p *Prober) IsReady() error {
+	p.readyMtx.Lock()
```
bwplotka: Let's Move to `sync.RWMutex` and do `Rlock` and `RUnlock` here (: There will be more readers than writerrs.

pkg/prober/prober.go
```diff
@@ -0,0 +1,150 @@
...
+	}
+}
+
+func (p *Prober) probeHandlerFunc(probeFunc func() error, probeType string) func(http.ResponseWriter, *http.Request) {
+	return func(w http.ResponseWriter, _ *http.Request) {
+		p.writeResponse(w, probeFunc, probeType)
+	}
+}
+
+// IsReady returns error if component is not ready and nil if it is.
```
bwplotka:
```suggestion
// IsReady returns error if component is not ready and nil otherwise
```

pkg/prober/prober.go
```diff
@@ -0,0 +1,150 @@
...
+func (p *Prober) IsReady() error {
+	p.readyMtx.Lock()
+	defer p.readyMtx.Unlock()
+	return p.readiness
+}
+
+// SetReady sets components status to ready.
+func (p *Prober) SetReady() {
+	if p.IsReady() != nil {
```
bwplotka: This is wrong I think as those two methods are locked so you can have a race. IF you need this log line, please add this to `SetNotReady`

pkg/prober/prober.go
```diff
@@ -0,0 +1,150 @@
...
+		level.Info(p.logger).Log("msg", "changing probe status", "status", "ready")
+	}
+	p.SetNotReady(nil)
+}
+
+// SetNotReady sets components status to not ready with given error as a cause.
+func (p *Prober) SetNotReady(err error) {
+	if err != nil && p.IsReady() == nil {
```
bwplotka: This is wrong from the same reason above, there can be a race.

Please lock here and then check `p.readiness` instead to check what to log. What do you think? (:

pkg/prober/prober.go
```diff
@@ -0,0 +1,150 @@
...
+// IsHealthy returns error if component is not healthy and nil if it is.
+func (p *Prober) IsHealthy() error {
+	p.healthyMtx.Lock()
+	defer p.healthyMtx.Unlock()
+	return p.healthiness
+}
+
+// SetHealthy sets components status to healthy.
+func (p *Prober) SetHealthy() {
+	if p.IsHealthy() != nil {
```
bwplotka: Exactly the same issue as for Readiness.

Let's move to `RWLock` and log under lock as well (it's only on write so should be fine)

pkg/prober/prober.go
```diff
@@ -0,0 +1,150 @@
...
+	}
+	p.healthyMtx.Lock()
+	defer p.healthyMtx.Unlock()
+	p.healthiness = err
+}
+
+// HandleIfReady if probe is ready calls the function otherwise returns 503.
+func (p *Prober) HandleIfReady(f http.HandlerFunc) http.HandlerFunc {
+	return func(w http.ResponseWriter, r *http.Request) {
+		p.readyMtx.Lock()
```
bwplotka: Why not reusing `p.IsReady`? (: In this case it's fine.

bwplotka: Especially that we don't know how long `f` is and what it does from this function perspecitve.

pkg/prober/prober.go
bwplotka: Because there is no need for this. Literally we will never in our code require dynamic reload of component. We don't have components that transforms into another components right? 

Every bit of genercity is introducing complexity. Since we don't need that now (YAGNI rule), we can avoid that complexity, right? (:

pkg/prober/prober.go
FUSAKLA: Great, thanks I didn't know there is also RWMutex, TIL :)

pkg/prober/prober.go
FUSAKLA: Ok, just so I understand where the race is. The `IsReady` locks but once finished it again unlocks it. And only after it returns and it does not return `nil` the `SetNotReady` is called which again locks and unlocks. Which part could go wrong here?

FUSAKLA: Sorry, now I get it.  Meanwhile I unlock it the status of prober could change.

pkg/prober/prober.go
FUSAKLA: Good catch. This could cause serious issues.

FUSAKLA:
@bwplotka thank you for all the valuable comments/suggestions and for answering all my stupid questions so I can learn from it, I really appreciate it. 

Is it ok with you this way?

pkg/prober/prober.go
FUSAKLA: Makes sense, thanks for for the explanation :)

bwplotka:
No worries, just tell if you are rdy for next review iteration (: I think it's rdy now?

FUSAKLA:
Yes, it should be ready I hope.

FUSAKLA:
@bwplotka if you find a while and could take a look at this? 

It's rebased and I added metrics for observing the Prober state if that is ok with uou
- `thanos_<component>_ready`
- `thanos_<component>_healthy` (wasn't sure about this one since it should be 1 whenever e serve metrics but on the other hand it felt weird not to expose it either)

pkg/prober/prober.go
```diff
@@ -0,0 +1,160 @@
...
+	"github.com/go-kit/kit/log"
+	"github.com/go-kit/kit/log/level"
+	"github.com/improbable-eng/thanos/pkg/component"
+	"github.com/prometheus/common/route"
+)
+
+const (
+	healthyEndpointPath  = "/-/healthy"
+	readyEndpointPath    = "/-/ready"
+	okProbeFmt           = "thanos %v is %v"
```
bwplotka: Can we inline all constants that are used only once? (: There is no point of those if they are not references more than once.

pkg/prober/prober.go
```diff
@@ -0,0 +1,160 @@
...
+}
+
+// RegisterInMux registers readiness and liveness probes to mux.
+func (p *Prober) RegisterInMux(mux *http.ServeMux) {
+	mux.HandleFunc(healthyEndpointPath, p.probeHandlerFunc(p.IsHealthy, "healthy"))
+	mux.HandleFunc(readyEndpointPath, p.probeHandlerFunc(p.IsReady, "ready"))
+}
+
+func (p *Prober) writeResponse(w http.ResponseWriter, probeFunc func() error, probeType string) {
+	err := probeFunc()
```
bwplotka: you can do `if err := probeFn(); err != nil ` here (: Also we usually do `Fn` suffix for function vars so  `probeFn`  would be nice? (: What do you think?

pkg/prober/prober.go
```diff
@@ -0,0 +1,160 @@
...
+	if err != nil {
+		http.Error(w, fmt.Sprintf(errorProbeFmt, p.component, probeType, err), probeErrorHTTPStatus)
+		return
+	}
+	if _, e := io.WriteString(w, fmt.Sprintf(okProbeFmt, p.component, probeType)); e == nil {
```
bwplotka: please let's keep `err` as error variable name! (:

bwplotka: Becuse of introducing the new name there is a bug here -> you log `"err", err` not `"err", e`  as it should be (:

pkg/prober/prober.go
```diff
@@ -0,0 +1,160 @@
...
+func NewProber(component component.Component, logger log.Logger, reg prometheus.Registerer) *Prober {
+	initialErr := fmt.Errorf(initialErrorFmt, component)
+
+	p := &Prober{
+		component:   component,
+		logger:      logger,
+		healthiness: initialErr,
+		readiness:   initialErr,
+		readyStateMetric: prometheus.NewGauge(prometheus.GaugeOpts{
+			Name: fmt.Sprintf("thanos_%s_ready", component),
```
bwplotka: TBH instead of that I would just do `prober_ready",  and then:
* have `label` component. This will allow for better, potential aggregations in future.
* you can have just `prober_ready` and do not specify `thanos` prefix. This allows to reuse this package outside of Thanos even (why not)!

Thanos can then on caller side do `prometheus.WrapRegistererWithPrefix("thanos"` (: 

What do you think?

same for `prober_healthy`

bwplotka:
BTW nice idea with metric :+1:

pkg/prober/prober.go
FUSAKLA: yes, inlined them probably will be easier to read :+1:

pkg/prober/prober.go
FUSAKLA: thanks for noticing, changed to `err`

pkg/prober/prober.go
FUSAKLA: I was thinking about the naming a lot bud than looked to the rest of packages and found everywhere hardcoded the thanos prefix so I wanted to stay consistent but I agree your suggestion is more variable.

Applied as suggested

FUSAKLA:
@bwplotka thanks for those comments! All should be resolved as you suggested.

pkg/prober/prober.go
bwplotka: Yes, we did not think about that in initial implementation, but now I would aim for no `thanos_` prefix (:

PR #1243, opened 2019-06-11
 pkg/receive: enable forwarding metrics

squat:
This PR enables metrics forwarding from one receive node to another.
The receive nodes construct hashrings from the given sd-files and
use these hashrings to select a node to which to forward a given time
series. Time series are batched together to ensure that for any incoming
write-request to a node, at most one outgoing write-request will be made
every other node in the hashring.

## Changes

* enable time series forwarding between receive nodes

## Verification

Added e2e tests for receive nodes in hashring


cmd/thanos/receive.go
```diff
@@ -97,13 +134,45 @@ func runReceive(
 		MaxBlockDuration:  model.Duration(time.Hour * 2),
 	}
 
+	var hashring receive.Hashring
+	// Run File Service Discovery and check for updates.
+	if fileSD != nil {
+		ctxRun, cancelRun := context.WithCancel(context.Background())
+		fileSDUpdates := make(chan []*targetgroup.Group)
+		go fileSD.Run(ctxRun, fileSDUpdates)
```
metalmatze: Not sure if this is a common thing in the discovery package, but what about checking errors here? Seems like this, if at all, would be an upstream improvement.

pkg/receive/handler.go
```diff
@@ -180,8 +191,82 @@ func (h *Handler) receive(w http.ResponseWriter, req *http.Request) {
...
+			return
+		}
+	}
+}
+
+// forward accepts a write request, batches its time series by
+// corresponding endpoint, and forwards them in parallel. It returns a write
+// request containing only the time series that correspond to
+// local handler.
+func (h *Handler) forward(ctx context.Context, tenant string, wreq *prompb.WriteRequest) (*prompb.WriteRequest, error) {
```
metalmatze: A few more comments inside this code would be nice for guidance.

pkg/receive/handler.go
```diff
@@ -180,8 +191,82 @@ func (h *Handler) receive(w http.ResponseWriter, req *http.Request) {
...
+		if _, ok := wreqs[endpoint]; !ok {
+			wreqs[endpoint] = &prompb.WriteRequest{}
+		}
+		wr := wreqs[endpoint]
+		wr.Timeseries = append(wr.Timeseries, wreq.Timeseries[i])
+	}
+
+	ec := make(chan error)
+	defer close(ec)
+	var n int
```
metalmatze: I am pretty unsure, what `n` in this case does. Would be nice to either document or giving it a more descriptive name.

pkg/receive/handler.go
```diff
@@ -180,8 +191,82 @@ func (h *Handler) receive(w http.ResponseWriter, req *http.Request) {
...
+				level.Error(h.logger).Log("msg", "forward request error", "err", err, "endpoint", endpoint)
+				ec <- err
+				return
+			}
+			ec <- nil
+		}(endpoint)
+	}
+
+	var errs error
+	for ; n > 0; n-- {
```
metalmatze: Couldn't we utilize a sync.WaitGroup here? I might be missing something more obvious though.
https://gobyexample.com/waitgroups

pkg/receive/hashring.go
```diff
@@ -108,7 +119,53 @@ func NewHashring(matcher Matcher, groups []*targetgroup.Group) Hashring {
...
+		}
+		h := m.hashrings[t].(*simpleHashring)
+		h.Targets = append(h.Targets, g.Targets...)
 	}
 	return m
 }
+
+// HashringsAreEquivalent indicates whether the hashrings created from
+// two sets of groups would be equivalent.
+func HashringsAreEquivalent(a, b []*targetgroup.Group) bool {
```
metalmatze: In the comment for this function. Could you give a quick example on why we actually need this? Like an (example) use case.

pkg/receive/handler.go
squat: We could and it originally did but it makes it more complex in reality. Plus it introduces an unnecessary second synchronization mechanism (the first being the error chan).

pkg/receive/handler.go
squat: `n` like `i` is pretty standard for a count. I can change it to `i` instead if that’s clearer

pkg/receive/handler.go
squat: It complicated things because in order to do `wg.Wait()` we need to put the error channel receiver in a new go routine (otherwise the waitgroups wouldn’t finish (unless the chan was buffered by n)) and then we need to close the chan in a different go routine in a way that doesn’t race with the data that is shared, i.e. the collected errs. Keeping the collection in the main goroutine simplifies the code IMO and spawns fewer gorouines.

pkg/receive/hashring.go
squat: It is used here: https://github.com/squat/thanos/blob/df3289e9e914846d08bd28987a368aac6083f137/cmd/thanos/receive.go#L152

This is required so we know if the hash ring configuration has changed. The way file sd works, every interval seconds, the file is re-read and an update is sent even if the file hasn’t changed. This should not cause the receiver to shut down because the configuration is the same. For that reason we need to check if two file sd updates are equivalent. If they are, we can ignore the update. If they are different, we shutdown gracefully.

cmd/thanos/receive.go
squat: What errors are you referring to? Run returns no errors https://github.com/prometheus/prometheus/blob/master/discovery/file/file.go#L221

pkg/receive/hashring.go
squat: I can summarize that in the comment

cmd/thanos/receive.go
```diff
@@ -97,13 +134,45 @@ func runReceive(
...
+		groups := <-fileSDUpdates
+		hashring = receive.NewHashring(receive.ExactMatcher, groups)
+		g.Add(func() error {
+			for {
+				select {
+				// If the SD changes, we need to shutdown.
+				case update := <-fileSDUpdates:
+					if !receive.HashringsAreEquivalent(groups, update) {
+						level.Info(logger).Log("msg", "file service discovery changed")
+						return nil
```
brancz: Is shutting down really the right thing to do here? I feel like not accepting writes would be more appropriate, as kubernetes or equivalent would just restart the process, and really what we expect is that something from the outside initiates a shutdown causing flushing + upload with a subsequent rollout.

pkg/receive/hashring.go
```diff
@@ -108,7 +119,53 @@ func NewHashring(matcher Matcher, groups []*targetgroup.Group) Hashring {
 		matcher:   matcher,
```
brancz: I thought about this again, and I feel like this should be pushed further down into the hashring itself, which will allow it to be more extensible in the future.

cmd/thanos/receive.go
metalmatze: I thought about the upstream function not returning some errors, but this seems good.

brancz: should this be a separate run group though?

pkg/receive/handler.go
metalmatze: Yes, I got what it is used for, just some more guidance on why we need to have it next to the loop below, would be nice. (Similar to what you wrote to me on Slack, we don't want to use a WaitGroup because... This improves error handling compared to WaitGroups...) Something like that.

pkg/receive/handler.go
```diff
@@ -17,12 +19,17 @@ import (
...
 	opentracing "github.com/opentracing/opentracing-go"
+	"github.com/pkg/errors"
 	"github.com/prometheus/client_golang/prometheus"
 	"github.com/prometheus/client_golang/prometheus/promhttp"
 	"github.com/prometheus/common/route"
 	promtsdb "github.com/prometheus/prometheus/storage/tsdb"
 )
 
+const (
+	tenantHeader = "THANOS_TENANT"
```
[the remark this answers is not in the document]
squat: Ok flag sounds good. But using the `X-` prefix has been deprecated by some rfcs

pkg/receive/hashring.go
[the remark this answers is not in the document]
squat: Hm what do you mean? If I understand you, we are doing exactly that here: we save the target group that produces a hashring, then compare it deeply to a new target group to see if they would produce equivalent rings. Note that doing a shallow comparison won’t work because every `interval` seconds the file is re-read, a new target group is generated which would never be the same as the original.

pkg/receive/handler.go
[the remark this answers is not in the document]
brancz: I haven't seen that, but yes configurable in any case.

pkg/receive/hashring.go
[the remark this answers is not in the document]
brancz: I'm saying, maybe comparing targetGroups themselves would be simpler?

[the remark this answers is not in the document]
squat: Ah sorry, despite what the name says, this function accepts two `[]*targetgroup.Group` as its parameters

cmd/thanos/receive.go
squat: Hmm putting it in a separate run group means it is async and we can’t wait to receive a target group on the chan until after `g.Run()`. This means we can’t construct a hashring directly with a target group but instead the hashring needs to accept a `Discoverer`. That’s doable, just means some rearchitecting.

pkg/receive/handler.go
squat: done :+1:

pkg/receive/handler.go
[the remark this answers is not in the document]
squat: perf, done :+1:

pkg/receive/hashring.go
```diff
@@ -13,11 +13,14 @@ import (
 
 const sep = '\xff'
 
-// Hashring finds the correct host to handle a given time series
+// tenantLabel is used to group discovered targets by tenant.
+const tenantLabel = "tenant"
```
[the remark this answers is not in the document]
squat: done :+1:

cmd/thanos/receive.go
squat: updated, now the handler will simply stop accepting write-requests when the hashring changes.

squat:
cc @brancz @metalmatze @bwplotka I addressed the comments you made the first time around so this is now ready for another round :muscle:

brancz:
We had some offline discussion to change this implementation a little bit. The revised version should actually be significantly simpler.

squat:
@brancz all the changes we talked about today are in now. The configuration management is much simpler for hashrings now :relieved:. We have some boilerplate for creating a configuration file watcher, but it is sane and self-contained.

pkg/receive/config.go
```diff
@@ -0,0 +1,167 @@
...
+			cw.refresh(ctx, ch)
+
+		case <-ticker.C:
+			// Setting a new watch after an update might fail. Make sure we don't lose
+			// those files forever.
+			cw.refresh(ctx, ch)
+
+		case err := <-cw.watcher.Errors:
+			if err != nil {
+				level.Error(cw.logger).Log("msg", "error watching file", "err", err)
```
brancz: let's have a metric for this

pkg/receive/config.go
```diff
@@ -0,0 +1,167 @@
...
+			}
+		}
+	}
+}
+
+// refresh reads the configured file and sends the hashring configuration on the channel.
+func (cw *ConfigWatcher) refresh(ctx context.Context, ch chan<- []HashringConfig) {
```
brancz: let's have a metric for this

pkg/receive/hashring.go
```diff
@@ -66,49 +66,106 @@ func hash(tenant string, ts *prompb.TimeSeries) uint64 {
...
 	cache     map[string]Hashring
-	hashrings map[string]Hashring
+	hashrings []Hashring
 	matcher   Matcher
+	cfg       []HashringConfig
+
+	// We need a mutex to guard concurrent access
+	// to the cache map, as this is both written to
+	// and read from.
+	mu sync.Mutex
```
brancz: we very rarely write to the cache and often read from it, this should be a RWMutex

pkg/receive/hashring.go
```diff
@@ -66,49 +66,106 @@ func hash(tenant string, ts *prompb.TimeSeries) uint64 {
...
 }
 
-// GetHost returns a hostname to handle the given tenant and time series.
-func (m matchingHashring) GetHost(tenant string, ts *prompb.TimeSeries) (string, error) {
-	if h, ok := m.cache[tenant]; ok {
-		return h.GetHost(tenant, ts)
+// Get returns a target to handle the given tenant and time series.
+func (mh *matchingHashring) Get(tenant string, ts *prompb.TimeSeries) (string, error) {
+	mh.mu.Lock()
```
brancz: once a RWMutex, this locking should be more granular

pkg/receive/hashring.go
```diff
@@ -66,49 +66,106 @@ func hash(tenant string, ts *prompb.TimeSeries) uint64 {
...
 }
+
+// HashringFromConfig creates multi-tenant hashrings from a
+// hashring configuration file watcher.
+// The configuration file is watched for updates.
+// Hashrings are returned on the updates channel.
+// Which nodes are assigned to a hashring is controlled
+// by the given tenant label.
+// Which tenant's hashring to use is determined by the Matcher.
+func HashringFromConfig(ctx context.Context, updates chan<- Hashring, matcher Matcher, cw *ConfigWatcher) {
```
brancz: The `matcher` should just be part of the configuration now, no?

pkg/receive/hashring.go
squat: yes absolutely :+1:

pkg/receive/hashring.go
squat: what do you mean exactly? we talked about the configuration only allowing for exact set matching, otherwise for empty set it is match all. This means we *could* remove the matcher from the func signature and always use exact match.

brancz: Precisely, let's just always use set-matching

pkg/receive/config.go
```diff
@@ -0,0 +1,206 @@
...
+			prometheus.CounterOpts{
+				Name: "thanos_receive_hashrings_file_refreshes_total",
+				Help: "The number of refreshes of the hashrings configuration file.",
+			}),
+	}, nil
+}
+
+// RegisterMetrics registers the configuration watcher's metrics
+// with the given Prometheus registerer.
+func (cw *ConfigWatcher) RegisterMetrics(r prometheus.Registerer) {
```
brancz: this is typically done in the `XYZNew` function and if `prometheus.Registerer` is nil, then it's just not, see here for example: https://github.com/improbable-eng/thanos/blob/e1d3b6297f194879d9f0653abb0479e30ffc2163/pkg/shipper/shipper.go#L110

pkg/receive/hashring.go
squat: Alright sounds good :) done

pkg/receive/hashring.go
```diff
@@ -66,49 +66,106 @@ func hash(tenant string, ts *prompb.TimeSeries) uint64 {
...
+	// every tenant in the configuration.
+	for i, h := range m.cfg {
+		// If the hashring has no tenants, then it is
+		// considered a default hashring and matches everything.
+		if len(h.Tenants) == 0 {
+			found = true
+		}
+		// Match against every tenant in the hashring.
+		for _, t := range h.Tenants {
+			if m.matcher.Match(tenant, t) {
```
brancz: Really sorry for going back and forth with this, but as we decided that for now there will be only exact matches, let’s turn this marching into a set-inclusion check against the h.Tenants (The set we can build within the new function of the hashring).

squat: Sounds good to me. I’ll away with the rest of the Matcher code and interface rather than leave it unused. Yeah?

brancz: Yes, perfect :+1:

PR #1244, opened 2019-06-11
Add downsampling metrics

mattrco:
When `compact` goes into its downsampling step, no app metrics are reported, which makes our dashboard look a bit bare for several hours (sometimes days if we have a backlog). Adding a couple of metrics means we can alert if no compactions or downsamples have happened in x period.

This is a little less clean than I'd like due to the two callers of `downsampleBucket` having slightly different contexts, however I think it's a reasonable compromise. Let me know what you think!

The metrics are labelled in the same way as compact metrics for ease of comparison.

## Changes

Add `thanos_compact_downsample_total` and `thanos_compact_downsample_failures_total` downsampling metrics.

## Verification

Tests are passing, will run with real workloads.

### DIFF
--- a/cmd/thanos/downsample.go
+++ b/cmd/thanos/downsample.go
@@ -13,6 +13,7 @@ import (
 	"github.com/go-kit/kit/log/level"
 	"github.com/improbable-eng/thanos/pkg/block"
 	"github.com/improbable-eng/thanos/pkg/block/metadata"
+	"github.com/improbable-eng/thanos/pkg/compact"
 	"github.com/improbable-eng/thanos/pkg/compact/downsample"
 	"github.com/improbable-eng/thanos/pkg/component"
 	"github.com/improbable-eng/thanos/pkg/objstore"
@@ -220,9 +221,12 @@ func downsampleBucket(
 			if m.MaxTime-m.MinTime < 40*60*60*1000 {
 				continue
 			}
+
 			if err := processDownsampling(ctx, logger, bkt, m, dir, 5*60*1000); err != nil {
+				metrics.downsampleFailures.WithLabelValues(compact.GroupKey(*m))
 				return errors.Wrap(err, "downsampling to 5 min")
 			}
+			metrics.downsamples.WithLabelValues(compact.GroupKey(*m))
 
 		case 5 * 60 * 1000:
 			missing := false
@@ -242,8 +246,10 @@ func downsampleBucket(
 				continue
 			}
 			if err := processDownsampling(ctx, logger, bkt, m, dir, 60*60*1000); err != nil {
+				metrics.downsampleFailures.WithLabelValues(compact.GroupKey(*m))
 				return errors.Wrap(err, "downsampling to 60 min")
 			}
+			metrics.downsamples.WithLabelValues(compact.GroupKey(*m))
 		}
 	}
 	return nil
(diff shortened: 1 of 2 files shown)


cmd/thanos/downsample.go
```diff
@@ -31,20 +32,50 @@ import (
...
+	m.downsamples = prometheus.NewCounterVec(prometheus.CounterOpts{
+		Name: "thanos_compact_downsample_total",
+		Help: "Total number of downsampling attempts.",
+	}, []string{"group"})
+	m.downsampleFailures = prometheus.NewCounterVec(prometheus.CounterOpts{
+		Name: "thanos_compact_downsample_failures_total",
+		Help: "Total number of failed downsampling attempts.",
+	}, []string{"group"})
+
+	if err := reg.Register(m.downsamples); err != nil {
```
mattrco: This could use `MustRegister` (as done for other metrics) and skip the error check in the caller.

bwplotka: Yes, `MustRegister` is ok here

cmd/thanos/downsample.go
```diff
@@ -31,20 +32,50 @@ import (
...
+		return nil, err
+	}
+	if err := reg.Register(m.downsampleFailures); err != nil {
```
bwplotka: Yup. `MustRegister` is fine here

mattrco:
Thanks for reviewing, have updated to use `MustRegister` 👍

cmd/thanos/downsample.go
```diff
@@ -31,20 +32,46 @@ import (
...
 
 	objStoreConfig := regCommonObjStoreFlags(cmd, "", true)
 
 	m[name] = func(g *run.Group, logger log.Logger, reg *prometheus.Registry, tracer opentracing.Tracer, _ bool) error {
-		return runDownsample(g, logger, reg, *dataDir, objStoreConfig)
+		return runDownsample(g, logger, reg, *httpAddr, *dataDir, objStoreConfig)
 	}
 }
 
+type DownsampleMetrics struct {
```
bwplotka: I think it can be private, right?

cmd/thanos/downsample.go
```diff
@@ -65,6 +92,8 @@ func runDownsample(
 		}
 	}()
 
+	metrics := newDownsampleMetrics(reg)
```
GiedriusS: Could we use `extprom.WrapRegistererWithPrefix("thanos_compact_downsample", reg)` here?

cmd/thanos/downsample.go
```diff
@@ -31,20 +32,46 @@ import (
...
 
+type DownsampleMetrics struct {
+	downsamples               *prometheus.CounterVec
+	downsampleFailures        *prometheus.CounterVec
+}
+
+func newDownsampleMetrics(reg *prometheus.Registry) *DownsampleMetrics {
+	m := new(DownsampleMetrics)
+
+	m.downsamples = prometheus.NewCounterVec(prometheus.CounterOpts{
```
GiedriusS: Then we would only have `total` here

cmd/thanos/downsample.go
bwplotka: I think it makes sense but with `thanos_compact` prefix, but not a blocker.

bwplotka:
Merging in this form, we can improve that later once we actually need to reuse this component.

bwplotka:
Thanks @mattrco !

PR #1248, opened 2019-06-12
Add web UI to display the state of Thanos bucket

jaseemabid:
Exposes a web interface for the state of Thanos buckets just like `pprof web`.
This is an early preview, the code is rough on the edges and could use some
review. See the screenshots for some examples.

**What works?**

- Can display the timeline of blocks per shard
- Can print error messages if any
- Reloads data from remote storage every 30m
- Reloads browser slightly smarter than that

**TODOS**

- Some download code is duplicated from `ls -o json`; refactor as needed
- Some web code is duplicated from query UI; refactor as needed
- Do we need any of the route prefix?
- Unfortunately learned a bit late that the JS libraries cannot be vendored. That sucks :(
- The tooltip is pretty basic and can contain a lot more useful info.
- Allow some kind of filtering and other bells and whistles to the UI, but this can be part of later PR. 

**Notes for the reviewer**

- goimports is slightly off on master, the changes in kingpin is not intentional
- Multiple `client.NewBucket` will panic; the downloader avoids that.
- `downloader` can be rewritten if required
- Asset reloading doesn't work with `DEBUG=1`, I had to add an extra `-debug`

**Default view:** 
Screenshot 2019-06-12 at 18 07 52

**Loading indication:**
Screenshot 2019-06-12 at 18 08 05

**Other errors:**
Screenshot 2019-06-12 at 18 19 08

### DIFF
--- a/cmd/thanos/bucket.go
+++ b/cmd/thanos/bucket.go
@@ -359,7 +359,7 @@ func refresh(ctx context.Context, logger log.Logger, bucketUI *ui.Bucket, durati
 
 	defer runutil.CloseWithLogOnErr(logger, bkt, "bucket client")
 
-	return runutil.Repeat(duration, ctx.Done(), func() error {
+	runutil.Forever(logger, duration, ctx.Done(), func() error {
 		iterCtx, iterCancel := context.WithTimeout(ctx, 5*time.Minute)
 		defer iterCancel()
 
@@ -379,6 +379,8 @@ func refresh(ctx context.Context, logger log.Logger, bucketUI *ui.Bucket, durati
 			return nil
 		})
 	})
+
+	return nil
 }
 
 func download(ctx context.Context, logger log.Logger, bkt objstore.Bucket) (blocks []metadata.Meta, err error) {
--- a/cmd/thanos/bucket.go
+++ b/cmd/thanos/bucket.go
@@ -12,9 +12,6 @@ import (
 	"text/template"
 	"time"
 
-	opentracing "github.com/opentracing/opentracing-go"
-	"github.com/prometheus/common/route"
-
 	"github.com/improbable-eng/thanos/pkg/block"
 	"github.com/improbable-eng/thanos/pkg/block/metadata"
 	"github.com/improbable-eng/thanos/pkg/objstore"
@@ -28,12 +25,14 @@ import (
 	"github.com/oklog/run"
 	"github.com/oklog/ulid"
 	"github.com/olekukonko/tablewriter"
+	opentracing "github.com/opentracing/opentracing-go"
 	"github.com/pkg/errors"
 	"github.com/prometheus/client_golang/prometheus"
+	"github.com/prometheus/common/route"
 	"github.com/prometheus/tsdb/labels"
 	"golang.org/x/text/language"
 	"golang.org/x/text/message"
-	"gopkg.in/alecthomas/kingpin.v2"
+	kingpin "gopkg.in/alecthomas/kingpin.v2"
 )
 
 var (
@@ -299,12 +298,7 @@ func registerBucketInspect(m map[string]setupFunc, root *kingpin.CmdClause, name
 	}
 }
 
-// registerBucketWeb exposes a web interface for the state of Thanos remote store just like `pprof web`
-//
-// TODO(jaseemabid):
-//
-// 1. Some download code is duplicated from `ls -o json`; refactor as needed
-// 2. Some web code is duplicated from query UI, refactor as needed
+// registerBucketWeb exposes a web interface for the state of remote store like `pprof web`
 func registerBucketWeb(m map[string]setupFunc, root *kingpin.CmdClause, name string, objStoreConfig *pathOrContent) {
 	cmd := root.Command("web", "Web interface for remote storage bucket")
 	bind := cmd.Flag("listen", "HTTP host:port to listen on").Default("0.0.0.0:8080").String()
@@ -316,7 +310,6 @@ func registerBucketWeb(m map[string]setupFunc, root *kingpin.CmdClause, name str
 		ctx, cancel := context.WithCancel(context.Background())
 
 		router := route.New()
-
 		bucketUI := ui.NewBucketUI(logger, *label)
 		bucketUI.Register(router)
 
@@ -356,7 +349,6 @@ func registerBucketWeb(m map[string]setupFunc, root *kingpin.CmdClause, name str
 
 // refresh metadata from remote storage periodically and update UI.
 func refresh(ctx context.Context, logger log.Logger, bucketUI *ui.Bucket, duration time.Duration, timeout time.Duration, name string, reg *prometheus.Registry, objStoreConfig *pathOrContent) error {
-
 	confContentYaml, err := objStoreConfig.Content()
 	if err != nil {
 		return err
@@ -368,9 +360,7 @@ func refresh(ctx context.Context, logger log.Logger, bucketUI *ui.Bucket, durati
 	}
 
 	defer runutil.CloseWithLogOnErr(logger, bkt, "bucket client")
-
 	return runutil.Repeat(duration, ctx.Done(), func() error {
-
 		return runutil.RetryWithLog(logger, time.Minute, ctx.Done(), func() error {
 			iterCtx, iterCancel := context.WithTimeout(ctx, timeout)
 			defer iterCancel()
--- a/cmd/thanos/bucket.go
+++ b/cmd/thanos/bucket.go
@@ -309,13 +309,14 @@ func registerBucketWeb(m map[string]setupFunc, root *kingpin.CmdClause, name str
 	cmd := root.Command("web", "Web interface for remote storage bucket")
 	bind := cmd.Flag("listen", "HTTP host:port to listen on").Default("0.0.0.0:8080").String()
 	interval := cmd.Flag("refresh", "Refresh interval").Default("30m").Duration()
+	label := cmd.Flag("label", "Prometheus label to use as timeline title").String()
 
 	m[name+" web"] = func(g *run.Group, logger log.Logger, reg *prometheus.Registry, _ opentracing.Tracer, _ bool) error {
 		ctx, cancel := context.WithCancel(context.Background())
 
 		router := route.New()
 
-		bucketUI := ui.NewBucketUI(logger)
+		bucketUI := ui.NewBucketUI(logger, *label)
 		bucketUI.Register(router)
 
 		if *interval < 5*time.Minute {
--- a/cmd/thanos/bucket.go
+++ b/cmd/thanos/bucket.go
@@ -318,6 +318,10 @@ func registerBucketWeb(m map[string]setupFunc, root *kingpin.CmdClause, name str
 		bucketUI := ui.NewBucketUI(logger)
 		bucketUI.Register(router)
 
+		if *interval < 5*time.Minute {
+			level.Warn(logger).Log("msg", "Refreshing more often than 5m could lead to large data transfers")
+		}
+
 		g.Add(func() error {
 			return refresh(ctx, logger, bucketUI, *interval, name, reg, objStoreConfig)
 		}, func(error) {
@@ -358,11 +362,6 @@ func refresh(ctx context.Context, logger log.Logger, bucketUI *ui.Bucket, durati
 	return runutil.Repeat(duration, ctx.Done(), func() error {
 		iterCtx, iterCancel := context.WithTimeout(ctx, 5*time.Minute)
 		defer iterCancel()
-
-		// If there are no errors and the data isn't stale, wait a while before updating again.
-		if bucketUI.Err == nil && time.Since(bucketUI.RefreshedAt) < (5*time.Minute) {
-			return nil
-		}
 
 		return runutil.RetryWithLog(logger, time.Minute, iterCtx.Done(), func() error {
 			blocks, err := download(iterCtx, logger, bkt)
--- a/cmd/thanos/bucket.go
+++ b/cmd/thanos/bucket.go
@@ -299,13 +299,12 @@ func registerBucketInspect(m map[string]setupFunc, root *kingpin.CmdClause, name
 	}
 }
 
-// RegisterBucketWeb exposes a web interface for the state of Thanos remote store just like `pprof web`
+// registerBucketWeb exposes a web interface for the state of Thanos remote store just like `pprof web`
 //
-// TODOS:
+// TODO(jaseemabid):
 //
 // 1. Some download code is duplicated from `ls -o json`; refactor as needed
 // 2. Some web code is duplicated from query UI, refactor as needed
-// 3. Do we need any of the route prefix?
 func registerBucketWeb(m map[string]setupFunc, root *kingpin.CmdClause, name string, objStoreConfig *pathOrContent) {
 	cmd := root.Command("web", "Web interface for remote storage bucket")
 	bind := cmd.Flag("listen", "HTTP host:port to listen on").Default("0.0.0.0:8080").String()
@@ -341,19 +340,17 @@ func registerBucketWeb(m map[string]setupFunc, root *kingpin.CmdClause, name str
 	}
 }
 
-// Refresh metadata from remote storage periodically and update UI
-func refresh(ctx context.Context, logger log.Logger, bucketUI *ui.Bucket, duration time.Duration, name string,
-	reg *prometheus.Registry, objStoreConfig *pathOrContent) error {
+// refresh metadata from remote storage periodically and update UI.
+func refresh(ctx context.Context, logger log.Logger, bucketUI *ui.Bucket, duration time.Duration, name string, reg *prometheus.Registry, objStoreConfig *pathOrContent) error {
 
 	confContentYaml, err := objStoreConfig.Content()
 	if err != nil {
 		return err
 	}
 
-	// A client is created for the entire duration of the program to avoid duplicate metrics registration.
 	bkt, err := client.NewBucket(logger, confContentYaml, reg, name)
 	if err != nil {
-		return err
+		return errors.Wrap(err, "bucket client")
 	}
 
 	defer runutil.CloseWithLogOnErr(logger, bkt, "bucket client")
@@ -367,8 +364,6 @@ func refresh(ctx context.Context, logger log.Logger, bucketUI *ui.Bucket, durati
 			return nil
 		}
 
-		// Occasional network failures are OK; retry until download succeeds. Exponential decay would work
-		// much better than a simple loop.
 		return runutil.RetryWithLog(logger, time.Minute, iterCtx.Done(), func() error {
 			blocks, err := download(iterCtx, logger, bkt)
 			if err != nil {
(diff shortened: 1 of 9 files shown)


bwplotka:
:tada: Awesome! I will review tomorrow (:

jaseemabid:
Just a heads up, the program crashed a few times for me and I'm fixing the error handling right now to make it a bit more robust. Will update the PR later today.

jaseemabid:
Heads up! Google is being sort of evil here and won't allow the JS to be vendored locally. I didn't see that coming :( 

https://developers.google.com/chart/interactive/faq?csw=1#offline

image

bwplotka:
Wow, that is quite unexpected! Tracking everywhere ;p

I think this is totally fine for first iteration. You need internet usually to access object storage as well, right? Let's stick to this and add GH issue improving this later. What do you think? @jaseemabid ?

filippog:
> How serious of an issue is it for Thanos? I personally would hate it to be not able to run this without internet or relying on big G. I might have to rewrite the UI now, but thankfully its not much code. /shrug

FWIW I'm in the same boat, self-hosting Thanos is important together with not requiring internet access.

jaseemabid:
Works for now @bwplotka, but I'd probably rewrite it soon. Could you review the backend code for now? We could get some feedback from users (give it a try in one of your envs?) and use all that to iterate on a v2. I know that we might need some kind of filtering on label, maybe zoom into specific timeranges and better toooltips for example.

jaseemabid:
I've had this version run on prod for a few days now, I'll report tomorrow if it had any errors or crashes and the resource usage.

cmd/thanos/bucket.go
```diff
@@ -291,6 +299,123 @@ func registerBucketInspect(m map[string]setupFunc, root *kingpin.CmdClause, name
 	}
 }
 
+// RegisterBucketWeb exposes a web interface for the state of Thanos remote store just like `pprof web`
```
bwplotka:
```suggestion
// registerBucketWeb exposes a web interface for the state of Thanos remote store just like `pprof web`
```

cmd/thanos/bucket.go
```diff
@@ -291,6 +299,123 @@ func registerBucketInspect(m map[string]setupFunc, root *kingpin.CmdClause, name
...
+//
+// TODOS:
```
bwplotka:
```suggestion
// TODO(jaseemabid):
```

cmd/thanos/bucket.go
```diff
@@ -291,6 +299,123 @@ func registerBucketInspect(m map[string]setupFunc, root *kingpin.CmdClause, name
...
+//
+// 1. Some download code is duplicated from `ls -o json`; refactor as needed
+// 2. Some web code is duplicated from query UI, refactor as needed
+// 3. Do we need any of the route prefix?
```
bwplotka: Let's drop it, if we will need we will get a separate use case (and hopefully PR) from users.

cmd/thanos/bucket.go
```diff
@@ -291,6 +299,123 @@ func registerBucketInspect(m map[string]setupFunc, root *kingpin.CmdClause, name
...
+func registerBucketWeb(m map[string]setupFunc, root *kingpin.CmdClause, name string, objStoreConfig *pathOrContent) {
+	cmd := root.Command("web", "Web interface for remote storage bucket")
+	bind := cmd.Flag("listen", "HTTP host:port to listen on").Default("0.0.0.0:8080").String()
+	interval := cmd.Flag("refresh", "Refresh interval in minutes").Default("30").Int()
```
bwplotka: Why not `.Duration` so user can choose unit themselves?

cmd/thanos/bucket.go
```diff
@@ -291,6 +299,123 @@ func registerBucketInspect(m map[string]setupFunc, root *kingpin.CmdClause, name
...
+			return errors.Wrap(http.Serve(l, router), "serve web")
+		}, func(error) {
+			runutil.CloseWithLogOnErr(logger, l, "http listener")
+		})
+
+		return nil
+	}
+}
+
+// Refresh metadata from remote storage periodically and update UI
```
bwplotka: Missing trailing period.
```suggestion
// Refresh metadata from remote storage periodically and update UI.
```

cmd/thanos/bucket.go
```diff
@@ -291,6 +299,123 @@ func registerBucketInspect(m map[string]setupFunc, root *kingpin.CmdClause, name
...
+func refresh(ctx context.Context, logger log.Logger, bucketUI *ui.Bucket, interval int, name string,
```
bwplotka: This is not correct wrapping for Go args. Either single line or each argument in separate line.

cmd/thanos/bucket.go
```diff
@@ -291,6 +299,123 @@ func registerBucketInspect(m map[string]setupFunc, root *kingpin.CmdClause, name
...
+	reg *prometheus.Registry, objStoreConfig *pathOrContent) error {
+
+	confContentYaml, err := objStoreConfig.Content()
+	if err != nil {
+		return err
+	}
+
+	// A client is created for the entire duration of the program to avoid duplicate metrics registration.
```
bwplotka: Can we remove this comment unless we are realy blocked on having separate client for each refresh? I don't see the issue for now. If we need we can fix metrics (:

cmd/thanos/bucket.go
```diff
@@ -291,6 +299,123 @@ func registerBucketInspect(m map[string]setupFunc, root *kingpin.CmdClause, name
...
+	bkt, err := client.NewBucket(logger, confContentYaml, reg, name)
+	if err != nil {
+		return err
```
bwplotka: some `errors.Wrap(err, "bucket client")` could potentially help narrowing down an error but not a blocker (:

cmd/thanos/bucket.go
```diff
@@ -291,6 +299,123 @@ func registerBucketInspect(m map[string]setupFunc, root *kingpin.CmdClause, name
...
+	}
+
+	defer runutil.CloseWithLogOnErr(logger, bkt, "bucket client")
+
+	delay := time.Duration(interval) * time.Minute
```
bwplotka: We could potentially inline this as there is only single usage but if we change the flag you can use flag directly.

cmd/thanos/bucket.go
```diff
@@ -291,6 +299,123 @@ func registerBucketInspect(m map[string]setupFunc, root *kingpin.CmdClause, name
...
+
+	return runutil.Repeat(delay, ctx.Done(), func() error {
+		iterCtx, iterCancel := context.WithTimeout(ctx, 5*time.Minute)
+		defer iterCancel()
+
+		// If there are no errors and the data isn't stale, wait a while before updating again.
+		if bucketUI.Err == nil && time.Since(bucketUI.RefreshedAt) < (5*time.Minute) {
```
bwplotka: Does it mean that delay cannot be lower than 5 minutes? Worth to add to flag description and maybe add some validation?

cmd/thanos/bucket.go
```diff
@@ -291,6 +299,123 @@ func registerBucketInspect(m map[string]setupFunc, root *kingpin.CmdClause, name
...
+			return nil
+		}
+
+		// Occasional network failures are OK; retry until download succeeds. Exponential decay would work
+		// much better than a simple loop.
```
bwplotka: :+1:  fine for now, but not sure if there is point of `Exponential decay would work
		// much better than a simple loop.` sentence. Either put `TODO(jaseemabid)` in front or skip it. 

I think simple retry WITH some error message is fine for now

cmd/thanos/bucket.go
```diff
@@ -291,6 +299,123 @@ func registerBucketInspect(m map[string]setupFunc, root *kingpin.CmdClause, name
...
+	if err != nil {
+		return err
+	}
+
+	defer runutil.CloseWithLogOnErr(logger, bkt, "bucket client")
+
+	delay := time.Duration(interval) * time.Minute
+
+	return runutil.Repeat(delay, ctx.Done(), func() error {
+		iterCtx, iterCancel := context.WithTimeout(ctx, 5*time.Minute)
```
bwplotka: I think we might need to configure timeout at some point as well? Also it's tricky as timeout will mean that we will wait until ... next interval which can be in X minutes? We could simplify flow here a bit I think. I would retry until success OR timeout with hard failure on server. Continuing and waiting for next interval tick is like retrying with backoff to me (:

cmd/thanos/bucket.go
```diff
@@ -291,6 +299,123 @@ func registerBucketInspect(m map[string]setupFunc, root *kingpin.CmdClause, name
...
+
+		meta, err := block.DownloadMeta(ctx, logger, bkt, id)
+		if err != nil {
+			return err
+		}
+
+		blocks = append(blocks, meta)
+		return nil
+	})
+
```
bwplotka: Not needed empty line. Also I would vote for `if err := bkt.Iter(... ; err != nil {` pattern here for error scoping.

cmd/thanos/bucket.go
```diff
@@ -291,6 +299,123 @@ func registerBucketInspect(m map[string]setupFunc, root *kingpin.CmdClause, name
...
+	if err != nil {
+		level.Error(logger).Log("err", err, "msg", "Failed to downloaded block metadata")
+		return blocks, err
+	}
+
+	level.Info(logger).Log("msg", "downloaded block metadata")
```
bwplotka: Can we add number of blocks e.g we syncronized? Also let's be explicit:

```suggestion
	level.Info(logger).Log("msg", "downloaded blocks meta.json", "num", len(blocks))
```

cmd/thanos/bucket.go
jaseemabid: I wasn't aware of a `.Duration()`, that does make things easier 👍

cmd/thanos/bucket.go
```diff
@@ -4,29 +4,36 @@ import (
...
 	"os"
 	"sort"
 	"strings"
 	"text/template"
 	"time"
 
-	"github.com/go-kit/kit/log"
+	opentracing "github.com/opentracing/opentracing-go"
+	"github.com/prometheus/common/route"
+
```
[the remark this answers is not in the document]
jaseemabid: I can manually fix this now, but `make format` isn't fixing this for me. Are you using some custom config? There is a diff on master now if you run `make format`.

cmd/thanos/bucket.go
jaseemabid: I was trying to keep this loop really simple and not make it download a lot of data from s3. Since compact activity is a fairly slow operation, I guess data that's stale by a few minutes is not a big issue?

jaseemabid:
@bwplotka I'vent seen that UI error before and I wonder if its because of some difference in the way data is returned from GCS vs s3?

cmd/thanos/bucket.go
jaseemabid: I removed this check and added a simple warning if the user updates at a frequency less than 5m.

cmd/thanos/bucket.go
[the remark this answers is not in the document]
bwplotka: Yes, yes it's manual work for now until https://github.com/improbable-eng/thanos/issues/984 is fixed.

cmd/thanos/bucket.go
bwplotka: This minimum value makes sense, but yes we can check it early on instead.

bwplotka:
There is none as we have everything in our `pkg/objstore` abstraction. Maybe something in `meta.json` that is unusual? 

How can I give you more data for this error?

bwplotka:
It is some JS error, so wonder if it's not some JS versions mismatch? Or browser? I use chrome on linux.

jaseemabid:
@bwplotka I've pushed 2 more commits to make the error handling even more robust. More details in the commit message. 

ed0590d Add runutil.Forever since runutil.Repeat stops at first error
566ac2b Make thanos bucket web refresh forever ignoring errors ..

jaseemabid:
This is roughly how the generated JS block looks like for me;

```
   var blocks = [{"version":1,"ulid":"01CT...;
         err =  null ,
         refreshedAt = "2019-06-19T12:58:31.770801942Z";
```

There is no length function in any of the new code we added, so it might be somewhere in the library and its hard to debug without a line number. Could you privately share the html file with me? I'd debug and throw it away.

bwplotka:
Nice, will give it a go

cmd/thanos/bucket.go
```diff
@@ -291,6 +299,113 @@ func registerBucketInspect(m map[string]setupFunc, root *kingpin.CmdClause, name
...
+			data, err := json.Marshal(blocks)
+			if err != nil {
+				bucketUI.Set("[]", err)
+				return err
+			}
+			bucketUI.Set(string(data), nil)
+			return nil
+		})
+	})
```
bwplotka: Missing return here.

pkg/runutil/runutil.go
```diff
@@ -70,6 +70,24 @@ func Repeat(interval time.Duration, stopc <-chan struct{}, f func() error) error
 	}
 }
 
+// Forever executes f every interval seconds until stopc is closed. Errors are logged and ignored.
+// It executes f once right after being called.
+func Forever(logger log.Logger, interval time.Duration, stopc <-chan struct{}, f func() error) {
```
bwplotka: Do we really need this? Why not `Repeat` with `func()` never returning error? (:

jaseemabid:
Quick update on this! @bwplotka mentioned that using a single label to annotate the timelines won't work for everyone so now we have an option to customize it!  ⚙️

From 0057ae6c0c9e794d478a19bc2856cd86def46601, 

> Allow customizing timeline title with label
> 
> For small environments with a unique label to identify each Prometheus replica, set the title with a `--label`. If multiple labels are required to duplicate shards, use the default to get an external legend. 

If you have a unique label (like I do), pass `--label prometheus_replica` and you get something like this. ✨

Screenshot 2019-06-27 at 18 12 49

If you need multiple labels, you can skip the `--label` option and get a full legend below like this 🎹 

Screenshot 2019-06-27 at 18 17 49

Admittedly this is only a first draft and in @bwplotka's words, a shows some very ametuer web design skills! 

But, _almost all of our thanos instability issues were due to huge compact backlog and this is the graph that helped us track missing blocks and make it work reliably again_. It's not pretty, but this UI might help you debug some nasty issues. 

I'd love to get this out to a bunch of users and get some feedback before reiterating again. Anyone with interest and skills in modern frontend programming could rewrite this in one afternoon and I'd be very happy when that happens. Till then hopefully this will be of some use to a bunch of people 🤞

cmd/thanos/bucket.go
bwplotka: not addressed (:

pkg/runutil/runutil.go
bwplotka: I would remove it IMO unless you have strong opinion on this (:

pkg/ui/templates/bucket.html
```diff
@@ -0,0 +1,55 @@
...
+    <script src="{{ pathPrefix }}/static/vendor/rickshaw/rickshaw.min.js?v={{ buildVersion }}"></script>
+    <script src="{{ pathPrefix }}/static/vendor/moment/moment.min.js?v={{ buildVersion }}"></script>
+    <script src="{{ pathPrefix }}/static/vendor/moment/moment-timezone-with-data.min.js?v={{ buildVersion }}"></script>
+    <script src="{{ pathPrefix }}/static/vendor/eonasdan-bootstrap-datetimepicker/bootstrap-datetimepicker.min.js?v={{ buildVersion }}"></script>
+    <script src="{{ pathPrefix }}/static/vendor/bootstrap3-typeahead/bootstrap3-typeahead.min.js?v={{ buildVersion }}"></script>
+    <script src="{{ pathPrefix }}/static/vendor/fuzzy/fuzzy.js?v={{ buildVersion }}"></script>
+
+    <script src="{{ pathPrefix }}/static/vendor/mustache/mustache.min.js?v={{ buildVersion }}"></script>
+    <script src="{{ pathPrefix }}/static/vendor/js/jquery.selection.js?v={{ buildVersion }}"></script>
+    <!-- <script src="{{ pathPrefix }}/static/vendor/js/jquery.hotkeys.js?v={{ buildVersion }}"></script> -->
```
bwplotka: Is this commented thing needed?

pkg/ui/templates/bucket.html
```diff
@@ -0,0 +1,55 @@
...
+     };
+    </script>
+
+    <div id="err" class="container" style="display: none;">
+        <div class="alert alert-warning" role="alert"></div>
+    </div>
+
+    <div class="container-fluid" id="Compactions" style="height: 100%; width: 100%;"></div>
+
+    <div id="legend" class="container-fluid" style="display: none; padding-top: 50px;">
```
bwplotka: Can we add comment with Github issue to fix this? (as to remove legend and make it more clear - do not require user to look what number means) (: Let's create that issue if not there yet!

pkg/runutil/runutil.go
jaseemabid: Fixed at f4ff6df2684d46baa4585f6cfd0115e6780bbb63

pkg/ui/templates/bucket.html
jaseemabid: @bwplotka Can track progress here? https://github.com/improbable-eng/thanos/issues/1246#issuecomment-506681398

pkg/ui/templates/bucket.html
jaseemabid: Removed bebabfa71f7166950d70cb91281882702e980525

pkg/ui/templates/bucket.html
bwplotka: I would say it would be nice to have new issue to be clear, but up to you (:

bwplotka: Let me know what's the plan as it's the last item before merge. All rest is good :+1:

cmd/thanos/bucket.go
```diff
@@ -291,6 +299,124 @@ func registerBucketInspect(m map[string]setupFunc, root *kingpin.CmdClause, name
...
+
+	bkt, err := client.NewBucket(logger, confContentYaml, reg, name)
+	if err != nil {
+		return errors.Wrap(err, "bucket client")
+	}
+
+	defer runutil.CloseWithLogOnErr(logger, bkt, "bucket client")
+
+	return runutil.Repeat(duration, ctx.Done(), func() error {
+
```
bwplotka: is this newline needed?

cmd/thanos/bucket.go
jaseemabid: @bwplotka 83dda592358cded322c450a2dda733c435615200 simplifies the loop again and a7012e04be717d5ea91cd20c56b46c9ba89d87d0 made the arguments configurable.

cmd/thanos/bucket.go
jaseemabid: 83dda592358cded322c450a2dda733c435615200 simplifies the loop again and a7012e04be717d5ea91cd20c56b46c9ba89d87d0 made the arguments configurable.  Does this look ok?

pkg/ui/templates/bucket.html
bwplotka: ping, let's add TODO if we can

jaseemabid:
@bwplotka I guess this is good to go! 

This is probably the longest time or iterations I've ever had to spend on a PR! 😂You were super helpful! Thank you so much :D

pkg/ui/templates/bucket.html
```diff
@@ -0,0 +1,40 @@
...
+    </script>
+
+    <div id="err" class="container" style="display: none;">
+        <div class="alert alert-warning" role="alert"></div>
+    </div>
+
+    <div class="container-fluid" id="Compactions" style="height: 100%; width: 100%;"></div>
+
+    <!--
+     TODO(jaseemabid): The legend is only a small temporary workaround till we have better ways of showing the labels.
```
bwplotka: perfect :+1:

jaseemabid:
I'm equally picky with PRs and I'm honestly happy that you did what you did. Cheers! 🍺

PR #1250, opened 2019-06-13
pkg/downsample: use a.count

GiedriusS:
When aggregating downsampled chunks use `a.count` which actually points to
the data that we want to use. I don't know why `a.sum` was used here
before as it seems wrong.

Also, the tests now AFAICT show proper aggregation values.


GiedriusS:
NVM, `a.sum` is used because we don't want to actually lose the previous data

PR #1251, opened 2019-06-13
cmd/thanos/receive: eliminate unused code

squat:
This commit simplifies the setup of the receive Handler. The
`reloadReady` variable was not actually doing anything and adds some
complexity. We can add this back in once reloading is actually
implemented.

## Changes
Remove some dead code in receive component setup.

### DIFF
--- a/cmd/thanos/receive.go
+++ b/cmd/thanos/receive.go
@@ -4,7 +4,6 @@ import (
 	"context"
 	"fmt"
 	"net"
-	"sync"
 	"time"
 
 	"github.com/go-kit/kit/log"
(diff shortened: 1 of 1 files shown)


cmd/thanos/receive.go
```diff
@@ -110,38 +109,13 @@ func runReceive(
...
-				case <-dbOpen:
-					break
-				case <-cancel:
-					reloadReady.Close()
-					return nil
-				}
-
-				reloadReady.Close()
-
+				<-dbOpen
```
brancz: for this to not leak, we need to also close this channel in the tsdb run group no?

squat: hmm now that I think about it, we could really eliminate this entire run group and put the single line:

```
webHandler.Ready()
```

in the tsdb run group, no?

squat: in any case, we do already close it: https://github.com/improbable-eng/thanos/pull/1251/files#diff-074deab4cb04b87eb84800f5bbd8221fL176

brancz: Not if we return right? https://github.com/improbable-eng/thanos/pull/1251/files#diff-074deab4cb04b87eb84800f5bbd8221fR144

squat: ah yes, you're completely right. I'll add a defer

brancz: But yes, if there is nothing else but the readiness thing, then let's just remove the group

squat: ok simplified some more :) ptal when you get a change

brancz:
Thanks!

PR #1256, opened 2019-06-14
[Documentation] Add initial sequence diagrams

cfchad:
## Changes

Adding sequence diagrams to thanos

## Verification

No code change


docs/sequence-diagrams/compact.puml
```diff
@@ -0,0 +1,7 @@
+@startuml compact
+thanos_compact -> thanos_compact_config: Evaluate retention policy to see how long to keep samples
```
bwplotka: This is not really what's happening in order. I think the easiest way for you would to be go to implementation and look it up. 

It is rather:
* Sync meta files
* Compact
   * plan
   * download planned blocks to be compacted
   * compact 
   * upload compacted
   * delete old
* Downsample
   * ... 
* Apply Retention 

I don't really get what `thanos_compact_ config` is - it's not really an actor for sequential graph.

This means that as this doc of steps is useful, not sure if sequential graph is useful. If any I would see 3 "actors". 
* Thanos compact process
* Object storage (let's name it like that)
* Local filesystem

What do you think?

docs/sequence-diagrams/README.md
```diff
@@ -0,0 +1,10 @@
+# Sequence Diagrams
```
bwplotka: TBH I would say we would love to:
* Have those diagrams defined in some place
* Have automation (as part of `make docs?`) that will generate png those from `.puml` files.
* png will be used in `docs/components` for each component. (: What do you think?

docs/sequence-diagrams/generate-uml.sh
```diff
@@ -0,0 +1,7 @@
+#!/bin/bash
+
+java -jar plantuml.jar sidecar.puml
```
bwplotka: mhm is there some sequential diagram language that does not require java? This is blocker for CI I assume.

docs/sequence-diagrams/query.puml
```diff
@@ -0,0 +1,6 @@
+@startuml query
+client -> thanos_query: Request for data
```
bwplotka: There is more to it, but it's a good start. 

There is also deduplication and filtering step.

docs/sequence-diagrams/rule.puml
```diff
@@ -0,0 +1,3 @@
+@startuml rule
+thanos_rule -> thanos_store: data
```
bwplotka: This is not helpful, and quite misleading ): There is many things happening here. Let's drop this for now and introduce in another PRs with details what do you think?

docs/sequence-diagrams/sidecar.puml
```diff
@@ -0,0 +1,8 @@
+@startuml
+thanos_sidecar -> prometheus_data_dir: On init create thanos.shipper.json
```
bwplotka: This roughly sense, just maybe `cloud_storage` -> `object storage`

Also `Write block to storage` cannot be just `data dir` to itself, I would introduce another actor: Prometheus (: 

and there are many more interactions like getting flags or external labels. But this is good a start. I would say splitting PR and focusing on each component would be the easiest way (:

docs/sequence-diagrams/store.puml
```diff
@@ -0,0 +1,4 @@
+@startuml
+thanos_store -> cloud_storage: request metadata
```
bwplotka: Definitely too small amount of details, again as rule let's drop it and focus on this in details in another PRs .

bwplotka:
@cfchad what about https://github.com/lmika/goseq ? It looks good enough and it's easily cross platform compatible.

bwplotka:
Or maybe https://github.com/yogendra/plantuml-go which supports `puml` files!

cfchad:
@bwplotka all good suggestions, most of the items that were completely off were meant to be placeholders for future editors. leaning towards plantuml-go as I have run into issues in the past with goseq. I will update these when I get some free time.

PR #1260, opened 2019-06-18
Query/Ruler: Add metric for number of endpoints resolved for each configured address.

devnev:
## Changes

Add `thanos_querier_store_apis_dns_provider_results` / `thanos_ruler_query_apis_dns_provider_results` metrics.

`count(dns_provider_results)` tells us how many addresses have been
configured through flags and file-SD, while `sum(dns_provider_results)`
gives us the number of endpoints actually produced from these addresses.

## Verification

Ran `thanos query --store=localhost:10901`, navigated to `localhost:10902/metrics` and found the entry:
```
# HELP thanos_querier_store_apis_dns_provider_results The number of resolved endpoints for each configured address
# TYPE thanos_querier_store_apis_dns_provider_results gauge
thanos_querier_store_apis_dns_provider_results{addr="localhost:10901"} 1
```

### DIFF
--- a/pkg/discovery/dns/provider.go
+++ b/pkg/discovery/dns/provider.go
@@ -20,6 +20,7 @@ type Provider struct {
 	resolved map[string][]string
 	logger   log.Logger
 
+	resolverAddrs         *prometheus.GaugeVec
 	resolverLookupsCount  prometheus.Counter
 	resolverFailuresCount prometheus.Counter
 }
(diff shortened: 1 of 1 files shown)


pkg/discovery/dns/provider.go
```diff
@@ -91,6 +97,7 @@ func (p *Provider) Resolve(ctx context.Context, addrs []string) {
 		p.resolverLookupsCount.Inc()
 		if err != nil {
 			// The DNS resolution failed. Continue without modifying the old records.
+			p.resolved[addr] = p.resolved[addr] // Ensure metrics capture the result even if empty.
```
bwplotka: I don't really get this part is, it intentional?  Isn't this noop?
It's tricky what should be there TBH ;p Should we set results for both query to empty and metrics to 0? should we only adjust metrics? Not sure yet.

*

devnev: This ensures that the map contains an entry for the address. So when we iterate over the map keys for the metric, we report that there was an address configured regardless of whether we got some results here, got no results, or continue using previous results.

PR #1261, opened 2019-06-19
pkg/receive: register handler metrics

squat:
The thanos receive component instruments its handler with some metrics
but these metrics are currently never registered.

## Changes

Register the receive handler metrics with the given registerer during creation. 

## Verification
Scrape the receive component's /metrics endpoint to verify that the handler metrics are now present.

cc @brancz @bwplotka


brancz:
Oh my. Second time in two days. Someone really should invest in some static tooling to try to catch common mistakes like this.

lgtm :+1:

brancz:
As a side note, but should be done in a separate PR .. we probably want metrics for requests total

Issue #1191, opened 2019-05-29
sidecar: exceeded sample limit


jjneely:
When running queries via the Query component, we often get partial results and errors back from various sidecars that look like the below.  Definitely a Prometheus/Sidecar IP and not a Store component.  See graphic.

![thanos-query-20190529](https://user-images.githubusercontent.com/10540/58584970-80ee1c00-8225-11e9-928d-e974ce2fa69c.png)

We broke out tcpdump to figure out the exact problem.  The Prometheus remote read API is returning a HTTP 400 status code to the Sidecar which propagates the result to Thanos.  The actual error message is below:

```
HTTP/1.1 400 Bad Request
Content-Type: text/plain; charset=utf-8
X-Content-Type-Options: nosniff
Date: Wed, 29 May 2019 14:41:01 GMT
Content-Length: 33

exceeded sample limit (50000000)
```

As we are trying to build a transition path from direct Prometheus queries to querying data via Thanos, we have not yet lowered the retention of data on our Prometheus VMs.  So they still have 30 days of data (and 60 in some cases).  It appears that some of our Prometheus instances are large enough that Prometheus remote read queries end up returning a large amount of data and at times exceed this limit.  It also seems to vastly slow down query performance.

The preferred thing to do here, obviously, is run Prometheus VMs with limited retention as data will exist in GCS/Store.  We did this in our devel environment and problems went away, query performance increased, and using downsampled results seemed to work much better.  However, I find myself in an interesting position.  I don't feel comfortable calling my Thanos Query service "production" yet, and I do not feel comfortable removing retention from team's Prometheus VMs without a production Thanos Query service running.  It feels like there may be other reasons folks may wish to keep longer than a day or 3 of retention on their Prometheus VMs as well.

I'm wondering if this can be solved by adding a command line option to the Sidecar that would provide a duration like `3d`.  When the Sidecar process is running it would advertise and use the most recent value of `mint` or now-3d as the `mint` argument for Prometheus's remote read API.  This would effectively limit queries from the Sidecar to the Prometheus DB to the last 3 days.  The remaining long term data would be provided by GCS.  This would provide the best of both worlds, limits queries against Prometheus, and allowing us to keep retention until we have teams moved over to the Thanos endpoints.

**Thanos, Prometheus and Golang version used**

Thanos 0.4.0 with Goland 1.12.1.

**What happened**

Prometheus remote read APIs return 400 errors when the exact same query would run on the native Prometheus VM without issue and quite quickly.

[...]

PR #1267, opened 2019-06-20
sidecar: time limit requests to Prometheus remote read api

jjneely:
* [ENHANCEMENT] Add --prometheus.retention=0d flag to sidecar to limit the TSDB data made available from Prometheus to the last given duration.  The default is 0 or no limit.

## Changes

Implement #1191

The Sidecar keeps track of the mint/maxt of the TSDB data that's it makes available via the Prometheus Remote Read API.  This limits the mint to the given duration, such as 1d or 3d, and makes available only the last 1d or 3d worth of data from the local Prometheus.  

This can vastly improve query response times if Prometheus has more data than this on disk and long term storage is also uploaded into object storage.  Querying the Store component is more efficient than large queries via the Remote Read API.  This can also fix 400 errors returned by the Remote Read API when `storage.remote.read-sample-limit` is hit.

## Verification

* Monitored remote read queries hitting Prometheus that they were for the correct time period.  
* Tested the Query component for errors and speed of query execution.

### DIFF
--- a/cmd/thanos/sidecar.go
+++ b/cmd/thanos/sidecar.go
@@ -37,6 +37,8 @@ func registerSidecar(m map[string]setupFunc, app *kingpin.Application, name stri
 	promURL := cmd.Flag("prometheus.url", "URL at which to reach Prometheus's API. For better performance use local network.").
 		Default("http://localhost:9090").URL()
 
+	promRetention := modelDuration(cmd.Flag("prometheus.retention", "A limit on how much retention to query from Prometheus. 0d means query all TSDB data found on disk").Default("0d"))
+
 	dataDir := cmd.Flag("tsdb.path", "Data directory of TSDB.").
 		Default("./data").String()
 
@@ -71,6 +73,7 @@ func registerSidecar(m map[string]setupFunc, app *kingpin.Application, name stri
 			*clientCA,
 			*httpBindAddr,
 			*promURL,
+			time.Duration(*promRetention),
 			*dataDir,
 			objStoreConfig,
 			rl,
@@ -90,6 +93,7 @@ func runSidecar(
 	clientCA string,
 	httpBindAddr string,
 	promURL *url.URL,
+	promRetention time.Duration,
 	dataDir string,
 	objStoreConfig *pathOrContent,
 	reloader *reloader.Reloader,
@@ -100,8 +104,9 @@ func runSidecar(
 
 		// Start out with the full time range. The shipper will constrain it later.
 		// TODO(fabxc): minimum timestamp is never adjusted if shipping is disabled.
-		mint: 0,
-		maxt: math.MaxInt64,
+		mint:  0,
+		maxt:  math.MaxInt64,
+		limit: promRetention,
 	}
 
 	confContentYaml, err := objStoreConfig.Content()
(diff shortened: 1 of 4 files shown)


bwplotka:
Hi, 

Do you think this is still needed if we make remote read more efficient as it's in progress now? https://github.com/improbable-eng/thanos/pull/1268

jjneely:
Perhaps?  I probably don't fully understand the benefits the streaming remote read API will bring.  But it seems that if you are using Stores and object buckets for metric storage, then you never want to ask the Sidecar/Prometheus for anything more than the last few hours of data.  Or do these changes make Sidecar/Prometheus more efficient to query than the Store/bucket?

I'm migrating from a world where we have 30 days or more retention on our Prometheus VMs into Thanos which moves the long term data into object storage.  I can't limit the retention on the Prometheus VMs until the Thanos Query component is production ready.  However, the Query component was requesting large date ranges from Prometheus that was blowing the remote read sample limit and the same date ranges from object storage.  It was the 400 HTTP status code responses from Prometheus that sent me down this quest.

This also looks like a Prometheus 2.10-ish upgrade would be required for this to work.

povilasv:
So regarding this PR I agree with @bwplotka, I don't think we need it, if we have streaming remote reads.

Regarding:

> This can vastly improve query response times if Prometheus has more data than this on disk and long term storage is also uploaded into object storage. Querying the Store component is more efficient than large queries via the Remote Read API. 

I don't really agree, if we have streamed remote reads, querying Prometheus should be faster and more efficient than Store, as Prometheus would be hitting local disk worst case scenario and Thanos Store - Object Store.

In my testing of https://github.com/thanos-io/thanos/pull/1077#issuecomment-495593281 I actually seen that query times are shorter when you skip Thanos Store for short periods of data and hit only Prometheus.

cmd/thanos/sidecar.go
```diff
@@ -37,6 +37,8 @@ func registerSidecar(m map[string]setupFunc, app *kingpin.Application, name stri
 	promURL := cmd.Flag("prometheus.url", "URL at which to reach Prometheus's API. For better performance use local network.").
 		Default("http://localhost:9090").URL()
 
+	promRetention := modelDuration(cmd.Flag("prometheus.retention", "A limit on how much retention to query from Prometheus. 0d means query all TSDB data found on disk").Default("0d"))
```
bwplotka: Curious about this. Should we do retention? I would rather see `storeapi.min-time` with similar model to specify both relative and absolute time as here: https://github.com/improbable-eng/thanos/pull/1077 

This is to keep it consistent with potential store gateway time partitioning.

jjneely: This was what I struggled with most -- how to name and handle this argument.  I'll take a look at implementing the suggestions.  I'm all about consistency.

povilasv: yy, if we are doing this then `--storeapi.min-time`, `--storeapi.max-time` sounds like good options.

You can copy paste the https://github.com/thanos-io/thanos/pull/1077/files#diff-dd29e6298d43e46bb651035051819cfcR14 class from my PR, to make it take same exact format.

bwplotka:
I think we have some decision mentioned here: https://github.com/thanos-io/thanos/issues/1191

Happy to approve this once rebased AND @povilasv comment will be addressed (: 

Sorry for the massive lag on this @jjneely ! Are you still around to continue on this?

bwplotka:
Addressed and rebased here: https://github.com/thanos-io/thanos/pull/1619 (:

PR #1269, opened 2019-06-21
Docs update on prometheus_rule_group_iterations_missed_total metric f…

homelessnessbo:
Signed-off-by: Ivan Kiselev <kiselev_ivan@pm.me>

## Changes

A little doc update on the ruler that comes from here https://improbable-eng.slack.com/archives/CA4UWKEEN/p1560784841270000

Mentioned metric `rule_group_iterations_missed_total_total ` helps to avoid a picture like this: 

Left part indicates a situation when eval interval is lower than actual eval time

Screenshot 2019-06-20 at 13 57 58

### DIFF
--- a/docs/components/rule.md
+++ b/docs/components/rule.md
@@ -11,7 +11,7 @@ _**NOTE:** It is recommended to keep deploying rules inside the relevant Prometh
 _The rule component should in particular not be used to circumvent solving rule deployment properly at the configuration management level._
 
 The rule component evaluates Prometheus recording and alerting rules against chosen query API via repeated `--query` (or FileSD via `--query.sd`). If more than one query is passed, round robin balancing is performed.
- 
+
 Rule results are written back to disk in the Prometheus 2.0 storage format. Rule nodes at the same time participate in the system as source store nodes, which means that they expose StoreAPI and upload their generated TSDB blocks to an object store.
 
 You can think of Rule as a simplified Prometheus that does not require a sidecar and does not scrape and do PromQL evaluation (no QueryAPI).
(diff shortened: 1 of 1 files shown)


docs/components/rule.md
```diff
@@ -84,20 +84,24 @@ indicate connection, incompatibility or misconfiguration problems.
...
+* `prometheus_rule_group_last_duration_seconds < prometheus_rule_group_interval_seconds`  If the difference is large, it means
+that rule evaluation took more time than the scheduled interval. It can indicate that your query backend (e.g Querier) takes too much time
+to evaluate the query, i.e. that it is not fast enough to fill the rule. This might indicate other problems like slow StoreAPis or
+too complex query expression in rule.
 
 * `thanos_rule_evaluation_with_warnings_total`. If you choose to use Rules and Alerts with [partial response strategy's](rule.md#partial-response)
 value as "warn", this metric will tell you how many evaluation ended up with some kind of warning. To see the actual warnings
 see WARN log level. This might suggest that those evaluations return partial response and might not be accurate.
 
+* `prometheus_rule_group_last_duration_seconds > prometheus_rule_group_interval_seconds`. With all the networking overhead of thanos - you might end up in a situation when actual group evaluation
```
bwplotka: Would argue if that `with all the networking overhead` comment is useful. You can end up in this situation for Prometheus as well. Also you might have local setup, so there might be not much of networking overhead, so it really depends.
 
```suggestion
* `prometheus_rule_group_last_duration_seconds > prometheus_rule_group_interval_seconds`.You might end up in a situation when actual group evaluation
```

docs/components/rule.md
```diff
@@ -84,20 +84,24 @@ indicate connection, incompatibility or misconfiguration problems.
...
+time is bigger than the evaluation interval which leads to an incomplete data set. If you see this alert firing - you most likely need to tune
+evaluation interval or divide rules to smaller groups. (Rules within one group are executed sequentially and different groups are executed in parallel)
```
bwplotka:
```suggestion
evaluation interval or divide rules to smaller groups. Note that rules within one group are executed sequentially in ordered manner. Groups are executed concurrently to each other.
```

homelessnessbo:
jeez christ. It's already there >_< https://github.com/improbable-eng/thanos/pull/1269/files#diff-0f0fe7c1be3f2ba3063727e4fd8bf733L87
I made such a long way to something that was always here.

PR #1270, opened 2019-06-21
receive: Replication

squat:
This commit adds a new replication feature to the Thanos receiver.
By default, replication is turned off, however, when replication is
enabled (replication factor >=2), the target node for a time series will
replicate the time series to the other nodes concurrently and
synchronously. If the replication to >= (rf+1)/2 nodes fails, the
original write request is failed.

## Changes

* add optional replication to the Thanos receive component
* add e2e tests to exercise replication

## Verification

* e2e tests
* manual testing to ensure replicated metrics appear in querier
* checking new metrics to ensure replication errors don't increase in regular operation

### DIFF
--- a/docs/proposals/approved/201812_thanos-remote-receive.md
+++ b/docs/proposals/approved/201812_thanos-remote-receive.md
@@ -26,7 +26,7 @@ Prometheus has the remote write API to send samples collected by a Prometheus se
 
 ## Architecture
 
-The Thanos receiver component seamlessly integrates into the rest of the Thanos components. It acts similarly to what is referred to in Thanos as a "source", in the current set of components, this is typically represented by the Thanos sidecar that is put next to Prometheus to ship tsdb blocks into object storage and reply to store API requests, however in the case of the Thanos receiver, the Thanos sidecar is not necessary anymore, as the data is replicated from the original Prometheus server to the Thanos receiver, and the Thanos receiver participates in the Thanos gossip mesh. The Prometheus server on a tenant’s infrastructure can therefore be completely vanilla and is just configured to replicate its time-series to the Thanos receiver.
+The Thanos receiver component seamlessly integrates into the rest of the Thanos components. It acts similarly to what is referred to in Thanos as a "source", in the current set of components, this is typically represented by the Thanos sidecar that is put next to Prometheus to ship tsdb blocks into object storage and reply to store API requests, however in the case of the Thanos receiver, the Thanos sidecar is not necessary anymore, as the data is replicated from the original Prometheus server to the Thanos receiver, and the Thanos receiver participates in the Thanos gossip mesh. The Prometheus server on a tenant's infrastructure can therefore be completely vanilla and is just configured to replicate its time-series to the Thanos receiver.
 
 Instead of directly scraping metrics, however, the Thanos receiver accepts Prometheus remote-write requests, and writes these into a local instance of the Prometheus tsdb. Once successfully committed to the tenant's tsdbs, the requests returns successfully. To prevent data leaking at the database level, each tenant has an individual tsdb instance, meaning a single Thanos receiver may manage multiple tsdb instances. The receiver answers Thanos store API requests and uploads built blocks of the Prometheus tsdb. Implementation-wise, this just requires wiring up existing components. As tenant's data within object storage are separate objects, it may be enough separation to have a single bucket for all tenants, however, this architecture supports any setup of tenant to object storage bucket combination.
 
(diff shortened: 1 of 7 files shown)


brancz:
As replication wasn't specified in the original design, could you move the paragraph from open questions to the "Rollout/scaling/failure of receiver nodes" section and add appropriate detail?

pkg/receive/handler.go
```diff
@@ -245,17 +313,37 @@ func (h *Handler) forward(ctx context.Context, tenant string, wreq *prompb.Write
...
 
+	var errs error
+	// Collect all the errors into a single error.
+	for _, err := range h.parallelizeRequests(ctx, tenant, replicas, wreqs, forwardRequestsTotal, forwardErrorsTotal) {
+		if err != nil {
+			if errs == nil {
+				errs = err
+				continue
+			}
+			errs = errors.Wrap(errs, err.Error())
```
brancz: MultiErrror makes more sense here no? In fact `.parallelizeRequsets` could immediately return a MultiError

pkg/receive/handler.go
```diff
@@ -49,16 +50,42 @@ var (
 		},
 		[]string{"handler"},
 	)
+	replicationRequestsTotal = prometheus.NewCounter(
```
brancz: let's move not only the registration, but also the initialization of these to the `NewHandler` function

pkg/receive/handler.go
squat: Sounds good! I didn’t realize we had a util for that :)

pkg/receive/handler.go
metalmatze: Then we should also pass the `prometheus.Registry`?!

pkg/receive/handler.go
```diff
@@ -288,13 +373,30 @@ func (h *Handler) forward(ctx context.Context, tenant string, wreq *prompb.Write
 				return
 			}
 			req.Header.Add(h.options.TenantHeader, tenant)
+			req.Header.Add(h.options.ReplicaHeader, strconv.FormatUint(replicas[endpoint], 10))
+
+			// Increment the counters as necessary now that
+			// the requests will go out.
+			defer func() {
+				requestCounter.Inc()
```
metalmatze: This will also be increased, even when there is an error. Is that intended?

pkg/receive/handler.go
brancz: Precisely.

pkg/receive/handler.go
squat: yes, this is tracking the _total_, number of requests. failed or successful. Using the error counter we can know the percentage of errors.

brancz: it's probably worth unifying in one metric with different labels for success/failure

metalmatze: Yep, I'd prefer having a unified counter. :+1:

squat: so don’t track total; instead use one counter vec with a result label for success or failure. When we want to know the percentage of failed request we do errors/(errors+successes)?

pkg/receive/handler.go
squat: The registry is already passed into NewHandler via the options struct :+1:

pkg/receive/handler.go
brancz: correct

docs/proposals/approved/201812_thanos-remote-receive.md
```diff
@@ -157,11 +165,29 @@ To start, exact matches of tenant IDs will used to distribute requests to receiv
...
+
+### Replication
+
+The Thanos receiver supports replication of received time-series to other receivers in the same hashring. The replication factor is controlled by setting a flag on the receivers and indicates the maximum number of copies of any time-series that should be stored in the hashring. If any time-series in a write request received by a Thanos receiver is not successfully written to at least `(REPLICATION_FATOR + 1)/2` nodes, the receiver responds with an error. For example, to attempt to store 3 copies of every time-series and ensure that every time-series is successfully written to at least 2 Thanos receivers in the target hashring, all receivers should be configured with the following flag:
```
bwplotka:
```suggestion
The Thanos receiver supports replication of received time-series to other receivers in the same hashring. The replication factor is controlled by setting a flag on the receivers and indicates the maximum number of copies of any time-series that should be stored in the hashring. If any time-series in a write request received by a Thanos receiver is not successfully written to at least `(REPLICATION_FACTOR + 1)/2` nodes, the receiver responds with an error. For example, to attempt to store 3 copies of every time-series and ensure that every time-series is successfully written to at least 2 Thanos receivers in the target hashring, all receivers should be configured with the following flag:
```

PR #1274, opened 2019-06-24
store/cache: remove iteration limit and reset on internal inconsistency

### LINKED ISSUE #1257: Store: Hundreds of errors in logs
Hi

I have hundreds of error in logs of store daemon, like:
<details>Logs
<p>

```
{"cacheType":"Postings","caller":"cache.go:254","curSize":8587457979,"itemSize":4253828,"iterations":500,"level":"error","maxItemSizeBytes":4294967296,"maxSizeBytes":8589934592,"msg":"After max sane iterations of LRU evictions, we still cannot allocate the item. Ignoring.","ts":"2019-06-17T07:07:37.983393384Z"}
```
</p>
</details>

I think this is not real problem. Maybe change log level to warning?

```
./thanos --version
thanos, version 0.5.0 (branch: HEAD, revision: 72820b3f41794140403fd04d6da82299f2c16447)
  build user:       root@7d72e9360b09
  build date:       20190606-10:49:10
  go version:       go1.12.5
```

abursavich:
## Changes

This affects adding new items to the cache:
1. It removes the max iterations restriction.  See https://github.com/improbable-eng/thanos/issues/1257#issuecomment-503824210 for reasoning.
2. If the item is small enough to fit but doesn't fit even after the cache has been emptied, the cache is reset to recover from an internal accounting error (this should never occur).

## Verification

Updated tests.

## Notes

Please let me know if you'd like a CHANGELOG entry.

Fixes #1257

### DIFF
--- a/pkg/store/cache/cache.go
+++ b/pkg/store/cache/cache.go
@@ -235,8 +235,6 @@ func (c *IndexCache) set(typ string, key cacheKey, val []byte) {
 // ensureFits tries to make sure that the passed slice will fit into the LRU cache.
 // Returns true if it will fit.
 func (c *IndexCache) ensureFits(size uint64, typ string) bool {
-	const saneMaxIterations = 500
-
 	if size > c.maxItemSizeBytes {
 		level.Debug(c.logger).Log(
 			"msg", "item bigger than maxItemSizeBytes. Ignoring..",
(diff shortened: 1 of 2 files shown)


abursavich:
That test replaced the LRU and changed its logic such that nothing could be evicted. I changed it such that it only breaks the current size accounting. However, I wouldn't normally suggest writing tests that reach into internal state and break invariants.

bwplotka:
.. and we fotgot about CHANGELOG. Something to remember to add during release.

Issue #1014, opened 2019-04-07
Compact: Offline deduplication


smalldirector:
Currently Thanos provides dedup function in Query component, it can merge the data on the fly from Prometheus HA pairs or remote object storage.
However, the Query dedup function is not query efficiently, because the replica's blocks in tsdb/object storage are not really merged so that the Query API has to load duplicated data in each request.

With current Prometheus TSDB design, it seems difficult to implement dedup function to merge blocks in each TSDB node on Thanos side.
However, it should be easily to implement it for the object storage as it is one centralized storage, and also we already have one Compactor component running on it.

Offering dedup function on object storage side will definitely help fast metrics query latency and reduce the object storage cost.

Do we have any plan to support such requirements in Thanos? If had, I would like to know your ideas about this feature.

Thanks.

bwplotka:
Hi :wave: thanks for raising this. I rename title to make it more clear, let me know if this makes sense. 

There was always idea like this to allow compactor to deduplicate blocks offline and reduce the storage and bit the query performance. I think I am happy to do so however, we need ourselves question, is the curret deduplication algorithm a correct one for everybody?

We have seen some reports claiming that our "penalty-based" deduplication algorithm is not working for all edge cases e.g: https://github.com/improbable-eng/thanos/issues/981 It's fair as our alghorithm is very basic. The problem with this feature and not perfect algorithm is that we will deduplicate data irreversibily (unless we back up those blocks somehow). 

Anyway, I think we are happy to add such feature to compactor at some point, given we can solve the gaps, for current algorithm (and add more tests/ test cases maybe).

smalldirector:
Thanks for the reply @bwplotka. I'm glad to hear that you also want to support this feature in the future. 
I'm working in eBay's monitoring infrastructure team, and we are leveraging Thanos to build eBay's high availability monitoring platform. Thanks for providing such great framework.

For the offline dedup function, we have already started to build it inside the Thanos compactor component. If needed, we are more than happy to contribute it back to Thanos community. Please feel free to let me know your thoughts as well.

bwplotka:
Of course! PRs would be welcome, especially if you have something proven from your prod (:

bwplotka:
Are you on slack BTW? (:

smalldirector:
[...]

PR #1275, opened 2019-06-24
Compact: Offline deduplication

smalldirector:
## Changes

As the discussion in #1014, implement a new feature in compactor which provides offline deduplication function for the data from different replicas.

The offline deduplication follow the same design with [query deduplication](https://thanos.io/components/query.md/#deduplication). The user needs to specify replica label by config `dedup.replica-label` before enable it, and the function uses current query deduplication algorithm(`penalty algorithm`) to merge data points come from different replicas.

The offline deduplication function is based on bucket level in remote storage, so the user needs to ensure that all replica data write to same bucket.

## Verification

- Compared data quality by tsdb read API.
- Compared the graph by Thanos query UI.

Below figures are the comparison for one sample metrics before dedup and after dedup. It defines one interval replica label `_agg_replica_` to represent the merged data.

Before Dedup
After Dedup

- Tested the dedup function online, and monitoring its metrics(especially memory usage).

For below figure, each of block size is around 3GB and we have two replicas for each block. As it is using streaming read/write way to operate block, so no OOM exception happens.

Metrics

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -17,6 +17,8 @@ We use *breaking* word for marking changes that are not backward compatible (rel
 
 - [#1253](https://github.com/improbable-eng/thanos/pull/1253) Add support for specifying a maximum amount of retries when using Azure Blob storage (default: no retries).
 
+- [#1275](https://github.com/improbable-eng/thanos/pull/1275) Added `offline dudeplication` function in Thanos compactor.
+
 ## [v0.5.0](https://github.com/improbable-eng/thanos/releases/tag/v0.5.0) - 2019.06.05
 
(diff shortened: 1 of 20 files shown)


bwplotka:
Why it was closed? Also can we ensure the PR is split into smaller bits if possible? It will ensure quicker review for sure (:

smalldirector:
@bwplotka Closed it as I want to have a clean commit history for the review. Actually I opened another new PR: https://github.com/improbable-eng/thanos/pull/1276. Can you please check the codes there?

Agree. Smaller PR will make the review easier. However, those codes are for one new feature `dedup function`, IMO keep everything in one PR will give a better context to know how it works there. 
Do you have any suggestion on how to split it?

PR #1276, opened 2019-06-24
Compact: Offline deduplication

smalldirector:
## Changes

As the discussion in #1014, implement a new feature in compactor which provides offline deduplication function for the data from different replicas.

The offline deduplication follow the same design with [query deduplication](https://thanos.io/components/query.md/#deduplication). The user needs to specify replica label by config `dedup.replica-label` before enable it, and the function uses current query deduplication algorithm(`penalty algorithm`) to merge data points come from different replicas.

The offline deduplication function is based on bucket level in remote storage, so the user needs to ensure that all replica data write to same bucket.

## Verification

- Compared data quality by tsdb read API.
- Compared the graph by Thanos query UI.

Below figures are the comparison for one sample metrics before dedup and after dedup. It defines one interval replica label `_agg_replica_` to represent the merged data.

Before Dedup
After Dedup

- Tested the dedup function online, and monitoring its metrics(especially memory usage).

For below figure, each of block size is around 3GB and we have two replicas for each block. As it is using streaming read/write way to operate block, so no OOM exception happens.

Metrics


smalldirector:
@bwplotka I submitted the PR for offline deduplication function, please help add reviewers and review the codes. Thanks.

bwplotka:
Thanks @smalldirector and we definitely need that (: Is there any way you can split this PR to smaller chunks in any way? It will will improve reviewing process majorily!

smalldirector:
@bwplotka Good question here. I was trying to split the PR before, however I didn't figure out one good way.  For my understanding, we need to keep the function work in the PR before it is merged. However my PR only provide one function `dedup`. 
Do you have any suggestion on how to break it?

smalldirector:
@bwplotka The dudep function support both of raw metrics and downsampling metrics, should I split it to two separate PRs(one is for raw metrics, another is for downsampling metrics)?

pkg/compact/dedup/replica.go
```diff
@@ -0,0 +1,258 @@
...
+	result := make([]*metadata.Meta, 0)
+	for id, b := range blocks {
+		if _, ok := remote[id]; ok {
+			result = append(result, b)
+		}
+	}
+	return NewReplicas(s.labelName, result)
+}
+
+func (s *ReplicaSyncer) download(ctx context.Context, id ulid.ULID) (*metadata.Meta, error) {
```
bwplotka: Not a fan of adding yet another Download method. Same with Syncer. Can we reuse those?

We already have problem with having multiple Syncer so we have https://github.com/thanos-io/thanos/issues/1335

pkg/compact/dedup/replica.go
```diff
@@ -0,0 +1,258 @@
...
+		logger:               logger,
+		metrics:              metrics,
+		bkt:                  bkt,
+		labelName:            labelName,
+		consistencyDelay:     consistencyDelay,
+		blockSyncConcurrency: blockSyncConcurrency,
+	}
+}
+
+func (s *ReplicaSyncer) Sync(ctx context.Context) (Replicas, error) {
```
bwplotka: We already have this for compactor. Can we reuse somehow?

pkg/compact/dedup/reader.go
```diff
@@ -0,0 +1,405 @@
...
+	logger  log.Logger
+	closers []io.Closer
+
+	ir tsdb.IndexReader
+	cr tsdb.ChunkReader
+
+	postings index.Postings
+}
+
+func NewBlockReader(logger log.Logger, resolution int64, blockDir string) (*BlockReader, error) {
```
bwplotka: Why we need this? We already have `tsdb.BlockReader`

pkg/compact/dedup/reader.go
```diff
@@ -0,0 +1,405 @@
...
+	return samples
+}
+
+type SampleType uint8
+
+const (
+	// samples from non-downsample blocks
+	RawSample SampleType = iota
+	// samples from downsample blocks, map to downsample AggrType
+	CountSample
```
bwplotka: Can we just reuse `AggrType`? This will avoid those mappings and simplify code (:

pkg/compact/dedup/reader.go
```diff
@@ -0,0 +1,405 @@
...
+		chks = append(chks, chunks.Meta{
+			MinTime: samples[start].timestamp,
+			MaxTime: samples[end-1].timestamp,
+			Chunk:   c,
+		})
+	}
+	return chks, nil
+}
+
+type SampleReader struct {
```
bwplotka: Hm. We deduplicate in `Iterators` interface currently... Wonder if we can reuse same algorithm as here: https://github.com/thanos-io/thanos/blob/master/pkg/query/iter.go#L375

Again - to reuse code to reduce overall complexity for understanding Thanos codebase (: and to gain benefits from improvements by changing just single place (:

pkg/compact/dedup/merger.go
```diff
@@ -0,0 +1,506 @@
...
+	"github.com/prometheus/tsdb/labels"
+	"github.com/thanos-io/thanos/pkg/block"
+	"github.com/thanos-io/thanos/pkg/block/metadata"
+	"github.com/thanos-io/thanos/pkg/compact"
+	"github.com/thanos-io/thanos/pkg/compact/downsample"
+	"github.com/thanos-io/thanos/pkg/objstore"
+	"github.com/thanos-io/thanos/pkg/query"
+)
+
+type TimeWindow struct {
```
bwplotka: We already have `tsdb.TimeRange{}`

Can we reuse it?

pkg/compact/dedup/merger.go
```diff
@@ -0,0 +1,506 @@
...
+	return fmt.Sprintf("BlockGroup{window: %s, blocks: %s}", g.window, builder.String())
+}
+
+func NewBlockGroup(window *TimeWindow, blocks []*metadata.Meta) *BlockGroup {
+	return &BlockGroup{window: window, blocks: blocks}
+}
+
+type BlockGroups []*BlockGroup
+
+func NewBlockGroups(replicas Replicas) BlockGroups {
```
bwplotka: This might be doable in Iterator as well. I think it would more understandable if we would have just one method/algo for this - for both offline and online deduplication. Online deduplication is available here: https://github.com/thanos-io/thanos/blob/master/pkg/query/iter.go#L304

pkg/compact/dedup/dedup_test.go
```diff
@@ -0,0 +1,153 @@
...
+					}
+					result[k][v.timestamp] = append(result[k][v.timestamp], v.value)
+				}
+			}
+		}
+	}
+	return result
+}
+
+func getBlockSampleSeries(t *testing.T, logger log.Logger, blockDir string) map[string][]*Sample {
```
bwplotka: I think sticking to iterator might be less complex / memory consuming then just buffering everything.

Keep in mind that this block might be 500GB big sometimes! We don't want to buffer all if possible. Iterator allows for parital flush on disk and streaming approach.

pkg/compact/dedup/replica.go
smalldirector: Sure, will figure out one proper way to reuse the functions in the `compact.Syncer`.
From the #1335, I saw we would start to unify the syncer APIs.  
I think if I can do this change after #1335, it will save some efforts.

pkg/compact/dedup/replica.go
smalldirector: same with above.

pkg/compact/dedup/reader.go
smalldirector: When try to get the sorted postings for one block, it needs to call four functions to make it: `OpenBlock(), Index(), Postings(), SortedPostings()`. Here just use this `BlockReader ` to provide a convenient way to access one block's postings and chunk reader.

pkg/compact/dedup/reader.go
smalldirector: Ok, let me figure out the way to handle raw sample with AggrType.

pkg/compact/dedup/reader.go
smalldirector: Actually this `SampleReader` doesn't do the deduplication logic. It just provides the function to read the samples under the specified time range.

smalldirector: And using `DedupSeriesIterator` to do the dedup logic: https://github.com/thanos-io/thanos/blob/62cc9bf4fedd601209d8197daf9163ae484afae1/pkg/compact/dedup/merger.go#L421
So both of online and offline deduplication are using the same `penalty` algorithm.

pkg/compact/dedup/merger.go
smalldirector: Good to know tsdb has `TimeRange `. Will change to use it.

pkg/compact/dedup/dedup_test.go
smalldirector: This function is only used in test cases, so I think we don't need to create the buffering issue here.

pkg/compact/dedup/merger.go
smalldirector: The `BlockGroups ` provides the function to group all the blocks under the specified time range. It doesn't provide any deduplication here.

smalldirector:
@bwplotka Thanks for your 1st round review. Totally agree with you that we should leverage more on the existed codes. For the time range and aggr type, I made the code changes. For the sync and downloading block, as I mentioned above, the #1335 will do the refactor task on those function. If I do it after the #1335, it will save us lots of efforts. Currently the offline deduplication logic follows the same dedup algorithm with online deduplication, so no code changes need to be done here.

bwplotka:
Cool @smalldirector , so is it ready for second review iteration? (:

smalldirector:
@bwplotka Yes, please help do the next round interview. For the meta syncing and block downloading, I would like to do it after #1335 is fixed.

bwplotka:
We need to get back to this finally (: cc @GiedriusS @povilasv @brancz 

Sorry for delay.

smalldirector:
Rebased the codes for the upcoming review.

cmd/thanos/compact.go
```diff
@@ -257,7 +266,14 @@ func runCompact(
 		level.Info(logger).Log("msg", "retention policy of 1 hour aggregated samples is enabled", "duration", retentionByResolution[compact.ResolutionLevel1h])
 	}
 
+	deduper := dedup.NewBucketDeduper(logger, reg, bkt, dedupDir, dedupReplicaLabel, consistencyDelay, blockSyncConcurrency)
+
 	f := func() error {
+		if isEnableDedup(enableDedup, dedupReplicaLabel) {
```
povilasv:
```suggestion
		if enableDedup && len(dedupReplicaLabel) > 0 {
```

cmd/thanos/compact.go
```diff
@@ -463,3 +479,7 @@ func generateIndexCacheFile(
 	}
 	return nil
 }
+
+func isEnableDedup(enableDedup bool, dedupReplicaLabel string) bool {
```
povilasv: Consider getting rid of this one line function, IMO no point in abstracting single line of code :)

pkg/compact/compact.go
```diff
@@ -665,7 +665,7 @@ type RetryError struct {
 	err error
 }
 
-func retry(err error) error {
+func Retry(err error) error {
```
povilasv: Does this need to be exported method? If yes add comment, but I would keep it internal if possible

pkg/compact/dedup/dedup.go
```diff
@@ -0,0 +1,79 @@
...
+	"time"
+
+	"github.com/go-kit/kit/log"
+	"github.com/go-kit/kit/log/level"
+	"github.com/pkg/errors"
+	"github.com/prometheus/client_golang/prometheus"
+	"github.com/thanos-io/thanos/pkg/objstore"
+)
+
+type BucketDeduper struct {
```
povilasv: We are missing comments on exported structs / methods / functions.

Reamer:
Hi @smalldirector,

At first, thank you for developing this feature. I tested your feature on my test environment and I will keep this running. If I encounter issues, i'll let you known.

# Used code
I've done a [merge with thanos v0.8.1](https://github.com/Reamer/thanos/commits/dedup_test).

# Before De-duplication
I used `thanos bucket web` with `--label=prometheus_replica`.
I have two prometheus statefulsets in a HA setup, created with prometheus-operator. One is called prometheus-k8s, the other prometheus-avm.
Pre_Dedup

# During De-duplication
For both prometheus setups I noticed a lot of warn message of the kind `find  empty downsample chunk`  during de-duplication of resolution 3600000.
Full error message:
```
resolution=3600000 metric in prometheus-k8s job:
{"caller":"reader.go:324","labels":"{__name__=\"admission_quota_controller_work_duration\",endpoint=\"https\",instance=\"10.20.15.13:443\",job=\"apiserver\",namespace=\"default\",quantile=\"0.99\",service=\"kubernetes\"}","level":"warn","msg":"find empty downsample chunk","ref":887438,"ts":"2019-10-16T07:16:23.650581433Z","type":"counter"}

resolution=3600000 metric in prometheus-avm job:
{"caller":"reader.go:324","labels":"{__name__=\"http_request_duration_microseconds\",endpoint=\"http\",handler=\"flags\",instance=\"10.130.17.27:10902\",job=\"thanos-metrics\",namespace=\"thanos\",pod=\"thanos-querier-657d5f7d9-qvzmq\",quantile=\"0.5\",service=\"thanos-metrics\"}","level":"warn","msg":"find empty downsample chunk","ref":51486867,"ts":"2019-10-17T05:57:53.170766602Z","type":"counter"}
```
first_dedbug_3600000_with_many_errors

De-duplication of resolution 0 (Raw-files) and resolution 300000 was done without warning or error messages.

# After De-duplication
After De-duplication was succeeded, the downsampling algorithm had starting to create new blocks for 300000 (A third block for 300000). I have no idea why that was happening. Maybe you or an other developer knows more. I interrupted this process and deleted all blocks of resolution 300000 and 3600000. After this I restarted compactor with downsampling again.

After downsampling overnight everything looks nearly perfect.
After_Dedup
After changing `thanos bucket web` config to `--label=prometheus` it looks perfect.
After_Dedup2

# Proposal to change from user perspective 
 - Checkout de-duplication algorithm for resolution 3600000.

homelessnessbo:
This would be just an awesome thing to have. Looking forward to it!

smalldirector:
Just came back from vacation. Thanks for review @povilasv, I will do the related code changes with comments. @Reamer Thanks for your testing, I will check your question.

cmd/thanos/compact.go
smalldirector: Ok, I can remove this function.

pkg/compact/compact.go
smalldirector: As this function is used in another package scope, I need to change it as public. I will add comment for this method.

pkg/compact/dedup/dedup.go
smalldirector: I will review all the codes and see whether I can change some exported structs / methods / functions to interval. If not, I will add comments for them. Thanks.

smalldirector:
@Reamer The logic for throwing the warning log `find empty downsample chunk` is pretty simple. The function tries to read chunk data with given downsampling aggr type, if no such chunk found, it will throw such logs. 
From your test, there is no `counter ` downsampling data for given label set. Can you write one simple block reader to test your downsampling block?

smalldirector:
@povilasv Done the changes with above comments, please do the next round review. Thanks.

Reamer:
Hi @smalldirector
I'm not a go developer, so I can't write any "simple" block reader.

smalldirector:
i see @Reamer, thanks again for trying this function.

GiedriusS:
Seems like a docs update is needed. Could we do that before the next round of reviews? (: @smalldirector

smalldirector:
@GiedriusS Fixed the doc issue.

bwplotka:
Nice, will review this week (:

smalldirector:
> Nice! I left some comments.
> 
> I think we need to address issues against deduplication algorithm first: #1788

Thanks for the comments @bwplotka . Agree, we should solve #1788 and #1394 firstly. Please let me know when the issues are resolved and I will do the related code changes.

Reamer:
Hi @smalldirector,
can I please rebase your commits with current master. I get the following compilation error.
```
go: extracting github.com/Azure/go-autorest/autorest/validation v0.2.1-0.20191028180845-3492b2aff503
github.com/thanos-io/thanos/pkg/compact/dedup imports
	github.com/prometheus/prometheus/tsdb/labels: module github.com/prometheus/prometheus@latest (v2.5.0+incompatible) found, but does not contain package github.com/prometheus/prometheus/tsdb/labels
make: *** [Makefile:132: deps] Error 1
The command '/bin/sh -c git update-index --refresh; make build' returned a non-zero code: 2
```
I merged your commits into my `dedup_test` branch which is based on the new tag 0.9.0.
https://github.com/Reamer/thanos/compare/v0.9.0...Reamer:dedup_test

Thanks for this feature

Reamer

smalldirector:
@Reamer Merged the latest codes and fixed the issues.

smalldirector:
Thanks for the suggestion. 
AFAIK, Thanos supports both of raw and downsampling data block, but there is no downsampling concept in Prometheus world. We may can use the function `chunks.MergeOverlappingChunks(chks) ` provided by Prometheus to do the merge logic, however it seems to be impossible to support `MergeOverlappingChunks(downsampling chks)` in Prometheus.
I don't think we can use this one function to handle both of raw and downsampling cases.

bwplotka:
> chunks.MergeOverlappingChunks(raw chks)

This should be fine as we hide downsampling chunks under common Prometheus chunk interface (:

smalldirector:
> > chunks.MergeOverlappingChunks(raw chks)
> 
> This should be fine as we hide downsampling chunks under common Prometheus chunk interface (:

Can anyone help make this change in Prometheus upstream?

bwplotka:
> Can anyone help make this change in Prometheus upstream?

In what sense help?  

I can help, but it's as easy as proposing a change to injecting interface/function for `chunks.MergeOverlappingChunks(raw chks)` to Compactor (: Happy to look on that tomorrow.

Reamer:
Hi,
I doesn't want, that this cool feature falls into sleep. @bwplotka Can we move forward here?

I get a merge conflict with [v0.10.1](https://github.com/thanos-io/thanos/releases/tag/v0.10.1). @smalldirector Can you please rebase your code to current master.

Thanks for your work

Reamer

bwplotka:
@Reamer @smalldirector We looked in details and we believe we can do 90% of this PR by just reusing existing components (with small upstream changes). Cc @metalmatze 

We will spend some time next week on this to show what we mean (: No work will be waste as grouping etc can be used from the code here!

vjsamuel:
@bwplotka curious, would you just add into this PR and add move the feature forward or would this PR be merged in and modified? we would like to see this feature in place as we see great value add. @smalldirector  thank you for continuing work on this.

bwplotka:
The feature is literally 100 lines of code if we would integrate all with TSDB code and reuse more, so I would say another PR. Are you ok with this @smalldirector ? Looks like Shuai is inactive as well recently.

Thank you for your hard work on this, I think we learned a lot with this implementation. Will close for now.

cc @metalmatze who is working on this, this week.

d-ulyanov:
waiting for this feature, @bwplotka is there another PR already?

bwplotka:
@metalmatze is working on the first iteration, I am preparing interface upstream, maybe as part of this: https://github.com/prometheus/prometheus/pull/5882

vjsamuel:
@bwplotka @metalmatze is there an update on this one?

Reamer:
https://github.com/prometheus/prometheus/pull/5882 -> https://github.com/prometheus/prometheus/pull/6990

bwplotka:
Yes, we are close to enable vertical compaction (which is a bit different than offline): https://github.com/thanos-io/thanos/pull/2250

Then for offline indeed it requires  prometheus/prometheus#6990 which is cloooooooose (:

PR #1277, opened 2019-06-25
docs: add integration docs

szalai1:
I am happy to add more information about the integration, but I believe if something is missing it  should be added to the original documentation of Geras.

## Changes

* add link to integrations.md from the main README.md
* start integrations.md doc

### DIFF
--- a/docs/integrations.md
+++ b/docs/integrations.md
@@ -7,11 +7,15 @@ slug: /integrations.md
 
 # Integrations 
 
-[StoreAPI](https://github.com/improbable-eng/thanos/blob/master/pkg/store/storepb/rpc.proto) is a common proto interface for gRPC component that can connect to [Querier](docs/components/query) in order to fetch the metric series. Natively Thanos implements [Sidecar](docs/components/sidecar) (local Prometheus data), [Ruler](docs/components/rule) and [Store gateway](docs/components/store). This solves fetching series from Prometheus or Prometheus TSDB format, however same interface can be used to fetch metrics from other storages. 
+## StoreAPI 
+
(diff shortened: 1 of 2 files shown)


docs/integrations.md
```diff
@@ -0,0 +1,9 @@
+# Integrations 
+
+## Geras (OpenTSDB integration)
```
bwplotka: Awesome. I would split this (even though it's one) in the sections. So let's add `## StoreAPI` then let's add something like

```
[StoreAPI](link to proto) is a common proto interface for gRPC component that can connect to [Querier](link to components/querier) in order to fetch the metric series. Natively Thanos implements [Sidecar](link to components/sidecar) (local Prometheus data), [Ruler](link to componenets/rule) and [Store gateway](link to components/store). This solves fetching series from Prometheus or Prometheus TSDB format, however same interface can be used to fetch metrics from other storages. 

Below you can find some public integrations with other systems through StoreAPI:

### OpenTSDB

(geras...) 

```
Can you be clear if it supports downsampled (aggregated chunks) or is it planned. It is totally fine if not (:

docs/integrations.md
```diff
@@ -0,0 +1,9 @@
+# Integrations
```
bwplotka: Can we add FontMatter as we have for other pages? This will expose this to website as well (: 

See https://thanos.io/contributing/how-to-contribute-to-docs.md/ for details (:

docs/integrations.md
szalai1: do you mean down sampled queries? Geras basically just translates the PromQL query to OpenTSDB query and passes the data. it's possible to aggregate the data on OpenTSDB level, but it's not supported at the moment.

docs/integrations.md
```diff
@@ -0,0 +1,17 @@
+---
+title: Integrations
+type: docs
+menu: thanos
+slug: /integrations.md
+---
+
+# Integrations 
+
```
bwplotka: Let's add `## StoreAPI` headline. There might be more types of integrations (:

bwplotka: Also we have broken links as found by CI: https://circleci.com/gh/improbable-eng/thanos/3979?utm_campaign=vcs-integration-link&utm_medium=referral&utm_source=github-build-link

docs/integrations.md
szalai1: mentioned it in the doc: 
| Although OpenTSDB is able to aggregte the data, it's not supported by Geras at the moment.

Issue #922, opened 2019-03-14
query: auto-downsampling causes inaccurate output of metrics (inflated values)


ottoyiu:
**Thanos, Prometheus and Golang version used**
```
image: improbable/thanos:v0.3.1
```
and 
```
image: improbable/thanos:v0.3.2
```
I was able to replicate it on both versions.

Prometheus 2.7.1

**What happened**
When `--query.auto-downsampling` is enabled on the query component, metrics beyond two days would be ballooned by multiples of the actual result. In this case, we've seen the metrics values go 10x.

PromQL: 
```
sum(dest_bps_in{hostname=~"$hostname", exported_namespace=~"$namespace"}) by (service_name, exported_namespace) * 8
```

Auto-downsampling enabled (grafana v5.3.4):
![Screenshot-2019-3-14 Grafana - (k8s) BalanceD Service At A Glance(1)](https://user-images.githubusercontent.com/2016437/54339527-53b0ab00-45f2-11e9-9070-1d5cb2ca7d34.png)

Auto-downsampling disabled (grafana v.5.3.4) - these metrics are accurate:
![Screenshot-2019-3-14 Grafana - (k8s) BalanceD Service At A Glance](https://user-images.githubusercontent.com/2016437/54339322-bbb2c180-45f1-11e9-869c-f77581363a17.png)

Another one with auto-downsampling enabled (grafana v6.0.1):
![Screenshot-2019-3-14 New dashboard - Grafana](https://user-images.githubusercontent.com/2016437/54340028-b22a5900-45f3-11e9-8c38-fb7427c0f586.png)

**What you expected to happen**
Metrics to be accurate irregardless of auto-downsampling is enabled or not.

**How to reproduce it (as minimally and precisely as possible)**:
```
        - --retention.resolution-raw=30d
        - --retention.resolution-5m=90d
        - --retention.resolution-1h=365d
```
on compactor

* Enable auto-downsampling, observe any metrics with 30 days window in Grafana. Metrics are inaccurate, and when zooming back in to a smaller window, the metrics become accurate again.
Disable auto-downsamping, observe any metrics with 30 days windows in Grafana. Metrics are accurate.

**Full logs to relevant components**
`thanos bucket inspect` output
<details>Logs
<p>

```
[...]

PR #1278, opened 2019-06-25
query/querier: fix sum() inflated values problem

GiedriusS:
Fixes the problem with `sum` and inflated values as outlined here: https://github.com/improbable-eng/thanos/issues/922.

The problem was the following: `sum` selects the last value of each time series in each window and adds the different dimensions up however our code asks for the downsampled sum value which is equal to all samples added up in either a 5m or 1h window, and adding those up, obviously, results in an inflated value. The practical result was that as if we applied `sum_over_time(...[5m])` on top. So the fix is to ask for the last sample in each window in the case of `sum` instead of the aggregated value.

Also adds an E2E test for a typical setup of one Sidecar connected + one or more
Thanos Store nodes. Tests and shows how the whole thing with `sum` really works.

Testing: wrote a query like `sum(kafka_log_log_value{topic="iam"})` with identical Sidecar/Store nodes and selected `Max 5m/1h downsampling`. With this fix the values look sane, without - not (it jumps up very high after the downsampled data comes into play).

Screenshot from 2019-06-26 10-43-51
Screenshot from 2019-06-26 10-44-01

### DIFF
--- a/pkg/query/iter.go
+++ b/pkg/query/iter.go
@@ -123,7 +123,7 @@ func (s *chunkSeries) Iterator() storage.SeriesIterator {
 		sit = newChunkSeriesIterator(its)
 	case resAggrSum:
 		for _, c := range s.chunks {
-			its = append(its, getFirstIterator(c.Raw, c.Sum))
+			its = append(its, getFirstIterator(c.Sum, c.Raw))
 		}
 		sit = newChunkSeriesIterator(its)
 	case resAggrMin:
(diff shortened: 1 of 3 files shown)


GiedriusS:
cc @jjneely. It's funny that we've arrived that the same conclusion even though I haven't looked at the commented that you've made in #922.

brancz:
Nice find. lgtm :+1:

jjneely:
Where is my approve button?  I want to hit it.  :-)  Thanks!

brancz:
I think we've reached sufficient consensus that this is correct. @bwplotka feel free to still review, but I'll go ahead and merge :)

PR #1280, opened 2019-06-26
Fixes #1272: fix map concurrent writes when rendering html templates in ui

### LINKED ISSUE #1272: thanos-query crashes with "concurrent map iteration and map write"

thanos, version 0.5.0 (branch: HEAD, revision: 72820b3f41794140403fd04d6da82299f2c16447)
  build user:       circleci@eeac5eb36061
  build date:       20190606-10:53:12
  go version:       go1.12.5

**What happened**

In one of k8s clusters that we run thanos-query in it crashes every couple of minutes with "fatal error: concurrent map iteration and map write" or "fatal error: concurrent map writes"

No crash :-)

I've no idea. I didn't manage to find anything that triggers it. Same problem was observed in 0.4.0. I'm not sure about 0.3.0.
thanos runs in GCP GKE cluster, query is deployed via our own helm chart. Crashing containers run:
```
  thanos query
      --log.level=debug
      --query.replica-label=prometheus_replica
      --grpc-server-tls-cert=/etc/certs/tls.crt
      --grpc-server-tls-key=/etc/certs/tls.key
      --store=dnssrv+_grpc._tcp.thanos-sidecars-prometheus.monitoring.svc
      --selector-label=location="REDACTED"
      --selector-label=stack="REDACTED"
      --selector-label=REDACTED
```

Same deployment (differs in selector-label values) crashes less in other GKE cluster and almost not at all in yet another GKE cluster, while receiving similar (very low) traffic via GRPC.

Those query instances serve as GRPC endpoints for global thanos-query (that runs in another, "observability" cluster and does not crash) to return recent data (older data is served from bucket). They are behind GCP load balancer (using http2 to communicate LB <-> thanos in GKE)

Example after-crash dump is here: https://gist.github.com/bjakubski/18a98f6f1fc2922e5056df3106fe1477

lx223:
## Changes
* use `template.Funcs` to overwrite prefix rendering function to avoid concurrent map writes

## Verification
Run `go run github.com/improbable-eng/thanos/cmd/thanos query`
And then concurrently HTTP GET "http://0.0.0.0:10902/graph" 1000 times

Before change:
It panics with concurrent map writes.
After change:
No longer panics.

### DIFF
--- a/pkg/ui/ui.go
+++ b/pkg/ui/ui.go
@@ -75,9 +75,9 @@ func (bu *BaseUI) executeTemplate(w http.ResponseWriter, name string, prefix str
 		return
 	}
 
-	bu.tmplFuncs["pathPrefix"] = func() string { return prefix }
-
-	t, err := template.New("").Funcs(bu.tmplFuncs).Parse(text)
+	t, err := template.New("").Funcs(bu.tmplFuncs).
+		Funcs(template.FuncMap{"pathPrefix": func() string { return prefix }}).
+		Parse(text)
 	if err != nil {
 		http.Error(w, err.Error(), http.StatusInternalServerError)
 		return


povilasv:
This makes sense, we were writing to map concurrently. Thanks for fixing this!

bwplotka:
That was quick @lx223 ! (:

bwplotka:
Since we probably copied this part from Prometheus I would check Prometheus codebase as well if they have same bug :thinking:

bwplotka:
Also we are missing CHANGELOG entry (: But it will be part of release shephard items to do.

lx223:
Ah, sorry. Should have added the entry in CHANGELOG. I'll put up another PR.

PR #1284, opened 2019-06-27
*: Allow exposing multiple label-sets

brancz:
## Changes 

This PR adds that the Info gRPC service now exposes potentially multiple label-sets. For all store types except `proxy` this doesn't change anything for now, it just adds the currently single label-set as the only entry in the list of label-sets. The `proxy` type, from now exposes it's stores' labels merged with its configured selector.

I'd like to ask that the `proxy.go` changes are particularly carefully reviewed.

## Verification

Adapted all tests and see them passing.

@squat @bwplotka @GiedriusS @povilasv @domgreen

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -19,7 +19,9 @@ We use *breaking* word for marking changes that are not backward compatible (rel
 
 - [#1248](https://github.com/improbable-eng/thanos/pull/1248) Add a web UI to show the state of remote storage.
 
-- [#1284](https://github.com/improbable-eng/thanos/pull/1284) Add support for multiple label-sets in Info gRPC service. This deprecates the single `Labels` slice of the `InfoResponse`.
+### Changed
+
+- [#1284](https://github.com/improbable-eng/thanos/pull/1284) Add support for multiple label-sets in Info gRPC service. This deprecates the single `Labels` slice of the `InfoResponse`. Old Thanos query components are compatible with new stores, but queriers must not be mixed versions.
 
 ## [v0.5.0](https://github.com/improbable-eng/thanos/releases/tag/v0.5.0) - 2019.06.05
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -19,6 +19,8 @@ We use *breaking* word for marking changes that are not backward compatible (rel
 
 - [#1248](https://github.com/improbable-eng/thanos/pull/1248) Add a web UI to show the state of remote storage.
 
+- [#1284](https://github.com/improbable-eng/thanos/pull/1284) Add support for multiple label-sets in Info gRPC service. This deprecates the single `Labels` slice of the `InfoResponse`.
+
 ## [v0.5.0](https://github.com/improbable-eng/thanos/releases/tag/v0.5.0) - 2019.06.05
 
 TL;DR: Store LRU cache is no longer leaking, Upgraded Thanos UI to Prometheus 2.9, Fixed auto-downsampling, Moved to Go 1.12.5 and more.
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -21,7 +21,7 @@ We use *breaking* word for marking changes that are not backward compatible (rel
 
 ### Changed
 
-- [#1284](https://github.com/improbable-eng/thanos/pull/1284) Add support for multiple label-sets in Info gRPC service. This deprecates the single `Labels` slice of the `InfoResponse`. Old Thanos query components are compatible with new stores, but queriers must not be mixed versions.
+- [#1284](https://github.com/improbable-eng/thanos/pull/1284) Add support for multiple label-sets in Info gRPC service. This deprecates the single `Labels` slice of the `InfoResponse`, in a future release backward compatible handling for the single set of Labels will be removed. Upgrading to v0.6.0 or higher is advised.
 
 ## [v0.5.0](https://github.com/improbable-eng/thanos/releases/tag/v0.5.0) - 2019.06.05
 
--- a/pkg/query/storeset.go
+++ b/pkg/query/storeset.go
@@ -33,7 +33,7 @@ type StoreSpec interface {
 	// If metadata call fails we assume that store is no longer accessible and we should not use it.
 	// NOTE: It is implementation responsibility to retry until context timeout, but a caller responsibility to manage
 	// given store connection.
-	Metadata(ctx context.Context, client storepb.StoreClient) (labelSets []storepb.LabelSet, labels []storepb.Label, mint int64, maxt int64, err error)
+	Metadata(ctx context.Context, client storepb.StoreClient) (labelSets []storepb.LabelSet, mint int64, maxt int64, err error)
 }
 
 type StoreStatus struct {
@@ -63,12 +63,16 @@ func (s *grpcStoreSpec) Addr() string {
 
 // Metadata method for gRPC store API tries to reach host Info method until context timeout. If we are unable to get metadata after
 // that time, we assume that the host is unhealthy and return error.
-func (s *grpcStoreSpec) Metadata(ctx context.Context, client storepb.StoreClient) (labelSets []storepb.LabelSet, labels []storepb.Label, mint int64, maxt int64, err error) {
+func (s *grpcStoreSpec) Metadata(ctx context.Context, client storepb.StoreClient) (labelSets []storepb.LabelSet, mint int64, maxt int64, err error) {
 	resp, err := client.Info(ctx, &storepb.InfoRequest{}, grpc.FailFast(false))
 	if err != nil {
-		return nil, nil, 0, 0, errors.Wrapf(err, "fetching store info from %s", s.addr)
+		return nil, 0, 0, errors.Wrapf(err, "fetching store info from %s", s.addr)
 	}
-	return resp.LabelSets, resp.Labels, resp.MinTime, resp.MaxTime, nil
+	if len(resp.LabelSets) == 0 && len(resp.Labels) > 0 {
+		resp.LabelSets = []storepb.LabelSet{{Labels: resp.Labels}}
+	}
+
+	return resp.LabelSets, resp.MinTime, resp.MaxTime, nil
 }
 
 // StoreSet maintains a set of active stores. It is backed up by Store Specifications that are dynamically fetched on
@@ -171,11 +175,7 @@ type storeRef struct {
 	logger log.Logger
 }
 
-func (s *storeRef) Update(labelSets []storepb.LabelSet, labels []storepb.Label, minTime int64, maxTime int64) {
-	if len(labelSets) == 0 && len(labels) > 0 {
-		labelSets = []storepb.LabelSet{{Labels: labels}}
-	}
-
+func (s *storeRef) Update(labelSets []storepb.LabelSet, minTime int64, maxTime int64) {
 	s.mtx.Lock()
 	defer s.mtx.Unlock()
 
@@ -296,14 +296,14 @@ func (s *StoreSet) getHealthyStores(ctx context.Context) map[string]*storeRef {
 			store, ok := s.stores[addr]
 			if ok {
 				// Check existing store. Is it healthy? What are current metadata?
-				labelSets, labels, minTime, maxTime, err := spec.Metadata(ctx, store.StoreClient)
+				labelSets, minTime, maxTime, err := spec.Metadata(ctx, store.StoreClient)
 				if err != nil {
 					// Peer unhealthy. Do not include in healthy stores.
 					s.updateStoreStatus(store, err)
 					level.Warn(s.logger).Log("msg", "update of store node failed", "err", err, "address", addr)
 					return
 				}
-				store.Update(labelSets, labels, minTime, maxTime)
+				store.Update(labelSets, minTime, maxTime)
 			} else {
 				// New store or was unhealthy and was removed in the past - create new one.
 				conn, err := grpc.DialContext(ctx, addr, s.dialOpts...)
@@ -322,8 +322,11 @@ func (s *StoreSet) getHealthyStores(ctx context.Context) map[string]*storeRef {
 					level.Warn(s.logger).Log("msg", "update of store node failed", "err", errors.Wrap(err, "initial store client info fetch"), "address", addr)
 					return
 				}
+				if len(resp.LabelSets) == 0 && len(resp.Labels) > 0 {
+					resp.LabelSets = []storepb.LabelSet{{Labels: resp.Labels}}
+				}
 				store.storeType = component.FromProto(resp.StoreType)
-				store.Update(resp.LabelSets, resp.Labels, resp.MinTime, resp.MaxTime)
+				store.Update(resp.LabelSets, resp.MinTime, resp.MaxTime)
 			}
 
 			mtx.Lock()
--- a/pkg/query/storeset.go
+++ b/pkg/query/storeset.go
@@ -33,7 +33,7 @@ type StoreSpec interface {
 	// If metadata call fails we assume that store is no longer accessible and we should not use it.
 	// NOTE: It is implementation responsibility to retry until context timeout, but a caller responsibility to manage
 	// given store connection.
-	Metadata(ctx context.Context, client storepb.StoreClient) (labelSets []storepb.LabelSet, mint int64, maxt int64, err error)
+	Metadata(ctx context.Context, client storepb.StoreClient) (labelSets []storepb.LabelSet, labels []storepb.Label, mint int64, maxt int64, err error)
 }
 
 type StoreStatus struct {
@@ -63,12 +63,12 @@ func (s *grpcStoreSpec) Addr() string {
 
 // Metadata method for gRPC store API tries to reach host Info method until context timeout. If we are unable to get metadata after
 // that time, we assume that the host is unhealthy and return error.
-func (s *grpcStoreSpec) Metadata(ctx context.Context, client storepb.StoreClient) (labelSets []storepb.LabelSet, mint int64, maxt int64, err error) {
+func (s *grpcStoreSpec) Metadata(ctx context.Context, client storepb.StoreClient) (labelSets []storepb.LabelSet, labels []storepb.Label, mint int64, maxt int64, err error) {
 	resp, err := client.Info(ctx, &storepb.InfoRequest{}, grpc.FailFast(false))
 	if err != nil {
-		return nil, 0, 0, errors.Wrapf(err, "fetching store info from %s", s.addr)
+		return nil, nil, 0, 0, errors.Wrapf(err, "fetching store info from %s", s.addr)
 	}
-	return resp.LabelSets, resp.MinTime, resp.MaxTime, nil
+	return resp.LabelSets, resp.Labels, resp.MinTime, resp.MaxTime, nil
 }
 
 // StoreSet maintains a set of active stores. It is backed up by Store Specifications that are dynamically fetched on
@@ -171,7 +171,11 @@ type storeRef struct {
 	logger log.Logger
 }
 
-func (s *storeRef) Update(labelSets []storepb.LabelSet, minTime int64, maxTime int64) {
+func (s *storeRef) Update(labelSets []storepb.LabelSet, labels []storepb.Label, minTime int64, maxTime int64) {
+	if len(labelSets) == 0 && len(labels) > 0 {
+		labelSets = []storepb.LabelSet{{Labels: labels}}
+	}
+
 	s.mtx.Lock()
 	defer s.mtx.Unlock()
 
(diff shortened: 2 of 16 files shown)


pkg/store/storepb/rpc.proto
```diff
@@ -42,9 +42,14 @@ enum StoreType {
 
 message InfoResponse {
   repeated Label labels = 1 [(gogoproto.nullable) = false];
```
povilasv: This one should be marked as deprecated no?

povilasv: something like `reserved 0;`

pkg/store/proxy.go
```diff
@@ -104,15 +123,69 @@ func (s *ProxyStore) Info(ctx context.Context, r *storepb.InfoRequest) (*storepb
...
-	for _, l := range s.selectorLabels {
-		res.Labels = append(res.Labels, storepb.Label{
-			Name:  l.Name,
-			Value: l.Value,
+	if len(stores) == 0 {
+		return res, nil
+	}
+
+	g, gctx := errgroup.WithContext(ctx)
+	respSender, respRecv, closeFn := newInfoRespCh(gctx, 10)
```
povilasv: I'm a bit worried about this part, whether it will actually work with all the time outs and partial response strategy.

So IMO more tests is needed.

Also maybe this could be simplified using mutex? Channels are a bit hard to deal with :/

brancz: I don't think there is a possibility for tolerating partial responses. We essentially need a consistent system for metadata like this, meaning either all succeed or none, and metadata stays unchanged until the next successful request.

Adding more tests is definitely a good idea, on that.

pkg/store/storepb/rpc.proto
```diff
@@ -42,9 +42,14 @@ enum StoreType {
...
-  int64 min_time        = 2;
-  int64 max_time        = 3;
-  StoreType storeType   = 4;
+  int64 min_time = 2;
```
bwplotka: Our manual pretty indent is gone ;p

not a blocker though.

pkg/store/storepb/rpc.proto
bwplotka: Let's comment as Deprecated

test/e2e/rule_test.go
```diff
@@ -241,6 +241,16 @@ func (a *failingStoreAPI) Info(context.Context, *storepb.InfoRequest) (*storepb.
 				Value: "store_api",
 			},
 		},
+		LabelSets: []storepb.LabelSet{
+			{
+				Labels: []storepb.Label{
```
bwplotka: Let's make test more ambitious and have multiple set here?

pkg/query/storeset.go
```diff
@@ -238,10 +239,12 @@ func (s *StoreSet) Update(ctx context.Context) {
...
 		//
 		// Sidecar will error out if it will be configured with empty external labels.
-		if len(store.Labels()) > 0 && externalLabelStores[externalLabelsFromStore(store)] != 1 {
+		externalLabels := externalLabelsFromStore(store)
+		storesWithExternalLabels := externalLabelStores[externalLabels]
+		if len(store.LabelSets()) > 0 && len(store.LabelSets()[0].Labels) > 0 && storesWithExternalLabels != 1 {
 			store.close()
 			s.updateStoreStatus(store, errors.New(droppingStoreMessage))
-			level.Warn(s.logger).Log("msg", droppingStoreMessage, "address", addr)
+			level.Warn(s.logger).Log("msg", droppingStoreMessage, "address", addr, "externalLabels", externalLabels, "duplicate-amount", storesWithExternalLabels)
```
bwplotka:
```suggestion
			level.Warn(s.logger).Log("msg", droppingStoreMessage, "address", addr, "extLset", externalLabels, "duplicates", storesWithExternalLabels)
```
I feel like short names are fine + let's be consistent on camelCase, snakeCase

pkg/query/storeset.go
```diff
@@ -238,10 +239,12 @@ func (s *StoreSet) Update(ctx context.Context) {
 		// No external labels means strictly store gateway or ruler and it is fine to have access to multiple instances of them.
 		//
 		// Sidecar will error out if it will be configured with empty external labels.
-		if len(store.Labels()) > 0 && externalLabelStores[externalLabelsFromStore(store)] != 1 {
+		externalLabels := externalLabelsFromStore(store)
+		storesWithExternalLabels := externalLabelStores[externalLabels]
+		if len(store.LabelSets()) > 0 && len(store.LabelSets()[0].Labels) > 0 && storesWithExternalLabels != 1 {
```
bwplotka: `len(store.LabelSets()[0].Labels) > 0 ` -> hm, what if `1` will have 0 elements? (:

I would skip this one. Let's make this natural, so we don't expect external labels to be empty? if empty the labelset is empty? What do you think?

pkg/store/tsdb.go
```diff
@@ -58,6 +58,14 @@ func (s *TSDBStore) Info(ctx context.Context, r *storepb.InfoRequest) (*storepb.
 			Value: l.Value,
 		})
 	}
+
+	// Until we deprecate the single labels in the reply, we just duplicate
+	// them here for migration/compatibility purposes.
+	res.LabelSets = []storepb.LabelSet{
+		{
+			Labels: res.Labels,
```
bwplotka: I would aim for invariant that if `res.Labels` are empty then `res.LabelSets` is `[]` - it will simplify interface a bit. WDYT?

pkg/query/storeset.go
```diff
@@ -331,16 +334,26 @@ func (s *StoreSet) getHealthyStores(ctx context.Context) map[string]*storeRef {
...
+				Name:  l.Name,
+				Value: l.Value,
+			})
+		}
+		sort.Sort(tsdbLabels)
+		tsdbLabelSetStrings = append(tsdbLabelSetStrings, tsdbLabels.String())
+	}
+	sort.Strings(tsdbLabelSetStrings)
+
+	s := ""
```
bwplotka: can we replace this with `return strings.Join(tsdbLabelSetStrings, ", )` or some other delimiter?

bwplotka: or just "" delimiter to have exactly same thing as you have here.

pkg/store/prometheus.go
```diff
@@ -98,6 +98,14 @@ func (p *PrometheusStore) Info(ctx context.Context, r *storepb.InfoRequest) (*st
 			Value: l.Value,
 		})
 	}
+
+	// Until we deprecate the single labels in the reply, we just duplicate
+	// them here for migration/compatibility purposes.
+	res.LabelSets = []storepb.LabelSet{
```
bwplotka: ditto as tsdb.go

pkg/store/proxy.go
```diff
@@ -104,15 +123,69 @@ func (s *ProxyStore) Info(ctx context.Context, r *storepb.InfoRequest) (*storepb
 	res.MaxTime = MaxTime
```
bwplotka: I know it is not yours, but can we fix above code? It looks weird that we loop over `range store` then check `len(stores)` 2 times. 

Also `MaxTime -> maxTime`, `MinTime ->  minTime`

pkg/store/proxy.go
```diff
@@ -104,15 +123,69 @@ func (s *ProxyStore) Info(ctx context.Context, r *storepb.InfoRequest) (*storepb
...
 	res.MinTime = MinTime
 
-	for _, l := range s.selectorLabels {
```
bwplotka: We need to remove selector flag as well then + CHANGELOG

pkg/store/tsdb.go
brancz: Yes agreed.

pkg/store/proxy.go
bwplotka: Hm.. Why we are doing this here? (: 

Querier already maintain healthy nodes in storeset. It does it by periodic `Info` calls. This always include labels as well. I think we can rely on this asyncronous Info calls here as well instead of duplicated, sync forwarding. What do you think? (:

I think some lag is fine with benefits of reduced number of requests + simplified code here.

pkg/store/proxy.go
```diff
@@ -74,6 +74,25 @@ func NewProxyStore(
 	return s
 }
 
+type ctxInfoRespSender struct {
+	ctx context.Context
+	ch  chan<- *storepb.InfoResponse
+}
+
+func newInfoRespCh(ctx context.Context, buffer int) (*ctxInfoRespSender, <-chan *storepb.InfoResponse, func()) {
```
bwplotka: might be not needed as commented above

pkg/store/proxy.go
```diff
@@ -193,7 +266,7 @@ func (s *ProxyStore) Series(r *storepb.SeriesRequest, srv storepb.Store_SeriesSe
 
 			sc, err := st.Series(seriesCtx, r)
 			if err != nil {
-				storeID := fmt.Sprintf("%v", storepb.LabelsToString(st.Labels()))
+				storeID := storepb.LabelSetsToString(st.LabelSets())
```
bwplotka: :+1:

pkg/store/proxy.go
```diff
@@ -388,18 +461,20 @@ func storeMatches(s Client, mint, maxt int64, matchers ...storepb.LabelMatcher)
 		return false, nil
 	}
 	for _, m := range matchers {
-		for _, l := range s.Labels() {
-			if l.Name != m.Name {
-				continue
-			}
+		for _, ls := range s.LabelSets() {
+			for _, l := range ls.Labels {
```
bwplotka: hm, something is wrong here - or if right we need to comment it better.

Essentially this means that matching here treat set with `AND` operator and I think we need `OR` rather, right?

as:
`[a=1, b=1][a=1, b=2]` labelset should be *matched* for `a=1, b=2` matcher, but it will be fitered out by `if !m.Matches(l.Value) {`

pkg/store/storepb/custom.go
```diff
@@ -177,3 +178,11 @@ func LabelsToString(lset []Label) string {
 	}
 	return "[" + strings.Join(s, ",") + "]"
 }
+
+func LabelSetsToString(lsets []LabelSet) string {
+	s := ""
+	for _, ls := range lsets {
```
bwplotka: I think convention in Thanos is to use `strings.Join` but here we might use `bytes.Buffer.WriteString` I guess? I think we don't care about performance though as it's only for printing. Not a blocker, just `+` for strings feels wrong in Golang although it's heavily optimized nowadays.

pkg/store/storepb/rpc.proto
bwplotka: :+1:

pkg/store/proxy.go
brancz: That's fair, will adapt.

pkg/query/storeset.go
brancz: :man_facepalming: you're absolutely right

pkg/query/storeset.go
brancz: Yes I was thinking something similar, and with the other invariant in tsdb.go and prometheus.go this works out nicely.

pkg/store/proxy.go
brancz: Yes you're right.

pkg/store/storepb/custom.go
brancz: Replaced with `strings.Join` based approached.

pkg/store/storepb/rpc.proto
brancz: I tried, but it started looking weirder and weirder. Let me know if it bothers you :)

pkg/store/storepb/rpc.proto
brancz: Done

test/e2e/rule_test.go
brancz: Done

pkg/store/tsdb.go
brancz: Done.

pkg/query/storeset.go
brancz: Done.

pkg/query/storeset.go
brancz: Done. It's much nicer this way.

pkg/query/storeset.go
brancz: Done.

pkg/store/prometheus.go
brancz: Done.

pkg/store/storepb/rpc.proto
brancz: Commented as deprecated.

pkg/store/proxy.go
brancz: Removed all the async code here and just assembled the label-sets with the cached store's label-sets.

pkg/query/storeset.go
```diff
@@ -215,6 +215,7 @@ func (s *StoreSet) Update(ctx context.Context) {
 	for _, st := range healthyStores {
 		externalLabelStores[externalLabelsFromStore(st)]++
 	}
+	level.Debug(s.logger).Log("externalLabelStores", fmt.Sprintf("%#+v", externalLabelStores))
```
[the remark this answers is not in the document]
brancz: Done.

pkg/store/proxy.go
brancz: Moved all `len(stores) == 0` to an early return before anything else.

pkg/store/proxy.go
brancz: Actually didn't remove this. If specified label-sets are merged with the selector labels specified.

pkg/store/proxy.go
brancz: Removed all of this async code.

pkg/store/proxy.go
brancz: Refactored this a bunch, I think this is much clearer now.

brancz:
I believe I have addressed all comments and rebased the branch to resolve conflicts with master. Let me know what you think :slightly_smiling_face: . Overall this is in a much better shape than the first time around I think! I'm feeling much more comfortable with it now. Thanks for the reviews!

CHANGELOG.md
```diff
@@ -19,6 +19,8 @@ We use *breaking* word for marking changes that are not backward compatible (rel
 
 - [#1248](https://github.com/improbable-eng/thanos/pull/1248) Add a web UI to show the state of remote storage.
 
+- [#1284](https://github.com/improbable-eng/thanos/pull/1284) Add support for multiple label-sets in Info gRPC service. This deprecates the single `Labels` slice of the `InfoResponse`.
```
bwplotka: Let's mention maybe what's the plan on deprecation. It looks like breaking change, while it is not. Worth to add details on that otherwise users will be like `Ok, but then all the components has to be upgraded to 0.6.0 if I use 0.6.0 Querier`?

pkg/store/proxy.go
```diff
@@ -81,38 +81,73 @@ func (s *ProxyStore) Info(ctx context.Context, r *storepb.InfoRequest) (*storepb
...
 
-	MinTime := int64(math.MaxInt64)
-	MaxTime := int64(0)
-
+	minTime := int64(math.MaxInt64)
+	maxTime := int64(0)
 	stores := s.stores()
+
+	if len(stores) == 0 {
+		res.MaxTime = math.MaxInt64
```
bwplotka: I am not sure about this. I know there is not regression to previous version (same logic), but does this make sense? ;p
Let's leave the comment maybe? `// Edge case: we have all of the data if there are no stores.` and add some `TODO`. 

Essentially not sure if we no store we should do either
* max time = MaxInt64 and min time = maxInt64 as nothing is there.
or
* max time = MaxInt64 and min Time MinInt64 if we want to have `whole range` ;p 

But we should not spend time on this in this PR I guess.

pkg/store/proxy.go
```diff
@@ -81,38 +81,73 @@ func (s *ProxyStore) Info(ctx context.Context, r *storepb.InfoRequest) (*storepb
...
-		MinTime = 0
-		MaxTime = math.MaxInt64
-	}
-
-	res.MaxTime = MaxTime
-	res.MinTime = MinTime
+	res.MaxTime = maxTime
+	res.MinTime = minTime
 
 	for _, l := range s.selectorLabels {
```
bwplotka: I assume we are not removing `selector Labels` for now?

pkg/store/proxy.go
```diff
@@ -81,38 +81,73 @@ func (s *ProxyStore) Info(ctx context.Context, r *storepb.InfoRequest) (*storepb
...
+			mergedLabelSet := mergeLabels(labelSet.Labels, s.selectorLabels)
+			ls := storepb.LabelsToPromLabels(mergedLabelSet)
+			sort.Sort(ls)
+			labelSets[ls.Hash()] = mergedLabelSet
+		}
+	}
+
+	res.LabelSets = make([]storepb.LabelSet, 0, len(labelSets))
+	for _, v := range labelSets {
+		res.LabelSets = append(res.LabelSets, storepb.LabelSet{Labels: v})
```
bwplotka: Worth to mention on interface that sets are not sorted?

pkg/store/proxy.go
```diff
@@ -81,38 +81,73 @@ func (s *ProxyStore) Info(ctx context.Context, r *storepb.InfoRequest) (*storepb
...
+	}
+
 	return res, nil
 }
 
+// mergeLabels merges label-sets a and b with label-set b having precedence.
+func mergeLabels(labelSet []storepb.Label, selector labels.Labels) []storepb.Label {
```
bwplotka: I guess `a []storepb.Label, b labels.Labels)` would makes sense then more by looking on comment?

pkg/store/proxy.go
```diff
@@ -381,14 +416,39 @@ func (s *streamSeriesSet) Err() error {
...
+// matchers.
 func storeMatches(s Client, mint, maxt int64, matchers ...storepb.LabelMatcher) (bool, error) {
 	storeMinTime, storeMaxTime := s.TimeRange()
 	if mint > storeMaxTime || maxt < storeMinTime {
 		return false, nil
 	}
+	return labelSetsMatch(s.LabelSets(), matchers)
+}
+
+// labelSetsMatch returns false if each label-set negative matches against the
```
bwplotka: I would change this to (find it more clear for me), but up to you (: 
```suggestion
// labelSetsMatch returns false if all label-set does not match the
```

pkg/store/storepb/rpc.proto
bwplotka: Nah, I think your time is way more precious to waste it on  manual linting (:

pkg/store/proxy.go
brancz: Would an issue be more appropriate than a TODO in code maybe?

brancz: I'm adding the "Edge case:" comment back and will open an issue to decide this as a follow up.

pkg/store/proxy.go
brancz: Added to the proto message and generated structs.

pkg/store/proxy.go
brancz: It somewhat depends on our decision around the changelog. If we decide all queriers must be >=v0.6.0 if any are, then this would be safe to remove.

pkg/query/storeset.go
```diff
@@ -32,14 +33,14 @@ type StoreSpec interface {
 	// If metadata call fails we assume that store is no longer accessible and we should not use it.
 	// NOTE: It is implementation responsibility to retry until context timeout, but a caller responsibility to manage
 	// given store connection.
-	Metadata(ctx context.Context, client storepb.StoreClient) (labels []storepb.Label, mint int64, maxt int64, err error)
+	Metadata(ctx context.Context, client storepb.StoreClient) (labelSets []storepb.LabelSet, labels []storepb.Label, mint int64, maxt int64, err error)
```
bwplotka: I don't really get why not just replacing with labelSets.

We can hide the compatibility logic internally, right?

pkg/query/storeset.go
```diff
@@ -162,27 +163,31 @@ type storeRef struct {
...
 	maxTime   int64
 
 	logger log.Logger
 }
 
-func (s *storeRef) Update(labels []storepb.Label, minTime int64, maxTime int64) {
-	s.mtx.RLock()
-	defer s.mtx.RUnlock()
+func (s *storeRef) Update(labelSets []storepb.LabelSet, labels []storepb.Label, minTime int64, maxTime int64) {
+	if len(labelSets) == 0 && len(labels) > 0 {
```
bwplotka: I would hide it behind Metadata interface.

pkg/store/proxy.go
bwplotka: Thanks, makes sense

pkg/store/proxy.go
bwplotka: :+1:  Let's leave for now

bwplotka:
Awesome! :ship: it.

Thanks :+1:

PR #1287, opened 2019-06-28
sidecar:  wait on ext labels before uploading

mjd95:
Addresses https://github.com/improbable-eng/thanos/issues/1239 by adding a step that will force a wait on external labels before starting the shipper (when upload is enabled).

Note that the case of `len(m.Labels())` is already checked in the sidecar when it initially syncs with Prom.  However, we don't currently hard-fail if `len(m.Labels())==0`as we retry indefinitely.  With this PR, we would now hard-fail in that case (after 2 minutes).

### DIFF
--- a/cmd/thanos/sidecar.go
+++ b/cmd/thanos/sidecar.go
@@ -248,7 +248,7 @@ func runSidecar(
 		g.Add(func() error {
 			defer runutil.CloseWithLogOnErr(logger, bkt, "bucket client")
 
-			extLabelsTimeout := 2 * time.Minute
+			extLabelsTimeout := 10 * time.Minute
 			extLabelsCtx, _ := context.WithTimeout(ctx, extLabelsTimeout)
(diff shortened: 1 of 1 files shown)


cmd/thanos/sidecar.go
```diff
@@ -248,6 +248,17 @@ func runSidecar(
 		g.Add(func() error {
 			defer runutil.CloseWithLogOnErr(logger, bkt, "bucket client")
 
+			extLabelsTimeout := 2 * time.Minute
```
bwplotka: Not sure if 2 min is ok, maybe worth to increase or expose via flag?

mjd95: Yeah fair point.  I've increased it, quite a lot actually to 10 minutes.  We should probably be conservative here (don't want to kill anything we shouldn't) and I'd rather avoid a flag if possible.  Feel free to adjust the value though, I might be missing some context on startup times :)

PR #1289, opened 2019-06-28
Added docs about new bucket web command.

bwplotka:

@jaseemabid


bwplotka:
Good point @FUSAKLA added extra info

PR #1297, opened 2019-07-02
feat compact: added readiness Prober

FUSAKLA:
## Changes
Used new Prober package in the Thanos compact to provide `/-/healthy` and `/-/ready` endpoints

I also added new `defatltHTTPListener` which registers the prober. I'm leaving the old `metricHTTPListenGroup` with `TODO` to remove it once all components are migrated to the new one to avoid touching different components in this PR.

Also I switched to using `component.Component` instead of passing the string from the main. 
It felt weird to have somewhere string and somewhere the component interface.

## Verification
Tested it and works as expected, also exposes metrics about Prober status 
```
# HELP thanos_prober_healthy Represents health status of the component Prober.
# TYPE thanos_prober_healthy gauge
thanos_prober_healthy{component="compact"} 1
# HELP thanos_prober_ready Represents readiness status of the component Prober.
# TYPE thanos_prober_ready gauge
thanos_prober_ready{component="compact"} 1
```

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -18,6 +18,7 @@ We use *breaking* word for marking changes that are not backward compatible (rel
 ### Changed
 
 - [#1338](https://github.com/thanos-io/thanos/pull/1338) Querier still warns on store API duplicate, but allows a single one from duplicated set. This is gracefully warn about the problematic logic and not disrupt immediately.
+- [#1297](https://github.com/improbable-eng/thanos/pull/1297) Added `/-/ready` and `/-/healthy` endpoints to Thanos compact.
 
 ### Fixed
 
--- a/cmd/thanos/compact.go
+++ b/cmd/thanos/compact.go
@@ -15,7 +15,7 @@ import (
 	"github.com/go-kit/kit/log"
 	"github.com/go-kit/kit/log/level"
 	"github.com/oklog/run"
-	opentracing "github.com/opentracing/opentracing-go"
+	"github.com/opentracing/opentracing-go"
 	"github.com/pkg/errors"
 	"github.com/prometheus/client_golang/prometheus"
 	"github.com/prometheus/tsdb"
@@ -23,10 +23,12 @@ import (
 	"github.com/thanos-io/thanos/pkg/block/metadata"
 	"github.com/thanos-io/thanos/pkg/compact"
 	"github.com/thanos-io/thanos/pkg/compact/downsample"
+	"github.com/thanos-io/thanos/pkg/component"
 	"github.com/thanos-io/thanos/pkg/objstore"
 	"github.com/thanos-io/thanos/pkg/objstore/client"
+	"github.com/thanos-io/thanos/pkg/prober"
 	"github.com/thanos-io/thanos/pkg/runutil"
-	kingpin "gopkg.in/alecthomas/kingpin.v2"
+	"gopkg.in/alecthomas/kingpin.v2"
 )
 
 var (
@@ -49,7 +51,7 @@ func (cs compactionSet) String() string {
 	return strings.Join(result, ", ")
 }
 
-// levels returns set of compaction levels not higher than specified max compaction level
+// levels returns set of compaction levels not higher than specified max compaction level.
 func (cs compactionSet) levels(maxLevel int) ([]int64, error) {
 	if maxLevel >= len(cs) {
 		return nil, errors.Errorf("level is bigger then default set of %d", len(cs))
@@ -62,13 +64,14 @@ func (cs compactionSet) levels(maxLevel int) ([]int64, error) {
 	return levels, nil
 }
 
-// maxLevel returns max available compaction level
+// maxLevel returns max available compaction level.
 func (cs compactionSet) maxLevel() int {
 	return len(cs) - 1
 }
 
-func registerCompact(m map[string]setupFunc, app *kingpin.Application, name string) {
-	cmd := app.Command(name, "continuously compacts blocks in an object store bucket")
+func registerCompact(m map[string]setupFunc, app *kingpin.Application) {
+	comp := component.Compact
+	cmd := app.Command(comp.String(), "continuously compacts blocks in an object store bucket")
 
 	haltOnError := cmd.Flag("debug.halt-on-error", "Halt the process if a critical compaction error is detected.").
 		Hidden().Default("true").Bool()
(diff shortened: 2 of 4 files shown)


FUSAKLA:
@bwplotka 
So first one of the series of Prober additions :tada: 

Formerly we discussed:
- Naming of the `defaultHTTPListener`
- If the default listener should
  - set healthy
  - set ready

I chose compact as a first one since it should be most straight forward :)

CHANGELOG.md
```diff
@@ -19,6 +19,8 @@ We use *breaking* word for marking changes that are not backward compatible (rel
 
 - [#1248](https://github.com/improbable-eng/thanos/pull/1248) Add a web UI to show the state of remote storage.
 
+- [#1297](https://github.com/improbable-eng/thanos/pull/1297) Added `/-/ready` and `/-healthy` endpoints to Thanos compact.
```
GiedriusS: It should be `/-/healthy` here

cmd/thanos/main.go
```diff
@@ -73,7 +75,7 @@ func main() {
 	registerStore(cmds, app, "store")
 	registerQuery(cmds, app, "query")
 	registerRule(cmds, app, "rule")
-	registerCompact(cmds, app, "compact")
+	registerCompact(cmds, app, component.Compact)
```
GiedriusS: I know that you have not introduced this parameter but I feel that in the long term this parameter should go away and we should just use `component.Compact` inside of the function itself since it is already designated for registering the `compact` command :smile:

GiedriusS:
Oh, and could we update `tutorials/kubernetes-demo/manifests/thanos-compactor.yaml` so that it would use these new end-points according to this https://kubernetes.io/docs/tasks/configure-pod-container/configure-liveness-readiness-probes/? :heart_eyes:

cmd/thanos/main.go
FUSAKLA: Sounds good to me, I thought about it too but I wasn't sure even with this change.
If we agree I'll be happy to change it. PR like this will come for every component so eventually I can change it for all of them.

CHANGELOG.md
FUSAKLA: thanks!

FUSAKLA:
@GiedriusS I knew I forgot about something, thanks for reminding me! :smile: Ill modify those manifests too.

cmd/thanos/main.go
FUSAKLA: changed in https://github.com/improbable-eng/thanos/pull/1297/commits/127ee71d2a18dd59d7f1606ceceb943db969c094 is this ok with you?

FUSAKLA:
@GiedriusS manifests should be fixed now PTAL

cmd/thanos/main.go
```diff
@@ -326,3 +328,27 @@ func metricHTTPListenGroup(g *run.Group, logger log.Logger, reg *prometheus.Regi
...
+	registerProfile(mux)
+	readinessProber.RegisterInMux(mux)
+
+	l, err := net.Listen("tcp", httpBindAddr)
+	if err != nil {
+		return errors.Wrap(err, "listen metrics address")
+	}
+
+	g.Add(func() error {
+		level.Info(logger).Log("msg", "Listening for metrics", "address", httpBindAddr)
```
bwplotka: Not yours but I think we aimed for lower case log messages so 
```suggestion
		level.Info(logger).Log("msg", "listening for metrics", "address", httpBindAddr)
```

cmd/thanos/compact.go
```diff
@@ -62,13 +64,14 @@ func (cs compactionSet) levels(maxLevel int) ([]int64, error) {
...
-// maxLevel returns max available compaction level
+// maxLevel returns max available compaction level.
 func (cs compactionSet) maxLevel() int {
 	return len(cs) - 1
 }
 
-func registerCompact(m map[string]setupFunc, app *kingpin.Application, name string) {
-	cmd := app.Command(name, "continuously compacts blocks in an object store bucket")
+func registerCompact(m map[string]setupFunc, app *kingpin.Application) {
+	component := component.Compact
```
bwplotka: `const component = component.Compact`?

FUSAKLA: hm it looks like struct or any other more complex data structure cannot be a constant IIUC?
I renamed the variable at least to `comp` so it does not collide with the package name.

cmd/thanos/main.go
FUSAKLA: yeah, changed it

FUSAKLA:
thanks @bwplotka for the cr!

FUSAKLA:
Yes, the healthy status is set by the `defaultHTTPListener` right from the time when the `/-/healthy` endpoint is available which happen to be the same time we start to listen for metrics. 

The readiness is left there just so it is somewhere. As you said in case of compactor it does not matter. 

IIUC you would move setting healthiness to the same place as readiness? IMHO this would not change anything since the server is not listening yet so setting healthy before it starts has no meaning and just would require to take out the generic liveness setting from the `defaultHTTPListener`. But maybe I'm missing out something?

povilasv:
I think we should just remove readiness, it doesn't make sense in compactor and we shouldn't expose it. Users should just use healthy. WDYT?

FUSAKLA:
TBH I'd rather leave it there. All other components will hopefully have it and user could be confused that it's missing. I'd just say it's ready right from the start? 

But if you prefer removing it I'll make it disappear :)

midnightconman:
> TBH I'd rather leave it there. All other components will hopefully have it and user could be confused that it's missing. I'd just say it's ready right from the start?
> 
> But if you prefer removing it I'll make it disappear :)

I agree let's leave it in. This makes helm / deployment logic simpler. Standards reduce cognitive load :-)

FUSAKLA:
@GiedriusS should be rebased now with all the conflicts resolved and changelog entry adjusted as well. Sorry for the delay

FUSAKLA:
hm.. weird. retriggered the CI and it passed and before it failed on Azure tests.
seems to be somehow flaky :/

GiedriusS:
LGTM, @bwplotka could you make the final call here in light of comments by Povilas?

povilasv:
@GiedriusS I agree let's reduce cognitive load and keep it ;)

povilasv:
Regarding 
> IIUC you would move setting healthiness to the same place as readiness? IMHO this would not change anything since the server is not listening yet so setting healthy before it starts has no meaning and just would require to take out the generic liveness setting from the defaultHTTPListener. But maybe I'm missing out something?

I would still set readiness and healthiness in the same place if possible, to reduce the cognitive burden on developers. But other than that LGTM

FUSAKLA:
Great, thanks for the review and merging this!

I'll start with PR for next thanos component to add the prober

PR #1302, opened 2019-07-03
runutil: add Exhaust* fns, initial users

GiedriusS:
## Changes

* `runutil` package now has an extra family of functions which exhaust an `io.ReadCloser` before closing them
* Internal Thanos code has been converted to use them where it makes sense i.e. `resp.Body` is used a.k.a. HTTP connections

I think that the old functions can still be there because we might have external users of this package and in some cases it just doesn't make sense to use them.

## Rationale

Lets us reuse keep-alive connections efficiently. Ticket: https://github.com/improbable-eng/thanos/issues/1206


GiedriusS:
@bwplotka do you agree with this direction? If yes, then I will convert the other users where this is applicable.

pkg/runutil/runutil.go
```diff
@@ -108,6 +115,12 @@ func CloseWithLogOnErr(logger log.Logger, closer io.Closer, format string, a ...
 	level.Warn(logger).Log("msg", "detected close error", "err", errors.Wrap(err, fmt.Sprintf(format, a...)))
 }
 
+// ExhaustCloseWithLogOnErr closes the io.ReadCloser with a log message on error but exhausts the reader before.
+func ExhaustCloseWithLogOnErr(logger log.Logger, readcloser io.ReadCloser, format string, a ...interface{}) {
+	var _, _ = io.Copy(ioutil.Discard, readcloser) // best effort
```
bwplotka: No needed comment IMO. also kill `var`
```suggestion
	 _, _ = io.Copy(ioutil.Discard, readCloser) 
```

pkg/runutil/runutil.go
```diff
@@ -108,6 +115,12 @@ func CloseWithLogOnErr(logger log.Logger, closer io.Closer, format string, a ...
 	level.Warn(logger).Log("msg", "detected close error", "err", errors.Wrap(err, fmt.Sprintf(format, a...)))
 }
 
+// ExhaustCloseWithLogOnErr closes the io.ReadCloser with a log message on error but exhausts the reader before.
+func ExhaustCloseWithLogOnErr(logger log.Logger, readcloser io.ReadCloser, format string, a ...interface{}) {
```
bwplotka:
```suggestion
func ExhaustCloseWithLogOnErr(logger log.Logger, r io.ReadCloser, format string, a ...interface{}) {
```
maybe?

GiedriusS: Renamed.

pkg/runutil/runutil.go
GiedriusS: Killed these :)

GiedriusS:
@bwplotka PTAL again when you'll have some time before merging. Thanks! :smile:

pkg/runutil/runutil.go
```diff
@@ -108,6 +115,12 @@ func CloseWithLogOnErr(logger log.Logger, closer io.Closer, format string, a ...
 	level.Warn(logger).Log("msg", "detected close error", "err", errors.Wrap(err, fmt.Sprintf(format, a...)))
 }
 
+// ExhaustCloseWithLogOnErr closes the io.ReadCloser with a log message on error but exhausts the reader before.
+func ExhaustCloseWithLogOnErr(logger log.Logger, r io.ReadCloser, format string, a ...interface{}) {
+	_, _ = io.Copy(ioutil.Discard, r)
```
bwplotka: Technically log on error might mean checking err here as well (:  But I think it's fine for now.

GiedriusS:
> I think we could potentially reduce number of methods here e.g by having just `Close` that accepts `io.Closer` but tries to cast to `io.Reader` and then optionally Copies? I think it is solid behavior. (:
> 
> What do you think? (:

It is a thing that we could do but the question here is do we want to force that behavior onto the external users of this package? IMHO it's not always useful to discard all of the input because you could never be sure what is actually in the `Read()` method of `io.Reader`.

pkg/runutil/runutil.go
GiedriusS: That's true but such an error message would be harmless and users would just ignore it, no?

bwplotka: No if you want to exhaust and that failed, right?

bwplotka:
True you are right. (:

PR #1306, opened 2019-07-04
prober: remove unused componentMtx

GiedriusS:
AFAICT there was an idea to use a general RWMutex componentMtx but then
it was removed in favor of two separate RWMutexes for the readiness and
healthiness probes.

Remove it from the struct to unblock the build.


FUSAKLA:
It was originaly used by getter and setter but those were dropped because it wont ever change so it wasn't needed to lock it and apparently I forgot to remove the unused mtx. Sorry for the trouble :/

PR #1310, opened 2019-07-05
Make it clear, explicit, and fail early on query with just external labels.

bwplotka:
Addresses https://github.com/improbable-eng/thanos/issues/1305


GiedriusS:
@bwplotka was this intentional that you've merged this into `master-build-fix`, not `master`? :smile:

PR #1314, opened 2019-07-09
.*: update prometheus client, remove deprecated InstrumentHandler and InstrumentHandlerFunc methods

### LINKED ISSUE #1304: Move away from deprecated InstrumentHandler{,Func} and update to the newest version of client_golang
Those two methods were deprecated in client_golang and recently completely removed. Move away from using them and also update to the newest client_golang version.

kakkoyun:
Updates `prometheus/client_golang`. Remove usages of deprecated methods by introducing wrapper methods which mimics behaviour the same behaviour using new APIs.

Fixes #1304

## Changes

* Updates `prometheus/client_golang`.
* Removes usages of `prometheus.InstrumentHandler` and `prometheus. InstrumentHandlerFunc`
* Introduces `extprom/NewInstrumentedHandler` and `extprom/NewInstrumentedHandlerFunc`.
* Removes `http_request_duration_microseconds` (Summary).
* Adds `http_request_duration_seconds` (Histogram).

New metrics:
```
# HELP http_request_duration_seconds Tracks the latencies for HTTP requests.
# TYPE http_request_duration_seconds histogram
# HELP http_request_size_bytes Tracks the size of HTTP requests.
# TYPE http_request_size_bytes summary
# HELP http_requests_total Tracks the number of HTTP requests.
# TYPE http_requests_total counter
# HELP http_response_size_bytes Tracks the size of HTTP responses.
# TYPE http_response_size_bytes summary
```

Old metrics:
```
# HELP http_request_duration_microseconds The HTTP request latencies in microseconds.
# TYPE http_request_duration_microseconds summary
# HELP http_request_size_bytes The HTTP request sizes in bytes.
# TYPE http_request_size_bytes summary
# HELP http_requests_total Total number of HTTP requests made.
# TYPE http_requests_total counter
# HELP http_response_size_bytes The HTTP response sizes in bytes.
# TYPE http_response_size_bytes summary
```

## Verification

By using `make test` and manual testing against query API, it seems like previous behaviour kept intact.


CHANGELOG.md
```diff
@@ -23,6 +23,8 @@ We use *breaking* word for marking changes that are not backward compatible (rel
 
 - [#1284](https://github.com/improbable-eng/thanos/pull/1284) Add support for multiple label-sets in Info gRPC service. This deprecates the single `Labels` slice of the `InfoResponse`, in a future release backward compatible handling for the single set of Labels will be removed. Upgrading to v0.6.0 or higher is advised.
 
+- [#1314](https://github.com/improbable-eng/thanos/pull/1314) Updates `prometheus/client_golang`. Remove usages of [deprecated methods](https://github.com/prometheus/client_golang/blob/505eaef017263e299324067d40ca2c48f6a2cf50/prometheus/http.go#L122) by introducing a wrapper package which mimics the same behaviour while using new APIs and fixes stated issues. As a result, removes `http_request_duration_microseconds` (Summary) and adds `http_request_duration_seconds` (Histogram).
```
bwplotka: I would focus on user facing issues, so literaly metrics changes. From user standpoint in does not matter what client library we use (: Can we trim this?

pkg/extprom/http/instrument_server.go
```diff
@@ -0,0 +1,107 @@
...
+}
+
+type serverInstrumentor struct {
+	requestDuration *prometheus.HistogramVec
+	requestSize     *prometheus.SummaryVec
+	requestsTotal   *prometheus.CounterVec
+	responseSize    *prometheus.SummaryVec
+}
+
+// NewServerInstrumentor provides default ServerInstrumentor
```
bwplotka: missing trailing period here and in other places.

pkg/extprom/http/instrument_server.go
```diff
@@ -0,0 +1,107 @@
...
+	"github.com/prometheus/client_golang/prometheus/promhttp"
+)
+
+// ServerInstrumentor holds necessary metrics to instrument an http.Server
+// and provides necessary behaviors
+type ServerInstrumentor interface {
+	// NewInstrumentedHandler wraps the given HTTP handler for instrumentation.
+	NewInstrumentedHandler(handlerName string, handler http.Handler) http.HandlerFunc
+	// NewInstrumentedHandlerFunc wraps the given function for instrumentation.
+	NewInstrumentedHandlerFunc(handlerName string, next func(http.ResponseWriter, *http.Request)) http.HandlerFunc
```
bwplotka: I would vote for simpler interface and removal of one of those methods. This one is to shallow IMO and there are many easy ways to tranform `handler` to `next func` and vice versa. (: 

Let's keep one way of doing things.

pkg/extprom/http/instrument_server.go
```diff
@@ -0,0 +1,107 @@
...
+import (
+	"net/http"
+
+	"github.com/prometheus/client_golang/prometheus"
+	"github.com/prometheus/client_golang/prometheus/promhttp"
+)
+
+// ServerInstrumentor holds necessary metrics to instrument an http.Server
+// and provides necessary behaviors
+type ServerInstrumentor interface {
```
bwplotka: Also not sure of `Instrumentor` name.. Isn't `Middleware` bit better and clearer? Not a blocker, might be by personal preference, up to you.

CHANGELOG.md
kakkoyun: Of course, I'll trim it.

pkg/extprom/http/instrument_server.go
GiedriusS: Maybe I am wrong but I think that if we would name these middlewares then there might be more confusion in the future when (if) we will switch to using Cortex's middlewares for caching.

kakkoyun: I wasn't also sure about `Instrumentor` name, it's not even an actual word. I'm ok to use `Middleware`.

@GiedriusS Since it will be under a particular namespace with package structure, maybe it won't create any confusion. What do you think?

kakkoyun: @bwplotka How about `InstrumentationMiddleware`? It'll go like `extpromhttp.NewInstrumentationMiddleware`, `NewInstrumentationMiddleware().NewHandler()` etc.

brancz:
InstrumentationMiddleware sounds good to me. As the package and context are revolving around http, I think it's pretty clear, and doesn't cause confusion.

pkg/extprom/http/instrument_server.go
bwplotka: Again, don't have strong opinion (: Happy with `Instrumentor`

bwplotka: > Maybe I am wrong but I think that if we would name these middlewares then there might be more confusion in the future when (if) we will switch to using Cortex's middlewares for caching.

Well that will be exactly the same middleware isn't it?

```
 ... handler http.Handler) http.HandlerFunc { ...
```

bwplotka:
Thanks! LGTM

PR #1311, opened 2019-07-08
compact/store: update Azure dependency to fix panics

wbh1:
Closes #1207
## Changes

This bumps the Azure Blob (cloud storage) library dependency to the latest version, which includes a fix I added to resolve panics occasionally encountered in Compact and Store components (https://github.com/Azure/azure-storage-blob-go/pull/135). 

## Verification

Verified compaction works and verified that buckets can be enumerated with commands like `thanos bucket web` and `thanos bucket ls`

PR #1315, opened 2019-07-09
add CHANGELOG entry for PR #1311

wbh1:
## Changes

Adds an entry to the CHANGELOG for the fix added in #1311, since I overlooked that initially :sweat_smile: 

## Verification

N/A

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -11,6 +11,9 @@ We use *breaking* word for marking changes that are not backward compatible (rel
 
 ## Unreleased.
 
+### Fixed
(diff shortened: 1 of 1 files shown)


wbh1:
@povilasv :smile:

wbh1:
It could be many different things causing the panics, honestly. Network connectivity, Azure availability, etc. It was caused by the Azure library having a part of a function that wasn't checking the value of `err` before proceeding to its next step.

I could clarify that the panics would happen while downloading the "blob" to a buffer, if that helps :)

bwplotka:
Ah. so let's write then `not checked errors` to be explicit (:

wbh1:
Fixed it. 

PS - congrats on CNCF acceptance :tada:

PR #1321, opened 2019-07-11
Make it clear, explicit, and fail early on query with just external labels

GiedriusS:
Merging it into `master`. I think @bwplotka accidentally merged it into a wrong branch which has a similar name to `master`. Original PR: https://github.com/improbable-eng/thanos/pull/1310


bwplotka:
Ups! Chained PR indeed, thanks for doin it properly.

PR #1327, opened 2019-07-13
api/series: return an empty array iff no results

### LINKED ISSUE #1325: query: Series API returns `null` data
Related to, but not quite,  #576

* Thanos `improbable/thanos:v0.5.0`
* Prometheus `prometheus/prometheus:v2.10.0`

**What happened?**

I queried the `series` api for a metric that doesn't exist and the data field of the response was `null`:

```
# curl http://thanos-query.monitoring.svc.cluster.local:9090/api/v1/series?match%5B%5D=no_exist | jq .
{
  "status": "success",
  "data": null
}
```

I expected it to return an empty array like the Prometheus `series` api does:

```
# curl http://prometheus-k8s-a.monitoring.svc.cluster.local:9090/api/v1/series?match%5B%5D=no_exist | jq .
{
  "status": "success",
  "data": []
}
```

Query the `series` api for a metric that does not exist.

We came across this issue debugging the prometheus-adapter.  We had recently switched the endpoint for it from Prometheus to the Thanos Querier.  It began hitting this [error](https://github.com/DirectXMan12/k8s-prometheus-adapter/blob/3e6d59813bb6135cf34fd5ad477850317dd9b24a/pkg/custom-provider/provider.go#L267).  Presumably because Thanos is returning `null` instead of an empty array.

A little hunting and I found that the series function for the api can in fact return `nil` if there are no series (see [here](https://github.com/improbable-eng/thanos/blob/7b36de5d19ec594f64b10a47f0bd85594a52c814/pkg/query/api/v1.go#L546-L554)).  The variable is never initialized as opposed to the Prometheus code [here](https://github.com/prometheus/prometheus/blob/d20e84d0fb64aff2f62a977adc8cfb656da4e286/web/api/v1/api.go#L503).  I can't however understand how it's making it back out in the [response](https://github.com/improbable-eng/thanos/blob/7b36de5d19ec594f64b10a47f0bd85594a52c814/pkg/query/api/v1.go#L158-L164).

This is probably not a big deal for most since usually you'll be querying for metrics that exist.  But operating on a multi-tenant cluster and can't guarantee that the metrics we're using for the adapter will always be there.  We're going to try a recording rule to fake the metric so we can continue with the switch to Thanos as a short term fix.  I'd be happy to do a pull request but like I said couldn't find how it's getting into the response and not returning a no content code.  A push in the right direction would be appreciated.

GiedriusS:
Prometheus' /series end-point returns an empty array no matter what: if
there were results or not. We need to follow the same principle in
Thanos as well because our users depend on it.

The same fix has been applied as here:
https://github.com/prometheus/prometheus/blob/fef150f1b5e48652ec6779e2129ae9a0cf13db8a/web/api/v1/api.go#L506

Tests have been updated to account for this difference and plus one test
has been added just for this case.

Closes #1325.

### DIFF
--- a/pkg/query/api/v1.go
+++ b/pkg/query/api/v1.go
@@ -543,7 +543,7 @@ func (api *API) series(r *http.Request) (interface{}, []error, *ApiError) {
 
 	set := storage.NewMergeSeriesSet(sets, nil)
 
-	var metrics []labels.Labels
+	metrics := []labels.Labels{}
 	for set.Next() {
 		metrics = append(metrics, set.At().Labels())
 	}
--- a/pkg/query/api/v1_test.go
+++ b/pkg/query/api/v1_test.go
@@ -302,6 +302,14 @@ func TestEndpoints(t *testing.T) {
 				labels.FromStrings("__name__", "test_metric2", "foo", "boo"),
 			},
 		},
+		// Series that does not exist should return an empty array.
+		{
+			endpoint: api.series,
+			query: url.Values{
+				"match[]": []string{`foobar`},
+			},
+			response: []labels.Labels{},
+		},
 		{
 			endpoint: api.series,
 			query: url.Values{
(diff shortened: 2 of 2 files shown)


bwplotka:
Nice!

PR #1332, opened 2019-07-17
refactor the postings handling

krasi-georgiev:
closes https://github.com/improbable-eng/thanos/issues/1205

Brings the latest improvement of `tsdb.PostingsForMatchers`.
Refactored to use `tsdb.PostingsForMatchers` directly.This way  can update tsdb without any changes to the code.  

Haven't done any optimizations or bench tests, just made the `TestBucketStore_e2e` test pass and wanted to get an early opinion. Will work on bench results and optimize of you think this change is an a good direction.


bwplotka:
Nice! We need refactor after org change (:

bwplotka:
I think we agreed offline that this is not worth pursuing, but it was really nice experiment, thanks for this @krasi-georgiev 

I think the key is problem is that we have to combine fetches against object storage. 

Closing this for now, as it's stale and there no plans to continue this experiment.

PR #1355, opened 2019-07-25
Move suggestion about admin API to appropriate place

### LINKED ISSUE #1354: promclient: Misleading error message about enabling Prometheus admin API

Thanos at commit f6992f7a752e0c3580a9f741bfd4720411b69651

**What happened**

https://github.com/thanos-io/thanos/blob/f6992f7a752e0c3580a9f741bfd4720411b69651/pkg/promclient/promclient.go#L79 suggests to set `--web.enable-admin-api` for Prometheus if a non-200 response is encountered. However, the endpoint used there is not part of the Prometheus admin API.

https://github.com/thanos-io/thanos/blob/f6992f7a752e0c3580a9f741bfd4720411b69651/pkg/promclient/promclient.go#L245 , however, does not contain that suggestion, despite dealing with an admin API endpoint (snapshot).

It should be the other way round.

beorn7:
Fixes #1354

### DIFF
--- a/pkg/promclient/promclient.go
+++ b/pkg/promclient/promclient.go
@@ -76,7 +76,7 @@ func ExternalLabels(ctx context.Context, logger log.Logger, base *url.URL) (labe
 	}
 
 	if resp.StatusCode != 200 {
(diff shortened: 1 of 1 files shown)


beorn7:
The CI failure seems unrelated.

PR #1358, opened 2019-07-29
Multipart features

deejay1:
Some S3 compatible setups may have a different limit for maximum
file size in multipart uploads, so allow for overriding it.

Signed-off-by: Lukasz Jernas <lukasz.jernas@allegro.pl>

* [ENHANCEMENT] Allow for setting minimal part size for S3 multipart uploads

## Changes

Added `part_size` parameter to S3 YAML config file in bytes to allow for setting different multipart part size for chunk upload.

## Verification

Added a config test and tested in a real world scenario.

### DIFF
--- a/pkg/objstore/s3/s3.go
+++ b/pkg/objstore/s3/s3.go
@@ -45,6 +45,7 @@ type Config struct {
 	PutUserMetadata map[string]string `yaml:"put_user_metadata"`
 	HTTPConfig      HTTPConfig        `yaml:"http_config"`
 	TraceConfig     TraceConfig       `yaml:"trace"`
+	PartSize        uint64            `yaml:"part_size,omitempty"`
 }
 
 type TraceConfig struct {
@@ -65,6 +66,7 @@ type Bucket struct {
 	client          *minio.Client
 	sse             encrypt.ServerSide
 	putUserMetadata map[string]string
+	partSize        uint64
 }
 
 // parseConfig unmarshals a buffer into a Config with default HTTPConfig values.
@@ -81,6 +83,12 @@ func parseConfig(conf []byte) (Config, error) {
 	if config.PutUserMetadata == nil {
 		config.PutUserMetadata = make(map[string]string)
 	}
+
+	// Use the default minio minPartSize if not set
+	if config.PartSize == 0 {
+		config.PartSize = 1024 * 1024 * 128
+	}
+
 	return config, nil
 }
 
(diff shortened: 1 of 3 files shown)


pkg/objstore/s3/s3.go
```diff
@@ -81,6 +83,12 @@ func parseConfig(conf []byte) (Config, error) {
 	if config.PutUserMetadata == nil {
 		config.PutUserMetadata = make(map[string]string)
 	}
+
+	// Use the default minio minPartSize if not set
```
bwplotka: missing trailing period (:

pkg/objstore/s3/s3.go
```diff
@@ -81,6 +83,12 @@ func parseConfig(conf []byte) (Config, error) {
...
+	if config.PartSize == 0 {
+		config.PartSize = 1024 * 1024 * 128
```
bwplotka: putting this `1024 * 1024 * 128` in some meaningfully named constant would be nice for a reader.

pkg/objstore/s3/s3.go
deejay1: fixed

pkg/objstore/s3/s3.go
deejay1: hope it's better now, although I couldn't really find a meaningful variable name which wouldn't require an explanation anyway.

pkg/objstore/s3/s3.go
```diff
@@ -81,6 +87,12 @@ func parseConfig(conf []byte) (Config, error) {
 	if config.PutUserMetadata == nil {
 		config.PutUserMetadata = make(map[string]string)
 	}
+
+	// Use the default minPartSize if not set.
```
bwplotka: This comment might be not necessary, but not a blocker.

pkg/objstore/s3/s3.go
```diff
@@ -32,6 +32,10 @@ import (
 // DirDelim is the delimiter used to model a directory structure in an object store bucket.
 const DirDelim = "/"
 
+// Minimum file size after which an HTTP multipart request should be used to upload objects to storage.
+// Set to 128 MiB as in the minio client.
+const minPartSize = 1024 * 1024 * 128
```
bwplotka:
```suggestion
const defaultMinPartSize = 1024 * 1024 * 128
```
.. might be more descriptive here, but not a blocker (:

deejay1: fixed

pkg/objstore/s3/s3.go
deejay1: fixed

FUSAKLA:
Great, thanks for this! We actually lately hit scenario where we need this feature :+1:

FUSAKLA:
maybe worth to add to the changelog?

Issue #1174, opened 2019-05-23
Ability to use multiple dedup labels


jpiper:
Let's say we have two external services

Service A
Service B

Which are scraped by two prometheus setups in two kubernetes clusters (so there's two prometheus instances in each site)

site X: P1 and P2
site Y: P1 and P2 

And then we have a "global" thanos querier that uses the store API of the thanos query nodes in both of these sites.

So we end up with metrics like
```
up{service="A",site="X",prometheus_replica="1"}
up{service="A",site="X",prometheus_replica="2"}
up{service="A",site="Y",prometheus_replica="1"}
up{service="A",site="Y",prometheus_replica="2"}
up{service="B",site="X",prometheus_replica="1"}
up{service="B",site="X",prometheus_replica="2"}
up{service="B",site="Y",prometheus_replica="1"}
up{service="B",site="Y",prometheus_replica="2"}
```

but we really just want
```
up{service="A"}
up{service="B"}
```

Where we want to dedup across site and prometheus_replica. What is the best approach here? Can the global thanos querier ask the site X and site Y to dedup, and then the global just dedups on site? Or is it possible to provide multiple dedup labels?

bwplotka:
Hey, thanks for this (: 

The current way is just use single label. In your case, literally remove prometheus_replica and just use `site` (: OR remove `site` and add `replica=siteX` (:  

We could add multi label, but what's the reason for that if you remove thie extra metadata immdiately on Thanos Query API? so why do you need this multiple labels info? If you will never filter on that? 

The only use case would be if you sometimes deduplicate and for other system you don't deduplicate, is that a case? (:

metalmatze:
I'm currently working on a setup that involves [kube-prometheus](https://github.com/coreos/kube-prometheus) and [kube-thanos](https://github.com/metalmatze/kube-thanos). 

Let's use `node_boot_time_seconds` as example.
We have 2 replicas of Prometheus running and collecting that metric from node-exporter. Next to that we have Thanos running queriers, stores and receive components. Once the Prometheus ingest a time series, they remote write those to the Thanos receive component, adding `prometheus_replica` as label.
Once we query the Thanos queriers they talk to the receive components and those reply with their additional `replica` label. Now we end up with both `replica` and `prometheus_replica` as labels we want to dedup. 

/cc @brancz @squat

bwplotka:
[...]

PR #1362, opened 2019-07-31
Query: allow multiple deduplication label.

krasi-georgiev:
Closes: https://github.com/thanos-io/thanos/issues/1174

## Changes

Can start the querier with multi replica labels like:
```
--query.replica-label=prometheus_replica --query.replica-label=service
```
querier api supports `replicaLabels` url param which will overwrite the `--query.replica-label` cli flag to allow  dynamic deduplication at query time.


krasi-georgiev:
ping @bwplotka 	@brancz 	@squat

pkg/query/api/v1.go
```diff
@@ -204,6 +204,17 @@ func (api *API) parseEnableDedupParam(r *http.Request) (enableDeduplication bool
 	return enableDeduplication, nil
 }
 
+func (api *API) parseReplicaLabelsParam(r *http.Request) (replicaLabels []string, _ *ApiError) {
+	if err := r.ParseForm(); err != nil {
+		return nil, &ApiError{ErrorInternal, errors.Wrap(err, "parse form")}
+	}
+
+	if len(r.Form["replicaLabels[]"]) > 0 {
+		replicaLabels = r.Form["match[]"]
```
brancz: This doesn't seem quite right, is this a copy paste mistake? I would have expected `r.Form["replicaLabels[]"]`.

pkg/query/iter.go
```diff
@@ -325,10 +325,10 @@ func (s *dedupSeriesSet) Next() bool {
 // replica label if it exists
 func (s *dedupSeriesSet) peekLset() labels.Labels {
 	lset := s.peek.Labels()
-	if lset[len(lset)-1].Name != s.replicaLabel {
+	if _, ok := s.replicaLabels[lset[len(lset)-1].Name]; !ok {
```
brancz: I would have expected that we need to verify each replica label here. I feel like I'm missing something about this logic though, can you explain what's happening here?

pkg/query/api/v1.go
krasi-georgiev: good catch!, Thanks!

pkg/query/iter.go
krasi-georgiev: If the last label is not in the `replicaLabels` list than no replica labels are set.
Replica labels are always at the end of the list so if the last one is not present in the `s.replicaLabels` map this guarantees that no replica labels need to be stripped.

If the last labels in the full list is a replica label than strip all replica labels by removing `len(s.replicaLabels)` labels from the full labels list.

pkg/query/api/v1.go
```diff
@@ -204,6 +204,17 @@ func (api *API) parseEnableDedupParam(r *http.Request) (enableDeduplication bool
 	return enableDeduplication, nil
 }
 
+func (api *API) parseReplicaLabelsParam(r *http.Request) (replicaLabels []string, _ *ApiError) {
+	if err := r.ParseForm(); err != nil {
+		return nil, &ApiError{ErrorInternal, errors.Wrap(err, "parse form")}
+	}
+
+	if len(r.Form["replicaLabels[]"]) > 0 {
```
bwplotka: Why not being consistent with every parse function:

```
const dedupParam = "dedup"
	enableDeduplication = true

	if val := r.FormValue(dedupParam); val != "" {
		var err error
		enableDeduplication, err = strconv.ParseBool(val)
		if err != nil {
			return false, &ApiError{errorBadData, errors.Wrapf(err, "'%s' parameter", dedupParam)}
		}
	}
	return enableDeduplication, nil
```
Essentially `ParseForm` vs `FormValue`?

pkg/query/api/v1.go
```diff
@@ -294,7 +310,7 @@ func (api *API) query(r *http.Request) (interface{}, []error, *ApiError) {
 	defer span.Finish()
 
 	begin := api.now()
-	qry, err := api.queryEngine.NewInstantQuery(api.queryableCreate(enableDedup, 0, enablePartialResponse, warningReporter), r.FormValue("query"), ts)
+	qry, err := api.queryEngine.NewInstantQuery(api.queryableCreate(enableDedup, replicaLabels, 0, enablePartialResponse, warningReporter), r.FormValue("query"), ts)
```
bwplotka: We are bit inconsistent here again. 

Everywhere else default values are returned from `parseXXXParam` if param is empty. Can we do the same for `parseReplicaLabelsParam`?

(current version gets default replica label from `api.queryEngine`

pkg/query/iter.go
bwplotka: Worth to comment this indeed.

pkg/query/iter.go
```diff
@@ -325,10 +325,10 @@ func (s *dedupSeriesSet) Next() bool {
...
 		return lset
 	}
-	return lset[:len(lset)-1]
+	return lset[:len(lset)-len(s.replicaLabels)]
```
bwplotka: Hm this does not looks right... What if series includes only one of the replica labels? It seems like you assume that either there are none or all replica labels specified on series.

pkg/query/querier.go
```diff
@@ -21,17 +21,22 @@ import (
...
 // partialResponse controls `partialResponseDisabled` option of StoreAPI and partial response behaviour of proxy.
-type QueryableCreator func(deduplicate bool, maxResolutionMillis int64, partialResponse bool, r WarningReporter) storage.Queryable
+type QueryableCreator func(deduplicate bool, replicaLabels []string, maxResolutionMillis int64, partialResponse bool, r WarningReporter) storage.Queryable
 
 // NewQueryableCreator creates QueryableCreator.
-func NewQueryableCreator(logger log.Logger, proxy storepb.StoreServer, replicaLabel string) QueryableCreator {
-	return func(deduplicate bool, maxResolutionMillis int64, partialResponse bool, r WarningReporter) storage.Queryable {
+func NewQueryableCreator(logger log.Logger, proxy storepb.StoreServer, replicaLabels []string) QueryableCreator {
+	return func(deduplicate bool, replicaLabelsOverwrite []string, maxResolutionMillis int64, partialResponse bool, r WarningReporter) storage.Queryable {
+		if len(replicaLabelsOverwrite) > 0 {
```
bwplotka: I would move this logic to `v1` (api) as all other parameters. What do you think?

pkg/query/querier_test.go
```diff
@@ -216,56 +216,129 @@ func TestQuerier_Series(t *testing.T) {
...
+					{Name: "a", Value: "1"},
+					{Name: "c", Value: "3"},
+					{Name: "d", Value: "4"},
+					{Name: "b", Value: "replica-1"},
+					{Name: "b1", Value: "replica-1"},
+				}},
+				{Labels: []storepb.Label{
+					{Name: "a", Value: "1"},
+					{Name: "c", Value: "4"},
+					{Name: "b", Value: "replica-1"},
```
bwplotka: Let's add tests for single replica label specified in this case as well (I think it's valid case (? )

pkg/query/iter.go
bwplotka: Not sure what semantics we should have here.. Thoughts @brancz ?

pkg/query/api/v1.go
krasi-georgiev: The golang API states that for multiple values need to use `ParseForm`

https://godoc.org/net/http#Request.FormValue

pkg/query/iter.go
brancz: Not sure either, I'm assuming it's similar to the other one, but at least needs a comment as well.

pkg/query/querier.go
brancz: Agreed.

pkg/query/iter.go
bwplotka: I would vote for any replica label not stricly all of them

brancz: Yeah I think the correct semantic would be to only subtract the ones that occur in the label-set, but maybe I'm missing something.

krasi-georgiev: yes that makes sense will refactor.

pkg/query/api/v1.go
bwplotka: fair. (:

pkg/query/iter.go
krasi-georgiev: I added a test for this as well.
Series that include only some of the replica labels are also de-duplicated.

see this test
https://github.com/thanos-io/thanos/pull/1362/files#diff-9dfbf115cb6790c1ff2c931fb91bf9c9R492

pkg/query/querier.go
krasi-georgiev: yes good idea, I missed that.
updated

pkg/query/querier_test.go
krasi-georgiev: yes good idea, added.

krasi-georgiev: can you double check the output now just to make sure that the ordering is what should be expected for the single replica label.

pkg/query/api/v1.go
krasi-georgiev: updated

docs/components/query.md
```diff
@@ -47,12 +49,28 @@ And we query for metric `up{job="prometheus",env="2"}` with this option we will
...
   * `up{job="prometheus",env="2",cluster="2"} 1`
 
-WITHOUT this replica flag (so deduplication turned off), we will get 3 results:
+WITHOUT this replica flag (deduplication turned off), we will get 3 results:
 
   * `up{job="prometheus",env="2",cluster="1",replica="A"} 1`
   * `up{job="prometheus",env="2",cluster="1",replica="B"} 1`
   * `up{job="prometheus",env="2",cluster="2",replica="A"} 1`
 
+### The same example with multiple replica labels:
```
bwplotka:
```suggestion
### The same output will be present for this example with multiple replica labels:
```

I think it might be clearer

go.mod
```diff
@@ -35,6 +35,7 @@ require (
 	github.com/prometheus/client_golang v1.1.0
 	github.com/prometheus/common v0.6.0
 	github.com/prometheus/prometheus v1.8.2-0.20190819201610-48b2c9c8eae2 // v1.8.2 is misleading as Prometheus does not have v2 module. This is pointing to one commit after 2.12.0.
+	github.com/prometheus/tsdb v0.10.0 // indirect
```
bwplotka: This is definitely wrong, did we import tsdb by accident here?

pkg/query/api/v1.go
```diff
@@ -185,6 +188,21 @@ func (api *API) parseEnableDedupParam(r *http.Request) (enableDeduplication bool
 	return enableDeduplication, nil
 }
 
+func (api *API) parseReplicaLabelsParam(r *http.Request) (replicaLabels []string, _ *ApiError) {
+	const replicaLabelsParam = "replicLabeals[]"
```
bwplotka: typo

So it means we don't have test for this in `v1_test.go` (: Can we add test for this?

pkg/query/api/v1.go
```diff
@@ -212,6 +230,7 @@ func (api *API) parsePartialResponseParam(r *http.Request) (enablePartialRespons
 	const partialResponseParam = "partial_response"
 	enablePartialResponse = api.enablePartialResponse
 
+	// Overwrite the cli flag when provided as a query parameter.
```
bwplotka: Did you mean to put this below?

bwplotka: Anyway I would be fine with just what we have in doc, so maybe we can drop this comment?

pkg/query/iter.go
```diff
@@ -360,10 +360,18 @@ func (s *dedupSeriesSet) Next() bool {
...
 	lset := s.peek.Labels()
-	if lset[len(lset)-1].Name != s.replicaLabel {
+	if len(s.replicaLabels) == 0 {
 		return lset
 	}
-	return lset[:len(lset)-1]
+	// Check how many replica labels are present so that these are removed.
+	var totalToRemove int
+	for index := 0; index < len(s.replicaLabels); index++ {
+		if _, ok := s.replicaLabels[lset[len(lset)-index-1].Name]; ok {
```
bwplotka: Nice, that might work nicely :+1:

pkg/query/querier_test.go
bwplotka: It looks good IMO.

bwplotka:
Looks like netlify OOMs (: https://app.netlify.com/sites/thanos-io/deploys/5d7f620fb4a56e00093ad620

krasi-georgiev:
> Why you put it here then and not for parseDownsamplingParamMillis and parseReplicaLabelsParam? I don't get it ):

I see what you mean now , yes that would be better to be as a comment to the function itself.

If I don't forget will open a PR.

PR #1363, opened 2019-07-31
receive: Hash-ring metrics

kakkoyun:
This PR adds new metrics to monitor hash-rings.
It exposes the number of nodes and the number of tenants per hash-rings.

## Changes

* Adds two new metrics to receive `thanos_receive_hashring_nodes` and `thanos_receive_hashring_tenants`.

## Verification

By using make test and manual testing against receive API, it seems like the previous behavior kept intact.

### DIFF
--- a/pkg/receive/config.go
+++ b/pkg/receive/config.go
@@ -96,6 +96,8 @@ func NewConfigWatcher(logger log.Logger, r prometheus.Registerer, path string, i
 			c.changesCounter,
 			c.errorCounter,
 			c.refreshCounter,
+			c.hashringNodesGauge,
+			c.hashringTenantsGauge,
 		)
 	}
 
--- a/pkg/receive/config.go
+++ b/pkg/receive/config.go
@@ -79,13 +79,13 @@ func NewConfigWatcher(logger log.Logger, r prometheus.Registerer, path string, i
 			}),
 		hashringNodesGauge: prometheus.NewGaugeVec(
 			prometheus.GaugeOpts{
-				Name: "thanos_receive_hashrings_nodes",
+				Name: "thanos_receive_hashring_nodes",
 				Help: "The number of nodes per hashring.",
 			},
 			[]string{"name"}),
 		hashringTenantsGauge: prometheus.NewGaugeVec(
 			prometheus.GaugeOpts{
-				Name: "thanos_receive_hashrings_tenants",
+				Name: "thanos_receive_hashring_tenants",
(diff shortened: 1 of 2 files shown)


kakkoyun:
cc @brancz

pkg/receive/config.go
```diff
@@ -75,6 +77,18 @@ func NewConfigWatcher(logger log.Logger, r prometheus.Registerer, path string, i
...
+				Name: "thanos_receive_hashring_nodes",
+				Help: "The number of nodes per hashring.",
+			},
+			[]string{"name"}),
+		hashringTenantsGauge: prometheus.NewGaugeVec(
+			prometheus.GaugeOpts{
+				Name: "thanos_receive_hashring_tenants",
+				Help: "The number of tenants per hashring.",
+			},
+			[]string{"name"}),
```
bwplotka: Just curious on cardinality of this - this change on restart time or is fairly consistent across restarts?

kakkoyun: This is consistent across restarts. `Name` comes from a manually maintained list which is fed through a configuration file.

pkg/receive/config.go
```diff
@@ -75,6 +77,18 @@ func NewConfigWatcher(logger log.Logger, r prometheus.Registerer, path string, i
 				Name: "thanos_receive_hashrings_file_refreshes_total",
 				Help: "The number of refreshes of the hashrings configuration file.",
 			}),
+		hashringNodesGauge: prometheus.NewGaugeVec(
+			prometheus.GaugeOpts{
+				Name: "thanos_receive_hashring_nodes",
```
bwplotka: nodes or endpoints? (: Don't have strong opinion.

pkg/receive/config.go
```diff
@@ -75,6 +77,18 @@ func NewConfigWatcher(logger log.Logger, r prometheus.Registerer, path string, i
...
+			[]string{"name"}),
+		hashringTenantsGauge: prometheus.NewGaugeVec(
+			prometheus.GaugeOpts{
+				Name: "thanos_receive_hashring_tenants",
+				Help: "The number of tenants per hashring.",
+			},
+			[]string{"name"}),
 	}
 
 	if r != nil {
```
bwplotka: I think those metrics will work bit better if we would register them below ^^

kakkoyun: 🤦‍♂  Thanks for pointing that out.

PR #1366, opened 2019-08-02
refactor: replicate write request in receive() if necessary

YaoZengzeng:
There is a race condition: the `h.hashring` could be set to nil in `Hashring(...)` without lock. So if a request pass the `testReady()` and then `h.hashring` be set to nil, the program would panic.

And if the request is not replicated yet, we should do it when we receive it instead of doing it in `parallelizeRequests` after some unnecessary calculation in `forward`. 

In addtion, the refactor make the locking of `h.hashring` more easy.

### DIFF
--- a/pkg/receive/handler.go
+++ b/pkg/receive/handler.go
@@ -255,10 +255,17 @@ func (h *Handler) receive(w http.ResponseWriter, r *http.Request) {
 
 	tenant := r.Header.Get(h.options.TenantHeader)
 
-	// Forward any time series as necessary. All time series
-	// destined for the local node will be written to the receiver.
-	// Time series will be replicated as necessary.
-	if err := h.forward(r.Context(), tenant, rep, &wreq); err != nil {
+	if !rep.replicated && h.options.ReplicationFactor > 1 {
+		// If the request is not yet replicated, let's replicate it.
+		// If the replication factor isn't greater than 1, let's
+		// just forward the requests.
+		err = h.replicate(r.Context(), tenant, &wreq)
+	} else {
+		// Forward any time series as necessary. All time series
+		// destined for the local node will be written to the receiver.
+		err = h.forward(r.Context(), tenant, rep, &wreq)
+	}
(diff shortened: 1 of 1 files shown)


YaoZengzeng:
@brancz PTAL

brancz:
cc @squat

squat:
@YaoZengzeng right now this PR introduces a sneaky bug.

Notice that the `replicate` func only looks for the hashring of the first timeseries of the write request parameter. This is done because the `replicate` func was only intended to be called AFTER all the individual timeseries in a write request were split off and forwarded to the correct nodes.

With this change set, `replicate` is called on the entire received write request, which can include time series meant for other nodes. Now when `parallelizeRequests` is called from `replicate` and the destination endpoint is the local node, the entire write request will be written locally, even timeseries that are not destined for the local node.

The race does indeed occur, but I don’t think this change is necessary. Let’s simply introduce an RW lock that locks only during the hashring access portions of the `replicate` and  `forward` funcs, ie the loops before `parallelizeRequests` is called.

YaoZengzeng:
@squat You are right.

The code is awesome, after splitting the time series with same hash result will be grouped together, they all will be sent to same node group.  So in `receive()`, we could use `&wreq.Timeseries[0]` to represent all the series in the group.

And if needed, I'd like to help to do the RWLock stuff :)

squat:
> So in receive(), we could use &wreq.Timeseries[0] to represent all the series in the group.

s/receive/replicate/, right?

YaoZengzeng:
`receive()` -> `replicate()` 😂

squat:
Agreed :) thanks for the enthusiasm and for taking a look close at the receive component! Let’s close this PR and start work on the race. If you are up for making an initial draft of the locking PR please go for it and I would be happy to review. We can also discuss and plan on the thanos channel of the cncf slack.

YaoZengzeng:
Close this PR and make another for race condition.

PR #1367, opened 2019-08-02
store: use a slice instead of a map for storing the symbols in an index

xrstf:
This PR attempts to reduce the memory used by the store component when initializing the local cache during startup. It transforms the `map[uint32]string` into a `[]string`, accepting that there might be many, many empty elements in the slice. This also makes it so that there is no difference between a non-existing label and an empty label.

Testing showed a reduction from 4.0 GiB to 3.6 GiB when loading 3 GiB data from S3 (on an XPS 13 running Ubuntu 18.04).

## Verification

I let the unit tests run, both with the original, my patched and and intentionally broken implementation to check if there are tests for the `lookupSymbol()` function at all.

### DIFF
--- a/pkg/block/index.go
+++ b/pkg/block/index.go
@@ -175,7 +175,7 @@ func WriteIndexCache(logger log.Logger, indexFn string, fn string) error {
 // ReadIndexCache reads an index cache file.
 func ReadIndexCache(logger log.Logger, fn string) (
 	version int,
-	symbols map[uint32]string,
+	symbols []string,
 	lvals map[string][]string,
 	postings map[labels.Label]index.Range,
 	err error,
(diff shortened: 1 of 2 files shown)


pkg/block/index.go
```diff
@@ -201,6 +201,14 @@ func ReadIndexCache(logger log.Logger, fn string) (
 	lvals = make(map[string][]string, len(v.LabelValues))
 	postings = make(map[labels.Label]index.Range, len(v.Postings))
 
+	maxSymbolID := 0
```
povilasv:
```suggestion
	var maxSymbolID uint32
```

pkg/block/index.go
```diff
@@ -201,6 +201,14 @@ func ReadIndexCache(logger log.Logger, fn string) (
...
+	for o := range v.Symbols {
+		if int(o) > maxSymbolID {
```
povilasv:
```suggestion
		if o > maxSymbolID {
```

pkg/store/bucket.go
```diff
@@ -1260,11 +1260,12 @@ func newBucketIndexReader(ctx context.Context, logger log.Logger, block *bucketB
 }
 
 func (r *bucketIndexReader) lookupSymbol(o uint32) (string, error) {
-	s, ok := r.block.symbols[o]
-	if !ok {
-		return "", errors.Errorf("bucketIndexReader: unknown symbol offset %d", o)
+	idx := int(o)
+	if idx < len(r.block.symbols) {
```
povilasv: Consider indenting errow flow as per https://github.com/golang/go/wiki/CodeReviewComments#indent-error-flow

i.e. checking for  `if idx  >= len(..)` and returning and error

pkg/block/index.go
xrstf: Good catch!

GiedriusS:
Shows an improvement of a couple hundred MB, thanks!

bwplotka:
Nice! Thanks!

bwplotka:
In fact there was idea to maintain the unified symbols across all the blocks in store (: But I think this is a good step anyway.

Issue #703, opened 2019-01-02
Expose query.max-samples flag for Thanos Querier.


bwplotka:
Thanks of upgrade of promQL package, we got new Prometheus change:
> [FEATURE] Make maximum number of samples per query configurable via `--query.max-samples` flag. #4513

Which gives ability to limit number of samples. This with max-concurrent queries, allows to calculate maximum amount of samples Querier can hold and set it accordingly.

Open question: Does max-samples helps with blocking OOMs of StoreAPI? At first glance, yes as we do max sample verify while we are iterating over messages from stream, so we will not fetch more much more samples than  `query.max-samples` define. (However store API implementation can buffer more e.g sidecar with non-streamed remote-read). 
 
*Acceptance Criteria:*
* expose query.max-samples for Thanos Querier
* Define default value. Should be unlimited (IMO. no)

CC @domgreen

GiedriusS:
@bwplotka does this even make sense? Could you PTAL at this comment here: https://github.com/thanos-io/thanos/pull/1369#issuecomment-518697874?

ppanyukov:
> Open question: Does max-samples helps with blocking OOMs of StoreAPI?

Short answer: no. Or more precisely, "it depends".

From my own tests:

- this is a difficult limit to use meaningfully;
- the actual memory allocated by PromQL very largely depends on the query and the cardinality;
- so querier can still OOM.

fpetkovski:
@GiedriusS do you think it makes sense to allow configure this (or a similar limit) in the [lazySeriesSet](https://github.com/thanos-io/thanos/blob/de0e3848ff6085acf89a5f77e053c555a2cce550/pkg/query/iter.go#L306)?

GiedriusS:
I think it's a bit too late there. I think the limit should apply before Recv() https://github.com/thanos-io/thanos/blob/main/pkg/store/proxy.go#L506

PR #1369, opened 2019-08-03
add flag max query samples

yeya24:

## Changes
fix https://github.com/thanos-io/thanos/issues/703

### DIFF
--- a/cmd/thanos/query.go
+++ b/cmd/thanos/query.go
@@ -6,12 +6,13 @@ import (
 	"math"
 	"net/http"
 	"path"
+	"strconv"
 	"time"
 
 	"github.com/go-kit/kit/log"
 	"github.com/go-kit/kit/log/level"
 	"github.com/oklog/run"
-	opentracing "github.com/opentracing/opentracing-go"
+	"github.com/opentracing/opentracing-go"
 	"github.com/pkg/errors"
 	"github.com/prometheus/client_golang/prometheus"
 	"github.com/prometheus/common/route"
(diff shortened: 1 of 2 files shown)


GiedriusS:
In the original issue @bwplotka says:
>>>
At first glance, yes as we do max sample verify while we are iterating over messages from stream, so we will not fetch more much more samples than query.max-samples define. 

Is this really true? Could you or someone point me to code where this happens? I'm afraid that the PromQL engine only sees the "end results", after the `Select()` gets everything from StoreAPI nodes and, optionally, deduplicates them.

GiedriusS:
@yeya24 do you have any comments on my previous question? @bwplotka WDYT?

yeya24:
How to reopen this?

bwplotka:
Just ask us (: Or help us maintaining Thanos and become Thanos maintainer or Triage (: Would you be interested @yeya24 ? (:

GiedriusS:
What about my comment? AIUI this limit is only being applied after the results from `Select()` have been retrieved. The `Select()` called is this one: https://github.com/thanos-io/thanos/blob/master/pkg/query/querier.go#L168. Inside of it, we already retrieve all of the results and only then pass them on to the PromQL engine which actually applies those limits. I tested out with `--query.max-samples=1` and a query `sum(up) by (job)`. Added a small `fmt.Println` there: `fmt.Println(len(resp.seriesSet))` and it prints out numbers like `7596` which means AIUI that those series have been loaded but they haven't been passed to the engine itself. After that, I get that error but that data is still in my querier.

yeya24:
@GiedriusS Thanks for the comment. I think you are right, these samples from  Select() are already in the Querier's memory and the help message in this flag is quite misleading. What about `Maximum number of samples in a single query evaluation`? Do you think should we change the name as well?

bwplotka:
Sorry, I totally missed this. You are right @GiedriusS I believe in this case we even have to calculate by estimating chunks again when they appear on stream... 🤔

GiedriusS:
Yep, that's why I was hesitant to merge this. Maybe let's revert this current version and remake it into something that uses a counter on the `Recv()` part? I feel like no matter how we would name this, it wouldn't bring much benefit because AIUI the main objective is to prevent OOM situations but currently it almost doesn't help with that :/ and I just want to avoid getting lots of tickets on github about how this option doesn't work

bwplotka:
Agree - let's revert. Sorry my bad - premature approve.

GiedriusS:
> Agree - let's revert. Sorry my bad - premature approve.

It's ok, no worries! :smile:

Issue #1331, opened 2019-07-16
S3/GCS: Upload inconsistency (missing chunk file)


0robustus1:
**Thanos, Prometheus and Golang version used**

* thanos: improbable/thanos:v0.4.0
  * go: 1.12.4
* prometheus: quay.io/prometheus/prometheus:v2.7.2
  * go: 1.11.5

**What happened**

We use an on-premise S3 store (in thanos via `type: S3` in the object store
config) that experienced availability issues for a 3 - 15 hour period (3 hours
of common connection issues, 15 hours of less common connection issues).
Multiple thanos components (shipper/sidecar, compactor) experienced timeouts
while awaiting responses from the S3 service.

This resulted in the shipper/sidecar not writing complete blocks.
We observed the following types of partially written blocks:

* *block* directory present, *meta.json* present, *index* and *chunks* directory missing.
* *block* directory present, *meta.json* present, *index* present, *chunks* directory missing.
* *block* directory present, *meta.json* present, *index* present, *chunks* directory present, but not all chunk files being present.

Those partially written (one might call them corrupted) blocks caused subsequent issues:

* store components failed queries over a timespan that would contain a partially written block
* compactor component crashed with non-zero exitcode (log messages before the crash below)

**What you expected to happen**

* sidecars do not give up in completing partially written blocks
* compactors don't crash on partially written blocks (either skipping them or removing them outright)
* stores don't fail queries on partially written blocks (e.g. by ignoring those blocks)

**How to reproduce it (as minimally and precisely as possible)**:

We have been running into issues trying to build a minimal reproducible scenario.
It would seem that it should be enough to have blocks with the mentioned criteria:

* *block* directory present, *meta.json* present, *index* and *chunks* directory missing.
* *block* directory present, *meta.json* present, *index* present, *chunks* directory present, but not all chunk files being present.

When trying this however we ran into the situation that these blocks did not run
into the compaction plan (see [here](https://github.com/improbable-eng/thanos/blob/master/pkg/compact/compact.go#L791-L845)).

It seems however that once this is considered in the compaction plan and [GatherIndexIssueStats](https://github.com/improbable-eng/thanos/blob/master/pkg/compact/compact.go#L822) is executed the compactor will fail if the *index* file is not present. If chunks are missing it will fail later in the prometheus tsdb compaction code.

**Partial logs to relevant components**

[...]

PR #1370, opened 2019-08-05
compact: handle partially written blocks

0robustus1:
originates from https://github.com/thanos-io/thanos/issues/1331

* [Changed] skip compacting of partially written blocks

## Changes

* Determine whether a block is only partially written (e.g. meta.json is present, but index file is missing)
* Expose a metric that counts found and therefore skipped partially written blocks (`thanos_compact_group_partially_written_blocks_detected_total`)
* Remove partially written blocks from the compaction plan before running prometheus/tsdb compaction on the final compaction plan.

## Verification

We encountered partially written blocks in our own thanos system. Adding this fix let the compactor skip the partially written block (instead of crashing/exiting). It was then later garbage collected.

I would like to provide an automated test for this but i am having trouble creating a minimal artificial error scenario in which to reproduce the crash.

### DIFF
--- a/pkg/block/index.go
+++ b/pkg/block/index.go
@@ -264,6 +264,59 @@ type Stats struct {
 	// OutOfOrderLabels represents the number of postings that contained out
 	// of order labels, a bug present in Prometheus 2.8.0 and below.
 	OutOfOrderLabels int
+	// IndexMissing represents whether the index could be found at all. In a
+	// partially written block (https://github.com/thanos-io/thanos/issues/1331)
+	// situation it might be missing.
+	IndexMissing bool
+	// ChunksReferencedInIndex lists all chunks that have been referenced within
+	// the index. It is used to determine whether there is a mismatch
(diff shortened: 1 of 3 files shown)


bwplotka:
Thanks for that. However before getting into full block scan / detection you propose it would good to fully understand what's wrong with your object storage. Even on full or partial unavailability the worst what can happen is that upload will fail in mid stage. Since meta.json is uploaded at the end, there is no possiblity that you can have missing index but uploaded meta.json. Can we try to figure out first why that happens?

* Is Prometheus filesystem unstable (e.g NFS?) in your case? 
* Is somehow S3 object storage eventually consistent and there was some super long delay? 

AFAIK write via minio client Go library should synchronous.

bwplotka:
Let's continue discussions on other issues, happy to proceed with better error on compactor side. Thanks for this!

0robustus1:
So this PR should be closed? Or should we adjust to just produce a different error message before exiting the compactor (meaning: remove the skip functionality and remove the metric)?

PR #1391, opened 2019-08-09
scripts.quickstart: Use environment variables for executables

f9n:


brancz:
I think we’re going a little abstract here. This is just a QuickStart script, I feel we’re at a point where if we introduce these things it’ll grow less useful for the QuickStart purpose and become a script people start using for prod deploys which we absolutely do not want people to do.

scripts/quickstart.sh
```diff
@@ -5,6 +5,11 @@
 
 trap 'kill 0' SIGTERM
 
+export MINIO_EXECUTABLE="minio"
```
bwplotka: Why not just variable without `export`?

f9n: I just want to use variables dynamically. I didn't care to define it as an environment variable.

bwplotka: So can we remove export then? (:

f9n: Sure :)

scripts/quickstart.sh
```diff
@@ -168,7 +175,7 @@ sleep 0.5
 # Start to query nodes.
 for i in `seq 1 2`
 do
-  ./thanos query \
+  $THANOS_EXECUTABLE query \
```
bwplotka: One more thing: Technically `${...}` is the pattern for variables. Can we be consistent here and on other variables you define here?

Issue #1318, opened 2019-07-10
*: Support multi-tenancy on object storage using custom prefix


edevouge:
# Support multi-tenancy on object storage using custom prefix

## Context / Use case 
On a multi-cluster kubernetes setup with multiple prometheus instances (one per namespace) using PrometheusOperator. Each prometheus instance embed its own thanos sidecar and for performance & scalability purpose data chuncks are sharded accross multiple s3 buckets.
For each prometheus instance, there is a dedicated thanos store and a thanos compactor working on its own bucket. One thanos query is setup to map/reduce promql queries across multiple stores and thanos-sidecars.

## Problems
- With this setup there is a proliferation of s3 buckets.
- Bucket creation lifecycle is not as dynamic as kubernetes resources (kubectl apply versus terraform apply). Require use of multiple tools.

## Question / feature request 
Is there a way to share a bucket across multiple thanos instances (keeping the sharding property) ?

Otherwise, a cool feature could be the ability to add a custom prefix on thanos objects (objstore.config). So we could take advantage of s3 bucket multi-tenancy.

+ e.g : `s3://<my-bucket-name>/**<custom-prefix>**/<thanos-objects>`

edevouge:
# Design proposal to discuss

## Scope :
  - Enrich `objstore` configuration to support a custom prefix in s3 object storage key name
  - Focus on AWS S3 implementation for a first contribution

## Expected result

[Before] Current flat bucket file tree :
```
  bucket-root
  ├── 01DH3V0E80CBEK8SK00A93SH7Z
  │    ├── chunks
  │    ├── index
  │    └── meta.json
  ├── 01DG4BDF8R3FDSZ3EM6MVHVTFX
  │    ├── chunks
  │    ├── index
  │    └── meta.json
  ├──  01DH3V01JXFHC0C9ABCM7MZGP4
  │    ├── chunks
  │    ├── index
  │    └── meta.json
  └── debug
       └── metas
           ├── 01DH3V0E80CBEK8SK00A93SH7Z.json
           ├── 01DG4BDF8R3FDSZ3EM6MVHVTFX.json
           └──  01DH3V01JXFHC0C9ABCM7MZGP4.json
```
[After] Expected prefixed bucket file tree :
```
  bucket-root
  └── my
      └── custom
          └── prefix
              ├── 01DH3V0E80CBEK8SK00A93SH7Z
              │    ├── chunks
              │    ├── index
              │    └── meta.json
[...]

PR #1392, opened 2019-08-09
feat(pkg/objstore/s3): support for prefixes

jaredallard:
* [] CHANGELOG Support for s3 prefixes

## Changes

I added support for s3 prefixes based on the amazing work proposed by #1318 

## Verification

Tested configuration changes via unit test, tested integration wise by running it.


jaredallard:
Going to fix the docs related stuff... as soon as I figure out how that works

jaredallard:
Ready to go :tada:

pkg/objstore/s3/s3.go
```diff
@@ -92,6 +94,11 @@ func parseConfig(conf []byte) (Config, error) {
 		config.PartSize = defaultMinPartSize
 	}
 
+	// here for testing
+	if config.Prefix != "" {
+		config.Prefix = strings.TrimSuffix(config.Prefix, DirDelim) + DirDelim
```
povilasv: Why is this needed?

pkg/objstore/s3/s3.go
```diff
@@ -208,6 +216,17 @@ func validate(conf Config) error {
 	if conf.AccessKey != "" && conf.SecretKey == "" {
 		return errors.New("no s3 secret_key specified while access_key is present in config file; either both should be present in config or envvars/IAM should be used.")
 	}
+
+	// ensure that the prefix isn't too long
+	if conf.Prefix != "" && len(conf.Prefix) >= 1024 {
+		return errors.New("prefix is too long (limited by Amazon at 1024 bytes long)")
```
povilasv: I don't think this is a good idea, as S3 isn't only amazon it's also minio, ceph and other s3 compatible stores, so if other implementations support long prefixes, this would break those.

pkg/objstore/s3/s3.go
```diff
@@ -208,6 +216,17 @@ func validate(conf Config) error {
...
+	}
+
+	// trim trailing slash and insert one
```
povilasv: Why are we doing this?

pkg/objstore/s3/s3.go
```diff
@@ -230,16 +249,17 @@ func (b *Bucket) Iter(ctx context.Context, dir string, f func(string) error) err
 		dir = strings.TrimSuffix(dir, DirDelim) + DirDelim
 	}
 
-	for object := range b.client.ListObjects(b.name, dir, false, ctx.Done()) {
+	for object := range b.client.ListObjects(b.name, b.objectPrefix+dir, false, ctx.Done()) {
+		keyName := strings.Replace(object.Key, b.objectPrefix, "", 1)
```
povilasv:
```suggestion
		keyName := strings.TrimPrefix(object.Key, b.objectPrefix)
```

pkg/objstore/s3/s3.go
jaredallard: for the configuration parsing test

pkg/objstore/s3/s3.go
jaredallard: normalization of paths for object keys, don't want two `/` on accident.

pkg/objstore/s3/s3.go
jaredallard: Good one, I should've caught that.

pkg/objstore/s3/s3.go
jaredallard: OK, I can remove that.

pkg/objstore/s3/s3.go
povilasv: Adding normal code just for tests doesn't sound like a good idea, consider removing this.

pkg/objstore/s3/s3.go
povilasv: Consider adding user validation, instead of trying to fix the error.

pkg/objstore/s3/s3.go
jaredallard: Alright.

pkg/objstore/s3/s3.go
jaredallard: Not sure I agree with that, I think this is a perfectly valid solution,

jaredallard:
@edevouge feel free to nullify this PR with your work, I needed to implement this for my own use cases, and I would much rather you get credit than I.

bwplotka:
So, it looks good, but why are we limiting ourselves to S3 implementation? If we are getting into this, I think it might be a part of our interface instead. E.g Adding `Prefix` to https://github.com/thanos-io/thanos/blob/2c5f2cde11f5cd100f147ad2e5d4dbeccbd604c5/pkg/objstore/client/factory.go#L32 

This will be then consistent across all object storages.

edevouge:
> @edevouge feel free to nullify this PR with your work, I needed to implement this for my own use cases, and I would much rather you get credit than I.

@jaredallard : I appreciate your contribution, but it would be great the next time you decided to implement a design proposal to post a small comment on the issue thread to explain your intention. Because, I was also writing some code for this feature and it's not really efficient to do the job twice. I'll let you finish the job.

jaredallard:
> So, it looks good, but why are we limiting ourselves to S3 implementation? If we are getting into this, I think it might be a part of our interface instead. E.g Adding `Prefix` to
> 
> https://github.com/thanos-io/thanos/blob/2c5f2cde11f5cd100f147ad2e5d4dbeccbd604c5/pkg/objstore/client/factory.go#L32
> 
> This will be then consistent across all object storages.

Good point, I'll rework this.

AnandDevan:
hi folks, is this waiting for the equivalent implementation on the other object stores for merging?
would like to see prefix supported for s3 soon

svyotov:
This looks like a good feature, do you think it will re-open?

pkg/objstore/s3/s3.go
```diff
@@ -87,6 +89,11 @@ func parseConfig(conf []byte) (Config, error) {
 		return Config{}, err
 	}
 
+	// here for testing
```
bwplotka, line 92: This looks wrong as @povilasv mentioned in https://github.com/thanos-io/thanos/pull/1392/files#r312902398

docs/storage.md
```diff
@@ -90,6 +90,7 @@ config:
     insecure_skip_verify: false
   trace:
     enable: false
+  prefix: ""
```
bwplotka, line 93: I would be curious and actually I think it's must-have to actually make this change on bucket level, not each bucket client level.. WDYT? (:

bwplotka:
Well it really depends on @jaredallard if he is willing to continue this change.

I believe to merge it we need to do it per global object storage config, not just S3, so all providers can reuse this prefix logic (: We can also discuss if it should be `Prefix` or `Prefixes` potentially on issue: https://github.com/thanos-io/thanos/issues/1318

Help wanted, we definitely would like to have this in!

jaredallard:
Hey, I'm likely to not drive this work to completion. I am no longer on the team that needed this change at my company (they have maintained a fork with this AFAIK). I may come and do it in my spare time, but I have a massive backlog of things to do outside of work and this is pretty low on that list for that reason 😓 

Sorry everyone!

bwplotka:
This is totally ok (: No worries, and good luck! 

It's best if you are transparent about the priorities, so someone else can take your work. :+1 Thanks!


so... let's keep it close and help wanted (:

Issue #1335, opened 2019-07-18
Handle (ignore) partial uploads on downsampling and apply retention cycles


bwplotka:
AC:
* Meta files sync should ignore not-complete blocks for downsampling and retention cycle.
* Removal of malformed block if any is ONLY on compaction cycle.

We have nice concurrent logic here: https://github.com/improbable-eng/thanos/blob/10f84a7d7f06813fda97e17d609a9969fcda037f/pkg/compact/compact.go#L169

We did not apply it here: https://github.com/improbable-eng/thanos/blob/10f84a7d7f06813fda97e17d609a9969fcda037f/cmd/thanos/downsample.go#L145

And not on retention: https://github.com/improbable-eng/thanos/blob/1d582af96b2cd412ade46dcfa07e6d40ffb3c176/pkg/compact/retention.go#L16

bwplotka:
It fixes compactor crashes like 
```
level=error ts=2019-07-15T21:09:56.108249354Z caller=main.go:182 msg="running command failed" err="error executing compaction: first pass of downsampling failed: retrieve bucket block metas: get meta for block 01DFVPYR64AGDW009PF99N2YFY: The specified key does not exist."
```

This is because when doing sync and sidecar uploading block we might hit block with no meta.json.

lx223:
Hey @bwplotka I can help with this.

bwplotka:
oh yes please @lx223 

Essentially we have 3 different sync implementation. I think we should have nice one meta syncer everyone can reuse: compactor for compating/downsampling/retention but also store gateway.

bwplotka:
Any progress @lx223 ?

lx223:
sorry for the delay; have been busy with work. should have something up for review soon.

lx223:
PR: https://github.com/thanos-io/thanos/pull/1394 PTAL

bwplotka:
Still valid I think, ref: https://cloud-native.slack.com/archives/CK5RSSC10/p1572453579422500

bwplotka:
It's getting quite important to get this in, another point of failure:
* https://github.com/thanos-io/thanos/issues/1655#issuecomment-559068163
* FIxes like https://github.com/thanos-io/thanos/pull/1807

bwplotka:
https://github.com/thanos-io/thanos/pull/1394 was our first attempt and can be used as a base (: Will get into this in free time, unless there are other volunteers (:

PR #1394, opened 2019-08-09
Unified fetching/syncing block metadata everywhere 

lx223:

## Changes

* add a new block meta fetcher that fetches block metadata stored in a bucket
* replace some of the existing block md fetching code with the new fetcher
* change the signature of `DownloadMeta` to use the more restrictive `BucketReader`

## Verification

Unit test

### DIFF
--- a/cmd/thanos/bucket.go
+++ b/cmd/thanos/bucket.go
@@ -36,6 +36,10 @@ import (
 	kingpin "gopkg.in/alecthomas/kingpin.v2"
 )
 
+const (
+	defaultBlockMDFetchConcurrency = 3
+)
+
 var (
 	issuesMap = map[string]verifier.Issue{
 		verifier.IndexIssueID:                verifier.IndexIssue,
@@ -175,28 +179,26 @@ func registerBucketLs(m map[string]setupFunc, root *kingpin.CmdClause, name stri
 
 		var (
 			format     = *output
-			printBlock func(id ulid.ULID) error
+			printBlock func(id ulid.ULID, md *metadata.Meta) error
 		)
 
 		switch format {
 		case "":
-			printBlock = func(id ulid.ULID) error {
+			printBlock = func(id ulid.ULID, md *metadata.Meta) error {
 				fmt.Fprintln(os.Stdout, id.String())
 				return nil
 			}
 		case "wide":
-			printBlock = func(id ulid.ULID) error {
-				m, err := block.DownloadMeta(ctx, logger, bkt, id)
-				if err != nil {
-					return err
+			printBlock = func(id ulid.ULID, md *metadata.Meta) error {
+				if md == nil {
+					return nil
 				}
-
-				minTime := time.Unix(m.MinTime/1000, 0)
-				maxTime := time.Unix(m.MaxTime/1000, 0)
+				minTime := time.Unix(md.MinTime/1000, 0)
+				maxTime := time.Unix(md.MaxTime/1000, 0)
 
 				if _, err = fmt.Fprintf(os.Stdout, "%s -- %s - %s Diff: %s, Compaction: %d, Downsample: %d, Source: %s\n",
-					m.ULID, minTime.Format("2006-01-02 15:04"), maxTime.Format("2006-01-02 15:04"), maxTime.Sub(minTime),
-					m.Compaction.Level, m.Thanos.Downsample.Resolution, m.Thanos.Source); err != nil {
+					md.ULID, minTime.Format("2006-01-02 15:04"), maxTime.Format("2006-01-02 15:04"), maxTime.Sub(minTime),
+					md.Compaction.Level, md.Thanos.Downsample.Resolution, md.Thanos.Source); err != nil {
 					return err
 				}
 				return nil
@@ -205,25 +207,22 @@ func registerBucketLs(m map[string]setupFunc, root *kingpin.CmdClause, name stri
 			enc := json.NewEncoder(os.Stdout)
 			enc.SetIndent("", "\t")
 
-			printBlock = func(id ulid.ULID) error {
-				m, err := block.DownloadMeta(ctx, logger, bkt, id)
-				if err != nil {
-					return err
+			printBlock = func(id ulid.ULID, md *metadata.Meta) error {
+				if md == nil {
+					return nil
 				}
-				return enc.Encode(&m)
+				return enc.Encode(&md)
 			}
 		default:
 			tmpl, err := template.New("").Parse(format)
 			if err != nil {
 				return errors.Wrap(err, "invalid template")
 			}
-			printBlock = func(id ulid.ULID) error {
-				m, err := block.DownloadMeta(ctx, logger, bkt, id)
-				if err != nil {
-					return err
+			printBlock = func(id ulid.ULID, md *metadata.Meta) error {
+				if md == nil {
+					return nil
 				}
-
-				if err := tmpl.Execute(os.Stdout, &m); err != nil {
+				if err := tmpl.Execute(os.Stdout, &md); err != nil {
 					return errors.Wrap(err, "execute template")
 				}
 				fmt.Fprintln(os.Stdout, "")
@@ -231,13 +230,18 @@ func registerBucketLs(m map[string]setupFunc, root *kingpin.CmdClause, name stri
 			}
 		}
 
-		return bkt.Iter(ctx, "", func(name string) error {
-			id, ok := block.IsBlockDir(name)
-			if !ok {
-				return nil
+		mdFetcher := block.NewMetadataFetcher(logger, defaultBlockMDFetchConcurrency, bkt)
+		mds, err := mdFetcher.Fetch(ctx)
+		if err != nil {
+			return err
+		}
+		for id, md := range mds {
+			if err := printBlock(id, md); err != nil {
+				return err
 			}
-			return printBlock(id)
-		})
+		}
+
+		return nil
 	}
 }
 
@@ -276,22 +280,8 @@ func registerBucketInspect(m map[string]setupFunc, root *kingpin.CmdClause, name
 		defer cancel()
 
 		// Getting Metas.
-		var blockMetas []*metadata.Meta
-		if err = bkt.Iter(ctx, "", func(name string) error {
-			id, ok := block.IsBlockDir(name)
-			if !ok {
-				return nil
-			}
-
-			m, err := block.DownloadMeta(ctx, logger, bkt, id)
-			if err != nil {
-				return err
-			}
-
-			blockMetas = append(blockMetas, &m)
-
-			return nil
-		}); err != nil {
+		blockMetas, err := download(ctx, logger, bkt)
+		if err != nil {
 			return err
 		}
 
@@ -383,25 +373,21 @@ func refresh(ctx context.Context, logger log.Logger, bucketUI *ui.Bucket, durati
 	})
 }
 
-func download(ctx context.Context, logger log.Logger, bkt objstore.Bucket) (blocks []metadata.Meta, err error) {
+func download(ctx context.Context, logger log.Logger, bkt objstore.Bucket) (blocks []*metadata.Meta, err error) {
 	level.Info(logger).Log("msg", "synchronizing block metadata")
-
-	if err = bkt.Iter(ctx, "", func(name string) error {
-		id, ok := block.IsBlockDir(name)
-		if !ok {
-			return nil
-		}
-
-		meta, err := block.DownloadMeta(ctx, logger, bkt, id)
-		if err != nil {
-			return err
-		}
-
-		blocks = append(blocks, meta)
-		return nil
-	}); err != nil {
+	mdFetcher := block.NewMetadataFetcher(logger, defaultBlockMDFetchConcurrency, bkt)
+	mds, err := mdFetcher.Fetch(ctx)
+	if err != nil {
 		level.Error(logger).Log("err", err, "msg", "Failed to downloaded block metadata")
-		return blocks, err
+		return nil, err
+	}
+
+	for _, md := range mds {
+		if md == nil {
+			continue
+		}
+
+		blocks = append(blocks, md)
 	}
 
 	level.Info(logger).Log("msg", "downloaded blocks meta.json", "num", len(blocks))
--- a/pkg/block/block.go
+++ b/pkg/block/block.go
@@ -163,7 +163,7 @@ func deleteDir(ctx context.Context, logger log.Logger, bkt objstore.Bucket, dir 
 
 // DownloadMeta downloads only meta file from bucket by block ID.
 // TODO(bwplotka): Differentiate between network error & partial upload.
-func DownloadMeta(ctx context.Context, logger log.Logger, bkt objstore.Bucket, id ulid.ULID) (metadata.Meta, error) {
+func DownloadMeta(ctx context.Context, logger log.Logger, bkt objstore.BucketReader, id ulid.ULID) (metadata.Meta, error) {
 	rc, err := bkt.Get(ctx, path.Join(id.String(), MetaFilename))
 	if err != nil {
 		return metadata.Meta{}, errors.Wrapf(err, "meta.json bkt get for %s", id.String())
--- a/pkg/block/mdfetcher.go
+++ b/pkg/block/mdfetcher.go
@@ -1,0 +1,103 @@
+package block
+
+import (
+	"context"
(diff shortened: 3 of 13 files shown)


lx223:
The CI test fails because of go mod error. It doesn't seem related to my change. Is it a known issue?

bwplotka:
Fixes https://github.com/thanos-io/thanos/issues/1335

@lx223 can you rebase, we fixed this master go.mod error (:

pkg/block/mdfetcher.go
```diff
@@ -0,0 +1,103 @@
...
+		concurrencyLimit: concurrencyLimit,
+		rwLock:           &sync.RWMutex{},
+		cachedBlockMDs:   make(map[ulid.ULID]*metadata.Meta),
+		bkt:              bkt,
+	}
+}
+
+// Fetch wil return the mapping from block ID to its metadata that are currently stored in the bucket passed in.
+// If the corresponding value for a block ID key is nil, no valid metadta is retrieved successfully.
+func (f *MetadataFetcher) Fetch(ctx context.Context) (map[ulid.ULID]*metadata.Meta, error) {
```
bwplotka: Why fetcher and not e.g `Syncer`? Especially if we maintain `cachedBlockMDs` for efficiency?

bwplotka: Let's mention if it's goroutine safe  (: It looks like it is, the question is ... should it be?

I think it would be better if you would not make `Fetch/Sync` goroutine safe. First of all, I don't think it is needed. Secondly, with this, we can be more efficient in terms of memory. We can allow to reuse exactly the same `map[ulid.ULID]*metadata.Meta` (pass as an argument and return updated one). This might save some allocations as there can be millions of objects potentially in the bucket. 

What do you think? With this I am not sure if the struct is needed, maybe just Function is enough? Don't have a strong opinion.

pkg/block/mdfetcher.go
```diff
@@ -0,0 +1,103 @@
...
+	"github.com/go-kit/kit/log"
+	"github.com/go-kit/kit/log/level"
+	"github.com/oklog/ulid"
+	"github.com/thanos-io/thanos/pkg/block/metadata"
+	"github.com/thanos-io/thanos/pkg/objstore"
+)
+
+type MetadataFetcher struct {
+	cachedBlockMDs map[ulid.ULID]*metadata.Meta
+	rwLock         *sync.RWMutex
```
bwplotka: Usually we just name it `mu`. `rw` is part of type so no need to put type info into variable. (: 

```suggestion
	mu         *sync.RWMutex
```

pkg/block/mdfetcher.go
```diff
@@ -0,0 +1,103 @@
+package block
```
bwplotka: `syncer`?

Same for file `mdsyncer`?

pkg/block/mdfetcher.go
```diff
@@ -0,0 +1,103 @@
...
+type MetadataFetcher struct {
+	cachedBlockMDs map[ulid.ULID]*metadata.Meta
+	rwLock         *sync.RWMutex
+
+	concurrencyLimit int
+	logger           log.Logger
+	bkt              objstore.Bucket
+}
+
+func NewMetadataFetcher(lg log.Logger, concurrencyLimit int, bkt objstore.Bucket) *MetadataFetcher {
```
bwplotka:
```suggestion
func NewMetadataFetcher(logger log.Logger, concurrencyLimit int, bkt objstore.Bucket) *MetadataFetcher {
```
Super minor nit, but `logger` is consistent across codebase.

pkg/block/mdfetcher.go
bwplotka: Be aware of this https://github.com/thanos-io/thanos/pull/1408 that changes this code as well (somehow we need to sync both PRs) cc @povilasv

pkg/block/mdfetcher.go
```diff
@@ -0,0 +1,103 @@
...
+		logger:           lg,
+		concurrencyLimit: concurrencyLimit,
+		rwLock:           &sync.RWMutex{},
+		cachedBlockMDs:   make(map[ulid.ULID]*metadata.Meta),
+		bkt:              bkt,
+	}
+}
+
+// Fetch wil return the mapping from block ID to its metadata that are currently stored in the bucket passed in.
+// If the corresponding value for a block ID key is nil, no valid metadta is retrieved successfully.
```
bwplotka: > If the corresponding value for a block ID key is nil, no valid metadta is retrieved successfully.

1. `s/metadta/metadata`
2. Nice idea to delegate the no-meta-json blocks on caller side. I think it makes sense.

However, maybe two maps make sense here then? 
```
(metas map[ulid.ULID]*metadata.Meta, noMeta []ulid.ULID, err  error)
```
This will be easier to use on caller side I think.

pkg/block/mdfetcher.go
```diff
@@ -0,0 +1,103 @@
...
+func (f *MetadataFetcher) Fetch(ctx context.Context) (map[ulid.ULID]*metadata.Meta, error) {
+	blockIDs := make(chan ulid.ULID, f.concurrencyLimit)
+	defer close(blockIDs)
+	remoteBlockMDs := make(map[ulid.ULID]*metadata.Meta, len(f.cachedBlockMDs))
+	wg := &sync.WaitGroup{}
+
+	for i := 0; i < f.concurrencyLimit; i++ {
```
bwplotka: Will review body of this function when we resolve other comments first (: 

Looks good from first glance, looks like you covered all the edge cases, nice! Thanks for tests as well!

bwplotka:
Also we need CHANGELOG for this I think as we not handle partial uploads carefully on downsampling/retention apply

bwplotka:
Hi :wave: 

Can we finish this? (: If you don't have time, can I grab it and finish?

pkg/block/mdfetcher.go
lx223: I am happy to rename to syncer if you feel strongly about it. I am alright with both. I chose fetch as I felt it is more about fetching the metadata from remote source. Keeping a cache is just an internal optimisation and might become an issue and get removed if the bucket contains too many objects.

Good point about about the memory allocation. I have a different perspective. I don't think it would add that much to memory as it is using metadata pointers right now. And if we were to reuse the same map, there would still have to be a map to keep what metadata are in the bucket in order to remove those cached but somehow-removed-from-remote-source ones. So the saving won't be that much. Besides, I think streaming metadata might be a better solution to scale with bigger number of objects in the bucket. What do you think?

pkg/block/mdfetcher.go
lx223: sorry, not following about point 1. what is `s/metadta/metadata`?

I thought about whether to distinguish those two cases. For the places I have replaced, I haven't found a case that would require these two being distinguished. Did I miss a case that would require this?

lx223:
@bwplotka  sorry about that! 

I'll change `pkg/store/bucket.go` and add a changelog tomorrow.

pkg/block/mdfetcher.go
bwplotka: > sorry, not following about point 1. what is s/metadta/metadata?

https://www.tutorialspoint.com/unix/unix-regular-expressions.htm

(:

bwplotka:
@lx223 Are you on CNCF slack? Join us so we can discuss quicker offline!

I would like to see this change ASAP as it will help with: https://github.com/thanos-io/thanos/issues/1331

pkg/block/mdfetcher.go
bwplotka: Hm, what about the case where we remove partially uploaded blocks? (noMeta + delay):

https://github.com/thanos-io/thanos/blob/865d5ec710b5651ead08a7dc73864a6f5dbeeaa9/pkg/compact/compact.go#L197

I think that's when noMeta is useful.

pkg/block/mdfetcher.go
bwplotka: No strong feeling to the naming. It really depends on what exactly interface we are going to apply as you are exploring streaming suggestion now (: So then `ForeachMeta` might make sense more?

>  I think streaming metadata might be a better solution to scale with bigger number of objects in the bucket. What do you think?

Definitely, however, this means we would need to revert to any of Iter concepts like

* passing a function to invoke for each meta
* `Iter` interface: 
```
// Iterator is a simple iterator that can only get the next value.
type Iterator interface {
	At() metadata.Meta)
	Err() error
	Next() bool
}
```

bwplotka:
Would love to see this work to be done as this would enormously help with dynamic system where components are restarted in different moments causing partial uploads to be very common (: Let me know if I can help with this and I would love to prioritize this.  @lx223 

I listed all the places were we can unify our sync/fetch logic.

bwplotka:
Relaed https://github.com/thanos-io/thanos/pull/1525

pkg/block/mdfetcher.go
lx223: I might be wrong here but I think this is the only case that this required and it is covered in the caller's logic - `thanos/pkg/compact/compact.go`. But it does sound like something useful in future. Tempted to add when we are seeing more such usages.

pkg/block/mdfetcher.go
lx223: resolved as it is related to the name discussion.

lx223:
@bwplotka just joined CNCF slack and have made changes to all those places you listed. PTAL

pkg/block/mdfetcher.go
lx223: I think we don't have to use stream for now unless the OOM issue is already big and urgent. I think we can keep this signature which meshes pretty well with the current usage and we can use this as the foundation to identify and make changes when we are ready to explore other options like streaming.

cmd/thanos/bucket.go
```diff
@@ -205,39 +207,41 @@ func registerBucketLs(m map[string]setupFunc, root *kingpin.CmdClause, name stri
...
 				fmt.Fprintln(os.Stdout, "")
 				return nil
 			}
 		}
 
-		return bkt.Iter(ctx, "", func(name string) error {
-			id, ok := block.IsBlockDir(name)
-			if !ok {
-				return nil
+		mdFetcher := block.NewMetadataFetcher(logger, defaultBlockMDFetchConcurrency, bkt)
```
lx223: didn't switch to the `download` function because the original logic suggests it requires block id even for partially uploaded blocks.

bwplotka:
Should we get back to this @lx223 ? (:

lx223:
@bwplotka sure :D WDYT about the current code? there is some rebasing to do. But, it probably won't change the main logic.

cmd/thanos/bucket.go
bwplotka: :+1:

pkg/block/block.go
```diff
@@ -163,7 +163,7 @@ func deleteDir(ctx context.Context, logger log.Logger, bkt objstore.Bucket, dir
 
 // DownloadMeta downloads only meta file from bucket by block ID.
 // TODO(bwplotka): Differentiate between network error & partial upload.
-func DownloadMeta(ctx context.Context, logger log.Logger, bkt objstore.Bucket, id ulid.ULID) (metadata.Meta, error) {
+func DownloadMeta(ctx context.Context, logger log.Logger, bkt objstore.BucketReader, id ulid.ULID) (metadata.Meta, error) {
```
bwplotka: :+1:

pkg/block/mdfetcher.go
```diff
@@ -0,0 +1,103 @@
...
+
+	f.mu.RLock()
+	existingMD, exist := f.cachedBlockMDs[blockID]
+	f.mu.RUnlock()
+
+	if exist {
+		return existingMD, nil
+	}
+
+	md, err := DownloadMeta(ctx, f.logger, f.bkt, blockID)
```
bwplotka: Wonder if we can leverage the file system as a cache, instead of downloading and keep in memory? Let's maybe add TODO at least (: It saves some bandwidth (?) on startup.

pkg/block/mdfetcher.go
bwplotka: Yea, I think I am happy with this. We already keep `cached block MD` anyway.

bwplotka: fetcher might be fine then

cmd/thanos/bucket.go
```diff
@@ -383,25 +373,21 @@ func refresh(ctx context.Context, logger log.Logger, bucketUI *ui.Bucket, durati
 	})
 }
 
-func download(ctx context.Context, logger log.Logger, bkt objstore.Bucket) (blocks []metadata.Meta, err error) {
+func download(ctx context.Context, logger log.Logger, bkt objstore.Bucket) (blocks []*metadata.Meta, err error) {
 	level.Info(logger).Log("msg", "synchronizing block metadata")
+	mdFetcher := block.NewMetadataFetcher(logger, defaultBlockMDFetchConcurrency, bkt)
+	mds, err := mdFetcher.Fetch(ctx)
+	if err != nil {
+		level.Error(logger).Log("err", err, "msg", "Failed to downloaded block metadata")
```
bwplotka: double error handling, please choose one

pkg/block/mdfetcher.go
```diff
@@ -0,0 +1,103 @@
...
+	defer close(blockIDs)
+	remoteBlockMDs := make(map[ulid.ULID]*metadata.Meta, len(f.cachedBlockMDs))
+	wg := &sync.WaitGroup{}
+
+	for i := 0; i < f.concurrencyLimit; i++ {
+		go func() {
+			for blockID := range blockIDs {
+				md, err := f.fetchBlockMDIfNotCached(ctx, wg, blockID)
+				if err != nil {
+					level.Debug(f.logger).Log("msg", "failed to fetch block metadata", "block", blockID, "error", err)
```
bwplotka: warning?

pkg/store/bucket.go
```diff
@@ -295,59 +294,38 @@ func (s *BucketStore) Close() (err error) {
...
 		}
-		return nil
-	})
-
-	close(blockc)
-	wg.Wait()
 
-	if err != nil {
-		return errors.Wrap(err, "iter")
+		if err := s.addBlock(ctx, blockID); err != nil {
```
bwplotka: This is important as we download meta.json twice. Once to memory once to disk... Can we improve this?

pkg/block/mdfetcher.go
```diff
@@ -0,0 +1,103 @@
...
+		case blockIDs <- id:
+		}
+
+		return nil
+	}); err != nil {
+		return nil, err
+	}
+	wg.Wait()
+
+	f.mu.Lock()
```
bwplotka: Why this lock? To make fetcher thread-safe? I think we need single flight here TBH with some expiry even (: Can you add `TODO(bwplotka): Add single flight with expiry if needed`

Also, I am not sure if we can really reuse this lock IMO...  I would rather have:
either 
* two locks, one for concurrency here and second for potential concurrent use of `Fetch` 
* remove this lock. I don't think we do currently fetch concurrently so we could remove this and comment that it's not thread safe

I would say let's add `TODO`

pkg/block/mdfetcher.go
squat: I think we should collect and/or exit on errors. Currently, if there was a transient network error we'll cache `nil` (negative result) and then never re-try downloading that metadata.

pkg/block/mdfetcher.go
```diff
@@ -0,0 +1,103 @@
...
+	for i := 0; i < f.concurrencyLimit; i++ {
+		go func() {
+			for blockID := range blockIDs {
+				md, err := f.fetchBlockMDIfNotCached(ctx, wg, blockID)
+				if err != nil {
+					level.Debug(f.logger).Log("msg", "failed to fetch block metadata", "block", blockID, "error", err)
+				}
+
+				f.mu.Lock()
+				remoteBlockMDs[blockID] = md
```
squat: I think there is a race currently: we call `wg.Done` before inserting the returned values in the map, meaning `wg.Done()` may return early and we may return an incomplete map to the caller.

bwplotka:
@lx223 let us know if you have time to finish this. This fixes quite annoying issue, plus @squat already touching this path for compactor syncer. 

If not we are happy to continue this for you (:

lx223:
@bwplotka  sry, might have enough time to finish it fast. Feel free to take it. :)

bwplotka:
Cool, closing this for now then, updating the issue.

PR #1395, opened 2019-08-10
feat sicedar: added readiness prober

FUSAKLA:
Signed-off-by: Martin Chodur <m.chodur@seznam.cz>

## Changes

_Next PR in the from the series of adding readiness and liveness probes._

Added readiness and liveness prober to the thanos sidecar.
It follows the logic of setting the `promUp` metric so when it successfully fetches config of the Prometheus it becomes ready.

The question is should it take into account status of the GRPC interface also?

## Verification

1.  started sidecar without running the Prometheus instance results in 
```
$ curl http://0.0.0.0:10902/-/ready
thanos sidecar is not ready. Reason: request flags against http://localhost:9090/api/v1/status/config: Get http://localhost:9090/api/v1/status/config: dial tcp 127.0.0.1:9090: connect: connection refused
``` 
2.  started the prometheus sidecar becomes ready
```
$ curl http://0.0.0.0:10902/-/ready
thanos sidecar is ready
```

3. stopped the prometheus instance again sidecar goes again to not ready state after max 30s which is the sidecar check interval
```
$ curl http://0.0.0.0:10902/-/ready
thanos sidecar is not ready. Reason: request flags against http://localhost:9090/api/v1/status/config: Get http://localhost:9090/api/v1/status/config: dial tcp 127.0.0.1:9090: connect: connection refused
```

### DIFF
--- a/cmd/thanos/sidecar.go
+++ b/cmd/thanos/sidecar.go
@@ -8,8 +8,6 @@ import (
 	"sync"
 	"time"
 
-	"github.com/thanos-io/thanos/pkg/prober"
-
 	"github.com/go-kit/kit/log"
 	"github.com/go-kit/kit/log/level"
 	"github.com/oklog/run"
@@ -21,6 +19,7 @@ import (
 	"github.com/thanos-io/thanos/pkg/block/metadata"
 	"github.com/thanos-io/thanos/pkg/component"
 	"github.com/thanos-io/thanos/pkg/objstore/client"
+	"github.com/thanos-io/thanos/pkg/prober"
 	"github.com/thanos-io/thanos/pkg/promclient"
 	"github.com/thanos-io/thanos/pkg/reloader"
 	"github.com/thanos-io/thanos/pkg/runutil"
@@ -34,8 +33,7 @@ import (
 const waitForExternalLabelsTimeout = 10 * time.Minute
 
 func registerSidecar(m map[string]setupFunc, app *kingpin.Application) {
-	comp := component.Sidecar
-	cmd := app.Command(comp.String(), "sidecar for Prometheus server")
+	cmd := app.Command(component.Sidecar.String(), "sidecar for Prometheus server")
 
 	grpcBindAddr, httpBindAddr, cert, key, clientCA := regCommonServerFlags(cmd)
 
@@ -57,7 +55,7 @@ func registerSidecar(m map[string]setupFunc, app *kingpin.Application) {
 
 	uploadCompacted := cmd.Flag("shipper.upload-compacted", "[Experimental] If true sidecar will try to upload compacted blocks as well. Useful for migration purposes. Works only if compaction is disabled on Prometheus.").Default("false").Hidden().Bool()
 
-	m[comp.String()] = func(g *run.Group, logger log.Logger, reg *prometheus.Registry, tracer opentracing.Tracer, _ bool) error {
+	m[component.Sidecar.String()] = func(g *run.Group, logger log.Logger, reg *prometheus.Registry, tracer opentracing.Tracer, _ bool) error {
 		rl := reloader.New(
 			log.With(logger, "component", "reloader"),
 			reloader.ReloadURLFromBase(*promURL),
@@ -80,7 +78,7 @@ func registerSidecar(m map[string]setupFunc, app *kingpin.Application) {
 			objStoreConfig,
 			rl,
 			*uploadCompacted,
-			comp,
+			component.Sidecar,
 		)
 	}
 }
--- a/cmd/thanos/sidecar.go
+++ b/cmd/thanos/sidecar.go
@@ -187,10 +187,8 @@ func runSidecar(
 				if err := m.UpdateLabels(iterCtx, logger); err != nil {
 					level.Warn(logger).Log("msg", "heartbeat failed", "err", err)
 					promUp.Set(0)
-					statusProber.SetNotReady(err)
 				} else {
 					promUp.Set(1)
-					statusProber.SetReady()
 					lastHeartbeat.Set(float64(time.Now().UnixNano()) / 1e9)
 				}
 
--- a/cmd/thanos/sidecar.go
+++ b/cmd/thanos/sidecar.go
@@ -8,10 +8,12 @@ import (
 	"sync"
 	"time"
 
+	"github.com/thanos-io/thanos/pkg/prober"
+
 	"github.com/go-kit/kit/log"
 	"github.com/go-kit/kit/log/level"
 	"github.com/oklog/run"
-	opentracing "github.com/opentracing/opentracing-go"
+	"github.com/opentracing/opentracing-go"
 	"github.com/pkg/errors"
 	"github.com/prometheus/client_golang/prometheus"
 	"github.com/prometheus/common/model"
@@ -26,13 +28,14 @@ import (
 	"github.com/thanos-io/thanos/pkg/store"
 	"github.com/thanos-io/thanos/pkg/store/storepb"
 	"google.golang.org/grpc"
-	kingpin "gopkg.in/alecthomas/kingpin.v2"
+	"gopkg.in/alecthomas/kingpin.v2"
 )
 
 const waitForExternalLabelsTimeout = 10 * time.Minute
 
-func registerSidecar(m map[string]setupFunc, app *kingpin.Application, name string) {
-	cmd := app.Command(name, "sidecar for Prometheus server")
+func registerSidecar(m map[string]setupFunc, app *kingpin.Application) {
+	comp := component.Sidecar
+	cmd := app.Command(comp.String(), "sidecar for Prometheus server")
 
 	grpcBindAddr, httpBindAddr, cert, key, clientCA := regCommonServerFlags(cmd)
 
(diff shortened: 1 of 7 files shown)


cmd/thanos/sidecar.go
```diff
@@ -8,10 +8,12 @@ import (
 	"sync"
 	"time"
 
+	"github.com/thanos-io/thanos/pkg/prober"
+
```
povilasv: remove the space :)

cmd/thanos/sidecar.go
```diff
@@ -26,13 +28,14 @@ import (
...
-	kingpin "gopkg.in/alecthomas/kingpin.v2"
+	"gopkg.in/alecthomas/kingpin.v2"
 )
 
 const waitForExternalLabelsTimeout = 10 * time.Minute
 
-func registerSidecar(m map[string]setupFunc, app *kingpin.Application, name string) {
-	cmd := app.Command(name, "sidecar for Prometheus server")
+func registerSidecar(m map[string]setupFunc, app *kingpin.Application) {
+	comp := component.Sidecar
```
bwplotka: I would maybe inline those, would be more verbose.

cmd/thanos/sidecar.go
```diff
@@ -176,8 +189,10 @@ func runSidecar(
 				if err := m.UpdateLabels(iterCtx, logger); err != nil {
 					level.Warn(logger).Log("msg", "heartbeat failed", "err", err)
 					promUp.Set(0)
+					readinessProber.SetNotReady(err)
```
bwplotka: Wonder if it's ready or rather healthy?

cmd/thanos/sidecar.go
```diff
@@ -117,6 +122,12 @@ func runSidecar(
 		uploads = false
 	}
 
+	readinessProber := prober.NewProber(comp, logger, prometheus.WrapRegistererWithPrefix("thanos_", reg))
```
bwplotka: it is actually both health and ready probe, right?  So can we remove all `readiness` mentions here?

cmd/thanos/sidecar.go
bwplotka: From: https://kubernetes.io/docs/tasks/configure-pod-container/configure-liveness-readiness-probes/
Liveness/Healthiness:
```
Many applications running for long periods of time eventually transition to broken states, and cannot recover except by being restarted. Kubernetes provides liveness probes to detect and remedy such situations.
```
Readiness:
```
Sometimes, applications are temporarily unable to serve traffic. For example, an application might need to load large data or configuration files during startup, or depend on external services after startup. In such cases, you don’t want to kill the application, but you don’t want to send it requests either. Kubernetes provides readiness probes to detect and mitigate these situations. A pod with containers reporting that they are not ready does not receive traffic through Kubernetes Services.
```
I would add those definitions to probe `SetReady` and `SetHealth` to be clear, maybe?

I believe we can set healthiness instead of readiness here, but also not sure how crash will help here.... (: More over I would retry couple of times first (e.g 3) to ensure we are not flaky. Maybe after 3rd heartbeat fail we should set to not ready / not healthy?

cmd/thanos/sidecar.go
FUSAKLA: Ok, I wasn't sure if name `prober` is understandable on it's own but you are right it's both even liveness. Will rename.

cmd/thanos/sidecar.go
FUSAKLA: I'd say the sidecar can easily recover from this state when the Prometheus instance goes back up or becomes responsible again. As you said, I'm not sure either if the restart helps here somehow. But probably some issue where the restart would help could occur. 

Maybe set it to not ready right after first fail and unhealthy after for example as you suggested 3 failed health-checks in a row or bit more? Personally I'd wait at least 2m for example? That could be sufficient for even bigger instance of Prometheus to restart.

Adding the comments sounds as a good idea, will do.

cmd/thanos/sidecar.go
FUSAKLA: should be gone now, thanks :)

cmd/thanos/sidecar.go
FUSAKLA: it collided with the package name as a just `prober` so changed it to `statusProber` if that is ok with you?

cmd/thanos/sidecar.go
FUSAKLA: Changed to `NotReady` right after health-check on Prometheus fails and `NotHealthy` after 3 minutes of consequential fails of the health-check. I wanted to use the number of failed health-checks but the check interval is different on startup and later when the sidecar is already running so I stick to the time to be consistent.

Does this make sense to you? I'm bit afraid to set it unhealthy right after one fail.

cmd/thanos/sidecar.go
FUSAKLA: I inlined it in the `registerSidecar` function, not sure if you want to inline it everywhere even in the `runSidecar` ?

FUSAKLA:
@povilasv thanks! there are some new changes just so you don't approve something you didn't see so maybe check that out :)

@bwplotka   Thanks for the suggestions! I tried some refactor of the sidecar liveness so PTAL.
Also I applied your other suggestions on the compactor also since the changes are the same as in the previous PR.

FUSAKLA:
those failing test are weird, not sure if related to my changes?

cmd/thanos/sidecar.go
povilasv: I would try to avoid this logic, as this is easily implementable in Kubernetes via:
```
timeoutSeconds
failureThreshold
initialDelaySeconds
```
Docs: https://kubernetes.io/docs/tasks/configure-pod-container/configure-liveness-readiness-probes/#configure-probes


Let's just make it simple.

So if there is no connection to Prometheus, we can't serve traffic, so it's definetely *NotReady*.

Now is it not healthy? IMO as we keep retrying for prometheus, we are healthy, but Prometheus isn't. So Prometheus healthchecks should restart it and make sure it's there and healthy, not sidecar's.

@bwplotka / @FUSAKLA  WDYT?

cmd/thanos/sidecar.go
```diff
@@ -139,6 +148,12 @@ func runSidecar(
 				}
 			}
 
+			// When the heartbeat to Prometheus fails, the sidecar is marked as not ready.
+			// But after `sinceLastSuccessfulHeartbeatLimit` duration of consequential fails it's marked also not healthy,
+			// so the orchestrator (if any) can try restarting it if it would help.
+			sinceLastSuccessfulHeartbeat := 0 * time.Minute
```
bwplotka: Thanks for those variables and comments. Super clear now :+1: 

However, are we sure about the value? If we cannot connect to Prometheus, we cannot use remote read either so I guess it make sense? (:

bwplotka: I think those values are quite strict, but also user can configure values on probe side just in case. https://kubernetes.io/docs/tasks/configure-pod-container/configure-liveness-readiness-probes/#configure-probes

cmd/thanos/sidecar.go
```diff
@@ -148,6 +163,10 @@ func runSidecar(
 						"err", err,
 					)
 					promUp.Set(0)
+					statusProber.SetNotReady(err)
+					if sinceLastSuccessfulHeartbeat >= sinceLastSuccessfulHeartbeatLimit {
+						statusProber.SetNotHealthy(err)
```
bwplotka: For initial retry we either need higher limit or just let's not do it maybe (?)

Essentially you can configure easily sidecar without any hierarchy in Kuberentes - this means that this can retry for some time, as Prometheus container is starting up maybe.. I think 5m or more even if you have slowness on startup.. 

I just want to be careful to not block anyone with this (:

pkg/prober/prober.go
```diff
@@ -37,6 +37,19 @@ type Prober struct {
 func NewProber(component component.Component, logger log.Logger, reg prometheus.Registerer) *Prober {
 	initialErr := fmt.Errorf(initialErrorFmt, component)
 
+	// From Kubernetes documentation https://kubernetes.io/docs/tasks/configure-pod-container/configure-liveness-readiness-probes/ :
```
bwplotka: Nice!

cmd/thanos/sidecar.go
bwplotka: Did not see that comment, sorry. will respond in more visible thread ;p

bwplotka:
Let's chat offline @FUSAKLA but overall problem with those probes is that we need to be careful. @povilasv mentioned that all of this can be configured anyone on Kubernetes side, but that's not that simple. Problems:

1. You can control those things.. timeouts, thresholds etc in Kubernetes, but maybe not in thousands of other solutions/load balancers
2. You can control those things... But only per container. This means that if you have 4 different cases in the application (like sidecar) when you make something not ready or not healthy.... You cannot do it yolo and ask someone to configure something on top. As they can only adjust all cases in same time there is No granularity.

bwplotka:
Regarding healthyness: https://twitter.com/miekg/status/1152911058627153920 

We are I think on the same page that we should do it ONLY if we are sure that restart will help.

FUSAKLA:
Yeah, so should we get back to the original state of just modifying the readiness based on the halthchecks? 

I wanted originally to avoid touching the liveness probe because it felt bit harsh but I understood that this was what you suggested. But I probably did not clearly understood what you meant, sorry for the confusement :confused:

FUSAKLA:
Along with the offline agreement, the readiness probe is only set after the initial Prometheus external labels fetch from the Prometheus. Once they are fetched it stays ready all the time.

If we hit any case in the future when more complex readiness handling will be needed it can be added then.

FUSAKLA:
@bwplotka this is done in the `defaultHTTPListener` constructor where the mux is started actually if that is ok?
https://github.com/thanos-io/thanos/blob/0a30a2e20881cc7d364437974cc13182aa0abf83/cmd/thanos/main.go#L362-L369

FUSAKLA:
But I found one bug with opposite condition on error logging in the Prober, fixed in https://github.com/thanos-io/thanos/pull/1395/commits/60897a8ee0008520a529e5abe8b47efe457a17c6  :roll_eyes:

pkg/prober/prober.go
bwplotka: Minor nit but this should part of struct TBH (for GoDocs)

FUSAKLA: good point, is it ok this way?

bwplotka:
Thanks @FUSAKLA !

FUSAKLA:
Great, thank you all for the comments and going through the discussions around the readiness!

There is only few components left :smile: 
If you have the time and will discuss another one, then I made next PR for the store https://github.com/thanos-io/thanos/pull/1460

PR #1406, opened 2019-08-13
shipper: add missing increment of uploadFailures metric

zhulongcheng:


pkg/shipper/shipper.go
```diff
@@ -326,6 +326,7 @@ func (s *Shipper) Sync(ctx context.Context) (uploaded int, err error) {
 		return nil
 	}); err != nil {
 		s.metrics.dirSyncFailures.Inc()
+		s.metrics.uploadFailures.Add(float64(uploadErrs))
```
povilasv: This won't really work as we return `nil` in most of the cases when we increment `uploadErrs`.

consider just incrementing the counter in those places.

zhulongcheng: OK.

Thanks @povilasv

PR #1408, opened 2019-08-13
Store: Add Time & duration based partitioning

povilasv:
## Changes

New start of https://github.com/thanos-io/thanos/pull/1077

## Verification

Multiple tests added.


cmd/thanos/store.go
```diff
@@ -49,7 +50,19 @@ func registerStore(m map[string]setupFunc, app *kingpin.Application, name string
...
 		Default("20").Int()
 
+	minTime := model.TimeOrDuration(cmd.Flag("min-time", "Start of time range limit to serve. Thanos Store serves only metrics, which happened later than this value. Option can be a constant time in RFC3339 format or time duration relative to current time, such as -1d or 2h45m. Valid duration units are ms, s, m, h, d, w, y.").
+		Default("0000-01-01T00:00:00Z"))
+
+	maxTime := model.TimeOrDuration(cmd.Flag("max-time", "End of time range limit to serve. Thanos Store serves only blocks, which happened eariler than this value. Option can be a constant time in RFC3339 format or time duration relative to current time, such as -1d or 2h45m. Valid duration units are ms, s, m, h, d, w, y.").
+		Default("9999-12-31T23:59:59Z"))
+
 	m[name] = func(g *run.Group, logger log.Logger, reg *prometheus.Registry, tracer opentracing.Tracer, debugLogging bool) error {
+		// Sanity check Time filters
```
bwplotka: missing trailing period

bwplotka: Also comment not needed IMO

pkg/store/bucket.go
```diff
@@ -309,6 +319,17 @@ func (s *BucketStore) SyncBlocks(ctx context.Context) error {
 		if err != nil {
 			return nil
 		}
+
+		inRange, err := s.isBlockInMinMaxRange(ctx, id)
```
bwplotka: I think the problem here is that we hit the issue of partial blocks... Something that https://github.com/thanos-io/thanos/pull/1394 touches. 

I thought we actually handle this in store gw properly (blocks without meta are excluded), seems not so not a regression.

In any case we will just warn so :+1:

pkg/store/bucket.go
```diff
@@ -377,6 +398,31 @@ func (s *BucketStore) numBlocks() int {
 	return len(s.blocks)
 }
 
+func (s *BucketStore) isBlockInMinMaxRange(ctx context.Context, id ulid.ULID) (bool, error) {
+	dir := filepath.Join(s.dir, id.String())
+
+	b := &bucketBlock{
```
bwplotka: Allocating the whole block for quick check is not great. I think we should just extract `loadMeta` method to separate function and user.

Also at some point we need some meta syncer/ matcher that deals with all of this, something similar to https://github.com/thanos-io/thanos/pull/1394 ): 

Also even with time slicing looks like we will download all meta.json files to disk. Is that what we want? Essentially time slices will not include meta.json files on disk.

pkg/store/bucket.go
```diff
@@ -482,6 +532,26 @@ func (s *BucketStore) Info(context.Context, *storepb.InfoRequest) (*storepb.Info
 	}, nil
 }
 
+func (s *BucketStore) normalizeMinTime(mint int64) int64 {
```
bwplotka: Can we rename it a bit? `normalize` does not tell much, I think we rather mean `limit` maybe?

pkg/store/bucket_e2e_test.go
```diff
@@ -433,3 +453,79 @@ func TestBucketStore_ManyParts_e2e(t *testing.T) {
...
+	ctx, cancel := context.WithCancel(context.Background())
+	defer cancel()
+	bkt := inmem.NewBucket()
+
+	dir, err := ioutil.TempDir("", "test_bucket_time_part_e2e")
+	testutil.Ok(t, err)
+	defer func() { testutil.Ok(t, os.RemoveAll(dir)) }()
+
+	series := []labels.Labels{
+		labels.FromStrings("a", "1", "b", "1"),
```
bwplotka: Can we reuse `prepareStoreWithTestBlocks`? Series looks exactly the same.

pkg/store/bucket.go
```diff
@@ -468,6 +514,10 @@ func (s *BucketStore) TimeRange() (mint, maxt int64) {
 			maxt = b.meta.MaxTime
 		}
 	}
+
+	mint = s.normalizeMinTime(mint)
```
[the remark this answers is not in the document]
bwplotka: but we do it in 2 places.. I think we just need to rename to something clear.

cmd/thanos/store.go
```diff
@@ -49,7 +50,19 @@ func registerStore(m map[string]setupFunc, app *kingpin.Application, name string
 	blockSyncConcurrency := cmd.Flag("block-sync-concurrency", "Number of goroutines to use when syncing blocks from object storage.").
 		Default("20").Int()
 
+	minTime := model.TimeOrDuration(cmd.Flag("min-time", "Start of time range limit to serve. Thanos Store serves only metrics, which happened later than this value. Option can be a constant time in RFC3339 format or time duration relative to current time, such as -1d or 2h45m. Valid duration units are ms, s, m, h, d, w, y.").
+		Default("0000-01-01T00:00:00Z"))
+
+	maxTime := model.TimeOrDuration(cmd.Flag("max-time", "End of time range limit to serve. Thanos Store serves only blocks, which happened eariler than this value. Option can be a constant time in RFC3339 format or time duration relative to current time, such as -1d or 2h45m. Valid duration units are ms, s, m, h, d, w, y.").
```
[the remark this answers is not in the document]
bwplotka: I think that makes sense to me.

bwplotka:
Sure, but not for this PR (:

pkg/store/bucket_e2e_test.go
povilasv: I splitted the `prepareTestBlocks` from `prepareStoreWithTestBlocks` so that we can easily configure Store and generate test blocks.

I tried to adjust `prepareStoreWithTestBlocks`, but I didn't like the newly added parameter for filtering and it would need quite a lot of changes due to changes in sync blocks https://github.com/thanos-io/thanos/blob/master/pkg/store/bucket_e2e_test.go#L155-L159

IMO that func is awesome for https://github.com/thanos-io/thanos/blob/master/pkg/store/bucket_e2e_test.go#L365, but doesn't really fit my use case, as I need custom checks in Store gw

pkg/store/bucket.go
povilasv: Good question, IMO those meta.json files are small in size, so why not? It will speed up our next meta.json lookup.

povilasv:
> Guys, what do you think about implementing the same in the sidecar?

For Sidecar I think we have a similar PR open https://github.com/thanos-io/thanos/pull/1267

pkg/store/bucket.go
```diff
@@ -377,6 +398,26 @@ func (s *BucketStore) numBlocks() int {
...
 
+func (s *BucketStore) isBlockInMinMaxRange(ctx context.Context, id ulid.ULID) (bool, error) {
+	dir := filepath.Join(s.dir, id.String())
+
+	err, meta := loadMeta(ctx, s.logger, s.bucket, dir, id)
+	if err != nil {
+		return false, err
+	}
+
+	// We check for blocks in configured minTime, maxTime range
```
bwplotka: missing trailing period

pkg/store/bucket_e2e_test.go
bwplotka: fair

cmd/thanos/store.go
```diff
@@ -49,7 +50,18 @@ func registerStore(m map[string]setupFunc, app *kingpin.Application, name string
 	blockSyncConcurrency := cmd.Flag("block-sync-concurrency", "Number of goroutines to use when syncing blocks from object storage.").
 		Default("20").Int()
 
+	minTime := model.TimeOrDuration(cmd.Flag("min-time", "Start of time range limit to serve. Thanos Store serves only metrics, which happened later than this value. Option can be a constant time in RFC3339 format or time duration relative to current time, such as -1d or 2h45m. Valid duration units are ms, s, m, h, d, w, y.").
```
bwplotka: `Default("0000-01-01T00:00:00Z"))` Is that lowest min time? What about `-9999-12-31T23:59:59Z` ?

povilasv:
```
    t, err = time.Parse(time.RFC3339, "-9999-12-31T23:59:59Z")
    if err != nil {
        fmt.Println(err)
    }
    fmt.Println(t)
```

gives:
```
parsing time "-9999-12-31T23:59:59Z" as "2006-01-02T15:04:05Z07:00": cannot parse "-9999-12-31T23:59:59Z" as "2006"        
```

bwplotka: I would set some explicit default value here (for `""` value), but we can do it in next PRs (:

PR #1380, opened 2019-08-07
Upgraded Prometheus to 2.11.1 and TSDB to 2.9.1

bwplotka:
Signed-off-by: Bartek Plotka <bwplotka@gmail.com>

Issue #1271, opened 2019-06-21
query: metric autocomplete drop down inserts wrong metric


jjneely:
**Thanos, Prometheus and Golang version used**

Thanos v0.5.0
Go: go1.12.6

**What happened**

When working with the Graph UI from the Query web interface, the PromQL entry textarea will suggest metric names after you type in several letters.

For example, typing in "node_memory" (presuming node-exporter metrics are scraped somewhere) will produce a drop down with many/all possible completions.  Selecting one fills in that completed metric name in the PromQL textarea.

However, (hard to show in pictures) if you have a long list of results and scroll or use PageDown to find the right metric, selecting that metric will insert a different metric name in the textarea.  Not the metric name that was selected.

**What you expected to happen**

The metric selected should be inserted into the PromQL text area.

**How to reproduce it (as minimally and precisely as possible)**:

In the PromQL text area type in "node_", scroll or PageDown to find a suggested metric toward the bottom of the list.  Select it.  Note that a different metric may have been inserted.

jjneely:
I think this is a Prometheus bug, and we just import the code.  Ugh.

bwplotka:
)): Worth to put this bug on Prometheus then?

Thanks for spotting - must be quite annoying.

jojohappy:
Hi, I'm interested in this problem. Let me try to reproduce it both of thanos and Prometheus.

jojohappy:
@jjneely @bwplotka I have deep dive into the code of `typeahead`, I guess that it is a bug on the [Bootstrap-3-Typeahead](https://github.com/bassjobsen/Bootstrap-3-Typeahead), meanwhile, it would be fine in Firefox 67.0.4 and Safari 12.1.1 in my laptop, but Chrome 74.0.3729.169 would hit the bug ): , I have submitted a [issue](https://github.com/bassjobsen/Bootstrap-3-Typeahead/issues/375) and will wait for response. So you can use Firefox or Safari instead.

PR #1409, opened 2019-08-14
pkg/ui: Upgrade bootstrap typeahead to v4.0.2

brancz:
## Changes

This PR updates bootstrap typeahead to v4.0.2

## Verification

Tested that typeahead still works.

@bwplotka @povilasv @GiedriusS @domgreen


Makefile
```diff
@@ -7,8 +7,6 @@ TMP_GOPATH        ?= /tmp/thanos-go
 GOBIN             ?= ${GOPATH}/bin
 GO111MODULE       ?= on
 export GO111MODULE
-GOPROXY           ?= https://proxy.golang.org
```
GiedriusS: Why are we removing the proxy here? (:

brancz: Otherwise we have a failing build, the same reason why we had to update to a newer tsdb here: https://github.com/thanos-io/thanos/pull/1380

But I didn't want to update to a newer tsdb for a patch release.

jojohappy:
@brancz could this PR solve the issue #1271?

Makefile
bwplotka: We need that. It's also only for 0.6.1

PR #1416, opened 2019-08-14
Add XING to adopters

boosty:
We are happily adding xing.com as an adopter 😃 
Thanks for this great project!


boosty:
Done ✅

boosty:
Thanks a lot!

Issue #1415, opened 2019-08-14
Freshly cloned Thanos now fails to build with `make build`


bwplotka:
Only `publish_master` does not work though: https://circleci.com/gh/thanos-io/thanos/4467

I can repro locally:
```
>> building binaries /home/bartek/Repos/thanos/.bin
 >   thanos
build github.com/thanos-io/thanos/cmd/thanos: cannot load github.com/prometheus/prometheus/tsdb: cannot find module providing package github.com/prometheus/prometheus/tsdb
!! command failed: build -o /home/bartek/Repos/thanos/thanos -ldflags -X github.com/prometheus/common/version.Version=0.6.0-master -X github.com/prometheus/common/version.Revision=e9028159404dc753e5e1fa4443ae135d72f09361 -X github.com/prometheus/common/version.Branch=publish-fix -X github.com/prometheus/common/version.BuildUser=bartek@bartek-lin-fgzbqq2.corp.improbable.io -X github.com/prometheus/common/version.BuildDate=20190814-14:53:04  -extldflags '-static' -a -tags netgo github.com/thanos-io/thanos/cmd/thanos: exit status 1
Makefile:110: recipe for target 'build' failed
make: *** [build] Error 1

```
Investigating.

bwplotka:
Somehow `go.mod` was reverted.

imluyuan:
[root@pfw06vv thanos]# make build
>> building binaries /bin
 >   thanos
# github.com/prometheus/tsdb/goversion
/root/go/pkg/mod/github.com/prometheus/tsdb@v0.8.0/goversion/init.go:17:9: undefined: _SoftwareRequiresGOVERSION1_12
!! command failed: build -o /home/luyuan/go/src/github.com/thanos-io/thanos/thanos -ldflags -X github.com/prometheus/common/version.Version=0.6.1 -X github.com/prometheus/common/version.Revision=acb1cb03fd5ecf3e044a97642be50897db11c34f -X github.com/prometheus/common/version.Branch=master -X github.com/prometheus/common/version.BuildUser=root@pfw06vv.ops.hkht.qihoo.net -X github.com/prometheus/common/version.BuildDate=20190815-04:34:32  -extldflags '-static' -a -tags netgo github.com/improbable-eng/thanos/cmd/thanos: exit status 2
make: *** [build] 错误 1
[root@pfw06vv thanos]#

imluyuan:
[root@gobuildtest thanos]# make build
[...]

PR #1421, opened 2019-08-15
Replace k8s.io dependencies with specific pinned version

metalmatze:

## Changes

Fix Go modules for #1415 

## Verification

Tried `make build` with both `proxy.golang.org` and `goproxy.io`. Work on my machine ™

/cc @bwplotka @brancz @kakkoyun


metalmatze:
These replaces are actually taken from the Prometheus Operator where we were running into the same issue before.

https://github.com/coreos/prometheus-operator/blob/d1f2d19573329739513c6c46f6f8cc876173ce29/go.mod#L42-L50

go.mod
```diff
@@ -49,3 +49,13 @@ require (
 	gopkg.in/fsnotify.v1 v1.4.7
 	gopkg.in/yaml.v2 v2.2.2
 )
+
+replace (
```
bwplotka: Do you have an actual explanation of why this is needed? Also, those versions are from where?

Also I  think Thanos don't depend on all of them (e.g `klog`?) (:

metalmatze: This overrides the dependencies to use the exact commits of client-go instead of the sem ver that Go modules would pick, as those are broken.

We might not need klog, however it's a dependency of client-go so it's probably better to pin it in that case.

bwplotka: ok, fair: https://github.com/thanos-io/thanos/blob/master/go.sum#L616

Can we move back to `proxy.golang.org` maybe as our Proxy? As agreed then?

bwplotka: Let's comment maybe as well why we have this.

bwplotka:
DCO @metalmatze (:

bwplotka:
Merged as critical fix.

PR #1429, opened 2019-08-16
Add automaxprocs

2nick:
## Changes

- Add uber-go/automaxprocs integration to automatically set maxprocs value according to cgroup limit or $GOMAXPROCS env variable

## Verification

Run any thanos component with --log.level=debug in docker container with cpus limit or not empty env variable GOMAXPROCS and see debug log message from library which value has been set for golang maxprocs.


bwplotka:
Why closed? (:

bwplotka:
My first question is.. is this needed?

As:
>  By default, Go programs run with GOMAXPROCS set to the number of cores available; in prior Go1.5 it defaulted to 1.

PR #1430, opened 2019-08-16
Add automaxprocs

2nick:
## Changes

Add uber-go/automaxprocs integration to automatically set maxprocs value according to cgroup limit or $GOMAXPROCS env variable

## Verification

Run any thanos component with --log.level=debug in docker container with cpus limit or not empty env variable GOMAXPROCS and see debug log message from library which value has been set for golang maxprocs.

### DIFF
--- a/cmd/thanos/main.go
+++ b/cmd/thanos/main.go
@@ -35,6 +35,7 @@ import (
 	"github.com/thanos-io/thanos/pkg/runutil"
 	"github.com/thanos-io/thanos/pkg/tracing"
 	"github.com/thanos-io/thanos/pkg/tracing/client"
+	"go.uber.org/automaxprocs/maxprocs"
 	"google.golang.org/grpc"
 	"google.golang.org/grpc/codes"
 	"google.golang.org/grpc/credentials"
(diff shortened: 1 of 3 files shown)


cmd/thanos/main.go
```diff
@@ -115,6 +116,16 @@ func main() {
 		logger = log.With(logger, "ts", log.DefaultTimestampUTC, "caller", log.DefaultCaller)
 	}
 
+	loggerAdapter := func(template string, args ...interface{}) {
+		level.Debug(logger).Log("msg", fmt.Sprintf(template, args))
+	}
+
+	undo, err := maxprocs.Set(maxprocs.Logger(loggerAdapter))
```
bwplotka: Can we comment on what are rationales? Why we are doing this? 

There is almost 0 docs on https://github.com/uber-go/automaxprocs

d-ulyanov: @2nick let's add more comments here, pls

d-ulyanov: @bwplotka yep, you're right, app running in container will see all CPUs on host. `maxprocs.Set()` calculates GOMAXPROCS more accurate because also uses cgroup limits.
It's pretty convenient to configure CPU limits in one place when you're setting them for container and don't mind about GOMAXPROCS at all :)

2nick: Added comment and force-pushed amended commit - will github switch to the new commit automatically or I need to make some additional actions? First time merge requesting so sorry for newbie questions. :)

GiedriusS: It's ok, GitHub will take care of this automatically :+1:

bwplotka:
Changelog item is missing, but we can add it later (before tomorrow's release)

PR #1431, opened 2019-08-16
Support max source resolution for instant queries

0robustus1:
* [] Added `thanos query --query.instant.default.max_source_resolution` flag to set a default value for maxSourceResolution on instant queries (as are for example used by the `thanos rule` component)
* [] Added support for the max_source_resolution query parameter for `/api/v1/query`

## Reasoning

We have configured our compactor with different retentions for different resolution levels:

```
thanos
  compact
  --log.level=debug
  --data-dir=/var/thanos/store
  --objstore.config-file=/s3/thanos.yaml
  --retention.resolution-raw=7d
  --retention.resolution-5m=30d
  --retention.resolution-1h=180d
  --wait
```

This causes a situation where the result of a `thanos rule` recording rule was
only taking the last 7-8 days of data into account, because instant queries were
always using a maxSourceResolution of 0 and therefore only using raw resolution
data.

In our case the query looked like this and was executed on a thanos ruler
because our prometheuses only have a relatively short retention of a few days:

```
(100 -
  sum_over_time (
    some_metric:errors_count:increase1d[30d]
  ) /
  sum_over_time (
    some_metric:total_count:increase1d[30d]
  ) * 100
)
```

The `thanos rule` component runs an instant query in order to retrieve a single
datapoint to set for the recording rule metric. Our query requires that data
from at least 30 days is taken into account. But by default it would only select
raw data and therefore only provide data from the last 7 days.

## Changes

This allows for `/api/v1/query` to accept the `?max_source_resolution` query parameter
(the documentation did not clearly state that it was only accepted for
`/api/v1/query_range` prior to this).
We also added a command line argument to set a default should the query parameter not be set.

We also propose the following (for later PRs):

* Amend the UI to allow setting the `max_source_resolution` for "Console"
queries (which are executing instant queries).
* Add a new field to the rule-yaml for thanos ruler to set the
  `max_source_resolution` on a per-recording-rule basis.

## Verification

* We've used a fork to fire individual instant queries with the `max_source_resolution`
  parameter set and compared it with the last result we got from a range query.
* We also deployed a version with the default value changed and observed that the recording
  rule was now producing the expected value.

### DIFF
--- a/cmd/thanos/query.go
+++ b/cmd/thanos/query.go
@@ -67,7 +67,7 @@ func registerQuery(m map[string]setupFunc, app *kingpin.Application, name string
 	replicaLabel := cmd.Flag("query.replica-label", "Label to treat as a replica indicator along which data is deduplicated. Still you will be able to query without deduplication using 'dedup=false' parameter.").
 		String()
 
-	defaultMaxSourceResolution := modelDuration(cmd.Flag("query.instant.default.max_source_resolution", "default value for max_source_resolution for instant queries. If not set, defaults to 0s only taking raw resolution into account. 1h can be a good value if you use instant queries over time ranges that incorporate times outside of your raw-retention.").Default("0s"))
+	instantDefaultMaxSourceResolution := modelDuration(cmd.Flag("query.instant.default.max_source_resolution", "default value for max_source_resolution for instant queries. If not set, defaults to 0s only taking raw resolution into account. 1h can be a good value if you use instant queries over time ranges that incorporate times outside of your raw-retention.").Default("0s"))
 
 	selectorLabels := cmd.Flag("selector-label", "Query selector labels that will be exposed in info endpoint (repeated).").
 		PlaceHolder("<name>=\"<value>\"").Strings()
@@ -156,7 +156,7 @@ func registerQuery(m map[string]setupFunc, app *kingpin.Application, name string
 			time.Duration(*dnsSDInterval),
 			*dnsSDResolver,
 			time.Duration(*unhealthyStoreTimeout),
-			time.Duration(*defaultMaxSourceResolution),
+			time.Duration(*instantDefaultMaxSourceResolution),
 		)
 	}
 }
@@ -274,7 +274,7 @@ func runQuery(
 	dnsSDInterval time.Duration,
 	dnsSDResolver string,
 	unhealthyStoreTimeout time.Duration,
-	defaultMaxSourceResolution time.Duration,
+	instantDefaultMaxSourceResolution time.Duration,
 ) error {
 	// TODO(bplotka in PR #513 review): Move arguments into struct.
 	duplicatedStores := prometheus.NewCounter(prometheus.CounterOpts{
@@ -405,7 +405,7 @@ func runQuery(
 		ins := extpromhttp.NewInstrumentationMiddleware(reg)
 		ui.NewQueryUI(logger, reg, stores, flagsMap).Register(router.WithPrefix(webRoutePrefix), ins)
 
-		api := v1.NewAPI(logger, reg, engine, queryableCreator, enableAutodownsampling, enablePartialResponse, defaultMaxSourceResolution)
+		api := v1.NewAPI(logger, reg, engine, queryableCreator, enableAutodownsampling, enablePartialResponse, instantDefaultMaxSourceResolution)
 
 		api.Register(router.WithPrefix(path.Join(webRoutePrefix, "/api/v1")), tracer, logger, ins)
 
--- a/cmd/thanos/query.go
+++ b/cmd/thanos/query.go
@@ -67,7 +67,7 @@ func registerQuery(m map[string]setupFunc, app *kingpin.Application, name string
 	replicaLabel := cmd.Flag("query.replica-label", "Label to treat as a replica indicator along which data is deduplicated. Still you will be able to query without deduplication using 'dedup=false' parameter.").
 		String()
 
-	defaultMaxSourceResolution := modelDuration(cmd.Flag("query.instant.default.max_source_resolution", "default value for max_source_resolution for instant queries. If not set, defaults to 0s only taking raw resolution into account.").Default("0s"))
+	defaultMaxSourceResolution := modelDuration(cmd.Flag("query.instant.default.max_source_resolution", "default value for max_source_resolution for instant queries. If not set, defaults to 0s only taking raw resolution into account. 1h can be a good value if you use instant queries over time ranges that incorporate times outside of your raw-retention.").Default("0s"))
 
 	selectorLabels := cmd.Flag("selector-label", "Query selector labels that will be exposed in info endpoint (repeated).").
 		PlaceHolder("<name>=\"<value>\"").Strings()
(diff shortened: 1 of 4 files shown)


brancz:
This looks right to me, but it'd be good if @bwplotka and or @GiedriusS have a look as well.

pkg/query/api/v1.go
```diff
@@ -203,13 +206,18 @@ func (api *API) parseEnableDedupParam(r *http.Request) (enableDeduplication bool
 }
 
 func (api *API) parseDownsamplingParamMillis(r *http.Request, step time.Duration) (maxResolutionMillis int64, _ *ApiError) {
+	// If no max_source_resolution is specified fit at least 5 samples between steps.
+	return api.parseDownsamplingParamMillisWithDefault(r, step/5)
```
povilasv: Consider removing this function, and just calling `api.parseDownsamplingParamMillisWithDefault(r, step/5)` directly.

cmd/thanos/query.go
```diff
@@ -67,6 +67,8 @@ func registerQuery(m map[string]setupFunc, app *kingpin.Application, name string
 	replicaLabel := cmd.Flag("query.replica-label", "Label to treat as a replica indicator along which data is deduplicated. Still you will be able to query without deduplication using 'dedup=false' parameter.").
 		String()
 
+	defaultMaxSourceResolution := modelDuration(cmd.Flag("query.instant.default.max_source_resolution", "default value for max_source_resolution for instant queries. If not set, defaults to 0s only taking raw resolution into account.").Default("0s"))
```
povilasv: Consider improving documentation here. What are good values for this flag ? Maybe some better guide for users experiencing similiar problem to yours.

Also are you sure that this impacts only `instant` queries?

Looking at the code you changed `parseDownsamplingParamMillis` to incorporate default, which is also used in range queries.

Maybe we should have one flag for both `instant` and range queries ?

0robustus1: The new function `parseDownsamplingParamMillisWithDefault` accepts the default as a parameter. For query_range that parameter is `step/5`. The `api.defaultInstantQueryMaxSourceResolution` is only used in the `query` function which corresponds to instant queries (as far as i read it).

Regarding one default for both: There has already been a default for range queries: `step/5`. As `?step` is a query parameter that is already defined by the prometheus API it provides better integration with clients than `maxSourceResolution` (at least currently). That is why i only defined the flag within the scope of instant queries. Long-Term the flag might not be necessary if we integrate `?max_source_resolution` for instant queries with clients such as grafana (or thanos' own thanos ruler).

Regarding docs: I added an example. Not sure if this bloats the documentation for a commandline flag too much though.

0robustus1:
Hi @GiedriusS, that might be a significantly more comprehensive and powerful solution.
It sounds however like a larger-scale change (that should then be adopted for all query types).

There is just one concern that comes to mind. Wouldn't the querier need to be aware of the retention configured for the different compaction levels? How else would it have translated the `30d` range selector to the correct maxSourceResolution (`1h`) in a retention setup like this:

```
thanos
  compact
  --log.level=debug
  --data-dir=/var/thanos/store
  --objstore.config-file=/s3/thanos.yaml
  --retention.resolution-raw=7d
  --retention.resolution-5m=11d
  --retention.resolution-1h=365d
  --wait
```

Configuring a `maxSourceResolution` via the flag would enable that.

0robustus1:
That's what we set it to in our setup now and would probably leave it at. I just wasn't sure whether to include it in the Pull-Request.

GiedriusS:
> Hi @GiedriusS, that might be a significantly more comprehensive and powerful solution.
> It sounds however like a larger-scale change (that should then be adopted for all query types).
> 
> There is just one concern that comes to mind. Wouldn't the querier need to be aware of the retention configured for the different compaction levels? How else would it have translated the `30d` range selector to the correct maxSourceResolution (`1h`) in a retention setup like this:
> 
> ```
> thanos
>   compact
>   --log.level=debug
>   --data-dir=/var/thanos/store
>   --objstore.config-file=/s3/thanos.yaml
>   --retention.resolution-raw=7d
>   --retention.resolution-5m=11d
>   --retention.resolution-1h=365d
>   --wait
> ```
> 
> Configuring a `maxSourceResolution` via the flag would enable that.

Indeed, it would be a much bigger change but not sure that introducing a new flag is the best solution. It's always easiest to do that but then after some time you end with a thing which has hundreds of knobs which is bad.

I think my original suggestion would work out because actually we do downsampling at fixed intervals. The options that you've pasted only control the retention part. I have outlined this in my other issue what I have started about how Thanos Store should always prefer higher resolution data.

I still would very much like @bwplotka to comment on this.

cmd/thanos/query.go
```diff
@@ -67,6 +67,8 @@ func registerQuery(m map[string]setupFunc, app *kingpin.Application, name string
 	replicaLabel := cmd.Flag("query.replica-label", "Label to treat as a replica indicator along which data is deduplicated. Still you will be able to query without deduplication using 'dedup=false' parameter.").
 		String()
 
+	defaultMaxSourceResolution := modelDuration(cmd.Flag("query.instant.default.max_source_resolution", "default value for max_source_resolution for instant queries. If not set, defaults to 0s only taking raw resolution into account. 1h can be a good value if you use instant queries over time ranges that incorporate times outside of your raw-retention.").Default("0s"))
```
bwplotka: I am happy with this if you make this flag `.Hidden` until we have a clear direction/decision on this matter.

bwplotka:
```suggestion
	instantDefaultMaxSourceResolution := modelDuration(cmd.Flag("query.instant.default.max_source_resolution", "default value for max_source_resolution for instant queries. If not set, defaults to 0s only taking raw resolution into account. 1h can be a good value if you use instant queries over time ranges that incorporate times outside of your raw-retention.").Default("0s"))
```

pkg/query/api/v1.go
```diff
@@ -202,14 +205,14 @@ func (api *API) parseEnableDedupParam(r *http.Request) (enableDeduplication bool
 	return enableDeduplication, nil
 }
 
-func (api *API) parseDownsamplingParamMillis(r *http.Request, step time.Duration) (maxResolutionMillis int64, _ *ApiError) {
+func (api *API) parseDownsamplingParamMillisWithDefault(r *http.Request, defaultVal time.Duration) (maxResolutionMillis int64, _ *ApiError) {
```
bwplotka: the `default` argument is clear enough so I think `parseDownsamplingParamMillis` is fine (: but not a blocker

pkg/query/api/v1_test.go
```diff
@@ -902,7 +902,7 @@ func TestParseDownsamplingParamMillis(t *testing.T) {
 		v.Set("max_source_resolution", test.maxSourceResolutionParam)
 		r := http.Request{PostForm: v}
 
-		maxResMillis, _ := api.parseDownsamplingParamMillis(&r, test.step)
+		maxResMillis, _ := api.parseDownsamplingParamMillisWithDefault(&r, test.step/5)
```
bwplotka: Please preserve the comment why 5

0robustus1:
I adjusted the code to the remaining review comments. The flag is now hidden.

bwplotka:
Actually, I was super wrong.. All things I mentioned with resolution per range works right now: https://prometheus.io/blog/2019/01/28/subquery-support/#subqueries 

I might think we need to change slightly implementation then for querier and use per selection `	Step int64  // Query step size in milliseconds.` instead of global, query one.

bwplotka:
Failed quick experiment [here](https://github.com/thanos-io/thanos/pull/1465) so merging this "hidden" workaround for now. 

Thanks for doing this! Would love to have your input on further work in this area @0robustus1

PR #1432, opened 2019-08-19
Fix location of testutil pkg after prometheus update.

johncming:
* bugfix after prometheus update. https://github.com/prometheus/prometheus/pull/5893

## Changes

prometheus tsdb/testutil has moved to util/testutil, so change it in thanos too

## Verification

1. run go vet first
2. run go test ./... -v

Signed-off-by: johncming <johncming@yahoo.com>


bwplotka:
Hm, I think if we want to fix this, we need to update deps as well, right? (:

johncming:
> Hm, I think if we want to fix this, we need to update deps as well, right? (:

Does it mean go mod？

go.mod
```diff
@@ -33,20 +47,27 @@ require (
 	github.com/opentracing/opentracing-go v1.1.0
 	github.com/pkg/errors v0.8.1
 	github.com/prometheus/client_golang v1.1.0
+	github.com/prometheus/client_model v0.0.0-20190812154241-14fe0d1b01d4 // indirect
 	github.com/prometheus/common v0.6.0
-	github.com/prometheus/prometheus v0.0.0-20190814090039-b5c833ca2194
```
jojohappy: Em... Why does prometheus rollback to 2.5.0? ):

go.mod
```diff
@@ -33,20 +47,27 @@ require (
...
+	github.com/prometheus/prometheus v2.5.0+incompatible
```
jojohappy: Em... Why does prometheus rollback to 2.5.0? ):

johncming: > Em... Why does prometheus rollback to 2.5.0? ):

No special version is specified. According  minimal version selection algorithm, 2.5 was selected. 

Considering the updated version, there may be other effects, please turn off this pr for the time being.

thanks.

https://github.com/golang/go/wiki/Modules#version-selection

PR #1437, opened 2019-08-19
Reduce CI memory usage; Added memstat implementatation to gather data.

bwplotka:
This is for baseline mostly for now. Will experiment bit in order to detect a leak in our go test.

First try: `TotalPeak RSS: (locally  445417)  697954-726349 Bytes vs.. far from worrying number.


bwplotka:
Did this in another PR.

PR #1439, opened 2019-08-20
pkg/rule: Fix bugs where rules were out of sync.

### LINKED ISSUE #1739: rule: Alerts will not be deleted after curl -X POST localhost:10902/-/reload.

0.8.1

**What happened**:
I deleted alerts file and used curl to reload thanos. But the alerts are still in thanos-ruler.

The alerts should be disappeared from thanos-ruler

`kubectl apply -f rules.yaml`
```---
apiVersion: v1
kind: ConfigMap
metadata:
  name: rule-test
  namespace: monitoring
data:
  test_rules_alerts.rules.yaml: |-
    groups:
    - name: test.alarm
      rules:
      - alert: TEST-ALARM
        expr: vector(1)
```
`curl -X POST localhost:10902/-/reload.`

Then the alert is in thanos-rule.
And now:

`kubectl delete -f rules.yaml`
`curl -X POST localhost:10902/-/reload.`

The alert is not disappeared.

```
{"caller":"rule.go:441","level":"info","msg":"reload rule files","numFiles":1,"ts":"2019-11-11T14:45:15.926305365Z"}

{"caller":"rule.go:441","level":"info","msg":"reload rule files","numFiles":0,"ts":"2019-11-11T14:45:27.957372263Z"}
```

johncming:
Thanos rule loading alerts about out-of-sync rule files

Specifically, the following scenario:

Rule file rule. Yaml, which contains some rules, through the reload command, thanos rule load rules.

If you clear the contents of rule. Yaml, but leave the file in, the old rules in thanos rule will not be cleared.

This bug leads to a common situation (all rules in one file)

If thanos rule has 100 rules, it can only clear 99 rules at most.

Closes #1739


pkg/rule/rule.go
```diff
@@ -113,7 +113,10 @@ func (r RuleGroup) MarshalYAML() (interface{}, error) {
 func (m *Managers) Update(dataDir string, evalInterval time.Duration, files []string) error {
 	var (
 		errs     = tsdberrors.MultiError{}
-		filesMap = map[storepb.PartialResponseStrategy][]string{}
+		filesMap = map[storepb.PartialResponseStrategy][]string{
```
bwplotka: Hm I think that will work, but to make it verbose, clearer and bit more efficient I would suggest add one more loop at the end:

```go
	for s, updater := range *m {
		if _, ok := filesMap[s]; ok {
			// We already updated those above.
			continue
		}
		// Rules removed update updater if needed.
		if len(updater.RuleGroups()) == 0 {
			continue
		}

		if err := updater.Update(evalInterval, []string{}, nil); err != nil {
			errs = append(errs, err)
		}
	}
```
What do you think? (:

johncming:
> Nice! I think that makes sense but have one suggestion
> 
> Thanks for helping 👍
> 
> On top of the suggestion below I think we really need more unit tests to:
> 
> * Detect those bug early
> * Do not repeat same bug anymore
> * Confirm fix works (:
> 
> Do you think we can add some tests?

Yes, it is necessary to add test cases.

pkg/rule/rule.go
johncming: > Hm I think that will work, but to make it verbose, clearer and bit more efficient I would suggest add one more loop at the end:
> 
> ```go
> 	for s, updater := range *m {
> 		if _, ok := filesMap[s]; ok {
> 			// We already updated those above.
> 			continue
> 		}
> 		// Rules removed update updater if needed.
> 		if len(updater.RuleGroups()) == 0 {
> 			continue
> 		}
> 
> 		if err := updater.Update(evalInterval, []string{}, nil); err != nil {
> 			errs = append(errs, err)
> 		}
> 	}
> ```
> 
> What do you think? (:

it 's more clear. how about merge the last loop with your loop?

GiedriusS:
@johncming do you have any free time to finish up this PR? :p

simonpasquier:
@johncming do you have time to rebase?

johncming:
> @johncming do you have time to rebase?

yes. I will finish this.

pkg/rule/rule_test.go
```diff
@@ -222,6 +225,47 @@ groups:
...
+
+	ruleMgrAbort.Run()
+	ruleMgrWarn.Run()
+	defer ruleMgrAbort.Stop()
+	defer ruleMgrWarn.Stop()
+
+	err = m.Update(1*time.Second, []string{
+		filepath.Join(dir, "no_strategy.yaml"),
+	})
+
```
simonpasquier: (nit) remove blank line

pkg/rule/rule_test.go
```diff
@@ -222,6 +225,47 @@ groups:
 	}
 }
 
+func TestUpdateAfterClear(t *testing.T) {
+	dir, err := ioutil.TempDir("", "test_rule_rule_groups")
+	testutil.Ok(t, err)
+	defer func() { testutil.Ok(t, os.RemoveAll(dir)) }()
+	err = os.MkdirAll(filepath.Join(dir, "subdir"), 0775)
```
simonpasquier: this line and the next one aren't needed AFAICT.

simonpasquier:
I think you'll need to update the CHANGELOG and reference https://github.com/thanos-io/thanos/issues/1739 too.

johncming:
> I think you'll need to update the CHANGELOG and reference #1739 too.

Updated.  

When I read the CHANGELOG, there was no place to put this reference #1739. Where should I put this reference #1739? Or I reference it in the pr conversations?

Thanks.

simonpasquier:
> When I read the CHANGELOG, there was no place to put this reference #1739. Where should I put this reference #1739? Or I reference it in the pr conversations?

Yes I meant the latter (adding to `Closes #1739` to the PR description).

johncming:
> > When I read the CHANGELOG, there was no place to put this reference #1739. Where should I put this reference #1739? Or I reference it in the pr conversations?
> 
> Yes, I meant the latter (adding to `Closes #1739` to the PR description).

Yes, it is updated.

daixiang0:
@johncming do you have to rebase?

johncming:
> @johncming can you rebase?

rebased.

CHANGELOG.md
```diff
@@ -44,6 +44,7 @@ Compactor now properly handles partial block uploads for all operation like rete
 - [#1882](https://github.com/thanos-io/thanos/pull/1882) Receive: upload to object storage as 'receive' rather than 'sidecar'.
 - [#1907](https://github.com/thanos-io/thanos/pull/1907) Store: Fixed the duration unit for the metric `thanos_bucket_store_series_gate_duration_seconds`.
 - [#1931](https://github.com/thanos-io/thanos/pull/1931) Compact: Fixed the compactor successfully exiting when actually an error occurred while compacting a blocks group.
+- [#1439](https://github.com/thanos-io/thanos/pull/1439) Rule: Fix bugs where rules were out of sync.
```
simonpasquier: It should go in the `Unreleased` section.

pkg/rule/rule.go
```diff
@@ -213,6 +213,21 @@ func (m *Manager) Update(evalInterval time.Duration, files []string) error {
 			continue
 		}
 	}
+
+	for s, mgr := range m.mgrs {
```
simonpasquier: it might be worth adding a comment explaining that this section removes the rules from a manager when a strategy has no more rule.

GiedriusS:
@johncming thanks for your persistence! Care to fix the code according to the last comments so that we could merge it and fix it?

lilic:
Any reason why this PR was never merged, was this issue fixed by another PR? Thanks! @johncming @bwplotka

simonpasquier:
@lilic it has been closed by the bot because it has seen no activity in a while. @johncming would you be able to reopen it and address the comments?

lilic:
@johncming  if you don't have time to address the comments for any reason, I can also takeover and pick up this PR from where you leftover with your commits of course? Thanks!

Issue #1442, opened 2019-08-20
Fix Flaky test


bwplotka:
```
reloader_test.go:269:

	exp: 5

	got: 3

--- FAIL: TestReloader_RuleApply (10.04s)
    reloader_test.go:238: Performing step number 1
    reloader_test.go:238: Performing step number 3
FAIL
FAIL	github.com/thanos-io/thanos/pkg/reloader	10.769s
```
Quite often spotted.

johncming:
https://github.com/thanos-io/thanos/blob/c35d47b41954dae0e9b5592cb327bf0896e78e26/pkg/reloader/reloader_test.go#L182

Reason: plus op is not an atomic operation.

jojohappy:
@bwplotka @johncming I guess that the reloader read the middle status of watched files, because of using `ioutil.WriteFile`.

```
// WriteFile writes data to a file named by filename.
// If the file does not exist, WriteFile creates it with permissions perm;
// otherwise WriteFile truncates it before writing.
```

bwplotka:
@johncming I think this part is ok as there is only a single writer.

@jojohappy you are right but this can happen normally as well so we should handle it properly.

IMO it is just a CI slowdown, so increasing timeout here https://github.com/thanos-io/thanos/issues/1442 (:

PR #1449, opened 2019-08-22
pkg/reloader: Fixed concurrent conflicts bug.

johncming:
Fix the bug mentioned in issue #1442

### DIFF
--- a/pkg/reloader/reloader_test.go
+++ b/pkg/reloader/reloader_test.go
@@ -167,8 +167,7 @@ func TestReloader_RuleApply(t *testing.T) {
 	l, err := net.Listen("tcp", "localhost:0")
 	testutil.Ok(t, err)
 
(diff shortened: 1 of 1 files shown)


jojohappy:
Thank you for your PR! But maybe it is not right point. I can still reproduce that issue with your PR.

johncming:
> Thank you for your PR! But maybe it is not right point. I can still reproduce that issue with your PR.

It's really hard to repeat. I repeated the testcase 100 times and there was no failed once.

bwplotka:
See my response [here](https://github.com/thanos-io/thanos/issues/1442#issuecomment-523867826) 

I think it was just a CI slow down so increased timeout to 10 seconds. Let's see. If that will happen we can reopen issue and continue this PR (: 

Let me know your thoughts, closing this for now @johncming

PR #1458, opened 2019-08-26
query/receive: Unify HTTP server instrumentation

kakkoyun:
This PR attempts to unify HTTP server instrumentation behavior across components and removes unnecessary metrics. Thanos Query and Receive now use common instrumentation middleware. 

PTAL #1420 for previous discussions.

## Changes

* Thanos Query no longer exposes `thanos_query_api_instant_query_duration_seconds`, `thanos_query_api_range_query_duration_second` metrics.
* Thanos Receive no longer exposes `thanos_http_request_duration_seconds`, `thanos_http_requests_total`, `thanos_http_response_size_bytes`.

## Verification

Local `make test` and run.

### DIFF
--- a/pkg/receive/handler.go
+++ b/pkg/receive/handler.go
@@ -21,10 +21,10 @@ import (
 	opentracing "github.com/opentracing/opentracing-go"
 	"github.com/pkg/errors"
 	"github.com/prometheus/client_golang/prometheus"
-	"github.com/prometheus/client_golang/prometheus/promhttp"
 	"github.com/prometheus/common/route"
 	promtsdb "github.com/prometheus/prometheus/storage/tsdb"
 	terrors "github.com/prometheus/prometheus/tsdb/errors"
+	extpromhttp "github.com/thanos-io/thanos/pkg/extprom/http"
 	"github.com/thanos-io/thanos/pkg/runutil"
 	"github.com/thanos-io/thanos/pkg/store/prompb"
 )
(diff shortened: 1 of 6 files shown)


kakkoyun:
cc @brancz @bwplotka @squat

pkg/receive/handler.go
```diff
@@ -103,35 +79,20 @@ func NewHandler(logger log.Logger, o *Options) *Handler {
 		),
 	}
 
-	router := route.New().WithInstrumentation(h.instrumentHandler)
-	h.router = router
+	ins := extpromhttp.NewNopInstrumentationMiddleware()
+	if o.Registry != nil {
+		ins = extpromhttp.NewInstrumentationMiddleware(o.Registry)
+		o.Registry.MustRegister(h.forwardRequestsTotal)
```
brancz: Did we always forget to register this or am I missing why this line is here?

brancz:
Sorry misclickes. Didn’t mean to approve, just comment.

pkg/receive/handler.go
kakkoyun: If I understood you correctly, It was registered before. I just removed the ones which come from `InstrumentationMiddleware`.
https://github.com/thanos-io/thanos/blob/8feb0798a68e2a11b9bd625fcc80842f0d04a758/pkg/receive/handler.go#L112-L119

brancz: Got it. Makes sense now.

PR #656, opened 2018-12-03
WIP: Added /-/healthy and /-/ready endpoints to all thanos components

FUSAKLA:
Signed-off-by: Martin Chodur <m.chodur@seznam.cz>
Closes https://github.com/improbable-eng/thanos/issues/644

Addition of liveness and readiness endpoints to all components.
Added package `prober` which olds information if the component is `healthy` and `ready`.
It can be registered to `Mux` or `Router` so that it can be used in the `metricHTTPListenGroup` in components which does not have own UI or in components that has own routing.

**store**
In the `store` the initial loading of cache was blocking start of the HTTP server thus it couldnt expose the liveness check. Because of that the initial cache update was moved to the `g` as an actor and only readiness of the thanos-store is set to true when the cache is updated for the first time.

**receive**
As [discussed on slack](https://improbable-eng.slack.com/archives/CADR0UWS2/p1553233498251300) the receive had two http interfaces which were merged together when adding the prober. 
As a side effect it resolves https://github.com/improbable-eng/thanos/issues/959

## Verification
Tests are passing and it was tested on started every tjanos component type.

PR #1460, opened 2019-08-26
feat store: added readiness and livenes prober

FUSAKLA:
Added the prober to yet another component. Store this time.

The store does the bucket initialization and sync synchronously, so the default http server will start up only once it is finished. Then it gets healthy(live) and once the gRPC server starts up the store gets also ready.

Personally I'd much rather see the liveness probe to answer right from the start during the bucket init since this way we leave the bucket sync duration to some initial back-off of an orchestrator. 

## Changes

- added prober to the thanos store

## Verification

Started it and verified it gets ready end healthy as expected.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -12,10 +12,11 @@ We use *breaking* word for marking changes that are not backward compatible (rel
 ## Unreleased
 
 ## Added
-- [#1540](https://github.com/thanos-io/thanos/pull/1540) Added `/-/ready` and `/-/healthy` endpoints to Thanos Downsample.
-- [#1538](https://github.com/thanos-io/thanos/pull/1538) Added `/-/ready` and `/-/healthy` endpoints to Thanos Rule.
-- [#1537](https://github.com/thanos-io/thanos/pull/1537) Added `/-/ready` and `/-/healthy` endpoints to Thanos Receive.
-- [#1534](https://github.com/thanos-io/thanos/pull/1534) Added `/-/ready` endpoint to Thanos Query.
+- [#1540](https://github.com/thanos-io/thanos/pull/1540) Thanos Downsample added `/-/ready` and `/-/healthy` endpoints.
+- [#1538](https://github.com/thanos-io/thanos/pull/1538) Thanos Rule added `/-/ready` and `/-/healthy` endpoints.
+- [#1537](https://github.com/thanos-io/thanos/pull/1537) Thanos Receive added `/-/ready` and `/-/healthy` endpoints.
+- [#1460](https://github.com/thanos-io/thanos/pull/1460) Thanos Store Added `/-/ready` and `/-/healthy` endpoints.
+- [#1534](https://github.com/thanos-io/thanos/pull/1534) Thanos Query Added `/-/ready` and `/-/healthy` endpoints.
 - [#1533](https://github.com/thanos-io/thanos/pull/1533) Thanos inspect now supports the timeout flag.
 
 ### Fixed
--- a/cmd/thanos/store.go
+++ b/cmd/thanos/store.go
@@ -8,20 +8,22 @@ import (
 	"github.com/go-kit/kit/log"
 	"github.com/go-kit/kit/log/level"
 	"github.com/oklog/run"
-	opentracing "github.com/opentracing/opentracing-go"
+	"github.com/opentracing/opentracing-go"
 	"github.com/pkg/errors"
 	"github.com/prometheus/client_golang/prometheus"
+	"github.com/thanos-io/thanos/pkg/component"
 	"github.com/thanos-io/thanos/pkg/model"
 	"github.com/thanos-io/thanos/pkg/objstore/client"
+	"github.com/thanos-io/thanos/pkg/prober"
 	"github.com/thanos-io/thanos/pkg/runutil"
 	"github.com/thanos-io/thanos/pkg/store"
 	storecache "github.com/thanos-io/thanos/pkg/store/cache"
-	kingpin "gopkg.in/alecthomas/kingpin.v2"
+	"gopkg.in/alecthomas/kingpin.v2"
 )
 
 // registerStore registers a store command.
-func registerStore(m map[string]setupFunc, app *kingpin.Application, name string) {
-	cmd := app.Command(name, "store node giving access to blocks in a bucket provider. Now supported GCS, S3, Azure, Swift and Tencent COS.")
+func registerStore(m map[string]setupFunc, app *kingpin.Application) {
+	cmd := app.Command(component.Store.String(), "store node giving access to blocks in a bucket provider. Now supported GCS, S3, Azure, Swift and Tencent COS.")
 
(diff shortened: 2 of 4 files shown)


cmd/thanos/store.go
```diff
@@ -106,104 +110,106 @@ func runStore(
...
 	syncInterval time.Duration,
 	blockSyncConcurrency int,
 	filterConf *store.FilterConfig,
 ) error {
-	{
-		confContentYaml, err := objStoreConfig.Content()
-		if err != nil {
-			return err
-		}
+	statusProber := prober.NewProber(component, logger, prometheus.WrapRegistererWithPrefix("thanos_", reg))
```
kakkoyun: Could we not wrap metrics with `thanos_` prefix in this PR, we could have it in another one. We have alerts and dashboards depend on this, better to do this change explicitly.

kakkoyun:
Hey @FUSAKLA, I realized that we miss liveness and healthiness checks for various components when I was working on https://github.com/metalmatze/kube-thanos/pull/42 and then I have seen your work #656, #1395 and #1297. 

We need probes in place for all other components as well. I'd like to push this effort further as fast as possible. I'm happy to help. I could create PRs for components which you haven't touch yet if it is also ok for you? I'd be happy to hear your opinion on those PRs as well.

What do you think?

FUSAKLA:
Hi @kakkoyun ! Yeah I didn't want to make more PRs at once since even this way it takes it's time to get these reviewed. Every one of them raises big discussion on when to report as ready and when as healthy. Thats also the reason it got eventually splitted up but help is always welcomed so feel free to add those PRS! I'd love to see this done finally :D
Thanks for the initiative!

kakkoyun:
@FUSAKLA Alright, greate! Let's do this then. I've already sent the first one and pinged you on #1534, PTAL.

It'd be great if we can push this one too.

CHANGELOG.md
```diff
@@ -11,6 +11,10 @@ We use *breaking* word for marking changes that are not backward compatible (rel
 
 ## Unreleased
 
+### Added
+- [#1460](https://github.com/thanos-io/thanos/pull/1460) Thanos store Added `/-/ready` and `/-/healthy` endpoints.
+
```
bwplotka: Kill line?

cmd/thanos/store.go
```diff
@@ -5,10 +5,13 @@ import (
 	"net"
 	"time"
 
+	"github.com/thanos-io/thanos/pkg/component"
+	"github.com/thanos-io/thanos/pkg/prober"
+
```
bwplotka: you need to kill this line and run `make format` ):

cmd/thanos/store.go
bwplotka: Hm are we changing anything? Before we did not have any probe metrics so I think this should be fine?

cmd/thanos/store.go
```diff
@@ -106,104 +110,106 @@ func runStore(
...
+	g.Add(func() error {
+		level.Info(logger).Log("msg", "Listening for StoreAPI gRPC", "address", grpcBindAddr)
+		statusProber.SetReady()
+		return errors.Wrap(s.Serve(l), "serve gRPC")
+	}, func(error) {
+		s.Stop()
+	})
+
+	if err := defaultHTTPListener(g, logger, reg, httpBindAddr, statusProber); err != nil {
+		return errors.Wrap(err, "create readiness prober")
```
bwplotka: technically we create HTTP Server not prober

cmd/thanos/store.go
FUSAKLA: Hmm as Bartek pointed out this does not change anything just adds new metrics. Is that an issue for you?

cmd/thanos/store.go
FUSAKLA: should be properly formatted now :crossed_fingers:

cmd/thanos/store.go
FUSAKLA: you are right, changed to `scheduling HTTP server`

FUSAKLA:
Thanks @bwplotka ! Should be resolved and rebased on master.

cmd/thanos/store.go
```diff
@@ -104,102 +106,104 @@ func runStore(
...
+	if err != nil {
+		return errors.Wrap(err, "create bucket client")
+	}
 
-		bkt, err := client.NewBucket(logger, confContentYaml, reg, component)
+	// Ensure we close up everything properly.
+	defer func() {
 		if err != nil {
-			return errors.Wrap(err, "create bucket client")
+			runutil.CloseWithLogOnErr(logger, bkt, "bucket client")
```
povilasv: I think we should remove `if err !=nil`, as it is run `in defer`,  `err` will be overwritten by non related code which happens after this line.

cmd/thanos/store.go
```diff
@@ -104,102 +106,104 @@ func runStore(
...
-				return nil
-			})
-
-			runutil.CloseWithLogOnErr(logger, bs, "bucket store")
-			return err
-		}, func(error) {
-			cancel()
+	ctx, cancel := context.WithCancel(context.Background())
+	g.Add(func() error {
+		defer runutil.CloseWithLogOnErr(logger, bkt, "bucket client")
```
povilasv: this is a bit confusing, so we were closing `bkt` earlier, why are we closing here again?

cmd/thanos/store.go
```diff
@@ -104,102 +106,104 @@ func runStore(
...
+	g.Add(func() error {
+		level.Info(logger).Log("msg", "Listening for StoreAPI gRPC", "address", grpcBindAddr)
+		statusProber.SetReady()
+		return errors.Wrap(s.Serve(l), "serve gRPC")
+	}, func(error) {
+		s.Stop()
+	})
+
+	if err := scheduleHTTPServer(g, logger, reg, statusProber, httpBindAddr, nil, component); err != nil {
+		return errors.Wrap(err, "scheduing HTTP server")
```
povilasv:
```suggestion
		return errors.Wrap(err, "schedule HTTP server")
```

cmd/thanos/store.go
FUSAKLA: I'd say it was meant to close the client if the `runStore` ended up with an error. 
Otherwise it should not close the client since others are still using it. 
Maybe that's why the `if err != nil` is there and the fact err will be overwritten by the following code was the intention?

But I did not touch this code actually, just removed the unneeded code block wrapping it.

cmd/thanos/store.go
FUSAKLA: commented here https://github.com/thanos-io/thanos/pull/1460#discussion_r326884636

FUSAKLA:
@kakkoyun PTAL I removed the `metricHTTPListenGroup` since it is not used anymore and unified the changelog

cmd/thanos/store.go
kakkoyun: lgtm

PR #1465, opened 2019-08-27
Choose lowest downsampling resolution with granular control.

bwplotka:
Playing with `Select` `storage.SelectParams.Step`

As a follow up on this: https://github.com/thanos-io/thanos/pull/1431 I was experimenting with subquery + resolution in order to automatically adjust what downsampling Thanos should use. This, in theory, sounds great but looks like I don't really understand PromQL `storage.Params.Step` field properly. Need to investigate more (is this param not propagated)?

TODO: 
* [] Proper design
* [] Investigate why the step is not propagated from `[...:<resolution>]`

### DIFF
--- a/pkg/query/querier_test.go
+++ b/pkg/query/querier_test.go
@@ -22,23 +22,180 @@ import (
 	"github.com/thanos-io/thanos/pkg/testutil"
 )
 
-func TestQueryableCreator_MaxResolution(t *testing.T) {
+type recordRequestsStoreServer struct {
+	// This field just exist to pseudo-implement the unused methods of the interface.
+	storepb.StoreServer
+
+	reqs []storepb.SeriesRequest
+}
+
+func (s *recordRequestsStoreServer) Series(r *storepb.SeriesRequest, _ storepb.Store_SeriesServer) error {
+	s.reqs = append(s.reqs, *r)
+	return nil
+}
+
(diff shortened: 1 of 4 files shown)


pkg/query/querier_test.go
```diff
@@ -22,23 +22,180 @@ import (
...
+							Name:  "__name__",
+							Value: "up",
+						},
+					},
+					Aggregates: []storepb.Aggr{storepb.Aggr_COUNT, storepb.Aggr_SUM},
+				},
+			},
+		},
+		{
+			query: "rate(up[5m:200s])",
```
bwplotka: This does not work as expected (in my mind). Step is still passed as 100 ms no matter what. cc @brian-brazil for help =D

bwplotka: Based my knowledge on: https://prometheus.io/blog/2019/01/28/subquery-support/#subqueries

brian-brazil: The step is the one for query_range. We're not passing any hints relating to subqueries currently.

bwplotka: Thanks! Will look deeper in PromQL if it's sane to actually propagate that hint & how.

damnever:
The step in query hints seems to be fixed now. The remaining issue is that we still can’t tune lookback delta for subqueries.

PR #1468, opened 2019-08-28
Store: Refactor run group to errgroup

povilasv:
Signed-off-by: Povilas Versockas <p.versockas@gmail.com>

## Changes

Changes bucket functions to use `errgroup` instead of `run.Group` 

`run.Group` concurrently runs all of the goroutines, waits until the first actor exits, invokes the interrupt functions, and finally returns control to the caller only once all actors have returned. 

This makes it that if first goroutine returns `nil`, and other goroutine returns an error, we won't receive it, as it gets thrown away:

https://github.com/oklog/run/blob/master/group.go#L51-L53

On the other hand `errgroup` - blocks until all function calls from the Go method have returned, then returns the first non-nil error (if any) from them.

This will return us an error if any of the goroutines fail

### DIFF
--- a/pkg/store/bucket.go
+++ b/pkg/store/bucket.go
@@ -18,7 +18,6 @@ import (
 
 	"github.com/go-kit/kit/log"
 	"github.com/go-kit/kit/log/level"
-	"github.com/oklog/run"
 	"github.com/oklog/ulid"
 	"github.com/pkg/errors"
 	"github.com/prometheus/client_golang/prometheus"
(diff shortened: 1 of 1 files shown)


pkg/store/bucket.go
```diff
@@ -775,10 +774,6 @@ func (s *BucketStore) Series(req *storepb.SeriesRequest, srv storepb.Store_Serie
 				mtx.Unlock()
 
 				return nil
-			}, func(err error) {
-				if err != nil {
```
bwplotka: You are aware that this here made sure that `run.Group` was behaving like `errgroup`?

bwplotka:
Should we rebase and merge? (:  @povilasv

PR #1469, opened 2019-08-28
store: Prevent EOF errors in Azure objstore

### LINKED ISSUE #1466: store: Azure hits EOF when reading last metric in a chunk
Replicated on v0.3.2, v0.4.0, v0.6.0, and v0.6.1

**What happened**
Store gets an unexpected EOF when trying to read in a chunk because the Azure library uses an `io.ReadFull` call with an ending point that is longer than the file.

This error does not get passed up the stack by oklog/run when the group is run in 
`func (s *BucketStore) Series(req *storepb.SeriesRequest, srv storepb.Store_SeriesServer) (err error)`, which made this more fun to debug :smile: 

Either Thanos to not request an ending point that is larger than the file, or the Azure library to handle it gracefully.

I don't think Thanos knows ahead of time how big the chunk is, so I'm leaning towards Azure's fault. This leads me to think that:
```
		parts := r.block.partitioner.Partition(len(offsets), func(i int) (start, end uint64) {
			return uint64(offsets[i]), uint64(offsets[i]) + maxChunkSize
		})
```

This may be a complete coincidence, but this metric is the exact last one I have in Prometheus when sorted alphabetically. Try using Azure Blob and querying your last metric?

I have already fixed this issue in the Azure library. I'm going to test it more, submit a PR to their repo, and then I'll submit a PR here to bump the version in go.mod

Additionally, I think it'd be worth looking into why using github.com/oklog/run doesn't report errors that the group encounters. Maybe we should add Error logging in the functions being run?

wbh1:
closes #1466 

* [#1469] Prevent EOF errors in Azure objstore

## Changes

Make Azure objstore's GetRange smarter by recognizing when the requested size will be greater than the remaining length of the file to be downloaded. This prevents EOF errors described in #1466 

## Verification

Verified compaction, initial download, and querying the store node all work (tested against Azure Blob with ~400GB of data)

### DIFF
--- a/pkg/objstore/azure/azure.go
+++ b/pkg/objstore/azure/azure.go
@@ -190,15 +190,19 @@ func (b *Bucket) getBlobReader(ctx context.Context, name string, offset, length 
 	}
 
 	var size int64
-	if length > 0 {
+	// If a length is specified and it won't go past the end of the file,
+	// then set it as the size.
+	if length > 0 && length <= props.ContentLength()-offset {
 		size = length
+		level.Debug(b.logger).Log("msg", "set size to length", "size", size, "length", length, "offset", offset, "name", name)
 	} else {
 		size = props.ContentLength() - offset
+		level.Debug(b.logger).Log("msg", "set size to go to EOF", "contentlength", props.ContentLength(), "size", size, "length", length, "offset", offset, "name", name)
 	}
 
 	destBuffer := make([]byte, size)
 
-	if err := blob.DownloadBlobToBuffer(context.Background(), blobURL.BlobURL, offset, length,
+	if err := blob.DownloadBlobToBuffer(context.Background(), blobURL.BlobURL, offset, size,
 		destBuffer, blob.DownloadFromBlobOptions{
 			BlockSize:   blob.BlobDefaultDownloadBlockSize,
 			Parallelism: uint16(3),


pkg/objstore/azure/azure.go
```diff
@@ -190,15 +190,19 @@ func (b *Bucket) getBlobReader(ctx context.Context, name string, offset, length
 	}
 
 	var size int64
-	if length > 0 {
+	// If a length is specified and it won't go past the end of the file,
+	// then set it as the size
```
bwplotka: missing trailing period in the sentence.

pkg/objstore/azure/azure.go
```diff
@@ -190,15 +190,19 @@ func (b *Bucket) getBlobReader(ctx context.Context, name string, offset, length
...
+	if length > 0 && length <= props.ContentLength()-offset {
```
bwplotka: This does not look right.

AFAIK there is no known case in immutable system that Thanos would ask for something that does not exists unless TSDB block is malformed (malformed index, index pointing to not existsing chunks). Is that the case?

Even in this case this change would delegate error to Thanos which would see wrong result from Azure anyway.

wbh1: hey @bwplotka, just pinging you for another look.

btw, congrats on the Redhat move! :smile:

devalexx: @bwplotka I can confirm that it works or me too

PR #1470, opened 2019-08-28
Release 0.7 RC Updating vesion and changelog

domgreen:
Updated VERSION and CHANGELOG to reflect changes that we see going out in this release.

---
This change is [Reviewable](https://reviewable.io/reviews/thanos-io/thanos/1470)


FUSAKLA:
maybe we could add some basic example how to format changelog entries so the release shepherd does not need to fix the changelog that much? it could possibly go to the https://github.com/thanos-io/thanos/blob/master/.github/PULL_REQUEST_TEMPLATE.md

bwplotka:
Yup, we could put more details good point @FUSAKLA 

It's a lot of manual job for us.

Issue #1093, opened 2019-04-29
Thanos receive component umbrella issue


brancz:
This is an umbrella issue to track the development of the accepted [proposal for the Thanos receive component](https://github.com/improbable-eng/thanos/blob/master/docs/proposals/approved/201812_thanos-remote-receive.md).

* [x] remote-write protocol accepting API https://github.com/improbable-eng/thanos/pull/811
* [x] upload blocks produced by local tsdb https://github.com/improbable-eng/thanos/pull/1011 https://github.com/improbable-eng/thanos/pull/1046
* [x] hashring for sharded storage (as a start to not be blocked by the next item, we expect that all requests are sent to all receive nodes and they persist only the time-series assigned to them) https://github.com/improbable-eng/thanos/pull/1217
* [x] split and forward requests for time-series to receive nodes according to hashring https://github.com/improbable-eng/thanos/pull/1243
* [x] create mechanism to force flush and upload, while not accepting writes (necessary for hashring resizing)
* [x] announce multiple sets of external labels through StoreAPI info response #1284 (this does not only influence the receive component, but will also allow sharding distinct sets of blocks by the store component as well, in addition to time sharding as introduced by https://github.com/improbable-eng/thanos/pull/1077)
* [x] add management of multiple tsdb instances based on write request header input
* [x] dynamo style replication and read/write quorum. This was only part of the "Open Questions" in the design doc, but it is becoming increasingly clear that this will need to be necessary, but likely still the last thing to implement. https://github.com/improbable-eng/thanos/pull/1270
* [x] component documentation

Let me know if any of this is unclear or needs more detail, should the design doc not have enough.

@bwplotka @domgreen @povilasv @metalmatze @squat @mxinden @paulfantom @s-urbaniak

bwplotka:
I think it would be nice to also document the `receive` like we do for other commands here: https://thanos.io/getting-started.md/

dawidmalina:
@bwplotka what is actual state of the receive implementation? I mean how far we are from having this ready for wider audience (at least for test purpose)?

dawidmalina:
Good to hear that 👍 but I'm still suspicious about definition of the word "everything". According to the number of todo tasks attached to this ticket we have only one topic related to management of multiple tsdb instances. Is this the only one to consider receive implementation "ready for a wider audience" or we still rather stay away and wait until other features will be implemented?

brancz:
[...]

PR #1480, opened 2019-08-31
cmd,pkg: flush storage when hashring changes

squat:
This PR implements one of the last tasks of the Receive proposal
(tracked in https://github.com/thanos-io/thanos/issues/1093).

In order to allow the hashring of receive nodes to scale at runtime, we
need to temporarily stop the storage, flush it, and upload the blocks.
This commit ensures that whenever a hashring change is detected, the
entire storage is flushed and the shipper is notified so that is uploads
any new blocks. It also ensures that when the receive component starts
up, any in-progress WAL is flushed and uploaded. This ensures that
new data that may belong to a different tenant is not mixed with old
data. Finally, it also ensures that the storage is flushed and uploaded
whenever the process shuts down.

## Changes
* flush storage and upload blocks on startup
* stop, flush, and restart storage whenever hashring changes
* flush storage and upload blocks on shutdown

## Verification
Manual testing of modifying hashrings at runtime and confirming
WAL is flushed to blocks.


cmd/thanos/receive.go
```diff
@@ -321,23 +367,145 @@ func runReceive(
...
+			}, func(error) {
+				cancel()
+				close(uploadDone)
+			})
+		}
 	}
 
+	// Before quitting, ensure the WAL is flushed.
+	defer func() {
+		db, err := tsdb.Open(
```
brancz: I don't see a `db.Close()` anywhere

cmd/thanos/receive.go
```diff
@@ -321,23 +367,145 @@ func runReceive(
...
 
-			return runutil.Repeat(30*time.Second, ctx.Done(), func() error {
-				if uploaded, err := s.Sync(ctx); err != nil {
-					level.Warn(logger).Log("err", err, "uploaded", uploaded)
-				}
+		{
+			// Run the uploader in a loop.
+			ctx, cancel := context.WithCancel(context.Background())
+			g.Add(func() error {
+				return runutil.Repeat(30*time.Second, ctx.Done(), func() error {
```
brancz: seems like we could just write to the "on demand" channel every 30 seconds no?

cmd/thanos/receive.go
```diff
@@ -251,34 +296,46 @@ func runReceive(
...
+				startGRPC <- struct{}{}
+			}
+			return nil
 		}, func(error) {
 			if s != nil {
 				s.Stop()
 			}
+			close(startGRPC)
 		})
+		// We need to be able to start and stop the gRPC server
```
brancz: I don't understand why we need this to be separate from the above run group. It seems the above one immediately triggers this one when sending to the `startGRPC` channel. What would the difference be if we just combined the two?

cmd/thanos/receive.go
```diff
@@ -142,54 +144,100 @@ func runReceive(
...
-			func(err error) {
-				if err := localStorage.Close(); err != nil {
-					level.Error(logger).Log("msg", "error stopping storage", "err", err)
+		// Before actually starting, we need to make sure
+		// the WAL is flushed.
+		startTimeMargin := int64(2 * time.Duration(tsdbCfg.MinBlockDuration).Seconds() * 1000)
+		db, err := tsdb.Open(
+			dataDir,
+			log.With(logger, "component", "tsdb"),
+			&unRegisterer{reg},
```
brancz: I assume we do this, because we instantiate the tsdb multiple times. Why don't we instead keep the reference to it?

squat: agh, that is a very good idea, I'll refactor

cmd/thanos/receive.go
squat: the gRPC server needs to be restartable, so this means that the routine that stops and starts the server should not block but `s.Serve(l)` is blocking, so we want to run the gRPC server in it's own go-routine. "combining the two" means that rather than use a run group we simply do `go func() {if err := s.Serve(l)...}()`. Either way there's a new go routine. The difference is that now the gRPC server can't cause the run group to exit when it experiences an error.

cmd/thanos/receive.go
squat: ack sounds nice :)

cmd/thanos/receive.go
squat: :+1:

cmd/thanos/receive.go
squat: hmm the issue I see with that is that now we have two concurrent writers to the chan and only one of them has control of closing the chan. This is means we will have racy behavior of writing to a closed chan

cmd/thanos/receive.go
squat: the problem is that we need to close the tsdb every time we flush it. once closed, we can't re-open the same reference. Is there another way?

cmd/thanos/receive.go
squat: the `db.Close()` is in the flush func

cmd/thanos/receive.go
krasi-georgiev: on another note it is best to open the db in Read only mode when doing the snapshot. In general snapshoting is safe on a "live" db, but IIRC there is a small chance that a compaction can start during the snapshot and lead to some unexpected results. 
Reltaed discussions: https://github.com/prometheus/tsdb/issues/346#issuecomment-520786524

snapshoting is not implemented for the read only mode, but it should be just few extra lines and I can open a PR for that.

krasi-georgiev: > 			&unRegisterer{reg},

do you even use the registerer? why not just pass nil?

cmd/thanos/receive.go
krasi-georgiev: Why not keep it in the block where the db is opened to avoid the confusion?

cmd/thanos/receive.go
squat: we can't pass nil because we actually do use the new DB. if we pass nil then we lose the metrics

cmd/thanos/receive.go
brancz: I don't see why it needs to be restarted. I'm probably totally missing it :slightly_smiling_face: .

cmd/thanos/receive.go
squat: Flushing has been moved to a new type with a `Flush` method. The method docstring documents this, so I think that is plenty

cmd/thanos/receive.go
squat: We need to restart because the actual tsdb instance used by in `tsdbStore := store.NewTSDBStore(log.With(logger, "component", "thanos-tsdb-store"), nil, localStorage.Get(), component.Receive, lset)` is rotated after ever re-sharding. So before every restart we create a new `tsdbStore`, make a new gRPC server from that `tsdbStore` instance and start the server

cmd/thanos/receive.go
squat: @krasi-georgiev yes, reading through the TSDB code I saw that indeed there is a potential race with the compaction routine. It would be great to add snapshotting support to the read-only DB to avoid the problem with the race. In that case the Flush method would:
1. close the DB
2. open a read only DB
3. snapshot
4. close the read-only DB

I can happily make the PR and cc you on it

brancz: we have compaction completely turned off the way we run tsdb in the receive component, wouldn't this be a non-issue because of that?

krasi-georgiev: @brancz 	I remember that when I was working on the db read only mode there was no easy way to disable compaction and ensure all mutable operations are disabled. 

@squat 	yep feel free to open the PR and I will review it quickly. should be just few lines.

pkg/receive/tsdb.go
```diff
@@ -0,0 +1,132 @@
...
+	defer f.mu.Unlock()
+	if f.stopped {
+		return nil
+	}
+	tmp := filepath.Join(f.Dir(), "tmp")
+	if err := os.Mkdir(tmp, os.ModePerm); err != nil && !os.IsExist(err) {
+		return errors.Wrap(err, "making tmp dir")
+	}
+	defer os.RemoveAll(tmp)
+	f.DisableCompactions()
```
krasi-georgiev: logically this is the right place, but `DisableCompactions` has no effect on started compactions.
For example a call to `DisableCompactions` in the middle of a running compaction will not stop it so a call to `Snapshot` would still happen in the middle of a running compaction.

krasi-georgiev:
why is the db run without a lock file?
https://github.com/thanos-io/thanos/pull/1480/files#diff-074deab4cb04b87eb84800f5bbd8221fL139

pkg/receive/tsdb.go
squat: Ack, we’ll switch over to the DBReadOnly implementation to make this work correctly :+1:

squat:
| why is the db run without a lock file?

Hm does this PR introduce a regression?

brancz:
It's not a regression, it's intentional as we always run this component in environments where the orchestration system ensure there is only one process connected to a volume, hence no need for a lockfile.

cmd/thanos/receive.go
brancz: got it, that makes perfect sense

pkg/receive/tsdb.go
krasi-georgiev: Yes `DBReadOnly` would the safest approach in the long run, but in this case it might also work if you disable the compaction as soon as you open the db.

cmd/thanos/receive.go
```diff
@@ -321,23 +359,78 @@ func runReceive(
...
 	level.Info(logger).Log("msg", "starting receiver")
 
 	return nil
 }
+
+// sendOrCancel allows us to signal a chan and still
+// quit early if the receiver has already quit.
+func sendOrCancel(cancel <-chan struct{}, ch chan<- struct{}) {
+	select {
+	case <-cancel:
```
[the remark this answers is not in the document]
squat: the idea here was to let the send be cancel-able, not simply non-blocking.

cmd/thanos/receive.go
```diff
@@ -140,56 +140,96 @@ func runReceive(
...
-				level.Info(logger).Log("msg", "server is ready to receive web requests.")
-				close(dbOpen)
-				<-cancel
-				return nil
-			},
-			func(err error) {
-				if err := localStorage.Close(); err != nil {
-					level.Error(logger).Log("msg", "error stopping storage", "err", err)
+		// Before actually starting, we need to make sure
+		// the WAL is flushed.
```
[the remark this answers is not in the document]
squat: Just for correctness; if for some reason the last time the receiver ran it quit unexpectedly and was not able to flush the WAL on shutdown, we want to make sure it is flushed so that the new blocks don't have mixed data.

cmd/thanos/receive.go
```diff
@@ -140,56 +140,96 @@ func runReceive(
...
+		startTimeMargin := int64(2 * time.Duration(tsdbCfg.MinBlockDuration).Seconds() * 1000)
+		if err := db.Open(); err != nil {
```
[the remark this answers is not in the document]
squat: Yes we could do that. I was worried that it would be confusing if flush left an open DB. Does you think it would be more expected for Flush to leave a running DB?

pkg/receive/tsdb.go
squat: xref https://github.com/prometheus/prometheus/pull/6006

cmd/thanos/receive.go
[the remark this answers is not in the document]
squat: Yes we could do that. I was worried that it would be confusing if flush left an open DB. Does you think it would be more expected for Flush to leave a running DB?

cmd/thanos/receive.go
```diff
@@ -321,23 +359,78 @@ func runReceive(
 
 		s := shipper.New(logger, reg, dataDir, bkt, func() labels.Labels { return lset }, metadata.ReceiveSource)
 
-		ctx, cancel := context.WithCancel(context.Background())
-		g.Add(func() error {
-			defer runutil.CloseWithLogOnErr(logger, bkt, "bucket client")
+		// Before starting, ensure any old blocks are uploaded.
+		if uploaded, err := s.Sync(context.Background()); err != nil {
```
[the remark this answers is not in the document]
squat: The whole idea is that we want this to block so nothing is ingested before old blocks are flushed and uploaded.

cmd/thanos/receive.go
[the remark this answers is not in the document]
krasi-georgiev: I think it is most logical that flush should: `open the db -> flush -> close the db`

cmd/thanos/receive.go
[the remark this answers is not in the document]
krasi-georgiev: don't you think that it will cause issues when you try to exit the receiver in the middle of an upload and it blocks here?

cmd/thanos/receive.go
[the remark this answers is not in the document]
krasi-georgiev: got it thanks.

cmd/thanos/receive.go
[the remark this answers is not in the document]
squat: I already had this discussion with @brancz on this PR, but I'm not sure where GitHub put that comment history :/ . The original implementation had a 10-minute timeout on the context and the suggestion was to remove that. Let's cover that question in a follow rather than go back and forth.

cmd/thanos/receive.go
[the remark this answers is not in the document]
squat: there are two  issues I see with that:

1. `f.Open()` returns an error, so we need to error check `f.Get()`
2. if flush expects a closed DB, then the caller has to call `close` before flushing.

Both of these mean more work for the caller. For example:

```go
// Using DB
if err := f.Close(); err != nil {
}
if err := f.Flush(); err != nil {
}
db, err := f.Get()
if err != nil {
}
// Using DB
```
instead of simply:
```go
// Using DB
if err := f.Flush(); err != nil {
}
db := f.Get()
// Using DB
```

cmd/thanos/receive.go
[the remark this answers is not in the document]
squat: The cancelable contexts by themselves are not enough, I think. Take the case of the run group managing the DB. Whenever the DB gets updated, it needs to be flushed and then it needs to signal to uploader that it should upload the blocks. Here the uploading and flushing is synchronous but not preemptible:
```go
for {
    select {
    case <- cancel:
        return nil
    case <- dbUpdate:
        db.Flush()
        uploadC <- struct{}{}
        <-uploadDone
    }
}
```
The uploading is blocking, this means that if we need want to cancel the run group for the db manager, it will have to wait for uploading to finish. Also, if the uploader has already exited, then no one will every receive on `uploadC` and the flushing run group will be blocked forever. However, thinking about this again, I think we can get away with simply using buffered chans. This will ensure we can synchronize the run groups and still not block when we need to cancel. I'll post that change, which should simplify things a bit (:

cmd/thanos/receive.go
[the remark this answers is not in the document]
krasi-georgiev: this can also happen internally, no?
```
if !db.closed {
	db.Close()
}
.
.
```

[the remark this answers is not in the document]
squat: Sure I just want to be clear about the API. I think it’s confusing if Flush leaves an open DB if the original state was closed. It may be clearest to say `Flush will always leave the storage in the same state it was in: open storage will stay open; closed storage will remain closed.`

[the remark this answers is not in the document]
krasi-georgiev: yep , that is a great idea, but can you imagine a case where the storage would be open when running a flush?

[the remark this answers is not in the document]
squat: The DB is open immediately before every flush (we are actively writing samples until we get the reshard signal and need to flush), someone has to close it, either the caller, or the flush method itself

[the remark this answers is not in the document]
krasi-georgiev: hm yeah, maybe now with the new tsdb.Flush method can just have just use it directly and not need the separate `FlushableStorage`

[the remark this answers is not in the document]
squat: Yes that’s what I was driving at earlier. It will reduce the code here, though we’ll have some more boilerplate in the caller since it will have to open a new DB after flushing. This type is convenient because we can have one single reference that takes care of re-opening itself after flushing if it was already open and the only thing the caller has to do is call `Open` once at the very beginning with all the args and then `Flush`

squat:
ping @krasi-georgiev @brancz PTAL again (:

pkg/receive/tsdb.go
```diff
@@ -0,0 +1,123 @@
...
+}
+
+// Close stops the storage.
+func (f *FlushableStorage) Close() error {
+	f.mu.Lock()
+	defer f.mu.Unlock()
+	if f.stopped {
+		return nil
+	}
+	defer func() { f.stopped = true }()
```
brancz: this reads quite tricky, even if I think this is correct, I think I'd prefer to expand this into a couple of lines

squat: ack, expanded this to be more legible (: no need to code-golf

pkg/receive/hashring.go
```diff
@@ -165,8 +165,10 @@ func newMultiHashring(cfg []HashringConfig) Hashring {
 // Hashrings are returned on the updates channel.
 // Which hashring to use for a tenant is determined
 // by the tenants field of the hashring configuration.
+// The updates chan is closed before exiting.
 func HashringFromConfig(ctx context.Context, updates chan<- Hashring, cw *ConfigWatcher) {
 	go cw.Run(ctx)
+	defer close(updates)
```
krasi-georgiev: isn't it a widely accepted that only  the sender should close the channel?

pkg/receive/tsdb.go
```diff
@@ -0,0 +1,129 @@
...
+	if err := f.DB.Close(); err != nil {
+		return err
+	}
+	f.stopped = true
+	return nil
+}
+
+// UnRegisterer is a Prometheus registerer that
+// ensures that collectors can be registered
+// by unregistering already-registered collectors.
```
krasi-georgiev: I am sure I will forget very soon why did we need this :smile:  so  maybe mention somewhere that this is used to not loose metrics between db flushes.

pkg/receive/handler.go
```diff
@@ -95,9 +87,12 @@ func NewHandler(logger log.Logger, o *Options) *Handler {
 	return h
 }
 
-// StorageReady marks the storage as ready.
-func (h *Handler) StorageReady() {
-	atomic.StoreUint32(&h.storageReady, 1)
+// Storage sets the writer and marks the storage as ready.
+// If the writer is nil, then the storage is marked as not ready.
+func (h *Handler) Storage(w *Writer) {
```
krasi-georgiev: Shouldn't this be called "SetWriter"?

and just comment that it needs to be called before using the endpoint

pkg/receive/hashring.go
squat: agreed, I thought it would be confusing that we don't close it in the config-watcher (hashring) case but we do close it in the single node case, but I'll add a comment comment to explain.

pkg/receive/tsdb.go
squat: :+1:

pkg/receive/handler.go
squat: naming is hard!

pkg/receive/hashring.go
squat: This goroutine IS the sender. In the single-node case, where there is no configuration watcher we have to close it ourselves.

krasi-georgiev: sorry for my ignorance, could you expand a bit more on that, 

I see no reason why it can't be closed somewhere in `runReceive` ?

squat: The reason we don't want to close it in `runReceive` is because by convention chans should always be closed by their sender to avoid races that `send on closed chan`. In this case, the `HashringFromConfig` func runs a goroutine that is the only sender of the given chan, and as such it should be the one to close it. Otherwise, we need to add another layer of synchronization to always ensure that the generator goroutine has returned *before* we close the given chan. This adds a lot of complexity for little benefit IMO.

squat: Ping @krasi-georgiev

krasi-georgiev: yes 100% correct, just got confused with the channel directions. Thanks for the patience!

krasi-georgiev:
LGTM

PR #1487, opened 2019-09-02
Some more specific info and setting sampler_param to 1 by default

sudeepknair:
Some more specific info which folks would find helpful and setting sampler_param to 1 by default.

People when referring to the yaml might end up setting `sampler_param` to `0` because of which no trace is captured. 

## Verification
Checked on MD editor


sudeepknair:
@bwplotka Please review

sudeepknair:
Sounds good @bwplotka

Issue #713, opened 2019-01-07
Swift upload: "new bucket block: list chunk files: Cannot extract names from response with content-type: []"


tivvit:
Hi

@sudhi-vm 

**Thanos, Prometheus and Golang version used**
thanos, version 0.2.1 (branch: HEAD, revision: 30e7cbdafd3ef01189f202945c2728fcf37e1cf1)
  build user:       root@79ffcf51ff9b
  build date:       20181227-15:44:56
  go version:       go1.11.4

**What happened**
I am having a problem with SWIFT store API

I can see it both in the sidecar and in store

**What you expected to happen**
The sidecar should be able to close files

The store does not even start

**How to reproduce it (as minimally and precisely as possible)**:
started sidecar with configured SWIFT

**Full logs to relevant components**
Here is the log from the sidecar
```
level=info ts=2019-01-07T13:00:32.764722917Z caller=shipper.go:201 msg="upload new block" id=01D0M6C05AQ76KYERF02E7JRNY
level=warn ts=2019-01-07T13:00:32.827598341Z caller=runutil.go:69 msg="detected close error" err="close file /data/thanos/upload/01D0M6C05AQ76KYERF02E7JRNY/meta.json: close /data/thanos/upload/01D0M6C05AQ76KYERF02E7JRNY/meta.json: file already closed"
level=warn ts=2019-01-07T13:00:36.581858197Z caller=runutil.go:69 msg="detected close error" err="close file /data/thanos/upload/01D0M6C05AQ76KYERF02E7JRNY/chunks/000001: close /data/thanos/upload/01D0M6C05AQ76KYERF02E7JRNY/chunks/000001: file already closed"
level=warn ts=2019-01-07T13:00:37.212250194Z caller=runutil.go:69 msg="detected close error" err="close file /data/thanos/upload/01D0M6C05AQ76KYERF02E7JRNY/index: close /data/thanos/upload/01D0M6C05AQ76KYERF02E7JRNY/index: file already closed"
level=warn ts=2019-01-07T13:00:37.28117964Z caller=runutil.go:69 msg="detected close error" err="close file /data/thanos/upload/01D0M6C05AQ76KYERF02E7JRNY/meta.json: close /data/thanos/upload/01D0M6C05AQ76KYERF02E7JRNY/meta.json: file already closed"
```
and from store
```
level=info ts=2019-01-07T13:08:19.353763442Z caller=factory.go:39 msg="loading bucket configuration"
[...]

PR #1489, opened 2019-09-03
swift: fix file already closed issue

FlorinPeter:
* Fix file already closed issue on Swift (#713)

## Changes

Fix file already closed issue by passing a reader that supports seek to the swift implementation.

## Verification
Tested locally without fix:
```
$ go test -v ./pkg/shipper/
=== RUN   TestShipper_SyncBlocks_e2e
=== RUN   TestShipper_SyncBlocks_e2e/inmem
level=info msg="upload new block" id=000000000006AFVGQT5ZYC0GEK
level=info msg="upload new block" id=0000000001ZW908PVKS1Q4ZYAZ
level=info msg="upload new block" id=0000000002YSHABVQ85AYZ8JHD
level=info msg="upload new block" id=00000000031DTZP60DNX4AF7Q0
level=info msg="upload new block" id=0000000004P46KJHJHGM7X98BR
level=info msg="upload new block" id=000000000611TXCKQ2TF8D1QKB
level=info msg="upload new block" id=0000000007Z3WV8K78BZR49HNH
level=info msg="upload new block" id=0000000008Z0XRX21VQY2QNAWS
level=info msg="upload new block" id=0000000009RPS55HT2KGSF7A5E
=== RUN   TestShipper_SyncBlocks_e2e/swift
level=info msg="upload new block" id=000000000006AFVGQT5ZYC0GEK
level=warn msg="detected close error" err="close file /var/folders/rl/1l703cjd6d93_4f6ktp13cpr0000gn/T/shipper-e2e-test093408693/thanos/upload/000000000006AFVGQT5ZYC0GEK/meta.json: close /var/folders/rl/1l703cjd6d93_4f6ktp13cpr0000gn/T/shipper-e2e-test093408693/thanos/upload/000000000006AFVGQT5ZYC0GEK/meta.json: file already closed"
level=warn msg="detected close error" err="close file /var/folders/rl/1l703cjd6d93_4f6ktp13cpr0000gn/T/shipper-e2e-test093408693/thanos/upload/000000000006AFVGQT5ZYC0GEK/chunks/0001: close /var/folders/rl/1l703cjd6d93_4f6ktp13cpr0000gn/T/shipper-e2e-test093408693/thanos/upload/000000000006AFVGQT5ZYC0GEK/chunks/0001: file already closed"
level=warn msg="detected close error" err="close file /var/folders/rl/1l703cjd6d93_4f6ktp13cpr0000gn/T/shipper-e2e-test093408693/thanos/upload/000000000006AFVGQT5ZYC0GEK/chunks/0002: close /var/folders/rl/1l703cjd6d93_4f6ktp13cpr0000gn/T/shipper-e2e-test093408693/thanos/upload/000000000006AFVGQT5ZYC0GEK/chunks/0002: file already closed"
level=warn msg="detected close error" err="close file /var/folders/rl/1l703cjd6d93_4f6ktp13cpr0000gn/T/shipper-e2e-test093408693/thanos/upload/000000000006AFVGQT5ZYC0GEK/index: close /var/folders/rl/1l703cjd6d93_4f6ktp13cpr0000gn/T/shipper-e2e-test093408693/thanos/upload/000000000006AFVGQT5ZYC0GEK/index: file already closed"
level=warn msg="detected close error" err="close file /var/folders/rl/1l703cjd6d93_4f6ktp13cpr0000gn/T/shipper-e2e-test093408693/thanos/upload/000000000006AFVGQT5ZYC0GEK/meta.json: close /var/folders/rl/1l703cjd6d93_4f6ktp13cpr0000gn/T/shipper-e2e-test093408693/thanos/upload/000000000006AFVGQT5ZYC0GEK/meta.json: file already closed"
...
```

With the fix:
```
$ go test -v ./pkg/shipper/
=== RUN   TestShipper_SyncBlocks_e2e
=== RUN   TestShipper_SyncBlocks_e2e/inmem
level=info msg="upload new block" id=000000000006AFVGQT5ZYC0GEK
level=info msg="upload new block" id=0000000001ZW908PVKS1Q4ZYAZ
level=info msg="upload new block" id=0000000002YSHABVQ85AYZ8JHD
level=info msg="upload new block" id=00000000031DTZP60DNX4AF7Q0
level=info msg="upload new block" id=0000000004P46KJHJHGM7X98BR
level=info msg="upload new block" id=000000000611TXCKQ2TF8D1QKB
level=info msg="upload new block" id=0000000007Z3WV8K78BZR49HNH
level=info msg="upload new block" id=0000000008Z0XRX21VQY2QNAWS
level=info msg="upload new block" id=0000000009RPS55HT2KGSF7A5E
=== RUN   TestShipper_SyncBlocks_e2e/swift
level=info msg="upload new block" id=000000000006AFVGQT5ZYC0GEK
level=info msg="upload new block" id=0000000001ZW908PVKS1Q4ZYAZ
level=info msg="upload new block" id=0000000002YSHABVQ85AYZ8JHD
level=info msg="upload new block" id=00000000031DTZP60DNX4AF7Q0
level=info msg="upload new block" id=0000000004P46KJHJHGM7X98BR
level=info msg="upload new block" id=000000000611TXCKQ2TF8D1QKB
level=info msg="upload new block" id=0000000007Z3WV8K78BZR49HNH
level=info msg="upload new block" id=0000000008Z0XRX21VQY2QNAWS
level=info msg="upload new block" id=0000000009RPS55HT2KGSF7A5E
--- PASS: TestShipper_SyncBlocks_e2e (35.63s)
    --- PASS: TestShipper_SyncBlocks_e2e/inmem (0.45s)
    foreach.go:48: THANOS_SKIP_GCS_TESTS envvar present. Skipping test against GCS.
    foreach.go:70: THANOS_SKIP_S3_AWS_TESTS envvar present. Skipping test against S3 AWS.
    foreach.go:86: THANOS_SKIP_AZURE_TESTS envvar present. Skipping test against Azure.
    swift.go:313: created temporary container for swift tests with name test_testshipper_syncblocks_e2e_456bf734519ee3fa
    --- PASS: TestShipper_SyncBlocks_e2e/swift (23.91s)
    foreach.go:118: THANOS_SKIP_TENCENT_COS_TESTS envvar present. Skipping test against Tencent COS.
=== RUN   TestShipper_SyncBlocksWithMigrating_e2e
=== RUN   TestShipper_SyncBlocksWithMigrating_e2e/v2.9.2
level=info msg="upload new block" id=000000000006AFVGQT5ZYC0GEK
level=info msg="upload new block" id=0000000001ZW908PVKS1Q4ZYAZ
level=info msg="upload new block" id=0000000002YSHABVQ85AYZ8JHD
level=info msg="upload new block" id=00000000031DTZP60DNX4AF7Q0
level=info msg="gathering all existing blocks from the remote bucket for check" id=0000000004P46KJHJHGM7X98BR
level=info msg="upload new block" id=0000000004P46KJHJHGM7X98BR
level=info msg="upload new block" id=0000000005H4QE51FCW58H8NBR
level=info msg="upload new block" id=000000000611TXCKQ2TF8D1QKB
level=info msg="upload new block" id=0000000007Z3WV8K78BZR49HNH
level=info msg="upload new block" id=0000000008Z0XRX21VQY2QNAWS
level=info msg="upload new block" id=0000000009RPS55HT2KGSF7A5E
--- PASS: TestShipper_SyncBlocksWithMigrating_e2e (4.82s)
    --- PASS: TestShipper_SyncBlocksWithMigrating_e2e/v2.9.2 (4.82s)
=== RUN   TestShipperTimestamps
--- PASS: TestShipperTimestamps (0.03s)
PASS
ok  	github.com/thanos-io/thanos/pkg/shipper	40.519s
```

### DIFF
--- a/pkg/objstore/swift/swift.go
+++ b/pkg/objstore/swift/swift.go
@@ -2,6 +2,7 @@
 package swift
 
 import (
+	"bytes"
 	"context"
 	"fmt"
 	"io"
@@ -19,7 +20,7 @@ import (
 	"github.com/gophercloud/gophercloud/pagination"
 	"github.com/pkg/errors"
 	"github.com/thanos-io/thanos/pkg/objstore"
-	yaml "gopkg.in/yaml.v2"
+	"gopkg.in/yaml.v2"
 )
 
 // DirDelim is the delimiter used to model a directory structure in an object store bucket.
@@ -149,7 +150,15 @@ func (c *Container) IsObjNotFoundErr(err error) bool {
 
 // Upload writes the contents of the reader as an object into the container.
 func (c *Container) Upload(ctx context.Context, name string, r io.Reader) error {
-	options := &objects.CreateOpts{Content: r}
+	// to avoid file already closed error pass a Reader that supports seek
+	buf := new(bytes.Buffer)
+	_, err := buf.ReadFrom(r)
+	if err != nil {
+		return err
+	}
+	readSeeker := strings.NewReader(buf.String())
+
+	options := &objects.CreateOpts{Content: readSeeker}
 	res := objects.Create(c.client, c.name, name, options)
 	return res.Err
 }


pkg/objstore/swift/swift.go
```diff
@@ -247,9 +256,7 @@ func configFromEnv() SwiftConfig {
 func validateForTests(conf SwiftConfig) error {
 	if conf.AuthUrl == "" ||
 		conf.Username == "" ||
-		conf.Password == "" ||
-		(conf.ProjectName == "" && conf.ProjectID == "") ||
-		conf.RegionName == "" {
+		conf.Password == "" {
```
sudhi-vm: @FlorinPeter One of `ProjectName`, `ProjectID` is mandatory for OpenStack client initialization. Any particular reason to relax these checks?

pkg/objstore/swift/swift.go
```diff
@@ -149,7 +150,15 @@ func (c *Container) IsObjNotFoundErr(err error) bool {
...
 // Upload writes the contents of the reader as an object into the container.
 func (c *Container) Upload(ctx context.Context, name string, r io.Reader) error {
-	options := &objects.CreateOpts{Content: r}
+	// to avoid file already closed error pass a Reader that supports seek
+	buf := new(bytes.Buffer)
+	_, err := buf.ReadFrom(r)
+	if err != nil {
+		return err
+	}
+	readSeeker := strings.NewReader(buf.String())
```
sudhi-vm: This issue seems to be due to a bug in the vendored gophercloud dependency and looks like it has been fixed upstream: https://github.com/rackspace/gophercloud/blob/master/openstack/objectstorage/v1/objects/requests.go#L219. Do you think you could re-vendor gophercloud and run your tests?

pkg/objstore/swift/swift.go
FlorinPeter: @sudhi-vm I relaxed it because I had some issues with my credentials:
```
export OS_AUTH_URL=https://some.url.com/v3
export OS_USERNAME=username
export OS_PASSWORD=password
export OS_PROJECT_NAME=some-projekt-name
export OS_USER_DOMAIN_NAME=some-user-domain-name
```
But it looks like I removed maybe to much ;)

pkg/objstore/swift/swift.go
FlorinPeter: sure I can check it

FlorinPeter: @sudhi-vm looks like gophercloud was moved from github.com/rackspace to https://github.com/gophercloud and it looks like the version 0.3.0 is the latest one.
I am not an io.Reader expert but I guess the ioutil.ReadAll will close the reader and causes this issue.
https://github.com/gophercloud/gophercloud/blob/master/openstack/objectstorage/v1/objects/requests.go#L198-L199

sudhi-vm:
retest ci

sudhi-vm:
@povilasv Good point. We upgraded the vendored library (gophercloud) to recent version locally, but it did not fix the issue. I will follow up with gophercloud on this.

As you mention, we could merge this and treat it as a short-term fix. Once the fix is identified and merged in gophercloud, we could consume the change and also upgrade the vendored library to more recent version.

GiedriusS:
Has this been reported upstream? (:

acukan:
Are there any progress about this issue?

acukan:
ping @sudhi-vm

GiedriusS:
FWIW I still haven't gotten around to this because I cannot get my hands on a working Swift cluster. Seems like the community projects which let you run Swift in a small deployment unit are buggy and/or something doesn't work :'(

GiedriusS:
Reported upstream at https://github.com/gophercloud/gophercloud/issues/1856.

jtopjian:
Hi folks, I'm currently looking at the issue @GiedriusS reported at https://github.com/gophercloud/gophercloud/issues/1856. There are some details to my current investigation over there. If anyone has any pointers, it would be very much appreciated!

jtopjian:
I was able to get the Thanos S3 e2e tests working and then dug into the minio library to compare behavior. I think I've determined what's going on - there's always room for error, though.

At least for the e2e tests, the calculated partSize is always 0 because the files are small and/or they're less than the default vaules of 128mb:

https://github.com/thanos-io/thanos/blob/master/pkg/objstore/s3/s3.go#L321-L325

Minio implements a 128mb size as a threshold on when to do multipart/streaming uploads:

https://github.com/minio/minio-go/blob/master/api-put-object.go#L171-L174

When the size of the object is less than the default 128mb or whatever was explicitly passed, the following happens:

https://github.com/minio/minio-go/blob/master/api-put-object-streaming.go#L335-L344

ta-da! The object data is being wrapped into something that doesn't implement `Close()`. Thus when the Transport closes the body, the original file pointer is prevented from being closed. If you remove the highlighted code, you'll see that you get the same error being reported in Swift. 

Minio is providing a lot of logic for how the data should be handled prior to being uploaded to S3. This amount of care is intentionally not in Gophercloud.

Gophercloud is meant to be a very low-level API. Its core goal is to model the OpenStack REST API with as little convenience and helper functions/methods as possible. This is because OpenStack is a very flexible set of projects and the APIs can be customized from one environment to the next. Implementing any sort of help/convenience opens the door to _a lot_ of help/convenience to account for everyone's needs.

Gophercloud's current Swift support provides all of the capabilities to implement the same kind of logic that Minio is doing internally - it just expects it to be done a layer above Gophercloud. 

We have a repository at https://github.com/gophercloud/utils that is meant to provide helpers/utilities for things that would otherwise be repeated a lot, including support for [uploading and downloading objects](https://github.com/gophercloud/utils/tree/master/openstack/objectstorage/v1/objects), too. I've also [committed a patch](https://github.com/gophercloud/utils/pull/108) to assist the different kinds of data streams to prevent the original/underlying pointer from being prematurely closed, too. 

While I tested this and it works, I don't think this object storage code sees a lot of use, so if you folks choose to adopt it, you might run into some bugs. Please feel free to submit PRs or at least a detailed report so this code can be made better -- you'd end up helping anyone else creating an OpenStack-compatible object storage app.

If you don't want to adopt the utils code, the solution that this PR is implementing is the best that can be done - possibly additionally implementing the same `NewSectionReader` logic in Minio.

Please let me know if anyone has any questions or comments.

GiedriusS:
> I was able to get the Thanos S3 e2e tests working and then dug into the minio library to compare behavior. I think I've determined what's going on - there's always room for error, though.
> 
> At least for the e2e tests, the calculated partSize is always 0 because the files are small and/or they're less than the default vaules of 128mb:
> 
> https://github.com/thanos-io/thanos/blob/master/pkg/objstore/s3/s3.go#L321-L325
> 
> Minio implements a 128mb size as a threshold on when to do multipart/streaming uploads:
> 
> https://github.com/minio/minio-go/blob/master/api-put-object.go#L171-L174
> 
> When the size of the object is less than the default 128mb or whatever was explicitly passed, the following happens:
> 
> https://github.com/minio/minio-go/blob/master/api-put-object-streaming.go#L335-L344
> 
> ta-da! The object data is being wrapped into something that doesn't implement `Close()`. Thus when the Transport closes the body, the original file pointer is prevented from being closed. If you remove the highlighted code, you'll see that you get the same error being reported in Swift.
> 
> Minio is providing a lot of logic for how the data should be handled prior to being uploaded to S3. This amount of care is intentionally not in Gophercloud.
> 
> Gophercloud is meant to be a very low-level API. Its core goal is to model the OpenStack REST API with as little convenience and helper functions/methods as possible. This is because OpenStack is a very flexible set of projects and the APIs can be customized from one environment to the next. Implementing any sort of help/convenience opens the door to _a lot_ of help/convenience to account for everyone's needs.
> 
> Gophercloud's current Swift support provides all of the capabilities to implement the same kind of logic that Minio is doing internally - it just expects it to be done a layer above Gophercloud.
> 
> We have a repository at https://github.com/gophercloud/utils that is meant to provide helpers/utilities for things that would otherwise be repeated a lot, including support for [uploading and downloading objects](https://github.com/gophercloud/utils/tree/master/openstack/objectstorage/v1/objects), too. I've also [committed a patch](https://github.com/gophercloud/utils/pull/108) to assist the different kinds of data streams to prevent the original/underlying pointer from being prematurely closed, too.
> 
> While I tested this and it works, I don't think this object storage code sees a lot of use, so if you folks choose to adopt it, you might run into some bugs. Please feel free to submit PRs or at least a detailed report so this code can be made better -- you'd end up helping anyone else creating an OpenStack-compatible object storage app.
> 
> If you don't want to adopt the utils code, the solution that this PR is implementing is the best that can be done - possibly additionally implementing the same `NewSectionReader` logic in Minio.
> 
> Please let me know if anyone has any questions or comments.

Thanks for your impressive analysis! :muscle: I have one small question: is there some small tool which would let us start a small `Swift` node for e2e testing purposes? Some similar to what `minio` has.

Now about this PR itself... maybe we shouldn't even log at all that a `Close()` had failed **iff** it has been closed already. I think it would be a better solution than all of this (potential) copying or creating new readers. Any other thoughts?

jtopjian:
> Thanks for your impressive analysis!

No problem at all 🙂 

> I have one small question: is there some small tool which would let us start a small Swift node for e2e testing purposes? Some similar to what minio has.

There are a few ways. What you're looking for is something called Swift AIO (all-in-one). There are a bunch of scripts around that will build this. One option is to use the official OpenStack [devstack](https://docs.openstack.org/devstack/latest/) project:

```
git clone https://opendev.org/openstack/devstack
cd devstack

cat > local.conf <<EOF
[[local|localrc]]
ADMIN_PASSWORD=secret
DATABASE_PASSWORD=\$ADMIN_PASSWORD
RABBIT_PASSWORD=\$ADMIN_PASSWORD
SERVICE_PASSWORD=\$ADMIN_PASSWORD
SWIFT_HASH=foobar
disable_all_services
enable_service key mysql s-proxy s-object s-container s-account
EOF

./stack.sh
```

The above will take around 5-6 minutes. When complete, you'll have a local OpenStack environment that only has Keystone (authentication) and Swift running:

```
source openrc
openstack catalog list
openstack container list
openstack container create foo
openstack container list
openstack object create foo local.conf
openstack object list foo
```

> Now about this PR itself... maybe we shouldn't even log at all that a Close() had failed iff it has been closed already. I think it would be a better solution than all of this (potential) copying or creating new readers. Any other thoughts?

IMO, I'm quite thankful that this was being logged. It allowed me to catch an inefficiency, enhance the `gophercloud/utils` code, and learn about readers in general. But I don't think having the object closed from Transport was actually doing any harm -- at least not this time. If it's possible to restrict this logging to only happen in debug mode, that might be a good tradeoff.

> As for the other library, I would love to switch myself but we probably don't have enough human resources for that :(

I totally understand - I'm in the same situation!

Issue #813, opened 2019-02-06
compaction: Be clear about storage retention and downsampling.


bwplotka:
Currently in Thanos compactor,  users can set retention for each resolution:
```
--retention.resolution-raw=...
--retention.resolution-5m=...
--retention.resolution-1h=...
```
If any is left `0` then it is unlimited.

It is natural for user to think that downsampling is something that magically zooms out the metrics and reduces size, so it's tempting to make raw retention super small (because it seems to be expensive) and rest super long as it is seems to be "just" cheap. However it's not that easy (as we can see for number of tickets related to this), so we need to document it better and make it easier to use. 

**Facts:**
* [5m downsampled data are created only after 40h](https://github.com/improbable-eng/thanos/blob/master/cmd/thanos/downsample.go#L172) and there is rather not worth to make it earlier then that.
* [1h downsampled data are created only after 10days](https://github.com/improbable-eng/thanos/blob/master/cmd/thanos/downsample.go#L193), and similar, it is not worth to make it earlier. 1h is created from 5m resolution.
* Downsampled data DOES not need to be smaller really (!) This is because downsampling to stay accurate generates 5 series from single series. It reduces the samples, so for 15s scrape interval, 5m resolution has 20x less samples, 1h has 240x less samples than raw. Depending on scrape interval the gain is smaller or bigger. The problem is that if you have really high cardinally (like several millions), you end up with 5x several millions of series (without overheat from strings size though, as those 4 additional does not have labels). So you might reduce samples (which are highly compressible)  in exchange for series (which are non compressible). It might happen that the size reduction is not that great as you expected, so the best is to check your downsampled block sizes on your own. However the whole point of downsampling is not really size. Is ability to cheaply draw *accurate* graph for long range queries like 1year. 
* Due to accuracy aggregations, special usage of downsampled data is recommended. We don't simply remove samples - that could mask meaningful info especially for gauges. This means that best effect you have with `<>_over_time` aggregations when using downsampled data.
* It is always recommended to leave RAW data with similar retention as others, because it is super valuable to be able to quickly zoom in the past. With just downsampled data you lose that ability, as it is designed only for long range queries. 

**Additionally:**
[...]

PR #1493, opened 2019-09-04
[Docs] - Compact: Add line clarifying that its also responsib…

gjtempleton:
…le for downsampling

## Changes

Added line clarifying that the compact component is also responsible for downsampling and documenting the magic numbers use to determine when to start. 

Partly addresses #813 which appears to be the only place the times downsampling begins are currently documented - have now seen a number of queries about whether or not the `compact` component is responsible for downsampling as well, and this should at leats make that info easier to find.

Was unsure whether to link back to #813 as the central place for discussion of this.

## Verification

N/A


bwplotka:
It helps with https://github.com/thanos-io/thanos/issues/813 indeed, but not fully fix I guess. Thanks anyway!

PR #1495, opened 2019-09-05
docs: added standardized port layout

### LINKED ISSUE #1479: create a port layout (documentation only) to standardize all thanos components on a single host.
I'm currently working on a FreeBSD port of Thanos, and would like to get a standardized layout for having all the thanos components running on a single host. 

Opening this per @FUSAKLA on Slack.

FUSAKLA:
Resolves #1479 

Signed-off-by: Martin Chodur <m.chodur@seznam.cz>

Added port layout for running on single host

Not sure where to put it in the docs so if you think of any better place or form please comment.

### DIFF
--- a/docs/getting-started.md
+++ b/docs/getting-started.md
@@ -244,21 +244,28 @@ Thanos also has a tutorial on deploying it to Kubernetes. We have a full page de
 
 We also have example Grafana dashboards [here](/examples/grafana/monitoring.md) and some [alerts](/examples/alerts/alerts.md) to get you started.
 
-## Single host port layout
-In case you need to run multiple Thanos components on the same host you
-can follow this standardized port layout to avoid port collisions. 
-> The port can be changed using the `--http-address` and `--grpc-address` flags.
+## Testing Thanos on Single Host
 
-| Component | Interface  | Port  |
-| --------- | ---------- | ----- |
-| Sidecar   | gRPC       | 10901 |
-| Sidecar   | HTTP       | 10902 |
-| Query     | gRPC       | 10903 |
-| Query     | HTTP       | 10904 |
-| Store     | gRPC       | 10905 |
-| Store     | HTTP       | 10906 |
-| Receive   | gRPC       | 10907 |
-| Receive   | HTTP       | 10908 |
-| Rule      | gRPC       | 10909 |
-| Rule      | HTTP       | 10910 |
-| Compact   | HTTP       | 10911 |
+We don't recommend running Thanos on a single node on production.
+Thanos is designed and built to run as a distributed system.
+Vanilla Prometheus might be totally enough for small setups.
+
+However, in case you want to play and run Thanos components
+on a single node, we recommend following the port layout:
+
+| Component | Interface               | Port  |
+| --------- | ----------------------- | ----- |
+| Sidecar   | gRPC                    | 10901 |
+| Sidecar   | HTTP                    | 10902 |
+| Query     | gRPC                    | 10903 |
+| Query     | HTTP                    | 10904 |
+| Store     | gRPC                    | 10905 |
+| Store     | HTTP                    | 10906 |
+| Receive   | gRPC (store API)        | 10907 |
+| Receive   | HTTP (remote write API) | 10908 |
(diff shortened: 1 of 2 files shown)


FUSAKLA:
@lrosenman PTAL if this is ok with you

lrosenman:
LGTM.  Would it make sense to also have a consolidated doc or list the "single-host port" in each components list?  (the docs really need some love here).

FUSAKLA:
Yeah, I thought about that if it would be better to have it separated in each component or on a single place. Having it at each component separately would require explain the motivation repeatably which on the other side I wanted to avoid since this case is more of a rare case since Thanos is aiming for the distributed architecture. 

Lets wait for others opinions and will see then. But the port layout will probably stay thus way so you can use it already so you are not blocked.

lrosenman:
My gut is saying this list is good where it is, but we should put a pointer on each components docs to this list maybe?  Not sure.

docs/getting-started.md
```diff
@@ -243,3 +243,22 @@ TBD
 Thanos also has a tutorial on deploying it to Kubernetes. We have a full page describing a standard deployment here.
 
 We also have example Grafana dashboards [here](/examples/grafana/monitoring.md) and some [alerts](/examples/alerts/alerts.md) to get you started.
+
+## Single host port layout
+In case you need to run multiple Thanos components on the same host you
```
bwplotka: Newline above.

docs/getting-started.md
```diff
@@ -243,3 +243,22 @@ TBD
...
+can follow this standardized port layout to avoid port collisions. 
+> The port can be changed using the `--http-address` and `--grpc-address` flags.
```
bwplotka: Why `>` ?

docs/getting-started.md
```diff
@@ -243,3 +243,22 @@ TBD
 Thanos also has a tutorial on deploying it to Kubernetes. We have a full page describing a standard deployment here.
 
 We also have example Grafana dashboards [here](/examples/grafana/monitoring.md) and some [alerts](/examples/alerts/alerts.md) to get you started.
+
+## Single host port layout
```
bwplotka: Can we change this to  maybe something like:

```
We don't recommend running Thanos on a single node on production. Thanos is designed and built to run as a distributed system. Vanilla Prometheus might be totally enough for small setups.

However, in case you want to play and run Thanos components on a single node, we recommend following the port layout: 

< TABLE> 

You can see example one-node setup [here](./scripts/quickstart.sh)
```
Plus, we might want to adjust script for those ports (:

bwplotka:
```suggestion
## Testing Thanos on Single Host
```
Maybe? (:

FUSAKLA: Not sure if the statement isn't bit too harsh because I can imagine using Thanos just as a global view solution when just sidecar and query is running on along with the Prometheus on the same host and having two of those hosts providing also HA for query but that is probably a corner case. 

I'll update it to your suggestions  and definitely also the script, I completely forgot about it thanks for reminding!

bwplotka: > Not sure if the statement isn't bit too harsh because I can imagine using Thanos just as a global view solution when just sidecar and query is running on along with the Prometheus on the same host and having two of those hosts providing also HA for query but that is probably a corner case.

Well for HA you probably want more than one Querier as well and not on the same host. Anyway feel free to adjust those suggestions to be less harsh, I definitely don't mean to be harsh (:

FUSAKLA: I meant two hosts and every one one of them running Prometheus, sidecar and query.
Harsh was a probably bit too harsh word :D but anyway I like the testing usage formulation so I'll stick with yours version.

FUSAKLA:
@bwplotka so the docs is changed PTAL and I updated the quickstart.sh

btw the `mc` default binary is really cool when it starts up the midnight commander :D

FUSAKLA:
the preview https://deploy-preview-1495--thanos-io.netlify.com/getting-started.md/#testing-thanos-on-single-host

lrosenman:
Thanks @FUSAKLA!  Looks good.

PR #1496, opened 2019-09-05
Allow setting TSDB block duration for receive service

fhalim:
Allow setting frequency at which TSDB blocks are created for the receive service.

## Changes

Extracted tsdb minBlockDuration and maxBlockDuration used in receive service into flag that defaults to the current value of 2h.

## Verification

Built and ran service locally to ensure that the flag takes effect as intended.

### DIFF
--- a/cmd/thanos/receive.go
+++ b/cmd/thanos/receive.go
@@ -143,6 +143,7 @@ func runReceive(
 		NoLockfile:        true,
 		MinBlockDuration:  tsdbBlockDuration,
 		MaxBlockDuration:  tsdbBlockDuration,
+		WALCompression:    true,
 	}
 
 	localStorage := &tsdb.ReadyStorage{}
--- a/cmd/thanos/receive.go
+++ b/cmd/thanos/receive.go
@@ -60,7 +60,7 @@ func registerReceive(m map[string]setupFunc, app *kingpin.Application, name stri
 
 	replicationFactor := cmd.Flag("receive.replication-factor", "How many times to replicate incoming write requests.").Default("1").Uint64()
 
-	tsdbBlockDuration := modelDuration(cmd.Flag("tsdb.blockduration", "Duration for local TSDB blocks").Default("2h"))
+	tsdbBlockDuration := modelDuration(cmd.Flag("tsdb.block-duration", "Duration for local TSDB blocks").Default("2h").Hidden())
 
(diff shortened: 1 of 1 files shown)


cmd/thanos/receive.go
```diff
@@ -60,6 +60,8 @@ func registerReceive(m map[string]setupFunc, app *kingpin.Application, name stri
 
 	replicationFactor := cmd.Flag("receive.replication-factor", "How many times to replicate incoming write requests.").Default("1").Uint64()
 
+	tsdbBlockDuration := modelDuration(cmd.Flag("tsdb.blockduration", "Duration for local TSDB blocks").Default("2h"))
```
bwplotka: Can we maybe make this hidden as it's not recommended to tweak this? Plus we keep more verbose naming (`-` between words)

```suggestion
	tsdbBlockDuration := modelDuration(cmd.Flag("tsdb.block-duration", "Duration for local TSDB blocks").Hidden().Default("2h"))
```

fhalim: The use case is that we generate an really high volume of metrics during a business day, so 2 hours of WAL add up to a pretty huge amount of disk space. That translates to expensive persistant volumes, backups etc. For the receive service, I'm planning to use it as a gateway to the object store, so it's desirable to reduce the duration enough that the average WAL backlog is small enough that we can restore it in a reasonable amount of time on failure.

I've marked the flag as hidden, and corrected the naming. Thanks!

krasi-georgiev: tsdb supports wal compression so maybe having it enabled would solve your use case without changing the block sizes?
In my tests this halves the size of the wal.

https://github.com/prometheus/prometheus/blob/937cc1a52af503d6be93119c6579228e43a60ec0/tsdb/db.go#L85-L86

brancz: I think Thanos receive should always unconditionally have compression enabled.

fhalim: WAL compression will definitely help reduce the disk size, reducing the amount of time restores will take. Updating the However, I'd still like to be able to control the maximum amount of time I depend on the block storage device for.

It seems that in Prometheus the max block duration defaults to 10% of the retention period which does not have a lower bound to it (https://github.com/prometheus/prometheus/blob/master/cmd/prometheus/main.go#L317). Is there a reason why that should not be allowed for the receive component?

PR #1525, opened 2019-09-16
Fixed partial delete issues on compactor; Added Upload/Delete tests.

bwplotka:
Fixes https://github.com/thanos-io/thanos/issues/1331

Problem that we are fixing is explained in the linked issue.

Also the problem will be fully fixed once this is also done: https://github.com/thanos-io/thanos/pull/1394 This is because although major parts of Thanos for syncing blocks is resilient to partial uploads, there are some points when we don't handle it properly (crash e.g compactor)
 
cc @lx223 @metalmatze 

### DIFF
--- a/pkg/block/block.go
+++ b/pkg/block/block.go
@@ -9,6 +9,9 @@ import (
 	"os"
 	"path"
 	"path/filepath"
+	"strings"
+
+	"github.com/go-kit/kit/log/level"
 
 	"github.com/thanos-io/thanos/pkg/block/metadata"
 
@@ -62,7 +65,7 @@ func Download(ctx context.Context, logger log.Logger, bucket objstore.Bucket, id
 func Upload(ctx context.Context, logger log.Logger, bkt objstore.Bucket, bdir string) error {
 	df, err := os.Stat(bdir)
 	if err != nil {
-		return errors.Wrap(err, "stat bdir")
+		return err
 	}
 	if !df.IsDir() {
 		return errors.Errorf("%s is not a directory", bdir)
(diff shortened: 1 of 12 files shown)


pkg/block/block.go
```diff
@@ -89,41 +92,66 @@ func Upload(ctx context.Context, logger log.Logger, bkt objstore.Bucket, bdir st
...
-func Delete(ctx context.Context, bucket objstore.Bucket, id ulid.ULID) error {
-	return objstore.DeleteDir(ctx, bucket, id.String())
+// Delete removes directory that is meant to be block directory.
+// NOTE: Always prefer this method for deleting blocks.
+//  * We have to delete block's files in the certain order (meta.json first)
+//  to ensure we don't end up with malformed partial blocks. Thanos system handles well partial blocks
+//  only if they don't have meta.json. If meta.json is present Thanos assumes valid block.
+//  * This avoids deleting empty dir (whole bucket) by mistake.
+func Delete(ctx context.Context, logger log.Logger, bkt objstore.Bucket, id ulid.ULID) error {
+	if err := bkt.Delete(ctx, path.Join(id.String(), MetaFilename)); err != nil {
```
metalmatze: Maybe it makes sense to put `path.Join(id.String(), MetaFilename)` into a variable, to avoid mismatches in the future with the second time below that you use this path.

bwplotka: :+1: 
Fixed

SuperQ:
Is there any race between the sidecar uploading (no meta.json yet) and the compactor deleting?

bwplotka:
> Is there any race between the sidecar uploading (no meta.json yet) and the compactor deleting?

Yea, detecting if the block is partially uploaded (malformed) vs upload is in progress is in current model quite tricky. We do that currently by just waiting for some time called `consistencyDelay` (configurable on compactor:https://github.com/thanos-io/thanos/blob/c83332f6aac03df3c9eeb5193c3d688da0f980f1/cmd/thanos/compact.go#L86). For now the default is 30m.

This has proven to be ok for all users (as per: no known issue because of that). Once upload takes more than 30 minutes, you might want to increase such consistency delay. This delay is also useful for object storages that are eventually consistent.

Good question is: can we warn/document these details about uploads longer than 30 minutes ):

SuperQ:
What about a simple locking method of renaming `meta.json` to `meta.json.deleted`, then cleanup the index/chunks, then finally delete the `meta.json.deleted` file.

bwplotka:
I merged already, but happy to discuss alternative ideas @SuperQ

> What about a simple locking method of renaming meta.json to meta.json.deleted, then cleanup the index/chunks, then finally delete the meta.json.deleted file.

Do you mean for deleting process? Not sure how is that better (: We want to get rid meta.json ASAP. Once we remove meta.json, block is nicely handled as malformed even if deletion would be disrupted at any point (e.g by pod crash/restart). In fact we would be fine by just deleting `meta.json` and allow compactor to assume as malformed block in next iteration.

bwplotka:
We also discussed some alternatives here: https://thanos.io/proposals/201901-read-write-operations-bucket.md/

bwplotka:
There is still some slight chance to restart compactor in a certain moment that will cause overlaps (in the worst case). For that to work we need to implement action items for design doc: https://thanos.io/proposals/201901-read-write-operations-bucket.md/

PR #1534, opened 2019-09-17
query: Add readiness probe to query

kakkoyun:
This PR,

* Adds `/-/ready` endpoint for readiness checks.

## Changes

* Adds `/-/ready` endpoint for readiness checks.
* Uses `prober.Prober` for readiness and liveness endpoints.
* Changes ~`defaultHTTPListener`~ `scheduleHTTPServer` to consume an existing `http.Handler`

## Verification

1. `make test`

2. Started `thanos query` and made a request to related endpoints.

```console
➜ curl http://0.0.0.0:10902/-/healthy
thanos query is healthy
```

```console
➜ curl http://0.0.0.0:10902/-/ready
thanos query is not ready. Reason: thanos query is initializing
```

```console
➜ curl http://0.0.0.0:10902/-/ready
thanos query is ready
```

### DIFF
--- a/cmd/thanos/main.go
+++ b/cmd/thanos/main.go
@@ -31,6 +31,7 @@ import (
 	"github.com/prometheus/client_golang/prometheus"
 	"github.com/prometheus/client_golang/prometheus/promhttp"
 	"github.com/prometheus/common/version"
+	"github.com/thanos-io/thanos/pkg/component"
 	"github.com/thanos-io/thanos/pkg/prober"
 	"github.com/thanos-io/thanos/pkg/runutil"
 	"github.com/thanos-io/thanos/pkg/store/storepb"
@@ -74,7 +75,7 @@ func main() {
 	cmds := map[string]setupFunc{}
 	registerSidecar(cmds, app)
 	registerStore(cmds, app, "store")
-	registerQuery(cmds, app, "query")
+	registerQuery(cmds, app)
 	registerRule(cmds, app, "rule")
 	registerCompact(cmds, app)
 	registerBucket(cmds, app, "bucket")
@@ -355,11 +356,14 @@ func metricHTTPListenGroup(g *run.Group, logger log.Logger, reg *prometheus.Regi
 
 // defaultHTTPListener starts a run.Group that servers HTTP endpoint with default endpoints providing Prometheus metrics,
 // profiling and liveness/readiness probes.
-func defaultHTTPListener(g *run.Group, logger log.Logger, reg *prometheus.Registry, httpBindAddr string, readinessProber *prober.Prober) error {
+func defaultHTTPListener(comp component.Component, g *run.Group, logger log.Logger, reg *prometheus.Registry, router http.Handler, httpBindAddr string, readinessProber *prober.Prober) error {
 	mux := http.NewServeMux()
 	registerMetrics(mux, reg)
 	registerProfile(mux)
 	readinessProber.RegisterInMux(mux)
+	if router != nil {
+		mux.Handle("/", router)
+	}
 
 	l, err := net.Listen("tcp", httpBindAddr)
 	if err != nil {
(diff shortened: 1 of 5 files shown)


kakkoyun:
cc @FUSAKLA @metalmatze @bwplotka

cmd/thanos/main.go
```diff
@@ -74,7 +75,7 @@ func main() {
 	cmds := map[string]setupFunc{}
 	registerSidecar(cmds, app)
 	registerStore(cmds, app, "store")
-	registerQuery(cmds, app, "query")
+	registerQuery(cmds, app)
```
bwplotka: I think this makes sense but inconsistent to other commands (:

cmd/thanos/main.go
```diff
@@ -355,24 +356,27 @@ func metricHTTPListenGroup(g *run.Group, logger log.Logger, reg *prometheus.Regi
...
+	}
 
 	l, err := net.Listen("tcp", httpBindAddr)
 	if err != nil {
 		return errors.Wrap(err, "listen metrics address")
 	}
 
 	g.Add(func() error {
-		level.Info(logger).Log("msg", "listening for metrics", "address", httpBindAddr)
+		level.Info(logger).Log("msg", fmt.Sprintf("listening for %s and metrics", comp.String()), "address", httpBindAddr)
```
bwplotka: I would not change this way. We aim for structured logging. While it might be bit confusing for human we want to spread variables as separate args like `address` `component` `msg`... what do you think? Then we don't need `fmt.Sprintf(`

cmd/thanos/main.go
```diff
@@ -355,24 +356,27 @@ func metricHTTPListenGroup(g *run.Group, logger log.Logger, reg *prometheus.Regi
...
 		readinessProber.SetHealthy()
-		return errors.Wrap(http.Serve(l, mux), "serve metrics")
+		return errors.Wrap(http.Serve(l, mux), fmt.Sprintf("serve %s and metrics", comp.String()))
```
bwplotka: Here we are fine with human message but instead of `fmt.Sprintf` you can use `errors.Wrapf`

cmd/thanos/main.go
```diff
@@ -355,24 +356,27 @@ func metricHTTPListenGroup(g *run.Group, logger log.Logger, reg *prometheus.Regi
...
 	}, func(err error) {
 		readinessProber.SetNotHealthy(err)
-		runutil.CloseWithLogOnErr(logger, l, "metric listener")
+		runutil.CloseWithLogOnErr(logger, l, fmt.Sprintf("%s and metric listener", comp.String()))
```
bwplotka: formatting is built-in here so you can remove fmt.Sprintf as well

cmd/thanos/main.go
```diff
@@ -355,24 +356,27 @@ func metricHTTPListenGroup(g *run.Group, logger log.Logger, reg *prometheus.Regi
 
 // defaultHTTPListener starts a run.Group that servers HTTP endpoint with default endpoints providing Prometheus metrics,
 // profiling and liveness/readiness probes.
-func defaultHTTPListener(g *run.Group, logger log.Logger, reg *prometheus.Registry, httpBindAddr string, readinessProber *prober.Prober) error {
+func defaultHTTPListener(comp component.Component, g *run.Group, logger log.Logger, reg *prometheus.Registry, router http.Handler, httpBindAddr string, readinessProber *prober.Prober) error {
```
bwplotka: This method is getting bit out of control. It's quite shallow. Maybe rename to `scheduleHTTPServer` instead?

cmd/thanos/query.go
```diff
@@ -408,29 +414,10 @@ func runQuery(
 
 		api.Register(router.WithPrefix(path.Join(webRoutePrefix, "/api/v1")), tracer, logger, ins)
 
-		router.Get("/-/healthy", func(w http.ResponseWriter, r *http.Request) {
```
bwplotka: Nice, thanks for reuse :+1:

cmd/thanos/main.go
GiedriusS: Isn't it just because we are in the middle of moving to the readiness probers in each component? Then they will all look the same :stuck_out_tongue:

cmd/thanos/main.go
```diff
@@ -355,24 +356,27 @@ func metricHTTPListenGroup(g *run.Group, logger log.Logger, reg *prometheus.Regi
 
 // defaultHTTPListener starts a run.Group that servers HTTP endpoint with default endpoints providing Prometheus metrics,
```
GiedriusS: Plus a small typo here so if you will rename it then you could fix it as well :smile:

cmd/thanos/main.go
kakkoyun: :D Exactly, in the end, they'll all look same.

kakkoyun:
@GiedriusS @bwplotka This is ready for another round. If this looks good I'll send similar PR's for other components today.

cmd/thanos/main.go
bwplotka: We are! let's go with this :+1:

PR #1537, opened 2019-09-18
receive: Add liveness and readiness probe

kakkoyun:
This PR,

* Adds `/-/healthy` endpoint for liveness checks.
* Adds `/-/ready` endpoint for readiness checks.

## Changes

* Adds `/-/healthy` endpoint for liveness checks.
* Adds `/-/ready` endpoint for readiness checks.
* Uses `prober.Prober` for readiness and liveness endpoints.

## Verification

1. `make test`

2. Started `thanos receive` and made a request to related endpoints.

```console
➜ curl http://0.0.0.0:10902/-/healthy
thanos receive is healthy%
```

```console
➜ curl http://0.0.0.0:10902/-/ready
thanos receive is not ready. Reason: thanos receive is initializing
```

```console
➜ curl http://0.0.0.0:10902/-/ready
thanos receive is ready%
```

### DIFF
--- a/cmd/thanos/receive.go
+++ b/cmd/thanos/receive.go
@@ -252,11 +252,11 @@ func runReceive(
 		)
 	}
 
-	level.Debug(logger).Log("msg", "setting up default http server")
+	level.Debug(logger).Log("msg", "setting up http server")
 	statusProber := prober.NewProber(comp, logger, prometheus.WrapRegistererWithPrefix("thanos_", reg))
-	// Initiate default HTTP listener providing metrics endpoint and readiness/liveness probes.
+	// Initiate HTTP listener providing metrics endpoint and readiness/liveness probes.
 	if err := scheduleHTTPServer(g, logger, reg, statusProber, httpBindAddr, nil, comp); err != nil {
-		return errors.Wrap(err, "schedule default HTTP server with probes")
+		return errors.Wrap(err, "schedule HTTP server with probes")
 	}
 
 	level.Debug(logger).Log("msg", "setting up grpc server")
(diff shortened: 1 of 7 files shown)


kakkoyun:
cc @FUSAKLA

cmd/thanos/receive.go
```diff
@@ -278,6 +284,7 @@ func runReceive(
 			s := newStoreGRPCServer(logger, reg, tracer, tsdbStore, opts)
 
 			level.Info(logger).Log("msg", "listening for StoreAPI gRPC", "address", grpcBindAddr)
+			statusProber.SetReady()
```
FUSAKLA: I think the receiver should probably not be ready until the TSDB is ready? Also not sure about the hashring and the receive interface is also not guarantied to be up at this point. 

Maybe this will require some more complex condition for the ready state :thinking:

kakkoyun: @FUSAKLA For the TSDB, it's ready at this stage if you check line `270`. It runs after `TSDB` is open.
For receive interface, I thought if something goes south it'll change liveness state so, readiness won't be needed.

I guess I need to double-check the hashring readiness.

I'll have another look at it.

kakkoyun:
@FUSAKLA @bwplotka I've updated the logic to set receive ready. Please have another look at it.

cmd/thanos/receive.go
```diff
@@ -277,6 +290,8 @@ func runReceive(
 			}
 			s := newStoreGRPCServer(logger, reg, tracer, tsdbStore, opts)
 
+			// Wait hashring to be ready before start serving metrics
+			<-hashringReady
```
squat: Why are we waiting for the hashring to be ready before serving metrics from the store? These things are entirely independent IMO

Issue #844, opened 2019-02-14
compactor: File name is too long error


mdferoz:
**Thanos, Prometheus and Golang version used**
-> prometheus instances run in kubernetes with thanos 0.3 side car and image tags used 2.7.1.

-> % thanos --version
type: S3
thanos, version 0.3.0 (branch: HEAD, revision: 837e9671737698bf1778a4a9abfebbf96117a0be)
  build user:       root@986454de7a63
  build date:       20190208-15:24:36
  go version:       go1.11.5

**What happened**
Running compact command fails against a known bucket containing our metrics data. The data itself is currently not huge, may be around 3GB so far. Thanos sidecar keep pushing the data in to our local s3 compatible storage and we were hoping to keep this data at lesser granularity for several years. Here is the command I used:

> % thanos compact --log.level=info --data-dir=/tmp/ --objstore.config-file=thanos-s3-bucket.conf --retention.resolution-raw=0d --retention.resolution-5m=180d --retention.resolution-1h=1000d --wait

**What you expected to happen**
I was hoping the compact command will process the data :)
**How to reproduce it (as minimally and precisely as possible)**:
Our data is collected by prometheus from a bunch of netdata agents but with limited metrics we have identified. This gets scraped every 15s at the moment and all targets are CentOS 6/7. I feel like that may not be as relevant, as this looks like something else while creating a directory locally.

**Full logs to relevant components**
level=info ts=2019-02-14T19:39:03.453025Z caller=factory.go:39 msg="loading bucket configuration"
level=info ts=2019-02-14T19:39:03.453635Z caller=compact.go:196 msg="retention policy of 5 min aggregated samples is enabled" duration=4320h0m0s
level=info ts=2019-02-14T19:39:03.453659Z caller=compact.go:199 msg="retention policy of 1 hour aggregated samples is enabled" duration=24000h0m0s
level=info ts=2019-02-14T19:39:03.453759Z caller=compact.go:281 msg="starting compact node"
level=info ts=2019-02-14T19:39:03.453862Z caller=compact.go:821 msg="start sync of metas"
level=info ts=2019-02-14T19:39:03.453839Z caller=main.go:308 msg="Listening for metrics" address=0.0.0.0:10902
level=info ts=2019-02-14T19:39:03.658623Z caller=compact.go:827 msg="start of GC"
[...]

PR #1544, opened 2019-09-19
Compactor: Fix for #844 - Ignore object if it is the current directory

jimbobby5:
It seems the s3 client has the (slightly unexpected) behaviour that the `ListObjects` call also returns the directory itself, which causes the recursion experienced in https://github.com/thanos-io/thanos/issues/844.

## Changes

Update bucket folder iteration logic to skip folder if it is the current folder.

## Verification

Deployed this version into test environment which was exhibiting the issue in https://github.com/thanos-io/thanos/issues/844


GiedriusS:
@jimbobby5 may I know what kind of S3 API implementation are you using? :thinking:

jimbobby5:
I believe this is using the minio s3 client.

GiedriusS:
@jimbobby5 I am talking about the server-side part because we don't have any issue like this here. Just curious. Overall the code LGTM.

jimbobby5:
Oh sorry I misunderstood the question. Have seen this on AWS as well as Quobyte.

pkg/objstore/s3/s3.go
```diff
@@ -239,6 +239,10 @@ func (b *Bucket) Iter(ctx context.Context, dir string, f func(string) error) err
 		if object.Key == "" {
 			continue
 		}
+		// The s3 client can also return the directory itself in the ListObjects call above
```
bwplotka: missing trailing period.

PR #1546, opened 2019-09-20
query: Experimental federate endpoint.

bwplotka:
This adds `/federate` endpoint directly to Querier.

**NOTE**: This PR is not intended to be merged for now. It's just exploring capabilities of integrating Thanos with older Prometheus APIs like federate.

:warning:  **Federate endpoint is not recommended as it tries to replicate a metric database (TSDB) with the requirement of the highly available network (double scrape).**

cc @brancz  @metalmatze @aditya-konarde

There are available docker images with those commits:

* `quay.io/thanos/thanos:v0.7.0-with-federate-b7177f666c`

### Why not separate service on top of StoreAPI?

Querier StoreAPI implementation does not do deduplication which might be handy, so, for now, we implemented that in querier itself. If separate service is needed, separate querier might be deployed.

### DIFF
--- a/cmd/thanos/query.go
+++ b/cmd/thanos/query.go
@@ -12,6 +12,8 @@ import (
 	"path"
 	"time"
 
+	"github.com/thanos-io/thanos/pkg/query/adapter"
+
 	"github.com/go-kit/kit/log"
 	"github.com/go-kit/kit/log/level"
 	grpc_middleware "github.com/grpc-ecosystem/go-grpc-middleware"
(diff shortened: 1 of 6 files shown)


midnightconman:
I was just looking for this yesterday... I have a use case where an external integration was designed as a prometheus scraper, so having a way to expose certain series by name in the common exposition format works nicely.

midnightconman:
What's the verdict on this feature?

brancz:
I believe we concluded this is not a good idea no @bwplotka? I think we can close this.

metalmatze:
Agreed. Please open again, if anything else was said.

PR #1548, opened 2019-09-20
Compact resolution/retention docs update. 

homelessnessbo:
## Changes

A little update of compactor documentation related to Resolution and retention with a hope that fewer people are going to get confused in the future. 

Related to #813


docs/components/compact.md
```diff
@@ -28,6 +28,18 @@ config:
 The compactor needs local disk space to store intermediate data for its processing. Generally, about 100GB are recommended for it to keep working as the compacted time ranges grow over time.
 On-disk data is safe to delete between restarts and should be the first attempt to get crash-looping compactors unstuck.
 
+# Resolution and retention
+
+Resolution - distance between data points on your graphs. E.g.
+
+raw - the same as scrape interval at the moment of data ingestion
```
bwplotka: Can you make those a bullet points? (with `*`)?

docs/components/compact.md
```diff
@@ -28,6 +28,18 @@ config:
...
+5m - data point is every 5 minutes
+1h - data point is every 1h
+
+Keep in mind, that the initial goal of downsampling is not saving a disk space. The goal of downsampling is providing an opportunity to get fast approximate results for range queries of big time intervals like months or years. In other words, if you set `--retention.resolution-raw` less then `--retention.resolution-5m` and `--retention.resolution-1h` - you might run into a problem of not being able to "zoom in" after you queried your data for 1 year interval.
```
bwplotka: Nice! Well put, some nits though:

* I would remove `approximate`. We made sure the data are mathematically correct (: So it should be correct.
* I removed ` after you queried your data for 1 year interval.` as I found this bit confusing -> you can zoom in even before I query long range query (: 

```suggestion
Keep in mind, that the initial goal of downsampling is not saving disk space. The goal of downsampling is providing an opportunity to get fast results for range queries of big time intervals like months or years. In other words, if you set `--retention.resolution-raw` less then `--retention.resolution-5m` and `--retention.resolution-1h` - you might run into a problem of not being able to "zoom in" to your historical data.
```

I would also add something about Querier being able to automatically deduce what resolution you need based on query `step` (for now). 

What do you think?

docs/components/compact.md
```diff
@@ -28,6 +28,18 @@ config:
...
+
+To avoid confusion - you might want to think about `raw` data as about "zoom in" opportunity. Considering the values for mentioned options - always think "Will I need to zoom in to the day 1 year ago?" if the answer "yes" - you most likely want to keep raw data for as long as 1h and 5m resolution, otherwise you'll be able to see only approximate representation of how your data looked like.
```
bwplotka: Nice!

In the same time can ask yourself (do I want to query long time ranges? If not I can disable downsampling by `debug.disable-downsampling` on compactor (or something like that ;p). Maybe worth to add?

docs/components/compact.md
bwplotka: Also, can we add something about expected size:

```
In fact, downsampling adds 2 more blocks for each raw block which are only slightly smaller or relatively similar size to raw block. This is required by internal downsampling implementation which to be mathematically correct holds various aggregations. This means that downsampling can increase the size of your storage a bit (~3x), but it gives massive advantage on querying long ranges. 
```

docs/components/compact.md
```diff
@@ -28,6 +28,18 @@ config:
...
+
```
bwplotka: I would also add:

```
Ideally, you will have equal retention set (or no retention at all) to all resolutions which allow both zoom in capabilities as well as performant long ranges queries. Since object storages are usually quite cheap, storage size might not matter that much.
```

docs/components/compact.md
homelessnessbo: @bwplotka I'd also elaborate on the exact list of aggregation so users would know what might be effectively used as a pre-calculated dataset.

Can you point me to where I can find such a list?

docs/components/compact.md
homelessnessbo: done

docs/components/compact.md
homelessnessbo: Added!

homelessnessbo:
@bwplotka so I'd add exact list of aggregations that happen, or at least link to the code where a user might find it and after that - merge it.

docs/components/compact.md
```diff
@@ -28,6 +28,16 @@ config:
...
 
+# Resolution and retention
+
+Resolution - distance between data points on your graphs. E.g.
+
+raw - the same as scrape interval at the moment of data ingestion
+5m - data point is every 5 minutes
+1h - data point is every 1h
+
+Keep in mind, that the initial goal of downsampling is not saving a disk space. The goal of downsampling is providing an opportunity to get fast approximate results for range queries of big time intervals like months or years. In other words, if you set `--retention.resolution-raw` less then `--retention.resolution-5m` and `--retention.resolution-1h` - you might run into a problem of not being able to "zoom in" after you queried your data for 1 year interval. To avoid confusion - you might want to think about `raw` data as about "zoom in" opportunity. Considering the values for mentioned options - always think "Will I need to zoom in to the day 1 year ago?" if the answer "yes" - you most likely want to keep raw data for as long as 1h and 5m resolution, otherwise you'll be able to see only approximate representation of how your data looked like.
```
j4ng5y: "is not saving a disk space." should probably read "is not to save disk space"

homelessnessbo:
Jesus heck how do I sign-off that nasty commit?

homelessnessbo:
So I executed a bunch of random git commands without any understanding of what's going on and now I have some changed that I did not do and that is already in master but all the checks are green. Have no clue what happened, but could you guys @gkope @bwplotka  approve and merge?

homelessnessbo:
oh I figured it. Gosh, I'm bad with git.

homelessnessbo:
I don't think I have merge permissions, so you gotta click the button folks

PR #1550, opened 2019-09-20
compact: add heuristic disk space check

GiedriusS:
Add a heuristic check to see if there is enough space on the disk to
even perform it. This is needed so that we would avoid downloading
potentially hundreds of gigabytes of data just to see that there is not
enough space. The check is the following: we add up the space that all
of the blocks take in the compaction group and then add another time the
size of the smallest block. It should work well enough.

Uses the new ObjectSize() method. Added a `diskusage` which was adopted
from https://github.com/ricochet2200/go-disk-usage which was old and unmaintained.
I'm not aware of any other libraries so perhaps it is a good idea to include this into
our code-base. It is needed to check how much space is free in a given
directory.

Refactor the code a bit so that metadata.InjectThanos() is removed and
we call Write() everywhere consistently.


pkg/block/metadata/meta.go
```diff
@@ -51,8 +51,9 @@ type Meta struct {
 
 // Thanos holds block meta information specific to Thanos.
 type Thanos struct {
-	Labels     map[string]string `json:"labels"`
-	Downsample ThanosDownsample  `json:"downsample"`
+	Labels      map[string]string `json:"labels"`
+	Downsample  ThanosDownsample  `json:"downsample"`
+	SizeInBytes *uint64           `json:"size"`
```
bwplotka: I guess this **does not** include meta file?

bwplotka: Worth to comment.

GiedriusS: Indeed, will do.

wrossmann:
Yes please, very much.

I just had the first high-level compaction try to run recently in the kubernetes deployment I just set up. For 16 hours the pod repeatedly overran its disk allocation, was evicted, and restarted. Ran up a solid $50 in S3 bandwidth before we noticed.

My one ask would be that the skipped compaction runs be reflected in the exported metrics somehow so that it can be track/alerted on.

GiedriusS:
Will rebase and add the comment about that. Need to double-check if all code paths use this function - AFAICT the answer is no so those parts should be refactored as well.

GiedriusS:
I believe this solves a pain point for our users (me included). @bwplotka PTAL again when you can! It has also turned into a nice refactoring since not all code paths were using `metadata.InjectThanos()` (where it makes sense, of course).

pkg/block/metadata/meta.go
```diff
@@ -53,6 +53,8 @@ type Meta struct {
 type Thanos struct {
 	Labels     map[string]string `json:"labels"`
 	Downsample ThanosDownsample  `json:"downsample"`
+	// Size of the block in bytes of the original block (before compaction and/or 1st upload).
```
bwplotka: This semantic is quite odd. What do you mean before compaction? Meta JSON is always bound to the block and block is immutable, so I think this detail is not necessary. Meta.json should state size in bytes related to the block is states metadata for (: 

Also curious, can we make it separate `indexSizeBytes`  and `chunkTotalSizeBytes`? Those are the biggest blocks that matter. This info will enormously help in knowing the file sizes in advance (: for other certain bucket operations.

I think I like it. Alternative is change bucket API to fetch some metadata (size), but this sounds like some easy way to achieve this.

pkg/block/metadata/meta.go
```diff
@@ -62,25 +64,41 @@ type ThanosDownsample struct {
...
+		return err
+	})
+	return size, err
+}
+
 // InjectThanos sets Thanos meta to the block meta JSON and saves it to the disk.
 // NOTE: It should be used after writing any block by any Thanos component, otherwise we will miss crucial metadata.
-func InjectThanos(logger log.Logger, bdir string, meta Thanos, downsampledMeta *tsdb.BlockMeta) (*Meta, error) {
-	newMeta, err := Read(bdir)
+func InjectThanos(logger log.Logger, bdir string, meta *Meta, downsampledMeta *tsdb.BlockMeta) (*Meta, error) {
```
bwplotka: I think we can simplify this function signature.

I think we would be fine with just `Write` that would add `index size` and `chunks total size` right?

Reasons:
* We never now use `downsampledMeta` with your change as we reuse `meta` (which is nice)
* It's actually the same as Write now, except size, and actually it's nice and expected behaviour.

pkg/block/metadata/meta.go
GiedriusS: Actually, let me invest a bit more time into seeing if we could add a new method to our remote object storage interface which would return the size of a specified object. Perhaps I will split it into a separate PR (if it's feasible to make it at all).

GiedriusS:
Actually, it shouldn't be hard to add a method like `ObjectSize()` to our objstore providers. Will prepare a PR soon. Let's postpone this until that.

pkg/compact/compact.go
```diff
@@ -743,6 +746,46 @@ func (cg *Group) areBlocksOverlapping(include *metadata.Meta, excludeDirs ...str
 	return nil
 }
 
+// isEnoughDiskSpace checks with a heuristic if there is enough disk space in the compaction group
+// in the specified directory. It adds up the space requirements of all the blocks in the compaction group
+// and then adds the smallest one again. The specified directory needs to have at least that amount of available bytes.
+// Doing this check prevents (in most cases) from downloading a lot of data and only then finding out that there is not
+// enough space.
+func (cg *Group) isEnoughDiskSpace(ctx context.Context, dir string) error {
```
squat: Do you think it might be more idiomatic to return `(bool, error)`? This way you can differentiate between a failure to calculate and a negative response? You could also do that with a specially typed err. Either way, doesn’t matter much for this use case

GiedriusS: I thought about this and arrived at exactly the same conclusion - it is unnecessary in this case unless we will want to add some kind of flag to skip this check. (:

pkg/block/metadata/meta.go
GiedriusS: Good catch! In fact, we had never passed the last argument to this function so I just went ahead and removed it, and renamed the existing users to use `Write()`.

pkg/compact/compact.go
```diff
@@ -743,6 +747,49 @@ func (cg *Group) areBlocksOverlapping(include *metadata.Meta, excludeDirs ...str
...
+	}
+
+	var sumTakenByBlocks uint64
+	var smallestBlockBytes uint64 = math.MaxUint64
+
+	for blid := range cg.blocks {
+		var thisBlockBytes uint64
+		err := cg.bkt.Iter(ctx, blid.String(), func(d string) error {
+			if strings.HasSuffix(d, objstore.DirDelim) {
+				return nil
```
bwplotka: I think this is wrong as compactor downloads all chunks as well, which in `chunks/` directory. This also reminds me there are no tests for this (: Do you mind adding some?

Also we might want really to be selective. E.g it might be quite confusing if someone copies `index` into same directory as `index.backup` and this will be summed here, however, compactor does not download it.  WDYT? (:

GiedriusS: I'm not sure how this part would skip the `chunks` directory - this is simply needed so that we would skip over things like `0123890ASDASDAS/chunks/` and only check `0123890ASDASDAS/chunks/00001`, etc.

Added some tests.

Also, I'm not sure that such a check that you are talking exists in the code at all besides the explicit handling of metadata files. AFAICT `block.Download()` is called that downloads everything that is inside there (: Could you please point me to where this check is done, @bwplotka? I feel like such a thing would be useful but probably should not be a part of this PR.

bwplotka: `Iter` is not recursive. This is what I meant (: 

> FAICT block.Download() is called that downloads everything that is inside there 

You are right, let's count all then :+1:

GiedriusS:
Any other concerns, comments?

bwplotka:
Looking

pkg/block/metadata/meta.go
```diff
@@ -65,27 +65,6 @@ type ThanosDownsample struct {
 	Resolution int64 `json:"resolution"`
 }
 
-// InjectThanos sets Thanos meta to the block meta JSON and saves it to the disk.
-// NOTE: It should be used after writing any block by any Thanos component, otherwise we will miss crucial metadata.
-func InjectThanos(logger log.Logger, bdir string, meta Thanos, downsampledMeta *tsdb.BlockMeta) (*Meta, error) {
```
bwplotka: What's wrong with this? This is really critical to not forget ):

pkg/compact/compact.go
```diff
@@ -583,6 +591,49 @@ func (cg *Group) areBlocksOverlapping(include *metadata.Meta, excludeDirs ...str
...
+	if err != nil {
+		return errors.Wrap(err, "get disk usage")
+	}
+
+	var sumTakenByBlocks uint64
+	var smallestBlockBytes uint64 = math.MaxUint64
+
+	for blid := range cg.blocks {
+		var thisBlockBytes uint64
+		err := cg.bkt.Iter(ctx, blid.String(), func(d string) error {
```
bwplotka: we can use `if err := ... ; err  != nil {` pattern.

bwplotka: What about `chunk` directory? It will be missed here.

pkg/compact/compact.go
```diff
@@ -583,6 +591,49 @@ func (cg *Group) areBlocksOverlapping(include *metadata.Meta, excludeDirs ...str
...
+func (cg *Group) isEnoughDiskSpace(ctx context.Context, dir string) error {
+	du, err := cg.getDiskUsage(dir)
+	if err != nil {
+		return errors.Wrap(err, "get disk usage")
+	}
+
+	var sumTakenByBlocks uint64
+	var smallestBlockBytes uint64 = math.MaxUint64
+
+	for blid := range cg.blocks {
```
bwplotka: This is wrong I think... `isEnoughSpace` should allow to pass plan for compaction as well as the base block for downsampling.

Why you iterate through whole group? Only those part of compaction will be downloaded.

pkg/compact/compact.go
```diff
@@ -656,6 +707,11 @@ func (cg *Group) compact(ctx context.Context, dir string, comp tsdb.Compactor) (
 		overlappingBlocks = true
 	}
 
+	// Heuristic check if there is enough disk space.
+	if err := cg.isEnoughDiskSpace(ctx, dir); err != nil {
```
bwplotka: This is wrong, we should download one block that's it.

pkg/compact/compact.go
```diff
@@ -773,12 +829,17 @@ func (cg *Group) compact(ctx context.Context, dir string, comp tsdb.Compactor) (
 	index := filepath.Join(bdir, block.IndexFilename)
 	indexCache := filepath.Join(bdir, block.IndexCacheFilename)
 
-	newMeta, err := metadata.InjectThanos(cg.logger, bdir, metadata.Thanos{
+	meta, err := metadata.Read(bdir)
+	if err != nil {
+		return false, ulid.ULID{}, errors.Wrapf(err, "read meta from %s", bdir)
+	}
+	meta.Thanos = metadata.Thanos{
```
bwplotka: I'd rather use Inject method here for consistency.

pkg/compact/compact_test.go
```diff
@@ -107,3 +118,85 @@ func TestGroupKey(t *testing.T) {
 		}
 	}
 }
+
+func Test_DiskUsage_Corner(t *testing.T) {
```
bwplotka: Corner :thinking: ?

pkg/diskusage/diskusage_posix.go
```diff
@@ -0,0 +1,28 @@
...
+	"syscall"
+
+	"github.com/pkg/errors"
+)
+
+// Get returns the Usage of a given path, or an error if usage data is
+// unavailable.
+func Get(path string) (Usage, error) {
+	var stat syscall.Statfs_t
+	err := syscall.Statfs(filepath.Clean(path), &stat)
```
bwplotka: `if err := (...); err != nil {` pattern available!

pkg/diskusage/diskusage_windows.go
```diff
@@ -0,0 +1,34 @@
+// Copyright (c) The Thanos Authors.
+// Licensed under the Apache License 2.0.
+
+// +build windows
```
bwplotka: thanks!

pkg/block/metadata/meta.go
GiedriusS: Yeah, my bad. I did a robotic job after rebasing on `master`. When I have first started this PR, `InjectThanos` was always called with `downsampledMeta == nil`. Now that you've added proper users, we should revert this change.

GiedriusS:
Will rework at some point in the future. Still want to avoid downloading hundreds of gigabytes when there is not enough disk space to do the compaction in the first place.

GiedriusS:
plz no mr. bot

bwplotka:
Still, would be nice to get back to this (:

GiedriusS:
:( nothing new to report on this. If anyone wants to pick it up and have any questions then feel free to pm me on Slack. I might finish this one day.

GiedriusS:
> :( nothing new to report on this. If anyone wants to pick it up and have any questions then feel free to pm me on Slack. I might finish this one day.

Same :(

bwplotka:
Let's keep it open, no worries @GiedriusS 🤗 

One way for moving this forward would be to add GH issue and close this PR and reference from issue so it's visible it's rdy to take!

someshkoli:
> > :( nothing new to report on this. If anyone wants to pick it up and have any questions then feel free to pm me on Slack. I might finish this one day.
> 
> Same :(

Can I pick this up and try to get it merged ? :smile:

GiedriusS:
> > > :( nothing new to report on this. If anyone wants to pick it up and have any questions then feel free to pm me on Slack. I might finish this one day.
> > 
> > 
> > Same :(
> 
> Can I pick this up and try to get it done ? 😄

Yep, feel free!

nicolastakashi:
Something that I can do to help with that @someshkoli ??

someshkoli:
> Something that I can do to help with that @someshkoli ??

@nicolastakashi, I havent actully done anything major yet on this, was working on another pr.
Will be starting now, If you wish to pick this up then go ahead :rocket:

Kot-o-pes:
This feature is essential for running large scale thanos deployments where bucket size can exceed 100tb, in this case compactor tries to download blocks, ends up with no space left on device instead of throttling downloads to match actual space left, we end up with a script which restart compactor on on space left error which is not the best way to run it

PR #1555, opened 2019-09-22
Adding doc explaining the importance of groups for compactor

Leulz:
## Changes

Added a `Groups` section in `compact.md` explaining the importance for the `external_labels` to be both unique and persistent across different Prometheis.

## Motivation

I have recently had to deal with quite a large data loss incident. 

I believe that it happened because we had a compactor running while the `external_labels` of the Prometheis that produced blocks for it had as `replica` a value that had a hash appended to it. 

That hash changed when a Prometheus version restarted for some reason, since it was a hash generated on startup. No more blocks from the group generated by the now-dead replica arrived, and so the old blocks never became big enough for them to be downsampled, and eventually they were deleted because of the retention policy we had.

So I think it's quite important to explain how groups are used by the compactor. I apologize if my explanation is incorrect in some way, I tried to read the code for the compactor, but I am not very familiar with Go.

### DIFF
--- a/docs/components/compact.md
+++ b/docs/components/compact.md
@@ -28,6 +28,17 @@ config:
 The compactor needs local disk space to store intermediate data for its processing. Generally, about 100GB are recommended for it to keep working as the compacted time ranges grow over time.
 On-disk data is safe to delete between restarts and should be the first attempt to get crash-looping compactors unstuck.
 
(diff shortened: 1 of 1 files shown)


Leulz:
@GiedriusS Done, thanks!

PR #1562, opened 2019-09-24
Fix downsampling option in querier URL

### LINKED ISSUE #1463: Downsampling option not contained in Thanos URL

<table class="table table-sm table-bordered table-striped table-hover">
      <tbody>
        <tr>
          <th scope="row">Version</th>
          <td>0.6.0</td>
        </tr>
        <tr>
          <th scope="row">Revision</th>
          <td>c70b80eb83e52f5013ed5ffab72464989c132883</td>
        </tr>
        <tr>
          <th scope="row">Branch</th>
          <td>HEAD</td>
        </tr>
        <tr>
          <th scope="row">BuildUser</th>
          <td>circleci@090532e54e70</td>
        </tr>
        <tr>
          <th scope="row">BuildDate</th>
          <td>20190718-11:16:14</td>
        </tr>
        <tr>
          <th scope="row">GoVersion</th>
          <td>go1.12.5</td>
        </tr>
      </tbody>
    </table>

**What happened**

When creating ad-hoc queries in Thanos UI, the URI contains all parameters required to reconstruct the queries. This is very helpful to send eg. a link to a co-worker to share an analysis, or to check from one Thanos environment to another by changing the URI.

Example partial URL:

    /graph?g0.range_input=1h&g0.expr=count(kube_pod_info)&g0.tab=0

Screenshot showing scaledown option

The URI misses the the auto scale down option, though, which has to be set up for all queries after opening the URI.

Downsampling option contained in the URI

Change downsampling option, observe URI

obiesmans:
Fixes #1463 

* Downsampling option in Querier UI is now reflected in the URL.

### DIFF
--- a/pkg/ui/static/js/graph.js
+++ b/pkg/ui/static/js/graph.js
@@ -531,10 +531,6 @@ Prometheus.Graph.prototype.submitQuery = function() {
   var rangeSeconds = self.parseDuration(self.rangeInput.val());
   var resolution = parseInt(self.queryForm.find("input[name=step_input]").val()) || Math.max(Math.floor(rangeSeconds / 250), 1);
   var maxSourceResolution = self.maxSourceResolutionInput.val()
-  // normalize max_source_resolution parameter
-  if ( maxSourceResolution == "auto") {
-	  maxSourceResolution = ""
-  }
   var endDate = self.getEndDate() / 1000;
   var moment = self.getMoment() / 1000;
 
--- a/pkg/ui/static/js/graph.js
+++ b/pkg/ui/static/js/graph.js
@@ -44,6 +44,9 @@ Prometheus.Graph.prototype.initialize = function() {
   if (self.options.tab === undefined) {
     self.options.tab = 1;
   }
+  if (self.options.max_source_resolution === undefined) {
+	  self.options.max_source_resolution = "0s";
+  }
 
   // Draw graph controls and container from Handlebars template.
 
@@ -93,6 +96,8 @@ Prometheus.Graph.prototype.initialize = function() {
   self.stacked = self.queryForm.find("input[name=stacked]");
   self.insertMetric = self.queryForm.find("select[name=insert_metric]");
   self.refreshInterval = self.queryForm.find("select[name=refresh]");
+  self.maxSourceResolutionInput = self.queryForm.find("select[name=max_source_resolution_input]");
+
 
   self.consoleTab = graphWrapper.find(".console");
   self.graphTab   = graphWrapper.find(".graph_container");
@@ -230,6 +235,8 @@ Prometheus.Graph.prototype.initialize = function() {
     self.partialResponse.val(self.isPartialResponseEnabled() ? '0' : '1');
     stylePartialResponseBtn();
   });
+
+  self.maxSourceResolutionInput.val(self.options.max_source_resolution);
 
   self.queryForm.submit(function() {
     self.consoleTab.addClass("reload");
@@ -377,7 +384,6 @@ Prometheus.Graph.prototype.getOptions = function() {
     "range_input",
     "end_input",
     "step_input",
-    "downsample_input",
     "stacked",
     "moment_input"
   ];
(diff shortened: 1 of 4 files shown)


obiesmans:
Amended the pr to use camelcase and have a max_source_resolution in URL, as in the documentation.

pkg/ui/static/js/graph_template.handlebar
```diff
@@ -115,7 +115,7 @@
                         <div class="prometheus_input_group pull-left">
                            <select name="max_source_resolution_input" title="Downsampling option" id="max_source_resolution_input{{id}}">
                               <option value="">Auto downsampling</option>
-                              <option selected="selected" value="0s">Only raw data</option>
+                              <option value="0s">Only Raw data</option>
```
GiedriusS: Why did you take off `selected` here and converted it into `Raw`? It is changing the defaults AFAICT :(

pkg/ui/static/js/graph.js
```diff
@@ -390,6 +393,9 @@ Prometheus.Graph.prototype.getOptions = function() {
       }
     }
   });
+
+  options.max_source_resolution = self.queryForm.find("select[name=max_source_resolution_input]").val();
```
GiedriusS: Maybe `select[name=max_source_resolution_input]` should be moved out somewhere because we are repeating ourselves?

GiedriusS:
Everything works, just those two comments and we should be good to go :smile:

pkg/ui/static/js/graph_template.handlebar
obiesmans: Converted Raw back, selected is still off, but now the code select "0s" if max_source_resolution is undefined. This keeps the default comportment to "Only raw data".

pkg/ui/static/js/graph.js
obiesmans: Refactored the 3 occurrences as `maxSourceResolutionInput`.

obiesmans:
Thanks a lot for the review ! Sorry I didn't reply sooner, was off for a few days.
I pushed an update that addresses @GiedriusS comments.

I also ran into issues where the URL params format didn't allow for empty params. 
I thought that correcting this would make maintaining against the upstream prometheus UI harder, so instead I put "auto" as keyword in the UI and normalized it back to an empty string when submitting the query. Is it OK with you ?

pkg/ui/static/js/graph.js
```diff
@@ -521,7 +530,11 @@ Prometheus.Graph.prototype.submitQuery = function() {
   var startTime = new Date().getTime();
   var rangeSeconds = self.parseDuration(self.rangeInput.val());
   var resolution = parseInt(self.queryForm.find("input[name=step_input]").val()) || Math.max(Math.floor(rangeSeconds / 250), 1);
-  var maxSourceResolution = self.queryForm.find("select[name=max_source_resolution_input]").val();
+  var maxSourceResolution = self.maxSourceResolutionInput.val()
+  // normalize max_source_resolution parameter
+  if ( maxSourceResolution == "auto") {
```
bwplotka: Why not instead supporting `auto` value in Thanos Querier server side? :thinking:

obiesmans: Yes I like the idea.
Should I add this to this PR or open another PR ?

obiesmans: Also remember that grafana users can't set this parameter and probably want the "auto" parameter as default.

pkg/ui/static/js/graph_template.handlebar
```diff
@@ -114,8 +114,8 @@
 
                         <div class="prometheus_input_group pull-left">
                            <select name="max_source_resolution_input" title="Downsampling option" id="max_source_resolution_input{{id}}">
-                              <option value="">Auto downsampling</option>
-                              <option selected="selected" value="0s">Only raw data</option>
+                              <option value="auto">Auto downsampling</option>
+                              <option value="0s">Only raw data</option>
```
bwplotka: Quick silly question, why we remove `selected` here again? It changes what is by default set, right?

GiedriusS: Makes sense IMHO that we keep all of that logic together in that file instead of the HTML code, thanks for pointing this out!

GiedriusS:
Also, double-checked the "instant query" side of this - everything should work well like before and now actually the user will be able to see that they are able to change this parameter -- it was "invisible" before.

obiesmans:
Thanks a lot for take the time to review this @GiedriusS !

Issue #1509, opened 2019-09-11
Thanos receive failing on receive "out of order sample"


R4scal:
thanos, version 0.7.0 (branch: HEAD, revision: c6eaf68bec32aefd280318eaef6121d6ddb20d10)

prometheus, version 2.12.0 (branch: HEAD, revision: 43acd0e2e93f9f70c49b2267efa0124f1e759e86)

**Setup**
prometheus (in kubernetes) + influxdb_exporter + thanos receive

**What happened**
1. In prometheus logs (1 record on 30-60 seconds with scrape every 15 seconds)
```
{"caller":"scrape.go:1164","component":"scrape manager","level":"warn","msg":"Error on ingesting out-of-order samples","num_dropped":11,"scrape_pool":"influxdb_exporters","target":"http://1.1.1.1:xxx/metrics","ts":"2019-09-11T07:58:16.910Z"}
```
2. In thanos receive logs (thouthands)
```
{"caller":"handler.go:307","component":"receive-handler","endpoint":"http://2.2.2.2:xxx/api/v1/receive","err":"failed to non-fast add: out of order sample","level":"error","msg":"storing locally","ts":"2019-09-11T07:59:37.697039374Z"}
```
3. After some time in prometheus logs
```
{"caller":"queue_manager.go:716","component":"remote","count":100,"err":"context canceled","level":"error","msg":"non-recoverable error","queue":"0:http://2.2.2.2:xxx/api/v1/receive","ts":"2019-09-11T08:10:04.482Z"}
```
In this scenario, the load on the processor and the network also increases significantly, since prometheus tries many times to send the same data

**What you expected to happen**
I think, need reproduce prometheus action, when receive "out of order sample" it's warning, not critical error, need drop this sample (with increment dropped metrics counter) and response 200 to write request

krasi-georgiev:
In general the problem here seems to be that the `influxdb_exporter` is misbehaving and adds incorrect timestamps.

@R4scal your suggestion would work, but a better solution would be to not even try to send these out of order samples in the first place. cc @cstyan @csmarchbanks 

@squat  tsdb has the error `tsdb.ErrOutOfOrderSample` so can also add this as a fail safe. Can match against that error,log a warning and just return 200 response. What do you think?

squat:
@krasi-georgiev Ack sounds good to me. I can make the PR quickly

csmarchbanks:
[...]

PR #1563, opened 2019-09-24
Fixes out of order samples handling

dtrejod:
Fixes: https://github.com/thanos-io/thanos/issues/1509

## Changes

* Return 409 (Conflict) when encountering out of order/duplicate samples
* Rollback TSDB changes when we hit a non-fast add failure

## Verification

Verified it resolved our out-of-order sample situation in our test env.


pkg/receive/writer.go
```diff
@@ -43,7 +44,16 @@ func (r *Writer) Receive(wreq *prompb.WriteRequest) error {
...
+			switch err {
+			case storage.ErrOutOfOrderSample:
+				level.Debug(r.logger).Log("msg", "Out of order sample", "lset", lset.String(), "sample", s.String())
+			case storage.ErrDuplicateSampleForTimestamp:
+				level.Debug(r.logger).Log("msg", "Duplicate sample for timestamp", "lset", lset.String(), "sample", s.String())
+			case storage.ErrOutOfBounds:
+				level.Debug(r.logger).Log("msg", "Out of bounds metric", "lset", lset.String(), "sample", s.String())
+			}
 			if err != nil {
+				app.Rollback()
```
krasi-georgiev: why do you need to rollback here?

Imagine the situation:
added sample 1
error on adding sample2

when you run a rollback you also discard sample 1

when you hit an error at app.Add it doesn't get added to the head anyway. I don't see why you need to run a rollback. Correct me if my assumption is wrong.

dtrejod: If my understanding of the mechanism is correct, then on any failed sample send we should be discarding all samples appended in that batch. 

You are correct -- currently sample1 is added, but sample2 isn't. After returning a [500 error code](https://github.com/prometheus/prometheus/blob/08c55c119f39093e18b2bb9cba5c5619dc4ea0e1/storage/remote/client.go#L112-L114), Prometheus considers the batch of samples sent recoverable, and retries all samples -- see [sendSamplesWithBackoff](https://github.com/prometheus/prometheus/blob/3344bb5c332838df21f87b6473c0764315f5ca71/storage/remote/queue_manager.go#L781). On the next retry there is now a conflict on sample1 since you already appended it.

krasi-georgiev: From my POV we only want to notify about this conflict and not cause Prometheus to try and resend these as sending the same samples will result in the same conflict. 
As mentioned in the original issue the remote write shouldn't even send these and it this should be fixed in Prometheus as well.

The idea with this change is to  ingest the samples and just ignore the conflict errors as these are harmless and notify Prometheus with `http.StatusConflict`. 
Now I haven't looked at the  Prometheus code for the remote write, but if it tries to resend the same samples even with the `http.StatusConflict` than a PR should be opened to change this behavior in Prometheus.

https://github.com/thanos-io/thanos/issues/1509#issuecomment-531261715
>What error code does Thanos return when failing due to an out of order error? Prometheus should only retry on 5XX errors, so returning a 4XX would probably be the correct behavior and avoid the retries.

@csmarchbanks's comment suggests that `4XX` are not retried so this change should work as expected.

I haven't looked at the tests, but I would definitely prefer to have some tests to ensure this behavior.

dtrejod: 👍 Yes, receiving a 4xx code results in Prometheus abandoning the batch and moving onto the next.

I updated my commit to reflect that we should add all samples possible even when encountering errors.

squat:
Was this tested with replication on or off? Looking at this quickly, it seems to me that if replication is enabled, then the out of order conflict will be swallowed by the replication error. We need to surface the reason that replication failed if we want to catch all of the out of order sample errors. In other words, we need to do deep inspection of error causes for nested wrapped multierrors (#1586).

pkg/receive/writer.go
```diff
@@ -41,17 +45,26 @@ func (r *Writer) Receive(wreq *prompb.WriteRequest) error {
...
+			case storage.ErrDuplicateSampleForTimestamp:
+				level.Debug(r.logger).Log("msg", "Duplicate sample for timestamp", "lset", lset.String(), "sample", s.String())
+			case storage.ErrOutOfBounds:
+				level.Debug(r.logger).Log("msg", "Out of bounds metric", "lset", lset.String(), "sample", s.String())
 			}
+			errs.Add(errors.Wrap(err, "failed to non-fast add"))
 		}
 	}
 
 	if err := app.Commit(); err != nil {
```
squat: if the errs slice is length > 0 then we should never even try to commit, no?

dtrejod:
My testing thus far was done with 3 receive nodes with no replication. I agree this gets more complicated when replication is turned on. In my head here's how I think error handling should happen:

**Scenario 1:**
No replication errors but some conflict samples, we return a 409. 

**Scenario 2:**
Replication error on 2 out of 3 nodes we return 500 -- regardless if we saw conflict samples. On the one node that was successful in processing the wreq, we add the samples to the db. Given the 500 return code, Promethteus will retry the same wreq. We can expect the one node that was successful the first time to encounter conflicts but the other two nodes to successfully process the wreq. Assuming no more replication errors however, a 409 is returned (conflicts from node one) and Prometheus skips to the next wreq. 

Does that sound reasonable? 

---

Also I think I figured out why we were encountering this issue. We had a Prometheus endpoint erroneously producing a metric endpoint that had a metric listed twice but with the value. For example:
```sh
$ curl -s https://<hostname>/metrics/prometheus |grep jvm_memory_heap_used
# HELP jvm_memory_heap_used 
# TYPE jvm_memory_heap_used gauge
jvm_memory_heap_used 5.33600144E8
# HELP jvm_memory_heap_used 
# TYPE jvm_memory_heap_used gauge
jvm_memory_heap_used 5.33600144E8
```

Apparently this is a valid case that is [silently ignored](https://github.com/prometheus/prometheus/blob/3344bb5c332838df21f87b6473c0764315f5ca71/tsdb/head.go#L1715-L1719)  in Prometheus. This type of error still results in the sample being [appended in the TSDB twice](https://github.com/prometheus/prometheus/blob/3344bb5c332838df21f87b6473c0764315f5ca71/tsdb/head.go#L873).

squat:
Even without considering replication, this PR needs to consider passing through 409s received from forwarding time series. Please take a look at case 2 from my comment above. Currently this is not addressed. See pr 1586 for a working implementation.

pkg/receive/writer.go
squat: ping @dtrejod

squat:
still one comment: https://github.com/thanos-io/thanos/pull/1563/files#r329580671

pkg/receive/writer.go
dtrejod: We don't want to rollback. Per discussion before -- https://github.com/thanos-io/thanos/pull/1563#discussion_r329066801 -- we should always commit all valid samples. We simply should ignore the samples in conflict and continue on processing.

dtrejod:
@squat / @krasi-georgiev Anything blocking this PR still?

squat:
lgtm

I think that there is a little bit of nuance in the replication error logic that we can improve in order to avoid false positives when discovering 409s. We can add that in a follow up. Namely, when the replicate func checks for errors, it should only return the 409 if the number of 409s in the multi-error was greater than or equal to the minimum replication threshold. Otherwise, the replication should be retried.

In order words, let's say we have replication factor 3. This means the minimum replication threshold is 2. 
* if there was only 1 error, then no matter what the error was, the replication will succeed and return a nil error;
* if there were two or more errors:
    * if just one of the errors was a 409, then we shouldn't return a 409 because this didn't really cause the failure, the 500 from the other error is responsible; the error should either be fixed in code, or prometheus should retry sending the sample;
    * if two or more errors were 409s, then the sample is definitely conflicting and should not be retried. There is no internal server error that could be fixed and so retrying will never work

GiedriusS:
> lgtm
> 
> I think that there is a little bit of nuance in the replication error logic that we can improve in order to avoid false positives when discovering 409s. We can add that in a follow up. Namely, when the replicate func checks for errors, it should only return the 409 if the number of 409s in the multi-error was greater than or equal to the minimum replication threshold. Otherwise, the replication should be retried.
> 
> In order words, let's say we have replication factor 3. This means the minimum replication threshold is 2.
> 
>     * if there was only 1 error, then no matter what the error was, the replication will succeed and return a nil error;
> 
>     * if there were two or more errors:
>       
>       * if just one of the errors was a 409, then we shouldn't return a 409 because this didn't really cause the failure, the 500 from the other error is responsible; the error should either be fixed in code, or prometheus should retry sending the sample;
>       * if two or more errors were 409s, then the sample is definitely conflicting and should not be retried. There is no internal server error that could be fixed and so retrying will never work

Could we create a new issue from this comment?

squat:
I would suggest we merge this as is (the current state of the PR is much better than without it). I can file the comment as an issue and I will happily make a follow up and also add a test.

Issue #1471, opened 2019-08-28
Reduce store memory usage by reworking block.indexCache


miekg:
See the umbrella bug on Store memory usage: #448

Lots of memory is being used to keep the indexCache in memory (this is in pkg/block/index.go). 
Instead of having a fixed in-ram cache of every index file that is loaded, we could maintain an LRU cache of partial index files on disk. Index files are designed to be mmaped so we can leverage the kernel page cache for ram caching of this data. This would replace the existing LRU index caches, and let us delegate more of the work to the prometheus tsdb libraries which already support this operating mode.

/cc @asuffield (to flesh out some more, as he has spend more cycles on this currently).

asuffield:
We get burned by having a relatively high number of label values over time, and the current index cache implementation pins all of these into ram, even though most are never used. We're up to nearly 40Gb of index.cache.json files now, and it grows over time because of continually changing label values. Prometheus itself doesn't mind because the number of unique values in any recent chunk of time is small enough to fit into ram, but thanos attempts to hold multiple copies of every value in history.

The ByteSlice interface in prometheus/tsdb would fit neatly onto a download-with-cache function, which would:
 - mmap a disk file for local backing store
 - if the index file is not already present, fetch part or all of it
 - return a byte slice into the mmaped file

This leans on the kernel page cache to do most of the ram juggling, and the thanos code can focus on managing disk usage. With that in place, I believe the index.cache.json files become unnecessary.

bwplotka:
Awesome! I like this generally as well as attempt to reuse as much as possible (ByteSlice from TSDB) :heart:

Just to clarify:

 Do you mean to directly hook into [ByteSlice](https://github.com/prometheus/tsdb/blob/0fdd93b0b936bcd9aefa67954821961aba16c7b3/index/index.go#L622) and same for chunks into form of: `ByteSlice -> Cache (mmap) -> Cache2 (Disk) -> ObjectStorage API. GetRange/Get`? 

This is great from the code complexity and readability standpoint, however, there are a couple of challenges, we might need to investigate first.

Before we touch those just one thing. It is super confusing and we should change the naming ASAP, but there are two index caches in Thanos. It looks like @miekg @asuffield are on the same page but just for further readers of this issue, there are
[...]

PR #1567, opened 2019-09-25
docs: add BucketFile proposal

zhulongcheng:
related to #1471


bwplotka:
Nice,  I am back from holidays, so will take a proper look today @zhulongcheng ! Thanks!

bwplotka:
cc @GiedriusS @ppanyukov

docs/proposals/201909_bucketfile.md
```diff
@@ -0,0 +1,329 @@
...
+
+Each time to fetch pending pages，BucketFile will reset  `pendingPages`, `pendingReaders` and `pendingChan`.
+
+### In-memory object
+
+Each IndexReader decodes [postings offset block](https://github.com/prometheus/prometheus/blob/master/tsdb/docs/format/index.md#postings-offset-table) and [label offset block](https://github.com/prometheus/prometheus/blob/master/tsdb/docs/format/index.md#label-offset-table) into memory object for sequential and  random accesses.  
+
+Each BucketFile has a `pages` bitmap, but the bitmap uses very small memory. 
+
+For inactive IndexReaders, if the inactive duration (`now.Sub(lastQueryTime)`) greats than a configured duration (e.g. 15 mins), then IndexReader will remove in-memory objects and close file descriptor and mmap in IndexFile.  This will reduce memory usage.
```
GiedriusS: Who will actually track this time?

docs/proposals/201909_bucketfile.md
```diff
@@ -0,0 +1,329 @@
...
+
+**NOTE**: The `pages` bitmap in BucketFile should always be kept in memory until the server is closed.
```
GiedriusS: Could you elaborate more on this? Why do we want this if files might disappear?

docs/proposals/201909_bucketfile.md
```diff
@@ -0,0 +1,329 @@
...
+Our goals here are:
+
+- A) Reduce memory usage
+- B) Reuse chunks fetched from object storage
+
+## Motivation
+
+Currently, the store gateway loads all [index.cache.json](https://github.com/thanos-io/thanos/blob/865d5ec710b5651ead08a7dc73864a6f5dbeeaa9/pkg/block/index.go#L40)  files into memory on startup, which easily hits OOM situations. On the other hand, some indexes are rarely or never queried, so we don't have to keep these indexes in memory.
+
+The store gateway fetches chunks from object storage for each query request. but fetched chunks are not reused for next requests.
```
GiedriusS: If we will `mmap` everything and create sparse files then won't this technically mean that there is kind of an assumption that we have "infinite" RAM and address space? How will `BucketFile` ensure that we won't pass a certain threshold? What will we do in situations under memory pressure? I think that ATM having a common pool of RAM for responses serves us well as a capacity planning mechanism but this could potentially mean that someone will just ask for random data from a bucket that could be terabytes in size, and we will hold all of that in memory?

docs/proposals/201909_bucketfile.md
```diff
@@ -0,0 +1,329 @@
...
+
+
+
+![BucketBlock](./img/bucketfile_in_block.jpg)
+
+The above sequence diagram shows how the BucketFile works with IndexReader and ChunksReader in a block:
+
+(BucketIndexFile and BucketChunksFile are represented by `BucketFile` )
+
+1. BucketStore calls `blockSeries`  method to retrieve series from a block.
```
GiedriusS: It might be me but I feel like all of this description might be a bit too verbose. We should focus on how `BucketFile` works

docs/proposals/201909_bucketfile.md
zhulongcheng: `IndexReaders` and `ChunksReader`

docs/proposals/201909_bucketfile.md
zhulongcheng: Inactive IndexReaders just close file descriptors, not remove local files. 

Inactive IndexReaders may be queried again.
Keep `pages` bitmap in memory, so we can reuse fetched content.

docs/proposals/201909_bucketfile.md
zhulongcheng: this helps to understand how `BucketFile` works with other components.

will remove if no need.

docs/proposals/201909_bucketfile.md
zhulongcheng: >  ask for random data from a bucket that could be terabytes in size, and we will hold all of that in memory?    

will not hold all of that in memory.  
`BucketFile` will read small size (e.g. 1 MiB) data from `io.Reader` and write to a local file each time, until  `io.Reader` EOF. 

> there is kind of an assumption that we have "infinite" RAM and address space?

If we fetch many blocks (> 1000 blocks) from object storage,
it is possible that the size of all mmaps exceeds the virtual address space (e.g. 128TiB) for the store gateway process.
I think [Thanos Sharding](https://github.com/thanos-io/thanos/blob/master/docs/proposals/201909_thanos_sharding.md) is a good way to resolve this.

GiedriusS: Yeah but typically resilient systems gracefully shed their load, no? I would rather have Thanos Store unmmap some files instead of crashing in the middle of the night when someone queries a bunch of disparate data. I am fairly confident that we should probably have some kind of strategy on how to deal with this kind of situation instead of crashing and telling our users that maybe they should split up the Thanos Store nodes.

docs/proposals/201909_bucketfile.md
bwplotka: Essentially what is here is just coding it straight away but in human language. That's hard to review (:  It's similar if we would just start coding without design. We might need to agree on high level things first.

docs/proposals/201909_bucketfile.md
```diff
@@ -0,0 +1,329 @@
...
+    pendingChan     chan error      // chan to notify callers the result of prefetch	
+}
+
+func (f *BucketFile) Prefetch(ranges []*Range) (chan error, error) {...}
+```
+
+BucketFile appends new pages to `pendingPages`, waits for next  time to fetch. So that we can combine multiple queries to ones.
+
+When fetching combined  pages,
+if an error occurs, BucketFile will send the error to callers via the `pendingChan` channel,
```
bwplotka: Error handling is really a low-level implementation detail and should be not part of the design as it does not add much to the topic. It's enough to say - there must be error handling for failed prefetches. And if single fetch is retried or crash whole request? Those questions should be covered in design, not what go channels to use, I think (:

docs/proposals/201909_bucketfile.md
bwplotka: `greats`?

bwplotka: there will be separate tracking for those two?

docs/proposals/201909_bucketfile.md
bwplotka: What we miss from design is human description of what `pages bitmat` is, how large it is, what the use case. Cant's see any description now. ):

docs/proposals/201909_bucketfile.md
```diff
@@ -0,0 +1,329 @@
...
+
+Each BucketFile has a `pages` bitmap, but the bitmap uses very small memory. 
+
+For inactive IndexReaders, if the inactive duration (`now.Sub(lastQueryTime)`) greats than a configured duration (e.g. 15 mins), then IndexReader will remove in-memory objects and close file descriptor and mmap in IndexFile.  This will reduce memory usage.
+
+**NOTE**: The `pages` bitmap in BucketFile should always be kept in memory until the server is closed.
+
+### Preload blocks on startup
+
+The latest blocks will be queried mostly.
```
bwplotka:
```suggestion
The latest blocks will be queried the most. 
```

docs/proposals/201909_bucketfile.md
```diff
@@ -0,0 +1,329 @@
...
+
+For inactive IndexReaders, if the inactive duration (`now.Sub(lastQueryTime)`) greats than a configured duration (e.g. 15 mins), then IndexReader will remove in-memory objects and close file descriptor and mmap in IndexFile.  This will reduce memory usage.
+
+**NOTE**: The `pages` bitmap in BucketFile should always be kept in memory until the server is closed.
+
+### Preload blocks on startup
+
+The latest blocks will be queried mostly. 
+
+Preload `the last N hours` (configurable) blocks on startup.
```
bwplotka: Interesting idea.

docs/proposals/201909_bucketfile.md
```diff
@@ -0,0 +1,329 @@
...
+This will fetch index files from object storage and decode bytes into in-memory objects.
+
+### Local Files
+
+Because the max size of a index file is 64 GiB, and  the max size of a chunks file is 128 MiB.
```
bwplotka: Really? I think chunks files are capped on 0.5GB

docs/proposals/201909_bucketfile.md
```diff
@@ -0,0 +1,329 @@
...
+3. Fetch everything starting at `start` till  `end`:    
+   `Range{start:N, end: M}` -> Request header `Range: bytes=M-N` 
+
+if providers support multiple ranges, likes `Range: bytes=500-700,60000-61000`, we need to implement a `BucketReader.GetRanges` API, then we can query multiple ranges in one request.
+
+See https://tools.ietf.org/html/rfc7233#section-2.1 for reference.
+
+## Risks
+
+Ensure all providers support `fetch last N bytes` and `fetch everything starting from offset`. Otherwise, we need to know the size of remote index file before fetching TOC block. We may get the size by fetching metadata from object storage.
```
bwplotka: Nice! Can we actually check .. now? Design phase is exactly perfect moment for this.

docs/proposals/201909_bucketfile.md
```diff
@@ -0,0 +1,329 @@
...
+  - implement `BucketReader.GetRanges` if providers support
+- B) Implement `BucketFile`
+  
+- C) Use `BucketFile` for `BucketIndexReader` and `BucketchunkReader`
+
+- D) Add Flags to enable `BucketFile` 
+
+### Future work
+
+Design a new index file for [postings offset entries](https://github.com/prometheus/prometheus/blob/master/tsdb/docs/format/index.md#postings-offset-table) and [label offset entries](https://github.com/prometheus/prometheus/blob/master/tsdb/docs/format/index.md#label-offset-table) for sequential and  random accesses.
```
bwplotka: Future work means not `next steps to implement this designed idea`  - it's rather: What's next if we implement this design?

docs/proposals/201909_bucketfile.md
```diff
@@ -0,0 +1,329 @@
...
+* https://github.com/thanos-io/thanos/issues/448 (store gateway OOM)
+
+## Summary
+
+This document describes the motivation and design of the `BucketFile` of the store component. 
+
+Our goals here are:
+
+- A) Reduce memory usage
+- B) Reuse chunks fetched from object storage
```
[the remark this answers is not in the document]
zhulongcheng: chunks and pieces of index.

docs/proposals/201909_bucketfile.md
zhulongcheng: providers: Azure, S3, COS, GCS, Swift

all these providers support `fetch last N bytes` and `fetch everything starting from offset`.

some changes in client side:
1. Upgrade GCS client pkg to version v0.45.0+
2. Update Swift [GetRange](https://github.com/thanos-io/thanos/blob/master/pkg/objstore/swift/swift.go#L121) implementation.

https://github.com/googleapis/google-cloud-go/blob/master/CHANGES.md

docs/proposals/201909_bucketfile.md
zhulongcheng: you are right. it should be 512MiB.
128MiB is the size of WAL file.

docs/proposals/201909_bucketfile.md
zhulongcheng: typo ):
`is greater than`

docs/proposals/201909_bucketfile.md
zhulongcheng: Use [Roaring bitmaps](http://roaringbitmap.org) implementation.

Compressed Size: 2.6 bits per int.

1 TiB = 1 * 1024 * 1024 * 1024 KiB = 1073741824 KiB
4 KiB pages:  1073741824 / 4 = 268435456
bitmap size: 268435456 * 3 bits = 805306368 bits = 805306368 / 8 / 1024 / 1024 = 96 MiB

so if we write 1 TiB to local files, then the size of bitmaps is 96 MiB.

zhulongcheng: if we write many bytes (e.g. 100 TiB) to local files, bitmaps will use 10 GiB memory.

so I think we also need to close the `pages` bitmap now. ):

docs/proposals/201909_bucketfile.md
zhulongcheng: Update: `BucketFile` will track the time.

IndexReaders hold decoded `postings offsets` and `label offsets` in memory. 
so IndexReaders need to track the time.
if one IndexReader is inactive, it will remove in-memory `postings offsets` and `label offsets`.

ChunksReader no need to track.

docs/proposals/201909_bucketfile.md
zhulongcheng: related links:
Azure: https://docs.microsoft.com/en-us/rest/api/storageservices/specifying-the-range-header-for-blob-service-operations

COS: https://cloud.tencent.com/developer/section/1189987

Swift: https://docs.openstack.org/api-ref/object-store/?expanded=get-object-content-and-metadata-detail#get-object-content-and-metadata

S3: https://docs.aws.amazon.com/zh_cn/AmazonS3/latest/API/API_GetObject.html

COS: https://cloud.google.com/storage/docs/xml-api/reference-headers#range

---

COS and Swift support multiple ranges (Range: bytes=A-B,M-N)

docs/proposals/201909_bucketfile.md
zhulongcheng: This needs to design a new search algorithm and file disk format. 

Although its goal is to reduce memory, it is independent and can be described in the next design.

Should describe details in this design？

bwplotka: Sorry I thought you mean something else. 

No, it should not be part of the design. I would state it as `Consider changing index file design for (...)` as we first need to agree on this. I am quite sure we would like to be compliant with native TSDB (:

docs/proposals/201909_bucketfile.md
zhulongcheng: Ok, will simplify this.

docs/proposals/201909_bucketfile.md
zhulongcheng: Add a helper struct `MmapTracker` to manage the size of all mmaps.

The`MmapTracker` is a global obj for store gw. 

For each `BucketFile`:
1. call `MmapTracker.Allocate(size)` before `mmap` a file
2. call `MmapTracker.Free(size)` after `unmmap` a file

`MmapTracker.allocate` returns an error if the current size is greater than the configurable `MaxMmapSize`.

How about this way?  @bwplotka @GiedriusS

docs/proposals/201909_bucketfile.md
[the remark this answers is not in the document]
zhulongcheng: updated.

docs/proposals/201909_bucketfile.md
```diff
@@ -0,0 +1,329 @@
...
+
+## Motivation
+
+Currently, the store gateway loads all [index.cache.json](https://github.com/thanos-io/thanos/blob/865d5ec710b5651ead08a7dc73864a6f5dbeeaa9/pkg/block/index.go#L40)  files into memory on startup, which easily hits OOM situations. On the other hand, some indexes are rarely or never queried, so we don't have to keep these indexes in memory.
```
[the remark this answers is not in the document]
zhulongcheng: fixed

docs/proposals/201909_bucketfile.md
zhulongcheng: fixed

docs/proposals/201909_bucketfile.md
zhulongcheng: updated.

docs/proposals/201909_bucketfile.md
zhulongcheng: fixed

docs/proposals/201909_bucketfile.md
zhulongcheng: The `Risks` section has been removed. see https://github.com/thanos-io/thanos/pull/1567/commits/3376912da28e15e5da77386f210e3a1fa1e347c4

docs/proposals/201909_bucketfile.md
bwplotka: Again, for design it's really enough to say `We will track mmaped file globally per store.`

And then knowing that there are important design questions to ask, e.g
* What if one request take all MaxMmapSIze?
* Who unmmaps file? Is it get unmmapped after each request?
* We fetch index and chunk pieces on demand but once we fetch all, are we mmapping the whole file? Potentially large one?

docs/proposals/201909_bucketfile.md
```diff
@@ -0,0 +1,287 @@
...
+
+Each time to fetch pending pages, BucketFile will reset `pendingPages`, `pendingReaders` and `pendingChan`.
+
+### In-memory objects
+
+Each IndexReader decodes [postings offset block](https://github.com/prometheus/prometheus/blob/master/tsdb/docs/format/index.md#postings-offset-table) and [label offset block](https://github.com/prometheus/prometheus/blob/master/tsdb/docs/format/index.md#label-offset-table) into in-memory object for sequential and random accesses.
+
+Each BucketFile has a `pages` bitmap, but the bitmap uses small memory. 
+
+For inactive IndexReaders and BucketFiles, if the inactive duration (`now.Sub(lastQueryTime)`) is greater than a configured duration (e.g. 15 mins), then IndexReaders will remove in-memory objects, and BucketFiles will remove in-memory bitmaps, close file descriptors and unmmap local files. This will reduce memory usage.
```
bwplotka: Nice, what about disk space? I would say we should try to have limited space as well.

BTW can we remove pieces of e.g index from memory or only full index? or from disk? E.g questions is can we make it sparse for things that are not used.. (Probably too much effort)

docs/proposals/201909_bucketfile.md
bwplotka: Preload what?  Full block to memory? last N hours of it? That sounds scary for large objects. Even downloading to disk will take ages. :thinking:

docs/proposals/201909_bucketfile.md
```diff
@@ -0,0 +1,287 @@
...
+The latest blocks will be queried the most. 
+
+Preload `the last N hours` (configurable) blocks on startup.
+This will fetch index files from object storage and decode bytes into in-memory objects.
+
+### Local files
+
+Because the max size of a index file is 64 GiB, and the default size of a chunks file is 512 MiB.
+
+So we can set the size of a local index file to 64 GiB, set the size of a local chunks file to 768 (512*1.5) MiB.
```
bwplotka: Do we really need to know/set max size of files here? Why?

docs/proposals/201909_bucketfile.md
```diff
@@ -0,0 +1,287 @@
...
+1. Fetch last N bytes:     
+     `Range{start:0, end: -N}` ->  Request header`Range: bytes=-N`     
+     **NOTE**: will not write to local file because we do not know the offset.
+2. Fetch everything starting from `start`:    
+    `Range{start:N, end: 0}	`  ->  Request heade `Range: bytes=N-` 
+
+3. Fetch everything starting at `start` till  `end`:    
+   `Range{start:N, end: M}` -> Request header `Range: bytes=M-N` 
+
+Now all providers (Azure, COS, GCS, S3, Swift) support `Fetch last N bytes`, `Fetch everything starting from offset` and `Fetch everything starting at start till end`.
```
bwplotka: Thanks for research, let's see first if we really need to have this working, maybe we can do it without those 3 special features of fetching ranges? 
Can you provide motivation here for them?

bwplotka: Essentially, index TOC knows exactly where things are in the first place, so we need just that :thinking:

docs/proposals/201909_bucketfile.md
zhulongcheng: Preload block index,  the `postings offset table` and `label offset table`.

The latest blocks are usually small.  
For large blocks, users can disable this configure option.

docs/proposals/201909_bucketfile.md
zhulongcheng: we do not know the size of the remote index file, so we need the Feature-1 `Fetch last N bytes` to fetch TOC.

we know the start of `postings offset table` block through [TOC](https://github.com/prometheus/tsdb/blob/master/docs/format/index.md#toc), but we do not know the end of this block, so we need the Feature-2 `Fetch everything starting from start` to fetch TOC.

docs/proposals/201909_bucketfile.md
zhulongcheng: `mmap` needs a size to allocate virtual address space. If we give a small size, when the offset of fetched data is greater than mmap size,  we need to re-mmap.

docs/proposals/201909_bucketfile.md
zhulongcheng: BucketFiles will also remove local files.  (will update the doc)
Maybe we can track disk space, just like tracking mmap size.

zhulongcheng: IndexReaders decode the `postings offset block` and `label offset block` into memory. 
we remove those in-memory objs.

The disk index is represented by BucketFile. If BucketFile is inactive, BucketFile will remove the disk index.

docs/proposals/201909_bucketfile.md
zhulongcheng: TODO: check overfetch.

GiedriusS:
How did this end? Could someone give a summary? I think this is quite important.

bwplotka:
Yes! Let's get back to this. Sorry for delay @zhulongcheng and thanks for the good work (: 

I think we need to improve this proposal and split it into smaller ones as we touch many improvements that we should add separately! For example, the optimized fetching index header can be beneficial for current flow as well! (: This allows iterating over those quicker.

Also, are you fine @zhulongcheng to maybe us maintainers to propose some PRs, commits to your PR so we can work on this together? 

I would say let's focus @zhulongcheng on this:

1. Treat this proposal mainly for the goal of adding **DISK CACHING** to Thanos. In particular, this is one of many possible disk cache formats we can use (: e.g we could use just https://github.com/peterbourgon/diskv (:  I think a disk cache + index cache composed in sparse file implementation is what https://github.com/thanos-io/thanos/issues/1471 proposes.
1. Extract constructing index-cache (the initial header) improvements we talk about in this proposal to the separate issue/proposal. I mean this part with fetching last item etc. I think it's super-valuable but we don't need sparse file implementation for this.
1. Open issue/proposal for the improvements to the TSDB index file itself you had @zhulongcheng in future work! This is an incredibly valuable idea that is totally fair (: I would call it `make index file object storage-friendly` (: 


Overall let's focus on sparse file implementation here in this PR maybe. Overall I have still concerns around:
* partitioning requests against object storage
* granular eviction of items. 

However, by splitting, we can keep the discussion focused! Thoughts?

bwplotka:
Particularly once split we can update: https://github.com/thanos-io/thanos/issues/1705

docs/proposals/201909_bucketfile.md
```diff
@@ -0,0 +1,287 @@
...
+  - Reimplement Swift `GetRange` 
+- B) Implement `BucketFile`
+  
+- C) Use `BucketFile` for `BucketIndexReader` and `BucketchunkReader`
+
+- D) Add Flags to enable `BucketFile` 
+
+### Future work
+
+Design a new index file for [postings offset entries](https://github.com/prometheus/prometheus/blob/master/tsdb/docs/format/index.md#postings-offset-table) and [label offset entries](https://github.com/prometheus/prometheus/blob/master/tsdb/docs/format/index.md#label-offset-table) for sequential and random accesses.
```
bwplotka: a new index-cache /index-header I guess? Yes! And it might be the first step /treated separately not future work (:

docs/proposals/201909_bucketfile.md
bwplotka: > we know the start of postings offset table block through TOC, but we do not know the end of this block, so we need the Feature-2 Fetch everything starting from start to fetch TOC.

yes, you are right here.

bwplotka:
Pulled composing index-cache to another issue: https://github.com/thanos-io/thanos/issues/1711

zhulongcheng:
> ... maybe us maintainers to propose some PRs, commits to your PR so we can work on this together?

Great! Please commit :)

docs/proposals/201909_bucketfile.md
zhulongcheng: removed.

PR #1569, opened 2019-09-25
tests: only adjust SysProcAttr on linux

alfred-landrum:
Fixes compilation and local test runs on darwin.

Signed-off-by: Alfred Landrum <alfred@leakybucket.org>


krasi-georgiev:
so this one can be close now.

brancz:
Sorry I missed this PR. #1584 does the same. Closing here.

PR #1234, opened 2019-06-07
objstore : implement Aliyun OSS

shaulboozhiao:
* [] CHANGELOG entry if change is relevant to the end user.

My company infra is worked with Aliyun Cloud , the cold-monitoring data have to be shipped to Aliyun OSS.This PR works to implement it. The code I submit can upload prometheus data to aliyun oss , it also works well to go with `thanos bucket ls/verify/inspect` although I only deploy it in test env , because I need to make `thanos` working with the whole framework ,like service discovery & load balance .I will work on this to make aliyun oss objstore work well . If you want something  such as doc, I will do it later . 

In order to make this PR, I pass over the `FOREACH` test , the print seem to be OK . although along  with this test ,  i jump over the "bucket creating & bucket delete" testing , I will do it next week .

ps. The object-config.file format like 
object.config-file format 
type: OSS
config:
  bucket: "XXXX"
  endpoint: "XXXX"
  secret_key: "XXXX"
  secret_id: "XXXX"

## Changes
pull-d67a69d & pull-4abb631 is the first try,  they work for implement aliyun oss objstore

pull-f973be4 can be ignored , I was stupid to push my local changed compiling  files `go.mod` & `go.sum`  to repository , so i have to restore them .

pull-453829 & pull-86c5548 is uploading some images , IGNORE THEM ALSO

## Verification

aliyun oss 
https://github.com/shaulboozhiao/thanos/blob/master/docs/img/thanos.jpg

`FOREACH` test result 
[thanos@XXX objtesting]$ go test acceptance_e2e_test.go  foreach.go 
ok      command-line-arguments  0.160s

and aliyun oss page 

foreach test 
https://github.com/shaulboozhiao/thanos/blob/master/docs/img/foreachtest.jpg

PR #1573, opened 2019-09-26
objstore : implement Aliyun OSS

wujinhu:
Open this PR as discussed in #1234 ， will update soon.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -160,8 +160,8 @@ docs: $(EMBEDMD) build
 .PHONY: check-docs
 check-docs: $(EMBEDMD) $(LICHE) build
 	@EMBEDMD_BIN="$(EMBEDMD)" scripts/genflagdocs.sh check
-	@$(LICHE) --recursive docs --exclude "cloud.tencent.com" --document-root .
-	@$(LICHE) --exclude "cloud.tencent.com|goreportcard.com" --document-root . *.md
+	@$(LICHE) --recursive docs --exclude "(cloud.tencent.com|alibabacloud.com)" --document-root .
+	@$(LICHE) --exclude "(cloud.tencent.com|alibabacloud.com)" --document-root . *.md
 
 # checks Go code comments if they have trailing period (excludes protobuffers and vendor files).
 # Comments with more than 3 spaces at beginning are omitted from the check, example: '//    - foo'.
--- a/Makefile
+++ b/Makefile
@@ -161,7 +161,7 @@ docs: $(EMBEDMD) build
 check-docs: $(EMBEDMD) $(LICHE) build
 	@EMBEDMD_BIN="$(EMBEDMD)" scripts/genflagdocs.sh check
 	@$(LICHE) --recursive docs --exclude "(cloud.tencent.com|alibabacloud.com)" --document-root .
-	@$(LICHE) --exclude "(cloud.tencent.com|alibabacloud.com)" --document-root . *.md
+	@$(LICHE) --exclude "(cloud.tencent.com|goreportcard.com|alibabacloud.com)" --document-root . *.md
 
 # checks Go code comments if they have trailing period (excludes protobuffers and vendor files).
 # Comments with more than 3 spaces at beginning are omitted from the check, example: '//    - foo'.
--- a/Makefile
+++ b/Makefile
@@ -203,17 +203,19 @@ test: check-git install-deps
 	@go install github.com/thanos-io/thanos/cmd/thanos
 	# Be careful on GOCACHE. Those tests are sometimes using built Thanos/Prometheus binaries directly. Don't cache those.
 	@rm -rf ${GOCACHE}
-	@echo ">> running all tests. Do export THANOS_SKIP_GCS_TESTS='true' or/and THANOS_SKIP_S3_AWS_TESTS='true' or/and THANOS_SKIP_AZURE_TESTS='true' and/or THANOS_SKIP_SWIFT_TESTS='true' and/or THANOS_SKIP_TENCENT_COS_TESTS='true' if you want to skip e2e tests against real store buckets"
+	@echo ">> running all tests. Do export THANOS_SKIP_GCS_TESTS='true' or/and THANOS_SKIP_S3_AWS_TESTS='true' or/and THANOS_SKIP_AZURE_TESTS='true' and/or THANOS_SKIP_SWIFT_TESTS='true' and/or THANOS_SKIP_ALIYUN_OSS_TESTS='true' and/or THANOS_SKIP_TENCENT_COS_TESTS='true' if you want to skip e2e tests against real store buckets"
 	@go test $(shell go list ./... | grep -v /vendor/);
 
 .PHONY: test-ci
 test-ci: export THANOS_SKIP_AZURE_TESTS = true
 test-ci: export THANOS_SKIP_SWIFT_TESTS = true
 test-ci: export THANOS_SKIP_TENCENT_COS_TESTS = true
+test-ci: export THANOS_SKIP_ALIYUN_OSS_TESTS = true
 test-ci:
 	@echo ">> Skipping AZURE tests"
 	@echo ">> Skipping SWIFT tests"
 	@echo ">> Skipping TENCENT tests"
+	@echo ">> Skipping ALIYUN tests"
 	$(MAKE) test
 
 .PHONY: test-local
--- a/docs/storage.md
+++ b/docs/storage.md
@@ -60,6 +60,7 @@ Current object storage client implementations:
 | [Azure Storage Account](./storage.md#azure) | Stable  (production usage) | yes       | @vglafirov   |
 | [OpenStack Swift](./storage.md#openstack-swift)      | Beta  (working PoCs, testing usage)               | no        | @sudhi-vm   |
 | [Tencent COS](./storage.md#tencent-cos)          | Beta  (testing usage)                   | no        | @jojohappy          |
+| [AliYun OSS](./storage.md#aliyun-oss)           | Beta  (testing usage)                   | no        | @shaulboozhiao,@wujinhu      |
 
 NOTE: Currently Thanos requires strong consistency (write-read) for object store implementation.
 
@@ -336,3 +337,20 @@ config:
 ```
 
 Set the flags `--objstore.config-file` to reference to the configuration file.
+
+##  AliYun OSS Configuration
+In order to use AliYun OSS object storage, you should first create a bucket with proper Storage Class , ACLs and get the access key on the AliYun cloud. Go to [https://www.alibabacloud.com/product/oss](https://www.alibabacloud.com/product/oss) for more detail.
+
+To use AliYun OSS object storage, please specify following yaml configuration file in `objstore.config*` flag.
+
+[embedmd]:# (flags/config_aliyunoss.txt $)
+```$
+type: ALIYUNOSS
+config:
+  endpoint: ""
+  bucket: ""
+  access_key_id: ""
+  access_key_secret: ""
+```
+
+Use --objstore.config-file to reference to this configuration file.
--- a/docs/storage.md
+++ b/docs/storage.md
@@ -343,7 +343,7 @@ In order to use AliYun OSS object storage, you should first create a bucket with
 
 To use AliYun OSS object storage, please specify following yaml configuration file in `objstore.config*` flag.
 
-[embedmd]:# (flags/config_aliyunoss.txt $)
+[embedmd]:# (flags/config_bucket_aliyunoss.txt $)
 ```$
 type: ALIYUNOSS
 config:
--- a/pkg/objstore/oss/oss.go
+++ b/pkg/objstore/oss/oss.go
@@ -114,7 +114,7 @@ func (b *Bucket) Upload(ctx context.Context, name string, r io.Reader) error {
 			for ; chunk < chunksnum; chunk++ {
 				part, err := uploadEveryPart(alioss.MaxPartSize, chunk+1)
 				if err != nil {
-					return errors.Wrap(err,"failed to upload every part")
+					return errors.Wrap(err, "failed to upload every part")
 				}
 				parts = append(parts, part)
 			}
--- a/pkg/objstore/oss/oss.go
+++ b/pkg/objstore/oss/oss.go
@@ -18,7 +18,6 @@ import (
 	"github.com/go-kit/kit/log"
 	"github.com/pkg/errors"
 	"github.com/thanos-io/thanos/pkg/objstore"
-	"github.com/thanos-io/thanos/pkg/runutil"
 	"gopkg.in/yaml.v2"
 )
 
@@ -302,11 +301,6 @@ func (b *Bucket) getRange(ctx context.Context, name string, off, length int64) (
 
 	resp, err := b.bucket.GetObject(name, opts...)
 	if err != nil {
-		return nil, err
-	}
-
-	if _, err := resp.Read(nil); err != nil {
-		runutil.CloseWithLogOnErr(b.logger, resp, "oss get range obj close")
 		return nil, err
 	}
 
--- a/pkg/objstore/oss/oss.go
+++ b/pkg/objstore/oss/oss.go
@@ -1,0 +1,380 @@
+package oss
+
+import (
+	"context"
+	"fmt"
+	"io"
+	"io/ioutil"
+	"math"
+	"math/rand"
+	"net/http"
+	"os"
+	"strconv"
+	"strings"
+	"testing"
+	"time"
+
+	alioss "github.com/aliyun/aliyun-oss-go-sdk/oss"
+
+	"github.com/go-kit/kit/log"
+	"github.com/pkg/errors"
+	"github.com/thanos-io/thanos/pkg/objstore"
+	"github.com/thanos-io/thanos/pkg/runutil"
+	"gopkg.in/yaml.v2"
+)
+
+type Config struct {
+	Endpoint        string `yaml:"endpoint"`
+	Bucket          string `yaml:"bucket"`
+	AccessKeyID     string `yaml:"access_key_id"`
+	AccessKeySecret string `yaml:"access_key_secret"`
+}
+
+type Bucket struct {
+	name   string
+	logger log.Logger
+	client *alioss.Client
+	config Config
+	bucket *alioss.Bucket
+}
+
+func NewTestBucket(t testing.TB) (objstore.Bucket, func(), error) {
+	c := configFromEnv()
+	err := func(conf Config) error {
+		if conf.Endpoint == "" {
+			return errors.New("no aliyun oss endpoint in config file")
+		}
+		if conf.AccessKeyID == "" {
+			return errors.New("access_key_id is not present in config file")
+		}
+		if conf.AccessKeySecret == "" {
+			return errors.New("access_key_secret is not present in config file")
+		}
+		return nil
+	}(c)
+	if err != nil {
+		return nil, nil, err
+	}
+	if c.Bucket != "" && os.Getenv("THANOS_ALLOW_EXISTING_BUCKET_USE") == "true" {
+		t.Log("ALIYUNOSS_BUCKET is defined. Normally this tests will create temporary bucket " +
+			"and delete it after test. Unset ALIYUNOSS_BUCKET env variable to use default logic. If you really want to run " +
+			"tests against provided (NOT USED!) bucket, set THANOS_ALLOW_EXISTING_BUCKET_USE=true.")
+		return NewTestBucketFromConfig(t, c, true)
+	}
+	return NewTestBucketFromConfig(t, c, false)
+}
+
+func calculateChunks(name string, r io.Reader) (int, int64, error) {
+	switch r.(type) {
+	case *os.File:
+		f, _ := r.(*os.File)
+		if fileInfo, err := f.Stat(); err == nil {
+			s := fileInfo.Size()
+			return int(math.Floor(float64(s) / alioss.MaxPartSize)), s % alioss.MaxPartSize, nil
+		}
+	case *strings.Reader:
+		f, _ := r.(*strings.Reader)
+		return int(math.Floor(float64(f.Size()) / alioss.MaxPartSize)), f.Size() % alioss.MaxPartSize, nil
+	}
+	return -1, 0, errors.New("unsupported implement of io.Reader")
+}
+
+// Upload the contents of the reader as an object into the bucket.
+func (b *Bucket) Upload(ctx context.Context, name string, r io.Reader) error {
+	chunksnum, lastslice, err := calculateChunks(name, r)
+	if err != nil {
+		return err
+	}
+
+	ncloser := ioutil.NopCloser(r)
+	switch chunksnum {
+	case 0:
+		if err := b.bucket.PutObject(name, ncloser); err != nil {
+			return errors.Wrap(err, "failed to upload oss object")
+		}
+	default:
+		{
+			init, err := b.bucket.InitiateMultipartUpload(name)
+			if err != nil {
+				return errors.Wrap(err, "failed to initiate multi-part upload")
+			}
+			chunk := 0
+			uploadEveryPart := func(everypartsize int64, cnk int) (alioss.UploadPart, error) {
+				prt, err := b.bucket.UploadPart(init, ncloser, everypartsize, cnk)
+				if err != nil {
+					if err := b.bucket.AbortMultipartUpload(init); err != nil {
+						return prt, errors.Wrap(err, "failed to upload multi-part chunk")
+					}
+				}
+				return prt, nil
+			}
+			var parts []alioss.UploadPart
+			for ; chunk < chunksnum; chunk++ {
+				part, err := uploadEveryPart(alioss.MaxPartSize, chunk+1)
+				if err != nil {
+					return err
+				}
+				parts = append(parts, part)
+			}
+			if lastslice != 0 {
+				part, err := uploadEveryPart(lastslice, chunksnum+1)
+				if err != nil {
+					return errors.Wrap(err, "failed to upload the last chunk")
+				}
+				parts = append(parts, part)
+			}
+			_, err = b.bucket.CompleteMultipartUpload(init, parts)
+			if err != nil {
+				return errors.Wrap(err, "failed to set multi-part upload completive")
+			}
+		}
+	}
+	return nil
+}
+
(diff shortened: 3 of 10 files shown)


wujinhu:
@bwplotka It seems a little difficult to make commits signed off, there is a lot of conflicts when I run the rebase command. I think it will be easier if I revert all those commits and submit a new one. And I will reference #1234 in the new one. What do you think of this?

woodliu:
@wujinhu What is the status now?  This guy _shaulboozhiao_ has deleted the code he had...
We are going to use Ali oss to support thanos, but can't find the supported code.

woodliu:
> @wujinhu What is the status now? This guy _shaulboozhiao_ has deleted the code he had...
> We are going to use Ali oss to support thanos, but can't find the supported code.

wujinhu:
@woodliu Sorry for the delay, will update this week.

woodliu:
> @wujinhu What is the status now? This guy _shaulboozhiao_ has deleted the code he had...
> We are going to use Ali oss to support thanos, but can't find the supported code.

I have try and merge the history code from your repo, it works, there's not much files changed, it seems that merge manually is a better way.

> @woodliu Sorry for the delay, will update this week.

I have try and merge the history code from your repo, it works, there's not much files changed, it seems that merge manually is a better way.

bwplotka:
Any luck? (: Feel free to squash the original commits into one.

wujinhu:
revert commits and apply patch!

wujinhu:
Sth is wrong with the patch, I am fixing!

wujinhu:
@bwplotka @jojohappy Please help review this PR, thanks!😆

pkg/objstore/oss/oss.go
```diff
@@ -0,0 +1,380 @@
...
+	if c.Bucket != "" && os.Getenv("THANOS_ALLOW_EXISTING_BUCKET_USE") == "true" {
+		t.Log("ALIYUNOSS_BUCKET is defined. Normally this tests will create temporary bucket " +
+			"and delete it after test. Unset ALIYUNOSS_BUCKET env variable to use default logic. If you really want to run " +
+			"tests against provided (NOT USED!) bucket, set THANOS_ALLOW_EXISTING_BUCKET_USE=true.")
+		return NewTestBucketFromConfig(t, c, true)
+	}
+	return NewTestBucketFromConfig(t, c, false)
+}
+
+func calculateChunks(name string, r io.Reader) (int, int64, error) {
```
povilasv: I think we shouldn't do this type castin, let's just take an interface which has `io.Reader` & `Size()` methods instead of supporting only `os.File` & `strings.Reader` types.

pkg/objstore/oss/oss.go
```diff
@@ -0,0 +1,380 @@
...
+)
+
+type Config struct {
+	Endpoint        string `yaml:"endpoint"`
+	Bucket          string `yaml:"bucket"`
+	AccessKeyID     string `yaml:"access_key_id"`
+	AccessKeySecret string `yaml:"access_key_secret"`
+}
+
+type Bucket struct {
```
povilasv: This is missing comment on exposed type

pkg/objstore/oss/oss.go
```diff
@@ -0,0 +1,380 @@
...
+	alioss "github.com/aliyun/aliyun-oss-go-sdk/oss"
+
+	"github.com/go-kit/kit/log"
+	"github.com/pkg/errors"
+	"github.com/thanos-io/thanos/pkg/objstore"
+	"github.com/thanos-io/thanos/pkg/runutil"
+	"gopkg.in/yaml.v2"
+)
+
+type Config struct {
```
povilasv: This is missing comment on exposed type

pkg/objstore/oss/oss.go
```diff
@@ -0,0 +1,380 @@
...
+				return errors.Wrap(err, "failed to initiate multi-part upload")
+			}
+			chunk := 0
+			uploadEveryPart := func(everypartsize int64, cnk int) (alioss.UploadPart, error) {
+				prt, err := b.bucket.UploadPart(init, ncloser, everypartsize, cnk)
+				if err != nil {
+					if err := b.bucket.AbortMultipartUpload(init); err != nil {
+						return prt, errors.Wrap(err, "failed to upload multi-part chunk")
+					}
+				}
```
povilasv: We are not returning error when UploadPart fails & AbortMultipartUpload succeeds.

The code should be something like:
```
		prt, err := b.bucket.UploadPart(init, ncloser, everypartsize, cnk)
				if errAbort != nil {
					if errAbort := b.bucket.AbortMultipartUpload(init); err != nil {
						return prt, errors.Wrap(err, "failed to upload abort multi-part chunk")
					}
                                       return prt, errors.Wrap(err, "failed to upload multi-part chunk")
				}
```

pkg/objstore/oss/oss.go
```diff
@@ -0,0 +1,380 @@
...
+				return prt, nil
+			}
+			var parts []alioss.UploadPart
+			for ; chunk < chunksnum; chunk++ {
+				part, err := uploadEveryPart(alioss.MaxPartSize, chunk+1)
+				if err != nil {
+					return err
```
povilasv:
```suggestion
					return errors.Wrap(err,"failed to upload every part")
```

pkg/objstore/oss/oss.go
```diff
@@ -0,0 +1,380 @@
...
+	c := Config{
+		Endpoint:        os.Getenv("ALIYUNOSS_ENDPOINT"),
+		Bucket:          os.Getenv("ALIYUNOSS_BUCKET"),
+		AccessKeyID:     os.Getenv("ALIYUNOSS_ACCESS_KEY_ID"),
+		AccessKeySecret: os.Getenv("ALIYUNOSS_ACCESS_KEY_SECRET"),
+	}
+	return c
+}
+
+func NewBucket(logger log.Logger, conf []byte, component string) (*Bucket, error) {
```
povilasv: We should add a comment on exported funciton

pkg/objstore/oss/oss.go
```diff
@@ -0,0 +1,380 @@
...
+		if err := yaml.Unmarshal(conf, &config); err != nil {
+			return Config{}, err
+		}
+
+		return config, nil
+	}(conf)
+
+	if err != nil {
+		return nil, errors.Wrap(err, "parse aliyun oss config file failed")
+	}
```
povilasv: Let's get rid of function creation and simplify this to:
```
	var config Config
		if err := yaml.Unmarshal(conf, &config); err != nil {
			return Config{}, errors.Wrap(err, "parse aliyun oss config file failed")
		}
```

pkg/objstore/oss/oss.go
```diff
@@ -0,0 +1,380 @@
...
+	if conf.AccessKeyID == "" {
+		return errors.New("access_key_id is not present in config file")
+	}
+	if conf.AccessKeySecret == "" {
+		return errors.New("access_key_secret is not present in config file")
+	}
+	return nil
+}
+
+func NewTestBucketFromConfig(t testing.TB, c Config, reuseBucket bool) (objstore.Bucket, func(), error) {
```
povilasv: I think this code should be in *_test.go files

pkg/objstore/oss/oss.go
jojohappy: Agree.

pkg/objstore/oss/oss.go
```diff
@@ -0,0 +1,380 @@
...
+	}
+
+	if conf.AccessKeyID == "" {
+		return errors.New("access_key_id is not present in config file")
+	}
+	if conf.AccessKeySecret == "" {
+		return errors.New("access_key_secret is not present in config file")
+	}
+	return nil
+}
```
jojohappy: Maybe we can simplify the checks, use one `if` statement instead.

pkg/objstore/oss/oss.go
```diff
@@ -0,0 +1,380 @@
...
+
+// Delete removes the object with the given name.
+func (b *Bucket) Delete(ctx context.Context, name string) error {
+	if err := b.bucket.DeleteObject(name); err != nil {
+		return errors.Wrap(err, "delete oss object")
+	}
+	return nil
+}
+
+func configFromEnv() Config {
```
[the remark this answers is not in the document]
wujinhu: OK, sounds good👍

pkg/objstore/oss/oss.go
```diff
@@ -0,0 +1,380 @@
...
+			return errors.New("access_key_id is not present in config file")
+		}
+		if conf.AccessKeySecret == "" {
+			return errors.New("access_key_secret is not present in config file")
+		}
+		return nil
+	}(c)
+	if err != nil {
+		return nil, nil, err
+	}
```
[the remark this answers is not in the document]
wujinhu: OK

pkg/objstore/oss/oss.go
wujinhu: OK

pkg/objstore/oss/oss.go
wujinhu: OK

pkg/objstore/oss/oss.go
wujinhu: Good catch!

pkg/objstore/oss/oss.go
wujinhu: OK

pkg/objstore/oss/oss.go
wujinhu: I am a little ambiguous about this. We take S3 as an example, could you please give me comments in more details. Thanks!

pkg/objstore/oss/oss.go
wujinhu: OK

pkg/objstore/oss/oss.go
wujinhu: OK

pkg/objstore/oss/oss.go
povilasv: :+1:

pkg/objstore/oss/oss.go
povilasv: I think we can do this in another PR. As this would require quite a lot of refactoring.

If we change the interface in https://github.com/thanos-io/thanos/blob/master/pkg/objstore/objstore.go#L26
 from:
```
Upload(ctx context.Context, name string, r io.Reader) error
```

to

```
type ReadSizer interface{
   io.Reader
   Size() int64
} 

Upload(ctx context.Context, name string, r ReadSizer) error
```

We would avoid this hack entirely.

@bwplotka what do you think?

wujinhu:
@povilasv @jojohappy Thanks for your comments, they are very helpful, and I updated the PR.

pkg/objstore/oss/oss.go
bwplotka: I guess there is no way to have this working without knowing the size beforehand? :thinking:

bwplotka: We could change interface, but the S3 example is bad - we should never rely on size as it block streaming case. However the truth is there is no case we need streaming currently for write (:

pkg/objstore/oss/oss.go
```diff
@@ -0,0 +1,380 @@
...
+
+type Bucket struct {
+	name   string
+	logger log.Logger
+	client *alioss.Client
+	config Config
+	bucket *alioss.Bucket
+}
+
+func NewTestBucket(t testing.TB) (objstore.Bucket, func(), error) {
```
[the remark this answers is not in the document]
bwplotka: Nope, this is the idiom for now. This is because _test.go files even exported methods are not exported between packages.

On the other hand this method is really the same across objectstores so we might want to refactor it to one (:

Makefile
```diff
@@ -160,8 +160,8 @@ docs: $(EMBEDMD) build
 .PHONY: check-docs
 check-docs: $(EMBEDMD) $(LICHE) build
 	@EMBEDMD_BIN="$(EMBEDMD)" scripts/genflagdocs.sh check
-	@$(LICHE) --recursive docs --exclude "cloud.tencent.com" --document-root .
-	@$(LICHE) --exclude "cloud.tencent.com|goreportcard.com" --document-root . *.md
+	@$(LICHE) --recursive docs --exclude "(cloud.tencent.com|alibabacloud.com)" --document-root .
+	@$(LICHE) --exclude "(cloud.tencent.com|alibabacloud.com)" --document-root . *.md
```
bwplotka: Can we leave `goreport`? We do this as those servers are in different locations thus very often the slow CI cannot reach it in a given time ):

docs/storage.md
```diff
@@ -60,6 +60,7 @@ Current object storage client implementations:
 | [Azure Storage Account](./storage.md#azure) | Stable  (production usage) | yes       | @vglafirov   |
 | [OpenStack Swift](./storage.md#openstack-swift)      | Beta  (working PoCs, testing usage)               | no        | @sudhi-vm   |
 | [Tencent COS](./storage.md#tencent-cos)          | Beta  (testing usage)                   | no        | @jojohappy          |
+| [AliYun OSS](./storage.md#aliyun-oss)           | Beta  (testing usage)                   | no        | @shaulboozhiao,@wujinhu      |
```
bwplotka: :+1:

docs/storage.md
```diff
@@ -336,3 +337,20 @@ config:
 ```
 
 Set the flags `--objstore.config-file` to reference to the configuration file.
+
+##  AliYun OSS Configuration
+In order to use AliYun OSS object storage, you should first create a bucket with proper Storage Class , ACLs and get the access key on the AliYun cloud. Go to [https://www.alibabacloud.com/product/oss](https://www.alibabacloud.com/product/oss) for more detail.
```
bwplotka: What proper `Storage Class` means?

pkg/objstore/oss/oss.go
```diff
@@ -0,0 +1,349 @@
...
+	"math/rand"
+	"net/http"
+	"os"
+	"strconv"
+	"strings"
+	"testing"
+	"time"
+
+	alioss "github.com/aliyun/aliyun-oss-go-sdk/oss"
+
```
bwplotka: Please kill this newline and run `make format` (Current tooling limitation)

pkg/objstore/oss/oss.go
bwplotka: Yea ok it looks like for multipart need to do on our own: https://github.com/aliyun/aliyun-oss-go-sdk/blob/d0a2d03230103776946b757cbc05aeecb523d9c9/oss/multipart_test.go#L180

pkg/objstore/oss/oss.go
```diff
@@ -0,0 +1,349 @@
...
+				parts = append(parts, part)
+			}
+			if lastslice != 0 {
+				part, err := uploadEveryPart(lastslice, chunksnum+1)
+				if err != nil {
+					return errors.Wrap(err, "failed to upload the last chunk")
+				}
+				parts = append(parts, part)
+			}
+			_, err = b.bucket.CompleteMultipartUpload(init, parts)
```
bwplotka: we can use `if err := ...; err != nil` pattern here (:

pkg/objstore/oss/oss.go
```diff
@@ -0,0 +1,349 @@
...
+		opt, err := b.setRange(off, off+length-1, name)
+		if err != nil {
+			return nil, err
+		}
+		opts = append(opts, opt)
+	}
+
+	resp, err := b.bucket.GetObject(name, opts...)
+	if err != nil {
+		return nil, err
```
bwplotka: should we wrap those errors here? (:

pkg/objstore/oss/oss.go
povilasv: The other option is to actually read all the bytes into memory and then we would know the size, which sucks for memory requirements. 

I would prefer to have `ReadSizer`, it actually helps upload case, as the uploader client can split the object into chunks, make it parallel, etc.

bwplotka: Well that's the thing, for streaming purposes you buffer let's say 128KB then if you buffer more you know you need multipart. While you traverse the file you stream parts until end of the reader (: I think that's what GCS is doing for example. Essentially you can implement this without knowing size beforehand.

bwplotka: In fact this will be required if we would like to have this: https://github.com/thanos-io/thanos/issues/1673

povilasv: Yeah you are right. Actually in this case we shouldn't need a Size() method, we can just do what you said and upload parts the way you described.

docs/storage.md
wujinhu: It is the same meaning with S3 https://aws.amazon.com/s3/storage-classes/?nc1=h_ls
It defines the storage type of data.

pkg/objstore/oss/oss.go
wujinhu: OK

pkg/objstore/oss/oss.go
wujinhu: OK

pkg/objstore/oss/oss.go
wujinhu: OK

Makefile
wujinhu: OK

wujinhu:
=== RUN   TestObjStore_AcceptanceTest_e2e
--- PASS: TestObjStore_AcceptanceTest_e2e (1.68s)
=== RUN   TestObjStore_AcceptanceTest_e2e/inmem
    --- PASS: TestObjStore_AcceptanceTest_e2e/inmem (0.00s)
    foreach.go:49: THANOS_SKIP_GCS_TESTS envvar present. Skipping test against GCS.
    foreach.go:70: THANOS_SKIP_S3_AWS_TESTS envvar present. Skipping test against S3 AWS.
    foreach.go:86: THANOS_SKIP_AZURE_TESTS envvar present. Skipping test against Azure.
    foreach.go:102: THANOS_SKIP_SWIFT_TESTS envvar present. Skipping test against swift.
    foreach.go:118: THANOS_SKIP_TENCENT_COS_TESTS envvar present. Skipping test against Tencent COS.
=== RUN   TestObjStore_AcceptanceTest_e2e/AliYun_oss
    --- PASS: TestObjStore_AcceptanceTest_e2e/AliYun_oss (0.63s)
PASS

Process finished with exit code 0

wujinhu:
=== RUN   TestBucketStore_e2e
--- PASS: TestBucketStore_e2e (26.03s)
=== RUN   TestBucketStore_e2e/inmem
    --- PASS: TestBucketStore_e2e/inmem (10.17s)
        bucket_e2e_test.go:398: Test with no index cache
        bucket_e2e_test.go:372: Run  0
        bucket_e2e_test.go:372: Run  1
        bucket_e2e_test.go:372: Run  2
        bucket_e2e_test.go:372: Run  3
        bucket_e2e_test.go:372: Run  4
        bucket_e2e_test.go:372: Run  5
        bucket_e2e_test.go:372: Run  6
        bucket_e2e_test.go:372: Run  7
        bucket_e2e_test.go:372: Run  8
        bucket_e2e_test.go:372: Run  9
        bucket_e2e_test.go:372: Run  10
        bucket_e2e_test.go:402: Test with large, sufficient index cache
        bucket_e2e_test.go:372: Run  0
        bucket_e2e_test.go:372: Run  1
        bucket_e2e_test.go:372: Run  2
        bucket_e2e_test.go:372: Run  3
        bucket_e2e_test.go:372: Run  4
        bucket_e2e_test.go:372: Run  5
        bucket_e2e_test.go:372: Run  6
        bucket_e2e_test.go:372: Run  7
        bucket_e2e_test.go:372: Run  8
        bucket_e2e_test.go:372: Run  9
        bucket_e2e_test.go:372: Run  10
        bucket_e2e_test.go:411: Test with small index cache
        bucket_e2e_test.go:372: Run  0
        bucket_e2e_test.go:372: Run  1
        bucket_e2e_test.go:372: Run  2
        bucket_e2e_test.go:372: Run  3
        bucket_e2e_test.go:372: Run  4
        bucket_e2e_test.go:372: Run  5
        bucket_e2e_test.go:372: Run  6
        bucket_e2e_test.go:372: Run  7
        bucket_e2e_test.go:372: Run  8
        bucket_e2e_test.go:372: Run  9
        bucket_e2e_test.go:372: Run  10
    foreach.go:49: THANOS_SKIP_GCS_TESTS envvar present. Skipping test against GCS.
    foreach.go:70: THANOS_SKIP_S3_AWS_TESTS envvar present. Skipping test against S3 AWS.
    foreach.go:86: THANOS_SKIP_AZURE_TESTS envvar present. Skipping test against Azure.
    foreach.go:102: THANOS_SKIP_SWIFT_TESTS envvar present. Skipping test against swift.
    foreach.go:118: THANOS_SKIP_TENCENT_COS_TESTS envvar present. Skipping test against Tencent COS.
=== RUN   TestBucketStore_e2e/AliYun_oss
    --- PASS: TestBucketStore_e2e/AliYun_oss (14.41s)
        bucket_e2e_test.go:398: Test with no index cache
        bucket_e2e_test.go:372: Run  0
        bucket_e2e_test.go:372: Run  1
        bucket_e2e_test.go:372: Run  2
        bucket_e2e_test.go:372: Run  3
        bucket_e2e_test.go:372: Run  4
        bucket_e2e_test.go:372: Run  5
        bucket_e2e_test.go:372: Run  6
        bucket_e2e_test.go:372: Run  7
        bucket_e2e_test.go:372: Run  8
        bucket_e2e_test.go:372: Run  9
        bucket_e2e_test.go:372: Run  10
        bucket_e2e_test.go:402: Test with large, sufficient index cache
        bucket_e2e_test.go:372: Run  0
        bucket_e2e_test.go:372: Run  1
        bucket_e2e_test.go:372: Run  2
        bucket_e2e_test.go:372: Run  3
        bucket_e2e_test.go:372: Run  4
        bucket_e2e_test.go:372: Run  5
        bucket_e2e_test.go:372: Run  6
        bucket_e2e_test.go:372: Run  7
        bucket_e2e_test.go:372: Run  8
        bucket_e2e_test.go:372: Run  9
        bucket_e2e_test.go:372: Run  10
        bucket_e2e_test.go:411: Test with small index cache
        bucket_e2e_test.go:372: Run  0
        bucket_e2e_test.go:372: Run  1
        bucket_e2e_test.go:372: Run  2
        bucket_e2e_test.go:372: Run  3
        bucket_e2e_test.go:372: Run  4
        bucket_e2e_test.go:372: Run  5
        bucket_e2e_test.go:372: Run  6
        bucket_e2e_test.go:372: Run  7
        bucket_e2e_test.go:372: Run  8
        bucket_e2e_test.go:372: Run  9
        bucket_e2e_test.go:372: Run  10
=== RUN   TestBucketStore_ManyParts_e2e
--- PASS: TestBucketStore_ManyParts_e2e (17.94s)
=== RUN   TestBucketStore_ManyParts_e2e/inmem
    --- PASS: TestBucketStore_ManyParts_e2e/inmem (6.57s)
        bucket_e2e_test.go:372: Run  0
        bucket_e2e_test.go:372: Run  1
        bucket_e2e_test.go:372: Run  2
        bucket_e2e_test.go:372: Run  3
        bucket_e2e_test.go:372: Run  4
        bucket_e2e_test.go:372: Run  5
        bucket_e2e_test.go:372: Run  6
        bucket_e2e_test.go:372: Run  7
        bucket_e2e_test.go:372: Run  8
        bucket_e2e_test.go:372: Run  9
        bucket_e2e_test.go:372: Run  10
    foreach.go:49: THANOS_SKIP_GCS_TESTS envvar present. Skipping test against GCS.
    foreach.go:70: THANOS_SKIP_S3_AWS_TESTS envvar present. Skipping test against S3 AWS.
    foreach.go:86: THANOS_SKIP_AZURE_TESTS envvar present. Skipping test against Azure.
    foreach.go:102: THANOS_SKIP_SWIFT_TESTS envvar present. Skipping test against swift.
    foreach.go:118: THANOS_SKIP_TENCENT_COS_TESTS envvar present. Skipping test against Tencent COS.
=== RUN   TestBucketStore_ManyParts_e2e/AliYun_oss
    --- PASS: TestBucketStore_ManyParts_e2e/AliYun_oss (9.57s)
        bucket_e2e_test.go:372: Run  0
        bucket_e2e_test.go:372: Run  1
        bucket_e2e_test.go:372: Run  2
        bucket_e2e_test.go:372: Run  3
        bucket_e2e_test.go:372: Run  4
        bucket_e2e_test.go:372: Run  5
        bucket_e2e_test.go:372: Run  6
        bucket_e2e_test.go:372: Run  7
        bucket_e2e_test.go:372: Run  8
        bucket_e2e_test.go:372: Run  9
        bucket_e2e_test.go:372: Run  10
=== RUN   TestBucketStore_TimePartitioning_e2e
--- PASS: TestBucketStore_TimePartitioning_e2e (6.19s)
    bucket_e2e_test.go:520: Run 0
PASS

Process finished with exit code 0

        bucket_e2e_test.go:372: Run  2
        bucket_e2e_test.go:372: Run  3
        bucket_e2e_test.go:372: Run  4
        bucket_e2e_test.go:372: Run  5
        bucket_e2e_test.go:372: Run  6
        bucket_e2e_test.go:372: Run  7
        bucket_e2e_test.go:372: Run  8
        bucket_e2e_test.go:372: Run  9
        bucket_e2e_test.go:372: Run  10
=== RUN   TestBucketStore_TimePartitioning_e2e
--- PASS: TestBucketStore_TimePartitioning_e2e (6.19s)
    bucket_e2e_test.go:520: Run 0
PASS

Process finished with exit code 0

wujinhu:
@bwplotka acceptance_e2e_test.go and bucket_e2e_test.go both are ok.

pkg/objstore/oss/oss.go
```diff
@@ -0,0 +1,347 @@
...
+	}
+
+	if _, err := resp.Read(nil); err != nil {
```
povilasv: Why are we reading nil bytes here?

wujinhu: You are right, I ran oss and s3 tests and there is no need here because oss sdk is not same with s3 sdk.
```go
// NotFoundObject error is revealed only after first Read. This does the initial GetRequest. Prefetch this here
// for convenience.
if _, err := r.Read(nil); err != nil {
	runutil.CloseWithLogOnErr(b.logger, r, "s3 get range obj close")

	// First GET Object request error.
	return nil, err
}
```

wujinhu:
@bwplotka @povilasv @jojohappy I think each comment has been addressed. 😆

PR #1581, opened 2019-09-27
Set minimum buffer size of BytesPool to maxChunk size

### LINKED ISSUE #1580: store: insufficient byte pool size when loading buckets
The thanos-querier of our installation failed yesterday, caused by an `allocate chunk bytes: pool exhausted` error.

Inspection revealed that the pool.BytesPool is using a range of pooled `[]byte` sizes from `2e5` (200KiB) to `5e6` (500MiB).

The `bucketChunkReader` requests `[]byte` of 16000 bytes that are handed the smallest possible pooled buffer (200KiB) causing each `[]byte` requested from the pool to occupy 184 KiB more than required for the read 16000 bytes.

This means that in our current installation with a max. pool size of 4GiB the pool can only accomodate 21.474 chunks (327,67 MiB), which is too small for chunks up to ~500 MiB.

**Used Thanos version: Thanos 0.7.0**

master branch still has these settings.

The store should either reduce the minimal sizer to the used size or ditch the BytesPool in favour of directly using sync.Pool for the required slices of the given size.

As far as I can tell, the Store is the only user of `pool.BytesPool` in the codebase.

jabbrwcky:
This PR fixes #1580 

It fixes the immediate issue, but it could be worthy to discuss if pool.BytesPool is necessary at all or could be replaced by a simpler direct usage of sync.Pool

* Match store bytes pool size to size of chunks used by chunkByteReader.

## Changes

* extract constant nested in method to package level
* change pool.BytesPool construction to use chunk size as smallest bucket size

## Verification

The store does no longer exhaust its byte pool size on larger buckets

### DIFF
--- a/pkg/store/bucket.go
+++ b/pkg/store/bucket.go
@@ -52,6 +52,8 @@ import (
 // Take a look at Figure 6 in this whitepaper http://www.vldb.org/pvldb/vol8/p1816-teller.pdf.
 const maxSamplesPerChunk = 120
 
+const maxChunkSize = 16000
+
 type bucketStoreMetrics struct {
 	blocksLoaded          prometheus.Gauge
 	blockLoads            prometheus.Counter
@@ -241,7 +243,7 @@ func NewBucketStore(
 		return nil, errors.Errorf("max concurrency value cannot be lower than 0 (got %v)", maxConcurrent)
 	}
 
-	chunkPool, err := pool.NewBytesPool(2e5, 50e6, 2, maxChunkPoolBytes)
+	chunkPool, err := pool.NewBytesPool(maxChunkSize, 50e6, 2, maxChunkPoolBytes)
 	if err != nil {
 		return nil, errors.Wrap(err, "create chunk pool")
 	}
@@ -1768,8 +1770,6 @@ func (r *bucketChunkReader) addPreload(id uint64) error {
 
 // preload all added chunk IDs. Must be called before the first call to Chunk is made.
 func (r *bucketChunkReader) preload(samplesLimiter *Limiter) error {
-	const maxChunkSize = 16000
-
 	var g run.Group
 
 	numChunks := uint64(0)


jabbrwcky:
> More sizes we have it defeats the purpose of sync pool indeed. But for queries that uses lot's of single chunks (e.g especially for downsampled data) having smaller pools might make sense. Any chance for some benchmarks before & after? (: In terms of memory usage for the same requests?

The memory usage is only positively affected as described in the linked issue.

The main issue is that the individual chunk files are fetched in parallel in 16k byte parts when the chunk is read.

So this one:

Screenshot 2019-10-02 at 16 12 08

would be fetched in ~ 1,244 concurrent requests - each downloader requests 16k from the byte pool which caused in the previous version to grab 248.75 MB from the byte pool (because the 16k request was allocated the smallest buffer size: 200k), with the changed version it is only 24,875 MB (smallest buffer: 20k)

The issue appears on primarily when quering as far as I can tell, because the byte chunks are allocated in https://github.com/thanos-io/thanos/blob/4c7ecd3f86c5ce1f59f54f03a6dceec82d21b77e/pkg/store/bucket.go#L1272
and only returned when the reader is closed in https://github.com/thanos-io/thanos/blob/4c7ecd3f86c5ce1f59f54f03a6dceec82d21b77e/pkg/store/bucket.go#L1887

For some of our index chunk files with 500+MB this amounted to a required pool size of 6.25GB to successfully read that chunk file, which blew our Max BytePool size of 4GB, resulting in the `byte pool exhausted` error when querying.

With the change it 'only' allocates 625 MB for reading the 500MB chunk, which is not optimal but OK and fixes the issue for us for now (on second thought I should have set the smallest bucket to 16000, which is pretty much the only used size).

bwplotka:
OK I think there are few misunderstandings but overall this fix makes sense (: 

* `000001` chunk file stores multiple chunks (!). At least one for each series. 
*  Partitioning tries to compose request against bucket (and that's the byte slice we take) into bigger one: https://github.com/thanos-io/thanos/blob/138909636249036fec1b6f3922317e7e01fb586b/pkg/store/bucket.go#L1789 This means that we will request a minimum 16k ONLY if for some reason our query touches only single chunk (it can happen though for short time range queries and when we touch downsampled data). 

> would be fetched in ~ 1,244 concurrent requests - each downloader requests 16k from the byte pool which caused in the previous version to grab 248.75 MB from the byte pool (because the 16k request was allocated the smallest buffer size: 200k), with the changed version it is only 19,904 MB (smallest buffer: 16k)

Partitioning makes this really unlikely - however still you can exhaust pool with multiple specific queries at the same time. 

> and only returned when the reader is closed in

Yes - which is done once the requests finish so it should be fine (?)

> For some of our index chunk files with 500+MB this amounted to a required pool size of 6.25GB to successfully read that chunk file, which blew our Max BytePool size of 4GB, resulting in the byte pool exhausted error when querying.

That makes sense, but I guess increasing the limit (and memory) will work.

> With the change it 'only' allocates the chunk size (plus a few bytes if the chunk is not evenly divisible by 16k) for reading the 500MB chunk, which is a lot better and fixes the issue for us for now

This means to me that your queries are hitting small chunks a lot as I mentioned before.

PR #1582, opened 2019-09-28
fix rule: fixed issue with alertmanager URL containing +

### LINKED ISSUE #1564: ruler: alertmanager.url `+` symbol in password for basic auth interprets start of url as lookup scheme
Encountered on Thanos v0.7.0

Running 
```bash
docker run thanosio/thanos:v0.7.0 rule 
    --alertmanagers.url=https://user:pass+word@domain.com 
    --query=sample.com
``` 
as password contains a '+', the part of the url up to the plus gets interpreted as a lookup scheme causing an error 
```
level=error ts=2019-09-24T15:43:08.687515899Z caller=rule.go:360 msg="refreshing alertmanagers failed" err="alertmanager resolve: invalid lookup scheme \"https://user:pass\""
```

FUSAKLA:
fixes #1564 

## Changes

- fixed issue when Alertmanager URL contains `+` which can be for example in credentials 

## Verification
- added tests which are passing
- tested to startup the rule with various URLs

### DIFF
--- a/cmd/thanos/rule.go
+++ b/cmd/thanos/rule.go
@@ -620,7 +620,7 @@ func parseAlertmanagerAddress(addr string) (qType dns.QType, parsedUrl *url.URL,
 	if err != nil {
 		return qType, nil, err
 	}
-	// The Scheme might contain DNS resolver type separated by + so we split it a part
+	// The Scheme might contain DNS resolver type separated by + so we split it a part.
 	if schemeParts := strings.Split(parsedUrl.Scheme, "+"); len(schemeParts) > 1 {
 		parsedUrl.Scheme = schemeParts[len(schemeParts)-1]
 		qType = dns.QType(strings.Join(schemeParts[:len(schemeParts)-1], "+"))
--- a/cmd/thanos/rule.go
+++ b/cmd/thanos/rule.go
@@ -20,7 +20,7 @@ import (
 	"github.com/go-kit/kit/log"
 	"github.com/go-kit/kit/log/level"
 	"github.com/oklog/run"
-	opentracing "github.com/opentracing/opentracing-go"
+	"github.com/opentracing/opentracing-go"
 	"github.com/pkg/errors"
 	"github.com/prometheus/client_golang/prometheus"
 	"github.com/prometheus/common/model"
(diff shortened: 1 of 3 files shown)


FUSAKLA:
Thanks! Yes, good point. The function handles just parsing not validation. The validation is done in the resolver package so it eventually errors out afterwards (below example). This way the rule does not necessarily know all the resolvers types and who knows, maybe there will be some QType with `+` :smile: 
But yeah, I'm ok with erroring out if you think it will be safer this way.

Example:
```
$ ./thanos rule --alertmanagers.url=foo+bar+http:user:pass+word@test.haha --query=http://test.dev
...
level=info ts=2019-09-28T13:12:36.055664442Z caller=rule.go:591 msg="starting rule node"
...
level=error ts=2019-09-28T13:12:36.055936546Z caller=rule.go:363 msg="refreshing alertmanagers failed" err="alertmanager resolve: invalid lookup scheme \"foo+bar\""

```

GiedriusS:
> Thanks! Yes, good point. The function handles just parsing not validation. The validation is done in the resolver package so it eventually errors out afterwards (below example). This way the rule does not necessarily know all the resolvers types and who knows, maybe there will be some QType with `+` smile
> But yeah, I'm ok with erroring out if you think it will be safer this way.
> 
> Example:
> 
> ```
> $ ./thanos rule --alertmanagers.url=foo+bar+http:user:pass+word@test.haha --query=http://test.dev
> ...
> level=info ts=2019-09-28T13:12:36.055664442Z caller=rule.go:591 msg="starting rule node"
> ...
> level=error ts=2019-09-28T13:12:36.055936546Z caller=rule.go:363 msg="refreshing alertmanagers failed" err="alertmanager resolve: invalid lookup scheme \"foo+bar\""
> ```

Ah, I haven't checked. Then it should be all good, thanks for the example output.

cmd/thanos/rule.go
```diff
@@ -614,22 +614,31 @@ func (s *alertmanagerSet) get() []*url.URL {
 
 const defaultAlertmanagerPort = 9093
 
+func parseAlertmanagerAddress(addr string) (qType dns.QType, parsedUrl *url.URL, err error) {
+	qType = ""
+	parsedUrl, err = url.Parse(addr)
+	if err != nil {
+		return qType, nil, err
+	}
+	// The Scheme might contain DNS resolver type separated by + so we split it a part
```
bwplotka: the missing trailing period in the comment (: 

(We should write some linter finally for this ;p)

FUSAKLA: :thinking: I wrote a hook for myself to Goland to check this but it's flaky on multiline comments.
But mainly I'm ignoring it apparently :smile:

FUSAKLA:
@bwplotka fixed and rebased, thanks for noticing :))

PR #1583, opened 2019-09-30
Store,Compactor: Thanos sharding

jojohappy:

## Changes

Implement partial parts of proposal #1501 :
* Add relabel config into Thanos, store gateway and compactor are supported now.
* Selecting blocks to serve depends on the result of block labels relabeling.
* For store gateway, expose advertise labels after relabeling.
* Allow the duplicate external labels for store gateway components.

Note: pre-filter has been implemented before this PR.

TODO:
- [x] Add unit test.
- [ ] Add documents for relabeling.

@bwplotka @brancz @povilasv @GiedriusS PTAL, thanks!

### DIFF
--- a/cmd/thanos/flags.go
+++ b/cmd/thanos/flags.go
@@ -77,7 +77,7 @@ func regSelectorRelabelFlags(cmd *kingpin.CmdClause) *extflag.PathOrContent {
 	return extflag.RegisterPathOrContent(
 		cmd,
 		"selector.relabel-config",
-		"YAML file that contains seletor relabeling configuration. It follows native Prometheus relabel-config syntax. See format details: https://prometheus.io/docs/prometheus/latest/configuration/configuration/#relabel_config ",
+		"YAML file that contains relabeling configuration that allows selecting blocks. It follows native Prometheus relabel-config syntax. See format details: https://prometheus.io/docs/prometheus/latest/configuration/configuration/#relabel_config ",
 		false,
 	)
 }
--- a/cmd/thanos/flags.go
+++ b/cmd/thanos/flags.go
@@ -72,3 +72,24 @@ func regCommonTracingFlags(app *kingpin.Application) *extflag.PathOrContent {
 		false,
 	)
 }
+
+func regSelectorRelabelFlags(cmd *kingpin.CmdClause) *pathOrContent {
+	fileFlagName := "selector.relabel-config-file"
+	contentFlagName := "selector.relabel-config"
+	help := "Path to YAML file that contains seletor relabeling configuration. It follows native Prometheus relabel-config syntax. See format details: https://prometheus.io/docs/prometheus/latest/configuration/configuration/#relabel_config "
+
+	selectorRelabelConfFile := cmd.Flag(fileFlagName, help).PlaceHolder("<seletor.relabel-config-yaml-path>").String()
+
+	help = fmt.Sprintf("Alternative to '%s' flag. Relabeling configuration in YAML. See format details: https://prometheus.io/docs/prometheus/latest/configuration/configuration/#relabel_config", fileFlagName)
+
+	selectorRelabelConf := cmd.Flag(contentFlagName, help).PlaceHolder("<selector.relabel-config-yaml>").String()
+
+	return &pathOrContent{
+		fileFlagName:    fileFlagName,
+		contentFlagName: contentFlagName,
+		required:        false,
+
+		path:    selectorRelabelConfFile,
+		content: selectorRelabelConf,
+	}
+}
--- a/cmd/thanos/flags.go
+++ b/cmd/thanos/flags.go
@@ -73,23 +73,11 @@ func regCommonTracingFlags(app *kingpin.Application) *extflag.PathOrContent {
 	)
 }
 
-func regSelectorRelabelFlags(cmd *kingpin.CmdClause) *pathOrContent {
-	fileFlagName := "selector.relabel-config-file"
-	contentFlagName := "selector.relabel-config"
-	help := "Path to YAML file that contains seletor relabeling configuration. It follows native Prometheus relabel-config syntax. See format details: https://prometheus.io/docs/prometheus/latest/configuration/configuration/#relabel_config "
-
-	selectorRelabelConfFile := cmd.Flag(fileFlagName, help).PlaceHolder("<seletor.relabel-config-yaml-path>").String()
-
-	help = fmt.Sprintf("Alternative to '%s' flag. Relabeling configuration in YAML. See format details: https://prometheus.io/docs/prometheus/latest/configuration/configuration/#relabel_config", fileFlagName)
-
-	selectorRelabelConf := cmd.Flag(contentFlagName, help).PlaceHolder("<selector.relabel-config-yaml>").String()
-
-	return &pathOrContent{
-		fileFlagName:    fileFlagName,
-		contentFlagName: contentFlagName,
-		required:        false,
-
-		path:    selectorRelabelConfFile,
-		content: selectorRelabelConf,
-	}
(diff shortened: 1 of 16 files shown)


bwplotka:
Nice! It looks awesome generally. Will try to look on it ASAP.

pkg/compact/compact.go
```diff
@@ -275,6 +280,15 @@ func (c *Syncer) downloadMeta(ctx context.Context, id ulid.ULID) (*metadata.Meta
 		return nil, errors.Wrapf(err, "downloading meta.json for %s", id)
 	}
 
+	// Check for block labels by relabeling.
```
bwplotka: Nice. However I would move it out of `downloadMeta` as we do more than method suggests which might be confusing. Let's do it after meta is downloaded? Also it's nice to do it after downloadMeta method as we check if it's not too fresh first.

pkg/query/storeset.go
```diff
@@ -360,6 +360,12 @@ func externalLabelsFromStore(store *storeRef) string {
 	}
 	sort.Strings(tsdbLabelSetStrings)
 
+	// Append the storeType to the end of list, because the store gateway will be exposed external labels,
+	// we allow the duplicate external labels with different components.
+	if store.storeType != nil {
+		tsdbLabelSetStrings = append(tsdbLabelSetStrings, store.storeType.String())
```
bwplotka: Hm, this sounds like a hack, there is no way to check if this guy is a `storeGW` inside checking deduplicated stuff?

pkg/store/bucket.go
```diff
@@ -456,6 +461,16 @@ func (s *BucketStore) addBlock(ctx context.Context, id ulid.ULID) (err error) {
...
 	h := lset.Hash()
 
+	// Check for block labels by relabeling.
+	// If output is empty, the block will be dropped.
+	processedLabels := relabel.Process(promlabels.FromMap(lset.Map()), s.relabelConfig...)
+	if processedLabels == nil {
+		level.Debug(s.logger).Log("msg", "dropping block(drop in relabeling)", "block", id)
+		return os.RemoveAll(dir)
+	}
+	b.labels = labels.FromMap(processedLabels.Map())
```
bwplotka: This is important because actually we get the labels after relabelling, which we agreed not doing in the first iteration (:

pkg/store/bucket.go
```diff
@@ -519,12 +534,34 @@ func (s *BucketStore) TimeRange() (mint, maxt int64) {
...
-	// Store nodes hold global data and thus have no labels.
-	return &storepb.InfoResponse{
+	res := &storepb.InfoResponse{
 		StoreType: component.Store.ToProto(),
 		MinTime:   mint,
 		MaxTime:   maxt,
-	}, nil
+	}
+
+	if len(s.relabelConfig) != 0 {
```
bwplotka: I would vote to precompute this every sync instead of computing this on every Info which is frequent (at least every 5s)

pkg/store/bucket.go
jojohappy: So should we expose the original labels from `meta.json`? Maybe the action of relabeling will be `labeldrop`.

pkg/query/storeset.go
jojohappy: Sorry, what do you mean?

bwplotka: We are adding some artificial labels to the labelset. Why? This sounds wrong? To allow duplicated store gateway labels we just need to avoid detecting this if `store.storeType` is Gateway, right? NOT

pkg/store/bucket.go
bwplotka: Yes!

It may yes. But what's the use case to drop only some labels?

In fact it would be more useful to add new labels! This is great but there are some limitiation for it (e.g relabelling cannot generate 2 lablesets) . For now in this PR I would say we care ONLY if it exists or not exists. Does it sounds fair? (:

bwplotka:
It would really nice to add tests particulary for sharding compactor and store GW. Maybe in `/test/e2e/`?

cmd/thanos/flags.go
```diff
@@ -72,3 +72,12 @@ func regCommonTracingFlags(app *kingpin.Application) *extflag.PathOrContent {
 		false,
 	)
 }
+
+func regSelectorRelabelFlags(cmd *kingpin.CmdClause) *extflag.PathOrContent {
+	return extflag.RegisterPathOrContent(
+		cmd,
+		"selector.relabel-config",
+		"YAML file that contains seletor relabeling configuration. It follows native Prometheus relabel-config syntax. See format details: https://prometheus.io/docs/prometheus/latest/configuration/configuration/#relabel_config ",
```
bwplotka:
```suggestion
		"YAML file that contains relabeling configuration that allows selecting blocks. It follows native Prometheus relabel-config syntax. See format details: https://prometheus.io/docs/prometheus/latest/configuration/configuration/#relabel_config ",
```

pkg/query/storeset.go
```diff
@@ -218,7 +218,10 @@ func (s *StoreSet) Update(ctx context.Context) {
 	// Record the number of occurrences of external label combinations for current store slice.
 	externalLabelOccurrencesInStores := map[string]int{}
 	for _, st := range healthyStores {
-		externalLabelOccurrencesInStores[externalLabelsFromStore(st)]++
+		if st.storeType != nil && (st.storeType.ToProto() == storepb.StoreType_QUERY ||
```
bwplotka: Why are we changing this? Let's maybe focus on line 258 as you did instead (:

pkg/query/storeset.go
```diff
@@ -252,7 +255,11 @@ func (s *StoreSet) Update(ctx context.Context) {
 		// Note: No external labels means strictly store gateway or ruler and it is fine to have access to multiple instances of them.
 		// Any other component will error out if it will be configured with empty external labels.
 		externalLabels := externalLabelsFromStore(store)
-		if len(store.LabelSets()) > 0 && externalLabelOccurrencesInStores[externalLabels] != 1 {
+		if len(store.LabelSets()) > 0 &&
+			store.storeType != nil &&
+			(store.storeType.ToProto() == storepb.StoreType_QUERY ||
```
bwplotka: Can we just exclude `StoreType_STORE`? It will be much safe. It's the only thing that we change here - store GW handing, right?

pkg/store/bucket.go
```diff
@@ -521,12 +536,32 @@ func (s *BucketStore) TimeRange() (mint, maxt int64) {
...
-	// Store nodes hold global data and thus have no labels.
-	return &storepb.InfoResponse{
+	res := &storepb.InfoResponse{
 		StoreType: component.Store.ToProto(),
 		MinTime:   mint,
 		MaxTime:   maxt,
-	}, nil
+	}
+
+	labelSets := make(map[uint64][]storepb.Label, len(s.blocks))
```
bwplotka: Again as I mentioned before - we can build that every sync instead of every Info call. Can we do that? (:

bwplotka: Info should be as fast as possible as it works as healthiness/heartbeat check.

bwplotka:
It would be also awesome to have some tests for compactor filtering and store GW exposing info + filtering. Maybe in next PR (:

pkg/query/storeset.go
jojohappy: Thanks! Done!

pkg/store/bucket.go
jojohappy: I will update today. Thanks!

bwplotka:
Tests fails ):

jojohappy:
@bwplotka That's green!

xiaozongyang:
Hi @jojohappy @bwplotka 
Do we have some examples of selecting blocks with relabel configs here?

Issue #1594, opened 2019-10-01
recieve: Panics


dtrejod:
**Thanos, Prometheus and Golang version used**
```sh
$ ./thanos --version
thanos, version 0.7.0 (branch: master, revision: baf2f3a03757c0260250d93aa8ef41a79a829bb6)
  build user:       circleci@54f32563bbc6
  build date:       20191001-13:19:04
  go version:       go1.12.5
```

Prometheus version 2.12.0

**What happened**

Go routine panic.

**What you expected to happen**

No go routine panic

**How to reproduce it (as minimally and precisely as possible)**:
Simply start a receiver --
```sh
./thanos receive --log.format=logfmt --log.level=debug --tsdb.path=./test/data --objstore.config-file=./test/bucket_config.yml --receive.replication-factor=1
```

**Full logs to relevant components**

<details>Logs
<p>
level=debug ts=2019-10-01T17:51:34.346053081Z caller=main.go:122 msg="maxprocs: Leaving GOMAXPROCS=[8]: CPU quota undefined"
level=info ts=2019-10-01T17:51:34.346457856Z caller=main.go:170 msg="Tracing will be disabled"
level=warn ts=2019-10-01T17:51:34.346500366Z caller=receive.go:143 component=receive msg="setting up receive; the Thanos receive component is EXPERIMENTAL, it may break significantly without notice"
level=debug ts=2019-10-01T17:51:34.346746109Z caller=receive.go:192 component=receive msg="setting up tsdb"
level=info ts=2019-10-01T17:51:34.350840387Z caller=head.go:509 component=receive component=tsdb component=tsdb msg="replaying WAL, this may take awhile"
level=info ts=2019-10-01T17:51:34.351074421Z caller=head.go:557 component=receive component=tsdb component=tsdb msg="WAL segment loaded" segment=0 maxSegment=4
level=info ts=2019-10-01T17:51:34.35128205Z caller=head.go:557 component=receive component=tsdb component=tsdb msg="WAL segment loaded" segment=1 maxSegment=4
level=info ts=2019-10-01T17:51:34.351416083Z caller=head.go:557 component=receive component=tsdb component=tsdb msg="WAL segment loaded" segment=2 maxSegment=4
level=info ts=2019-10-01T17:51:34.352002711Z caller=head.go:557 component=receive component=tsdb component=tsdb msg="WAL segment loaded" segment=3 maxSegment=4
[...]

PR #1595, opened 2019-10-01
cmd/thanos/receive: fix close chan panic

squat:
This commit fixes a panic caused by sending on the chan which is closed
too early. This chan was closed when the setup func returned, rather
than when the group's goroutine returned.

Fixes: #1594

### DIFF
--- a/cmd/thanos/receive.go
+++ b/cmd/thanos/receive.go
@@ -370,25 +370,12 @@ func runReceive(
 			return err
 		}
 
-		// Ensure we close up everything properly.
-		defer func() {
-			if err != nil {
-				runutil.CloseWithLogOnErr(logger, bkt, "bucket client")
-			}
-		}()
-
 		s := shipper.New(logger, reg, dataDir, bkt, func() labels.Labels { return lset }, metadata.ReceiveSource)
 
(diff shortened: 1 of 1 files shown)


dtrejod:
You also close the `uploadC` channel twice.

https://github.com/thanos-io/thanos/blob/6abe080c99dbf45a7e826565b5eb17d727a494ab/cmd/thanos/receive.go#L207
https://github.com/thanos-io/thanos/blob/6abe080c99dbf45a7e826565b5eb17d727a494ab/cmd/thanos/receive.go#L413

squat:
Yes you’re totally right. This is an unrelated issue but a bug nonetheless. I’ll address this in the same pr

dtrejod:
Confirmed these changes resolved the panics. However hit another issue where the TSDB doesn't start back up after a flush event. Looks to me this cancel call closes the `upload on demand` function prematurely --

https://github.com/thanos-io/thanos/blob/e7d8707c4e94ae10758980cd2de174fb5bab473e/cmd/thanos/receive.go#L422

squat:
Yeah you’re right. That cancel is leftover from a previous iteration where the upload used a cancellable context and `cancel()` needed to be called after the upload to clean up.

PR #1599, opened 2019-10-02
chore: added makefile target to check Go code comments trailing periods

FUSAKLA:
@bwplotka :))

It should check only last line if there are multiple comment lines.

For special cases allows to disable the check when leaving more than 3 spaces after the `//` 
example: `//    test`

Also, adds separate commit which fixes all currently invalid comments.

Example output:
image

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -158,9 +158,19 @@ check-docs: $(EMBEDMD) $(LICHE) build
 	@$(LICHE) --recursive docs --exclude "cloud.tencent.com" --document-root .
 	@$(LICHE) --exclude "cloud.tencent.com|goreportcard.com" --document-root . *.md
 
+# checks Go code comments if they have trailing period (excludes protobuffers and vendor files).
+.PHONY: check-comments
+check-comments:
+	@printf ">> checking Go comments trailing periods\n\n\n"
+	@if grep -Przo --color --include \*.go --exclude \*.pb.go --exclude bindata.go --exclude-dir vendor '\n.*\s+//\s{0,3}[^\s][^\n]+[^.?!:]{2}\n[ \t]*[^/\s].*\n' ./; \
+	 then \
+	 	printf "\n\n\n Error: Found comments without trailing period. Please make sure to add them.\n\n\n." \
+	 	&& false; \
+	 fi;
+
 # format formats the code (including imports format).
 .PHONY: format
-format: $(GOIMPORTS)
+format: $(GOIMPORTS) check-comments
 	@echo ">> formatting code"
 	@$(GOIMPORTS) -w $(FILES_TO_FMT)
 
--- a/Makefile
+++ b/Makefile
@@ -159,7 +159,7 @@ check-docs: $(EMBEDMD) $(LICHE) build
 	@$(LICHE) --exclude "cloud.tencent.com|goreportcard.com" --document-root . *.md
 
 # checks Go code comments if they have trailing period (excludes protobuffers and vendor files).
-# Comments with more than 3 spaces at begining are ommited from the check, example: '//    - foo'.
+# Comments with more than 3 spaces at beginning are omitted from the check, example: '//    - foo'.
 .PHONY: check-comments
 check-comments:
 	@printf ">> checking Go comments trailing periods\n\n\n"
--- a/Makefile
+++ b/Makefile
@@ -159,6 +159,7 @@ check-docs: $(EMBEDMD) $(LICHE) build
 	@$(LICHE) --exclude "cloud.tencent.com|goreportcard.com" --document-root . *.md
 
 # checks Go code comments if they have trailing period (excludes protobuffers and vendor files).
+# Comments with more than 3 spaces at begining are ommited from the check, example: '//    - foo'.
 .PHONY: check-comments
 check-comments:
 	@printf ">> checking Go comments trailing periods\n\n\n"
--- a/Makefile
+++ b/Makefile
@@ -162,7 +162,7 @@ check-docs: $(EMBEDMD) $(LICHE) build
 .PHONY: check-comments
 check-comments:
 	@printf ">> checking Go comments trailing periods\n\n\n"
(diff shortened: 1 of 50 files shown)


Makefile
```diff
@@ -158,9 +158,19 @@ check-docs: $(EMBEDMD) $(LICHE) build
 	@$(LICHE) --recursive docs --exclude "cloud.tencent.com" --document-root .
 	@$(LICHE) --exclude "cloud.tencent.com|goreportcard.com" --document-root . *.md
 
+# checks Go code comments if they have trailing period (excludes protobuffers and vendor files).
+.PHONY: check-comments
+check-comments:
+	@printf ">> checking Go comments trailing periods\n\n\n"
+	@if grep -Przo --color --include \*.go --exclude \*.pb.go --exclude bindata.go --exclude-dir vendor '\n.*\s+//\s{0,3}[^\s][^\n]+[^.?!:]{2}\n[ \t]*[^/\s].*\n' ./; \
+	 then \
+	 	printf "\n\n\n Error: Found comments without trailing period. Please make sure to add them.\n\n\n." \
```
bwplotka:
```suggestion
	 	printf "\n\n\n Error: Found comments without trailing period. Comments has to be full sentences.\n\n\n." \
```

Makefile
```diff
@@ -158,9 +158,19 @@ check-docs: $(EMBEDMD) $(LICHE) build
 	@$(LICHE) --recursive docs --exclude "cloud.tencent.com" --document-root .
 	@$(LICHE) --exclude "cloud.tencent.com|goreportcard.com" --document-root . *.md
 
+# checks Go code comments if they have trailing period (excludes protobuffers and vendor files).
```
bwplotka: Worth to mention that we are more strict in Thanos and requires that.

Plus I think if you want this special case for `// ` with spaces it should mentioned. However TBH I would drop it (: I restrict comments to be full sentence always...

cmd/thanos/rule_test.go
```diff
@@ -20,27 +20,27 @@ func Test_parseFlagLabels(t *testing.T) {
...
 			expectErr: true,
 		},
 		{
-			s:         []string{`1abelName="LabelVal"`}, //Unsupported labelname
+			s:         []string{`1abelName="LabelVal"`}, //Unsupported labelname.
 			expectErr: true,
 		},
 		{
-			s:         []string{`label_Name"LabelVal"`}, //Missing "=" seprator
+			s:         []string{`label_Name"LabelVal"`}, //Missing "=" seprator.
```
bwplotka: Can we filter `//COMENT` as well? As it's not readable (:

pkg/alert/alert.go
```diff
@@ -203,7 +203,7 @@ func (q *Queue) Push(alerts []*Alert) {
 	q.pushed.Add(float64(len(alerts)))
 
 	// Attach external labels and drop excluded labels before sending.
-	// TODO(bwplotka): User proper relabelling with https://github.com/thanos-io/thanos/issues/660
+	// TODO(bwplotka): User proper relabelling with https://github.com/thanos-io/thanos/issues/660 .
```
bwplotka: This is inconsistent. Above you added `.` without space (:  It's up to you but let's make it consistent.

pkg/reloader/example_test.go
```diff
@@ -49,5 +49,5 @@ func ExampleReloadURLFromBase() {
 		log.Fatal(err)
 	}
 	fmt.Println(reloader.ReloadURLFromBase(u))
-	// Output: http://localhost:9090/-/reload
+	//    Output: http://localhost:9090/-/reload
```
bwplotka: This is the only exception where we used `// <multiple spaces>` I think we can add trailing period, rem spaces and remove supporting this exception.. What do you think?

FUSAKLA:
Oh, sorry it does not correct anything the script. 
I meant "it" as the PR referring to the commit https://github.com/thanos-io/thanos/pull/1599/commits/72ce608487c0ab463c32ec9adab3e6e2975d53fc

But if we agree it could also add the period automatically. But then it would be bit more complicated.

pkg/alert/alert.go
bwplotka: There is even full discussion about this ;p https://ux.stackexchange.com/questions/17331/should-a-full-sentence-html-link-include-the-period-in-the-linked-text

I would say no space?

bwplotka:
Oh I misunderstod `Also, adds separate commit which fixes all currently invalid comments.` xD I thought there is some tooling for that.

I think check is perfect for now :+1:

pkg/alert/alert.go
FUSAKLA: Ok, no space it is :+1:

Makefile
FUSAKLA: True that it should be mentioned if it stays :+1: 

I'm bit afraid to drop it. In case there was some issue with the regular then it will be easy to workaround it without blocking unrelated PRs until it's fixed since this is not that critical. 
Also I'm not sure about the lists and the edge case with testing. 

But if you say so, I can remove it.

Where would you announce this change? Is slack enough or put it to the changelog?

pkg/reloader/example_test.go
FUSAKLA: I think some lists uses it also such as
```
// foo:
//   - bar
//   - bar
...
```

Not sure if lists should also end with period?

FUSAKLA:
So now required leading space in comment after `//` and removed the spaces after URLs.

Makefile
bwplotka: >  then it will be easy to work around it without blocking unrelated PRs until it's fixed since this is not that critical.

It's always that with linting. But you can always workaround by dropping check temporarily. 

I would only mention it in thanos-dev channel. Not even CHANGELOG. It's only affecting code readability and documenting code by comments. (: so it should be fine. If it would be too strict we can always loosen this.

> Also I'm not sure about the lists and the edge case with testing.

Agree, but let's be opinionated in order to give readability.

bwplotka:
Thanks!

PR #1605, opened 2019-10-03
Moved pathOrContent to separate package.

bwplotka:

This removes some complexity, plus it allows other projects to use this (e.g https://github.com/thanos-io/thanosbench)

### DIFF
--- a/docs/components/bucket.md
+++ b/docs/components/bucket.md
@@ -42,21 +42,23 @@ Flags:
       --version            Show application version.
       --log.level=info     Log filtering level.
       --log.format=logfmt  Log format to use.
-      --tracing.config-file=<tracing.config-yaml-path>
-                           Path to YAML file that contains tracing
+      --tracing.config-file=<file-path>
+                           Path to YAML file with tracing configuration. See
+                           format details:
+                           https://thanos.io/tracing.md/#configuration
+      --tracing.config=<content>
+                           Alternative to 'tracing.config-file' flag (lower
+                           priority). Content of YAML file with tracing
                            configuration. See format details:
                            https://thanos.io/tracing.md/#configuration
-      --tracing.config=<tracing.config-yaml>
-                           Alternative to 'tracing.config-file' flag. Tracing
-                           configuration in YAML. See format details:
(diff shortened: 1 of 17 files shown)


docs/components/bucket.md
```diff
@@ -101,23 +103,23 @@ Flags:
...
+      --tracing.config-file=<file-path>
                                Path to YAML file that contains tracing
                                configuration. See format details:
                                https://thanos.io/tracing.md/#configuration
-      --tracing.config=<tracing.config-yaml>
+      --tracing.config=<content>
                                Alternative to 'tracing.config-file' flag.
-                               Tracing configuration in YAML. See format
-                               details:
+                               Content of YAML file that contains tracing
```
FUSAKLA: nit: I understand that it's composed of the generic prefix and flag help but the resulting wording "content of file that contains" sounds bit weird to me.  "Content of YAML file with tracing configuration." maybe? 
But perhaps it is just me.

pkg/extflag/pathorcontent.go
```diff
@@ -0,0 +1,70 @@
...
+type CmdClause interface {
+	Flag(name, help string) *kingpin.FlagClause
+}
+
+// RegisterPathOrContent registers PathOrContent flag in kingpinCmdClause.
+func RegisterPathOrContent(cmd CmdClause, flagName string, help string, required bool) *PathOrContent {
+	fileFlagName := fmt.Sprintf("%s-file", flagName)
+	contentFlagName := flagName
+
+	fileHelp := fmt.Sprintf("Path to %s", help)
```
FUSAKLA:
```suggestion
	fileHelp := fmt.Sprintf("Path to %s (has priority over the content flag)", help)
```
Might be worth to add the priority somewhere but not sure if this is the right way/place.

docs/components/bucket.md
bwplotka: happy to change

pkg/extflag/pathorcontent.go
bwplotka: I will add

PR #1606, opened 2019-10-04
receiver: Avoid flushing the WAL twice on startup

dtrejod:
Avoid flushing the WAL twice when starting the receiver. Instead we wait for the hashring to be loaded which will force a WAL flush. When stopping the receiver we also stop the DB before flushing to avoid re-opening the DB unnecessarily.

Signed-off-by: Devin Trejo <dtrejo@palantir.com>

* [] I added CHANGELOG entry for this change.

### DIFF
--- a/cmd/thanos/receive.go
+++ b/cmd/thanos/receive.go
@@ -149,12 +149,6 @@ func runReceive(
 		MaxBlockDuration:  tsdbBlockDuration,
 		WALCompression:    true,
 	}
-	db := receive.NewFlushableStorage(
-		dataDir,
-		log.With(logger, "component", "tsdb"),
-		reg,
-		tsdbCfg,
-	)
 
 	localStorage := &tsdb.ReadyStorage{}
 	webHandler := receive.NewHandler(log.With(logger, "component", "receive-handler"), &receive.Options{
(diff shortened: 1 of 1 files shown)


cmd/thanos/receive.go
```diff
@@ -193,29 +193,28 @@ func runReceive(
...
 			defer func() {
-				if err := db.Flush(); err != nil {
-					level.Warn(logger).Log("err", err, "msg", "failed to flush storage")
-					return
-				}
 				if err := db.Close(); err != nil {
 					level.Warn(logger).Log("err", err, "msg", "failed to close storage")
 					return
 				}
+				if err := db.Flush(); err != nil {
```
krasi-georgiev: Even if the current implementation allows that a flush can be called after `db.Close` I don't think any part of the code should use it this way. I don't think that should use any api after calling `db.Close`

cmd/thanos/receive.go
```diff
@@ -193,29 +193,28 @@ func runReceive(
...
-		}
 		g.Add(func() error {
 			defer close(dbReady)
 			defer close(uploadC)
 
+			// Before actually starting, we need to make sure the
+			// WAL is flushed. The WAL is flushed after the
+			// hashring ring is loaded.
+			startTimeMargin := int64(2 * time.Duration(tsdbCfg.MinBlockDuration).Seconds() * 1000)
+			if err := db.Open(); err != nil {
```
squat: With this change, the `db` instance is never used outside of this run group. Let's move the definition of the db instance into this block to make it more local.

cmd/thanos/receive.go
```diff
@@ -193,29 +193,28 @@ func runReceive(
...
-			return errors.Wrap(err, "flushing storage")
-		}
 		g.Add(func() error {
 			defer close(dbReady)
 			defer close(uploadC)
 
+			// Before actually starting, we need to make sure the
+			// WAL is flushed. The WAL is flushed after the
+			// hashring ring is loaded.
+			startTimeMargin := int64(2 * time.Duration(tsdbCfg.MinBlockDuration).Seconds() * 1000)
```
squat: IMO the run groups should be as much about the action of running something as possible, with these config parameters out of the way. Can we move this variable definition out of the run group but keep it in the block? i.e. line 195?

cmd/thanos/receive.go
dtrejod: Sure.. A question then, replaying the WAL on each open is quite expensive for us since we are ingesting hundreds of thousands of samples per send (potentially more in the the future).

Is there any impact on chunk upload if were to reduce the `tsdb.block-duration` by an order of magnitude. Say to 30m or smaller? This should help with overall memory consumption and WAL replay speeds.

squat: The WAL is completely empty after every flush (in fact the directory doesn’t exist). So there shouldn’t be any WAL-replaying cost when re-opening the DB after a flush.

PR #1608, opened 2019-10-04
OS X fix for `make check-comments`

ppanyukov:
OS X fix for https://github.com/thanos-io/thanos/pull/1599

- OS X grep is different from Linux and original PR didn't work
- To avoid complex OS detection in Makefile, move to external script
- Also to avoid OS X specific grep/regex (it's complex enough!), use reliable Linux distro in Docker

People may not like the dependency on Docker, of course.

No CHANGELOG, Change is not relevant to the end user.

## Changes

- Makefile (edit)
- build-check-comments.sh (add)
- Change is not relevant to the end user.

Question: should `build-check-comments.sh` live in `scripts` dir?

## Verification

Manual tests on OS X and in Linux docker containers. Tested for overrides etc.
Tested adding bad comments to trigger the check. All seems to work.

### DIFF
--- a/build-check-comments.sh
+++ b/build-check-comments.sh
@@ -7,8 +7,8 @@ set -eu
 # to make sure we work on Linux and OS X and oh even Windows! :)
 # Main issue is minor incompatibies between grep on various platforms.
 #
-# Linux: set USE_DOCKER=yes to force run in docker
-# All other platforms: uses docker by default. Set USE_DOCKER=no to override.
+# Linux: set USE_DOCKER=yes or USE_DOCKER=false to force run in docker
+# All other platforms: uses docker by default. Set USE_DOCKER=no or USE_DOCKER=false to override.
 #
 # Checks Go code comments if they have trailing period (excludes protobuffers and vendor files).
 # Comments with more than 3 spaces at beginning are omitted from the check, example: '//    - foo'.
@@ -29,8 +29,8 @@ set -eu
 declare ROOT_DIR=$(cd $(dirname ${BASH_SOURCE}) && pwd)
 declare THIS_SCRIPT="${ROOT_DIR}/$(basename "${BASH_SOURCE}")"
 
-# Image to use if we do docker-based commands, centos is most stable right?
-declare IMAGE="centos:7"
+# Image to use if we do docker-based commands. NB: busybox is no good for this.
+declare IMAGE="debian:9-slim"
 
 # User can explicitly ask to run in docker
 declare USE_DOCKER=${USE_DOCKER:=""}
@@ -40,13 +40,13 @@ declare USE_DOCKER=${USE_DOCKER:=""}
 if test "Linux" != "$(uname || true)"
 then
   # Allow overriding for non-linux platforms
-  if test "no" != "${USE_DOCKER}"
+  if test "no" != "${USE_DOCKER}" && test "false" != "${USE_DOCKER}"
   then
     USE_DOCKER="yes"
   fi
 fi
 
-if test "yes" == "${USE_DOCKER}"
+if test "yes" == "${USE_DOCKER}" || test "true" == "${USE_DOCKER}"
 then
     # Make sure we only attach TTY if we have it, CI builds won't have it.
     declare TTY_FLAG=""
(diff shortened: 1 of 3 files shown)


build-check-comments.sh
```diff
@@ -0,0 +1,105 @@
...
+#	    app := kingpin.New(filepath.Base(os.Args[0]), "A block storage based long-term storage for Prometheus")
+#   }
+
+
+# Dir of this script
+declare ROOT_DIR=$(cd $(dirname ${BASH_SOURCE}) && pwd)
+declare THIS_SCRIPT="${ROOT_DIR}/$(basename "${BASH_SOURCE}")"
+
+# Image to use if we do docker-based commands, centos is most stable right?
+declare IMAGE="centos:7"
```
FUSAKLA: Maybe I'd use some smaller image if possible? for example some `slim` version of debian?

build-check-comments.sh
```diff
@@ -0,0 +1,105 @@
...
+    # UID and GID for Linux only (this won't work on OS X anyway).
+    declare USER_FLAG=""
+    if test "Linux" == "$(uname || true)"
+    then
+        USER_FLAG="-u $(id -u):$(id -g)"
+    fi
+
+    (
+      set -x
+      docker run \
```
FUSAKLA: Maybe we could add a check if docker is available and if not, bail out with comment why it is needed?

build-check-comments.sh
```diff
@@ -0,0 +1,105 @@
...
+if test "Linux" != "$(uname || true)"
+then
+  # Allow overriding for non-linux platforms
+  if test "no" != "${USE_DOCKER}"
+  then
+    USE_DOCKER="yes"
+  fi
+fi
+
+if test "yes" == "${USE_DOCKER}"
```
FUSAKLA: Maybe check with `-z "${USE_DOCKER}"` ? Or add also `true` as an option but that's just a nit.

FUSAKLA:
@bwplotka PTAL if this workaround is ok with you.

build-check-comments.sh
ppanyukov: Good suggestion, I was wondering which image to use, busybox didn't work for obvious reasons. Is `debian:9-slim` ok?

build-check-comments.sh
ppanyukov: Adding `true` is a good idea. I probably will do that. I will have a look at `-z` although it might not work in this scenario. I'm not a big fan of these really, as it adds mental strain on me to remember and understand what those shortcut flags mean.

build-check-comments.sh
FUSAKLA: Yes, that sounds good :+1:

build-check-comments.sh
ppanyukov: currently we get this if docker isn't there:

> ./build-check-comments.sh: line 71: docker: command not found

is this good enough as it is? 

otherwise we'd rely in `which` which incidentally also may not be present, like in `centos:7` docker image for example and which may lead to false positives (have `docker` no `which`). :) Case in point, running in `centos:7` base image:

```
[root@98cad6a96d96 thanos]# yum install -y make git
[root@98cad6a96d96 thanos]# make build
make: which: Command not found
No git binary found.
make: *** [check-git] Error 1
```

I will add a `printf` saying `Make sure Docker is installed if you you get errors`

build-check-comments.sh
FUSAKLA: Sorry I off course confused those :) I meant `-n` which should match non-empty string.

ppanyukov: Well perhaps lets steer clear of these `-z` and `-n` flags? For clarity sake?

build-check-comments.sh
ppanyukov: Done

build-check-comments.sh
ppanyukov: Not using which, but added `printf`, is this better?

```
Mayas-MacBook-Pro:thanos philip$ make check-comments

 This will run in Docker.
 If you get error, ensure Docker is installed.

+ docker run ...
```

FUSAKLA: Ok, true about the checking that it would bring another complexity. Ok from me :+1:

ppanyukov:
> Nice! Thanks.
> 
> This makes sense to me, but can we move the script to `scripts`? (:

All done :)

PR #1609, opened 2019-10-04
Add a limit in the graph code for number of metrics

dgl:

## Changes

Improvements to handle large numbers of metric names in the graph UI.

This merges my upstream (but pending) changes in Prometheus Graph UI (https://github.com/prometheus/prometheus/pull/6100, discussed a bit in https://github.com/prometheus/prometheus/issues/5919 too).

## Verification

Tested query UI with 1 million metrics (via Geras) in Chrome.

### DIFF
--- a/pkg/ui/static/js/graph.js
+++ b/pkg/ui/static/js/graph.js
@@ -3,7 +3,14 @@ var graphTemplate;
 
 var INPUT_DEBOUNCE_WAIT = 500; // ms
 var SECOND = 1000;
+// Do not show the metrics selector if there are more than this many metrics.
+const METRIC_SELECTOR_LIMIT = 5000;
+// If there are more than this many metrics do not do fuzzy matching.
+const FUZZY_MATCHING_LIMIT = 10000;
 
+/**
+ * Graph
+*/
 Prometheus.Graph = function(element, options, handleChange, handleRemove) {
   this.el = element;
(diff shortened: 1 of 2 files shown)


pkg/ui/static/js/graph.js
```diff
@@ -332,29 +344,44 @@ Prometheus.Graph.prototype.populateInsertableMetrics = function() {
   });
 };
 
+Prometheus.Graph.prototype.makeFuzzyResult = function(self, query, source) {
+  // If fuzzyResult has already been calculated for this query, don't calculate again
```
bwplotka: missing trailing period (:

dgl: Fixed.

GiedriusS:
The code looks good, need to try out on my cluster to see if it helps - I wonder if just the parsing of the JSON won't crash my browser.

GiedriusS:
On an environment with ~1.5 million metrics, this definitely fixes the problem (tested it out) just that there is an interesting artifact: the dropdown disappears after a small hiccup when the JSON finishes downloading and the parsing begins :smile: 

Lets see what Prometheus's maintainers will say.

GiedriusS:
Funny how they are talking about it but still no one reviewed your PR on the Prometheus side :crying_cat_face:

PR #1618, opened 2019-10-09
[QUERY] Add metric for status of stores for query

OGKevin:
Signed-off-by: Kevin Hellemun <17928966+OGKevin@users.noreply.github.com>

## Changes

1. Add a new metric `thanos_store_status` for query component that indicates if a store is up or down for the query.

## Verification

```
# HELP thanos_store_status The status of stores as indicated on the /stores page. 1 is UP 0 is DOWN
# TYPE thanos_store_status gauge
thanos_store_status{labelset="",labelset_hash="da39a3ee5e6b4b0d3255bfef95601890afd80709"} 0
```
Screenshot

### DIFF
--- a/pkg/query/storeset.go
+++ b/pkg/query/storeset.go
@@ -456,7 +456,7 @@ func (s *StoreSet) getHealthyStores(ctx context.Context, stores map[string]*stor
 	return healthyStores
 }
 
-func (s *StoreSet) updateStoreStatus(store *storeRef, err error) {
+func (s *StoreSet) updateStoreStatus(store *storeRef, storeUnhealthyReasonError error) {
 	s.storesStatusesMtx.Lock()
 	defer s.storesStatusesMtx.Unlock()
 
@@ -466,14 +466,19 @@ func (s *StoreSet) updateStoreStatus(store *storeRef, err error) {
 		status = *prev
 	}
 
-	status.LastError = err
+	storeLabelsSetHash, err := storepb.LabelSetsToHash(store.labelSets)
+	if err != nil {
+		level.Error(s.logger).Log("msg", "create hash from label sets failed", "err", err)
+	}
+
+	status.LastError = storeUnhealthyReasonError
 	status.LastCheck = time.Now()
 	s.storeStatus.WithLabelValues(
 		storepb.LabelSetsToString(store.labelSets),
-		storepb.LabelSetsToHash(store.labelSets),
+		storeLabelsSetHash,
 	).Set(0)
 
-	if err == nil {
+	if storeUnhealthyReasonError == nil {
 
 		mint, maxt := store.TimeRange()
 		status.LabelSets = store.LabelSets()
@@ -482,7 +487,7 @@ func (s *StoreSet) updateStoreStatus(store *storeRef, err error) {
 		status.MaxTime = maxt
 		s.storeStatus.WithLabelValues(
 			storepb.LabelSetsToString(store.labelSets),
-			storepb.LabelSetsToHash(store.labelSets),
+			storeLabelsSetHash,
 		).Set(1)
 	}
 
--- a/pkg/query/storeset.go
+++ b/pkg/query/storeset.go
@@ -164,7 +164,9 @@ type StoreSet struct {
 	storesMetric *storeSetNodeCollector
 
 	// Map of statuses used only by UI.
-	storeStatuses         map[string]*StoreStatus
+	storeStatuses map[string]*StoreStatus
+	storeStatus   *prometheus.GaugeVec
+
 	unhealthyStoreTimeout time.Duration
 }
 
@@ -177,8 +179,17 @@ func NewStoreSet(
 	unhealthyStoreTimeout time.Duration,
 ) *StoreSet {
 	storesMetric := newStoreSetNodeCollector()
+	storeStatus := prometheus.NewGaugeVec(prometheus.GaugeOpts{
+		Name: "thanos_store_status",
+		Help: "The status of stores as indicated on the /stores page. 1 is UP 0 is DOWN",
+	},
+		[]string{
+			"address",
+		},
+	)
 	if reg != nil {
 		reg.MustRegister(storesMetric)
+		reg.MustRegister(storeStatus)
 	}
 
 	if logger == nil {
@@ -193,6 +204,7 @@ func NewStoreSet(
 		storeSpecs:            storeSpecs,
 		dialOpts:              dialOpts,
 		storesMetric:          storesMetric,
+		storeStatus:           storeStatus,
 		gRPCInfoCallTimeout:   10 * time.Second,
 		stores:                make(map[string]*storeRef),
 		storeStatuses:         make(map[string]*StoreStatus),
@@ -455,6 +467,7 @@ func (s *StoreSet) updateStoreStatus(store *storeRef, err error) {
 
 	status.LastError = err
 	status.LastCheck = time.Now()
+	s.storeStatus.WithLabelValues(store.addr).Set(0)
 
 	if err == nil {
 
@@ -463,6 +476,7 @@ func (s *StoreSet) updateStoreStatus(store *storeRef, err error) {
 		status.StoreType = store.StoreType()
 		status.MinTime = mint
 		status.MaxTime = maxt
+		s.storeStatus.WithLabelValues(store.addr).Set(1)
 	}
 
 	s.storeStatuses[store.addr] = &status
--- a/pkg/query/storeset.go
+++ b/pkg/query/storeset.go
@@ -184,7 +184,8 @@ func NewStoreSet(
 		Help: "The status of stores as indicated on the /stores page. 1 is UP 0 is DOWN",
 	},
 		[]string{
-			"address",
+			"labelset",
+			"labelset_hash",
 		},
 	)
 	if reg != nil {
@@ -467,7 +468,10 @@ func (s *StoreSet) updateStoreStatus(store *storeRef, err error) {
 
 	status.LastError = err
 	status.LastCheck = time.Now()
-	s.storeStatus.WithLabelValues(store.addr).Set(0)
+	s.storeStatus.WithLabelValues(
+		storepb.LabelSetsToString(store.labelSets),
+		storepb.LabelSetsToHash(store.labelSets),
+	).Set(0)
 
 	if err == nil {
 
@@ -476,7 +480,10 @@ func (s *StoreSet) updateStoreStatus(store *storeRef, err error) {
 		status.StoreType = store.StoreType()
 		status.MinTime = mint
 		status.MaxTime = maxt
-		s.storeStatus.WithLabelValues(store.addr).Set(1)
+		s.storeStatus.WithLabelValues(
+			storepb.LabelSetsToString(store.labelSets),
+			storepb.LabelSetsToHash(store.labelSets),
+		).Set(1)
 	}
 
 	s.storeStatuses[store.addr] = &status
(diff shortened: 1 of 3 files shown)


pkg/query/storeset.go
```diff
@@ -375,12 +386,14 @@ func (s *StoreSet) updateStoreStatus(store *storeRef, err error) {
 
 	status.LastError = err
 	status.LastCheck = time.Now()
+	s.storeStatus.WithLabelValues(store.addr).Set(0)
```
bwplotka: IMO I would update this metric in `func (s *StoreSet) Update(ctx context.Context) {` - it would be much more explicit.

pkg/query/storeset.go
```diff
@@ -128,12 +129,21 @@ func NewStoreSet(
 		Name: "thanos_store_nodes_grpc_connections",
 		Help: "Number indicating current number of gRPC connection to store nodes. This indicates also to how many stores query node have access to.",
 	})
+	storeStatus := prometheus.NewGaugeVec(prometheus.GaugeOpts{
+		Name: "thanos_store_status",
+		Help: "The status of stores as indicated on the /stores page. 1 is UP 0 is DOWN",
+	},
+		[]string{
+			"address",
```
bwplotka: That's the main problem we want to avoid. Putting potentially dynamic addresses to the metrics which increases cardinality.

What's wrong again with:
```
ar nodeInfoDesc = prometheus.NewDesc(
	"thanos_store_node_info",
	"Number of nodes with the same external labels identified by their hash. If any time-series is larger than 1, external label uniqueness is not true",
	[]string{"external_labels"}, nil,
)
```
It has exactly the same logic you have given here just encoded bit vaguely in collector logic - it might be not clear. Anyway in my opinion the problem must be somewhere else.. Maybe the cluster was down for a short time?

However, I am happy with you metric that will **replace** `thanos_store_node_info` as we can improve name and help. It's not only meta info - it also gives you healthiness info. However I would stick to `external_labels` - I think even for your use case it's just enough to have external labels, right?

OGKevin: The problem with `thanos_store_node_info ` is:

1. its not a bool value, which means it becomes a little more complex to eval simply "can this query talk to a store with labels set X"
2. it records how many stores are available with the same label set
3. If a store becomes unhealthy, its removed from the list of healthy stores which means in the re-construction of the list, it won't be exposed in the metric as down which means it will never expose a metric with 0. 
```
externalLabelOccurrencesInStores := map[string]int{}
		externalLabelOccurrencesInStores[externalLabelsFromStore(st)]++
	}
```
4. In our setup we have 1 entry point per cluster e.g. `cluster-1.mb.io` and I just want to know if "Global Query" considers this particular store healthy or not.

OGKevin: > That's the main problem we want to avoid. Putting potentially dynamic addresses to the metrics which increases cardinality.

I don't see any other solution as the external label set is not always available in case if the first request results in an error. The address is the only constant we have.

pkg/query/storeset.go
OGKevin: In the Update method, there are a lot of blocks that calls the `updateStoreStatus` method. Setting the store status metric literally needs to happen every time `updateStoreStatus` is called in `Update`. 

So with this in mind, it made more sense to place this piece of code in the `updateStoreStatus` method as this metric literally represent the status that is being updated in this method. 

Placing it in the Update method also means that the value needs to be set manually and not via logic. E.g. 

```
s.updateStoreStatus(store, nil)
s.storeStatus.WithLabelValues(store.addr).Set(1)

s.updateStoreStatus(store, someError)
s.storeStatus.WithLabelValues(store.addr).Set(0)
```

Instead of based on logic:

```
	s.storeStatus.WithLabelValues(store.addr).Set(0)

	if err == nil {
		s.storeStatus.WithLabelValues(store.addr).Set(1)
	}

```

With this in mind, do you still think It needs to go in the Update method?

pkg/query/storeset.go
OGKevin: @bwplotka I've pushed using label set and its hash as labels.

bwplotka:
We updated the metrics in recent PR. Can you rebase and take a look? I
think we really would be fine with just `thanos_store_nodes_grpc_connections`.
`addr` is not really nice as label. Think about it as IP address - it
changes in most of the setups so it does not give you much info. It's even
hard to track service behind IP address.. I would strongly suggest avoiding
this, thus I am sceptical of this change. Happy to hear feedback from
others as well (:

OGKevin:
Rebased and pushed.

>  I think we really would be fine with just `thanos_store_nodes_grpc_connections`. `addr` is not really nice as label

I agree and that's why I've removed it as you suggested. It now uses the LabelSet and its hash. I think my argument on why `thanos_node_info ` can not be used is still strong and has not had a good counter yet.

`thanos_store_nodes_grpc_connections ` tells me to how many stores this query is talking to. It does not give any info on what these stores are. So, even if I do `thanos_store_nodes_grpc_connections != EXPECTED_NUMBER_OF_STORES` I would not know which store went down, nor if its it ALL in 1 cluster (e.g. im fine if 1 goes down but not all 5) etc etc. 

Thats why, still, I think the only nice solution to monitor this is with a literal representation off the `/store` page data in a metric based on label set.

bwplotka:
> thanos_store_nodes_grpc_connections != EXPECTED_NUMBER_OF_STORES I would not know which store is unreachable, nor if its it ALL in 1 cluster (e.g. im fine if 1 is unreachable but not all 5) etc etc.

This is now grouped per external labels and StoreAPI type, so you know exactly what cluster it is, right?  You can easily set up alerts for:
* Expected number of stores per labelset and type SIDECAR (e.g which shows you exactly how many per cluster are present) 
* If you use DNS discovery: expected DNS results from the discovery as well if needed.

This should be well enough for alerting right? You narrowed storeAPI to external labels and type. You can then go to StoreAPI or check logs for exact address (which might be an IP that changes all the time). So what's missing exactly? (:

OGKevin:
> This is now grouped per external labels and StoreAPI type, so you know exactly what cluster it is, right?

The code seems to have been changed recently to add this info, so now I could indeed know which cluster and store type it is.

> So what's missing exactly? (:
> thanos_store_nodes_grpc_connections != EXPECTED_NUMBER_OF_STORES 

EXPECTED_NUMBER_OF_STORES is a hardcoded value here that I rather do not want to use because we spin up/down stores as there are needed in a cluster depending on the load etc. And seeing that `thanos_store_node_info` is now deprecated and refers to itself in the deprecation message (not sure if this is intentional)

https://github.com/thanos-io/thanos/blob/5bfe54eae7952478e3759a84a51dda232b0c5512/pkg/query/storeset.go#L98-L102

there is no longer a way to know "How many stores with external_labels should be available for query". So that one could do

`(thanos_store_nodes_grpc_connections / thanos_store_node_info) * 100 < 5` as this would mean most of the stores are now unreachable for the query (with the correct aggregations).

So unless im missing something, I still don't see a way to implement this with the current metrics. And a simple bool metric at a POV from query would still be a nice and simple solution. 

We already have alerts if X% of stores or whatever are down. But in a scenario where everything is actually up and running, but let's say Nginx has been misconfigured, query can not talk to the stores as the http requests are misrouted. So I want to get notified if thanos query can not reach all StoreAPI's in cluster X, not being able to reach a few is fine.

OGKevin:
Actually, now that I've changed my MR to use label set rather then the IP address, my new metric will make no sense. 

Imagine that there are 2 stores with the same label set. One is down and one is up, my metric's value will be determined by the last store that gets updated. Unless all the label sets are unique which is most likely not the case. 

So, if the metric I've added can not use IP address due to high cardinality, then I think using labels will cause for unpredictable behaviour 🤔

bwplotka:
> Imagine that there are 2 stores with the same label set. One is down and one is up, my metric's value will be determined by the last store that gets updated. Unless all the label sets are unique which is most likely not the case.

Then you will have this metric for exactly this STORE and extlset = X set to 2 or 1 depending how many are "live" right? So you can assert/alert on this. Sure, to tell what exactly address is failing you need to dive deeper, but for alerting purposes it should be enough. Am I wrong here? And yes labels doesn't need to be unique for certain cases. One is Store GW time slicing, so if you have couple of store GWs for different time ranges.

> hen I think using labels will cause for unpredictable behaviour 

What's unpredictable again?

OGKevin:
> Then you will have this metric for exactly this STORE and extlset = X set to 2 or 1 depending how many are "live" right?

No, I'm. referring to the metric I just added in this MR. My bad if I was not clear. So the value will basically be True or False.

> What's unpredictable again?

Because this MR introduces a bool metric based on label set. The bool can be set too True, because the 1st is healthy, and then to False, because the second store is unhealthy. And to fix this it needs to be instrumented with the address as label.


So yea, due to the changes made to `thanos_store_nodes_grpc_connections`, it could work however, I can not find a replacement for `thanos_store_info` now that this one is deprecated to know how many stores with `ExtLset` _should_ be available. I don't want to hardcode the number.

So its either a bool metric, like the one I added, based on address OR a percentage like: 

> (thanos_store_nodes_grpc_connections / thanos_store_node_info) * 100 < 5

or more explicitly phrased:

`(thanos_stores_up_for_query / thanos_total_stores_that_should_be_available_for_query) * 100 < 5`

and ofc both metrics need to have the external labels instrumented as label in the metric. 

So, this MR can be closed if the metric added in this MR `thanos_store_status` is not allowed to be instrumented with address. And discussion would probably need to be continued elsewhere.

pkg/query/storeset.go
```diff
@@ -177,8 +179,18 @@ func NewStoreSet(
 	unhealthyStoreTimeout time.Duration,
 ) *StoreSet {
 	storesMetric := newStoreSetNodeCollector()
+	storeStatus := prometheus.NewGaugeVec(prometheus.GaugeOpts{
+		Name: "thanos_store_status",
+		Help: "The status of stores as indicated on the /stores page. 1 is UP 0 is DOWN",
+	},
+		[]string{
+			"labelset",
```
bwplotka: I think those two has exactly the same info so we need to choose one.

pkg/query/storeset.go
```diff
@@ -453,16 +466,29 @@ func (s *StoreSet) updateStoreStatus(store *storeRef, err error) {
...
 
-	if err == nil {
+	if storeUnhealthyReasonError == nil {
 
 		mint, maxt := store.TimeRange()
 		status.LabelSets = store.LabelSets()
 		status.StoreType = store.StoreType()
 		status.MinTime = mint
 		status.MaxTime = maxt
+		s.storeStatus.WithLabelValues(
```
bwplotka: Hm this is hiding the truth as there can be many store APIs with the same labelset. So it's not only 0 or 1. This status can mean status for many storeAPIs essentially. It has to be number gauge with + and - 

... and this is exactly what `thanos_store_nodes_grpc_connections` does right?

OGKevin: Yea that's what I tried to explain here: 

> Because this MR introduces a bool metric based on label set. The bool can be set too True, because the 1st is healthy, and then to False, because the second store is unhealthy. And to fix this it needs to be instrumented with the address as label.

OGKevin:
Closing, as address can not be used for instrumentation which makes the goal of the added metric useless.

OGKevin:
> I think that's a fair point. What about a different strategy? StoreAPI is unhealthy if we suddenly can't talk to it via gRPC. This will be stated by grpc_client_handled_total grpc_method=Info and error grpc_code so you could assert on this, right? (:

this metric is a "generic" metric and does not expose labels I need, such as the label set. So there is no way to match `thanos_store_nodes_grpc_connections` with `grpc_client_handled_total `.

Basically, next to  `thanos_store_nodes_grpc_connections` I need a metric like "thanos_store_nodes_registerd" or something that tells me, how many stores are "registerd" with this certain label set so I know how many of them or up or down. So I can make an alert based on percentage.

PR #1627, opened 2019-10-10
pkg/receive: fix regression and add testing for receive external API to avoid future regressions

squat:
This commit adds tests for the Thanos receive component's external HTTP
API. This rather large PR was motivated by fixing a regression that was
introduced in 19e59ef5369c2134ae3295bd29b86b05b82cb980, where returning
a MultiError threw off the error counting logic in the request
parallelization func, which could cause the wrong HTTP status code to be
returned. In order to protect against these regressions, we need
thorough testing of the external API.

cc @brancz @bwplotka

### DIFF
--- a/pkg/receive/handler.go
+++ b/pkg/receive/handler.go
@@ -29,6 +29,13 @@ import (
 	"github.com/thanos-io/thanos/pkg/tracing"
 )
 
+const (
+	// DefaultTenantHeader is the default header used to designate the tenant making a write request.
+	DefaultTenantHeader = "THANOS-TENANT"
+	// DefaultReplicaHeader is the default header used to designate the replica count of a write request.
+	DefaultReplicaHeader = "THANOS-REPLICA"
+)
+
 // conflictErr is returned whenever an operation fails due to any conflict-type error.
 var conflictErr = errors.New("conflict")
 
@@ -326,6 +333,15 @@ func (h *Handler) parallelizeRequests(ctx context.Context, tenant string, replic
 					err = errors.New("storage is not ready")
 				} else {
 					err = h.writer.Write(wreqs[endpoint])
+					// When a MultiError is added to another MultiError, the error slices are concatenated, not nested.
+					// To avoid breaking the counting logic, we need to flatten the error.
+					if errs, ok := err.(terrors.MultiError); ok {
+						if countCause(errs, isConflict) > 0 {
+							err = errors.Wrap(conflictErr, errs.Error())
+						} else {
+							err = errors.New(errs.Error())
+						}
+					}
 				}
 				h.mtx.RUnlock()
 				if err != nil {
--- a/pkg/receive/handler_test.go
+++ b/pkg/receive/handler_test.go
@@ -1,11 +1,19 @@
 package receive
 
 import (
+	"bytes"
 	"net/http"
+	"net/http/httptest"
 	"strconv"
+	"sync"
 	"testing"
 
+	"github.com/go-kit/kit/log"
+	"github.com/gogo/protobuf/proto"
+	"github.com/golang/snappy"
 	"github.com/pkg/errors"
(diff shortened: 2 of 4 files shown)


pkg/receive/handler_test.go
```diff
@@ -120,3 +128,346 @@ func TestCountCause(t *testing.T) {
...
+				&fakeAppendable{
+					appender: newFakeAppender(conflictErrFn, nil, commitErrFn, nil),
+				},
+				&fakeAppendable{
+					appender: newFakeAppender(conflictErrFn, nil, commitErrFn, nil),
+				},
+			},
+		},
+		{
+			name:              "size 3 with replication and one faulty",
```
brancz: I think I'd be worth it to add a few more cases where 2/3 appenders fail with an error.

squat: ack, added more tests for 1/3 faulty receivers and 2/3 faulty receivers with different mixes of errors, which indeed are some of the more interesting cases.

pkg/receive/handler.go
```diff
@@ -326,6 +333,18 @@ func (h *Handler) parallelizeRequests(ctx context.Context, tenant string, replic
 					err = errors.New("storage is not ready")
 				} else {
 					err = h.writer.Write(wreqs[endpoint])
+					// When a MultiError is added to another MultiError, the error slices are concatenated, not nested.
+					// To avoid breaking the counting logic, we need to flatten the error.
+					if errs, ok := err.(terrors.MultiError); ok {
+						switch {
+						case errs.Err() == nil:
```
bwplotka: do we really need that? empty MultiError is a nil:

```
/ Err returns the error list as an error or nil if it is empty.
func (es MultiError) Err() error {
	if len(es) == 0 {
		return nil
	}
	return es
}

```

bwplotka: And if we don't we probable can have just `if`

pkg/receive/handler.go
```diff
@@ -326,6 +333,18 @@ func (h *Handler) parallelizeRequests(ctx context.Context, tenant string, replic
 					err = errors.New("storage is not ready")
 				} else {
 					err = h.writer.Write(wreqs[endpoint])
+					// When a MultiError is added to another MultiError, the error slices are concatenated, not nested.
+					// To avoid breaking the counting logic, we need to flatten the error.
+					if errs, ok := err.(terrors.MultiError); ok {
```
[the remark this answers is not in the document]
squat: the commit message mentions:
```
fixing a regression that was
introduced in 19e59ef, where returning
a MultiError threw off the error counting logic in the request
parallelization func, which could cause the wrong HTTP status code to be
returned
```

pkg/receive/handler_test.go
```diff
@@ -120,3 +128,346 @@ func TestCountCause(t *testing.T) {
...
+			i = 0
+		}
+		return err
+	}
+}
+
+// makeRequest is a helper to make a correct request against a remote write endpoint given a request.
+func makeRequest(t *testing.T, h *Handler, tenant string, wreq *prompb.WriteRequest) int {
+	buf, err := proto.Marshal(wreq)
+	if err != nil {
```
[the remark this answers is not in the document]
squat: I prefer doing it this way, but I prefer to conform to the codebase if that is the pattern Thanos follows

pkg/receive/handler.go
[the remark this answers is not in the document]
squat: I can reword for clarity

pkg/receive/handler_test.go
brancz: Nice :+1:

pkg/receive/handler.go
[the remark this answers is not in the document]
bwplotka: :+1: thx for fixing.

Issue #1625, opened 2019-10-10
Querier OOM: introduce and manage memory limits


ppanyukov:
This is to address **one specific issue with Querier**. It is the first step on the much longer path to improve memory usage in both Store Gateway and Querier.

```
thanos, version 0.7.0 (branch: master, revision: ed045447fef8c4b5ff62697559463549ff522f31)
  build user:       philip@Mayas-MacBook-Pro.local
  build date:       20191010-13:52:26
  go version:       go1.13.1
```

Specific issue:

- Querier allocates **unbounded** amounts of memory during query execution.
- This can and does lead to OOM in resource-limited container environments (k8s etc).
- The issue prevents use of the product as it's just too unstable.

Proposed first-step fix:

- Introduce memory limits:
    - Overall max mem for all queries.
    - Individual max mem for individual queries.
- Actively manage mem accounting during query execution.
- Fail the query when we know the memory limits will be breached.

Initial introduction to product:

- Opt-in for those who want it (at first).

Effects of the fix:

- Benefit: No more OOM of central Querier caused by any one query issued by one user.
- Possible Cost: Potentially some queries which succeed now may start failing due to introduced caps.

------

# Root cause

- Unbounded memory allocations in one place:
- File: `github.com/thanos-io/thanos/pkg/query/querier.go`
- Function:
    ```
    func (s *seriesServer) Send(r *storepb.SeriesResponse) error {
    }
    ```
- This function is called in a loop.

- Instrumetation shows these typical memory stats per each 1000 calls:

```
    seriesServer.Send
      HeapAlloc  : 3.97G, delta: 1.16G
      HeapObjects: 32.15M, delta: 4.62M
      > call count       - 6.84K
      > len(s.seriesSet) - 6.84K

    seriesServer.Send
      HeapAlloc  : 5.13G, delta: 1.16G
      HeapObjects: 36.77M, delta: 4.62M
      > call count       - 7.81K
      > len(s.seriesSet) - 7.81K
```

- The array `s.seriesSet` grows by `~1GB` and `4.62M` objects with each 1000 calls.

- Introducing memory accounting in this function may be a good place for the fix.

# Repro

- **Data**
    - synthetically generated
    - data size: 
        - metrics: 200
        - targets: 100
        - sample interval: 15s
        - total time interval: 10 hours

- **Query**
    - `count({__name__=~".+"}) by (__name__)`
    - time span: all generated data

- **Memory profile**
    - SG: 20GB peak
[...]

PR #1631, opened 2019-10-11
Memory limiter in Querier

ppanyukov:
**DRAFT - NOT FOR MERGE**

This PR is a first demo of the fix for #1625.

It is here for:

- Demonstrate the nature of the fix;
- Discuss and agree if this is how we broadly want to implement it. Or some other way.

Note:
- The code for the fix can probably be in better place;
- There are some hard-coded values and TODOs for now.

## Changes

Small changes to `github.com/thanos-io/thanos/pkg/store/proxy.go`. 

## Verification

Using manual test. Querier logs show this:

```
level=error ts=2019-10-11T10:31:34.215101Z caller=proxy.go:203 err="OOM preventor: exceeded max memory. Max: 2048.00M; Current: 2048.23M"
level=error ts=2019-10-11T10:31:34.215354Z caller=engine.go:596 msg="error selecting series set" err="proxy Series(): OOM preventor: exceeded max memory. Max: 2048.00M; Current: 2048.23M"
level=warn ts=2019-10-11T10:31:34.215528Z caller=proxy.go:447 err="failed to receive any data from Addr: localhost:10901 LabelSets: [] Mint: 1569762000000 Maxt: 1569798000001: context canceled" msg="returning partial response"
```

### DIFF
--- a/pkg/store/proxy.go
+++ b/pkg/store/proxy.go
@@ -5,6 +5,7 @@ import (
 	"fmt"
 	"io"
 	"math"
+	"runtime"
 	"sort"
 	"strings"
 	"sync"
@@ -181,6 +182,31 @@ func (s ctxRespSender) send(r *storepb.SeriesResponse) {
 // Series returns all series for a requested time range and label matcher. Requested series are taken from other
 // stores and proxied to RPC client. NOTE: Resulted data are not trimmed exactly to min and max time range.
 func (s *ProxyStore) Series(r *storepb.SeriesRequest, srv storepb.Store_SeriesServer) error {
+	// TODO(ppanyukov): hardcoded mem limit for now, where is best place to plug config for this?
+	const maxMem = uint64(2 * 1024 * 1024 * 1024) // 2GB
+
+	// checkMem returns an error if the current memory allocations exceeded max.
+	checkMem := func() error {
+		var nowMem uint64
+		{
+			// TODO(ppanyukov): ReadMemStats does stopTheWorld. Is this acceptable?
+			currentM := &runtime.MemStats{}
+			runtime.ReadMemStats(currentM)
+			nowMem = currentM.HeapAlloc
+		}
+
+		if nowMem >= maxMem {
+			// format in MB.
+			maxMemS := fmt.Sprintf("%.2fM", float64(maxMem)/1024.0/1024.0)
+			nowMemS := fmt.Sprintf("%.2fM", float64(nowMem)/1024.0/1024.0)
+			err := fmt.Errorf("OOM preventor: exceeded max memory. Max: %s; Current: %s", maxMemS, nowMemS)
+			level.Error(s.logger).Log("err", err)
+			return err
+		}
+
+		return nil
+	}
+
 	match, newMatchers, err := matchesExternalLabels(r.Matchers, s.selectorLabels)
 	if err != nil {
 		return status.Error(codes.InvalidArgument, err.Error())
@@ -280,9 +306,21 @@ func (s *ProxyStore) Series(r *storepb.SeriesRequest, srv storepb.Store_SeriesSe
 		return mergedSet.Err()
 	})
 
+	// All code above is fast and light on memory.
+	// The real work is in the loop that follows.
+	// Before we do anything expense, check we should even start.
+	if err := checkMem(); err != nil {
+		return err
+	}
+
 	for resp := range respRecv {
 		if err := srv.Send(resp); err != nil {
 			return status.Error(codes.Unknown, errors.Wrap(err, "send series response").Error())
+		}
+
+		// bail if exceeded mem.
+		if err := checkMem(); err != nil {
+			return err
 		}
 	}
 
@@ -479,6 +517,8 @@ func labelSetMatches(ls storepb.LabelSet, matchers []storepb.LabelMatcher) (bool
 	}
 	return true, nil
 }
+
+// TODO:
 
 // LabelNames returns all known label names.
 func (s *ProxyStore) LabelNames(ctx context.Context, r *storepb.LabelNamesRequest) (


pkg/store/proxy.go
```diff
@@ -280,10 +306,22 @@ func (s *ProxyStore) Series(r *storepb.SeriesRequest, srv storepb.Store_SeriesSe
...
+		return err
+	}
+
 	for resp := range respRecv {
 		if err := srv.Send(resp); err != nil {
 			return status.Error(codes.Unknown, errors.Wrap(err, "send series response").Error())
 		}
+
+		// bail if exceeded mem.
+		if err := checkMem(); err != nil {
```
bwplotka: I think we might to that on receive rather.. just thinking loudly

ppanyukov:
@bwplotka 

> ...This will indeed potentially keep Querier running by failing ALL the requests pending in Querier.

Good point. There are really many flaws with this naive simple approach really. The only aim here is to prevent OOM as step1 and then see what else we can do to make things better.

Ideas how to take this further:

- Some kind of logic where we kill one query in the hope this frees up room for others. For example, if memcheck detect nomem condition, it returns error to the first caller. Then waits a bit, runs GC, while blocking other callers. All this in the hope that the query that was killed freed up enough memory for others to proceed. This probably relatively easy to do.

- Per-query accounting. Would be nice. But I think it's quite difficult to do without very impactful changes. At least I don't have a clear picture of how this can be done at the moment without more digging around. The only idea I have is some kind of custom arena memory manager (basically massive byte slice) which gives out sub-slices and through which everything goes.



> BTW, how is that different from OOM? ((: 

OOM is a bug, and this is a feature! :}} 

I see your point though. The difference is:

- OOM gives a really bad impression to the users; Here at least we try to do a bit better.
- OOM is not always easy to see -- stuff just dies. Depending on how/where you run it, need to look at different places to check if it's OOM or something else. Here we have log message.
- Other things may depend on the Querier. Do they get "connection reset by peer" when it OOMs? Here they won't.
- Restart may be quick, but it still takes time. Also, in k8s it will be another pod, which will do +1 for cardinality of metrics. Here we avoid it.

Not arguing much with what you pointed out, this fix is a stop-gap safety valve feature really, not much more at this point.

pkg/store/proxy.go
```diff
@@ -181,6 +182,31 @@ func (s ctxRespSender) send(r *storepb.SeriesResponse) {
...
 // stores and proxied to RPC client. NOTE: Resulted data are not trimmed exactly to min and max time range.
 func (s *ProxyStore) Series(r *storepb.SeriesRequest, srv storepb.Store_SeriesServer) error {
+	// TODO(ppanyukov): hardcoded mem limit for now, where is best place to plug config for this?
+	const maxMem = uint64(2 * 1024 * 1024 * 1024) // 2GB
+
+	// checkMem returns an error if the current memory allocations exceeded max.
+	checkMem := func() error {
+		var nowMem uint64
+		{
+			// TODO(ppanyukov): ReadMemStats does stopTheWorld. Is this acceptable?
```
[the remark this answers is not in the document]
ppanyukov: I'm not sure there is a way to avoid it? Unless maybe we ask OS instead? Or we use completely our own allocator. Or the queries can tell us exactly how much memory they are consuming and freeing?

pkg/store/proxy.go
ppanyukov: is this the wrong place to return an error?

pkg/store/proxy.go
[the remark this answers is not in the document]
bwplotka: qeuries?

pkg/store/proxy.go
bwplotka: Return error, sure, but checking - it sounds more natural to check it when you receive the data

pkg/store/proxy.go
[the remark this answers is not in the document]
ppanyukov: Yeah, if we have each query itself know know exactly how much memory they need, and tell some central accounting service when add acquire and release memory, we could do it.

pkg/store/proxy.go
ppanyukov: Where would that be? Sorry I'm not as familiar with codebase yet :) 

Do you mean inside `srv.Send(resp)` perhaps? Does it even matter? To me this does look like receiver of sorts, since this loop reads out from `respRecv` channel and the var is called response?

bwplotka:
It looks to me that we are focusing on the opposite direction of the value. We are implementing some blackbox OOM killer, instead of trying to understand the data and what's happening, how much data we are actually fetching vs allocating, where we can stream more. But that's only my opinion. I think I am happy for you to experiment with this as long as it does not introduce a huge maintenance burden/problem over time for Thanos project (:  f you/users feel it's valuable, we can definitely try it out and check how it goes, add more logic to this blackbox solution.

> Per-query accounting. Would be nice. But I think it's quite difficult to do without very impactful changes. At least I don't have a clear picture of how this can be done at the moment without more digging around. The only idea I have is some kind of custom arena memory manager (basically massive byte slice) which gives out sub-slices and through which everything goes.

Yup, that is called `heap` (: There is sync Pool that tries to help if you have similar sizes of the byte slices. There is also https://github.com/glycerine/offheap but it might be too complex FWTW. There is also this: https://github.com/VictoriaMetrics/fastcache

ppanyukov:
I agree there is a lot that can be done in this space. Hopefully issue #1625 is just the first step as it does at least highlights that all responses from SG look to be copied into internal array before given out to the client (unless I forgot to mention it there?). 

Maybe we should build a list of things to explore and think about, along these lines:

- why does it not stream out the results? is it something to do with deduplication perhaps? aggregation?
- what else can be done to optimise this?
- can we do something to prevent knowingly bad queries?
- can SG help make queries better by taking doing slightly more work?
- etc

Do people think this would be the right approach to take?

povilasv:
I agree with @bwplotka here I think we are going in the opposite direction here with this memory limiter. I would rather we have some kind of solution in a style of https://landing.google.com/sre/sre-book/chapters/addressing-cascading-failures/#xref_cascading-failure_preventing-server-overload

Maybe adding a limited queue instead of processing all requests, giving requests with larger time windows bigger cost's etc?

I would also look at what Cortex people did to prevent overload, maybe we can reuse some of their stuff.

GiedriusS:
I concur with others. It doesn't seem to me like stopping the world and calculating the allocations is a good way to go forward with this. Besides the comments about streaming more and so on, I had a small idea: we should handle panics in the separate goroutines which handle requests and disable overcommit to prevent the Linux kernel from allocating memory opportunistically. Then we would fail huge requests early. WDYT?

ppanyukov:
So, if we could put limits on individual queries that would be better, right? :) 

Had a super quick look at Cortex, here is the config right there in the querier. Looks like they have limits on max concurrent queries and max samples. 

```
// Config contains the configuration require to create a querier
type Config struct {
	MaxConcurrent            int
	Timeout                  time.Duration
	MaxSamples               int
	IngesterMaxQueryLookback time.Duration
....
}
```

Am I even looking in the right place? 

https://github.com/cortexproject/cortex/blob/20d3af237bff4db91ca20d850ab5413faf1949d7/pkg/querier/querier.go#L20

bwplotka:
> Looks like they have limits on max concurrent queries and max samples.

Yup, but that's not enough. Store GW had this and it was not enough. I think on top of this there might be max `series` which could be interesting. Cardinality is the largest source of memory use. Also, keep in mind that users start to use frontend query caching. This changes a characteristic of traffic against Querier a lot, as it splits queries by day - maybe we can take that as an advantage?

ppanyukov:
> I had a small idea: we should handle panics in the separate goroutines which handle requests and disable overcommit to prevent the Linux kernel from allocating memory opportunistically. Then we would fail huge requests early. WDYT?

@GiedriusS, could you elaborate please?

ppanyukov:
I will close this as it's clearly not the way to go. Instead I've opened a new discussion issue to have a central list the ideas and talk through them in one place.  See https://github.com/thanos-io/thanos/issues/1649

If there are better ways to do it rather than GitHub issues, please let me know, I'm more than happy to switch to that.

Issue #443, opened 2018-07-24
Migrate objstore package to use new Go cloud


cstyan:
The Go team announce the Go cloud packages today: https://blog.golang.org/go-cloud

To reduce the amount of internal code in objstore it could make use of the `go-cloud/blob`, or if we don't need `objstore/inmem` then objstore could be replaced entirely.

Curious what you guys think @Bplotka @fabxc, if this would be useful I'd like to work on it.

Bplotka:
Thanks for willing to help!

I would be very very careful about this. IMO:
- We aim for stability right now, so moving to package that is online only 1d might be not the greatest idea. 
- We need to first clearly enumerate advantages and disadvantages of moving to it.

For example:
> To reduce the amount of internal code in objstore

...might be not true at the end. There is risk that we will end up writing even more LOC to satisfy our current interface.

To sum up, I think we should definitely wait for this package to be proven in wild (and prod) by users that are willing to experiment more. What do you think? (:

BTW there are lots of way more important issues we could use your help! (: Feel free to look on any issues with prefix label `need help`

cstyan:
> We aim for stability right now, so moving to package that is online only 1d might be not the greatest idea.

Agreed! They even say it's not 100% stable and shouldn't be used in production yet. 

> There is risk that we will end up writing even more LOC to satisfy our current interface.

That would be disappointing. I think I'll give it a shot just to see what the current version of library would provide and what the code would look like. No intention of merging though.

>BTW there are lots of way more important issues we could use your help! (: Feel free to look on any issues with prefix label need help
Will do :)

Bplotka:
Awesome, thanks! Closing this for now.

bwplotka:
I think it's slowly time to reevaluate this (:

cstyan:
:+1: I can look into it, but not for a few days at least

bwplotka:
That would be interesting experiment. 

The main acceptance criteria are:
- Having proposal on how to migrate to blob interface - what changes vs our Bucket interface that would require.

The reasons why we potentially would do it is basically this:
- Adding more providers that would fit into blob interface would allow to have them more useful (and maintained) by anyone who use well known go cloud blob interface. That would ensure Thanos can focus on main goals. They could even migrate into their own repositories.

domgreen:
I think this is a great idea of getting a proposal document of what we would need to do to transition over.

[...]

PR #1648, opened 2019-10-14
Added CI testing against S3 bucket.

bwplotka:
* Again, only for master/release builds and PRs from person with write-access to Thanos.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -26,7 +26,7 @@ jobs:
           command: |
             if ! [ -z ${GCP_PROJECT} ]; then
               echo $GOOGLE_APPLICATION_CREDENTIALS_CONTENT > $GOOGLE_APPLICATION_CREDENTIALS
-              echo "Awesome! GCS integration tests are enabled."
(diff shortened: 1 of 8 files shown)


brancz:
Looks like we've got some failures. Happy to review once resolved.

bwplotka:
Yea, the reason is that `minio-go` client does not work if you don't pass  `os.File` as `io.Reader` which is really bad. We touched this problem at some point ago, we should either:

* Be clear we require the size of upload before op: add `ReaderSizer` interface or something. Now it is super hidden. In fact we always upload files, we never stream without knowing size, so that would do... This will block though any attempts to do some streaming more in future.
* Fix minio-go to work without file size (exploring)
* Just burn all of it and use AWS SDK or better.. go-cloud! (wonder if go cloud S3 driver works for other S3-compatible backends)

bwplotka:
Resolved it quickly.

Let's discuss moving to go-cloud here: https://github.com/thanos-io/thanos/issues/443.

PR #1654, opened 2019-10-15
pkg/receive: remove flushed WAL

squat:
This commit ensures that we delete the WAL after it has been flushed to
a block. Flushing the WAL simply creates a block but does not remove the
WAL directory or its contents. This means that once the DB is re-opened,
new samples are added to the same WAL. Flushing the WAL again does not
result in blocks with overlapping time ranges because the flushing logic
guards against this
(https://github.com/prometheus/prometheus/blob/master/tsdb/db.go#L300).
Nevertheless, we should delete the WAL after flushing it to ensure that
flushed samples are not needlessly re-processed. Also, once multi-TSDB
support is added, holding old samples in the WAL could cause problems.

## Verification
Ran thanos receive locally and ensured that after several starts and stops, blocks are created but the WAL is empty.

cc @bwplotka @brancz @krasi-georgiev

### DIFF
--- a/pkg/receive/tsdb.go
+++ b/pkg/receive/tsdb.go
@@ -1,6 +1,8 @@
 package receive
 
 import (
+	"os"
+	"path/filepath"
 	"sync"
 
 	"github.com/go-kit/kit/log"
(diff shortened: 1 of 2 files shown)


brancz:
I feel like it would be nice to have a test verifying that this is not putting us at risk of data loss.

squat:
@brancz ack, added test to ensure that opening a db, adding samples, flushing, and the querying returns the same samples.

brancz:
Very nice! :+1:

metalmatze:
Could we release v8.0.2 to ship this? :thinking:

Issue #1655, opened 2019-10-16
Store Gateway increased startup/baseline memory usage from v0.7.0 to v0.8.1


bvis:
Ping: @bwplotka

**Thanos, Prometheus and Golang version used**:

Thanos: quay.io/thanos/thanos:v0.8.1
Prometheus: prom/prometheus:v2.13.0

**Object Storage Provider**:
```
$ cat /config/thanos.yaml
type: GCS
...
```

**What happened**:
I've updated all the Thanos stack I'm using, from the v0.7.0 to v0.8.1. Everything works as expected (I've seen incredible memory savings in Thanos Sidecar, amazing job!) except the Store Gateway component.

It simply does not start and dies with a `2` exit code.

**What you expected to happen**:
It starts as before!

**How to reproduce it**:
It's happening in our productive environment, our staging stack, which uses its own bucket,  works correctly with the same parameters. 

Parameters:

```store
      --data-dir=/data
      --log.level=debug
      --index-cache-size=500MB
      --chunk-pool-size=2GB
      --grpc-server-tls-cert=/etc/pki/thanos/tls.crt
      --grpc-server-tls-client-ca=/etc/pki/thanos/tls.ca
      --grpc-server-tls-key=/etc/pki/thanos/tls.key
      --objstore.config-file=/config/thanos.yaml
```
**Full logs to relevant components**:
Excerpt of the pod describe:
```
Last State:     Terminated
      Reason:       Error
      Exit Code:    2
      Started:      Tue, 15 Oct 2019 19:12:56 +0200
      Finished:     Tue, 15 Oct 2019 19:23:23 +0200
```

<details>Logs of v0.8.1
<p>

```
level=debug ts=2019-10-16T08:00:41.916594638Z caller=main.go:120 msg="maxprocs: Leaving GOMAXPROCS=[8]: CPU quota undefined"
level=info  ts=2019-10-16T08:00:41.924413863Z caller=main.go:168 msg="Tracing will be disabled"
level=info  ts=2019-10-16T08:00:41.924630438Z caller=factory.go:39 msg="loading bucket configuration"
level=info  ts=2019-10-16T08:00:41.929235164Z caller=cache.go:172 msg="created index cache" maxItemSizeBytes=262144000 maxSizeBytes=524288000 maxItems=math.MaxInt64
level=debug ts=2019-10-16T08:00:41.981473465Z caller=store.go:162 msg="initializing bucket store"
```

</p>
</details>

[...]

PR #1656, opened 2019-10-16
store: Start metric and status probe HTTP server as earlier as possible

### LINKED ISSUE #1653: store gateway: Start metric server from the very start of the proces.
Store Gateway exposes its /metric page only after initial sync. We should do it sooner. Since we have readiness probe now we can do it properly earlier!

We need to fix it as part of https://github.com/thanos-io/thanos/issues/1471 anyway, so this needs to be done ASAP.

kakkoyun:
As a result of a couple of issues that we had by running Thanos Store with liveness probes on Kubernetes, this PR attempts to fix the liveness probe issues by moving them earlier in the start-up sequence. Now, metrics and status probe HTTP server starts as earlier as possible and returns success from `/-/healthy` endpoint and sets its status ready when everything else properly bootstrapped.

A good read on the subject: https://srcco.de/posts/kubernetes-liveness-probes-are-dangerous.html

## Changes

* starts metric and status probe HTTP server for Store
* attempts to fix #1653 

## Verification

* `make local-test`
* `MINIO_ENABLED=1 ./scripts/quickstart.sh` while debug log enabled, and observed start sequence.

Thanos Store start sequence logs:
```text
level=info name=store ts=2019-10-17T12:52:19.814229Z caller=main.go:170 msg="Tracing will be disabled"
level=info name=store ts=2019-10-17T12:52:19.814515Z caller=factory.go:39 msg="loading bucket configuration"
level=info name=store ts=2019-10-17T12:52:19.814916Z caller=cache.go:172 msg="created index cache" maxItemSizeBytes=131072000 maxSizeBytes=262144000 maxItems=math.MaxInt64
level=info name=store ts=2019-10-17T12:52:19.815104Z caller=main.go:257 msg="disabled TLS, key and cert must be set to enable"
level=info name=store ts=2019-10-17T12:52:19.815301Z caller=store.go:252 msg="starting store node"
level=info name=store ts=2019-10-17T12:52:19.815336Z caller=store.go:198 msg="initializing bucket store"
level=info name=store ts=2019-10-17T12:52:19.815359Z caller=main.go:353 msg="listening for requests and metrics" component=store address=0.0.0.0:10906
level=info name=store ts=2019-10-17T12:52:19.815421Z caller=prober.go:143 msg="changing probe status" status=healthy
level=info name=store ts=2019-10-17T12:52:19.818189Z caller=store.go:203 msg="bucket store ready" init_duration=3.131623ms
level=info name=store ts=2019-10-17T12:52:19.818247Z caller=store.go:245 msg="listening for StoreAPI gRPC" address=0.0.0.0:10905
level=info name=store ts=2019-10-17T12:52:19.818287Z caller=prober.go:114 msg="changing probe status" status=ready
```

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -17,7 +17,7 @@ We use *breaking* word for marking changes that are not backward compatible (rel
 
 ### Fixed
 
-- [#1656](https://github.com/thanos-io/thanos/pull/1656) Thanos Store now starts metric and status probe HTTP server earlier in its start-up sequence. `/-/healthy` now endpoint starts to respond with success earlier. `/metrics` endpoint now starts serving earlier as well, as a result for your readiness probes you should rely on  `/-/ready` rather than `/metrics`.
+- [#1656](https://github.com/thanos-io/thanos/pull/1656) Thanos Store now starts metric and status probe HTTP server earlier in its start-up sequence. `/-/healthy` endpoint now starts to respond with success earlier. `/metrics` endpoint starts serving metrics earlier as well. Make sure to point your readiness probes to the `/-/ready` endpoint rather than `/metrics`.
 
 ## [v0.8.1](https://github.com/thanos-io/thanos/releases/tag/v0.8.1) - 2019.10.14
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -17,7 +17,7 @@ We use *breaking* word for marking changes that are not backward compatible (rel
 
 ### Fixed
 
-- [#1656](https://github.com/thanos-io/thanos/pull/1656) Thanos components now starts metric and status probe HTTP server earlier in their start-up sequence.
+- [#1656](https://github.com/thanos-io/thanos/pull/1656) Thanos Store now starts metric and status probe HTTP server earlier in its start-up sequence. `/-/healthy` now endpoint starts to respond with success earlier. `/metrics` endpoint now starts serving earlier as well, as a result for your readiness probes you should rely on  `/-/ready` rather than `/metrics`.
 
 ## [v0.8.1](https://github.com/thanos-io/thanos/releases/tag/v0.8.1) - 2019.10.14
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -15,6 +15,10 @@ We use *breaking* word for marking changes that are not backward compatible (rel
 
 - [#1660](https://github.com/thanos-io/thanos/pull/1660) Add a new `--prometheus.ready_timeout` CLI option to the sidecar to set how long to wait until Prometheus starts up.
 
+### Fixed
+
+- [#1656](https://github.com/thanos-io/thanos/pull/1656) Thanos components now starts metric and status probe HTTP server earlier in their start-up sequence.
+
 ## [v0.8.1](https://github.com/thanos-io/thanos/releases/tag/v0.8.1) - 2019.10.14
 
 ### Fixed
@@ -23,12 +27,12 @@ We use *breaking* word for marking changes that are not backward compatible (rel
   * NOTE: `thanos_store_nodes_grpc_connections` metric is now per `external_labels` and `store_type`. It is a recommended  metric for Querier storeAPIs. `thanos_store_node_info` is marked as obsolete and will be removed in next release.
   * NOTE2: Store Gateway is now advertising artificial: `"@thanos_compatibility_store_type=store"` label. This is to have the current Store Gateway compatible with Querier pre v0.8.0.
   This label can be disabled by hidden `debug.advertise-compatibility-label=false` flag on Store Gateway.
- 
+
 ## [v0.8.0](https://github.com/thanos-io/thanos/releases/tag/v0.8.0) - 2019.10.10
 
 Lot's of improvements this release! Noteworthy items:
 - First Katacoda tutorial! 🐱
-- Fixed Deletion order causing Compactor to produce not needed 👻 blocks with missing random files. 
+- Fixed Deletion order causing Compactor to produce not needed 👻 blocks with missing random files.
 - Store GW memory improvements (more to come!).
 - Querier allows multiple deduplication labels.
 - Both Compactor and Store Gateway can be **sharded** within the same bucket using relabelling!
@@ -42,7 +46,7 @@ both Prometheus and sidecar with Thanos: https://prometheus.io/blog/2019/10/10/r
 
 - [#1619](https://github.com/thanos-io/thanos/pull/1619) Thanos sidecar allows to limit min time range for data it exposes from Prometheus.
 - [#1583](https://github.com/thanos-io/thanos/pull/1583) Thanos sharding:
-  - Add relabel config (`--selector.relabel-config-file` and `selector.relabel-config`) into Thanos Store and Compact components. 
+  - Add relabel config (`--selector.relabel-config-file` and `selector.relabel-config`) into Thanos Store and Compact components.
 Selecting blocks to serve depends on the result of block labels relabeling.
   - For store gateway, advertise labels from "approved" blocks.
 - [#1540](https://github.com/thanos-io/thanos/pull/1540) Thanos Downsample added `/-/ready` and `/-/healthy` endpoints.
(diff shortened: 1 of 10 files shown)


kakkoyun:
cc @bwplotka @brancz  @FUSAKLA @squat @metalmatze

squat:
Looking at the code changes, it seems to me that the only thing that has changed is the lexicographic ordering of the function calls but that the server is still started at exactly the same time. Note that the HTTP servers are assigned to the run group and that they only actually start serving requests when the run group itself is started.

kakkoyun:
@squat Thanks for pointing it. I somehow missed that. I'll have another hard look at it.

cmd/thanos/store.go
```diff
@@ -126,7 +126,12 @@ func runStore(
 	selectorRelabelConf *extflag.PathOrContent,
 	advertiseCompatibilityLabel bool,
 ) error {
+	level.Debug(logger).Log("msg", "setting up http server")
+	// Initiate HTTP listener providing metrics endpoint and readiness/liveness probes.
 	statusProber := prober.NewProber(component, logger, prometheus.WrapRegistererWithPrefix("thanos_", reg))
+	if err := scheduleHTTPServer(g, logger, reg, statusProber, httpBindAddr, nil, component); err != nil {
```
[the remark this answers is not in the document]
kakkoyun: I have updated `scheduleHTTPServer`, now it spawns a separate goroutine to start Server and syncs it with rungroup. Could you have a look at it now?

cmd/thanos/downsample.go
```diff
@@ -95,7 +95,13 @@ func runDownsample(
 	}()
 
 	metrics := newDownsampleMetrics(reg)
+	level.Debug(logger).Log("msg", "setting up http server")
```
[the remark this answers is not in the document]
bwplotka: Hmm, maybe (: What's the purpose of it?

cmd/thanos/main.go
```diff
@@ -349,13 +349,29 @@ func scheduleHTTPServer(g *run.Group, logger log.Logger, reg *prometheus.Registr
 		return errors.Wrap(err, "listen metrics address")
 	}
 
-	g.Add(func() error {
-		level.Info(logger).Log("msg", "listening for requests and metrics", "component", comp.String(), "address", httpBindAddr)
+	srv := http.Server{Handler: mux}
```
bwplotka: Hm, why this and not the opposite, which is taking `InitialSync` into `g.Add` group? Essentially this breaks convention - spins up extra go routine out of run group which is not easy to handle/cancel etc. 

What's the blocker of moving `InitialSync` instead? :thinking:

kakkoyun: Nothing, I believe. I just didn't understand what were you saying in your first comment, now I got it. I'll try to move `InitialSync` to the rungroup.

Maybe one argument could be, adding this change in here enables also the other components' probes. However, you are right it's better to stick with the conventions and avoid synchronization issues which could be introduced by this additional goroutine.

bwplotka:
Wonder if we can also add more logs to the startup as well, since you are touching this part as per: https://github.com/thanos-io/thanos/issues/1655

cmd/thanos/store.go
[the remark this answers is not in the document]
kakkoyun: Now `InitSync` is included in the rungroup.

cmd/thanos/main.go
```diff
@@ -355,6 +355,7 @@ func scheduleHTTPServer(g *run.Group, logger log.Logger, reg *prometheus.Registr
 		return errors.Wrapf(http.Serve(l, mux), "serve %s and metrics", comp.String())
 	}, func(err error) {
 		readinessProber.SetNotHealthy(err)
+		l.Close()
```
bwplotka: listener is Closed down below

cmd/thanos/store.go
```diff
@@ -185,29 +189,45 @@ func runStore(
...
-			if err := bs.SyncBlocks(ctx); err != nil {
-				level.Warn(logger).Log("msg", "syncing blocks failed", "err", err)
+	// bucketStoreReady signals when bucket store is ready.
+	bucketStoreReady := make(chan struct{})
+	{
+		begin := time.Now()
+		ctx, cancel := context.WithCancel(context.Background())
+		g.Add(func() error {
+			level.Info(logger).Log("msg", "initializing bucket store")
+			if err := bs.InitialSync(ctx); err != nil {
```
bwplotka: Can we just put this in the same group as 
```
err := runutil.Repeat(syncInterval, ctx.Done(), func() error {
				if err := bs.SyncBlocks(ctx); err != nil {
					level.Warn(logger).Log("msg", "syncing blocks failed", "err", err)
				}
				return nil
			})
```
I can't see really the reason why it has to be separate..

cmd/thanos/downsample.go
```diff
@@ -95,7 +95,13 @@ func runDownsample(
 	}()
 
 	metrics := newDownsampleMetrics(reg)
+
 	statusProber := prober.NewProber(comp, logger, prometheus.WrapRegistererWithPrefix("thanos_", reg))
+	// Initiate HTTP listener providing metrics endpoint and readiness/liveness probes.
```
squat: To make this PR a little bit smaller, could we revert these changes that are only moving code around?

cmd/thanos/query.go
```diff
@@ -342,6 +342,36 @@ func runQuery(
 			stores.Close()
 		})
 	}
+	// Start query API + UI, metrics and status probe HTTP server.
+	statusProber := prober.NewProber(comp, logger, reg)
```
squat: same as above

cmd/thanos/receive.go
```diff
@@ -174,6 +173,12 @@ func runReceive(
 		upload = false
 	}
 
+	statusProber := prober.NewProber(comp, logger, prometheus.WrapRegistererWithPrefix("thanos_", reg))
```
squat: same as above

cmd/thanos/rule.go
```diff
@@ -483,6 +482,40 @@ func runRule(
 			close(cancel)
 		})
 	}
+	// Start query API + UI, metrics and status probe HTTP server.
+	statusProber := prober.NewProber(comp, logger, prometheus.WrapRegistererWithPrefix("thanos_", reg))
```
squat: same

cmd/thanos/sidecar.go
```diff
@@ -109,6 +109,16 @@ func runSidecar(
 	comp component.Component,
 	limitMinTime thanosmodel.TimeOrDurationValue,
 ) error {
+	promUp := prometheus.NewGauge(prometheus.GaugeOpts{
```
squat: same

cmd/thanos/store.go
```diff
@@ -185,29 +189,35 @@ func runStore(
...
-	g.Add(func() error {
-		defer runutil.CloseWithLogOnErr(logger, bkt, "bucket client")
-
-		err := runutil.Repeat(syncInterval, ctx.Done(), func() error {
-			if err := bs.SyncBlocks(ctx); err != nil {
-				level.Warn(logger).Log("msg", "syncing blocks failed", "err", err)
+	// bucketStoreReady signals when bucket store is ready.
+	bucketStoreReady := make(chan struct{})
+	{
+		begin := time.Now()
```
bwplotka: This is weird as it shows also non init bucket operation time.. Wonder if we should not keep it together in run.Group?

docs/components/store.md
```diff
@@ -136,6 +136,12 @@ Thanos Store Gateway might not get new blocks immediately, as Time partitioning
...
 We recommend having overlapping time ranges with Thanos Sidecar and other Thanos Store gateways as this will improve your resiliency to failures.
 
-Thanos Querier deals with overlapping time series by merging them together. 
+Thanos Querier deals with overlapping time series by merging them together.
 
 Filtering is done on a Chunk level, so Thanos Store might still return Samples which are outside of `--min-time` & `--max-time`.
+
+## Probes
+
+- Thanos Store exposes two endpoints for probing.
```
bwplotka: Can we add: 

`NOTE: Metric endpoint starts immediately so, make sure you set up readiness probe on designated HTTP `/-/ready` path.`

CHANGELOG.md
```diff
@@ -11,6 +11,10 @@ We use *breaking* word for marking changes that are not backward compatible (rel
 
 ## Unreleased
 
+### Fixed
+
+- [#1656](https://github.com/thanos-io/thanos/pull/1656) Thanos Store now starts metric and status probe HTTP server earlier in its start-up sequence. `/-/healthy` now endpoint starts to respond with success earlier. `/metrics` endpoint now starts serving earlier as well, as a result for your readiness probes you should rely on  `/-/ready` rather than `/metrics`.
```
FUSAKLA:
```suggestion
- [#1656](https://github.com/thanos-io/thanos/pull/1656) Thanos Store now starts metric and status probe HTTP server earlier in its start-up sequence. `/-/healthy` endpoint now starts to respond with success earlier. `/metrics` endpoint starts serving metrics earlier as well. Make sure to point your readiness probes to the `/-/ready` endpoint rather than `/metrics`.
```
Just tiny nits on wording

bwplotka:
Good work @kakkoyun thanks! :heart:

PR #1660, opened 2019-10-17
Add a new prometheus.ready_timeout CLI option to the sidecar

bandesz:

## Changes

The current timeout while the Prometheus instance is starting up is a constant 10 minutes.
As reading the WAL can take a long time we would like to set a custom timeout value, so the Thanos sidecar container is not erroring out every 10 minutes.

## Verification

I ran `make test` with the in-memory store tests.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -11,6 +11,10 @@ We use *breaking* word for marking changes that are not backward compatible (rel
 
 ## Unreleased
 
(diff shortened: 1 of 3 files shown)


bwplotka:
Can you sign the commit as DCO bot suggests? (:

bandesz:
@bwplotka sure, done!

Issue #1661, opened 2019-10-17
compactor: weird directory structure


Reamer:
**Thanos, Prometheus and Golang version used**:
```
/ $ /bin/thanos --version
thanos, version 0.8.1 (branch: HEAD, revision: a328f4562f77deefcd990330681a62f7de049e67)
  build user:       root@my_company-thanos-1-build
  build date:       20191016-07:10:48
  go version:       go1.13.1
```
**Object Storage Provider**:
S3 with a minio setup

**What happened**:
Compaction directory structure is weird when a label includes a `/`. The additional `/` produces a subfolder.

```
ls -l /data/compact/
total 20
drwxr-sr-x    4 10003300 10003300      4096 Oct 17 13:03 0@{cluster="s-openshift.my_company.de",prometheus="openshift-monitoring
drwxr-sr-x    4 10003300 10003300      4096 Oct 17 13:03 300000@{cluster="s-openshift.my_company.de",prometheus="openshift-monitoring
drwxr-sr-x    4 10003300 10003300      4096 Oct 17 13:03 3600000@{cluster="s-openshift.my_company.de",prometheus="openshift-monitoring
```

**What you expected to happen**:

A nice folder without weird subfolders to chunks

```
ls -l /data/compact/3600000@{cluster="s-openshift.my_company.de",prometheus="openshift-monitoring/my_company",prometheus_replica="_agg_replica_"}:
total 36
drwxr-sr-x    2 10003300 10003300      4096 Oct 17 13:03 01DQC4N9T0YNFZZR27RD5GCAG5
drwxr-sr-x    2 10003300 10003300      4096 Oct 17 13:03 01DQC4PQ6NTRH3BJYQGJ1S6CKK
drwxr-sr-x    2 10003300 10003300      4096 Oct 17 13:03 01DQC4R5NBDYY241MHJW0MN74D
drwxr-sr-x    2 10003300 10003300      4096 Oct 17 13:03 01DQC4SYR49E14SAN1245ZZZA2
drwxr-sr-x    2 10003300 10003300      4096 Oct 17 13:03 01DQC4VGDG3Y9YS46FZTZ9EPJR
drwxr-sr-x    2 10003300 10003300      4096 Oct 17 13:03 01DQC4WYWR0ZY2PR6ZTXTHJQ1C
drwxr-sr-x    2 10003300 10003300      4096 Oct 17 13:03 01DQC4YB870WD2X7DD9V09SYGG
```

[...]

Issue #1499, opened 2019-09-06
Compactor: Compaction of Raw Data without cleaning local Disk


Jakob3xD:
**Thanos, Prometheus and Golang version used**
Thanos: v0.7.0 Prometheus: latest (v2.12.0)

**What happened**
When compacting Raw data the local disk wont clean after a compaction is done.
In my case this will cause the compactor to use more then 1TB of Space when stating a lvl 4 compaction.

**What you expected to happen**
I expected that the compactor clears the data on his local disk after he uploaded the compacted block to the object storage like it is done on the down-sampled data.
It was expected that the compactor crashes once the 1TB disk was full.

**How to reproduce it (as minimally and precisely as possible)**:
Some blocks form several Sidecars which needs to compact.
![image](https://user-images.githubusercontent.com/29206900/64410804-1eaa8000-d08c-11e9-8b5d-1c5e4ef2163f.png)
The one larger peak at 11:15 is the finish of the level 3 compaction.
Each small peak you see is a new compaction to level 4.
**Full logs to relevant components**
Once the disk is full:
level=error ts=2019-09-05T16:51:30.809934432Z caller=main.go:213 msg="running command failed" err="error executing compaction: compaction failed: compaction failed for group 0@{monitor=\"node\",prom=\"prom-30\",region=\"fsn1\",replica=\"a\"}: download block 01DM09QBNY0SDHGPT6J71FPST2: copy object to file: write /tmp/thanos/compact/0@{monitor=\"node\",prom=\"prom-30\",region=\"fsn1\",replica=\"a\"}/01DM09QBNY0SDHGPT6J71FPST2/chunks/000002: no space left on device"

**Anything else we need to know**

krasi-georgiev:
When you run the sidecar to upload the blocks to remote storage you should have the compaction disabled
see: https://github.com/thanos-io/thanos/blob/master/docs/components/sidecar.md#example-basic-deployment

If you need the compaction enabled for some reason you can set the Prometheus size or time retention flag so that older blocks get deleted.

Jakob3xD:
@krasi-georgiev 
I wasn't talking about the Prometheus compactor.  I mean the Thanos Compactor running over the object storage.

krasi-georgiev:
@Jakob3xD 	oops, sorry for the confusion.
I just had a quick look and the compactor does indeed clear the tmp folder before every compaction.

[...]

PR #1666, opened 2019-10-18
Fixed compactor tests; Moved to full e2e compact test; Cleaned metrics.

bwplotka:
PR is big but I recommend review commit by commit (: 

Fixes: https://github.com/thanos-io/thanos/issues/1499
Fixes: https://github.com/thanos-io/thanos/issues/1661

cc @krasi-georgiev as I changed your tests (one was broken, some I moved to proper, full compact_e2e)


pkg/compact/compact.go
```diff
@@ -592,8 +595,16 @@ func (cg *Group) Resolution() int64 {
 // Compact plans and runs a single compaction against the group. The compacted result
 // is uploaded into the bucket the blocks were retrieved from.
 func (cg *Group) Compact(ctx context.Context, dir string, comp tsdb.Compactor) (bool, ulid.ULID, error) {
+	cg.compactionRuns.Inc()
```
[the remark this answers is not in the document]
bwplotka: How? It is `attempts`?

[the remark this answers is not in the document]
bwplotka: But it's fair happy to move it at the end.

bwplotka:
Thanks for looking into this metric "swap" @brancz - I am not 100% particularly sure if we won't confuse users, but I think what users was expecting seeing `compactions_total` was actually compactions that resulted in block (:

pkg/compact/compact.go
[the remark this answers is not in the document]
brancz: How about we have both `started` and `completed`? :)

CHANGELOG.md
```diff
@@ -15,6 +15,11 @@ We use *breaking* word for marking changes that are not backward compatible (rel
 
 - [#1660](https://github.com/thanos-io/thanos/pull/1660) Add a new `--prometheus.ready_timeout` CLI option to the sidecar to set how long to wait until Prometheus starts up.
 
+### Changed
+
+- [](https://github.com/thanos-io/thanos/pull/) `thanos_compact_group_compactions_total` now counts block compactions, so operation that resulted in compacted block. The old behaviour
```
squat:
```suggestion
- [](https://github.com/thanos-io/thanos/pull/) `thanos_compact_group_compactions_total` now counts block compactions, so operations that resulted in a compacted block. The old behaviour
```

CHANGELOG.md
```diff
@@ -15,6 +15,11 @@ We use *breaking* word for marking changes that are not backward compatible (rel
...
+is now exposed by new metric: `thanos_compact_group_compaction_runs_total` which counts compaction runs overall.
```
squat:
```suggestion
is now exposed by a new metric: `thanos_compact_group_compaction_runs_total`, which counts compaction runs overall.
```

pkg/compact/compact.go
```diff
@@ -115,7 +110,15 @@ func newSyncerMetrics(reg prometheus.Registerer) *syncerMetrics {
 
 	m.compactions = prometheus.NewCounterVec(prometheus.CounterOpts{
 		Name: "thanos_compact_group_compactions_total",
-		Help: "Total number of group compactions attempts.",
+		Help: "Total number of group compactions attempts, resulted with new block.",
```
squat:
```suggestion
		Help: "Total number of group compaction attempts that resulted in a new block.",
```

pkg/compact/compact.go
```diff
@@ -115,7 +110,15 @@ func newSyncerMetrics(reg prometheus.Registerer) *syncerMetrics {
...
+	}, []string{"group"})
+	m.compactionRunsStarted = prometheus.NewCounterVec(prometheus.CounterOpts{
+		Name: "thanos_compact_group_compaction_runs_started_total",
+		Help: "Total number of group compactions run attempts.",
```
squat:
```suggestion
		Help: "Total number of group compaction attempts.",
```

pkg/compact/compact.go
```diff
@@ -115,7 +110,15 @@ func newSyncerMetrics(reg prometheus.Registerer) *syncerMetrics {
...
+	}, []string{"group"})
+	m.compactionRunsFinished = prometheus.NewCounterVec(prometheus.CounterOpts{
+		Name: "thanos_compact_group_compaction_runs_finished_total",
+		Help: "Total number of group compactions run finished. This also includes compactor group runs that resulted with no compaction.",
```
squat:
```suggestion
		Help: "Total number of group compaction runs finished. This also includes compactor group runs that resulted in no compaction.",
```

pkg/compact/compact.go
```diff
@@ -342,12 +347,12 @@ func (c *Syncer) removeIfMetaMalformed(ctx context.Context, id ulid.ULID) (remov
...
 // the downsampling resolution and the block's labels.
-func GroupKey(meta metadata.Meta) string {
-	return groupKey(meta.Thanos.Downsample.Resolution, labels.FromMap(meta.Thanos.Labels))
+func GroupKey(meta metadata.Thanos) string {
+	return groupKey(meta.Downsample.Resolution, labels.FromMap(meta.Labels))
 }
 
 func groupKey(res int64, lbls labels.Labels) string {
-	return fmt.Sprintf("%d@%s", res, lbls)
+	return fmt.Sprintf("%d@%s", res, fmt.Sprintf("%v", lbls.Hash()))
```
squat: Why two nested `Sprintf` calls?

```suggestion
	return fmt.Sprintf("%d@%d", res, lbls.Hash())
```

pkg/compact/compact.go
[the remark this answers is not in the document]
bwplotka: done

pkg/objstore/inmem/inmem.go
```diff
@@ -86,6 +92,8 @@ func (b *Bucket) Get(_ context.Context, name string) (io.ReadCloser, error) {
 		return nil, errors.New("inmem: object name is empty")
 	}
 
+	b.mtx.RLock()
+	defer b.mtx.RUnlock()
```
squat: We could be a bit more specific with the unlocking here as there is no need to lock while reading the actual file.

pkg/objstore/inmem/inmem.go
```diff
@@ -100,6 +108,8 @@ func (b *Bucket) GetRange(_ context.Context, name string, off, length int64) (io
...
```
squat: same as above

pkg/objstore/inmem/inmem.go
```diff
@@ -16,7 +17,9 @@ import (
 var errNotFound = errors.New("inmem: object not found")
 
 // Bucket implements the store.Bucket and shipper.Bucket interfaces against local memory.
+// methods from Bucket interface are thread-save. Object are assumed immutable.
 type Bucket struct {
+	mtx     sync.RWMutex
```
squat: Nothing is ever write-locking this mutex, so what is this mutex actually protecting against?

pkg/objstore/inmem/inmem.go
```diff
@@ -16,7 +17,9 @@ import (
 var errNotFound = errors.New("inmem: object not found")
 
 // Bucket implements the store.Bucket and shipper.Bucket interfaces against local memory.
+// methods from Bucket interface are thread-save. Object are assumed immutable.
```
squat:
```suggestion
// methods from Bucket interface are thread-safe. Object are assumed to be immutable.
```

pkg/compact/compact.go
bwplotka: :man_facepalming:

pkg/objstore/inmem/inmem.go
bwplotka: true, but we don't read - we just return the reader. ;p so it does not matter. It is not safe concurrent in case for deletions anyway

squat: yes exactly

Issue #1649, opened 2019-10-14
Discussion: Improve memory use for queries


ppanyukov:
This is a tracker issue for **memory utilisation in Thanos to serve Queries** and the steps we want to take to make things better.

Core of the issue:

- Both **Querier** and **Store Gateway** allocate unbounded amounts of memory during query execution.
- This can and does lead to OOM in resource-limited container environments (k8s etc).

What we want is:

- A list of ideas to consider and try;
- Status of each idea based on discussions, tests etc;

# Repro

**Base build**
```
thanos, version 0.7.0 (branch: master, revision: ed045447fef8c4b5ff62697559463549ff522f31)
  build user:       philip@Mayas-MacBook-Pro.local
  build date:       20191010-13:52:26
  go version:       go1.13.1
```

**Data**
- synthetically generated (uses `thanosbench blockgen`)
- data size: 
    - metrics: 200
    - targets: 100
    - sample interval: 15s
    - total time interval: 10 hours

**Query**
- `count({__name__=~".+"}) by (__name__)`
- time span: all generated data
- curl: 

```
curl 'http://localhost:11902/api/v1/query_range?query=count(%7B__name__%3D~%22.%2B%22%7D)%20by%20(__name__)&dedup=true&partial_response=true&start=1568634900&end=1569844500&step=4838&max_source_resolution=0s&_=1570622122482'
```

**Memory profile**
- MacBook Pro 16GB RAM
- Store Gateway: 20GB peak
- Querier: 10GB peak before time-out
- See chart
 

**Chart**
- Green is Storage Gateway; Red is Querier
- Note that the query timed out so does not show Querier in its entirety.

<img width="687" alt="Screenshot 2019-10-10 at 14 42 08" src="https://user-images.githubusercontent.com/300031/66575827-8746ba00-eb6e-11e9-9d8a-758af9b02b51.png">

# List of ideas

Per-query limits and flow optimisations looks to be the way to go.

[1.] ~~Simple global memory limiter in Querier.~~ Rejected. See https://github.com/thanos-io/thanos/pull/1631 .

   [1.1] Simple global memory limiter in **SG**? I've just realised that the OOM problem actually may be in **SG** that causes most pain, not in Querier.  Worth considering?

[2.] Limit concurrent queries, max samples, and series(?). As per Cortex: https://github.com/thanos-io/thanos/pull/1631#issuecomment-541616487

[3.] Frontend query caching. See https://github.com/thanos-io/thanos/issues/1006

[...]

PR #1667, opened 2019-10-18
[Demo] Query/SG: chopped queries

ppanyukov:
# DEMO. NOT FOR MERGE

**TL;DR:** *NOT WORTH IT* (unless I did something dumb, in which case can revisit)

---

This is a small demo of the chunked queries:

- Part of the `[8.]` item in https://github.com/thanos-io/thanos/issues/1649
- PR is for historical purposes to record the results.

**Change:**

- The change is in querier.
- Given time interval `[mint, maxt]`, chop it into smaller `1h` intervals `[mint, mint+1h], [mint+1h, mint+2h] ... [maxt-1h, maxt]`.
- Ask Store Gateway (SG) to give results for each smaller chunk.
- Assemble results in the querier.

**Purpose:**

- See if this has positive impact on memory utilisation in both Querier and SG components;
- See if query execution time is affected in any negative way.

**Results:**

- SG memory utilisation is really much better (flat line).
- Querier is way way higher though.
- Also it's quite a bit slower.
- Query used: `count({__name__=~".+"}) by (__name__)`
- See stats and charts.

```
ORIGINAL IMPLEMENTATION:

    real    1m30.330s
    user    0m0.014s
    sys     0m0.021s

    Peak MEM:
        SG  22GB
        Q   16GB


CHUNKED:

    real    2m13.879s
    user    0m0.014s
    sys     0m0.022s

    Peak MEM:
        SG  6GB
        Q   25GB

```

**Conclusion:**

*NOT WORTH IT* 

- Looks to shift the problem from SG to Querier at the cost of latency.
- Questionable benefit.
- Unless I did something dumb, in which case can revisit later.

** Chart **

Screenshot 2019-10-18 at 14 27 13

### DIFF
--- a/pkg/store/proxy.go
+++ b/pkg/store/proxy.go
@@ -184,16 +184,18 @@ func (s ctxRespSender) send(r *storepb.SeriesResponse) {
 func (s *ProxyStore) Series(r *storepb.SeriesRequest, srv storepb.Store_SeriesServer) error {
 	if os.Getenv("CHUNKED") != "" {
 		level.Info(s.logger).Log("msg", "func (s *ProxyStore) Series -> CHUNKED")
-		return s.SeriesChunked(r, srv)
+		return s.seriesChunked(r, srv)
 	}
 
 	level.Info(s.logger).Log("msg", "func (s *ProxyStore) Series -> ORIG")
-	return s.SeriesOrig(r, srv)
+	return s.seriesOrig(r, srv)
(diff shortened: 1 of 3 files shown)


ppanyukov:
Closing as it's for historical purposes only.

bwplotka:
Nice! Good work, fun fact is that overall it is less memory (:

PR #1668, opened 2019-10-18
cmd/thanos/receive: remote-write client+server TLS

squat:
This commit gives the Thanos receive component the capability to use TLS
in both the remote-write client and server. This means that Thanos
receive can now authenticate all requests.

In order to accomplish this change, this commit abstracts the majority
of the logic of `defaultGRPCServerOpts` into a reusable func for gRPC
and HTTP servers and creates a similar func for TLS client
configurations.

cc @bwplotka @brancz, this is following our conversation yesterday about missing authn and encryption options for Thanos receive.

### DIFF
--- a/cmd/thanos/main.go
+++ b/cmd/thanos/main.go
@@ -246,17 +246,29 @@ func registerMetrics(mux *http.ServeMux, g prometheus.Gatherer) {
 	mux.Handle("/metrics", promhttp.HandlerFor(g, promhttp.HandlerOpts{}))
 }
 
-func defaultGRPCServerOpts(logger log.Logger, cert, key, clientCA string) ([]grpc.ServerOption, error) {
+func defaultGRPCTLSServerOpts(logger log.Logger, cert, key, clientCA string) ([]grpc.ServerOption, error) {
 	opts := []grpc.ServerOption{}
+	tlsCfg, err := defaultTLSServerOpts(log.With(logger, "protocol", "gRPC"), cert, key, clientCA)
+	if err != nil {
+		return opts, err
+	}
+	if tlsCfg != nil {
+		opts = append(opts, grpc.Creds(credentials.NewTLS(tlsCfg)))
+	}
+	return opts, nil
+}
 
+func defaultTLSServerOpts(logger log.Logger, cert, key, clientCA string) (*tls.Config, error) {
 	if key == "" && cert == "" {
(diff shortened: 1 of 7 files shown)


cmd/thanos/main.go
```diff
@@ -248,16 +248,28 @@ func registerMetrics(mux *http.ServeMux, g prometheus.Gatherer) {
 
 func defaultGRPCServerOpts(logger log.Logger, cert, key, clientCA string) ([]grpc.ServerOption, error) {
```
bwplotka: Looks like it is literally `defaultGRPCTLSServerOpts` right? no other opts (:

cmd/thanos/main.go
```diff
@@ -288,10 +298,58 @@ func defaultGRPCServerOpts(logger log.Logger, cert, key, clientCA string) ([]grp
...
 
-		level.Info(logger).Log("msg", "gRPC server TLS client verification enabled")
+		level.Info(logger).Log("msg", "server TLS client verification enabled")
+	}
+
+	return tlsCfg, nil
+}
+
+func defaultTLSClientOpts(logger log.Logger, secure bool, cert, key, caCert, serverName string) (*tls.Config, error) {
+	if !secure {
```
bwplotka: I know this does not make sense from a high level, but I think... let's remove this. With this change if someone use `secure false ` and key and cert set we silently ignore key and cert. I think failing would be better.

squat: in that case should we remove the `-secure` flag as well? I also think it is confusing to explicitly have to opt into client TLS in addition to specifying the certs. but it's a breaking change

bwplotka: I think yes - `secure` only makes sense if you have certs (?)

squat:
Agreed, it is over constrained otherwise.

cmd/thanos/main.go
squat: hmm I'm looking at this again and I think the actual problem here is that every gRPC ClientConn created by the querier re-uses the same set of `[]DialOpt`, which must be either secure OR insecure, but not both. For this reason, the `grpc-client-tls-secure` flag is used to toggle off ALL client TLS features. Another option would be to use the insecure DialOpt unless ANY client TLS flag is provided.

squat:
ping @bwplotka (:

PR #1671, opened 2019-10-21
cmd/thanos/main.go: simplify pprof registration

squat:
This commit simplifies the registration of pprof HTTP endpoints. The
pprof.Index handler automatically takes care of delegating to the
correct handler for each profile depending on the request path:
https://golang.org/src/net/http/pprof/pprof.go?s=8862:9042#L260
The following profiles are handled:
https://golang.org/src/net/http/pprof/pprof.go?s=7565:8570#L248
Note that this also includes the `allocs` profile, which was previously
not explicitly added.

## Verification

Built binary and verified all pprof endpoints continue to function.

cc @bwplotka

### DIFF
--- a/cmd/thanos/main.go
+++ b/cmd/thanos/main.go
@@ -236,10 +236,6 @@ func registerProfile(mux *http.ServeMux) {
 	mux.HandleFunc("/debug/pprof/profile", pprof.Profile)
 	mux.HandleFunc("/debug/pprof/symbol", pprof.Symbol)
 	mux.HandleFunc("/debug/pprof/trace", pprof.Trace)
-	mux.Handle("/debug/pprof/block", pprof.Handler("block"))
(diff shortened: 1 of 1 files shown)


cmd/thanos/main.go
```diff
@@ -236,10 +236,6 @@ func registerProfile(mux *http.ServeMux) {
 	mux.HandleFunc("/debug/pprof/profile", pprof.Profile)
 	mux.HandleFunc("/debug/pprof/symbol", pprof.Symbol)
 	mux.HandleFunc("/debug/pprof/trace", pprof.Trace)
```
bwplotka: so those are not added by  `index`?

squat: ack, only the profiles accessible via `pprof.Handler(...)` are delegated.

PR #1672, opened 2019-10-21
[WIP] allow server cert rotation without a restart

krasi-georgiev:
I did a quick test and reading the file from disk on every request is ok sinse the kernel puts this in the pagecache so in practice it reads the file from memory when not changed and reads it from disk when it is replaced with a new one.

- [x] tests
- [x] refactor to reduce the performance overhead
- [x] try to implement something similar for the CA rotation.
- [ ] the CA rotation works only for the client CAs and not for the server CAs. Missing `GetServerConfig` as mentioned in https://github.com/golang/go/issues/16066#issuecomment-250606132
Tracking issue: https://github.com/golang/go/issues/22836
should be able to use `VerifyPeerCertificate` for this - https://github.com/golang/go/issues/22836#issuecomment-584968697

related modules that use the same idea.
https://godoc.org/golang.org/x/crypto/acme/autocert
https://github.com/johanbrandhorst/certify

I did a quick test and reading the same file from disk is not a problem as the kernel caches it in the memory when the file is not changed.

<details>
<summary>Test for reading the same file again doesn't hit the disk</summary>

```
func Test_io(t *testing.T) {
	pid := os.Getpid()
	for {
		_, err := ioutil.ReadFile("env")
		testutil.Ok(t, err)

		io, err := ioutil.ReadFile("/proc/" + strconv.Itoa(pid) + "/io")
		testutil.Ok(t, err)

		fmt.Println(string(io))
		time.Sleep(500 * time.Millisecond)
	}
}
```

</details>


brancz:
This reads and parses the certificates on every request. I feel it would be appropriate to do this periodically instead and atomically swap them.

krasi-georgiev:
> This reads and parses the certificates on every request. I feel it would be appropriate to do this periodically instead and atomically swap them.

Yep, and it is the second item on my to do list in the PR description

>  refactor to reduce the performance overhead

I already refactored it to compare mod times and only when different will reload the certificated.

by the way currently the Prometheus scrape client does this with every request which I agree should be improved.
https://github.com/prometheus/common/blob/2afc453a9de43c46dadd462998d096e52cd4e4e3/config/http_config.go#L348

brancz:
Sorry was too hasty. All good :)

pkg/tls/tls.go
```diff
@@ -0,0 +1,149 @@
...
+		if !certPool.AppendCertsFromPEM(caPEM) {
+			return nil, errors.New("adding CA file to the client pool")
+		}
+
+		caCert, err := tls.X509KeyPair(caPEM, caPEM)
+		if err != nil {
+			return nil, errors.Wrap(err, "building client CA certificate")
+		}
+
+		// ????? test if true...... This is a pointer it will update the underlying tls config pool.
```
GiedriusS: I guess this is still a work in progress because of these comments?

bwplotka:
Do we have a story for compatible rotation? e.g how this will work between server and client? (:

pkg/tls/tls.go
krasi-georgiev: yep one of the items in my todo list
> test if the CA pool gets updated with the ponter

brancz:
Even if a single file, CAs can usually be loaded as bundles so in reality there are multiple CAs in one file.

krasi-georgiev:
now after writing the tests I realized that there is another problem.
`getConfigForClient` and `getCertificate` are called only with the  tls handshake and NOT when sending or receiving data which means that for long running data flow it will never reload the certs as it will reuse the same connection and no new tls handshake will happen.
In this PR I have also added a cli so users can set different keep alive setting with `MaxConnectionAge`. 

New tls handshake is initiated when the connection is interrupted or when the `MaxConnectionAge` expires.

So with this new info I think there are 3 options:
1. keep the current implementation with a cli flag for `MaxConnectionAge`
2. run some loop which will restart the server gracefully or drop all connections when it detects tls cert files changes.
3. provide a reload endpoint and leave this to external config management.

Again I think this is important mainly for the thanos receiver as I imagine that with all other components a new connection and a new tls handshake would be created with every request so it will reload the certs as expected.

cc @bwplotka, @s-urbaniak

krasi-georgiev:
@squat , @bwplotka @s-urbaniak waiting for some feedback on which options do you think it best for this implementation.

s-urbaniak:
@krasi-georgiev Sorry for the late reply. I am voting for option "1. keep the current implementation with a cli flag for MaxConnectionAge"

I believe this is a sensible setting and less disruptive than the other options. Also, it does not require external coordination (someone has to implement that timeout).

Finally cert rotation happens in timespans of minutes, maybe even hours so a MaxConnectionAge in that timespan sounds sensible to me too.

@bwplotka @brancz Do you have additional thoughts or objections?

bwplotka:
>  the CA rotation works only for the client CAs and not for the server CAs. Missing GetServerConfig as mentioned in golang/go#16066 (comment)

How did we work around this one?

So I think MaxConnectionAge here might be too short by default (`1m`), as kind of defies any KeepAlive logic.. but maybe something longer would make sense. Also something I would like to check if `MaxConnectionAge` is actually accounted for during long-running requests, although I think we don't do those longer than X minutes. 

What's the requirement for those cert rotations in terms of delay? I think ~minutes is ok, so maybe we could start with this, but larger `MaxConnectionAge`

s-urbaniak:
> What's the requirement for those cert rotations in terms of delay? I think ~minutes is ok, so maybe we could start with this, but larger MaxConnectionAge

agreed, the ~minutes ballpark sounds good to me.

krasi-georgiev:
After a bit of a struggle the tls rotation is implemented with full e2e tests.

As mentioned in the first PR comment server CA rotation is missing becasue it is yet not added in the golang std library so I left a TODO note to add it when it is added in the golang std library. I don't think that this is a blocker for this PR and I will follow up the tracking golang issues for updates.

@squat I also noticed that there are no e2e tests for the reciever TLS, should I add it to this PR or you prefer to open another PR after this one is merged?

pkg/tls/tls.go
```diff
@@ -0,0 +1,213 @@
...
+
+	tlsCfg := &tls.Config{
+		RootCAs: certPool,
+	}
+
+	if serverName != "" {
+		tlsCfg.ServerName = serverName
+	}
+
+	if (keyPath != "") != (certPath != "") {
```
squat: This check is duplicated from line 129

pkg/tls/tls.go
```diff
@@ -0,0 +1,213 @@
...
+		mngr := &clientTLSManager{
+			certPath: certPath,
+			keyPath:  keyPath,
+		}
+		// TODO (krasi) implement CA rotation when added in the std.
+		// With the current implementation the CA's can't be auto reloaded without a restart.
+		tlsCfg.GetClientCertificate = mngr.getClientCertificate
+		level.Info(logger).Log("msg", "TLS client authentication enabled")
+	}
+	return tlsCfg, nil
```
squat: nit: add a newline before this return for legibility

test/e2e/tls_test.go
```diff
@@ -0,0 +1,199 @@
...
+	"github.com/fortytw2/leaktest"
+	"github.com/go-kit/kit/log"
+	"github.com/thanos-io/thanos/pkg/testutil"
+	"google.golang.org/grpc"
+	"google.golang.org/grpc/credentials"
+	"google.golang.org/grpc/keepalive"
+
+	thTLS "github.com/thanos-io/thanos/pkg/tls"
+
+	pb "google.golang.org/grpc/examples/features/proto/echo"
```
squat: nit: import ordering

test/e2e/tls_test.go
```diff
@@ -0,0 +1,198 @@
...
+	SubjectKeyId: []byte{1, 2, 3},
+	ExtKeyUsage:  []x509.ExtKeyUsage{x509.ExtKeyUsageClientAuth, x509.ExtKeyUsageServerAuth},
+	KeyUsage:     x509.KeyUsageDigitalSignature,
+}
+
+// genCerts generates certificates and writes those to the provided paths.
+// When the CA file already exists it is not overwritten and
+// it is used to sign the certificates.
+func genCerts(t *testing.T, certPath, privkeyPath, caPath, serverName string) {
+	caPrivKey, err := rsa.GenerateKey(rand.Reader, 1024)
```
squat: nit: only generate if no CA is found

pkg/tls/tls.go
```diff
@@ -0,0 +1,213 @@
...
+	"time"
+
+	"github.com/go-kit/kit/log"
+
+	"github.com/go-kit/kit/log/level"
+	"github.com/pkg/errors"
+)
+
+// NewServerConfig returns a new server tls config with auto reloading certificates.
+// It reload the key and certificates when the local mod time of any file has changed.
```
squat:
```suggestion
// It reloads the key and certificates when the local mod time of any file has changed.
```

pkg/tls/tls.go
```diff
@@ -0,0 +1,213 @@
...
+			return nil, errors.New("adding CA file to the client pool")
+		}
+		m.cltCAmodTime = statCA.ModTime()
+	}
+
+	return m.cltTLSConfig, nil
+}
+
+// NewClientConfig returns a new client tls config with auto reloading certificates.
+// It reload the key and certificates when the local mod time of any file has changed.
```
squat:
```suggestion
// It reloads the key and certificates when the local mod time of any file has changed.
```

krasi-georgiev:
@squat thanks for the review, addressed all comments.

didn't answer my original question though :smiley_cat: 

> @squat I also noticed that there are no e2e tests for the reciever TLS, should I add it to this PR or you prefer to open another PR after this one is merged?

squat:
let's do it in a new PR to keep the concerns separate :) I can make the PR

krasi-georgiev:
@brancz @bwplotka any other comments for this PR?

krasi-georgiev:
ping @bwplotka

GiedriusS:
@krasi-georgiev are there any numbers on the performance impact of the current implementation? :eyes: does Windows exhibit the same behavior with caching?

krasi-georgiev:
> @krasi-georgiev are there any numbers on the performance impact of the current implementation? 

probably, but only for single syscall to get the `ModTime()` so I would say negligible. If anyone wants to disable this can set a high enough  `--grpc-server-max-connection-age` so that the connection is persistent and it never checks for new certs.


>  does Windows exhibit the same behavior with caching?

I would say yes, but is windows even suported by Thanos?
Last time I tried to run the tests under windoes these failed in many places.

xunleii:
Thanks for this PR, it is awesome to automatically reload certificates when they are modified. 

However, I have a problem on Kubernetes; when we mount a secret or a configmap as a folder, Thanos doesn't refresh automatically the certificates. I think this is due to the check of the file's modification time: because Kubernetes creates symbolic links when we mount a secret or a configmap as a volume (in order to refresh the mounted files when the secret/configmap changes), these links are never modified.

A possible workaround for that is to follow symbolic links before getting the file's stats (like in https://github.com/thanos-io/thanos/pull/1672/files#diff-ce85c8e1672751ec281d5c28749989b7R79).

bwplotka:
Still much needed. Looking for more time to review it properly. Help wanted.

bwplotka:
It is indeed (:

bwplotka:
still valid (:

bwplotka:
For review, help wanted.

Kind Regards,
Bartek Płotka (@bwplotka)

krasi-georgiev:
stale, someone else needs to take over.

Tachone:
> now after writing the tests I realized that there is another problem. `getConfigForClient` and `getCertificate` are called only with the tls handshake and NOT when sending or receiving data which means that for long running data flow it will never reload the certs as it will reuse the same connection and no new tls handshake will happen. In this PR I have also added a cli flag so users can set the keep alive setting with `MaxConnectionAge` which enforces a new TLS hanshake when `MaxConnectionAge` expires. New tls handshake is initiated when the connection is interrupted or when the `MaxConnectionAge` expires.
> 
> So with this new info I think there are 3 options:
> 
> 1. keep the current implementation with a cli flag for `MaxConnectionAge`
> 2. run some loop which will restart the server gracefully or drop all connections when it detects tls cert files changes.
> 3. provide a reload endpoint and leave reloading  to external config management.
> 
> Again I think this is important mainly for the thanos receiver as I imagine that with all other components a new connection and a new tls handshake would be created with every request so it will reload the certs as expected.
> 
> cc @bwplotka, @s-urbaniak

Why do we need to set `MaxConnectionAge`, Is there any problem with the long connection not being closed? the old certificate is no longer used on  the established connection, am i wrong?

PR #1675, opened 2019-10-22
querier: Actually use select Start,End params during select.

bwplotka:
This will definitely helps with offset queries.

Same was done on https://github.com/cortexproject/cortex/pull/1012

### DIFF
--- a/pkg/query/api/v1.go
+++ b/pkg/query/api/v1.go
@@ -503,7 +503,8 @@ func (api *API) series(r *http.Request) (interface{}, []error, *ApiError) {
 	}
 
 	// TODO(bwplotka): Support downsampling?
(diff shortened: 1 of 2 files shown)


squat:
nice catch! looks good, just need to update a test, it seems.

```
--- FAIL: TestEndpoints (0.02s)
    --- FAIL: TestEndpoints/end=-1&match%5B%5D=test_metric2&start=-2 (0.00s)
        v1_test.go:831: Response does not match, expected:
            []
            got:
            [{__name__="test_metric2", foo="boo"}]
FAIL
FAIL	github.com/thanos-io/thanos/pkg/query/api	0.033s
```

bwplotka:
It was actually bug, so test helped. Should be fine now :+1:

brian-brazil:
This is okay, as long as you're using a new enough version of Prometheus that actually sends these.

bwplotka:
Good point, we plan to fetch latest Prometheus soon as well.

PR #1679, opened 2019-10-23
Upload blocks with the lowest minTime first

### LINKED ISSUE #1670: sidecar: Upload blocks from oldest
Ref: https://cloud-native.slack.com/archives/CK5RSSC10/p1571650833206600

It can happen during long network partition that uploaded blocks in the wrong order can cause compactor to miss one block during compaction. This results in overlapping blocks - compactor halts in that case. Read path works.

AC:
* Sort local blocks based on min/max time
* Optional: Add repair for verifier with vertical compaction - this will allow manual repair for those case (the alternative is to change a bit different meta to the overlapped block OR remove it)

obiesmans:
Fixes #1670

Signed-off-by: Olivier Biesmans <olivier.biesmans@blablacar.com>

## Changes

Sidecar now uploads blocks with the lowest minTime first

### DIFF
--- a/pkg/shipper/shipper_test.go
+++ b/pkg/shipper/shipper_test.go
@@ -3,8 +3,10 @@ package shipper
 import (
 	"io/ioutil"
 	"math"
+	"math/rand"
 	"os"
 	"path"
+	"sort"
 	"testing"
 
 	"github.com/go-kit/kit/log"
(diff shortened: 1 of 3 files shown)


obiesmans:
Work in progress. But feedback is already welcome. 
I'd like to add a repair for `thanos bucket verify` althought I have no idea how to fix the vertical compression issues except by deleting one of the overlapping blocks.

pkg/shipper/shipper_test.go
```diff
@@ -77,3 +79,132 @@ func TestShipperTimestamps(t *testing.T) {
...
+						MinTime: int64(i * 1000),
+						Version: 1,
+					},
+				},
+			),
+		)
+	}
+	rand.Shuffle(len(metas), func(i, j int) {
+		metas[i], metas[j] = metas[j], metas[i]
+	})
```
bwplotka: Why are we not resetting timer here?

pkg/shipper/shipper_test.go
```diff
@@ -77,3 +79,132 @@ func TestShipperTimestamps(t *testing.T) {
...
+	b.ResetTimer()
+
+	shipper := New(nil, nil, dir, nil, nil, metadata.TestSource)
+	if err := shipper.iterBlockMetas(func(m *metadata.Meta) error {
```
bwplotka: Can we remove this function? I think you left it to compare benchmarks but actually the proper flow would be to create a benchmark for old thing, then have new commit with new function and tested in the same benchmark name so we can use something like this: https://github.com/prometheus/prombench/tree/master/funcbench (: or https://godoc.org/golang.org/x/tools/cmd/benchcmp

bwplotka:
Also can we have BlaBla car badge on our website? (: https://thanos.io/ ([here](https://thanos.io/contributing/how-to-contribute-to-docs.md/#adopters-logos) is how to add it)

pkg/shipper/shipper_test.go
obiesmans: Didn't know about those tools, I'll remove the old `iterBlockMetas()`.

obiesmans:
For the benchmarks, there's basically no difference : the number of blocks is quite low and the blocks are almost sorted. 
While `sort.Slice()` relies on quicksort, which doesn't do very well on almost sorted data, I don't think it's worth looking further on a low number of blocks (100 2-hours blocks is already more that 8 days retention).


```
goos: darwin
goarch: amd64
pkg: github.com/thanos-io/thanos/pkg/shipper
BenchmarkIterBlockMetas-8         	1000000000	         0.0236 ns/op
BenchmarkIterSortedBlockMetas-8   	1000000000	         0.0207 ns/op
PASS
ok  	github.com/thanos-io/thanos/pkg/shipper	9.861s
```

pkg/shipper/shipper_test.go
obiesmans: Nice catch.

pkg/shipper/shipper_test.go
obiesmans: Deleted !

pkg/shipper/shipper_test.go
```diff
@@ -77,3 +79,85 @@ func TestShipperTimestamps(t *testing.T) {
...
+						ULID:    id,
+						MaxTime: int64((i + 1) * 1000),
+						MinTime: int64(i * 1000),
+						Version: 1,
+					},
+				},
+			),
+		)
+	}
+	rand.Shuffle(len(metas), func(i, j int) {
```
bwplotka: I would vote for actually organizing initial metas ourselves. Relying on some random shuffling is not really deterministic.

pkg/shipper/shipper_test.go
```diff
@@ -77,3 +79,85 @@ func TestShipperTimestamps(t *testing.T) {
...
+		return nil
+	}); err != nil {
+		testutil.Ok(t, err)
+	}
+	testutil.Equals(t, sort.SliceIsSorted(metas, func(i, j int) bool {
+		return metas[i].BlockMeta.MinTime < metas[j].BlockMeta.MinTime
+	}), true)
+}
+
+func BenchmarkIterBlockMetas(b *testing.B) {
```
bwplotka: They are much more interesting functions we could profile then this - it is fairly simple, but no harm in keeping this (:

pkg/shipper/shipper_test.go
obiesmans: Ok.

pkg/shipper/shipper_test.go
obiesmans: Yes I went overboard with this ;)

PR #1682, opened 2019-10-24
fix: oss range in 32bit os

mengskysama:


bwplotka:
Thanks, but wait for @wujinhu to approve as well. (:

GiedriusS:
Thanks to all! ❤

PR #1684, opened 2019-10-25
Adopter blablacar

obiesmans:
Add [Blablacar](https://www.blablacar.com/) as an adopter.


bwplotka:
:+1:

PR #1680, opened 2019-10-24
.*: Add new http-grace-period flag

kakkoyun:
This PR adds a new CLI flag to components which serve HTTP, called `--http-grace-period`.
Also refactors existing code and introduces a new HTTP server package, to increase the clarity of the `run.Group` scheduling.

cc @bwplotka @FUSAKLA 

Signed-off-by: Kemal Akkoyun <kakkoyun@gmail.com>

* [x] I added CHANGELOG entry for this change.
* [ ] Change is not relevant to the end-user.

## Changes

* Adds new `--http-grace-period` CLI flag to components which serve HTTP.
* Removes `scheduleHTTPServer` function and introduces a new package to handle HTTP serving.

## Verification

* `make local-test`
* `./scripts/quickstart.sh`

PR #1689, opened 2019-10-28
server: shutdown the actual http server

vmg:
Hi friends! It seems like #1680 introduced a regression that was preventing Thanos from shutting down cleanly. This fixes it. I did not add a CHANGELOG entry because, well, it's a regression which we fixed early. Let me know if you need me to add one.

## Changes

Commit message inline for your convenience:

```
PR #1680 introduced graceful handling for the HTTP server in Thanos, but
the graceful `Shutdown` call was being performed on an `http.Server`
instance that was *not* running at all. The actual server that was
listening for requests was started through `http.Serve`, so there was no
reference to the server struct that we could use to shut it down. This
was causing all of Thanos to freeze after receiving an exit signal,
because the run-group for the HTTP server would never finalize.

This seems like an oversight because the `(*Server).srv` field was being
properly initialized with an HTTP server. Fix this by calling
`ListenAndServe` on our initialized server.

Signed-off-by: Vicent Marti <vmg@strn.cat>
```

## Verification

Manually verified that `thanos` can now be shutdown with a shutdown signal.

cc @antonio

### DIFF
--- a/pkg/server/http.go
+++ b/pkg/server/http.go
@@ -2,7 +2,6 @@ package server
 
 import (
 	"context"
-	"net"
 	"net/http"
 	"net/http/pprof"
 	"time"
@@ -22,9 +21,8 @@ type Server struct {
 	comp   component.Component
 	prober *prober.Prober
 
-	mux      *http.ServeMux
-	srv      *http.Server
-	listener net.Listener
+	mux *http.ServeMux
+	srv *http.Server
 
 	opts options
 }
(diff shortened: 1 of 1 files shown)


kakkoyun:
@vmg Thanks a lot for fixing that. I guess I made a mistake while trying to ensure parity with the old implementation.

vmg:
:bow::heart:

PR #1697, opened 2019-10-29
shipper check number of chunk files in remote match local block

yobro:
## Changes

When shipper uploads blocks, it is possible that not all chunk files will upload successfully, but the metafile and block will exist in block storage when a failure occurs, such as network or block storage outage. In such a case, when the shipper syncs again it will not see the failed block in its 'uploaded blocks' list, but will see the block and meta file in block storage and assume it has already been uploaded.

Added logic to the shipper sync method to check that the same number of chunk files exist in block storage as well as the existence of the block and meta file. 

## Verification

1. ran: make test-local
2. ran: 2 prometheus, 2 sidecars, and block storage service. Then removed a chunk file from uploaded block and verified the file was re-uploaded.

### DIFF
--- a/pkg/shipper/shipper.go
+++ b/pkg/shipper/shipper.go
@@ -296,7 +296,22 @@ func (s *Shipper) Sync(ctx context.Context) (uploaded int, err error) {
 			return errors.Wrap(err, "check exists")
 		}
 		if ok {
-			return nil
+			// Check that same number of chunk files match meta file, this will cause partially failed transfers to re-upload
+			seqFiles, err := ioutil.ReadDir(path.Join(s.dir, m.ULID.String(), block.ChunksDirname))
+			if err != nil {
+				return errors.Wrap(err, "counting local chunk files")
+			}
+			if err = s.bucket.Iter(ctx, path.Join(m.ULID.String(), block.ChunksDirname), func(string) error {
+				if len(seqFiles) > 0 {
+					seqFiles = seqFiles[1:]
+				}
+				return nil
+			}); err != nil {
+				return errors.Wrap(err, "counting remote chunk files")
+			}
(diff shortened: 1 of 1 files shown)


yobro:
Yes I see now that the meta file is uploaded last, however I would argue that this change would provide some level of self healing in case of a failure during compaction. For such an infrequent operation it would have no impact, but would help maintain data consistency.

bwplotka:
> self-healing in case of a failure during compaction. 

You mean during sidecar upload rather?

I totally understand this point, but there is some harm to add it if it's totally unnecessary. We have chosen to do "full block" check by checking meta.json as it is uploaded as a last one. 

Some partial ensuring like this:

* increases some net traffic a bit
* does not check everything (e.g index file)
* does not check if chunk or any file is not partially uploaded (!) 
* is not consistent with many other upload cases we do (compactor, repair processes)

In my eyes, overall if uploading meta.json is not enough or there is a bug: it will hit us in many other places anyway, so we want to rely on one safeguard and make it strong. If we start to add small safeguards in many places we:
* Maybe loose focus on the main safeguard as there are extra checks so yolo (: 
* Get lost what is checked and where, so it's prone to bugs.

Super silly comparison but in a similar way, we don't log the same error 3 times, just in case we did not log it in this place, and just in case we did not handle error in another, we log here as well, etc (: 

Sorry for being super strict, but this is my opinion in order to keep the project clean and understandable! (:

yobro:
I suppose this cannot happen though since compactor won’t delete anything until after upload completes anyway. Will close request.

yobro:
Sorry missed your last comment. Yes you are right. No worries.

GiedriusS:
I probably ran into that issue that you've mentioned @bwplotka, thanks for the comments here - makes sense and I agree.

Issue #1657, opened 2019-10-16
Add UI (e.g bucket web) to Compactor


bwplotka:
I would like to propose an optional UI to the compactor, which will render what blocks we have access to same as `bucket web`.

AC:
* Compactor having UI with similar view as `thanos bucket web`

jojohappy:
What is the difference between `thanos bucket web`?

squat:
I think the main difference is that it runs as a daemon and you can access it whenever. I am a big fan of surfacing this UI easily ((((:

bwplotka:
Yes and everyone needs compactor anyway, so it would fit perfectly.

squat:
ack, I started hacking on this last week, I can put up a PR soon (:

bwplotka:
:heart:

vsxen:
how to build gstatic all js&cs file into thanos binary?

vsxen:
hi @squarejaw    how to build gstatic js&cs file into thanos binary , can you help me ?

bwplotka:
Not yet done and in progress.

bwplotka:
@squat

PR #1702, opened 2019-10-31
cmd/thanos/bucket: expose metrics

squat:
Currently, the bucket web command generates and registers metrics but
they are never actually exposed. This commit ensures that the metrics
are exposed and leverages the recently created server package for
consistency.

This cleanup also helps prepare for the upcoming changes for
https://github.com/thanos-io/thanos/issues/1657.

Changelog entry is already covered by #1680 

## Verification
Launched bucket web component before and after change and ensured that metrics are exposed.

### DIFF
--- a/pkg/ui/bucket.go
+++ b/pkg/ui/bucket.go
@@ -6,8 +6,8 @@ import (
 	"time"
 
 	"github.com/go-kit/kit/log"
-	"github.com/prometheus/common/route"
 	extpromhttp "github.com/thanos-io/thanos/pkg/extprom/http"
+	"github.com/thanos-io/thanos/pkg/server"
 )
 
 // Bucket is a web UI representing state of buckets as a timeline.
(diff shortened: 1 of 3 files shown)


pkg/ui/bucket.go
```diff
@@ -30,13 +30,13 @@ func NewBucketUI(logger log.Logger, label string) *Bucket {
...
-func (b *Bucket) Register(r *route.Router, ins extpromhttp.InstrumentationMiddleware) {
+func (b *Bucket) Register(s server.Server, ins extpromhttp.InstrumentationMiddleware) {
 	instrf := func(name string, next func(w http.ResponseWriter, r *http.Request)) http.HandlerFunc {
 		return ins.NewHandler(name, http.HandlerFunc(next))
 	}
 
-	r.Get("/", instrf("root", b.root))
-	r.Get("/static/*filepath", instrf("static", b.serveStaticAsset))
+	s.Handle("/", instrf("root", b.root))
+	s.Handle("/static/*filepath", instrf("static", b.serveStaticAsset))
```
metalmatze: Why change these specifically?

squat: :/ because the `Server` type doesn't have a `Get` method, only a `Handle` and to be consistent internally, it simplifies things to re-use the server package.

kakkoyun: `Server` is just a wrapper around `http.Server` that why it resonates with its methods. However, since `prometheus/common/route.Router` implements `http.Handler` you can pass it down to the `Server`, if you want to preserve behaviour.

squat: I guess we could do that but I think it would just add an extra layer of complexity

Issue #1568, opened 2019-09-25
counter reset on aggregation boundaries


alfred-landrum:
My colleague @aponjavic and I have a question about the counter aggregation implementation: We were wondering if the following scenario could occur in real world usage:
- 2 consecutive aggregation chunks, A1 and A2 are created
- a counter reset occurs at the boundary between A1 and A2
- the counter state for A2 is *greater* than the last raw for A1  

In this scenario, a `CounterSeriesIterator` wouldn't treat the A1 -> A2 transition as a reset. I can force this scenario by modifying the unit tests in downsample_test.go; for example, using raw values of:

`{t: 5, v: 95}, {t: 15, v: 100}, {t: 25, v: 97}, {t: 35, v: 107}`

and forcing aggregation at 20 time unit boundaries causes 2 aggregation chunks, whose values are:

`[]downsample.sample{downsample.sample{t:15, v:100}, downsample.sample{t:15, v:100}, downsample.sample{t:35, v:107}, downsample.sample{t:35, v:107}}`

Since the value of 107 is > 100, it seems that any iteration over these wouldn't track the reset that occurred between time 15 and time 25.

If this can actually occur, we imagine that adding a "first raw" sample point in A2 for {t: 25: v: 97}, and then ensuring its passed through the aggregation iterations, would address the issue.

Please let us know if this is a real possibility, or if we've misunderstood how counter downsampling works!

brian-brazil:
I don't think that's the output that `downsampleRaw` would produce, it'd be `{t: 5, v: 95}, {t: 15, v: 100}, {t: 25, v: 197}, {t: 35, v: 107}`

I did notice that https://github.com/thanos-io/thanos/blob/99bc7d29ffe836224afcd6ed5f0efe91d3458035/pkg/compact/downsample/downsample.go#L594 may cause a call stack overflow, a for loop would be better.

alfred-landrum:
@brian-brazil : thanks for looking at this. Are the points you posted from a single aggregation chunk? If not, where did the aggregation boundary fall?

brian-brazil:
Oops, got my timestamps wrong. That's two counter chunk, with two samples each.

brian-brazil:
I'm not seeing that in the code, https://github.com/thanos-io/thanos/blob/bd9aa1b4be3bb5d841cb7271c29d02ebb5eb5168/pkg/compact/downsample/downsample.go#L582 should catch it. Maybe a unittest would demonstrate better?

[...]

PR #1709, opened 2019-11-03
store the first raw value of a chunk during downsampling

alfred-landrum:
As discussed in #1568, storing only the last raw value
of a chunk will lose a counter reset when:
a) the reset occurs at a chunk boundary, and
b) the last raw value of the earlier chunk is less than
the first aggregated value of the later chunk.

This commit stores the first raw value of a chunk during
the initial raw aggregation, and retains it during
subsequent aggregations. This is similar to the existing
handling for the last raw value of a chunk.

With this change, when counterSeriesIterator iterates over
a chunk boundary, it will see the last raw value of the
earlier chunk, then the first raw value of the later chunk,
and then the first aggregated value of the later chunk. The
first raw value will always be less than or equal to the
first aggregated value, so the only difference in
counterSeriesIterator's output will be the possible detection
of a reset and an extra sample after the chunk boundary.

Fixes: https://github.com/thanos-io/thanos/issues/1568

Signed-off-by: Alfred Landrum <alfred@leakybucket.org>


pkg/compact/downsample/downsample.go
```diff
@@ -306,14 +312,17 @@ func downsampleRaw(data []sample, resolution int64) []chunks.Meta {
...
-	)
+
+	mint, maxt := data[0].t, data[len(data)-1].t
+	// We assume a raw resolution of 1 minute. In practice it will often be lower
+	// but this is sufficient for our heuristic to produce well-sized chunks.
+	numChunks := targetChunkCount(mint, maxt, 1*60*1000, resolution, len(data))
+	return downsampleRawLoop(data, resolution, numChunks)
+}
+
+func downsampleRawLoop(data []sample, resolution int64, numChunks int) []chunks.Meta {
```
bwplotka: ~hm, why extracting? (: ~

Ok, for test purposes, fair (:

pkg/compact/downsample/downsample.go
```diff
@@ -289,7 +289,13 @@ func (b *aggrChunkBuilder) add(t int64, aggr *aggregator) {
 	b.added++
 }
 
-func (b *aggrChunkBuilder) finalizeChunk(lastT int64, trueSample float64) {
+func (b *aggrChunkBuilder) firstRawSample(firstT int64, trueSample float64) {
```
bwplotka: those functions are quite shallow, and really the same. Can we maybe just inline with the comment? We use it twice, sure, but if we would inline them it might be even clearer?

pkg/compact/downsample/downsample.go
```diff
@@ -379,18 +390,20 @@ func downsampleBatch(data []sample, resolution int64, add func(int64, *aggregato
...
-		res       = make([]chunks.Meta, 0, numChunks)
-		batchSize = len(chks) / numChunks
-	)
+	numChunks := targetChunkCount(mint, maxt, inRes, outRes, numSamples)
+	return downsampleAggrLoop(chks, buf, outRes, numChunks)
+}
+
+func downsampleAggrLoop(chks []*AggrChunk, buf *[]sample, resolution int64, numChunks int) ([]chunks.Meta, error) {
+	// We downsample aggregates only along chunk boundaries. This is required for counters
+	// to be downsampled correctly since a chunks' last counter value is the true last value
```
brian-brazil: chunk's

pkg/compact/downsample/downsample_test.go
```diff
@@ -23,6 +23,124 @@ import (
...
+	tests := []test{
+		{
+			// In this test case, counter resets occur at the
+			// boundaries between the t=49,t=99 and t=99,t=149
+			// windows, and the values in the t=49, t=99, and
+			// t=149 windows are high enough that the resets
+			// will only be accounted for if the first raw value
+			// of a chunk is maintained during aggregation.
+			// See #1568 for more details.
+			[]sample{
```
brian-brazil: Explicit field names here would be more readable

pkg/compact/downsample/downsample.go
alfred-landrum: Yes, it was for testing :). The `targetChunkCount` call makes crafting the right set of test data a little tricky otherwise.

alfred-landrum:
@bwplotka : regarding your question: I didn't observe this directly: my colleague @aponjavic spotted the potential issue as we were studying how Thanos implements downsampling. So I don't have an easy setup that repros the issue & the fix.

pkg/compact/downsample/downsample.go
GiedriusS: Maybe but IMHO this split up is clear too since literally the function's name tells you what's happening :smile: up to you.

PR #1717, opened 2019-11-05
pkg/receive: eliminate duplicate tracing

squat:
The receive component currently has duplicate spans for requests handled
by its HTTP server. This is due to the fact that the server is
instrumented once using the global singleton tracer and then a second
time with the tracer configured via the CLI. This commit eliminates the
duplicate instrumentation via the global singleton in favor of the
explicit specified tracer, which is consistent with the practice of
other Thanos components.

Example of duplicate span:
image

Note that:
* both `/receive HTTP[server]` and `POST /api/v1/receive` represent the same operation; and that
* the span created by the global singleton, `POST /api/v1/receive` is erroneously setting the HTTP status code to `0`.

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -31,7 +31,6 @@ require (
 	github.com/oklog/run v1.0.0
 	github.com/oklog/ulid v1.3.1
 	github.com/olekukonko/tablewriter v0.0.1
-	github.com/opentracing-contrib/go-stdlib v0.0.0-20190519235532-cf7a6c988dc9
 	github.com/opentracing/basictracer-go v1.0.0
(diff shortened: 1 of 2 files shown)


squat:
@metalmatze the tracing configuration is also typically passed explicitly, e.g. https://github.com/thanos-io/thanos/blob/master/cmd/thanos/query.go#L382,
https://github.com/thanos-io/thanos/blob/master/cmd/thanos/query.go#L400.

The receive component was the only component with a reference to the globaltracer, however this seems to have been left over from the initial implementation of the receiver.

PR #1721, opened 2019-11-05
cmd/thanos/receive: reduce WAL replays at startup

squat:
Every time thanos receive is started, it has to replay the WAL three
times, namely:
1. open the TSDB;
2. close the TSDB; open the ReadOnly TSDB and Flush; and
3. open the TSDB

These WAL replays can take a very long time if the WAL has lots of data.
With the fix from #1654, the third time will be instantaneous because
the WAL will be empty. That still leaves two potentially long WAL
replays. We can cut this down to just one long replay if we do the
following operations instead:
1. with a closed TSDB, open the ReadOnly TSDB and Flush; and
2. open the TSDB

Now, the second step will be a fast replay because the WAL is empty,
leaving just one potentially expensive WAL replay.

This commit eliminates explicit opening of the writable TSDB during
startup, and instead opens it after flushing the read-only TSDB.

cc @metalmatze @bwplotka @brancz

### DIFF
--- a/cmd/thanos/receive.go
+++ b/cmd/thanos/receive.go
@@ -230,6 +230,9 @@ func runReceive(
 			defer close(dbReady)
 			defer close(uploadC)
 
+			// Before actually starting, we need to make sure the
+			// WAL is flushed. The WAL is flushed after the
+			// hashring is loaded.
 			db := receive.NewFlushableStorage(
 				dataDir,
 				log.With(logger, "component", "tsdb"),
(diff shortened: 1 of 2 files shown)


brancz:
Not sure if legitimately or not, but CI is killing the job because nothing new in logs for 10mins.

squat:
Saw that. I am running locally and it’s happy. Can we re-kick CI?

brancz:
Re-ran a few times now, it seems we just have to extend the timeout if that's possible on circleci.

squat:
Ill do some digging

brancz:
Nicely found!

PR #1727, opened 2019-11-06
cmd/thanos/receive: avoid deadlock

squat:
While debugging #1721, I found that when thanos receive bails, there is
a race in a select statement, where the non-returning branch may be
chosen. This branch will deadlock if selected twice because the channel
reader has already exited. The way to prevent this is by checking if
we need to exit on every loop.

cc @brancz @bwplotka


brancz:
Oops...

PR #1746, opened 2019-11-14
docs: Query Memory Limits proposal

ppanyukov:
Proposal to introduce query memory limits in Querier (Q) and Store Gateway (SG) components.

~I have an implementation of this design almost ready to submit for a review as a follow up / in parallel if people generally agree on this feature.~

Draft implementation of this design is here: https://github.com/thanos-io/thanos/pull/1747

### DIFF
--- a/docs/proposals/201911_thanos_query_mem_limits.md
+++ b/docs/proposals/201911_thanos_query_mem_limits.md
@@ -1,0 +1,299 @@
+---
+title: Thanos Query Memory Limits
+type: proposal
+menu: proposals
+status: pending
+owner: ppanyukov
+---
+
+### Related Tickets
+
+Quite a few related to OOM and improving memory utilisation:
+
+* https://github.com/thanos-io/thanos/issues/448
+* https://github.com/thanos-io/thanos/issues/1625 (previously failed attempt and PR)
+* https://github.com/thanos-io/thanos/issues/1649 (discussion with ideas)
+* https://github.com/thanos-io/thanos/issues/1705 (another tracking issue)
+
+
+## Summary
+
+This document describes the motivation and design of memory accounting and limits for query execution in Thanos. This affects both Querier (Q) and Store Gateway (SG) components.
+
+## Motivation
+
+**TL;DR: Availability is the main issue.**
+
+Both Store Gateway (SG) and Querier (Q) OOMs at query time. Why is this a problem?
+
+* Thanos is used by multiple customers (users).
+* One query issued by one customer can OOM SG or Q, which kills all other queries.
+* This same query gets retried many times.
+* Effectively becoming a DOS attack.
+* Making the entire service completely unavailable to everyone.
+
+Conclusion:
+* Availability of the service is the main issue.
+
+## Top-level idea
+
+Introduce memory accounting and limits in SG and Q components. This will ensure that one query does not take over all resources. Combined with the limit on the number of concurrent queries (already available) should result in a much better multi-user story.
+
+There are two main limits in both Q and SG components. They are conceptually very similar.
+
+* Querier
+    * (x) Query Pipe Limit: when querying we can contact multiple StoreAPIs, limit the size of bytes we receive and keep from each source.
+    * (y) Query Map Size: limit the number of bytes that Q can merge from all sources (StoreAPI), cancel requests if the merged result becomes too big
+* Store Gateway:
+    * (x) Query Pipe Limit: when querying we can read from multiple blocks, limit the size of bytes we receive and keep from each block.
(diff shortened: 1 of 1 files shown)


docs/proposals/201911_thanos_query_mem_limits.md
```diff
@@ -0,0 +1,299 @@
...
+The approach to this is using sizes of structs and objects that make up the query results. Code is like this:
+
+**Querier: receiving frames**
+
+```golang
+// Used in startStreamSeriesSet(), pkg/store/proxy.go, 
+getApproxSize := func(r *storepb.SeriesResponse) int64 {
+    size := int64(0)
+    size += int64(unsafe.Sizeof(storepb.SeriesResponse{}))
+    size += int64(len(r.XXX_unrecognized))
```
bwplotka: Didn't we disable those fields?

ppanyukov: We did, but proposal was written *before* that PR was merged :)

Obviously, `XXX_unrecognized` won't be taken into account now.

docs/proposals/201911_thanos_query_mem_limits.md
```diff
@@ -0,0 +1,299 @@
...
+    for _, id := range ps {
+        frameSize := getFrameSize()
+
+        // check with local limiter first
+        if err := localLimiter.Add(frameSize); err != nil {
+            return err
+        }
+
+        // now add to total
+        if err := queryTotalLimiter.Add(frameSize); err != nil {
```
bwplotka: Postings are not the only source of memory consumption. I assume we will do similar for series.

In fact while having series references we can be smart as we know the number of the chunks we are going fetch and we can add this before fetch even (:

docs/proposals/201911_thanos_query_mem_limits.md
```diff
@@ -0,0 +1,299 @@
...
+
+func blockSeries(..., queryTotalLimiter) {
+    // Main memory users are lebels and aggrChunk. But not all. Only showing main parts
+
+    // Have local limiter here.
+    localLimiter limit.Limiter = limit.NewLimiterNoLock(limit.QueryPipeLimit())
+    
+    ps, err := indexr.ExpandedPostings(matchers)
+    for _, id := range ps {
+        frameSize := getFrameSize()
```
bwplotka: What's frame here?

docs/proposals/201911_thanos_query_mem_limits.md
```diff
@@ -0,0 +1,299 @@
...
+
+        // check with local limiter first
+        if err := localLimiter.Add(frameSize); err != nil {
```
bwplotka: Does it need to be a separate function to the totalLimiter? (: I think we can do all in one function

metalmatze:
How do we expose these changes to our users? I suppose by flags, but what do they look like?

ppanyukov:
So, there is an accompanying draft implementation PR as well, it might be easier to look at the real code there, as most code in this doc is for "illustration purpose only" to give a broad idea of how I propose things to work.

WRT exposing to users: the draft implementation currently uses env vars as the least invasive mechanism. Env vars also allow running built versions of Thanos with and without this feature without having to modify scripts etc. An open question of course. // @metalmatze

docs/proposals/201911_thanos_query_mem_limits.md
ppanyukov: It's whatever stuff (labels, chunks etc) we create and allocate memory for while doing each posting.

docs/proposals/201911_thanos_query_mem_limits.md
ppanyukov: There are so many ways to do this limiter/counter. Conceptually, they would all do this kind of thing though: 1) add to local limiter and check;  2) add to total limiter and check.

Unless I misunderstood the question?

docs/proposals/201911_thanos_query_mem_limits.md
ppanyukov: It would be nice if we could do some magic one-liner calculation. Probably best to look at the real code in https://github.com/thanos-io/thanos/pull/1747/files ?

bwplotka:
Hi! After some consideration, I think the major concern about this is that those limits (e.g 100MB per query) can use much more memory due to the nature of Go memory management (e.g slice reallocation always doubled capacity and  GC eventual release). This can massively surprise users and effectively make it even harder to do any reasonable capacity planning. While I was initially positive on this, after some discussions I might think if we cannot be precise, and it requires from user to do load tests anyway, we might want to stick to more size-abstract limits, like samples/chunks/series etc.

Issue #1659, opened 2019-10-17
Disable dependabot and add process for updating all deps after each release 


bwplotka:
We are kind of behind dependencies and being spammed by 
It would be nice to configure dependabot to do all in one PR. I can't see such an option so I have created a support ticket.

Should it be a release shepherd responsibility?

AC:
* After every release, we upgrade all our dependencies.

cc @brancz @domgreen @GiedriusS @metalmatze @adrien-f @FUSAKLA opinions?

brancz:
I do what is it that is keeping us from merging these PRs? It seems like continuously doing this is a better idea then at once. Do we not have the confidence? Is it too much?

Generally speaking this still seems like a good idea in theory, I want to understand what's making it not useful in practice (and/or if we _could_ make it useful :slightly_smiling_face: ).

bwplotka:
Sure, so problems with current approach:

* Each merge of such dep PR creates conflict in go.mod (in most cases). Bot re submitting another dep PRs in this case, but we have to wait for CI etc. At least 20m wait time for each PR. For ~50 deps it takes in worst case it takes 16h (: Not fun.
* We can have some dependency hell with indirect deps. This means that one PR can update one dep, but in another we have to downgrade. It might be not very efficient to takle on each dep basis

brancz:
It's probably worth giving the dependabot folks this feedback. I'd say let's make it release shepherd responsibility, that seems like a reasonable thing (of course one off bumps are not discouraged).

bwplotka:
Ticket sent, no response

metalmatze:
5 days later, did you get a response in the meantime?

bwplotka:
Nope: It is also Github support as it was acquired. No response at all ):

bwplotka:
Actually forget about a support ticket, look at this: https://github.com/dependabot/feedback/issues/5 :eyes: 

Will describe our use case and will check if we can help in any way.

PR #1748, opened 2019-11-15
Updated all deps.

bwplotka:
As mentioned in https://github.com/thanos-io/thanos/issues/1659 let's update our deps at once. 

Should we disable dependabot now? I will close all its PRs for now. Still, it can be nice to be updated about all new releases (:


squat:
How many dependabot PRs are we actually merging? Do you mean keeping it enabled just to be notified about releases? It's a little noisy IMO for that :p

bwplotka:
Not sure, we probably have to reduce the noise.

In terms of release notes, I double-checked Prometheus upgrade and nothing really affecting Thanos users was merged (:

PR #1756, opened 2019-11-18
Add default label "prometheus" to bucket web.

cmanzi:

## Changes
Add a default label to `thanos web bucket` of `prometheus`. This is a common external label and would make a good default for most users so that the UI will show the different `prometheus` labels by default instead of integers 1,2,3, etc.


cmanzi:
> I thought something like `replica` was more common (: At least that is what is used in the documentation, AFAICT.

Ok. Just trying to add some sort of default to help out people like me who might get confused at the lack of labeling in a default install.

PR #1761, opened 2019-11-19
cmd/thanos/rule: simplify if block

simonpasquier:

## Changes

Refactor the if loop checking whether the shipper is enabled or not to be easier to read.

## Verification

`make test-local` runs fine.

### DIFF
--- a/cmd/thanos/rule.go
+++ b/cmd/thanos/rule.go
@@ -567,13 +567,7 @@ func runRule(
 		return err
 	}
 
-	uploads := true
-	if len(confContentYaml) == 0 {
-		level.Info(logger).Log("msg", "No supported bucket was configured, uploads will be disabled")
-		uploads = false
-	}
-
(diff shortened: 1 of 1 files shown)


cmd/thanos/rule.go
```diff
@@ -604,6 +598,8 @@ func runRule(
 		}, func(error) {
 			cancel()
 		})
+	} else {
+		level.Info(logger).Log("msg", "No supported bucket was configured, uploads will be disabled")
```
bwplotka: nice, can we also fix the sentence one we are here?  (should be lower case all)
```suggestion
		level.Info(logger).Log("msg", "no supported bucket was configured, uploads will be disabled")
```

bwplotka:
Plase sign the commit (: (DCO bot)

simonpasquier:
Ah that old DCO friend :smile:

PR #1785, opened 2019-11-25
pkg/rule: retain original path for rule files

simonpasquier:
This change ensures that the /api/v1/rules endpoint returns the original path of the rule file instead of the path of the temporary file generated by Thanos ruler.

## Changes

The mapping from the original rule filenames to the temporary filenames is stored in the `github.com/thanos-io/thanos/pkg/rule.Manager` struct. 

## Verification

Added end-to-end tests.

### DIFF
--- a/test/e2e/spinup_test.go
+++ b/test/e2e/spinup_test.go
@@ -10,7 +10,7 @@ import (
 	"net"
 	"os"
 	"os/exec"
-	"path"
+	"path/filepath"
 	"strconv"
 	"syscall"
 	"testing"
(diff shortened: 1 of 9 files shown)


simonpasquier:
The CI failure doesn't seem to be related to the change.

test/e2e/spinup_test.go
```diff
@@ -304,35 +304,17 @@ receivers:
...
-func rule(http, grpc address, rules []string, am address, queryAddresses []address, queryFileSDAddresses []address) *serverScheduler {
-	return ruleWithDir(http, grpc, "", rules, am, queryAddresses, queryFileSDAddresses)
-}
-
-func ruleWithDir(http, grpc address, dir string, rules []string, am address, queryAddresses []address, queryFileSDAddresses []address) *serverScheduler {
+func rule(http, grpc address, ruleDir string, am address, queryAddresses []address, queryFileSDAddresses []address) *serverScheduler {
 	return &serverScheduler{
 		HTTP: http,
 		GRPC: grpc,
 		schedule: func(workDir string) (Exec, error) {
```
bwplotka: Much, cleaner thanks :+1:

PR #1786, opened 2019-11-25
test/e2e: start Alertmanager without clustering

simonpasquier:

## Changes

When running the end-to-end tests locally, Alertmanager may fail to start if another instance of Alertmanager is already running. This is because both instances bind to the default clustering port (9094). Since e2e tests don't rely on Alertmanager clustering, we disable clustering by setting `--cluster.listen-address=""`.

## Verification

End-to-end tests should pass as before.


bwplotka:
:+1:

PR #1789, opened 2019-11-26
Improve store timeouts

IKSIN:
We want to increase stability Thanos infrastructure with many stores/sidecars (49 in our case) when some stores respond slowly.

## Changes

* [ ] Fix and improve TestProxyStore_SeriesSlowStores:
  * [ ] Fixed broken test
  * [ ] Add some use-cases
  * [ ]  check on elapsed time on get data from stores.
* [ ] Change logic in ProxyStore:
  * [ ] store response timeout moved from MergedSet.Next() function to goroutine in startStreamSeriesSet to check timeouts in parallel
  * [ ] add ProxyStore metrics

More details in comments.

## Verification

* [ ] Tests
* [ ] On production in our system with 49 store backends already 2 weeks.

### DIFF
--- a/pkg/store/proxy.go
+++ b/pkg/store/proxy.go
@@ -411,18 +411,12 @@ func startStreamSeriesSet(
 			frameTimeoutCtx, cancel := frameCtx(s.responseTimeout)
 			defer cancel()
 			var rr *recvResponse
-
-			var err error
 			select {
 			case <-ctx.Done():
-				close(done)
-				err = errors.Wrap(ctx.Err(), fmt.Sprintf("failed to receive any data from %s", s.name))
-				s.handleErr(err)
+				s.handleErr(errors.Wrapf(ctx.Err(), "failed to receive any data from %s", s.name), done)
 				return
 			case <-frameTimeoutCtx.Done():
-				close(done)
-				err = errors.Wrap(frameTimeoutCtx.Err(), fmt.Sprintf("failed to receive any data in %s from %s", s.responseTimeout.String(), s.name))
-				s.handleErr(err)
+				s.handleErr(errors.Wrapf(ctx.Err(), "failed to receive any data in %s from %s", s.responseTimeout.String(), s.name), done)
 				return
 			case rr = <-rCh:
 			}
@@ -434,8 +428,7 @@ func startStreamSeriesSet(
 
 			if rr.err != nil {
 				wrapErr := errors.Wrapf(rr.err, "receive series from %s", s.name)
-				s.handleErr(wrapErr)
-				close(done)
+				s.handleErr(wrapErr, done)
 				return
 			}
 			numResponses++
@@ -450,8 +443,10 @@ func startStreamSeriesSet(
 	return s
 }
 
-func (s *streamSeriesSet) handleErr(err error) {
+func (s *streamSeriesSet) handleErr(err error, done chan struct{}) {
+	defer close(done)
 	s.closeSeries()
+
 	if s.partialResponse {
 		level.Warn(s.logger).Log("err", err, "msg", "returning partial response")
 		s.warnCh.send(storepb.NewWarnSeriesResponse(err))
--- a/pkg/store/proxy.go
+++ b/pkg/store/proxy.go
@@ -395,12 +395,12 @@ func startStreamSeriesSet(
 		}()
 
 		rCh := make(chan *recvResponse)
-		recvCancel := make(chan bool)
+		done := make(chan struct{})
 		go func() {
 			for {
 				r, err := s.stream.Recv()
 				select {
-				case <-recvCancel:
+				case <-done:
 					close(rCh)
 					return
 				case rCh <- &recvResponse{r: r, err: err}:
@@ -415,12 +415,12 @@ func startStreamSeriesSet(
 			var err error
 			select {
 			case <-ctx.Done():
-				close(recvCancel)
+				close(done)
 				err = errors.Wrap(ctx.Err(), fmt.Sprintf("failed to receive any data from %s", s.name))
 				s.handleErr(err)
 				return
 			case <-frameTimeoutCtx.Done():
-				close(recvCancel)
+				close(done)
 				err = errors.Wrap(frameTimeoutCtx.Err(), fmt.Sprintf("failed to receive any data in %s from %s", s.responseTimeout.String(), s.name))
 				s.handleErr(err)
 				return
@@ -428,14 +428,14 @@ func startStreamSeriesSet(
 			}
 
 			if rr.err == io.EOF {
-				close(recvCancel)
+				close(done)
 				return
 			}
 
 			if rr.err != nil {
 				wrapErr := errors.Wrapf(rr.err, "receive series from %s", s.name)
 				s.handleErr(wrapErr)
-				close(recvCancel)
+				close(done)
 				return
 			}
 			numResponses++
--- a/pkg/store/proxy.go
+++ b/pkg/store/proxy.go
@@ -198,12 +198,7 @@ func newRespCh(ctx context.Context, buffer int) (*ctxRespSender, <-chan *storepb
 }
 
 func (s ctxRespSender) send(r *storepb.SeriesResponse) {
-	select {
-	case <-s.ctx.Done():
-		return
-	case s.ch <- r:
-		return
-	}
+	s.ch <- r
 }
 
 // Series returns all series for a requested time range and label matcher. Requested series are taken from other
@@ -348,6 +343,21 @@ type streamSeriesSet struct {
 	closeSeries     context.CancelFunc
 }
 
+type recvResponse struct {
+	r   *storepb.SeriesResponse
+	err error
+}
+
+func startFrameCtx(responseTimeout time.Duration) (context.Context, context.CancelFunc) {
+	frameTimeoutCtx := context.Background()
+	var cancel context.CancelFunc
+	if responseTimeout != 0 {
+		frameTimeoutCtx, cancel = context.WithTimeout(frameTimeoutCtx, responseTimeout)
+		return frameTimeoutCtx, cancel
+	}
+	return frameTimeoutCtx, nil
+}
+
 func startStreamSeriesSet(
 	ctx context.Context,
 	logger log.Logger,
@@ -384,14 +394,34 @@ func startStreamSeriesSet(
 			}
 		}()
 		for {
-			r, err := s.stream.Recv()
+			frameTimeoutCtx, cancel := startFrameCtx(s.responseTimeout)
+			if cancel != nil {
+				defer cancel()
+			}
+			rCh := make(chan *recvResponse, 1)
+			var rr *recvResponse
+			go func() {
+				r, err := s.stream.Recv()
+				rCh <- &recvResponse{r: r, err: err}
+			}()
 
-			if err == io.EOF {
+			select {
+			case <-ctx.Done():
+				s.timeoutHandling(true, ctx)
+				return
+			case <-frameTimeoutCtx.Done():
+				s.timeoutHandling(false, frameTimeoutCtx)
+				return
+			case rr = <-rCh:
+			}
+			close(rCh)
+
+			if rr.err == io.EOF {
 				return
 			}
 
-			if err != nil {
-				wrapErr := errors.Wrapf(err, "receive series from %s", s.name)
+			if rr.err != nil {
+				wrapErr := errors.Wrapf(rr.err, "receive series from %s", s.name)
 				if partialResponse {
 					s.warnCh.send(storepb.NewWarnSeriesResponse(wrapErr))
 					return
@@ -402,59 +432,40 @@ func startStreamSeriesSet(
 				s.errMtx.Unlock()
 				return
 			}
-
 			numResponses++
 
-			if w := r.GetWarning(); w != "" {
+			if w := rr.r.GetWarning(); w != "" {
 				s.warnCh.send(storepb.NewWarnSeriesResponse(errors.New(w)))
 				continue
 			}
-
-			select {
-			case s.recvCh <- r.GetSeries():
-				continue
-			case <-ctx.Done():
-				return
-			}
-
+			s.recvCh <- rr.r.GetSeries()
 		}
 	}()
 	return s
 }
 
+func (s *streamSeriesSet) timeoutHandling(isQueryTimeout bool, ctx context.Context) {
+	var err error
+	if isQueryTimeout {
+		err = errors.Wrap(ctx.Err(), fmt.Sprintf("failed to receive any data from %s", s.name))
+	} else {
+		err = errors.Wrap(ctx.Err(), fmt.Sprintf("failed to receive any data in %s from %s", s.responseTimeout.String(), s.name))
+	}
+	s.closeSeries()
+	if s.partialResponse {
+		level.Warn(s.logger).Log("err", err, "msg", "returning partial response")
+		s.warnCh.send(storepb.NewWarnSeriesResponse(err))
+		return
+	}
+	s.errMtx.Lock()
+	s.err = err
+	s.errMtx.Unlock()
+}
+
 // Next blocks until new message is received or stream is closed or operation is timed out.
 func (s *streamSeriesSet) Next() (ok bool) {
-	ctx := s.ctx
-	timeoutMsg := fmt.Sprintf("failed to receive any data from %s", s.name)
-
-	if s.responseTimeout != 0 {
-		timeoutMsg = fmt.Sprintf("failed to receive any data in %s from %s", s.responseTimeout.String(), s.name)
-
-		timeoutCtx, done := context.WithTimeout(s.ctx, s.responseTimeout)
-		defer done()
-		ctx = timeoutCtx
-	}
-
-	select {
-	case s.currSeries, ok = <-s.recvCh:
-		return ok
-	case <-ctx.Done():
-		// closeSeries to shutdown a goroutine in startStreamSeriesSet.
-		s.closeSeries()
-
-		err := errors.Wrap(ctx.Err(), timeoutMsg)
-		if s.partialResponse {
-			level.Warn(s.logger).Log("err", err, "msg", "returning partial response")
-			s.warnCh.send(storepb.NewWarnSeriesResponse(err))
-			return false
-		}
-		s.errMtx.Lock()
-		s.err = err
-		s.errMtx.Unlock()
-
-		level.Warn(s.logger).Log("err", err, "msg", "partial response disabled; aborting request")
-		return false
-	}
+	s.currSeries, ok = <-s.recvCh
+	return ok
 }
 
 func (s *streamSeriesSet) At() ([]storepb.Label, []storepb.AggrChunk) {
--- a/pkg/store/proxy.go
+++ b/pkg/store/proxy.go
@@ -348,14 +348,14 @@ type recvResponse struct {
 	err error
 }
 
-func startFrameCtx(responseTimeout time.Duration) (context.Context, context.CancelFunc) {
+func frameCtx(responseTimeout time.Duration) (context.Context, context.CancelFunc) {
 	frameTimeoutCtx := context.Background()
 	var cancel context.CancelFunc
 	if responseTimeout != 0 {
 		frameTimeoutCtx, cancel = context.WithTimeout(frameTimeoutCtx, responseTimeout)
 		return frameTimeoutCtx, cancel
 	}
-	return frameTimeoutCtx, nil
+	return frameTimeoutCtx, func() {}
 }
 
 func startStreamSeriesSet(
@@ -393,43 +393,49 @@ func startStreamSeriesSet(
 				emptyStreamResponses.Inc()
 			}
 		}()
-		for {
-			frameTimeoutCtx, cancel := startFrameCtx(s.responseTimeout)
-			if cancel != nil {
-				defer cancel()
-			}
-			rCh := make(chan *recvResponse, 1)
-			var rr *recvResponse
-			go func() {
-				r, err := s.stream.Recv()
-				rCh <- &recvResponse{r: r, err: err}
-				close(rCh)
-			}()
 
+		rCh := make(chan *recvResponse)
+		recvCancel := make(chan bool)
+		go func() {
+			for {
+				r, err := s.stream.Recv()
+				select {
+				case <-recvCancel:
+					close(rCh)
+					return
+				case rCh <- &recvResponse{r: r, err: err}:
+				}
+			}
+		}()
+		for {
+			frameTimeoutCtx, cancel := frameCtx(s.responseTimeout)
+			defer cancel()
+			var rr *recvResponse
+
+			var err error
 			select {
 			case <-ctx.Done():
-				s.timeoutHandling(true, ctx)
+				close(recvCancel)
+				err = errors.Wrap(ctx.Err(), fmt.Sprintf("failed to receive any data from %s", s.name))
+				s.handleErr(err)
 				return
 			case <-frameTimeoutCtx.Done():
-				s.timeoutHandling(false, frameTimeoutCtx)
+				close(recvCancel)
+				err = errors.Wrap(frameTimeoutCtx.Err(), fmt.Sprintf("failed to receive any data in %s from %s", s.responseTimeout.String(), s.name))
+				s.handleErr(err)
 				return
 			case rr = <-rCh:
 			}
 
 			if rr.err == io.EOF {
+				close(recvCancel)
 				return
 			}
 
 			if rr.err != nil {
 				wrapErr := errors.Wrapf(rr.err, "receive series from %s", s.name)
-				if partialResponse {
-					s.warnCh.send(storepb.NewWarnSeriesResponse(wrapErr))
-					return
-				}
-
-				s.errMtx.Lock()
-				s.err = wrapErr
-				s.errMtx.Unlock()
+				s.handleErr(wrapErr)
+				close(recvCancel)
 				return
 			}
 			numResponses++
@@ -444,13 +450,7 @@ func startStreamSeriesSet(
 	return s
 }
 
-func (s *streamSeriesSet) timeoutHandling(isQueryTimeout bool, ctx context.Context) {
-	var err error
-	if isQueryTimeout {
-		err = errors.Wrap(ctx.Err(), fmt.Sprintf("failed to receive any data from %s", s.name))
-	} else {
-		err = errors.Wrap(ctx.Err(), fmt.Sprintf("failed to receive any data in %s from %s", s.responseTimeout.String(), s.name))
-	}
+func (s *streamSeriesSet) handleErr(err error) {
 	s.closeSeries()
 	if s.partialResponse {
 		level.Warn(s.logger).Log("err", err, "msg", "returning partial response")
--- a/pkg/store/proxy.go
+++ b/pkg/store/proxy.go
@@ -403,6 +403,7 @@ func startStreamSeriesSet(
 			go func() {
 				r, err := s.stream.Recv()
 				rCh <- &recvResponse{r: r, err: err}
+				close(rCh)
 			}()
 
 			select {
@@ -414,7 +415,6 @@ func startStreamSeriesSet(
 				return
 			case rr = <-rCh:
 			}
-			close(rCh)
 
 			if rr.err == io.EOF {
 				return
--- a/pkg/store/proxy_test.go
+++ b/pkg/store/proxy_test.go
@@ -5,6 +5,7 @@ package store
 
 import (
 	"context"
+	"fmt"
 	"io"
 	"math"
 	"os"
@@ -49,6 +50,7 @@ func (c *testClient) String() string {
 func (c *testClient) Addr() string {
 	return "testaddr"
 }
+
 func TestProxyStore_Info(t *testing.T) {
 	defer leaktest.CheckTimeout(t, 10*time.Second)()
 
@@ -412,32 +414,6 @@ func TestProxyStore_Series(t *testing.T) {
 			},
 			expectedErr: errors.New("fetch series for [name:\"ext\" value:\"1\" ] test: error!"),
 		},
-		{
-			title: "use no chunk to only get labels",
-			storeAPIs: []Client{
-				&testClient{
-					StoreClient: &mockedStoreAPI{
-						RespSeries: []*storepb.SeriesResponse{
-							storeSeriesResponse(t, labels.FromStrings("a", "a")),
-						},
-					},
-					minTime:   1,
-					maxTime:   300,
-					labelSets: []storepb.LabelSet{{Labels: []storepb.Label{{Name: "ext", Value: "1"}}}},
-				},
-			},
-			req: &storepb.SeriesRequest{
-				MinTime:    1,
-				MaxTime:    300,
-				Matchers:   []storepb.LabelMatcher{{Name: "ext", Value: "1", Type: storepb.LabelMatcher_EQ}},
-				SkipChunks: true,
-			},
-			expectedSeries: []rawSeries{
-				{
-					lset: []storepb.Label{{Name: "a", Value: "a"}},
-				},
-			},
-		},
 	} {
 
 		if ok := t.Run(tc.title, func(t *testing.T) {
@@ -488,7 +464,7 @@ func TestProxyStore_SeriesSlowStores(t *testing.T) {
 		expectedWarningsLen int
 	}{
 		{
-			title: "partial response disabled one thanos query is slow to respond",
+			title: "partial response disabled; 1st store is slow, 2nd store is fast;",
 			storeAPIs: []Client{
 				&testClient{
 					StoreClient: &mockedStoreAPI{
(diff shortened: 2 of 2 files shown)


IKSIN:
Old ProxyStore logic:



New ProxyStore logic:

IKSIN:
This PR fixes the case when 2 or more stores are responding slowly.
It is the case if, for example,
a) common S3 storage degraded -> all stores become slow
b) multiple prom instances deployed to the same host and the host is under high CPU pressure.
new messages

This PR also fixes double timeout in case of warnings

Also, we’ve separated RT metrics with/without payload:
before this change we’ve seen pretty low RT for all stores because most of queries doesn’t return any data. But if store return any payload - RT usually is much more.

This PR change works well in most of cases because if store responding slowly - it’s usually since first data chunk.

Example:
You have 3 stores backed by Ceph and 3 fast proms/sidecars
Ceph is degraded and responding very slowly (~infinite time).
Query timeout = 2m, response timeout = 10s
As it was before:
We will wait at least 10s X 3 because we’ve started response timeout timer inside MergedSet -> Next()
Now:
We will wait at least 10s, because timeout works in parallel way for all stores.

d-ulyanov:
mishka_vodka_balalayka_

GiedriusS:
Seems like you have rebased this on `master` but you have recreated all of the commits with you as the author. Could we please fix this and only leave the proper commits before a review?

d-ulyanov:
@GiedriusS sure

pkg/store/proxy.go
```diff
@@ -49,12 +63,37 @@ type ProxyStore struct {
 	selectorLabels labels.Labels
 
 	responseTimeout time.Duration
+	metrics         *proxyStoreMetrics
+}
+
+func newProxyStoreMetrics(reg *prometheus.Registry) *proxyStoreMetrics {
+	var m proxyStoreMetrics
+
+	m.firstRecvDuration = prometheus.NewSummaryVec(prometheus.SummaryOpts{
```
bwplotka: Should we really use summaries here? (: Can we switch to histograms, maybe?

pkg/store/proxy.go
```diff
@@ -49,12 +63,37 @@ type ProxyStore struct {
...
+		Name:       "thanos_proxy_first_recv_duration",
+		Help:       "Time to get first part data from store(ms).",
+		Objectives: map[float64]float64{0.5: 0.05, 0.9: 0.01, 0.99: 0.001},
+		MaxAge:     2 * time.Minute,
+	}, []string{"store", "payload"})
+
+	m.timeoutRecvCount = prometheus.NewCounterVec(prometheus.CounterOpts{
+		Name: "thanos_proxy_timeout_recv_count",
+		Help: "Timeout recv count.",
```
bwplotka: Not really helpful help (:

pkg/store/proxy.go
```diff
@@ -49,12 +63,37 @@ type ProxyStore struct {
...
 	responseTimeout time.Duration
+	metrics         *proxyStoreMetrics
+}
+
+func newProxyStoreMetrics(reg *prometheus.Registry) *proxyStoreMetrics {
+	var m proxyStoreMetrics
+
+	m.firstRecvDuration = prometheus.NewSummaryVec(prometheus.SummaryOpts{
+		Name:       "thanos_proxy_first_recv_duration",
+		Help:       "Time to get first part data from store(ms).",
```
bwplotka: Probably should be float64 seconds

pkg/store/proxy.go
```diff
@@ -49,12 +63,37 @@ type ProxyStore struct {
...
+		Objectives: map[float64]float64{0.5: 0.05, 0.9: 0.01, 0.99: 0.001},
+		MaxAge:     2 * time.Minute,
+	}, []string{"store", "payload"})
```
bwplotka: We talked about `store` in metrics - this might leak cardinaltity (changing IP address), so I think we have to hook it to external labels and store TYPE as we do in storeset.

pkg/store/proxy.go
bwplotka: `Time to first byte` is what we can call it

IKSIN:
https://github.com/thanos-io/thanos/pull/1804 recreated pull request.

pkg/store/proxy.go
```diff
@@ -344,11 +398,55 @@ func startStreamSeriesSet(
...
+			rCh := make(chan *recvResponse, 1)
+			var rr *recvResponse
+			go func() {
+				t0 := time.Now()
+				r, err := s.stream.Recv()
+				if isFirstRecv {
+					if r.Size() > 0 {
+						metrics.withPayload.Observe(time.Since(t0).Seconds())
+					} else {
+						metrics.withoutPayload.Observe(time.Since(t0).Seconds())
```
bwplotka: Can we just skip the recv without payload? We care about time to first byte IMO

pkg/store/proxy.go
```diff
@@ -344,11 +398,55 @@ func startStreamSeriesSet(
...
 		defer close(s.recvCh)
-
 		for {
-			r, err := s.stream.Recv()
+			var cancel context.CancelFunc
+			if s.responseTimeout != 0 {
+				ctx, cancel = context.WithTimeout(ctx, s.responseTimeout)
+				defer cancel()
+			}
+			rCh := make(chan *recvResponse, 1)
```
bwplotka: Instead of allocating this every time we can reuse this channel

pkg/store/proxy.go
```diff
@@ -344,11 +398,55 @@ func startStreamSeriesSet(
...
 	wg.Add(1)
 	go func() {
+		isFirstRecv := true
 		defer wg.Done()
 		defer close(s.recvCh)
-
 		for {
-			r, err := s.stream.Recv()
+			var cancel context.CancelFunc
+			if s.responseTimeout != 0 {
```
bwplotka: Just curious: can we have this timeout for a whole response? Do we really need it per frame? Plus we might allocate a lot here so canceling per stream frame would be nice.

pkg/store/proxy.go
```diff
@@ -344,11 +398,55 @@ func startStreamSeriesSet(
...
+					return
+				}
+				s.errMtx.Lock()
+				s.err = err
+				s.errMtx.Unlock()
+				return
+			case rr = <-rCh:
+			}
+			close(rCh)
+			err := rr.err
```
bwplotka: Can we use those directly? Do we need those local variables?

pkg/store/proxy.go
```diff
@@ -344,11 +398,55 @@ func startStreamSeriesSet(
...
+					} else {
+						metrics.withoutPayload.Observe(time.Since(t0).Seconds())
+					}
+					isFirstRecv = false
+				}
+				rCh <- &recvResponse{r: r, err: err}
+			}()
+
+			select {
+			case <-ctx.Done():
```
bwplotka: I think we might want to separate response timeout context vs the given context. Just want to avoid this when client cancels request or we have overall query timeout.

pkg/store/proxy.go
IKSIN: We have too large external labels in our thanos-store's, therefore external labels is not comfortable. Also external_labels can have different order time from time. Maybe we can think about comfortable way to identify stores? For example, by bucket_name?

bwplotka: See storeset, we already do that (we also sort it), so if that's the problem we need to fix it everywhere (:

pkg/store/proxy.go
IKSIN: We observe strange data from this metrics (we set 10s timeout):
Снимок экрана 2019-11-27 в 17 44 21
Therefore, we want to save without_payload metrics, for investigate problems in future.

pkg/store/proxy.go
IKSIN: If we want to use timeout for a whole response:
1. We must to set such timeout ~query response. in this case we lose opportunity to fast cancel response from slow store (for example)
2. We need to increase channel buffer (for real parallel read from stores)

pkg/store/proxy.go
IKSIN: Снимок экрана 2019-11-27 в 17 58 38
I can't imagine comfortable work with such external_labels, presented as json string

bwplotka: Can you show querier /metrics page?

In this case, we need to fix this in separate PR `stores hash as you already have metric with this label: https://github.com/thanos-io/thanos/blob/7e11afe64af0c096743a3de8a594616abf52be45/pkg/query/storeset.go#L94 (:

pkg/store/proxy.go
bwplotka: Wait, this is because it's EOF which we should filter out. This the normal response when the server closes the stream, it should be filtered out. This is also received when context is canceled or timeout was triggered that's why you see 10s I think.

pkg/store/proxy.go
bwplotka: Yes, you are right all is connected and dependent.

I actually think we should increase buffer at some point, to some degree , but not here. Let's keep it per frame.

bwplotka: It would be nice to have timeout naming adjusted then to mention frame right now it is:

> If a Store doesn't send any data in this specified duration then a Store will be ignored and partial data will be returned if it's enabled. 0 disables timeout.

Maybe to:

`If a Store doesn't send any frame in this specified duration then a Store will be ignored and partial data will be returned if it's enabled. 0 disables timeout.`

bwplotka:
I think I addressed all the comments (:

pkg/store/proxy.go
IKSIN: `thanos_store_node_info{external_labels="{dc=\"m9\",env=\"production\",prometheus=\"monitoring/prometheus-apps-0\",prometheus_replica=\"A\"}"} 1
thanos_store_node_info{external_labels="{dc=\"m9\",env=\"production\",prometheus=\"monitoring/prometheus-apps-0\",prometheus_replica=\"A\"},{dc=\"m9\",env=\"production\",prometheus=\"monitoring/prometheus-apps-0\",prometheus_replica=\"B\"},{dc=\"m9\",env=\"production\",prometheus=\"monitoring/prometheus-apps-1\",prometheus_replica=\"A\"},{dc=\"m9\",env=\"production\",prometheus=\"monitoring/prometheus-apps-1\",prometheus_replica=\"B\"},{dc=\"m9\",env=\"production\",prometheus=\"monitoring/prometheus-apps-2\",prometheus_replica=\"A\"},{dc=\"m9\",env=\"production\",prometheus=\"monitoring/prometheus-apps-2\",prometheus_replica=\"B\"},{dc=\"m9\",env=\"production\",prometheus=\"monitoring/prometheus-apps-3\",prometheus_replica=\"A\"},{dc=\"m9\",env=\"production\",prometheus=\"monitoring/prometheus-apps-3\",prometheus_replica=\"B\"},{dc=\"m9\",env=\"production\",prometheus=\"monitoring/prometheus-apps-4\",prometheus_replica=\"A\"},{dc=\"m9\",env=\"production\",prometheus=\"monitoring/prometheus-apps-4\",prometheus_replica=\"B\"},{dc=\"m9\",env=\"production\",prometheus=\"monitoring/prometheus-apps-5\",prometheus_replica=\"A\"},{dc=\"m9\",env=\"production\",prometheus=\"monitoring/prometheus-apps-5\",prometheus_replica=\"B\"}"} 1
thanos_store_node_info{external_labels="{dc=\"m9\",env=\"production\",prometheus=\"monitoring/prometheus-apps-0\",prometheus_replica=\"B\"}"} 1
thanos_store_node_info{external_labels="{dc=\"m9\",env=\"production\",prometheus=\"monitoring/prometheus-apps-1\",prometheus_replica=\"A\"}"} 1
thanos_store_node_info{external_labels="{dc=\"m9\",env=\"production\",prometheus=\"monitoring/prometheus-apps-1\",prometheus_replica=\"B\"}"} 1
thanos_store_node_info{external_labels="{dc=\"m9\",env=\"production\",prometheus=\"monitoring/prometheus-apps-2\",prometheus_replica=\"A\"}"} 1
thanos_store_node_info{external_labels="{dc=\"m9\",env=\"production\",prometheus=\"monitoring/prometheus-apps-2\",prometheus_replica=\"B\"}"} 1
thanos_store_node_info{external_labels="{dc=\"m9\",env=\"production\",prometheus=\"monitoring/prometheus-apps-3\",prometheus_replica=\"A\"}"} 1
thanos_store_node_info{external_labels="{dc=\"m9\",env=\"production\",prometheus=\"monitoring/prometheus-apps-3\",prometheus_replica=\"B\"}"} 1
thanos_store_node_info{external_labels="{dc=\"m9\",env=\"production\",prometheus=\"monitoring/prometheus-apps-4\",prometheus_replica=\"A\"}"} 1
thanos_store_node_info{external_labels="{dc=\"m9\",env=\"production\",prometheus=\"monitoring/prometheus-apps-4\",prometheus_replica=\"B\"}"} 1
thanos_store_node_info{external_labels="{dc=\"m9\",env=\"production\",prometheus=\"monitoring/prometheus-apps-5\",prometheus_replica=\"A\"}"} 1
thanos_store_node_info{external_labels="{dc=\"m9\",env=\"production\",prometheus=\"monitoring/prometheus-apps-5\",prometheus_replica=\"B\"}"} 1
thanos_store_node_info{external_labels="{dc=\"m9\",env=\"production\",prometheus=\"monitoring/prometheus-hardware-0\",prometheus_replica=\"A\"}"} 1
thanos_store_node_info{external_labels="{dc=\"m9\",env=\"production\",prometheus=\"monitoring/prometheus-hardware-0\",prometheus_replica=\"A\"},{dc=\"m9\",env=\"production\",prometheus=\"monitoring/prometheus-hardware-0\",prometheus_replica=\"B\"},{dc=\"mts\",env=\"production\",prometheus_replica=\"prometheus01z1.h.o3.ru\"}"} 1
thanos_store_node_info{external_labels="{dc=\"m9\",env=\"production\",prometheus=\"monitoring/prometheus-hardware-0\",prometheus_replica=\"B\"}"} 1
thanos_store_node_info{external_labels="{dc=\"m9\",env=\"production\",prometheus=\"monitoring/prometheus-ingress-0\",prometheus_replica=\"A\"}"} 1
thanos_store_node_info{external_labels="{dc=\"m9\",env=\"production\",prometheus=\"monitoring/prometheus-ingress-0\",prometheus_replica=\"A\"},{dc=\"m9\",env=\"production\",prometheus=\"monitoring/prometheus-ingress-0\",prometheus_replica=\"B\"},{dc=\"m9\",env=\"production\",prometheus=\"monitoring/prometheus-ingress-1\",prometheus_replica=\"A\"},{dc=\"m9\",env=\"production\",prometheus=\"monitoring/prometheus-ingress-1\",prometheus_replica=\"B\"},{dc=\"m9\",env=\"production\",prometheus=\"monitoring/prometheus-other-0\",prometheus_replica=\"A\"},{dc=\"m9\",env=\"production\",prometheus=\"monitoring/prometheus-other-0\",prometheus_replica=\"B\"},{dc=\"m9\",env=\"production\",prometheus=\"monitoring/prometheus-system-0\",prometheus_replica=\"A\"},{dc=\"m9\",env=\"production\",prometheus=\"monitoring/prometheus-system-0\",prometheus_replica=\"B\"}"} 1
thanos_store_node_info{external_labels="{dc=\"m9\",env=\"production\",prometheus=\"monitoring/prometheus-ingress-0\",prometheus_replica=\"B\"}"} 1
thanos_store_node_info{external_labels="{dc=\"m9\",env=\"production\",prometheus=\"monitoring/prometheus-ingress-1\",prometheus_replica=\"A\"}"} 1
thanos_store_node_info{external_labels="{dc=\"m9\",env=\"production\",prometheus=\"monitoring/prometheus-ingress-1\",prometheus_replica=\"B\"}"} 1
thanos_store_node_info{external_labels="{dc=\"m9\",env=\"production\",prometheus=\"monitoring/prometheus-other-0\",prometheus_replica=\"A\"}"} 1
thanos_store_node_info{external_labels="{dc=\"m9\",env=\"production\",prometheus=\"monitoring/prometheus-other-0\",prometheus_replica=\"B\"}"} 1
thanos_store_node_info{external_labels="{dc=\"m9\",env=\"production\",prometheus=\"monitoring/prometheus-system-0\",prometheus_replica=\"A\"}"} 1
thanos_store_node_info{external_labels="{dc=\"m9\",env=\"production\",prometheus=\"monitoring/prometheus-system-0\",prometheus_replica=\"B\"}"} 1
# HELP thanos_store_nodes_grpc_connections Number of gRPC connection to Store APIs. Opened connection means healthy store APIs available for Querier.
# TYPE thanos_store_nodes_grpc_connections gauge
thanos_store_nodes_grpc_connections{external_labels="{dc=\"m9\",env=\"production\",prometheus=\"monitoring/prometheus-apps-0\",prometheus_replica=\"A\"}",store_type="sidecar"} 1
thanos_store_nodes_grpc_connections{external_labels="{dc=\"m9\",env=\"production\",prometheus=\"monitoring/prometheus-apps-0\",prometheus_replica=\"A\"},{dc=\"m9\",env=\"production\",prometheus=\"monitoring/prometheus-apps-0\",prometheus_replica=\"B\"},{dc=\"m9\",env=\"production\",prometheus=\"monitoring/prometheus-apps-1\",prometheus_replica=\"A\"},{dc=\"m9\",env=\"production\",prometheus=\"monitoring/prometheus-apps-1\",prometheus_replica=\"B\"},{dc=\"m9\",env=\"production\",prometheus=\"monitoring/prometheus-apps-2\",prometheus_replica=\"A\"},{dc=\"m9\",env=\"production\",prometheus=\"monitoring/prometheus-apps-2\",prometheus_replica=\"B\"},{dc=\"m9\",env=\"production\",prometheus=\"monitoring/prometheus-apps-3\",prometheus_replica=\"A\"},{dc=\"m9\",env=\"production\",prometheus=\"monitoring/prometheus-apps-3\",prometheus_replica=\"B\"},{dc=\"m9\",env=\"production\",prometheus=\"monitoring/prometheus-apps-4\",prometheus_replica=\"A\"},{dc=\"m9\",env=\"production\",prometheus=\"monitoring/prometheus-apps-4\",prometheus_replica=\"B\"},{dc=\"m9\",env=\"production\",prometheus=\"monitoring/prometheus-apps-5\",prometheus_replica=\"A\"},{dc=\"m9\",env=\"production\",prometheus=\"monitoring/prometheus-apps-5\",prometheus_replica=\"B\"}",store_type="store"} 1
thanos_store_nodes_grpc_connections{external_labels="{dc=\"m9\",env=\"production\",prometheus=\"monitoring/prometheus-apps-0\",prometheus_replica=\"B\"}",store_type="sidecar"} 1
thanos_store_nodes_grpc_connections{external_labels="{dc=\"m9\",env=\"production\",prometheus=\"monitoring/prometheus-apps-1\",prometheus_replica=\"A\"}",store_type="sidecar"} 1
thanos_store_nodes_grpc_connections{external_labels="{dc=\"m9\",env=\"production\",prometheus=\"monitoring/prometheus-apps-1\",prometheus_replica=\"B\"}",store_type="sidecar"} 1
thanos_store_nodes_grpc_connections{external_labels="{dc=\"m9\",env=\"production\",prometheus=\"monitoring/prometheus-apps-2\",prometheus_replica=\"A\"}",store_type="sidecar"} 1
thanos_store_nodes_grpc_connections{external_labels="{dc=\"m9\",env=\"production\",prometheus=\"monitoring/prometheus-apps-2\",prometheus_replica=\"B\"}",store_type="sidecar"} 1
thanos_store_nodes_grpc_connections{external_labels="{dc=\"m9\",env=\"production\",prometheus=\"monitoring/prometheus-apps-3\",prometheus_replica=\"A\"}",store_type="sidecar"} 1
thanos_store_nodes_grpc_connections{external_labels="{dc=\"m9\",env=\"production\",prometheus=\"monitoring/prometheus-apps-3\",prometheus_replica=\"B\"}",store_type="sidecar"} 1
thanos_store_nodes_grpc_connections{external_labels="{dc=\"m9\",env=\"production\",prometheus=\"monitoring/prometheus-apps-4\",prometheus_replica=\"A\"}",store_type="sidecar"} 1
thanos_store_nodes_grpc_connections{external_labels="{dc=\"m9\",env=\"production\",prometheus=\"monitoring/prometheus-apps-4\",prometheus_replica=\"B\"}",store_type="sidecar"} 1
thanos_store_nodes_grpc_connections{external_labels="{dc=\"m9\",env=\"production\",prometheus=\"monitoring/prometheus-apps-5\",prometheus_replica=\"A\"}",store_type="sidecar"} 1
thanos_store_nodes_grpc_connections{external_labels="{dc=\"m9\",env=\"production\",prometheus=\"monitoring/prometheus-apps-5\",prometheus_replica=\"B\"}",store_type="sidecar"} 1
thanos_store_nodes_grpc_connections{external_labels="{dc=\"m9\",env=\"production\",prometheus=\"monitoring/prometheus-hardware-0\",prometheus_replica=\"A\"}",store_type="sidecar"} 1
thanos_store_nodes_grpc_connections{external_labels="{dc=\"m9\",env=\"production\",prometheus=\"monitoring/prometheus-hardware-0\",prometheus_replica=\"A\"},{dc=\"m9\",env=\"production\",prometheus=\"monitoring/prometheus-hardware-0\",prometheus_replica=\"B\"},{dc=\"mts\",env=\"production\",prometheus_replica=\"prometheus01z1.h.o3.ru\"}",store_type="store"} 1
thanos_store_nodes_grpc_connections{external_labels="{dc=\"m9\",env=\"production\",prometheus=\"monitoring/prometheus-hardware-0\",prometheus_replica=\"B\"}",store_type="sidecar"} 1
thanos_store_nodes_grpc_connections{external_labels="{dc=\"m9\",env=\"production\",prometheus=\"monitoring/prometheus-ingress-0\",prometheus_replica=\"A\"}",store_type="sidecar"} 1
thanos_store_nodes_grpc_connections{external_labels="{dc=\"m9\",env=\"production\",prometheus=\"monitoring/prometheus-ingress-0\",prometheus_replica=\"A\"},{dc=\"m9\",env=\"production\",prometheus=\"monitoring/prometheus-ingress-0\",prometheus_replica=\"B\"},{dc=\"m9\",env=\"production\",prometheus=\"monitoring/prometheus-ingress-1\",prometheus_replica=\"A\"},{dc=\"m9\",env=\"production\",prometheus=\"monitoring/prometheus-ingress-1\",prometheus_replica=\"B\"},{dc=\"m9\",env=\"production\",prometheus=\"monitoring/prometheus-other-0\",prometheus_replica=\"A\"},{dc=\"m9\",env=\"production\",prometheus=\"monitoring/prometheus-other-0\",prometheus_replica=\"B\"},{dc=\"m9\",env=\"production\",prometheus=\"monitoring/prometheus-system-0\",prometheus_replica=\"A\"},{dc=\"m9\",env=\"production\",prometheus=\"monitoring/prometheus-system-0\",prometheus_replica=\"B\"}",store_type="store"} 1
thanos_store_nodes_grpc_connections{external_labels="{dc=\"m9\",env=\"production\",prometheus=\"monitoring/prometheus-ingress-0\",prometheus_replica=\"B\"}",store_type="sidecar"} 1
thanos_store_nodes_grpc_connections{external_labels="{dc=\"m9\",env=\"production\",prometheus=\"monitoring/prometheus-ingress-1\",prometheus_replica=\"A\"}",store_type="sidecar"} 1
thanos_store_nodes_grpc_connections{external_labels="{dc=\"m9\",env=\"production\",prometheus=\"monitoring/prometheus-ingress-1\",prometheus_replica=\"B\"}",store_type="sidecar"} 1
thanos_store_nodes_grpc_connections{external_labels="{dc=\"m9\",env=\"production\",prometheus=\"monitoring/prometheus-other-0\",prometheus_replica=\"A\"}",store_type="sidecar"} 1
thanos_store_nodes_grpc_connections{external_labels="{dc=\"m9\",env=\"production\",prometheus=\"monitoring/prometheus-other-0\",prometheus_replica=\"B\"}",store_type="sidecar"} 1
thanos_store_nodes_grpc_connections{external_labels="{dc=\"m9\",env=\"production\",prometheus=\"monitoring/prometheus-system-0\",prometheus_replica=\"A\"}",store_type="sidecar"} 1
thanos_store_nodes_grpc_connections{external_labels="{dc=\"m9\",env=\"production\",prometheus=\"monitoring/prometheus-system-0\",prometheus_replica=\"B\"}",store_type="sidecar"} 1`
All metrics with `external_labels`. It looks ugly and unreadable, also json string to hard to process in grafana (for naming graphs)

pkg/store/proxy.go
IKSIN: Sorry, I don't understand... I think that we must to see minimal RT for first byte without payload...

pkg/store/proxy.go
IKSIN: I try to increase buffer to fit all response.
Increase buffer affected only usage memory, but not increase RT, independed on slow/fast stores )

pkg/store/proxy.go
bwplotka: It's crazy!

I think, still this should be resolved outside of this PR (: 

Again, I think `address` might add some cardinality, but I am tempted to allow it under some flag...

pkg/store/proxy.go
bwplotka: No, I EOF is when server closes a stream, but we account that as `first byte without payload` - I believe that's wrong

pkg/store/proxy.go
bwplotka: In my opinion, this memory will be buffered anyway at some point, but anyway, let's not introduce this here. I think we all like this PR in such logic: To make sure we timeout on the first byte from the slow store instead precisely.

bwplotka:
I think this PR makes sense, just some suggestions, any movement here? (:

d-ulyanov:
I suppose we'll continue with this PR on next week, @IKSIN is on vacations currently :)

pkg/store/proxy.go
IKSIN: I think that `without_payload` == `EOF on first byte` and it's represetnatine metric...

pkg/store/proxy.go
IKSIN: We need to wait last recv on timeout for close channel for send unused data frame, therefore we can't reuse firs channel

IKSIN:
PR updated

pkg/store/proxy.go
```diff
@@ -256,10 +301,21 @@ func (s *ProxyStore) Series(r *storepb.SeriesRequest, srv storepb.Store_SeriesSe
...
+				storeTypeStr = st.StoreType().String()
+			}
+			metrics := &storeRequestMetrics{
+				withPayload:       s.metrics.timeToFirstByte.WithLabelValues(st.LabelSetsString(), storeTypeStr, withPayloadLabel),
+				withoutPayload:    s.metrics.timeToFirstByte.WithLabelValues(st.LabelSetsString(), storeTypeStr, withoutPayloadLabel),
+				frameTimeoutCount: s.metrics.frameTimeoutCount.WithLabelValues(st.LabelSetsString(), storeTypeStr),
+				queryTimeoutCount: s.metrics.queryTimeoutCount.WithLabelValues(st.LabelSetsString(), storeTypeStr),
+			}
+
+			seriesSet = append(seriesSet, startStreamSeriesSet(seriesCtx, s.logger, closeSeries,
```
GiedriusS: This should be under the comment, no? (:

pkg/store/proxy.go
```diff
@@ -49,12 +67,41 @@ type ProxyStore struct {
...
+
+	m.timeToFirstByte = prometheus.NewHistogramVec(prometheus.HistogramOpts{
+		Name:    "thanos_proxy_time_to_first_byte_seconds",
+		Help:    "Time to get 1st frame from store.",
+		Buckets: []float64{0.01, 0.1, 0.3, 0.6, 1, 3, 6, 9, 20, 30, 60, 90, 120, 240, 360, 720},
+	}, []string{"external_labels", "store_type", "payload"})
+
+	m.frameTimeoutCount = prometheus.NewCounterVec(prometheus.CounterOpts{
+		Name: "thanos_proxy_frame_timeouted_requests_count",
+		Help: "Number of requests with expired store.respone-timeout",
```
GiedriusS: Typo.

pkg/store/proxy.go
GiedriusS: Could you add a check for `err` here before sending it back? It probably is equal to `io.EOF` in such case and it should be filtered out.

pkg/store/proxy.go
```diff
@@ -344,18 +406,51 @@ func startStreamSeriesSet(
...
+		isFirstRecv := true
 		defer wg.Done()
 		defer close(s.recvCh)
-
 		for {
-			r, err := s.stream.Recv()
+			frameTimeoutCtx := context.Background()
+			var cancel context.CancelFunc
+			if s.responseTimeout != 0 {
+				frameTimeoutCtx, cancel = context.WithTimeout(frameTimeoutCtx, s.responseTimeout)
```
GiedriusS: Could we move out the construction of this context out of this function? It should probably make things easier to understand. WDYT, @bwplotka? This part is really becoming complex :(

pkg/store/proxy.go
IKSIN: No, It's not to be commented. Maybe you mean add a comment?

pkg/store/proxy.go
IKSIN: I don't see a reasons to check on some errors here. All errors processed out of this goroutine. This goroutine needed only for get data async.

pkg/store/proxy.go
GiedriusS: There is a comment on the 317 line:
```
 			// Schedule streamSeriesSet that translates gRPC streamed response 
...
```
Could this be moved under that comment or removed?

pkg/store/proxy.go
GiedriusS: What are your thoughts on this, @IKSIN ?

pkg/store/proxy.go
```diff
@@ -256,10 +301,21 @@ func (s *ProxyStore) Series(r *storepb.SeriesRequest, srv storepb.Store_SeriesSe
 				continue
 			}
 
+			var storeTypeStr string
+			if st.StoreType() != nil {
+				storeTypeStr = st.StoreType().String()
+			}
```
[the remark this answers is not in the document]
IKSIN: refactored

pkg/store/proxy.go
IKSIN: moved to function

pkg/store/proxy.go
IKSIN: Moved under the comment

IKSIN:
updated

pkg/store/proxy.go
```diff
@@ -49,12 +67,41 @@ type ProxyStore struct {
...
+	m.timeToFirstByte = prometheus.NewHistogramVec(prometheus.HistogramOpts{
+		Name:    "thanos_proxy_time_to_first_byte_seconds",
+		Help:    "Time to get 1st frame from store.",
+		Buckets: []float64{0.01, 0.1, 0.3, 0.6, 1, 3, 6, 9, 20, 30, 60, 90, 120, 240, 360, 720},
+	}, []string{"external_labels", "store_type", "payload"})
+
+	m.frameTimeoutCount = prometheus.NewCounterVec(prometheus.CounterOpts{
+		Name: "thanos_proxy_frame_timeouted_requests_count",
+		Help: "Number of requests with expired store.response-timeout",
+	}, []string{"external_labels", "store_type"})
```
povilasv: Not sure if having `external_labels` is good idea as it might be a huge string with multiple key value pairs. I don't think anyone would query `thanos_proxy_frame_timeouted_requests_count{external_labels="a=b c=d"}` ?

pkg/store/proxy.go
```diff
@@ -49,12 +67,41 @@ type ProxyStore struct {
...
+	m.queryTimeoutCount = prometheus.NewCounterVec(prometheus.CounterOpts{
+		Name: "thanos_proxy_query_timeouted_requests_count",
+		Help: "Number of requests with expired query.timeout",
+	}, []string{"external_labels", "store_type"})
```
povilasv: ditto

pkg/store/proxy.go
```diff
@@ -49,12 +67,41 @@ type ProxyStore struct {
...
+}
+
+func newProxyStoreMetrics(reg *prometheus.Registry) *proxyStoreMetrics {
+	var m proxyStoreMetrics
+
+	m.timeToFirstByte = prometheus.NewHistogramVec(prometheus.HistogramOpts{
+		Name:    "thanos_proxy_time_to_first_byte_seconds",
+		Help:    "Time to get 1st frame from store.",
+		Buckets: []float64{0.01, 0.1, 0.3, 0.6, 1, 3, 6, 9, 20, 30, 60, 90, 120, 240, 360, 720},
+	}, []string{"external_labels", "store_type", "payload"})
```
povilasv: I think having histogram with  `external_labels` might have huge cardinality, consider getting rid of it.

pkg/store/proxy.go
povilasv: I also think we don't need `without_payload` metrics, let's just have time_to_first_byte metrics. What are you going to differently when you see `without_payload`  vs `with_payload` metrics increasing?

pkg/store/proxy.go
IKSIN: Look for https://github.com/thanos-io/thanos/pull/1789#discussion_r351232976 discuss please )

pkg/store/proxy.go
GiedriusS: I agree here. This metric roughly says "time until a node tells us that it doesn't have any metrics". In what cases do you think it could be useful? `payload` is a bit misleading as well, IMHO. Payload is "the actual information or message in transmitted data" and even if we get EOF it still is some kind of data. Even if we will keep this around I'd suggest renaming this to perhaps `response_kind` that could be `no_metrics` or `metrics`.

pkg/store/proxy.go
GiedriusS: Yep, and we have rejected PRs with these labels in the past because our users could be using service discovery mechanisms with highly dynamic i.e. temporary external labels.

IKSIN:
Updated. Get rid of metrics. ProxyStore timeouts only.

pkg/store/proxy.go
```diff
@@ -347,16 +356,35 @@ func startStreamSeriesSet(
...
-			r, err := s.stream.Recv()
+			frameTimeoutCtx, cancel := startFrameCtx(s.responseTimeout)
+			if cancel != nil {
+				defer cancel()
+			}
+			rCh := make(chan *recvResponse, 1)
+			var rr *recvResponse
+			go func() {
+				r, err := s.stream.Recv()
+				rCh <- &recvResponse{r: r, err: err}
```
povilasv: Consider adding `ctx.Done()` check in here.

I think the error is in the this scenario:

1. stream.Recv() blocks waiting for data
2. Client cancels request ctx.Done()
3. we close the channel (line 380)
4. stream.Recv() finishes and tries to push data into closed channel which should panic out.

bwplotka:
Let's focus on that next week to merge (:

pkg/store/proxy.go
IKSIN: Therefore used buffered channel. Handing ctx.Done in goroutine too complicated code in my opinion.

povilasv: I don't understand what this has to do with buffered channels. Sending data to closed channel will cause panics with or without buffered channels.

IKSIN: Oh, sorry, buffered channel fix another problem. In your use-case: goroutine can not send data to closed channel, because it's create for every recv. We can create channel outside of loop and try to reuse it, but it will complicated code in my opinion.

IKSIN: Oh, sorry, buffered channel solve another problem)

In your use-case:
1. stream.Recv() blocks waiting for data
2. Client cancels request ctx.Done()
3. we close the channel (line 380) - only if we read from channel (in other select cases we only `return`), therefor pointer on rCh was saved in goroutine and closed(via gabarge collector) after close goroutine.
4. stream.Recv() finishes and tries to push data into not closed channel.

But I move close channel into goroutine to code will be more understandable.

IKSIN:
@povilasv @bwplotka Any comments?

pkg/store/proxy.go
```diff
@@ -348,6 +343,21 @@ type streamSeriesSet struct {
 	closeSeries     context.CancelFunc
 }
 
+type recvResponse struct {
+	r   *storepb.SeriesResponse
+	err error
+}
+
+func startFrameCtx(responseTimeout time.Duration) (context.Context, context.CancelFunc) {
+	frameTimeoutCtx := context.Background()
```
bwplotka: I would really suggest to actually get context from argument, (so from `startStreamSeriesSet` )

pkg/store/proxy.go
```diff
@@ -348,6 +343,21 @@ type streamSeriesSet struct {
 	closeSeries     context.CancelFunc
 }
 
+type recvResponse struct {
+	r   *storepb.SeriesResponse
+	err error
+}
+
+func startFrameCtx(responseTimeout time.Duration) (context.Context, context.CancelFunc) {
```
bwplotka:
```suggestion
func frameCtx(responseTimeout time.Duration) (context.Context, context.CancelFunc) {
```

pkg/store/proxy.go
```diff
@@ -384,14 +394,34 @@ func startStreamSeriesSet(
 			}
 		}()
 		for {
-			r, err := s.stream.Recv()
+			frameTimeoutCtx, cancel := startFrameCtx(s.responseTimeout)
+			if cancel != nil {
```
bwplotka: just defer cancel always and return  `func() {}` not nil

pkg/store/proxy.go
```diff
@@ -384,14 +394,34 @@ func startStreamSeriesSet(
...
+				defer cancel()
+			}
+			rCh := make(chan *recvResponse, 1)
+			var rr *recvResponse
+			go func() {
```
bwplotka: So we care about the first frame right? or timeout for all frames?

bwplotka: In any case I think we need to start this go routine before `for` loop and have here `for` loop as well.

This will make sure that we only have one 2 go routines running: one waiting for recv or context cancel, second for reading.

Current implementation will constantly allocate new channel and go routine. For 1000 frames x 100 concurrent queries this might matter.


I wish we have benchmarks for querier ))): Like we do for Store now: https://github.com/thanos-io/thanos/blob/88f6be883b66036c589c5697034e3d99c0848b3b/pkg/store/bucket_test.go#L1071

Actually added issue: https://github.com/thanos-io/thanos/issues/2105

pkg/store/proxy.go
```diff
@@ -402,59 +432,40 @@ func startStreamSeriesSet(
...
-	if s.responseTimeout != 0 {
-		timeoutMsg = fmt.Sprintf("failed to receive any data in %s from %s", s.responseTimeout.String(), s.name)
-
-		timeoutCtx, done := context.WithTimeout(s.ctx, s.responseTimeout)
-		defer done()
-		ctx = timeoutCtx
+func (s *streamSeriesSet) timeoutHandling(isQueryTimeout bool, ctx context.Context) {
+	var err error
+	if isQueryTimeout {
+		err = errors.Wrap(ctx.Err(), fmt.Sprintf("failed to receive any data from %s", s.name))
```
bwplotka: What about just passing error to propagate instead here? and rename method to `handleErr`?

pkg/store/proxy.go
bwplotka: Ignore me you already check the parent one, so all good

IKSIN:
@bwplotka PR updated) w/o benchmarks(

pkg/store/proxy.go
```diff
@@ -383,78 +393,79 @@ func startStreamSeriesSet(
 				emptyStreamResponses.Inc()
 			}
 		}()
+
+		rCh := make(chan *recvResponse)
+		recvCancel := make(chan bool)
```
povilasv:
```suggestion
		done := make(chan struct{})
```

I think this is more common for cancelation channels, more info -> https://blog.golang.org/pipelines#TOC_6.

IKSIN: fixed

bwplotka:
Looking!

pkg/store/proxy.go
```diff
@@ -383,78 +393,79 @@ func startStreamSeriesSet(
 				emptyStreamResponses.Inc()
```
bwplotka: I don't we want to increment this if the context was actually cancelled.. right?

pkg/store/proxy.go
```diff
@@ -383,78 +393,79 @@ func startStreamSeriesSet(
...
+			frameTimeoutCtx, cancel := frameCtx(s.responseTimeout)
+			defer cancel()
+			var rr *recvResponse
 
-			if err == io.EOF {
+			var err error
+			select {
+			case <-ctx.Done():
+				close(done)
+				err = errors.Wrap(ctx.Err(), fmt.Sprintf("failed to receive any data from %s", s.name))
```
bwplotka: Why separate `err` variable?

also can we use `Wrapf` instead of sprintf?

pkg/store/proxy.go
```diff
@@ -383,78 +393,79 @@ func startStreamSeriesSet(
...
+				s.handleErr(err)
+				return
+			case <-frameTimeoutCtx.Done():
+				close(done)
+				err = errors.Wrap(frameTimeoutCtx.Err(), fmt.Sprintf("failed to receive any data in %s from %s", s.responseTimeout.String(), s.name))
```
bwplotka: ditto

pkg/store/proxy.go
```diff
@@ -383,78 +393,79 @@ func startStreamSeriesSet(
...
-				s.errMtx.Lock()
-				s.err = wrapErr
-				s.errMtx.Unlock()
+			if rr.err == io.EOF {
+				close(done)
 				return
 			}
 
+			if rr.err != nil {
+				wrapErr := errors.Wrapf(rr.err, "receive series from %s", s.name)
```
bwplotka: ditto

pkg/store/proxy.go
```diff
@@ -383,78 +393,79 @@ func startStreamSeriesSet(
...
+				s.handleErr(wrapErr)
+				close(done)
```
bwplotka: can we close done in `handleErr`?

pkg/store/proxy.go
IKSIN: Yes, therefore `numResponses++` only on recv processed.

IKSIN:
@bwplotka Updated!)

pkg/store/proxy.go
bwplotka: Why another var? (: We can inline this.. not a blocker though.

d-ulyanov:
Hooray! :)
Thank you for review, @bwplotka  @GiedriusS  @povilasv !
🍻

Issue #1779, opened 2019-11-22
ruler: silently mangles rules read from config (subqueries)


bjakubski:
Thanos official docker 0.8.1.
Also checked with master 8a8f576a go 1.13

**What happened**:

When investigating thanos ruler alerting rules not firing I've noticed that thanos mangled all my rules that used subquery with offset.

Rules shown (and executed) in ruler are different than ones in config files with rules.

**What you expected to happen**:

rules read cannot be altered and should be executed as-is

**How to reproduce it (as minimally and precisely as possible)**:

I've reduced offending query to a simple case:
```
rate(some_metric[1h:5m] offset 1d)
```
it is nonsensical, but it is enough to demonstrate a problem
Case can be reproduced in single command

```
thanos rule --query foobar --rule-file <(echo 'groups: [ {name: test, rules: [{ alert: Test, expr: "rate(some_metric[1h:5m] offset 1d)" }] } ]')
```

After accessing localhost:10902 one can see that instead of
```
rate(some_metric[1h:5m] offset 1d)
```
thanos ruler shows
```
rate(some_metric[1h:5m])
```

While this minimal example has nonsensical expr: my real query demonstrating this problem works as expected under normal circumstances i.e. if I go to thanos query and execute my query with "offset" added back where it should be I get proper results.

I was only able to reproduce this problem when there in query there both subquery and offset used.

This is quite serious as it is a *silent* failure. Nothing visible happens. But queries are mangled and (in my case) they never alert when they should if queries were untouched.

bjakubski:
This seems to be prometheus problem. Same thing happens with bare prometheus.
I'll try to confirm/report upstream tomorrow

bwplotka:
Yup, I played with this on Prometheus side as well and it looks really similar issue is there.

bjakubski:
https://github.com/prometheus/prometheus/pull/6373

PR #1790, opened 2019-11-26
Update prometheus for subqueries support in ruler.

bwplotka:
* Upgraded Prometheus to the latest.
   * The changeset is large due to: https://github.com/prometheus/prometheus/pull/6326
* Added regression test.

Fixes https://github.com/thanos-io/thanos/issues/1779

cc @bjakubski @simonpasquier


pkg/store/matchers.go
```diff
@@ -2,32 +2,28 @@ package store
...
-		return labels.NewRegexpMatcher(m.Name, "^(?:"+m.Value+")$")
+		return labels.NewMatcher(labels.MatchRegexp, m.Name, m.Value)
 
 	case storepb.LabelMatcher_NRE:
-		m, err := labels.NewRegexpMatcher(m.Name, "^(?:"+m.Value+")$")
-		if err != nil {
-			return nil, err
-		}
-		return labels.Not(m), nil
+		return labels.NewMatcher(labels.MatchNotRegexp, m.Name, m.Value)
```
squat: Same

pkg/store/matchers.go
```diff
@@ -2,32 +2,28 @@ package store
...
-		return labels.NewEqualMatcher(m.Name, m.Value), nil
+		return labels.NewMatcher(labels.MatchEqual, m.Name, m.Value)
 
 	case storepb.LabelMatcher_NEQ:
-		return labels.Not(labels.NewEqualMatcher(m.Name, m.Value)), nil
+		return labels.NewMatcher(labels.MatchNotEqual, m.Name, m.Value)
 
 	case storepb.LabelMatcher_RE:
-		return labels.NewRegexpMatcher(m.Name, "^(?:"+m.Value+")$")
+		return labels.NewMatcher(labels.MatchRegexp, m.Name, m.Value)
```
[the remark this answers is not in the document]
bwplotka: We do not need as this is done already in `NewMatcher`:

```go
// NewMatcher returns a matcher object.
func NewMatcher(t MatchType, n, v string) (*Matcher, error) {
	m := &Matcher{
		Type:  t,
		Name:  n,
		Value: v,
	}
	if t == MatchRegexp || t == MatchNotRegexp {
		re, err := regexp.Compile("^(?:" + v + ")$")
		if err != nil {
			return nil, err
		}
		m.re = re
	}
	return m, nil
}
```

(:

pkg/store/matchers.go
bwplotka: same answer (:

PR #1792, opened 2019-11-26
objstore: add ObjectSize() implementation

GiedriusS:
Adds a new method ObjectSize(context.Context, string) method to our
object storage interface which returns the size of the specified object
in uint64 without actually downloading all of it.

This is a pre-requisite to one check that we want to add (https://github.com/thanos-io/thanos/pull/1550).

Add a check to our E2E tests to see if it works.

### DIFF
--- a/pkg/objstore/objtesting/acceptance_e2e_test.go
+++ b/pkg/objstore/objtesting/acceptance_e2e_test.go
@@ -42,6 +42,11 @@ func TestObjStore_AcceptanceTest_e2e(t *testing.T) {
 		testutil.Ok(t, err)
 		testutil.Equals(t, "@test-data@", string(content))
 
+		// Check if we can get the correct size.
+		sz, err := bkt.ObjectSize(ctx, "id1/obj_1.some")
+		testutil.Ok(t, err)
+		testutil.Assert(t, sz == 11, "expected size to be equal to 11")
+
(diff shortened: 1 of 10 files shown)


pkg/objstore/objtesting/acceptance_e2e_test.go
```diff
@@ -42,6 +42,11 @@ func TestObjStore_AcceptanceTest_e2e(t *testing.T) {
 		testutil.Ok(t, err)
 		testutil.Equals(t, "@test-data@", string(content))
 
+		// Check if we can get the correct size.
```
bwplotka: Can we check if ObjectSize for non existing block properly errors? Also if we can find assert NotFound on such operation if block is missing? The latter is not critical, but if we cannot assume so, we should document that as a method comment (:

GiedriusS: Added a test for this before the initial upload and cleaned up the functions so that `IsObjNotFoundErr(error)` would be properly triggered in such cases.

GiedriusS:
Merging this as I have addressed the comments and since it doesn't change the behaviour in any way ATM. Will rebase https://github.com/thanos-io/thanos/pull/1550 on these methods now.

bwplotka:
Thanks!

PR #1796, opened 2019-11-27
docs: Update getting-started.md

wbh1:

## Changes

Update docs based off feedback received in #thanos channel on CNCF Slack about confusion trying to run the `go get` command without enabling Go module support.

## Verification

N/A


bwplotka:
Stale, but also no longer needed as we use Go version with go modules enabled by default ^^ Thanks for contribution anyway!

PR #1797, opened 2019-11-27
update image tag

daixiang0:

## Changes

Update image tag in doc. 0.3.2 is too old to use some cool new feature.

## Verification

Test pass at local.


daixiang0:
The original one still work but some feature not match docs in https://thanos.io/ @bwplotka

PR #1798, opened 2019-11-27
remove docker hub part

daixiang0:

## Changes

Since we've moved to quay and do not update docker hub, remove all docker hub part.

I see that we mirror it in docker hub in changelog, but in fact, we have not mirrored it recently refer to https://hub.docker.com/r/improbable/thanos/tags.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -70,9 +70,7 @@ jobs:
       - run: ln -s .build/linux-amd64/thanos thanos
       - run: make docker
       - run: docker run thanos --help
-      # Upload to both dockerhub and quay.io.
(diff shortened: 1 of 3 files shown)


bwplotka:
Thanks for this, but we mirror it properly on https://hub.docker.com/r/thanosio/thanos (: 

How we can be more clear it's not Improbable link anymore? (:

daixiang0:
@bwplotka how about add comment in https://hub.docker.com/r/improbable/thanos?

bwplotka:
Good idea!

@domgreen Can you update README of  https://hub.docker.com/r/improbable/thanos? I don't have access to this.

I think we can close this then.

domgreen:
I can't change it myself will try to find who does have access tomorrow 👍

PR #1801, opened 2019-11-27
clean up all white noise

daixiang0:

## Changes

Remove all white noise.


daixiang0:
If possible, i want to add a tool for this when any docs updated.

bwplotka:
Yes, please, can we add some linting for his? Maybe in this PR? (:

daixiang0:
i would do it, where is the best to add this? In circleci config file?

bwplotka:
Yes

daixiang0:
@bwplotka mind have a look?

scripts/cleanup-white-noise.sh
```diff
@@ -0,0 +1,2 @@
+#!/bin/bash
+sed -i 's/[ \t]*$//' "$@"
```
bwplotka: Nice! I think we might want to have it in `Makefile` as well. We can hook it on the CI in later PR.

daixiang0: format `*.go`?

daixiang0:
ping @bwplotka

daixiang0:
mind review again? @bwplotka

daixiang0:
mind review again? @bwplotka

daixiang0:
I want to do those changes to make target more clear:

1. `docs` only generate docs
2. `format` only call `go format -w`
3. `docs-check`  would call `docs` target first then clean white noise
4. `format-check` would call `format` target first then clean white noise

Wait for your feedback @bwplotka

bwplotka:
That's not what I proposed (: 

`*-check` would always VERIFY and return non 0 signal code if something is not done. Actually for `format` we have `lint` already so let's use that (: 
`format` and `docs` would `fix` things in you local repo for you to commit it and include in the PR (: 


So:
* `docs`: generate docs + **clean whitespace in docs**
* `format`: format + **clean whitespace in Go code**
* `docs-check` if docs not generated and **if any whitespace not cleaned** return error
* `lint` if Go file not formatted / lint found issues/ misspell or **whitespeaces not cleaned** return error (:  

In bold font things that are would like to change in this PR (: Does it make sense? (:

daixiang0:
@bwplotka thanks for your patience, please review again.

daixiang0:
@bwplotka thanks for your patience, please review again.

PR #1805, opened 2019-11-27
Fix typo in package name

antonio:
Just a quick fix for a typo I noticed while reading the source code.

## Changes

Update the package import

## Verification

N/A

### DIFF
--- a/pkg/compact/compact.go
+++ b/pkg/compact/compact.go
@@ -17,7 +17,7 @@ import (
 	"github.com/pkg/errors"
 	"github.com/prometheus/client_golang/prometheus"
 	"github.com/prometheus/prometheus/pkg/labels"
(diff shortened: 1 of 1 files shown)


pkg/compact/compact.go
```diff
@@ -17,7 +17,7 @@ import (
 	"github.com/pkg/errors"
 	"github.com/prometheus/client_golang/prometheus"
 	"github.com/prometheus/prometheus/pkg/labels"
-	promlables "github.com/prometheus/prometheus/pkg/labels"
+	promlabels "github.com/prometheus/prometheus/pkg/labels"
```
bwplotka: Ups, I missed this while upgrading to Prometheus with unified labels packages. This line can be removed and we can use `labels` below (: Thanks for spotting!

PR #1807, opened 2019-11-27
Store Fix: Removed dir on loadmeta failure

thorfour:
If loadmeta fails during SyncBlocks, the block directory is left behind,
resulting in subsequent failures to read the meta file since loadmeta just looks for the directory to exist.

FailureToLoadMeta unit test

## Changes

Calls `os.RemoveAll` on the blockdir when returning error from load meta

## Verification

Added unit test that checks if a sync against a block that's missing a meta.json gets removed.

### DIFF
--- a/pkg/store/bucket.go
+++ b/pkg/store/bucket.go
@@ -343,7 +343,8 @@ func (s *BucketStore) SyncBlocks(ctx context.Context) error {
 		bdir := path.Join(s.dir, id.String())
 		meta, err := loadMeta(ctx, s.logger, s.bucket, bdir, id)
 		if err != nil {
-			return errors.Wrap(os.RemoveAll(bdir), "load meta")
+			level.Warn(s.logger).Log("msg", "error loading meta", "block", id, "err", err)
+			return os.RemoveAll(bdir)
 		}
 
 		inRange, err := s.isBlockInMinMaxRange(ctx, meta)
--- a/pkg/store/bucket.go
+++ b/pkg/store/bucket.go
@@ -343,7 +343,7 @@ func (s *BucketStore) SyncBlocks(ctx context.Context) error {
 		bdir := path.Join(s.dir, id.String())
 		meta, err := loadMeta(ctx, s.logger, s.bucket, bdir, id)
 		if err != nil {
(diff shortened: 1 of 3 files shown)


pkg/store/bucket.go
```diff
@@ -343,7 +343,7 @@ func (s *BucketStore) SyncBlocks(ctx context.Context) error {
 		bdir := path.Join(s.dir, id.String())
 		meta, err := loadMeta(ctx, s.logger, s.bucket, bdir, id)
 		if err != nil {
-			return errors.Wrap(err, "load meta")
+			return errors.Wrap(os.RemoveAll(bdir), "load meta")
```
squat: Hmm I don’t think this is quite right. This change will suppress the load meta error if the directory cleanup is successful. If we get an error loading metas, we want to return that error, not the error concerning cleanup. Cleanup should be a defer that logs any error.

thorfour: Do we want to return the error? Wouldn't that cause the Iter function to immediately abort the iteration?

squat: Aborting the iteration is the current logic

thorfour: Sure.. but it would be aborting an Iter function due to a transient error (block is still being written), when it could continue on syncing the remaining blocks.

This change would bring this more in-line with the functions below it which also log the error and clean up the directory.

squat: I’m not necessarily against changing the logic, it simply isn’t what is done today and was not mentioned in your PR. Nevertheless, this change still suppresses the cause of the error without logging it. It also mixes up the causes of the errors. It would wrap a directory deletion error with `load meta` as though the deletion were the cause of failing to load the meta.

thorfour: I understand what you're saying. I've updated it to be more in-line with the functions below it. Log the error, and return the `os.RemoveAll` error.

thorfour:
@bwplotka you're right, just realized I had an older version of that file, that my IDE was presenting which had this instead of the line checking for the meta file.
```
if _, err := os.Stat(dir); os.IsNotExist(err) {
```

Egg on my face. Closing

PR #1823, opened 2019-12-02
Added special mode where ruler evaluates in past.

bwplotka:
Use cases: 
* Recording rules for X old data
* Starting point for backfill feature in Thanos. 

This trick works only if the evaluations are within the same 2h timestamps. This produces
blocks properly in the head for older samples.

This can be actually extended to full backfill one-off job at some point as long as it's done in
same time flow for all rules.

Probably makes sense mostly with vertical compaction enabled to join small blocks together for efficiency, but still
will work as it is now.

Started running ruler at 18:30 - we can see backfilled data for 15:30:
Above Prometheus, below Ruler:
image

Rule: 
image

Full ruler view: 
image

Will run it for a longer time let's see how blocking cut will work, but looking good. It's quite crazy though!

### DIFF
--- a/cmd/thanos/rule.go
+++ b/cmd/thanos/rule.go
@@ -111,6 +111,8 @@ func registerRule(m map[string]setupFunc, app *kingpin.Application) {
 	dnsSDResolver := cmd.Flag("query.sd-dns-resolver", "Resolver to use. Possible options: [golang, miekgdns]").
 		Default("golang").Hidden().String()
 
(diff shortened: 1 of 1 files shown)


squat:
Very cool! We may need to document the limitation of evaluations being within the same 2hr

bwplotka:
This is documented in flag, but I am not sure if we should merge this, it was a quick experiment only.

metalmatze:
Yes, let's wait and see.

PR #1824, opened 2019-12-03
[bugfix] pkg/reloader: add return which miss.

johncming:
https://github.com/thanos-io/thanos/blob/bd9aa1b4be3bb5d841cb7271c29d02ebb5eb5168/pkg/reloader/reloader.go#L292-L294

`return` statement lost.

### DIFF
--- a/pkg/reloader/reloader.go
+++ b/pkg/reloader/reloader.go
@@ -291,6 +291,7 @@ func (r *Reloader) apply(ctx context.Context) error {
 		return nil
 	}); err != nil {
 		level.Error(r.logger).Log("msg", "Failed to trigger reload. Retrying.", "err", err)
+		return err
 	}
 
 	return nil


johncming:
> I think it was intended. See the `Retrying` in the log line (:
> 
> This is as it will be retried by another interval tick.

`apply` is mainly used for `watch`, where retry is already available. If there are still errors after retry, can consider the errors unrecoverable for a short period of time and let the program terminate which is better.

povilasv:
Yup the error checking logic here https://github.com/thanos-io/thanos/blob/bc4fee6294145e375c528589bccdc98ee94c0cc2/pkg/runutil/runutil.go#L92-L95 should retry `retryInterval` times and then return the error and right now it loops forever

bwplotka:
Yes, this might be confusing but see a few lines above: 
`// Retry trigger reload until it succeeded or next tick is near.` And by tick we mean here this tick: https://github.com/thanos-io/thanos/blob/bd9aa1b4be3bb5d841cb7271c29d02ebb5eb5168/pkg/reloader/reloader.go#L164 

..not runutil one @povilasv 

It's not timeout of rules apply. It's rather a continuous retry combined with watch interval.

 *If we want to introduce timeout and fail* we need to do it properly instead of just adding error here. For example allowing user to have special timeout for it.

 I would vote howevr for not failing hard here. Reason is that Prometheus can be restarting, replaying WAL e.g and you don't want cascading failure on sidecar in this case.

povilasv:
@bwplotka :+1: Makes sense.

johncming:
> Yes, this might be confusing but see a few lines above:
> `// Retry trigger reload until it succeeded or next tick is near.` And by tick we mean here this tick:
> 
> https://github.com/thanos-io/thanos/blob/bd9aa1b4be3bb5d841cb7271c29d02ebb5eb5168/pkg/reloader/reloader.go#L164
> 
> ..not runutil one @povilasv
> 
> It's not timeout of rules apply. It's rather a continuous retry combined with watch interval.
> 
> _If we want to introduce timeout and fail_ we need to do it properly instead of just adding error here. For example allowing user to have special timeout for it.
> 
> I would vote howevr for not failing hard here. Reason is that Prometheus can be restarting, replaying WAL e.g and you don't want cascading failure on sidecar in this case.

Failure after retry should be the more intuitive way of sidecar pattern. If the main instance goes down and sidecar fails after retrying for a while,it can find the problem early, especially in the k8s environment, so that the pod can be rescheduled quickly.

bwplotka:
If the main instance goes down (Prometheus), it does not scrape the data, and you should have other means of meta monitoring (e.g DeadManSwitch alert, other Prometheus) to detect this state, then just realy on lack reloading or inability to query through sidecar.

> Sidecar is not responsible for the up-and-down of Prometheus, so it makes little sense for the sidecar to survive a master instance outage

True, but if we design sidecar let's say proxy like envoy, does it fail if the main instance is down as well? No, it just returns failure on request. It does not make sense to fail on every blip of main instance, it just increases the burden, reduces debuggability in case of the incident and increases chances of cascading failures.

> Prometheus in the sidecar scenario would not have a large wal (2 hours), and the slow response from replay wal should not exist

Can you elaborate? Prometheus with sidecar has slow replay as well, as it has exactly the same WAL (: 

I think its a general sidecar behavior question asked here and it's totally fair to agree on those and have consistent behavior. This is somehow similar to another PR you have [here](https://github.com/thanos-io/thanos/pull/1834). Let's have some agreement (:

johncming:
> If the main instance goes down (Prometheus), it does not scrape the data, and you should have other means of meta monitoring (e.g DeadManSwitch alert, other Prometheus) to detect this state, then just realy on lack reloading or inability to query through sidecar.
> 
> > Sidecar is not responsible for the up-and-down of Prometheus, so it makes little sense for the sidecar to survive a master instance outage
> 
> True, but if we design sidecar let's say proxy like envoy, does it fail if the main instance is down as well? No, it just returns failure on request. It does not make sense to fail on every blip of main instance, it just increases the burden, reduces debuggability in case of the incident and increases chances of cascading failures.
> 
> > Prometheus in the sidecar scenario would not have a large wal (2 hours), and the slow response from replay wal should not exist
> 
> Can you elaborate? Prometheus with sidecar has slow replay as well, as it has exactly the same WAL (:
> 
> I think its a general sidecar behavior question asked here and it's totally fair to agree on those and have consistent behavior. This is somehow similar to another PR you have [here](https://github.com/thanos-io/thanos/pull/1834). Let's have some agreement (:



> If the main instance goes down (Prometheus), it does not scrape the data, and you should have other means of meta monitoring (e.g DeadManSwitch alert, other Prometheus) to detect this state, then just realy on lack reloading or inability to query through sidecar.
> 
> > Sidecar is not responsible for the up-and-down of Prometheus, so it makes little sense for the sidecar to survive a master instance outage
> 
> True, but if we design sidecar let's say proxy like envoy, does it fail if the main instance is down as well? No, it just returns failure on request. It does not make sense to fail on every blip of main instance, it just increases the burden, reduces debuggability in case of the incident and increases chances of cascading failures.
> 
> > Prometheus in the sidecar scenario would not have a large wal (2 hours), and the slow response from replay wal should not exist
> 
> Can you elaborate? Prometheus with sidecar has slow replay as well, as it has exactly the same WAL (:
> 
> I think its a general sidecar behavior question asked here and it's totally fair to agree on those and have consistent behavior. This is somehow similar to another PR you have [here](https://github.com/thanos-io/thanos/pull/1834). Let's have some agreement (:

Yes ,after these discussions, the sidecar model is understood differently. The sidecar that I understand is the sidecar pattern, the symbiotic relationship. You understand the proxy or ambassador pattern. The literal meaning of sidecar is easy to misunderstand, but I still think the sidecar model might be more suitable, thanos-sidecar&prometheus  is 1:1 not 1:N. Take kube-proxy as an example. After the machine is hung, kube-proxy cannot run naturally.

https://docs.microsoft.com/en-us/azure/architecture/patterns/sidecar

PR #1825, opened 2019-12-03
Store: make index cache shared cache friendly

### LINKED ISSUE #1815: Make the IndexCache shared cache friendly
Currently, the `IndexCache` is an in-memory cache whose exposed interface allow to cache single Postings and Series. In Cortex we've an experimental TSDB blocks storage integration based on Thanos and we would like to introduce a shared index cache (based on Memcached).

The current `IndexCache` interface is not shared cache friendly, due to the fact that the exposed functions allow to query/cache single items. In a shared cache implementation, this would require a single round trip for each cache key. We would like to change the `IndexCache` to `MultiPostings()` / `MultiSeries()` to make it shared cache friendly and - before working on it - we would like to reach some consensus.

## Proposal

We propose to switch the `IndexCache` to the following interface:

```
type IndexCache interface {
    // MultiPostings fetches multiple postings - each identified by a labels set -
    // and return a map containing cache hits. In case of no cache hit, an empty
    // map is returned.
    MultiPostings(b ulid.ULID, keys []labels.Label) (map[label.Label][]byte, error)

    // MultiSeries fetches multiple series - each identified by ID - from the cache
    // and return a map containing cache hits. In case of no cache hit, an empty
    // map is returned.
    MultiSeries(b ulid.ULID, ids []uint64) (map[uint64][]byte, error)

    SetPostings(b ulid.ULID, l labels.Label, v []byte)
    SetSeries(b ulid.ULID, id uint64, v []byte)
}
```

- Read path: load postings and series - for a single block - within a single round trip
- Write path: keep setters for single cache keys because (a) ease to integrate (nothing to do) (b) memcached doesn't support multi set (c) no synchronous requirements in the cache write path, so multiple `SetSeries()` for the same block can be shortly buffered in the cache implementation and then stored with a single multi set API call (ie. supported by Redis)

_We currently propose to only change the `IndexCache` interface to allow Cortex using a custom index cache based on Memcached, but we're open to work on adding a shared cache support to Thanos too in subsequent PRs if there's some interest._

pracucci:
The `IndexCache` is an in-memory cache whose exposed interface allow to cache single Postings and Series. In Cortex we've an experimental TSDB blocks storage integration based on Thanos and we would like to introduce a shared index cache (based on Memcached).

The current `IndexCache` interface is not shared cache friendly, due to the fact that the exposed functions allow to fetch/store single items. In this PR I'm proposing to switch `IndexCache` to an interface which allows to fetch multiple postings and series at once, in order to make it shared cache friendly.

Fixed #1815 

## Changes

- Renamed store `cache.go` to `inmemory.go`
- Refactored `IndexCache` to multiple fetch postings and series within a single call
- Prefixed fetch/store functions with `Fetch` and `Store` for clarity

Due to the file renaming, the diff is quite difficult to follow, reason why I've splitted into two separate commits (it should help).

## Verification

Unit and manual tests.


pkg/store/cache/inmemory.go
```diff
@@ -0,0 +1,314 @@
...
+	c.curSize = 0
+}
+
+// StorePostings sets the postings identified by the ulid and label to the value v,
+// if the postings already exists in the cache it is not mutated.
+func (c *InMemoryIndexCache) StorePostings(blockID ulid.ULID, l labels.Label, v []byte) {
+	c.set(cacheTypePostings, cacheKey{blockID, cacheKeyPostings(l)}, v)
+}
+
+func (c *InMemoryIndexCache) FetchMultiPostings(blockID ulid.ULID, keys []labels.Label) (hits map[labels.Label][]byte, misses []labels.Label) {
```
bwplotka: A comment would be nice on exported method (:

pracucci:
Thanks @bwplotka for your review! What's about now?

bwplotka:
Thanks!

PR #1826, opened 2019-12-03
comment whole alert rule

daixiang0:

Comment the alert rule config file format for beginners.

### DIFF
--- a/docs/components/rule.md
+++ b/docs/components/rule.md
@@ -43,6 +43,90 @@ For Ruler the read path is distributed, since most likely Ruler is querying Than
 This means that **query failure** are more likely to happen, that's why clear strategy on what will happen to alert and during query
 unavailability is the key.
 
+
+## Configuring Rules
+
+
+Rule files use YAML, the syntax of a rule file is:
+
+```
+groups:
+  [ - <rule_group> ]
+```
+
+A simple example rules file would be:
+
+```
+groups:
+  - name: example
+    rules:
+    - record: job:http_inprogress_requests:sum
+      expr: sum(http_inprogress_requests) by (job)
+```
+
+<rule_group>
+
+```
+# The name of the group. Must be unique within a file.
+name: <string>
+
+# How often rules in the group are evaluated.
+[ interval: <duration> | default = global.evaluation_interval ]
+
+rules:
+  [ - <rule> ... ]
+```
+
+Thanos supports two types of rules which may be configured and then evaluated at regular intervals: recording rules and alerting rules.
+
+### Recording Rules
+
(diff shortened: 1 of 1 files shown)


examples/alerts/alerts.md
```diff
@@ -1,6 +1,19 @@
 # Alerts
 
 Here are some example alerts configured for Kubernetes environment.
+The whole alert rule config file is as following:
+```
```
bwplotka: New line has to before the code snippet (:

examples/alerts/alerts.md
```diff
@@ -1,6 +1,19 @@
...
+groups:
+- name: rule1
```
bwplotka:
```suggestion
- name: alert_group1
```
and ditto

daixiang0:
It is in https://thanos.io/getting-started.md/#operating, not sure whether move to rule component.

daixiang0:
mind review again? @bwplotka

examples/alerts/alerts.md
bwplotka: 1. Newline is not addressed. Markdown requires this. 
2. The reason why it would be better on the ruler, as those are alerts/rules *specifically* in Thanos Ruler. Examples of alerts here are mainly to operate Thanos, but it does not mean you have to set up alerts on Thanos Ruler straight away! (:  
3. Let's mention recording rules as well (:

daixiang0:
What are recording rules? Could share more details? @bwplotka

squat:
@daixiang0 take a look at the Prometheus docs: https://prometheus.io/docs/prometheus/latest/configuration/recording_rules/

daixiang0:
@bwplotka @squat Add how to configure rule config, including recording rule and alerting rules. please review it.

daixiang0:
@bwplotka @squat any input?

daixiang0:
@bwplotka @squat mind have a look?

bwplotka:
Thanks for this PR (: 

I think we are already fixing this here: https://github.com/thanos-io/thanos/pull/1908 

The PR is really similar, just that we have single function vs splitted into two here (where one function is not really used), let's merge one PR to fix current problem, I hope you don't mind merging https://github.com/thanos-io/thanos/pull/1908 as it's already ready to go (:

docs/components/rule.md
```diff
@@ -43,6 +43,90 @@ For Ruler the read path is distributed, since most likely Ruler is querying Than
...
+# Labels to add or overwrite before storing the result.
+labels:
+  [ <labelname>: <labelvalue> ]
+```
+
+### Alerting Rules
+
+The syntax for alerting rules is:
+
+```
```
bwplotka: We could refer just the rules and alerts docs from Prometheus.. but this will do as well (: Thanks.

PR #1827, opened 2019-12-03
add partial response strategy values

daixiang0:

- add support value
- remove duplicate link


bwplotka:
Thanks (:

PR #1834, opened 2019-12-04
cmd/thanos: add labels len check when running.

johncming:
The external labels check of Prometheus may also become illegal during operation after first passing.

https://github.com/thanos-io/thanos/blob/7e11afe64af0c096743a3de8a594616abf52be45/cmd/thanos/sidecar.go#L208-L227

### DIFF
--- a/cmd/thanos/sidecar.go
+++ b/cmd/thanos/sidecar.go
@@ -223,6 +223,10 @@ func runSidecar(
 					lastHeartbeat.Set(float64(time.Now().UnixNano()) / 1e9)
 				}
 
+				if len(m.Labels()) == 0 {
+					return errors.New("no external labels configured on Prometheus server, uniquely identifying external labels must be configured")
+				}
+
 				return nil
 			})
 		}, func(error) {


johncming:
> True, but do we really want to fail here and thus introducing unavailability for querying as well as write path (uploading block)? In the same way, we don't crash sidecar as soon as we cannot reach prometheus (aka `heatbeat`)
> 
> I would start with warning here or maybe _NOT_ updating labels if empty?

With thanos sidecar, it is a serious mistake to remove the external label of Prometheus at run time, crash sidecar is a way.

https://github.com/thanos-io/thanos/blob/7e11afe64af0c096743a3de8a594616abf52be45/cmd/thanos/sidecar.go#L208-L210

bwplotka:
> With thanos sidecar, it is a serious mistake to remove the external label of Prometheus at run time, crash sidecar is a way. Let serious errors be reported as soon as possible

How serious? Serious enough to disrupt query availability? In the worst case, it will be caught in compactor when compacting and advertised badly. Still recoverable by editing meta.json manually in S3. I agree it can be inconvenient. We should have metric and alert for that, the same way Prometheus does not accept the corrupted configuration and just increment metric (: 

Also if we would avoid to update labels, we have old behaviour as well and similar to broken configuration on Prometheus. 

Happy to discuss it more with others. @GiedriusS @brancz @squat ? 

BTW overall thanks @johncming for contributions, it's definitely worth to have clear consistent behavior for those (:

johncming:
> > With thanos sidecar, it is a serious mistake to remove the external label of Prometheus at run time, crash sidecar is a way. Let serious errors be reported as soon as possible
> 
> How serious? Serious enough to disrupt query availability? In the worst case, it will be caught in compactor when compacting and advertised badly (broader than it can). Still recoverable by editing meta.json manually in S3. I agree it can be inconvenient. We should have a metric and alert for that, the same way Prometheus does not accept the corrupted configuration and just increment metric (:
> 
> Also if we would avoid updating empty labels, we have old behavior as well and similar to the broken configuration on Prometheus.
> 
> Happy to discuss it more with others. @GiedriusS @brancz @squat ?
> 
> BTW overall thanks @johncming for contributions, it's definitely worth to have clear consistent behavior for those (:

Whether the upload is enabled or not affects the running state of thanos sidecar.

Three of the following
1. Disable upload. Sidecar can run forever if it writes the wrong address for Prometheus
2. Disable upload, sidecar will exit quickly if sidecar points to the address of Prometheus (but no external labels are configured)
3. Enable upload. If sidecar points to the address of Prometheus (but with external labels configured), sidecar will exit after a while

If the user does not configure metrics, thanos sidecar will be surprised by its strange behavior

if upload enabled and prometheus missing external labels, thanos sidecar will exit.

https://github.com/thanos-io/thanos/blob/e94478c81c2615adde63ed56930ee0114b7b3c5a/cmd/thanos/sidecar.go#L293-L297

daixiang0:
@bwplotka any input?

PR #1835, opened 2019-12-04
comment url scheme

daixiang0:

Since now alertmanager url is auto parse into url.URL, the scheme should exist and not empty, or would parse warn:
```
thanos rule --alertmanagers.url=localhost:9093 --rule-file=rule.yaml --query=localhost:19999
...
level=warn ts=2019-12-04T03:45:28.597397054Z caller=alert.go:342 msg="sending alerts failed" alertmanager= numAlerts=2 err="send request to \"localhost:///api/v1/alerts\": Post 
localhost:///api/v1/alerts: unsupported protocol scheme \"localhost\""
```

Without scheme, this warn is error indeed since can not send alerts forever.

### DIFF
--- a/cmd/thanos/rule.go
+++ b/cmd/thanos/rule.go
@@ -81,7 +81,7 @@ func registerRule(m map[string]setupFunc, app *kingpin.Application) {
 	tsdbRetention := modelDuration(cmd.Flag("tsdb.retention", "Block retention time on local disk.").
 		Default("48h"))
 
-	alertmgrs := cmd.Flag("alertmanagers.url", "Alertmanager replica URLs to push firing alerts. Ruler claims success if push to at least one alertmanager from discovered succeeds. The scheme may be prefixed with 'dns+' or 'dnssrv+' to detect Alertmanager IPs through respective DNS lookups. The port defaults to 9093 or the SRV record's value. The URL path is used as a prefix for the regular Alertmanager API path.").
+	alertmgrs := cmd.Flag("alertmanagers.url", "Alertmanager replica URLs to push firing alerts. Ruler claims success if push to at least one alertmanager from discovered succeeds. The scheme should not be empty e.g `http` might be used. The scheme may be prefixed with 'dns+' or 'dnssrv+' to detect Alertmanager IPs through respective DNS lookups. The port defaults to 9093 or the SRV record's value. The URL path is used as a prefix for the regular Alertmanager API path.").
 		Strings()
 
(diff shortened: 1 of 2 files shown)


daixiang0:
Not sure update doc or code since there is a plan to use original alertmanager send.

bwplotka:
What about putting just `http`? (:

daixiang0:
Do you mean force specify this option start with `http` and abandon dns thing?

bwplotka:
No, I mean if you put `http` it will work as expected, while this changes the doc in a way that you have to use `dns` which is not true

daixiang0:
Of course use `http://localhost:9093` works.
This pr means can not specify alertmanger url without scheme while https may not support.

cmd/thanos/rule.go
```diff
@@ -81,7 +81,7 @@ func registerRule(m map[string]setupFunc, app *kingpin.Application) {
 	tsdbRetention := modelDuration(cmd.Flag("tsdb.retention", "Block retention time on local disk.").
 		Default("48h"))
 
-	alertmgrs := cmd.Flag("alertmanagers.url", "Alertmanager replica URLs to push firing alerts. Ruler claims success if push to at least one alertmanager from discovered succeeds. The scheme may be prefixed with 'dns+' or 'dnssrv+' to detect Alertmanager IPs through respective DNS lookups. The port defaults to 9093 or the SRV record's value. The URL path is used as a prefix for the regular Alertmanager API path.").
+	alertmgrs := cmd.Flag("alertmanagers.url", "Alertmanager replica URLs to push firing alerts. Ruler claims success if push to at least one alertmanager from discovered succeeds. The scheme should be not empty, The scheme may be prefixed with 'dns+' or 'dnssrv+' to detect Alertmanager IPs through respective DNS lookups. The port defaults to 9093 or the SRV record's value. The URL path is used as a prefix for the regular Alertmanager API path.").
```
bwplotka:
```suggestion
	alertmgrs := cmd.Flag("alertmanagers.url", "Alertmanager replica URLs to push firing alerts. Ruler claims success if push to at least one alertmanager from discovered succeeds. The scheme should be not empty e.g `http` might be used. The scheme may be prefixed with 'dns+' or 'dnssrv+' to detect Alertmanager IPs through respective DNS lookups. The port defaults to 9093 or the SRV record's value. The URL path is used as a prefix for the regular Alertmanager API path.").
```

bwplotka: What about this? (:

daixiang0: It is clear enough for users, done.

squat: Grammar nit: `should not be empty`

daixiang0:
mind review again? @bwplotka

PR #1836, opened 2019-12-04
pkg/query: add lock sync.

johncming:
There are concurrent access conflicts, `s.cleanUpStoreStatuses(stores)` need lock sync.

https://github.com/thanos-io/thanos/blob/7e11afe64af0c096743a3de8a594616abf52be45/pkg/query/storeset.go#L376-L381

goroutine 1
https://github.com/thanos-io/thanos/blob/7e11afe64af0c096743a3de8a594616abf52be45/cmd/thanos/query.go#L301

goroutine 2
https://github.com/thanos-io/thanos/blob/7e11afe64af0c096743a3de8a594616abf52be45/cmd/thanos/query.go#L333

### DIFF
--- a/pkg/query/storeset.go
+++ b/pkg/query/storeset.go
@@ -376,9 +376,8 @@ func (s *StoreSet) Update(ctx context.Context) {
 	s.storesMetric.Update(stats)
 	s.storesMtx.Lock()
 	s.stores = stores
-	s.storesMtx.Unlock()
-
 	s.cleanUpStoreStatuses(stores)
+	s.storesMtx.Unlock()
 }
 
 func (s *StoreSet) getHealthyStores(ctx context.Context, stores map[string]*storeRef) map[string]*storeRef {


johncming:
> Thanks!
> 
> I don't think this is needed.
> What's the reason, what triggered this change? Any particular failure? (:
> 
> `cleanUpStoreStatuses` does not use`s.stores` and we never update s.stores in place, always regenerate new. Plus we have another mutex for statuses inside.

but it updates `storeStatuses` which should sync access.
https://github.com/thanos-io/thanos/blob/7e11afe64af0c096743a3de8a594616abf52be45/pkg/query/storeset.go#L508-L522

bwplotka:
That's why we have
```
s.storesStatusesMtx.Lock() 
defer s.storesStatusesMtx.Unlock() 
```

johncming:
> That's why we have
> 
> ```
> s.storesStatusesMtx.Lock() 
> defer s.storesStatusesMtx.Unlock() 
> ```

The `store` parameter is not protected.

Consider the following execution order:

1. goroutine 1 running

```
	s.storesMetric.Update(stats)
	s.storesMtx.Lock()
	s.stores = stores
	s.storesMtx.Unlock()
```

2. goroutine 2 running

```
	s.storesMetric.Update(stats)
	s.storesMtx.Lock()
	s.stores = stores
	s.storesMtx.Unlock()
```

3. goroutine 1 running

```
	s.cleanUpStoreStatuses(stores)
```

but the current value of `s.stores` is from goroutine 2 not goroutine 1 which cause problem.

bwplotka:
That's also not correct as we have this on top of `Update` (:  
```
func (s *StoreSet) Update(ctx context.Context) {
	s.updateMtx.Lock()
	defer s.updateMtx.Unlock()
```

PR #1837, opened 2019-12-04
skip the error of unmarshal for index-cache file

### LINKED ISSUE #1810: Clean partially downloaded index-cache.json
Looks like we missed this: https://cloud-native.slack.com/archives/CK5RSSC10/p1574932985449600

In case of crash do index-cache.JSON being partially downloaded/created we have garbage that store GW cannot handle right now. This requires manual cleanup of local disk which is not great.

AC:
* Handle partial index-cache JSON on store GW

fatsheep9146:
fix #1810

### DIFF
--- a/pkg/block/index.go
+++ b/pkg/block/index.go
@@ -32,6 +32,10 @@ const (
 	IndexCacheVersion1 = iota + 1
 )
 
+const (
+	UnmarshalErrorMsg = "unmarshal index cache"
+)
+
 type postingsRange struct {
 	Name, Value string
 	Start, End  int64
@@ -194,7 +198,7 @@ func ReadIndexCache(logger log.Logger, fn string) (
 	}
 
 	if err = json.Unmarshal(bytes, &v); err != nil {
-		return 0, nil, nil, nil, errors.Wrap(err, "unmarshal index cache")
+		return 0, nil, nil, nil, errors.Wrap(err, UnmarshalErrorMsg)
 	}
 
 	strs := map[string]string{}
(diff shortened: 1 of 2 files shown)


fatsheep9146:
@bwplotka

pkg/block/index.go
```diff
@@ -239,6 +243,10 @@ func ReadIndexCache(logger log.Logger, fn string) (
 	return v.Version, symbols, lvals, postings, nil
 }
 
+func IsUnmarshalError(err error) bool {
+	return strings.Contains(err.Error(), UnmarshalErrorMsg)
```
bwplotka: I think we can do better than just parsing the message.

I would propose sentinel error instead:

`IndexCacheUnmarshalError = errors.New("....")`

and then on line 201:  `errors.Wrap(IndexCacheUnmarshalError, err.Error())`

and then instead of this function user can do `if !os.IsNotExist(errors.Cause(err)) && errors.Cause(err) =! block.IndexCacheUnmarshalError{` hm? (:

The problem with string parsing is that it could be unsafe e.g we could catch some unexpected errors there. Let's be explicit.

fatsheep9146: That sounds reasonable, I will improve it like that. (:

fatsheep9146:
Sorry for my delay, I originally meant to sentinel all errors in package `github.com/thanos-io/thanos/pkg/block`, but I found that there are too many. And so far I cannot tell which error should be exported or not. So I just sentinel `IndexCacheUnmarshalError`. What do you think about this? @bwplotka

PR #1838, opened 2019-12-04
*: support authentication and TLS for Alertmanager

### LINKED ISSUE #606: [Feature] Rule: Add authentication for alertmanagers
It would be great to add possibility to authenticate to Alertmanagers from Rule node when sending alerts the same way as Prometheus allows.

With current configuration using `--alertmanagers.url` it would have to be common authentication configuration for all those AM I suppose.

Prometheus now supports authentication using client certificate, BasicAuth and bearer token.
I'm not saying there should be all of them but at least one of them would be great.

simonpasquier:
Closes #606 

## Changes

This change adds support for authentication with basic auth, client
certificates and bearer tokens. It also enables to configure TLS
settings for the Alertmanager endpoints.

Most of the work leverages the existing Prometheus configuration format
and code. In particular TLS certificate files are automatically reloaded
whenever they change.

## Verification

End-to-end tests added to cover various HTTP client configurations (TLS, authentication) and file SD integration.

### DIFF
--- a/cmd/thanos/rule.go
+++ b/cmd/thanos/rule.go
@@ -296,14 +296,14 @@ func runRule(
 		return err
 	}
 	var (
-		alertingcfg alert.AlertingConfig
+		alertingCfg alert.AlertingConfig
 		alertmgrs   []*alert.Alertmanager
 	)
 	if len(alertmgrsConfigYAML) > 0 {
 		if len(alertmgrURLs) != 0 {
 			return errors.New("--alertmanagers.url and --alertmanagers.config* flags cannot be defined at the same time")
 		}
-		alertingcfg, err = alert.LoadAlertingConfig(alertmgrsConfigYAML)
+		alertingCfg, err = alert.LoadAlertingConfig(alertmgrsConfigYAML)
 		if err != nil {
 			return err
 		}
@@ -314,13 +314,13 @@ func runRule(
 			if err != nil {
 				return err
 			}
-			alertingcfg.Alertmanagers = append(alertingcfg.Alertmanagers, cfg)
+			alertingCfg.Alertmanagers = append(alertingCfg.Alertmanagers, cfg)
 		}
 	}
-	if len(alertingcfg.Alertmanagers) == 0 {
+	if len(alertingCfg.Alertmanagers) == 0 {
 		level.Warn(logger).Log("msg", "no alertmanager configured")
 	}
-	for _, cfg := range alertingcfg.Alertmanagers {
+	for _, cfg := range alertingCfg.Alertmanagers {
 		am, err := alert.NewAlertmanager(logger, cfg)
 		if err != nil {
 			return err
@@ -421,11 +421,11 @@ func runRule(
 	}
 	// Run the alert sender.
 	{
-		doers := make([]alert.AlertmanagerDoer, len(alertmgrs))
+		clients := make([]alert.AlertmanagerClient, len(alertmgrs))
 		for i := range alertmgrs {
-			doers[i] = alertmgrs[i]
+			clients[i] = alertmgrs[i]
 		}
-		sdr := alert.NewSender(logger, reg, doers)
+		sdr := alert.NewSender(logger, reg, clients)
 		ctx, cancel := context.WithCancel(context.Background())
 
 		g.Add(func() error {
--- a/cmd/thanos/rule.go
+++ b/cmd/thanos/rule.go
@@ -328,7 +328,7 @@ func runRule(
 		dns.ResolverType(dnsSDResolver),
 	)
 	for _, cfg := range alertingCfg.Alertmanagers {
-		// Each Alertmanager client needs its own DNS provider.
+		// Each Alertmanager client has a different list of targets thus each needs its own DNS provider.
 		am, err := alert.NewAlertmanager(logger, cfg, amProvider.Clone())
 		if err != nil {
 			return err
--- a/cmd/thanos/rule.go
+++ b/cmd/thanos/rule.go
@@ -297,7 +297,7 @@ func runRule(
 	)
 	if len(alertmgrsConfigYAML) > 0 {
 		if len(alertmgrURLs) != 0 {
-			level.Warn(logger).Log("msg", "ignoring --alertmanagers.url flag because --alertmanagers.config* flag is also defined")
+			return errors.New("--alertmanagers.url and --alertmanagers.config* flags cannot be defined at the same time")
 		}
 		alertingcfg, err = alert.LoadAlertingConfig(alertmgrsConfigYAML)
 		if err != nil {
--- a/cmd/thanos/rule.go
+++ b/cmd/thanos/rule.go
@@ -317,11 +317,19 @@ func runRule(
 			alertingCfg.Alertmanagers = append(alertingCfg.Alertmanagers, cfg)
 		}
 	}
+
 	if len(alertingCfg.Alertmanagers) == 0 {
 		level.Warn(logger).Log("msg", "no alertmanager configured")
 	}
+
+	amProvider := dns.NewProvider(
+		logger,
+		extprom.WrapRegistererWithPrefix("thanos_ruler_alertmanagers_", reg),
+		dns.ResolverType(dnsSDResolver),
+	)
 	for _, cfg := range alertingCfg.Alertmanagers {
-		am, err := alert.NewAlertmanager(logger, cfg)
+		// Each Alertmanager client needs its own DNS provider.
+		am, err := alert.NewAlertmanager(logger, cfg, amProvider.Clone())
 		if err != nil {
 			return err
 		}
@@ -394,8 +402,6 @@ func runRule(
 	}
 	// Discover and resolve Alertmanager addresses.
 	{
-		resolver := dns.NewResolver(dns.ResolverType(dnsSDResolver).ToResolver(logger))
-
 		for i := range alertmgrs {
 			am := alertmgrs[i]
 			ctx, cancel := context.WithCancel(context.Background())
@@ -408,10 +414,7 @@ func runRule(
 
 			g.Add(func() error {
 				return runutil.Repeat(alertmgrsDNSSDInterval, ctx.Done(), func() error {
-					if err := am.Update(ctx, resolver); err != nil {
-						level.Error(logger).Log("msg", "refreshing alertmanagers failed", "err", err)
-						alertMngrAddrResolutionErrors.Inc()
-					}
+					am.Resolve(ctx)
 					return nil
 				})
 			}, func(error) {
--- a/cmd/thanos/rule.go
+++ b/cmd/thanos/rule.go
@@ -4,7 +4,6 @@ import (
 	"context"
 	"fmt"
 	"math/rand"
-	"net"
 	"net/http"
 	"net/url"
 	"os"
@@ -13,7 +12,6 @@ import (
 	"path/filepath"
 	"strconv"
 	"strings"
-	"sync"
 	"syscall"
 	"time"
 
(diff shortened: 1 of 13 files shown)


docs/components/rule.md
```diff
@@ -265,3 +281,35 @@ Flags:
...
 
 ```
+
+## Configuration
+
+### Alertmanager
+
+The configuration format supported by the `--alertmanagers.config` and `--alertmanagers.config-file` flags is the following:
+
+[embedmd]:# (../flags/config_rule_alerting.txt yaml)
```
bwplotka: nice!

scripts/cfggen/main.go
```diff
@@ -113,16 +121,12 @@ func checkForOmitEmptyTagOptionRec(v reflect.Value) error {
...
 			}
 		}
 
 	case reflect.Ptr:
-		if !v.IsValid() {
-			return errors.New("nil pointers are not allowed in configuration.")
-		}
-
-		return errors.New("nil pointers are not allowed in configuration.")
+		return errors.New("nil pointers are not allowed in configuration")
```
bwplotka: lol this was funny (:

simonpasquier:
@bwplotka it's ready now.

pkg/alert/client_test.go
```diff
@@ -0,0 +1,183 @@
...
+		},
+		{
+			address: "ssh+http://localhost",
+			expected: AlertmanagerConfig{
+				StaticAddresses: []string{"localhost"},
+				Scheme:          "ssh+http",
+			},
+		},
+		{
+			address: "dns+https://localhost/path/prefix/",
```
bwplotka: Currently, we did not even support paths, so that's nice :+1:  worth to mention in CHANGELOG I think.

pkg/alert/client_test.go
```diff
@@ -0,0 +1,183 @@
...
+	}{
+		{
+			address: "http://localhost:9093",
+			expected: AlertmanagerConfig{
+				StaticAddresses: []string{"localhost:9093"},
+				Scheme:          "http",
+			},
+		},
+		{
+			address: "https://am.example.com",
```
bwplotka: Currently, we did not support HTTPS, so that's nice as well :+1:  worth to mention in CHANGELOG I think, similar to comment below

docs/service-discovery.md
```diff
@@ -77,6 +77,8 @@ Again, the `<path>` can be a glob pattern.
 
 The flag `--query.sd-interval=<5m>` can be used to change the fallback re-read interval.
 
+`Thanos Rule` also supports the configuration of Alertmanager endpoints using YAML with the `--alertmanagers.config=<content>` and `--alertmanagers.config-file=<path>` flags.
```
bwplotka:
```suggestion
`Thanos Rule` also supports the configuration of Alertmanager endpoints using YAML with the `--alertmanagers.config=<content>` and `--alertmanagers.config-file=<path>` flags in the `FileSDfiles` section.
```

docs/service-discovery.md
```diff
@@ -33,7 +33,7 @@ The repeatable flag `--store=<store>` can be used to specify a `StoreAPI` that `
 
 The repeatable flag `--query=<query>` can be used to specify a `QueryAPI` that `Thanos Rule` should use.
 
-The repeatable flag `--alertmanager.url=<alertmanager>` can be used to specify a `Alertmanager API` that `Thanos Rule` should use.
+The repeatable flag `--alertmanagers.url=<alertmanager>` can be used to specify a `Alertmanager API` that `Thanos Rule` should use.
```
bwplotka:
```suggestion
`Thanos Rule` also supports the configuration of Alertmanager endpoints using YAML with the `--alertmanagers.config=<content>` and `--alertmanagers.config-file=<path>` flags in the `StaticAddress` section.
```

cmd/thanos/rule.go
```diff
@@ -83,8 +81,10 @@ func registerRule(m map[string]setupFunc, app *kingpin.Application) {
 
 	alertmgrs := cmd.Flag("alertmanagers.url", "Alertmanager replica URLs to push firing alerts. Ruler claims success if push to at least one alertmanager from discovered succeeds. The scheme should not be empty e.g `http` might be used. The scheme may be prefixed with 'dns+' or 'dnssrv+' to detect Alertmanager IPs through respective DNS lookups. The port defaults to 9093 or the SRV record's value. The URL path is used as a prefix for the regular Alertmanager API path.").
```
bwplotka: I think this is no longer true: `The scheme should not be empty e.g `http` might be used.` as your unit tests shows (:

pkg/alert/client_test.go
```diff
@@ -0,0 +1,183 @@
...
+		{
+			address: "dns+https://localhost/path/prefix/",
+			expected: AlertmanagerConfig{
+				StaticAddresses: []string{"dns+localhost"},
+				Scheme:          "https",
+				PathPrefix:      "/path/prefix/",
+			},
+		},
+		{
+			address: "http://user:pass@localhost:9093",
```
bwplotka: Same here I think - a new feature to `alertmanagers.url` flag.

I believe it would be nice to support this flag on top of configuration (we already document it as normal flag and not deprecated) so that's why it would be nice to mention this in CHANGELOG (:

pkg/alert/client_test.go
```diff
@@ -0,0 +1,183 @@
...
+					},
+				},
+				StaticAddresses: []string{"localhost:9093"},
+				Scheme:          "http",
+			},
+		},
+	} {
+		t.Run(tc.address, func(t *testing.T) {
+			cfg, err := BuildAlertmanagerConfig(nil, tc.address, time.Duration(0))
+			if tc.err {
```
bwplotka: hm, no error case though (: Can we create one or two?

test/e2e/rule_test.go
```diff
@@ -62,19 +70,321 @@ func createRuleFiles(t *testing.T, dir string) {
...
+}
+
+func writeAlertmanagerFileSD(t *testing.T, path string, addrs ...string) {
+	group := targetgroup.Group{Targets: []model.LabelSet{}}
+	for _, addr := range addrs {
+		group.Targets = append(group.Targets, model.LabelSet{model.LabelName(model.AddressLabel): model.LabelValue(addr)})
+	}
+
+	b, err := yaml.Marshal([]*targetgroup.Group{&group})
+	if err != nil {
```
bwplotka: Why not `testutil.Ok` here?

bwplotka: and below?

test/e2e/rule_test.go
```diff
@@ -62,19 +70,321 @@ func createRuleFiles(t *testing.T, dir string) {
...
+		return
+	}
+
+	err = ioutil.WriteFile(path+".tmp", b, 0660)
+	if err != nil {
+		t.Errorf("failed to write file SD configuration: %v", err)
+		return
+	}
+
+	err = os.Rename(path+".tmp", path)
```
bwplotka: let's inline

test/e2e/rule_test.go
```diff
@@ -62,19 +70,321 @@ func createRuleFiles(t *testing.T, dir string) {
...
+	rulesDir, err := ioutil.TempDir("", "rules")
+	defer os.RemoveAll(rulesDir)
+	testutil.Ok(t, err)
+	createRuleFiles(t, rulesDir)
+
+	qAddr := a.New()
+	r := rule(a.New(), a.New(), rulesDir, amCfg, []address{qAddr}, nil)
+	q := querier(qAddr, a.New(), []address{r.GRPC}, nil)
+
+	ctx, cancel := context.WithTimeout(context.Background(), 1*time.Minute)
```
bwplotka: `1*time.Minute` This might be not enough for our sometimes slow CI, let's make it 3m

test/e2e/rule_test.go
```diff
@@ -62,19 +70,321 @@ func createRuleFiles(t *testing.T, dir string) {
...
+		}
+		if len(alrts) != 0 {
+			return errors.Errorf("unexpected alerts length %d", len(alrts))
+		}
+
+		return nil
+	}))
+
+	// Update the Alertmanager file service discovery configuration.
+	writeAlertmanagerFileSD(t, filepath.Join(amDir, "targets.yaml"), am.HTTP.HostPort())
```
bwplotka:
```suggestion
	writeRulerAlertmanagerFileSD(t, filepath.Join(amDir, "targets.yaml"), am.HTTP.HostPort())
```

test/e2e/rule_test.go
```diff
@@ -62,19 +70,321 @@ func createRuleFiles(t *testing.T, dir string) {
...
+			return err
+		}
+		if len(alrts) != 0 {
+			return errors.Errorf("unexpected alerts length %d", len(alrts))
+		}
+
+		return nil
+	}))
+
+	// Update the Alertmanager file service discovery configuration.
```
bwplotka: It sounds like we are updating some `Alertmanager` file SD  not Ruler\s file SD for alertmanager (: Can we clarify a bit?

test/e2e/rule_test.go
```diff
@@ -62,19 +70,321 @@ func createRuleFiles(t *testing.T, dir string) {
...
+		cancel()
+		return
+	}
+
+	defer func() {
+		cancel()
+		<-exit
+	}()
+
+	// Wait for a couple of evaluations.
```
bwplotka: can we comment on what we wait?

test/e2e/rule_test.go
```diff
@@ -62,19 +70,321 @@ func createRuleFiles(t *testing.T, dir string) {
...
+		}
+
+		return nil
+	}))
+}
+
+func TestRuleAlertmanagerFileSD(t *testing.T) {
+	a := newLocalAddresser()
+
+	am := alertManager(a.New())
```
bwplotka: What do you think about this and using alertmanager Mock? I like e2e compatibility check against Alertmanager. I guess it would be too hard to use proper alertmanager in `TestRuleAlertmanagerHTTPClient` as well? (:

pkg/alert/client.go
```diff
@@ -0,0 +1,362 @@
...
+			case <-ctx.Done():
+				return
+			}
+		}
+	}()
+	wg.Wait()
+}
+
+// Update refreshes and resolves the list of targets.
+func (a *Alertmanager) Update(ctx context.Context, resolver dns.Resolver) error {
```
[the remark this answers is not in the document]
simonpasquier: Yes it makes sense. I'll have to extend a bit `dns.Provider` to support default port for A lookups.

pkg/alert/client_test.go
simonpasquier: You could already pass `--alertmanagers.url https://foo.example.com`?

test/e2e/rule_test.go
bwplotka: This makes sense totally, worth to comment maybe? (: 

> Side-note: with the Alertmanager v2 API and its Open API specification, it's even less needed to run a "real" Alertmanager server as you can generate the server code and probably hook into it from the e2e tests.

From API perspective yes (methods, required parameters etc), but it's always nice to have e2e tests against actual implementation. This useful to check against hidden invariants etc.

test/e2e/rule_test.go
bwplotka: still 1m?

cmd/thanos/rule.go
bwplotka: You are right, ignore me (:

test/e2e/rule_test.go
simonpasquier: Agreed. It would be worth revisiting this part when we add support for Alertmanager API v2. I tried quickly to hack something with `httputil.ReverseProxy` but I failed short.

bwplotka:
Rdy for review? (:

simonpasquier:
yep

simonpasquier:
Thanks a lot for the speedy reviews!

pkg/alert/client.go
```diff
@@ -0,0 +1,343 @@
...
+const (
+	defaultAlertmanagerPort = 9093
+	alertPushEndpoint       = "/api/v1/alerts"
+	contentTypeJSON         = "application/json"
+)
+
+var userAgent = fmt.Sprintf("Thanos/%s", version.Version)
+
+type AlertingConfig struct {
+	Alertmanagers []AlertmanagerConfig `yaml:"alertmanagers"`
```
FUSAKLA: Just thinking.. It might be nice to also use the `alert_relabel_configs` instead of the `--alert.label-drop` ?
That would provide higher variability to user.
Also adding external labels to the config instead of the `--label` flag?
Not sure how fare we want to take this configuration @bwplotka WDYT?
(Sorry for late comments, I had no time lately)

PR #1839, opened 2019-12-04
Added proposal for moving to index-header binary format.

bwplotka:
Some verifications in TODO, otherwise. Feedback welcome.

Definitely not a silver bullet here, but rather change in a good direction (:

### DIFF
--- a/docs/proposals/201912_thanos_binary_index_header.md
+++ b/docs/proposals/201912_thanos_binary_index_header.md
@@ -68,14 +68,14 @@ This design is trying to address those 4 problems.
 * Reduce confusion between index-cache.json and IndexCache for series and postings.
 * Do not keep large pieces of the blocks (e.g symbols, label values) in the memory for all blocks. Be able to load it quickly in query time from disk.
 * Use constant amount of memory for building index-header/index-cache; download only required pieces of index from bucket and compute new TOC file.
-* Leverage mmap and let OS to unload unused pages from the physical memory as needed to satisfy the queries against the block.
+* Leverage mmap and let OS unload unused pages from the physical memory as needed to satisfy the queries against the block.
  This is due to random access against label values and postings offsets.
 
 ## No Goals
 
-* Full, no initial startup as designed in [Cortex, no initial block sync](https://github.com/thanos-io/thanos/issues/1813)
+* Removing initial startup for Thanos Store Gateway completely as designed in [Cortex, no initial block sync](https://github.com/thanos-io/thanos/issues/1813)
   * However this proposal might a step towards that as we might be able to load index-cache/index quickly on demand from disk.
-  * In the same time be able to load `index-header` in query time directly from bucket is not a goal of this proposal.
+  * At the same time be able to load `index-header` in query time directly from bucket is not a goal of this proposal.
 * Decreasing size. While it would be nice to use less space; our aim is latency of building/loading the block. That might be correlated with size, but not necessarily (e.g when extra considering compression)
 
 ## Verification
@@ -98,8 +98,9 @@ The process for building this will be as follows:
   * Get the posting table and copy it to the local file.
   * Get the label offset table and copy it to the local file.
   * Get the label index and copy it to the local file.
+  * Open question: How to get [Posting size to fetch](./201912_thanos_binary_index_header.md#posting-size-to-fetch)?
   
-With that effort building time and resources should be compared with downloading the prebuild `index-header` from bucket. This allows us to reduce complexity of the system as 
+With that effort building time and resources should be compared with downloading the prebuilt `index-header` from the bucket. This allows us to reduce complexity of the system as 
 we don't need to cache that in object storage by compactor or even sidecar.
 
 Thanks to this format we can reuse most of the [FileReader](https://github.com/prometheus/prometheus/blob/de0a772b8e7d27dc744810a1a693d97be027049a/tsdb/index/index.go#L664) code
@@ -114,15 +115,38 @@ This proposes following algorithm:
 
 * blocks newer then X (e.g 2w) being always loaded
 * blocks older than X will be loaded on demand on query time and kept cached until evicted.
-* Background eviction process unloads blocks without pending reads to the amount of Y (e.g 100 blocks) in LRU fashion. 
+* background eviction process unloads blocks without pending reads to the amount of Y (e.g 100 blocks) in LRU fashion. 
  
 Both X and Y being configurable.
  
 ## Risks
 
+### Posting size to fetch
+
+While idea of combing different pieces of TSDB index as our index-header is great, unfortunately we heavily rely 
+on information about size of each posting represented as `postingRange.End`. 
+
+We need to know that apriori to know how to partition and how many bytes we need to fetch from the storage to get each posting: https://github.com/thanos-io/thanos/blob/7e11afe64af0c096743a3de8a594616abf52be45/pkg/store/bucket.go#L1567
+
+To calculate those sizes we use [`indexr.PostingsRanges()`](https://github.com/thanos-io/thanos/blob/7e11afe64af0c096743a3de8a594616abf52be45/pkg/block/index.go#L156-L155) which
+scans through `posting` section of the TSDB index. Having to fetch whole postings section just to get size of each posting 
+makes this proposal less valuable as we still need to download big part of index and traverse through it instead of what we propose in [#Proposal](./201912_thanos_binary_index_header.md#proposal)
+
+For series we don't know the exact size either, however we estimate max size of each series to be 64*1024. It depends on sane number of label pairs and chunks per series.
+We have really only one potential case when this was too low: https://github.com/thanos-io/thanos/issues/552. Decision about series this was made here: https://github.com/thanos-io/thanos/issues/146
+
+For postings it's more tricky as it depends on number of series in which given label pair exists. For worst case it can be even million label-pair for popular pairs like `__name__=http_request_http_duration_bucket` etc.
+
+We have few options:
+
+* Encode this posting size in TSDB index `PostingOffset`. 
+* Scan postings to fetch, which is something we wanted to avoid when building `index-header` without downloading full TSDB index.
+* Estimate some large value (will overfetch)
+* Estimate casual value and retry fetching remaining size on demand for data consistency.
+
 ### Unexpected memory usage
 
-Users cares the most about surprising spikes in memory usage. Currently Store Gateway caches the whole index-cache.json. While it's silly to do so for all blocks, this 
+Users care the most about surprising spikes in memory usage. Currently the Store Gateway caches the whole index-cache.json. While it's silly to do so for all blocks, this 
 will happen anyway if query will span over large number of blocks and series. This means that while baseline memory will be reduced, baseline vs request memory difference will be even more noticeable.
 
 This tradeoff is acceptable, due to total memory used for all operation should be much smaller. Additionally such query spanning all of the blocks and series are unlikely and should be blocked by simple `sample` limit.
--- a/docs/proposals/201912_thanos_binary_index_header.md
+++ b/docs/proposals/201912_thanos_binary_index_header.md
@@ -1,0 +1,191 @@
+---
+title: Binary Format for Index-cache; Renaming to Index-header.
+type: proposal
+menu: proposals
+status: proposed
+owner: bwplotka
+---
+
+### Related Tickets
+
+* https://github.com/thanos-io/thanos/issues/942 (format changes)
+* https://github.com/thanos-io/thanos/issues/1711 (pulling index in smarter way)
+* https://github.com/thanos-io/thanos/pull/1013 (initial investigation of different formats)
+* https://github.com/thanos-io/thanos/issues/448 (OOM)
+* https://github.com/thanos-io/thanos/issues/1705 (LTS umbrella issue)
+
+## Summary
+
+This short document describes the motivation and the design of a new format that is meant to replace `index-cache.json` we have currently.
+
+We also propose renaming index-cache to *index-header* due to name collision with index cache for postings and series.
+
+## Motivation
+
+Currently the Store Gateway component has to be aware of all the blocks (modulo sharding & time partitioning) in the bucket.
+For each block that we want to serve metrics from, the Store Gateway has to have that block `synced` in order to:
+
+* Know what postings are matching given query matchers.
+* Know what label values are for names.
+* Know how to un-intern symbols.
+
+The `sync` process includes:
+
+* Download meta.json or fetch from disk if cached.
+* Check if index-cache.json is present on the disk. If not: 
+  * Check if index-cache.json is present in the bucket. If not: 
+    * Download the whole TSDB index file, mmap all of it.
+    * Build index-cache.json
+    else:
+    * Download index-cache.json from bucket
+* Load whole index-cache.json to memory and keep it for lifetime of the block.
+
+The current, mentioned [index-cache.json][index-cache-code] holds block’s:
+* [TOC][toc]
+* [Symbols][symbols]
+* LabelValues:
+  * Calculated from all [LabelsIndicesTable][labels-indices-table], that each points to [LabelIndex][label-index] entry.
+  * Used for matching against request selectors. 
+* [Postings Offsets][postings-offsets]
+
+There are few problems with this approach: 
+
+1. Downloading the whole TSDB index file (which can be from few MB up to 64GB or even more when limit will be increased soon) is expensive and might be slow.
+2. Building such index-cache requires at least 2x `index-cache.json` size of memory.
+3. Parsing large index-cache.json files (which can be from few KB to few GB) might be slow.
+4. We keep the full content of index-cache file in memory for all blocks.
+   * Some blocks are never queried. 
+   * Some posting offsets and label values within each block are never used.
+
+1, 2 & 3 contributes to Store Gateway startup being slow and resource consuming: https://github.com/thanos-io/thanos/issues/448
+
+4. causes unnecessary cache for unused blocks.
+
+This design is trying to address those 4 problems. 
+
+## Goals
+    
+* Reduce confusion between index-cache.json and IndexCache for series and postings.
+* Do not keep large pieces of the blocks (e.g symbols, label values) in the memory for all blocks. Be able to load it quickly in query time from disk.
+* Use constant amount of memory for building index-header/index-cache; download only required pieces of index from bucket and compute new TOC file.
+* Leverage mmap and let OS to unload unused pages from the physical memory as needed to satisfy the queries against the block.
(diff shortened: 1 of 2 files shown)


docs/proposals/201912_thanos_binary_index_header.md
```diff
@@ -0,0 +1,191 @@
...
+4. causes unnecessary cache for unused blocks.
+
+This design is trying to address those 4 problems. 
+
+## Goals
+    
+* Reduce confusion between index-cache.json and IndexCache for series and postings.
+* Do not keep large pieces of the blocks (e.g symbols, label values) in the memory for all blocks. Be able to load it quickly in query time from disk.
+* Use constant amount of memory for building index-header/index-cache; download only required pieces of index from bucket and compute new TOC file.
+* Leverage mmap and let OS to unload unused pages from the physical memory as needed to satisfy the queries against the block.
```
metalmatze:
```suggestion
* Leverage mmap and let OS unload unused pages from the physical memory as needed to satisfy the queries against the block.
```

docs/proposals/201912_thanos_binary_index_header.md
```diff
@@ -0,0 +1,191 @@
...
+ This is due to random access against label values and postings offsets.
+
+## No Goals
+
+* Full, no initial startup as designed in [Cortex, no initial block sync](https://github.com/thanos-io/thanos/issues/1813)
```
metalmatze: I don't really get this sentence. Could we rephrase it?

docs/proposals/201912_thanos_binary_index_header.md
```diff
@@ -0,0 +1,191 @@
...
+  * However this proposal might a step towards that as we might be able to load index-cache/index quickly on demand from disk.
+  * In the same time be able to load `index-header` in query time directly from bucket is not a goal of this proposal.
```
metalmatze:
```suggestion
  * At the same time be able to load `index-header` in query time directly from bucket is not a goal of this proposal.
```

docs/proposals/201912_thanos_binary_index_header.md
```diff
@@ -0,0 +1,191 @@
...
+* blocks older than X will be loaded on demand on query time and kept cached until evicted.
+* Background eviction process unloads blocks without pending reads to the amount of Y (e.g 100 blocks) in LRU fashion. 
+ 
+Both X and Y being configurable.
+ 
+## Risks
+
+### Unexpected memory usage
+
+Users cares the most about surprising spikes in memory usage. Currently Store Gateway caches the whole index-cache.json. While it's silly to do so for all blocks, this
```
metalmatze:
```suggestion
Users care the most about surprising spikes in memory usage. Currently, the Store Gateway caches the whole index-cache.json. While it's silly to do so for all blocks, this 
```

brancz:
Love the gist of this. I haven't had the chance to do a very thorough review, but I think this is only natural to do. Nice job!

docs/proposals/201912_thanos_binary_index_header.md
```diff
@@ -0,0 +1,191 @@
...
+
+Step 1: Replace `index-cache.json` with `index-header`: 
+  * Implement building `index-header` file from pieces of TSDB index.
+  * Allow Store Gateway to build it instead of `index-cache.json` to disk.
+  * Allow Store Gateway to use `index-header` instead of `index-cache.json` for queries.
+  * Load those files using mmap
+  * Remove building `index-cache.json` in compactor.
+Step 2: Load on demand in query time:
+  * Allow Store Gateway to load `index-header` on demand from disk in query time.
+  * Add background job to unmmap/unload `index-header` for blocks that are unused (LRU).
```
GiedriusS: Not just unused but the least recently used blocks if the memory limit is violated, right?

docs/proposals/201912_thanos_binary_index_header.md
```diff
@@ -0,0 +1,191 @@
...
+The process for building this will be as follows:
+
+* Thanks to https://github.com/thanos-io/thanos/pull/1792 we can check final size of index and scan for TOC file.
+* With TOC:
+  * Get the symbols table and copy it to the local file.
+  * Get the posting table and copy it to the local file.
+  * Get the label offset table and copy it to the local file.
+  * Get the label index and copy it to the local file.
+  
+With that effort building time and resources should be compared with downloading the prebuild `index-header` from bucket. This allows us to reduce complexity of the system as
```
GiedriusS:
```suggestion
With that effort building time and resources should be compared with downloading the prebuilt `index-header` from the bucket. This allows us to reduce complexity of the system as 
```

docs/proposals/201912_thanos_binary_index_header.md
```diff
@@ -0,0 +1,191 @@
...
+
+How to micro-benchmark such change? mmaping is outside of Go run time, which counts allocations etc.
+ 
+### Do we need to mmap?
+
+mmap adds a lot of complexity and confusion especially around monitoring it's memory usage as it does not appear on Go profiles.
+
+While mmap is great for random access against a big file, in fact the current implementation of the [FileReader](https://github.com/prometheus/prometheus/blob/de0a772b8e7d27dc744810a1a693d97be027049a/tsdb/index/index.go#L664)
+reallocates symbols, offsets and label name=value pairs while reading. This really defies the purpose of mmap as we want to combine all info in a dense few sequential sections of index-header binary format, 
+this file will be mostly read sequentially. Still, label values can be accessed randomly, that's why we propose to start with mmap straight away.
```
GiedriusS: But with this, I guess, we would be tied to platforms which support mmap(2). That would exclude Windows AFAIK. It's OK with me, just something to be aware of.

docs/proposals/201912_thanos_binary_index_header.md
```diff
@@ -0,0 +1,191 @@
...
+* https://github.com/thanos-io/thanos/issues/1711 (pulling index in smarter way)
+* https://github.com/thanos-io/thanos/pull/1013 (initial investigation of different formats)
+* https://github.com/thanos-io/thanos/issues/448 (OOM)
+* https://github.com/thanos-io/thanos/issues/1705 (LTS umbrella issue)
+
+## Summary
+
+This short document describes the motivation and the design of a new format that is meant to replace `index-cache.json` we have currently.
+
+We also propose renaming index-cache to *index-header* due to name collision with index cache for postings and series.
```
GiedriusS: 👍

docs/proposals/201912_thanos_binary_index_header.md
```diff
@@ -0,0 +1,191 @@
...
+## Verification
+
+* Compare memory and latency usage for startup without index-header being generated on compactor
+* Benchmark for loading the `index-header` vs `index-cache.json` into Store GW.
+
+## Proposal
+
+TSDB index is in binary [format](https://github.com/prometheus/prometheus/blob/master/tsdb/docs/format/index.md). 
+
+To allow reduced resource consumption and effort when building (1), (2), "index-header" for blocks we plan to reuse similar format for sections like symbols, label indices and
```
GiedriusS: Not sure if I have missed this detail but what is the actual format that this proposes? Protobufs?

docs/proposals/201912_thanos_binary_index_header.md
brancz: In prometheus this seems to work: https://github.com/prometheus/prometheus/blob/master/tsdb/fileutil/mmap_windows.go

GiedriusS: TIL about that, that's nice! We should definitely re-use that :p disregard this message, then.

bwplotka: All is good on this front.

docs/proposals/201912_thanos_binary_index_header.md
bwplotka: How we would know this?

docs/proposals/201912_thanos_binary_index_header.md
bwplotka: A custom binary format was planned, composed with those sections as mentioned, however, I might find a blocker to this ):

docs/proposals/201912_thanos_binary_index_header.md
```diff
@@ -0,0 +1,215 @@
...
+
+* blocks newer then X (e.g 2w) being always loaded
+* blocks older than X will be loaded on demand on query time and kept cached until evicted.
+* background eviction process unloads blocks without pending reads to the amount of Y (e.g 100 blocks) in LRU fashion. 
+ 
+Both X and Y being configurable.
+ 
+## Risks
+
+### Posting size to fetch
```
bwplotka: Added this which we missed initially in this design. Let me know if it's clear. 

TL;DR: we scan `postings` form TSDB index when building `index-cache` and without changing TSDB index this will be still required building `index-header`

@brancz @GiedriusS

docs/proposals/201912_thanos_binary_index_header.md
```diff
@@ -0,0 +1,215 @@
...
+This means maintaining pool with limited number of `index-header` files loaded at the time in Store Gateway.
+With an LRU logic we should be able to deduce which of blocks should be unloaded and left on the disk. 
+
+This proposes following algorithm:
+
+* blocks newer then X (e.g 2w) being always loaded
+* blocks older than X will be loaded on demand on query time and kept cached until evicted.
+* background eviction process unloads blocks without pending reads to the amount of Y (e.g 100 blocks) in LRU fashion. 
+ 
+Both X and Y being configurable.
```
pracucci: I understand the proposed algorithm would be easier to implement, but the only real downside of keeping all `index-header` loaded is the memory limit. From the UX perspective would be great being able to set a memory limit for the `index-header` and having Thanos automatically adjusting X and Y based on actual indexes size.

docs/proposals/201912_thanos_binary_index_header.md
```diff
@@ -0,0 +1,215 @@
...
+makes this proposal less valuable as we still need to download big part of index and traverse through it instead of what we propose in [#Proposal](./201912_thanos_binary_index_header.md#proposal)
+
+For series we don't know the exact size either, however we estimate max size of each series to be 64*1024. It depends on sane number of label pairs and chunks per series.
+We have really only one potential case when this was too low: https://github.com/thanos-io/thanos/issues/552. Decision about series this was made here: https://github.com/thanos-io/thanos/issues/146
+
+For postings it's more tricky as it depends on number of series in which given label pair exists. For worst case it can be even million label-pair for popular pairs like `__name__=http_request_http_duration_bucket` etc.
+
+We have few options:
+
+* Encode this posting size in TSDB index `PostingOffset`.
```
pracucci: I'm not sure this is a real option. Even if such change will be accepted and merged (and I don't think), you would have to deal with all previously created blocks, which invalidates this option.

docs/proposals/201912_thanos_binary_index_header.md
```diff
@@ -0,0 +1,215 @@
...
+
+For series we don't know the exact size either, however we estimate max size of each series to be 64*1024. It depends on sane number of label pairs and chunks per series.
+We have really only one potential case when this was too low: https://github.com/thanos-io/thanos/issues/552. Decision about series this was made here: https://github.com/thanos-io/thanos/issues/146
+
+For postings it's more tricky as it depends on number of series in which given label pair exists. For worst case it can be even million label-pair for popular pairs like `__name__=http_request_http_duration_bucket` etc.
+
+We have few options:
+
+* Encode this posting size in TSDB index `PostingOffset`. 
+* Scan postings to fetch, which is something we wanted to avoid when building `index-header` without downloading full TSDB index.
```
pracucci: I genuinely believe this option just invalidates the proposal. We shouldn't download the postings section at all while building the `index-header`.

docs/proposals/201912_thanos_binary_index_header.md
```diff
@@ -0,0 +1,215 @@
...
+* Estimate some large value (will overfetch)
+* Estimate casual value and retry fetching remaining size on demand for data consistency.
```
pracucci: If I'm not missing anything, the postings `End` is only used in `bucketIndexReader.fetchPostings()`, affecting only the postings which have not been fetched from the cache.

Let's consider to use a casual value. The partitioner will partition into 1+ parts, which will get fetched. Upon fetching and processing all of them, we're able to calculate the end of each posting and - if there's anyone for which some partial data is missing - we can fetch all the missing data in a second iteration, so the overhead introduced in the worst case scenario shouldn't be greater than 2x compared to today.

Given we're already over fetching due to how the max gap in the partitioner works, and given when accessing the bucket store the request latency impacts more than the bandwidth (ie. if you benchmark reading 1KB and 1MB, reading 1MB is not 1024x slower than reading 1KB) and given the work I'm doing to add memcached support for the index cache, this option doesn't look particularly bad to me, given the benefits this proposal bring.

docs/proposals/201912_thanos_binary_index_header.md
```diff
@@ -0,0 +1,215 @@
...
+
+mmap adds a lot of complexity and confusion especially around monitoring it's memory usage as it does not appear on Go profiles.
+
+While mmap is great for random access against a big file, in fact the current implementation of the [FileReader](https://github.com/prometheus/prometheus/blob/de0a772b8e7d27dc744810a1a693d97be027049a/tsdb/index/index.go#L664)
+reallocates symbols, offsets and label name=value pairs while reading. This really defies the purpose of mmap as we want to combine all info in a dense few sequential sections of index-header binary format, 
+this file will be mostly read sequentially. Still, label values can be accessed randomly, that's why we propose to start with mmap straight away. 
+
+### Is initial long startup, really a problem?
+
+After initial startup with persistent disk, next startup should be quick due to cached files on disk for old blocks.
```
pracucci: > After initial startup with persistent disk, next startup should be quick due to cached files on disk for old blocks.

Another use case is scaling. When you scale up your cluster because of high load, you want the store to be ready as quickly as possible. The initial long startup delays it.

docs/proposals/201912_thanos_binary_index_header.md
```diff
@@ -0,0 +1,215 @@
...
+1, 2 & 3 contributes to Store Gateway startup being slow and resource consuming: https://github.com/thanos-io/thanos/issues/448
+
+4. causes unnecessary cache for unused blocks.
+
+This design is trying to address those 4 problems. 
+
+## Goals
+    
+* Reduce confusion between index-cache.json and IndexCache for series and postings.
+* Do not keep large pieces of the blocks (e.g symbols, label values) in the memory for all blocks. Be able to load it quickly in query time from disk.
```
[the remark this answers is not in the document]
bwplotka: will do thanks

docs/proposals/201912_thanos_binary_index_header.md
bwplotka: No really. We need to pull PostingOffset anyway. (:

docs/proposals/201912_thanos_binary_index_header.md
bwplotka: Yes fully agree

docs/proposals/201912_thanos_binary_index_header.md
pracucci: I thought the option was to change the TSDB index to add the size/end to the postings offset table, but I guess I'm wrong. Can you elaborate on this option, please?

docs/proposals/201912_thanos_binary_index_header.md
bwplotka: Thanks for this - I have similar thoughts. This is still doable indeed. I think  I agree we should take is a preffered option so far.

docs/proposals/201912_thanos_binary_index_header.md
bwplotka: yup, adding

docs/proposals/201912_thanos_binary_index_header.md
GiedriusS: Sorry, I meant some kind of limit in terms of their size on disk because it would be really nice to avoid RAM usage blowing up in case of some queries. Having all of it in RAM is horrible, yes, but OTOH it gives the users a relatively certain thing: we know that at least that part is more or less equal over a period of time.

docs/proposals/201912_thanos_binary_index_header.md
```diff
@@ -0,0 +1,215 @@
...
+* blocks older than X will be loaded on demand on query time and kept cached until evicted.
+* background eviction process unloads blocks without pending reads to the amount of Y (e.g 100 blocks) in LRU fashion. 
+ 
+Both X and Y being configurable.
+ 
+## Risks
+
+### Posting size to fetch
+
+While idea of combing different pieces of TSDB index as our index-header is great, unfortunately we heavily rely
```
pracucci: Can't we infer the `End` looking at the `Start` of the next posting, sorting postings by `Start`?

For each posting we have label name, label value and offset to its series list in the postings section. The postings section contains all postings, one after the other one. If we sort the postings table by offset, the end range for the posting at position X should be equal to the start range of the posting at position X+1. The last posting will end at the position of the label index table.

Am I missing anything?

docs/proposals/201912_thanos_binary_index_header.md
bwplotka: We got some answers (:

bwplotka:
Starting implementation (: 

cc @GiedriusS @pracucci if you want to take another look.

docs/proposals/201912_thanos_binary_index_header.md
bwplotka: As discussed offline added to the doc (:

docs/proposals/201912_thanos_binary_index_header.md
bwplotka: Added all of options including those not choosen

docs/proposals/201912_thanos_binary_index_header.md
```diff
@@ -0,0 +1,217 @@
...
+
+Thanos will build/compose all index-headers on the startup for now, however in theory we can load and build those blocks on demand.
+Given the minimal memory that each loaded block should take now, this is described as [Future Work](./201912_thanos_binary_index_header.md#Future-Work)
+
+### Length of Posting to fetch
+
+While idea of combing different pieces of TSDB index as our index-header is great, unfortunately we heavily rely 
+on information about size of each posting represented as `postingRange.End`. 
+
+We need to know that apriori to know how to partition and how many bytes we need to fetch from the storage to get each posting: https://github.com/thanos-io/thanos/blob/7e11afe64af0c096743a3de8a594616abf52be45/pkg/store/bucket.go#L1567
```
pracucci: > We need to know that apriori to know how to partition

I guess you mean `We need to know apriori how to partition`

bwplotka: Will fix in next PR (:

PR #1841, opened 2019-12-05
add kubernetes comment for service discovery

daixiang0:

Add an example for service discovery in kubernetes.


bwplotka:
Thanks!

Issue #1177, opened 2019-05-27
docs: Add check to CI to make sure links in markdown resolve correctly


LiliC:
Currently, as part of CI the only tests are for the website links to resolve correctly, there should also be a check for links in all markdown files as well. See full discussion https://github.com/improbable-eng/thanos/pull/1176#issuecomment-496219345.

I know of https://github.com/crawford/marker and it works well for us in https://github.com/operator-framework/operator-sdk/ but I am sure there are other ones out there for this.

bwplotka:
Awesome, we would love to have this. Marked this issue as `help wanted` (: 

Thanks @LiliC

Avni-Sharma:
Hi, Can I work on this?

GiedriusS:
> Hi, Can I work on this?

Sure, feel free to ask for help if you have any questions in our Slack. PRs welcome!

bwplotka:
Just to clarify. We already have [`liche`](https://github.com/raviqqe/liche) in `make dock-check`. However, looks like this does not detect all the broken links like mentioned here: https://github.com/thanos-io/thanos/pull/1176/files#diff-264a4415d8565eda4a7572e6522acc07R69 (`/pkg/store/storepb/rpc.pb.go)` and things like markdown sections: `query.md#custom-response-fields)` 

AC:

* We check both broken external links as well as broken markdown sections and local files

PR #1845, opened 2019-12-06
Check markdown links

Avni-Sharma:
Fix for https://github.com/thanos-io/thanos/issues/1177

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -163,6 +163,11 @@ check-docs: $(EMBEDMD) $(LICHE) build
 	@$(LICHE) --recursive docs --exclude "(cloud.tencent.com|alibabacloud.com)" --document-root .
 	@$(LICHE) --exclude "(cloud.tencent.com|goreportcard.com|alibabacloud.com)" --document-root . *.md
 
+# check-markdown-links checks if markdown files have broken links.
+.PHONY: check-markdown-links
+check-markdown-links:
+	./scripts/marker
+
 # checks Go code comments if they have trailing period (excludes protobuffers and vendor files).
 # Comments with more than 3 spaces at beginning are omitted from the check, example: '//    - foo'.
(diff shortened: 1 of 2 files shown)


Makefile
```diff
@@ -163,6 +163,11 @@ check-docs: $(EMBEDMD) $(LICHE) build
 	@$(LICHE) --recursive docs --exclude "(cloud.tencent.com|alibabacloud.com)" --document-root .
 	@$(LICHE) --exclude "(cloud.tencent.com|goreportcard.com|alibabacloud.com)" --document-root . *.md
 
+# check-markdown-links checks if markdown files have broken links.
+.PHONY: check-markdown-links
+check-markdown-links:
+	./scripts/marker
```
brancz: I'm not a fan of checking in binaries like that. I think it's fair to ask people to install it themselves.

squat: I understand the desire to make this easy to use, but unfortunately it’s not portable. What if someone isn’t running Linux on AMD64?

bwplotka:
Thanks!

I added more context to https://github.com/thanos-io/thanos/issues/1177 Essentially we already have `liche` in `make doc-check` doing this, but indeed it looks it's not fully capable. 

Suggestions to this PR:

1. It would be nice to test on some broken links the capabilities of liche vs marker you want to introduce. Then we should either replace [liche](https://github.com/raviqqe/liche) or add `marker` as extra tool. Alternatively, we can contribute put issue on liche on what exactly is not working, I am surprised it does not handle cases like ours.
2. Definitely, we have to have either script which installs this (unfortunately it's not Go so it's bit more tricky) or add comment how to install it.
3. CI has to have this bundled as well. Preferable installing linux x64 binary on start or baking this in process of `make docker-ci`

PR #1848, opened 2019-12-07
returns error messages when trigger reload with http

00arthur00:
Signed-off-by: arthur yang <yang_yapo@126.com>
keep same logic with prometheus


cmd/thanos/rule.go
```diff
@@ -476,8 +479,9 @@ func runRule(
 				signal.Notify(c, syscall.SIGHUP)
 				select {
 				case <-c:
+					rc := make(chan error, 1)
 					select {
-					case reload <- struct{}{}:
+					case reload <- rc:
```
bwplotka: Let's log error for this case?

cmd/thanos/rule.go
```diff
@@ -420,17 +420,18 @@ func runRule(
 	}
 
 	// Handle reload and termination interrupts.
-	reload := make(chan struct{}, 1)
+	reload := make(chan chan error, 1)
```
bwplotka: Can we add comment? e.g
```
         // Reload is channel passing another error channel for deferred error feedback if reload succeded or not.
```

cmd/thanos/rule.go
00arthur00: @bwplotka  I've committed my change, please help to review. Thanks.

00arthur00:
@bwplotka Any update?

daixiang0:
@bwplotka ping

00arthur00:
@bwplotka ping

kakkoyun:
Hey @00arthur00, we've recently updated `SIGHUP`handling, see #2139. I think you need to rebase and update your changes.

Please feel free to ping me, I can review this one for you.

00arthur00:
@kakkoyun  Sure. Thanks.

00arthur00:
@kakkoyun Please help review. Thanks.

cmd/thanos/rule.go
```diff
@@ -472,49 +472,26 @@ func runRule(
...
 	{
 		cancel := make(chan struct{})
-		reload <- struct{}{} // Initial reload.
-
 		g.Add(func() error {
+			// Initialize rules.
+			if err := reloadRules(logger, ruleFiles, ruleMgr, evalInterval, configSuccess, configSuccessTime, rulesLoaded); err != nil {
+				level.Error(logger).Log("msg", "initialize rules failed", "err", err)
+				// Return when initialize with invalid pattern error.
+				if _, ok := err.(*errInvalidPattern); ok {
```
kakkoyun: Consider using recently introduced errors API. https://blog.golang.org/go1.13-errors
```go
// Similar to:
//   if e, ok := err.(*QueryError); ok { … }
var e *QueryError
if errors.As(err, &e) {
    // err is a *QueryError, and e is set to the error's value
}
```

cmd/thanos/rule.go
```diff
@@ -758,3 +738,47 @@ func addDiscoveryGroups(g *run.Group, c *http_util.Client, interval time.Duratio
...
 }
+
+type errInvalidPattern struct {
+	err error
+	pat string
+}
+
+func (e *errInvalidPattern) Error() string {
+	return errors.Wrapf(e.err, "retrieving rule files failed. Ignoring file. pattern %s", e.pat).Error()
+}
```
kakkoyun: Missing white space after function body.

cmd/thanos/rule.go
```diff
@@ -758,3 +738,47 @@ func addDiscoveryGroups(g *run.Group, c *http_util.Client, interval time.Duratio
...
+func reloadRules(logger log.Logger,
```
kakkoyun: This function has too many parameters, it makes it harder to read. And most of the parameters are metrics, consider using a struct to collect metrics as it had done in https://github.com/thanos-io/thanos/blob/021f623e7f3b29121f425ec2ce10e4e96377963a/cmd/thanos/downsample.go#L52-L55

and pass it around.

cmd/thanos/rule.go
```diff
@@ -472,49 +472,26 @@ func runRule(
...
 		cancel := make(chan struct{})
-		reload <- struct{}{} // Initial reload.
-
 		g.Add(func() error {
+			// Initialize rules.
+			if err := reloadRules(logger, ruleFiles, ruleMgr, evalInterval, configSuccess, configSuccessTime, rulesLoaded); err != nil {
+				level.Error(logger).Log("msg", "initialize rules failed", "err", err)
+				// Return when initialize with invalid pattern error.
+				if _, ok := err.(*errInvalidPattern); ok {
+					return err
```
kakkoyun: This changes behaviour compared to the previous implementation. Previously, it was logging and proceeding to other rule files.
```go
if err != nil {	
	// The only error can be a bad pattern.	
	level.Error(logger).Log("msg", "retrieving rule files failed. Ignoring file.", "pattern", pat, "err", err)	
	continue	
}
```

If this is what we want, you should update the CHANGELOG accordingly.
You should update CHANGELOG anyway because you are changing the behaviour of the HTTP API.

@bwplotka

00arthur00: @kakkoyun  This comes from orignal logic. I wrote it to keep same logic with orignal one.

kakkoyun: No actually. That's what I'm trying to tell, it doesn't keep the same logic _if I'm not mistaken_. If you check your `reloadRules` function, in a for-loop when execution encounters an error, it returns it so it breaks out from with the first encounter of an error.
```go
var files []string
for _, pat := range ruleFiles {
    fs, err := filepath.Glob(pat)
    if err != nil {
        // Check errInvalidPattern when initialize.
        return &errInvalidPattern{err, pat}
    }

    files = append(files, fs...)
}
```

The old code, just prints an error and continue with the execution and checks other files.

00arthur00: @kakkoyun Sorry for my misunderstanding. Cannot remeber why change like this, since it's long time.
Maybe the change is reasonable.  What do you thank?

cmd/thanos/rule.go
00arthur00: @kakkoyun  Added a newline. If you think OK, please help to resovle this conversation.

cmd/thanos/rule.go
00arthur00: Sure. Will look into it.

cmd/thanos/rule.go
00arthur00: @kakkoyun Hi, the code comes from prometheus. Please have a look at the link 
https://github.com/prometheus/prometheus/blob/fdbb18b445285b69d005d88d2c1edb83f5c1b5e5/cmd/prometheus/main.go#L459.

kakkoyun: The change is definitely reasonable, all I'm saying is that we should document the behaviour change. And of course, if also the maintainers agree.

cmd/thanos/rule.go
kakkoyun: All good, but I also don't see a button to resolve conversation.

00arthur00: @kakkoyun Under the convsersation, there's a button. Maybe only the submitter can see it. Will click the button.

cmd/thanos/rule.go
00arthur00: @kakkoyun  Committed. Please help to review. Thanks.

cmd/thanos/rule.go
bwplotka: From my side, I think @kakkoyun is right.

> You should update CHANGELOG anyway because you are changing the behaviour of the HTTP API.

Yes

> The old code, just prints an error and continue with the execution and checks other files.

My opinion is that we should fail and proceed to another rule. No point in failing everything and stop as it does not crash (and should not!) ruler. Also we should be consistent to Prometheus behaviour. (: Can we change the logic here to what was previously? @00arthur00

00arthur00: @bwplotka Thanks for your reply. I will revert the logic. I.e, keep log and no termination.

cmd/thanos/rule.go
00arthur00: @kakkoyun According to the coverstaion, I should not change the behavior the the rule reloading. This code will be removed.

cmd/thanos/rule.go
```diff
@@ -200,6 +201,55 @@ func registerRule(m map[string]setupFunc, app *kingpin.Application) {
...
+		}, []string{"strategy"},
+	)
+	m.ruleEvalWarnings.WithLabelValues(strings.ToLower(storepb.PartialResponseStrategy_ABORT.String()))
+	m.ruleEvalWarnings.WithLabelValues(strings.ToLower(storepb.PartialResponseStrategy_WARN.String()))
+
+	reg.MustRegister(m.configSuccess)
+	reg.MustRegister(m.configSuccessTime)
+	reg.MustRegister(m.duplicatedQuery)
+	reg.MustRegister(m.rulesLoaded)
+	reg.MustRegister(m.ruleEvalWarnings)
```
kakkoyun: Better to check if the registry is nil or not.
```suggestion
if (reg != nil) {
	reg.MustRegister(m.configSuccess)
	reg.MustRegister(m.configSuccessTime)
	reg.MustRegister(m.duplicatedQuery)
	reg.MustRegister(m.rulesLoaded)
	reg.MustRegister(m.ruleEvalWarnings)
}
```

cmd/thanos/rule.go
```diff
@@ -472,49 +490,22 @@ func runRule(
...
 	// Handle reload and termination interrupts.
-	reload := make(chan struct{}, 1)
 	{
 		cancel := make(chan struct{})
-		reload <- struct{}{} // Initial reload.
-
 		g.Add(func() error {
+			// Initialize rules.
+			if err := reloadRules(logger, ruleFiles, ruleMgr, evalInterval, metrics); err != nil {
+				level.Error(logger).Log("msg", "initialize rules failed", "err", err)
```
kakkoyun: For the initial load, we can actually fail and return the combined multierror.

cmd/thanos/rule.go
kakkoyun: @bwplotka I think we should error and exit out if rule file includes faulty rules **when we are starting component for the first time**. And then for any subsequent update, it should just log the error and continue running without any crash. What do you think?

cmd/thanos/rule.go
kakkoyun: I don't have strong opinions about it, both works for me. In any case, I've pinged @bwplotka on the thread for the final decision.

cmd/thanos/rule.go
```diff
@@ -200,6 +201,55 @@ func registerRule(m map[string]setupFunc, app *kingpin.Application) {
 	}
 }
 
+// RuleMetrics defines thanos rule metrics.
+type RuleMetrics struct {
```
bwplotka: :+1: 
Thanks!

cmd/thanos/rule.go
```diff
@@ -200,6 +201,55 @@ func registerRule(m map[string]setupFunc, app *kingpin.Application) {
...
+		Name: "thanos_rule_config_last_reload_successful",
+		Help: "Whether the last configuration reload attempt was successful.",
+	})
+	m.configSuccessTime = prometheus.NewGauge(prometheus.GaugeOpts{
+		Name: "thanos_rule_config_last_reload_success_timestamp_seconds",
+		Help: "Timestamp of the last successful configuration reload.",
+	})
+	m.duplicatedQuery = prometheus.NewCounter(prometheus.CounterOpts{
+		Name: "thanos_rule_duplicated_query_addresses_total",
+		Help: "The number of times a duplicated query addresses is detected from the different configs in rule",
```
bwplotka: missing trailing period ;p

cmd/thanos/rule.go
```diff
@@ -200,6 +201,55 @@ func registerRule(m map[string]setupFunc, app *kingpin.Application) {
...
+	})
+	m.rulesLoaded = prometheus.NewGaugeVec(
+		prometheus.GaugeOpts{
+			Name: "thanos_rule_loaded_rules",
+			Help: "Loaded rules partitioned by file and group",
```
bwplotka: ditto

cmd/thanos/rule.go
```diff
@@ -200,6 +201,55 @@ func registerRule(m map[string]setupFunc, app *kingpin.Application) {
...
+	configSuccessTime prometheus.Gauge
+	duplicatedQuery   prometheus.Counter
+	rulesLoaded       *prometheus.GaugeVec
+	ruleEvalWarnings  *prometheus.CounterVec
+}
+
+func newRuleMetrics(reg *prometheus.Registry) *RuleMetrics {
+	m := new(RuleMetrics)
+
+	m.configSuccess = prometheus.NewGauge(prometheus.GaugeOpts{
```
bwplotka: Do you want to be our super hero (level: legendary)? Let's start using new pattern here as decided here: https://github.com/thanos-io/thanos/issues/2102

so `promauto.With(reg).NewGauge... and all here :hugs:

cmd/thanos/rule.go
```diff
@@ -200,6 +201,55 @@ func registerRule(m map[string]setupFunc, app *kingpin.Application) {
...
+	)
+	m.ruleEvalWarnings = prometheus.NewCounterVec(
+		prometheus.CounterOpts{
+			Name: "thanos_rule_evaluation_with_warnings_total",
+			Help: "The total number of rule evaluation that were successful but had warnings which can indicate partial error.",
+		}, []string{"strategy"},
+	)
+	m.ruleEvalWarnings.WithLabelValues(strings.ToLower(storepb.PartialResponseStrategy_ABORT.String()))
+	m.ruleEvalWarnings.WithLabelValues(strings.ToLower(storepb.PartialResponseStrategy_WARN.String()))
+	if reg != nil {
```
bwplotka: ... then with `promauto` we don't need this (:

cmd/thanos/rule.go
bwplotka: Yes, initial can fail indeed (:

cmd/thanos/rule.go
```diff
@@ -472,49 +490,22 @@ func runRule(
...
-				level.Debug(logger).Log("msg", "configured rule files", "files", strings.Join(ruleFiles, ","))
-				var files []string
-				for _, pat := range ruleFiles {
-					fs, err := filepath.Glob(pat)
-					if err != nil {
-						// The only error can be a bad pattern.
-						level.Error(logger).Log("msg", "retrieving rule files failed. Ignoring file.", "pattern", pat, "err", err)
-						continue
+					if err := reloadRules(logger, ruleFiles, ruleMgr, evalInterval, metrics); err != nil {
+						level.Error(logger).Log("msg", "reload rules by sighup failed", "err", err)
```
bwplotka: Not only sighup TBH, can be HTTP `/-/reload` right? So let's fix message here (:

cmd/thanos/rule.go
```diff
@@ -472,49 +490,22 @@ func runRule(
...
-				}
-
-				configSuccess.Set(1)
-				configSuccessTime.SetToCurrentTime()
-
-				rulesLoaded.Reset()
-				for _, group := range ruleMgr.RuleGroups() {
-					rulesLoaded.WithLabelValues(group.PartialResponseStrategy.String(), group.File(), group.Name()).Set(float64(len(group.Rules())))
+				case <-cancel:
+					return errors.New("canceled")
```
bwplotka: I know it's not yours, but can we move to `context.WithCancel(..)` as we do everywhere else?

cmd/thanos/rule.go
```diff
@@ -564,7 +555,10 @@ func runRule(
 		}
 
 		router.WithPrefix(webRoutePrefix).Post("/-/reload", func(w http.ResponseWriter, r *http.Request) {
-			reload <- struct{}{}
+			if err := reloadRules(logger, ruleFiles, ruleMgr, evalInterval, metrics); err != nil {
```
bwplotka: Hm... so why we wait on reloadSignal above if we just reload stuff here? :thinking:

cmd/thanos/rule.go
bwplotka: Looks good now :+1:

cmd/thanos/rule.go
```diff
@@ -564,7 +555,10 @@ func runRule(
 		}
 
 		router.WithPrefix(webRoutePrefix).Post("/-/reload", func(w http.ResponseWriter, r *http.Request) {
-			reload <- struct{}{}
```
00arthur00: @bwplotka The line is removed. So we don't wait the reload signal at all.

cmd/thanos/rule.go
00arthur00: @bwplotka 
> reload <- struct{}{}

This line is removed.

cmd/thanos/rule.go
00arthur00: Checked the promauto package. The promauto.With function exists in master branch only. There may be something unstable there. Not sure we should use the master branch.
@bwplotka  What's your suggestion?

cmd/thanos/rule.go
00arthur00: @bwplotka @kakkoyun Let's focus on the reload process first in this PR. Since the failure in iniliazation is breaking change.
After the reload related code is merged, I will start a new PR for the initialization failure only.
Any suggestion?

cmd/thanos/rule.go
00arthur00: @bwplotka Yes. These code hanlde sighup only.

cmd/thanos/rule.go
bwplotka: I maintain master, let's update maybe? (:

00arthur00: @bwplotka Cool. Updated, please help check.

cmd/thanos/rule.go
bwplotka: cool, happy with that.

cmd/thanos/rule.go
bwplotka: Would be nice to treat signal and HTTP reload exactly the same way (as it was before). Why not? We could reuse `reloadSignal` channel.

cmd/thanos/rule.go
bwplotka: See my comment above.

bwplotka: I don't think we need to have separate err message for HTTP and signal (:

cmd/thanos/rule.go
00arthur00: > Would be nice to treat signal and HTTP reload exactly the same way (as it was before). Why not? We could reuse `reloadSignal` channel.

@bwplotka If so, we need another channel to receive the error message for webhandler. So a new struct should wrap reloadSignal and errMsg.  Maybe it is redundant for the current implementation.
BTW, If you persist, I will follow your comment, and any suggestion on the "a new struct"  solution?

cmd/thanos/rule.go
00arthur00: concrete error message helps troubleshooting.

bwplotka: Ok, then let's maybe use another channel?

cmd/thanos/rule.go
00arthur00: Add a new struct to handler webhandler. Since we have a reloadSignal as an input paramter, we need select reloadSignal always.
Please help review. Thanks.
@bwplotka

cmd/thanos/rule.go
```diff
@@ -200,6 +202,55 @@ func registerRule(m map[string]setupFunc, app *kingpin.Application) {
...
+		}, []string{"strategy"},
+	)
+	m.ruleEvalWarnings.WithLabelValues(strings.ToLower(storepb.PartialResponseStrategy_ABORT.String()))
+	m.ruleEvalWarnings.WithLabelValues(strings.ToLower(storepb.PartialResponseStrategy_WARN.String()))
+
+	return m
+}
+
+// reloadErr handlers webhandler and returns err message.
+type reloadErr struct {
```
bwplotka: I would just remove this type and just use `chan error` directly? Why we need this struct?

cmd/thanos/rule.go
bwplotka: LGTM!

PR #1856, opened 2019-12-09
receive: close DBReadOnly after flushing

### LINKED ISSUE #1855: Receive: Possible memory leak with TSDB growth

I have tested with both 0.8.1 and 0.9.0. Most recent version output is as follows:

```
thanos, version 0.9.0 (branch: HEAD, revision: 0833cad83db8b257a2275ec83a3d034c73659056)
  build user:       circleci@8e3da52515a2
  build date:       20191203-17:03:13
  go version:       go1.13.1
```

**What happened**: 
When receiver runs for a long period of time it's memory grows until reaching OOM kills

A steady memory footprint

Run thanos receive with a retention time of 5x the block size (default 2h). When the first WAL flush occurs you will see the memory usage start to grow. With each WAL flush there is exponentially more memory usage. 

image

I initially found this problem when our servers started an OOM loop. I tried to clear the WAL and restart to no avail. The only solution was to delete some local TSDB blocks. With each block I deleted it lowered the memory footprint. I finally restarted fresh and found the memory usage to be level and consistent until the first WAL flush at which point there was a small spike (probably from the flush) and then a slope. 

image

You can see the memory growth here. I have the blocks set to 1h which explains the spikes.

**UPDATE**: it appears that restarting exacerbates the problem. I didn't change the hashring.json or do anything but restart the thanos service. 
image

blockloop:
We are running thanos receive in production with 1w retention. When the receiver starts up it consumes about 20% memory on each of our 32 nodes. After the initial WAL flush the memory footprint progressively grows over time. I originally reported this in CNCF Slack. I initially tried to delete the WAL and restart the nodes but that didn’t solve the issue. The only solution was to clear out some TSDB blocks

Potentially fixes #1855 

## Changes

Close the DBReadOnly connection after `FlushableStorage.Flush()`

## Verification

I’ve not been able to verify this just yet. I’ll try to build the image with this patch and see if I see an improvement.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -10,6 +10,10 @@ NOTE: As semantic versioning states all 0.y.z releases can contain breaking chan
 We use *breaking* word for marking changes that are not backward compatible (relates only to v0.y.z releases.)
 
 ## Unreleased
+### Fixed
+
+- [#1856](https://github.com/thanos-io/thanos/pull/1856) Receive: close DBReadOnly after flushing to fix a memory leak.
+
 ### Added
 - [#1852](https://github.com/thanos-io/thanos/pull/1852) Add support for `AWS_CONTAINER_CREDENTIALS_FULL_URI` by upgrading to minio-go v6.0.44
 
--- a/pkg/receive/tsdb.go
+++ b/pkg/receive/tsdb.go
@@ -92,6 +92,7 @@ func (f *FlushableStorage) Flush() error {
 	if err != nil {
 		return errors.Wrap(err, "opening read-only DB")
 	}
+	defer ro.Close()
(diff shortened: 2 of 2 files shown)


blockloop:
Interesting you mention that. We have a watcher that monitors our nodes and updates the hashring.json file. I noticed some changes it picked up a few times. Is it bad for the hashring to update frequently? I have it do a `list | sort` on the available nodes and write to the hashring.json file.

squat:
nice :)

No, it should absolutely not be a problem to update the file frequently. The thanos receive config watcher checks the hash of the file to check it the content has actually changed; if not, then no update is triggered and no cost is incurred.

On the other hand, if the file is actually changing very frequently, then you will be triggering lots of TSDB flushes, which could be costly.

PR #1862, opened 2019-12-10
objstore/s3, docs/storage.md: Added 'path' field for S3 objstore

DLag:
Added 'path' field for S3 objstore to allow storing data into a sub-folder of the buckets.

Also updated storage.md doc to explain how to use it.


bwplotka:
Hi, I think we need to have a general consensus if we want this in Thanos (:  

We already have this: 

Related: 
* https://github.com/thanos-io/thanos/pull/1392
* https://github.com/thanos-io/thanos/issues/1318

I would suggest dicsussing on https://github.com/thanos-io/thanos/issues/1318 with your use cases and plan to implement (:

PR #1867, opened 2019-12-11
add user-agent in http header.

johncming:
add user-agent in http header.


pkg/alert/alert.go
```diff
@@ -17,8 +17,9 @@ import (
 	"github.com/go-kit/kit/log/level"
 	"github.com/pkg/errors"
 	"github.com/prometheus/client_golang/prometheus"
-
+	"github.com/prometheus/common/version"
```
squat: This sends the Prometheus version, not the Thanos version

bwplotka:
Would be nice to have CHANGELOG for this (:

PR #1868, opened 2019-12-11
add lock.

johncming:


johncming:
> Oops, sorry my bad - double-checked: this variable isn't actually being modified at all: see by doing `git grep "capacity ="` so this is perfectly OK to do. Unless I have missed something, this PR is invalid.

Yes, there are no write operations after creation.

PR #1870, opened 2019-12-11
persist settings in query UI

### LINKED ISSUE #1743: query: UI doesn't persist Thanos-specific settings
When using the Thanos query UI, the additional settings like "deduplication", "partial response", etc are not persisted to the URL the same way settings like "stacked" are. This makes it difficult to share URLs.

Also problematic is that the UI defaults to "Only raw data", which can be expensive on large queries.

yeya24:

## Changes

Fixes #1743

### DIFF
--- a/pkg/ui/templates/query_menu.html
+++ b/pkg/ui/templates/query_menu.html
@@ -1,25 +1,25 @@
 {{define "nav"}}
     <nav class="navbar fixed-top navbar-expand-sm navbar-dark bg-dark">
-      <div class="container-fluid">
-        <button type="button" class="navbar-toggler" data-toggle="collapse" data-target="#nav-content" aria-expanded="false" aria-controls="nav-content" aria-label="Toggle navigation">
-          <span class="navbar-toggler-icon"></span>
-        </button>
-        <a class="navbar-brand" href="{{ pathPrefix }}/">Thanos</a>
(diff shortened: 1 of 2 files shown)


pkg/ui/templates/query_menu.html
```diff
@@ -9,7 +9,7 @@
           <ul class="navbar-nav">
             <li class="nav-item"><a class="nav-link" href="{{ pathPrefix }}/graph">Graph</a></li>
             <li class="nav-item"><a class="nav-link" href="{{ pathPrefix }}/stores">Stores</a></li>
-	    <li class="nav-item dropdown">
+	        <li class="nav-item dropdown">
```
squat: I think this indentation is unneeded

yeya24: Thanks! I just format this page.

yeya24:
Sorry, just forget to make the two options as default, let me modify this PR.

pkg/ui/templates/query_menu.html
squat: thanks (: looks like we have a bit of a mix of 2sp and 4sp in those templates. I'll make a followup to standardize it.

GiedriusS: :+1:

PR #1876, opened 2019-12-12
Clean up white noise and format all codes

daixiang0:

Based on #1801, finish format and lint for current codes and docs.

## Changes

- clean up all white noise calling `make docs`
- fix comment issue calling `make format`:
```


./pkg/objstore/oss/oss.go:
func (b *Bucket) ObjectSize(ctx context.Context, name string) (uint64, error) {
	// https://github.com/aliyun/aliyun-oss-go-sdk/blob/cee409f5b4d75d7ad077cacb7e6f4590a7f2e172/oss/bucket.go#L668
	m, err := b.bucket.GetObjectMeta(name)




 Error: Found comments without trailing period. Comments has to be full sentences.
```

### DIFF
--- a/docs/components/compact.md
+++ b/docs/components/compact.md
@@ -9,7 +9,7 @@ menu: components
 The compactor component of Thanos applies the compaction procedure of the Prometheus 2.0 storage engine to block data stored in object storage.
 It is generally not semantically concurrency safe and must be deployed as a singleton against a bucket.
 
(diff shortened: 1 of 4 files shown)


daixiang0:
Yep, i run those cmd by hand then fix them. I will dive into it to find why CI does not work as expected. Maybe those files never changed since CI enable.

PR #1878, opened 2019-12-12
fix syntax error

daixiang0:

To fix below error introduced by #1801 :
```
# make lint
>> linting all of the Go files GOGC=
>> detecting misspells
>> detecting white noise
/bin/sh: -c: line 0: conditional binary operator expected
/bin/sh: -c: line 0: syntax error near `diff-files'
/bin/sh: -c: line 0: `if [[ ! git diff-files --quiet --ignore-submodules -- ]]; then \'
```

Correct syntax:
```
# [[ ! git diff-files --quiet --ignore-submodules -- ]] && echo yes
-bash: conditional binary operator expected
-bash: syntax error near `diff-files'
# ! git diff-files --quiet --ignore-submodules -- && echo yes
yes
# [[ ! $(git diff-files --quiet --ignore-submodules --) ]] && echo yes
yes
```

Also, update index before diff or diff would not work as expected.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -181,7 +181,7 @@ check-docs: $(EMBEDMD) $(LICHE) build
 	@$(LICHE) --recursive docs --exclude "(cloud.tencent.com|alibabacloud.com)" --document-root .
 	@$(LICHE) --exclude "(cloud.tencent.com|goreportcard.com|alibabacloud.com)" --document-root . *.md
 	@find -type f -name "*.md" | xargs scripts/cleanup-white-noise.sh
-	@if [[ ! git diff-files --quiet --ignore-submodules -- ]]; then \
+	@if ! git diff-files --quiet --ignore-submodules -- ; then \
 		echo >&2 "please clean up white noise in all docs"; \
 		exit 1; \
 	fi
@@ -304,7 +304,7 @@ lint: check-git $(GOLANGCILINT) $(MISSPELL)
 	@find . -type f | grep -v vendor/ | grep -vE '\./\..*' | xargs $(MISSPELL) -error
 	@echo ">> detecting white noise"
 	@find . -type f \( -name "*.md" -o -name "*.go" \) | xargs scripts/cleanup-white-noise.sh
-	@if [[ ! git diff-files --quiet --ignore-submodules -- ]]; then \
+	@if ! git diff-files --quiet --ignore-submodules -- ; then \
 		echo >&2 "please clean up white noise in all docs or Go files"; \
 		exit 1; \
 	fi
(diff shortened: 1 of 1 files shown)


daixiang0:
@bwplotka mind have a look?

Makefile
```diff
@@ -181,7 +181,8 @@ check-docs: $(EMBEDMD) $(LICHE) build
 	@$(LICHE) --recursive docs --exclude "(cloud.tencent.com|alibabacloud.com)" --document-root .
 	@$(LICHE) --exclude "(cloud.tencent.com|goreportcard.com|alibabacloud.com)" --document-root . *.md
 	@find -type f -name "*.md" | xargs scripts/cleanup-white-noise.sh
-	@if [[ ! git diff-files --quiet --ignore-submodules -- ]]; then \
+	@git update-index -q --ignore-submodules --refresh
```
squat: do we need these new lines?

daixiang0: Yes, need to update index before do git diff.

Makefile
```diff
@@ -96,23 +96,17 @@ define fetch_go_bin_version
...
 
     @if ! git diff-files --quiet --ignore-submodules --; then \
-        echo >&2 "cannot $1: you have unstaged changes."; \
-        git diff-files --name-status -r --ignore-submodules -- >&2; \
-        echo >&2 "Please commit or stash them."; \
+        echo >&2 "$(if $1,$1,You have unstaged changes. Please commit or stash them.)"; \
         exit 1; \
     fi
-
-    @if ! git diff-index --cached --quiet HEAD --ignore-submodules --; then \
```
bwplotka: This function was also used previously to ensure a clean git state. Is it really safe to allow `staged` then?

Essentially this function is used for other cases then just check if something was changed in for format/clean whitespaces (:

daixiang0: In fact, format or clean whitespaces does not exit 1 but change directly, so git state check works.
With default help info, we can handle most case i think.

bwplotka: > In fact, format or clean whitespaces does not exit 1 but change directly, so git state check works.

I know and that's ok. We use this function to exit with 1.

Ok, checked and we are the only users of this function now. This means that it's fine to change it in the way you want but either let's rename it or get back the check for `staged` changes as the function name is misleading now: In this form, the function does not check clean working tree properly.

daixiang0: I split them into 2 function, may be more clear.

bwplotka: Then let's remove `ensure_no_staged` if we don't use it. (: Why we cannot keep all in same function and just not remove staged one?

daixiang0: I think the staged one is more useful at local, no staged change could exist in a pr, so i split them.
I would like to add ensure_no_staged in `make docs` and `make format` if possible.

daixiang0:
@bwplotka mind have a look?

Makefile
bwplotka: but we never use it  (:

squat:
@daixing0 I’m so sorry, I merged #1908 which now caused a conflict here 😔 sorry for the trouble

bwplotka:
I thought I commented but seems not: We had a clash and both PRs this and https://github.com/thanos-io/thanos/pull/1908 was fixing the same thing in very similar way. Changes were making sense on both, so we had to choose one (: Since https://github.com/thanos-io/thanos/pull/1908 was already ready we merged that. Hope you don't mind!

Thanks for you work anyway, I think we can close this PR.

PR #1881, opened 2019-12-12
Add memcached support to index cache

pracucci:
This PR proposes to introduce memcached support for the index cache. Few reasons why memcached may make sense:

1. Do not wipe out the index cache on thanos store restarts/redeploys
2. Scale the cache size above the single thanos store memory limits (ie. a cluster of memcached nodes)
3. _Use it in Cortex, where we're experimenting a TSDB storage based on some Thanos components and we currently have the "bucket store" replicated across different nodes and a shared cache may benefit_

**Open questions**:
- Should `MaxGetMultiBatchConcurrency` be global (like right now) or a per `MultiGet()` call limit? I would be more lean towards the latter (see TODO in `pkg/cacheutil/memcached_client.go`)

## Changes

- Moved index cache key struct outside of the file containing the in-memory cache backend because generic
- Added `MemcachedIndexCache` support, with the backend client inspired by Cortex, with some substantial differences:
  - DNS discovery based on the Thanos one
  - `SetAsync()` instead of `Set()`

## Verification

Manual and unit tests.

### DIFF
--- a/cmd/thanos/store.go
+++ b/cmd/thanos/store.go
@@ -40,7 +40,7 @@ func registerStore(m map[string]setupFunc, app *kingpin.Application) {
 		Default("250MB").Bytes()
 
 	indexCacheConfig := extflag.RegisterPathOrContent(cmd, "index-cache.config",
-		"YAML file that contains index cache configuration. See format details: https://thanos.io/components/store.md/",
+		"YAML file that contains index cache configuration. See format details: https://thanos.io/components/store.md/#index-cache",
 		false)
 
 	chunkPoolSize := cmd.Flag("chunk-pool-size", "Maximum size of concurrently allocatable bytes for chunks.").
--- a/cmd/thanos/store.go
+++ b/cmd/thanos/store.go
@@ -36,11 +36,11 @@ func registerStore(m map[string]setupFunc, app *kingpin.Application) {
 	dataDir := cmd.Flag("data-dir", "Data directory in which to cache remote blocks.").
 		Default("./data").String()
 
-	indexCacheSize := cmd.Flag("index-cache-size", "Maximum size of items held in the in-memory index cache. Ignored if --index-cache.config option is specified.").
+	indexCacheSize := cmd.Flag("index-cache-size", "Maximum size of items held in the in-memory index cache. Ignored if --index-cache.config or --index-cache.config-file option is specified.").
 		Default("250MB").Bytes()
 
 	indexCacheConfig := extflag.RegisterPathOrContent(cmd, "index-cache.config",
-		"YAML file that contains index cache configuration. See format details: https://thanos.io/caching.md/",
+		"YAML file that contains index cache configuration. See format details: https://thanos.io/components/store.md/",
 		false)
 
 	chunkPoolSize := cmd.Flag("chunk-pool-size", "Maximum size of concurrently allocatable bytes for chunks.").
--- a/cmd/thanos/store.go
+++ b/cmd/thanos/store.go
@@ -194,7 +194,7 @@ func runStore(
 		indexCache, err = storecache.NewIndexCache(logger, indexCacheContentYaml, reg)
 	} else {
 		indexCache, err = storecache.NewInMemoryIndexCacheWithConfig(logger, reg, storecache.InMemoryIndexCacheConfig{
-			MaxSizeBytes:     indexCacheSizeBytes,
+			MaxSizeBytes:     storecache.Bytes(indexCacheSizeBytes),
 			MaxItemSizeBytes: storecache.DefaultInMemoryIndexCacheConfig.MaxItemSizeBytes,
 		})
 	}
--- a/cmd/thanos/store.go
+++ b/cmd/thanos/store.go
@@ -36,8 +36,12 @@ func registerStore(m map[string]setupFunc, app *kingpin.Application) {
 	dataDir := cmd.Flag("data-dir", "Data directory in which to cache remote blocks.").
 		Default("./data").String()
 
-	indexCacheSize := cmd.Flag("index-cache-size", "Maximum size of items held in the index cache.").
+	indexCacheSize := cmd.Flag("index-cache-size", "Maximum size of items held in the in-memory index cache. Ignored if --index-cache.config option is specified.").
 		Default("250MB").Bytes()
+
+	indexCacheConfig := extflag.RegisterPathOrContent(cmd, "index-cache.config",
+		"YAML file that contains index cache configuration. See format details: https://thanos.io/caching.md/",
+		false)
 
 	chunkPoolSize := cmd.Flag("chunk-pool-size", "Maximum size of concurrently allocatable bytes for chunks.").
 		Default("2GB").Bytes()
@@ -77,6 +81,7 @@ func registerStore(m map[string]setupFunc, app *kingpin.Application) {
 			logger,
 			reg,
 			tracer,
+			indexCacheConfig,
 			objStoreConfig,
 			*dataDir,
 			*grpcBindAddr,
@@ -110,6 +115,7 @@ func runStore(
 	logger log.Logger,
 	reg *prometheus.Registry,
 	tracer opentracing.Tracer,
+	indexCacheConfig *extflag.PathOrContent,
 	objStoreConfig *extflag.PathOrContent,
 	dataDir string,
 	grpcBindAddr string,
@@ -169,6 +175,11 @@ func runStore(
 		return err
 	}
 
+	indexCacheContentYaml, err := indexCacheConfig.Content()
+	if err != nil {
+		return errors.Wrap(err, "get content of index cache configuration")
+	}
+
 	// Ensure we close up everything properly.
 	defer func() {
 		if err != nil {
@@ -176,13 +187,17 @@ func runStore(
 		}
 	}()
 
-	// TODO(bwplotka): Add as a flag?
-	maxItemSizeBytes := indexCacheSizeBytes / 2
-
-	indexCache, err := storecache.NewInMemoryIndexCache(logger, reg, storecache.Opts{
-		MaxSizeBytes:     indexCacheSizeBytes,
-		MaxItemSizeBytes: maxItemSizeBytes,
-	})
+	// Create the index cache loading its config from config file, while keeping
+	// backward compatibility with the pre-config file era.
+	var indexCache storecache.IndexCache
+	if len(indexCacheContentYaml) > 0 {
+		indexCache, err = storecache.NewIndexCache(logger, indexCacheContentYaml, reg)
+	} else {
+		indexCache, err = storecache.NewInMemoryIndexCacheWithConfig(logger, reg, storecache.InMemoryIndexCacheConfig{
+			MaxSizeBytes:     indexCacheSizeBytes,
+			MaxItemSizeBytes: storecache.DefaultInMemoryIndexCacheConfig.MaxItemSizeBytes,
+		})
+	}
 	if err != nil {
 		return errors.Wrap(err, "create index cache")
 	}
--- a/cmd/thanos/store.go
+++ b/cmd/thanos/store.go
@@ -194,8 +194,8 @@ func runStore(
 		indexCache, err = storecache.NewIndexCache(logger, indexCacheContentYaml, reg)
 	} else {
 		indexCache, err = storecache.NewInMemoryIndexCacheWithConfig(logger, reg, storecache.InMemoryIndexCacheConfig{
-			MaxSizeBytes:     storecache.Bytes(indexCacheSizeBytes),
-			MaxItemSizeBytes: storecache.DefaultInMemoryIndexCacheConfig.MaxItemSizeBytes,
+			MaxSize:     storecache.Bytes(indexCacheSizeBytes),
+			MaxItemSize: storecache.DefaultInMemoryIndexCacheConfig.MaxItemSize,
 		})
 	}
 	if err != nil {
--- a/docs/components/store.md
+++ b/docs/components/store.md
@@ -217,12 +217,13 @@ config:
 
   # Maximum number of concurrent batch executions when fetching keys from memcached.
   # Each batch can fetch up to max_get_multi_batch_size keys.
-  max_get_multi_batch_concurrency: 20
+  max_get_multi_batch_concurrency: 0
 
   # Maximum number of keys a single underlying GetMulti() operation should run. If
   # more keys are specified, internally keys are splitted into multiple batches and
   # fetched concurrently up to max_get_multi_batch_concurrency parallelism.
-  max_get_multi_batch_size: 1024
+  # If set to 0, the max batch size is unlimited.
+  max_get_multi_batch_size: 0
 
   # DNS discovery update interval.
   dns_provider_update_interval: 10s
--- a/docs/components/store.md
+++ b/docs/components/store.md
@@ -198,7 +198,7 @@ The `memcached` index cache allows to use [Memcached](https://memcached.org) as 
 ```yaml
 type: MEMCACHED
 config:
-  addrs: []
+  addresses: []
   timeout: 0s
   max_idle_connections: 0
   max_async_concurrency: 0
@@ -210,7 +210,7 @@ config:
 
 The **required** settings are:
 
-- `addrs`: list of memcached addresses, that will get resolved with the [DNS service discovery](../service-discovery.md/#dns-service-discovery) provider.
+- `addresses`: list of memcached addresses, that will get resolved with the [DNS service discovery](../service-discovery.md/#dns-service-discovery) provider.
 
 While the remaining settings are **optional**:
 
--- a/docs/components/store.md
+++ b/docs/components/store.md
@@ -203,7 +203,7 @@ config:
   max_idle_connections: 0
   max_async_concurrency: 0
   max_async_buffer_size: 0
-  max_get_multi_batch_concurrency: 0
+  max_get_multi_concurrency: 0
   max_get_multi_batch_size: 0
   dns_provider_update_interval: 0s
 ```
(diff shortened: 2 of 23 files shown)


pkg/cacheutil/memcached_client.go
```diff
@@ -0,0 +1,382 @@
...
+		go c.getMultiQueueProcessLoop()
+	}
+
+	return c, nil
+}
+
+func (c *memcachedClient) Stop() {
+	close(c.stop)
+
+	// Wait until all workers have terminated.
```
thorfour: do we also want to wait for the resolve addrs loop?

pkg/cacheutil/memcached_client.go
```diff
@@ -0,0 +1,382 @@
...
+	// Channel used to notify internal goroutines when they should quit.
+	stop chan struct{}
+
+	// Channel used to enqueue async operations.
+	asyncQueue   chan func()
+	asyncWorkers sync.WaitGroup
+
+	// Channel used to enqueue get multi operations.
+	getMultiQueue   chan *memcachedGetMultiBatch
+	getMultiWorkers sync.WaitGroup
```
thorfour: why not use a single wg for all spawned bg routines?

pkg/cacheutil/memcached_client.go
```diff
@@ -0,0 +1,382 @@
...
+		items, err := c.client.GetMulti(keys)
+		if err != nil {
+			return nil, err
+		}
+
+		return []map[string]*memcache.Item{items}, nil
+	}
+
+	// Split input keys into batches and schedule a job for it.
+	results := make(chan *memcachedGetMultiResult)
```
thorfour: why not size this based on numResults to keep from blocking the multi queue?

pkg/cacheutil/memcached_client.go
pracucci: I thought about it, but I don't see any benefit waiting for it. Do you have any thought on it?

pkg/cacheutil/memcached_client.go
pracucci: Definitely. Fixed.

pkg/cacheutil/memcached_client.go
pracucci: Makes sense. Fixed.

pkg/cacheutil/memcached_client.go
```diff
@@ -0,0 +1,382 @@
...
+	defaultMaxGetMultiBatchSize        = 1024
+)
+
+var (
+	errMemcachedAsyncBufferFull = errors.New("the async buffer is full")
+	errMemcachedConfigNoAddrs   = errors.New("no memcached addrs provided")
+)
+
+// MemcachedClient is a high level client to interact with memcached.
+type MemcachedClient interface {
```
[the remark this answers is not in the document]
pracucci: > Do we need another interface?

The interface is currently used to mock it in `pkg/store/cache/memcached_test.go`. It may also be need to add multi-tenancy support in Cortex: an option may be having 1 single underlying memcached client and wrapping it through a proxy for each tenant having the proxy prefixing the keys with the tenant ID (but no decision has been taken and I'm open to suggestions).

> I think we way we do it is to accept interfaces but return structs as advised by Go (:

Right, done.

pkg/cacheutil/memcached_client.go
```diff
@@ -0,0 +1,382 @@
...
+)
+
+const (
+	defaultTimeout                     = 100 * time.Millisecond
+	defaultMaxIdleConnections          = 100
+	defaultMaxAsyncConcurrency         = 20
+	defaultMaxAsyncBufferSize          = 10000
+	defaultDNSProviderUpdateInterval   = 10 * time.Second
+	defaultMaxGetMultiBatchConcurrency = 20
+	defaultMaxGetMultiBatchSize        = 1024
```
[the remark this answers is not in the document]
pracucci: Not really. I picked it from the current value we're setting it in our Cortex clusters, for the same feature. Any suggestion?

pkg/cacheutil/memcached_client.go
```diff
@@ -0,0 +1,382 @@
...
+	"time"
+
+	"github.com/bradfitz/gomemcache/memcache"
+	"github.com/go-kit/kit/log"
+	"github.com/go-kit/kit/log/level"
+	"github.com/thanos-io/thanos/pkg/discovery/dns"
+)
+
+const (
+	defaultTimeout                     = 100 * time.Millisecond
```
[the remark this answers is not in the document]
pracucci: Maybe. Generally, we want a cache being significantly faster than the backend storage. Assuming a 1ms networking latency to comunicate to a memcached cluster running within the same cloud region, 100ms should be enough (ie. it's what we set in our Cortex clusters).

However, I do see that it's probably a low value for a default. What would you suggest?

pkg/cacheutil/memcached_client.go
```diff
@@ -0,0 +1,382 @@
...
+func (c *memcachedClient) GetMulti(keys []string) (map[string][]byte, error) {
+	batches, err := c.getMultiBatched(keys)
+	if err != nil {
+		if len(batches) == 0 {
+			return nil, err
+		}
+
+		// In case we have both results and an error, it means some batch requests
+		// failed and other succeeded. In this case we prefer to log it and move on,
+		// given returning from results from the cache is better than returning
```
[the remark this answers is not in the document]
pracucci: > Actually do we really need this second layer (memcachedClient)? We have already dependency injection for client backend itself (Set, GetMulti). While separation of methods is ok, I am not sure if we need really another separate struct/interface

I would suggest to keep a generic `MemcachedClient` separated from the index cache to allow us to use memcached for other caching purposes as well in the future (ie. I have some ideas I would like to propose which are outside the scope of this PR).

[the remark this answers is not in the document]
pracucci: > what do you think on making this interface to not return error at all? Quite odd to log error here on the partial result but log on client side + metric if no result

Agree. It was odd. I have removed the `error` from the return values and introduced metrics and tracing in `MemcachedClient` (we're not able anymore to track failures from the index cache implementation).

pracucci:
@bwplotka Thanks for your initial review! I've addressed your feedback (or commented otherwise), added the config file and written some doc. May you take another look, please?

pkg/cacheutil/memcached_client.go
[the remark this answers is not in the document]
bwplotka: Nope let's start with something

pkg/cacheutil/memcached_client.go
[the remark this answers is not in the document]
bwplotka: Let's gather some data and start with something (:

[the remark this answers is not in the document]
pracucci: In our Cortex clusters running on GCE we measure a 99th memcached latency of about 15ms (pretty constant) and max latency between 50ms and 250ms. I agree a default timeout should be conservative, so I'm going to set it to 500ms here, and then can be adjusted by the user (ie. in our cluster I don't see a good reason of setting it above 100ms, given we want to skip the cache if it's too slow for any reason).

pkg/cacheutil/memcached_client.go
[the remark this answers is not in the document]
pracucci: I've pushed a commit taking a different approach:
- The value of `0` disables it
- The default is disabled and can be enabled by the user if hits any memcached limit

pkg/store/cache/cache.go
```diff
@@ -1,22 +1,79 @@
...
+	case cacheKeySeries:
+		return 16 + 8 // ULID + uint64.
+	}
+	return 0
+}
+
+func (c cacheKey) string() string {
+	switch c.key.(type) {
+	case cacheKeyPostings:
+		// Do not use non cryptographically hash functions to avoid hash collisions
```
bwplotka:
```suggestion
		// Use cryptographically hash functions to avoid hash collisions
```

pkg/store/cache/cache.go
```diff
@@ -1,22 +1,79 @@
...
+		// which would end up in wrong query results
```
bwplotka: missing trailing period

pkg/store/cache/cache_test.go
```diff
@@ -0,0 +1,99 @@
...
+		})
+	}
+}
+
+func TestCacheKey_string_ShouldGuaranteeReasonablyShortKeyLength(t *testing.T) {
+	t.Parallel()
+
+	uid := ulid.MustNew(1, nil)
+
+	tests := map[string]struct {
```
bwplotka: Usually we do list for deterministic order, but that's fine here

pkg/cacheutil/memcached_client.go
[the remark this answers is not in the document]
bwplotka: It sounds like a YAGNI here. Does it mean creating distributing each client implementation for each block? (: Sounds like something we can add unless we know clear decision here, but let's discuss

pkg/cacheutil/memcached_client.go
```diff
@@ -0,0 +1,448 @@
...
+		span, _ := tracing.StartSpan(ctx, "memcached_set")
+		err := c.client.Set(&memcache.Item{
+			Key:        key,
+			Value:      value,
+			Expiration: int32(time.Now().Add(ttl).Unix()),
+		})
+		span.Finish()
+		if err != nil {
+			c.failures.WithLabelValues(opSet).Inc()
+			level.Warn(c.logger).Log("msg", fmt.Sprintf("failed to store item with key %s to memcached", key), "err", err)
```
bwplotka: What about having key as a separate struct in structured log? (:

pkg/cacheutil/memcached_client.go
```diff
@@ -0,0 +1,448 @@
...
+		for key, item := range items {
+			hits[key] = item.Value
+		}
+	}
+
+	return hits
+}
+
+func (c *memcachedClient) getMultiBatched(ctx context.Context, keys []string) ([]map[string]*memcache.Item, error) {
+	// Do not batch if the input keys are less then the max batch size.
```
bwplotka:
```suggestion
	// Do not batch if the input keys are less than the max batch size.
```

pkg/cacheutil/memcached_client.go
```diff
@@ -0,0 +1,448 @@
...
+	defer close(results)
+
+	go func() {
+		for batchStart := 0; batchStart < len(keys); batchStart += batchSize {
+			batchEnd := batchStart + batchSize
+			if batchEnd > len(keys) {
+				batchEnd = len(keys)
+			}
+
+			c.getMultiQueue <- &memcachedGetMultiBatch{
```
bwplotka: We lost the discussion on GitHub but I guess at the end we decided on per client concurrency for now? 

Thinking loudly: if we choose this I would say we need good observability: how to find latency caused by starvation of this queue? The latency metric is per single call to Memcache (which is fair for from mameched client perspective).

I think tracing would help as we already create some spans here e.g in `getMultiSingle` so should be enough..

pkg/store/cache/cache.go
```diff
@@ -1,22 +1,79 @@
...
+		return cacheTypeSeries
+	}
+	return "<unknown>"
+}
+
+func (c cacheKey) size() uint64 {
+	switch k := c.key.(type) {
+	case cacheKeyPostings:
+		// ULID + 2 slice headers + number of chars in value and name.
+		return 16 + 2*sliceHeaderSize + uint64(len(k.Value)+len(k.Name))
```
thorfour: does the ulid size of 16 exist anywhere as a const? Maybe at least make it a const in this file

docs/components/store.md
```diff
@@ -151,3 +163,67 @@ Filtering is done on a Chunk level, so Thanos Store might still return Samples w
...
+
+The `in-memory` index cache is enabled by default and its max size can be configured through the flag `--index-cache-size`.
+
+Alternatively, the `in-memory` index cache can also by configured using `--index-cache.config-file` to reference to the configuration file or `--index-cache.config` to put yaml config directly:
+
+```
+type: in-memory
+config:
+  # Maximum number of bytes the cache can contain (defaults to 250MB).
+  max_size_bytes: 262144000
```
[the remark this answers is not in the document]
GiedriusS: Perhaps the same kingpin code can be reused here: https://godoc.org/github.com/alecthomas/units#Base2Bytes ?

pkg/cacheutil/jump_hash.go
```diff
@@ -0,0 +1,18 @@
+package cacheutil
+
+// JumpHash consistently chooses a hash bucket number in the range
+// [0, numBuckets) for the given key. numBuckets must be >= 1.
+//
+// Copied from github.com/dgryski/go-jump/blob/master/jump.go (MIT license)
+func JumpHash(key uint64, numBuckets int) int32 {
```
GiedriusS: Unless I am missing something but does this need to be exposed?

pkg/cacheutil/memcached_client.go
GiedriusS: Maybe our current gate https://github.com/thanos-io/thanos/blob/master/pkg/store/gate.go could be adopted here somehow? It would be nice to have some metric here because otherwise all users would need to have some kind of tracing enabled.

pkg/cacheutil/memcached_client.go
```diff
@@ -0,0 +1,448 @@
...
+				ctx:     ctx,
+				keys:    keys[batchStart:batchEnd],
+				results: results,
+			}
+		}
+	}()
+
+	// Wait for all batch results. In case of error, we keep
+	// track of the last error occurred.
+	items := make([]map[string]*memcache.Item, 0, numResults)
```
GiedriusS: Maybe `sync.Pool` could be used here?

cmd/thanos/store.go
```diff
@@ -36,9 +36,13 @@ func registerStore(m map[string]setupFunc, app *kingpin.Application) {
 	dataDir := cmd.Flag("data-dir", "Data directory in which to cache remote blocks.").
 		Default("./data").String()
 
-	indexCacheSize := cmd.Flag("index-cache-size", "Maximum size of items held in the index cache.").
+	indexCacheSize := cmd.Flag("index-cache-size", "Maximum size of items held in the in-memory index cache. Ignored if --index-cache.config or --index-cache.config-file option is specified.").
 		Default("250MB").Bytes()
 
+	indexCacheConfig := extflag.RegisterPathOrContent(cmd, "index-cache.config",
```
GiedriusS: 👍

pkg/cacheutil/memcached_client.go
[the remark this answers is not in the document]
GiedriusS: That sounds good to me. @bwplotka what about you?

docs/components/store.md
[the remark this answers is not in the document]
pracucci: I've introduced `storecache.Bytes`, whose yaml marshalling is based on `units.Base2Bytes`, to allow to configure in-memory index cache size using byte units.

What's your take?

docs/components/store.md
```diff
@@ -151,3 +163,67 @@ Filtering is done on a Chunk level, so Thanos Store might still return Samples w
...
+  max_item_size_bytes: 131072000
+```
+
+### Memcached index cache
+
+The `memcached` index cache allows to use [Memcached](https://memcached.org) as cache backend. This cache type is configured using `--index-cache.config-file` to reference to the configuration file or `--index-cache.config` to put yaml config directly:
+
+```
+type: memcached
+config:
```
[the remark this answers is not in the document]
pracucci: Sure. I've added the support to `cfggen` and documented the config options below the config file format.

What's your take?

pkg/cacheutil/memcached_client.go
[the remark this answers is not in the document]
pracucci: I'm not getting if the comment is about (a) removing the `MemcachedClient` interface or (b) merging `memcachedClient` with `MemcachedIndexCache`?

pkg/cacheutil/memcached_client.go
bwplotka: I like the metric standpoint, that's a good point. Do you think @pracucci we can adapt the existing way of queuing requests via `gate` in this way? as suggested by @GiedriusS ? I don't have a strong opinion but can see those pros & cons:

Advantages:
* Reuse
* Latency + queue size metric
* We can switch from global to per request limit easily `gate` here and there.

Downsides:
* Each gated request requires a new goroutine. Bit more goroutine overhead and coordination.

I think for this iteration I would be happy with what we have now.

pkg/cacheutil/memcached_client.go
bwplotka: Maybe, but fine to play with it with some microbenchmark on later PRs.

pkg/store/cache/units.go
```diff
@@ -0,0 +1,29 @@
+package storecache
+
+import (
+	"github.com/alecthomas/units"
+)
+
+// Bytes is a data type which supports yaml serialization/deserialization
```
bwplotka: Thanks for this :+1:

bwplotka:
I think I am happy to merge this in this form (after rebase) (: 

@squat @thorfour @brancz anyone wants to take a look before we do that?

pkg/cacheutil/memcached_client.go
pracucci: I personally see a big value in having such metrics. Also the gate would simplify the implementation (less code).

> We lost the discussion on GitHub but I guess at the end we decided on per client concurrency for now?

Re-iterating on this, I think the current implementation is confusing. We're currently limiting the batched `GetMulti()`  but not non-batched `GetMulti()`. For example, if you have `MaxGetMultiBatchSize=1000` and `MaxAsyncConcurrency=10` and you concurrently call 100 times `MemcachedClient.GetMulti(1K keys)` we end up with 100 concurrent underlying requests (each with 1K keys for a total of 100K keys). However, if you call once `MemcachedClient.GetMulti(100K keys)` we end up with 10 concurrent underlying requests, each with 1K keys, until all 100K keys are fetched.

I see three options:
1. Keep it as is (but weird)
2. Enforce the concurrency limit even to non-batched requests
  - Pro: the concurrency limit is an effective max number of concurrent fetch requests
  - Con: a single request with a bunch of keys will slow down everything else because of the queue
3. Switch to a per-request gate
  - Pro: a single request with a bunch of keys doesn't slow down others
  - Con: there's no way to set a cap on the max number of connections towards memcached

I'm currently more keen to #2, cause it looks the safest option.

pracucci: > Do you think @pracucci we can adapt the existing way of queuing requests via gate in this way? as suggested by @GiedriusS ?

I think so and I believe it would make sense. Observability - at this stage - is more important than goroutine overhead and we could implement it in a way that goroutines are created only for batched requests (so as far as the batching is disabled or requests do not exceed the max batch size, there will be no goroutine overhead).

pracucci: In [this commit](https://github.com/thanos-io/thanos/pull/1881/commits/ee88dd5c269ab5caf76da449dc42399a67c8e10c) I've:
1. Switched the max get multi concurrency to a gate
2. Applied the gate also to non batched requests

What's your take?

pracucci:
Thanks @bwplotka for your extensive review and sorry for this big PR. Next time I will address changes in a more iterative way. I went through a self review and further manual tests and it looks good to me, but I will be happy to re-iterate on it in case of further feedback.

> I think I am fine with adding this as long as there will be someone to maintain this code other than a few of us from maintainer team

Sure. I'm OK maintaining it. Feel free to assign me memcached-related issues in the next future.

GiedriusS:
Last small conflict needs to be solved before the final review. Thanks for your work!

pracucci:
@GiedriusS I've rebased `master` and fixed the conflict. Thanks!

bwplotka:
Thanks @pracucci

PR #1882, opened 2019-12-12
cmd/thanos/receive: dont upload as sidecar

squat:
As pointed out by a user on the CNCF Slack, the receiver is currently
uploading to object storage as the sidecar.

cc @metalmatze @brancz 

### DIFF
--- a/cmd/thanos/receive.go
+++ b/cmd/thanos/receive.go
@@ -164,7 +164,7 @@ func runReceive(
 	replicaHeader string,
 	replicationFactor uint64,
 	tsdbBlockDuration model.Duration,
-	comp component.Component,
+	comp component.SourceStoreAPI,
 ) error {
 	logger = log.With(logger, "component", "receive")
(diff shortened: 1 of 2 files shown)


cmd/thanos/receive.go
```diff
@@ -407,7 +407,7 @@ func runReceive(
 	if upload {
 		// The background shipper continuously scans the data directory and uploads
 		// new blocks to Google Cloud Storage or an S3-compatible storage service.
-		bkt, err := client.NewBucket(logger, confContentYaml, reg, component.Sidecar.String())
+		bkt, err := client.NewBucket(logger, confContentYaml, reg, component.Receive.String())
```
brancz: We have the “comp” variable that we can use here no?

squat: this func expects a `component.SourceStoreAPI`, not just a `component.Component`. I could type assert here or change the function signature of `runReceive`

PR #1887, opened 2019-12-13
deduplicate addresses.

johncming:
Multiple groups of addresses may have duplicates


pkg/discovery/cache/cache.go
```diff
@@ -38,12 +38,21 @@ func (c *Cache) Update(tgs []*targetgroup.Group) {
...
 // Addresses returns all the addresses from all target groups present in the Cache.
 func (c *Cache) Addresses() []string {
+	var addresses []string
+	var unique map[string]struct{}
+
 	c.Lock()
 	defer c.Unlock()
-	var addresses []string
+
+	unique = make(map[string]struct{})
```
bwplotka: Do you mind adding test case for it? (:

johncming: ok.

PR #1891, opened 2019-12-13
Rename components

### LINKED ISSUE #1871: Move components names to `Ruler`, `Compactor`, `Querier` (`-er`)
It's maybe a minor but we should be consistent on naming. In the documentation, code and `thanos` command we have `rule, compact, downample, query` vs in some docs, deployments and most of the presentations we are leaking `Querier`, `Ruler`, `Compactor` 

This conflicts a bit with `Store Gateway` but we already have some discussion about renaming it: https://github.com/thanos-io/thanos/issues/1646

I would vote `-er` as it's more correct (aka `go` way of naming objects as well) and it avoids name collisions like `rule` (recording/alerting rule) vs `rule` (the component responsible for evaluating rules). 

**AC**: 
 < to agree>

Any thoughts?

yeya24:

## Changes

Fixes #1871

### DIFF
--- a/docs/design.md
+++ b/docs/design.md
@@ -16,14 +16,14 @@ The Prometheus metric data model and the 2.0 storage format ([spec][tsdb-format]
 Thanos is a clustered system of components with distinct and decoupled purposes. Clustered components can be categorized as follows:
 
 * Metric sources
-* Stores
+* Store Gateways
 * Queriers
 
 ### Metric Sources
 
 A data source is a very generalized definition of a component that produces or collects metric data. Source advertise this data in the cluster to potential clients. Metric data can be retrieved via a well-known gRPC service.
 
-Thanos provides two components that act as data sources: the Prometheus sidecar and rule nodes.
+Thanos provides three components that act as data sources: the Prometheus sidecar, Ruler and Receiver.
 
 The sidecar implements the gRPC service on top of Prometheus' [HTTP and remote-read APIs][prom-http-api]. The rule node directly implements it on top of the Prometheus storage engine it is running.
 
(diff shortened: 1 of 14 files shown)


yeya24:
Forgot this. I will rebase it tomorrow!

bwplotka:
@yeya24, before you rebase have you seen our final decision about naming? https://thanos.io/contributing/#components-naming

kakkoyun:
@yeya24 Correct me if I'm wrong, but the decision was actually to keep command names as is in the code, no?

yeya24:
@bwplotka  @kakkoyun 
Thanks! Seems I need to update my PR first!
Just to confirm what place should I change:
- Docs
- Code comments

And a small question: What about `receive`? I don' t see it in https://thanos.io/contributing/#components-naming. Do I need to rename it to `receiver`?

kakkoyun:
@yeya24 It should be `receive` as well, as in command. `thanos receive`. We can also update the that in the document.

yeya24:
> @yeya24 It should be `receive` as well, as in command. `thanos receive`. We can also update the that in the document.

Thanks. I will pin you after I finish.

docs/design.md
```diff
@@ -54,7 +54,7 @@ The meta.json is updated during upload time on sidecars.
 
 ```
 ┌────────────┬─────────┐         ┌────────────┬─────────┐     ┌─────────┐
-│ Prometheus │ Sidecar │   ...   │ Prometheus │ Sidecar │     │   Rule  │
+│ Prometheus │ Sidecar │   ...   │ Prometheus │ Sidecar │     │  Ruler  │
```
yeya24: Maybe we need to update the graph to add `Receiver`

yeya24:
I think it is ready. Based on https://thanos.io/contributing/#components-naming, most changes are just renaming in docs.

PR #1895, opened 2019-12-14
pkg/receive: simplify hashring file

squat:
This commit simplifies the receive/hashring.go file a bit. Most
significantly it changes two sequential `if`s that can never both be
true, to an `if/else`.

### DIFF
--- a/pkg/receive/hashring.go
+++ b/pkg/receive/hashring.go
@@ -7,9 +7,8 @@ import (
 	"sort"
 	"sync"
 
-	"github.com/prometheus/prometheus/prompb"
-
 	"github.com/cespare/xxhash"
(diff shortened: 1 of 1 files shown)


pkg/receive/hashring.go
```diff
@@ -120,12 +119,9 @@ func (m *multiHashring) GetN(tenant string, ts *prompb.TimeSeries, n uint64) (st
 		// considered a default hashring and matches everything.
 		if t == nil {
 			found = true
-		}
-		if _, ok := t[tenant]; ok {
+		} else if _, ok := t[tenant]; ok {
 			found = true
 		}
-		// If the hashring has no tenants, then it is
```
squat: this exact comment is already written above at 120, where it belongs

PR #1897, opened 2019-12-16
bugfix: adjust the timeout.

johncming:
repeat timeout:
https://github.com/thanos-io/thanos/blob/2f29975982e25632a3eaccaf47d2401dcbfeef4b/cmd/thanos/query.go#L299-L303

update grpc timeout:
https://github.com/thanos-io/thanos/blob/2f29975982e25632a3eaccaf47d2401dcbfeef4b/pkg/query/storeset.go#L406

the `grpc timeout` should be lower than `repeat timeout`.

### DIFF
--- a/pkg/query/storeset.go
+++ b/pkg/query/storeset.go
@@ -193,7 +193,7 @@ func NewStoreSet(
 		storeSpecs:            storeSpecs,
 		dialOpts:              dialOpts,
 		storesMetric:          storesMetric,
-		gRPCInfoCallTimeout:   10 * time.Second,
+		gRPCInfoCallTimeout:   5 * time.Second,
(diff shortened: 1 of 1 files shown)


johncming:
> I think it makes sense, but let's round it to 5s.
> 
> This does not make any problem for `Repeat` itself, but indeed we might cause less often update because of the slow store.
> 
> Have you noticed something like this already?

Yes, 5s is good.

In my enviroment, query is slower than before when a store down. Then found the timeout values. Indeed, may be is not bug in the beginning, but after a longer time, storesets will older than expected which is hard to know latency.

bwplotka:
Can we have set  this to 5s then?

johncming:
> Can we have set this to 5s then?

fixed.

PR #1902, opened 2019-12-17
sidecar: don't require Prometheus min/max block durations to be equal

cstyan:
@bwplotka

Adds a flag that allows users to disable the requirement that their Prometheus instances have equal min/max block durations if they have enabled block upload. 

As discussed on slack; because only uncompacted blocks are uploaded, in the event of an upload failure or network issues that last longer than 4h there is the potential for data loss (in the bucket storage) if blocks are compacted by Prometheus during that time

Signed-off-by: Callum Styan <callumstyan@gmail.com>
cc @gouthamve @beorn7 

### DIFF
--- a/cmd/thanos/sidecar.go
+++ b/cmd/thanos/sidecar.go
@@ -60,6 +60,8 @@ func registerSidecar(m map[string]setupFunc, app *kingpin.Application) {
 
 	uploadCompacted := cmd.Flag("shipper.upload-compacted", "If true sidecar will try to upload compacted blocks as well. Useful for migration purposes. Works only if compaction is disabled on Prometheus. Do it once and then disable the flag when done.").Default("false").Bool()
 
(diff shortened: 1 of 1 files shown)


cmd/thanos/sidecar.go
```diff
@@ -60,6 +60,8 @@ func registerSidecar(m map[string]setupFunc, app *kingpin.Application) {
 
 	uploadCompacted := cmd.Flag("shipper.upload-compacted", "If true sidecar will try to upload compacted blocks as well. Useful for migration purposes. Works only if compaction is disabled on Prometheus. Do it once and then disable the flag when done.").Default("false").Bool()
 
+	ignoreBlockSize := cmd.Flag("shipper.ignore-unequal-block-size", "If true sidecar will not require prometheus min and max block size flags to be set to the same value. Only use this if you want to keep long retention and compaction enabled on your Prometheus instance, as in the worst case it can result in ~2h data loss for your Thanos bucket storage.").Default("false").Bool()
```
bwplotka: Let's make it `.Hidden()` for now and not put in CHANGELOG (:

cstyan: I also removed it from `docs/components/sidecar.md`

Issue #1774, opened 2019-11-21
Remote read in Store API has a high overhead


nextus:
**Thanos, Prometheus and Golang version used**:

```
thanos, version 0.8.1 (branch: HEAD, revision: bd8278859b2321aaaa7514edde764816cc039d34)
  build user:       root@2227d9a2fdb1
  build date:       20191014-12:03:55
  go version:       go1.13.1
```
```
prometheus, version 2.14.0 (branch: HEAD, revision: edeb7a44cbf745f1d8be4ea6f215e79e651bfe19)
  build user:       root@df2327081015
  build date:       20191111-14:27:12
  go version:       go1.13.4
```
```
/usr/bin/thanos sidecar --log.level=info --log.format=logfmt --grpc-address=0.0.0.0:10901 --http-address=0.0.0.0:10902 --prometheus.url=http://localhost:9090/prometheus --tsdb.path=/opt/prometheus/data --objstore.config-file=/etc/thanos/s3.yaml
/usr/bin/thanos query --log.level=info --log.format=logfmt --grpc-address=0.0.0.0:10903 --http-address=0.0.0.0:10904 --query.partial-response --store=localhost:10901
/opt/prometheus/prometheus --config.file=/etc/prometheus/prometheus.yml --web.external-url=http://0.0.0.0:9090/prometheus/ --storage.tsdb.path=/opt/prometheus/data --storage.tsdb.retention.time=6h --storage.tsdb.max-block-duration=2h --storage.tsdb.min-block-duration=2h --query.timeout=30s --query.max-concurrency=8 --web.enable-admin-api
```

**Object Storage Provider**:
S3

**What happened**:
Response time is 10 times slower using Store API than using Prometheus API endpoint for specific requests.

**What you expected to happen**:
Less overhead in response time using Store API.

**How to reproduce it (as minimally and precisely as possible)**:
Compare response time during querying data with large amount of samples.

**Anything else we need to know**:
I have built a very simple setup to rid of any overhead by network and other components:
`Thanos query --> Thanos sidecar --> Prometheus`
Each of it runs on the same server and  loopback interface is used for IPC.

So there are two queries which are self-explanatory:
```
$ curl -sw '%{time_total} %{size_download}\n' -o /dev/null 'http://localhost:10904/api/v1/series?match[]=http%3Ainternal_request_client%3Aresponses_by_path' 
9.869466 27584266
[...]

PR #1904, opened 2019-12-18
Add skipChunk option in series API

### LINKED ISSUE #1822: Add no-chunks `Store.Series` option to StoreAPI
Querier HTTP `/series` [API](https://prometheus.io/docs/prometheus/latest/querying/api/#finding-series-by-label-matchers) is returning metadata, so series matched by labels.

For Thanos this uses StoreAPI.Series which also returns full chunks for whole requested time period. However the only labelset is required. 

I would propose:
* Extend Series method of StoreAPI with some `no-chunk` bool or enum. This is to let know all implementations to do not send any samples, just labels of series. We might adjust the  response, or just stream series with zero chunks. I think the latter makes sense.
* Sidecar then based on this can decide to use series API of Prometheus instead of remote read (:

yeya24:

## Changes

Fixes #1822 #1774
Add a no-chunk bool in series API to return labels instead of the whole series. Only in querier /api/v1/series API, this option will set to true, otherwise it set to false.

## Verification

I have already added test cases for all stores. 

For the response time using no-chunk, I tested in my local env.

Using Thanos v0.9.0
``` bash
curl -sw '%{time_total} %{size_download}\n' -o /dev/null 'http://localhost:10904/api/v1/series?' -d 'match[]={job="prometheus"}'
0.297961 81680
```

Using this PR
``` bash
curl -sw '%{time_total} %{size_download}\n' -o /dev/null 'http://localhost:10908/api/v1/series?' -d 'match[]={job="prometheus"}'
0.024901 81680
```


CHANGELOG.md
```diff
@@ -18,6 +18,7 @@ We use *breaking* word for marking changes that are not backward compatible (rel
 ### Added
 - [#1852](https://github.com/thanos-io/thanos/pull/1852) Add support for `AWS_CONTAINER_CREDENTIALS_FULL_URI` by upgrading to minio-go v6.0.44
 - [#1854](https://github.com/thanos-io/thanos/pull/1854) Update Rule UI to support alerts count displaying and filtering.
+- [#1904](https://github.com/thanos-io/thanos/pull/1904) Add a no-chunk option in Store Series API.
```
GiedriusS: Perhaps lets mention that this greatly improves performance of the `/api/v1/series` endpoint?

pkg/store/proxy_test.go
```diff
@@ -420,6 +454,9 @@ func TestProxyStore_Series(t *testing.T) {
 
 			s := newStoreSeriesServer(context.Background())
 
+			if tc.req.NoChunk == true {
```
GiedriusS: You probably forgot to remove this? :P

pkg/store/prometheus.go
```diff
@@ -621,3 +649,64 @@ func (p *PrometheusStore) LabelValues(ctx context.Context, r *storepb.LabelValue
...
+	u.Path = path.Join(u.Path, "/api/v1/series")
+
+	q := u.Query()
+	q.Add("match[]", matchersToMetric(matchers))
+	u.RawQuery = q.Encode()
+
+	q.Add("start", string(startTime))
+	q.Add("end", string(endTime))
+
+	req, err := http.NewRequest("GET", u.String(), nil)
```
GiedriusS: You can use `http.NewRequestWithContext` here to shorten things :P

pkg/store/matchers.go
```diff
@@ -33,3 +35,64 @@ func translateMatchers(ms []storepb.LabelMatcher) (res []*labels.Matcher, err er
...
+		return m.Name + `!="` + m.Value + `"`
+
+	case storepb.LabelMatcher_RE:
+		return m.Name + `=~"` + m.Value + `"`
+
+	case storepb.LabelMatcher_NRE:
+		return m.Name + `!~"` + m.Value + `"`
+	}
+
+	return ""
```
GiedriusS: Maybe lets `panic` here so that it would act as an assertion that this for sure cannot be reached?

pkg/store/prometheus.go
GiedriusS: Also... can we use `POST` here to avoid potential character limits with big queries?

pkg/store/matchers.go
```diff
@@ -33,3 +35,64 @@ func translateMatchers(ms []storepb.LabelMatcher) (res []*labels.Matcher, err er
...
+		if i != len(nameMatchers)-1 {
+			n += ", "
+		}
+		res += n
+	}
+
+	return "{" + res + "}"
+}
+
+func labelMatcherToString(m storepb.LabelMatcher) string {
```
GiedriusS: Maybe let's reuse https://github.com/thanos-io/thanos/blob/master/pkg/store/matchers.go#L9 here + https://godoc.org/github.com/prometheus/tsdb/labels#Matcher ?

pkg/store/prometheus.go
yeya24: Oh, make sense! Thanks 😄

pkg/store/matchers.go
yeya24: wow... I didn't even find this is already implemented 😢 Thanks, I will switch to it.

pkg/store/prometheus.go
yeya24: I recheck the POST thing. Looks like https://github.com/prometheus/client_golang/blob/master/api/prometheus/v1/api.go also uses GET.

pkg/store/proxy_test.go
yeya24: Thanks!

pkg/store/matchers.go
yeya24: TBH, I am not sure whether there is a case to not match anything? Since the matchers parameters are passed from HTTP request, so it will validate the user input to ensure its validity

CHANGELOG.md
yeya24: Done

pkg/store/storepb/rpc.proto
```diff
@@ -92,6 +92,8 @@ message SeriesRequest {
 
   // TODO(bwplotka): Move Thanos components to use strategy instead. Including QueryAPI.
   PartialResponseStrategy partial_response_strategy = 7;
+
+  bool no_chunk = 8;
```
bwplotka: This requires a comment.

Overall are we happy with naming and this API field? Wonder how extendible this is. Overall having bool field to RPC is sometimes a [code/design smell](https://martinfowler.com/bliki/FlagArgument.html).

I think I am ok with this here as it's a caller decision to no skip chunks so maybe field should be named:

* `bool skip_chunks` or `bool discard_chunks`? (:

yeya24: I prefer `skip_chunks`. Do we need to make this as an enum? Currently I don't see some possible extensions of this field.

pkg/store/matchers.go
```diff
@@ -33,3 +35,53 @@ func translateMatchers(ms []storepb.LabelMatcher) (res []*labels.Matcher, err er
...
+	)
+
+	matchers, err := translateMatchers(ms)
+	if err != nil {
+		return "", err
+	}
+
+	for _, m := range matchers {
+		if m.Name == "__name__" {
+			nameMatchers = append(nameMatchers, m)
```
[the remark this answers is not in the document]
yeya24: Indeed!

yeya24:
@bwplotka  Thanks for your detailed review. I have addressed all of your comments.

pkg/store/matchers.go
```diff
@@ -33,3 +33,21 @@ func translateMatchers(ms []storepb.LabelMatcher) (res []*labels.Matcher, err er
...
+// matchersToString converts label matchers to string format.
+func matchersToString(ms []storepb.LabelMatcher) (string, error) {
+	var res string
+	matchers, err := translateMatchers(ms)
+	if err != nil {
+		return "", err
+	}
+
+	for i, m := range matchers {
+		res += m.String()
```
bwplotka: Generally, we do `strings.Join` but this is fine for now (:

PR #1905, opened 2019-12-18
add go vet to lint

### LINKED ISSUE #1783: pkg: go vet issues due to Seek signature
Hello! `io.Seeker` is one of the more core Go interfaces that should be followed for any method named `Seek`. For examle, go vet currently returns:

```
$ go vet ./...
# github.com/thanos-io/thanos/pkg/compact/downsample
pkg/compact/downsample/downsample.go:620:34: method Seek(x int64) bool should have signature Seek(int64, int) (int64, error)
pkg/compact/downsample/downsample_test.go:495:27: method Seek(int64) bool should have signature Seek(int64, int) (int64, error)
# github.com/thanos-io/thanos/pkg/query
pkg/query/iter.go:221:26: method Seek(int64) bool should have signature Seek(int64, int) (int64, error)
pkg/query/iter.go:237:34: method Seek(t int64) (ok bool) should have signature Seek(int64, int) (int64, error)
pkg/query/iter.go:288:32: method Seek(t int64) (ok bool) should have signature Seek(int64, int) (int64, error)
pkg/query/iter.go:519:32: method Seek(t int64) bool should have signature Seek(int64, int) (int64, error)
pkg/query/querier_test.go:688:26: method Seek(t int64) bool should have signature Seek(int64, int) (int64, error)
```

I don't know what your API compatibility rules and such are, but I would encourage giving this Seek either a different name or following the io.Seeker interface. Alternatively, you could consider whitelisting this and running vet in CI, but I would encourage complying with it to avoid unexpected surprises.

daixiang0:

## Changes

Fix #1783

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -298,6 +298,8 @@ web: web-pre-process $(HUGO)
 #
 # to debug big allocations during linting.
 lint: check-git $(GOLANGCILINT) $(MISSPELL)
+	@echo ">> examining all of the Go files"
+	@go vet -stdmethods=false
 	@echo ">> linting all of the Go files GOGC=${GOGC}"
 	@$(GOLANGCILINT) run
 	@echo ">> detecting misspells"
--- a/Makefile
+++ b/Makefile
@@ -299,7 +299,7 @@ web: web-pre-process $(HUGO)
 # to debug big allocations during linting.
 lint: check-git $(GOLANGCILINT) $(MISSPELL)
 	@echo ">> examining all of the Go files"
-	@go vet -stdmethods=false
+	@go vet -stdmethods=false ./pkg/... ./cmd/... && go vet doc.go
 	@echo ">> linting all of the Go files GOGC=${GOGC}"
 	@$(GOLANGCILINT) run
 	@echo ">> detecting misspells"


daixiang0:
I do not find docs about how to ignore vendor dir with flag or it would check vendor dir, and i test at local as first commit, all are fine.

daixiang0:
And if i specify all go files using FILES_TO_FMT, it would fail:
```
>> examining all of the Go files
named files must all be in one directory; have ./cmd/thanos/ and ./scripts/cfggen/
```

squat:
That’s because go vet takes a list of packages, not files

daixiang0:
I test a repo with vendor dir existed, no issue reported from vendor dir.

bwplotka:
yes, but we have to exclude `vendor` - we don't want to block dep update only because there is some vet issue inside dep.

bwplotka:
> I test a repo with vendor dir existed, no issue reported from vendor dir.

Not sure if that proves that we are safe though (: I would expect some deps actually having some vet issues, so maybe (:

daixiang0:
For now, we have only one vendor dir `./pkg/ui/static/vendor` where no go files exist.

squat:
the golang vendor directory is not checked into the repo but it is created whenever the binary is compiled. the directory will be expected to be there. in fact it's present in my local checkout of the repo pretty much always :)

daixiang0:
@bwplotka @squat i specify dir to skip vendor.

PR #1908, opened 2019-12-18
Makefile: fix lint and check-docs targets

simonpasquier:
Those errors didn't trigger a CI failure because `quay.io/thanos/thanos-ci` (or rather `circleci/golang`) uses `sh` as the interpreter which doesn't fail on incorrect test expressions.

```
$ if [[ ! ls ezae; ]]; then echo yes; fi
/bin/sh: 14: [[: not found  
/bin/sh: 14: ]]: not found  
$ echo $?
0
$ ps 
  PID TTY          TIME CMD 
    1 pts/0    00:00:00 sh 
 7457 pts/0    00:00:00 ps
$ bash
circleci@47ec4d9adda4:/go/thanos$ if [[ ! ls ezae; ]]; then echo yes; fi
bash: conditional binary operator expected
bash: syntax error near `;'
circleci@47ec4d9adda4:/go/thanos$ echo $?
2
```

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -181,10 +181,7 @@ check-docs: $(EMBEDMD) $(LICHE) build
 	@$(LICHE) --recursive docs --exclude "(cloud.tencent.com|alibabacloud.com)" --document-root .
 	@$(LICHE) --exclude "(cloud.tencent.com|goreportcard.com|alibabacloud.com)" --document-root . *.md
 	@find -type f -name "*.md" | xargs scripts/cleanup-white-noise.sh
-	@if [[ ! git diff-files --quiet --ignore-submodules -- ]]; then \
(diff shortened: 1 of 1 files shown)


squat:
Hmm looks like now it’s working too well. CI doesn’t pass because there are some things that need cleanup it seems

simonpasquier:
Yes working on it.

simonpasquier:
That's fixed.

bwplotka:
Do we have a PR clash with: https://github.com/thanos-io/thanos/pull/1878 ? (:

bwplotka:
Merging this, commented why https://github.com/thanos-io/thanos/pull/1878 

Thanks both!

PR #1914, opened 2019-12-19
Simplify running tests with excluding certain providers.

bwplotka:
Still by default run all. Making it more difficult to skip makes sure we don't forget to run those.

### DIFF
--- a/pkg/objstore/objtesting/foreach.go
+++ b/pkg/objstore/objtesting/foreach.go
@@ -3,8 +3,10 @@ package objtesting
 import (
 	"io/ioutil"
 	"os"
+	"strings"
 	"testing"
 
+	"github.com/thanos-io/thanos/pkg/objstore/client"
 	"github.com/thanos-io/thanos/pkg/objstore/filesystem"
 
 	"github.com/thanos-io/thanos/pkg/objstore"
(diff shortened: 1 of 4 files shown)


pkg/objstore/objtesting/foreach.go
```diff
@@ -18,23 +20,38 @@ import (
 	"github.com/thanos-io/thanos/pkg/testutil"
 )
 
+// IsObjStoreSkipped returns true if given provider ID is found in THANOS_TEST_OBJSTORE_BLACKLIST array delimited by comma e.g:
+// THANOS_TEST_OBJSTORE_BLACKLIST=GCS,S3,AZURE,SWIFT,COS,ALIYUNOSS
+func IsObjStoreSkipped(t *testing.T, provider client.ObjProvider) bool {
+	if e, ok := os.LookupEnv("THANOS_TEST_OBJSTORE_BLACKLIST"); !ok {
```
squat: Naming nit: can we call this instead `THANOS_TEST_OBJSTORE_SKIP`? In general I think we’d like to avoid using blacklist and whitelist

squat:
100% behind the idea of making skipping these tests easier tho

pkg/objstore/objtesting/foreach.go
bwplotka: It's a list though (: Maybe `skip list`?

squat: Sure but in spoken English you could say “please skip tests A B and C” there’s no need for the word list IMO. I think this is also common in Unix tools. Plus, why make the flag name any longer and harder to type :p

bwplotka: fair, done (:

PR #1916, opened 2019-12-20
Makefile: use gsed on OSX to keep GNU sed compatibility

pracucci:
OSX's `sed` has the `-i` flag which is not compatible with GNU `sed`, causing commands like `make docs` to fail on OSX. A popular option for OSX users is to install `gsed` via `brew`.

In this PR I'm suggesting to:

1. Use `gsed` when available, falling back to `sed`
2. Always use `find .`  to get it working on OSX too

These changes should be safe for Linux users.

## Changes

- Makefile: use gsed on OSX to keep GNU sed compatibility

## Verification

Manual tests.


pracucci:
The failing test looks unrelated, and due to a temporary network failure:
```
docs/governance.md
	ERROR	https://couchdb.apache.org/bylaws.html#lazy
		Dialing to the given TCP address timed out
```

squat:
Saw this yesterday too. Will restart the tests

PR #1923, opened 2019-12-23
Pin quay.io/prometheus/busybox docker image

### LINKED ISSUE #911: Pin quay.io/prometheus/busybox docker image
We cannot do port-forward for Thanos images recently due to:

```
Handling connection for 43709
Handling connection for 43709
E0312 13:56:25.445092   19402 portforward.go:329] an error occurred forwarding 43709 -> 19191: error forwarding port 19191 to pod simstoregw-0_infra, uid : unable to do port forwarding: socat not found.
E0312 13:56:55.018961   19402 portforward.go:269] error creating error stream for port 43709 -> 19191: Timeout occured
```

This is because we pin to latest `FROM quay.io/prometheus/busybox:latest`. We should aim on pinning to something that has `socat` and allows us to do port forwading

AC: 
* Find base image with `sockat` and proper tooling (e.g net tools with `curl` `git` and `dig` is must have)
* Pin to certain version

Non-goals:
* Making sure image is super small, that's not a requirement. But smaller = better.

daixiang0:

## Changes

Fix #911

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,4 +1,5 @@
-FROM quay.io/prometheus/busybox:latest
+ARG SHA="0c38f63cbe19e40123668a48c36466ef72b195e723cbfcbe01e9657a5f14cec6"
+FROM quay.io/prometheus/busybox@sha256:${SHA}
 LABEL maintainer="The Thanos Authors"
 
 COPY /thanos_tmp_for_docker /bin/thanos
--- a/Dockerfile.multi-stage
+++ b/Dockerfile.multi-stage
@@ -14,7 +14,8 @@ RUN git update-index --refresh; make build
 
 # -----------------------------------------------------------------------------
 
-FROM quay.io/prometheus/busybox:latest
+ARG SHA="0c38f63cbe19e40123668a48c36466ef72b195e723cbfcbe01e9657a5f14cec6"
+FROM quay.io/prometheus/busybox@sha256:${SHA}
 LABEL maintainer="The Thanos Authors"
 
 COPY --from=builder /go/bin/thanos /bin/thanos
--- a/Makefile
+++ b/Makefile
@@ -5,6 +5,20 @@ DOCKER_IMAGE_REPO ?= quay.io/thanos/thanos
 DOCKER_IMAGE_TAG  ?= $(subst /,-,$(shell git rev-parse --abbrev-ref HEAD))-$(shell date +%Y-%m-%d)-$(shell git rev-parse --short HEAD)
 DOCKER_CI_TAG     ?= test
 
+SHA=''
+arch = $(shell uname -m)
+# Run `DOCKER_CLI_EXPERIMENTAL=enabled docker manifest inspect quay.io/prometheus/busybox:latest` to get SHA
+# Update at 2020.07.06
+ifeq ($(arch), x86_64)
+    # amd64
+    SHA="248b7ec76e03e6b4fbb796fc3cdd2f91dad45546a6d7dee61c322475e0e8a08f"
+else ifeq ($(arch), armv8)
+    # arm64
+    SHA="69508e8fdc516eacbacc0379c03c971e3043706cc8211e6bddb35d903edc3628"
+else
+    echo >&2 "only support amd64 or arm64 arch" && exit 1
+endif
+
 # Ensure everything works even if GOPATH is not set, which is often the case.
 # The `go env GOPATH` will work for all cases for Go 1.8+.
 GOPATH            ?= $(shell go env GOPATH)
@@ -132,7 +146,7 @@ docker: build
 	@echo ">> copying Thanos from $(PREFIX) to ./thanos_tmp_for_docker"
 	@cp $(PREFIX)/thanos ./thanos_tmp_for_docker
 	@echo ">> building docker image 'thanos'"
-	@docker build -t "thanos" .
+	@docker build -t "thanos" --build-arg SHA=$(SHA) .
 	@rm ./thanos_tmp_for_docker
 else
 docker: docker-multi-stage
@@ -142,7 +156,7 @@ endif
 docker-multi-stage: ## Builds 'thanos' docker image using multi-stage.
 docker-multi-stage:
 	@echo ">> building docker image 'thanos' with Dockerfile.multi-stage"
-	@docker build -f Dockerfile.multi-stage -t "thanos" .
+	@docker build -f Dockerfile.multi-stage -t "thanos" --build-arg SHA=$(SHA) .
 
 .PHONY: docker-push
 docker-push: ## Pushes 'thanos' docker image build to "$(DOCKER_IMAGE_REPO):$(DOCKER_IMAGE_TAG)".


Dockerfile
```diff
@@ -1,4 +1,5 @@
-FROM quay.io/prometheus/busybox:latest
+# pin busybox version
```
bwplotka: No need for this. Maybe it would be rather interesting to comment on data of which this image is from (:

daixiang0: comment https://quay.io/repository/prometheus/busybox?tab=tags?

bwplotka: Still, I can't really see that certain sha256 there :thinking:

daixiang0: As commit said, you can see sha256 in https://quay.io/repository/prometheus/busybox?tab=history, i use the latest tag at Nov 13, 2019.

bwplotka: https://quay.io/repository/prometheus/busybox/manifest/sha256:598bed4d1a87712e9d5e5863c0458c20161362d4f41233ee262e44ae69f21b33

Manifest not found ):

daixiang0:
In fact, all sha256 in web page are invalid. Use below cmd to get true sha256:
```
[root@test ~]# docker manifest inspect quay.io/prometheus/busybox:latest
{
   "schemaVersion": 2,
   "mediaType": "application/vnd.docker.distribution.manifest.list.v2+json",
   "manifests": [
      {
         "mediaType": "application/vnd.docker.distribution.manifest.v2+json",
         "size": 737,
         "digest": "sha256:0c38f63cbe19e40123668a48c36466ef72b195e723cbfcbe01e9657a5f14cec6",
         "platform": {
            "architecture": "amd64",
            "os": "linux"
         }
      },
      {
         "mediaType": "application/vnd.docker.distribution.manifest.v2+json",
         "size": 737,
         "digest": "sha256:3ab46d9f23bc36c511f3206ebdd40bfb5a63663acc0f3fc02ca24f5dbfb762a2",
         "platform": {
            "architecture": "arm64",
            "os": "linux"
         }
      },
      {
         "mediaType": "application/vnd.docker.distribution.manifest.v2+json",
         "size": 737,
         "digest": "sha256:d887a41d091ffeff30962e1dc5f61fc0cc672ac9432086fef4775a1a48c24a75",
         "platform": {
            "architecture": "arm",
            "os": "linux"
         }
      },
      {
         "mediaType": "application/vnd.docker.distribution.manifest.v2+json",
         "size": 738,
         "digest": "sha256:359fd3f29ee5518256f47f8e84008ad03e4e934c34e2fb8dead380022f8ce092",
         "platform": {
            "architecture": "ppc64le",
            "os": "linux"
         }
      },
      {
         "mediaType": "application/vnd.docker.distribution.manifest.v2+json",
         "size": 738,
         "digest": "sha256:5dfaf3f5106471b7179f88277becd66e2444f59bbb6ba2461a188eef090c91b6",
         "platform": {
            "architecture": "s390x",
            "os": "linux"
         }
      }
   ]
}```

**Comment by daixiang0** [comment]
ping @bwplotka

**Comment by squat** [comment]
Pinning to a SHA like this will break building this container on non-amd64 machines :/ . I don't think we should apply this change.

**Comment by daixiang0** [comment]
Since there is no tag for all arch, we need to specify each SHA for each arch we support, then auto check arch.

File: Dockerfile.multi-stage [hunk 2]
```diff
@@ -9,7 +9,8 @@ RUN git update-index --refresh; make build
 
 # -----------------------------------------------------------------------------
 
-FROM quay.io/prometheus/busybox:latest
+# sha256 refer to https://quay.io/repository/prometheus/busybox?tab=tags
+FROM quay.io/prometheus/busybox@sha256:0c38f63cbe19e40123668a48c36466ef72b195e723cbfcbe01e9657a5f14cec6
```
squat: We should probably also parameterize this dockerfile

daixiang0:
@bwplotka mind have a look?

daixiang0:
any input for this?

Makefile
```diff
@@ -5,6 +5,19 @@ DOCKER_IMAGE_REPO ?= quay.io/thanos/thanos
...
 
+SHA=''
+arch = $(shell uname -m)
+# Run `docker manifest inspect quay.io/prometheus/busybox:latest` to get SHA
+ifeq ($(arch), x86_64)
+    # amd64
+    SHA="0c38f63cbe19e40123668a48c36466ef72b195e723cbfcbe01e9657a5f14cec6"
+else ifeq ($(arch), armv8)
+	# arm64
+    SHA="3ab46d9f23bc36c511f3206ebdd40bfb5a63663acc0f3fc02ca24f5dbfb762a2"
```
squat: I think these may be a little cumbersome to maintain and error prone

squat: Consider something more automatic like

```make
SHA=$(shell docker manifest inspect $(BASE_IMAGE) | jq '.manifests[] | select(.platform.architecture == "$(arch)'") | if .platform | has("variant") then select(.platform.variant == "'$(variant)") else . end | .digest' -r)
```

daixiang0: It is hard to do this, please look at https://quay.io/repository/prometheus/busybox?tag=latest&tab=tags.
Busybox has no tag expect latest or SHA.
We want to pin the busybox version, we can not use latest, the SHA is the only choice.
Since SHA is diffienert for each arch, hardcode here maybe look ugly but fit this case.

daixiang0: We can comment here about how to get the SHA.

squat: The code I shared with you automatically determines the correct sha for the selected arch. The problem is that simply hardcoding a sha value will definitely get out of date and be difficult to maintain. The wrong sha *will* be chosen at one point for an arch even if we write the steps down. Generating it automatically is the right thing to do IMO

daixiang0: I know your code works, but the `BASE_IMAGE` does not exist indeed.
In above [comment](https://github.com/thanos-io/thanos/pull/1923#issuecomment-568635587), i use busybox:latest as `BASE_IMAGE`.
As i said, busybox has no tags expect `latest` or SHA, and SHA is diffienert for each arch. When busybox update, your code would update all SHA which does not meet our needs(We do not always want to update SHA when busybox update, do we?).

daixiang0: I think generating it automatically is helpful when we confirm that we need to update busybox version. To use a fixed busybox version, hardcode is necessary.

daixiang0:
hold

daixiang0:
@squat @bwplotka any input?

daixiang0:
update

daixiang0:
ping @bwplotka

squat:
looks like we had a flake

daixiang0:
@bwplotka any chance to merge it?

daixiang0:
ping @thanos-io/thanos-maintainers

daixiang0:
@bwplotka any chance to merge it?

daixiang0:
@squat @bwplotka any input?

bwplotka (approved):
Super nice! Thank you :+1:

PR #1924, opened 2019-12-23
use go module to download pkgs

daixiang0:

## Changes

Optimize download logic.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -86,7 +86,7 @@ define fetch_go_bin_version
 	@mkdir -p $(TMP_GOPATH)
 
 	@echo ">> fetching $(1)@$(2) revision/version"
-	@GOPATH=$(TMP_GOPATH) GO111MODULE=on go get -u $(1)@$(2)
+	@GOBIN='$(TMP_GOPATH)/bin' GOPATH=$(TMP_GOPATH) GO111MODULE=on go get -u $(1)@$(2)
 	@mv -- '$(TMP_GOPATH)/bin/$(shell basename $(1))' '$(GOBIN)/$(shell basename $(1))-$(2)'
 	@echo ">> produced $(GOBIN)/$(shell basename $(1))-$(2)"
 endef
--- a/Makefile
+++ b/Makefile
@@ -81,13 +81,12 @@ MINIO_SERVER            ?=$(GOBIN)/minio-$(MINIO_SERVER_VERSION)
 # arguments:
 # $(1): Install path. (e.g github.com/campoy/embedmd)
 # $(2): Tag or revision for checkout.
-# cd TMP_GOPATH to ensure do not use or edit thanos go.mod file.
 define fetch_go_bin_version
 	@mkdir -p $(GOBIN)
 	@mkdir -p $(TMP_GOPATH)
 
 	@echo ">> fetching $(1)@$(2) revision/version"
-	@cd -- $(TMP_GOPATH) && GOPATH=$(TMP_GOPATH) GO111MODULE=on go get -u $(1)@$(2)
+	@GOPATH=$(TMP_GOPATH) GO111MODULE=on go get -u $(1)@$(2)
 	@mv -- '$(TMP_GOPATH)/bin/$(shell basename $(1))' '$(GOBIN)/$(shell basename $(1))-$(2)'
 	@echo ">> produced $(GOBIN)/$(shell basename $(1))-$(2)"
 endef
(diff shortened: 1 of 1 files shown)


daixiang0:
On amd64 centos, with go version go1.13.4 linux/amd64.

daixiang0:
ping @bwplotka

daixiang0:
ping @bwplotka

kakkoyun:
@daixiang0 @bwplotka I believe rather than just flipping `GO111MODULE` we should introduce a `tools.go` and track our tool dependencies in our modules.  As it's described in https://github.com/golang/go/wiki/Modules#how-can-i-track-tool-dependencies-for-a-module and shown in https://github.com/go-modules-by-example/index/blob/master/010_tools/README.md

With this approach, we can simplify the Makefile significantly. We can get rid of the custom scripts., and versioning of the tools in Makefile.

I'm happy to help with it. I can address it in another PR or it could be handled in this one by @daixiang0 

WDYT @bwplotka?

kakkoyun:
cc @squat

Makefile
```diff
@@ -81,23 +81,14 @@ MINIO_SERVER            ?=$(GOBIN)/minio-$(MINIO_SERVER_VERSION)
...
-	@if [ ! -d '$(TMP_GOPATH)/src/$(1)' ]; then \
-    GOPATH='$(TMP_GOPATH)' GO111MODULE='off' go get -d -u '$(1)/...'; \
-  else \
-    CDPATH='' cd -- '$(TMP_GOPATH)/src/$(1)' && git fetch; \
-  fi
-	@CDPATH='' cd -- '$(TMP_GOPATH)/src/$(1)' && git checkout -f -q '$(2)'
-	@echo ">> installing $(1)@$(2)"
-	@GOBIN='$(TMP_GOPATH)/bin' GOPATH='$(TMP_GOPATH)' GO111MODULE='off' go install '$(1)'
+	@GOPATH=$(TMP_GOPATH) GO111MODULE=on go get -u $(1)@$(2)
 	@mv -- '$(TMP_GOPATH)/bin/$(shell basename $(1))' '$(GOBIN)/$(shell basename $(1))-$(2)'
```
bwplotka: We need to export GOBIN to be TMP_GOPATH/bin in order for this to work.

Essentially it has to work if someone has `GOBIN` specified.

daixiang0:
@bwplotka mind have a look?

daixiang0:
any input for this?

kakkoyun:
@daixiang0, I think this doesn't suffice the Acceptance criteria as @bwplotka told. You can try the `tools.go` with build flags approach that I've suggested above. However, keep in mind that it won't completely solve the problem here. Because it will download or vendor the tool dependencies and it will use go.{mod|sum} to track those dependencies, even though it won't try to build those binaries unless you explicitly say so. 

See https://github.com/golang/go/wiki/Modules#how-can-i-track-tool-dependencies-for-a-module

One thing you could try to create a separate package and `go module` for the tools, and use build flags to ignore them for regular operations. And you can try to build those explicitly when needed in development time or CI.

I haven't tried this to be honest, but in theory, it should work :)

daixiang0:
@kakkoyun I think the easiest way is to `git checkout go.{mod|sum}`, since it is in a function.
I add a help info in case users change those files and not committed.

kakkoyun:
@daixiang0 I think there's a cleaner way with go `1.14` release. Direct quote from the [release notes](https://golang.org/doc/go1.14):
> -modfile=file is a new flag that instructs the go command to read (and possibly write) an alternate go.mod file instead of the one in the module root directory. A file named go.mod must still be present in order to determine the module root directory, but it is not accessed. When -modfile is specified, an alternate go.sum file is also used: its path is derived from the -modfile flag by trimming the .mod extension and appending .sum

My vote on waiting for the upgrade and create a dedicated directory and go.mod file for the external tools.

daixiang0:
hold

daixiang0:
@bwplotka i see you add samiliar func in #2233, does it mean we support those 2 ways?

kakkoyun:
@daixiang0 An example implementation to use a separate dedicated `go.mod` file to use for binary dependencies https://github.com/coreos/prometheus-operator/pull/3176/files.

Also, @bwplotka cooks something aside to fix this particular problem https://github.com/bwplotka/gobin. So If we want to go down this path we should close this one.

bwplotka:
Yes, I am on it. Will propose a PR soon @daixiang0 if that's ok

bwplotka:
Oh yea I added @daixiang0 XD 

but it is workaround still. But definitely it's the same as this you wrote. Sorry for missing this ): Thanks for your work, but we plan to try https://github.com/bwplotka/gobin (project is not fully ready yet!)

daixiang0:
Would close this as it is finished.

bwplotka:
Yup! done (:

PR #1933, opened 2020-01-02
Add a wal compression flag in rule and receive

yeya24:

## Changes

Add a WAL compression option in rule and receive

### DIFF
--- a/cmd/thanos/receive.go
+++ b/cmd/thanos/receive.go
@@ -13,7 +13,6 @@ import (
 	opentracing "github.com/opentracing/opentracing-go"
 	"github.com/pkg/errors"
 	"github.com/prometheus/client_golang/prometheus"
-	"github.com/prometheus/common/model"
 	"github.com/prometheus/prometheus/pkg/labels"
 	"github.com/prometheus/prometheus/storage/tsdb"
 	"github.com/thanos-io/thanos/pkg/block/metadata"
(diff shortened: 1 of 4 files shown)


yeya24:
Should it be a flag? Receiver sets this option to `true` by default. Is there any use case that people want to set it to false?

bwplotka:
It is an option in Prometheus, plus false by default. (:

yeya24:
OK. I will add the flag for ruler and receiver.

yeya24:
I have added a flag for rule and receive. Should I add a changelog?

squat:
Definitely worth mentioning in changelog if the default has changed

brancz:
WAL compression is intended to become the default in Prometheus, we just can't do it in Prometheus yet as that would be too breaking of a change. I personally think we should just always have WAL compression enabled, the CPU footprint of receive is so tiny it's the right trade-off in my opinion.

tl;dr I think we should leave things as they are in this regard.

bwplotka:
Should we make it true by default then?

brancz:
I'm not sure this should be configurable at all, but yes if this config option exists then it should definitely be true by default.

cmd/thanos/receive.go
```diff
@@ -73,6 +72,8 @@ func registerReceive(m map[string]setupFunc, app *kingpin.Application) {
 
 	tsdbBlockDuration := modelDuration(cmd.Flag("tsdb.block-duration", "Duration for local TSDB blocks").Default("2h").Hidden())
 
+	walCompression := cmd.Flag("tsdb.wal-compression", "Compress the tsdb WAL.").Default("false").Bool()
```
bwplotka: Let's have true by default then

yeya24: Thanks, done.

bwplotka: We thank you (:

PR #1934, opened 2020-01-03
Added block.MetaFetcher logic for resilient sync of meta files.

bwplotka:
This is meant to replace least 5 inconsistent meta.json syncs places across components.

Usage of those will be done in follow up chained PRs.

In the first commit: Transactional gauge allowing to have atomic gauge across different labels.

### DIFF
--- a/pkg/block/fetcher.go
+++ b/pkg/block/fetcher.go
@@ -191,7 +191,7 @@ func (s *MetaFetcher) loadMeta(ctx context.Context, id ulid.ULID) (*metadata.Met
 	r, err := s.bkt.Get(ctx, metaFile)
 	if s.bkt.IsObjNotFoundErr(err) {
 		// Meta.json was deleted between bkt.Exists and here.
-		return nil, errors.Wrapf(ErrorSyncMetaNotFound, "%v ", err)
+		return nil, errors.Wrapf(ErrorSyncMetaNotFound, "%v", err)
 	}
 	if err != nil {
 		return nil, errors.Wrapf(err, "get meta file: %v", metaFile)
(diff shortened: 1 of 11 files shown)


pkg/block/fetcher.go
```diff
@@ -0,0 +1,410 @@
...
+	defer runutil.CloseWithLogOnErr(s.logger, r, "close bkt meta get")
+
+	metaContent, err := ioutil.ReadAll(r)
+	if err != nil {
+		return nil, errors.Wrapf(err, "read meta file: %v", metaFile)
+	}
+
+	m := &metadata.Meta{}
+	if err := json.Unmarshal(metaContent, m); err != nil {
+		return nil, errors.Wrapf(ErrorSyncMetaCorrupted, "meta.json %v unmarshal: %v ", metaFile, err)
```
GiedriusS: Small nit: that space at the end seems unnecessary. The same happened on some other lines as well.

pkg/block/fetcher.go
```diff
@@ -0,0 +1,410 @@
...
+
+	corruptedMeta = "corrupted-meta-json"
+	noMeta        = "no-meta-json"
+	loadedMeta    = "loaded"
+	failedMeta    = "failed"
+
+	// Filter's label values.
+	labelExcludedMeta = "label-excluded"
+	timeExcludedMeta  = "time-excluded"
+	TooFreshMeta      = "too-fresh"
```
GiedriusS: Does this need to be exposed?

pkg/block/fetcher.go
```diff
@@ -0,0 +1,410 @@
...
+	for id, m := range metas {
+		if m.MaxTime >= f.minTime.PrometheusTimestamp() && m.MinTime <= f.maxTime.PrometheusTimestamp() {
+			continue
+		}
+		synced.WithLabelValues(timeExcludedMeta).Inc()
+		delete(metas, id)
+	}
+}
+
+var _ MetaFetcherFilter = (&LabelShardedMetaFilter{}).Filter
```
GiedriusS: Also lets not forget to remove this and the other one. I presume some linter was complaining 😄

bwplotka: Why removing this? (: This is ensuring Filter meets interface

pkg/block/fetcher.go
bwplotka: Yes - used in https://github.com/thanos-io/thanos/pull/1937 (:

pkg/block/fetcher.go
GiedriusS: Ah, right, that's ok, I guess :+1:

PR #1939, opened 2020-01-06
*: support TLS and authentication for Thanos Ruler queries

### LINKED ISSUE #1778: Ruler: Add authentication options query
version:0.8.1

        http://user:password@xxxx.com/api/v1/query?dedup=true&partial_response=false&query=probe_failed_due_to_http_status_code+%21%3D+0&time=2019-11-22T14%3A37%3A10.387097931%2B08%3A00
        level=error ts=2019-11-22T06:37:10.53514552Z caller=rule.go:775 err="unmarshal query instant response: invalid character '<' looking for beginning of value" 

I copy the url to postman and it returns a json result. it seems the ruler can not perform a basic auth request

simonpasquier:
Closes #1778

## Changes

Similar to #1838, this PR adds TLS and authentication support to Thanos Ruler for the query API endpoints. The YAML configuration format is very similar to the one used for configuring Alertmanager.

## Verification

I have adapted the end-to-end tests to use the new parameters. The tests already exercised static addresses and file SD.


cmd/thanos/rule.go
```diff
@@ -256,10 +268,56 @@ func runRule(
...
+		dns.ResolverType(dnsSDResolver),
+	)
+	var queryClients []*http_util.FanoutClient
+	for _, cfg := range queryCfg {
+		c, err := http_util.NewClient(cfg.HTTPClientConfig, "query")
+		if err != nil {
+			return err
 		}
+		c.Transport = tracing.HTTPTripperware(logger, c.Transport)
+		queryClient, err := http_util.NewFanoutClient(logger, cfg.EndpointsConfig, c, queryProvider.Clone())
```
bwplotka: Fanout? I think the logic for rule evaluation might be different. It's fanout = broadcast approach.

cmd/thanos/rule.go
```diff
@@ -322,13 +361,21 @@ func runRule(
...
 	)
+	var alertmgrs []*alert.Alertmanager
 	for _, cfg := range alertingCfg.Alertmanagers {
+		c, err := http_util.NewClient(cfg.HTTPClientConfig, "alertmanager")
+		if err != nil {
+			return err
+		}
 		// Each Alertmanager client has a different list of targets thus each needs its own DNS provider.
-		am, err := alert.NewAlertmanager(logger, cfg, amProvider.Clone())
+		amClient, err := http_util.NewFanoutClient(logger, cfg.EndpointsConfig, c, amProvider.Clone())
```
bwplotka: We can add tracing Tripperware as well I think

cmd/thanos/rule.go
```diff
@@ -440,38 +461,11 @@ func runRule(
...
-				}
-			}
-		}, func(error) {
-			cancelUpdate()
-			close(fileSDUpdates)
-		})
+	// Discover and resolve query addresses.
+	{
+		for _, c := range queryClients {
+			addToGroup(g, c, dnsSDInterval)
```
bwplotka: Not consistent with alertmanager `addToGroup` placement. Let's decide on one approach if we can (: 

Reason: I spent some time trying to understand if we missed `addToGroup` for query or not (:

cmd/thanos/rule.go
bwplotka: Yea, the logic is ok, just the Name of client might be confusing. It's not Fanout for Querier.

cmd/thanos/rule.go
```diff
@@ -99,6 +97,8 @@ func registerRule(m map[string]setupFunc, app *kingpin.Application) {
 	queries := cmd.Flag("query", "Addresses of statically configured query API servers (repeatable). The scheme may be prefixed with 'dns+' or 'dnssrv+' to detect query API servers through respective DNS lookups.").
 		PlaceHolder("<query>").Strings()
 
+	queryConfig := extflag.RegisterPathOrContent(cmd, "query.config", "YAML file that contains query API servers configuration. See format details: https://thanos.io/components/rule.md/#configuration. If defined, it takes precedence over the '--query' and '--query.sd-files' flags.", false)
+
 	fileSDFiles := cmd.Flag("query.sd-files", "Path to file that contain addresses of query peers. The path can be a glob pattern (repeatable).").
```
bwplotka:
```suggestion
	fileSDFiles := cmd.Flag("query.sd-files", "Path to file that contains addresses of query API servers. The path can be a glob pattern (repeatable).").
```

cmd/thanos/rule.go
```diff
@@ -738,36 +719,49 @@ func queryFunc(
...
 
-			if err != nil {
-				level.Error(logger).Log("err", err, "query", q)
-			} else {
-				if len(warns) > 0 {
-					ruleEvalWarnings.WithLabelValues(strings.ToLower(partialResponseStrategy.String())).Inc()
-					// TODO(bwplotka): Propagate those to UI, probably requires changing rule manager code ):
-					level.Warn(logger).Log("warnings", strings.Join(warns, ", "), "query", q)
+				if err != nil {
+					level.Error(logger).Log("err", err, "query", q)
```
bwplotka: let's do `continue` instead of `else`  - bit more readable

cmd/thanos/rule.go
```diff
@@ -738,36 +719,49 @@ func queryFunc(
...
+						ruleEvalWarnings.WithLabelValues(strings.ToLower(partialResponseStrategy.String())).Inc()
+						// TODO(bwplotka): Propagate those to UI, probably requires changing rule manager code ):
+						level.Warn(logger).Log("warnings", strings.Join(warns, ", "), "query", q)
+					}
+					return v, nil
 				}
-				return v, nil
 			}
 		}
 		return nil, errors.Errorf("no query peer reachable")
```
bwplotka:
```suggestion
		return nil, errors.Errorf("no query API server reachable")
```

The `peer` name is obsolete: Came from gossip times. (:

cmd/thanos/rule.go
```diff
@@ -738,36 +719,49 @@ func queryFunc(
...
 	}
 }
+
+func addToGroup(g *run.Group, c *http_util.FanoutClient, interval time.Duration) {
```
bwplotka: It's not only add, it adds for DNS & file discovery, so maybe: 

```suggestion
func addDiscoveryGroups(g *run.Group, c *http_util.FanoutClient, interval time.Duration) {
```

cmd/thanos/rule.go
```diff
@@ -738,36 +719,49 @@ func queryFunc(
 	}
 
 	return func(ctx context.Context, q string, t time.Time) (promql.Vector, error) {
```
bwplotka: I hope we can proper timeout set on caller top this (:

pkg/http/http.go
```diff
@@ -114,3 +124,138 @@ func (u userAgentRoundTripper) RoundTrip(r *http.Request) (*http.Response, error
...
+	return fileSDConfig, nil
+}
+
+type AddressProvider interface {
+	Resolve(context.Context, []string)
+	Addresses() []string
+}
+
+// FanoutClient represents a client that can send requests to a cluster of HTTP-based endpoints.
+type FanoutClient struct {
```
bwplotka: Again, fanout might be bad wording here, why not `just` Client?

cmd/thanos/rule.go
```diff
@@ -738,36 +719,49 @@ func queryFunc(
...
-
-			span, ctx := tracing.StartSpan(ctx, spanID)
-			v, warns, err := promclient.PromqlQueryInstant(ctx, logger, u, q, t, promclient.QueryOptions{
-				Deduplicate:             true,
-				PartialResponseStrategy: partialResponseStrategy,
-			})
-			span.Finish()
+		for _, i := range rand.Perm(len(queriers)) {
+			querier := queriers[i]
+			c := promclient.NewClient(logger, querier)
```
bwplotka: Why not creating this at the start of ruler? (:

pkg/http/http.go
```diff
@@ -114,3 +124,138 @@ func (u userAgentRoundTripper) RoundTrip(r *http.Request) (*http.Response, error
...
+		prefix:          cfg.PathPrefix,
+		staticAddresses: cfg.StaticAddresses,
+		fileSDCache:     cache.New(),
+		fileDiscoverers: discoverers,
+		provider:        provider,
+	}, nil
+}
+
+// Do executes an HTTP request with the underlying HTTP client.
+func (f *Client) Do(req *http.Request) (*http.Response, error) {
```
bwplotka: nit: `f` , probably should be `c` ;p

test/e2e/rule_test.go
```diff
@@ -83,6 +103,15 @@ func serializeAlertingConfiguration(t *testing.T, cfg ...alert.AlertmanagerConfi
 	return b
 }
 
+func serializeQueryConfiguration(t *testing.T, cfg ...query.Config) []byte {
+	t.Helper()
+	b, err := yaml.Marshal(&cfg)
+	if err != nil {
```
bwplotka: `testutil.Ok` is preferred. Additionally `Errorf` only marks the test as failed but continues the flow, which is wrong here (will cause panics)

pkg/http/http.go
simonpasquier: nice spotting!

test/e2e/rule_test.go
simonpasquier: done

bwplotka:
Rebase needed ):

Issue #1935, opened 2020-01-03
Please upgrade to 2.15.1 for Prometheus dependency of go Mod


liguozhong:
github.com/prometheus/prometheus v1.8.2-0.20191126064551-80ba03c67da1 // Prometheus master v2.14.0

Prometheus master v2.15.1

bwplotka:
Yup, that's the plan. We are waiting on 2.15.2 release though, fixing compaction on Windows: https://github.com/prometheus/prometheus/pull/6547

bwplotka:
Also help wanted (:

daixiang0:
@bwplotka happy to take this.

bwplotka:
I depend on this for https://github.com/thanos-io/thanos/pull/1839 so I will take it if you don't mind (:

daixiang0:
That is fine :)

bwplotka:
But again, let's wait on upstream fix that we discussed on Prometheus IRC

liguozhong:
![image](https://user-images.githubusercontent.com/9583245/71874594-4a826300-315d-11ea-9223-b0d4b984dc12.png)
2.15.2 release ，thanks @bwplotka

Issue #1928, opened 2019-12-27
Compactor OOM


acarsercan:
Hi,

- Thanos v0.8.1
- banzai thanos helm chart - https://github.com/banzaicloud/banzai-charts/blob/master/thanos/templates/compact-deployment.yaml
- S3 bucket 200GB

I been using thanos behind a prometheus-operator deployment for over a month. During holiday period (out of hours) memory usage has skyrocketed. I'm still unable to pin down what is causing it.

Compactor pod will OOM every 15-20 minutes, going beyond the 8GB limit

I'm going to guess there is a setting or retention which I just hit.

Any ideas?

Log level debug
```

 level=debug ts=2019-12-27T16:32:34.090393606Z caller=main.go:101 msg="maxprocs: Updating GOMAXPROCS=[2]: determined from CPU quota"
 level=info ts=2019-12-27T16:32:34.090586411Z caller=main.go:149 msg="Tracing will be disabled"
 level=info ts=2019-12-27T16:32:34.090762472Z caller=factory.go:43 msg="loading bucket configuration"
 level=info ts=2019-12-27T16:32:34.105826606Z caller=compact.go:270 msg="retention policy of raw samples is enabled" duration=720h0m0s
 level=info ts=2019-12-27T16:32:34.105858745Z caller=compact.go:273 msg="retention policy of 5 min aggregated samples is enabled" duration=2880h0m0s
 level=info ts=2019-12-27T16:32:34.105869173Z caller=compact.go:276 msg="retention policy of 1 hour aggregated samples is enabled" duration=8760h0m0s
 level=info ts=2019-12-27T16:32:34.105882534Z caller=compact.go:361 msg="starting compact node"
 level=info ts=2019-12-27T16:32:34.105895111Z caller=prober.go:107 msg="changing probe status" status=ready
 level=info ts=2019-12-27T16:32:34.105967171Z caller=prober.go:127 msg="changing probe status" status=healthy
 level=info ts=2019-12-27T16:32:34.10600956Z caller=http.go:53 service=http/server component=compact msg="listening for requests and metrics" address=0.0.0.0:10902
 level=info ts=2019-12-27T16:32:34.109593018Z caller=compact.go:1083 msg="start sync of metas"
 level=debug ts=2019-12-27T16:32:34.175253071Z caller=compact.go:288 msg="download meta" block=01DTR7TX8GGMXM1Z3WQ2JKHC3J
[...]

Issue #1771, opened 2019-11-21
Thanos compactor stuck on downsampling?


Jonnymcc:
**Thanos, Prometheus and Golang version used**: image: quay.io/thanos/thanos:v0.8.0

**Object Storage Provider**: S3

**What happened**: Compactor started downsampling and appeared to make no progress. It eventually hit the memory limit (40GB) for the pod in Kubernetes and is terminated.

**What you expected to happen**: A log message saying that a block has been downsampled. My understanding is that each block would be downloaded, downsampled, and then uploaded back to the bucket.

**How to reproduce it (as minimally and precisely as possible)**: Not sure.

**Full logs to relevant components**:

```
level=info ts=2019-11-21T00:19:33.796596397Z caller=compact.go:1069 msg="start of GC"
level=info ts=2019-11-21T00:19:33.797001158Z caller=compact.go:1075 msg="start of compaction"
level=info ts=2019-11-21T00:19:33.831505526Z caller=compact.go:264 msg="compaction iterations done"
level=info ts=2019-11-21T00:19:33.831529668Z caller=compact.go:271 msg="start first pass of downsampling"
level=info ts=2019-11-21T00:28:48.542147196Z caller=downsample.go:257 msg="downloaded block" id=01DRGCN74V2S71A5VAR5GS9ERX duration=9m14.046528213s
```
This is where the log ends.

**Anything else we need to know**:
Bucket total size > 1TB
Number of objects in bucket ~ 10.5k
Total size of 01DRGCN74V2S71A5VAR5GS9ERX in bucket is 40.2GB

bwplotka:
It's OOM and while we optimized the downsampling process a lot, there still has to be more memory than biggest block.

Issue #722, opened 2019-01-10
Compact RAM calculation


arnitolog:
I have one of the blocks which is ~7GB of size. And Compact can't do anything with it. It is OOM killed every time. Is there any way to instruct Compact to process big blocks piecemeal
Here is part of meta.json file:
	"version": 1,
	"ulid": "01CRYM9RYN78DFKZ73XB4H4WCP",
	"minTime": 1537401600000,
	"maxTime": 1538611200000,
	"stats": {
		"numSamples": 4489155614,
		"numSeries": 2088751,
		"numChunks": 39194144
	},
here is graphs for compact pod:
![image](https://user-images.githubusercontent.com/34036079/50997191-97bcfc00-152c-11e9-826e-47ba28ea0d9a.png)
here is an error:
level=info ts=2019-01-10T20:46:34.747142839Z caller=compact.go:822 msg="start of GC"
level=info ts=2019-01-10T20:46:34.78464068Z caller=compact.go:207 msg="compaction iterations done"
level=info ts=2019-01-10T20:46:34.784681757Z caller=compact.go:214 msg="start first pass of downsampling"
level=info ts=2019-01-10T20:48:27.547013191Z caller=downsample.go:212 msg="downloaded block" id=01CRYM9RYN78DFKZ73XB4H4WCP duration=1m52.526675454s
fatal error: runtime: out of memory

GiedriusS:
Compactor now supports sharding so you could use that as a temporary workaround but we could definitely provide some guidance on how much RAM is needed since it seems like this is a recurring topic: https://github.com/thanos-io/thanos/issues/1928

PR #1947, opened 2020-01-07
Updated Prometheus deps to master around 2.15.2 tag.

bwplotka:
Fixes: https://github.com/thanos-io/thanos/issues/1935
Fixes: https://github.com/thanos-io/thanos/issues/1928
Fixes: https://github.com/thanos-io/thanos/issues/722
Fixes: https://github.com/thanos-io/thanos/issues/1771

Needed for index-header but mainly this makes compaction and downsampling process taking constant memory! It will take more time though as a tradeoff.

Now it should take at max:
* ~30th of all symbols*2 and number of postings per loaded block into memory.
* ~Largest series with its chunks and all label names during compaction.

Help in verifying this on a bigger setup would be nice! (: 


bwplotka:
Did you go through this PR @povilasv in details ? ;p 

I made same mistake again (did similar in previous also approved PRs):

> `- []() Upgraded Prometheus dependencies to v2.15.2. This includes:` 

It's non-obvious PR so want to be careful here. (:

povilasv:
I did 2x times and IMO it's ok :+1: 

> Did you go through this PR @povilasv in details ? ;p
> 
I did 2x times and IMO it's ok :+1: but obviously needs some review more familiar with the code base :)

pkg/block/indexheader/json_reader.go
```diff
@@ -56,6 +58,50 @@ func (b realByteSlice) Sub(start, end int) index.ByteSlice {
...
+// The table gets initialized with sync.Once but may still cause a race
+// with any other use of the crc32 package anywhere. Thus we initialize it
+// before.
+var castagnoliTable *crc32.Table
+
+func init() {
+	castagnoliTable = crc32.MakeTable(crc32.Castagnoli)
+}
+
+// ReadSymbols reads the symbol table fully into memory and allocates proper strings for them.
```
squat: nit: func name mismatch

bwplotka:
@povilasv double checking (: Thanks for review!

Issue #1873, opened 2019-12-11
index.cache.json files way bigger than common indexes


vlantier:
**Thanos, Prometheus and Golang version used**:
Thanos: 0.9.0
Go: 1.13.1 
Prometheus: 2.14

**Object Storage Provider**: GCS

**What happened**:
Our `index.cache.json` files for a couple of days are being created greater than the normal index files. This is expected?

Comparing with 5 days agor we have index files of 1G and 300/400MB `cache.json`. No we have index with 1.8G and caches with 1.7.

I think that the big files are responsaible for the actual overload in `thanos-store` that we have, because only when we configure `min-size` and `max-size` for the correspondent days that we have the bigger cache they crash.

The compactor is running without errors.

**What you expected to happen**:
The previous behavior: Cache files way small than index files

**Anything else we need to know**:

bwplotka:
Yes, the reason behind that is that index.cache.json is unoptimized e.g strings are not interned. With huge number of labels and long strings this might be the case, but your case looks really extreme. 

We are working on this to resolve your issue: https://github.com/thanos-io/thanos/pull/1839

vlantier:
I see.
Thanks for the responde @bwplotka, I'll be watching this issue..

With the new Prometheus 2.14 we saw isome `Highest Cardinality Metric Names` and `Highest Cardinality Labels` who may impact in this matter. Those metrics are from home-made exporters and we are working on it.

But something is odd because those metrics are there for a long time already. May you have some tips about where to look?

Issue #1750, opened 2019-11-16
Thanos store OOM


salapat11:
**Thanos, Prometheus and Golang version used**:
Thanos: 0.6.0
Prometheus: 2.10.0

**Object Storage Provider**: S3

**What happened**:
Thanos Store won't start.  It runs for 2 mins and crashes with OOM.  Increased the memory to 64 GB and will still fail.  Compactor is running and generating index.cache.json files.  

Bucket size: 202.4 GiB.  Total Objects: 4952
Biggest index.cache.json: 3.2 GiB

- --index-cache-size=20GB
- --chunk-pool-size=40GB

GiedriusS:
Hi, any reason you are running such an old version? Please try the `master` and/or `0.8.1` and see if it is still reproducible.

salapat11:
Same issue with v0.8.1.

hbokh:
Indeed. Using the same version here.

```
# thanos --version
thanos, version 0.8.1 (branch: HEAD, revision: bd8278859b2321aaaa7514edde764816cc039d34)
  build user:       root@2227d9a2fdb1
  build date:       20191014-12:03:55
  go version:       go1.13.1
```

Total objects is a little over 8000.

Running on a VM, I had to go from 2GB --> 4GB --> 8GB --> 16GB of memory before the OOM-killer was not an issue anymore!
Now `thanos store` is using 12.3GB of RAM.

MarcMielke:
Same issue with 0.9.0. I wonder what the relations between `Bucket size`, `Total Objects`,`--index-cache-size` and `--chunk-pool-size` are - as to how to come up with a formula indicating the proper memory requirements. Even if it's only an estimate.

draeron:
> Same issue with 0.9.0. I wonder what the relations between `Bucket size`, `Total Objects`,`--index-cache-size` and `--chunk-pool-size` are - as to how to come up with a formula indicating the proper memory requirements. Even if it's only an estimate.

we're having the same issue, it'd be really useful if the documentation would contains some informations about ball park values to set.

hbokh:
Things seem to have gotten a bit worse with v0.9.0...
I upgraded at 14:13CET and this is what Grafana shows:

![Screenshot 2019-12-24 at 15 32 58](https://user-images.githubusercontent.com/519955/71416856-b3001900-2662-11ea-86c0-3770193b7591.png)

[...]

Issue #1711, opened 2019-11-04
Improve composing index-header by pulling only required pieces of index.


bwplotka:
Orthogonal to [sparse file proposal](https://github.com/thanos-io/thanos/issues/1471) we may need to explore composing index-cache/index-header from parts of the index.

We do not compose the block only when:
* No index-cache.json is in the local filesystem
* No index-cache.json is in object storage.

Currently, we create index-cache by mmaping whole index file which might cause excessive resource usage, especially if we syncronize many blocks at the same time without locally cached objects. 

The proposal is to do pull pieces of the index. To do so we need TOC which is at the end of the file as well as other pieces. This might mean changes in object store API: https://github.com/thanos-io/thanos/pull/1567/files#diff-c4e16588aa4d5a179f37f9a55308eb7eR245

If the process will be quick enough to do per request, we might move out of startup towards request time for this, improving startup time.

bwplotka:
This is orthogonal to: https://github.com/thanos-io/thanos/issues/1471

bwplotka:
This is well suitable to be designed and done together with: https://github.com/thanos-io/thanos/issues/942

PR #1952, opened 2020-01-07
Added binary index header implementation.

bwplotka:
This PR adds index-header implementation based on [this design](https://thanos.io/proposals/201912_thanos_binary_index_header.md/). 

It adds a separate `indexheader.Binary*` structs and methods allowing to build and read index-header in binary format.

Fixes: https://github.com/thanos-io/thanos/issues/1711
Fixes: https://github.com/thanos-io/thanos/issues/1873
Fixes: https://github.com/thanos-io/thanos/issues/1750
Fixes: https://github.com/thanos-io/thanos/issues/1655
Fixes: https://github.com/thanos-io/thanos/issues/1471
Fixes: https://github.com/thanos-io/thanos/issues/1455
Fixes: https://github.com/thanos-io/thanos/issues/942
Fixes: https://github.com/thanos-io/thanos/issues/448
Fixes: https://github.com/thanos-io/thanos/issues/1750

cc @brian-brazil @codesome as you are super familiar with the TSDB format.
cc @pracucci 

## Stats

Size difference:

10k series for my autogenerated block:

```
-rw-r--r-- 1 bwplotka bwplotka 6.1M Jan 10 13:20 index
-rw-r--r-- 1 bwplotka bwplotka  23K Jan 10 13:20 index.cache.json
-rw-r--r-- 1 bwplotka bwplotka 9.2K Jan 10 13:20 index-header
```

For production block 8mln series, also similar gain.

```
-rw-r--r-- 1 bwplotka bwplotka 1.9G Jan 10 13:29 index
-rw-r--r-- 1 bwplotka bwplotka 287M Jan 10 13:29 index.cache.json
-rw-r--r-- 1 bwplotka bwplotka 122M Jan 10 13:29 index-header
```

NOTE: Size is smaller, but it's not what we are trying to optimize for. Nevertheless
PostingOffsets and Symbols takes a significant amount of bytes. The only downsides of size
is the fact that to create such index-header we have to fetch those two parts ~60MB each from object storage.
Idea for another improvement at some point would if it will become a problem: Cache only 32th of the posting ranges and fetch gaps between on demand
on query time (with some cache).

Real-time latencies for creation and loading (without network traffic):

For 10k block it's similar for both (ms/micros), for 8mln we can spot the difference:

index-header:

* write 134.197732ms
* read 415.971774ms

index-cache.json:

* write 6.712496338s
* read 6.112222132s

## Go Benchmarks:

Before comparing I changed names to correlate tests with `benchcmp`:

```
BenchmarkJSONReader-12-> BenchmarkRead-12 old
BenchmarkBinaryReader-12 -> BenchmarkRead-12 new
BenchmarkJSONWrite-12 -> BenchmarkWrite-12 old
BenchmarkBinaryWrite-12  -> BenchmarkWrite-12 new
```

### 10k series block:

```
benchmark             old ns/op     new ns/op     delta
BenchmarkRead-12      591780        66613         -88.74%
BenchmarkWrite-12     2458454       6532651       +165.72%

benchmark             old allocs     new allocs     delta
BenchmarkRead-12      2306           629            -72.72%
BenchmarkWrite-12     1995           64             -96.79%

benchmark             old bytes     new bytes     delta
BenchmarkRead-12      150904        32976         -78.15%
BenchmarkWrite-12     161501        73412         -54.54%
```

CPU time for smaller index file is interesting. Value is low anyway. Might be
something to follow up.

### 8mln series (index takes 2GB so not committed to git):

```
benchmark             old ns/op      new ns/op     delta
BenchmarkRead-12      7026290474     552913402     -92.13%
BenchmarkWrite-12     6480769814     276441977     -95.73%

benchmark             old allocs     new allocs     delta
BenchmarkRead-12      20100014       5501312        -72.63%
BenchmarkWrite-12     18263356       64             -100.00%

benchmark             old bytes      new bytes     delta
BenchmarkRead-12      1873789526     406021516     -78.33%
BenchmarkWrite-12     2385193317     74187         -100.00%
```

NOTE: Benchmarks for query time and CHANGELOG will go in next PR.


pkg/block/indexheader/header_test.go
```diff
@@ -44,8 +72,29 @@ func TestReaders(t *testing.T) {
...
+	testutil.Ok(t, err)
+	defer func() { testutil.Ok(t, os.RemoveAll(tmpDir)) }()
+
+	// Copy block index version 1 for backward compatibility.
+	m, err := metadata.Read("./testdata/index_format_v1")
+	testutil.Ok(t, err)
+	testutil.Copy(t, filepath.Join(tmpDir, m.ULID.String()), "./testdata/index_format_v1")
+
+	t.Run("binary", func(t *testing.T) {
+	})
```
bwplotka: in progress (:

pkg/block/indexheader/header.go
```diff
@@ -1,19 +1,35 @@
...
 	// TODO(bwplotka): Move to PostingsOffsets(name string, value ...string) []index.Range and benchmark.
-	PostingsOffset(name string, value string) index.Range
+	PostingsOffset(name string, value string) (index.Range, error)
+
+	// LookupSymbol returns string based on given reference.
+	// Error is return if the symbol can't be found.
 	LookupSymbol(o uint32) (string, error)
-	LabelValues(name string) []string
+
+	// LabelValues returns all label values for given label name or error if not found.
```
GiedriusS: The comment here seems a bit wrong: we return an error when an actual error had occurred, not when we haven't found any label values.

pkg/block/indexheader/binary_reader.go
```diff
@@ -0,0 +1,802 @@
...
+// polynomial may be easily changed in one location at a later time, if necessary.
+func newCRC32() hash.Hash32 {
+	return crc32.New(castagnoliTable)
+}
+
+// BinaryTOC is a table of content for index-header file.
+type BinaryTOC struct {
+	// Symbols holds start to the same symbols section as index related to this index header.
+	Symbols uint64
+	// PostingsTable holds start to the the same Postings Offset Table section as index related to this index header.
```
GiedriusS: The comment should start here with `PostingsOffsetTable`.

pkg/block/indexheader/binary_reader.go
```diff
@@ -0,0 +1,802 @@
...
+		r.postingsV1 = map[string]map[string]index.Range{}
+
+		var prevRng index.Range
+		if err := index.ReadOffsetTable(r.b, r.toc.PostingsOffsetTable, func(key []string, off uint64, _ int) error {
+			if len(key) != 2 {
+				return errors.Errorf("unexpected key length for posting table %d", len(key))
+			}
+			// TODO(bwplotka): This is wrong, probably we have to sort.
+			if lastKey != nil {
+				prevRng.End = int64(off + 4)
```
GiedriusS: In a lot of places we add `4` to different offsets. I guess that's because of how big crc32 is. I think the code's readability could be improved if we would move that constant `4` into a variable. `crc32.Size` could probably be reused.

pkg/block/indexheader/binary_reader.go
```diff
@@ -0,0 +1,802 @@
...
+		// Earlier V1 formats don't have a sorted postings offset table, so
+		// load the whole offset table into memory.
+		r.postingsV1 = map[string]map[string]index.Range{}
+
+		var prevRng index.Range
+		if err := index.ReadOffsetTable(r.b, r.toc.PostingsOffsetTable, func(key []string, off uint64, _ int) error {
+			if len(key) != 2 {
+				return errors.Errorf("unexpected key length for posting table %d", len(key))
+			}
+			// TODO(bwplotka): This is wrong, probably we have to sort.
```
GiedriusS: So should we sort this or not? :smile:

pkg/block/indexheader/header_test.go
```diff
@@ -20,53 +31,362 @@ func TestReaders(t *testing.T) {
...
 
-func testReader(t *testing.T, r Reader) {
-	testutil.Equals(t, 2, r.IndexVersion())
-	exp := []string{"1", "2", "3", "4", "a", "b"}
-	for i := range exp {
-		r, err := r.LookupSymbol(uint32(i))
+func compareIndexToHeader(t *testing.T, indexByteSlice index.ByteSlice, headerReader Reader) {
+	indexReader, err := index.NewReader(indexByteSlice)
+	testutil.Ok(t, err)
+	defer func() { _ = indexReader.Close() }()
```
GiedriusS: Shouldn't we check that this succeeds?

pkg/block/indexheader/binary_reader.go
bwplotka: done, but it's not crc, but size of length of posting.

pkg/block/indexheader/binary_reader.go
bwplotka: not ;p

pkg/block/indexheader/header_test.go
bwplotka: no, because it's a Prometheus indexReader and we are not testing that one.

pkg/store/bucket.go
```diff
@@ -1514,6 +1548,65 @@ func (r *bucketIndexReader) fetchPostings(groups []*postingGroup) error {
...
 }
 
+func resizePostings(b []byte) ([]byte, error) {
+	d := encoding.Decbuf{B: b}
+	n := d.Be32int()
+	if d.Err() != nil {
+		return nil, errors.Wrap(d.Err(), "read postings list")
+	}
+	// 4 for posting length, then n * 4, foreach each big endian posting.
+	return b[:4+n*4], nil
```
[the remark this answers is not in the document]
bwplotka: nice catch!

pkg/store/bucket.go
```diff
@@ -1492,20 +1526,20 @@ func (r *bucketIndexReader) fetchPostings(groups []*postingGroup) error {
 			r.stats.postingsFetchedSizeSum += int(length)
 
 			for _, p := range ptrs[i:j] {
-				c := b[p.ptr.Start-start : p.ptr.End-start]
-
-				_, fetchedPostings, err := r.dec.Postings(c)
+				// index-header can estimate endings, which means we need to resize the endings.
```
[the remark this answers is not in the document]
bwplotka: We can't due to arbitrary padding that can occur - it's fine though, overfetching is usually small.

We can only exactly resize once we know number of entries from fetched posting.

pkg/block/indexheader/binary_reader.go
```diff
@@ -0,0 +1,802 @@
...
+
+// NewBinaryReader loads or builds new index-header if not present on disk.
+func NewBinaryReader(ctx context.Context, logger log.Logger, bkt objstore.BucketReader, dir string, id ulid.ULID) (*BinaryReader, error) {
+	binfn := filepath.Join(dir, id.String(), block.IndexHeaderFilename)
+	br, err := newFileBinaryReader(binfn)
+	if err == nil {
+		return br, nil
+	}
+
+	level.Warn(logger).Log("msg", "failed to read index-header from disk; recreating", "path", binfn, "err", err)
```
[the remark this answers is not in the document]
bwplotka: :+1:  Definitely! I would say even debug.

pkg/block/indexheader/binary_reader.go
```diff
@@ -0,0 +1,802 @@
...
+
+	if err := WriteBinary(ctx, bkt, id, binfn); err != nil {
```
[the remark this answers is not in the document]
bwplotka: hm.. let's start with elapsed time in log line WDYT?

pkg/block/indexheader/binary_reader.go
```diff
@@ -0,0 +1,802 @@
...
+				}
+				valueIndex++
+				if valueIndex == len(values) {
+					break
+				}
+				value = values[valueIndex]
+			}
+			if i+1 == len(e) {
+				for i := range tmpRngs {
+					tmpRngs[i].End = r.indexLastPostingEnd
```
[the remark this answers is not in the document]
bwplotka: Wow :heart: You just found a major bug here! Nice - I fixed it and made tests more strict! Awesome!

bwplotka:
Thanks for reviews! All should be addressed.

@GiedriusS: 

> The cherry on top of the pie would be to add some kind of flag to Thanos Compact which would migrate all of the existing JSON files into the binary ones and improve the binary reader to try to download them from remote object storage. But I guess that could be added when we will move this from the experimental stage.

I would say, remove all index-cache JSON file as we build all on-demand on store Gateway (: 

@pracucci kudos for finding quite a major bug, I think the Series() micro-benchmark would find this.. if we would have any (: TODO for next PR to iterate on index-header.

bwplotka:
Should be rdy to go, so we can iterate on it later on :+1:

It is hidden behind an experimental flag for now. 

 @squat @brancz

pkg/block/indexheader/binary_reader.go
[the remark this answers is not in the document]
pracucci: Agree

pkg/block/indexheader/binary_reader.go
```diff
@@ -0,0 +1,817 @@
...
+	}
+
+	level.Debug(logger).Log("msg", "failed to read index-header from disk; recreating", "path", binfn, "err", err)
+
+	start := time.Now()
+	if err := WriteBinary(ctx, bkt, id, binfn); err != nil {
+		return nil, errors.Wrap(err, "write index header")
+	}
+
+	level.Debug(logger).Log("msg", "built index-header file", "path", binfn, "elapsed", time.Since(start))
```
pracucci: Given we don't have any metric whenever an `index-header` is created, if there's a bug which re-creates the `index-header` for a block each time (because the reader detects it as corrupted) it would be very difficult to notice, unless you run Thanos with debug logging. Switching this log to `Info` may help.

bwplotka: Right. In this case we need a metric (:

bwplotka: Will add that in next PR. (:

farvour:
This feature really helped. Much better than mmapping the really large index json files that were being generated by our large prometheus cluster. Combined with the memcached index caching I can finally lower the memory requirements significantly in our clusters for store gateway.

Hope to see more improvements.

bwplotka:
Btw did you run memcached as well? Any feedback on that part?

der-eismann:
Just to give some feedback: The memory consumption of our store gateway sank from 28,9 GB to 27,2 GB, which IMHO isn't such a huge improvement. But we still have no idea how the `index-cache-size` and `chunk-pool-size` options affect this. We don't use memcached yet.
However the CPU usage is now significantly lower, reduced from 0,5 core spikes to 0,1 core spikes.

PR #1955, opened 2020-01-08
sidecar: allow custom http con pool size fix #1953

### LINKED ISSUE #1953: sidecar: increase MaxIdleConnsPerHost or allow it to be configured
The thanos sidecar is mounted to a single Prometheus instance where it establishes an HTTP connection for queries against that machine. It's using the `http.DefaultTransport` when connecting to that client which has `MaxIdleConnsPerHost` set to 2 by default [see DefaultMaxIdleConnsPerHost](https://golang.org/pkg/net/http/). We run a large, heavily used cluster and we're receiving about 1.7k queries per second from some machines. Because of this load we are constantly over the connection pool limit so new connections are being established and destroyed. This causes our machines to run low on ports since each connection sits in a TIME_WAIT state for 60s, consuming a port. 

Since the connection pool size depends entirely on the setup I think there should be a flag to allow us to run sidecar with custom connection pool size. Perhaps `--prometheus.connection-pool-size`?

Here are some examples showing the TIME_WAIT connections being created in excess in the second example: https://tleyden.github.io/blog/2016/11/21/tuning-the-go-http-client-library-for-load-testing/

This graph shows HTTP requests between sidecar and Prometheus on our machines. 
image

blockloop:

## Changes

Allow custom connection pool size for Thanos sidecar to prevent excessive tcp connections on machines with very high load. More details are laid out in #1953 

## Verification

A build from my fork is currently running in production on our machines and handling ~1700 RPS with very little connection cycling.

```
$ netstat -a | awk '{ print $6 }' | grep -P '[A-Z]' | sort | uniq -c | sort -r
    951 ESTABLISHED
    166 TIME_WAIT
    146 CONNECTED
    133 CLOSE_WAIT
    129 FIN_WAIT2
     53 STREAM
     12 LISTEN
      4 FIN_WAIT1
      1 SEQPACKET
      1 I-Node
      1 Foreign
```

## Other Info

HTTP connections

image

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -51,6 +51,7 @@ Compactor now properly handles partial block uploads for all operation like rete
 - [#1833](https://github.com/thanos-io/thanos/pull/1833) `--shipper.upload-compacted` flag has been promoted to non hidden, non experimental state. More info available [here](docs/quick-tutorial.md#uploading-old-metrics).
 - [#1867](https://github.com/thanos-io/thanos/pull/1867) Ruler: now sets a `Thanos/$version` `User-Agent` in requests
 - [#1887](https://github.com/thanos-io/thanos/pull/1887) Service discovery now deduplicates targets between different target groups
+- [#1933](https://github.com/thanos-io/thanos/pull/1933) Add a flag `--tsdb.wal-compression` to configure whether to enable tsdb wal compression in ruler and receiver.
 
 ## [v0.9.0](https://github.com/thanos-io/thanos/releases/tag/v0.9.0) - 2019.12.03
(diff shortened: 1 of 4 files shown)


blockloop:
Hey @bwplotka I'll make these changes and push another PR. This was from the master branch of my fork

bwplotka:
Sure, hope you can address my comments there (:

blockloop:
Yep will do. I already separated the two flags in the PR I made to receive. I was going to do that here also. I'll send up another PR shortly.

blockloop:
reopened as #1969

Issue #1710, opened 2019-11-03
receive: create a gRPC remote_write endpoint


blockloop:
I understand the compatibility reasons for creating a remote_write endpoint that adheres to the Prometheus standard. Would you consider enabling a gRPC endpoint to enable pushing metrics from arbitrary gRPC sources? Ideally it would use a gRPC Stream to improve performance.

brancz:
I think I would want something like this to be experimental, but if we can prove it to have a significant improvement over the existing one, then I'd even go a step further and attempt to make Prometheus send remote write via gRPC.

blockloop:
I've started work on this here: https://github.com/blockloop/thanos/tree/bjones/receive/use-grpc

PR #1966, opened 2020-01-08
receive: allow custom connection pooling

blockloop:

## Changes

Receive: enable custom connection pool sizing. With the default configuration machines with very high load (over 500 requests per second) will cycle TCP connections too frequently and run out of available ports. See more [here](https://tleyden.github.io/blog/2016/11/21/tuning-the-go-http-client-library-for-load-testing/)

## Verification

Running a custom build in production. Also see #1955

### DIFF
--- a/cmd/thanos/receive.go
+++ b/cmd/thanos/receive.go
@@ -18,6 +18,7 @@ import (
 	"github.com/thanos-io/thanos/pkg/block/metadata"
 	"github.com/thanos-io/thanos/pkg/component"
 	"github.com/thanos-io/thanos/pkg/extflag"
+	"github.com/thanos-io/thanos/pkg/exthttp"
 	"github.com/thanos-io/thanos/pkg/objstore/client"
 	"github.com/thanos-io/thanos/pkg/prober"
 	"github.com/thanos-io/thanos/pkg/receive"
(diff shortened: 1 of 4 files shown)


cmd/thanos/receive.go
```diff
@@ -71,7 +71,8 @@ func registerReceive(m map[string]setupFunc, app *kingpin.Application) {
 
 	replicationFactor := cmd.Flag("receive.replication-factor", "How many times to replicate incoming write requests.").Default("1").Uint64()
 
-	tsdbBlockDuration := modelDuration(cmd.Flag("tsdb.block-duration", "Duration for local TSDB blocks").Default("2h").Hidden())
```
squat: ah wait, this should still be here

blockloop: yeah I added something from another commit on accident :stuck_out_tongue: 
fix coming in a sec :smile:

pkg/receive/handler.go
```diff
@@ -76,6 +78,9 @@ func NewHandler(logger log.Logger, o *Options) *Handler {
 
 	transport := http.DefaultTransport.(*http.Transport)
```
bwplotka: Same comment applies here: https://github.com/thanos-io/thanos/pull/1955#discussion_r364338651

blockloop:
I think the build needs to be kicked @bwplotka

brancz:
Somewhat related, I was thinking we should switch the internal communications to use gRPC, which could eventually evolve to also be the externally accepting write request API. I’m mentioning this as that would probably make this obsolete. What do you think?

blockloop:
I agree. I think this should be gRPC streams all the way through. I have a branch on my fork with a lot of that work complete. https://github.com/blockloop/thanos/tree/bjones/grpc-remote-write

squat:
Happy to review a PR switching the internal forwarding to gRPC. I say
switching rather than adding as there is no use for two internal forwarding
mechanisms.

blockloop:
@squat, I think that makes sense. Right now since forwarding and receiving are the same code that would require a bit of a change. I can change gears a bit and start on splitting away forwarding from receiving HTTP remote writes. That might be the better option. Ideally the gRPC endpoint would also accept remote writes externally as well.

squat:
Let me know if you want to pair. I think it should be quite straight
forward to essentially keep the same forward func and simply have two
handlers (one gRPC and one HTTP) that both call the same forwarding func.

blockloop:
Yeah I think a RemoteWrite method needs to be added to `store.TSDBStore` though, right?

blockloop:
Maybe we should shift the conversation to https://github.com/thanos-io/thanos/issues/1710 and close this PR?

squat:
Yeah it’ll have to be a new service that has the store methods as well as the RemoteWrite method

blockloop:
Closing this in favor of #1970

PR #1970, opened 2020-01-08
receive: use gRPC for forwarding messages

### LINKED ISSUE #1710: receive: create a gRPC remote_write endpoint

blockloop:

## Changes

Receive: use gRPC for forwarding messages between receive peers to significantly increase throughput. This also introduces the ability for users to use gRPC to write messages to Thanos via gRPC.

## Verified

A working build can be run with this docker-compose configuration https://gist.github.com/blockloop/637bd0c8c9295178b67812f43e661419

The images used in that gist have been built from this branch. I have run it and verified that everything is working. 

Here is a screenshot of Thanos query from that setup:

image

Fixes #1710


pkg/receive/writable_store.go
```diff
@@ -0,0 +1,40 @@
...
+
+// NewReadWriteStore creates a tsdb store that's also writable.
+func NewReadWriteStore(r *store.TSDBStore, w *Writer) *ReadWriteStore {
+	return &ReadWriteStore{
+		TSDBStore: r,
+		w:         w,
+	}
+}
+
+func (rw *ReadWriteStore) RemoteWrite(ctx context.Context, r *storepb.WriteRequest) (*storepb.WriteResponse, error) {
```
squat: Right now if anything is received on the gRPC service it will _always_ be written to the local disk without checking if it should be forwarded. Once we choose to have clients directly remote write gRPC to a receive endpoint, this will no longer work for us.

pkg/store/storepb/rpc.proto
```diff
@@ -39,6 +40,21 @@ service Store {
...
+service WriteableStore {
+  // WriteRequest allows you to write metrics to this store via remote write
+  rpc RemoteWrite(WriteRequest) returns (WriteResponse) {}
+}
+
+message WriteResponse {
+}
+
+message WriteRequest {
+  string tenant = 1;
```
squat: tenant and replica are currently not used at all by the RemoteWrite rpc, as it doesn't check and forward further. Are they placeholders or is the plan to use them?

pkg/receive/writable_store.go
blockloop: Yeah I need to do some more changes to account for that. I’ll continue that part tomorrow

pkg/store/storepb/rpc.proto
blockloop: Yep they’ll be used to forward requests. I gotta continue that part tomorrow 🙂

pkg/store/storepb/rpc.proto
```diff
@@ -39,6 +40,21 @@ service Store {
...
+  int64 replica = 2;
```
brancz: I'm curious how you think we should handle the zero value of this. I feel like we may need to either work with non-zero index or figure some other strategy out.

squat: Ack, with http, the zero value was inferred by checking if the header was present or not. Changing it to index-1 would be simple however, we should then ensure that the http path’s header also becomes 1-indexed to eliminate confusion.

pkg/server/grpc/grpc.go
```diff
@@ -127,3 +127,16 @@ func (s *Server) Shutdown(err error) {
 		cancel()
 	}
 }
+
+// ReadWriteStoreServer is a StoreServer and a WriteableStoreServer.
+type ReadWriteStoreServer interface {
+	storepb.StoreServer
+	storepb.WriteableStoreServer
```
squat: I am wondering again about exposing these on the same gRPC service. If we do, we have to make sure to document that exposing the receive gRPC to a WAN to enable writes means that one tenant could read any other tenant’s metrics

pkg/store/storepb/rpc.proto
blockloop: Is the replica header 0 indexed? I think 1 indexed would be the better option given the zero value situation.

squat:
It is today. We should change both (gRPC and http) to 1-indexed

pkg/receive/handler.go
```diff
@@ -377,25 +385,27 @@ func (h *Handler) parallelizeRequests(ctx context.Context, tenant string, replic
...
+
 			// Create a span to track the request made to another receive node.
 			span, ctx := tracing.StartSpan(ctx, "thanos_receive_forward")
 			defer span.Finish()
 
-			// Actually make the request against the endpoint
-			// we determined should handle these time series.
-			var res *http.Response
-			res, err = h.client.Do(req.WithContext(ctx))
+			_, err = cl.RemoteWrite(ctx, &storepb.WriteRequest{
```
squat: Let's keep this comment here

pkg/receive/handler.go
```diff
@@ -479,5 +504,47 @@ func isConflict(err error) bool {
...
+type peerGroup struct {
+	dialOpts []grpc.DialOption
+	cache    map[string]storepb.WriteableStoreClient
+	m        sync.Mutex
+
+	// dialer is used for testing.
+	dialer func(ctx context.Context, target string, opts ...grpc.DialOption) (conn *grpc.ClientConn, err error)
+}
+
+func (p *peerGroup) Get(ctx context.Context, addr string) (storepb.WriteableStoreClient, error) {
```
squat: can we keep this unexported?

pkg/receive/handler.go
```diff
@@ -479,5 +504,47 @@ func isConflict(err error) bool {
...
+		cache:    map[string]storepb.WriteableStoreClient{},
+		m:        sync.Mutex{},
+		dialer:   grpc.DialContext,
+	}
+}
+
+type peerGroup struct {
+	dialOpts []grpc.DialOption
+	cache    map[string]storepb.WriteableStoreClient
+	m        sync.Mutex
```
squat: if we are making thousands of requests a second, then pretty soon every node in the hashring will have already messaged every other node. Since we won't be writing so often to the cache, I think we  could benefit greatly from using a RW mutex here

pkg/store/storepb/rpc.proto
```diff
@@ -39,6 +40,21 @@ service Store {
...
+  repeated prometheus.TimeSeries timeseries = 3 [(gogoproto.nullable) = false];
```
squat: how about putting this field first so that clients, e.g. a modified prometheus, that only know about the prometheus.TimeSeries field can send to us and we interpret tenant as `''` and replica as `0`?

pkg/receive/handler.go
blockloop: Yeah we can change it. The struct isn't exported so it doesn't change much though.

pkg/receive/handler.go
squat: I think this is missing the first part of the comment

blockloop: Fixed

pkg/server/grpc/grpc.go
GiedriusS: I guess that's acceptable for now 🤔

blockloop:
Rebase complete @GiedriusS

CHANGELOG.md
```diff
@@ -14,6 +14,7 @@ We use *breaking* word for marking changes that are not backward compatible (rel
 - [#1969](https://github.com/thanos-io/thanos/pull/1969) Sidecar: allow setting http connection pool size via flags
 - [#1967](https://github.com/thanos-io/thanos/issues/1967) Receive: Allow local TSDB compaction
 - [#1975](https://github.com/thanos-io/thanos/pull/1975) Store Gateway: fixed panic caused by memcached servers selector when there's 1 memcached node
+- [#1970](https://github.com/thanos-io/thanos/issues/1970) Receive: Use gRPC for forwarding requests between peers
```
brancz: I think we need an item here that indicates that people need to change their hashring files to be just addresses that the gRPC client is built from, instead of the full http endpoint.

Issue #1974, opened 2020-01-09
store: crashes with new memcached config


kamilhristov:
**Thanos version used**: v0.10.0-rc.0

**Object Storage Provider**: AWS S3

**What happened**: Thanos Store crashes when trying to query any data.

**Arguments**:

`thanos store --index-cache.config-file=/app/cache-config.yaml --chunk-pool-size=6GB --objstore.config-file=/app/objstore-config.yaml`

**Config**:

```
type: MEMCACHED
config:
  addresses: ["*********.cache.amazonaws.com:11211"]
```

**Logs**:
```
level=info ts=2020-01-09T16:58:44.91368086Z caller=factory.go:43 msg="loading bucket configuration"
level=info ts=2020-01-09T16:58:45.096521711Z caller=factory.go:30 msg="loading index cache configuration"
level=info ts=2020-01-09T16:58:45.114047201Z caller=memcached.go:54 msg="created memcached index cache"
level=info ts=2020-01-09T16:58:45.114407301Z caller=options.go:20 protocol=gRPC msg="disabled TLS, key and cert must be set to enable"
level=info ts=2020-01-09T16:58:45.114663961Z caller=store.go:288 msg="starting store node"
level=info ts=2020-01-09T16:58:45.114779469Z caller=store.go:243 msg="initializing bucket store"
level=info ts=2020-01-09T16:58:45.115289288Z caller=prober.go:127 msg="changing probe status" status=healthy
level=info ts=2020-01-09T16:58:45.115307464Z caller=http.go:53 service=http/server component=store msg="listening for requests and metrics" address=0.0.0.0:10902
level=info ts=2020-01-09T16:58:45.312703838Z caller=fetcher.go:361 component=block.MetaFetcher msg="successfully fetched block metadata" duration=197.897143ms cached=40 returned=40 partial=0
level=info ts=2020-01-09T16:59:37.571359888Z caller=store.go:249 msg="bucket store ready" init_duration=52.456553581s
level=info ts=2020-01-09T16:59:37.571749739Z caller=grpc.go:98 service=gRPC/server component=store msg="listening for StoreAPI gRPC" address=0.0.0.0:10901
level=info ts=2020-01-09T16:59:37.571618464Z caller=prober.go:107 msg="changing probe status" status=ready
[...]

PR #1975, opened 2020-01-09
Fixed crash in the memcached servers selector when there's only 1 server

pracucci:
Fixes https://github.com/thanos-io/thanos/issues/1974

## Verification

Unit tests.

### DIFF
--- a/pkg/cacheutil/memcached_server_selector_test.go
+++ b/pkg/cacheutil/memcached_server_selector_test.go
@@ -76,7 +76,7 @@ func TestMemcachedJumpHashSelector_PickServer(t *testing.T) {
 		actualAddr, err := s.PickServer(test.key)
 
 		if test.expectedErr != nil {
-			testutil.NotOk(t, err)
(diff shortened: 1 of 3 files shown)


pkg/cacheutil/memcached_server_selector_test.go
```diff
@@ -37,6 +37,54 @@ func TestNatSort(t *testing.T) {
...
+
+	s := MemcachedJumpHashSelector{}
+
+	for _, test := range tests {
+		testutil.Ok(t, s.SetServers(test.addrs...))
+
+		actualAddr, err := s.PickServer(test.key)
+
+		if test.expectedErr != nil {
+			testutil.NotOk(t, err)
```
GiedriusS: We forgot to check that `err` is equal to `test.expectedErr` here :P

pracucci: Thanks for noticing it! Fixed

bwplotka:
Thanks for this quick reaction :heart:

PR #1984, opened 2020-01-10
api/v1: add cache-control header

GiedriusS:
Let's set `Cache-Control` to `no-store` if there were any warnings while
the query was being executed i.e. it is a partial response or we got
some other problem.

Seems like this works successfully in tandem with https://github.com/cortexproject/cortex/pull/1974.

Tested by turning off one sidecar which leads to warnings and then Cortex Query Frontend did not save anything into the cache when querying.


bwplotka:
We can't merge it, until you undraft it (:

GiedriusS:
> We can't merge it, until you undraft it (:

I know, didn't expect such positive responses and I wasn't fast enough :smile:

Issue #1983, opened 2020-01-10
Unable to fetch data after compaction


nextus:
**Thanos, Prometheus and Golang version used**:

```
thanos, version 0.9.0 (branch: HEAD, revision: 0833cad83db8b257a2275ec83a3d034c73659056)
  build user:       root@e65730470597
  build date:       20191203-16:50:47
  go version:       go1.13.1
```

**Object Storage Provider**:

AWS S3

**What happened**:
Can't access to the historical data from the object storage after compaction (looks similar to #552 and #146):
`Mint: 1571750427740 Maxt: 1578584483987: rpc error: code = Aborted desc = fetch series for block 01DY61NKQNEYFM6BFPDJXZRPQX: preload series: invalid remaining size 65536, expected 71774`
Furthermore, it seems persistent for specific workload: prometheus scrapes node_exporter metrics from the hosts. Right now I have three l4 blocks (Nov 15 - Nov 28, Nov 28 - Dec 12, Dec 12 - Dec 26) and several l3 (Dec 26 - Jan 07). After compaction, I still can fetch historical data from the former block (Nov 15 - Nov 28), but not from the others. Query fails with this response from the Store: https://github.com/thanos-io/thanos/blob/9c8443583a4b5a0c245c8add4fa6ff7a71fc443d/pkg/store/bucket.go#L1566
I have two replicas but the problem exists for each of them. The reason why old compacted block accessible could be in some bug introduced in 0.8 or 0.9 releases: I updated Thanos version for all components on 11 Dec.

**What you expected to happen**:

**How to reproduce it (as minimally and precisely as possible)**:

**Full logs to relevant components**:

There is Compactor logs related to one corrupted block (nothing useful actually):
```
[...]

PR #1985, opened 2020-01-10
store: Refetch series if longer than expected.

bwplotka:
Fixes: https://github.com/thanos-io/thanos/issues/1983

cc @pracucci as we discussed this offline at some point (:

### DIFF
--- a/pkg/objstore/inmem/inmem.go
+++ b/pkg/objstore/inmem/inmem.go
@@ -1,14 +1,13 @@
 package inmem
 
 import (
+	"bytes"
 	"context"
 	"io"
-	"sort"
-	"sync"
-
-	"bytes"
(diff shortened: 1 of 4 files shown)


pkg/objstore/inmem/inmem.go
```diff
@@ -123,6 +122,10 @@ func (b *Bucket) GetRange(_ context.Context, name string, off, length int64) (io
 		return ioutil.NopCloser(bytes.NewReader(file[off:])), nil
 	}
 
+	if length <= 0 {
```
squat: I think the error message does not agree with the condition here. Message says “smaller than 0” but the condition is <=0

bwplotka: Done

Issue #1941, opened 2020-01-06
Use block.MetaFetcher for all meta json syncs


bwplotka:
This is to make sure we have consistent meta.json fetch logic for all command (e.g graceful partial block support)
AC:

* [ ] Use block.MetaFetcher inside verify command
* [ ] Use block.MetaFetcher inside bucket command

daixiang0:
I would like to take this:)

bwplotka:
It's also a good first task for @khyatisoneji 

She is our student as a part of CNCF Community Bridge program, and it sounds like a good introduction if you don't mind @daixiang0

daixiang0:
Never mind, welcome @khyatisoneji

PR #1988, opened 2020-01-11
bucket: use block.MetaFetcher to fetch blocks

khyatisoneji:
Issue link: https://github.com/thanos-io/thanos/issues/1941

## Changes

use `block.metaFetcher` in `bucket` command

### DIFF
--- a/cmd/thanos/bucket.go
+++ b/cmd/thanos/bucket.go
@@ -25,6 +25,7 @@ import (
 	"github.com/thanos-io/thanos/pkg/compact"
 	"github.com/thanos-io/thanos/pkg/component"
 	"github.com/thanos-io/thanos/pkg/extflag"
+	"github.com/thanos-io/thanos/pkg/extprom"
 	extpromhttp "github.com/thanos-io/thanos/pkg/extprom/http"
 	"github.com/thanos-io/thanos/pkg/objstore"
 	"github.com/thanos-io/thanos/pkg/objstore/client"
@@ -37,6 +38,8 @@ import (
 	"golang.org/x/text/message"
 	kingpin "gopkg.in/alecthomas/kingpin.v2"
 )
+
+const extpromPrefix = "thanos_bucket_"
 
 var (
 	issuesMap = map[string]verifier.Issue{
@@ -123,10 +126,15 @@ func registerBucketVerify(m map[string]setupFunc, root *kingpin.CmdClause, name 
 			issues = append(issues, issueFn)
 		}
 
+		fetcher, err := block.NewMetaFetcher(logger, fetcherConcurrency, bkt, "", extprom.WrapRegistererWithPrefix(extpromPrefix, reg))
+		if err != nil {
+			return err
+		}
+
 		if *repair {
-			v = verifier.NewWithRepair(logger, bkt, backupBkt, issues)
+			v = verifier.NewWithRepair(logger, bkt, backupBkt, fetcher, issues)
 		} else {
-			v = verifier.New(logger, bkt, issues)
+			v = verifier.New(logger, bkt, fetcher, issues)
 		}
 
 		var idMatcher func(ulid.ULID) bool = nil
@@ -167,6 +175,11 @@ func registerBucketLs(m map[string]setupFunc, root *kingpin.CmdClause, name stri
 			return err
 		}
 
+		fetcher, err := block.NewMetaFetcher(logger, fetcherConcurrency, bkt, "", extprom.WrapRegistererWithPrefix(extpromPrefix, reg))
+		if err != nil {
+			return err
+		}
+
 		// Dummy actor to immediately kill the group after the run function returns.
 		g.Add(func() error { return nil }, func(error) {})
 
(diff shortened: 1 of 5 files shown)


cmd/thanos/bucket.go
```diff
@@ -234,15 +235,18 @@ func registerBucketLs(m map[string]setupFunc, root *kingpin.CmdClause, name stri
...
 		}
 
-		if err := bkt.Iter(ctx, "", func(name string) error {
-			id, ok := block.IsBlockDir(name)
-			if !ok {
-				return nil
-			}
+		metas, _, metaFetchErr := fetcher.Fetch(ctx)
+		// For partial view allow adding new blocks at least.
+		if metaFetchErr != nil && metas == nil {
```
bwplotka: In Go we stick to `err` variable name for errors (: Even if there is already an `err` variable - that's fine as long as you don't need differentiate those.

Also, do we need to check if `metas` == nil? In Go there is a pattern that if `err` is not nil, caller has to ignore rest of the arguments as they are incorrect... due to error (:

cmd/thanos/bucket.go
```diff
@@ -234,15 +235,18 @@ func registerBucketLs(m map[string]setupFunc, root *kingpin.CmdClause, name stri
...
+			return metaFetchErr
+		}
+
+		for id, meta := range metas {
 			objects++
-			return printBlock(id)
-		}); err != nil {
-			return errors.Wrap(err, "iter")
+			err := printBlock(id, meta)
```
bwplotka: Super small nit: if you have just single return argument and it is an error there is nice pattern that simplifies this:

```
if err := printBlock(...); err != nil {
   return errors.Wrap(err, "print")
} 
```

cmd/thanos/bucket.go
```diff
@@ -406,25 +412,17 @@ func refresh(ctx context.Context, logger log.Logger, bucketUI *ui.Bucket, durati
...
-			return nil
-		}
-
-		meta, err := block.DownloadMeta(ctx, logger, bkt, id)
-		if err != nil {
-			return err
-		}
+	metas, _, metaFetchErr := fetcher.Fetch(ctx)
+	// For partial view allow adding new blocks at least.
+	if metaFetchErr != nil && metas == nil {
```
bwplotka: ditto

cmd/thanos/bucket.go
```diff
@@ -406,25 +412,17 @@ func refresh(ctx context.Context, logger log.Logger, bucketUI *ui.Bucket, durati
 	})
 }
 
-func download(ctx context.Context, logger log.Logger, bkt objstore.Bucket) (blocks []metadata.Meta, err error) {
+func download(ctx context.Context, logger log.Logger, bkt objstore.Bucket, fetcher *block.MetaFetcher) (blocks []metadata.Meta, err error) {
```
bwplotka: I think we can return map here and operate on map on download caller to avoid unnecessary allocations/operations. The reason is rather readability not performance.

pkg/verifier/index_issue.go
```diff
@@ -25,15 +25,16 @@ const IndexIssueID = "index_issue"
...
 
-	err := bkt.Iter(ctx, "", func(name string) error {
-		id, ok := block.IsBlockDir(name)
-		if !ok {
-			return nil
-		}
-
+	metas, _, metaFetchErr := fetcher.Fetch(ctx)
+	// For partial view allow adding new blocks at least.
+	if metaFetchErr != nil && metas == nil {
```
bwplotka: ditto

pkg/verifier/overlapped_blocks.go
```diff
@@ -45,28 +45,20 @@ func OverlappedBlocksIssue(ctx context.Context, logger log.Logger, bkt objstore.
...
-
-		metas[compact.GroupKey(m.Thanos)] = append(metas[compact.GroupKey(m.Thanos)], m.BlockMeta)
-		return nil
-	})
-	if err != nil {
-		return nil, err
+func fetchOverlaps(ctx context.Context, logger log.Logger, bkt objstore.Bucket, fetcher *block.MetaFetcher) (map[string]tsdb.Overlaps, error) {
+	metas, _, metaFetchErr := fetcher.Fetch(ctx)
+	// For partial view allow adding new blocks at least.
+	if metaFetchErr != nil && metas == nil {
```
bwplotka: ditto

pkg/verifier/overlapped_blocks.go
```diff
@@ -45,28 +45,20 @@ func OverlappedBlocksIssue(ctx context.Context, logger log.Logger, bkt objstore.
...
+		return nil, metaFetchErr
+	}
+	
+	groupMetasMap := map[string][]tsdb.BlockMeta{}
+	for _, meta := range metas {
```
bwplotka: :+1:

cmd/thanos/bucket.go
khyatisoneji: the caller function is marshaling the returned array, is it ok if we return map instead of array now?
Will there be any change in the caller because of this?

bwplotka: In this case probably no (: ignore my comment, you are right :+1:

cmd/thanos/bucket.go
```diff
@@ -178,22 +191,18 @@ func registerBucketLs(m map[string]setupFunc, root *kingpin.CmdClause, name stri
...
 			objects    = 0
-			printBlock func(id ulid.ULID) error
+			printBlock func(m *metadata.Meta) error
 		)
 
 		switch format {
 		case "":
-			printBlock = func(id ulid.ULID) error {
+			printBlock = func(m *metadata.Meta) error {
+				id := m.ULID
```
bwplotka: Is this variable really required? I would just inline `m.ULID`.

bwplotka:
As a last think: Can you test manually? e.g `thanos bucket web` on some prepared bucket?

khyatisoneji:
Screenshot 2020-01-12 at 9 02 57 AM
The web component gives this output

bwplotka:
Thank you! Good job 🥇

PR #1990, opened 2020-01-12
*: update for 0.10.0 release

GiedriusS:
Clean up the `CHANGELOG.md` file, update the katacoda scripts, etc. for
the `0.10.0` release.


metalmatze:
There were no changes to metrics right? I'm asking because now we also ship the thanos-mixin. :blush:

GiedriusS:
Good catch, let me double-check. I will also follow up with a PR to improve the release docs to add this step.

GiedriusS:
It seems like all of the metrics changes have been noted in the `CHANGELOG.md` file and they are not used in any of the mixins so this should be good to go :+1: I will update the date to today in both PRs and merge+release on green.

PR #1999, opened 2020-01-15
Toggle SkipVerify for StoreAPI 

cyrus-mc:
I am deploying Prometheus in multiple private (internal) clusters. The Thanos sidecar is exposed via the Nginx Ingress Controller. The ability to support GRPC is supported only over TLS. The ingress is created with a self-signed certificate (as I would prefer not to have to manage certificates) therefore I require the ability to skip verification of the certificate when connecting to the StoreAPI.
## Changes

Add new bool option `grpc-client-skip-verify`, which defaults to false to maintain backwards compatibility, that when set to true skips certificate validation. 

## Verification

Tested against a StoreAPI exposed with a self-signed certificate.

### DIFF
--- a/cmd/thanos/receive.go
+++ b/cmd/thanos/receive.go
@@ -191,7 +191,7 @@ func runReceive(
 	if err != nil {
 		return err
 	}
-	dialOpts, err := extgrpc.StoreClientGRPCOpts(logger, reg, tracer, rwServerCert != "", false, rwClientCert, rwClientKey, rwClientServerCA, rwClientServerName)
+	dialOpts, err := extgrpc.StoreClientGRPCOpts(logger, reg, tracer, rwServerCert != "", rwClientCert, rwClientKey, rwClientServerCA, rwClientServerName, false)
 	if err != nil {
 		return err
 	}
--- a/cmd/thanos/receive.go
+++ b/cmd/thanos/receive.go
@@ -187,11 +187,11 @@ func runReceive(
 	if err != nil {
 		return err
 	}
-	rwTLSClientConfig, err := tls.NewClientConfig(logger, rwClientCert, rwClientKey, rwClientServerCA, rwClientServerName)
+	rwTLSClientConfig, err := tls.NewClientConfig(logger, rwClientCert, rwClientKey, rwClientServerCA, rwClientServerName, false)
 	if err != nil {
 		return err
 	}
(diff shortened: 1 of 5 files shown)


cmd/thanos/receive.go
```diff
@@ -187,11 +187,11 @@ func runReceive(
 	if err != nil {
 		return err
 	}
-	rwTLSClientConfig, err := tls.NewClientConfig(logger, rwClientCert, rwClientKey, rwClientServerCA, rwClientServerName)
+	rwTLSClientConfig, err := tls.NewClientConfig(logger, rwClientCert, rwClientKey, rwClientServerCA, rwClientServerName, false)
```
brancz: This shouldn’t blindly be disabled, doing proper TLS certificate provisioning should be encouraged by default.

cyrus-mc: @brancz False there isn't blindly disabling it. It is disabling the skip verification. So it keeps the default behavior of validating certs. 

The reason I don't want to have to do real certs is that in order to generate certs automatically (say through ACM or K8s cert-mannager) it requires you use a public DNS zone. Meaning I would be putting entries like prometheus-0.company-domain.com that resolve to private IPs in a public zone. While that doesn't allow you to get to that server I have always been opposed to putting private resolution in public zones.

cyrus-mc: @brancz I took your opinion and actually decided to go with a private CA and generate certs against that for the private domain, then just share the CA cert with Thanos.

In the end I still think this PR is useful for the case you mentioned - dev.

cyrus-mc:
> I personally think you should still do proper TLS certificates but I think for dev env this could be useful and is generally a standard thing that exists so I’m not entirely against it. That said I think this should have buy from at least one other maintainer I think.

@brancz I took your opinion and actually decided to go with a private CA and generate certs against that for the private domain, then just share the CA cert with Thanos.

In the end I still think this PR is useful for the case you mentioned - dev.

cmd/thanos/receive.go
brancz: oJ5vMr]ig-D72Yfc/JgV'Q&BA8)?TN5?

brancz: That makes sense, but we still want this to be based on a flag that the user passes no?

bwplotka:
I think private CA is the answer, even for dev - so I would say let's close this PR.

Anyway thanks for the nice work @cyrus-mc and discussion!

brancz:
I assume you’re suggesting something like mkcert? I think agree with that. But we should make sure to document a workflow with it.

PR #2001, opened 2020-01-15
receive: add receive writer metrics.

johncming:
https://github.com/thanos-io/thanos/blob/ac4967e044f41606bf094a053b8d345c251dfebc/pkg/receive/writer.go#L21-L24

`Writer` has no metrics for below counter:

https://github.com/thanos-io/thanos/blob/ac4967e044f41606bf094a053b8d345c251dfebc/pkg/receive/writer.go#L35-L37

Add metrics for `Writer` to watch the values.

### DIFF
--- a/cmd/thanos/receive.go
+++ b/cmd/thanos/receive.go
@@ -286,7 +286,7 @@ func runReceive(
 					}
 					level.Info(logger).Log("msg", "tsdb started")
 					localStorage.Set(db.Get(), startTimeMargin)
-					webHandler.SetWriter(receive.NewWriter(log.With(logger, "component", "receive-writer"), localStorage))
+					webHandler.SetWriter(receive.NewWriter(log.With(logger, "component", "receive-writer"), reg, localStorage))
 					statusProber.Ready()
 					level.Info(logger).Log("msg", "server is ready to receive web requests.")
 					dbReady <- struct{}{}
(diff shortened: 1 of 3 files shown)


pkg/receive/writer.go
```diff
@@ -21,13 +22,37 @@ type Appendable interface {
 type Writer struct {
 	logger log.Logger
 	append Appendable
+
+	numOutOfOrder  prometheus.Gauge
+	numDuplicates  prometheus.Gauge
```
squat: These looks like they should probably all be counters instead of gauges

cmd/thanos/receive.go
```diff
@@ -286,7 +286,7 @@ func runReceive(
 					}
 					level.Info(logger).Log("msg", "tsdb started")
 					localStorage.Set(db.Get(), startTimeMargin)
-					webHandler.SetWriter(receive.NewWriter(log.With(logger, "component", "receive-writer"), localStorage))
+					webHandler.SetWriter(receive.NewWriter(log.With(logger, "component", "receive-writer"), reg, localStorage))
```
squat: Have you tested this on a receive that flushes it’s database? I don’t think this should work. Receive creates a new writer, and thus would register metrics, every time every time it flushes it’s DB. But if you try to register the same metrics multiple times, the registerer will panic. I think you will need to use the receive.UnRegisterer here.

PR #2003, opened 2020-01-15
Support downsampling for /series.

IKSIN:
Signed-off-by: Aleksey Sin <asin@ozon.ru>

## Use-case
Grafana need to get variables from /series. It do not work when we have only downsampled data (we using other retentions for every resolutions).

## Changes

Support downsampling for /series API method. Based on `defaultInstantQueryMaxSourceResolution`

## Verification

Grafana can get data for variables.

### DIFF
--- a/pkg/query/api/v1.go
+++ b/pkg/query/api/v1.go
@@ -505,8 +505,7 @@ func (api *API) series(r *http.Request) (interface{}, []error, *ApiError) {
 		return nil, nil, apiErr
 	}
 
-	// TODO(bwplotka): Support downsampling?
-	q, err := api.queryableCreate(enableDedup, replicaLabels, 0, enablePartialResponse, true).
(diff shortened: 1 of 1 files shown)


GiedriusS:
> This is kind of weird as series does not fetch chunks... I think the fix here is to actually look up all blocks even downsampled for all series? Instead of this?
> 
> What if instead put `maxSourceResolution` = max.Int64`? I think that should include all series... (:
> What do you think?

Yep, we could just do this here. It would reduce the confusion for the users. :+1:

IKSIN:
> This is kind of weird as series does not fetch chunks... I think the fix here is to actually look up all blocks even downsampled for all series? Instead of this?
> 
> What if instead put `maxSourceResolution` = max.Int64`? I think that should include all series... (:
> What do you think?

Yes, I think it good idea! PR Updated)

IKSIN:
@bwplotka Any comments?)

GiedriusS:
I have tested this out - works properly and now returns all data. Thanks!

PR #2006, opened 2020-01-16
compactor: Clear up wording around 0d retention times in docs

moskyb:
(This is a docs-only change - does it need to go in the changelog? If so, I'm more than happy to add it in 🙂)

## Changes

I was confused by the wording around 0d rerentions in the docs - does 0d retention mean that samples aren't retained at all, or that they're not evicted? - so I went through the code, and as far as I can tell, 0d retention means that samples aren't evicted.

I went through the docs and the CLI help text, and added explanations that 0d retention means that samples are never evicted.

## Verification

- Ran `make build`, and then `./thanos compactor -h`, verifying the the new help text was there
- Ran `make web`, then opened up `website/public/components/compact.md/index.html`, and verified that the text I'd added to the docs was present

### DIFF
--- a/docs/components/compact.md
+++ b/docs/components/compact.md
@@ -47,6 +47,8 @@ There's also a case when you might want to disable downsampling at all with `deb
 
 Ideally, you will have equal retention set (or no retention at all) to all resolutions which allow both "zoom in" capabilities as well as performant long ranges queries. Since object storages are usually quite cheap, storage size might not matter that much, unless your goal with thanos is somewhat very specific and you know exactly what you're doing.
 
+Not setting this flag, or setting it to `0d`, i.e. `--retention.resolution-X=0d`, will mean that samples at the `X` resolution level will be kept forever.
+
 ## Storage space consumption
 
 In fact, downsampling doesn't save you any space but instead it adds 2 more blocks for each raw block which are only slightly smaller or relatively similar size to raw block. This is required by internal downsampling implementation which to be mathematically correct holds various aggregations. This means that downsampling can increase the size of your storage a bit (~3x), but it gives massive advantage on querying long ranges.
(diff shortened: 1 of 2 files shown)


docs/components/compact.md
```diff
@@ -47,6 +47,8 @@ There's also a case when you might want to disable downsampling at all with `deb
 
 Ideally, you will have equal retention set (or no retention at all) to all resolutions which allow both "zoom in" capabilities as well as performant long ranges queries. Since object storages are usually quite cheap, storage size might not matter that much, unless your goal with thanos is somewhat very specific and you know exactly what you're doing.
 
+Setting no retention, ie `--retention.resolution-something=0d`, will mean samples at the `something` resolution level will be kept forever.
```
GiedriusS:
```suggestion
Setting no retention, i.e. `--retention.resolution-X=0d`, will mean that samples at the `X` resolution level will be kept forever.
```
Mostly about the missing `.` after `i` and `e` :smile: WDYT?

moskyb: `resolution-X` is way clearer than `resolution-something` - thanks! I'll make this change :)

docs/components/compact.md
```diff
@@ -47,6 +47,8 @@ There's also a case when you might want to disable downsampling at all with `deb
 
 Ideally, you will have equal retention set (or no retention at all) to all resolutions which allow both "zoom in" capabilities as well as performant long ranges queries. Since object storages are usually quite cheap, storage size might not matter that much, unless your goal with thanos is somewhat very specific and you know exactly what you're doing.
 
+Setting no retention, i.e. `--retention.resolution-X=0d`, will mean that samples at the `X` resolution level will be kept forever.
```
bwplotka: I will also mention default behaviour if you do not specify this flag at all (: 

```suggestion
Not setting this flag or setting it to `0d`, i.e. `--retention.resolution-X=0d`, will mean that samples at the `X` resolution level will be kept forever.
```

moskyb: Great point - I've included this in my latest commit.

GiedriusS: Are you sure? GH's UI still shows the old text.

moskyb: Oh whoops! I just forgot to push it 😶 it's there now :)

PR #2008, opened 2020-01-16
querier, receiver, sidecar, store: Add gRPC health check endpoints

### LINKED ISSUE #1980: Support the GRPC Health protocol
While trying to implement service discovery for Thanos components using Consul, we've run into an issue where we're unable to healthcheck Thanos using gRPC because it doesn't support the standard gRPC health protocol. (`grpc.health.v1.Health`) which would allow load balancers and the like to be able to run gRPC healthchecks successfully.

Is this something which is on the radar for future releases?

kakkoyun:
This PR adds gRPC Health handlers to gRPC servers.
Attempts to fix #1980.

## Changes

* Adds gRPC Health handlers to gRPC servers.

## Verification

Tested using https://github.com/grpc-ecosystem/grpc-health-probe

1. ` MINIO_ENABLED=1 ./scripts/quickstart.sh`
1. Use health probe to check
```console
➜ grpc-health-probe -addr=localhost:9999 -connect-timeout 250ms -rpc-timeout 100ms
timeout: failed to connect service "localhost:9999" within 250ms

➜ grpc-health-probe -addr=localhost:10905 -connect-timeout 250ms -rpc-timeout 100ms
status: SERVING

➜ grpc-health-probe -addr=localhost:10903 -connect-timeout 250ms -rpc-timeout 100ms
status: SERVING

➜ grpc-health-probe -addr=localhost:10913 -connect-timeout 250ms -rpc-timeout 100ms
status: SERVING
```


pkg/server/grpc/grpc.go
```diff
@@ -79,6 +81,8 @@ func New(logger log.Logger, reg prometheus.Registerer, tracer opentracing.Tracer
 	storepb.RegisterStoreServer(s, storeSrv)
 	met.InitializeMetrics(s)
 
+	grpc_health.RegisterHealthServer(s, health.NewServer())
```
GiedriusS: Doesn't it mean that we would always be reporting the health as `SERVING`? Maybe it would be a good idea to tie this with the `prober` package?

kakkoyun: I wanted to keep things simple, I assumed if it’s running it’s serving and it’d be enough. However I get your point I’ll integrate it with prober.

kakkoyun:
@GiedriusS I did another iteration on gRPC health checks. Rather than integrating with prober, I have added grpcHealthServer separately. Not all the components serve gRPC, so they don't use prober. I'm not very sure about API, I'm considering adding additional `Resume` and `Pause` methods to `grpcserver.Server` and not expose `health.Server`. WDYT?

@squat Also I'd be glad if you can also have a look at this.

pkg/server/grpc/grpc.go
```diff
@@ -79,10 +82,18 @@ func New(logger log.Logger, reg prometheus.Registerer, tracer opentracing.Tracer
 	storepb.RegisterStoreServer(s, storeSrv)
 	met.InitializeMetrics(s)
 
+	if h == nil {
+		h = health.NewServer()
+	}
+
+	h.SetServingStatus("", grpc_health.HealthCheckResponse_NOT_SERVING)
```
bwplotka: This should be really done in Health constructor

pkg/server/grpc/grpc.go
```diff
@@ -29,13 +31,14 @@ type Server struct {
...
 	srv      *grpc.Server
+	h        *health.Server
 	listener net.Listener
 
 	opts options
 }
 
 // New creates a new Server.
-func New(logger log.Logger, reg prometheus.Registerer, tracer opentracing.Tracer, comp component.Component, storeSrv storepb.StoreServer, opts ...Option) *Server {
+func New(logger log.Logger, reg prometheus.Registerer, tracer opentracing.Tracer, comp component.Component, h *health.Server, storeSrv storepb.StoreServer, opts ...Option) *Server {
```
bwplotka: I think we should pass just interface to health Server not the pointer to exact implementation

pkg/server/grpc/grpc.go
```diff
@@ -135,8 +147,8 @@ type ReadWriteStoreServer interface {
 }
 
 // NewReadWrite creates a new server that can be written to.
-func NewReadWrite(logger log.Logger, reg prometheus.Registerer, tracer opentracing.Tracer, comp component.Component, storeSrv ReadWriteStoreServer, opts ...Option) *Server {
-	s := New(logger, reg, tracer, comp, storeSrv, opts...)
+func NewReadWrite(logger log.Logger, reg prometheus.Registerer, tracer opentracing.Tracer, comp component.Component, h *health.Server, storeSrv ReadWriteStoreServer, opts ...Option) *Server {
```
bwplotka: Uuu, can we do the same for health one?

cmd/thanos/sidecar.go
```diff
@@ -269,13 +275,15 @@ func runSidecar(
...
-		s := grpcserver.New(logger, reg, tracer, comp, promStore,
+		grpcHealthSrv := health.NewServer()
+		s := grpcserver.New(logger, reg, tracer, comp, grpcHealthSrv, promStore,
 			grpcserver.WithListen(grpcBindAddr),
 			grpcserver.WithGracePeriod(grpcGracePeriod),
 			grpcserver.WithTLSConfig(tlsCfg),
 		)
 		g.Add(func() error {
 			statusProber.Ready()
+			grpcHealthSrv.Resume()
```
bwplotka: It's already difficult to not forget about proper statusProber management, not we have two probes to be instrumented. 

What do you think about:
* Wrapping health server as readiness prober
* Adding joinedProber which will regsiter both http Probes and our gRPC readiness probe (named health, but I assumed it will have rather readiness logic)
* Create joined Prober with gRPC + http for places with gRPC traffic as `Ready`, `NotReady` methods are already correctly instrumented there

WDYT?

kakkoyun: I really like the idea 👍 I'll work on it.

kakkoyun:
@bwplotka I did another iteration on it. How does it look? Any suggestions?

cmd/thanos/query.go
```diff
@@ -308,9 +308,12 @@ func runQuery(
 			cancel()
 		})
 	}
-	// Start query API + UI HTTP server.
 
-	statusProber := prober.New(comp, logger, reg)
+	grpcProbe := prober.NewGRPC(comp, logger)
+	httpProbe := prober.NewHTTP(comp, logger, reg)
```
bwplotka: missing wrapper?

pkg/prober/grpc.go
```diff
@@ -0,0 +1,54 @@
...
+
+// HealthServer returns a gRPC health server which responds readiness and liveness checks.
+func (p *GRPCProbe) HealthServer() *health.Server {
+	return p.h
+}
+
+// Ready sets components status to ready.
+func (p *GRPCProbe) Ready() {
+	p.h.SetServingStatus("", grpc_health.HealthCheckResponse_SERVING)
+	level.Info(p.logger).Log("msg", "changing probe status", "status", "ready")
```
bwplotka: Hm, wonder if it's consistent and clean to have log item added to gRPC methods but metrics added to HTTP methods. It's kind of weird. I would either:

* add single logging/metric prober (so prober responsible only to log/metric) and always combine all to the others
* add log line and metric to combine prober and use it everywhere. WDYT?

kakkoyun: I'll find a way to neatly combine observability layer for prober.

poolski:
Thank you! 😍

bwplotka:
Please give us feedback if that works for you 💪🤞

poolski:
It works perfectly! Thank you 👍

PR #2009, opened 2020-01-17
store: add consistency delay to fetch blocks

khyatisoneji:
Add consistency delay in store component to fetch blocks

### DIFF
--- a/cmd/thanos/compact.go
+++ b/cmd/thanos/compact.go
@@ -16,7 +16,6 @@ import (
 	"github.com/go-kit/kit/log"
 	"github.com/go-kit/kit/log/level"
 	"github.com/oklog/run"
-	"github.com/oklog/ulid"
 	"github.com/opentracing/opentracing-go"
 	"github.com/pkg/errors"
 	"github.com/prometheus/client_golang/prometheus"
@@ -184,17 +183,11 @@ func runCompact(
 		Name: "thanos_compactor_iterations_total",
 		Help: "Total number of iterations that were executed successfully.",
 	})
-	consistencyDelayMetric := prometheus.NewGaugeFunc(prometheus.GaugeOpts{
-		Name: "thanos_consistency_delay_seconds",
-		Help: "Configured consistency delay in seconds.",
-	}, func() float64 {
-		return consistencyDelay.Seconds()
-	})
 	partialUploadDeleteAttempts := prometheus.NewCounter(prometheus.CounterOpts{
 		Name: "thanos_compactor_aborted_partial_uploads_deletion_attempts_total",
 		Help: "Total number of started deletions of blocks that are assumed aborted and only partially uploaded.",
 	})
-	reg.MustRegister(halted, retried, iterations, consistencyDelayMetric, partialUploadDeleteAttempts)
+	reg.MustRegister(halted, retried, iterations, partialUploadDeleteAttempts)
 
 	downsampleMetrics := newDownsampleMetrics(reg)
 
@@ -247,15 +240,18 @@ func runCompact(
 		}
 	}()
 
-	metaFetcher, err := block.NewMetaFetcher(logger, 32, bkt, "", extprom.WrapRegistererWithPrefix("thanos_", reg),
+	duplicateBlocksFilter := block.NewDeduplicateFilter()
+	prometheusRegisterer := extprom.WrapRegistererWithPrefix("thanos_", reg)
+	metaFetcher, err := block.NewMetaFetcher(logger, 32, bkt, "", prometheusRegisterer,
 		block.NewLabelShardedMetaFilter(relabelConfig).Filter,
-		(&consistencyDelayMetaFilter{logger: logger, consistencyDelay: consistencyDelay}).Filter,
+		block.NewConsistencyDelayMetaFilter(logger, consistencyDelay, prometheusRegisterer).Filter,
+		duplicateBlocksFilter.Filter,
 	)
 	if err != nil {
 		return errors.Wrap(err, "create meta fetcher")
 	}
 
-	sy, err := compact.NewSyncer(logger, reg, bkt, metaFetcher, blockSyncConcurrency, acceptMalformedIndex, false)
+	sy, err := compact.NewSyncer(logger, reg, bkt, metaFetcher, duplicateBlocksFilter, blockSyncConcurrency, acceptMalformedIndex, false)
 	if err != nil {
 		return errors.Wrap(err, "create syncer")
 	}
@@ -390,25 +386,6 @@ func runCompact(
 	level.Info(logger).Log("msg", "starting compact node")
 	statusProber.Ready()
 	return nil
-}
-
-type consistencyDelayMetaFilter struct {
-	logger           log.Logger
-	consistencyDelay time.Duration
-}
-
-func (f *consistencyDelayMetaFilter) Filter(metas map[ulid.ULID]*metadata.Meta, synced block.GaugeLabeled, _ bool) {
-	for id, meta := range metas {
-		if ulid.Now()-id.Time() < uint64(f.consistencyDelay/time.Millisecond) &&
-			meta.Thanos.Source != metadata.BucketRepairSource &&
-			meta.Thanos.Source != metadata.CompactorSource &&
-			meta.Thanos.Source != metadata.CompactorRepairSource {
-
-			level.Debug(f.logger).Log("msg", "block is too fresh for now", "block", id)
-			synced.WithLabelValues(block.TooFreshMeta).Inc()
-			delete(metas, id)
-		}
-	}
 }
 
 // genMissingIndexCacheFiles scans over all blocks, generates missing index cache files and uploads them to object storage.
--- a/cmd/thanos/store.go
+++ b/cmd/thanos/store.go
@@ -81,6 +81,9 @@ func registerStore(m map[string]setupFunc, app *kingpin.Application) {
 	enableIndexHeader := cmd.Flag("experimental.enable-index-header", "If true, Store Gateway will recreate index-header instead of index-cache.json for each block. This will replace index-cache.json permanently once it will be out of experimental stage.").
 		Hidden().Default("false").Bool()
 
+	consistencyDelay := modelDuration(cmd.Flag("consistency-delay", "Minimum age of all blocks before they are being read.").
+		Default("30m"))
+
 	m[component.Store.String()] = func(g *run.Group, logger log.Logger, reg *prometheus.Registry, tracer opentracing.Tracer, _ <-chan struct{}, debugLogging bool) error {
 		if minTime.PrometheusTimestamp() > maxTime.PrometheusTimestamp() {
 			return errors.Errorf("invalid argument: --min-time '%s' can't be greater than --max-time '%s'",
@@ -116,6 +119,7 @@ func registerStore(m map[string]setupFunc, app *kingpin.Application) {
 			selectorRelabelConf,
 			*advertiseCompatibilityLabel,
 			*enableIndexHeader,
+			time.Duration(*consistencyDelay),
 		)
 	}
 }
@@ -148,6 +152,7 @@ func runStore(
 	selectorRelabelConf *extflag.PathOrContent,
 	advertiseCompatibilityLabel bool,
 	enableIndexHeader bool,
+	consistencyDelay time.Duration,
 ) error {
 	grpcProbe := prober.NewGRPC()
 	httpProbe := prober.NewHTTP()
@@ -220,9 +225,11 @@ func runStore(
 		return errors.Wrap(err, "create index cache")
 	}
 
-	metaFetcher, err := block.NewMetaFetcher(logger, fetcherConcurrency, bkt, dataDir, extprom.WrapRegistererWithPrefix("thanos_", reg),
+	prometheusRegisterer := extprom.WrapRegistererWithPrefix("thanos_", reg)
+	metaFetcher, err := block.NewMetaFetcher(logger, fetcherConcurrency, bkt, dataDir, prometheusRegisterer,
 		block.NewTimePartitionMetaFilter(filterConf.MinTime, filterConf.MaxTime).Filter,
 		block.NewLabelShardedMetaFilter(relabelConfig).Filter,
+		block.NewConsistencyDelayMetaFilter(logger, consistencyDelay, prometheusRegisterer).Filter,
 		block.NewDeduplicateFilter().Filter,
 	)
 	if err != nil {
--- a/docs/components/store.md
+++ b/docs/components/store.md
@@ -27,7 +27,8 @@ In general about 1MB of local disk space is required per TSDB block stored in th
 
 ## Flags
 
-[embedmd]:# (flags/store.txt $)
+[embedmd]: # "flags/store.txt $"
+
 ```$
 usage: thanos store [<flags>]
 
@@ -137,6 +138,7 @@ Flags:
                                  Prometheus relabel-config syntax. See format
                                  details:
                                  https://prometheus.io/docs/prometheus/latest/configuration/configuration/#relabel_config
+      --consistency-delay=30m    Minimum age of all blocks before they are being read.
 
 ```
 
@@ -179,7 +181,8 @@ The `in-memory` index cache is enabled by default and its max size can be config
 
 Alternatively, the `in-memory` index cache can also by configured using `--index-cache.config-file` to reference to the configuration file or `--index-cache.config` to put yaml config directly:
 
-[embedmd]:# (../flags/config_index_cache_in_memory.txt yaml)
+[embedmd]: # "../flags/config_index_cache_in_memory.txt yaml"
+
 ```yaml
 type: IN-MEMORY
 config:
@@ -196,7 +199,8 @@ All the settings are **optional**:
 
 The `memcached` index cache allows to use [Memcached](https://memcached.org) as cache backend. This cache type is configured using `--index-cache.config-file` to reference to the configuration file or `--index-cache.config` to put yaml config directly:
 
-[embedmd]:# (../flags/config_index_cache_memcached.txt yaml)
+[embedmd]: # "../flags/config_index_cache_memcached.txt yaml"
+
 ```yaml
 type: MEMCACHED
 config:
@@ -224,13 +228,12 @@ While the remaining settings are **optional**:
 - `max_get_multi_batch_size`: maximum number of keys a single underlying operation should fetch. If more keys are specified, internally keys are splitted into multiple batches and fetched concurrently, honoring `max_get_multi_concurrency`. If set to `0`, the batch size is unlimited.
 - `dns_provider_update_interval`: the DNS discovery update interval.
 
-
 ## Index Header
 
 In order to query series inside blocks from object storage, Store Gateway has to know certain initial info about each block such as:
 
-* symbols table to unintern string values
-* postings offset for posting lookup
+- symbols table to unintern string values
+- postings offset for posting lookup
 
 In order to achieve so, on startup for each block `index-header` is built from pieces of original block's index and stored on disk.
 Such `index-header` file is then mmaped and used by Store Gateway.
--- a/pkg/block/fetcher.go
+++ b/pkg/block/fetcher.go
@@ -417,7 +417,7 @@ func (f *LabelShardedMetaFilter) Filter(metas map[ulid.ULID]*metadata.Meta, sync
 
 // DeduplicateFilter is a MetaFetcher filter that filters out older blocks that have exactly the same data.
 type DeduplicateFilter struct {
-	DuplicateIDs []ulid.ULID
+	duplicateIDs []ulid.ULID
 }
 
 // NewDeduplicateFilter creates DeduplicateFilter.
@@ -428,16 +428,30 @@ func NewDeduplicateFilter() *DeduplicateFilter {
 // Filter filters out duplicate blocks that can be formed
 // from two or more overlapping blocks that fully submatches the source blocks of the older blocks.
 func (f *DeduplicateFilter) Filter(metas map[ulid.ULID]*metadata.Meta, synced GaugeLabeled, _ bool) {
-	root := NewNode(&metadata.Meta{
-		BlockMeta: tsdb.BlockMeta{
-			ULID: ulid.MustNew(uint64(0), nil),
-		},
-	})
+	var wg sync.WaitGroup
 
-	metaSlice := []*metadata.Meta{}
+	metasByResolution := make(map[int64][]*metadata.Meta)
 	for _, meta := range metas {
-		metaSlice = append(metaSlice, meta)
+		res := meta.Thanos.Downsample.Resolution
+		metasByResolution[res] = append(metasByResolution[res], meta)
 	}
+
+	for res := range metasByResolution {
+		wg.Add(1)
+		go func(res int64) {
+			defer wg.Done()
+			f.filterForResolution(NewNode(&metadata.Meta{
+				BlockMeta: tsdb.BlockMeta{
+					ULID: ulid.MustNew(uint64(0), nil),
+				},
+			}), metasByResolution[res], metas, res, synced)
+		}(res)
+	}
+
+	wg.Wait()
+}
+
+func (f *DeduplicateFilter) filterForResolution(root *Node, metaSlice []*metadata.Meta, metas map[ulid.ULID]*metadata.Meta, res int64, synced GaugeLabeled) {
 	sort.Slice(metaSlice, func(i, j int) bool {
 		ilen := len(metaSlice[i].Compaction.Sources)
 		jlen := len(metaSlice[j].Compaction.Sources)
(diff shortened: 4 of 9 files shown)


khyatisoneji:
Although this is a rough implementation of adding filters in store, I wanted to get feedback on this approach.

I have added two filters:
a. Consistency filter: This is to account for eventual consistency for storing data in external storage
b. Garbage Blocks filter: This is to filter out the blocks that will be deleted by compactor, so these blocks should not be synced by store.

Please let me know if there are any additional filters that I should apply and also if this is in the right direction.
I will add tests and also make the code more structured.

cmd/thanos/store.go
```diff
@@ -75,6 +78,9 @@ func registerStore(m map[string]setupFunc, app *kingpin.Application) {
 
 	selectorRelabelConf := regSelectorRelabelFlags(cmd)
 
+	consistencyDelay := modelDuration(cmd.Flag("consistency-delay", fmt.Sprintf("Minimum age of fresh (non-compacted) blocks before they are being processed. Malformed blocks older than the maximum of consistency-delay and %v will be removed.", compact.PartialUploadThresholdAge)).
```
GiedriusS: I feel like the second sentence doesn't belong here since Thanos Store doesn't actually perform the deletion itself. My suggestion:
```suggestion
	consistencyDelay := modelDuration(cmd.Flag("consistency-delay", "Minimum age of fresh (non-compacted) blocks before they are being processed.").
```

cmd/thanos/store.go
```diff
@@ -140,7 +147,16 @@ func runStore(
...
 	advertiseCompatibilityLabel bool,
+	consistencyDelay time.Duration,
 ) error {
+	consistencyDelayMetric := prometheus.NewGaugeFunc(prometheus.GaugeOpts{
+		Name: "thanos_consistency_delay_seconds",
+		Help: "Configured consistency delay in seconds.",
+	}, func() float64 {
+		return consistencyDelay.Seconds()
+	})
+	reg.MustRegister(consistencyDelayMetric)
```
GiedriusS: Since we have `NewConsistencyDelayMetaFilter` now perhaps the metrics could be registered there? We could pass the registry as: `extprom.WrapRegistererWithPrefix("thanos_", reg)` and then use a `component.Component` as the `Namespace`, and only then `consistency_delay_seconds` as the metric's name. I think it would be cleaner and less duplication between `cmd/thanos/store.go` and `cmd/thanos/compact.go`. What do you think?

cmd/thanos/store.go
```diff
@@ -206,9 +222,12 @@ func runStore(
 		return errors.Wrap(err, "create index cache")
 	}
 
+	garbageBlocksFinder := garbagecollector.NewGarbageBlocksFinder()
 	metaFetcher, err := block.NewMetaFetcher(logger, fetcherConcurrency, bkt, dataDir, extprom.WrapRegistererWithPrefix("thanos_", reg),
 		block.NewTimePartitionMetaFilter(filterConf.MinTime, filterConf.MaxTime).Filter,
 		block.NewLabelShardedMetaFilter(relabelConfig).Filter,
+		block.NewConsistencyDelayMetaFilter(logger, consistencyDelay).Filter,
+		garbagecollector.NewGarbageBlocksFilter(garbageBlocksFinder).Filter,
```
GiedriusS: :clap: :+1:

khyatisoneji: Adding the garbageblocks filter is a good addition right?

GiedriusS: Yep! It should reduce RAM usage in certain cases.

cmd/thanos/store.go
khyatisoneji: I agree with this, I have made the change

khyatisoneji:
@GiedriusS the tests are still breaking, can you retrigger them? Also please check if that it because of code changes

pkg/block/fetcher.go
```diff
@@ -407,3 +407,40 @@ func (f *LabelShardedMetaFilter) Filter(metas map[ulid.ULID]*metadata.Meta, sync
...
+// ConsistencyDelayMetaFilter is a MetaFetcher filter that filters out blocks that are created before a specified consistency delay.
+type ConsistencyDelayMetaFilter struct {
+	logger           log.Logger
+	consistencyDelay time.Duration
+}
+
+// NewConsistencyDelayMetaFilter creates ConsistencyDelayMetaFilter.
+func NewConsistencyDelayMetaFilter(logger log.Logger, consistencyDelay time.Duration, reg prometheus.Registerer) *ConsistencyDelayMetaFilter {
+	consistencyDelayMetric := prometheus.NewGaugeFunc(prometheus.GaugeOpts{
+		Name: "thanos_consistency_delay_seconds",
```
GiedriusS: Since the `prometheus.Registerer` already has the `thanos_` prefix, it is not needed here.

cmd/thanos/store.go
```diff
@@ -206,9 +212,12 @@ func runStore(
 		return errors.Wrap(err, "create index cache")
 	}
 
+	garbageBlocksFinder := garbagecollector.NewGarbageBlocksFinder()
 	metaFetcher, err := block.NewMetaFetcher(logger, fetcherConcurrency, bkt, dataDir, extprom.WrapRegistererWithPrefix("thanos_", reg),
 		block.NewTimePartitionMetaFilter(filterConf.MinTime, filterConf.MaxTime).Filter,
 		block.NewLabelShardedMetaFilter(relabelConfig).Filter,
+		block.NewConsistencyDelayMetaFilter(logger, consistencyDelay, extprom.WrapRegistererWithPrefix("thanos_", reg)).Filter,
```
GiedriusS: Since you will need to touch the code again then we could move out `extprom.WrapRegistererWithPrefix` into a separate variable and use it on this line and on L216.

pkg/testutil/prometheus.go
```diff
@@ -304,10 +320,15 @@ func createBlock(
...
 	numSamples int,
 	mint, maxt int64,
+	blockDelay time.Duration,
 	extLset labels.Labels,
 	resolution int64,
 	tombstones bool,
 ) (id ulid.ULID, err error) {
+	if err != nil {
+		return id, errors.Wrap(err, "create block id")
+	}
```
yeya24: Should this error be moved below Line 401, I guess?

pkg/compact/compact_e2e_test.go
```diff
@@ -29,6 +29,10 @@ import (
 	"github.com/thanos-io/thanos/pkg/testutil"
 )
 
+const (
+	BlockDelay = 30 * time.Minute
```
yeya24: Is it necessary to expose this const publicly?

test/e2e/store_gateway_test.go
```diff
@@ -22,6 +22,10 @@ import (
 	yaml "gopkg.in/yaml.v2"
 )
 
+const (
+	BlockDelay = 30 * time.Minute
```
yeya24: Same with the previous one

pkg/compact/compact_e2e_test.go
khyatisoneji: Made this change

pkg/testutil/prometheus.go
khyatisoneji: Can you check why the tests are breaking?

khyatisoneji:
The `TestStoreGateway` test has been failing.

The reason is that in this test, we are creating the block by calling tsdb LeveledCompactor's Write function.
In this function, it creates the folder using `ulid.Now()`, but here we want to use `ulid.Now() - consistencyDelay` as the folder to write to, since we want to use the default value of consistencyDelay in the test.

Is there anyway in which I can provide the id of the block to write to tsdb? Otherwise we have to use consistencyDelay as 0.

yeya24:
> The `TestStoreGateway` test has been failing.
> 
> The reason is that in this test, we are creating the block by calling tsdb LeveledCompactor's Write function.
> In this function, it creates the folder using `ulid.Now()`, but here we want to use `ulid.Now() - consistencyDelay` as the folder to write to, since we want to use the default value of consistencyDelay in the test.
> 
> Is there anyway in which I can provide the id of the block to write to tsdb? Otherwise we have to use consistencyDelay as 0.

AFAIK, there is no way to customize block id when you want to create a block in TSDB

khyatisoneji:
What would be an alternate way to test a block id that was created consistencyDelay time ago?

Our default value of consistencyDelay is 30m, so we want to only fetch the blocks which are older than 30m. If we use consistencyDelay as 30m, then the call to fetch blocks fails saying that the block is too fresh.
We cannot also mock the `NewConsistencyDelayMetaFilter` function.

Is there a way to test the retrieval of blocks that are old?

pkg/testutil/e2eutil/prometheus.go
```diff
@@ -277,6 +280,134 @@ func (p *Prometheus) Appender() tsdb.Appender {
 	return p.db.Appender()
 }
 
+// CreateEmptyBlock produces empty block like it was the case before fix: https://github.com/prometheus/tsdb/pull/374.
+// (Prometheus pre v2.7.0).
+func CreateEmptyBlock(dir string, blockDelay time.Duration, mint int64, maxt int64, extLset labels.Labels, resolution int64) (ulid.ULID, error) {
```
GiedriusS: Is `blockDelay` ever not `0`? AFAICT there aren't any current users and everything uses `CreateBlockWithBlockDelay`. Do you plan on using this?

cmd/thanos/store.go
```diff
@@ -81,6 +82,9 @@ func registerStore(m map[string]setupFunc, app *kingpin.Application) {
 	enableIndexHeader := cmd.Flag("experimental.enable-index-header", "If true, Store Gateway will recreate index-header instead of index-cache.json for each block. This will replace index-cache.json permanently once it will be out of experimental stage.").
 		Hidden().Default("false").Bool()
 
+	consistencyDelay := modelDuration(cmd.Flag("consistency-delay", "Minimum age of fresh (non-compacted) blocks before they are being processed.").
```
bwplotka: It is not only for non-compacted, right?

bwplotka: processed? I would say just read.

cmd/thanos/compact.go
```diff
@@ -247,15 +241,17 @@ func runCompact(
 		}
 	}()
 
-	metaFetcher, err := block.NewMetaFetcher(logger, 32, bkt, "", extprom.WrapRegistererWithPrefix("thanos_", reg),
+	prometheusRegisterer := extprom.WrapRegistererWithPrefix("thanos_", reg)
+	metaFetcher, err := block.NewMetaFetcher(logger, 32, bkt, "", prometheusRegisterer,
 		block.NewLabelShardedMetaFilter(relabelConfig).Filter,
-		(&consistencyDelayMetaFilter{logger: logger, consistencyDelay: consistencyDelay}).Filter,
+		block.NewConsistencyDelayMetaFilter(logger, consistencyDelay, prometheusRegisterer).Filter,
```
bwplotka: No `garbagecollector.NewGarbageBlocksFilter(garbageBlocksFinder).Filter,`?

pkg/block/fetcher.go
```diff
@@ -506,3 +506,40 @@ func contains(s1 []ulid.ULID, s2 []ulid.ULID) bool {
...
+		logger:           logger,
+		consistencyDelay: consistencyDelay,
+	}
+}
+
+// Filter filters out blocks that filters blocks that have are created before a specified consistency delay.
+func (f *ConsistencyDelayMetaFilter) Filter(metas map[ulid.ULID]*metadata.Meta, synced GaugeLabeled, _ bool) {
+	for id, meta := range metas {
+		if ulid.Now()-id.Time() < uint64(f.consistencyDelay/time.Millisecond) &&
+			meta.Thanos.Source != metadata.BucketRepairSource &&
```
bwplotka: I think for Store Gateway at some point we don't need those special cases. However since we don't have Delete delay yet on compactor, let's leave it here. Maybe adding TODO would nice here?

pkg/compact/garbagecollector/garbage_collector.go
```diff
@@ -0,0 +1,98 @@
...
+	parents := map[ulid.ULID]ulid.ULID{}
+
+	for id, meta := range blocks {
+		// Skip any block that has a different resolution.
+		if meta.Thanos.Downsample.Resolution != resolution {
+			continue
+		}
+
+		// For each source block we contain, check whether we are the highest priority parent block.
+		for _, sid := range meta.Compaction.Sources {
```
bwplotka: This is confusing for me. How this is really different to our Duplicate block filter? :thinking: Isn't this exactly the same?

pkg/compact/garbagecollector/garbage_collector.go
```diff
@@ -0,0 +1,98 @@
...
+	return &GarbageBlocksFilter{
+		garbageBlocksFinder: garbageBlocksFinder,
+	}
+}
+
+func (f *GarbageBlocksFilter) Filter(metas map[ulid.ULID]*metadata.Meta, synced block.GaugeLabeled, _ bool) {
+	for _, res := range []int64{
+		downsample.ResLevel0, downsample.ResLevel1, downsample.ResLevel2,
+	} {
+		garbageIds, _ := f.garbageBlocksFinder.GarbageBlocks(res, metas)
```
bwplotka: This is 1:1 the same as our duplicate block filter, right?

pkg/testutil/e2eutil/prometheus.go
bwplotka: What about, 
```suggestion
func CreateEmptyBlock(dir string, creationTime time.Time, mint int64, maxt int64, extLset labels.Labels, resolution int64) (ulid.ULID, error) {
```
? It is more verbose. block delay feels like you want to upload it with delay or something... Let's be explicit.

pkg/testutil/e2eutil/prometheus.go
```diff
@@ -277,6 +280,134 @@ func (p *Prometheus) Appender() tsdb.Appender {
...
+	ctx context.Context,
+	dir string,
+	series []labels.Labels,
+	numSamples int,
+	mint, maxt int64,
+	blockDelay time.Duration,
+	extLset labels.Labels,
+	resolution int64,
+) (id ulid.ULID, err error) {
+	blockID, err := createBlock(ctx, dir, series, numSamples, mint, maxt, extLset, resolution, false)
```
bwplotka: It looks really similar to `CreateEmptyBlock`.

Can we reuse more code based on `numSamples`?

cmd/thanos/compact.go
bwplotka: Also this is exactly the same as our duplicate block filter :thinking:

cmd/thanos/store.go
```diff
@@ -220,10 +226,14 @@ func runStore(
...
 
-	metaFetcher, err := block.NewMetaFetcher(logger, fetcherConcurrency, bkt, dataDir, extprom.WrapRegistererWithPrefix("thanos_", reg),
+	garbageBlocksFinder := garbagecollector.NewGarbageBlocksFinder()
+	prometheusRegisterer := extprom.WrapRegistererWithPrefix("thanos_", reg)
+	metaFetcher, err := block.NewMetaFetcher(logger, fetcherConcurrency, bkt, dataDir, prometheusRegisterer,
 		block.NewTimePartitionMetaFilter(filterConf.MinTime, filterConf.MaxTime).Filter,
 		block.NewLabelShardedMetaFilter(relabelConfig).Filter,
+		block.NewConsistencyDelayMetaFilter(logger, consistencyDelay, prometheusRegisterer).Filter,
 		block.NewDeduplicateFilter().Filter,
+		garbagecollector.NewGarbageBlocksFilter(garbageBlocksFinder).Filter,
```
bwplotka: Wait... we already have `block.NewDeduplicateFilter().Filter,`... so why garbage collector is needed?

pkg/compact/compact.go
```diff
@@ -249,67 +252,8 @@ func (s *Syncer) GarbageCollect(ctx context.Context) error {
...
-		}
-
-		ids = append(ids, id)
-	}
-	return ids, nil
-}
-
 func (s *Syncer) garbageCollect(ctx context.Context, resolution int64) error {
-	garbageIds, err := s.GarbageBlocks(resolution)
+	garbageIds, err := s.garbageBlocksFinder.GarbageBlocks(resolution, s.blocks)
```
bwplotka: I believe this will do nothing as we have duplicate block filter already. We need to use that instead and remove GarbageCollect method. I thought that was the idea from the beginning (:

cmd/thanos/compact.go
khyatisoneji: added duplicateBlocksFilter

docs/components/store.md
```diff
@@ -27,7 +27,8 @@ In general about 1MB of local disk space is required per TSDB block stored in th
 
 ## Flags
 
-[embedmd]:# (flags/store.txt $)
+[embedmd]: # "flags/store.txt $"
```
khyatisoneji: changes done by vscode automatically, please advice if I should remove this or not

bwplotka: That's fine as long as embedding works.

pkg/compact/compact.go
khyatisoneji: removed the garbage filter

pkg/testutil/e2eutil/prometheus.go
khyatisoneji: I tried doing this, but getting an exception while calling tsdb Compactor.write function.
It throws exception when it calls `populateBlock` function.

pkg/block/fetcher.go
khyatisoneji: added todo, please review

bwplotka:
I think we missed an important things: What if those blocks have different
external labels?

pkg/block/fetcher.go
```diff
@@ -428,15 +428,30 @@ func NewDeduplicateFilter() *DeduplicateFilter {
...
+					ULID: ulid.MustNew(uint64(0), nil),
+				},
+			}), metas, res, synced)
+		}(res)
+	}
 
+	wg.Wait()
+}
+
+func (f *DeduplicateFilter) filter(root *Node, metas map[ulid.ULID]*metadata.Meta, resolution int64, synced GaugeLabeled) {
```
bwplotka: Small suggestion for readability:
```suggestion
func (f *DeduplicateFilter) filterForResolution(root *Node, metas map[ulid.ULID]*metadata.Meta, res int64, synced GaugeLabeled) {
```

pkg/block/fetcher.go
```diff
@@ -428,15 +428,30 @@ func NewDeduplicateFilter() *DeduplicateFilter {
...
 // from two or more overlapping blocks that fully submatches the source blocks of the older blocks.
 func (f *DeduplicateFilter) Filter(metas map[ulid.ULID]*metadata.Meta, synced GaugeLabeled, _ bool) {
-	root := NewNode(&metadata.Meta{
-		BlockMeta: tsdb.BlockMeta{
-			ULID: ulid.MustNew(uint64(0), nil),
-		},
-	})
+	var wg sync.WaitGroup
+	for _, res := range []int64{
+		int64(0), int64(5 * 60 * 1000), int64(60 * 60 * 1000),
```
bwplotka: Hm, this will work. But what about doing something more generic and building metaSlice as `map[int64][]*metadata.Meta{}`? And then filterForResolution? So if we change resolutions someday it will be still fully working?

pkg/compact/compact.go
```diff
@@ -231,89 +233,7 @@ func (s *Syncer) GarbageCollect(ctx context.Context) error {
...
-	return ids, nil
-}
-
-func (s *Syncer) garbageCollect(ctx context.Context, resolution int64) error {
-	garbageIds, err := s.GarbageBlocks(resolution)
-	if err != nil {
-		return err
-	}
-
+	garbageIds := s.duplicateBlocksFilter.GetDuplicateIDs()
```
bwplotka: hm.. it would be nice to comment on `GarbageCollect` that Sync is required for this to work.

pkg/testutil/e2eutil/prometheus.go
bwplotka: What kind of panic?

test/e2e/store_gateway_test.go
```diff
@@ -75,18 +75,19 @@ func TestStoreGateway(t *testing.T) {
 	series := []labels.Labels{
 		labels.FromStrings("a", "1", "b", "2"),
 	}
+
```
bwplotka: I think we are missing test cases for our dedup filter to make sure resolution grouping works, right?

khyatisoneji: Done

pkg/testutil/e2eutil/prometheus.go
khyatisoneji: so when we are creating a block without any samples, tsdb returns block id as ulid.ULID{} which is like 0000000000000000.
It is from this code: https://github.com/prometheus/prometheus/blob/master/tsdb/compact.go#L492

So we try to fetch block with id as 0, and thus fail to get its `meta.json` file and fail to populate Thanos labels and it panics.

pkg/block/fetcher.go
```diff
@@ -428,16 +428,33 @@ func NewDeduplicateFilter() *DeduplicateFilter {
...
-		},
-	})
+	var wg sync.WaitGroup
 
-	metaSlice := []*metadata.Meta{}
+	metasByResolutionMap := make(map[int64][]*metadata.Meta)
 	for _, meta := range metas {
-		metaSlice = append(metaSlice, meta)
+		res := meta.Thanos.Downsample.Resolution
+		if metasByResolutionMap[res] == nil {
```
bwplotka: You don't need this. NO map item for `[]*metadata.Meta` means default value... so just nil slice. `append` can handle nil slice just fine.

pkg/block/fetcher.go
```diff
@@ -428,16 +428,33 @@ func NewDeduplicateFilter() *DeduplicateFilter {
...
 func (f *DeduplicateFilter) Filter(metas map[ulid.ULID]*metadata.Meta, synced GaugeLabeled, _ bool) {
-	root := NewNode(&metadata.Meta{
-		BlockMeta: tsdb.BlockMeta{
-			ULID: ulid.MustNew(uint64(0), nil),
-		},
-	})
+	var wg sync.WaitGroup
 
-	metaSlice := []*metadata.Meta{}
+	metasByResolutionMap := make(map[int64][]*metadata.Meta)
```
bwplotka: Go is a language that is type safe, so we don't need add suffix to name what `type` this variable is. IDE and code already show us that this is a `map`. No need for this suffix.

pkg/block/fetcher.go
```diff
@@ -428,16 +428,33 @@ func NewDeduplicateFilter() *DeduplicateFilter {
...
+			metasByResolutionMap[res] = []*metadata.Meta{}
+		}
+		metasByResolutionMap[res] = append(metasByResolutionMap[res], meta)
+	}
+
+	for res := range metasByResolutionMap {
+		wg.Add(1)
+		go func(res int64) {
+			defer wg.Done()
+			f.filterForResolution(NewNode(&metadata.Meta{
```
bwplotka: nice, very smart and clean :+1:

pkg/block/fetcher.go
```diff
@@ -463,6 +480,12 @@ func (f *DeduplicateFilter) Filter(metas map[ulid.ULID]*metadata.Meta, synced Ga
 	}
 }
 
+// GetDuplicateIDs returns slice of block ids
+// that are filtered out by DeduplicateFilter.
+func (f *DeduplicateFilter) GetDuplicateIDs() []ulid.ULID {
```
bwplotka: Get/Set prefixes for Go language are not preferred, so just 
```suggestion
func (f *DeduplicateFilter) DuplicateIDs() []ulid.ULID {
```
...is totally fine.

pkg/block/fetcher.go
```diff
@@ -463,6 +480,12 @@ func (f *DeduplicateFilter) Filter(metas map[ulid.ULID]*metadata.Meta, synced Ga
...
+	return f.DuplicateIDs
```
bwplotka: if you have getter for this, let's make it private.

pkg/block/fetcher.go
```diff
@@ -506,3 +529,42 @@ func contains(s1 []ulid.ULID, s2 []ulid.ULID) bool {
...
+	return &ConsistencyDelayMetaFilter{
+		logger:           logger,
+		consistencyDelay: consistencyDelay,
+	}
+}
+
+// Filter filters out blocks that filters blocks that have are created before a specified consistency delay.
+func (f *ConsistencyDelayMetaFilter) Filter(metas map[ulid.ULID]*metadata.Meta, synced GaugeLabeled, _ bool) {
+	for id, meta := range metas {
+		// TODO(khyatisoneji): Remove the checks about Thanos Source
```
bwplotka: Well... This is more tricky than just this. But let's talk on later PRs.

pkg/block/fetcher_test.go
```diff
@@ -396,18 +396,32 @@ func TestTimePartitionMetaFilter_Filter(t *testing.T) {
 
 }
 
+type sourcesAndResolution struct {
```
bwplotka: :+1:

pkg/testutil/e2eutil/prometheus.go
bwplotka: Then why not change the ulid?

bwplotka: Anyway, I think it's fine for this PR. We can reuse more later on next one :+1:

khyatisoneji: it doesn't return the ulid that it creates, also it doesn't create any block if numSamples = 0

pkg/block/fetcher.go
```diff
@@ -506,3 +526,42 @@ func contains(s1 []ulid.ULID, s2 []ulid.ULID) bool {
...
+		//  by implementing delete delay to fetch metas.
+		if ulid.Now()-id.Time() < uint64(f.consistencyDelay/time.Millisecond) &&
```
bwplotka: Let's add another todo: `TODO(bwplotka): Check consistency delay based on file upload / modification time instead of ULID`.

khyatisoneji: done

bwplotka:
:clap: :clap: :clap:

PR #2012, opened 2020-01-20
receive: Add support for TSDB per tenant

brancz:

## Changes

This PR adds multi TSDB support for the receive component.

## Verification

Tested manually and e2e tests.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -252,7 +252,7 @@ test-e2e: docker
 	@echo ">> running /test/e2e tests."
 	# NOTE(bwplotka):
 	# * If you see errors on CI (timeouts), but not locally, try to add -parallel 1 to limit to single CPU to reproduce small 1CPU machine.
-	@go test -failfast -timeout 10m -v ./test/e2e/...
+	@go test -failfast -timeout 10m $(RUN_ARGS) -v ./test/e2e/...
 
 .PHONY: install-deps
 install-deps: ## Installs dependencies for integration tests. It installs supported versions of Prometheus and alertmanager to test against in integration tests.
--- a/Makefile
+++ b/Makefile
@@ -16,6 +16,8 @@ export GO111MODULE
 GOPROXY           ?= https://proxy.golang.org
 export GOPROXY
 
+GOTEST_OPTS ?= -failfast -timeout 10m -v
+
 # Tools.
 EMBEDMD           ?= $(GOBIN)/embedmd-$(EMBEDMD_VERSION)
 # v2.0.0
@@ -252,7 +254,7 @@ test-e2e: docker
 	@echo ">> running /test/e2e tests."
 	# NOTE(bwplotka):
 	# * If you see errors on CI (timeouts), but not locally, try to add -parallel 1 to limit to single CPU to reproduce small 1CPU machine.
-	@go test -failfast -timeout 10m $(RUN_ARGS) -v ./test/e2e/...
+	@go test $(GOTEST_OPTS) ./test/e2e/...
 
 .PHONY: install-deps
 install-deps: ## Installs dependencies for integration tests. It installs supported versions of Prometheus and alertmanager to test against in integration tests.
--- a/cmd/thanos/receive.go
+++ b/cmd/thanos/receive.go
@@ -20,18 +20,17 @@ import (
 	"github.com/prometheus/prometheus/storage/tsdb"
 	kingpin "gopkg.in/alecthomas/kingpin.v2"
 
-	"github.com/thanos-io/thanos/pkg/block/metadata"
 	"github.com/thanos-io/thanos/pkg/component"
 	"github.com/thanos-io/thanos/pkg/extflag"
 	"github.com/thanos-io/thanos/pkg/extgrpc"
 	"github.com/thanos-io/thanos/pkg/extprom"
+	"github.com/thanos-io/thanos/pkg/objstore"
 	"github.com/thanos-io/thanos/pkg/objstore/client"
 	"github.com/thanos-io/thanos/pkg/prober"
 	"github.com/thanos-io/thanos/pkg/receive"
 	"github.com/thanos-io/thanos/pkg/runutil"
 	grpcserver "github.com/thanos-io/thanos/pkg/server/grpc"
 	httpserver "github.com/thanos-io/thanos/pkg/server/http"
-	"github.com/thanos-io/thanos/pkg/shipper"
 	"github.com/thanos-io/thanos/pkg/store"
 	"github.com/thanos-io/thanos/pkg/tls"
 )
@@ -203,7 +202,43 @@ func runReceive(
 		return err
 	}
 
-	dbs := receive.NewMultiTSDB(dataDir, logger, reg, tsdbOpts, lset, tenantLabelName)
+	confContentYaml, err := objStoreConfig.Content()
+	if err != nil {
+		return err
+	}
+	upload := true
+	if len(confContentYaml) == 0 {
+		level.Info(logger).Log("msg", "No supported bucket was configured, uploads will be disabled")
+		upload = false
+	}
+
+	if upload && tsdbOpts.MinBlockDuration != tsdbOpts.MaxBlockDuration {
+		if !ignoreBlockSize {
+			return errors.Errorf("found that TSDB Max time is %s and Min time is %s. "+
+				"Compaction needs to be disabled (tsdb.min-block-duration = tsdb.max-block-duration)", tsdbOpts.MaxBlockDuration, tsdbOpts.MinBlockDuration)
+		}
+		level.Warn(logger).Log("msg", "flag to ignore min/max block duration flags differing is being used. If the upload of a 2h block fails and a tsdb compaction happens that block may be missing from your Thanos bucket storage.")
+	}
+
+	var bkt objstore.Bucket
+	if upload {
+		// The background shipper continuously scans the data directory and uploads
+		// new blocks to Google Cloud Storage or an S3-compatible storage service.
+		bkt, err = client.NewBucket(logger, confContentYaml, reg, comp.String())
+		if err != nil {
+			return err
+		}
+	}
+
+	dbs := receive.NewMultiTSDB(
+		dataDir,
+		logger,
+		reg,
+		tsdbOpts,
+		lset,
+		tenantLabelName,
+		bkt,
+	)
 	writer := receive.NewWriter(log.With(logger, "component", "receive-writer"), dbs)
 	webHandler := receive.NewHandler(log.With(logger, "component", "receive-handler"), &receive.Options{
 		Writer:            writer,
@@ -226,24 +261,6 @@ func runReceive(
 		grpcProbe,
 		prober.NewInstrumentation(comp, logger, extprom.WrapRegistererWithPrefix("thanos_", reg)),
 	)
-
-	confContentYaml, err := objStoreConfig.Content()
-	if err != nil {
-		return err
-	}
-	upload := true
-	if len(confContentYaml) == 0 {
-		level.Info(logger).Log("msg", "No supported bucket was configured, uploads will be disabled")
-		upload = false
-	}
-
-	if upload && tsdbOpts.MinBlockDuration != tsdbOpts.MaxBlockDuration {
-		if !ignoreBlockSize {
-			return errors.Errorf("found that TSDB Max time is %s and Min time is %s. "+
-				"Compaction needs to be disabled (tsdb.min-block-duration = tsdb.max-block-duration)", tsdbOpts.MaxBlockDuration, tsdbOpts.MinBlockDuration)
-		}
-		level.Warn(logger).Log("msg", "flag to ignore min/max block duration flags differing is being used. If the upload of a 2h block fails and a tsdb compaction happens that block may be missing from your Thanos bucket storage.")
-	}
 
 	// Start all components while we wait for TSDB to open but only load
 	// initial config and mark ourselves as ready after it completed.
@@ -438,18 +455,9 @@ func runReceive(
 	}
 
 	if upload {
-		// The background shipper continuously scans the data directory and uploads
-		// new blocks to Google Cloud Storage or an S3-compatible storage service.
-		bkt, err := client.NewBucket(logger, confContentYaml, reg, comp.String())
-		if err != nil {
-			return err
-		}
-
-		s := shipper.New(logger, reg, dataDir, bkt, func() labels.Labels { return lset }, metadata.ReceiveSource)
-
-		// Before starting, ensure any old blocks are uploaded.
-		if uploaded, err := s.Sync(context.Background()); err != nil {
-			level.Warn(logger).Log("err", err, "failed to upload", uploaded)
+		level.Debug(logger).Log("msg", "upload enabled")
+		if err := dbs.Upload(context.Background()); err != nil {
+			level.Warn(logger).Log("err", err)
 		}
 
 		{
@@ -457,8 +465,8 @@ func runReceive(
 			ctx, cancel := context.WithCancel(context.Background())
 			g.Add(func() error {
 				return runutil.Repeat(30*time.Second, ctx.Done(), func() error {
-					if uploaded, err := s.Sync(ctx); err != nil {
-						level.Warn(logger).Log("err", err, "uploaded", uploaded)
+					if err := dbs.Upload(ctx); err != nil {
+						level.Warn(logger).Log("err", err)
 					}
 
 					return nil
@@ -479,8 +487,8 @@ func runReceive(
 				// Before quitting, ensure all blocks are uploaded.
 				defer func() {
 					<-uploadC
-					if uploaded, err := s.Sync(context.Background()); err != nil {
-						level.Warn(logger).Log("err", err, "failed to upload", uploaded)
+					if err := dbs.Upload(context.Background()); err != nil {
+						level.Warn(logger).Log("err", err)
 					}
 				}()
 				defer close(uploadDone)
@@ -494,8 +502,8 @@ func runReceive(
 					case <-ctx.Done():
 						return nil
 					case <-uploadC:
-						if uploaded, err := s.Sync(ctx); err != nil {
-							level.Warn(logger).Log("err", err, "failed to upload", uploaded)
+						if err := dbs.Upload(ctx); err != nil {
+							level.Warn(logger).Log("err", err)
 						}
 						uploadDone <- struct{}{}
 					}
--- a/cmd/thanos/receive.go
+++ b/cmd/thanos/receive.go
@@ -202,32 +202,28 @@ func runReceive(
 		return err
 	}
 
+	var bkt objstore.Bucket
 	confContentYaml, err := objStoreConfig.Content()
 	if err != nil {
 		return err
 	}
-	upload := true
-	if len(confContentYaml) == 0 {
-		level.Info(logger).Log("msg", "No supported bucket was configured, uploads will be disabled")
-		upload = false
-	}
-
-	if upload && tsdbOpts.MinBlockDuration != tsdbOpts.MaxBlockDuration {
-		if !ignoreBlockSize {
-			return errors.Errorf("found that TSDB Max time is %s and Min time is %s. "+
-				"Compaction needs to be disabled (tsdb.min-block-duration = tsdb.max-block-duration)", tsdbOpts.MaxBlockDuration, tsdbOpts.MinBlockDuration)
-		}
-		level.Warn(logger).Log("msg", "flag to ignore min/max block duration flags differing is being used. If the upload of a 2h block fails and a tsdb compaction happens that block may be missing from your Thanos bucket storage.")
-	}
-
-	var bkt objstore.Bucket
+	upload := len(confContentYaml) > 0
 	if upload {
+		if tsdbOpts.MinBlockDuration != tsdbOpts.MaxBlockDuration {
+			if !ignoreBlockSize {
+				return errors.Errorf("found that TSDB Max time is %s and Min time is %s. "+
+					"Compaction needs to be disabled (tsdb.min-block-duration = tsdb.max-block-duration)", tsdbOpts.MaxBlockDuration, tsdbOpts.MinBlockDuration)
+			}
+			level.Warn(logger).Log("msg", "flag to ignore min/max block duration flags differing is being used. If the upload of a 2h block fails and a tsdb compaction happens that block may be missing from your Thanos bucket storage.")
+		}
 		// The background shipper continuously scans the data directory and uploads
-		// new blocks to Google Cloud Storage or an S3-compatible storage service.
+		// new blocks to object storage service.
 		bkt, err = client.NewBucket(logger, confContentYaml, reg, comp.String())
 		if err != nil {
 			return err
 		}
+	} else {
+		level.Info(logger).Log("msg", "no supported bucket was configured, uploads will be disabled")
 	}
 
 	dbs := receive.NewMultiTSDB(
@@ -310,9 +306,8 @@ func runReceive(
 						uploadC <- struct{}{}
 						<-uploadDone
 					}
-					level.Info(logger).Log("msg", "tsdb started")
 					statusProber.Ready()
-					level.Info(logger).Log("msg", "server is ready to receive web requests")
+					level.Info(logger).Log("msg", "tsdb started, and server is ready to receive web requests")
 					dbReady <- struct{}{}
 				}
 			}
@@ -411,9 +406,13 @@ func runReceive(
 					s.Shutdown(errors.New("reload hashrings"))
 				}
 
-				multiStore := store.NewMultiTSDBStore(logger, reg, comp, dbs.TSDBStores)
 				rw := store.ReadWriteTSDBStore{
-					StoreServer:          multiStore,
+					StoreServer: store.NewMultiTSDBStore(
+						logger,
+						reg,
+						comp,
+						dbs.TSDBStores,
+					),
 					WriteableStoreServer: webHandler,
 				}
 
@@ -456,8 +455,8 @@ func runReceive(
 
 	if upload {
 		level.Debug(logger).Log("msg", "upload enabled")
-		if err := dbs.Upload(context.Background()); err != nil {
-			level.Warn(logger).Log("err", err)
+		if err := dbs.Sync(context.Background()); err != nil {
+			level.Warn(logger).Log("msg", "initial upload failed", "err", err)
 		}
 
 		{
@@ -465,8 +464,8 @@ func runReceive(
 			ctx, cancel := context.WithCancel(context.Background())
 			g.Add(func() error {
 				return runutil.Repeat(30*time.Second, ctx.Done(), func() error {
-					if err := dbs.Upload(ctx); err != nil {
-						level.Warn(logger).Log("err", err)
+					if err := dbs.Sync(ctx); err != nil {
+						level.Warn(logger).Log("msg", "interval upload failed", "err", err)
 					}
 
 					return nil
@@ -487,8 +486,8 @@ func runReceive(
 				// Before quitting, ensure all blocks are uploaded.
 				defer func() {
 					<-uploadC
-					if err := dbs.Upload(context.Background()); err != nil {
-						level.Warn(logger).Log("err", err)
+					if err := dbs.Sync(context.Background()); err != nil {
+						level.Warn(logger).Log("msg", "on demnad upload failed", "err", err)
 					}
 				}()
 				defer close(uploadDone)
@@ -502,7 +501,7 @@ func runReceive(
 					case <-ctx.Done():
 						return nil
 					case <-uploadC:
-						if err := dbs.Upload(ctx); err != nil {
+						if err := dbs.Sync(ctx); err != nil {
 							level.Warn(logger).Log("err", err)
 						}
 						uploadDone <- struct{}{}
--- a/cmd/thanos/receive.go
+++ b/cmd/thanos/receive.go
@@ -72,6 +72,10 @@ func registerReceive(m map[string]setupFunc, app *kingpin.Application) {
 
 	tenantHeader := cmd.Flag("receive.tenant-header", "HTTP header to determine tenant for write requests.").Default(receive.DefaultTenantHeader).String()
 
+	defaultTenantID := cmd.Flag("receive.default-tenant-id", "Default tenant ID to use when none is provided via a header.").Default(receive.DefaultTenant).String()
+
+	tenantLabelName := cmd.Flag("receive.tenant-label-name", "Label name through which the tenant will be announced.").Default(receive.DefaultTenantLabel).String()
+
 	replicaHeader := cmd.Flag("receive.replica-header", "HTTP header specifying the replica number of a write request.").Default(receive.DefaultReplicaHeader).String()
 
 	replicationFactor := cmd.Flag("receive.replication-factor", "How many times to replicate incoming write requests.").Default("1").Uint64()
@@ -144,6 +148,8 @@ func registerReceive(m map[string]setupFunc, app *kingpin.Application) {
 			cw,
 			*local,
 			*tenantHeader,
+			*defaultTenantID,
+			*tenantLabelName,
 			*replicaHeader,
 			*replicationFactor,
 			comp,
@@ -179,6 +185,8 @@ func runReceive(
 	cw *receive.ConfigWatcher,
 	endpoint string,
 	tenantHeader string,
+	defaultTenantID string,
+	tenantLabelName string,
 	replicaHeader string,
 	replicationFactor uint64,
 	comp component.SourceStoreAPI,
@@ -186,7 +194,6 @@ func runReceive(
 	logger = log.With(logger, "component", "receive")
 	level.Warn(logger).Log("msg", "setting up receive; the Thanos receive component is EXPERIMENTAL, it may break significantly without notice")
 
-	localStorage := &tsdb.ReadyStorage{}
 	rwTLSConfig, err := tls.NewServerConfig(log.With(logger, "protocol", "HTTP"), rwServerCert, rwServerKey, rwServerClientCA)
 	if err != nil {
 		return err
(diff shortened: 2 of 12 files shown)


pkg/store/multitsdb.go
```diff
@@ -0,0 +1,156 @@
...
+
+	return resp, nil
+}
+
+// Series returns all series for a requested time range and label matcher. The returned data may
+// exceed the requested time bounds.
+func (s *MultiTSDBStore) Series(r *storepb.SeriesRequest, srv storepb.Store_SeriesServer) error {
+	stores := s.tsdbStores()
+	for _, store := range stores {
+		err := store.Series(r, srv)
```
bwplotka: `if err := ...; err != nil { ` pattern is available here

bwplotka: Hm..... does this works? Don't we need some proxyStore here?

squat:
Still in progress

brancz:
hopefully will soon get back to working on this

pkg/receive/tsdb_test.go
```diff
@@ -48,6 +48,9 @@ func TestFlushableStorage(t *testing.T) {
 
 		// Append data to the WAL.
 		app := db.Appender()
+		if err != nil {
```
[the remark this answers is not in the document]
brancz: actually there was no error to test in the first place :rofl:

pkg/store/multitsdb.go
```diff
@@ -0,0 +1,156 @@
...
+	seenLabelSets := map[uint64]struct{}{}
+	labelSets := []storepb.LabelSet{}
+	for _, info := range infos {
+		for _, labelSet := range info.LabelSets {
+			ls := storepb.LabelsToPromLabels(labelSet.Labels)
+			sort.Sort(ls)
+			h := ls.Hash()
+			_, seen := seenLabelSets[h]
+			if !seen {
+				seenLabelSets[h] = struct{}{}
```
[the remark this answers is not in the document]
brancz: yeah true, this way we actually also have predictable allocations, which is a nice side effect

pkg/store/multitsdb.go
brancz: Picking this up now.

pracucci:
According to an offline discussion with @brancz, the plan would be to keep a flat structure in bucket, so that every tenant TSDB block is stored within the same prefix.

Have you considered a structure like `bucket://<tenant-id>/<block-id>`? The benefit would be having a storage compatible with Cortex (it's how we store per-tenant blocks in the bucket).

brancz:
This is now finally ready for true review. All intended functionality is now available. I intentionally kept it focused on the receive component for now. I could see a future where we make the ruler also multitsdb capable, but I think that needs more thought so I didn't want to prematurely overgeneralize.

pkg/receive/multitsdb.go
```diff
@@ -0,0 +1,269 @@
...
+			wg.Done()
+		}()
+	}
+
+	wg.Wait()
+	return merr.Err()
+}
+
+func (t *MultiTSDB) Upload(ctx context.Context) error {
+	if !t.upload {
```
bwplotka, line 125: I think this should be handled from outside, no? It sounds like opportunity for too many ways of doing stuff and threat for consistency (: I would remove this and just control this from outside only. And return error here if you want on bkt=nil case

pkg/receive/multitsdb.go
```diff
@@ -0,0 +1,269 @@
...
+
+	errmtx := &sync.Mutex{}
+	merr := terrors.MultiError{}
+	wg := &sync.WaitGroup{}
+	for tenant, s := range t.shippers {
+		level.Debug(t.logger).Log("msg", "uploading block for tenant", "tenant", tenant)
+		s := s
+		wg.Add(1)
+		go func() {
+			if uploaded, err := s.Sync(ctx); err != nil {
```
bwplotka, line 140: Wow the key would be to have instrumentation to tell which TSDB was slowest in sync and affecting others etc

pkg/receive/multitsdb.go
```diff
@@ -0,0 +1,269 @@
...
+				errmtx.Lock()
+				merr.Add(fmt.Errorf("failed to upload %d: %w", uploaded, err))
```
bwplotka, line 142:
```suggestion
				merr.Add(errors.Wrap(err, "sync %d", uploaded))
```

As per https://github.com/thanos-io/thanos/blob/8828b3b797ae1b75a19c0ccc4024b8c212c1adc9/docs/contributing/coding-style-guide.md#wrap-errors-for-more-context-dont-repeat-failed--there

pkg/receive/multitsdb.go
```diff
@@ -0,0 +1,269 @@
...
+			wg.Done()
+		}()
+	}
+
+	wg.Wait()
+	return merr.Err()
+}
+
+func (t *MultiTSDB) openTSDBs() error {
+	files, err := ioutil.ReadDir(t.dataDir)
```
bwplotka, line 154: why not inlining in  `Open`  method? Looks like used once.

pkg/receive/multitsdb.go
```diff
@@ -0,0 +1,269 @@
...
+
+	var g errgroup.Group
+	for _, f := range files {
+		f := f
+		if !f.IsDir() {
+			continue
+		}
+
+		g.Go(func() error {
+			tenantId := f.Name()
```
bwplotka, line 167: inline `f.Name`?

pkg/receive/multitsdb.go
```diff
@@ -0,0 +1,269 @@
...
+	return g.Wait()
+}
+
+func (t *MultiTSDB) TSDBStores() map[string]*store.TSDBStore {
+	t.mtx.RLock()
+	res := make(map[string]*store.TSDBStore, len(t.stores))
+	for k, v := range t.stores {
+		res[k] = v
+	}
+	defer t.mtx.RUnlock()
```
bwplotka, line 182: hmm odd to defer something as last step

pkg/receive/multitsdb.go
```diff
@@ -0,0 +1,269 @@
...
+	return res
+}
+
+func (t *MultiTSDB) getOrLoadTenant(tenantID string) (*tsdb.ReadyStorage, error) {
+	// Fast path, as creating tenants is a very rare operation.
```
bwplotka, line 187: Fast if rare? :thinking:  that sounds weird. So who we care about fast then?

cmd/thanos/receive.go
```diff
@@ -283,28 +300,25 @@ func runReceive(
 
 					level.Info(logger).Log("msg", "updating DB")
 
-					if err := db.Flush(); err != nil {
+					if err := dbs.Flush(); err != nil {
```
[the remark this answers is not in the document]
brancz: I thought about this as well. I think we should do that as a follow up as it essentially means we need to carefully dissect the configuration and if done incorrectly we are risking loosing data or having inconsistent data.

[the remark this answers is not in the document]
bwplotka: Agree, just want to make sure we are thinking about it. Worth to add a  TODO or issue?

pkg/store/multitsdb.go
```diff
@@ -0,0 +1,284 @@
...
+		recv:   make(chan *storepb.Series),
+		errMtx: &sync.Mutex{},
+	}
+}
+
+func (s *seriesSetServer) Context() context.Context {
+	return s.ctx
+}
+
+func (s *seriesSetServer) Run(store *TSDBStore, r *storepb.SeriesRequest) {
```
bwplotka, line 113:
```suggestion
func (s *seriesSetServer) Series(store *TSDBStore, r *storepb.SeriesRequest) {
```

pkg/store/multitsdb.go
```diff
@@ -0,0 +1,284 @@
...
+}
+
+func (s *seriesSetServer) At() ([]storepb.Label, []storepb.AggrChunk) {
+	if s.cur == nil {
+		return nil, nil
+	}
+	return s.cur.Labels, s.cur.Chunks
+}
+
+func (s *seriesSetServer) Err() error {
```
bwplotka, line 151: just curious why only Err have mtxs?

pkg/store/multitsdb.go
```diff
@@ -0,0 +1,284 @@
...
+	s.errMtx.Lock()
+	defer s.errMtx.Unlock()
+	return s.err
+}
+
+// Series returns all series for a requested time range and label matcher. The returned data may
+// exceed the requested time bounds.
+func (s *MultiTSDBStore) Series(r *storepb.SeriesRequest, srv storepb.Store_SeriesServer) error {
```
bwplotka, line 159: This looks good to me, but ... it also looks similar to proxy? Why not using proxy itself?

pkg/store/multitsdb.go
```diff
@@ -0,0 +1,284 @@
...
+	})
+
+	for resp := range respRecv {
+		if err := srv.Send(resp); err != nil {
+			return status.Error(codes.Unknown, errors.Wrap(err, "send series response").Error())
+		}
+	}
+
+	if err := g.Wait(); err != nil {
+		level.Error(s.logger).Log("err", err)
```
bwplotka, line 217: Error handled twice. I know we miss request logging, but worth to add it everywhere then?

pkg/store/multitsdb.go
```diff
@@ -0,0 +1,284 @@
...
+		if err != nil {
+			return nil, err
+		}
+
+		for _, l := range r.Values {
+			values[l] = struct{}{}
+		}
+
+		for _, l := range r.Warnings {
+			warnings[l] = struct{}{}
```
bwplotka, line 276: do we want really deduplicate warnings :thinking:  same for names

cmd/thanos/receive.go
```diff
@@ -196,11 +202,51 @@ func runReceive(
 		return err
 	}
 
+	confContentYaml, err := objStoreConfig.Content()
+	if err != nil {
+		return err
+	}
+	upload := true
+	if len(confContentYaml) == 0 {
+		level.Info(logger).Log("msg", "No supported bucket was configured, uploads will be disabled")
```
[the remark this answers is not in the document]
brancz: will fix in any place that I'm touching anyways, otherwise agreed, we should do a 1 time sweep

cmd/thanos/receive.go
```diff
@@ -196,11 +202,51 @@ func runReceive(
...
+	if upload && tsdbOpts.MinBlockDuration != tsdbOpts.MaxBlockDuration {
+		if !ignoreBlockSize {
+			return errors.Errorf("found that TSDB Max time is %s and Min time is %s. "+
+				"Compaction needs to be disabled (tsdb.min-block-duration = tsdb.max-block-duration)", tsdbOpts.MaxBlockDuration, tsdbOpts.MinBlockDuration)
+		}
+		level.Warn(logger).Log("msg", "flag to ignore min/max block duration flags differing is being used. If the upload of a 2h block fails and a tsdb compaction happens that block may be missing from your Thanos bucket storage.")
+	}
+
+	var bkt objstore.Bucket
+	if upload {
```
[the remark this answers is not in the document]
brancz: was trying to change the smallest amount of things, but happy to consolidate

cmd/thanos/receive.go
```diff
@@ -196,11 +202,51 @@ func runReceive(
...
+		// The background shipper continuously scans the data directory and uploads
+		// new blocks to Google Cloud Storage or an S3-compatible storage service.
```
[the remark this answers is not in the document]
brancz: Just kept things from before, but will adapt.

cmd/thanos/receive.go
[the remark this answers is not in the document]
brancz: Actually the upload functionality has a dependency on the `dbs` reference, so we need to at least defer it until after creation of it, but dbs itself requires the bucket, and `upload` is used in various other places. I simplified a little bit, but the general structure is needed as is. If you have another suggestion, I'm all ears (still collecting things from other comments and will push everything together later).

cmd/thanos/receive.go
```diff
@@ -440,27 +455,18 @@ func runReceive(
...
-		if err != nil {
-			return err
-		}
-
-		s := shipper.New(logger, reg, dataDir, bkt, func() labels.Labels { return lset }, metadata.ReceiveSource)
-
-		// Before starting, ensure any old blocks are uploaded.
-		if uploaded, err := s.Sync(context.Background()); err != nil {
-			level.Warn(logger).Log("err", err, "failed to upload", uploaded)
+		level.Debug(logger).Log("msg", "upload enabled")
```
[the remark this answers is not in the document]
brancz: `dbs` requires it for instantiation

cmd/thanos/receive.go
```diff
@@ -440,27 +455,18 @@ func runReceive(
...
+		if err := dbs.Upload(context.Background()); err != nil {
```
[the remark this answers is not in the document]
brancz: Will rename to `Sync`.

pkg/receive/handler.go
```diff
@@ -526,6 +528,13 @@ func isConflict(err error) bool {
 		status.Code(err) == codes.AlreadyExists
 }
 
+// isNotReady returns whether or not the given error represents a not ready error.
+func isNotReady(err error) bool {
+	return err == tsdb.ErrNotReady ||
+		err.Error() == strconv.Itoa(http.StatusConflict) ||
```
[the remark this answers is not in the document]
brancz: frankly I don't remember myself, I think this was a mistake

pkg/receive/multitsdb.go
```diff
@@ -0,0 +1,269 @@
...
+	logger          log.Logger
+	reg             prometheus.Registerer
+	tsdbCfg         *tsdb.Options
+	tenantLabelName string
+	labels          labels.Labels
+	bucket          objstore.Bucket
+	upload          bool
+
+	mtx         *sync.RWMutex
+	dbs         map[string]*FlushableStorage
```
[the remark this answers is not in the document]
brancz: Yeah my initial thought (and parts of the implementation) were that these are distinct concerns and may be individually optimizable (some of these at some point were slices instead of maps). But the further I got, the clearer it actually became that this is precisely our _tenant_ definition. So I'll adapt to make that more clear.

pkg/receive/handler.go
```diff
@@ -358,26 +359,22 @@ func (h *Handler) parallelizeRequests(ctx context.Context, tenant string, replic
...
+				if errs, ok := err.(terrors.MultiError); ok {
+					if countCause(errs, isConflict) > 0 {
+						err = errors.Wrap(conflictErr, errs.Error())
+					} else if countCause(errs, isNotReady) > 0 {
+						err = tsdb.ErrNotReady
+					} else {
+						err = errors.New(errs.Error())
 					}
 				}
-				h.mtx.RUnlock()
```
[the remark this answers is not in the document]
brancz: Can you elaborate what you mean here? The mutex is used in various other places as well and is still necessary in those places if that's what you mean.

pkg/receive/multitsdb.go
brancz: sounds good to me

pkg/receive/multitsdb.go
```diff
@@ -0,0 +1,269 @@
...
+func (t *MultiTSDB) Open() error {
+	if err := os.MkdirAll(t.dataDir, 0777); err != nil {
+		return err
+	}
+
+	return t.openTSDBs()
+}
+
+func (t *MultiTSDB) Close() error {
+	t.mtx.Lock()
```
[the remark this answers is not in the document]
brancz: Double open or double close should never happen.

pkg/receive/multitsdb.go
```diff
@@ -0,0 +1,269 @@
...
+		shippers:        map[string]*shipper.Shipper{},
+		labels:          labels,
+		tenantLabelName: tenantLabelName,
+		bucket:          bucket,
+		upload:          bucket != nil,
+	}
+}
+
+func (t *MultiTSDB) Open() error {
+	if err := os.MkdirAll(t.dataDir, 0777); err != nil {
```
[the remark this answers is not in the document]
brancz: Things are already more granularly locked in the functions called.

pkg/receive/multitsdb.go
brancz: We already have metrics per shipper, what else do you think we need?

pkg/receive/multitsdb.go
brancz: To me the original comment makes more sense, but no strong opinion.

pkg/receive/multitsdb.go
```diff
@@ -0,0 +1,269 @@
...
+			dataDir,
+			logger,
+			reg,
+			t.tsdbCfg,
+		)
+
+		if err := s.Open(); err != nil {
+			level.Error(logger).Log("msg", "failed to open tsdb", "err", err)
+			t.mtx.Lock()
+			delete(t.appendables, tenantID)
```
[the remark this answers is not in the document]
brancz: Why and how would that happen? The way we lock things, this is the only path that would ever get here and exactly once.

pkg/receive/multitsdb.go
```diff
@@ -0,0 +1,269 @@
...
+		t.dbs[tenantID] = s
+		t.shippers[tenantID] = ship
+		t.mtx.Unlock()
+	}()
+
+	return rs, nil
+}
+
+func (t *MultiTSDB) TenantAppendable(tenantID string) (Appendable, error) {
+	return t.getOrLoadTenant(tenantID)
```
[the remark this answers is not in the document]
brancz: This is primarily to have a clear interface where we _only_ return an `Appendable` to the outer world, but have the real tsdb instance internally. This is also changing slightly as I'm introducing a real `tenant` struct as a follow up from your comments.

pkg/receive/multitsdb.go
```diff
@@ -0,0 +1,269 @@
...
+		tstore := store.NewTSDBStore(
+			logger,
+			reg,
+			s.Get(),
+			component.Receive,
+			lbls,
+		)
+
+		t.mtx.Lock()
+		rs.Set(s.Get(), int64(2*time.Duration(t.tsdbCfg.MinBlockDuration).Seconds()*1000))
```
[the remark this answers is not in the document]
brancz: Twice as in locking twice? If that's what you meant, then yes I noticed this as well, removing :)

pkg/receive/writer.go
```diff
@@ -21,29 +22,44 @@ type Appendable interface {
...
+	}
+
+	app, err := s.Appender()
+	if err == tsdb.ErrNotReady {
+		return err
+	}
 	if err != nil {
 		return errors.Wrap(err, "get appender")
 	}
+	if app == nil {
```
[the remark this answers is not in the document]
brancz: you're right it can't, it's already covered by the not ready error in the ready storage

pkg/store/multitsdb.go
```diff
@@ -0,0 +1,284 @@
...
+		if err != nil {
+			return nil, err
+		}
+		infos = append(infos, info)
+	}
+
+	resp.MinTime = infos[0].MinTime
+	resp.MaxTime = infos[0].MaxTime
+
+	for i := 1; i < len(infos); i++ {
```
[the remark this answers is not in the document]
brancz: Yeah that's what my initial implementation did, but we can actually make a number of assumptions when we just deal with multiple TSDBs, I think generalizing further than what's already reused here would be premature.

pkg/store/multitsdb.go
brancz: It doesn't. It's a leftover from intermediate work. Good catch.

pkg/store/multitsdb.go
brancz: I implemented both, and the proxy implementation got very complex and felt wrong as I was faking stores to be real grpc servers. There are some similarities, especially in the Series function, but there are a number of assumptions that we can make because we are in a more tightly scoped scenario. I opted for this one as it ended up being the simpler implementation of the two.

pkg/store/multitsdb.go
brancz: Duplicate labelnames in the response would be more than confusing. Same with warnings, it doesn't seem helpful to the user if the same warning shows up multiple times.

cmd/thanos/receive.go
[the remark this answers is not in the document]
brancz: Modified this a bit, let me know if this is better.

cmd/thanos/receive.go
```diff
@@ -283,28 +300,25 @@ func runReceive(
...
 					}
-					if err := db.Open(); err != nil {
+					if err := dbs.Open(); err != nil {
 						return errors.Wrap(err, "opening storage")
 					}
 					if upload {
 						uploadC <- struct{}{}
 						<-uploadDone
 					}
 					level.Info(logger).Log("msg", "tsdb started")
```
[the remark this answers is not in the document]
brancz: Done.

cmd/thanos/receive.go
```diff
@@ -397,9 +410,10 @@ func runReceive(
 				if s != nil {
 					s.Shutdown(errors.New("reload hashrings"))
 				}
-				tsdbStore := store.NewTSDBStore(log.With(logger, "component", "thanos-tsdb-store"), nil, localStorage.Get(), comp, lset)
+
+				multiStore := store.NewMultiTSDBStore(logger, reg, comp, dbs.TSDBStores)
```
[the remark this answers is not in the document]
brancz: Done

cmd/thanos/receive.go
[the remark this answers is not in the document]
brancz: Done

cmd/thanos/receive.go
```diff
@@ -481,8 +487,8 @@ func runReceive(
 				// Before quitting, ensure all blocks are uploaded.
 				defer func() {
 					<-uploadC
-					if uploaded, err := s.Sync(context.Background()); err != nil {
-						level.Warn(logger).Log("err", err, "failed to upload", uploaded)
+					if err := dbs.Upload(context.Background()); err != nil {
+						level.Warn(logger).Log("err", err)
```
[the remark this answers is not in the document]
brancz: done

pkg/receive/multitsdb.go
```diff
@@ -0,0 +1,269 @@
...
+
+type MultiTSDB struct {
+	dataDir         string
+	logger          log.Logger
+	reg             prometheus.Registerer
+	tsdbCfg         *tsdb.Options
+	tenantLabelName string
+	labels          labels.Labels
+	bucket          objstore.Bucket
+	upload          bool
```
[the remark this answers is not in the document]
brancz: done

pkg/receive/multitsdb.go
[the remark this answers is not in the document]
brancz: Moved to a `tenant` struct. Let me know what you think.

pkg/receive/multitsdb.go
```diff
@@ -0,0 +1,269 @@
...
+
+	mtx         *sync.RWMutex
+	dbs         map[string]*FlushableStorage
+	appendables map[string]*tsdb.ReadyStorage
+	stores      map[string]*store.TSDBStore
+	shippers    map[string]*shipper.Shipper
+}
+
+func NewMultiTSDB(dataDir string, l log.Logger, reg prometheus.Registerer, tsdbCfg *tsdb.Options, labels labels.Labels, tenantLabelName string, bucket objstore.Bucket) *MultiTSDB {
```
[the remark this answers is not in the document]
brancz: Done.

pkg/receive/multitsdb.go
```diff
@@ -0,0 +1,269 @@
...
+		tsdbCfg:         tsdbCfg,
+		mtx:             &sync.RWMutex{},
+		dbs:             map[string]*FlushableStorage{},
+		stores:          map[string]*store.TSDBStore{},
+		appendables:     map[string]*tsdb.ReadyStorage{},
+		shippers:        map[string]*shipper.Shipper{},
+		labels:          labels,
+		tenantLabelName: tenantLabelName,
+		bucket:          bucket,
+		upload:          bucket != nil,
```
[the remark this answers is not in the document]
brancz: done

pkg/receive/multitsdb.go
brancz: done, I don't think `bkt = nil` case needs to be errored on, all `Sync` calls from the outside already respect this

pkg/receive/multitsdb.go
brancz: Done

pkg/receive/multitsdb.go
brancz: done

pkg/receive/multitsdb.go
brancz: done

pkg/receive/multitsdb.go
brancz: yes that was accidental, good catch

brancz: fixed

Makefile
```diff
@@ -252,7 +252,7 @@ test-e2e: docker
 	@echo ">> running /test/e2e tests."
 	# NOTE(bwplotka):
 	# * If you see errors on CI (timeouts), but not locally, try to add -parallel 1 to limit to single CPU to reproduce small 1CPU machine.
-	@go test -failfast -timeout 10m -v ./test/e2e/...
```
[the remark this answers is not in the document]
brancz: Nice, yes I like that as well. Done.

pkg/receive/multitsdb.go
```diff
@@ -0,0 +1,269 @@
...
+			t.tsdbCfg,
+		)
+
+		if err := s.Open(); err != nil {
+			level.Error(logger).Log("msg", "failed to open tsdb", "err", err)
+			t.mtx.Lock()
+			delete(t.appendables, tenantID)
+			delete(t.stores, tenantID)
+			t.mtx.Unlock()
+			if err := s.Close(); err != nil {
```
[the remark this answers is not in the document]
brancz: yes nice, done

pkg/receive/writer.go
```diff
@@ -21,29 +22,44 @@ type Appendable interface {
...
 
+type TenantStorage interface {
+	TenantAppendable(string) (Appendable, error)
+}
+
 type Writer struct {
-	logger log.Logger
-	append Appendable
+	logger    log.Logger
+	multitsdb TenantStorage
```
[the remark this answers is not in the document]
brancz: done

pkg/receive/writer.go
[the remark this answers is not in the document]
brancz: removed.

pkg/store/multitsdb.go
```diff
@@ -0,0 +1,284 @@
...
+	"github.com/prometheus/client_golang/prometheus"
+	"github.com/thanos-io/thanos/pkg/component"
+	"github.com/thanos-io/thanos/pkg/store/storepb"
+	"golang.org/x/sync/errgroup"
+	"google.golang.org/grpc"
+	"google.golang.org/grpc/codes"
+	"google.golang.org/grpc/status"
+)
+
+type MultiTSDBStore struct {
```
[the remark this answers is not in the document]
brancz: added comments, let me know if those are sufficient

pkg/store/multitsdb.go
brancz: done

pkg/store/multitsdb.go
brancz: removed

brancz:
@bwplotka addressed all your comments I think, or asked for clarification/verification on your end.

One more thing I'd like to add, but rather do in a follow up is the automatic migration from existing data dirs to the specified default tenant, that way people don't need to perform anything manually when rolling out this new version.

cmd/thanos/receive.go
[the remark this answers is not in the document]
bwplotka: :+1:

cmd/thanos/receive.go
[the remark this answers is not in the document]
bwplotka: :+1:

pkg/receive/handler.go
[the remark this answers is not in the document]
bwplotka: Yes that's what I meant. I was to lazy to pull code and check myself, and no vet will find the unused struct field (:

pkg/receive/multitsdb.go
```diff
@@ -0,0 +1,297 @@
...
+	tsdbCfg         *tsdb.Options
+	tenantLabelName string
+	labels          labels.Labels
+	bucket          objstore.Bucket
+
+	mtx     *sync.RWMutex
+	tenants map[string]*tenant
+}
+
+type tenant struct {
```
bwplotka, line 43: nice!

pkg/receive/multitsdb.go
```diff
@@ -0,0 +1,297 @@
...
+	tsdbCfg *tsdb.Options
+
+	readyS *tsdb.ReadyStorage
```
bwplotka, line 46: TBH all of those could be be embedded types, but not a big deal

bwplotka: so you would not need special methods for those like `storage`

bwplotka: Unless all of them have the same method like `close` ;p

pkg/receive/multitsdb.go
bwplotka: I guess that's good for a start.

pkg/receive/multitsdb.go
bwplotka: fair, now when I understood code it makes sense. But just out of nowhere for first time reader it can surprise (: Mabe just my English is broken

pkg/store/multitsdb.go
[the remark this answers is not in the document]
bwplotka: Yea, can see this. Cool let's start with this.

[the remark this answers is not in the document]
bwplotka: or you know at least some day we could reuse only info etc

pkg/store/multitsdb.go
bwplotka: ack :+1:

pkg/store/multitsdb.go
```diff
@@ -0,0 +1,275 @@
...
+			closeFn()
+		}()
+
+		for tenant, store := range stores {
+			store := store
+			seriesCtx, closeSeries := context.WithCancel(gctx)
+			seriesCtx = grpc_opentracing.ClientAddContextTags(seriesCtx, opentracing.Tags{
+				"tenant": tenant,
+			})
+			defer closeSeries()
```
bwplotka, line 185:
```suggestion
			defer cancelSeries()
```

As we usually cancel context not close (:

pkg/store/multitsdb.go
bwplotka: I disagree, they relate to different DBs, no?

pkg/receive/multitsdb.go
brancz: We need to protect reading them with the mutex, and we would have public functions that don't really make sense in their mix. I think the point is that it's encapsulated, and potentially we can expand this encapsulation to have the tenant object take care of writing/appending directly to rate limit etc.

pkg/store/multitsdb.go
```diff
@@ -178,12 +183,12 @@ func (s *MultiTSDBStore) Series(r *storepb.SeriesRequest, srv storepb.Store_Seri
...
 			store := store
-			seriesCtx, closeSeries := context.WithCancel(gctx)
+			seriesCtx, cancelSeries := context.WithCancel(gctx)
 			seriesCtx = grpc_opentracing.ClientAddContextTags(seriesCtx, opentracing.Tags{
 				"tenant": tenant,
 			})
-			defer closeSeries()
-			ss := newSeriesSetServer(seriesCtx, respSender)
+			defer cancelSeries()
+			ss := newTenantSeriesSetServer(seriesCtx, tenant, respSender)
```
bwplotka, line 191: What about `warnings` on series? are those wrapped with prefix?

brancz: yeah it was already the case, but slightly inconsistent (just generically wrapped the error and used the same in warnings and error handling, but that's not how we prefix other warnings in other places, so will do that instead)

brancz: done

pkg/receive/handler.go
```diff
@@ -526,6 +528,12 @@ func isConflict(err error) bool {
 		status.Code(err) == codes.AlreadyExists
 }
 
+// isNotReady returns whether or not the given error represents a not ready error.
+func isNotReady(err error) bool {
```
squat, line 532: :+1:

pkg/receive/multitsdb.go
```diff
@@ -0,0 +1,297 @@
...
+func (t *tenant) set(tstore *store.TSDBStore, fs *FlushableStorage, ship *shipper.Shipper) {
+	t.readyS.Set(fs.Get(), int64(2*time.Duration(t.tsdbCfg.MinBlockDuration).Seconds()*1000))
+	t.mtx.Lock()
+	t.fs = fs
+	t.s = tstore
+	t.ship = ship
+	t.mtx.Unlock()
+}
+
+func NewMultiTSDB(
```
squat, line 93: Nit: would be nice to document all of these exported types with comments, but we can do in a follow up

pkg/receive/multitsdb.go
```diff
@@ -0,0 +1,297 @@
...
+	for k, tenant := range t.tenants {
+		s := tenant.store()
+		if s != nil {
+			res[k] = s
+		}
+	}
+	return res
+}
+
+func (t *MultiTSDB) getOrLoadTenant(tenantID string) (*tenant, error) {
```
squat, line 216: nit: for clarity, consider calling this `getOrCreateTenant`

pkg/receive/writer.go
```diff
@@ -21,26 +22,38 @@ type Appendable interface {
 	Appender() (storage.Appender, error)
 }
 
+type TenantStorage interface {
+	TenantAppendable(string) (Appendable, error)
```
squat, line 26: nit: we could add the parameter name to help hint that the string is a tenant name

pkg/store/multitsdb.go
```diff
@@ -0,0 +1,285 @@
...
+	"context"
+	"fmt"
+	"sync"
+
+	"github.com/go-kit/kit/log"
+	grpc_opentracing "github.com/grpc-ecosystem/go-grpc-middleware/tracing/opentracing"
+	"github.com/opentracing/opentracing-go"
+	"github.com/pkg/errors"
+	"github.com/prometheus/client_golang/prometheus"
+	"github.com/thanos-io/thanos/pkg/component"
```
squat, line 16: completely a nit: as a follow up let's clean up these imports

cmd/thanos/receive.go
```diff
@@ -250,24 +274,13 @@ func runReceive(
...
 		cancel := make(chan struct{})
-		startTimeMargin := int64(2 * time.Duration(tsdbOpts.MinBlockDuration).Seconds() * 1000)
 		g.Add(func() error {
 			defer close(dbReady)
 			defer close(uploadC)
 
-			// Before actually starting, we need to make sure the
-			// WAL is flushed. The WAL is flushed after the
-			// hashring is loaded.
-			db := receive.NewFlushableStorage(
```
squat, line 261: We can ignore this initial flushing because now we know for sure that tenant data will not be mixed, whereas before it could have been :+1:

PR #2013, opened 2020-01-20
compact: implement delay to delete compacted block

khyatisoneji:
## Changes

Add delay to delete the compacted block from compaction

### DIFF
--- a/pkg/compact/compact.go
+++ b/pkg/compact/compact.go
@@ -42,6 +42,8 @@ type Syncer struct {
 	fetcher                  block.MetadataFetcher
 	mtx                      sync.Mutex
 	blocks                   map[ulid.ULID]*metadata.Meta
+	blocksToDelete           map[ulid.ULID]time.Time
+	consistencyDelay         int64
 	blockSyncConcurrency     int
 	metrics                  *syncerMetrics
 	acceptMalformedIndex     bool
@@ -120,7 +122,7 @@ func newSyncerMetrics(reg prometheus.Registerer) *syncerMetrics {
 
 // NewMetaSyncer returns a new Syncer for the given Bucket and directory.
 // Blocks must be at least as old as the sync delay for being considered.
-func NewSyncer(logger log.Logger, reg prometheus.Registerer, bkt objstore.Bucket, fetcher block.MetadataFetcher, blockSyncConcurrency int, acceptMalformedIndex bool, enableVerticalCompaction bool) (*Syncer, error) {
+func NewSyncer(logger log.Logger, reg prometheus.Registerer, bkt objstore.Bucket, fetcher block.MetadataFetcher, blockSyncConcurrency int, consistencyDelay int64, acceptMalformedIndex bool, enableVerticalCompaction bool) (*Syncer, error) {
 	if logger == nil {
 		logger = log.NewNopLogger()
 	}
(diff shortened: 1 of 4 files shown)


khyatisoneji:
I would like to know If I am going in the correct direction with this implementation.
I wanted to add a filter to remove the older blocks only after consistencyDelay.

Since metaFetcher is used by syncer, I had to create an AddFilter function to add filter to metaFetcher.

Please add your comments if there are any changes required.

khyatisoneji:
I was thinking of adding a map of ULID to the time the block was marked for deletion.
This map would be shared between all the components that call `block.Delete`, the problem here is that since this map is in memory if the compact process restarts, we lose this information in the map.
So after the process restart, the compact process would delete the block directly instead of waiting for `consistencyDelay` time. This would result in the same state which is happening now that the store component may try to retrieve a block that is deleted.

Please share some points on how we can make it resilient to such scenarios.

pkg/compact/compact.go
```diff
@@ -326,14 +330,34 @@ func (s *Syncer) garbageCollect(ctx context.Context, resolution int64) error {
 			return retry(errors.Wrapf(err, "delete block %s from bucket", id))
 		}
 
-		// Immediately update our in-memory state so no further call to SyncMetas is needed
-		// after running garbage collection.
-		delete(s.blocks, id)
+		// Mark the blocks as to be removed from the blocks map so that
+		// it is not considered for the next round of compaction.
+		s.blocksToDelete[id] = time.Now().UTC()
 		s.metrics.garbageCollectedBlocks.Inc()
```
bwplotka: Is it really garbage collected? We might think of reworking metrics in our case.

pkg/compact/compact.go
```diff
@@ -326,14 +330,34 @@ func (s *Syncer) garbageCollect(ctx context.Context, resolution int64) error {
 			return retry(errors.Wrapf(err, "delete block %s from bucket", id))
 		}
 
-		// Immediately update our in-memory state so no further call to SyncMetas is needed
-		// after running garbage collection.
-		delete(s.blocks, id)
+		// Mark the blocks as to be removed from the blocks map so that
+		// it is not considered for the next round of compaction.
+		s.blocksToDelete[id] = time.Now().UTC()
```
bwplotka: Something is wrong here.. We already delete the block (above), so why this is map is named `to Delete`? (:

bwplotka: So I believe we are actually deleting it immediately so we don't delay anything at the moment in this PR :thinking: We might want to change this.

pkg/compact/compact.go
```diff
@@ -326,14 +330,34 @@ func (s *Syncer) garbageCollect(ctx context.Context, resolution int64) error {
...
+		}
+
+		// Wait for consistencyDelay for the compacted block to be successfully uploaded to bucket.
+		if ulid.Now()-ulid.Timestamp(blockMarkedTime) < uint64(s.consistencyDelay) {
+			level.Debug(s.logger).Log("msg", "block has been marked for deletion", "block", id)
+			synced.WithLabelValues(block.MarkedForDeletion).Inc()
+			delete(metas, id)
+		} else {
+			// If the compacted block is successfully uploaded, remove the old block from the group.
+			delete(s.blocks, id)
```
bwplotka: So we remove this from group.. but don't we want actually want to remove the block from the bucket? Also doing here is weird. It is a filter of metaFetcher. Filter can be run any time. I believe meta Fetcher filter might be capable to NOT return blocks that are scheduled to be removed, but BASED on Source ULID on meta.json not on some map stored in memory. This is because when compactor crashes this information will be lost.

Let's talk about this offline we can setup a call (:

khyatisoneji:
Closing this PR since the approach is not resilient to failures. Will work on a new PR which has the changes as discussed

PR #2015, opened 2020-01-20
Fix invalid start/end params in series api

### LINKED ISSUE #2011: query: error in #1904 - /series with SkipChunks not use start/end args.
thanos v0.10.0 with #1904 

**Object Storage Provider**:

**What happened**:
/series return all data (not considering selected time range in Grafana)

Return data only in selected time range

Request /series to prometheus with start/end args and compare with same request to thanos-query

I try to set `SkipChunks=false` and get vallid response

yeya24:

## Changes

Fixes #2011

cc @IKSIN  @GiedriusS 
## Verification

I have already tested both in Grafana and local API request.

```
curl -g 'http://localhost:10908/api/v1/series?' -d 'match[]={__name__="up"}'
{"status":"success","data":[{"__name__":"up","instance":"localhost:9090","job":"prometheus","replica":"test"},{"__name__":"up","instance":"localhost:9091","job":"prometheus","replica":"test"}]}

curl -g 'http://localhost:10908/api/v1/series?' -d 'match[]={__name__="up"}' -d 'start=1579551916'   
{"status":"success","data":[{"__name__":"up","instance":"localhost:9091","job":"prometheus","replica":"test"}]}
```

### DIFF
--- a/pkg/store/prometheus.go
+++ b/pkg/store/prometheus.go
@@ -12,8 +12,10 @@ import (
 	"net/url"
 	"path"
 	"sort"
+	"strconv"
 	"strings"
 	"sync"
+	"time"
 
 	"github.com/go-kit/kit/log"
 	"github.com/go-kit/kit/log/level"
@@ -23,6 +25,7 @@ import (
 	"github.com/pkg/errors"
 	"github.com/prometheus/common/version"
 	"github.com/prometheus/prometheus/pkg/labels"
+	"github.com/prometheus/prometheus/pkg/timestamp"
 	"github.com/prometheus/prometheus/prompb"
 	"github.com/prometheus/prometheus/storage/remote"
 	"github.com/prometheus/prometheus/tsdb/chunkenc"
@@ -663,9 +666,9 @@ func (p *PrometheusStore) seriesLabels(ctx context.Context, matchers []storepb.L
 	}
 
 	q.Add("match[]", metric)
+	q.Add("start", formatTime(timestamp.Time(startTime)))
+	q.Add("end", formatTime(timestamp.Time(endTime)))
 	u.RawQuery = q.Encode()
-	q.Add("start", string(startTime))
-	q.Add("end", string(endTime))
 
 	req, err := http.NewRequest("GET", u.String(), nil)
 	if err != nil {
@@ -714,4 +717,8 @@ func (p *PrometheusStore) seriesLabels(ctx context.Context, matchers []storepb.L
 	}
 
 	return m.Data, nil
+}
+
+func formatTime(t time.Time) string {
+	return strconv.FormatFloat(float64(t.Unix())+float64(t.Nanosecond())/1e9, 'f', -1, 64)
 }
--- a/pkg/model/timeduration.go
+++ b/pkg/model/timeduration.go
@@ -42,7 +42,7 @@ func (tdv *TimeOrDurationValue) Set(s string) error {
 	return nil
 }
 
-// String returns either tume or duration.
+// String returns either time or duration.
 func (tdv *TimeOrDurationValue) String() string {
 	switch {
 	case tdv.Time != nil:
--- a/pkg/store/prometheus_test.go
+++ b/pkg/store/prometheus_test.go
@@ -199,6 +199,8 @@ func TestPrometheusStore_SeriesLabels_e2e(t *testing.T) {
 	testutil.Ok(t, err)
 	_, err = a.Add(labels.FromStrings("a", "d", "job", "test"), baseT+300, 3)
 	testutil.Ok(t, err)
+	_, err = a.Add(labels.FromStrings("job", "test"), baseT+400, 4)
+	testutil.Ok(t, err)
 	testutil.Ok(t, a.Commit())
 
 	ctx, cancel := context.WithCancel(context.Background())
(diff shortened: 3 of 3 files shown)


pkg/store/prometheus.go
```diff
@@ -663,9 +666,9 @@ func (p *PrometheusStore) seriesLabels(ctx context.Context, matchers []storepb.L
 	}
 
 	q.Add("match[]", metric)
+	q.Add("start", formatTime(timestamp.Time(startTime)))
```
yeya24: I am not sure if there is a better way to format the timestamp from int64 directly? I am not quite familiar with this

bwplotka: Looks like time.RFC3339Nano is supported so maybe `startTime.Format(time.RFC3339Nano)`?

yeya24: I have tried several times but still doesn't work, the generated timestamp is not valid in Prometheus side. Are you OK with the current implementation? @bwplotka  
I think there is no big difference between the two ways.

bwplotka: It would be super nice to make it valid, I can't see reason why it would not work:

The code is literally this:
```
func parseTime(s string) (time.Time, error) {
	if t, err := strconv.ParseFloat(s, 64); err == nil {
		s, ns := math.Modf(t)
		ns = math.Round(ns*1000) / 1000
		return time.Unix(int64(s), int64(ns*float64(time.Second))), nil
	}
	if t, err := time.Parse(time.RFC3339Nano, s); err == nil {

		return t, nil
	}

	// Stdlib's time parser can only handle 4 digit years. As a workaround until
	// that is fixed we want to at least support our own boundary times.
	// Context: https://github.com/prometheus/client_golang/issues/614
	// Upstream issue: https://github.com/golang/go/issues/20555
	switch s {
	case minTimeFormatted:
		return minTime, nil
	case maxTimeFormatted:
		return maxTime, nil
	}
	return time.Time{}, errors.Errorf("cannot parse %q to a valid timestamp", s)
}
```

so you can test even locally for that.

bwplotka: https://github.com/prometheus/prometheus/blob/9d9bc524e5e893287cfeeab9995252f54b0624a8/web/api/v1/api.go#L1400

yeya24: Thanks for your help! But unfortunately, it still doesn't work properly. I use `timestamp.Time(startTime).Format(time.RFC3339Nano)`.
If I specify a start time, I will get the result:
``` go
➜  ~ curl -g 'http://localhost:10908/api/v1/series?' -d 'match[]={__name__="up"}' -d 'start=1579551916'
{"status":"success","data":[{"__name__":"up","instance":"localhost:9090","job":"prometheus","replica":"test"},{"__name__":"up","instance":"localhost:9091","job":"prometheus","replica":"test"}]}
```
But If I dont specify any time range, it will use the PrometheusStore.minTime as the start time, this seems invalid in Prometheus side
``` go
➜  ~ curl -g 'http://localhost:10908/api/v1/series?' -d 'match[]={__name__="up"}'                     
{"status":"success","data":[],"warnings":["receive series from Addr: localhost:10905 LabelSets: [name:\"replica\" value:\"test\" ] Mint: -62167219200000 Maxt: 9223372036854775807: rpc error: code = Internal desc = request Prometheus server failed, code 400 Bad Request"]}
```

yeya24: Thanks for your help! But unfortunately, it still fails in some cases. I am not sure what's wrong. I used `timestamp.Time(startTime).Format(time.RFC3339Nano)`.
If I specify a start time and end time, I will get the result, which means the RFC3339 stamp works:
``` go
➜  ~ curl -g 'http://localhost:10908/api/v1/series?' -d 'match[]={__name__="up"}' -d 'start=1579551916' -d 'end=1579600000'
{"status":"success","data":[{"__name__":"up","instance":"localhost:9091","job":"prometheus","replica":"test"}]}
```
But If I dont specify any time range, it will use the PrometheusStore.minTime and maxTime as the start time and end time, this seems invalid in Prometheus side. How to handle this?
``` go
➜  ~ curl -g 'http://localhost:10908/api/v1/series?' -d 'match[]={__name__="up"}'
{"status":"success","data":[],"warnings":["receive series from Addr: localhost:10905 LabelSets: [name:\"replica\" value:\"test\" ] Mint: -62167219200000 Maxt: 9223372036854775807: rpc error: code = Internal desc = request Prometheus server failed, code 400 Bad Request"]}
```

FUSAKLA: Hm, interesting. I tried it on your branch with using the `timestamp.Time(startTime).Format(time.RFC3339Nano)` formatting and it works correctly with and without the time range :thinking:

bwplotka: Interesting.. looks like this means we specify maxInt64 and minInt64 then. 

Maybe checking if it is maxInt then skipping specifying end and same if start is minInt then skip specifying start?

GiedriusS: @yeya24 maybe let's add another test-case which would test this out so that we would know for sure if this is reproducible?

yeya24: @GiedriusS  Added one. From that testcase, using `timestamp.Time(startTime).Format(time.RFC3339Nano)` will fail. I think it is OK to use the current one.

GiedriusS: It is unfortunate that Golang's stdlib cannot parse all of the values that one can format in `time.RFC3339Nano`: https://github.com/golang/go/issues/20555. However, I'd vote for @bwplotka's suggestion here: maybe lets just not add these parameters if they haven't been specified? Prometheus then handles all of this accordingly. I think we'd avoid more problems down the line. WDYT?

sylr:
I've built this PR and unfortunately it did not fix the problem for me.

Grafana queries that populate variables returned more items that they should have.

pkg/store/prometheus.go
yeya24: Thanks for pointing this out! However, I am still not sure about that edge case. Currently the minTime of sidecar can be configured via a flag. But its default value `"0000-01-01T00:00:00Z"` is invalid at the Prometheus side. How to avoid this case? And what if users set some other time that cannot be parsed as well?

yeya24:
> I've built this PR and unfortunately it did not fix the problem for me.
> 
> Grafana queries that populate variables returned more items that they should have.

Hi, @sylr 
Could you please share with me how you test this?
I am not quite familiar with Grafana and I only found the **variables** page calls the **series** API.

Recent 30 min of label_values(up, instance)
image

Several days ago data of label_values(up, instance). The result is different from the previous one.
image

bwplotka:
Thank you @yeya24 for quick reaction!

sylr:
Ok so I'd like to correct my previous comment as I thought the problem was located only on the querier, it was not according to my new observations:

`0.10.0` on the sidecar and `0.10.0` on the querier -> NOT OK
`0.10.0` on the sidecar and `0.8.1` on the querier -> OK
`master+this` on the sidecar and `0.10.0` on the querier -> NOT OK
`master+this` on the sidecar and `master+this` on the querier -> OK
`0.10.0` on the sidecar and `master+this` on the querier -> NOT OK

PR #2017, opened 2020-01-21
mixin/thanos: fix typo in alert name

simonpasquier:


squat:
Seems we got a flake. Re-running

simonpasquier:
> Seems we got a flake

I'll investigate it as it's not the first time I see this and it's in the Ruler e2e tests.

PR #2021, opened 2020-01-21
Add _total suffix to two counter metrics

yeya24:

## Changes

I guess we need `_total` in counter metrics


yeya24:
Sure. I will add the changelog. But I didn't find any related rules in the mixin. And couldn't find that in the kube-thanos repo as well

Issue #1611, opened 2019-10-08
Reduce unnecessary StoreAPI series calls


brancz:
Thanos currently performs a "possibly in set" check in order to exclude certain StoreAPIs from calls. This check currently only excludes a StoreAPI if all sets of external labels indicate that time-series selected by a certain query will never include data from this StoreAPI. While this check only has true positives, the potential for empty responses is still large, as it requires the query to truly select on the labels that a StoreAPI happens to expose.

In order to expand these "possibly in set" checks, I propose to have StoreAPIs additionally expose a probabilistic datastructure(s) via the `Info` API next to the sets of external labels. Datastructures such as [Bloom filters](https://en.wikipedia.org/wiki/Bloom_filter) or [Cuckoo filters](https://en.wikipedia.org/wiki/Cuckoo_filter), could be a good candidate. This way we could in a space efficient way minimize unnecessary calls to StoreAPIs.

This would not only improve performance, but also reliability, as any additional and unnecessary network request has the potential to cause what seems like a partial response even when it's not.

In order to verify further, that this would be useful, I would propose to instrument the querier with metrics to verify that these types of empty StoreAPI series requests are actually happening and therefore impacting performance, otherwise this optimization is likely not worth it.

@bwplotka @domgreen @GiedriusS @metalmatze @squat @kakkoyun

GiedriusS:
In theory, it sounds good but I wonder about the performance implications of keeping that bloom map up-to-date, especially in the case of Sidecar/Prometheus. This probably makes the most sense in environments where are a bunch of disparate Thanos Query instances and there is one "on top" which is being queried.

bwplotka:
In my eyes there will just small and very often used portion of labels maintained by such filter, so it should be fine.

brancz:
I think we have yet to instrument the proxy store type (used with the query component) to validate whether and how much this is happening. I do suspect that it will highly vary across environments.

brancz:
With #2030 merged we should be able to at least identify the potential for this.

jojohappy:
Maybe it's still valid.

brancz:
Looking at the metrics in our prod environment, the metrics seem to show that we do _a lot_ of requests that return empty responses.

brancz:
As mentioned in the previous comment, there is definitely room for improvement still.

GiedriusS:
Still valid.

yeya24:
[...]

PR #2030, opened 2020-01-22
pkg/store: Expose metric about empty stream responses in proxy store

brancz:

## Changes

This adds a metric to track how often we do requests that end up in empty responses. Related to #1611.

## Verification

Did queries that stores didn't have data for, and verified that indeed the `/metrics` endpoint reflects this.

@bwplotka @squat @GiedriusS

### DIFF
--- a/pkg/store/proxy.go
+++ b/pkg/store/proxy.go
@@ -15,6 +15,7 @@ import (
 	grpc_opentracing "github.com/grpc-ecosystem/go-grpc-middleware/tracing/opentracing"
 	"github.com/opentracing/opentracing-go"
 	"github.com/pkg/errors"
+	"github.com/prometheus/client_golang/prometheus"
 	"github.com/prometheus/prometheus/pkg/labels"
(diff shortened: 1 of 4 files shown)


pkg/store/proxy.go
```diff
@@ -348,6 +374,12 @@ func startStreamSeriesSet(
 		defer wg.Done()
 		defer close(s.recvCh)
 
+		i := 0
```
GiedriusS: I feel like a more descriptive name like `responses` would be better. I am way too used to seeing `i` as a placeholder variable in for-loops for the index :smile: Here we do an infinite loop so it probably doesn't fit as well.

brancz: agreed

GiedriusS:
There was an attempt to add a similar thing here as well: https://github.com/thanos-io/thanos/pull/1789

GiedriusS:
> Giedrius, not really connected to other issues, it's not about slow stores but rather stores that were asked for data that they don't have.

Hence the word 'similar' :smile:

PR #2035, opened 2020-01-23
Cut 0.10.1

bwplotka:
@GiedriusS reverted the fix metric name regression in examples as they are really ONLY examples, so no point in adding them to patch release IMO (:


brancz:
I'd be ok with keeping the mixin part of releases, but don't feel too strongly.

metalmatze:
Having the mixin as part of releases would definitely elevate their importance and thus hopefully increase their quality. Afterall those alerts are what people are going to rely on moving forward.

bwplotka:
Why? No one look on those on release branch - everyone do that on the master, right :thinking:

brancz:
We tend to use the same mixin version of etcd, ceph, etc. as the version of the software itself. I feel it would be useful to version along, then we can try to do changed to the mixin whenever we do changes on master, that way it's easy to know which mixin is compatible with which version.

VERSION
```diff
@@ -1 +1 @@
-0.10.0
```
squat: :p

bwplotka:
Fair points @brancz @metalmatze let's touch it on our Thanos sync. We need to decide then if e.g we make a mistake and discrepancy between mixin and code appear, will it require patch release?

PR #2047, opened 2020-01-26
Fix bucket e2e test

### LINKED ISSUE #2046: Flaky Thanos Bucket Store E2E test
See https://circleci.com/gh/thanos-io/thanos/6803, for this error, we don't need to check the number of chunks of series response for each testcase, we just need to make sure the number of chunks is 0 when `skipChunks` option set to true.

yeya24:

## Changes

Fixes #2046 
As I mentioned in #2046, we just need to check the chunk length when `SkipChunks` set to true.

### DIFF
--- a/pkg/store/bucket_e2e_test.go
+++ b/pkg/store/bucket_e2e_test.go
@@ -191,9 +191,8 @@ func testBucketStore_e2e(t *testing.T, ctx context.Context, s *storeSuite) {
 
 	// TODO(bwplotka): Add those test cases to TSDB querier_test.go as well, there are no tests for matching.
 	for i, tcase := range []struct {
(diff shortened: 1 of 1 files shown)


bwplotka:
Thanks for this but I disagree with that PR. (: I think it's nice to check number of chunks. And for SkipChunks we just expect 0 which is nice to know if SkipChunks works or not. (: 

Also how it fixes the flakiness?

yeya24:
> Thanks for this but I disagree with that PR. (: I think it's nice to check number of chunks. And for SkipChunks we just expect 0 which is nice to know if SkipChunks works or not. (:
> 
> Also how it fixes the flakiness?

Thanks for the review. I will investigate more and find a better way to fix it if the flaky test is valid.

Issue #2039, opened 2020-01-24
Sampling for Elastic APM tracing


igorwwwwwwwwwwwwwwwwwwww:
This may change with the switch to OpenTelemetry (https://github.com/thanos-io/thanos/issues/1972).

It looks like the Elastic APM tracing backend in thanos does not currently support sampling. Without sampling, every request will be traced, which can be quite expensive.

The Elastic APM client library _does_ support sampling via [`Tracer.SetSampler`](https://godoc.org/go.elastic.co/apm#Tracer.SetSampler), so adding support for that might not be that hard.

PR #2049, opened 2020-01-27
tracing: Support sampling on Elastic APM

igorwwwwwwwwwwwwwwwwwwww:
This is a patch to address thanos-io/thanos#2039.

Tracing every request can be quite expensive. By introducing a sample rate (similar to what stackdriver and jaeger already support), the overhead and trace storage cost can be controlled.

This patch adds sampling for Elastic APM tracing.

## Changes

* tracing: Support sampling on Elastic APM with new `sample_rate` setting

## Verification

* I still need to run a proper integration test. Is there a standalone-ish way of running thanos locally that would enable such testing?
* Since the default value for sample rate is `0`, this will break anyone currently using the elastic APM backend. It might make sense to have a proper default value of `1` -- this applies to various settings on other tracing backends.


igorwwwwwwwwwwwwwwwwwwww:
Managed to verify that this works.

Surprisingly, the APM library will still emit empty traces for non-sampled events (basically a root span without children), but that appears to be by design.

PR #2051, opened 2020-01-27
Pass prom registry to shipper from ruler.

### LINKED ISSUE #2050: Ruler does not expose shipper metrics

0.10.1

**Object Storage Provider**: 
GCS

**What happened**:
Ruler failed to upload to GCS bucket, by reading the code, it should increase and expose the following metric:

https://github.com/thanos-io/thanos/blob/4a23b36a017961a2ab6b78c4ca585a680d922a0d/pkg/shipper/shipper.go#L338-L340

However, this does not happen due to missing prom register:
https://github.com/thanos-io/thanos/blob/4a23b36a017961a2ab6b78c4ca585a680d922a0d/cmd/thanos/rule.go#L632

Ruler to expose this metric.

1. run ruler with bad bucket config 
2. let the upload fail
3. check if metric is exposed, this should not be the case

OGKevin:
https://github.com/thanos-io/thanos/issues/2050
Fixes #2050


OGKevin:
@bwplotka  Its ready for review now 😊

PR #2057, opened 2020-01-28
Store: Document --chunk-pool-size better

### LINKED ISSUE #2042: Store: Document --chunk-pool-size better
It's not immediately apparent if the chunk pool size is a in-memory cache of the chunks, or if it's an on disk cache. The index cache option was updated and clearly states it's an in-memory cache whereas chunk pool is ambiguous.

daixiang0:
Fix #2042

### DIFF
--- a/cmd/thanos/store.go
+++ b/cmd/thanos/store.go
@@ -47,7 +47,7 @@ func registerStore(m map[string]setupFunc, app *kingpin.Application) {
 		"YAML file that contains index cache configuration. See format details: https://thanos.io/components/store.md/#index-cache",
 		false)
 
-	chunkPoolSize := cmd.Flag("chunk-pool-size", "Maximum size of concurrently allocatable bytes for chunks.").
+	chunkPoolSize := cmd.Flag("chunk-pool-size", "Maximum size of concurrently allocatable bytes reserved strictly to reuse for chunks in memory.").
 		Default("2GB").Bytes()
 
 	maxSampleCount := cmd.Flag("store.grpc.series-sample-limit",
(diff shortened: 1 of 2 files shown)


daixiang0:
@squat how about `Maximum size of concurrently allocatable bytes for chunks in memory, not cache.`?

GiedriusS:
> @squat how about `Maximum size of concurrently allocatable bytes for chunks in memory, not cache.`?

Not sure about the `, not cache.` part because I can imagine there will be a question about what kind of cache are we exactly talking about but otherwise sounds good. It seems to me like the goal here is to put emphasis on the fact that we are talking about the memory in general. Let's see what @squat and others think.

squat:
I agree with @GiedriusS :+1:

daixiang0:
I think it means some fixed memory during thanos running, not the cache memory.

bwplotka:
That's the problem @daixiang0. It's not a fixed memory. It does not allocate that much as you put, it's only use a maximum of configured memory for reusable bytes pool. And if it will go beyond that for chunks it will fail the request. In theory. In practice, I really could not see our:

```
// ErrPoolExhausted is returned if a pool cannot provide the request bytes.
var ErrPoolExhausted = errors.New("pool exhausted")
```

But that's another story (:

daixiang0:
@bwplotka does it mean that memory leak or the speed of system clean cache memory is faster than thanos consumes?

bwplotka:
No, it means opposite: garbage collection (cleaning) is slower than our allocations, even though we don't need some memory. That's why we pool memory.

PR #2063, opened 2020-01-28
Added Copyright header to all Go and Proto source files.

bwplotka:
As per: https://github.com/cncf/foundation/blob/master/copyright-notices.md#copyright-notices

### DIFF
--- a/pkg/testutil/sysprocattr.go
+++ b/pkg/testutil/sysprocattr.go
@@ -1,3 +1,6 @@
+// Copyright (c) The Thanos Authors.
+// Licensed under the Apache License 2.0.
+
 // +build !linux
 
 package testutil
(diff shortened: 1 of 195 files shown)


pkg/testutil/sysprocattr.go
```diff
@@ -1,3 +1,6 @@
+// Copyright (c) The Thanos Authors.
+// Licensed under the Apache License 2.0.
+
 // +build !linux
```
squat: This is causing format errors in the file now :///

```txt


./pkg/testutil/sysprocattr.go:

// +build !linux

package testutil



 Error: Found comments without trailing period. Comments has to be full sentences.


make: *** [Makefile:195: check-comments] Error 1
```

bwplotka:
PTAL @squat

PR #2065, opened 2020-01-28
mixin, docs: Rename Compact -> Compactor

kakkoyun:
This PR renames `compact` to `compactor` in mixin and docs.

As discussed in https://github.com/thanos-io/thanos/issues/1871

## Verification

* `make examples`
* `make example-rules-lint`
* `make docs`
* `make check-docs`

### DIFF
--- a/examples/alerts/alerts.yaml
+++ b/examples/alerts/alerts.yaml
@@ -3,7 +3,7 @@ groups:
   rules:
   - alert: ThanosCompactorMultipleCompactsAreRunning
     annotations:
-      message: You should never run more than one Thanos Compact at once. You have
+      message: You should never run more than one Thanos Compactor at once. You have
         {{ $value }}
     expr: sum(up{job=~"thanos-compactor.*"}) > 1
     for: 5m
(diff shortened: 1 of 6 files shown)


examples/alerts/alerts.yaml
```diff
@@ -3,23 +3,23 @@ groups:
   rules:
   - alert: ThanosCompactorMultipleCompactsAreRunning
```
yeya24: Is it better to use `ThanosCompactorMultipleCompactorsAreRunning`

mixin/thanos/alerts/compactor.libsonnet
```diff
@@ -12,7 +12,7 @@
           {
             alert: 'ThanosCompactorMultipleCompactsAreRunning',
```
yeya24: Same here

kakkoyun:
@yeya24 From the discussions on the issue, I got the feeling that we won't change metric names yet.
I might be wrong.

@bwplotka What do you think? Is it safe to rename metrics at this stage?

Issue #1708, opened 2019-11-01
Move retention apply to the beginning of Compactor loop 


bwplotka:
We apply retention at the end, we probably should move it before any compaction. It was by design to make sure lower resolution can be downsampled, but TBH it does not really make much sense. If you configure the retention in a way that raw goes away too quickly this can hit you anyway. So I cannot see the blocker now (: 

We can save a lot especially if retention is desired to be e.g 3 months but we left the data without compactor for 6 months (: We can avoid unnecessary compactions and downsampling.

AC: 
* Move retention applies to the beginning of Thanos compact loop.

GiedriusS:
>  If you configure the retention in a way that raw goes away too quickly this can hit you anyway.

That's true but I would really love to make Thanos Compact scream loudly if this happens as it indicates some kind of error, no? Especially if downsampling is enabled but the retention policies prevent us from downsampling data.

bwplotka:
I think we might need to make it a bit more complex.

Let's consider this case:
* raw retention is 1month
* downsampled retention is 2y
* compactor is not running for 2 months.

Now what will happen with this change is that we will have a gap for downsampled data. between `[2mo, 1mo]` as we remove raw data that was dependent on. 

I think we could improve the flow only for case where raw = res5m = res1h retention is equal. In this case, removal on the start might make sense. 

Or in other words: We can only remove up to res1h retention at the start. Does it make sense? Other opinions are welcome (:

GiedriusS:
Reopening since there is a new PR which tries to address this.

kakkoyun:
Still ongoing. Let's keep it fresh for a little more.

GiedriusS:
In progress.

AllenzhLi:
Maybe this issue should be more attention.

heyitsmdr:
Is this still on-going by any chance?

yeya24:
This can also address the problem mentioned in https://github.com/thanos-io/thanos/pull/4406

PR #2068, opened 2020-01-28
Move retention at the start of the compactor loop.

obiesmans:
Signed-off-by: Olivier Biesmans <olivier.biesmans@blablacar.com>

This PR tries to address issue #1708, with the fixes discussed on [#1708](https://github.com/thanos-io/thanos/issues/1708#issuecomment-553981466)

Feedback is welcome, especially on how to test those changes.

 I added CHANGELOG entry for this change.
 Change is not relevant to the end user

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -81,7 +81,7 @@ anymore.
 - [#2450](https://github.com/thanos-io/thanos/pull/2450) Store: regex-set optimization for `label=~"a|b|c"` matchers.
 - [#2475](https://github.com/thanos-io/thanos/pull/2475) Store: remove incorrect optimizations for queries with `=~".*"` and `!=~".*"` matchers.
 
-- [#1735](https://github.com/thanos-io/thanos/pull/1735) Retention policies are now applied at the start of the compactor loop. Retentions policies that break downsampling are now logged as warnings.
+- [#1735](https://github.com/thanos-io/thanos/pull/1735) Retention policies are now applied at the start of the compactor loop. Retention policies that break downsampling are now logged as warnings.
 
 ## [v0.12.0](https://github.com/thanos-io/thanos/releases/tag/v0.12.0) - 2020.04.15
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -81,6 +81,8 @@ anymore.
 - [#2450](https://github.com/thanos-io/thanos/pull/2450) Store: regex-set optimization for `label=~"a|b|c"` matchers.
 - [#2475](https://github.com/thanos-io/thanos/pull/2475) Store: remove incorrect optimizations for queries with `=~".*"` and `!=~".*"` matchers.
 
+- [#1735](https://github.com/thanos-io/thanos/pull/1735) Retention policies are now applied at the start of the compactor loop. Retentions policies that break downsampling are now logged as warnings.
+
 ## [v0.12.0](https://github.com/thanos-io/thanos/releases/tag/v0.12.0) - 2020.04.15
 
 ### Fixed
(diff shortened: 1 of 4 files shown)


cmd/thanos/compact.go
```diff
@@ -306,6 +324,7 @@ func runCompact(
 		if err := compact.ApplyRetentionPolicyByResolution(ctx, logger, bkt, retentionByResolution); err != nil {
 			return errors.Wrap(err, fmt.Sprintf("retention failed"))
 		}
+
```
daixiang0: Mind delete this new line?

obiesmans: not at all :)

cmd/thanos/compact.go
```diff
@@ -268,15 +268,33 @@ func runCompact(
 
 	if retentionByResolution[compact.ResolutionLevelRaw].Seconds() != 0 {
 		level.Info(logger).Log("msg", "retention policy of raw samples is enabled", "duration", retentionByResolution[compact.ResolutionLevelRaw])
+		if retentionByResolution[compact.ResolutionLevelRaw].Seconds() <= downsample.DownsampleRange0 {
+			level.Warn(logger).Log("msg", "retention policy of raw samples is too low to generate 5 min aggregated samples", "duration", retentionByResolution[compact.ResolutionLevelRaw])
```
GiedriusS: This is good :+1:

cmd/thanos/compact.go
```diff
@@ -268,15 +268,33 @@ func runCompact(
...
+			level.Warn(logger).Log("msg", "retention policy of 5 min aggregated samples is too low to generate 1h aggregated samples", "duration", retentionByResolution[compact.ResolutionLevel5m])
+		}
 	}
 	if retentionByResolution[compact.ResolutionLevel1h].Seconds() != 0 {
 		level.Info(logger).Log("msg", "retention policy of 1 hour aggregated samples is enabled", "duration", retentionByResolution[compact.ResolutionLevel1h])
 	}
 
 	f := func() error {
+		// At first, we need at least to be able to genetes 1h resolution metrics
+		// from the lower resolutions, so remove only up to 1h resolution retention.
```
GiedriusS: I feel like there is an implicit assumption here that retention times are non-decreasing and "correct". What if the user sets the 1h retention to some very small value? Won't we delete some data that we are not supposed to? At least right now we would generate it but, of course, we would delete it immediately :smile: 

Also, I thought we were going to only apply this at the beginning if the retention times are configured to be the same as otherwise we run the risk of losing data. Is there something missing as to why you are avoiding that check? :smile: 

How did you test this out?

obiesmans: Nice catch. I'll add the cases of non-decreasing somewhere, and add tests as well.

obiesmans:
Well, sorry I haven't found time to work on this yet ; I'll try to do this next week !

CHANGELOG.md
```diff
@@ -137,6 +139,7 @@ Compactor now properly handles partial block uploads for all operation like rete
 
 - [#1852](https://github.com/thanos-io/thanos/pull/1852) Add support for `AWS_CONTAINER_CREDENTIALS_FULL_URI` by upgrading to minio-go v6.0.44
 - [#1854](https://github.com/thanos-io/thanos/pull/1854) Update Rule UI to support alerts count displaying and filtering.
+<<<<<<< HEAD
```
GiedriusS, line 142: Seems like there's a remnant here of a merge conflict being solved. Could you remove this?

CHANGELOG.md
```diff
@@ -20,6 +20,8 @@ We use *breaking* word for marking changes that are not backward compatible (rel
 
 ### Changed
 
+- [#1735](https://github.com/thanos-io/thanos/pull/1735) Retention policies are now applied at the start of the compactor loop. Retentions policies that break downsampling are now logged as warnings.
```
GiedriusS, line 23:
```suggestion
- [#1735](https://github.com/thanos-io/thanos/pull/1735) Retention policies are now applied at the start of the compactor loop. Retention policies that break downsampling are now logged as warnings.
```

pkg/compact/retention.go
```diff
@@ -17,6 +18,8 @@ import (
 	"github.com/thanos-io/thanos/pkg/objstore"
 )
 
+type RetentionPolicy map[ResolutionLevel]time.Duration
```
GiedriusS, line 21: Could we add a comment for this exported type?

obiesmans: Will do !

CHANGELOG.md
obiesmans: Ooopsie !

PR #2069, opened 2020-01-28
Lower restriction for website scripts.

bwplotka:
`Refused to execute inline script because it violates the following Content Security Policy directive: "script-src thanos.io". Either the 'unsafe-inline' keyword, a hash ('sha256-3qFt4qPvMCWVUpjUxP5X57GBKae6RHYZ0rMjn9WuNF4='), or a nonce ('nonce-...') is required to enable inline execution.`

Fixed by not restricting by domain, just TLS.


bwplotka:
I mean I could enumerate 10 different domains.. or download those scripts. Something to later on. Let's fix the website now.

PR #2071, opened 2020-01-28
fetcher: add filter to remove old blocks if new overlapping blocks found

khyatisoneji:
## Changes

Add Filter to filter out old blocks if new overlapping blocks found that fully submatch the old blocks

Please let me know if this is the right direction to go in terms of filtering out the old blocks

### DIFF
--- a/pkg/block/fetcher.go
+++ b/pkg/block/fetcher.go
@@ -11,6 +11,7 @@ import (
 	"os"
 	"path"
 	"path/filepath"
+	"sort"
 	"sync"
 	"time"
 
@@ -21,6 +22,7 @@ import (
 	"github.com/prometheus/client_golang/prometheus"
 	"github.com/prometheus/prometheus/pkg/labels"
 	"github.com/prometheus/prometheus/pkg/relabel"
+	"github.com/prometheus/prometheus/tsdb"
 	tsdberrors "github.com/prometheus/prometheus/tsdb/errors"
 	"github.com/prometheus/prometheus/tsdb/fileutil"
 	"github.com/thanos-io/thanos/pkg/block/metadata"
@@ -50,6 +52,7 @@ const (
 	labelExcludedMeta = "label-excluded"
 	timeExcludedMeta  = "time-excluded"
 	TooFreshMeta      = "too-fresh"
+	duplicateMeta     = "duplicate"
 )
 
 func newSyncMetrics(r prometheus.Registerer) *syncMetrics {
@@ -84,6 +87,7 @@ func newSyncMetrics(r prometheus.Registerer) *syncMetrics {
 		[]string{failedMeta},
 		[]string{labelExcludedMeta},
 		[]string{timeExcludedMeta},
+		[]string{duplicateMeta},
 	)
 	if r != nil {
 		r.MustRegister(
@@ -409,4 +413,96 @@ func (f *LabelShardedMetaFilter) Filter(metas map[ulid.ULID]*metadata.Meta, sync
 		synced.WithLabelValues(labelExcludedMeta).Inc()
 		delete(metas, id)
 	}
+}
+
+// DeduplicateFilter is a MetaFetcher filter that filters out older blocks that have exactly the same data.
+type DeduplicateFilter struct {
+	DuplicateIDs []ulid.ULID
+}
+
+// NewDeduplicateFilter creates DeduplicateFilter.
+func NewDeduplicateFilter() *DeduplicateFilter {
+	return &DeduplicateFilter{}
+}
+
+// Filter filters out duplicate blocks that can be formed
+// from two or more overlapping blocks that fully submatches the source blocks of the older blocks.
+func (f *DeduplicateFilter) Filter(metas map[ulid.ULID]*metadata.Meta, synced GaugeLabeled, _ bool) {
+	root := NewNode(&metadata.Meta{
+		BlockMeta: tsdb.BlockMeta{
+			ULID: ulid.MustNew(uint64(0), nil),
+		},
+	})
+
+	metaSlice := []*metadata.Meta{}
+	for _, meta := range metas {
+		metaSlice = append(metaSlice, meta)
+	}
+	sort.Slice(metaSlice, func(i, j int) bool {
+		ilen := len(metaSlice[i].Compaction.Sources)
+		jlen := len(metaSlice[j].Compaction.Sources)
+
+		if ilen == jlen {
+			return metaSlice[i].ULID.Compare(metaSlice[j].ULID) < 0
+		}
+
+		return ilen-jlen > 0
+	})
+
+	for _, meta := range metaSlice {
+		addNodeBySources(root, NewNode(meta))
+	}
+
+	duplicateULIDs := getNonRootIDs(root)
+	for _, id := range duplicateULIDs {
+		if metas[id] != nil {
+			f.DuplicateIDs = append(f.DuplicateIDs, id)
+		}
+		synced.WithLabelValues(duplicateMeta).Inc()
+		delete(metas, id)
+	}
+}
+
+func addNodeBySources(root *Node, add *Node) bool {
+	var rootNode *Node
+	for _, node := range root.Children {
+		parentSources := node.Compaction.Sources
+		childSources := add.Compaction.Sources
+
+		// Block exists with same sources, add as child.
+		if contains(parentSources, childSources) && contains(childSources, parentSources) {
+			node.Children = append(node.Children, add)
+			return true
+		}
+
+		// Block's sources are present in other block's sources, add as child.
+		if contains(parentSources, childSources) {
+			rootNode = node
+			break
+		}
+	}
+
+	// Block cannot be attached to any child nodes, add it as child of root.
+	if rootNode == nil {
+		root.Children = append(root.Children, add)
+		return true
+	}
+
+	return addNodeBySources(rootNode, add)
+}
+
+func contains(s1 []ulid.ULID, s2 []ulid.ULID) bool {
+	for _, a := range s2 {
+		found := false
+		for _, e := range s1 {
+			if a.Compare(e) == 0 {
+				found = true
+				break
+			}
+		}
+		if !found {
+			return false
+		}
+	}
+	return true
 }
--- a/pkg/block/fetcher_test.go
+++ b/pkg/block/fetcher_test.go
@@ -7,10 +7,12 @@ import (
 	"bytes"
 	"context"
 	"encoding/json"
+	"fmt"
 	"io/ioutil"
 	"os"
 	"path"
 	"path/filepath"
+	"runtime"
 	"sort"
 	"testing"
 	"time"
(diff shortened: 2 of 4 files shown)


pkg/block/fetcher.go
```diff
@@ -410,3 +411,52 @@ func (f *LabelShardedMetaFilter) Filter(metas map[ulid.ULID]*metadata.Meta, sync
...
+	return &CompactedBlockFilter{}
+}
+
+// Filter filters out blocks that filters out blocks that have newer block
+// from two or more overlapping blocks that fully submatches the source blocks of the older blocks.
+func (f *CompactedBlockFilter) Filter(metas map[ulid.ULID]*metadata.Meta, synced GaugeLabeled, _ bool) {
+	for id, meta := range metas {
+		sources := make([]ulid.ULID, len(metas))
+		for _, sourceID := range meta.Compaction.Sources {
+			if sourceID == id {
```
bwplotka: Let's use `sourceID.Compare(id) == 0` to be safe maybe (:

pkg/block/fetcher.go
```diff
@@ -410,3 +411,52 @@ func (f *LabelShardedMetaFilter) Filter(metas map[ulid.ULID]*metadata.Meta, sync
...
+			}
+			sources = append(sources, sourceMeta.Compaction.Sources...)
+		}
+		sort.Slice(sources, func(i, j int) bool {
+			return sources[i].Compare(sources[j]) < 0
+		})
+		if areSlicesEqual(sources, meta.Compaction.Sources) {
+			for _, sourceID := range meta.Compaction.Sources {
+				if sourceID != id {
+					delete(metas, sourceID)
```
bwplotka: Next to deleting, we probably want store the deleted metas and check if we are ready to delete those blocks from bucket (:

pkg/block/fetcher.go
```diff
@@ -410,3 +411,52 @@ func (f *LabelShardedMetaFilter) Filter(metas map[ulid.ULID]*metadata.Meta, sync
 		delete(metas, id)
 	}
 }
+
+// CompactedBlockFilter is a MetaFetcher filter that filters out blocks that have newer block
```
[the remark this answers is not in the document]
bwplotka: Not addressed.

pkg/block/fetcher_test.go
```diff
@@ -393,3 +401,271 @@ func TestTimePartitionMetaFilter_Filter(t *testing.T) {
...
+				ULID(3): {
+					BlockMeta: tsdb.BlockMeta{
+						ULID: ULID(3),
+						Compaction: tsdb.BlockMetaCompaction{
+							Sources: []ulid.ULID{ULID(3)},
+						},
+					},
+				},
+			},
+			expected: func(input map[ulid.ULID]*metadata.Meta) map[ulid.ULID]*metadata.Meta {
```
bwplotka: Why this complex signature for expected field? Why not just expect certain slice of ulid.ULID? (:

pkg/block/fetcher_test.go
```diff
@@ -393,3 +401,271 @@ func TestTimePartitionMetaFilter_Filter(t *testing.T) {
...
 }
+
+func TestCompactedBlockFilter_Filter(t *testing.T) {
+	for _, tcase := range []struct {
+		name     string
+		input    map[ulid.ULID]*metadata.Meta
+		expected func(input map[ulid.ULID]*metadata.Meta) map[ulid.ULID]*metadata.Meta
+	}{
+		{
+			name: "3 non compacted metas in bucket",
```
bwplotka: It does not matter if they are really compacted .. It still does not say to me if it's overlapping or not

pkg/block/fetcher_test.go
```diff
@@ -393,3 +401,271 @@ func TestTimePartitionMetaFilter_Filter(t *testing.T) {
...
+			expected: func(input map[ulid.ULID]*metadata.Meta) map[ulid.ULID]*metadata.Meta {
+				return map[ulid.ULID]*metadata.Meta{
+					ULID(1): input[ULID(1)],
+					ULID(2): input[ULID(2)],
+					ULID(3): input[ULID(3)],
+				}
+			},
+		},
+		{
+			name: "compacted meta in bucket",
```
bwplotka: meta is compacted? Let's be precise (:

pkg/block/fetcher_test.go
```diff
@@ -393,3 +401,271 @@ func TestTimePartitionMetaFilter_Filter(t *testing.T) {
...
+			expected: func(input map[ulid.ULID]*metadata.Meta) map[ulid.ULID]*metadata.Meta {
+				return map[ulid.ULID]*metadata.Meta{
+					ULID(4): input[ULID(4)],
+				}
+			},
+		},
+		{
+			name: "two blocks with same sources",
+			input: map[ulid.ULID]*metadata.Meta{
+				ULID(3): {
```
bwplotka: Configuring test cases should be clear and simple. I think we don't need to have the whole `metadata.Meta` We can just configure as input: `map[ulid.ULID]tsdb.BlockMetaCompaction` and that's it. We don't really configure anything else. Then we can create a proper map to pass to filter later on. 

However, piece with test cases should be concise and without many repetitions.

pkg/block/fetcher_test.go
```diff
@@ -393,3 +401,271 @@ func TestTimePartitionMetaFilter_Filter(t *testing.T) {
...
+				},
+			},
+			expected: func(input map[ulid.ULID]*metadata.Meta) map[ulid.ULID]*metadata.Meta {
+				return map[ulid.ULID]*metadata.Meta{
+					ULID(5): input[ULID(5)],
+				}
+			},
+		},
+		{
+			name: "3 non compacted metas and compacted meta of level 2 in bucket",
```
bwplotka: again, meta is never technically compacted -> we need to consistent and clear in naming here. Also compaction does not matter here: Just what constains same sources.

For example this case could be named: `Block 4 contains 1, 2, 3`

Also I believe :
```
ULID: ULID(4),
						Compaction: tsdb.BlockMetaCompaction{
							Sources: []ulid.ULID{ULID(1), ULID(2), ULID(3), ULID(4)},
						},
```
.. is wrong. This will never happen as 4 is compacted block, so it will be never part of sources I believe (That's why I linked the Prometheus code which does it, to double check) (:

pkg/block/fetcher_test.go
```diff
@@ -393,3 +401,271 @@ func TestTimePartitionMetaFilter_Filter(t *testing.T) {
...
+				},
+			},
+			expected: func(input map[ulid.ULID]*metadata.Meta) map[ulid.ULID]*metadata.Meta {
+				return map[ulid.ULID]*metadata.Meta{
+					ULID(13): input[ULID(13)],
+				}
+			},
+		},
+		{
+			name: "compacted blocks with overlapping sources",
```
bwplotka: How is this different to other cases?

pkg/block/fetcher_test.go
```diff
@@ -393,3 +401,271 @@ func TestTimePartitionMetaFilter_Filter(t *testing.T) {
...
+				},
+			},
+			expected: func(input map[ulid.ULID]*metadata.Meta) map[ulid.ULID]*metadata.Meta {
+				return map[ulid.ULID]*metadata.Meta{
+					ULID(6): input[ULID(6)],
+				}
+			},
+		},
+	} {
+		f := NewCompactedBlockFilter()
```
bwplotka: Can we add case with 2 blocks having exactly the same sources?

pkg/block/forest.go
```diff
@@ -0,0 +1,121 @@
...
+}
+
+func (f *Forest) addVertex(id ulid.ULID, startVertex *Vertex, metas map[ulid.ULID]*metadata.Meta) bool {
+	var rootVertex *Vertex
+	newVertex := NewVertex(id, metas[id])
+	for _, vertex := range startVertex.Children {
+		rootSources := metas[vertex.ID].Compaction.Sources
+		sources := metas[id].Compaction.Sources
+
+		if f.containsSources(rootSources, sources) && f.containsSources(sources, rootSources) { // a block exists with same sources.
```
bwplotka: This line is quite long so maybe we can put comment above? And make it a full sentence? (Capital first letter)

pkg/block/forest.go
```diff
@@ -0,0 +1,121 @@
...
+	}
+
+	return newitems
+}
+
+func contains(s1 []ulid.ULID, s2 []ulid.ULID) bool {
+	for _, a := range s2 {
+		found := false
+		for _, e := range s1 {
+			if a.Compare(e) == 0 {
```
bwplotka: Nice. I think we can optimize a bit but this willl do for now (: 

We can create some go benchmark for this in next PR :muscle:

cmd/thanos/compact.go
```diff
@@ -247,9 +247,11 @@ func runCompact(
 		}
 	}()
 
+	compactedBlockFilter := block.NewCompactedBlockFilter()
 	metaFetcher, err := block.NewMetaFetcher(logger, 32, bkt, "", extprom.WrapRegistererWithPrefix("thanos_", reg),
 		block.NewLabelShardedMetaFilter(relabelConfig).Filter,
 		(&consistencyDelayMetaFilter{logger: logger, consistencyDelay: consistencyDelay}).Filter,
+		compactedBlockFilter.Filter,
```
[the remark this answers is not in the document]
khyatisoneji: Syncer is not used by store component, so we need to add the filter there explicitly then

cmd/thanos/compact.go
```diff
@@ -247,9 +247,11 @@ func runCompact(
 		}
 	}()
 
+	compactedBlockFilter := block.NewCompactedBlockFilter()
```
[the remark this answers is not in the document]
khyatisoneji: renamed this to duplicatefilter.

pkg/block/fetcher.go
```diff
@@ -410,3 +411,49 @@ func (f *LabelShardedMetaFilter) Filter(metas map[ulid.ULID]*metadata.Meta, sync
...
+// from two or more overlapping blocks that fully submatches the source blocks of the older blocks.
+type CompactedBlockFilter struct {
+	IdsToDelete []ulid.ULID
```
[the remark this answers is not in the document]
khyatisoneji: renamed this as DuplicateIDs

pkg/block/fetcher.go
```diff
@@ -410,3 +411,49 @@ func (f *LabelShardedMetaFilter) Filter(metas map[ulid.ULID]*metadata.Meta, sync
...
+		}
+
+		return ilen-jlen > 0
+	})
+
+	for _, meta := range metaSlice {
+		forest.AddMeta(meta.ULID, metas)
+	}
+
+	ulidsToDelete := forest.GetNonRootIDs()
```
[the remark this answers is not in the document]
khyatisoneji: It is returning slice, can you please give more context on your concern

pkg/block/fetcher_test.go
khyatisoneji: Yes, its added with test case name: `two compacted blocks with same sources`

khyatisoneji:
@bwplotka I have addressed some concerns that you have made:
1. Removing forest and having only vertex
2. Name test cases with compacted blocks and not compacted metas.
3. Removed the function for expected metas, currently only comparing the slices.
4. Renamed ulidsToDelete with duplicateIds
5. Rename CompactedBlockFilter with DeduplicateFilter

I have one suggestion that we should address deletion in a separate PR. Let's narrow down scope of this PR to only filter out the duplicate blocks.
Deletion of the blocks, as you mentioned should be done with checks mentioned in the proposal, so adding them in a separate PR would be better.

bwplotka:
> I have one suggestion that we should address deletion in a separate PR. Let's narrow down scope of this PR to only filter out the duplicate blocks.
Deletion of the blocks, as you mentioned should be done with checks mentioned in the proposal, so adding them in a separate PR would be better.

Very good idea. But in this case we should not probably hook this filter yet in Compactor I think. We can hook in Store Gateway though.

pkg/block/fetcher.go
```diff
@@ -410,3 +411,48 @@ func (f *LabelShardedMetaFilter) Filter(metas map[ulid.ULID]*metadata.Meta, sync
...
+func NewDeduplicateFilter() *DeduplicateFilter {
+	return &DeduplicateFilter{
+		DuplicateIDs: []ulid.ULID{},
+	}
+}
+
+// Filter filters out duplicate blocks that can be formed
+// from two or more overlapping blocks that fully submatches the source blocks of the older blocks.
+func (f *DeduplicateFilter) Filter(metas map[ulid.ULID]*metadata.Meta, synced GaugeLabeled, _ bool) {
+	forest := NewVertex(ulid.MustNew(uint64(0), nil), nil)
```
bwplotka: still forest name, maybe just `root`?

pkg/block/forest.go
```diff
@@ -0,0 +1,109 @@
...
+
+package block
+
+import (
+	"github.com/oklog/ulid"
+	"github.com/thanos-io/thanos/pkg/block/metadata"
+)
+
+// Vertex type implements a vertex of a Forest.
+type Vertex struct {
```
bwplotka: filename is still old (forest)

pkg/block/forest.go
```diff
@@ -0,0 +1,109 @@
...
+func (f *Vertex) iterate(vertex *Vertex) []ulid.ULID {
+	var ulids []ulid.ULID
+	ulids = append(ulids, vertex.ID)
+	for _, childVertex := range vertex.Children {
+		ulids = append(ulids, f.iterate(childVertex)...)
+	}
+	return ulids
+}
+
+func (f *Vertex) add(id ulid.ULID, startVertex *Vertex, metas map[ulid.ULID]*metadata.Meta) bool {
```
bwplotka: `add` is still cryptic. At least can we have a comment what this method does? (: Because it's not just vertex - it's a vertex that 

Also:
```suggestion
func (v *Vertex) add(id ulid.ULID, startVertex *Vertex, metas map[ulid.ULID]*metadata.Meta) bool {
```

pkg/block/forest.go
```diff
@@ -0,0 +1,109 @@
...
+
+	if rootVertex == nil {
+		startVertex.Children = append(startVertex.Children, newVertex)
+		return true
+	}
+
+	return f.add(id, rootVertex, metas)
+}
+
+func (f *Vertex) containsSources(rootSources, sources []ulid.ULID) bool {
```
bwplotka: Does it have to be method? Also, I don't get why we need this at all this method does nothing ): It might too shallow

pkg/block/fetcher.go
```diff
@@ -410,3 +411,48 @@ func (f *LabelShardedMetaFilter) Filter(metas map[ulid.ULID]*metadata.Meta, sync
 		delete(metas, id)
 	}
 }
+
+// DeduplicateFilter is a MetaFetcher filter that filters out duplicate blocks.
```
bwplotka:
```suggestion
// DeduplicateFilter is a MetaFetcher filter that filters out older blocks that have exactly the same data.
```

pkg/block/fetcher.go
```diff
@@ -410,3 +411,48 @@ func (f *LabelShardedMetaFilter) Filter(metas map[ulid.ULID]*metadata.Meta, sync
...
+
+		if ilen == jlen {
+			return metaSlice[i].ULID.Compare(metaSlice[j].ULID) < 0
+		}
+
+		return ilen-jlen > 0
+	})
+
+	for _, meta := range metaSlice {
+		root.AddMeta(meta.ULID, metas)
```
bwplotka: Still, I would vote to move it as a separate function (:

pkg/block/fetcher_test.go
bwplotka: not addressed

pkg/block/fetcher_test.go
bwplotka: ditto

pkg/block/node.go
```diff
@@ -0,0 +1,104 @@
...
+// Licensed under the Apache License 2.0.
+
+package block
+
+import (
+	"github.com/oklog/ulid"
+	"github.com/thanos-io/thanos/pkg/block/metadata"
+)
+
+// Node type implements a node of a Forest.
```
bwplotka: forest still ):

khyatisoneji: Suggestions? Node of a tree?

khyatisoneji: Aren't both test cases relevant? The first one is for case where two blocks have same sources and second one is where the blocks have overlapping sources.

pkg/block/node.go
```diff
@@ -0,0 +1,87 @@
...
+func (n *Node) iterate(node *Node) []ulid.ULID {
+	var ulids []ulid.ULID
+	ulids = append(ulids, node.ID)
+	for _, childNode := range node.Children {
+		ulids = append(ulids, n.iterate(childNode)...)
+	}
+	return ulids
+}
+
+func (n *Node) add(id ulid.ULID, startNode *Node, metas map[ulid.ULID]*metadata.Meta, areEqual, isChildNode compareNodes) bool {
```
bwplotka: Ok, I said like 4 times to move this from node and it's still there and you did not explain why you are against this (: 

Maybe I can send PR to your PR to show you how it can be done and show adventages? (:

bwplotka: The main problem is that this is just a `node` 

It's not  `hierarchicalBlocksNode` or anything like this. Method is also just `add` which suggests nothing is done expect just adding a node to the tree ... somewhere. This is might be not clear for readers of this code so we need to improve this.

Second argument why I believe it would nice to move `add` to separate function is extensibility. What if we would like to use our tree with meta.jsons data struct for other algorithms? We can't because `add` has certain logic baked in.

pkg/block/fetcher_test.go
```diff
@@ -393,3 +393,159 @@ func TestTimePartitionMetaFilter_Filter(t *testing.T) {
 	testutil.Equals(t, expected, input)
 
 }
+
+func TestDeduplicateFilter_Filter(t *testing.T) {
+	for _, tcase := range []struct {
+		name     string
+		input    map[ulid.ULID]tsdb.BlockMetaCompaction
```
bwplotka: Maybe just `map[ulid.ULID][]ulid.ULID`?

pkg/block/fetcher.go
```diff
@@ -410,3 +412,95 @@ func (f *LabelShardedMetaFilter) Filter(metas map[ulid.ULID]*metadata.Meta, sync
...
+	for _, meta := range metaSlice {
+		addNodeBySources(root, NewNode(meta))
+	}
+
+	duplicateULIDs := getNonRootIDs(root)
+	for _, id := range duplicateULIDs {
+		if metas[id] != nil {
+			f.DuplicateIDs = append(f.DuplicateIDs, id)
+		}
+		// TODO(khyati): Increment synced metrics with proper state.
```
bwplotka: todo

pkg/block/fetcher_test.go
```diff
@@ -393,3 +393,117 @@ func TestTimePartitionMetaFilter_Filter(t *testing.T) {
...
+			expected: []ulid.ULID{
+				ULID(1),
+				ULID(2),
+				ULID(3),
+			},
+		},
+		{
+			name: "compacted block with sources in bucket",
+			input: map[ulid.ULID][]ulid.ULID{
+				ULID(4): []ulid.ULID{ULID(1), ULID(2), ULID(3)},
```
bwplotka: Let's put other blocks here that we don't expect to be compacted

pkg/block/fetcher_test.go
```diff
@@ -393,3 +393,117 @@ func TestTimePartitionMetaFilter_Filter(t *testing.T) {
...
+			},
+			expected: []ulid.ULID{
+				ULID(4),
+			},
+		},
+		{
+			name: "two compacted blocks with same sources",
+			input: map[ulid.ULID][]ulid.ULID{
+				ULID(3): []ulid.ULID{ULID(1), ULID(2)},
```
bwplotka: ditto, let's mix it

pkg/block/fetcher_test.go
```diff
@@ -393,3 +393,117 @@ func TestTimePartitionMetaFilter_Filter(t *testing.T) {
...
+				ULID(4): []ulid.ULID{ULID(1), ULID(2)},
+			},
+			expected: []ulid.ULID{
+				ULID(3),
+			},
+		},
+		{
+			name: "two compacted blocks with overlapping sources",
+			input: map[ulid.ULID][]ulid.ULID{
+				ULID(4): []ulid.ULID{ULID(1), ULID(2)},
```
bwplotka: ditto

pkg/block/fetcher_test.go
```diff
@@ -393,3 +393,117 @@ func TestTimePartitionMetaFilter_Filter(t *testing.T) {
...
+				ULID(5): []ulid.ULID{ULID(1), ULID(2), ULID(3)},
```
bwplotka: Can we reorder ULIDs in Source to check if that will work?

pkg/block/fetcher_test.go
```diff
@@ -393,3 +393,117 @@ func TestTimePartitionMetaFilter_Filter(t *testing.T) {
...
+				ULID(4): []ulid.ULID{ULID(1), ULID(2), ULID(3)},
+			},
+			expected: []ulid.ULID{
+				ULID(4),
+			},
+		},
+		{
+			name: "3 compacted blocks of level 2 and one compacted block of level 3 in bucket",
+			input: map[ulid.ULID][]ulid.ULID{
+				ULID(10): []ulid.ULID{ULID(1), ULID(2), ULID(3)},
```
bwplotka: ditto, let's mix with some block that should stay

pkg/block/fetcher_test.go
```diff
@@ -393,3 +393,117 @@ func TestTimePartitionMetaFilter_Filter(t *testing.T) {
...
+			expected: []ulid.ULID{
+				ULID(13),
+			},
+		},
+		{
+			name: "compacted blocks with overlapping sources",
+			input: map[ulid.ULID][]ulid.ULID{
+				ULID(1): []ulid.ULID{ULID(1)},
+				ULID(5): []ulid.ULID{ULID(1), ULID(2)},
+				ULID(6): []ulid.ULID{ULID(1), ULID(2), ULID(3), ULID(4)},
```
bwplotka: Let's reorder those a bit to make it more tricky. Yes we put it in map anyway, but sometimes map tends to be.. accidently ordered (:

pkg/block/fetcher_test.go
```diff
@@ -393,3 +393,117 @@ func TestTimePartitionMetaFilter_Filter(t *testing.T) {
...
+				ULID(8): []ulid.ULID{ULID(1), ULID(2), ULID(3), ULID(4)},
+			},
+			expected: []ulid.ULID{
+				ULID(6),
+			},
+		},
+	} {
+		f := NewDeduplicateFilter()
+		if ok := t.Run(tcase.name, func(t *testing.T) {
+			synced := prometheus.NewGaugeVec(prometheus.GaugeOpts{}, []string{"state"})
```
bwplotka: We can assert on this metric! 

See our talk: https://are-you-testing-your-observability.now.sh/

Click `right` until 26 slide and check it out (:

pkg/block/fetcher_test.go
```diff
@@ -393,3 +393,117 @@ func TestTimePartitionMetaFilter_Filter(t *testing.T) {
...
+			},
+		},
+		{
+			name: "compacted blocks with overlapping sources",
+			input: map[ulid.ULID][]ulid.ULID{
+				ULID(1): []ulid.ULID{ULID(1)},
+				ULID(5): []ulid.ULID{ULID(1), ULID(2)},
+				ULID(6): []ulid.ULID{ULID(1), ULID(2), ULID(3), ULID(4)},
+				ULID(7): []ulid.ULID{ULID(1), ULID(2), ULID(3)},
+				ULID(8): []ulid.ULID{ULID(1), ULID(2), ULID(3), ULID(4)},
```
bwplotka: Let's also swap `ULID(7)` to `ULID(8)` (: Make it more tricky!

pkg/block/fetcher_test.go
```diff
@@ -393,3 +393,117 @@ func TestTimePartitionMetaFilter_Filter(t *testing.T) {
...
+			},
+			expected: []ulid.ULID{
```
bwplotka: Can we add one more case?

The case would be something like 

`10 containing  1,2,6,7`
`11 containing  1,2,6,8`
(: 
And maybe:
`10 containing  1,2,6,7`
`11 containing  1,2,3`

khyatisoneji:
Added the requested changes:
1. implemented todo to add sync metric for filtered duplicates
2. Reordered ulids in test
3. Mixed compacted and non compacted ulids in test
4. Added requested test additions
5. added test to test metric for filtered duplicates.

pkg/block/fetcher_test.go
```diff
@@ -393,3 +395,178 @@ func TestTimePartitionMetaFilter_Filter(t *testing.T) {
...
+			compareSliceWithMapKeys(t, metas, tcase.expected)
+			testutil.Equals(t, float64(inputLen-len(tcase.expected)), promtest.ToFloat64(synced.WithLabelValues(duplicateMeta)))
+		}); !ok {
+			return
+		}
+	}
+}
+
+func compareSliceWithMapKeys(tb testing.TB, m map[ulid.ULID]*metadata.Meta, s []ulid.ULID) {
+	_, file, line, _ := runtime.Caller(1)
```
bwplotka: Wow this is epic!  Probably bit over-engineered, but... thanks!  (:

Issue #1651, opened 2019-10-15
Response caching for Thanos


bwplotka:
This is an umbrella issue for things we might need to do/contribute in order to have smooth Thanos response caching.

As you probably know Cortex query-frontend [now supports Prometheus API](https://www.youtube.com/watch?v=eyBbImSDOrI), so - it can be put on top of Thanos. Still, there are some cases we have to/can improve for Thanos usage:

Must:
* [ ] Understanding underlying block structure and downsampling handling. Query frontend splits by day. This does not make sense for our downsampling which helps to run queries for large time ranges. Also split by day when querying 2w block might be not that efficient as most of the index lookup has to be 14x duplicated. However, splitting allows for easier load balancing and distributing the queries across system. This means that we might want maybe a bit smarter query split. Some idea would be:

```
// thanosQuerySplitIntervalFn returns split interval based on query step. This is based on the fact that
// Thanos in auto-downsampling mode will choose also use step when choosing what resolution to use.
//
// Splitting rationales: 
// Cortex splits by day and it's quite a good ratio of chunks per series to fetch. For 1 day we will have approx 24 chunks/series (30s scrape). Also 
// To max keep 24 chunks per series for 5m resolution it means 10 days split, for 1h resolution 120 days.
var thanosIntervalFn = func(r Request)time.Duration {
	step := time.Duration(r.Step()) * time.Millisecond
	if step > 30 * time.Minute {
		return 120 * day
	}
	if step > 5 * time.Hour {
		return 10 * day
	}
	return day
}
```
This might be some improvement, but still, there are some edge cases:
* Fallback to raw chunks if downsampled blocks are still missing
* Downsampled blocks appear only after 40h (5m) and 10d (1h), so this logic might need to consider time as well.
* Split by day efficiency might differ across time, depending on how many blocks are touched.

I think there is still no clear idea of what's the best practice there. Ideally, we would split based on the underlying block structure, but there is no easy way to have that info during query planning level.

Maybe treating query splitting and splitting response a separate steps make more sense? Right now response caching relies on splitting as the cache key depends on queried start time. This is to avoid lookup logic in the caching layer. 

* [ ] Allow setting custom query params to Cortex frontend
   * [ ] Auto downsampling has to be used for the above interval based on step logic.
   * [x] Partial response must NOT be cached. This can be achieved by simply setting `--query.no`partial-response` however we can also add some logic for caching to be avoided if the response has some special header. 
[...]

PR #2072, opened 2020-01-28
query: add 'sticky' store nodes

GiedriusS:
Add the ability to have 'sticky' store nodes - we will always consider them available and retain their clients. This makes it possible to have consistent partial responses when a node goes down when we know that it must be up at all times. The relevant part from the documentation which explains more:

```
Thanos Query periodically checks the health of the StoreAPI nodes that it knows about via the `Info()` gRPC call. However, if one is failing then it is not considered a part of the active set of StoreAPI nodes. To make them always part of the active set, you need to make them "sticky" - their last available information will be retained even in the face of a failure of a health-check.

This is useful in the cases where you have some kind of caching layer in front of Thanos Query i.e. Cortex's `query-frontend` and you know that certain nodes must always be alive. It allows you to get a partial response when one of the sticky nodes goes down.

To make a node sticky you need to add a suffix `+sticky` to the end of the address.

Sticky nodes have a yellow `UP` status in the `Stores` page if we have failed to check their data but we still consider them available.
```

Ad-hoc tests:

Tested locally. Turned off one store node which made it get removed from the `healthyStores`. Sending queries for a short time still gave an error but it went away after the refresh delay.

Added a `+sticky` suffix then. Shut off the same node and tried sending queries again. Got errors that Thanos Query couldn't connect to that node. Disabling partial response made it into an error.

Looking for feedback on:
* Does it even make sense?! :smile: 
* The code, ofc
* Is it okay to have this as a suffix? Or should we make static targets 'sticky' by default? But that would change Thanos Query's behavior for our users and maybe someone wants to stay with the current behavior. IMHO we should leave this possibility to add a suffix.

WIP:
* Still want to add a test for the parsing part
* Looking for feedback

Relevant issue: https://github.com/thanos-io/thanos/issues/1651.

### DIFF
--- a/pkg/query/storeset.go
+++ b/pkg/query/storeset.go
@@ -48,6 +48,7 @@ type StoreStatus struct {
 	StoreType component.StoreAPI
 	MinTime   int64
 	MaxTime   int64
+	Sticky    bool
 }
 
 type grpcStoreSpec struct {
@@ -67,7 +68,7 @@ func (s *grpcStoreSpec) Addr() string {
 }
 
 func (s *grpcStoreSpec) Sticky() bool {
-	return s.Sticky()
+	return s.sticky
 }
 
 // Metadata method for gRPC store API tries to reach host Info method until context timeout. If we are unable to get metadata after
@@ -359,7 +360,7 @@ func (s *StoreSet) Update(ctx context.Context) {
 
 		st.Close()
 		delete(stores, addr)
-		s.updateStoreStatus(st, errors.New(unhealthyStoreMessage))
+		s.updateStoreStatus(st, st.sticky, errors.New(unhealthyStoreMessage))
 		level.Info(s.logger).Log("msg", unhealthyStoreMessage, "address", addr, "extLset", st.LabelSetsString())
 	}
 
@@ -383,7 +384,7 @@ func (s *StoreSet) Update(ctx context.Context) {
 		stats[st.StoreType()][st.LabelSetsString()]++
 
 		stores[addr] = st
-		s.updateStoreStatus(st, nil)
+		s.updateStoreStatus(st, st.sticky, nil)
 		level.Info(s.logger).Log("msg", "adding new storeAPI to query storeset", "address", addr, "extLset", extLset)
 	}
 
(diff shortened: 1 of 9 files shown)


pkg/query/storeset.go
```diff
@@ -36,6 +36,8 @@ type StoreSpec interface {
 	// NOTE: It is implementation responsibility to retry until context timeout, but a caller responsibility to manage
 	// given store connection.
 	Metadata(ctx context.Context, client storepb.StoreClient) (labelSets []storepb.LabelSet, mint int64, maxt int64, storeType component.StoreAPI, err error)
+	// Sticky tells us whether the StoreSpec should always be considered healthy.
+	Sticky() bool
```
bwplotka:
```suggestion
	IsSticky() bool
```
to be consistent.

pkg/query/storeset.go
```diff
@@ -428,25 +440,29 @@ func (s *StoreSet) getHealthyStores(ctx context.Context, stores map[string]*stor
 					// Close only if new. Unhealthy `s.stores` will be closed later on.
 					st.Close()
 				}
-				s.updateStoreStatus(st, err)
-				level.Warn(s.logger).Log("msg", "update of store node failed", "err", errors.Wrap(err, "getting metadata"), "address", addr)
+				level.Warn(s.logger).Log("msg", "update of store node failed", "err", errors.Wrap(err, "getting metadata"), "address", addr, "sticky", spec.Sticky())
+				if seenAlready && spec.Sticky() {
+					mtx.Lock()
+					healthyStores[addr] = st
```
bwplotka: This feels very wrong (: 

Putting unhealthy thing in healthy map

bwplotka:
I would love to hear others' opinions because this is something we should do from start probably even. @brancz @squat @povilasv @FUSAKLA @domgreen @IKSIN @d-ulyanov

GiedriusS:
> I would love to hear others' opinions because this is something we should do from start probably even. @brancz @squat @povilasv @FUSAKLA @domgreen @IKSIN @d-ulyanov

Since it seems like there are different options and opinions here so lets put this PR on hold for a bit and let me make a proposal PR instead where we would make the decision on how to proceed.

GiedriusS:
The proposal which suggests going further with this: https://github.com/thanos-io/thanos/pull/2086.

d-ulyanov:
Hi, guys and thank you for mentioning us!
Totally agree with @bwplotka - I would make all stores "sticky".

As a user I expect following behaviuor:
As we're adding stores just as a simple list to Thanos Query config - we expect all of them work.
If one of them is not responding - any response should be marked as partial: we don't know for sure is retrieved data consistent or not (hope in the future we'll invent some approach how to know are  requested metrics located on unhealthy Store or not).

If partial response enabled - as usual I'm expecting to get warning, otherwise - request must fail.
Implementation could be simple - let's make StoreSet keep 2 lists - all stores and healthy stores. If their lenght is not equal - mark response as partial in all cases (in the future we can improve this logic).

Also, I would not add any logic related to "sticky" to DNSProvider - I'm not sure that its should think about anything except for DNS resolution, maybe it would be better to add some wrapper around DNS provider.

GiedriusS:
I will close this and open another one with the agreed implementation on the proposal (:

PR #2076, opened 2020-01-29
Makefile: optimize jsonnet-format target

daixiang0:

Fix following error:
```
# make examples
find . -name 'vendor' -prune -o -name '*.libsonnet' -print -o -name '*.jsonnet' -print | \
	xargs -n 1 -- jsonnetfmt -n 2 --max-blank-lines 2 --string-style s --comment-style s -i
xargs: jsonnetfmt: No such file or directory
```

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -359,6 +359,9 @@ JSONNET_FMT := jsonnetfmt -n 2 --max-blank-lines 2 --string-style s --comment-st
 
 .PHONY: jsonnet-format
 jsonnet-format:
+	@which jsonnetfmt 2>/dev/null || ( \
+		echo "Cannot find jsonnetfmt command, please install from https://github.com/google/jsonnet/releases. If your C++ does not support GLIBCXX_3.4.20, please use xxx-in-container target like jsonnet-format-in-container." && exit 1
+	)
 	find . -name 'vendor' -prune -o -name '*.libsonnet' -print -o -name '*.jsonnet' -print | \
 		xargs -n 1 -- $(JSONNET_FMT) -i
 


Makefile
```diff
@@ -358,7 +358,7 @@ jsonnet-vendor: $(JSONNET_BUNDLER) jsonnetfile.json jsonnetfile.lock.json
 JSONNET_FMT := jsonnetfmt -n 2 --max-blank-lines 2 --string-style s --comment-style s
 
 .PHONY: jsonnet-format
-jsonnet-format:
+jsonnet-format: $(JSONNET)
```
squat: Does that target dependency actually install `jsonnetfmt`? It looks like it only installs `jsonnet`

daixiang0: please wait ...

daixiang0: @squat just find that jsonnetfmt seems cannot be installed by go install:
```
# go install 'github.com/google/go-jsonnet/cmd/jsonnetfmt'
go: finding github.com/google/go-jsonnet v0.14.0
go: downloading github.com/google/go-jsonnet v0.14.0
go: extracting github.com/google/go-jsonnet v0.14.0
can't load package: package github.com/google/go-jsonnet/cmd/jsonnetfmt: module github.com/google/go-jsonnet@latest found (v0.14.0), but does not contain package github.com/google/go-jsonnet/cmd/jsonnetfmt

```
Need more time to fix it.
How to mark this pr as draft?

Makefile
```diff
@@ -425,6 +426,9 @@ $(PROTOC):
 $(JSONNET):
 	$(call fetch_go_bin_version,github.com/google/go-jsonnet/cmd/jsonnet,$(JSONNET_VERSION))
 
+$(JSONNET):
```
kakkoyun: @daixiang0  This supposed to be `JSONNETFMT` or better `JSONNET_FMT`, but then you have to rename things here and there.
`jsonnetfmt` is not a go binary by the way so I suggest something like this.

```suggestion
$(JSONNETFMT):
	cd tmp && curl -Lso - https://github.com/google/jsonnet/archive/v${JSONNET_VERSION}.tar.gz | \
    tar xfz - -C . && \
    cd jsonnet-${JSONNET_VERSION} && \
    make && mv jsonnetfmt $(BIN_DIR) && \
    rm -rf /tmp/jsonnet-${JSONNET_VERSION}
```

Probably you'll need to define `BIN_DIR`.

Makefile
```diff
@@ -358,7 +359,7 @@ jsonnet-vendor: $(JSONNET_BUNDLER) jsonnetfile.json jsonnetfile.lock.json
 JSONNET_FMT := jsonnetfmt -n 2 --max-blank-lines 2 --string-style s --comment-style s
 
 .PHONY: jsonnet-format
-jsonnet-format:
+jsonnet-format: $(JSONNETFMT)
 	find . -name 'vendor' -prune -o -name '*.libsonnet' -print -o -name '*.jsonnet' -print | \
```
kakkoyun:
```suggestion
	@PATH=$$PATH:$$(pwd)/$(BIN_DIR) find . -name 'vendor' -prune -o -name '*.libsonnet' -print -o -name '*.jsonnet' -print | \
```

daixiang0:
Since even we download binary, it may not work as expected, i would check the cmd and comment here.

Issue #984, opened 2019-03-27
Fix `goimports`.


bwplotka:
AC:
* `make format` or `gomiports` should fix blank lines.

Eg: 
```
 	"net/http"
 	"path"
 	"regexp"
 	"sort"
 
+	thanosrule "github.com/improbable-eng/thanos/pkg/rule"
+
 	"github.com/improbable-eng/thanos/pkg/store/storepb"
 
 	"github.com/go-kit/kit/log"
 	"github.com/prometheus/client_golang/prometheus"
 	"github.com/prometheus/common/route"
```
`goimports` does NOT fixes those, see why: https://github.com/golang/go/issues/26246

This is quite annoying. I think this was fixed, but then reverted. Anyway we should ensure this is fixed.

jojohappy:
I saw it had been fixed with pinned `goimports`. Any update? @bwplotka

bwplotka:
Still valid.

bwplotka:
Same problem.

bwplotka:
Still broken

bwplotka:
Still goimports is leaving blank lines.

bwplotka:
eh eh (:

PR #2077, opened 2020-01-29
testutil: Moving all e2e utils to separate package, added TB union.

bwplotka:

### DIFF
--- a/cmd/thanos/main_test.go
+++ b/cmd/thanos/main_test.go
@@ -9,9 +9,6 @@ import (
 	"os"
 	"path"
 	"path/filepath"
-
-	"github.com/thanos-io/thanos/pkg/block/metadata"
-
 	"testing"
 	"time"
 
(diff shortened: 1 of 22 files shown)


cmd/thanos/main_test.go
```diff
@@ -11,6 +11,7 @@ import (
 	"path/filepath"
 
 	"github.com/thanos-io/thanos/pkg/block/metadata"
+	"github.com/thanos-io/thanos/pkg/testutil/e2eutil"
```
kakkoyun: nit: Shall we group the imports? Does it even matter for the tests?

pkg/testutil/testorbench.go
```diff
@@ -0,0 +1,71 @@
...
+
+func (tb *UnionTB) ResetTimer() {
+	if b, ok := tb.TB.(*testing.B); ok {
+		b.ResetTimer()
+		return
+	}
+}
+
+func (tb *UnionTB) IsBenchmark() bool {
+	if _, ok := tb.TB.(*testing.B); ok {
```
kakkoyun: I'm totally sure if this makes easier to read or not, but the following could also be used
```suggestion
	_, ok := tb.TB.(*testing.B)
	return ok
```

cmd/thanos/main_test.go
bwplotka: we should but we don't have tools for that so it's tedious and I am lazy (: 

@jojohappy if you were asking if still valid: https://github.com/thanos-io/thanos/issues/984

pkg/testutil/testorbench.go
```diff
@@ -0,0 +1,71 @@
...
+	if b, ok := tb.TB.(*testing.B); ok {
+		return b.N
+	}
+	return 1
+}
+
+func (tb *UnionTB) ResetTimer() {
+	if b, ok := tb.TB.(*testing.B); ok {
+		b.ResetTimer()
+		return
```
GiedriusS: very minor nit: not needed return :P

pkg/testutil/testorbench.go
bwplotka: Makes sense :+1:

Issue #1830, opened 2019-12-03
compact: Hanging upload


bwplotka:
As referred in https://gitlab.com/gitlab-com/gl-infra/infrastructure/issues/

TODO: Get more info, type of storage, last log line from @SuperQ

cstyan:
As discussed on Slack, looks like we've just run into the same issue.

We're using GCS, last log line is:

```
level=info ts=2019-11-30T04:53:44.700533615Z caller=compact.go:440 msg="compact blocks" count=6 mint=1574899200000 maxt=1575072000000 ulid=01DTX9XYZZDSWKT2BRYM59CGT9 sources="[01DTS3ZE5QSVSC0D06PPH81ZKA 01DTSRJJXYNV5H9DZ806QT0JND 01DTTM1F3G68Y1107WNEA30XN7 01DTVFH73PH5AABGDM293P9V8K 01DTWAZY17ZXK645R1EP3YFST9 01DTX6DMWGMH65R10QX1QK7TV7]" duration=10m17.751658532s
```

So it looks like the upload is stuck, no log lines about those source blocks being deleted.

bwplotka:
Do you think we can grab the pprof goroutines if that will happen again @cstyan @SuperQ

SuperQ:
Yes, I'll see what I can do.

bwplotka:
Still valid, pinning.

bwplotka:
Still we need some timeout.

bwplotka:
Still timeout pliz (:

bwplotka:
Still todo

bwplotka:
The task is quite easy: To ensure timeouts for uploads. (:

SuperQ:
Still TODO.

PR #2079, opened 2020-01-30
store: S3 object store via AWS SDK

cyrus-mc:
This PR allows supporting of the following setup:

* AWS Account A
    * A S3 bucket with a policy allowing Account B to PutObject
* AWS Account B/C/D/...
   * An IAM Role with the policy PutObject on bucket in Account A

Various Prometheus + Thanos clusters are running in Account B/C/D/.. and writing their storage to S3 bucket in Account A. In Account A I have Thanos Querier, Compact, etc running.

In the above scenario the default ACL assigned to an object only allows the owner access (Account B). Account A has no permissions to that object thus the Thanos components can't read the contents uploaded by the sub-accounts.

The current S3 implementation coded against Minio doesn't support setting ACL (as Minio server doesn't support ACL). The other option would be to use an assume role but the Minio client doesn't support this either. 

This implementation of the S3 storage allows setting an ACL on PutObject.

## Changes

* Added object storage for S3 utilizing the AWS SDK

## Verification

Currently only manual.

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -16,6 +16,7 @@ require (
 	github.com/alecthomas/units v0.0.0-20190924025748-f65c72e2690d
 	github.com/aliyun/aliyun-oss-go-sdk v2.0.4+incompatible
 	github.com/armon/go-metrics v0.3.0
+	github.com/aws/aws-sdk-go v1.25.48
 	github.com/baiyubin/aliyun-sts-go-sdk v0.0.0-20180326062324-cfa1a18b161f // indirect
 	github.com/bradfitz/gomemcache v0.0.0-20190913173617-a41fca850d0b
 	github.com/cespare/xxhash v1.1.0
--- a/pkg/objstore/aws/aws.go
+++ b/pkg/objstore/aws/aws.go
@@ -21,7 +21,6 @@ import (
 	"github.com/aws/aws-sdk-go/service/s3"
 	"github.com/go-kit/kit/log"
 	"github.com/go-kit/kit/log/level"
-	"github.com/minio/minio-go/v6/pkg/encrypt"
 	"github.com/pkg/errors"
 	"github.com/prometheus/common/model"
 	"github.com/thanos-io/thanos/pkg/objstore"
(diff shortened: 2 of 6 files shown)


cyrus-mc:
This PR is not complete. I am new to golang so figured I would open the PR to elicit feedback.

bwplotka:
I totally see the reasoning, but this needs some discussion before allowing separate AWS SDK implementation. Mainly because it's not the best to have two things solving the same problem (minio client and AWS SDK client for S3 backend). 

However, given some timeout issues we seen with minio client:https://github.com/thanos-io/thanos/issues/1830.. maybe it's worth considering.

cyrus-mc:
@bwplotka 

I know the minio client will add assume role capabilities in the coming weeks. But (reading other GitHub issues) they have shot down adding ACL since the server will never support ACL. 

I will continue to develop this implementation as the discussion continues.

cyrus-mc:
@bwplotka 

I have completed development of this implementation. I believe it to be feature compatible with the minio implementation.

bwplotka:
Cool. There was one good question done by @kakkoyun 

Why not using just https://github.com/google/go-cloud? It uses SDK and it will allow us to remove also GCS, Azure implemetations (: Would that work?

cyrus-mc:
> Cool. There was one good question done by @kakkoyun
> 
> Why not using just https://github.com/google/go-cloud? It uses SDK and it will allow us to remove also GCS, Azure implemetations (: Would that work?

Not familiar with go-cloud.

cyrus-mc:
@bwplotka Two question.

First, the failed build. This seems to be something with the build pipeline in general and not related to any code changes I made (I see other PR suffer this same error).

Second, I have been running this in production for a week or so now. Ran into a few issues which I have fixed. Was the idea of using minio due to the fact you could then use Minio as an S3 backend? Or other S3 compatible backends? 

I believe my implementation will still work in those situations.

povilasv:
In general I agree with @bwplotka we should look for alternatives for minio client and trying them before actually adding this, as this will cause confusion (which S3 to use?), this will add more dependencies we need to take care of etc. 

Let's try go-cloud first if there are no better alternatives which seem to do the job, we can consider this as a viable option, but then maybe deprecate the old one?

bwplotka:
Sad we could merge this ): But we really need to have a proper solution with go-cloud instead I believe.

Thanks for your work @cyrus-mc anyway!

jtweaver:
Was a solution for this ever implemented? We have essentially the same scenario as OP and are just realizing the account with the bucket (bucket owner, not object owner) cannot access the objects pushed by the other accounts.

PR #2085, opened 2020-01-31
Display partial response strategy in ruler UI

yeya24:

## Changes

Support displaying partial response strategy in ruler UI

## Verification

image

image

### DIFF
--- a/pkg/rule/rule.go
+++ b/pkg/rule/rule.go
@@ -6,17 +6,21 @@ package thanosrule
 import (
 	"crypto/sha256"
 	"fmt"
+	html_template "html/template"
 	"io/ioutil"
 	"os"
 	"path/filepath"
+	"sort"
 	"strings"
 	"sync"
 	"time"
 
 	"github.com/pkg/errors"
+	"github.com/prometheus/common/model"
 	"github.com/prometheus/prometheus/pkg/rulefmt"
 	"github.com/prometheus/prometheus/rules"
 	tsdberrors "github.com/prometheus/prometheus/tsdb/errors"
+	"github.com/prometheus/prometheus/util/strutil"
 	"github.com/thanos-io/thanos/pkg/store/storepb"
 	"gopkg.in/yaml.v2"
 )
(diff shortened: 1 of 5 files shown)


pkg/rule/rule.go
```diff
@@ -33,11 +37,109 @@ func (g Group) OriginalFile() string {
...
+
+func (g Group) GetAlertingRules() []*AlertingRule {
+	var alerts []*AlertingRule
+	for _, rule := range g.Rules() {
+		if alertingRule, ok := rule.(*rules.AlertingRule); ok {
+			alerts = append(alerts, &AlertingRule{alertingRule, g.PartialResponseStrategy})
+		}
+	}
+	sort.Slice(alerts, func(i, j int) bool {
+		return alerts[i].State() > alerts[j].State() ||
```
GiedriusS: Nice touch :+1:

pkg/rule/rule.go
```diff
@@ -33,11 +37,109 @@ func (g Group) OriginalFile() string {
...
+	})
+	return alerts
+}
+
 type AlertingRule struct {
 	*rules.AlertingRule
 	PartialResponseStrategy storepb.PartialResponseStrategy
 }
 
+func (r *AlertingRule) HTMLSnippet(pathPrefix string) html_template.HTML {
```
GiedriusS: Shouldn't we document this and other new public functions?

yeya24: I am not sure, these functions are only used in UI, I will add some comments.

pkg/rule/rule.go
```diff
@@ -33,11 +37,114 @@ func (g Group) OriginalFile() string {
...
 
+// GetRules returns a list of rules, which is used in UI rules page.
+func (g *Group) GetRules() []rules.Rule {
+	var rs []rules.Rule
+	for _, rule := range g.Rules() {
+		if r, ok := rule.(*rules.AlertingRule); ok {
+			rs = append(rs, &AlertingRule{r, g.PartialResponseStrategy})
+		} else if r, ok := rule.(*rules.RecordingRule); ok {
+			rs = append(rs, &RecordingRule{r, g.PartialResponseStrategy})
+		}
```
povilasv: consider doing a type switch https://golang.org/doc/effective_go.html#type_switch

pkg/rule/rule.go
```diff
@@ -33,11 +37,114 @@ func (g Group) OriginalFile() string {
...
+	}
+
+	byt, err := yaml.Marshal(ar)
+	if err != nil {
+		return html_template.HTML(fmt.Sprintf("error marshaling alerting rule: %q", html_template.HTMLEscapeString(err.Error())))
+	}
+	return html_template.HTML(byt)
+}
+
+type RecordingRule struct {
```
povilasv: Let's add a comment or make this unexported

pkg/rule/rule.go
yeya24: Thanks!

pkg/rule/rule.go
yeya24: Done. Please take a look agian

pkg/rule/rule.go
```diff
@@ -33,11 +37,115 @@ func (g Group) OriginalFile() string {
...
 
+// GetRules returns a list of rules, which is used in UI rules page.
+func (g *Group) GetRules() []rules.Rule {
+	var rs []rules.Rule
+	for _, rule := range g.Rules() {
+		switch r := rule.(type) {
+		case *rules.AlertingRule:
+			rs = append(rs, &AlertingRule{r, g.PartialResponseStrategy})
+		case *rules.RecordingRule:
+			rs = append(rs, &recordingRule{r, g.PartialResponseStrategy})
```
povilasv: having AlertingRule exported and recordingRule unexported looks strange, I think it would be good to have both exported or unexported for consistency :)

Exporting it allows other users to use it as package, unexporting it hides away from that. Not sure which one is the best in this case tbh

yeya24: Since AlertingRule must be public, I will make it recordingRule public as well and add a comment

yeya24:
CI failed but unrelated

GiedriusS:
I have double-checked one thing and my last concern is that currently we define `partial_response_strategy` on the rule group level, not on a specific alerting rule. Would it be a lot of effort to move this to the actual table? I'm afraid people will get confused about why the alerting/recording rules do not work if they will copy them from this UI as-is.
Also, maybe it would be worth to hide this value if it is `ABORT` as that's the default?

yeya24:
> I have double-checked one thing and my last concern is that currently we define `partial_response_strategy` on the rule group level, not on a specific alerting rule. Would it be a lot of effort to move this to the actual table? I'm afraid people will get confused about why the alerting/recording rules do not work if they will copy them from this UI as-is. 

I totally agree with you! Partial response strategy is a rule group level config instead of per rule level. It definitely should be on the rule group side, but to be honest I don't come up with the appropriate place to put it in the UI currently. 

Put the strategy field on a specific rule field is confusing to users. So maybe it is better to close/hold this PR until I find a better way.

> Alternatively, we could support overriding those values at the alerting/recording rule level.

It is possible but I am not sure if it is something we want to do. @bwplotka

yeya24:
> I have double-checked one thing and my last concern is that currently we define `partial_response_strategy` on the rule group level, not on a specific alerting rule. Would it be a lot of effort to move this to the actual table? I'm afraid people will get confused about why the alerting/recording rules do not work if they will copy them from this UI as-is. Alternatively, we could support overriding those values at the alerting/recording rule level.
> Also, maybe it would be worth to hide this value if it is `ABORT` as that's the default?

Thanos ruler `/api/v1/rules` API adds the `partial_response_strategy` in both rule group, per-rule level and even alert level. Do you think it's OK? @GiedriusS  I think this is quite confusing as well
``` JSON
groups: [
          {
            "name": "AlertmanagerMembersInconsistent",
            "query": "alertmanager_cluster_members{job=\"alertmanager-main\",namespace=\"monitoring\"} != on(service) group_left() count by(service) (alertmanager_cluster_members{job=\"alertmanager-main\",namespace=\"monitoring\"})",
            "duration": 300,
            "labels": {
              "severity": "critical"
            },
            "annotations": {
              "message": "Alertmanager has not found all other members of the cluster."
            },
            "alerts": [
              {
                "labels": {
                  "alertname": "AlertmanagerMembersInconsistent",
                  "endpoint": "web",
                  "instance": "192.168.11.5:9093",
                  "job": "alertmanager-main",
                  "namespace": "monitoring",
                  "pod": "alertmanager-main-1",
                  "prometheus": "monitoring/k8s",
                  "service": "alertmanager-main",
                  "severity": "critical"
                },
                "annotations": {
                  "message": "Alertmanager has not found all other members of the cluster."
                },
                "state": "pending",
                "activeAt": "2020-02-04T14:56:55.05551088Z",
                "value": "2e+00",
                "partial_response_strategy": "ABORT"
              }
            ],
            "health": "ok",
            "type": "alerting",
            "partial_response_strategy": "ABORT"
          }
        ],
        "interval": 30,
        "partial_response_strategy": "ABORT"
      }
]
```

yeya24:
> That's nice, can we still have it? ;p

I am thinking about moving this to the React UI, not in the original one. Is it okay?

bwplotka:
Definitely!

yeya24:
Valid. But let me go back to this after our frontend moving to React

yeya24:
/remind me about this pr in 1 month

yeya24:
valid.

yeya24:
Valid. Since the new rule UI is done, I will finish this soon.

PR #2086, opened 2020-01-31
Add proposal for improving Thanos Query healthiness handling

GiedriusS:
Adds a document which proposes improving the healthiness handling of the
store nodes in Thanos Query.

### DIFF
--- a/docs/proposals/202001_thanos_query_health_handling.md
+++ b/docs/proposals/202001_thanos_query_health_handling.md
@@ -60,6 +60,16 @@ After careful consideration and with the rationale in this proposal, we have dec
 
 * Fixing the cache handling in the cases where a **new** StoreAPI gets added to the active store set.
 
+This deserves a separate discussion and/or proposal. The issue when adding a completely new store node via service discovery is that a new node may suddenly provide new information in the past. In this paragraph when we are saying "new" it means new in terms of the data that it provides. Generally over time only a limited number of Prometheus instances will be providing data that can only change in the future (relatively to the current time).
+
+When this happens, we will need to most likely somehow signal the caching layer that it needs to drop (some of the) results cache that it has depending on:
+
+* Time ranges of a new node.
+* Its external labels.
+* The data that the node itself has by using some kind of hashing mechanism.
+
+The way this will need to be done should be as generic as possible so the design and solution of this is still an open question that this proposal does not solve.
+
 ## Verification
 
 * Unit tests which would fire up a dummy StoreAPI and check different scenarios.
--- a/docs/proposals/202001_thanos_query_health_handling.md
+++ b/docs/proposals/202001_thanos_query_health_handling.md
@@ -1,0 +1,86 @@
+---
+title: Thanos Query store nodes healthiness handling
+type: proposal
+menu: proposals
+status: proposed
+owner: GiedriusS
+---
+
+### Related Tickets
+
+* https://github.com/thanos-io/thanos/issues/1651 (Response caching for Thanos)
+* https://github.com/thanos-io/thanos/pull/2072 (query: add 'sticky' nodes)
+* https://github.com/cortexproject/cortex/pull/1974 (frontend: implement cache control)
+* https://github.com/thanos-io/thanos/pull/1984 (api/v1: add cache-control header)
+
+## Summary
+
+This proposal document describes how currently the healthiness of store nodes is handled and why it should be changed to work better in terms of end response caching.
+
+It explores a few options that are available - weighs their pros and cons, and finally proposes one final variant which seems to be the best out of the possible ones.
+
+## Motivation
+
+Currently Thanos Query updates the list of healthy store nodes every 5 seconds. It does that by sending the `Info()` call via gRPC. The last successful check is noted in the `LastCheck` field. At this point if it fails then we note the error and remove it from the set of active store nodes. If that succeeds then it becomes a part of the active store set.
+
+After `--store.unhealthy-timeout` passes since `LastCheck` then it also gets removed from the UI. At this point we would forget about it completely.
+
+Every time a query is executed, we consult the active store set and send the query to all of the nodes which match according to the external labels and the min/max times that they advertise. However, if a node goes down according to our previous definition then in 5 seconds we will not send anything to it anymore. This means that we won't even be able to control this via the partial response options since no query is sent to those nodes in the first place.
+
+This is problematic in the cache of end-response caching. If we know that certain StoreAPI nodes should always be up then we always want to have not just an error (if partial response is disabled) but also the appropriate `Cache-Control` header in the response. But right now we would only get it for maximum 5 seconds after a store node would go down.
+
+Thus, this logic needs to be changed somehow. There are a few possible options:
+
+* `--store.unhealthy-timeout` could be made to apply to this case as well - we could still consider it a part of the active store set while it is still visible in the UI
+* Another option could be introduced such as `--store.hold-timeout` which would be `--store.unhealthy-timeout`'s brother and we would hold the StoreAPI nodes for `max(hold_timeout, unhealthy_timeout)`.
+* Another option such as `--store.strict-mode` could be introduced which means that we would always retain the last information of the StoreAPI nodes after the first successful check
+* The StoreAPI node specification format that is used in `--store` could be extended to include another flag which would let specify the previous option per-specific node
+
+Lets look through their pros and cons:
+
+* In general it would be nice to avoid adding new options so the first option seems the most attractive in this regard but it is also the most invasive one;
+* Second option increases the configuration complexity the most since you would have to think about the other option `--store.unhealthy-timeout` as well while setting it;
+* The last option is the most complex from code's perspective but it is also least invasive; however it has some downsides like the syntax for specifying store nodes becomes really ugly and hard to understand because we would have not only the DNS query type in there but also a special marker to enable that mode
+
+If we were to graph these choices in terms of theirs incisiveness and complexity it would look something like this:
+
+```text
+Most incisive / Least Complex ------------ Least incisive / Most Complex
+#1           #2                                      #4
+           #3
+```
+
+At first there was an initial attempt to add the last option but it was met with some discussion if this is really the way to go forward. After careful consideration and with the rationale in this proposal we could go with the most invasive choice. But it also probably makes the most sense since it would logically follow that as long as a node is visible in the UI then we send a query to it.
+
+## Goals
+
+* Update the health check logic to properly handle the case when a node disappears in terms of a caching layer
+
+## No Goals
+
+* Fixing the cache handling in the cases where a **new** StoreAPI gets added to the active store set
+
+## Verification
+
+* Unit tests which would fire up a dummy StoreAPI and check different scenarios
+* Ad-hoc testing
+
+## Proposal
+
+* Keep store nodes around in the active store set in Thanos Query until `--store.unhealthy-timeout` passes
+
+## Risk
+
+* Users might start having partial responses with this implemented if they have certain nodes configured that are always down. However, one might argue that if the nodes go down then something like DNS service discovery needs to be used which would dynamically add and remove those nodes
+
+## Work Plan
+
+* Implement the change in logic not to remove nodes from the active store set immediately after one `Info()` gRPC call failure but until `--store.unhealthy-timeout` passes
+* Implement tests with dummy store nodes
+* Document the new behavior
+
+## Future Work
+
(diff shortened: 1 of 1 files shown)


daixiang0:
It sounds like something in S/C architecture, would it a good idea that support active push/reconnect by client(here means store) . It may result in coupling but fix control issue in reconnect case i think.

d-ulyanov:
What was initial motivation of having `--store.unhealthy-timeout`?
Maybe I don't understand something, but I don't see any reason for cleaning up unhealthy stores from the store list at all.

bwplotka:
Thanks for this design doc @GiedriusS :heart: 

My opinion stays the same as I mentioned here: https://github.com/thanos-io/thanos/pull/2072#pullrequestreview-349905672

I think we should add `query.strict-mode` which makes Querier to never clear any store that was added statically.

The problem is with the hosts that are dynamically added with DNS, or FileSD. I don't believe we can be strict about those.

GiedriusS:
So it seems like no one else has any comments and we have at least some kind of minimum consensus. Thus, I will update this PR to reflect this decision and set the status to `accepted`. Does that sound about right?

bwplotka:
Happy with that. Let's update and I will ping some people for opinion tomorrow (:

GiedriusS:
Let's talk about this in the upcoming maintainers sync :P

bwplotka:
@GiedriusS let's put it on the agenda then

docs/proposals/202001_thanos_query_health_handling.md
```diff
@@ -0,0 +1,86 @@
...
+### Related Tickets
+
+* https://github.com/thanos-io/thanos/issues/1651 (Response caching for Thanos)
+* https://github.com/thanos-io/thanos/pull/2072 (query: add 'sticky' nodes)
+* https://github.com/cortexproject/cortex/pull/1974 (frontend: implement cache control)
+* https://github.com/thanos-io/thanos/pull/1984 (api/v1: add cache-control header)
+
+## Summary
+
+This proposal document describes how currently the healthiness of store nodes is handled and why it should be changed to work better in terms of end response caching.
```
bwplotka:
```suggestion
This proposal document describes how currently the healthiness of store nodes is handled and why it should be changed (e.g. to improve response caching and avoid surprises)
```

docs/proposals/202001_thanos_query_health_handling.md
```diff
@@ -0,0 +1,86 @@
...
+
+It explores a few options that are available - weighs their pros and cons, and finally proposes one final variant which seems to be the best out of the possible ones.
```
bwplotka:
```suggestion
It explores a few options that are available - weighs their pros and cons, and finally proposes one final variant which we believe is the best out of the possible ones.
```

docs/proposals/202001_thanos_query_health_handling.md
```diff
@@ -0,0 +1,86 @@
...
+
+Every time a query is executed, we consult the active store set and send the query to all of the nodes which match according to the external labels and the min/max times that they advertise. However, if a node goes down according to our previous definition then in 5 seconds we will not send anything to it anymore. This means that we won't even be able to control this via the partial response options since no query is sent to those nodes in the first place.
+
+This is problematic in the cache of end-response caching. If we know that certain StoreAPI nodes should always be up then we always want to have not just an error (if partial response is disabled) but also the appropriate `Cache-Control` header in the response. But right now we would only get it for maximum 5 seconds after a store node would go down.
+
+Thus, this logic needs to be changed somehow. There are a few possible options:
+
+* `--store.unhealthy-timeout` could be made to apply to this case as well - we could still consider it a part of the active store set while it is still visible in the UI
```
bwplotka: Let's actually number then as we referenced them as numbers afterwards.

docs/proposals/202001_thanos_query_health_handling.md
```diff
@@ -0,0 +1,86 @@
...
+
+If we were to graph these choices in terms of theirs incisiveness and complexity it would look something like this:
+
+```text
+Most incisive / Least Complex ------------ Least incisive / Most Complex
+#1           #2                                      #4
+           #3
+```
+
+At first there was an initial attempt to add the last option but it was met with some discussion if this is really the way to go forward. After careful consideration and with the rationale in this proposal we could go with the most invasive choice. But it also probably makes the most sense since it would logically follow that as long as a node is visible in the UI then we send a query to it.
```
bwplotka: Does full story really matters? ;p 

I would just remove sentence:

> At first there was an initial attempt to add the last option but it was met with some discussion if this is really the way to go forward.

docs/proposals/202001_thanos_query_health_handling.md
```diff
@@ -0,0 +1,86 @@
...
+
+## Goals
+
+* Update the health check logic to properly handle the case when a node disappears in terms of a caching layer
+
+## No Goals
+
+* Fixing the cache handling in the cases where a **new** StoreAPI gets added to the active store set
```
bwplotka: Missing trailing period in all of those sentence above, here and below

docs/proposals/202001_thanos_query_health_handling.md
```diff
@@ -0,0 +1,86 @@
...
+
+## Verification
+
+* Unit tests which would fire up a dummy StoreAPI and check different scenarios
+* Ad-hoc testing
+
+## Proposal
+
+* Keep store nodes around in the active store set in Thanos Query until `--store.unhealthy-timeout` passes
```
bwplotka: Is this what we have chosen?

docs/proposals/202001_thanos_query_health_handling.md
```diff
@@ -0,0 +1,86 @@
...
+## No Goals
+
+* Fixing the cache handling in the cases where a **new** StoreAPI gets added to the active store set
+
+## Verification
+
+* Unit tests which would fire up a dummy StoreAPI and check different scenarios
+* Ad-hoc testing
+
+## Proposal
```
bwplotka: Can we try to think about some solution (or future solution) for discovered notes as @brancz suggested?

bwplotka: Or at least let's mention we mean static nodes here and move SD nodes to `Future Work`

docs/proposals/202001_thanos_query_health_handling.md
```diff
@@ -0,0 +1,86 @@
...
+
+* Keep store nodes around in the active store set in Thanos Query until `--store.unhealthy-timeout` passes
+
+## Risk
+
+* Users might start having partial responses with this implemented if they have certain nodes configured that are always down. However, one might argue that if the nodes go down then something like DNS service discovery needs to be used which would dynamically add and remove those nodes
```
bwplotka: trailing period but also I don't really understand this sentence, can you rephrase?: 

>  However, one might argue that if the nodes go down then something like DNS service discovery needs to be used which would dynamically add and remove those nodes

docs/proposals/202001_thanos_query_health_handling.md
```diff
@@ -0,0 +1,86 @@
...
+
+## Work Plan
+
+* Implement the change in logic not to remove nodes from the active store set immediately after one `Info()` gRPC call failure but until `--store.unhealthy-timeout` passes
```
bwplotka: Do we really agreed on this? I thought we agreed on `strict mode`

docs/proposals/202001_thanos_query_health_handling.md
```diff
@@ -0,0 +1,86 @@
...
+* Implement tests with dummy store nodes
+* Document the new behavior
+
+## Future Work
+
+* Handle the case when a new node appears in terms of the end-result cache
+  1. Need to somehow signal the upper layer to clear the end-result cache
```
bwplotka: Yup :+1:

docs/proposals/202001_thanos_query_health_handling.md
```diff
@@ -0,0 +1,86 @@
+---
+title: Thanos Query store nodes healthiness handling
+type: proposal
+menu: proposals
+status: proposed
```
bwplotka: Let's make it `accepted` when we update the doc to agreed solution.

docs/proposals/202001_thanos_query_health_handling.md
GiedriusS: I haven't changed this yet 😄

docs/proposals/202001_thanos_query_health_handling.md
GiedriusS: No, I still haven't updated this yet. I am about to do that 😄

docs/proposals/202001_thanos_query_health_handling.md
GiedriusS: Added this :+1:

docs/proposals/202001_thanos_query_health_handling.md
GiedriusS: Yep, added this as well :+1: I have used that language previously since I have decided that myself personally, and now we have decided to use this option so this is appropriate.

docs/proposals/202001_thanos_query_health_handling.md
GiedriusS: Converted into numbers :+1: makes sense

docs/proposals/202001_thanos_query_health_handling.md
GiedriusS: Removed :P

docs/proposals/202001_thanos_query_health_handling.md
GiedriusS: I think I have added the trailing period everywhere. Checked with a simple: `/bin/grep -E -e "\.$" -v ./docs/proposals/202001_thanos_query_health_handling.md`. LMK if I have missed something.

docs/proposals/202001_thanos_query_health_handling.md
GiedriusS: That's what the blurb in the "Future Work" section at the end is about. Do you think we should mention something more? I have added there some more text that there we are talking about the cases where SD is used.

docs/proposals/202001_thanos_query_health_handling.md
GiedriusS: Rephrased it with your suggestion + some extra text. Let me know if it makes sense.

docs/proposals/202001_thanos_query_health_handling.md
```diff
@@ -0,0 +1,86 @@
...
+
+* Implement the new flag `--store.strict-mode` in Thanos Query which will make it keep around statically defined nodes.
+* Implement tests with dummy store nodes.
+* Document the new behavior.
+
+## Future Work
+
+* Handle the case when a new node appears in terms of the end-result cache i.e. when using SD:
+  1. Need to somehow signal the upper layer to clear the end-result cache. Ideally we would only clear the relevant parts.
+  1. Perhaps some kind of hashing can be used for this.
```
povilasv:
```suggestion
  2. Perhaps some kind of hashing can be used for this.
```

docs/proposals/202001_thanos_query_health_handling.md
```diff
@@ -0,0 +1,86 @@
...
+
+* Add a new flag to Thanos Query `--store.strict-mode` which will make it always retain the last successfully retrieved information via the `Info()` gRPC method of **statically** defined nodes and thus always consider them part of the active store set.
+
+## Risk
+
+* Users might have problems removing the store nodes from the active store set since they will be there forever with this option set. However, one might argue that if the nodes go down then something like DNS service discovery needs to be used which would dynamically add and remove those nodes.
+
+## Work Plan
+
+* Implement the new flag `--store.strict-mode` in Thanos Query which will make it keep around statically defined nodes.
```
povilasv: Consider adding more info around whether we will make it default behaviour or opt-in?

GiedriusS: Added `It will be disabled by default to reduce surprises when upgrading.` here. I think that should be explicit enough :)

docs/proposals/202001_thanos_query_health_handling.md
bwplotka: Where? I think we need to explain it better what are the challenges with service discovery.

GiedriusS: I feel like there was some miscommunication. It was only briefly mentioned in the end before this. I really wanted to keep it brief since all of this will probably come in a separate proposal. Now I have added 4 paragraphs about this under "No Goals". I really think that we shouldn't elaborate more on that in this document since we are not trying to solve that problem now.

PR #2090, opened 2020-02-03
Document downsample component

### LINKED ISSUE #1938: Docs for downsample cmd
I think we need docs for command `downsample` as well.

daixiang0:
Fix #1938


daixiang0:
Now it is a subcmd for thanos too which could act as a service i think. Maybe comment it as a build-in feature in Compator too?

bwplotka:
No, it cannot run like a service. 

If you want to run as a service then  you should use compactor indeed (: 

What I am saying is that I would love to move this as a subcommand of `thanos bucket` instead.

daixiang0:
Get it, how about add an option for Compator to make it work as a tool? If move it as subcommand of `thanos bucket`, it looks a bit repeated.

yeya24:
I agree with @bwplotka  that it is better to become a subcommand of `thanos bucket`.

daixiang0:
@bwplotka hi, what need i change? Move this as a subcommand of thanos bucket and move this document?

bwplotka:
> @bwplotka hi, what need i change? Move this as a subcommand of thanos bucket and move this document?

That would be amazing TBH (: let's double-check with the team.

@squat @metalmatze @brancz @FUSAKLA @GiedriusS  @povilasv  WDYT?

squat:
That sounds like a great idea to me:)

povilasv:
Definetely agree :+1: I remember `thanos downsample` was confusing when I was first learning Thanos. I had questions like: is it another thanos component? Do I need to run it? etc

So I think moving this onto bucket makes more sense

metalmatze:
Agreed. :)

daixiang0:
@bwplotka done, mind have a look?

pkg/downsample/downsample.go
```diff
@@ -0,0 +1,300 @@
...
+	}, []string{"group"})
+	m.downsampleFailures = promauto.With(reg).NewCounterVec(prometheus.CounterOpts{
+		Name: "thanos_compact_downsample_failures_total",
+		Help: "Total number of failed downsampling attempts.",
+	}, []string{"group"})
+
+	return m
+}
+
+func RunDownsample(
```
bwplotka: hm I think this should rather land in `cmd` to be consistent with other commands

daixiang0: Do you mean just keep `cmd/downsample.go` and call it in bucket? Or keep a part of  `cmd/downsample.go`?

bwplotka: I would say part of `bucket.go`?

daixiang0: I have changed it, if anything wrong, please tell me.

bwplotka:
Ah it is also kind of braking change so let's add CHANGELOG in next PR (:

daixiang0:
Track as #2278

docs/components/bucket.md
```diff
@@ -389,3 +392,71 @@ Flags:
       --single-run               Run replication only one time, then exit.
 
 ```
+
+### Downsample
+
+The downsample component of Thanos implements the downsample API on top of historical data in an object storage bucket. It continuously downsamples blocks in an object store bucket as a service.
+
+```bash
+$ thanos downsample \
```
belm0, line 401: should this be `thanos bucket downsample`, matching usage below?

GiedriusS: @belm0, feel free to submit a PR which fixes this :P

PR #2095, opened 2020-02-04
Add benchmark scripts

daixiang0:

Now we have some benchmarks but not clear to users who want to test performance.

### DIFF
--- a/scripts/benchmark/expandedpostings.sh
+++ b/scripts/benchmark/expandedpostings.sh
@@ -1,0 +1,19 @@
+#!/usr/bin/env bash
+
+DIR="$(dirname $(readlink -e -v ${BASH_SOURCE[0]}))"
+
+cd ${DIR}/../../
+
+RUN="expandedpostings"
+mkdir -p ${DIR}/bench_outs/${RUN}
+
+for TEST in BenchmarkBucketIndexReader_ExpandedPostings/binary_header BenchmarkBucketIndexReader_ExpandedPostings/index_json; do
+    echo "Running ${TEST} run ${RUN}"
+    go test -bench=${TEST} -run=^$ -benchmem -memprofile ${DIR}/bench_outs/${RUN}/memprofile${TEST//\//-}.out -timeout 2h -benchtime 60s ./pkg/store > ${DIR}/bench_outs/${RUN}/bench${TEST//\//-}.out
+done
+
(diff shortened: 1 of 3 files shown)


scripts/benchmark/expandedpostings.sh
```diff
@@ -0,0 +1,19 @@
...
+
+RUN="expandedpostings"
+mkdir -p ${DIR}/bench_outs/${RUN}
+
+for TEST in BenchmarkBucketIndexReader_ExpandedPostings/binary_header BenchmarkBucketIndexReader_ExpandedPostings/index_json; do
+    echo "Running ${TEST} run ${RUN}"
+    go test -bench=${TEST} -run=^$ -benchmem -memprofile ${DIR}/bench_outs/${RUN}/memprofile${TEST//\//-}.out -timeout 2h -benchtime 60s ./pkg/store > ${DIR}/bench_outs/${RUN}/bench${TEST//\//-}.out
+done
+
+sed 's/BenchmarkBucketIndexReader_ExpandedPostings\/binary_header/BenchmarkBucketIndexReader_ExpandedPostings/g' ${DIR}/bench_outs/${RUN}/benchBenchmarkBucketIndexReader_ExpandedPostings-binary_header.out > ${DIR}/bench_outs/${RUN}/bench_new.out
```
bwplotka: So this part is related to comparing index-header to index-cache.json. Since we move to index-header, we will remove index-cache.json support soon.

That's why I would rather suggest refactoring this into script that will run benchmark on master vs here - but that might involve bit git checkout mangling. Not sure if worth to check them in and maintain...

daixiang0: It is related to thanos version indeed, check them in may be more easily for users.

bwplotka:
Hey, I would close this in favor of https://github.com/prometheus/test-infra/tree/master/funcbench which is already supported and maintained by us. (: It's really similar but in Go. What do you think? :hugs:

daixiang0:
Cool~
Is there any docs about how to run thanos benchmark with this tool?

bwplotka:
There are none ):

bwplotka:
Oh wait I added `Makefile` at some point: https://github.com/thanos-io/thanos/pull/2233 with all of this (: But PR is still not ready, feel free to continue (:

PR #2096, opened 2020-02-04
pkg/store/proxy.go: optimize help info

### LINKED ISSUE #2091: Ability to query store-gateway without active sidecar
Hi, this is a kind feature request. 
I current scheme, thanos query does not allow you query anything from a store if a sidecar that uploaded data into this store terminated already. 
My proposal is have ability to query historical data from store does even sidecar that uploaded it don't exist already. Mostly we need it for short-lifetime geo-distributed k8s clusters that make's some testing, analytics and send result to some bucket and then cluster will shortly terminated.

daixiang0:
Fix #2091 

There is not only one case for query failure:
- no storeAPIs available
- no storeAPIs matched
- include all above cases

Also, querier use not only storer, but also sidecar, may be other, use storeAPIs here is better.

### DIFF
--- a/pkg/store/proxy.go
+++ b/pkg/store/proxy.go
@@ -294,7 +294,7 @@ func (s *ProxyStore) Series(r *storepb.SeriesRequest, srv storepb.Store_SeriesSe
 		level.Debug(s.logger).Log("msg", strings.Join(storeDebugMsgs, ";"))
 		if len(seriesSet) == 0 {
 			// This is indicates that configured StoreAPIs are not the ones end user expects.
-			err := errors.New("No store matched for this query")
+			err := errors.New("No StoreAPIs matched for this query")
 			level.Warn(s.logger).Log("err", err, "stores", strings.Join(storeDebugMsgs, ";"))
 			respSender.send(storepb.NewWarnSeriesResponse(err))
 			return nil


pkg/store/proxy.go
```diff
@@ -294,7 +294,7 @@ func (s *ProxyStore) Series(r *storepb.SeriesRequest, srv storepb.Store_SeriesSe
 		level.Debug(s.logger).Log("msg", strings.Join(storeDebugMsgs, ";"))
 		if len(seriesSet) == 0 {
 			// This is indicates that configured StoreAPIs are not the ones end user expects.
-			err := errors.New("No store matched for this query")
+			err := errors.New("No available StoreAPIs matched for this query")
```
jojohappy: Thanks for that! But I disagree this change. It's still confused when the stores are all available(IMO that means all of stores are healthy in the Stores page).

daixiang0: It includes all case in my comment in the commit i think.

bwplotka: Yes, but I agree with @jojohappy 

It feels like available can be confusing. What about: 
```suggestion
			err := errors.New("No StoreAPI matched for this query")
```

PR #2113, opened 2020-02-10
Added `thanos bucket replicate`

daixiang0:

Implement [thanos-replicate](https://github.com/observatorium/thanos-replicate.git) to support replicate from one object storage to another.

TODO:
* [x] Add document for replcate.
* [x] Add mixin about replcate.


yeya24:
Hmm, do we plan to add `thanos-replicate` here?

daixiang0:
Yep, here is a discuss in the [slack](https://cloud-native.slack.com/archives/CK5RSSC10/p1581074820366800).

yeya24:
> Yep, here is a discuss in the [slack](https://cloud-native.slack.com/archives/CK5RSSC10/p1581074820366800).

Thanks for clarifying 😄

cmd/thanos/bucket.go
```diff
@@ -375,6 +379,38 @@ func registerBucketWeb(m map[string]setupFunc, root *kingpin.CmdClause, name str
 	}
 }
 
+func registerBucketReplicate(m map[string]setupFunc, root *kingpin.CmdClause, name string, objStoreConfig *extflag.PathOrContent) {
+	cmd := root.Command("replicate", "Replicate data from one object storage to another")
```
bwplotka:
```suggestion
	cmd := root.Command("replicate", "Replicate data from one object storage to another. NOTE: Currently it works only with Thanos blocks (meta.json has to have Thanos metadata).")
```

cmd/thanos/bucket.go
```diff
@@ -375,6 +379,38 @@ func registerBucketWeb(m map[string]setupFunc, root *kingpin.CmdClause, name str
...
+	matcherStrs := cmd.Flag("matcher", "Only blocks whose labels match this matcher will be replicated.").PlaceHolder("key=\"value\"").Strings()
+	singleRun := cmd.Flag("single-run", "Run replication only one time, then exit.").Default("false").Bool()
+
+	m[name+" replicate"] = func(g *run.Group, logger log.Logger, reg *prometheus.Registry, tracer opentracing.Tracer, _ bool) error {
+		matchers, err := replicater.ParseFlagMatchers(*matcherStrs)
+		if err != nil {
+			return errors.Wrap(err, "parse block label matchers")
+		}
+
+		return replicater.RunReplicate(
```
bwplotka:
```suggestion
		return replicator.RunReplicate(
```

docs/components/bucket.md
```diff
@@ -315,3 +318,70 @@ Flags:
       --timeout=5m           Timeout to download metadata from remote storage
 
 ```
+
+### replicate
+
+`bucket replicate` is used to replicate buckets from one object storage to another.
```
bwplotka:
```suggestion
`bucket replicate` is used to replicate buckets from one object storage to another.

NOTE: Currently it works only with Thanos blocks (meta.json has to have Thanos metadata).
```

go.mod
```diff
@@ -73,6 +73,7 @@ require (
 	github.com/prometheus/common v0.7.0
 	github.com/prometheus/procfs v0.0.6 // indirect
 	github.com/prometheus/prometheus v1.8.2-0.20200110114423-1e64d757f711 // master ~ v2.15.2
+	github.com/prometheus/tsdb v0.10.0
```
bwplotka: This is wrong, something imports obsolete tsdb. Please check imports and move all to `github.com/prometheus/prometheus`

pkg/replicater/scheme.go
```diff
@@ -0,0 +1,381 @@
...
+		labelSelector:   labelSelector,
+		logger:          logger,
+		resolutionLevel: resolutionLevel,
+		compactionLevel: compactionLevel,
+	}
+}
+
+// Filter return true if block is non-compacted and matches selector.
+func (bf *BlockFilter) Filter(b *metadata.Meta) bool {
+	blockLabels := labels.FromMap(b.Thanos.Labels)
```
bwplotka: If labels are empty let's error out. (or return false plus error log line)

pkg/replicater/scheme.go
```diff
@@ -0,0 +1,381 @@
...
+		metrics:     metrics,
+	}
+}
+
+func (rs *replicationScheme) execute(ctx context.Context) error {
+	availableBlocks := []*metadata.Meta{}
+
+	level.Debug(rs.logger).Log("msg", "scanning blocks available blocks for replication")
+
+	if err := rs.fromBkt.Iter(ctx, "", func(name string) error {
```
bwplotka: Let's add `TODO: Use block.MetaFetcher with filters instead.` (:  Let's not do to all in the same PR!

cmd/thanos/bucket.go
daixiang0: I use "replicater" as package name, i google it and "replcater" and "replicator" are both right grammar, how to choose?

pkg/replicater/scheme.go
daixiang0: Also, there are some TODO in original repo, need to move all here?

cmd/thanos/bucket.go
```diff
@@ -375,6 +379,38 @@ func registerBucketWeb(m map[string]setupFunc, root *kingpin.CmdClause, name str
 	}
 }
 
+func registerBucketReplicate(m map[string]setupFunc, root *kingpin.CmdClause, name string, objStoreConfig *extflag.PathOrContent) {
+	cmd := root.Command("replicate", "Replicate data from one object storage to another. NOTE: Currently it works only with Thanos blocks (meta.json has to have Thanos metadata).")
+	httpMetricsBindAddr, _ := regHTTPFlags(cmd)
+	toObjStoreConfig := regCommonObjStoreFlags(cmd, "-to", false, "The object storage which replicate data to.")
+	resolution := cmd.Flag("resolution", "Only blocks with this resolution will be replicated.").Default(strconv.FormatInt(downsample.ResLevel0, 10)).Int64()
```
yeya24: A small question, is it better to use `resolution level` instead of the milliseconds value here?

daixiang0: It is int value:
```
      --resolution=0             Only blocks with this resolution will be replicated.

```

CHANGELOG.md
```diff
@@ -28,6 +28,7 @@ We use *breaking* word for marking changes that are not backward compatible (rel
 - [#2030](https://github.com/thanos-io/thanos/pull/2030) Query: Add `thanos_proxy_store_empty_stream_responses_total` metric for number of empty responses from stores.
 - [#2049](https://github.com/thanos-io/thanos/pull/2049) Tracing: Support sampling on Elastic APM with new sample_rate setting.
 - [#2008](https://github.com/thanos-io/thanos/pull/2008) Querier, Receiver, Sidecar, Store: Add gRPC [health check](https://github.com/grpc/grpc/blob/master/doc/health-checking.md) endpoints.
+- [#2113](https://github.com/thanos-io/thanos/pull/2113) Bucket: Implement replicate.
```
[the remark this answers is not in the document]
daixiang0: Does it mean i need to write what this pr does and not in the changelog doc?

pkg/replicater/scheme.go
daixiang0: The mixin part would finish in other PR.

cmd/thanos/bucket.go
```diff
@@ -375,6 +379,38 @@ func registerBucketWeb(m map[string]setupFunc, root *kingpin.CmdClause, name str
...
+	compaction := cmd.Flag("compaction", "Only blocks with this compaction level will be replicated.").Default("1").Int()
+	matcherStrs := cmd.Flag("matcher", "Only blocks whose labels match this matcher will be replicated.").PlaceHolder("key=\"value\"").Strings()
```
GiedriusS: I think it would be worth saying here that we are only taking about external labels and that only exact matches are specified. Something like this:
```suggestion
	matcherStrs := cmd.Flag("matcher", "Only blocks whose external labels exactly match this matcher will be replicated.").PlaceHolder("key=\"value\"").Strings()
```

pkg/replicater/scheme.go
```diff
@@ -0,0 +1,387 @@
...
+// loadMeta loads the meta.json from the origin bucket and returns the meta
+// struct as well as if failed, whether the failure was due to the meta.json
+// not being present or partial. The distinction is important, as if missing or
+// partial, this is just a temporary failure, as the block is still being
+// uploaded to the origin bucket.
+func loadMeta(ctx context.Context, bucket objstore.BucketReader, id ulid.ULID) (*metadata.Meta, bool, error) {
+	src := path.Join(id.String(), thanosblock.MetaFilename)
+
+	r, err := bucket.Get(ctx, src)
+	if bucket.IsObjNotFoundErr(err) {
```
GiedriusS: Don't we forget to do `r.Close()` here if the object has not been found?

cmd/thanos/bucket.go
GiedriusS: It seems like Oxford's dictionary has no entry for `replicater`: https://www.oxfordlearnersdictionaries.com/spellcheck/english/?q=replicater
Only `replicator`: https://www.oxfordlearnersdictionaries.com/spellcheck/english/?q=replicator

Also, I'm not a native English speaker but `replicator` seems the correct version here to me.

daixiang0: This web helps me! Thanks a lot!

pkg/replicater/scheme.go
daixiang0: Good catch!

daixiang0: Good catch, thanks!

daixiang0: @GiedriusS refer to [test](https://circleci.com/gh/thanos-io/thanos/7171?utm_campaign=vcs-integration-link&utm_medium=referral&utm_source=github-build-link), if close here, it would raise a error, the older one is fine i think, so revert change.

GiedriusS: You are correct, it seems like there is some kind of implicit invariant here in our object storage implementations that no `io.ReadCloser` is returned if the object doesn't exist.

cmd/thanos/bucket.go
```diff
@@ -375,6 +379,38 @@ func registerBucketWeb(m map[string]setupFunc, root *kingpin.CmdClause, name str
 	}
 }
 
+func registerBucketReplicate(m map[string]setupFunc, root *kingpin.CmdClause, name string, objStoreConfig *extflag.PathOrContent) {
+	cmd := root.Command("replicate", "Replicate data from one object storage to another. NOTE: Currently it works only with Thanos blocks (meta.json has to have Thanos metadata).")
```
GiedriusS: Minor nit: we can use `block.MetaFilename` here.

cmd/thanos/bucket.go
GiedriusS: I was thinking maybe it would be useful to do an `Enum` param here and list all of the three possible different options? Then it would be clearer to our users. Otherwise, they'd have to go into the code to know the possible values. Not sure if we have those millisecond values documented anywhere.

pkg/replicator/scheme.go
```diff
@@ -0,0 +1,387 @@
...
+	originMetaFile, err := rs.fromBkt.Get(ctx, metaFile)
+	if err != nil {
+		return fmt.Errorf("get meta file from origin bucket: %w", err)
+	}
+
+	defer runutil.CloseWithLogOnErr(rs.logger, originMetaFile, "close original meta file")
+
+	targetMetaFile, err := rs.toBkt.Get(ctx, metaFile)
+	if targetMetaFile != nil {
+		defer runutil.CloseWithLogOnErr(rs.logger, targetMetaFile, "close target meta file")
```
GiedriusS: I guess now that we know that we have that invariant, we should continue the typical pattern here instead of this:
```
	targetMetaFile, err := rs.toBkt.Get(ctx, metaFile)
	if err != nil {
		return fmt.Errorf("get meta file from target bucket: %w", err)
	}
        defer runutil.CloseWithLogOnErr(rs.logger, targetMetaFile, "close target meta file")
        ...
```
Or is there some reason you wrote this in such a way?

pkg/replicator/scheme.go
```diff
@@ -0,0 +1,387 @@
...
+
+	return nil
+}
+
+// loadMeta loads the meta.json from the origin bucket and returns the meta
+// struct as well as if failed, whether the failure was due to the meta.json
+// not being present or partial. The distinction is important, as if missing or
+// partial, this is just a temporary failure, as the block is still being
+// uploaded to the origin bucket.
+func loadMeta(ctx context.Context, bucket objstore.BucketReader, id ulid.ULID) (*metadata.Meta, bool, error) {
```
GiedriusS: I feel like you could use our `MetaFetcher` here and for the 2nd return argument you could check if the returned error is `block.ErrorSyncMetaNotFound`.

cmd/thanos/bucket.go
daixiang0: It is int value indeed ...

cmd/thanos/bucket.go
daixiang0: Sorry, do not get your meanning :(

cmd/thanos/bucket.go
daixiang0: Comment in docs/components/compact.md i think, maybe not very clear.
I will use `Enum` here.

cmd/thanos/bucket.go
GiedriusS: That's the constant that has the value "meta.json" so we could do fmt.Sprintf here :P

daixiang0: Get it, thanks for your patience.

pkg/replicator/scheme.go
daixiang0: This `if` ensure targetMetaFile is not nil, the following `if` would handle error.

pkg/replicator/scheme.go
```diff
@@ -0,0 +1,389 @@
...
+			return nil, true, fmt.Errorf("fetch meta: %w", err)
+		}
+	}
+
+	m, ok := metas[id]
+	if !ok {
+		return nil, true, fmt.Errorf("fetch meta: %w", err)
+	}
+
+	if metas[id].Version != metadata.MetaVersion1 {
```
GiedriusS: Isn't this already checked in `loadMeta` inside of `MetaFetcher.loadMeta`? :P

daixiang0: Oh, i see, thanks!

brancz:
>  Add mixin about replicate.

I think it would be nice to have this. I'm happy with just transferring what we have in the other repo.

daixiang0:
Yep, it is in the TODO list, and i will do it.

pkg/replicator/replicater.go
```diff
@@ -0,0 +1,200 @@
+// Copyright (c) The Thanos Authors.
+// Licensed under the Apache License 2.0.
+
+package replicator
```
brancz: I think go best practice would rather be "replicate".

daixiang0: ping @bwplotka @GiedriusS

cmd/thanos/bucket.go
```diff
@@ -377,6 +381,46 @@ func registerBucketWeb(m map[string]setupFunc, root *kingpin.CmdClause, name str
...
+		strconv.FormatInt(downsample.ResLevel0, 10),
+		strconv.FormatInt(downsample.ResLevel1, 10),
+		strconv.FormatInt(downsample.ResLevel2, 10)}
+}
+
+func registerBucketReplicate(m map[string]setupFunc, root *kingpin.CmdClause, name string, objStoreConfig *extflag.PathOrContent) {
+	cmd := root.Command("replicate", fmt.Sprintf("Replicate data from one object storage to another. NOTE: Currently it works only with Thanos blocks (%v has to have Thanos metadata).", block.MetaFilename))
+	httpMetricsBindAddr, _ := regHTTPFlags(cmd)
+	toObjStoreConfig := regCommonObjStoreFlags(cmd, "-to", false, "The object storage which replicate data to.")
+	resolution := cmd.Flag("resolution", "Only blocks with this resolution will be replicated.").Default(strconv.FormatInt(downsample.ResLevel0, 10)).HintAction(listResLevel).Int64()
```
bwplotka: hm, I think we might need actually `Int64s` for it as we might want to replicate more than one resolution. Let's add `TODO` for now.

Also `resolution-level` ?

```suggestion
        // TODO(bwplotka): Allow to replicate many compaction levels.
	resolution := cmd.Flag("resolution-level", "Only blocks with this resolution will be replicated.").Default(strconv.FormatInt(downsample.ResLevel0, 10)).HintAction(listResLevel).Int64()
```

cmd/thanos/bucket.go
```diff
@@ -377,6 +381,46 @@ func registerBucketWeb(m map[string]setupFunc, root *kingpin.CmdClause, name str
...
+	compaction := cmd.Flag("compaction", "Only blocks with this compaction level will be replicated.").Default("1").Int()
```
bwplotka: `compaction-level`? 

And same TODO as above needed

bwplotka:
```suggestion
        // TODO(bwplotka): Allow to replicate many compaction levels.
	compaction := cmd.Flag("compaction-level", "Only blocks with this compaction level will be replicated. 1 means non-compacted block, straight from source e.g. Prometheus.").Default("1").Int()
```

pkg/replicator/scheme.go
```diff
@@ -0,0 +1,384 @@
...
+
+	return nil
+}
+
+// loadMeta loads the meta.json from the origin bucket and returns the meta
+// struct as well as if failed, whether the failure was due to the meta.json
+// not being present or partial. The distinction is important, as if missing or
+// partial, this is just a temporary failure, as the block is still being
+// uploaded to the origin bucket.
+func loadMeta(ctx context.Context, rs *replicationScheme, id ulid.ULID) (*metadata.Meta, bool, error) {
```
bwplotka: Let's add `TODO(bwplotka): Move to block.MetaFetcher.`

pkg/replicator/replicater.go
```diff
@@ -0,0 +1,200 @@
...
+	logger = log.With(logger, "component", "replicate")
+
+	level.Debug(logger).Log("msg", "setting up metric http listen-group")
+
+	// register a metrics http service.
+	httpProbe := prober.NewHTTP()
+	s := http.New(logger, reg, component.Replicate, httpProbe)
+
+	g.Add(func() error {
+		level.Info(logger).Log("msg", "Listening for metrics", "address", httpMetricsBindAddr)
```
bwplotka: It's missing our probes (ready/healthy). Do you mind looking on another commands and copy similar here? (:

pkg/replicator/replicater.go
bwplotka: Yes, I agree. Sorry @daixiang0 for initially asking you to rename to `replicator`. `replicate` makes sense as per:  https://thanos.io/contributing/#components-naming

pkg/replicator/scheme.go
daixiang0: In fact, it calls MetaFetcher indeed.

cmd/thanos/bucket.go
bwplotka: Not resolved

cmd/thanos/bucket.go
bwplotka: Not resolved.

mixin/thanos/dashboards/replicate.libsonnet
```diff
@@ -0,0 +1,58 @@
+local g = import '../lib/thanos-grafana-builder/builder.libsonnet';
+
+{
+  local thanos = self,
+  replicator+:: {
```
bwplotka: `replicator?`

bwplotka: @daixiang0 please look carefully on details. It's sometimes hard but details matter (:

cmd/thanos/bucket.go
daixiang0: In fact it does.

cmd/thanos/bucket.go
daixiang0: In fact it does.

mixin/thanos/dashboards/replicate.libsonnet
daixiang0: I miss the mixin part since they are in process indeed, sorry for pushing them :)

daixiang0:
@bwplotka hi, i do not understand why those codes are old when you review, they were resolved.
Anyway, i have confirmed them again.
For mixin part, i am still working on it.

bwplotka:
Thank You!

brancz:
Can you mark this PR as work in progress until it’s reviewable in its entirety? Thanks! :)

CHANGELOG.md
```diff
@@ -29,6 +29,7 @@ We use *breaking* word for marking changes that are not backward compatible (rel
 - [#2049](https://github.com/thanos-io/thanos/pull/2049) Tracing: Support sampling on Elastic APM with new sample_rate setting.
 - [#2008](https://github.com/thanos-io/thanos/pull/2008) Querier, Receiver, Sidecar, Store: Add gRPC [health check](https://github.com/grpc/grpc/blob/master/doc/health-checking.md) endpoints.
 - [#2145](https://github.com/thanos-io/thanos/pull/2145) Tracing: track query sent to prometheus via remote read api.
+- [#2113](https://github.com/thanos-io/thanos/pull/2113) Bucket: Added `thanos bucket replicate` (mixin todo).
```
brancz: looks like this was added now?

kakkoyun: Ow, I've missed this TODO. All good if we'll address it in another issue. Already added files mislead me.

daixiang0: Sorry, my mistake.

daixiang0:
@kakkoyun just rename or comment it in description?

daixiang0:
@kakkoyun please take a look :)

cmd/thanos/bucket.go
bwplotka: How? I proposed `resolution-level` (: I might be wrong, but you should comment that I am wrong, because of X not ignore it. But I am pretty sure you just did not notice (:

daixiang0: I thought that you mean add more level support for both resolution and compaction, so i added two TODOs here.

daixiang0: Kinpin supports `Int` but not `Int64`, so still need to use `String` here.

CHANGELOG.md
```diff
@@ -29,6 +29,7 @@ We use *breaking* word for marking changes that are not backward compatible (rel
 - [#2049](https://github.com/thanos-io/thanos/pull/2049) Tracing: Support sampling on Elastic APM with new sample_rate setting.
 - [#2008](https://github.com/thanos-io/thanos/pull/2008) Querier, Receiver, Sidecar, Store: Add gRPC [health check](https://github.com/grpc/grpc/blob/master/doc/health-checking.md) endpoints.
 - [#2145](https://github.com/thanos-io/thanos/pull/2145) Tracing: track query sent to prometheus via remote read api.
+- [#2113](https://github.com/thanos-io/thanos/pull/2113) Bucket: Added `thanos bucket replicate`.
```
bwplotka: It got into wrong release section, so we need to be careful when merging release-0.11 to master @metalmatze

PR #2123, opened 2020-02-12
docs: Document Thanos Receive

muryoutaisuu:

## Changes

1. added Thanos receive component documentation page

### DIFF
--- a/docs/components/receive.md
+++ b/docs/components/receive.md
@@ -1,0 +1,98 @@
+---
+title: Receive
+type: docs
+menu: components
+---
+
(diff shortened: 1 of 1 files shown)


muryoutaisuu:
Task of #1093 

pinging @brancz because of little discussion on slack.

This is a very slim draft for the receive component documentation. But, as already discussed on slack, the hashring (loadbalancing & tenancy) functionality feels like the biggest part of this component, with which I'm neither acquainted enough nor do we use it. Sorry therefore!

PR #2125, opened 2020-02-12
Close index-header and block properly, even on error case.

bwplotka:
Especially for binary index-header we were not closing mmaped file handle.


bwplotka:
We might need some unit tests for closing.. Deferring this for later work: https://github.com/thanos-io/thanos/issues/2126

PR #2131, opened 2020-02-13
add full overlap block detection algorithm

khyatisoneji:


bwplotka:
I changed name while merging to `proposal: Updated RW proposal`

PR #2135, opened 2020-02-13
dynamic aws credentials support

kush-patel-hs:
DO NOT MERGE. This is a proposal design and needs work.  

## Changes

Will watch for aws credential changes

## Verification

If we are good with this design then we can work on verification

### DIFF
--- a/pkg/objstore/s3/s3.go
+++ b/pkg/objstore/s3/s3.go
@@ -151,7 +151,7 @@ func NewBucketWithConfig(logger log.Logger, config Config, component string) (*B
 		startCredentialExpirer(creds, filename)
 	}
 
-	client, err := minio.NewWithCredentials(config.Endpoint, credentials.NewChainCredentials(chain), !config.Insecure, config.Region)
+	client, err := minio.NewWithCredentials(config.Endpoint, creds, !config.Insecure, config.Region)
 	if err != nil {
 		return nil, errors.Wrap(err, "initialize s3 client")
 	}
--- a/pkg/objstore/s3/s3.go
+++ b/pkg/objstore/s3/s3.go
@@ -148,7 +148,9 @@ func NewBucketWithConfig(logger log.Logger, config Config, component string) (*B
 				filename = filepath.Join(homeDir, ".aws", "credentials")
 			}
 		}
-		startCredentialExpirer(creds, filename)
+		if err := startCredentialExpirer(creds, filename); err != nil {
+			logger.Log("msg", "failed to start credential expirer", err)
+		}
 	}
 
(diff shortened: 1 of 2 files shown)


pkg/objstore/s3/s3.go
```diff
@@ -134,6 +138,17 @@ func NewBucketWithConfig(logger log.Logger, config Config, component string) (*B
 				},
 			},
 		}
+		creds = credentials.NewChainCredentials(chain)
+
+		// We will watch for credential changes for non-static credentials
+		filename := os.Getenv("AWS_SHARED_CREDENTIALS_FILE")
```
kush-patel-hs: We should probably also get filename from config if we can. Then provide that to FileAWSCredentials. I really hate having to duplicate the FileAWSCredentials logic to determine filename here.

pkg/objstore/s3/s3.go
```diff
@@ -192,6 +207,46 @@ func (b *Bucket) Name() string {
...
+		return nil
+	}
+
+	watcher, err := fsnotify.NewWatcher()
+	if err != nil {
+		return err
+	}
+
+	// Launch the watch which will expire the credentials when fired
+	go func() {
```
kush-patel-hs: We need to kill this go routine when thanos quit or dies. Is there a standard way thanos does this atm for other go routines?

pkg/objstore/s3/s3.go
```diff
@@ -134,6 +138,17 @@ func NewBucketWithConfig(logger log.Logger, config Config, component string) (*B
...
+
+		// We will watch for credential changes for non-static credentials
+		filename := os.Getenv("AWS_SHARED_CREDENTIALS_FILE")
+		if filename == "" {
+			homeDir, err := homedir.Dir()
+			if err == nil {
+				filename = filepath.Join(homeDir, ".aws", "credentials")
+			}
+		}
+		startCredentialExpirer(creds, filename)
```
kush-patel-hs: log if there is an error starting the watcher maybe

kush-patel-hs:
Thanks for taking a look! For our use case it is what we want. What we do is we have our service. Then we have a "Vault Sidecar" which refreshes AWS credentials and will write the refreshed credentials to a directory. So we wouldn't be changing the Thanos config when this happens and would be a lot of work for us to do. So we would rely on this else block https://github.com/thanos-io/thanos/pull/2135/files#diff-bef8b149473cb313205c67f945d249ceR131 and reloadable thanos config would not help unless we reload ~/.aws/credentials (or the file path in the env var).

kush-patel-hs:
closing

PR #2139, opened 2020-02-14
*: Handle SIGHUP

### LINKED ISSUE #1333: Handle SIGHUP event

thanos: v0.5.0

**What happened**

Sending a SIGHUP signal to sidecar, store gateway, querier or compactor crashes it.

They should not crash. Maybe they should also reload configuration (?)

```sh
$ kill -SIGHUP <thanos>
```

I think thanos should at least not crash upon receiving a SIGHUP signal.

I understand that flags are used to configure thanos components, but it would be nice to have a way to reload configuration files immediately with a SIGHUP signal. This is what most applications do (eg: prometheus). Potential use case: Have an ansible handler sending the SIGHUP signal.

kakkoyun:
This PR handles received SIGHUP signals and dispatches reload events to sub-commands to handle.
Only the ruler utilizes this right now. Receiver and Ruler implement their own file watchers to reload configurations.

Fixes #1333.

## Verification

* `make test-local`
* MINIO_ENABLED=1 ./scripts/quickstart.sh
  - `kill -HUP <thanos-pid>`

### DIFF
--- a/cmd/thanos/main.go
+++ b/cmd/thanos/main.go
@@ -233,8 +233,8 @@ func reload(logger log.Logger, cancel <-chan struct{}, r chan<- struct{}) error 
 			level.Info(logger).Log("msg", "caught signal. Reloading.", "signal", s)
 			select {
 			case r <- struct{}{}:
+				level.Info(logger).Log("msg", "relaod dispatched.")
 			default:
-				level.Info(logger).Log("msg", "already reloading.")
 			}
 		case <-cancel:
 			return errors.New("canceled")
--- a/cmd/thanos/main.go
+++ b/cmd/thanos/main.go
@@ -178,7 +178,8 @@ func main() {
 		})
 	}
 
-	reloadCh := make(chan struct{})
+	// Create a signal channel to dispatch reload events to sub-commands.
+	reloadCh := make(chan struct{}, 1)
 
 	if err := cmds[cmd](&g, logger, metrics, tracer, reloadCh, *logLevel == "debug"); err != nil {
 		level.Error(logger).Log("err", errors.Wrapf(err, "%s command failed", cmd))
--- a/cmd/thanos/main.go
+++ b/cmd/thanos/main.go
@@ -32,7 +32,7 @@ const (
 	logFormatJson   = "json"
 )
 
-type setupFunc func(*run.Group, log.Logger, *prometheus.Registry, opentracing.Tracer, bool) error
+type setupFunc func(*run.Group, log.Logger, *prometheus.Registry, opentracing.Tracer, <-chan struct{}, bool) error
 
 func main() {
 	if os.Getenv("DEBUG") != "" {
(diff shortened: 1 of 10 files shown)


cmd/thanos/main.go
```diff
@@ -178,7 +178,9 @@ func main() {
 		})
 	}
 
-	if err := cmds[cmd](&g, logger, metrics, tracer, *logLevel == "debug"); err != nil {
+	reloadCh := make(chan struct{})
```
kakkoyun: This should be a buffered channel with a capacity of 1. Will fix it.

cmd/thanos/main.go
```diff
@@ -211,3 +224,21 @@ func interrupt(logger log.Logger, cancel <-chan struct{}) error {
...
+
+func reload(logger log.Logger, cancel <-chan struct{}, r chan<- struct{}) error {
+	c := make(chan os.Signal, 1)
+	signal.Notify(c, syscall.SIGHUP)
+	for {
+		select {
+		case s := <-c:
+			level.Info(logger).Log("msg", "caught signal. Reloading.", "signal", s)
+			select {
+			case r <- struct{}{}:
```
bwplotka: What if no one is receiving here?

bwplotka: I think we mind this 
```
{
		cancel := make(chan struct{})
		g.Add(func() error {
			return reload(logger, cancel, reloadCh)
		}, func(error) {
			close(cancel)
		})
	}
```
To be per command maybe? Or noop otherwise? That's the reason and also we might think of redesigning commands and the number of arguments is becoming a problem (:

kakkoyun: > What if no one is receiving here?

When no one is receiving that means (or if someone's receiving), we just allocate 4 bytes of memory for the full lifecycle of the program. It's unnecessary if no one receiving of course. Empty structs allocate zero bytes of memory, fortunately. (so allocation is just for the channel)
https://play.golang.org/p/2QPjVBQZSGU

That being said I'll have another look at it to avoid unnecessary allocation.

> To be per command maybe? Or noop otherwise? 

I will have another look at this, find a clean way to achieve it.
I just don't want to handle signals outside the `main.go`, it makes things harder to pinpoint.

> That's the reason and also we might think of redesigning commands and the number of arguments is becoming a problem (:

Yes, we definitely need to find more elegant solution to this parameter problem. As once a great man said:

> “If you have a procedure with 10 parameters, you probably missed some.”
> —Allan Perlis

🤪

bwplotka: > When no one is receiving that means (or if someone's receiving), we just allocate 4 bytes of memory for the full lifecycle of the program

I don't think only 4 bytes. I am more worried on signal channel.Whatt if no one is receiving and I will send 100mln of SIGHUP signals?

bwplotka: For redesign we probably need to finally move to some structure (: I remember Cortex had some quite nice modules structure.. worth to get inspired? Cc @gouthamve

kakkoyun: > Whatt if no one is receiving and I will send 100mln of SIGHUP signals?
That’s why it’s in an additional select with default case so that it can just continue on if the buffered (reload) channel is full. So memory will always be constant, it’ll just loop over 100mln.

bwplotka: You are right, I missed this default case. Thanks! ((:

kakkoyun:
@bwplotka Should I move handling logic per subcommand? or is this good enough?

PR #2143, opened 2020-02-17
Makefile: add faillint to lint target - Issue 2142

### LINKED ISSUE #2142: Incorporate failinit.
Hi, 

I would propose adding https://github.com/fatih/faillint/blob/master/README.md to our 'make lint' rule to avoid some inconsistencies like some packages we would like to not import:

* errors (we use pkg/errors)
* prometheus/tsdb (it was archived and moved to Prometheus module)
* prometheus/prometheus/pkg/testutils (we have our own)

Help wanted 🤗

kadern0:
Fixes #2142 

(I've included a line at the bottom of the Added section)

## Changes

Added failint package with the paths specified within the issue description to 'make lint'

## Verification
Manual execution of the command and output verification.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -425,9 +425,7 @@ $(MINIO_SERVER):
 	$(call fetch_go_bin_version,github.com/minio/minio,$(MINIO_SERVER_VERSION))
 
 $(FAILLINT):
-	@go get github.com/fatih/faillint@$(FAILLINT_VERSION)
-	@mv $(GOBIN)/faillint $(FAILLINT)
-	@go mod tidy
+	$(call fetch_go_bin_version,github.com/fatih/faillint,$(FAILLINT_VERSION))
 
 $(PROMS):
 	$(foreach ver,$(PROM_VERSIONS),$(call fetch_go_bin_version,github.com/prometheus/prometheus/cmd/prometheus,$(ver)))
--- a/Makefile
+++ b/Makefile
@@ -78,6 +78,10 @@ ALERTMANAGER            ?= $(GOBIN)/alertmanager-$(ALERTMANAGER_VERSION)
 MINIO_SERVER_VERSION    ?= RELEASE.2018-10-06T00-15-16Z
 MINIO_SERVER            ?=$(GOBIN)/minio-$(MINIO_SERVER_VERSION)
 
+FAILLINT_VERSION        ?= v1.0.1
+FAILLINT                ?=$(GOBIN)/faillint-$(FAILLINT_VERSION)
+MODULES_TO_AVOID        ?=errors,github.com/prometheus/tsdb,github.com/prometheus/prometheus/pkg/testutils
+
 # fetch_go_bin_version downloads (go gets) the binary from specific version and installs it in $(GOBIN)/<bin>-<version>
 # arguments:
 # $(1): Install path. (e.g github.com/campoy/embedmd)
@@ -290,7 +294,9 @@ web: web-pre-process $(HUGO)
 #      --mem-profile-path string   Path to memory profile output file
 #
 # to debug big allocations during linting.
-lint: check-git $(GOLANGCILINT) $(MISSPELL)
+lint: check-git $(GOLANGCILINT) $(MISSPELL) $(FAILLINT)
+	@echo ">> verifying modules being imported"
+	@$(FAILLINT) -paths $(MODULES_TO_AVOID) ./...
 	@echo ">> examining all of the Go files"
 	@go vet -stdmethods=false ./pkg/... ./cmd/... && go vet doc.go
 	@echo ">> linting all of the Go files GOGC=${GOGC}"
@@ -417,6 +423,11 @@ $(ALERTMANAGER):
 
 $(MINIO_SERVER):
 	$(call fetch_go_bin_version,github.com/minio/minio,$(MINIO_SERVER_VERSION))
+
+$(FAILLINT):
+	@go get github.com/fatih/faillint@$(FAILLINT_VERSION)
+	@mv $(GOBIN)/faillint $(FAILLINT)
+	@go mod tidy
 
 $(PROMS):
 	$(foreach ver,$(PROM_VERSIONS),$(call fetch_go_bin_version,github.com/prometheus/prometheus/cmd/prometheus,$(ver)))
--- a/go.mod
+++ b/go.mod
@@ -66,7 +66,7 @@ require (
 	github.com/onsi/gomega v1.7.1 // indirect
 	github.com/opentracing/basictracer-go v1.0.0
 	github.com/opentracing/opentracing-go v1.1.0
-	github.com/pkg/errors v0.8.1
+	github.com/pkg/errors v0.9.1
 	github.com/prometheus/alertmanager v0.20.0
 	github.com/prometheus/client_golang v1.2.1
 	github.com/prometheus/client_model v0.0.0-20190812154241-14fe0d1b01d4
@@ -92,6 +92,7 @@ require (
 	golang.org/x/sys v0.0.0-20191113165036-4c7a9d0fe056 // indirect
 	golang.org/x/text v0.3.2
 	golang.org/x/time v0.0.0-20191024005414-555d28b269f0 // indirect
+	golang.org/x/tools v0.0.0-20200207224406-61798d64f025 // indirect
 	google.golang.org/api v0.14.0
 	google.golang.org/appengine v1.6.5 // indirect
 	google.golang.org/genproto v0.0.0-20191115194625-c23dd37a84c9
--- a/pkg/block/fetcher.go
+++ b/pkg/block/fetcher.go
@@ -6,7 +6,6 @@ package block
 import (
 	"context"
 	"encoding/json"
-	stderrors "errors"
 	"io/ioutil"
 	"os"
 	"path"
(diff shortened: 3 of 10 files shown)


kadern0:
This is the output of the command:

```
make lint
>> verifying packages being imported
/home/kaderno/go/src/github.com/improbable-eng/thanos/pkg/block/fetcher.go:9:12: package "errors" shouldn't be imported
/home/kaderno/go/src/github.com/improbable-eng/thanos/pkg/receive/hashring.go:8:2: package "errors" shouldn't be imported
/home/kaderno/go/src/github.com/improbable-eng/thanos/pkg/cacheutil/memcached_client_test.go:8:2: package "errors" shouldn't be imported
/home/kaderno/go/src/github.com/improbable-eng/thanos/pkg/query/api/v1_test.go:22:2: package "errors" shouldn't be imported
/home/kaderno/go/src/github.com/improbable-eng/thanos/pkg/runutil/example_test.go:8:2: package "errors" shouldn't be imported
/home/kaderno/go/src/github.com/improbable-eng/thanos/pkg/store/cache/memcached_test.go:8:2: package "errors" shouldn't be imported
/home/kaderno/go/src/github.com/improbable-eng/thanos/pkg/store/storepb/custom_test.go:7:2: package "errors" shouldn't be imported
```
All these packages import "errors", so they should be updated before merging this PR or builds will fail.

Makefile
```diff
@@ -78,6 +78,8 @@ ALERTMANAGER            ?= $(GOBIN)/alertmanager-$(ALERTMANAGER_VERSION)
 MINIO_SERVER_VERSION    ?= RELEASE.2018-10-06T00-15-16Z
 MINIO_SERVER            ?=$(GOBIN)/minio-$(MINIO_SERVER_VERSION)
 
+IMPORTS_TO_AVOID				?=errors,prometheus/tsdb,prometheus/prometheus/pkg/testutils
```
bwplotka: I think it would be nice to have full modules path. E.g not everything is hosted in `github.com` domain.

Also: 
```suggestion
MODULES_TO_AVOID				?=errors,prometheus/tsdb,prometheus/prometheus/pkg/testutils
```

Makefile
```diff
@@ -283,14 +285,18 @@ web: web-pre-process $(HUGO)
 	# TODO(bwplotka): Make it --gc
 	@cd $(WEB_DIR) && HUGO_ENV=production $(HUGO) --config hugo.yaml --minify -v -b $(WEBSITE_BASE_URL)
 
+.PHONY: failint
+failint:
```
bwplotka: Please let's add this to `lint` command. We already have too many commands ;p

Makefile
```diff
@@ -283,14 +285,18 @@ web: web-pre-process $(HUGO)
...
+	@echo ">> verifying packages being imported"
+	@faillint -paths $(IMPORTS_TO_AVOID) ./...
```
bwplotka: What about vendor? I think we should exclude vendor

Makefile
bwplotka: I I know you added this as lint dep, but I  think it might be more explicit when inlined in `lint` command.

Makefile
bwplotka: Also let's install `failint` as we do for other Go binaries.

go.mod
```diff
@@ -66,7 +66,7 @@ require (
 	github.com/onsi/gomega v1.7.1 // indirect
 	github.com/opentracing/basictracer-go v1.0.0
 	github.com/opentracing/opentracing-go v1.1.0
-	github.com/pkg/errors v0.8.1
+	github.com/pkg/errors v0.9.1
```
bwplotka: Why we upgraded this module on the way? It looks like irrelevant change.

Makefile
kadern0: It makes moving it inside of the `lint` command although then we would have to wait for the whole linting to complete whenever we want to test (not a huge deal).

Makefile
kadern0: I've followed the "Dependency management" section under the contributing guide. I've realized that if I run `make deps` it removes the `faillint` entries from `go.sum` and `go.mod`

go.mod
kadern0: Don't know when this was changed. I thought it was some automatic dependency resolution.

Makefile
kadern0: If we don't want to run it across all folders, we would have to specify the files and directories manually (or a for loop): pkg, cmd, test, scripts and doc.go. Other option could be, opening a PR to the faillint project with this idea.

What do you think?

bwplotka: ah.. TBH can we do that? Can we ask question on upstream project how to avoid vendor dir?

bwplotka: Maybe it already support this?

kadern0: Current version has no option like that. I'm going to ask in the upstream project to see what they recommend.

kadern0: According to the project developer, faillint excludes the `vendor` directory automatically.

daixiang0:
Would it be better to change pr title as "Makefile: add faillint to lint target"?

kadern0:
> Would it be better to change pr title as "Makefile: add faillint to lint target"?

Yeah, I like it better. Should I keep the reference to the issue No.?

Makefile
bwplotka: I think that's ok (:

bwplotka: You can put it in front of it if you think it's fast.

Makefile
bwplotka: Thanks for checking :heart:

bwplotka: And that's ok - check the binaries like `$(MINIO_SERVER)` and how it install it and version. (: Can we mimic that?

go.mod
bwplotka: Let's remove then (:

CHANGELOG.md
```diff
@@ -678,3 +678,4 @@ Initial version to have a stable reference before [gossip protocol removal](/doc
 - Bucket commands.
 - Downsampling support for UI.
 - Grafana dashboards for Thanos components.
+- Failint for imported packages verification.
```
bwplotka: I think we should not touch 0.1.0 release  ;p

Also this is a tooling, not really relevant for Thanos users, so we don't need changelog entry. (:

Makefile
kadern0: I think I've fixed this, please let me know if you were thinking something different when you said to put it in front.

Makefile
kadern0: Done (but check it, please)!

go.mod
kadern0: Removed

CHANGELOG.md
kadern0: I was following the guide :) Removed!

kadern0: About not touching the 0.1.0 version, should I change the PR to a different branch instead of master?

bwplotka: No. Master is fine. Just you put the changelog entry at the bottom, but actually releases goes from newest to oldest.

Anyway, there should be no entry.

kadern0:
One of the tests was failing because of incorrect signing off of commits.  The other test fails because of what I've mentioned in my second comment on this PR. There are 7 packages importing `errors`, thus, the linting stage fails. Now that the "setup" is working, it might be appropriate, fixing these files, what do you think?

Also, yes, I've tested the correct behavior of the code (with different modules and also placing them under a directory called `vendor`) and it seems to be working as expected.

kadern0:
If I update this line:
`MODULES_TO_AVOID        ?=errors,prometheus/tsdb,prometheus/prometheus/pkg/testutils`
to:
`MODULES_TO_AVOID        ?=errors=pkg/errors,prometheus/tsdb,prometheus/prometheus/pkg/testutils`

Then, in the output we get a suggestion of the correct module to import:
`improbable-eng/thanos/pkg/block/fetcher.go:9:12: package "errors" shouldn't be imported, suggested: "pkg/errors"` 

Should we add the recommended modules for each of the ones to avoid?

bwplotka:
That would be amazing! Nice idea @kadern0!

But please let's use the full path for imports if we can?

Looks like we detected errors!
```
 /go/src/github.com/thanos-io/thanos/pkg/block/fetcher.go:9:12: package "errors" shouldn't be imported
/go/src/github.com/thanos-io/thanos/pkg/receive/hashring.go:8:2: package "errors" shouldn't be imported
/go/src/github.com/thanos-io/thanos/pkg/cacheutil/memcached_client_test.go:8:2: package "errors" shouldn't be imported
/go/src/github.com/thanos-io/thanos/pkg/query/api/v1_test.go:22:2: package "errors" shouldn't be imported
/go/src/github.com/thanos-io/thanos/pkg/runutil/example_test.go:8:2: package "errors" shouldn't be imported
/go/src/github.com/thanos-io/thanos/pkg/store/cache/memcached_test.go:8:2: package "errors" shouldn't be imported
/go/src/github.com/thanos-io/thanos/pkg/store/storepb/custom_test.go:7:2: package "errors" shouldn't be imported
```

Do you mind changing those?

Good job, can't wait to merge it! (:

kadern0:
I've changed all the imports but now circleci is failing for some reason I don't understand. Could you please have a look?

Makefile
```diff
@@ -418,6 +424,11 @@ $(ALERTMANAGER):
 $(MINIO_SERVER):
 	$(call fetch_go_bin_version,github.com/minio/minio,$(MINIO_SERVER_VERSION))
 
+$(FAILLINT):
+	@go get github.com/fatih/faillint@$(FAILLINT_VERSION)
+	@mv $(GOBIN)/faillint $(FAILLINT)
+	@go mod tidy
```
yeya24: Maybe the problem is that the go mod file is changed because of downloading faillint. What about using `fetch_go_bin_version` like we downloading other binaries?

kadern0: Fixed

go.mod
bwplotka: This is actually needed (:

pkg/block/fetcher.go
```diff
@@ -185,7 +184,7 @@ func (s *MetaFetcher) loadMeta(ctx context.Context, id ulid.ULID) (*metadata.Met
 			return m, nil
 		}
 
-		if !stderrors.Is(err, os.ErrNotExist) {
+		if !errors.Is(err, os.ErrNotExist) {
```
bwplotka: FYI: This actually works thanks to: https://github.com/pkg/errors/commit/6d954f502eb89cd315e4baae5b0e0db516d6f787

PR #2150, opened 2020-02-18
UI: rework graph page

vankop:
Signed-off-by: Ivan Kopeykin <ikopeykin@ozon.ru>

## Changes

1. Throttling user input,
2. Add minimum query length == 3,
3. Maximum amount of suggestions to show from 1000 to 20/30,
4. Improve performance of fuzzy matching a lot (less allocations, run matching on previous results if possible)

## Verification

Tested in our environment with more than 500_000 metrics

/cc @d-ulyanov


d-ulyanov:
We've fixed UI a bit in our fork, because UI getting stuck always right after opening. For us, this bug is a huge pain, so we decided to fix it and share.

I'm not sure that it's legal to make such changes in Thanos without making it in Prometheus UI first, just curious what do you think @bwplotka ?

GiedriusS:
I believe this is yet another attempt at implementing https://github.com/prometheus/prometheus/issues/5421 and https://github.com/prometheus/prometheus/issues/5919. We should really probably do this or something similar. It seems like this https://github.com/prometheus/prometheus/pull/6100 upstream never got any attention. Maybe @bwplotka has some new information regarding it? If not, then we should probably really finally do this ourselves, half a year after.

bwplotka:
Yes, let's do this. The conclusion upstream was: "Let's fix this in
ReactUI"

Happy to fix this in both React and old one for Thanos!

vankop:
@GiedriusS regarding to https://github.com/prometheus/prometheus/pull/6100 I can also fix initial page freezing in this PR. In upstream it fixed [here](https://github.com/prometheus/prometheus/pull/6100/files#diff-29559b5de4c061ac1d6785e74d286ae4R279-R284)

d-ulyanov:
@vankop please, rebase and let's merge it 🚀
I'm so happy to work with Thanos UI after that fix :)

GiedriusS:
Still valid. Seems like there's a conflict :/

PR #2162, opened 2020-02-20
compactor: Put important compaction logs as info; init object store metrics.

bwplotka:

### DIFF
--- a/pkg/objstore/objstore.go
+++ b/pkg/objstore/objstore.go
@@ -188,6 +188,16 @@ func Exists(ctx context.Context, bkt Bucket, src string) (bool, error) {
 	return true, nil
 }
 
+const (
+	iterOp     = "iter"
+	sizeOp     = "objectsize"
+	getOp      = "get"
+	getRangeOp = "get_range"
+	existsOp   = "exists"
+	uploadOp   = "upload"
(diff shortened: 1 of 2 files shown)


pkg/objstore/objstore.go
```diff
@@ -219,6 +229,12 @@ func BucketWithMetrics(name string, b Bucket, r prometheus.Registerer) Bucket {
 	}
 	if r != nil {
 		r.MustRegister(bkt.ops, bkt.opsFailures, bkt.opsDuration, bkt.lastSuccessfullUploadTime)
+		for _, op := range []string{iterOp, sizeOp, getOp, getRangeOp, existsOp, uploadOp, deleteOp} {
+			bkt.ops.WithLabelValues(op)
+			bkt.opsFailures.WithLabelValues(op)
+			bkt.opsDuration.WithLabelValues(op)
+		}
+		bkt.lastSuccessfullUploadTime.WithLabelValues(b.Name())
```
GiedriusS: I know you haven't introduced this but maybe we should fix this typo as well while changing this so that it would be `lastSuccessfulUploadTime`? :P

bwplotka: :+1:

PR #2163, opened 2020-02-20
receive: Validate hashring at initialization

### LINKED ISSUE #2160: receiver: Validate hashring config.
See: https://github.com/thanos-io/thanos/pull/2138#discussion_r381330638

kakkoyun:
This PR checks if the hashring config is empty and fails if it is.

Fixes #2160 

cc @squat @bwplotka 

## Changes

* checks if the hashring config is empty and fails if it is.

## Verification

* `make test-local`
* Run `REMOTE_WRITE_ENABLED=1 MINIO_ENABLED=1 ./scripts/quickstart.sh` with several modified versions of hashring file.

### DIFF
--- a/cmd/thanos/receive.go
+++ b/cmd/thanos/receive.go
@@ -321,9 +321,7 @@ func runReceive(
 		if cw != nil {
 			ctx, cancel := context.WithCancel(context.Background())
 			g.Add(func() error {
-				// TODO: If config is empty we never start receiver. We might want to fail in this case.
-				receive.HashringFromConfig(ctx, updates, cw)
-				return nil
+				return receive.HashringFromConfig(ctx, updates, cw)
 			}, func(error) {
 				cancel()
 			})
--- a/cmd/thanos/receive.go
+++ b/cmd/thanos/receive.go
@@ -320,7 +320,7 @@ func runReceive(
 		updates := make(chan receive.Hashring, 1)
 
 		if cw != nil {
-			// Before scheduling a job, check if hashring config is valid.
+			// Check the hashring configuration on before running the watcher.
 			if err := cw.ValidateConfig(); err != nil {
 				close(updates)
 				return errors.Wrap(err, "failed to validate hashring configuration file")
--- a/cmd/thanos/receive.go
+++ b/cmd/thanos/receive.go
@@ -318,7 +318,14 @@ func runReceive(
 		// In the single-node case, which has no configuration
 		// watcher, we close the chan ourselves.
 		updates := make(chan receive.Hashring, 1)
+
 		if cw != nil {
+			// Before scheduling a job, check if hashring config is valid.
+			if err := cw.ValidateConfig(); err != nil {
+				close(updates)
+				return errors.Wrap(err, "failed to validate hashring configuration file")
+			}
+
 			ctx, cancel := context.WithCancel(context.Background())
 			g.Add(func() error {
 				return receive.HashringFromConfig(ctx, updates, cw)
--- a/pkg/receive/config.go
+++ b/pkg/receive/config.go
@@ -10,7 +10,6 @@ import (
 	"encoding/json"
 	"io/ioutil"
 	"os"
-	"reflect"
 	"time"
 
 	"github.com/go-kit/kit/log"
@@ -47,8 +46,8 @@ type ConfigWatcher struct {
 	hashringNodesGauge   *prometheus.GaugeVec
 	hashringTenantsGauge *prometheus.GaugeVec
 
-	// last is the last known configuration.
-	last []HashringConfig
+	// lastConfigHash is the last known hash of the loaded configuration.
+	lastConfigHash float64
 }
 
 // NewConfigWatcher creates a new ConfigWatcher.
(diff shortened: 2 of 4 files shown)


pkg/receive/config.go
```diff
@@ -21,6 +20,10 @@ import (
 	"gopkg.in/fsnotify.v1"
 )
 
+type EmptyHashringError string
+
+func (e EmptyHashringError) Error() string { return "hashring is empty" }
```
squat: This should not be a type but rather a new variable eg
```go
var EmptyHashringConfig = errors.New(...)
```

pkg/receive/config.go
```diff
@@ -21,6 +20,10 @@ import (
 	"gopkg.in/fsnotify.v1"
 )
 
+type EmptyHashringError string
```
squat: Let's add a comment here

squat: Should probably be empty config error than hashring

cmd/thanos/receive.go
```diff
@@ -315,11 +315,17 @@ func runReceive(
 		// In the single-node case, which has no configuration
 		// watcher, we close the chan ourselves.
 		updates := make(chan receive.Hashring, 1)
+
 		if cw != nil {
+			// Before scheduling a job, check if hashring config is valid.
```
squat: Scheduling which job? To clarify maybe consider "Check the hashring configurat on before running the watcher."

pkg/receive/config.go
```diff
@@ -180,55 +184,56 @@ func (cw *ConfigWatcher) C() <-chan []HashringConfig {
 	return cw.ch
 }
 
-// readFile reads the configured file and returns content of configuration file.
-func (cw *ConfigWatcher) readFile() ([]byte, error) {
-	fd, err := os.Open(cw.path)
+// ValidateConfig returns an error if configuration that's being watched is not valid.
```
squat: "if the configuration"

pkg/receive/config.go
```diff
@@ -271,6 +276,28 @@ func (cw *ConfigWatcher) stop() {
 	level.Debug(cw.logger).Log("msg", "hashring configuration watcher stopped")
 }
 
+// readFile reads the configured file and returns content of configuration file.
+func (cw *ConfigWatcher) readFile() ([]byte, error) {
```
squat: s/configured/configuration/

pkg/receive/config.go
kakkoyun: I did this on purpose/ to use `errors.As` in the tests. I needed a type because it was more convenient to use `errors.As` and error types from `encoding/json` package. See `config_test.go`.

I didn't want to just test against the existence of an error, I wanted to be more specific.

Also even though go standard library uses package-level variables as sentinel errors, I'm not a fan. I think using constant errors or types is better for these type of errors.
 
All those being said, I also find it a bit ugly, however, I couldn't find a more elegant way, I'm open to suggestion.

squat: This is actually quite clean and easy to achieve thanks to the new errors package using errors.Is. Here is an example I put together: https://play.golang.org/p/3itSVfLjtNX

kakkoyun: Yes, for the known error from this package, this is a more elegant solution and easy to test with

However, the problem arises when you want to test against wrapped errors, which we use in `config_test.go`.  

```go
config, err := cw.parseConfig(cfgContent)
if err != nil {
	return nil, 0, errors.Wrap(err, "failed to load configuration file")
}
// ...

// config_test.go
// ...
{
	name: "unparsable config",
	cfg:  struct{}{},
	err:  json.UnsupportedTypeError{},
},

// ...
if err := cw.ValidateConfig(); err != nil && !errors.As(err, &tc.err) {
	t.Errorf("case %q: got unexpected error: %v", tc.name, err)
	continue
}
```
To test against wrapped error type I've used `errors.As`, hence the  empty config error's defined as a type.

squat: Note in the example I have you it IS a wrapped error. We wrap the FooError with `fmt.Errorf("bar: %w", err)`. errors.Is let's you check if the error that was wrapped is a specific error

kakkoyun: Yes, I'm well aware of that. I couldn't explain my problem clearly enough. I don't have the error value from the underlying `json` package, I only have the type that's why I couldn't use `errors.Is` for that specific case. 

I'll find another way to solve this, it's solely for testing anyway. 
Thanks for the feedback.

pkg/receive/hashring.go
```diff
@@ -165,19 +165,19 @@ func newMultiHashring(cfg []HashringConfig) Hashring {
...
+func HashringFromConfig(ctx context.Context, updates chan<- Hashring, cw *ConfigWatcher) error {
 	defer close(updates)
+	go cw.Run(ctx)
 
 	for {
 		select {
 		case cfg, ok := <-cw.C():
 			if !ok {
-				return
+				return errors.New("hashring config watcher stopped")
```
squat: Can we make one tiny change here and add `unexpectedly` to the end? Just to highlight that this should only ever occur if the config watcher stopped irregularly. The only other exit scenario is a planned exit when the context is cancelled.

Issue #946, opened 2019-03-19
Support additional SSE methods for S3


taharah:
**Thanos, Prometheus and Golang version used**

**Thanos:** 0.3.2
**Prometheus:** 2.7.2
**Golang**: 1.11.5

**What happened**

Currently, the S3 bucket client only supports using the default key for SSE [[1](https://github.com/improbable-eng/thanos/blob/master/pkg/objstore/s3/s3.go#L151-L153)]. 

**What you expected to happen**

The S3 bucket client should be able to also support SSE-KMS and SSE-C encryption [[2](https://godoc.org/github.com/minio/minio-go/pkg/encrypt#ServerSide)]. 

**How to reproduce it (as minimally and precisely as possible)**:

N/A

**Full logs to relevant components**

N/A 

**Anything else we need to know**

I took a look at the code and this does introduce the opportunity to modify the existing SSE configuration in order to be a little cleaner as well as support the additional encryption methods.

For instance, the current configuration is this:

```yaml
type: S3
config:
  encrypt_sse: false
```

I think it would be nice if we changed the configuration to something like the following:

```yaml
type: S3
config:
  server_side_encryption:
    enabled: false
    kms_key_id: ""  # Used for SSE-KMS encrytpion
    encryption_key: ""  # Used for SSE-C encryption
```

NOTE: I've included only the relevant portions in the above examples.

However, the change could be made passive by just adding the parameters to the existing configuration at the top level.

adrien-f:
Hi :wave: !

This is an interesting feature. So we need to support three cases:

* If no `kms_key_id` and no `encryption_key` but `enabled` then `NewSSE`
* If `kms_key_id` and `enabled` then `NewSSEKMS`. Do you know what's the context parameter for ?
* If `encryption_key` and `enabled` then `NewSSEC`

Seems clear enough ! What do you think ?

taharah:
@adrien-f that looks good to me and what I was thinking. Additionally, the [KMS encryption context](https://docs.aws.amazon.com/kms/latest/developerguide/concepts.html?shortFooter=true#encrypt_context) is an optional set of key/value pairs that you can send with the requests.

jujugrrr:
hi, it seems like a basic feature if you use S3 with server side encryption. @adrien-f Do you think it would be possible to work on this?

@taharah did you just use the default sse after all?

bwplotka:
Valid request, help wanted (: 

:+1:

jujugrrr:
[...]

PR #2170, opened 2020-02-22
#946: Implement SSE-KMS and SSE-C for AWS S3

alex-laties:
Implements #946 

* [] I added CHANGELOG entry for this change.
* [] Change is not relevant to the end user.

## Changes

Added support for parsing and passing along AWS S3 SSE-KMS and SSE-C mode configuration details. Seemed pretty easy.

Adds the following config vars:
 - `sse_s3` - essentially the new `encrypt_sse`, enables SSE-S3 mode.
 - `sse_kms_id` - enables SSE-KMS mode when set.
 - `sse_kms_context` - optional map of string keys and values for the KMS Context.
 - `sse_c_key` - enables SSE-C mode when set. expects a filepath.

`encrypt_sse` is kept as an alias for `sse_s3`, as that's what `encrypt_sse` currently does, and it seems silly to break valid configuration files from prior to this change for little reason.

Note that KMS and C mode both supersede and override SSE-S3 mode currently. This allows for syntax like:
```
encrypt_sse: true
sse_c_key: "/path/to/my/key"
```
to work as expected.

KMS and C mode are mutually exclusive, and trying to enable both currently throws an error.
E.G. 
```
# this is invalid and will throw an error during validation
sse_c_key: "/path/to/my/key"
sse_kms_id: "myid"
```

## Verification

Haven't yet really... would like some help from @jujugrrr to validate the code does what I think it does before going any further really.

### DIFF
--- a/pkg/objstore/s3/s3.go
+++ b/pkg/objstore/s3/s3.go
@@ -9,6 +9,7 @@ import (
 	"crypto/tls"
 	"fmt"
 	"io"
+	"io/ioutil"
 	"net"
 	"net/http"
 	"os"
@@ -34,6 +35,8 @@ import (
 // DirDelim is the delimiter used to model a directory structure in an object store bucket.
 const DirDelim = "/"
 
+var ErrSSEConfig = fmt.Errorf("sse_kms_id and sse_c_key are both set")
+
 var DefaultConfig = Config{
 	PutUserMetadata: map[string]string{},
 	HTTPConfig: HTTPConfig{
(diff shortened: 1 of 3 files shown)


pkg/objstore/s3/s3.go
```diff
@@ -53,13 +56,28 @@ type Config struct {
...
+	//	encrypt_sse - An alias for sse_s3.
+	// 	sse_s3 			- When true, enables SSE-S3. ignored if KMS or C are enabled. `encrypt_sse` is kept as a legacy alias for this.
+	//.	sse_c_key		- When set to a filepath, enables SSE-C. Overrides `sse_s3`.
+	//	sse_kms_id	- When set, enables SSE-KMS. Overrides `sse_s3`. A KMS context can be optionally set via `sse_kms_context`.
+	//	sse_kms_context  - If `sse_kms_id` is set, attaches the given context to all KMS requests.
+	//
+	// NOTE: If both sse_c_key and sse_kms_id are set, we throw an error, as it's unclear what the operator wants.
+	// See https://docs.aws.amazon.com/AmazonS3/latest/dev/ServerSideEncryptionCustomerKeys.html
+	// for more details on the SSE-S3, SSE-KMS, and SSE-C options.
+	SSEEncryptionLegacy bool              `yaml:"encrypt_sse"`
```
bwplotka: I think we want something like this, but what about, since we are in struct make it more generic? The problem with this approach is that the relation if user will see ALL three of them is not really sure. 

It should be clear that you can either set `sse_s3`, `see_c` or `see_kms` if possible... (: 

Or at least let's create `SSE SSEConfig` field and put all of those in such struct. I think that would work. Also let's remove legacy field and add CHANGELOG item for removal of this. What do you think?

arshvin:
It seems like this feature should work but probably not in my environment. During attempts to upload the new block into AWS S3 store or to read `meta.json` file of already uploaded block with help by AWS CLI, I only get 403 http-error. Unfortunately I have no access to AWS Console to touch the knobs but from the other side I've tried to put files to and get files from this bucket with SSE-KMS with help by peace of code which uses the same version of the minio-go labrary (6.0.49). What's a miserable week (

PR #2172, opened 2020-02-24
Added hashmod support for store sharding.

bwplotka:


pkg/block/fetcher.go
```diff
@@ -401,14 +401,25 @@ func NewLabelShardedMetaFilter(relabelConfig []*relabel.Config) *LabelShardedMet
 	return &LabelShardedMetaFilter{relabelConfig: relabelConfig}
 }
 
+const blockIDLabel = "__block_id"
+
 // Filter filters out blocks that filters blocks that have no labels after relabelling.
```
squat: This comment is quite confusing to me. Do the filtered blocks filter out other blocks with no labels? or is there an extra `filters out blocks` here?

pkg/block/fetcher.go
```diff
@@ -401,14 +401,25 @@ func NewLabelShardedMetaFilter(relabelConfig []*relabel.Config) *LabelShardedMet
 	return &LabelShardedMetaFilter{relabelConfig: relabelConfig}
 }
 
+const blockIDLabel = "__block_id"
```
squat: Do we have a list of the various different magic label names we use? I'm worried about adding more of these in undocumented places. At one point we had something like `@thanos...` but I forget if we took it out

brancz: https://github.com/thanos-io/thanos/blob/master/docs/proposals/201909_thanos_sharding.md#relabelling

pkg/block/fetcher.go
```diff
@@ -401,14 +401,25 @@ func NewLabelShardedMetaFilter(relabelConfig []*relabel.Config) *LabelShardedMet
...
 // Filter filters out blocks that filters blocks that have no labels after relabelling.
 func (f *LabelShardedMetaFilter) Filter(metas map[ulid.ULID]*metadata.Meta, synced GaugeLabeled, _ bool) {
+	lbls := make(labels.Labels, 1)
 	for id, m := range metas {
-		if processedLabels := relabel.Process(labels.FromMap(m.Thanos.Labels), f.relabelConfig...); processedLabels != nil {
-			continue
+		lbls[0] = labels.Label{Name: blockIDLabel, Value: id.String()}
+		lbls = lbls[:1]
+		if len(lbls) < len(m.Thanos.Labels) {
+			lbls = make([]labels.Label, 0, len(m.Thanos.Labels))
```
squat: this logic is a bit funny I think. If `m.Thanos.Labels` has length `2`, then we throw away the block ID label we set earlier. But if it has length `1`, then we keep the block ID label and allocate a new backing array when we append that one single label. Is that what we want to do?

pkg/block/fetcher.go
squat: thanks :)

squat: The proposal was already `complete` but that label was previously unused in the repo. So I take it we simply reserved the labels

brancz: yes we simply missed to implement it back then

pkg/block/fetcher.go
brancz: Yeah it looks like in this case we wouldn't end up having the block id label I think. I can see what you were trying to do but is this really that performance sensitive that we need to safe the allocations here?

pkg/block/fetcher.go
bwplotka: This was part of the https://thanos.io/proposals/201909_thanos_sharding.md/ and yes we need docs! :hugs:

pkg/block/fetcher.go
brancz: Yes I think there is a double `filters out blocks`.

pkg/block/fetcher.go
bwplotka: Good point. I wanted to save allocs, will fix

bwplotka:
Addressed all, thanks for detailed review :+1:

pkg/block/fetcher.go
squat: nit, `special label` rather than `special block`

pkg/block/fetcher.go
```diff
@@ -401,14 +401,26 @@ func NewLabelShardedMetaFilter(relabelConfig []*relabel.Config) *LabelShardedMet
...
+const blockIDLabel = "__block_id"
+
+// Filter filters out blocks that have no labels after relabelling of each block external (Thanos) labels.
 func (f *LabelShardedMetaFilter) Filter(metas map[ulid.ULID]*metadata.Meta, synced GaugeLabeled, _ bool) {
+	var lbls labels.Labels
 	for id, m := range metas {
-		if processedLabels := relabel.Process(labels.FromMap(m.Thanos.Labels), f.relabelConfig...); processedLabels != nil {
-			continue
+		lbls = lbls[:0]
+		if len(lbls) < len(m.Thanos.Labels)+1 {
```
brancz: you reset the length in the line above, I assume you meant to do `cap` on the left side here?

squat: This check is now unnecessary. Here, `lbls` will ALWAYS be 0 length, and `len(m.Thanos.Labels)+1` will ALWAYS be at least 1, so the make will always be called. I think we can skip the `if` and simply always call the `lbls = make...`

bwplotka: I wanted cap :man_facepalming:

brancz: or just remove all of this and just reset lbls at the beginning of the for body, then append will take care of growing the array when needed

bwplotka: Wanted to reuse the arr. Fixed now

pkg/block/fetcher.go
bwplotka: done

bwplotka:
At some point, we need benchmarks for those filters. Maybe we would need to make this more concurrently etc.

TODO: https://github.com/thanos-io/thanos/issues/2174

brancz:
I assume you’re mostly talking about fetching the block meta. Relabeling is cheap, fast and cacheable.

bwplotka:
No, I mean filtering itself. Especially the graph-based one for `duplicated` blocks (: And yes, relabelling is maybe cacheable but we don't do that, do we?  =D Anyway, added issue with `help wanted` - would be nice to have.

brancz:
Makes sense.

PR #2176, opened 2020-02-24
mixin: Modify alert severities and thresholds

kakkoyun:
This PR modifies alert a couple of alert severities.

cc @brancz 

## Changes

* Update severities for Thanos Ruler alerts
* Update severities for Thanos Receiver alerts
* Update severities and the threshold for Thanos Querier alerts

## Verification

* `make examples`

### DIFF
--- a/examples/alerts/alerts.md
+++ b/examples/alerts/alerts.md
@@ -336,7 +336,7 @@ rules:
       }} seconds for instant queries.
   expr: |
     (
-      histogram_quantile(0.99, sum by (job, le) (http_request_duration_seconds_bucket{job=~"thanos-query.*", handler="query"})) > 10
+      histogram_quantile(0.99, sum by (job, le) (http_request_duration_seconds_bucket{job=~"thanos-query.*", handler="query"})) > 90
     and
       sum by (job) (rate(http_request_duration_seconds_bucket{job=~"thanos-query.*", handler="query"}[5m])) > 0
     )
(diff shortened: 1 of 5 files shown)


examples/alerts/alerts.md
```diff
@@ -106,7 +106,7 @@ rules:
     sum by (job) (rate(thanos_rule_evaluation_with_warnings_total{job=~"thanos-rule.*"}[5m])) > 0
   for: 15m
   labels:
-    severity: warning
+    severity: info
```
kakkoyun: Downgraded to `info` because it won't affect anything operational, Thanos ruler will continue to operate regardlessly.

examples/alerts/alerts.md
```diff
@@ -336,7 +336,7 @@ rules:
       }} seconds for instant queries.
   expr: |
     (
-      histogram_quantile(0.99, sum by (job, le) (http_request_duration_seconds_bucket{job=~"thanos-query.*", handler="query"})) > 10
+      histogram_quantile(0.99, sum by (job, le) (http_request_duration_seconds_bucket{job=~"thanos-query.*", handler="query"})) > 90
```
kakkoyun: I have increased the threshold to `90` where default query timeout is `120` seconds.

examples/alerts/alerts.md
```diff
@@ -403,7 +403,7 @@ rules:
     )
   for: 5m
   labels:
-    severity: critical
+    severity: warning
```
kakkoyun: Downgraded to `warning` because  certain amount of failures happen when rolling out receivers,

metalmatze:
I think in all other upstream mixin we usually only have warning and critical severities. Are we sure we want to introduce this?

metalmatze:
Don't get me wrong, I think it's nice to have them as info, just thinking about following most other projects.

kakkoyun:
@metalmatze I don't actually know a lot on this aspect. I have looked for canonical levels for alerts couldn't find one. Is there a place I can check for? Otherwise, I would like to introduce this level and I can also add documentation for these levels.

examples/alerts/alerts.md
bwplotka: Hm. Still if we are not rolling, we should have it critical, right? Then downgrading might be not great. If we are fine with user to have higher latency as we roll out we should have that in our alerting as well. I guess this means having this alert being critical but less strict. WDYT? (:

brancz: this should always be warning in my opinion, these are internal requests, the thing that should only ever be critical really for the receive components is the inability to write data, which we do already have :)

metalmatze: The thing is. With data-replication one can easily tolerate up to 20% of internal errors, for example, and to the outside, a user won't even notice.

bwplotka: Wait. That's not error. That's latency.

PR #2184, opened 2020-02-25
pkg/query: Add more detail to data retrieval tracing

brancz:

## Changes

Add matchers, min and max time as span tags for data retrieval of the querier.

## Verification

n/a as instrumentation

@bwplotka @squat @metalmatze @GiedriusS @povilasv

### DIFF
--- a/pkg/query/querier.go
+++ b/pkg/query/querier.go
@@ -9,6 +9,8 @@ import (
 	"strings"
 
 	"github.com/go-kit/kit/log"
+	grpc_opentracing "github.com/grpc-ecosystem/go-grpc-middleware/tracing/opentracing"
+	"github.com/opentracing/opentracing-go"
 	"github.com/pkg/errors"
 	"github.com/prometheus/prometheus/pkg/labels"
 	"github.com/prometheus/prometheus/storage"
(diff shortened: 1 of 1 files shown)


metalmatze:
This is just adding more context to the spans to follow, correct?

brancz:
correct

pkg/query/querier.go
```diff
@@ -185,6 +187,16 @@ func (q *querier) Select(params *storage.SelectParams, ms ...*labels.Matcher) (s
 	}
 	queryAggrs, resAggr := aggrsFromFunc(params.Func)
 
+	matchers := make([]string, len(ms))
```
bwplotka: I think this can work for now, but what about doing it properly on GRPC StoreAPI level? Potentially as interceptor even? :thinking:

bwplotka: I wonder if there is some existing interceptor for it TBH..

brancz: Yeah I was thinking about this. For now I want it to concentrate on the most essential things we need. I'll keep thinking about this though.

bwplotka: There is none: Added issue: https://github.com/grpc-ecosystem/go-grpc-middleware/issues/273

PR #2185, opened 2020-02-25
Skip deleting files that we just deleted

hsmade:
We see this happening with Swift. Because the consistency of swift is eventual, swift sometimes didn't process the deletion of the meta file yet, and so it turns up in the bkt.Iter(). The second deletion then causes a 404 and compaction fails.

### DIFF
--- a/pkg/block/block.go
+++ b/pkg/block/block.go
@@ -147,10 +147,7 @@ func Delete(ctx context.Context, logger log.Logger, bkt objstore.Bucket, id ulid
 
 	// Delete the bucket, but skip the metaFile, if found. As we just deleted that.
 	return deleteDirRec(ctx, logger, bkt, id.String(), func(name string) bool {
-		if name == metaFile {
-			return true
-		}
-		return false
+		return name == metaFile
 	})
 }
 
--- a/pkg/block/block.go
+++ b/pkg/block/block.go
@@ -156,6 +156,11 @@ func deleteDir(ctx context.Context, logger log.Logger, bkt objstore.Bucket, dir 
 		if strings.HasSuffix(name, objstore.DirDelim) {
 			return deleteDir(ctx, logger, bkt, name)
 		}
+		metaFile := path.Join(id.String(), MetaFilename)
+		if name == metaFile {
+			// skip this file if we found it, it was already deleted
+			continue
+		}
 		if err := bkt.Delete(ctx, name); err != nil {
 			return err
 		}
--- a/pkg/block/block.go
+++ b/pkg/block/block.go
@@ -156,7 +156,7 @@ func deleteDir(ctx context.Context, logger log.Logger, bkt objstore.Bucket, dir 
 		if strings.HasSuffix(name, objstore.DirDelim) {
 			return deleteDir(ctx, logger, bkt, name)
 		}
-		metaFile := path.Join(id.String(), MetaFilename)
+		metaFile := path.Join(dir, MetaFilename)
 		if name == metaFile { // the metaFile was already deleted in Delete(), but might still appear in the Iter() list
 			level.Debug(logger).Log("msg", "skipping deletion of meta file, as it should already be deleted", "file", name, "bucket", bkt.Name())
 			return nil
--- a/pkg/block/block.go
+++ b/pkg/block/block.go
@@ -157,7 +157,7 @@ func deleteDir(ctx context.Context, logger log.Logger, bkt objstore.Bucket, dir 
 			return deleteDir(ctx, logger, bkt, name)
 		}
 		metaFile := path.Join(dir, MetaFilename)
-		if name == metaFile { // the metaFile was already deleted in Delete(), but might still appear in the Iter() list
+		if name == metaFile { // the metaFile was already deleted in Delete(), but might still appear in the Iter() list.
 			level.Debug(logger).Log("msg", "skipping deletion of meta file, as it should already be deleted", "file", name, "bucket", bkt.Name())
 			return nil
 		}
--- a/pkg/block/block.go
+++ b/pkg/block/block.go
@@ -145,13 +145,13 @@ func Delete(ctx context.Context, logger log.Logger, bkt objstore.Bucket, id ulid
 		level.Debug(logger).Log("msg", "deleted file", "file", metaFile, "bucket", bkt.Name())
 	}
 
-	// Delete the bucket, but skip the metaFile, if found. As we just deleted that.
+	// Delete the bucket, but skip the metaFile as we just deleted that. This is required for eventual object storages (list after write).
 	return deleteDirRec(ctx, logger, bkt, id.String(), func(name string) bool {
 		return name == metaFile
 	})
 }
 
-// deleteDirRec removes all objects prefixed with dir from the bucket. It skips objects that return true for the passed keep func
+// deleteDirRec removes all objects prefixed with dir from the bucket. It skips objects that return true for the passed keep function.
 // NOTE: For objects removal use `block.Delete` strictly.
 func deleteDirRec(ctx context.Context, logger log.Logger, bkt objstore.Bucket, dir string, keep func(name string) bool) error {
 	return bkt.Iter(ctx, dir, func(name string) error {
@@ -160,7 +160,6 @@ func deleteDirRec(ctx context.Context, logger log.Logger, bkt objstore.Bucket, d
 			return deleteDirRec(ctx, logger, bkt, name, keep)
 		}
 		if keep(name) {
-			level.Debug(logger).Log("msg", "skipping deletion of object, as requested by keep()", "file", name, "bucket", bkt.Name())
 			return nil
 		}
 		if err := bkt.Delete(ctx, name); err != nil {
--- a/pkg/block/block.go
+++ b/pkg/block/block.go
@@ -145,20 +145,25 @@ func Delete(ctx context.Context, logger log.Logger, bkt objstore.Bucket, id ulid
 		level.Debug(logger).Log("msg", "deleted file", "file", metaFile, "bucket", bkt.Name())
 	}
 
-	return deleteDir(ctx, logger, bkt, id.String())
+	// Delete the bucket, but skip the metaFile, if found. As we just deleted that.
+	return deleteDirRec(ctx, logger, bkt, id.String(), func(name string) bool {
+		if name == metaFile {
+			return true
+		}
+		return false
+	})
 }
 
-// deleteDir removes all objects prefixed with dir from the bucket.
+// deleteDirRec removes all objects prefixed with dir from the bucket. It skips objects that return true for the passed keep func
 // NOTE: For objects removal use `block.Delete` strictly.
-func deleteDir(ctx context.Context, logger log.Logger, bkt objstore.Bucket, dir string) error {
+func deleteDirRec(ctx context.Context, logger log.Logger, bkt objstore.Bucket, dir string, keep func(name string) bool) error {
 	return bkt.Iter(ctx, dir, func(name string) error {
 		// If we hit a directory, call DeleteDir recursively.
 		if strings.HasSuffix(name, objstore.DirDelim) {
-			return deleteDir(ctx, logger, bkt, name)
+			return deleteDirRec(ctx, logger, bkt, name, keep)
 		}
-		metaFile := path.Join(dir, MetaFilename)
-		if name == metaFile { // the metaFile was already deleted in Delete(), but might still appear in the Iter() list.
-			level.Debug(logger).Log("msg", "skipping deletion of meta file, as it should already be deleted", "file", name, "bucket", bkt.Name())
+		if keep(name) {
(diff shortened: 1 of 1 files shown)


hsmade:
`{
  "caller": "fetcher.go:189",
  "component": "block.MetaFetcher",
  "level": "debug",
  "msg": "download meta",
  "name": "01E065XF3TXB1VQ765VJS9QTBT/meta.json",
  "ts": "2020-02-25T13:38:12.902114856Z"
}
{
  "blocks": "[data/compact/0@16967372497332894226/01E065XF3TXB1VQ765VJS9QTBT data/compact/0@16967372497332894226/01E067MCXV4G386C5R7SZF9KWM data/compact/0@16967372497332894226/01E069BAQZWB9DYZRDR20RJS6G data/compact/0@16967372497332894226/01E06B28HY6X0FD6M57BFMQHDY]",
  "caller": "compact.go:733",
  "compactionGroup": "0@16967372497332894226",
  "duration": "54.278347188s",
  "level": "debug",
  "msg": "downloaded and verified blocks",
  "ts": "2020-02-25T13:49:02.610198004Z"
}
{
  "caller": "compact.go:441",
  "count": 4,
  "duration": "1m1.290166319s",
  "level": "info",
  "maxt": 1580760000000,
  "mint": 1580752800000,
  "msg": "compact blocks",
  "sources": "[01E065XF3TXB1VQ765VJS9QTBT 01E067MCXV4G386C5R7SZF9KWM 01E069BAQZWB9DYZRDR20RJS6G 01E06B28HY6X0FD6M57BFMQHDY]",
  "ts": "2020-02-25T13:50:03.900468827Z",
  "ulid": "01E1Y9QBSJ3NHQMSRAJ6FQXPX1"
}
{
  "blocks": "[data/compact/0@16967372497332894226/01E065XF3TXB1VQ765VJS9QTBT data/compact/0@16967372497332894226/01E067MCXV4G386C5R7SZF9KWM data/compact/0@16967372497332894226/01E069BAQZWB9DYZRDR20RJS6G data/compact/0@16967372497332894226/01E06B28HY6X0FD6M57BFMQHDY]",
  "caller": "compact.go:764",
  "compactionGroup": "0@16967372497332894226",
  "duration": "1m1.33904902s",
  "level": "debug",
  "msg": "compacted blocks",
  "overlapping_blocks": false,
  "ts": "2020-02-25T13:50:03.94934599Z"
}
{
  "caller": "compact.go:834",
  "compactionGroup": "0@16967372497332894226",
  "level": "info",
  "msg": "deleting compacted block",
  "old_block": "01E065XF3TXB1VQ765VJS9QTBT",
  "ts": "2020-02-25T13:50:32.309739616Z"
}
{
  "bucket": "thanos",
  "caller": "block.go:140",
  "compactionGroup": "0@16967372497332894226",
  "file": "01E065XF3TXB1VQ765VJS9QTBT/meta.json",
  "level": "debug",
  "msg": "deleted file",
  "ts": "2020-02-25T13:50:32.907953626Z"
}
{
  "bucket": "thanos",
  "caller": "block.go:157",
  "compactionGroup": "0@16967372497332894226",
  "file": "01E065XF3TXB1VQ765VJS9QTBT/chunks/000001",
  "level": "debug",
  "msg": "deleted file",
  "ts": "2020-02-25T13:50:33.533423488Z"
}
{
  "bucket": "thanos",
  "caller": "block.go:157",
  "compactionGroup": "0@16967372497332894226",
  "file": "01E065XF3TXB1VQ765VJS9QTBT/index",
  "level": "debug",
  "msg": "deleted file",
  "ts": "2020-02-25T13:50:33.655638721Z"
}
{
  "caller": "prober.go:117",
  "level": "warn",
  "msg": "changing probe status",
  "reason": "error executing compaction: compaction failed: compaction failed for group 0@16967372497332894226: delete old block from bucket: delete block 01E065XF3TXB1VQ765VJS9QTBT from bucket: Resource not found",
  "status": "not-ready",
  "ts": "2020-02-25T13:50:34.386430986Z"
}
{
  "caller": "http.go:78",
  "component": "compact",
  "err": "error executing compaction: compaction failed: compaction failed for group 0@16967372497332894226: delete old block from bucket: delete block 01E065XF3TXB1VQ765VJS9QTBT from bucket: Resource not found",
  "level": "info",
  "msg": "internal server shutdown",
  "service": "http/server",
  "ts": "2020-02-25T13:50:34.38667029Z"
}
{
  "caller": "prober.go:137",
  "level": "info",
  "msg": "changing probe status",
  "reason": "error executing compaction: compaction failed: compaction failed for group 0@16967372497332894226: delete old block from bucket: delete block 01E065XF3TXB1VQ765VJS9QTBT from bucket: Resource not found",
  "status": "not-healthy",
  "ts": "2020-02-25T13:50:34.386719215Z"
}
{
  "caller": "main.go:194",
  "err": "error executing compaction: compaction failed: compaction failed for group 0@16967372497332894226: delete old block from bucket: delete block 01E065XF3TXB1VQ765VJS9QTBT from bucket: Resource not found",
  "level": "error",
  "msg": "running command failed",
  "ts": "2020-02-25T13:50:34.386763391Z"
}`

bwplotka:
Wow, this eventual consistency is really bad... (:

But good to know!

 @khyatisoneji do you mind taking a look at this PR, how we can fix this? You have quite good experience with working with eventual consistency systems so far :heart: ! 

Looks like yet another reason to finish this: https://thanos.io/proposals/201901-read-write-operations-bucket.md/

bwplotka:
However still I can't see the problem. Even with inconsistency Iter should not repeat same file :thinking: Is it really the case now?

hsmade:
The swift logs tell us so, I'll ask our cloud team for some logs to add to this PR.

leeuwenrjj:
> However still I can't see the problem. 
> Even with inconsistency Iter should not repeat same file 🤔 Is it really the case now?

The problem is the code first starts with deleting metadata.json outside of the iter loop in "Delete()"
Next it lists all items in the bucket (where also the metadata.json is located) and does the iter loop over all files it got from the list.
So if the delete is not fully processed on all nodes with the bucket index you might end up with a list that still contains the metadata.json. Now the iter items will also try to delete it but the original delete actually is processed and the object store will throw an error.

Note that Amazon S3 will have the exact same issue, from the amazon docs [here](https://docs.aws.amazon.com/AmazonS3/latest/dev/Introduction.html) :
* A process deletes an existing object and immediately lists keys within its bucket. Until the deletion is fully propagated, Amazon S3 might list the deleted object.

hsmade:
(stripped) logs from swift:

```
{
  "_index": "logstash-2020.02.25",
  "_type": "doc",
  "_source": {
    "timestamp": "2020-02-25T12:21:33.408687+00:00",
    "data": {
      "_SYSTEMD_UNIT": "openstack-swift-proxy.service",
      "msg": "xx.xx.xx.xx yy.yy.yy.yy 25/Feb/2020/12/21/33 DELETE /v1/xxxxxxxxxxxxx/thanos/01E062FKFT3YMZFGCQGZSTYKVF/meta.json HTTP/1.0 204 - gophercloud/2.0.0",
      "_SOURCE_REALTIME_TIMESTAMP": "1582633293407785"
    },
    "@timestamp": "2020-02-25T12:21:33.409Z",
  }
}

{
  "_index": "logstash-2020.02.25",
  "_type": "doc",
  "_source": {
    "timestamp": "2020-02-25T12:21:34.111697+00:00",
    "data": {
      "_SYSTEMD_UNIT": "openstack-swift-proxy.service",
      "msg": "xx.xx.xx.xx yy.yy.yy.yy 25/Feb/2020/12/21/34 DELETE /v1/xxxxxxxxxxxxx/thanos/01E062FKFT3YMZFGCQGZSTYKVF/meta.json HTTP/1.0 404 - gophercloud/2.0.0",
      "_SOURCE_REALTIME_TIMESTAMP": "1582633294111419",
    },
    "@timestamp": "2020-02-25T12:21:34.112Z",
  }
}
```

khyatisoneji:
I think having this extra check would be good since we are deleting the `meta.json` first and then iterating on the files in the folder. There can be cases when the file delete call succeeds, but a GET call still returns the file.

bwplotka:
I mean... to me it looks like we want to actually use Delayed Delete always? @khyatisoneji

khyatisoneji:
Just for clarification, by delayed delete do you mean that we don't straightaway delete the block right?

This case is about `block.Delete` function which first deletes `meta.json` file inside the block and then iterates over all files in the block. This scenario occurs when even after deletion of `meta.json` file, we get `meta.json` file in the `bkt.Iter` function call.

So this particular concern is not about deleting the blocks after some delay. Even if we delete the block after some delay, this scenario can happen

hsmade:
Just tested this PR in our integration env, and where we had errors almost every hour, they're now gone.

pkg/block/block.go
```diff
@@ -156,6 +156,11 @@ func deleteDir(ctx context.Context, logger log.Logger, bkt objstore.Bucket, dir
 		if strings.HasSuffix(name, objstore.DirDelim) {
 			return deleteDir(ctx, logger, bkt, name)
 		}
+		metaFile := path.Join(dir, MetaFilename)
```
bwplotka: OK while I see the problem, but do you this exact fix here really makes sense?

`deleteDir` is responsible for deleting a "directory" in object storage, right? So this is what the user expects. Now we suddenly ignore meta.json file some reason. Do you think this is safe? (: I think this will be extremely surprising and might introduce bug someday when someone will try to reuse this function for something else than now... And we love to avoid surprises in Thanos project! (: WDYT? You should never assume usage without explicitly mentioning it in name/comment. ):

In this case, what would work is to pass to `deleteDir` some `keep func(name string)bool` argument, so users can decide on their own. Also can we rename `deleteDir` to `deleteDirRec` to be more explicit :hugs:

hsmade: Having a whole function to decide on what file might stay or not, might overcomplicate things though. What about a slice of strings? 
Also, what crossed my mind: what about ignoring the 404 in swift? The contract of `objstore.swift.Delete` is that it should remove the object, the 404 tells you it's not there, so the contract is fulfilled.

bwplotka: Why overcomplicate? It will have fewer lines then what you added now (: 

```
if keep(name) {
   return nil 
}
```

bwplotka: > what about ignoring the 404 in swift?

Ignornig on what level?

bwplotka: We like 404 to be an error, as you remove something which you expect to be there. We can consider and discuss it, but for here it will super easy to add `keep` function (:

pkg/block/block.go
```diff
@@ -145,16 +145,23 @@ func Delete(ctx context.Context, logger log.Logger, bkt objstore.Bucket, id ulid
...
-func deleteDir(ctx context.Context, logger log.Logger, bkt objstore.Bucket, dir string) error {
+func deleteDirRec(ctx context.Context, logger log.Logger, bkt objstore.Bucket, dir string, keep func(name string) bool) error {
 	return bkt.Iter(ctx, dir, func(name string) error {
 		// If we hit a directory, call DeleteDir recursively.
 		if strings.HasSuffix(name, objstore.DirDelim) {
-			return deleteDir(ctx, logger, bkt, name)
+			return deleteDirRec(ctx, logger, bkt, name, keep)
+		}
+		if keep(name) {
+			level.Debug(logger).Log("msg", "skipping deletion of object, as requested by keep()", "file", name, "bucket", bkt.Name())
```
bwplotka: This works as intended so I would prefer no log line even.

pkg/block/block.go
```diff
@@ -145,16 +145,23 @@ func Delete(ctx context.Context, logger log.Logger, bkt objstore.Bucket, id ulid
...
 
-	return deleteDir(ctx, logger, bkt, id.String())
+	// Delete the bucket, but skip the metaFile, if found. As we just deleted that.
+	return deleteDirRec(ctx, logger, bkt, id.String(), func(name string) bool {
+		return name == metaFile
+	})
 }
 
-// deleteDir removes all objects prefixed with dir from the bucket.
+// deleteDirRec removes all objects prefixed with dir from the bucket. It skips objects that return true for the passed keep func
```
bwplotka:
```suggestion
// deleteDirRec removes all objects prefixed with dir from the bucket. It skips objects that return true for the passed keep function.
```

pkg/block/block.go
```diff
@@ -145,16 +145,23 @@ func Delete(ctx context.Context, logger log.Logger, bkt objstore.Bucket, id ulid
 		level.Debug(logger).Log("msg", "deleted file", "file", metaFile, "bucket", bkt.Name())
 	}
 
-	return deleteDir(ctx, logger, bkt, id.String())
+	// Delete the bucket, but skip the metaFile, if found. As we just deleted that.
```
bwplotka:
```suggestion
	// Delete the bucket, but skip the metaFile as we just deleted that. This is required for eventual object storages (list after write).
```

PR #2193, opened 2020-02-26
Simplify and clarify small bits in the release process

metalmatze:
Just a few small things I encountered while cutting the RCs (:


metalmatze:
Formatting in Markdown is hard. But after 3 force pushes I finally got the correct result on Netlify.
Also :heart: Netlify's preview. Amazing!

PR #2201, opened 2020-02-28
Added Type to rules requests as it is on HTTP API.

bwplotka:

cc @s-urbaniak


s-urbaniak:
good catch :+1: 
lgtm

PR #2202, opened 2020-02-28
tracing: Simplified creation of spans.

bwplotka:
cc @s-urbaniak (: 

### DIFF
--- a/pkg/store/bucket.go
+++ b/pkg/store/bucket.go
@@ -825,14 +825,13 @@ func debugFoundBlockSetOverview(logger log.Logger, mint, maxt, maxResolutionMill
 
 // Series implements the storepb.StoreServer interface.
 func (s *BucketStore) Series(req *storepb.SeriesRequest, srv storepb.Store_SeriesServer) (err error) {
-	{
-		span, _ := tracing.StartSpan(srv.Context(), "store_query_gate_ismyturn")
-		err := s.queryGate.IsMyTurn(srv.Context())
-		span.Finish()
-		if err != nil {
-			return errors.Wrapf(err, "failed to wait for turn")
-		}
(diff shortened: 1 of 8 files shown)


pkg/store/bucket.go
```diff
@@ -962,16 +958,18 @@ func (s *BucketStore) Series(req *storepb.SeriesRequest, srv storepb.Store_Serie
 			}
 
 			if err := srv.Send(storepb.NewSeriesResponse(&series)); err != nil {
-				return status.Error(codes.Unknown, errors.Wrap(err, "send series response").Error())
+				err = status.Error(codes.Unknown, errors.Wrap(err, "send series response").Error())
```
s-urbaniak: i believe this still needs a `return`, no?

s-urbaniak: scratch that, you need a capturing `var err error` before the `if` block :)

pkg/store/bucket.go
```diff
@@ -962,16 +958,18 @@ func (s *BucketStore) Series(req *storepb.SeriesRequest, srv storepb.Store_Serie
 			}
 
 			if err := srv.Send(storepb.NewSeriesResponse(&series)); err != nil {
```
s-urbaniak: sorry, i was blind, i guess this just needs `if err = srv.Send...`

s-urbaniak: otherwise the shadowed `err` will be an ineffectual assignment as CI complained :-)

s-urbaniak:
We just need to fix two shadowed `err`s, else lgtm! :tada:

pkg/store/bucket.go
bwplotka: Wow! Thanks `go vet` :+1:

PR #2217, opened 2020-03-05
[WIP] compact: Add vertical-deduplication for replica-labels as MetaFetcher filter

metalmatze:

_Need to create the PR first to get a number :grin:_

## Changes

Add MetaFetcher filter called ReplicaLabelsFilter in: https://github.com/thanos-io/thanos/pull/2217/commits/c6828f9fd9fc1867f6dd4caa3c9fcf01d2e071d3
Other commits are to add e2e tests.

## Verification

This is the reason why it's still WIP. I want to finish the end to end tests, but I get minio errors. I'm not able to fix it right now.
```
compact_test.go:60: unexpected error: upload file /tmp/e2e_integration_test767365421/tmp/01E2NBJ5Q70000000000000000/chunks/000001 as 01E2NBJ5Q70000000000000000/chunks/000001: upload s3 object: Get "http://e2e_test_compact-minio-80/thanos/?location=": dial tcp: lookup e2e_test_compact-minio-80: no such host
```

/cc @brancz @bwplotka @squat

### DIFF
--- a/cmd/thanos/compact.go
+++ b/cmd/thanos/compact.go
@@ -121,6 +121,9 @@ func registerCompact(m map[string]setupFunc, app *kingpin.Application) {
 	compactionConcurrency := cmd.Flag("compact.concurrency", "Number of goroutines to use when compacting groups.").
 		Default("1").Int()
 
+	dedupReplicaLabels := cmd.Flag("offline-deduplication.replica-labels", "Label to treat as a replica indicator of blocks that can be deduplicated. This will merge multiple replica blocks into one. This process is irrevertable. Experminteal").
+		Hidden().String()
+
 	selectorRelabelConf := regSelectorRelabelFlags(cmd)
 
 	m[component.Compact.String()] = func(g *run.Group, logger log.Logger, reg *prometheus.Registry, tracer opentracing.Tracer, _ <-chan struct{}, _ bool) error {
@@ -144,6 +147,7 @@ func registerCompact(m map[string]setupFunc, app *kingpin.Application) {
 			*maxCompactionLevel,
 			*blockSyncConcurrency,
 			*compactionConcurrency,
+			*dedupReplicaLabels,
 			selectorRelabelConf,
 		)
 	}
@@ -168,6 +172,7 @@ func runCompact(
 	maxCompactionLevel int,
 	blockSyncConcurrency int,
 	concurrency int,
+	dedupReplicaLabels string,
 	selectorRelabelConf *extflag.PathOrContent,
 ) error {
 	halted := prometheus.NewGauge(prometheus.GaugeOpts{
@@ -242,10 +247,13 @@ func runCompact(
 
 	duplicateBlocksFilter := block.NewDeduplicateFilter()
 	prometheusRegisterer := extprom.WrapRegistererWithPrefix("thanos_", reg)
+	replicaLabelFilter := block.ReplicaLabelsFilter{ReplicaLabels: strings.Split(dedupReplicaLabels, ",")}
+
 	metaFetcher, err := block.NewMetaFetcher(logger, 32, bkt, "", prometheusRegisterer,
 		block.NewLabelShardedMetaFilter(relabelConfig).Filter,
 		block.NewConsistencyDelayMetaFilter(logger, consistencyDelay, prometheusRegisterer).Filter,
 		duplicateBlocksFilter.Filter,
+		replicaLabelFilter.Filter,
 	)
 	if err != nil {
 		return errors.Wrap(err, "create meta fetcher")
--- a/pkg/block/fetcher.go
+++ b/pkg/block/fetcher.go
@@ -52,6 +52,7 @@ const (
 	timeExcludedMeta  = "time-excluded"
 	tooFreshMeta      = "too-fresh"
 	duplicateMeta     = "duplicate"
+	replicaExclude    = "replica-exclude"
 )
 
 func newSyncMetrics(r prometheus.Registerer) *syncMetrics {
@@ -531,6 +532,23 @@ func contains(s1 []ulid.ULID, s2 []ulid.ULID) bool {
 		}
 	}
 	return true
+}
+
+type ReplicaLabelsFilter struct {
+	ReplicaLabels []string
+}
+
+func (f *ReplicaLabelsFilter) Filter(metas map[ulid.ULID]*metadata.Meta, synced GaugeLabeled, view bool) {
+	for u, meta := range metas {
+		labels := meta.Thanos.Labels
+		for _, replicaLabel := range f.ReplicaLabels {
+			if _, exists := labels[replicaLabel]; exists {
+				delete(labels, replicaLabel)
+				synced.WithLabelValues(replicaExclude).Inc()
+			}
+		}
+		metas[u].Thanos.Labels = labels
+	}
 }
 
 // ConsistencyDelayMetaFilter is a MetaFetcher filter that filters out blocks that are created before a specified consistency delay.
--- a/pkg/block/fetcher_test.go
+++ b/pkg/block/fetcher_test.go
@@ -847,6 +847,29 @@ func TestDeduplicateFilter_Filter(t *testing.T) {
 	}
 }
 
+func TestReplicaLabel_Filter(t *testing.T) {
+	rf := ReplicaLabelsFilter{ReplicaLabels: []string{"replica", "rule_replica"}}
+
+	input := map[ulid.ULID]*metadata.Meta{
+		ULID(1): {Thanos: metadata.Thanos{Labels: map[string]string{"message": "something"}}},
+		ULID(2): {Thanos: metadata.Thanos{Labels: map[string]string{"replica": "cluster1", "message": "something"}}},
+		ULID(3): {Thanos: metadata.Thanos{Labels: map[string]string{"replica": "cluster1", "rule_replica": "rule1", "message": "something"}}},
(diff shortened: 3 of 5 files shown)


cmd/thanos/compact.go
```diff
@@ -121,6 +121,9 @@ func registerCompact(m map[string]setupFunc, app *kingpin.Application) {
 	compactionConcurrency := cmd.Flag("compact.concurrency", "Number of goroutines to use when compacting groups.").
 		Default("1").Int()
 
+	dedupReplicaLabels := cmd.Flag("offline-deduplication.replica-labels", "Label to treat as a replica indicator of blocks that can be deduplicated. This will merge multiple replica blocks into one. This process is irrevertable. Experminteal").
```
bwplotka:
```suggestion
	dedupReplicaLabels := cmd.Flag("deduplication.replica-label", "Label to treat as a replica indicator of blocks that can be deduplicated. This will merge multiple replica blocks into one (repeated flag). This process is irreversible. Experimental").
```

cmd/thanos/compact.go
```diff
@@ -121,6 +121,9 @@ func registerCompact(m map[string]setupFunc, app *kingpin.Application) {
...
+		Hidden().String()
```
bwplotka:
```suggestion
		Hidden().Strings()
```

cmd/thanos/compact.go
```diff
@@ -242,10 +247,13 @@ func runCompact(
 
 	duplicateBlocksFilter := block.NewDeduplicateFilter()
 	prometheusRegisterer := extprom.WrapRegistererWithPrefix("thanos_", reg)
+	replicaLabelFilter := block.ReplicaLabelsFilter{ReplicaLabels: strings.Split(dedupReplicaLabels, ",")}
```
bwplotka:
```suggestion
	replicaLabelFilter := block.ReplicaLabelFilter{ReplicaLabels: dedupReplicaLabels}
```

bwplotka: Can we inline this?

pkg/block/fetcher.go
```diff
@@ -533,6 +534,23 @@ func contains(s1 []ulid.ULID, s2 []ulid.ULID) bool {
 	return true
 }
 
+type ReplicaLabelsFilter struct {
```
bwplotka:
```suggestion
type ReplicaLabelConstructingFakeMetaFilter struct {
```
;p

pkg/block/fetcher.go
```diff
@@ -533,6 +534,23 @@ func contains(s1 []ulid.ULID, s2 []ulid.ULID) bool {
...
+	ReplicaLabels []string
```
bwplotka: It's configuration, so maybe make it priv? + constructor?

pkg/block/fetcher.go
```diff
@@ -533,6 +534,23 @@ func contains(s1 []ulid.ULID, s2 []ulid.ULID) bool {
...
+}
+
+func (f *ReplicaLabelsFilter) Filter(metas map[ulid.ULID]*metadata.Meta, synced GaugeLabeled, view bool) {
+	for u, meta := range metas {
+		labels := meta.Thanos.Labels
+		for _, replicaLabel := range f.ReplicaLabels {
+			if _, exists := labels[replicaLabel]; exists {
+				delete(labels, replicaLabel)
```
bwplotka: We need comment that we modify external lable of existing block, constructing fake meta.json. This might be surprising so we need to be extra careful

pkg/block/fetcher.go
bwplotka: Actually as @kakkoyun found it's not a filter, so maybe different name

pkg/block/fetcher.go
```diff
@@ -533,6 +534,23 @@ func contains(s1 []ulid.ULID, s2 []ulid.ULID) bool {
...
+				synced.WithLabelValues(replicaExclude).Inc()
```
bwplotka: This is wrong then I think? At least not consistent with other filters.

Usually we add some label like this and increment when we filter block.

bwplotka: Anything that is not filtered, will be incremented by len(not filtered) {loaded}

bwplotka: Maybe a separate metrics here?

pkg/block/fetcher.go
bwplotka: "Modifier"?

pkg/block/fetcher.go
```diff
@@ -533,6 +534,23 @@ func contains(s1 []ulid.ULID, s2 []ulid.ULID) bool {
 	return true
 }
 
+type ReplicaLabelsFilter struct {
+	ReplicaLabels []string
+}
+
+func (f *ReplicaLabelsFilter) Filter(metas map[ulid.ULID]*metadata.Meta, synced GaugeLabeled, view bool) {
```
bwplotka:
```suggestion
func (f *ReplicaLabelsFilter) PostFetch<SOMETHING>(metas map[ulid.ULID]*metadata.Meta, synced GaugeLabeled, view bool) {
```

pkg/block/fetcher_test.go
```diff
@@ -847,6 +847,29 @@ func TestDeduplicateFilter_Filter(t *testing.T) {
 	}
 }
 
+func TestReplicaLabel_Filter(t *testing.T) {
+	rf := ReplicaLabelsFilter{ReplicaLabels: []string{"replica", "rule_replica"}}
```
bwplotka: Can we have a test with no replicalabels?

cmd/thanos/compact.go
```diff
@@ -242,10 +247,13 @@ func runCompact(
 
 	duplicateBlocksFilter := block.NewDeduplicateFilter()
```
bwplotka: Somewhere we need to enable vertical compaction flag IF replica flag is enabled on LeveledCompactor

test/e2e/compact_test.go
```diff
@@ -0,0 +1,69 @@
...
+	series := []labels.Labels{labels.FromStrings("a", "1", "b", "2")}
+	extLset := labels.FromStrings("ext1", "value1", "replica", "1")
+
+	bkt, err := s3.NewBucketWithConfig(l, s3Config, "test-feed")
+	testutil.Ok(t, err)
+
+	dir := filepath.Join(s.SharedDir(), "tmp")
+
+	now := time.Now()
+	id1, err := e2eutil.CreateBlockWithBlockDelay(ctx, dir, series, 10, timestamp.FromTime(now), timestamp.FromTime(now.Add(2*time.Hour)), 30*time.Minute, extLset, 0)
```
bwplotka: We don''t need or want delay I think

test/e2e/compact_test.go
```diff
@@ -0,0 +1,69 @@
...
+
+	s3Config := s3.Config{
+		Bucket:    bucket,
+		AccessKey: e2edb.MinioAccessKey,
+		SecretKey: e2edb.MinioSecretKey,
+		Endpoint:  m.NetworkHTTPEndpoint(),
+		Insecure:  true,
+	}
+
+	series := []labels.Labels{labels.FromStrings("a", "1", "b", "2")}
```
bwplotka: Let's maybe put block creation to separate function, because we need to create tons of them for the tests

bwplotka: Potentiall would be nice to have a table test for different tests
* Offline Dedup
* Vertical dedup (let's focus in this)
* Downs
* Compacti
* retention 

etc

cmd/thanos/compact.go
bwplotka: Let's make sure it's documented that it's ONLY vertical for now and what it means.

cmd/thanos/compact.go
bwplotka: Let's add comment to `Hidden` why it's hidden

test/e2e/compact_test.go
```diff
@@ -0,0 +1,69 @@
...
+
+	testutil.Ok(t, objstore.UploadDir(ctx, l, bkt, path.Join(dir, id1.String()), id1.String()))
+
+	compact, err := e2ethanos.NewCompact(s.SharedDir(), "compact", client.BucketConfig{
+		Type:   client.S3,
+		Config: s3Config,
+	})
+	testutil.Ok(t, err)
+	testutil.Ok(t, s.StartAndWaitReady(compact))
+
```
bwplotka: On each run of table test, we need to create some blocks, then assert something. 

Open question: In table test: Should we clean up the bucket or clean up +  restart compactor.

test/e2e/e2ethanos/services.go
```diff
@@ -313,3 +313,32 @@ func NewStoreGW(sharedDir string, name string, bucketConfig client.BucketConfig,
...
+	if err := os.MkdirAll(dir, 0777); err != nil {
+		return nil, errors.Wrap(err, "create compact dir")
+	}
+
+	bktConfigBytes, err := yaml.Marshal(bucketConfig)
+	if err != nil {
+		return nil, errors.Wrapf(err, "generate compact config file: %v", bucketConfig)
+	}
+
+	fmt.Println(string(bktConfigBytes))
```
bwplotka: Kill this

pkg/block/fetcher.go
kakkoyun: When that I've checked it again I realize all other `MetaFetcherFilter`s do similar modifications. To be consistent we can keep the name.

kakkoyun:
Picking up from where @metalmatze left off in https://github.com/thanos-io/thanos/pull/2250.

PR #2219, opened 2020-03-05
bucket: improve shard label handling

MacroPower:

## Changes

* Allow bucket web UI to continue labeling old blocks, even if they don't match up with current labeling conventions.

This allows someone to edit their external labels, and still see all their data inside the bucket web component. Currently, the bucket web component will just throw an error (in bucket.js) since some of the data is unlabeled (assuming that you pass `--label`).

The bucket component already handles shards with and without labels, this change simply allows you to mix the two. Any unlabeled shards will be assigned a number, exactly how the component behaves when no `--label` is passed. After the unlabeled shards are listed, any labeled shards will be listed.

## Verification

Tested with bucket containing some data with no external labels.
image

It is merged with newer data containing external labels.

### DIFF
--- a/pkg/ui/static/js/bucket.js
+++ b/pkg/ui/static/js/bucket.js
@@ -33,16 +33,13 @@ function draw() {
                 // Title is the first column of the timeline.
                 //
                 // A unique Prometheus label that identifies each shard is used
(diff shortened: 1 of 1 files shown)


bwplotka:
Sorry for lag in review ):

PR #1982, opened 2020-01-10
*: support Alertmanager API v2

simonpasquier:
* [X] I added CHANGELOG entry for this change.
* [ ] Change is not relevant to the end user.

## Changes

Added a new parameter to the alerting configuration similar to what we have in the Prometheus configuration.

## Verification

End-to-end tests modified to verify both v1 and v2 endpoints.

PR #2226, opened 2020-03-06
query: support store config file

eswdd:
Closes https://github.com/thanos-io/thanos/issues/977

## Changes

This change adds support for reading stores from an external configuration file, with TLS and file SD configuration variable between sets of stores. The configuration file allows specification of one or more configurations in a similar manner to #1939 and #1982 . Have decided to mandate naming the configurations so that we can label metrics with these names since the primary use case driving this seems to be TLS vs non-TLS and the feeling around our team is that we'd like to see metrics split along these lines.

It is backwards compatible with existing flags for specifying stores, TLS and file SD.

## Verification

e2e tests have been updated to use the external store config file.

### DIFF
--- a/cmd/thanos/query.go
+++ b/cmd/thanos/query.go
@@ -230,7 +230,7 @@ func runQuery(
 		var err error
 		storesConfig, err = store.LoadConfig(storeConfigYAML)
 		if err != nil {
-			return err
+			return errors.Wrap(err, "loading store config")
 		}
 	} else {
 		for _, addr := range storeAddrs {
--- a/cmd/thanos/query.go
+++ b/cmd/thanos/query.go
@@ -293,7 +293,6 @@ func runQuery(
 		fileSDCache := cache.New()
 		dnsProvider := dns.NewProvider(
 			logger,
-			// todo: sml: looks a bit iffy
 			extprom.WrapRegistererWith(
 				map[string]string{"config_name": config.Name},
 				extprom.WrapRegistererWithPrefix("thanos_querier_store_apis_", reg),
--- a/cmd/thanos/query.go
+++ b/cmd/thanos/query.go
@@ -361,8 +361,8 @@ func runQuery(
 				Timeout:    queryTimeout,
 			},
 		)
-		grpcProbe = prober.NewGRPC()
-		httpProbe = prober.NewHTTP()
+		grpcProbe    = prober.NewGRPC()
+		httpProbe    = prober.NewHTTP()
 		statusProber = prober.Combine(
 			httpProbe,
 			grpcProbe,
--- a/cmd/thanos/query.go
+++ b/cmd/thanos/query.go
@@ -227,7 +227,6 @@ func runQuery(
 	})
 	reg.MustRegister(duplicatedStores)
 
-
 	var storesConfig []store.Config
 	if len(storeConfigYAML) > 0 {
 		var err error
@@ -242,26 +241,26 @@ func runQuery(
 			}
 		}
 
-		endpointsConfig := store.EndpointsConfig {
+		endpointsConfig := store.EndpointsConfig{
 			StaticAddresses: storeAddrs,
 		}
 		if fileSDConfig != nil {
 			endpointsConfig.FileSDConfigs = []file.SDConfig{*fileSDConfig}
 		}
 		storeConfig := store.Config{
-			Name: "default",
+			Name:            "default",
 			EndpointsConfig: endpointsConfig,
 		}
 		if secure {
-			storeConfig.TlsConfig = &store.TlsConfig {
-				Cert: cert,
-				Key: key,
-				CaCert: caCert,
+			storeConfig.TlsConfig = &store.TlsConfig{
+				Cert:       cert,
+				Key:        key,
+				CaCert:     caCert,
 				ServerName: serverName,
 			}
 		}
 
-		storesConfig = []store.Config{ storeConfig }
+		storesConfig = []store.Config{storeConfig}
 	}
 
 	// todo: sml: remove - for debugging
@@ -312,12 +311,11 @@ func runQuery(
 			logger,
 			// todo: sml: looks a bit iffy
 			extprom.WrapRegistererWith(
-				map[string]string{"config_name":config.Name},
+				map[string]string{"config_name": config.Name},
 				extprom.WrapRegistererWithPrefix("thanos_querier_store_apis_", reg),
-		    ),
+			),
 			dns.ResolverType(dnsSDResolver),
 		)
-
 
 		stores := query.NewStoreSet(
 			logger,
@@ -354,11 +352,10 @@ func runQuery(
 		// Run File Service Discovery and update the store set when the files are modified.
 		if config.EndpointsConfig.FileSDConfigs != nil && len(config.EndpointsConfig.FileSDConfigs) > 0 {
 			var fileSDUpdates chan []*targetgroup.Group
-			ctxRun, cancelRun := context.WithCancel(context.Background())
-
 			fileSDUpdates = make(chan []*targetgroup.Group)
 
 			for _, fsdConfig := range config.EndpointsConfig.FileSDConfigs {
+				ctxRun, cancelRun := context.WithCancel(context.Background())
 				fileSD := file.NewDiscovery(&fsdConfig, logger)
 				g.Add(func() error {
 					fileSD.Run(ctxRun, fileSDUpdates)
--- a/cmd/thanos/query.go
+++ b/cmd/thanos/query.go
@@ -286,20 +286,11 @@ func runQuery(
 		)
 	)
 
-	// run through once so we don't leave funcs hanging
-	perConfigDialOpts := make(map[string][]grpc.DialOption)
 	for _, config := range storesConfig {
-
 		dialOpts, err := extgrpc.StoreClientGRPCOptsFromTlsConfig(logger, config.Name, reg, tracer, config.TlsConfig)
 		if err != nil {
 			return errors.Wrap(err, "building gRPC client")
 		}
-		perConfigDialOpts[config.Name] = dialOpts
-	}
-
-	for _, config := range storesConfig {
-
-		dialOpts := perConfigDialOpts[config.Name]
 
 		fileSDCache := cache.New()
 		dnsProvider := dns.NewProvider(
--- a/cmd/thanos/query.go
+++ b/cmd/thanos/query.go
@@ -233,31 +233,10 @@ func runQuery(
 			return errors.Wrap(err, "loading store config")
 		}
 	} else {
-		for _, addr := range storeAddrs {
-			if addr == "" {
-				return errors.New("static store address cannot be empty")
-			}
+		storeConfig, err := store.NewConfig(storeAddrs, fileSDConfig, secure, cert, key, caCert, serverName)
+		if err != nil {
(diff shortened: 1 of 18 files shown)


bwplotka:
Nice! Marking this as higher priority, would be nice to have review of this. 


Can you mark this un-wip when ready for review?

cc @krasi-georgiev and @s-urbaniak , @simonpasquier  for some help in review (:

eswdd:
Yup. Will be working on it later today or this evening.

eswdd:
@bwplotka have cleaned it up so all the PR checks pass and removed any debugging I had from development. Now available for review.

cmd/thanos/query.go
```diff
@@ -210,21 +225,84 @@ func runQuery(
...
 
-	dialOpts, err := extgrpc.StoreClientGRPCOpts(logger, reg, tracer, secure, cert, key, caCert, serverName)
-	if err != nil {
-		return errors.Wrap(err, "building gRPC client")
+	var storesConfig []store.Config
+	if len(storeConfigYAML) > 0 {
+		var err error
+		storesConfig, err = store.LoadConfig(storeConfigYAML)
+		if err != nil {
+			return err
```
s-urbaniak: Let's wrap the error here with some context, otherwise a very low level error will potentially bubble up.

bwplotka:
Awesome. All - even non-maintainers, please help in the review if you want this! e.g @lilic @paulfantom 

I will have time to look on this from next week. ):

cmd/thanos/query.go
```diff
@@ -210,21 +225,84 @@ func runQuery(
...
+		if secure {
+			storeConfig.TlsConfig = &store.TlsConfig{
+				Cert:       cert,
+				Key:        key,
+				CaCert:     caCert,
+				ServerName: serverName,
+			}
+		}
+
+		storesConfig = []store.Config{storeConfig}
```
s-urbaniak: suggestion: could we make the above block a constructor method in `store.NewConfig(storeAddrs, fileSDConfig, tlsConfig)` which potentially also returns an error?

cmd/thanos/query.go
```diff
@@ -210,21 +225,84 @@ func runQuery(
...
 	}
 
-	fileSDCache := cache.New()
-	dnsProvider := dns.NewProvider(
-		logger,
-		extprom.WrapRegistererWithPrefix("thanos_querier_store_apis_", reg),
-		dns.ResolverType(dnsSDResolver),
+	var storeSets []*query.StoreSet
```
s-urbaniak: `storeSets` is referenced in `allClients` but initialized later in the range over `storesConfig` which can be hard to reason about. Let me think if we can simplify this.

cmd/thanos/query.go
```diff
@@ -210,21 +225,84 @@ func runQuery(
...
+			promql.EngineOpts{
+				Logger:        logger,
+				Reg:           reg,
+				MaxConcurrent: maxConcurrentQueries,
+				// TODO(bwplotka): Expose this as a flag: https://github.com/thanos-io/thanos/issues/703.
+				MaxSamples: math.MaxInt32,
+				Timeout:    queryTimeout,
+			},
+		)
 	)
```
s-urbaniak: ok, played around locally, i suggest to move the whole `allClients` var block down after the for block in line 287, so we:
1. initialize `storesConfig`
2. initialize `storeSets`
3. initialize `proxy`, `queryableCreator`, and `engine`

that gives it (at least locally) a more sequential initialization flow and is easier to reason about, wdyt?

cmd/thanos/receive.go
```diff
@@ -194,7 +194,7 @@ func runReceive(
 	if err != nil {
 		return err
 	}
-	dialOpts, err := extgrpc.StoreClientGRPCOpts(logger, reg, tracer, rwServerCert != "", rwClientCert, rwClientKey, rwClientServerCA, rwClientServerName)
+	dialOpts, err := extgrpc.StoreClientGRPCOpts(logger, nil, reg, tracer, rwServerCert != "", rwClientCert, rwClientKey, rwClientServerCA, rwClientServerName)
```
s-urbaniak: unrelated to this PR (we should tackle this ): this rather lonely looking `nil` here (clientInstance) is a bit hard to reason about, maybe we could combine this in the effort done in #2267, cc @PhilipGough @kakkoyun ?

pkg/extgrpc/client.go
```diff
@@ -12,18 +12,38 @@ import (
...
 	"github.com/thanos-io/thanos/pkg/tls"
 	"github.com/thanos-io/thanos/pkg/tracing"
 	"google.golang.org/grpc"
 	"google.golang.org/grpc/credentials"
 )
 
+func StoreClientGRPCOptsFromTlsConfig(logger log.Logger, clientInstance string, reg *prometheus.Registry, tracer opentracing.Tracer, tlsConfig *store.TlsConfig) ([]grpc.DialOption, error) {
+	if tlsConfig != nil {
+		return StoreClientGRPCOpts(logger, &clientInstance, reg, tracer, true, tlsConfig.Cert, tlsConfig.Key, tlsConfig.CaCert, tlsConfig.ServerName)
+	} else {
```
s-urbaniak: let's just omit the `else`

pkg/extgrpc/client.go
```diff
@@ -12,18 +12,38 @@ import (
...
-func StoreClientGRPCOpts(logger log.Logger, reg *prometheus.Registry, tracer opentracing.Tracer, secure bool, cert, key, caCert, serverName string) ([]grpc.DialOption, error) {
-	grpcMets := grpc_prometheus.NewClientMetrics()
-	grpcMets.EnableClientHandlingTimeHistogram(
-		grpc_prometheus.WithHistogramBuckets([]float64{0.001, 0.01, 0.1, 0.3, 0.6, 1, 3, 6, 9, 20, 30, 60, 90, 120}),
+func StoreClientGRPCOpts(logger log.Logger, clientInstance *string, reg *prometheus.Registry, tracer opentracing.Tracer, secure bool, cert, key, caCert, serverName string) ([]grpc.DialOption, error) {
+	var grpcMets *grpc_prometheus.ClientMetrics
+	histogramBuckets := grpc_prometheus.WithHistogramBuckets(
+		[]float64{0.001, 0.01, 0.1, 0.3, 0.6, 1, 3, 6, 9, 20, 30, 60, 90, 120},
 	)
+	if clientInstance != nil {
```
s-urbaniak: Instead of passing a string pointer, let's simply pass a value and decide on empty string `""`. Go doesn't have optionals, but signalling the absence of a value in Go often is idiomatically done in Go simply by signalling it's zero value, which in the case of `string` is empty string `""`.

pkg/query/storeset_test.go
```diff
@@ -521,7 +521,7 @@ func TestStoreSet_Update_NoneAvailable(t *testing.T) {
 	st.CloseOne(initialStoreAddr[0])
 	st.CloseOne(initialStoreAddr[1])
 
-	storeSet := NewStoreSet(nil, nil, func() (specs []StoreSpec) {
+	storeSet := NewStoreSet(nil, "default", nil, func() (specs []StoreSpec) {
```
s-urbaniak: instead of passing `"default"` here, let's set it in the `NewStoreSet` constructor, and just pass empty string `""` here. 

Rationale: The fact that `"default"` is set is an implementation detail, constructor invokers can simply pass empty string and do not care, plus we won't forget to set `"default"` if we invoke `NewStoreSet` somewhere else.

pkg/store/config.go
```diff
@@ -0,0 +1,71 @@
...
+
+import (
+	"fmt"
+
+	"gopkg.in/yaml.v2"
+
+	"github.com/prometheus/prometheus/discovery/file"
+)
+
+type Config struct {
```
s-urbaniak: Do you mind to add some godoc?

pkg/store/config.go
```diff
@@ -0,0 +1,71 @@
...
+	Name            string          `yaml:"name"`
+	TlsConfig       *TlsConfig      `yaml:"tls_config"`
+	EndpointsConfig EndpointsConfig `yaml:",inline"`
+}
+
+type TlsConfig struct {
```
s-urbaniak: ditto (albeit obvious), but maybe describe the tls config of what is this actually.

pkg/store/config.go
```diff
@@ -0,0 +1,71 @@
...
+	Cert string `yaml:"cert_file"`
+	// TLS Key for the client's certificate.
+	Key string `yaml:"key_file"`
+	// TLS CA Certificates to use to verify gRPC servers.
+	CaCert string `yaml:"ca_file"`
+	// Server name to verify the hostname on the returned gRPC certificates. See https://tools.ietf.org/html/rfc4366#section-3.1
+	ServerName string `yaml:"server_name"`
+}
+
+type EndpointsConfig struct {
```
s-urbaniak: ditto

s-urbaniak:
@eswdd thank you a lot for the contribution! :tada: I left a few comments.

eswdd:
Thanks @s-urbaniak, will take a look today/tomorrow and get back to you.

cmd/thanos/query.go
eswdd: yup, makes a lot of sense. although when I did it locally it made no sense having a var block then a bunch of variables initialised with the probers and then the code block below, so I combined the prober initialisation into the var block too. looks cleaner IMO.

cmd/thanos/receive.go
eswdd: It has of course now turned into a lonely looking `""`

eswdd:
@s-urbaniak have implemented all the suggestions, PTAL

s-urbaniak:
For me this is looking great, thank you! :tada: If there is a last nit, I would love to see some unmarshaling test (especially with the DefaultConfig handling) in `pkg/store/config.go`.

else this LGTM

Leaving the final pair of eyes to @bwplotka

bwplotka:
Sorry for lag, let's wait until release is done... we are working heavily on it, then I will look on it closely, promise (: 

Let's rebase, but I think rebase can wait until the merge release-0.12 -> master will happen. It might introduce some conflicts anyway.

Thanks @eswdd for patience!

eswdd:
@bwplotka sorry, was off last week, so only just seen this - would have wanted to get this into the release but I guess too late now. When do you expect to cut the next release?

simonpasquier:
According to the [release planning](https://thanos.io/release-process.md/#cadence), the first RC of v0.13.0 is slated for the 13th of May.

bwplotka:
Let's get back to this... can you rebase @eswdd ?

bwplotka:
It's huge change plus lots of things changes since last rebase, so unlikely to fit 0.13

amrmahdi:
Thanks @eswdd for this change, very much needed. 
@bwplotka we are at 0.14.0 train now, would this make it to 0.14.0 ?

bwplotka:
Sorry for huge delay on this. Let's get back to this. We might have more time this week! (: (PromCon was last week).

Do you mind reading?

s-urbaniak:
@bwplotka i think this is still useful, PTAL :)

bwplotka:
We need rebase :thinking:

markmsmith:
FWIW, I'm still interested in this.  Hopefully the original author can check in before the stale bot kills it.

markmsmith:
Is this still blocked on a reabase from @eswdd ?

PR #2227, opened 2020-03-07
added insecureDialer to allow simple TLS-connections without needing any TLS-config

Zenlil:
## Changes

Added a network dialer that supports plain TLS for HTTPS connections (without client or server-certs).
If checks if the address ends with ":443" and if so uses tls.Dial instead of net.Dial

## Verification

Have several own k8 clusters using nginx-ingress with cert-manager.
Added the folliwing arguments to my Thanos Query:
--store=thanos-sidecar-0.cluster1.mydomain.com:443
--store=thanos-sidecar-1.cluster1.mydomain.com:443
--store=thanos-sidecar-0.cluster2.mydomain.com:443
--store=thanos-sidecar-1.cluster2.mydomain.com:443
a.s.o

Works really well i.m.o

### DIFF
--- a/pkg/extgrpc/client.go
+++ b/pkg/extgrpc/client.go
@@ -4,7 +4,11 @@
 package extgrpc
 
 import (
+	"context"
+	gotls "crypto/tls"
 	"math"
+	"net"
+	"strings"
 
 	"github.com/go-kit/kit/log"
 	"github.com/go-kit/kit/log/level"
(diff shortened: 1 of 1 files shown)


lilic:
cc @bwplotka do you mind having a look at this, thanks!

Zenlil:
I did think that doing the fix where the grpc.Dial is done would be a better solution, but that's done in a completely different location, and one that do not know if there is a `WithInsecure`in the dial-options or not.

If doing it there we need a way to propagate the `secure`-flag from the `StoreClientGRPCOpts` all the way to the `grpc.Dial` and affect the dial-options.

And, since there is multiple-tls-configs on the way I thought this could work for the "low-handing-fruit"-use-case.

bwplotka:
Well temporary here would mean to enforce some option that has to be maintained then removed while breaking some compatibility. It would be nice to avoid this (:

PR #2228, opened 2020-03-07
Blocked classic prometheus constructors, moved all to promauto; Removed unnecessary printfs.

bwplotka:
This is thanks to https://github.com/fatih/faillint/issues/7

### DIFF
--- a/pkg/alert/alert.go
+++ b/pkg/alert/alert.go
@@ -23,6 +23,7 @@ import (
 	"github.com/pkg/errors"
 	"github.com/prometheus/alertmanager/api/v2/models"
 	"github.com/prometheus/client_golang/prometheus"
+	"github.com/prometheus/client_golang/prometheus/promauto"
 	"github.com/prometheus/prometheus/pkg/labels"
 
 	"github.com/thanos-io/thanos/pkg/runutil"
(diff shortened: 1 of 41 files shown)


pkg/alert/alert.go
```diff
@@ -133,34 +134,31 @@ func NewQueue(logger log.Logger, reg prometheus.Registerer, capacity, maxBatchSi
...
 			Help: "Total number of alerts pushed to the queue.",
 		}),
-		popped: prometheus.NewCounter(prometheus.CounterOpts{
+		popped: promauto.With(reg).NewCounter(prometheus.CounterOpts{
 			Name: "thanos_alert_queue_alerts_popped_total",
 			Help: "Total number of alerts popped from the queue.",
 		}),
 	}
-	capMetric := prometheus.NewGaugeFunc(prometheus.GaugeOpts{
+	_ = promauto.With(reg).NewGaugeFunc(prometheus.GaugeOpts{
```
squat: Is it better to do `_ = func()` or just `func()`?

bwplotka: Well I do `_` to indicate that this returns value, but we don't need it. Do you think it's wrong pattern?

bwplotka: I like it as it avoid surprises (: But happy to change this in the following PRs :hugs:

squat: I just wanted to make sure we were being consistent in the project one way or the other so :+1:

PR #2231, opened 2020-03-07
Bucket Web - Sort chunks by thanos.downsample.resolution for better grouping

kfdm:
## Changes

Sort series on bucket viewer page to better group compaction levels

Before

before

After

after

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -14,8 +14,9 @@ We use *breaking* word for marking changes that are not backward compatible (rel
 ### Fixed
 
 - [#2238](https://github.com/thanos-io/thanos/pull/2238) Ruler: Fixed Issue #2204 bug in alert queue signalling filled up queue and alerts were dropped
+- [#2231](https://github.com/thanos-io/thanos/pull/2231) Bucket Web - Sort chunks by thanos.downsample.resolution for better grouping
 
-## [v0.11.0](https://github.com/thanos-io/thanos/releases/tag/v0.11.0-rc.1) - 2020.03.02
+## [v0.11.0](https://github.com/thanos-io/thanos/releases/tag/v0.11.0) - 2020.03.02
 
 ### Fixed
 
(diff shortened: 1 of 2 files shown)


kfdm:
Ok, should be rebased 👍

CHANGELOG.md
```diff
@@ -14,6 +14,7 @@ We use *breaking* word for marking changes that are not backward compatible (rel
 ### Fixed
 
 - [#2238](https://github.com/thanos-io/thanos/pull/2238) Ruler: Fixed Issue #2204 bug in alert queue signalling filled up queue and alerts were dropped
+- [#2231](https://github.com/thanos-io/thanos/pull/2231) Bucket Web - Sort chunks by thanos.downsample.resolution for better grouping
```
bwplotka: Is this really Fixed? I think we might rather put it to `changed`?

CHANGELOG.md
```diff
@@ -14,6 +14,7 @@ We use *breaking* word for marking changes that are not backward compatible (rel
...
 
 ## [v0.11.0](https://github.com/thanos-io/thanos/releases/tag/v0.11.0-rc.1) - 2020.03.02
```
bwplotka: It's not your bug, can you help and fix this link? :hugs: 

```suggestion
## [v0.11.0](https://github.com/thanos-io/thanos/releases/tag/v0.11.0) - 2020.03.02
```

CHANGELOG.md
kfdm: I think it was previously sorted properly but was changed when things migrated to block.MetaFetcher

GiedriusS: Yes, I think it should be in `Fixed`. The fetching of different metas happens concurrently now and I think previously that wasn't the case.

kfdm:
Rebased once more to make sure the DOC should still be there.

CHANGELOG.md
GiedriusS: I agree with @kfdm here. I had to recheck the code and it seems like @kfdm is correct.

PR #2233, opened 2020-03-08
Add extra benchmark GithubAction job to CI for each PRs, master push and releases.

bwplotka:
Fixes https://github.com/thanos-io/thanos/pull/2095

With this we run ALL benchmarks:

* For All PRs: Run all go  bench benchmarks for the current PR and compare with `master`
* For push to master: Run all go  bench benchmarks in master and compare with `latest release"
* For new release: Run all go  bench benchmarks for the release and compare with `the previous minor release`

Benchmark runs should be cached as well.

### DIFF
--- a/.github/workflows/benchmark-pr.yaml
+++ b/.github/workflows/benchmark-pr.yaml
@@ -13,17 +13,13 @@ jobs:
         uses: actions/setup-go@v1
         with:
           go-version: 1.13.6
-
       - name: Check out code into the Go module directory.
         uses: actions/checkout@v2
-
-      - name: Give me just shared dir lol GitHub Action!
+      - name: Give me just shared dir GitHub Action!
         run: mkdir /tmp/cache
-
       - uses: actions/cache@v1
         with:
           path: /tmp/cache
           key: benchmark-results
-
       - name: Run selected benchmarks and compare with newest release.
-        run: make bench THANOS_BENCH_FUNC="BenchmarkSeries|BenchmarkBucketIndexReader_ExpandedPostings"
+        run: make bench THANOS_BENCH_FUNC="BenchmarkSeries"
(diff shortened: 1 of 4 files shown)


krasi-georgiev:
We tried this as well, but GH actions give very different results for CPU every time even without code changes so these would be good only for RAM benchmarking and not for CPU

bwplotka:
Also: `BenchmarkDedupSeriesIterator/equal-2         	fatal error: runtime: out of memory` ;p

Anyway, this needs more work... 

I won't have much time for this in following week, anyone is free to continue on this (:

.github/workflows/benchmark-pr.yaml
```diff
@@ -0,0 +1,26 @@
+name: benchmark
+
+on: pull_request
```
yeya24: We can use `path-ignore` to ignore only docs/mixins change
```
on:
  push:
    paths-ignore:
    - 'docs/**'
```

bwplotka: Actually wow! We can divide those things in current CI.. just we need to move to GH actions! That's amazing.

I think we should create an GH issuse for that (:

bwplotka: https://github.com/thanos-io/thanos/issues/2257

bwplotka:
Ok trying again, this time with limited number of benchmarks (:

daixiang0:
Add a tool or something to compare results could be better i think.

bwplotka:
We have already such tool. See
https://github.com/prometheus/test-infra/tree/master/funcbench

What's wrong with that?

bwplotka:
e2e is crying with the amount of time and resources our benchmarks need so far, so we need to find some solution. 

Help wanted, anyone can hop on this PR and help :hugs: 

cc @daixiang0

daixiang0:
@bwplotka i would learn Github action first then look this.

daixiang0:
@bwplotka seems i can not push to this branch, i test in another pr.

PR #2237, opened 2020-03-09
pkg/store/storepb: fix wrong rule reference

s-urbaniak:
Currently we recursively reference rules instead of recording rules.

cc @bwplotka


bwplotka:
Resolving conflicts time!

PR #2238, opened 2020-03-09
ruler: Fix #2204 bug where alert queue is unpoppable causing full queue and dropped alerts

### LINKED ISSUE #2204: Dropping alerts from ruler

robincw-gr:

## Changes

Fixes Issue #2204. The bug is that `alert.Queue.Push` can be called multiple times to add alerts to the queue, but it only provides a single signal on `morec` for `alert.Queue.Pop` to know that the queue has changed, i.e. there are alerts waiting. Further calls to `Push` can fill the queue and use `select` ... `default:` to skip signalling if the `morec` channel's buffer (size 1) is full. However, Pop will only remove `maxBatchSize` alerts from the queue and further calls to `Pop` block because the signal has been read. Therefore Pop fails to remove remaining alerts until `Push` has been called at least once again to restore the signal on `morec`.
This fix adds to `alert.Queue.Pop` so that if the queue is not empty after popping it restores the signal it read on the `morec` channel to indicate there are still more alerts it hasn't removed.

## Verification

1. This bug was noticed in production where the `thanos_alert_queue_alerts_pushed_total` metric was increasing in a short timespan but the `thanos_alert_queue_alerts_popped_total` was not keeping up, and `thanos_alert_queue_length` grew to almost 10,000 and remained there.

2. This change adds a test to ensure that after multiple calls to `alert.Queue.Push` to add more than `maxBatchSize` alerts to the queue, multiple calls to `alert.Queue.Pop` will work until all the pushed alerts have been popped.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -11,6 +11,10 @@ We use *breaking* word for marking changes that are not backward compatible (rel
 
 ## Unreleased
 
+### Fixed
+
+- [#TODO](https://github.com/thanos-io/thanos/pull/#TODO) Ruler: Fixed Issue #2204 bug in alert queue signalling filled up queue and alerts were dropped
+
 ## [v0.11.0](https://github.com/thanos-io/thanos/releases/tag/v0.11.0-rc.1) - 2020.03.02
 
 ### Fixed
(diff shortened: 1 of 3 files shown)


robincw-gr:
Would you like to take a look @yeya24 ?

PR #2239, opened 2020-03-09
tutorials: fix incorrect query

kryptonite303:
You would have to query `prometheus_tsdb_head_series` instead of `sum(prometheus_tsdb_head_series)` in order to get the 5 results when deduplicating.


kryptonite303:
Fixed DCO

PR #2240, opened 2020-03-09
cmd/thanos/query: add initial rules support

s-urbaniak:
This adds basic support for the rule parameter into the Thanos Query component. Marking as WIP as I'd like to have feedback about whether the trajectory actually is fit.

Bear with me, I am new here :-)

cc @bwplotka


cmd/thanos/query.go
```diff
@@ -76,6 +76,9 @@ func registerQuery(m map[string]setupFunc, app *kingpin.Application) {
 	stores := cmd.Flag("store", "Addresses of statically configured store API servers (repeatable). The scheme may be prefixed with 'dns+' or 'dnssrv+' to detect store API servers through respective DNS lookups.").
 		PlaceHolder("<store>").Strings()
 
+	rules := cmd.Flag("rule", "Addresses of statically configured rules API servers (repeatable). The scheme may be prefixed with 'dns+' or 'dnssrv+' to detect store API servers through respective DNS lookups.").
```
bwplotka: Worth to mention that RulesAPI are now required to be storeAPI as well? (info endpoint needed).

Alternatively... We could define INFO endpoint for Rules as well (same as StoreAPI) and treat those separatedly. I think I kind of like this idea :thinking:

cmd/thanos/query.go
```diff
@@ -215,30 +217,46 @@ func runQuery(
...
 				specs = removeDuplicateStoreSpecs(logger, duplicatedStores, specs)
 
 				return specs
 			},
+			func() (specs []query.RuleSpec) {
+				for _, addr := range dnsRuleProvider.Addresses() {
+					specs = append(specs, query.NewGRPCStoreSpec(addr))
+				}
+
+				// no need to remove duplicates, as rule apis are a subset of store apis.
```
bwplotka:
```suggestion
				// NOTE(s-urbaniak): No need to remove duplicates, as rule apis are a subset of store apis.
```

cmd/thanos/query.go
```diff
@@ -289,7 +307,8 @@ func runQuery(
 					}
 					fileSDCache.Update(update)
 					stores.Update(ctxUpdate)
-					dnsProvider.Resolve(ctxUpdate, append(fileSDCache.Addresses(), storeAddrs...))
+					dnsStoreProvider.Resolve(ctxUpdate, append(fileSDCache.Addresses(), storeAddrs...))
+					// rules apis do not support file service discovery as of now.
```
bwplotka: Full sentence please (First letter uppercase) 

Our CI check did find  that? :thinking:

cmd/thanos/query.go
```diff
@@ -403,3 +423,21 @@ func removeDuplicateStoreSpecs(logger log.Logger, duplicatedStores prometheus.Co
 	}
 	return deduplicated
 }
+
+type stringSlice []string
+
+// duplicate returns a duplicate string in the given stringSlice
```
bwplotka: ditto

cmd/thanos/query.go
```diff
@@ -403,3 +423,21 @@ func removeDuplicateStoreSpecs(logger log.Logger, duplicatedStores prometheus.Co
...
+// or empty string if none was found.
+func (ss stringSlice) duplicate() string {
```
bwplotka: What about just passing slice? Not sure if it's worth to create custom stringSlice type for it, cannot see advantage

pkg/query/storeset.go
```diff
@@ -207,6 +218,9 @@ func NewStoreSet(
 type storeRef struct {
 	storepb.StoreClient
 
+	// if not nil, then this store also supports rules API.
```
bwplotka: full sentence

pkg/query/storeset.go
```diff
@@ -418,7 +438,14 @@ func (s *StoreSet) getHealthyStores(ctx context.Context, stores map[string]*stor
 					level.Warn(s.logger).Log("msg", "update of store node failed", "err", errors.Wrap(err, "dialing connection"), "address", addr)
 					return
 				}
-				st = &storeRef{StoreClient: storepb.NewStoreClient(conn), cc: conn, addr: addr, logger: s.logger}
+
+				var rule storepb.RulesClient
+
+				if _, ok := ruleAddrSet[addr]; ok {
```
bwplotka: What's confusing here, is that if someone put something in `--rules` but not `--stores` it will be silently dropped

pkg/query/storeset_test.go
```diff
@@ -27,7 +27,12 @@ var testGRPCOpts = []grpc.DialOption{
 }
 
 type testStore struct {
-	info storepb.InfoResponse
+	info      storepb.InfoResponse
+	ruleGroup *storepb.RuleGroup
+}
+
+func (s *testStore) Rules(_ *storepb.RulesRequest, srv storepb.Rules_RulesServer) error {
+	return srv.Send(storepb.NewRuleGroupResponse(s.ruleGroup))
```
bwplotka: Hm.. this is quite inconsistent. We don't test Series this way. I would either do Series the same as move rule group testing somewhere else (as it is with Series).

Most likely e2e is the correct place

pkg/store/proxy_test.go
```diff
@@ -1138,6 +1144,7 @@ func TestProxyStore_LabelNames(t *testing.T) {
 				nil,
 				nil,
 				func() []Client { return tc.storeAPIs },
+				nil,
```
bwplotka: I think here would be actually the correct place to test it, no?

cmd/thanos/query.go
s-urbaniak: whopsie

cmd/thanos/query.go
s-urbaniak: it was just easier for my eyes, I have no strong opinions here, will change to parameter instead of type :+1:

pkg/query/storeset.go
s-urbaniak: Indeed, that is a fair edge case. I will implement a small assertion bailing out if `--rules` is given but not `--store`.

pkg/query/storeset_test.go
s-urbaniak: ah thanks, i really wasn't sure either if this is the right place.

pkg/store/proxy_test.go
s-urbaniak: sounds good :+1:

cmd/thanos/query.go
s-urbaniak: I think the INFO endpoint idea is great. Generally as far as I understood we could leverage the INFO endpoint on RulesAPI to communicate back if this store endpoint actually is capable to serve RulesAPI. As far as I see we would have two cases:

- sidecar, rule: these components signal back the capability to server the RulesAPI.
- query, store, receive: these components signal back that they do not support the RulesAPI and always error out on the actual remote methods.

Does that sound reasonable at all?

cmd/thanos/query.go
s-urbaniak: fixed

cmd/thanos/query.go
s-urbaniak: fixed

cmd/thanos/query.go
s-urbaniak: fixed

pkg/store/proxy_test.go
s-urbaniak: Ah i understand this area now better. Well, I explicitely wanted to test storeset handling but I can add an more thorough test here indeed.

cmd/thanos/query.go
s-urbaniak: @bwplotka just to understand: if we don't want to use INFO endpoint for discoverability, what would be the benefit of implementing it redundantly?

pkg/query/storeset.go
s-urbaniak: I believe we cannot handle this at startup time because of the other discoverability flags, so instead logging out a warning as per https://github.com/thanos-io/thanos/pull/2240/commits/c054ca8dc8bff6e7632c5c0239d1dea212a8ca99

cmd/thanos/query.go
bwplotka: The only benefit would be: To have a service that ONLY implements RulesAPI and we can healthcheck it.

pkg/query/storeset.go
bwplotka: :+1:

pkg/store/prometheus.go
```diff
@@ -91,6 +91,10 @@ func NewPrometheusStore(
 	return p, nil
 }
 
+func (p *PrometheusStore) RulesInfo(context.Context, *storepb.RulesInfoRequest) (*storepb.RulesInfoResponse, error) {
```
bwplotka, line 94: Discussed offline ):

cmd/thanos/query.go
```diff
@@ -65,7 +65,7 @@ func registerQuery(m map[string]setupFunc, app *kingpin.Application) {
 	maxConcurrentQueries := cmd.Flag("query.max-concurrent", "Maximum number of queries processed concurrently by query node.").
 		Default("20").Int()
 
-	replicaLabels := cmd.Flag("query.replica-label", "Labels to treat as a replica indicator along which data is deduplicated. Still you will be able to query without deduplication using 'dedup=false' parameter.").
+	queryReplicaLabels := cmd.Flag("query.replica-label", "Labels to treat as a replica indicator along which data is deduplicated. Still you will be able to query without deduplication using 'dedup=false' parameter.").
```
bwplotka, line 68: Worth to mention that we use same labels for rules dedup?

cmd/thanos/query.go
bwplotka, line 79: We might want to be clear that Rules API has to implement Info as well to work

cmd/thanos/query.go
```diff
@@ -215,30 +217,46 @@ func runQuery(
...
 				specs = removeDuplicateStoreSpecs(logger, duplicatedStores, specs)
 
 				return specs
 			},
+			func() (specs []query.RuleSpec) {
+				for _, addr := range dnsRuleProvider.Addresses() {
+					specs = append(specs, query.NewGRPCStoreSpec(addr))
+				}
+
+				// NOTE(s-urbaniak): No need to remove duplicates, as rule apis are a subset of store apis.
```
bwplotka, line 251: hmm, really?

bwplotka: I thought this is fine by design, just you need to make sure RulesAPI implements StoreAPI as well?

cmd/thanos/query.go
povilasv, line 79:
```suggestion
	rules := cmd.Flag("rule", "Addresses of statically configured rules API servers (repeatable). The scheme may be prefixed with 'dns+' or 'dnssrv+' to detect rule API servers through respective DNS lookups.").
```

cmd/thanos/query.go
```diff
@@ -190,9 +192,9 @@ func runQuery(
 	maxConcurrentQueries int,
 	queryTimeout time.Duration,
 	storeResponseTimeout time.Duration,
-	replicaLabels []string,
+	queryReplicaLabels []string,
 	selectorLset labels.Labels,
-	storeAddrs []string,
+	storeAddrs []string, ruleAddrs []string,
```
povilasv, line 197:
```suggestion
	storeAddrs []string,
	ruleAddrs []string,
```

cmd/thanos/query.go
```diff
@@ -369,7 +389,7 @@ func runQuery(
 		}
 
 		// TODO: Add rules API implementation when ready.
```
povilasv, line 391: remove the TODO :dancing_men:

cmd/thanos/query.go
```diff
@@ -404,3 +424,19 @@ func removeDuplicateStoreSpecs(logger log.Logger, duplicatedStores prometheus.Co
...
 	return deduplicated
 }
+
+// Duplicate returns a duplicate string in the given string slice
+// or empty string if none was found.
+func duplicate(ss []string) string {
+	set := map[string]struct{}{}
+
+	for _, s := range []string(ss) {
+		if _, ok := set[s]; ok {
```
povilasv, line 434: what if multiple duplicates in string slice?

docs/components/query.md
```diff
@@ -327,6 +327,11 @@ Flags:
                                  prefixed with 'dns+' or 'dnssrv+' to detect
                                  store API servers through respective DNS
                                  lookups.
+      --rule=<rule> ...          Addresses of statically configured rules API
+                                 servers (repeatable). The scheme may be
+                                 prefixed with 'dns+' or 'dnssrv+' to detect
+                                 store API servers through respective DNS
```
povilasv, line 333:
```suggestion
                                 Rule API servers through respective DNS
```

pkg/query/storeset.go
```diff
@@ -418,11 +437,18 @@ func (s *StoreSet) getHealthyStores(ctx context.Context, stores map[string]*stor
...
+					rule = storepb.NewRulesClient(conn)
+				}
+
+				st = &storeRef{StoreClient: storepb.NewStoreClient(conn), rule: rule, cc: conn, addr: addr, logger: s.logger}
 			}
 
 			// Check existing or new store. Is it healthy? What are current metadata?
-			labelSets, minTime, maxTime, storeType, err := spec.Metadata(ctx, st.StoreClient)
+			labelSets, minTime, maxTime, Type, err := spec.Metadata(ctx, st.StoreClient)
```
povilasv, line 451:
```suggestion
			labelSets, minTime, maxTime, type, err := spec.Metadata(ctx, st.StoreClient)
```

local variables should be lower case

pkg/store/proxy.go
```diff
@@ -50,13 +50,127 @@ type Client interface {
...
+	for _, g := range groups {
+		if err := srv.Send(storepb.NewRuleGroupRulesResponse(g)); err != nil {
+			return status.Error(codes.Unknown, errors.Wrap(err, "send rules response").Error())
+		}
+	}
+
+	return nil
+}
+
+func dedupGroups(groups []*storepb.RuleGroup) []*storepb.RuleGroup {
```
povilasv, line 100: I thought we are going to do streaming dedup, not convert everything into slices and then dedup?

cmd/thanos/query.go
s-urbaniak: indeed, fixed :+1:

cmd/thanos/query.go
s-urbaniak: Correct, currently RulesAPI is a strict subset of StoreAPI, and we currently rely on that invariant.

pkg/query/storeset.go
s-urbaniak: oh doh, that was an old remainder of some recursive regex search/replace, thanks!

pkg/store/proxy.go
s-urbaniak: I would love to leave real stream based deduplication as an exercise for the future, as currently I see the problem, that we cannot really stream back to the client any rule, until we received the youngest rule and/or (firing) alert from all replicas.

cmd/thanos/query.go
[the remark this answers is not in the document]
s-urbaniak: indeed, thank you, fixed :+1:

pkg/query/api/v1.go
```diff
@@ -28,6 +28,8 @@ import (
 	"strconv"
 	"time"
 
+	"github.com/prometheus/prometheus/rules"
+
```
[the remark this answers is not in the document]
s-urbaniak: indeed, fixed

pkg/query/storeset.go
```diff
@@ -443,6 +469,12 @@ func (s *StoreSet) getHealthyStores(ctx context.Context, stores map[string]*stor
 	}
 	wg.Wait()
 
+	for ruleAddr := range ruleAddrSet {
+		if _, ok := storeAddrSet[ruleAddr]; !ok {
+			level.Warn(s.logger).Log("msg", "ignored rule store", "address", ruleAddr)
```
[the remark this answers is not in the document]
s-urbaniak: Re: I was thinking of this as well, but I am wondering if `thanos_rule_nodes_grpc_connections` is needed as Rules API is a strict subset of Store API, wdyt?

cmd/thanos/query.go
s-urbaniak: indeed, fixed :+1:

cmd/thanos/query.go
s-urbaniak: oki :+1:

cmd/thanos/query.go
s-urbaniak: weeeee :tada: :-)

cmd/thanos/query.go
s-urbaniak: I have to admit I took the existing code mechanically, but the user will only be informed about the first duplicate. As far as I can we are retaining the existing mechanism. Happy to enhance this here though, wdyt?

docs/components/query.md
s-urbaniak: regenerated :+1:

pkg/store/proxy.go
povilasv: ok future sounds good :+1:

cmd/thanos/query.go
povilasv: we can rename it to `firstDuplicate` and LGTM :D

pkg/store/proxy.go
```diff
@@ -50,13 +50,127 @@ type Client interface {
...
+		rs := &rulesStream{
+			client:  rulesClient,
+			request: req,
+			channel: respChan,
+			server:  srv,
+		}
+		g.Go(func() error { return rs.receive(gctx) })
+	}
+
+	for resp := range respChan {
```
GiedriusS, line 80: Does it work? What happens when all of the senders finish sending their data? AFAICT no one closes this channel. My tests also show that goroutines get stuck on:

```
16 @ 0x437580 0x405f17 0x405c1b 0x16df6f0 0xafbe49 0x1891f6e 0xe2a82f 0xe306c4 0xe32474 0xe2a82f 0x188485b 0xe2a82f 0xe2a9ef 0xab9dee 0xabbf76 0xac9de1 0x4675e1
#	0x16df6ef	github.com/thanos-io/thanos/pkg/store.(*ProxyStore).Rules+0x2cf						/home/gstatkevicius/dev/thanos/pkg/store/proxy.go:80
#	0xafbe48	github.com/thanos-io/thanos/pkg/store/storepb._Rules_Rules_Handler+0x108				/home/gstatkevicius/dev/thanos/pkg/store/storepb/rpc.pb.go:1210
#	0x1891f6d	github.com/grpc-ecosystem/go-grpc-middleware/recovery.StreamServerInterceptor.func1+0xad		/home/gstatkevicius/go/pkg/mod/github.com/grpc-ecosystem/go-grpc-middleware@v1.1.0/recovery
```

pkg/store/proxy.go
```diff
@@ -50,13 +50,127 @@ type Client interface {
...
+		}
+
+		if w := rule.GetWarning(); w != "" {
+			if err := stream.server.Send(storepb.NewWarningRulesResponse(errors.New(w))); err != nil {
+				return errors.Wrapf(err, "sending rules warning to server %v", stream.server)
+			}
+			continue
+		}
+
+		stream.channel <- rule.GetGroup()
```
GiedriusS, line 170: Don't you need to check here if the context has ended? There might not be any readers left

pkg/query/api/v1.go
```diff
@@ -626,3 +637,54 @@ func (api *API) labelNames(r *http.Request) (interface{}, []error, *ApiError) {
...
+		apiRuleGroup := &storepb.RuleGroup{
+			Name:                              grp.Name,
+			File:                              grp.File,
+			Interval:                          grp.Interval,
+			EvaluationDurationSeconds:         grp.EvaluationDurationSeconds,
+			LastEvaluation:                    grp.LastEvaluation,
+			DeprecatedPartialResponseStrategy: grp.DeprecatedPartialResponseStrategy,
+			PartialResponseStrategy:           grp.PartialResponseStrategy,
+		}
+
```
povilasv, line 669:
```suggestion
apiRuleGroup.Rules = make(,len(..))
```
Consider pereallocating rule slice -> https://www.bwplotka.dev/2020/how-thanos-would-program-in-go/#pre-allocating-slices-and-maps

pkg/query/rules.go
```diff
@@ -0,0 +1,62 @@
...
+	proxy storepb.RulesServer
+}
+
+func (rr *rulesRetriever) RuleGroups(ctx context.Context) ([]*storepb.RuleGroup, storage.Warnings, error) {
+	resp := &rulesServer{ctx: ctx}
+
+	if err := rr.proxy.Rules(&storepb.RulesRequest{
+		PartialResponseStrategy: storepb.PartialResponseStrategy_ABORT,
+	}, resp); err != nil {
+		return nil, nil, errors.Wrap(err, "proxy RuleGroups()")
```
povilasv, line 30: consider not captilazing error strings https://github.com/golang/go/wiki/CodeReviewComments#error-strings

pkg/query/rules.go
```diff
@@ -0,0 +1,62 @@
...
+
+type rulesRetriever struct {
+	proxy storepb.RulesServer
+}
+
+func (rr *rulesRetriever) RuleGroups(ctx context.Context) ([]*storepb.RuleGroup, storage.Warnings, error) {
+	resp := &rulesServer{ctx: ctx}
+
+	if err := rr.proxy.Rules(&storepb.RulesRequest{
+		PartialResponseStrategy: storepb.PartialResponseStrategy_ABORT,
```
povilasv, line 28: This should be configureable, no?

s-urbaniak:
@GiedriusS @povilasv please bear with me, it is still WIP :-) I'll ping once I remove the tag, thanks for the early feedback!!! :heart:

pkg/store/proxy.go
s-urbaniak: yes, i just found it too :+1: added initial e2e tests to cover this. we need to start a goroutine and wait for the errgroup to finish before closing the response channel.

pkg/store/proxy.go
s-urbaniak: indeed, great catch :+1:

pkg/query/api/v1.go
s-urbaniak: agree, that's a good one :+1:

pkg/query/rules.go
s-urbaniak: hmm ... I wanted to stay consistent with https://github.com/thanos-io/thanos/blob/4240e3c743ba56969d32ad818a9208de0732f691/pkg/query/querier.go#L208

wdyt?

pkg/query/rules.go
s-urbaniak: yes :+1: I was just impatient and wanted to have something working :)

s-urbaniak:
@GiedriusS @povilasv @bwplotka I added an initial (very simple) e2e test, spinning up a Thanos Querier + Sidecar (so far the two components implementing the API). So far no deduplication logic yet, bear with me :-)

Also ... a general neutral statement from someone contributing the first time: I really really enjoy the code base and the great e2e testing framework capabilities here, so much goodness here and easy to start locally :heart:

pkg/query/rules.go
s-urbaniak: plus (note to myself): once I have also configurability injected here, this is where I want to implement the deduplication, and of course it needs unit tests here too.

pkg/store/proxy.go
s-urbaniak: rationale of why you observed the above: effectively the response channel never got closed ;-)

pkg/query/storeset.go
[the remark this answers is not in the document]
s-urbaniak: @simonpasquier as an after thought, I think this is a good suggestion :+1: Implementing it now too.

s-urbaniak:
@GiedriusS @povilasv @bwplotka Do you believe we could merge this into the feature branch? It would make logistics easier for me to submit additional PRs for the actual deduplication which I am working on right now. Thank you! :-)

bwplotka:
Let's go! :muscle: 

We can address potential comments in next PRs!

bwplotka:
AFAIK all comments by @povilasv and @GiedriusS  were fixed.

PR #2242, opened 2020-03-09
Made storepb.RuleGroups a source of truth for rules API (Go, JSON, proto). Added tests.

bwplotka:
cc @s-urbaniak


pkg/store/storepb/rules_custom.go
```diff
@@ -0,0 +1,132 @@
...
+		r := &Alert{}
+		if err := json.Unmarshal(entry, r); err != nil {
+			return errors.Wrapf(err, "rule: alerting rule unmarshal: %v", string(entry))
+		}
+
+		m.Result = &Rule_Alert{Alert: r}
+	case "":
+		return errors.Errorf("rule: no type field provided: %v", string(entry))
+	default:
+		return errors.Errorf("rule: unknown type field provided %s; %v", decider.Type, string(entry))
```
s-urbaniak: a bummer that this isn't supported natively, but this looks actually quite ok!

s-urbaniak:
CI complains something about `go.mod` :thinking: 
else LGTM :tada:

s-urbaniak:
i think the `github.com/gogo/status` dependency in `go.mod` is not necessary.

pkg/store/storepb/rules_custom.go
bwplotka: I guess it would work for int not string field out of the box

bwplotka: Wait, I thought we are talking about enums. Then you probably mean https://github.com/gogo/protobuf/issues/623? I think there will be never native support due to custom `type` thing in JSON (unless it will become a standard).

PR #2252, opened 2020-03-10
query: add strict mode flag

GiedriusS:

## Changes

Adds `--store.strict-mode` flag to Thanos Query as agreed on https://thanos.io/proposals/202001_thanos_query_health_handling.md/. The changes seem straightforward:
* Refactored some places where we refer to "healthy" stores - now they are not necessarily healthy but active stores to which we will send the queries;
* Refactored to separate specified store addrs into two sets - dynamically specified and statically. This should make the flow more straightforward i.e. no changes were needed to the resolvers' part;
* `StoreSpec` has a new method `Static()` which tells us if a node has been specified statically. This seems to be the right abstraction layer.

Finally, marks the proposal as "done".

## Verification

Unit tests pass which cover the new functionality.


cmd/thanos/query.go
```diff
@@ -70,6 +70,8 @@ func registerQuery(m map[string]setupFunc, app *kingpin.Application) {
 	replicaLabels := cmd.Flag("query.replica-label", "Labels to treat as a replica indicator along which data is deduplicated. Still you will be able to query without deduplication using 'dedup=false' parameter.").
 		Strings()
 
+	strictMode := cmd.Flag("store.strict-mode", "Enable strict mode which makes Thanos Query always keep statically specified StoreAPIs around.").Default("false").Bool()
```
bwplotka: What it means "keeping something around" here? I think we need to explicitly elaborate this.

cmd/thanos/query.go
```diff
@@ -222,14 +226,20 @@ func runQuery(
 		dns.ResolverType(dnsSDResolver),
 	)
 
+	staticStores, dynamicStores := dns.FilterStaticNodes(storeAddrs...)
```
bwplotka: This is weird that filterstatic nodes function gives you both static and dynamic.. 🤔

pkg/query/storeset_test.go
```diff
@@ -450,7 +453,7 @@ func TestStoreSet_Update(t *testing.T) {
 	discoveredStoreAddr = append(discoveredStoreAddr, stores2.StoreAddresses()...)
 
 	// New stores should be loaded.
-	storeSet.Update(context.Background())
+	storeSet.Update(context.Background(), false)
```
bwplotka: Generally a good rule is to avoid Boolean variables in function. Especially in this case it might be better to just have different function... WDYT?

GiedriusS: I agree that an enum such as `QuerierEvictionPolicy` with values like `StrictMode` and `LaxMode` would make things cleaner here. However, you've mentioned that you'd like to see another function for this. Could you elaborate on that?

cmd/thanos/query.go
GiedriusS: I agree. We should probably rename it to `FilterNodes` or something. WDYT?

cmd/thanos/query.go
GiedriusS: What about:
```
Enable strict mode which makes Thanos Query always keep statically specified StoreAPIs even if we had failed to check their health. This is useful when you have a caching layer on top and want to get partial responses if one of your nodes goes down that you know for sure that it needs to be up. 
```
?

pkg/query/storeset_test.go
GiedriusS: @bwplotka ping.

GiedriusS: @bwplotka ping.

GiedriusS:
> Thanks all good as implemented as in design heart
> 
> However, I just realized that there might be a better UX even if we would allow instead of strict-mode flag adding a new `--store-strict` next to `--store` so you can fine granularly add certain stores as strict. I definitely like this more... It looks like with your implementation it's the matter of few tweaks to make this happen.
> 
> WDYT @GiedriusS ?
> 
> Sorry for this, but I literally did not think about this idea when we were reviewing proposal (:

It's ok but I hope that you understand where the frustration might come from. Without the other discussions that we've had, I can only see marginal UX improvement where the HELP message would be close to the place where the StoreAPIs are defined. I think it's debatable whether it will make sense to have __some__ statically specified StoreAPIs with Cortex's `query-frontend` on top *without* the strict mode. With TLS (`--grpc-client-tls-secure`, https://github.com/thanos-io/thanos/issues/977) it's a totally different case and that's where it would make more sense.

All in all, I will remake this PR with your suggestion and update the proposal. I really want everyone to be happy with this and other things that we do hence why I have made that proposal in the first place.

PR #2253, opened 2020-03-10
*: Fix misuse of pkg/errors.Errorf and error directive

kakkoyun:
`%w` directive is only valid for recently introduces go native error wrapping. `pkg/errors` uses `%v`.
This PR fixes discovered issues in the code base.

## Verification

* `make test-local`
* `make test-e2e`

### DIFF
--- a/pkg/prober/http_test.go
+++ b/pkg/prober/http_test.go
@@ -72,7 +72,7 @@ func TestHTTPProberMuxRegistering(t *testing.T) {
 
 	var g run.Group
 	g.Add(func() error {
-		return errors.Errorf("serve probes %w", http.Serve(l, mux))
+		return errors.Errorf("serve probes %v", http.Serve(l, mux))
 	}, func(err error) {
 		t.Fatalf("server failed: %v", err)
 	})
--- a/pkg/prober/http_test.go
+++ b/pkg/prober/http_test.go
@@ -30,7 +30,7 @@ func TestHTTPProberReadinessInitialState(t *testing.T) {
 }
 
 func TestHTTPProberHealthyStatusSetting(t *testing.T) {
-	testError := errors.Errorf("test error")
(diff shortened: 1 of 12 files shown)


pkg/replicate/replicater.go
```diff
@@ -171,7 +171,7 @@ func RunReplicate(
 		level.Info(logger).Log("msg", "running replication attempt")
 
 		if err := newReplicationScheme(logger, metrics, blockFilter, fromBkt, toBkt, reg).execute(ctx); err != nil {
-			return errors.Errorf("replication execute: %w", err)
+			return errors.Errorf("replication execute: %v", err)
```
GiedriusS: Why don't we replace these with `errors.Wrap`?

pkg/prober/http_test.go
```diff
@@ -72,7 +72,7 @@ func TestHTTPProberMuxRegistering(t *testing.T) {
 
 	var g run.Group
 	g.Add(func() error {
-		return errors.Errorf("serve probes %w", http.Serve(l, mux))
+		return errors.Errorf("serve probes %v", http.Serve(l, mux))
```
GiedriusS: This seems like the only legit usage of `errors.Errorf` in these changed lines.

pkg/replicate/replicater.go
kakkoyun: Actually, yes. This approach would be better. I haven't dug that into what they do while I was changing those. I'm gonna fix them.

bwplotka: Yes, Wrap would be better ((:

kakkoyun:
@GiedriusS I've fixed misuse of `errors.Errorf` for the existing changes. Also traversed the code base and fixed other misuses that caught by my eyes.

I'd be glad if you could take another look at it.

Issue #2248, opened 2020-03-10
cmd/thanos: Refactor run*(...) functions.


bwplotka:
Hi :wave: 

Right now all our components can be run as a single commands. We do this via those ugly `run` functions with a sometimes a large number of arguments (one per each flag). e.g. https://github.com/thanos-io/thanos/blob/master/cmd/thanos/query.go#L132 It would be nice to fix this while preserving flags. The short fix would be to create config structs for each command that will be populated by flags, and then pass it to `run` functions?

AC:
* No more functions with 20+ arguments in Thanos.

bwplotka:
Wonder if this helps https://github.com/peterbourgon/ff

bwplotka:
@s-urbaniak as you were complaining about this :smiling_imp:

PhilipGough:
@bwplotka I don't mind taking a stab at this if thats ok

bwplotka:
Sure @PhilipGough :heart: 

Let us know here @PhilipGough what would be idea of end look .e.g. how you plan to achieve this. (:

PhilipGough:
@bwplotka I did some work on the `sidecar` component only to get some feedback. See https://github.com/thanos-io/thanos/pull/2267

Basically, the idea I had was to decompose into smaller structs with related config and compose the config for the component from these.

Config structs are built by passing all flags into a func. The benefit here is that we wouldn't need to duplicate this info in each component, the downside (I haven't checked if its required) is the lack of ability to customise defaults in each component. Should the defaults always be the same?

I took a look at the package you linked and while it would probably work, would require a change from kingpin lib to flag package in stdlib. Maybe this is worth considering but I would have thought unlikely? 

I didn't see that the kingpin library had a way to populate struct fields directly during parse unfortunately

bwplotka:
Almost done (:

soniasingla:
@bwplotka i would like to play with this bug 😺

PhilipGough:
@soniasingla I have some completed already but feel free to take any of the others. I am working through `query` at the moment, `compact` and `sidecar` are done.

prmsrswt:
Still valid

bwplotka:
As @MalloZup pointed, some commands are still to be done 🤗

MalloZup:
ack

bwplotka:
Maybe @idoqo, you want help here? Sounds like it's defined and we need this badly 🤗

bwplotka:
[...]

PR #2267, opened 2020-03-13
Refactor of commands and flag parsing

PhilipGough:
Introduce config structs for sidecar component.

## Changes

As per #2248  - this change parses flags into structs before passing into the sidecar run func

### DIFF
--- a/cmd/thanos/config.go
+++ b/cmd/thanos/config.go
@@ -1,0 +1,127 @@
+// Copyright (c) The Thanos Authors.
+// Licensed under the Apache License 2.0.
+
+package main
+
+import (
+	"net/url"
+	"time"
+
+	"github.com/prometheus/common/model"
+	"gopkg.in/alecthomas/kingpin.v2"
+)
+
+type grpcConfig struct {
+	bindAddress    string
+	gracePeriod    model.Duration
+	tlsSrvCert     string
+	tlsSrvKey      string
+	tlsSrvClientCA string
+}
+
+func (gc *grpcConfig) registerFlag(cmd *kingpin.CmdClause) *grpcConfig {
+	cmd.Flag("grpc-address",
+		"Listen ip:port address for gRPC endpoints (StoreAPI). Make sure this address is routable from other components.").
+		Default("0.0.0.0:10901").StringVar(&gc.bindAddress)
+	cmd.Flag("grpc-grace-period",
+		"Time to wait after an interrupt received for GRPC Server.").
+		Default("2m").SetValue(&gc.gracePeriod)
+	cmd.Flag("grpc-server-tls-cert",
+		"TLS Certificate for gRPC server, leave blank to disable TLS").
+		Default("").StringVar(&gc.tlsSrvCert)
+	cmd.Flag("grpc-server-tls-key",
+		"TLS Key for the gRPC server, leave blank to disable TLS").
+		Default("").StringVar(&gc.tlsSrvKey)
+	cmd.Flag("grpc-server-tls-client-ca",
+		"TLS CA to verify clients against. If no client CA is specified, there is no client verification on server side. (tls.NoClientCert)").
+		Default("").StringVar(&gc.tlsSrvClientCA)
+	return gc
+}
+
+type httpConfig struct {
+	bindAddress string
+	gracePeriod model.Duration
+}
+
+func (hc *httpConfig) registerFlag(cmd *kingpin.CmdClause) *httpConfig {
+	cmd.Flag("http-address",
+		"Listen host:port for HTTP endpoints.").
+		Default("0.0.0.0:10902").StringVar(&hc.bindAddress)
+	cmd.Flag("http-grace-period",
+		"Time to wait after an interrupt received for HTTP Server.").
+		Default("2m").SetValue(&hc.gracePeriod)
+	return hc
+}
+
+type prometheusConfig struct {
+	url          *url.URL
+	readyTimeout time.Duration
+}
+
+func (pc *prometheusConfig) registerFlag(cmd *kingpin.CmdClause) *prometheusConfig {
+	cmd.Flag("prometheus.url",
+		"URL at which to reach Prometheus's API. For better performance use local network.").
+		Default("http://localhost:9090").URLVar(&pc.url)
+	cmd.Flag("prometheus.ready_timeout",
+		"Maximum time to wait for the Prometheus instance to start up").
+		Default("10m").DurationVar(&pc.readyTimeout)
+	return pc
+}
+
+type connConfig struct {
+	maxIdleConns        int
+	maxIdleConnsPerHost int
+}
+
+func (cc *connConfig) registerFlag(cmd *kingpin.CmdClause) *connConfig {
+	cmd.Flag("receive.connection-pool-size",
+		"Controls the http MaxIdleConns. Default is 0, which is unlimited").
+		IntVar(&cc.maxIdleConns)
+	cmd.Flag("receive.connection-pool-size-per-host",
+		"Controls the http MaxIdleConnsPerHost").
+		Default("100").IntVar(&cc.maxIdleConnsPerHost)
+	return cc
+}
+
+type tsdbConfig struct {
+	path string
+}
+
+func (tc *tsdbConfig) registerFlag(cmd *kingpin.CmdClause) *tsdbConfig {
+	cmd.Flag("tsdb.path", "Data directory of TSDB.").Default("./data").StringVar(&tc.path)
+	return tc
+}
+
+type reloaderConfig struct {
+	confFile        string
+	envVarConfFile  string
+	ruleDirectories []string
+}
+
+func (rc *reloaderConfig) registerFlag(cmd *kingpin.CmdClause) *reloaderConfig {
+	cmd.Flag("reloader.config-file",
+		"Config file watched by the reloader.").
+		Default("").StringVar(&rc.confFile)
+	cmd.Flag("reloader.config-envsubst-file",
+		"Output file for environment variable substituted config file.").
+		Default("").StringVar(&rc.envVarConfFile)
+	cmd.Flag("reloader.rule-dir",
+		"Rule directories for the reloader to refresh (repeated field).").
+		StringsVar(&rc.ruleDirectories)
+	return rc
+}
+
+type shipperConfig struct {
+	uploadCompacted bool
+	ignoreBlockSize bool
+}
+
+func (sc *shipperConfig) registerFlag(cmd *kingpin.CmdClause) *shipperConfig {
+	cmd.Flag("shipper.upload-compacted",
+		"If true sidecar will try to upload compacted blocks as well. Useful for migration purposes. Works only if compaction is disabled on Prometheus. Do it once and then disable the flag when done.").
+		Default("false").BoolVar(&sc.uploadCompacted)
+	cmd.Flag("shipper.ignore-unequal-block-size",
+		"If true sidecar will not require prometheus min and max block size flags to be set to the same value. Only use this if you want to keep long retention and compaction enabled on your Prometheus instance, as in the worst case it can result in ~2h data loss for your Thanos bucket storage.").
+		Default("false").Hidden().BoolVar(&sc.ignoreBlockSize)
+	return sc
+}
--- a/cmd/thanos/flags.go
+++ b/cmd/thanos/flags.go
@@ -7,9 +7,8 @@ import (
 	"fmt"
 	"strings"
 
-	"github.com/thanos-io/thanos/pkg/extflag"
-
 	"github.com/prometheus/common/model"
+	"github.com/thanos-io/thanos/pkg/extflag"
 	"gopkg.in/alecthomas/kingpin.v2"
 )
 
--- a/cmd/thanos/sidecar.go
+++ b/cmd/thanos/sidecar.go
@@ -44,47 +44,17 @@ import (
 
 func registerSidecar(m map[string]setupFunc, app *kingpin.Application) {
 	cmd := app.Command(component.Sidecar.String(), "sidecar for Prometheus server")
-
-	httpBindAddr, httpGracePeriod := regHTTPFlags(cmd)
-	grpcBindAddr, grpcGracePeriod, grpcCert, grpcKey, grpcClientCA := regGRPCFlags(cmd)
-
-	promURL := cmd.Flag("prometheus.url", "URL at which to reach Prometheus's API. For better performance use local network.").
-		Default("http://localhost:9090").URL()
-
-	promReadyTimeout := cmd.Flag("prometheus.ready_timeout", "Maximum time to wait for the Prometheus instance to start up").
-		Default("10m").Duration()
-
-	connectionPoolSize := cmd.Flag("receive.connection-pool-size", "Controls the http MaxIdleConns. Default is 0, which is unlimited").Int()
-	connectionPoolSizePerHost := cmd.Flag("receive.connection-pool-size-per-host", "Controls the http MaxIdleConnsPerHost").Default("100").Int()
-
-	dataDir := cmd.Flag("tsdb.path", "Data directory of TSDB.").
-		Default("./data").String()
-
-	reloaderCfgFile := cmd.Flag("reloader.config-file", "Config file watched by the reloader.").
-		Default("").String()
-
-	reloaderCfgOutputFile := cmd.Flag("reloader.config-envsubst-file", "Output file for environment variable substituted config file.").
-		Default("").String()
-
-	reloaderRuleDirs := cmd.Flag("reloader.rule-dir", "Rule directories for the reloader to refresh (repeated field).").Strings()
-
-	objStoreConfig := regCommonObjStoreFlags(cmd, "", false)
-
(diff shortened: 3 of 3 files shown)


cmd/thanos/sidecar.go
```diff
@@ -44,100 +43,53 @@ import (
...
 			component.Sidecar,
 			*minTime,
-			*connectionPoolSize,
-			*connectionPoolSizePerHost,
 		)
 	}
 }
 
 func runSidecar(
+	conf sidecarComponentConfig,
```
kakkoyun: I  believe it'd be better to put conf as the last argument when initializing a component. If you check the code base, you'd see injected dependencies come first.

cmd/thanos/config.go
```diff
@@ -0,0 +1,60 @@
...
+	"net/url"
+	"time"
+
+	"github.com/prometheus/common/model"
+	"github.com/thanos-io/thanos/pkg/extflag"
+	thanosmodel "github.com/thanos-io/thanos/pkg/model"
+)
+
+type grpcConfig struct {
+	bindAddress    *string
```
kakkoyun: I think when we should have the values in the config rather than pointers. It'd be cleaner in that way, otherwise, we'd end up lots of value extactions.
We can leave pointers if they're optional.

bwplotka:
Started: https://github.com/bwplotka/flagarize

Hopefully can finish something this weekend :muscle:  :heart:

PhilipGough:
@kakkoyun, @bwplotka - thanks for the feedback. I'll be away for a few days so likely the middle of next week before I get back to take a look at this properly.

@bwplotka linked library looks promising. I can look at integrating that if you think it is cleaner to define all the details as struct tags?

bwplotka:
Which exactly link you mean?

PhilipGough:
@bwplotka sorry, this lib https://github.com/octago/sflags

bwplotka:
Yes, I researched it a bit and I don't like it really. It is too generic
and will not work with our custom flag structs. That's why I started
https://github.com/bwplotka/flagarize. Should be ready when you are back!
(:

bwplotka:
I released v0.9.0 of https://github.com/bwplotka/flagarize... should we try to adopt it? 

I already refactored Prometheus for it: https://github.com/prometheus/prometheus/pull/7026

GiedriusS:
> I released v0.9.0 of https://github.com/bwplotka/flagarize... should we try to adopt it?
> 
> I already refactored Prometheus for it: [prometheus/prometheus#7026](https://github.com/prometheus/prometheus/pull/7026)

Thanks for the blog post - the idea itself looks good to me. Is this ready for review? Seems like the CI doesn't like this.

bwplotka:
I believe this is not. Philip wanted to move to flagarize.

PhilipGough:
@GiedriusS , @bwplotka hey, no not ready for review. I just updated one of the components for feedback on how it would look. I'll introduce flagarize and updated the other components asap

bwplotka:
Any update? (:

cmd/thanos/config.go
```diff
@@ -0,0 +1,61 @@
...
+
+type sidecarComponentConfig struct {
+	grpc       grpcConfig
+	http       httpConfig
+	prometheus prometheusConfig
+	tsdb       tsdbConfig
+	connection connectionConfig
+	reloader   reloaderConfig
+	shipper    shipperConfig
+	objStore   *extflag.PathOrContent
```
bwplotka, line 60: This needs some work or we should use `flagarize` PathOrContent instead (and remove our flag)

cmd/thanos/sidecar.go
```diff
@@ -45,70 +46,31 @@ import (
...
-
-	ignoreBlockSize := cmd.Flag("shipper.ignore-unequal-block-size", "If true sidecar will not require prometheus min and max block size flags to be set to the same value. Only use this if you want to keep long retention and compaction enabled on your Prometheus instance, as in the worst case it can result in ~2h data loss for your Thanos bucket storage.").Default("false").Hidden().Bool()
-
 	minTime := thanosmodel.TimeOrDuration(cmd.Flag("min-time", "Start of time range limit to serve. Thanos sidecar will serve only metrics, which happened later than this value. Option can be a constant time in RFC3339 format or time duration relative to current time, such as -1d or 2h45m. Valid duration units are ms, s, m, h, d, w, y.").
 		Default("0000-01-01T00:00:00Z"))
 
 	m[component.Sidecar.String()] = func(g *run.Group, logger log.Logger, reg *prometheus.Registry, tracer opentracing.Tracer, _ <-chan struct{}, _ bool) error {
```
bwplotka, line 52: I think we can now totally remove this item. Instead of registration we can invoke run directly I think :+1:

cmd/thanos/config.go
```diff
@@ -0,0 +1,61 @@
...
+import (
+	"net/url"
+	"time"
+
+	"github.com/bwplotka/flagarize"
+	"github.com/prometheus/common/model"
+)
+
+type grpcConfig struct {
+	BindAddress    string         `flagarize:"name=Grpc-address|help=Listen ip:port address for gRPC endpoints (StoreAPI). Must be routable from other components.|default=0.0.0.0:10901"`
```
brancz, line 15: I'm not a huge fan of the struct tags in this way. We're throwing away all the type safety and language characteristics that we could previously use with kingpin. Why can't we capture the information kept in the struct tags here in real structs?

PhilipGough: @brancz Thats the approach I took in the initial commit https://github.com/thanos-io/thanos/pull/2267/commits/2562b321fbd8d7dd6073f65299272c67d824c1af

I just integrated the library after the first review as suggested but am happy to revert if thats the consensus

brancz: I think I would personally prefer that approach.

PhilipGough: @brancz - I can let yourself and @bwplotka battle that one out in that case :D

bwplotka: I can see lack of type safety, but the advantage is huge - we have finally a way to tie struct filed directly, in the same readable place to a flag. In the same way we can do at some point with `yaml` or json parsing if we choose to have config based configuration. 

Also we literally don't need type safety here, because we always check on our CI the issue on the very start by doing `thanos --help` on every command. (make docs). So overall it does not matter if the error on registering flags is in compile-time or runtime. That's why, while I love type safety, here I am able to sacrifice is for a very good reason.

bwplotka: If you still disagree it would be nice to have opinions from other maintainers as well e.g: @squat @kakkoyun @povilasv @GiedriusS

bwplotka: Overall yea, let's discuss and ensure we all agree here. I can see this being "bit surprising" approach (:

povilasv: I'm also not a big fan of struct tags, but I think we can work around it by having good linter in place, which would complain if it can't parse the tag and explain whats wrong.

pracucci: _My very personal perspective:_

To my understanding, the plan is to introduce a linter to do the compiler job just because we want to specify CLI flags as struct tags instead of code. Why do we want it? Because "we have finally a way to tie struct filed directly, in the same readable place to a flag".

As you probably know, in Cortex we basically achieve the same result attaching a `RegisterFlags()` function to each config struct and we recursively call `RegisterFlags()` on sub-config structs. Despite there's more code involved, I personally find it a safer approach (and avoids to have to spend further time writing a linter).

Example:
https://github.com/cortexproject/cortex/blob/9245c3120d7b16117aaffcb383a6347d002475d1/pkg/compactor/compactor.go#L52-L61

bwplotka: Awesome, thanks for the feedback. Looks like flagarize is not the best option here. I can totally agree with your points. As always there is some power in type safety and more important: autocompletion. (: 

@PhilipGough sorry for taking this back and forth and thanks for giving us the opportunity to see how flagarize would work in Thanos (:

To me looks like we settled on just:

```go
type ...Config struct {
    httpAddress string
}
c := ...Config{}
cmd.Flag( "name", "help").Default().StringVar(&c.httpAddress)
``` 

It's almost similar to what Cortex has except one thing: I am pretty sure we should never define flags outside main package.

BTW,  I only wished @brancz would have the same opinion for other languages and patterns... ;p

povilasv: I've sort of like flagarize, as we would have only one place where we define the config, instead of flags & struct.
The linter thing doesn't have to happen now, so the PR & approach to me looks good.

brancz: I actually quite like the pattern in Cortex.

bwplotka: Let's discuss it more I feel like we have 3 options already (: 

1. Flagarize
2. Struct with typed flag registration in main
3.  Struct with typed flag registration in each component.

I believe flag should never be part of package. We can then limit the surfaces we should care  for each package. If suddenly add flags we add much more things we need to care for in terms of complexity, compatibility etc on top of what we already have with metrics and logs... Imagine e.g reloader package is used in Thanos and X projects (e.g part of prometheus-operator is using it). If we would add flag registration to reloader flag we introduce huge pain in renaming, what should be default should be hidden or not? Quite a pain ): 

So for me no 2 wins, we could think about flagarize more, but looks like 2 is good step in this direction we all are comfortable to take IMO (:

brancz: I'm happy with that as a first step.

pracucci: > I believe flag should never be part of package

That's what happen with flagarize too. Being coupled with the config struct, whenever you move it to a package flags definition moves along it.

> 2. Struct with typed flag registration in main

Even if this option misses the point of "we have finally a way to tie struct filed directly, in the same readable place to a flag", I still see it valuable to move config to `struct` 👍

bwplotka: > That's what happen with flagarize too

Well that's the point we should never use even flagarize outside of main (:

povilasv: I have a bit different opinion, one thing having multiple structs / config options in packages gives us that you can add build tags which could be compiled for embedded stuff. 

For example I'm using s3 and I don't need all that azure stuff and it's dependencies. This would make binary size way smaller, etc. 

One example of this pattern is node exporter actually does command line flags in the different packages and it's fine:

https://github.com/prometheus/node_exporter/blob/master/collector/systemd_linux.go#L41-L48

https://github.com/prometheus/node_exporter/blob/master/collector/filesystem_linux.go#L38-L41

bwplotka: Interesting. @povilasv you are right but we can already do that as we hide all the plugins with the config pathOrContent yaml (:

cmd/thanos/config.go
```diff
@@ -0,0 +1,63 @@
...
+	envVarConfFile  string
+	ruleDirectories []string
+}
+
+type shipperConfig struct {
+	uploadCompacted bool
+	ignoreBlockSize bool
+}
+
+type sidecarComponentConfig struct {
```
bwplotka, line 53: Ideally all of those would be in each command file (:

cmd/thanos/flags.go
```diff
@@ -73,3 +75,99 @@ func regSelectorRelabelFlags(cmd *kingpin.CmdClause) *extflag.PathOrContent {
...
+			Strings(),
+	}
+}
+
+func parseFlagsToShipperConf(cmd *kingpin.CmdClause) shipperConfig {
+	return shipperConfig{
+		uploadCompacted: *cmd.Flag("shipper.upload-compacted",
+			"If true sidecar will try to upload compacted blocks as well. Useful for migration purposes. Works only if compaction is disabled on Prometheus. Do it once and then disable the flag when done.").
+			Default("false").Bool(),
+		ignoreBlockSize: *cmd.Flag("shipper.ignore-unequal-block-size",
```
bwplotka, line 159: So to make it work we need opposite: We need `BoolVar(&ignoreBlockSize)`.

How it works is that you register variables here first. Kingpin either returns address of the variable that the value will be AFTER parsing or it allows us to tell kingpin where is the address (using `*Var()` methods). Let's do the later so our config struct will fill (:

cmd/thanos/flags.go
```diff
@@ -73,3 +75,99 @@ func regSelectorRelabelFlags(cmd *kingpin.CmdClause) *extflag.PathOrContent {
...
+		envVarConfFile: *cmd.Flag("reloader.config-envsubst-file",
+			"Output file for environment variable substituted config file.").
+			Default("").String(),
+		ruleDirectories: *cmd.Flag("reloader.rule-dir",
+			"Rule directories for the reloader to refresh (repeated field).").
+			Strings(),
+	}
+}
+
+func parseFlagsToShipperConf(cmd *kingpin.CmdClause) shipperConfig {
```
bwplotka, line 154: We can follow Cortex way of doing thing and have this as just `RegisterFlag` method on shipperConfig struct directly.


```suggestion
func (c *shipperConfig) RegisterFlag(cmd *kingpin.CmdClause) {
```

cmd/thanos/sidecar.go
```diff
@@ -457,3 +398,39 @@ func (s *promMetadata) Timestamps() (mint int64, maxt int64) {
...
+	tsdb         tsdbConfig
+	connection   connectionConfig
+	reloader     reloaderConfig
+	shipper      shipperConfig
+	objStore     extflag.PathOrContent
+	limitMinTime thanosmodel.TimeOrDurationValue
+}
+
+func newSidecarConfig(cmd *kingpin.CmdClause) *sidecarConfig {
+	var grpcConf grpcConfig
```
bwplotka, line 415: We don't need any of this var definition, we can use directly sidecar config (: 

I think what would look better is just RegisterFlag method for sidecar.


Then we can create sidecar struct and do `RegisterFlags`. Then in that method we can run all RegisterFlags of each field for this struct.  Those method has to have pointer receivers to make a difference.

bwplotka: It's bit repetitive, but good enough (:

cmd/thanos/sidecar.go
```diff
@@ -458,3 +400,29 @@ func (s *promMetadata) Timestamps() (mint int64, maxt int64) {
...
+func (sc *sidecarConfig) registerFlag(cmd *kingpin.CmdClause) *sidecarConfig {
+	sc.http.registerFlag(cmd)
+	sc.grpc.registerFlag(cmd)
+	sc.prometheus.registerFlag(cmd)
+	sc.connection.registerFlag(cmd)
+	sc.tsdb.registerFlag(cmd)
+	sc.reloader.registerFlag(cmd)
+	sc.objStore = *regCommonObjStoreFlags(cmd, "", false)
+	sc.shipper.registerFlag(cmd)
+	sc.limitMinTime = *thanosmodel.TimeOrDuration(cmd.Flag("min-time", "Start of time range limit to serve. Thanos sidecar will serve only metrics, which happened later than this value. Option can be a constant time in RFC3339 format or time duration relative to current time, such as -1d or 2h45m. Valid duration units are ms, s, m, h, d, w, y.").
```
PhilipGough, line 425: @bwplotka are you happy with how this and the one above it are being parsed at this stage?

PhilipGough: actually I see `limitMinTime` isn't being set correctly yet

cmd/thanos/config.go
```diff
@@ -0,0 +1,127 @@
...
+	cmd.Flag("prometheus.url",
+		"URL at which to reach Prometheus's API. For better performance use local network.").
+		Default("http://localhost:9090").URLVar(&pc.url)
+	cmd.Flag("prometheus.ready_timeout",
+		"Maximum time to wait for the Prometheus instance to start up").
+		Default("10m").DurationVar(&pc.readyTimeout)
+	return pc
+}
+
+type connectionConfig struct {
```
bwplotka, line 71:
```suggestion
type connConfig struct {
```

cmd/thanos/sidecar.go
```diff
@@ -458,3 +400,29 @@ func (s *promMetadata) Timestamps() (mint int64, maxt int64) {
...
+	prometheus   prometheusConfig
+	connection   connectionConfig
+	tsdb         tsdbConfig
+	reloader     reloaderConfig
+	objStore     extflag.PathOrContent
+	shipper      shipperConfig
+	limitMinTime thanosmodel.TimeOrDurationValue
+}
+
+func (sc *sidecarConfig) registerFlag(cmd *kingpin.CmdClause) *sidecarConfig {
```
bwplotka, line 416: Looking forward for some tool for this, but let's do in separate PR (:

bwplotka:
So much cleaner :heart: 

@brancz @kakkoyun @squat @GiedriusS looks good? (:

cmd/thanos/sidecar.go
```diff
@@ -44,112 +44,52 @@ import (
...
 			component.Sidecar,
-			*minTime,
-			*connectionPoolSize,
-			*connectionPoolSizePerHost,
 		)
 	}
 }
 
 func runSidecar(
+	conf sidecarConfig,
```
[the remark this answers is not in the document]
PhilipGough: Yep, I had actually fixed this previously but missed it when I reverted back. will fix

bwplotka:
:muscle: @PhilipGough Thanks!

PR #2268, opened 2020-03-13
pkg/store/cache/inmemory.go: Use maxInt instead of math.MaxInt64

pstibrany:
Thanos uses `math.MaxInt64` constant for `int` parameter, which fails when building linux/arm builds. Found via vendoring Thanos in Cortex, and vendoring Cortex in Loki, which does linux/arm builds.

I've run `GOOS=linux GOARCH=arm go build ./cmd/thanos/` with this fix. Without the fix, it fails with

```
$ GOOS=linux GOARCH=arm go build ./cmd/thanos/
# github.com/thanos-io/thanos/pkg/store/cache
pkg/store/cache/inmemory.go:164:22: constant 9223372036854775807 overflows int
```

### DIFF
--- a/pkg/store/cache/inmemory.go
+++ b/pkg/store/cache/inmemory.go
@@ -5,7 +5,6 @@ package storecache
 
 import (
 	"context"
-	"math"
 	"reflect"
 	"sync"
 	"unsafe"
@@ -27,6 +26,8 @@ var (
 		MaxItemSize: 125 * 1024 * 1024,
 	}
 )
+
+const maxInt = int(^uint(0) >> 1)
 
 type InMemoryIndexCache struct {
 	mtx sync.Mutex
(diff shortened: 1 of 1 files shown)


daixiang0:
Check https://github.com/thanos-io/thanos/blob/master/.promu.yml#L18-L22, now we do not support 32it OS.
@bwplotka WDYT? Is any plan to support 32bit OS?

brancz:
As this is equivalent on 64bit binaries I see no reason why not to do this. To me this looks good but I agree this should probably be a larger decision as I doubt this is the only place we do an assumption nor will it be the last, so we should have consensus now.

cc @thanos-io/thanos-maintainers wdyt?

bwplotka:
Yes, let's bring this to our Team sync, but also would be nice as a topic for some Thanos community meeting! 

We essentially need to know if there is any need for 32bit.

Looks like we kind of....have to if Loki indirectly imports Thanos (: 

Let's merge anyway to unblock you @pstibrany

bwplotka:
Also yes, fixes like this are awesome - we might want some helper for this. Using such thing everywhere and setting 32bit cross-build for testing purposes might be a low effort and no impact.

pstibrany:
Thank you Bartek for fast merge!

PR #2271, opened 2020-03-13
bucket: handle instances where no blocks are loaded

### LINKED ISSUE #2260: bucket: thanos.blocks is null
**Thanos bucket version used**: v0.11 & master

**Object Storage Provider**: S3

**What happened**: 

~~We had an issue with overlapping data in one of our blocks. Bucket web worked fine while this data existed for some time. However, after running compactor (and removing a few blocks that it complained about), Bucket web no longer loads.~~ Edit, there was an issue with my object store config.

When trying to load bucket web, we see an error in the console that `thanos.blocks is null`. And therefore (of course) there's no way to `.sort()`.

* Bucket web should display a helpful error message

Handling this could be pretty simple (see example), but I have no idea what error message would be most apt.

```
# bucket.js

function draw() {
    if (thanos.refreshedAt == "0001-01-01T00:00:00Z") {
        thanos.err = "Synchronizing blocks from remote storage";
    }
    else if (!thanos.blocks)
    {
        thanos.err = "No blocks are loaded";
    }
    ...
```

Or some other more elegant solution, any suggestions?

* Create Thanos instance
* Load bucket web interface using an empty bucket

MacroPower:
Closes #2260

## Changes

Current situation is explained in #2260

* If the thanos bucket component loads 0 blocks without errors, pass an empty array rather than nil.
* In the bucket web UI, display a message that 0 blocks are loaded.

## Verification

When bucket web has zero blocks:

`caller=bucket.go:481 msg="downloaded blocks meta.json" num=0`

The following is displayed with no JS errors:

image

If anyone has any better ideas of how to handle this, please let me know. Thanks!


pkg/ui/static/js/bucket.js
```diff
@@ -7,6 +7,9 @@ function draw() {
     if (thanos.refreshedAt == "0001-01-01T00:00:00Z") {
         thanos.err = "Synchronizing blocks from remote storage";
     }
+    else if (thanos.blocks.length === 0) {
```
GiedriusS: We should check here also if `thanos.err` is not null because otherwise, we will hide the "real" error here because our backend gives us an empty array when an error happens.

cmd/thanos/bucket.go
```diff
@@ -472,6 +472,8 @@ func download(ctx context.Context, logger log.Logger, bkt objstore.Bucket, fetch
 		return nil, err
 	}
 
+	blocks = []metadata.Meta{}
```
GiedriusS: Also, if we are doing this then converting this and `err` into "ordinary" variables seems like a nice refactoring that you could do since you'll revise this PR either way.

```suggestion
	blocks := []metadata.Meta{}
```

And then remove it from the function's definition. :P Would look cleaner, right?

pkg/ui/static/js/bucket.js
MacroPower: @GiedriusS You mean if `thanos.err` *is* null? e.g. `!thanos.err && thanos.blocks.length == 0`

cmd/thanos/bucket.go
MacroPower: Yes, completely. I will do this.

pkg/ui/static/js/bucket.js
GiedriusS: Yes, that's what I mean :+1:

MacroPower:
Doing some testing right now. If all goes well I will remove the WIP tag and re-request review.

MacroPower:
@GiedriusS Done!

GiedriusS:
> @GiedriusS Done!

Are you sure you've done that on the newest version? Still shows that there's a conflict :man_shrugging:

GiedriusS:
Thanks!

PR #2272, opened 2020-03-14
Sharding: document supported relabel action and add store gateway backgroud

### LINKED ISSUE #2259: store gw, compactor: Document supported actions for relabel config
Even though [compactor](https://thanos.io/components/compact.md/#flags) and [store gateway](https://thanos.io/components/store.md/#flags) provides a flag for relabeling configuration (`--selector.relabel-config`) and it says in the documentation that it behaves like Prometheus (https://prometheus.io/docs/prometheus/latest/configuration/configuration/#relabel_config), it actually doesn't utilize label mutation actions, like `replace`, `labeldrop` etc. It's only used to keep or drop a block. 

https://github.com/thanos-io/thanos/blob/8a6746298ec39ccb40108e5517535e944d3920d3/pkg/block/fetcher.go#L400-L415

Documentation needs to be more clear about this. 
_Even better,  raise an error when one of the inefective rules specified._

cc @bwplotka 

**What happened**:

Specified a relabel config in Store and Compact, to drop a certain external label from blocks, and it didn't do anything. 

I have expected Thanos to raise an error or log if it fails to execute a relabel action.
(Or actually do it, however apparently this is a design choice, so error would be good.)

* Specify `--selector.relabel-config` flag in Compactor, Store that uses a `labeldrop` action.
* Execute a query with a corresponding selector.

daixiang0:

Fix #2259

### DIFF
--- a/docs/sharding.md
+++ b/docs/sharding.md
@@ -17,10 +17,17 @@ This means that with time you might want to scale out certain components e.g:
 
 * Compactor: Larger number of objects does not matter much, however compactor has to scale (CPU, network, disk, memory) with number of Sources pushing blocks to the object storage.
 
+* Store Gateway: Queries against store gateway which are touching large number of Sources might be expensive, so it has to scale up with number of Sources if we assume those queries.
+    * Orthogonally we did not advertise any labels on Store Gateway's Info. This means that querier was not able to do any pre-filtering, so all store gateways in system are always touched for each query.
+
 # Relabelling
 
 Similar to [promtail](https://github.com/grafana/loki/blob/master/docs/promtail.md#scrape-configs) this config will follow native
 [Prometheus relabel-config](https://prometheus.io/docs/prometheus/latest/configuration/configuration/#relabel_config) syntax.
+
+Now, thanos only support following relabel action:
+* keep
+* drop
 
 The relabel config defines filtering process done on **every** synchronization with object storage.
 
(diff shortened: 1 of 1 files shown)


docs/sharding.md
```diff
@@ -17,11 +17,18 @@ This means that with time you might want to scale out certain components e.g:
...
+    * Orthogonally we did not advertise any labels on Store Gateway's Info. This means that querier was not able to do any pre-filtering, so all store gateways in system are always touched for each query.
+
 # Relabelling
 
 Similar to [promtail](https://github.com/grafana/loki/blob/master/docs/promtail.md#scrape-configs) this config will follow native
 [Prometheus relabel-config](https://prometheus.io/docs/prometheus/latest/configuration/configuration/#relabel_config) syntax.
 
+Now, thanos only support following relabel action:
+* keep
+* drop
```
bwplotka: We also support `hashmod` but it's tricky:

* `hashmod(externa labels)` is ok for all components
* `hashmod (__block_id)` is only allowed for store gateway. We should actually block this in code as well. Also we can link example to our setup here:  https://github.com/observatorium/configuration/blob/master/environments/openshift/manifests/observatorium-template.yaml#L1511

docs/sharding.md
```diff
@@ -17,11 +17,21 @@ This means that with time you might want to scale out certain components e.g:
...
 # Relabelling
 
 Similar to [promtail](https://github.com/grafana/loki/blob/master/docs/promtail.md#scrape-configs) this config will follow native
 [Prometheus relabel-config](https://prometheus.io/docs/prometheus/latest/configuration/configuration/#relabel_config) syntax.
 
+Now, thanos only support following relabel actions:
+* keep
+* drop
+* hashmod
+    * `externa labels` for all components
```
kakkoyun: @daixiang0 I guess you've got a typo here. This should be `external labels`.

bwplotka: Good catch, missed it (:

daixiang0: Oh i missed it.

daixiang0: Fix as #2273

PR #2274, opened 2020-03-16
fix: Ignore the OS-X Trash

kushthedude:

## Changes

- Ignoring the OS-X Trash which can be randomly generated using the project


kushthedude:
@bwplotka Please have a look here!

bwplotka:
Thanks (:

PR #2275, opened 2020-03-16
fix: Discrepancies in go.sum after setting up project

kushthedude:
Signed-off-by: kushthedude <kushthedude@gmail.com>

## Changes

- There were discrepancies in the go.sum after setting up a project on macOS Catalina. I have pushed the changes for the same to fix it up.


kushthedude:
@bwplotka Please review it !

bwplotka:
@kushthedude can you stop pinging me personally? There are more maintainers than just me, plus I already have the notification from this PR, so don't worry - if I will have time I will take a look. (:

bwplotka:
>  they are not checked by any osx vendor test

@squat How do you know those are for osx vendor test?

bwplotka:
If that's true we need long term fix indeed.

daixiang0:
I think it is better to update go.sum and go.mod in container, since it is a clean env.

kushthedude:
> If that's true we need long term fix indeed.

What can be the possible way to handle the following ?

squat:
I suspect they are macos deps somehow because our vendor test is passing in ci and the op wrote that this happens on macos Catalina. But it could also be simply due to a Golang version mismatch. @kushthedude, what golang version are you running locally?

kushthedude:
> I suspect they are macos deps somehow because our vendor test is passing in ci and the op wrote that this happens on macos Catalina. But it could also be simply due to a Golang version mismatch. @kushthedude, what golang version are you running locally?


go1.13.7 darwin/amd64

PR #2284, opened 2020-03-18
Replace objstore.Exists function calls with bkt.Exists

khyatisoneji:
Signed-off-by: khyatisoneji <khyatisoneji5@gmail.com>

## Changes

* Since Bucket Interface already has Exists function, it is better to use `bkt.Exists` directly rather than using `objstore.Exists`. 
This should also address the CI pipeline failures seen here: https://circleci.com/gh/thanos-io/thanos/7875


bwplotka:
oh forgot about my GIF promise:

Issue #2281, opened 2020-03-17
rule: query parameter is not encoded properly in v0.11.0


tobiaszheller:
**Thanos, Prometheus and Golang version used**:
Thanos v0.11.0

**What happened**:

After upgrading thanos from v0.10.0 to v.0.11.0 thanos rule has stopped working.
I am using following setting for query:
```
--query=envoy-internal:80/thanos-query
```
It was working fine before asking for url `http://envoy-internal:80/thanos-query/api/v1/query`.
After upgrading to v0.11.0 query url with after hostname is not encoded properly:
`http://envoy-internal:80%2Fthanos-query/api/v1/query`.

**What you expected to happen**:

Encode url with path after hostname in valid way.

**How to reproduce it (as minimally and precisely as possible)**:

```
	scheme := "http"
	addr := "localhost:80/thanos"
	prefix := ""
	base := &url.URL{
		Scheme: scheme,
		Host:   addr,
		Path:   path.Join("/", prefix),
	}
	fmt.Println(base.String())
```
This is ☝️ more or less current implementation of parsing query URL which results in following output: `http://localhost:80%2Fthanos/`

I can prepare PR with fix.

bwplotka:
Valid, help wanted, whoever has time (: Thanks for report! 

cc @simonpasquier @pgier maybe?

bwplotka:
> I can prepare PR with fix.

Just noticed. Yes please @tobiaszheller fix would be nice

tobiaszheller:
@bwplotka I have take a closer looks and it seems that it will require bigger changes: https://github.com/thanos-io/thanos/blob/5cb3c8758f07fa65ad6a0e97ea69732b14136adb/pkg/http/http.go#L133-L144
EndpointsConfig component expects slice of static addresses but only single PathPrefix.

I am not sure if I can modify that component without break compatibility. I saw that it also used in unmarshaling from config file, right?

simonpasquier:
what about changing this part of the code and create one `query.Config` instance per `queryAddrs` item?

https://github.com/thanos-io/thanos/blob/2d7b214376d5ff2c1b322cecc9b96ed1127dddd7/cmd/thanos/rule.go#L298-L320

Something like:

```go
 for _, addr := range queryAddrs { 
 	if addr == "" { 
 		return errors.New("static querier address cannot be empty") 
 	} 
        // TODO: parse addr to extract path.
	queryCfg = append(queryCfg,
		query.Config{
[...]

PR #2288, opened 2020-03-19
rule: fix query addr parsing

tobiaszheller:

## Changes

- Fixes issue #2281 bug in ruler with parsing query addr with path prefix
Note: Right now `querySDFiles` will be added as new element to `query.Config` slice as suggested in https://github.com/thanos-io/thanos/issues/2281#issuecomment-600546052.

## Verification

I have added unit test for parsing config, however I haven't tested it live. Is there and easy way of doing it?

### DIFF
--- a/cmd/thanos/rule.go
+++ b/cmd/thanos/rule.go
@@ -288,14 +288,13 @@ func runRule(
 	metrics := newRuleMetrics(reg)
 
 	var queryCfg []query.Config
+	var err error
 	if len(queryConfigYAML) > 0 {
-		var err error
 		queryCfg, err = query.LoadConfigs(queryConfigYAML)
 		if err != nil {
 			return err
 		}
 	} else {
-		var err error
 		queryCfg, err = query.BuildQueryConfig(queryAddrs)
 		if err != nil {
 			return err
--- a/cmd/thanos/rule.go
+++ b/cmd/thanos/rule.go
@@ -295,10 +295,10 @@ func runRule(
 			return err
 		}
 	} else {
-		for _, addr := range queryAddrs {
-			if addr == "" {
-				return errors.New("static querier address cannot be empty")
-			}
+		var err error
+		queryCfg, err = query.BuildQueryConfig(queryAddrs)
+		if err != nil {
+			return err
 		}
 
 		// Build the query configuration from the legacy query flags.
@@ -308,16 +308,15 @@ func runRule(
 				Files:           querySDFiles,
 				RefreshInterval: model.Duration(querySDInterval),
 			})
-		}
-		queryCfg = append(queryCfg,
-			query.Config{
-				EndpointsConfig: http_util.EndpointsConfig{
-					Scheme:          "http",
-					StaticAddresses: queryAddrs,
-					FileSDConfigs:   fileSDConfigs,
+			queryCfg = append(queryCfg,
+				query.Config{
+					EndpointsConfig: http_util.EndpointsConfig{
+						Scheme:        "http",
+						FileSDConfigs: fileSDConfigs,
+					},
 				},
-			},
-		)
+			)
+		}
 	}
 
 	queryProvider := dns.NewProvider(
--- a/pkg/query/config.go
+++ b/pkg/query/config.go
@@ -4,8 +4,12 @@
 package query
 
 import (
+	"fmt"
+	"net/url"
+
 	"gopkg.in/yaml.v2"
 
+	"github.com/pkg/errors"
 	http_util "github.com/thanos-io/thanos/pkg/http"
 )
 
@@ -38,4 +42,26 @@ func LoadConfigs(confYAML []byte) ([]Config, error) {
 		return nil, err
 	}
 	return queryCfg, nil
+}
+
+// BuildQueryConfig initializes and returns an Query client configuration from a static address.
+func BuildQueryConfig(queryAddrs []string) ([]Config, error) {
+	configs := make([]Config, 0, len(queryAddrs))
+	for _, addr := range queryAddrs {
+		if addr == "" {
+			return nil, errors.New("static querier address cannot be empty")
+		}
+		u, err := url.Parse(fmt.Sprintf("http://%s", addr))
+		if err != nil {
+			return nil, errors.Wrapf(err, "failed to parse addr %q", addr)
+		}
+		configs = append(configs, Config{
+			EndpointsConfig: http_util.EndpointsConfig{
+				Scheme:          u.Scheme,
+				StaticAddresses: []string{u.Host},
+				PathPrefix:      u.Path,
+			},
+		})
+	}
+	return configs, nil
 }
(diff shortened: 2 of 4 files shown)


tobiaszheller:
@bwplotka is this image automatically built on docker somehow? Or how can I test it live?

bwplotka:
No, you can build on your own: `make docker && make docker-push DOCKER_IMAGE_REPO=<your-repo> DOCKER_IMAGE_TAG=<wanted tag>`

We build only for releases and once you merge to master.

cmd/thanos/rule.go
```diff
@@ -295,10 +295,10 @@ func runRule(
 			return err
 		}
 	} else {
-		for _, addr := range queryAddrs {
-			if addr == "" {
-				return errors.New("static querier address cannot be empty")
-			}
+		var err error
```
bwplotka: let's change returned `) error {` to `) (err error) {` so you don't need this and on line 292

pkg/query/config.go
```diff
@@ -39,3 +43,25 @@ func LoadConfigs(confYAML []byte) ([]Config, error) {
 	}
 	return queryCfg, nil
 }
+
+// BuildQueryConfig initializes and returns an Query client configuration from a static address.
```
bwplotka:
```suggestion
// BuildQueryConfig returns a query client configuration from a static address.
```

Essentially I don't think we initialize anything (:

pkg/query/config.go
```diff
@@ -39,3 +43,25 @@ func LoadConfigs(confYAML []byte) ([]Config, error) {
...
+func BuildQueryConfig(queryAddrs []string) ([]Config, error) {
+	configs := make([]Config, 0, len(queryAddrs))
+	for _, addr := range queryAddrs {
+		if addr == "" {
+			return nil, errors.New("static querier address cannot be empty")
+		}
+		u, err := url.Parse(fmt.Sprintf("http://%s", addr))
```
bwplotka: hmmm what if someone will want HTTPS?

bwplotka: or essentially what if someone will pass with `http://query.com?`

tobiaszheller: I have implemented it in a way how it worked in v0.10.
But maybe you are right, that it should also support passing scheme. I will try to think about something and will get back to you when it's ready.

bwplotka: Yes, I am not saying what was there was the best (: That's why we always look for improvement (:

cmd/thanos/rule.go
tobiaszheller: I have move `var err error` next to definition of `var queryCfg []query.Config`.

pkg/query/config.go
tobiaszheller: right, fixed

pkg/query/config.go
tobiaszheller: I went with checking if `://` is present. If yes, schema is parse. If not, http is added as schema.
I don't have an idea how to do it more clever, apparently it is not that easy looking at google results.

pkg/query/config_test.go
```diff
@@ -0,0 +1,85 @@
...
+
+func TestBuildQueryConfig(t *testing.T) {
+	for _, tc := range []struct {
+		desc      string
+		addresses []string
+		err       bool
+		expected  []Config
+	}{
+		{
+			desc:      "signe addr without path",
```
simonpasquier: s/signe/single/?

pkg/query/config_test.go
```diff
@@ -0,0 +1,85 @@
...
+					EndpointsConfig: http.EndpointsConfig{
+						StaticAddresses: []string{"localhost:9094"},
+						Scheme:          "http",
+						PathPrefix:      "/prefix",
+					},
+				},
+			},
+		},
+		{
+			desc:      "signe addr with path and http scheme",
```
simonpasquier: ditto

pkg/query/config_test.go
```diff
@@ -0,0 +1,85 @@
...
+			addresses: []string{"http://localhost:9093"},
+			expected: []Config{{
+				EndpointsConfig: http.EndpointsConfig{
+					StaticAddresses: []string{"localhost:9093"},
+					Scheme:          "http",
+				},
+			}},
+		},
+		{
+			desc:      "signe addr with path and https scheme",
```
simonpasquier: ditto

pkg/query/config_test.go
```diff
@@ -0,0 +1,85 @@
...
+					Scheme:          "https",
+				},
+			}},
+		},
+		{
+			desc:      "invalid addr",
+			addresses: []string{"this is not a valid addr"},
+			err:       true,
+		},
+	} {
```
simonpasquier: you can add a case with for the empty string.

pkg/query/config.go
```diff
@@ -39,3 +44,29 @@ func LoadConfigs(confYAML []byte) ([]Config, error) {
...
+	configs := make([]Config, 0, len(queryAddrs))
+	for _, addr := range queryAddrs {
+		if addr == "" {
+			return nil, errors.New("static querier address cannot be empty")
+		}
+		// If addr is missing schema, add http.
+		if !strings.Contains(addr, "://") {
+			addr = fmt.Sprintf("http://%s", addr)
+		}
+		u, err := url.Parse(addr)
```
simonpasquier: we could also verify that the scheme is `http` or `https`? It would catch typos like `ttp://example.com` which is a valid URL. 

@bwplotka WDYT?

pkg/query/config.go
```diff
@@ -39,3 +44,29 @@ func LoadConfigs(confYAML []byte) ([]Config, error) {
...
+
+// BuildQueryConfig returns a query client configuration from a static address.
+func BuildQueryConfig(queryAddrs []string) ([]Config, error) {
+	configs := make([]Config, 0, len(queryAddrs))
+	for _, addr := range queryAddrs {
+		if addr == "" {
+			return nil, errors.New("static querier address cannot be empty")
+		}
+		// If addr is missing schema, add http.
+		if !strings.Contains(addr, "://") {
```
simonpasquier: Rather than checking the presence of `://`, maybe parse the URL a first time here and check whether `u.Scheme` is empty or not?

tobiaszheller: The problem is that if you omit scheme, it will parse your domian as scheme:
https://play.golang.org/p/doagHkCbLZu
Or can you suggestion how it can be done in other way @simonpasquier? From what I have checked on Go issues parse works like that for "invalid url" - the ones with missing sheme.

simonpasquier: You're right, sorry for the misleading comment.

pkg/query/config.go
```diff
@@ -39,3 +44,29 @@ func LoadConfigs(confYAML []byte) ([]Config, error) {
 	}
 	return queryCfg, nil
 }
+
+// BuildQueryConfig returns a query client configuration from a static address.
+func BuildQueryConfig(queryAddrs []string) ([]Config, error) {
+	configs := make([]Config, 0, len(queryAddrs))
+	for _, addr := range queryAddrs {
+		if addr == "" {
+			return nil, errors.New("static querier address cannot be empty")
```
povilasv: consider adding more context here, if we 2nd queryAddr is empty, it would be helpful to tell that 2nd adress is wrong

pkg/query/config.go
tobiaszheller: I have added support for it.

bwplotka: Thanks :+1:

PR #2292, opened 2020-03-19
Enrich Memcached client logs

pracucci:
While investigating some memcached errors I found a couple of additional logs (which I'm submitting in this PR) quite useful:

1. Size of a key on store failure
2. Number of keys (and their first key to see what's fetching) on get multi failure

## Changes

- Enriched Memcached client logs

## Verification

Manual tests.

### DIFF
--- a/pkg/cacheutil/memcached_client.go
+++ b/pkg/cacheutil/memcached_client.go
@@ -263,7 +263,7 @@ func (c *memcachedClient) SetAsync(ctx context.Context, key string, value []byte
 		})
 		if err != nil {
 			c.failures.WithLabelValues(opSet).Inc()
-			level.Warn(c.logger).Log("msg", "failed to store item to memcached", "key", key, "err", err)
+			level.Warn(c.logger).Log("msg", "failed to store item to memcached", "key", key, "size_bytes", len(value), "err", err)
 			return
 		}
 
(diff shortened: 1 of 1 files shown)


pkg/cacheutil/memcached_client.go
```diff
@@ -263,7 +263,7 @@ func (c *memcachedClient) SetAsync(ctx context.Context, key string, value []byte
 		})
 		if err != nil {
 			c.failures.WithLabelValues(opSet).Inc()
-			level.Warn(c.logger).Log("msg", "failed to store item to memcached", "key", key, "err", err)
+			level.Warn(c.logger).Log("msg", "failed to store item to memcached", "key", key, "size_bytes", len(value), "err", err)
```
bwplotka:
```suggestion
			level.Warn(c.logger).Log("msg", "failed to store item to memcached", "key", key, "sizeBytes", len(value), "err", err)
```

pkg/cacheutil/memcached_client.go
```diff
@@ -272,9 +272,13 @@ func (c *memcachedClient) SetAsync(ctx context.Context, key string, value []byte
...
 
 func (c *memcachedClient) GetMulti(ctx context.Context, keys []string) map[string][]byte {
+	if len(keys) == 0 {
+		return nil
+	}
+
 	batches, err := c.getMultiBatched(ctx, keys)
 	if err != nil {
-		level.Warn(c.logger).Log("msg", "failed to fetch items from memcached", "err", err)
+		level.Warn(c.logger).Log("msg", "failed to fetch items from memcached", "num_keys", len(keys), "first_key", keys[0], "err", err)
```
bwplotka:
```suggestion
		level.Warn(c.logger).Log("msg", "failed to fetch items from memcached", "numKeys", len(keys), "firstKey", keys[0], "err", err)
```
Maybe it's silly but we stick to camel Case everywhere in Thanos in logs for keys.. (:

pracucci: Sure! I wasn't aware of that. Now I do :)

PR #2297, opened 2020-03-20
store: added option to reencode and compress postings before storing them to the cache

pstibrany:

## Changes

This PR introduces ability to compress postings before storing them into cache (in memory or shared cache). It reencodes postings to use more efficient encoding (diff + varint) and then uses snappy to compress it even more. It typically reduces original postings size to about 1/10th. New flag is disabled by default.

## Verification

- unit tests for encoding/decoding postings
- manual testing hitting the code paths in `fetchPostings` method, and stepping through it in debugger

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -24,7 +24,7 @@ We use *breaking* word for marking changes that are not backward compatible (rel
 - [#2250](https://github.com/thanos-io/thanos/pull/2250) Compactor: Enable vertical compaction for offline deduplication (Experimental). Uses `--deduplication.replica-label` flag to specify the replica label to deduplicate on (Hidden). Please note that this uses a NAIVE algorithm for merging (no smart replica deduplication, just chaining samples together). This works well for deduplication of blocks with **precisely the same samples** like produced by Receiver replication. We plan to add a smarter algorithm in the following weeks.
 - [#1714](https://github.com/thanos-io/thanos/pull/1714) Run the bucket web UI in the compact component when it is run as a long-lived process.
 - [#2304](https://github.com/thanos-io/thanos/pull/2304) Store: Added `max_item_size` config option to memcached-based index cache. This should be set to the max item size configured in memcached (`-I` flag) in order to not waste network round-trips to cache items larger than the limit configured in memcached.
-- [#2297](https://github.com/thanos-io/thanos/pull/2297) Store Gateway: Add `--experimental.enable-postings-compression` flag to enable reencoding and compressing postings before storing them into cache. Compressed postings take about 10% of the original size.
+- [#2297](https://github.com/thanos-io/thanos/pull/2297) Store Gateway: Add `--experimental.enable-index-cache-postings-compression` flag to enable reencoding and compressing postings before storing them into cache. Compressed postings take about 10% of the original size.
 
 ### Changed
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -24,6 +24,7 @@ We use *breaking* word for marking changes that are not backward compatible (rel
 - [#2250](https://github.com/thanos-io/thanos/pull/2250) Compactor: Enable vertical compaction for offline deduplication (Experimental). Uses `--deduplication.replica-label` flag to specify the replica label to deduplicate on (Hidden). Please note that this uses a NAIVE algorithm for merging (no smart replica deduplication, just chaining samples together). This works well for deduplication of blocks with **precisely the same samples** like produced by Receiver replication. We plan to add a smarter algorithm in the following weeks.
 - [#1714](https://github.com/thanos-io/thanos/pull/1714) Run the bucket web UI in the compact component when it is run as a long-lived process.
 - [#2304](https://github.com/thanos-io/thanos/pull/2304) Store: Added `max_item_size` config option to memcached-based index cache. This should be set to the max item size configured in memcached (`-I` flag) in order to not waste network round-trips to cache items larger than the limit configured in memcached.
+- [#2297](https://github.com/thanos-io/thanos/pull/2297) Store Gateway: Add `--experimental.enable-postings-compression` flag to enable reencoding and compressing postings before storing them into cache. Compressed postings take about 10% of the original size.
 
 ### Changed
 
--- a/cmd/thanos/store.go
+++ b/cmd/thanos/store.go
@@ -81,6 +81,9 @@ func registerStore(m map[string]setupFunc, app *kingpin.Application) {
 	enableIndexHeader := cmd.Flag("experimental.enable-index-header", "If true, Store Gateway will recreate index-header instead of index-cache.json for each block. This will replace index-cache.json permanently once it will be out of experimental stage.").
 		Hidden().Default("false").Bool()
 
+	enablePostingsCompression := cmd.Flag("experimental.enable-postings-compression", "If true, Store Gateway will reencode and compress postings before storing them into cache. Compressed postings take about 10% of the original size.").
+		Hidden().Default("false").Bool()
+
 	consistencyDelay := modelDuration(cmd.Flag("consistency-delay", "Minimum age of all blocks before they are being read. Set it to safe value (e.g 30m) if your object storage is eventually consistent. GCS and S3 are (roughly) strongly consistent.").
 		Default("0s"))
 
@@ -126,6 +129,7 @@ func registerStore(m map[string]setupFunc, app *kingpin.Application) {
 			selectorRelabelConf,
 			*advertiseCompatibilityLabel,
 			*enableIndexHeader,
+			*enablePostingsCompression,
 			time.Duration(*consistencyDelay),
 			time.Duration(*ignoreDeletionMarksDelay),
 		)
@@ -160,6 +164,7 @@ func runStore(
 	selectorRelabelConf *extflag.PathOrContent,
 	advertiseCompatibilityLabel bool,
 	enableIndexHeader bool,
+	enablePostingsCompression bool,
 	consistencyDelay time.Duration,
 	ignoreDeletionMarksDelay time.Duration,
 ) error {
@@ -265,6 +270,7 @@ func runStore(
 		filterConf,
 		advertiseCompatibilityLabel,
 		enableIndexHeader,
+		enablePostingsCompression,
 	)
 	if err != nil {
 		return errors.Wrap(err, "create object storage store")
--- a/cmd/thanos/store.go
+++ b/cmd/thanos/store.go
@@ -81,7 +81,7 @@ func registerStore(m map[string]setupFunc, app *kingpin.Application) {
 	enableIndexHeader := cmd.Flag("experimental.enable-index-header", "If true, Store Gateway will recreate index-header instead of index-cache.json for each block. This will replace index-cache.json permanently once it will be out of experimental stage.").
 		Hidden().Default("false").Bool()
 
-	enablePostingsCompression := cmd.Flag("experimental.enable-postings-compression", "If true, Store Gateway will reencode and compress postings before storing them into cache. Compressed postings take about 10% of the original size.").
+	enablePostingsCompression := cmd.Flag("experimental.enable-index-cache-postings-compression", "If true, Store Gateway will reencode and compress postings before storing them into cache. Compressed postings take about 10% of the original size.").
 		Hidden().Default("false").Bool()
 
 	consistencyDelay := modelDuration(cmd.Flag("consistency-delay", "Minimum age of all blocks before they are being read. Set it to safe value (e.g 30m) if your object storage is eventually consistent. GCS and S3 are (roughly) strongly consistent.").
--- a/pkg/store/bucket.go
+++ b/pkg/store/bucket.go
@@ -223,6 +223,7 @@ type BucketStore struct {
 
 	// Reencode postings using diff+varint+snappy when storing to cache.
 	// This makes them smaller, but takes extra CPU and memory.
+	// When used with in-memory cache, memory usage should decrease overall, thanks to postings being smaller.
 	enablePostingsCompression bool
 }
 
--- a/pkg/store/bucket.go
+++ b/pkg/store/bucket.go
@@ -1608,12 +1608,11 @@ func (r *bucketIndexReader) fetchPostings(keys []labels.Label) ([]index.Postings
 
 				if r.block.enablePostingsCompression {
 					// Reencode postings before storing to cache. If that fails, we store original bytes.
+					// This can only fail, if postings data was somehow corrupted,
+					// and there is nothing we can do about it. It's not worth reporting here.
 					data, err := diffVarintSnappyEncode(newBigEndianPostings(pBytes[4:]))
 					if err == nil {
 						storeData = data
-					} else {
-						// This can only fail, if postings data was somehow corrupted,
-						// and there is nothing we can do about it. It's not worth reporting here.
 					}
 				}
 
--- a/pkg/store/bucket.go
+++ b/pkg/store/bucket.go
@@ -98,14 +98,11 @@ type bucketStoreMetrics struct {
 	queriesLimit          prometheus.Gauge
 	seriesRefetches       prometheus.Counter
 
-	cachedPostingsCompressions             prometheus.Counter
-	cachedPostingsCompressionErrors        prometheus.Counter
-	cachedPostingsOriginalSizeBytes        prometheus.Counter
-	cachedPostingsCompressedSizeBytes      prometheus.Counter
-	cachedPostingsCompressionTimeSeconds   prometheus.Counter
-	cachedPostingsDecompressions           prometheus.Counter
-	cachedPostingsDecompressionErrors      prometheus.Counter
-	cachedPostingsDecompressionTimeSeconds prometheus.Counter
+	cachedPostingsCompressions           *prometheus.CounterVec
+	cachedPostingsCompressionErrors      *prometheus.CounterVec
+	cachedPostingsCompressionTimeSeconds *prometheus.CounterVec
+	cachedPostingsOriginalSizeBytes      prometheus.Counter
+	cachedPostingsCompressedSizeBytes    prometheus.Counter
 }
 
 func newBucketStoreMetrics(reg prometheus.Registerer) *bucketStoreMetrics {
@@ -190,14 +187,18 @@ func newBucketStoreMetrics(reg prometheus.Registerer) *bucketStoreMetrics {
 		Help: fmt.Sprintf("Total number of cases where %v bytes was not enough was to fetch series from index, resulting in refetch.", maxSeriesSize),
 	})
 
-	m.cachedPostingsCompressions = promauto.With(reg).NewCounter(prometheus.CounterOpts{
+	m.cachedPostingsCompressions = promauto.With(reg).NewCounterVec(prometheus.CounterOpts{
 		Name: "thanos_bucket_store_cached_postings_compressions_total",
 		Help: "Number of postings compressions before storing to index cache.",
-	})
-	m.cachedPostingsCompressionErrors = promauto.With(reg).NewCounter(prometheus.CounterOpts{
+	}, []string{"op"})
+	m.cachedPostingsCompressionErrors = promauto.With(reg).NewCounterVec(prometheus.CounterOpts{
 		Name: "thanos_bucket_store_cached_postings_compression_errors_total",
 		Help: "Number of postings compression errors.",
-	})
+	}, []string{"op"})
+	m.cachedPostingsCompressionTimeSeconds = promauto.With(reg).NewCounterVec(prometheus.CounterOpts{
+		Name: "thanos_bucket_store_cached_postings_compression_time_seconds",
+		Help: "Time spent compressing postings before storing them into postings cache.",
+	}, []string{"op"})
 	m.cachedPostingsOriginalSizeBytes = promauto.With(reg).NewCounter(prometheus.CounterOpts{
 		Name: "thanos_bucket_store_cached_postings_original_size_bytes_total",
 		Help: "Original size of postings stored into cache.",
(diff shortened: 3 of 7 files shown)


pkg/store/bucket.go
```diff
@@ -1544,7 +1564,20 @@ func (r *bucketIndexReader) fetchPostings(groups []*postingGroup) error {
...
+				storeData := pBytes
+
+				if r.block.enablePostingsCompression {
+					// Reencode postings before storing to cache. If that fails, we store original bytes.
+					// This can only fail, if postings data was somehow corrupted,
+					// and there is nothing we can do about it. It's not worth reporting here.
+					data, err := diffVarintSnappyEncode(newBigEndianPostings(pBytes[4:]))
+					if err == nil {
+						storeData = data
+					}
```
bwplotka: What if cache is malformed and will give error?

pkg/store/postings_codec.go
```diff
@@ -0,0 +1,139 @@
...
+	encoding_binary "encoding/binary"
+	"fmt"
+
+	"github.com/golang/snappy"
+	"github.com/pkg/errors"
+	"github.com/prometheus/prometheus/tsdb/index"
+)
+
+const (
+	// these headers should not be prefix of each other
```
bwplotka: Please make sure comments are full sentence (I think `make lint` should detect this)

pkg/store/postings_codec.go
```diff
@@ -0,0 +1,139 @@
...
+	codecHeaderRaw    = "diff+varint+raw"
+	codecHeaderSnappy = "diff+varint+snappy"
+)
+
+func isDiffVarintEncodedPostings(input []byte) bool {
+	return bytes.HasPrefix(input, []byte(codecHeaderRaw)) || bytes.HasPrefix(input, []byte(codecHeaderSnappy))
+}
+
+func diffVarintSnappyEncode(p index.Postings) ([]byte, error) {
```
bwplotka: Looks quite shallow to me ):

pkg/store/postings_codec.go
```diff
@@ -0,0 +1,139 @@
...
+	return diffVarintEncode(p, true)
+}
+
+func diffVarintEncode(p index.Postings, useSnappy bool) ([]byte, error) {
+	varintBuf := make([]byte, encoding_binary.MaxVarintLen64)
+
+	buf := bytes.Buffer{}
+
+	// if we're returning raw data, write the header to the buffer, and then return buffer directly
```
bwplotka: comment not a full sentence.

bwplotka: and everywhere in this file.

pkg/store/postings_codec.go
```diff
@@ -0,0 +1,139 @@
...
+
+func isDiffVarintEncodedPostings(input []byte) bool {
+	return bytes.HasPrefix(input, []byte(codecHeaderRaw)) || bytes.HasPrefix(input, []byte(codecHeaderSnappy))
+}
+
+func diffVarintSnappyEncode(p index.Postings) ([]byte, error) {
+	return diffVarintEncode(p, true)
+}
+
+func diffVarintEncode(p index.Postings, useSnappy bool) ([]byte, error) {
```
bwplotka: Can we use https://github.com/prometheus/prometheus/blob/cf76daed2f778e97ea70be1bc65ac5f9c5d6d83f/tsdb/encoding/encoding.go#L31 instead of this function?

pkg/store/postings_codec.go
pstibrany: You mean the comment? Well, I can add some, although the function name is pretty self-explanatory to me ;-)

pkg/store/postings_codec.go
pstibrany: `make lint` failed on me on my mac :-( But I will check errors by CI.

pkg/store/bucket.go
pstibrany: Here we are encoding postings data fetched from remote storage. If that data is somehow corrupted (`newBigEndianPostings` return error, or postings entries are not in increasing order), we will store original data (as fetched from storage) to the cache. This is what code did already before this PR.

When same corrupted data is later used, queries will return with errors.

pstibrany:
I've made it through all lint checks 🎉

pkg/store/postings_codec.go
bwplotka: No, I mean function is to shallow. It does not add anything. Abstractions should be deeper and have smaller interfaces otherwise it's just a noise (IMO).

In this case for boolean argument, IDE gives us hint these days so I think we can kill this function.

pstibrany: Ah, I see what you mean. I didn't want to call `diffVarintEncode` with constant `true` argument from bucket.go, so I created this function. You're right that it doesn't add much. False branch (no snappy) is there mostly for testing and to see if compression adds anything or not.

pkg/store/bucket.go
```diff
@@ -220,6 +220,10 @@ type BucketStore struct {
 	advLabelSets             []storepb.LabelSet
 	enableCompatibilityLabel bool
 	enableIndexHeader        bool
+
+	// Reencode postings using diff+varint+snappy when storing to cache.
+	// This makes them smaller, but takes extra CPU and memory.
```
bwplotka: `memory` is tricky here. Because if you in-mem cache it might reduce overall memory a lot.

pkg/store/bucket.go
```diff
@@ -1472,7 +1483,16 @@ func (r *bucketIndexReader) fetchPostings(groups []*postingGroup) error {
 				r.stats.postingsTouched++
 				r.stats.postingsTouchedSizeSum += len(b)
 
-				_, l, err := r.dec.Postings(b)
+				// Even if this instance is not using compression, there may be compressed
+				// entries in the cache written by other stores.
+				var l index.Postings
```
bwplotka: again `var (` available.

pkg/store/bucket.go
```diff
@@ -1472,7 +1483,16 @@ func (r *bucketIndexReader) fetchPostings(groups []*postingGroup) error {
...
+				var err error
+				if isDiffVarintEncodedPostings(b) {
+					l, err = diffVarintDecode(b)
+				} else {
```
bwplotka: I would love to avoid `else` everywhere possible... We can do this here as well with duplicated `g.Fill`.. up to you (: It's just whenever I see `else` in Go it's 99% not needed.

pkg/store/bucket.go
bwplotka: Fair :+1:

pkg/store/postings_codec.go
```diff
@@ -0,0 +1,134 @@
...
+	"bytes"
+
+	"github.com/golang/snappy"
+	"github.com/pkg/errors"
+	"github.com/prometheus/prometheus/tsdb/encoding"
+	"github.com/prometheus/prometheus/tsdb/index"
+)
+
+const (
+	// These headers should not be a prefix of each other.
```
bwplotka: of each other?

bwplotka: what does it mean?

pkg/store/postings_codec.go
```diff
@@ -0,0 +1,134 @@
...
+	codecHeaderRaw    = "diff+varint+raw"
```
bwplotka:
```suggestion
	codecHeaderRaw    = "diff+varint"
```
?

pkg/store/postings_codec.go
```diff
@@ -0,0 +1,134 @@
...
+// isDiffVarintEncodedPostings returns true, if input looks like it has been encoded by diff+varint(+snappy) codec.
+func isDiffVarintEncodedPostings(input []byte) bool {
+	return bytes.HasPrefix(input, []byte(codecHeaderRaw)) || bytes.HasPrefix(input, []byte(codecHeaderSnappy))
+}
+
+// diffVarintEncode encodes postings into diff+varint representation and optional snappy compression.
+func diffVarintEncode(p index.Postings, useSnappy bool) ([]byte, error) {
+	buf := encoding.Encbuf{}
+
+	// If we're returning raw data, write the header to the buffer, and then return buffer directly.
```
bwplotka: I don't understand this comment really.. maybe placement is not correct? let's remove it, I totally understand this function without this.

pkg/store/postings_codec.go
```diff
@@ -0,0 +1,134 @@
...
+	if !useSnappy {
+		buf.PutString(codecHeaderRaw)
+	}
+
+	prev := uint64(0)
+	for p.Next() {
+		v := p.At()
+		if v < prev {
```
bwplotka: What if v i `0`?

pkg/store/postings_codec.go
```diff
@@ -0,0 +1,134 @@
...
+			// Postings entries must be in increasing order.
```
bwplotka: Not needed comment, it should rather part of error

pkg/store/postings_codec.go
```diff
@@ -0,0 +1,134 @@
...
+			return nil, errors.Errorf("decreasing entry, val: %d, prev: %d", v, prev)
+		}
+
+		buf.PutUvarint64(v - prev)
```
bwplotka: why we do delta? For better compression?

bwplotka: This one is very much needed to be commented then ;p

pkg/store/postings_codec.go
```diff
@@ -0,0 +1,134 @@
...
+		prev = v
+	}
+
```
bwplotka: not needed new line... In the same way we don't do newline between `err := invoke` and `if err != nil`

CHANGELOG.md
```diff
@@ -20,6 +20,7 @@ We use *breaking* word for marking changes that are not backward compatible (rel
 ### Added
 
 - [#2265](https://github.com/thanos-io/thanos/pull/2265) Compactor: Add `--wait-interval` to specify compaction wait interval between consecutive compact runs when `--wait` enabled.
+- [#2297](https://github.com/thanos-io/thanos/pull/2297) Store Gateway: Add `--experimental.enable-postings-compression` flag to enable reencoding and compressing postings before storing them into cache. Compressed postings take about 10% of the original size.
```
bwplotka: This is not the only thing that changed. We changed overall the format of cached postings even with flag is set to false, right?

pkg/store/postings_codec.go
```diff
@@ -0,0 +1,134 @@
...
+		headerLen = len(codecHeaderRaw)
+	case bytes.HasPrefix(input, []byte(codecHeaderSnappy)):
+		headerLen = len(codecHeaderSnappy)
+		compressed = true
+	default:
+		return nil, errors.New("header not found")
+	}
+
+	raw := input[headerLen:]
+	if compressed {
```
bwplotka: just put it in `case bytes.HasPrefix(input, []byte(codecHeaderSnappy)):`? No need for `compressed` bool

pkg/store/postings_codec.go
```diff
@@ -0,0 +1,134 @@
...
+		raw, err = snappy.Decode(nil, raw)
+		if err != nil {
+			return nil, errors.Errorf("snappy decode: %w", err)
+		}
+	}
+
+	return &diffVarintPostings{buf: &encoding.Decbuf{B: raw}}, nil
+}
+
+// Implementation of index.Postings based on diff+varint encoded data.
```
bwplotka: comment should start with struct name

pkg/store/postings_codec.go
```diff
@@ -0,0 +1,134 @@
...
+
+		buf.PutUvarint64(v - prev)
+		prev = v
+	}
+
+	if p.Err() != nil {
+		return nil, p.Err()
+	}
+
+	if !useSnappy {
```
bwplotka: hmmm do we really need `useSnappy`? We always use snappy.

pkg/store/bucket.go
pstibrany: Added extra `// When used with in-memory cache, memory usage should decrease overall, thanks to postings being smaller.` sentence to the comment.

pkg/store/bucket.go
pstibrany: OK :) Personally I find `var (` ugly

pkg/store/bucket.go
pstibrany: > It's just whenever I see else in Go it's 99% not needed.

That's a ... interesting... statement :)

Here we would need to duplicate entire
```
				if err != nil {
					return errors.Wrap(err, "decode postings")
				}
				g.Fill(j, l)
```

block. I feels like extra repetition to me, but whatever works...

pkg/store/postings_codec.go
pstibrany: This is exactly what would not work ;-) Since we compare prefix of the input byte array, HasPrefix(input, "diff+varint") returns true for both snappy and not snappy version. I wanted prefix to be unique.

pkg/store/postings_codec.go
pstibrany: ok

pkg/store/postings_codec.go
pstibrany: Snappy compression on its own doesn't really help, because there is no repetition. Delta makes values small, and Varint is very efficient at encoding small values (values < 128 are encoded as single byte, values < 16384 are encoded as two bytes). It's delta + using varint encoding that reduces size significantly (to about 20% of original), snappy then halves it to 10% of the original.

pstibrany: Well, "diff" (or "delta", perhaps we should rename it) is already a strong indication of what we're doing here :)

CHANGELOG.md
pstibrany: Not really. If flag is set to false, then this store will only write original postings (as found in the index) to the cache. However, if store finds compressed entry in the cache, it will understand it and decompress.

pkg/store/postings_codec.go
pstibrany: By using bool flag, we can avoid `HasPrefix` comparison which is costly. But I think I'll remove non-snappy version completely, as you suggested in other comment.

pkg/store/postings_codec.go
pstibrany: I *believe* that contract of `index.Postings` is that values are increasing [it is not explicit about it, although the file format is and much of the code using postings requires this property]. That implies that `p.At()` can only be 0 as the very first value, which is ok (`prev` is 0 as well at that point, and `v < prev` is not true).

pkg/store/bucket.go
bwplotka: Thanks!

pkg/store/bucket.go
bwplotka: Curious why, but yea - not a blocker (:

pkg/store/bucket.go
bwplotka: > That's a ... interesting... statement :)

Yes, but I can totally prove this (:

bwplotka: Not a blocker here as well, up to you (:

pkg/store/bucket.go
pstibrany: Because of extra indentation. But I've used var block here, as you suggested :)

pkg/store/postings_codec.go
bwplotka: Ok, now is the question why do we need both. Can we only support snappy?

pkg/store/postings_codec.go
bwplotka: fair point. Thanks

pkg/store/postings_codec.go
bwplotka: :+1: Thanks!

bwplotka: Maybe strong indication but what you said here:

> Snappy compression on its own doesn't really help, because there is no repetition. Delta makes values small, and Varint is very efficient at encoding small values (values < 128 are encoded as single byte, values < 16384 are encoded as two bytes). It's delta + using varint encoding that reduces size significantly (to about 20% of original), snappy then halves it to 10% of the original.

Is important and will be forgotten if it's only on GH PR comment.

CHANGELOG.md
bwplotka: Fair (:

pkg/store/postings_codec.go
bwplotka: yes please (:

pkg/store/postings_codec.go
pstibrany: Two headers and comment is now removed.

pkg/store/postings_codec.go
pstibrany: Yes, you're right.

pkg/store/postings_codec.go
pstibrany: Fair enough, I'll put this comment into the file.

pkg/store/postings_codec.go
pstibrany: It's gone now :)

pkg/store/bucket.go
```diff
@@ -1544,7 +1567,20 @@ func (r *bucketIndexReader) fetchPostings(groups []*postingGroup) error {
...
 				groups[p.groupID].Fill(p.keyID, newBigEndianPostings(pBytes[4:]))
-				r.block.indexCache.StorePostings(r.ctx, r.block.meta.ULID, groups[p.groupID].keys[p.keyID], pBytes)
+
+				storeData := pBytes
+
+				if r.block.enablePostingsCompression {
+					// Reencode postings before storing to cache. If that fails, we store original bytes.
+					// This can only fail, if postings data was somehow corrupted,
+					// and there is nothing we can do about it. It's not worth reporting here.
+					data, err := diffVarintSnappyEncode(newBigEndianPostings(pBytes[4:]))
```
pracucci: Have you considered adding a couple of metrics with postings raw size and the compressed one, to measure the compression ratio in the wild?

pkg/store/postings_codec.go
```diff
@@ -0,0 +1,135 @@
...
+// number encoding. On top of that, we apply Snappy compression.
+//
+// On its own, Snappy compressing raw postings doesn't really help, because there is no
+// repetition in raw data. Using diff (delta) between postings entries makes values small,
+// and Varint is very efficient at encoding small values (values < 128 are encoded as
+// single byte, values < 16384 are encoded as two bytes). Diff + varint reduces postings size
+// significantly (to about 20% of original), snappy then halves it to ~10% of the original.
+
+const (
+	codecHeaderSnappy = "dvs" // As in "diff+varint+snappy".
```
bwplotka: :+1:

cmd/thanos/store.go
```diff
@@ -81,6 +81,9 @@ func registerStore(m map[string]setupFunc, app *kingpin.Application) {
 	enableIndexHeader := cmd.Flag("experimental.enable-index-header", "If true, Store Gateway will recreate index-header instead of index-cache.json for each block. This will replace index-cache.json permanently once it will be out of experimental stage.").
 		Hidden().Default("false").Bool()
 
+	enablePostingsCompression := cmd.Flag("experimental.enable-postings-compression", "If true, Store Gateway will reencode and compress postings before storing them into cache. Compressed postings take about 10% of the original size.").
```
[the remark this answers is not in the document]
bwplotka: good point, but since we don't have any other postings storage it's fine for me in smaller form.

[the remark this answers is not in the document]
bwplotka: Also I assume at some point we will enable it by default and remove this flag

pkg/store/bucket.go
```diff
@@ -1544,7 +1567,20 @@ func (r *bucketIndexReader) fetchPostings(groups []*postingGroup) error {
...
+					if err == nil {
```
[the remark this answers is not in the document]
bwplotka: We will know as querying would fail?

pkg/store/bucket.go
bwplotka: Good point

cmd/thanos/store.go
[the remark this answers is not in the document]
pstibrany: Bartek, if you don't mind, I will change it. It will be clearer.

pkg/store/bucket.go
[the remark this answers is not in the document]
pstibrany: That was my thinking here... if it we get error here, it means postings was corrupted, so it will still be corrupted few lines later when it's used, and error will be reported there.

pkg/store/postings_codec.go
```diff
@@ -0,0 +1,135 @@
...
+// and applies snappy compression on the result.
+// Returned byte slice starts with codedHeaderSnappy header.
+func diffVarintSnappyEncode(p index.Postings) ([]byte, error) {
+	buf, err := diffVarintEncodeNoHeader(p)
+	if err != nil {
+		return nil, err
+	}
+
+	// Make result buffer large enough to hold our header and compressed block.
+	result := make([]byte, len(codecHeaderSnappy)+snappy.MaxEncodedLen(len(buf)))
```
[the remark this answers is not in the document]
pstibrany: No, because both `copy` and `snappy.Encode` only checks for length of the supplied buffer, not its capacity.

cmd/thanos/store.go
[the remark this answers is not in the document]
bwplotka: Sure, go for it :+1:

pkg/store/bucket.go
pstibrany: I've added following metrics:
- number of compressions, and compression failures 
- total original size and compressed size
- total compression time

pstibrany: Oh, one more thing... I've moved compression outside of the lock. However, storing data back to cache is still done with lock held, which looks wrong. I haven't fixed that part, because this PR ir not about it, but it looks like it should be moved outside as well. I haven't looked into more details here (like, does it break anything?)

pstibrany: Also added number of decompressions, decompression errors, and total decompression time (snappy part only, we don't time decoding varints and adding values together)

pstibrany:
> Awesome. LGTM! Will merge after a conflict is resolved 👍 👍 👍

Rebased, and also added metrics for compression and decompression (count, errors, time, original vs compressed size).

pkg/store/bucket.go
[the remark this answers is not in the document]
pstibrany: These errors are now tracked via metric.

bwplotka:
Thanks for your hard work @pstibrany ! Sorry for the need for some many rebases ):

pstibrany:
No worries, thanks for your help Bartek!

PR #2304, opened 2020-03-23
Added max_item_size to Memcached client

pracucci:

## Changes

While running the memcached index cache in Cortex we've find out that whenever the item size in larger than the max size allowed by memcached (defaults to 1MB, can increase with the `-I` option) the client will waste resources doing a full round-trip to memcached just to see the `Set()` operation to fail because the item size is larger than the limit.

In this PR I'm proposing to add a new option to memcached client config to allow to set the `max_item_size` so that we skip the set operation at all if the item size is greater.

## Verification

Unit tests.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -23,7 +23,7 @@ We use *breaking* word for marking changes that are not backward compatible (rel
 - [#2265](https://github.com/thanos-io/thanos/pull/2265) Compactor: Add `--wait-interval` to specify compaction wait interval between consecutive compact runs when `--wait` enabled.
 - [#2250](https://github.com/thanos-io/thanos/pull/2250) Compactor: Enable vertical compaction for offline deduplication (Experimental). Uses `--deduplication.replica-label` flag to specify the replica label to deduplicate on (Hidden). Please note that this uses a NAIVE algorithm for merging (no smart replica deduplication, just chaining samples together). This works well for deduplication of blocks with **precisely the same samples** like produced by Receiver replication. We plan to add a smarter algorithm in the following weeks.
 - [#1714](https://github.com/thanos-io/thanos/pull/1714) Run the bucket web UI in the compact component when it is run as a long-lived process.
-- [#2304](https://github.com/thanos-io/thanos/pull/2304) Store: Added `max_item_size` config option to memcached-based index cache. This should be set to the max item size configured in memcached (`-I` flag) in order to waste network round-trips to cache items larger than the limit configured in memcached.
+- [#2304](https://github.com/thanos-io/thanos/pull/2304) Store: Added `max_item_size` config option to memcached-based index cache. This should be set to the max item size configured in memcached (`-I` flag) in order to not waste network round-trips to cache items larger than the limit configured in memcached.
 
 ### Changed
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -23,6 +23,7 @@ We use *breaking* word for marking changes that are not backward compatible (rel
 - [#2265](https://github.com/thanos-io/thanos/pull/2265) Compactor: Add `--wait-interval` to specify compaction wait interval between consecutive compact runs when `--wait` enabled.
 - [#2250](https://github.com/thanos-io/thanos/pull/2250) Compactor: Enable vertical compaction for offline deduplication (Experimental). Uses `--deduplication.replica-label` flag to specify the replica label to deduplicate on (Hidden). Please note that this uses a NAIVE algorithm for merging (no smart replica deduplication, just chaining samples together). This works well for deduplication of blocks with **precisely the same samples** like produced by Receiver replication. We plan to add a smarter algorithm in the following weeks.
 - [#1714](https://github.com/thanos-io/thanos/pull/1714) Run the bucket web UI in the compact component when it is run as a long-lived process.
+- [#2304](https://github.com/thanos-io/thanos/pull/2304) Store: Added `max_item_size` config option to memcached-based index cache. This should be set to the max item size configured in memcached (`-I` flag) in order to waste network round-trips to cache items larger than the limit configured in memcached.
 
 ### Changed
 
--- a/Makefile
+++ b/Makefile
@@ -261,7 +261,7 @@ test-e2e-ci: docker
 	@rm -rf ./test/e2e/e2e_integration_test*
 	@echo ">> running /test/e2e tests."
 	@go clean -testcache
-	@go test -failfast -parallel 1 -timeout 5m -v ./test/e2e/...
+	@go test -failfast -parallel 1 -timeout 10m -v ./test/e2e/...
 
 .PHONY: install-deps
 install-deps: ## Installs dependencies for integration tests. It installs supported versions of Prometheus and alertmanager to test against in integration tests.
--- a/Makefile
+++ b/Makefile
@@ -261,7 +261,7 @@ test-e2e-ci: docker
 	@rm -rf ./test/e2e/e2e_integration_test*
 	@echo ">> running /test/e2e tests."
 	@go clean -testcache
-	@go test -failfast -parallel 1 -timeout 10m -v ./test/e2e/...
+	@go test -failfast -parallel 1 -timeout 5m -v ./test/e2e/...
 
 .PHONY: install-deps
 install-deps: ## Installs dependencies for integration tests. It installs supported versions of Prometheus and alertmanager to test against in integration tests.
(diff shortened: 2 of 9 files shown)


pkg/cacheutil/memcached_client_test.go
```diff
@@ -102,27 +105,62 @@ dns_provider_update_interval: 1s
...
 
-	testutil.Equals(t, 2.0, prom_testutil.ToFloat64(client.operations.WithLabelValues(opSet)))
-	testutil.Equals(t, 0.0, prom_testutil.ToFloat64(client.operations.WithLabelValues(opGetMulti)))
-	testutil.Equals(t, 0.0, prom_testutil.ToFloat64(client.failures.WithLabelValues(opSet)))
+		testutil.Equals(t, 2.0, prom_testutil.ToFloat64(client.operations.WithLabelValues(opSet)))
+		testutil.Equals(t, 1.0, prom_testutil.ToFloat64(client.operations.WithLabelValues(opGetMulti)))
+		testutil.Equals(t, 0.0, prom_testutil.ToFloat64(client.failures.WithLabelValues(opSet)))
+		testutil.Equals(t, 0.0, prom_testutil.ToFloat64(client.skipped.WithLabelValues(opSet, reasonMaxItemSize)))
+	}
+
```
povilasv: Consider splitting these into two tests instead of using blocks?

pkg/store/cache/inmemory.go
```diff
@@ -18,6 +18,8 @@ import (
 	"github.com/prometheus/client_golang/prometheus/promauto"
 	"github.com/prometheus/prometheus/pkg/labels"
 	"gopkg.in/yaml.v2"
+
```
povilasv: remove the space please

pkg/model/units_test.go
```diff
@@ -1,13 +1,14 @@
...
 
-package storecache
+package model
 
 import (
 	"testing"
 
-	"github.com/thanos-io/thanos/pkg/testutil"
 	"gopkg.in/yaml.v2"
+
```
povilasv: remove the blank line please

docs/components/store.md
```diff
@@ -212,6 +212,7 @@ config:
   max_idle_connections: 0
   max_async_concurrency: 0
   max_async_buffer_size: 0
+  max_item_size: 0
```
pstibrany: Would it make sense for this option to be available also for in-memory cache?

pkg/cacheutil/memcached_client.go
```diff
@@ -250,6 +264,12 @@ func (c *memcachedClient) Stop() {
 }
 
 func (c *memcachedClient) SetAsync(ctx context.Context, key string, value []byte, ttl time.Duration) (err error) {
+	// Skip hitting memcached at all if the item is bigger than the max allowed size.
+	if c.config.MaxItemSize > 0 && len(value) > int(c.config.MaxItemSize) {
```
pstibrany: `int(c.config.MaxItemSize)` seems unsafe as it can cut off higher 4 bytes of the value on 32-bit systems. On the other hand, MaxItemSize is very unlikely to be bigger than 2^31, so not worth fixing.

docs/components/store.md
pracucci: Yes, it's already available:
https://github.com/thanos-io/thanos/blob/d89a497d03a3d41781c26cf6f248f214f08bace7/pkg/store/cache/inmemory.go#L57

pkg/cacheutil/memcached_client.go
pracucci: Unless I'm missing anything, a slice in Go is limited to an `int` size, so I'm not sure `len(value)` could be ever greater than `int`. Am I missing anything?

pracucci:
Thanks @povilasv and @pstibrany for your review. I should have addressed your comments. May you take a look, please?

pkg/cacheutil/memcached_client.go
pracucci: From an offline discussion:

> It's c.config.MaxItemSize that can be bigger than 2^31. For example, say that c.config.MaxItemSize = 2^31 + 10, then conversion to int will make it negative number

You're right. Fixed!

cmd/thanos/store.go
```diff
@@ -14,6 +14,9 @@ import (
 	"github.com/pkg/errors"
 	"github.com/prometheus/client_golang/prometheus"
 	"github.com/prometheus/prometheus/pkg/relabel"
+	"gopkg.in/alecthomas/kingpin.v2"
+	yaml "gopkg.in/yaml.v2"
+
```
povilasv: let's remove blank line

pracucci: Done. I would suggest you to add a linter on imports so that it's easier to enforce your preferred styling 😉

povilasv: We do have a ticket for that https://github.com/thanos-io/thanos/issues/984, I think none of the linters supports this atm

pstibrany: Yes, extra new lines are tricky. Golangci-lint can check for imports ordering and grouping at least.

docs/components/store.md
GiedriusS: Maybe let's recommend users to set this to `1MiB` as that is the default memcached's limit instead of disabling it?

pracucci: Would you just change this example config or the default value in the code?

GiedriusS: What do you mean by **or** in your sentence? We could add `omitempty` on `MaxItemSize` & change the value in `defaultMemcachedClientConfig` to `1MiB` plus change it here to be `1MiB`. I think our users would be better off if we would recommend sane values with the default memcached configuration :-)

pracucci: Makes sense to me. I've changed the default from `0` to `1MB` (which is the default in Memcached). What's your take?

Makefile
```diff
@@ -261,7 +261,7 @@ test-e2e-ci: docker
 	@rm -rf ./test/e2e/e2e_integration_test*
 	@echo ">> running /test/e2e tests."
 	@go clean -testcache
-	@go test -failfast -parallel 1 -timeout 5m -v ./test/e2e/...
+	@go test -failfast -parallel 1 -timeout 10m -v ./test/e2e/...
```
GiedriusS: I assume you've talked with @bwplotka privately about this one (https://github.com/thanos-io/thanos/pull/2313). Should be OK, I guess :P

docs/components/store.md
GiedriusS: My take is that this is awesome :+1:

CHANGELOG.md
```diff
@@ -22,6 +22,7 @@ We use *breaking* word for marking changes that are not backward compatible (rel
 
 - [#2265](https://github.com/thanos-io/thanos/pull/2265) Compactor: Add `--wait-interval` to specify compaction wait interval between consecutive compact runs when `--wait` enabled.
 - [#2250](https://github.com/thanos-io/thanos/pull/2250) Compactor: Enable vertical compaction for offline deduplication (Experimental). Uses `--deduplication.replica-label` flag to specify the replica label to deduplicate on (Hidden). Please note that this uses a NAIVE algorithm for merging (no smart replica deduplication, just chaining samples together). This works well for deduplication of blocks with **precisely the same samples** like produced by Receiver replication. We plan to add a smarter algorithm in the following weeks.
+- [#2304](https://github.com/thanos-io/thanos/pull/2304) Store: Added `max_item_size` config option to memcached-based index cache. This should be set to the max item size configured in memcached (`-I` flag) in order to waste network round-trips to cache items larger than the limit configured in memcached.
```
GiedriusS:
```suggestion
- [#2304](https://github.com/thanos-io/thanos/pull/2304) Store: Added `max_item_size` config option to memcached-based index cache. This should be set to the max item size configured in memcached (`-I` flag) in order not to waste network round-trips to cache items larger than the limit configured in memcached.
```

pracucci:
Thanks @GiedriusS for your prompt review. I should have addressed the comments.

bwplotka:
Don't worry about e2e now, we have flaky tests currently. Fixing now: https://github.com/thanos-io/thanos/pull/2313

pkg/store/cache/inmemory.go
```diff
@@ -52,9 +53,9 @@ type InMemoryIndexCache struct {
 // InMemoryIndexCacheConfig holds the in-memory index cache config.
 type InMemoryIndexCacheConfig struct {
 	// MaxSize represents overall maximum number of bytes cache can contain.
-	MaxSize Bytes `yaml:"max_size"`
+	MaxSize model.Bytes `yaml:"max_size"`
```
bwplotka: sorry for being late to the party, but why we are not using the same thing as Prometheus use, so ` units.Base2Bytes`?

pracucci: `model.Bytes` is based on `units.Base2Bytes` but adds YAML marshalling support. Am I missing anything?

bwplotka:
Thanks ! :muscle:

pracucci:
> Thanks !

Thanks to you!

PR #2307, opened 2020-03-24
git: ignore cmd/thanos/thanos

00arthur00:


squat:
hi @00arthur00 thanks for submitting this, however I don't think we really need this gitignore rule. When you run `make build` the binary is placed in the root of the repo and is already covered by a gitignore rule.

00arthur00:
@squat  Thanks. Will close the PR.

Issue #1299, opened 2019-07-02
Decide & apply what time range query without stores should announce


brancz:
This is a follow up from https://github.com/improbable-eng/thanos/pull/1284#discussion_r299038494

If a query component has no stores discovered, it currently announces that it has data from time range `0` to `math.MaxInt64`. @bwplotka mentioned that this should either be all of time, which would be `math.MinInt64` to `math.MaxInt64` or nothing, which could be represented as `math.MaxInt64` to `math.MaxInt64`.

I think either of those two make more sense than what we currently have, but I'm tending towards the latter, to not do unneeded requests to a store.

brancz:
We have yet to decide on this.

bwplotka:
I like `or nothing, which could be represented as math.MaxInt64 to math.MaxInt64` but maybe in form or `[0:0]` It is just more readable for humans... WDYT?

brancz:
Happy with 0, 0. Will do a PR.

PR #2310, opened 2020-03-24
pkg/store: Report no data if no stores discovered

brancz:

## Changes

When no stores are discovered, then timespan from 0 to 0 unixtime is reported.

Closes https://github.com/thanos-io/thanos/issues/1299.

## Verification

n/a

@thanos-io/thanos-maintainers


bwplotka:
Looks like test has to be adjusted (L

Issue #2315, opened 2020-03-24
query: Inconsistent "thanos_status" metric naming


MacroPower:
**Thanos, Prometheus and Golang version used**:

Thanos: 0.11
Promethus: 2.16
Golang: 1.13.6

**What happened**:

I was exploring some of the internal Thanos metrics, and noticed some inconsistencies with the query component.

Each component has a `thanos_status` metric, except query, where it is called `status`.

Results from `http://<query>/metrics`:
```
# HELP status Represents status (0 indicates success, 1 indicates failure) of the component.
# TYPE status gauge
status{check="healthy",component="query"} 1
status{check="ready",component="query"} 1
```

Results from `http://<compact/bucket/sidecar/store>/metrics`:
```
# HELP thanos_status Represents status (0 indicates success, 1 indicates failure) of the component.
# TYPE thanos_status gauge
thanos_status{check="healthy",component="sidecar"} 1
thanos_status{check="ready",component="sidecar"} 1
```

**What you expected to happen**:

Metric naming should be consistent across all components.

MacroPower:
> 0 indicates success, 1 indicates failure

Additionally, this #HELP is the inverse of what it should be.

jojohappy:
Hi, sorry for the inconsistent metrics, we will fix soon.

> > 0 indicates success, 1 indicates failure
> 
> Additionally, this #HELP is the inverse of what it should be.
This HELP message has been changed in this PR #2216 .

PR #2319, opened 2020-03-25
Fixed inconsistent metrics naming and method

jojohappy:
Signed-off-by: jojohappy <sarahdj0917@gmail.com>

fixes: #2315 


kakkoyun:
@bwplotka I guess this was a mistake, it supposed to be `thanos_status`. It's not documented anywhere else. Anyway, this comes from an issue so people use it, better to add an entry to CHANGELOG. WDYT @jojohappy?

bwplotka:
Yea, like `fixed`

PR #2325, opened 2020-03-25
tesutil: Enchanced testutil, refactored for our needs.

bwplotka:
Changed LICENSE as we no longer use version we copied back then.
Most of it was reimplemented.

Why?
* Much richer diff (inspired by testify packages
* Consistent API
* Less indentation.

### DIFF
--- a/pkg/testutil/testutil.go
+++ b/pkg/testutil/testutil.go
@@ -1,24 +1,5 @@
-// The MIT License (MIT)
-
-// Copyright (c) 2014 Ben Johnson
-
-// Permission is hereby granted, free of charge, to any person obtaining a copy
-// of this software and associated documentation files (the "Software"), to deal
-// in the Software without restriction, including without limitation the rights
-// to use, copy, modify, merge, publish, distribute, sublicense, and/or sell
-// copies of the Software, and to permit persons to whom the Software is
-// furnished to do so, subject to the following conditions:
-
-// The above copyright notice and this permission notice shall be included in all
-// copies or substantial portions of the Software.
-
-// THE SOFTWARE IS PROVIDED "AS IS", WITHOUT WARRANTY OF ANY KIND, EXPRESS OR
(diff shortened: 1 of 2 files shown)


bwplotka:
As discussed https://cloud-native.slack.com/archives/CL25937SP/p1585155073025700 we might also want to upstream this to Prometheus.

bwplotka:
done: https://github.com/prometheus/prometheus/pull/7053

pkg/testutil/testutil.go
```diff
@@ -29,50 +10,121 @@ import (
...
+	diff, _ := difflib.GetUnifiedDiffString(difflib.UnifiedDiff{
+		A:        difflib.SplitLines(e),
+		B:        difflib.SplitLines(a),
+		FromFile: "Expected",
+		FromDate: "",
+		ToFile:   "Actual",
+		ToDate:   "",
+		Context:  1,
+	})
+	return "\n\nDiff:\n" + diff
```
GiedriusS: We should probably have:
```suggestion
	return "Diff:\n" + diff
```
And let the caller decide on the exact formatting. Maybe not a problem for us in Thanos but if someone else will start using this then probably it makes sense to remove those two `\n\n` and move them to callers so that the output for any other potential user would not be malformed. What do you think?

bwplotka: Fair

bwplotka:
Merging as I need this, will fix the minor nit in another PR.

pkg/testutil/testutil.go
bwplotka: Actually that's not fair... I don't think it makes sense as if there is no DIFF (because type is not correct or something) we would have unnecessary new lines.

bwplotka: Happy to fix in another PR if decided to, but I don't think this suggestion makes sense @GiedriusS  ):

PR #2337, opened 2020-03-29
query: add --store-strict flag

GiedriusS:
Replaces: https://github.com/thanos-io/thanos/pull/2252.

Add a new flag called `--store-strict` as agreed per
https://thanos.io/proposals/202001_thanos_query_health_handling.md/

I have updated the proposal to reflect the reality.

Third time's the charm, I believe it :-)

Now the flag is called `--store-strict` which only accepts statically
defined nodes. I guess the code is even simpler now.

I have also fixed one small issue where `%w` was used in
`errors.Errorf`. Couldn't compile Thanos locally with Go 1.14 without
this fix.

Tested with unit tests + ad-hoc local testing.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -20,6 +20,7 @@ We use *breaking* word for marking changes that are not backward compatible (rel
 
 ### Added
 
+- [#2252](https://github.com/thanos-io/thanos/pull/2252) Query: add new `--store.strict-mode` flag. More information available [here](/docs/proposals/202001_thanos_query_health_handling.md).
 - [#2265](https://github.com/thanos-io/thanos/pull/2265) Compactor: Add `--wait-interval` to specify compaction wait interval between consecutive compact runs when `--wait` enabled.
 - [#2250](https://github.com/thanos-io/thanos/pull/2250) Compactor: Enable vertical compaction for offline deduplication (Experimental). Uses `--deduplication.replica-label` flag to specify the replica label to deduplicate on (Hidden). Please note that this uses a NAIVE algorithm for merging (no smart replica deduplication, just chaining samples together). This works well for deduplication of blocks with **precisely the same samples** like produced by Receiver replication. We plan to add a smarter algorithm in the following weeks.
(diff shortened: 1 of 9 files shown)


CHANGELOG.md
```diff
@@ -20,6 +20,7 @@ We use *breaking* word for marking changes that are not backward compatible (rel
 
 ### Added
 
+- [#2252](https://github.com/thanos-io/thanos/pull/2252) Query: add new `--store.strict-mode` flag. More information available [here](/docs/proposals/202001_thanos_query_health_handling.md).
```
bwplotka: Looks like this was not updated. (:

GiedriusS: Oops, updated. Thanks for the quick review!

GiedriusS:
> Amazing +1 LGTM, thanks for this.
> 

Thanks for your patience and understanding!

> I just have another idea. What about changing this to... I am joking upside_down_face upside_down_face xD

I was about to write that it's too bad that GitHub doesn't have a troll face emoji but apparently it exists: :trollface: ... yay :smile:  

Hmm... what about renaming it to `--strict-store` since it seems like that this is an attributive adjective (https://www.ucl.ac.uk/internet-grammar/adjectiv/attribut.htm)?

Actually, I double-checked and we have some inconsistency already:
* `--web.route-prefix`
* `--web.external-prefix`
* `--web.prefix-header`

:trollface: 

> 
> Sorry about that. But I truly believe last-minute ideas that I think even aligns with what you wanted from the beginning (!). It is not a bad thing (: But sorry for adding more work on you. And please always challenge my ideas.
> 
> CHANGELOG is not rdy, otherwise let's merge!

Fixed this.

Also, I have tested this on top of the latest `master` with my recent changes and it seems like everything works how it is expected to. I have missed one detail in the comment on my last PR - that `s.stores` is always populated with the nodes specified via this option and that the code passes around the same `map` between functions.

bwplotka:
:tada: 

It is like 100th feature for Thanos 0.12.0 so it will be quite epic release ;p

PR #2339, opened 2020-03-29
ui: fix store never removed from /stores page bug

GiedriusS:
We need to update `LastCheck` only if the error is non-nil. That field
is used in the cleanup function to know when to remove the StoreAPI from
the UI. If we always update it, even if an error has happened, that
means that `--store.unhealthy-timeout` is never respected.
*But* this only affects the UI part.

## Verification

* Local testing;
* Amended unit tests

## Note

Funnily enough, https://github.com/thanos-io/thanos/pull/2337 will have to probably be modified a bit because the UI AFAICT will continuously remove/add nodes specified via `--store-strict` if they go down.


bwplotka:
Thanks! :heart:

PR #2343, opened 2020-03-30
Allow more connection reuse than the default of 2

kubicgruenfeld:

## Changes

Due to https://github.com/golang/go/issues/13801 Connection reuse currently seems not to work as expected. This also got fixed for Prometheus in https://github.com/prometheus/prometheus/pull/3592 and for Cortex in https://github.com/cortexproject/cortex/pull/2268, since Cortex is now using the implementation of object store of thanos i may hit this behaviour.

I am not sure how to test this, if someone could help to test, this would be great.

## Verification

See above

### DIFF
--- a/pkg/objstore/s3/s3.go
+++ b/pkg/objstore/s3/s3.go
@@ -149,6 +149,7 @@ func NewBucketWithConfig(logger log.Logger, config Config, component string) (*B
 			DualStack: true,
 		}).DialContext,
 		MaxIdleConns:          100,
+		MaxIdleConnsPerHost:   100,
 		IdleConnTimeout:       time.Duration(config.HTTPConfig.IdleConnTimeout),
 		TLSHandshakeTimeout:   10 * time.Second,
 		ExpectContinueTimeout: 1 * time.Second,


pkg/exthttp/transport.go
```diff
@@ -20,6 +20,7 @@ func NewTransport() *http.Transport {
 		}).DialContext,
 		ForceAttemptHTTP2:     true,
 		MaxIdleConns:          100,
+		MaxIdleConnsPerHost:   100,
```
GiedriusS: If I am not mistaken this one is already covered by the `connectionPoolSizePerHost` option in `cmd/thanos/sidecar.go` that later gets passed down to `NewPrometheusStore()` :P So this should probably be removed.

pkg/objstore/s3/s3.go
```diff
@@ -149,6 +149,7 @@ func NewBucketWithConfig(logger log.Logger, config Config, component string) (*B
 			DualStack: true,
 		}).DialContext,
 		MaxIdleConns:          100,
+		MaxIdleConnsPerHost:   100,
```
GiedriusS: Since the default in our code-base is already this is probably OK. We should probably opt to have this value hard-coded, not sure if anyone will want to change it. Also, I wonder how this will affect performance. There are most likely way too many variables to have objective data.

kubicgruenfeld:
If i got this right, the values in pkg/exthttp/transport.go are overwritten by cmd/thanos/sidecar.go. So i would just remove it from transport.go and only left it for pkg/objstore/s3/s3.go. I did not understand how this is related to https://github.com/thanos-io/thanos/pull/1967.

bwplotka:
Ignore, me I though we have connection pool in receive as well, looks like not. 

Well we could keep the exthttp default with 100 idle connection TBH if that's what we expect to use in prod, WDYT?

kubicgruenfeld:
> Ignore, me I though we have connection pool in receive as well, looks like not.
> 
> Well we could keep the exthttp default with 100 idle connection TBH if that's what we expect to use in prod, WDYT?

@GiedriusS Do you agree?

blockloop:
#1966 added those flags to the receive component, but was closed in favor of #1970

GiedriusS:
> > Ignore, me I though we have connection pool in receive as well, looks like not.
> > Well we could keep the exthttp default with 100 idle connection TBH if that's what we expect to use in prod, WDYT?
> 
> @GiedriusS Do you agree?

Yes, IMHO it should be good. I don't have any concrete benchmark numbers for now but `100` doesn't look very big to me and it is definitely an improvement over only `2` connections.  If this ever becomes a bottleneck then we could think of ways to pass down a user-specified number to the S3 remote object storage provider.

So, let's remove the change to `pkg/exthttp/transport.go` and merge this? AFAICT there aren't any other suggestions.

bwplotka:
+1

pkg/exthttp/transport.go
kubicgruenfeld: I removed it from here.

PR #2346, opened 2020-03-30
website: Enabled two scripts to fix Google analytics.

bwplotka:

Verification: Now on F12 network page we can see call to Google Analyitcs.


bwplotka:
With this we can see Google Analytics call:

image

PR #2351, opened 2020-03-31
Added Workfront as adopter

ryaneorth:
* [ x] Change is not relevant to the end user.

## Changes
Added Workfront as an adopter.


ryaneorth:
We started using Thanos and love it! Thank you for the awesome product!

PR #2354, opened 2020-03-31
fetcher: Made metaFetcher go routine safe; Fixed multiple bucket UI +fetcher issues.

bwplotka:
Fixed https://github.com/thanos-io/thanos/issues/2349
Fixed races (we were reusing fetcher by both bucket UI and compaction syncs...
Fixed logging
Added singleflight to ensure we don't synchronize too often.

Signed-off-by: Bartlomiej Plotka <bwplotka@gmail.com>

cc @squat Would be awesome to merge this before release as it's a major race :/ That might interact with what compact sees.

PR #2357, opened 2020-04-01
store & compact: For components that operates on blocks - expose the UI on /loaded-blocks

bwplotka:


bwplotka:
Also fixed comments from https://github.com/thanos-io/thanos/pull/2354

bwplotka:
All works (: 

image


PTAL @squat

metalmatze:
The gist of this is: Store now has the Bucket UI for the blocks it has loaded, correct?

bwplotka:
Yes, in more extended way:

* Store now has the Bucket UI for the blocks it has loaded in `/loaded`
* Compactor has two UIs:
   * `/loaded` for what Compactor is seeing (e.g based on sharding) 
   * `/global` for what overall we have in whatever form in the bucket (:

PR #2364, opened 2020-04-02
mixin: Remove unused jobPrefix field

lilic:

## Changes

Remove jobPrefix from the thanos mixin, as it's not used anywhere.

## Verification

After removing the field from the libs I did a `grep -r 'jobPrefix' mixin/thanos/alerts` to confirm it's really not used anywhere.

---
Not sure if this needs a CHANELOG entry? 🤔


povilasv:
I'll merge once CI finishes, thanks! :1st_place_medal:

PR #2365, opened 2020-04-02
.*: Don't report "object not found" as error for Get/GetRange

pstibrany:

## Changes

Don't treat "object not found" as failures when reporting metrics for Get/GetRange operations.

Recently introduced `IgnoreDeletionMarkFilter` uses Get operation to check if block is marked for deletion, and since many blocks are not, bucket store returns "object not found" error. Since `metricBucket` reports this as failed operation, it looks as if there were many failed operations. But this is expected outcome for Get/GetRange, so I don't think it should be reported as failure.

Alternatively, we can modify `IgnoreDeletionMarkFilter` to check for file first, but that would require two operations.

## Verification

See decreased number of reported failures `thanos_objstore_bucket_operation_failures_total` for `Get` operation.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -19,6 +19,7 @@ We use *breaking* word for marking changes that are not backward compatible (rel
 - [#2254](https://github.com/thanos-io/thanos/pull/2254) Bucket: Fix metrics registered multiple times in bucket replicate
 - [#2271](https://github.com/thanos-io/thanos/pull/2271) Bucket Web: Fixed Issue #2260 bucket passes null when storage is empty
 - [#2339](https://github.com/thanos-io/thanos/pull/2339) Query: fix a bug where `--store.unhealthy-timeout` was never respected
+- [#2365](https://github.com/thanos-io/thanos/pull/2365) .*: don't report "object not found" error as failure via `thanos_objstore_bucket_operation_failures_total` for `get` and `get_range` operations.
 
 ### Added
 
--- a/pkg/objstore/objstore.go
+++ b/pkg/objstore/objstore.go
@@ -276,7 +276,9 @@ func (b *metricBucket) Get(ctx context.Context, name string) (io.ReadCloser, err
 
 	rc, err := b.bkt.Get(ctx, name)
 	if err != nil {
-		b.opsFailures.WithLabelValues(getOp).Inc()
+		if !b.bkt.IsObjNotFoundErr(err) {
+			b.opsFailures.WithLabelValues(getOp).Inc()
+		}
 		return nil, err
 	}
 	return newTimingReadCloser(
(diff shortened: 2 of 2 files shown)


pstibrany:
`check-doc` target failed with 

```
docs/getting-started.md
	ERROR	https://fosdem.org/2019/schedule/event/thanos_transforming_prometheus_to_a_global_scale_in_a_seven_simple_steps/
		Dialing to the given TCP address timed out
```

I believe it's unrelated to my change.

GiedriusS:
> https://fosdem.org/2019/schedule/event/thanos_transforming_prometheus_to_a_global_scale_in_a_seven_simple_steps/

Indeed it seems unrelated. Reran the CI.

pstibrany:
> IMHO it would be cleaner to do this in `deletionmark.go` itself because some code _might_ actually try to `Get` something that the code assumes that some object exists, and that should probably be reflected in the metrics. For example, the user could _still_ run into consistency issues even with all of these safeguards in place but with this `if` it would be impossible to detect such a problem easily.

I've created https://github.com/thanos-io/thanos/pull/2369 that implements alternative fix in `ReadDeletionMark` function. Personally I weakly prefer this PR, but either one is fine. Thank you.

pracucci:
> IMHO it would be cleaner to do this in deletionmark.go itself because some code might actually try to Get something that the code assumes that some object exists, and that should probably be reflected in the metrics.

From my perspective, bucket level metrics should track a failure only when a bucket-related failure occurs. If the object does not exist, it's not a bucket-level error (ie. networking error, storage unavailable, etc...). I personally think this PR is better than #2369.

bwplotka:
Looking.

NotFound is a problem indeed - it's most of the time user error, but for us we used to treat it as a server error if it's unexpected.

bwplotka:
One way of doing it is to always use `Exists` method before.. but that seems like overhead for most of the times ):

pstibrany:
> One way of doing it is to always use `Exists` method before.. but that seems like overhead for most of the times ):

#2369 :)

bwplotka:
or actually you did that version too https://github.com/thanos-io/thanos/pull/2369

Will reach you offline

bwplotka:
I think I like https://github.com/thanos-io/thanos/pull/2369 more, because we will be still notified for cases that not found is unexpected...

Other option is to extend the API a bit and add something like `NewBucketNotFoundIsOk` ;p

bwplotka:
Slack discussion: https://cloud-native.slack.com/archives/CL25937SP/p1585903230146500

@pracucci Re https://github.com/thanos-io/thanos/pull/2365#issuecomment-608260426

You are right, but we still need to have some metric if this happens in unexpectedly. Fourth alternative is another metric (:

bwplotka:
Alternative we agrred offline for: https://github.com/thanos-io/thanos/pull/2370 Let me know if that's make sense :hugs:

pstibrany:
#2370 seems to fix the problem as well. Thanks!

bwplotka:
Thanks to you for starting discussion and ideas :+1:

pstibrany:
Thanks Bartek!

PR #2369, opened 2020-04-02
.*: Verify if deletion marker file exists before using it.

pstibrany:
This is an alternative to #2365. Please see the problem description there. This PR modifies `ReadDeletionMark` function to use `Exists` call before reading deletion marker file. This helps to keep `thanos_objstore_bucket_operation_failures_total` down, since missing deletion marker file is not really a failure.

### DIFF
--- a/pkg/block/metadata/deletionmark.go
+++ b/pkg/block/metadata/deletionmark.go
@@ -48,6 +48,17 @@ type DeletionMark struct {
 func ReadDeletionMark(ctx context.Context, bkt objstore.BucketReader, logger log.Logger, dir string) (*DeletionMark, error) {
 	deletionMarkFile := path.Join(dir, DeletionMarkFilename)
 
+	// BucketReader.Get reports missing file as a failure (via metrics),
+	// but since most blocks don't have this marker file, it skews metrics unnecessarily.
+	// By using exists first, we avoid that.
+	exists, err := bkt.Exists(ctx, deletionMarkFile)
+	if err != nil {
+		return nil, err
+	}
+	if !exists {
+		return nil, ErrorDeletionMarkNotFound
+	}
+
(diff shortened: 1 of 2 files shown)


pkg/block/metadata/deletionmark.go
```diff
@@ -48,6 +48,17 @@ type DeletionMark struct {
 func ReadDeletionMark(ctx context.Context, bkt objstore.BucketReader, logger log.Logger, dir string) (*DeletionMark, error) {
 	deletionMarkFile := path.Join(dir, DeletionMarkFilename)
 
+	// BucketReader.Get reports missing file as a failure (via metrics),
+	// but since most blocks don't have this marker file, it skews metrics unnecessarily.
+	// By using exists first, we avoid that.
+	exists, err := bkt.Exists(ctx, deletionMarkFile)
```
pracucci, line 54: I'm dubious about having to do another API call just to check if the object exists, when we can get the same information from `Get()`. I personally believe #2365 works better ([my take here](https://github.com/thanos-io/thanos/pull/2365#issuecomment-608260426)).

bwplotka:
See discussion here: https://github.com/thanos-io/thanos/pull/2365 and on slack https://cloud-native.slack.com/archives/CL25937SP/p1585903230146500

pkg/block/metadata/deletionmark.go
bwplotka: ack

bwplotka:
Alternative we agrred offline for: https://github.com/thanos-io/thanos/pull/2370 Let me know if that's make sense :hugs:

bwplotka:
Fix merged https://github.com/thanos-io/thanos/pull/2383

bwplotka:
Thanks for help on this :heart:

PR #2370, opened 2020-04-03
objstore: Added WithExpectedErrs which allows to control instrumentation (e.g not increment failures for expected not found).

bwplotka:
This 3rd alternative to below PRs that was discussed offline: https://cloud-native.slack.com/archives/CL25937SP/p1585903230146500

Closes https://github.com/thanos-io/thanos/pull/2365 and https://github.com/thanos-io/thanos/pull/2369

This allows to not wake up on-call in the middle of the night, because of the expected, properly handled case (:

Thanks @pracucci and @pstibrany for starting this work and discussion. We were aware of this for like 12 months, but only with your discussions, we clarified what is expected here :hugs:  Your input was super valuable.

## Changes

Added new interface:

```go
type InstrumentedBucket interface {
	Bucket

	// WithExpectedErrs allows to specify a filter that marks certain errors as expected, so it will not wake up
	// on-call engineer (:
	WithExpectedErrs(IsOpFailureExpectedFunc) Bucket
```

Also: Has to move inmem to objstore for testing.

### DIFF
--- a/pkg/block/fetcher.go
+++ b/pkg/block/fetcher.go
@@ -142,7 +142,7 @@ type MetadataModifier interface {
 type BaseFetcher struct {
 	logger      log.Logger
 	concurrency int
-	bkt         objstore.BucketReader
+	bkt         objstore.InstrumentedBucketReader
 
 	// Optional local directory to cache meta.json files.
 	cacheDir string
@@ -152,7 +152,7 @@ type BaseFetcher struct {
 }
 
 // NewBaseFetcher constructs BaseFetcher.
-func NewBaseFetcher(logger log.Logger, concurrency int, bkt objstore.BucketReader, dir string, reg prometheus.Registerer) (*BaseFetcher, error) {
+func NewBaseFetcher(logger log.Logger, concurrency int, bkt objstore.InstrumentedBucketReader, dir string, reg prometheus.Registerer) (*BaseFetcher, error) {
 	if logger == nil {
 		logger = log.NewNopLogger()
 	}
@@ -180,7 +180,7 @@ func NewBaseFetcher(logger log.Logger, concurrency int, bkt objstore.BucketReade
 }
 
 // NewMetaFetcher returns meta fetcher.
-func NewMetaFetcher(logger log.Logger, concurrency int, bkt objstore.BucketReader, dir string, reg prometheus.Registerer, filters []MetadataFilter, modifiers []MetadataModifier) (*MetaFetcher, error) {
+func NewMetaFetcher(logger log.Logger, concurrency int, bkt objstore.InstrumentedBucketReader, dir string, reg prometheus.Registerer, filters []MetadataFilter, modifiers []MetadataModifier) (*MetaFetcher, error) {
 	b, err := NewBaseFetcher(logger, concurrency, bkt, dir, reg)
 	if err != nil {
 		return nil, err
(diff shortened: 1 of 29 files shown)


pkg/objstore/azure/azure.go
```diff
@@ -64,7 +64,7 @@ func (conf *Config) validate() error {
 	return nil
 }
 
-// NewBucket returns a new Bucket using the provided Azure config.
+// NewInMemBucket returns a new Bucket using the provided Azure config.
```
GiedriusS, line 67: Seems like an unrelated change?

pkg/objstore/client/factory.go
```diff
@@ -40,9 +40,9 @@ type BucketConfig struct {
 	Config interface{} `yaml:"config"`
 }
 
-// NewBucket initializes and returns new object storage clients.
+// NewInMemBucket initializes and returns new object storage clients.
```
GiedriusS, line 43: Seems like an unrelated change?

pkg/objstore/filesystem/filesystem.go
```diff
@@ -43,7 +43,7 @@ func NewBucketFromConfig(conf []byte) (*Bucket, error) {
 	return NewBucket(c.Directory)
 }
 
-// NewBucket returns a new filesystem.Bucket.
+// NewInMemBucket returns a new filesystem.Bucket.
```
GiedriusS, line 46: Ditto.

pkg/objstore/gcs/gcs.go
```diff
@@ -41,7 +41,7 @@ type Bucket struct {
 	closer io.Closer
 }
 
-// NewBucket returns a new Bucket against the given bucket handle.
+// NewInMemBucket returns a new Bucket against the given bucket handle.
```
GiedriusS, line 44: Ditto.

pkg/objstore/oss/oss.go
```diff
@@ -150,7 +150,7 @@ func (b *Bucket) ObjectSize(ctx context.Context, name string) (uint64, error) {
 	return 0, errors.New("content-length header not found")
 }
 
-// NewBucket returns a new Bucket using the provided oss config values.
+// NewInMemBucket returns a new Bucket using the provided oss config values.
```
GiedriusS, line 153: Ditto.

pkg/objstore/azure/azure.go
bwplotka: :man_facepalming:

pkg/block/fetcher.go
```diff
@@ -236,7 +236,7 @@ func (f *BaseFetcher) loadMeta(ctx context.Context, id ulid.ULID) (*metadata.Met
 		}
 	}
 
-	r, err := f.bkt.Get(ctx, metaFile)
+	r, err := f.bkt.ReaderWithExpectedErrs(f.bkt.IsObjNotFoundErr).Get(ctx, metaFile)
```
pracucci, line 239: Is this expected to cover the case of a partially uploaded block (while still uploading)?

bwplotka: Yes

bwplotka:
Exactly this + Giedrius addressed comments was merged in release 0.12 and will be ported to master soon: https://github.com/thanos-io/thanos/commit/c05daeeace272a94fae81fd0c15ff70adab50e10

Issue #2213, opened 2020-03-03
Query latency increased after upgrade Thanos v0.10.1 ->v 0.11.0


IKSIN:
**Thanos, Prometheus and Golang version used**:
Thanos: v0.11.0
Golang: 1.13.8

**Object Storage Provider**:
CEPH (S3)

**What happened**:
Query latency increased

**What you expected to happen**:
Same latency that on v0.10.1

**How to reproduce it (as minimally and precisely as possible)**:

**Full logs to relevant components**:

**Anything else we need to know**:

IKSIN:
<img width="1636" alt="Screenshot 2020-03-03 at 16 51 23" src="https://user-images.githubusercontent.com/2960110/75784162-d9cd8080-5d72-11ea-9939-f37518693b59.png">
<img width="1581" alt="Screenshot 2020-03-03 at 14 44 04" src="https://user-images.githubusercontent.com/2960110/75784175-dc2fda80-5d72-11ea-8df9-2968ea46aef5.png">

bwplotka:
Thanks for the reports.

Let's eliminate the unknown. The major changes to querying was:

* store timeout: you said you are running on ported 0.10.1 + this and it was fine, but are we sure ported commit was actually the same?
* index-header: However if not enabled, it should change nothing. Maybe we actually touched something? Will look.
* Refetch: We added this: https://github.com/thanos-io/thanos/pull/1985 Let's check for log msg: series size exceeded expected size; refetching or `thanos_bucket_store_series_refetches_total`
* Minio client update .. :man_shrugging: Worth to look?

Worth to check:

* If the same queries are running.
* CPU Profile differences

NOTE: You show graph with high latency metric for store GW, but this latency is really depending on Querier (due to our streaming round-robin behavior), so we don't really know if it's store being slow, or querier or yet another Store API :/

bwplotka:
The easiest way would be to go back on master container images and do bi search. Maybe without index-header first?

bwplotka:
Refetches are not the problem: https://cloud-native.slack.com/archives/CL25937SP/p1583249306101000?thread_ts=1583243694.092700&cid=CL25937SP

@d-ulyanov

GiedriusS:
@IKSIN any updates on your investigation? I ran into the same problem and tried to do some digging but still couldn't find anything specific. I concur: refetches and JSON/binary index headers don't make a difference. I was wondering perhaps you've found something.

d-ulyanov:
[...]

PR #2371, opened 2020-04-03
binaryHeader: Fixed partial write issue for index-header.

bwplotka:
Fixes https://github.com/thanos-io/thanos/issues/2213

This cause was indicated as the potential root cause for regression of latency and also causes the potentially critical issue
for store GW, where manual delete of index-header from local storage was required.

This might be considered as a blocker for 0.12, so it would be worth it to port it to 0.12 TBH @squat WDYT?

### DIFF
--- a/pkg/block/indexheader/binary_reader.go
+++ b/pkg/block/indexheader/binary_reader.go
@@ -71,20 +71,21 @@ type BinaryTOC struct {
 }
 
 // WriteBinary build index-header file from the pieces of index in object storage.
-func WriteBinary(ctx context.Context, bkt objstore.BucketReader, id ulid.ULID, fn string) (err error) {
+func WriteBinary(ctx context.Context, bkt objstore.BucketReader, id ulid.ULID, filename string) (err error) {
 	ir, indexVersion, err := newChunkedIndexReader(ctx, bkt, id)
 	if err != nil {
 		return errors.Wrap(err, "new index reader")
 	}
+	tmpFilename := filename + ".tmp"
 
 	// Buffer for copying and encbuffers.
 	// This also will control the size of file writer buffer.
 	buf := make([]byte, 32*1024)
-	bw, err := newBinaryWriter(fn, buf)
+	bw, err := newBinaryWriter(tmpFilename, buf)
 	if err != nil {
 		return errors.Wrap(err, "new binary index header writer")
 	}
-	defer runutil.CloseWithErrCapture(&err, bw, "close binary writer for %s", fn)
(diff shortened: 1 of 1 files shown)


bwplotka:
cc @mkabischev

pkg/block/indexheader/binary_reader.go
```diff
@@ -71,20 +71,21 @@ type BinaryTOC struct {
...
 	if err != nil {
 		return errors.Wrap(err, "new index reader")
 	}
+	tmpFilename := filename + ".tmp"
 
 	// Buffer for copying and encbuffers.
 	// This also will control the size of file writer buffer.
 	buf := make([]byte, 32*1024)
-	bw, err := newBinaryWriter(fn, buf)
+	bw, err := newBinaryWriter(tmpFilename+".tmp", buf)
```
vitkovskii, line 84: `.tmp` added twice?

bwplotka: Fixed. :man_facepalming:

pkg/block/indexheader/binary_reader.go
```diff
@@ -71,20 +71,21 @@ type BinaryTOC struct {
...
 	// Buffer for copying and encbuffers.
 	// This also will control the size of file writer buffer.
 	buf := make([]byte, 32*1024)
-	bw, err := newBinaryWriter(fn, buf)
+	bw, err := newBinaryWriter(tmpFilename, buf)
 	if err != nil {
 		return errors.Wrap(err, "new binary index header writer")
 	}
-	defer runutil.CloseWithErrCapture(&err, bw, "close binary writer for %s", fn)
+	defer runutil.CloseWithErrCapture(&err, bw, "close binary writer for %s", tmpFilename)
```
GiedriusS, line 88: Since `binaryWriter.Close()` calls `FileWriter.Close()` which does `file.Sync()` and `file.Flush()` maybe it would be worth being extra careful and explicitly calling it before `os.Rename`?

bwplotka: yes!

bwplotka: Thanks for this good point

squat:
if we want this in v0.12.0, should we instead PR into release-0.12?

bwplotka:
Yes, will do with @GiedriusS fixes.

brancz:
Correct me if I'm wrong, but I think we can close this as we're going to be merging this into master when we merge `release-0.12` back into master no?

squat:
ack we can close this now @brancz :+1:

PR #2379, opened 2020-04-05
Fix markdown docs style and link in the sharding doc

yeya24:

## Changes

- Update the promtail link
- We also support the default `replace` relabel action
- Fix the markdown style issue in the website

## Verification

Before: https://thanos.io/sharding.md/

After: https://deploy-preview-2379--thanos-io.netlify.com/sharding.md/

### DIFF
--- a/docs/sharding.md
+++ b/docs/sharding.md
@@ -22,14 +22,18 @@ This means that with time you might want to scale out certain components e.g:
 
 # Relabelling
 
-Similar to [promtail](https://github.com/grafana/loki/blob/master/docs/promtail.md#scrape-configs) this config will follow native
+Similar to [promtail](https://github.com/grafana/loki/blob/master/docs/clients/promtail/configuration.md#relabel_config) this config will follow native
 [Prometheus relabel-config](https://prometheus.io/docs/prometheus/latest/configuration/configuration/#relabel_config) syntax.
 
 Now, thanos only support following relabel actions:
+
 * keep
+
 * drop
+
 * hashmod
     * `external labels` for all components
+
     * `__block_id` for store gateway, see this [example](https://github.com/observatorium/configuration/blob/bf1304b0d7bce2ae3fefa80412bb358f9aa176fb/environments/openshift/manifests/observatorium-template.yaml#L1514-L1521)
 
 The relabel config defines filtering process done on **every** synchronization with object storage.


docs/sharding.md
```diff
@@ -22,14 +22,20 @@ This means that with time you might want to scale out certain components e.g:
 
 # Relabelling
 
-Similar to [promtail](https://github.com/grafana/loki/blob/master/docs/promtail.md#scrape-configs) this config will follow native
+Similar to [promtail](https://github.com/grafana/loki/blob/master/docs/clients/promtail/configuration.md#relabel_config) this config will follow native
 [Prometheus relabel-config](https://prometheus.io/docs/prometheus/latest/configuration/configuration/#relabel_config) syntax.
 
 Now, thanos only support following relabel actions:
+
+* replace
```
bwplotka, line 30: We don't support relabel, do we?

yeya24: Thanks, you are right. I am not familiar with sharding. I just realized that we can have the relabel config file with actions like `replace`, `labelmap`, etc. But they are all meaningless in the label meta filter.

yeya24: Do you think it is not that user-friendly? Maybe we can add a check in the parsing phase of the relabel config file. And say we don't support these actions.

bwplotka: What do you mean it's not user friendly? You are asking why `relabel` does not work? It's because label name or values do not matter. What matters only if they exists or not after relabel.

yeya24: Yes. You are right. But the relabel config file we use is exactly the same as Prometheus, and we use https://github.com/thanos-io/thanos/blob/1a2edf19475b854181e5abaa4b6f1c9caf2a1547/pkg/block/fetcher.go#L537 to process the labels. So if we have some other actions in the relabel config, like using `replace` or `labelmap`, there is no any error or log to show that these actions are not supported.

Many users might think this is the same as Prometheus' s relabel just like me. But it is actually just a filter. Maybe we can give them some hints that these actions are meaningless. That's what I mean about not that user-friendly. Correct me if I am wrong.

bwplotka: True. We can totally error on those special labels in separate step, no? (: Before processing

bwplotka: On configuration change, even

yeya24: Do you think this worth a PR? Or we can open a good-first issue for it

bwplotka: Yes yes! Please do :heart: and explain AC (Acceptance Criteria)

PR #2385, opened 2020-04-06
compact: Made MarkForDeletion less strict; Added more debugability to block deletion logic, made meta sync explicit.

bwplotka:
Also:

* Changed order: Now BestEffortCleanAbortedPartialUploads is before DeleteMarkedBlocks.
* Increment markedForDeletion counter only when we actually uploaded block.
* Fixed logging issues.

### DIFF
--- a/pkg/block/fetcher.go
+++ b/pkg/block/fetcher.go
@@ -189,8 +189,8 @@ func NewMetaFetcher(logger log.Logger, concurrency int, bkt objstore.Instrumente
 }
 
 // NewMetaFetcher transforms BaseFetcher into actually usable *MetaFetcher.
-func (f *BaseFetcher) NewMetaFetcher(reg prometheus.Registerer, filters []MetadataFilter, modifiers []MetadataModifier) *MetaFetcher {
-	return &MetaFetcher{metrics: newFetcherMetrics(reg), wrapped: f, filters: filters, modifiers: modifiers}
+func (f *BaseFetcher) NewMetaFetcher(reg prometheus.Registerer, filters []MetadataFilter, modifiers []MetadataModifier, logTags ...interface{}) *MetaFetcher {
+	return &MetaFetcher{metrics: newFetcherMetrics(reg), wrapped: f, filters: filters, modifiers: modifiers, logger: log.With(f.logger, logTags...)}
 }
 
 var (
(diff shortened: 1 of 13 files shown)


bwplotka:
TODO: 
* Our fix https://github.com/thanos-io/thanos/commit/e1b1d716adfc838f776036ba132090c7644bed9e and https://github.com/thanos-io/thanos/commit/c05daeeace272a94fae81fd0c15ff70adab50e10 requires CHANGELOG entry.

This one I think not as this is something we introduced in 0.11+

bwplotka:
Should be fine now.

pkg/block/fetcher.go
```diff
@@ -189,8 +189,8 @@ func NewMetaFetcher(logger log.Logger, concurrency int, bkt objstore.Instrumente
 }
 
 // NewMetaFetcher transforms BaseFetcher into actually usable *MetaFetcher.
-func (f *BaseFetcher) NewMetaFetcher(reg prometheus.Registerer, filters []MetadataFilter, modifiers []MetadataModifier) *MetaFetcher {
-	return &MetaFetcher{metrics: newFetcherMetrics(reg), wrapped: f, filters: filters, modifiers: modifiers}
+func (f *BaseFetcher) NewMetaFetcher(reg prometheus.Registerer, filters []MetadataFilter, modifiers []MetadataModifier, logTags ...interface{}) *MetaFetcher {
```
squat, line 192: I think this is a bit funny. A nicer API would be for `NewMetaFetcher` to accept a `log.Logger`. Not necessarily a blocker

pkg/compact/compact.go
```diff
@@ -228,16 +246,18 @@ func (s *Syncer) GarbageCollect(ctx context.Context) error {
...
 	begin := time.Now()
 
-	duplicateIDs := s.duplicateBlocksFilter.DuplicateIDs()
+	// Ignore filter exists before deduplicate filter.
 	deletionMarkMap := s.ignoreDeletionMarkFilter.DeletionMarkBlocks()
+	duplicateIDs := s.duplicateBlocksFilter.DuplicateIDs()
 
 	// GarbageIDs contains the duplicateIDs, since these blocks can be replaced with other blocks.
-	// We also remove ids present in deletionMarkMap since these blocks are already marked for deletion.
+	// We also remove ids present in deletionMarkMapAND  01E4NVH3TCJB3QA3WNANTYH66F since these blocks are already marked for deletion.
```
squat, line 254: What does this mean exactly?

bwplotka: Copy pasta, sry.

pkg/block/fetcher.go
bwplotka: I want to share same logger. Let's maybe merge and iterate over it (:

bwplotka:
Addressed (:

PR #2386, opened 2020-04-06
binaryHeader: fixed "invalid size" error with non existing label value

### LINKED ISSUE #2213: Query latency increased after upgrade Thanos v0.10.1 ->v 0.11.0

mkabischev:
Fixes #2213 

There is a problem in `sort.Search` usage that causes floating error when try to get offsets for non existing combination of name & values.

### DIFF
--- a/pkg/block/indexheader/binary_reader.go
+++ b/pkg/block/indexheader/binary_reader.go
@@ -685,14 +685,10 @@ func (r BinaryReader) postingsOffset(name string, values ...string) ([]index.Ran
 		value := values[valueIndex]
 
 		i := sort.Search(len(e.offsets), func(i int) bool { return e.offsets[i].value >= value })
-		if i == len(e.offsets) {
-			// We're past the end.
+		if i == len(e.offsets) || e.offsets[i].value != value {
 			break
 		}
-		if i > 0 && e.offsets[i].value != value {
-			// Need to look from previous entry.
-			i--
-		}
+
 		// Don't Crc32 the entire postings offset table, this is very slow
 		// so hope any issues were caught at startup.
 		d := encoding.NewDecbufAt(r.b, int(r.toc.PostingsOffsetTable), nil)


pkg/block/indexheader/binary_reader.go
```diff
@@ -685,14 +685,10 @@ func (r BinaryReader) postingsOffset(name string, values ...string) ([]index.Ran
 		value := values[valueIndex]
 
 		i := sort.Search(len(e.offsets), func(i int) bool { return e.offsets[i].value >= value })
-		if i == len(e.offsets) {
-			// We're past the end.
+		if i == len(e.offsets) || e.offsets[i].value != value {
 			break
 		}
-		if i > 0 && e.offsets[i].value != value {
```
bwplotka, line 692: I think we still need this (:

mkabischev: why? Really I didn't get this logic.

bwplotka: How we decrement the i then?

bwplotka: Looks like This is still not clear, I think what would help is some extra unit test to verify

mkabischev: Why do we need to decrement it? I mean that there is no value found then we need `break`.

bwplotka: The whole idea behind this code is to find matching offset OR the largest offset that is before wanted value. Then we iterate one by one to find exact offset.

This is because binary index stores only 1/32 of offsets (to not keep all in memory). Problem is that our tests have only 3 postings most likely that's why we test nothing..

It's extremely concerning that our tests passed given that this version will literally skip 31/32 postings ((:

mkabischev:
We found the root cause of error in another place, so close this PR. Fix will be in another one.

PR #2390, opened 2020-04-07
store: Fixed binary header bug that was causing all postings to be kept in memory instead of 1/32 as we meant.

bwplotka:
Spotted by @mkabischev! Thanks to you and @d-ulyanov as well! Epic finding :+1: :muscle: 

Test output before fix:

```go
					testutil.Equals(t, 1, br.version)
					testutil.Equals(t, 2, br.indexVersion)
					testutil.Equals(t, &BinaryTOC{Symbols: headerLen, PostingsOffsetTable: 66}, br.toc)
					testutil.Equals(t, int64(626), br.indexLastPostingEnd)
					testutil.Equals(t, 8, br.symbols.Size())
					testutil.Equals(t, map[string]*postingValueOffsets{
						"": {
							offsets:       []postingOffset{{value: "", tableOff: 4}},
							lastValOffset: 392,
						},
						"a": {
							offsets: []postingOffset{
								{value: "1", tableOff: 9},
								{value: "11", tableOff: 16},
								{value: "12", tableOff: 24},
								{value: "2", tableOff: 32},
								{value: "3", tableOff: 39},
								{value: "4", tableOff: 46},
								{value: "5", tableOff: 53},
								{value: "6", tableOff: 60},
								{value: "7", tableOff: 67},
								{value: "8", tableOff: 74},
								{value: "9", tableOff: 81},
							},
							lastValOffset: 572,
						},
						"longer-string": {
							offsets:       []postingOffset{{value: "1", tableOff: 88}},
							lastValOffset: 622,
						},
					}, br.postings)
					testutil.Equals(t, 0, len(br.postingsV1))
					testutil.Equals(t, 2, len(br.nameSymbols))
```

### DIFF
--- a/pkg/block/indexheader/binary_reader.go
+++ b/pkg/block/indexheader/binary_reader.go
@@ -42,8 +42,6 @@ const (
 	// MagicIndex are 4 bytes at the head of an index-header file.
 	MagicIndex = 0xBAAAD792
 
-	symbolFactor = 32
-
 	postingLengthFieldSize = 4
 )
 
@@ -428,11 +426,12 @@ type BinaryReader struct {
 	c io.Closer
 
 	// Map of LabelName to a list of some LabelValues's position in the offset table.
-	// The first and last values for each name are always present.
+	// The first and last values for each name are always present, we keep only `symbolsFactor` part of for rest.
 	postings map[string]*postingValueOffsets
 	// For the v1 format, labelname -> labelvalue -> offset.
 	postingsV1 map[string]map[string]index.Range
 
+	// Symbols struct that keeps only `symbolsFactor` part in the memory, looks up via mmap rest.
 	symbols     *index.Symbols
 	nameSymbols map[uint32]string // Cache of the label name symbol lookups,
 	// as there are not many and they are half of all lookups.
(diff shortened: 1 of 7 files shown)


pkg/block/indexheader/binary_reader.go
```diff
@@ -428,11 +426,12 @@ type BinaryReader struct {
 	c io.Closer
 
 	// Map of LabelName to a list of some LabelValues's position in the offset table.
-	// The first and last values for each name are always present.
+	// The first and last values for each name are always present, we keep only `symbolsFactor` part of for rest.
```
squat, line 429: there is no identifier in code with the name `symbolsFactor` anymore. Maybe we should write this as `1/postingOffsetsInMemSampling` or something

bwplotka: good catch! Fixed

bwplotka: Good catch, thanks!

bwplotka:
cc @pracucci @pstibrany

bwplotka:
Looks like some users like to add more labels: 

image


At this point, I am starting to think... Do we really need to index ALL labels in the series in Prometheus? ;p 

cc @brian-brazil @beorn7 @brancz @SuperQ

brian-brazil:
> Do we really need to index ALL labels in the series in Prometheus? ;p 

If you're willing to have queries get massively slower we could of course drop it ;p

bwplotka:
But that's the thing, most of labels are not used.

beorn7:
If you can predict which labels will be queried and for which it is unlikely, you could optimize the indexing based on that prediction…

brian-brazil:
One thing I've pondered is ordering the postings by cardinality when we're evaluating them, but I'm not sure if it'd actually help much.

PR #2393, opened 2020-04-08
store: Fixed critical bug, when certain not-existing values queried was causing "invalid size" error inside binary index-header

bwplotka:
The reason, why we could not reproduce it locally, was that for most of non-existing values, we were lucky that buffer was still long enough and we could read and decode some (malformed) variadic type. For certain rare cases, the buffer was not long enough or varint was failing.

Fixed and spotted thanks to amazing @mkabischev!

* Added more regression tests for the binary header.

Without the fix it fails with:
```
            header_test.go:154: header_test.go:154:

                	exp: range not found

                	got: get postings offset entry: invalid size
```


bwplotka:
With this I believe can kick v0.12.0-rc1 :muscle:

bwplotka:
cc @squat

Issue #2391, opened 2020-04-07
ThanosRuleTSDBNotIngestingSamples firing without rules


brancz:
As mentioned in the code comment:

https://github.com/thanos-io/thanos/blob/1a2edf19475b854181e5abaa4b6f1c9caf2a1547/mixin/thanos/alerts/rule.libsonnet#L174-L175

This alert currently fires even if there are no rules loaded. We should ensure that it only fires if we are performing rule evaluations that should have written data.

@thanos-io/thanos-maintainers 

cc @lilic as she found this firing incorrectly in one of our clusters.

brancz:
An easy metric we could add, although it wouldn't be 100% correct would be having a metric for the number of rule files and or rules loaded, so we can make the alert only fire if there is actually a rule in the first place, however, there could be recording rules that legitimately don't write any data, for example if they have a `> $threshold` filter.

bwplotka:
Wow so actually it happened ;p

brancz:
Poking around at the metrics that we have, I'm thinking that we should maybe change the alert definition all together, to instead of monitoring writing data, we observe rule evaluations. Something along the lines of:

```
sum without(strategy) (rate(prometheus_rule_evaluations_total{job="observatorium-thanos-rule"}[10m])) <= 0
```

Plus exposing the amount of rules loaded we can conditionally make this fire only if there are actually rules to be evaluated.

PR #2394, opened 2020-04-08
rule: Add metric exposing the amount of rules loaded

brancz:

## Changes

Add gauge metric that exposes the amount of rules loaded by the rule manager. I went with the gaugefunc strategy and actual rules as opposed to rule files as in theory there could be an empty rule file (or just empty rule group) and the alert would falsely fire in that case.

If people prefer I'm also happy to point this at the 0.12 release, but I'm also fine with waiting for 0.13.

Related to https://github.com/thanos-io/thanos/issues/2391. With this metric we could make the alert more stable as described in the issue.

## Verification

`curl`ed the metrics endpoint and the metric was there :) 

@thanos-io/thanos-maintainers 

cc @lilic


bwplotka:
What's wrong with existing `thanos_rule_loaded_rules`?

brancz:
Nevermind, I didn't see that one :) 

Closing.

PR #2398, opened 2020-04-08
mixin: Change critical rule alert to be symptom based

brancz:

## Changes

This change makes the critical (typically paging) alert more symptom
based, rather than observing data written to disk. Additionally after
this change the alert will only fire if there are actually rules loaded.

Additionally to no rules loaded the previous alert was also prone to rules that legitimately are not writing data.

Fixes https://github.com/thanos-io/thanos/issues/2391

## Verification

Verified against an incident we had, plus verified that this doesn't fire against a Thanos Rule instance that does not have any rules loaded.

@thanos-io/thanos-maintainers @lilic

### DIFF
--- a/mixin/thanos/alerts/rule.libsonnet
+++ b/mixin/thanos/alerts/rule.libsonnet
@@ -171,15 +171,15 @@
             },
           },
           {
-            // NOTE: This alert will give false positive if no rules are configured.
-            alert: 'ThanosRuleTSDBNotIngestingSamples',
+            alert: 'ThanosNoRuleEvaluations',
             annotations: {
-              message: 'Thanos Rule {{$labels.job}} did not ingest any samples for the last 15 minutes.',
+              message: 'Thanos Rule {{$labels.job}} did not perform any rule evaluations in the past 2 minutes.',
             },
             expr: |||
-              sum by (job) (rate(prometheus_tsdb_head_samples_appended_total{%(selector)s}[5m])) <= 0
+              sum(rate(prometheus_rule_evaluations_total{%(selector)s}[2m])) <= 0
+                and
+              sum(thanos_rule_loaded_rules{%(selector)s}) > 0
             ||| % thanos.rule,
-            'for': '10m',
             labels: {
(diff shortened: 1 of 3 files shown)


mixin/thanos/alerts/rule.libsonnet
```diff
@@ -171,15 +171,15 @@
...
-            // NOTE: This alert will give false positive if no rules are configured.
-            alert: 'ThanosRuleTSDBNotIngestingSamples',
+            alert: 'ThanosNoRuleEvaluations',
             annotations: {
-              message: 'Thanos Rule {{$labels.job}} did not ingest any samples for the last 15 minutes.',
+              message: 'Thanos Rule {{$labels.job}} did not perform any rule evaluations in the past 2 minutes.',
             },
             expr: |||
-              sum by (job) (rate(prometheus_tsdb_head_samples_appended_total{%(selector)s}[5m])) <= 0
+              sum(rate(prometheus_rule_evaluations_total{%(selector)s}}[2m])) <= 0
```
aditya-konarde, line 179: Is this setting an expectation that users are only doing rule evaluations at an interval of < 2 minutes? 

Do we instead want a variable that defaults to 2m in the range selector? 

I can't immediately think of an example of a rule that's evaluated over such infrequent intervals, but that may be a case for some consumers. Wdyt?

brancz: The evaluation interval is for a rule group, so the expectation would be that even if the evaluation interval is 2m, at least 1 rule would have been attempted every 2 mins. And yes same as with scrapes, evaluation intervals higher than 2m don't really make sense. I'd prefer to not make things configurable prematurely.

aditya-konarde: Sounds good, thanks for clarifying :)

brancz:
> I guess this metric was not incremented for our incident case?

Define "this"?

bwplotka:
`prometheus_rule_evaluations_total`

brancz:
Correct.

PR #2407, opened 2020-04-09
Query: Add servername to grpc dial options for DNS stores (Fixes #1507)

### LINKED ISSUE #1507: thanos+ingress-nginx+grpc: impossible setup due missing host header
quay.io/thanos/thanos:v0.7.0

**What happened**
i setup 2 kubernetes clusters, thanos query is in one cluster (and a local prometheus+sidecar) and need to query the remote kubernetes cluster thanos sidecar, all running in AWS (but not using eks)
I created one ingress-nginx with support for grpc with this config:
```
---
apiVersion: extensions/v1beta1
kind: Ingress
metadata:
  name: monitoring-ingress
  namespace: monitoring
  annotations:
    kubernetes.io/ingress.class: "nginx"
spec:
  rules:
  - host: prometheus-k8s-live-a.ops.example.com
    http:
      paths:
      - path: /
        backend:
          serviceName: prometheus-k8s-live-a
          servicePort: 9090
  - host: prometheus-k8s-live-b.ops.example.com
    http:
      paths:
      - path: /
        backend:
          serviceName: prometheus-k8s-live-b
          servicePort: 9090
  tls:
  - hosts:
    - prometheus-k8s-live-a.ops.example.com
    - prometheus-k8s-live-b.ops.example.com
---
apiVersion: extensions/v1beta1
kind: Ingress
metadata:
  annotations:
    kubernetes.io/ingress.class: "nginx"
    nginx.ingress.kubernetes.io/ssl-redirect: "true"
    nginx.ingress.kubernetes.io/backend-protocol: "GRPC"
  name: grpc-ingress
  namespace: monitoring
spec:
  rules:
  - host: sidecar-k8s-live-a.ops.example.com
    http:
      paths:
      - backend:
          serviceName: sidecar-k8s-live-a
          servicePort: 10911
  - host: sidecar-k8s-live-b.ops.example.com
    http:
      paths:
      - backend:
          serviceName: sidecar-k8s-live-b
          servicePort: 10911
  tls:
  - hosts:
      - sidecar-k8s-live-a.ops.example.com
      - sidecar-k8s-live-b.ops.example.com
```

thanos query is using 
```
--store=sidecar-k8s-live-a.ops.example.com.:443
--store=sidecar-k8s-live-a.ops.example.com.:443
```

I can connect to the prometheus url, but the sidecar grpc fail in thanos query.
looking to the nginx logs i can see the query arriving in http2, but returning 400. Doing a curl i can get a 503, but probably just because it is not really grpc. Changing the ingress-nginx logs to show the host header, i can see that curl is sending the correct host header, but for thanos query the logs show only `_`, it is either sending a empty one or a `_`.

I wanted to share the ingress to receive the https requests for prometheus and the grpc and using the host to redirect the request to the correct service. Sadly thanos query fail to send the host header and so the nginx can't apply the virtual_host search and servers the request from the default site.

<details>Logs
<p>

```
172.27.119.135 - [172.27.119.135] - - [10/Sep/2019:15:02:40 +0000] "PRI * HTTP/2.0" 400 163 "-" "-" 0 0.001 [] [] - - - - 477873c7a336618ccf06cf9c03fe8d97
172.27.119.135 - [172.27.119.135] - - [10/Sep/2019:15:02:40 +0000] "PRI * HTTP/2.0" 400 163 "-" "-" 0 0.003 [] [] - - - - c32e68975e91159a64326b55d4b72934
2019/09/10 15:02:40 [error] 1137#1137: *7155 upstream rejected request with error 2 while reading response header from upstream, client: 172.26.81.74, server: sidecar-k8s-live-a.ops.example.com, request: "PRI / HTTP/1.1", upstream: "grpc://100.96.136.200:10911", host: "sidecar-k8s-live-a.ops.example.com"
172.26.81.74 - [172.26.81.74] - - [10/Sep/2019:15:02:40 +0000] "PRI / HTTP/1.1" 502 163 "-" "curl/7.58.0" 189 0.002 [monitoring-sidecar-k8s-live-a-10911] [] 100.96.136.200:10911 0 0.004 502 4e08c4e8c6d8df148c5bc3a68d61ccf9

```
here we can see that the thanos query requests do not trigger the virtual_host, but the curl  one, with host, is redirected to thanos sidecar
</p>
</details>

j3p0uk:
Closes #1507 

## Changes

To avoid needing a query per remote cluster, get the name to add to
the dial options from the dns provider when making the grpc connection.

## Verification

Building via make docker and running a query against multiple remote query services through nginx proxies.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -11,6 +11,8 @@ We use *breaking* word for marking changes that are not backward compatible (rel
 
 ## Unreleased
 
+- [#2407](https://github.com/thanos-io/thanos/pull/2407) Query: Add servername to grpc dial options for DNS stores (Fixes #1507)
+
 ### Fixed
 
 - [#2288](https://github.com/thanos-io/thanos/pull/2288) Ruler: Fixes issue #2281 bug in ruler with parsing query addr with path prefix
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -11,7 +11,7 @@ We use *breaking* word for marking changes that are not backward compatible (rel
 
 ## Unreleased
 
-- [#2407](https://github.com/thanos-io/thanos/pull/2407) Query: Add servername to grpc dial options for DNS stores (Fixes #1507)
+- [#2407](https://github.com/thanos-io/thanos/pull/2407) Query: Add servername to gRPC dial options for DNS stores (Fixes #1507)
 
 ### Fixed
 
--- a/cmd/thanos/query.go
+++ b/cmd/thanos/query.go
@@ -56,7 +56,7 @@ func registerQuery(m map[string]setupFunc, app *kingpin.Application) {
 	key := cmd.Flag("grpc-client-tls-key", "TLS Key for the client's certificate").Default("").String()
 	caCert := cmd.Flag("grpc-client-tls-ca", "TLS CA Certificates to use to verify gRPC servers").Default("").String()
 	serverName := cmd.Flag("grpc-client-server-name", "Server name to verify the hostname on the returned gRPC certificates. See https://tools.ietf.org/html/rfc4366#section-3.1").Default("").String()
-	dnsServerName := cmd.Flag("grpc-client-dns-server-name", "For stores that are DNS, use the dns name as the server name for connection.  Needed when proxying through nginx").Default("false").Bool()
+	dnsServerName := cmd.Flag("grpc-client-dns-server-name", "For stores that are DNS, use the dns name as the server name for connection.  Needed when proxying through nginx, and only applicable if using the secure flag").Default("false").Bool()
 
 	webRoutePrefix := cmd.Flag("web.route-prefix", "Prefix for API and UI endpoints. This allows thanos UI to be served on a sub-path. This option is analogous to --web.route-prefix of Promethus.").Default("").String()
 	webExternalPrefix := cmd.Flag("web.external-prefix", "Static prefix for all HTML links and redirect URLs in the UI query web interface. Actual endpoints are still served on / or the web.route-prefix. This allows thanos UI to be served behind a reverse proxy that strips a URL sub-path.").Default("").String()
@@ -108,6 +108,10 @@ func registerQuery(m map[string]setupFunc, app *kingpin.Application) {
 	storeResponseTimeout := modelDuration(cmd.Flag("store.response-timeout", "If a Store doesn't send any data in this specified duration then a Store will be ignored and partial data will be returned if it's enabled. 0 disables timeout.").Default("0ms"))
 
 	m[comp.String()] = func(g *run.Group, logger log.Logger, reg *prometheus.Registry, tracer opentracing.Tracer, _ <-chan struct{}, _ bool) error {
+		if *dnsServerName && !*secure {
+			return errors.New("grpc-client-dns-server-name specified without grpc-client-tls-secure")
+		}
+
 		selectorLset, err := parseFlagLabels(*selectorLabels)
 		if err != nil {
 			return errors.Wrap(err, "parse federation labels")
@@ -260,6 +264,7 @@ func runQuery(
 			},
 			dialOpts,
 			unhealthyStoreTimeout,
+			dnsServerName,
 			cert,
 			key,
 			caCert,
--- a/cmd/thanos/query.go
+++ b/cmd/thanos/query.go
@@ -56,7 +56,7 @@ func registerQuery(m map[string]setupFunc, app *kingpin.Application) {
 	key := cmd.Flag("grpc-client-tls-key", "TLS Key for the client's certificate").Default("").String()
 	caCert := cmd.Flag("grpc-client-tls-ca", "TLS CA Certificates to use to verify gRPC servers").Default("").String()
 	serverName := cmd.Flag("grpc-client-server-name", "Server name to verify the hostname on the returned gRPC certificates. See https://tools.ietf.org/html/rfc4366#section-3.1").Default("").String()
-	dnsServerName := cmd.Flag("grpc-client-dns-server-name", "For stores that are DNS, use the dns name as the server name for connection.  Needed when proxying through nginx, and only applicable if using the secure flag").Default("false").Bool()
+	dnsServerName := cmd.Flag("grpc-client-dns-server-name", "For stores that are DNS, use the dns name as the server name for connection. Only applicable if using the secure flag").Default("false").Bool()
 
 	webRoutePrefix := cmd.Flag("web.route-prefix", "Prefix for API and UI endpoints. This allows thanos UI to be served on a sub-path. This option is analogous to --web.route-prefix of Promethus.").Default("").String()
 	webExternalPrefix := cmd.Flag("web.external-prefix", "Static prefix for all HTML links and redirect URLs in the UI query web interface. Actual endpoints are still served on / or the web.route-prefix. This allows thanos UI to be served behind a reverse proxy that strips a URL sub-path.").Default("").String()
--- a/cmd/thanos/query.go
+++ b/cmd/thanos/query.go
@@ -56,6 +56,7 @@ func registerQuery(m map[string]setupFunc, app *kingpin.Application) {
 	key := cmd.Flag("grpc-client-tls-key", "TLS Key for the client's certificate").Default("").String()
 	caCert := cmd.Flag("grpc-client-tls-ca", "TLS CA Certificates to use to verify gRPC servers").Default("").String()
 	serverName := cmd.Flag("grpc-client-server-name", "Server name to verify the hostname on the returned gRPC certificates. See https://tools.ietf.org/html/rfc4366#section-3.1").Default("").String()
+	dnsServerName := cmd.Flag("grpc-client-dns-server-name", "For stores that are DNS, use the dns name as the server name for connection.  Needed when proxying through nginx").Default("false").Bool()
 
 	webRoutePrefix := cmd.Flag("web.route-prefix", "Prefix for API and UI endpoints. This allows thanos UI to be served on a sub-path. This option is analogous to --web.route-prefix of Promethus.").Default("").String()
 	webExternalPrefix := cmd.Flag("web.external-prefix", "Static prefix for all HTML links and redirect URLs in the UI query web interface. Actual endpoints are still served on / or the web.route-prefix. This allows thanos UI to be served behind a reverse proxy that strips a URL sub-path.").Default("").String()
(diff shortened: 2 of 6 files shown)


pkg/query/storeset.go
```diff
@@ -420,7 +445,13 @@ func (s *StoreSet) getActiveStores(ctx context.Context, stores map[string]*store
...
 			if !seenAlready {
 				// New store or was unactive and was removed in the past - create new one.
-				conn, err := grpc.DialContext(ctx, addr, s.dialOpts...)
+				tlsCfg, err := tls.NewClientConfig(s.logger, s.cert, s.key, s.caCert, spec.ServerName())
+				if err != nil {
+					level.Warn(s.logger).Log("msg", "update of store node failed", "err", errors.Wrap(err, "setting TLS"), "address", addr)
+					return
+				}
+				dialOpts := append(s.dialOpts, grpc.WithTransportCredentials(credentials.NewTLS(tlsCfg)))
+				conn, err := grpc.DialContext(ctx, addr, dialOpts...)
```
povilasv, line 454: I believe this makes an assumption that store is listening with TLS & won't work in insecure mode. Could you please verify that?

j3p0uk: Yes, That would be absolutely true.  I need to conditionally do this based on the combination of the secure flag and the new grpc-client-dns-server-name flag.

EvertonSA:
looks legit, can we have this approved and released?

pkg/query/storeset.go
```diff
@@ -420,7 +448,17 @@ func (s *StoreSet) getActiveStores(ctx context.Context, stores map[string]*store
 			st, seenAlready := stores[addr]
 			if !seenAlready {
 				// New store or was unactive and was removed in the past - create new one.
-				conn, err := grpc.DialContext(ctx, addr, s.dialOpts...)
+				dialOpts := s.dialOpts
+				// Update the dialOpts if we are asked to add dnsServerName.
+				if s.dnsServerName {
```
squat, line 453: Hmm this seems a bit odd. The NewStoreSet func already accepts a slice of dialopts. I think we should expect that the dialopts has the correct WithTransportCredentials already set, no? ie correctly configure this option in cmd/Thanos/query rather than leaking configuration into this struct. Also, it looks to me like the server name is already set in the dialopts generated with extgrpc https://github.com/j3p0uk/thanos/blob/40526f52f54d4501737e5246c0e71e56dd7e0b2d/pkg/extgrpc/client.go#L56

CHANGELOG.md
```diff
@@ -11,6 +11,8 @@ We use *breaking* word for marking changes that are not backward compatible (rel
 
 ## Unreleased
 
+- [#2407](https://github.com/thanos-io/thanos/pull/2407) Query: Add servername to grpc dial options for DNS stores (Fixes #1507)
```
squat, line 14: nit: s/grpc/gRPC/

cmd/thanos/query.go
```diff
@@ -56,6 +56,7 @@ func registerQuery(m map[string]setupFunc, app *kingpin.Application) {
 	key := cmd.Flag("grpc-client-tls-key", "TLS Key for the client's certificate").Default("").String()
 	caCert := cmd.Flag("grpc-client-tls-ca", "TLS CA Certificates to use to verify gRPC servers").Default("").String()
 	serverName := cmd.Flag("grpc-client-server-name", "Server name to verify the hostname on the returned gRPC certificates. See https://tools.ietf.org/html/rfc4366#section-3.1").Default("").String()
+	dnsServerName := cmd.Flag("grpc-client-dns-server-name", "For stores that are DNS, use the dns name as the server name for connection.  Needed when proxying through nginx, and only applicable if using the secure flag").Default("false").Bool()
```
squat, line 59: Nit: we use single spaces after periods in this repo

squat: This sounds like it should be more common than just for NGINX. I'd like to avoid naming one single proxy if this is a generic problem with ingress

pkg/query/storeset.go
```diff
@@ -38,6 +40,9 @@ type StoreSpec interface {
 	Metadata(ctx context.Context, client storepb.StoreClient) (labelSets []storepb.LabelSet, mint int64, maxt int64, storeType component.StoreAPI, err error)
 	// StrictStatic returns true if the StoreAPI has been statically defined and it is under a strict mode.
 	StrictStatic() bool
+	// ServerName returns StoreAPI ServerName for the store spec.
```
squat, line 43: ... returns the StoreAPI's server name...

pkg/query/storeset.go
```diff
@@ -38,6 +40,9 @@ type StoreSpec interface {
...
+	// It is needed to get to a StoreAPI behind an nginx proxy.
```
squat, line 44: Same question about specifying NGINX

pkg/query/storeset.go
```diff
@@ -53,12 +58,19 @@ type StoreStatus struct {
 type grpcStoreSpec struct {
 	addr         string
 	strictstatic bool
+	serverName   string
+}
+
+// NewGRPCStoreSpecServerName creates store pure gRPC spec with a Server Name.
```
squat, line 64: s/Server Name/server name/

pkg/query/storeset.go
squat: Also, taking out this new dial opt and server name bool from this struct would eliminate the need for extending the StoreSpec interface and adding the extra field to the grpcStoreSpec struct

j3p0uk: The bug, as it stands, is that the arguments provided to the query at start time do not allow for individual server names per configured store.  The store set gets populated with the resolved addresses of the stores in the store spec and not the original server name, as far as I could tell.  Given this, passing through the flag to generate the TLS portion of the dialOpts on demand at connection time by retrieving the server name from the dns provider.  I passed minimal data through to accomplish that.

GiedriusS: Indeed this is a bit weird because with this we not only have two places where these options are set but also this newly introduced option overrides the other one. It should probably be refactored so that `query.NewGRPCStoreSpec*` would hook into some function which would give it a list of opts to use. We could discuss it on #thanos-dev.
Also, I wonder if there any downsides to enabling this kind of feature by default for everyone, and then only allow overriding the server name via the option that exists already? That would probably simplify things too.

j3p0uk: Sure, happy to discuss further.

My thoughts are that this looks odd because a StoreSet is a set of Stores, and the code assumes that each of the stores shares a particular set of characteristics, which may not be true (and is the basis for this bug fix).  The StoreSet currently holds attributes that can differ on a per-store basis.  The ServerName is one such attribute.

squat: Yes, but I think we need to refactor this because the current approach causes the StoreSet to have duplicate `WithTransportCredentials` dialopts, one of which overrides the other. This makes it hard to reason about the code IMO

j3p0uk: Chat in slack...

Sounds like implementation of this would be something like removing dialOpts from the StoreSet, and adding it to the StoreSpec.
We should be able to generate the dialOpts at creation of the StoreSpec array in the StoreSet, so would only be once at startup.
Need to decide on what flags should exist, and how would they interact and how to handle address-specified stores and DNS SRV records.

For the SRV case, I think the correct behaviour would be to enable lookup of the A record resolved from the SRV lookup.  This data is currently lost during resolution, as only the IPs are returned to the provider from resolver, and stored under the key of the original store spec.
To improve the DNS Provider lookup of address to server name, we could pass back the data to stash in the provider as a reverse lookup cache.  This would allow SRV to do addr -> A lookup, and others to do addr -> host lookup, simplifying the Provider.ServerName() function.

I can drop the “--grpc-client-dns-server-name” flag and then use “dns” in the “--grpc-client-server-name” to trigger addition of server name from DNS in the dialOpts, as “dns” is likely an illegal server name anyway.

GiedriusS: @bwplotka @squat ping ^

cmd/thanos/query.go
```diff
@@ -56,6 +56,7 @@ func registerQuery(m map[string]setupFunc, app *kingpin.Application) {
 	key := cmd.Flag("grpc-client-tls-key", "TLS Key for the client's certificate").Default("").String()
 	caCert := cmd.Flag("grpc-client-tls-ca", "TLS CA Certificates to use to verify gRPC servers").Default("").String()
 	serverName := cmd.Flag("grpc-client-server-name", "Server name to verify the hostname on the returned gRPC certificates. See https://tools.ietf.org/html/rfc4366#section-3.1").Default("").String()
+	dnsServerName := cmd.Flag("grpc-client-dns-server-name", "For stores that are DNS, use the dns name as the server name for connection. Only applicable if using the secure flag").Default("false").Bool()
```
povilasv, line 59: what if we change to his to string and instead of resolving the server name from dns we use this value instead?

cmd/thanos/query.go
```diff
@@ -240,7 +247,11 @@ func runQuery(
 			func() (specs []query.StoreSpec) {
 				// Add DNS resolved addresses.
 				for _, addr := range dnsProvider.Addresses() {
-					specs = append(specs, query.NewGRPCStoreSpec(addr, false))
+					if dnsServerName {
+						specs = append(specs, query.NewGRPCStoreSpecServerName(addr, false, dnsProvider.ServerName(addr)))
```
povilasv, line 251:
```suggestion
						specs = append(specs, query.NewGRPCStoreSpecServerName(addr, false, serverName)
```

can we just use serverName provided from `--grpc-client-server-name`  instead of relying on dns?

j3p0uk: Currently, the code as-is will use a provided server-name.  This works when using a single store from the querier.  But adding multiple store options to collect data from many stores will not be able to use individual server names as the code currently stands.

The change intended here is to allow for multiple stores to each use different server names.

j3p0uk:
Needs updating with new design as detailed in https://github.com/thanos-io/thanos/pull/2407#discussion_r414743260 and

kakkoyun:
> Needs updating with new design as detailed in [#2407 (comment)](https://github.com/thanos-io/thanos/pull/2407#discussion_r414743260) 

@j3p0uk Friendly ping. Are you still planning to work on this?

j3p0uk:
> > Needs updating with new design as detailed in [#2407 (comment)](https://github.com/thanos-io/thanos/pull/2407#discussion_r414743260)
> 
> @j3p0uk Friendly ping. Are you still planning to work on this?

I want to, but I haven't yet found the time to do so, and don't see myself getting to it in the short term :(

kakkoyun:
All good. I just asked so that we can keep this open.

billylindeman:
Is this still being tracked? @kakkoyun @j3p0uk I just hit this issue today

PR #2121, opened 2020-02-12
ui: add React UI from upstream Prometheus

adrien-f:
* [X] I added CHANGELOG entry for this change.
* [ ] Change is not relevant to the end user.

## Changes

* Add React UI accesible from the Query only for the moment as only the graph endpoint has been migrated.

I've been putting Thanos related modifications in the `ui/react-app/src/thanos` directory but there are some modifications in the root files. The goal is to be able to just copy paste the source from upstream and let the Thanos directory do the rest to ease upgrades.

## Verification

image

PR #2412, opened 2020-04-10
ui: add new React UI from Prometheus

prmsrswt:

## Changes

This PR is a continuation of #2121 . I have updated the React UI to incorporate new changes from upstream Prometheus. Also, as #2121 this is currently only accessible from the Query component.

## Verification
- Tested by running a dev environment locally.


GiedriusS:
Thanks for this! I haven't checked the code yet and I have played a bit with this locally. The CI seems to fail and there is a problem if you start a bare Thanos Query like this:

```bash
./thanos query
```

Then going to http://localhost:10902/new/graph?g0.expr=&g0.tab=1&g0.stacked=0&g0.range_input=1h and clicking 'Graph' leads to a blank page. What could be the problem?

I'd recommend setting up your editor so that it would run `goimports` & all of the other linters automatically on save. It will save you a lot of headaches in the long-term. Let me know if we could help you somehow.

prmsrswt:
> Then going to http://localhost:10902/new/graph?g0.expr=&g0.tab=1&g0.stacked=0&g0.range_input=1h and clicking 'Graph' leads to a blank page. What could be the problem?


There is a lack of error handling in the code. When you click on the Graph tab a `fetch` request is fired for endpoint `/api/v1/label/__name__/values` and the results are used without checking if it is `null`.

In the production build of React, it shows a blank page if an irrecoverable error has occurred, that's why we get the blank page. You can view the actual error in development build (run `./thanos query` and then in `pkg/ui/react-app` run `yarn install && yarn start`).

I didn't add the error handling because for now, I wanted to keep it as close to upstream as possible.

Currently, you need have some metrics for it to work. For development I ran `prometheus` scraping itself locally, a `thanos sidecar` and then `thanos query` using the `sidecar` as a store.

> I'd recommend setting up your editor so that it would run goimports & all of the other linters automatically on save. 

Also, for the CI part, I am not getting why it failed. Maybe I am missing something but the imports look fine to me. I use VSCode and it always coughs up an error if am not using an import. I am getting trouble running the `make lint` command locally. I am working on a laptop with 8 gigs of RAM and whenever I run `make lint` it eats up all my RAM and gets OOM killed. Any idea what should I do here?

squat:
@prmsrswt not sure about lint, this works fine on my linux laptop as well as in CI with much less than 8gb ram. let's investigate a bit. Any idea what command it hangs on?

Also, we need to remove the merge commit (https://github.com/thanos-io/thanos/pull/2412/commits/9120f145b5495b4a8fcec5d5fbd0e3b242cde568) as it conflicts and makes it hard to review. Can you instead rebase your branch on master (`git rebase -i master`)?

prmsrswt:
It hangs on `golangci-lint`. I was able to run the linter after a fresh reboot with most of the RAM free.

I removed the merge commit and I have rebased it on the master. The CI is still failing because it can't find `yarn`, I guess that would be easy to solve. :)

pkg/query/api/v1.go
```diff
@@ -178,6 +183,8 @@ type queryData struct {
 	Warnings []error `json:"warnings,omitempty"`
 }
 
+type storesData map[string][]query.StoreStatus
```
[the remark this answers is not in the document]
prmsrswt: This change was made by Adrien so I can just guess. What I understand is `api.stores` returns a map with Store's Type as `key` and array of the store's statuses as it's `value`, while `api.storeSet.GetStoreStatus()` returns a flat array of statuses.

I guess this was done because it would make sense from the perspective of an api client, as the current approach would give the client a json object with Store's name as key and an array of status as it's value, which is easier to consume imo.

.gitignore
```diff
@@ -25,6 +25,9 @@ website/public/
 website/docs-pre-processed/
 !website/data
 
+# React build assets
+pkg/ui/static/react
```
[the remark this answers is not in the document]
prmsrswt: There's already a `.gitignore` file in the `pkg/ui/react-app` directory, which have `node_modules` and many other required patterns for a javascript project in it.

prmsrswt:
> Also worth noting that on Ubuntu (and Debian probably) the yarn package manager's executable is called yarnpkg, not yarn so some extra documentation on this would be nice.

Maybe I am wrong but the executable was always `yarn` in my case. Even in the official documentation it's called `yarn` and not `yarnpkg` https://classic.yarnpkg.com/en/docs/install/#debian-stable 

> Lastly, there are some compilation warnings of the React app:
Do we want to get rid of them before merging or converting into errors?

Yeah sure, I would make the required changes.

prmsrswt:
Thanks for the review :heart:

I think you are right about the dropdown. Also, I tried to keep this PR as close as to the original one. This would have been harder to review if there were any more changes.

Makefile
```diff
@@ -82,6 +82,12 @@ MINIO_SERVER            ?=$(GOBIN)/minio-$(MINIO_SERVER_VERSION)
 FAILLINT_VERSION        ?= v1.2.0
 FAILLINT                ?=$(GOBIN)/faillint-$(FAILLINT_VERSION)
 
+REACT_APP_PATH = pkg/ui/react-app
+REACT_APP_SOURCE_FILES = $(wildcard $(REACT_APP_PATH)/public/* $(REACT_APP_PATH)/src/* $(REACT_APP_PATH)/tsconfig.json)
+REACT_APP_OUTPUT_DIR = pkg/ui/static/react
+REACT_APP_NODE_MODULES_PATH = $(REACT_APP_PATH)/node_modules
+REACT_APP_NPM_LICENSES_TARBALL = "npm_licenses.tar.bz2"
```
squat, line 89: This seems to be unused, do we need it?

pkg/ui/react-app/public/manifest.json
```diff
@@ -0,0 +1,15 @@
+{
+  "short_name": "Thanos UI",
+  "name": "Thanos Server Web Interface",
```
squat, line 3: nit: we could make these
```json
{
    "short_name": "Thanos",
    "name": "Thanos web interface"
}

File: pkg/ui/react-app/src/App.css [hunk 5]
```diff
@@ -0,0 +1,253 @@
...
+.graph-controls .range-input input {
+  width: 50px;
+}
+
+.time-input input {
+  border-right: none;
+}
+
+.time-input {
+  width: 270px !important;
```
squat, line 126: Is there a good reason to use `!important` here?

pkg/ui/react-app/src/App.tsx
```diff
@@ -0,0 +1,38 @@
+import React, { FC } from 'react';
+import Navigation from './thanos/Navbar';
```
squat, line 2: shall we move this down to the local imports

pkg/ui/react-app/src/App.tsx
```diff
@@ -0,0 +1,38 @@
...
+import { Container } from 'reactstrap';
+
+import './App.css';
+import { Router, Redirect } from '@reach/router';
```
squat, line 6: and move this up?

pkg/ui/react-app/src/pages/graph/GraphControls.tsx
```diff
@@ -0,0 +1,150 @@
...
+import { Button, ButtonGroup, Form, InputGroup, InputGroupAddon, Input } from 'reactstrap';
+
+import { FontAwesomeIcon } from '@fortawesome/react-fontawesome';
+import { faPlus, faMinus, faChartArea, faChartLine } from '@fortawesome/free-solid-svg-icons';
+
+import TimeInput from './TimeInput';
+import { parseRange, formatRange } from '../../utils';
+
+interface GraphControlsProps {
+  range: number;
```
squat, line 11: should one of these be `startTime` instead?

pkg/ui/react-app/src/pages/graph/Panel.tsx
```diff
@@ -0,0 +1,322 @@
...
+  useDeduplication: boolean;
+  usePartialResponse: boolean;
+  pastQueries: string[];
+  metricNames: string[];
+  removePanel: () => void;
+  onExecuteQuery: (query: string) => void;
+}
+
+interface PanelState {
+  data: any; // TODO: Type data.
```
squat, line 29: can we type this explicitly?

pkg/ui/react-app/public/manifest.json
prmsrswt: sure!

pkg/ui/react-app/src/App.css
prmsrswt: This is taken from upstream as it is.

pkg/ui/react-app/src/App.tsx
prmsrswt: done

pkg/ui/react-app/src/App.tsx
prmsrswt: and done!

Makefile
prmsrswt: I guess not. I have removed it for now.

pkg/query/api/v1.go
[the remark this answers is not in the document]
GiedriusS: Yes, it's all good with the actual type but we can convert:

```golang
638     statuses := make(storesData)                                                                                                                                                                               
```

Into:

```golang
638     statuses := make(map[string][]query.StoreStatus)
```

There's no need to have that type alias since there is nothing else consuming it. I think it's a case of YAGNI: https://en.wikipedia.org/wiki/You_aren%27t_gonna_need_it. If we will need it then we can always add it back.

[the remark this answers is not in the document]
prmsrswt: Oh got it :+1: I'll update the PR.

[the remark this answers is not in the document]
GiedriusS: This was solved so marking as 'Resolved'.

.gitignore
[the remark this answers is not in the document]
GiedriusS: Thank you, haven't noticed it somehow originally. Marking as resolved :+1:

Makefile
GiedriusS: This has been removed so marking as resolved.

pkg/ui/react-app/src/pages/graph/GraphControls.tsx
GiedriusS: This has been moved as-is from https://github.com/prometheus/prometheus/blob/d996ba20ec9c7f1808823a047ed9d5ce96be3d8f/web/ui/react-app/src/pages/graph/GraphControls.tsx#L11 which I think is a good thing since it makes this PR easier to review. Could we make a small PR upstream to fix it there? I think this PR can go either way with this.

pkg/ui/react-app/src/pages/graph/Panel.tsx
GiedriusS: This has been moved as-is from https://github.com/prometheus/prometheus/blob/d996ba20ec9c7f1808823a047ed9d5ce96be3d8f/web/ui/react-app/src/pages/graph/Panel.tsx#L27 which I think is a good thing since it makes this PR easier to review. Could we make a small PR upstream to fix it there @prmsrswt? I think this PR can go forward either way with this.

pkg/ui/react-app/src/App.css
GiedriusS: Indeed: https://github.com/prometheus/prometheus/blob/d996ba20ec9c7f1808823a047ed9d5ce96be3d8f/web/ui/react-app/src/App.css#L126

GiedriusS:
Awesome work @prmsrswt! :+1: I will take a final look in a bit but AFAICT this can be merged. One thing for the future: could we please avoid force pushes and overriding last commits on huge PRs such as this? Because otherwise, it is hard to follow what has changed between one review and another. :smile: 
Also, I really think that we should keep this as close as possible to the Prometheus code for now because it's the initial import of the new UI. Could you please a small PR there with the suggestions that @squat wrote here?

Issue #270, opened 2018-04-06
Describe verify/repair commands


bwplotka:
- How to use it
- How to extend/modify it
- What is that for

domgreen:
@Bplotka can I close down this issue?

Bplotka:
Not really. We need to also describe all of the verify/repair "issues" we have: https://github.com/improbable-eng/thanos/blob/master/cmd/thanos/bucket.go#L25 but you can leave that to me. You did the major part though, thanks!

jabbrwcky:
Also the documentation should specify prerequisites for running verifier.

I stumbled across the log message `GLOBAL COMPACTOR SHOULD __NOT__ BE RUNNING ON THE SAME BUCKET` issued by the verifier, and was wondering if my setup is wrong.

I dug a little bit and found that this message is a blanket message output here: https://github.com/thanos-io/thanos/blob/7e11afe64af0c096743a3de8a594616abf52be45/pkg/verifier/verify.go#L49

There is no documentation or any easily identifiable issue that explains the why of this message.

I assume that the verifier *with repair enabled* should not run concurrently with the compactor, but this is just a guess.

I think that running the verifier read-only with a running compactor should not have negative impact, in which case it might be enough to show this message only if repairs are enabled.

In any case the documentation should refelect when to stop the compactor before running verify.

PR #2418, opened 2020-04-13
bucket verify: document that compactor should be disabled

belm0:

## Changes

added note to `bucket verify` docs about disabling the compactor when using `--repair`

### DIFF
--- a/docs/components/bucket.md
+++ b/docs/components/bucket.md
@@ -172,6 +172,8 @@ Example:
 $ thanos bucket verify --objstore.config-file="..."
 ```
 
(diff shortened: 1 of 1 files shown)


belm0:
this was mentioned in https://github.com/thanos-io/thanos/issues/270#issuecomment-572050744, which was unfortunately closed by stalebot

Regarding the "GLOBAL COMPACTOR SHOULD __NOT__ BE RUNNING ON THE SAME BUCKET" check, really I think it should only be made when the `--repair` option is used, and furthermore it should abort the verify run if the compactor is running.  Anyone second that?

PR #2429, opened 2020-04-14
Update coding-style-guide.md

bitfield:
Add link to 'script' library.


bwplotka:
Thanks for this, but I think Go is really easy to make script even without any 3rd party deps. However, your library is definitely useful for executing complex pieped one lines on UNIX. Not sure if it fits the coding style though here :thinking:

kakkoyun:
In addition to what @bwplotka said, the style guide includes the things that we already practised so far. As far as I can tell github.com/bitfield/script is not used anywhere in the codebase.

bitfield:
Understood. Note that this is not a change to the style guide, as such; I'm not suggesting you _require_ users to write their scripts with `script`. It's just a suggestion which may help people who would otherwise need to use bash.

bitfield:
> Is there anything you can do to help move it forward?

No; it needs a decision from the maintainers on whether to accept or reject the suggestion.

squat:
Thank you for the contribution :) but I think we will close this PR for now since the maintainers do not use this library and thus can't recommend it to new contributors in the style guide. We can revisit later if we see a need for it

PR #2433, opened 2020-04-14
docs/release-process: make shell command copyable

squat:
In general, I think it is easier for users of guides when shell commands
are listed without a preceeding `$`, otherwise the commands cannot be
directly copied and pasted into a terminal.


bwplotka:
Agree I always have to manually remove those ;p

Issue #2415, opened 2020-04-11
Agree on dependency update policy.


bwplotka:
We are a hard dependency for the Corex.

This means that also when we update or not our dependencies, it makes a huge difference for Cortex @pracucci @pstibrany Recently on the Cortex community call @pstibrany mentioned they cannot upgrade Prometheus because of Thanos still being on ~2.16. We dod that as want to upgrade all in once when we have ChunkIterator ready. Peter said it's not urgent but this makes me think we should have some procedure in place to:

1. Ensure that Cortex can upgrade certain dep (or downgrade?) to unblock themselves
1. Ensure that Thanos can upgrade or downgrade certain dep and not impacting Cortex.

Any ideas? @pracucci @gouthamve @pstibrany @brancz ? :hugs:

bwplotka:
I think it really comes down to communication, maybe would be nice to define some steps who to ask where etc...

bwplotka:
Looks like fluid communication is the way to go. (: Works so far!

PR #2434, opened 2020-04-14
proposal: Added proposal for new Thanos component: Thanos Frontend.

bwplotka:

### DIFF
--- a/docs/proposals/202004_embedd_cortex_frontend.md
+++ b/docs/proposals/202004_embedd_cortex_frontend.md
@@ -9,12 +9,13 @@ owner: bwplotka
 ### Related Tickets
 
 * Response caching: https://github.com/thanos-io/thanos/issues/1651
-* Moving query frontend to separate repo: https://github.com/cortexproject/cortex/issues/1672
+* Moving query frontend to separate repo: https://github.com/cortexproject/Cortex/issues/1672
+* Discussion about naming: https://cloud-native.slack.com/archives/CK5RSSC10/p1586939369171300
 
 ## Summary
 
-This proposal describes addition of a new Thanos command (component) into `cmd/thanos` called `frontend`.
-This component will literally import a certain version of Cortex [frontend package](https://github.com/cortexproject/cortex/tree/4410bed704e7d8f63418b02b328ddb93d99fad0b/pkg/querier/frontend).
+This proposal describes addition of a new Thanos command (component) into `cmd/thanos` called `query-frontend`
+This component will literally import a certain version of Cortex [frontend package](https://github.com/cortexproject/Cortex/tree/4410bed704e7d8f63418b02b328ddb93d99fad0b/pkg/querier/frontend).
 
 We will go through rationales, and potential alternatives.
 
@@ -31,8 +32,10 @@ to be deployed in front of Prometheus Query API in order to ensure:
 
 Since the nature of Cortex backend is really similar to Thanos, with exactly the same PromQL API, and long term capabilities, the caching
 work done for Cortex fits to Thanos. Given also our good collaboration in the past, it feels natural to reuse Cortex's code.
-We even started discussion to move it to separate repo, but there was no motivation towards this, for a good reason.
-At the end we were advertising to use cortex query frontend on production on top of Thanos and this works considerably well, with some
+We even started discussion to move it to separate repo, but there was no motivation towards this, since there is no issue on using
+the Cortex one, as Cortex is happy to take generalized contributions.
+
+At the end we were advertising to use Cortex query frontend on production on top of Thanos and this works considerably well, with some
 problems on edge cases and for downsampled data as mentioned [here](https://github.com/thanos-io/thanos/issues/1651).
 
 However, we realized recently that asking users to install suddenly Cortex component on top of Thanos system is extremely confusing:
@@ -43,25 +46,26 @@ have flags and different subcommand for each component.
 * There are many Cortex specific configuration items which can confuse Thanos user and increase complexity overall.
 * We have many ideas how to improve Cortex Query Frontend on top of Thanos, but adding Thanos specific configuration options will increase
 complexity on Cortex side as well.
-* Cortex has no good example or tutorial on how to use frontend either. We have only [Observatorium example](https://github.com/observatorium/configuration/blob/master/environments/openshift/manifests/observatorium-template.yaml#L515).
+* Cortex has no good example or tutorial on how to use frontend either. We have only [Observatorium example](https://github.com/observatorium/configuration/blob/5129a8beb9507f29aec05566ca9a0f2ad82bbf76/environments/openshift/manifests/observatorium-template.yaml#L515).
 
 All of this were causing confusion and questions like [this](https://cloud-native.slack.com/archives/CK5RSSC10/p1586504362400300?thread_ts=1586492170.387900&cid=CK5RSSC10).
 
-At the end we decided with Thanos and Cortex maintainers that the ultimately it would be awesome to create a new Thanos subcommand called `frontend`.
+At the end we decided with Thanos and Cortex maintainers that, ultimately, it would be awesome to create a new Thanos service called `query-frontend`.
 
 ## Use Cases
 
-* User can cache responses for query range
-* User can use the same configuration patterns as rest of Thanos components.
-* Thanos Developer does not need to write and maintain custom response cache logic.
+* User can cache responses for query range.
+* User can split query range queries.
+* User can rate limit and retry range queries.
 
 ## Goals of this design
 
 * Enable response caching that will easy to use for Thanos users.
 * Keep it extensible and scalable for future improvements like advanced query planning, queuing, rate limiting etc.
 * Reuse as much as possible between projects, contribute.
+* Use the same configuration patterns as rest of Thanos components.
 
-## No Goals
+## Non Goals
 
 * Create Thanos specific response caching from scratch.
 
@@ -82,23 +86,13 @@ This command will be placeholder for any query planning or queueing logic that w
 
 To make this happen we will propose a small refactor in Cortex code to avoid unnecessary package dependencies.
 
-### Open Questions
-
-* Is `thanos frontend` a valid name? I feel it's short and verbose enough.
-
-Other, considered options:
-* `loadbalance`, `lb`
-* `ingress`
-* `edge`
-* `planner`
-
 ### Alternatives
 
 #### Don't add anything, document Cortex query frontend and add examples of usage
 
 Unfortunately we tried this path already without success. Reasons were mentioned in [Motivation](202004_embedd_cortex_frontend.md#Motivation)
 
-#### Add response caching to Querier itself, in the same binary.
+#### Add response caching to Querier itself, in the same process.
 
 This will definitely simplify deployment if Querier would allow caching directly. However, this way is not really scalable.
 
@@ -120,11 +114,11 @@ Overall, [Reusing is caring](https://www.bwplotka.dev/2020/how-to-became-oss-mai
 ## Work Plan
 
 1. Refactor [IndexCacheConfig](https://github.com/thanos-io/thanos/blob/55cb8ca38b3539381dc6a781e637df15c694e50a/pkg/store/cache/factory.go#L32) to generic cache config so we can reuse.
-Make it implement cortex cache.Cache interface.
+Make it implement Cortex cache.Cache interface.
 1. Add necessary changes to Cortex frontend
     * Metric generalization (they are globals now).
     * Avoid unnecessary dependencies.
-1. Add `thanos frontend` subcommand.
+1. Add `thanos frontend` subcommand. Call it `query-frontend` in docs and in the communication.
 1. Add proper e2e test using cache.
 1. Document new subcommand
 1. Add to [kube-thanos](https://github.com/thanos-io/kube-thanos)
--- a/docs/proposals/202004_embedd_cortex_frontend.md
+++ b/docs/proposals/202004_embedd_cortex_frontend.md
@@ -1,0 +1,119 @@
+---
+title: Adding a New Thanos Component that Embeds Cortex Query Frontend
+type: proposal
+menu: proposals
+status: in-review
+owner: bwplotka
+---
+
+### Related Tickets
+
+* Response caching: https://github.com/thanos-io/thanos/issues/1651
+* Moving query frontend to separate repo: https://github.com/cortexproject/cortex/issues/1672
+
+## Summary
+
+This proposal describes addition of a new Thanos command (component) into `cmd/thanos` called `frontend`.
+This component will literally import a certain version of Cortex [frontend package](https://github.com/cortexproject/cortex/tree/4410bed704e7d8f63418b02b328ddb93d99fad0b/pkg/querier/frontend).
+
+We will go through rationales, and potential alternatives.
+
+## Motivation
+
+[Cortex Frontend](https://www.youtube.com/watch?v=eyBbImSDOrI&t=2s) was introduced by Tom in August 2019. It was designed
+to be deployed in front of Prometheus Query API in order to ensure:
+
+* Query split by time.
+* Query step alignment.
+* Query retry logic
+* Query limit logic
+* Query response cache in memory, Memcached or Redis.
+
+Since the nature of Cortex backend is really similar to Thanos, with exactly the same PromQL API, and long term capabilities, the caching
+work done for Cortex fits to Thanos. Given also our good collaboration in the past, it feels natural to reuse Cortex's code.
+We even started discussion to move it to separate repo, but there was no motivation towards this, for a good reason.
+At the end we were advertising to use cortex query frontend on production on top of Thanos and this works considerably well, with some
+problems on edge cases and for downsampled data as mentioned [here]()
+
+However, we realized recently that asking users to install suddenly Cortex component on top of Thanos system is extremely confusing:
+
+* Cortex has totally different way of configuring services. It requires to decide what module you have in single YAML file. Thanos in opposite
+have flags and different subcommand for each component.
+* Cortex has bit different way of configuring memcached, which is inconsistent with what we have in Thanos Stote Gateway.
+* There are many Cortex specific configuration items which can confuse Thanos user and increase complexity overall (e.g. )
+* We have many ideas how to improve Cortex Query Frontend on top of Thanos, but adding Thanos specific configuration options will increase
+complexity on Cortex side as well.
+* Cortex has no good example or tutorial on how to use frontend either. We have only [Observatorium example](https://github.com/observatorium/configuration/blob/master/environments/openshift/manifests/observatorium-template.yaml#L515).
+
+All of this were causing confusion and questions like [this](https://cloud-native.slack.com/archives/CK5RSSC10/p1586504362400300?thread_ts=1586492170.387900&cid=CK5RSSC10).
+
+At the end we decided with Thanos and Cortex maintainers that the ultimately it would be awesome to create a new Thanos subcommand called `frontend`.
+
+## Use Cases
+
+* User can cache responses for query range
+* User can use the same configuration patterns as rest of Thanos components.
+* Thanos Developer does not need to write and maintain custom response cache logic.
+
+## Goals of this design
+
+* Enable response caching that will easy to use for Thanos users.
+* Keep it extensible and scalable for future improvements like advanced query planning, queuing, rate limiting etc.
+* Reuse as much as possible between projects, contribute.
+
+## No Goals
+
+* Create Thanos specific response caching from scratch.
+
+## Proposal
+
+The idea is to create `thanos frontend` component that allows specifying following options:
+
+* `--query-range.split-interval`, `time.Duration`
+* `--query-range.max-retries-per-request`, `int`, default = `5`
+* `--query-range.disable-step-align`, `bool`
+* `--query-range.response-cache-ttl` `time.Duration` default = `1m`
+* `--query-range.response-cache-config(-file)` `pathorcontent` + [CacheConfig](https://github.com/thanos-io/thanos/blob/55cb8ca38b3539381dc6a781e637df15c694e50a/pkg/store/cache/factory.go#L32)
+
+We plan to have in-mem, fifo and memcached support for now. Cache config will be exactly the same as the one used for Store Gateway.
+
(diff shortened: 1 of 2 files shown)


docs/proposals/202004_embedd_cortex_frontend.md
```diff
@@ -0,0 +1,119 @@
...
+* `--query-range.max-retries-per-request`, `int`, default = `5`
+* `--query-range.disable-step-align`, `bool`
+* `--query-range.response-cache-ttl` `time.Duration` default = `1m`
+* `--query-range.response-cache-config(-file)` `pathorcontent` + [CacheConfig](https://github.com/thanos-io/thanos/blob/55cb8ca38b3539381dc6a781e637df15c694e50a/pkg/store/cache/factory.go#L32)
+
+We plan to have in-mem, fifo and memcached support for now. Cache config will be exactly the same as the one used for Store Gateway.
+
+### Open Questions
+
+* Is `thanos frontend` a valid name? I feel it's short and verbose enough.
```
brancz, line 82: I have a feeling that people new to the project are going to be confused about this. Why not `query-cache`?

docs/proposals/202004_embedd_cortex_frontend.md
```diff
@@ -0,0 +1,119 @@
...
+
+### Related Tickets
+
+* Response caching: https://github.com/thanos-io/thanos/issues/1651
+* Moving query frontend to separate repo: https://github.com/cortexproject/cortex/issues/1672
+
+## Summary
+
+This proposal describes addition of a new Thanos command (component) into `cmd/thanos` called `frontend`.
+This component will literally import a certain version of Cortex [frontend package](https://github.com/cortexproject/cortex/tree/4410bed704e7d8f63418b02b328ddb93d99fad0b/pkg/querier/frontend).
```
brancz, line 17: Entirely practical concern, will this make https://github.com/thanos-io/thanos/issues/2415 an even bigger pain? :thinking: 

It doesn't seem like it would be a problem for Thanos, but would Cortex even still be able to import Thanos?

bwplotka: Good point, so far cyclic modules are totally fine. As long as we don't introcude cyclil package dependency it should be fine :thinking: 

Anyway if we bump into any problem we will need  to reopen: https://github.com/cortexproject/cortex/issues/1672 (:

docs/proposals/202004_embedd_cortex_frontend.md
bwplotka: Hm, I disagree here. While we can discuss better ideas `query-cache` is particularly not perfect one. It clash with `query` so again huge confusion. Additonally this component introduces much more than cache. I would rather call it.. `lb`? or `loadbalance`? It's just a frontend proxy with features.

Maybe:
* `loadbalance`
* `lb`
* `ingress` 
(:

bwplotka: Also one-worder would be nice

docs/proposals/202004_embedd_cortex_frontend.md
```diff
@@ -0,0 +1,121 @@
...
+
+## Work Plan
+
+1. Refactor [IndexCacheConfig](https://github.com/thanos-io/thanos/blob/55cb8ca38b3539381dc6a781e637df15c694e50a/pkg/store/cache/factory.go#L32) to generic cache config so we can reuse.
+1. Add necessary changes to Cortex frontend
+    * Metric generalization (they are globals now).
+1. Add `thanos frontend` subcommand.
+1. Add proper e2e test using cache.
+1. Document new subcommand
+1. Add to [kube-thanos](https://github.com/thanos-io/kube-thanos0)
```
prmsrswt, line 117: Minor typo at the en of URL here. Should be `/kube-thanos` instead `/kube-thanos0` :)

docs/proposals/202004_embedd_cortex_frontend.md
pracucci: > As long as we don't introcude cyclil package dependency it should be fine

I'm not super sure there are no cyclic package dependencies. I think the problem is caused by indirect dependencies. Which packages would Thanos directly import? `pkg/querier/frontend` and `pkg/querier/queryrange`? The last one is quite problematic, because has many internal dependencies.

Moreover, recent developments in the query-frontend have introduced the sum sharding, so we would end up importing a bunch of logic related to Cortex chunks which Thanos doesn't really need.

I would suggest to work on a refactoring and generalisation to avoid this problem. Some Cortex-specific logic should be plugged (via interfaces) only in Cortex.

docs/proposals/202004_embedd_cortex_frontend.md
```diff
@@ -0,0 +1,121 @@
...
+
+* Query split by time.
+* Query step alignment.
+* Query retry logic
+* Query limit logic
+* Query response cache in memory, Memcached or Redis.
+
+Since the nature of Cortex backend is really similar to Thanos, with exactly the same PromQL API, and long term capabilities, the caching
+work done for Cortex fits to Thanos. Given also our good collaboration in the past, it feels natural to reuse Cortex's code.
+We even started discussion to move it to separate repo, but there was no motivation towards this, for a good reason.
```
pracucci, line 34: A separate package without a deep indirect dependency tree within Cortex could be a feasible intermediate solution.

docs/proposals/202004_embedd_cortex_frontend.md
```diff
@@ -0,0 +1,121 @@
...
+At the end we were advertising to use cortex query frontend on production on top of Thanos and this works considerably well, with some
+problems on edge cases and for downsampled data as mentioned [here]()
```
pracucci, line 36: Did you mean to link this https://github.com/thanos-io/thanos/issues/1651 ?

docs/proposals/202004_embedd_cortex_frontend.md
```diff
@@ -0,0 +1,121 @@
...
+
+However, we realized recently that asking users to install suddenly Cortex component on top of Thanos system is extremely confusing:
+
+* Cortex has totally different way of configuring services. It requires to decide what module you have in single YAML file. Thanos in opposite
+have flags and different subcommand for each component.
+* Cortex has bit different way of configuring memcached, which is inconsistent with what we have in Thanos Stote Gateway.
```
pracucci, line 42: Cortex memcached client now supports Thanos service discovery too. However, given we're going to vendor it I would advocate to use Thanos memcached client (we just need to make it implement the `cache.Cache` Cortex interface).

docs/proposals/202004_embedd_cortex_frontend.md
```diff
@@ -0,0 +1,121 @@
...
+* Create Thanos specific response caching from scratch.
+
+## Proposal
+
+The idea is to create `thanos frontend` component that allows specifying following options:
+
+* `--query-range.split-interval`, `time.Duration`
+* `--query-range.max-retries-per-request`, `int`, default = `5`
+* `--query-range.disable-step-align`, `bool`
+* `--query-range.response-cache-ttl` `time.Duration` default = `1m`
```
pracucci, line 75: Why only 1m default?

docs/proposals/202004_embedd_cortex_frontend.md
pracucci: My take:
- Frontend can be confused with "UI frontend" (as of Today, I wouldn't call it frontend in Cortex)
- `query-cache` looks reductive. Even if now the primary use case could be caching, we may end up putting in this component other read path ingress logic

> loadbalance
> lb

I think it's reductive too (because doesn't just do load balancing).

> ingress

Better option to me.

docs/proposals/202004_embedd_cortex_frontend.md
```diff
@@ -0,0 +1,121 @@
...
+I think this does not need to be explained. Response caching has proven to be not trivial. It's really amazing that we
+have opportunity to work towards something that works with experts in the field like @tomwilkie and others from Loki and Cortex Team.
+
+Overall, [Reusing is caring](https://www.bwplotka.dev/2020/how-to-became-oss-maintainer/#5-want-more-help-give-back-help-others).
+
+## Work Plan
+
+1. Refactor [IndexCacheConfig](https://github.com/thanos-io/thanos/blob/55cb8ca38b3539381dc6a781e637df15c694e50a/pkg/store/cache/factory.go#L32) to generic cache config so we can reuse.
+1. Add necessary changes to Cortex frontend
+    * Metric generalization (they are globals now).
```
pracucci, line 113: I recently moved the query-frontend metrics to `promauto.With(reg)` but:
1. Still have the `cortex_` prefix (we need to inject it from outside)
2. Some internal dependencies may have globals

_I volunteer for this refactoring once this proposal is approved_

docs/proposals/202004_embedd_cortex_frontend.md
bwplotka: Hm, yea - this might require more work then ):

docs/proposals/202004_embedd_cortex_frontend.md
bwplotka: :+1:

docs/proposals/202004_embedd_cortex_frontend.md
bwplotka: yup

docs/proposals/202004_embedd_cortex_frontend.md
brancz: > Also one-worder would be nice

Agreed, but if we can't find something better, I wouldn't mind it too much.

> query-cache looks reductive. Even if now the primary use case could be caching, we may end up putting in this component other read path ingress logic

Agreed.

Since a very common target platform is Kubernetes, I'm not too sure about ingress either. Maybe Cortex's naming of `query-frontend` isn't all that bad?

docs/proposals/202004_embedd_cortex_frontend.md
bwplotka: :+1:

bwplotka: Added to action plan

docs/proposals/202004_embedd_cortex_frontend.md
bwplotka: Messed up flags I meant --query-range.response-cache-max-freshness. Should be fixed now.

docs/proposals/202004_embedd_cortex_frontend.md
bwplotka: `ingress` looks good, just curious if it's not mistaken / confused by Kubernetes ingress ):

bwplotka: What about... `edge`?

bwplotka: `planner`?

docs/proposals/202004_embedd_cortex_frontend.md
bwplotka: :+1:

bwplotka: Thanks!

docs/proposals/202004_embedd_cortex_frontend.md
bwplotka: `pre-query`

bwplotka: naming is hard ):

bwplotka: Asked on slack for some opinions, let's look deeper. I don't like query-frontend. Still exacty same confusion with both UI and queried.

adrianmoisey: > Since a very common target platform is Kubernetes, I'm not too sure about ingress either. Maybe Cortex's naming of `query-frontend` isn't all that bad?

I like `query-frontend`. It seems descriptive and it matches what Cortex are doing.

docs/proposals/202004_embedd_cortex_frontend.md
squat: > for a good reason.

Maybe worth mentioning the reason (:

docs/proposals/202004_embedd_cortex_frontend.md
```diff
@@ -0,0 +1,134 @@
...
+
+However, we realized recently that asking users to install suddenly Cortex component on top of Thanos system is extremely confusing:
+
+* Cortex has totally different way of configuring services. It requires deciding what module you have in single YAML file. Thanos in opposite
+have flags and different subcommand for each component.
+* Cortex has bit different way of configuring memcached, which is inconsistent with what we have in Thanos Store Gateway.
+* There are many Cortex specific configuration items which can confuse Thanos user and increase complexity overall.
+* We have many ideas how to improve Cortex Query Frontend on top of Thanos, but adding Thanos specific configuration options will increase
+complexity on Cortex side as well.
+* Cortex has no good example or tutorial on how to use frontend either. We have only [Observatorium example](https://github.com/observatorium/configuration/blob/master/environments/openshift/manifests/observatorium-template.yaml#L515).
```
squat, line 46: probably good to grab a specific commit SHA here to avoid broken links

docs/proposals/202004_embedd_cortex_frontend.md
```diff
@@ -0,0 +1,134 @@
...
+
+All of this were causing confusion and questions like [this](https://cloud-native.slack.com/archives/CK5RSSC10/p1586504362400300?thread_ts=1586492170.387900&cid=CK5RSSC10).
+
+At the end we decided with Thanos and Cortex maintainers that the ultimately it would be awesome to create a new Thanos subcommand called `frontend`.
+
+## Use Cases
+
+* User can cache responses for query range
+* User can use the same configuration patterns as rest of Thanos components.
```
squat, line 55: This doesn't sound like a use-case to me but more like a goal of the design

docs/proposals/202004_embedd_cortex_frontend.md
```diff
@@ -0,0 +1,134 @@
...
+* Thanos Developer does not need to write and maintain custom response cache logic.
```
squat, line 56: Same here, this is not a use-case IMO

docs/proposals/202004_embedd_cortex_frontend.md
```diff
@@ -0,0 +1,134 @@
...
+
+## Goals of this design
+
+* Enable response caching that will easy to use for Thanos users.
+* Keep it extensible and scalable for future improvements like advanced query planning, queuing, rate limiting etc.
+* Reuse as much as possible between projects, contribute.
+
+## No Goals
```
squat, line 64: Nit: s/No Goals/Non-goals/

docs/proposals/202004_embedd_cortex_frontend.md
squat: `proxy`?

docs/proposals/202004_embedd_cortex_frontend.md
bwplotka: :+1:

docs/proposals/202004_embedd_cortex_frontend.md
```diff
@@ -0,0 +1,134 @@
...
+
+#### Add response caching to Querier itself, in the same binary.
+
+This will definitely simplify deployment if Querier would allow caching directly. However, this way is not really scalable.
+
+Furthermore, eventually frontend will be responsible for more than just caching. It is meant to do query planning like splitting or even
+advanced query parallelization (query sharding). This might mean future improvements in terms of query scheduling, queuing and retrying.
+This means that at some point we would need an ability to scale query part and caching/query planner totally separately.
+
+Last but not least splitting queries allows to perform request in parallel. Only if used in single binary we can achieve load balancing of those requests.
```
squat, line 109: > Only if used in single binary we can achieve load balancing of those requests.

Why is using a single binary the only way to achieve load balancing? This seems orthogonal

bwplotka: That's the point it's not - at least not easily. Imagine we have just Querier and caching inside. To me it means that caching can split query into 100 bits, and really just schedule all parts into same Querier since it's single binary... unless we try to coordinate with others somehow. That's my point. Maybe I can be more clear about that (:

docs/proposals/202004_embedd_cortex_frontend.md
```diff
@@ -0,0 +1,134 @@
...
+
+I think this does not need to be explained. Response caching has proven to be not trivial. It's really amazing that we
+have opportunity to work towards something that works with experts in the field like @tomwilkie and others from Loki and Cortex Team.
+
+Overall, [Reusing is caring](https://www.bwplotka.dev/2020/how-to-became-oss-maintainer/#5-want-more-help-give-back-help-others).
+
+## Work Plan
+
+1. Refactor [IndexCacheConfig](https://github.com/thanos-io/thanos/blob/55cb8ca38b3539381dc6a781e637df15c694e50a/pkg/store/cache/factory.go#L32) to generic cache config so we can reuse.
+Make it implement cortex cache.Cache interface.
```
squat, line 123: nit, this doc has lots of references to `cortex` in lowercase. It would be nice to do s/cortex/Cortex/g (excluding URLs etc)

docs/proposals/202004_embedd_cortex_frontend.md
squat: do you mean `same process`? Same binary has nothing to do with that IMO

docs/proposals/202004_embedd_cortex_frontend.md
bwplotka: Proxy might be very confusing as Querier is literally a proxy, nothing more ;p

bwplotka: Looks like majority on slack like consistency :heart:  so `query-frontend`: https://cloud-native.slack.com/archives/CK5RSSC10/p1586942580182000

Of course @brancz was right again (:

bwplotka: I propose to call in code `frontend` but call in docs everywhere query-frontend. Same as we do with store. (:

bwplotka: thoughts?

docs/proposals/202004_embedd_cortex_frontend.md
bwplotka: Added to work plan.

docs/proposals/202004_embedd_cortex_frontend.md
bwplotka: yea

bwplotka: sorry my bad

bwplotka:
All should be addressed @brancz @squat @pracucci

docs/proposals/202004_embedd_cortex_frontend.md
```diff
@@ -0,0 +1,128 @@
...
+Overall, [Reusing is caring](https://www.bwplotka.dev/2020/how-to-became-oss-maintainer/#5-want-more-help-give-back-help-others).
+
+## Work Plan
+
+1. Refactor [IndexCacheConfig](https://github.com/thanos-io/thanos/blob/55cb8ca38b3539381dc6a781e637df15c694e50a/pkg/store/cache/factory.go#L32) to generic cache config so we can reuse.
+Make it implement Cortex cache.Cache interface.
+1. Add necessary changes to Cortex frontend
+    * Metric generalization (they are globals now).
+    * Avoid unnecessary dependencies.
+1. Add `thanos frontend` subcommand. Call it `query-frontend` in docs and in the communication.
```
GiedriusS, line 121: I thought the command will be `thanos query-frontend` i.e. the same name? :confused:

docs/proposals/202004_embedd_cortex_frontend.md
```diff
@@ -0,0 +1,128 @@
...
+* Create Thanos specific response caching from scratch.
+
+## Proposal
+
+The idea is to create `thanos frontend` component that allows specifying following options:
+
+* `--query-range.split-interval`, `time.Duration`
+* `--query-range.max-retries-per-request`, `int`, default = `5`
+* `--query-range.disable-step-align`, `bool`
+* `--query-range.response-cache-ttl` `time.Duration`
```
GiedriusS, line 79: What will this option do? Can't find it in Cortex's query-frontend docs (https://cortexmetrics.io/docs/configuration/prometheus-frontend/, https://cortexmetrics.io/docs/configuration/arguments/, https://cortexmetrics.io/docs/configuration/configuration-file/)

docs/proposals/202004_embedd_cortex_frontend.md
bwplotka: Please read carefully the previous discussions: https://github.com/thanos-io/thanos/pull/2434#discussion_r409016044

docs/proposals/202004_embedd_cortex_frontend.md
bwplotka: Validity. It will be well described in help.

docs/proposals/202004_embedd_cortex_frontend.md
pracucci: > I propose to call in code frontend

I'm OK calling it frontend in the code, but a bit dubious about calling the cmd `thanos frontend` (I understand this is the same trade-off picked for `store`, but at the beginning I found it confusing too). I would personally pick clarity on the cmd, so `thanos query-frontend` (given this is the naming we're going to use in the doc), at the cost of having to type 6 more characters 😉

brancz: I think the whole discussion here was about the command, so I still think the command should be query-frontend.

docs/proposals/202004_embedd_cortex_frontend.md
brancz: I was actually under the same impression

bwplotka: Yea let's discuss. I think this what we did with `store-gateway`, so wanted to be consistent. Happy to change

bwplotka: Talked offline, let's move to `query-frontend` subcommand as well. No harm, but only can help.

bwplotka:
Thanks all for rapid reviews :tada:

PR #2025, opened 2020-01-22
Use latest golang 1.13 docker image available

sylr:
* [ ] I added CHANGELOG entry for this change.
* [ ] Change is not relevant to the end user.

PR #2440, opened 2020-04-15
.circleci: use consistent ci image tags

squat:
We were not using the latest thanos-ci image tag for every part of the
CI pipeline: we were using 0.3.0 for tests but 0.2.0 for all builds.

fixes: https://github.com/thanos-io/thanos/pull/2025#issuecomment-613969516


squat:
ah very cool idea, let me go ahead and do that :)))

PR #2446, opened 2020-04-16
Fixed memcached client metrics initialization

pracucci:

## Changes

- Fixed memcached client metrics initialization


squat:
@pracucci by "fixing initialization" we mean we are pre-initializing the label map?

pracucci:
> @pracucci by "fixing initialization" we mean we are pre-initializing the label map?

@squat Correct, so that if the memcached client is used its metrics will be exported even before the 1st request. We kept the same approach in other places, ie:
https://github.com/thanos-io/thanos/blob/c77d7ca57f6b444539e3fd466951008960740f67/pkg/store/cache/inmemory.go#L99-L100

PR #2457, opened 2020-04-17
Cache deletion mark files together with meta files.

pstibrany:
## Changes

This PR adds caching of deletion mark files, similar to how meta.json files are cached. Deletion mark files are stored at the same place.

In order to make this work, deletion mark files are fetched at the same time as `meta.json` files. They are fetched for each block, stored in a map and on the disk (if they exist), and then passed to filters.

Open questions:
- is this a viable approach? 
- do we want to avoid fetching deletion mark files, if no filter uses them? We can do that for example by adding `RequiresDeletionMarks() bool` function to `Filter` interface, to figure out if we need them, and passing the OR-ed result from all filters to `BaseFetcher`.

I personally think this idea can be extended so that Fetcher returns mark files as well. Compactor could use that information, instead of relying on `ignoreDeletionMarkFilter` like it does today. But that would be a different PR.

## Verification

Not tested yet, and unit tests are broken. I will continue working on tests once I know this is a valid approach.

### DIFF
--- a/pkg/block/fetcher.go
+++ b/pkg/block/fetcher.go
@@ -75,7 +75,7 @@ const (
 	// Modified label values.
 	replicaRemovedMeta = "replica-label-removed"
 
-	// Default values for caching deletion marks in memory
+	// Default values for caching deletion marks in memory.
 	defaultDeletionMarkPositiveCacheEntryTTL = 1 * time.Hour
 	defaultDeletionMarkNegativeCacheEntryTTL = 5 * time.Minute
 )
@@ -156,8 +156,8 @@ type BaseFetcher struct {
 	concurrency int
 	bkt         objstore.InstrumentedBucketReader
 
-	// How long to cache deletion mark cache entries (positive -- when deletion mark exists, and negative -- when it doesn't exist)
-	// Note that real TTL is computed as: TTL/2 + random(TTL)
+	// How long to cache deletion mark cache entries (positive -- when deletion mark exists, and negative -- when it doesn't exist).
+	// Note that real TTL is computed as: TTL/2 + random(TTL).
 	deletionMarkPositiveCacheEntryTTL time.Duration
 	deletionMarkNegativeCacheEntryTTL time.Duration
 
--- a/pkg/block/fetcher.go
+++ b/pkg/block/fetcher.go
@@ -7,6 +7,7 @@ import (
 	"context"
 	"encoding/json"
 	"io/ioutil"
+	"math/rand"
 	"os"
 	"path"
 	"path/filepath"
@@ -26,12 +27,13 @@ import (
 	"github.com/prometheus/prometheus/tsdb"
 	tsdberrors "github.com/prometheus/prometheus/tsdb/errors"
 	"github.com/prometheus/prometheus/tsdb/fileutil"
+	"golang.org/x/sync/errgroup"
+
 	"github.com/thanos-io/thanos/pkg/block/metadata"
 	"github.com/thanos-io/thanos/pkg/extprom"
 	"github.com/thanos-io/thanos/pkg/model"
 	"github.com/thanos-io/thanos/pkg/objstore"
 	"github.com/thanos-io/thanos/pkg/runutil"
-	"golang.org/x/sync/errgroup"
 )
 
 type fetcherMetrics struct {
@@ -137,8 +139,6 @@ type MetadataModifier interface {
 	Modify(ctx context.Context, metas map[ulid.ULID]*metadata.Meta, modified *extprom.TxGaugeVec, incompleteView bool) error
 }
 
-const deletionMarkRecheckTime = 1 * time.Hour
-
 type deletionMark struct {
 	nextCheck time.Time
 	mark      metadata.DeletionMark
@@ -150,6 +150,8 @@ type BaseFetcher struct {
 	logger      log.Logger
 	concurrency int
 	bkt         objstore.InstrumentedBucketReader
+
+	deletionMarkRecheckTime time.Duration
 
 	// Optional local directory to cache meta.json and deletion mark files.
 	cacheDir string
@@ -184,6 +186,7 @@ func NewBaseFetcher(logger log.Logger, concurrency int, bkt objstore.Instrumente
 			Name:      "base_syncs_total",
 			Help:      "Total blocks metadata synchronization attempts by base Fetcher",
 		}),
+		deletionMarkRecheckTime: 1 * time.Hour, // TODO: configurable?
 	}, nil
 }
 
@@ -308,7 +311,7 @@ func (f *BaseFetcher) loadDeletionMark(ctx context.Context, id ulid.ULID, now ti
 		m, err := metadata.ReadDeletionMarkFromLocalDir(cachedBlockDir)
 		if err == nil {
 			return &deletionMark{
-				nextCheck: now.Add(deletionMarkRecheckTime),
+				nextCheck: now.Add(f.deletionMarkRecheckTime/2 + time.Duration(rand.Int63n(f.deletionMarkRecheckTime.Nanoseconds()))),
 				mark:      *m,
 			}, nil
 		}
@@ -337,7 +340,7 @@ func (f *BaseFetcher) loadDeletionMark(ctx context.Context, id ulid.ULID, now ti
 		}
 	}
 	return &deletionMark{
-		nextCheck: now.Add(deletionMarkRecheckTime),
+		nextCheck: now.Add(f.deletionMarkRecheckTime/2 + time.Duration(rand.Int63n(f.deletionMarkRecheckTime.Nanoseconds()))),
 		mark:      *m,
 	}, nil
 }
--- a/pkg/block/fetcher.go
+++ b/pkg/block/fetcher.go
@@ -76,7 +76,7 @@ const (
 	replicaRemovedMeta = "replica-label-removed"
 
 	// Default value for caching deletion marks in memory.
-	defaultDeletionMarkPositiveCacheEntryTTL = 1 * time.Hour
+	defaultDeletionMarkCacheEntryTTL = 1 * time.Hour
 )
 
 func newFetcherMetrics(reg prometheus.Registerer) *fetcherMetrics {
@@ -155,8 +155,8 @@ type BaseFetcher struct {
 	bkt         objstore.InstrumentedBucketReader
 
 	// How long to cache deletion mark cache entries.
-	// Note that next check time is computed as: TTL/2 + random(TTL).
-	deletionMarkPositiveCacheEntryTTL time.Duration
+	// Note that next check time is computed as: TTL/2 + random(TTL), so this is an average TTL, not max.
+	deletionMarkCacheEntryTTL time.Duration
 
 	// Optional local directory to cache meta.json and deletion mark files.
 	cacheDir string
@@ -192,7 +192,7 @@ func NewBaseFetcher(logger log.Logger, concurrency int, bkt objstore.Instrumente
 			Help:      "Total blocks metadata synchronization attempts by base Fetcher",
 		}),
 		// TODO: configurable?
-		deletionMarkPositiveCacheEntryTTL: defaultDeletionMarkPositiveCacheEntryTTL,
+		deletionMarkCacheEntryTTL: defaultDeletionMarkCacheEntryTTL,
 	}, nil
 }
 
@@ -292,7 +292,7 @@ func (f *BaseFetcher) loadMeta(ctx context.Context, id ulid.ULID) (*metadata.Met
 }
 
 func (f *BaseFetcher) newCachedDeletionMark(m metadata.DeletionMark, now time.Time) *cachedDeletionMark {
-	ttl := f.deletionMarkPositiveCacheEntryTTL
+	ttl := f.deletionMarkCacheEntryTTL
 
 	return &cachedDeletionMark{
 		nextCheck: now.Add(ttl/2 + time.Duration(rand.Int63n(ttl.Nanoseconds()))),
@@ -387,7 +387,8 @@ func (r response) metasCopy() map[ulid.ULID]*metadata.Meta {
 	return metas
 }
 
-func (r response) filterMarksCopy() map[ulid.ULID]*metadata.DeletionMark {
+// deletionMarksCopyForFilter makes a copy of deletion marks map, suitable for passing to Filter method.
+func (r response) deletionMarksCopyForFilter() map[ulid.ULID]*metadata.DeletionMark {
 	marks := make(map[ulid.ULID]*metadata.DeletionMark, len(r.marks))
 	for id, m := range r.marks {
 		marks[id] = &m.mark
@@ -444,7 +445,7 @@ func (f *BaseFetcher) fetchMetadata(ctx context.Context, now time.Time) (respons
 						resp.marks[id] = mark
 					}
 					if markErr != nil {
-						resp.metaErrs.Add(metaErr)
+						resp.metaErrs.Add(markErr)
 					}
 				}()
 			}
@@ -537,7 +538,7 @@ func (f *BaseFetcher) fetch(ctx context.Context, metrics *fetcherMetrics, filter
 
 	// Copy as same response might be reused by different goroutines.
 	metas := resp.metasCopy()
-	marks := resp.filterMarksCopy()
+	marks := resp.deletionMarksCopyForFilter()
 
 	metrics.synced.WithLabelValues(failedMeta).Set(float64(len(resp.metaErrs)))
 	metrics.synced.WithLabelValues(noMeta).Set(resp.noMetas)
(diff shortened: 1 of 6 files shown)


pkg/block/fetcher.go
```diff
@@ -130,7 +130,7 @@ type MetadataFetcher interface {
 }
 
 type MetadataFilter interface {
-	Filter(ctx context.Context, metas map[ulid.ULID]*metadata.Meta, synced *extprom.TxGaugeVec, incompleteView bool) error
+	Filter(ctx context.Context, metas map[ulid.ULID]*metadata.Meta, markers map[ulid.ULID]*metadata.DeletionMark, synced *extprom.TxGaugeVec, incompleteView bool) error
```
bwplotka, line 133: do you think this is needed for the filter? Wonder if we make deletion mark first citizen in fetcher rather :thinking:

bwplotka:
As commented above, but also:

> do we want to avoid fetching deletion mark files, if no filter uses them? We can do that for example by adding RequiresDeletionMarks() bool function to Filter interface, to figure out if we need them, and passing the OR-ed result from all filters to BaseFetcher.

I would say we always have to fetch, that's why I would think rather deletion marks should be done before fitlers even

bwplotka:
Otherwise make sense, I guess once we see deletion mark we cache it and no longer check for it right?

bwplotka:
question is: Do we even need to delete something

pstibrany:
> I would say we always have to fetch, that's why I would think rather deletion marks should be done before fitlers even

That's exactly what the code does now (https://github.com/pstibrany/thanos/blob/3a2276aa182a918e888413c0cac3b088788a87fc/pkg/block/fetcher.go#L377).

However, there are some places in UI where no filters are used, and I was thinking that perhaps in those cases we don't need to fetch deletion mark files.

pstibrany:
> Otherwise make sense, I guess once we see deletion mark we cache it and no longer check for it right?

Similar to meta.json, we only check if it still exists on the storage, but don't read it anymore. If it does, we use local cache. If it doesn't, we report that mark file no longer exist (even if we have it locally).

https://github.com/pstibrany/thanos/blob/3a2276aa182a918e888413c0cac3b088788a87fc/pkg/block/fetcher.go#L280

pkg/block/fetcher.go
pstibrany: Only one filter uses them. Alternative I considered was to pass fetcher to the filter, but that seems weird. Since fetcher already has all the markers, it can as well just pass them to filters.

pstibrany:
> question is: Do we even need to delete some deletion mark to undo deletion

This would still work! `loadDeletionMark` verifies whether deletion mark file exist on remote storage, before reusing cached copy

pstibrany:
> This would still work! `loadDeletionMark` verifies whether deletion mark file exist on remote storage, before reusing cached copy

Which in retrospect doesn't save that much, since before this PR it's one operation to read the mark file, after this PR it's one operation to check if it exists (+ another one to read if, it it's not yet cached). 🤔

pkg/block/fetcher.go
bwplotka: We can start with that indeed then

bwplotka:
Exactly. I think with deletion marks we can go much further. We don't expect them to be changed ever. So we can cache once and always ignore such block. For readers they can ignore also meta.json  after ignore time (:

bwplotka:
As agree offline I will try to embed this code into fix for https://github.com/thanos-io/thanos/issues/2459

bwplotka:
Maybe in two steps

bwplotka:
Actually let's have quick fix for v0.12.1 maybe.

pstibrany:
`BaseFetcher` now caches both positive and negative (non-existant marks) results for loading deletion marks. Cached entries will be returned without any rechecking for some time (existing marks are returned for ~1h, non-existing marks for ~5m). Next check time is randomized to avoid rechecking all deletion marks at the same time.

Cache entry time-to-live durations are currently hardcoded, but can be made configurable if required.

pstibrany:
Back to draft... I've added caching of negative results too, but that has some consequences that I'd rather not tackle in this PR. I will remove that and then un-draft it.

bwplotka:
negative result?

pstibrany:
> negative result?

When deletion marker doesn’t exist on the remote storage, my PR was caching that info for a short while.

pstibrany:
> `BaseFetcher` now caches both positive and negative (non-existant marks) results for loading deletion marks. Cached entries will be returned without any rechecking for some time (existing marks are returned for ~1h, non-existing marks for ~5m). Next check time is randomized to avoid rechecking all deletion marks at the same time.
> 
> Cache entry time-to-live durations are currently hardcoded, but can be made configurable if required.

`BaseFetcher` now only caches positive results for ~1h. Next check time is still randomized, but TTL is hard-coded.

pkg/block/fetcher.go
```diff
@@ -353,19 +472,16 @@ func (f *BaseFetcher) fetchMetadata(ctx context.Context) (interface{}, error) {
 	})
 
 	if err := eg.Wait(); err != nil {
-		return nil, errors.Wrap(err, "BaseFetcher: iter bucket")
+		return resp, errors.Wrap(err, "BaseFetcher: iter bucket")
```
pracucci, line 475: Could you explain me this change, please?

pkg/block/fetcher.go
```diff
@@ -72,6 +74,9 @@ const (
 
 	// Modified label values.
 	replicaRemovedMeta = "replica-label-removed"
+
+	// Default value for caching deletion marks in memory.
+	defaultDeletionMarkPositiveCacheEntryTTL = 1 * time.Hour
```
pracucci, line 79: I would remove `Positive` from naming here.

pkg/block/fetcher.go
```diff
@@ -274,8 +291,84 @@ func (f *BaseFetcher) loadMeta(ctx context.Context, id ulid.ULID) (*metadata.Met
 	return m, nil
 }
 
+func (f *BaseFetcher) newCachedDeletionMark(m metadata.DeletionMark, now time.Time) *cachedDeletionMark {
+	ttl := f.deletionMarkPositiveCacheEntryTTL
+
+	return &cachedDeletionMark{
+		nextCheck: now.Add(ttl/2 + time.Duration(rand.Int63n(ttl.Nanoseconds()))),
```
pracucci, line 298: So the effective one is between ttl/2 and ttl*1.5. Is it what we want right? I would have expected `deletionMarkPositiveCacheEntryTTL` to be the upper bound, not the average TTL, but I don't have strong opinions as far as we document it.

pkg/block/fetcher.go
```diff
@@ -301,33 +411,42 @@ func (f *BaseFetcher) fetchMetadata(ctx context.Context) (interface{}, error) {
...
+					default:
+						resp.metaErrs.Add(metaErr)
+					}
+
+					// Handle deletion mark.
+					if mark != nil {
+						resp.marks[id] = mark
+					}
+					if markErr != nil {
+						resp.metaErrs.Add(metaErr)
```
pracucci, line 447: You should add `markErr` not `metaErr`

pkg/block/fetcher.go
```diff
@@ -286,12 +379,29 @@ type response struct {
...
-func (f *BaseFetcher) fetchMetadata(ctx context.Context) (interface{}, error) {
+func (r response) metasCopy() map[ulid.ULID]*metadata.Meta {
+	metas := make(map[ulid.ULID]*metadata.Meta, len(r.metas))
+	for id, m := range r.metas {
+		metas[id] = m
+	}
+	return metas
+}
+
+func (r response) filterMarksCopy() map[ulid.ULID]*metadata.DeletionMark {
```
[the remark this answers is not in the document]
pstibrany: It's not a straightforward copy (that would return *cachedDeletionMark), but preparation for filter function, hence the name. I can improve the name.

pkg/block/fetcher.go
```diff
@@ -130,23 +135,33 @@ type MetadataFetcher interface {
...
+	Filter(ctx context.Context, metas map[ulid.ULID]*metadata.Meta, deletionMarks map[ulid.ULID]*metadata.DeletionMark, synced *extprom.TxGaugeVec, incompleteView bool) error
 }
 
 type MetadataModifier interface {
 	Modify(ctx context.Context, metas map[ulid.ULID]*metadata.Meta, modified *extprom.TxGaugeVec, incompleteView bool) error
 }
 
+type cachedDeletionMark struct {
+	nextCheck time.Time
+	mark      metadata.DeletionMark
```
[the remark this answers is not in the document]
pstibrany: It cannot be nil, and we don't need to modify it, and it's a small struct. But pointer would work too of course.

pkg/block/fetcher.go
```diff
@@ -469,15 +585,20 @@ type MetaFetcher struct {
 //
 // Returned error indicates a failure in fetching metadata. Returned meta can be assumed as correct, with some blocks missing.
 func (f *MetaFetcher) Fetch(ctx context.Context) (metas map[ulid.ULID]*metadata.Meta, partial map[ulid.ULID]error, err error) {
-	metas, partial, err = f.wrapped.fetch(ctx, f.metrics, f.filters, f.modifiers)
+	metas, _, partial, err = f.FetchWithMarks(ctx)
```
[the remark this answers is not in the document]
pstibrany: We have briefly discussed this earlier in the PR. One possibility would be to ask each filter if it needs deletion marks, and only fetch them if there is any such filter. But we concluded that we should always fetch deletion marks.

In the future, we could use these fetched marks in compactor, instead of relying on `ignoreDeletionMarkFilter`.

pkg/block/fetcher.go
pstibrany: `fetchMetadata` was returning `interface{}` previously, which could be nil, but now it returns `response` (not a pointer), so we need to return some value.

This change was done to make testing easier.

pkg/block/fetcher.go
pstibrany: I mostly wanted to avoid doing all rechecks at the same time, but to spread them a little. You're right that TTL could be made upper-bound instead, and then we can compute effective TTL for example as random between [TTL/2, TTL]. I also don't have strong opinions on this. I think the naming is bit confusing here.

pkg/block/fetcher.go
pstibrany: Oh, good catch!

pkg/block/metadata/deletionmark.go
```diff
@@ -74,3 +81,39 @@ func ReadDeletionMark(ctx context.Context, bkt objstore.InstrumentedBucketReader
...
+func WriteDeletionMarkToLocalDir(logger log.Logger, dir string, mark *DeletionMark) error {
+	data, err := json.Marshal(mark)
+	if err != nil {
+		return errors.Wrap(err, "json encode deletion mark")
+	}
+
+	p := filepath.Join(dir, DeletionMarkFilename)
+	tmp := p + ".tmp"
+
+	err = ioutil.WriteFile(tmp, data, 0600)
```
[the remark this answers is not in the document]
pstibrany: OK, sounds good. This is value before umask, so it's ok to use `0666`.

pkg/block/fetcher.go
```diff
@@ -469,15 +585,20 @@ type MetaFetcher struct {
...
+	return metas, partial, err
+}
+
+func (f *MetaFetcher) FetchWithMarks(ctx context.Context) (metas map[ulid.ULID]*metadata.Meta, marks map[ulid.ULID]*metadata.DeletionMark, partial map[ulid.ULID]error, err error) {
```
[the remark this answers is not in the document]
pstibrany: Fine with me. I consider this method to be temporary. I think `Filter` should return deletion marks too, but I didn't want to change it in this PR, as that would touch many other places. To avoid that I've instead introduced this method.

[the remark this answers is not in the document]
pstibrany: I will make it unexported, as it's only used by tests now, and add short comment.

pkg/block/fetcher.go
```diff
@@ -130,23 +135,33 @@ type MetadataFetcher interface {
...
 // Go-routine safe.
 type BaseFetcher struct {
 	logger      log.Logger
 	concurrency int
 	bkt         objstore.InstrumentedBucketReader
 
-	// Optional local directory to cache meta.json files.
+	// How long to cache deletion mark cache entries.
+	// Note that next check time is computed as: TTL/2 + random(TTL).
+	deletionMarkPositiveCacheEntryTTL time.Duration
```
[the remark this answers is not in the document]
pstibrany: `Positive` is a leftover, removed now.

pstibrany:
Thanks for review @pracucci, I've addressed your feedback.

pkg/block/fetcher.go
pracucci: I think could be made a bit more clear if we would have two hardcoded values: min TTL and max TTL. Then the actual TTL is a random between min and max.

pkg/block/fetcher.go
```diff
@@ -286,12 +379,30 @@ type response struct {
...
+		metas[id] = m
+	}
+	return metas
+}
+
+// deletionMarksCopyForFilter makes a copy of deletion marks map, suitable for passing to Filter method.
+func (r response) deletionMarksCopyForFilter() map[ulid.ULID]*metadata.DeletionMark {
+	marks := make(map[ulid.ULID]*metadata.DeletionMark, len(r.marks))
+	for id, m := range r.marks {
+		marks[id] = &m.mark
```
pracucci, line 394: The reason why I commented about storing by copy/reference in `cachedDeletionMark` is due to the factor that here we work with pointers, while in the cache we store by copy. It just looked weird to me, but I'm not feeling strong about this.

pkg/block/fetcher.go
```diff
@@ -301,33 +412,42 @@ func (f *BaseFetcher) fetchMetadata(ctx context.Context) (interface{}, error) {
...
-					continue
+				meta, metaErr := f.loadMeta(ctx, id)
+
+				var (
+					mark    *cachedDeletionMark
+					markErr error
+				)
+				if metaErr == nil {
+					mark, markErr = f.loadDeletionMark(ctx, id, now)
 				}
```
pracucci, line 423: Thinking loudly: the contract of this `fetchMetadata()` function is that it returns found metas also in case of any error fetching any meta (so it's OK returning a partial view). I'm wondering if it's correct to never fetch the deletion mark in case of any error: ie. should we fetch it anyway in case of `ErrorSyncMetaCorrupted`? I don't have an answer honestly, but I would like to have a discussion on it.

pkg/block/fetcher.go
```diff
@@ -353,19 +473,16 @@ func (f *BaseFetcher) fetchMetadata(ctx context.Context) (interface{}, error) {
...
 	}
 
 	// Only for complete view of blocks update the cache.
-	cached := make(map[ulid.ULID]*metadata.Meta, len(resp.metas))
-	for id, m := range resp.metas {
-		cached[id] = m
-	}
-	f.cached = cached
+	f.cached = resp.metasCopy()
+	f.marks = resp.marks // no need to copy, as it's not going to be modified
```
pracucci, line 485: Given deletion marks are relatively small (compared to metas), could just be more future-proof always copying them?

pkg/block/fetcher.go
```diff
@@ -286,12 +379,30 @@ type response struct {
...
+func (r response) metasCopy() map[ulid.ULID]*metadata.Meta {
+	metas := make(map[ulid.ULID]*metadata.Meta, len(r.metas))
+	for id, m := range r.metas {
+		metas[id] = m
+	}
+	return metas
+}
+
+// deletionMarksCopyForFilter makes a copy of deletion marks map, suitable for passing to Filter method.
+func (r response) deletionMarksCopyForFilter() map[ulid.ULID]*metadata.DeletionMark {
```
pracucci, line 391: Looks a bit weird to me calling it `ForFilter()`. The usage we're going to do with a copy of deletion marks shouldn't be reflected in the function name IMO.

pkg/block/fetcher_test.go
```diff
@@ -235,12 +236,46 @@ func TestMetaFetcher_Fetch(t *testing.T) {
...
+
+					// This mark will be ignored.
+					testutil.Ok(t, bkt.Upload(ctx, path.Join(ULID(3).String(), metadata.DeletionMarkFilename), bytes.NewBufferString("not a valid deletion-mark.json")))
+				},
+
+				// from previous tests...
+				expectedMetas:         ULIDs(1, 3, 6),
+				expectedCorruptedMeta: ULIDs(5),
+				expectedNoMeta:        ULIDs(4, 2),
+				expectedMetaErr:       errors.New("incomplete view: unexpected meta file: 00000000070000000000000000/meta.json version: 20"),
```
pracucci, line 268: Shouldn't we also expect an error for the corrupted deletion mark?

pstibrany:
We have deployed updated Cortex with this PR in today, and results are somewhat disappointing...

Screenshot 2020-04-22 at 10 00 55

If we focus on get and exists operations only, we can see that we now issue even more operations to the storage:

Screenshot 2020-04-22 at 10 02 20

It seems that reason for this is a combination of several factors:
- New store-gateway component in scanning all blocks, and uses metadata filter that removes "unwanted" blocks
- We now fetch deletion marks *before filters are run*. At the same time, in this PR we don't cache non-existant marks

Previously deletion marks were only fetched for blocks that passed "is this block interesting" filter, which resulted in fewer operations.

pstibrany:
In the light of these findings, let's not merge this yet. I'd like to write a design document to suggest more generic caching approach around storage.

PR #2463, opened 2020-04-18
add gofmt -s step to makefile and golangci

dvrkps:
Signed-off-by: Davor Kapsa <davor.kapsa@gmail.com>

### DIFF
--- a/.golangci.yml
+++ b/.golangci.yml
@@ -32,6 +32,7 @@ linters:
     - deadcode
     - errcheck
     - goconst
+    - gofmt
     - goimports
     - gosimple
     - govet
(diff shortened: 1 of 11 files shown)


pracucci:
> This is nice, but until we change [`Makefile format`](https://github.com/thanos-io/thanos/blob/58458f1b0a652b6ba7f98cd72dbe9272af9c7982/Makefile#L198) command which is run by default, this might be violated in next PR....

As a corollary, we should also enforce it via `golangci-lint`, adding the `gofmt` linter to `.golangci.yml`. What's your take @bwplotka ?

bwplotka:
Definitely!

bwplotka:
Can you help with this @dvrkps ?

dvrkps:
Sure, i will try.

PR #2464, opened 2020-04-18
cmd/thanos: clean gosimple S1039

dvrkps:
Signed-off-by: Davor Kapsa <davor.kapsa@gmail.com>


bwplotka:
Same as https://github.com/thanos-io/thanos/pull/2463#pullrequestreview-395926865 

It would be really amazing if we can add this check to `make lint` and validate on PR :hugs:  

WDYT @dvrkps ?

Issue #2459, opened 2020-04-17
v0.12.0 regression: Partial blocks are never deleted; spam in logs for double deletion mark.


bwplotka:
It was hard to track down, but looks like 4 unexpected things are happening on v0.12.0:

1. Partial blocks (no meta.json) are never cleaned up: 
![image](https://user-images.githubusercontent.com/6950331/79596563-b1a6a080-80d8-11ea-84c7-05849e10ae4b.png)

2. We have the spam of double mark deletion for partial upload delete attempt (for each partial block):

```
l=info ts=2020-04-17T16:50:14.305651352Z caller=clean.go:52 msg="deleted aborted partial upload" block=01E5Q7MP483QSDSX036X4TZ5SD thresholdAge=48h0m0s
level=info ts=2020-04-17T16:50:14.305659796Z caller=clean.go:47 msg="found partially uploaded block; marking for deletion" block=01E50RCWDZ5P83NRH4R0GXVGCC
level=warn ts=2020-04-17T16:50:14.312057891Z caller=block.go:140 msg="requested to mark for deletion, but file already exists; this should not happen; investigate" err="file 01E50RCWDZ5P83NRH4R0GXVGCC/deletion-mark.json already exists in bucket"
level=info ts=2020-04-17T16:50:14.312073215Z caller=clean.go:52 msg="deleted aborted partial upload" block=01E50RCWDZ5P83NRH4R0GXVGCC thresholdAge=48h0m0s
level=info ts=2020-04-17T16:50:14.312081259Z caller=clean.go:47 msg="found partially uploaded block; marking for deletion" block=01E5695C03VK6FB89G49T5MRB4
level=warn ts=2020-04-17T16:50:14.317892887Z caller=block.go:140 msg="requested to mark for deletion, but file already exists; this should not happen; investigate" err="file 01E5695C03VK6FB89G49T5MRB4/deletion-mark.json already exists in bucket"
level=info ts=2020-04-17T16:50:14.317907097Z caller=clean.go:52 msg="deleted aborted partial upload" block=01E5695C03VK6FB89G49T5MRB4 thresholdAge=48h0m0s
level=info ts=2020-04-17T16:50:14.317914967Z caller=clean.go:47 msg="found partially uploaded block; marking for deletion" block=01E5J2V84SKW9GYMH8B88M5NR9
```

3. We have spam of double mark deletions for garbage collection attempts (thousands+ as well):

```
[...]

Issue #2470, opened 2020-04-20
*: Use modified time to detect partial uploads.


bwplotka:
Problem:

Current flow for detecting partial blocks looks like this:

### Partial aborted upload

| Compactor                                                                       | Producer (e.g. sidecar, or previous compactor run)                 |
|---------------------------------------------------------------------------------|--------------------------------------------------------------------|
|                                                                                 | Uploads block, fails before meta.json was finished                 |
|                                                                                 | Producer is being removed (e.g downscale, cluster shutdown)        |
|                                                                                 | Partial block present on object storage (no / malformed meta.json) |
| Sync metas, sees partial block                                                  |                                                                    |
| What's the age of ULID? Is it older than PartialUploadThresholdAge? (2d) -> NO  |                                                                    |
| (2d passes)                                                                     |                                                                    |
| Sync metas, sees partial block                                                  |                                                                    |
| What's the age of ULID? Is it older than PartialUploadThresholdAge? (2d) -> YES |                                                                    |
| Block is deleted from object storage.                                           |                                                                    |

Our mechanism works **generally** but there are also cases it does not:

### Producer down for 2d

| Compactor                                                                       | Producer (e.g. sidecar, or previous compactor run)                                             |
|---------------------------------------------------------------------------------|------------------------------------------------------------------------------------------------|
|                                                                                 | Block was produced by Prometheus                                                               |
|                                                                                 | Uploads block, fails before meta.json was finished                                             |
|                                                                                 | Partial block present on object storage (no / malformed meta.json)                             |
|                                                                                 | Sidecar is down for some reason / Object storage connection is down                            |
| Sync metas, sees partial block                                                  |                                                                                                |
| What's the age of ULID? Is it older than PartialUploadThresholdAge? (2d) -> NO  |                                                                                                |
| (2d passes)                                                                     |                                                                                                |
|                                                                                 | All is back again, yay! Let's retry upload!                                                    |
| Sync metas, sees partial block                                                  |                                                                                                |
| What's the age of ULID? Is it older than PartialUploadThresholdAge? (2d) -> YES |                                                                                                |
| Block is deleted from object storage.                                           |                                                                                                |
|                                                                                 | Upload of missing meta.json succeeded, yay! Let's remove the local instance of this block. |
| WTH Block with just meta.json?                                                  | WTH Block with just meta.json?                                                                 |

[...]

PR #2472, opened 2020-04-20
Reverted addition of deletion mark for partial uploads.

bwplotka:
Fixes https://github.com/thanos-io/thanos/issues/2459 (quick fix).

This keeps the logic from the 0.11.0 which was good enough.

Some improvement for future: https://github.com/thanos-io/thanos/issues/2470

Let's merge this to master once 0.12.1 is released.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -13,8 +13,9 @@ We use *breaking* word for marking changes that are not backward compatible (rel
 
 ### Fixed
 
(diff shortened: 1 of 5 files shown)


squat:
This definitely needs a changelog entry to document both the fix and the behavior change :+1:

bwplotka:
ah yes, good point. Also tests

bwplotka:
I somehow thought we are still in release candidate process ;p

bwplotka:
Let's kick 0.12.1 ? @squat

Issue #2471, opened 2020-04-20
thanos store gateway: fatal error: concurrent map writes 


hawran:
**Thanos, Prometheus and Golang version used**:
thanos: 0.12.0
prometheus: 2.17.1
go: 1.13

**Object Storage Provider**:
ceph s3

**What happened**:
A couple of restarts during last two days because of internal errors.

**What you expected to happen**:
No such restarts.

**How to reproduce it (as minimally and precisely as possible)**:

**Full logs to relevant components**:
https://gist.github.com/hawran/f87f6b03aceac64553d9b2cd8e6c2532

**Anything else we need to know**:

bwplotka:
Ack, looks like a bug to ship to 0.12.1  @squat 

Will look on this after breakfast

squat:
+1 thanks for the report. will investigate

bwplotka:
Fixed on `release-0.12`, should ported to master soon :hugs: 

Thanks for quick report!

PR #2474, opened 2020-04-20
pkg/block/fetcher: fix concurrent map usage

squat:
Fixes: #2471

This commit fixes an issue where multiple goroutines in the block
fetcher filtering were concurrently accessing the same map. The
goroutines were concurrently writing AND reading to the shared metas
map. This commit guards this concurrent access by giving the
DeduplicateFilter struct a mutex.

### DIFF
--- a/pkg/block/fetcher.go
+++ b/pkg/block/fetcher.go
@@ -550,6 +550,7 @@ var _ MetadataFilter = &DeduplicateFilter{}
 // Not go-routine safe.
 type DeduplicateFilter struct {
 	duplicateIDs []ulid.ULID
+	mu           sync.Mutex
 }
 
 // NewDeduplicateFilter creates DeduplicateFilter.
@@ -603,11 +604,13 @@ func (f *DeduplicateFilter) filterForResolution(root *Node, metaSlice []*metadat
 
 	duplicateULIDs := getNonRootIDs(root)
 	for _, id := range duplicateULIDs {
+		f.mu.Lock()
 		if metas[id] != nil {
 			f.duplicateIDs = append(f.duplicateIDs, id)
 		}
 		synced.WithLabelValues(duplicateMeta).Inc()
 		delete(metas, id)
+		f.mu.Unlock()
 	}
 }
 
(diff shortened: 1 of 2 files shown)


pkg/block/fetcher.go
```diff
@@ -603,11 +604,13 @@ func (f *DeduplicateFilter) filterForResolution(root *Node, metaSlice []*metadat
 
 	duplicateULIDs := getNonRootIDs(root)
 	for _, id := range duplicateULIDs {
+		f.mu.Lock()
 		if metas[id] != nil {
 			f.duplicateIDs = append(f.duplicateIDs, id)
 		}
 		synced.WithLabelValues(duplicateMeta).Inc()
 		delete(metas, id)
+		f.mu.Unlock()
```
pracucci, line 613: This could be right before `synced.WithLabelValues(duplicateMeta).Inc()`. I think it would be more clear the intention, given we're not trying to synchronise on `metas`, just on `duplicateIDs`.

bwplotka:
sorry for quick merge @pracucci

pkg/block/fetcher.go
squat: hmm I am not sure I agree. The map that we need to synchronize is `metas`. `duplicateIDs`  is a slice and is not the source of the panic `fatal error: concurrent map writes`

squat: Also, note that we need the lock around both the read AND the write of the map

pracucci: Sorry for my previous comment. I was wrong, you're right. All good!

khyatisoneji:
Thanks for the PR @squat . For my learning, can you please tell me why this issue is not present with other filters? According to the stacktrace posted in the issue, this is because of concurrent map writes, we are deleting from map in all filters. Shouldn't the mutex be present in all filters?

squat:
@khyatisoneji this is because the other filters are not concurrently accessing maps, i.e. they aren't spawning multiple goroutines to access a single map of metas

PR #2475, opened 2020-04-20
Remove optimizations for label=~".*" and label!~".*".

pstibrany:

## Changes

This PR removes shortcuts for `label=~".*"` and `label!~".*"` matchers. Unless default in Prometheus changes such that `.` matches newline as well, these optimizations are not correct. See https://github.com/prometheus/prometheus/pull/6996#issuecomment-616473253

## Verification

Only removes shortcuts for specific matchers. Nothing else is changed.


bwplotka:
):

bwplotka:
But thanks for the work anyway!

bwplotka:
Good point on slack channel -> We probably have to have this in 0.12.1 as well

bwplotka:
Will cherry pick the PR

bwplotka:
Ignore me - it was not released yet (:

PR #2478, opened 2020-04-20
store, compact: Validate relabel actions

### LINKED ISSUE #2468: Validate the sharding relabel action
As discussed [here](https://github.com/thanos-io/thanos/pull/2379#discussion_r404418937), currently [thanos sharding](https://thanos.io/sharding.md/#relabelling) only supports three relabel actions. However, we use the relabel process function from Prometheus, which accepts other relabel actions as well, but actually these unsupported actions don't do anything.

It would be nice to have a check/validation on the relabel config. If the action is not supported by thanos, we can tell users this is not supported and return error.

AC:
- Check the relabel config before the relabel process. If the configured action is not supported, tell the reason in the log and return error

ranjithkumar007:
Fixes #2468 
Signed-off-by: ranjithkumar007 <ranjith.dakshana2015@gmail.com>

## Changes
- Returned an error if the relabel config is not supported currently by Thanos.

## Verification
- Verified correctness by adding a unit test.

I am very new to Go(Wrote hello world in Go a few days back). Though this is a small PR, any suggestions regarding the code style would be really great.


yeya24:
Thanks for your contribution. Really nice work!
But I am wondering is it better to put the validation phase here? https://github.com/thanos-io/thanos/blob/209fd88cbb66385c6a7d239b7e2c179094a747ee/cmd/thanos/store.go#L356

ranjithkumar007:
> Thanks for your contribution. Really nice work!
> But I am wondering is it better to put the validation phase here?
> 
> https://github.com/thanos-io/thanos/blob/209fd88cbb66385c6a7d239b7e2c179094a747ee/cmd/thanos/store.go#L356

Thanks for the quick review. Yes, I think [`store.go`](https://github.com/thanos-io/thanos/blob/209fd88cbb66385c6a7d239b7e2c179094a747ee/cmd/thanos/store.go#L356) is a much better place for validation.

Actually, I came across [a piece of code](https://github.com/thanos-io/thanos/blob/209fd88cbb66385c6a7d239b7e2c179094a747ee/pkg/block/fetcher_test.go#L315) in `fetcher_test.go` where parsing from yaml is done directly inline. I think I should also update such instances with a call to `parseRelabelConfig`?

Also,  Is `store_test.go`(doesn't exist now) the right place for adding tests to `parseRelabelConfig`,?

ranjithkumar007:
ping @yeya24

yeya24:
Sorry for the late response. To me, I think it is better to add the test at `store_test.go` since we just need to test `parseRelabelConfig`. However, I am not sure whether `parseRelabelConfig` will be used in other places in the future, which we need to support all the relabel actions. Then we need to change that code again.

Maybe @bwplotka has a better idea. WDYT @ranjithkumar007 ?

ranjithkumar007:
> Sorry for the late response

No worries.

> To me, I think it is better to add the test at `store_test.go` since we just need to test `parseRelabelConfig`. However, I am not sure whether `parseRelabelConfig` will be used in other places in the future, which we need to support all the relabel actions. Then we need to change that code again.
> 
> Maybe @bwplotka has a better idea. WDYT @ranjithkumar007 ?

Yeah. I also think `store_test.go` might be the right way.
With a quick grep of usages of `yaml.Unmarshal`  across the codebase, I see there are lots of places where different types of configs are parsed. I am not sure if its a good idea but maybe we can have such parsing of configs at a single place with a param to decide on the type of config?

ranjithkumar007:
> To me, I think it is better to add the test at `store_test.go` since we just need to test `parseRelabelConfig`

On second thought, I think it might be better to move `parseRelabelConfig` to `pkg`(not sure which sub dir exactly, maybe `store`?) as this functionality is needed in places like [fetcher_test.go](https://github.com/thanos-io/thanos/blob/209fd88cbb66385c6a7d239b7e2c179094a747ee/pkg/block/fetcher_test.go#L315). If it is used in other places in future, this can be easily imported.
thoughts @yeya24 ?

pkg/block/fetcher.go
```diff
@@ -529,6 +537,15 @@ const blockIDLabel = "__block_id"
 // Filter filters out blocks that have no labels after relabelling of each block external (Thanos) labels.
 func (f *LabelShardedMetaFilter) Filter(_ context.Context, metas map[ulid.ULID]*metadata.Meta, synced *extprom.TxGaugeVec, _ bool) error {
 	var lbls labels.Labels
+	supportedActions := map[relabel.Action]struct{}{relabel.Keep: {}, relabel.Drop: {}, relabel.HashMod: {}}
+
+	for _, cfg := range f.relabelConfig {
+		if _, isPresent := supportedActions[cfg.Action]; !isPresent {
+			level.Error(f.logger).Log("msg", "unsupported relabel action", "action", cfg.Action)
```
bwplotka, line 544: I think this is too late for validation.. Why cannot we validate those during startup? (: And fail command totally (in `cmd/thanos/...`

ranjithkumar007: Thanks for the review.
Agreed, will update asap.

cmd/thanos/store_test.go
```diff
@@ -0,0 +1,27 @@
...
+      - cluster
+    `))
+	testutil.Ok(t, err)
+
+	_, err = parseRelabelConfig([]byte(`
+    - action: labelmap
+      regex: "A"
+    `))
+	testutil.NotOk(t, err)
+	testutil.Equals(t, "unsupported relabel action: labelmap", err.Error())
```
ranjithkumar007, line 26: Should I use table tests even for couple of tests?

pkg/block/fetcher.go
ranjithkumar007: Done.

cmd/thanos/store.go
```diff
@@ -359,6 +359,13 @@ func parseRelabelConfig(contentYaml []byte) ([]*relabel.Config, error) {
 	if err := yaml.Unmarshal(contentYaml, &relabelConfig); err != nil {
 		return nil, errors.Wrap(err, "parsing relabel configuration")
 	}
+	supportedActions := map[relabel.Action]struct{}{relabel.Keep: {}, relabel.Drop: {}, relabel.HashMod: {}}
+
+	for _, cfg := range relabelConfig {
+		if _, isPresent := supportedActions[cfg.Action]; !isPresent {
```
bwplotka, line 365: nit: We usually use `ok` as pattern here. Let's be consistent, although `isPresent` is really nice as well.
```suggestion
		if _, ok := supportedActions[cfg.Action]; !ok {
```

cmd/thanos/store_test.go
```diff
@@ -0,0 +1,27 @@
...
+
+package main
+
+import (
+	"testing"
+
+	"github.com/thanos-io/thanos/pkg/testutil"
+)
+
+func Test_ParseRelabelConfig(t *testing.T) {
```
bwplotka, line 12:
```suggestion
func TestStore_ParseRelabelConfig(t *testing.T) {
```

cmd/thanos/store_test.go
bwplotka: Thanks for asking. 

No, no need (:

ranjithkumar007:
> I think we need exactly the same thing in `compact.go`.

`compact.go` [uses](https://github.com/thanos-io/thanos/blob/master/cmd/thanos/compact.go#L281) the same `parseRelabelConfig` function. So, no change might be required for this.
Other places of unmarshalling relabel yaml I found are in test files: [fetcher_test.go](https://github.com/thanos-io/thanos/blob/5b64108a1e1954f6d3ce056df23da24aac6af841/pkg/block/fetcher_test.go#L315), [bucket_test.go](https://github.com/thanos-io/thanos/blob/67ca6935823810ccf73cbfb1dfbe94fd9100815e/pkg/store/bucket_test.go#L714)
But these are inside `pkg/block` and `pkg/store` respectively. 
So, what would be the best place for keeping `parseRelabelConfig`?

yeya24:
> > I think we need exactly the same thing in `compact.go`.
> 
> `compact.go` [uses](https://github.com/thanos-io/thanos/blob/master/cmd/thanos/compact.go#L281) the same `parseRelabelConfig` function. So, no change might be required for this.
> Other places of unmarshalling relabel yaml I found are in test files: [fetcher_test.go](https://github.com/thanos-io/thanos/blob/5b64108a1e1954f6d3ce056df23da24aac6af841/pkg/block/fetcher_test.go#L315), [bucket_test.go](https://github.com/thanos-io/thanos/blob/67ca6935823810ccf73cbfb1dfbe94fd9100815e/pkg/store/bucket_test.go#L714)
> But these are inside `pkg/block` and `pkg/store` respectively.
> So, what would be the best place for keeping `parseRelabelConfig`?

Nice. Sorry for the late response.

Maybe you can make the logic below a helper function and reuse it. WDYT?

```
	supportedActions := map[relabel.Action]struct{}{relabel.Keep: {}, relabel.Drop: {}, relabel.HashMod: {}}

	for _, cfg := range relabelConfig {
		if _, ok := supportedActions[cfg.Action]; !ok {
			return nil, errors.Errorf("unsupported relabel action: %v", cfg.Action)
		}
	}
```

yeya24:
> > I think we need exactly the same thing in `compact.go`.
> 
> `compact.go` [uses](https://github.com/thanos-io/thanos/blob/master/cmd/thanos/compact.go#L281) the same `parseRelabelConfig` function. So, no change might be required for this.
> Other places of unmarshalling relabel yaml I found are in test files: [fetcher_test.go](https://github.com/thanos-io/thanos/blob/5b64108a1e1954f6d3ce056df23da24aac6af841/pkg/block/fetcher_test.go#L315), [bucket_test.go](https://github.com/thanos-io/thanos/blob/67ca6935823810ccf73cbfb1dfbe94fd9100815e/pkg/store/bucket_test.go#L714)
> But these are inside `pkg/block` and `pkg/store` respectively.
> So, what would be the best place for keeping `parseRelabelConfig`?

Nice. Sorry for the late response. 

Looks very good. Regarding other places of unmarshalling relabel yaml, what about moving the `parseRelabelConfig` function to `pkg/block`, then updating  `fetcher_test.go, bucket_test.go` to use that function?

WDYT @ranjithkumar007

ranjithkumar007:
> what about moving the `parseRelabelConfig` function to `pkg/block`, then updating `fetcher_test.go, bucket_test.go` to use that function?

Okay, sounds good to me. Does that mean `parseRelabelConfig` should be moved to `pkg/block/fetcher.go`?

yeya24:
> > what about moving the `parseRelabelConfig` function to `pkg/block`, then updating `fetcher_test.go, bucket_test.go` to use that function?
> 
> Okay, sounds good to me. Does that mean `parseRelabelConfig` should be moved to `pkg/block/fetcher.go`?

Sounds good.

yeya24:
Any updates @ranjithkumar007 ?

ranjithkumar007:
> Any updates @ranjithkumar007 ?

@yeya24 sorry for not updating here. I was busy lately.

```
fetcher_test.go:379             unexpected error: parsing relabel configuration: yaml: line 8: found character that cannot start any token
```
I am not really sure why there is above error on CI. I haven't changed the yaml, just updated the parsing function.  any idea why this might be failing?

ranjithkumar007:
PTAL @yeya24 .
failing e2e tests don't look relevant to this PR.

bwplotka:
I restarted e2e tests, let's see.

ranjithkumar007:
Tests have passed :tada:

yeya24:
@ranjithkumar007 Hey, would you mind rebasing master? Sorry for the late, @bwplotka  @kakkoyun It would be great if you can have a chance to give :eye: on it

ranjithkumar007:
Updated 👍

PR #2485, opened 2020-04-21
Log: Add app and service fields

alvaroaleman:
These allow to filter for logs from Thanos as a whole or a specific
component.


brancz:
I agree. This is responsibility of the mechanism picking up the logs.

PR #2494, opened 2020-04-21
[WIP] Chunks caching

pstibrany:

## Changes

This PR adds initial support for chunks caching into bucketChunkReader.

In TSDB, “chunk” is a small part of bigger file called “segment” (also “chunks file”). [Segment files](https://github.com/prometheus/prometheus/blob/master/tsdb/docs/format/chunks.md) are stored in “chunks” directory inside blocks directory. Each segment file is max 512 MiB, and single file may contain many chunks. There is a limit to single chunk size, which is 16000 bytes. Each segment file has a numeric filename (starting from 1, but otherwise unused), and a sequence number (starting from 0, and is determined by listing all segment files in “chunks” directory, and sorting them).

Chunks can start anywhere in the file and are typically very small, even though block chunk reader currently always fetches 16 KB from chunk start position – simply because it doesn’t know chunk length before it reads the chunk from storage. From my quick check of single block on one of our clusters, this block covering 1-day of data was using chunks with 99th percentile chunk size being ~700 bytes, and max chunk size being 3740 bytes.

Requests for individual chunks are grouped together using `gapBasedPartitioner` (already exists before this PR). This partitioner is configured to merge multiple chunk requests if they are close to each other – where “close” is defined as being within 512 KB of the segment file.

While we could cache individual chunks, given how small they are, that seems impractical. We could also cache 512 KB of segment files, but the problem is that partitioner doesn’t do any alignment when it merges requests, which makes caching very difficult.

This PR introduces ChunksCache, which allows `bucketChunkReader` to store and fetch back parts of single segment file. To make caching work, these parts must be aligned to `part size` (also defined in the interface). This PR already has code for *using* the chunks cache, but there is no implementation of chunks cache yet :) 

Posting as a draft to gather feedback on design.

Missing parts:
* [ ] Add memcache-based cache implementation
* [ ] Unit tests
* [ ] Metrics
* [ ] Manual testing
* [ ] Integration testing (?)

## Verification

This PR is not yet finished. Cache only exists as an interface, and there are no cache implementations yet :)


pstibrany:
After discussion with @pracucci we would like to propose more generic caching solution in front of the object store. Closing this for now. May reopen later if other approach is not suited for this use case.

bwplotka:
We should be careful with this then. Generic caches already exists for
object storages.. maybe we should use that instead?

pracucci:
Even if the PR has been closed, let me share a quick comment just for the records.

> While we could cache individual chunks, given how small they are, that seems impractical.

Why impractical? Some thoughts:
1. We do it in Cortex, and works (but Cortex chunks contain 6h of samples each)
2. Memcached works better with small items
3. Memcached multi-get can absorb the overhead of single small gets

_Not against doing sub-object caching (which I think is interesting) but I think we shouldn't exclude per-chunk caching apriori._

pracucci:
On the other side, sub-object caching with fixed size (ie. 16KB) and offset alignment has some important advantages:
1. No (or very reduced) memory fragmentation in memcached
2. Fits better with the Thanos partitioner
3. Significantly less cache requests for heavy requests (it's true the overhead can be absorbed with multi-get but the less requests the better)

pstibrany:
> > While we could cache individual chunks, given how small they are, that seems impractical.
> 
> Why impractical? Some thoughts:
> 
> 1. We do it in Cortex, and works (but Cortex chunks contain 6h of samples each)
> 2. Memcached works better with small items
> 3. Memcached multi-get can absorb the overhead of single small gets

I will share my comments from internal slack discussion:

In block that covers 1d, half of the chunks are less than 190 bytes. 95% of chunks are less than 375 bytes. These chunks seem to be very tiny.  Thanos always fetches entire 16k, starting from chunk start offset. So it would be nice to keep that entire fetched 16K, and not just small chunk (beginning of that). Plus, if Thanos finds out that there are chunks within 512KB offsets, it fetches entire range [first chunk position .... last chunk position + 16K). If we only cached individual chunks, we would be throwing away lot of fetched data.

pracucci:
I agree we should give a try to sub-object caching first. Looks more promising than per-chunk caching.

Issue #2479, opened 2020-04-20
Expose queried block IDs from the BucketStore


pracucci:
In Cortex we recently introduced the `store-gateway` service ([proposal](https://cortexmetrics.io/docs/proposals/blocks-storage-sharding/)) which, similarly to Thanos, sits in front of the bucket and is used to shard blocks across a pool of `store-gateway` instances (we also support a replication factor for HA in the read path).

The next step on the Cortex side would be introducing a consistency check. Basically, due to the lack of a strong coordination there's no guarantee that the set of `store-gateway` instances picked by a querier effectively loaded the exact set of blocks the querier expects (ie. ring hash change could be not propagated yet, blocks could still be loading after a resharding, etc). For this reason, we would like the querier to double check the queried blocks through `store-gateway` which is a piece of information missing right now.

Since the `store-gateway` internally uses the `BucketStore` we would need to expose the list of queried block IDs from the `Series()` API. So far, we're importing Thanos protobuf in Cortex in order to guarantee the `Series()` API endpoint exposed by Cortex `store-gateway` is compatible with Thanos, but it's a soft requirement (even if personally desirable).

I personally see a couple of options:
1. Add `info` to `storepb.SeriesResponse` (in addition to `series` and `warnings`). The `info` would contain an entry with the list of queried block IDs. The `info` response wouldn't be included by default, but enabled though a new boolean flag in `SeriesRequest`. This option would allow Cortex to keep full API compatibility with Thanos. One downside is that only `BucketStore` would support it (OK for Cortex, maybe KO for Thanos).
2. Refactor `BucketStore.Series()` into `BucketStore.SeriesWithInfo(req, srv) (SeriesInfo, error)` (not exposed through the protobuf) and having `BucketStore.Series()` just calling `SeriesWithInfo()` and ignoring the returned info. This way Cortex `store-gateway` could directly call `BucketStore.SeriesWithInfo()` instead of `BucketStore.Series()`.

I've also considered using `Info()` but there are couple of downsides:
1. We would like the list of block IDs queried for a specific `Series()` request and not the entire list of block IDs loaded in the `BucketStore`
2. There would be no guarantee that the returned block IDs are the one loaded once `Series()` was called

Thoughts?

bwplotka:
Thanks for this, I am curious what others think ( @GiedriusS @brancz ?), but IMO:

[...]

PR #2500, opened 2020-04-22
Merge v0.12.1 into master

squat:
This PR merges the release-0.12 branch back into master to get all of the 0.12.1 changes in master.

cc @thanos-io/thanos-maintainers

PR #2501, opened 2020-04-22
Gracefully handle additional oneof fields in SeriesResponse

pracucci:

## Changes

Following up the discussion we had in #2479, in this PR I'm proposing to gracefully handle additional `oneof` fields we may add in the future to `SeriesResponse`, without assuming that it either contains warnings or series (may also contain new fields an older client doesn't known).

## Verification

(Already existing)  tests.

### DIFF
--- a/pkg/store/proxy.go
+++ b/pkg/store/proxy.go
@@ -434,11 +434,15 @@ func startStreamSeriesSet(
 				s.warnCh.send(storepb.NewWarnSeriesResponse(errors.New(w)))
 				continue
 			}
-			select {
-			case s.recvCh <- rr.r.GetSeries():
-			case <-ctx.Done():
-				s.handleErr(errors.Wrapf(ctx.Err(), "failed to receive any data from %s", s.name), done)
-				return
+
+			if series := rr.r.GetSeries(); series != nil {
+				select {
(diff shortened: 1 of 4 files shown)


pkg/store/proxy.go
```diff
@@ -434,11 +434,15 @@ func startStreamSeriesSet(
...
-				return
+
+			if series := rr.r.GetSeries(); series != nil {
+				select {
+				case s.recvCh <- series:
+				case <-ctx.Done():
+					s.handleErr(errors.Wrapf(ctx.Err(), "failed to receive any data from %s", s.name), done)
+					return
+				}
+				continue
```
bwplotka, line 445: do we need continue?

pracucci: No. It's useless as, technically, is now useless the `continue` in the "warnings branch" above. I've removed both for consistency, but if you have a different opinion I can reiterate on it.

pracucci:
> I think this is worth a CHANGELOG entry, but let's do on separate PR or during the release as it will create conflict with this #2500

I've rebased `master` and added it. PTAL if you think it's clear enough.

bwplotka:
Thanks!

PR #2502, opened 2020-04-22
Added hints support to store protobuf

pracucci:

## Changes

Following up the conversation in #2479, in this PR I'm proposing to introduce the `hints` field to the `SeriesResponse` protobuf definition along with a `hints_enabled` option in the `SeriesRequest`. It's currently not used by anyone, but this PR setup the foundation to allow specific store implementations to extend the `SeriesResponse` with hints about how series have been queried.

I've also took the change to remove the "Hacky hack" from `genproto.sh`.

## Verification

Existing tests.


pkg/store/storepb/rpc.proto
```diff
@@ -132,6 +138,13 @@ message SeriesResponse {
     /// warning is considered an information piece in place of series for warning purposes.
     /// It is used to warn query customer about suspicious cases or partial response (if enabled).
     string warning = 2;
+
+    /// hints in an opaque data structure that can be used to carry information about how the series
```
pstibrany, line 142: s/in/is/

pstibrany: I would say it simply additional information from store, without specifying what it is. If store wants to send current time in it, it can do that.

pracucci: Thanks @pstibrany for your feedback! Updated 👍

pkg/store/storepb/rpc.proto
```diff
@@ -114,6 +115,11 @@ message SeriesRequest {
 
   // skip_chunks controls whether sending chunks or not in series responses.
   bool skip_chunks = 8;
+
+  // hints_enabled controls whether sending hints or not in the series responses. Hints
```
bwplotka, line 119: Do we really need this? I would avoid extending API if possible. Especially bool options.

The alternative (which actually avoid lots of boilerplate) is to just add hidden flag for store GW to enable sending hints in the message. This will satisfy all needs and will be perfectly compatible without Store API noise

pracucci: I've updated it based on our offline discussion:
- Removed `hints_enabled` field from `SeriesRequest`
- We'll have an internal flag in `NewBucketStore()` in a subsequent PR (the one that will add hints support to bucket store)

PR #2504, opened 2020-04-22
Adjust memcached operation buckets

kakkoyun:
Depending on the memcached client configuration, particularly `MaxGetMultiBatchSize` and `MaxItemSize`, Memcached operation duration varies; hence this PR adjusts buckets to cover more latency classes.

https://github.com/thanos-io/thanos/blob/45a6bc4ed39e7f7b14f99fedbc62ac414b745bb4/pkg/cacheutil/memcached_client.go#L35-L45

cc @brancz @pracucci @pstibrany 

## Changes

* Adjust `thanos_memcached_operation_duration_seconds ` metric histogram buckets.

## Verification

* `make build`
* `MINIO_ENABLED=1 ./scripts/quickstart.sh`
* Check `curl localhost:10906/metrics`

### DIFF
--- a/pkg/cacheutil/memcached_client.go
+++ b/pkg/cacheutil/memcached_client.go
@@ -240,7 +240,7 @@ func newMemcachedClient(
 		Name:        "thanos_memcached_operation_duration_seconds",
 		Help:        "Duration of operations against memcached.",
 		ConstLabels: prometheus.Labels{"name": name},
-		Buckets:     []float64{0.001, 0.005, 0.01, 0.025, 0.05, 0.1, 0.2, 0.5, 1},
+		Buckets:     []float64{0.001, 0.005, 0.01, 0.025, 0.05, 0.1, 0.2, 0.5, 1, 3, 6, 10},
 	}, []string{"operation"})
 	c.duration.WithLabelValues(opGetMulti)
 	c.duration.WithLabelValues(opSet)


pkg/cacheutil/memcached_client.go
```diff
@@ -240,7 +240,7 @@ func newMemcachedClient(
 		Name:        "thanos_memcached_operation_duration_seconds",
 		Help:        "Duration of operations against memcached.",
 		ConstLabels: prometheus.Labels{"name": name},
-		Buckets:     []float64{0.001, 0.005, 0.01, 0.025, 0.05, 0.1, 0.2, 0.5, 1},
+		Buckets:     []float64{0.01, 0.1, 0.3, 0.6, 1, 3, 6, 9, 20, 30, 60, 90, 120, 240, 360, 720},
```
pracucci, line 243: 720 seconds? The duration is tracked for a single operation (does not include the "waiting time" for limited concurrency on the client side), so anything above 1s looks weird to me (I mean, could happen, but then there's something wrong in the setup/networking/memcached and it's not about the metric bucket). What's the rationale of setting such high buckets?

kakkoyun: I've sent the PR to start a discussion, of course, an issue would have been more appropriate, will do next time.

Anyway, although I might have gone overboard with  720 😅I think by just configuring batch size, one can easily exceed the 1s limit for `getmulti` operation. I'm happy to adjust them. (Even though sync happens in shorter intervals, compactor runs every 5m by default. So maybe 300 is a more natural limit for the worst-case scenario.) 

And with 3 sharded Thanos Stores and the following configuration,
```diff
"dns_provider_update_interval": "10s"
"max_async_buffer_size": 10000000
"max_async_concurrency": 100
+"max_get_multi_batch_size": 100
-"max_get_multi_batch_size": 1000
"max_get_multi_concurrency": 900
"max_idle_connections": 1000
"max_item_size": "1MiB"
"timeout": "10s"
```
we have observed high latencies so far. 

Screenshot 2020-04-22 18 40 21

We reduced batched size at a certain point, you can see the improvement on P50 graph

Screenshot 2020-04-22 18 42 26

pracucci: If you have a timeout of 10s, the operation should never take longer than 10s (+ a small delta). I personally see any timeout longer than 10s quite useless. In Cortex, in our large clusters we set a memcached timeout of <= 1s. If the cache is slow (for any reason) it's OK directly going to the backend store.

My take is:
- The largest bucket here should not go over 10s
- I would keep previously defined buckets and add few more buckets after 1s and until 10s (ie. 3, 5, 10)

kakkoyun: Good points! Thanks for the feedback. I will fix accordingly.

PR #2508, opened 2020-04-22
Dockerfile: Leveraging docker layer caching

ankitjain28may:
Reformatting `Dockerfile.muti-stage` to leverage the Docker Layer Caching, it reduces the build time up to 30% for recurring builds. First-time build will take the same time.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -22,6 +22,7 @@ We use *breaking* word for marking changes that are not backward compatible (rel
 ### Changed
 
 - [#2505](https://github.com/thanos.io/thanos/pull/2505) Store: remove obsolete `thanos_store_node_info` metric.
+- [#2508](https://github.com/thanos.io/thanos/pull/2508) Dockerfile: Leveraging docker layer caching.
 
 ## [v0.12.1](https://github.com/thanos-io/thanos/releases/tag/v0.12.1) - 2020.04.20
 
--- a/Dockerfile.multi-stage
+++ b/Dockerfile.multi-stage
@@ -1,9 +1,14 @@
 FROM golang:1.13.6-alpine3.11 as builder
 
-ADD . $GOPATH/src/github.com/thanos-io/thanos
 WORKDIR $GOPATH/src/github.com/thanos-io/thanos
-
+# Change in the docker context invalidates the cache so to leverage docker
+# layer caching, moving update and installing apk packages above COPY cmd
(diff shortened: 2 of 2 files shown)


Dockerfile.multi-stage
```diff
@@ -1,10 +1,11 @@
 FROM golang:1.13.6-alpine3.11 as builder
 
-ADD . $GOPATH/src/github.com/thanos-io/thanos
 WORKDIR $GOPATH/src/github.com/thanos-io/thanos
 
 RUN apk update && apk upgrade && apk add --no-cache alpine-sdk
 
+ADD . $GOPATH/src/github.com/thanos-io/thanos
```
povilasv, line 7: Consider adding a comment why would this improve docker layer caching, so that in future we know to leave it here.

ankitjain28may: Thanks for reviewing, I have added the comments.

Dockerfile.multi-stage
```diff
@@ -1,9 +1,14 @@
...
 WORKDIR $GOPATH/src/github.com/thanos-io/thanos
-
+# Change in the docker context invalidates the cache so to leverage docker
+# layer caching, moving update and installing apk packages above COPY cmd
+# More info https://docs.docker.com/develop/develop-images/dockerfile_best-practices/#leverage-build-cache 
 RUN apk update && apk upgrade && apk add --no-cache alpine-sdk
+# Replaced ADD with COPY as add is generally to download content form link or tar files
+# while COPY supports the basic copying of local files into the container.
+# https://docs.docker.com/develop/develop-images/dockerfile_best-practices/#add-or-copy
+COPY . .
```
GiedriusS, line 11: I presume we use `.` here because of `WORKDIR` above? Maybe it would be worth still keeping the whole path explicitly to make it more clear?

CHANGELOG.md
```diff
@@ -9,6 +9,16 @@ NOTE: As semantic versioning states all 0.y.z releases can contain breaking chan
 
 We use *breaking* word for marking changes that are not backward compatible (relates only to v0.y.z releases.)
 
+## Unreleased
```
GiedriusS, line 12: This section has been added in `master`, please rebase.

Dockerfile.multi-stage
ankitjain28may: Yes, because we have already added the `WORKDIR` above. Sure, I will explicitly add it to be more clear.

ankitjain28may:
@GiedriusS I have made the required changes :)

PR #2515, opened 2020-04-23
cmd: rule: do not wrap reload endpoint with '/'

### LINKED ISSUE #2514: rule: `/-/reload` is inaccessible on v0.12.1 when no prefix has been specified
Relevant code part:

https://github.com/thanos-io/thanos/blob/master/cmd/thanos/rule.go#L553-L564

I wrote this in kind of pseudo-code, I hope that's understandable.

The prefix after all of the wrapping that we use for API end-points is `/{USER_STRING}api/v1`. In the `/-/reload` case after all of the wrapping there, the prefix that we use when registering a route is `/{USER_STRING}`. Then, the `/-/reload` is registered on top of that prefix. The crux of the issue is that the prefix, in this case, ends with a `/` when no `USER_STRING` has been specified leading to a 404 when `/-/reload` is accessed.

Testing on `v0.12.1`:

```
$ curl --fail -X POST http://localhost:${HTTP_PORT}/-/reload
curl: (22) The requested URL returned error: 404 Not Found
```

With a simple change to:
```
		router.WithPrefix(webRoutePrefix).Post("-/reload", func(w http.ResponseWriter, r *http.Request) {
```

It works again:
```
{"caller":"rule.go:783","level":"info","msg":"reload rule files","numFiles":1,"ts":"2020-04-23T15:09:45.569659899Z"}
```

But this "fix" doesn't handle the case when the user passes us some string. If some string has been passed then the route needs to be _with_ the prefix `/`. If no string has been passed then it needs to be _without_ the prefix `/`. 

*Or* just remove this `router.WithPrefix(webRoutePrefix)` part on that line because it is already handled by us previously.

GiedriusS:
Do not wrap the router with `/` on the `/-/reload` endpoint. Otherwise,
it is inaccessible when no prefix has been specified by the user.

Fixes #2514.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -14,6 +14,7 @@ We use *breaking* word for marking changes that are not backward compatible (rel
 ### Fixed
 
 - [#2501](https://github.com/thanos-io/thanos/pull/2501) Query: gracefully handle additional fields in `SeriesResponse` protobuf message that may be added in the future.
+- [#2515](https://github.com/thanos-io/thanos/pull/2515) Rule: do not wrap reload endpoint with `/`. Makes `/-/reload` accessible again when no prefix has been specified.
 
 ### Added
 
--- a/cmd/thanos/rule.go
+++ b/cmd/thanos/rule.go
@@ -561,7 +561,7 @@ func runRule(
 			router = router.WithPrefix(webRoutePrefix)
 		}
 
-		router.WithPrefix(webRoutePrefix).Post("/-/reload", func(w http.ResponseWriter, r *http.Request) {
+		router.Post("/-/reload", func(w http.ResponseWriter, r *http.Request) {
 			reloadMsg := make(chan error)
 			reloadWebhandler <- reloadMsg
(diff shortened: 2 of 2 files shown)


simonpasquier:
LGTM. Shouldn't it be against `release-012` and then backported to `master`?

GiedriusS:
> LGTM. Shouldn't it be against `release-012` and then backported to `master`?

I haven't checked whether this bug was introduced in 0.12 before creating this, sorry. But either order should work albeit with more effort on the person who is doing the release. I'm fine with reopening it to a different branch. @squat care to take a look?

lilic:
Should we add a test for reload that it picks up new rules, wdyt?

squat:
@GiedriusS let's make this PR to the release-0.12 branch to we can cut a 0.12.2 with this fix in it, yeah?

bwplotka:
:+1: 

Some e2e test for rule reload would be nice indeed.

GiedriusS:
:+1: I will open up a PR against `release-0.12` with a simple test-case.

PR #2516, opened 2020-04-23
Added hints support to BucketStore.Series()

### LINKED ISSUE #2479: Expose queried block IDs from the BucketStore

pracucci:

## Changes

Following up the discussion in #2479, in this PR I'm adding hints support to `BucketStore.Series()`.

Highlights:
- Hints are enabled via feature flag (not exposed on the CLI because hints are currently unused by Thanos)
- The hints exposed by `BucketStore` are defined in a protobuf so that future versioning is easier
- The `Block` is a struct (protobuf message) instead of a `string` make it easier to extend in the future (if necessary)

Closes #2479

## Verification

Unit tests.


pkg/store/bucket.go
```diff
@@ -257,6 +259,9 @@ type BucketStore struct {
 	// When used with in-memory cache, memory usage should decrease overall, thanks to postings being smaller.
 	enablePostingsCompression   bool
 	postingOffsetsInMemSampling int
+
+	// Enables hits in the Series() response.
```
bwplotka, line 263: Let's not enable hits (: 


```suggestion
	// Enables hints in the Series() response.
```

pkg/store/bucket.go
```diff
@@ -891,6 +899,11 @@ func (s *BucketStore) Series(req *storepb.SeriesRequest, srv storepb.Store_Serie
 		for _, b := range blocks {
 			b := b
 
+			// Keep track of queried blocks.
+			if s.enableSeriesHints {
+				hints.AddQueriedBlock(b.meta.ULID.String())
```
bwplotka, line 904: Can we move the above comment here? This only happens inside `if` :hugs:

pkg/store/bucket.go
```diff
@@ -1000,6 +1013,21 @@ func (s *BucketStore) Series(req *storepb.SeriesRequest, srv storepb.Store_Serie
 
 		err = nil
 	})
+
+	if s.enableSeriesHints {
+		var anyHints *types.Any
```
bwplotka, line 1018: Helper might be nice for this in `custom.go` in `hintspb`

bwplotka: (marshal I mean)

pkg/store/bucket_e2e_test.go
```diff
@@ -169,6 +169,7 @@ func prepareStoreWithTestBlocks(t testing.TB, dir string, bkt objstore.Bucket, m
 		true,
 		true,
 		DefaultPostingOffsetInMemorySampling,
+		false,
```
bwplotka, line 172: I would just enable. This should work anyway, right? Thast's the point of compatibility (:

pkg/store/bucket_test.go
```diff
@@ -1617,3 +1631,106 @@ func TestSeries_OneBlock_InMemIndexCacheSegfault(t *testing.T) {
...
+		10000,
+		10,
+		false,
+		10,
+		nil,
+		false,
+		true,
+		true,
+		DefaultPostingOffsetInMemorySampling,
+		true)
```
bwplotka, line 1692:
```suggestion
		true,
)
```

pkg/store/hintspb/custom.go
```diff
@@ -0,0 +1,10 @@
+// Copyright (c) The Thanos Authors.
+// Licensed under the Apache License 2.0.
+
+package hintspb
+
+func (m *SeriesResponseHints) AddQueriedBlock(id string) {
```
bwplotka, line 6: I would keep it typed and then decided how to format (.String()) below. It will make sure we will pass ULID nothing else (:

pkg/store/hintspb/hints.proto
```diff
@@ -0,0 +1,28 @@
...
+option (gogoproto.unmarshaler_all) = true;
+option (gogoproto.goproto_getters_all) = false;
+
+// Do not generate XXX fields to reduce memory.
+option (gogoproto.goproto_unkeyed_all) = false;
+option (gogoproto.goproto_unrecognized_all) = false;
+option (gogoproto.goproto_sizecache_all) = false;
+
+message SeriesResponseHints {
+    // queried_blocks is the list of blocks that have been queried.
```
bwplotka, line 22: gRPC comments for fields and messages should start with 3 dashes. I learned that at Improbable... but at this point I am not sure why though, would need to dive deeper ;p @mwitkow might know ^^

Let's change for consistency now?

Issue #2413, opened 2020-04-10
compact: downsample stopped making progress (compacted blocks migrated from prometheus too small)


belm0:
**Thanos, Prometheus and Golang version used**:  Thanos image:tag v0.11.0

**Object Storage Provider**:  GCS

**What happened**:  thanos-compact downsample stopped making progress (5m incomplete, no 1h).  The situation has continued for 10 hours.

bucket UI apppears as follows:
![Screen Shot 2020-04-11 at 8 39 09 AM](https://user-images.githubusercontent.com/1708631/79029815-f8743200-7bd0-11ea-9773-7a90ecfb6ad4.png)

**What you expected to happen**: compactor makes progress towards steady-state of downsampled data

**How to reproduce it (as minimally and precisely as possible)**:

newly deployed compactor job against 5 months of data

compact args:
```
         - "compact"
         - "--log.level=info"
         - "--log.format=logfmt"
         - "--http-address=0.0.0.0:10902"
         - "--objstore.config-file=/etc/config/object-store.yaml"
         - "--data-dir=/var/thanos/compact"
         - "--consistency-delay=30m"
         - "--retention.resolution-raw=0d"
         - "--retention.resolution-5m=0d"
         - "--retention.resolution-1h=0d"
         - "--block-sync-concurrency=20"
         - "--compact.concurrency=1"
         - "--wait"
```

**Full logs to relevant components**:

Compactor iterations have settled to doing no work.  There are no warnings or errors in the logs.  Compaction operations are still happening normally.

Log for latest iteration:
<details>Logs
<p>

```
E 2020-04-10T23:40:59.535995275Z level=info ts=2020-04-10T23:40:59.535727872Z caller=clean.go:50 msg="cleaning of aborted partial uploads done"
 
E 2020-04-10T23:40:59.535910379Z level=info ts=2020-04-10T23:40:59.535671469Z caller=fetcher.go:368 component=block.MetaFetcher msg="successfully fetched block metadata" duration=51.342219ms cached=52 returned=52 partial=0
 
E 2020-04-10T23:40:59.484518859Z level=info ts=2020-04-10T23:40:59.484316539Z caller=clean.go:25 msg="started cleaning of aborted partial uploads"
 
E 2020-04-10T23:40:59.484512229Z level=info ts=2020-04-10T23:40:59.484300199Z caller=retention.go:41 msg="optional retention apply done"
 
[...]

PR #2519, opened 2020-04-24
sidecar: document caveat of --shipper.upload_compacted

belm0:
Document a known issue of the experimental `--shipper.upload_compacted` flag, where the different compaction schedules of Prometheus and Thanos is not accounted for and may affect higher-order compaction and downsampling.

See #2413 

Signed-off-by: John Belmonte <john@neggie.net>

### DIFF
--- a/docs/components/sidecar.md
+++ b/docs/components/sidecar.md
@@ -79,6 +79,9 @@ To use this, the Prometheus compaction needs to be disabled. This can be done by
 - `--storage.tsdb.min-block-duration=2h`
 - `--storage.tsdb.max-block-duration=2h`
 
+A known issue is that Prometheus and Thanos use different compaction schedules, so Compact may
+get confused when trying to further compact or downsample compacted blocks from Prometheus.
+
 ## Flags
 
 [embedmd]:# (flags/sidecar.txt $)


docs/components/sidecar.md
```diff
@@ -79,6 +79,9 @@ To use this, the Prometheus compaction needs to be disabled. This can be done by
 - `--storage.tsdb.min-block-duration=2h`
 - `--storage.tsdb.max-block-duration=2h`
 
+A known issue is that Prometheus and Thanos use different compaction schedules, so Compact may
```
bwplotka, line 82: This is nice, can you link issue as well?

However, for documentation would be nice to be clear what we mean by "confused".. until we know roughly what can go wrong this piece of documentation will be only source of further questions.

belm0:
> Nice, some suggestion.
> 
> BTW I also don't think it should be confused (: It actually works as designed. The downsampling will kick at some point where blocks are packed into 2w no ?

As I've observed, the 162h blocks from prometheus do not get further compacted.

When we are mixing compacted blocks from different schedules, it means that blocks at the same compaction level have different sizes, and a block and level N may be bigger than a block with level N+1.  I don't know how the prometheus compaction library behaves in those circumstances, and whether the behavior is even defined.

docs/components/sidecar.md
belm0: A agree the language is vague, but it's because we actually don't know.  See my other comment about not knowing how the compaction library deals with blocks mixed together from different compaction schedules.

I guess that's why `upload_compacted` is considered experimental...

bwplotka: Well then we should either learn or not describe until we learn.

We have some idea though as you mentioned below

bwplotka:
So TSDB compaction should be smart enough to fill the gap for bigger level compaction if it does not overlap. So there might be a case IMO where.

lv4 = 2w
you have 3x 6d blocks. In this case none of them will be ever compacted.

But BTW. This is totally fine! (: We should not be worried much. Yes it will not downsample, but with you change we should be fine. 

We can add more tests here to know those edge cases: https://github.com/prometheus/prometheus/blob/5225f3ff5047e7c36a2beb1ae0376b788f07086b/tsdb/compact_test.go#L166 

((:

docs/components/sidecar.md
```diff
@@ -79,6 +79,9 @@ To use this, the Prometheus compaction needs to be disabled. This can be done by
...
+get confused when trying to further compact or downsample compacted blocks from Prometheus.
```
bwplotka, line 83:
```suggestion
get confused when trying to further compact or downsample compacted blocks from Prometheus.
This is similar to why enabling local compaction with upload is tricky.

This can be overcome by enabling experimental vertical compaction.
```

PR #2523, opened 2020-04-24
Added thanos tools serve command that allows to serve JSON files as StoreAPI.

bwplotka:
See: 

image

The plan is to add remote read support as well at some point with separate PR.

It would be awesome if before we will add remote read we can move forward Prometheus support of streamed remote read: https://github.com/prometheus/prometheus/issues/5926 
Motivated awesome student @Sudhar287 for this contribution :hugs: 


bwplotka:
I think this is actually quite useful (:

PR #2524, opened 2020-04-25
Apply webrouteprefix on debug endpoints

### LINKED ISSUE #2486: Web route prefix is not applied on debug endpoints.
In Prometheus, if the web external path is set, then all Prometheus routes will be updated. There is a good example here https://www.robustperception.io/using-external-urls-and-proxies-with-prometheus

```
You should be aware that with this external URL, the /prometheus/ path prefix will be required for all HTTP access to Prometheus. 

The /metrics will be on http://localhost:9090/prometheus/metrics for example.
```

However, in thanos, if web route prefix is set to `/thanos`, only `/api/v1` endpoints will be updated to `/thanos/api/v1`. But other endpoints like `/metrics`, still serve on `/metrics`, we should add web route prefix to these endpoints as well.

AC:
- Endpoints like `/metrics` or `/-/ready` can work with web route prefix just like Prometheus.

ranjithkumar007:
Fixes #2486 
Signed-off-by: ranjithkumar007 <ranjith.dakshana2015@gmail.com>

## Verification
Verified locally by checking the URLs with and without webrouteprefix.


ranjithkumar007:
Also, there seems some issue with js/CSS not being loaded properly when webrouteprefix is provided. Attached screenshots below

Without any webrouteprefix:
Screenshot from 2020-04-25 16-51-55

With webrouteprefix
Screenshot from 2020-04-25 16-51-52

yeya24:
The current solution for this is when web route prefix is provided, then web external prefix should also be provided, to make the static resources in the valid path.

cmd/thanos/bucket.go
```diff
@@ -335,7 +335,8 @@ func registerBucketWeb(m map[string]setupFunc, root *kingpin.CmdClause, name str
 			prober.NewInstrumentation(comp, logger, extprom.WrapRegistererWithPrefix("thanos_", reg)),
 		)
 
-		srv := httpserver.New(logger, reg, comp, httpProbe,
+		srv := httpserver.New(logger, reg, comp, "/",
```
yeya24, line 338: Do we want `webRoutePrefix` for all components, instead of just using `/` here? cc @GiedriusS @bwplotka

ranjithkumar007:
> The current solution for this is when web route prefix is provided, then web external prefix should also be provided, to make the static resources in the valid path.

Got it. So, I think the above screenshots shouldn't be an issue. 
We can log/inform to specify external prefix as well to view static resources when the user gives a web route prefix?

cmd/thanos/bucket.go
ranjithkumar007: I am not sure if it is required for all components. 
@bwplotka any inputs here?

cmd/thanos/compact.go
```diff
@@ -247,7 +247,8 @@ func runCompact(
 		prober.NewInstrumentation(component, logger, extprom.WrapRegistererWithPrefix("thanos_", reg)),
 	)
 
-	srv := httpserver.New(logger, reg, component, httpProbe,
+	srv := httpserver.New(logger, reg, component, "/",
```
bwplotka, line 250: Why not we route as well?

pkg/server/http/http.go
```diff
@@ -30,16 +30,20 @@ type Server struct {
...
 
 // New creates a new Server.
-func New(logger log.Logger, reg *prometheus.Registry, comp component.Component, prober *prober.HTTPProbe, opts ...Option) *Server {
+func New(logger log.Logger, reg *prometheus.Registry, comp component.Component, webRoutePrefix string, prober *prober.HTTPProbe, opts ...Option) *Server {
 	options := options{}
 	for _, o := range opts {
 		o.apply(&options)
 	}
 
+	if webRoutePrefix == "/" {
```
bwplotka, line 39: We can skip that if we would just use `path.Join` instead of `+` in `registerProfiler` (:

ranjithkumar007: Thanks! will update.

cmd/thanos/compact.go
ranjithkumar007: I checked existing compacter [flags](https://thanos.io/components/compact.md/#flags) in which there was `web.route-prefix`. I was not sure whether I should add `web.route-prefix` flag to all components.

ranjithkumar007: Should I add `web.route-prefix` to all components?

ranjithkumar007: ping

ranjithkumar007:
failing e2e unrelated to PR

squat:
Rerunning :)

yeya24:
@ranjithkumar007 E2E still failing.

Would you mind fixing this e2e test? 

```
##[error]    TestQueryRoutePrefix: query_test.go:184: query_test.go:184:
        
         unexpected error: the service querier-1 is not ready; err: got no expected status code: 404, expected: 200
      
```

I think the problem is here https://github.com/thanos-io/thanos/blob/master/test/e2e/e2ethanos/services.go#L170, you should add the web route prefix to readiness probe path as well.

ranjithkumar007:
> @ranjithkumar007 E2E still failing.
> 
> Would you mind fixing this e2e test?
> 
> ```
> ##[error]    TestQueryRoutePrefix: query_test.go:184: query_test.go:184:
>         
>          unexpected error: the service querier-1 is not ready; err: got no expected status code: 404, expected: 200
>       
> ```
> 
> I think the problem is here https://github.com/thanos-io/thanos/blob/master/test/e2e/e2ethanos/services.go#L170, you should add the web route prefix to readiness probe path as well.

Thanks @yeya24 .
Got it.. I will fix that.

cmd/thanos/compact.go
ranjithkumar007: May be we can do it in a separate PR.

prmsrswt: I think it would be nice to have `--web.route-prefix` in other components as well, but let's leave it for a separate PR.

ranjithkumar007:
Anything left here?

squat:
@ranjithkumar007 sorry we let this get so stale. Can you please rebase?

CC @prmsrswt for review considering web route prefix changes

prmsrswt:
There are some changes in route and external prefix handling in `master` branch which are not present in this PR so I'll test again after a rebase, but I think these changes are good to go :100:.

yeya24:
@ranjithkumar007 Hello. Sorry for the delay, but this pr is pretty close now. Can you please rebase master and fix the conflict? Thank you!

prmsrswt:
The new component Query Frontend also uses the `httpserver.New` function which was modified in this PR. @ranjithkumar007 can you update the code to use `web.route-prefix` there as well? Thanks.

yeya24:
@prmsrswt  Thanks for this!  It is fine for Query frontend now. I need to investigate whether route prefix breaks anything in Query frontend first. If it doesn't bring any risk, then I will add it in another pr,

prmsrswt:
The CI failures are because Query Frontend now has to use the modified HTTP server, which expects web route prefix as an argument. As @yeya24 said, we can pass an empty string for now in the case of Query Frontend and then investigate it further if it makes sense.

yeya24:
> The CI failures are because Query Frontend now has to use the modified HTTP server, which expects web route prefix as an argument. As @yeya24 said, we can pass an empty string for now in the case of Query Frontend and then investigate it further if it makes sense.

Oh yes, I missed that the function signature is changed :+1:

kakkoyun:
@ranjithkumar007 Friendly ping. Any update on this?

kakkoyun:
@ranjithkumar007 Any update? We can take this over if you're busy.

@yeya24 @prmsrswt What do you think? Shall we take this one over?

ranjithkumar007:
> @ranjithkumar007 Any update? We can take this over if you're busy.
 

Sorry, I wasn't getting time outside of work. Please feel free to take over.

Issue #2404, opened 2020-04-09
Query does not log undiscoverable store


belm0:
**Thanos, Prometheus and Golang version used**:  v0.11.0
**Object Storage Provider**:  GCS

**What happened**:
Specified Store and Sidecar storage endpoints on the command command line using DNS resolver.  The Sidecar endpoint was incorrect.  Query logged the addition of the Store endpoint, but nothing about the Sidecar endpoint.  The Query `/stores` page shows only the Store endpoint.
```
        --store.sd-dns-resolver=miekgdns
        --store=dnssrv+_grpc._tcp.thanos-store-grpc.default.svc.cluster.local
        --store=dnssrv+_grpc._tcp.thanos-sidecar-grpc.default.svc.cluster.local
```

**What you expected to happen**:
Query logs an error regarding DNS resolution of the bad endpoint.  The `/stores` page provides information about the unresolved endpoint.

bwplotka:
Agree, good point. :+1:  Marking as bug, help wanted to fix it :hugs:

yashrsharma44:
Shall I go about solving this issue?

bwplotka:
yes please!

yashrsharma44:
Great! I will hop on! :stuck_out_tongue:

yashrsharma44:
Hey @bwplotka! I was going through the issue, and after reproducing it, I observed the following - 
* The DNS resolution raises an error([here](https://github.com/thanos-io/thanos/blob/master/pkg/discovery/dns/provider.go#L123)) where it seems to log the error and move on with other dns resolution.
I was thinking of passing the error to this [function](https://github.com/thanos-io/thanos/blob/master/cmd/thanos/query.go#L239) where if it detects an error, it logs an error from the `Query` component.
What do you think about the approach?

bwplotka:
Nice, so if it already logs an error.. what is the problem we are trying to solve then? (:

yashrsharma44:
So the error raised was from the resolver.go file, where it does the resolution of dns, but somehow that error is not propagated to the query component. So I think we need to propagate the error, as I didn't see any errors raised in the logs of query component.

bwplotka:
What do you mean no propagate? There is literally `level.Error(p.logger).Log("msg", "dns resolution failed", "addr", addr, "err", err)`  log line :thinking:

yashrsharma44:
[...]

PR #2525, opened 2020-04-26
Added logging for dns resolution error in the Query component as well

yashrsharma44:
Signed-off-by: Yash <yashrsharma44@gmail.com>

## Changes

This PR addresses the issue  #2404 

## Verification
I had run the changes, and the Query component also logs the failure of the dns resolution. Earlier, only the `provider.go` logged the issue, now both are logging the error.

```
yash@kmaster  thanos-io/thanos   query-component-issue  ./thanos query --store.sd-dns-resolver=miekgdns --store=dnssrv+_grpc._tcp.thanos-store-grpc.default.svc.cluster.local
level=info ts=2020-04-26T08:10:36.149858286Z caller=main.go:152 msg="Tracing will be disabled"
level=info ts=2020-04-26T08:10:36.150778772Z caller=options.go:23 protocol=gRPC msg="disabled TLS, key and cert must be set to enable"
level=info ts=2020-04-26T08:10:36.151007023Z caller=query.go:408 msg="starting query node"
level=info ts=2020-04-26T08:10:36.151180057Z caller=intrumentation.go:60 msg="changing probe status" status=healthy
level=info ts=2020-04-26T08:10:36.151220838Z caller=http.go:56 service=http/server component=query msg="listening for requests and metrics" address=0.0.0.0:10902
level=info ts=2020-04-26T08:10:36.151303554Z caller=intrumentation.go:48 msg="changing probe status" status=ready
level=info ts=2020-04-26T08:10:36.151383662Z caller=grpc.go:106 service=gRPC/server component=query msg="listening for StoreAPI gRPC" address=0.0.0.0:10901
level=error ts=2020-04-26T08:11:38.153041054Z caller=provider.go:127 msg="dns resolution failed" addr=dnssrv+_grpc._tcp.thanos-store-grpc.default.svc.cluster.local err="lookup SRV records \"_grpc._tcp.thanos-store-grpc.default.svc.cluster.local\": could not resolve \"_grpc._tcp.thanos-store-grpc.default.svc.cluster.local\": all servers responded with errors to at least one search domain. Errs ;could not resolve _grpc._tcp.thanos-store-grpc.default.svc.cluster.local.: no servers returned a viable answer. Errs ;resolution against server 127.0.0.1 for _grpc._tcp.thanos-store-grpc.default.svc.cluster.local.: exchange: read udp 127.0.0.1:49740->127.0.0.1:53: i/o timeout"
level=error ts=2020-04-26T08:11:38.153395929Z caller=query.go:328 msg="failed to resolve addresses for storeAPIs" err="lookup SRV records \"_grpc._tcp.thanos-store-grpc.default.svc.cluster.local\": could not resolve \"_grpc._tcp.thanos-store-grpc.default.svc.cluster.local\": all servers responded with errors to at least one search domain. Errs ;could not resolve _grpc._tcp.thanos-store-grpc.default.svc.cluster.local.: no servers returned a viable answer. Errs ;resolution against server 127.0.0.1 for _grpc._tcp.thanos-store-grpc.default.svc.cluster.local.: exchange: read udp 127.0.0.1:49740->127.0.0.1:53: i/o timeout"

```

### Comments
I would like to discuss this approach, and see if this approach matches the requirements, as this is my first technical PR to Thanos! :nerd_face:

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -18,6 +18,8 @@ We use *breaking* word for marking changes that are not backward compatible (rel
 - [#2568](https://github.com/thanos-io/thanos/pull/2568) Query: Does not close the connection of strict, static nodes if establishing a connection had succeeded but Info() call failed
 - [#2615](https://github.com/thanos-io/thanos/pull/2615) Rule: Fix bugs where rules were out of sync.
 - [#2548](https://github.com/thanos-io/thanos/pull/2548) Query: Fixed rare cases of double counter reset accounting when querying `rate` with deduplication enabled.
+- [#2501](https://github.com/thanos-io/thanos/pull/2501) Query: gracefully handle additional fields in `SeriesResponse` protobuf message that may be added in the future.
+- [#2525](https://github.com/thanos-io/thanos/pull/2525) Query: Fixed logging for dns resolution error in the `Query` component.
 
 ### Added
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -18,7 +18,6 @@ We use *breaking* word for marking changes that are not backward compatible (rel
 - [#2568](https://github.com/thanos-io/thanos/pull/2568) Query: Does not close the connection of strict, static nodes if establishing a connection had succeeded but Info() call failed
 - [#2615](https://github.com/thanos-io/thanos/pull/2615) Rule: Fix bugs where rules were out of sync.
 - [#2548](https://github.com/thanos-io/thanos/pull/2548) Query: Fixed rare cases of double counter reset accounting when querying `rate` with deduplication enabled.
-- [#2501](https://github.com/thanos-io/thanos/pull/2501) Query: gracefully handle additional fields in `SeriesResponse` protobuf message that may be added in the future.
 - [#2525](https://github.com/thanos-io/thanos/pull/2525) Query: Fixed logging for dns resolution error in the `Query` component.
 
 ### Added
--- a/cmd/thanos/query.go
+++ b/cmd/thanos/query.go
@@ -304,7 +304,9 @@ func runQuery(
 					}
 					fileSDCache.Update(update)
 					stores.Update(ctxUpdate)
-					dnsProvider.Resolve(ctxUpdate, append(fileSDCache.Addresses(), storeAddrs...))
+					if err := dnsProvider.Resolve(ctxUpdate, append(fileSDCache.Addresses(), storeAddrs...)); err != nil {
+						level.Error(logger).Log("msg", "failed to resolve addresses for storeAPIs", "err", err)
+					}
 				case <-ctxUpdate.Done():
 					return nil
 				}
@@ -319,7 +321,9 @@ func runQuery(
 		ctx, cancel := context.WithCancel(context.Background())
 		g.Add(func() error {
 			return runutil.Repeat(dnsSDInterval, ctx.Done(), func() error {
-				dnsProvider.Resolve(ctx, append(fileSDCache.Addresses(), storeAddrs...))
+				if err := dnsProvider.Resolve(ctx, append(fileSDCache.Addresses(), storeAddrs...)); err != nil {
+					level.Error(logger).Log("msg", "failed to resolve addresses for storeAPIs", "err", err)
+				}
 				return nil
 			})
 		}, func(error) {
--- a/cmd/thanos/rule.go
+++ b/cmd/thanos/rule.go
@@ -748,8 +748,7 @@ func addDiscoveryGroups(g *run.Group, c *http_util.Client, interval time.Duratio
 
 	g.Add(func() error {
 		return runutil.Repeat(interval, ctx.Done(), func() error {
-			c.Resolve(ctx)
-			return nil
+			return c.Resolve(ctx)
 		})
 	}, func(error) {
 		cancel()
--- a/pkg/cacheutil/memcached_client.go
+++ b/pkg/cacheutil/memcached_client.go
@@ -464,7 +464,7 @@ func (c *memcachedClient) resolveAddrs() error {
 
 	// If some of the dns resolution fails, log the error.
 	if err := c.dnsProvider.Resolve(ctx, c.config.Addresses); err != nil {
-		level.Error(c.logger).Log("msg", "failed to resolve addresses for storeAPIs", "addresses", strings.Join(c.config.Addresses,","), "err", err)
+		level.Error(c.logger).Log("msg", "failed to resolve addresses for storeAPIs", "addresses", strings.Join(c.config.Addresses, ","), "err", err)
 	}
 	// Fail in case no server address is resolved.
 	servers := c.dnsProvider.Addresses()
--- a/pkg/cacheutil/memcached_client.go
+++ b/pkg/cacheutil/memcached_client.go
@@ -462,8 +462,10 @@ func (c *memcachedClient) resolveAddrs() error {
 	ctx, cancel := context.WithTimeout(context.Background(), 10*time.Second)
 	defer cancel()
 
-	c.dnsProvider.Resolve(ctx, c.config.Addresses)
-
+	// If some of the dns resolution fails, log the error.
+	if err := c.dnsProvider.Resolve(ctx, c.config.Addresses); err != nil {
+		level.Error(c.logger).Log("msg", "failed to resolve addresses for storeAPIs", "err", err)
+	}
 	// Fail in case no server address is resolved.
 	servers := c.dnsProvider.Addresses()
 	if len(servers) == 0 {
--- a/pkg/cacheutil/memcached_client.go
+++ b/pkg/cacheutil/memcached_client.go
@@ -5,6 +5,7 @@ package cacheutil
 
 import (
 	"context"
+	"strings"
 	"sync"
 	"time"
 
--- a/pkg/cacheutil/memcached_client.go
+++ b/pkg/cacheutil/memcached_client.go
@@ -464,7 +464,7 @@ func (c *memcachedClient) resolveAddrs() error {
 
 	// If some of the dns resolution fails, log the error.
 	if err := c.dnsProvider.Resolve(ctx, c.config.Addresses); err != nil {
-		level.Error(c.logger).Log("msg", "failed to resolve addresses for storeAPIs", "err", err)
+		level.Error(c.logger).Log("msg", "failed to resolve addresses for storeAPIs", "addresses", strings.Join(c.config.Addresses,","), "err", err)
 	}
 	// Fail in case no server address is resolved.
 	servers := c.dnsProvider.Addresses()
--- a/pkg/discovery/dns/provider.go
+++ b/pkg/discovery/dns/provider.go
@@ -13,6 +13,7 @@ import (
 	"github.com/go-kit/kit/log/level"
 	"github.com/prometheus/client_golang/prometheus"
 	"github.com/prometheus/client_golang/prometheus/promauto"
+	tsdberrors "github.com/prometheus/prometheus/tsdb/errors"
 	"github.com/thanos-io/thanos/pkg/discovery/dns/miekgdns"
 	"github.com/thanos-io/thanos/pkg/extprom"
 )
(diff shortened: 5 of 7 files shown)


pkg/discovery/dns/provider.go
```diff
@@ -125,6 +127,7 @@ func (p *Provider) Resolve(ctx context.Context, addrs []string) {
 			level.Error(p.logger).Log("msg", "dns resolution failed", "addr", addr, "err", err)
 			// Use cached values.
 			p.RLock()
+			errorList = append(errorList, err)
```
ranjithkumar007, line 130: I think this change can be moved outside of Rlock.

yashrsharma44: I was thinking of putting it outside the lock, the consistency of the error in the list does not matter. Thanks for the suggestion!

yashrsharma44:
> Could we please refactor this to use `errors.MultiError` instead of an array of `error`? AFAIK we generally try to avoid using `[]error` as you can see in the current code-base. That other type provides a nicer and unified format IMHO. The code would simpler as well since the for-loop would be contained inside of `MultiError.Error()` + we'd have one, simple message about all failures instead of spam :P

Thanks for the review! I guess that sounds elegant, as the error log certainly seems a bit clogged. Making the changes right away! :smile:

CHANGELOG.md
```diff
@@ -14,6 +14,7 @@ We use *breaking* word for marking changes that are not backward compatible (rel
 ### Fixed
 
 - [#2501](https://github.com/thanos-io/thanos/pull/2501) Query: gracefully handle additional fields in `SeriesResponse` protobuf message that may be added in the future.
+- [#2525](https://github.com/thanos-io/thanos/pull/2525) Query: Added logging for dns resolution error in the `Query` component.
```
bwplotka, line 17: So fixed or added? (; You put this in `Fixed` section , but we said "Added" here... ^^

I think it's fixed as we fix issue

cmd/thanos/query.go
```diff
@@ -304,7 +304,9 @@ func runQuery(
 					}
 					fileSDCache.Update(update)
 					stores.Update(ctxUpdate)
-					dnsProvider.Resolve(ctxUpdate, append(fileSDCache.Addresses(), storeAddrs...))
+					if err := dnsProvider.Resolve(ctxUpdate, append(fileSDCache.Addresses(), storeAddrs...)); err.Err() != nil {
+						level.Error(logger).Log("msg", "failed to resolve addresses for storeAPIs", "err", err.Error())
```
bwplotka, line 308:
```suggestion
						level.Error(logger).Log("msg", "failed to resolve addresses for storeAPIs", "err", err)
```

bwplotka: Logger will know what to do

cmd/thanos/query.go
```diff
@@ -304,7 +304,9 @@ func runQuery(
 					}
 					fileSDCache.Update(update)
 					stores.Update(ctxUpdate)
-					dnsProvider.Resolve(ctxUpdate, append(fileSDCache.Addresses(), storeAddrs...))
+					if err := dnsProvider.Resolve(ctxUpdate, append(fileSDCache.Addresses(), storeAddrs...)); err.Err() != nil {
```
bwplotka, line 307: There is quite big bug here, but CI should help you to find what. 

I will not say exactly what's wrong for educational purpose, please try to get from CI what's wrong you should be able to quickly spot this! :hugs: Let me know if you are blocked.

cmd/thanos/rule.go
```diff
@@ -748,8 +748,8 @@ func addDiscoveryGroups(g *run.Group, c *http_util.Client, interval time.Duratio
 
 	g.Add(func() error {
 		return runutil.Repeat(interval, ctx.Done(), func() error {
-			c.Resolve(ctx)
-			return nil
+			err := c.Resolve(ctx)
```
bwplotka, line 751: Nice! No need for separate variable though.

Please see https://thanos.io/contributing/coding-style-guide.md/#avoid-defining-variables-used-only-once

pkg/cacheutil/memcached_client.go
```diff
@@ -462,11 +462,10 @@ func (c *memcachedClient) resolveAddrs() error {
 	ctx, cancel := context.WithTimeout(context.Background(), 10*time.Second)
 	defer cancel()
 
-	c.dnsProvider.Resolve(ctx, c.config.Addresses)
-
+	err := c.dnsProvider.Resolve(ctx, c.config.Addresses)
 	// Fail in case no server address is resolved.
 	servers := c.dnsProvider.Addresses()
-	if len(servers) == 0 {
+	if len(servers) == 0 || err != nil {
```
bwplotka, line 468: I think we want to always check error, no?

pkg/discovery/dns/provider.go
```diff
@@ -107,8 +108,10 @@ func GetQTypeName(addr string) (qtype string, name string) {
 // Resolve stores a list of provided addresses or their DNS records if requested.
 // Addresses prefixed with `dns+` or `dnssrv+` will be resolved through respective DNS lookup (A/AAAA or SRV).
 // defaultPort is used for non-SRV records when a port is not supplied.
-func (p *Provider) Resolve(ctx context.Context, addrs []string) {
+func (p *Provider) Resolve(ctx context.Context, addrs []string) tsdberrors.MultiError {
 	resolvedAddrs := map[string][]string{}
+	errorList := tsdberrors.MultiError{}
```
bwplotka, line 113: We use to use just `errs` name for those

pkg/discovery/dns/provider.go
```diff
@@ -120,6 +123,7 @@ func (p *Provider) Resolve(ctx context.Context, addrs []string) {
 		resolved, err := p.resolver.Resolve(ctx, name, QType(qtype))
 		p.resolverLookupsCount.Inc()
 		if err != nil {
+			errorList.Add(err)
 			// The DNS resolution failed. Continue without modifying the old records.
 			p.resolverFailuresCount.Inc()
 			level.Error(p.logger).Log("msg", "dns resolution failed", "addr", addr, "err", err)
```
bwplotka, line 129: Ok this wrong. 

The problem is that we handle errors TWICE. One time here, then we pass `errs` somewhere. This will mean that we probably will log errors couple of times and just add confusion... Same with metric which is even more risky. Let's handle error just once. Either handle it all (just log), or pass further... not both! (: 

What do you think?

pkg/discovery/dns/provider_test.go
```diff
@@ -32,63 +32,69 @@ func TestProvider(t *testing.T) {
...
 
-	prv.Resolve(ctx, []string{"any+a", "any+b", "any+c"})
-	result = prv.Addresses()
-	sort.Strings(result)
-	testutil.Equals(t, ips, result)
-	testutil.Equals(t, 3, promtestutil.CollectAndCount(prv.resolverAddrs))
-	testutil.Equals(t, float64(2), promtestutil.ToFloat64(prv.resolverAddrs.WithLabelValues("any+a")))
-	testutil.Equals(t, float64(2), promtestutil.ToFloat64(prv.resolverAddrs.WithLabelValues("any+b")))
-	testutil.Equals(t, float64(1), promtestutil.ToFloat64(prv.resolverAddrs.WithLabelValues("any+c")))
+	if err := prv.Resolve(ctx, []string{"any+x"}); err != nil {
```
bwplotka, line 62: Why this pattern? 

With this if there is error on ALL resolves tests will be still green.
 Why not using just `testutil.Ok(t, prv.Resolve...)`?

pkg/http/http.go
```diff
@@ -163,7 +164,7 @@ func (c FileSDConfig) convert() (file.SDConfig, error) {
 }
 
 type AddressProvider interface {
-	Resolve(context.Context, []string)
+	Resolve(context.Context, []string) tsdberrors.MultiError
```
bwplotka, line 167: You should never return concrete type of error in Go.

```suggestion
	Resolve(context.Context, []string) errors
```

pkg/http/http.go
```diff
@@ -259,6 +260,8 @@ func (c *Client) Discover(ctx context.Context) {
 }
 
 // Resolve refreshes and resolves the list of targets.
-func (c *Client) Resolve(ctx context.Context) {
-	c.provider.Resolve(ctx, append(c.fileSDCache.Addresses(), c.staticAddresses...))
+func (c *Client) Resolve(ctx context.Context) tsdberrors.MultiError {
+	// Skip the error as it would be logged in the corresponding component.
```
bwplotka, line 264: So why we return error then?

bwplotka: Also what do you mean skip? We are not skipping , we are passing it further so handling properly

CHANGELOG.md
yashrsharma44: Sure, some sloppy grammar from my side :P

cmd/thanos/query.go
yashrsharma44: Challenge accepted!

yashrsharma44:
Thanks for the suggestion, I will make the changes :smile:

pkg/discovery/dns/provider.go
yashrsharma44: I think that we should not log error in the `dns/provider.go`, rather the error should be logged by the function requesting it, so that we can clearly log where the error is coming from.

bwplotka: If we log, and we do that only for this error, do we really need to return this error? Does the caller ("requesting function") needs to act in specific way for those errors?

yashrsharma44: Okay, so I forgot to mention that, I am not logging the error _now_ in `provider.go`, rather I am returning it(in my current change that I will push), and the receiver would handle that error(by logging it).

yashrsharma44: > Does the caller ("requesting function") needs to act in specific way for those errors?

If we can log in the scope of the requesting function, that would be helpful in understanding which function request resulted in the error. Other than that, there is no specific behaviour that I am looking forward to.

bwplotka: I understand this, but still what would be the solid answer to the question: 

> If we log, and we do that only for this error, do we really need to return this error? Does the caller ("requesting function") needs to act in specific way for those errors?

(: ?

pkg/discovery/dns/provider_test.go
yashrsharma44: Great! I didn't had a look at the `testutil` package, the construct is quite elegant.

pkg/http/http.go
yashrsharma44: Sure, reverted it.

pkg/http/http.go
yashrsharma44: I have changed it.

pkg/discovery/dns/provider.go
yashrsharma44: Sure.

pkg/cacheutil/memcached_client.go
yashrsharma44: Sure, made a note.

cmd/thanos/rule.go
yashrsharma44: Thanks!

cmd/thanos/query.go
yashrsharma44: Sure

pkg/discovery/dns/provider.go
yashrsharma44: > If we log, and we do that only for this error, do we really need to return this error? Does the caller ("requesting function") needs to act in specific way for those errors?

As I have understood this question, I think we should return the error and the requesting function should log the error. If you have found any function which is doing both, then please highlight that, I might have missed that :sweat_smile:

pkg/cacheutil/memcached_client.go
```diff
@@ -462,8 +462,10 @@ func (c *memcachedClient) resolveAddrs() error {
 	ctx, cancel := context.WithTimeout(context.Background(), 10*time.Second)
 	defer cancel()
 
-	c.dnsProvider.Resolve(ctx, c.config.Addresses)
-
+	// If some of the dns resolution fails, log the error.
+	if err := c.dnsProvider.Resolve(ctx, c.config.Addresses); err != nil {
+		level.Error(c.logger).Log("msg", "failed to resolve addresses for storeAPIs", "err", err)
```
bwplotka, line 467: Maybe worth mentioning addresses?
```suggestion
		level.Error(c.logger).Log("msg", "failed to resolve addresses for storeAPIs", "addresses", strings.Join(c.config.Addresses,","), "err", err)
```

pkg/discovery/dns/provider.go
```diff
@@ -143,6 +147,11 @@ func (p *Provider) Resolve(ctx context.Context, addrs []string) {
 	p.resolverAddrs.Submit()
 
 	p.resolved = resolvedAddrs
+	if len(errs) > 0 {
```
bwplotka, line 150: Power tip: we can do just returns. `errs.Err()` (:  Without check! (check will be done on err)

pkg/cacheutil/memcached_client.go
yashrsharma44: Sure, that looks neat!

pkg/discovery/dns/provider.go
yashrsharma44: Ohh great, let me add it up!

bwplotka:
Thanks!

PR #2526, opened 2020-04-26
compact: add label when no labels left after deduplication

sepich:

## Changes
Fixes https://github.com/thanos-io/thanos/issues/2550
Per https://github.com/thanos-io/thanos/issues/1014#issuecomment-605873793 i'm adding new flag `--deduplication.label-if-empty=label=value` to handle situation, when there are no labels left after deduplication.

## Verification

Basic tests added.
Tested manually on our data, collected by 3 thanos-receive with only `--label=replica="$(NAME)"`


sepich:
Sure, will rewrite then to use first label name from list set in `--deduplication.replica-label`, and then automatically add `first_label_name=deduped` in this case.
I think it is better than `constant_replica_label`, because user probably do not have this constant in `--query.replica-label` on querier. Which will lead to metrics have additional label at some point in time (corresponding to compaction start)

bwplotka:
Good idea @kakkoyun !

sepich:
I can add `level.Warn` log, but for us it is usual case - so this would just spam the logs.
Could you please clarify how it "affect people", so we consider adding additional label to current `replica`.

bwplotka:
> I can add level.Warn log, but for us it is usual case - so this would just spam the logs.

Really? It's every compaction so it's fine. And also you should have unique label per each "stream" as it will easier to scale to more streams some day. That's why log .

bwplotka:
Thanks!

Issue #2401, opened 2020-04-08
querier: Rate over deduplicated counter from many replicas can lead to double reset account.


bwplotka:
Found by GitLab, we were investigating offline with @SuperQ 

Their issue: https://gitlab.com/gitlab-com/gl-infra/infrastructure/-/issues/9293

This can be only reproducible with large rates `[30m+]` which means it has to do with chunks ordering or overlaps.

bwplotka:
Some potential issue is that overlaps are not handled really well: https://github.com/thanos-io/thanos/pull/2400/files

bwplotka:
It can be something related to fact that GitLab is using Store GW in HA without loadbalancer (querying both in same time), so the data is duplicated and unsorted (chunks) for sure.

SuperQ:
@bwplotka We initially reproduced the issue with direct to Prometheus sidecars, not Store GW. These are in HA pairs as well.

bwplotka:
Got some test data from @SuperQ so hopefully will be able to repro locally :hugs: 

Fingers crossed :crossed_fingers:

bwplotka:
Looks like it is dup of https://github.com/thanos-io/thanos/issues/1326 - let's continue discussion there.

bwplotka:
Actually let's not be so  sure, this might be different (here deduplication does not cause it)

bwplotka:
@SuperQ this repro is so amazing can explore all details. Definitely we change overlapping and unsorted chunks. We should be able to find a problem in our algorithm soon, thanks!

BTW... I kind of overengineered (as you can imagine) and wrote `thanos tools storeapi serve --json=<file x>` which can serve JSON (protobuf based) and as Store API :tada: 

So I can get your file (actually anything generated by `grpcurl` and put into `thanos tools storeapi serve --json` , run querier and connect `storeapi serve` as --store, and see your results: 

![image](https://user-images.githubusercontent.com/6950331/80226032-c5578700-8643-11ea-923c-c921dd20f0d1.png)

bwplotka:
Tooling looks like it works, but I think we don't have enough chunks to repro it :thinking: 

Tried all sorts of time ranges, steps and rate ranges.. no luck:

![image](https://user-images.githubusercontent.com/6950331/80238264-08225a80-8656-11ea-9ffe-3000e5296eb8.png)

All good everywhere...

cc @SuperQ

SuperQ:
[...]

PR #2528, opened 2020-04-27
  querier: Added regressions tests for counter missed reset bug.

bwplotka:
PR with just tests, not fix yet.

Reproduces: https://github.com/thanos-io/thanos/issues/2401

    * Added regressions tests for CounterSeriesIterator; Simplified aggregators.
    * Fixes edge dedup cases for Next and added tests for deduplication.
    * Refactored downsampling tests, added more realistic cases.
    * Added check for duplicated chunks during downsampling.
    * Removed duplicates for efficiency on promSeriesSet.

Will add quick fix in separate PR for clarity.


bwplotka:
Findings so far:

After Querier Select (Counter + dedup) our data looks like have malformed interval.
```
all: avg = 209.830508 timeavg =  1791445.377119
[1,2]/3: avg = 217.369128 timeavg =  1125001.791946
3/3: avg = 196.139535 timeavg =  690765.604651
2/2: avg = 204.257353 timeavg =  1051698.242647
```

This is causing unexpected output like this:

```
	t.Run("Rate=30mStep=500s", func(t *testing.T) {
			q, err := e.NewRangeQuery(&mockedQueryable{q}, `rate(gitlab_transaction_cache_read_hit_count_total[30m])`, timestamp.Time(realSeriesWithStaleMarkerMint).Add(30*time.Minute), timestamp.Time(realSeriesWithStaleMarkerMaxt), 500*time.Second)
			testutil.Ok(t, err)
			defer q.Close()

			r := q.Exec(context.Background())
			testutil.Ok(t, r.Err)
			testutil.Assert(t, len(r.Warnings) == 0)

			vec, err := r.Matrix()
			testutil.Ok(t, err)
			testutil.Equals(t, promql.Matrix{
				{Metric: expectedLset, Points: []promql.Point{
					{T: 1587691800000, V: 14.464425770308123},
					{T: 1587692300000, V: 14.763025210084033},
					{T: 1587692800000, V: 291.105652426067}, // ??
					{T: 1587693300000, V: 290.89236811640285},
				}},
			}, vec)
		})
```

bwplotka:
I am keeping adding findings and removing because they were wrong :facepalm:

bwplotka:
Got it!

Delta of timestamps in the broken dataset:
```
15000
15000
61358 // ??
14990
15000
```

And values!
```
269
213
207
114  
49
50
70
36
51
39
53
-7  // Wut? Counter decrease?
96
211
196
217
```

bwplotka:
Minimal repro case is as follows:

```
{ 	                // Regression test against https://github.com/thanos-io/thanos/issues/2401.
			// Two counter series, when one (initially chosen) series is having hiccup (few dropped samples), while second is live.,
			input: []series{
				{
					lset: labels.Labels{{Name: "replica", Value: "01"}},
					samples: []sample{
						{10000, 8.0}, // Smaller timestamp, this will be chosen. CurrValue = 8.0
						{20000, 9.0}, // Same. CurrValue = 9.0
						// {Gap} app reset. No sample, because stale marker but removed by downsample.CounterIterator.
						{50001, 9 + 1.0}}, // Next after 20000+1 has a bit higher than timestamp then in second series. Penalty 5000 will be added.
				}, {
					lset: labels.Labels{{Name: "replica", Value: "02"}},
					samples: []sample{
						{10001, 8.0}, // Penalty 5000 will be added.
						// 20001 was app reset. No sample, because stale marker but removed by downsample.CounterIterator. Penalty 2 * (20000 - 10000) will be added.
						// 30001 no sample. Within penalty, ignored.
						{45001, 8.0 + 0.5}, // Smaller timestamp, this will be chosen. CurrValue = 8.5 which is smaller THAN last chosen value.
					},
				},
			},
			exp: []series{
				{
					lset:    labels.Labels{},
					// Outcome for rate: Double counter reset accounted.
					samples: []sample{{10000, 8}, {20000, 9}, {45001, 8.5}},
				},
			},
			dedupLabels: map[string]struct{}{
				"replica": {},
			},
		},
```

[`downsample.CounterSeriesIterator`](https://github.com/thanos-io/thanos/blob/f8bbcbbe4bfc144dec81096b95cd63af9ef2d584/pkg/compact/downsample/downsample.go#L564) is something that is only triggered for RAW or Aggregated COUNTER chunks and ONLY when `rate` was used. This is to ensure we properly detect resets in aggregated data. This is mainly for specific edge samples in aggregated data. Eventually reset is accounted.

The problem is that in certain replica cases when reset is accounted... We can have a decremented counter, which means fake counter reset.

cc @brancz @squat @SuperQ 

Thinking about how to fix.. we have few options:

1. Use  `downsample.CounterSeriesIterator` only on downsampled data. I think this is not the best because this iterator was actually designed for both. Mainly I am rather convinced that we can have the same case on downsampled, replicated data.
2. Pass the options `isCounter` to dedup. We know this when `rate` is used. (Well we know that it should be counter). Thanks of that we can try to make sure NO value decrease should ever be spotted.
3. We can change CounterSeriesIterator to add "fake resets" so we can be sure that deduplication iterator is always working on full overview. If the replicated application scrape was removed, dedup will see stale marker etc, so everything Prometheus do.

I think the 3 is the cleanest, but I need to yet dive more on how and IF we can fake stale marker, counter reset. (:

bwplotka:
I pushed my tests reproducing all, but it definitely needs lots of cleanup before merge (:

brancz:
4. separate iterator for raw and downsampled

Torn between 3 & 4. Happy with either I think.

bwplotka:
Nice @brancz, let's discuss on issue: https://github.com/thanos-io/thanos/issues/2401

I think 3 is a quick win here.

bwplotka:
Should be ready for review. Will add a fix in next chained PR to not obfuscate it with a large number of tests (we needed them so much I think!)

So this PR does not change much, it's low risk - mostly tests.

cc @squat @kakkoyun @brancz @povilasv @GiedriusS

bwplotka:
Long term fix: https://github.com/thanos-io/thanos/issues/2547

pkg/query/iter.go
```diff
@@ -153,33 +184,47 @@ func (s *chunkSeries) Iterator() storage.SeriesIterator {
...
-	case resAggrAvg:
+		return newBoundedSeriesIterator(sit, s.mint, s.maxt)
+	}
+
+	if len(s.aggrs) != 2 {
+		return errSeriesIterator{err: errors.Errorf("unexpected result aggregate type %v", s.aggrs)}
+	}
+
+	switch {
+	case s.aggrs[0] == storepb.Aggr_SUM && s.aggrs[1] == storepb.Aggr_COUNT,
```
GiedriusS, line 225: When is this case possible? `aggrsFromFunc` only returns the 2nd case.

pkg/compact/downsample/downsample.go
```diff
@@ -551,16 +565,21 @@ type sample struct {
...
 // Counter aggregation chunks must have the first and last values from their
 // original raw series: the first raw value should be the first value encoded
 // in the chunk, and the last raw value is encoded by the duplication of the
-// previous sample's timestamp.  As iteration occurs between chunks, the
+// previous sample's timestamp. As iteration occurs between chunks, the
 // comparison between the last raw value of the earlier chunk and the first raw
 // value of the later chunk ensures that counter resets between chunks are
 // recognized and that the correct value delta is calculated.
+//
+// It handles overlapped chunks (removes overlaps.
```
GiedriusS, line 580:
```suggestion
// It handles overlapped chunks (removes overlaps).
```

pkg/query/iter.go
bwplotka: This is average

pkg/compact/downsample/downsample_test.go
```diff
@@ -162,185 +162,389 @@ func TestExpandChunkIterator(t *testing.T) {
...
+					AggrMax:     {{t: 1587690299999, v: 465870}, {t: 1587690599999, v: 469951}, {t: 1587690899999, v: 474726}, {t: 1587691199999, v: 479368}, {t: 1587691499999, v: 483566}, {t: 1587691799999, v: 487787}, {t: 1587692099999, v: 492065}, {t: 1587692399999, v: 496245}, {t: 1587692699999, v: 496544}, {t: 1587692999999, v: 6010}, {t: 1587693299999, v: 10242}, {t: 1587693590791, v: 14956}},
+					AggrCounter: {{t: 1587690005791, v: 461968}, {t: 1587690299999, v: 465870}, {t: 1587690599999, v: 469951}, {t: 1587690899999, v: 474726}, {t: 1587691199999, v: 479368}, {t: 1587691499999, v: 483566}, {t: 1587691799999, v: 487787}, {t: 1587692099999, v: 492065}, {t: 1587692399999, v: 496245}, {t: 1587692699999, v: 498647}, {t: 1587692999999, v: 502554}, {t: 1587693299999, v: 506786}, {t: 1587693590791, v: 511500}, {t: 1587693590791, v: 14956}},
+				},
+			},
+			resolution: ResLevel2,
+
+			expected: []map[AggrType][]sample{
+				{
+					0x0: {{t: 1587693590791, v: 236}},
```
GiedriusS, line 397: Could we use named constants here? :smile:

pkg/compact/downsample/downsample_test.go
```diff
@@ -162,185 +162,389 @@ func TestExpandChunkIterator(t *testing.T) {
...
+			}
 
-// testDownsample inserts the input into a block and invokes the downsampler with the given resolution.
-// The chunk ranges within the input block are aligned at 500 time units.
-func testDownsample(t *testing.T, data []*downsampleTestSet, meta *metadata.Meta, resolution int64) {
-	t.Helper()
+			for i := int64(0); i < 120; i++ {
+				d.inAggr[0][AggrCounter] = append(d.inAggr[0][AggrCounter], sample{t: i, v: float64(i)})
+				d.inAggr[1][AggrCounter] = append(d.inAggr[1][AggrCounter], sample{t: 120 + i, v: float64(120 + i)})
+				d.inAggr[2][AggrCounter] = append(d.inAggr[2][AggrCounter], sample{t: 240 + i, v: float64(240 + i)})
```
GiedriusS, line 417: Not a biggie but I think here you forgot to fill the 4th element in the `d.inAggr` array with samples.

pkg/compact/downsample/downsample_test.go
```diff
@@ -162,185 +162,389 @@ func TestExpandChunkIterator(t *testing.T) {
...
+				series = append(series, pall.At())
+			}
+			testutil.Ok(t, pall.Err())
+			testutil.Equals(t, 1, len(series))
 
-	for pall.Next() {
-		id := pall.At()
+			var lset labels.Labels
+			var chks []chunks.Meta
+			testutil.Ok(t, indexr.Series(series[0], &lset, &chks))
```
GiedriusS, line 478: Very minor suggestion: we could be a bit paranoic here and check if `lset` has the right labels.

pkg/compact/downsample/downsample_test.go
bwplotka: Yes :hugs:

pkg/compact/downsample/downsample_test.go
bwplotka: :+1:  Good point!

pkg/query/querier_test.go
```diff
@@ -482,9 +931,15 @@ func TestSortReplicaLabel(t *testing.T) {
 	}
 }
 
+const hackyStaleMarker = float64(-99999999)
```
GiedriusS, line 934: Could we please add a comment on _why_ this is needed? It's clear that NaN != NaN but it would be nice to know in the future quickly why this is needed.

bwplotka: Just for testutil.Equals to work, but good point (:

pkg/query/iter.go
bwplotka: ah I know what you mean. You are right but it's easy to swap and assume it will handled well, so I would vote to be extra safe here.

bwplotka:
Thanks for quick and detailed review @GiedriusS  on this big PR. Should be all addressed :+1:

bwplotka:
Thanks for the effort of reviewing this so quickly @GiedriusS (huge PR)! :heart:

Issue #415, opened 2018-07-05
sidecar: Handle or do not allow `delete_series` for Thanos with object store backup


bwplotka:
Basically https://github.com/prometheus/prometheus/blob/master/docs/querying/api.md#delete-series is unsafe to run with Thanos sidecar that uploads blocks to bucket because:
- Prometheus rewrites *ALL* blocks because they are immutable. This causes Thanos sidecar to upload new ones (without removal of old ones) which causes compactor to stop because of overlap.
- Race conditions (rewrite during Thanos shipper upload)

It is worth to note that it does not make sense much to do `delete_series` on local Prometheus with Thanos backup, because most likely all blocks are already uploaded at the moment when Prometheus replaces these with rewritten blocks. So deleted series will be seen only locally. Plus every older block will have still that `series`

Solution:
- We don't want to replace the blocks in sidecar. We don't want keep whole process coordination free, so only compactor is allowed to remove blocks on regular basis (compaction process)
- We could instead of specifying which ULIDs where deployed in Thanos meta, specify also time ranges, and ignore or error out the rewritten blocks with new ULID and same time range.

Thanks `[slack]@jean-louis` for spotting this!

dupondje:
Great we found the root cause :)

Even better would be (in a dream scenario), that you can do delete_series in Thanos. And that Thanos deletes series in the bucket also.

But it might be a good start to let the compactor just delete the new/adjusted blocks on each run.

Bplotka:
> But it might be a good start to let the compactor just delete the new/adjusted blocks on each run.

What do you mean? I think it is way more clear to block this behaviour and be explicit about it. (not sure yet how). Basically if we do this ("let the compactor just delete the new/adjusted blocks on each run.") we have few problems:
- Maybe these new blocks are not because of `delete_series` but some other configuration errors? (e.g same external labels)
- User is confident that actually she/he deleted these series everywhere, but actually, fresh series are only deleted.
- What if user deleted series before Thanos being able to upload some of the blocks? Really big place of inconsistencies.

All of this makes me think to do either:
a. Block or explicitly do not allow this logic
b. Block or explicitly do not allow this logic, but implement some batch job code that will delete series from the whole history (from bucket) - actually user can implement that code on its own even now.
c. Integrate `delete series` with the whole stack (sidecar + compactor) - but I think that is way to complex for the actual value.

**Can we first enumerate what are the use cases for `delete series`?** 
[...]

PR #2530, opened 2020-04-27
Add deleteSeries skeleton to return bad request

darshanime:

Thanos currently does not support `delete_series` api, see https://github.com/thanos-io/thanos/issues/415. This PR adds a skeleton endpoint explicitly mentioning the api is not supported yet as [suggested earlier](https://github.com/thanos-io/thanos/issues/415#issuecomment-403399284).
 
We will eventually implement the API by adding in functionality in the handler. 

### Changes
- return `400 Bad Request` on `delete_series` API
### Verification
```
curl -X POST "http://thanos_querier:10902/api/v1/delete_series?match[]=scrape_duration_seconds"  -v
*   Trying 172.24.0.5...
* TCP_NODELAY set
* Connected to thanos_querier (172.24.0.5) port 10902 (#0)
> POST /api/v1/delete_series?match[]=scrape_duration_seconds HTTP/1.1
> Host: thanos_querier:10902
> User-Agent: curl/7.60.0
> Accept: */*
>
< HTTP/1.1 400 Bad Request
< Access-Control-Allow-Headers: Accept, Accept-Encoding, Authorization, Content-Type, Origin
< Access-Control-Allow-Methods: GET, OPTIONS
< Access-Control-Allow-Origin: *
< Access-Control-Expose-Headers: Date
< Cache-Control: no-store
< Content-Type: application/json
< Vary: Accept-Encoding
< Date: Mon, 27 Apr 2020 17:06:40 GMT
< Content-Length: 89
<
{"status":"error","errorType":"bad_data","error":"delete_series is not implemented yet"}
* Connection #0 to host thanos_querier left intact
```

### DIFF
--- a/pkg/query/api/v1.go
+++ b/pkg/query/api/v1.go
@@ -166,6 +166,8 @@ func (api *API) Register(r *route.Router, tracer opentracing.Tracer, logger log.
 	r.Get("/series", instr("series", api.series))
 	r.Post("/series", instr("series", api.series))
 
+	r.Post("/delete_series", instr("delete_series", api.deleteSeries))
+
 	r.Get("/labels", instr("label_names", api.labelNames))
 	r.Post("/labels", instr("label_names", api.labelNames))
 }
@@ -534,6 +536,10 @@ func (api *API) series(r *http.Request) (interface{}, []error, *ApiError) {
 		return nil, nil, &ApiError{errorExec, set.Err()}
 	}
 	return metrics, warnings, nil
+}
+
+func (api *API) deleteSeries(r *http.Request) (interface{}, []error, *ApiError) {
+	return nil, nil, &ApiError{errorBadData, errors.New("delete_series is not implemented yet")}
 }
 
 func Respond(w http.ResponseWriter, data interface{}, warnings []error) {
--- a/pkg/query/api/v1_test.go
+++ b/pkg/query/api/v1_test.go
@@ -786,6 +786,13 @@ func TestEndpoints(t *testing.T) {
 			errType: errorBadData,
 			method:  http.MethodPost,
 		},
+		// deleteSeries should return bad request.
+		{
+			endpoint: api.deleteSeries,
+			query:    url.Values{},
+			method:   http.MethodPost,
+			errType:  errorBadData,
+		},
 	}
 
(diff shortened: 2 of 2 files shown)


darshanime:
@bwplotka I have added a high level approach on the analysis doc - https://docs.google.com/document/d/1L8PJu0ScDTrWp0mfPKYu2GcJU19nkQKLe1joD4vNncg/edit

What do you think about this first step?

yeya24:
Thanks for starting this and the design doc!
However, I think we should put this API to compact, instead of thanos query. In the previous design, compact does the job to delete series so it is more appropriate to put the API there

darshanime:
>In the previous design, compact does the job to delete series  

Oh, I was not aware of any previous designs on this, where can I read more about it?

I thought the querier might be a good place beacuse:
- It can be called from clients directly, if for eg Grafana was to add the feature to be able to delete series from dashboards directly.
- It can orchestrate the execution of `delete_series` in sidecars and compactor.  

What do you think?

bwplotka:
As first step this is good, but let's talk about design on the issue itself, WDYT?

bwplotka:
Actually I reverted this, sorry I merged prematurely: https://github.com/thanos-io/thanos/pull/2551

We don't know where to put it yet etc. We need design first. I think compactor might be better place.

yeya24:
> > In the previous design, compact does the job to delete series
> 
> Oh, I was not aware of any previous designs on this, where can I read more about it?

There were some discussions in the issue. 

> I thought the querier might be a good place beacuse:
> 
> * It can be called from clients directly, if for eg Grafana was to add the feature to be able to delete series from dashboards directly.
> * It can orchestrate the execution of `delete_series` in sidecars and compactor.
> 
> What do you think?

IMO `delete_series` won't be supported by Grafana, since this API is more related to `admin`, not just query, so it is dangerous to put it in Grafana.

In the context of `delete_series`, I think we only talk about deleting series in obj store? For the series in Prometheus, we can already use Prometheus' API.

But that's a good point, there is no way to delete data in Thanos Rule and Receive, it is good to support `delete_series` API like Prometheus on the two as well

bwplotka:
:+1: 

Again, a design in our proposal docs would be nice.

PR #2532, opened 2020-04-28
Chunks caching at bucket level

pstibrany:
This PR implements chunks caching as described in ["Caching in Thanos"](https://docs.google.com/document/d/1knNXQfDJlQDA9Dr9R9OVFvdWS83sXd0kLEFdv21IxCM/edit#) document.

Chunks caching is done at Bucket level. Since bucket doesn't really know about chunks, and Thanos simply uses GetRange call on Bucket to fetch chunks, this implementation splits such requests into smaller "blocks" (with size of 16000 bytes), and stores each block into cache individually.

To enable chunks caching, one needs to configure `store` to use caching bucket using `--experimental.caching-bucket.config=store_cache.yaml` option. File looks like this:

```yaml
backend: memcached
backend_config:
  addresses:
    - localhost:11211

caching_config:
  chunk_block_size: 16000
  max_chunks_get_range_requests: 3
```

Only memcached client is supported at the moment. If `--caching-bucket.config-file` is not passed to the store (default), no chunks caching is done. (Index cache is unaffected by this)

## Verification

Tested manually (running queries against store with caching bucket) and using unit tests.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -19,6 +19,7 @@ We use *breaking* word for marking changes that are not backward compatible (rel
 
 - [#2502](https://github.com/thanos-io/thanos/pull/2502) Added `hints` field to `SeriesResponse`. Hints in an opaque data structure that can be used to carry additional information from the store and its content is implementation specific.
 - [#2521](https://github.com/thanos-io/thanos/pull/2521) Sidecar: add `thanos_sidecar_reloader_reloads_failed_total`, `thanos_sidecar_reloader_reloads_total`, `thanos_sidecar_reloader_watch_errors_total`, `thanos_sidecar_reloader_watch_events_total` and `thanos_sidecar_reloader_watches` metrics.
+- [#2532](https://github.com/thanos-io/thanos/pull/2532) Store: Added caching bucket, that can cache chunks into shared memcached. This can speed up querying and reduce number of requests to object storage.
 
 ### Changed
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -19,7 +19,7 @@ We use *breaking* word for marking changes that are not backward compatible (rel
 
 - [#2502](https://github.com/thanos-io/thanos/pull/2502) Added `hints` field to `SeriesResponse`. Hints in an opaque data structure that can be used to carry additional information from the store and its content is implementation specific.
 - [#2521](https://github.com/thanos-io/thanos/pull/2521) Sidecar: add `thanos_sidecar_reloader_reloads_failed_total`, `thanos_sidecar_reloader_reloads_total`, `thanos_sidecar_reloader_watch_errors_total`, `thanos_sidecar_reloader_watch_events_total` and `thanos_sidecar_reloader_watches` metrics.
-- [#2532](https://github.com/thanos-io/thanos/pull/2532) Store: Added caching bucket, that can cache chunks into shared memcached. This can speed up querying and reduce number of requests to object storage.
+- [#2532](https://github.com/thanos-io/thanos/pull/2532) Store: Added hidden option for experimental caching bucket, that can cache chunks into shared memcached. This can speed up querying and reduce number of requests to object storage.
 
 ### Changed
 
--- a/cmd/thanos/store.go
+++ b/cmd/thanos/store.go
@@ -55,6 +55,10 @@ func registerStore(m map[string]setupFunc, app *kingpin.Application) {
 		"YAML file that contains index cache configuration. See format details: https://thanos.io/components/store.md/#index-cache",
 		false)
 
+	cachingBucketConfig := extflag.RegisterPathOrContent(cmd, "experimental.caching-bucket.config",
+		"YAML file that contains configuration for caching bucket. Experimental feature, with high risk of changes. See format details: https://thanos.io/components/store.md/#TBD",
+		false)
+
 	chunkPoolSize := cmd.Flag("chunk-pool-size", "Maximum size of concurrently allocatable bytes reserved strictly to reuse for chunks in memory.").
 		Default("2GB").Bytes()
 
@@ -149,6 +153,7 @@ func registerStore(m map[string]setupFunc, app *kingpin.Application) {
 			*webExternalPrefix,
 			*webPrefixHeaderName,
 			*postingOffsetsInMemSampling,
+			cachingBucketConfig,
 		)
 	}
 }
@@ -179,6 +184,7 @@ func runStore(
 	ignoreDeletionMarksDelay time.Duration,
 	externalPrefix, prefixHeader string,
 	postingOffsetsInMemSampling int,
+	cachingBucketConfig *extflag.PathOrContent,
 ) error {
 	grpcProbe := prober.NewGRPC()
 	httpProbe := prober.NewHTTP()
@@ -212,6 +218,18 @@ func runStore(
 	bkt, err := client.NewBucket(logger, confContentYaml, reg, component.String())
 	if err != nil {
 		return errors.Wrap(err, "create bucket client")
+	}
+
+	cachingBucketConfigYaml, err := cachingBucketConfig.Content()
+	if err != nil {
+		return errors.Wrap(err, "get caching bucket configuration")
+	}
+	if len(cachingBucketConfigYaml) > 0 {
+		newbkt, err := storecache.NewCachingBucketFromYaml(cachingBucketConfigYaml, bkt, logger, reg)
+		if err != nil {
+			return errors.Wrap(err, "create caching bucket")
+		}
+		bkt = newbkt
 	}
 
 	relabelContentYaml, err := selectorRelabelConf.Content()
--- a/cmd/thanos/store.go
+++ b/cmd/thanos/store.go
@@ -225,11 +225,10 @@ func runStore(
 		return errors.Wrap(err, "get caching bucket configuration")
 	}
 	if len(cachingBucketConfigYaml) > 0 {
-		newbkt, err := storecache.NewCachingBucketFromYaml(cachingBucketConfigYaml, bkt, logger, reg)
+		bkt, err = storecache.NewCachingBucketFromYaml(cachingBucketConfigYaml, bkt, logger, reg)
 		if err != nil {
 			return errors.Wrap(err, "create caching bucket")
 		}
-		bkt = newbkt
 	}
 
 	relabelContentYaml, err := selectorRelabelConf.Content()
--- a/cmd/thanos/store.go
+++ b/cmd/thanos/store.go
@@ -55,7 +55,7 @@ func registerStore(m map[string]setupFunc, app *kingpin.Application) {
 		"YAML file that contains index cache configuration. See format details: https://thanos.io/components/store.md/#index-cache",
 		false)
 
-	cachingBucketConfig := extflag.RegisterPathOrContent(cmd, "store.caching-bucket.config",
+	cachingBucketConfig := extflag.RegisterPathOrContent(extflag.HiddenCmdClause(cmd), "store.caching-bucket.config",
 		"YAML that contains configuration for caching bucket. Experimental feature, with high risk of changes. See format details: https://thanos.io/components/store.md/#caching-bucket",
 		false)
 
--- a/cmd/thanos/store.go
+++ b/cmd/thanos/store.go
@@ -55,8 +55,8 @@ func registerStore(m map[string]setupFunc, app *kingpin.Application) {
 		"YAML file that contains index cache configuration. See format details: https://thanos.io/components/store.md/#index-cache",
 		false)
 
-	cachingBucketConfig := extflag.RegisterPathOrContent(cmd, "experimental.caching-bucket.config",
-		"YAML file that contains configuration for caching bucket. Experimental feature, with high risk of changes. See format details: https://thanos.io/components/store.md/#caching-bucket",
+	cachingBucketConfig := extflag.RegisterPathOrContent(cmd, "caching-bucket.config",
+		"YAML that contains configuration for caching bucket. Experimental feature, with high risk of changes. See format details: https://thanos.io/components/store.md/#caching-bucket",
 		false)
 
 	chunkPoolSize := cmd.Flag("chunk-pool-size", "Maximum size of concurrently allocatable bytes reserved strictly to reuse for chunks in memory.").
--- a/cmd/thanos/store.go
+++ b/cmd/thanos/store.go
@@ -56,7 +56,7 @@ func registerStore(m map[string]setupFunc, app *kingpin.Application) {
 		false)
 
 	cachingBucketConfig := extflag.RegisterPathOrContent(cmd, "experimental.caching-bucket.config",
-		"YAML file that contains configuration for caching bucket. Experimental feature, with high risk of changes. See format details: https://thanos.io/components/store.md/#TBD",
+		"YAML file that contains configuration for caching bucket. Experimental feature, with high risk of changes. See format details: https://thanos.io/components/store.md/#caching-bucket",
 		false)
 
 	chunkPoolSize := cmd.Flag("chunk-pool-size", "Maximum size of concurrently allocatable bytes reserved strictly to reuse for chunks in memory.").
--- a/cmd/thanos/store.go
+++ b/cmd/thanos/store.go
@@ -55,7 +55,7 @@ func registerStore(m map[string]setupFunc, app *kingpin.Application) {
 		"YAML file that contains index cache configuration. See format details: https://thanos.io/components/store.md/#index-cache",
 		false)
 
-	cachingBucketConfig := extflag.RegisterPathOrContent(cmd, "caching-bucket.config",
+	cachingBucketConfig := extflag.RegisterPathOrContent(cmd, "store.caching-bucket.config",
 		"YAML that contains configuration for caching bucket. Experimental feature, with high risk of changes. See format details: https://thanos.io/components/store.md/#caching-bucket",
 		false)
 
--- a/docs/components/store.md
+++ b/docs/components/store.md
@@ -232,6 +232,34 @@ While the remaining settings are **optional**:
 - `max_item_size`: maximum size of an item to be stored in memcached. This option should be set to the same value of memcached `-I` flag (defaults to 1MB) in order to avoid wasting network round trips to store items larger than the max item size allowed in memcached. If set to `0`, the item size is unlimited.
 - `dns_provider_update_interval`: the DNS discovery update interval.
 
+## Caching Bucket
+
+Thanos Store Gateway supports a "caching bucket" with chunks caching to speed up loading of chunks from TSDB blocks. Currently only memcached "backend" is supported:
+
+```yaml
+backend: memcached
+backend_config:
+  addresses:
+    - localhost:11211
+
+caching_config:
+  chunk_block_size: 16000
+  max_chunks_get_range_requests: 3
+  chunk_object_size_ttl: 24h
+  chunk_block_ttl: 24h
+```
+
+`backend_config` field for memcached supports all the same configuration as memcached for [index cache](#memcached-index-cache).
+
+`caching_config` is a configuration for chunks cache.
+
+- `chunk_block_size`: size of segment of chunks object that is stored to the cache. This is the smallest unit that chunks cache is working with.
+- `max_chunks_get_range_requests`: how many "get range" sub-requests may cache perform to fetch missing blocks.
+- `chunk_object_size_ttl`: how long to keep information about chunk file length in the cache.
+- `chunk_block_ttl`: how long to keep individual blocks in the cache.
+
+Note that chunks cache is an experimental feature, and these fields may be renamed or removed completely in the future.
+
 ## Index Header
 
 In order to query series inside blocks from object storage, Store Gateway has to know certain initial info about each block such as:
--- a/docs/components/store.md
+++ b/docs/components/store.md
@@ -251,7 +251,7 @@ caching_config:
 
 `backend_config` field for memcached supports all the same configuration as memcached for [index cache](#memcached-index-cache).
 
-`caching_config` is a configuration for chunks cache.
+`caching_config` is a configuration for chunks cache and supports the following optional settings:
 
 - `chunk_block_size`: size of segment of chunks object that is stored to the cache. This is the smallest unit that chunks cache is working with.
 - `max_chunks_get_range_requests`: how many "get range" sub-requests may cache perform to fetch missing blocks.
--- a/docs/components/store.md
+++ b/docs/components/store.md
@@ -243,20 +243,20 @@ backend_config:
     - localhost:11211
 
 caching_config:
-  chunk_block_size: 16000
+  chunk_subrange_size: 16000
   max_chunks_get_range_requests: 3
   chunk_object_size_ttl: 24h
-  chunk_block_ttl: 24h
+  chunk_subrange_ttl: 24h
 ```
 
 `backend_config` field for memcached supports all the same configuration as memcached for [index cache](#memcached-index-cache).
 
 `caching_config` is a configuration for chunks cache and supports the following optional settings:
 
-- `chunk_block_size`: size of segment of chunks object that is stored to the cache. This is the smallest unit that chunks cache is working with.
-- `max_chunks_get_range_requests`: how many "get range" sub-requests may cache perform to fetch missing blocks.
+- `chunk_subrange_size`: size of segment of chunks object that is stored to the cache. This is the smallest unit that chunks cache is working with.
+- `max_chunks_get_range_requests`: how many "get range" sub-requests may cache perform to fetch missing subranges.
 - `chunk_object_size_ttl`: how long to keep information about chunk file length in the cache.
-- `chunk_block_ttl`: how long to keep individual blocks in the cache.
+- `chunk_subrange_ttl`: how long to keep individual subranges in the cache.
 
 Note that chunks cache is an experimental feature, and these fields may be renamed or removed completely in the future.
 
--- a/pkg/cache/cache.go
+++ b/pkg/cache/cache.go
@@ -15,6 +15,7 @@ type Cache interface {
 	// Note that individual byte buffers may be retained by the cache!
 	Store(ctx context.Context, data map[string][]byte, ttl time.Duration)
 
-	// Fetch multiple keys from cache.
-	Fetch(ctx context.Context, keys []string) (found map[string][]byte, missing []string)
+	// Fetch multiple keys from cache. Returns map of input keys to data.
+	// If key isn't in the map, data for given key was not found.
+	Fetch(ctx context.Context, keys []string) map[string][]byte
 }
--- a/pkg/cache/cache.go
+++ b/pkg/cache/cache.go
@@ -1,0 +1,16 @@
+package cache
+
+import (
+	"context"
+	"time"
+)
+
+// Generic cache
+type Cache interface {
+	// Store data into the cache. If data for given key is nil, data is removed from cache instead.
+	// Only positive ttl values are used.
+	Store(ctx context.Context, data map[string][]byte, ttl time.Duration)
+
+	// Fetch multiple keys from cache.
+	Fetch(ctx context.Context, keys []string) (found map[string][]byte, missing []string)
+}
--- a/pkg/cache/cache.go
+++ b/pkg/cache/cache.go
@@ -1,3 +1,6 @@
+// Copyright (c) The Thanos Authors.
+// Licensed under the Apache License 2.0.
+
 package cache
 
 import (
@@ -7,8 +10,9 @@ import (
 
 // Generic cache
 type Cache interface {
-	// Store data into the cache. If data for given key is nil, data is removed from cache instead.
-	// Only positive ttl values are used.
+	// Store data into the cache.
+	//
+	// Note that individual byte buffers may be retained by the cache!
 	Store(ctx context.Context, data map[string][]byte, ttl time.Duration)
 
 	// Fetch multiple keys from cache.
--- a/pkg/cache/cache.go
+++ b/pkg/cache/cache.go
@@ -8,7 +8,7 @@ import (
 	"time"
 )
 
-// Generic cache
+// Generic best-effort cache.
 type Cache interface {
 	// Store data into the cache.
 	//
--- a/pkg/cache/memcached.go
+++ b/pkg/cache/memcached.go
@@ -1,0 +1,85 @@
+// Copyright (c) The Thanos Authors.
+// Licensed under the Apache License 2.0.
+
+package cache
+
+import (
+	"context"
+	"time"
+
+	"github.com/go-kit/kit/log"
+	"github.com/go-kit/kit/log/level"
+	"github.com/prometheus/client_golang/prometheus"
+	"github.com/prometheus/client_golang/prometheus/promauto"
+
+	"github.com/thanos-io/thanos/pkg/cacheutil"
+)
+
+// MemcachedCache is a memcached-based cache.
+type MemcachedCache struct {
+	logger    log.Logger
+	memcached cacheutil.MemcachedClient
+
+	// Metrics.
+	requests prometheus.Counter
+	hits     prometheus.Counter
+}
+
+// NewMemcachedCache makes a new MemcachedCache.
+func NewMemcachedCache(logger log.Logger, memcached cacheutil.MemcachedClient, reg prometheus.Registerer) *MemcachedCache {
+	c := &MemcachedCache{
+		logger:    logger,
+		memcached: memcached,
+	}
+
+	c.requests = promauto.With(reg).NewCounter(prometheus.CounterOpts{
+		Name: "thanos_store_memcached_requests_total",
+		Help: "Total number of items requests to the memcached.",
+	})
+
+	c.hits = promauto.With(reg).NewCounter(prometheus.CounterOpts{
+		Name: "thanos_store_memcached_hits_total",
+		Help: "Total number of items requests to the cache that were a hit.",
+	})
+
+	level.Info(logger).Log("msg", "created memcached cache")
+
+	return c
+}
+
+// Store data identified by keys.
+// The function enqueues the request and returns immediately: the entry will be
+// asynchronously stored in the cache.
+func (c *MemcachedCache) Store(ctx context.Context, data map[string][]byte, ttl time.Duration) {
+	for key, val := range data {
+		if err := c.memcached.SetAsync(ctx, key, val, ttl); err != nil {
+			level.Error(c.logger).Log("msg", "failed to cache postings in memcached", "err", err)
+		}
+	}
+}
+
+// FetchMultiPostings fetches multiple postings - each identified by a label -
+// and returns a map containing cache hits, along with a list of missing keys.
+// In case of error, it logs and return an empty cache hits map.
+func (c *MemcachedCache) Fetch(ctx context.Context, keys []string) (map[string][]byte, []string) {
+	// Fetch the keys from memcached in a single request.
+	c.requests.Add(float64(len(keys)))
+	results := c.memcached.GetMulti(ctx, keys)
+	if len(results) == 0 {
+		return nil, keys
+	}
+
+	// Construct the resulting hits map and list of missing keys. We iterate on the input
+	// list of labels to be able to easily create the list of ones in a single iteration.
+	var misses []string
+	for _, key := range keys {
+		_, ok := results[key]
+		if !ok {
+			misses = append(misses, key)
+			continue
+		}
+	}
+
+	c.hits.Add(float64(len(results)))
+	return results, misses
+}
--- a/pkg/cache/memcached.go
+++ b/pkg/cache/memcached.go
@@ -26,20 +26,22 @@ type MemcachedCache struct {
 }
 
 // NewMemcachedCache makes a new MemcachedCache.
-func NewMemcachedCache(logger log.Logger, memcached cacheutil.MemcachedClient, reg prometheus.Registerer) *MemcachedCache {
+func NewMemcachedCache(name string, logger log.Logger, memcached cacheutil.MemcachedClient, reg prometheus.Registerer) *MemcachedCache {
 	c := &MemcachedCache{
 		logger:    logger,
 		memcached: memcached,
 	}
 
 	c.requests = promauto.With(reg).NewCounter(prometheus.CounterOpts{
-		Name: "thanos_cache_memcached_requests_total",
-		Help: "Total number of items requests to the memcached.",
+		Name:        "thanos_cache_memcached_requests_total",
+		Help:        "Total number of items requests to memcached.",
+		ConstLabels: prometheus.Labels{"name": name},
 	})
 
 	c.hits = promauto.With(reg).NewCounter(prometheus.CounterOpts{
-		Name: "thanos_cache_memcached_hits_total",
-		Help: "Total number of items requests to the cache that were a hit.",
+		Name:        "thanos_cache_memcached_hits_total",
+		Help:        "Total number of items requests to the cache that were a hit.",
+		ConstLabels: prometheus.Labels{"name": name},
 	})
 
 	level.Info(logger).Log("msg", "created memcached cache")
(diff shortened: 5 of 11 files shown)


pkg/objstore/caching/readers.go
```diff
@@ -0,0 +1,245 @@
...
+	// Offset for next read, used to find correct block to return data from.
+	readOffset int64
+
+	// Remaining data to return from this reader. Once zero, this reader reports EOF.
+	remaining int64
+}
+
+func newBlocksReader(blockSize int64, blocks blocksProvider, readOffset, remaining int64) *blocksReader {
+	return &blocksReader{
+		blockSize:  blockSize,
```
pracucci, line 39: In other places it's called `chunkBlockSize`.

pkg/objstore/caching/caching_bucket.go
```diff
@@ -0,0 +1,290 @@
...
+		chunkBlockTTL:        24 * time.Hour,
+
+		cachedBytes: promauto.With(reg).NewCounter(prometheus.CounterOpts{
+			Name: "thanos_objstore_caching_bucket_cached_chunk_bytes_total",
+			Help: "Total number of chunk blocks used from cache",
+		}),
+		fetchedBytes: promauto.With(reg).NewCounter(prometheus.CounterOpts{
+			Name: "thanos_objstore_caching_bucket_fetched_chunk_bytes_total",
+			Help: "Total number of chunk blocks fetched from storage",
+		}),
```
pracucci, line 65: If we pick the path I've suggested to make this implementation specific to the thanos store (and thus specific cache chunks) we can make these metrics more specific.

pkg/objstore/caching/readers.go
```diff
@@ -0,0 +1,245 @@
...
+		bytesCounter: bytesCounter,
+	}
+}
+
+func (c cachedBlocks) blockAt(offset int64) ([]byte, error) {
+	b := c.blocks[c.offsetsKeys[offset]]
+	if b == nil {
+		return nil, errors.Errorf("cachedBlocks: block for offset %d not found", offset)
+	}
+	c.bytesCounter.Add(float64(len(b)))
```
pracucci, line 103: This counter overcounts. The problem is that:
1. `blockAt()` can be potentially called multiple times for the same block (depends on the buffer passed to `Read()`)
2. If it's the first or last block, the reader may actually read less then whole bucket size

A different option may have `blockAt()` returning 3rd parameter to let the caller know whether the block has been picked from the cache or not. The caller can then correctly increment the counter when the block comes from the cache.

pkg/objstore/caching/readers.go
```diff
@@ -0,0 +1,245 @@
...
+
+func (c cachedBlocks) close() error {
+	return nil
+}
+
+// Blocks provider that reads each block from bucket and also stores it into cache.
+type fetchingBlocksProvider struct {
+	blockSize int64
+
+	blockOffset int64  // Offset of the current block/
```
pracucci, line 115:
```suggestion
	blockOffset int64  // Offset of the current block.
```

bwplotka:
You are way too fast with this =D 

We can't catch up @pstibrany ;p Just joined Google doc party with late comments: https://docs.google.com/document/d/1knNXQfDJlQDA9Dr9R9OVFvdWS83sXd0kLEFdv21IxCM/edit# Sorry for lag!

I know @brancz wanted to add some input as point as well. (: 


Overall commented the TL;DR of my comments but will comment here as well:

----- 

Amazing, thanks for this proposal!

This is really good, my only suggestion is that it looks like we can really create a generic client-side cache layer for object storage (or maybe server-side at some point).

I think for a start we can create a generic cache layer for Bucket interface backed up by memcached. Maybe we can find a caching logic that will work for both metadata, postings/series, and chunk caching? The idea is to simplify this layer which will also allow extensibility and easy maintenance.

Then we can iterate to move it different backends OR maybe totally move as localized server-side cache for the whole cluster... ((:

pkg/cache/cache.go
```diff
@@ -0,0 +1,20 @@
...
+
+// Generic cache
+type Cache interface {
+	// Store data into the cache.
+	//
+	// Note that individual byte buffers may be retained by the cache!
+	Store(ctx context.Context, data map[string][]byte, ttl time.Duration)
+
+	// Fetch multiple keys from cache.
+	Fetch(ctx context.Context, keys []string) (found map[string][]byte, missing []string)
```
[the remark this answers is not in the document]
pstibrany: You mean on `Fetch`, `Store` or both?

I think keeping them error-less better shows the intent that these are best-effort caches. It also simplifies client. Key itself is typically quite information-rich already. But I can be convinced.

pkg/objstore/caching/caching_bucket.go
```diff
@@ -0,0 +1,290 @@
...
+	}
+}
+
+func cachingKeyObjectSize(name string) string {
+	return fmt.Sprintf("size:%s", name)
+}
+
+func cachingKeyObjectBlock(name string, start int64, end int64) string {
+	return fmt.Sprintf("block:%s:%d:%d", name, start, end)
+}
```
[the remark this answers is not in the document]
pstibrany: I don't think it will be a problem here, where name looks like `/01E1TR0Q8MKHM3BC3JX1JE6DP3/chunks/000007` (40 bytes) plus two offsets in 512 MiB file (`536870756:536870756` is 19 bytes), but it's something to keep in mind.

pkg/objstore/caching/caching_bucket.go
```diff
@@ -0,0 +1,290 @@
...
+		length = int64(size - uint64(offset))
+	}
+
+	startOffset := (offset / cb.chunkBlockSize) * cb.chunkBlockSize
+	endOffset := ((offset + length) / cb.chunkBlockSize) * cb.chunkBlockSize
+	if (offset+length)%cb.chunkBlockSize > 0 {
+		endOffset += cb.chunkBlockSize
+	}
+
+	blocks := (endOffset - startOffset) / cb.chunkBlockSize
```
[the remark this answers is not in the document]
pstibrany: Files in "chunks" directory are called segment files already. It seems that all good names are used already. 😞

pkg/objstore/caching/readers.go
pstibrany: I see `blocksReader` as being generic reader implementation, while `chunkBlockSize` is related to chunks.

pkg/objstore/caching/readers.go
pstibrany: Correct. This is fixed by simpler reader implementation.

pkg/objstore/caching/caching_bucket.go
```diff
@@ -0,0 +1,290 @@
+// Copyright (c) The Thanos Authors.
+// Licensed under the Apache License 2.0.
+
+package caching
```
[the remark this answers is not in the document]
pstibrany: Caching bucket has been moved to storecache, and is available for configuration only when running store.

pkg/objstore/caching/caching_bucket.go
pstibrany: I've renamed them to `thanos_store_caching_bucket_cached_chunk_bytes_total` and `thanos_store_caching_bucket_fetched_chunk_bytes_total`.

pstibrany:
I think this is ready for another review.

pstibrany:
docs-tests failed with

```
docs/community.md
 
	ERROR	https://zoom.us/j/96476403386
 
		Error when reading response headers: small read buffer. Increase ReadBufferSize. Buffer size=4096, contents: "HTTP/1.1 200 \r\nDate: Thu, 30 Apr 2020 08:52:17 GMT\r\nContent-Type: text/html;charset=utf-8\r\nTransfer-Encoding: chunked\r\nConnection: keep-alive\r\nServer: ZOOM\r\nX-Zm-Trackingid: WEB_b1b0451a7c7a6d0f1339b4"..."=44766e66c4f14818a82eb55b073f8388; Domain=.zoom.us; Path=/; Secure\r\nContent-Language: en-US\r\nStrict-Transport-Security: max-age=31536000\r\nX-Xss-Protection: 1; mode=block\r\nReferrer-Policy: strict-origi"
 ```

which seems unrelated to my changes.

**Comment by bwplotka** [comment]
Please rebase, I think it's fixed on master.

File: pkg/cache/cache.go (hunk 5 above)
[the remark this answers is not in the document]
Comment (pracucci) [reply pkg/cache/cache.go:19]: I did mean both, but I see your point about making the cache a best-effort. My only concern is that we can't accurately log/track errors inside a generic cache implementation, while if we had this information in the caller we may track it more accurately. However, I do agree that the interface design looks clearer/easier without returning an `error`.

File: pkg/store/cache/caching_bucket.go [hunk 9]
```diff
@@ -0,0 +1,389 @@
...
+	keys := make([]string, 0, blocks)
+	lastBlockLength := 0
+
+	for o := startOffset; o < endOffset; o += cb.config.ChunkBlockSize {
+		end := o + cb.config.ChunkBlockSize
+		if end > int64(size) {
+			end = int64(size)
+		}
+
+		lastBlockLength = int(end - o)
```
pracucci, line 200: Despite correct, looks a bit weird updating the `lastBlockLength` at every `for` iteration when actually it's just the last iteration the correct one. You can compute the `lastBlockLength` in a 1-line when you define the variable above. Have you considered doing it?

pkg/store/cache/caching_bucket.go
```diff
@@ -0,0 +1,389 @@
...
+	startOffset := (offset / cb.config.ChunkBlockSize) * cb.config.ChunkBlockSize
+	endOffset := ((offset + length) / cb.config.ChunkBlockSize) * cb.config.ChunkBlockSize
+	if (offset+length)%cb.config.ChunkBlockSize > 0 {
+		endOffset += cb.config.ChunkBlockSize
+	}
+
+	blocks := (endOffset - startOffset) / cb.config.ChunkBlockSize
+
+	offsetKeys := make(map[int64]string, blocks)
+	keys := make([]string, 0, blocks)
```
pracucci, line 191: This is unused

pkg/store/cache/caching_bucket.go
```diff
@@ -0,0 +1,389 @@
...
+		length = int64(size - uint64(offset))
+	}
+
+	startOffset := (offset / cb.config.ChunkBlockSize) * cb.config.ChunkBlockSize
+	endOffset := ((offset + length) / cb.config.ChunkBlockSize) * cb.config.ChunkBlockSize
+	if (offset+length)%cb.config.ChunkBlockSize > 0 {
+		endOffset += cb.config.ChunkBlockSize
+	}
+
+	blocks := (endOffset - startOffset) / cb.config.ChunkBlockSize
```
pracucci, line 188: `blocks` > `numBlocks` may be clearer

pkg/store/cache/caching_bucket.go
```diff
@@ -0,0 +1,389 @@
...
+	}
+
+	// Try to get all blocks from the cache.
+	hits, _ := cb.cache.Fetch(ctx, keys)
+
+	for _, b := range hits {
+		cb.cachedChunkBytes.Add(float64(len(b)))
+	}
+
+	if len(hits) < len(keys) {
```
pracucci, line 214: You can remove `keys` at all if you just do `if len(hits) < blocks`

pkg/store/cache/caching_bucket.go
```diff
@@ -0,0 +1,389 @@
...
+	missing = mergeRanges(missing, 0) // Merge adjacent ranges.
+	// Keep merging until we have only max number of ranges (= requests).
+	for limit := cb.config.ChunkBlockSize; cb.config.MaxChunksGetRangeRequests > 0 && len(missing) > cb.config.MaxChunksGetRangeRequests; limit = limit * 2 {
+		missing = mergeRanges(missing, limit)
+	}
+
+	var hitsMutex sync.Mutex
+
+	// Run parallel queries for each missing range. Fetched data is stored into 'hits' map, protected by hitsMutex.
+	eg, nctx := errgroup.WithContext(ctx)
```
pracucci, line 252: To keep consistency with the rest of the code I would suggest to rename `eg` to `g` and `nctx` to `gctx`

pkg/store/cache/caching_bucket.go
```diff
@@ -0,0 +1,389 @@
...
+			for off := m.start; off < m.end && nctx.Err() == nil; off += cb.config.ChunkBlockSize {
+				key := cacheKeys[off]
+				if key == "" {
+					return errors.Errorf("fetching range [%d, %d]: caching key for offset %d not found", m.start, m.end, off)
+				}
+
+				// We need a new buffer for each block, both for storing into hits, and also for caching.
+				blockData := make([]byte, cb.config.ChunkBlockSize)
+				n, err := io.ReadFull(r, blockData)
+				if err == io.ErrUnexpectedEOF && (off+cb.config.ChunkBlockSize) == endOffset && n == lastBlockLength {
```
pracucci, line 271: I think always considering `ErrUnexpectedEOF` safe for the last block is risky. Why don't we initialise `blockData` accurately for the last block and we never check for `err == io.ErrUnexpectedEOF`?

pkg/store/cache/caching_bucket.go
```diff
@@ -0,0 +1,389 @@
...
+				if err != nil {
+					return errors.Wrapf(err, "fetching range [%d, %d]", m.start, m.end)
+				}
+
+				hitsMutex.Lock()
+				hits[key] = blockData
+				hitsMutex.Unlock()
+
+				cb.fetchedChunkBytes.Add(float64(len(blockData)))
+				cb.cache.Store(nctx, map[string][]byte{key: blockData}, cb.config.ChunkBlockTTL)
```
pracucci, line 285: Due to how the `mergeRanges`  work, this block could already be in `hits`. In such case, we should definitely not restore it in the cache (it's a waste of resources).

I'm wondering what we should do with the `fetchedChunkBytes` metric. On one side it's data we actually re-fetched from the storage so increasing it could be correct. On the other side if you expect `fetchedChunkBytes + cachedChunkBytes` to be equal to the total size of data loaded from chunks then increasing it here is not correct. What do you think if we split `fetchedChunkBytes` into two: "fetched" and "refetched" (the latter would be the total number of bytes we wasted loading just to keep the number of storage API calls low)?

pkg/store/cache/caching_bucket.go
```diff
@@ -0,0 +1,389 @@
...
+		return nil, errors.New("cache is nil")
+	}
+
+	return &CachingBucket{
+		config: chunks,
+		bucket: b,
+		cache:  c,
+		logger: logger,
+
+		cachedChunkBytes: promauto.With(reg).NewCounter(prometheus.CounterOpts{
```
pracucci, line 73: From my perspective, would be nice to also see counters tracking the number of sub-object blocks:
- Total number of sub-object blocks requested
- Total number of sub-object blocks cache hit

pkg/cache/memcached.go
```diff
@@ -0,0 +1,84 @@
...
+	}
+
+	// Construct the resulting hits map and list of missing keys. We iterate on the input
+	// list of labels to be able to easily create the list of ones in a single iteration.
+	var misses []string
+	for _, key := range keys {
+		_, ok := results[key]
+		if !ok {
+			misses = append(misses, key)
+			continue
```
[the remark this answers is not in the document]
pstibrany: I actually think returning `misses` is not necessary. Anymore. One can get that information by iterating `keys` and checking returned map. I will remove it completely.

pkg/cache/memcached.go
```diff
@@ -0,0 +1,84 @@
...
+func (c *MemcachedCache) Fetch(ctx context.Context, keys []string) (map[string][]byte, []string) {
+	// Fetch the keys from memcached in a single request.
+	c.requests.Add(float64(len(keys)))
+	results := c.memcached.GetMulti(ctx, keys)
+	if len(results) == 0 {
+		return nil, keys
+	}
+
+	// Construct the resulting hits map and list of missing keys. We iterate on the input
+	// list of labels to be able to easily create the list of ones in a single iteration.
```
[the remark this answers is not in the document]
pstibrany: Removed completely. Client can easily do this if needed.

pkg/store/cache/caching_bucket.go
```diff
@@ -0,0 +1,389 @@
...
+	MaxChunksGetRangeRequests int `yaml:"max_chunks_get_range_requests"`
+
+	// TTLs for various cache items.
+	ChunkObjectSizeTTL time.Duration `yaml:"chunk_object_size_ttl"`
+	ChunkBlockTTL      time.Duration `yaml:"chunk_block_ttl"`
+}
+
+func DefaultCachingBucketConfig() CachingBucketConfig {
+	return CachingBucketConfig{
+		ChunkBlockSize:            16000, // Equal to max chunk size.
```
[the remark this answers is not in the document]
pstibrany: Unfortunately this would lead to import cycle. We could move `maxChunkSize` to some other package, I'm not sure it's worth it though.

```
import cycle not allowed
package github.com/thanos-io/thanos/cmd/thanos
	imports github.com/thanos-io/thanos/pkg/query
	imports github.com/thanos-io/thanos/pkg/store
	imports github.com/thanos-io/thanos/pkg/store/cache
	imports github.com/thanos-io/thanos/pkg/store
```

pkg/store/cache/caching_bucket.go
```diff
@@ -0,0 +1,389 @@
...
+		ChunkObjectSizeTTL:        24 * time.Hour,
+		ChunkBlockTTL:             24 * time.Hour,
+		MaxChunksGetRangeRequests: 1,
```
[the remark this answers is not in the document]
pstibrany: I'll go with 3 for no specific reason. I can change it.

pkg/store/cache/caching_bucket.go
```diff
@@ -0,0 +1,389 @@
...
+}
+
+var chunksMatcher = regexp.MustCompile(`^.*/chunks/\d+$`)
+
+func isTSDBChunkFile(name string) bool {
+	return chunksMatcher.MatchString(name)
+}
+
+func (cb *CachingBucket) GetRange(ctx context.Context, name string, off, length int64) (io.ReadCloser, error) {
+	if isTSDBChunkFile(name) && off >= 0 && length > 0 {
```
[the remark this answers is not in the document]
pstibrany: > Why do we need to add the check `off >= 0 && length > 0`? Some thoughts:
> 
> 1. Can we consider `off < 0` == offset zero?
> 2. Given the first thing we do in `getRangeChunkFile()` is reading the object size, could we just set `length = object size` if `length < 0`?

Both conditions you describe are incorrect and I think hands-off approach is better. We let underlying bucket handle it.

Specifically I think that normalizing incorrect values is a bad idea, because then the behavior changes when used with and without caching bucket.

pkg/store/cache/caching_bucket.go
```diff
@@ -0,0 +1,389 @@
...
+		return nil, errors.Wrapf(err, "failed to get size of chunk file: %s", name)
+	}
+
+	// If length goes over object size, adjust length. We use it later to limit number of read bytes.
+	if uint64(offset+length) > size {
+		length = int64(size - uint64(offset))
+	}
+
+	startOffset := (offset / cb.config.ChunkBlockSize) * cb.config.ChunkBlockSize
+	endOffset := ((offset + length) / cb.config.ChunkBlockSize) * cb.config.ChunkBlockSize
```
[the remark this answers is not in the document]
pstibrany: "offset" is simply a position. I will rename it to `startRange`/`endRange` though.

pkg/store/cache/caching_bucket.go
pstibrany: `keys` are used in `hits := cb.cache.Fetch(ctx, keys)` call.

pkg/store/cache/caching_bucket.go
pstibrany: > I think always considering `ErrUnexpectedEOF` safe for the last block is risky. 

There is one extra check -- whether `n` matches `lastBlockLength`, so there is no risk.

> Why don't we initialise `blockData` accurately for the last block and we never check for `err == io.ErrUnexpectedEOF`?

Yes, it will be nicer.

pkg/store/cache/caching_bucket.go
pstibrany: I think "blocks" in caching bucket are implementation detail that helps caching, but I don't think we should leak that to user. I would therefore suggest we keep bytes instead. WDYT?

We can definitely add metric for "requested".

pkg/store/cache/caching_bucket.go
pstibrany: I will add counter for "refetched" bytes.

pkg/store/cache/caching_bucket.go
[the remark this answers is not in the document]
pracucci: I haven't considered the cyclic dependency. The correlation here is quite low with the `maxChunkSize` so I'm OK to keep it as i.

docs/components/store.md
```diff
@@ -232,6 +232,34 @@ While the remaining settings are **optional**:
...
+caching_config:
+  chunk_block_size: 16000
+  max_chunks_get_range_requests: 3
+  chunk_object_size_ttl: 24h
+  chunk_block_ttl: 24h
+```
+
+`backend_config` field for memcached supports all the same configuration as memcached for [index cache](#memcached-index-cache).
+
+`caching_config` is a configuration for chunks cache.
```
pracucci, line 254:
```suggestion
`caching_config` is a configuration for chunks cache and supports the following optional settings:
```

pkg/store/cache/caching_bucket.go
pracucci: I see your point. I agree. The new metrics you've added, already covers several uses cases. Further metrics can always be added in the future.

pkg/store/cache/caching_bucket.go
```diff
@@ -0,0 +1,430 @@
...
+
+	// Start and end range are block-aligned offsets into object, that we're going to read.
+	startRange := (offset / cb.config.ChunkBlockSize) * cb.config.ChunkBlockSize
+	endRange := ((offset + length) / cb.config.ChunkBlockSize) * cb.config.ChunkBlockSize
+	if (offset+length)%cb.config.ChunkBlockSize > 0 {
+		endRange += cb.config.ChunkBlockSize
+	}
+
+	// The very last block in the object may have length that is not divisible by block size.
+	lastBlockOffset := int64(-1)
```
pracucci, line 216: Despite correct (due to how the current logic works) setting it to the correct offset may be more future proof. I think you can easily set it to `endRange - cb.config.ChunkBlockSize` thus covering all cases.

pkg/store/cache/caching_bucket.go
```diff
@@ -0,0 +1,430 @@
...
+			Name: "thanos_store_bucket_cache_requested_chunk_bytes_total",
+			Help: "Total number of requested bytes for chunk data ",
+		}),
+		cachedChunkBytes: promauto.With(reg).NewCounter(prometheus.CounterOpts{
+			Name: "thanos_store_bucket_cache_cached_chunk_bytes_total",
+			Help: "Total number of chunk bytes used from cache",
+		}),
+		fetchedChunkBytes: promauto.With(reg).NewCounter(prometheus.CounterOpts{
+			Name: "thanos_store_bucket_cache_fetched_chunk_bytes_total",
+			Help: "Total number of chunk bytes fetched from storage",
```
pracucci, line 89:
```suggestion
			Help: "Total number of chunk bytes fetched from storage because missing from cache",
```

pkg/store/cache/caching_bucket.go
```diff
@@ -0,0 +1,430 @@
...
+
+	return &CachingBucket{
+		config: chunks,
+		bucket: b,
+		cache:  c,
+		logger: logger,
+
+		requestedChunkBytes: promauto.With(reg).NewCounter(prometheus.CounterOpts{
+			Name: "thanos_store_bucket_cache_requested_chunk_bytes_total",
+			Help: "Total number of requested bytes for chunk data ",
```
pracucci, line 81:
```suggestion
			Help: "Total number of requested bytes for chunk data",
```

pracucci, line 81: Could you add a dot (`.`) at the end of each `Help` missing it, to keep consistency with Thanos codebase, please?

pstibrany:
We have some numbers to share from testing this PR (as of commit 9821c29).

We were testing with following query from our dashboard with 30 days time range:
```sum by (status) (label_replace(label_replace(rate(cortex_request_duration_seconds_count{cluster=~"dev-us-central1/agent|dev-us-central1", job=~"(loki-dev|loki-boltdb-shipper|default|cortex-tsdb-dev|cortex-enterprise|cortex-chunks)/query-frontend"}[2400s]), "status", "${1}xx", "status_code", "([0-9]).."), "status", "${1}",   "status_code", "([a-z]+)"))```

This query needs to fetch about 20 MiB of chunks data.

We used separate cache for index-cache and chunks-cache. Only chunks cache was emptied between tests, not chunks-cache. We also used query-frontend in our setup, which split the query into smaller time ranges that were executed in parallel, but with caching of results disabled.

Times were measured with `time` command, timing `curl`.

| empty cache | with full cache |
|--|--|
| 1.934s | 0.822s |
| 1.732s | 0.827s |
| 1.818s | 0.819s |

Here are two screenshots, one showing trace with empty cache, with many 'multicache gets', followed by 'set' for fetched blocks (`cortex-store-gateway` is doing job of `store` component in Thanos)

Screenshot 2020-05-05 at 12 16 14

Trace with cached data:

Screenshot 2020-05-05 at 12 13 55

File: CHANGELOG.md [hunk 27]
```diff
@@ -19,6 +19,7 @@ We use *breaking* word for marking changes that are not backward compatible (rel
 
 - [#2502](https://github.com/thanos-io/thanos/pull/2502) Added `hints` field to `SeriesResponse`. Hints in an opaque data structure that can be used to carry additional information from the store and its content is implementation specific.
 - [#2521](https://github.com/thanos-io/thanos/pull/2521) Sidecar: add `thanos_sidecar_reloader_reloads_failed_total`, `thanos_sidecar_reloader_reloads_total`, `thanos_sidecar_reloader_watch_errors_total`, `thanos_sidecar_reloader_watch_events_total` and `thanos_sidecar_reloader_watches` metrics.
+- [#2532](https://github.com/thanos-io/thanos/pull/2532) Store: Added caching bucket, that can cache chunks into shared memcached. This can speed up querying and reduce number of requests to object storage.
```
bwplotka, line 22:
```suggestion
- [#2532](https://github.com/thanos-io/thanos/pull/2532) Store: Added hidden option for experimental caching bucket, that can cache chunks into shared memcached. This can speed up querying and reduce number of requests to object storage. 
```

bwplotka, line 22: I guess no in-mem option is planned :thinking: 

I think that is a good pattern and simplify a lot.

cmd/thanos/store.go
```diff
@@ -55,6 +55,10 @@ func registerStore(m map[string]setupFunc, app *kingpin.Application) {
 		"YAML file that contains index cache configuration. See format details: https://thanos.io/components/store.md/#index-cache",
 		false)
 
+	cachingBucketConfig := extflag.RegisterPathOrContent(cmd, "experimental.caching-bucket.config",
```
bwplotka, line 58: I know we were doing opposite but now when I think of it I think it would be nice to have this is hidden but normal flag name. 

It's painful to change flag names for users ... (: 

so maybe just 
```suggestion
	cachingBucketConfig := extflag.RegisterPathOrContent(cmd, "store.caching-bucket.config",
```

cmd/thanos/store.go
```diff
@@ -55,6 +55,10 @@ func registerStore(m map[string]setupFunc, app *kingpin.Application) {
...
+		"YAML file that contains configuration for caching bucket. Experimental feature, with high risk of changes. See format details: https://thanos.io/components/store.md/#caching-bucket",
```
bwplotka, line 59:
```suggestion
		"YAML that contains configuration for caching bucket. Experimental feature, with high risk of changes. See format details: https://thanos.io/components/store.md/#caching-bucket",
```

This helper will add file or content descriptions

cmd/thanos/store.go
```diff
@@ -214,6 +220,17 @@ func runStore(
 		return errors.Wrap(err, "create bucket client")
 	}
 
+	cachingBucketConfigYaml, err := cachingBucketConfig.Content()
+	if err != nil {
+		return errors.Wrap(err, "get caching bucket configuration")
+	}
+	if len(cachingBucketConfigYaml) > 0 {
+		bkt, err = storecache.NewCachingBucketFromYaml(cachingBucketConfigYaml, bkt, logger, reg)
```
bwplotka, line 228: :+1:

docs/components/store.md
```diff
@@ -232,6 +232,34 @@ While the remaining settings are **optional**:
 - `max_item_size`: maximum size of an item to be stored in memcached. This option should be set to the same value of memcached `-I` flag (defaults to 1MB) in order to avoid wasting network round trips to store items larger than the max item size allowed in memcached. If set to `0`, the item size is unlimited.
 - `dns_provider_update_interval`: the DNS discovery update interval.
 
+## Caching Bucket
```
bwplotka, line 235: Thanks for docs straight away!

pkg/store/cache/caching_bucket.go
```diff
@@ -0,0 +1,444 @@
...
+
+	return cb, nil
+}
+
+func (cb *CachingBucket) Close() error {
+	return cb.bucket.Close()
+}
+
+func (cb *CachingBucket) Upload(ctx context.Context, name string, r io.Reader) error {
+	return cb.bucket.Upload(ctx, name, r)
```
bwplotka, line 118: What do you think about embedding objstore.Bucket so we can totally avoid writting those boilerplate methods?

pkg/store/cache/caching_bucket.go
```diff
@@ -0,0 +1,444 @@
...
+}
+
+func (cb *CachingBucket) Delete(ctx context.Context, name string) error {
+	return cb.bucket.Delete(ctx, name)
+}
+
+func (cb *CachingBucket) Name() string {
+	return "caching: " + cb.bucket.Name()
```
bwplotka, line 126: hmmmmm I think Name is what we want but does not matter much, I think we only log this

pkg/store/cache/caching_bucket.go
```diff
@@ -0,0 +1,444 @@
...
+}
+
+func (cb *CachingBucket) WithExpectedErrs(expectedFunc objstore.IsOpFailureExpectedFunc) objstore.Bucket {
+	if ib, ok := cb.bucket.(objstore.InstrumentedBucket); ok {
```
bwplotka, line 130: Can this happen?

bwplotka: I think it's perferred to use NopInstrumentedBucket or something like this we created in `objstore`

CHANGELOG.md
pstibrany: We can surely add it, if it is needed, but it will help only in small installations.

cmd/thanos/store.go
pstibrany: How can we hide it?

pkg/store/cache/caching_bucket.go
pstibrany: It is extra boilerplate, but I think it's also cleaner design-wise. Plus I'm going to add logic to other methods in next PR (for metadata caching).

pkg/store/cache/caching_bucket.go
pstibrany: Should I change this or not?

pkg/store/cache/caching_bucket.go
pstibrany: Can you please explain what you mean, and if you would like me to do anything here? :) Thanks!

CHANGELOG.md
bwplotka: Let's skip it , it's good practice with Memcached for now

cmd/thanos/store.go
bwplotka: I guess by adding `RegisterHiddenPathOrContent` or something (:

pkg/store/cache/caching_bucket.go
bwplotka: This is fine, you can overwrite the method always. I think it really depends what is the consistent approach. If we have boilerplate like this and only one method is changes we always embed, but it's up to you here, not a blocker.

pkg/store/cache/caching_bucket.go
bwplotka: :shrug:

pkg/store/cache/caching_bucket.go
bwplotka: Ignore me I misunderstood the code all good here

pkg/store/cache/caching_bucket.go
```diff
@@ -0,0 +1,444 @@
...
+		// Make a copy, but replace bucket with instrumented one.
+		res := &CachingBucket{}
+		*res = *cb
```
bwplotka, line 133: Do we nead a copy?

pkg/store/cache/caching_bucket.go
bwplotka: But hm I would really just do embedding. The key here is that is will show us CLEARLY what caching is adding on top of normal obj store ):

pkg/store/cache/caching_bucket.go
```diff
@@ -0,0 +1,444 @@
...
+			hits = map[string][]byte{}
+		}
+
+		err := cb.fetchMissingChunkBlocks(ctx, name, startRange, endRange, offsetKeys, hits, lastBlockOffset, lastBlockLength)
+		if err != nil {
+			return nil, err
+		}
+	}
+
+	return newBlocksReader(cb.config.ChunkBlockSize, offsetKeys, hits, offset, length), nil
```
bwplotka, line 270: Very dangerous in Thanos to name a block reader, something that has nothing to do with TSDB block. Can we find different naming? (:

pkg/store/cache/caching_bucket.go
```diff
@@ -0,0 +1,444 @@
...
+}
+
+type rng struct {
+	start, end int64
+}
+
+// Fetches missing blocks and stores them into "hits" map.
+func (cb *CachingBucket) fetchMissingChunkBlocks(ctx context.Context, name string, startRange, endRange int64, cacheKeys map[int64]string, hits map[string][]byte, lastBlockOffset int64, lastBlockLength int) error {
```
bwplotka, line 278: even if this is not exported it's really nice if comments starts with method name.

bwplotka, line 278: what's the default chunks block size? :thinking:

pkg/store/cache/caching_bucket.go
```diff
@@ -0,0 +1,444 @@
...
+
+func cachingKeyObjectSize(name string) string {
+	return fmt.Sprintf("size:%s", name)
+}
+
+func cachingKeyObjectBlock(name string, start int64, end int64) string {
+	return fmt.Sprintf("block:%s:%d:%d", name, start, end)
+}
+
+// io.ReadCloser implementation that uses in-memory blocks.
```
bwplotka, line 376:
```suggestion
// blocksReader is an io.Reader implementation that uses in-memory blocks.
```

bwplotka: but again I would strongly suggest using some other wording like `segment`

pkg/store/cache/caching_bucket_factory.go
```diff
@@ -0,0 +1,59 @@
...
+	"github.com/thanos-io/thanos/pkg/objstore"
+)
+
+type BucketCacheProvider string
+
+const MemcachedBucketCacheProvider BucketCacheProvider = "memcached"
+
+type CachingBucketWithBackendConfig struct {
+	Type          BucketCacheProvider `yaml:"backend"`
+	BackendConfig interface{}         `yaml:"backend_config"`
```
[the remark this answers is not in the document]
bwplotka: Cool, let's maybe add TODO to rename others?

pkg/store/cache/caching_bucket_factory.go
```diff
@@ -0,0 +1,59 @@
...
+	"gopkg.in/yaml.v2"
+
+	cache "github.com/thanos-io/thanos/pkg/cache"
+	"github.com/thanos-io/thanos/pkg/cacheutil"
+	"github.com/thanos-io/thanos/pkg/objstore"
+)
+
+type BucketCacheProvider string
+
+const MemcachedBucketCacheProvider BucketCacheProvider = "memcached"
```
bwplotka, line 20: Would be nice to comment all exported types.

pstibrany:
> LGTM, just I would love if we could change `block` into `segment` to not create confusion with TSDB blocks.

Problem with “segment” is that it already has a meaning in TSDB, and I don’t want to reuse it.

I’ve found some synonyms to [“chunk”](https://www.thesaurus.com/browse/chunk?s=t) that may work: hunk, lump, piece, portion, ... I personally like hunk, because it’s subword of “chunk” :) WDYT?

> Does it work well with partitioning? Looks like it partition by chunk size and partitioning is not aware of this 🤔

Partitioner merges multiple small requests into bigger one. That helps to save on number of requests, which is good. Chunking cache then “rounds” starting/ending offsets to hunk (!) size, and caches them individually.

I think it works together just fine, I’m not sure what could be improved.  We certainly don’t want to skip gap-sized partitioner. From caching perspective, it may be useful to reduce gap check (it defaults to 512 KiB if I remember correctly), but that would possibly increase number of requests, which is not what we want. Ideas are welcome.

pstibrany:
@bwplotka Thanks for your review! I will address your feedback tomorrow.

pracucci:
> I’ve found some synonyms to “chunk” that may work: hunk, lump, piece, portion, ... I personally like hunk, because it’s subword of “chunk” :) WDYT?

@pstibrany @bwplotka I'm personally dubious about "hunk" because of few reasons:
1. Finger memory risks to type "chunk"
2. While reading code, it's easy to misread "hunk" as "chunk"
3. "hunk" looks like a typo of "chunk"

I already had this discussion with Peter and at the beginning I was also suggesting "segment", but I also agree that's not much better then "block" given both names already mean something different in TSDB. Have you considered `subrange`? What we're caching is practically an offset-aligned subrange, it looks clear and doesn't clash with any other naming we already have.

cmd/thanos/store.go
pstibrany: Oh, are you suggesting to add `RegisterHiddenPathOrContent` that will "hide" the command line option – how? By giving it "hidden" prefix? I'm little confused about what you're suggesting here.

pkg/store/cache/caching_bucket.go
pstibrany: OK

pkg/store/cache/caching_bucket.go
pstibrany: keeping as-is for now then

pkg/store/cache/caching_bucket.go
pstibrany: This code copies cache, config, logger and metrics, but replaces bucket with "instrumented one". Without doing copy, we would need to copy all fields manually.

pkg/store/cache/caching_bucket.go
pstibrany: embedding will expose underlying bucket in a public field, instead of private (not a big deal, just saying)

pkg/store/cache/caching_bucket.go
pstibrany:
```go
func DefaultCachingBucketConfig() CachingBucketConfig {
	return CachingBucketConfig{
		ChunkBlockSize: 16000, // Equal to max chunk size.
```

cmd/thanos/store.go
pstibrany: Hidden by using new `extflag.HiddenCmdClause(cmd)` like this:

```
	cachingBucketConfig := extflag.RegisterPathOrContent(extflag.HiddenCmdClause(cmd), "store.caching-bucket.config", ...)
```

WDYT?

pkg/store/cache/caching_bucket.go
pstibrany: Renamed to "subrange" everywhere in this code.

pkg/store/cache/caching_bucket_factory.go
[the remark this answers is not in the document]
pstibrany: Created issue for this: https://github.com/thanos-io/thanos/issues/2570

pstibrany:
I believe I've now addressed all feedback, and this PR is hopefully ready for final (?) review and merge!

pstibrany:
Thanks a lot for your review efforts! You can look forward to more caching PRs real soon :)

bwplotka:
Yes, especially around meta fetching (:

PR #2533, opened 2020-04-28
cmd: rule: do not wrap reload endpoint with prefix twice

GiedriusS:
Let's make `/-/reload` accessible again by not wrapping the router with the prefix twice. Add a simple test-case to automatically check whether this works.

## Changes

* Do not wrap `/-/reload` with the prefix twice

## Verification

`make test-e2e`

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -9,11 +9,12 @@ NOTE: As semantic versioning states all 0.y.z releases can contain breaking chan
 
 We use *breaking* word for marking changes that are not backward compatible (relates only to v0.y.z releases.)
 
-## Unreleased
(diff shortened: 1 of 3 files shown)


CHANGELOG.md
```diff
@@ -9,6 +9,10 @@ NOTE: As semantic versioning states all 0.y.z releases can contain breaking chan
 
 We use *breaking* word for marking changes that are not backward compatible (relates only to v0.y.z releases.)
 
+## [v0.12.2](https://github.com/thanos-io/thanos/releases/tag/v0.12.2) - 2020.04.XX
+
```
simonpasquier, line 13: does it need a `### Fixed` header?

GiedriusS: Thanks, I have added it :+1:

Issue #2344, opened 2020-03-30
compact: Allows backing up blocks instead of straight removing to another bucket.


bwplotka:
AC:
* Compactor is moving blocks to separate object storage instead of removing them during all( ??) block rewrites.

Something to consider which came up in the discussion with our users. (:

GiedriusS:
> AC:
> 
>     * Compactor is moving blocks to separate object storage instead of removing them during all block rewrites.
> 
> 
> Something to consider which came up in the discussion with our users. (:

I guess it would be extra nice to have this when we will have series deletion. :+1:

bwplotka:
Definitely. That's the idea, since the storing itself is cheap

bwplotka:
This needs further discussion / design as per https://github.com/thanos-io/thanos/pull/2541#issuecomment-622447561

It's silly to just backup compaction blocks as not data is removed. However it might make sense to back up removals / vertical complications (:

bwplotka:
As per discussion on our public meeting, we would aim for back up mode, where we back up ALL operations that can lead to data removal, which is:
* Vertical / offline dedup  /  per series retention compaction (not normal one!)
* Retention deletion
* Delete series someday

cc @hitha1234 @darshanime

bwplotka:
still valid

jujugrrr:
hi @bwplotka , not sure it's the best issue to ask the question, but I couldn't find a recommended way to backup Thanos data. Will this issue allow to use the compactor to do this?

I've seen we can do `thanos bucket replicate` this as well, but is it not reliable because the compactor might corrupt data while it's replicating?

jujugrrr:
> hi @bwplotka , not sure it's the best issue to ask the question, but I couldn't find a recommended way to backup Thanos data(for disaster recovery). Will this issue allow to use the compactor to do this?
> 
> I've seen we can do `thanos bucket replicate` this as well, but is it not reliable because the compactor might corrupt data while it's replicating?

Would S3 bucket replication + `thanos bucket verify --repair` be a better approach. The replication won't delete files in the destination bucket  when deleted from the source bucket, but the verify might fix the overlap/duplicate created by the compactor from the source bucket?

PR #2541, opened 2020-04-30
Allow blocks to be retained after compaction

darshanime:

Add ability to retain blocks after compaction. 
Fixes https://github.com/thanos-io/thanos/issues/2344

### DIFF
--- a/cmd/thanos/compact.go
+++ b/cmd/thanos/compact.go
@@ -149,6 +149,8 @@ func registerCompact(m map[string]setupFunc, app *kingpin.Application) {
 	webPrefixHeaderName := cmd.Flag("web.prefix-header", "Name of HTTP request header used for dynamic prefixing of UI links and redirects. This option is ignored if web.external-prefix argument is set. Security risk: enable this option only if a reverse proxy in front of thanos is resetting the header. The --web.prefix-header=X-Forwarded-Prefix option can be useful, for example, if Thanos UI is served via Traefik reverse proxy with PathPrefixStrip option enabled, which sends the stripped prefix value in X-Forwarded-Prefix header. This allows thanos UI to be served on a sub-path.").Default("").String()
 	label := cmd.Flag("bucket-web-label", "Prometheus label to use as timeline title in the bucket web UI").String()
 
+	retainCompactedBlocks := cmd.Flag("retain-compacted-blocks", "Do not remove old blocks even after they are safely compacted, but move them to another directory").Default("false").Bool()
+
 	m[component.Compact.String()] = func(g *run.Group, logger log.Logger, reg *prometheus.Registry, tracer opentracing.Tracer, _ <-chan struct{}, _ bool) error {
 		return runCompact(g, logger, reg,
 			*httpAddr,
@@ -177,6 +179,7 @@ func registerCompact(m map[string]setupFunc, app *kingpin.Application) {
 			*label,
 			*webExternalPrefix,
 			*webPrefixHeaderName,
+			*retainCompactedBlocks,
 		)
 	}
 }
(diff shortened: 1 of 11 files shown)


cmd/thanos/compact.go
```diff
@@ -149,6 +149,8 @@ func registerCompact(m map[string]setupFunc, app *kingpin.Application) {
 	webPrefixHeaderName := cmd.Flag("web.prefix-header", "Name of HTTP request header used for dynamic prefixing of UI links and redirects. This option is ignored if web.external-prefix argument is set. Security risk: enable this option only if a reverse proxy in front of thanos is resetting the header. The --web.prefix-header=X-Forwarded-Prefix option can be useful, for example, if Thanos UI is served via Traefik reverse proxy with PathPrefixStrip option enabled, which sends the stripped prefix value in X-Forwarded-Prefix header. This allows thanos UI to be served on a sub-path.").Default("").String()
 	label := cmd.Flag("bucket-web-label", "Prometheus label to use as timeline title in the bucket web UI").String()
 
+	retainCompactedBlocks := cmd.Flag("retain-compacted-blocks", "Do not remove old blocks even after they are safely compacted, but move them to another directory").Default("false").Bool()
```
bwplotka, line 152: Cool, we were thinking about this, but actually we though about separate bucket :thinking: 
Thoughts?

Essentially backup mode

belm0:
document the use case in the compactor docs?

darshanime:
change to retain in separate bucket ✔️ 
added a note about usecase in docs ✔️

docs/components/compact.md
```diff
@@ -49,6 +49,8 @@ Ideally, you will have equal retention set (or no retention at all) to all resol
 
 Not setting this flag, or setting it to `0d`, i.e. `--retention.resolution-X=0d`, will mean that samples at the `X` resolution level will be kept forever.
 
+Optionally, you can also retain blocks which are compacted into larger blocks. You can supply `--objstore-retain-compacted-blocks-to-bucket.config` and Thanos will move compacted blocks to the specified bucket. This can be useful if you do not want to lose blocks during rewrites, which can happen in compaction, delete series etc.
```
belm0, line 52: it's not clear-- does this apply to only the original blocks, or also to higher-order compactions?

if the goal is to keep a backup of the original prometheus blocks, it seems the latter is not necessary

bwplotka:
Thanks for this but I think we need to have larger discussion first.

For me backing up all of the blocks might be not efficient - we are actually replicate tons of things for backup purposes, which is just not needed. 

The perfect way TBH would be to use `thanos bucket replicate` to always backup 2h blocks to some cold storage if you want an overall backup.

If you want backup only for retention then just stating NO retention and adding some `min` time for compactor might make more sense.

What's the use case for you? I would say it's a perfect item to discuss on our Thanos public community meeting if you want to join: https://thanos.io/community.md/

darshanime:
Hi @bwplotka 👋 

To be honest, I am not an end user of Thanos, but am trying to contribute to Thanos so as to participate and learn more about building distributed systems. 

I came across the issue #2344 and thought I'd try to implement it since it relates to `delete_series` (which is a project I am hoping to get selected for in Community Bridge), . 
I should have asked for more details/prepared a document to get concensus before working on it, will keep that in mind going forward 👍

bwplotka:
Thanks for this.

It's a good pattern to first describe quickly how you plan to do it so we can quickly discuss it.

Anyway, this is only my opinion, maybe it's fine to duplicate backup... Some opinions from other @thanos-io/thanos-maintainers would be nice :hugs:

brancz:
I think in particular because we can never know when compaction happens it's probably safest to put backup logic into it, even if that somewhat overlaps with the replication tooling we provide (as those don't have the same safety guarantees as the compactor backup process would be able to provide).

darshanime:
Hi, can this be closed? 
Agree a better PR can be made after getting consensus on what the best retention strategy would be.

kakkoyun:
Closing this as @darshanime requested.

PR #2544, opened 2020-04-30
Clear duplicateIDs at the beginning of Filter.

pstibrany:
This PR fixes problem with duplicate blocks field not being cleared at the beginning of Filter.

If not cleared, such blocks are then reused between runs. Once deleted due to deletion delay, deletion markers are recreated, and on next meta sync, they are deleted again (as partial blocks).

## Verification

Added unit test.

Closes issue #2459.

To be merged into release-0.12 as suggested by @bwplotka

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -13,7 +13,7 @@ We use *breaking* word for marking changes that are not backward compatible (rel
 
 ### Fixed
 
-- [#2459](https://github.com/thanos-io/thanos/issues/2459) Compact: Fixed issue with old blocks being marked and deleted in a (slow) loop. 
+- [#2459](https://github.com/thanos-io/thanos/issues/2459) Compact: Fixed issue with old blocks being marked and deleted in a (slow) loop.
 
 ## [v0.12.1](https://github.com/thanos-io/thanos/releases/tag/v0.12.1) - 2020.04.20
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -9,6 +9,12 @@ NOTE: As semantic versioning states all 0.y.z releases can contain breaking chan
 
 We use *breaking* word for marking changes that are not backward compatible (relates only to v0.y.z releases.)
 
+## Next release
+
+### Fixed
+
+- [#2459](https://github.com/thanos-io/thanos/issues/2459) Compact: Fixed issue with old blocks being marked and deleted in a (slow) loop. 
+
 ## [v0.12.1](https://github.com/thanos-io/thanos/releases/tag/v0.12.1) - 2020.04.20
 
 ### Fixed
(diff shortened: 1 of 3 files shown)


CHANGELOG.md
```diff
@@ -9,6 +9,12 @@ NOTE: As semantic versioning states all 0.y.z releases can contain breaking chan
 
 We use *breaking* word for marking changes that are not backward compatible (relates only to v0.y.z releases.)
 
+## Next release
```
bwplotka, line 12:
```suggestion
## Unreleased
```

pkg/block/fetcher.go
```diff
@@ -561,6 +561,8 @@ func NewDeduplicateFilter() *DeduplicateFilter {
 // Filter filters out duplicate blocks that can be formed
 // from two or more overlapping blocks that fully submatches the source blocks of the older blocks.
 func (f *DeduplicateFilter) Filter(_ context.Context, metas map[ulid.ULID]*metadata.Meta, synced *extprom.TxGaugeVec, _ bool) error {
+	f.duplicateIDs = nil
```
bwplotka, line 564: Can we reuse array? (:

```suggestion
	f.duplicateIDs = f.duplicateIDs[:0]
```

bwplotka: As per: https://thanos.io/contributing/coding-style-guide.md/#reuse-arrays

pkg/compact/compact_e2e_test.go
```diff
@@ -416,3 +417,101 @@ func createAndUpload(t testing.TB, bkt objstore.Bucket, blocks []blockgenSpec) (
 	}
 	return metas
 }
+
+func Test_Issue2459_e2e(t *testing.T) {
```
bwplotka, line 421: So what we do usually is just normal Test name and comment that `// Regression test for #xx issue`

bwplotka: Wonder if we can embed this into some case for `fetcher_test` rather. What do you think? (: Just simpler.

pkg/compact/compact_e2e_test.go
```diff
@@ -416,3 +417,101 @@ func createAndUpload(t testing.TB, bkt objstore.Bucket, blocks []blockgenSpec) (
...
+		rem, err := listBlocksMarkedForDeletion(ctx, bkt)
+		testutil.Ok(t, err)
+
+		sort.Slice(rem, func(i, j int) bool {
+			return rem[i].Compare(rem[j]) < 0
+		})
+
+		testutil.Equals(t, ids, rem)
+
+		// delete source blocks
```
bwplotka, line 486: plus full sentence  comments. please (:

pstibrany:
Thanks for review Bartek, I've addressed your feedback.

pkg/compact/compact_e2e_test.go
bwplotka: Maybe it was found, but it's best to test the contract of the filter not how we mitigated it outside (:  Not a blocker.

PR #2548, opened 2020-04-30
querier: Adjust deduplication for counters when querying for PromQL rates.

bwplotka:
cc @SuperQ 

Chained over: https://github.com/thanos-io/thanos/pull/2528

Fixes:  https://github.com/thanos-io/thanos/issues/2401


bwplotka:
Works perfectly now :heart: 

image

bwplotka:
Let's not merge before https://github.com/thanos-io/thanos/pull/2528 merges.

pkg/query/iter.go
```diff
@@ -587,3 +593,40 @@ func (it *dedupSeriesIterator) Err() error {
 	}
 	return it.b.Err()
 }
+
+// counterDedupAdjustSeriesIterator is used when we deduplicate counter.
+// It makes sure we always adjust for the latest seen last counter value for all replicas.
```
brancz, line 598: for clarity I think it would be good to elaborate _how_ it adjusts, as in the strategy/heuristic, I think I can tell by reading the code and following the investigation etc., but a comment to confirm would be helpful for me in 3 months :)

bwplotka: Yes, good point, will do. Thanks!

bwplotka: Done: https://github.com/thanos-io/thanos/pull/2548/commits/e903228a93b93015271ec9a3c5bb3dfd308e4b4e

Does it make sense?

bwplotka:
Let's merge once https://github.com/thanos-io/thanos/pull/2528 get green mark.

pkg/query/iter.go
```diff
@@ -587,3 +593,56 @@ func (it *dedupSeriesIterator) Err() error {
...
+		SeriesIterator: iter,
+		lastV:          -1 * math.MaxFloat64,
+	}
+
+}
+
+func (it *counterDedupAdjustSeriesIterator) Next() bool {
+	if it.SeriesIterator.Next() {
+		_, v := it.SeriesIterator.At()
+		if it.lastV > v {
```
GiedriusS, line 636: Hmm, I was thinking but maybe a stupid question: shouldn't we set `it.adjust` back to 0 if we will jump back to a replica with a higher counter value again at some point during the iteration? Wouldn't that lead to higher counter values than usual without such an adjustment?

bwplotka: Hmm, you might be right. Adjust has to be cumulative, but per replica not globally, otherwise, we are adjusting true value... This might mean we need to

1. Actually hook inside the dedup iterator and adjust per replica (fine)
2, Or reset adjust as you propose (easier, but we lost the info that replica was behind).

Also if we do 1 it's a bit broken anyway as we know about WHEN & WHAT to adjust only if we are lucky that replica changes (: Still it's better than what we had before, plus better than any solution in the world for this (e.g dedup on ingest Cortex have)

cc @brancz @brian-brazil 

Will think about it a bit more, thanks for spotting this!

bwplotka: I fixed this problem, but changed code a little bit, please review @GiedriusS and thanks for spotting this! I added test cases for this

bwplotka:
PTAL @GiedriusS @brancz

bwplotka:
Let's merge on Monday. I plan to add benchmarks for this, plus consider to rewrite to heap approach, plus few other fixes, but let's do on later PR maybe (:

bwplotka:
Let's merge without squash (:

bwplotka:
And of course I squashed ;p

PR #2557, opened 2020-05-04
Multitsdb migration

brancz:

## Changes

This adds automatic migration of legacy storage to multi-tsdb disk layout, where the existing storage is migrated to the default tenant.

This builds on top of https://github.com/thanos-io/thanos/pull/2012, so until that merges, this includes those changes, but only the latest commit is relevant.

## Verification

Tried upgrading locally from a legacy TSDB disk layout.

@bwplotka @squat @metalmatze @krasi-georgiev @blockloop

### DIFF
--- a/cmd/thanos/receive.go
+++ b/cmd/thanos/receive.go
@@ -6,7 +6,9 @@ package main
 import (
 	"context"
 	"fmt"
+	"io/ioutil"
 	"os"
+	"path"
 	"strings"
 	"time"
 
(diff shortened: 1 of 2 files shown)


cmd/thanos/receive.go
```diff
@@ -226,6 +228,13 @@ func runReceive(
 		level.Info(logger).Log("msg", "no supported bucket was configured, uploads will be disabled")
 	}
 
+	// TODO(brancz): remove after a couple of versions
+	// Migrate non-multi-tsdb capable storage to multi-tsdb disk layout.
+	err = migrateLegacyStorage(logger, dataDir, defaultTenantID)
```
squat, line 233: nit: I think we typically make this a one-liner

PR #2558, opened 2020-05-04
Refactored proto generation and separated store from rules APIs.

bwplotka:
Just refactor, no logic was changed.

NOTE: This is major change to how we build our photos (:  ( I think good one, thanks for this we can import and reuse things, while maintaining different Go packages.

cc @s-urbaniak 


s-urbaniak:
lgtm :+1:

bwplotka:
With this we separated rules and store APIs and servers into different directories :hugs: 

image

bwplotka:
Giving time for others to review this. @s-urbaniak let's build up on top of this PR (:

pkg/promclient/promclient.go
```diff
@@ -46,6 +48,8 @@ var (
 		http.StatusServiceUnavailable:  codes.Unavailable,
 		http.StatusInternalServerError: codes.Internal,
 	}
+
+	ThanosUserAgent = fmt.Sprintf("Thanos/%s", version.Version)
```
GiedriusS, line 52: `pkg/http/http.go` has the same exact user-agent string. Maybe it would be worth reusing that?

pkg/query/storeset.go
```diff
@@ -215,14 +216,15 @@ func NewStoreSet(
 	return ss
 }
 
+// TODO(bwplotka): Consider moving out of this package and renam as it also supports rules API.
```
GiedriusS, line 219:
```suggestion
// TODO(bwplotka): Consider moving this out of this package and renaming it as it also supports rules API.
```

pkg/rules/prometheus_test.go
```diff
@@ -0,0 +1,178 @@
...
+				},
+			},
+			// TODO(bwplotka): Potentially add more cases.
+		},
+	} {
+		t.Run("", func(t *testing.T) {
+			srv := &rulesServer{ctx: ctx}
+			err = promRules.Rules(&rulespb.RulesRequest{}, srv)
+			if tcase.expectedErr != nil {
+				testutil.NotOk(t, err)
```
GiedriusS, line 152: Minor thing but this is unnecessary since we check for equality on the next line and `tcase.expectedErr` is guaranteed not to be `nil`. BTW, great test :smile:

pkg/rules/proxy.go
```diff
@@ -0,0 +1,153 @@
...
+)
+
+// Proxy implements rulespb.Rules that fanouts requests to given rulespb.Rules and deduplication on the way.
+type Proxy struct {
+	logger log.Logger
+	rules  func() []rulespb.RulesClient
+}
+
+// NewProxy returns new rules.Proxy.
+func NewProxy(logger log.Logger, rules func() []rulespb.RulesClient) *Proxy {
```
GiedriusS, line 28: Minor thing but the StoreAPI method is called `NewProxyStore` so maybe let's call this `NewProxyRule`?

pkg/rules/proxy.go
```diff
@@ -0,0 +1,153 @@
...
+	rules, err := stream.client.Rules(ctx, stream.request)
+	if err != nil {
+		err = errors.Wrapf(err, "fetching rules from rules client %v", stream.client)
+
+		if stream.request.PartialResponseStrategy == storepb.PartialResponseStrategy_ABORT {
+			return err
+		}
+
+		if err := stream.server.Send(rulespb.NewWarningRulesResponse(err)); err != nil {
+			return err
```
GiedriusS, line 116: Shouldn't we return here the `err` from the previous `if` no matter what? Because if sending this succeeds but calling `Rules()` previously does _not_ then we will continue on using `rules`.

pkg/rules/proxy_test.go
```diff
@@ -0,0 +1,157 @@
...
+				},
+				{
+					Name: "c",
+				},
+			},
+		},
+	} {
+		t.Run(tc.name, func(t *testing.T) {
+			got := dedupGroups(tc.groups)
+			if !reflect.DeepEqual(tc.want, got) {
```
GiedriusS, line 152: Is there some reason not to use `testutil.Equals`? :eye:

pkg/promclient/promclient.go
bwplotka: good point

pkg/rules/proxy.go
bwplotka: yup

bwplotka: oh actually it will be done on next PR

bwplotka: actually let's leave it Proxy. Store thing should be fixed IMO

pkg/rules/proxy.go
bwplotka: yup! Valid bug

pkg/rules/proxy_test.go
bwplotka: just copy pasta

bwplotka:
All addressed and rebased, please PTAL @s-urbaniak

cmd/thanos/rule.go
```diff
@@ -40,8 +40,8 @@ import (
 	"github.com/thanos-io/thanos/pkg/prober"
 	"github.com/thanos-io/thanos/pkg/promclient"
 	"github.com/thanos-io/thanos/pkg/query"
-	thanosrule "github.com/thanos-io/thanos/pkg/rule"
-	v1 "github.com/thanos-io/thanos/pkg/rule/api"
+	v1 "github.com/thanos-io/thanos/pkg/rules/api"
+	thanosmanager "github.com/thanos-io/thanos/pkg/rules/manager"
```
s-urbaniak, line 44: nit: maybe let's call the alias simply `rulesmanager` as `thanosmanager.Manager` as the context is lost.

pkg/rules/api/v1_test.go
```diff
@@ -66,7 +66,7 @@ type rulesRetrieverMock struct {
 	testing *testing.T
 }
 
-func (m rulesRetrieverMock) RuleGroups() []thanosrule.Group {
+func (m rulesRetrieverMock) RuleGroups() []manager.Group {
```
s-urbaniak, line 69: same here: let's call this alias `rulesmanager` so the context for `rulesmanager.Group` is not lost.

cmd/thanos/rule.go
bwplotka: good point

bwplotka:
done @s-urbaniak

pkg/rules/manager/rule.go
```diff
@@ -1,7 +1,7 @@
 // Copyright (c) The Thanos Authors.
 // Licensed under the Apache License 2.0.
 
-package thanosrule
+package manager
```
s-urbaniak, line 4: i wonder if we can name this more concrete. This implementation "manages" rule files, correct? Maybe even `rules/file` makes more sense here then? The struct `file.Manager` could stay, stating that it is "managing" those rule files.

bwplotka: maybe but TBH it's mainly a mangers that groups Prom ones.

s-urbaniak:
just naming nits, else lgtm :+1:

pkg/rules/manager/rule.go
s-urbaniak: sounds good then

bwplotka: actually we change this in next PR ;p so let's merge

PR #2559, opened 2020-05-04
Merge 0.12 into master

squat:


VERSION
```diff
@@ -1 +1 @@
-0.12.1
+0.12.2
```
metalmatze, line 1: IIRC we had something like `0.13.0-dev` on master last time I merged back. Not sure if we "enforce" this somehow?!

squat: :/ hm we don't have a strict process for this. master currently says `0.12.1` so this should be good to merge now. We can make a follow up PR to bump master to 0.13.0-dev and we should maybe put this in the release docs. WDYT?

bwplotka: yes please. 


We removed this because `promu` was handling this, but not anyone ):

bwplotka: let's fix in separate PR with release doc fix :+1:

bwplotka:
I did squash :facepalm: 

I guess we should revert and do normal merge? ):

PR #2571, opened 2020-05-07
Allow 3rd parties to customise objstore client prefix

pracucci:

## Changes

Cortex is going to use the Thanos objstore clients beyond the blocks storage and we would like to be able to customise the metrics prefix. I this PR I propose to add the `thanos_` prefix in the objstore client factory, so that whoever embed Thanos objstore can directly istantiate specific clients with custom metrics prefix.

## Verification

Unit tests.

### DIFF
--- a/pkg/objstore/client/factory.go
+++ b/pkg/objstore/client/factory.go
@@ -12,6 +12,7 @@ import (
 	"github.com/go-kit/kit/log/level"
 	"github.com/pkg/errors"
 	"github.com/prometheus/client_golang/prometheus"
+	"github.com/thanos-io/thanos/pkg/extprom"
 	"github.com/thanos-io/thanos/pkg/objstore"
 	"github.com/thanos-io/thanos/pkg/objstore/azure"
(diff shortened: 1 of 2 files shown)


pkg/objstore/client/factory.go
```diff
@@ -76,5 +76,11 @@ func NewBucket(logger log.Logger, confContentYaml []byte, reg prometheus.Registe
 	if err != nil {
 		return nil, errors.Wrap(err, fmt.Sprintf("create %s client", bucketConf.Type))
 	}
-	return objstore.BucketWithMetrics(bucket.Name(), bucket, reg), nil
+
+	var bucketReg prometheus.Registerer
+	if reg != nil {
```
bwplotka, line 81: Actually you can just pass it, no need for check IF you will use `exthttp.WrapRegisterer..`

pracucci: Ups, right. Thanks, done!

pracucci:
_Closing this to try to not mangle metrics prefixes at all in Cortex._

PR #2579, opened 2020-05-08
Metadata caching in bucket

pstibrany:
This PR does several things at once:

- Adds caching to more methods in `CachingBucket` (Get, Exists, Iter, ObjectSize)
- Makes caching bucket fully configurable: caching for each operation is configured via `Cache<OperationName>` method. This takes config name (used as label in metrics), cache, matcher function that is used to match object names, and configuration used for matching objects.
- Factory configures caching bucket to support:
    - Chunks caching: via `GetRange` method
    - Meta files caching (meta.json, deletion mark) via `Get` and `Exists` method
    - Caching of `Iter` for root directory
    - Caching of index files via `GetRange` and `ObjectSize` (this is mostly an example how to do it. We can discuss whether to do it now, or we wait)

Factory uses user-supplied configuration to configure operations based on how Thanos works.
Notice that factory uses single cache object for all of these, but cache is still passed as extra argument since we plan to pass different caches for different configurations in Cortex.

Opening PR to get feedback on design. I think code is also ready for review, but documentation hasn't been updated yet. Note that existing metrics have been modified, but configuration has been preserved and extended (there is no config for index cache, those values are just hardcoded. We can discuss whether to provide index cache config now, or in another PR which will also document situation around two different approaches to index cache).

/cc @pracucci @bwplotka 

* [ ] I have updated documentation.

## Verification

Unit tests for now. I have also tested earlier version of the code integrated into Cortex in our dev cluster, but will retest again.

### DIFF
--- a/pkg/store/cache/caching_bucket.go
+++ b/pkg/store/cache/caching_bucket.go
@@ -6,6 +6,7 @@ package storecache
 import (
 	"context"
 	"encoding/binary"
+	"encoding/json"
 	"fmt"
 	"io"
 	"io/ioutil"
@@ -14,6 +15,8 @@ import (
 	"time"
 
 	"github.com/go-kit/kit/log"
+	"github.com/go-kit/kit/log/level"
+	"github.com/golang/snappy"
 	"github.com/pkg/errors"
 	"github.com/prometheus/client_golang/prometheus"
 	"github.com/prometheus/client_golang/prometheus/promauto"
@@ -40,6 +43,9 @@ type CachingBucketConfig struct {
 	// TTLs for various cache items.
 	ChunkObjectSizeTTL time.Duration `yaml:"chunk_object_size_ttl"`
 	ChunkSubrangeTTL   time.Duration `yaml:"chunk_subrange_ttl"`
+
+	// How long to cache result of Iter call.
+	IterTTL time.Duration `yaml:"iter_ttl"`
 }
 
 func DefaultCachingBucketConfig() CachingBucketConfig {
@@ -48,6 +54,8 @@ func DefaultCachingBucketConfig() CachingBucketConfig {
 		ChunkObjectSizeTTL:        24 * time.Hour,
 		ChunkSubrangeTTL:          24 * time.Hour,
 		MaxChunksGetRangeRequests: 3,
+
+		IterTTL: 5 * time.Minute,
 	}
 }
 
@@ -67,6 +75,9 @@ type CachingBucket struct {
 
 	objectSizeRequests prometheus.Counter
 	objectSizeHits     prometheus.Counter
+
+	iterRequests  prometheus.Counter
+	iterCacheHits prometheus.Counter
 }
 
 func NewCachingBucket(b objstore.Bucket, c cache.Cache, chunks CachingBucketConfig, logger log.Logger, reg prometheus.Registerer) (*CachingBucket, error) {
@@ -103,6 +114,15 @@ func NewCachingBucket(b objstore.Bucket, c cache.Cache, chunks CachingBucketConf
 			Name: "thanos_store_bucket_cache_objectsize_hits_total",
 			Help: "Number of object size hits for objects.",
 		}),
+
+		iterRequests: promauto.With(reg).NewCounter(prometheus.CounterOpts{
+			Name: "thanos_store_bucket_cache_iter_requests_total",
+			Help: "Number of Iter requests for directories.",
+		}),
+		iterCacheHits: promauto.With(reg).NewCounter(prometheus.CounterOpts{
+			Name: "thanos_store_bucket_cache_iter_cache_hits_total",
+			Help: "Number of Iter requests served from cache.",
+		}),
 	}
 
 	cb.fetchedChunkBytes.WithLabelValues(originBucket)
@@ -130,6 +150,70 @@ func (cb *CachingBucket) WithExpectedErrs(expectedFunc objstore.IsOpFailureExpec
 
 func (cb *CachingBucket) ReaderWithExpectedErrs(expectedFunc objstore.IsOpFailureExpectedFunc) objstore.BucketReader {
 	return cb.WithExpectedErrs(expectedFunc)
+}
+
+func (cb *CachingBucket) Iter(ctx context.Context, dir string, f func(string) error) error {
+	cb.iterRequests.Inc()
+
+	key := cachingKeyIter(dir)
+
+	data := cb.cache.Fetch(ctx, []string{key})
+	if data[key] != nil {
+		list, err := decodeIterResult(data[key])
+		if err == nil {
+			cb.iterCacheHits.Inc()
+
+			for _, n := range list {
+				err = f(n)
+				if err != nil {
+					return err
+				}
+			}
+			return nil
+		} else {
+			// This should not happen.
+			level.Warn(cb.logger).Log("msg", "failed to decode cached Iter result", "err", err)
+		}
+	}
+
+	// Iteration can take a while (esp. since it calls function), and iterTTL is generally low.
+	// We will compute TTL based time when iteration started.
+	iterTime := time.Now()
+	var list []string
+	err := cb.Bucket.Iter(ctx, dir, func(s string) error {
+		list = append(list, s)
+		return f(s)
+	})
+
+	remainingTTL := cb.config.IterTTL - time.Since(iterTime)
+	if err == nil && remainingTTL > 0 {
+		data := encodeIterResult(list)
+		if data != nil {
+			cb.cache.Store(ctx, map[string][]byte{key: data}, remainingTTL)
+		}
+	}
+	return err
+}
+
+// Iter results should compress nicely, especially in subdirectories.
+func encodeIterResult(files []string) []byte {
+	data, err := json.Marshal(files)
+	if err != nil {
+		return nil
+	}
+
+	return snappy.Encode(nil, data)
+}
+
+func decodeIterResult(data []byte) ([]string, error) {
+	decoded, err := snappy.Decode(nil, data)
+	if err != nil {
+		return nil, err
+	}
+
+	var list []string
+	err = json.Unmarshal(decoded, &list)
+	return list, err
 }
 
 var chunksMatcher = regexp.MustCompile(`^.*/chunks/\d+$`)
@@ -342,6 +426,10 @@ func cachingKeyObjectSize(name string) string {
 
 func cachingKeyObjectSubrange(name string, start int64, end int64) string {
 	return fmt.Sprintf("subrange:%s:%d:%d", name, start, end)
+}
+
+func cachingKeyIter(name string) string {
+	return fmt.Sprintf("iter:%s", name)
 }
 
 // Reader implementation that uses in-memory subranges.
--- a/pkg/store/cache/caching_bucket.go
+++ b/pkg/store/cache/caching_bucket.go
@@ -36,6 +36,12 @@ const (
 
 	existsTrue  = "true"
 	existsFalse = "false"
+
+	operationGet        = "get"
+	operationGetRange   = "getrange"
+	operationIter       = "iter"
+	operationExists     = "exists"
+	operationObjectSize = "objectsize"
 )
 
 var errObjNotFound = errors.Errorf("object not found")
@@ -164,19 +170,19 @@ func (cb *CachingBucket) addOperationConfig(operationName, configName string, cf
 }
 
 func (cb *CachingBucket) CacheIter(configName string, cache cache.Cache, matcher func(string) bool, cfg CachingBucketConfig) {
-	cb.addOperationConfig("iter", configName, newCacheConfig(cfg, configName, matcher, cache))
+	cb.addOperationConfig(operationIter, configName, newCacheConfig(cfg, configName, matcher, cache))
 }
 
 func (cb *CachingBucket) CacheExists(configName string, cache cache.Cache, matcher func(string) bool, cfg CachingBucketConfig) {
-	cb.addOperationConfig("exists", configName, newCacheConfig(cfg, configName, matcher, cache))
+	cb.addOperationConfig(operationExists, configName, newCacheConfig(cfg, configName, matcher, cache))
 }
 
 func (cb *CachingBucket) CacheGet(configName string, cache cache.Cache, matcher func(string) bool, cfg CachingBucketConfig) {
-	cb.addOperationConfig("get", configName, newCacheConfig(cfg, configName, matcher, cache))
+	cb.addOperationConfig(operationGet, configName, newCacheConfig(cfg, configName, matcher, cache))
 }
 
 func (cb *CachingBucket) CacheGetRange(configName string, cache cache.Cache, matcher func(string) bool, cfg CachingBucketConfig) {
-	cb.addOperationConfig("getrange", configName, newCacheConfig(cfg, configName, matcher, cache))
+	cb.addOperationConfig(operationGetRange, configName, newCacheConfig(cfg, configName, matcher, cache))
 
 	cb.requestedGetRangeBytes.WithLabelValues(configName)
 	cb.fetchedGetRangeBytes.WithLabelValues(originCache, configName)
@@ -185,7 +191,7 @@ func (cb *CachingBucket) CacheGetRange(configName string, cache cache.Cache, mat
 }
 
 func (cb *CachingBucket) CacheObjectSize(configName string, cache cache.Cache, matcher func(name string) bool, cfg CachingBucketConfig) {
-	cb.addOperationConfig("objectsize", configName, newCacheConfig(cfg, configName, matcher, cache))
+	cb.addOperationConfig(operationObjectSize, configName, newCacheConfig(cfg, configName, matcher, cache))
 }
 
 func (cb *CachingBucket) findCacheConfig(configs []*operationConfig, objectName string) (configName string, cfg *operationConfig) {
@@ -218,12 +224,12 @@ func (cb *CachingBucket) ReaderWithExpectedErrs(expectedFunc objstore.IsOpFailur
 }
 
 func (cb *CachingBucket) Iter(ctx context.Context, dir string, f func(string) error) error {
-	cfgName, cfg := cb.findCacheConfig(cb.operationConfigs["iter"], dir)
+	cfgName, cfg := cb.findCacheConfig(cb.operationConfigs[operationIter], dir)
 	if cfg == nil {
 		return cb.Bucket.Iter(ctx, dir, f)
 	}
 
-	cb.operationRequests.WithLabelValues("iter", cfgName).Inc()
+	cb.operationRequests.WithLabelValues(operationIter, cfgName).Inc()
 
 	key := cachingKeyIter(dir)
 
@@ -231,7 +237,7 @@ func (cb *CachingBucket) Iter(ctx context.Context, dir string, f func(string) er
 	if data[key] != nil {
 		list, err := decodeIterResult(data[key])
 		if err == nil {
-			cb.operationHits.WithLabelValues("iter", cfgName).Inc()
+			cb.operationHits.WithLabelValues(operationIter, cfgName).Inc()
 
 			for _, n := range list {
 				err = f(n)
@@ -287,12 +293,12 @@ func decodeIterResult(data []byte) ([]string, error) {
 }
 
 func (cb *CachingBucket) Exists(ctx context.Context, name string) (bool, error) {
-	cfgName, cfg := cb.findCacheConfig(cb.operationConfigs["exists"], name)
+	cfgName, cfg := cb.findCacheConfig(cb.operationConfigs[operationExists], name)
 	if cfg == nil {
 		return cb.Bucket.Exists(ctx, name)
 	}
 
-	cb.operationRequests.WithLabelValues("exists", cfgName).Inc()
+	cb.operationRequests.WithLabelValues(operationExists, cfgName).Inc()
 
 	key := cachingKeyExists(name)
 	hits := cfg.cache.Fetch(ctx, []string{key})
@@ -300,10 +306,10 @@ func (cb *CachingBucket) Exists(ctx context.Context, name string) (bool, error) 
 	if ex := hits[key]; ex != nil {
 		switch string(ex) {
 		case existsTrue:
-			cb.operationHits.WithLabelValues("exists", cfgName).Inc()
+			cb.operationHits.WithLabelValues(operationExists, cfgName).Inc()
 			return true, nil
 		case existsFalse:
-			cb.operationHits.WithLabelValues("exists", cfgName).Inc()
+			cb.operationHits.WithLabelValues(operationExists, cfgName).Inc()
 			return false, nil
 		default:
 			level.Warn(cb.logger).Log("msg", "unexpected cached 'exists' value", "val", string(ex))
@@ -338,25 +344,25 @@ func (cb *CachingBucket) storeExistsCacheEntry(ctx context.Context, cachingKey s
 }
 
 func (cb *CachingBucket) Get(ctx context.Context, name string) (io.ReadCloser, error) {
-	cfgName, cfg := cb.findCacheConfig(cb.operationConfigs["get"], name)
+	cfgName, cfg := cb.findCacheConfig(cb.operationConfigs[operationGet], name)
 	if cfg == nil {
 		return cb.Bucket.Get(ctx, name)
 	}
 
-	cb.operationRequests.WithLabelValues("get", cfgName).Inc()
+	cb.operationRequests.WithLabelValues(operationGet, cfgName).Inc()
 
 	key := cachingKeyContent(name)
 	existsKey := cachingKeyExists(name)
 
 	hits := cfg.cache.Fetch(ctx, []string{key, existsKey})
 	if hits[key] != nil {
-		cb.operationHits.WithLabelValues("get", cfgName).Inc()
+		cb.operationHits.WithLabelValues(operationGet, cfgName).Inc()
 		return ioutil.NopCloser(bytes.NewReader(hits[key])), nil
 	}
 
 	// If we know that file doesn't exist, we can return that. Useful for deletion marks.
 	if ex := hits[existsKey]; ex != nil && string(ex) == existsFalse {
-		cb.operationHits.WithLabelValues("get", cfgName).Inc()
+		cb.operationHits.WithLabelValues(operationGet, cfgName).Inc()
 		return nil, errObjNotFound
 	}
 
@@ -395,7 +401,7 @@ func (cb *CachingBucket) GetRange(ctx context.Context, name string, off, length 
 		return cb.Bucket.GetRange(ctx, name, off, length)
 	}
 
-	cfgName, cfg := cb.findCacheConfig(cb.operationConfigs["getrange"], name)
+	cfgName, cfg := cb.findCacheConfig(cb.operationConfigs[operationGetRange], name)
 	if cfg == nil {
 		return cb.Bucket.GetRange(ctx, name, off, length)
 	}
@@ -411,7 +417,7 @@ func (cb *CachingBucket) GetRange(ctx context.Context, name string, off, length 
 }
 
 func (cb *CachingBucket) ObjectSize(ctx context.Context, name string) (uint64, error) {
-	lname, cfg := cb.findCacheConfig(cb.operationConfigs["objectsize"], name)
+	lname, cfg := cb.findCacheConfig(cb.operationConfigs[operationObjectSize], name)
 	if cfg == nil {
 		return cb.Bucket.ObjectSize(ctx, name)
 	}
@@ -422,11 +428,11 @@ func (cb *CachingBucket) ObjectSize(ctx context.Context, name string) (uint64, e
 func (cb *CachingBucket) cachedObjectSize(ctx context.Context, name string, cfgName string, cache cache.Cache, ttl time.Duration) (uint64, error) {
 	key := cachingKeyObjectSize(name)
 
-	cb.operationRequests.WithLabelValues("objectsize", cfgName).Add(1)
+	cb.operationRequests.WithLabelValues(operationObjectSize, cfgName).Inc()
 
 	hits := cache.Fetch(ctx, []string{key})
 	if s := hits[key]; len(s) == 8 {
-		cb.operationHits.WithLabelValues("objectsize", cfgName).Add(1)
+		cb.operationHits.WithLabelValues(operationObjectSize, cfgName).Inc()
 		return binary.BigEndian.Uint64(s), nil
 	}
 
@@ -443,7 +449,7 @@ func (cb *CachingBucket) cachedObjectSize(ctx context.Context, name string, cfgN
 }
 
 func (cb *CachingBucket) cachedGetRange(ctx context.Context, name string, offset, length int64, cfgName string, cfg *operationConfig) (io.ReadCloser, error) {
-	cb.operationRequests.WithLabelValues("getrange", cfgName)
+	cb.operationRequests.WithLabelValues(operationGetRange, cfgName)
 	cb.requestedGetRangeBytes.WithLabelValues(cfgName).Add(float64(length))
 
 	size, err := cb.cachedObjectSize(ctx, name, cfgName, cfg.cache, cfg.ObjectSizeTTL)
@@ -496,7 +502,7 @@ func (cb *CachingBucket) cachedGetRange(ctx context.Context, name string, offset
 		totalCachedBytes += int64(len(b))
 		cb.fetchedGetRangeBytes.WithLabelValues(originCache, cfgName).Add(float64(len(b)))
 	}
-	cb.operationHits.WithLabelValues("getrange", cfgName).Add(float64(totalCachedBytes) / float64(totalRequestedBytes))
+	cb.operationHits.WithLabelValues(operationGetRange, cfgName).Add(float64(totalCachedBytes) / float64(totalRequestedBytes))
 
 	if len(hits) < len(keys) {
 		if hits == nil {
--- a/pkg/store/cache/caching_bucket.go
+++ b/pkg/store/cache/caching_bucket.go
@@ -11,6 +11,7 @@ import (
 	"io"
 	"io/ioutil"
 	"regexp"
+	"strings"
 	"sync"
 	"time"
 
@@ -22,6 +23,7 @@ import (
 	"github.com/prometheus/client_golang/prometheus/promauto"
 	"golang.org/x/sync/errgroup"
 
+	"github.com/thanos-io/thanos/pkg/block/metadata"
 	"github.com/thanos-io/thanos/pkg/cache"
 	"github.com/thanos-io/thanos/pkg/objstore"
 	"github.com/thanos-io/thanos/pkg/runutil"
@@ -31,6 +33,12 @@ import (
 const (
 	originCache  = "cache"
 	originBucket = "bucket"
+
+	existsTrue  = "true"
+	existsFalse = "false"
+
+	metaFilenameSuffix         = "/" + metadata.MetaFilename
+	deletionMarkFilenameSuffix = "/" + metadata.DeletionMarkFilename
 )
 
 type CachingBucketConfig struct {
@@ -46,6 +54,11 @@ type CachingBucketConfig struct {
 
 	// How long to cache result of Iter call.
 	IterTTL time.Duration `yaml:"iter_ttl"`
+
+	// Meta files (meta.json and deletion mark file) caching config.
+	MetaFilesCachingEnabled bool          `yaml:"metafile_caching_enabled"`
+	MetaFileExistsTTL       time.Duration `yaml:"metafile_exists_ttl"`
+	MetaFileDoesntExistTTL  time.Duration `yaml:"metafile_doesnt_exist_ttl"`
 }
 
 func DefaultCachingBucketConfig() CachingBucketConfig {
@@ -56,6 +69,10 @@ func DefaultCachingBucketConfig() CachingBucketConfig {
 		MaxChunksGetRangeRequests: 3,
 
 		IterTTL: 5 * time.Minute,
+
+		MetaFilesCachingEnabled: true,
+		MetaFileExistsTTL:       10 * time.Minute,
+		MetaFileDoesntExistTTL:  3 * time.Minute,
 	}
 }
 
@@ -78,6 +95,9 @@ type CachingBucket struct {
 
 	iterRequests  prometheus.Counter
 	iterCacheHits prometheus.Counter
+
+	metaFileExistsRequests  prometheus.Counter
+	metaFileExistsCacheHits prometheus.Counter
 }
 
 func NewCachingBucket(b objstore.Bucket, c cache.Cache, chunks CachingBucketConfig, logger log.Logger, reg prometheus.Registerer) (*CachingBucket, error) {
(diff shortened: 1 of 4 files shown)


pkg/store/cache/caching_bucket.go
```diff
@@ -132,35 +221,221 @@ func (cb *CachingBucket) ReaderWithExpectedErrs(expectedFunc objstore.IsOpFailur
...
+	if ex := hits[existsKey]; ex != nil && string(ex) == existsFalse {
+		cb.operationHits.WithLabelValues(operationGet, cfgName).Inc()
+		return nil, errObjNotFound
+	}
+
+	getTime := time.Now()
+	reader, err := cb.Bucket.Get(ctx, name)
+	if err != nil {
+		if cb.Bucket.IsObjNotFoundErr(err) {
+			// Cache that object doesn't exist.
```
ranjithkumar007, line 374: small typo: Cache object that doesn't exist.

pstibrany: I think it's correct the way it is. We want to cache information that object doesn't exist here.

ranjithkumar007: Ah got it.

pkg/store/cache/caching_bucket.go
```diff
@@ -26,92 +29,178 @@ import (
...
+	config = "config"
+
 	originCache  = "cache"
 	originBucket = "bucket"
-)
 
-type CachingBucketConfig struct {
-	// Basic unit used to cache chunks.
-	ChunkSubrangeSize int64 `yaml:"chunk_subrange_size"`
+	existsTrue  = "true"
```
bwplotka, line 37: `strconv.ParseBool(true)`?

pkg/store/cache/caching_bucket.go
```diff
@@ -26,92 +29,178 @@ import (
...
+
+type existsConfig struct {
+	existsTTL, doesntExistTTL time.Duration
+}
 
-	requestedChunkBytes prometheus.Counter
-	fetchedChunkBytes   *prometheus.CounterVec
-	refetchedChunkBytes *prometheus.CounterVec
+type getConfig struct {
+	contentTTL, existsTTL, doesntExistTTL time.Duration
```
bwplotka, line 80: we could embed `existsConfig` for reuse.

pkg/store/cache/caching_bucket.go
```diff
@@ -26,92 +29,178 @@ import (
...
-	objectSizeRequests prometheus.Counter
-	objectSizeHits     prometheus.Counter
+type getRangeConfig struct {
+	subrangeSize               int64
+	objectSizeTTL, subrangeTTL time.Duration
+	maxSubRequests             int
 }
 
-func NewCachingBucket(b objstore.Bucket, c cache.Cache, chunks CachingBucketConfig, logger log.Logger, reg prometheus.Registerer) (*CachingBucket, error) {
+type objectSizeCfg time.Duration
```
bwplotka, line 89: Let's either do everywhere `Cfg` or `Config` (:

pkg/store/cache/caching_bucket.go
```diff
@@ -26,92 +29,178 @@ import (
...
+
+// NewCachingBucket creates caching bucket with no configuration. Various "Cache*" methods configure
+// this bucket to cache results of individual bucket methods. Configuration must be set before
```
bwplotka, line 92: I think it would be type safe if we could pass configuration ONLY via constructor. Any blocker for that? Especially if we cannot change it in runtime.

pkg/store/cache/caching_bucket.go
```diff
@@ -26,92 +29,178 @@ import (
...
-
-	cb.fetchedChunkBytes.WithLabelValues(originBucket)
-	cb.fetchedChunkBytes.WithLabelValues(originCache)
-	cb.refetchedChunkBytes.WithLabelValues(originCache)
+		operationConfigs: map[string][]*operationConfig{},
+
+		requestedGetRangeBytes: promauto.With(reg).NewCounterVec(prometheus.CounterOpts{
+			Name: "thanos_store_bucket_cache_getrange_requested_bytes_total",
+			Help: "Total number of bytes requested via GetRange.",
+		}, []string{config}),
```
bwplotka, line 108: what is config here.. you mean "config"?

pkg/store/cache/caching_bucket.go
```diff
@@ -26,92 +29,178 @@ import (
...
+		operationHits: promauto.With(reg).NewCounterVec(prometheus.CounterOpts{
+			Name: "thanos_store_bucket_cache_operation_hits_total",
+			Help: "Number of operations served from cache for given config.",
+		}, []string{"operation", "config"}),
+	}
 
 	return cb, nil
 }
 
+func (cb *CachingBucket) addOperationConfig(operationName, configName string, matcher func(string) bool, cache cache.Cache, opConfig interface{}) {
```
bwplotka, line 131: It would be nice to actually have it via constructor. We can create some `cacheConfigs` type if need with this method (:

bwplotka: so we can avoid panics

pkg/store/cache/caching_bucket.go
```diff
@@ -26,92 +29,178 @@ import (
...
+	cb.fetchedGetRangeBytes.WithLabelValues(originBucket, configName)
+	cb.refetchedGetRangeBytes.WithLabelValues(originCache, configName)
+}
+
+// CacheObjectSize configures caching of "ObjectSize" operation for matching files.
+func (cb *CachingBucket) CacheObjectSize(configName string, cache cache.Cache, matcher func(name string) bool, ttl time.Duration) {
+	cb.addOperationConfig(operationObjectSize, configName, matcher, cache, objectSizeCfg(ttl))
+}
+
+func (cb *CachingBucket) findCacheConfig(configs []*operationConfig, objectName string) (configName string, cache cache.Cache, cfg interface{}) {
```
bwplotka, line 195: I would consider this method quite shallow. It's not adding much, let's just inline, or better .. replace configs with map?

pkg/store/cache/caching_bucket.go
```diff
@@ -132,35 +221,221 @@ func (cb *CachingBucket) ReaderWithExpectedErrs(expectedFunc objstore.IsOpFailur
...
-var chunksMatcher = regexp.MustCompile(`^.*/chunks/\d+$`)
+func (cb *CachingBucket) Iter(ctx context.Context, dir string, f func(string) error) error {
+	cfgName, cache, cfg := cb.findCacheConfig(cb.operationConfigs[operationIter], dir)
+	if cfg == nil {
+		return cb.Bucket.Iter(ctx, dir, f)
+	}
+	iterCfg := cfg.(iterConfig)
+
+	cb.operationRequests.WithLabelValues(operationIter, cfgName).Inc()
+
```
bwplotka, line 232: nit: Let's dense it a bit? We have new line every line...

pkg/store/cache/caching_bucket.go
```diff
@@ -132,35 +221,221 @@ func (cb *CachingBucket) ReaderWithExpectedErrs(expectedFunc objstore.IsOpFailur
...
+	key := cachingKeyIter(dir)
+
+	data := cache.Fetch(ctx, []string{key})
+	if data[key] != nil {
+		list, err := decodeIterResult(data[key])
```
bwplotka, line 237: Why not tie-ing this method to `iterConfig` itself? Easier to version encodings (:

pkg/store/cache/caching_bucket.go
```diff
@@ -132,35 +221,221 @@ func (cb *CachingBucket) ReaderWithExpectedErrs(expectedFunc objstore.IsOpFailur
...
+		if err == nil {
+			cb.operationHits.WithLabelValues(operationIter, cfgName).Inc()
+
```
bwplotka, line 240:
```suggestion
```

pkg/store/cache/caching_bucket.go
```diff
@@ -132,35 +221,221 @@ func (cb *CachingBucket) ReaderWithExpectedErrs(expectedFunc objstore.IsOpFailur
...
+	cfgName, cache, cfg := cb.findCacheConfig(cb.operationConfigs[operationIter], dir)
+	if cfg == nil {
+		return cb.Bucket.Iter(ctx, dir, f)
+	}
+	iterCfg := cfg.(iterConfig)
+
+	cb.operationRequests.WithLabelValues(operationIter, cfgName).Inc()
+
+	key := cachingKeyIter(dir)
+
```
bwplotka, line 234:
```suggestion
```

pkg/store/cache/caching_bucket.go
```diff
@@ -132,35 +221,221 @@ func (cb *CachingBucket) ReaderWithExpectedErrs(expectedFunc objstore.IsOpFailur
...
 }
 
-var chunksMatcher = regexp.MustCompile(`^.*/chunks/\d+$`)
+func (cb *CachingBucket) Iter(ctx context.Context, dir string, f func(string) error) error {
+	cfgName, cache, cfg := cb.findCacheConfig(cb.operationConfigs[operationIter], dir)
+	if cfg == nil {
+		return cb.Bucket.Iter(ctx, dir, f)
+	}
+	iterCfg := cfg.(iterConfig)
+
```
bwplotka, line 230:
```suggestion
```

pkg/store/cache/caching_bucket.go
```diff
@@ -132,35 +221,221 @@ func (cb *CachingBucket) ReaderWithExpectedErrs(expectedFunc objstore.IsOpFailur
...
+	if data[key] != nil {
+		list, err := decodeIterResult(data[key])
+		if err == nil {
+			cb.operationHits.WithLabelValues(operationIter, cfgName).Inc()
+
+			for _, n := range list {
+				err = f(n)
+				if err != nil {
+					return err
+				}
```
bwplotka, line 245:
```suggestion
				if err :=  f(n); err != nil {
					return err
				}
```

pkg/store/cache/caching_bucket.go
```diff
@@ -132,35 +221,221 @@ func (cb *CachingBucket) ReaderWithExpectedErrs(expectedFunc objstore.IsOpFailur
...
+			}
+			return nil
+		} else {
```
bwplotka, line 248: No need for this else

pkg/store/cache/caching_bucket.go
```diff
@@ -132,35 +221,221 @@ func (cb *CachingBucket) ReaderWithExpectedErrs(expectedFunc objstore.IsOpFailur
...
+			// This should not happen.
```
bwplotka, line 249: mhm why? memcached can malform things, we could create incompatible encodings etc...

pkg/store/cache/caching_bucket.go
```diff
@@ -132,35 +221,221 @@ func (cb *CachingBucket) ReaderWithExpectedErrs(expectedFunc objstore.IsOpFailur
...
+			level.Warn(cb.logger).Log("msg", "failed to decode cached Iter result", "err", err)
+		}
+	}
 
-func isTSDBChunkFile(name string) bool {
-	return chunksMatcher.MatchString(name)
+	// Iteration can take a while (esp. since it calls function), and iterTTL is generally low.
+	// We will compute TTL based time when iteration started.
```
bwplotka, line 255: :+1:

pkg/store/cache/caching_bucket.go
```diff
@@ -132,35 +221,221 @@ func (cb *CachingBucket) ReaderWithExpectedErrs(expectedFunc objstore.IsOpFailur
...
+		list = append(list, s)
+		return f(s)
+	})
+
+	remainingTTL := time.Duration(iterCfg) - time.Since(iterTime)
+	if err == nil && remainingTTL > 0 {
+		data := encodeIterResult(list)
+		if data != nil {
+			cache.Store(ctx, map[string][]byte{key: data}, remainingTTL)
+		}
```
bwplotka, line 268:
```suggestion
		if data := encodeIterResult(list); data != nil {
			cache.Store(ctx, map[string][]byte{key: data}, remainingTTL)
		}
```

pkg/store/cache/caching_bucket.go
```diff
@@ -132,35 +221,221 @@ func (cb *CachingBucket) ReaderWithExpectedErrs(expectedFunc objstore.IsOpFailur
...
+	}
+	return err
+}
+
+// Iter results should compress nicely, especially in subdirectories.
+func encodeIterResult(files []string) []byte {
```
bwplotka, line 274: probably should part of `cacheIter`

pkg/store/cache/caching_bucket_factory.go
```diff
@@ -30,6 +38,41 @@ type CachingBucketWithBackendConfig struct {
 	CachingBucketConfig CachingBucketConfig `yaml:"caching_config"`
 }
 
+type CachingBucketConfig struct {
+	// Basic unit used to cache chunks.
+	ChunkSubrangeSize int64 `yaml:"chunk_subrange_size"`
```
bwplotka, line 43: Ok, so I am super confused why you did all this nice work of making the options typed for each operation and then here, we just accept EVERYTHING (: 

I would either stick to this (if you configure options for specific operation, skip unrelated fields) and then do the same in underlying code, OR let's do typed config here as well.


Since you started nice typed thing on code level, maybe it's worth to actually **reuse** those `getConfig` `iterConfig`  and parse that directly (add yaml tags there)? (: WDYT?

pstibrany: > Ok, so I am super confused why you did all this nice work of making the options typed for each operation and then here, we just accept EVERYTHING (:

My thinking is this: I don’t think user should be configuring parameters of individual operations (get, iter, ...), which is what those small structs do. User-provided configuration is higher level, and it is then used to configure individual operations based on our expert knowledge of what Thanos is doing. (lines 113-122 in this file).

See for example how single MetafileExistsTTL is used in both Get and Exists operation config. I don’t think user should be able to do this.

pkg/store/cache/caching_bucket_factory.go
```diff
@@ -30,6 +38,41 @@ type CachingBucketWithBackendConfig struct {
...
+
+	// Maximum number of GetRange requests issued by this bucket for single GetRange call. Zero or negative value = unlimited.
+	MaxChunksGetRangeRequests int `yaml:"max_chunks_get_range_requests"`
+
+	// TTLs for various cache items.
+	ChunkObjectSizeTTL time.Duration `yaml:"chunk_object_size_ttl"`
+	ChunkSubrangeTTL   time.Duration `yaml:"chunk_subrange_ttl"`
+
+	// How long to cache result of Iter call in root directory.
+	RootIterTTL time.Duration `yaml:"root_iter_ttl"`
```
pracucci, line 53: From the user perspective this is the "blocks iteration". The fact blocks are stored in the root is more an implementation detail (and we may soon support custom prefix support to the bucket). For this reason I would call it `blocks_iter_ttl`.

pkg/store/cache/caching_bucket_factory.go
```diff
@@ -30,6 +38,41 @@ type CachingBucketWithBackendConfig struct {
...
+		ChunkSubrangeSize:         16000, // Equal to max chunk size.
+		ChunkObjectSizeTTL:        24 * time.Hour,
+		ChunkSubrangeTTL:          24 * time.Hour,
+		MaxChunksGetRangeRequests: 3,
+
+		RootIterTTL: 5 * time.Minute,
+
+		MetafileExistsTTL:      10 * time.Minute,
+		MetafileDoesntExistTTL: 3 * time.Minute,
+		MetafileContentTTL:     1 * time.Hour,
```
pracucci, line 72: Metadata file is expected to be immutable, so I would rase the default to 24h like chunks.

pkg/store/cache/caching_bucket_factory.go
```diff
@@ -60,5 +103,35 @@ func NewCachingBucketFromYaml(yamlContent []byte, bucket objstore.Bucket, logger
...
+	}
+
+	cbc := config.CachingBucketConfig
+	// Configure cache.
+	cb.CacheGetRange("chunks", c, isTSDBChunkFile, cbc.ChunkSubrangeSize, cbc.ChunkObjectSizeTTL, cbc.ChunkSubrangeTTL, cbc.MaxChunksGetRangeRequests)
+	cb.CacheExists("metafile", c, isMetaFile, cbc.MetafileExistsTTL, cbc.MetafileDoesntExistTTL)
+	cb.CacheGet("metafile", c, isMetaFile, cbc.MetafileContentTTL, cbc.MetafileExistsTTL, cbc.MetafileDoesntExistTTL)
+
+	// Cache Iter requests for root.
+	cb.CacheIter("dir", c, func(dir string) bool { return dir == "" }, cbc.RootIterTTL)
```
pracucci, line 118: I would define a standalone function for `func(dir string) bool { return dir == "" }` just for readability.

pkg/store/cache/caching_bucket.go
```diff
@@ -190,75 +466,80 @@ func (cb *CachingBucket) getRangeChunkFile(ctx context.Context, name string, off
...
 	}
 
 	// Try to get all subranges from the cache.
-	hits := cb.cache.Fetch(ctx, keys)
+	totalCachedBytes := int64(0)
+	hits := cache.Fetch(ctx, keys)
 	for _, b := range hits {
-		cb.fetchedChunkBytes.WithLabelValues(originCache).Add(float64(len(b)))
+		totalCachedBytes += int64(len(b))
+		cb.fetchedGetRangeBytes.WithLabelValues(originCache, cfgName).Add(float64(len(b)))
```
pracucci, line 506: I would increase this just once, after the `for`, using `totalCachedBytes`

pkg/store/cache/caching_bucket.go
```diff
@@ -190,75 +466,80 @@ func (cb *CachingBucket) getRangeChunkFile(ctx context.Context, name string, off
...
 	}
+	cb.operationHits.WithLabelValues(operationGetRange, cfgName).Add(float64(totalCachedBytes) / float64(totalRequestedBytes))
```
pracucci, line 508: Shouldn't we increase by `len(hits)`?

pkg/store/cache/caching_bucket.go
```diff
@@ -132,35 +221,221 @@ func (cb *CachingBucket) ReaderWithExpectedErrs(expectedFunc objstore.IsOpFailur
...
+func (cb *CachingBucket) Get(ctx context.Context, name string) (io.ReadCloser, error) {
+	cfgName, cache, cfg := cb.findCacheConfig(cb.operationConfigs[operationGet], name)
+	if cfg == nil {
+		return cb.Bucket.Get(ctx, name)
+	}
+	getCfg := cfg.(getConfig)
+
+	cb.operationRequests.WithLabelValues(operationGet, cfgName).Inc()
+
+	key := cachingKeyContent(name)
```
pracucci, line 355: Naming it `contentKey` could simplify reading this code

pkg/store/cache/caching_bucket.go
```diff
@@ -26,92 +29,178 @@ import (
 )
 
 const (
+	config = "config"
```
pracucci, line 32: I think this complicates reading the code because the reader woudln't expect the `config` variable contains the `"config"` label name. I would keep it simpler and just remove this var.

pkg/store/cache/caching_bucket.go
```diff
@@ -26,92 +29,178 @@ import (
...
+	existsFalse = "false"
 
-	// Maximum number of GetRange requests issued by this bucket for single GetRange call. Zero or negative value = unlimited.
-	MaxChunksGetRangeRequests int `yaml:"max_chunks_get_range_requests"`
-
-	// TTLs for various cache items.
-	ChunkObjectSizeTTL time.Duration `yaml:"chunk_object_size_ttl"`
-	ChunkSubrangeTTL   time.Duration `yaml:"chunk_subrange_ttl"`
-}
+	operationGet        = "get"
```
pracucci, line 40: I would compress all these variable names, changing `operation` into `op`

pkg/store/cache/caching_bucket.go
pstibrany: Sounds good, I will make it work that way.

pkg/store/cache/caching_bucket.go
pstibrany: There was `const config="config"` on the top, but it's now inlined.

pkg/store/cache/caching_bucket.go
pstibrany: I've refactored config-setup into separate object (now called `CachingBucketConfig`), but I kept panics :) They are protection for coding errors, not config errors.

pkg/store/cache/caching_bucket.go
pstibrany: It is used from multiple places. In fact, I've now added more methods like this, to avoid using generic map for all kinds of configs.

pkg/store/cache/caching_bucket.go
pstibrany: Good idea, will do that.

pkg/store/cache/caching_bucket.go
pstibrany: I like to keep code breath a little bit :) It makes it easier for me to read it. But yeah, whatever works..

pkg/store/cache/caching_bucket.go
pstibrany: I mean under normal conditions. Removed comment.

pkg/store/cache/caching_bucket_factory.go
pstibrany: What do you mean by "we just accept EVERYTHING"? I explained my reasoning in previous comment, wdyt?

pkg/store/cache/caching_bucket.go
pstibrany: I'd like for `operationHits` metric to be comparable to `operationRequests`, which is why I compute ratio of cached vs requested bytes. WDYT?

pkg/store/cache/caching_bucket_factory.go
pstibrany: This works for now, since `BaseFetcher` uses `Exists` method to verify if meta.json file exists before fetching it. If it switches to `Get` only, we may need to reconsider this default value.

pkg/store/cache/caching_bucket.go
pstibrany: Added `IterCodec` interface with `JsonIterCodec` implementation, and removed snappy compression from it. For Thanos, compression wouldn't help much since there is little repetition in block names in root ("") dir.

pkg/store/cache/caching_bucket.go
pstibrany: Fixed by using `len(hits) / len(keys)`.

pstibrany:
Thank you guys for feedback! I've now addressed it, I'd be happy if you could take a look again.

CHANGELOG entry and documentation is still missing, I will add it asap.

pstibrany:
Please notice how `CachingBucketConfig` is now used for passing operations configuration to the cache, while user-exposed configuration lives in `CachingWithBackendConfig`, which is used to build `CachingBucketConfig` in the `NewCachingBucketFromYaml` function.

pkg/store/cache/caching_bucket_factory.go
```diff
@@ -60,5 +95,40 @@ func NewCachingBucketFromYaml(yamlContent []byte, bucket objstore.Bucket, logger
...
+	cfg.CacheGetRange("chunks", c, isTSDBChunkFile, config.ChunkSubrangeSize, config.ChunkObjectSizeTTL, config.ChunkSubrangeTTL, config.MaxChunksGetRangeRequests)
+	cfg.CacheExists("metafile", c, isMetaFile, config.MetafileExistsTTL, config.MetafileDoesntExistTTL)
+	cfg.CacheGet("metafile", c, isMetaFile, config.MetafileContentTTL, config.MetafileExistsTTL, config.MetafileDoesntExistTTL)
+
+	// Cache Iter requests for root.
+	cfg.CacheIter("dir", c, isBlocksRootDir, config.BlocksIterTTL, JsonIterCodec{})
+
+	// Enabling index caching (example).
+	cfg.CacheObjectSize("index", c, isIndexFile, time.Hour)
+	cfg.CacheGetRange("index", c, isIndexFile, 32000, time.Hour, 24*time.Hour, 3)
```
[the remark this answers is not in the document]
pstibrany: This is an example how to use it for caching index, but I would prefer to address that in a separate PR, while also making decision about further direction on index cache.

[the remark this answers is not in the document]
pracucci: Then I would remove it from this PR.

pkg/store/cache/caching_bucket.go
bwplotka: ah thanks XD it confused me

pkg/store/cache/caching_bucket.go
```diff
@@ -409,3 +575,16 @@ func (c *subrangesReader) subrangeAt(offset int64) ([]byte, error) {
 	}
 	return b, nil
 }
+
+// JsonIterCodec encodes iter results into JSON. Suitable for root dir.
+type JsonIterCodec struct{}
```
bwplotka, line 580: either JSON or `json` I think is the rule here.

pkg/store/cache/caching_bucket.go
```diff
@@ -132,35 +131,184 @@ func (cb *CachingBucket) ReaderWithExpectedErrs(expectedFunc objstore.IsOpFailur
...
+	// We will compute TTL based time when iteration started.
+	iterTime := time.Now()
+	var list []string
+	err := cb.Bucket.Iter(ctx, dir, func(s string) error {
+		list = append(list, s)
+		return f(s)
+	})
+
+	remainingTTL := cfg.ttl - time.Since(iterTime)
+	if err == nil && remainingTTL > 0 {
```
bwplotka, line 168: `remainingTTL > 0` 

I might think this as error case if users defines ttl smaller than actual iteration time (:

bwplotka: I would log warning at least + metric.

Thanks for caring for this BTW! and spotting this special case :muscle:

pkg/store/cache/caching_bucket.go
```diff
@@ -132,35 +131,184 @@ func (cb *CachingBucket) ReaderWithExpectedErrs(expectedFunc objstore.IsOpFailur
...
+		if data, err := cfg.codec.Encode(list); err == nil {
+			cfg.cache.Store(ctx, map[string][]byte{key: data}, remainingTTL)
+		} else {
```
bwplotka, line 171: As you seen I love explicitlness and don't like `else` that's why I would suggest returning `nil` explicitly instead of else's (:

bwplotka: Of course feel free to disagree

pkg/store/cache/caching_bucket.go
```diff
@@ -132,35 +131,184 @@ func (cb *CachingBucket) ReaderWithExpectedErrs(expectedFunc objstore.IsOpFailur
...
+	key := cachingKeyExists(name)
+	hits := cfg.cache.Fetch(ctx, []string{key})
+
+	if ex := hits[key]; ex != nil {
+		exists, err := strconv.ParseBool(string(ex))
+		if err == nil {
+			cb.operationHits.WithLabelValues(opExists, cfgName).Inc()
+			return exists, nil
+		}
+		level.Warn(cb.logger).Log("msg", "unexpected cached 'exists' value", "val", string(ex))
```
bwplotka, line 195: worth to mention `key`?

pkg/store/cache/caching_bucket.go
```diff
@@ -132,35 +131,184 @@ func (cb *CachingBucket) ReaderWithExpectedErrs(expectedFunc objstore.IsOpFailur
...
+		list = append(list, s)
+		return f(s)
+	})
+
+	remainingTTL := cfg.ttl - time.Since(iterTime)
+	if err == nil && remainingTTL > 0 {
+		if data, err := cfg.codec.Encode(list); err == nil {
+			cfg.cache.Store(ctx, map[string][]byte{key: data}, remainingTTL)
+		} else {
+			level.Warn(cb.logger).Log("msg", "failed to encode Iter result", "err", err)
```
bwplotka, line 172: worth to mention `key`?

pkg/store/cache/caching_bucket.go
```diff
@@ -132,35 +131,184 @@ func (cb *CachingBucket) ReaderWithExpectedErrs(expectedFunc objstore.IsOpFailur
...
+		if err == nil {
+			cb.operationHits.WithLabelValues(opIter, cfgName).Inc()
+			for _, n := range list {
+				if err := f(n); err != nil {
+					return err
+				}
+			}
+			return nil
+		}
+		level.Warn(cb.logger).Log("msg", "failed to decode cached Iter result", "err", err)
```
bwplotka, line 155: worth to mention `key`?

pkg/store/cache/caching_bucket.go
```diff
@@ -132,35 +131,184 @@ func (cb *CachingBucket) ReaderWithExpectedErrs(expectedFunc objstore.IsOpFailur
...
+	} else {
+		ttl = doesntExistTTL - time.Since(ts)
+	}
+
+	if ttl > 0 {
+		cache.Store(ctx, map[string][]byte{cachingKey: []byte(strconv.FormatBool(exists))}, ttl)
+	}
+}
+
+func (cb *CachingBucket) Get(ctx context.Context, name string) (io.ReadCloser, error) {
```
bwplotka, line 220: those files are HUGE!

bwplotka: we do this  a lot on compactor

pkg/store/cache/caching_bucket.go
```diff
@@ -132,35 +131,184 @@ func (cb *CachingBucket) ReaderWithExpectedErrs(expectedFunc objstore.IsOpFailur
...
+	hits := cfg.cache.Fetch(ctx, []string{contentKey, existsKey})
+	if hits[contentKey] != nil {
+		cb.operationHits.WithLabelValues(opGet, cfgName).Inc()
+		return ioutil.NopCloser(bytes.NewReader(hits[contentKey])), nil
+	}
+
+	// If we know that file doesn't exist, we can return that. Useful for deletion marks.
+	if ex := hits[existsKey]; ex != nil {
+		if exists, err := strconv.ParseBool(string(ex)); err == nil && !exists {
+			cb.operationHits.WithLabelValues(opGet, cfgName).Inc()
```
bwplotka, line 240: can we store those information in exists cache actually?

bwplotka: and share?

bwplotka: actually we do lol, but the metrics I would expect to be incremented for exists then,.... same below

pkg/store/cache/caching_bucket.go
```diff
@@ -132,35 +131,184 @@ func (cb *CachingBucket) ReaderWithExpectedErrs(expectedFunc objstore.IsOpFailur
...
+	defer runutil.CloseWithLogOnErr(cb.logger, reader, "CachingBucket.Get(%q)", name)
+
+	data, err := ioutil.ReadAll(reader)
+	if err != nil {
+		return nil, err
+	}
+
+	ttl := cfg.contentTTL - time.Since(getTime)
+	if ttl > 0 {
+		cfg.cache.Store(ctx, map[string][]byte{contentKey: data}, ttl)
```
bwplotka, line 264: ditto with too small ttl

Also I am not sure if it makes sense to be size unbounded.....

bwplotka: :thinking:

pkg/store/cache/caching_bucket.go
```diff
@@ -132,35 +131,184 @@ func (cb *CachingBucket) ReaderWithExpectedErrs(expectedFunc objstore.IsOpFailur
...
+		if cb.Bucket.IsObjNotFoundErr(err) {
+			// Cache that object doesn't exist.
+			storeExistsCacheEntry(ctx, existsKey, false, getTime, cfg.cache, cfg.existsTTL, cfg.doesntExistTTL)
+		}
+
+		return nil, err
+	}
+	defer runutil.CloseWithLogOnErr(cb.logger, reader, "CachingBucket.Get(%q)", name)
+
+	data, err := ioutil.ReadAll(reader)
```
bwplotka, line 257: Ok, I think this is wrong. The interface was made to allow you to stream download. We just break it here. Ideally, we want a special readers handling this. This will be complicated and add some work, so I would say, let's leave with this but add proper TODOs and warning comments for this structure. Given it's used for meta jsons it's ok, but it's easy to miss this hidden knowledge (since it's generic wrapper) and use it for huge files.

bwplotka: Actually wait a second. You just did all the hard work with cachedGetSubrange!

Let's just use this then? (:

pkg/store/cache/caching_bucket_factory.go
bwplotka: I see. I missed the fact that this is hardcoded per funcionalities. Cool then.

pkg/store/cache/caching_bucket_factory.go
```diff
@@ -60,5 +95,36 @@ func NewCachingBucketFromYaml(yamlContent []byte, bucket objstore.Bucket, logger
 		return nil, errors.Errorf("unsupported cache type: %s", config.Type)
 	}
 
-	return NewCachingBucket(bucket, c, config.CachingBucketConfig, logger, reg)
+	cfg := NewCachingBucketConfig()
+
+	// Configure cache.
+	cfg.CacheGetRange("chunks", c, isTSDBChunkFile, config.ChunkSubrangeSize, config.ChunkObjectSizeTTL, config.ChunkSubrangeTTL, config.MaxChunksGetRangeRequests)
+	cfg.CacheExists("metafile", c, isMetaFile, config.MetafileExistsTTL, config.MetafileDoesntExistTTL)
```
bwplotka, line 102:
```suggestion
	cfg.CacheExists("meta.jsons", c, isMetaFile, config.MetafileExistsTTL, config.MetafileDoesntExistTTL)
```

pkg/store/cache/caching_bucket_factory.go
```diff
@@ -60,5 +95,36 @@ func NewCachingBucketFromYaml(yamlContent []byte, bucket objstore.Bucket, logger
...
+	cfg.CacheGet("metafile", c, isMetaFile, config.MetafileContentTTL, config.MetafileExistsTTL, config.MetafileDoesntExistTTL)
```
bwplotka, line 103: "meta.jsons"?

pkg/store/cache/caching_bucket_factory.go
```diff
@@ -60,5 +95,36 @@ func NewCachingBucketFromYaml(yamlContent []byte, bucket objstore.Bucket, logger
...
+
+	return cb, nil
+}
+
+var chunksMatcher = regexp.MustCompile(`^.*/chunks/\d+$`)
+
+func isTSDBChunkFile(name string) bool { return chunksMatcher.MatchString(name) }
+
+func isMetaFile(name string) bool {
+	return strings.HasSuffix(name, metaFilenameSuffix) || strings.HasSuffix(name, deletionMarkFilenameSuffix)
```
bwplotka, line 121: I feel like `metaFilenameSuffix` and `deletionMarkFilenameSuffix` are used only once... worth to maybe inline those here? :thinking:

pkg/store/cache/caching_bucket.go
```diff
@@ -28,86 +31,82 @@ import (
...
 
-	objectSizeRequests prometheus.Counter
-	objectSizeHits     prometheus.Counter
+	operationConfigs  map[string][]*operationConfig
+	operationRequests *prometheus.CounterVec
+	operationHits     *prometheus.CounterVec
 }
 
-func NewCachingBucket(b objstore.Bucket, c cache.Cache, chunks CachingBucketConfig, logger log.Logger, reg prometheus.Registerer) (*CachingBucket, error) {
+// NewCachingBucket creates new caching bucket with provided configuration. Configuration should not be
```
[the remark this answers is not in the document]
pstibrany: I explicitly left it there, as I was too lazy to implement copy of the config in NewCachingBucket for all those maps. Comment still applies. WDYT?

pkg/store/cache/caching_bucket.go
pstibrany: I use the same `ttl > 0` pattern in all methods. If we add logging + metric here, then I think it should go to all of them.

pkg/store/cache/caching_bucket.go
pstibrany:
```		if data, err := cfg.codec.Encode(list); err == nil {
			cfg.cache.Store(ctx, map[string][]byte{key: data}, remainingTTL)
			return nil
		}
		level.Warn(cb.logger).Log("msg", "failed to encode Iter result", "err", err)
```

version would have a subtle bug -- using of incorrect `err` variable. But I can make it bit more explicit.

pkg/store/cache/caching_bucket.go
pstibrany: > those files are HUGE!

Not sure what you're saying here. If we want to cache file, we need to fetch it into memory first. `Get` is only cached for `meta.json` and deletion mark files. For all other files, `Get` goes directly to bucket, with no caching.

pkg/store/cache/caching_bucket.go
pstibrany: > but the metrics I would expect to be incremented for exists then,.... same below

I was thinking about this, but I think it would be confusing to report it for `exists` operation... this is really `get` operation, but reusing cached information about whether file exist. I want 'requests' and 'hits' metrics to be comparable... if we increase 'hits' for exists, we also need to increase 'requests' ... but then 'requests' will not match real bucket usage, which is the confusing bit. Personally I think it's ok the way it is.

pkg/store/cache/caching_bucket_factory.go
pstibrany: I suppose Go compiler is smart enough not to do concatenation on each call.

pkg/store/cache/caching_bucket.go
pstibrany: Reader used in cachedGetSubrange is different – it fetches all blocks into memory first.

I've now modified `Get` to use streaming, and buffer read object in the memory up to configured size. If object is read fully, we cache it (under single key, unlike `GetRange`, which caches each subrange under separate key).

pkg/store/cache/caching_bucket.go
pstibrany: Not anymore. PTAL.

pkg/store/cache/caching_bucket.go
pstibrany: `Get` is now streaming.

pkg/store/cache/caching_bucket.go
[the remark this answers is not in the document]
bwplotka: Cool makes sense. It's totally fine :+1:

pkg/store/cache/caching_bucket.go
bwplotka: Agree, but would be nice to log+metric, but we can do as follow up

pkg/store/cache/caching_bucket.go
bwplotka: :+1: Up to you, sorry for being a pain (:

pkg/store/cache/caching_bucket.go
bwplotka: > . For all other files, Get goes directly to bucket, with no caching.

Well, yes, but it's easy to miss. Think about users of this package.

bwplotka: Thanks for making it streaming

pkg/store/cache/caching_bucket.go
bwplotka: Then I might think we should adjust request for exists as well (: 

But happy to leave it like this and adjust later if needed.

pkg/store/cache/caching_bucket_factory.go
bwplotka: Yes, I know, I am just seeing this as reader complexity. Human reading this code has to jump 100 lines to ensure something for no reason (variable not reused)

bwplotka:
:tada:

Issue #2249, opened 2020-03-10
cmd: Ability to run multiple commands in single process.


bwplotka:
It would be nice to be able to run many components as a single process. In a similar model, you can run Cortex. This is not a critical blocker but for demo purposes being able to run Querier + Compactor + Store in the same container for introduction demo would be a nice touch. (: 

Cortex is using this library: https://github.com/cortexproject/cortex/tree/master/pkg/util/services  It's great, but I wonder if such complexity is needed. For us, just different `run.Group` would do. The main problem is with discovery between those components in the same binary and configuration since we base on flags. Probably to solve this we would need to move all to config file...

The issue for some initial discussion on this.

bwplotka:
Demo would be the only use case TBH, so quesiton is... is it worth pursuing.

jojohappy:
Update.

darshanime:
If the only usecase is easy demos, why not add a docker compose setup for thanos? 
It'll be as simple as [`docker-compose up`](https://docs.docker.com/compose/reference/up/) to get the entire stack up and running (using minio as bucket store would make it independent of external dependencies)

bwplotka:
Looks like docker compose just landed on thanos-community org: https://github.com/prmsrswt/thanos-docker-compose

I think that's good enough for now..

I am not fan of having single comment to "rule them all", as so far we have feedback that what Loki and Cortex have causes major confusion here & there. Worth to rething. I think docker compose can help here greatly (:

bwplotka:
I strongly believe that claiming that something is easier only because you can demo it easier with single process (but does not make sense for any realistic setups) is not a valid statement. 

Let's close this for now then :hugs:

bwplotka:
There is actually one more use case: Just small home-size project e.g smart home telemetry etc (:

bwplotka:
TODO. Nice as some good first issue potentially (:

Kind Regards,
Bartek Płotka (@bwplotka)

SuperQ:
It's not just demos where this is useful. Small installs for users that are outside of Kubernetes need a lot of orchestration could use it.

PR #2583, opened 2020-05-11
Add docker compose setup for local development, demo

darshanime:

Closes https://github.com/thanos-io/thanos/issues/2249

Add a docker-compose setup for local dev environment, demos etc. 

* [ ] I added documentation for this new flow

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -12,6 +12,7 @@ kube/.minikube
 
 # Ignore e2e working dirs.
 data/
+!development/data
 test/e2e/e2e_integration_test*
 
 # Ignore promu artifacts.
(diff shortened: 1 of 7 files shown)


darshanime:
@bwplotka the PR is not yet done, but wanted some early feedback on the idea

brancz:
I agree this seems to be best done in a separate project.

bwplotka:
Happy to set up the project for https://github.com/thanos-community/ if you want.

darshanime:
sure, that sounds good @bwplotka 👍

bwplotka:
I mean, this also mean we need longer term maintainer for  such project. Also my worry is that such project will be super small.

Maybe something like deployments would do better :thinking:

prmsrswt:
I think having a sample `docker-compose.yml` would be of great help. Maybe we can add the example to the documentation.

darshanime:
As disussed in community call, this is not required currently. 
Can I close this?

PR #2596, opened 2020-05-13
Update Prometheus to cd73b3d33e064bbd846fc7a26dc8c313d46af382

codesome:
As we are blocked on some breaking changes to be addressed before we can upgrade Prometheus to the latest master, this PR attempts to upgrade Prometheus upto this commit https://github.com/prometheus/prometheus/commit/cd73b3d33e064bbd846fc7a26dc8c313d46af382 which hopefully does not have any breaking changes and helps the TSDB block based ingesters in Cortex :)


codesome:
I would have expected `go.sum` file to have changed, but that is not the case. What is more worrying is that, `go.mod` and `go.sum` in upstream point to different commit. (upstream `go.mod` says `b90be6f32a33` and `go.sum` says `1e64d757f711`, and the CI picks up `1e64d757f711`.

codesome:
Marco pointed me out to the override in go.mod, makes sense now, removing it and trying

pracucci:
There's an overrided in `go.mod` `replace` which you should change as well.

cmd/thanos/rule.go
```diff
@@ -73,10 +73,10 @@ func registerRule(m map[string]setupFunc, app *kingpin.Application) {
 		Default("1m"))
 	evalInterval := modelDuration(cmd.Flag("eval-interval", "The default evaluation interval to use.").
 		Default("30s"))
-	tsdbBlockDuration := modelDuration(cmd.Flag("tsdb.block-duration", "Block duration for TSDB block.").
```
bwplotka, line 76: Just curious what happened with those? The above `eval` still have modelDuration

cmd/thanos/tools.go
```diff
@@ -58,17 +63,66 @@ func checkRulesFiles(logger log.Logger, files *[]string) error {
...
 
 type ThanosRuleGroups struct {
 	Groups []ThanosRuleGroup `yaml:"groups"`
 }
 
-func checkRules(logger log.Logger, filename string) (int, errors.MultiError) {
+// Validate validates all rules in the rule groups.
+// This is copied from Prometheus.
+// TODO: Replace this with upstream implementation after https://github.com/prometheus/prometheus/issues/7128 is fixed.
+func (g *ThanosRuleGroups) Validate() (errs []error) {
```
bwplotka, line 77: :+1:

cmd/thanos/tools.go
```diff
@@ -99,31 +151,31 @@ func checkRules(logger log.Logger, filename string) (int, errors.MultiError) {
...
-			)
-		}
-		promRuleGroups.Groups = append(
-			promRuleGroups.Groups,
-			group,
-		)
-	}
-	return promRuleGroups
-}
+//func thanosRuleGroupsToPromRuleGroups(ruleGroups ThanosRuleGroups) rulefmt.RuleGroups {
```
bwplotka, line 154: What's this?

go.mod
```diff
@@ -69,8 +71,6 @@ require (
 replace (
 	// Mitigation for: https://github.com/Azure/go-autorest/issues/414
 	github.com/Azure/go-autorest => github.com/Azure/go-autorest v12.3.0+incompatible
-	// Make sure Cortex is not forcing us to some other Prometheus version.
-	github.com/prometheus/prometheus => github.com/prometheus/prometheus v1.8.2-0.20200110114423-1e64d757f711 // master ~ v2.15.2
```
bwplotka, line 73: Should we have this for current version?

pkg/compact/downsample/downsample.go
```diff
@@ -695,6 +695,12 @@ func (it *AverageChunkIterator) Next() bool {
 	return true
 }
 
+func (it *AverageChunkIterator) Seek(t int64) bool {
+	for it.t < t && it.Next() {
```
bwplotka, line 699: that's cheating =D

pkg/receive/multitsdb.go
```diff
@@ -302,3 +304,106 @@ func (t *MultiTSDB) TenantAppendable(tenantID string) (Appendable, error) {
 	}
 	return tenant.readyStorage(), nil
 }
+
+// ErrNotReady is returned if the underlying storage is not ready yet.
+var ErrNotReady = errors.New("TSDB not ready")
+
+// ReadyStorage implements the Storage interface while allowing to set the actual
+// storage at a later point in time.
+type ReadyStorage struct {
```
bwplotka, line 313: It's because it is private now? Can you add `TODO` to upstream making it back public on Prometheus or something?

pkg/rule/api/v1.go
```diff
@@ -92,7 +92,7 @@ func (api *API) rules(r *http.Request) (interface{}, []error, *qapi.ApiError) {
 				enrichedRule = alertingRule{
 					Name:                    rule.Name(),
 					Query:                   rule.Query().String(),
-					Duration:                rule.Duration().Seconds(),
+					Duration:                rule.HoldDuration().Seconds(),
```
bwplotka, line 95: Hold?

pkg/rule/rule.go
```diff
@@ -42,8 +43,18 @@ type RuleGroups struct {
 	Groups []RuleGroup `yaml:"groups"`
 }
 
+// ruleGroup is a list of sequentially evaluated recording and alerting rules.
```
bwplotka, line 46: comment name obsolete

pkg/rule/rule.go
```diff
@@ -132,10 +143,10 @@ func (r RuleGroup) MarshalYAML() (interface{}, error) {
 	}
 
 	rs := struct {
-		RuleGroup               rulefmt.RuleGroup `yaml:",inline"`
-		PartialResponseStrategy *string           `yaml:"partial_response_strategy,omitempty"`
+		RuleGroup               PromRuleGroup `yaml:",inline"`
```
bwplotka, line 146: BTW this will kill us @s-urbaniak XD

bwplotka:
> As we are blocked on some breaking changes to be addressed before we can upgrade Prometheus to the latest master, this PR attempts to upgrade Prometheus upto this commit prometheus/prometheus@cd73b3d which hopefully does not have any breaking changes and helps the TSDB block based ingesters in Cortex :)

Why not up to latest? 

I guess then @krasi-georgiev want to take over?

cmd/thanos/tools.go
codesome: Because I copied `Validate()` method temporarily, this method was unused. I kept it in comment because once we switch back to using upstream `Validate()`, we will need it again. Do you want me to remove it altogether?

go.mod
codesome: Makes sense, I didn't think about that. I will add that.

pkg/receive/multitsdb.go
codesome: It is removed upstream :( so had to copy it from the old version

pkg/rule/api/v1.go
codesome: The method was changed upstream. There is no `Duration()` now. I will verify this again.

codesome:
> Why not up to latest?

That has more breaking changes which will take time to fix here, and idea was to get some improvements from Prometheus soon. And yes, Krasi will take over after this one.

codesome:
Also, I have some ugly hacks in place right now to fix some breaking changes. Once I fix the e2e tests, I plan to review stuff again and do a cleanup.
One such breaking change was replacing `Rule` with `RuleNode` in a `RuleGroup` upstream. The `RuleNode` upstream uses a yaml Node instead of a string like in `Rule`, which is painful when using it here, so I added a `PromRuleGroup` which does not use `RuleNode`.

codesome:
> Do you think there was something significant we pulled with this.

I will take a look at the changelogs and add the significate changes :)

cmd/thanos/rule.go
codesome: Thanks for pointing out! `.Int64()` is definitely wrong as the is not an integer, I need to use `Duration()`. And to answer why not `modelDuration`, we need these as int64 at the end, so having it as `model.Duration` will lead to lot of type conversions.

pkg/compact/downsample/downsample.go
codesome: But works :) would this cause any performance issues?

cmd/thanos/tools.go
bwplotka: Well git is versioning system you know... we don't need another backuping with comments (:

pkg/receive/multitsdb.go
bwplotka: Noooo it's not removed. It's part of main.go which is something we can pull back to reuse in Prometheus (|:

pkg/rule/api/v1.go
bwplotka: Thanks!

bwplotka:
This makes totally sense @codesome , let's do step by step. Thanks for helping :heart:

pkg/rule/api/v1.go
codesome: Yes, the name was changed, that's all.

cmd/thanos/tools.go
codesome: Makes sense XD

pkg/compact/downsample/downsample.go
bwplotka: We don't use it I think, so I would error out (`it.err = errors.New("seek used, but not implemented")`)

pkg/receive/multitsdb.go
codesome: Oh, I did not know that, I will add a TODO

bwplotka:
Looks like essentially we have something wrong with flags still. e2e test:
```
=== CONT  TestReceive/hashring
Starting receive-1
receive-1: Error parsing commandline arguments: time: unknown unit d in duration 15d
receive-1: usage: thanos receive [<flags>]
receive-1: Accept Prometheus remote write API requests and write to local tsdb`
```

codesome:
I guess its the `modelDuration` method after all. I will fall back to using them :)

codesome:
Tests are passing and changelog entries added!

CHANGELOG.md
```diff
@@ -31,6 +31,9 @@ We use *breaking* word for marking changes that are not backward compatible (rel
 - [2513](https://github.com/thanos-io/thanos/pull/2513) Tools: Moved `thanos bucket` commands to `thanos tools bucket`, also
 moved `thanos check rules` to `thanos tools rules-check`. `thanos tools rules-check` also takes rules by `--rules` repeated flag not argument
 anymore.
+- [#2596](https://github.com/thanos-io/thanos/pull/2596) Update to Prometheus [@cd73b3d33e064bbd846fc7a26dc8c313d46af382](https://github.com/prometheus/prometheus/commit/cd73b3d33e064bbd846fc7a26dc8c313d46af382) which falls in between v2.17.0 and v2.18.0.
+    - TSDB now supports ssolation of append and queries.
```
kakkoyun, line 35:
```suggestion
    - TSDB now supports isolation of append and queries.
```
I believe this is a typo.

bwplotka:
@codesome looks like some conflicts this morning. Let's fix & merge, and then I will do 0.13 release (: :muscle: 

Thanks!

cmd/thanos/tools.go
```diff
@@ -58,17 +63,66 @@ func checkRulesFiles(logger log.Logger, files *[]string) error {
...
+
+	for _, g := range g.Groups {
+		if g.Name == "" {
+			errs = append(errs, errors.New("Groupname should not be empty"))
+		}
+
+		if _, ok := set[g.Name]; ok {
+			errs = append(
+				errs,
+				fmt.Errorf("groupname: \"%s\" is repeated in the same file", g.Name),
```
squat, line 88: nit, rather than `\"%s\"`, we can use `%q` to quote a string :)

bwplotka:
LGTM, Let's merge on green (:

bwplotka:
Thanks!

@codesome @pracucci @pstibrany it's unblock Cortex I guess.

PR #2597, opened 2020-05-13
pkg/ui: bump jQuery to 3.5.0

squat:

cc @prmsrswt @GiedriusS


prmsrswt:
Looks good, but the tests are failing because `yarn.lock` was not updated to reflect the changes in `package.json`. i think you just need to run `yarn` in `pkg/ui/react-app` and commit the updated `yarn.lock` file.

PR #2604, opened 2020-05-14
Cleanup flags for compact component

PhilipGough:
Introduce config structs for sidecar component.

## Changes

As per #2248  - this change parses flags into structs before passing into the compact run func


bwplotka:
Can you rebase? The fix will ensure we don't broke the help (:

PhilipGough:
@bwplotka done

kakkoyun:
@PhilipGough Nice 👏

PR #2611, opened 2020-05-14
Fixed make docs 

bwplotka:

cc @Harshitha1234 @PhilipGough  this is what I mean (:

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,6 +1,6 @@
 FROM quay.io/prometheus/busybox:latest
 LABEL maintainer="The Thanos Authors"
 
-COPY thanos /bin/thanos
+COPY /thanos_tmp_for_docker /bin/thanos
 
 ENTRYPOINT [ "/bin/thanos" ]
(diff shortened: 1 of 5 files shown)


Dockerfile
```diff
@@ -1,6 +1,6 @@
 FROM quay.io/prometheus/busybox:latest
 LABEL maintainer="The Thanos Authors"
 
-COPY thanos /bin/thanos
+COPY /thanos_tmp_for_docker /bin/thanos
```
squat, line 4: What issue does using this temporary thanos binary solve?

bwplotka: The issue: 
* We accidently were using wrong binary in genflagdocs vs used by docker. This is because `make build` was building to current dir `./thanos` vs install-deps step is installing to GOBIN/thanos via `go install`

This is why I would love to be explicit (: and always use GOBIN. However either my dockerfile sucks or docker just cannot get anything outside current dir and recursive sub dirs.

squat: ack got it :+1: your Docker skillz are not bad, it is an (intentional) limitation of the Docker build tooling

bwplotka:
This is however still broken and not fixing much ;p

bwplotka:
Ok, found issue, it was wrong embedmd hook

PR #2612, opened 2020-05-15
mixin: Alert on receive not uploading recent data

brancz:

## Changes

Add an alert on receive not performing uploads of data.

## Verification

Wrote this rule based on data of a recent incident. So I checked whether it would have fired in that situation.

@bwplotka @krasi-georgiev @squat @metalmatze @kakkoyun

### DIFF
--- a/mixin/thanos/alerts/receive.libsonnet
+++ b/mixin/thanos/alerts/receive.libsonnet
@@ -93,6 +93,17 @@
               severity: 'warning',
             },
           },
+          {
+            alert: 'ThanosReceiveNoUpload',
+            annotations: {
+              message: 'Thanos Receive {{$labels.job}} has not uploaded latest data to object storage.',
+            },
+            expr: 'increase(thanos_shipper_uploads_total{%(selector)s}[2h]) == 0' % thanos.receive,
+            'for': '30m',
+            labels: {
+              severity: 'warning',
+            },
+          },
         ],
       },
     ],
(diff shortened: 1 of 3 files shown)


mixin/thanos/alerts/receive.libsonnet
```diff
@@ -93,6 +93,17 @@
               severity: 'warning',
             },
           },
+          {
+            alert: 'ThanosReceiveNoUpload',
```
bwplotka, line 97: worth to mention this is only ok for receive with upload option on.... :thinking:

brancz: are you suggesting a comment?

brancz: or conditionally add the rule if uploads are enabled?

kakkoyun: Not want to make things complicated but if we only initialize the shipper for the receive when upload is enabled and check the existence of the metric for the alert, we could achieve this. 
However I think we should defer this until this creates an issue. We can just add a comment.
WDYT?

brancz: hah you're right, this should already not fire because we only create the shipper and register metrics if enabled

brancz:
I'll move this along, but happy to do a follow up if there are any other requested changes! :)

PR #2618, opened 2020-05-15
Documentation and CHANGELOG.md entry for #2579.

pstibrany:
This PR adds CHANGELOG entry and documentation for PR #2579.

### DIFF
--- a/docs/components/store.md
+++ b/docs/components/store.md
@@ -274,7 +274,7 @@ While the remaining settings are **optional**:
 
 ## Caching Bucket
 
-Thanos Store Gateway supports a "caching bucket" with chunks caching to speed up loading of chunks from TSDB blocks. Currently only memcached "backend" is supported:
+Thanos Store Gateway supports a "caching bucket" with chunks and metadata caching to speed up loading of chunks from TSDB blocks. Currently only memcached "backend" is supported:
 
 ```yaml
 backend: memcached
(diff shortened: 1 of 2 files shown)


pstibrany:
> Perfect! Just DCO & LGTM!

🤦

pstibrany:
> @pstibrany One thing I have realized, we mentioned and documented config file to use it, however the flag itself is not documented and it is hidden (if it hasn't changed). This might create confusion. Should we also make it visible or at least mention it somewhere?

That's a good point. I will put it into documentation and changelog entry.

docs/components/store.md
```diff
@@ -274,30 +274,44 @@ While the remaining settings are **optional**:
...
+metafile_exists_ttl: 2h
+metafile_doesnt_exist_ttl: 15m
+metafile_content_ttl: 24h
+metafile_max_size: 1MiB
 ```
 
 `backend_config` field for memcached supports all the same configuration as memcached for [index cache](#memcached-index-cache).
 
-`caching_config` is a configuration for chunks cache and supports the following optional settings:
+`caching_config` is a configuration for cache and supports the following optional settings for chunks caching:
```
GiedriusS, line 300: But `caching_config` key doesn't exist, right? :thinking: So, I think this line should be reworded.

pstibrany: You're correct. I've removed mention of `caching_config`.

docs/components/store.md
```diff
@@ -274,31 +274,45 @@ While the remaining settings are **optional**:
...
 - `max_chunks_get_range_requests`: how many "get range" sub-requests may cache perform to fetch missing subranges.
 - `chunk_object_size_ttl`: how long to keep information about chunk file length in the cache.
 - `chunk_subrange_ttl`: how long to keep individual subranges in the cache.
 
-Note that chunks cache is an experimental feature, and these fields may be renamed or removed completely in the future.
+Following options are used for metadata caching (meta.json files, deletion mark files, iteration result):
+
+- `blocks_iter_ttl`: how long to cache result of iterating blocks.
+- `metafile_exists_ttl`: how long to cache information about whether meta.json or deletion mark file exists
+- `metafile_doesnt_exist_ttl`: how long to cache information about whether meta.json or deletion mark file doesn't exist
```
pracucci, line 311:
```suggestion
- `metafile_doesnt_exist_ttl`: how long to cache information about whether meta.json or deletion mark file doesn't exist.
```

PR #2621, opened 2020-05-18
receive: Only wait for write quorum

brancz:

## Changes

This patch modifies receive replication slightly, in that it doesn't
always wait for all requests to complete anymore. If quorum amount of
replication requests were successful it now does not wait for the
remaining request to finish as it's not necessary to reach quorum
anymore. In error cases where quorum is not reached, it still continues
to wait for all requests to finish in an attempt to return a quorum
error.

Additionally this patch moves log lines printed in the parallelize
requests function to debug logging. Calling functions already print the
resulting error(s), so this was previously just noise, even in cases
where requests actually succeeded.

Let me know if you think there should be a changelog entry for this, in reality there is not really a user noticable change, other than less noisy logs and lower latency for requests.

## Verification

* Tested various failure cases manually with the quickstart script to ensure that errors are still printed appropriately.
* Duplicated the existing replication tests to 1) just wait for quorum 2) run tests plus consistency delay as on a test run there should never be any problem, verifying overall correctness.

@bwplotka @krasi-georgiev @metalmatze @squat @kakkoyun

### DIFF
--- a/pkg/receive/handler.go
+++ b/pkg/receive/handler.go
@@ -13,6 +13,7 @@ import (
 	"net/http"
 	"strconv"
 	"sync"
+	"time"
 
 	"github.com/go-kit/kit/log"
 	"github.com/go-kit/kit/log/level"
@@ -67,6 +68,7 @@ type Options struct {
 	Tracer            opentracing.Tracer
 	TLSConfig         *tls.Config
 	DialOpts          []grpc.DialOption
+	ForwardTimeout    time.Duration
 }
 
 // Handler serves a Prometheus remote write receiving HTTP endpoint.
@@ -324,32 +326,38 @@ func (h *Handler) forward(ctx context.Context, tenant string, r replica, wreq *p
 	}
 	h.mtx.RUnlock()
 
-	return h.parallelizeRequests(ctx, tenant, replicas, wreqs)
+	return h.fanoutForward(ctx, tenant, replicas, wreqs, len(wreqs))
 }
 
-// parallelizeRequests parallelizes a given set of write requests.
-// The function only returns when all requests have finished
-// or the context is canceled.
-func (h *Handler) parallelizeRequests(ctx context.Context, tenant string, replicas map[string]replica, wreqs map[string]*prompb.WriteRequest) error {
+// writeQuorum returns minimum number of replicas that has to confirm write success before claiming replication success.
+func (h *Handler) writeQuorum() int {
+	return int((h.options.ReplicationFactor / 2) + 1)
+}
+
+// fanoutForward fanouts concurrently given set of write requests. It returns status immediately when quorum of
+// requests succeeds or fails or if context is cancelled.
+func (h *Handler) fanoutForward(ctx context.Context, tenant string, replicas map[string]replica, wreqs map[string]*prompb.WriteRequest, successThreshold int) error {
 	ec := make(chan error)
-	defer close(ec)
-	// We don't wan't to use a sync.WaitGroup here because that
-	// introduces an unnecessary second synchronization mechanism,
-	// the first being the error chan. Plus, it saves us a goroutine
-	// as in order to collect errors while doing wg.Wait, we would
-	// need a separate error collection goroutine.
-	var n int
+
+	var wg sync.WaitGroup
 	for endpoint := range wreqs {
-		n++
+		wg.Add(1)
+
 		// If the request is not yet replicated, let's replicate it.
 		// If the replication factor isn't greater than 1, let's
 		// just forward the requests.
 		if !replicas[endpoint].replicated && h.options.ReplicationFactor > 1 {
 			go func(endpoint string) {
-				ec <- h.replicate(ctx, tenant, wreqs[endpoint])
+				defer wg.Done()
+				if err := h.replicate(ctx, tenant, wreqs[endpoint]); err != nil {
+					ec <- errors.Wrap(err, "replicate write request")
+					return
+				}
+				ec <- nil
 			}(endpoint)
 			continue
 		}
+
 		// If the endpoint for the write request is the
 		// local node, then don't make a request but store locally.
 		// By handing replication to the local node in the same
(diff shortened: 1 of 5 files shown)


pkg/receive/handler.go
```diff
@@ -375,15 +393,15 @@ func (h *Handler) parallelizeRequests(ctx context.Context, tenant string, replic
 						err = errors.New(errs.Error())
 					}
 				}
-				if err != nil {
-					level.Error(h.logger).Log("msg", "storing locally", "err", err, "endpoint", endpoint)
-				}
-				ec <- err
+				ec <- errors.Wrapf(err, "storagin locally, endpoint %v", endpoint)
```
kakkoyun, line 396:
```suggestion
				ec <- errors.Wrapf(err, "storing locally, endpoint %v", endpoint)
```

brancz:
@kakkoyun good point! Added! :)

pkg/receive/handler.go
```diff
@@ -324,29 +326,39 @@ func (h *Handler) forward(ctx context.Context, tenant string, r replica, wreq *p
 	}
 	h.mtx.RUnlock()
 
-	return h.parallelizeRequests(ctx, tenant, replicas, wreqs)
+	n, ec := h.parallelizeRequests(ctx, tenant, replicas, wreqs)
```
bwplotka, line 329: why not just `for err := range <-ec`? without n?

pkg/receive/handler.go
```diff
@@ -464,20 +465,57 @@ func (h *Handler) replicate(ctx context.Context, tenant string, wreq *prompb.Wri
...
-		if uint64(len(errs)) >= (h.options.ReplicationFactor+1)/2 {
-			return errors.Wrap(err, "did not meet replication threshold")
+	ctx, cancel := context.WithTimeout(ctx, h.options.ForwardTimeout)
+	success := uint64(0)
+
+	n, ec := h.parallelizeRequests(ctx, tenant, replicas, wreqs)
+	defer func() {
+		go func() {
+			for {
+				err, more := <-ec
```
bwplotka, line 475:
```
defer cancel()
for err := range <-ec {
```
Would do the work I think.

pkg/receive/handler.go
```diff
@@ -375,15 +389,15 @@ func (h *Handler) parallelizeRequests(ctx context.Context, tenant string, replic
 						err = errors.New(errs.Error())
 					}
 				}
-				if err != nil {
-					level.Error(h.logger).Log("msg", "storing locally", "err", err, "endpoint", endpoint)
-				}
-				ec <- err
+				ec <- errors.Wrapf(err, "storing locally, endpoint %v", endpoint)
```
bwplotka, line 392: I think we should just return if no error?

pkg/receive/handler.go
```diff
@@ -324,29 +326,39 @@ func (h *Handler) forward(ctx context.Context, tenant string, r replica, wreq *p
...
 		n++
 		// If the request is not yet replicated, let's replicate it.
 		// If the replication factor isn't greater than 1, let's
 		// just forward the requests.
 		if !replicas[endpoint].replicated && h.options.ReplicationFactor > 1 {
+			wg.Add(1)
 			go func(endpoint string) {
-				ec <- h.replicate(ctx, tenant, wreqs[endpoint])
+				defer wg.Done()
+				ec <- errors.Wrap(h.replicate(ctx, tenant, wreqs[endpoint]), "could not replicate write request")
```
bwplotka, line 361: Should pass error only on error?

pkg/receive/handler.go
bwplotka: I think I think you rely on `err != nil ` on caller side, maybe I would do it here for readability, but not a blocker.

pkg/receive/handler.go
bwplotka: Ok because you relied on err != nil on caller side.... and you forgot about it here, I would really recommend my suggestion in comment above =D

pkg/receive/handler.go
```diff
@@ -464,20 +465,57 @@ func (h *Handler) replicate(ctx context.Context, tenant string, wreq *prompb.Wri
...
+				}
+			}
+			errs.Add(err)
+			n--
+			if n > 0 {
+				// Not done yet need to keep going.
+				continue
+			}
+
+			if uint64(countCause(errs, isNotReady)) >= (h.options.ReplicationFactor+1)/2 {
```
bwplotka, line 506: Maybe worth to keep this important number in some function.. `(h.options.ReplicationFactor+1)/2 `

pkg/receive/handler.go
```diff
@@ -464,20 +465,57 @@ func (h *Handler) replicate(ctx context.Context, tenant string, wreq *prompb.Wri
...
+		case <-ctx.Done():
+			return ctx.Err()
+		case err := <-ec:
+			if err == nil {
+				success++
+				if success >= (h.options.ReplicationFactor/2)+1 {
+					return nil
+				}
+			}
+			errs.Add(err)
```
bwplotka, line 499: I am so confused, why we are passing err nil to multiError?

bwplotka: (it can flow through above)

pkg/receive/handler.go
bwplotka: And yes, I understand this is how we know if it was success or not.. but maybe we can come up with cleaner API

pkg/receive/handler.go
brancz: `errors.Wrap` returns nil if the input err is nil

pkg/receive/handler.go
brancz: errors.Wrap returns nil if the input err is nil

pkg/receive/handler.go
brancz: nil errors are not actually added

pkg/receive/handler.go
brancz: The problem is that we want to drain the channel, and stop when it's actually empty and closed, so we need to have the `more` returned.

Any suggestions for a better API? The problem is that we need to know in advance how many potential results we would be getting from the channel.

pkg/receive/handler.go
brancz: we need to know the potential number of results so we know when we have reached quorum, it could also return the quorum amount instead, but that would really result in the same code

brancz:
> Also you use a lot this errors.Wrap(nil, .... and tsdbError.MultiError.Add(nil) and it just looks too scary to me. It might be just me, but I feel like it should be antipatterns, mentioned in code style ): It's just scares reader a lot (: For me it's just opportunity for easier errors.

I don't see anything in the style guide that's violated here.

The only other API that I could think of that wouldn't end up in just a rearrangement of the current code is using two channels, one for errors one for successes, but that would complicate draining them a lot.

pkg/receive/handler.go
bwplotka: But `for ... <-ec` has exactly the same semantics, no?

bwplotka: It will stop iterating when it's closed. If empty `err, more := <-ec` will not unblock either, it will block

pkg/receive/handler.go
bwplotka: I am totally aware of that but IMO it's extremely confusing and prone to error. Plus adds major overhead on critical path.

pkg/receive/handler.go
bwplotka: Again: I am totally aware of that but IMO it's extremely confusing and prone to error.

bwplotka:
> I don't see anything in the style guide that's violated here.

Yea, I am proposing to add that.

Plus only this PR is doing so, 100% of Thanos codebase is not putting nil to multierror and does not wrap nils with message.

bwplotka:
Let me think about API, I totally see the aim of it, maybe we can find something cleaner

pkg/receive/handler.go
```diff
@@ -358,82 +366,116 @@ func (h *Handler) parallelizeRequests(ctx context.Context, tenant string, replic
...
+		success int
+		errs    terrors.MultiError
+	)
+	for {
+		select {
+		case <-ctx.Done():
+			return ctx.Err()
+		case err, more := <-ec:
+			if !more {
+				return errs
```
bwplotka, line 462: I was thinking about this case, but I though that this will never happen (: We either have success or errors quorum kind of in my previous version (:

pkg/receive/handler.go
```diff
@@ -464,20 +506,22 @@ func (h *Handler) replicate(ctx context.Context, tenant string, wreq *prompb.Wri
...
+	ctx, cancel := context.WithTimeout(ctx, h.options.ForwardTimeout)
+	defer cancel()
+
+	quorum := h.writeQuorum()
+	err := h.fanoutForward(ctx, tenant, replicas, wreqs, quorum)
+	if countCause(err, isNotReady) >= quorum {
+		return tsdb.ErrNotReady
 	}
-	return errors.Wrap(err, "could not replicate write request")
+	if countCause(err, isConflict) >= quorum {
```
bwplotka, line 517: I still think it should len - quorum

brancz: Let's take the example of replication factor 3, which has a quorum factor of 2, and we get 1 conflict error. `3-2=1`, so we would be returning conflict, even though write quorum was met.

pkg/receive/handler.go
brancz: The attempt here is to return the best possible error, for example in a 3x replication, and 2x return tsdb-not-ready/unavailable, and 1 conflict, would end up with a generalized error when in reality a retry is likely to resolve the error.

pkg/receive/handler.go
bwplotka: Ok, then it as to be `> len(reqs) - quorum`, right? (not `>=`)
 
I think we are both right if we assume that quorum is always +1 than half. This is however very depending on quorum value.... This algorithm should never assume things like that. Let's say quorum is 1 for some reason, with replication 3, then this logic will not hold true, vs `> len(reqs) - quorum` is always correct. (:

brancz: > This is however very depending on quorum value....

bwplotka: We talked offline.  Not a blocker so merging.

bwplotka:
Thanks!

PR #2624, opened 2020-05-18
mixin: Dynamically alert on forward request depending on replication factor

kakkoyun:

## Verification

* `make examples`
* Manually with ad-hoc queries against a running cluster.

cc @squat @brancz @metalmatze

### DIFF
--- a/mixin/alerts/receive.libsonnet
+++ b/mixin/alerts/receive.libsonnet
@@ -3,7 +3,6 @@
   receive+:: {
     selector: error 'must provide selector for Thanos Receive alerts',
     httpErrorThreshold: 5,
-    forwardErrorThreshold: 5,
     refreshErrorThreshold: 0,
     p99LatencyThreshold: 10,
   },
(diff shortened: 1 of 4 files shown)


mixin/alerts/receive.libsonnet
```diff
@@ -56,10 +55,14 @@
                 sum by (job) (rate(thanos_receive_forward_requests_total{result="error", %(selector)s}[5m]))
               /
                 sum by (job) (rate(thanos_receive_forward_requests_total{%(selector)s}[5m]))
-              * 100 > %(forwardErrorThreshold)s
+              )
+              >
+              (
+                max by (job) ((thanos_receive_replication_factor+1) / 2)
```
brancz, line 61: If I'm understanding this right, then I think you need to use `floor` on this, as we are not doing integer divisions here.

kakkoyun:
@brancz Do you see any blockers?

brancz:
lgtm

PR #2575, opened 2020-05-07
Rename index cache YAML fields

ranjithkumar007:
Fixes #2570 
Signed-off-by: Ranjith Kumar <ranjith.dakshana2015@gmail.com>

* [x] I added CHANGELOG entry for this change.
* [ ] Change is not relevant to the end user.

## Changes
1. Rename index cache YAML fields from `type` to `backend`, `config` to `backend_config`
2. Update docs to reflect the change in YAML fields

## Verification
for Change 1, Not required may be?
for Change 2. I checked using `make web-serve`.

cc @pstibrany

PR #2625, opened 2020-05-18
Rename caching_bucket yaml fields

ranjithkumar007:
Relevant discussion #2575 

Signed-off-by: Ranjith Kumar <ranjith.dakshana2015@gmail.com>

## Changes
1. Rename caching bucket YAML fields from backend to type, backend_config to config 
2. Update docs to reflect the change in YAML fields

## Verification
Verified documentation using `make web-serve` .

Caching at bucket level was not part of an earlier release. So, this shouldn't update changelog, right?


bwplotka:
Thanks!@

PR #2630, opened 2020-05-19
Proposed rule to avoid wrapping nil error.

bwplotka:
As per discussion on https://github.com/thanos-io/thanos/pull/2621/files#r427146885

### DIFF
--- a/docs/contributing/coding-style-guide.md
+++ b/docs/contributing/coding-style-guide.md
@@ -288,6 +288,46 @@ in future. There was even Go 2 proposal for [disabling this in the language itse
 
 Similar to this problem is the package name shadowing. While it is less dangerous, it can cause similar issues, so avoid package shadowing if you can.
 
+#### Avoid Wrapping nil With Error Message; Avoid Passing nil to MultiError
+
+It's very tempting to use "magic" functionality of e.g errors.Wrap that returns nil if wrapped error equals nil as it created more compacted code.
+Avoid that. Consider following case:
+
+<table>
+<tbody>
+<tr><th>Avoid 🔥</th></tr>
+<tr><td>
+
+```go
+    return errors.Wrapf(doSomethingOrReturnError(var1), "do something for %v", var1)
+```
+
+</td></tr>
+<tr><th>Better 🤓</th></tr>
+<tr><td>
+
+```go
+    if err := doSomethingOrReturnError(var1); err != nil {
+    	return errors.Wrapf(err, "do something for %v", var1)
+    }
+    return nil
+```
+
+</td></tr>
+</tbody></table>
+
+This can cause pain in long term maintenance because of several reasons:
+
+* If project or even consumer of library will swap to different errors package, the implementation might NOT handle nil in a way
+we expect it, because wrapping nil does not make sense.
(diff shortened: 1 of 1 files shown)


brancz:
I'm very torn on these. The first time I saw the `errors.Wrap` function used I was also kind of surprised, but once you know it, it's actually a nice tool to write precise and short code that clearly describes the error path. I mean .. these APIs handle these cases in that way intentionally.

I won't block this but I personally prefer to use these APIs to their full extend.

docs/contributing/coding-style-guide.md
```diff
@@ -288,6 +288,65 @@ in future. There was even Go 2 proposal for [disabling this in the language itse
...
+It's very tempting to use "magic" functionality of e.g errors.Wrap that returns nil if wrapped error equals nil as it created more compacted code.
+Avoid that. Consider following case:
+
+<table>
+<tbody>
+<tr><th>Avoid 🔥</th></tr>
+<tr><td>
+
+```go
+  	if !replicas[endpoint].replicated && h.options.ReplicationFactor > 1 {
```
squat, line 302: I think this example is way too complicated to illustrate the point of returning a wrapped error. it distracts from the point

bwplotka: agree

bwplotka:
yup @brancz I only want to start discussion. Let's not merge until we all are happy.

docs/contributing/coding-style-guide.md
metalmatze: Yes. I have a hard time figuring out what's happening too. Still haven't found it.

A nice idea for such examples would also be to use diffs.

Essentially like this, to give an example. Obviously we lose Go syntax highlighting. Not sure what's better.
```diff
func example() error {
-	return foo()
+	if err := foo(); err != nil {
+		return err
+	}
+	return nil
```

bwplotka: Simplified

bwplotka: diff would be nice, but for consistency let's keep it simple for now

bwplotka:
BTW @brancz we were thinking about writing our own error pkg. In this project I was thinking to literally return error on nil. That's why the whole discussion.

This will make sure we don't panic: https://github.com/thanos-io/thanos/pull/2630

bwplotka:
Another argument I should put maybe to the list that nil comparison is hard. Sometimes you have variable not equal to `nil` but to `Type(nil)`.. etc

pracucci:
I don't have a strong opinion on this.

I personally find easier to read an explicit `if err != nil` (like this proposal outlines) but I also agree with @brancz that once you get used to the fact that wrappers handle `nil` error, then it's pretty convenient to use.

bwplotka:
Just for you to be aware I broke my own rule in Prometheus and could not stop writing:

```
if err != nil {
			merr := tsdb_errors.MultiError{err}
			merr.Add(primary.Close())
			for _, q := range secondaries {
				merr.Add(q.Close())
			}
			return nil, merr.Err()
		}
```

:smiley: I think I will definitely lose the rule for MultiTSDB I like it. But I would still keep this rule for `Wraps` (:

brancz:
I see it the other way around :upside_down_face:

PR #2631, opened 2020-05-19
Moved CI to dockerhub.

bwplotka:

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -4,7 +4,7 @@ version: 2
 defaults: &defaults
   docker:
     # Built by Thanos make docker-ci
-    - image: quay.io/thanos/thanos-ci:go1.14.2-node
+    - image: thanosio/thanos-ci:v1.1-go1.14.2-node
 jobs:
   test:
     <<: *defaults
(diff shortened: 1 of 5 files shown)


bwplotka:
We will need to live with broken cross-build step: https://github.com/prometheus/promu/issues/182

metalmatze:
Rather than changing everything over to another registry, I would say we should give some more time and the problem will resolve itself on its own (at least from our point of view).
We don't migrate off of GitHub just because it has issue (I know it's not really a good metaphor, but I hope it gets my point across).

bwplotka:
It's was 5minute job, so why we cannot do it?

bwplotka:
I am pretty sure it will come back soon, just as a project we don't care much

metalmatze:
If you feel like it. Go ahead. Don't want to block. Just stated that it feels a bit rushed to me (although that might be my personal opinion). 😉

bwplotka:
Looks like we are back, but let's keep this PR handy. It's not a such huge pain to more registries as it does not matter for this project. The only problem would be release, that's it.

PR #2632, opened 2020-05-20
Fixed panic in compactor

pracucci:

## Changes

If the compactor group key check fails, it currently wraps an error which is actually `nil` and this causes a panic. This PR fixes it.

## Verification

Manual tests.


bwplotka:
I don't think these panics... Wrap handles nils just fine. 

We actually had related discussion here: https://github.com/thanos-io/thanos/pull/2630

pracucci:
> I don't think these panics... Wrap handles nils just fine.

But `halt()`  doesn't.

bwplotka:
`halt that nil` (:

PR #2633, opened 2020-05-20
TSDB supports flushing natively so no need of the flushable storage adapter.

krasi-georgiev:
Now using the new head compaction method instead of the read only flush. The head compaction locks the db compaction to ensure that flashing the head and compaction don't run at the same time. this simplifies the receiver code quite a bit.

Changelog entry not required as this changes just an internal implementation and it doesn't add or remove anything a user should know about.


brancz:
cc @squat

krasi-georgiev:
the prom/tsdb upgrade has been done and merged in a separate PR so this one become about removing the flushable storage adapted as we can do this natively in tsdb now.

pkg/receive/tsdb_test.go
```diff
@@ -16,74 +17,96 @@ import (
...
-func TestFlushableStorage(t *testing.T) {
+// TestHeadFlushIntoBlock ensures that the head flush
+// creates a block that is ready for shipping and
+// that flushing the storage does not cause data loss.
+// This test:
+// * opens a storage;
+// * appends values;
+// * flushes the storage wal/head; and
+// * queries the storage to ensure the samples are present from the flushed block.
+func TestHeadFlushIntoBlock(t *testing.T) {
```
squat, line 33: I don't think these unit tests really make sense any more. They used to ensure that our flushable storage implementation actually worked, but now we are essentially just replicating unit tests that should exist within prometheus. This is really testing foreign code; no Thanos packages/types are exercised here, right? Let's remove this IMO, otherwise we could do this is simply creating more maintenance overhead.

krasi-georgiev: done

cmd/thanos/receive.go
```diff
@@ -429,7 +429,7 @@ func runReceive(
 					WriteableStoreServer: webHandler,
 				}
 
-				s = grpcserver.NewReadWrite(logger, &receive.UnRegisterer{Registerer: reg}, tracer, comp, grpcProbe, rw,
+				s = grpcserver.NewReadWrite(logger, reg, tracer, comp, grpcProbe, rw,
```
squat, line 432: have you tested this change? if you trigger a hashring change and a new grpc server is created and registers its metrics against the registerer, then this will panic

pkg/receive/multitsdb.go
```diff
@@ -259,15 +261,15 @@ func (t *MultiTSDB) getOrLoadTenant(tenantID string, blockingStart bool) (*tenan
 			)
 		}
 
-		s := NewFlushableStorage(
+		s, err := tsdb.Open(
 			dataDir,
 			logger,
 			reg,
```
squat, line 267: same here, the FlushableStorage implementation ensured that a each opened tsdb could re-register metrics every time that a database reload was triggered. Doing two `tsdb.Open`s should cause two calls withing the `tsdb` package to `MustRegister`, which will cause a panic, no?

cmd/thanos/receive.go
krasi-georgiev: yes correct I overlooked this.

PR #2634, opened 2020-05-20
store: Fix inconsistent cache backend types

kakkoyun:

## Changes

* Change memcached BucketCacheProvider constant type to `MEMCACHED` to be consistent with index cache backend.

https://github.com/thanos-io/thanos/blob/8bdd15e032c24fb98ed9ec69fa42f16376a9101e/pkg/store/cache/factory.go#L18-L23

* Make it case-insensitive to prevent future issues.

## Verification

* `make test-local`

cc @pstibrany

### DIFF
--- a/docs/components/store.md
+++ b/docs/components/store.md
@@ -285,7 +285,7 @@ Thanos Store Gateway supports a "caching bucket" with chunks and metadata cachin
 Currently only memcached "backend" is supported:
 
 ```yaml
-type: memcached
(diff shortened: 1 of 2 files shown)


bwplotka:
One more thing. If we merge here this will not be part of 0.13.0. I guess this is ok?

pracucci:
> One more thing. If we merge here this will not be part of 0.13.0. I guess this is ok?

Given bucket caching is marked as experimental, I would say it's OK.

PR #2646, opened 2020-05-22
Refactoring to extract blocks grouping logic out of syncer

pracucci:

## Changes

In Cortex we're experimenting a different grouping logic, in order to be able to concurrently compact blocks belonging to non-overlapping time ranges. So far we've got a good result and we would like to move it upstream.

In order to continue to use the great Thanos `BucketCompactor`, we would need to inject a custom blocks grouping logic.

In this PR I'm proposing to extract the syncer grouping logic into a dedicated component (`Grouper`), moving the syncer `Groups()` logic to a `DefaultGrouper` implementation. Logic and metrics haven't changed (if they did, it's a bug).

/cc @pstibrany 

## Verification

Existing tests.

### DIFF
--- a/pkg/compact/compact.go
+++ b/pkg/compact/compact.go
@@ -49,8 +49,6 @@ type Syncer struct {
 	partial                  map[ulid.ULID]error
 	blockSyncConcurrency     int
 	metrics                  *syncerMetrics
-	acceptMalformedIndex     bool
-	enableVerticalCompaction bool
 	duplicateBlocksFilter    *block.DeduplicateFilter
 	ignoreDeletionMarkFilter *block.IgnoreDeletionMarkFilter
 }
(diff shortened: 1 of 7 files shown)


pkg/compact/compact.go
```diff
@@ -645,11 +703,6 @@ func (cg *Group) compact(ctx context.Context, dir string, comp tsdb.Compactor) (
 			return false, ulid.ULID{}, errors.Wrapf(err, "read meta from %s", pdir)
 		}
 
-		cgKey, groupKey := cg.Key(), GroupKey(meta.Thanos)
```
pracucci, line 648: I've removed this circuit breaker because, from my perspective, it shouldn't be the single group responsabilty to know how blocks should be grouped together (that's Grouper responsability).

pkg/compact/compact.go
```diff
@@ -286,11 +198,127 @@ func (s *Syncer) GarbageCollect(ctx context.Context) error {
 	return nil
 }
 
+// Grouper is responsible to group all known blocks into sub groups which are safe to be
+// compacted concurrently.
+type Grouper interface {
+	// Groups returns the compaction groups for all blocks currently known to the syncer.
+	// It creates all groups from the scratch on every call.
+	Groups(blocks map[ulid.ULID]*metadata.Meta) (res []*Group, err error)
```
pstibrany, line 206: Can we add comment about result slice of groups -- is order important?

pkg/compact/compact.go
```diff
@@ -286,11 +198,127 @@ func (s *Syncer) GarbageCollect(ctx context.Context) error {
...
+
 // Group captures a set of blocks that have the same origin labels and downsampling resolution.
 // Those blocks generally contain the same series and can thus efficiently be compacted.
 type Group struct {
 	logger                      log.Logger
 	bkt                         objstore.Bucket
+	key                         string
 	labels                      labels.Labels
 	resolution                  int64
 	mtx                         sync.Mutex
```
pstibrany, line 324: It's not clear to me why Group needs a mutex in the first place. It is fully constructed in one goroutine, before dispatched to goroutines that do compaction. (Dispatching creates happens-before relation, and group is not modified afterwards).

It's not part of this PR, but I think this mutex can be dropped entirely. (Separate PR is fine as well)

pracucci:
Thanks @pstibrany for your feedback! I will wait any maintainer feedback as well and then I will address all comments at once.

pkg/compact/compact.go
bwplotka: Makes sense to me (:

bwplotka:
Looks you did not rebase so you run older jobs. Since it's passing, let's merge.

PR #2648, opened 2020-05-22
Allow using multiple memcached clients at the same time.

pstibrany:
This PR fixes issue when user tries to configure both caching bucket and index cache at the same time. That creates two independent memcached clients, which try to create same metrics into the same registry, which failed before. This PR fixes that.

## Verification

Unit test that fails on master. Also manually tested by configuring Thanos store to use both caches -- it fails with master build, but works with this patch.

/cc @kakkoyun (since you reported it to me)


bwplotka:
I am cherry picking it

PR #2652, opened 2020-05-24
Docs: fix Pull Request Process in CONTRIBUTING.md

### LINKED ISSUE #2651: Docs: Pull Request Process described in CONTRIBUTING.md doesn't work as expected
The Pull Request Process as described in CONTRIBUTING.md gives us following steps:
```
$ mkdir -p $GOPATH/src/github.com/thanos-io
$ cd $GOPATH/src/github.com/thanos-io
$ git clone https://github.com/<your_github_id>/thanos.git
$ cd thanos
$ git remote add upstream https://github.com/thanos-io/thanos.git
$ git remote update
$ git merge upstream/master
$ make build
$ ./thanos -h
```

There are some minor problems here.

```
$ mkdir -p $GOPATH/src/github.com/thanos-io
```
In most of the cases, the environment variable GOPATH isn't set, so this command tries to create a folder at `/src/github.com/thanos-io` which is not what we want.
A possible solution is to run `GOPATH=$(go env GOPATH)` first.

Also, as we use Go modules we actually don't need to clone the repo inside the GOPATH, it works fine irrespective of the folder's location.

```
$ make build
$ ./thanos -h
```
Here, `./thanos -h` fails because after #2611 the `make build` outputs the `thanos` binary in `$GOBIN` instead of the current folder. Maybe we can run this using `$GOPATH/bin/thanos -h` but I guess the proper way would be to have the `$GOPATH/bin` directory in the `PATH` variable.

prmsrswt:

## Changes
Fixes #2651

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -99,6 +99,7 @@ component of the Thanos distributed system. We recommend:
 1. Fork thanos-io/thanos.git and start development from your own fork. Here are sample steps to setup your development environment:
 
 ```console
+$ GOPATH=$(go env GOPATH)
 $ mkdir -p $GOPATH/src/github.com/thanos-io
 $ cd $GOPATH/src/github.com/thanos-io
 $ git clone https://github.com/<your_github_id>/thanos.git
@@ -107,7 +108,8 @@ $ git remote add upstream https://github.com/thanos-io/thanos.git
 $ git remote update
 $ git merge upstream/master
 $ make build
-$ ./thanos -h
+$ export PATH=$PATH:$GOPATH/bin
+$ thanos -h
 ```
 
 1. Signing your work: DCO (Developer Certificate of Origin) Process.


CONTRIBUTING.md
```diff
@@ -107,7 +108,8 @@ $ git remote add upstream https://github.com/thanos-io/thanos.git
 $ git remote update
 $ git merge upstream/master
 $ make build
-$ ./thanos -h
+$ export PATH=$PATH:$GOPATH/bin
+$ thanos -h
```
soniasingla, line 112: @prmsrswt `thanos -h` could be replaced with `thanos -h flags/command/args` 
Also, additionally we can link it to `scripts/quickstart.sh` to get the basic idea of which components to run :) 

WDYT @daixiang0 @bwplotka ?

bwplotka: > thanos -h could be replaced with thanos -h flags/command/args

What is ` thanos -h flags/command/args`? :thinking: 

> Also, additionally we can link it to scripts/quickstart.sh to get the basic idea of which components to run :)

Yes!

soniasingla: > > thanos -h could be replaced with thanos -h flags/command/args
> 
> What is ` thanos -h flags/command/args`? 🤔
 
We need to mention the arguments/flags/commands whichever we are using after `thanos -h` 
for ex : `sidecar` , `store` are inculded under commands :)

bwplotka: hmmm you can but what's wrong with printing all of them? (: `thanos -h` just works and essentially it's the way to test if you build Thanos properly for a start.

daixiang0: I think the sub command test would be done with `make docs`, as @bwplotka said, it is enough for a start.

PR #2662, opened 2020-05-26
Document series labels sorting in SeriesResponse

pracucci:

## Changes

After a brief [discussion on public Slack](https://cloud-native.slack.com/archives/CL25937SP/p1590491672334400) and after a look at the `Series()` implementation, I think we can safely document that response series labels are guaranteed to be sorted by name.

## Verification

N/A


bwplotka:
Thanks, just ignore lint for now https://github.com/thanos-io/thanos/issues/2663

bwplotka:
Did not know (: SGTM then. Thanks!

PR #2666, opened 2020-05-27
Added Series request hints support to BucketStore

### LINKED ISSUE #2660: Add request hints to BucketStore
In Cortex, given a query, we're running a query plan in the querier to figure out which blocks should be queried and then we fetch the `Series()` from all store-gateway instances holding the affected blocks (a store-gateway runs the Thanos `BucketStore`).

Due to how the blocks sharding and replication works, there're real cases where we end up querying the same block from multiple store-gateway instances. For example if SG1 contains blocks A,B,C and SG2 contains blocks B,C,D and we have to query blocks C,D the querier fetches `Series()` both from SG1 and SG2. The block D will be queried only on SG2, but block C will be queried both from SG1 and SG2. This leads to a waste of computational resources.

Given we run a plan in the querier, we would like to be able to tell `BucketStore` which blocks to query, in order to efficiently distribute the workload across nodes and not fetch series from the same block twice (or even 3 times, given we run with a blocks replication factor = 3) for the same query.

Like we did in https://github.com/thanos-io/thanos/issues/2479, I would like to propose to introduce **request hints**:

```
message SeriesRequest {
  // [already existing fields]

  /// hints is an opaque data structure that can be used to carry additional information.
  /// The content of this field and whether it's supported depends on the
  /// implementation of a specific store. 
  google.protobuf.Any hints = 9;
}
```

Like we did for the response, `hints` would be an opaque structure whose content depends on the specific store. The `BucketStore` would support these hints:

```
message SeriesRequestHints {
    /// filter_blocks is the list of blocks that should be queried. Any other block loaded in the
    /// store can be skipped. If the list is empty, no filtering is applied.
    repeated Block filter_blocks = 1 [(gogoproto.nullable) = false];
}
```

The required changes in `BucketStore.Series()` would be minimal (few lines of code) and shouldn't make the `BucketStore.Series()` implementation harder to maintain over the time.

pracucci:

## Changes

As detailed in #2660, in this PR I'm proposing to add hints support to `SeriesRequest`, similarly to what we did in the `SeriesResponse`. The request hints proposed in the PR allows to let the client to optionally specify the list of block IDs that should be queried. This would allow Cortex to build a more optimised query planner and executor.

Fixes #2660.

## Verification

Unit tests.


pkg/store/bucket.go
```diff
@@ -874,14 +874,21 @@ func (s *BucketStore) Series(req *storepb.SeriesRequest, srv storepb.Store_Serie
...
+		g, gctx  = errgroup.WithContext(ctx)
+		reqHints = &hintspb.SeriesRequestHints{}
+		resHints = &hintspb.SeriesResponseHints{}
 	)
 
+	if s.enableSeriesHints && req.Hints != nil {
+		if err := types.UnmarshalAny(req.Hints, reqHints); err != nil {
+			return status.Error(codes.InvalidArgument, errors.Wrap(err, "unmarshal series request hints").Error())
+		}
+	}
```
pstibrany, line 890: Shouldn't unknown hints be simply ignored?

pracucci: I don't think so. If a client sends a request, the request should be valid. If it's invalid, it's a client issue and it should be fixed.

brancz:
lgtm

pkg/store/hintspb/hints.proto
```diff
@@ -18,6 +18,12 @@ option (gogoproto.goproto_unkeyed_all) = false;
 option (gogoproto.goproto_unrecognized_all) = false;
 option (gogoproto.goproto_sizecache_all) = false;
 
+message SeriesRequestHints {
+    /// include_blocks is the list of blocks that should be queried. Any other block loaded in the
+    /// store can be skipped if the list is non empty. If the list is empty, no filtering is applied.
+    repeated Block include_blocks = 1 [(gogoproto.nullable) = false];
```
bwplotka, line 24: Crazy idea: To me it really looks like we might want.... BlockMatchers? no? (:

Essentially I am worried that you want to have more flexible filter logic at som point (regexp, != etc). Matcher for arbitrary labels (like we do use reliable) where for now we will have external labels PLUS __block_id (same as we use in relabelling) would do the work.

What do you think? (:

pkg/store/storepb/rpc.proto
```diff
@@ -102,6 +102,11 @@ message SeriesRequest {
 
   // skip_chunks controls whether sending chunks or not in series responses.
   bool skip_chunks = 8;
+
+  // hints is an opaque data structure that can be used to carry additional information.
+  // The content of this field and whether it's supported depends on the
+  // implementation of a specific store.
+  google.protobuf.Any hints = 9;
```
bwplotka, line 109: LGTM

pkg/store/hintspb/hints.proto
pracucci: I think it's a smart idea. Let me give it a try.

pkg/store/bucket.go
```diff
@@ -863,14 +863,27 @@ func (s *BucketStore) Series(req *storepb.SeriesRequest, srv storepb.Store_Serie
...
+		reqHintsMatchers = []*labels.Matcher(nil)
 	)
 
+	if req.Hints != nil {
+		reqHints := &hintspb.SeriesRequestHints{}
+		if err := types.UnmarshalAny(req.Hints, reqHints); err != nil {
+			return status.Error(codes.InvalidArgument, errors.Wrap(err, "unmarshal series request hints").Error())
+		}
+
+		reqHintsMatchers, err = promclient.TranslateMatchers(reqHints.BlockMatchers)
```
bwplotka, line 881: side note: we keep useful helpers in `storepb/custom.go` would be nice to move `TranslateMatchers` there as well at some point :thinking:

bwplotka: Also can make sure those are block matchers (in name)
```suggestion
		reqBlockMatchers, err = promclient.TranslateMatchers(reqHints.BlockMatchers)
```

pkg/store/bucket.go
```diff
@@ -1284,6 +1300,15 @@ func newBucketBlock(
 		enablePostingsCompression: enablePostingsCompression,
 	}
 
+	// Translate the block's labels and inject the block ID as a label
+	// to allow to match blocks also by ID.
+	b.labels = labels.FromMap(meta.Thanos.Labels)
+	b.labels = append(b.labels, labels.Label{
```
bwplotka, line 1306: I hope this is not advertised? (: 

Also that's funny, because technically you could now compose query to filter blocks as well, no?

bwplotka: (because we match normal requests against block external labels.)  I think that's not what we want...

pracucci: > I hope this is not advertised? (:

It's not.

> Also that's funny, because technically you could now compose query to filter blocks as well, no?

I don't think so. We only apply block matchers in the request hints (if any). The block's meta external labels are never manipulated, but a copy is done (via `labels.FromMap()`).

pkg/store/bucket.go
pracucci: If the request hints contains the wrong protobuf message, the unmarshalling will fail and it's expected (I've added a test now). If it contains the right message but with some extra unknown fields, that's fine.

pkg/store/bucket.go
```diff
@@ -1284,6 +1302,15 @@ func newBucketBlock(
 		enablePostingsCompression: enablePostingsCompression,
 	}
 
+	// Translate the block's labels and inject the block ID as a label
+	// to allow to match blocks also by ID.
+	b.matcherLabels = labels.FromMap(meta.Thanos.Labels)
+	b.matcherLabels = append(b.matcherLabels, labels.Label{
```
bwplotka, line 1308: Feels a bit weird to me, what about:
```suggestion
	b.matcherLabels = append(labels.FromMap(meta.Thanos.Labels), labels.Label{
```

pkg/store/bucket.go
```diff
@@ -1344,6 +1371,16 @@ func (b *bucketBlock) chunkReader(ctx context.Context) *bucketChunkReader {
 	return newBucketChunkReader(ctx, b)
 }
 
+// matchLabels verifies whether the block matches the given matchers.
+func (b *bucketBlock) matchLabels(matchers ...*labels.Matcher) bool {
```
bwplotka, line 1375:
```suggestion
func (b *bucketBlock) matchRelabelLabels(matchers ...*labels.Matcher) bool {
```

pkg/store/bucket.go
```diff
@@ -1284,6 +1302,15 @@ func newBucketBlock(
 		enablePostingsCompression: enablePostingsCompression,
 	}
 
+	// Translate the block's labels and inject the block ID as a label
+	// to allow to match blocks also by ID.
+	b.matcherLabels = labels.FromMap(meta.Thanos.Labels)
```
bwplotka, line 1307:
```suggestion
	b.extLabelsAndBlockID = labels.FromMap(meta.Thanos.Labels)
```

no better idea :shrug: 
Maybe `relabelLabels` would do to say that we are it's the same as doing relabel

pkg/store/bucket.go
```diff
@@ -1256,6 +1271,9 @@ type bucketBlock struct {
 	seriesRefetches prometheus.Counter
 
 	enablePostingsCompression bool
+
+	// Block's labels used by block-level matchers to filter blocks to query.
```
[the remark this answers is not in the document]
pracucci: I would expect increase memory utilisation being very small, given these are block-level labels.

pkg/store/bucket.go
```diff
@@ -881,20 +894,21 @@ func (s *BucketStore) Series(req *storepb.SeriesRequest, srv storepb.Store_Serie
...
 		if s.debugLogging {
 			debugFoundBlockSetOverview(s.logger, req.MinTime, req.MaxTime, req.MaxResolutionWindow, bs.labels, blocks)
 		}
 
 		for _, b := range blocks {
 			b := b
 
-			if s.enableSeriesHints {
+			// Check whether this block should be included in the query.
+			if len(reqBlockMatchers) > 0 && !b.matchLabels(reqBlockMatchers...) {
```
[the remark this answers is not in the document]
pracucci: Yes, good point. I've moved the logic to `getFor()`. What's your take?

pracucci:
Thanks @bwplotka for your review. I should have addressed / replied to your comments. What's your take?

pkg/store/bucket.go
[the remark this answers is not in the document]
bwplotka: can agree, relooking (:

PR #2668, opened 2020-05-27
Improve binary index-header documentation

### LINKED ISSUE #2154: docs/store: Document binaryindex-header usage and characteristics.
It would be nice to do before release.

AC:
* Explain how to use index-header for end-user.
* Explain characteristics (low baseline usage, but query spikes are expected)
* Note that it is still experimental.

cc @metalmatze @squat @brancz

pracucci:

## Changes

In this PR I've moved index-header details (ie. format doc) from the Store Gateway doc to a new Operational Guide (from my perspective, the index-header format is not something the entry level user should be aware of). I've also added a mention about how the index header is build and the impact on file descriptors.

What else do you suggest to document?

Fixes #2154.

## Verification

N/A


docs/components/store.md
```diff
@@ -319,61 +319,3 @@ Following options are used for metadata caching (meta.json files, deletion mark
 - `metafile_max_size`: maximum size of cached meta.json and deletion mark file. Larger files are not cached.
 
 Note that chunks and metadata cache is an experimental feature, and these fields may be renamed or removed completely in the future.
-
```
bwplotka, line 322: Can we link those two pages at least? :hugs:

docs/operating/binary-index-header.md
```diff
@@ -0,0 +1,72 @@
...
+---
+
+# Binary index-header
+
+In order to query series inside blocks from object storage, Store Gateway has to know certain initial info about each block such as:
+
+- symbols table to unintern string values
+- postings offset for posting lookup
+
+In order to achieve so, on startup for each block `index-header` is built from pieces of original block's index and stored on disk.
```
bwplotka, line 15: Let's clearly mention that store does not UPLOAD any of this. It's a common misconception.

docs/operating/binary-index-header.md
```diff
@@ -0,0 +1,72 @@
...
+└─────────────────────────────────────────┘
+```
+
+## How the index-header is built
+
+The [Store Gateway](../components/store.md) periodically scans the bucket to look for new and deleted blocks. For each new block found, the Gateway stores the index-header on the local disk, building it with specific sections of the block's index downloaded using GET byte range requests.
+
+## Impact on number of open file descriptors
+
+The Store Gateway stores each block's index-header on the local disk and loads it via mmap. This means that the Gateway keeps a file descriptor for each loaded block. If your Thanos setup has many blocks in the bucket, the Gateway may hit the `file-max` ulimit (maximum number of open file descriptions by a process); in such case, we recommend increasing the limit on your system.
```
bwplotka, line 72: Thanks for those.

I would add one more thing: Characteristics. We cache only 1/32 of symbols and 1/32 of series in memory then we use CPU to lookup further if needed. I think in the issue you fix we wanted to mention this.. Thanks! :hugs:

docs/components/store.md
pracucci: Done. What do you think?

docs/operating/binary-index-header.md
pracucci: Done. What do you think?

docs/operating/binary-index-header.md
pracucci: It wasn't clear to me what were the "Characteristics" to outline. I've added a new "Impact on CPU, memory and disk" paragraph. What's your take?

PR #2669, opened 2020-05-27
Added internal option to disable verification in compactor

pracucci:

## Changes

To my understanding, the block's index/chunks verification done by the compactor (at least the one done on source blocks) is used to try to spot (and a case repair) known Prometheus issues from the past.

I think this make much sense from the Thanos perspective (in a setup running the sidecar, at least), because blocks can be generated from any Prometheus version. In Cortex this looks less critical to me because all blocks are generated by a TSDB version we control (similarly to Thanos receive).

Since verification is quite slow on large blocks (or a large number of source blocks), I would like to investigate what you think if we add an internal option to disable it.

## Verification

Existing tests.

### DIFF
--- a/cmd/thanos/compact.go
+++ b/cmd/thanos/compact.go
@@ -238,6 +238,7 @@ func runCompact(
 			ignoreDeletionMarkFilter,
 			blocksMarkedForDeletion,
 			conf.blockSyncConcurrency,
+			true, // Always enable verification.
 			conf.acceptMalformedIndex, enableVerticalCompaction)
 		if err != nil {
 			return errors.Wrap(err, "create syncer")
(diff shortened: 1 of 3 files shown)


pracucci:
Re-thinking this, what do you think if we do not expose any internal option (or hidden CLI flag) and we do two things:
1. Completely remove the verification on the output block (we assume compactor works fine)
2. We always skip verification when for blocks whose source is receiver?

/cc @brancz

pracucci:
> Re-thinking this, what do you think if we do not expose any internal option (or hidden CLI flag) and we do two things:
> 
> 1. Completely remove the verification on the output block (we assume compactor works fine)
> 2. We always skip verification when for blocks whose source is receiver?

@bwplotka @brancz What's your take about this? If it makes sense I may work on it.

pracucci:
_Closing for now._

PR #2675, opened 2020-05-28
WIP: docs/proposals/receive: Improved routing and distribution config

brancz:

## Changes

After having run Thanos receive at significant scale for some time, we have come to realize that the way we route series through hashrings has limitations, that we would like to address before moving it out of experimental state.

The general problem is, that any 1 remote write request being sufficiently large produces N internal requests in a hashrings (N being the hashring size). This ultimately results in a hashring only being able to scale to the number of requests a single instance can process.

This is WIP to discuss possible solutions, as to how to distribute series of the same tenant across multiple hashrings. I have sketched out what a potential replacement configuration could look like, which decouples routing from hashring definitions. Essentially first a series traverses the routing tree, to find which hashring-group it should be ingested into, and then using the `partition_keys` (identifying a labels within those series) that decide which hashring within the hashring group the series is forwarded to and ultimately ingested into. Partitioning rules could become arbitrarily complex but I think we should start with something simple and just concatenate all label-values of the `partition_keys` and perform consistent hashring to decide which hashring within the hashring-group to choose. The idea being, we want to keep series close to each other if they are of the same origin. A configuration I could image being common could be partition keys being configured to: cluster, instance.

Please let me know your thoughts.

@thanos-io/thanos-maintainers

### DIFF
--- a/docs/proposals/201812_thanos-remote-receive.md
+++ b/docs/proposals/201812_thanos-remote-receive.md
@@ -106,23 +106,40 @@ Hard tenants in the Thanos receiver are configured in a configuration file. Chan
 
 A sample of the configuration of tenants and their respective infrastructure:
 
-```json
-[
-    {
-        "hashring": "tenant-a",
-        "endpoints": ["tenant-a-1.metrics.local:19291/api/v1/receive", "tenant-a-2.metrics.local:19291/api/v1/receive"],
-        "tenants": ["tenant-a"]
-    },
-    {
-        "hashring": "tenants-b-c",
-        "endpoints": ["tenant-b-c-1.metrics.local:19291/api/v1/receive", "tenant-b-c-2.metrics.local:19291/api/v1/receive"],
-        "tenants": ["tenant-b", "tenant-c"]
-    },
-    {
-        "hashring": "soft-tenants",
-        "endpoints": ["http://soft-tenants-1.metrics.local:19291/api/v1/receive"]
-    }
-]
+```yaml
+route:
+- hashring_group: soft-tenants
+  partition_keys:
+  - tenant_id
+  routes:
+  - hashring_group: tenant-a
+    match:
+      tenant_id: tenant-a
+  - hashring_group: tenant-b-c
+    match:
+      tenant_id: tenant-b-c
+hashring_groups:
+- hashring_group: tenant-a
+  hashrings:
+  - hashring: tenant-a-1
+    endpoints:
+    - tenant-a-1-1.metrics.local:19291/api/v1/receive
+    - tenant-a-1-2.metrics.local:19291/api/v1/receive
+  - hashring: tenant-a-2
+    endpoints:
+    - tenant-a-2-1.metrics.local:19291/api/v1/receive
+    - tenant-a-2-2.metrics.local:19291/api/v1/receive
+- hashring_group: tenant-b-c
+  hashrings:
+  - hashring: tenant-b-c
+    endpoints:
+    - tenant-b-c-1.metrics.local:19291/api/v1/receive
+    - tenant-b-c-2.metrics.local:19291/api/v1/receive
+- hashring_group: soft-tenants
+  hashrings:
+  - hashring: soft-tenants
+    endpoints:
+    - soft-tenants-1.metrics.local:19291/api/v1/receive
 ```
 
 To start, exact matches of tenant IDs will be used to distribute requests to receive endpoints. Should it be necessary, more sophisticated mechanisms can be added later. When a request is received, the tenant specified in the request is tested against the configured allowed tenants for each hashring until an exact match is found. If a hashring specifies no explicit tenants, then any tenant is considered a valid match; this allows for a cluster to provide soft-tenancy. Requests whose tenant ID matches no other hashring explicitly, will automatically land in this soft tenancy hashring. If no matching hashring is found and no soft tenancy is configured, the receiver responds with an error.


docs/proposals/201812_thanos-remote-receive.md
```diff
@@ -106,23 +106,40 @@ Hard tenants in the Thanos receiver are configured in a configuration file. Chan
...
-        "hashring": "tenants-b-c",
-        "endpoints": ["tenant-b-c-1.metrics.local:19291/api/v1/receive", "tenant-b-c-2.metrics.local:19291/api/v1/receive"],
-        "tenants": ["tenant-b", "tenant-c"]
-    },
-    {
-        "hashring": "soft-tenants",
-        "endpoints": ["http://soft-tenants-1.metrics.local:19291/api/v1/receive"]
-    }
-]
+```yaml
```
bwplotka, line 109: I like this change already (:

squat: :smile_cat:

docs/proposals/201812_thanos-remote-receive.md
```diff
@@ -106,23 +106,40 @@ Hard tenants in the Thanos receiver are configured in a configuration file. Chan
...
+  - tenant_id
+  routes:
+  - hashring_group: tenant-a
+    match:
+      tenant_id: tenant-a
+  - hashring_group: tenant-b-c
+    match:
+      tenant_id: tenant-b-c
+hashring_groups:
+- hashring_group: tenant-a
```
krasi-georgiev, line 122:
```suggestion
- name: tenant-a
```
less stutter

the same in other places.

docs/proposals/201812_thanos-remote-receive.md
```diff
@@ -106,23 +106,40 @@ Hard tenants in the Thanos receiver are configured in a configuration file. Chan
...
+  hashrings:
+  - hashring: tenant-a-1
```
krasi-georgiev, line 124:
```suggestion
  - name: tenant-a-1
```

docs/proposals/201812_thanos-remote-receive.md
```diff
@@ -106,23 +106,40 @@ Hard tenants in the Thanos receiver are configured in a configuration file. Chan
...
-        "endpoints": ["tenant-b-c-1.metrics.local:19291/api/v1/receive", "tenant-b-c-2.metrics.local:19291/api/v1/receive"],
-        "tenants": ["tenant-b", "tenant-c"]
-    },
-    {
-        "hashring": "soft-tenants",
-        "endpoints": ["http://soft-tenants-1.metrics.local:19291/api/v1/receive"]
-    }
-]
+```yaml
+route:
```
krasi-georgiev, line 110: I don't see an example config for a hard tenant?

docs/proposals/201812_thanos-remote-receive.md
```diff
@@ -106,23 +106,40 @@ Hard tenants in the Thanos receiver are configured in a configuration file. Chan
...
+- hashring_group: soft-tenants
```
krasi-georgiev, line 111: I don't see an example config for a hard tenant?

docs/proposals/201812_thanos-remote-receive.md
```diff
@@ -106,23 +106,40 @@ Hard tenants in the Thanos receiver are configured in a configuration file. Chan
...
+  partition_keys:
+  - tenant_id
+  routes:
```
krasi-georgiev, line 114: having `route` at the top level and `routes` here is a bit confusing. how about:
```suggestion
  target:
```

docs/proposals/201812_thanos-remote-receive.md
squat: The other hashring groups are hard tenants, notice that this one has no `match` section`, which means it is a catch-all. Also note that the hashring group name `soft-tenants` has no magic properties

docs/proposals/201812_thanos-remote-receive.md
brancz: the point here is that routes are nest-able, much like the alertmanager routing tree

docs/proposals/201812_thanos-remote-receive.md
brancz: makes sese

docs/proposals/201812_thanos-remote-receive.md
krasi-georgiev: aah , then the identation is wrong, no? It seems like the other `hashring_group` are under `hashring_group: soft-tenants`

docs/proposals/201812_thanos-remote-receive.md
krasi-georgiev: got it, some commenting would be nice to avoid the confusion I guess.

docs/proposals/201812_thanos-remote-receive.md
squat: not quite, the `hashring_groups` field is for defining a list of all hashrings. The `route` tree is for assigning hashring groups to routes

krasi-georgiev: aah I get it now, but some commenting would be nice.

docs/proposals/201812_thanos-remote-receive.md
brancz: yes makes sense

metalmatze:
I guess there are still some things to be renamed etc, let's see if Frederic is going to address them in the future :)

brancz:
I don't expect myself to get back to this in the short term if anyone wants to take this over, please feel free to do so!

squat:
Ack, i'd be happy to take over, I've been doing some thinking on the subject :))

yeya24:
Does anything block merge this proposal? Is the proposal done?

brancz:
I'm not totally convinced anymore that this is the right thing as opposed to sub-rings like it's done in shuffle-sharding.

bwplotka:
@brancz Makes sense, wonder if it makes sense to set status to `rejected` if that's the case and merge this? (: It's good to document why we did not do some idea (:

PR #2676, opened 2020-05-28
2020 05 tracing bucket

pstibrany:
This PR adds bucket and cache interactions to the traces.

This PR also fixes bug in Thanos tracing code, which first finished the span and only then called the function with given span.

## Verification

Tested manually, see screenshot with small part of example trace.

Screenshot 2020-05-28 at 15 40 09

### DIFF
--- a/pkg/objstore/tracing.go
+++ b/pkg/objstore/tracing.go
@@ -48,7 +48,7 @@ func (t TracingBucket) Get(ctx context.Context, name string) (io.ReadCloser, err
 
 func (t TracingBucket) GetRange(ctx context.Context, name string, off, length int64) (io.ReadCloser, error) {
 	span, spanCtx := tracing.StartSpan(ctx, "bucket_getrange")
-	span.LogKV("name", name, "off", off, "length", length)
+	span.LogKV("name", name, "offset", off, "length", length)
 
 	r, err := t.Bucket.GetRange(spanCtx, name, off, length)
 	if err != nil {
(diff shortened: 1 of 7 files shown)


pkg/objstore/tracing.go
```diff
@@ -0,0 +1,143 @@
...
+func (t TracingBucket) Delete(ctx context.Context, name string) (err error) {
+	tracing.DoWithSpan(ctx, "bucket_delete", func(spanCtx context.Context, span opentracing.Span) {
+		span.LogKV("name", name)
+		err = t.bkt.Delete(spanCtx, name)
+	})
+	return
+}
+
+func (t TracingBucket) Name() string {
+	return "tracing: " + t.bkt.Name()
```
pracucci, line 93: Not sure if we should rename this. Why not just returning `t.bkt.Name()`? We're just adding instrumentation, without any logic change.

pkg/objstore/tracing.go
```diff
@@ -0,0 +1,143 @@
...
+	read int
+}
+
+func (t *tracingReadCloser) Read(p []byte) (int, error) {
+	n, err := t.r.Read(p)
+	if n > 0 {
+		t.read += n
+	}
+	if err != nil && err != io.EOF && t.s != nil {
+		t.s.LogKV("err", err)
```
pracucci, line 127: Shouldn't we finish the span here?

pstibrany: We could do that. My idea was that `Close` should always be called, so we better `Finish` there.

pkg/objstore/tracing.go
pstibrany: This method is only called by metrics bucket, which is not used on top of tracking bucket. Personally I would suggest removing this method completely.

PR #2679, opened 2020-05-28
receive: Let forward requests run until timeout

kakkoyun:

We have observed increase in forward error rate, even though no increase in incoming error rates.
This happens because we cancel the request as soon as we reach the quorum.
This PR lets forward requests run for replication until it actually timeouts.

Screenshot 2020-05-28 17 24 18

## Changes

* Cancel request context only an error occurred and let it optimistically run until it timeouts.

## Verification

* `make test-local`

cc @brancz @squat


kakkoyun:
@brancz @squat Can someone rerun the checks? They failed because of quay incident.

squat:
Lgtm let's gooooo

PR #2684, opened 2020-05-29
change code of conduct to CNCF

povilasv:
There is nothing wrong with current code of conduct a part from being a bit outdated. So I propose instead of having our own code of conduct, we "import" CNCF and add additional information about enforcement.

I believe it's easier for our users to learn/read one Code of Conduct.

This is similar to what other has done (Prometheus, Cortex).

 https://github.com/prometheus/prometheus/blob/master/code-of-conduct.md
https://github.com/cortexproject/cortex/blob/master/code-of-conduct.md

### DIFF
--- a/CODE_OF_CONDUCT.md
+++ b/CODE_OF_CONDUCT.md
@@ -1,53 +1,11 @@
-# Contributor Covenant Code of Conduct
+# Thanos Community Code of Conduct
 
-## Our Pledge
+Thanos follows the [CNCF Code of Conduct](https://github.com/cncf/foundation/blob/master/code-of-conduct.md). Additionally our team members can be contacted directly for code of conduct abuse, as described below.
 
-In the interest of fostering an open and welcoming environment, we as contributors and maintainers pledge to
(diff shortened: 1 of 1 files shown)


CODE_OF_CONDUCT.md
```diff
@@ -1,53 +1,11 @@
...
-include using an official project e-mail address, posting via an official social media account,
-or acting as an appointed representative at an online or offline event. Representation of a project
-may be further defined and clarified by project maintainers.
-
-## Enforcement
+# Enforcement
 
 Instances of abusive, harassing, or otherwise unacceptable behavior may be reported by contacting
-the project team members (e.g @bplotka @domgreen, @Fabian Reinartz) at Thanos slack workspace.
+the project team members (e.g @bwplotka, @povilasv) at CNCF slack workspace.
```
pracucci, line 8: What do you think if we link "CNCF Slack workspace" to https://slack.cncf.io/ ?

povilasv: Fixed

Issue #2405, opened 2020-04-09
query: Optionally make selector labels be required in matchers


dgl:
In Prometheus [remote_read](https://prometheus.io/docs/operating/configuration/#remote_read) configuration there is a `required_matchers` option. This is very similar to selectors in Thanos, but if a matcher is missing from a query Thanos currently sends the query to it (behaviour explained in #44).

It would be nice to be able to specify that a matcher must be present to query a particular store. This would allow to limit queries that are sent to a remote store. Ideally this would be configurable per store defined on query.

Our use case is combined with [Geras](http://github.com/G-Research/geras), potentially expensive Store API queries are made to Geras, even if we know Geras won't have results. This gets more complex when multiple parts of the Thanos hierarchy reference Geras, for example a global thanos-query may reference thanos-query in two clusters, we want queries to go to them without any restriction (i.e. normal Thanos behaviour), but that would mean multiple queries going to Geras (which is fine, because they get deduped, but it hurts latency). Being able to require `{source="geras"}` is specified in queries means we have a way to avoid extra unnecessary queries.

```
                                                +-----------------+
  +------------------------+  {source="geras"}  |                 |
  |                        +------------------->+                 |
  |                        |   (required)       |                 |
  |     Global             |                    |      Geras      |
  |     thanos-query       |                    |                 |
  |                        |                    |                 |
  |                        |                    |                 |
  +--+--------------+------+                    +-------+---------+
     |              |                                   ^
     |              |                                   |
     |              |  {source="thanos"}                |
     |              |  (not required in query)          | {source="geras"}
     |              |                                   | (required)
     |              |                                   |
     v              v                                   |
+----+------+   +---+----------+                        |
|           |   |              +------------------------+
|  thanos   |   |   thanos     |
|  cluster1 |   |   cluster2   |
|           |   |              |
+-----------+   +--------------+
```

The architecture could look something like this, we can use Geras as a custom Store API at either the global or cluster level, by specifying `{source="geras"}`, but because the source label is set to thanos (via `--selector`) on the cluster level thanos-query instances, we don't send queries to the wrong layer of the hierarchy (i.e. each layer can access geras locally, but not via a remote thanos-query instance).

I believe this should be generally applicable as a way to reduce latency in any Thanos configurations with high latency between components though. (We have been emulating this with multiple query instances with different store backends, where ruler talks to a different query backend than users do, but this is quite difficult to reason about, whereas this uses label matching in a more understandable way.)

[...]

PR #2688, opened 2020-05-30
query: Added a new flag --required-matcher

Harshitha1234:
Signed-off-by: Harshitha1234 <harshithachowdary.t17@iiits.in>

## Changes
#2405 Task 2: Added a new flag `--required-matcher` in the querier

## Verification

Tested it locally on dev environment

### DIFF
--- a/cmd/thanos/query.go
+++ b/cmd/thanos/query.go
@@ -74,7 +74,7 @@ func registerQuery(m map[string]setupFunc, app *kingpin.Application) {
 	selectorLabels := cmd.Flag("selector-label", "Query selector labels that will be exposed in info endpoint (repeated).").
 		PlaceHolder("<name>=\"<value>\"").Strings()
 
-	requiredMatcher := cmd.Flag("required-matcher", "Matchers that will be required in order to processes any request to `Series` (or `Select`  via query HTTP). If request matches will not have matchers required, query will return error. (repeated).").
+	requiredMatcher := cmd.Flag("required-matcher", "Experimental: Matchers that will be required in order to processes any request to `Series` (or `Select`  via query HTTP). If request matches will not have matchers required, query will return error. (repeated).").
 		PlaceHolder("<name>=\"<value>\"").Strings()
 
 	stores := cmd.Flag("store", "Addresses of statically configured store API servers (repeatable). The scheme may be prefixed with 'dns+' or 'dnssrv+' to detect store API servers through respective DNS lookups.").
--- a/cmd/thanos/query.go
+++ b/cmd/thanos/query.go
@@ -73,7 +73,10 @@ func registerQuery(m map[string]setupFunc, app *kingpin.Application) {
 
 	selectorLabels := cmd.Flag("selector-label", "Query selector labels that will be exposed in info endpoint (repeated).").
 		PlaceHolder("<name>=\"<value>\"").Strings()
-
+	
+	requiredMatcher := cmd.Flag("required-matcher", "If required matchers are specified in a querier then the querier processes a query only if these labels are present in the query request.").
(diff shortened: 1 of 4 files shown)


pkg/store/proxy.go
```diff
@@ -198,9 +201,37 @@ func (s ctxRespSender) send(r *storepb.SeriesResponse) {
 	s.ch <- r
 }
 
+func matchesRequiredMatchers(ms []storepb.LabelMatcher, requiredMatcher labels.Labels) (bool, error) {
```
yeya24, line 204: This function is almost the same as the function `matchesExternalLabels` below.  What about using that function directly? Maybe we can rename `matchesExternalLabels` to sth else. WDYT?

Harshitha1234: Yeah we can use `matchesExternalLabels` with the following logic
```go
ok, newmatchers, err := matchesExternalLabels(r.Matchers, s.requiredMatcher)
if err != nil {
	return status.Error(codes.InvalidArgument, err.Error())
}
if !ok {
	return nil
}
if len(newmatchers) > len(r.Matchers)-len(s.requiredMatcher) {
	return nil
}
```
WDYT?

cmd/thanos/query.go
```diff
@@ -74,6 +74,9 @@ func registerQuery(m map[string]setupFunc, app *kingpin.Application) {
 	selectorLabels := cmd.Flag("selector-label", "Query selector labels that will be exposed in info endpoint (repeated).").
 		PlaceHolder("<name>=\"<value>\"").Strings()
 
+	requiredMatcher := cmd.Flag("required-matcher", "If required matchers are specified in a querier then the querier processes a query only if these labels are present in the query request.").
```
bwplotka, line 77:
```suggestion
	requiredMatcher := cmd.Flag("required-matcher", "Matchers that will be required in order to processes any request to `Series` (or `Select`  via query HTTP). If request matches will not have matchers required, query will return error. (repeated).").
```

pkg/store/proxy.go
bwplotka: Hm, I don't think this is the same. This is totally agnostic to external matchers at current state.

dgl:
@bwplotka yes, I think this is useful. It may be worth marking the flag as "experimental" as the more useful per store config will look a bit different.

cmd/thanos/query.go
```diff
@@ -74,6 +74,9 @@ func registerQuery(m map[string]setupFunc, app *kingpin.Application) {
 	selectorLabels := cmd.Flag("selector-label", "Query selector labels that will be exposed in info endpoint (repeated).").
 		PlaceHolder("<name>=\"<value>\"").Strings()
 
+	requiredMatcher := cmd.Flag("required-matcher", "Matchers that will be required in order to processes any request to `Series` (or `Select`  via query HTTP). If request matches will not have matchers required, query will return error. (repeated).").
```
bwplotka, line 77: `.Hidden()`

```suggestion
	requiredMatcher := cmd.Flag("required-matcher", "Experimental: Matchers that will be required in order to processes any request to `Series` (or `Select`  via query HTTP). If request matches will not have matchers required, query will return error. (repeated).").
```

+ Let's make it matcher (:

Harshitha1234:
Maybe we could implement this at the storeAPI level in the component where we are using this feature (For example., Geras) as it could be much better than the current feature at per Series level  and maybe we could have the required-matcher feature at per Store level which would really be quite helpful as @dgl suggested. WDYT?

bwplotka:
Yea, we re-review this work and TBH as @Harshitha1234 said, this small feature is something you can totally do now in Geras, right @dgl ? TBH it looks like this feature would be (until we have per client configuration) only for your use case :hugs: (:

PR #2689, opened 2020-05-30
Adding TOC on pages with more than 400 words

### LINKED ISSUE #2512: website (hugo): Add TOC on some pages larger than X words
AC:
 * Setup: https://gohugo.io/content-management/toc/#template-example-toc-partial

This might mean we need to move to new markdown engine.

soniasingla:
Signed-off-by: soniasingla <soniasingla.1812@gmail.com>

## Changes

Fixes #2512 
This PR aims to add TOC to `getting-started.md` page

### DIFF
--- a/go.sum
+++ b/go.sum
@@ -361,6 +361,7 @@ github.com/gobuffalo/attrs v0.0.0-20190224210810-a9411de4debd/go.mod h1:4duuawTq
 github.com/gobuffalo/depgen v0.0.0-20190329151759-d478694a28d3/go.mod h1:3STtPUQYuzV0gBVOY3vy6CfMm/ljR4pABfrTeHNLHUY=
 github.com/gobuffalo/depgen v0.1.0/go.mod h1:+ifsuy7fhi15RWncXQQKjWS9JPkdah5sZvtHc2RXGlg=
 github.com/gobuffalo/envy v1.6.15/go.mod h1:n7DRkBerg/aorDM8kbduw5dN3oXGswK5liaSCx4T5NI=
+github.com/gobuffalo/envy v1.7.0 h1:GlXgaiBkmrYMHco6t4j7SacKO4XUjvh5pwXh0f4uxXU=
 github.com/gobuffalo/envy v1.7.0/go.mod h1:n7DRkBerg/aorDM8kbduw5dN3oXGswK5liaSCx4T5NI=
 github.com/gobuffalo/flect v0.1.0/go.mod h1:d2ehjJqGOH/Kjqcoz+F7jHTBbmDb38yXA598Hb50EGs=
 github.com/gobuffalo/flect v0.1.1/go.mod h1:8JCgGVbRjJhVgD6399mQr4fx5rRfGKVzFjbj6RE/9UI=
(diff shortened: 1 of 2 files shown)


soniasingla:
Yes, that would be best. I'm working on it :)

soniasingla:
This pull request is setting up the automatic TOC above the header. I am blocked after this :)

website/layouts/_default/baseof.html
```diff
@@ -26,6 +26,11 @@
 </head>
 
 <body>
+    {{ if or (eq .Params.toc true) (and (gt .WordCount 400) (ne .Params.toc false)) }}
+    <div class="column is-narrow">
+      {{ .TableOfContents }}
```
bwplotka, line 31: Lol, that was easy =D

website/layouts/_default/baseof.html
```diff
@@ -26,6 +26,11 @@
 </head>
 
 <body>
+    {{ if or (eq .Params.toc true) (and (gt .WordCount 400) (ne .Params.toc false)) }}
```
bwplotka, line 29: :muscle:

soniasingla:
@bwplotka before adding this code in HTML file, i tried by adding this code in a seperate new file but it doesn't reflect changes. In my opinion, the problem was we need to tell where to implement this TOC. I read the official documentation but can't find anything relevant to this :) 

> Ok, I think what we need to do is to NOT change HTML but actually all markdown pages. (:

website/layouts/_default/baseof.html
soniasingla: ❤️

soniasingla:
Adding this code in `single.html` file (maybe)

```
{{ if or (eq .Params.toc true) (and (gt .WordCount 400) (ne .Params.toc false)) }} 
{{ .TableOfContents }}
{{ end }}
```
And, on the pages we want a TOC, setting `toc=true` might work.

+++
...
toc = true
...

I tried this method, but am having doubts with `single.html` file because the directory structure is bit confusing :) Using this method, provides us an advantage that our `HTML` structure is not disturbed :D

Reference : https://www.codeooze.com/webdesign/hugo-toc/

WDYT? @bwplotka @GiedriusS @povilasv 🙂

bwplotka:
No no (: We need to put this in Markdown instead (: Because this is where we want to render it (: 

Now to automate it we can have some quick Perl to make it in scripts or we can put it manually to each markdown file in `docs` (:

povilasv:
> {{ if or (eq .Params.toc true) (and (gt .WordCount 400) (ne .Params.toc false)) }} 
> {{ .TableOfContents }}
> {{ end }}

I like this approach :+1:

povilasv:
In general I like this, but design seems a bit funky, maybe let's add a header and somehow make a list be less spaced out 

image

soniasingla:
May be we can add something like `Working on Thanos` or `Thanos Documentation` or as header. WDYT? @povilasv :)

povilasv:
I was thinking more like:

## Table Of Content
...

## Getting Started

bwplotka:
I think what suggests POV makes sense, otherwise LGTM! :heart:

bwplotka:
Amazing, let's merge it, thanks!

In separate PRs it would be awesome to fix few more things:
* Some links does not work: in case of header being link
* Some sides have manually created TOC, we should remove those :hugs: (https://deploy-preview-2689--thanos-io.netlify.app/contributing/coding-style-guide.md/)

Thanks :muscle:

PR #2687, opened 2020-05-30
shipper: do not check for existence if not needed

GiedriusS:
Skip over already compacted blocks and do not check for their existence
if not needed i.e. appropriate flags have not been specified.

Closes #2686.

Signed-off-by: Giedrius Statkevičius <giedriuswork@gmail.com>

* [ ] I added CHANGELOG entry for this change.
* [x] Change is not relevant to the end user.

## Changes

I moved one check a bit further up so that we would not do unnecessary existence checks.

## Verification

Improved tests that fail without this change.

PR #2696, opened 2020-06-01
shipper: do not check for existence if not needed (#2687)

bwplotka:
Skip over already compacted blocks and do not check for their existence
if not needed i.e. appropriate flags have not been specified.


bwplotka:
Cherry-picked from master.

bwplotka:
https://github.com/thanos-io/thanos/pull/2687

bwplotka:
Actually I don't think it's 0.13 regression so let's drop it.

PR #2603, opened 2020-05-13
store/proxy: Deduplicate chunks on StoreAPI level. Recommend chunk sorting for StoreAPI.

bwplotka:
Blocked by: https://github.com/thanos-io/thanos/pull/2548

Partially fixes: https://github.com/thanos-io/thanos/issues/2303

It should make certain calls where we have overlapped data from StoreAPIs much faster and more efficient.

## Changes: 

* Cases like overlapped data from store and sidecar and 1:1 duplicates are now optimized by removal of exact duplicates as soon as it's possible, on proxy level (fanout receive). This case was highly visible on GitLab repro data and exists in most of the Thanos setups. cc @SuperQ 
* Also this adds logic for merging the same series together on proxy level instead of `select`. This allows for better dedup efficiency.

Signed-off-by: Bartlomiej Plotka <bwplotka@gmail.com>

Issue #2546, opened 2020-04-30
querier: Remove exact chunk duplicates on StoreAPI level, potentially merging frames together.


bwplotka:
This needs some benchmarks as well. Found in https://github.com/thanos-io/thanos/pull/2528

Duplicates are handled ok in terms of data correctness but it's not very efficient. In @SuperQ repro data we found 3x chunk duplications. This is when you have duplicated StoreAPI as well as sidecar + store GW overlap. (:  We can totally trim this on proxy level instead of concating all.

AC: 
* proxy.go returns chunks without duplications
* frames are also "compacted" on up to 1MB (or defined size).

bwplotka:
`proxy.go returns chunks without duplications` was done:  https://github.com/thanos-io/thanos/pull/2603

The defined size AC was not done and it was actually part of another ticket as well plus the idea was abandoned.

PR #2710, opened 2020-06-03
store/proxy: Deduplicate chunks on StoreAPI level. Recommend chunk sorting for StoreAPI + Optimized iter chunk dedup.

bwplotka:
Actually do properly the https://github.com/thanos-io/thanos/issues/2546

This is a rebased version of https://github.com/thanos-io/thanos/pull/2603 - It was wrongly merged to chained, already merged PR. + small fix to avoid `.String()`

This also has to be cherry picked to 0.13

### DIFF
--- a/pkg/query/iter.go
+++ b/pkg/query/iter.go
@@ -62,10 +62,8 @@ func (s *promSeriesSet) Next() bool {
 		return s.currChunks[i].MinTime < s.currChunks[j].MinTime
 	})
 
-	// newChunkSeriesIterator will handle overlaps well, however we don't need to iterate over those samples,
-	// removed early duplicates here.
-	// TODO(bwplotka): Remove chunk duplicates on proxy level as well to avoid decoding those.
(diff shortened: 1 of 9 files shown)


bwplotka:
See benchmarks here: https://github.com/thanos-io/thanos/pull/2603#issuecomment-629651189

Also we desperatedly needs Query benchmarks (there are in some old forgotten PR...)

pkg/query/iter.go
```diff
@@ -81,7 +79,7 @@ func removeExactDuplicates(chks []storepb.AggrChunk) []storepb.AggrChunk {
 	ret = append(ret, chks[0])
 
 	for _, c := range chks[1:] {
-		if ret[len(ret)-1].String() == c.String() {
+		if ret[len(ret)-1].Compare(c) == 0 {
```
bwplotka, line 82: @brancz solving regression here.

Issue #2582, opened 2020-05-10
query: add tests for endpoints + prefix tests


GiedriusS:
Coming from https://github.com/thanos-io/thanos/issues/2514#issuecomment-618513836.

We should add some similar tests to what Prometheus has here: https://github.com/prometheus/prometheus/blob/84b4d079c8714be8e8ad071a35b0391df270364c/web/web_test.go#L92. AFAICT it should be pretty easy to do:

* Refactor the route registering and starting of the HTTP server into a separate function
* Call it from a test & check all of the endpoints
* Add a test on what happens if a prefix has been specified

Should prevent https://github.com/thanos-io/thanos/issues/2581. Making a separate issue for this because this should encompass all of the endpoints that we expose to our users.

bwplotka:
:+1: 

Help wanted

tpaschalis:
I'll take a stab if there's no objection!

kakkoyun:
@GiedriusS Does https://github.com/thanos-io/thanos/pull/2714 satisfy all the needs here?

squat:
hm not quite, I think we should still test for combinations for external web prefix + web prefix. We identified last week that these are not yet completely correctly handled

GiedriusS:
External/route prefix tests are being added in https://github.com/thanos-io/thanos/pull/2800.

GiedriusS:
They have been added, closing.

PR #2714, opened 2020-06-04
query: Add e2e test for web route prefix

### LINKED ISSUE #2581: query: --web.route-prefix makes UI inaccessible

```
./thanos --version                                                                                                                
thanos, version 0.12.2 (branch: master, revision: 060ad9552dbcce9ea50218025a913ad8e99b4471)
  build user:       gstatkevicius@gstatkevicius-desktop
  build date:       20200510-20:54:11
  go version:       go1.14.2
```

**Object Storage Provider**:

N/A

**What happened**:

I have specified `--web.route-prefix` as `FOO` by: `--web.route-prefix=FOO`. Then, I went to http://localhost:10902/FOO.

Expected the user interface to load. Instead, I got a 404.

Run `./thanos query --web.route-prefix="test"`. Go to http://localhost:10902/test.

N/A

ATM when you specify `FOO` as `--web.route-prefix` then:
* A permanent redirect is registered on `/` to `/FOO`
* On `/FOO` a permanent redirect is registered via `q.root` to `/graph`
* And `/graph` is actually registered on `/FOO/graph`

Making an ad-hoc change to `q.root` like this:

```
// Root redirects "/" requests to "/graph", taking into account the path prefix value.
func (q *Query) root(w http.ResponseWriter, r *http.Request) {
	prefix := GetWebPrefix(q.logger, q.externalPrefix, q.prefixHeader, r)

	http.Redirect(w, r, path.Join(prefix, "/test", "/graph"), http.StatusFound)
}
```

Makes the UI available on http://localhost:10902/test again to some degree.

The assets are still being loaded from `/static`, not `/test/static` but that's another problem. I'd expect to see a fully functional UI but that hasn't worked on 0.11.x as well. To get a fully functional UI, you also need to specify `--web.external-prefix="/test"`. But maybe that's by design.

prmsrswt:

## Changes
Add a new e2e test scenario for testing route prefix. I used [chromedp](https://github.com/chromedp/chromedp) to test failed network requests. Right now this test would fail, because we are using absolute URLs for static assets in our templates, and hence we get `404 Not Found`. This is because we are still trying to get assets from `/static/*` while it should be `/{{ routePrefix }}/static/*` as we are using a route prefix.

This isn't exactly what we required in #2582 but it serves the same purpose, and is better in some cases. But I think there's still merit in having unit tests for route prefix and similar.

Fixes #2581 

## Verification
Tested the e2e test by running locally with different scenarios.


test/e2e/query_test.go
```diff
@@ -167,6 +170,53 @@ func TestQuery(t *testing.T) {
...
+	ctx, cancel := chromedp.NewContext(context.Background())
+	defer cancel()
+
+	var networkErrors []string
+
+	// Listen for failed network requests and push them to an array.
+	chromedp.ListenTarget(ctx, func(ev interface{}) {
+		switch ev := ev.(type) {
+		case *network.EventLoadingFailed:
+			networkErrors = append(networkErrors, ev.ErrorText)
```
squat, line 199: Is chromedp guaranteeing to run this func sequentially? Otherwise we might have data races with concurrent access overwriting the networkErrors slice.

test/e2e/query_test.go
```diff
@@ -167,6 +170,53 @@ func TestQuery(t *testing.T) {
...
+	err = chromedp.Run(ctx,
+		network.Enable(),
+		chromedp.Navigate("http://"+q.HTTPEndpoint()+"/test/graph"),
+		chromedp.WaitVisible(`body`),
+	)
+	testutil.Ok(t, err)
+
+	err = func() error {
+		if len(networkErrors) > 0 {
+			return fmt.Errorf("some network requests failed")
```
squat, line 212: if we just return this error, then we won't know _why_ chromedp is failing. How about concatenating the error strings?
```go
return fmt.Errorf("some network requests failed: %s", strings.Join(networkErrors, "; "))
```

test/e2e/query_test.go
prmsrswt: Yes this is synchronous. You have to use goroutines if you want this to handle concurrently. They also recommend using goroutines almost all the time, but it was only an append statement so thought it would only add unnecessary complexity.

test/e2e/query_test.go
prmsrswt: Got it. I'll push the changes.

test/e2e/query_test.go
squat: ack good. the main thing I wanted to ensure was that we wouldn't have a data race. we aren't doing any blocking actions, we operating synchronously is fine :)

pkg/ui/ui.go
```diff
@@ -170,6 +170,12 @@ func GetWebPrefix(logger log.Logger, externalPrefix, prefixHeader string, r *htt
 		level.Warn(logger).Log("msg", "Could not parse value of UI external prefix", "prefix", prefix, "err", err)
 	}
 
+	// To use relative URLs we need the prefix to have trailing "/" so that we don't have to add the
+	// "/" everywhere in templates.
+	if prefix != "" {
+		prefix = prefix + "/"
```
squat, line 176: What if a prefix is given that already ends in `/`? Can we ensure we only have one trailing slash?

prmsrswt: That is already being done in the sanitizePrefix function.

PR #2716, opened 2020-06-04
compact: do not delete compact dir if halt error occurs

GiedriusS:
Let's leave the compactDir if a HaltError occurs to aid in debugging
problems. Without this, it is impossible to inspect the state on disk
after a HaltError occurs.

## Changes

Added a named return for the error in a function. Checked if it is not a halting error before trying to delete `compactDir`.

## Verification

e2e test improvements that fail without this change.

### DIFF
--- a/pkg/compact/compact.go
+++ b/pkg/compact/compact.go
@@ -453,7 +453,7 @@ func (cg *Group) Compact(ctx context.Context, dir string, comp tsdb.Compactor) (
 	subDir := filepath.Join(dir, cg.Key())
 
 	defer func() {
-		if rerr != nil {
+		if IsHaltError(rerr) {
 			return
 		}
 		if err := os.RemoveAll(subDir); err != nil {
--- a/pkg/compact/compact.go
+++ b/pkg/compact/compact.go
@@ -447,12 +447,15 @@ func (cg *Group) Resolution() int64 {
 
 // Compact plans and runs a single compaction against the group. The compacted result
 // is uploaded into the bucket the blocks were retrieved from.
-func (cg *Group) Compact(ctx context.Context, dir string, comp tsdb.Compactor) (bool, ulid.ULID, error) {
+func (cg *Group) Compact(ctx context.Context, dir string, comp tsdb.Compactor) (shouldRerun bool, compID ulid.ULID, rerr error) {
 	cg.compactionRunsStarted.Inc()
 
 	subDir := filepath.Join(dir, cg.Key())
 
 	defer func() {
+		if rerr != nil {
+			return
+		}
 		if err := os.RemoveAll(subDir); err != nil {
 			level.Error(cg.logger).Log("msg", "failed to remove compaction group work directory", "path", subDir, "err", err)
 		}
@@ -879,8 +882,11 @@ func NewBucketCompactor(
 }
 
 // Compact runs compaction over bucket.
-func (c *BucketCompactor) Compact(ctx context.Context) error {
+func (c *BucketCompactor) Compact(ctx context.Context) (rerr error) {
 	defer func() {
+		if IsHaltError(rerr) {
+			return
+		}
 		if err := os.RemoveAll(c.compactDir); err != nil {
 			level.Error(c.logger).Log("msg", "failed to remove compaction work directory", "path", c.compactDir, "err", err)
 		}
--- a/test/e2e/compact_test.go
+++ b/test/e2e/compact_test.go
@@ -8,6 +8,7 @@ import (
 	"context"
 	"encoding/json"
 	"fmt"
+	"io"
 	"net/http"
 	"os"
 	"path"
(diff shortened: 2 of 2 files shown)


GiedriusS:
Oops, my bad - forgot to push the latest changes. e2e tests should pass now.

> lgtm; if we have halt errors then we don't really care about the compact dir growing in size since there are likely bigger problems, right?

The whole process will hang at that point. And there is a:

```go
  935         // Clean up the compaction temporary directory at the beginning of every compaction loop.                                                                                                              
  936         if errRemove := os.RemoveAll(c.compactDir); errRemove != nil {                                                                                                                                         
  937             return errors.Wrap(errRemove, "clean up the compaction temporary directory")                                                                                                                       
  938         } 
```

That ensures we start with a fresh directory.

pkg/compact/compact.go
```diff
@@ -879,10 +879,13 @@ func NewBucketCompactor(
 }
 
 // Compact runs compaction over bucket.
-func (c *BucketCompactor) Compact(ctx context.Context) error {
+func (c *BucketCompactor) Compact(ctx context.Context) (err error) {
 	defer func() {
-		if err := os.RemoveAll(c.compactDir); err != nil {
-			level.Error(c.logger).Log("msg", "failed to remove compaction work directory", "path", c.compactDir, "err", err)
+		if IsHaltError(err) {
```
squat, line 884: hmm actually something here is a little funny.. the `err` variable is never actually used or set, so this is always be false, no?

GiedriusS:
> hmm actually something here is a little funny.. the err variable is never actually used or set, so `IsHaltError` will always return false, no?

No, that's not needed - the named variables get their values according to what's after `return` if they have been explicitly written. For example https://play.golang.org/p/WsvFWYw1iAp. 

Actually, we even have a paragraph about this in our style guide: https://thanos.io/contributing/coding-style-guide.md/#use-named-return-parameters-carefully :smile:

squat:
well, our style guide mentions named returns but says nothing of this particular language feature. In any case, this feature is news to me

pkg/compact/compact.go
```diff
@@ -879,10 +879,13 @@ func NewBucketCompactor(
 }
 
 // Compact runs compaction over bucket.
-func (c *BucketCompactor) Compact(ctx context.Context) error {
+func (c *BucketCompactor) Compact(ctx context.Context) (err error) {
```
pracucci, line 882: I think the whole diff would have been easier if you just named the returned error differently, like `returnErr`.

pkg/compact/compact.go
```diff
@@ -915,40 +918,40 @@ func (c *BucketCompactor) Compact(ctx context.Context) error {
 						continue
 					}
 
-					if IsIssue347Error(err) {
-						if err := RepairIssue347(workCtx, c.logger, c.bkt, c.sy.metrics.blocksMarkedForDeletion, err); err == nil {
+					if IsIssue347Error(errCompact) {
+						if errRepair := RepairIssue347(workCtx, c.logger, c.bkt, c.sy.metrics.blocksMarkedForDeletion, err); errRepair == nil {
```
pracucci, line 922: The `err` passed to `RepairIssue347()` is wrong. It should be `errCompact` (but I would suggest to reconsider just naming differently the returned param, otherwise it's too easy to introduce such bugs).

GiedriusS: Good catch and indeed you are correct that it would be easier to read. I will update.

pkg/compact/compact.go
```diff
@@ -447,12 +447,15 @@ func (cg *Group) Resolution() int64 {
...
 // Compact plans and runs a single compaction against the group. The compacted result
 // is uploaded into the bucket the blocks were retrieved from.
-func (cg *Group) Compact(ctx context.Context, dir string, comp tsdb.Compactor) (bool, ulid.ULID, error) {
+func (cg *Group) Compact(ctx context.Context, dir string, comp tsdb.Compactor) (shouldRerun bool, compID ulid.ULID, rerr error) {
 	cg.compactionRunsStarted.Inc()
 
 	subDir := filepath.Join(dir, cg.Key())
 
 	defer func() {
+		if rerr != nil {
```
pracucci, line 456: Is it correct doing it for any error?

GiedriusS: I think so:

```go
	if err := os.RemoveAll(subDir); err != nil {
		return false, ulid.ULID{}, errors.Wrap(err, "clean compaction group dir")
	}
```
Most likely the same operation will fail again and the deletion of it in `defer` is best-effort either way.


```go
	if err := os.MkdirAll(subDir, 0777); err != nil {
		return false, ulid.ULID{}, errors.Wrap(err, "create compaction group dir")
	}
```

This is not a `HaltError` and then in such case, the caller will clean up the parent directory in a best-effort manner. Is there something, in particular, that doesn't look correct to you?

pracucci: This function can return also if `cg.compact()` returns error. There are many reasons why `cg.compact()` could return error, and it can also return an `HalfError`. The `os.RemoveAll()` done in this `defer` is used to cleanup the blocks downloaded (or created) by `cg.compact()`. For this reason, I'm wondering what we want to achieve with this change. If it's all about preserving the disk state on halt error, then I think we could also check if the HaltError here as well.

PR #2718, opened 2020-06-04
deps: Updated minio-go dependency to v6.0.56 to add two region endpoints

theunisbotha:
This PR is a cherry-pick of #2705, which updates the minio-go dependency to [v6.0.56](https://github.com/minio/minio-go/tree/v6.0.56). This essentially adds two new region endpoints for AWS S3.

Would it be possible for this change to be included in the `0.13.0` release?
We are planning to switch over to the newly added regions in AWS S3, but until the dependency is added, we are stuck using on-prem storage. 

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -11,6 +11,10 @@ We use *breaking* word for marking changes that are not backward compatible (rel
 
 ## Unreleased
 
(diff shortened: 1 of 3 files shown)


theunisbotha:
Thank you! 😁

PR #2725, opened 2020-06-05
store: Expose bucket index operation duration histogram

kakkoyun:
I would like to expose bucket fetch operation durations to fine-tune caching layer for store.

## Changes

* Add metrics to expose already collected fetch durations.

## Verification

* `make build`

### DIFF
--- a/pkg/store/bucket.go
+++ b/pkg/store/bucket.go
@@ -1631,8 +1631,8 @@ type postingPtr struct {
 // It returns one postings for each key, in the same order.
 // If postings for given key is not fetched, entry at given index will be nil.
 func (r *bucketIndexReader) fetchPostings(keys []labels.Label) ([]index.Postings, error) {
-	start := time.Now()
-	defer r.block.metrics.postingsLookupDuration.Observe(time.Since(start).Seconds())
+	timer := prometheus.NewTimer(r.block.metrics.postingsLookupDuration)
+	defer timer.ObserveDuration()
 
 	var ptrs []postingPtr
 
@@ -1851,8 +1851,8 @@ func (it *bigEndianPostings) length() int {
 }
 
 func (r *bucketIndexReader) PreloadSeries(ids []uint64) error {
-	start := time.Now()
-	defer r.block.metrics.seriesLookupDuration.Observe(float64(time.Since(start)))
+	timer := prometheus.NewTimer(r.block.metrics.seriesLookupDuration)
+	defer timer.ObserveDuration()
 
 	// Load series from cache, overwriting the list of ids to preload
 	// with the missing ones.
--- a/pkg/store/bucket.go
+++ b/pkg/store/bucket.go
@@ -226,13 +226,13 @@ func newBucketStoreMetrics(reg prometheus.Registerer) *bucketStoreMetrics {
 	})
 
 	m.postingsFetchDuration = promauto.With(reg).NewHistogram(prometheus.HistogramOpts{
-		Name:    "thanos_bucket_store_series_fetch_duration_seconds",
+		Name:    "thanos_bucket_store_postings_fetch_duration_seconds",
 		Help:    "Time it takes to fetch postings to respond a query.",
 		Buckets: []float64{0.001, 0.01, 0.1, 0.3, 0.6, 1, 3, 6, 9, 20, 30, 60, 90, 120},
 	})
 	m.seriesFetchDuration = promauto.With(reg).NewHistogram(prometheus.HistogramOpts{
-		Name:    "thanos_bucket_store_postings_fetch_duration_seconds",
-		Help:    "Time it takes to fetch sub-results to respond a query.",
+		Name:    "thanos_bucket_store_series_fetch_duration_seconds",
+		Help:    "Time it takes to fetch series to respond a query.",
 		Buckets: []float64{0.001, 0.01, 0.1, 0.3, 0.6, 1, 3, 6, 9, 20, 30, 60, 90, 120},
 	})
 	m.chunksFetchDuration = promauto.With(reg).NewHistogram(prometheus.HistogramOpts{
@@ -987,7 +987,7 @@ func (s *BucketStore) Series(req *storepb.SeriesRequest, srv storepb.Store_Serie
 		s.metrics.cachedPostingsOriginalSizeBytes.Add(float64(stats.cachedPostingsOriginalSizeSum))
 		s.metrics.cachedPostingsCompressedSizeBytes.Add(float64(stats.cachedPostingsCompressedSizeSum))
 		s.metrics.postingsFetchDuration.Observe(stats.postingsFetchDurationSum.Seconds())
-		s.metrics.chunksFetchDuration.Observe(stats.postingsFetchDurationSum.Seconds())
+		s.metrics.chunksFetchDuration.Observe(stats.chunksFetchDurationSum.Seconds())
 		s.metrics.seriesFetchDuration.Observe(stats.seriesFetchDurationSum.Seconds())
 
 		level.Debug(s.logger).Log("msg", "stats query processed",
--- a/pkg/store/bucket.go
+++ b/pkg/store/bucket.go
@@ -106,6 +106,10 @@ type bucketStoreMetrics struct {
 	queriesDropped        prometheus.Counter
 	seriesRefetches       prometheus.Counter
 
+	postingsFetchDuration prometheus.Histogram
+	seriesFetchDuration   prometheus.Histogram
+	chunksFetchDuration   prometheus.Histogram
+
 	cachedPostingsCompressions           *prometheus.CounterVec
 	cachedPostingsCompressionErrors      *prometheus.CounterVec
 	cachedPostingsCompressionTimeSeconds *prometheus.CounterVec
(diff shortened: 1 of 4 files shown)


pkg/store/bucket.go
```diff
@@ -213,6 +217,22 @@ func newBucketStoreMetrics(reg prometheus.Registerer) *bucketStoreMetrics {
 		Help: "Compressed size of postings stored into cache.",
 	})
 
+	m.postingsFetchDuration = promauto.With(reg).NewHistogram(prometheus.HistogramOpts{
+		Name:    "thanos_bucket_store_series_fetch_duration_seconds",
```
pracucci, line 221:
```suggestion
		Name:    "thanos_bucket_store_postings_fetch_duration_seconds",
```

pkg/store/bucket.go
```diff
@@ -213,6 +217,22 @@ func newBucketStoreMetrics(reg prometheus.Registerer) *bucketStoreMetrics {
...
+		Help:    "Time it takes to fetch postings to respond a query.",
+		Buckets: []float64{0.001, 0.01, 0.1, 0.3, 0.6, 1, 3, 6, 9, 20, 30, 60, 90, 120},
+	})
+	m.seriesFetchDuration = promauto.With(reg).NewHistogram(prometheus.HistogramOpts{
+		Name:    "thanos_bucket_store_postings_fetch_duration_seconds",
+		Help:    "Time it takes to fetch sub-results to respond a query.",
```
pracucci, line 227:
```suggestion
		Name:    "thanos_bucket_store_series_fetch_duration_seconds",
		Help:    "Time it takes to fetch series to respond a query.",
```

pkg/store/bucket.go
```diff
@@ -961,6 +981,9 @@ func (s *BucketStore) Series(req *storepb.SeriesRequest, srv storepb.Store_Serie
 		s.metrics.cachedPostingsCompressionTimeSeconds.WithLabelValues("decode").Add(stats.cachedPostingsDecompressionTimeSum.Seconds())
 		s.metrics.cachedPostingsOriginalSizeBytes.Add(float64(stats.cachedPostingsOriginalSizeSum))
 		s.metrics.cachedPostingsCompressedSizeBytes.Add(float64(stats.cachedPostingsCompressedSizeSum))
+		s.metrics.postingsFetchDuration.Observe(stats.postingsFetchDurationSum.Seconds())
+		s.metrics.chunksFetchDuration.Observe(stats.postingsFetchDurationSum.Seconds())
```
pracucci, line 985:
```suggestion
		s.metrics.chunksFetchDuration.Observe(stats.chunksFetchDurationSum.Seconds())
```

kakkoyun:
@pracucci Thanks a lot for correcting my sloppy work 🙈 I haven't finished my morning coffee yet.

pkg/store/bucket.go
```diff
@@ -961,6 +981,9 @@ func (s *BucketStore) Series(req *storepb.SeriesRequest, srv storepb.Store_Serie
...
+		s.metrics.seriesFetchDuration.Observe(stats.seriesFetchDurationSum.Seconds())
```
pracucci: I see the value in exposing timings, but I'm a bit dubious about these timings. These timings are tracked and summed up for every single block, but different blocks are queried concurrently. What you get is the sum of timings across all blocks (which again, are queried concurrently) so from an operation perspective I'm not sure how we can get a real value from these timings because they're influenced by how many blocks you query in a given query.

For example, let's say you have to query 2 blocks (A and B). To fetch postings it take 1s for each block, so the sum is 2s, but since they've been queried concurrently the real time it took was 1s.

kakkoyun: Good point. I'm gonna mark this as work in-progress and try to improve how to track these stats.

pkg/store/bucket.go
```diff
@@ -213,6 +217,22 @@ func newBucketStoreMetrics(reg prometheus.Registerer) *bucketStoreMetrics {
 		Help: "Compressed size of postings stored into cache.",
 	})
 
+	m.postingsFetchDuration = promauto.With(reg).NewHistogram(prometheus.HistogramOpts{
+		Name:    "thanos_bucket_store_postings_fetch_duration_seconds",
+		Help:    "Time it takes to fetch postings to respond a query.",
```
bwplotka, line 222:
```suggestion
		Help:    "Time it takes to fetch postings to respond a query.",
```

Can we elaborate what fetch means? (it means fetch from object storage with cache on top)

Also do you think we can put those in one metric and have label `op` like we have for other metrics? (:

kakkoyun: Yes, of course. Good idea.

pkg/store/bucket.go
bwplotka: That is quite important information yes. However, in fact we have some of this information. Maybe there is a way to correlate this latency with number of postings fetched during that time etc? Would that work? :thinking: 


For me the goal is simple: We want to know if Memcached caching is efficient or not. We want to know if we change object storage if has effect in performance or not. Maybe we want to divide this latency by that number of bytes we fetched overall in given time etc (: There must be some good enough thing we can get from it :thinking: 

Any further ideas @brancz @squat ? :thinking:

pkg/store/bucket.go
```diff
@@ -221,6 +226,14 @@ func newBucketStoreMetrics(reg prometheus.Registerer) *bucketStoreMetrics {
 		Help: "Compressed size of postings stored into cache.",
 	})
 
+	m.indexOperationDuration = promauto.With(reg).NewHistogramVec(prometheus.HistogramOpts{
+		Name: "thanos_bucket_store_index_operation_duration_seconds",
```
pracucci, line 230: [just a question] Have you considered two different metrics, one for postings and one for series? Like `thanos_bucket_store_index_series_lookup_duration_seconds` and ``thanos_bucket_store_index_postings_lookup_duration_seconds`.

pkg/store/bucket.go
```diff
@@ -1610,6 +1629,9 @@ type postingPtr struct {
 // It returns one postings for each key, in the same order.
 // If postings for given key is not fetched, entry at given index will be nil.
 func (r *bucketIndexReader) fetchPostings(keys []labels.Label) ([]index.Postings, error) {
+	start := time.Now()
+	defer r.block.metrics.indexOperationDuration.WithLabelValues(opFetchPostings).Observe(time.Since(start).Seconds())
```
pracucci, line 1633: Should include the time it takes to fetch from cache? Currently it does. If yes, I would mention it in the metric description.

pkg/store/bucket.go
```diff
@@ -1827,6 +1849,9 @@ func (it *bigEndianPostings) length() int {
 }
 
 func (r *bucketIndexReader) PreloadSeries(ids []uint64) error {
+	start := time.Now()
+	defer r.block.metrics.indexOperationDuration.WithLabelValues(opLoadSeries).Observe(float64(time.Since(start)))
```
pracucci, line 1853: Should include the time it takes to fetch from cache? Currently it does. If yes, I would mention it in the metric description.

pkg/store/bucket.go
kakkoyun: Yes, the idea is to use this metric to also benchmark the performance of the index cache. Either different implementation or different configuration of the cache.

I will mention it in the metric name.

pkg/store/bucket.go
kakkoyun: That's what I had done at first. I don't have any strong feelings about it. If you think it would be more clear to have 2 different metrics for this, I'm happy to change it.

pkg/store/bucket.go
kakkoyun: @bwplotka Since the main goal is to have metric to help us to fine-tune caching layer for index lookups. I have changed the implementation and focused on where index cache actually used. Please have a look at it. And let me know if you have any concerns? (on performance etc.)

pkg/store/bucket.go
pracucci: I don't have strong opinions too. We have different metrics for series/postings right now so, maybe, splitting it slightly more consistent with existing metrics but, again, no hard feelings here 😉

pkg/store/bucket.go
```diff
@@ -1610,6 +1631,9 @@ type postingPtr struct {
 // It returns one postings for each key, in the same order.
 // If postings for given key is not fetched, entry at given index will be nil.
 func (r *bucketIndexReader) fetchPostings(keys []labels.Label) ([]index.Postings, error) {
+	start := time.Now()
+	defer r.block.metrics.postingsLookupDuration.Observe(time.Since(start).Seconds())
```
pracucci, line 1635: Isn't `time.Since(start).Seconds()` evaluated immediately? Should you do:

```
defer func() {
    r.block.metrics.postingsLookupDuration.Observe(time.Since(start).Seconds())
}
```

kakkoyun: You have x-ray eyes. Great catch 👍 
TIL there's a timer API https://github.com/prometheus/client_golang/blob/master/prometheus/timer.go#L20
I've fixed it using that.

pkg/store/bucket.go
```diff
@@ -221,6 +224,18 @@ func newBucketStoreMetrics(reg prometheus.Registerer) *bucketStoreMetrics {
 		Help: "Compressed size of postings stored into cache.",
 	})
 
+	m.seriesLookupDuration = promauto.With(reg).NewHistogram(prometheus.HistogramOpts{
+		Name:    "thanos_bucket_store_index_series_lookup_duration_seconds",
+		Help:    "Time it takes to lookup series from a bucket to respond a query. It also includes the cache fetch and store operations.",
+		Buckets: []float64{0.001, 0.01, 0.1, 0.3, 0.6, 1, 3, 6, 9, 20, 30, 60, 90, 120},
+	})
+
+	m.postingsLookupDuration = promauto.With(reg).NewHistogram(prometheus.HistogramOpts{
```
bwplotka, line 233: Can we have just label? It will much easier to aggregate, plus we already do that for other statistics, WDYT?

I think also bucket can be adjusted for timeout. BTW What lookup means is quite vague by this help :thinking:  Maybe we can be explicit more.

pkg/store/bucket.go
```diff
@@ -221,6 +224,18 @@ func newBucketStoreMetrics(reg prometheus.Registerer) *bucketStoreMetrics {
...
+		Name:    "thanos_bucket_store_index_postings_lookup_duration_seconds",
```
bwplotka, line 234:
```suggestion
		Name:    "thanos_bucket_store_index_cachable_fetches_duration_seconds",
```

I would add `cachable` to put highlight that it includes cache (:

pkg/store/bucket.go
kakkoyun: @bwplotka The first version was using a single histogram with labels https://github.com/thanos-io/thanos/pull/2725/commits/8a46f33ac88716c559ac1c0eeefe29143abc04e6. Then I decided to change it, just to be consistent with what we already have in here.

> BTW What lookup means is quite vague by this help 🤔 Maybe we can be explicit more.

I'm gonna change the name as you suggested and try to clarify the description.

> I think also bucket can be adjusted for timeout. 

I've put the boundary at the query timeout. Which timeout value is used for this?

kakkoyun: @bwplotka friendly ping.

bwplotka: ok for now :+1:

PR #2726, opened 2020-06-05
docs: Add receive command docs

kakkoyun:

## Changes

* Add documentation for `receive` command
* Add documentation for multi-tenancy

## Verification

* `make docs`
* `make check-docs`

cc @brancz @squat

ref: https://github.com/thanos-io/thanos/issues/1093

### DIFF
--- a/docs/components/receive.md
+++ b/docs/components/receive.md
@@ -1,0 +1,166 @@
+---
+title: Receiver
+type: docs
+menu: components
+---
+
+# Receiver
+
+The `thanos receive` command implements the [Prometheus Remote Write API](https://prometheus.io/docs/prometheus/latest/configuration/configuration/#remote_write). It builds on top of existing Prometheus servers and retains their usefulness while extending their functionality with long-term-storage, horizontal scalability, and downsampling. The [Thanos Sidecar](./sidecar.md) is not sufficient for this, as the system would always lag the block length behind (typically 2 hours), which would prevent the most common query pattern of Prometheus: real-time queries of very recent data.
+
+This component is only recommended for uses for whom pushing is the only viable solution, for example, analytics use cases or cases where the data ingestion must be client-initiated, such as software as a service type environments.
+
+This component is not recommended in order to achieve a global view of data of a single-tenant, for those cases the sidecar based approach with layered Thanos Queriers is recommended. Multi-tenancy may also be achievable if ingestion is not user-controlled, as then enforcing of labels, for example using the [prom-label-proxy](https://github.com/openshift/prom-label-proxy) (please thoroughly understand the mechanism if intending to employ this mechanism, as the wrong configuration could leak data).
+
+Also, users are asked to note the [various pros and cons of pushing metrics](https://docs.google.com/document/d/1H47v7WfyKkSLMrR8_iku6u9VB73WrVzBHb2SB6dL9_g/edit#heading=h.2v27snv0lsur).
+
+For more information please check out [initial design proposal](../proposals/201812_thanos-remote-receive.md).
+For further information on tuning Prometheus Remote Write [see remote write tuning document](https://prometheus.io/docs/practices/remote_write/).
+
+# Example
+
+```bash
+thanos receive \
+    --tsdb.path        "/path/to/receive/data/dir" \
+    --grpc-address 0.0.0.0:10907 \
+    --http-address 0.0.0.0:10909 \
+    --receive.replication-factor 1 \
+    --label "receive_replica=\"0\"" \
+    --receive.local-endpoint 127.0.0.1:10907 \
+    --receive.hashrings-file ./data/hashring.json \
+    --remote-write.address 0.0.0.0:10908 &
+    --objstore.config-file  "bucket.yml"
+```
+
+The example content of `bucket.yml`:
+
+```yaml
+type: GCS
+config:
+  bucket: example-bucket
+```
+
+The example content of `hashring.json`:
+
+```json
+[
+    {
+        "endpoints": [
+            "127.0.0.1:10907",
+            "127.0.0.1:11907",
+            "127.0.0.1:12907"
+        ]
+    }
+]
+```
+
+## Flags
+
+[embedmd]:# (flags/receive.txt $)
+```$
+usage: thanos receive [<flags>]
+
+Accept Prometheus remote write API requests and write to local tsdb
+(EXPERIMENTAL, this may change drastically without notice)
+
+Flags:
+  -h, --help                     Show context-sensitive help (also try
+                                 --help-long and --help-man).
+      --version                  Show application version.
+      --log.level=info           Log filtering level.
+      --log.format=logfmt        Log format to use. Possible options: logfmt or
+                                 json.
+      --tracing.config-file=<file-path>
+                                 Path to YAML file with tracing configuration.
+                                 See format details:
+                                 https://thanos.io/tracing.md/#configuration
+      --tracing.config=<content>
+                                 Alternative to 'tracing.config-file' flag
+                                 (lower priority). Content of YAML file with
+                                 tracing configuration. See format details:
+                                 https://thanos.io/tracing.md/#configuration
+      --http-address="0.0.0.0:10902"
+                                 Listen host:port for HTTP endpoints.
+      --http-grace-period=2m     Time to wait after an interrupt received for
+                                 HTTP Server.
+      --grpc-address="0.0.0.0:10901"
+                                 Listen ip:port address for gRPC endpoints
+                                 (StoreAPI). Make sure this address is routable
+                                 from other components.
+      --grpc-grace-period=2m     Time to wait after an interrupt received for
+                                 GRPC Server.
+      --grpc-server-tls-cert=""  TLS Certificate for gRPC server, leave blank to
+                                 disable TLS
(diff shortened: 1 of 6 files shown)


docs/components/sidecar.md
```diff
@@ -13,7 +13,7 @@ In details:
...
 
-  NOTE: This still does NOT mean that Prometheus can be fully stateless, because if it crashes and restarts you will lose ~2 hours of metrics, so persistent disk for Prometheus is highly recommended. The closest to stateless you can get is using remote write (which Thanos experimentally supports, see [this](../proposals/201812_thanos-remote-receive.md). Remote write has other risks and consequences, and still if crashed you loose in positive case seconds of metrics data, so persistent disk is recommended in all cases.
+  NOTE: This still does NOT mean that Prometheus can be fully stateless, because if it crashes and restarts you will lose ~2 hours of metrics, so persistent disk for Prometheus is highly recommended. The closest to stateless you can get is using remote write (which Thanos experimentally supports, see [Receiver](./receive.md). Remote write has other risks and consequences, and still if crashed you loose in positive case seconds of metrics data, so persistent disk is recommended in all cases.
```
povilasv, line 16:
```suggestion
  NOTE: This still does NOT mean that Prometheus can be fully stateless, because if it crashes and restarts you will lose ~2 hours of metrics, so persistent disk for Prometheus is highly recommended. The closest to stateless you can get is using remote write (which Thanos supports, see [Receiver](./receive.md). Remote write has other risks and consequences, and still if crashed you loose in positive case seconds of metrics data, so persistent disk is recommended in all cases.
```

is it still experimental?

docs/components/receive.md
```diff
@@ -0,0 +1,166 @@
+---
+title: Receiver
+type: docs
+menu: components
+---
+
+# Receiver
+
+The `thanos receive` command implements the [Prometheus Remote Write API](https://prometheus.io/docs/prometheus/latest/configuration/configuration/#remote_write). It builds on top of existing Prometheus servers and retains their usefulness while extending their functionality with long-term-storage, horizontal scalability, and downsampling. The [Thanos Sidecar](./sidecar.md) is not sufficient for this, as the system would always lag the block length behind (typically 2 hours), which would prevent the most common query pattern of Prometheus: real-time queries of very recent data.
```
metalmatze, line 9: I don't think the last sentence describes this particularly good.
I'm not even fully grasping what you're trying to say.

docs/components/sidecar.md
kakkoyun: That's a good question actually. Maybe by adding these docs we could say it's not experimental anymore. There's a minor issue that has raised by @brancz https://github.com/thanos-io/thanos/issues/1093#issuecomment-630349045

If we think this could be handled incrementally, I'm happy to remove `experimental` mentions, in this PR. What do you think @thanos-io/thanos-maintainers ?

docs/components/receive.md
```diff
@@ -0,0 +1,166 @@
...
+
+This component is only recommended for uses for whom pushing is the only viable solution, for example, analytics use cases or cases where the data ingestion must be client-initiated, such as software as a service type environments.
```
metalmatze, line 11:
```suggestion
We recommend this component to users who can only push into a Thanos due to air-gapped environments. Examples are: analytic use cases or cases where the data ingestion must be client-initiated, such as software as a service type environments.
```

I don't see why the receiver is particularily good for analytic use case compared to the sidecar.

docs/components/receive.md
```diff
@@ -0,0 +1,166 @@
...
+
+This component is not recommended in order to achieve a global view of data of a single-tenant, for those cases the sidecar based approach with layered Thanos Queriers is recommended. Multi-tenancy may also be achievable if ingestion is not user-controlled, as then enforcing of labels, for example using the [prom-label-proxy](https://github.com/openshift/prom-label-proxy) (please thoroughly understand the mechanism if intending to employ this mechanism, as the wrong configuration could leak data).
```
metalmatze, line 13:
```suggestion
We don't recommend the Thanos Receiver in order to achieve a global view of data of a single-tenant. For such use cases, the sidecar based approach with layered Thanos Queriers is recommended. Multi-tenancy may also be achievable if ingestion is not user-controlled, as then enforcing of labels, for example using the [prom-label-proxy](https://github.com/openshift/prom-label-proxy) (please thoroughly understand the mechanism if intending to employ this mechanism, as the wrong configuration could leak data).
```

We we really need to touch on multi-tenancy here?

docs/components/receive.md
```diff
@@ -0,0 +1,166 @@
...
+
+Also, users are asked to note the [various pros and cons of pushing metrics](https://docs.google.com/document/d/1H47v7WfyKkSLMrR8_iku6u9VB73WrVzBHb2SB6dL9_g/edit#heading=h.2v27snv0lsur).
```
metalmatze, line 15:
```suggestion
Additionally, we would like to ask users to note the [various pros and cons of pushing metrics](https://docs.google.com/document/d/1H47v7WfyKkSLMrR8_iku6u9VB73WrVzBHb2SB6dL9_g/edit#heading=h.2v27snv0lsur).
```

metalmatze: Even better is probably something like:
> Please note the various pros and cons of pushing metrics.

docs/components/receive.md
kakkoyun: I've cheated and copied that over proposal document 😅 Happy to remove or paraphrase it.

docs/components/receive.md
kakkoyun: We don't I can remove it. From the proposal as well.

docs/components/sidecar.md
bwplotka: I think for now it's still "experimental" as we want to add one important feature. However we put that for discussion on next Prom Community Meeting cc @brancz

docs/components/receive.md
bwplotka: Agree

bwplotka: At least for now, there is no difference for analytics

docs/components/receive.md
bwplotka: I think it is awesome Kemal touched multitenancy, but I would elaborate more on this part... This sentence might be confusing

docs/components/sidecar.md
bwplotka: I think at least that was my impression, let's ask Frederic. To me it's production ready and I think we can make sure feature we add is compatible

kakkoyun: I've already taken the initiative and removed experimental mentions.

docs/components/receive.md
squat: Hmm for the sentence starting with `It builds on top of existing Prometheus servers...` I think is a little bit fuzzy for people new to Thanos and who aren't familiar with everything. Also if Thanos builds on top of Prometheus. Maybe we should collaborate on a Google doc so we can iterate faster. Wdyt?

kakkoyun: @squat Yes, please. Greate idea. https://docs.google.com/document/d/1H4psm3ancNRyUfg4UilroAEKb96zm_dldH6UnVoLB6A/edit#

docs/components/receive.md
```diff
@@ -0,0 +1,163 @@
...
+title: Receiver
+type: docs
+menu: components
+---
+
+# Receiver
+
+The `thanos receive` command implements the [Prometheus Remote Write API](https://prometheus.io/docs/prometheus/latest/configuration/configuration/#remote_write). It builds on top of existing Prometheus TSDB and retains their usefulness while extending their functionality with long-term-storage, horizontal scalability, and downsampling. It exposes the StoreAPI so that [Thanos Queriers](./query.md) can query received metrics in real-time. The [Thanos Sidecar](./sidecar.md) is not sufficient for this, as the system would always lag the block length behind (typically 2 hours).
+
+We recommend this component to users who can only push into a Thanos due to air-gapped environments. Please note the various pros and cons of pushing metrics.
```
bwplotka, line 11:
```suggestion
We recommend this component to users who can only push into a Thanos due to air-gapped, or egress only environments. Please note the various pros and cons of pushing metrics.
```

Otherwise well put :muscle:

docs/components/receive.md
```diff
@@ -0,0 +1,163 @@
...
+
+Thanos Receive supports multi-tenancy by using labels. We don't recommend the Thanos Receiver in order to achieve a global view of data of a single-tenant. For such use cases, the [Thanos Sidecar](./sidecar.md) based approach with layered [Thanos Queriers](./query.md) is recommended. Note that, multi-tenancy may also be achievable if ingestion is not user-controlled, as then enforcing of labels, for example using the [prom-label-proxy](https://github.com/openshift/prom-label-proxy) (please thoroughly understand the mechanism if intending to employ this mechanism, as the wrong configuration could leak data).
```
bwplotka, line 13: This kind of suggest again that you can achieve multitenancy with receiver only, which is not true. Let's maybe just save this section to some `Multitenancy` Operating doc? :thinking: I would here only say:
```suggestion
Thanos Receive supports multi-tenancy by using labels.  See [Multitenancy documentation here](link).
```

kakkoyun: I have started a doc and linked. FYI it's just bare bones. I can send an another PR to improve it.

Issue #2645, opened 2020-05-22
index out of range [-1]


maulikjs:
**Thanos Version used**: v0.12.2

**Object Storage Provider**: Ceph S3

**What happened**: When I run queries against historic dates, for certain dates we get a response 
`unexpected error: runtime error: index out of range [-1]`

{'field': 'subscribed_cores', 'date': datetime.date(2019, 12, 27), 'runtime': 1.5100421905517578, 'text': '{"status":"error","errorType":"execution","error":"unexpected error: runtime error: index out of range [-1]"}\n', 'value': None, 'status': 'error', 'error': 'unexpected error: runtime error: index out of range [-1]'}

Other than that we also sometimes get an empty response if not the error but never the actual value.

{'field': 'subscribed_cores', 'date': datetime.date(2019, 12, 27), 'runtime': 1.227376699447632, 'text': '{"status":"success","data":{"resultType":"vector","result":[]}}\n', 'value': None, 'status': 'success', 'error': None}

**What you expected to happen**: Get the value for that date.

GiedriusS:
Is there some kind of stack trace printed with the error, perhaps in the Thanos Query logs or somewhere else? Where is this printed in general? Because these don't seem similar to our log format. Do you have a way of reproducing it? :thinking:

maulikjs:
Hey @GiedriusS , 
I was running queries against the API where I got these responses,
I manually also went to the UI and ran the query for the date and got the same output in a error box.
Let me see if I can replicate it and get a screenshot and the pod logs from that.

maulikjs:
![image](https://user-images.githubusercontent.com/22432723/82701813-da253b80-9c3e-11ea-847e-81ade8ff97f8.png)

maulikjs:
```
level=info ts=2020-05-22T19:13:37.505197904Z caller=frontend.go:166 msg="slow query" org_id=fake url="http://telemeter-lts.datahub.redhat.com/api/v1/query?query=round%28avg_over_time%28sum%28sum+by+%28_id%29+%28max+without+%28prometheus%2Creceive%2Cinstance%29+%28+cluster%3Acapacity_cpu_cores%3Asum%29%29+%2B+on+%28_id%29+group_left%28_blah%29+%280+%2A+%28id_version_ebs_account_internal%3Acluster_subscribed%7B%7D+%3D%3D+1%29%29%29%5B1d%3A30m%5D%29%29&time=1587945600.0 " time_taken=6.143809981s
[...]

PR #2728, opened 2020-06-05
querier: Dedup series is now replica label agnostic and simpler. Fixed panic seen when using larger number of replicas and small series.

bwplotka:
Fixes https://github.com/thanos-io/thanos/issues/2645


pkg/query/querier.go
```diff
@@ -226,28 +226,41 @@ func (q *querier) Select(_ bool, hints *storage.SelectHints, ms ...*labels.Match
...
 }
 
-// sortDedupLabels re-sorts the set so that the same series with different replica
-// labels are coming right after each other.
-func sortDedupLabels(set []storepb.Series, replicaLabels map[string]struct{}) {
-	for _, s := range set {
+// trimReplicaLabels removed replica labels from all series and re-sorts the set so that the same series are coming right after each other.
+func trimReplicaLabels(set []storepb.Series, replicaLabels map[string]struct{}) {
+	for si := range set {
+		lset := set[si].Labels
```
yashrsharma44, line 236: Can we name the `lset` to `labelSet`? Sounds more intuitive

bwplotka: I would stick to `lset` I know it's confusing but we are using this name everywhere in Thanos, Prometheus and Cortex. Plus we use `lset` ONLY for label sets (: So with this after learning that `lset` might make sense. What do you think? (:

yashrsharma44: Sure, makes sense to follow the convention 😁

bwplotka:
It should be good to review, just some tests I need to fix: `FAIL: TestQuerier_Select/select_overlapping_data_with_partial_error/dedup=true/querier.Select (0.00s)`

bwplotka:
Should be good to go now.

PR #2665, opened 2020-05-26
fix swift: fixed missing Content-Type headers issue

FUSAKLA:
Signed-off-by: Martin Chodur <m.chodur@seznam.cz>
Fixes https://github.com/thanos-io/thanos/issues/713

@sudhi-vm PTAL as a maintainer of the Swift storage integration

* [x] I added CHANGELOG entry for this change.
* [ ] Change is not relevant to the end user.

## Changes
1. Enabling the [`Full` option in ListOpts](https://godoc.org/github.com/gophercloud/gophercloud/openstack/objectstorage/v1/objects#ListOpts). This should provide missing headers for the gophercloud library to be able to determine content-type of the response. 
1. Fix `GetRange` which did not handle the `length: -1` correctly as not set but used it leading to incorrect range `bytes=1--1` instead of `bytes=1-`. See [the `Range` parameter of the `GET` method in the Swift API docs](https://docs.openstack.org/api-ref/object-store/?expanded=get-object-content-and-metadata-detail#get-object-content-and-metadata).

## Verification

Before, two acceptance e2e tests for object storage were failing
```
=== RUN   TestObjStore_AcceptanceTest_e2e
=== PAUSE TestObjStore_AcceptanceTest_e2e
=== CONT  TestObjStore_AcceptanceTest_e2e
--- FAIL: TestObjStore_AcceptanceTest_e2e (0.94s)
=== RUN   TestObjStore_AcceptanceTest_e2e/inmem
    --- PASS: TestObjStore_AcceptanceTest_e2e/inmem (0.00s)
=== RUN   TestObjStore_AcceptanceTest_e2e/filesystem
=== PAUSE TestObjStore_AcceptanceTest_e2e/filesystem
    TestObjStore_AcceptanceTest_e2e: foreach.go:32: GCS found in THANOS_TEST_OBJSTORE_SKIP array. Skipping.
    TestObjStore_AcceptanceTest_e2e: foreach.go:32: S3 found in THANOS_TEST_OBJSTORE_SKIP array. Skipping.
    TestObjStore_AcceptanceTest_e2e: foreach.go:32: AZURE found in THANOS_TEST_OBJSTORE_SKIP array. Skipping.
=== CONT  TestObjStore_AcceptanceTest_e2e/filesystem
    --- PASS: TestObjStore_AcceptanceTest_e2e/filesystem (0.01s)
=== RUN   TestObjStore_AcceptanceTest_e2e/swift
[...]

PR #2732, opened 2020-06-07
objstore/swift: use ncw/swift and support large files for OpenStack Swift

### LINKED ISSUE #2678: Issue with large files using Swift storage
While migrating from MinIO to Swift I found out that the Gophercloud library does not support large file support out of the box. Swift has by default limit of 5GB for file (our setup has 500MB).
More info here https://docs.openstack.org/swift/latest/overview_large_objects.html

I'd say this is pretty big limitation since the indexes can grow to enormous file size.

We could implement the Large File support with gophercloud ourselves but it's definitely not trivial.
There is also gophercloud/utils where is helper function for upload which implements the support but it requires the file to be on the disk (regarding out bucket interface it would require changes or writing to tmp file) but this could be possibly changed upstream. Also it has no releases and does not look much maintained. https://github.com/gophercloud/utils/blob/master/openstack/objectstorage/v1/objects/upload.go#L119

Other possibility is to use different library for the Swift.
Probably only one which we could consider to use is https://github.com/ncw/swift
It supports those Large Files so should allow Thanos to upload gig indexes.

@sudhi-vm @bwplotka WDYT?

FUSAKLA:
Resolves #2678 
Closes #2665 

As described in the issue, switch to new library to communicate with swift which supports large file types.

I also changed the maintainer of the swift integration to me since if this would be accepted, there would be not much of the old codebase left so it felt fair not to leave the burden on @sudhi-vm.

It supports also working with the large objects which was limit of the previous library.

## Changes
- switched to new library for Swift API
- added support for large files (description in docs)
- added timeouts for Swift API 

## Verification

All tests are passing, tried running it as a sidecar, compactor and store for now on real workload.
Works well but will continue testing it with higher load and  bigger blocks. 

Eventually would like to migrate production on it.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -196,9 +196,10 @@ sse_config:
 - [#2739](https://github.com/thanos-io/thanos/pull/2739) Changed `bucket tool bucket verify` `--id-whitelist` flag to `--id`.
 - [#2748](https://github.com/thanos-io/thanos/pull/2748) Upgrade Prometheus to [@66dfb951c4ca](https://github.com/prometheus/prometheus/commit/66dfb951c4ca2c1dd3f266172a48a925403b13a5) which is after v2.19.0.
   - PromQL now allow us to executed concurrent selects.
-- [#2732](https://github.com/thanos-io/thanos/pull/2732) Swift: Switched to a new library [ncw/swift]() providing large objects support.
+- [#2732](https://github.com/thanos-io/thanos/pull/2732) Swift: Switched to a new library [ncw/swift](https://github.com/ncw/swift) providing large objects support.
    By default, segments will be uploaded to the same container directory `segments/` if the file is bigger than `1GB`.
    To change the defaults see [the docs](./docs/storage.md#openstack-swift).
+
 
 ### Added
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -24,6 +24,9 @@ We use _breaking :warning:_ to mark changes that are not backward compatible (re
 ### Changed
 
 - [#3496](https://github.com/thanos-io/thanos/pull/3496) s3: Respect SignatureV2 flag for all credential providers.
+- [#2732](https://github.com/thanos-io/thanos/pull/2732) Swift: Switched to a new library [ncw/swift](https://github.com/ncw/swift) providing large objects support.
+   By default, segments will be uploaded to the same container directory `segments/` if the file is bigger than `1GB`.
+   To change the defaults see [the docs](./docs/storage.md#openstack-swift).
 
 ## [v0.17.0](https://github.com/thanos-io/thanos/releases/tag/v0.17.0) - 2020.11.18
 
@@ -68,9 +71,6 @@ We use _breaking :warning:_ to mark changes that are not backward compatible (re
   - `thanos_replicate_origin_meta_loads_total` can be replaced by `blocks_meta_synced{state="loaded"}`.
   - `thanos_replicate_origin_partial_meta_reads_total` can be replaced by `blocks_meta_synced{state="failed"}`.
 - [#3309](https://github.com/thanos-io/thanos/pull/3309) Compact: _breaking :warning:_ Rename metrics to match naming convention. This includes metrics starting with `thanos_compactor` to `thanos_compact`, `thanos_querier` to `thanos_query` and `thanos_ruler` to `thanos_rule`.
-- [#2732](https://github.com/thanos-io/thanos/pull/2732) Swift: Switched to a new library [ncw/swift](https://github.com/ncw/swift) providing large objects support.
-   By default, segments will be uploaded to the same container directory `segments/` if the file is bigger than `1GB`.
-   To change the defaults see [the docs](./docs/storage.md#openstack-swift).
 
 ## [v0.16.0](https://github.com/thanos-io/thanos/releases) - 2020.10.26
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -68,6 +68,9 @@ We use _breaking :warning:_ to mark changes that are not backward compatible (re
   - `thanos_replicate_origin_meta_loads_total` can be replaced by `blocks_meta_synced{state="loaded"}`.
   - `thanos_replicate_origin_partial_meta_reads_total` can be replaced by `blocks_meta_synced{state="failed"}`.
 - [#3309](https://github.com/thanos-io/thanos/pull/3309) Compact: _breaking :warning:_ Rename metrics to match naming convention. This includes metrics starting with `thanos_compactor` to `thanos_compact`, `thanos_querier` to `thanos_query` and `thanos_ruler` to `thanos_rule`.
+- [#2732](https://github.com/thanos-io/thanos/pull/2732) Swift: Switched to a new library [ncw/swift](https://github.com/ncw/swift) providing large objects support.
+   By default, segments will be uploaded to the same container directory `segments/` if the file is bigger than `1GB`.
+   To change the defaults see [the docs](./docs/storage.md#openstack-swift).
 
 ## [v0.16.0](https://github.com/thanos-io/thanos/releases) - 2020.10.26
 
@@ -196,10 +199,6 @@ sse_config:
 - [#2739](https://github.com/thanos-io/thanos/pull/2739) Changed `bucket tool bucket verify` `--id-whitelist` flag to `--id`.
 - [#2748](https://github.com/thanos-io/thanos/pull/2748) Upgrade Prometheus to [@66dfb951c4ca](https://github.com/prometheus/prometheus/commit/66dfb951c4ca2c1dd3f266172a48a925403b13a5) which is after v2.19.0.
   - PromQL now allow us to executed concurrent selects.
-- [#2732](https://github.com/thanos-io/thanos/pull/2732) Swift: Switched to a new library [ncw/swift](https://github.com/ncw/swift) providing large objects support.
-   By default, segments will be uploaded to the same container directory `segments/` if the file is bigger than `1GB`.
-   To change the defaults see [the docs](./docs/storage.md#openstack-swift).
-
 
 ### Added
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -196,6 +196,9 @@ sse_config:
 - [#2739](https://github.com/thanos-io/thanos/pull/2739) Changed `bucket tool bucket verify` `--id-whitelist` flag to `--id`.
 - [#2748](https://github.com/thanos-io/thanos/pull/2748) Upgrade Prometheus to [@66dfb951c4ca](https://github.com/prometheus/prometheus/commit/66dfb951c4ca2c1dd3f266172a48a925403b13a5) which is after v2.19.0.
   - PromQL now allow us to executed concurrent selects.
+- [#TBA](https://github.com/thanos-io/thanos/pull/TBA) Swift: Switched to a new library [ncw/swift]() providing large objects support.
+   By default, segments will be uploaded to the same container directory `segments/` if the file is bigger than `1GB`.
+   To change the defaults see [the docs](./docs/storage.md#openstack-swift).
 
 ### Added
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -196,7 +196,7 @@ sse_config:
 - [#2739](https://github.com/thanos-io/thanos/pull/2739) Changed `bucket tool bucket verify` `--id-whitelist` flag to `--id`.
 - [#2748](https://github.com/thanos-io/thanos/pull/2748) Upgrade Prometheus to [@66dfb951c4ca](https://github.com/prometheus/prometheus/commit/66dfb951c4ca2c1dd3f266172a48a925403b13a5) which is after v2.19.0.
   - PromQL now allow us to executed concurrent selects.
-- [#TBA](https://github.com/thanos-io/thanos/pull/TBA) Swift: Switched to a new library [ncw/swift]() providing large objects support.
+- [#2732](https://github.com/thanos-io/thanos/pull/2732) Swift: Switched to a new library [ncw/swift]() providing large objects support.
    By default, segments will be uploaded to the same container directory `segments/` if the file is bigger than `1GB`.
    To change the defaults see [the docs](./docs/storage.md#openstack-swift).
 
--- a/docs/storage.md
+++ b/docs/storage.md
@@ -340,7 +340,7 @@ Various examples for OpenStack authentication can be found in the [official docu
 By default, OpenStack Swift has a limit for maximum file size of 5 GiB. Thanos index files are often larger than that.
 To resolve this issue, Thanos uses [Static Large Objects (SLO)](https://docs.openstack.org/swift/latest/overview_large_objects.html)
 which are uploaded as segments. These are by default put into the `segments` directory of the same container.
-Default limit for using SLO is 1 GiB which is also maximal size of the segment.
+The default limit for using SLO is 1 GiB which is also the maximum size of the segment.
 If you don't want to use the same container for the segments
 (best practise is to use `<container_name>_segments` to avoid polluting listing of the container objects)
 you can use the `large_file_segments_container_name` option to override the default and put the segments to other container.
--- a/docs/storage.md
+++ b/docs/storage.md
@@ -50,7 +50,7 @@ Current object storage client implementations:
 | [Google Cloud Storage](./storage.md#gcs)                                                           | Stable             | Production Usage      | yes               | @bwplotka               |
 | [AWS/S3](./storage.md#s3) (and all S3-compatible storages e.g disk-based [Minio](https://min.io/)) | Stable             | Production Usage      | yes               | @bwplotka               |
 | [Azure Storage Account](./storage.md#azure)                                                        | Stable             | Production Usage      | no                | @vglafirov              |
-| [OpenStack Swift](./storage.md#openstack-swift)                                                    | Beta (working PoC) | Production Usage      | yes               | @sudhi-vm               |
+| [OpenStack Swift](./storage.md#openstack-swift)                                                    | Beta (working PoC) | Production Usage      | yes               | @FUSAKLA                |
 | [Tencent COS](./storage.md#tencent-cos)                                                            | Beta               | Production Usage      | no                | @jojohappy              |
 | [AliYun OSS](./storage.md#aliyun-oss)                                                              | Beta               | Production Usage      | no                | @shaulboozhiao,@wujinhu |
 | [Local Filesystem](./storage.md#filesystem)                                                        | Stable             | Testing and Demo only | yes               | @bwplotka               |
@@ -331,16 +331,25 @@ config:
 
 #### OpenStack Swift
 
-Thanos uses [gophercloud](http://gophercloud.io/) client to upload Prometheus data into [OpenStack Swift](https://docs.openstack.org/swift/latest/).
+Thanos uses [ncw/swift](https://github.com/ncw/swift) client to upload Prometheus data into [OpenStack Swift](https://docs.openstack.org/swift/latest/).
 
 Below is an example configuration file for thanos to use OpenStack swift container as an object store.
 Note that if the `name` of a user, project or tenant is used one must also specify its domain by ID or name.
 Various examples for OpenStack authentication can be found in the [official documentation](https://developer.openstack.org/api-ref/identity/v3/index.html?expanded=password-authentication-with-scoped-authorization-detail#password-authentication-with-unscoped-authorization).
 
+By default, OpenStack Swift has a limit for maximum file size of 5 GiB. Thanos index files are often larger than that.
+To resolve this issue, Thanos uses [Static Large Objects (SLO)](https://docs.openstack.org/swift/latest/overview_large_objects.html)
+which are uploaded as segments. These are by default put into the `segments` directory of the same container.
+Default limit for using SLO is 1 GiB which is also maximal size of the segment.
+If you don't want to use the same container for the segments
+(best practise is to use `<container_name>_segments` to avoid polluting listing of the container objects)
+you can use the `large_file_segments_container_name` option to override the default and put the segments to other container.
+
 [embedmd]:# (flags/config_bucket_swift.txt yaml)
 ```yaml
 type: SWIFT
 config:
+  auth_version: 0
   auth_url: ""
   username: ""
   user_domain_name: ""
@@ -355,6 +364,11 @@ config:
   project_domain_name: ""
   region_name: ""
   container_name: ""
+  large_object_chunk_size: 1073741824
+  large_object_segments_container_name: ""
+  retries: 3
+  connect_timeout: 10s
+  timeout: 5m
 ```
 
 #### Tencent COS
--- a/docs/storage.md
+++ b/docs/storage.md
@@ -344,6 +344,8 @@ The default limit for using SLO is 1 GiB which is also the maximum size of the 
 If you don't want to use the same container for the segments
 (best practise is to use `<container_name>_segments` to avoid polluting listing of the container objects)
 you can use the `large_file_segments_container_name` option to override the default and put the segments to other container.
+_In rare cases you can switch to [Dynamic Large Objects (DLO)](https://docs.openstack.org/swift/latest/overview_large_objects.html)
+by setting the `use_dynamic_large_objects` to true, but use it with caution since it even more relies on eventual consistency._
 
 [embedmd]:# (flags/config_bucket_swift.txt yaml)
 ```yaml
@@ -369,6 +371,7 @@ config:
   retries: 3
   connect_timeout: 10s
   timeout: 5m
+  use_dynamic_large_objects: false
 ```
 
 #### Tencent COS
--- a/pkg/objstore/swift/swift.go
+++ b/pkg/objstore/swift/swift.go
@@ -203,7 +203,7 @@ func (c *Container) Name() string {
 
 // Iter calls f for each entry in the given directory. The argument to f is the full
 // object name including the prefix of the inspected directory.
-func (c *Container) Iter(ctx context.Context, dir string, f func(string) error) error {
+func (c *Container) Iter(_ context.Context, dir string, f func(string) error) error {
 	if dir != "" {
 		dir = strings.TrimSuffix(dir, string(DirDelim)) + string(DirDelim)
 	}
@@ -233,11 +233,11 @@ func (c *Container) get(name string, headers swift.Headers, checkHash bool) (io.
 }
 
 // Get returns a reader for the given object name.
-func (c *Container) Get(ctx context.Context, name string) (io.ReadCloser, error) {
+func (c *Container) Get(_ context.Context, name string) (io.ReadCloser, error) {
 	return c.get(name, swift.Headers{}, true)
 }
 
-func (c *Container) GetRange(ctx context.Context, name string, off, length int64) (io.ReadCloser, error) {
+func (c *Container) GetRange(_ context.Context, name string, off, length int64) (io.ReadCloser, error) {
 	// Set Range HTTP header, see the docs https://docs.openstack.org/api-ref/object-store/?expanded=show-container-details-and-list-objects-detail,get-object-content-and-metadata-detail#id76.
 	bytesRange := fmt.Sprintf("bytes=%d-", off)
 	if length != -1 {
@@ -247,7 +247,7 @@ func (c *Container) GetRange(ctx context.Context, name string, off, length int64
 }
 
 // Attributes returns information about the specified object.
-func (c *Container) Attributes(ctx context.Context, name string) (objstore.ObjectAttributes, error) {
+func (c *Container) Attributes(_ context.Context, name string) (objstore.ObjectAttributes, error) {
 	if name == "" {
 		return objstore.ObjectAttributes{}, errors.New("Object name cannot be empty")
 	}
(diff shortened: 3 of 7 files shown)


CHANGELOG.md
```diff
@@ -24,6 +24,9 @@ We use *breaking* word for marking changes that are not backward compatible (rel
 - [#2667](https://github.com/thanos-io/thanos/pull/2667) Store: removed support to the legacy `index.cache.json`. The hidden flag `--store.disable-index-header` was removed.
 - [#2667](https://github.com/thanos-io/thanos/pull/2667) Compact: the deprecated flag `--index.generate-missing-cache-file` and the metric `thanos_compact_generated_index_total` were removed.
 - [2603](https://github.com/thanos-io/thanos/pull/2603) Store/Querier: Significantly optimize cases where StoreAPIs or blocks returns exact overlapping chunks (e.g Store GW and sidecar or brute force Store Gateway HA).
+- [#TBA](https://github.com/thanos-io/thanos/pull/TBA) Swift: Switched to a new library [ncw/swift]() providing large objects support.
```
bwplotka, line 27: Let's add this PR number :hugs:

bwplotka: plus link

docs/storage.md
```diff
@@ -297,16 +297,25 @@ config:
...
 
-Thanos uses [gophercloud](http://gophercloud.io/) client to upload Prometheus data into [OpenStack Swift](https://docs.openstack.org/swift/latest/).
+Thanos uses [ncw/swift](https://github.com/ncw/swift) client to upload Prometheus data into [OpenStack Swift](https://docs.openstack.org/swift/latest/).
 
 Below is an example configuration file for thanos to use OpenStack swift container as an object store.
 Note that if the `name` of a user, project or tenant is used one must also specify its domain by ID or name.
 Various examples for OpenStack authentication can be found in the [official documentation](https://developer.openstack.org/api-ref/identity/v3/index.html?expanded=password-authentication-with-scoped-authorization-detail#password-authentication-with-unscoped-authorization).
 
+By default, OpenStack Swift has a limit for maximum file size of 5 GiB. Thanos index files are often larger than that.
+To resolve this issue, Thanos uses [Static Large Objects (SLO)](https://docs.openstack.org/swift/latest/overview_large_objects.html)
```
bwplotka, line 307: wow :hugs:

pkg/objstore/swift/swift.go
```diff
@@ -8,75 +8,193 @@ import (
...
+		return nil, errors.Wrap(err, "swift authentication")
+	}
+
+	if _, err := ensureContainer(connection, sc.ContainerName, createContainer); err != nil {
 		return nil, err
 	}
+	if sc.SegmentContainerName == "" {
+		sc.SegmentContainerName = sc.ContainerName
+	} else {
+		if _, err := ensureContainer(connection, sc.SegmentContainerName, createContainer); err != nil {
```
bwplotka, line 185:
```suggestion
	} else if _, err := ensureContainer(connection, sc.SegmentContainerName, createContainer); err != nil {
```

pkg/objstore/swift/swift.go
```diff
@@ -87,206 +205,138 @@ func (c *Container) Name() string {
...
 func (c *Container) Exists(ctx context.Context, name string) (bool, error) {
-	err := objects.Get(c.client, c.name, name, nil).Err
-	if err == nil {
-		return true, nil
-	}
-
-	if _, ok := err.(gophercloud.ErrDefault404); ok {
-		return false, nil
+	_, _, err := c.connection.Object(c.name, name)
+	if err != nil {
```
bwplotka, line 268: I would be explicit here (: 

```go
if c.IsObjNotFoundErr(err) {
  return false, nil
}

if err != nil {
  return false, errors.Wrap(err, "swift check if file exists")
}

return true, nil
```

bwplotka: bit safer (:

FUSAKLA: less indentation is always better :+1: :)

FUSAKLA:
thanks for the _swift_ review :grin: 

I'd like to wait some time to test it out more to be sure it works fine. 
Also I'd love to have some other Swift user to test it out but it seems like nobody active is here around :/ 

Regarding the consistency issues. 
I'm aware of Swift eventual consistency but with recently added deletion marks and delay after upload I believe it could work. My biggest concern is probably downsampling of blocks which were recently created by compaction.
But this could be mitigated by separating those by some time span? :thinking: 

Am I missing out some other cases where it would could blow up?

CHANGELOG.md
```diff
@@ -30,6 +30,9 @@ We use *breaking* word for marking changes that are not backward compatible (rel
 - [#2739](https://github.com/thanos-io/thanos/pull/2739) Changed `bucket tool bucket verify` `--id-whitelist` flag to `--id`.
 - [#2748](https://github.com/thanos-io/thanos/pull/2748) Upgrade Prometheus to [@66dfb951c4ca](https://github.com/prometheus/prometheus/commit/66dfb951c4ca2c1dd3f266172a48a925403b13a5) which is after v2.19.0.
     - PromQL now allow us to executed concurrent selects.
+- [#2732](https://github.com/thanos-io/thanos/pull/2732) Swift: Switched to a new library [ncw/swift]() providing large objects support.
```
GiedriusS, line 33: Seems like we are missing a link here.

docs/storage.md
```diff
@@ -297,16 +297,25 @@ config:
...
+which are uploaded as segments. These are by default put into the `segments` directory of the same container.
+Default limit for using SLO is 1 GiB which is also maximal size of the segment.
```
GiedriusS, line 309: Minor grammar nit: we want a noun here.

```suggestion
The default limit for using SLO is 1 GiB which is also the maximum size of the segment.
```

pkg/objstore/swift/swift.go
```diff
@@ -8,75 +8,191 @@ import (
...
-	DomainName        string `yaml:"domain_name"`
-	ProjectID         string `yaml:"project_id"`
-	ProjectName       string `yaml:"project_name"`
-	ProjectDomainID   string `yaml:"project_domain_id"`
-	ProjectDomainName string `yaml:"project_domain_name"`
-	RegionName        string `yaml:"region_name"`
-	ContainerName     string `yaml:"container_name"`
+const DirDelim = '/'
+
+var DefaultConfig = Config{
```
GiedriusS, line 32: Maybe it would be worth adding `AuthVersion` here with the value `2`? AFAIK there are only two auth versions available and `2` is the newer one, so adding this default value here would mean that all of the Swift users won't need to add it.

FUSAKLA: Hm, Actually there are 3 versions, see for example http://doc.isilon.com/ECS/3.1/DataAccessGuide/ecs_c_oapi_swift_authentication_introduction.html

But we could do better with the default value. The library docs states: `Set to 1, 2 or 3 or leave at 0 for autodetect` so `0` might be the way WDYT?

FUSAKLA: Changed to `0` now

CHANGELOG.md
FUSAKLA: Thanos for noticing, added.

FUSAKLA:
Thanks for another review! :) 
The testing looks promising. I see no obvious issues, but unfortunately our cluster has some issues with the load my testing recording rules are generating so there are some issues caused by this from time to time. I'm now waiting for the maintainers of the cluster to tune it up a bit so there are no errors at all and I'm 100% sure it's OK.

I'm fairly confident the uploading/reading/listing etc are working OK.
My biggest concern causing some unexpected breaking change in the configuration of the bucket.
I hope not but I failed to find any previous user of the Swift integration to verify it :confused:

FUSAKLA:
I'll rebase once https://github.com/thanos-io/thanos/pull/2838 to verify the tests still passes even after the refactor

pkg/objstore/swift/swift.go
GiedriusS: Makes sense, thank you!

FUSAKLA:
I finally found a while to take a look at the failing tests. 
Great idea with running the tests in the CI @GiedriusS , caught two issues actually (one of them regression during CR) :tada: 

Now it's all green so hopefully everything is OK :crossed_fingers: 

I was afraid to merge this right away but it has been a while and it looks like there are no other active users of Swift so I'd go ahead and merge this if you are not against it?

pkg/objstore/swift/swift.go
```diff
@@ -82,215 +206,141 @@ func (c *Container) Name() string {
...
+		objects, err := c.connection.ObjectNames(c.name, opts)
 		if err != nil {
-			return false, err
+			return objects, errors.Wrap(err, "swift list object names")
 		}
-		for _, objectName := range objectNames {
-			if err := f(objectName); err != nil {
-				return false, err
+		for _, object := range objects {
+			if object == SegmentsDir {
```
FUSAKLA, line 219: @bwplotka @GiedriusS PTAL this feels bit hacky, Swift puts the object segments to the `segments` directory in the configured segments container(bucket). To avoid breaking change and requiring another container, the default segments container is the same as for the original data. This means that in the bucket is directory `segments` full of the file parts. 

In one of the tests we check listing of all objects in the bucket, this is an issue in this case since there is additional object now causing the test to fail. As a solution I added skip of this one particular directory but I'm not sure it this is the right way and it feels bit hacky :/

FUSAKLA:
Rebased once again :recycle:

pkg/objstore/swift/swift.go
```diff
@@ -82,215 +206,141 @@ func (c *Container) Name() string {
...
-	// The project can be in another domain than the user, which is indicated by setting either projectDomainName or projectDomainID.
-	switch {
-	case sc.ProjectDomainName != "":
-		authOpts.Scope = &gophercloud.AuthScope{
-			DomainName: sc.ProjectDomainName,
+	if err != nil {
+		return errors.Wrap(err, "swift failed to create file")
+	}
+	defer runutil.CloseWithLogOnErr(c.logger, file, "swift upload obj close")
+	switch f := file.(type) {
```
GiedriusS, line 310: Is there a reason not to use `io.Copy()`?

pkg/objstore/swift/swift.go
```diff
@@ -82,215 +206,141 @@ func (c *Container) Name() string {
...
+	case io.ReaderFrom:
+		if _, err := f.ReadFrom(r); err != nil {
+			return errors.Wrap(err, "swift failed to write uploaded file")
 		}
-	case sc.ProjectDomainID != "":
-		authOpts.Scope = &gophercloud.AuthScope{
-			DomainID: sc.ProjectDomainID,
+	default:
+		buf, err := ioutil.ReadAll(r)
```
GiedriusS, line 316: I believe this might lead to a very huge buffer. Streaming reading might significantly improve performance. This is related to the previous question.

GiedriusS (commented):
Sorry for the delay in the review. Some questions/nits in line.

pkg/objstore/swift/swift.go
```diff
@@ -8,71 +8,195 @@ import (
...
+	ProjectID            string `yaml:"project_id"`
+	ProjectName          string `yaml:"project_name"`
+	ProjectDomainID      string `yaml:"project_domain_id"`
+	ProjectDomainName    string `yaml:"project_domain_name"`
+	RegionName           string `yaml:"region_name"`
+	ContainerName        string `yaml:"container_name"`
+	ChunkSize            int64  `yaml:"large_object_chunk_size"`
+	SegmentContainerName string `yaml:"large_object_segments_container_name"`
+	Retries              int    `yaml:"retries"`
+	ConnectTimeout       string `yaml:"connect_timeout"`
```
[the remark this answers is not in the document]
FUSAKLA: Definitely, should be changed now, thanks makes parsing easier! :+1:

pkg/objstore/swift/swift.go
```diff
@@ -8,71 +8,195 @@ import (
...
+		Retries:        sc.Retries,
+		Region:         sc.RegionName,
+		AuthVersion:    sc.AuthVersion,
+		Tenant:         sc.ProjectName,
+		TenantId:       sc.ProjectID,
+		TenantDomain:   sc.ProjectDomainName,
+		TenantDomainId: sc.ProjectDomainID,
+	}
+	if sc.ConnectTimeout != "" {
+		connectTimeout, err := time.ParseDuration(sc.ConnectTimeout)
```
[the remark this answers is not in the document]
FUSAKLA: done

pkg/objstore/swift/swift.go
```diff
@@ -8,71 +8,195 @@ import (
...
-		name:   sc.ContainerName,
-	}, nil
+	container := Container{
+		logger:            logger,
+		name:              sc.ContainerName,
+		connection:        connection,
+		chunkSize:         sc.ChunkSize,
+		segmentsContainer: sc.SegmentContainerName,
+	}
+	return &container, nil
```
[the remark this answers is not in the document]
FUSAKLA: :+1:

pkg/objstore/swift/swift.go
FUSAKLA: TBH the PR is quite old and I do not remember but I do not see any obvious reason not to :thinking: 

Changed it to `io.Copy` and will see hot the tests end but definitely good point :)

pkg/objstore/swift/swift.go
FUSAKLA: changed to io.copy

FUSAKLA:
@GiedriusS Thanks for going through it in detail. All comments were valid and should be fixed now.

Tests related to Swift are passing but other seems to be flaky probably? :confused: 

PTAL

pkg/objstore/swift/swift.go
FUSAKLA: Ok, I'll leave it like it is

FUSAKLA:
@GiedriusS do you think we could move on with this? I'll rebase

hsmade:
We're bumping into the same issue. We can offer to test on our QA envs, but there we might not have big enough buckets to 'need' the splitting up on uploads.

kakkoyun:
@FUSAKLA Could you rebase again? I can help with the review and push this forward.

pkg/objstore/swift/swift.go
```diff
@@ -9,69 +9,177 @@ import (
...
+	if err := c.ApplyEnvironment(); err != nil {
+		return nil, err
+	}
+
+	config := Config{
+		AuthVersion:          c.AuthVersion,
+		AuthUrl:              c.AuthUrl,
+		Password:             c.ApiKey,
+		Username:             c.UserName,
+		UserId:               c.UserId,
```
GiedriusS, line 85: Very minor thing but in this `Config` we should set `ChunkSize` to `DefaultConfig.ChunkSize`, right? Otherwise it is always zero if `SWIFT_CHUNK_SIZE` is empty.

pkg/objstore/swift/swift.go
```diff
@@ -9,69 +9,177 @@ import (
...
+	connection        *swift.Connection
+	chunkSize         int64
+	segmentsContainer string
 }
 
 func NewContainer(logger log.Logger, conf []byte) (*Container, error) {
 	sc, err := parseConfig(conf)
 	if err != nil {
-		return nil, err
+		return nil, errors.Wrap(err, "swift parse configs")
```
GiedriusS, line 141:
```suggestion
		return nil, errors.Wrap(err, "swift parse config")
```

pkg/objstore/swift/swift.go
```diff
@@ -82,215 +190,130 @@ func (c *Container) Name() string {
...
+		return objects, nil
 	})
 }
 
-// Get returns a reader for the given object name.
-func (c *Container) Get(ctx context.Context, name string) (io.ReadCloser, error) {
+func (c *Container) get(name string, headers swift.Headers, checkHash bool) (io.ReadCloser, error) {
 	if name == "" {
-		return nil, errors.New("error, empty container name passed")
+		return nil, errors.New("Object name cannot be empty")
```
GiedriusS, line 216:
```suggestion
		return nil, errors.New("object name cannot be empty")
```

pkg/objstore/swift/swift.go
```diff
@@ -82,215 +190,130 @@ func (c *Container) Name() string {
...
+	return c.get(name, swift.Headers{"Range": bytesRange}, false)
 }
 
 // Attributes returns information about the specified object.
-func (c *Container) Attributes(ctx context.Context, name string) (objstore.ObjectAttributes, error) {
-	response := objects.Get(c.client, c.name, name, nil)
-	headers, err := response.Extract()
+func (c *Container) Attributes(_ context.Context, name string) (objstore.ObjectAttributes, error) {
+	if name == "" {
+		return objstore.ObjectAttributes{}, errors.New("Object name cannot be empty")
```
GiedriusS, line 242:
```suggestion
		return objstore.ObjectAttributes{}, errors.New("object name cannot be empty")
```

pkg/objstore/swift/swift.go
```diff
@@ -82,215 +190,130 @@ func (c *Container) Name() string {
...
-	if authOpts.Scope != nil {
-		switch {
-		case sc.ProjectName != "":
-			authOpts.Scope.ProjectName = sc.ProjectName
-		case sc.ProjectID != "":
-			authOpts.Scope.ProjectID = sc.ProjectID
-		}
+	defer runutil.CloseWithLogOnErr(c.logger, file, "swift upload obj close")
+	if _, err := io.Copy(file, r); err != nil {
+		return errors.Wrap(err, "failed to write uploaded file to swift")
```
GiedriusS, line 295:
```suggestion
		return errors.Wrap(err, "downloading swift object")
```

pkg/objstore/swift/swift.go
```diff
@@ -299,30 +322,34 @@ func NewTestContainer(t testing.TB) (objstore.Bucket, func(), error) {
...
-
-	c.name = tmpContainerName
-	t.Log("created temporary container for swift tests with name", tmpContainerName)
+	t.Log("created temporary container for swift tests with name", c.Name())
 
 	return c, func() {
 		objstore.EmptyBucket(t, context.Background(), c)
-		if err := c.deleteContainer(tmpContainerName); err != nil {
-			t.Logf("deleting container %s failed: %s", tmpContainerName, err)
+		time.Sleep(time.Second)
```
GiedriusS, line 347: What's the reason for this? :P

GiedriusS (commented):
Very sorry for the long delay in reviewing this. Looks good, just some minor nits.

FUSAKLA:
Ill adress all the comments ASAP, can't wait to see this merged :smile:

pkg/objstore/swift/swift.go
```diff
@@ -82,215 +190,130 @@ func (c *Container) Name() string {
...
-	// Support for cross-domain scoping (user in different domain than project).
-	// If a userDomainName or userDomainID is given, the user is scoped to this domain.
-	switch {
-	case sc.UserDomainName != "":
-		authOpts.DomainName = sc.UserDomainName
-	case sc.UserDomainID != "":
-		authOpts.DomainID = sc.UserDomainID
+	var file io.WriteCloser
+	if size >= c.chunkSize {
+		file, err = c.connection.StaticLargeObjectCreateFile(&swift.LargeObjectOpts{
```
deniszh, line 280: Could we optionally do this with `DynamicLargeObjectCreateFile` too (with some switch and SLO by default)? I know that SLO is used here for consistency, but e.g. our Swift do not support SLO but support DLO.
Much appreciated with effort BTW!

pkg/objstore/swift/swift.go
FUSAKLA: Right, I missed that, thanks!

pkg/objstore/swift/swift.go
FUSAKLA: hm, there is uploading going on? so maybe `uploading swift object` ?  but otherwise :+1: from me for the suggestion

pkg/objstore/swift/swift.go
FUSAKLA: :neutral_face: TBH I do not remember after that time :sweat_smile: 

I'll investigate, maybe there was an issue with the eventual consistency for tests :/

pkg/objstore/swift/swift.go
FUSAKLA: Hi, I suppose it won't hurt to add the support for DLO, but I'll add warning to the docs about it's usage and consequences to scare off the users :)

pkg/objstore/swift/swift.go
FUSAKLA: fixed in https://github.com/thanos-io/thanos/pull/2732/commits/7a51f585a7184cc9a0c5a73ffcd77b23896da026

pkg/objstore/swift/swift.go
FUSAKLA: resolved in https://github.com/thanos-io/thanos/pull/2732/commits/7a51f585a7184cc9a0c5a73ffcd77b23896da026

pkg/objstore/swift/swift.go
FUSAKLA: Added here https://github.com/thanos-io/thanos/pull/2732/commits/7a51f585a7184cc9a0c5a73ffcd77b23896da026

pkg/objstore/swift/swift.go
FUSAKLA: changed here https://github.com/thanos-io/thanos/pull/2732/commits/7a51f585a7184cc9a0c5a73ffcd77b23896da026 to `uploading swift object`, hope that's ok

pkg/objstore/swift/swift.go
FUSAKLA: The DLO support is added here https://github.com/thanos-io/thanos/commit/7a51f585a7184cc9a0c5a73ffcd77b23896da026, check it out.

pkg/objstore/swift/swift.go
```diff
@@ -9,69 +9,188 @@ import (
...
+var DefaultConfig = Config{
+	AuthVersion:    0, // Means autodetect of the auth API version by the library.
+	ChunkSize:      1024 * 1024 * 1024,
+	Retries:        3,
+	ConnectTimeout: model.Duration(10 * time.Second),
+	Timeout:        model.Duration(5 * time.Minute),
+}
+
+// TODO(FUSAKLA): Added to avoid breaking dependency of Cortex which uses the original struct name SwiftConfig.
+type SwiftConfig Config
```
FUSAKLA, line 46: Hm, we managed to get to an interesting situation.. 

Cortex stared to use the old `"github.com/thanos-io/thanos/pkg/objstore/swift".SwiftConfig` which I'm renaming to just `Config`.
But we are obviously embedding Cortex code also so this leads to issue that I'm not able to build id.  

This workaround adds the `SwiftConfig` type just to make it working but I'm not sure if this is a way to go.
Or should I leave the Name of the config to `SwiftConfig`?

 @GiedriusS I'd be really glad for any advice here

FUSAKLA:
Not sure if the e2e test fail is related, does not look like it

deniszh:
Btw, we have agreement in
https://github.com/cortexproject/cortex/pull/3303 to fix Swift support in
Cortex when this PR will be merged.

FUSAKLA:
Cool! thanks for info :tada: :)

pkg/objstore/swift/swift.go
FUSAKLA: The tests passes even with the DLO as I tested it now so hopefully it will work out for you :)

pkg/objstore/swift/swift.go
FUSAKLA: Tested it without the timeout and the tests are passing so I'll just drop it :)
Thanks for noticing it.

FUSAKLA:
Thanks @GiedriusS! 
PTAL once again hopefully the last round :)

pkg/objstore/swift/swift.go
```diff
@@ -9,69 +9,188 @@ import (
 	"fmt"
 	"io"
 	"os"
+	"strconv"
 	"strings"
 	"testing"
+	"time"
+
+	"github.com/prometheus/common/model"
+
```
bwplotka, line 18: formatting is needed, but that's fine

pkg/objstore/swift/swift.go
```diff
@@ -9,69 +9,188 @@ import (
...
-	DomainName        string `yaml:"domain_name"`
-	ProjectID         string `yaml:"project_id"`
-	ProjectName       string `yaml:"project_name"`
-	ProjectDomainID   string `yaml:"project_domain_id"`
-	ProjectDomainName string `yaml:"project_domain_name"`
-	RegionName        string `yaml:"region_name"`
-	ContainerName     string `yaml:"container_name"`
+const (
+	DirDelim    = '/'
+	SegmentsDir = "segments/"
```
bwplotka, line 34: Can we comment? What's this?

pkg/objstore/swift/swift.go
bwplotka: There is a semi-good way to handle this. I will respond with some blog post I think. (: For now, this is fine :+1:

bwplotka, line 46:
```suggestion
type SwiftConfig = Config
```
This makes it clear that it's an alias

pkg/objstore/swift/swift.go
```diff
@@ -9,69 +9,188 @@ import (
...
+}
 
-	client, err := openstack.NewObjectStorageV1(provider, gophercloud.EndpointOpts{
-		Region: sc.RegionName,
-	})
-	if err != nil {
+func NewContainerFromConfig(logger log.Logger, sc *Config, createContainer bool) (*Container, error) {
+	connection := connectionFromConfig(sc)
+	if err := connection.Authenticate(); err != nil {
+		return nil, errors.Wrap(err, "swift authentication")
```
bwplotka, line 175: Normally we don't need to add `swift` to wrap. In the same way, we don't need to add `NewContainerFromConfig swift...`. This is because in Thanos we wrap "whatever we do now" with context, not "where we do it"

pkg/objstore/swift/swift.go
```diff
@@ -82,215 +201,135 @@ func (c *Container) Name() string {
...
 // Exists checks if the given object exists.
-func (c *Container) Exists(ctx context.Context, name string) (bool, error) {
-	err := objects.Get(c.client, c.name, name, nil).Err
-	if err == nil {
-		return true, nil
-	}
-
-	if _, ok := err.(gophercloud.ErrDefault404); ok {
-		return false, nil
+func (c *Container) Exists(_ context.Context, name string) (bool, error) {
```
bwplotka, line 266: None of those support context cancellation?

pkg/objstore/swift/swift.go
bwplotka: still not changed here though

pkg/objstore/swift/swift.go
```diff
@@ -82,215 +201,135 @@ func (c *Container) Name() string {
...
 	}
-	if authOpts.Scope != nil {
-		switch {
-		case sc.ProjectName != "":
-			authOpts.Scope.ProjectName = sc.ProjectName
-		case sc.ProjectID != "":
-			authOpts.Scope.ProjectID = sc.ProjectID
-		}
+	if err != nil {
+		return errors.Wrap(err, "swift failed to create file")
```
bwplotka, line 307:
```suggestion
		return errors.Wrap(err, "create file")
```

bwplotka: Maybe it would nice to get those near each create method so we know if it's statc, dynamic or objectCreate

bwplotka (approved):
LGTM, just small nits.

pkg/objstore/swift/swift.go
FUSAKLA: Great, thanks! I'd be happy to read about that :)

pkg/objstore/swift/swift.go
FUSAKLA: Unfortunately not. The connection lets you pass own transport, but that does not help :/

pkg/objstore/swift/swift.go
FUSAKLA: Moved them right to the creating of the file so the wrap can distinguish SLO, DLO and normal object

pkg/objstore/swift/swift.go
FUSAKLA: Right, I removed the swift reference from all the error wrappings.

pkg/objstore/swift/swift.go
FUSAKLA: Comment added, hopefully it's understandable

pkg/objstore/swift/swift.go
FUSAKLA: sorry, should be fixed now :crossed_fingers:

FUSAKLA:
Thanks for looking at it once again! :+1: 
I hope I addressed all that I could.

Regarding the context cancellation.. I'm not sure we can do much about it here.

FUSAKLA:
@bwplotka @GiedriusS PTAL, should be up to date again
e2e tests not sure why :/

GiedriusS (approved):
The code looks good but I'm kind of hesitant to merge this because the new library doesn't support `context.Context` and that means we have to add yet another timeout parameter. @FUSAKLA perhaps you know if there are any plans to add support for contexts to the library?

FUSAKLA:
@GiedriusS Yep, that is not great but the current state of the swift integration is almost unusable because of the file size limitation.

I opened an issue on the library and it showed up it's already considered (it was originally created before Google context library was created :)) ) so I'd send a PR there to add it.

Considering the current status of the Swift integration and how long is this PR pending, I'd be glad to merge this and ASAP send PR to add the context support once done in the library if you wouldn't mind. Swift is still marked as experimental in the docs so maybe the lack of context is acceptable? 

See the issues: 
https://github.com/ncw/swift/issues/161
https://github.com/ncw/swift/issues/159

bwplotka:
Happy with this approach given the inusability of previous integration. Thanks for this work and sorry for lag in review.

Good work!

PR #2740, opened 2020-06-08
Refactor run command for query component

PhilipGough:

## Changes

@bwplotka  - Related to #2248 and refactors the query component

### DIFF
--- a/cmd/thanos/config.go
+++ b/cmd/thanos/config.go
@@ -137,13 +137,24 @@ func (sc *shipperConfig) registerFlag(cmd *kingpin.CmdClause) *shipperConfig {
 type webConfig struct {
 	externalPrefix   string
 	prefixHeaderName string
+	routePrefix      webRoute
+}
+
+type webRoute struct {
+	prefix string
 }
 
 func (wc *webConfig) registerFlag(cmd *kingpin.CmdClause) *webConfig {
 	cmd.Flag("web.external-prefix",
-		"Static prefix for all HTML links and redirect URLs in the bucket web UI interface. Actual endpoints are still served on / or the web.route-prefix. This allows thanos bucket web UI to be served behind a reverse proxy that strips a URL sub-path.").
+		"Static prefix for all HTML links and redirect URLs in the web UI interface. Actual endpoints are still served on / or the web.route-prefix. This allows thanos web UI to be served behind a reverse proxy that strips a URL sub-path.").
 		Default("").StringVar(&wc.externalPrefix)
 	cmd.Flag("web.prefix-header", "Name of HTTP request header used for dynamic prefixing of UI links and redirects. This option is ignored if web.external-prefix argument is set. Security risk: enable this option only if a reverse proxy in front of thanos is resetting the header. The --web.prefix-header=X-Forwarded-Prefix option can be useful, for example, if Thanos UI is served via Traefik reverse proxy with PathPrefixStrip option enabled, which sends the stripped prefix value in X-Forwarded-Prefix header. This allows thanos UI to be served on a sub-path.").
 		Default("").StringVar(&wc.externalPrefix)
 	return wc
+}
+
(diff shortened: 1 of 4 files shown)


cmd/thanos/config.go
```diff
@@ -11,6 +11,7 @@ import (
 	"gopkg.in/alecthomas/kingpin.v2"
 )
 
+// server config
```
bwplotka, line 14: What do you mean server?

cmd/thanos/config.go
```diff
@@ -129,13 +130,24 @@ func (sc *shipperConfig) registerFlag(cmd *kingpin.CmdClause) *shipperConfig {
 type webConfig struct {
 	externalPrefix   string
 	prefixHeaderName string
+	routePrefix      webRoute
+}
+
+type webRoute struct {
+	prefix string
```
bwplotka, line 137: do we need separate struct? :thinking:

cmd/thanos/query.go
```diff
@@ -479,3 +366,100 @@ func firstDuplicate(ss []string) string {
 
 	return ""
 }
+
+type queryConfig struct {
+	http httpConfig
+	//grpc configs
```
bwplotka, line 372: do we need this comment? (:

cmd/thanos/config.go
PhilipGough: I originally embedded the client config for grpc here but since it was only used in `query` I moved it back out, will remove this comment

cmd/thanos/config.go
PhilipGough: I think we do because the web config is used across components but only a couple fo the use the route prefix, so while it is linked to the webconfig it is optional, and if we embed it then we must register it for all if that makes sense?

Did you have an alternative suggestion?

cmd/thanos/query.go
PhilipGough: need, no :) however all the config in the commented block is related to grpc so was just highlighting that, happy to remove if you want.

cmd/thanos/config.go
PhilipGough: removed

cmd/thanos/query.go
PhilipGough: removed

cmd/thanos/config.go
bwplotka: I think this is ok :hugs:

docs/components/query.md
```diff
@@ -216,12 +216,6 @@ type queryData struct {
 Additional field is `Warnings` that contains every error that occurred that is assumed non critical. `partial_response`
 option controls if storeAPI unavailability is considered critical.
 
-### Concurrent Selects
```
bwplotka, line 219: Ups we lost this during rebase :see_no_evil:

docs/components/query.md
```diff
@@ -28,10 +28,10 @@ Thanos Querier essentially allows to aggregate and optionally deduplicate multip
 
 Since for Querier "a backend" is anything that implements gRPC StoreAPI we can aggregate data from any number of the different storages like:
 
-* Prometheus (see [Sidecar](./sidecar.md))
```
bwplotka, line 31: same here

MalloZup:
still valid

kakkoyun:
@MalloZup Do you want to take it over?

MalloZup:
@kakkoyun  I will try to take over partially o completely on the issue so I'm ok to do the refactor. :+1: ok to me

metonymic-smokey:
@bwplotka could I take over refactoring `query.go` since I think that still lacks a `registerFlag` function and a `queryConfig` type of struct?
I have previously refactored `tools` and `tools_bucket` similarly.

bwplotka:
Yes please!

Kind Regards,
Bartek Płotka ***@***.***)

PR #2746, opened 2020-06-10
feat compact: added flgs to disable compaction and retention

FUSAKLA:
This PR adds 2 new flags `--compaction.disable` and `--retention.disable` to the compactor.
They follow the same logic as already present flag `--downsampling.disable`.

Even though it's usage is discouraged in the help, it allows user to customize his setup and run each part of the compactor job separately. 

I'm personally experimenting with Swift storage because I do not have much other options and this should help to mitigate issues with it's eventual consistency by running those parts separately. 

## Changes

- added new flag `--compaction.disable`
- added new flag `--retention.disable`

## Verification

tested it locally, it really does nothing if everything disabled :grimacing:


FUSAKLA:
CI failed because infracloud doesn't like me :frowning_face: 
image

cmd/thanos/compact.go
```diff
@@ -480,6 +490,12 @@ func (cc *compactConfig) registerFlag(cmd *kingpin.CmdClause) *compactConfig {
 		"as querying long time ranges without non-downsampled data is not efficient and useful e.g it is not possible to render all samples for a human eye anyway").
 		Default("false").BoolVar(&cc.disableDownsampling)
 
+	cmd.Flag("compaction.disable", "Disables compaction. This is not recommended "+
+		"as compaction optimizes the stored data in bucket").
+		Default("false").BoolVar(&cc.disableCompaction)
+
+	cmd.Flag("retention.disable", "Disables retention.").Default("false").BoolVar(&cc.disableRetention)
```
povilasv, line 497: we don't need this as default doesn't do any retention `0d` value

cmd/thanos/compact.go
```diff
@@ -480,6 +490,12 @@ func (cc *compactConfig) registerFlag(cmd *kingpin.CmdClause) *compactConfig {
 		"as querying long time ranges without non-downsampled data is not efficient and useful e.g it is not possible to render all samples for a human eye anyway").
 		Default("false").BoolVar(&cc.disableDownsampling)
 
+	cmd.Flag("compaction.disable", "Disables compaction. This is not recommended "+
```
povilasv, line 493: why are we deploying compactor without compaction? can you describe the usecase

FUSAKLA: As stated in the PR description. I'm using Swift and it's eventually consistent.
My concern is that running for example downsampling right after compaction could lead to higher chance of consistency issues. 
This way I could run compaction, downsampling and retention separately with some delay.
So definitely not disabling compaction just kind of extending variability of the deployment model.   
But I know it's weird use case probably, I'd understand if this would be rejected :)

cmd/thanos/compact.go
FUSAKLA: Yeh, but there is still the process of deleting already marked blocks for deletion which happens regardless on the resolution retention.

povilasv: maybe then let's change the flag to disable just the marked blocks?

povilasv: I don't like having two ways doing the same thing.

cmd/thanos/compact.go
povilasv: Hmm I'm not against this then. Let's see what others think.

PR #2747, opened 2020-06-10
Upgrade Prometheus to fix merge adapter bug

pracucci:

## Changes

Upgrade Prometheus to include the merge adapter bug fix https://github.com/prometheus/prometheus/pull/7369, introduced in Thanos by https://github.com/thanos-io/thanos/pull/2596.

This bug shouldn't affect Thanos (reason why I haven't opened a PR against the `0.13` branch) but it's safer to upgrade the vendored Prometheus. The reason why this bug shouldn't affect Thanos is because `storage.NewMergeSeriesSet()` is only used by the series API, but the resulting `SeriesSet` is iterated sequentially (the bug is triggered when the result of an iteration is retained).

## Verification

_Existing tests_

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -47,7 +47,7 @@ require (
 	github.com/prometheus/client_golang v1.6.0
 	github.com/prometheus/client_model v0.2.0
 	github.com/prometheus/common v0.10.0
-	github.com/prometheus/prometheus v1.8.2-0.20200601152113-3268eac2ddda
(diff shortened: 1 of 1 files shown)


pracucci:
_This PR did nothing, due to override in `go.mod`. The real Prometheus upgrade will happen in https://github.com/thanos-io/thanos/pull/2748._

PR #2748, opened 2020-06-10
Upgrade Prometheus to latest master and reflect changes

kakkoyun:

## Changes

* Upgrade prometheus to https://github.com/prometheus/prometheus/commit/66dfb951c4ca2c1dd3f266172a48a925403b13a5
* Reflect changes

## Verification

* `make build`
* `make test-local`


pkg/query/api/v1.go
```diff
@@ -550,7 +544,7 @@ func (api *API) series(r *http.Request) (interface{}, []error, *ApiError) {
 	if set.Err() != nil {
 		return nil, nil, &ApiError{errorExec, set.Err()}
 	}
-	return metrics, warnings, nil
+	return metrics, set.Warnings(), nil
```
squat, line 547: it looks like nothing in this function returns a single warning anymore; why not change the function signature to eliminate this extra returned value?

kakkoyun: How do you mean? Because of the flow changes warnings returned by the set. And we want to return those warnings to the user, so we have those in the handler function signature. I don't quite get it. Could you elaborate on this?

squat: ah, nevermind, I thought we were always returning an empty list of warning here, but I see we are generating the warnings from the set

PR #2754, opened 2020-06-12
ui: Add Stores page to React UI

prmsrswt:

## Changes
* Migrate the stores page to the React UI

## Verification
Tested locally using both development build of react and the `thanos` binary with production build of react.


GiedriusS:
It crashes for me because in my API response I have `{}` for one `lastError`:

```
Error: Objects are not valid as a React child (found: object with keys {}). If you meant to render a collection of children, use an array instead.
    in span (created by Badge)
    in Badge (at StorePoolPanel.tsx:58)
    in td (at StorePoolPanel.tsx:58)
    in tr (at StorePoolPanel.tsx:47)
    in tbody (at StorePoolPanel.tsx:40)
    in table (created by Table)
    in Table (at StorePoolPanel.tsx:32)
    in div (created by Transition)
    in Transition (created by Collapse)
    in Collapse (at StorePoolPanel.tsx:31)
    in div (created by Container)
    in Container (at StorePoolPanel.tsx:27)
    in StorePoolPanel (at Stores.tsx:16)
    in StoreContent (at withStatusIndicator.tsx:45)
    in Unknown (at Stores.tsx:30)
    in Stores (at App.tsx:34)
```

There are some interesting `error`s, it seems. If I convert it into a string explicitly, I get:

```
fetching store info from XXX:13903: rpc error: code = DeadlineExceeded desc = latest balancer error: connection error: desc = "transport: Error while dialing dial tcp: lookup XXX on YYY:53: no such host"
```

So, I'm certain that it has some weird `MarshalJSON()` function that returns an empty `map`. My suggestion is to return `string`s from the API explicitly. But, I guess, we can improve the API in future PR. WDYT?

pkg/ui/react-app/src/thanos/pages/stores/StorePoolPanel.tsx
```diff
@@ -0,0 +1,69 @@
...
+              return (
+                <tr key={name}>
+                  <td>{name}</td>
+                  <td>
+                    <Badge color={color}>{health.toUpperCase()}</Badge>
+                  </td>
+                  <td>
+                    <StoreLabels labelSet={labelSets} />
+                  </td>
+                  <td>{formatTime(minTime)}</td>
```
GiedriusS, line 55: What about very huge values here? Should we print the infinity sign here if it is huge?

pkg/ui/react-app/src/thanos/pages/stores/StorePoolPanel.tsx
```diff
@@ -0,0 +1,69 @@
...
+                  <td>{formatTime(maxTime)}</td>
```
GiedriusS, line 56: Ditto here.

pkg/ui/react-app/src/thanos/pages/stores/StorePoolPanel.tsx
```diff
@@ -0,0 +1,69 @@
...
+                  <td>{formatRelative(lastCheck, now())} ago</td>
```
GiedriusS, line 57: My guess is that we should apply the same logic here because a StoreAPI might be down for a very _long_ time theoretically, albeit probably not practically.

pkg/ui/react-app/src/thanos/pages/stores/StorePoolPanel.tsx
prmsrswt: I was checking the values and I am getting 
```json
{
    "maxTime": 9223372036854775807,
    "minTime": -62167219200000,
}
```
in the response. The `minTime` is negative and `maxTime` is a very large value, so I thought I need to look deeper, and see from where I am getting these values.

pkg/ui/react-app/src/thanos/pages/stores/StorePoolPanel.tsx
prmsrswt: Oh, gotcha. I haven't considered that but I think it makes sense to have the same check in place here.

prmsrswt:
> So, I'm certain that it has some weird MarshalJSON() function that returns an empty map. My suggestion is to return strings from the API explicitly. But, I guess, we can improve the API in future PR. WDYT?

Returning string explicitly is a good idea. And yeah sure, I'll keep a note of it and fix this in a future PR.

pkg/ui/react-app/src/thanos/pages/stores/StorePoolPanel.tsx
prmsrswt: Okay I think I found the source.
```go
var (
	minTime = time.Unix(math.MinInt64/1000+62135596801, 0)
	maxTime = time.Unix(math.MaxInt64/1000-62135596801, 999999999)
)
```
These are kinda constant values so I guess we can just check for these values and put infinity sign there.

prmsrswt:
I am not sure what do we need to use as `MIN_TIME` constant. Maybe `0`?

Also, please correct me if I am wrong. What I understand from this is that we should check if a particular field is greater than `MAX_TIME` and show infinity sign, else if it is less than the `MIN_TIME`, we should show minus infinity, otherwise just show the parsed date-time.

pkg/ui/react-app/src/thanos/pages/stores/StorePoolPanel.tsx
```diff
@@ -0,0 +1,80 @@
...
+                  <td>
+                    <Badge color={color}>{health.toUpperCase()}</Badge>
+                  </td>
+                  <td>
+                    <StoreLabels labelSet={labelSets} />
+                  </td>
+                  <td>{minTime >= MAX_TIME ? <FontAwesomeIcon icon={faInfinity} /> : formatTime(minTime)}</td>
+                  <td>{maxTime >= MAX_TIME ? <FontAwesomeIcon icon={faInfinity} /> : formatTime(maxTime)}</td>
+                  <td>
+                    {parseTime(lastCheck) >= MAX_TIME ? (
```
GiedriusS, line 62: I was more thinking about 
```suggestion
                    {parseTime(Math.abs(now() - lastCheck)) >= MAX_TIME ? (
```
To be defensive here :P

prmsrswt: `lastCheck` is actually in ISO format (I think, it's a string nonetheless). Guess we need to do
```
Math.abs(now() - parseTime(lastCheck))
```
here.

pkg/ui/react-app/src/thanos/pages/stores/StorePoolPanel.test.tsx
```diff
@@ -0,0 +1,132 @@
...
+        } else {
+          expect(td.text()).toBe(formatTime(maxTime));
+        }
+      });
+
+      it('renders lastCheck', () => {
+        const td = row.find({ 'data-testid': 'lastCheck' });
+        expect(td).toHaveLength(1);
+
+        if (parseTime(lastCheck) >= MAX_TIME) {
```
GiedriusS, line 115: AFAICT the only condition that we don't check in the tests is this but we can add another case in a follow-up PR

pkg/ui/react-app/src/thanos/pages/stores/StorePoolPanel.tsx
```diff
@@ -0,0 +1,84 @@
...
+                <tr key={name}>
+                  <td data-testid="endpoint">{name}</td>
+                  <td data-testid="health">
+                    <Badge color={color}>{health.toUpperCase()}</Badge>
+                  </td>
+                  <td data-testid="storeLabels">
+                    <StoreLabels labelSets={labelSets} />
+                  </td>
+                  <td data-testid="minTime">
+                    {minTime >= MAX_TIME ? <FontAwesomeIcon icon={faInfinity} /> : formatTime(minTime)}
```
squat, line 60: strictly speaking should this rather be `>` instead of `>=`?

pkg/ui/react-app/src/thanos/pages/stores/StorePoolPanel.tsx
```diff
@@ -0,0 +1,84 @@
...
+                  </td>
+                  <td data-testid="maxTime">
+                    {maxTime >= MAX_TIME ? <FontAwesomeIcon icon={faInfinity} /> : formatTime(maxTime)}
```
squat, line 63: same

pkg/ui/react-app/src/thanos/pages/stores/StoreLabels.tsx
```diff
@@ -0,0 +1,21 @@
...
+import { Badge, ListGroup, ListGroupItem } from 'reactstrap';
+import { Labels } from './store';
+
+export type StoreLabelsProps = { labelSet: Labels[] };
+
+export const StoreLabels: FC<StoreLabelsProps> = ({ labelSet }) => {
+  return (
+    <ListGroup>
+      {labelSet.map(({ labels }, idx) => (
+        <ListGroupItem key={idx}>
```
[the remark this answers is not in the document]
GiedriusS: We haven't found any other loops that use some other name and other code uses `idx` as well

pkg/ui/react-app/src/thanos/pages/stores/StorePoolPanel.tsx
GiedriusS: Talked about this in our sync - this is OK for now. It would be worth investigating whether momentjs has some kind of constant or function that we could compare against i.e. highest possible date that it supports.

PR #2759, opened 2020-06-14
ui: Add error boundary in the React UI

prmsrswt:

## Changes
- Add a error boundary to the React UI. Now in case of an irrecoverable error, users would see a "Something went wrong" page instead of a blank screen.
- Fix the crash when there is no store connected to the Querier. The crash was because the API response for `/api/v1/label/__name__/values` was
```json
{
  "status": "success",
  "data": null
}
```
Here, the UI expects `data` to be an array.
Fixed this by sending an empty array (`[]`) instead of `null`

## Verification
Tested locally by emulating error conditions.

## Screenshot
Error Boundary

### DIFF
--- a/pkg/ui/react-app/src/thanos/pages/errorBoundary/ErrorBoundary.tsx
+++ b/pkg/ui/react-app/src/thanos/pages/errorBoundary/ErrorBoundary.tsx
@@ -28,15 +28,15 @@ class ErrorBoundary extends React.Component<{}, ErrorState> {
         <Container fluid className={styles.container}>
           <h1>Aaaah! Something went wrong.</h1>
           <h3>
-            Please file an issue at&nbsp;
+            Please file an issue in the&nbsp;
             <a href="https://github.com/thanos-io/thanos/issues" target="_blank" rel="noreferrer noopener">
               Thanos issue tracker.
             </a>
           </h3>
-          <Button color="link" id="toggler" className={styles.detailsBtn} style={{ fontSize: '1.2em' }}>
+          <Button color="link" id="error-details-toggler" className={styles.detailsBtn}>
             View error details.
           </Button>
-          <UncontrolledCollapse toggler="#toggler" className={styles.errorDiv}>
+          <UncontrolledCollapse toggler="#error-details-toggler" className={styles.errorDiv}>
             <span>{this.state.error && this.state.error.toString()}</span>
             {this.state.errorInfo.componentStack}
           </UncontrolledCollapse>
--- a/pkg/ui/react-app/src/thanos/pages/errorBoundary/ErrorBoundary.tsx
+++ b/pkg/ui/react-app/src/thanos/pages/errorBoundary/ErrorBoundary.tsx
@@ -1,0 +1,50 @@
+import React, { ErrorInfo } from 'react';
+import { Container, UncontrolledCollapse, Button } from 'reactstrap';
+import styles from './ErrorBoundary.module.css';
+
+interface ErrorState {
+  error: Error | null;
+  errorInfo: ErrorInfo | null;
+}
+
+class ErrorBoundary extends React.Component<{}, ErrorState> {
+  constructor(props: {}) {
+    super(props);
+    this.state = {
+      error: null,
+      errorInfo: null,
+    };
+  }
+
+  componentDidCatch(error: Error, errorInfo: ErrorInfo): void {
+    this.setState({
+      error,
+      errorInfo,
+    });
+  }
+  render(): React.ReactNode {
+    if (this.state.errorInfo) {
+      return (
+        <Container fluid className={styles.container}>
+          <h1>Aaaah! Something went wrong.</h1>
+          <h3>
+            Please file an issue at&nbsp;
+            <a href="https://github.com/thanos-io/thanos/issues" target="_blank" rel="noreferrer noopener">
+              Thanos issue tracker.
+            </a>
+          </h3>
+          <Button color="link" id="toggler" className={styles.detailsBtn} style={{ fontSize: '1.2em' }}>
+            View error details.
+          </Button>
+          <UncontrolledCollapse toggler="#toggler" className={styles.errorDiv}>
+            <span>{this.state.error && this.state.error.toString()}</span>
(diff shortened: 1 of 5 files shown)


pkg/strutil/merge.go
```diff
@@ -12,7 +12,7 @@ import (
 // while removing all duplicates.
 func MergeSlices(a ...[]string) []string {
 	if len(a) == 0 {
-		return nil
+		return make([]string, 0)
```
bwplotka, line 15: This really not the best pattern, can we avoid this?

pkg/ui/react-app/src/thanos/pages/errorBoundary/ErrorBoundary.tsx
```diff
@@ -0,0 +1,50 @@
...
+    this.setState({
+      error,
+      errorInfo,
+    });
+  }
+  render(): React.ReactNode {
+    if (this.state.errorInfo) {
+      return (
+        <Container fluid className={styles.container}>
+          <h1>Aaaah! Something went wrong.</h1>
```
bwplotka, line 29: Just curious, wonder if inline HTML makes sense. What would be the benefit between this and a separate file for HTML (:

pkg/strutil/merge.go
prmsrswt: Okay, so I did this because when there isn't a store API connected the Thanos Querier, it returns `{ "data": null }` because the function `MergeSlices` returns `nil`. This causes the UI to error out.

What I wanted to do was to return `{ "data": [] }` instead, so that we don't get an error in the UI saying `data` is null, expected array, and as we can see in this playground https://play.golang.org/p/zDLyXcc8ie the only way to return an empty array is to use `make` to create the array.

Can you suggest me some alternative to this? One solution that came to my mind is to check if the array is nil and then return an empty array instead in the `pkg/query/api/v1.go`
```go

func (api *API) labelNames(r *http.Request) (interface{}, []error, *ApiError) {
	// redacted

	names, warnings, err := q.LabelNames()
	if err != nil {
		return nil, nil, &ApiError{errorExec, err}
	}

	// names would be nil in case there is no store api connected to querier
	if names == nil {
		names = make([]string, 0)
	}

	return names, warnings, nil
}
```

pkg/ui/react-app/src/thanos/pages/errorBoundary/ErrorBoundary.tsx
prmsrswt: Sorry, I think I am not able to understand what are you trying to say. This isn't actually inline HTML, it's JSX and it's how we are supposed to work in React. There's only a single `index.html` in the whole React project. :)

bwplotka: yea :facepalm: Looks good then (:

pkg/strutil/merge.go
bwplotka: Yes, it would be nice to have some layer of conversion on UI side. The reason is that we will not remember to put everywhere allocated slice as it's not something you do in Go, because you just kind of waste CPU / mem. (: So would be nice to handle that on UI layer globally. WDYT?

prmsrswt: Okay, got it. I'll add a null check on the UI side. But as we are using almost the same code for the graph page as upstream, I think it's not a great idea to significantly diverge from the code. Maybe we can upstream this null check too. Also I guess it's fine to go with
```go
if names == nil {
	names = make([]string, 0)
}
```
in `api.go`, this way we won't have an allocated slice for any other consumer of `MergeSlice` function, but only in `labelNames` and that too only when we need it.

GiedriusS: @bwplotka the convention is to return empty arrays in such cases: https://stackoverflow.com/questions/6045895/should-a-restful-api-return-404-for-arrays-of-objects, https://stackoverflow.com/questions/13366730/proper-rest-response-for-empty-table, https://groups.google.com/forum/#!topic/api-craft/wngl_ZKONyk, etc.

GiedriusS: Ah, I see, this was added to our utility function, that's why you had commented. My bad.

pkg/ui/react-app/src/thanos/pages/errorBoundary/ErrorBoundary.tsx
```diff
@@ -0,0 +1,50 @@
...
+          <h3>
+            Please file an issue at&nbsp;
```
squat, line 31: Tiny grammar nit:
Please file an issue in the&nbsp;

pkg/ui/react-app/src/thanos/pages/errorBoundary/ErrorBoundary.tsx
```diff
@@ -0,0 +1,50 @@
...
+            <a href="https://github.com/thanos-io/thanos/issues" target="_blank" rel="noreferrer noopener">
+              Thanos issue tracker.
+            </a>
+          </h3>
+          <Button color="link" id="toggler" className={styles.detailsBtn} style={{ fontSize: '1.2em' }}>
```
squat, line 36: If this is the only instance of the detailsBtn class, why split the font size out of the css? To keep the css class more generic?

prmsrswt: The internal bootstrap classes were getting priority over the `detailsBtn` class, so I wasn't able to modify the font-size here. I had two options, either use `!important` on use inline CSS.

I think maybe there is a third solution too. I can increase the specificity of `detailsBtn` class, by declaring it as `button.detailsBtn` or something.

prmsrswt: Using `button.detailsBtn` instead of just `.detailsBtn` worked :tada:

squat: yes let's avoid !important for sure. making it more specialized :+1:

squat:
looks great. See,s that we need to fix a linting error. This may require regeneration of the bindata.go

Issue #2753, opened 2020-06-10
Sidecar and Compact: Out of order block uploads lead to overlaps and Compact critical errors


gburek-fastly:
**Thanos, Prometheus and Golang version used**:

thanos, version 0.12.2 (branch: HEAD, revision: 52e10c6e0f644ea98fd057e7fbece828d8dd07c7)
  build user:       root@c1a6cf60f03d
  build date:       20200430-16:24:03
  go version:       go1.13.6

prometheus, version 2.18.1 (branch: HEAD, revision: ecee9c8abfd118f139014cb1b174b08db3f342cf)
  build user:       root@2117a9e64a7e
  build date:       20200507-16:51:47
  go version:       go1.14.2

**Object Storage Provider**:

Google Cloud Storage

**What happened**:

Compact ran into an overlapped block error:

```
Jun 05 08:09:29 caller=compact.go:460 msg="critical error detected; halting" err="compaction: group 0@15585302070356120162: pre compaction overlap check: overlaps found while gathering blocks. [mint: 1591185600000, maxt: 1591192800000, range: 2h0m0s, blocks: 2]: <ulid: 01EA1R7GMDZP9XBPQBYBJPES3W, mint: 1591171200000, maxt: 1591200000000, range: 8h0m0s>, <ulid: 01E9XB4ECJ3NPQ08PBS7XHAXZ6, mint: 1591185600000, maxt: 1591192800000, range: 2h0m0s>"
```

Looking at the source system, it appears that it was was isolated from the internet due to an upstream network failure for several days. Once the network was accessible, it began uploading blocks, but failed on 01E9XB4ECJ3NPQ08PBS7XHAXZ6 at 07:52:19. That block was later uploaded at 08:05:51:

```
Jun 03 11:00:12 caller=shipper.go:347 msg="upload new block" id=01E9WXCZWJA9TXDEY0JJ6KACNR
Jun 03 13:00:41 caller=sidecar.go:340 err="iter local block metas: check exists: Get https://storage.googleapis.com/storage/v1/b/<bucket>/o/01E9X48Q4J2X91Y6WPDW6ST3T7%2Fmeta.json?alt=json&prettyPrint=false&projection=full: oauth2: cannot fetch token: Post https://oauth2.googleapis.com/token: dial tcp: i/o timeout" uploaded=0
...
[...]

PR #2762, opened 2020-06-15
Cut 0.13.0 release.

bwplotka:


bwplotka:
NOTE: I still investigate https://github.com/thanos-io/thanos/issues/2753 if we can fix it quickly in 0.13

bwplotka:
Found root cause of https://github.com/thanos-io/thanos/issues/2753

Releasing without fix for now as it's not 0.13 regression

bwplotka:
Let's actually fix https://github.com/thanos-io/thanos/issues/2753 before release

PR #2765, opened 2020-06-15
shipper: Be strict about upload order unless it's specified so & cut v0.13.0-rc.2

bwplotka:
This is actually a quite real case for potential overlaps in Thanos system, so fixing before 0.13. 

Fixes https://github.com/thanos-io/thanos/issues/2753

Thanks @gburek-fastly for all pointers, it helped us to narrow this down :muscle: 


bwplotka:
I mentioned this in flag help. Why I created an option to still enable it:

* In case if you don't have the persistent disk you kind of wants to upload as soon as possible. Also this allows you to have block with malformed metadata and it will be just ignored if you set this flag to true (well ignored + error log, metric). Also with `upload-compacted` you might want best effort upload as soon as possible.
* For all those releases we were allowing out of order uploads, so someone might have rely on it (?)

Those arguments are not strong I agree, I am happy to reconsider this... Maybe moving this flag to `Hidden` for now? 

Also technically we should do rc.3 with this not full release. Thoughts?

CHANGELOG.md
```diff
@@ -26,6 +26,7 @@ We use *breaking* word for marking changes that are not backward compatible (rel
 - [#2416](https://github.com/thanos-io/thanos/pull/2416) Bucket: Fixed issue #2416 bug in `inspect --sort-by` doesn't work correctly in all cases.
 - [#2719](https://github.com/thanos-io/thanos/pull/2719) Query: `irate` and `resets` use now counter downsampling aggregations.
 - [#2705](https://github.com/thanos-io/thanos/pull/2705) minio-go: Added support for `af-south-1` and `eu-south-1` regions.
+- [#2753](https://github.com/thanos-io/thanos/issues/2753) Sidecar,Receive,Rule: Fixed cause for compactor overlapping blocks in upload error cases.
```
squat, line 29: nit on comment formatting: we should have spaces between component names

cmd/thanos/receive.go
```diff
@@ -89,6 +89,12 @@ func registerReceive(m map[string]setupFunc, app *kingpin.Application) {
 
 	walCompression := cmd.Flag("tsdb.wal-compression", "Compress the tsdb WAL.").Default("true").Bool()
 
+	allowOutOfOrderUpload := cmd.Flag("shipper.allow-out-of-order-uploads",
+		"If true shipper will skip failed block uploads in given iteration and retry later. This means that some newer blocks might uploaded sooner than older."+
```
squat, line 93: Let's keep these as full sentences to make it more readable. 
```suggestion
		"If true, shipper will skip failed block uploads in the given iteration and retry later. This means that some newer blocks might be uploaded sooner than older blocks."+
```

cmd/thanos/receive.go
```diff
@@ -89,6 +89,12 @@ func registerReceive(m map[string]setupFunc, app *kingpin.Application) {
...
+			"This will trigger compaction without those blocks, as a resulted create 'valid overlap situation'. Set it to true if you have vertical compaction enabled and wish to upload blocks as soon as possible without caring"+
```
squat, line 94:
```suggestion
			"This will trigger compaction without those blocks and as a result will create a 'valid overlap situation'. Set it to true if you have vertical compaction enabled and wish to upload blocks as soon as possible without caring"+
```

squat: just for clarification, what is a 'valid overlap situation'?

docs/operating/troubleshooting.md
```diff
@@ -29,13 +28,15 @@ Checking producers log for such ULID, and checking meta.json (e.g if sample stat
 
 ### Reasons
 
+- You are running Thanos (sidecar, ruler or receive) older than 0.13.0. During transient upload errors there was possibility to have overlaps caused by compactor not being aware of all blocks See: [this](https://github.com/thanos-io/thanos/issues/2753)
```
squat, line 31:
```suggestion
- You are running Thanos (sidecar, ruler or receive) older than 0.13.0. During transient upload errors there is a possibility to have overlaps caused by the compactor not being aware of all blocks See: [this](https://github.com/thanos-io/thanos/issues/2753)
```

pkg/shipper/shipper.go
```diff
@@ -272,70 +278,76 @@ func (s *Shipper) Sync(ctx context.Context) (uploaded int, err error) {
...
 				level.Error(s.logger).Log("msg", "found overlap or error during sync, cannot upload compacted block", "err", err)
 				uploadErrs++
-				return nil
+				continue
 			}
 		}
 
 		if err := s.upload(ctx, m); err != nil {
-			level.Error(s.logger).Log("msg", "shipping failed", "block", m.ULID, "err", err)
+			if !s.allowOutOfOrderUploads {
```
squat, line 326: :+1:

pkg/shipper/shipper.go
```diff
@@ -378,15 +390,12 @@ func (s *Shipper) upload(ctx context.Context, meta *metadata.Meta) error {
...
 
-// iterBlockMetas calls f with the block meta for each block found in dir
-// sorted by minTime asc. It logs an error and continues if it cannot access a
-// meta.json file.
-// If f returns an error, the function returns with the same error.
-func (s *Shipper) iterBlockMetas(f func(m *metadata.Meta) error) error {
-	var metas []*metadata.Meta
+// blockMetasFromOldest returns the block meta of each block found in dir
+// sorted by minTime asc.
+func (s *Shipper) blockMetasFromOldest() (metas []*metadata.Meta, _ error) {
```
squat, line 395: nice, I think this is a good simplification

brancz:
> Also technically we should do rc.3 with this not full release. Thoughts?

Yes I think rc.3 is the right thing to do here.

I think a hidden flag for now sounds good.

cmd/thanos/receive.go
bwplotka: I changed to just `an overlap situation`

PR #2767, opened 2020-06-15
pkg/rules/manager: expose original file field in groups

s-urbaniak:
Currently, the original file field is a private field.
This causes it to render in the Thanos Rules UI.
This fixes it by making that field public.

## Changes

- Exposed original file field in `pkg/rules.Group` struct.

## Verification

Screenshot before this change shows that Rules are not being rendered:

image

Additionally the following warn error message is rendered in log files:

```
level=warn ts=2020-06-15T11:05:16.146463886Z caller=ui.go:152 component=rules msg="template expansion failed" err="template: :82:26: executing \"content\" at <.OriginalFile>: can't evaluate field OriginalFile in type rules.Group"
```

Screenshot with this change applied:

image

/cc @bwplotka @brancz

### DIFF
--- a/pkg/rules/manager.go
+++ b/pkg/rules/manager.go
@@ -33,14 +33,14 @@ const tmpRuleDir = ".tmp-rules"
 
 type Group struct {
 	*rules.Group
-	originalFile            string
+	OriginalFile            string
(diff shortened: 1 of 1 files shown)


squat:
thanks :)

PR #2768, opened 2020-06-15
Treating GCS EOF failures in the read path as IsObjNotFoundErr w/ logging

jrepp:
Upon startup of thanos store on GCS it's common to see read failures on the deletion mark:

caller=grpc.go:137 service=gRPC/server component=store msg="internal server shutdown"
    err="bucket store initial sync: sync block: filter metas: get file: ***/deletion-mark.json: Get ***/deletion-mark.json: unexpected EOF"
caller=main.go:212
    err="Get ***/deletion-mark.json: unexpected EOF get file: ***/deletion-mark.json
thanos/pkg/block/metadata.ReadDeletionMark
thanos/pkg/block/metadata/deletionmark.go:56
thanos/pkg/block.(*IgnoreDeletionMarkFilter).Filter

These failures cause crash loops in our GKE clusters where storage will commonly need 20+ restarts before becoming available.

After this change we still see intermittent transient failures on startup (mostly due to rate limiting) but typically no more than 2.

Now treat eof (wrapped) as a not-found-error but log the stack trace for analysis

This is the underlying issue tracking the eof error in GCS:

https://github.com/googleapis/google-cloud-go/issues/1832

### DIFF
--- a/pkg/objstore/gcs/gcs.go
+++ b/pkg/objstore/gcs/gcs.go
@@ -167,7 +167,11 @@ func (b *Bucket) Delete(ctx context.Context, name string) error {
 
 // IsObjNotFoundErr returns true if error means that object is not found. Relevant to Get operations.
 func (b *Bucket) IsObjNotFoundErr(err error) bool {
-	return err == storage.ErrObjectNotExist
+	eof :=  errors.Is(err, io.ErrUnexpectedEOF)
+	if eof {
+		b.logger.Log("method", "IsObjNotFoundErr", "err", err, "stack", fmt.Sprintf("%+v", errors.WithStack(err)))
+	}
+	return err == storage.ErrObjectNotExist || eof
 }
 
 func (b *Bucket) Close() error {


bwplotka:
Ok, I totally get your PR as workaround for bug in GCS. This is however should be handled like workaround, so:

* There should be clear comment stating so with TODO and link to issue tracking this
* Should be some .Hidden flag controlling this behaviour and turned off by default.  So we can control this without compiling. 

Alternatively, you can change it quickly on some fork, but it makes sense to have this workaround for everyone :hugs: 

What do you think?

jrepp:
> Ok, I totally get your PR as workaround for bug in GCS. And it makes sense, thanks for this. This is however should be handled like workaround, so:
> 
> * There should be clear comment stating so with TODO and link to issue tracking this
> * Should be some .Hidden flag controlling this behaviour and turned off by default.  So we can control this without compiling.
> 
> Alternatively, you can change it quickly on some fork, but it makes sense to have this workaround for everyone 🤗
> 
> What do you think?

Those are great ideas. I'll make those changes. I'll need to look for some other .Hidden flags to reference.

On my fork I also tried the latest google cloud and storage APIs. They still haven't sorted out the eof issue but some additional connection reset issues that we were seeing are gone on latest.

pkg/objstore/gcs/gcs.go
```diff
@@ -167,7 +167,11 @@ func (b *Bucket) Delete(ctx context.Context, name string) error {
 
 // IsObjNotFoundErr returns true if error means that object is not found. Relevant to Get operations.
 func (b *Bucket) IsObjNotFoundErr(err error) bool {
-	return err == storage.ErrObjectNotExist
+	eof :=  errors.Is(err, io.ErrUnexpectedEOF)
```
pracucci, line 170: I don't think it's safe to assume that a `ErrUnexpectedEOF` always mean "object doesn't exit". I think you could get an unexpected EOF even if the connection is truncated while downloading the content or maybe even other scenarios.

jrepp: Yes, I agree - and I think the google developers realize the same thing. If anything EOF should is a retriable error. I am open to other suggestions on how to get the thanks objstore code to run reliably on GCS. At the moment this work-around is getting our systems up and running at scale.

What is the level of testing that is available for different storage solutions? I saw there are some e2e tests. Does the Thanos project have environments for doing comprehensive objstore tests and collecting measurements for issues like this?

PR #2773, opened 2020-06-16
receive: Improve receive error logging

kakkoyun:

## Changes

* Add request-id middleware and use it for remote writes.
* Add tenant context to logs.
* Add missing endpoint to logs.

## Verification

* `make test-local`

### DIFF
--- a/pkg/server/http/middleware/request_id.go
+++ b/pkg/server/http/middleware/request_id.go
@@ -1,0 +1,42 @@
+// Copyright (c) The Thanos Authors.
+// Licensed under the Apache License 2.0.
+
+package middleware
+
+import (
+	"context"
+	"math/rand"
+	"net/http"
+	"time"
+
+	"github.com/oklog/ulid"
+)
+
+type ctxKey int
+
+const reqIDKey = ctxKey(0)
+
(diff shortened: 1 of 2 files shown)


pkg/server/http/middleware/request_id.go
```diff
@@ -0,0 +1,42 @@
+// Copyright (c) The Thanos Authors.
+// Licensed under the Apache License 2.0.
+
+package middleware
```
bwplotka, line 4: cc @yashrsharma44

yashrsharma44: Looks great, however, do we want to correlate the trace-id with request-id? Maybe my idea would be suitable for another PR for correlating them.
Otherwise looks great :100:

kakkoyun: We can try correlate or use a single id for both. We need to check how actually different middleware interacts and what happens if tracing is not enabled. For the scope of this PR, I want to keep track of execution of async forward requests in the scope a single request.

brancz: Aligning everything with the W3C trace context would probably work, but doesn't block this I would say.

PR #2775, opened 2020-06-17
Add memcached server address to Set() error

pracucci:

## Changes

Today I was investigating the error `memcache: unexpected response line from \"set\": \"SERVER_ERROR Out of memory during read\\r\\n\` in Cortex and I had some hard time figuring out which was the memcached server.

In this PR I propose to inject the server address to the Set() error.

Note: the same thing is difficult to do with the GetMulti() because we may potentially touch many (sometimes all) memcached server during a multi get request.

## Verification

Existing tests.

### DIFF
--- a/pkg/cacheutil/memcached_client.go
+++ b/pkg/cacheutil/memcached_client.go
@@ -5,6 +5,7 @@ package cacheutil
 
 import (
 	"context"
+	"net"
 	"strings"
 	"sync"
 	"time"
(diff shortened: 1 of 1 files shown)


pkg/cacheutil/memcached_client.go
```diff
@@ -463,3 +464,8 @@ func (c *memcachedClient) resolveAddrs() error {
 
 	return c.selector.SetServers(servers...)
 }
+
+func (c *memcachedClient) getServerAddrByKey(key string) net.Addr {
+	addr, _ := c.selector.PickServer(key)
```
squat, line 469: could we be strict check the error? Or at least document why we are 100% confident the error does not need checking?

povilasv: IMO if this fails we just log empty addr, which is fine, no?

pracucci: > IMO if this fails we just log empty addr, which is fine, no?

Correct. That was my intention. What's your take @squat ?

squat: that's OK but nothing is preventing anyone from using this internally and getting surprised later when something doesn't work and there's no error. Any reason this really needs to be a new func vs simply inlined above 291? Otherwise I would suggest to add this exact thing as a comment :)

pracucci: Right. I removed the function. What's about now?

PR #2779, opened 2020-06-18
ui: Fix external prefix not having a trailing slash

prmsrswt:
There was a regression introduced in #2714 where we expected the `externalPrefix` to have a trailing slash but actually missed adding it there. This PR fixes that bug.

## Changes

Add a trailing slash to `externalPrefix`. Also, sanitize `externalPrefix` as well.

## Verification
Tested locally with multiple combinations of `web.route-prefix` and `web.external-prefix`


bwplotka:
Thanks for helping :heart:

Issue #1598, opened 2019-10-02
Delete series for object storage.


bwplotka:
**Goals:**
* Allow Delete series from blocks kept in an object storage

**Use cases:**
* GDPR
* Delete corrupted / cardinality bomb series
* Series based retention! (e.g aggregates: https://github.com/thanos-io/thanos/issues/903)

**Related:**
* https://github.com/thanos-io/thanos/issues/415 - /delete_series for sidecar
* https://github.com/thanos-io/thanos/issues/903 - aggr based retention

It's a dry issue, design to come. (:

bwplotka:
Challenges:
* How to handle long-running job for that? Should it be compactor? How to coordinate with the compactor if not? How to measure progress?
* Is it even worth to add tombstones logic? 
  * For GDPR we care of data complete removal, not just unvisible
  * For aggr retention as well - we care about data size. Tombstones will only slow all down and add size.

SuperQ:
My opinion:
* The compactor seems like the right place for this.
* I think following the tombstones logic is fine. It works well for Prometheus itself.
* Adding the `clean_tombstones` endpoint/function is probably the right way to deal with purging if necessary.

bwplotka:
Yup @SuperQ 

As we talked offline tombstones are probably no go, as they are not immutable. At least not in the form that is implemented in Prometheus. We might be though able to add deletion requests to each block, then having a compactor to include that during compaction and lazy rewrite. It would be nice to have some solution that will also fit https://github.com/thanos-io/thanos/issues/903

bwplotka:
Still valid. Help wanted.

SidharthBansal:
Hello, I am looking forward to contributing to this for GSoC 2020. How shall I start?

infa-kparida:
This will be a super useful feature for decommissioning  prometheus clusters.Looking forward to it +1

bwplotka:
Good point! 

Looks like we did not find anyone for GSoC, we will try with Summer Community Bridge. (:

SidharthBansal:
What is the timeline for `Summer Community Bridge` and what is the link where I can apply?

bwplotka:
@idvoretskyi might know more.

idvoretskyi:
@bwplotka we'll announce this later in April/early May.

SidharthBansal:
Great thanks

idvoretskyi:
[...]

PR #2780, opened 2020-06-18
Proposal: Delete series for object storage

Harshitha1234:

## Changes

Design proposal for delete series #1598

### DIFF
--- a/docs/proposals/deletions_object_storage.md
+++ b/docs/proposals/deletions_object_storage.md
@@ -41,7 +41,7 @@ The main motivation for considering deletions in the object storage are the foll
 *   **Perspectives to deal with Compaction of blocks having tombstones:**
     *   **Block with tombstones** Have a threshold to perform deletions on the compacted blocks ([In Prometheus](https://github.com/prometheus/prometheus/blob/f0a439bfc5d1f49cec113ee9202993be4b002b1b/tsdb/compact.go#L213), the blocks with big enough time range, that have >5% tombstones, are considered for compaction.) We solve the tombstones, if the tombstones are greater than than the threshold and then perform compaction. If not we attach the tombstone file to the new block. If multiple blocks are being compacted, we merge the tombstone files of the blocks whose threshold is not met.
     *   **Block with deletion-mark.json i.e., entire block marked for deletion:** Returns an error message as the entire block is going to be deleted.
-*   **Perspectives to deal with Downsampling of blocks having tombstones:** 
+*   **Perspectives to deal with Downsampling of blocks having tombstones:**
     *   **Block with tombstones:** If the tombstones are less than the threshold we copy the tombstone file and attach it to the new downsampled block else we solve the tombstones and downsample the block. And the downsampled block with tombstones during its next compaction would again have the same cases as with the compaction of a block with tombstones.
     *   **Blocks without tombstones:** Downsampling happens...
 
--- a/docs/proposals/deletions_object_storage.md
+++ b/docs/proposals/deletions_object_storage.md
@@ -1,0 +1,83 @@
+# Delete series for object storage
+
+
+**Date:**	&lt;2020-06-17>
+**Type:**   Proposal
+**Menu:**   Proposals
+**Status:** Draft
+**Authors:** 	&lt;Harshitha Chowdary Thota, Matthias Loibl, Bartek Plotka>
+**Tickets:**
+1. [https://github.com/thanos-io/thanos/issues/1598](https://github.com/thanos-io/thanos/issues/1598) (main)
+2. [https://github.com/thanos-io/thanos/issues/903](https://github.com/thanos-io/thanos/issues/903)
+
+
+# Summary
+
+This design document proposes deletion of series for object storage in Thanos. This feature mainly causes changes in the Compactor and Store components and the object storage itself where the changes are expected to be reflected.
+
+
+# Motivation
+
+The main motivation for considering deletions in the object storage are the following use cases
+
+
+
+*   **Accidental insertion of confidential data:** This is a scenario where a user accidentally inserts confidential data and wants to delete it immediately. In this case the user expects his/her request to accelerate an immediate deletion of the series pertaining to the blocks concerned with that specific data to be deleted.
+*   **GDPR:** Masking data and eventual deletion is expected.
+*   **Deletions to sustain user requirements:** Let’s assume the user has some data which leads to some unexpected results or causes performance degradation (due to high cardinality) and the user wants to restore the previous data set-up for obtaining the desired results. In this scenario, the user would want to send a request to mask the data for the time being as there isn’t a high priority requirement to delete the data but eventually during the compaction the user can expect the data to be deleted by the compactor not leading to any major performance issues.
+*   **Helps achieving series based retention (e.g rule aggregation) [#903](https://github.com/thanos-io/thanos/issues/903).**
+
+
+## Goals
+
+
+
+*   Unblock users and allow series deletion in the object storage using tombstones.
+*   Allowing undeletes for a default time duration and extending it as per the user requirements.
+*   Dealing with already downsampled and to be downsampled blocks with tombstones.
+*   Performing all the above operations at admin level.
+
+
+# Proposed Approach
+
+
+
+*   We propose to implement deletions using the tombstones approach.
+*   A user is expected to enter the following details for performing deletions:
+    *   **external labels**
+    *   **start timestamp**
+    *   **end timestamp** (start and end timestamps of the series data the user expects to be deleted)
+    *   **maximum waiting duration for performing deletions** where a default value is considered if explicitly not specified by the user. (only after this time passes the deletions are performed by the compactor in the next compaction cycle)
+*   The details are entered via a deletions API (good to have a web UI interface) and they are processed by the compactor to create a tombstone file, if there’s a match for the details entered. Afterwards the tombstone file is uploaded to object storage making it accessible to other components.
+*   If the data with the requested details couldn’t be found in the storage an error message is returned back as a response to the request.
+*   Store Gateway masks the series on processing the tombstones from the object storage.
+*   **Perspectives to deal with Downsampling of blocks having tombstones:**
+    *   **Block with tombstones and max duration to perform deletes passed:** The compactor should first check the maximum duration to perform deletions for that block and if the proposed maxtime passed the deletions are first performed and then the downsampling occurs.
+    *   **Block with tombstones and max duration still hasn’t passed:** Perform compaction.
+    *   **Performing deletes on already compacted blocks:** Have a threshold to perform deletions on the compacted blocks (Prometheus has 5%)
+*   For undoing deletions of a time series there are two proposed ways
+    *   API to undelete a time series - maybe delete the whole tombstones file?
+    *   “Imaginary” deletion that can delete other tombstones 
+
+Considerations :
+
+
+
+*   Tombstones should be append only, so that we can solve these during compaction.
+*   We don’t want to add this feature to the sidecar. The sidecar is expected to be kept lightweight.
+
+
+## Alternatives
+
+
+
+1. External tool operating to perform deletions on the object storage.(Details still to be discussed)
+
+
+## Action Plan
+
+
+
+*   Add the deletion API (probably compactor) that only creates tombstones
+*   Store Gateway should be able to mask based on the tombstones from object storage
+*   A web UI for the deletion API?
--- a/docs/proposals/deletions_object_storage.md
+++ b/docs/proposals/deletions_object_storage.md
@@ -50,6 +50,7 @@ During the implementation phase of this proposal one of the first problems we ca
 
 ##### Proposed Alternative Approaches: 
 (1) To have a separate store like component with index cache.
+
 (2) To have a different format for the tombstones. Instead of having <seriesRef, min, max> maybe we can have <matcher, min, max>
 * **Idea 1:** Having a CLI tool for dealing with deletions.
 * **Idea 2:** To have all tombstones at some single place and they are used while performing compaction or during query time.
--- a/docs/proposals/deletions_object_storage.md
+++ b/docs/proposals/deletions_object_storage.md
@@ -41,11 +41,24 @@ The main motivation for considering deletions in the object storage are the foll
 *   **Perspectives to deal with Compaction of blocks having tombstones:**
     *   **Block with tombstones** Have a threshold to perform deletions on the compacted blocks ([In Prometheus](https://github.com/prometheus/prometheus/blob/f0a439bfc5d1f49cec113ee9202993be4b002b1b/tsdb/compact.go#L213), the blocks with big enough time range, that have >5% tombstones, are considered for compaction.) We solve the tombstones, if the tombstones are greater than than the threshold and then perform compaction. If not we attach the tombstone file to the new block. If multiple blocks are being compacted, we merge the tombstone files of the blocks whose threshold is not met.
     *   **Block with deletion-mark.json i.e., entire block marked for deletion:** Returns an error message as the entire block is going to be deleted.
-*   **Perspectives to deal with Downsampling of blocks having tombstones:**
+*   **Perspectives to deal with Downsampling of blocks:**
     *   **Block with tombstones:** If the tombstones are less than the threshold we copy the tombstone file and attach it to the new downsampled block else we solve the tombstones and downsample the block. And the downsampled block with tombstones during its next compaction would again have the same cases as with the compaction of a block with tombstones.
     *   **Blocks without tombstones:** Downsampling happens...
 
-Considerations :
+##### Problems during implementations and possible ideas to solve :
+During the implementation phase of this proposal one of the first problems we came across was that we had to pull the index of all the blocks for creating or appending a tombstone.
+
+##### Proposed Alternative Approaches: 
+(1) To have a separate store like component with index cache.
+(2) To have a different format for the tombstones. Instead of having <seriesRef, min, max> maybe we can have <matcher, min, max>
+* **Idea 1:** Having a CLI tool for dealing with deletions.
+* **Idea 2:** To have all tombstones at some single place and they are used while performing compaction or during query time.
+* **Idea 3:** If we want to have an API for deletions, one way is to have the Store API configured in such a way that the compactor can use it to check for a match.
+*Edge case:* Do we rebuild all the rows of a tombstone? Because we need to be aware that the tombstone is being included(or not) in the given Store API call.
+
+*We're starting with (2) Idea 2*
+
+#### Considerations :
 
 *   Tombstones should be append only, so that we can solve tombstones and rewrite the blocks by performing changes. The old block and the tombstones are deleted during compaction.
 *   We don’t want to add this feature to the sidecar. The sidecar is expected to be kept lightweight.
--- a/docs/proposals/deletions_object_storage.md
+++ b/docs/proposals/deletions_object_storage.md
@@ -9,7 +9,7 @@ owner: bwplotka, Harshitha1234, metalmatze
 ### Ticket: https://github.com/thanos-io/thanos/issues/1598
 ## Summary
 
-This design document proposes deletion of series for object storage in Thanos. This feature mainly causes changes in the Compactor and Store components and the object storage itself where the changes are expected to be reflected.
+This design document proposes deletion of series for object storage in Thanos. This feature mainly causes changes in the Store component and masks the series from objstore corresponding to the user deletion requests and does not trigger actual deletions in the object storage which is proposed to be the future work extension of the current proposal.
 
 ## Motivation
 
@@ -23,7 +23,6 @@ The main motivation for considering deletions in the object storage are the foll
 ## Goals
 
 *   Unblock users and allow series deletion in the object storage using tombstones.
-*   Deal with compaction and downsampling of blocks with tombstones.
 *   Performing deletions at admin level.
 
 ## Proposed Approach
@@ -55,7 +54,6 @@ The main motivation for considering deletions in the object storage are the foll
 *   The entered details are processed by the CLI tool to create a tombstone file (unique for a request and irrespective of the presence of series), and the file is uploaded to the object storage making it accessible to all components.
 *   **Filename optimization**: The filename is created from the hash of matchers, minTime and maxTime. This helps re-write an existing tombstone, whenever a same request is made in the future hence avoiding duplication of the same request. (NOTE: Requests which entail common deletions still creates different tombstones.)
 *   Store Gateway masks the series on processing the global tombstone files from the object storage. At chunk level, whenever there's a match with the data corresponding to atleast one of the tombstones, we skip the chunk, potentially resulting in the masking of chunk.
-*   During compaction or downsampling, we check if the blocks being considered have series corresponding to a tombstone file, if so we delete the data and continue with the process.
 
 ## Considerations
 
@@ -94,7 +92,6 @@ Reasons for not choosing 4 i.e., **Custom format global - appendable file**:
 
 *   Add a CLI tool which creates tombstones
 *   Store Gateway should be able to mask based on the tombstones from object storage
-*   Compactor solves the tombstones as per the proposed approach
 
 ## Challenges
 
@@ -104,6 +101,7 @@ Reasons for not choosing 4 i.e., **Custom format global - appendable file**:
 
 ## Future Work
 
+*   Performing actual deletions in the object storage and rewriting the blocks based on tombstones by the compactor.
 *   Have a max waiting duration feature for performing deletions where a default value is considered if explicitly not specified by the user. (only after this time passes the deletions are performed by the compactor in the next compaction cycle)
 *   Have the undoing deletions feature and there are two proposed ways
     *   API to undelete a time series - maybe delete the whole tombstones file?
--- a/docs/proposals/deletions_object_storage.md
+++ b/docs/proposals/deletions_object_storage.md
@@ -2,7 +2,7 @@
 title: Delete series for object storage
 type: proposal
 menu: proposals
-status: draft
+status: approved
 owner: bwplotka, Harshitha1234, metalmatze
 ---
 
@@ -23,9 +23,8 @@ The main motivation for considering deletions in the object storage are the foll
 ## Goals
 
 *   Unblock users and allow series deletion in the object storage using tombstones.
-*   Allowing undeletes for a default time duration and extending it as per the user requirements.
-*   Dealing with already downsampled and to be downsampled blocks with tombstones.
-*   Performing all the above operations at admin level.
+*   Deal with compaction and downsampling of blocks with tombstones.
+*   Performing deletions at admin level.
 
 ## Proposed Approach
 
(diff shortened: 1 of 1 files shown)


Harshitha1234:
cc @metalmatze @bwplotka :)

pkg/cacheutil/memcached_client.go
```diff
@@ -269,7 +269,7 @@ func (c *memcachedClient) Stop() {
 	c.workers.Wait()
 }
 
-func (c *memcachedClient) SetAsync(ctx context.Context, key string, value []byte, ttl time.Duration) error {
+func (c *memcachedClient) SetAsync(_ context.Context, key string, value []byte, ttl time.Duration) error {
```
GiedriusS, line 272: Seems like the changes in this file are unrelated? :eye:

docs/proposals/deletions_object_storage.md
```diff
@@ -0,0 +1,86 @@
...
+*   The details are entered via a deletions API (good to have a web UI interface) and they are processed by the compactor to create a tombstone file, if there’s a match for the details entered. Afterwards the tombstone file is uploaded to object storage making it accessible to other components.
+*   If the data with the requested details couldn’t be found in the storage an error message is returned back as a response to the request.
+*   Store Gateway masks the series on processing the tombstones from the object storage.
+*   **Perspectives to deal with Downsampling of blocks having tombstones:**
+    *   **Block with tombstones and max duration to perform deletes passed:** The compactor should first check the maximum duration to perform deletions for that block and if the proposed maxtime passed the deletions are first performed and then the downsampling occurs.
+    *   **Block with tombstones and max duration still hasn’t passed:** Perform compaction.
+    *   **Performing deletes on already compacted blocks:** Have a threshold to perform deletions on the compacted blocks (Prometheus has 5%)
```
GiedriusS, line 60: Could we elaborate more on what is this threshold?

docs/proposals/deletions_object_storage.md
```diff
@@ -0,0 +1,86 @@
...
+
+*   Tombstones should be append only, so that we can solve these during compaction.
+*   We don’t want to add this feature to the sidecar. The sidecar is expected to be kept lightweight.
+
+
+## Alternatives
+
+
+
+1. External tool operating to perform deletions on the object storage.(Details still to be discussed)
```
GiedriusS, line 77: It's probably unrealistic to have this since the tool would have to coordinate its actions with potentially more than 1 compactor running on a bucket. IMHO it's not worth considering this as a feasible option.

docs/proposals/deletions_object_storage.md
```diff
@@ -0,0 +1,86 @@
...
+*   We propose to implement deletions using the tombstones approach.
+*   A user is expected to enter the following details for performing deletions:
+    *   **external labels**
+    *   **start timestamp**
+    *   **end timestamp** (start and end timestamps of the series data the user expects to be deleted)
+    *   **maximum waiting duration for performing deletions** where a default value is considered if explicitly not specified by the user. (only after this time passes the deletions are performed by the compactor in the next compaction cycle)
+*   The details are entered via a deletions API (good to have a web UI interface) and they are processed by the compactor to create a tombstone file, if there’s a match for the details entered. Afterwards the tombstone file is uploaded to object storage making it accessible to other components.
+*   If the data with the requested details couldn’t be found in the storage an error message is returned back as a response to the request.
+*   Store Gateway masks the series on processing the tombstones from the object storage.
+*   **Perspectives to deal with Downsampling of blocks having tombstones:**
```
GiedriusS, line 57: I guess it's also worth noting that we shouldn't consider blocks for the deletion API operations that have a `deletion-mark.json` i.e. about to be completely deleted themselves.

docs/proposals/deletions_object_storage.md
```diff
@@ -0,0 +1,86 @@
...
+    *   **Block with tombstones and max duration to perform deletes passed:** The compactor should first check the maximum duration to perform deletions for that block and if the proposed maxtime passed the deletions are first performed and then the downsampling occurs.
+    *   **Block with tombstones and max duration still hasn’t passed:** Perform compaction.
+    *   **Performing deletes on already compacted blocks:** Have a threshold to perform deletions on the compacted blocks (Prometheus has 5%)
+*   For undoing deletions of a time series there are two proposed ways
+    *   API to undelete a time series - maybe delete the whole tombstones file?
```
GiedriusS, line 62: What if a deletion happens during a compaction pass? In that case, we most likely should return an error to the caller, right?

docs/proposals/deletions_object_storage.md
```diff
@@ -0,0 +1,86 @@
...
+    *   “Imaginary” deletion that can delete other tombstones 
+
+Considerations :
+
+
+
+*   Tombstones should be append only, so that we can solve these during compaction.
```
GiedriusS, line 69: `Solve these` ... what is "these"? :P

docs/proposals/deletions_object_storage.md
```diff
@@ -0,0 +1,86 @@
...
+*   We don’t want to add this feature to the sidecar. The sidecar is expected to be kept lightweight.
```
GiedriusS, line 70: :+1: concur with this

docs/proposals/deletions_object_storage.md
```diff
@@ -0,0 +1,86 @@
...
+This design document proposes deletion of series for object storage in Thanos. This feature mainly causes changes in the Compactor and Store components and the object storage itself where the changes are expected to be reflected.
+
+
+# Motivation
+
+The main motivation for considering deletions in the object storage are the following use cases
+
+
+
+*   **Accidental insertion of confidential data:** This is a scenario where a user accidentally inserts confidential data and wants to delete it immediately. In this case the user expects his/her request to accelerate an immediate deletion of the series pertaining to the blocks concerned with that specific data to be deleted.
```
GiedriusS, line 28: We can use gender-neutral language here plus there's a missing comma:

```suggestion
*   **Accidental insertion of confidential data:** This is a scenario where a user accidentally inserts confidential data and wants to delete it immediately. In this case, the user expects their request to accelerate an immediate deletion of the series pertaining to the blocks concerned with that specific data to be deleted.
```

docs/proposals/deletions_object_storage.md
Harshitha1234: Yeah! Definitely that's a good point

docs/proposals/deletions_object_storage.md
```diff
@@ -0,0 +1,86 @@
...
+*   Performing all the above operations at admin level.
+
+
+# Proposed Approach
+
+
+
+*   We propose to implement deletions using the tombstones approach.
+*   A user is expected to enter the following details for performing deletions:
+    *   **external labels**
```
pracucci, line 50: Why external labels and not metric labels?

docs/proposals/deletions_object_storage.md
```diff
@@ -0,0 +1,86 @@
...
+    *   **start timestamp**
+    *   **end timestamp** (start and end timestamps of the series data the user expects to be deleted)
+    *   **maximum waiting duration for performing deletions** where a default value is considered if explicitly not specified by the user. (only after this time passes the deletions are performed by the compactor in the next compaction cycle)
+*   The details are entered via a deletions API (good to have a web UI interface) and they are processed by the compactor to create a tombstone file, if there’s a match for the details entered. Afterwards the tombstone file is uploaded to object storage making it accessible to other components.
```
pracucci, line 54: > The details are entered via a deletions API (good to have a web UI interface) and they are processed by the compactor

This means the API is exposed by the compactor. I personally see the compactor more an "off-the-grid" component and I'm not sure that design-wise it's good that it will expose this API.

docs/proposals/deletions_object_storage.md
```diff
@@ -0,0 +1,86 @@
...
+*   If the data with the requested details couldn’t be found in the storage an error message is returned back as a response to the request.
```
pracucci, line 55: > If the data with the requested details couldn’t be found in the storage an error message is returned back as a response to the request.

This means the deletion API request is synchronous. The problem is that on large setups, processing this request may take much time and the request could timeout before done. I think an async approach would be more robust:

1. Deletion API returns a request ID
2. A second API allows to GET the status of a request ID

docs/proposals/deletions_object_storage.md
```diff
@@ -0,0 +1,86 @@
...
+1. External tool operating to perform deletions on the object storage.(Details still to be discussed)
+
+
+## Action Plan
+
+
+
+*   Add the deletion API (probably compactor) that only creates tombstones
+*   Store Gateway should be able to mask based on the tombstones from object storage
+*   A web UI for the deletion API?
```
pracucci, line 86: I would keep this proposal smaller and remove the UI part form this proposal.

pkg/cacheutil/memcached_client.go
pracucci: This was done by me in the PR #2775. I think this proposal branch has been created from that PR branch or something like this.

docs/proposals/deletions_object_storage.md
metalmatze: This threshold is something that Prometheus does as well. We need to check for the exact value and maybe should want to link here?

docs/proposals/deletions_object_storage.md
metalmatze: You think we should remove it from the proposal entirely?

docs/proposals/deletions_object_storage.md
```diff
@@ -0,0 +1,86 @@
+# Delete series for object storage
+
+**Date:**	&lt;2020-06-17>
```
bwplotka, line 3: It has to have certain format, please other proposals, something like:

```
---
title: Thanos Remote Write
type: proposal
menu: proposals
status: accepted
owner: brancz
---
```

docs/proposals/deletions_object_storage.md
metalmatze: The compactor is one of the very few components (next to Sidecar, Receiver, Ruler) that has write capabilities to object storage. At the same time it is only ever running once per object storage, so that we have a single writer to the deletion tombstones. 
Additionally, by design, we thought that having the compactor "off-the-grid" and thus only accessible by an administrator solves a few permission problems right away.

We'll add this point to the proposal.

docs/proposals/deletions_object_storage.md
metalmatze: We thought that the creation of tombstones is synchronous and should be able to respond fairly quickly. The deletion itself of times series is still asynchronous and will eventually happen.
The API has an option to list all tombstones and through that get their status.

Because deletion is asynchronous and the creation of tombstones synchronous but fairly quick, we don't think the overhead of a asynchronous approach for the tombstones creation is justified.

docs/proposals/deletions_object_storage.md
metalmatze: For now we want to start with the 5% that Prometheus has. 
We want to follow up on that threshold on IRC/Matrix and ask why it was chosen. In the future, we can always be smarter about this.

docs/proposals/deletions_object_storage.md
metalmatze: The compactor is continuously running compactions and return errors to the caller might result in a lot of errors all the time, which would be annoying.

For now, we are going to check how Prometheus handles this.

docs/proposals/deletions_object_storage.md
metalmatze: tombstones, will update.

docs/proposals/deletions_object_storage.md
metalmatze: Totally, this proposal in itself is quite complex and we will remove the UI from for now. Later, we will follow up on this.

Harshitha1234:
Thank you for the feedback all the points were really worth noting :muscle:

Harshitha1234:
@bwplotka @GiedriusS @metalmatze @pracucci :hugs:

docs/proposals/deletions_object_storage.md
Harshitha1234: Adding this to the Prometheus community meeting, happening next week :)

docs/proposals/deletions_object_storage.md
```diff
@@ -0,0 +1,66 @@
...
+    *   **maximum waiting duration for performing deletions** where a default value is considered if explicitly not specified by the user. (only after this time passes the deletions are performed by the compactor in the next compaction cycle)
+*   The details are entered via a deletions API and they are processed by the compactor to create a tombstone file, if there’s a match for the details entered. Afterwards the tombstone file is uploaded to object storage making it accessible to other components.
+*   **Why Compactor**? : The compactor is one of the very few components (next to Sidecar, Receiver, Ruler) that has write capabilities to object storage. At the same time it is only ever running once per object storage, so that we have a single writer to the deletion tombstones.
+*   If the data with the requested details couldn’t be found in the storage an error message is returned back as a response to the request.
+*   Store Gateway masks the series on processing the tombstones from the object storage.
+*   **Perspectives to deal with Downsampling of blocks having tombstones:**
```
bwplotka, line 43: ` **Perspectives to deal with Downsampling of blocks having tombstones:**` Unfinished?

bwplotka: Oh you mean `Compactions` maybe?

bwplotka, line 43:
```suggestion
*   **Perspectives to deal with `compaction` of blocks having tombstones:**
```

docs/proposals/deletions_object_storage.md
```diff
@@ -0,0 +1,66 @@
...
+    *   **Block with tombstones and max duration to perform deletes passed:** The compactor should first check the maximum duration to perform deletions for that block and if the proposed maxtime passed the deletions are first performed and then the downsampling occurs.
```
bwplotka, line 44:
```suggestion
    *   **Block with tombstones and max duration to perform deletes passed:** The compactor should first check the maximum duration to perform deletions for that block and if the proposed maxtime passed the deletions are first performed and then the compaction occurs.
```

docs/proposals/deletions_object_storage.md
```diff
@@ -0,0 +1,66 @@
...
+
+## Proposed Approach
+
+*   We propose to implement deletions using the tombstones approach.
+*   To start off we use the Prometheus tombstone file format.
+*   A user is expected to enter the following details for performing deletions:
+    *   **label matchers**
+    *   **start timestamp**
+    *   **end timestamp** (start and end timestamps of the series data the user expects to be deleted)
+    *   **maximum waiting duration for performing deletions** where a default value is considered if explicitly not specified by the user. (only after this time passes the deletions are performed by the compactor in the next compaction cycle)
```
bwplotka, line 38: I think I am missing what this is, can we reword?

Harshitha1234: Moved this feature to the future work :)

docs/proposals/deletions_object_storage.md
Harshitha1234: My bad! :facepalm:

Harshitha1234:
Made new changes as per the discussions in my weekly meeting with @metalmatze @bwplotka. :) 

TODOs from our meeting:
- Add more info about what we do with compactions
- What to do if compaction level is max and there are tombstones
- Postpone undeletes feature & max waiting period feature considered while performing actual deletions & separate component to later proposal
- Keep design clear what’s now what’s later.
- Downsampling : copy tombstones

docs/proposals/deletions_object_storage.md
```diff
@@ -0,0 +1,84 @@
...
+*   **Deletions to sustain user requirements:** Let’s assume the user has some data which leads to some unexpected results or causes performance degradation (due to high cardinality) and the user wants to restore the previous data set-up for obtaining the desired results. In this scenario, the user would want to send a request to mask the data for the time being as there isn’t a high priority requirement to delete the data but eventually during the compaction the user can expect the data to be deleted by the compactor not leading to any major performance issues.
+*   **Helps achieving series based retention (e.g rule aggregation) [#903](https://github.com/thanos-io/thanos/issues/903).**
+
+## Goals
+
+*   Unblock users and allow series deletion in the object storage using tombstones.
+*   Deal with compaction and downsampling of blocks with tombstones.
+*   Performing deletions at admin level.
+
+## Proposed Approach
```
pracucci, line 29: How to deal with series that have been already ingested in the receiver, but not compacted into a block uploaded to the bucket yet?

yeya24: This proposal is only about deleting series in objstore. So maybe we need to a separate series deletion API for deleting series in TSDB components (receiver, ruler), what do you think?

docs/proposals/deletions_object_storage.md
```diff
@@ -0,0 +1,84 @@
...
+*Edge case:* Do we rebuild all the rows of a tombstone? Because we need to be aware that the tombstone is being included(or not) in the given Store API call.
+
+*We're starting with (2) Idea 2*
+
+#### Considerations :
+
+*   Tombstones should be append only, so that we can solve tombstones and rewrite the blocks by performing changes. The old block and the tombstones are deleted during compaction.
+*   We don’t want to add this feature to the sidecar. The sidecar is expected to be kept lightweight.
+
+## Alternatives
```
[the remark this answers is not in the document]
Harshitha1234: Yeah.. Sure! :)

docs/proposals/deletions_object_storage.md
```diff
@@ -0,0 +1,84 @@
...
+*   **Why Compactor**? : The compactor is one of the very few components (next to Sidecar, Receiver, Ruler) that has write capabilities to object storage. At the same time it is only ever running once per object storage, so that we have a single writer to the deletion tombstones.
+*   If the data with the requested details couldn’t be found in the storage an error message is returned back as a response to the request.
+*   Store Gateway masks the series on processing the tombstones from the object storage.
+*   **Perspectives to deal with Compaction of blocks having tombstones:**
+    *   **Block with tombstones** Have a threshold to perform deletions on the compacted blocks ([In Prometheus](https://github.com/prometheus/prometheus/blob/f0a439bfc5d1f49cec113ee9202993be4b002b1b/tsdb/compact.go#L213), the blocks with big enough time range, that have >5% tombstones, are considered for compaction.) We solve the tombstones, if the tombstones are greater than than the threshold and then perform compaction. If not we attach the tombstone file to the new block. If multiple blocks are being compacted, we merge the tombstone files of the blocks whose threshold is not met.
```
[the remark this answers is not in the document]
bwplotka: Yes. I think the 5% rule should apply only for blocks that have a max level of compaction. Otherwise, all should be applied during compaction as it's essentially almost free to apply deletions during compaction.

The only caveat was about some `delay` time for tombstones to ensure undelete. This functionality was hower decided to be handled in later efforts, right?

docs/proposals/deletions_object_storage.md
```diff
@@ -0,0 +1,84 @@
...
+    *   **Block with deletion-mark.json i.e., entire block marked for deletion:** Returns an error message as the entire block is going to be deleted.
+*   **Perspectives to deal with Downsampling of blocks:**
+    *   **Block with tombstones:** If the tombstones are less than the threshold we copy the tombstone file and attach it to the new downsampled block else we solve the tombstones and downsample the block. And the downsampled block with tombstones during its next compaction would again have the same cases as with the compaction of a block with tombstones.
+    *   **Blocks without tombstones:** Downsampling happens...
+
+##### Problems during implementations and possible ideas to solve :
+During the implementation phase of this proposal one of the first problems we came across was that we had to pull the index of all the blocks for creating or appending a tombstone.
```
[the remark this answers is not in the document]
bwplotka: Yea, Compactor is not responsible for that. It does not load or need to load all the blocks in the system in order to craft tombstones. That's why we cannot use Prometheus format tombstone and have to figure out something else. If we would required to load indexes for ALL blocks in order to create deletion request - this will never scale. 

We need to delegate logic of figuring our which series ID to delete in leafs for querying and during compaction for rewrite.

The tricky part is to know which blocks have the data to be deleted. We thought about simple Series with `skipChunk` options

docs/proposals/deletions_object_storage.md
bwplotka: One thing at the time (: 

I think for this we might need just repeated deletion call - easy if deletion is one of action  (as it should be)

docs/proposals/deletions_object_storage.md
```diff
@@ -0,0 +1,84 @@
...
+## Goals
+
+*   Unblock users and allow series deletion in the object storage using tombstones.
+*   Deal with compaction and downsampling of blocks with tombstones.
+*   Performing deletions at admin level.
+
+## Proposed Approach
+
+*   We propose to implement deletions using the tombstones approach.
+*   To start off we use the Prometheus tombstone file format.
```
bwplotka, line 32: Let's make sure to point out why this is not possible, and let's put that to Alternatives

docs/proposals/deletions_object_storage.md
```diff
@@ -0,0 +1,84 @@
...
+##### Proposed Alternative Approaches:
+(1) To have a separate store like component with index cache.
+
+(2) To have a different format for the tombstones. Instead of having <seriesRef, min, max> maybe we can have <matcher, min, max>
+* **Idea 1:** Having a CLI tool for dealing with deletions.
+* **Idea 2:** To have all tombstones at some single place and they are used while performing compaction or during query time.
+* **Idea 3:** If we want to have an API for deletions, one way is to have the Store API configured in such a way that the compactor can use it to check for a match.
+*Edge case:* Do we rebuild all the rows of a tombstone? Because we need to be aware that the tombstone is being included(or not) in the given Store API call.
+
+*We're starting with (2) Idea 2*
```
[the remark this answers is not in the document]
bwplotka: Yes, let's be clear about this.

docs/proposals/deletions_object_storage.md
```diff
@@ -0,0 +1,84 @@
...
+    *   **Block with deletion-mark.json i.e., entire block marked for deletion:** Returns an error message as the entire block is going to be deleted.
+*   **Perspectives to deal with Downsampling of blocks:**
+    *   **Block with tombstones:** If the tombstones are less than the threshold we copy the tombstone file and attach it to the new downsampled block else we solve the tombstones and downsample the block. And the downsampled block with tombstones during its next compaction would again have the same cases as with the compaction of a block with tombstones.
+    *   **Blocks without tombstones:** Downsampling happens...
+
+##### Problems during implementations and possible ideas to solve :
+During the implementation phase of this proposal one of the first problems we came across was that we had to pull the index of all the blocks for creating or appending a tombstone.
+
+##### Proposed Alternative Approaches:
```
bwplotka, line 51: Normally we would do 

`## Alternative` sections with couple of alternative, but with clear separation. e.g

```
## Proposed Solution

* CLI.
* Custom format global - single file per request

.. elaborate on how and what challenges / open questions are.
 
## Alternatives

### Where Deletion API sits
 
1. Compactor: 
  * Pros: Easiest as it's single writer in Thanos (but not for Cortex)
  * Cons: Does not have quick access to querying or block indexes
1. Store GW:
  * Pros: Does have quick access to querying or block indexes
  * Cons: There would many writers, and store never was about to write things, it only needed read access

### How we store tombstone and in what format

1. Prometheus Format per block
1. Prometheus Format global
1. Custom format per block
1. Custom format global - appendable file

Comment (bwplotka) [reply docs/proposals/deletions_object_storage.md:51]: We can explain that `* Custom format global - single file per request` idea is good because:
* https://cloud-native.slack.com/archives/CL25937SP/p1595954850430300
* We can easily have multiple writers
* No need to have index data involved

## Chalenges
* How to "compact" / "remove" those files when applied.
* Do we want those deletion to by applied only during compaction or also for already compacted blocks. If yes how? 5%? And how to tell that when is 5%? What component would do that?
* Any rate limiting, what if there are too many files? Would that scale?

Comment (bwplotka) [reply docs/proposals/deletions_object_storage.md:51]: Plus other ideas to talk about later like:

Optimization for file names?

File: docs/proposals/deletions_object_storage.md [hunk 24]
```diff
@@ -0,0 +1,84 @@
...
+
+(2) To have a different format for the tombstones. Instead of having <seriesRef, min, max> maybe we can have <matcher, min, max>
+* **Idea 1:** Having a CLI tool for dealing with deletions.
+* **Idea 2:** To have all tombstones at some single place and they are used while performing compaction or during query time.
+* **Idea 3:** If we want to have an API for deletions, one way is to have the Store API configured in such a way that the compactor can use it to check for a match.
+*Edge case:* Do we rebuild all the rows of a tombstone? Because we need to be aware that the tombstone is being included(or not) in the given Store API call.
+
+*We're starting with (2) Idea 2*
+
+#### Considerations :
```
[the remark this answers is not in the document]
Harshitha1234: It was intended for the original proposal :)

bwplotka:
Hey, can we finalize this and make sure it's clean for other ppl? We just share it to Cortex devs, they would be happy to contribute (:

Harshitha1234:
Yes, sure. PTAL I've made the latest changes based on our last discussion about the proposal, if something doesn't seem right please let me know 🙂

docs/proposals/deletions_object_storage.md
```diff
@@ -0,0 +1,100 @@
...
+*   During compaction or downsampling, we check if the blocks being considered have series corresponding to a tombstone file, if so we delete the data and continue with the process.
+
+## Open Questions
+*   Optimization for file names?
+*   Do we create a tombstone irrespective of the presence of the series with the given matcher? Or should we check for the existence of series that corresponds to the deletion request?
+*   Should we consider having a threshold for compaction or downsampling where we check all the tombstones and calculate the percentage of deletions?
+
+## Considerations
+
+*   The old blocks and the tombstones are deleted during compaction time.
```
pracucci, line 52: I think this opens to a race condition.

Above it's mentioned this:
> Store Gateway masks the series on processing the global tombstone files from the object storage.

If the store gateway scans deleted tombstones before the old blocks are effectively offloaded by the store gateway (we do soft delete first, apply a delay and then delete hard delete them) we may open to a time window during which the tombstone is deleted, the store gateway removed it from its local map, the old block is still queried and we do re-expose deleted series.

docs/proposals/deletions_object_storage.md
```diff
@@ -0,0 +1,100 @@
...
+## Goals
+
+*   Unblock users and allow series deletion in the object storage using tombstones.
+*   Deal with compaction and downsampling of blocks with tombstones.
+*   Performing deletions at admin level.
+
+## Proposed Approach
+
+*   We propose to implement deletions via the tombstones approach using a CLI tool.
+*   A tombstone is proposed to be a **Custom format global - single file per request**.
```
pracucci, line 32: I haven't found the exact format of this file documented. Anyway, I would add the timestamp of the deletion request inside as well. Would allow to build hard deletion grace periods, without having to read it from the object storage (object creation).

metalmatze (approved):
Would be nice to finally merge. Seems like we at least agree on the first steps so far :blush:

docs/proposals/deletions_object_storage.md
bwplotka: still missing @Harshitha1234

docs/proposals/deletions_object_storage.md
```diff
@@ -0,0 +1,100 @@
...
+*   **Why Custom format global - single file per request**? :
+    *   https://cloud-native.slack.com/archives/CL25937SP/p1595954850430300
```
bwplotka, line 34: ? 

We need more details, can we use full sentence?

bwplotka: Also design is to make sure we have full view, so I would rather explain instead of linking some further reading (:

docs/proposals/deletions_object_storage.md
```diff
@@ -0,0 +1,100 @@
...
+    *   We can easily have multiple writers
+    *   No need to have index data of the blocks
+*   A user is expected to enter the following details for performing deletions:
+    *   **label matchers**
+    *   **start timestamp**
+    *   **end timestamp** (start and end timestamps of the series data the user expects to be deleted)
```
bwplotka, line 40: What about things like `reason` or `author`? (:

docs/proposals/deletions_object_storage.md
```diff
@@ -0,0 +1,100 @@
...
+*   The entered details are processed by the CLI tool to create a tombstone file (unique for a request), and then the tombstone file is uploaded to the object storage making it accessible to all components.
+*   Store Gateway masks the series on processing the global tombstone files from the object storage.
+*   During compaction or downsampling, we check if the blocks being considered have series corresponding to a tombstone file, if so we delete the data and continue with the process.
+
+## Open Questions
+*   Optimization for file names?
```
bwplotka, line 46: Can you describe more why we would need this? (:

docs/proposals/deletions_object_storage.md
```diff
@@ -0,0 +1,100 @@
...
+*   Do we create a tombstone irrespective of the presence of the series with the given matcher? Or should we check for the existence of series that corresponds to the deletion request?
```
bwplotka, line 47: I think this should be moved to proposal - we would love to avoid doing this.

bwplotka: We could have this in alternatives

docs/proposals/deletions_object_storage.md
```diff
@@ -0,0 +1,100 @@
...
+*   Should we consider having a threshold for compaction or downsampling where we check all the tombstones and calculate the percentage of deletions?
```
bwplotka, line 48: Yes - this should be covered as well here :thinking: 

As per - how to delete stuff from already compacted blocks section

docs/proposals/deletions_object_storage.md
bwplotka: Yes, it has to be delayed-deleted or something else... It's not easy, because:

* How we know there are no more stuff to delete then in mentioned block?
* What if someone backfills block that touches deleted period?
* What if min-time is .. not specified? Does it has to be?


Behaviour on all those questions have to be defined in this proposal, explicit and simple to understand for user of deletion API I think ):

bwplotka:
Let's get back to this... is it ready for review? :hugs:

Harshitha1234:
Yes, it is :)

Harshitha1234:
Still valid! @bwplotka Are there any key changes that we expect in this proposal? :)

bwplotka (approved):
No, just implement this 🤗

bwplotka:
Merged! Thanks @Harshitha1234 🤗

Harshitha1234:
> No, just implement this 🤗

Oh yeah! Let's do this then!! 💪

yeya24:
Tombstone files are global. Then is there a way to achieve some multi-tenancy or fine-grained control in this case?
What if I only want to delete series `{xxx}` in blocks with external labels `cluster=eu`? Or only want to delete series in raw blocks but not in downsampled blocks?

These cases cannot be easily handled with global tombstone files.

PR #2783, opened 2020-06-19
Introducing zero alloc labelpb, using it everywhere; Cleaned matchers.

brancz:

## Changes

We regularly run queries over weeks of data and millions (~20M) series in total. Doing so allows us to see various inefficiecies of Thanos pretty clearly in profiles, for example:

Screenshot from 2020-06-19 07-43-29

This PR attempts to port [Cortex's optimization](https://github.com/cortexproject/cortex/blob/55a4fcfbc07903fd722e9c72659fd8ff9a873f84/pkg/ingester/client/cortex.proto#L135) on [proto unmarshaling of Labels](https://github.com/cortexproject/cortex/blob/55a4fcfbc07903fd722e9c72659fd8ff9a873f84/pkg/ingester/client/timeseries.go#L73-L74), as the querier seems to spend a good amount of time just doing that (roughly 1/5 just on label unmarshaling, and majority of that is the byte slice to string conversion).

I time boxed myself on this, as I have a lot of other things to do, so if someone wants to pick this up, that would be fantastic :) 

## Verification

WIP :) 
.
@thanos-io/thanos-maintainers


pkg/store/storepb/labeladapter.go
```diff
@@ -0,0 +1,203 @@
...
+	"strings"
+	"unsafe"
+
+	"github.com/prometheus/prometheus/pkg/labels"
+)
+
+// LabelAdapter is a labels.Label that can be marshalled to/from protos.
+type LabelAdapter labels.Label
+
+func yoloString(buf []byte) string {
```
bwplotka, line 15: Awesome, but we can rename and add docs now (:

pkg/store/storepb/labeladapter.go
```diff
@@ -0,0 +1,203 @@
+package storepb
```
bwplotka, line 1: Curious if it's not the time to move this outside of storepb (rules has labels as well)

pkg/store/storepb/types.proto
```diff
@@ -33,7 +33,7 @@ message Chunk {
 }
 
 message Series {
-  repeated Label labels     = 1 [(gogoproto.nullable) = false];
+  repeated Label labels     = 1 [(gogoproto.nullable) = false, (gogoproto.customtype) = "LabelAdapter"];
```
bwplotka, line 36: WOW this is epic :muscle: 

I did not know you can do it ;p

brancz:
@bwplotka do you know if anyone is continuing this work?

bwplotka:
Yes. 

Me. (: Now.

bwplotka:
zero copy label marshall & umarshal

image

copy label: 

image

Results are insane!

```
/tmp/___BenchmarkLabelsMarshallUnmarshall_in_github_com_thanos_io_thanos_pkg_store_storepb_zcpylabels__1_ -test.v -test.bench ^\QBenchmarkLabelsMarshallUnmarshall\E$ -test.run ^$
goos: linux
goarch: amd64
pkg: github.com/thanos-io/thanos/pkg/store/storepb/zcpylabels
BenchmarkLabelsMarshallUnmarshall
BenchmarkLabelsMarshallUnmarshall/copy
BenchmarkLabelsMarshallUnmarshall/copy-12         	       5	 212321964 ns/op	369203625 B/op	 2000046 allocs/op
BenchmarkLabelsMarshallUnmarshall/zerocopy
BenchmarkLabelsMarshallUnmarshall/zerocopy-12     	       7	 143116871 ns/op	273203604 B/op	      46 allocs/op
PASS

Process finished with exit code 0
```

pkg/store/storepb/zcpylabels/label.go
```diff
@@ -0,0 +1,218 @@
...
+}
+
+// LabelsFromPromLabels converts Thanos proto no alloc labels to Prometheus labels in type unsafe manner.
+// It reuses the same memory. Caller should abort using passed labels.Labels.
+func LabelsFromPromLabels(lset labels.Labels) []Label {
+	return *(*[]Label)(unsafe.Pointer(&lset))
+}
+
+// LabelsToPromLabels converts Prometheus labels to Thanos proto no alloc labels in type unsafe manner.
+// It reuses the same memory. Caller should abort using passed []NoAllocLabels
```
squat, line 25: nit: there is no type called `NoAllocLabels`

bwplotka:
This PR has some last unit tests to be fixed and I am on it, but should be ready to review.

It is huge, but mostly rename from `storepb.Label` to `labelpb.Label` (: 

PTAL @brancz @kakkoyun @yeya24 

Benchmarks to come.

pkg/store/labelpb/label.go
```diff
@@ -0,0 +1,280 @@
...
+		i = encodeVarintTypes(data, i, uint64(len(m.Name)))
+		i--
+		data[i] = 0xa
+	}
+	return len(data) - i, nil
+}
+
+func (m *Label) Unmarshal(data []byte) error {
+	l := len(data)
+	iNdEx := 0
```
kakkoyun, line 84: 🤔 A new kind of case convention?

kakkoyun (commented):
It looks alright.... But I don't think this is easy on eyes and easy to review. I can tell the changes are somewhat straight forward but too many lines of diffs 😅 

Other than that DCO is missing and there are some failing CI stages. 

I'll continue to review.

pkg/store/labelpb/label.go
bwplotka: Copied from protobuf marshal - so we left that this way.

bwplotka:
I will reduce a diff for ease of use a bit

bwplotka:
PTAL @brancz @kakkoyun @squat Should be fine now.

bwplotka:
The key ideas of this PR:

* Proto Deserialization of labels is zero alloc. This is what `labelspb` is for. It also supports JSON marshall to Prometheus API supported JSON format.
* We try to use as much as possible Promethues `labels.Labels`, as the main labels package for functionality reasons.
* Some aliases with todos were left for easier PR iterations, will tackle in next PRs.
* Matchers were cleaned as well - we use labels.Matchers as much as possible as well everywhere.

brancz:
Honestly, if you can canary this and it looks good, then I'm happy to merge this :)

kakkoyun (approved):
As already mentioned, this is really hard to review. Let's go ahead and merge it, so that we can test it as a new release candidate. We can iterate on it if we discover any issues.

LGTM

bwplotka:
Soo I assume it was not reviewed fully?  (: 

I need to admit, that I am bit sad. I would rather expect this to take some time or maybe reviewers forcing me to split it more vs just merging. I know you trust me, but there are definitely some things I might have do wrong. 

> , our chance to spot something in the >2k lines are very unlikely 

Why? There is not time preassure, and there are other methods (e.g pair review).

Overall, I think this should not happen at all. I hope that you at least looked briefly (:

kakkoyun:
I actually reviewed it (as much as I can). However I might have missed things. It's really hard to catch anything with this amount of changes, and also these type of changes (relatively straightforward type conversions. I think for these we need to trust our test suite and benchmarks. 

I think splitting would only help to pinpoint the scope of a regression (if we have one), because they are all the same type of change at the end.

But for the future we can try to introduce a check to actually block a PR if it's too large. WDYT?

brancz:
I did a review as thoroughly as I could, I'm just saying that any type of review of this can (I believe) only go so far. All I was trying to say is, we can review this as many rounds as we want, I didn't think we were going to get closer to something "more mergable" than the state it was already in, and realistically only running this will tell us.

It's the same as SLOs, I'm already confident enough with how it has been reviewed, and any gain in confidence would be exponentially more expensive, and yet only actually running it will tell if there are still bugs. I think we've done enough.

bwplotka:
Ok, cool, just want to make sure it was reviewed in some way (:

bwplotka:
Tested on staging, working pretty well (:

Kind Regards,
Bartek Płotka (@bwplotka)

PR #2785, opened 2020-06-19
compact/tools: add hidden option for retrying downloads

GiedriusS:
In certain environments, especially private, on-premise clouds,
sometimes transient errors can happen due to misconfiguration or some
reason. In my case, the S3 end-point sometimes had spuriously reset the TCP
connection. Without functionality like this, it was almost impossible
for Thanos Compact to finish its work considering sometimes we have to
download hundreds of gigabytes of files.

Curiously enough, only the downloading path was affected so this only
adds retries for that direction.

Opening this up without any tests to see if it is something we want to
add to Thanos or we should implement
https://github.com/thanos-io/thanos/issues/2784? Or maybe both?

## Changes

Added exponential backoff with a configurable amount of retries to `block.Download`

## Verification

N/A for now

### DIFF
--- a/cmd/thanos/compact.go
+++ b/cmd/thanos/compact.go
@@ -291,7 +291,7 @@ func runCompact(
 	}
 
 	compactMainFn := func() error {
-		if err := compactor.Compact(ctx); err != nil {
+		if err := compactor.Compact(ctx, conf.downloadRetries); err != nil {
 			return errors.Wrap(err, "compaction")
 		}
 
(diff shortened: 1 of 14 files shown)


cmd/thanos/compact.go
```diff
@@ -499,6 +500,9 @@ func (cc *compactConfig) registerFlag(cmd *kingpin.CmdClause) *compactConfig {
 		"This works well for deduplication of blocks with **precisely the same samples** like produced by Receiver replication.").
 		Hidden().StringsVar(&cc.dedupReplicaLabels)
 
+	cmd.Flag("download.retries", "How many time to retry downloading a block with exponential backoff before giving up").
```
bwplotka, line 503: hm again, this is similar to the https://github.com/thanos-io/thanos/pull/2908 discussion. Shouldn't this be applied by upload/download client itself?

bwplotka:
:cry:

PR #2786, opened 2020-06-19
continue compact group even if occur a error

ifooth:
compact group is independent of each other, even if one error does not affect other compression, this is very necessary for massive data, and is very effective for my company's scenario

Signed-off-by: joelei <thezero12@hotmail.com>

## Changes

- continue compact group even if occur a error
- show compact ratio, useful for a long compact

## Verification

Tested it on prod environment, 20T data, 60K blocks, with continue compact, successfully reduced to 6k blocks

### DIFF
--- a/pkg/compact/compact.go
+++ b/pkg/compact/compact.go
@@ -959,7 +959,7 @@ func (c *BucketCompactor) Compact(ctx context.Context) (rerr error) {
 
 		level.Info(c.logger).Log("msg", "start of compactions")
 		st := time.Now()
-		compactCount := 0
+		progressCount := 0
 
 		// Send all groups found during this pass to the compaction workers.
 		var groupErrs terrors.MultiError
@@ -968,8 +968,8 @@ func (c *BucketCompactor) Compact(ctx context.Context) (rerr error) {
 			case groupErr := <-errChan:
 				groupErrs.Add(groupErr)
 			case groupChan <- g:
-				compactCount++
-				level.Info(c.logger).Log("msg", "start compact group", "group", g.Key(), "compactRatio", fmt.Sprintf("%d/%d", compactCount, len(groups)), "duration", time.Since(st))
+				progressCount++
+				level.Info(c.logger).Log("msg", "get compact group", "group", g.Key(), "progressRate", fmt.Sprintf("%d/%d", progressCount, len(groups)))
 			}
 		}
 		close(groupChan)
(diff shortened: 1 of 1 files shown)


pkg/compact/compact.go
```diff
@@ -958,16 +958,18 @@ func (c *BucketCompactor) Compact(ctx context.Context) (rerr error) {
...
+		compactCount := 0
 
 		// Send all groups found during this pass to the compaction workers.
 		var groupErrs terrors.MultiError
-	groupLoop:
 		for _, g := range groups {
 			select {
 			case groupErr := <-errChan:
 				groupErrs.Add(groupErr)
-				break groupLoop
```
pracucci, line 969: I kinda agree on not breaking, but I think we should continue breaking in case of `IsHaltError(groupErr)`.

pkg/compact/compact.go
```diff
@@ -958,16 +958,18 @@ func (c *BucketCompactor) Compact(ctx context.Context) (rerr error) {
...
 			case groupChan <- g:
+				compactCount++
+				level.Info(c.logger).Log("msg", "start compact group", "group", g.Key(), "compactRatio", fmt.Sprintf("%d/%d", compactCount, len(groups)), "duration", time.Since(st))
```
pracucci, line 972: I'm OK adding a log at the beginning of a group compaction, but this is not the right place to log the begin of a compaction. It should be logged at beginning of `Group.Compact()`. I'm also not much sure about the value of `compactRatio` and `duration` to be honest. What's the use case for that?

pkg/compact/compact.go
ifooth: Thanks for your comment, en, I think it is a good way to continue compact
- one loop, compact all group as expected, for lot's of block, this is very useful(if break on any halt error, the following groups may not be executed at once)
- halt or retry will decide in external function (whether alarm, etc.), see the screenshot below

image

pkg/compact/compact.go
ifooth: yeah, it's not a message for group, it's  for log for compact loop, very useful for lot's of groups and system manager have no idea the process of compact
- compactRation may be confuse `compact`, i will change to `progressRate`
- remove `duration`, and add a `finish of compactions` log message

thanks for your comment again

ifooth:
@pracucci Please review this PR again, let me know if has any question，I have time to improve it recently

PR #2788, opened 2020-06-21
Upgrade grpc middleware to v2

yashrsharma44:
Upgraded the `grpc_middleware` from `v1` to `v2`
Currently using the release candidate version - https://github.com/grpc-ecosystem/go-grpc-middleware/releases/tag/v2.0.0-rc.2

## Changes

Updated the files that imported the old grpc middleware module

## Verification
`make test`

## Why it is needed

The grpc middleware currently went through a change in the internal module structure, while maintaining the same [API](https://github.com/grpc-ecosystem/go-grpc-middleware/issues/275), introducing lot more features, while keeping the **initial API same.**

We would also utilise the logging middleware which **we don't currently use**, but is quite simple to use in the v2, so makes sense to upgrade.

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -27,6 +27,7 @@ require (
 	github.com/googleapis/gax-go v2.0.2+incompatible
 	github.com/gophercloud/gophercloud v0.11.0
 	github.com/grpc-ecosystem/go-grpc-middleware v1.1.0
+	github.com/grpc-ecosystem/go-grpc-middleware/v2 v2.0.0-rc.2
 	github.com/grpc-ecosystem/go-grpc-prometheus v1.2.0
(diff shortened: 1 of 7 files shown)


povilasv:
as far as I can tell it's still v2-rc2 so let's not merge this until they do a proper stable  release 

https://github.com/grpc-ecosystem/go-grpc-middleware/releases

yashrsharma44:
Sure, let me update the details.

bwplotka:
I don't think there are changes logically. V2 aims for compatibility, it's just rewrite with removal of certain options that we don't use.

Thus I believe for now, no changelog required (:

yashrsharma44:
This PR is superseded by #2907

PR #2796, opened 2020-06-23
Implement CLI Flags page in React UI

prmsrswt:

## Changes
- New API endpoint `/status/flags` in `query` to expose provided cli flags.
- Updated `thanos/Navbar.tsx` in React UI to handle dropdowns as well.

## Verification
Tested locally, with both development and production build.


bwplotka:
Nice, BTW I just got idea: We could run Thanos components for UI purposes on netlfy so we can preview of all your amazing work! :heart:

GiedriusS:
> Nice, BTW I just got idea: We could run Thanos components for UI purposes on netlfy so we can preview of all your amazing work! heart

We talked about this on our sync: that's definitely something we could explore but for this, a publicly available instance of Thanos Query would be needed, and that would take much more time than just deploying the resulting pages onto Netlify. :+1: Prometheus could also benefit from something like that.

cmd/thanos/query.go
```diff
@@ -141,6 +141,19 @@ func registerQuery(m map[string]setupFunc, app *kingpin.Application) {
...
+
+		// Exclude kingpin default flags to expose only Thanos ones.
+		boilerplateFlags := kingpin.New("", "").Version("")
+
+		for _, f := range cmd.Model().Flags {
+			if boilerplateFlags.GetFlag(f.Name) != nil {
+				continue
+			}
+
+			flagsMap[f.Name] = f.Value.String()
```
GiedriusS, line 154: We should skip over setting this if `f.Value.String()` is empty otherwise we will get a bunch of options in the API response with keys but empty values.

prmsrswt: I was thinking that, but I thought it would still be helpful if someone wants to see that if a particular flag has been set or not. wdyt?

GiedriusS: Double checked and it seems like Prometheus does the same. I guess it looked a bit weirder on Thanos because there are more flags without any default values. But, you are correct, let's leave these for clarity.

GiedriusS:
Let's resolve the conflicts and I think this should be good to go :tada:

bwplotka:
> We talked about this on our sync: that's definitely something we could explore but for this, a publicly available instance of Thanos Query would be needed, and that would take much more time than just deploying the resulting pages onto Netlify. +1 Prometheus could also benefit from something like that.

Yes, this will need bit more work(:

PR #2800, opened 2020-06-24
query: Fix external prefix and add an e2e test for it

prmsrswt:

## Changes
- Test `--web.external-prefix` using a simple reverse proxy
- Test both `--web.external-prefix` and `--web.route-prefix` in a single scenario, i.e. when both are set simultaneously.

### Changes in `externalPrefix` and `routePrefix` handling
- Now, `--web.route-prefix` defaults to `--web.external-prefix`. This behavior allows a user to use `thanos` with and without a reverse proxy if `--web.external-prefix` is set, and it is in line with what Prometheus does.
- Updated both, the old UI and the new React UI  to account for `--web.external-prefix` 

~~Turns out, `--web.external-prefix` is still broken, and as expected, these tests would fail unless we fix it.~~
It should work now.

This PR changes how `--web.external-prefix` works, so it can make the UI inaccessible to some users who were relying on some workarounds or hacks before (like setting `--web.external-prefix="."`).

## Verification
Ran `GOTEST_OPTS="-run TestQuery" make test-e2e` locally to test all e2e tests related to Query component.


prmsrswt:
Now that we are more in line with what Prometheus does, the following guides also works for Thanos.

- https://www.robustperception.io/using-external-urls-and-proxies-with-prometheus
- https://www.robustperception.io/external-urls-and-path-prefixes

The only difference is that, instead of using the Prometheus flag `--web.external-url` one needs to use `--web.external-prefix` flag in Thanos. So `prometheus --web.external-url http://localhost:19090/prometheus/` becomes `thanos query --web.external-prefix=thanos` for Thanos.

test/e2e/query_test.go
```diff
@@ -166,23 +167,85 @@ func TestQuery(t *testing.T) {
...
+		nil,
+		"",
+		externalPrefix,
+	)
+	testutil.Ok(t, err)
+	testutil.Ok(t, s.StartAndWaitReady(q))
+
+	querierURL := urlParse(t, "http://"+q.HTTPEndpoint()+"/"+externalPrefix)
+
+	querierProxy := httptest.NewServer(e2ethanos.NewSingleHostReverseProxy(querierURL, externalPrefix))
```
GiedriusS, line 215: Let's add `defer querierProxy.Close()` here

test/e2e/query_test.go
```diff
@@ -166,23 +167,85 @@ func TestQuery(t *testing.T) {
...
+		nil,
+		routePrefix,
+		externalPrefix,
+	)
+	testutil.Ok(t, err)
+	testutil.Ok(t, s.StartAndWaitReady(q))
+
+	querierURL := urlParse(t, "http://"+q.HTTPEndpoint()+"/"+routePrefix)
+
+	querierProxy := httptest.NewServer(e2ethanos.NewSingleHostReverseProxy(querierURL, externalPrefix))
```
GiedriusS, line 243: Ditto.

test/e2e/e2ethanos/helpers.go
```diff
@@ -19,3 +23,38 @@ func CleanScenario(t *testing.T, s *e2e.Scenario) func() {
...
+		req.URL.Path = singleJoiningSlash(target.Path, strings.TrimPrefix(req.URL.Path, "/"+externalPrefix))
+
+		if targetQuery == "" || req.URL.RawQuery == "" {
+			req.URL.RawQuery = targetQuery + req.URL.RawQuery
+		} else {
+			req.URL.RawQuery = targetQuery + "&" + req.URL.RawQuery
+		}
+
+		if _, ok := req.Header["User-Agent"]; !ok {
+			// Explicitly disable User-Agent so it's not set to default value.
```
GiedriusS, line 55: This doesn't seem to match with what is happening here. We set it to `""` if it does _not_ exist already. What's the rationale behind this?

cmd/thanos/query.go
```diff
@@ -139,6 +139,14 @@ func registerQuery(m map[string]setupFunc, app *kingpin.Application) {
 			fileSD = file.NewDiscovery(conf, logger)
 		}
 
+		if *webRoutePrefix == "" {
+			*webRoutePrefix = *webExternalPrefix
+		}
+
+		if *webRoutePrefix != *webExternalPrefix {
+			level.Warn(logger).Log("msg", "different values for --web.route-prefix and --web.external-prefix detected. WebUI may not work without a reverse-proxy")
```
GiedriusS, line 147: Kudos for this :+1:

test/e2e/query_test.go
prmsrswt: Gotcha. Also a question, can we use `t.Cleanup` here instead of `defer`?

test/e2e/e2ethanos/helpers.go
prmsrswt: This function is exactly same as https://golang.org/src/net/http/httputil/reverseproxy.go?s=3655:3716#L105
So I didn't thought about it much.

I can't do URL rewrite using `httputil.NewSingleHostReverseProxy` so I just modified it for my needs. The only difference between both of them is that I trim the `externalPrefix` from `url.Path` in my implementation.

test/e2e/query_test.go
GiedriusS: Sure, that's even better.

test/e2e/e2ethanos/helpers.go
GiedriusS: The commit message here https://github.com/golang/go/commit/6a6c1d9841a1957a2fd292df776ea920ae38ea00 explains my question. Thanks for the pointer!

test/e2e/query_test.go
prmsrswt: Fixed this :)

test/e2e/query_test.go
prmsrswt: Addressed this.

prmsrswt:
Seems like Netlify is having some problems.

docs/operating/reverse-proxy.md
```diff
@@ -0,0 +1,70 @@
+---
+title: Running Thanos behind a reverse proxy
+type: docs
+menu: operating
+slug: /reverse-proxy.md
+---
+
+# Running Thanos behind a reverse proxy
+
+Many times you would like to use a [reverse proxy](https://www.nginx.com/resources/glossary/reverse-proxy-server/) in front of Thanos. There are several reasons to use a reverse proxy such that SSL termination (serve Thanos over HTTPS) and basic authentication.
```
GiedriusS, line 10:
```suggestion
There are many reasons to use a [reverse proxy](https://www.nginx.com/resources/glossary/reverse-proxy-server/) in front of Thanos. For example, SSL termination (serve Thanos over HTTPS) and basic authentication. This small guide will tell you how to do that.
```

docs/operating/reverse-proxy.md
```diff
@@ -0,0 +1,70 @@
...
+
+## External Prefix and Route
+
+Before continuing, lets first take a look at two cli flags provided by Thanos, `--web.external-prefix` and `--web.route-prefix`. The External Prefix is useful when you want to prefix all requests from UI with some path. Normally, the web UI would load all static assets from `/static/...`, do API calls on `/api/v1/...` etc. But if we use `--web.external-prefix="thanos"` the UI would prefix each request with `/thanos`. It would try to load all assets from `/thanos/static/...`, do API calls on `/thanos/api/v1/...` and so on. One thing to note here is that `--web.external-prefix` only prefixes the requests and redirects with the specified value, but Thanos is still listening on the root, not the specified sub-path i.e. the API is still accessible at `/api/v1/...` and not at `/thanos/api/v1`. This is where `--web.route-prefix` comes in. If you set `--web.route-prefix="thanos"` every route would get prefixed with the specified value. For example, the API will be accessible on `/thanos/api/v1`. As this is the most common use case when using the `--web.external-prefix`, the default value of `--web.route-prefix` is the value of `--web.external-prefix`.
```
GiedriusS, line 14:
```suggestion
Before continuing, let's first take a look at two CLI flags provided by Thanos, `--web.external-prefix` and `--web.route-prefix`. The external prefix is useful when you want to prefix all requests from the UI with some path. Normally, the web UI would load all static assets from `/static/...`, do API calls on `/api/v1/...` etc. But if we use `--web.external-prefix="thanos"` the UI would prefix each request with `/thanos`. It would try to load all assets from `/thanos/static/...`, do API calls on `/thanos/api/v1/...` and so on. One thing to note here is that `--web.external-prefix` only prefixes the requests and redirects with the specified value, but Thanos is still listening on the root, not the specified sub-path i.e. the API is still accessible at `/api/v1/...` and not at `/thanos/api/v1`. This is where `--web.route-prefix` comes in. If you set `--web.route-prefix="thanos"` every route would get prefixed with the specified value. For example, the API will be accessible on `/thanos/api/v1`. As this is the most common use case when using the `--web.external-prefix`, the default value of `--web.route-prefix` is the value of `--web.external-prefix`.
```

docs/operating/reverse-proxy.md
```diff
@@ -0,0 +1,70 @@
...
+
+> Note: Using different values for `--web.external-prefix` and `--web.route-prefix` can lead to the web UI not working properly if it is accessed directly (without a reverse proxy).
```
GiedriusS, line 16: Is there a reason to quote here?

```suggestion
Note: Using different values for `--web.external-prefix` and `--web.route-prefix` can lead to the web UI not working properly if it is accessed directly (without a reverse proxy).
```

docs/operating/reverse-proxy.md
```diff
@@ -0,0 +1,70 @@
...
+
+## Examples
+
+Lets look into some example scenarios. All examples are using Nginx as a reverse proxy here.
```
GiedriusS, line 20:
```suggestion
Let's look into some example scenarios. All examples are using nginx as a reverse proxy here.
```
https://en.wikipedia.org/wiki/Nginx
> stylized as NGINX or nginx or NginX

docs/operating/reverse-proxy.md
```diff
@@ -0,0 +1,70 @@
...
+
+### Serving Thanos on a subdomain
+
+Serving a Thanos component on the root of a subdomain is pretty straight-forward. Let's say you want to run Thanos querier behind a Nginx server, accessible on domain `thanos.example.com`. A basic Nginx configuration would look like this:
```
GiedriusS, line 24:
```suggestion
Serving a Thanos component on the root of a subdomain is pretty straight-forward. Let's say you want to run Thanos Query behind a nginx reverse proxy, accessible on domain `thanos.example.com`. A basic nginx configuration would look like this:
```

docs/operating/reverse-proxy.md
```diff
@@ -0,0 +1,70 @@
...
+        listen 80;
+        server_name thanos.example.com;
+
+        location / {
+            proxy_pass    http://localhost:10902/;
+        }
+    }
+}
+
+events {}
```
GiedriusS, line 38: Is there a reason to have this here? Only having the `http` block should be sufficient. Plus, this is always empty.

docs/operating/reverse-proxy.md
```diff
@@ -0,0 +1,70 @@
...
+```
+
+### Serving Thanos on a sub-path
+
+Things become a little tricky when you want to serve Thanos on a sub-path. Let's say, you want to run Thanos querier behind an Nginx server, accessible on the URL `http://example.com/thanos`. The Thanos web UI depends on it being accessed on the same URL as Thanos itself is listening. This is because the UI needs to know the URL from where to load static assets and what URL to use in links or redirects. If Thanos is behind a reverse proxy, particularly one where Thanos is not at the root, this doesn't work so well.
```
GiedriusS, line 43:
```suggestion
Things become a little tricky when you want to serve Thanos on a sub-path. Let's say, you want to run Thanos Query behind an nginx server, accessible on the URL `http://example.com/thanos`. The Thanos web UI depends on it being accessed on the same URL as Thanos itself is listening. This is because the UI needs to know the URL from where to load static assets and what URL to use in links or redirects. If Thanos is behind a reverse proxy, particularly one where Thanos is not at the root, this doesn't work so well.
```

docs/operating/reverse-proxy.md
```diff
@@ -0,0 +1,70 @@
...
+
+To tackle this problem, Thanos provides a flag `--web.external-prefix`.
+
+Let's say we have Thanos querier running on the usual port, we need Nginx running with the following configuration:
```
GiedriusS, line 47:
```suggestion
Let's say we have Thanos Query running on the usual port, we need nginx running with the following configuration:
```

docs/operating/reverse-proxy.md
```diff
@@ -0,0 +1,70 @@
...
+        listen 80;
+        server_name example.com;
+
+        location /thanos/ {
+            proxy_pass    http://localhost:10902/thanos/;
+        }
+    }
+}
+
+events {}
```
GiedriusS, line 61: Remove this?

docs/operating/reverse-proxy.md
```diff
@@ -0,0 +1,70 @@
...
+```
+
+With this configuration, you can access Thanos querier on `http://example.com/thanos`. Notice that as we are using `http://localhost:10902/thanos/` as the reverse proxy target, every request path will be prefixed with `/thanos`. To make this work we need t run Thanos querier like
```
GiedriusS, line 64:
```suggestion
With this configuration, you can access Thanos Query on `http://example.com/thanos`. Notice that because we are using `http://localhost:10902/thanos/` as the reverse proxy target, every request path will be prefixed with `/thanos`. To make this work we need to run Thanos Query like this:
```

docs/operating/reverse-proxy.md
```diff
@@ -0,0 +1,70 @@
...
+
+```
+thanos query --web.external-prefix="thanos"
+```
+
+It should be noted that now the `/thanos` path prefix would be required for all HTTP access to Thanos. The `/metrics` endpoint would be accessible at `http://localhost:10902/thanos/metrics`.
```
GiedriusS, line 70:
```suggestion
It should be noted that now the `/thanos` path prefix would be required for all HTTP access to Thanos. For example, the `/metrics` endpoint would be accessible at `http://localhost:10902/thanos/metrics`.
```

docs/operating/reverse-proxy.md
prmsrswt: Oh I just did this because many blogs use quote for Notes, but yeah I agree, it isn't needed here.

docs/operating/reverse-proxy.md
prmsrswt: Oh gotcha, I'll change it in other places as well.

docs/operating/reverse-proxy.md
prmsrswt: I got this from the Robust Perception blog, I researched a bit and I am pretty sure it isn't needed so I am going to remove this.

docs/operating/reverse-proxy.md
```diff
@@ -0,0 +1,66 @@
+---
+title: Running Thanos behind a reverse proxy
+type: docs
+menu: operating
+slug: /reverse-proxy.md
+---
+
+# Running Thanos behind a reverse proxy
+
+There are many reasons to use a [reverse proxy](https://www.nginx.com/resources/glossary/reverse-proxy-server/) in front of Thanos. For example, SSL termination (serve Thanos over HTTPS) and basic authentication. This small guide will tell you how to do that.
```
squat, line 10: Nit, the sentence starting with "For example" has no independent clause, ie it's not a full sentence. Let's join it with the previous sentence

squat: Nit s/SSL/TLS/

cmd/thanos/query.go
squat, line 147: Nit, let's be consistent and use `web UI` rather than `WebUI`

test/e2e/e2ethanos/helpers.go
squat, line 55: In understand why a general purpose library like go's net/http would need this, buy why do we need this logic?

prmsrswt: Gotcha, I will remove the unnecessary bits.

docs/operating/reverse-proxy.md
prmsrswt: Sure :+1: .

GiedriusS:
Oops, fixing up the CHANGELOG in https://github.com/thanos-io/thanos/pull/2833.

Issue #2805, opened 2020-06-25
Compactor/S3: sometimes segments are not being fully downloaded ("segment doesn't include enough bytes")


JensErat:
**Thanos, Prometheus and Golang version used**:

Thanos 0.13 rc2, currently upgrading to 0.13 but commits do not indicate any relevant change since the tag was crafted

**Object Storage Provider**:

Internal Cloudian S3 service

**What happened**:

Compactor halts with following error message:

`level=error ts=2020-06-25T07:57:55.183632304Z caller=compact.go:375 msg="critical error detected; halting" err="compaction: group 0@9955244503410235132: compact blocks [/var/thanos/compact/data/compact/0@9955244503410235132/01EAQGV34VQ8PWVKXJ4V2K8M5B /var/thanos/compact/data/compact/0@9955244503410235132/01EAQQPTCT5ABJSXRBMG53V8ZC /var/thanos/compact/data/compact/0@9955244503410235132/01EAQYJHMZ9R0V8M5Y3GPG0NWD /var/thanos/compact/data/compact/0@9955244503410235132/01EAR5E8WTX7MJP7PG1DQMPY4W]: write compaction: iterate compaction set: chunk 41943009 not found: segment doesn't include enough bytes to read the chunk - required:41943044, available:41943040"`

After restarting compactor, compacting continues successfully (usually hours later, we only get warned if compactor does not run for quite some time -- so I cannot tell whether this is a pretty short inconsistency window and lots of bad luck involved).

**What you expected to happen**:

If this is really a transient issue based on eventual consistency issues, retry and don't halt.

**How to reproduce it (as minimally and precisely as possible)**:

Slightly overload your S3 service and wait. We have a fail rate of estimated once a week for a single compactor. Mostly happens in installations with some network distance (ping >= 200ms)

**Full logs to relevant components**:

<details>Logs
<p>

```
[...]
level=info ts=2020-06-25T07:42:56.091077077Z caller=block.go:157 group="0@{monitor=\"prometheus\", replica=\"10\"}" groupKey=0@7398988455995530398 msg="block has been marked for deletion" block=01EAQYJKPJS17YPE28VZ48JF7Q
level=info ts=2020-06-25T07:42:56.11841702Z caller=compact.go:792 group="0@{monitor=\"prometheus\", replica=\"10\"}" groupKey=0@7398988455995530398 msg="marking compacted block for deletion" old_block=01EAR5EAY80Z79JCDWQBAVVGJ4
[...]

PR #2806, opened 2020-06-25
compact: set unready if halted

JensErat:
If comapctor is halted, it should not report ready state any more.

### DIFF
--- a/.bingo/go.mod
+++ b/.bingo/go.mod
@@ -1,1 +1,3 @@
 module _ // Fake go.mod auto-created by 'bingo' for go -moddir compatibility with non-Go projects. Commit this file, together with other .mod files.
+
+go 1.13
--- a/cmd/thanos/compact.go
+++ b/cmd/thanos/compact.go
@@ -357,6 +357,7 @@ func runCompact(
 				if conf.haltOnError {
 					level.Error(logger).Log("msg", "critical error detected; halting", "err", err)
 					halted.Set(1)
+					httpProbe.NotHealthy(fmt.Errorf("critical error detected; halting: %v", err))
 					select {}
 				} else {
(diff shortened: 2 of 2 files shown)


JensErat:
Seems I was somehow to fast in sending the PR through the command line (actually wanted to add some more notes on exactly that this might be subject of discussion regarding whether this is wanted), but seems I actually sent the first copy instead of cancelling...

I guess in this case you probably should remove the health check? In this case Kubernetes will not restart it any more. I know I can also set the hidden debug flag (actually, that's what I'm deploying right now because of #2805), but saying "compactor is ready" although it actually failed feels slightly wrong.

GiedriusS:
> Thanks! But does this mean that Kubernetes will restart it, right?
> 
> So this means this change is the same as the setting --no-halt-on-error flag (: No? We actually want still access UI's and everything is working just fine on halt, just compaction is halted, so from HTTP and UI perspective it's ready. What do you think? thinking

I concur with @bwplotka here. That flag should suffice. Plus, the whole idea is that when Thanos Compactor is halted like this, it should still remain up so that you could go and troubleshoot what had happened.

bwplotka:
Yes. Essentially if we mark it as unready we could equally just not halt it and crash (: Same effect.

.bingo/go.mod
```diff
@@ -1 +1,3 @@
-module _ // Fake go.mod auto-created by 'bingo' for go -moddir compatibility with non-Go projects. Commit this file, together with other .mod files.
\ No newline at end of file
+module _ // Fake go.mod auto-created by 'bingo' for go -moddir compatibility with non-Go projects. Commit this file, together with other .mod files.
+
+go 1.13
```
kakkoyun, line 3: On the other note, bingo requires at least go 1.14

PR #2809, opened 2020-06-25
Query: always return a string in the `lastError` field

GiedriusS:
While testing out the new React UI, I have found out that some errors
return empty dicts when marshalled into the JSON format. The response
looks like this:

```
{..., "lastError": {}}
```

And then, obviously, the UI fails to parse that. Add a
`stringifiedError` type with tests which ensures that we always get a
string.

## Changes

Add a small, new type which ensures that we always get strings in the `lastError` field of `/api/v1/stores`.

## Verification

Unit tests.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -19,7 +19,7 @@ We use *breaking* word for marking changes that are not backward compatible (rel
 - [#2728](https://github.com/thanos-io/thanos/pull/2728) Query: Fixed panics when using larger number of replica labels with short series label sets.
 - [#2787](https://github.com/thanos-io/thanos/pull/2787) Update Prometheus mod to pull in prometheus/prometheus#7414.
 - [#2807](https://github.com/thanos-io/thanos/pull/2807) Store: decreased memory allocations while querying block's index.
-- []() Query: `/api/v1/stores` now guarantees to return a string in the `lastError` field 
+- [#2809](https://github.com/thanos-io/thanos/pull/2809) Query: `/api/v1/stores` now guarantees to return a string in the `lastError` field
 
 ### Changed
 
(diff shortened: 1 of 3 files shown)


GiedriusS:
cc @prmsrswt

CHANGELOG.md
```diff
@@ -19,6 +19,7 @@ We use *breaking* word for marking changes that are not backward compatible (rel
 - [#2728](https://github.com/thanos-io/thanos/pull/2728) Query: Fixed panics when using larger number of replica labels with short series label sets.
 - [#2787](https://github.com/thanos-io/thanos/pull/2787) Update Prometheus mod to pull in prometheus/prometheus#7414.
 - [#2807](https://github.com/thanos-io/thanos/pull/2807) Store: decreased memory allocations while querying block's index.
+- [#2809](https://github.com/thanos-io/thanos/pull/2809) Query: `/api/v1/stores` now guarantees to return a string in the `lastError` field
```
bwplotka, line 22:
```suggestion
- [#2809](https://github.com/thanos-io/thanos/pull/2809) Query: `/api/v1/stores` now guarantees to return a string in the `lastError` field,
```

pkg/query/storeset.go
```diff
@@ -47,10 +48,26 @@ type RuleSpec interface {
 	Addr() string
 }
 
+// stringifiedError forces the error to be a string
+// when marshalled into a JSON.
+type stringifiedError struct {
```
bwplotka, line 53:
```suggestion
type stringError struct {
```

bwplotka: Maybe simper? (:

CHANGELOG.md
GiedriusS: You probably mean a full stop here? :P

pkg/query/storeset.go
GiedriusS: Sure, sounds good :+1:

PR #2812, opened 2020-06-26
store: Preallocate output buffer when encoding postings.

pstibrany:
This PR modifies `diffVarintSnappyEncode` method to preallocate output buffer based on number of passed postings. This helps to avoid reallocations.

```
name                      old time/op    new time/op    delta
EncodePostings/10000-4       114µs ± 1%     106µs ± 1%   -6.55%  (p=0.009 n=3+9)
EncodePostings/100000-4     1.11ms ± 1%    1.05ms ± 0%   -5.39%  (p=0.007 n=3+10)
EncodePostings/1000000-4    11.6ms ± 2%    10.7ms ± 0%   -7.21%  (p=0.009 n=3+9)

name                      old alloc/op   new alloc/op   delta
EncodePostings/10000-4      48.5kB ± 0%    13.6kB ± 0%  -71.98%  (p=0.000 n=3+10)
EncodePostings/100000-4      506kB ± 0%     131kB ± 0%  -74.08%  (p=0.000 n=3+10)
EncodePostings/1000000-4    5.86MB ± 0%    1.25MB ± 0%  -78.62%  (p=0.000 n=3+10)

name                      old allocs/op  new allocs/op  delta
EncodePostings/10000-4        18.0 ± 0%       2.0 ± 0%  -88.89%  (p=0.000 n=3+10)
EncodePostings/100000-4       26.0 ± 0%       2.0 ± 0%  -92.31%  (p=0.000 n=3+10)
EncodePostings/1000000-4      36.0 ± 0%       2.0 ± 0%  -94.44%  (p=0.000 n=3+10)
```

Notes for reviewers: I chose to pass expected number of postings as another parameter. Alternative I considered was introducing an interface with `PostingsCount` method, which may look nicer, but would add hidden dependency into `diffVarintSnappyEncode` -- for postings implementing the interface, it would be faster (use less allocs) than for postings not implementing it. Passing extra parameter is more explicit.

* [NA] I added CHANGELOG entry for this change.

## Verification

Added benchmark.

### DIFF
--- a/pkg/store/postings_codec.go
+++ b/pkg/store/postings_codec.go
@@ -33,9 +33,9 @@ func isDiffVarintSnappyEncodedPostings(input []byte) bool {
 // diffVarintSnappyEncode encodes postings into diff+varint representation,
 // and applies snappy compression on the result.
 // Returned byte slice starts with codecHeaderSnappy header.
-// Values argument is expected number of postings, used for preallocating buffer.
-func diffVarintSnappyEncode(p index.Postings, values int) ([]byte, error) {
-	buf, err := diffVarintEncodeNoHeader(p, values)
+// Length argument is expected number of postings, used for preallocating buffer.
+func diffVarintSnappyEncode(p index.Postings, length int) ([]byte, error) {
+	buf, err := diffVarintEncodeNoHeader(p, length)
 	if err != nil {
 		return nil, err
 	}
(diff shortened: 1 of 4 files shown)


pkg/store/postings_codec.go
```diff
@@ -33,8 +33,9 @@ func isDiffVarintSnappyEncodedPostings(input []byte) bool {
 // diffVarintSnappyEncode encodes postings into diff+varint representation,
 // and applies snappy compression on the result.
 // Returned byte slice starts with codecHeaderSnappy header.
-func diffVarintSnappyEncode(p index.Postings) ([]byte, error) {
-	buf, err := diffVarintEncodeNoHeader(p)
+// Values argument is expected number of postings, used for preallocating buffer.
+func diffVarintSnappyEncode(p index.Postings, values int) ([]byte, error) {
```
pracucci, line 37: Maybe renaming `values` in `length` could be a bit more clear.

bwplotka: :+1:

pstibrany: ok, `length` it is.

pstibrany:
> 💪
> 
> Motto for the week: PREALLOCATE ;p

Indeed :)

pstibrany:
Test and doc failures seem to be unrelated to my changes :-(

bwplotka:
Thanks!

PR #2813, opened 2020-06-26
Improve ThanosReceiveNoUpload to only alert on current instances

metalmatze:
* [x] We don't have a CHANGELOG entry for mixin.

## Changes

Screenshot from 2020-06-26 17-51-22-blurred

## Verification

Old query on the left and new query on the right. One can clearly see how the old query is still alerting on the long gone Pods, while we filter those on the right side and the number of NoUpload instances decreases while restarting Pods.

### DIFF
--- a/examples/alerts/alerts.md
+++ b/examples/alerts/alerts.md
@@ -459,7 +459,10 @@ rules:
   annotations:
     message: Thanos Receive {{$labels.job}} has not uploaded latest data to object
       storage.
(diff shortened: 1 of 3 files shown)


brancz:
Docs need regeneration

metalmatze:
Seems like it's also broken if there are multiple tenants per statefulset...

krasi-georgiev:
LGTM

PR #2814, opened 2020-06-29
go.mod: Bump Prometheus to current latest

brancz:

## Changes

Bump TSDB.

## Verification

Tests.

@thanos-io/thanos-maintainers


GiedriusS:
Does this contain some important fixes for our users?

brancz:
Mainly performance improvements, without running it ourselves we don't know how significant those are though. We run latest Thanos in prod all the time, so we want to just get into the habit of bumping TSDB more often so we avoid making big jumps.

brancz:
Oh and it includes https://github.com/prometheus/prometheus/pull/7472 which allows moving forward with https://github.com/thanos-io/thanos/blob/master/docs/proposals/202005_scalable-rule-storage.md

bwplotka:
Looks like a valid test failure @brancz 

It's because in tests we are using label values which are not sorted now, we should use SortedLabelValues instead :hugs: 

Worth to check our TSDB method we use in StoreAPI for TSDB impl, we probably need to sort to be compliant. (if we want)

brancz:
I'm not sure what the consequence of this is, I haven't followed those changes very closely on TSDB. From what I can tell sorted LabelValues is not exposed on the storage interfaces though so I'm not sure how to call them explicitly. Suggestions?

brancz:
Couldn't figure out how to re-run netlify jobs, so re-triggered by amending last commit.

bwplotka:
We can ignore netlify, they have some incident

PR #2817, opened 2020-06-29
receive: Added more observability, fixed leaktest, to actually check leaks ):

bwplotka:
Reason: Missing (), probably we need linter for this.


brancz:
Oops, yeah that's an easy one to do.

bwplotka:
Should be good to go. I assumed that during reload we don't want to close anything (e.g on Flush). Is that ok?

PR #2819, opened 2020-06-30
Update ThanosReceiveNoUpload to select sum == 0

metalmatze:
Right now the alert actually fires in wrong conditions.
We select all sum(instances  == 0) which is always 0...
Instead we want to sum(instances) == 0

Thanks for finding this @krasi-georgiev 


krasi-georgiev:
works as expected in our staging so when tests are passing LGTM

PR #2828, opened 2020-07-01
ci: Manually download promu in crossbuild stage

kakkoyun:

Fix failed crossbuild CI job: https://app.circleci.com/pipelines/github/thanos-io/thanos/2008/workflows/f96cefcb-e91c-4f73-8eca-5c1c20b2ac46/jobs/10115

For context: https://cloud-native.slack.com/archives/CL25937SP/p1593613071221800

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -43,8 +43,10 @@ jobs:
       PROMU_VERSION: "0.5.0"
     steps:
       - checkout
-      - run: curl -L "https://github.com/prometheus/promu/releases/download/v${PROMU_VERSION}/promu-${PROMU_VERSION}.$(go env GOOS)-$(go env GOARCH).tar.gz" | tar --strip-components=1 -xzf - -C ${GOBIN}
       - run: source .bingo/variables.env
+      - run: mkdir -p ${GOBIN}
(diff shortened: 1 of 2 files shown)


.circleci/config.yml
```diff
@@ -40,8 +40,12 @@ jobs:
     working_directory: /home/circleci/.go_workspace/src/github.com/thanos-io/thanos
     environment:
       GOBIN: "/home/circleci/.go_workspace/go/bin"
+      PROMU_VERSION: "0.5.0"
     steps:
       - checkout
+      - run: curl -L "https://github.com/prometheus/promu/releases/download/v${PROMU_VERSION}/promu-${PROMU_VERSION}.$(go env GOOS)-$(go env GOARCH).tar.gz" | tar --strip-components=1 -xzf - -C ${GOBIN}
+      - run: source .bingo/variables.env
```
bwplotka, line 47: You don't want to source those things BEFORE running curl? :thinking:

Makefile
```diff
@@ -113,7 +113,7 @@ build: check-git deps $(PROMU)
 
 .PHONY: crossbuild
 crossbuild: ## Builds all binaries for all platforms.
-crossbuild: deps $(PROMU)
+crossbuild: | $(PROMU)
```
kakkoyun, line 116: Pipe makes sure that dependency doesn't run if the file already exists.
https://www.gnu.org/software/make/manual/html_node/Prerequisite-Types.html

bwplotka: I think even without pipe that is happening. I guess it does not check mod file?

kakkoyun: I have tested locally somehow, it still tries to execute bingo install task. Just to be sure.

PR #2831, opened 2020-07-02
increase ThanosReceiveNoUpload time  check to 3h

krasi-georgiev:


krasi-georgiev:
cc @brancz @metalmatze

PR #2832, opened 2020-07-02
ui: react: Add runtime and build info page

prmsrswt:

## Changes
- Implement API endpoints `/status/runtimeinfo` and `/api/buildinfo`
- Enable the `Runtime & Build Information` page in react UI

## Verification
Tested locally using both production and development build.

## Screenshot
Runtime & Build Info page

### DIFF
--- a/cmd/thanos/query.go
+++ b/cmd/thanos/query.go
@@ -492,10 +492,11 @@ func runQuery(
 	return nil
 }
 
-func runtimeInfo() v1.RuntimeInfo {
+func runtimeInfo(logger log.Logger) v1.RuntimeInfo {
 	cwd, err := os.Getwd()
 	if err != nil {
 		cwd = "<error retrieving current working directory>"
+		level.Warn(logger).Log("msg", "failed to retrieve current working directory", "err", err)
 	}
 
 	status := v1.RuntimeInfo{
--- a/cmd/thanos/query.go
+++ b/cmd/thanos/query.go
@@ -429,10 +429,10 @@ func runQuery(
 			CWD = "<error retrieving current working directory>"
 			level.Warn(logger).Log("msg", "failed to retrieve current working directory", "err", err)
 		}
-		
+
 		birth := time.Now()
 
-		var runtimeInfo v1.RuntimeInfoFn = func (logger log.Logger) v1.RuntimeInfo {
+		var runtimeInfo v1.RuntimeInfoFn = func(logger log.Logger) v1.RuntimeInfo {
 			status := v1.RuntimeInfo{
 				StartTime:      birth,
 				CWD:            CWD,
--- a/cmd/thanos/query.go
+++ b/cmd/thanos/query.go
@@ -424,9 +424,29 @@ func runQuery(
 			GoVersion: version.GoVersion,
 		}
 
+		CWD, err := os.Getwd()
+		if err != nil {
+			CWD = "<error retrieving current working directory>"
+			level.Warn(logger).Log("msg", "failed to retrieve current working directory", "err", err)
+		}
+		
+		birth := time.Now()
+
+		var runtimeInfo v1.RuntimeInfoFn = func (logger log.Logger) v1.RuntimeInfo {
+			status := v1.RuntimeInfo{
+				StartTime:      birth,
+				CWD:            CWD,
+				GoroutineCount: runtime.NumGoroutine(),
+				GOMAXPROCS:     runtime.GOMAXPROCS(0),
+				GOGC:           os.Getenv("GOGC"),
+				GODEBUG:        os.Getenv("GODEBUG"),
+			}
+			return status
+		}
+
 		ins := extpromhttp.NewInstrumentationMiddleware(reg)
 		// TODO(bplotka in PR #513 review): pass all flags, not only the flags needed by prefix rewriting.
-		ui.NewQueryUI(logger, reg, stores, webExternalPrefix, webPrefixHeaderName).Register(router, ins)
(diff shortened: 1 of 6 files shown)


cmd/thanos/query.go
```diff
@@ -470,6 +484,24 @@ func runQuery(
 	return nil
 }
 
+func runtimeInfo() v1.RuntimeInfo {
+	cwd, err := os.Getwd()
+	if err != nil {
+		cwd = "<error retrieving current working directory>"
```
squat, line 490: should we log the error? Or maybe put the error in the returned string?

prmsrswt: I think logging it would make more sense. I got this from https://github.com/thanos-io/thanos/blob/master/pkg/ui/query.go#L47-L50

Should we log the error there too? And I guess the log-level should be `warn`, or do we need to use log-level error for this?

squat: yes warn sounds good IMO, it as it doesn't cause the entire request to fail

prmsrswt: Done :+1:

pkg/query/api/v1.go
```diff
@@ -87,6 +87,26 @@ func (e *ApiError) Error() string {
 	return fmt.Sprintf("%s: %s", e.Typ, e.Err)
 }
 
+// ThanosVersion contains build information about Thanos.
+type ThanosVersion struct {
```
GiedriusS, line 91: Isn't this exactly equivalent to the struct in `pkg/ui/query.go`? Maybe let's re-use it?

cmd/thanos/query.go
```diff
@@ -431,6 +443,8 @@ func runQuery(
 			flagsMap,
 			instantDefaultMaxSourceResolution,
 			maxConcurrentQueries,
+			runtimeInfo,
+			buildInfo,
```
GiedriusS, line 447: Is there a reason to have this as an argument? Sounds like YAGNI. But, instead of moving this down to where it is actually needed, it's probably an even better idea to re-use this variable and pass it to `NewQueryUI()` as well, down to `q.status()` where `thanosVersion` is used. Then we would ensure that the same information is available in both UIs.

pkg/query/api/v1.go
prmsrswt: Yeah, sure.

cmd/thanos/query.go
prmsrswt: Oh got it. So if we pass the `runtimeInfo` and `buildInfo` to `NewQueryUI()` then which `thanosVersion` struct should we remove. The one in /pkg/ui/query.go` or should I remove the one I added as you said before?

prmsrswt:
> how is this not a user-facing change?
Oh I thought the new UI is still in beta :sweat_smile: But yeah, maybe we should add an entry in `CHANGELOG` for this.

cmd/thanos/query.go
GiedriusS: I suppose let's remove the old one since it is now a part of the API, and then let's make those two variables in `cmd/thanos/query.go` and pass them down to where they are used to ensure that we don't do duplicate work

prmsrswt: Sure

cmd/thanos/query.go
```diff
@@ -478,6 +492,25 @@ func runQuery(
...
+func runtimeInfo(logger log.Logger) v1.RuntimeInfo {
+	cwd, err := os.Getwd()
+	if err != nil {
+		cwd = "<error retrieving current working directory>"
+		level.Warn(logger).Log("msg", "failed to retrieve current working directory", "err", err)
+	}
+
+	status := v1.RuntimeInfo{
+		StartTime:      time.Now().UTC(),
+		CWD:            cwd,
```
prmsrswt, line 504: Oops! There is a bug here that completely slipped past me. Every time the API gets a request for status, we run `runtimeinfo()` to get updated values for `GoroutineCount` etc. but we are also setting `StartTime` to `time.now().UTC()` So we are sending current time instead of the start time :open_mouth: 

We can make this a method of the API itself like Prometheus is doing it here https://github.com/prometheus/prometheus/blob/98a69b77d1eda71c20836cbebf83b956c68de3a9/web/web.go#L817 but then, I am not sure how can we expose this info to `NewQueryUI()` as well.

@GiedriusS @squat wdyt?

GiedriusS: Sorry for the late response. What about making the function `runtimeInfo` in-line that "catches" `CWD` (not necessarily but it will most likely never change, right?) and `Birth` from the environment, and then passing it to both `ui.NewQueryUI()` and `v1.NewAPI()`? Something like this: https://play.golang.org/p/KG6Rf_vns32.

cmd/thanos/query.go
```diff
@@ -412,9 +415,38 @@ func runQuery(
...
+		var runtimeInfo v1.RuntimeInfoFn = func() v1.RuntimeInfo {
+			status := v1.RuntimeInfo{
+				StartTime:      birth,
+				CWD:            CWD,
+				GoroutineCount: runtime.NumGoroutine(),
+				GOMAXPROCS:     runtime.GOMAXPROCS(0),
+				GOGC:           os.Getenv("GOGC"),
+				GODEBUG:        os.Getenv("GODEBUG"),
+			}
+			return status
```
GiedriusS, line 444: We could just return this struct directly, no need for a variable. But I'd guess that the compiler already does this optimization for us. Plus, I'm surprised that none of our linters caught this. Perhaps we could improve the linters to catch functions that look like:
```go
func foo() test {
  a := test{}
  return a
}
```

In the distant future? :smile:

PR #2835, opened 2020-07-02
ui: react: Add a dropdown for selecting max source resolution

prmsrswt:

## Changes
- Add a dropdown in the new React UI to select max source resolution

## Verification
Tested locally and verified the fetch requests


pkg/ui/react-app/src/pages/graph/GraphControls.tsx
```diff
@@ -142,6 +144,19 @@ class GraphControls extends Component<GraphControlsProps> {
             <FontAwesomeIcon icon={faChartArea} fixedWidth />
           </Button>
         </ButtonGroup>
+
+        <Input
+          type="select"
+          value={this.props.maxSourceResolution}
+          onChange={e => this.props.onChangeMaxSourceResolution(e.target.value)}
```
squat, line 151: in general we try to avoid binding callbacks at render time with the arrow function in order to not trigger a diff. Do you think this is a small enough component that we can afford a diff? Or should we pre-bind?

prmsrswt: I was not aware of this. By pre-binding do you mean something like this?
```js
constructor(props) {
  super(props);

  this.handleMaxSourceResChange = this.handleMaxSourceResChange.bind(this);
}

handleMaxSourceResChange(event) {
  this.props.onChangeMaxSourceResolution(event.target.value)
}

render(
  <Input onChange={this.handleMaxSourceResChange} />
)
```
Sorry for the silly questions, I don't have much experience with class-based components.

cmd/thanos/query.go
```diff
@@ -433,15 +433,14 @@ func runQuery(
 		birth := time.Now()
 
 		var runtimeInfo v1.RuntimeInfoFn = func() v1.RuntimeInfo {
-			status := v1.RuntimeInfo{
+			return v1.RuntimeInfo{
```
GiedriusS, line 436: Thank you for this :+1:

pkg/ui/react-app/src/pages/graph/GraphControls.tsx
```diff
@@ -142,6 +154,19 @@ class GraphControls extends Component<GraphControlsProps> {
             <FontAwesomeIcon icon={faChartArea} fixedWidth />
           </Button>
         </ButtonGroup>
+
+        <Input
+          type="select"
+          value={this.props.maxSourceResolution}
+          onChange={this.handleMaxSourceResChange}
```
squat, line 161: :+1: perfect

Issue #2846, opened 2020-07-06
Query: /api/v1/stores panics when store is healthy - v0.14.0-rc.0


cuotos:
Thanos v0.14.0-rc.0

When query is connected to a healthy Store, viewing /stores shows 
![image](https://user-images.githubusercontent.com/3035161/86639250-88f3c000-bfd0-11ea-8cd7-8118f7c63ef9.png)

accessing /api/v1/stores causes a panic. 

Issue was introduced in https://github.com/thanos-io/thanos/pull/2809. 

When nil error is passed to `updateStoreStatus(store *storeRef, err error)` results in a nil pointer here
`status.LastError = &stringError{originalErr: err}`

I've got a PR that i've tried to unit tests the entire function so should catch this. 

I do not know all the details of how this worked before, I can only assume that *MOST of the errors that get passed to this will have custom Marshal methods, and this fix was to catch those errors that do not, as these will normally Marshal to {}.

Let me know if I've mis understood this.

PR #2847, opened 2020-07-06
Query: correctly marshal errors to JSON and ignore if nil

cuotos:

## Changes

Fixes: https://github.com/thanos-io/thanos/issues/2846
Passing nil error to updateStoreStatus causes a panic. This is displayed to the user in the Stores ui when the Store is healthy

Added higher level unit test to catch this, and hopefully clarified existing test.


cuotos:
thought id completely message up my testing, panicked and closed this. turns out i had another instance running.... have created a new PR

PR #2852, opened 2020-07-07
receive: Track replications

kakkoyun:
This PR adds a new metric to track replication behaviour. Previously, `thanos_receive_forward_requests_total` was used. However, that metric is now misleading because we have recently changed how replication behaves (https://github.com/thanos-io/thanos/pull/2621).

## Changes

* Add a new metric to track replicated request `thanos_receive_replications_total` https://github.com/thanos-io/thanos/pull/2852/commits/fd0497105ab76fdc2cbcfd0c74f4a642f389ec45
* Add alerts and dashboards for the new metric https://github.com/thanos-io/thanos/pull/2852/commits/0aa495da1bb6e44c56c2653c55e8c4bc70b1e62c
* Modify alerts for the `thanos_forward_requests_total` https://github.com/thanos-io/thanos/pull/2852/commits/0aa495da1bb6e44c56c2653c55e8c4bc70b1e62c

## Verification

* `make test-local`
* `make examples`
* `make example-rules-lint`

cc @krasi-georgiev

### DIFF
--- a/examples/alerts/rules.yaml
+++ b/examples/alerts/rules.yaml
@@ -7,7 +7,6 @@ groups:
       /
         sum(rate(grpc_client_started_total{job=~"thanos-query.*", grpc_type="unary"}[5m]))
       )
-    labels: {}
     record: :grpc_client_failures_per_unary:sum_rate
   - expr: |
       (
@@ -15,7 +14,6 @@ groups:
       /
         sum(rate(grpc_client_started_total{job=~"thanos-query.*", grpc_type="server_stream"}[5m]))
       )
-    labels: {}
     record: :grpc_client_failures_per_stream:sum_rate
   - expr: |
       (
@@ -23,7 +21,6 @@ groups:
       /
         sum(rate(thanos_querier_store_apis_dns_lookups_total{job=~"thanos-query.*"}[5m]))
       )
-    labels: {}
     record: :thanos_querier_store_apis_dns_failures_per_lookup:sum_rate
   - expr: |
       histogram_quantile(0.99,
@@ -47,7 +44,6 @@ groups:
       /
         rate(grpc_server_started_total{job=~"thanos-receive.*", grpc_type="unary"}[5m])
       )
-    labels: {}
     record: :grpc_server_failures_per_unary:sum_rate
   - expr: |
       sum(
@@ -55,7 +51,6 @@ groups:
       /
         rate(grpc_server_started_total{job=~"thanos-receive.*", grpc_type="server_stream"}[5m])
       )
-    labels: {}
     record: :grpc_server_failures_per_stream:sum_rate
   - expr: |
       sum(
@@ -63,7 +58,6 @@ groups:
       /
         rate(http_requests_total{handler="receive", job=~"thanos-receive.*"}[5m])
       )
-    labels: {}
     record: :http_failure_per_request:sum_rate
   - expr: |
       histogram_quantile(0.99,
@@ -78,7 +72,6 @@ groups:
       /
         sum(rate(thanos_receive_replications_total{job=~"thanos-receive.*"}[5m]))
       )
-    labels: {}
     record: :thanos_receive_replication_failure_per_requests:sum_rate
   - expr: |
       (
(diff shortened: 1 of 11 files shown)


pkg/receive/handler.go
```diff
@@ -100,12 +100,18 @@ func NewHandler(logger log.Logger, o *Options) *Handler {
...
+		forwardRequests: promauto.With(o.Registry).NewCounterVec(
 			prometheus.CounterOpts{
 				Name: "thanos_receive_forward_requests_total",
 				Help: "The number of forward requests.",
 			}, []string{"result"},
 		),
+		replications: promauto.With(o.Registry).NewCounterVec(
+			prometheus.CounterOpts{
+				Name: "thanos_receive_replications_total",
+				Help: "The number of replication operation.",
```
bwplotka, line 112:
```suggestion
				Help: "The number of replication operations.",
```

bwplotka: I would also add more info like
```suggestion
				Help: "The number of replication operations started by the receiver. The success of replication is fulfilled when a quorum is met",
```

bwplotka:
In other words what I mean is that `http_requests_total{handler="receive"}`  is really the same as our replication metric you add here no?

kakkoyun:
@bwplotka Yes, eventually that's the result. We can use that metric instead of adding a new one. I just wanted to be more explicit about the metric.

> I think what would fix this problem would be actually starting new context for forwarding and then cancel it only if timeout OR ((not yet success or failure) && parent context is cancelled). WDYT?

For me, this is just a misleading metric issue. Rather than detaching the forward requests from the incoming request's context, I would suggest to catch this case and handle it different than an error.

bwplotka:
You cannot catch cancellation in any other way than having separate context  ):

bwplotka:
It's not error handling, it's premature context cancel.

kakkoyun:
@bwplotka @squat Closing this one. 

We have talked with @bwplotka offline and decided to fix it in a simpler manner. I'll send another PR.

kakkoyun:
@squat has pointed out offline that we have multiple replication batch operations per request so `http_requests_total{handler="receive"}` is not enough to do the trick. Thus opened this again.

examples/alerts/rules.yaml
```diff
@@ -72,6 +72,14 @@ groups:
     labels:
       quantile: "0.99"
     record: :http_request_duration_seconds:histogram_quantile
+  - expr: |
+      (
+        sum(rate(thanos_receive_replications_total{result="error", job=~"thanos-receive.*"}[5m]))
+      /
+        sum(rate(thanos_receive_replications_total{job=~"thanos-receive.*"}[5m]))
+      )
+    labels: {}
```
brancz, line 81: nit: no need for this line if we're not adding any labels

pkg/receive/handler.go
```diff
@@ -100,12 +100,18 @@ func NewHandler(logger log.Logger, o *Options) *Handler {
...
+		forwardRequests: promauto.With(o.Registry).NewCounterVec(
 			prometheus.CounterOpts{
 				Name: "thanos_receive_forward_requests_total",
 				Help: "The number of forward requests.",
 			}, []string{"result"},
 		),
+		replications: promauto.With(o.Registry).NewCounterVec(
+			prometheus.CounterOpts{
+				Name: "thanos_receive_replications_total",
+				Help: "The number of replication operations started by the receiver. The success of replication is fulfilled when a quorum is met.",
```
bwplotka, line 112:
```suggestion
				Help: "The number of replication operations done by the receiver. The success of replication is fulfilled when a quorum is met.",
```

pkg/receive/handler.go
```diff
@@ -374,16 +380,21 @@ func (h *Handler) fanoutForward(ctx context.Context, tenant string, replicas map
...
 				tracing.DoInSpan(ctx, "receive_replicate", func(ctx context.Context) {
 					err = h.replicate(ctx, tenant, wreqs[endpoint])
 				})
 				if err != nil {
+					h.replications.WithLabelValues("error").Inc()
 					ec <- errors.Wrapf(err, "replicate write request, endpoint %v", endpoint)
 					return
 				}
+
+				h.replications.WithLabelValues("success").Inc()
```
bwplotka, line 394: I would suggest adding `success` and `error` to constant vars so we can reduce chance of typo increasing cardinality and creating massive confusion ;p

PR #2854, opened 2020-07-07
receive: Detach fanout context from request context

kakkoyun:

This PR makes the best-effort fan-outs more explicit by detaching the fan-out context requests' context from the incoming request context. 

## Changes

* Detach the fan-out context requests' context from the incoming request context. Make the best-effort fan-outs more explicit.
* Removed redundant `govet` call, `golangci-lint` already runs it 0f5b22556a7827f4246435a4145e8c311b278cb3

## Verification

* `make test`

### DIFF
--- a/pkg/receive/handler.go
+++ b/pkg/receive/handler.go
@@ -357,6 +357,18 @@ func (h *Handler) writeQuorum() int {
 // fanoutForward fans out concurrently given set of write requests. It returns status immediately when quorum of
 // requests succeeds or fails or if context is canceled.
 func (h *Handler) fanoutForward(ctx context.Context, tenant string, replicas map[string]replica, wreqs map[string]*prompb.WriteRequest, successThreshold int) error {
+	var errs terrors.MultiError
+
+	ctx, cancel := context.WithTimeout(tracing.CopyTraceContext(context.Background(), ctx), h.options.ForwardTimeout)
+	defer func() {
+		if errs != nil {
+			// NOTICE: The cancel function is not used on all paths intentionally,
+			// if there is no error when quorum successThreshold is reached,
+			// let forward requests to optimistically run until timeout.
+			cancel()
+		}
+	}()
+
 	logger := log.With(h.logger, "tenant", tenant)
 	if id, ok := middleware.RequestIDFromContext(ctx); ok {
 		logger = log.With(logger, "request-id", id)
@@ -397,7 +409,7 @@ func (h *Handler) fanoutForward(ctx context.Context, tenant string, replicas map
 			go func(endpoint string) {
 				defer wg.Done()
 				var err error
-				tracing.DoInSpan(ctx, "receive_tsdb_write", func(ctx context.Context) {
+				tracing.DoInSpan(ctx, "receive_tsdb_write", func(_ context.Context) {
 					err = h.writer.Write(tenant, wreqs[endpoint])
 				})
 				if err != nil {
@@ -478,10 +490,7 @@ func (h *Handler) fanoutForward(ctx context.Context, tenant string, replicas map
 		}()
 	}()
 
-	var (
-		success int
-		errs    terrors.MultiError
-	)
+	var success int
 	for {
 		select {
 		case <-ctx.Done():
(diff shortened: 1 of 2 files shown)


pkg/receive/handler.go
```diff
@@ -533,28 +533,39 @@ func (h *Handler) replicate(ctx context.Context, tenant string, wreq *prompb.Wri
...
-	ctx, cancel := context.WithTimeout(ctx, h.options.ForwardTimeout)
-	defer func() {
-		// If there is no error, let forward requests optimistically run until timeout.
+	var (
+		quorum       = h.writeQuorum()
+		fctx, cancel = context.WithTimeout(context.Background(), h.options.ForwardTimeout)
+		errChan      = make(chan error)
+	)
+	go func() {
+		close(errChan)
```
bwplotka, line 542: Closing before doing anything sounds wrong (: You mean defer here?

bwplotka: Also I  think we need to defer from main go routine, otherwise we will deadlock on write to errChan

kakkoyun: 🙈 It's what I meant. 
Yeah, I'm gonna fix it.

kakkoyun:
Linter fails because of a missing `.` at the end of a comment (`//nolint:govet`). And we need this directive because we are intentionally leaking a context to make fan-out requests best effort. 

I have plans to improve our CI for the comment-trailing-dot check.

kakkoyun:
Merging https://github.com/thanos-io/thanos/pull/2859 would solve the CI problem.

kakkoyun:
Removed redundant `govet` call, `golangci-lint` already runs it 0f5b22556a7827f4246435a4145e8c311b278cb3

pkg/receive/handler.go
```diff
@@ -533,28 +533,39 @@ func (h *Handler) replicate(ctx context.Context, tenant string, wreq *prompb.Wri
...
 
-	quorum := h.writeQuorum()
-	err = h.fanoutForward(ctx, tenant, replicas, wreqs, quorum)
-	if countCause(err, isNotReady) >= quorum {
-		return tsdb.ErrNotReady
-	}
-	if countCause(err, isConflict) >= quorum {
-		return errors.Wrap(conflictErr, "did not meet success threshold due to conflict")
+		// If there is no error, quorum reached, let forward requests optimistically run until timeout.
+		return nil //nolint:govet
```
brancz, line 564: I don't think I understand why the `nolint` comment is necessary here? The `err != nil` branch is already covered, and will return no matter what the `err` value is. I feel like I'm missing something.

kakkoyun: `govet` here complains that for this return branch the context is not cancelled.

brancz: I see, can you explain exactly this in the comment? The connection between the comment above it and the nolint wasn't obvious to me (maybe it should have been but I would guess that future me would make the same mistake)

kakkoyun: Added comments to explain the reasoning behind `nolint`s.

brancz: I apologize for being picky on this, but for the sake of it being extremely obvious, can you add to the comments something along the lines of "which is why the no-lint comment is required and intentional here."

brancz:
(another re-review from @bwplotka would be great though)

pkg/receive/handler.go
```diff
@@ -533,28 +533,45 @@ func (h *Handler) replicate(ctx context.Context, tenant string, wreq *prompb.Wri
...
-	if countCause(err, isNotReady) >= quorum {
-		return tsdb.ErrNotReady
-	}
-	if countCause(err, isConflict) >= quorum {
-		return errors.Wrap(conflictErr, "did not meet success threshold due to conflict")
+		// If there is no error, quorum reached, let forward requests to optimistically run until timeout.
+		// NOTICE: This return statement may be reached without using the cancel var defined on line 540,
+		// this is the intended behavior since we want to forward requests to optimistically run until timeout,
+		// which is why the nolint comment is required and intentional here.
+		return nil //nolint:govet
```
bwplotka, line 570: so... why nolint is here if we removed go vet?

kakkoyun: We didn't remove the `go vet`. We removed redundant `go vet`. `golangci-lint` runs it already, we were running it again as a separate command.

kakkoyun:
Alright, let me try to find out a cleaner approach and maybe pushing down the logic in `fanoutForward`. Stay tuned 😄

pkg/receive/handler.go
```diff
@@ -357,6 +357,18 @@ func (h *Handler) writeQuorum() int {
 // fanoutForward fans out concurrently given set of write requests. It returns status immediately when quorum of
 // requests succeeds or fails or if context is canceled.
 func (h *Handler) fanoutForward(ctx context.Context, tenant string, replicas map[string]replica, wreqs map[string]*prompb.WriteRequest, successThreshold int) error {
+	var errs terrors.MultiError
+
+	ctx, cancel := context.WithTimeout(tracing.CopyTraceContext(context.Background(), ctx), h.options.ForwardTimeout)
```
kakkoyun, line 362: Having timeout here rather than `replicate` is actually semantically more sound to cover forward request timeouts without replication enabled.

bwplotka, line 362: Yes. Make sense. Can we not shadow ctx variable for verbocity? :hugs:

pkg/receive/handler.go
```diff
@@ -357,6 +357,18 @@ func (h *Handler) writeQuorum() int {
...
+	defer func() {
+		if errs != nil {
```
bwplotka, line 364: Wonder if `errs.Err()` is not safer

pkg/receive/handler.go
```diff
@@ -356,9 +356,21 @@ func (h *Handler) writeQuorum() int {
 
 // fanoutForward fans out concurrently given set of write requests. It returns status immediately when quorum of
 // requests succeeds or fails or if context is canceled.
-func (h *Handler) fanoutForward(ctx context.Context, tenant string, replicas map[string]replica, wreqs map[string]*prompb.WriteRequest, successThreshold int) error {
+func (h *Handler) fanoutForward(pctx context.Context, tenant string, replicas map[string]replica, wreqs map[string]*prompb.WriteRequest, successThreshold int) error {
+	var errs terrors.MultiError
+
+	fctx, cancel := context.WithTimeout(tracing.CopyTraceContext(context.Background(), pctx), h.options.ForwardTimeout)
```
brancz, line 362: > tracing.CopyTraceContext

Neat! I had not known about this one previously!

PR #2858, opened 2020-07-08
Fixed --store.grpc.series-sample-limit

### LINKED ISSUE #2845: Thanos store: improve series-sample-limit
The Thanos store supports the `-store.grpc.series-sample-limit` CLI flag, whose description is:
> Maximum amount of samples returned via a single Series call. 0 means no limit. NOTE: For efficiency we take 120 as the number of samples in chunk (it cannot be bigger than that), so the actual number of samples might be lower, even though the maximum could be hit.

However, looking at the implementation, it doesn't limit the "samples returned via a single Series call" but the samples read from each block. If a single `Series()` call fetch samples from N blocks, the limit is checked for each block and the actual number of samples returned could be up to `N * limit` (where `N` is the number of blocks queried).

I would like to propose to improve it. 

### Proposal

I propose to **internally** change `SampleLimiter` into `ChunksLimiter`, limiting the number of chunks instead of samples (which is how it's actually implemented, because the number of samples per chunks is just estimated). The `ChunksLimiter` interface is as follow:

```
type ChunksLimiter interface {
        // Reserve num chunks out of the total number of chunks.
	Reserve(num uint64) error
}
```

The `Reserve()` function increases by `num` the number of chunks fetched, so that multiple calls to `Reserve()` (one for each block) will increase the total number of chunks fetched until it will eventually hit the limit. A **new limiter is created for each `Series()` call**, so that limits are isolated to individual `Series()` calls.

The `-store.grpc.series-sample-limit` flag value (if > 0) will be used to compute the actual number of chunks to limit to, which is `maxChunks = (maxSampleCount / maxSamplesPerChunk) + 1`.

pracucci:

## Changes

In this PR I've changed the `--store.grpc.series-sample-limit` implementation to apply the limit to the sum of "samples" fetched across all queried blocks instead of individually apply the limit to each block.

Few notes:
- Internally implemented as a chunks limiter (because that's what we limit on)
- Introduced `ChunksLimiterFactory` to not pass a static limit to `NewBucketStore` because I would like to use it in Cortex too and in Cortex we have dynamic limits (limits can be reloaded live, without restarting the process)

Fixes #2845.

## Verification

Existing + new unit tests.

### DIFF
--- a/pkg/store/bucket.go
+++ b/pkg/store/bucket.go
@@ -54,13 +54,13 @@ import (
 )
 
 const (
-	// maxSamplesPerChunk is approximately the max number of samples that we may have in any given chunk. This is needed
+	// MaxSamplesPerChunk is approximately the max number of samples that we may have in any given chunk. This is needed
 	// for precalculating the number of samples that we may have to retrieve and decode for any given query
 	// without downloading them. Please take a look at https://github.com/prometheus/tsdb/pull/397 to know
 	// where this number comes from. Long story short: TSDB is made in such a way, and it is made in such a way
 	// because you barely get any improvements in compression when the number of samples is beyond this.
 	// Take a look at Figure 6 in this whitepaper http://www.vldb.org/pvldb/vol8/p1816-teller.pdf.
-	maxSamplesPerChunk = 120
+	MaxSamplesPerChunk = 120
 	maxChunkSize       = 16000
 	maxSeriesSize      = 64 * 1024
 
@@ -240,9 +240,9 @@ type BucketStore struct {
 	// Query gate which limits the maximum amount of concurrent queries.
 	queryGate gate.Gate
 
-	// samplesLimiter limits the number of samples per each Series() call.
-	samplesLimiter SampleLimiter
-	partitioner    partitioner
+	// chunksLimiterFactory creates a new limiter used to limit the number of chunks fetched by each Series() call.
+	chunksLimiterFactory ChunksLimiterFactory
+	partitioner          partitioner
 
 	filterConfig             *FilterConfig
 	advLabelSets             []storepb.LabelSet
@@ -269,7 +269,7 @@ func NewBucketStore(
 	indexCache storecache.IndexCache,
 	queryGate gate.Gate,
 	maxChunkPoolBytes uint64,
-	maxSampleCount uint64,
+	chunksLimiterFactory ChunksLimiterFactory,
 	debugLogging bool,
 	blockSyncConcurrency int,
 	filterConfig *FilterConfig,
@@ -287,7 +287,6 @@ func NewBucketStore(
 		return nil, errors.Wrap(err, "create chunk pool")
 	}
 
-	metrics := newBucketStoreMetrics(reg)
 	s := &BucketStore{
 		logger:                      logger,
 		bkt:                         bkt,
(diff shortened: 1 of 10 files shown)


pkg/store/bucket.go
```diff
@@ -722,12 +721,16 @@ func blockSeries(
 			s.refs = append(s.refs, meta.Ref)
 		}
 		if len(s.chks) > 0 {
+			if err := chunksLimiter.Reserve(uint64(len(s.chks))); err != nil {
```
bwplotka, line 724: We could potentially limit chunk bytes instead TBH, but this is good for a start

pkg/store/limiter.go
```diff
@@ -4,19 +4,31 @@
...
+
 	"github.com/pkg/errors"
 	"github.com/prometheus/client_golang/prometheus"
 )
 
-type SampleLimiter interface {
-	Check(num uint64) error
+type ChunksLimiter interface {
+	// Reserve num chunks out of the total number of chunks enforced by the limiter.
+	// Returns an error if the limit has been exceeded. This function must be multi
```
bwplotka, line 16:
```suggestion
	// Returns an error if the limit has been exceeded. This function must be
```

bwplotka: We always refer as to `goroutine safe` same as `"thread safe"`

pkg/store/limiter.go
```diff
@@ -4,19 +4,31 @@
...
+type ChunksLimiterFactory func(failedCounter prometheus.Counter) ChunksLimiter
+
 // Limiter is a simple mechanism for checking if something has passed a certain threshold.
 type Limiter struct {
-	limit uint64
+	limit    uint64
+	reserved uint64
 
-	// Counter metric which we will increase if Check() fails.
+	// Counter metric which we will increase if limit is exceeded.
```
bwplotka, line 30: I guess this line was suppose to be below failedOnce

pkg/store/limiter.go
```diff
@@ -25,14 +37,23 @@ func NewLimiter(limit uint64, ctr prometheus.Counter) *Limiter {
...
 	if l.limit == 0 {
 		return nil
 	}
-	if num > l.limit {
-		l.failedCounter.Inc()
-		return errors.Errorf("limit %v violated (got %v)", l.limit, num)
+	if reserved := atomic.AddUint64(&l.reserved, num); reserved > l.limit {
+		// We need to protect from the counter being incremented twice due to concurrency
+		// while calling Reserve().
+		l.failedOnce.Do(l.failedCounter.Inc)
```
bwplotka, line 48: :+1:

pkg/store/bucket.go
pracucci: Or both. In Cortex we would need by count :)

pracucci:
Thanks @bwplotka for your feedback. I should have addressed the comments and fixed the doc.

pkg/store/bucket.go
pracucci: Opened an issue: https://github.com/thanos-io/thanos/issues/2861

bwplotka:
Looks like my manual rebase through Github UI (there were conflicts) was not successful..  :cry: I tried ;p

bwplotka:
Thanks!

xiaozongyang:
Hi @pracucci,
I'm using this feature in my production environment, thx for you great job!

And I'm wondering that do we have some metrics about loaded/fetched samples per-query? So we can determine a query is rejected by this series limit or other reason.

PR #2862, opened 2020-07-08
ci: Enable a set of linters to catch unused code

kakkoyun:

## Changes

* Enabled  `unparam`, `unused` and `varcheck`
* Applied fixes accordingly
* Added notes for more complicated or critical ones for now

## Verification

* `make go-lint`

### DIFF
--- a/.golangci.yml
+++ b/.golangci.yml
@@ -43,6 +43,9 @@ linters:
     - unused
     - varcheck
     - godot
+    - unparam
+    - unused
+    - varcheck
 
 linters-settings:
   errcheck:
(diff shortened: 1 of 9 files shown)


.golangci.yml
```diff
@@ -43,6 +43,9 @@ linters:
     - unused
     - varcheck
     - godot
+    - unparam
+    - unused
+    - varcheck
```
pracucci, line 48: These two were already in the list, a couple of entries above.

kakkoyun: 🙈 I messed up something while rebasing, will fix it.

PR #2871, opened 2020-07-10
Add comment about prometheus sync with cortex

sylr:
## Changes

As Thanos depends on Cortex and Cortex depends of Thanos and both depends on Prometheus, upgrading prometheus on one side can cause build failures on the other side if breaking changes are shipped within the new version of prometheus.

I believe that syncing the upgrades of prometheus between Thanos & Cortex will make it more manageable.

CC https://github.com/cortexproject/cortex/pull/2864

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -47,6 +47,8 @@ require (
 	github.com/prometheus/client_golang v1.6.0
 	github.com/prometheus/client_model v0.2.0
 	github.com/prometheus/common v0.10.0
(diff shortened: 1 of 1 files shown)


bwplotka:
Well, comment is not needed, we just have to agree on that on both projects. 

Or actually much better: Create automation job for this!

sylr:
Automation Job would be super nice.

bwplotka:
Cool let's think about automation there. Dependabot could be configured here.

PR #2883, opened 2020-07-11
Use modification time to filter metas in ConsistencyDelayMetaFilter

### LINKED ISSUE #2585: ConsistencyDelayMetaFilter should filter on the block upload time and not the block creation time
The `ConsistencyDelayMetaFilter` filter currently filters out blocks based on the ULID time, which is the time when the block has been created:
https://github.com/thanos-io/thanos/blob/cfced59c652aca5c311baa22b519e107a20ba422/pkg/block/fetcher.go#L727

This means that if that actual consistency delay is the configured delay minus the upload time. If the delay is X but the block upload time is > X, the consistency delay is effectively ignored.

To correct approach would be checking the consistency delay against the "block upload completed timestamp" and, to do so, checking the `meta.json` creation time could be fine. However, object stores stores the "updated time" and not the "creation time", so I'm wondering how safe would be using it. Thoughts?

yeya24:

## Changes

Fixes #2585 

I use metafile's modification to do filtering.

## Verification

Unit tests updated.


yeya24:
Umm, I got what's wrong. Need to update E2E tests as well https://github.com/thanos-io/thanos/blob/master/test/e2e/store_gateway_test.go#L81!

But it seems not easy to change modification time in E2E tests. Any ideas?

pkg/block/fetcher.go
```diff
@@ -706,10 +706,16 @@ func (r *ReplicaLabelRemover) Modify(_ context.Context, metas map[ulid.ULID]*met
...
 	consistencyDelay time.Duration
+	bkt              objstore.InstrumentedBucketReader
 }
 
 // NewConsistencyDelayMetaFilter creates ConsistencyDelayMetaFilter.
-func NewConsistencyDelayMetaFilter(logger log.Logger, consistencyDelay time.Duration, reg prometheus.Registerer) *ConsistencyDelayMetaFilter {
+func NewConsistencyDelayMetaFilter(
+	logger log.Logger,
+	consistencyDelay time.Duration,
+	bkt objstore.InstrumentedBucketReader,
```
bwplotka, line 716: If we want we could just add interface for Attributes so the usability is scoped here (it's important to know if component has access to writing/reading blocks or not.

pkg/block/fetcher.go
```diff
@@ -723,16 +729,27 @@ func NewConsistencyDelayMetaFilter(logger log.Logger, consistencyDelay time.Dura
...
 		logger:           logger,
 		consistencyDelay: consistencyDelay,
+		bkt:              bkt,
 	}
 }
 
 // Filter filters out blocks that filters blocks that have are created before a specified consistency delay.
-func (f *ConsistencyDelayMetaFilter) Filter(_ context.Context, metas map[ulid.ULID]*metadata.Meta, synced *extprom.TxGaugeVec) error {
+func (f *ConsistencyDelayMetaFilter) Filter(ctx context.Context, metas map[ulid.ULID]*metadata.Meta, synced *extprom.TxGaugeVec) error {
+	// If consistencyDelay is 0, skip filtering.
```
bwplotka, line 738: I think this comment is not needed this is clear :+1:

pkg/block/fetcher.go
```diff
@@ -723,16 +729,27 @@ func NewConsistencyDelayMetaFilter(logger log.Logger, consistencyDelay time.Dura
...
+		attributes, err := f.bkt.Attributes(ctx, metaFile)
+		if err != nil {
+			return errors.Wrapf(err, "get object attributes of %s", metaFile)
+		}
+
 		// TODO(khyatisoneji): Remove the checks about Thanos Source
 		//  by implementing delete delay to fetch metas.
-		// TODO(bwplotka): Check consistency delay based on file upload / modification time instead of ULID.
-		if ulid.Now()-id.Time() < uint64(f.consistencyDelay/time.Millisecond) &&
+		if ulid.Now()-ulid.Timestamp(attributes.LastModified) < uint64(f.consistencyDelay.Milliseconds()) &&
```
bwplotka, line 752: :call_me_hand:

pkg/block/fetcher_test.go
```diff
@@ -1033,7 +1113,11 @@ func TestConsistencyDelayMetaFilter_Filter_0(t *testing.T) {
 		m := newTestFetcherMetrics()
 		expected := map[ulid.ULID]*metadata.Meta{}
 		// Only certain sources and those with 30m or more age go through.
-		for i, id := range u.created {
+		for i, obj := range objectsWithAttrs {
+			// obj.Name is Ulid/meta.json, the block id is the dir name
```
bwplotka, line 1117: missing trailing period

pkg/block/fetcher_test.go
```diff
@@ -949,79 +949,159 @@ func compareSliceWithMapKeys(tb testing.TB, m map[ulid.ULID]*metadata.Meta, s []
...
+		},
+		{
+			&metadata.Meta{Thanos: metadata.Thanos{Source: metadata.BucketRepairSource}},
+			objstore.ObjectAttributes{LastModified: now.Add(-2 * time.Hour)},
+		},
+	}
+
+	// input is a metadata map, works as an input parameter for Filter method
+	input := make(map[ulid.ULID]*metadata.Meta, len(objs))
+	// objectsWithAttrs is an array of object content and attributes, used for buckets mocking
```
bwplotka, line 1079: ditto and above

pkg/objstore/inmem.go
```diff
@@ -26,6 +26,14 @@ type InMemBucket struct {
 	attrs   map[string]ObjectAttributes
 }
 
+// ObjectWithAttrs is a struct defines object name, content and attributes.
+// Can be used together with NewInMemBucketWithObjectsAndAttrs for mocking.
+type ObjectWithAttrs struct {
```
bwplotka, line 31: Do we need this? We could just add pass a map[string][]byte and build attributes automatically, no? (:

pracucci:
> Actually @pracucci reminded me that we might want to cache last modified state on meta json discovery in memory before we merge this 🤗

To elaborate a bit more:

Without memcached attributes caching, it will do a bunch of Bucket API calls. If we would store the `LastModified` in the `metadata.Meta`, we would load it only once, when the block is discovered, and then the `baseFetcher` will always reuse the cached version.

And also with memcached attributes caching, it does unnecessary calls there (think about a multi-tenant cluster with a tens or hundred of thousands of blocks). 

Storing the `LastModified` in the `metadata.Meta` may be a bit of a design stretch, so I'm not sure it's the right approach. Better ideas are welcome.

pracucci:
> Storing the LastModified in the metadata.Meta may be a bit of a design stretch, so I'm not sure it's the right approach. Better ideas are welcome.

However, if the `LastModified` would be available in the `metadata.Meta` we would be able to leverage on it in the query results consistency check done in Cortex, where we're currently employing a custom solution due to the lack of it.

pkg/objstore/inmem.go
yeya24: Sorry, I don't understand the point. What do you mean by `build attributes automatically`? I need to mock the time, so passing attributes is necessary I think,

pkg/block/fetcher.go
yeya24: I agree. Umm, I cannot name this interface :rofl:  `AttrReader`?

pkg/objstore/inmem.go
yeya24: Umm, if you were talking about the case that lastModified time is added in meta.json, then it is good to just pass `[]byte` here.

pkg/block/fetcher.go
```diff
@@ -735,21 +764,26 @@ func NewConsistencyDelayMetaFilter(
...
 	}
 
 	for id, meta := range metas {
-		metaFile := path.Join(id.String(), MetaFilename)
-		attributes, err := f.bkt.Attributes(ctx, metaFile)
-		if err != nil {
-			return errors.Wrapf(err, "get object attributes of %s", metaFile)
+		lastModifiedTime := meta.Thanos.LastModified
+		// Will this happen?
+		if lastModifiedTime.IsZero() {
```
yeya24, line 774: Umm, this seems unlikely to happen, but I am not sure if it is good to remove it.

yeya24:
Hello @pracucci  I updated the code to cache this time in meta file, PTAL take a look when you get a chance.

yeya24:
@kakkoyun Hello, I am not sure I have time to work on it this week. But since this change is big and it is changing the meta json format, so I am not sure whether I am in the right direction or not. It is better to have some 👁️  from @pracucci first.

pkg/block/fetcher.go
pracucci: I would simplify this, as follow:
- Remove this `if` branch, so that the filter doesn't need the objstore client
- If last modified is zero (for whatever reason), then we can just fallback to the old logic, which is reading the timestamp from the block ID (which is the timestamp when the block creation started)

pkg/block/metadata/meta.go
```diff
@@ -59,6 +60,9 @@ type Thanos struct {
 
 	// Source is a real upload source of the block.
 	Source SourceType `json:"source"`
+
+	// LastModified records the last modified time of meta.json
+	LastModified time.Time `json:"lastModified,omitempty"`
```
pracucci, line 65: Should be serialized to JSON? I don't think so. This field should be populated only in-memory, so I would not serialize it in the JSON.

pkg/block/fetcher.go
```diff
@@ -217,14 +217,45 @@ func (f *BaseFetcher) loadMeta(ctx context.Context, id ulid.ULID) (*metadata.Met
 		return nil, ErrorSyncMetaNotFound
 	}
 
+	writeMetaFunc := func(m *metadata.Meta) {
```
pracucci, line 220: I think the `meta.json` should never be rewritten (we should treat each object in the bucket as immutable). We should just read the LastModified and just save it in the in-memory Meta structure.

yeya24: I thought having lastModifiedTime in `meta.json` saves 1 obj store attributes call if we lose the in-memory Meta structure (restart, etc). Rewritten this also ensures all `meta.json` files have lastModifiedTime field including the old meta files.

I will update the pr as you suggested.

pkg/block/fetcher.go
yeya24: Do we still want to keep the `timestamp based block ID` way? I thought we want to fully change to use `lastModifiedTime` based on the comment

``` 
// TODO(bwplotka): Check consistency delay based on file upload / modification time instead of ULID.
```

If it is this case, then we don't need to have `lastModifiedTime` in the json file and the logic would be much easier! Thanks.

pracucci:
@yeya24 Would you have time to reiterate on this PR?

yeya24:
Sorry for the delay. I will try to work on it this or next week.

yeya24:
Hi, @pracucci I updated this pr. Please take a look when you get a chance, thanks!

yeya24:
Since we are using the modification time now, the compactor E2E tests fail because we used `30m` consistency-delay before and the blocks ID cannot be fake. So I just set the `consistency-delay` to `1s` and this time tests should pass now.

@pracucci Please take a look when you have time, thanks.

PR #2888, opened 2020-07-13
pkg/rules/rulespb: fix JSON marshaling for nil slices

simonpasquier:
Go's `encoding/json` turns nil slices into `null` instead of `[]`. This
is problematic for consumers expecting that the `alerts` field is a JSON
array. The same is true for empty rule groups.

* [X] Change is not relevant to the end user (the feature hasn't been released yet).

## Changes

The solution is to enforce that nil slices are converted to zero-length
slices when encoding.

## Verification

I've updated the unit tests to cover the issue.

### DIFF
--- a/pkg/query/api/v1_test.go
+++ b/pkg/query/api/v1_test.go
@@ -1359,6 +1359,7 @@ func TestRulesHandler(t *testing.T) {
 			EvaluationTime: all[3].GetAlert().EvaluationDurationSeconds,
 			Duration:       all[3].GetAlert().DurationSeconds,
 			Annotations:    nil,
+			Alerts:         []*testpromcompatibility.Alert{},
 			Type:           "alerting",
 		},
 	}
--- a/pkg/rules/rulespb/custom.go
+++ b/pkg/rules/rulespb/custom.go
@@ -177,6 +177,15 @@ func (r1 *Rule) Compare(r2 *Rule) int {
 	return 0
 }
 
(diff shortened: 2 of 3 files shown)


bwplotka:
Tests are failing for a good reason I think (:

simonpasquier:
@bwplotka I've fixed tests in ` pkg/query/api/v1_test.go` indeed. The e2e failure is more cryptic for me...

```
2020-07-13T09:51:15.5978723Z ##[error]    TestQueryExternalPrefixWithoutReverseProxy: query_test.go:278: [31mquery_test.go:278:
2020-07-13T09:51:15.5979200Z         
2020-07-13T09:51:15.5980293Z          unexpected error: some network requests failed: net::ERR_NETWORK_CHANGED; net::ERR_NETWORK_CHANGED; net::ERR_NETWORK_CHANGED; net::ERR_NETWORK_CHANGED; net::ERR_NETWORK_CHANGED; net::ERR_NETWORK_CHANGED; net::ERR_NETWORK_CHANGED[39m
2020-07-13T09:51:15.5980510Z         
2020-07-13T09:51:15.5980809Z Killing querier-1
2020-07-13T09:51:15.5980982Z Error: No such network: e2e_test_query
2020-07-13T09:51:15.5981064Z 
2020-07-13T09:51:15.5981286Z Unable to remove docker network e2e_test_query : exit status 1
```

bwplotka:
Looks like flak, retried

bwplotka:
Thanks!

PR #2892, opened 2020-07-14
receive: Overhaul start up sequence

kakkoyun:

## Changes

* Add more logs for receive start and shutdown sequence.
* Merge uploader routines (on-demand and recurring).
* Receive now fails if initial upload fails.

## Verification

* `make test-local`

### DIFF
--- a/cmd/thanos/receive.go
+++ b/cmd/thanos/receive.go
@@ -15,13 +15,13 @@ import (
 	"github.com/go-kit/kit/log"
 	"github.com/go-kit/kit/log/level"
 	"github.com/oklog/run"
-	opentracing "github.com/opentracing/opentracing-go"
+	"github.com/opentracing/opentracing-go"
 	"github.com/pkg/errors"
 	"github.com/prometheus/client_golang/prometheus"
 	"github.com/prometheus/client_golang/prometheus/promauto"
 	"github.com/prometheus/prometheus/pkg/labels"
 	"github.com/prometheus/prometheus/tsdb"
-	kingpin "gopkg.in/alecthomas/kingpin.v2"
+	"gopkg.in/alecthomas/kingpin.v2"
 
 	"github.com/thanos-io/thanos/pkg/component"
 	"github.com/thanos-io/thanos/pkg/extflag"
@@ -83,6 +83,8 @@ func registerReceive(m map[string]setupFunc, app *kingpin.Application) {
 	replicationFactor := cmd.Flag("receive.replication-factor", "How many times to replicate incoming write requests.").Default("1").Uint64()
 
 	forwardTimeout := modelDuration(cmd.Flag("receive-forward-timeout", "Timeout for forward requests.").Default("5s").Hidden())
+
+	uploadTimeout := modelDuration(cmd.Flag("receive.upload-timeout", "Timeout for the initial and last upload request.").Default("10m").Hidden())
 
 	tsdbMinBlockDuration := modelDuration(cmd.Flag("tsdb.min-block-duration", "Min duration for local TSDB blocks").Default("2h").Hidden())
 	tsdbMaxBlockDuration := modelDuration(cmd.Flag("tsdb.max-block-duration", "Max duration for local TSDB blocks").Default("2h").Hidden())
@@ -163,8 +165,9 @@ func registerReceive(m map[string]setupFunc, app *kingpin.Application) {
 			*replicaHeader,
 			*replicationFactor,
 			time.Duration(*forwardTimeout),
-			comp,
+			time.Duration(*uploadTimeout),
 			*allowOutOfOrderUpload,
+			comp,
 		)
 	}
 }
@@ -202,8 +205,9 @@ func runReceive(
 	replicaHeader string,
 	replicationFactor uint64,
 	forwardTimeout time.Duration,
-	comp component.SourceStoreAPI,
+	uploadTimeout time.Duration,
 	allowOutOfOrderUpload bool,
+	comp component.SourceStoreAPI,
 ) error {
 	logger = log.With(logger, "component", "receive")
 	level.Warn(logger).Log("msg", "setting up receive")
@@ -294,6 +298,7 @@ func runReceive(
 
 	level.Debug(logger).Log("msg", "setting up tsdb")
 	{
+		log.With(logger, "component", "storage")
 		dbUpdatesStarted := promauto.With(reg).NewCounter(prometheus.CounterOpts{
 			Name: "thanos_receive_multi_db_updates_attempted_total",
 			Help: "Number of Multi DB attempted reloads with flush and potential upload due to hashring changes",
@@ -311,12 +316,15 @@ func runReceive(
 
 			// Before quitting, ensure the WAL is flushed and the DBs are closed.
 			defer func() {
+				level.Info(logger).Log("msg", "shutting down Multi TSDB")
 				if err := dbs.Flush(); err != nil {
-					level.Warn(logger).Log("err", err, "msg", "failed to flush storage")
+					level.Error(logger).Log("err", err, "msg", "failed to flush storage")
 				}
 				if err := dbs.Close(); err != nil {
-					level.Warn(logger).Log("err", err, "msg", "failed to close multi db")
+					level.Error(logger).Log("err", err, "msg", "failed to close storage")
+					return
 				}
+				level.Info(logger).Log("msg", "Multi TSDB is closed")
 			}()
 
 			for {
@@ -328,7 +336,7 @@ func runReceive(
 						return nil
 					}
 					dbUpdatesStarted.Inc()
-					level.Info(logger).Log("msg", "updating Multi DB")
+					level.Info(logger).Log("msg", "updating Multi TSDB")
 
 					if err := dbs.Flush(); err != nil {
 						return errors.Wrap(err, "flushing storage")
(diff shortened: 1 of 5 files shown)


cmd/thanos/receive.go
```diff
@@ -84,6 +84,8 @@ func registerReceive(m map[string]setupFunc, app *kingpin.Application) {
 
 	forwardTimeout := modelDuration(cmd.Flag("receive-forward-timeout", "Timeout for forward requests.").Default("5s").Hidden())
 
+	uploadTimeout := modelDuration(cmd.Flag("receive.upload-timeout", "Timeout for the initial and last upload request.").Default("10m").Hidden())
```
bwplotka, line 87: It's very tricky if for initial and last. Why not all? (:

cmd/thanos/receive.go
```diff
@@ -311,12 +316,15 @@ func runReceive(
...
 				if err := dbs.Flush(); err != nil {
-					level.Warn(logger).Log("err", err, "msg", "failed to flush storage")
+					level.Error(logger).Log("err", err, "msg", "failed to flush storage")
 				}
 				if err := dbs.Close(); err != nil {
-					level.Warn(logger).Log("err", err, "msg", "failed to close multi db")
+					level.Error(logger).Log("err", err, "msg", "failed to close storage")
+					return
 				}
+				level.Info(logger).Log("msg", "Multi TSDB is closed")
```
bwplotka, line 327: Above you switched from multi TSDB to storage and here is MultiTSDB naming? Is it on purpose?

bwplotka: Also worth to mention status of flush? I know it looks logic not but worth to check if there is any log on successful flush (:

cmd/thanos/receive.go
```diff
@@ -489,20 +497,27 @@ func runReceive(
 	}
 
 	if upload {
-		level.Debug(logger).Log("msg", "upload enabled")
-		if err := dbs.Sync(context.Background()); err != nil {
-			level.Warn(logger).Log("msg", "initial upload failed", "err", err)
+		logger := log.With(logger, "component", "uploader")
+		{
+			level.Info(logger).Log("msg", "upload enabled, starting initial sync")
+			ctx, cancel := context.WithTimeout(context.Background(), uploadTimeout)
```
bwplotka, line 503: Nice, very good idea, but wonder if it should be per block or overall. 

It's hard to figure out from flag name. It looks like per block but what if someone had receiver with no upload with retention 1y and suddenly enable upload? It expects longer upload time

cmd/thanos/receive.go
```diff
@@ -489,20 +497,27 @@ func runReceive(
...
+			if err := dbs.Sync(ctx); err != nil {
+				cancel()
+				level.Warn(logger).Log("msg", "initial upload failed", "err", err)
+			}
+			cancel()
```
bwplotka, line 508: cancel will be done twice on error

cmd/thanos/receive.go
```diff
@@ -489,20 +497,27 @@ func runReceive(
...
+			level.Info(logger).Log("msg", "initial sync done")
```
bwplotka, line 509: This will be also logged on err, do we want this?

cmd/thanos/receive.go
squat, line 87: Nit: s/last/final/

cmd/thanos/receive.go
```diff
@@ -489,20 +497,27 @@ func runReceive(
...
-		level.Debug(logger).Log("msg", "upload enabled")
-		if err := dbs.Sync(context.Background()); err != nil {
-			level.Warn(logger).Log("msg", "initial upload failed", "err", err)
+		logger := log.With(logger, "component", "uploader")
+		{
+			level.Info(logger).Log("msg", "upload enabled, starting initial sync")
+			ctx, cancel := context.WithTimeout(context.Background(), uploadTimeout)
+			if err := dbs.Sync(ctx); err != nil {
+				cancel()
+				level.Warn(logger).Log("msg", "initial upload failed", "err", err)
```
bwplotka, line 506: Curious if this should not trigger crash :thinking:

cmd/thanos/receive.go
```diff
@@ -489,20 +497,27 @@ func runReceive(
...
 			ctx, cancel := context.WithCancel(context.Background())
 			g.Add(func() error {
 				return runutil.Repeat(30*time.Second, ctx.Done(), func() error {
+					level.Debug(logger).Log("msg", "recurring upload starting")
 					if err := dbs.Sync(ctx); err != nil {
-						level.Warn(logger).Log("msg", "interval upload failed", "err", err)
+						level.Warn(logger).Log("msg", "recurring upload failed", "err", err)
 					}
-
+					level.Debug(logger).Log("msg", "upload done")
```
bwplotka, line 520:
```suggestion
					level.Debug(logger).Log("msg", "recurring upload done")
```

What about adding `elapsed` time?

cmd/thanos/receive.go
```diff
@@ -521,9 +536,15 @@ func runReceive(
 				// Before quitting, ensure all blocks are uploaded.
 				defer func() {
 					<-uploadC
```
bwplotka, line 538: Wow what's this ;p I did not know we have it. In such case why we don't merge those two goroutines into one? With above repeat. I don't think we want 2 syncs running in same time.

cmd/thanos/receive.go
squat, line 520: I think this will be a bit confusing for users. If there's and error, first they'll see a log saying "something failed" and immediately afterwards they'll see "everything is done". Maybe we should make this an if-else

bwplotka: commented as well on similar thing. Let's be clear and avoid saying done on error :hugs:

cmd/thanos/receive.go
```diff
@@ -521,9 +536,15 @@ func runReceive(
...
-						level.Warn(logger).Log("msg", "on demnad upload failed", "err", err)
+					level.Info(logger).Log("msg", "uploading the last cut block before exiting")
+					dctx, dCancel := context.WithTimeout(context.Background(), uploadTimeout)
+					if err := dbs.Sync(dctx); err != nil {
+						dCancel()
+						level.Error(logger).Log("msg", "on demand upload failed", "err", err)
+						return
 					}
+					dCancel()
+					level.Info(logger).Log("msg", "the last cut block is uploaded")
```
squat, line 547: nit: s/is/was/

cmd/thanos/receive.go
kakkoyun: Yep, we can use for all for sure.

cmd/thanos/receive.go
kakkoyun: I'll unify them, you're right. I want to reserve `Mutli TSDB` for the package.

cmd/thanos/receive.go
kakkoyun: Actually good point. We can just stop if we can't upload it.

pkg/server/http/http.go
```diff
@@ -77,7 +78,9 @@ func (s *Server) Shutdown(err error) {
 
 	if err := s.srv.Shutdown(ctx); err != nil {
 		level.Error(s.logger).Log("msg", "internal server shut down failed", "err", err)
+		return
 	}
+	level.Info(s.logger).Log("msg", "internal server is shutdown", "err", err)
```
bwplotka, line 83:
```suggestion
	level.Info(s.logger).Log("msg", "internal server is shutdown gracefully", "err", err)
```

pkg/server/grpc/grpc.go
```diff
@@ -147,6 +149,7 @@ func (s *Server) Shutdown(err error) {
 	case <-stopped:
 		cancel()
 	}
+	level.Info(s.logger).Log("msg", "internal server is shutdown", "err", err)
```
bwplotka, line 152: ditto

cmd/thanos/receive.go
```diff
@@ -489,57 +499,69 @@ func runReceive(
...
 
 	if upload {
-		level.Debug(logger).Log("msg", "upload enabled")
-		if err := dbs.Sync(context.Background()); err != nil {
-			level.Warn(logger).Log("msg", "initial upload failed", "err", err)
+		logger := log.With(logger, "component", "uploader")
+		upload := func(ctx context.Context) error {
+			level.Debug(logger).Log("msg", "upload starting")
+			start := time.Now()
+			ctx, cancel := context.WithTimeout(ctx, uploadTimeout)
```
bwplotka, line 506: Again, as I mentioned. This is tricky as this timeout is not per block but per either thousands or one block, so not sure if this timeout is useful / not harmful :thinking:  Worth to make it per block? (on objstore level)?

So something we always wanted: https://github.com/thanos-io/thanos/issues/1830

cmd/thanos/receive.go
```diff
@@ -489,57 +499,69 @@ func runReceive(
...
 					case <-uploadC:
-						if err := dbs.Sync(ctx); err != nil {
-							level.Warn(logger).Log("err", err)
+						// Upload on demand.
+						if err := upload(ctx); err != nil {
+							level.Warn(logger).Log("msg", "on demand upload failed", "err", err)
 						}
 						uploadDone <- struct{}{}
+					case <-tick.C:
+						if err := upload(ctx); err != nil {
```
bwplotka, line 562: I assume that this is fine to requests upload via `uploadC` and wait until this one finishes and just perform next `upload` straight aways? Should be quick check and if nothing new then it's fine.. I think it's fine...

cmd/thanos/receive.go
kakkoyun: I've tried to address this issue at https://github.com/thanos-io/thanos/pull/2908

cmd/thanos/receive.go
kakkoyun: I've tried to address this issue at #2908

PR #2895, opened 2020-07-15
compact: fix increment of thanos_compact_downsample_total counter

belm0:
`thanos_compact_downsample_total` metric was not incremented for ResLevel1


squat:
Thanks a lot @belm0 :) this is a pretty common mistake and it would be great to have some static analysis to catch these issues

belm0:
> it would be great to have some static analysis to catch these

That would be hard, since it's valid to end at WithLabelValues() when initializing know label sets.

PR #2899, opened 2020-07-15
receive: Unify error handling

kakkoyun:
The Receiver has a couple of similar error modes that could be originated from different resources (i.e local tsdb, forward requests for other peers in the ring), this PR attempts to unify how errors are returned from those paths and how the errors handled, especially for the nested errors, in order to provide consistent fail mode behaviour and make the code easier to follow.

One other reason was, for some cases because of the wrapped errors, it was wrongly categorising the root cause of the errors, this PR also attempts to fix it.

## Changes

* Unify error handling paths for fan-out forward requests.

## Verification

* `make test-local`

### DIFF
--- a/pkg/receive/handler.go
+++ b/pkg/receive/handler.go
@@ -11,6 +11,7 @@ import (
 	stdlog "log"
 	"net"
 	"net/http"
+	"sort"
 	"strconv"
 	"sync"
 	"time"
@@ -56,8 +57,8 @@ const (
 )
 
 var (
-	// conflictErr is returned whenever an operation fails due to any conflict-type error.
-	conflictErr = errors.New("conflict")
+	// errConflict is returned whenever an operation fails due to any conflict-type error.
+	errConflict = errors.New("conflict")
 
 	errBadReplica  = errors.New("replica count exceeds replication factor")
 	errNotReady    = errors.New("target not ready")
@@ -269,13 +270,7 @@ func (h *Handler) handleRequest(ctx context.Context, rep uint64, tenant string, 
 	// Forward any time series as necessary. All time series
 	// destined for the local node will be written to the receiver.
 	// Time series will be replicated as necessary.
-	if err := h.forward(ctx, tenant, r, wreq); err != nil {
-		if countCause(err, isConflict) > 0 {
-			return conflictErr
-		}
-		return err
-	}
-	return nil
+	return h.forward(ctx, tenant, r, wreq)
 }
 
 func (h *Handler) receiveHTTP(w http.ResponseWriter, r *http.Request) {
@@ -316,14 +311,18 @@ func (h *Handler) receiveHTTP(w http.ResponseWriter, r *http.Request) {
 	}
 
 	err = h.handleRequest(ctx, rep, tenant, &wreq)
-	switch err {
+	if err != nil {
+		level.Debug(h.logger).Log("msg", "failed to handle request", "err", err)
+	}
+
+	switch rootCause(err) {
 	case nil:
 		return
 	case errNotReady:
 		http.Error(w, err.Error(), http.StatusServiceUnavailable)
 	case errUnavailable:
 		http.Error(w, err.Error(), http.StatusServiceUnavailable)
-	case conflictErr:
+	case errConflict:
 		http.Error(w, err.Error(), http.StatusConflict)
 	case errBadReplica:
 		http.Error(w, err.Error(), http.StatusBadRequest)
@@ -425,7 +424,7 @@ func (h *Handler) fanoutForward(pctx context.Context, tenant string, replicas ma
 				})
 				if err != nil {
 					h.replications.WithLabelValues(labelError).Inc()
-					ec <- errors.Wrapf(err, "replicate write request, endpoint %v", endpoint)
+					ec <- errors.Wrapf(err, "replicate write request for endpoint %v", endpoint)
 					return
 				}
 
@@ -453,16 +452,8 @@ func (h *Handler) fanoutForward(pctx context.Context, tenant string, replicas ma
 				if err != nil {
 					// When a MultiError is added to another MultiError, the error slices are concatenated, not nested.
 					// To avoid breaking the counting logic, we need to flatten the error.
-					if errs, ok := err.(errutil.MultiError); ok {
-						if countCause(errs, isConflict) > 0 {
-							err = errors.Wrap(conflictErr, errs.Error())
-						} else if countCause(errs, isNotReady) > 0 {
-							err = errNotReady
-						} else {
-							err = errors.New(errs.Error())
-						}
-					}
-					ec <- errors.Wrapf(err, "storing locally, endpoint %v", endpoint)
+					level.Debug(h.logger).Log("msg", "local tsdb write failed", "err", err.Error())
+					ec <- errors.Wrapf(cause(err, 1), "store locally for endpoint %v", endpoint)
 					return
 				}
 				ec <- nil
@@ -614,18 +605,8 @@ func (h *Handler) replicate(ctx context.Context, tenant string, wreq *prompb.Wri
 	quorum := h.writeQuorum()
 	// fanoutForward only returns an error if successThreshold (quorum) is not reached.
 	if err := h.fanoutForward(ctx, tenant, replicas, wreqs, quorum); err != nil {
-		if countCause(err, isNotReady) >= quorum {
-			return errors.Wrap(errNotReady, "replicate: quorum not reached")
-		}
-		if countCause(err, isConflict) >= quorum {
-			return errors.Wrap(conflictErr, "replicate: quorum not reached")
-		}
-		if countCause(err, isUnavailable) >= quorum {
-			return errors.Wrap(errUnavailable, "replicate: quorum not reached")
-		}
-		return errors.Wrap(err, "unexpected error, before quorum is reached")
+		return errors.Wrap(cause(err, quorum), "quorum not reached")
 	}
-
 	return nil
 }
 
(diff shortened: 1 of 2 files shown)


pkg/receive/handler.go
```diff
@@ -592,43 +582,102 @@ func (h *Handler) RemoteWrite(ctx context.Context, r *storepb.WriteRequest) (*st
...
-	return err == tsdb.ErrNotReady ||
+	return err == errNotReady ||
+		err == tsdb.ErrNotReady ||
 		status.Code(err) == codes.Unavailable
 }
 
+// countCause counts the number of errors within the given error slice
+// whose causes satisfy the given function.
+// cause will inspect the error's cause or, if the error is a MultiError,
+// the cause of each contained error but will not traverse any deeper.
```
squat, line 607: This comment is no longer accurate: this function _does_ traverse deeply, in fact it can recurse infinitely, which is new functionality compared to before.

pkg/receive/handler_test.go
```diff
@@ -94,47 +87,137 @@ func TestCountCause(t *testing.T) {
...
+			err: errors.Wrap(errors.Wrap(terrors.MultiError([]error{
 				storage.ErrOutOfOrderSample,
 				errors.New("foo"),
 				errors.New("bar"),
+			}), "baz"), "qux"),
+			threshold: 1,
+			exp:       errConflict,
+		},
+		{
+			name: "deep nested matching multierror",
```
squat, line 153: I'm just noting that this is new functionality compared to how error handling worked before

kakkoyun:
@squat I have removed the nested counting, it should be good now.

pkg/receive/handler.go
```diff
@@ -629,6 +599,77 @@ func isNotReady(err error) bool {
...
+func countCause(errs []error, f func(error) bool) int {
```
bwplotka, line 606:
```suggestion
func countCauses(errs []error, f func(error) bool) int {
```

pkg/receive/handler.go
```diff
@@ -629,6 +599,77 @@ func isNotReady(err error) bool {
...
+}
+
+func (a expectedErrors) Len() int           { return len(a) }
+func (a expectedErrors) Swap(i, j int)      { a[i], a[j] = a[j], a[i] }
+func (a expectedErrors) Less(i, j int) bool { return a[i].count < a[j].count }
+
+// cause returns a known error which has occurred more than the given threshold.
+// cause will inspect the error's cause or, if the error is a MultiError,
+// the cause of each contained error but will not traverse any deeper.
+func cause(err error, threshold int) error {
```
bwplotka, line 629: I think we can work on this name ... I think it could be more verbose. It means it can masks error, right? and you use it in wrap which can be confusing a lot...

pkg/receive/handler.go
```diff
@@ -629,6 +599,77 @@ func isNotReady(err error) bool {
...
+	if err == nil {
+		return nil
+	}
+	unwrappedErr := errors.Cause(err)
+	errs, ok := unwrappedErr.(terrors.MultiError)
+	if !ok {
+		errs = []error{unwrappedErr}
+	}
+	if threshold > 0 {
+		expErrs := expectedErrors{
```
bwplotka, line 639: You know that for those two error types, we can just remove `countCause` and implement all directly here and it will use less lines and lower complexity? (: Or am I missing something?

pkg/receive/handler.go
```diff
@@ -592,34 +581,15 @@ func (h *Handler) RemoteWrite(ctx context.Context, r *storepb.WriteRequest) (*st
...
 func isConflict(err error) bool {
 	if err == nil {
 		return false
 	}
-	return err == conflictErr ||
+	return err == errConflict ||
 		err == storage.ErrDuplicateSampleForTimestamp ||
 		err == storage.ErrOutOfOrderSample ||
 		err == storage.ErrOutOfBounds ||
-		err.Error() == strconv.Itoa(http.StatusConflict) ||
```
bwplotka, line 622: We don't need this one?

pkg/receive/handler.go
```diff
@@ -629,6 +599,77 @@ func isNotReady(err error) bool {
...
+	}
+	return errors.Wrap(err, "unexpected error")
+}
+
+// rootCause extracts the root cause of a fan-out error.
+func rootCause(err error) error {
+	if err == nil {
+		return nil
+	}
+	err = errors.Cause(cause(err, 1))
```
bwplotka, line 665: I am so confused here.

pkg/receive/handler.go
```diff
@@ -629,6 +599,77 @@ func isNotReady(err error) bool {
...
+	errs, ok := err.(terrors.MultiError)
+	if !ok {
+		return err
+	}
+	return errors.New(errs.Error())
```
bwplotka, line 670: and here even more xD

pkg/receive/handler_test.go
```diff
@@ -94,47 +87,137 @@ func TestCountCause(t *testing.T) {
...
 				errors.New("bar"),
 			}), "baz"),
-			f:   isConflict,
-			out: 0,
+			exp: errors.New("3 errors: 2 errors: qux; rpc error: code = AlreadyExists desc = conflict; foo; bar"),
 		},
 	} {
-		if n := countCause(tc.err, tc.f); n != tc.out {
-			t.Errorf("test case %s: expected %d, got %d", tc.name, tc.out, n)
+		got := rootCause(tc.err)
```
bwplotka, line 210: What about old good `testutil.Equals`?

pkg/receive/handler.go
```diff
@@ -629,6 +599,77 @@ func isNotReady(err error) bool {
 		status.Code(err) == codes.Unavailable
 }
 
+// countCause counts the number of errors within the given error slice
+// whose causes satisfy the given function.
+// cause will inspect the error's cause or, if the error is a MultiError,
```
bwplotka, line 604: is it important? 

because you know if we use `someunkownimportpackage.YoloError` it will be not traversed either ;p

pkg/receive/handler.go
kakkoyun: We don't use `http` for forwarding anymore. This is just a left-over.

pkg/receive/handler_test.go
kakkoyun: `testutil.Equals` won't help in this case. We are not using sentinel or predefined errors. It never returns true.

bwplotka: Well usually we do
```
if tc.exp != nil {
     testutil.NotOk(...
     testutil.Equals(t, tc.exp.Errors(),...
     continue
}
testutil.Ok(...
```

pkg/receive/handler.go
```diff
@@ -702,6 +667,58 @@ type retryState struct {
...
+}
+
+func (a expectedErrors) Len() int           { return len(a) }
+func (a expectedErrors) Swap(i, j int)      { a[i], a[j] = a[j], a[i] }
+func (a expectedErrors) Less(i, j int) bool { return a[i].count < a[j].count }
+
+// findCause returns a sentinel error that has occurred more than the given threshold.
+// It will inspect the error's cause if the error is a MultiError,
+// It will return cause of each contained error but will not traverse any deeper.
+func findCause(err error, threshold int) error {
```
brancz, line 683: If I'm not mistaken there are exactly two types of calls to this function, with threshold being 1 or quorum. I think it would be useful to have distinct function names for the two cases, so as a caller I don't have to understand the parameter meanings.

kakkoyun:
@squat @bwplotka Could you please have another look at this?

kakkoyun:
> @squat @bwplotka Could you please have another look at this?

@squat @bwplotka friendly ping!

brancz (approved):
lgtm but I think I would like another pair of eyes on this

kakkoyun:
@bwplotka I think I had addressed your requests for this one. Could you have another look?

kakkoyun:
@bwplotka friendly ping.

pkg/receive/handler.go
```diff
@@ -702,6 +667,56 @@ type retryState struct {
...
+func determineWriteErrorCause(err error, threshold int) error {
+	if err == nil {
+		return nil
+	}
+	unwrappedErr := errors.Cause(err)
+	errs, ok := unwrappedErr.(errutil.MultiError)
+	if !ok {
+		errs = []error{unwrappedErr}
+	}
+	if threshold > 0 {
```
bwplotka, line 692: To decrease complexity we could return early here on `threshold == 0`?

pkg/receive/handler.go
```diff
@@ -702,6 +667,56 @@ type retryState struct {
...
+		}
+		for _, exp := range expErrs {
+			exp.count = 0
+			for _, err := range errs {
+				if exp.cause(errors.Cause(err)) {
+					exp.count++
+				}
+			}
+		}
+		sort.Sort(sort.Reverse(expErrs))
```
bwplotka, line 706: Worth a comment why sorting is useful for humans (:

pkg/receive/handler.go
```diff
@@ -702,6 +667,56 @@ type retryState struct {
...
+		}
+	}
+	if len(errs) == 0 {
+		return nil
+	}
+	return err
+}
+
+// writeErrorCause extracts the root cause of a write error.
+func writeErrorCause(err error) error { return determineWriteErrorCause(err, 1) }
```
bwplotka, line 718: Maybe bit too shallow

bwplotka (approved):
LGTM, just small nits 👍🏽

bwplotka (commented):
Thanks 💪🏽

PR #2900, opened 2020-07-15
receive: Forward Request exponential back-off for unavailable nodes

kakkoyun:
This PR will help to reduce unnecessary forward requests and noisy logs while rolling out new versions for Thanos Receive.

*  [x] Change is not relevant to the end-user.

## Changes

* Implement exponential backoffs for forwarding requests when nodes are unavailable.

## Verification

* `make test-local`

### DIFF
--- a/pkg/receive/handler.go
+++ b/pkg/receive/handler.go
@@ -19,6 +19,7 @@ import (
 	"github.com/go-kit/kit/log/level"
 	"github.com/gogo/protobuf/proto"
 	"github.com/golang/snappy"
+	"github.com/jpillora/backoff"
 	"github.com/mwitkow/go-conntrack"
 	"github.com/opentracing/opentracing-go"
 	"github.com/pkg/errors"
@@ -54,10 +55,13 @@ const (
 	labelError   = "error"
 )
 
-// conflictErr is returned whenever an operation fails due to any conflict-type error.
-var conflictErr = errors.New("conflict")
+var (
+	// conflictErr is returned whenever an operation fails due to any conflict-type error.
+	conflictErr   = errors.New("conflict")
+	errBadReplica = errors.New("replica count exceeds replication factor")
 
-var errBadReplica = errors.New("replica count exceeds replication factor")
+	errNotReady = errors.New("target not ready")
+)
 
 // Options for the web Handler.
 type Options struct {
@@ -83,9 +87,11 @@ type Handler struct {
 	options  *Options
 	listener net.Listener
 
-	mtx      sync.RWMutex
-	hashring Hashring
-	peers    *peerGroup
+	mtx        sync.RWMutex
+	hashring   Hashring
+	peers      *peerGroup
+	expBackoff backoff.Backoff
+	backoffs   map[string]*peerState
 
 	forwardRequests   *prometheus.CounterVec
 	replications      *prometheus.CounterVec
@@ -103,6 +109,12 @@ func NewHandler(logger log.Logger, o *Options) *Handler {
 		router:  route.New(),
 		options: o,
 		peers:   newPeerGroup(o.DialOpts...),
+		expBackoff: backoff.Backoff{
+			Factor: 2,
+			Min:    100 * time.Millisecond,
+			Max:    30 * time.Second,
+			Jitter: true,
+		},
 		forwardRequests: promauto.With(o.Registry).NewCounterVec(
 			prometheus.CounterOpts{
 				Name: "thanos_receive_forward_requests_total",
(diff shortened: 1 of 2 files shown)


squat:
It looks to me like this PR doesn't really depend at all on #2899, does it? I think this one would actually be a lot less difficult to merge alone before 2899

kakkoyun:
@squat They share an error type which I can extract from other PR. I'll do it.

pkg/receive/handler.go
```diff
@@ -295,12 +303,16 @@ func (h *Handler) receiveHTTP(w http.ResponseWriter, r *http.Request) {
...
 	err = h.handleRequest(ctx, rep, tenant, &wreq)
-	switch err {
+	if err != nil {
+		level.Debug(h.logger).Log("msg", "failed to handle request", "err", err)
+	}
+
+	switch rootCause(err) {
 	case nil:
 		return
-	case tsdb.ErrNotReady:
```
bwplotka, line 301: Why this? (:

kakkoyun: The new error encapsulates both of the cases, tsdb not ready and service unavailable. IMHO It's better not to expose a sentinel error that belongs to a dependency.

pkg/receive/handler.go
```diff
@@ -54,10 +55,13 @@ const (
 	labelError   = "error"
 )
 
-// conflictErr is returned whenever an operation fails due to any conflict-type error.
-var conflictErr = errors.New("conflict")
+var (
+	// conflictErr is returned whenever an operation fails due to any conflict-type error.
+	conflictErr   = errors.New("conflict")
```
bwplotka, line 60: I know it's not your issue but it's bit inconsistent naming (err in the end). Can we clean this up while we are here? (:

pkg/receive/handler.go
bwplotka: The new error encapsulates both of the cases, tsdb not ready and service unavailable. 

Is this what we want? :thinking: I think then `errUnavailable` might more precise :thinking: 

>  IMHO It's better not to expose a sentinel error that belongs to a dependency.

No strong feelings, you might be right

pkg/receive/handler.go
```diff
@@ -625,10 +664,18 @@ func isConflict(err error) bool {
...
 func isNotReady(err error) bool {
-	return err == tsdb.ErrNotReady ||
+	return err == errNotReady ||
+		err == tsdb.ErrNotReady ||
 		status.Code(err) == codes.Unavailable
 }
 
+// peerState encapsulates the number of request attempt made against a peer and,
+// next allowed time for the next attempt.
+type peerState struct {
```
bwplotka, line 674: I think we don 't need to give usage context to type name. Something that this allows is rather preferred. For example: `retryState`

and the variable that holds map of it can be named `peerStates` (:

pkg/receive/handler.go
```diff
@@ -470,6 +485,15 @@ func (h *Handler) fanoutForward(pctx context.Context, tenant string, replicas ma
 				ec <- errors.Wrapf(err, "get peer connection for endpoint %v", endpoint)
 				return
 			}
+
+			b, ok := h.backoffs[endpoint]
```
bwplotka, line 489: I think this need to be mutexed, no? :thinking:

pkg/receive/handler.go
```diff
@@ -481,6 +505,21 @@ func (h *Handler) fanoutForward(pctx context.Context, tenant string, replicas ma
...
+						if b, ok := h.backoffs[endpoint]; ok {
+							b.attempt++
+							dur := h.expBackoff.ForAttempt(b.attempt)
+							b.nextAllowed = time.Now().Add(dur)
+							level.Debug(h.logger).Log("msg", "target unavailable backing off", "for", dur)
+						} else {
+							h.backoffs[endpoint] = &peerState{nextAllowed: time.Now().Add(h.expBackoff.ForAttempt(0))}
+						}
+						h.mtx.Unlock()
+					}
```
bwplotka, line 521: I think we could delete(endpoint) from it for cleannes on ok state, no?

pkg/receive/handler.go
kakkoyun: I have already fixed this issue on another PR #2899

pkg/receive/handler.go
kakkoyun: Yes, you're super right. I've missed this one.

pkg/receive/handler.go
kakkoyun: Added the delete logic in the success case. Also, we clear the states when the hashring get updated.

pkg/receive/handler.go
```diff
@@ -481,9 +511,28 @@ func (h *Handler) fanoutForward(pctx context.Context, tenant string, replicas ma
 				})
 			})
 			if err != nil {
+				// Check if peer connection is unavailable, don't attempt to send requests constantly.
+				if st, ok := status.FromError(err); ok {
+					if st.Code() == codes.Unavailable {
+						h.mtx.Lock()
+						if b, ok := h.peerStates[endpoint]; ok {
+							b.attempt++
```
bwplotka, line 519: there could be method in `retryState` for all of this (:

Issue #2444, opened 2020-04-15
Do not append custom prefix for metrics from imported packages for no reason.


bwplotka:
In many many places, we make sure that the metric has `thanos` prefix even though it comes from different packages outside of Thanos, e.g: https://github.com/thanos-io/thanos/blob/8f492a9f073f819019dd9f044e346a1e1fa730bc/cmd/thanos/compact.go#L246

This is wrong. I just learned that recently, because we cannot reuse the same dashboards and alerts as the same binaries (e.g Prometheus) that uses exactly the same code. 

This problem was also described by Brian long time ago https://www.robustperception.io/target-labels-not-metric-name-prefixes 

Let's fix it! This will be breaking change for ppl so let's be careful to mention this in CHANGELOG and update mixins (: 

AC:
* Don't change prefix unless we have to significantly change metric usage (add labels, register many times, etc)

bwplotka:
We had a good offline conversation with @pracucci and these  are some takeaways:

1. (advantage) Keeping old system prefix is preferable for reuse and understanding the metric technicality and detailed semantics
2. (downside)  Keeping old system prefix is definitely super confusing for users. A typical user would think "What are e.g Thanos metrics? So metrics exposed by Thanos?", so they will type `thanos_` and think that only those are from Thanos (it's not true, probably it's only 70% of it as we are still inconsistent).

To me:
Point (1) gives extreme benefits. IMO it is extremely important to maintain the old system metric name. There is nothing worse than misunderstanding what you are looking on. To fix point (2), I would suggest extending UI in some way to include a quick option to look on metric names for certain job/pod label value e.g  `{__ name__=~".*" , job="thanos"}... So users can quickly have the answer that question.

cc @brancz @brian-brazil @beorn7 any thoughts? :thinking: and @prmsrswt @juliusv for some UI idea!

brian-brazil:
2 is what's covered by the blog post, better tooling has always been the best way to approach that rather than misusing metric names as a crutch.

bwplotka:
I think we need waaaaay more than your old blog post. We don't want only advanced super experienced people to use Prometheus/Thanos (and I mean USE in terms of use metrics from those, not operating those). So, we need to be clearer and louder than this. More docs, examples, talks, tools would definitely will help. 

So I assume such UI enhancement might make sense? :thinking:

juliusv:
[...]

PR #2905, opened 2020-07-17
Wrap meta fetcher metrics with thanos prefix in replicate

yeya24:

## Changes

When I looked at the replicate metrics, I found that these fetch metrics are not wrapped. So I just add `thanos_` prefix in this pr.

Screenshot from 2020-07-17 17-13-45

## Verification

With this pr,

Screenshot from 2020-07-17 17-15-06

### DIFF
--- a/pkg/replicate/replicator.go
+++ b/pkg/replicate/replicator.go
@@ -151,7 +151,8 @@ func RunReplicate(
 		Help: "The Duration of replication runs split by success and error.",
 	}, []string{"result"})
 
-	fetcher, err := thanosblock.NewMetaFetcher(logger, 32, fromBkt, "", reg, nil, nil)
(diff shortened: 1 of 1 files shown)


yeya24:
For the prefix, one problem I saw. We are using a different prefix `thanos_bucket_` for bucket tools commands [here](https://github.com/thanos-io/thanos/blob/master/cmd/thanos/tools_bucket.go#L373), is it necessary to have different prefixes?

bwplotka:
Hey, this is tricky. The important thing to note is:  https://github.com/thanos-io/thanos/issues/2444

Let's discuss there (:

yeya24:
Close this pr because of https://cloud-native.slack.com/archives/CL25937SP/p1595924398405200

PR #2906, opened 2020-07-17
Refactor replicate execution method to not iterate the origin bucket

yeya24:

## Changes

In the previous implementation, we are iterating the origin bucket, and each time, we use metaFetcher to fetch all meta files. But this seems not necessary, we can just fetch meta files once at the beginning.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -28,6 +28,12 @@ We use *breaking :warning:* to mark changes that are not backward compatible (re
 - [#3331](https://github.com/thanos-io/thanos/pull/3331) Disable Azure blob exception logging
 - [#3341](https://github.com/thanos-io/thanos/pull/3341) Disable Azure blob syslog exception logging
 
+### Changed
+
+- [#2906](https://github.com/thanos-io/thanos/pull/2906) Tools: Refactor Bucket replicate execution. Removed all `thanos_replicate_origin_.*` metrics.
+    - `thanos_replicate_origin_meta_loads_total` can be replaced by `blocks_meta_synced{state="loaded"}`.
+    - `thanos_replicate_origin_partial_meta_reads_total` can be replaced by `blocks_meta_synced{state="failed"}`.
+
 ## [v0.16.0](https://github.com/thanos-io/thanos/releases) - 2020.10.26
 
 Highlights:
--- a/pkg/replicate/scheme.go
+++ b/pkg/replicate/scheme.go
@@ -168,7 +168,7 @@ func (rs *replicationScheme) execute(ctx context.Context) error {
 	availableBlocks := []*metadata.Meta{}
 
 	metas, partials, err := rs.fetcher.Fetch(ctx)
-	if err != nil && metas == nil {
+	if err != nil {
 		return err
 	}
 
--- a/pkg/replicate/scheme.go
+++ b/pkg/replicate/scheme.go
@@ -117,10 +117,6 @@ type replicationScheme struct {
 }
 
 type replicationMetrics struct {
-	originIterations  prometheus.Counter
-	originMetaLoads   prometheus.Counter
-	originPartialMeta prometheus.Counter
-
 	blocksAlreadyReplicated prometheus.Counter
 	blocksReplicated        prometheus.Counter
 	objectsReplicated       prometheus.Counter
@@ -128,18 +124,6 @@ type replicationMetrics struct {
 
 func newReplicationMetrics(reg prometheus.Registerer) *replicationMetrics {
 	m := &replicationMetrics{
-		originIterations: promauto.With(reg).NewCounter(prometheus.CounterOpts{
-			Name: "thanos_replicate_origin_iterations_total",
-			Help: "Total number of objects iterated over in the origin bucket.",
-		}),
-		originMetaLoads: promauto.With(reg).NewCounter(prometheus.CounterOpts{
-			Name: "thanos_replicate_origin_meta_loads_total",
-			Help: "Total number of meta.json reads in the origin bucket.",
-		}),
-		originPartialMeta: promauto.With(reg).NewCounter(prometheus.CounterOpts{
-			Name: "thanos_replicate_origin_partial_meta_reads_total",
-			Help: "Total number of partial meta reads encountered.",
-		}),
 		blocksAlreadyReplicated: promauto.With(reg).NewCounter(prometheus.CounterOpts{
 			Name: "thanos_replicate_blocks_already_replicated_total",
 			Help: "Total number of blocks skipped due to already being replicated.",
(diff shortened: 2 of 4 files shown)


pkg/replicate/scheme.go
```diff
@@ -117,29 +118,13 @@ type replicationScheme struct {
...
 	objectsReplicated       prometheus.Counter
 }
 
 func newReplicationMetrics(reg prometheus.Registerer) *replicationMetrics {
 	m := &replicationMetrics{
-		originIterations: promauto.With(reg).NewCounter(prometheus.CounterOpts{
-			Name: "thanos_replicate_origin_iterations_total",
-			Help: "Total number of objects iterated over in the origin bucket.",
-		}),
-		originMetaLoads: promauto.With(reg).NewCounter(prometheus.CounterOpts{
```
GiedriusS, line 135: Do we really want to remove this and the following metric? May still be useful. The previous one definitely needs to go, though.

pkg/replicate/scheme.go
```diff
@@ -183,45 +168,20 @@ func newReplicationScheme(
...
-			level.Info(rs.logger).Log("msg", "block meta not uploaded yet. Skipping.", "block_uuid", id.String())
-			return nil
-		}
-		if err != nil {
-			return errors.Wrapf(err, "load meta for block %v from origin bucket", id.String())
-		}
+	metas, partials, err := rs.fetcher.Fetch(ctx)
+	if err != nil && metas == nil {
+		return err
+	}
```
GiedriusS, line 174: Perhaps we should log the error if `metas != nil && err != nil`?

pkg/replicate/scheme.go
yeya24: Thanks for the quick review! This one `thanos_replicate_origin_meta_loads_total` is the same as the fetcher metric `thanos_blocks_meta_synced{state="loaded"}`

GiedriusS: Yep! You are correct. Let's document that in the `CHANGELOG.md` file and it should be good to go :+1:

pkg/replicate/scheme.go
yeya24: It is good to log the error here, but kind of duplicate.

I just check the code, only [here](https://github.com/thanos-io/thanos/blob/master/pkg/block/fetcher.go#L450) returns metas and error at the same time. In this case, partials != nil, so we will log the partial meta errors later.

yeya24:
Since this pr updates replicate metrics, I also need to update our mixins. It might make sense to merge #2905 first.

yeya24:
> AFAICT the CI failure seems related to the changes

Thanks, the previous problem was caused by the `make examples-in-container` target, and that should be fixed it in the latest commit :smile: .

pkg/replicate/scheme.go
OGKevin: By removing `thanos_replicate_origin_meta_loads_total`   and replacing it with `thanos_blocks_meta_synced` there is no way to distinct metrics from the 2 fetchers 🤔

OGKevin: So we either need to put back the origin metrics, or expose a label in the fetchers to distinguish them. E.g bucket name.

pkg/replicate/scheme.go
```diff
@@ -59,6 +59,7 @@ func NewBlockFilter(
 
 // Filter return true if block is non-compacted and matches selector.
 func (bf *BlockFilter) Filter(b *metadata.Meta) bool {
+	// TODO(bwplotka): Allow injecting custom labels as shipper does.
```
OGKevin, line 62: I don't think this todo should be here tho, it should be in replicator. As replicator should be able to inject labels like shipper does and not this block filter. This block filter is the last filter that will remove blocks without labels. So labels should be injected before they reach this filter.

pkg/replicate/scheme.go
```diff
@@ -183,45 +168,20 @@ func newReplicationScheme(
...
-			rs.metrics.originPartialMeta.Inc()
-			level.Info(rs.logger).Log("msg", "block meta not uploaded yet. Skipping.", "block_uuid", id.String())
-			return nil
-		}
-		if err != nil {
-			return errors.Wrapf(err, "load meta for block %v from origin bucket", id.String())
-		}
+	metas, partials, err := rs.fetcher.Fetch(ctx)
+	if err != nil && metas == nil {
+		return err
```
OGKevin, line 173: Should this error not be wrapped?

pkg/replicate/scheme.go
yeya24: Sorry, I didn't get the point. What do you mean by `distinct metrics from the 2 fetchers`? If the fetcher metrics are exposed by 2 components, then they must have different job labels? Maybe I misunderstood your question. WDYT?

pkg/replicate/scheme.go
yeya24: Good catch! I agree it should be in the replicator.

pkg/replicate/scheme.go
yeya24: Thanks, wrapping it seems better. I will update it later.

pkg/replicate/scheme.go
OGKevin: > If the fetcher metrics are exposed by 2 components
this would be correct, but not in this case. In this case 2 fetchers are being used in the same component. Fetchers for the 2 buckets.

OGKevin: But maybe it is still save the expose the bucket name as a static label in the fetcher metrics?

yeya24: I don't know the answer. But this is not part of this pr and fetcher metrics are used in store and compactor as well, so if you think it is necessary to have another label we could have another issue to track it.

pkg/replicate/scheme.go
yeya24: Just checked it and seems it is not necessary to wrap it because the error is already wrapped in the baseFetcher. Other components (store, compact) don't wrap it as well.

yeya24:
Can I get a review for this? I think it is close.

kakkoyun (approved):
lgtm

pkg/replicate/scheme.go
bwplotka: Yup, let's propose this in other Pr if needed

pkg/replicate/scheme.go
bwplotka: hm the `if err != nil && metas == nil {`  feels like code smell to me though. It just looks wrong from outside which can lead to bugs in future:

* Inside implementation can change, and metas can be produced without partial for some reason - then we will miss an error.
* Even if we say it makes sense, maybe another person will come here and refactor it anyway as it looks like error can be missed.

On other hand, shouldn't replicate FAIL on any error spotted? Even if we failed to read one block? Sometimes we heavily rely on replication status so this has to be heavily tested and controlled, ideally with failed metric being somewhere (:

pkg/replicate/scheme.go
```diff
@@ -183,45 +167,20 @@ func newReplicationScheme(
...
+	}
 
-		if len(meta.Thanos.Labels) == 0 {
-			// TODO(bwplotka): Allow injecting custom labels as shipper does.
-			level.Info(rs.logger).Log("msg", "block meta without Thanos external labels set. This is not allowed. Skipping.", "block_uuid", id.String())
-			return nil
-		}
+	for id, partialError := range partials {
+		level.Info(rs.logger).Log("msg", "failed to fetch block meta. Skipping.", "block_uuid", id.String(), "err", partialError)
```
bwplotka, line 176: Is this failure? This is actually valid state of blocks being potentially uploaded, no?

bwplotka: Something like `level.Info(rs.logger).Log("msg", "block meta not uploaded yet. Skipping.", "block_uuid", id.String())	
			r`

bwplotka (commented):
Nice, some comments though, before merging (:

kakkoyun:
@yeya24 This looks fine to me. You need to rebase I believe.
Ping @bwplotka are you satisfied with the state of the PR now?

pkg/replicate/scheme.go
bwplotka: > corrupted case, is it an error or not?

For current flow it's not error. We consider corrupted meta same as no meta.json as the reason is usually same - partial upload (:

bwplotka (approved):
Responded to comment, that still needs to be addressed, otherwise LGTM!

Thanks!

pkg/replicate/scheme.go
yeya24: Thanks! Updated.

CHANGELOG.md
```diff
@@ -28,6 +28,12 @@ We use *breaking :warning:* to mark changes that are not backward compatible (re
 - [#3331](https://github.com/thanos-io/thanos/pull/3331) Disable Azure blob exception logging
 - [#3341](https://github.com/thanos-io/thanos/pull/3341) Disable Azure blob syslog exception logging
 
+### Changed
+
+- [#2906](https://github.com/thanos-io/thanos/pull/2906) Tools: Refactor Bucket replicate execution. Removed all `thanos_replicate_origin_.*` metrics.
```
bwplotka, line 33: Technically not need to mention refactor, just metric changes (:

bwplotka (approved):
Thanks!

Changelog entry could be better, but we can fix it later.

OGKevin:
Whoop 🥳 🎊 Ill work on #2979.

OGKevin:
Nice work @yeya24

bwplotka:
:muscle:

PR #2908, opened 2020-07-20
[WIP] *: Add Bucket With Timeout

### LINKED ISSUE #1830: compact: Hanging upload

kakkoyun:

Fixes #1830.

## TODO

* [ ] Add tests

## Changes

* Add options to control the execution time of the bucket operations.

## Verification

* `make test-local`


cmd/thanos/receive.go
```diff
@@ -71,21 +71,17 @@ func registerReceive(m map[string]setupFunc, app *kingpin.Application) {
...
-
 	replicationFactor := cmd.Flag("receive.replication-factor", "How many times to replicate incoming write requests.").Default("1").Uint64()
-
-	forwardTimeout := modelDuration(cmd.Flag("receive-forward-timeout", "Timeout for forward requests.").Default("5s").Hidden())
+	forwardTimeout := modelDuration(cmd.Flag("receive-forward-timeout", "Timeout for each forward request.").Default("5s").Hidden())
 
 	tsdbMinBlockDuration := modelDuration(cmd.Flag("tsdb.min-block-duration", "Min duration for local TSDB blocks").Default("2h").Hidden())
 	tsdbMaxBlockDuration := modelDuration(cmd.Flag("tsdb.max-block-duration", "Max duration for local TSDB blocks").Default("2h").Hidden())
+
+	blockUploadTimeout := modelDuration(cmd.Flag("shipper.block-upload-timeout", "Timeout for each block upload request.").Default("10m").Hidden())
```
bwplotka, line 84: Let's mention breaking change that now timeouts is added and it has 10m. WDYT?

I would make it even larger, like 30m as this is totally fine for Thanos system. Anything beyond 2h really might cause issues.

bwplotka: I would even add this as non hidden flag

kakkoyun: I made it public and changed the default to `30m`.

kakkoyun:
> I would say this is fine, however I would seeing this as rather bucket layer timeout 🤔
And this timeout should be ensure by each client provider. However since we cannot really assure that each implementation does that properly (can we?) Mabe multiple layers of timeouts would do.

As you said, this could be harder to push down to bucket layer. But I can have another look at it.

> The only problem in this PR is that we also upload blocks as the compactor, and we cannot use similar upload timeout - would be nice to make it consistent.

One other place we can actually use for this timeout and it is `block.Upload`. This would actually make it consistent for compactor and any other component as well. I'll also check the possibility of doing so.

bwplotka:
> As you said, this could be harder to push down to bucket layer. But I can have another look at it.


What about pushing it at least to `block.Upload` helper?

bwplotka:
lols you meantioned it already here https://github.com/thanos-io/thanos/pull/2908#issuecomment-662887173

yes, this idea makes sesne

pkg/objstore/timeout.go
```diff
@@ -0,0 +1,100 @@
...
+	Get:        5 * time.Second,
+	GetRange:   30 * time.Second,
+	Exists:     5 * time.Second,
+	Upload:     10 * time.Minute,
+	Delete:     5 * time.Second,
+	Attributes: 1 * time.Minute,
+}
+
+type Timeout struct {
+	Iter       time.Duration `yaml:"iter"`
```
kakkoyun, line 23: I'm looking for feedbacks for the default timeout values.

povilasv: maybe default disabled?

povilasv (approved):

pkg/objstore/timeout.go
kakkoyun: Yes, definitely could be a good strategy.

krasi-georgiev (commented):
Is there a test for these timeouts?

docs/storage.md
```diff
@@ -96,6 +96,14 @@ config:
...
     kms_encryption_context: {}
     encryption_key: ""
+timeout:
+  iter: 0s
+  get: 5s
+  get_range: 30s
+  exists: 5s
+  upload: 10m0s
+  delete: 5s
+  attributes: 1m0s
```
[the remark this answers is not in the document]
kakkoyun: These actions are pretty well-known. However, we can add additional documentation. I suspect most likely only upload timeout value will be used if we provide sane defaults.

kakkoyun:
@krasi-georgiev Thanks for the feedback, I'll add tests for these changes and additional documentation. I'm marking it as WIP again.

pkg/objstore/timeout.go
kakkoyun: > As I mentioned in one of the previous Thanos community calls, I personally believe that whenever there's a data transfer involved (download/upload) we should timeout based on minimum speed and not time. It's very difficult to fine-tune these timeouts otherwise.

This is also on my agenda. I wanted to have a simpler solution as well. But you have a fair point that it'd be harder to come up with those default values. I'll park this one and focus on bandwidth/minimum speed based implementation and then we can decide which one to go with.

krasi-georgiev: ooh yeah I very much like this idea as well!

PR #2913, opened 2020-07-20
Add FLAT ZONE as adopter

itsx:

Hi!

We use Thanos in production almost one year at FLAT ZONE and would love to be listed as an adopter on your website.

Thanks for your awesome work :smiley: :clap:


bwplotka:
Thanks for your support, LGTM! :muscle:

PR #2915, opened 2020-07-21
receive, rule: Lock TSDB directories 

kakkoyun:
This PR attempts to prevent any concurrent access to TSDB directories on a single session.

## Changes

* Lock TSDB dir for Receiver
* Lock TSDB dir for Ruler

## TODO

* Introduce a flag or a default mode to clean up flags, in case of a dirty shutdown in the next startup.

## Verification

* `make test-local`
* `make test-e2e`

### DIFF
--- a/cmd/thanos/receive.go
+++ b/cmd/thanos/receive.go
@@ -87,7 +87,7 @@ func registerReceive(m map[string]setupFunc, app *kingpin.Application) {
 	tsdbMinBlockDuration := modelDuration(cmd.Flag("tsdb.min-block-duration", "Min duration for local TSDB blocks").Default("2h").Hidden())
 	tsdbMaxBlockDuration := modelDuration(cmd.Flag("tsdb.max-block-duration", "Max duration for local TSDB blocks").Default("2h").Hidden())
 	walCompression := cmd.Flag("tsdb.wal-compression", "Compress the tsdb WAL.").Default("true").Bool()
-	noLockFile := cmd.Flag("tsdb.no-lockfile", "Do not create lockfile in TSDB data directory.").Default("false").Bool()
+	noLockFile := cmd.Flag("tsdb.no-lockfile", "Do not create lockfile in TSDB data directory. In any case, the lockfiles will be deleted on next startup.").Default("false").Bool()
 
 	ignoreBlockSize := cmd.Flag("shipper.ignore-unequal-block-size", "If true receive will not require min and max block size flags to be set to the same value. Only use this if you want to keep long retention and compaction enabled, as in the worst case it can result in ~2h data loss for your Thanos bucket storage.").Default("false").Hidden().Bool()
 	allowOutOfOrderUpload := cmd.Flag("shipper.allow-out-of-order-uploads",
@@ -304,9 +304,9 @@ func runReceive(
 			Help: "Number of Multi DB completed reloads with flush and potential upload due to hashring changes",
 		})
 
-		level.Debug(logger).Log("msg", "cleaning storage lock files")
-		if err := dbs.Clean(); err != nil {
-			return errors.Wrap(err, "cleaning storage lock files")
+		level.Debug(logger).Log("msg", "removing storage lock files if any")
+		if err := dbs.RemoveLockFilesIfAny(); err != nil {
+			return errors.Wrap(err, "remove storage lock files")
 		}
 
 		// TSDBs reload logic, listening on hashring changes.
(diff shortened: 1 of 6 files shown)


brancz:
For multi-tsdb this lgtm, but I'm not entirely sure about the ruler, why do we think we need this there? I think it would be ok to have configurable, so people can make this decision themselves, much like with Prometheus (this goes for both receive and ruler), as locking like this may already be handled by kubernetes for example.

kakkoyun:
@brancz It was just premature optimization for Ruler, I can revert it. We can add a flag to control this behaviour. What should be the default? Prometheus uses lock file by default, should we as well?

brancz:
I believe creating the lock by default is correct, with the option to opt out. As long this is clear in the changelog, this is fine in my opinion. Would be good to be confirmed by @bwplotka though.

cmd/thanos/receive.go
```diff
@@ -86,10 +86,10 @@ func registerReceive(m map[string]setupFunc, app *kingpin.Application) {
...
 	tsdbMaxBlockDuration := modelDuration(cmd.Flag("tsdb.max-block-duration", "Max duration for local TSDB blocks").Default("2h").Hidden())
-	ignoreBlockSize := cmd.Flag("shipper.ignore-unequal-block-size", "If true receive will not require min and max block size flags to be set to the same value. Only use this if you want to keep long retention and compaction enabled, as in the worst case it can result in ~2h data loss for your Thanos bucket storage.").Default("false").Hidden().Bool()
-
 	walCompression := cmd.Flag("tsdb.wal-compression", "Compress the tsdb WAL.").Default("true").Bool()
+	noLockFile := cmd.Flag("tsdb.no-lockfile", "Do not create lockfile in TSDB data directory.").Default("false").Bool()
 
+	ignoreBlockSize := cmd.Flag("shipper.ignore-unequal-block-size", "If true receive will not require min and max block size flags to be set to the same value. Only use this if you want to keep long retention and compaction enabled, as in the worst case it can result in ~2h data loss for your Thanos bucket storage.").Default("false").Hidden().Bool()
```
bwplotka, line 92:
```suggestion
	ignoreBlockSize := cmd.Flag("shipper.ignore-unequal-block-size", "If true receive will not require min and max block size flags to be set to the same value. Only use this if you want to keep long retention and compaction enabled, as in the worst case it can result in ~2h data loss for your Thanos bucket storage.").Default("false").Hidden().Bool()
```
Hm... this is actually something we can improve. Let's add TODO and ensure there is issue for this. We can leverage vertical compaciton just fine in this case.

cmd/thanos/receive.go
```diff
@@ -303,6 +303,11 @@ func runReceive(
 			Help: "Number of Multi DB completed reloads with flush and potential upload due to hashring changes",
 		})
 
+		level.Debug(logger).Log("msg", "cleaning storage lock files")
+		if err := dbs.Clean(); err != nil {
```
bwplotka, line 307: Looks scary! ;p 

Also hm....... I think this needs explanation in the flag, that lock file is removed on the start always.

NOTE: Maybe worth to add log msg if we have lock file still? (indicating hard shutdown)

bwplotka: Also we need the same on ruler for now

bwplotka:
```suggestion
		if err := dbs.RemoveLockFilesIfAny(); err != nil {
```

cmd/thanos/rule.go
```diff
@@ -642,6 +650,21 @@ func runRule(
...
+	if err != nil {
+		return err
+	}
+	if err := os.Remove(filepath.Join(absdir, "lock")); err != nil {
+		if os.IsNotExist(err) {
+			return nil
+		}
+		return err
+	}
+	level.Info(logger).Log("msg", "a leftover lockfile found and removed")
```
bwplotka, line 664: nice and explict! :+1:

pkg/receive/multitsdb.go
```diff
@@ -213,6 +214,32 @@ func (t *MultiTSDB) Sync(ctx context.Context) error {
...
+		}
+		return err
+	}
+
+	merr := terrors.MultiError{}
+	for _, fi := range fis {
+		if !fi.IsDir() {
+			continue
+		}
+		if err := os.Remove(filepath.Join(t.defaultTenantDataDir(fi.Name()), "lock")); err != nil {
```
bwplotka, line 231: I would reuse the `removeLockfileIfAny`, because... why not? (:

PR #2613, opened 2020-05-15
Replace BucketReader.ObjectSize() with .Attributes()

pracucci:
* [ ] I added CHANGELOG entry for this change.
* [x] Change is not relevant to the end user.

## Changes

As first step to address #2585, in this PR I've replaced `BucketReader.ObjectSize()` with `Attributes()` in order to also be able to read the object last modified timestamp via the `BucketReader`.

/cc @pstibrany 

## Verification

Unit tests.

PR #2916, opened 2020-07-21
Fixed time format for Last-Modified header

SabaPing:
Fixed time format for Last-Modified header when parsing oss/cos attributes.

Signed-off-by: Sabaping <bbq.yuki@outlook.com>

[COS api doc reference](https://intl.cloud.tencent.com/document/product/436/7729)
[OSS api doc reference](https://www.alibabacloud.com/help/doc-detail/31985.htm)

## Changes

1. `clientutil.ParseLastModified` function accepts optional format parameter
2. `Bucket.Attributes` function, for cos/oss object storage, passes additional `time.RFC1123` when calls `clientutil.ParseLastModified`.

## Verification

Unit test.

### DIFF
--- a/pkg/objstore/clientutil/parse.go
+++ b/pkg/objstore/clientutil/parse.go
@@ -34,10 +34,14 @@ func ParseContentLength(m http.Header) (int64, error) {
 }
 
 // ParseLastModified returns the timestamp parsed from the Last-Modified
-// HTTP header in input (expected to be in the RFC3339 format).
-// Passing an optional second parameter to override the default format.
-func ParseLastModified(m http.Header, f ...string) (time.Time, error) {
-	const name = "Last-Modified"
+// HTTP header in input.
+// Passing an second parameter, named f, to specify the time format.
+// If f is empty then RFC3339 will be used as default format.
+func ParseLastModified(m http.Header, f string) (time.Time, error) {
+	const (
+		name          = "Last-Modified"
+		defaultFormat = time.RFC3339
+	)
 
 	v, ok := m[name]
 	if !ok {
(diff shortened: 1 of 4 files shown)


SabaPing:
#2613

SabaPing:
Fixed unit tests

pkg/objstore/clientutil/parse.go
```diff
@@ -35,7 +35,8 @@ func ParseContentLength(m http.Header) (int64, error) {
 
 // ParseLastModified returns the timestamp parsed from the Last-Modified
 // HTTP header in input (expected to be in the RFC3339 format).
-func ParseLastModified(m http.Header) (time.Time, error) {
+// Passing an optional second parameter to override the default format.
+func ParseLastModified(m http.Header, f ...string) (time.Time, error) {
```
bwplotka, line 39: I would vote for making this argument explicit. The problem by this way of doing `optional` arguments it's just it's end being quick hacky and surprising. 

It's totally ok to pass everywhere `time.RFC3339` or even allow passing "" and setting to time.RFC3339 if that happens.

pkg/objstore/cos/cos.go
```diff
@@ -105,7 +106,7 @@ func (b *Bucket) Attributes(ctx context.Context, name string) (objstore.ObjectAt
 		return objstore.ObjectAttributes{}, err
 	}
 
-	mod, err := clientutil.ParseLastModified(resp.Header)
+	mod, err := clientutil.ParseLastModified(resp.Header, time.RFC1123)
```
bwplotka, line 109: Can we add comment why this is needed? :hugs: 

Same below

SabaPing:
Refactored code and improved readability based on @bwplotka 's comments.

bwplotka:
Thanks. I think we need some CHANGELOG for this, but we can add it during release time.

wujinhu:
LGTM❤️  @SabaPing  @kakkoyun

PR #2921, opened 2020-07-22
.*: Add header for remote read version

geobeau:

## Changes

Added a http header to the remote_read request.
While Prometheus itself doesn't care about this header, for some third party remote_read compatible api, it is required.
The version match the one made by Prometheus when performing a remote read (see https://github.com/prometheus/prometheus/blob/v2.20.0-rc.1/storage/remote/client.go#L237) .

## Verification

Manually tested again a third party API (Cortex)


bwplotka:
Thanks!

PR #2926, opened 2020-07-22
api: Refactor API into a new package

prmsrswt:

## Changes
- Consolidated the API logic into a new package.
- Implement API in the Compactor.
- Add `/blocks` endpoint to return blocks metadata.
- Expose the same API from Block Viewer.

## Verification
Tested Locally using docker-compose setup.


cmd/thanos/compact.go
```diff
@@ -81,16 +83,30 @@ func registerCompact(m map[string]setupFunc, app *kingpin.Application) {
 	conf.registerFlag(cmd)
 
 	m[component.Compact.String()] = func(g *run.Group, logger log.Logger, reg *prometheus.Registry, tracer opentracing.Tracer, _ <-chan struct{}, _ bool) error {
-		return runCompact(g, logger, reg, component.Compact, *conf)
+		flagsMap := map[string]string{}
+
+		// Exclude kingpin default flags to expose only Thanos ones.
+		boilerplateFlags := kingpin.New("", "").Version("")
```
bwplotka, line 89: sounds like some helper would be nice for this (:

pkg/api/compact/v1.go
```diff
@@ -0,0 +1,68 @@
...
+	instr := api.GetInstr(tracer, logger, ins)
+
+	r.Get("/blocks", instr("blocks", capi.blocks))
+}
+
+func (capi *CompactAPI) blocks(r *http.Request) (interface{}, []error, *api.ApiError) {
+	return capi.blocksInfo, nil, nil
+}
+
+// Set updates the blocks metadata in API
```
bwplotka, line 56: This is new API for block viewer? :thinking:

pkg/api/compact/v1.go
```diff
@@ -0,0 +1,68 @@
+// Copyright (c) The Thanos Authors.
+// Licensed under the Apache License 2.0.
+
+package v1
+
+import (
```
bwplotka, line 6: At this point I would really be careful - it would be nice to have sort of apis like:
* query APIs + UI
* Rules APIs + UI
* Base APIs + UI (flags, metrics endpoints etc)
* Blocks APIs. 

Naming this compact is rather specifying WHO is using this not WHAT this package gives. I think it should be the latter :hugs: 
Then we can specify WHO is using this in each component by registering those which are needed

cmd/thanos/compact.go
prmsrswt: Yep, this was on my list, but I am not sure what would be the best place to write the helper.

pkg/api/compact/v1.go
prmsrswt: Yep, this started as implementing the API for block viewer, but there wasn't a good place to write the API logic, as it wasn't coupled to a specific component. So we decided to refactor the API code in a separate package.

pkg/api/compact/v1.go
prmsrswt: I agree, before the separate package, I wrote the API as a part of the compact package, and then during the refactor it just slipped my mind. I'd rename it as blocks API.

cmd/thanos/compact.go
bwplotka: any (: something in this main package for a start would do (:

PR #2929, opened 2020-07-23
mixin: Fix alert about unhealthy sidecar

hwoarang:
The alert was giving the wrong information as the $value contained
the number of pods that failing to send heartbeat instead of the actual
number of seconds that each sidecar was being unhealthy.
    
Also the 5 minute interval is probably too low as on large deployments
prometheus could take much longer to come up online and for sidecar to
become actually useful.
    
As such, we can simply subtract the timestamp of the last heartbeat from
the current time and fire if we are lagging for more than 10 minutes.

This is also consistent with the current unit tests.

## Changes

Improve alert for sidecar to provide correct information and not fire too soon

## Verification

Deployed locally and did a prometheus update and it did not fire anymore

### DIFF
--- a/examples/alerts/alerts.md
+++ b/examples/alerts/alerts.md
@@ -275,7 +275,7 @@ rules:
     message: Thanos Sidecar {{$labels.job}} {{$labels.pod}} is unhealthy for {{ $value
       }} seconds.
   expr: |
-    count(time() - max(thanos_sidecar_last_heartbeat_success_time_seconds{job=~"thanos-sidecar.*"}) by (job, pod) >= 300) > 0
+    time() - max(thanos_sidecar_last_heartbeat_success_time_seconds{job=~"thanos-sidecar.*"}) by (job, pod) >= 600
   labels:
     severity: critical
 ```
--- a/mixin/alerts/sidecar.libsonnet
+++ b/mixin/alerts/sidecar.libsonnet
@@ -27,7 +27,7 @@
               message: 'Thanos Sidecar {{$labels.job}} {{$labels.pod}} is unhealthy for {{ $value }} seconds.',
             },
             expr: |||
-              count(time() - max(thanos_sidecar_last_heartbeat_success_time_seconds{%(selector)s}) by (job, pod) >= 300) > 0
+              time() - max(thanos_sidecar_last_heartbeat_success_time_seconds{%(selector)s}) by (job, pod) >= 600
             ||| % thanos.sidecar,
             labels: {
               severity: 'critical',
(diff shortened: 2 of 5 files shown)


hwoarang:
> lgtm 1st_place_medal It needs rebasing otherwise we can merge

@kakkoyun  thank you for the approval. I have just rebased it

examples/alerts/alerts.md
```diff
@@ -275,7 +275,7 @@ rules:
     message: Thanos Sidecar {{$labels.job}} {{$labels.pod}} is unhealthy for {{ $value
```
kakkoyun, line 275: When I checked it again since we removed `pod` aggregation, `{{$labels.pod}}` won't be available.

examples/alerts/alerts.md
```diff
@@ -275,7 +275,7 @@ rules:
...
       }} seconds.
   expr: |
-    count(time() - max(thanos_sidecar_last_heartbeat_success_time_seconds{job=~"thanos-sidecar.*"}) by (job, pod) >= 300) > 0
+    time() - max(thanos_sidecar_last_heartbeat_success_time_seconds{job=~"thanos-sidecar.*"}) by (job) >= 600
```
kakkoyun, line 278: How about keeping `pod` aggregation? It'd make it easier to pin point the source.

kakkoyun:
@hwoarang Friendly ping.

hwoarang:
@kakkoyun apologies for the delay but due to holidays, connectivity and time are at a premium :) nevertheless I will try to get to it as soon as possible.

examples/alerts/alerts.md
hwoarang: @kakkoyun sure I can do that.

hwoarang: @kakkoyun because we are using the max() operator I am not sure how we can use the pod aggregation. Right now the expression finds the pod with the max latency. If we use `by (job,pod)` then the `max()` has no effect as pod+job combinations are unique anyway. Do I miss something?

examples/alerts/alerts.md
kakkoyun: @hwoarang I could be wrong here. Better to test it using the alert tests we have.

hwoarang:
@kakkoyun I believe I have addressed all your concerns now

mixin/alerts/sidecar.libsonnet
```diff
@@ -27,7 +27,7 @@
               message: 'Thanos Sidecar {{$labels.job}} {{$labels.pod}} is unhealthy for {{ $value }} seconds.',
             },
             expr: |||
-              count(time() - max(thanos_sidecar_last_heartbeat_success_time_seconds{%(selector)s}) by (job, pod) >= 300) > 0
+              time() - max(thanos_sidecar_last_heartbeat_success_time_seconds{%(selector)s}) by (job) >= 600
```
kakkoyun, line 30: This is why tests faild I believe
```suggestion
              time() - max(thanos_sidecar_last_heartbeat_success_time_seconds{%(selector)s}) by (job, pod) >= 600
```

hwoarang: The change looks correct but I do not know how this was related to the e2e test failures. I pushed again

kakkoyun:
@hwoarang Thanks a lot 🙏

Issue #2919, opened 2020-07-22
querier: Allow passing allow-list of store (IPs?) that should be used for query for debug purposes; Add multiselect on UI


bwplotka:
It would be awesome to add the ability to query only a subset of stores to detect which one is slow/ problematic etc (: It should be hidden option only for ad-hoc query.

geobeau:
I am interested in doing it.

After looking at how to implement it, I have a design question: the query api uses the proxy as a `storepb.StoreServer` and the logic behind the store selection is done after calling `storepb.StoreServer.Series()`, however there is no way of passing an allowlist without extending the interface. Would be okay to change querier.proxy to use directly `ProxyStore` in place of `storepb.StoreServer`?

bwplotka:
Awesome. I think it would be nice to avoid passing this in protobuf, but we could technically create some hints.

I would also avoid running just ProxyStore and changing methods. 

We have two options:
1. Pass through context as a best-effort thing
2. Extend protobuf

I would start for now with 1 and then 2 long term. WDYT? (:

geobeau:
Pass through context seems good for me. 👍

geobeau:
Hello, can you take a look at #2931 ? It's still in WIP as I need to finish the testing but I would like your input first.

bwplotka:
I did, and left some comments. Good approach overall!

bwplotka:
It's now done in https://github.com/thanos-io/thanos/pull/2931 and documented.

PR #2931, opened 2020-07-23
Query: Allow passing a `storeMatcher[]` to select matching stores when debugging the querier

geobeau:
* [ ] Get the general design validated.
* [x] Update tests.
* [x] Update documentation.

## Changes

Solves the api part of #2919
Add a new parameter on the API to pass an allow list of stores.
Only stores matching this list will be able to be used for the query.

## Verification

WIP: manual testing and partial unit tests

### DIFF
--- a/docs/components/query.md
+++ b/docs/components/query.md
@@ -222,6 +222,26 @@ Thanos Querier has the ability to perform concurrent select request per query. I
 The maximum number of concurrent requests are being made per query is controller by `query.max-concurrent-select` flag.
 Keep in mind that the maximum number of concurrent queries that are handled by querier is controlled by `query.max-concurrent`. Please consider implications of combined value while tuning the querier.
 
+### Store filtering
+
+It's possible to provide a set of matchers to the Querier api to select specific stores to be used during the query using the `storeMatch[]` parameter. It is useful when debugging a slow/broken store.
+It uses the same format as the matcher of [Prometheus' federate api](https://prometheus.io/docs/prometheus/latest/querying/api/#finding-series-by-label-matchers).
+Note that at the moment the querier only supports the `__address__` which contain the address of the store as it is shown on the `/stores` endoint of the UI.
+
+Example:
+```
+- targets:
+  - prometheus-foo.thanos-sidecar:10901
+  - prometheus-bar.thanos-sidecar:10901
+```
+
+```
+http://localhost:10901/api/v1/query?query=up&dedup=true&partial_response=true&storeMatch={__address__=~"prometheus-foo.*"}
+```
+
+Will only return metrics from `prometheus-foo.thanos-sidecar:10901`
+
+
 ## Expose UI on a sub-path
 
 It is possible to expose thanos-query UI and optionally API on a sub-path.
--- a/pkg/query/querier.go
+++ b/pkg/query/querier.go
@@ -17,6 +17,7 @@ import (
 	"github.com/prometheus/prometheus/storage"
 
 	"github.com/thanos-io/thanos/pkg/gate"
+	"github.com/thanos-io/thanos/pkg/store"
 	"github.com/thanos-io/thanos/pkg/store/storepb"
 	"github.com/thanos-io/thanos/pkg/tracing"
 )
@@ -27,17 +28,18 @@ import (
 // replicaLabels at query time.
 // maxResolutionMillis controls downsampling resolution that is allowed (specified in milliseconds).
 // partialResponse controls `partialResponseDisabled` option of StoreAPI and partial response behavior of proxy.
-type QueryableCreator func(deduplicate bool, replicaLabels []string, maxResolutionMillis int64, partialResponse, skipChunks bool) storage.Queryable
+type QueryableCreator func(deduplicate bool, replicaLabels []string, storeMatchers [][]storepb.LabelMatcher, maxResolutionMillis int64, partialResponse, skipChunks bool) storage.Queryable
 
 // NewQueryableCreator creates QueryableCreator.
 func NewQueryableCreator(logger log.Logger, reg prometheus.Registerer, proxy storepb.StoreServer, maxConcurrentSelects int, selectTimeout time.Duration) QueryableCreator {
 	keeper := gate.NewKeeper(reg)
 
-	return func(deduplicate bool, replicaLabels []string, maxResolutionMillis int64, partialResponse, skipChunks bool) storage.Queryable {
+	return func(deduplicate bool, replicaLabels []string, storeMatchers [][]storepb.LabelMatcher, maxResolutionMillis int64, partialResponse, skipChunks bool) storage.Queryable {
 		return &queryable{
 			logger:               logger,
 			reg:                  reg,
 			replicaLabels:        replicaLabels,
+			storeMatchers:        storeMatchers,
 			proxy:                proxy,
 			deduplicate:          deduplicate,
 			maxResolutionMillis:  maxResolutionMillis,
@@ -54,6 +56,7 @@ type queryable struct {
 	logger               log.Logger
 	reg                  prometheus.Registerer
 	replicaLabels        []string
+	storeMatchers        [][]storepb.LabelMatcher
 	proxy                storepb.StoreServer
 	deduplicate          bool
 	maxResolutionMillis  int64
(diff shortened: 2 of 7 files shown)


pkg/query/api/v1.go
```diff
@@ -141,6 +141,7 @@ type API struct {
 	enableQueryPartialResponse bool
 	enableRulePartialResponse  bool
 	replicaLabels              []string
+	storeAllowList             []string
```
bwplotka, line 144: Let's be explicit here. What about?
```suggestion
	storeAddressAllowlist             []string
```

pkg/query/api/v1.go
```diff
@@ -269,6 +270,21 @@ func (api *API) parseReplicaLabelsParam(r *http.Request) (replicaLabels []string
 	return replicaLabels, nil
 }
 
+func (api *API) parseStoreAllowListParam(r *http.Request) (storeAllowList []string, _ *ApiError) {
```
bwplotka, line 273: It's fine to treat `Allowlist` as one word so no need to do camel case for that.

But again it would be nice to put explicit `Address` there (:

pkg/query/api/v1.go
```diff
@@ -269,6 +270,21 @@ func (api *API) parseReplicaLabelsParam(r *http.Request) (replicaLabels []string
...
+	const storeAllowListParam = "storeAllowList[]"
```
bwplotka, line 274: Now when I think of it I wonder if something like matchers would be quite neat to have.
Because when we add allowlist we will soon have request "can I have denylist"? ;p

I mean something like this API have: https://prometheus.io/docs/prometheus/latest/querying/api/#finding-series-by-label-matchers

For example:
 
```
&storeMatch[]={__address__=~"1\.2\.3\.\d*"}
```
This will give us quite amazing flexibility as in future we can add more "labels" to maybe filter by source, health, or any other metadata ... what do you think? (:

pkg/query/api/v1.go
```diff
@@ -688,7 +719,7 @@ func (api *API) labelNames(r *http.Request) (interface{}, []error, *ApiError) {
 		return nil, nil, apiErr
 	}
 
-	q, err := api.queryableCreate(true, nil, 0, enablePartialResponse, false).Querier(ctx, math.MinInt64, math.MaxInt64)
+	q, err := api.queryableCreate(true, nil, nil, 0, enablePartialResponse, false).Querier(ctx, math.MinInt64, math.MaxInt64)
```
bwplotka, line 722: technically this param could be everywhere (:

pkg/store/proxy.go
```diff
@@ -255,8 +260,14 @@ func (s *ProxyStore) Series(r *storepb.SeriesRequest, srv storepb.Store_SeriesSe
 			// NOTE: all matchers are validated in matchesExternalLabels method so we explicitly ignore error.
 			var ok bool
 			tracing.DoInSpan(gctx, "store_matches", func(ctx context.Context) {
+				storeAllowList := make(map[string]struct{})
```
bwplotka, line 263: let's add todo to consider this in proto

pkg/query/api/v1.go
geobeau: Maybe it could rely on the same system as the external labels match:
- When a store is registered we add a label for `__thanos_store_address__` containing the address (and maybe a `__thanos_store_health__` as well)
- Then the user add the label in its regular query and we match it using the regular system
- Before we dispatch the query to the other store, we drop all the `__thanos_store_.*` labels

Pro:
- easy to implement
- easy to use

Con:
- magic is done is the background
- we reserve some labels

What do you think?

bwplotka: Yea I was thinking about this as well, but this sounds scary. I think this one will need more thoughts. Let me propagate this question among maintainers and devs. I think the matcher idea is less intrusive.

We could in theory add those labels only in some `--debug` mode :thinking:

bwplotka: We definitely cannot inject them as normal flow as it will surprise Querier clients. It's important to remember that users already build alerts etc so data consistency is the key

bwplotka: wonder what e.g @brian-brazil would say about our ideas =D

brian-brazil: I see no problems here with storeMatch - though I wonder if you really need regexes if this is just for ad-hoc debugging.

Tying this to general label semantics sounds like it's over-complicating things, and would likely cause problems down time line (plus is technically breaking as `__thanos_store_address__` is a valid label value that could come from Prometheus).

geobeau: Alright, let's go for `storeMatch` then! 👍

bwplotka: Yes. Although I really like the fact that you could aggregate, join etc all metadata. ;p

pkg/query/api/v1.go
geobeau: Indeed, I missed it!

pkg/query/api/v1.go
geobeau: I updated the PR. Much cleaner to use matcher instead of a regular allow list!

geobeau:
@bwplotka could you review it again? :)

pkg/query/querier.go
```diff
@@ -250,9 +256,20 @@ func (q *querier) selectFn(ctx context.Context, hints *storage.SelectHints, ms .
 	if err != nil {
 		return nil, errors.Wrap(err, "convert matchers")
 	}
+	var storeMatchers [][]storepb.LabelMatcher
+	for _, stms := range q.storeMatchers {
+		stm, err := storepb.TranslatePromMatchers(stms...)
```
bwplotka, line 261: Hm what about translating them in build time (start time of query) not query time? (:

pkg/store/proxy.go
```diff
@@ -498,14 +509,40 @@ func (s *streamSeriesSet) Err() error {
...
 	return labelSetsMatch(s.LabelSets(), matchers)
 }
 
+// storeMatch return true if the store's metadata labels match the storeMatcher.
+func storeMatchMetadata(s Client, storeMatcher [][]storepb.LabelMatcher) (bool, error) {
+	clientLabels := generateMetadataClientLabels(s)
+	if len(storeMatcher) != 0 {
+		res := false
+		for _, stm := range storeMatcher {
+			stmMatch, err := labelSetMatches(clientLabels, stm)
```
bwplotka, line 530: I think we can simplify this error handling, not sure why simple matching can produce error tbh ;p

pkg/store/proxy.go
```diff
@@ -498,14 +509,40 @@ func (s *streamSeriesSet) Err() error {
...
+			}
+			res = res || stmMatch
+		}
+		return res, nil
+	}
+	return true, nil
+}
+
+func generateMetadataClientLabels(s Client) storepb.LabelSet {
+	l := storepb.Label{Name: "__address__", Value: s.Addr()}
```
bwplotka, line 542: Looks like shallow function to me ):

bwplotka:
Rebase needed though ):

docs/components/query.md
```diff
@@ -222,6 +222,24 @@ Thanos Querier has the ability to perform concurrent select request per query. I
 The maximum number of concurrent requests are being made per query is controller by `query.max-concurrent-select` flag.
 Keep in mind that the maximum number of concurrent queries that are handled by querier is controlled by `query.max-concurrent`. Please consider implications of combined value while tuning the querier.
 
+### Store filtering
+
+It's possible to provide a set of matchers to the Querier api to select specific stores to be used during the query using the `storeMatch[]` parameter. It is useful when debugging a slow/broken store. It uses the same format as the matcher of [Prometheus' federate api](https://prometheus.io/docs/prometheus/latest/querying/api/#finding-series-by-label-matchers). Note that at the moment the querier only support the `__address__` which contains the address of the store.
```
bwplotka, line 227:
```suggestion
It's possible to provide a set of matchers to the Querier api to select specific stores to be used during the query using the `storeMatch[]` parameter. It is useful when debugging a slow/broken store. It uses the same format as the matcher of [Prometheus' federate api](https://prometheus.io/docs/prometheus/latest/querying/api/#finding-series-by-label-matchers). Note that at the moment the querier only support the `__address__` which contains the address of the store that were ... 
```
I think it would be actually IP address that is visible as Endpoint on Store API. Not necessarily domain as user can configure `dns://prometheus-foo.thanos-sidecar:10901` (:

pkg/store/proxy.go
geobeau: Is it better? The function is very similar to `labelSetsMatch()`

pkg/store/proxy.go
geobeau: It's in a separate function because I wanted to separate the matching from the generation of labels. It also make the function easier to extend with new labels or be used elsewhere.

docs/components/query.md
geobeau: Is it better like this?

bwplotka: yes! thx!

bwplotka: Can we link the docs & rebase? Then it's more than enough for merge :muscle: Thanks!

geobeau:
Rebased!

bwplotka:
Thanks!

geobeau:
Nice, thanks for the reviews!

PR #2934, opened 2020-07-24
docs: Add blog post from Beat

hwoarang:
Add blog post from Beat which uses Thanos to monitor kubernetes
clusters and AWS resources.

## Changes

Add blog post from Beat


hwoarang:
> Looks good however the DCO is missing! Care to add it? (:

oops added it :)

PR #2943, opened 2020-07-27
store/memcached: Mitigate bloating logs when async buffer is full

kakkoyun:

* [X] Change is not relevant to the end-user.

## Changes

* Change the log level for async buffer full error.
* Add a label to track the occurrence of the error.

## Verification

* `make test-local`

cc @pstibrany

### DIFF
--- a/pkg/cache/memcached.go
+++ b/pkg/cache/memcached.go
@@ -68,7 +68,7 @@ func (c *MemcachedCache) Store(ctx context.Context, data map[string][]byte, ttl 
 	}
 
 	if firstErr != nil {
-		level.Warn(c.logger).Log("msg", "failed to store one or more items into memcached", "failed", failed, "firstErr", firstErr)
+		level.Debug(c.logger).Log("msg", "failed to store one or more items into memcached", "failed", failed, "firstErr", firstErr)
 	}
 }
 
--- a/pkg/cache/memcached.go
+++ b/pkg/cache/memcached.go
@@ -68,7 +68,7 @@ func (c *MemcachedCache) Store(ctx context.Context, data map[string][]byte, ttl 
 	}
 
 	if firstErr != nil {
-		level.Debug(c.logger).Log("msg", "failed to store one or more items into memcached", "failed", failed, "firstErr", firstErr)
+		level.Warn(c.logger).Log("msg", "failed to store one or more items into memcached", "failed", failed, "firstErr", firstErr)
 	}
 }
 
--- a/pkg/cacheutil/memcached_client.go
+++ b/pkg/cacheutil/memcached_client.go
@@ -24,9 +24,10 @@ import (
 )
 
 const (
-	opSet             = "set"
-	opGetMulti        = "getmulti"
-	reasonMaxItemSize = "max-item-size"
+	opSet                 = "set"
+	opGetMulti            = "getmulti"
+	reasonMaxItemSize     = "max-item-size"
+	reasonAsyncBufferFull = "async-buffer-full"
 )
 
 var (
@@ -233,6 +234,8 @@ func newMemcachedClient(
 	}, []string{"operation", "reason"})
 	c.skipped.WithLabelValues(opGetMulti, reasonMaxItemSize)
 	c.skipped.WithLabelValues(opSet, reasonMaxItemSize)
+	c.skipped.WithLabelValues(opGetMulti, reasonAsyncBufferFull)
+	c.skipped.WithLabelValues(opSet, reasonAsyncBufferFull)
 
 	c.duration = promauto.With(reg).NewHistogramVec(prometheus.HistogramOpts{
 		Name:    "thanos_memcached_operation_duration_seconds",
@@ -276,7 +279,7 @@ func (c *memcachedClient) SetAsync(_ context.Context, key string, value []byte, 
 		return nil
 	}
 
-	return c.enqueueAsync(func() {
+	err := c.enqueueAsync(func() {
 		start := time.Now()
 		c.operations.WithLabelValues(opSet).Inc()
 
@@ -297,6 +300,13 @@ func (c *memcachedClient) SetAsync(_ context.Context, key string, value []byte, 
 
 		c.duration.WithLabelValues(opSet).Observe(time.Since(start).Seconds())
 	})
+
+	if err == errMemcachedAsyncBufferFull {
+		c.skipped.WithLabelValues(opSet, reasonAsyncBufferFull).Inc()
+		level.Debug(c.logger).Log("msg", "failed to store item to memcached, check your configuration", "err", err, "size", len(c.asyncQueue))
+		return nil
+	}
+	return err
 }
 
 func (c *memcachedClient) GetMulti(ctx context.Context, keys []string) map[string][]byte {
--- a/pkg/cacheutil/memcached_client.go
+++ b/pkg/cacheutil/memcached_client.go
@@ -234,7 +234,6 @@ func newMemcachedClient(
 	}, []string{"operation", "reason"})
 	c.skipped.WithLabelValues(opGetMulti, reasonMaxItemSize)
 	c.skipped.WithLabelValues(opSet, reasonMaxItemSize)
-	c.skipped.WithLabelValues(opGetMulti, reasonAsyncBufferFull)
 	c.skipped.WithLabelValues(opSet, reasonAsyncBufferFull)
 
 	c.duration = promauto.With(reg).NewHistogramVec(prometheus.HistogramOpts{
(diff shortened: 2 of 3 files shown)


pkg/cacheutil/memcached_client.go
```diff
@@ -297,6 +300,13 @@ func (c *memcachedClient) SetAsync(_ context.Context, key string, value []byte,
 
 		c.duration.WithLabelValues(opSet).Observe(time.Since(start).Seconds())
 	})
+
+	if err == errMemcachedAsyncBufferFull {
+		c.skipped.WithLabelValues(opSet, reasonAsyncBufferFull).Inc()
```
pracucci, line 305: Could you initialise it in `newMemcachedClient()` similarly to what we do with `c.skipped.WithLabelValues(opSet, reasonMaxItemSize)`, please?

pkg/cacheutil/memcached_client.go
```diff
@@ -297,6 +300,13 @@ func (c *memcachedClient) SetAsync(_ context.Context, key string, value []byte,
...
+		level.Debug(c.logger).Log("msg", "failed to store item to memcached, check your configuration", "err", err, "size", len(c.asyncQueue))
```
pracucci, line 306:
```suggestion
		level.Debug(c.logger).Log("msg", "failed to store item to memcached because the async buffer is full", "err", err, "size", len(c.asyncQueue))
```

pkg/cache/memcached.go
```diff
@@ -68,7 +68,7 @@ func (c *MemcachedCache) Store(ctx context.Context, data map[string][]byte, ttl
 	}
 
 	if firstErr != nil {
-		level.Warn(c.logger).Log("msg", "failed to store one or more items into memcached", "failed", failed, "firstErr", firstErr)
+		level.Debug(c.logger).Log("msg", "failed to store one or more items into memcached", "failed", failed, "firstErr", firstErr)
```
pracucci, line 71: Is it really problematic to you? We find it quite handy in Cortex :)

pkg/cacheutil/memcached_client.go
kakkoyun: Already did.

pkg/cache/memcached.go
kakkoyun: Actually I've forgotten to clean this one. Since we handled it in Memcached client layer we can keep it as Warn.
It has created issues for a couple of users when they misconfigured their buffer size. Do you want to log async buffer over flows?

GiedriusS: I thought this was useful too but the spam is too much. I was thinking more about adding some kind of rate-limiting to this place so that the logs wouldn't appear too many times.

kakkoyun: @GiedriusS @pracucci I have changed level back to `Warn` again, however this PR still omits async buffer overflow error. How should we proceed with this? Rate-limiting or logging every X message for this error is ok for me, would you prefer that?

pkg/cacheutil/memcached_client.go
```diff
@@ -233,6 +234,8 @@ func newMemcachedClient(
 	}, []string{"operation", "reason"})
 	c.skipped.WithLabelValues(opGetMulti, reasonMaxItemSize)
 	c.skipped.WithLabelValues(opSet, reasonMaxItemSize)
+	c.skipped.WithLabelValues(opGetMulti, reasonAsyncBufferFull)
```
pracucci, line 237: This shouldn't be a case. Gets are always sync, while sets are async.

pkg/cache/memcached.go
pracucci: > Rate-limiting or logging every X message for this error is ok for me, would you prefer that?

I would prefer rate-limiting, preferably with the next error log will also display the count of omitted ones. All in all, what you want to protect from is a flood of logs; logging 100% of messages is fine if the rate is low.

pkg/cacheutil/memcached_client.go
```diff
@@ -414,6 +416,8 @@ func (c *memcachedClient) enqueueAsync(op func()) error {
 	case c.asyncQueue <- op:
 		return nil
 	default:
+		c.skipped.WithLabelValues(opSet, reasonAsyncBufferFull).Inc()
+		level.Debug(c.logger).Log("msg", "the async buffer is full", "size", len(c.asyncQueue))
```
bwplotka, line 420: Code smell: Returning error and logging :thinking:

pkg/logging/limit.go
```diff
@@ -0,0 +1,30 @@
...
+func (r *rateLimitedLogger) Log(keyvals ...interface{}) error {
+	r.occurrence++
+	if r.limiter.Allow() {
+		return log.With(r.logger, "occurrence", r.occurrence).Log(keyvals...)
+	}
+	return nil
+}
+
+// Limit creates a new logger with rate limiting using given time window and burst parameters.
+func Limit(logger log.Logger, every time.Duration, b int) log.Logger {
```
bwplotka, line 28: It's not `every` but `limit` (: Every would mean it will produce logs in interval.

pkg/logging/limit.go
```diff
@@ -0,0 +1,30 @@
...
+	"golang.org/x/time/rate"
+)
+
+type rateLimitedLogger struct {
+	logger     log.Logger
+	limiter    *rate.Limiter
+	occurrence int
+}
+
+func (r *rateLimitedLogger) Log(keyvals ...interface{}) error {
```
bwplotka, line 19: We probably need to be bit smarter here.

* I would try to limit the same messages. Doing `level.Warn(logging.With(logger, limit1)).Log` does the work well so all good
* I would somehow don't lose the info about lost messages as we rely on them. Either some metrics for dropped logs or better: Another field to this message saying `occured` to mention how many of those occurred during last log time.

pkg/logging/limit.go
```diff
@@ -0,0 +1,30 @@
...
+	r.occurrence++
```
bwplotka, line 20: Has to be atomic most likely

pkg/logging/limit.go
kakkoyun: How about now?

pkg/logging/limit.go
```diff
@@ -0,0 +1,70 @@
...
+
+	m           sync.Mutex
+	occurrences map[string]*occurrence
+}
+
+// Log limits the amount of events logged and keep tracks of occurrences of messages.
+func (r *rateLimitedLogger) Log(keyvals ...interface{}) error {
+	var msg string
+	for i := range keyvals {
+		if i%2 == 0 && keyvals[i] == messageKey {
```
bwplotka, line 33: This is something it would be awesome to avoid. buffering tons of memory in logger does not sounds like a good idea (: 

Having rate-limiting for single messages sounds better to me.

pkg/cacheutil/memcached_client.go
```diff
@@ -414,6 +416,7 @@ func (c *memcachedClient) enqueueAsync(op func()) error {
 	case c.asyncQueue <- op:
 		return nil
 	default:
+		c.skipped.WithLabelValues(opSet, reasonAsyncBufferFull).Inc()
```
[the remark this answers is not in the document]
bwplotka: In case of this metric.. can we just drop `level.Warn(limitedLogger).Log("msg", "failed to store one or more items into memcached", "failed", failed, "firstErr", firstErr)`?

It sounds to me like metric is enough.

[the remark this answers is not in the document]
bwplotka: This is again why handling error multiple times is tricky as you can by accident replicate this error many times (we inc metric and return err in same time)

pkg/logging/limit.go
```diff
@@ -0,0 +1,70 @@
...
+	}
+	if r.limiter.Allow() {
+		if msg != "" {
+			r.m.Lock()
+			defer r.m.Unlock()
+
+			o := r.occurrences[msg]
+			defer func() { o.lastSeen = o.current }()
+
+			return log.With(r.logger, "occurred", o.current, "since last seen", o.current-o.lastSeen).Log(keyvals...)
```
bwplotka, line 54: All of this to me means like we are re- implementing just counter.. Sorry for shitty suggestion ;p I think it should be just metric then.

pkg/cache/memcached.go
```diff
@@ -67,8 +68,9 @@ func (c *MemcachedCache) Store(ctx context.Context, data map[string][]byte, ttl
 		}
 	}
 
+	limitedLogger := logging.Limit(c.logger, time.Second, 1)
```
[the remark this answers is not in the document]
bwplotka: or actually remove... (: as we can tell  that from metric already

pkg/cache/memcached.go
bwplotka: Looks like I proposed something opposite (https://github.com/thanos-io/thanos/pull/2943#discussion_r461648356), let's talk offline ;p

PR #2944, opened 2020-07-27
mixin: Separate Read/Write path dashboards for receive

kakkoyun:

## Changes

* Separate Read/Write path dashboards for Receiver.
* Remove detailed panels.
* Clean up.

## TODO

* Add screenshots.

## Verification

* `make examples` and import Grafana.

### DIFF
--- a/examples/dashboards/overview.json
+++ b/examples/dashboards/overview.json
@@ -62,10 +62,10 @@
                "steppedLine": false,
                "targets": [
                   {
-                     "expr": "sum(label_replace(rate(http_requests_total{namespace=\"$namespace\",job=~\"thanos-query.*\",handler=\"query\"}[$interval]),\"status_code\", \"${1}xx\", \"code\", \"([0-9])..\")) by (job, status_code)",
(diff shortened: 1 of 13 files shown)


kakkoyun:
@yeya24 What we observed from our use case, most of the times they are not used. I moved method based aggregations to the panels we have and removed the detailed panels. I'm happy to separate the PRs. 

I just want to reduce confusion and create a single path for troubleshooting through dashboards.
I'm open to discussions. What do you think? Do you see value having them around?

kakkoyun:
@yeya24 Since you already imported them, I'd be glad if you could add some screenshots 😅

yeya24:
> @yeya24 What we observed from our use case, most of the times they are not used. I moved method based aggregations to the panels we have and removed the detailed panels. I'm happy to separate the PRs.
> 
> I just want to reduce confusion and create a single path for troubleshooting through dashboards.
> I'm open to discussions. What do you think? Do you see value having them around?

Makes sense! 

> @yeya24 Since you already imported them, I'd be glad if you could add some screenshots

Will do.

yeya24:
Screenshots for receive.

Screenshot from 2020-07-28 06-05-54
Screenshot from 2020-07-28 06-06-16

kakkoyun:
@yeya24 Thanks for the screenshots.

kakkoyun:
@brancz What do you think?

Issue #2496, opened 2020-04-22
Thanos sidecar: internal server shutdown after changes within a config file (via a configmap)


hawran:
**Thanos, Prometheus and Golang version used**:
thanos: 0.12.0
prometheus: 2.17.1
go: 1.13

NB I'm aware of the https://github.com/thanos-io/thanos/releases/tag/v0.12.1 release, however regarding fixes within that release I presume this issue is still here.

**Object Storage Provider**:
ceph s3

**What happened**:
A thanos sidecar gets restarted from time to time when a config file has been changed (as a reloaded configmap).

**What you expected to happen**:
No restarts after such changes.

**How to reproduce it (as minimally and precisely as possible)**:
Run a pod with both prometheus and thanos sidecar containers and update a config file (a configmap).

**Full logs to relevant components**:
_RESTARTS_
```
Time kubernetes.pod.name kubernetes.container.name log
April 21st 2020, 15:09:05.046 prometheus-1 thanos level=warn ts=2020-04-21T13:09:05.046664068Z caller=intrumentation.go:54 msg="changing probe status" status=not-ready reason="build hash: lstat /etc/prometheus/rules/..2020_04_17_06_44_43.813421346/config.yaml: no such file or directory"
April 21st 2020, 15:09:05.047 prometheus-1 thanos level=warn ts=2020-04-21T13:09:05.047028972Z caller=intrumentation.go:54 msg="changing probe status" status=not-ready reason="build hash: lstat /etc/prometheus/rules/..2020_04_17_06_44_43.813421346/config.yaml: no such file or directory"
April 21st 2020, 15:09:05.047 prometheus-1 thanos level=info ts=2020-04-21T13:09:05.046979769Z caller=intrumentation.go:66 msg="changing probe status" status=not-healthy reason="build hash: lstat /etc/prometheus/rules/..2020_04_17_06_44_43.813421346/config.yaml: no such file or directory"
April 21st 2020, 15:09:05.047 prometheus-1 thanos level=info ts=2020-04-21T13:09:05.046947111Z caller=http.go:81 service=http/server component=sidecar msg="internal server shutdown" err="build hash: lstat /etc/prometheus/rules/..2020_04_17_06_44_43.813421346/config.yaml: no such file or directory"
[...]

PR #2952, opened 2020-07-28
sidecar: fix sidecar crashing with lstat errors during a reload from a configmap update

ebabani:
During a reload it's possible that files are added/removed while
we walk the directory and we try to lstat a file that doesn't exist.

This is especially common if running in k8s with the configuration mounted
as a configmap, and the configmap contents change.

With this change we log when this happens but don't return the error so the
reloader process doesn't crash.

## Changes

Add an error check during a reload for when a watch triggers on a file that gets removed. This fixes the issue reported in https://github.com/thanos-io/thanos/issues/2496

## Verification

Added a test to reproduce the initial error and to verify the fix

### DIFF
--- a/pkg/reloader/reloader.go
+++ b/pkg/reloader/reloader.go
@@ -218,8 +218,6 @@ func (r *Reloader) Watch(ctx context.Context) error {
 		}
 
 		if err := r.apply(ctx); err != nil {
-			// Critical error.
-			// TODO(bwplotka): There is no need to get process down in this case and decrease availability, handle the error in different way.
 			return err
 		}
 	}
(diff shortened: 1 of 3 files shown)


ebabani:
Linter didn't like the Canadian spelling of behaviour. Pushing a fix.

pkg/reloader/reloader.go
```diff
@@ -218,9 +218,15 @@ func (r *Reloader) Watch(ctx context.Context) error {
...
-			// Critical error.
-			// TODO(bwplotka): There is no need to get process down in this case and decrease availability, handle the error in different way.
-			return err
+			switch err := errors.Cause(err).(type) {
+			case *os.PathError:
+				r.watchErrors.Inc()
+				level.Error(r.logger).Log("msg", "path error", "err", err)
+			default:
+				// Critical error.
+				// TODO(bwplotka): There is no need to get process down in this case and decrease availability, handle the error in different way.
```
kakkoyun, line 227: Maybe we should still address this as well. There's no need to get the process down. Maybe decision can be made by the caller side.

pkg/reloader/reloader.go
```diff
@@ -218,9 +218,15 @@ func (r *Reloader) Watch(ctx context.Context) error {
 		}
 
 		if err := r.apply(ctx); err != nil {
-			// Critical error.
-			// TODO(bwplotka): There is no need to get process down in this case and decrease availability, handle the error in different way.
-			return err
+			switch err := errors.Cause(err).(type) {
+			case *os.PathError:
+				r.watchErrors.Inc()
```
kakkoyun, line 223: To make these errors more visible, we can add labels to `watchErrors` and increment counter by the type.

ebabani:
Will push an update tomorrow. Need to think about how to handle the errors, there's a lot of different filesystem related ones.

ebabani:
I was thinking a bit more about this and reloader errors feel pretty important since this component is responsible for rendering the final config.

With my original approach I was only handling one type of filesystem error but I bet there's others. I updated this PR to add a short retry mechanism when calculating the rules hash and it hits an error. If the error still happens after the retries it's worth returning it to the caller instead of potentially swallowing it and reporting a metric that someone might not be checking.

I'm not fully satisfied since the retry is not yet configurable(couldn't fine a nice way to pass that config) but feels better than the old approach.

kakkoyun:
@ebabani Could you have another look at this after the recent changes? We have an upcoming release, it'd be great to have this. We plan to cut it in today or tomorrow.

ebabani:
@kakkoyun I verified that with the recent retry updates in #2996 we avoid the original lstat problem. Since it's a temporary issue when symlinks are changing retrying on config update errors is good enough.

I think this PR can be closed since #2996 does essentially the same thing (with a nicer config 🙂  )

kakkoyun:
@ebabani Great to hear that the issue is fixed for you. Can we make use of the test that we have in this PR to prevent any regressions? If so we can merge the PR only with the tests.

ebabani:
That would be covered by the test for general FS errors at https://github.com/thanos-io/thanos/blob/master/pkg/reloader/reloader_test.go#L145-L151

kakkoyun:
Closing this since the issue fixed by #2996

PR #2955, opened 2020-07-29
fix the grpcurl script

krasi-georgiev:
Signed-off-by: Krasi Georgiev <8903888+krasi-georgiev@users.noreply.github.com>

### DIFF
--- a/scripts/insecure_grpcurl_series.sh
+++ b/scripts/insecure_grpcurl_series.sh
@@ -52,11 +52,10 @@ SERIES_REQUEST='{
 
 GOGOPROTO_ROOT="$(GO111MODULE=on go list -f '{{ .Dir }}' -m github.com/gogo/protobuf)"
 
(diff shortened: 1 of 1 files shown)


scripts/insecure_grpcurl_series.sh
```diff
@@ -52,11 +52,10 @@ SERIES_REQUEST='{
 
 GOGOPROTO_ROOT="$(GO111MODULE=on go list -f '{{ .Dir }}' -m github.com/gogo/protobuf)"
 
-pushd ${DIR}/../pkg/store/storepb/
```
bwplotka, line 55: curious why this was not working :thinking: It's the same thing.

krasi-georgiev: yes this change is not relevant, just thought it is more simple as not everyone(including me) is familiar with pushd and popd

Issue #2945, opened 2020-07-27
Fetch remaining data if chunk is bigger than 16000 bytes


pstibrany:
We have found a case where compactor can generate chunks [bigger than 16k](https://github.com/prometheus/prometheus/issues/5862). Currently when Thanos finds that it has loaded less than data than chunk length, it [returns error](https://github.com/thanos-io/thanos/blob/a64884b19dbe7a6c2613078a07409c89ce4e2508/pkg/store/bucket.go#L2029).

Instead of returning error, I propose to fetch remaining data instead in such case. Thoughts?

(This has been discussed at https://cloud-native.slack.com/archives/CL25937SP/p1595842211384500 as well)

GiedriusS:
Fixed by https://github.com/thanos-io/thanos/pull/2956

PR #2956, opened 2020-07-29
Fetching large chunks

pstibrany:

## Changes

This PR fixes issue with fetching chunks that are bigger than 16000 bytes. Under some circumstances it is possible to come across chunks like that, and under some conditions Thanos may currently fail to fetch such chunks. This PR adds fetching of remaining chunk data from the bucket.

## Verification

Added unit test. Also tested manually against real block found in our test cluster.

Fixes issue https://github.com/thanos-io/thanos/issues/2945.

### DIFF
--- a/pkg/store/bucket.go
+++ b/pkg/store/bucket.go
@@ -2026,21 +2026,29 @@ func (r *bucketChunkReader) preload() error {
 	return g.Wait()
 }
 
+// loadChunks will read range [start, end] from the segment file with sequence number seq.
+// This data range covers chunks starting at supplied offsets.
 func (r *bucketChunkReader) loadChunks(ctx context.Context, offs []uint32, seq int, start, end uint32) error {
-	begin := time.Now()
+	fetchBegin := time.Now()
 
 	b, err := r.block.readChunkRange(ctx, seq, int64(start), int64(end-start))
 	if err != nil {
 		return errors.Wrapf(err, "read range for %d", seq)
 	}
 
+	locked := true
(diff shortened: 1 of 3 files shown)


pkg/store/bucket.go
```diff
@@ -2050,11 +2058,44 @@ func (r *bucketChunkReader) loadChunks(ctx context.Context, offs []uint32, seq i
...
+		if len(cb) >= chLen {
+			r.chunks[chunkRef] = rawChunk(cb[n:chLen])
+			continue
 		}
-		cid := uint64(seq<<32) | uint64(o)
-		r.chunks[cid] = rawChunk(cb[n : n+int(l)+1])
+
+		// If we didn't fetch enough data for the chunk, fetch more. This can only really happen for last
+		// chunk in the list of fetched chunks, otherwise partitioner would merge fetch ranges together.
+		r.mtx.Unlock()
```
bwplotka, line 2074: Worth to mention what this mutex is saving from. Probably changing name would be nice. I know it was like this before, but maybe we can improve that part. I assume it's for stats, right?

pstibrany: It's both for stats and storing chunks. I will document and rename it.

pstibrany: Also `chunkBytes`.

bwplotka:
Thanks!

pracucci:
I'm joining the party very late, but good job @pstibrany ! 👏

Issue #2938, opened 2020-07-24
Rules HTTP API weirdness: nulls, different values for "same" field, lastEvaluation set to 0000


GiedriusS:
On top of:

https://github.com/thanos-io/thanos/pull/2865/commits/b2f1485e3239be07c93e68b404070d48b214965b from #2865 

With:
```
TZ=UTC PROMETHEUS_EXECUTABLE=~/dev/prometheus/prometheus ./scripts/quickstart.sh
```

Going to `localhost:19999/api/v1/rules?type=alert` gives:

```
{"status":"success","data":{"groups":[{"name":"example","file":"data/rules.yml","rules":null,"interval":30,"evaluationTime":0,"lastEvaluation":"0001-01-01T00:00:00Z","partial_response_strategy":"WARN","partialResponseStrategy":"ABORT"}]}}
```

Three things seem off to me:
* `rules` is null even though the branch contains https://github.com/thanos-io/thanos/pull/2888 which supposedly fixes this problem
* Two fields `partial_response_strategy` and `partialResponseStrategy` that should show the same value but they are different in this case :thinking: 
* `lastEvaluation` is _always_ set to `0001-01-01T00:00:00Z`

yeya24:
@GiedriusS Is this fixed by https://github.com/thanos-io/thanos/pull/2925 already? 😄

PR #2957, opened 2020-07-29
rules: fix silly mistakes in the rules API

GiedriusS:
Fixes https://github.com/thanos-io/thanos/issues/2938 and is a follow up to https://github.com/thanos-io/thanos/pull/2925.

Changes:
* Copying all fields properly to the group-level as well;
* Set TZ to UTC explicitly in other `time.Time` fields to avoid a panic;
* Make sure that `RuleGroup` always has an empty array in the `rules`
field because that's customary and that is what the new React UI
expects;
* Remove `DeprecatedPartialResponseStrategy` so that we wouldn't
have essentially two fields in the response of `/api/v1/rules` that could
_potentially_ show different values because one of them is locked to
always be `WARN`.

The new e2e test case tests out the first change written in this list and the last one
is covered by unit tests that were changed to work with these modifications. 


pkg/rules/rulespb/rpc.proto
```diff
@@ -73,7 +73,6 @@ message RuleGroup {
     google.protobuf.Timestamp last_evaluation = 6 [(gogoproto.jsontag) = "lastEvaluation", (gogoproto.stdtime) = true, (gogoproto.nullable) = false ];
 
     // Thanos specific.
-    PartialResponseStrategy DeprecatedPartialResponseStrategy = 7 [(gogoproto.jsontag) = "partial_response_strategy"];
```
bwplotka, line 76: Yea, you are right ;p

Issue #1811, opened 2019-11-28
querier: Limit LabelNames; LabelValues to certain time period.


bwplotka:
Something that Cortex is doing is that `LabelNames;LabelValues` by default are limited to active series only on ingesters (so let's say in Thanos world to Thanos receiver or Prometheus for latest 2h). This is due to the fact that when you query you are usually interested in autocompletion for fresh metrics.

We are now exploring if we can apply the same logic for Thanos e.g limit `LabelNames` and `LabelValues` and `Series` (by default) on Querier QueryAPI to 2h. 

IF we don't do it, StoreGateway will have to load portions of the blocks in memory for full-time retention, for all store APIs, which is something we wanted to avoid. It was briefly mentioned here: https://github.com/thanos-io/thanos/issues/39 Also currently this is limited to raw data only, so it's already scoped for certain setups with retention for older blocks.

ppanyukov:
Do I understand correctly this does not affect the query execution and any memory limits for that?

bwplotka:
Which memory limits you have in mind?

ppanyukov:
The limits on memory taken during query execution, as this PR attempts to address.

https://github.com/thanos-io/thanos/pull/1747

bwplotka:
Yes, I still don't see how limits improve those methods. This issue is required to have fulfill this: https://github.com/thanos-io/thanos/issues/1813

pracucci:
I would be glad to work on this. Few questions:

- Should the period be configurable?
- Should we introduce a breaking change enforcing a default of `2h`, or should we keep an unlimited default just introducing an option to slim it down?

> We are now exploring if we can apply the same logic for Thanos e.g limit LabelNames and LabelValues and Series (by default) on Querier QueryAPI to 2h.

FYI, Cortex doesn't enforce any short default for the `Series`. It currently uses Prometheus defaults, which is the entire history. Do we want to change this in Thanos, and enforce a default of `2h` for `Series` as well?

bwplotka:
Nice! 
> Should the period be configurable?
yes
> Should we introduce a breaking change enforcing a default of 2h, or should we keep an unlimited default just introducing an option to slim it down?
We can introduce a breaking change if needed, but we need to be explicit in changelog. The question is, should we?

[...]

PR #2964, opened 2020-07-31
Add time range parameters to labels API

yeya24:

## Changes

Fixes: #1811

* [x] Add changelog
* [x] Update unit tests and E2E tests.


yeya24:
This pr is ready for review!

pkg/api/query/v1.go
```diff
@@ -394,12 +388,27 @@ func (qapi *QueryAPI) labelValues(r *http.Request) (interface{}, []error, *api.A
...
 
+	start, err := parseTimeParam(r, "start", minTime)
+	if err != nil {
+		return nil, nil, &api.ApiError{Typ: api.ErrorBadData, Err: err}
+	}
+	end, err := parseTimeParam(r, "end", maxTime)
+	if err != nil {
+		return nil, nil, &api.ApiError{Typ: api.ErrorBadData, Err: err}
+	}
+	// Prometheus doesn't check this, do we need to?
```
bwplotka, line 399: Sounds useful, no? (: 

Please ask comments on PR itself not in comment (: So we can still merge PR if code is ok.

pkg/store/storepb/rpc.proto
```diff
@@ -144,6 +144,10 @@ message LabelNamesRequest {
 
   // TODO(bwplotka): Move Thanos components to use strategy instead. Including QueryAPI.
   PartialResponseStrategy partial_response_strategy = 2;
+
+  int64 start = 3;
```
bwplotka, line 148: :call_me_hand:

pkg/api/query/v1.go
GiedriusS: I think we should and it's probably worth contributing such a check upstream IMHO

yeya24:
I have addressed the comment. Please take a look again. :smile:

PR #2968, opened 2020-07-31
ci: Enable prealloc

kakkoyun:

This PR enables `prealloc` linter to hint possible pre-allocations. Tests and complex code pieces are skipped.

## Changes

* Enable prealloc
* Fix code accordingly

## Verification

* `make go-lint`

### DIFF
--- a/pkg/receive/handler_test.go
+++ b/pkg/receive/handler_test.go
@@ -23,9 +23,10 @@ import (
 	"github.com/prometheus/prometheus/pkg/labels"
 	"github.com/prometheus/prometheus/storage"
 	terrors "github.com/prometheus/prometheus/tsdb/errors"
+	"google.golang.org/grpc"
+
 	"github.com/thanos-io/thanos/pkg/store/storepb"
 	"github.com/thanos-io/thanos/pkg/store/storepb/prompb"
-	"google.golang.org/grpc"
 )
 
 func TestCountCause(t *testing.T) {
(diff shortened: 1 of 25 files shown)


pkg/receive/handler_test.go
```diff
@@ -145,7 +146,7 @@ func newHandlerHashring(appendables []*fakeAppendable, replicationFactor uint64)
 			Hashring: "test",
 		},
 	}
-	var handlers []*Handler
+	var handlers []*Handler //nolint:prealloc
```
bwplotka, line 149: kind of painful ;p

kakkoyun:
@bwplotka Yeah, for the parts that don't need any allocation, it could be tedious. However, it helps review to catch those things actually.

bwplotka:
I am not sure about this (: Let's see what would be opinions of others on this, seems strict (: @squat @brancz @GiedriusS @povilasv @yeya24

GiedriusS:
Seems like this would bring marginal gain in performance while hurting readability and I could see this being a bit annoying. I wonder if we could devise some smart way to only check code with this in hot code paths because that's where it matters the most? :s

kakkoyun:
I don't have strong feelings about this. If it doesn't provide any value I can close it.

brancz:
I'd be interested in knowing if what @GiedriusS suggested is possible. I think this makes a lot of sense for those parts of the code where small performance tunings make a big difference.

kakkoyun:
As far as I can find out there's not enough knobs to configure this. One can run it against the files on the path manually. So closing this one, since it has gotten enough traction. And I agree that it could be annoying.

PR #2969, opened 2020-07-31
ci: Add shellcheck and shfmt

kakkoyun:

## Changes

* Add https://github.com/mvdan/sh
* Add https://github.com/koalaman/shellcheck
* Fix scripts accordingly

## Verification

* `make shell-format`
* `make shell-lint`


scripts/build-react-app.sh
```diff
@@ -5,13 +5,13 @@
 set -e
 set -u
 
-if ! [[ "$0" =~ "scripts/build-react-app.sh" ]]; then
-       echo "must be run from repository root"
-       exit 255
+if ! [[ $0 == "scripts/build-react-app.sh" ]]; then
```
GiedriusS, line 8: `Makefile` has:
```
 79        @./scripts/build-react-app.sh                                                                                                                                                                             
```
And `$0` in this case is `./scripts/build-react-app.sh` so I'm not sure how this works :thinking: at least on my PC it doesn't work. Bash version is:
```
GNU bash, version 5.0.17(1)-release (x86_64-redhat-linux-gnu)
```

kakkoyun: 🤔 🤔  I wonder how I didn't come across this. I can send a follow up PR to fix it.

GiedriusS: It would be nice. I wonder too how our CI didn't catch this. Probably because this target is run only manually? But then how we check in our bin data that there aren't any unstaged changes? If you won't then I'll check it out myself when I'll have some time :smile:

PR #2972, opened 2020-08-03
Remove parseStoreMatchers in query labelNames API

yeya24:

## Changes

store matchers only relate to the `Series` gRPC method, so it shouldn't be added in the labelNames API.


yeya24:
> Why not? (: It also goes through various stores, no? It's nice that all those fanout procies allows debug matching like this?

Yes, supporting store matches for other APIs are good. But since that is not supported, it is strange to parse store matchers in the labelNames API. If we want to implement it, we can have separate issues and I will close this pr.

PR #2973, opened 2020-08-04
Add thanos query frontend sub command

yeya24:

## Changes

The current code is almost the same as the Cortex frontend. I just removed the sharding middleware because seems that is specific for Cortex.

TODO:
- [ ] Implement Cortex cache.Cache interface, this should be done in another pr.
- [x] Unit tests and E2E tests.
- [x] Support parsing Thanos specific query parameters such as `dedup`, `max_source_resolution`, `partial_response`, etc. These params should be part of the cache key because these affect the query results. We need to implement Thanos codec to decode/encode Thanos query requests. https://github.com/yeya24/thanos/pull/159


cmd/thanos/query-frontend.go
```diff
@@ -0,0 +1,174 @@
...
+func (c *queryRangeConfig) registerFlag(cmd *kingpin.CmdClause) {
+	c.respCacheConfig.registerFlag(cmd)
+
+	cmd.Flag("query-range.cache-results", "Cache query range results.").Default("false").
+		BoolVar(&c.cacheResults)
+
+	cmd.Flag("query-range.split-interval", "Split queries by an interval and execute in parallel, 0 disables it.").
+		Default("24h").DurationVar(&c.splitInterval)
+
+	cmd.Flag("query-range.max-retries-per-request", "Maximum number of retries for a single request; beyond this, the downstream error is returned.").
```
bwplotka, line 64: So retries does not work for us API. (sad), so we need to rework this and for now maybe remove this.

cmd/thanos/query-frontend.go
```diff
@@ -0,0 +1,174 @@
...
+		Default("5").IntVar(&c.maxRetries)
+
+	cmd.Flag("query-range.disable-step-align", "Disable aligning incoming queries' start and end with their step.").
```
bwplotka, line 67: Let's remove this option and only expose this if someone will need. In practice you really want true. Plus Grafana does that by default

cmd/thanos/query-frontend.go
yeya24: This is used in query range as well https://github.com/cortexproject/cortex/blob/master/pkg/querier/queryrange/roundtrip.go#L192, not only in the queue mode. Can you please check it?

yeya24:
Thanos Query API supports many custom parameters that Prometheus doesn't support, like `dedup`, `partial_response`, `max_source_resolution`, `storeMatch`, etc.

IMHO Thanos should implement its own Request type just like the [loki request](https://github.com/grafana/loki/blob/master/pkg/querier/queryrange/codec.go). Is it better for me to include it in this pr, or in another pr?

brancz:
> IMHO Thanos should implement its own Request type just like the loki request. Is it better for me to include it in this pr, or in another pr?

I haven't had the chance to review this PR in itself but I think since this work will be on-going I think I would prefer to merge a minimum state that works good enough :tm:, and then iterate on it. So to be explicit, I am for this, but in a follow up PR.

cmd/thanos/query-frontend.go
```diff
@@ -0,0 +1,207 @@
...
+}
+
+type responseCacheConfig struct {
+	cacheMaxFreshness time.Duration
+	fifoCache         fifoCacheConfig
+}
+
+func (c *responseCacheConfig) registerFlag(cmd *kingpin.CmdClause) {
+	c.fifoCache.registerFlag(cmd)
+	cmd.Flag("query-range.response-cache-max-freshness", "Most recent allowed cacheable result, to prevent caching very recent results that might still be in flux.").
```
bwplotka, line 55: I believe this should be overall general config not fifo? :thinking:

cmd/thanos/query-frontend.go
```diff
@@ -0,0 +1,207 @@
...
+		Default("1m").DurationVar(&c.cacheMaxFreshness)
+}
+
+// fifoCacheConfig defines configurations for Cortex fifo cache.
+type fifoCacheConfig struct {
+	maxSizeBytes units.Base2Bytes
```
bwplotka, line 61: BTW I think we should have those as cache client now and use our cache flags like in: https://github.com/thanos-io/thanos/blob/82cca562ce9e04e6ed6cbef41be71fc4b2d88207/cmd/thanos/store.go#L54

cmd/thanos/query-frontend.go
```diff
@@ -0,0 +1,207 @@
...
+func registerQueryFrontend(m map[string]setupFunc, app *kingpin.Application) {
+	comp := component.QueryFrontend
+	cmd := app.Command(comp.String(), "query frontend")
+	conf := &queryFrontendConfig{}
+	conf.registerFlag(cmd)
+
+	m[comp.String()] = func(g *run.Group, logger log.Logger, reg *prometheus.Registry, _ opentracing.Tracer, _ <-chan struct{}, _ bool) error {
+
+		return runQueryFrontend(
+			g,
```
bwplotka, line 114: no need for this syle - I think it can fit a line (:

cmd/thanos/query-frontend.go
```diff
@@ -0,0 +1,207 @@
...
+	}
+	defer fe.Close()
+
+	limits := queryfrontend.NewLimits(
+		conf.queryRangeConfig.maxQueryParallelism,
+		time.Duration(conf.queryRangeConfig.maxQueryLength),
+		conf.queryRangeConfig.respCacheConfig.cacheMaxFreshness,
+	)
+
+	// TODO(yeya24): support other cache when available.
```
bwplotka, line 151: :+1:

yeya24:
Some build errors got after rebasing master.

```
# github.com/cortexproject/cortex/pkg/querier/queryrange
../../../go/pkg/mod/github.com/cortexproject/cortex@v1.2.1-0.20200805064754-d8edc95e2c91/pkg/querier/queryrange/value.go:92:2: invalid case parser.ValueTypeVector in switch on promRes.Data.ResultType (mismatched types parser.ValueType and string)
../../../go/pkg/mod/github.com/cortexproject/cortex@v1.2.1-0.20200805064754-d8edc95e2c91/pkg/querier/queryrange/value.go:92:2: invalid case parser.ValueTypeMatrix in switch on promRes.Data.ResultType (mismatched types parser.ValueType and string)
!! command failed: build -o /home/yeya24/go/bin/thanos -ldflags -X github.com/prometheus/common/version.Version=0.14.0 -X github.com/prometheus/common/version.Revision=f3b29ab18ede95baad53cda564b490916efb5407 -X github.com/prometheus/common/version.Branch=add-thanos-query-frontend -X github.com/prometheus/common/version.BuildUser=yeya24@yeya24 -X github.com/prometheus/common/version.BuildDate=20200807-21:34:12  -extldflags '-static' -a -tags netgo github.com/thanos-io/thanos/cmd/thanos: exit status 2
make: *** [Makefile:114: build] Error 1
```

Seems the Prometheus dependency needs to be updated on the Cortex side as well,

cmd/thanos/query-frontend.go
bwplotka: You are right, was wrong (: Cool!

bwplotka:
That might be true. Here comes cyclic deps (:

bwplotka:
Let's propose a PR on their side cc @pracucci

yeya24:
> Let's propose a PR on their side cc @pracucci

I will prepare a pr.

pkg/api/queryfrontend/v1.go
```diff
@@ -0,0 +1,50 @@
...
+	r.Post("/query_range", instr("query_range", handleFunc))
+
+	r.Get("/label/:name/values", instr("/label/:name/values", handleFunc))
+
+	r.Get("/series", instr("series", handleFunc))
+	r.Post("/series", instr("series", handleFunc))
+
+	r.Get("/labels", instr("labels", handleFunc))
+	r.Post("/labels", instr("labels", handleFunc))
+}
```
yeya24, line 50: Is it necessary to add `stores` and `rules` APIs here?

cmd/thanos/query-frontend.go
yeya24: Yes, you are right. It is a general cache config so I named it `query-range.response-cache-max-freshness`.

cmd/thanos/query-frontend.go
yeya24: Done.

yeya24:
I have updated my pr to support the response cache config file. PTAL when you have time. For the Cortex dependency error, I opened a pr https://github.com/cortexproject/cortex/pull/3000 already.

CHANGELOG.md
```diff
@@ -35,6 +35,7 @@ We use *breaking* word for marking changes that are not backward compatible (rel
 - [#2865](https://github.com/thanos-io/thanos/pull/2865) ui: Migrate Thanos Ruler UI to React
 - [#2964](https://github.com/thanos-io/thanos/pull/2964) Query: Add time range parameters to label APIs. Add `start` and `end` fields to Store API `LabelNamesRequest` and `LabelValuesRequest`.
 - [#2996](https://github.com/thanos-io/thanos/pull/2996) Sidecar: Add `reloader_config_apply_errors_total` metric. Add new flags `--reloader.watch-interval`, and `--reloader.retry-interval`.
+- [#2973](https://github.com/thanos-io/thanos/pull/2973) Add Thanos Query Frontend component.
```
bwplotka, line 38: Docs, would be nice, maybe in next PR

cmd/thanos/query-frontend.go
```diff
@@ -0,0 +1,184 @@
...
+	cmd.Flag("query-range.max-retries-per-request", "Maximum number of retries for a single request; beyond this, the downstream error is returned.").
+		Default("5").IntVar(&c.maxRetries)
+
+	cmd.Flag("query-range.max-query-length", "Limit the query time range (end - start time) in the query-frontend, 0 disables it.").
+		Default("0").SetValue(&c.maxQueryLength)
+
+	cmd.Flag("query-range.max-query-parallelism", "Maximum number of queries will be scheduled in parallel by the frontend.").
+		Default("14").IntVar(&c.maxQueryParallelism)
+
+	cmd.Flag("query-range.response-cache-max-freshness", "Most recent allowed cacheable result, to prevent caching very recent results that might still be in flux.").
```
bwplotka, line 62: We need better help flag description - in separate PR is ok

bwplotka: essentially let's describe why this is needed

pkg/api/queryfrontend/v1.go
bwplotka: well depends how we handle this - I think query frontend was doing pass through for all not defined endpoints, but we should test against it. If that's the case then why we define labels, series, values if not cached?

pkg/queryfrontend/cache/inmemory.go
```diff
@@ -0,0 +1,40 @@
+// Copyright (c) The Thanos Authors.
+// Licensed under the Apache License 2.0.
+
+package cache
```
bwplotka, line 4: `inmem`?

pkg/api/queryfrontend/v1.go
bwplotka: Can we check it? (:

pkg/queryfrontend/cache/inmemory.go
bwplotka: ah it's the same package... I think it's fine for now

pkg/queryfrontend/roundtrip.go
```diff
@@ -0,0 +1,125 @@
...
+	return func(next http.RoundTripper) http.RoundTripper {
+		queryRangeTripper := queryrange.NewRoundTripper(next, codec, queryRangeMiddleware...)
+		return frontend.RoundTripFunc(func(r *http.Request) (*http.Response, error) {
+			isQueryRange := strings.HasSuffix(r.URL.Path, "/query_range")
+			op := labelQuery
+			if isQueryRange {
+				op = labelQueryRange
+			}
+
+			queriesCount.WithLabelValues(op).Inc()
```
bwplotka, line 108: this is weird. I think we put this this tripperware for `query`, `query range` but also` /label` `series` `label/.../values`?

pkg/queryfrontend/roundtrip_test.go
```diff
@@ -0,0 +1,222 @@
...
+func (r *fakeRoundTripper) RoundTrip(h *http.Request) (*http.Response, error) {
+	h.URL.Scheme = "http"
+	h.URL.Host = r.host
+	return http.DefaultTransport.RoundTrip(h)
+}
+
+// TestRoundTripRetryMiddleware tests the retry middleware.
+func TestRoundTripRetryMiddleware(t *testing.T) {
+	testRequest := &queryrange.PrometheusRequest{
+		Path:  "/api/v1/query_range",
```
bwplotka, line 71: What if path is totally different? not query range? Can we test it?

pkg/queryfrontend/roundtrip_test.go
```diff
@@ -0,0 +1,222 @@
...
+		name             string
+		splitInterval    time.Duration
+		req              queryrange.Request
+		handlerAndResult func() (*int, http.Handler)
+		expected         int
+	}{
+		{
+			name: "disable split",
+			req: &queryrange.PrometheusRequest{
+				Path:  "/api/v1/query_range",
```
bwplotka, line 166: again, can we test different paths? just to see what to expect

test/e2e/query_frontend_test.go
```diff
@@ -0,0 +1,193 @@
...
+
+		testutil.Ok(t, queryFrontend.WaitSumMetricsWithOptions(
+			e2e.Equals(1),
+			[]string{"thanos_query_frontend_queries_total"},
+			e2e.WithLabelMatchers(labels.MustNewMatcher(labels.MatchEqual, "op", "query_range"))),
+		)
+		testutil.Ok(t, queryFrontend.WaitSumMetrics(e2e.Equals(1), "cortex_cache_fetched_keys"))
+		testutil.Ok(t, queryFrontend.WaitSumMetrics(e2e.Equals(0), "cortex_cache_hits"))
+		testutil.Ok(t, queryFrontend.WaitSumMetrics(e2e.Equals(1), "querier_cache_added_new_total"))
+		testutil.Ok(t, queryFrontend.WaitSumMetrics(e2e.Equals(1), "querier_cache_added_total"))
```
bwplotka, line 120: can we check if backend (querier) was touched? (: using some grpc_client conn metric on querier (: WDYT?

bwplotka:
Also still somehow tests flakes, wonder if it's SWIFT flakiness

yeya24:
> `well depends how we handle this - I think query frontend was doing pass through for all not defined endpoints, but we should test against it. If that's the case then why we define labels, series, values if not cached?`?

Not passing through. Please see https://github.com/cortexproject/cortex/issues/2742. Now only defined endpoints are accessible. That's why the frontend needs to register other endpoints.

Only query range results are cached. The workflow is:
1. Non-query range but defined endpoints -> pass through to downstream URL. So no middlewares will be applied here.
2. query range -> go through all defined middlewares.

> `this is weird. I think we put this this tripperware for query, query range but also /label series label/.../values?`

Thanks, this metric is the same as Cortex. I agree we should separate labels for other endpoints as well.

yeya24:
Seems we got another flaky test https://app.circleci.com/pipelines/github/thanos-io/thanos/2764/workflows/74964958-9ce1-4041-b3b9-4ea659a1764e/jobs/10997

test/e2e/query_frontend_test.go
yeya24: Good idea! Let me update it.

yeya24: What about using `grpc_server_handled_total{method="xxx"}`?

yeya24: I think maybe checking `http_requests_total` on querier is also fine?

bwplotka:
In this case let's do following:

* Use handler in query_range and query since we will soon modify it to use our tripperware
* For everything else let's have (SEPARATE!) passthrough handler. WDYT? (: Also I am surprised it does not work as our UI works just fine on query frontend :thinking: 
TBH I don't feel like we need allowlist all of endpoints. Maybe we can catch query + query range for now and pass through rest? 
Make it more transparent (:

yeya24:
> In this case let's do following:
> 
> * Use handler in query_range and query since we will soon modify it to use our tripperware
> * For everything else let's have (SEPARATE!) passthrough handler. WDYT? (: Also I am surprised it does not work as our UI works just fine on query frontend thinking
>   TBH I don't feel like we need allowlist all of endpoints. Maybe we can catch query + query range for now and pass through rest?
>   Make it more transparent (:

Hello, I agree that we shouldn't limit the endpoints here and it is better to pass through all other routes to downstream querier to make sure everything works fine (like Grafana).

But I just found it not as easy as I thought to deal with the default route with `github.com/prometheus/common/route` package, so I removed the router and added the routing logic to the tripperware. WDYT? This way we don't need to have other routers. This works but TBH I don't know whether this is a good design or not.

``` go
	return func(next http.RoundTripper) http.RoundTripper {
		queryRangeTripper := queryrange.NewRoundTripper(next, codec, queryRangeMiddleware...)
		return frontend.RoundTripFunc(func(r *http.Request) (*http.Response, error) {
			switch r.URL.Path {
			case "/api/v1/query":
				if r.Method == http.MethodGet || r.Method == http.MethodPost {
					queriesCount.WithLabelValues(labelQuery).Inc()
				}
			case "api/v1/query_range":
				if r.Method == http.MethodGet || r.Method == http.MethodPost {
					queriesCount.WithLabelValues(labelQueryRange).Inc()
					return queryRangeTripper.RoundTrip(r)
				}
			default:
			}
			return next.RoundTrip(r)
		})
	}, nil
```

pkg/queryfrontend/roundtrip.go
yeya24: Just refactored this part. Since we only keep `query` and `query_range` endpoints, so we don't need to think about other endpoints anymore.

cmd/thanos/query-frontend.go
yeya24: Make sense, we can also help update the flag description in Cortex as well.

yeya24:
PR is updated and I added more test cases. PTAL

test/e2e/query_frontend_test.go
bwplotka: yup :+1:

bwplotka:
Happy with whatever makes sense more (: Will look today

cmd/thanos/query-frontend.go
```diff
@@ -0,0 +1,187 @@
...
+		)
+
+		injectf := func(f http.HandlerFunc) http.HandlerFunc {
+			hf := http.HandlerFunc(func(w http.ResponseWriter, r *http.Request) {
+				// Cortex frontend middlewares require orgID.
+				f.ServeHTTP(w, r.WithContext(user.InjectOrgID(r.Context(), "fake")))
+			})
+			return hf
+		}
+		srv.Handle("/", injectf(fe.Handler().ServeHTTP))
```
bwplotka, line 170: :+1:

test/e2e/query_frontend_test.go
```diff
@@ -0,0 +1,236 @@
...
+		testutil.Ok(t, q.WaitSumMetricsWithOptions(
+			e2e.Equals(1),
+			[]string{"http_requests_total"},
+			e2e.WithLabelMatchers(labels.MustNewMatcher(labels.MatchEqual, "handler", "query_range"))),
+		)
+	})
+
+	t.Run("same range query, cache hit.", func(t *testing.T) {
+		// Run the same range query again, the result can be retrieved from cache directly.
+		rangeQuery(
```
bwplotka, line 160: can we check that other endpoints are pass through?

yeya24: I already tested `labelNames` and `labelValues` here https://github.com/thanos-io/thanos/blob/master/test/e2e/query_frontend_test.go#L99-L118. Is it not enough?

Issue #2971, opened 2020-08-03
All result timestamps off by 1ms for certain floating-point input timestamps


juliusv:
This Cortex issue is relevant for Thanos as well: https://github.com/cortexproject/cortex/issues/2932

I'm getting the same timestamp behavior for the latest Thanos release v0.14.0, when passing in range query timestamps (`start` / `stop` of e.g. `1595846748.806` being parsed as `1595846748.805`).

bwplotka:
Thanks!

We just need this: https://github.com/prometheus/prometheus/pull/4941

bwplotka:
Quick fix is easy, help wanted

PR #2976, opened 2020-08-04
Query: Better rounding for incoming query timestamps.

johncming:

## Changes

Fixes: #2971

Use same fix like https://github.com/prometheus/prometheus/pull/4941

## Verification

1. Add a test case in `TestParseTime` and run it.
2. Run `make` via ci.


bwplotka:
Thank you for so quick pr! 😻

PR #2979, opened 2020-08-04
REPLICATOR: Add TimePartitionMetaFilter to thanos replicate.

OGKevin:
This will allow time based replication.

Its blocked by #2906 

## Changes

* Added TimePartitionMetaFilter for thanos replication, this means that replicate can now be time based.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -22,6 +22,7 @@ We use *breaking :warning:* to mark changes that are not backward compatible (re
 - [#3312](https://github.com/thanos-io/thanos/pull/3312) s3: add list_objects_version config option for compatibility.
 - [#3356](https://github.com/thanos-io/thanos/pull/3356) Query Frontend: Add a flag to disable step alignment middleware for query range.
 - [#3378](https://github.com/thanos-io/thanos/pull/3378) Ruler: added the ability to send queries via the HTTP method POST. Helps when alerting/recording rules are extra long because it encodes the actual parameters inside of the body instead of the URI. Thanos Ruler now uses POST by default unless `--query.http-method` is set `GET`.
+- [#2979](https://github.com/thanos-io/thanos/pull/2979) Replicator: Add the ability to be albe to replicate blocks within a time frame by passing --min-time and --max-time
 
 ### Fixed
 - [#3257](https://github.com/thanos-io/thanos/pull/3257) Ruler: Prevent Ruler from crashing when using default DNS to lookup hosts that results in "No such hosts" errors.
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -22,7 +22,7 @@ We use *breaking :warning:* to mark changes that are not backward compatible (re
 - [#3312](https://github.com/thanos-io/thanos/pull/3312) s3: add list_objects_version config option for compatibility.
 - [#3356](https://github.com/thanos-io/thanos/pull/3356) Query Frontend: Add a flag to disable step alignment middleware for query range.
 - [#3378](https://github.com/thanos-io/thanos/pull/3378) Ruler: added the ability to send queries via the HTTP method POST. Helps when alerting/recording rules are extra long because it encodes the actual parameters inside of the body instead of the URI. Thanos Ruler now uses POST by default unless `--query.http-method` is set `GET`.
-- [#2979](https://github.com/thanos-io/thanos/pull/2979) Replicator: Add the ability to be albe to replicate blocks within a time frame by passing --min-time and --max-time
+- [#2979](https://github.com/thanos-io/thanos/pull/2979) Replicator: Add the ability to replicate blocks within a time frame by passing --min-time and --max-tim
 
 ### Fixed
 - [#3257](https://github.com/thanos-io/thanos/pull/3257) Ruler: Prevent Ruler from crashing when using default DNS to lookup hosts that results in "No such hosts" errors.
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -22,7 +22,7 @@ We use *breaking :warning:* to mark changes that are not backward compatible (re
 - [#3312](https://github.com/thanos-io/thanos/pull/3312) s3: add list_objects_version config option for compatibility.
 - [#3356](https://github.com/thanos-io/thanos/pull/3356) Query Frontend: Add a flag to disable step alignment middleware for query range.
 - [#3378](https://github.com/thanos-io/thanos/pull/3378) Ruler: added the ability to send queries via the HTTP method POST. Helps when alerting/recording rules are extra long because it encodes the actual parameters inside of the body instead of the URI. Thanos Ruler now uses POST by default unless `--query.http-method` is set `GET`.
-- [#2979](https://github.com/thanos-io/thanos/pull/2979) Replicator: Add the ability to replicate blocks within a time frame by passing --min-time and --max-tim
+- [#2979](https://github.com/thanos-io/thanos/pull/2979) Replicator: Add the ability to replicate blocks within a time frame by passing --min-time and --max-time
 
 ### Fixed
 - [#3257](https://github.com/thanos-io/thanos/pull/3257) Ruler: Prevent Ruler from crashing when using default DNS to lookup hosts that results in "No such hosts" errors.
--- a/cmd/thanos/tools_bucket.go
+++ b/cmd/thanos/tools_bucket.go
@@ -14,8 +14,6 @@ import (
 	"text/template"
 	"time"
 
-	"github.com/thanos-io/thanos/pkg/model"
-
 	"github.com/go-kit/kit/log"
 	"github.com/go-kit/kit/log/level"
 	"github.com/oklog/run"
@@ -38,6 +36,7 @@ import (
 	"github.com/thanos-io/thanos/pkg/extprom"
 	extpromhttp "github.com/thanos-io/thanos/pkg/extprom/http"
 	"github.com/thanos-io/thanos/pkg/logging"
+	"github.com/thanos-io/thanos/pkg/model"
 	"github.com/thanos-io/thanos/pkg/objstore"
 	"github.com/thanos-io/thanos/pkg/objstore/client"
 	"github.com/thanos-io/thanos/pkg/prober"
--- a/cmd/thanos/tools_bucket.go
+++ b/cmd/thanos/tools_bucket.go
@@ -446,6 +446,7 @@ func registerBucketReplicate(app extkingpin.AppClause, objStoreConfig *extflag.P
 		Default("0000-01-01T00:00:00Z"))
 	maxTime := model.TimeOrDuration(cmd.Flag("max-time", "End of time range limit to replicate. Thanos Replicate will replicate only metrics, which happened earlier than this value. Option can be a constant time in RFC3339 format or time duration relative to current time, such as -1d or 2h45m. Valid duration units are ms, s, m, h, d, w, y.").
 		Default("9999-12-31T23:59:59Z"))
+	deleteOldBlocks := cmd.Flag("delete-old-blocks", "Delete blocks that are older then max-time.").Default("false").Bool()
 
 
 	cmd.Setup(func(g *run.Group, logger log.Logger, reg *prometheus.Registry, tracer opentracing.Tracer, _ <-chan struct{}, _ bool) error {
@@ -474,6 +475,7 @@ func registerBucketReplicate(app extkingpin.AppClause, objStoreConfig *extflag.P
 			*singleRun,
 			minTime,
 			maxTime,
+			*deleteOldBlocks,
 		)
 	})
 }
--- a/cmd/thanos/tools_bucket.go
+++ b/cmd/thanos/tools_bucket.go
@@ -7,13 +7,14 @@ import (
 	"context"
 	"encoding/json"
 	"fmt"
-	"github.com/thanos-io/thanos/pkg/model"
 	"os"
 	"sort"
 	"strconv"
 	"strings"
 	"text/template"
 	"time"
+
+	"github.com/thanos-io/thanos/pkg/model"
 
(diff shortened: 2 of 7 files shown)


pkg/block/metadata/deletionmark.go
```diff
@@ -32,7 +32,7 @@ var ErrorDeletionMarkNotFound = errors.New("deletion-mark.json not found")
 // or the deletion-mark.json file is not a valid json file.
 var ErrorUnmarshalDeletionMark = errors.New("unmarshal deletion-mark.json")
 
-// DeletionMark stores block id and when block was marked for deletion.
+// DeletionMark stores block id| and when block was marked for deletion.
```
OGKevin, line 35:
```suggestion
// DeletionMark stores block id and when block was marked for deletion.
```

cmd/thanos/tools_bucket.go
```diff
@@ -438,6 +440,11 @@ func registerBucketReplicate(m map[string]setupFunc, root *kingpin.CmdClause, na
...
 	matcherStrs := cmd.Flag("matcher", "Only blocks whose external labels exactly match this matcher will be replicated.").PlaceHolder("key=\"value\"").Strings()
 	singleRun := cmd.Flag("single-run", "Run replication only one time, then exit.").Default("false").Bool()
+	minTime := model.TimeOrDuration(cmd.Flag("min-time", "Start of time range limit to replicate. Thanos Replicate will replicate only metrics, which happened later than this value. Option can be a constant time in RFC3339 format or time duration relative to current time, such as -1d or 2h45m. Valid duration units are ms, s, m, h, d, w, y.").
+		Default("0000-01-01T00:00:00Z"))
+	maxTime := model.TimeOrDuration(cmd.Flag("max-time", "End of time range limit to replicate. Thanos Replicate will replicate only metrics, which happened earlier than this value. Option can be a constant time in RFC3339 format or time duration relative to current time, such as -1d or 2h45m. Valid duration units are ms, s, m, h, d, w, y.").
+		Default("9999-12-31T23:59:59Z"))
+	deleteOldBlocks := cmd.Flag("delete-old-blocks", "Delete blocks that are older then max-time.").Default("false").Bool()
```
brancz, line 447: I don't think replicate should ever be deleting data

OGKevin: Eh this should be min-time actually and not max time 🤦 . 

But the idea here is that, replicate will only replicate data that is inside the window of --max and --min time, e.g. -10h to -30d, to ensure that the bucket only has the data inside this window, there must be a flag to delete the data that falls out of this window.

bwplotka: Yes, let's use compactor for that.

OGKevin: @bwplotka how if I may ask? Should replicator still mark the blocks for deletion? In our particular case, we were not planning on running compactor on the bucket.

OGKevin: Plus its an optional flag 😔

pkg/block/block.go
```diff
@@ -123,6 +123,20 @@ func cleanUp(logger log.Logger, bkt objstore.Bucket, id ulid.ULID, err error) er
 
 // MarkForDeletion creates a file which stores information about when the block was marked for deletion.
 func MarkForDeletion(ctx context.Context, logger log.Logger, bkt objstore.Bucket, id ulid.ULID, markedForDeletion prometheus.Counter) error {
+	return markForDeletion(ctx, logger, bkt, id, time.Now(), markedForDeletion)
```
bwplotka, line 126: This function looks very shallow to me, maybe we can avoid this and inline, WDYT? (:

pkg/replicate/replicator.go
```diff
@@ -160,8 +164,24 @@ func RunReplicate(
 	}, []string{"result"})
 	replicationRunDuration.WithLabelValues(labelSuccess)
 	replicationRunDuration.WithLabelValues(labelError)
+	blocksCleaned := promauto.With(reg).NewCounter(prometheus.CounterOpts{
```
bwplotka, line 167: Let's remove the cleaning part from this PR as we discussed. It's not needed, part of replicator functionality I believe.

pkg/replicate/replicator.go
```diff
@@ -187,10 +209,16 @@ func RunReplicate(
 		logger := log.With(logger, "replication-run-id", ulid.String())
 		level.Info(logger).Log("msg", "running replication attempt")
 
-		if err := newReplicationScheme(logger, metrics, blockFilter, fetcher, fromBkt, toBkt, reg).execute(ctx); err != nil {
+		if err := newReplicationScheme(logger, metrics, blockFilter, fetcher, fromBkt, toBkt, reg, maxTime, deleteOldBlocks).execute(ctx); err != nil {
 			return errors.Wrap(err, "replication execute")
 		}
 
+		if deleteOldBlocks {
```
bwplotka, line 216: to remove

pkg/block/block.go
OGKevin: I don't understand what you mean with inline. 

I did it this way to make it easier to preserve current behaviour of the function and make it backwards compatible. Instead of changing the signature to add an optional time argument.

cmd/thanos/tools_bucket.go
OGKevin: Just to be clear, this flag is suppose to delete the old blocks from the `toBkt` to maintain the retention set on the min-max flags. 

So your suggestion is to add a functionality to compactor that goes over all the blocks and deletes the ones that are outside of the retention? I thought the whole point of compactor is to compact data. To me it makes more sense to add this to replicator as you want to replicate and have data in a certain time window. 

So to me it makes more sense to have replicator mark blocks for deletion and then let compactor or maybe even a whole new tool dedicated to clean up, delete these blocks. 🤔 And then compactor also marks things for deletion and then this cleanup tool ends up deleting them.

OGKevin: Because in the end, we don't need compactor to compact data on the `toBkt`.

OGKevin: Ok I see that compactor can run without downsampling and actually just delete old data. So I guess compactor can do the job.

OGKevin:
Im waiting on the depended MR to get merged to continue on this one. As the dependent MR will most likely introduce more merging conflicts.

cmd/thanos/tools_bucket.go
OGKevin: After testing this setup with compactor, I think this will not work. Im running into the overlap issue as explained here: https://github.com/thanos-io/thanos/blob/master/docs/operating/troubleshooting.md#overlaps.

And the only explanation I have is because we are no running 2 compactors on the same set of blocks. 

We are replicating from bucket 1 to bucket 2, bucket 1 has compactor running on it originally. We were not planning on running compactor on bucket 2 because its already running on bucket 1. But because of your suggestion to use compactor with downsampling disabled and only use it for retention, it still halts when trying to compact blocks due to overlap.


I still think that

>So to me it makes more sense to have replicator mark blocks for deletion and then let compactor or maybe even a whole new tool dedicated to clean up, delete these blocks. 🤔 And then compactor also marks things for deletion and then this cleanup tool ends up deleting them.

is the way to go. Move the deletion of marked blocks to another tool and let compactor and replicator mark blocks for deletion. WDYT?

bwplotka: Let's talk offline (:

OGKevin:
@bwplotka can the stale label be removed? As this is blocked on #2906

kakkoyun:
@yeya24 FYI this is also blocked by #2906

kakkoyun:
@OGKevin In the meanwhile, could you maybe rebase and also sign DCO?

OGKevin:
@bwplotka I think its ready for 👀!

CHANGELOG.md
```diff
@@ -22,6 +22,7 @@ We use *breaking :warning:* to mark changes that are not backward compatible (re
 - [#3312](https://github.com/thanos-io/thanos/pull/3312) s3: add list_objects_version config option for compatibility.
 - [#3356](https://github.com/thanos-io/thanos/pull/3356) Query Frontend: Add a flag to disable step alignment middleware for query range.
 - [#3378](https://github.com/thanos-io/thanos/pull/3378) Ruler: added the ability to send queries via the HTTP method POST. Helps when alerting/recording rules are extra long because it encodes the actual parameters inside of the body instead of the URI. Thanos Ruler now uses POST by default unless `--query.http-method` is set `GET`.
+- [#2979](https://github.com/thanos-io/thanos/pull/2979) Replicator: Add the ability to be albe to replicate blocks within a time frame by passing --min-time and --max-time
```
bwplotka, line 25:
```suggestion
- [#2979](https://github.com/thanos-io/thanos/pull/2979) Replicator: Add the ability to replicate blocks within a time frame by passing --min-time and --max-time
```

pkg/replicate/replicator.go
```diff
@@ -161,7 +164,15 @@ func RunReplicate(
...
 	replicationRunDuration.WithLabelValues(labelError)
 
-	fetcher, err := thanosblock.NewMetaFetcher(logger, 32, fromBkt, "", reg, nil, nil)
+	fetcher, err := thanosblock.NewMetaFetcher(
+		logger,
+		32,
+		fromBkt,
+		"",
+		reg,
+		[]thanosblock.MetadataFilter{thanosblock.NewTimePartitionMetaFilter(*minTime, *maxTime)},
```
bwplotka, line 173: Can we add this only if minTime and maxTime are passed? (:

pkg/replicate/scheme.go
```diff
@@ -200,8 +197,8 @@ func (rs *replicationScheme) execute(ctx context.Context) error {
 
 // ensureBlockIsReplicated ensures that a block present in the origin bucket is
 // present in the target bucket.
-func (rs *replicationScheme) ensureBlockIsReplicated(ctx context.Context, id ulid.ULID) error {
-	blockID := id.String()
+func (rs *replicationScheme) ensureBlockIsReplicated(ctx context.Context, meta *metadata.Meta) error {
```
bwplotka, line 200: Is this needed? ULID seemed enough? :thinking: 

Ensuring function interface is as small as possible gives usually more readable code (: (e.g you are sure that function does not modify / read other fields than ID)

pkg/replicate/scheme.go
```diff
@@ -136,19 +137,15 @@ func newReplicationMetrics(reg prometheus.Registerer) *replicationMetrics {
 			Name: "thanos_replicate_objects_replicated_total",
 			Help: "Total number of objects replicated.",
 		}),
+		blocksMarkedForDeletion: promauto.With(reg).NewCounter(prometheus.CounterOpts{
```
bwplotka, line 140: Is this needed?

pkg/replicate/scheme.go
```diff
@@ -136,19 +137,15 @@ func newReplicationMetrics(reg prometheus.Registerer) *replicationMetrics {
...
-func newReplicationScheme(
-	logger log.Logger,
-	metrics *replicationMetrics,
-	blockFilter blockFilterFunc,
-	fetcher thanosblock.MetadataFetcher,
-	from objstore.InstrumentedBucketReader,
-	to objstore.Bucket,
-	reg prometheus.Registerer,
-) *replicationScheme {
+func newReplicationScheme(logger log.Logger, metrics *replicationMetrics, blockFilter blockFilterFunc, fetcher thanosblock.MetadataFetcher, from objstore.InstrumentedBucketReader, to objstore.Bucket, reg prometheus.Registerer) *replicationScheme {
```
bwplotka, line 148: Why this change?

bwplotka (commented):
Looks good generally, some suggestions.

pkg/replicate/scheme.go
OGKevin: ah this is no longer needed 🙈

pkg/replicate/scheme.go
OGKevin: ah yea this can be reverted. I needed more data before.

pkg/replicate/replicator.go
OGKevin: Wouldn't the behaviour be the same? As the max and min are almost infinite. What is the reason for only passing it if its set to non default value?

CHANGELOG.md
```diff
@@ -22,7 +22,7 @@ We use *breaking :warning:* to mark changes that are not backward compatible (re
 - [#3312](https://github.com/thanos-io/thanos/pull/3312) s3: add list_objects_version config option for compatibility.
 - [#3356](https://github.com/thanos-io/thanos/pull/3356) Query Frontend: Add a flag to disable step alignment middleware for query range.
 - [#3378](https://github.com/thanos-io/thanos/pull/3378) Ruler: added the ability to send queries via the HTTP method POST. Helps when alerting/recording rules are extra long because it encodes the actual parameters inside of the body instead of the URI. Thanos Ruler now uses POST by default unless `--query.http-method` is set `GET`.
-- [#2979](https://github.com/thanos-io/thanos/pull/2979) Replicator: Add the ability to be albe to replicate blocks within a time frame by passing --min-time and --max-time
+- [#2979](https://github.com/thanos-io/thanos/pull/2979) Replicator: Add the ability to replicate blocks within a time frame by passing --min-time and --max-tim
```
kakkoyun, line 25:
```suggestion
- [#2979](https://github.com/thanos-io/thanos/pull/2979) Replicator: Add the ability to replicate blocks within a time frame by passing --min-time and --max-time
```

OGKevin: 🤦🏾 thx.

pkg/replicate/replicator.go
bwplotka: Just no risk ;p

bwplotka (approved):
Amazing! Looks good to me :+1: 

Thanks!

PR #2980, opened 2020-08-04
ui: Migrate block viewer to React

prmsrswt:

## Changes
- Migrated the Block Viewer to React

## Verification
Added new unit tests.

## Screenshots
image

image


Makefile
```diff
@@ -76,7 +76,7 @@ $(REACT_APP_NODE_MODULES_PATH): $(REACT_APP_PATH)/package.json $(REACT_APP_PATH)
 
 $(REACT_APP_OUTPUT_DIR): $(REACT_APP_NODE_MODULES_PATH) $(REACT_APP_SOURCE_FILES)
 	   @echo ">> building React app"
-	   @./scripts/build-react-app.sh
+	   @scripts/build-react-app.sh
```
GiedriusS, line 79: I like the workaround here to avoid the problem in this script on `master` :smile:

pkg/ui/react-app/src/thanos/pages/blocks/BlockDetails.tsx
```diff
@@ -0,0 +1,70 @@
...
+}
+
+export const BlockDetails: FC<BlockDetailsProps> = ({ block, selectBlock }) => {
+  return (
+    <div className={`${styles.blockDetails} ${block && styles.open}`}>
+      {block && (
+        <>
+          <div className={styles.detailsTop}>
+            <p>{block.ulid}</p>
+            <button className={styles.closeBtn} onClick={(): void => selectBlock(undefined)}>
```
GiedriusS, line 20: The close button is a bit too big, at least on my screen. Sometimes it doesn't even show up because of how big it is. Also, maybe it would be nice to make BlockDetails close automatically whenever the user clicks somewhere on the page?

pkg/ui/react-app/src/thanos/Navbar.tsx
```diff
@@ -40,6 +40,8 @@ const navConfig: { [component: string]: (NavConfig | NavDropDown)[] } = {
     { name: 'Alerts', uri: '/new/alerts' },
     { name: 'Rules', uri: '/new/rules' },
   ],
+  bucket: [{ name: 'Blocks', uri: '/new/blocks' }],
+  compact: [{ name: 'Blocks', uri: '/new/blocks' }],
```
squat, line 44: This is copy-paste right?

pkg/ui/react-app/src/thanos/pages/blocks/TimeRange.tsx
```diff
@@ -0,0 +1,49 @@
...
+  viewMinTime: number;
+  viewMaxTime: number;
+  gridMinTime: number;
+  gridMaxTime: number;
+  onChange: React.Dispatch<React.SetStateAction<[number, number]>>;
+}
+
+const TimeRange: FC<TimeRangeProps> = ({ viewMinTime, viewMaxTime, gridMinTime, gridMaxTime, onChange }) => {
+  const marks = useMemo(() => {
+    const NUM_MARKS = 10;
```
squat, line 19: This should likely live outside of the function as it doesn't depend on the func running

pkg/ui/react-app/src/thanos/pages/blocks/blocks.module.css
```diff
@@ -0,0 +1,124 @@
...
+  box-sizing: border-box;
+}
+
+.blockSpan:hover, .blockSpan:focus {
+  outline: none;
+}
+
+/* block colors */
+.res-0 {
+  background: turquoise;
```
squat, line 106: Are we using these colors already?

pkg/ui/react-app/src/thanos/Navbar.tsx
prmsrswt: Yes. Like for now, bucket web and compactor has same menus.

pkg/ui/react-app/src/thanos/pages/blocks/blocks.module.css
prmsrswt: I just chose some colors that came to my mind.

Makefile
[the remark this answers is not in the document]
prmsrswt: I saw a discussion around it somewhere and did this when I needed to build. Should I leave it as it is?

pkg/ui/react-app/src/thanos/pages/blocks/Blocks.tsx
```diff
@@ -0,0 +1,81 @@
...
+}
+
+export const BlocksContent: FC<{ data: BlockListProps }> = ({ data }) => {
+  const [selectedBlock, selectBlock] = useState<Block>();
+
+  const { blocks, label } = data;
+
+  const blockPools = useMemo(() => sortBlocks(blocks, label), [blocks, label]);
+  const [gridMinTime, gridMaxTime] = useMemo(() => {
+    let gridMinTime = blocks[0].minTime;
```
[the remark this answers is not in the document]
prmsrswt: I am not handling the errors yet. And I encountered it just today when the sync hasn't completed in the backend yet. I'll fix it.

pkg/ui/react-app/src/thanos/pages/blocks/BlockDetails.tsx
prmsrswt: Oh, I'd lower the size of the icon, maybe we can use `&times;` native HTML entity instead of the icon.

> Also, maybe it would be nice to make BlockDetails close automatically whenever the user clicks somewhere on the page?

I think it wouldn't be that straightforward to implement as we'd need to use a click handler on everything except blocks. I'd look into it further.

prmsrswt:
@GiedriusS 

> Different colors for different combos of compaction levels and resolutions

We now have different colors for diff resolutions, but I still need to figure out what we need to do about different compaction levels. One idea is to have a darker shade of color for a higher compaction level and a lighter shade of same color for a lower compaction level.

> Darker shades for overlapping blocks

It slipped my mind, thanks for reminding me. I have some ideas on how to implement this.

> The size of the block in the pop-up on the right

Ah yes. But we need to first have this data in the API, maybe we can tackle this in a separate PR, WDYT?

bwplotka:
Important part that was discussed offline: How overlap with 10 blocks looks like e.g?

prmsrswt:
@bwplotka 
> Important part that was discussed offline: How overlap with 10 blocks looks like e.g?

For now, if some blocks are overlapping, the overlapping part will turn darker in shade compared to the rest of the blocks in the same row. The shade doesn't depend on the number of overlapping blocks though, so if there are blocks overlapping or 10, the effect will be the same.

In the future, we can move the overlapping blocks in a new row. This will make more sense if someday we start supporting vertical compaction.

pkg/ui/react-app/src/thanos/pages/blocks/BlockDetails.tsx
```diff
@@ -0,0 +1,68 @@
...
+  selectBlock: React.Dispatch<React.SetStateAction<Block | undefined>>;
+}
+
+export const BlockDetails: FC<BlockDetailsProps> = ({ block, selectBlock }) => {
+  return (
+    <div className={`${styles.blockDetails} ${block && styles.open}`}>
+      {block && (
+        <>
+          <div className={styles.detailsTop}>
+            <p>{block.ulid}</p>
```
squat, line 17: a `<p>` doesn't seem like the most semantic tag for this; maybe a `span` or a `h#`

squat (approved):
Overall this looks great to me :)))) Let's merge this and iterate more

pkg/ui/react-app/src/thanos/pages/blocks/BlockDetails.tsx
GiedriusS: Seems like a very minor thing but let's change that and merge this? :tada:

prmsrswt: Yes, `span` is much more semantic.

PR #2981, opened 2020-08-05
Bump vendored cortex version

yeya24:

## Changes

This pr bumps Cortex version to https://github.com/cortexproject/cortex/commit/d8edc95e2c91ba947ae81e40881ddac34865560e. 

This is split from pr #2973.


yeya24:
Seems the change of `WaitSumMetrics` makes E2E metrics test flaky.

pracucci:
> WaitSumMetrics was changed at pr https://github.com/cortexproject/cortex/pull/2469/files#diff-e939a4fd75edf18ffb2c32b6bbca3679R523. Now it returns an error for non-existing metrics.

@yeya24 FYI, I've just merged another PR https://github.com/cortexproject/cortex/pull/2522 which introduced functional options to the `WaitSumMetricsWithOptions()`, including an option `WaitMissingMetrics` which you could use to change the behaviour and having the framework to wait when a metric is missing instead of fast failing.

yeya24:
> > WaitSumMetrics was changed at pr https://github.com/cortexproject/cortex/pull/2469/files#diff-e939a4fd75edf18ffb2c32b6bbca3679R523. Now it returns an error for non-existing metrics.
> 
> @yeya24 FYI, I've just merged another PR [cortexproject/cortex#2522](https://github.com/cortexproject/cortex/pull/2522) which introduced functional options to the `WaitSumMetricsWithOptions()`, including an option `WaitMissingMetrics` which you could use to change the behaviour and having the framework to wait when a metric is missing instead of fast failing.

Thanks @pracucci! Ci passed :rocket:

test/e2e/compact_test.go
```diff
@@ -509,9 +509,6 @@ func TestCompactWithStoreGateway(t *testing.T) {
 		testutil.Ok(t, c.WaitSumMetrics(e2e.Equals(3), "thanos_compact_group_compaction_runs_started_total"))
 		testutil.Ok(t, c.WaitSumMetrics(e2e.Equals(2), "thanos_compact_group_compaction_runs_completed_total"))
 
-		testutil.Ok(t, c.WaitSumMetrics(e2e.Equals(0), "thanos_compact_downsample_total"))
```
bwplotka, line 512: Why this was removed. (:

yeya24: This metric and `thanos_compact_downsample_failures_total` are both missing. They are not even initialized to 0 because in this test case, the compactor halts so it doesn't actually run the downsample logic @bwplotka

bwplotka: well that's the key here (: We don't expect so we set expectancy to 0

bwplotka: but whatever, we can think about it later

PR #2987, opened 2020-08-05
memcached: Add more metrics for memcached client

kakkoyun:

Add more metrics for memcached to benchmark the performance.

## Changes

* Add an info metric to expose memcached client configuration
* Record expected errors (especially timeout errors) in failure metric
* Change log level to debug since cache writes are best effort and we have a metric for it
* Add a metric to track the size of the values stored in the cache.

## Verification

* `make test-local`

### DIFF
--- a/pkg/cacheutil/memcached_client.go
+++ b/pkg/cacheutil/memcached_client.go
@@ -31,7 +31,7 @@ const (
 	reasonAsyncBufferFull = "async-buffer-full"
 	reasonMalformedKey    = "malformed-key"
 	reasonTimeout         = "timeout"
-	reasonMotStored       = "not-stored"
+	reasonNotStored       = "not-stored"
 	reasonCacheMiss       = "cache-miss"
 	reasonConflict        = "conflict"
 	reasonOther           = "other"
@@ -160,7 +160,7 @@ type memcachedClient struct {
 	failures   *prometheus.CounterVec
 	skipped    *prometheus.CounterVec
 	duration   *prometheus.HistogramVec
-	valueSize  *prometheus.HistogramVec
+	dataSize   *prometheus.HistogramVec
 }
 
 type memcachedGetMultiResult struct {
@@ -255,7 +255,7 @@ func newMemcachedClient(
 	c.failures.WithLabelValues(opGetMulti, reasonCacheMiss)
 	c.failures.WithLabelValues(opGetMulti, reasonOther)
 	c.failures.WithLabelValues(opSet, reasonTimeout)
-	c.failures.WithLabelValues(opSet, reasonMotStored)
+	c.failures.WithLabelValues(opSet, reasonNotStored)
 	c.failures.WithLabelValues(opSet, reasonCacheMiss)
 	c.failures.WithLabelValues(opSet, reasonConflict)
 	c.failures.WithLabelValues(opSet, reasonOther)
@@ -276,17 +276,17 @@ func newMemcachedClient(
 	c.duration.WithLabelValues(opGetMulti)
 	c.duration.WithLabelValues(opSet)
 
-	c.valueSize = promauto.With(reg).NewHistogramVec(prometheus.HistogramOpts{
-		Name: "thanos_memcached_operation_value_size_bytes",
-		Help: "Tracks the size of the values stored in and fetched from memcached.",
+	c.dataSize = promauto.With(reg).NewHistogramVec(prometheus.HistogramOpts{
+		Name: "thanos_memcached_operation_data_size_bytes",
+		Help: "Tracks the size of the data stored in and fetched from memcached.",
 		Buckets: []float64{
 			32, 256, 512, 1024, 32 * 1024, 256 * 1024, 512 * 1024, 1024 * 1024, 32 * 1024 * 1024, 256 * 1024 * 1024, 512 * 1024 * 1024,
 		},
 	},
 		[]string{"operation"},
 	)
-	c.valueSize.WithLabelValues(opGetMulti)
-	c.valueSize.WithLabelValues(opSet)
+	c.dataSize.WithLabelValues(opGetMulti)
+	c.dataSize.WithLabelValues(opSet)
 
 	// As soon as the client is created it must ensure that memcached server
 	// addresses are resolved, so we're going to trigger an initial addresses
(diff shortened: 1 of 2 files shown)


pkg/cacheutil/memcached_client.go
```diff
@@ -244,6 +276,18 @@ func newMemcachedClient(
 	c.duration.WithLabelValues(opGetMulti)
 	c.duration.WithLabelValues(opSet)
 
+	c.valueSize = promauto.With(reg).NewHistogramVec(prometheus.HistogramOpts{
```
kakkoyun, line 279: Now that I think maybe I should have split this metric. It could be confusing because for set operation it only tracks one item, for get it track multiple. What do you think?

pkg/cacheutil/memcached_client.go
```diff
@@ -244,6 +276,18 @@ func newMemcachedClient(
...
+		Name: "thanos_memcached_operation_value_size_bytes",
```
pracucci, line 280: Maybe `thanos_memcached_operation_data_size_bytes`? Give a `GetMulti()` works with multiple values, renaming `value` into `data` is more correct?

pkg/cacheutil/memcached_client.go
pracucci: I personally think it's fine having the operation as label. It's consistent with other metrics. It's the size of the data of a single operation/call.

pkg/cacheutil/memcached_client.go
```diff
@@ -28,6 +29,12 @@ const (
 	opGetMulti            = "getmulti"
 	reasonMaxItemSize     = "max-item-size"
 	reasonAsyncBufferFull = "async-buffer-full"
+	reasonMalformedKey    = "malformed-key"
+	reasonTimeout         = "timeout"
+	reasonMotStored       = "not-stored"
```
pracucci, line 34:
```suggestion
	reasonNotStored       = "not-stored"
```

pkg/cacheutil/memcached_client.go
```diff
@@ -288,15 +332,34 @@ func (c *memcachedClient) SetAsync(_ context.Context, key string, value []byte,
...
+			)
+
+			var e *memcache.ConnectTimeoutError
+			switch {
+			case errors.As(err, &e):
+				c.failures.WithLabelValues(opSet, reasonTimeout).Inc()
+			case errors.Is(err, memcache.ErrNotStored):
+				c.failures.WithLabelValues(opSet, reasonMotStored).Inc()
+			case errors.Is(err, memcache.ErrCASConflict):
+				c.failures.WithLabelValues(opSet, reasonConflict).Inc()
```
pracucci, line 353: Is this a real case for our client?

pkg/cacheutil/memcached_client.go
```diff
@@ -288,15 +332,34 @@ func (c *memcachedClient) SetAsync(_ context.Context, key string, value []byte,
...
+			case errors.Is(err, memcache.ErrCacheMiss):
+				c.failures.WithLabelValues(opSet, reasonCacheMiss).Inc()
```
pracucci, line 355: Can the "Set" operation fail because of a cache miss? 🤔

pkg/cacheutil/memcached_client.go
```diff
@@ -288,15 +332,34 @@ func (c *memcachedClient) SetAsync(_ context.Context, key string, value []byte,
...
+				"server", serverAddr,
+				"err", err,
+			)
+
+			var e *memcache.ConnectTimeoutError
+			switch {
+			case errors.As(err, &e):
+				c.failures.WithLabelValues(opSet, reasonTimeout).Inc()
+			case errors.Is(err, memcache.ErrNotStored):
+				c.failures.WithLabelValues(opSet, reasonMotStored).Inc()
```
pracucci, line 351: What's this error about?

pkg/cacheutil/memcached_client.go
```diff
@@ -410,8 +473,23 @@ func (c *memcachedClient) getMultiSingle(ctx context.Context, keys []string) (it
 	c.operations.WithLabelValues(opGetMulti).Inc()
 	items, err = c.client.GetMulti(keys)
 	if err != nil {
-		c.failures.WithLabelValues(opGetMulti).Inc()
+		var e *memcache.ConnectTimeoutError
+		switch {
+		case errors.As(err, &e):
+			c.failures.WithLabelValues(opGetMulti, reasonTimeout).Inc()
+		case errors.Is(err, memcache.ErrCacheMiss):
+			c.failures.WithLabelValues(opGetMulti, reasonCacheMiss).Inc()
```
pracucci, line 481: Don't think the `GetMulti` can fail because of this. Can it?

pkg/cacheutil/memcached_client.go
kakkoyun: 🙈

pkg/cacheutil/memcached_client.go
kakkoyun: These are the possible return values from the `memcache` library.

pkg/cacheutil/memcached_client.go
kakkoyun: Apparently from the code path of library it's possible. In any case, I'm gonna double check it.

pracucci: Looks weird to me.

kakkoyun:
@pracucci I have fixed errors as much as I can. I couldn't find exact documentation on possible error/response per command. This is what I could figured out from `memcache` package docs.

metalmatze:
I don't have much context, but looks good!

pkg/cacheutil/memcached_client.go
pracucci: I honestly have never seen this error for the `GetMulti()`. I believe this is a thing of the `Get()` only. Anyway, if you feel strong about it, let's keep it, but could be superfluous.

kakkoyun: I don't have any strong feelings about it. This is me just being too precautious. I have no experience on this and couldn't find documentation, I'll take your word for it and remove it. No need for additional cardinality on for this.

kakkoyun:
@pracucci Thanks a lot for thorough review 🙏

pracucci:
@kakkoyun Thanks to you for being so open to feedback! 🚀

Issue #2986, opened 2020-08-05
Unexpected errors in quickstart.sh


mneverov:
There are several issues in the `quickstart.sh`:

* error when run with minio disabled `thanos: error: unexpected`.
This happens due to empty [OBJSTORECFG](https://github.com/thanos-io/thanos/blob/master/scripts/quickstart.sh#L177) being parsed as a separate command line argument - empty string

* minio is enabled by default. This is confusing because I would expect it to be disabled because of default `0` value. 
Moreover, it is not possible to disable it, because [this](https://github.com/thanos-io/thanos/blob/master/scripts/quickstart.sh#L8) will replace env variable `MINIO_ENABLED=""` with `MINIO_ENABLED=0`, hence minio is always enabled.

* export for [S3_ENDPOINT](https://github.com/thanos-io/thanos/blob/master/scripts/quickstart.sh#L45) will add this variable to the env (no surprises here) if running with minio enabled.
The subsequent run with minio disabled however will run [thanos bucket](https://github.com/thanos-io/thanos/blob/master/scripts/quickstart.sh#L268) against non existing storage bucket

PR #2988, opened 2020-08-05
Fix quickstart

mneverov:

## Changes

Fixes: #2986 

Changes in `quickstart.sh`:
* `MINIO_ENABLED` is disabled by default
* `S3_ENDPOINT` is empty by default and filled in only when minio is enabled, hence [thanos does not run store](https://github.com/thanos-io/thanos/blob/master/scripts/quickstart.sh#L168)
* thanos arguments are parsed correctly (no empty arguments `OBJSTORECFG`, and `STORES`)
* thanos arguments are dumped in console on parse error. **NOTE** please object if you think it might be a security issue - i'll remove it

I didn't add anything to the changelog, since the last ~10 commits to the quickstart.sh are not reflected there.

## Verification

Verified manually by running `PROMETHEUS_EXECUTABLE=~/go/src/github.com/prometheus/prometheus/prometheus THANOS_EXECUTABLE=~/go/bin/thanos MINIO_ENABLED="" ./scripts/quickstart.sh`.

log is attached.
[log.txt](https://github.com/thanos-io/thanos/files/5028594/log.txt)

Note, that there are still errors `level=error name=query-0 ts=2020-08-05T13:14:02.164403579Z caller=logger.go:22 msg="failed to flush Jaeger spans to server: write udp 127.0.0.1:54187->127.0.0.1:6831: write: connection refused"`, that might need investigation

### DIFF
--- a/cmd/thanos/main.go
+++ b/cmd/thanos/main.go
@@ -65,7 +65,7 @@ func main() {
 
 	cmd, err := app.Parse(os.Args[1:])
 	if err != nil {
-		fmt.Fprintln(os.Stderr, errors.Wrapf(err, "Error parsing commandline arguments"))
+		fmt.Fprintln(os.Stderr, errors.Wrapf(err, "Error parsing commandline arguments: %v", os.Args))
 		app.Usage(os.Args[1:])
 		os.Exit(2)
 	}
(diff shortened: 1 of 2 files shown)


GiedriusS:
Maybe @kakkoyun you could help out in reviewing this? :hugs:

PR #2991, opened 2020-08-05
*: Reuse operation name constants for consistency

kakkoyun:

*  [x] The Changelog entry added.

## Verification

* `make test-local`


pkg/store/cache/caching_bucket.go
```diff
@@ -29,12 +29,6 @@ import (
 const (
 	originCache  = "cache"
 	originBucket = "bucket"
-
-	opGet        = "get"
-	opGetRange   = "getrange"
```
kakkoyun, line 34: This was actually mismatching what we have already for bucket.

pracucci: Doh! I think we should mention this change in the CHANGELOG

bwplotka: Any mixin / alert we could change?

kakkoyun:
@bwplotka Mixin seems ok. We don't have any usage. So merging it.

pkg/store/cache/caching_bucket.go
kakkoyun: Nothing, I could find.

Issue #2803, opened 2020-06-25
failing make test-e2e 


kjge:
**Thanos, and Golang version used:**
thanos master branch last commit 02ab6b84908e5458699655ab064a78afc0f9da9b
go version go1.14.4 darwin/amd64
mac 10.15.5
docker version 19.03.8

**What happened:**

running ```make test-e2e```  gives following error

```
unexpected error: the service querier-1 is not ready; err: Get "http://localhost:32808/-/ready": dial tcp [::1]:32808: connect: connection refused

unexpected error: failed to get mapping for port as container e2e_test_receive_replication_with_outage-receive-1 exited: exit status 1: docker container receive-1 failed to start: exit status 1

unexpected error: failed to get mapping for port as container e2e_test_query-receive-1 exited: exit status 1: docker container receive-1 failed to start: exit status 1

unexpected error: failed to get mapping for port as container e2e_test_receive_hashring-receive-1 exited: exit status 1: docker container receive-1 failed to start: exit status 1


```

even though  ```make docker``` is successful , can someone point me a direction?

bwplotka:
I would look on docker container logs before the error happens. Anything interesting being printeD? (:

kjge:
I checked the docker log before but didn't find anything that stands out

```
level=info ts=2020-07-07T02:51:53.619Z caller=main.go:231 msg="Starting Alertmanager" version="(version=0.20.0, branch=HEAD, revision=f74be0400a6243d10bb53812d6fa408ad71ff32d)"
level=info ts=2020-07-07T02:51:53.620Z caller=main.go:232 build_context="(go=go1.13.5, user=root@00c3106655f8, date=20191211-14:13:14)"
level=info ts=2020-07-07T02:51:53.624Z caller=cluster.go:161 component=cluster msg="setting advertise address explicitly" addr=172.26.0.2 port=9094
level=info ts=2020-07-07T02:51:53.642Z caller=cluster.go:623 component=cluster msg="Waiting for gossip to settle..." interval=2s
level=info ts=2020-07-07T02:51:53.705Z caller=coordinator.go:119 component=configuration msg="Loading configuration file" file=/shared/data/am/1/config.yaml
level=info ts=2020-07-07T02:51:53.714Z caller=coordinator.go:131 component=configuration msg="Completed loading of configuration file" file=/shared/data/am/1/config.yaml
[...]

PR #3000, opened 2020-08-07
change: change test-e2e target to depend on docker-multi-stage

jtlisi:

## Changes

~~Update the Makefile `test-e2e` target to depend on the `docker-multi-stage` instead of `docker` target. This ensures people running on Mac OS X will be able to run the integration tests easily.~~

I updated the `make docker` command to run as normal if the OS is linux and the Arch is x86_64. Otherwise `make docker` will default to `make docker-multi-stage`. 

## Verification

Ran `make test-e2e`

**Note**: This does not change the fact that images built using `make docker` will not work on Mac OS X. Also if other Mac users are not experiencing this issue, I would be interested to know if there is something misconfigured with my local environment causing this to fail.

Fixes: #2803

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -210,7 +210,7 @@ test-local:
 
 .PHONY: test-e2e
 test-e2e: ## Runs all Thanos e2e docker-based e2e tests from test/e2e. Required access to docker daemon.
-test-e2e: docker
+test-e2e: docker-multi-stage
 	@echo ">> cleaning docker environment."
 	@docker system prune -f --volumes
 	@echo ">> cleaning e2e test garbage."
--- a/Makefile
+++ b/Makefile
@@ -124,14 +124,19 @@ deps: ## Ensures fresh go.mod and go.sum.
 	@go mod tidy
 	@go mod verify
 
+
 .PHONY: docker
 docker: ## Builds 'thanos' docker with no tag.
+ifeq ($(OS)_$(ARCH), linux_x86_64)
 docker: build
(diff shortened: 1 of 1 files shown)


Makefile
```diff
@@ -210,7 +210,7 @@ test-local:
 
 .PHONY: test-e2e
 test-e2e: ## Runs all Thanos e2e docker-based e2e tests from test/e2e. Required access to docker daemon.
-test-e2e: docker
```
bwplotka, line 213: I think it makes sense. I also now remember why `make docker`  does not work for you - you build for Mac and copy binary to linux based image on docker, so that's why.

Let's have this but also ,can you comment on `make docker` that it work on Linux only, -or even better: Fail if you are on different ARCH/ OS? ((:

jtlisi: Yes, haha. That makes sense.

**Proposal**: Another option I was considering, adjusting the `make docker` target to default to `make docker-multi-stage` on non linux x86_64 machine. Something along the lines of:

```make
.PHONY: docker
docker: ## Builds 'thanos' docker with no tag.
ifeq ($(OS)_$(ARCH), linux_x86_64)
docker: build
	@echo ">> copying Thanos from $(PREFIX) to ./thanos_tmp_for_docker"
	@cp $(PREFIX)/thanos ./thanos_tmp_for_docker
	@echo ">> building docker image 'thanos'"
	@docker build -t "thanos" .
	@rm ./thanos_tmp_for_docker
else
docker: docker-multi-stage
endif
```

PR #3003, opened 2020-08-08
promclient: Support setting http headers in query range requests

### LINKED ISSUE #2997: Support customized headers in promclient HTTP requests
Currently in the `promclient` package, it is not supported to add HTTP headers when sending Get requests https://github.com/thanos-io/thanos/blob/master/pkg/promclient/promclient.go#L100.

Cortex frontend supports a special HTTP header `Cache-Control` to control whether to cache results or not https://github.com/cortexproject/cortex/blob/master/pkg/querier/queryrange/results_cache.go#L203. So it would be nice to support specifying the headers in `promclient` when sending query range requests, then we can test the cache behavior in the E2E test.

yeya24:

## Changes

Fix #2997 

Support setting HTTP headers in promclient get2xx method.

### DIFF
--- a/pkg/promclient/promclient.go
+++ b/pkg/promclient/promclient.go
@@ -97,13 +97,16 @@ func NewWithTracingClient(logger log.Logger, userAgent string) *Client {
 	)
 }
 
-func (c *Client) get2xx(ctx context.Context, u *url.URL) (_ []byte, _ int, err error) {
+func (c *Client) get2xx(ctx context.Context, u *url.URL, additionalHeaders ...[2]string) (_ []byte, _ int, err error) {
 	req, err := http.NewRequest(http.MethodGet, u.String(), nil)
 	if err != nil {
 		return nil, 0, errors.Wrap(err, "create GET request")
 	}
 	if c.userAgent != "" {
 		req.Header.Set("User-Agent", c.userAgent)
+	}
+	for _, headers := range additionalHeaders {
+		req.Header.Set(headers[0], headers[1])
 	}
 
 	resp, err := c.Do(req.WithContext(ctx))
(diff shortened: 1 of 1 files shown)


pkg/promclient/promclient.go
```diff
@@ -97,14 +97,17 @@ func NewWithTracingClient(logger log.Logger, userAgent string) *Client {
 	)
 }
 
-func (c *Client) get2xx(ctx context.Context, u *url.URL) (_ []byte, _ int, err error) {
+func (c *Client) get2xx(ctx context.Context, u *url.URL, additionalHeaders map[string]string) (_ []byte, _ int, err error) {
```
bwplotka, line 100: Wonder if `headers ...[2]string` wouldn't do (:  WDYT?

yeya24: Umm, I don't have a strong preference here. But seems using [2]string is better than a map because it uses less memory.

yeya24: Let me update the pr.

pkg/promclient/promclient.go
```diff
@@ -97,14 +97,17 @@ func NewWithTracingClient(logger log.Logger, userAgent string) *Client {
 	)
 }
 
-func (c *Client) get2xx(ctx context.Context, u *url.URL) (_ []byte, _ int, err error) {
+func (c *Client) get2xx(ctx context.Context, u *url.URL, additionalHeaders [][2]string) (_ []byte, _ int, err error) {
```
bwplotka, line 100:
```suggestion
func (c *Client) get2xx(ctx context.Context, u *url.URL, headers ...[2]string) (_ []byte, _ int, err error) {
```

That's kind of a point (:  to make those optional

yeya24: Oh, good point. Thanks! Let me update it.

yeya24:
Close this pr. Please see https://github.com/thanos-io/thanos/issues/2997#issuecomment-674538937

PR #3006, opened 2020-08-08
Tools: Add CLI to create tombstones in obj store

Harshitha1234:
Signed-off-by: Harshitha1234 <harshithachowdary.t17@iiits.in>

## Changes
- Add `thanos tools delete` command tool to create tombstones in object storage

## TODO
* [x] Refactor tombstone logic into its own package.
* [x] Write tests.

### DIFF
--- a/cmd/thanos/tools_delete.go
+++ b/cmd/thanos/tools_delete.go
@@ -4,32 +4,18 @@
 package main
 
 import (
-	"context"
-	"encoding/json"
-	"io/ioutil"
-	"os"
-	"strconv"
-	"time"
-
-	"github.com/cespare/xxhash"
 	"github.com/go-kit/kit/log"
 	"github.com/oklog/run"
 	opentracing "github.com/opentracing/opentracing-go"
 	"github.com/prometheus/client_golang/prometheus"
 	"github.com/prometheus/prometheus/promql/parser"
-	"github.com/thanos-io/thanos/pkg/model"
-	"github.com/thanos-io/thanos/pkg/runutil"
-
-	"github.com/thanos-io/thanos/pkg/objstore"
-	"github.com/thanos-io/thanos/pkg/objstore/client"
 	kingpin "gopkg.in/alecthomas/kingpin.v2"
-)
 
-type Tombstone struct {
-	Matcher string `json:"matcher"`
-	MinTime int64  `json:"minTime"`
-	MaxTime int64  `json:"maxTime"`
-}
+	"github.com/thanos-io/thanos/pkg/model"
+	"github.com/thanos-io/thanos/pkg/objstore/client"
+	"github.com/thanos-io/thanos/pkg/runutil"
+	"github.com/thanos-io/thanos/pkg/tombstone"
+)
 
 func registerDelete(m map[string]setupFunc, app *kingpin.CmdClause, pre string) {
 	cmd := app.Command("delete", "Delete series command")
@@ -64,26 +50,9 @@ func registerDelete(m map[string]setupFunc, app *kingpin.CmdClause, pre string) 
 			return err
 		}
 
-		ts := Tombstone{Matcher: *matcher, MinTime: minTime.PrometheusTimestamp(), MaxTime: maxTime.PrometheusTimestamp()}
-		b, err := json.Marshal(ts)
-		if err != nil {
-			return err
-		}
+		ts := tombstone.NewTombstone(*matcher, minTime.PrometheusTimestamp(), maxTime.PrometheusTimestamp())
 
-		tmpDir := os.TempDir()
-
-		tsPath := tmpDir + "/tombstone.json"
-		err = ioutil.WriteFile(tsPath, b, 0644)
-		if err != nil {
-			return err
-		}
-
-		ctx, cancel := context.WithTimeout(context.Background(), 5*time.Minute)
-		defer cancel()
-
-		hash := xxhash.Sum64(b)
-
-		err = objstore.UploadFile(ctx, logger, bkt, tsPath, "tombstones/"+strconv.FormatUint(hash, 10)+".json")
+		err = tombstone.UploadTombstone(ts, bkt, logger)
 		if err != nil {
 			return err
 		}
--- a/cmd/thanos/tools_delete.go
+++ b/cmd/thanos/tools_delete.go
@@ -18,17 +18,21 @@ import (
 )
 
 func registerDelete(m map[string]setupFunc, app *kingpin.CmdClause, pre string) {
-	cmd := app.Command("delete", "Delete series command")
+	cmd := app.Command("delete", "Delete series command for the object storage. NOTE: Currently it only performs Store API masking in the object storage at chunk level with respect to the tombstones created by the user (Doesn't actually delete the data in objstore).")
 
-	matcher := cmd.Flag("matcher", "The string representing label matchers").Default("").String()
+	matchers := cmd.Flag("matchers", "The string representing label matchers").Default("").String()
 
 	objStoreConfig := regCommonObjStoreFlags(cmd, "", true)
 
-	minTime := model.TimeOrDuration(cmd.Flag("min-time", "Start of time range limit to delete. Option can be a constant time in RFC3339 format or time duration relative to current time, such as -1d or 2h45m. Valid duration units are ms, s, m, h, d, w, y.").
+	minTime := model.TimeOrDuration(cmd.Flag("min-time", "Start of time range limit to delete. Option can be a constant time in RFC3339 format or time duration relative to current time, such as -1d (Calculates the actual timestamp at the tombstone creation time) or 2h45m. Valid duration units are ms, s, m, h, d, w, y.").
 		Default("0000-01-01T00:00:00Z"))
 
-	maxTime := model.TimeOrDuration(cmd.Flag("max-time", "End of time range limit to delete. Option can be a constant time in RFC3339 format or time duration relative to current time, such as -1d or 2h45m. Valid duration units are ms, s, m, h, d, w, y.").
+	maxTime := model.TimeOrDuration(cmd.Flag("max-time", "End of time range limit to delete. Option can be a constant time in RFC3339 format or time duration relative to current time, such as -1d (Calculates the actual timestamp at the tombstone creation time) or 2h45m. Valid duration units are ms, s, m, h, d, w, y.").
 		Default("9999-12-31T23:59:59Z"))
+
+	author := cmd.Flag("author", "Author of the deletion request").Default("not specified").String()
+
+	reason := cmd.Flag("reason", "Reason to perform the deletion request").Default("not specified").String()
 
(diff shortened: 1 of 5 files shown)


cmd/thanos/tools_delete.go
```diff
@@ -0,0 +1,93 @@
...
+	"github.com/prometheus/prometheus/promql/parser"
+	"github.com/thanos-io/thanos/pkg/model"
+	"github.com/thanos-io/thanos/pkg/runutil"
+
+	"github.com/thanos-io/thanos/pkg/objstore"
+	"github.com/thanos-io/thanos/pkg/objstore/client"
+	kingpin "gopkg.in/alecthomas/kingpin.v2"
+)
+
+type Tombstone struct {
```
bwplotka, line 28: Probably we need Thanos `tombstone` package (:

cmd/thanos/tools_delete.go
```diff
@@ -0,0 +1,93 @@
...
+
+	objStoreConfig := regCommonObjStoreFlags(cmd, "", true)
+
+	minTime := model.TimeOrDuration(cmd.Flag("min-time", "Start of time range limit to delete. Option can be a constant time in RFC3339 format or time duration relative to current time, such as -1d or 2h45m. Valid duration units are ms, s, m, h, d, w, y.").
+		Default("0000-01-01T00:00:00Z"))
+
+	maxTime := model.TimeOrDuration(cmd.Flag("max-time", "End of time range limit to delete. Option can be a constant time in RFC3339 format or time duration relative to current time, such as -1d or 2h45m. Valid duration units are ms, s, m, h, d, w, y.").
+		Default("9999-12-31T23:59:59Z"))
+
+	m[pre+" delete"] = func(g *run.Group, logger log.Logger, reg *prometheus.Registry, _ opentracing.Tracer, _ <-chan struct{}, _ bool) error {
```
bwplotka, line 47: This is nice, it will work, but I would put most of this function into separate package function so we can test it properly with unit tests. WDYT?

you can use `inmen.New` as dummy object storage (:

Harshitha1234: Yeah sure!  :)

cmd/thanos/tools_delete.go
Harshitha1234: Yes

Harshitha1234:
@bwplotka @metalmatze :)

cmd/thanos/tools_delete.go
```diff
@@ -0,0 +1,62 @@
...
+		g.Add(func() error { return nil }, func(error) {})
+
+		_, err = parser.ParseMetricSelector(*matcher)
+		if err != nil {
+			return err
+		}
+
+		ts := tombstone.NewTombstone(*matcher, minTime.PrometheusTimestamp(), maxTime.PrometheusTimestamp())
+
+		err = tombstone.UploadTombstone(ts, bkt, logger)
```
bwplotka, line 55: what about... `return tombstone.Upload....` (:

cmd/thanos/tools_delete.go
```diff
@@ -0,0 +1,62 @@
...
+
+		// Dummy actor to immediately kill the group after the run function returns.
+		g.Add(func() error { return nil }, func(error) {})
+
+		_, err = parser.ParseMetricSelector(*matcher)
+		if err != nil {
+			return err
+		}
+
+		ts := tombstone.NewTombstone(*matcher, minTime.PrometheusTimestamp(), maxTime.PrometheusTimestamp())
```
bwplotka, line 53: I think single tombstones can have more than one matcher? :thinking:

cmd/thanos/tools_delete.go
```diff
@@ -0,0 +1,62 @@
...
+	kingpin "gopkg.in/alecthomas/kingpin.v2"
+
+	"github.com/thanos-io/thanos/pkg/model"
+	"github.com/thanos-io/thanos/pkg/objstore/client"
+	"github.com/thanos-io/thanos/pkg/runutil"
+	"github.com/thanos-io/thanos/pkg/tombstone"
+)
+
+func registerDelete(m map[string]setupFunc, app *kingpin.CmdClause, pre string) {
+	cmd := app.Command("delete", "Delete series command")
```
bwplotka, line 21: worth to mention
* from where? 
* is this deleted immediately?
* where are the detailed docs?

pkg/tombstone/tombstone.go
```diff
@@ -0,0 +1,74 @@
...
+	"github.com/cespare/xxhash"
+	"github.com/go-kit/kit/log"
+	"github.com/prometheus/prometheus/pkg/timestamp"
+
+	"github.com/thanos-io/thanos/pkg/objstore"
+)
+
+const (
+	// TombstoneDir is the name of directory to upload tombstones.
+	TombstoneDir = "tombstones"
```
bwplotka, line 24: maybe... `thanos/tombstones?` :thinking:

pkg/tombstone/tombstone.go
```diff
@@ -0,0 +1,74 @@
...
+)
+
+// Tombstone represents a tombstone.
+type Tombstone struct {
+	Matcher      string `json:"matcher"`
```
bwplotka, line 29: I think we need slice here

pkg/tombstone/tombstone.go
```diff
@@ -0,0 +1,74 @@
...
+func UploadTombstone(tombstone Tombstone, bkt objstore.Bucket, logger log.Logger) error {
+	b, err := json.Marshal(tombstone)
+	if err != nil {
+		return err
+	}
+
+	tmpDir := os.TempDir()
+
+	tsPath := tmpDir + "/tombstone.json"
+	err = ioutil.WriteFile(tsPath, b, 0644)
```
bwplotka, line 61:
```
if err := .... ; err != nil {
```
pattern available (:

pkg/tombstone/tombstone.go
```diff
@@ -0,0 +1,74 @@
...
+	if err != nil {
+		return err
+	}
+
+	ctx, cancel := context.WithTimeout(context.Background(), 5*time.Minute)
+	defer cancel()
+
+	err = objstore.UploadFile(ctx, logger, bkt, tsPath, path.Join(TombstoneDir, GenName(tombstone)))
```
bwplotka, line 69: just return?

pkg/tombstone/tombstone.go
```diff
@@ -0,0 +1,74 @@
...
+	CreationTime int64  `json:"creationTime"`
+}
+
+// NewTombstone returns a new instance of Tombstone.
+func NewTombstone(matcher string, minTime, maxTime int64) Tombstone {
+	return Tombstone{
+		Matcher:      matcher,
+		MinTime:      minTime,
+		MaxTime:      maxTime,
+		CreationTime: timestamp.FromTime(time.Now()),
```
bwplotka, line 41: Similar to Alertmanager Silencers wonder if we can allow specifying human readable reason why like `reason` field
and `author` field where people can optionally specify author of deletion.. :thinking: Maybe too much (: 

cc @metalmatze

pkg/tombstone/tombstone.go
```diff
@@ -0,0 +1,74 @@
...
+	}
+}
+
+// GenName generates file name based on Matcher, MinTime and MaxTime of a tombstone.
+func GenName(ts Tombstone) string {
+	hash := xxhash.Sum64([]byte(ts.Matcher + string(ts.MinTime) + string(ts.MaxTime)))
```
bwplotka, line 47: interesting (: What's the purpose of this name? Was this logic part of the design?  :thinking: 

The only problem is that it will cause weird errors when someone will try to delete same things. Maybe that's ok - we should at least note it in design or comment here

pkg/tombstone/tombstone.go
```diff
@@ -0,0 +1,74 @@
...
+
+	tmpDir := os.TempDir()
+
+	tsPath := tmpDir + "/tombstone.json"
+	err = ioutil.WriteFile(tsPath, b, 0644)
+	if err != nil {
+		return err
+	}
+
+	ctx, cancel := context.WithTimeout(context.Background(), 5*time.Minute)
```
bwplotka, line 66: :+1:

pkg/tombstone/tombstone_test.go
```diff
@@ -0,0 +1,23 @@
...
+
+	"github.com/thanos-io/thanos/pkg/objstore"
+	"github.com/thanos-io/thanos/pkg/testutil"
+)
+
+func TestUploadTombstone(t *testing.T) {
+	bkt := objstore.WithNoopInstr(objstore.NewInMemBucket())
+
+	{
+		SampleTombstone := NewTombstone("up{a=\"b\"}", 00, 9999999)
```
bwplotka, line 19: does it has to be upper case?

bwplotka (approved):
Amazing, some comments, but overall LGTM (:

cmd/thanos/tools_delete.go
```diff
@@ -0,0 +1,62 @@
...
+
+	matcher := cmd.Flag("matcher", "The string representing label matchers").Default("").String()
+
+	objStoreConfig := regCommonObjStoreFlags(cmd, "", true)
+
+	minTime := model.TimeOrDuration(cmd.Flag("min-time", "Start of time range limit to delete. Option can be a constant time in RFC3339 format or time duration relative to current time, such as -1d or 2h45m. Valid duration units are ms, s, m, h, d, w, y.").
```
bwplotka, line 27: Does this mean the deletion can be from -10d to -1d? That does not make much sense, right? (: I think we should be explicit in help that -1d will calculate the actual timestamp now

pkg/tombstone/tombstone_test.go
Harshitha1234: Nope :p

cmd/thanos/tools_delete.go
Harshitha1234: Updated the flag help :+1:

pkg/tombstone/tombstone.go
Harshitha1234: If a same request is repeated in the future, this logic helps overwrite the previous tombstone file and avoids duplicate tombstones. Yeah sure, will add this to the design proposal. :slightly_smiling_face:

pkg/tombstone/tombstone.go
Harshitha1234: Added the author and reason flags :+1:

pkg/tombstone/tombstone.go
Harshitha1234: We will be slicing the `matchers` during store API masking and while tombstone creation, we store it as a string. :slightly_smiling_face:

pkg/tombstone/tombstone.go
Harshitha1234: Updated :+1:

cmd/thanos/tools_delete.go
Harshitha1234: Yes and we check the validity of the user input for the `matchers` using 
`_, err = parser.ParseMetricSelector(*matchers)` and store them in the form of a string.

PR #3007, opened 2020-08-10
Made version 0.1 and 0.2 to render docs on website

thisisobate:

## Changes

Made version 0.1 and 0.2 to render docs on website by changing the file name and adding frontmatter to files that doesn't have them.

## Verification
Works on local machine


bwplotka:
Talked offine and decided for better, non dynamic approach (:

Issue #3001, opened 2020-08-07
Same url (with max time in future) got diffenent data from Prometheus  and  thanos query 


husoule:
**What happened**:
I want a query which endtime is a future time. If I play with Prometheus directly, I see the latest result being less then now + lookback-delta. This is as expected。 But via thanos query, it got the latest result data of future 5m.

via thanos query: 
![image](https://user-images.githubusercontent.com/5336915/89661464-4ee64880-d905-11ea-9605-78ab06332eb7.png)

Prometheus directly:
![image](https://user-images.githubusercontent.com/5336915/89661963-f5cae480-d905-11ea-9dc9-8e2b9f3b8fab.png)

**What you expected to happen**:
get the same data

**Anything else we need to know**:

**Environment**:
- OS (e.g. from /etc/os-release): centos7
- Kernel (e.g. `uname -a`): Linux VM_120_12_centos 3.10.107-1 
- Others: 
prometheus version: v2.18.1
thanos query version:  thanos:v0.12.0
thanos query args:

        - query
        - --log.level=debug
        - --query.auto-downsampling
        - --grpc-address=0.0.0.0:10901
        - --http-address=0.0.0.0:9090
        - --query.partial-response
        - --query.replica-label=prometheus_replica
        - --query.replica-label=rule_replica

bwplotka:
Ack, might be the deduplication (: But kind of when you ask for future - this is what deduplication interpolates ;p

bwplotka:
How critical this is? Why it would cause issues?

husoule:
> How critical this is? Why it would cause issues?

I have a stress test service platform for many users . Use prometheus+thanos+grafana to collect and display graphs of stress test data. The stress test data is more accurate in terms of time, but the user actually saw the data in the chart that has not yet occurred, which is unacceptable. So I adjusted prometheus' --query.lookback-delta=5s to reduce future data. And prometheus is as expected, but there will still be data for the next 5m after thanos. So in fact, in my scenario, future data is not needed. I don't know how to not get the data of future time via thaos query.

PR #3010, opened 2020-08-10
querier: bugfix - data gaps when switching iterators

krasi-georgiev:
closes: https://github.com/thanos-io/thanos/issues/3001

Promql has a default 5m look back delta so if the penalty causes the
duration between the samples to be greater than 5m this will show as
gaps in the visualisation and when using the count function returns a
lower number then expected.

For example:
time is in mins

series 1
```
itA  t1 0.1  t3 0.3
itB  t1 0.1  t3 0.3 t4 0.4  t9 0.9
```

`it.penB = 2 * (ta - it.lastT), 2 * (3 - 1) = 4 `

```
merged t1 0.1  t3 0.3  t9 0.9
```

t4 is missing because the it.penB is 4 so it skips it.

When the time between  `t3 0.3`  and  `t9 0.9` is more then 5min(the default look
back) so when visualizing this it shows as gaps in the graph.

~When the penalty is resolution based the seek for itB always advances NO
more then the requested time so doesn’t skip samples when switching
between iterators.~ 

UPDATE: step based penalty wouldn't work when using promql functions. Adjusting the look back solves the problem so PR is updated to add it as a configurable flag.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -16,6 +16,7 @@ We use *breaking :warning:* word for marking changes that are not backward compa
 * [#3095](https://github.com/thanos-io/thanos/pull/3095) Rule: update manager when all rule files are removed.
 * [#3098](https://github.com/thanos-io/thanos/pull/3098) ui: Fix Block Viewer for Compactor and Store
 * [#3105](https://github.com/thanos-io/thanos/pull/3105) Query: Fix overwriting maxSourceResolution when auto downsampling is enabled.
+* [#3010](https://github.com/thanos-io/thanos/pull/3010) Querier: Added a flag to override the default look back delta in promql. The flag should be set to at least 2 times the slowest scrape interval or left unset to use the Prometheus defaults of 5min.
 
 ## [v0.15.0-rc.0](https://github.com/thanos-io/thanos/releases/tag/v0.15.0-rc.0) - 2020.08.26
 
--- a/cmd/thanos/query.go
+++ b/cmd/thanos/query.go
@@ -70,6 +70,8 @@ func registerQuery(m map[string]setupFunc, app *kingpin.Application) {
 	maxConcurrentQueries := cmd.Flag("query.max-concurrent", "Maximum number of queries processed concurrently by query node.").
 		Default("20").Int()
 
+	lookbackDelta := cmd.Flag("query.lookback-delta", "The maximum lookback duration for retrieving metrics during expression evaluations. PromQL always evaluates the query for the certain timestamp (query range timestamps are deduced by step). Since scrape intervals might be different, PromQL looks back for given amount of time to get latest sample. If it exceeds the maximum lookback delta it assumes series is stale and returns none (a gap). This is why lookback delta should be set to at least 2 times of the slowest scrape interval. If unset it will use the promql default of 5m").Duration()
+
 	maxConcurrentSelects := cmd.Flag("query.max-concurrent-select", "Maximum number of select requests made concurrently per a query.").
 		Default("4").Int()
 
@@ -175,6 +177,7 @@ func registerQuery(m map[string]setupFunc, app *kingpin.Application) {
 			*maxConcurrentQueries,
 			*maxConcurrentSelects,
 			time.Duration(*queryTimeout),
+			*lookbackDelta,
 			time.Duration(*defaultEvaluationInterval),
 			time.Duration(*storeResponseTimeout),
 			*queryReplicaLabels,
@@ -222,6 +225,7 @@ func runQuery(
 	maxConcurrentQueries int,
 	maxConcurrentSelects int,
 	queryTimeout time.Duration,
+	lookbackDelta time.Duration,
 	defaultEvaluationInterval time.Duration,
 	storeResponseTimeout time.Duration,
 	queryReplicaLabels []string,
@@ -312,6 +316,7 @@ func runQuery(
 				NoStepSubqueryIntervalFn: func(rangeMillis int64) int64 {
 					return defaultEvaluationInterval.Milliseconds()
 				},
+				LookbackDelta: lookbackDelta,
 			},
 		)
 	)
--- a/docs/components/query.md
+++ b/docs/components/query.md
@@ -367,6 +367,19 @@ Flags:
       --query.timeout=2m         Maximum time to process query by query node.
       --query.max-concurrent=20  Maximum number of queries processed
                                  concurrently by query node.
+      --query.lookback-delta=QUERY.LOOKBACK-DELTA
+                                 The maximum lookback duration for retrieving
+                                 metrics during expression evaluations. PromQL
+                                 always evaluates the query for the certain
+                                 timestamp (query range timestamps are deduced
+                                 by step). Since scrape intervals might be
+                                 different, PromQL looks back for given amount
+                                 of time to get latest sample. If it exceeds the
+                                 maximum lookback delta it assumes series is
+                                 stale and returns none (a gap). This is why
+                                 lookback delta should be set to at least 2
+                                 times of the slowest scrape interval. If unset
+                                 it will use the promql default of 5m
       --query.max-concurrent-select=4
                                  Maximum number of select requests made
                                  concurrently per a query.
--- a/pkg/query/iter.go
+++ b/pkg/query/iter.go
@@ -302,7 +302,7 @@ type chunkSeriesIterator struct {
 func newChunkSeriesIterator(cs []chunkenc.Iterator) chunkenc.Iterator {
 	if len(cs) == 0 {
 		// This should not happen. StoreAPI implementations should not send empty results.
-		return errSeriesIterator{}
+		return errSeriesIterator{err: errors.Errorf("store returned an empty result")}
 	}
 	return &chunkSeriesIterator{chunks: cs}
 }
@@ -503,7 +503,8 @@ func (it noopAdjustableSeriesIterator) adjustAtValue(float64) {}
 // Replica 1 counter scrapes: 20    30    40    Nan      -     0     5
 // Replica 2 counter scrapes:    25    35    45     Nan     -     2
 //
-// Now for downsampling purposes we are accounting the resets so our replicas before going to dedup iterator looks like this:
+// Now for downsampling purposes we are accounting the resets(rewriting the samples value)
+// so our replicas before going to dedup iterator looks like this:
 //
 // Replica 1 counter total: 20    30    40   -      -     40     45
 // Replica 2 counter total:    25    35    45    -     -     47
@@ -648,7 +649,7 @@ func (it *dedupSeriesIterator) Seek(t int64) bool {
 	// Don't use underlying Seek, but iterate over next to not miss gaps.
 	for {
 		ts, _ := it.At()
-		if ts > 0 && ts >= t {
+		if ts >= t {
 			return true
 		}
 		if !it.Next() {


pkg/api/query/v1.go
```diff
@@ -486,6 +486,15 @@ func (qapi *QueryAPI) series(r *http.Request) (interface{}, []error, *api.ApiErr
 		return nil, nil, &api.ApiError{Typ: api.ErrorBadData, Err: err}
 	}
 
+	var step time.Duration
+	if t := r.FormValue("step"); t != "" {
```
bwplotka, line 490: Series never had this, no? 

So why we are adding this? (https://prometheus.io/docs/prometheus/latest/querying/api/#finding-series-by-label-matchers)

pkg/query/iter.go
```diff
@@ -548,16 +554,18 @@ type dedupSeriesIterator struct {
 
 	penA, penB int64
 	useA       bool
+	hints      *storage.SelectHints
 }
 
-func newDedupSeriesIterator(a, b adjustableSeriesIterator) *dedupSeriesIterator {
+func newDedupSeriesIterator(a, b adjustableSeriesIterator, hints *storage.SelectHints) *dedupSeriesIterator {
```
bwplotka, line 560: Let's avoid tieing low-level algorithm into some struct with many fields. We only care about step - which also is far from being precise.

CHANGELOG.md
```diff
@@ -13,6 +13,7 @@ We use *breaking* word for marking changes that are not backward compatible (rel
 
 ### Fixed
 
+- [#3010](https://github.com/thanos-io/thanos/pull/3010) Querier: Data gaps when switching iterators.
```
kakkoyun, line 16:
```suggestion
- [#3010](https://github.com/thanos-io/thanos/pull/3010) Querier: Prevent data gaps while switching iterators.
```

pkg/query/querier_test.go
```diff
@@ -506,105 +575,127 @@ func TestQuerier_Select(t *testing.T) {
...
-				// This is required for testutil.Equals to work deterministically.
-				v = hackyStaleMarker
-			}
-			expectedCpy = append(expectedCpy, sample{t: s.t, v: v})
+func jsonToSeries(t *testing.T, filename string) []series {
+	file, err := ioutil.ReadFile(filename)
+	testutil.Ok(t, err)
+
+	data := Response{}
+	err = json.Unmarshal(file, &data)
```
bwplotka, line 619: you can inline

pkg/query/querier_test.go
```diff
@@ -506,105 +575,127 @@ func TestQuerier_Select(t *testing.T) {
...
+	for _, ser := range data.Data.Results {
+		var lbls labels.Labels
+		for n, v := range ser.Metric {
+			lbls = append(lbls, labels.Label{
+				Name:  string(n),
+				Value: string(v),
+			})
 		}
-		testutil.Equals(t, expectedCpy, samples, "samples for series %v does not match", i)
+		// Label names need to be sorted
```
bwplotka, line 631: missing trailing period

pkg/query/querier_test.go
```diff
@@ -506,105 +575,127 @@ func TestQuerier_Select(t *testing.T) {
...
-							w := res.Warnings()
-							testutil.Equals(t, 1, len(w))
-							testutil.Equals(t, tcase.expectedWarning, w[0].Error())
+				resolution := time.Duration(tcase.hints.Step) * time.Millisecond
+				t.Run(fmt.Sprintf("dedup=%v, resolution=%v", sc.dedup, resolution.String()), func(t *testing.T) {
+					var actual []series
+					// Boostrap a local store and pass the data through promql.
+					{
+						g := gate.New(2)
+						qCreator := &mockedQueryable{
```
bwplotka, line 585: Why we cannot do as previously? It was much cleaner IMO
```
timeout := 100 * time.Second	
		g := gate.New(2)	
		q := newQuerier(context.Background(), logger, nil, realSeriesWithStaleMarkerMint, realSeriesWithStaleMarkerMaxt, []string{"replica"}, nil, s, false, 0, true, false, g, timeout)	
		t.Cleanup(func() {	
			testutil.Ok(t, q.Close())	
		})	

		e := promql.NewEngine(promql.EngineOpts{	
			Logger:     logger,	
			Timeout:    timeout,	
			MaxSamples: math.MaxInt64,	
		})
```

pkg/query/querier_test.go
```diff
@@ -506,105 +575,127 @@ func TestQuerier_Select(t *testing.T) {
...
-}
 
-type mockedQueryable struct {
-	q storage.Querier
+	// Sort the series by their labels.
+	sort.Slice(sers, func(i, j int) bool {
+		return labels.Compare(sers[i].lset, sers[j].lset) <= 0
+	})
+
+	return sers
```
bwplotka, line 653: Never used before shortcut, let be consistent, maybe `ss` or `s`? or just `series`?

pkg/query/querier_test.go
```diff
@@ -1252,39 +1216,42 @@ func BenchmarkDedupSeriesIterator(b *testing.B) {
...
-			s1 = append(s1, sample{t: int64(i * 10000), v: 1})
-		}
-		for i := 0; i < b.N; i++ {
-			s2 = append(s2, sample{t: int64(i * 10000), v: 2})
-		}
-		run(b, s1, s2)
-	})
-	b.Run("fixed-delta", func(b *testing.B) {
-		var s1, s2 []sample
+	for _, step := range []int{1000, 5000, 10000} {
```
bwplotka, line 1220: Did you manage to gather some data/ results?

pkg/query/querier_test.go
```diff
@@ -483,13 +528,37 @@ func TestQuerier_Select(t *testing.T) {
...
+			// Regression test against https://github.com/thanos-io/thanos/issues/2890.
+			name: "when switching replicas make sure the time window between samples is never bigger then the lookback delta",
+			storeAPI: func() storepb.StoreServer {
+				s, err := store.NewLocalStoreFromJSONMmappableFile(logger, component.Debug, nil, "./testdata/issue2890-seriesresponses.json", store.ScanGRPCCurlProtoStreamMessages)
+				testutil.Ok(t, err)
+				return s
+			}(),
+			equivalentQuery: `cluster_version{}`,
+			replicaLabels:   []string{"replica"},
+			matchers: []*labels.Matcher{{
```
bwplotka, line 549: are those even used?

pkg/query/querier_test.go
```diff
@@ -653,152 +744,6 @@ func (s *mockedSeriesIterator) Next() bool {
...
-					{T: 1587692100000, V: 14.136842105263158}, {T: 1587692200000, V: 14.39298245614035}, {T: 1587692300000, V: 15.014035087719297}, {T: 1587692400000, V: 14.112280701754386}, {T: 1587692500000, V: 9.421065148148147}, {T: 1587692600000, V: 6.421368067203301},
-					{T: 1587692700000, V: 8.252631578947367}, {T: 1587692800000, V: 11.721237543747266}, {T: 1587692900000, V: 13.842105263157894}, {T: 1587693000000, V: 13.153509064307995}, {T: 1587693100000, V: 13.378947368421052}, {T: 1587693200000, V: 14.03157894736842},
-					{T: 1587693300000, V: 14.147368421052631}, {T: 1587693400000, V: 15.343159785693985}, {T: 1587693500000, V: 15.90877192982456}, {T: 1587693600000, V: 15.761403508771927},
-				}},
-			}, vec)
-		})
-		t.Run("Rate=30mStep=500s", func(t *testing.T) {
-			q, err := e.NewRangeQuery(&mockedQueryable{q}, `rate(gitlab_transaction_cache_read_hit_count_total[30m])`, timestamp.Time(realSeriesWithStaleMarkerMint).Add(30*time.Minute), timestamp.Time(realSeriesWithStaleMarkerMaxt), 500*time.Second)
```
bwplotka, line 716: Can't see 30m range test..

pkg/query/querier_test.go
```diff
@@ -848,16 +793,9 @@ func TestSortReplicaLabel(t *testing.T) {
 	}
 }
 
-const hackyStaleMarker = float64(-99999999)
-
 func expandSeries(t testing.TB, it chunkenc.Iterator) (res []sample) {
 	for it.Next() {
 		t, v := it.At()
-		// Nan != Nan, so substitute for another value.
```
bwplotka, line 856: Let's leave it - otherwise we will have nondeterministic result someday as Nans are tricky. What's the blocker for leaving those? (:

pkg/query/iter.go
```diff
@@ -575,6 +582,10 @@ func (it *dedupSeriesIterator) Next() bool {
 	// Advance both iterators to at least the next highest timestamp plus the potential penalty.
 	if it.aok {
 		it.aok = it.a.Seek(it.lastT + 1 + it.penA)
+		// When A is exhausted the seek for B below should be without any penalty
```
bwplotka, line 585: missing trailing period

bwplotka: Also this does not make sense, - why we treat A differently than B?

pkg/query/iter.go
```diff
@@ -604,31 +615,20 @@ func (it *dedupSeriesIterator) Next() bool {
...
+	// also not pick too many samples for the current resolution.
+	// Promql has a default look back delta of 5min so if the penalty causes the duration
+	// between 2 samples to be greater then the default look back it causes data gaps.
 	if it.useA {
-		if it.lastT != math.MinInt64 {
-			it.penB = 2 * (ta - it.lastT)
-		} else {
-			it.penB = initialPenalty
-		}
+		it.penB = it.step
```
bwplotka, line 624: Ok, I am pretty sure step based algo will not work. ): 

Let's take following query: `sum_over_time(my_metric[1h])` for 4d with step 1h. Let's say we have 15s scrape interval of this metric and 2 replicas with rollout in between. So let's say:

`a` replica samples (timestamps after start time): 0, 1, 2, 3,..... 59m58s, 59m59s, 1h, <rollout take 10m>, 1h20m, 1h20m1s, 1h20m2s...
`b` replica samples 1, 2, 3, 4, ..... every second always

Now our algo will choose replica `a` first and add step penalty (`1h`). At the last sample before rollout we will make penalty of B = step (1h), so next iteration we have:

1h20m form a, and 2h from b. This is a gap that could be totally avoided because `b` has that data. This will be highly visible on PromQL result despite the large step, as certain queries cares about each samples  - this has to be reliable ):

bwplotka (changes requested):
OK, this is critical path so let's take it carefully. I am not fan of changing all the tests and carefully adjusted test cases. 

Please be aware that we planned to have this algo stabilized to ensure we can run offline deduplication safely. With this we just start from scratch as we changed a bit - plus we change unit tests. This is quite risky.

I checked all and I don't think step based dedup will work ):

As commented below: 

Let's take following query: sum_over_time(my_metric[1h]) for 4d with step 1h. Let's say we have 15s scrape interval of this metric and 2 replicas with rollout in between. So let's say:

a replica samples (timestamps after start time): 0, 1, 2, 3,..... 59m58s, 59m59s, 1h, <rollout take 10m>, 1h20m, 1h20m1s, 1h20m2s...
b replica samples 1, 2, 3, 4, ..... every second always

Now our algo will choose replica a first and add step penalty (1h). At the last sample before rollout we will make penalty of B = step (1h), so next iteration we have:

1h20m form a, and 2h from b. This is a gap that could be totally avoided because b has that data. This will be highly visible on PromQL result despite the large step, as certain queries care about each samples - this has to be reliable ):


cc @brian-brazil if you have time.

Let's also do research on how others are doing it, but I believe the algo should be scrape interval-based - we need to assume it based on seen samples. I don't see any other way right now.


Curious if we could team up with promxy creator @jacksontj - we are in open source after all :hugs: 
We definitely work on something similar, he has some interesting approach to parallel queries on leafs: https://github.com/jacksontj/promxy/blob/master/pkg/proxystorage/proxy.go#L179 but cannot find any deduplication specifics.

brian-brazil:
> I checked all and I don't think step based dedup will work ):

As a matter of general principle, I agree. Due to races etc. you're going to need to look at 2-3 intervals minimum.

> I believe the algo should be scrape interval-based - we need to assume it based on seen samples. I don't see any other way right now.

Indeed, and there will always have to be a gap left for safety when you switch replicas, otherwise you're going to have issues with counters.

krasi-georgiev:
Turns out the deduplication alg works as expected and the problem is that the data was sent at an interval of 4.5min and the dedup alg uses 2xdelta so 9min which is more then the default  lookBackDelta in promql(5min).

Changed the PR to make the look back configurable.

cmd/thanos/query.go
```diff
@@ -70,6 +70,8 @@ func registerQuery(m map[string]setupFunc, app *kingpin.Application) {
 	maxConcurrentQueries := cmd.Flag("query.max-concurrent", "Maximum number of queries processed concurrently by query node.").
 		Default("20").Int()
 
+	lookbackDelta := cmd.Flag("query.lookback-delta", "The maximum lookback duration for retrieving metrics during expression evaluations.").Duration()
```
bwplotka, line 73: Can we describe it better? It's not maximum but actually lookback delta right? And also we should explain what it means for PromQL or link to docs :hugs:

CHANGELOG.md
```diff
@@ -52,6 +52,7 @@ sse_config:
 - [#2892](https://github.com/thanos-io/thanos/pull/2892) Receive: Receiver fails when the initial upload fails.
 - [#2865](https://github.com/thanos-io/thanos/pull/2865) ui: Migrate Thanos Ruler UI to React
 - [#2964](https://github.com/thanos-io/thanos/pull/2964) Query: Add time range parameters to label APIs. Add `start` and `end` fields to Store API `LabelNamesRequest` and `LabelValuesRequest`.
+- [#3010](https://github.com/thanos-io/thanos/pull/3010) Querier: Added a flag to set the default look back delta.
```
bwplotka, line 55: Can we move this to `unreleased`?

bwplotka (changes requested):
Thanks, LGTM, just minor nit in terms of description, also let's fix changelog.

And please remove `fixes` stuff - it does not fix any of the issues mentioned I think.

CHANGELOG.md
bwplotka: ditto

bwplotka (commented):
Still not addressed comments ):

cmd/thanos/query.go
krasi-georgiev: Added some more details, let me know what you think.

Unfortunately, there are official Prom docs about this flag.

CHANGELOG.md
krasi-georgiev: done

kakkoyun:
@krasi-georgiev We can include this in the next release candidate, so could you baed ut on `release-0.15` branch?

CHANGELOG.md
```diff
@@ -16,6 +16,7 @@ We use *breaking :warning:* word for marking changes that are not backward compa
 * [#3095](https://github.com/thanos-io/thanos/pull/3095) Rule: update manager when all rule files are removed.
 * [#3098](https://github.com/thanos-io/thanos/pull/3098) ui: Fix Block Viewer for Compactor and Store
 * [#3105](https://github.com/thanos-io/thanos/pull/3105) Query: Fix overwriting maxSourceResolution when auto downsampling is enabled.
+* [#3010](https://github.com/thanos-io/thanos/pull/3010) Querier: Added a flag to override the default look back delta in promql. The flag should be set to at least 2 times the slowest scrape interval.
```
bwplotka, line 19: Let's not forget to fix this on @kakkoyun PR (:

cmd/thanos/query.go
```diff
@@ -70,6 +70,8 @@ func registerQuery(m map[string]setupFunc, app *kingpin.Application) {
 	maxConcurrentQueries := cmd.Flag("query.max-concurrent", "Maximum number of queries processed concurrently by query node.").
 		Default("20").Int()
 
+	lookbackDelta := cmd.Flag("query.lookback-delta", "The maximum lookback duration for retrieving metrics during expression evaluations. For example with a look back of 7min and a scrape interval of 10min, promql will fill in the gap only for 7min and will show the other 3min as a gap. Should be set to at least 2 times the slowest scrape interval.").Duration()
```
bwplotka, line 73: Awesome, this is much better! 

We can be clear what it means in details and also add default here! (: 

```suggestion
	lookbackDelta := cmd.Flag("query.lookback-delta", "The maximum lookback duration for retrieving metrics during expression evaluations. PromQL always evaluates the query for the certain timestamp (query range timestamps are deduced by step). Since scrape intervals might be different, PromQL looks back for given amount of time to get latest sample. If it exceeds the maximum lookback delta it assumes series is stale and returns none (a gap). This is why lookback delta should be set to at least 2 times of the slowest scrape interval.").Default("5m").Duration()
```

WDYT?

bwplotka (approved):
One suggestion, otherwise LGTM! :muscle:

docs/components/query.md
```diff
@@ -367,6 +367,14 @@ Flags:
       --query.timeout=2m         Maximum time to process query by query node.
       --query.max-concurrent=20  Maximum number of queries processed
                                  concurrently by query node.
+      --query.lookback-delta=QUERY.LOOKBACK-DELTA
```
kakkoyun, line 370: @krasi-georgiev You need to run `make docs` again, to update this.

krasi-georgiev: done

kakkoyun (approved):
LGTM 🥇

PR #3011, opened 2020-08-10
Fixed shellcheck to be used from Gobin.

bwplotka:
Somehow for make lint was erroring with:

```
>> linting all of the shell script files
make: /home/bwplotka/Repos/thanos/tmp/bin/shellcheck: Command not found
make: *** [Makefile:303: shell-lint] Error 127
```

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -4,6 +4,7 @@ FILES_TO_FMT      ?= $(shell find . -path ./vendor -prune -o -name '*.go' -print
 DOCKER_IMAGE_REPO ?= quay.io/thanos/thanos
 DOCKER_IMAGE_TAG  ?= $(subst /,-,$(shell git rev-parse --abbrev-ref HEAD))-$(shell date +%Y-%m-%d)-$(shell git rev-parse --short HEAD)
 DOCKER_CI_TAG     ?= test
+SHELLCHECK ?= $(GOBIN)/shellcheck
 
 # Ensure everything works even if GOPATH is not set, which is often the case.
 # The `go env GOPATH` will work for all cases for Go 1.8+.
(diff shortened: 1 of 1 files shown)


bwplotka:
just guessing now really, ;p

Makefile
```diff
@@ -169,7 +169,7 @@ check-docs: $(EMBEDMD) $(LICHE) build
 .PHONY:shell-format
 shell-format: $(SHFMT)
 	@echo ">> formatting shell scripts"
-	@$(SHFMT) -i 2 -ci -w -s $(shell find . -type f -name "*.sh" -not -path "*vendor*" -not -path "tmp/*")
+	@$(SHFMT)shell -i 2 -ci -w -s $( find . -type f -name "*.sh" -not -path "*vendor*" -not -path "tmp/*")
```
kakkoyun, line 172: This doesn't seem right.
`make: /tmp/.bin/shfmt-v3.1.2shell: Command not found`

Makefile
```diff
@@ -231,6 +231,7 @@ install-deps: $(ALERTMANAGER) $(MINIO) $(PROMETHEUS_ARRAY)
 	@echo ">>GOBIN=$(GOBIN)"
 
 .PHONY: docker-ci
+docker-ci: export BIN_DIR= $(shell pwd)/tmp/bin
```
kakkoyun, line 234: Why do we need this if you are replacing `BIN_DIR` with `GOBIN`?

Makefile
bwplotka: yea this is totally wrong, just typo

Makefile
bwplotka: No idea, red herring - although relying on yet another bin dir seems wrong

PR #3018, opened 2020-08-11
query-frontend: Implement Thanos Codec to encode/decode Thanos requests

yeya24:

Currently in the frontend, we are using Cortex's PrometheusCodec to encode/decode HTTP requests. But this doesn't work well with Thanos since it doesn't support parsing Thanos specific query parameters, such as `partial_response`, `max_source_resolution`, `dedup`, etc.

This PR adds the implementation of Thanos Codec to support parsing these params, and then we can use them as part of the cache key or somewhere else.

TODO:
- More tests

Not sure about storeMatchers, since store matchers are mostly used for debugging, do we need to cache requests in this case?


cmd/thanos/query-frontend.go
```diff
@@ -62,6 +66,9 @@ func (c *queryRangeConfig) registerFlag(cmd *kingpin.CmdClause) {
 	cmd.Flag("query-range.response-cache-max-freshness", "Most recent allowed cacheable result, to prevent caching very recent results that might still be in flux.").
 		Default("1m").DurationVar(&c.cacheMaxFreshness)
 
+	cmd.Flag("query-range.partial-response", "Enable partial response for queries if no partial_response param is specified. --no-query.partial-response for disabling.").
```
kakkoyun, line 69: I guest it should be `--no-query-range`, no?
```suggestion
	cmd.Flag("query-range.partial-response", "Enable partial response for queries if no partial_response param is specified. --no-query-range.partial-response for disabling.").
```

yeya24: Good catch! Let me update the pr.

pkg/queryfrontend/codec.go
```diff
@@ -0,0 +1,307 @@
...
+		URL:        u,
+		Body:       http.NoBody,
+		Header:     http.Header{},
+	}
+
+	return req.WithContext(ctx), nil
+}
+
+func (c codec) DecodeResponse(ctx context.Context, r *http.Response, req queryrange.Request) (queryrange.Response, error) {
+	return c.prometheusCodec.DecodeResponse(ctx, r, req)
```
bwplotka, line 158: We could in theory embed Prometheus codec so we don't need to even implement those methods (:

pkg/queryfrontend/queryrange.proto
```diff
@@ -0,0 +1,31 @@
...
+package queryfrontend;
+
+import "gogoproto/gogo.proto";
+import "google/protobuf/duration.proto";
+import "store/storepb/types.proto";
+
+option (gogoproto.marshaler_all) = true;
+option (gogoproto.unmarshaler_all) = true;
+
+message ThanosRequest {
```
bwplotka, line 15: Hmmmmmmm sad we need to do it. Can we reuse SeriesRequest instead mabye? :thinking:

yeya24: What's the benefit of reusing it? SeriesRequest has some unnecessary fields like `aggregates` and `skipt_chunks` and `hints`. Also we also need `replicaLabels`, `dedup` fields in `ThanosRequest`. What do you think?

pkg/queryfrontend/codec.go
yeya24: Thanks, good catch! Let me update it.

pkg/queryfrontend/queryrange.proto
bwplotka: Well it allows to reduce number of protobuf code we have, and the code itself is non trivial.

And I don't think extra fields are the problem, just rather the fact that it's different level -> matchers vs query.

And the benefit of reuse is quite important. You change request for StoreAPI - most likely this one has to change as well.

Why do we need to have this protobuf BTW? It's all HTTP API? I believe if we want to suddenly have all HTTP API coded in protobuf (which nice idea BTW), probably we need to do it properly. Name it query API and make sure we can generate docs, and we have same structs for other HTTP APIs  (btw we have already LabelRequest, etc)

So after Storepb and this is really really similar if not the same (minus,plus few fields).

yeya24: Thanks, you are right. It is not necessary to have a protobuf for it. Using one struct is enough. Let me update it.

yeya24:
Proto is only used in Cortex and it seems that we don't really need it currently. So I removed the protobuf and update the pr. PTAL

pkg/queryfrontend/codec.go
```diff
@@ -0,0 +1,244 @@
...
+}
+
+func NewThanosCodec(partialResponse bool) *codec {
+	return &codec{
+		Codec:           queryrange.PrometheusCodec,
+		partialResponse: partialResponse,
+	}
+}
+
+func (c codec) DecodeRequest(_ context.Context, r *http.Request) (queryrange.Request, error) {
```
bwplotka, line 43: Amazing, wonder if we can use it in our Querier web logic even

pkg/queryfrontend/cache_splitter.go
```diff
@@ -0,0 +1,24 @@
...
+
+// constSplitter is a utility for using a constant split interval when determining cache keys.
+type constSplitter time.Duration
+
+// TODO(yeya24): Add other request params as request key.
+// GenerateCacheKey generates a cache key based on the Request and interval.
+func (t constSplitter) GenerateCacheKey(_ string, r queryrange.Request) string {
+	currentInterval := r.GetStart() / time.Duration(t).Milliseconds()
+	if tr, ok := r.(*ThanosRequest); ok {
+		return fmt.Sprintf("%s:%d:%d:%d", tr.Query, tr.Step, currentInterval, tr.MaxSourceResolution)
```
bwplotka, line 21: Hmmmm I would be careful here. E.g maxSourceResolution `1` and `2` will have exactly the same result, so no need to cache it differently.

Happy to start with this and improve later though.

Worth to add TODO here?

bwplotka (approved):
LGTM, just some thoughts. I think still it's good to be merged.

PR #3019, opened 2020-08-11
memcached: Handle unexpected network timeout errors

kakkoyun:

In addtition to `memcache: connect timeout to XXX.XXX.XXX.XXX:11211`, we have been seeing `read tcp XXX.XXX.XXX.XXX:59812->XXX.XXX.XXX.XXX:11211: i/o timeout`. This PR aims to address them.

## Changes

* Handle network timeout errors.

## Verification

* `make test-local`

### DIFF
--- a/pkg/cacheutil/memcached_client.go
+++ b/pkg/cacheutil/memcached_client.go
@@ -5,6 +5,7 @@ package cacheutil
 
 import (
 	"context"
+	"net"
 	"strconv"
 	"strings"
 	"sync"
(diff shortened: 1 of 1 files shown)


pkg/cacheutil/memcached_client.go
```diff
@@ -492,6 +473,27 @@ func (c *memcachedClient) getMultiSingle(ctx context.Context, keys []string) (it
...
+	var connErr *memcache.ConnectTimeoutError
+	var netErr net.Error
+	switch {
+	case errors.As(err, &connErr):
+		c.failures.WithLabelValues(op, reasonTimeout).Inc()
+	case errors.As(err, &netErr):
+		if netErr.Timeout() {
+			c.failures.WithLabelValues(op, reasonTimeout).Inc()
+		} else {
+			c.failures.WithLabelValues(op, reasonOther).Inc()
```
kakkoyun, line 486: I can introduce another label to cover `net.Error`. WDYT?

metalmatze: Sure. Doesn't sound too bad in terms of cardinality. Is there a specific reason we would need it right now?

kakkoyun: Noop. That's why I went with other. I just need to catch the timeout errors.

pracucci: Agree having a dedicated label value would be great here.

brancz: I don't feel this is problematic, it seems only helpful

pkg/cacheutil/memcached_client.go
```diff
@@ -492,6 +473,27 @@ func (c *memcachedClient) getMultiSingle(ctx context.Context, keys []string) (it
 	return items, err
 }
 
+func (c *memcachedClient) handleError(op string, err error) {
```
pracucci, line 476: Maybe "trackError" is more accurate then "handleError", given it just track metrics?

PR #3021, opened 2020-08-12
Add docs for query frontend

### LINKED ISSUE #3015: query-frontend: Docs
cc @yeya24

yeya24:

## Changes

Fixes #3015

### DIFF
--- a/docs/components/query-frontend.md
+++ b/docs/components/query-frontend.md
@@ -48,12 +48,13 @@ Query Frontend calculates the required subqueries and executes them in parallel 
 
 Currently, only in-memory cache (fifo cache) is supported. An example config:
 
+[embedmd]:# (../flags/config_response_cache_in_memory.txt yaml)
 ```yaml
-type: in-memory
+type: IN-MEMORY
 config:
-  max_size: 1MiB
-  max_size_items: 1000
-  validity: 6h
+  max_size: ""
+  max_size_items: 0
+  validity: 0s
 ```
 
 ### Slow Query Log
--- a/docs/components/query-frontend.md
+++ b/docs/components/query-frontend.md
@@ -1,0 +1,128 @@
+---
+title: Query Frontend
+type: docs
+menu: components
+---
+
+# Query Frontend
+
+The `thanos query-frontend` command implements a service that can be put in front of Thanos Queriers to improve the read path. It is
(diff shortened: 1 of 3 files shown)


docs/components/query-frontend.md
```diff
@@ -0,0 +1,128 @@
...
+  validity: 6h
+```
+
+### Slow Query Log
+
+Query Frontend supports `--query-frontend.log_queries_longer_than` flag to log queries running longer some duration.
+
+## Naming
+
+Naming is hard :) Please check [here](https://github.com/thanos-io/thanos/pull/2434#discussion_r408300683) to see why we chose `query-frontend` as the name.
```
pracucci, line 65: Should this be part of the component doc?

bwplotka: Yes, it's one of components, no?

pracucci: Yes. I was wondering if doc should focus on "why" as well (eg. in this case why the name has been picked).

bwplotka: Ah, good point. I think this is good enough - no better place I could think of. The reason is that ppl already ask why this name - so we don't want to repeat the work ;p

docs/components/query-frontend.md
```diff
@@ -0,0 +1,128 @@
...
+Query Frontend supports a retry mechanism to retry query when HTTP requests are failing. There is a `--query-range.max-retries-per-request` flag to limit the maximum retry times.
+
+### Caching
+
+Query Frontend supports caching query results and reuses them on subsequent queries. If the cached results are incomplete,
+Query Frontend calculates the required subqueries and executes them in parallel on downstream queriers. Query Frontend can optionally align queries with their step parameter to improve the cacheability of the query results.
+
+Currently, only in-memory cache (fifo cache) is supported. An example config:
+
+```yaml
```
bwplotka, line 51: this can be autogenerated. Can we do this in this PR maybe? (:

yeya24: I didn't know we can auto-generate this before. Let me take a look how to do it.

bwplotka: yup, looks on object storage page (:

yeya24: When investigating this, I found actually some of our embedmd configs are wrong like `[embedmd]: # "../flags/config_index_cache_in_memory.txt yaml"`, this syntax is not supported. So I also changed them in that commit

bwplotka: Thanks!

Issue #3023, opened 2020-08-12
Replace leaktest with goleak


bwplotka:
As Prometheus did we probably want to replace leaktest with https://github.com/uber-go/goleak with per package pattern:

```
func TestMain(m *testing.M) {
	goleak.VerifyTestMain(m)
}
```

This way we don't need to put defer on every unit test which we often forget. It will harder to detect WHERE the leak happens but for CI we care more if it happens.

bwplotka:
Help wanted!

mneverov:
hi, i'd like to work on this

bwplotka:
Please help! :heart:

Issue #234, opened 2018-03-05
Investigate potential go routine leak for testutil.NewBucketStore


bwplotka:
(Test more before putting issue on https://github.com/GoogleCloudPlatform/google-cloud-go/tree/master/storage package)

Reason:

<details><summary>This does not leak</summary>

```go
func NewObjectStoreBucket(t testing.TB) (*gcs.Bucket, func(), error) {
	_, ok := os.LookupEnv("GCP_PROJECT")
	// TODO(fabxc): make it run against a mock store if no actual bucket is configured.
	if !ok {
		t.Skip("test can only run against a defined GCP_PROJECT")
	}

	ctx, cancel := context.WithCancel(context.Background())
	gcsClient, err := storage.NewClient(ctx)
	if err != nil {
		cancel()
		return nil, nil, err
	}
	src := rand.NewSource(time.Now().UnixNano())
	name := fmt.Sprintf("test_%s_%x", strings.ToLower(t.Name()), src.Int63())
	fmt.Println(name)
	bkt := gcsClient.Bucket(name)
	//err = bkt.Create(ctx, project, nil)
	//if err != nil {
	//	cancel()
	//	gcsClient.Close()
	//	return nil, nil, err
	//}

	return gcs.NewBucket(name, bkt, nil), func() {
		//deleteAllBucket(t, ctx, bkt)
		cancel()
		gcsClient.Close()
	}, nil
}

func deleteAllBucket(t testing.TB, ctx context.Context, bkt *storage.BucketHandle) {
	var wg sync.WaitGroup

	objs := bkt.Objects(ctx, nil)
	for {
		oi, err := objs.Next()
		if err == iterator.Done {
			break
		}
		if err != nil {
			t.Logf("iterating over bucket objects failed: %s", err)
			wg.Wait()
			return
		}

		wg.Add(1)
		go func() {
			if err := bkt.Object(oi.Name).Delete(ctx); err != nil {
				t.Logf("deleting object %s failed: %s", oi.Name, err)
			}
			wg.Done()
		}()
	}
	wg.Wait()

	err := bkt.Delete(ctx)
	if err != nil {
		t.Logf("deleting bucket failed: %s", err)
	}
}
```
</details>

However, un-commenting either `bkt.Create` or `deleteAllBucket` or both leads to leaks:
<details><summary>Leaktest</summary>

```
	leaktest.go:90: leaktest: timed out checking goroutines
[...]

PR #3029, opened 2020-08-13
Replace Leaktest with Goleak

mneverov:

## Changes

Fixes: #3023
Replaced leaktest with goleak library.
Exposed [stop method](https://github.com/thanos-io/thanos/pull/3029/files#diff-2f6756629a1dc79c6131d406178ec0b4R242) in config watcher to correctly shut it down in case of invalid configuration.
Thanos still has goroutine leaks in storage tests with cloud providers, that's why I didn't use `VerifyTestMain` everywhere as it resulted in failed `block` and `store` tests ([see the pipeline](https://app.circleci.com/pipelines/github/thanos-io/thanos/2853/workflows/6f7cf06e-83cc-4e32-a24c-dfa3fa511f1d/jobs/11097)).
I found that [this issue](https://github.com/googleapis/google-cloud-go/issues/1025) was closed, but the `gcs` code was not adjusted. It seems that idle connections should be idle for a while for performance reasons and it is in general ok.
Also for the reference see [this issue](https://github.com/thanos-io/thanos/issues/234).
Goleak allows to exclude top functions, so `internal/poll.runtime_pollWait` might be added to the list of ignored functions, but it might hide real problems in other tests.

### DIFF
--- a/cmd/thanos/receive.go
+++ b/cmd/thanos/receive.go
@@ -373,6 +373,7 @@ func runReceive(
 		if cw != nil {
 			// Check the hashring configuration on before running the watcher.
 			if err := cw.ValidateConfig(); err != nil {
+				cw.Stop()
 				close(updates)
 				return errors.Wrap(err, "failed to validate hashring configuration file")
 			}
--- a/go.mod
+++ b/go.mod
@@ -16,7 +16,6 @@ require (
 	github.com/davecgh/go-spew v1.1.1
 	github.com/facette/natsort v0.0.0-20181210072756-2cd4dd1e2dcb
 	github.com/fatih/structtag v1.1.0
-	github.com/fortytw2/leaktest v1.3.0
 	github.com/fsnotify/fsnotify v1.4.7
 	github.com/go-kit/kit v0.10.0
 	github.com/go-openapi/strfmt v0.19.5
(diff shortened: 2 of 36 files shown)


mneverov:
@bwplotka please take a look

bwplotka:
> I found that this issue was closed, but the gcs code was not adjusted. It seems that idle connections should be idle for a while for performance reasons and it is in general ok.

Nice, let's reopen

bwplotka:
> Thanos still has goroutine leaks in storage tests with cloud providers, that's why I didn't use VerifyTestMain everywhere as it resulted in failed block and store tests (see the pipeline).

Can we ignore some go routines then? Explicitly? (: goleaks allows that.

bwplotka:
> Goleak allows to exclude top functions, so internal/poll.runtime_pollWait might be added to the list of ignored functions, but it might hide real problems in other tests.

I see your point - true that's bad, Let's keep the record on things that still are leaking in GH issues at least then (e.g add goleakMain for xxx package)

mneverov:
filed 3 issues: #3040 #3041 #3042

mneverov:
closed by mistake :scream:

CHANGELOG.md
```diff
@@ -50,6 +50,7 @@ We use *breaking* word for marking changes that are not backward compatible (rel
 - [#2931](https://github.com/thanos-io/thanos/pull/2931) Query: Allow passing a `storeMatcher[]` to select matching stores when debugging the querier. See [documentation](https://thanos.io/components/query.md/#store-filtering)
 - [#2991](https://github.com/thanos-io/thanos/pull/2991) store: `operation` label value `getrange` changed to `get_range` for `thanos_store_bucket_cache_operation_requests_total` and `thanos_store_bucket_cache_operation_hits_total` to be consistent with bucket operation metrics.
 - [#2876](https://github.com/thanos-io/thanos/pull/2876) Receive,Ruler: Updated TSDB and switched to ChunkIterators instead of sample one, which avoids unnecessary decoding / encoding.
+- [#3029](https://github.com/thanos-io/thanos/pull/3029) Replace leaktest with goleak
```
bwplotka, line 53: Let's remove this, it's not user impacting change (: (not implicitly)

mneverov: removed, thanks

PR #3030, opened 2020-08-13
Add instrumentation middlewares to query frontend

yeya24:

## Changes

This pr adds instrument middleware to query frontend handler, including logging and tracing.

The tracing middleware is important because we don't have a start span before query range tripperware. So range query starts from step_align span and there are no spans for other queries.

## Verification

Before

Screenshot from 2020-08-13 12-56-23

After

Query parameters are also logged in the first span.

Screenshot from 2020-08-13 12-58-10

Other queries are also working.

Screenshot from 2020-08-13 12-59-40

### DIFF
--- a/cmd/thanos/query-frontend.go
+++ b/cmd/thanos/query-frontend.go
@@ -7,6 +7,7 @@ import (
 	"net/http"
 	"time"
 
+	"github.com/NYTimes/gziphandler"
 	"github.com/cortexproject/cortex/pkg/querier/frontend"
 	"github.com/cortexproject/cortex/pkg/querier/queryrange"
 	"github.com/go-kit/kit/log"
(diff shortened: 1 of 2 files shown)


yeya24:
Hello @yashrsharma44, would you mind also giving a review when you get a chance? Since this pr also includes the new logging middleware

yeya24:
Still one question I have: do we need to implement `/status/*` endpoints for query frontend? If so, it cannot access downstream `/status/*` endpoints, or we need to set some special prefixs

yeya24:
Seems the two CI failures are not related to my change.

kakkoyun:
@yeya24 Let's not bother with those endpoint in this iteration. We can address later on, if they are needed. Our main aim for this PR is to have tracing starting from query frontend.

PR #3031, opened 2020-08-13
block: precalculate hashes if enabled and use them during compaction (downloading)

GiedriusS:

## Changes

Added a change to `objstore.DownloadDir` so that it would be possible to ignore certain directories. Added functions for calculating the hashes of files and directories. Used them inside of `shipper` & other places. Wired up the new `shipper` changes and all of the other writers to command-line parameters so that the users can optionally enable them. There should be no performance impact if none of the flags (the default) have been enabled.

## Verification

Unit tests. Manual testing. A new e2e test.


CHANGELOG.md
```diff
@@ -32,6 +32,7 @@ We use *breaking* word for marking changes that are not backward compatible (rel
 
 ### Added
 
+- [#3031](https://github.com/thanos-io/thanos/pull/3031) Compact/Sidecar/other writers: added `--add-hashes` and `--shipper.add-hashes` respectively. If enabled, they calculate SHA256 hashes of each file in a block before uploading them. If those hashes exist in the `meta.json` file then Compact becomes smart and does not download the files if they already exist on disk. This also means that the data directory passed to Thanos Compact is only *cleared once at boot*. So, if you, for example, use persistent volumes on k8s and your Thanos Compact crashes or fails to make an iteration properly then the last downloaded files are not wiped from the disk. The directories are only wiped again after a successful iteration.
```
bwplotka, line 35: Does it infinitely save data on disk and never delete it? :thinking: This is what I read from this

cmd/thanos/compact.go
```diff
@@ -268,12 +268,28 @@ func runCompact(
...
 
+	// Clean-up and create a fresh state at the beginning.
 	if err := os.RemoveAll(downsamplingDir); err != nil {
 		cancel()
 		return errors.Wrap(err, "clean working downsample directory")
 	}
 
-	grouper := compact.NewDefaultGrouper(logger, bkt, conf.acceptMalformedIndex, enableVerticalCompaction, reg, blocksMarkedForDeletion, garbageCollectedBlocks)
+	if err := os.RemoveAll(compactDir); err != nil {
+		cancel()
```
bwplotka, line 278: I think it's time for 
```
defer func() {
   if err  != nil {
      cancel()
   }
}()
```
(:

CHANGELOG.md
bwplotka: >  So, if you, for example, use persistent volumes on k8s and your Thanos Compact crashes or fails to make an iteration properly then the last downloaded files are not wiped from the disk

From the code it looks like they ARE wiped in fact, no?

cmd/thanos/compact.go
```diff
@@ -449,6 +465,7 @@ func runCompact(
 type compactConfig struct {
 	haltOnError                                    bool
 	acceptMalformedIndex                           bool
+	addHashes                                      bool
```
bwplotka, line 468: I am a fan of avoiding bool options. What about `hashFunction` setting when you set what kind of hash to use (: And by default it will be none?

cmd/thanos/compact.go
```diff
@@ -495,6 +512,9 @@ func (cc *compactConfig) registerFlag(cmd *kingpin.CmdClause) {
 	cmd.Flag("retention.resolution-1h", "How long to retain samples of resolution 2 (1 hour) in bucket. Setting this to 0d will retain samples of this resolution forever").
 		Default("0d").SetValue(&cc.retentionOneHr)
 
+	cmd.Flag("add-hashes", "Add hashes to meta file after producing new blocks which permits avoiding downloading some files twice albeit at some performance cost.").
```
bwplotka, line 515: `downloading some file twice`  is vague - does it mean Thanos downloads files couple of times without checking? Let's maybe not specify what's the purpose - there are many reasons. Mainly to ensure data integrity.

Can we at some point share what is that performance cost and during what moment? ( :

CHANGELOG.md
bwplotka: Ok I think we need to improve this (really 2x reduce) description, and just say "on-crash & recovery we don't delete data anymore but we are able to start up without redownload.

pkg/block/block.go
```diff
@@ -42,9 +42,28 @@ const (
...
 )
 
+// downloadDirFn is a function type that a function which downloads a directory from remote object
+// storage must implement. Used for mocking.
+type downloadDirFn func(context.Context, log.Logger, objstore.BucketReader, string, string, []string) error
+
 // Download downloads directory that is mean to be block directory.
-func Download(ctx context.Context, logger log.Logger, bucket objstore.Bucket, id ulid.ULID, dst string) error {
-	if err := objstore.DownloadDir(ctx, logger, bucket, id.String(), dst); err != nil {
+func Download(ctx context.Context, logger log.Logger, bucket objstore.Bucket, m *metadata.Meta, dst string, downloadDir downloadDirFn) error {
```
bwplotka, line 50: no need for full metadata really - just hashes, right?

pkg/block/block.go
```diff
@@ -42,9 +42,28 @@ const (
 	DebugMetas = "debug/metas"
 )
 
+// downloadDirFn is a function type that a function which downloads a directory from remote object
+// storage must implement. Used for mocking.
+type downloadDirFn func(context.Context, log.Logger, objstore.BucketReader, string, string, []string) error
```
bwplotka, line 47: This is bit confusing. Can we just have deleteDir function directly used and mock object storage?

pkg/block/metadata/hash.go
```diff
@@ -0,0 +1,100 @@
...
+	"io"
+	"os"
+	"path/filepath"
+	"strings"
+
+	"github.com/pkg/errors"
+)
+
+// HashKind indicates what type of hash it is.
+type HashKind string
```
bwplotka, line 19: It's not kind but maybe `function`?

```suggestion
type HashFunc string
```
or maybe type?

pkg/block/metadata/hash.go
```diff
@@ -0,0 +1,100 @@
...
+
+const (
+	// SHA256Kind shows that SHA256 has been used to generate the hash.
+	SHA256Kind HashKind = "SHA256"
+)
+
+// ObjectHash stores the hash of an object in the object storage.
+type ObjectHash struct {
+	Kind  HashKind
```
bwplotka, line 28:
```suggestion
	Func  HashFunc
```
? or type?

pkg/block/metadata/hash.go
```diff
@@ -0,0 +1,100 @@
...
+		if _, err := io.Copy(h, f); err != nil {
+			return ObjectHash{}, errors.Wrap(err, "copying")
+		}
+
+		return ObjectHash{
+			Kind:  SHA256Kind,
+			Value: hex.EncodeToString(h.Sum(nil)),
+		}, nil
+	default:
+		panic(fmt.Sprintf("hash kind %v is not supported; fill a bug, please", k))
```
bwplotka, line 58: We have error - let's use this instead.

The reason is compatibility. How would you add new function? (: Rollout Thanos in exactly same moment in all clusters?

pkg/block/metadata/hash.go
```diff
@@ -0,0 +1,100 @@
...
+	}); err != nil {
+		return ret, errors.Wrap(err, "walk")
+	}
+
+	return ret, nil
+}
+
+// NominalIgnoredFiles is a list of files that you want to typically ignore
+// when calculating the hashes of a TSDB block.
+var NominalIgnoredFiles = []string{DeletionMarkFilename, MetaFilename}
```
bwplotka, line 100: Why?

pkg/compact/compact.go
```diff
@@ -885,7 +898,9 @@ func NewBucketCompactor(
 // Compact runs compaction over bucket.
 func (c *BucketCompactor) Compact(ctx context.Context) (rerr error) {
 	defer func() {
-		if IsHaltError(rerr) {
+		// Leave the compact directory for inspection if it is a halt error
```
bwplotka, line 901: so we don't check halt. I think you mean any error, which is ok - as long it's not full disk error ;p

bwplotka: also everywhere else

pkg/objstore/objstore.go
```diff
@@ -200,15 +200,21 @@ func DownloadFile(ctx context.Context, logger log.Logger, bkt BucketReader, src,
...
 
 	var downloadedFiles []string
 	if err := bkt.Iter(ctx, src, func(name string) error {
 		if strings.HasSuffix(name, DirDelim) {
-			return DownloadDir(ctx, logger, bkt, name, filepath.Join(dst, filepath.Base(name)))
+			return DownloadDir(ctx, logger, bkt, name, filepath.Join(dst, filepath.Base(name)), ignoredPaths)
+		}
+		for _, ignoredPath := range ignoredPaths {
+			if name == ignoredPath {
+				level.Info(logger).Log("msg", "not downloading again because the hash of the file on disk matches", "file", name)
```
pracucci, line 215: This is a generic implementation which just takes in input `ignoredPaths`. In my opinion, this function should know nothing about the fact it will be called with `ignoredPaths` based on content hash comparison. For this reason, I would generalise this message to not mention the hashing thing.

pkg/block/block.go
```diff
@@ -42,9 +42,28 @@ const (
...
+
 // Download downloads directory that is mean to be block directory.
-func Download(ctx context.Context, logger log.Logger, bucket objstore.Bucket, id ulid.ULID, dst string) error {
-	if err := objstore.DownloadDir(ctx, logger, bucket, id.String(), dst); err != nil {
+func Download(ctx context.Context, logger log.Logger, bucket objstore.Bucket, m *metadata.Meta, dst string, downloadDir downloadDirFn) error {
+	ignoredPaths := []string{}
+	if len(m.Thanos.Hashes) > 0 {
```
pracucci, line 52: No need for this `if`. The next `for` will be a no-op if `m.Thanos.Hashes` is empty.

pkg/block/metadata/hash.go
```diff
@@ -0,0 +1,100 @@
...
+			Value: hex.EncodeToString(h.Sum(nil)),
+		}, nil
+	default:
+		panic(fmt.Sprintf("hash kind %v is not supported; fill a bug, please", k))
+	}
+}
+
+// GetHashDir calculates the hashes of the given kind of all files inside of dir.
+// Ignores the given files.
+func GetHashDir(dir string, k HashKind, ignore ...string) (map[string]ObjectHash, error) {
```
pracucci, line 64: Given this is `CalculateHash` while iterating over a directory, I'm wondering if we should name this function `CalculateHashDir`

pracucci (commented):
Thanks for working on this! The overall design LGTM. I agree with Bartek's comments plus I've added a couple of minor ones. Thanks!

pkg/block/metadata/hash.go
GiedriusS: Well, the deletion mark could actually just be removed from here since it will never appear while a block is being produced. However, I think the `meta.json` file should be definitely excluded because there will be confusion because the hash will most likely change *after* adding it to `m.Thanos.Hashes`.

pkg/verifier/safe_delete.go
```diff
@@ -62,7 +62,12 @@ func BackupAndDelete(ctx context.Context, logger log.Logger, bkt, backupBkt objs
 
 	// Download the TSDB block.
 	dir := filepath.Join(tempdir, id.String())
-	if err := block.Download(ctx, logger, bkt, id, dir); err != nil {
+
+	meta, err := block.DownloadMeta(ctx, logger, bkt, id)
```
bwplotka, line 66: Why we need to do this?

pkg/shipper/shipper.go
```diff
@@ -358,6 +360,12 @@ func (s *Shipper) upload(ctx context.Context, meta *metadata.Meta) error {
 		meta.Thanos.Labels = lset.Map()
 	}
 	meta.Thanos.Source = s.source
+	hashes, err := metadata.CalculateHashDir(updir, metadata.SHA256Func, metadata.NominalIgnoredFiles...)
```
bwplotka, line 363:
```suggestion
	meta.Thanos.Hashes, err = metadata.CalculateHashDir(updir, metadata.SHA256Func, metadata.NominalIgnoredFiles...)
```

pkg/block/metadata/hash.go
bwplotka: I mean, I would just rename it tbh or just use `meta.json` explicitly somewhere. `Nominal` does not mean anything to me :hugs: 

But overall it makes sense to skip hashing meta.json, it would hard... to achieve (:

bwplotka (approved):
LGTM, just minor comments.

cmd/thanos/receive.go
```diff
@@ -94,6 +95,9 @@ func registerReceive(app *extkingpin.App) {
 			"This can trigger compaction without those blocks and as a result will create an overlap situation. Set it to true if you have vertical compaction enabled and wish to upload blocks as soon as possible without caring"+
 			"about order.").
 		Default("false").Hidden().Bool()
+	hashFunc := cmd.Flag("shipper.hash-func",
+		"Specify which hash function to use when calculating the hashes of produced files. If no function has been specified, it does not happen. This permits avoiding downloading some files twice albeit at some performance cost. Possible values are: \"\", \"SHA256\".").
+		Default("").Enum("", "SHA256")
```
prmsrswt, line 100: As this flag is used in multiple places, maybe we should extract out it in a function in `cmd/thanos/flags.go`.

pkg/block/metadata/meta.go
```diff
@@ -59,6 +59,9 @@ type Thanos struct {
 
 	// Source is a real upload source of the block.
 	Source SourceType `json:"source"`
+
+	// Hashes is a map of hashes inside of the block.
```
prmsrswt, line 63: I guess you mean, `// Hashes is a map of hashes of files inside the block.`

prmsrswt (approved):
Just some minor nits, everything else looks good to me!

cmd/thanos/receive.go
GiedriusS: I agree, let's do it. However, how should we call it? Simply `--hash-func`? WDYT @prmsrswt, @bwplotka ?

pkg/block/metadata/meta.go
GiedriusS: :smile: yes, indeed. Thank you for spotting this! :heart:

pkg/verifier/safe_delete.go
GiedriusS: Good point, I have removed this change.

cmd/thanos/receive.go
prmsrswt: I guess `--hash-func` would work.

prmsrswt:
One more thing that came to my mind was, after using this feature, if someone decides to switch the hashing algorithm, al previous hashes would become unusable and we will trigger a redownload even though we have the correct files.

I guess we can also store the hashing algorithm used to create the hash alongside it so that we can calculate and match hashes using that algorithm irrespective of what algorithm is currently selected. This will also help us in identifying the hashing algorithm at places like Block Viewer and then showing some info to the user.

GiedriusS:
Working on rebasing this.

GiedriusS:
@bwplotka @prmsrswt @pracucci I have rebased this on `master`. PTAL! :muscle: This version looks a bit cleaner to me because we now have that `Files` array in `meta.json`.

GiedriusS:
I believe this is ready for another round of reviews. There are some potential cleanups to be had by exposing `gatherFileStats` and reusing it in `e2eutil` however that leads to circular dependencies which I'd like to solve in a separate PR as to not get distracted here. It's already a couple of months for this PR ;) 

I also played around with this locally. An excerpt from a `meta.json` file:

```json
                        {                                                                                                                                                                                  
                                "rel_path": "index",                                                                                                                                                       
                                "size_bytes": 212315,
                                "hash": {                                                            
                                        "Func": "SHA256",
                                        "Value": "c7fbe6c03c8c2a89e8417a6124fe37452b7b22a2b78934d9b25bd38815c81c1f"
                                }
                        },
                        {                                                                            
                                "rel_path": "meta.json"
                        }
```

I have double-checked that the hash is correct with `sha256sum`. PTAL @thanos-io/thanos-maintainers  :heart:

pkg/block/block.go
```diff
@@ -45,12 +45,40 @@ const (
...
+	}
+
+	ignoredPaths := []string{MetaFilename}
+	for _, fl := range m.Thanos.Files {
+		if fl.Hash == nil || fl.RelPath == "" {
+			continue
+		}
+		actualHash, err := metadata.CalculateHash(filepath.Join(dst, fl.RelPath), fl.Hash.Func)
+		if err != nil {
+			logger.Log("failed to calculate hash of %s due to %v, continuing", fl.RelPath, err)
```
bwplotka, line 67:
```suggestion
			level.Error(logger).Log("msg", "failed to calculate hash when downloading; re-downloading", "relPath", fl.RelPath, "err", err)
```
I think we should have some lint to check wrong log usage (:

bwplotka: Added https://github.com/thanos-io/thanos/issues/3680

pkg/block/block.go
```diff
@@ -45,12 +45,40 @@ const (
...
+		return err
+	}
+	m, err := metadata.ReadFromDir(dst)
+	if err != nil {
+		return errors.Wrapf(err, "reading meta from %s", dst)
+	}
+
+	ignoredPaths := []string{MetaFilename}
+	for _, fl := range m.Thanos.Files {
+		if fl.Hash == nil || fl.RelPath == "" {
```
bwplotka, line 62:
```suggestion
		if fl.Hash == nil || fl.Hash == metadata.None || fl.RelPath == "" {
```

pkg/block/metadata/hash_test.go
```diff
@@ -0,0 +1,31 @@
...
+
+import (
+	"io/ioutil"
+	"os"
+	"testing"
+
+	"github.com/thanos-io/thanos/pkg/testutil"
+)
+
+func TestHashSmoke(t *testing.T) {
```
bwplotka, line 14: `smoke`



Did not hear that word for some time... ;p Smoke test is valid naming though (:

pkg/compact/compact.go
```diff
@@ -874,10 +881,25 @@ func NewBucketCompactor(
 	}, nil
 }
 
+// removeAllNonCGDirs removes all subdirectories of the given directory
+// that are not part of the given compaction groups, and all files.
+// This is needed in the case when the compaction process resumes without
+// restarting the whole process but the blocks do not exist in the remote
+// object storage anymore.
+func removeAllNonCGDirs(groups []*Group, compactDir string) error {
```
bwplotka, line 889: Same, let's inline

pkg/objstore/inmem.go
```diff
@@ -24,14 +24,18 @@ type InMemBucket struct {
 	mtx     sync.RWMutex
 	objects map[string][]byte
 	attrs   map[string]ObjectAttributes
+
+	downloadedCount     map[string]uint
```
bwplotka, line 28: I would vote for just using MetricBucket wrapper instead. You can just use count metric (not needed per file) right?

The first reason is that we already have an implementation for tracking this. Secondly, it's because it's not testing bucket (by name), it's in mem bucket. While it's recommended for testing usage, let's keep it production-grade if we can (: 

WDYT?

pkg/objstore/objstore.go
```diff
@@ -201,15 +201,21 @@ func DownloadFile(ctx context.Context, logger log.Logger, bkt BucketReader, src,
 }
 
 // DownloadDir downloads all object found in the directory into the local directory.
-func DownloadDir(ctx context.Context, logger log.Logger, bkt BucketReader, src, dst string) error {
+func DownloadDir(ctx context.Context, logger log.Logger, bkt BucketReader, originalSrc, src, dst string, ignoredPaths []string) error {
```
bwplotka, line 204:
```suggestion
func DownloadDir(ctx context.Context, logger log.Logger, bkt BucketReader, originalSrc, src, dst string, ignorePaths ...string) error {
```

bwplotka: For compatibility and indication that those are optional 🤗

pkg/objstore/objstore.go
```diff
@@ -201,15 +201,21 @@ func DownloadFile(ctx context.Context, logger log.Logger, bkt BucketReader, src,
...
 
 	var downloadedFiles []string
 	if err := bkt.Iter(ctx, src, func(name string) error {
 		if strings.HasSuffix(name, DirDelim) {
-			return DownloadDir(ctx, logger, bkt, name, filepath.Join(dst, filepath.Base(name)))
+			return DownloadDir(ctx, logger, bkt, originalSrc, name, filepath.Join(dst, filepath.Base(name)), ignoredPaths)
+		}
+		for _, ignoredPath := range ignoredPaths {
+			if ignoredPath == strings.TrimPrefix(name, string(originalSrc)+DirDelim) {
+				level.Info(logger).Log("msg", "not downloading again because a provided path matches this one", "file", name)
```
bwplotka, line 216: Do we need this log line? 🤔 Might be spammy? Maybe debug level?

pkg/runutil/runutil.go
```diff
@@ -158,3 +159,41 @@ func ExhaustCloseWithErrCapture(err *error, r io.ReadCloser, format string, a ..
 
 	*err = merr.Err()
 }
+
+// DeleteAllExceptDirs deletes all files and directories inside the given
+// dir except for the ignoreDirs directories.
+func DeleteAllExceptDirs(dir string, ignoreDirs []string) error {
```
bwplotka, line 165:
```suggestion
func DeleteAll(dir string, ignoreDirs ..string) error {
```

This would be clear enough IMO WDYT?

pkg/testutil/e2eutil/prometheus.go
```diff
@@ -475,10 +479,39 @@ func createBlock(
...
+		err := filepath.Walk(blockDir, func(path string, info os.FileInfo, err error) error {
+			if info.IsDir() {
+				return filepath.SkipDir
+			}
+			paths = append(paths, path)
+			return nil
+		})
+		if err != nil {
+			return id, errors.Wrapf(err, "walking %s", dir)
+		}
```
bwplotka, line 495:
```suggestion
		if err := filepath.Walk(blockDir, func(path string, info os.FileInfo, err error) error {
			if info.IsDir() {
				return filepath.SkipDir
			}
			paths = append(paths, path)
			return nil
		}); err != nil {
			return id, errors.Wrapf(err, "walking %s", dir)
		}
```

pkg/testutil/e2eutil/prometheus.go
```diff
@@ -344,8 +344,9 @@ func CreateBlock(
 	mint, maxt int64,
 	extLset labels.Labels,
 	resolution int64,
+	addHashes bool,
```
bwplotka, line 347: I would vote for consistency and getting hashFunc instead like everywhere.

bwplotka (changes requested):
Not sure if this changed overtime a lot or it's just me being rested after some holidays, but I found more suggestions to make, sorry for that 🙈 

Thanks for this. It looks generally ok, just some style / readability suggestions, and one major question.

CHANGELOG.md
```diff
@@ -15,6 +15,7 @@ We use _breaking :warning:_ to mark changes that are not backward compatible (re
 ### Added
 
 - [#3469](https://github.com/thanos-io/thanos/pull/3469) StoreAPI: Added `hints` field to `LabelNamesRequest` and `LabelValuesRequest`. Hints in an opaque data structure that can be used to carry additional information from the store and its content is implementation specific.
+- [#3031](https://github.com/thanos-io/thanos/pull/3031) Compact/Sidecar/other writers: added `--hash-func`. If some function has been specified, writers calculate hashes using that function of each file in a block before uploading them. If those hashes exist in the `meta.json` file then Compact does a smart thing and does not download the files if they already exist on disk. This also means that the data directory passed to Thanos Compact is only *cleared once at boot* or *if everything succeeds*. So, if you, for example, use persistent volumes on k8s and your Thanos Compact crashes or fails to make an iteration properly then the last downloaded files are not wiped from the disk. The directories that were created the last time are only wiped again after a successful iteration or if the previously picked up blocks have disappeared.
```
MalloZup, line 18: `nit` I wonder if `does a smart thing` belong to the changelog.

Imho I would just remove it in sake of formal language.

pkg/block/metadata/hash.go
```diff
@@ -0,0 +1,60 @@
...
+
+// CalculateHash calculates the hash of the given type.
+func CalculateHash(p string, hf HashFunc) (ObjectHash, error) {
+	switch hf {
+	case SHA256Func:
+		f, err := os.Open(p)
+		if err != nil {
+			return ObjectHash{}, errors.Wrap(err, "opening file")
+		}
+		defer f.Close()
```
MalloZup, line 45: I guess here we could apply the Coding-style guide?
https://github.com/thanos-io/thanos/blob/master/docs/contributing/coding-style-guide.md#defers-dont-forget-to-check-returned-errors

pkg/compact/compact.go
```diff
@@ -874,10 +881,25 @@ func NewBucketCompactor(
...
+		ignoreDirs = append(ignoreDirs, gr.Key())
+	}
+	return runutil.DeleteAllExceptDirs(compactDir, ignoreDirs)
+}
+
 // Compact runs compaction over bucket.
 func (c *BucketCompactor) Compact(ctx context.Context) (rerr error) {
 	defer func() {
-		if IsHaltError(rerr) {
+		// Leave the compact directory for inspection if it is a halt error
```
MalloZup, line 900: this comment although appreciated, from my pov we could improve its readability.

Especially the 2nd line, has imho to much logic word. (or if it is not not, then) which maybe could simplified the sentence and improve readability imho

pkg/runutil/runutil.go
MalloZup: yep

MalloZup (commented):
good work ! :+1:  I just left some minor comment and took the occasion to review from curiosity pov to learn more the code .. :grin:

cmd/thanos/tools_bucket.go
```diff
@@ -483,9 +483,11 @@ func registerBucketDownsample(app extkingpin.AppClause, objStoreConfig *extflag.
 	httpAddr, httpGracePeriod := extkingpin.RegisterHTTPFlags(cmd)
 	dataDir := cmd.Flag("data-dir", "Data directory in which to cache blocks and process downsamplings.").
 		Default("./data").String()
+	hashFunc := cmd.Flag("hash-func", "Specify which hash function to use when calculating the hashes of produced files. If no function has been specified, it does not happen. This permits avoiding downloading some files twice albeit at some performance cost. Possible values are: \"\", \"SHA256\".").
```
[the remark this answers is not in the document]
GiedriusS: Because it doesn't make sense to add it for both writers _and_ readers. It would not mean anything for components that only read and would only raise confusion e.g. `bucket ls`. As for separating this into a helper or something - maybe. I had an idea that we could separate the uploading part from `Shipper` into a separate struct/package and use that inside of shipper + here. But that is a completely separate story that I think we can do later given that such a refactoring does not impact the end user at all.

CHANGELOG.md
GiedriusS: Removed!

cmd/thanos/compact.go
```diff
@@ -286,11 +291,23 @@ func runCompact(
 		downsamplingDir = path.Join(conf.dataDir, "downsample")
 	)
 
+	// Clean-up and create a fresh state at the beginning.
 	if err := os.RemoveAll(downsamplingDir); err != nil {
```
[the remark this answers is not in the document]
GiedriusS: You are correct - we can take a step further and do not delete these at the beginning due to the reasons you've mentioned.

GiedriusS:
@bwplotka @MalloZup I have fixed this PR according to your comments, added more extensive unit tests + added a e2e test which has caught some nasty bugs. I truly believe this is ready for prime time. PTAL! :hugs:

cmd/thanos/compact.go
```diff
@@ -376,15 +384,15 @@ func runCompact(
...
+			if err := downsampleBucket(ctx, logger, downsampleMetrics, bkt, sy.Metas(), downsamplingDir, metadata.HashFunc(conf.hashFunc)); err != nil {
 				return errors.Wrap(err, "first pass of downsampling failed")
 			}
 
 			level.Info(logger).Log("msg", "start second pass of downsampling")
 			if err := sy.SyncMetas(ctx); err != nil {
 				return errors.Wrap(err, "sync before second pass of downsampling")
 			}
-			if err := downsampleBucket(ctx, logger, downsampleMetrics, bkt, sy.Metas(), downsamplingDir); err != nil {
+			if err := downsampleBucket(ctx, logger, downsampleMetrics, bkt, sy.Metas(), downsamplingDir, metadata.HashFunc(conf.hashFunc)); err != nil {
```
bwplotka, line 395: I don't think you need to do it, won't it work without casting in this case? 🤔

cmd/thanos/downsample.go
```diff
@@ -158,15 +159,18 @@ func downsampleBucket(
...
+	hashFunc metadata.HashFunc,
+) (rerr error) {
 	if err := os.MkdirAll(dir, 0777); err != nil {
 		return errors.Wrap(err, "create dir")
 	}
 
 	defer func() {
+		// Leave the downsample directory for inspection if it is a halt error
+		// or if it is not then so that possibly we would not have to download everything again.
+		if rerr != nil {
```
bwplotka, line 171: What if it's disk full issue?

pkg/block/block.go
```diff
@@ -43,14 +43,44 @@ const (
 	DebugMetas = "debug/metas"
 )
 
-// Download downloads directory that is mean to be block directory.
+// Download downloads directory that is mean to be block directory. If any of the files
+// have a hash calculated in the meta file and it matches with what is in the destination path then
+// we do not download it. We always download the meta file.
```
bwplotka, line 48:
```suggestion
// we do not download it. We always re-download the meta file.
```

pkg/compact/compact.go
```diff
@@ -877,7 +884,10 @@ func NewBucketCompactor(
 // Compact runs compaction over bucket.
 func (c *BucketCompactor) Compact(ctx context.Context) (rerr error) {
 	defer func() {
-		if IsHaltError(rerr) {
+		// Do not remove the compactDir if an error has occurred
```
bwplotka, line 887: 👍🏽

pkg/compact/compact.go
```diff
@@ -950,6 +955,17 @@ func (c *BucketCompactor) Compact(ctx context.Context) (rerr error) {
 			return errors.Wrap(err, "build compaction groups")
 		}
 
+		ignoreDirs := []string{}
+		for _, gr := range groups {
+			ignoreDirs = append(ignoreDirs, gr.Key())
+		}
+
+		files, _ := ioutil.ReadDir(c.compactDir)
```
bwplotka, line 963: I think we need to check err, no?

pkg/compact/compact.go
```diff
@@ -950,6 +955,17 @@ func (c *BucketCompactor) Compact(ctx context.Context) (rerr error) {
...
+		level.Warn(c.logger).Log("files", fmt.Sprintf("%+v", files), "ignoringdirs", fmt.Sprintf("%+v", ignoreDirs))
```
bwplotka, line 964: Let's add `msg` to tell what this is warning about (:

pkg/store/bucket_test.go
```diff
@@ -1054,7 +1054,7 @@ func uploadTestBlock(t testing.TB, tmpDir string, bkt objstore.Bucket, series in
 		Source:     metadata.TestSource,
 	}, nil)
 	testutil.Ok(t, err)
-	testutil.Ok(t, block.Upload(context.Background(), logger, bkt, filepath.Join(tmpDir, "tmp", id.String())))
+	testutil.Ok(t, block.Upload(context.Background(), logger, bkt, filepath.Join(tmpDir, "tmp", id.String()), metadata.NoneFunc))
```
bwplotka, line 1057: There might be some time for Uploader struct , but not in this PR 🤗

bwplotka (approved):
Only minor logging nit, otherwise this looks good to me. Good tests and logic. 👍🏽  LGTM after this tiny log fix.

cmd/thanos/compact.go
GiedriusS: Unfortunately but it wouldn't work without casting:
```
cannot use conf.hashFunc (variable of type string) as metadata.HashFunc value in argument to downsampleBucket
```

pkg/compact/compact.go
GiedriusS: Actually, I have added this while trying to debug e2e tests. It's not needed at all so I have removed it.

pkg/compact/compact.go
GiedriusS: This was good, old printf-debugging in e2e tests that I forgot to remove and hence not needed since I have figured out the problem since. The culprit was that I had forgotten to remove the `replica` label before trying to get the grouping key for the tests, and then the directory that I had been expecting was removed.

pkg/store/bucket_test.go
GiedriusS: Yep :+1: it is starting to get a lot of arguments so we should refactor it into something more understandable

cmd/thanos/downsample.go
GiedriusS: Then it's not a problem because we have a consistent order of work with https://github.com/thanos-io/thanos/pull/3843 in place + https://github.com/thanos-io/thanos/blob/6877bd17aa3ebee44304aee4b2c6582be6f860cc/pkg/compact/compact.go#L312-L314 this exists already. This means that we are sure that after a retry, we are going to restart the work on the same blocks. If the blocks would disappear then accordingly we would get rid of them and then Thanos Compactor can continue its work. Does it make sense, @bwplotka ?

bwplotka: Makes sense 👍🏽

bwplotka (approved):
LGTM!

Rebase and feel free to merge 🤗

GiedriusS:
:tada:

PR #3032, opened 2020-08-14
Support memcached for query frontend

yeya24:

## Changes

Add support for Memcached cache for query frontend.


pkg/queryfrontend/cache/memcached.go
```diff
@@ -0,0 +1,145 @@
...
+		memcached: memcached,
+		validity:  validity,
+	}
+
+	if c.validity == 0 {
+		level.Info(logger).Log("msg", "memcached cache valid time set to 0, use 24 hours instead")
+		c.validity = memcachedDefaultTTL
+	}
+
+	c.requests = promauto.With(reg).NewCounter(prometheus.CounterOpts{
```
kakkoyun, line 83: We have the same metrics for store cache https://github.com/thanos-io/thanos/blob/798f40b5e7f7bffb5c9713ad72bc70ce2af7e411/pkg/cache/memcached.go#L35

This won't create a registration issue, however better to add a constant label (e.g name) to create distinction between them.

yeya24: Makes sense. If I add the const label, it is necessary to also add the `name` label to the store cache metrics as well?

If it is this case, then maybe let me rename the two metrics to `thanos_response_cache_memcached_requests_total` and `thanos_response_cache_memcached_hits_total`

kakkoyun: Store cache metrics already have this I believe. You just need to add the name here as well. 

> If it is this case, then maybe let me rename the two metrics to thanos_response_cache_memcached_requests_total and thanos_response_cache_memcached_hits_total

Let's go with the name approach.

yeya24: Thank you, let me update my pr.

pkg/cache/memcached.go
```diff
@@ -81,3 +81,42 @@ func (c *MemcachedCache) Fetch(ctx context.Context, keys []string) map[string][]
...
 }
+
+// mockedMemcachedClient is a mocked memcached client for testing.
+type mockedMemcachedClient struct {
+	cache             map[string][]byte
+	mockedGetMultiErr error
+}
+
+// NewMockedMemcachedClient returns a mocked memcached client.
+func NewMockedMemcachedClient(mockedGetMultiErr error) *mockedMemcachedClient {
```
squat, line 92: tiny nit: maybe it's clear already that the error is mocked because it

pkg/cache/memcached.go
```diff
@@ -81,3 +81,42 @@ func (c *MemcachedCache) Fetch(ctx context.Context, keys []string) map[string][]
...
+	return &mockedMemcachedClient{
+		cache:             map[string][]byte{},
+		mockedGetMultiErr: mockedGetMultiErr,
```
squat, line 95: tiny nit: maybe it's clear already that the error is mocked because it is a field on the `mockedMemcachedClient` and we can shorten this to `getMultiError`

squat: hm i see this is from before the PR

pkg/queryfrontend/cache/memcached.go
```diff
@@ -0,0 +1,145 @@
...
+}
+
+// newMemcachedCache makes a new MemcachedCache and returns Cortex ResultsCacheConfig.
+func newMemcachedCache(conf []byte, logger log.Logger, reg prometheus.Registerer) (*queryrange.ResultsCacheConfig, error) {
+	var config MemcachedResponseCacheConfig
+	if err := yaml.UnmarshalStrict(conf, &config); err != nil {
+		return nil, err
+	}
+
+	memcachedClientConfig, err := yaml.Marshal(config.Memcached)
```
squat, line 49: Why are we re-marshaling the inlined config.Memcached field? We could directly pass `conf` to the `cacheutil.NewMemcachedClient` func and save this extra step, no?

pkg/queryfrontend/cache/memcached.go
```diff
@@ -0,0 +1,145 @@
...
+	reg prometheus.Registerer,
+) *MemcachedCache {
+	c := &MemcachedCache{
+		logger:    logger,
+		memcached: memcached,
+		validity:  validity,
+	}
+
+	if c.validity == 0 {
+		level.Info(logger).Log("msg", "memcached cache valid time set to 0, use 24 hours instead")
```
squat, line 79: `use 24 hours instead` this sounds like a warning to me

pkg/queryfrontend/cache/memcached.go
```diff
@@ -0,0 +1,145 @@
...
+		c.validity = memcachedDefaultTTL
+	}
+
+	c.requests = promauto.With(reg).NewCounter(prometheus.CounterOpts{
+		Name: "thanos_cache_memcached_requests_total",
+		Help: "Total number of items requests to memcached.",
```
squat, line 85: Is this really `items requests`? Or could it be `Total number of requests to memcached.`

pkg/queryfrontend/cache/memcached.go
```diff
@@ -0,0 +1,145 @@
...
+	})
+
+	c.hits = promauto.With(reg).NewCounter(prometheus.CounterOpts{
+		Name: "thanos_cache_memcached_hits_total",
+		Help: "Total number of items requests to the cache that were a hit.",
```
squat, line 90: same here

pkg/queryfrontend/cache/memcached.go
```diff
@@ -0,0 +1,145 @@
...
+// Store data identified by keys.
+// The function enqueues the request and returns immediately: the entry will be
+// asynchronously stored in the cache.
+func (c *MemcachedCache) Store(ctx context.Context, keys []string, bufs [][]byte) {
+	var (
+		firstErr error
+		failed   int
+	)
+
+	for i, key := range keys {
```
squat, line 107: Implicit here is that  `bufs` MUST be at least as long as `keys`, otherwise the program will panic. Do we want to check this early in the function and log an error?

squat: Or just document?

pkg/queryfrontend/cache/memcached.go
yeya24: This metric is for `items requests`, same as https://github.com/thanos-io/thanos/blob/master/pkg/cache/memcached.go#L35.

Yes, it is not very clear about the naming.

pkg/queryfrontend/cache/memcached.go
squat: Or maybe accept as a parameter a map of type `map[string][][]byte` instead?

pkg/queryfrontend/cache/memcached.go
yeya24: Ditto.

pkg/queryfrontend/cache/memcached.go
squat: yes, I suspected this was the case. It reads a little bit funny, like we changed our mind while writing the sentence, but not too serious.

pkg/queryfrontend/cache/memcached.go
yeya24: > Or maybe accept as a parameter a map of type map[string][][]byte instead?

I cannot since this is the cache interface from Cortex... And I need to implement it. https://github.com/cortexproject/cortex/blob/master/pkg/chunk/cache/cache.go#L21

> Implicit here is that bufs MUST be at least as long as keys, otherwise the program will panic. Do we want to check this early in the function and log an error?

Umm, I don't know if this case would happen. But I agree with you that it is good to check length first

squat: It's a public method on an exposed member, so it could be called by anyone; so I think it would make sense to check it in our func

pkg/cache/memcached.go
bwplotka: Also Having mock in production code is not the best idea. Let's either put it inside `_test` file or create another package.

pkg/queryfrontend/cache/memcached.go
bwplotka: :+1:

pkg/queryfrontend/cache/memcached.go
```diff
@@ -0,0 +1,145 @@
+// Copyright (c) The Thanos Authors.
+// Licensed under the Apache License 2.0.
+
+package cache
```
bwplotka, line 4: Do we need separate client? We kind of have the bucket cache with this already? :thinking:

yeya24: The bucket memcached cache already has `Store` and `Fetch` method, same name but different function signatures. So I cannot implement the Cortex interface on it.

yeya24:
I am already reusing the memcached client config, please check https://github.com/thanos-io/thanos/pull/3032/files#diff-8b5ab7fe39fd04754734fceb6a9966dfR38.

There is another validity field here because the Cortex cache interface `Store` method needs to specify a ttl.

pkg/queryfrontend/cache/memcached.go
yeya24: You are right! Thanks.

yeya24:
I updated this pr. PTAL again.

pkg/queryfrontend/cache/memcached.go
bwplotka: Hm, I am not sure about it - I think we can reuse them - let me check

bwplotka: As talked offline, we can propose proper interface change (:

kakkoyun: @bwplotka @yeya24 I'd love to contribute to this interface as well. I had some struggles while integrating with groupcache so please include me into the discussions as well.

yeya24: Current interface

``` go
// Thanos
type Cache interface {
	Store(ctx context.Context, data map[string][]byte, ttl time.Duration)
	Fetch(ctx context.Context, keys []string) map[string][]byte
}

// Cortex
type Cache interface {
	Store(ctx context.Context, key []string, buf [][]byte)
	Fetch(ctx context.Context, keys []string) (found []string, bufs [][]byte, missing []string)
	Stop()
}
```

After checking some Cortex cache code, I think a better interface might be below. What do you think? IMHO `ttl` is specific to different cache, so it is better to not put it in the interface.

``` go
type Cache interface {
	Store(ctx context.Context, data map[string][]byte)
	Fetch(ctx context.Context, keys []string) (results map[string][]byte, missing []string)
	Stop()
}

```

bwplotka: This looks good, cc @krasi-georgiev who might take it now

bwplotka (changes requested):
The goal is to reuse same clients Cortex have.

kakkoyun:
Closing for the sake of #3114 cc @yeya24 @krasi-georgiev

PR #3049, opened 2020-08-18
ui: Fix "too many redirects" when using query frontend

### LINKED ISSUE #3043: New Query UI doesn't work with query frontend

Thanos master.

**What happened**:

The new query UI doesn't work with query frontend.

Setup:

Prometheus + Sidecar + Query + Query frontend

```bash

## Querier
thanos query --http-address="localhost:9091" --store="localhost:10903"

## Query frontend
thanos query-frontend --query-frontend.downstream-url="http://localhost:9091" --query-range.response-cache-config-file=test-conf --query-range.max-retries-per-request=0
```

When clicking `New UI` in the Query frontend address `http://localhost:10902`, I got `ERR_TOO_MANY_REDIRECTS`.
Screenshot from 2020-08-17 16-03-30

New UI should work in the above setup. 

Btw I also tested Prometheus + Query frontend, and set downstreamURL to prometheus addr. The Prometheus new UI works well. So I am wondering that is this problem coming from the new UI itself, not query frontend?

Screenshot from 2020-08-17 16-06-07

prmsrswt:

## Changes
Fixes #3043 

When using the UI via `query-frontend` we were getting redirected from `/new/` to `/new/` itself, which causes an infinite loop.
We had a redirect from `/new` to `/new/` (notice the trailing slash) and I suspect this the reason why we are getting this problem. The Query Frontend passes all requests other than `query_range` directly to upstream but I think that it is removing the trailing slash somehow. So when we request `/new/` the query frontend requests `/new` from upstream and get a 302 Redirection from upstream for `/new/` resulting in an infinite loop.

## Verification
Tested locally.

### DIFF
--- a/pkg/ui/query.go
+++ b/pkg/ui/query.go
@@ -84,7 +84,7 @@ func (q *Query) Register(r *route.Router, ins extpromhttp.InstrumentationMiddlew
 	// and which breaks users with a --web.route-prefix that deviates from the path derived
 	// from the external URL.
 	r.Get("/new", func(w http.ResponseWriter, r *http.Request) {
-		http.Redirect(w, r, path.Join(GetWebPrefix(q.logger, q.externalPrefix, q.prefixHeader, r), "new")+"/", http.StatusFound)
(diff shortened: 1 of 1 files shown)


yeya24:
Thank you for the quick fix :100: @prmsrswt

PR #3062, opened 2020-08-21
Website: add Talend as adopter

jleloup:

## Changes

We are using Thanos as part of Talend cloud monitoring for a bit more than a year now including in production so it is time to declare it publicly 😃 

I would like to add Talend logo on Thanos website in the adopters list.


GiedriusS (commented):
Thank you for using Thanos and your contribution! However, the logo is quite large as you can see in the preview and goes out of the bounds of one box. Perhaps there is some rectangular version of this logo or just a smaller one?

jleloup:
Sadly there is no rectangular version of this logo though I can reduce its size to make it fit.

jleloup:
Allright so with some adjustments I think it is ready to go.
I initially didn't saw this Netlify preview: that's pretty neat 😃

squat (approved):
thanks a lot, @jleloup :))

Issue #3060, opened 2020-08-20
ci: make docs-check in CI takes forever sometimes.


bwplotka:
Usually what we seen it's the `liche` taking ages to check links. We might want to disable this step or try something else ):

bwplotka:
Help wanted!

bwplotka:
E.g: https://github.com/thanos-io/thanos/pull/3059

bwplotka:
Ok it's too annoying, I am disabling it for now ): See: https://github.com/thanos-io/thanos/issues/3060

bwplotka:
Let's try to update it first: https://github.com/thanos-io/thanos/pull/3070

PR #3067, opened 2020-08-25
Updated readme link to prepend tip to fix Page Not Found; Disabled Liche temporarily. 

mllu:
* [ ] I added CHANGELOG entry for this change. (not sure which section to put under `Fixed` or `Changed`?)
* [ ] Change is not relevant to the end user.

## Changes

trying to fix the incorrect readme links that showing `Page Not Found` which were changed by this PR https://github.com/thanos-io/thanos/pull/2802/files#diff-6b3df26d750e09528185b365bbd99a76R72-R77 as shown below

Screen Shot 2020-08-24 at 10 18 05 PM

## Verification

click links from github readme

PR #3070, opened 2020-08-25
Updated liche.

bwplotka:
Hopefully improves https://github.com/thanos-io/thanos/issues/3060


bwplotka:
Wow, so many links are broken now cc @thisisobate 

I think it's worth to spend some time in adjusting our script for creating links ):

thisisobate:
> Wow, so many links are broken now cc @thisisobate
> 
> I think it's worth to spend some time in adjusting our script for creating links ):

wow...would look into this now

bwplotka:
The problem is that probably our liche script should check if the links will work on master website, not on master :thinking:

thisisobate:
okay...looking into this

squat:
but was it at least faster?

bwplotka:
Yup

bwplotka:
Will be part of https://github.com/thanos-io/thanos/pull/3067/

PR #3072, opened 2020-08-25
added future work to versioning proposal

thisisobate:
Signed-off-by: thisisobate <obasiuche62@gmail.com>

## Changes
Added future work section in versioning proposal

## Verification
Works on local

### DIFF
--- a/docs/proposals/202005_version_documentation.md
+++ b/docs/proposals/202005_version_documentation.md
@@ -2,7 +2,7 @@
 title: 'Building A Versioning Plugin For Thanos'
 type: proposal
 menu: proposals
-status: approved
+status: complete
 owner: thisisobate
 Date: May 2020
 ---
--- a/docs/proposals/202005_version_documentation.md
+++ b/docs/proposals/202005_version_documentation.md
@@ -86,11 +86,11 @@ We want to add an additional `versioned` folder within the website's `tmp` direc
     |- tmp
         |- public
         |- docs-pre-processed
-        |- versioned
-            |- master
-            |- version 0.13.0
-            |- version 0.12.2
-            |- other-folder-for-other-releases
+            |- versioned
+                |- master
+                |- version 0.13.0
+                |- version 0.12.2
+                |- other-folder-for-other-releases
 ```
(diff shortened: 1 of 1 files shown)


thisisobate:
@squat @bwplotka 
PTAL: Some addition to the proposal docs

docs/proposals/202005_version_documentation.md
```diff
@@ -86,11 +86,11 @@ We want to add an additional `versioned` folder within the website's `tmp` direc
     |- tmp
         |- public
         |- docs-pre-processed
-        |- versioned
-            |- master
-            |- version 0.13.0
-            |- version 0.12.2
-            |- other-folder-for-other-releases
+            |- versioned
```
bwplotka, line 89: That's not true, right? (:

bwplotka (approved):
LGTM, just dir is bit different at the end, but maybe those details does not matter that much here (: 

Also can we mark status of this as `completed`? (:

docs/proposals/202005_version_documentation.md
thisisobate: True...it's just that I think it's best we identify where those versioned folders are located (which in this case, exists in docs-pre-processed)

Anyways...those details don't matter much

thisisobate:
Fixed! @bwplotka @squat

docs/proposals/202005_version_documentation.md
```diff
@@ -162,3 +162,7 @@ With a regular expression. Instead of the developer manually checking out the in
 ## Summary
 
 We understand [Cortex](https://github.com/cortexproject/cortex/pull/2349) is working on this as well so we are learning from their approach and knowledge.
+
+## Future Work
+
+We hope to rewrite the versioning plugin using golang. We care alot about code maintainability and because we use golang as our primary programming language, it will be easier for developers to contribute to this plugin and make it even better.
```
squat, line 168:
```suggestion
We hope to rewrite the versioning plugin using Golang. We care a lot about code maintainability and because we use Golang as our primary programming language, it will be easier for developers to contribute to this plugin and make it even better.
```

thisisobate: Ack

bwplotka (approved):
Thanks! LGTM :+1:

PR #3075, opened 2020-08-25
Store: Mask series data from the objstore based on tombstones

Harshitha1234:

## Changes
- Store masks the series data from objstore based on the user created tombstones
- Added tombstones reader in tombstone package

## TODO
* [ ] Write tests.

## Verification

Checked manually

### DIFF
--- a/cmd/thanos/tools_delete.go
+++ b/cmd/thanos/tools_delete.go
@@ -4,32 +4,18 @@
 package main
 
 import (
-	"context"
-	"encoding/json"
-	"io/ioutil"
-	"os"
-	"strconv"
-	"time"
-
-	"github.com/cespare/xxhash"
 	"github.com/go-kit/kit/log"
 	"github.com/oklog/run"
 	opentracing "github.com/opentracing/opentracing-go"
 	"github.com/prometheus/client_golang/prometheus"
 	"github.com/prometheus/prometheus/promql/parser"
-	"github.com/thanos-io/thanos/pkg/model"
-	"github.com/thanos-io/thanos/pkg/runutil"
-
-	"github.com/thanos-io/thanos/pkg/objstore"
-	"github.com/thanos-io/thanos/pkg/objstore/client"
 	kingpin "gopkg.in/alecthomas/kingpin.v2"
-)
 
-type Tombstone struct {
-	Matcher string `json:"matcher"`
-	MinTime int64  `json:"minTime"`
-	MaxTime int64  `json:"maxTime"`
-}
+	"github.com/thanos-io/thanos/pkg/model"
+	"github.com/thanos-io/thanos/pkg/objstore/client"
+	"github.com/thanos-io/thanos/pkg/runutil"
+	"github.com/thanos-io/thanos/pkg/tombstone"
+)
 
 func registerDelete(m map[string]setupFunc, app *kingpin.CmdClause, pre string) {
 	cmd := app.Command("delete", "Delete series command")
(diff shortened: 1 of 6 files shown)


Harshitha1234:
cc @bwplotka @metalmatze  :)

pkg/store/bucket.go
```diff
@@ -891,6 +926,13 @@ func (s *BucketStore) Series(req *storepb.SeriesRequest, srv storepb.Store_Serie
 
 	s.mtx.RLock()
 
+	var ts []tombstone.Tombstone
+	//Loading all tombstones from objstore
+	ts, err = tombstone.ReadTombstones(s.bkt, s.logger)
```
pracucci, line 931: Performance wise this could be impactful. What if the bucket store keeps a background process to sync tombstones and keeping them in memory, so that at query time all tombstones are already loaded in memory without blocking to read them?

cmd/thanos/tools_delete.go
```diff
@@ -0,0 +1,61 @@
...
+		bkt, err := client.NewBucket(logger, confContentYaml, reg, pre+" delete")
+		if err != nil {
+			return err
+		}
+		defer runutil.CloseWithLogOnErr(logger, bkt, "tools delete")
+
+		// Dummy actor to immediately kill the group after the run function returns.
+		g.Add(func() error { return nil }, func(error) {})
+
+		_, err = parser.ParseMetricSelector(*matchers)
```
bwplotka, line 52: ` if _, err := parse.Parse...; err != nil {` pattern available

pkg/store/bucket.go
bwplotka: Maybe we could use https://github.com/thanos-io/thanos/blob/c5dd131a422f58b110463a9bb5766eb654bcf3f4/pkg/store/cache/caching_bucket_factory.go#L67

pkg/store/bucket.go
```diff
@@ -716,6 +719,25 @@ func blockSeries(
 			return s.lset[i].Name < s.lset[j].Name
 		})
 
+		validTombstones := make([]tombstone.Tombstone, 0)
+		for _, ts := range tombstones {
+			tsMatchers, err := parser.ParseMetricSelector(ts.Matchers)
```
bwplotka, line 724: Would be nice avoid extra parsing during query time, instead we preprocess tombstones for quicker querying

pkg/store/bucket.go
```diff
@@ -716,6 +719,25 @@ func blockSeries(
 			return s.lset[i].Name < s.lset[j].Name
 		})
 
+		validTombstones := make([]tombstone.Tombstone, 0)
```
metalmatze, line 722: If we don't allocate a specific amount, then we can simply do 
```suggestion
		var validTombstones []tombstone.Tombstone
```

pkg/store/bucket.go
```diff
@@ -716,6 +719,25 @@ func blockSeries(
...
+		for _, ts := range tombstones {
+			tsMatchers, err := parser.ParseMetricSelector(ts.Matchers)
+			if err != nil {
+				return nil, nil, err
+			}
+			tsPostings, err := indexr.ExpandedPostings(tsMatchers)
```
bwplotka, line 728: While this is correct, there might be more optimal solution. 


Instead of doing matchers to postings lookup we could just match labels that we have vs matchers that tombstones specifies. Thanks to this we don't need fetch potentially thousands of unrelated postings during this lookup.

bwplotka: Let's say we query for `cluster=us1`


Deleted `cluster=~"us.*"`

so we here we have in `lset` cluster=us1 but still we  (maybe there are 100000 us cluster) would fetch 10000 of those postings here.

pkg/store/bucket.go
```diff
@@ -724,6 +746,19 @@ func blockSeries(
 				break
 			}
 
+			mask := false
+			for _, ts := range validTombstones {
+				if meta.MaxTime > ts.MinTime && meta.MaxTime <= ts.MaxTime {
```
bwplotka, line 751: We had some method somewhere for checking if intervals of tombstones overlaps or not, maybe we can use that.

Just make sure add TODO to rewrite chunk on partial overlap with tombstone.

pkg/store/bucket.go
```diff
@@ -891,6 +926,13 @@ func (s *BucketStore) Series(req *storepb.SeriesRequest, srv storepb.Store_Serie
 
 	s.mtx.RLock()
 
+	var ts []tombstone.Tombstone
+	//Loading all tombstones from objstore
```
bwplotka, line 930:
```suggestion
	// Loading all tombstones from objstore.
```

pkg/store/bucket.go
bwplotka: Can we encapsulate some of this in Tombstone methods? (:

pkg/tombstone/tombstone_test.go
```diff
@@ -0,0 +1,23 @@
...
+
+import (
+	"testing"
+
+	"github.com/go-kit/kit/log"
+
+	"github.com/thanos-io/thanos/pkg/objstore"
+	"github.com/thanos-io/thanos/pkg/testutil"
+)
+
```
bwplotka, line 14: I am missing some tests for deletions! (:

bwplotka (commented):
Amazing work, some suggestions though after our offline meeting :hugs:

pkg/store/bucket.go
pracucci: Yes, would be an improvement. However, tombstones look to be a very good candidate for async scanning (especially when there are a bunch of them).

Harshitha1234:
Still valid! Needs some changes and I'm working on it :)

PR #3076, opened 2020-08-25
feat: Allow callers to the minio objstore to inject a RoundTripper

jtlisi:
# Inject Roundtripper to Minio Bucket using Config

The reason for this change was to provide a bit more control over the transport level of this client. I noticed that the default Minio transport has diverged significantly from this implementation. Instead of exposing more values as HTTPConfigs or introducing changes to the default config, I thought it made more sense to give the package caller extra control over the client.

* [ ] I added CHANGELOG entry for this change. **Question:** There is no change from a user perspective, should I still add an entry?

## Changes

- Create a `DefaultTransport` function that generates an `*http.Transport` from the Config as it currently exists. This function was made public since it allows a caller to generate a Transport that respects the provided config but then tweak values within the transport.
- Add a `Transport http.Roundtripper` field to the `s3.Config` struct that will replace the default transport if set

## Verification

I am assuming no unexpected behavior due to the nature of the change as long as the existing test cases pass.

### DIFF
--- a/pkg/objstore/s3/s3.go
+++ b/pkg/objstore/s3/s3.go
@@ -72,6 +72,9 @@ type Config struct {
 	// PartSize used for multipart upload. Only used if uploaded object size is known and larger than configured PartSize.
 	PartSize  uint64    `yaml:"part_size"`
 	SSEConfig SSEConfig `yaml:"sse_config"`
+
+	// Allow upstream callers to inject a round tripper
+	Transport http.RoundTripper `yaml:"-"`
 }
 
 // SSEConfig deals with the configuration of SSE for Minio. The following options are valid:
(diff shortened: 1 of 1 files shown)


bwplotka:
> There is no change from a user perspective, should I still add an entry?

No (:

pkg/objstore/s3/s3.go
```diff
@@ -72,6 +72,9 @@ type Config struct {
 	// PartSize used for multipart upload. Only used if uploaded object size is known and larger than configured PartSize.
 	PartSize  uint64    `yaml:"part_size"`
 	SSEConfig SSEConfig `yaml:"sse_config"`
+
+	// Allow upstream callers to inject a round tripper
+	Transport http.RoundTripper `yaml:"-"`
```
bwplotka, line 77: Hm, I would either inject it via some Setter or at least in HTTPConfig as there is functional overlap with that.

bwplotka (commented):
Nice, but before injecting custom round tripper can we learn what's the reason? If the reason is our obsolete default transport, we would use some help in making it better :hugs:

bwplotka:
>  Instead of exposing more values as HTTPConfigs or introducing changes to the default config

If changes makes sense, then why not to default config? (:

pkg/objstore/s3/s3.go
jtlisi: That makes sense, I'll move it to the HTTPConfig

pkg/objstore/s3/s3.go
```diff
@@ -157,36 +189,20 @@ func NewBucketWithConfig(logger log.Logger, config Config, component string) (*B
...
-				KeepAlive: 30 * time.Second,
-				DualStack: true,
-			}).DialContext,
-
-			MaxIdleConns:          100,
-			MaxIdleConnsPerHost:   100,
-			IdleConnTimeout:       time.Duration(config.HTTPConfig.IdleConnTimeout),
-			TLSHandshakeTimeout:   10 * time.Second,
-			ExpectContinueTimeout: 1 * time.Second,
-			// The ResponseHeaderTimeout here is the only change
```
pracucci, line 177: Any reason why the refactoring lost this comment?

pracucci (approved):
LGTM (modulo my question and Bartek's feedback)

pkg/objstore/s3/s3.go
jtlisi: The whole comment itself is no longer true since this config and the default minio config have diverged significantly. However, I will re-add the section explaining why the ResponseHeaderTimeout is set.

squat (approved):
lgtm

bwplotka (changes requested):
Still not applied comments:

> Nice, but before injecting custom round tripper can we learn what's the reason? If the reason is our obsolete default transport, we would use some help in making it better hugs

(:

jtlisi:
> Still not applied comments:

Sorry about that, I didn't find a chance yesterday to update the PR. It should be good to go.

bwplotka (approved):
LGTM, Thanks!

PR #3085, opened 2020-08-27
Allow no trimming in the equal test util with an env var.

krasi-georgiev:


pkg/testutil/testutil.go
```diff
@@ -76,15 +77,19 @@ func Equals(tb testing.TB, exp, act interface{}, v ...interface{}) {
...
+	tb.Fatalf("\033[31m%s:%d:"+msg+"\n\n\texp: %#v\n\n\tgot: %#v%s\033[39m\n\n", filepath.Base(file), line, trim(exp), trim(act), trim(diff(exp, act)))
 }
 
-func sprintfWithLimit(act string, v ...interface{}) string {
-	s := fmt.Sprintf(act, v...)
-	if len(s) > 1000 {
-		return s[:1000] + "...(output trimmed)"
+func trim(s interface{}) string {
+	str := fmt.Sprintf("%v", s)
+	_, noTrim := os.LookupEnv("NO_TRIM")
```
bwplotka, line 85: What about just commenting out in code if you really need this. It's rare that trillions line of output is really useful - normally it just explodes your tests. :thinking:

bwplotka (changes requested):
Hm, I would just edit code if you need full diff, wdyt?

pkg/testutil/testutil.go
krasi-georgiev: I would rather have it this way to switch quickly between trimmed and non-trimmed, otherwise if I change the code when needed will forget and add to a PR by mistake.

Do you see any problem with this approach?

krasi-georgiev: ping @bwplotka

bwplotka: I would say the hidden envars like this tend to be 100% forgotten. I would say, let's increase size a little bit and work towards automatic printing only things that matters instead. If you put this accidently in a PR, we can just find it in code review (:

krasi-georgiev: ok

PR #3088, opened 2020-08-27
Refactored cmd invokations for readability & reuse.

bwplotka:

### DIFF
--- a/pkg/extflag/pathorcontent.go
+++ b/pkg/extflag/pathorcontent.go
@@ -21,12 +21,12 @@ type PathOrContent struct {
 	content *string
 }
 
-type CmdClause interface {
+type FlagClause interface {
 	Flag(name, help string) *kingpin.FlagClause
 }
 
 // RegisterPathOrContent registers PathOrContent flag in kingpinCmdClause.
-func RegisterPathOrContent(cmd CmdClause, flagName string, help string, required bool) *PathOrContent {
+func RegisterPathOrContent(cmd FlagClause, flagName string, help string, required bool) *PathOrContent {
 	fileFlagName := fmt.Sprintf("%s-file", flagName)
 	contentFlagName := flagName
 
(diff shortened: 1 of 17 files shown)


pkg/extflag/pathorcontent.go
```diff
@@ -21,12 +21,12 @@ type PathOrContent struct {
 	content *string
 }
 
-type CmdClause interface {
+type FlagClause interface {
```
prmsrswt, line 24: Just a doubt :sweat_smile: Isn't this same as `extkingpin.FlagClause`? If yes then why aren't we using `extkinpin.FlagClause` instead of declaring it here again?

prmsrswt (approved):
LGTM :fire:

pkg/extflag/pathorcontent.go
bwplotka: `doubt` you mean `question`? (:+1: 

The reason is that `FlagClause` on kingpin is a specific implementation and we need just a lightweight interface.

prmsrswt: Yes, I meant a question.

I was talking about this interface defined in `pkg/extkingpin/app.go` on Line 14-16 .
```go

type FlagClause interface {
	Flag(name, help string) *kingpin.FlagClause
}
```
I think they are the same, so how is one lightweight?

squat: Yes, we declare this interface again in thanos/pkg/extkingpin. Can we solve this with just one declaration?

bwplotka: ack

bwplotka: Ok, I would just leave it. We use this package a lot in other project - would be nice to not require extra packages. Currently extkingpin requires a lot of extra stuff. I would just copy it (as per Go: little copying is better than importing).

Issue #2861, opened 2020-07-08
Store: limit queried chunks by bytes


pracucci:
In the PR https://github.com/thanos-io/thanos/pull/2858 we've introduced `ChunksLimiter` in the Thanos store to limit the number of chunks fetched by a single `Series()` call.

It could make sense to add a limit based on "chunks bytes" instead of "chunks count" (on in addition to "chunks count" if Cortex will need to support it) as a easier unit to work when doing capacity planning and fine tuning the Thanos store.

pracucci:
_Still valid, some help wanted (should be easy to do)._

mneverov:
Hi, I'd like to work on this.

@pracucci should a new limiter be added or should it replace the existing one?

pracucci:
> @pracucci should a new limiter be added or should it replace the existing one?

I would be glad if we could keep support for limiting by number of chunks as well (at least in the code) because we're using it in Cortex.

pracucci:
Still valid. The PR #3089 addresses it.

PR #3089, opened 2020-08-27
Limit queried chunks by bytes

mneverov:
Fixes: #2861

Signed-off-by: Max Neverov <neverov.max@gmail.com>

## Changes

Add possibility to limit queried chunks by bytes via `store.grpc.series-sample-size-limit` flag.

### DIFF
--- a/cmd/thanos/store.go
+++ b/cmd/thanos/store.go
@@ -69,7 +69,7 @@ func registerStore(app *extkingpin.App) {
 		Default("0").Uint()
 
 	maxSampleSize := cmd.Flag("store.grpc.series-sample-size-limit",
-		"Maximum size of samples returned via a single Series call. The Series call fails if this limit is exceeded. 0 means no limit.").
+		"Maximum size of samples returned (in bytes) via a single Series call. The Series call fails if this limit is exceeded. 0 means no limit.").
 		Default("0").Uint()
 
 	maxConcurrent := cmd.Flag("store.grpc.series-max-concurrency", "Maximum number of concurrent Series calls.").Default("20").Int()
@@ -297,7 +297,7 @@ func runStore(
 		store.WithChunksSizeLimit(store.NewChunksLimiterFactory(maxSampleSize)),
 	}
 
-	bs, err := store.New(
+	bs, err := store.NewBucketStoreWithOptions(
 		logger,
 		reg,
 		bkt,
--- a/cmd/thanos/store.go
+++ b/cmd/thanos/store.go
@@ -68,6 +68,10 @@ func registerStore(app *extkingpin.App) {
 		"Maximum amount of samples returned via a single Series call. The Series call fails if this limit is exceeded. 0 means no limit. NOTE: For efficiency the limit is internally implemented as 'chunks limit' considering each chunk contains 120 samples (it's the max number of samples each chunk can contain), so the actual number of samples might be lower, even though the maximum could be hit.").
 		Default("0").Uint()
 
+	maxSampleSize := cmd.Flag("store.grpc.series-sample-size-limit",
+		"Maximum size of samples returned via a single Series call. The Series call fails if this limit is exceeded. 0 means no limit.").
+		Default("0").Uint()
+
 	maxConcurrent := cmd.Flag("store.grpc.series-max-concurrency", "Maximum number of concurrent Series calls.").Default("20").Int()
 
 	objStoreConfig := extkingpin.RegisterCommonObjStoreFlags(cmd, "", true)
@@ -133,6 +137,7 @@ func registerStore(app *extkingpin.App) {
 			uint64(*indexCacheSize),
 			uint64(*chunkPoolSize),
 			uint64(*maxSampleCount),
+			uint64(*maxSampleSize),
 			*maxConcurrent,
 			component.Store,
 			debugLogging,
@@ -169,7 +174,7 @@ func runStore(
 	grpcGracePeriod time.Duration,
 	grpcCert, grpcKey, grpcClientCA, httpBindAddr string,
 	httpGracePeriod time.Duration,
-	indexCacheSizeBytes, chunkPoolSizeBytes, maxSampleCount uint64,
+	indexCacheSizeBytes, chunkPoolSizeBytes, maxSampleCount, maxSampleSize uint64,
 	maxConcurrency int,
 	component component.Component,
 	verbose bool,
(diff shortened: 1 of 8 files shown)


pkg/store/bucket.go
```diff
@@ -769,6 +774,10 @@ func blockSeries(
 		return nil, nil, errors.Wrap(err, "preload chunks")
 	}
 
+	if err := chunksSizeLimiter.Reserve(uint64(chunkr.stats.seriesFetchedSizeSum)); err != nil {
```
bwplotka, line 777: As per your question:

> Another question is is it a proper way to limit based on a chunk reader stats? https://github.com/thanos-io/thanos/pull/3089/files#diff-a75f50a9f5bf5b21a862e4e7c6bd1576R777
Or is it better to pass the limiter to the preload function?

That's a good one, it really depends on one thing: Do we want to return partial data if limit is exceeded or fail everything? (:

mneverov: My take on that is that a user might be surprised when find out that the returned response was partial. 
It might make sense to create a config for that and apply for both: limiting by a number of samples and by byte size.

bwplotka: Oh yes, we have warnings for that. Everything with warning is assumed partial - then querier decide if it's error or not with partialResponseStrategy request option (:

mneverov:
To the reviewer:
A [new bucket store constructor](https://github.com/thanos-io/thanos/pull/3089/files#diff-a75f50a9f5bf5b21a862e4e7c6bd1576R374) name suggestions are greatly appreciated.
I reused a [chunk reader stats ](https://github.com/thanos-io/thanos/pull/3089/files#diff-a75f50a9f5bf5b21a862e4e7c6bd1576R849) instead of passing the limiter to the preload function because it was refactored out in #2858 
I [reused the same metric](https://github.com/thanos-io/thanos/pull/3089/files#diff-a75f50a9f5bf5b21a862e4e7c6bd1576R981) for both chunks and chunks size limits. It is possible that this metric will be updated twice for the same query. What's better: to use different metric or to introduce a global lock to update it?

pkg/store/bucket.go
```diff
@@ -305,6 +345,49 @@ func NewBucketStore(
...
+		enablePostingsCompression,
+		postingOffsetsInMemSampling,
+		enableSeriesResponseHints,
+		bucketStoreOpts...,
+	)
+}
+
+// New creates a new bucket backed store that implements the store API against
+// an object store bucket. It is optimized to work against high latency backends.
+func New(
```
OGKevin, line 374: I would then just rename this to newBucketStore if there is no need to have it exported as its only used in this package.

OGKevin: Ah I see it now. Then I would suggest `NewBucketStoreWithOptions` and set a deprecation warning on the old method. Then slowly refactor. Once that is done, rename back to `NewBucketStore`.

mneverov: thanks!

cmd/thanos/store.go
```diff
@@ -68,6 +68,10 @@ func registerStore(app *extkingpin.App) {
 		"Maximum amount of samples returned via a single Series call. The Series call fails if this limit is exceeded. 0 means no limit. NOTE: For efficiency the limit is internally implemented as 'chunks limit' considering each chunk contains 120 samples (it's the max number of samples each chunk can contain), so the actual number of samples might be lower, even though the maximum could be hit.").
 		Default("0").Uint()
 
+	maxSampleSize := cmd.Flag("store.grpc.series-sample-size-limit",
+		"Maximum size of samples returned via a single Series call. The Series call fails if this limit is exceeded. 0 means no limit.").
```
pracucci, line 72: I would clarify what's the unit here. Bytes?

pkg/store/bucket.go
```diff
@@ -876,16 +971,22 @@ func (s *BucketStore) Series(req *storepb.SeriesRequest, srv storepb.Store_Serie
...
+		res               []storepb.SeriesSet
+		mtx               sync.Mutex
+		g, gctx           = errgroup.WithContext(ctx)
+		resHints          = &hintspb.SeriesResponseHints{}
+		reqBlockMatchers  []*labels.Matcher
+		chunksLimiter     = s.chunksLimiterFactory(s.metrics.queriesDropped)
+		chunksSizeLimiter = s.chunksSizeLimiterFactory(s.metrics.queriesDropped)
 	)
 
+	chunksSizeLimiter, err = chunksSizeLimiter.NewWithFailedCounterFrom(chunksLimiter)
```
pracucci, line 985: Could you clarify why you need `NewWithFailedCounterFrom()`?

mneverov: @pracucci thanks for reviewing this.
Chunks limiter and chunks size limiter should have common `failedOnce` to avoid concurrent update of the `failedCounter`.

mneverov: more info on that: `store` does not know at the moment of `ChunksLimit` and `ChunksSizeLimit` creation what metrics will be used. Metrics are the part of the `BucketStore`. That is why store operates limiter factories and bucket creates new chunk size limiter with the sync.Once shared between limiters. Hope that helps

kakkoyun:
Hey @mneverov, shall we reopen this? Do you plan to work on it?

kakkoyun:
@mneverov 👍 Reopening it and passing it to @bwplotka @pracucci. We can either merge if it's good to go or close it again.

pkg/store/bucket.go
pracucci: I think `NewWithFailedCounterFrom()` is a bit overengineered.

There's no problem updating a metric concurrently, but we may want to distinguish the reason why a query was dropped. An option may be adding a "reason" label to `queriesDropped` and passing `s.metrics.queriesDropped.WithLabelValues("<reason>")` to both factories.

pkg/store/bucket.go
```diff
@@ -305,6 +346,49 @@ func NewBucketStore(
...
+		enablePostingsCompression,
+		postingOffsetsInMemSampling,
+		enableSeriesResponseHints,
+		bucketStoreOpts...,
+	)
+}
+
+// NewBucketStoreWithOptions creates a new bucket backed store that implements the store API against
+// an object store bucket. It is optimized to work against high latency backends.
+func NewBucketStoreWithOptions(
```
pracucci, line 375: @bwplotka What do you think about this? I've mixed feelings about it, just to avoid adding another argument to `NewBucketStore()`. Maybe we can rollback options in this PR and follow up the discussion about functional options in https://github.com/thanos-io/thanos/pull/3931?

pracucci, line 375: @bwplotka What do you think about this? I've mixed feelings about it, just to avoid adding another argument to `NewBucketStore()`. Maybe we can rollback options in this PR and follow up the discussion about functional options in https://github.com/thanos-io/thanos/pull/3931?

pracucci (commented):
Sorry for the extreme low review. Overall changes LGTM, but I left a couple of comments.

mneverov:
hi @pracucci ,
thanks for the review! All comments are valid.
It's been a while since the PR was submitted and I stuck in rebasing master branch. I think it is easier to close this PR and reimplement the functionality on top of updated bucket.go. 
Sorry you wasted your time on this, I should have closed it long time ago.

PR #3094, opened 2020-08-28
made docsearch attribute reflect version of page

thisisobate:
Signed-off-by: thisisobate <obasiuche62@gmail.com>

## Changes
Made docsearch attribute reflect the version of the page

## Verification
Works on local

### DIFF
--- a/website/layouts/_default/baseof.html
+++ b/website/layouts/_default/baseof.html
@@ -2,9 +2,10 @@
 <html lang="en">
 
 <head>
+    {{ $version     := index (split .File.Path "/") 0 }}
     <meta http-equiv="content-type" content="text/html;charset=utf-8">
     <meta name="docsearch:language" content="en" />
-    <meta name="docsearch:version" content="1.0.0" />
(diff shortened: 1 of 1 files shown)


website/layouts/_default/baseof.html
```diff
@@ -2,9 +2,10 @@
 <html lang="en">
 
 <head>
+    {{ $version     := index (split .File.Path "/") 0 }}
```
squat, line 5: Tiny nit: why do we need so many spaces here between the `$version` and the `:=`?

squat (approved):
Overall this change makes sense. One question, will this result in meta-tags with values like `release-0.2` or `v0.2`?

website/layouts/_default/baseof.html
thisisobate: oh...my bad
would fix up

thisisobate:
it will result in meta tags with values v0.2

squat (approved):

squat:
cool, just tested this and it works perfectly in the preview :)

thisisobate:
Awesome!

PR #3098, opened 2020-08-28
ui: Fix Block Viewer for Compactor and Store

### LINKED ISSUE #3093: v0.15.0-rc.0 Compactor Block Viewer: Does not work
Looks like compactor `global` view does not forward API and css etc calls to correct endpoint: 

image

image

Even links point to just path without actual endpoint: 
image

cc @prmsrswt

prmsrswt:

Fixes #3093

## Changes
- Modified the Blocks API to serve loaded blocks and global blocks
- Fix Block Viewer not working because of incorrect handling of the external prefix.

## Verification
Tested locally for Bucket web, Compactor, and Store using docker-compose setup.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -14,6 +14,7 @@ We use *breaking :warning:* word for marking changes that are not backward compa
 ### Fixed
 
 * [#3095](https://github.com/thanos-io/thanos/pull/3095) Rule: update manager when all rule files are removed.
+* [#3098](https://github.com/thanos-io/thanos/pull/3098) ui: Fix Block Viewer for Compactor and Store
 
 ## [v0.15.0-rc.0](https://github.com/thanos-io/thanos/releases/tag/v0.15.0-rc.0) - 2020.08.26
 
--- a/pkg/api/blocks/v1.go
+++ b/pkg/api/blocks/v1.go
@@ -18,9 +18,10 @@ import (
 
 // BlocksAPI is a very simple API used by Thanos Block Viewer.
 type BlocksAPI struct {
-	baseAPI    *api.BaseAPI
-	logger     log.Logger
-	blocksInfo *BlocksInfo
+	baseAPI          *api.BaseAPI
+	logger           log.Logger
+	globalBlocksInfo *BlocksInfo
+	loadedBlocksInfo *BlocksInfo
 }
 
 type BlocksInfo struct {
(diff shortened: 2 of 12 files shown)


CHANGELOG.md
```diff
@@ -14,6 +14,7 @@ We use *breaking :warning:* word for marking changes that are not backward compa
 ### Fixed
 
 * [#3095](https://github.com/thanos-io/thanos/pull/3095) Rule: update manager when all rule files are removed.
+* [#3098](https://github.com/thanos-io/thanos/pull/3098) ui: Fix Block Viewer for Compactor and Store
```
bwplotka, line 17: No need for this, as it's fixing the release directly. But that's ok I will adjust it.

pkg/api/blocks/v1.go
```diff
@@ -51,19 +56,32 @@ func (bapi *BlocksAPI) Register(r *route.Router, tracer opentracing.Tracer, logg
 }
 
 func (bapi *BlocksAPI) blocks(r *http.Request) (interface{}, []error, *api.ApiError) {
-	return bapi.blocksInfo, nil, nil
+	viewParam := r.URL.Query().Get("view")
+	if viewParam == "loaded" {
+		return bapi.loadedBlocksInfo, nil, nil
+	}
+	return bapi.globalBlocksInfo, nil, nil
```
bwplotka, line 63: This feels weird from perspective on components that does not have global :thinking: 

Also I think it's on purpose that in code we talk about `global` view and in React / UI we have `blocks`?

bwplotka (approved):
Looks good only some few soft comments. Thanks for super quick fix!

pkg/api/blocks/v1.go
prmsrswt: I agree it is weird. But in our last 1:1, me, @squat, and @GiedriusS discussed this and decided to go with this. As we have the same API but different needs in different components (Bucket viewer: a global view of blocks, Compactor: both global and loaded and Store: loaded only) we came up with two solutions:

- Modify the API logic so that we can register a single type of API multiple times in a single component with different prefixes. This means that in the Compactor, we can create two instances of Blocks API on two different prefixes (let's say `/api/v1/loaded` and `/api/v1/global`) much like how we handled the old UI.
- Extend the blocks API to have two types of data (for Global and Loaded) always and add a query parameter in the existing `/api/v1/blocks` endpoint.

The first solution is more generic but it requires us to modify logic for all of the API packages. The second solution was simpler to implement and only affected the Blocks API.

CHANGELOG.md
prmsrswt: Thanks!

PR #3100, opened 2020-08-28
Fixed tip links.

bwplotka:

This does not fix everything, for that we need proper script (in making). This should fix basics at least (tip).


bwplotka:
Merging to fix latest docs.

cc @thisisobate

PR #3101, opened 2020-08-31
Improve doc about exposing UI sub-path behind reverse proxy with nginx ingress

shay-berman:
…

Signed-off-by: Shay Berman <shay.berman@redislabs.com>

query

## Changes

Adding clarification of how to set dynamic path configuration with Ingress nginx inside the section `Expose UI on a sub-path` [link](https://github.com/thanos-io/thanos/blob/master/docs/components/query.md#expose-ui-on-a-sub-path).
Currently the detail describe on this section is more focus on Trafix and less on nginx ingress.
So I just added an example of Igress (it was hard to find this info, and took me quite time to understand how to do it, so why not to help everyone with simple example).

## Verification

I used the ingress object in the example and setting of the thanos query as mentioned, then I access to `http://MY-HOST-FQDN/thanos` and it works well.

### DIFF
--- a/docs/components/query.md
+++ b/docs/components/query.md
@@ -260,6 +260,30 @@ If `PathPrefixStrip: /some-path` option or `traefik.frontend.rule.type: PathPref
 Kubernetes Ingress annotation is set, then `Traefik` writes the stripped prefix into X-Forwarded-Prefix header.
 Then, `thanos query --web.prefix-header=X-Forwarded-Prefix` will serve correct HTTP redirects and links prefixed by the stripped path.
 
+For ingress controller nginx, use the flag `--web.prefix-header=X-Forwarded-Prefix` for thanos query
+and use the below example of Ingress configuration to access Thanos on `/thanos` prefix:
+```
+apiVersion: extensions/v1beta1
+kind: Ingress
+metadata:
+  annotations:
+    kubernetes.io/ingress.class: nginx
+    nginx.ingress.kubernetes.io/rewrite-target: /$1
+    nginx.ingress.kubernetes.io/x-forwarded-prefix: /thanos
+  labels:
+    app.kubernetes.io/name: thanos-query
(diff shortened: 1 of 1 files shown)


docs/components/query.md
```diff
@@ -260,6 +260,30 @@ If `PathPrefixStrip: /some-path` option or `traefik.frontend.rule.type: PathPref
 Kubernetes Ingress annotation is set, then `Traefik` writes the stripped prefix into X-Forwarded-Prefix header.
 Then, `thanos query --web.prefix-header=X-Forwarded-Prefix` will serve correct HTTP redirects and links prefixed by the stripped path.
 
+For ingress controller nginx, use the flag `--web.prefix-header=X-Forwarded-Prefix` for thanos query
```
povilasv, line 263: Which ingress controller are we talking about?
https://github.com/kubernetes/ingress-nginx
or 
https://www.nginx.com/products/nginx/kubernetes-ingress-controller/

?

Just use the correct name ;)

docs/components/query.md
```diff
@@ -260,6 +260,30 @@ If `PathPrefixStrip: /some-path` option or `traefik.frontend.rule.type: PathPref
...
+and use the below example of Ingress configuration to access Thanos on `/thanos` prefix:
+```
+apiVersion: extensions/v1beta1
```
povilasv, line 266: this is deprecated version, please use the stable one.

docs/components/query.md
povilasv, line 263: Let's also add a subheading and talk more about kubernetes and the example.

prmsrswt:
We also have a separate page in the `Operating` section of our documentation (https://thanos.io/tip/operating/reverse-proxy.md) for running Thanos behind a reverse proxy. I think it'd be better if we can update the examples there and then link it in the Querier's documentation.

PR #3108, opened 2020-09-01
pkg/reloader: remove mention of rule directories

simonpasquier:

## Changes

The `reloader` package doesn't care whether the watched directories
contain rule files or something else. This change removes the references
to "rule directories" and uses "watched directories" instead.

The motivation is that prometheus-operator use 2 reloaders:
* [prometheus-config-reloader](https://quay.io/repository/prometheus-operator/prometheus-config-reloader) that watches the Prometheus configuration file, it leverages the `github.com/thanos-io/thanos/pkg/reloader` package
* [configmap-reload](https://hub.docker.com/r/jimmidyson/configmap-reload) that watches Prometheus rule files and Alertmanager configuration files, it is a separate project and implements its own logic to detect file changes.

Eventually we may consolidate and have only one reloader instance (presumably prometheus-config-reloader). Having to provide a `RulesDir` option to reload Alertmanager is slightly inconsistent.

## Verification

No unit or e2e test added since there's no functional change.

### DIFF
--- a/pkg/reloader/reloader.go
+++ b/pkg/reloader/reloader.go
@@ -9,9 +9,9 @@
 // 	* Watch on changes against certain file e.g (`cfgFile`).
 // 	* Optionally, specify different different output file for watched `cfgFile` (`cfgOutputFile`).
 // 	This will also try decompress the `cfgFile` if needed and substitute ALL the envvars using Kubernetes substitution format: (`$(var)`)
-// 	* Watch on changes against certain directories (`ruleDires`).
+// 	* Watch on changes against certain directories (`watchedDirs`).
 //
-// Once any of those two changes Prometheus on given `reloadURL` will be notified, causing Prometheus to reload configuration and rules.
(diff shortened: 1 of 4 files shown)


pkg/reloader/reloader.go
```diff
@@ -220,10 +220,10 @@ func (r *Reloader) Watch(ctx context.Context) error {
 
 	r.watches.Set(float64(len(watchables)))
 	level.Info(r.logger).Log(
-		"msg", "started watching config file and recursively rule dirs for changes",
+		"msg", "started watching config file and recursively dirs for changes",
```
squat, line 223: nit: 
```suggestion
		"msg", "started watching config file and directories recursively for changes",
```

squat (approved):
lgtm :) one small note

pkg/reloader/reloader.go
simonpasquier: Done. I've removed `recursively` because the reloader doesn't watch directories recursively.

squat (approved):
lgtm
re-running flaky test

PR #3109, opened 2020-09-01
Control response cache behavior based on query request

yeya24:

## Changes

1. Update Cortex to master https://github.com/cortexproject/cortex/commit/255ff33069609358d77de1d81f6e91aa19c129ac
2. Add `shouldCache` function to check whether go to cache or not. If a request has store matches, then we shouldn't get results from cache.
3. We can also control cache behavior based on the cache control header. 

One thing to note is that in https://github.com/cortexproject/cortex/pull/3051, Cortex uses a third-party fork for gomemcache.

## Verification

Unit test added.


povilasv (approved):
This LGTM, but we have a failing CI

kakkoyun (approved):
lgtm

PR #3110, opened 2020-09-01
Fix flaky query frontend split interval unit test

### LINKED ISSUE #3084: Flaky QueryFrontend Unit Test
```
--- FAIL: TestRoundTripSplitIntervalMiddleware (0.00s)
    --- FAIL: TestRoundTripSplitIntervalMiddleware/split_to_2_requests (0.00s)
        roundtrip_test.go:269: roundtrip_test.go:269:
            
            	exp: 2
            
            	got: 1
            
            
FAIL
FAIL	github.com/thanos-io/thanos/pkg/queryfrontend	0.037s
```

yeya24:

## Changes

Fix #3084 

In the previous test case, I used `counter` to mock HTTP handler and check how many times the handler is being called.

``` go
func counter() (*int, http.Handler) {
	count := 0
	var lock sync.Mutex
	return &count, http.HandlerFunc(func(w http.ResponseWriter, r *http.Request) {
		lock.Lock()
		defer lock.Unlock()
		count++
	})
}
```

But this mock handler doesn't return a meaningful result to query frontend so it might cause other requests canceled at the same time, which causes the flaky.  https://github.com/cortexproject/cortex/blob/master/pkg/querier/queryrange/limits.go#L104

So if we return promql response which is valid for query frontend, then the test case should work.

## Verification

Used the function below to test it works as expected.

``` go
func TestFlaky(t *testing.T) {
	testRequest := &ThanosRequest{
		Path:  "/api/v1/query_range",
		Start: 0,
		End:   2 * hour,
		Step:  10 * seconds,
	}

	codec := NewThanosCodec(true)
	tpw, err := NewTripperWare(&fakeLimits{}, nil, codec, nil,
		time.Hour, 0, nil, log.NewNopLogger())
	testutil.Ok(t, err)

	rt, err := newFakeRoundTripper()
	testutil.Ok(t, err)
	defer rt.Close()

	for i := 0; i < 10000; i++ {
		c, handler := promqlResults(false)
		rt.setHandler(handler)

		ctx := user.InjectOrgID(context.Background(), "1")
		httpReq, err := codec.EncodeRequest(ctx, testRequest)
		testutil.Ok(t, err)

		_, err = tpw(rt).RoundTrip(httpReq)
		testutil.Ok(t, err)

		testutil.Equals(t, 2, *c)
	}
}
```

### DIFF
--- a/pkg/queryfrontend/roundtrip_test.go
+++ b/pkg/queryfrontend/roundtrip_test.go
@@ -86,12 +86,11 @@ func TestRoundTripRetryMiddleware(t *testing.T) {
 	codec := NewThanosCodec(true)
 
 	for _, tc := range []struct {
-		name             string
-		maxRetries       int
-		req              queryrange.Request
-		handlerAndResult func() (*int, http.Handler)
-		expectedError    bool
-		expected         int
+		name       string
+		maxRetries int
+		req        queryrange.Request
+		fail       bool
+		expected   int
 	}{
 		{
 			name:       "not query range, retry won't be triggered 1.",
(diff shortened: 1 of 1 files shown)


kakkoyun (approved):
lgtm. Good work, thanks for jumping on it.

bwplotka (approved):

bwplotka:
Thanks!

PR #3114, opened 2020-09-02
Use more of the cortex frontend module to deduplicate some code and add the memcached backend

krasi-georgiev:
taking over for Bens PR :https://github.com/thanos-io/thanos/pull/3032

Signed-off-by: Krasi Georgiev <8903888+krasi-georgiev@users.noreply.github.com>


cmd/thanos/query-frontend.go
```diff
@@ -111,51 +91,40 @@ func runQueryFrontend(
 	logger log.Logger,
 	reg *prometheus.Registry,
 	tracer opentracing.Tracer,
-	conf *queryFrontendConfig,
+	cfg *config,
 	comp component.Component,
 ) error {
 
-	if len(conf.downstreamURL) == 0 {
-		return errors.New("downstream URL should be configured")
```
yeya24, line 119: This is still needed. If no downstreamURL is set, the query frontend will switch to the queue mode, which is not supported by Thanos.

cmd/thanos/query-frontend.go
```diff
@@ -16,93 +16,73 @@ import (
...
-	queryRangeConfig queryRangeConfig
-
-	downstreamURL        string
-	compressResponses    bool
-	LogQueriesLongerThan time.Duration
+type config struct {
+	http httpConfig
+	queryfrontend.Config
 
+	contentOrURL           extflag.PathOrContent
```
yeya24, line 39: Maybe use `responseCache` here? It is not `URL`

cmd/thanos/query-frontend.go
```diff
@@ -111,51 +91,40 @@ func runQueryFrontend(
...
 	defer fe.Close()
 
-	limits := queryfrontend.NewLimits(
-		conf.queryRangeConfig.maxQueryParallelism,
-		time.Duration(conf.queryRangeConfig.maxQueryLength),
-		conf.queryRangeConfig.cacheMaxFreshness,
-	)
-
-	respCacheContentYaml, err := conf.queryRangeConfig.respCacheConfig.Content()
+	limits, err := validation.NewOverrides(cfg.Limits, nil)
```
yeya24, line 117: It is not good to use Cortex's limits since it is too complex and you will find it adds more fields https://github.com/thanos-io/thanos/pull/3114/files#diff-d06787af93a387061cc1d2fe8067558eR87-R122 in the config file. 

Would you mind using our own? There are only 3 interface so not much code for us to maintain.

docs/components/query-frontend.md
```diff
@@ -45,21 +45,135 @@ Query Frontend supports a retry mechanism to retry query when HTTP requests are
...
+    max_metadata_per_metric: 0
+    max_global_metadata_per_user: 0
+    max_global_metadata_per_metric: 0
+    max_chunks_per_query: 0
+    max_query_length: 0s
+    max_query_parallelism: 0
+    cardinality_limit: 0
+    max_cache_freshness: 0s
+    per_tenant_override_config: ""
+    per_tenant_override_period: 0s
```
yeya24, line 122: The limits config is too complex, can you use our own limit?

docs/components/query-frontend.md
```diff
@@ -45,21 +45,135 @@ Query Frontend supports a retry mechanism to retry query when HTTP requests are
...
+          max_size_items: 0
+          validity: 0s
+          size: 0
+        prefix: ""
+      max_freshness: 0s
+    cache_results: false
+    max_retries: 0
+    parallelise_shardable_queries: false
+  frontend:
+    max_outstanding_per_tenant: 0
```
yeya24, line 168: The bad thing of using Cortex's config is that we will see many unnecessary fields in the config. Like `max_outstanding_per_tenant` and `parallelise_shardable_queries` both not used in Thanos.

krasi-georgiev:
> Though this pr is still a draft, I just take an initial look. Hope you don't mind.
> 
> We are reusing Cortex config directly here. My idea is:
> 
> We can reuse the Cortex cache config. But it is not very good to reuse its `Limits` and `QueryRange` config, they contain some unnecessary fields for Thanos so what about using our own? It wouldn't be too much code to maintain, just some structs/type conversion between Cortex and Thanos.

Yep I removed the Limits we can stil use the constructor though.

cmd/thanos/query-frontend.go
krasi-georgiev: yep, renamed to contentOrPath

cmd/thanos/query-frontend.go
krasi-georgiev: thanks re-added

cmd/thanos/query-frontend.go
krasi-georgiev: Yep I thought the same thing but we can still reuse the `validation.NewOverrides` no?

docs/components/query-frontend.md
```diff
@@ -100,32 +147,31 @@ Flags:
...
                               Listen host:port for HTTP endpoints.
       --http-grace-period=2m  Time to wait after an interrupt received for HTTP
                               Server.
       --query-frontend.downstream-url="http://localhost:9090"
                               URL of downstream Prometheus Query compatible API.
-      --query-frontend.compress-responses
+      --query-frontend.compress-http-responses
                               Compress HTTP responses.
-      --query-frontend.log_queries_longer_than=0
+      --query-frontend.log-queries-longer-than=0
```
yeya24, line 174: Thanks for this! Didn't notice that I used `_` before.

docs/components/query-frontend.md
```diff
@@ -89,7 +136,7 @@ Flags:
                               priority). Content of YAML file with tracing
                               configuration. See format details:
                               https://thanos.io/tip/tracing.md/#configuration
-      --query-range.split-interval=24h
+      --query-range.split-queries-by-interval=24h
```
yeya24, line 139: `split-interval` is fine IMHO, but not sure.

If we want to split other endpoints queries in the future (labels, series, etc), do we need separate split intervals for different endpoints?

pkg/queryfrontend/roundtrip.go
```diff
@@ -22,13 +21,15 @@ const (
 	labelQueryRange = "query_range"
 )
 
+// NewTripperware returns a Tripperware configured with middlewares to
+// limit, align, split, retry and cache requests.
+// Not using the cortex one as it uses  query parallelisations based on
+// storage sharding configuration and query ASTs.
```
yeya24, line 27: Perfect! Would you mind putting `cache requests` before `retry`? Because cache middleware is before retry. But just a nit.

cmd/thanos/query-frontend.go
yeya24: I think we can. If we need to extend the limits interface in the future, we can easily switch to our own.

docs/components/query-frontend.md
krasi-georgiev: I don't want to diverge too much from the cortex configs as they also have the Validate function which is nice. Maybe these changes can be done in the cortex module so that the config is stripped down a bit.

docs/components/query-frontend.md
krasi-georgiev: again I want this to be the same the flags upstream

docs/components/query-frontend.md
krasi-georgiev: ok after looking at this again I think you are right it will be much better to duplicate only the fields we use. Refactored, could you have another look and then I will start working on the tests

pkg/queryfrontend/roundtrip.go
krasi-georgiev: done

kakkoyun:
@krasi-georgiev I believe this will add support for memcached as a cache backend. Could you please update the title and changelog accordingly?

CHANGELOG.md
```diff
@@ -11,6 +11,8 @@ We use *breaking* word for marking changes that are not backward compatible (rel
 
 ## Unreleased
 
+- [#3032](https://github.com/thanos-io/thanos/pull/3032) Query Frontend: It now uses the corex module fully and it support all backends that are supported there.
```
kakkoyun, line 14:
```suggestion
- [#3032](https://github.com/thanos-io/thanos/pull/3032) Query Frontend: It now uses the Cortex module fully and it support all backends that are supported there.
```

Maybe specify a list for supported backends?

krasi-georgiev: Updated thanks

krasi-georgiev:
> @krasi-georgiev I believe this will add support for memcached as a cache backend. Could you please update the title and changelog accordingly?

updated thanks

docs/components/query-frontend.md
```diff
@@ -45,21 +45,49 @@ Query Frontend supports a retry mechanism to retry query when HTTP requests are
...
+    max_size_items: 0
+    validity: 0s
+    size: 0
+  prefix: ""
 ```
 
 ### Slow Query Log
 
-Query Frontend supports `--query-frontend.log_queries_longer_than` flag to log queries running longer some duration.
+Query Frontend supports `--query-frontend.log_queries_longer_than` flag to log queries running longer then some duration.
```
yeya24, line 93: `then` -> `than`

pkg/queryfrontend/config.go
```diff
@@ -0,0 +1,56 @@
...
+	Limits     Limits
+	QueryRange QueryRange
+	Frontend   Frontend
+}
+
+// Validate validates the config.
+func (cfg *Config) Validate() error {
+	err := errors.Wrapf(cfg.QueryRange.Validate(), "query range config validation")
+	err = errors.Wrapf(cfg.Frontend.Validate(), "frontend config validation")
+	return err
```
yeya24, line 55: What about inline line 54 and 55?

cmd/thanos/query-frontend.go
```diff
@@ -111,51 +92,49 @@ func runQueryFrontend(
...
-		return errors.Wrap(err, "get content of response cache configuration")
-	}
-
-	var cacheConfig *queryrange.ResultsCacheConfig
-	if len(respCacheContentYaml) > 0 {
-		cacheConfig, err = cache.NewResponseCacheConfig(respCacheContentYaml)
-		if err != nil {
-			return errors.Wrap(err, "create response cache")
-		}
+		return errors.Wrap(err, "initialiase limits")
```
yeya24, line 129: A small typo here. `initialise`

cmd/thanos/query-frontend.go
```diff
@@ -111,51 +92,49 @@ func runQueryFrontend(
...
-		conf.queryRangeConfig.cacheMaxFreshness,
-	)
-
-	respCacheContentYaml, err := conf.queryRangeConfig.respCacheConfig.Content()
+	limits, err := validation.NewOverrides(validation.Limits{
+		MaxQueryLength:      cfg.Limits.MaxQueryLength,
+		MaxQueryParallelism: cfg.Limits.MaxQueryParallelism,
+		MaxCacheFreshness:   cfg.Limits.MaxCacheFreshness,
+	}, nil)
 	if err != nil {
```
yeya24, line 128: Is it necessary to check error here. `validation.NewOverrides` always returns nil.

pkg/queryfrontend/roundtrip.go
```diff
@@ -62,16 +64,11 @@ func NewTripperWare(
 		)
 	}
 
-	if cacheConfig != nil {
-		// constSplitter will panic when splitQueryInterval is 0.
-		if splitQueryInterval == 0 {
-			return nil, errors.New("cannot create results cache middleware when split interval is 0")
-		}
-
+	if queryRange.ResultsCacheConfig != (queryrange.ResultsCacheConfig{}) {
```
yeya24, line 67: This check is still needed here, otherwise it panics when splitting queries.

```
		if splitQueryInterval == 0 {
			return nil, errors.New("cannot create results cache middleware when split interval is 0")
		}
```

pkg/queryfrontend/config.go
krasi-georgiev: prefer to keep it like this to make it easier to read.

cmd/thanos/query-frontend.go
krasi-georgiev: We don't know that this will remain like this. The implementation can change.

pkg/queryfrontend/roundtrip.go
krasi-georgiev: Thanks! added to the config validation logic.

cmd/thanos/query-frontend.go
krasi-georgiev: done, thanks

cmd/thanos/query-frontend.go
yeya24: Ahh, makes sense.

cmd/thanos/query-frontend.go
```diff
@@ -16,93 +16,74 @@ import (
 	"github.com/opentracing/opentracing-go"
 	"github.com/pkg/errors"
 	"github.com/prometheus/client_golang/prometheus"
-	"github.com/prometheus/common/model"
-	"github.com/thanos-io/thanos/pkg/extkingpin"
 	"github.com/weaveworks/common/user"
+	"gopkg.in/yaml.v2"
 
+	"github.com/cortexproject/cortex/pkg/util/validation"
```
yeya24, line 22: One small nit. Can you please move this import above?

yeya24 (commented):
Thanks for all the work! :1st_place_medal: Can I have some other :eyes:  to review this pr @kakkoyun @squat?

From my point of view, it is all good, just need to update the tests. Also some flags name and configs are changed, it would be good to add these changes to the changelog.

The Cortex cache config is a little bit complex, it would be good to add a Cortex doc link https://cortexmetrics.io/docs/configuration/configuration-file/. What do you think?

cmd/thanos/query-frontend.go
krasi-georgiev: updated

krasi-georgiev:
> Also some flags name and configs are changed, it would be good to add these changes to the changelog.

yep updated the changelog.

> The Cortex cache config is a little bit complex, it would be good to add a Cortex doc link cortexmetrics.io/docs/configuration/configuration-file. What do you think?

Yep good idea added in the readme.

cmd/thanos/query-frontend.go
```diff
@@ -10,99 +10,79 @@ import (
...
 	extpromhttp "github.com/thanos-io/thanos/pkg/extprom/http"
 	"github.com/thanos-io/thanos/pkg/logging"
 	"github.com/thanos-io/thanos/pkg/prober"
 	"github.com/thanos-io/thanos/pkg/queryfrontend"
-	"github.com/thanos-io/thanos/pkg/queryfrontend/cache"
 	httpserver "github.com/thanos-io/thanos/pkg/server/http"
 	"github.com/thanos-io/thanos/pkg/server/http/middleware"
 	"github.com/thanos-io/thanos/pkg/tracing"
+	"github.com/weaveworks/common/user"
+	"gopkg.in/yaml.v2"
```
yeya24, line 32: Can you please order the two import statements?

yeya24 (commented):
E2E tests failed, but I am not sure whether this is related or not. @kakkoyun Can you please help rerun the tests?

cmd/thanos/query-frontend.go
krasi-georgiev: these are auto sorted by goimports

docs/components/query-frontend.md
```diff
@@ -44,22 +44,55 @@ Query Frontend supports a retry mechanism to retry query when HTTP requests are
...
-
-[embedmd]:# (../flags/config_response_cache_in_memory.txt yaml)
+[embedmd]:# (../flags/config_frontend_cache.txt yaml)
 ```yaml
-type: IN-MEMORY
-config:
-  max_size: ""
-  max_size_items: 0
-  validity: 0s
+- enable_fifocache: false
```
kakkoyun, line 54: These configuration format kind of bothers me. Is this really what we want to do? 
This configuration format not idiomatic to Thanos. And also we are just exposing a configuration API from a dependency of ours directly. 
Can we do something about this?

docs/components/query-frontend.md
```diff
@@ -44,22 +44,55 @@ Query Frontend supports a retry mechanism to retry query when HTTP requests are
...
+    max_size_items: 0
+    validity: 0s
+    size: 0
+  prefix: ""
 ```
 
 ### Slow Query Log
 
-Query Frontend supports `--query-frontend.log_queries_longer_than` flag to log queries running longer some duration.
+Query Frontend supports `--query-frontend.log_queries_longer_than` flag to log queries running longer than some duration.
```
kakkoyun, line 95: This flag is renamed.
```suggestion
Query Frontend supports `--query-frontend.log-queries-longer-than` flag to log queries running longer than some duration.
```

cmd/thanos/query-frontend.go
```diff
@@ -10,99 +10,79 @@ import (
...
 	// partialResponseStrategy is the default strategy used
 	// when parsing thanos query request.
 	partialResponseStrategy bool
 }
 
-func (c *queryRangeConfig) registerFlag(cmd extkingpin.FlagClause) {
-	cmd.Flag("query-range.split-interval", "Split queries by an interval and execute in parallel, 0 disables it.").
-		Default("24h").SetValue(&c.splitInterval)
+func registerQueryFrontend(app *extkingpin.App) {
+	comp := component.QueryFrontend
```
kakkoyun, line 47: Why are we changing all these flag names? What's the motivation behind it?

kakkoyun (commented):
It looks solid. However I've a couple of questions about the overall user faced changes.
Maybe we could minimize user face changes. 
Especially I'm not sure about the configuration format we are exposing.

docs/components/query-frontend.md
krasi-georgiev: changing this means more maintenance, I agree that it doesn't look nice, but I think we should work towards changing this upstream in Thanos. keeping it tied together with Cortex means that people using the cortex front end can switch with their existing configs.

cmd/thanos/query-frontend.go
krasi-georgiev: because I think it makes sense to use exactly the same names as in Cortex

docs/components/query-frontend.md
krasi-georgiev: thanks updated

cmd/thanos/query-frontend.go
kakkoyun: WDYT? @yeya24 What was your motivation when you first changed these?

docs/components/query-frontend.md
kakkoyun: Does Cortex use similar configuration format as Thanos does? Can we make it happen? Otherwise I'm willing to take the responsibility of maintenance burden in order to keep Thanos configuration API consistent.

cmd/thanos/query-frontend.go
kakkoyun: Also because of these changes. We have to update kube-thanos as well. My vote on keeping them as is. Unless we want to make thanos query frontend a drop-in replacement for cortex frontend, which I don't see any value. We are already sharing a common codebase and if we can implement something on upstream we'll be implementing it.

krasi-georgiev: aha yeah good point about kube-thanos, but OTOH `query-frontend.log-queries-longer-than` needs changing anyway as it was with underscores so kube-thanos needs to be changed anyway.

docs/components/query-frontend.md
krasi-georgiev: This is the first step to make things more consistent, and my idea is to slowly contribute upstream to make things better. Also, the next step would be using the same code for the memcache client - at the moment Cortex has its own and Thanos uses a slightly different version.

Btw why do you think this change makes the API inconsistent?

kakkoyun: The previous style where we define type and specify the properties what we use across the board. Existence of a config enables corresponding backend. I find it more idiomatic for Thanos. e.g https://thanos.io/v0.15/components/store/

cmd/thanos/query-frontend.go
yeya24: I named these flags based on the original proposal https://thanos.io/tip/proposals/202004_embedd_cortex_frontend.md/. 

I agree with @kakkoyun it is not necessary to be the same as Cortex flags and to do this we have to change a lot.

docs/components/query-frontend.md
yeya24: This config is completely different from the Thanos way. Cortex supports configuring multiple cache backend to be a tiered cache, while Thanos can only configure one.

I think it is hard to convert this config to some Thanos format

krasi-georgiev: yeah I completely forgot about that -  another reason to keep it with the cortex configs logic.

bwplotka: Interesting. How one would configure tiered cache in Cortex then even? :thinking: 

I would aim for our config type though - even if it reproduces a bit of configuration. We would rather aim for 1:1 configuration to other caching usages (index, bucket on store), then aiming for Cortex configuration style I think. The good question to be answered, is how the tiered cache if we want this would look like. I think I would see this as multiple items in this file with different types so all find on our side.

No need to force Cortex clients to have same config as well - for this we would need bigger discussion and maintaining just static configuration structs for better discoverability, readability sounds like not a bad idea to me (: But there is some work yes - let's do here whatever we can (e.g just Memcached) and produce issues to add rest.

WDYT? @krasi-georgiev 

And thanks for this work!

pkg/queryfrontend/roundtrip.go
```diff
@@ -22,13 +22,15 @@ const (
 	labelQueryRange = "query_range"
 )
 
+// NewTripperware returns a Tripperware configured with middlewares to
+// limit, align, split,cache requests and retry.
+// Not using the cortex one as it uses  query parallelisations based on
+// storage sharding configuration and query ASTs.
 func NewTripperWare(
```
bwplotka, line 29:
```suggestion
func NewTripperware(
```

bwplotka (commented):
Nice! Some comments for a start :hugs:

cmd/thanos/query-frontend.go
krasi-georgiev: > Sure, but how flag naming changes anything? Still, I don't get the changes to those flags in some cases. The underscore -> hyphen makes sense but some naming was adjusted for readability reasons.

ooh yes you are right only the config struct passes through some validation.

krasi-georgiev:
> No need to force Cortex clients to have same config as well - for this we would need bigger discussion and maintaining just static configuration structs for better discoverability, readability sounds like not a bad idea to me (: But there is some work yes - let's do here whatever we can (e.g just Memcached) and produce issues to add rest.

ok we can do it this way as well, what about the last argument for a drop-in replacement of the cortex binary. Keeping the configs the same will allow people to just switch from cortex to thanos-query-frontend.

I guess to unblock this can just change to the  Thanos idiomatic config style and then open an issue on the Cortex side to see if we can come up with some middle ground.

cmd/thanos/query-frontend.go
```diff
@@ -10,99 +10,79 @@ import (
...
+	cmd := app.Command(comp.String(), "query frontend")
+	cfg := &config{}
+
+	cfg.http.registerFlag(cmd)
+
+	cmd.Flag("query-range.split-queries-by-interval", "Split queries by an interval and execute in parallel, 0 disables it.").
```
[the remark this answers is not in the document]
krasi-georgiev: reverted all cli flag renamings

cmd/thanos/query-frontend.go
```diff
@@ -10,99 +10,79 @@ import (
...
-	c.http.registerFlag(cmd)
+	cfg.cachePathOrContent = *extflag.RegisterPathOrContent(cmd, "query-range.cache-config", "YAML file that contains response cache configuration.", false)
 
 	cmd.Flag("query-frontend.downstream-url", "URL of downstream Prometheus Query compatible API.").
-		Default("http://localhost:9090").StringVar(&c.downstreamURL)
+		Default("http://localhost:9090").StringVar(&cfg.Frontend.DownstreamURL)
 
-	cmd.Flag("query-frontend.compress-responses", "Compress HTTP responses.").
-		Default("false").BoolVar(&c.compressResponses)
+	cmd.Flag("query-frontend.compress-http-responses", "Compress HTTP responses.").
```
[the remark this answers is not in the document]
yeya24: I agree no need to change here. It is already very clear.

bwplotka:
> ok we can do it this way as well, what about the last argument for a drop-in replacement of the cortex binary. Keeping the configs the same will allow people to just switch from cortex to thanos-query-frontend.

True, but also introduces lots of inconsistencies for those who were waiting until Thanos itself have this component.

> One thing I like that in Cortex side is that all configs are also converted to cli flags, so you can decide when to use a flag and when yaml files. Not sure how useful is this actually in production.

Depends. To us I think flags are superior if we can have them, also "one way of doing things" is kind of violated then. And if we want to consider this we need to be consisent and move everything else (:

krasi-georgiev:
refactored to be consistent with the other Thanos cache components

please take another look @kakkoyun @bwplotka @yeya24

kakkoyun (approved):
LGTM. Let's what others also think.

test/e2e/query_test.go
```diff
@@ -458,6 +458,7 @@ func labelValues(t *testing.T, ctx context.Context, addr, label string, start, e
 	}))
 }
 
+//nolint:unparam
```
bwplotka, line 461: Why this? Can we just fix this function? (:

bwplotka (approved):
Awesome, minor nits only otherwise LGTM, thanks!

test/e2e/query_test.go
yeya24: This one was added by me in my original pr. The error is  

Screenshot from 2020-09-15 17-57-38

I just use the same query in E2E and I think it is okay.

pkg/queryfrontend/config.go
```diff
@@ -0,0 +1,149 @@
...
+
+// Config holds the query frontend configs.
+type Config struct {
+	// PartialResponseStrategy is the default strategy used
+	// when parsing thanos query request.
+	PartialResponseStrategy bool
+
+	CortexFrontendConfig     *cortexfrontend.Config
+	CortexLimits             *cortexvalidation.Limits
+	CortexResultsCacheConfig *queryrange.ResultsCacheConfig
```
yeya24, line 125: Not a blocker. Is it necessary to add `Cortex` prefix to these fields?

pkg/queryfrontend/config.go
```diff
@@ -0,0 +1,149 @@
...
+					UpdateInterval: config.Memcached.DNSProviderUpdateInterval,
+				},
+				Background: cortexcache.BackgroundConfig{
+					WriteBackBuffer:     config.Memcached.MaxAsyncBufferSize,
+					WriteBackGoroutines: config.Memcached.MaxAsyncConcurrency,
+				},
+			},
+		}, nil
+	default:
+		return nil, errors.Errorf("index cache with type %s is not supported", cacheConfig.Type)
```
yeya24, line 113: s/index/response/g

pkg/queryfrontend/roundtrip.go
```diff
@@ -40,8 +40,13 @@ func NewTripperWare(
 	queriesCount.WithLabelValues(labelQuery)
 	queriesCount.WithLabelValues(labelQueryRange)
 
+	overrides, err := validation.NewOverrides(*config.CortexLimits, nil)
```
yeya24, line 43: Umm, is it better to rename `overrides` to just `limits`? `overrides` is only meaningful in Cortex.

pkg/queryfrontend/config.go
```diff
@@ -0,0 +1,149 @@
...
+	RequestLoggingDecision string
+
+	SplitQueriesByInterval time.Duration
+	MaxRetries             int
+}
+
+// Validate a fully initialized config.
+func (cfg *Config) Validate() error {
+	if cfg.CortexResultsCacheConfig != nil {
+		if cfg.SplitQueriesByInterval <= 0 {
```
yeya24, line 137: Do we want to also include the `=0` case? Umm, currently `cfg.SplitQueriesByInterval = 0` disables query splitting. I think this change makes sense since splitting is the core of query frontend, but we need to update the flag description. https://github.com/thanos-io/thanos/pull/3114/files#diff-ca5cdb3d99d056da38d7221457666c9aR52

scripts/cfggen/main.go
```diff
@@ -95,8 +97,8 @@ func main() {
 		}
 	}
 
-	for typ, config := range responseCacheConfigs {
-		if err := generate(responsecache.ResponseCacheConfig{Type: typ, Config: config}, generateName("response_cache_", string(typ)), *outputDir); err != nil {
+	for typ, config := range queryfrontendCacheConfigs {
+		if err := generate(queryfrontend.CacheProviderConfig{Type: typ, Config: config}, generateName("query_frontend_cache_", string(typ)), *outputDir); err != nil {
```
yeya24, line 101: Why not using `response_cache` here?

test/e2e/e2ethanos/services.go
```diff
@@ -406,18 +406,18 @@ func NewCompactor(sharedDir string, name string, bucketConfig client.BucketConfi
 	return compactor, nil
 }
 
-func NewQueryFrontend(name string, downstreamURL string, respCacheConf cache.ResponseCacheConfig) (*e2e.HTTPService, error) {
-	respCacheConfigBytes, err := yaml.Marshal(respCacheConf)
+func NewQueryFrontend(name string, downstreamURL string, cacheConfig queryfrontend.CacheProviderConfig) (*e2e.HTTPService, error) {
+	cacheConfigBytes, err := yaml.Marshal(cacheConfig)
 	if err != nil {
-		return nil, errors.Wrapf(err, "generate response cache config file: %v", respCacheConf)
+		return nil, errors.Wrapf(err, "marshal query frontend config file: %v", cacheConfig)
```
yeya24, line 412: Don't need to change here. This is the response cache config.

yeya24 (commented):
Some very small nits, overall LGTM :rocket:

pkg/queryfrontend/config.go
yeya24: The doc needs to update as well https://github.com/thanos-io/thanos/blob/master/docs/components/query-frontend.md#splitting

pkg/queryfrontend/config.go
krasi-georgiev: when reading the code later it is a bit more clear

test/e2e/e2ethanos/services.go
krasi-georgiev: updated

scripts/cfggen/main.go
krasi-georgiev: updated

pkg/queryfrontend/config.go
krasi-georgiev: updated - have a look

pkg/queryfrontend/roundtrip.go
krasi-georgiev: updated

pkg/queryfrontend/config.go
krasi-georgiev: fixed

test/e2e/query_test.go
krasi-georgiev: yep leaving it as is.

pkg/queryfrontend/config.go
```diff
@@ -0,0 +1,149 @@
...
+					Parallelism: config.Memcached.MaxGetMultiConcurrency,
+					BatchSize:   config.Memcached.MaxGetMultiBatchSize,
+				},
+				MemcacheClient: cortexcache.MemcachedClientConfig{
+					Timeout:        config.Memcached.Timeout,
+					MaxIdleConns:   config.Memcached.MaxIdleConnections,
+					Addresses:      strings.Join(config.Memcached.Addresses, ","),
+					UpdateInterval: config.Memcached.DNSProviderUpdateInterval,
+				},
+				Background: cortexcache.BackgroundConfig{
```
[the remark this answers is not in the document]
krasi-georgiev: Anything looks strange here?

cmd/thanos/query-frontend.go
```diff
@@ -8,101 +8,80 @@ import (
...
-type queryFrontendConfig struct {
-	http             httpConfig
-	queryRangeConfig queryRangeConfig
-
-	downstreamURL        string
-	compressResponses    bool
-	LogQueriesLongerThan time.Duration
-
-	requestLoggingDecision string
+type config struct {
```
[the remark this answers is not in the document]
krasi-georgiev: done

pkg/queryfrontend/roundtrip.go
```diff
@@ -50,10 +55,12 @@ func NewTripperWare(
 		queryrange.StepAlignMiddleware,
 	)
 
-	if splitQueryInterval != 0 {
+	codec := NewThanosCodec(config.PartialResponseStrategy)
+
+	if config.SplitQueriesByInterval != 0 {
```
yeya24, line 60: Is this still necessary? It is already checked in https://github.com/thanos-io/thanos/pull/3114/files#diff-98bce32adf2d7692d381e0688ccaab15R137

yeya24 (approved):
LGTM!

PR #3115, opened 2020-09-02
cmd: compact: clean partial / marked blocks concurrently

GiedriusS:
## Changes

Clean partially uploaded and blocks marked for deletion concurrently
with the whole compaction/downsampling process. One iteration could
potentially take a few days so it should be nice to periodically clean
unneeded blocks in the background. Without this, there are huge spikes
in block storage usage. The spike's size depends on how long it takes to
complete one iteration.

The implementation of this is simple - factored out the deletion part
into a separate function. It is called at the end of an iteration +
concurrently if `--wait` has been specified. Add a mutex to protect from
concurrent runs.

## Verification

Updated e2e tests.


cmd/thanos/compact.go
```diff
@@ -295,6 +296,35 @@ func runCompact(
...
+		defer cleanMtx.Unlock()
+
+		// No need to resync before partial uploads and delete marked blocks. Last sync should be valid.
+		compact.BestEffortCleanAbortedPartialUploads(ctx, logger, sy.Partial(), bkt, partialUploadDeleteAttempts, blocksCleaned, blockCleanupFailures)
+		if err := blocksCleaner.DeleteMarkedBlocks(ctx); err != nil {
+			return errors.Wrap(err, "error cleaning marked blocks")
+		}
+
+		if err := sy.SyncMetas(ctx); err != nil {
+			level.Error(logger).Log("msg", "failed to sync metas", "err", err)
```
yeya24, line 312: Can I ask why not return an error here?

GiedriusS: Because IMHO it would be a bit of an overkill to stop the whole compaction process in case an error would happen during such a harmless operation i.e. when `--wait` has been specified. What do you think?

yeya24: Makes sense to me! Thanks.

cmd/thanos/compact.go
```diff
@@ -295,6 +296,35 @@ func runCompact(
...
+	var cleanMtx sync.Mutex
+	// TODO(GiedriusS): we could also apply retention policies here but the logic would be a bit more complex.
+	cleanPartialMarked := func() error {
+		cleanMtx.Lock()
+		defer cleanMtx.Unlock()
+
+		// No need to resync before partial uploads and delete marked blocks. Last sync should be valid.
+		compact.BestEffortCleanAbortedPartialUploads(ctx, logger, sy.Partial(), bkt, partialUploadDeleteAttempts, blocksCleaned, blockCleanupFailures)
+		if err := blocksCleaner.DeleteMarkedBlocks(ctx); err != nil {
+			return errors.Wrap(err, "error cleaning marked blocks")
```
yeya24, line 308: One nit, is it better to be `cleaning marked blocks`? `error` seems redundant.

yeya24 (approved):
Overall it looks very good! Just one small nit, thanks!

cmd/thanos/compact.go
bwplotka: Yes, let's do this. Still valid comment :+1:

cmd/thanos/compact.go
```diff
@@ -295,6 +296,35 @@ func runCompact(
 		level.Info(logger).Log("msg", "retention policy of 1 hour aggregated samples is enabled", "duration", retentionByResolution[compact.ResolutionLevel1h])
 	}
 
+	var cleanMtx sync.Mutex
+	// TODO(GiedriusS): we could also apply retention policies here but the logic would be a bit more complex.
```
bwplotka, line 300: Agree :+1:

cmd/thanos/compact.go
```diff
@@ -416,6 +441,23 @@ func runCompact(
...
+		if conf.cleanupBlocksInterval > 0 {
+			g.Add(func() error {
+				// Wait the whole period at the beginning because we've executed this on boot.
+				select {
+				case <-time.After(conf.cleanupBlocksInterval):
+				case <-ctx.Done():
+				}
+				return runutil.Repeat(conf.cleanupBlocksInterval, ctx.Done(), func() error {
+					return cleanPartialMarked()
+				})
```
bwplotka, line 455:
```suggestion
				return runutil.Repeat(conf.cleanupBlocksInterval, ctx.Done(), cleanPartialMarked)
```

cmd/thanos/compact.go
```diff
@@ -416,6 +441,23 @@ func runCompact(
 
 		srv.Handle("/", r)
 
+		// Periodically remove partial blocks and blocks marked for deletion
+		// since one iteration potentially could take a long time.
+		if conf.cleanupBlocksInterval > 0 {
+			g.Add(func() error {
+				// Wait the whole period at the beginning because we've executed this on boot.
```
bwplotka, line 448: So... why not just removing this and removing boot time execution? (: Same stuff right?

cmd/thanos/compact.go
```diff
@@ -507,6 +550,8 @@ func (cc *compactConfig) registerFlag(cmd extkingpin.FlagClause) {
 		Default("20").IntVar(&cc.blockSyncConcurrency)
 	cmd.Flag("block-viewer.global.sync-block-interval", "Repeat interval for syncing the blocks between local and remote view for /global Block Viewer UI.").
 		Default("1m").DurationVar(&cc.blockViewerSyncBlockInterval)
+	cmd.Flag("compact.cleanup-interval", "How often we should clean up partially uploaded blocks and blocks with deletion mark in the background when --wait has been enabled. Setting it to \"0s\" disables it - the cleaning will only happen at the end of an iteration.").
```
bwplotka, line 553: :+1:

bwplotka (approved):
Awesome, thanks! 

Couple of comments but overall LGTM :muscle:

prmsrswt (approved):
Nice, LGTM!

bwplotka:
Where we are with this? I want to cut 0.16.0-rc.0 tomorrow :hugs:

cmd/thanos/compact.go
GiedriusS: No, not really, because we want to give our users the ability to disable this concurrent cleaning for some reason i.e. to save up on S3 GET/LIST operation costs

cmd/thanos/compact.go
GiedriusS: Added this. Thanks!

cmd/thanos/compact.go
bwplotka: It sounds wrong that we do more complex code only because we don't want to change tests :thinking:

bwplotka (approved):
LGTM for now, but I think we could improve a bit in future (: But not a blocker, LGTM!

Thanks :+1:

cmd/thanos/compact.go
GiedriusS: The tests would be much more complex and probably out of the scope of this PR. Actually, it's not just that, I think it's nice that we do this at least once. Imagine where someone doesn't use `--wait` and the whole Thanos Compact process ended before the clean-up has happened. Space usage would never go down in the remote object storage even though it could. And the user then could be charged more as a result of this not happening.

GiedriusS:
I guess since the approvals are there and I have cleaned up the CHANGELOG.md, I'll merge this. I also ran this for a bit locally with `--wait-interval=1s --wait --compact.cleanup-interval=1s` locally and it seems to work (tm). Will probably deploy this into production quite quickly because the saw-tooth pattern is becoming really spiky due to the fact that one iteration is taking a long time :( I hope that's ok with everyone! Thanks for the reviews :heart:

PR #3121, opened 2020-09-03
receive: Add ability to consume contents of the hashrings file directly

### LINKED ISSUE #3026: receive: Use PathOrContent file for the hashring
This: https://github.com/thanos-io/thanos/blob/a64884b19dbe7a6c2613078a07409c89ce4e2508/cmd/thanos/receive.go#L67

kakkoyun:

Fixes #3026.

## Changes

* Added `--receive.hashrings` alternative to `receive.hashrings-file` flag (lower priority). Content of JSON file that contains the hashring configuration.

## Verification

* `make test-local`
* `make test-e2e`

### DIFF
--- a/cmd/thanos/receive.go
+++ b/cmd/thanos/receive.go
@@ -106,14 +106,6 @@ func registerReceive(app *extkingpin.App) {
 			return errors.New("no external labels configured for receive, uniquely identifying external labels must be configured (ideally with `receive_` prefix); see https://thanos.io/tip/thanos/storage.md#external-labels for details.")
 		}
 
-		var cw *receive.ConfigWatcher
-		if *hashringsFile != "" {
-			cw, err = receive.NewConfigWatcher(log.With(logger, "component", "config-watcher"), reg, *hashringsFile, *refreshInterval)
-			if err != nil {
-				return err
-			}
-		}
-
 		tsdbOpts := &tsdb.Options{
 			MinBlockDuration:  int64(time.Duration(*tsdbMinBlockDuration) / time.Millisecond),
 			MaxBlockDuration:  int64(time.Duration(*tsdbMaxBlockDuration) / time.Millisecond),
@@ -376,8 +368,8 @@ func runReceive(
 		// watcher, we close the chan ourselves.
 		updates := make(chan receive.Hashring, 1)
 
-		// The Hashrings config file path given initializing config watcher.
-		if configPath, err := hashringsFile.Path(); err != nil {
+		// The Hashrings config file path is given initializing config watcher.
+		if configPath, err := hashringsFile.Path(); err == nil && configPath != "" {
 			cw, err := receive.NewConfigWatcher(log.With(logger, "component", "config-watcher"), reg, configPath, *refreshInterval)
 			if err != nil {
 				return errors.Wrap(err, "failed to initialize config watcher")
@@ -397,6 +389,7 @@ func runReceive(
 				cancel()
 			})
 		} else {
+			// The Hashrings config file path is not given, so initialize using content..
 			configContent, err := hashringsFile.Content()
 			if err != nil {
 				return errors.Wrap(err, "failed to read hashrings configuration file")
--- a/cmd/thanos/receive.go
+++ b/cmd/thanos/receive.go
@@ -65,7 +65,8 @@ func registerReceive(app *extkingpin.App) {
 
 	retention := extkingpin.ModelDuration(cmd.Flag("tsdb.retention", "How long to retain raw samples on local storage. 0d - disables this retention.").Default("15d"))
 
-	hashringsFile := extflag.RegisterPathOrContent(cmd, "receive.hashrings", "File that contains the hashring configuration.", false)
+	hashringsFilePath := cmd.Flag("receive.hashrings-file", "Path to file that contains the hashring configuration. A watcher is initialized to watch changes and update the hashring dynamically.").PlaceHolder("<path>").String()
+	hashringsFileContent := cmd.Flag("receive.hashrings", "Alternative to 'receive.hashrings-file' flag (lower priority). Content of file that contains the hashring configuration.").PlaceHolder("<content>").String()
 
 	refreshInterval := extkingpin.ModelDuration(cmd.Flag("receive.hashrings-file-refresh-interval", "Refresh interval to re-read the hashring configuration file. (used as a fallback)").
 		Default("5m"))
@@ -151,7 +152,8 @@ func registerReceive(app *extkingpin.App) {
 			tsdbOpts,
 			*ignoreBlockSize,
 			lset,
-			hashringsFile,
+			*hashringsFilePath,
+			*hashringsFileContent,
 			refreshInterval,
 			*localEndpoint,
 			*tenantHeader,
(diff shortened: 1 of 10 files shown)


cmd/thanos/receive.go
```diff
@@ -63,8 +65,7 @@ func registerReceive(app *extkingpin.App) {
 
 	retention := modelDuration(cmd.Flag("tsdb.retention", "How long to retain raw samples on local storage. 0d - disables this retention.").Default("15d"))
 
-	hashringsFile := cmd.Flag("receive.hashrings-file", "Path to file that contains the hashring configuration.").
-		PlaceHolder("<path>").String()
+	hashringsFile := extflag.RegisterPathOrContent(cmd, "receive.hashrings", "JSON file that contains the hashring configuration.", false)
```
squat, line 68: nit: I don't think we _have_ to specify JSON here, this may turn into YAML in the future anyways

squat:
Looking forward to taking a close look at this :+1:

kakkoyun:
Friendly ping @squat 
Also somewhat related https://github.com/thanos-io/thanos/pull/3138

bwplotka (approved):
Thanks! LGTM :muscle: perfect.

bwplotka (approved):
LGTM from my side.

pkg/extflag/pathorcontent.go
```diff
@@ -44,30 +44,45 @@ func RegisterPathOrContent(cmd FlagClause, flagName string, help string, require
...
 	if len(content) == 0 && p.required {
-		return nil, errors.Errorf("flag %s or %s is required for running this command and content cannot be empty.", fileFlagName, contentFlagName)
+		return nil, errors.Errorf("flag %s or %s is required for running this command and content cannot be empty.", fileFlagName, p.flagName)
 	}
 
 	return content, nil
 }
+
+// Path returns the path of the file. Flag that specifies path has priority.
+// It returns error if the required flag is set to true.
```
squat, line 75: It also returns errors if the required flag is not set to true but both a path and content are given.

cmd/thanos/receive.go
```diff
@@ -369,7 +364,13 @@ func runReceive(
 		// watcher, we close the chan ourselves.
 		updates := make(chan receive.Hashring, 1)
 
-		if cw != nil {
+		// The Hashrings config file path given initializing config watcher.
+		if configPath, err := hashringsFile.Path(); err != nil {
```
squat, line 368: this logic seems backwards to me. If we receive an error getting the path then we continue and try to use the path? When `Path()` returns an error, the returned path is `""` so this will not work. Am I understanding it incorrectly?

pkg/receive/config.go
```diff
@@ -238,50 +245,43 @@ func (cw *ConfigWatcher) refresh(ctx context.Context) {
...
+			level.Error(logger).Log("msg", "failed to close file", "err", err, "path", path)
 		}
 	}()
 
 	return ioutil.ReadAll(fd)
 }
 
 // parseConfig parses the raw configuration content and returns a HashringConfig.
-func (cw *ConfigWatcher) parseConfig(content []byte) ([]HashringConfig, error) {
+func parseConfig(content []byte) ([]HashringConfig, error) {
```
squat, line 284: IMO I'm not sure it's worth restructuring the whole file to reuse a func that only does a json unmarshal but it's not a blocker

cmd/thanos/receive.go
kakkoyun: Thanks for catching this. Fixed the issue.

kakkoyun:
@squat Thanks for the great review as always. I have addressed your concerns, could please take another look at it?

cmd/thanos/receive.go
```diff
@@ -383,15 +384,34 @@ func runReceive(
 
 			ctx, cancel := context.WithCancel(context.Background())
 			g.Add(func() error {
-				return receive.HashringFromConfig(ctx, updates, cw)
+				return receive.HashringFromConfigWatcher(ctx, updates, cw)
 			}, func(error) {
 				cancel()
 			})
 		} else {
+			// The Hashrings config file path is not given, so initialize using content..
```
bwplotka, line 392:
```suggestion
			// The Hashrings config file path is not given, so initialize using content.
```

pkg/extflag/pathorcontent.go
```diff
@@ -44,30 +44,47 @@ func RegisterPathOrContent(cmd FlagClause, flagName string, help string, require
...
+		return nil, errors.Errorf("flag %s or %s is required for running this command and content cannot be empty.", fileFlagName, p.flagName)
 	}
 
 	return content, nil
 }
+
+// Path returns the path of the given file that has passed to the flag.
+// It returns error if both a path and content are given.
+// It returns error if the required flag is set to true and path is empty.
+func (p *PathOrContent) Path() (string, error) {
```
bwplotka, line 78: Hmmmm...

So `Content()` is a bit different semantically. We aim for Content to be always there unless flag is optional.

With Path.. we have one more case, of user not specifying path at al, but specifying content. While technically this method should be understandable there are so many cases to tests to check which of those 6 cases we have (content passed, path passed, filag not used) x2 (required or not)

While all of this we have also this important message that is hidden in terms of hashring file:

IF you specify via content it will be used. If specify via path it will be dynamically reloadable. There are 2 problems with this:

1. This is surprising. We never reload content. How user can know we do it now? And let's say we are explicit we reload hashring. How user will know that other flags then hashrings are not dynamically reloadable?
2. What if I want via path but don't want to dynamically reload? For example I want to put file, start receive and remove (for some reason). Maybe rare, but someone could be surprised this is not possible. Might other usecases as well.

I think we need to think this through. I have some ideas, but before I state them,  maybe you have as well? 🤗 Do you see the problem?

bwplotka (changes requested):
Love it especially extra tests. But I think we can do better on pathOrContent API 🤗 Let's do it properly as it's quite flexible flag, so would be nice to have clean API.

pkg/extflag/pathorcontent.go
kakkoyun: > This is surprising. We never reload content. How user can know we do it now? And let's say we are explicit we reload hashring. How user will know that other flags then hashrings are not dynamically reloadable?

I understand the concern. Maybe overloading the `PathOrContent` for the Receive's use case is not the best way.
Receive has a special way to handle the given path as you already mentioned. It initializes a watcher for the given path.

> What if I want via path but don't want to dynamically reload? For example I want to put file, start receive and remove (for some reason). Maybe rare, but someone could be surprised this is not possible. Might other usecases as well.

This is a problem we already have. This PR shouldn't be aiming to solve this. However, I can add more documentation to clarify this behaviour.

How about rather than overloading the `PathOrContent`, we implement dedicated logic in the receive. Specific to receive? Would this work? WDYT?

bwplotka: hm what about, creating a second constructor?

So we can have:

* `something` or `something-file` as one constructor
* `something` and `something-file` and `something-reloadable-file` if you use another constructor

So receive will use latter (: Then we can actually embed file watching in the second constructor, directly in pathOrContent package 🤗 

WDYT?

kakkoyun: I like the idea. I would love to work on it on a separate PR to introduce this feature in a more generalized form.
How about merging this PR as is and then creating an issue for it? As I told you I can work on it. WDYT?

kakkoyun: @bwplotka As we talked off-line, do we have a consensus? Can we proceed with this?

bwplotka (approved):
Well done with short term thing 👍🏽 

It will have the same flag spec 🤗

pkg/extflag/pathorcontent.go
bwplotka: YES! love it. Thanks!

PR #3125, opened 2020-09-04
[WIP] store: Add Groupcache as a cache backend

### LINKED ISSUE #2962: Research groupcache as alternative cache client to memcached
I think something which would easy up Thanos deployment with cache is https://github.com/golang/groupcache This is project done my Memcached author and it's replicated in mem cache. The only tradeoff is gossip between replicas which is not that basd to setup.

To research and benchmark (:

kakkoyun:
WIP

## TODO
- [  ] Add groupcache
- [  ] Configure groupcache
- [  ]  Fix merge issues
- [  ] Fix dependency issues

Fixes #2962


brancz:
Let's say I were to run thanos store shards in HA, would we duplicate the cache in each instance? Or would cache locality make this even better as I understand groupcache can understand hot keys and distribute them. Generally I think this is really cool and valuable to experiment with the trade-offs.

bwplotka:
uuuu <3

> Let's say I were to run thanos store shards in HA, would we duplicate the cache in each instance? Or would cache locality make this even better as I understand groupcache can understand hot keys and distribute them. Generally I think this is really cool and valuable to experiment with the trade-offs.

I expect it to duplicate (as much as you give), with fetch from peer in a miss case before actually looking into obj storage etc (:

yeya24:
@krasi-georgiev did some refactorings on the query frontend cache in #3114 so we will use all Cortex cache package for query frontend later. So if we want to have this for query frontend, we might need to add it to Cortex.

Does GroupCache support any cache key expiration mechanism? I think this is a requirement to use it with the query frontend.

kakkoyun:
@yeya24 AFAIK, it doesn't have a TTL mechanism. I'm just experimenting with the idea. I don't think it's mature enough to propose something to Cortex at this stage. We have to try it first. I'll try to come up with a way to use it only in Thanos. 
Thanks a lot for checking out.

PR #3137, opened 2020-09-08
mixin: adding ability to have configurable namespace and selector

ahmednasir91:
Currently the thanos mixins are built with specific namespace and selector and have somewhat inconsistency in the selectors between alerts and dashboards.

In dashboards the queries are mostly based on `namespace` and `job` labels however in alerts its mostly using the `selector` property.

## Changes
To make the dashboards consistent, the following changes are done

- using `selector` variable instead of `job=~{jobPrefix}`
- adding ability to use `namespace` label for cases where the label name for `namespace` is different for example the prometheus stable chart, it relabels `namespace` -> `kubernetes_namespace`.

Looking forward for the feedback, if the changes makes sense and any improvement required.

Thank you

Will update these once the changes are acceptable to reviewers

## Verification

Installed on test system and works perfectly fine.

### DIFF
--- a/mixin/README.md
+++ b/mixin/README.md
@@ -100,7 +100,7 @@ This project is intended to be used as a library. You can extend and customize d
   dashboard+:: {
     prefix: 'Thanos / ',
     tags: ['thanos-mixin'],
-    namespaceQuery: 'kube_pod_info',
+    namespaceMetric: 'kube_pod_info',
   },
 }
 ```
--- a/mixin/dashboards/bucket_replicate.libsonnet
+++ b/mixin/dashboards/bucket_replicate.libsonnet
@@ -16,14 +16,14 @@ local g = import '../lib/thanos-grafana-builder/builder.libsonnet';
         .addPanel(
           g.panel('Rate') +
           g.qpsErrTotalPanel(
-            'thanos_replicate_replication_runs_total{result="error", namespace="$namespace",%(selector)s}' % thanos.bucket_replicate,
-            'thanos_replicate_replication_runs_total{namespace="$namespace",%(selector)s}' % thanos.bucket_replicate,
+            'thanos_replicate_replication_runs_total{result="error", %(namespaceLabel)s="$namespace",%(selector)s}' % thanos.bucket_replicate,
+            'thanos_replicate_replication_runs_total{%(namespaceLabel)s="$namespace",%(selector)s}' % thanos.bucket_replicate,
           )
         )
         .addPanel(
           g.panel('Errors', 'Shows rate of errors.') +
           g.queryPanel(
-            'sum(rate(thanos_replicate_replication_runs_total{result="error", namespace="$namespace",%(selector)s}[$interval])) by (result)' % thanos.bucket_replicate,
+            'sum(rate(thanos_replicate_replication_runs_total{result="error", %(namespaceLabel)s="$namespace",%(selector)s}[$interval])) by (result)' % thanos.bucket_replicate,
             '{{result}}'
           ) +
           { yaxes: g.yaxes('percentunit') } +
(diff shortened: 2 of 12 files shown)


mixin/README.md
```diff
@@ -100,7 +100,7 @@ This project is intended to be used as a library. You can extend and customize d
   dashboard+:: {
     prefix: 'Thanos / ',
     tags: ['thanos-mixin'],
-    namespaceQuery: 'kube_pod_info',
+    namespaceMetric: 'kube_pod_info',
```
kakkoyun, line 103: This could be a fully-fledged query, thus the name.

mixin/dashboards/bucket_replicate.libsonnet
```diff
@@ -6,31 +6,32 @@ local g = import '../lib/thanos-grafana-builder/builder.libsonnet';
     jobPrefix: error 'must provide job prefix for Thanos Bucket Replicate dashboard',
     selector: error 'must provide selector for Thanos Bucket Replicate dashboard',
     title: error 'must provide title for Thanos Bucket Replicate dashboard',
+    namespaceLabel: error 'must provide namespace label', 
   },
   grafanaDashboards+:: {
-    'bucket_replicate.json':
+    'thanos-bucket_replicate.json':
```
kakkoyun, line 12: What is the motivation behind add `thanos` prefix? Did it break something?

mixin/dashboards/compact.libsonnet
```diff
@@ -18,7 +19,7 @@ local g = import '../lib/thanos-grafana-builder/builder.libsonnet';
             'Shows rate of execution for compactions against blocks that are stored in the bucket by compaction group.'
           ) +
           g.queryPanel(
-            'sum(rate(thanos_compact_group_compactions_total{namespace="$namespace",job=~"$job"}[$interval])) by (job, group)',
+            'sum(rate(thanos_compact_group_compactions_total{%(namespaceLabel)s="$namespace",%(selector)s}[$interval])) by (job, group)' % thanos.compact,
```
kakkoyun, line 22: I'm not sure about job to selector change. It's specified as `job` here because we also aggregate on job for these queries and we have a dropdown for this variable. Rather than removing job we should have additional `selector` if it's needed.

kakkoyun (changes requested):
Thanks a lot for contribution.

I have left a couple of comments.
We should elaborate more on the motivation behind these changes. The alert selectors and dashboards selectors don't have to be perfectly in line IMHO. For the dashboards, `job` is kind of assumed.

That said, we are aware that the dashboards are far from being perfect and we are working on a couple of ideas with @metalmatze to improve these dashboards.

mixin/dashboards/bucket_replicate.libsonnet
ahmednasir91: When using with other mixins like the kubernetes-mixin. It makes sense to have a prefix as the files are in the same directory (the generated json files). For `bucket_replicate.json` its fine but for some names like `overview.json` it names sense to have some prefix.

The kubernetes-mixin is also following the approach of adding `k8s-` as a prefix
https://github.com/kubernetes-monitoring/kubernetes-mixin/blob/master/dashboards/resources/node.libsonnet#L32

What do you think?

mixin/dashboards/compact.libsonnet
ahmednasir91: Makes sense, I can selector as an additional one.

mixin/README.md
ahmednasir91: Currently in the grafana builder this is just defined as a metric name, so if we specify `namespaceQuery: 'kube_pod_info{app="..."}'` it will rather appear as `kube_pod_info\{app=\"...\"\}{}` in the json which makes it invalid.

https://github.com/thanos-io/thanos/blob/master/mixin/dashboards/compact.libsonnet#L133

ahmednasir91:
closing this PR will create separate PRs as i'm doing three things in one PR.

PR #3146, opened 2020-09-09
*: improve latency when streaming series from Prometheus

simonpasquier:

## Changes

I've found that when requesting many series (in the order of ten
thousands), the Thanos sidecar spends half of its time computing the
number of received series. To calculate the number of series, it needs
to build a label-based identifier for each chunked series and compare it
with the previous identifier. Eventually this number is only used for
logging and tracing so it doesn't feel like it's worth the penalty.

This change adds an histogram metric,
`thanos_sidecar_prometheus_store_received_frames`, which tracks the number
of frames per request received from the Prometheus remote read API
(buckets: 10, 100, 1000, 10000, 100000). It can be used to evaluate
whether expensive Series requests are performed.

## Verification

CPU profile before this PR (half of the time is spent in `prompb.(*Label).String`)

thanos_sidecar_before_cpu_profile

CPU profile after this PR

thanos_sidecar_after_cpu_profile

CPU graph showing the 50% improvement (the patched version of the Thanos sidecar has been deployed around 15:50)

https://snapshot.raintank.io/dashboard/snapshot/mufpH3k2yRJUYLM284SXESwVLG0Ns14i?orgId=2

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -15,6 +15,7 @@ We use *breaking :warning:* to mark changes that are not backward compatible (re
 
 - [#3133](https://github.com/thanos-io/thanos/pull/3133) Query: Allow passing a `storeMatch[]` to Labels APIs. Also time range metadata based store filtering is supported on Labels APIs.
 - [#3154](https://github.com/thanos-io/thanos/pull/3154) Query Frontend: Add metric `thanos_memcached_getmulti_gate_queries_max`.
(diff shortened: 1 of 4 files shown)


simonpasquier:
https://github.com/prometheus-operator/prometheus-operator/pull/3457 is the prometheus-operator PR including this change.

simonpasquier:
@bwplotka friendly ping :)

bwplotka (approved):
Wow. 

Amazing. Just curious how you found this? (: Just putting `timers` manually?

EDIT: Nvm, read your description. Thanks!

bwplotka:
Thanks, amazingly spotted :heart:

Issue #3119, opened 2020-09-03
grafana: Ensure Dashboard Vars Depending on LabelNames; LabelValues; Series are limited to certain time period / Add ability to set default in querier


bwplotka:
Currently unlimited time range for Label and LabelValues and Series Thanos APIs return and touch all the data as .. expected. This is great, but it does not make much sense for most of the cases.

We recently added time range parameters to those API endpoints for both Thanos & Prometheus: https://github.com/thanos-io/thanos/issues/1811 However we need to make sure that Grafana utilizes those. Otherwise Grafana always load years worth of series and labels every time params are refreshed.

Let's:
1. Add issue on Grafana side
2. Allow setting a sane default if time params are not specified. This is what Cortex team is doing - they allow fetching last 2h for non specified time ranges.

We also expect those endpoints to be cached in next iterations of Query Frontend (cc @kakkoyun ) but still this issue is valid.

Help wanted (:

kakkoyun:
xref: https://github.com/grafana/grafana/issues/11184

kakkoyun:
> Add issue on Grafana side
https://github.com/grafana/grafana/issues/27537

kakkoyun:
> Allow setting a sane default if time params are not specified. This is what Cortex team is doing - they allow fetching last 2h for non specified time ranges.

https://github.com/thanos-io/thanos/pull/3147

kakkoyun:
I'll handle what's left for this issue soon.

kakkoyun:
/remind me to finish this in a week

yeya24:
@kakkoyun Was this issue solved already by #3147?

kakkoyun:
@yeya24 Only a minor part is left, we need to adjust the default values for the interval. I want to do it for 0.20.

/remind me to finish this next Monday

PR #3147, opened 2020-09-09
querier: Add a flag to limit time range for metadata APIs (Labels and Series)

### LINKED ISSUE #1811: querier: Limit LabelNames; LabelValues to certain time period.

kakkoyun:

WIP

Ref #3119
Fixes #1811 

### Changes

* Add `query.metadata.default-time-range` flag to specify the default look-back duration for retrieving labels through Labels APIs when time range parameters are not specified.

### Verification

* `make test-local`

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -25,7 +25,7 @@ We use *breaking :warning:* to mark changes that are not backward compatible (re
 ### Changed
 
 - [#3136](https://github.com/thanos-io/thanos/pull/3136) Sidecar: Add metric `thanos_sidecar_reloader_config_apply_operations_total` and rename metric `thanos_sidecar_reloader_config_apply_errors_total` to `thanos_sidecar_reloader_config_apply_operations_failed_total`.
-- [#3154](https://github.com/thanos-io/thanos/pull/3154) Query: Add metric `thanos_query_gate_queries_max`. Remove metric `thanos_query_concurrent_selects_gate_queries_in_flight`.
+- [#3154](https://github.com/thanos-io/thanos/pull/3154) Querier: Add metric `thanos_query_gate_queries_max`. Remove metric `thanos_query_concurrent_selects_gate_queries_in_flight`.
 - [#3154](https://github.com/thanos-io/thanos/pull/3154) Store: Rename metric `thanos_bucket_store_queries_concurrent_max` to `thanos_bucket_store_series_gate_queries_max`.
 
 ## [v0.15.0](https://github.com/thanos-io/thanos/releases) - 2020.09.07
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -20,6 +20,7 @@ We use *breaking :warning:* to mark changes that are not backward compatible (re
 - [#3133](https://github.com/thanos-io/thanos/pull/3133) Query: Allow passing a `storeMatch[]` to Labels APIs. Also time range metadata based store filtering is supported on Labels APIs.
 - [#3154](https://github.com/thanos-io/thanos/pull/3154) Query Frontend: Add metric `thanos_memcached_getmulti_gate_queries_max`.
 - [#3146](https://github.com/thanos-io/thanos/pull/3146) Sidecar: Add `thanos_sidecar_prometheus_store_received_frames` histogram metric.
+- [#3147](https://github.com/thanos-io/thanos/pull/3147) Querier: Add `query.labels.lookback-delta` flag to specify the default lookback duration for retrieving labels through Labels APIs when time range parameters are not specified.
 
 ### Changed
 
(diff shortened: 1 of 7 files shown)


CHANGELOG.md
```diff
@@ -13,9 +13,10 @@ We use *breaking :warning:* to mark changes that are not backward compatible (re
 
 ### Added
 
-- [#3133](https://github.com/thanos-io/thanos/pull/3133) Query: Allow passing a `storeMatch[]` to Labels APIs. Also time range metadata based store filtering is supported on Labels APIs.
+- [#3133](https://github.com/thanos-io/thanos/pull/3133) Querier: Allow passing a `storeMatch[]` to Labels APIs. Also time range metadata based store filtering is supported on Labels APIs.
 - [#3154](https://github.com/thanos-io/thanos/pull/3154) Query Frontend: Add metric `thanos_memcached_getmulti_gate_queries_max`.
```
yeya24, line 17: Wait, why this is related to `Query Frontend` :laughing:  Would you mind updating it in this pr?

yeya24 (approved):
Looks perfect!

CHANGELOG.md
kakkoyun: It doesn't change anything in `Query Frontend`. Just renamed a changelog entry.

yeya24: Why the changelog title is `Query Frontend`, I am confused... It is `pkg/gate`, so should be `Querier`?

kakkoyun: Now, I get it. I was focusing on the changes that I've made. I can fix it as well.

prmsrswt (approved):
LGTM!

kakkoyun:
friendly ping @bwplotka

cmd/thanos/query.go
```diff
@@ -81,6 +82,8 @@ func registerQuery(app *extkingpin.App) {
 
 	instantDefaultMaxSourceResolution := modelDuration(cmd.Flag("query.instant.default.max_source_resolution", "default value for max_source_resolution for instant queries. If not set, defaults to 0s only taking raw resolution into account. 1h can be a good value if you use instant queries over time ranges that incorporate times outside of your raw-retention.").Default("0s").Hidden())
 
+	defaultLabelLookbackDelta := cmd.Flag("query.labels.api-lookback-delta", "The default lookback duration for retrieving labels through Labels API when the range parameters are not specified. The zero value means range covers the time since the beginning.").Default("0s").Duration()
```
bwplotka, line 85: Ok, I would not overuse `lookback-delta` naming. This is really specific to PromQL and it means quite a specific thing -> not queried time range, but rather the time when the series is considered dead. Maybe just `query.metadata.default-time-range`?

pkg/api/query/v1.go
```diff
@@ -659,3 +603,74 @@ func NewRulesHandler(client rules.UnaryClient, enablePartialResponse bool) func(
...
+var (
+	minTime = time.Unix(math.MinInt64/1000+62135596801, 0)
+	maxTime = time.Unix(math.MaxInt64/1000-62135596801, 999999999)
+)
+
+func parseLabelTimeRange(r *http.Request, defaultLabelLookbackDelta time.Duration) (time.Time, time.Time, error) {
+	// If start and end time not specified as query parameter, we get the range from the beginning of time by default.
+	var defaultStart, defaultEnd time.Time
+	if defaultLabelLookbackDelta == 0 {
+		defaultStart = minTime
```
bwplotka, line 616: `infMinTime`? `as `minTime` is often used as `startTime`. Same for max (:

pkg/api/query/v1.go
```diff
@@ -659,3 +603,74 @@ func NewRulesHandler(client rules.UnaryClient, enablePartialResponse bool) func(
...
+	start, err := parseTimeParam(r, "start", defaultStart)
+	if err != nil {
+		return time.Time{}, time.Time{}, &api.ApiError{Typ: api.ErrorBadData, Err: err}
+	}
+	end, err := parseTimeParam(r, "end", defaultEnd)
+	if err != nil {
+		return time.Time{}, time.Time{}, &api.ApiError{Typ: api.ErrorBadData, Err: err}
+	}
+	if end.Before(start) {
+		err := errors.New("end timestamp must not be before start time")
```
bwplotka, line 633: https://thanos.io/tip/coding-style-guide.md/#avoid-defining-variables-used-only-once (:

cmd/thanos/query.go
bwplotka: I propose `metadata` as I would love to cover `series` as well (on top of `labels, label_value`) I think it makes sense to group together, no need for flag for each of those :thinking: WDYT?

bwplotka (commented):
Amazing, thanks! Some suggestion, otherwise LGTM :+1:

kakkoyun:
@bwplotka Addressed the raised issues and included `series` API time range as well.

bwplotka (approved):
LGTM, :+1: 

Thanks! :muscle: 

I think it would be amazing to actually propose changing the default for this being ~1d or something in next release (:

yeya24 (approved):

PR #3148, opened 2020-09-10
tracing: Add spans for outgoing calls made by the memcached client

igorwwwwwwwwwwwwwwwwwwww:

## Changes

This adds tracing instrumentation to the memcached client. That should help diagnose issues with the cache, be it latency or cache effectiveness.

## Verification

Still need to figure out a good way to test this.


igorwwwwwwwwwwwwwwwwwwww:
Note that the upstream memcached client does not support opentracing, see https://github.com/bradfitz/gomemcache/issues/84.

pkg/cacheutil/memcached_client.go
```diff
@@ -340,11 +345,16 @@ func (c *memcachedClient) SetAsync(_ context.Context, key string, value []byte,
 		start := time.Now()
 		c.operations.WithLabelValues(opSet).Inc()
 
+		span, _ := tracing.StartSpan(ctx, "memcached_client_set")
```
igorwwwwwwwwwwwwwwwwwwww, line 348: Note that by its async nature, this span will appear later in the trace, after its parent has already finished. AFAIK most tracing tools should handle this case ok.

SuperQ (approved):
Nice

pkg/cacheutil/memcached_client.go
```diff
@@ -373,6 +384,10 @@ func (c *memcachedClient) SetAsync(_ context.Context, key string, value []byte,
 }
 
 func (c *memcachedClient) GetMulti(ctx context.Context, keys []string) map[string][]byte {
+	span, _ := tracing.StartSpan(ctx, "memcached_get_multi")
```
brancz, line 387: If I'm not mistaken we used to have this, but this created so many spans that it was impossible to make sense of the resulting trace. I believe @kakkoyun was working on this previously.

kakkoyun: I can't recall this. I guess you recall the spans related to mutli-tsdb actions. However you have a point we should make sure if this created too many spans.

kakkoyun (commented):
This looks solid. Thanks a lot for contributing.

I believe in order to truly track the performance issues with the cache, we need to make sure encapsulating actions also have necessary spans. AFAIK, we have spans for caching bucket however I'm not sure if we have spans already in place for index-caching. It'd be great if you could make sure we don't miss any of the spans for both index-cache and caching bucket.

WDYT?

bwplotka:
I recall the traces being unreadable with thousands on requests against Memcached for single request, but we can enable this if needed - maybe worth to enable this with some paam?

pracucci:
> I recall the traces being unreadable with thousands on requests against Memcached for single request

Yes, this was a deal breaker for us (Cortex). Maybe we could add an option to enable these spans, so that we can keep it disabled in Cortex?

bwplotka (changes requested):
Yes, if we want this, it has to be under flag / options (:

igorwwwwwwwwwwwwwwwwwwww:
@bwplotka I've brought this up-to-date with `master` and made the tracing configurable and opt-in, please take another look :)

PR #3154, opened 2020-09-11
pkg/gate: Prefix gate metrics for selects

simonpasquier:

## Changes

The metrics on the gates for concurrent selects were exposed without a prefix (e.g. `gate_queries_in_flight` and `gate_duration_seconds`) unlike the gate metrics for the query API. The first commit 60950ff adds the `thanos_query_concurrent_selects_` prefix.
After more thinking, I felt that exposing the number of in-flight queries doesn't make much sense for concurrent selects because each query creates a new gate. Thus we can have (max number of concurrent selects per query x max number of concurrent queries) selects in flight and it doesn't tell whether a query is blocked on too many selects or not. I've added another commit 
722181c that removes the `thanos_query_concurrent_selects_gate_queries_in_flight` metric.

The following metrics have also been added to record the maximum number of concurrent requests per gate:
* `thanos_query_gate_queries_max`
* `thanos_bucket_store_series_gate_queries_max`, previously known as `thanos_bucket_store_queries_concurrent_max.`
* `thanos_memcached_getmulti_gate_queries_max`

## Verification

Tested manually.

### DIFF
--- a/cmd/thanos/query.go
+++ b/cmd/thanos/query.go
@@ -32,6 +32,7 @@ import (
 	"github.com/thanos-io/thanos/pkg/extgrpc"
 	"github.com/thanos-io/thanos/pkg/extprom"
 	extpromhttp "github.com/thanos-io/thanos/pkg/extprom/http"
+	"github.com/thanos-io/thanos/pkg/gate"
 	"github.com/thanos-io/thanos/pkg/logging"
 	"github.com/thanos-io/thanos/pkg/prober"
 	"github.com/thanos-io/thanos/pkg/query"
@@ -426,11 +427,10 @@ func runQuery(
 
 		ins := extpromhttp.NewInstrumentationMiddleware(reg)
 		// TODO(bplotka in PR #513 review): pass all flags, not only the flags needed by prefix rewriting.
-		ui.NewQueryUI(logger, reg, stores, webExternalPrefix, webPrefixHeaderName).Register(router, ins)
+		ui.NewQueryUI(logger, stores, webExternalPrefix, webPrefixHeaderName).Register(router, ins)
 
 		api := v1.NewQueryAPI(
 			logger,
-			reg,
 			stores,
 			engine,
 			queryableCreator,
@@ -442,7 +442,10 @@ func runQuery(
 			queryReplicaLabels,
 			flagsMap,
 			instantDefaultMaxSourceResolution,
-			maxConcurrentQueries,
+			gate.New(
+				extprom.WrapRegistererWithPrefix("thanos_query_concurrent_", reg),
+				maxConcurrentQueries,
+			),
 		)
 
 		api.Register(router.WithPrefix("/api/v1"), tracer, logger, ins, logMiddleware)
--- a/cmd/thanos/query.go
+++ b/cmd/thanos/query.go
@@ -307,7 +307,7 @@ func runQuery(
 		rulesProxy       = rules.NewProxy(logger, stores.GetRulesClients)
 		queryableCreator = query.NewQueryableCreator(
 			logger,
-			extprom.WrapRegistererWithPrefix("thanos_query_concurrent_selects_", reg),
+			extprom.WrapRegistererWithPrefix("thanos_query_", reg),
 			proxy,
 			maxConcurrentSelects,
 			queryTimeout,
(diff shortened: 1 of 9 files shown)


cmd/thanos/query.go
```diff
@@ -304,8 +305,14 @@ func runQuery(
 		)
 		proxy            = store.NewProxyStore(logger, reg, stores.Get, component.Query, selectorLset, storeResponseTimeout)
 		rulesProxy       = rules.NewProxy(logger, stores.GetRulesClients)
-		queryableCreator = query.NewQueryableCreator(logger, reg, proxy, maxConcurrentSelects, queryTimeout)
-		engine           = promql.NewEngine(
+		queryableCreator = query.NewQueryableCreator(
+			logger,
+			extprom.WrapRegistererWithPrefix("thanos_query_concurrent_selects_", reg),
```
kakkoyun, line 310: Right now, this registry only used for concurrent selects. However, it could be used for others as well. I think we should wrap the registry this early.

pkg/gate/gate.go
```diff
@@ -12,67 +12,138 @@ import (
 	promgate "github.com/prometheus/prometheus/pkg/gate"
 )
 
-// Gate is an interface that mimics prometheus/pkg/gate behavior.
+var (
```
kakkoyun, line 15: Although I get the reasoning behind this change, the whole idea behind introducing the keeper was to remove metrics as package level variables. I wonder could we introduce some other pattern to avoid this.

kakkoyun (commented):
Thanks for the contributions. Looking good.
I have a couple of comments.

pkg/gate/gate.go
simonpasquier: these are only variables for metric description. The metrics themselves are created in `InstrumentGateDuration()` and `InstrumentGateInFlight`.

cmd/thanos/query.go
simonpasquier: Good point. I'll think about it.

pkg/gate/gate.go
kakkoyun: Oh yes, sorry 🤦‍♂️

cmd/thanos/query.go
simonpasquier: done

pkg/gate/gate.go
```diff
@@ -12,67 +12,144 @@ import (
...
 // Keeper is used to create multiple gates sharing the same metrics.
+//
+// Deprecated: when Keeper is used to create several gates, the metric tracking
+// the number of in-flight metric isn't meaningful because it is hard to say
+// whether requests are being blocked or not. For clients that call
+// gate.(*Keeper).NewGate only once, it is recommended to use gate.New()
+// instead. Otherwise it is recommended to use the
+// github.com/prometheus/prometheus/pkg/gate package directly and wrap the
+// returned gate with gate.InstrumentGateDuration().
 type Keeper struct {
```
kakkoyun, line 58: Is this still used anywhere in the codebase? If it's not the case, why not remove it directly? I don't think we have any downstream dependents.

pkg/query/querier.go
```diff
@@ -54,28 +60,26 @@ func NewQueryableCreator(logger log.Logger, reg prometheus.Registerer, proxy sto
...
-	reg                  prometheus.Registerer
 	replicaLabels        []string
 	storeMatchers        [][]storepb.LabelMatcher
 	proxy                storepb.StoreServer
 	deduplicate          bool
 	maxResolutionMillis  int64
 	partialResponse      bool
 	skipChunks           bool
-	gateKeeper           *gate.Keeper
+	gateFn               func() gate.Gate
```
kakkoyun, line 70: non-blocker suggestion: How about `gateProvider` or `gateProviderFn`?
```suggestion
	gateProvider               func() gate.Gate
```

pkg/gate/gate.go
simonpasquier: unfortunately we have, I should have pointed out in the PR.
https://github.com/cortexproject/cortex/blob/f1b95c68d3af4f715b11bd33c53ebaf83260f3bb/pkg/storegateway/bucket_stores.go#L72

simonpasquier: and given that Thanos imports Cortex too, it creates a circular dependency hence the need to keep `Keeper` :)

kakkoyun (approved):
🥇 Thanks a lot for addressing everything 💜

PR #3159, opened 2020-09-12
Extending GH issues staleness to kick in only after 2mo.

bwplotka:
I think we need to give more time for community to find time comment/contribute on interesting issues, let's extend staleness period a bit (: 


kakkoyun (approved):
lgtm

PR #3161, opened 2020-09-13
added key prop in Navigation

### LINKED ISSUE #3160: React-UI: "key" prop missing in Navigation
Thanos: v0.15.0
Prometheus: v2.21.0
Go: v1.15.1

**Object Storage Provider**:
NA

**What happened**:
To determine whether or not the component should be re-rendered, React uses the key prop which was missing in: https://github.com/thanos-io/thanos/blob/e31e93be758fc4c1e5befff5aa2e3f5ca2340062/pkg/ui/react-app/src/thanos/Navbar.tsx#L64.
Screenshot 2020-09-13 at 8 18 11 PM

Having unique key props to components so that React can handle the minimal DOM change.

Check `console.log` for when navbar loads.

NA

Made a PR for this issue: https://github.com/thanos-io/thanos/pull/3161

kunal-kushwaha:

## Changes
Fixes #3160 
Added `key` prop in `UncontrolledDropdown` & `DropdownItem` components of `Navigation` so that React can handle the minimal DOM change.

## Verification
The warnings shown in #3160 disappeared.

### DIFF
--- a/pkg/ui/react-app/src/thanos/Navbar.tsx
+++ b/pkg/ui/react-app/src/thanos/Navbar.tsx
@@ -84,13 +84,13 @@ const Navigation: FC<PathPrefixProps & NavigationProps> = ({ pathPrefix, thanosC
               );
 
             return (
-              <UncontrolledDropdown nav inNavbar>
+              <UncontrolledDropdown key={config.name} nav inNavbar>
                 <DropdownToggle nav caret>
                   {config.name}
                 </DropdownToggle>
                 <DropdownMenu>
                   {config.children.map(c => (
-                    <DropdownItem tag={Link} to={`${pathPrefix}${c.uri}`}>
+                    <DropdownItem key={c.uri} tag={Link} to={`${pathPrefix}${c.uri}`}>
                       {c.name}
                     </DropdownItem>
                   ))}


squat (approved):
Thanks :)))

prmsrswt:
I know I am late here but there's one thing that we missed in this PR.
After changing the React UI, you need to run `make assets` to generate a production build of the react code and get it included in the `bindata.go` file. This wasn't done here so the issue is not actually fixed in the Thanos binary we ship.

This is not a major problem and would be solved whenever someone runs the `make assets` command in future as the code is already there. Also, I guess we can use some kind of GH Action to warn us when the React code is modified but `bindata.go` isn't.

bwplotka:
It's actually as easy as adding `make assets` to `lint` test (:

bwplotka:
So it will detect modified code.

kunal-kushwaha:
> I know I am late here but there's one thing that we missed in this PR.
> After changing the React UI, you need to run `make assets` to generate a production build of the react code and get it included in the `bindata.go` file. This wasn't done here so the issue is not actually fixed in the Thanos binary we ship.
> 
> This is not a major problem and would be solved whenever someone runs the `make assets` command in future as the code is already there. Also, I guess we can use some kind of GH Action to warn us when the React code is modified but `bindata.go` isn't.

Ah my bad! I'll push it along with another commit for a different issue if not done already 👍

PR #3179, opened 2020-09-16
Store: If request ctx has an error we do not increment opsFailures counter

### LINKED ISSUE #3149: Store: Cancelled/Aborted GRPC Requests Increment thanos_objstore_bucket_operation_failures_total
**Thanos**: v0.13.0
**Prometheus**: 2.15.2
**Golang**: 1.14.1

**Object Storage Provider**:
GCS

**What happened**:
We noticed a continued increases in the `thanos_objstore_bucket_operation_failures_total` metric while not seeing errors in the log file. After looking at other metrics it appears that when a request from query is timed out/cancelled/aborted, the `thanos_objstore_bucket_operation_failures_total` metric increases. We also confirmed this from the GCS side by seeing an increased rate of `CANCELLED` API calls around the times we noticed that the bucket operation failures metric increased.

Screen Shot 2020-09-10 at 10 43 58 AM

`thanos_objstore_bucket_operation_failures_total` not to increase.

Run a large query from querier that will hit its query timeout.

I can include logs if desired but they just show normal block caching operations. Even in debug mode there is nothing about a failure.

ipstatic:
Signed-off-by: Jarod Watkins <jarod@42lines.net>

Resolves #3149 

## Changes

If a request is cancelled or times out we will no longer increment the ops failure counter.

### DIFF
--- a/pkg/objstore/objstore.go
+++ b/pkg/objstore/objstore.go
@@ -325,7 +325,7 @@ func (b *metricBucket) Attributes(ctx context.Context, name string) (ObjectAttri
 	start := time.Now()
 	attrs, err := b.bkt.Attributes(ctx, name)
 	if err != nil {
-		if !b.isOpFailureExpected(err) {
+		if !b.isOpFailureExpected(err) && ctx.Err() == nil {
 			b.opsFailures.WithLabelValues(op).Inc()
 		}
 		return attrs, err
@@ -340,7 +340,7 @@ func (b *metricBucket) Get(ctx context.Context, name string) (io.ReadCloser, err
 
 	rc, err := b.bkt.Get(ctx, name)
 	if err != nil {
-		if !b.isOpFailureExpected(err) {
+		if !b.isOpFailureExpected(err) && ctx.Err() == nil {
 			b.opsFailures.WithLabelValues(op).Inc()
 		}
 		return nil, err
@@ -360,7 +360,7 @@ func (b *metricBucket) GetRange(ctx context.Context, name string, off, length in
 
 	rc, err := b.bkt.GetRange(ctx, name, off, length)
 	if err != nil {
-		if !b.isOpFailureExpected(err) {
+		if !b.isOpFailureExpected(err) && ctx.Err() == nil {
 			b.opsFailures.WithLabelValues(op).Inc()
 		}
 		return nil, err
@@ -381,7 +381,7 @@ func (b *metricBucket) Exists(ctx context.Context, name string) (bool, error) {
 	start := time.Now()
 	ok, err := b.bkt.Exists(ctx, name)
 	if err != nil {
-		if !b.isOpFailureExpected(err) {
+		if !b.isOpFailureExpected(err) && ctx.Err() == nil {
 			b.opsFailures.WithLabelValues(op).Inc()
 		}
 		return false, err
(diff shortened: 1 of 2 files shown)


pkg/objstore/objstore.go
```diff
@@ -412,7 +412,7 @@ func (b *metricBucket) Delete(ctx context.Context, name string) error {
 
 	start := time.Now()
 	if err := b.bkt.Delete(ctx, name); err != nil {
-		if !b.isOpFailureExpected(err) {
+		if !b.isOpFailureExpected(err) && ctx.Err() == nil {
```
bwplotka, line 415:
```suggestion
		if !b.isOpFailureExpected(err) && ctx.Err() != context.Canceled {
```

bwplotka (commented):
I believe we need to just skip `context.Canceled`. 

I think timeouts might be a legit failures, no?

pkg/objstore/objstore.go
pracucci: Agree timeouts are legit failures, but this comment apply to any place, not just here, right? Also, is there a good reason why we don't we the `context.Canceled` check `isOpFailureExpected()`?

ipstatic: From what I have observed, cancelled requests can cause timeouts:
Screen Shot 2020-09-10 at 10 43 58 AM

If you notice, query cancels the request but store reports it as aborted. In my opinion that should be reported as cancelled, and not increase the error count.

ipstatic: or does Aborted != timeout?

ipstatic: > Agree timeouts are legit failures, but this comment apply to any place, not just here, right? Also, is there a good reason why we don't we the `context.Canceled` check `isOpFailureExpected()`?

If that is a better place for this then I have no problems making the change.

ipstatic:
Is there a way for me to view the circleci results without giving them access to my Github account?

pkg/objstore/objstore.go
bwplotka: `isOpFailureExpected` is really only for user specified function what to expect. I believe Cancel is valid to be ignored no matter what user specifies, no? (:

bwplotka:
You shold have acces to logs, no? https://circleci.com/gh/thanos-io/thanos/11705?utm_campaign=vcs-integration-link&utm_medium=referral&utm_source=github-build-link

ipstatic:
When I attempt to visit that URL, I get redirected to login via Github/Bitbucket and it requires read/write access to both public and private repos. I just didn't know if there was a way to view the logs "ready only" or not having to login at all.

bwplotka:
Try incognito mode - works just fine (:

pkg/objstore/objstore.go
pracucci: You're right @bwplotka 👍

ipstatic: I've updated the PR to use that context.

pracucci (approved):
LGTM

bwplotka (approved):
Amazing, thanks!

Some other solution at some point is to have error status as label (: 

Good enough for now +1

ipstatic:
Thank you both!

PR #3191, opened 2020-09-19
query-frontend: Update documentation for query-fronted in-memory cache

### LINKED ISSUE #3189: Document query frontend in-memory cache config better
There is an example in-memory cache config for query frontend in the document https://github.com/thanos-io/thanos/blob/master/docs/components/query-frontend.md#in-memory. However, it is not very clear how to configure these values, what is the correct format, etc.

AC, we should document 4 items:

1. The format of `max_size`. Based on the flag in Cortex https://github.com/cortexproject/cortex/blob/master/pkg/chunk/cache/fifo_cache.go#L43, `A unit suffix (KB, MB, GB) may be applied.`

2. Based on https://github.com/cortexproject/cortex/blob/master/pkg/chunk/cache/fifo_cache.go#L105, if `max_size` and `max_size_items` are not set, then the cache won't be created. We should document this case clearly.

3. If one of `max_size` and `max_size_items` is configured, then there is no limit for another field. 

4. Maybe we can also add a link for the example in kube-thanos https://github.com/thanos-io/kube-thanos/blob/master/examples/all/manifests/thanos-query-frontend-deployment.yaml#L50-L54

For more context, please check the slack thread https://cloud-native.slack.com/archives/CK5RSSC10/p1600454916006900.

daksh-sagar:
query-frontend: Add the documentation about the in memory cache config in  `docs/components/query-frontend.md`

Fixes #3189 

## Changes

Update documentation for the in-memory cache config.
## Verification

ran `make web-serve` locally

### DIFF
--- a/docs/components/query-frontend.md
+++ b/docs/components/query-frontend.md
@@ -58,6 +58,14 @@ config:
   max_size_items: 0
   validity: 0s
 ```
+`max_size: ` Maximum memory size of the cache in bytes. A unit suffix (KB, MB, GB) may be applied.
+
+**_NOTE:** If both `max_size` and `max_size_items` are not set, then the *cache* would not be created.
+
+If either of `max_size` or `max_size_items` is set, then there is not limit on other field.
+For example - only set `max_size_item` to 1000, then `max_size` is unlimited. Similarly, if only `max_size` is set, then `max_size_items` is unlimited.
+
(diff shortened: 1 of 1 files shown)


daksh-sagar:
@yeya24 This is what the docs look like after the changes. Please let me know if any changes are required, would be happy to do so.
docs_full

yeya24 (approved):
This looks perfect! Thanks for the contribution. @bwplotka @kakkoyun

kakkoyun (approved):
🥇 LGTM. Thanks a lot.

PR #3192, opened 2020-09-20
chore: Include half-open block maxtime for labels API

yeya24:

## Changes

TSDB block time range is half open, which is [b.meta.MinTime, b.meta.MaxTime). So when checking whether a block time is  overlapped with a time range, we can include the max time case. Added a `overlapsClosedInterval` method for the block which is the same as https://github.com/prometheus/prometheus/blob/b521612042/tsdb/block.go#L615.


prmsrswt (approved):
Looks good! LGTM.

bwplotka (approved):
Amazing :heart: 

Thanks for being so careful here, we missed this.

cc @kakkoyun

PR #3195, opened 2020-09-20
change from switch case to if else with a check on the url path suffix

Oghenebrume50:
Signed-off-by: Oghenebrume50 <raphlbrume@gmail.com>

## Changes
Changed the check on the exact matching part for the URL to a match on a specified suffix

## Verification
I queried the API using this `curl "http://localhost:9093/thanos/api/v1/query_range?query=time()&dedup=true&partial_response=true&start=1600269819.107&end=1600273419.107&step=14&max_source_resolution=0s&_=1600273386626"` and got an okay response. 

I am not exactly sure if this is the correct solution for this but I am open to better clarification

### DIFF
--- a/pkg/queryfrontend/roundtrip.go
+++ b/pkg/queryfrontend/roundtrip.go
@@ -5,9 +5,9 @@ package queryfrontend
 
 import (
 	"net/http"
-	"time"
 	"strings"
-	
+	"time"
+
 	"github.com/cortexproject/cortex/pkg/querier/frontend"
 	"github.com/cortexproject/cortex/pkg/querier/queryrange"
 	"github.com/cortexproject/cortex/pkg/util/validation"
--- a/pkg/queryfrontend/roundtrip.go
+++ b/pkg/queryfrontend/roundtrip.go
@@ -6,7 +6,8 @@ package queryfrontend
 import (
 	"net/http"
 	"time"
-
+	"strings"
+	
 	"github.com/cortexproject/cortex/pkg/querier/frontend"
 	"github.com/cortexproject/cortex/pkg/querier/queryrange"
 	"github.com/cortexproject/cortex/pkg/util/validation"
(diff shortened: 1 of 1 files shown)


pkg/queryfrontend/roundtrip.go
```diff
@@ -15,6 +15,7 @@ import (
 	"github.com/pkg/errors"
 	"github.com/prometheus/client_golang/prometheus"
 	"github.com/prometheus/client_golang/prometheus/promauto"
+	"strings"
```
yeya24, line 18: CI failed. Can you please group the import? Thanks!

Oghenebrume50: https://github.com/Oghenebrume50/thanos/blob/fcd80a2ec730eacc29f49a1076f802704e5d632c/pkg/queryfrontend/roundtrip.go#L6-L19

@yeya24 would this be a proper way to group the imports? I am new to Go  and Thanos and I am not so clear on how to do the grouping

Oghenebrume50:
I am not clear why the CI build is failing, I need some help to understand that properly.

pkg/queryfrontend/roundtrip.go
kakkoyun: Standard library imports should be grouped together and should be the top of the file. We have tooling for it to check on your development machine. Please take a look at the Makefile, you'll find them. (`goimports`)

Oghenebrume50: Thank you, I have done the corrections

bwplotka (approved):
This will do for now, thanks :+1:

bwplotka:
Thank you!

I seen some bug issue for this, you can link it in the PR so it's automatically closed when this is merge (:

Oghenebrume50:
Got it, thank you

PR #3204, opened 2020-09-22
mixin: Use sidecar's metric timestamp for healthcheck

hwoarang:
During prometheus updates the alert was firing because the metric was
initialized with a value of '0' before the first heartbeat was sent. As
such, the evaluation of the alert results into actually taking just the
value of time() into consideration which led to misleading information
about the health of the sidecar.
    
As the thanos_sidecar_last_heartbeat_success_time_seconds metric is
effectively just a timestamp that resets on new deployments, we can
simply wrap it around the timestamp() function which should return
almost the same value of the metric itself with the added benefit that
heartbeat resets will be ignored.
    
Signed-off-by: Markos Chandras <markos@chandras.me>

### DIFF
--- a/mixin/alerts/sidecar.libsonnet
+++ b/mixin/alerts/sidecar.libsonnet
@@ -39,11 +39,11 @@
           {
             alert: 'ThanosSidecarUnhealthy',
             annotations: {
-              description: 'Thanos Sidecar {{$labels.job}} {{$labels.instance}} is unhealthy for {{ $value }} seconds.',
+              description: 'Thanos Sidecar {{$labels.job}} {{$labels.pod}} is unhealthy for more than {{ $value }} seconds.',
               summary: 'Thanos Sidecar is unhealthy.',
             },
             expr: |||
-              time() - max(thanos_sidecar_last_heartbeat_success_time_seconds{%(selector)s}) by (job, instance) >= 600
+              time() - max(timestamp(thanos_sidecar_last_heartbeat_success_time_seconds{%(selector)s})) by (job,pod) >= 240
             ||| % thanos.sidecar,
             labels: {
               severity: 'critical',
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -15,6 +15,7 @@ We use _breaking :warning:_ to mark changes that are not backward compatible (re
 ### Added
 
 ### Fixed
+- [#3204](https://github.com/thanos-io/thanos/pull/3204) Mixin: Use sidecar's metric timestamp for healthcheck.
 
 ### Changed
 
--- a/examples/alerts/alerts.md
+++ b/examples/alerts/alerts.md
@@ -324,12 +324,12 @@ rules:
     severity: critical
 - alert: ThanosSidecarUnhealthy
   annotations:
-    description: Thanos Sidecar {{$labels.job}} {{$labels.instance}} is unhealthy
-      for {{ $value }} seconds.
+    description: Thanos Sidecar {{$labels.job}} {{$labels.pod}} is unhealthy for more
+      than {{ $value }} seconds.
     runbook_url: https://github.com/thanos-io/thanos/tree/main/mixin/runbook.md#alert-name-thanossidecarunhealthy
     summary: Thanos Sidecar is unhealthy.
   expr: |
-    time() - max(thanos_sidecar_last_heartbeat_success_time_seconds{job=~"thanos-sidecar.*"}) by (job, instance) >= 600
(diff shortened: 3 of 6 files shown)


mixin/alerts/sidecar.libsonnet
```diff
@@ -31,6 +31,7 @@
             expr: |||
               time() - max(thanos_sidecar_last_heartbeat_success_time_seconds{%(selector)s}) by (job, pod) >= 600
             ||| % thanos.sidecar,
+            'for': '2m'
```
kakkoyun, line 34: Hmmm we already checking if the last heartbeat is 5min ago and then with this it'll be another 2m. I think we should consider another way to prevent counter resets. WDYT?

hwoarang: @kakkoyun but the first alert is using a different metric right? If you think both serve the same purpose maybe we can remove the second one?

But to me the first one is checking the liveness of the sidecar and the second one the overall health. I also do not understand the comment about 7 minutes in total. Both alerts should be evaluated in parallel. So for this particular change we just wait 2 minutes to ensure that sidecar is fully up and had the time to send a heartbeat so we do not fire for no reason.

What do you think?

kakkoyun: I wasn't referring to the first one. The first one checking the liveness of the Prometheus that we run `Sidecar` as sidecar. So they serve different purpose.

What I'm refer is `600` in `time() - max(thanos_sidecar_last_heartbeat_success_time_seconds{%(selector)s}) by (job, pod) >= 600`, so this triggers an alert if the last heartbeat 600 seconds ago from the current time. Hence I said 5m + 2m. 

I'm just suggesting to use something more sophisticated(?) to prevent flappy alerts by catching the counter resets.
I don't have a concrete suggestion right now I have to check docs and think about a way to achieve it.

hwoarang: Oh I see. Yes in fact the delay would be 10m + 2m but I understand your point. I will look into it.

hwoarang: As the `thanos_sidecar_last_heartbeat_success_time_seconds` only contains a timestamp which is updated whenever the sidecar is healthy, what do you think of using the timestamp() function? something like

`time() - max(timestamp(thanos_sidecar_last_heartbeat_success_time_seconds{%(selector)s})) by (job,pod) >= 600`

This way we will check when the metric itself was updated last instead of what value it contains as we know that it always contains a timestamp value except when it resets. Its value and its timestamp attribute they should always be the same (again, except resets).

hwoarang: I have updated the code based on :point_up:

hwoarang:
Update PR to match latest master branch

@kakkoyun any thoughts here?

hwoarang:
@kakkoyun @bwplotka would it be possible to merge this? If not, do you have any feedback on how to bring it into mergeable state? :) Thank you

kakkoyun (approved):
LGTM

kakkoyun:
@hwoarang This looks good to me in principle, however, you need to generate docs and make sure this fulfils the expected behaviour with tests. CI already points out where it falls short. You can find all necessary task as Make actions.

kakkoyun (changes requested):
Requesting changes as commented above.

hwoarang:
> @hwoarang This looks good to me in principle, however, you need to generate docs and make sure this fulfils the expected behaviour with tests. CI already points out where it falls short. You can find all necessary task as Make actions.

@kakkoyun thank you for your input and apologies for taking a while to address your concerns. Tests are passing now as I had to refactor tests a little bit as we are now effectively testing for a different kind of alert. Please let me know your thoughts. Thank you

hwoarang:
The only tests that seems to fail in circle CI is `--- FAIL: TestBucketStore_ManyParts_e2e (0.20s)` which I believe is not related to this PR

kakkoyun:
Hey @hwoarang, enabled the auto-merge and approved. Please fix the conflicts for the Changelog.

kakkoyun (approved):

hwoarang:
> Hey @hwoarang, enabled the auto-merge and approved. Please fix the conflicts for the Changelog.

Hello @kakkoyun . Thank you so much. I will have this ready today.

kakkoyun:
@hwoarang This PR has introduced regressions around the pod to instance label renamings. It's my bad to mark as auto-merge. I didn't anticipate this. I'll send a consecutive PR to fix the issues.

hwoarang:
@kakkoyun really sorry that I missed that locally. I assumed the green CI was a good indication :)

PR #3207, opened 2020-09-23
Support snappy compression in query frontend cache

yeya24:

## Changes

This pr adds snappy compression support to the query frontend and bumps Cortex dependency to the latest master.


cmd/thanos/query_frontend.go
```diff
@@ -70,6 +70,9 @@ func registerQueryFrontend(app *extkingpin.App) {
 
 	cfg.CachePathOrContent = *extflag.RegisterPathOrContent(cmd, "query-range.response-cache-config", "YAML file that contains response cache configuration.", false)
 
+	cmd.Flag("query-range.compression", "Use compression in results cache. Supported values are: 'snappy' and '' (disable compression).").
```
bwplotka, line 73: Can we elaborate when it is used? On response? or when put in Memcached ?

bwplotka: Also, maybe `query-range.compression-type`?

bwplotka (approved):
:+1: 

LGTM just one question.

cmd/thanos/query_frontend.go
yeya24: Yes, it is used before storing the response to Memcached. It doesn't make sense to use it when only fifo-cache is enabled. I should document that.

I am okay with either `query-range.compression-type` or `query-range.compression`. The flag is named to `compression` in Cortex

bwplotka: > Yes, it is used before storing the response to Memcached. It doesn't make sense to use it when only fifo-cache is enabled. I should document that.

But it will work for other non FIFO backends as well, right? :thinking:

yeya24: Yes, it compresses the response and then sends it to the results cache backend. If it is fifo-cache, because snappy cache and fifo cache are in memory,  so not needed.

bwplotka: Technically it could be needed -> It might reduce memory consumed as well (: 
But yea, the question is what is available right now (:

yeya24: Yeah you are right. Sorry, my mistake. It is a tradeoff between CPU and Memory & network traffic

bwplotka: No need to be sorry - I think it's extremely useful. The only question is, do we need this for query range or some global cache option? (for future labels and instant)?

I think we need it for all so maybe some generic: 
```suggestion
	cmd.Flag("cache-compression", "Use compression in results cache. Supported values are: 'snappy' and '' (disable compression).").
```

prmsrswt (approved):
Nice! LGTM.

cmd/thanos/query_frontend.go
yeya24: I think the option can be global. But considering other caching configs, I am not sure. This is another problem and maybe we can have a separate issue to talk about it. The main question is, can we use different cache backends and different cache configs for each endpoint we want to cache? @bwplotka @kakkoyun 

For example, it is not very good to reuse the same cache config because the response sizes are different for range queries and instant queries. They might need different limits and different max_items configs.

yeya24:
Umm, this flaky test is new. Is it related to the new TSDB update? @bwplotka Can you please take a look? https://app.circleci.com/jobs/github/thanos-io/thanos/11833

```
Creating 1 50000-sample series in /tmp/testorbench-multitsdbseries256659744/0
Creating 1 50000-sample series in /tmp/testorbench-multitsdbseries256659744/1
Creating 1 50000-sample series in /tmp/testorbench-multitsdbseries256659744/2
Creating 1 50000-sample series in /tmp/testorbench-multitsdbseries256659744/3
Creating 50000 1-sample series in /tmp/testorbench-proxyseries345881914/0
Creating 50000 1-sample series in /tmp/testorbench-proxyseries345881914/1
Creating 50000 1-sample series in /tmp/testorbench-proxyseries345881914/2
Creating 50000 1-sample series in /tmp/testorbench-proxyseries345881914/3
Creating 5000 2-sample series in /tmp/testorbench-proxyseries834212689/0
Creating 5000 2-sample series in /tmp/testorbench-proxyseries834212689/1
Creating 5000 2-sample series in /tmp/testorbench-proxyseries834212689/2
Creating 5000 2-sample series in /tmp/testorbench-proxyseries834212689/3
Creating 1 50000-sample series in /tmp/testorbench-proxyseries876832380/0
Creating 1 50000-sample series in /tmp/testorbench-proxyseries876832380/1
Creating 1 50000-sample series in /tmp/testorbench-proxyseries876832380/2
Creating 1 50000-sample series in /tmp/testorbench-proxyseries876832380/3
--- FAIL: TestTSDBStore_Series_SplitSamplesIntoChunksWithMaxSizeOf120 (0.23s)
    prometheus_test.go:608: prometheus_test.go:608:
        
        	exp: 1093
        
        	got: 1092
        
        
Creating 2 300-sample series in /tmp/test011123234
Creating 2 300-sample series in /tmp/test011123234
Creating 2 300-sample series in /tmp/test354000153
Creating 2 300-sample series in /tmp/test354000153
```

CHANGELOG.md
```diff
@@ -21,6 +21,7 @@ We use *breaking :warning:* to mark changes that are not backward compatible (re
 - [#3154](https://github.com/thanos-io/thanos/pull/3154) Query Frontend: Add metric `thanos_memcached_getmulti_gate_queries_max`.
 - [#3146](https://github.com/thanos-io/thanos/pull/3146) Sidecar: Add `thanos_sidecar_prometheus_store_received_frames` histogram metric.
 - [#3147](https://github.com/thanos-io/thanos/pull/3147) Querier: Add `query.metadata.default-time-range` flag to specify the default metadata time range duration for retrieving labels through Labels and Series API when the range parameters are not specified. The zero value means range covers the time since the beginning.
+- [#3207](https://github.com/thanos-io/thanos/pull/3207) Query Frontend: Add `query-range.compression` flag to use compression in the query frontend cache.
```
kakkoyun, line 24:
```suggestion
- [#3207](https://github.com/thanos-io/thanos/pull/3207) Query Frontend: Add `cache-compression-type` flag to use compression in the query frontend cache.
```

kakkoyun (commented):
tiny nit, otherwise lgtm.

CHANGELOG.md
yeya24: Thanks for the catch! Please take a look again.

cmd/thanos/query_frontend.go
bwplotka: Totally agree. To me, something will be the same, some things different (:

bwplotka (approved):
Lgtm!

bwplotka:
Thanks! ♥️

PR #3228, opened 2020-09-25
doc: Correct CNCF project name

rytswd:
I noticed GitHub repo's description still says "CNCF Sandbox project" rather than "CNCF Incubating project" - but I could not propose a change for it via PR.

This PR is only to raise that point, but while at it, I also found the note of "CNCF Incubation project" being used, which I thought would make more sense with "CNCF Incubating project" as used in CNCF website https://www.cncf.io/projects/.

## Changes

`s/Incubation/Incubating/g`

## Verification

Via Netlify build

### DIFF
--- a/README.md
+++ b/README.md
@@ -19,7 +19,7 @@ Thanos is a set of components that can be composed into a highly available metri
 system with unlimited storage capacity, which can be added seamlessly on top of existing
 Prometheus deployments.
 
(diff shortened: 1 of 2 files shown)


yeya24 (approved):
LGTM. Thanks!

prmsrswt (approved):
Thanks!

prmsrswt:
Fixed the repo description as well :v:

Issue #3194, opened 2020-09-20
receive: relax quorum in case of 2 nodes


sepich:
**Thanos, Prometheus and Golang version used**:
v0.15.0

**Object Storage Provider**:
GCS

**What happened**:
Using single node thanos-receive and want to improve reliability a bit by adding the second node. Unfortunately due to 
https://github.com/thanos-io/thanos/blob/6fa6b4a18988ed11dd9392b399cde341ac36dc47/pkg/receive/handler.go#L386
I have to add 2 more nodes, to survive one node down.

**What you expected to happen**:
I understand the reasoning behind desire to have at least 2 copies of data, when you are SaaS and running on spot-instances.
But there could be other situations, for on-premises it could be "good enough" to have only single copy of data. It could be feasible to add a second node to make it possible to do maintenance with first one. But adding 2 more nodes could be "too much".
I propose to relax quorum to `1` for the case of `--receive.replication-factor=2` and add usual split-brain warning to the docs.
Because right now there is insist on majority at write-path, but read-path does not care about majority and would work with any amount of copies of data. Vertical compaction also works fine for 2 copies, which have not overlapping holes.

**How to reproduce it (as minimally and precisely as possible)**:
Here is docker-compose with 2 thanos-receive and prometheus with grafana
https://github.com/sepich/compose-playground/tree/thanos-quorum
Issue could be observed via `docker-compose stop thanos-receive-1`

**Full logs to relevant components**:
Prometheus logs:
```
ts=2020-09-20T13:07:30.335Z caller=dedupe.go:112 component=remote level=warn remote_name=5204bb url=http://thanos-receive-0:19291/api/v1/receive msg="Failed to send batch, retrying" err="server returned HTTP status 500 Internal Server Error: 2 errors: replicate write request, endpoint thanos-receive-1:10901: unexpected error, before quorum is reached: backing off forward request for endpoint thanos-receive-1:10901: target not available; replicate write request, endpoint thanos-receive-0:10901:"
```
Thanos receive-0 logs:
```
[...]

PR #3231, opened 2020-09-25
receive: relax quorum for 2 replicas

sepich:
Signed-off-by: Alexander Ryabov <alexander.ryabov@jetbrains.com>

## Changes

Fixes https://github.com/thanos-io/thanos/issues/3194

## Verification

Tests added. Also, we have working 2-node installation with the same patch.


prmsrswt (approved):
Looks good and well tested. @squat and @bwplotka any thoughts?

squat:
Bartek and I chatted about this last week. I don't think that this is not what exactly what we want. I think a more complete solution would be to implement a non-strict replication mode, where only one replica is required, no matter the replication factor. This would of course cover the case from this issue, and also all others

Issue #2577, opened 2020-05-08
Rule: add ability to send POST requests


GiedriusS:
Feature request description
---

Thanos Rule currently sends everything with GET requests. We should add an option to send the requests as POST. Accordingly, the logs should be improved to also log the POST (form) parameters.

Rationale
---

The alerting / recording rules' expressions might be very long and not fit into the URI.

yashrsharma44:
If no one is taking this issue, I am having a try :smile:

yashrsharma44:
Hey @GiedriusS !
I was trying to wrap my head around the use case of this feature, so can you elaborate on what are the requirements and acceptance criteria for this issue?
##
Also can you elaborate more on this?
>The alerting / recording rules' expressions might be very long and not fit into the URI.

GiedriusS:
> Hey @GiedriusS !
> I was trying to wrap my head around the use case of this feature, so can you elaborate on what are the requirements and acceptance criteria for this issue?
> 
> Also can you elaborate more on this?
> 
> > The alerting / recording rules' expressions might be very long and not fit into the URI.

There's no need for formal acceptance criteria - it's not work for us. At least for the majority of us.

Different implementations of HTTP browsers/servers have different limits on URL lengths. For example: https://technomanor.wordpress.com/2012/04/03/maximum-url-size/. How it stands right now, Thanos Rule encode everything into the URL when sending a request:

https://github.com/thanos-io/thanos/blob/2262f2601a2b04d0be7a951d1d962e698520c44e/pkg/promclient/promclient.go#L339

The recording or alerting rule might be too long and then it will get truncated when sending a request. That's why support for encoding request parameters via POST on `/api/v1/query` (and others) has been added. So:

* A new parameter needs to appear that controls whether GET or POST is used. We could theoretically convert everyone to use GET but very old versions do not support POST;
* The documentation needs to have verbiage about this;
* The tests need to be improved to test out this case.

GiedriusS:
Still very valid & help welcome!

jlr52:
@GiedriusS 

Hi I would love to contribute on this.

Does the ask only apply to `promclient.go` or does it apply to all queries in the repo that have post supported  in https://github.com/thanos-io/thanos/blob/0aa297bd00816ed5efe5d6dea117cb7059a3d7b1/pkg/query/api/v1.go like the following

```
	r.Get("/query", instr("query", api.query))
	r.Post("/query", instr("query", api.query))
[...]

PR #3233, opened 2020-09-26
First attempt to add ability to send POST requests in promclient

mritunjaysharma394:
Signed-off-by: Mritunjay Sharma <mritunjaysharma394@gmail.com>

## Changes

In file `thanos-io/thanos/pkg/promclient/promclient.go`, I added a new function
that is trying to handle POST requests. 
This is the first draft, feedback appreciated to
get a better head-start.

This PR is the first attempt to handle https://github.com/thanos-io/thanos/issues/2577

## Verification

Not tested yet, it is the first draft to know if the right approach is being
followed or not and to seek more clarity on doubts.

### DIFF
--- a/pkg/promclient/promclient.go
+++ b/pkg/promclient/promclient.go
@@ -122,6 +122,36 @@ func (c *Client) get2xx(ctx context.Context, u *url.URL) (_ []byte, _ int, err e
 	return body, resp.StatusCode, nil
 }
 
+//First attempt to add ability to send POST requests [issue #2577]
+// Doubt on which piece of code required to handle parameter/flag that
+// will give the option to choose between GET and POST
+func (c *Client) post2xx(ctx context.Context, u *url.URL, skipHead bool) (_ []byte, _ int, err error) {
+	req, err := http.NewRequest(
+		http.MethodPost,
+		u.String(),
+		strings.NewReader(url.Values{"skip_head": []string{strconv.FormatBool(skipHead)}}.Encode()),
+	)
+	if err != nil {
+		return nil, 0, errors.Wrap(err, "create POST request")
+	}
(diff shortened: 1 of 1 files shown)


pkg/promclient/promclient.go
```diff
@@ -122,6 +122,36 @@ func (c *Client) get2xx(ctx context.Context, u *url.URL) (_ []byte, _ int, err e
 	return body, resp.StatusCode, nil
 }
 
+//First attempt to add ability to send POST requests [issue #2577]
```
kakkoyun, line 125: nit
```suggestion
// First attempt to add ability to send POST requests [issue #2577]
```

pkg/promclient/promclient.go
```diff
@@ -122,6 +122,36 @@ func (c *Client) get2xx(ctx context.Context, u *url.URL) (_ []byte, _ int, err e
...
+// Doubt on which piece of code required to handle parameter/flag that
+// will give the option to choose between GET and POST
+func (c *Client) post2xx(ctx context.Context, u *url.URL, skipHead bool) (_ []byte, _ int, err error) {
```
kakkoyun, line 128: I'd love to see this method used in action. For example, I'd probably want to pass the content that I'd like to send.

kakkoyun (commented):
Looking good. Thanks for doing it. 
I have left some comments.

Rather than proposing a code piece that hasn't been used it would be great to see it in action. You could add some test and example. But the ideally you could directly try to fix #2577 in a PR so that we see this code in action.

GiedriusS:
Thanks for the review! I thought it should be easier - as simple as adding another parameter to the function which does the call to pass either GET or POST. I hope you don't mind if I take the implementation of this feature over.

PR #3234, opened 2020-09-26
ui: Fix assets not loading when `--web.prefix-header` is used

### LINKED ISSUE #3226: [Query] UI served behind reverse proxy on a subpath broken on 0.15.0
Thanos version: v.15.0

I updated from 0.14.0 to 0.15.0 and thanos query UI in not working correctly. I run thanos behind an nginx reverse proxy under a subpath. The way I've been running it so far is using the `--web.prefix-header=X-Forwarded-Prefix`

Here is my ingress:

```yaml
apiVersion: extensions/v1beta1
kind: Ingress
metadata:
  annotations:
    nginx.ingress.kubernetes.io/rewrite-target: /$1
    nginx.ingress.kubernetes.io/x-forwarded-prefix: /thanos
  name: thanos-query-http
spec:
  rules:
  - host: foo.bar
    http:
      paths:
      - backend:
          serviceName: thanos-query-http
          servicePort: 10902
        path: /thanos/?(.*)
```
Here is what I get on v0.15.0:

image

prmsrswt:

## Changes
- Fix UI not working when using `--web.prefix-header` to set external prefix. Fixes #3226 

This was caused by incorrect handling of the prefix value we got from the header. If the header had `/` as a suffix, we were still adding another `/` as a suffix to it, resulting in `//prefix` which translates to an external network request.

## Verification
Tested it using a reverse proxy with prefix-stripping and the header `X-Forwarded-Prefix` set.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -13,7 +13,7 @@ We use *breaking :warning:* to mark changes that are not backward compatible (re
 
 ### Fixed
 
-- [#3234](https://github.com/thanos-io/thanos/pull/3234) UI: Fix assets not loading when `--web.external-header` is used.
+- [#3234](https://github.com/thanos-io/thanos/pull/3234) UI: Fix assets not loading when `--web.prefix-header` is used.
 
 ### Added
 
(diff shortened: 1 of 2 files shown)


squat (approved):

amrmahdi:
I verified this fix is working. @prmsrswt @bwplotka Can we release this change?

PR #3235, opened 2020-09-26
ui: Fix labelSets in Stores page

prmsrswt:

#2783 modified the `LabelSets` field of struct `StoreStatus` in `/pkg/query/storeset.go`, which broke the `stores.html` template for old UI and it also changed the structure of it's marshaled JSON, which resulted in the crashing of Stores page in new UI as well.

## Changes
- Update the Stores page in both old UI and new UI to match the modified struct.

## Verification
All React unit tests passed with the new response. Also tested manually.


bwplotka (approved):
Thanks! ♥️♥️♥️

Ah those non type safe script languages and templates ):

Issue #3221, opened 2020-09-24
Block Viewer: Show size of the blocks in the UI


onprem:
The size of a block is a useful metric. We can show things like the overall size of a block, the size of it's index, the total size occupied by all blocks, etc. in the Block Viewer itself.

aribalam:
I see that Prometheus has an API exposed to query currently stored TSDB Blocks (via `/api/query/v1/blocks`). But I am unable to find details of it in the Prometheus or Thanos docs. I am guessing the above required data will be queried using that API? 
Is there any relevant docs or resources where I can find details of such an API.

roidelapluie:
Note that recently Prometheus BlockReader gained Size() function.

bwplotka:
> I see that Prometheus has an API exposed to query currently stored TSDB Blocks (via /api/query/v1/blocks). But I am unable to find details of it in the Prometheus or Thanos docs. I am guessing the above required data will be queried using that API?
Is there any relevant docs or resources where I can find details of such an API.

We have other APIs for this and yes, it should be better documented - only block viewer is using it.

> Note that recently Prometheus BlockReader gained Size() function.

This is great but not very helpful for this issue, as we store all blocks in object storage (:

roidelapluie:
> > I see that Prometheus has an API exposed to query currently stored TSDB Blocks (via /api/query/v1/blocks). But I am unable to find details of it in the Prometheus or Thanos docs. I am guessing the above required data will be queried using that API?
> > Is there any relevant docs or resources where I can find details of such an API.
> 
> We have other APIs for this and yes, it should be better documented - only block viewer is using it.
> 
> > Note that recently Prometheus BlockReader gained Size() function.
> 
> This is great but not very helpful for this issue, as we store all blocks in object storage (:

object storage does not implement the BlockReader interface?

bwplotka:
Not directly.

roidelapluie:
yes

zerefwayne:
Hi @roidelapluie @prmsrswt I would like to take this up if it's still unresolved.

saswatamcode:
@prmsrswt Is this issue still open? Would love to take this up!

douglascamata:
Closing this as done with PR #7233, related to #7205.

PR #3238, opened 2020-09-27
[WIP] compactor: Sets block size compaction limit

### LINKED ISSUE #3068: Limit size of blocks to X bytes on compaction.
This is to ensure the scalability of the compactor.

Tricky part: How to shard block, what label to put, or maybe vertical compaction?

someshkoli:
Currently WIP and aims to resolve #3068 
Partially solves #3221 as block size is required here too. 

### DIFF
--- a/pkg/block/block.go
+++ b/pkg/block/block.go
@@ -229,8 +229,8 @@ type Size struct {
 	ChunkSize int64 `json:"chunkSize"`
 }
 
-// GetSize Returns the size of a block.
-func GetSize(ctx context.Context, logger log.Logger, bkt objstore.Bucket, id ulid.ULID) (Size, error) {
+// GetSize Returns the size of a block in bytes.
+func GetSize(ctx context.Context, bkt objstore.Bucket, id ulid.ULID) (Size, error) {
 	var chunkSize int64
 	indexAttr, err := bkt.Attributes(ctx, path.Join(id.String(), "index"))
 	if err != nil {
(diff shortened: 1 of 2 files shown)


bwplotka:
Amazing, thanks for helping on this - we need help here.

Wonder what's the plan here? Detecting size is important, but how we ensure proper sharding of block? (:

someshkoli:
@bwplotka ,As for now I'm done limiting compaction acc to the block index size. 
For sharding I'm planning to do the following, would improvise on this, also let me know if the implementation is wrong 
-  compact a blocks 
-  check final block size (newly generated)
  - if greater than x(64 gib for now)
    - partition it in two roughly equal parts ( do this recursively till all the block sizes are < x )
    - write new blocks and move forward
 - if less than 64 move forward as usual

This way our blocks are compacted to its maximum and also we prevent it from exceeding the limit

roidelapluie:
Note that recently Prometheus BlockReader gained Size() function.

someshkoli:
> Note that recently Prometheus BlockReader gained Size() function.

yeah I did pick that up (disscussed here: #3254) but having some trouble in testing the changes on the same.

bwplotka:
I am moving with https://github.com/thanos-io/thanos/issues/3068 and will probably use the code from this PR if you don't mind :hugs:

someshkoli:
> I am moving with https://github.com/thanos-io/thanos/issues/3068 and will probably use the code from this PR if you don't mind :hugs: 

Yes sure ✌️

bwplotka:
Awesome, closing for now then, but will use your code (:

PR #3239, opened 2020-09-27
React-app: Handle routes on the client itself instead of `ui.go`

### LINKED ISSUE #3216: Remove the need for specifying React UI's paths in `ui.go`
We are using client-side routing in the new React UI. This means that for all the client-side routes (ex. `/new/graph`, `/new/stores`, etc.) in the new UI, we need to serve the same `index.html`. Right now we maintain an array of paths in the Go code, named `reactAppPaths`. If the incoming request's path is found in this array, we directly serve the `index.html` file, otherwise, we return `404` or a file if one exists with the given path.

Instead of maintaining the array, we can replace `404` with the `index.html` file. This is a common model in serving single-page applications. If the incoming request corresponds to a static asset, return that asset, otherwise return the `index.html` file. If the given route doesn't exist, then we can show a 404 page directly on the client-side, using a default route in React.

daksh-sagar:
Fixes #3216 

## Changes

Remove `reactPaths` from ui.go
Create `NotFound` component in the react-app and render it when a route is not found.


daksh-sagar:
@prmsrswt

pkg/ui/react-app/src/pages/notFound/NotFound.tsx
```diff
@@ -0,0 +1,9 @@
+import React, { FC } from 'react';
```
prmsrswt, line 1: Can you move this component to `src/thanos/pages`? All of our custom pages are there.

daksh-sagar: Ofcourse !
Is everything else fine ?

prmsrswt: Yep, everything else seems to be alright. I'll test it out locally first and then let's get this merged.

prmsrswt:
Sorry for not responding fast enough. We already have these changes in the codebase so I am going to close this PR. That being said we still appreciate your effort (very much) and will try to not waste it like this in the future.

PR #3245, opened 2020-09-28
query-frontend: Extract org id from from request headers

### LINKED ISSUE #3177: query-frontend: populate org_id in slow query logs
The OrgId passed to Cortex is hard-coded to "fake":
https://github.com/thanos-io/thanos/blob/5aac47760359503358191caeb73780f87aa94e38/cmd/thanos/query-frontend.go#L203

When logs are produced to the slow query log, it results in logs which contain `org_id=fake`, e.g:
```level=info ts=2020-09-16T13:33:14.670646Z caller=frontend.go:206 org_id=fake msg="slow query detected" method=GET host=localhost:10902 path=/api/v1/query_range time_taken=890.912705ms ...```

When using Cortex Query Frontend (which populates org_id based on the contents of the X-Scope-OrgID header), there was extra value in these logs.  The org_id field is useful for identifying sources of problematic queries which take a long time to execute.

Does anyone have any thoughts on populating that org_id log field with the contents of a request header?

### PR DESCRIPTION
**Pull request by neilfordyce**
Signed-off-by: Neil Fordyce <neilfordyce@skyscanner.net>

## Changes
1. Add `query-frontend.org-id-header` flag to specify HTTP header(s) to populate slow query log org_id field: Fixes #3177 

## Verification
Executed query-frontend pointed at a local querier with this configuration
`thanos query-frontend --http-address=127.0.0.1:9093 --query-frontend.downstream-url=http://localhost:9090 --query-frontend.log-queries-longer-than="-1s" --query-frontend.org-id-header="x-grafana-user" --query-frontend.org-id-header="x-scope-orgid"`

Ran query with expected headers
```
curl --request GET \
  --url 'http://localhost:9093/api/v1/query_range?query=up&dedup=true&partial_response=true&start=1594422000&end=1594508400&step=600&max_source_resolution=0s' \
  --header 'x-grafana-user: grafana-username'
```

grafana-username shows in org_id field of query frontend logs  
```
level=info ts=2020-09-28T13:11:22.645878Z caller=frontend.go:220 org_id=grafana-username msg="slow query detected" method=GET host=localhost:9093 path=/api/v1/query_range time_taken=811.734669ms param_dedup=true param_partial_response=true param_start=1594422000 param_end=1594508400 param_step=600 param_max_source_resolution=0s param_query=up
```

Further testing summary
| Input headers | org_id field contents | comment |
| -------------- | -------------------- | ---------- | 
| --header 'x-grafana-user: grafana-username' --header 'X-Scope-OrgId: another-org-id' | grafana-username | check first arg specified has precedence |
| --header 'X-Scope-OrgId: another-org-id' | another-org-id | check second arg works if first arg header not present |
| no headers | anonymous | check fallback if no headers match |

### DIFF
--- a/cmd/thanos/query_frontend.go
+++ b/cmd/thanos/query_frontend.go
@@ -242,10 +242,7 @@ func runQueryFrontend(
 	return nil
 }
 
-func extractOrgId(
-	conf *queryFrontendConfig,
-	r *http.Request,
-) string {
+func extractOrgId(conf *queryFrontendConfig, r *http.Request) string {
 	for _, header := range conf.orgIdHeaders {
 		headerVal := r.Header.Get(header)
 		if headerVal != "" {
--- a/cmd/thanos/query_frontend.go
+++ b/cmd/thanos/query_frontend.go
@@ -116,7 +116,8 @@ func registerQueryFrontend(app *extkingpin.App) {
 	cmd.Flag("query-frontend.log-queries-longer-than", "Log queries that are slower than the specified duration. "+
 		"Set to 0 to disable. Set to < 0 to enable on all queries.").Default("0").DurationVar(&cfg.CortexFrontendConfig.LogQueriesLongerThan)
 
-	cmd.Flag("query-frontend.org-id-header", "Request header names used to set the org id field in the slow query log (repeated flag). "+
+	cmd.Flag("query-frontend.org-id-header", "Request header names used to identify the source of slow queries (repeated flag). "+
+		"The values of the header will be added to the org id field in the slow query log. "+
 		"If multiple headers match the request, the first matching arg specified will take precedence. "+
 		"If no headers match 'anonymous' will be used.").PlaceHolder("<http-header-name>").StringsVar(&cfg.orgIdHeaders)
 
--- a/cmd/thanos/query_frontend.go
+++ b/cmd/thanos/query_frontend.go
@@ -36,6 +36,7 @@ import (
 type queryFrontendConfig struct {
 	http httpConfig
 	queryfrontend.Config
+	orgIdHeaders         []string
 }
 
 func registerQueryFrontend(app *extkingpin.App) {
@@ -114,6 +115,10 @@ func registerQueryFrontend(app *extkingpin.App) {
 
 	cmd.Flag("query-frontend.log-queries-longer-than", "Log queries that are slower than the specified duration. "+
 		"Set to 0 to disable. Set to < 0 to enable on all queries.").Default("0").DurationVar(&cfg.CortexFrontendConfig.LogQueriesLongerThan)
+
+	cmd.Flag("query-frontend.org-id-header", "Request header names used to set the org id field in the slow query log (repeated flag). "+
+		"If multiple headers match the request, the first matching arg specified will take precedence. "+
+		"If no headers match 'anonymous' will be used.").PlaceHolder("<http-header-name>").StringsVar(&cfg.orgIdHeaders)
 
 	cmd.Flag("log.request.decision", "Request Logging for logging the start and end of requests. LogFinishCall is enabled by default. LogFinishCall : Logs the finish call of the requests. LogStartAndFinishCall : Logs the start and finish call of the requests. NoLogCall : Disable request logging.").Default("LogFinishCall").EnumVar(&cfg.RequestLoggingDecision, "NoLogCall", "LogFinishCall", "LogStartAndFinishCall")
 
@@ -199,6 +204,7 @@ func runQueryFrontend(
 
 		instr := func(f http.HandlerFunc) http.HandlerFunc {
 			hf := http.HandlerFunc(func(w http.ResponseWriter, r *http.Request) {
+				orgId := extractOrgId(cfg, r)
 				name := "query-frontend"
 				api.SetCORS(w)
 				ins.NewHandler(
@@ -213,7 +219,7 @@ func runQueryFrontend(
 						),
 					),
 					// Cortex frontend middlewares require orgID.
-				).ServeHTTP(w, r.WithContext(user.InjectOrgID(r.Context(), "fake")))
+				).ServeHTTP(w, r.WithContext(user.InjectOrgID(r.Context(), orgId)))
 			})
 			return hf
 		}
@@ -234,4 +240,17 @@ func runQueryFrontend(
 	level.Info(logger).Log("msg", "starting query frontend")
 	statusProber.Ready()
 	return nil
+}
+
+func extractOrgId(
+	conf *queryFrontendConfig,
+	r *http.Request,
+) string {
+	for _, header := range conf.orgIdHeaders {
+		headerVal := r.Header.Get(header)
+		if headerVal != "" {
+			return headerVal
+		}
+	}
+	return "anonymous"
 }
--- a/cmd/thanos/query_frontend.go
+++ b/cmd/thanos/query_frontend.go
@@ -36,7 +36,7 @@ import (
 type queryFrontendConfig struct {
 	http httpConfig
 	queryfrontend.Config
-	orgIdHeaders         []string
+	orgIdHeaders []string
 }
 
 func registerQueryFrontend(app *extkingpin.App) {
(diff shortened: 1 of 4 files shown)

### CODE REVIEW

**Comment by yeya24** [comment]
Thanks for the work. The CI failure is not related.

**Comment by krasi-georgiev** [comment]
@neilfordyce can you explain when is this needed/useful?

File: cmd/thanos/query_frontend.go [hunk 1]
```diff
@@ -190,3 +196,16 @@ func runQueryFrontend(
 	statusProber.Ready()
 	return nil
 }
+
+func extractOrgId(
+	conf *queryFrontendConfig,
+	r *http.Request,
+) string {
```
yeya24, line 203: One small nit, we can inline this.

cmd/thanos/query_frontend.go
```diff
@@ -82,6 +83,10 @@ func registerQueryFrontend(app *extkingpin.App) {
 	cmd.Flag("query-frontend.log-queries-longer-than", "Log queries that are slower than the specified duration. "+
 		"Set to 0 to disable. Set to < 0 to enable on all queries.").Default("0").DurationVar(&cfg.CortexFrontendConfig.LogQueriesLongerThan)
 
+	cmd.Flag("query-frontend.org-id-header", "Request header names used to set the org id field in the slow query log (repeated flag). "+
```
yeya24, line 86: I am thinking about having a default org-id-header `X-Scope-OrgId` and users can also specify other headers as well. WDYT?

yeya24 (commented):
Some small nits. Overall it looks very good! Would you mind also documenting this?

neilfordyce:
Absolutely, apologies for the missing context.  We find it helpful to identify users and services who are creating (slow) queries.  In that way, we can find the source of slow queries and alter the queries, e.g. if they are continually causing memory issues for queriers.  It's a bit of a stopgap to help reduce problems large queries cause with high memory use; but more generally, it's useful to be able to see who is running what queries.

cmd/thanos/query_frontend.go
neilfordyce: Yeah, sounds like a sensible default :+1:.  Will that help with compatibility with Cortex things perhaps?

krasi-georgiev:
Why not use some of the standard HTTP headers to identify the source that creates these slow queries?
Host, Origin, Referer, Via?

neilfordyce:
We definitely could do that, but the best header to use would likely vary depending on the client.  E.g. most people interacting with our Thanos setup are logged in Grafana users, so it's handy to identify them via X-Grafana-User.   Other users of Thanos might not be so interested in the specific user (or they might be anonymous in Grafana) so they could fallback to another standard header.

krasi-georgiev:
I would let one of the maintainers(cc @bwplotka, @kakkoyun ) decide on that, but I would say let's add something basic for now that solves your use-case and if needed can revisit later to make it configurable. 

Would the `origin` header solve your use case?

neilfordyce:
I thought origin might be enough, but it appears like Grafana doesn't forward the origin when it's configured to proxy requests through the data source proxy.  

This is the complete set of headers I can see from a request coming from the Prometheus data source.  x-forwarded-for might be helpful, but an IP address isn't as helpful as a user name. ```{ host: 'echo-proxy:80', 'user-agent': 'Grafana/7.0.6', accept: 'application/json, text/plain, */*', 'accept-encoding': 'gzip, deflate', 'accept-language': 'en-US,en;q=0.5', 'x-forwarded-for': '192.168.48.1, 192.168.48.1', 'x-grafana-nocache': 'true', 'x-grafana-org-id': '1', 'x-grafana-user': 'neil' }```

If I change the datasource to go direct from the Browser there is an origin header, so we'd be able to see slow requests were coming from Grafana. But again probably not as helpful as seeing the user name in the logs.
```{ host: 'localhost:8090', 'user-agent': 'Mozilla/5.0 (Macintosh; Intel Mac OS X 10.15; rv:80.0) Gecko/20100101 Firefox/80.0', accept: 'application/json, text/plain, */*', 'accept-language': 'en-US,en;q=0.5', 'accept-encoding': 'gzip, deflate', referer: 'http://localhost:3000/datasources/edit/6/', origin: 'http://localhost:3000', connection: 'keep-alive' }```

**Comment by krasi-georgiev** [comment]
hm, yeah I see what you mean.

But isn't an IP a more unique ID here? It is not unlikely that many grafana installations will use the admin username which will cause the same log label from different sources.

**Comment by krasi-georgiev** [comment]
I see host is present in both modes here

**Comment by krasi-georgiev** [comment]
btw could this be an issue with Grafana not adding the origin header?

https://github.com/grafana/grafana/issues/22869

**Comment by neilfordyce** [comment]
We only allow logged in users in our Grafana setup, so the x-grafana-user header is always set and gives us exactly what we want.  IP address could be helpful in some setups, but for others it might only reveal the address of a NAT which would be the same for all users.  My thought was, different headers will make sense in different contexts.  This might be why Cortex have the user id header fixed to X-Scope-OrgId, then leaves it up to a reverse proxy to allow users to fill out that header whatever logic they want from the request.

**Comment by krasi-georgiev** [comment]
Aaah I see what you mean using IP or Origin header will always be the same when the request comes from the same grafana instance. Using the grafana username will indeed give you a better understanding of which use is sending these heavy queries.

File: cmd/thanos/query_frontend.go [hunk 3]
```diff
@@ -82,6 +83,10 @@ func registerQueryFrontend(app *extkingpin.App) {
...
+		"If multiple headers match the request, the first matching arg specified will take precedence. "+
```
krasi-georgiev, line 87: I don't understand this bit? Can you add test for this case to make it a bit more clear?

neilfordyce: Thanks, yeah this sounded clearer in my head; I'll try to make the behaviour clearer in the help string.  The idea is that if `--query-frontend.org-id-header` is used multiple times, we try to extract a header from the request in order the command line args were specified.

There are test cases to check the precedence behaviour in query_frontend_test.go, e.g. we check that when both `--query-frontend.org-id-header=x-server-id` and `query-frontend.org-id-header=x-grafana-user` are specified that org id is set to the value of the x-server-id header.    However, happy to split that into a few test functions with different names if it would be clearer what's being tested?

Also, I noticed that this file has changed a bit since writing it and after rebasing it seems bit out of place.  Considering moving the new logic into pkg/queryfrontend. WDYT?

docs/components/query-frontend.md
```diff
@@ -161,6 +161,12 @@ Flags:
                               Log queries that are slower than the specified
                               duration. Set to 0 to disable. Set to < 0 to
                               enable on all queries.
+      --query-frontend.org-id-header=<http-header-name> ...
```
neilfordyce, line 164: @yeya24  - thanks for the feedback.  Regarding docs, I'm considering adding something to explain the new flag in the `slow-query-log` section of query-frontend.md.  And probably some godoc to explain the new function.  Is that the kind of thing you had in mind?

cmd/thanos/query_frontend.go
krasi-georgiev: Thanks for clarifying, now looking at the test it is clear.
I think the test location is fine, but if you prefer can move it.

krasi-georgiev:
LGTM

docs/components/query-frontend.md
yeya24: You can simply add this description to the flag and I think it would be clear :+1:

yeya24:
Sorry for the late response. I think this feature is nice to have. @neilfordyce Can you please fix the conflicts?

docs/components/query-frontend.md
neilfordyce: Not sure I follow.  I think it's already like that.  This description is the one generated from the flag in cmd/thanos/query_frontend.go.

neilfordyce:
Thanks for the bump. This slipped my mind.  Sure thing, I've rebased and fixed the conflicts.  The test failure seems to be for unrelated reasons.

docs/components/query-frontend.md
yeya24: Yes, I mean you can add something like `This ord id will be added to the slow-query-log` to the flag description directly and then regenerate the doc

neilfordyce: Made an update to the flag description.  Hope it's clearer, but open to suggestions for a better way to describe it.

yeya24 (approved):
Thanks. I think it is clear. @bwplotka @kakkoyun Can you please also take a look at this pr?

kakkoyun (approved):
🥇 LGTM. Thanks for your contribution.

PR #3246, opened 2020-09-28
website: Changed download button to releases to be clearer.

bwplotka:


thisisobate (approved):
I think it's cool for us to implement this change. Although I tried checking other open source projects website and I found out they use "downloads" instead. 

I think using "releases" clearly defines a user's expectation so...LGTM! 👍

prmsrswt (approved):
LGTM! CI is failing but it shouldn't as the changes are totally unrelated :man_shrugging:.

PR #3248, opened 2020-09-28
Fix bucket replicate file path in dashboards doc

yeya24:

## Changes

The file name updated in https://github.com/thanos-io/thanos/pull/2781 and the doc should be updated as well.

Caught by our markdown checker.  https://github.com/thanos-io/thanos/pull/3245/checks?check_run_id=1177044878


prmsrswt (approved):
Nice. After so many false positives, it's refreshing to see that it actually caught something :smile:

metalmatze (approved):
Very cool!

PR #3251, opened 2020-09-28
Bucket Web UI: Added web.external-prefix to Bucket Web UI

### LINKED ISSUE #3211: Thanos bucket Web: Provide support to flag web.route-prefix which help to route UI properly
<!--
Template relevant to bug reports only!

Keep issue title verbose enough and add prefix telling
about what components it touches e.g "query:" or ".*:"
-->

<!--
In case of issues related to exact bucket implementation, please ping corresponded maintainer from list here: https://github.com/thanos-io/thanos/blob/master/docs/storage.md
-->

Using official docker image: thanosio/thanos:v0.15.0

<!--
Output of "thanos --version" or docker image:tag used.
(Double-check if all deployed components/services have expected versions)

If you are using custom build from master branch, have you checked out the tip of the master?
-->

**Object Storage Provider**: Ceph storage blocks

**What happened**: 
I am using Thanos tools bucket web with `--web.external-prefix=/thanos/bucket`. Once I hit URL http://localhost:9090/thanos/bucket (k8s service being forwarded to 9090 from 10902 internally), I get 404. When I hit http://localhost:9090 then it shows HTML links only, and when I click on any link it routes me via a prefix path which is okay but again 404. When going through the documentation, I found 
```
      --web.external-prefix=""  Static prefix for all HTML links and redirect
                                URLs in the bucket web UI interface. Actual
                                endpoints are still served on / or the
                                web.route-prefix. This allows thanos bucket web
                                UI to be served behind a reverse proxy that
                                strips a URL sub-path.
```
Flag`web.route-prefix` is mentioned but bucket web does not support it. 
I faced a similar issue with the querier and it got resolved when I used `web.route-prefix` flag. 
When I am trying to use this flag for Thanos tools bucket web, it says there is not support for the flag `web.route-prefix`. 

Bucket web UI should be loaded behind the proxy and this can be done by providing support of `web.route-prefix` flag which is missing as of now. 

1
```
      containers:
        - name: thanos-bucketweb
          image: thanosio/thanos:v0.15.0
          imagePullPolicy: IfNotPresent
          args:
            - tools
            - bucket 
            - web  
            - --log.level=info
            - --http-address=0.0.0.0:10902
            - --web.external-prefix=thanos/bucket
            - --refresh=30m
            - --objstore.config-file=/etc/thanos-store/ceph-object-storage.yaml
```
2 http://localhost:9090/thanos/bucket `404 page not found `
3 http://localhost:9090 `HTML UI is displayed and then 404 page not found after clicking a link`

After providing  '--web.route-prefix': 
```
│ Error parsing commandline arguments: [/bin/thanos tools bucket web --log.level=info --http-address=0.0.0.0:10902 --web.route-prefix=thanos/bucket --web.external-prefix=thanos/bucket --refresh=30m --objstore.config-file=/etc/thanos-store/ceph-object-storage.yaml]: unknown long flag '--we │
│ thanos: error: unknown long flag '--web.route-prefix'                                                                                                                                                                                                                                           │
│ stream closed
```
<!--
Uncomment if you would like to post collapsible logs:

<details>Logs
<p>

```

```

</p>
</details>
-->

I see a similar issue with compact web #2727. If it is fixed then bucket web UI can be fixed easily. 
<!--
Uncomment and fill if you use not casual environment or if it might be relevant.

**Environment**:
- OS (e.g. from /etc/os-release):
- Kernel (e.g. `uname -a`):
- Others:

-->

aribalam:

## Changes

Fixes #3211 

The router was not prefixed with the `webExternalPrefix` string before the routes for the Bucket Web UI was created. Moreover, the `bucket-menu.html` template was missing the `pathPrefix` in the link to new UI. 

The question remains, as to the Bucket Web UI should also support the `web.route-prefix` flag?

Also, the code for handling external prefixes and route prefix were becoming sort of repeating, with the same thing happening in querier too. Maybe we could create a function to the Base UI to automatically attach the required prefixes to the router.

@prmsrswt @bwplotka Any thoughts? 

## Verification

Started Bucket Web UI with `--web.external-prefix=thanos/bucket`. All links and assets were working fine.

### DIFF
--- a/cmd/thanos/tools_bucket.go
+++ b/cmd/thanos/tools_bucket.go
@@ -347,6 +347,8 @@ func registerBucketWeb(app extkingpin.AppClause, objStoreConfig *extflag.PathOrC
 		router := route.New()
 		ins := extpromhttp.NewInstrumentationMiddleware(reg)
 
+		router = router.WithPrefix("/" + strings.Trim(*webExternalPrefix, "/"))
+
 		bucketUI := ui.NewBucketUI(logger, *label, *webExternalPrefix, *webPrefixHeaderName, "", component.Bucket)
 		bucketUI.Register(router, true, ins)
 
--- a/cmd/thanos/tools_bucket.go
+++ b/cmd/thanos/tools_bucket.go
@@ -347,7 +347,9 @@ func registerBucketWeb(app extkingpin.AppClause, objStoreConfig *extflag.PathOrC
 		router := route.New()
 		ins := extpromhttp.NewInstrumentationMiddleware(reg)
 
-		router = router.WithPrefix("/" + strings.Trim(*webExternalPrefix, "/"))
+		if *webExternalPrefix != "" {
+			router = router.WithPrefix("/" + strings.Trim(*webExternalPrefix, "/"))
+		}
 
 		bucketUI := ui.NewBucketUI(logger, *label, *webExternalPrefix, *webPrefixHeaderName, "", component.Bucket)
 		bucketUI.Register(router, true, ins)
--- a/pkg/ui/templates/bucket_menu.html
+++ b/pkg/ui/templates/bucket_menu.html
@@ -11,7 +11,7 @@
                     <a class="nav-link" href="https://thanos.io/tip/thanos/getting-started.md/" target="_blank">Help</a>
                 </li>
                 <li class="nav-item">
-                    <a class="nav-link" href="/new/" target="_blank">New UI</a>
+                    <a class="nav-link" href="{{ pathPrefix }}/new/" target="_blank">New UI</a>
                 </li>
             </ul>
         </div>


prmsrswt (commented):
Looks good on first look. You still forgot to run `make assets` here though XD.

aribalam:
@prmsrswt Done! Wish there was some kind of auto-commit for that! :')

cmd/thanos/tools_bucket.go
```diff
@@ -347,6 +347,10 @@ func registerBucketWeb(app extkingpin.AppClause, objStoreConfig *extflag.PathOrC
 		router := route.New()
 		ins := extpromhttp.NewInstrumentationMiddleware(reg)
 
+		if *webExternalPrefix != "" {
+			router = router.WithPrefix("/" + strings.Trim(*webExternalPrefix, "/"))
```
prmsrswt, line 351: This is not what `external-prefix` is used for. This is what `route-prefix` is for. The default value of `route-prefix` is equal to `external-prefix` though. Take a look at how it is handled in `query.go`.

aribalam: The issue is that the bucket web currently does not pick up `route-prefix` through cmd flags. Therefore the router is not getting prefixed before creating the UI paths, as is happening in `query.go`. I am not sure whether bucket web should support `route-prefix` as it was not mentioned in the docs. If it should, then I could add code for picking `route-prefix` through cmd flags then proceed similar to `query.go`.

PR #3254, opened 2020-09-28
Show block sizes in block viewer

someshkoli:

## Changes

- Implemented GetSize which returns total chunk size & index size for a  block
- Adds block size to the api response in blocks api
- Updated UI to display block (chunk and index) size

thanos_block_view

## Verification
- tested manually for now
- React tests passed for changes

### DIFF
--- a/cmd/thanos/compact.go
+++ b/cmd/thanos/compact.go
@@ -230,8 +230,12 @@ func runCompact(
 				duplicateBlocksFilter,
 			}, []block.MetadataModifier{block.NewReplicaLabelRemover(logger, conf.dedupReplicaLabels)},
 		)
-		cf.UpdateOnChange(func(blocks []metadata.Meta, err error) {
-			compactorView.Set(blocks, err)
+		cf.UpdateOnChange(func(metas []metadata.Meta, err error) {
+			blocks, blkerr := getBlockMetaAndSize(metas, bkt)
+			compactorView.Set(metas, err)
+			if blkerr != nil {
+				api.SetLoaded([]blocksAPI.BlockInfo{}, blkerr)
+			}
 			api.SetLoaded(blocks, err)
 		})
 		sy, err = compact.NewSyncer(
@@ -414,8 +418,12 @@ func runCompact(
 		// Separate fetcher for global view.
 		// TODO(bwplotka): Allow Bucket UI to visualize the state of the block as well.
 		f := baseMetaFetcher.NewMetaFetcher(extprom.WrapRegistererWithPrefix("thanos_bucket_ui", reg), nil, nil, "component", "globalBucketUI")
-		f.UpdateOnChange(func(blocks []metadata.Meta, err error) {
-			global.Set(blocks, err)
+		f.UpdateOnChange(func(metas []metadata.Meta, err error) {
+			blocks, blkerr := getBlockMetaAndSize(metas, bkt)
+			compactorView.Set(metas, err)
+			if blkerr != nil {
+				api.SetLoaded([]blocksAPI.BlockInfo{}, blkerr)
+			}
 			api.SetGlobal(blocks, err)
 		})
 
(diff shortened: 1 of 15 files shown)


daksh-sagar (commented):
The react-app uses Prettier as its code formatter, so you should probably install it in your ide and format the code with it, this will resolve formatting errors 🙂

cmd/thanos/compact.go
```diff
@@ -230,8 +230,12 @@ func runCompact(
 				duplicateBlocksFilter,
 			}, []block.MetadataModifier{block.NewReplicaLabelRemover(logger, conf.dedupReplicaLabels)},
 		)
-		cf.UpdateOnChange(func(blocks []metadata.Meta, err error) {
-			compactorView.Set(blocks, err)
+		cf.UpdateOnChange(func(metas []metadata.Meta, err error) {
+			blocks, blkerr := getBlockMetaAndSize(metas, bkt)
```
krasi-georgiev, line 234: I think instead of calculating the block size when fetching from the storage It should calculate the block size before uploading it to block storage so that the uploaded block metafile has the size already in in.


@bwplotka what do you think?

krasi-georgiev:
> tested manually for now
we should def have some golang tests for this.

pkg/api/blocks/v1.go
```diff
@@ -24,11 +25,16 @@ type BlocksAPI struct {
 	loadedBlocksInfo *BlocksInfo
 }
 
+type BlockInfo struct {
```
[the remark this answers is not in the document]
someshkoli: Had thought about it but was not sure if was a good idea to store it in the metadata.Meta

pkg/ui/react-app/src/thanos/pages/blocks/__testdata__/testdata.ts
```diff
@@ -4,2057 +4,3067 @@ export const sampleAPIResponse: { status: string; data: BlockListProps } = {
...
-            resolution: 0,
-          },
-          labels: {
-            monitor: 'prometheus_two',
-          },
-          source: 'compactor',
-        },
-        ulid: '01EEB0ZRSQDJW51W11V4R6YP4T',
-        version: 1,
+        meta: {
```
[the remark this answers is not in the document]
someshkoli: Its just adding the block size. Actually its the test data and has array of blocks so the blocksize is added to all of the blocks in the array.

someshkoli:
> > tested manually for now
> > we should def have some golang tests for this.

yes, will add go tests soon

cmd/thanos/compact.go
someshkoli: yeah that seems to be fair, that way we can eliminate a few things and read size from the meta file itself whenever required

prmsrswt: > I think instead of calculating the block size when fetching from the storage It should calculate the block size before uploading it to block storage so that the uploaded block metafile has the size already in in.

This seems like a nice idea. We will save a lot of network requests. The only thing that we need to think about is how to handle old data. The old blocks won't have any size field in their meta.json.

krasi-georgiev: hhm, good question, I think since it is a new feature it is ok to NOT show the size for older blocks

someshkoli: @krasi-georgiev, the method you mentioned requires pool which in case is `chunkenc.Pool`  instance and thanos bucket uses `Pool` instance ?

krasi-georgiev: you can pass nil

pkg/api/blocks/v1.go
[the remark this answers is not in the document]
krasi-georgiev: I think it is ok

GiedriusS:
@someshkoli any updates on this?

someshkoli:
> @someshkoli any updates on this?

Nope :/ stuck at a part where I need to check if the changes are reflected but not sure how to do that. Done with pushing size to meta before uploading it but couldn't check if its actually being added

PR #3256, opened 2020-09-29
Website: fixed linking issue for v016 docs on website

thisisobate:

## Changes
fixed linking issue for v016 docs on website.

### DIFF
--- a/docs/community.md
+++ b/docs/community.md
@@ -2,7 +2,6 @@
 title: Community
 type: docs
 menu: contributing
-aliases: [/community.md/]
 ---
 
 # General
--- a/docs/community.md
+++ b/docs/community.md
@@ -2,6 +2,7 @@
 title: Community
 type: docs
 menu: contributing
+aliases: [/community.md/]
 ---
 
 # General
--- a/docs/design.md
+++ b/docs/design.md
@@ -2,7 +2,6 @@
 title: Design
 type: docs
 menu: thanos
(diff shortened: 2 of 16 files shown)


thisisobate:
PTAL: @bwplotka @squat @prmsrswt

squat (approved):

squat:
Is this a dupe of #3258? Do we need this in the 0.15 docs too? Maybe we can rather merge this into the release branch. and then merge the release branch into master as part of our regular release flow

thisisobate:
This is beneficial for only future releases. #3258 was made specifically for release-0.16 docs because it has already been cut from master so `aliases` frontmatter settings can't be found in release-0.16 docs.

thisisobate:
I think we should consider merging #3258 to master before we could consider this PR since #3258 promises to solve the problem at hand.

Also, couldn't view netlify preview in checks so I don't know if it's right to merge #3258

bwplotka (changes requested):
This is bit tedious. What's wrong with the approach on hugo.yaml?

thisisobate:
I tried the approach on hugo.yaml but it wasn't working. 

This was what I did: 
```
v0.16: "/:section /:filename.md" 

```

bwplotka (approved):

bwplotka:
This is it! Good for quick short term fix :+1:

PR #3257, opened 2020-09-29
[Ruler][DNS] Don't propagate no such host error if using default resolver

### LINKED ISSUE #3186: Thanos ruler dies on startup if AM DNS can't be resolved

thanosio/thanos:v0.14.0

**What happened**:
Some details/backstory can be found here: https://github.com/prometheus/alertmanager/issues/2372
TL;DR, thanos ruler dies at startup if it cannot revolve AM DNS.

Ruler to boot up and at leats process rules.

```
docker run -p10902:10902 thanosio/thanos:v0.14.0 rule --alertmanagers.url=dns+http://k3jhut933.messagebird.io:9093/alertmanager --query=mon.messagbird.io/query
```
<details>Logs
<p>

```
level=info ts=2020-09-18T07:32:40.602076815Z caller=main.go:151 msg="Tracing will be disabled"
level=info ts=2020-09-18T07:32:40.605531707Z caller=head.go:645 component=tsdb msg="Replaying WAL and on-disk memory mappable chunks if any, this may take a while"
level=info ts=2020-09-18T07:32:40.607499462Z caller=head.go:706 component=tsdb msg="WAL segment loaded" segment=0 maxSegment=0
level=info ts=2020-09-18T07:32:40.607541953Z caller=head.go:709 component=tsdb msg="WAL replay completed" duration=1.973454ms
level=info ts=2020-09-18T07:32:40.609255532Z caller=options.go:23 component=rules protocol=gRPC msg="disabled TLS, key and cert must be set to enable"
level=info ts=2020-09-18T07:32:40.609579143Z caller=grpc.go:92 component=rules service=gRPC/server component=rule msg="registering as gRPC StoreAPI and RulesAPI"
level=info ts=2020-09-18T07:32:40.610248039Z caller=rule.go:638 component=rules msg="no supported bucket was configured, uploads will be disabled"
level=info ts=2020-09-18T07:32:40.610285622Z caller=rule.go:641 component=rules msg="starting rule node"
level=info ts=2020-09-18T07:32:40.610460329Z caller=intrumentation.go:60 component=rules msg="changing probe status" status=healthy
level=info ts=2020-09-18T07:32:40.610495283Z caller=http.go:56 component=rules service=http/server component=rule msg="listening for requests and metrics" address=0.0.0.0:10902
level=info ts=2020-09-18T07:32:40.610988764Z caller=rule.go:791 component=rules msg="reload rule files" numFiles=0
level=info ts=2020-09-18T07:32:40.612768561Z caller=intrumentation.go:48 component=rules msg="changing probe status" status=ready
level=info ts=2020-09-18T07:32:40.612830648Z caller=grpc.go:119 component=rules service=gRPC/server component=rule msg="listening for serving gRPC" address=0.0.0.0:10901
level=info ts=2020-09-18T07:32:40.620800493Z caller=manager.go:888 component=rules msg="Stopping rule manager..."
level=info ts=2020-09-18T07:32:40.620844692Z caller=manager.go:898 component=rules msg="Rule manager stopped"
level=info ts=2020-09-18T07:32:40.620853668Z caller=manager.go:888 component=rules msg="Stopping rule manager..."
level=info ts=2020-09-18T07:32:40.620857918Z caller=manager.go:898 component=rules msg="Rule manager stopped"
level=warn ts=2020-09-18T07:32:40.620866143Z caller=intrumentation.go:54 component=rules msg="changing probe status" status=not-ready reason="lookup IP addresses \"k3jhut933.messagebird.io\": lookup k3jhut933.messagebird.io on 192.168.65.1:53: no such host"
level=info ts=2020-09-18T07:32:40.620883785Z caller=grpc.go:138 component=rules service=gRPC/server component=rule msg="gracefully stopping internal server"
level=info ts=2020-09-18T07:32:40.620983285Z caller=grpc.go:150 component=rules service=gRPC/server component=rule msg="internal server shutdown" err="lookup IP addresses \"k3jhut933.messagebird.io\": lookup k3jhut933.messagebird.io on 192.168.65.1:53: no such host"
level=warn ts=2020-09-18T07:32:40.621015491Z caller=intrumentation.go:54 component=rules msg="changing probe status" status=not-ready reason="lookup IP addresses \"k3jhut933.messagebird.io\": lookup k3jhut933.messagebird.io on 192.168.65.1:53: no such host"
level=info ts=2020-09-18T07:32:40.621064992Z caller=http.go:81 component=rules service=http/server component=rule msg="internal server shutdown" err="lookup IP addresses \"k3jhut933.messagebird.io\": lookup k3jhut933.messagebird.io on 192.168.65.1:53: no such host"
level=info ts=2020-09-18T07:32:40.621109322Z caller=intrumentation.go:66 component=rules msg="changing probe status" status=not-healthy reason="lookup IP addresses \"k3jhut933.messagebird.io\": lookup k3jhut933.messagebird.io on 192.168.65.1:53: no such host"
level=error ts=2020-09-18T07:32:40.621870546Z caller=main.go:211 err="lookup IP addresses \"k3jhut933.messagebird.io\": lookup k3jhut933.messagebird.io on 192.168.65.1:53: no such host\nrule command failed\nmain.main\n\t/go/src/github.com/thanos-io/thanos/cmd/thanos/main.go:211\nruntime.main\n\t/usr/local/go/src/runtime/proc.go:203\nruntime.goexit\n\t/usr/local/go/src/runtime/asm_amd64.s:1373"
```

</p>
</details>

Could not reproduce on MacOS.

When passing `--query.sd-dns-resolver=miekgdns`, ruler does not die.

OGKevin:
Signed-off-by: Kevin Hellemun <17928966+OGKevin@users.noreply.github.com>

Fixes #3186


OGKevin:
Shall there be a debug log entry when this happens 🤔 ?

pkg/discovery/dns/resolver.go
```diff
@@ -71,7 +72,10 @@ func (s *dnsSD) Resolve(ctx context.Context, name string, qtype QType) ([]string
 		}
 		ips, err := s.resolver.LookupIPAddr(ctx, host)
 		if err != nil {
-			return nil, errors.Wrapf(err, "lookup IP addresses %q", host)
+			dnsErr, ok := err.(*net.DNSError)
+			if !(s.resolverType == GolangResolverType && ok && dnsErr.IsNotFound) {
```
bwplotka, line 76: Do we need to differentiate by resolver? I think it's enough that we cast to error type, no? (:

bwplotka (changes requested):
Thanks!

It's not the only a ruler who is using this, also Querier and other components  and projects including Cortex.

I wonder if this makes sense, maybe putting it behind some option would do :thinking: 

No host found is when you cannot reach DNS host and this is definitely a configuration, server or just access failure. Imagine you are rolling to new cluster and somehow pod cannot each DNS server. By masking this error we will never know about this error right? 

cc @pracucci @pstibrany

bwplotka:
What exactly you want to achieve @OGKevin what's the use case? (:

OGKevin:
AFAIK, Host not found is also returned when the DNS server **_is_** reachable. Referent to this [article](https://www.dnsknowledge.com/whatis/nxdomain-non-existent-domain-2/) for example.

```
➜  ~ host k3jhut933.messagebird.io
Host k3jhut933.messagebird.io not found: 3(NXDOMAIN)
➜  ~ host google.com  
google.com has address 172.217.17.78
google.com has IPv6 address 2a00:1450:4001:817::200e
google.com mail is handled by 30 alt2.aspmx.l.google.com.
google.com mail is handled by 20 alt1.aspmx.l.google.com.
google.com mail is handled by 10 aspmx.l.google.com.
google.com mail is handled by 50 alt4.aspmx.l.google.com.
google.com mail is handled by 40 alt3.aspmx.l.google.com.
```

The DNS server is working fine, the DNS just does not exits, hence the NXDOMAIN which in go results in a `Host not found` error.

The issue linked in #3186 has more context of the use case that we want to avoid.

pkg/discovery/dns/resolver.go
OGKevin: Yes we do as explained here: https://github.com/thanos-io/thanos/issues/3186#issuecomment-700730106

The behaviour of the diff resolvers are different when the host does not exists.

OGKevin: >I wonder if this makes sense, maybe putting it behind some option would do 🤔 

I disagree purely based on the diff behaviour of the resolvers when NXDOMAIN (or similar errors) happens.

OGKevin: >By masking this error we will never know about this error right?
In this case we could also add warning logs so that the error is not totally masked, it just prevents a crash.

bwplotka: `(*net.DNSError)` is never used in miekg so we can just check the cast error and add comment (which is exactly what you said): 

```
miekgdns and the default DNS server have different behaviors when a host can not be resolved. The default DNS resolved returns a err no such host while the miekgdns returns a nil response and a nil error.
```

(: 

Resolver type is not needed. WDYT?

bwplotka (commented):
Ok, I checked a bit and it looks like indeed IsNotFound is returned when `EAI_NONAME` is given by Linux resolver ([getaddrinfo](https://man7.org/linux/man-pages/man3/getaddrinfo.3.html)) so you might be right. 

Still I would consider this as an error, potentially a configuration error case. (imagine you made a typo in crafting a DNS target for alertmananger).
 
I think the main question regarding your issue is really, should a ruler just mention the error and continue running or not. This decision might orthogonal to this change and in fact depends what it resolves for (key functionality or not) :thinking: 

Some more opinions welcome, but I would rather stop crashing ruler on fail like this on start, but still log & instrument it as failure overall.

OGKevin:
The problem might not be a config issue only. If Alertmanagers was up and running fine, and the DNS name becomes invalid because all pods are down. Ruler wil crashloop without making any config change to ruler.

So IMO Ruler should for sure not crash because Alertmanagers can't be resolved. Because ruler is not only responsible for evaluating alerting rules. I also agree that there should be a log entry saying that DNS resolution failed.

On the other issues, e.g. how it impacts the other components I can't make a clear assessment. What we could do indeed is, add a flag and set this to "true" by default for ruler, and false for the other components so that the behaviour only changes for Ruler.

bwplotka:
> . Ruler wil crashloop without making any config change to ruler.

Hm, crashlooop will actually apply all config changes right?

OGKevin:
I mean more like, ruller will crashloop not because of a config change made by a human to ruler. Ruler will go in crashloop because a dependency, Alertmanager, is down and k8s service FQDN returns NXDOMAIN (or equivalent) and dies. The subsequently, ruler won't start up because of the DNS resolution failure and eventually ends up in a crash loop.

https://github.com/thanos-io/thanos/blob/e4941a5c174f40741806aa81936176b3cb3160cb/cmd/thanos/rule.go#L418
https://github.com/thanos-io/thanos/blob/e4941a5c174f40741806aa81936176b3cb3160cb/cmd/thanos/rule.go#L792-L798

pkg/discovery/dns/provider.go
```diff
@@ -55,9 +55,9 @@ func (t ResolverType) ToResolver(logger log.Logger) ipLookupResolver {
 
 // NewProvider returns a new empty provider with a given resolver type.
 // If empty resolver type is net.DefaultResolver.
-func NewProvider(logger log.Logger, reg prometheus.Registerer, resolverType ResolverType) *Provider {
+func NewProvider(logger log.Logger, reg prometheus.Registerer, resolverType ResolverType, returnErrOnNotFound bool) *Provider {
```
OGKevin, line 58: hmm it seems changing the signature of this method is breaking code in vendor 😅 
```
# github.com/cortexproject/cortex/pkg/chunk/cache
../../vendor/github.com/cortexproject/cortex/pkg/chunk/cache/memcached_client.go:116:30: not enough arguments in call to dns.NewProvider
	have ("github.com/go-kit/kit/log".Logger, prometheus.Registerer, dns.ResolverType)
	want ("github.com/go-kit/kit/log".Logger, prometheus.Registerer, dns.ResolverType, bool)
# github.com/cortexproject/cortex/pkg/ring/kv/memberlist
../../vendor/github.com/cortexproject/cortex/pkg/ring/kv/memberlist/memberlist_client.go:295:34: not enough arguments in call to dns.NewProvider
	have ("github.com/go-kit/kit/log".Logger, "github.com/prometheus/client_golang/prometheus".Registerer, dns.ResolverType)
	want ("github.com/go-kit/kit/log".Logger, "github.com/prometheus/client_golang/prometheus".Registerer, dns.ResolverType, bool)

Compilation finished with exit code 2

```

OGKevin:
I the link that is being reported as dead is not actually dead. I dont quite understand why e2e failed 🤔

pkg/discovery/dns/resolver.go
```diff
@@ -71,7 +78,14 @@ func (s *dnsSD) Resolve(ctx context.Context, name string, qtype QType) ([]string
 		}
 		ips, err := s.resolver.LookupIPAddr(ctx, host)
 		if err != nil {
-			return nil, errors.Wrapf(err, "lookup IP addresses %q", host)
+			dnsErr, ok := err.(*net.DNSError)
+			// https://github.com/thanos-io/thanos/issues/3186
+			// Default DNS resolver can make thanos components crash if DSN resolutions results in EAI_NONAME.
```
OGKevin, line 83: Typo DNS
```suggestion
			// Default DNS resolver can make thanos components crash if DNS resolutions results in EAI_NONAME.
```

pkg/discovery/dns/provider.go
```diff
@@ -53,11 +53,38 @@ func (t ResolverType) ToResolver(logger log.Logger) ipLookupResolver {
...
 
+// Deprecated. Use NewProviderWithReturnOnErrorIfNotFound instead.
+//
 // NewProvider returns a new empty provider with a given resolver type.
 // If empty resolver type is net.DefaultResolver.
+// TODO(OGKevin): Refactor code to use new method and eventually rename NewProviderWithReturnOnErrorIfNotFound back to NewProvider.
 func NewProvider(logger log.Logger, reg prometheus.Registerer, resolverType ResolverType) *Provider {
 	p := &Provider{
-		resolver: NewResolver(resolverType.ToResolver(logger)),
+		resolver: NewResolver(resolverType.ToResolver(logger), logger, true),
```
pstibrany, line 63: Can this call `NewProviderWithReturnOnErrorIfNotFound` with `true` flag? I don’t quite see reason for duplicating entire method.

pkg/discovery/dns/resolver.go
```diff
@@ -71,7 +78,14 @@ func (s *dnsSD) Resolve(ctx context.Context, name string, qtype QType) ([]string
...
+			// the flag returnErrOnNotFound can be used to prevent such crash.
+			if !(!s.returnErrOnNotFound && ok && dnsErr.IsNotFound) {
+				return nil, errors.Wrapf(err, "lookup IP addresses %q", host)
+			}
```
pstibrany, line 87: I would remove double-negation in the condition and reorder it in logical sequence:

```suggestion
			if dnsErr, ok := err.(*net.DNSError); !ok || !dnsErr.IsNotFound || s.returnErrOnNotFound {
				return nil, errors.Wrapf(err, "lookup IP addresses %q", host)
			}
```

(and remove line 81)

pstibrany (commented):
Change to `dns` package looks good to me.

pkg/discovery/dns/provider.go
OGKevin: Changed, thanks.

pkg/discovery/dns/resolver.go
OGKevin: Thanks.

pkg/discovery/dns/provider.go
```diff
@@ -53,11 +53,20 @@ func (t ResolverType) ToResolver(logger log.Logger) ipLookupResolver {
 	return r
 }
 
+// Deprecated. Use NewProviderWithReturnOnErrorIfNotFound instead.
+//
 // NewProvider returns a new empty provider with a given resolver type.
 // If empty resolver type is net.DefaultResolver.
+// TODO(OGKevin): Refactor code to use new method and eventually rename NewProviderWithReturnOnErrorIfNotFound back to NewProvider.
 func NewProvider(logger log.Logger, reg prometheus.Registerer, resolverType ResolverType) *Provider {
+	return NewProviderWithReturnOnErrorIfNotFound(logger, reg, resolverType, true)
```
bwplotka, line 62: I am pretty sure there are better ways to handle this then this. Something not involving super long method name and boolean flag (:

pkg/discovery/dns/resolver.go
```diff
@@ -71,7 +78,13 @@ func (s *dnsSD) Resolve(ctx context.Context, name string, qtype QType) ([]string
 		}
 		ips, err := s.resolver.LookupIPAddr(ctx, host)
 		if err != nil {
-			return nil, errors.Wrapf(err, "lookup IP addresses %q", host)
+			// https://github.com/thanos-io/thanos/issues/3186
```
bwplotka, line 81: Why issue link here?

pkg/discovery/dns/resolver.go
```diff
@@ -71,7 +78,13 @@ func (s *dnsSD) Resolve(ctx context.Context, name string, qtype QType) ([]string
...
+			// Default DNS resolver can make thanos components crash if DNS resolutions results in EAI_NONAME.
```
bwplotka, line 82: This comment does not make sense, to me. Why telling a story in some deep Thanos package what happened? There might be more consumers of this than just  Thanos (e.g Cortex).

Let's focus on what user of this package can expect from the piece. And let's comment only if something is surprising. (:

pkg/discovery/dns/resolver.go
bwplotka: Let's delete IMO

pkg/discovery/dns/resolver.go
bwplotka: I would only say:

```
We exclude error from std Golang resolver for the case of the domain (e.g `NXDOMAIN`) not being found by DNS server. Since `miekg` does not consider this as an error,  when the host cannot be found, empty slice will be returned.
```

WDYT?

pkg/discovery/dns/provider.go
```diff
@@ -53,11 +53,20 @@ func (t ResolverType) ToResolver(logger log.Logger) ipLookupResolver {
...
+}
+
+// NewProviderWithReturnOnErrorIfNotFound returns a new empty provider with a given resolver type.
+// If empty resolver type is net.DefaultResolver.
+func NewProviderWithReturnOnErrorIfNotFound(logger log.Logger, reg prometheus.Registerer, resolverType ResolverType, returnErrOnNotFound bool) *Provider {
```
bwplotka, line 67: Let's just remove this and make this an normal (expected) behaviour.

pkg/discovery/dns/resolver.go
```diff
@@ -71,7 +78,13 @@ func (s *dnsSD) Resolve(ctx context.Context, name string, qtype QType) ([]string
...
+			// the flag returnErrOnNotFound can be used to prevent such crash.
+			if dnsErr, ok := err.(*net.DNSError); !ok || !dnsErr.IsNotFound || s.returnErrOnNotFound {
+				return nil, errors.Wrapf(err, "lookup IP addresses %q", host)
+			}
+			level.Error(s.logger).Log("msg", "failed to lookup IP addresses", "host", host, "err", err)
```
bwplotka, line 87: Let's do `res = nil` here, just in case there is something weird coming from resolver in error case

bwplotka, line 87: Let's delete this and if only, let's have one for both miekg and Go DNS.

pkg/discovery/dns/resolver.go
```diff
@@ -106,6 +119,12 @@ func (s *dnsSD) Resolve(ctx context.Context, name string, qtype QType) ([]string
 		return nil, errors.Errorf("invalid lookup scheme %q", qtype)
 	}
 
+	// https://github.com/thanos-io/thanos/issues/3186
+	// This happens when miekg is used as resolver. When the host cannot be found, nothing is returned.
```
bwplotka, line 123:
```suggestion
```

pkg/discovery/dns/resolver.go
```diff
@@ -106,6 +119,12 @@ func (s *dnsSD) Resolve(ctx context.Context, name string, qtype QType) ([]string
...
+	if res == nil && err == nil {
+		level.Warn(s.logger).Log("msg", "IP address lookup yielded no results nor errors", "host", host)
```
bwplotka, line 125:
```suggestion
		level.Warn(s.logger).Log("msg", "IP address lookup yielded no results. No host found or no addresses found", "host", host)
```

bwplotka (commented):
Awesome, thanks for offline chat, I think this makes sense, but let's make it a normal behaviour.

Some tests would be nice as well (:

pkg/discovery/dns/resolver.go
OGKevin: hmm I don't quite understand this comment :thinking:. Can you elaborate a lillte bit more?

pkg/discovery/dns/provider.go
```diff
@@ -55,9 +55,10 @@ func (t ResolverType) ToResolver(logger log.Logger) ipLookupResolver {
 
 // NewProvider returns a new empty provider with a given resolver type.
 // If empty resolver type is net.DefaultResolver.
+// TODO(OGKevin): Refactor code to use new method and eventually rename NewProviderWithReturnOnErrorIfNotFound back to NewProvider.
```
OGKevin, line 58:
```suggestion

```

bwplotka (approved):
I think I am fine with this consistency check, thanks!

bwplotka:
Can you rebase :hugs:

OGKevin:
@bwplotka  I could not rebase as I merged master in this branch before. Rebasing now will cause a headache. Do you want me to squash this to 1 commit?

bwplotka (approved):

bwplotka:
Nah no need, GH squashes this. Thanks and sorry for long discussion (:

damnever:
Do you have a release date for this? The rule will crash due to alert managers restart.

PR #3261, opened 2020-09-30
Segment files

pstibrany:

## Changes

This PR adds `segment_files` field into `meta.json` file (Thanos section), which lists all segment files for a block. This helps to avoid call to storage to get this list. This small optimization helps to reduce amount of LIST operations in clusters with many created blocks.

## Verification

- Added unit test for shipper, added check in compactor test
- Verified that new code path in `store/bucket.go` is hit by existing tests.

### DIFF
--- a/pkg/block/block.go
+++ b/pkg/block/block.go
@@ -222,3 +222,21 @@ func IsBlockDir(path string) (id ulid.ULID, ok bool) {
 	id, err := ulid.Parse(filepath.Base(path))
 	return id, err == nil
 }
+
+// Returns list of segment files for given block. Paths are relative to the chunks directory.
+// In case of errors, nil is returned.
+func GetSegmentFiles(blockDir string) []string {
+	chunksDir := filepath.Join(blockDir, ChunksDirname)
+
+	files, err := ioutil.ReadDir(chunksDir)
+	if err != nil {
+		return nil
+	}
+
+	// ReadDir returns files in sorted order already.
+	var result []string
+	for _, f := range files {
+		result = append(result, f.Name())
+	}
+	return result
+}
--- a/pkg/block/block.go
+++ b/pkg/block/block.go
@@ -223,7 +223,7 @@ func IsBlockDir(path string) (id ulid.ULID, ok bool) {
 	return id, err == nil
 }
 
-// Returns list of segment files for given block. Paths are relative to the chunks directory.
+// GetSegmentFiles returns list of segment files for given block. Paths are relative to the chunks directory.
 // In case of errors, nil is returned.
 func GetSegmentFiles(blockDir string) []string {
 	chunksDir := filepath.Join(blockDir, ChunksDirname)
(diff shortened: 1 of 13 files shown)


pkg/block/block.go
```diff
@@ -222,3 +222,21 @@ func IsBlockDir(path string) (id ulid.ULID, ok bool) {
 	id, err := ulid.Parse(filepath.Base(path))
 	return id, err == nil
 }
+
+// Returns list of segment files for given block. Paths are relative to the chunks directory.
```
bwplotka, line 226: For goDoc/go.dev to work properly: 

```suggestion
// GetSegmentFiles returns list of segment files for given block. Paths are relative to the chunks directory.
```

pkg/block/metadata/meta.go
```diff
@@ -59,6 +59,10 @@ type Thanos struct {
 
 	// Source is a real upload source of the block.
 	Source SourceType `json:"source"`
+
+	// List of segment files (in chunks directory), in sorted order.
+	// Not used if nil/empty.
```
bwplotka, line 64: Maybe let's just mention `optional`. Also, I would be somehow careful with empty. Empty / nil cases should be probably distinct. Block with no segments ... is kind of valid block, no? (:  But in worst case we do LIST call so not that bad.

```suggestion
	// List of segment files (in chunks directory), in sorted order. Optional. 
	// If nil, it means that that creator did not provide this information.
```

pkg/store/bucket.go
```diff
@@ -1330,12 +1330,20 @@ func newBucketBlock(
...
-		return nil
-	}); err != nil {
-		return nil, errors.Wrap(err, "list chunk files")
+	if len(meta.Thanos.SegmentFiles) > 0 {
+		b.chunkObjs = make([]string, 0, len(meta.Thanos.SegmentFiles))
+
+		for _, sf := range meta.Thanos.SegmentFiles {
+			b.chunkObjs = append(b.chunkObjs, path.Join(meta.ULID.String(), block.ChunksDirname, sf))
+		}
+	} else {
```
bwplotka, line 1339: No need for else, we can just return here (: 


https://thanos.io/tip/contributing/coding-style-guide.md/#control-structure-prefer-early-returns-and-avoid-else

bwplotka (approved):
Dear Peter.

Amazing. LGTM. :muscle:

pkg/block/metadata/meta.go
pstibrany: Shipper doesn’t upload blocks with no segment files.

pstibrany: > Shipper doesn’t upload blocks with no segment files.

My point was that block with no segments should not appear in Thanos. Also no segments = no chunks = no samples. Such block has no usable data. 

I will keep "Optional" part from your suggestion.

pstibrany:
Dear Bartek,

thanks for your review. I've updated the PR with your suggestions. Please take a look at your convenience.

With love,
-Peter

pracucci (approved):
I double checked if the new property is set in every place and I think so. This is apparently a small optimisation, but according to our calculations it would have a big impact to reduce bucket listing operations. Thanks @pstibrany !

PR #3262, opened 2020-09-30
Added Configuration Files page in the UI

### LINKED ISSUE #3217: ui: Add "Configuration Files" page
Right now we have a page to show the CLI flags passed and the Runtime and Build information. It would be nice to have a page showing the contents of configuration files passed to a component as well. For example, the object storage configuration, or the rules config file.

**Some things to keep in mind:**
- In Thanos, we use a special type of flag called `PathOrContent`. The user can pass either path to a file or the contents of the file directly.
- We can have multiple configuration files in a component (or no files at all).
- Some configuration files may contain secrets in them. Ideally, we would like to hide/sensor those secrets when showing the files on the Web UI.

aribalam:

## Changes

Fixes #3217 

Added Configuration Files page in the path `/configfiles` in React paths. 
The BaseAPI was changed to accommodate a map (from config name to its YAML content as string). The content of the config files was added to the map (for each component in `cmd/thanos`) before passing it to its API.

The Config Maps have been added for Bucket Web, Store, Rule, and Compactor.

The Screenshots are below -

Thanos Bucket Web

Thanos Store

### DIFF
--- a/cmd/thanos/compact.go
+++ b/cmd/thanos/compact.go
@@ -218,7 +218,7 @@ func runCompact(
 		"/loaded",
 		component,
 	)
-	api := blocksAPI.NewBlocksAPI(logger, conf.label, flagsMap)
+	api := blocksAPI.NewBlocksAPI(logger, conf.label, flagsMap, make(map[string]string))
 	var sy *compact.Syncer
 	{
 		// Make sure all compactor meta syncs are done through Syncer.SyncMeta for readability.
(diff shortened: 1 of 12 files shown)


aribalam:
@prmsrswt Could you take a look at this?

prmsrswt (commented):
Nice work! The only suggestion I have is maybe we can use monospace font-family for the config files.

There was one more point in the original issue that we also want to mask the secrets in these configuration files. How are you planning to tackle that?

You can look into how Prometheus does it as they also exposes thier config files in the UI with secrets masked.

Issue #3230, opened 2020-09-25
query-frontend: Split and cache metadata API responses by time interval


kakkoyun:
We recently added time range parameters to the LabelNames, LabelValues and Series API endpoints for both Thanos Query API. 
Now we expect those endpoints to be cached by Query Frontend to increase query scalability. In order to achieve this goal we need to introduce splitting request by time ranges.

### Tasks

* [ ] Introduce Split by time interval for Label (LabelNames and Label Values) API
* [ ] Cache Label API responses
* [ ] Introduce Split by time interval for Series API
* [ ] Cache Series API response

Help wanted :)

cc @yeya24

yeya24:
I am currently working on this issue already. As I mentioned [here](https://github.com/thanos-io/thanos/pull/3207#discussion_r493808806), some configs should be global, and some configs should be specific to different endpoints.

For multiple endpoints in the query frontend to work, my idea is to have two tripperware first.
1. Query range tripperware, the same as the current code. It has `limits`, `step_align`, `split_interval`, `results_cache` and `retry`.

2. Metadata tripperware, not sure about the naming. This tripperware handles `label names`, `label values` and `series` queriers and I think they can use the same middlewares. For caching, it would be better have a separate response cache config for this tripperware like `--metadata.response-cache`.

For flags, currently most of them have a `query-range` prefix, do we want to rename some of them to make them global? Otherwise it would be very confusing.

```
      --query-range.split-interval=24h  
                              Split queries by an interval and execute in parallel, 0 disables it.
      --query-range.max-retries-per-request=5  
                              Maximum number of retries for a single request; beyond this, the downstream error is returned.
      --query-range.max-query-length=0  
                              Limit the query time range (end - start time) in the query-frontend, 0 disables it.
      --query-range.max-query-parallelism=14  
                              Maximum number of queries will be scheduled in parallel by the frontend.
      --query-range.response-cache-max-freshness=1m  
                              Most recent allowed cacheable result, to prevent caching very recent results that might still be in flux.
      --query-range.response-cache-config-file=<file-path>  
                              Path to YAML file that contains response cache configuration.
      --query-range.response-cache-config=<content>  
                              Alternative to 'query-range.response-cache-config-file' flag (lower priority). Content of YAML file that contains response cache
                              configuration.
```

yeya24:
Any ideas are welcomed and let's discuss these options in this pr :).

PR #3276, opened 2020-10-05
Split labels/series API endpoints in query frontend

yeya24:

## Changes

This pr adds splitting and retry middleware support for all metadata API endpoints (label_name, label_value, series). This is the first step of #3230. Will add caching support in another pr since this pr is already very large.

## Verification

Tests updated.


pkg/queryfrontend/roundtrip.go
```diff
@@ -33,40 +36,117 @@ func NewTripperware(
...
+func getOperation(r *http.Request) string {
+	if r.Method == http.MethodGet || r.Method == http.MethodPost {
+		switch {
+		case strings.HasSuffix(r.URL.Path, "/api/v1/query"):
+			return instantQueryOp
+		case strings.HasSuffix(r.URL.Path, "/api/v1/query_range"):
+			return rangeQueryOp
+		case strings.HasSuffix(r.URL.Path, "/api/v1/labels"):
+			return labelNamesOp
+		case strings.HasSuffix(r.URL.Path, "/values") && strings.Contains(r.URL.Path, "/api/v1/label"):
```
kakkoyun, line 114: A regex matcher maybe?

pkg/queryfrontend/roundtrip.go
```diff
@@ -33,40 +36,117 @@ func NewTripperware(
...
+
+func newQueryRangeTripperware(
+	config Config,
+	limits queryrange.Limits,
+	codec *queryRangeCodec,
+	reg prometheus.Registerer,
+	logger log.Logger,
+) (frontend.Tripperware, error) {
 	queryRangeMiddleware := []queryrange.Middleware{queryrange.LimitsMiddleware(limits)}
+	instrumentMetrics := queryrange.NewInstrumentMiddlewareMetrics(reg)
```
kakkoyun, line 132:
```suggestion
	m := queryrange.NewInstrumentMiddlewareMetrics(reg)
```

kakkoyun (commented):
Great work 🎉 I just skimmed over and looking good so far.

pkg/queryfrontend/roundtrip.go
```diff
@@ -20,60 +21,148 @@ import (
...
+	switch op := getOperation(req); op {
+	case instantQueryOp:
+		r.queriesCount.WithLabelValues(instantQueryOp).Inc()
+	case rangeQueryOp:
+		r.queriesCount.WithLabelValues(rangeQueryOp).Inc()
+		return r.queryRange.RoundTrip(req)
+	case labelNamesOp, labelValuesOp, seriesOp:
+		r.queriesCount.WithLabelValues(op).Inc()
+		return r.metadata.RoundTrip(req)
+	default:
```
kakkoyun, line 103: How about adding an `Unknown` label for this case?

kakkoyun (approved):
@yeya24 Great work. It has taken some time to review the whole the PR. Overall it looks amazing. Thanks a lot for the contribution.

I have only added a small nit. Other than that we should test by integrating it with whole stack. It would be really nice to run it with `quickstart` script. WDYT?

pkg/queryfrontend/roundtrip.go
yeya24: I think not needed. Since query frontend is just a proxy for downstream querier, so all other requests are unknown. Like the requests for frontend resources... Then the unknown requests would be a lot.

yeya24:
> Other than that we should test by integrating it with whole stack. It would be really nice to run it with quickstart script. WDYT?

Yes, that's what I want as well. Since this task is trivial, I will open an issue and let someone else contribute.

yeya24:
While I am working on the caching part, I find that all response format must be protobuf because they will be marshaled in the results cache middlware.

I pushed another commit to switch the labels and series response to protobuf.

bwplotka (commented):
Amazing work, some post merge comments :hugs:

Issue #2608, opened 2020-05-14
query: Staleness problem


IKSIN:
**Thanos, Prometheus and Golang version used**:
thanos: v0.12.0

**Object Storage Provider**:
private CEPH (S3)

**What happened**:
See on `end_input` time and `resolution`:
<img width="1421" alt="Снимок экрана 2020-05-14 в 16 34 52" src="https://user-images.githubusercontent.com/2960110/81941650-c99f1080-9601-11ea-8588-d663b429de14.png">
<img width="1423" alt="Снимок экрана 2020-05-14 в 16 35 02" src="https://user-images.githubusercontent.com/2960110/81941676-cefc5b00-9601-11ea-8bd2-f6a2fda1be25.png">
Staleness functionality in prometheus library get rid of some points returned from thanos-stores.

**What you expected to happen**:
Return all data from store on any time_range

**How to reproduce it (as minimally and precisely as possible)**:
see on screenshots.

**Full logs to relevant components**:

**Anything else we need to know**:
I think, that we have few ways to resolve problem:
1. Update prometheus library and set `LookbackDelta` parameter > 5 min (need check)
2. Update query and move/duplicate points to needed timestamp. (Interpolate data for PromQL)
3. Update prometheus library to return all points from stores.

daixiang0:
What is the ceph version?

IKSIN:
CEPH does not matter, because thanos-stores return all data to query, and only in prometheus library points marked as staleness.

bwplotka:
Nice, thanks for this. Funnily enough we just talked about this exact problem with @juliusv (: 

We need different lookbackDelta for different resolution I think, right? @juliusv

bwplotka:
ref: https://matrix.to/#/!WaUKIfoqfiyWQhenET:matrix.org/$1589300436157120ZqIrC:matrix.org?via=matrix.org

juliusv:
At a minimum, it would be good to add the `--query.lookback-delta` that we have in Prometheus to Thanos as well. However, since it's a global setting, it would apply to all time series, even the ones that are scraped a intervals <5m. Normally you wouldn't want to set this lookback delta higher than needed for everything, as that will result in old samples being returned for quite long (although explicit staleness markers already help with that).

IKSIN:
I think that we need in dynamic lookback-delta inpdepended on resolution, forexample resolution/2.

bwplotka:
[...]

PR #3267, opened 2020-10-01
query: chain interpolating iterator (#2608)

krya-kryak:
This PR attempts to fix https://github.com/thanos-io/thanos/issues/2608
As of now it lacks unit tests, those will be added shortly. Meanwhile, any feedback is appreciated.

Signed-off-by: Vladimir Kononov <krya-kryak@users.noreply.github.com>

PR #3277, opened 2020-10-05
query: introduce dynamic lookback interval

krya-kryak:
Closes https://github.com/thanos-io/thanos/issues/2608
This allows queries with large step to make use of downsampled data.

## Changes

* Added CLI option for query

## Verification

This patch was tested with 0.15 branch (minor conflict solved manually), works as expected

Signed-off-by: Vladimir Kononov <krya-kryak@users.noreply.github.com>


cmd/thanos/query.go
```diff
@@ -317,20 +321,30 @@ func runQuery(
...
-				// TODO(bwplotka): Expose this as a flag: https://github.com/thanos-io/thanos/issues/703.
-				MaxSamples: math.MaxInt32,
-				Timeout:    queryTimeout,
-				NoStepSubqueryIntervalFn: func(rangeMillis int64) int64 {
-					return defaultEvaluationInterval.Milliseconds()
+		subqueryIntervalFunc = func(int64) int64 {
+			return defaultEvaluationInterval.Milliseconds()
+		}
+		totalEngines int
+		newEngine    = func(ld time.Duration) *promql.Engine {
```
bwplotka, line 328: I am a bit confused, do we need two functions? What about `engineFunc`? I guess you wanted to have large number of args in function, but it might be cleaner to embed `engineFunc` here, not embedding this function in `engineFunc`. WDYT?

cmd/thanos/query.go
```diff
@@ -73,6 +74,7 @@ func registerQuery(app *extkingpin.App) {
 		Default("20").Int()
 
 	lookbackDelta := cmd.Flag("query.lookback-delta", "The maximum lookback duration for retrieving metrics during expression evaluations. PromQL always evaluates the query for the certain timestamp (query range timestamps are deduced by step). Since scrape intervals might be different, PromQL looks back for given amount of time to get latest sample. If it exceeds the maximum lookback delta it assumes series is stale and returns none (a gap). This is why lookback delta should be set to at least 2 times of the slowest scrape interval. If unset it will use the promql default of 5m.").Duration()
+	dynamicLookbackDelta := cmd.Flag("query.dynamic-lookback-delta", "Allow for larger lookback duration for queries based on resolution.").Default("false").Bool()
```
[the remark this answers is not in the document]
bwplotka: And then we probably want raw delta default to be configurable.

cmd/thanos/query.go
```diff
@@ -536,3 +550,36 @@ func firstDuplicate(ss []string) string {
...
+	var (
+		engines       = make([]*promql.Engine, len(deltas))
+		defaultEngine = newEngine(lookbackDelta)
+		ld            = lookbackDelta.Milliseconds()
+	)
+	engines[0] = defaultEngine
+	for i, d := range deltas[1:] {
+		if ld < d {
+			// Convert delta from milliseconds to time.Duration (nanoseconds).
+			engines[i+1] = newEngine(time.Duration(d * 1_000_000))
```
bwplotka, line 568: This looks weird and should fail our lints: `1_000_000` - what about using `timestamp.Time(...)`

cmd/thanos/query.go
```diff
@@ -536,3 +550,36 @@ func firstDuplicate(ss []string) string {
...
 }
+
+func engineFunc(newEngine func(time.Duration) *promql.Engine, lookbackDelta time.Duration, dynamicLookbackDelta bool) func(int64) *promql.Engine {
+	deltas := []int64{downsample.ResLevel0}
+	if dynamicLookbackDelta {
+		deltas = []int64{downsample.ResLevel0, downsample.ResLevel1, downsample.ResLevel2}
+	}
+	var (
+		engines       = make([]*promql.Engine, len(deltas))
+		defaultEngine = newEngine(lookbackDelta)
```
bwplotka, line 561: Default? Why not just raw? resolution 0 is raw, so this should be default one, no? :thinking:

bwplotka (commented):
Great job! Thanks for this :+1: 

Overall to me, it makes sense, just some implementation suggestion etc. :hugs: 

NOTE: Let's reduce amount of engines we create - can we create just 3 engines (number of resolution we have) and just create promql.Engine wrapper that routes to correct one based on maxSourceResolution? :thinking:

cmd/thanos/query.go
krya-kryak: yeah, that does look better

cmd/thanos/query.go
krya-kryak: You're absolutely right, having two functions is confusing. Got some coffee, merged engineFunc and newEngine into single engineFactory. IMO things got a lot more unerstandable. I hope this isn't just coffee speaking.

cmd/thanos/query.go
krya-kryak: replaced with standard `time.Duration(d) * time.Millisecond`. Prometheus'es `timestamp` seems unhelpful here, as it's `Duration` we need, not `Time`

cmd/thanos/query.go
```diff
@@ -317,20 +321,30 @@ func runQuery(
...
-				Timeout:    queryTimeout,
-				NoStepSubqueryIntervalFn: func(rangeMillis int64) int64 {
-					return defaultEvaluationInterval.Milliseconds()
+		subqueryIntervalFunc = func(int64) int64 {
+			return defaultEvaluationInterval.Milliseconds()
+		}
+		totalEngines int
+		newEngine    = func(ld time.Duration) *promql.Engine {
+			var r prometheus.Registerer = reg
+			if dynamicLookbackDelta {
```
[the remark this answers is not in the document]
krya-kryak: Added a TODO for now

cmd/thanos/query.go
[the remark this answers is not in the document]
krya-kryak: `lookbackDelta` is still there, and it's still usable. Or are you talking about something else?

krya-kryak:
> NOTE: Let's reduce amount of engines we create - can we create just 3 engines (number of resolution we have) and just create promql.Engine wrapper that routes to correct one based on maxSourceResolution? 🤔

Well, this seems to be more or less similar to how it is done already - we create at most three engines, but choose correct one with an anonymous function call (note `pkg/api/query` change) instead of a wrapper. It seemed like a less intrusive way to implement, please correct me if I'm wrong.

cmd/thanos/query.go
```diff
@@ -73,6 +74,7 @@ func registerQuery(app *extkingpin.App) {
 		Default("20").Int()
 
 	lookbackDelta := cmd.Flag("query.lookback-delta", "The maximum lookback duration for retrieving metrics during expression evaluations. PromQL always evaluates the query for the certain timestamp (query range timestamps are deduced by step). Since scrape intervals might be different, PromQL looks back for given amount of time to get latest sample. If it exceeds the maximum lookback delta it assumes series is stale and returns none (a gap). This is why lookback delta should be set to at least 2 times of the slowest scrape interval. If unset it will use the promql default of 5m.").Duration()
+	dynamicLookbackDelta := cmd.Flag("query.dynamic-lookback-delta", "Allow for larger lookback duration for queries based on resolution.").Hidden().Default("true").Bool()
```
bwplotka, line 77: Quite busy recently with KubeCon talks, but I was thinking about one more thing we might want to check before deciding on this.

We actually now the best the resolution of the data... when we return it. Store GW depending on options can either return all 5m resolutions, or sometimes full raw resolution or sometimes mix: a bit of raw and a bit of 1h resolution. Wonder if this solution will work for those cases.. and what we actually expect from this? :thinking:

bwplotka (commented):
One important question that came to my mind before going full for this solution (:

bwplotka:
Wait, which one is the current fix / better? (: https://github.com/thanos-io/thanos/pull/3267 or this one?

krya-kryak:
> Wait, which one is the current fix / better? (: #3267 or this one?
I really prefer this one over #3267 for several reasons:
1) It doesn't mess with metrics staleness as much
2) It's less heavy on resources
3) It doesn't introduce rounding errors to query results

I'll just close that one.

cmd/thanos/query.go
krya-kryak: > We actually now the best the resolution of the data... when we return it. Store GW depending on options can either return all 5m resolutions, or sometimes full raw resolution or sometimes mix: a bit of raw and a bit of 1h resolution. Wonder if this solution will work for those cases.. and what we actually expect from this? 🤔

Well, result with mixed resolutions was the original case @IKSIN reported. I can foresee one behaviour change there - for range queries with sufficiently large `step` we might (but not necessarily will, dedpending on data and requested window) detect stale timeseries a bit later than we should have done. On the other hand, I doubt anyone will notice it given the requested step should be over 25m / 5h for dynamic lookback interval to even kick in.

Moreover, this adds the possibility to actually set a lower lookback interval for raw queries without sacrificing the downsampled views, allowing for more precise results when it trully maters (low `step` requested).

bwplotka: Hm, I think this makes sense - so you say lookback delta can be too large in some cases, but it's a tradeoff with auto-resolution :thinking: 
I think that might make good sense for now (:

cmd/thanos/query.go
```diff
@@ -536,3 +539,69 @@ func firstDuplicate(ss []string) string {
...
+		engines      = make([]*promql.Engine, len(deltas))
+		ld           = eo.LookbackDelta.Milliseconds()
+		totalEngines int
+	)
+	wrapReg := func() prometheus.Registerer {
+		wr := extprom.WrapRegistererWith(map[string]string{"engine": strconv.Itoa(totalEngines)}, eo.Reg)
+		totalEngines++
+		return wr
+	}
+	rawEngine := newEngine(promql.EngineOpts{
```
bwplotka, line 568: Can we remove this and init all engines in exactly the same way (no matter if raw or not raw) (:

bwplotka: It might simplify the code

bwplotka (commented):
Awesome! Some ideas for simplifying the code still, but otherwise LGTM (:

cmd/thanos/query.go
```diff
@@ -536,3 +539,69 @@ func firstDuplicate(ss []string) string {
...
+		deltas = []int64{downsample.ResLevel0, downsample.ResLevel1, downsample.ResLevel2}
+	}
+	var (
+		engines      = make([]*promql.Engine, len(deltas))
+		ld           = eo.LookbackDelta.Milliseconds()
+		totalEngines int
+	)
+	wrapReg := func() prometheus.Registerer {
+		wr := extprom.WrapRegistererWith(map[string]string{"engine": strconv.Itoa(totalEngines)}, eo.Reg)
+		totalEngines++
```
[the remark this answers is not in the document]
krya-kryak: got rid of it altogether in favour of explicit argument

cmd/thanos/query.go
krya-kryak: Thanks for the suggestion. Done.

cmd/thanos/query.go
```diff
@@ -536,3 +539,55 @@ func firstDuplicate(ss []string) string {
...
+	resolutions := []int64{downsample.ResLevel0}
+	if dynamicLookbackDelta {
+		resolutions = []int64{downsample.ResLevel0, downsample.ResLevel1, downsample.ResLevel2}
+	}
+	var (
+		engines = make([]*promql.Engine, len(resolutions))
+		ld      = eo.LookbackDelta.Milliseconds()
+	)
+	wrapReg := func(engineNum int) prometheus.Registerer {
+		return extprom.WrapRegistererWith(map[string]string{"engine": strconv.Itoa(engineNum)}, eo.Reg)
```
bwplotka, line 563: I would just inline this (:

bwplotka (approved):
LGTM! Thanks :+1: 

Sorry for delayed review :hugs:

cmd/thanos/query.go
bwplotka: Not a blocker (:

PR #3284, opened 2020-10-07
compact: Add a first pass of retention at the start of the compactor loop.

obiesmans:
This PR tries to address issue #1708, with the fixes discussed there.

We cannot totally move the retention policy apply at the start of the compact loop, but we can add a first pass than only apply a retention that is safe, in that it'll enable us to downsample the other resolution levels.

Tested : actually I've stumbled upon the problem in a real-world scenario : 
Screenshot 2020-10-07 at 10 13 45


kakkoyun:
@obiesmans Could you please rebase this PR? I promise I'll personally pay my attention to it

PR #3285, opened 2020-10-07
compact: fix disable downsampling flag name in compact.md

Sayrus:

## Changes

The flag used for disabling downsampling is now `--downsampling.disable` and one point in the documentation was not matching this. A really small fix but I hope it will prevent someone from using the wrong flag.

## Verification

Not relevant


prmsrswt (approved):
Thanks. There seems to be a conflict, can you rebase on master? LGTM after this.

Sayrus:
Indeed, there was a modification on the same line on master!

PR #3289, opened 2020-10-07
receiver/compact/store/sidecar: Added a provision in the bucket config to add a `prefix`

dsayan154:
This PR is to address issue [ #1318 ](https://github.com/thanos-io/thanos/issues/1318).

## Changes

- Added `prefix` to BucketConfig struct in factory.go
- Wrote wrapper function prefixedBucket wrapper struct and related wrapper functions in objstore.go

## Verification
Not yet verified, this is for the maintainers to review, if this is the right approach to address this issue. I am struggling to add unit tests for this, so any hints would be helpful. But, I can test it manually if the overall feedback on this approach is positive.


docs/storage.md
```diff
@@ -409,4 +415,36 @@ This is mainly useful for testing and demos.
 type: FILESYSTEM
 config:
   directory: ""
+prefix: ""
+```
+
+## Prefix
+
+Now Thanos also supports multi-tenancy at the object storage using custom prefix. Check [this](https://github.com/thanos-io/thanos/issues/1318) issue. This feature is implemented in such a way that, no client level changes are required. All object storage implementations support this `prefix` key in their object storage config yaml.
```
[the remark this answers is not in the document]
bwplotka: It would be definitely nice to mention `WHY` someone would use this field.

I would NOT mention multi-tenancy use case, let's talk on issue more to clarify the purpose, will add comment on https://github.com/thanos-io/thanos/issues/1318

pkg/objstore/objstore.go
```diff
@@ -437,6 +437,87 @@ func (b *metricBucket) Name() string {
 	return b.bkt.Name()
 }
 
+type prefixedBucket struct {
```
bwplotka, line 440: nice idea :+1:

pkg/objstore/objstore.go
```diff
@@ -437,6 +437,87 @@ func (b *metricBucket) Name() string {
...
+	bkt    Bucket
+	prefix string
+}
+
+//NewPrefixedBucket returns a new prefixedBucket.
```
bwplotka, line 445:
```suggestion
// NewPrefixedBucket returns a new prefixedBucket.
```

bwplotka (commented):
Some comments!

Amazing and clean work BTW :heart:

pkg/objstore/objstore.go
dsayan154: Thanks @bwplotka

dsayan154: Thanks to @pracucci for this [comment](https://github.com/thanos-io/thanos/issues/1318#issuecomment-660851708) and a detailed explanation over Slack

pkg/objstore/objstore.go
```diff
@@ -437,6 +437,87 @@ func (b *metricBucket) Name() string {
 	return b.bkt.Name()
 }
 
+type prefixedBucket struct {
+	bkt    Bucket
```
kakkoyun, line 441: How about embedding `Bucket` into this struct so that you don't have to implement all the passthrough functions?

kakkoyun (commented):
Looks good. Thanks ❤️  

I have a couple of suggestions.

* How about embedding `Bucket` into new struct?
* Could we maybe add some end-to-end tests?
* We need a changelog entry.

docs/storage.md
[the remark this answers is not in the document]
bwplotka: Still to do (:

pkg/objstore/client/factory.go
```diff
@@ -38,6 +38,7 @@ const (
 type BucketConfig struct {
 	Type   ObjProvider `yaml:"type"`
 	Config interface{} `yaml:"config"`
+	Prefix string      `yaml:"prefix" default:""`
```
bwplotka, line 41: :+1:

pkg/objstore/objstore.go
bwplotka: Yes, this would work well :+1: The downside is that we can miss prefix for newly added method but we should not need to optimize for this

bwplotka (approved):
LGTM, just minor comments to address, then we can merge! (: 

Thanks!

docs/storage.md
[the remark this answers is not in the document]
dsayan154: Sorry for the delay, I didn't get chance to work on this for some other priority. Should be done by tomorrow..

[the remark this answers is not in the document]
dsayan154: Yes, `Prefix: "/tenant-0"` should work too.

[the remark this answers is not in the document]
dsayan154: Updated the docs in the latest commit

pkg/objstore/objstore.go
dsayan154: resolving this as their are no action points to address

pkg/objstore/client/factory.go
dsayan154: Thanks @bwplotka :)

pkg/objstore/client/factory.go
```diff
@@ -38,6 +38,8 @@ const (
 type BucketConfig struct {
 	Type   ObjProvider `yaml:"type"`
 	Config interface{} `yaml:"config"`
+	// Prefix field allows adding an optional prefix to block (`/<ulid>`) and block files which are uploaded by any block "producer" (e.g sidecar, ruler, receiver).
+	Prefix string `yaml:"prefix" default:""` //
```
pracucci, line 42:
```suggestion
	Prefix string `yaml:"prefix" default:""`
```

pkg/objstore/objstore.go
```diff
@@ -437,6 +437,87 @@ func (b *metricBucket) Name() string {
...
+}
+
+func (pbkt *prefixedBucket) Upload(ctx context.Context, name string, r io.Reader) (err error) {
+	pname := filepath.Join(pbkt.prefix, name)
+	err = pbkt.bkt.Upload(ctx, pname, r)
+	return
+}
+
+func (pbkt *prefixedBucket) Delete(ctx context.Context, name string) (err error) {
+	pname := filepath.Join(pbkt.prefix, name)
```
pracucci, line 478: You should juse `path.Join()` and not `filepath.Join()`. The former guarantees paths are joined with `/`, which is what object stores expect.

_Same comment applies across the change set._

pkg/objstore/objstore.go
```diff
@@ -437,6 +437,87 @@ func (b *metricBucket) Name() string {
...
+	err = pbkt.bkt.Delete(ctx, pname)
+	return
+}
+
+func (pbkt *prefixedBucket) Close() error {
+	return pbkt.bkt.Close()
+}
+
+func (pbkt *prefixedBucket) Iter(ctx context.Context, dir string, f func(string) error) error {
```
pracucci, line 487: Would be possible to add a unit test on this?

pkg/objstore/objstore.go
```diff
@@ -437,6 +437,87 @@ func (b *metricBucket) Name() string {
...
+	pdir := filepath.Join(pbkt.prefix, dir)
+	if pbkt.prefix != "" {
+		return pbkt.bkt.Iter(ctx, pdir, func(s string) error {
+			return f(strings.Join(strings.Split(s, pbkt.prefix)[1:], "/"))
```
pracucci, line 491: Shouldn't be `f(strings.Join(strings.Split(s, "/")[1:], "/"))`?

pkg/objstore/objstore.go
pracucci: Aso the `.Join()` is used quite a lot. You could add an help `prefixedBucket.nameWithPrefix(name string)`

dsayan154:
Sorry for such a late response on this PR, I was busy with office deliverables. I hope to address these comments by next weekend.

pkg/objstore/client/factory.go
dsayan154: Thanks for this I'll update the PR

pkg/objstore/objstore.go
dsayan154: yeah, might be good to add a helper function but I don't know how will it reduce the redundancy of the code, as I'll have to call the helper function same number of times as I have called filepath.Join().
But I'll update the code with your suggestion. I think it will add consistency to the code.

pkg/objstore/objstore.go
dsayan154: Working on this.

pkg/objstore/objstore.go
dsayan154: I think this is fine. but would be more sure after writing a unit test

pkg/objstore/objstore.go
dsayan154: Hi @pracucci ,
I have added few unit tests for the Iter(), can you please review this? the tests are running fine locally.

pkg/objstore/objstore.go
dsayan154: The units tests for Prefixed bucket Iter are running fine locally.

ashuec90:
Getting below error while building thanos docker image after taking this PR.
```
>> building Thanos binary in /go/bin
 >   thanos
# github.com/thanos-io/thanos/pkg/objstore
pkg/objstore/objstore.go:488:10: cannot use &prefixedBucket literal (type *prefixedBucket) as type Bucket in return argument:
	*prefixedBucket does not implement Bucket (wrong type for Iter method)
		have Iter(context.Context, string, func(string) error) error
		want Iter(context.Context, string, func(string) error, ...IterOption) error
pkg/objstore/objstore.go:493:2: cannot use pbkt (type *prefixedBucket) as type Bucket in return argument:
	*prefixedBucket does not implement Bucket (wrong type for Iter method)
		have Iter(context.Context, string, func(string) error) error
		want Iter(context.Context, string, func(string) error, ...IterOption) error
!! command failed: build -o /go/bin/thanos -ldflags -X github.com/prometheus/common/version.Version=0.20.0-dev -X github.com/prometheus/common/version.Revision=82ea6ff33a65a109c6d8b70ef3687479a9f04325 -X github.com/prometheus/common/version.Branch=added_prefix -X github.com/prometheus/common/version.BuildUser=root@237fa8b74dcf -X github.com/prometheus/common/version.BuildDate=20210316-08:19:18  -extldflags '-static' -a -tags netgo github.com/thanos-io/thanos/cmd/thanos: exit status 2
make: *** [Makefile:130: build] Error 1
```

sfc-gh-jburns:
What would it take to get this done?

CHANGELOG.md
```diff
@@ -112,6 +114,16 @@ Highlights:
 - [#3114](https://github.com/thanos-io/thanos/pull/3114) Query Frontend: Added support for Memacached cache.
   - **breaking** Renamed flag `log_queries_longer_than` to `log-queries-longer-than`.
 - [#3166](https://github.com/thanos-io/thanos/pull/3166) UIs: Added UI for passing a `storeMatch[]` parameter to queries.
+<<<<<<< HEAD
```
kakkoyun, line 117: We need to clean these

kakkoyun (changes requested):

PR #3294, opened 2020-10-08
add some desc for the memcached cache for thanos-query-frontend

iyacontrol:

## Changes

add some desc  for the memcached cache for thanos-query-frontend


yeya24 (commented):
Thanks for the pr!

Since the query frontend cache configuration is similar to https://github.com/thanos-io/thanos/blob/master/docs/components/store.md#memcached-index-cache, what about adding a link to that document and remove the duplicates?

The `expiration` is newly added so we can keep it.

iyacontrol:
> Thanks for the pr!
> 
> Since the query frontend cache configuration is similar to https://github.com/thanos-io/thanos/blob/master/docs/components/store.md#memcached-index-cache, what about adding a link to that document and remove the duplicates?
> 
> The `expiration` is newly added so we can keep it.


If the cache configuration in these two places is indeed the same, then the connection is indeed the most appropriate

docs/components/compact.md
```diff
@@ -43,7 +43,7 @@ Keep in mind, that the initial goal of downsampling is not saving disk space (Re
...
 
-There's also a case when you might want to disable downsampling at all with `debug.disable-downsampling`. You might want to do it when you know for sure that you are not going to request long ranges of data (obviously, because without downsampling those requests are going to be much much more expensive than with it). A valid example of that case is when you only care about the last couple weeks of your data or use it only for alerting, but if that's your case - you also need to ask yourself if you want to introduce Thanos at all instead of just vanilla Prometheus?
+There's also a case when you might want to disable downsampling at all with `--downsampling.disable`. You might want to do it when you know for sure that you are not going to request long ranges of data (obviously, because without downsampling those requests are going to be much much more expensive than with it). A valid example of that case is when you only care about the last couple weeks of your data or use it only for alerting, but if that's your case - you also need to ask yourself if you want to introduce Thanos at all instead of just vanilla Prometheus?
```
yeya24, line 46: Very nice catch!

docs/components/query-frontend.md
```diff
@@ -85,6 +85,11 @@ config:
   expiration: 0s
 ```
 
+`expiration` Expiration specifies memcached cache valid time , If set to 0s, so using a default of 24 hours expiration time
```
yeya24, line 88: Some style issues. Would be good to change to:

`expiration` specifies Memcached cache valid time. If set to 0s, it will use 24 hours expiration time by default.

yeya24 (approved):
Overall looks good! Thanks for your contribution.

docs/components/query-frontend.md
iyacontrol: done it.

kakkoyun (approved):
💯 LGTM. Thanks a lot for the contribution.

PR #3299, opened 2020-10-09
Support web.route-prefix in query frontend

metalmatze:
I'm running the Thanos Querier on `/thanos/` as I run it behind some auth proxy that needs the prefix for routing between Prometheus, Thanos, Alertmanager and Grafana.

Currently, I'm unable to put the Query Frontend between the auth proxy and the Thanos Querier.

I'll add a changelog once this is somewhat ready to merge.


yeya24:
With this https://github.com/thanos-io/thanos/pull/3195 fix, you can already access the query frontend and a querier with route prefix configured. Just go to `/thanos/` in the query frontend then it should be accessible.

Does this satisfy your use case? The problem of having one single web route prefix in the query frontend is that we might have multiple downstream queriers with different prefix at downstream.

metalmatze:
Ah sweet. I was looking through previous changes but didn't see this one being related. Let me try this one with the new release candidate then! :+1:

metalmatze:
Tried v0.16.0-rc.0 and this works perfectly out of the box now! :+1:

PR #3310, opened 2020-10-13
[Docs] Add indent in remote write example

kubadawczynski:
Add indent in remote write example


bwplotka (commented):
Both versions works (this is how YAML works) (:

soniasingla:
@bwplotka we can close this :)

Issue #3291, opened 2020-10-07
Minio v6 -> v7 breaks for Ceph due to ListObjectsV2


sevagh:
Hello. When upgrading from Thanos 0.14.0 to 0.16.0-rc.0 (and 0.15.0 should have the same issue), this breaks with Ceph.

This is due to the minio client upgrade from v6 to v7: https://github.com/thanos-io/thanos/pull/2970/files

This breaks for me because the version of Ceph we operate has trouble with the ListObjectsV2 operation (if you look this up it's a known issue for minio/Ceph) - e.g. https://github.com/minio/minio-go/issues/1130

I can hotfix my own local clone of Thanos to add this:
```
diff --git a/pkg/objstore/s3/s3.go b/pkg/objstore/s3/s3.go
index 923a4253..26cfcc5d 100644
--- a/pkg/objstore/s3/s3.go
+++ b/pkg/objstore/s3/s3.go
@@ -309,6 +309,7 @@ func (b *Bucket) Iter(ctx context.Context, dir string, f func(string) error) err
        opts := minio.ListObjectsOptions{
                Prefix:    dir,
                Recursive: false,
+               UseV1: true,
        }
```
Maybe Ceph users are a small population that this doesn't matter.

sevagh:
As for any practical suggestions - maybe allow a "UseV1 = true" config key in the AWS bucket config file?

zygiss:
We're also running on Ceph and are affected by this.

I like the idea to have a bucket config option to revert to V1 of the ListObjects API in the interim, while we upgrade our Ceph installs.

sevagh:
I made a PR for this.

GiedriusS:
The PR has been merged so closing this! Thank you for this fix.

PR #3312, opened 2020-10-13
objstore/s3: Add list_objects_version to S3 config

sevagh:
Signed-off-by: Sevag Hanssian <sevagh@protonmail.com>

## Changes

I added an optional config "list_objects_version: string" to the S3 configuration.

## Verification

I have a bucket in production which requires ListObjectsV1 to work (I need to deploy a custom binary with minio UseV1 set to true). After making this PR, I tested "thanos bucket tools web" with the new config option and it works.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -19,6 +19,7 @@ We use *breaking :warning:* to mark changes that are not backward compatible (re
 - [#3315](https://github.com/thanos-io/thanos/pull/3315) Query Frontend: Support results caching for label names, label values and series requests.
 - [#3346](https://github.com/thanos-io/thanos/pull/3346) Ruler UI: Fix a bug preventing the /rules endpoint from loading.
 - [#3115](https://github.com/thanos-io/thanos/pull/3115) compact: now deletes partially uploaded and blocks with deletion marks concurrently. It does that at the beginning and then every `--compact.cleanup-interval` time period. By default it is 5 minutes.
+- [#3312](https://github.com/thanos-io/thanos/pull/3312) s3: add list_objects_version config option for compatibility.
 
 ### Fixed
 - [#3257](https://github.com/thanos-io/thanos/pull/3257) Ruler: Prevent Ruler from crashing when using default DNS to lookup hosts that results in "No such hosts" errors.
--- a/docs/storage.md
+++ b/docs/storage.md
@@ -90,6 +90,7 @@ config:
     insecure_skip_verify: false
   trace:
     enable: false
+  list_objects_version: ""
   part_size: 134217728
   sse_config:
     type: ""
@@ -109,6 +110,8 @@ You can configure the timeout settings for the HTTP client by setting the `http_
 Please refer to the documentation of [the Transport type](https://golang.org/pkg/net/http/#Transport) in the `net/http` package for detailed information on what each option does.
 
 `part_size` is specified in bytes and refers to the minimum file size used for multipart uploads, as some custom S3 implementations may have different requirements. A value of `0` means to use a default 128 MiB size.
+
+Set `list_objects_version: "v1"` for S3 compatible APIs that don't support ListObjectsV2 (e.g. some versions of Ceph). Default value (`""`) is equivalent to `"v2"`.
 
 For debug and testing purposes you can set
 
--- a/pkg/objstore/s3/s3.go
+++ b/pkg/objstore/s3/s3.go
@@ -59,16 +59,17 @@ var DefaultConfig = Config{
 
 // Config stores the configuration for s3 bucket.
 type Config struct {
-	Bucket          string            `yaml:"bucket"`
-	Endpoint        string            `yaml:"endpoint"`
-	Region          string            `yaml:"region"`
-	AccessKey       string            `yaml:"access_key"`
-	Insecure        bool              `yaml:"insecure"`
-	SignatureV2     bool              `yaml:"signature_version2"`
-	SecretKey       string            `yaml:"secret_key"`
-	PutUserMetadata map[string]string `yaml:"put_user_metadata"`
-	HTTPConfig      HTTPConfig        `yaml:"http_config"`
-	TraceConfig     TraceConfig       `yaml:"trace"`
+	Bucket             string            `yaml:"bucket"`
+	Endpoint           string            `yaml:"endpoint"`
+	Region             string            `yaml:"region"`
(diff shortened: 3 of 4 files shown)


sevagh:
Solves https://github.com/thanos-io/thanos/issues/3291

kakkoyun (approved):
LGTM. Thanks for your contributions.

Let's wait for another approval and then we can merge it.

sevagh:
> ci/circleci: test — Your tests failed on CircleCI 
I can't see what failed here. Is it related to my code?

docs/storage.md
```diff
@@ -110,6 +111,8 @@ Please refer to the documentation of [the Transport type](https://golang.org/pkg
 
 `part_size` is specified in bytes and refers to the minimum file size used for multipart uploads, as some custom S3 implementations may have different requirements. A value of `0` means to use a default 128 MiB size.
 
+Set `list_objects_v1: true` for S3 compatible APIs that don't support ListObjectsV2 (e.g. some versions of Ceph).
```
bwplotka, line 114: I would vote for `list_objects_version: "v1"` with default being `"v2"`. It might be much more flexible, WDYT? (:

bwplotka (commented):
Thanks! One suggestion for flexibility and LGTM (:

docs/storage.md
sevagh: So @bwplotka should I make the switch? I'll defer to your opinion.

My only preference for this PR is that it gets merged into main sooner rather than later so I can deploy a real upstream Thanos (and not my own binary :))

bwplotka: Yes please! Let's switch. It's a good practice overall for the future (what if we change S3 client to different one?) 

That's the last suggestion, I think once it's done we are good to merge :hugs:

sevagh: Question. I am choosing to set it as a string in the `type Config struct`, and leave it as a bool in `type Bucket string` (which contains a minio.Client as a field, so it makes sense to have a minio-specific in that struct).

The function NewBucketWithConfig is a good place to store the logic of parsing `""|"v2"` or `v1` into the appropriate bool value. Should I break the parsing logic into a function so I can unit test it?
```
var useListObjectsV1 bool
if config.ListObjectsVersion == "" || config.ListObjectsVersion == "v2" {
    useListObjectsV1 = false
} else if config.ListObjectsVersion == "v1" {
    useListObjectsV1 = true
} else {
    listObjectsErrMsg := errors.Errorf("Unsupported version %q was provided. Supported values are v1, v2", config.ListObjectsVersion)
    return nil, errors.Wrap(listObjectsErrMsg, "Initialize s3 client list objects version")
}
```

sevagh: It's more clear if I push commit (just did).

pkg/objstore/s3/s3.go
```diff
@@ -246,13 +248,24 @@ func NewBucketWithConfig(logger log.Logger, config Config, component string) (*B
...
+	var useListObjectsV1 bool
+	if config.ListObjectsVersion == "" || config.ListObjectsVersion == "v2" {
+		useListObjectsV1 = false
+	} else if config.ListObjectsVersion == "v1" {
+		useListObjectsV1 = true
+	} else {
+		listObjectsErrMsg := errors.Errorf("Unsupported version %q was provided. Supported values are v1, v2", config.ListObjectsVersion)
+		return nil, errors.Wrap(listObjectsErrMsg, "Initialize s3 client list objects version")
+	}
+
```
sevagh, line 260: I'm wondering if this code needs a unit test.

bwplotka: Probably, but also this code can be simplified to:

```
	if config.ListObjectsVersion =! "" && c config.ListObjectsVersion =! "v1" && config.ListObjectsVersion != "v2" {
		return nil, errors.Errorf(Initialize s3 client list objects version: Unsupported version %q was provided. Supported values are v1, v2", config.ListObjectsVersion)
       }
       return Bucket{ 
           .....
           listObjectsV1:  config.ListObjectsVersion == "v1",
}
```

bwplotka (approved):
Small suggestion and LGTM (:

pkg/objstore/s3/s3.go
sevagh: Nice one, wish I wrote it that way the first time.

sevagh:
I rebased on the tip of master to fix the CHANGELOG conflict. This should be OK to merge if it passes the checks?

CHANGELOG.md
```diff
@@ -19,11 +19,13 @@ We use *breaking :warning:* to mark changes that are not backward compatible (re
 - [#3315](https://github.com/thanos-io/thanos/pull/3315) Query Frontend: Support results caching for label names, label values and series requests.
 - [#3346](https://github.com/thanos-io/thanos/pull/3346) Ruler UI: Fix a bug preventing the /rules endpoint from loading.
 - [#3115](https://github.com/thanos-io/thanos/pull/3115) compact: now deletes partially uploaded and blocks with deletion marks concurrently. It does that at the beginning and then every `--compact.cleanup-interval` time period. By default it is 5 minutes.
+- [#3312](https://github.com/thanos-io/thanos/pull/3312) s3: add list_objects_version config option for compatibility.
 
 ### Fixed
 - [#3257](https://github.com/thanos-io/thanos/pull/3257) Ruler: Prevent Ruler from crashing when using default DNS to lookup hosts that results in "No such hosts" errors.
 - [#3331](https://github.com/thanos-io/thanos/pull/3331) Disable Azure blob exception logging
 - [#3341](https://github.com/thanos-io/thanos/pull/3341) Disable Azure blob syslog exception logging
+- [#3276](https://github.com/thanos-io/thanos/pull/3276) Query Frontend: Support query splitting and retry for labels and series requests.
```
bwplotka, line 28: Hm, I think rebase went wrong here?

bwplotka (approved):
LGTM, just one thing - rebase went wrong?

CHANGELOG.md
sevagh: The CHANGELOG looks different from the time I had to modify it: https://github.com/thanos-io/thanos/commit/57076a58680c8a318aefa2630ec6319f549acd56#diff-06572a96a58dc510037d5efa622f9bec8519bc1beab13c9f251e97e657a9d4ed

This has me confused, TBH.

sevagh: I just fixed it manually.

bwplotka:
Thanks!

PR #3321, opened 2020-10-15
Fix thanos.shipper.json uploaded blocks

pracucci:

## Changes

In Cortex we would like to use `thanos.shipper.json` as source of truth to know if a specific block has been shipped to the storage. We found an edge case where `thanos.shipper.json` doesn't actually get updated with the whole list of blocks shipper and this PR should fix it.

Change is not relevant to the end user because doesn't affect Thanos.

## Verification

N/A


yeya24 (approved):
LGTM

kakkoyun (approved):
💯 LGTM. Thanks a lot :)

pstibrany (commented):
LGTM.

Issue #3057, opened 2020-08-19
store panics with "slice bounds out of range"


csonp:
**Thanos, Prometheus and Golang version used**:

prometheus/prometheus:v2.19.0
thanos/thanos:v0.14.0

**Object Storage Provider**:

S3

**What happened**:

When running a query with a 90 days timespan, the store panics.

**Full logs to relevant components**:

Logs from store

<details>
<pre>
level=debug ts=2020-08-19T07:50:51.943767339Z caller=main.go:102 msg="maxprocs: Leaving GOMAXPROCS=[2]: CPU quota undefined"
level=info ts=2020-08-19T07:50:51.943987303Z caller=main.go:151 msg="Tracing will be disabled"
level=info ts=2020-08-19T07:50:51.944096933Z caller=factory.go:46 msg="loading bucket configuration"
level=info ts=2020-08-19T07:50:51.944459648Z caller=inmemory.go:172 msg="created in-memory index cache" maxItemSizeBytes=131072000 maxSizeBytes=262144000 maxItems=maxInt
level=info ts=2020-08-19T07:50:51.944766235Z caller=options.go:23 protocol=gRPC msg="disabled TLS, key and cert must be set to enable"
level=info ts=2020-08-19T07:50:51.944835531Z caller=grpc.go:95 service=gRPC/server component=store msg="registering as gRPC StoreAPI"
level=info ts=2020-08-19T07:50:51.945174296Z caller=store.go:374 msg="starting store node"
level=info ts=2020-08-19T07:50:51.945296817Z caller=store.go:321 msg="initializing bucket store"
level=info ts=2020-08-19T07:50:51.946285789Z caller=intrumentation.go:60 msg="changing probe status" status=healthy
level=info ts=2020-08-19T07:50:51.946304337Z caller=http.go:56 service=http/server component=store msg="listening for requests and metrics" address=0.0.0.0:10902
level=info ts=2020-08-19T07:50:52.504771373Z caller=fetcher.go:452 component=block.BaseFetcher msg="successfully synchronized block metadata" duration=559.439362ms cached=65 returned=65 partial=0
level=debug ts=2020-08-19T07:50:52.506595828Z caller=bucket.go:447 msg="loading new block" id=01EFGC2D7YY1A38MY13AH8DHSA
level=debug ts=2020-08-19T07:50:52.506719259Z caller=bucket.go:447 msg="loading new block" id=01EF0Y4DX4XX57WEVFS39F8Z7R
[...]

PR #3322, opened 2020-10-15
store: Prevent store from panic after range object storage

MegaEle:
Signed-off-by: MegaEle <ajmy789456123@gmail.com>

## Changes

When the store uses the object store as the s3 interface, there is a probability that the loadChunks or loadSeries method panic will appear in the query. It is found that the s3 interface does not return complete data, and the store should return an error instead of panic

## Verification

I compared it after build the modified code. Under the same circumstances, the store will not panic, and the query returns the following error like:
 Mint: 1601021768826 Maxt: 1602734400000: rpc error: code = Aborted desc = fetch series for block 01EMMWANSPNASD41ER54W9B2WQ: preload chunks: read range for 0: can't read chunk for required length, required 2723780, actually 48762

### DIFF
--- a/pkg/store/bucket.go
+++ b/pkg/store/bucket.go
@@ -1369,7 +1369,11 @@ func (b *bucketBlock) readIndexRange(ctx context.Context, off, length int64) ([]
 	if _, err := buf.ReadFrom(r); err != nil {
 		return nil, errors.Wrap(err, "read range")
 	}
-	return buf.Bytes(), nil
+	internalBuf := buf.Bytes()
+	if int64(len(internalBuf)) != length {
+		return nil, fmt.Errorf("can't read index for required length, required %d, actually %d", length, len(internalBuf))
+	}
+	return internalBuf, nil
 }
 
 func (b *bucketBlock) readChunkRange(ctx context.Context, seq int, off, length int64) (*[]byte, error) {
@@ -1396,6 +1400,10 @@ func (b *bucketBlock) readChunkRange(ctx context.Context, seq int, off, length i
 		return nil, errors.Wrap(err, "read range")
 	}
 	internalBuf := buf.Bytes()
+	if int64(len(internalBuf)) != length {
+		b.chunkPool.Put(c)
+		return nil, fmt.Errorf("can't read chunk for required length, required %d, actually %d", length, len(internalBuf))
+	}
 	return &internalBuf, nil
 }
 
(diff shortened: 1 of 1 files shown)


pkg/store/bucket.go
```diff
@@ -1908,7 +1909,11 @@ func (r *bucketIndexReader) loadSeries(ctx context.Context, ids []uint64, refetc
 	r.mtx.Unlock()
 
 	for i, id := range ids {
-		c := b[id-start:]
+		seriesOffset := id - start
+		if int(seriesOffset) > indexLength {
```
pracucci, line 1913: This just checks the offset, but the following code could overread anyway. Am I missing anything?

pracucci (commented):
Hi!

> When the store uses the object store as the s3 interface, there is a probability that the loadChunks or loadSeries method panic will appear in the query. It is found that the s3 interface does not return complete data, and the store should return an error instead of panic

I would be interested into hearing about the issue. The protection you propose here is ok (unless will turn out to add too much complexity) but I'm not sure I understand the root cause of this issue, which is more interesting to me. Have you opened an issue with details about it?

pkg/store/bucket.go
MegaEle: Hi, I found the following issue, we encountered the same issue.
https://github.com/thanos-io/thanos/issues/3057

When I debugged, I found that when using readIndexRange and readChunkRange, the length of the data is not necessarily equal to or longer than the required length of the data, but when we use it, we will set the offset according to the length, so there will be situations .

I use the s3 interface and the underlying ceph storage. I am still looking for ways to solve the root cause of the incorrect data length, but there is no progress. In addition, I hope the store will not panic in this case

I tried to modify the readChunkRange and readIndexRange methods, but I found that other upstream methods do not require the returned data length. So I only modified the place where panic occurs : )

GiedriusS: @MegaEle do you use radosgw, perchance? I have seen the same thing a few times but never got to it. It's probably some kind of bug there in the API in that it doesn't follow the spec. Have you ever found a way to consistently reproduce it?

sevagh: Not sure if the issue I just encountered is also relevant: https://github.com/thanos-io/thanos/issues/3339

I use Radosgw/ceph to confirm what @GiedriusS says

GiedriusS:
I think https://github.com/thanos-io/thanos/pull/3621 is a more reasonable defensive coding measure than this one because that other change adds a checker that is closer to the actual problem.

GiedriusS:
This has been fixed by https://github.com/thanos-io/thanos/pull/3795, thank you for your contribution!

PR #3325, opened 2020-10-15
Improve store to not load chunks from obj store for Series API

yeya24:

## Changes

Using the similar idea of https://github.com/prometheus/prometheus/pull/8050 to only fetch labels for `/api/v1/series` API. Changes:

1. Only sort and dedup labels when there is at least one chunk in the specified time range. If no chunk is valid, then not necessary to sort labels.

2. Series query not load chunks.

### DIFF
--- a/pkg/store/bucket.go
+++ b/pkg/store/bucket.go
@@ -741,12 +741,13 @@ func blockSeries(
 
 		// Reserve chunksLimiter if we save chunks.
 		if len(s.chks) > 0 {
+			hasValidChunk = true
 			if err := chunksLimiter.Reserve(uint64(len(s.chks))); err != nil {
 				return nil, nil, errors.Wrap(err, "exceeded chunks limit")
 			}
 		}
 
-		if hasValidChunk || len(s.chks) > 0 {
+		if hasValidChunk {
 			for _, l := range lset {
 				// Skip if the external labels of the block overrule the series' label.
 				// NOTE(fabxc): maybe move it to a prefixed version to still ensure uniqueness of series?
--- a/pkg/store/bucket.go
+++ b/pkg/store/bucket.go
@@ -797,8 +797,7 @@ func blockSeriesWithoutChunks(
 		return nil, nil, errors.Wrap(err, "preload series")
 	}
 
-	// Transform all series into the response types and mark their relevant chunks
-	// for preloading.
+	// Transform all series into the response types.
 	var (
 		res  []seriesEntry
 		lset labels.Labels
--- a/pkg/store/bucket.go
+++ b/pkg/store/bucket.go
@@ -712,18 +712,6 @@ func blockSeries(
 			refs: make([]uint64, 0, len(chks)),
 			chks: make([]storepb.AggrChunk, 0, len(chks)),
 		}
-		for _, l := range lset {
-			// Skip if the external labels of the block overrule the series' label.
-			// NOTE(fabxc): maybe move it to a prefixed version to still ensure uniqueness of series?
-			if extLset[l.Name] != "" {
-				continue
-			}
-			s.lset = append(s.lset, l)
-		}
-		for ln, lv := range extLset {
-			s.lset = append(s.lset, labels.Label{Name: ln, Value: lv})
-		}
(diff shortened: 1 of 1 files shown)


pracucci:
@gouthamve FYI, given you're working on this on Cortex side too.

pkg/store/bucket.go
```diff
@@ -772,6 +773,75 @@ func blockSeries(
 	return newBucketSeriesSet(res), indexr.stats.merge(chunkr.stats), nil
 }
 
+// blockSeriesWithoutChunks is the same as blockSeries function except it only
+// fetches labels and doesn't touch chunks. This is used in /api/v1/series API.
+func blockSeriesWithoutChunks(
```
pracucci, line 778: Do we need to duplicate `blockSeries()`? How more dirty would be `blockSeries()` code if we add a "skipChunks" parameter to that function? Or could we make it implicitly skipping chunks if `chunkr` is `nil`?

pracucci (commented):
Thanks for working on it! I have a design question.

pkg/store/bucket.go
yeya24: Thanks, let me try

yeya24: Done. Please take a look!

yeya24:
Tested this pr using our tool https://github.com/observatorium/up to send `/api/v1/series` request.

Red line is master and green line is this pr. Seems this is faster.

Screenshot from 2020-10-19 11-59-34

GiedriusS (approved):
Looks good to me! Much cleaner now that it is in the same function, IMHO. But let's wait for another review because this is a relatively critical path

pkg/store/bucket.go
```diff
@@ -712,19 +712,9 @@ func blockSeries(
...
-			}
-			s.lset = append(s.lset, l)
-		}
-		for ln, lv := range extLset {
-			s.lset = append(s.lset, labels.Label{Name: ln, Value: lv})
-		}
-		sort.Sort(s.lset)
 
+		// find is used to check whether there is at least one chunk in the required time range.
+		var find bool
```
kakkoyun, line 717: Should this be named `skipped` or `found`? 
```suggestion
		var skipped bool
```
The comment says:
> find is used to check whether there is at least one chunk in the required time range.

But it's only set the `true` when it's skipped. I'm a bit confused here about the naming. Could we be more clear here?

pkg/store/bucket.go
```diff
@@ -742,15 +738,35 @@ func blockSeries(
 			})
 			s.refs = append(s.refs, meta.Ref)
 		}
-		if len(s.chks) > 0 {
-			if err := chunksLimiter.Reserve(uint64(len(s.chks))); err != nil {
-				return nil, nil, errors.Wrap(err, "exceeded chunks limit")
+		if find || len(s.chks) > 0 {
+			// Reserve chunksLimiter if we save chunks.
+			if len(s.chks) > 0 {
```
kakkoyun, line 743: I think moving this outside the enclosing `if` could make it more readable.

kakkoyun (changes requested):
Looks great. I have only a couple of nits.

pkg/store/bucket.go
yeya24: Thanks. This makes sense!

pkg/store/bucket.go
yeya24: Done

pkg/store/bucket.go
```diff
@@ -742,15 +738,36 @@ func blockSeries(
 			})
 			s.refs = append(s.refs, meta.Ref)
 		}
+
+		// Reserve chunksLimiter if we save chunks.
 		if len(s.chks) > 0 {
 			if err := chunksLimiter.Reserve(uint64(len(s.chks))); err != nil {
```
kakkoyun, line 744: How about setting hasValidChunk to true in the clause and just checking it in below `if`? Would this work?

yeya24: Done.

kakkoyun (approved):
💯 🥇 💜

Issue #2565, opened 2020-05-05
Thanos compactor filling the disk with stack traces on Azure


joaosilva15:
**Thanos, Prometheus and Golang version used**:

Thanos: 012.2
Go: 1.13.6

**Object Storage Provider**:

Azure

**What happened**:
After the upgrade to the newest version of thanos compactor the logging has been filled with stack traces from the azure sdk.
The error states that the deletion-mark.json cannot be found and that the api returned 404 (as expected as the blocks are not to be deleted)

**What you expected to happen**:

No stack traces on the logs as the error is expected.

**How to reproduce it (as minimally and precisely as possible)**:

Running the compactor against the blob storage is enough in our case.

**Full logs to relevant components**:

```
[...]

PR #3331, opened 2020-10-16
Disable Azure blob exception logging

pharaujo:
Ever since the introduction of the deletion marker, Thanos compact has been filling the logs with Azure blob storage exceptions. As discussed in #2565, the error logging should be handled by Thanos itself, so this patch turns off standard logging from the Azure library.

However, [due to the way logging is set up in the library][1], it will always log errors to syslog, regardless of how `ShouldLog` is configured. As such, and until that can be a configurable behavior, care should be taken to handle that from the syslog side as well.

[1]: https://github.com/Azure/azure-storage-blob-go/blob/48358e1de5110852097ebbc11c53581d64d47300/azblob/zc_policy_request_log.go#L100-L102

Signed-off-by: Pedro Araujo <phcrva@gmail.com>

## Changes

Disabled exception logging from the Azure Blob storage module

## Verification

Compiled locally and ran `thanos compact` against an Azure blob storage container.

### DIFF
--- a/pkg/objstore/azure/helpers.go
+++ b/pkg/objstore/azure/helpers.go
@@ -10,6 +10,7 @@ import (
 	"regexp"
 	"time"
 
+	"github.com/Azure/azure-pipeline-go/pipeline"
 	blob "github.com/Azure/azure-storage-blob-go/azblob"
 )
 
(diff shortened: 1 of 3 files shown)


pkg/objstore/azure/helpers.go
```diff
@@ -34,6 +35,12 @@ func getContainerURL(ctx context.Context, conf Config) (blob.ContainerURL, error
 	p := blob.NewPipeline(c, blob.PipelineOptions{
 		Retry:     retryOptions,
 		Telemetry: blob.TelemetryOptions{Value: "Thanos"},
+		RequestLog: blob.RequestLogOptions{
+			LogWarningIfTryOverThreshold: -1,
```
bwplotka, line 39: Can we comment why -1 here? (:

bwplotka (approved):
LGTM, just would be nice to comment on surprises. Value -1 turning off something is super weird (: 

Thanks!

pkg/objstore/azure/helpers.go
pharaujo: Comment added.

pharaujo:
> LGTM, just would be nice to comment on surprises. Value -1 turning off something is super weird (:

Yeah, I agree. The whole logging method is super weird, it forces logging even if you explicitly request for it to not log. I plan on at least opening an issue for that on their end.

pharaujo:
@bwplotka I didn't change anything since the previous commit that would cause the failed test, only added a comment as discussed. How should I proceed?

kakkoyun (approved):
LGTM

PR #3342, opened 2020-10-20
cmd/thanos: Extract logger instantiation to pkg/logging

brancz:

## Changes

Extract logger instantiation to the `pkg/logging` package.

## Verification

Moved code, and still compiles :) 

@thanos-io/thanos-maintainers

### DIFF
--- a/pkg/logging/logger.go
+++ b/pkg/logging/logger.go
@@ -1,0 +1,55 @@
+// Copyright (c) The Thanos Authors.
+// Licensed under the Apache License 2.0.
+
+package logging
+
+import (
+	"fmt"
+	"os"
+
+	"github.com/go-kit/kit/log"
+	"github.com/go-kit/kit/log/level"
+)
+
+const (
+	LogFormatLogfmt = "logfmt"
+	LogFormatJSON   = "json"
+)
+
+// NewLogger returns a log.Logger that prints in the provided format with a UTC
+// timestamp and the caller of the log entry.
+func NewLogger(logFormat, debugName string) log.Logger {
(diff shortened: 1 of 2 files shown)


kakkoyun (approved):
🥇 LGTM

pkg/logging/logger.go
```diff
@@ -0,0 +1,46 @@
...
+	case "error":
+		lvl = level.AllowError()
+	case "warn":
+		lvl = level.AllowWarn()
+	case "info":
+		lvl = level.AllowInfo()
+	case "debug":
+		lvl = level.AllowDebug()
+	default:
+		panic("unexpected log level")
```
squat, line 31: I realize that we had the same behavior before, but I don't think that panicking is the right action to take when a user mistypes a flag. If this was only possible in a program error, then I would agree, but since it's possible when typing `--log.level=eror` I think it's too severe.

brancz: Refactored this a bit, let me know what you think about this approach.

pkg/logging/logger.go
```diff
@@ -0,0 +1,55 @@
...
+import (
+	"fmt"
+	"os"
+
+	"github.com/go-kit/kit/log"
+	"github.com/go-kit/kit/log/level"
+)
+
+const (
+	LogFormatLogfmt = "logfmt"
```
bwplotka, line 15:
```suggestion
	FormatLogfmt = "logfmt"
```
We can use package name now :hugs:

bwplotka (approved):
LGTM!

pkg/logging/logger.go
brancz: hmm not sure, I actually think it's clearer with the prefix

PR #3358, opened 2020-10-24
Cleanup some used by Cortex methods in compact.go

yeya24:

### DIFF
--- a/pkg/compact/compact.go
+++ b/pkg/compact/compact.go
@@ -394,56 +394,9 @@ func (cg *Group) Add(meta *metadata.Meta) error {
 	return nil
 }
 
-// IDs returns all sorted IDs of blocks in the group.
-func (cg *Group) IDs() (ids []ulid.ULID) {
-	cg.mtx.Lock()
-	defer cg.mtx.Unlock()
-
-	for id := range cg.blocks {
-		ids = append(ids, id)
-	}
-	sort.Slice(ids, func(i, j int) bool {
-		return ids[i].Compare(ids[j]) < 0
-	})
-	return ids
-}
-
-// MinTime returns the min time across all group's blocks.
-func (cg *Group) MinTime() int64 {
-	cg.mtx.Lock()
-	defer cg.mtx.Unlock()
-
-	min := int64(0)
-	for _, b := range cg.blocks {
-		if b.MinTime < min {
-			min = b.MinTime
-		}
-	}
-	return min
-}
-
-// MaxTime returns the max time across all group's blocks.
-func (cg *Group) MaxTime() int64 {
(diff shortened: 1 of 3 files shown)


pkg/compact/compact.go
```diff
@@ -394,58 +394,11 @@ func (cg *Group) Add(meta *metadata.Meta) error {
...
-	cg.mtx.Lock()
-	defer cg.mtx.Unlock()
-
-	max := int64(0)
-	for _, b := range cg.blocks {
-		if b.MaxTime < max {
-			max = b.MaxTime
-		}
-	}
-	return max
```
yeya24, line 436: This always returns 0.

pkg/compact/compact.go
```diff
@@ -394,58 +394,11 @@ func (cg *Group) Add(meta *metadata.Meta) error {
...
-	cg.mtx.Lock()
-	defer cg.mtx.Unlock()
-
-	min := int64(0)
-	for _, b := range cg.blocks {
-		if b.MinTime < min {
-			min = b.MinTime
-		}
-	}
-	return min
```
yeya24, line 422: This always returns 0

pracucci (changes requested):
Would you mind keeping these functions, please? We're leveraging on it in Cortex / Grafana Metrics Enterprise.

bwplotka:
Yup, we need to think about external users of our packages.

yeya24:
Thanks, I will close this pr. My question is that the `MaxTime` and `MinTime` methods seem wrong. Doesn't they always return 0?

pkg/compact/compact.go
bwplotka: You are right - Let's fix it. `min := int64(0)` should be `min := math.MaxInt64`

pkg/compact/compact.go
bwplotka: You are right - Let's fix it. max := int64(0) should be max := math.MinInt64 Also it should be ` b.MaxTime > max`

I don't think it was well implemented so wonder how those were used....  cc @pracucci 

Let's add tests as well? (:

bwplotka (commented):
Well yea, it's totally bugged, let's at least fix it. Thanks for spotting this! :hugs:

pkg/compact/compact.go
pracucci: They're definitely wrong 🤔  I will double check it where we use them. Thanks!

yeya24:
OK. Let me fix them in this pr.

prmsrswt (approved):

pracucci (approved):
Thanks!

bwplotka (approved):
Thanks!

PR #3359, opened 2020-10-24
Only alloc labelsets in bucket series entry when it is a Series query

yeya24:

## Changes

It is not necessary to alloc `s.refs` and `s.chks` when we don't need to load chunks.

### DIFF
--- a/pkg/store/bucket.go
+++ b/pkg/store/bucket.go
@@ -707,10 +707,10 @@ func blockSeries(
 		if err := indexr.LoadedSeries(id, &lset, &chks); err != nil {
 			return nil, nil, errors.Wrap(err, "read series")
 		}
-		s := seriesEntry{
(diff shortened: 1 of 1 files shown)


yeya24:
Seems the ut failure is not related.

bwplotka (approved):
Yes :+1: 

Thanks!

bwplotka:
All of this is nice, but we need to have some benchmarks results after each small improvements like this. Let's focus to have eas simple framework as possible for this.

Wonder if you can run (for your own practice as well) some benchmarks for Series benchmark we wrote? (before and after)? Probably we need to indeed extend / copy this for skipChunk option as well. Not a blocker (:

yeya24:
Thanks, I will add the Series benchmark for skipChunk in the next pr.

Issue #3113, opened 2020-09-02
query-frontend http: invalid Read on closed Body


bwplotka:
`level=warn ts=2020-09-02T10:32:32.374998059Z caller=frontend.go:198 org_id=fake traceID=1ea32ae91e7a739b msg="unable to parse form for request" err="http: invalid Read on closed Body"`

cc @yeya24

yeya24:
@bwplotka Already opened an issue on the Cortex side. https://github.com/cortexproject/cortex/issues/3111

melkorm:
Fixed in https://github.com/cortexproject/cortex/pull/3276

bwplotka:
Thanks @melkorm !

bwplotka:
Actually let's close once we pull Cortex dep! (:

yeya24:
Thanks a lot for the work @melkorm!

PR #3370, opened 2020-10-27
Add query-frontend.max-body-size to correctly parse query response body

yeya24:

## Changes

Fixes https://github.com/thanos-io/thanos/issues/3113

This pr bumps Cortex dependency. 

Prometheus is updated to commit https://github.com/prometheus/prometheus/commits/master?before=daf382e4a9f5ca380b2b662c8e60755a56675f14+35&branch=master.


GiedriusS:
I'm thinking maybe we need to provide more guidance on this parameter, in the docs somewhere? It will be probably unclear from the help message to our users. When should this value be changed? If it shouldn't then maybe it shouldn't be tunable at all?

yeya24:
@GiedriusS Thanks! This makes sense. 

Based on the original [pr](https://github.com/cortexproject/cortex/commit/747d5950481b859c80c1d6727a2066a5f1e2f02c#diff-06572a96a58dc510037d5efa622f9bec8519bc1beab13c9f251e97e657a9d4edR51), now query frontend reads every request and this flag is used to control the max request body size to read.

This will be useful when a POST request body exceeds 10MB. Umm, do you think it is good to be tunable? Maybe I should make it not tunable since 10MB is a very large number for the request body I think.

yeya24:
I updated the pr and removed that flag. PTAL again, thanks!

bwplotka (approved):
Thanks! :+1: 

e2e tests fails only.

yeya24:
That e2e test case seems flaky. Will take a look later.

bwplotka:
LGTM!

PR #3382, opened 2020-10-30
Upgraded Prometheus and Cortex dependencies

pracucci:

## Changes

We fixed a TSDB WAL corruption which we urgently need to backport to Cortex and deploy (we already got 2 incidents this week because of this).

Prometheus has recently done a refactoring of `MultiError`, which introduced a circular breaking change in Cortex and Thanos. To solve this, I've interned `MultiError` into Cortex first (https://github.com/cortexproject/cortex/pull/3425) and in this PR I'm upgrading Cortex and Prometheus. Thanos does an use the `MultiError` data type assertion in compactor and receiver, but the data type is not exposed anymore, so to unblock this situation I've interned the previous version of `MultiError` in Thanos, in order to gain some time to discuss how to better address it, while being able to rollout the WAL corruption fix to Cortex.

Thoughts?

## Verification

Existing tests.


bwplotka (approved):
Thanks! That looks like ok approach to me :+1: 

Some CI failures though.

bwplotka:
Let me think about priv type problem :thi

bwplotka:
Double checked and this looks fair to me.

I would use Prometheus tsdb multi error though everywhere, except where we assert error type. We should probably use local type for this to explicitness. We can fix later this readability suggestion :+1: 

LGTM

PR #3383, opened 2020-10-30
scripts/quickstart: connect Ruler to Query, fix recording rule

GiedriusS:
Move Thanos Ruler execution a bit earlier, before Thanos Query, so that
we could connect it to Thanos Query using the STORES variable. Also, fix
the recording rule so that it would use a more "universal" metric. On my
machine I couldn't find `http_inprogress_requests` so it didn't work
OOB.

## Verification

Ran the script and was able to query the result of the recording rule via Thanos Query on http://localhost:10904/ that is connected to Thanos Ruler.

### DIFF
--- a/scripts/quickstart.sh
+++ b/scripts/quickstart.sh
@@ -74,8 +74,8 @@ cat >data/rules.yml <<-EOF
 	groups:
 	  - name: example
 	    rules:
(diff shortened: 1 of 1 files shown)


yeya24 (approved):
LGTM

kakkoyun (approved):
🥇

PR #3384, opened 2020-10-30
Refactored verifier for easier verifiers addition.

bwplotka:

### DIFF
--- a/cmd/thanos/tools_bucket.go
+++ b/cmd/thanos/tools_bucket.go
@@ -51,18 +51,12 @@ import (
 const extpromPrefix = "thanos_bucket_"
 
 var (
-	issuesMap = map[string]verifier.Issue{
-		verifier.IndexIssueID:                verifier.IndexIssue,
-		verifier.OverlappedBlocksIssueID:     verifier.OverlappedBlocksIssue,
-		verifier.DuplicatedCompactionIssueID: verifier.DuplicatedCompactionIssue,
-	}
-	allIssues = func() (s []string) {
-		for id := range issuesMap {
-			s = append(s, id)
-		}
-
-		sort.Strings(s)
-		return s
+	issuesVerifiersRegistry = verifier.Registry{
+		Verifiers: []verifier.Verifier{verifier.OverlappedBlocksIssue{}},
+		VerifierRepairers: []verifier.VerifierRepairer{
+			verifier.IndexKnownIssues{},
+			verifier.DuplicatedCompactionBlocks{},
+		},
 	}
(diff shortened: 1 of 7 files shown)


cmd/thanos/tools_bucket.go
```diff
@@ -130,38 +126,32 @@ func registerBucketVerify(app extkingpin.AppClause, objStoreConfig *extflag.Path
...
 		if err != nil {
 			return err
 		}
 
-		if *repair {
-			v = verifier.NewWithRepair(logger, reg, bkt, backupBkt, fetcher, time.Duration(*deleteDelay), issues)
-		} else {
-			v = verifier.New(logger, reg, bkt, fetcher, time.Duration(*deleteDelay), issues)
-		}
+		ctx := context.Background()
```
kakkoyun, line 139: nit: Shall we move closer to where it's first used?

pkg/verifier/verify.go
```diff
@@ -18,85 +19,154 @@ import (
...
+type VerifierRepairer interface {
+	IssueID() string
+	VerifyRepair(ctx context.Context, conf Config, idMatcher func(ulid.ULID) bool, repair bool) error
 }
 
-func newVerifierMetrics(reg prometheus.Registerer) *verifierMetrics {
-	var m verifierMetrics
+// Config is an verifier config.
+type Config struct {
+	Logger      log.Logger
```
kakkoyun, line 34: I think this shouldn't be part of the config semantically. We should pass logger and metrics directly, (I know it'd mess up the interface however it's more sound in that way) or rather in `Context` even though it's not safe enough. 

We have similar patterns in the code base, but this is something that disturbs me. We should find a better pattern do pass this type of global dependencies (GlobalContext?).

kakkoyun (commented):
Looks good. I have added a couple of comments.

pkg/verifier/verify.go
bwplotka: I thought about context but it's also wrong ;p

bwplotka: any suggestions?

bwplotka:
Will fix in another PR (: Have something now on top of this :hugs:

pkg/verifier/verify.go
kakkoyun: In this particular implementation I'd suggest:
1. Not to log or increment in the implementations of the interfaces. Try to do that on the wrapper modules
2. Or just create another struct and pass that to the implementations. 

WDYT?

kakkoyun: Or we could use constructors to pass down those dependencies.

PR #3392, opened 2020-11-02
query-frontend: make POST-request to downstream url for query-range

2nick:
query-frontend: make POST-request to downstream url for query-range to avoid HTTP 414 request-URI too large

### DIFF
--- a/pkg/queryfrontend/queryrange_codec.go
+++ b/pkg/queryfrontend/queryrange_codec.go
@@ -4,6 +4,7 @@
 package queryfrontend
 
 import (
+	"bytes"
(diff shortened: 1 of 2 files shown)


yeya24 (approved):
Thanks for the pr! This makes sense. LGTM.

Would you mind also updating the labels API in https://github.com/thanos-io/thanos/blob/master/pkg/queryfrontend/labels_codec.go#L174? You can do this in a separate pr.

2nick:
I'll do it as separate pr in few days

bwplotka:
Thanks! :heart:

PR #3396, opened 2020-11-03
meta.json: Added meta.Thanos.Files section to meta.JSON; Added version

bwplotka:
Not necessarily required, but it would be nice to have to reduce Attr API calls and have it for debugging purposes.

cc @pstibrany as I marked one meta.json field as Deprecated (although not plan to remove it yet)

### DIFF
--- a/pkg/block/block.go
+++ b/pkg/block/block.go
@@ -14,6 +14,7 @@ import (
 	"os"
 	"path"
 	"path/filepath"
+	"sort"
 	"strings"
 	"time"
 
@@ -66,6 +67,7 @@ func Download(ctx context.Context, logger log.Logger, bucket objstore.Bucket, id
 // It makes sure cleanup is done on error to avoid partial block uploads.
 // It also verifies basic features of Thanos block.
 // TODO(bplotka): Ensure bucket operations have reasonable backoff retries.
+// NOTE: Upload updates `meta.Thanos.File` section.
 func Upload(ctx context.Context, logger log.Logger, bkt objstore.Bucket, bdir string) error {
 	df, err := os.Stat(bdir)
 	if err != nil {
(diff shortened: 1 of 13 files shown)


pkg/block/metadata/meta.go
```diff
@@ -54,14 +54,30 @@ type Meta struct {
...
+	// Sorted by relative path.
+	// Useful to avoid API call to get size of each file, as well as for debugging purposes.
+	// Optional, added in v0.17.0.
+	Files []File `json:"files,omitempty"`
+}
+
+type File struct {
+	RelPath string `json:"rel_path"`
+	// SizeBytes is optional (e.g meta.json does not show size).
+	SizeBytes int64 `json:"size_bytes,omitempty"`
```
prmsrswt, line 80: We are using `camelCase` in all other parts of the `meta.json` so I think it makes sense to have these properties in camelcase as well.

I noticed that the `segment_files` is also not in camel case, I am not sure how much impact it will have on existing installations if we change it to `segmentFiles`.

prmsrswt (approved):
Just a minor nit. Everything else looks good to me. :)

pkg/block/metadata/meta.go
bwplotka: > I noticed that the segment_files is also not in camel case, I am not sure how much impact it will have on existing installations if we change it to segmentFiles.

It will be :hankey: =D

pstibrany (commented):
Thanks Bartek for notification. I will review it in more detail. On first quick read I am just wondering about simplifying of the json format to

```
"files": {
    "meta.json": {},
    "index": {
        "length": 401
    },
    "chunks/000001": {
        "length": 3751
    },
}
```

Also please let's keep `"segment_files"` as-is, or renaming with backwards compatibility, so that blocks with existing segment files will still benefit from this optimization.

bwplotka:
> Thanks Bartek for notification. I will review it in more detail. On first quick read I am just wondering about simplifying of the json format to

What's the benefit? I think it's quite nice that we have it sorted (same way segments files are)

and yea, we are not renaming (:

pstibrany:
> What's the benefit? I think it's quite nice that we have it sorted (same way segments files are)

We need sorting for segment files, because of how they are used (in chunkrefs, also as replacement of Iter call).

Here we are not replacing Iter call. Map of file to attributes is therefore enough, and makes resulting JSON easier to read.

pstibrany:
Btw, maps are sorted when serialized to JSON.

pkg/block/block.go
```diff
@@ -91,6 +93,20 @@ func Upload(ctx context.Context, logger log.Logger, bkt objstore.Bucket, bdir st
...
+	if err != nil {
+		return errors.Wrap(err, "gather meta file stats")
+	}
+
+	metaEncoded := bytes.Buffer{}
+	if err := meta.Write(&metaEncoded); err != nil {
+		return errors.Wrap(err, "encode meta file")
+	}
+
 	if err := objstore.UploadFile(ctx, logger, bkt, path.Join(bdir, MetaFilename), path.Join(DebugMetas, fmt.Sprintf("%s.json", id))); err != nil {
```
pstibrany, line 106: Here we upload meta.json file which isn't updated with files.

pstibrany (commented):
> We need sorting for segment files, because of how they are used (in chunkrefs, also as replacement of Iter call).
> 
> Here we are not replacing Iter call. Map of file to attributes is therefore enough, and makes resulting JSON easier to read.

I see that segment files are also put into "files" array, which makes my argument about not needing sorting pointless.

However I still maintain that JSON would be easier to read, and also shorter, which helps caching.

pkg/block/block.go
bwplotka: Nice catch, moved metaEncoded just for this purpose but did not use it :scream:

bwplotka:
> However I still maintain that JSON would be easier to read, and also shorter, which helps caching.

I mean we can use map, and have special serialization to maintain order into slice, but does it really matter on JSON output? :thinking:  Just curious if it's worth complexity.

pstibrany (approved):
LGTM overall (assuming fixed debug meta.json). Thanks!

PR #3398, opened 2020-11-03
Add default value to query frontend memcached config

yeya24:

## Changes

This pr adds default value to query frontend Memcached configuration. For more details, please take a look at slack thread https://cloud-native.slack.com/archives/CK5RSSC10/p1604403545480000


pkg/queryfrontend/config.go
```diff
@@ -27,6 +27,21 @@ const (
 	MEMCACHED ResponseCacheProvider = "MEMCACHED"
 )
 
+var (
+	defaultMemcachedConfig = MemcachedResponseCacheConfig{
```
filippog, line 31: Since the default config is available now, should the e2e tests change as well to not duplicate the configuration?

yeya24: Hi, reusing the default config is a good idea. But I think we want this default config private, so we have to duplicate a little bit.

filippog: That's fair! SGTM

bwplotka (approved):
Thanks!

Hopefully will make it easier to use memcache config :hugs:

Issue #1424, opened 2019-08-15
compactor - "write postings: exceeding max size of 64GiB"


claytono:
**Thanos, Prometheus and Golang version used**

Prometheus version: 2.11.1

```
thanos, version 0.6.0 (branch: master, revision: 579671460611d561fcf991182b3f60e91ea04843)
  build user:       root@packages-build-1127f72e-908b-40dc-b9e1-f95dcacaaa62-sck6t
  build date:       20190806-16:13:48
  go version:       go1.12.7
```

**What happened**

When running the compactor on we get the following output (sanitized to remove internal hostnames):

```
2019-08-13T15:13:54-07:00 observability-compactor-123456 user:notice thanos-compact-observability-036[20057]: level=error ts=2019-08-13T22:13:54.054751118Z caller=main.go:199 msg="running command failed" err="compaction failed: compaction failed for group 0@{host=\"observability-prom-A98765\"}: compact blocks [thanos-compact-tDlEs/compact/0@{host=\"observability-prom-A98765\"}/01DHRV2ADD0DGW2V0WB84D472P thanos-compact-tDlEs/compact/0@{host=\"observability-prom-B98765\"}/01DHTN0B5GK4M9K1YF344C599J thanos-compact-tDlEs/compact/0@{host=\"observability-prom-B98765\"}/01DHTYSE4G3CPN6PT73JD4NFPC thanos-compact-tDlEs/compact/0@{host=\"observability-prom-B98765\"}/01DHV863B7P32AR6CM46ZJCJDR thanos-compact-tDlEs/compact/0@{host=\"observability-prom-B98765\"}/01DHY92FF48P4CPQMGPVMEVJE8 thanos-compact-tDlEs/compact/0@{host=\"observability-prom-B98765\"}/01DHYJM1E1F58STJERMF6PWY0M]: 2 errors: write compaction: write postings: write postings: exceeding max size of 64GiB; exceeding max size o
2019-08-13T15:13:54-07:00 observability-compactor-123456 user:notice thanos-compact-observability-036[20057]: f 64GiB"
```

**What you expected to happen**

Compaction would succeed, or skip these block without error, since they cannot be compacted.

**How to reproduce it (as minimally and precisely as possible)**:

Try to compact blocks that result in the `postings` part of the resulting index to exceed 64G

**Full logs to relevant components**

See log lines above.  Let me know if anything else would be useful, but I think that's all we have that is relevant

**Anything else we need to know**

[...]

PR #3409, opened 2020-11-04
compact: Added support for no-compact markers in planner.

bwplotka:
The planner algo was adapted to avoid unnecessary changes to
blocks caused by excluded blocks, so we can quickly switch to
different planning logic in the next iteration.

Fixes: https://github.com/thanos-io/thanos/issues/1424

The idea is to have `no-compact-mark.json` in each block that has to be excluded from compaction.  cc @SuperQ 


bwplotka:
Example `no-compact-mark.json`: `{"id":"00000000030000000000000000","version":1,"reason":"index-size-exceeding","details":"yolo"}`

pkg/compact/compact.go
```diff
@@ -984,3 +985,50 @@ func (c *BucketCompactor) Compact(ctx context.Context) (rerr error) {
...
+
+// NewGatherNoCompactionMarkFilter creates GatherNoCompactionMarkFilter.
+func NewGatherNoCompactionMarkFilter(logger log.Logger, bkt objstore.InstrumentedBucketReader) *GatherNoCompactionMarkFilter {
+	return &GatherNoCompactionMarkFilter{
+		logger: logger,
+		bkt:    bkt,
+	}
+}
+
+// DeletionMarkBlocks returns block ids that were marked for deletion.
```
yeya24, line 1008: `// NoCompactMarkedBlocks returns block ids that were marked for no compaction.`

pkg/compact/planner.go
```diff
@@ -7,47 +7,69 @@ import (
...
 var _ Planner = &tsdbBasedPlanner{}
 
-// NewTSDBBasedPlanner is planner with the same functionality as Prometheus' TSDB plus special handling of excluded blocks.
+// NewTSDBBasedPlanner is planner with the same functionality as Prometheus' TSDB.
 // TODO(bwplotka): Consider upstreaming this to Prometheus.
-// It's the same functionality just without accessing filesystem, and special handling of excluded blocks.
+// It's the same functionality just without accessing filesystem.
 func NewTSDBBasedPlanner(logger log.Logger, ranges []int64) *tsdbBasedPlanner {
-	return &tsdbBasedPlanner{logger: logger, ranges: ranges}
+	return &tsdbBasedPlanner{logger: logger, ranges: ranges, noCompBlocksFunc: func() map[ulid.ULID]*metadata.NoCompactMark { return make(map[ulid.ULID]*metadata.NoCompactMark) }}
```
yeya24, line 28: What about not inlining this line? It is too long.

pkg/compact/planner.go
```diff
@@ -7,47 +7,69 @@ import (
...
+	res := selectOverlappingMetas(notExcludedMetasByMinTime)
 	if len(res) > 0 {
 		return res, nil
 	}
-
 	// No overlapping blocks, do compaction the usual way.
-	// We do not include a recently created block with max(minTime), so the block which was just created from WAL.
-	// This gives users a window of a full block size to piece-wise backup new data without having to care about data overlap.
+
+	// We do not include a recently created block with max(minTime), so the block which was just uploaded to bucket.
```
yeya24, line 54: Why a block with the max `minTime` is just uploaded?  It is possible that it was updated several days ago, but has the max `minTime`

pkg/compact/planner.go
```diff
@@ -56,19 +78,17 @@ func (p *tsdbBasedPlanner) Plan(_ context.Context, metasByMinTime []*metadata.Me
 
 // selectMetas returns the dir metas that should be compacted into a single new block.
 // If only a single block range is configured, the result is always nil.
-func selectMetas(ranges []int64, metasByMinTime []*metadata.Meta) []*metadata.Meta {
+func selectMetas(ranges []int64, noCompactMarked map[ulid.ULID]*metadata.NoCompactMark, metasByMinTime []*metadata.Meta) []*metadata.Meta {
```
yeya24, line 81: Can you please have a unit test for this function? The logic is not very easy to review...

yeya24 (commented):
Only several nits. Amazing work!

pkg/compact/planner.go
bwplotka: It's copied from Prometheus ): 

Also there are full unit tests on Planner side, which should be enough. WDYT?

I agree code is not very clear. We need to redesign all long term - this is exactly same as Prometheus now

pkg/compact/planner.go
bwplotka: This is exactly what this comment says right?  That it has max of minTimes of all blocks, no?

bwplotka:
Addressed comments @yeya24 PTAL (:

pkg/compact/planner.go
yeya24: Nice

pkg/compact/planner.go
yeya24: For example, if we stop adding new blocks to a bucket, and the latest block in the bucket was uploaded 3 days ago. So we also want to exclude the latest block in this case?

The comment mentions `recently created block` so seems to check the block creation time makes more sense. Maybe I miss something

bwplotka: We mean the freshest TBH

also the whole check is really not super needed for us. It's relevant to Prometheus, but let's keep it :hugs:

yeya24: Ok. If it is relevant to Prometheus then it makes sense :+1:

yeya24 (approved):
LGTM!

PR #3411, opened 2020-11-05
e2e test for multitenancy WORK IN PROGRESS

Chans321:
Signed-off-by: Chans321 <tsschand@gmail.com>

### DIFF
--- a/test/e2e/receive_test.go
+++ b/test/e2e/receive_test.go
@@ -15,7 +15,34 @@ import (
 	"github.com/thanos-io/thanos/pkg/receive"
 	"github.com/thanos-io/thanos/pkg/testutil"
 	"github.com/thanos-io/thanos/test/e2e/e2ethanos"
+
+	"net/http"
+	"net/http/httputil"
+	"net/url"
+	"log"
 )
+
+type ReverseProxyConfig struct {
+	tenant_id string
+	port      string
+	target    string
+}
+
+func generateProxy(conf ReverseProxyConfig) {
+	originurl, _ := url.Parse(conf.target)
+	director := func(req *http.Request) {
+		req.Header.Add("tenant_id", conf.tenant_id)
+		req.Header.Add("X-Origin-Host", originurl.Host)
+		req.URL.Scheme = "http"
+		req.URL.Host = originurl.Host
+	}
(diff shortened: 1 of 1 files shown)


test/e2e/receive_test.go
```diff
@@ -240,4 +267,55 @@ func TestReceive(t *testing.T) {
...
+		conf1 := ReverseProxyConfig{
+			tenant_id: "tenant-1",
+			port:      ":9097",
+			target:    e2ethanos.RemoteWriteEndpoint(r1.NetworkEndpoint(8080)),
+		}
+		q, err := e2ethanos.NewQuerier(s.SharedDir(), "1", []string{r1.GRPCNetworkEndpoint()}, nil, nil, "", "")
+		testutil.Ok(t, err)
+		testutil.Ok(t, s.StartAndWaitReady(q))
+		go generateProxy(conf1)
+		prom1, _, err := e2ethanos.NewPrometheus(s.SharedDir(), "1", defaultPromConfig("prom1", 0, "http://localhost:9097/api/v1/receive", ""), e2ethanos.DefaultPrometheusImage())
```
brancz, line 303: I think what's happening is that because the Prometheus is started in its own docker container, "localhost:9097" is actually not the same as the reverse proxy started by this test.

Maybe @bwplotka or @pracucci have a suggestion how to do something like this best with the e2e test framework?

bwplotka: As in other tests, you can use `r1.HTTPEndpoint()` to get actual address. (:

Chans321: Ok. I will change it to this. But the error which im facing is "No storeAPI matched for this query".

bwplotka: Ok, I think I see the problem. We would need to inject ourselves into the docker network to inject  proxy server being run in unit test. I think it's doable, but it might simpler to create a container for this proxy first :hugs: Just creating a small service with container having such code like here would do there work: https://github.com/thanos-io/thanos/blob/17343c16b820a6fd5444e4258f4264ed1de065cc/test/e2e/e2ethanos/services.go#L117

brancz:
I think we can close this since https://github.com/thanos-io/thanos/pull/3474 is merged no?

PR #3414, opened 2020-11-05
Query Frontend: Set CORS header

hwoarang:
Set CORS to the query frontend component in order to be able
to freely access it from Grafana's living elsewhere.

Signed-off-by: Markos Chandras <markos@chandras.me>

## Changes

Set `Access-Control-Allow-Origin` Header to Query Frontend

## Verification

- `make test-e2e-local`
- Tested also with local grafana connected to a local Thanos frontend using our internal Thanos querier and everything worked fine. No CORS errors anymore.

### DIFF
--- a/cmd/thanos/query_frontend.go
+++ b/cmd/thanos/query_frontend.go
@@ -19,6 +19,7 @@ import (
 	"github.com/prometheus/client_golang/prometheus"
 	"github.com/weaveworks/common/user"
 
+	"github.com/thanos-io/thanos/pkg/api"
 	"github.com/thanos-io/thanos/pkg/component"
 	"github.com/thanos-io/thanos/pkg/extflag"
 	"github.com/thanos-io/thanos/pkg/extkingpin"
@@ -199,6 +200,7 @@ func runQueryFrontend(
 		instr := func(f http.HandlerFunc) http.HandlerFunc {
 			hf := http.HandlerFunc(func(w http.ResponseWriter, r *http.Request) {
 				name := "query-frontend"
+				api.SetCORS(w)
 				ins.NewHandler(
 					name,
 					logMiddleware.HTTPMiddleware(
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -28,6 +28,7 @@ We use *breaking :warning:* to mark changes that are not backward compatible (re
 - [#3257](https://github.com/thanos-io/thanos/pull/3257) Ruler: Prevent Ruler from crashing when using default DNS to lookup hosts that results in "No such hosts" errors.
 - [#3331](https://github.com/thanos-io/thanos/pull/3331) Disable Azure blob exception logging
 - [#3341](https://github.com/thanos-io/thanos/pull/3341) Disable Azure blob syslog exception logging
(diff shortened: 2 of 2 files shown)


GiedriusS:
Does this also work when you use POST? Because we have:

```go
var corsHeaders = map[string]string{
	"Access-Control-Allow-Headers":  "Accept, Accept-Encoding, Authorization, Content-Type, Origin",
	"Access-Control-Allow-Methods":  "GET, OPTIONS",
	"Access-Control-Allow-Origin":   "*",
	"Access-Control-Expose-Headers": "Date",
}
```

hwoarang:
> Does this also work when you use POST? Because we have:
> 
> ```go
> var corsHeaders = map[string]string{
> 	"Access-Control-Allow-Headers":  "Accept, Accept-Encoding, Authorization, Content-Type, Origin",
> 	"Access-Control-Allow-Methods":  "GET, OPTIONS",
> 	"Access-Control-Allow-Origin":   "*",
> 	"Access-Control-Expose-Headers": "Date",
> }
> ```

@GiedriusS that's an interesting question. I wonder how it works right now with Querier. I am looking into this.

hwoarang:
Even now (thanos 0.16), POST is not allowed from what I see

```
~$ curl -X POST --head https://thanos.XXXX
HTTP/1.1 405 Method Not Allowed
Allow: GET, OPTIONS
Content-Type: text/plain; charset=utf-8
Date: Thu, 05 Nov 2020 16:43:36 GMT
Vary: Accept-Encoding
Vary: Accept-Encoding
X-Content-Type-Options: nosniff
Content-Length: 19
Connection: keep-alive
```
:thinking:

hwoarang:
Actually I should have used `Add()` to only append the extra header to the frontend client instead of resetting everything. Now it looks OK.

hwoarang:
`make test-e2e-local` fails randomly but I do not thin it's related to this PR. I believe the test-e2e test failure here is similar.

cmd/thanos/query_frontend.go
```diff
@@ -199,6 +199,8 @@ func runQueryFrontend(
 		instr := func(f http.HandlerFunc) http.HandlerFunc {
 			hf := http.HandlerFunc(func(w http.ResponseWriter, r *http.Request) {
 				name := "query-frontend"
+				// Set CORS to frontend client
+				w.Header().Set("Access-Control-Allow-Origin", "*")
```
bwplotka, line 203: Can we do `api.SetCORS` instead? (:  And no need for comment :hugs: 

Thanks!

bwplotka: Ref: https://github.com/bwplotka/thanos/blob/987b74670cd5ab8bab15cd35fccc4c85dd36ce9e/pkg/api/api.go#L118

bwplotka (commented):
One suggestion and LGTM (:

cmd/thanos/query_frontend.go
hwoarang: @bwplotka I initially did it like you said but I thought that maybe I shouldn't set  more that what I actually needed. Let me change that again.

bwplotka (approved):
LGTM, thanks!

PR #3419, opened 2020-11-05
Fix empty label names response

yeya24:

## Changes

To be consistent with Prometheus API. If no label names found, return an empty array instead of nothing.


bwplotka (approved):
LGTM, thanks :+1:

PR #3427, opened 2020-11-09
'Proxy bases' or 'proxy based'?

kiefersmith:
Line 23 - I think this is meant to be 'proxy based'

## Changes

Simple spelling correction.

## Verification

I didn't test it, but it probably works.


bwplotka (approved):
Yes! Can you fix DCO check? You need to sign commit. 🤗

Thanks!

PR #3424, opened 2020-11-09
pkg/logging: Fix logging caller

brancz:
* [ ] I added CHANGELOG entry for this change.
* [x] Change is not relevant to the end user.

## Changes

Fix logging caller tag.

## Verification

Running the e2e tests to see.

@thanos-io/thanos-maintainers

PR #3428, opened 2020-11-10
*: Cleanup and simplify logging instantiation

brancz:
Signed-off-by: Frederic Branczyk <fbranczyk@gmail.com>

## Changes

After https://github.com/thanos-io/thanos/pull/3424, and re-reading all the surrounding code, I feel like this makes more sense, and we don't have to fiddle with the log caller if it were to ever change in go-kit, since we can use the default correctly now (again).

@thanos-io/thanos-maintainers

### DIFF
--- a/pkg/logging/logger.go
+++ b/pkg/logging/logger.go
@@ -4,7 +4,6 @@
 package logging
 
 import (
-	"fmt"
 	"os"
 
 	"github.com/go-kit/kit/log"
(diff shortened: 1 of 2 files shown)


pkg/logging/logger.go
```diff
@@ -16,32 +15,12 @@ const (
 	LogFormatJSON   = "json"
 )
 
-// NewLogger returns a log.Logger that prints in the provided format with a UTC
```
kakkoyun, line 19: nit: We should have the comment for `godoc` since it's an exposed function.

kakkoyun (commented):
LGTM 🥇 Simpler is better :)

Only one comment about the godoc comments.

pkg/logging/logger.go
brancz: My bad, not sure what made me delete this. Added it back and updated it with the current state.

brancz:
I assume the netlify failures can safely be ignored.

PR #3429, opened 2020-11-10
[RFC/WIP] receive: do not add tenant label if its value is DefaultTenant

GiedriusS:
We have one peculiar use-case for Thanos Receive: we send off
some metrics from a Prometheus instance to a Thanos Receive instance so
that it would be persisted for a longer time. We achieve that by adding
some special external labels that are later picked up by Thanos
Compactor with the appropriate retention policies.

Also, we use the `tenant` label already and have no intention of using
Thanos Receive in a multi-tenant mode. We'd like to avoid running a
Prometheus/Sidecar pair on top of another one that is running in
federation. That simplifies the deployment and reduces the resource
usage as well, most likely.

Since the multiTSDB functionality got introduced into Thanos Receive
from 0.13.x, I believe, it started adding a mandatory label `tenant`
with the tenant's ID that clashes with our already-used `tenant` label.

In this PR, I propose _not_ adding a `tenant` label if it is equal to
`default-tenant`. Since the tenant's ID is deeply entrenched in the code
base now, I think this might be reasonable trade-off to make.

Other options include adding another flag which controls what label
_name_ to use. And, if it would be `""` then we could not add that at
all. Or maybe someone else has an even better idea? Let me know.

## Verification

`make test-e2e-local` but fixing the tests is TBD.

### DIFF
--- a/pkg/receive/multitsdb.go
+++ b/pkg/receive/multitsdb.go
@@ -255,8 +255,11 @@ func (t *MultiTSDB) TSDBStores() map[string]storepb.StoreServer {
 }
 
 func (t *MultiTSDB) startTSDB(logger log.Logger, tenantID string, tenant *tenant) error {
+	lbls := t.labels
 	reg := prometheus.WrapRegistererWith(prometheus.Labels{"tenant": tenantID}, t.reg)
-	lbls := append(t.labels, labels.Label{Name: t.tenantLabelName, Value: tenantID})
+	if tenantID != DefaultTenant {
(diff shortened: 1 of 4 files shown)


GiedriusS:
Actually, I worked around this issue by switching from `--label="mymagiclabel=foo"` to:
```
--receive.tenant-label-name=mymagiclabel --receive.default-tenant-id="foo"
```

And, as a result, the `tenant` label does not appear, and I still have the label that I want. I guess we can close this until a problem occurs again.

PR #3431, opened 2020-11-11
Store-gateway: add support to lazy mmap index-headers

pracucci:

## Changes
As discussed during the Thanos community meeting, I would like to propose to add support for lazy mmaping of index-headers in the store-gateway with the ability to automatically unload them after X time of inactivity. This PR aims to do it.

**Draft** because:
- CHANGELOG is missing

## Verification

Unit tests.

### DIFF
--- a/cmd/thanos/store.go
+++ b/cmd/thanos/store.go
@@ -107,8 +107,8 @@ func registerStore(app *extkingpin.App) {
 		"Default is 24h, half of the default value for --delete-delay on compactor.").
 		Default("24h"))
 
-	lazyIndexReaderEnabled := cmd.Flag("store.enable-index-header-lazy-reader", "If true, Store Gateway will lazy memory map index-header only once required by a query.").
-		Hidden().Default("false").Bool()
+	lazyIndexReaderEnabled := cmd.Flag("store.enable-index-header-lazy-reader", "If true, Store Gateway will lazy memory map index-header only once the block is required by a query.").
+		Default("false").Bool()
 
 	lazyIndexReaderIdleTimeout := cmd.Flag("store.index-header-lazy-reader-idle-timeout", "If index-header lazy reader is enabled and this idle timeout setting is > 0, memory map-ed index-headers will be automatically released after 'idle timeout' inactivity.").
 		Hidden().Default("5m").Duration()
--- a/cmd/thanos/store.go
+++ b/cmd/thanos/store.go
@@ -107,6 +107,12 @@ func registerStore(app *extkingpin.App) {
 		"Default is 24h, half of the default value for --delete-delay on compactor.").
 		Default("24h"))
 
+	lazyIndexReaderEnabled := cmd.Flag("store.enable-index-header-lazy-loading", "If true, Store Gateway will lazy memory map index-header only once required by a query.").
+		Hidden().Default("false").Bool()
+
+	lazyIndexReaderIdleTimeout := cmd.Flag("store.lazy-index-header-idle-timeout", "If index-header lazy loading is enabled and this idle timeout setting is > 0, memory map-ed index-headers will be automatically released after 'idle timeout' inactivity.").
+		Hidden().Default("5m").Duration()
+
 	webExternalPrefix := cmd.Flag("web.external-prefix", "Static prefix for all HTML links and redirect URLs in the bucket web UI interface. Actual endpoints are still served on / or the web.route-prefix. This allows thanos bucket web UI to be served behind a reverse proxy that strips a URL sub-path.").Default("").String()
 	webPrefixHeaderName := cmd.Flag("web.prefix-header", "Name of HTTP request header used for dynamic prefixing of UI links and redirects. This option is ignored if web.external-prefix argument is set. Security risk: enable this option only if a reverse proxy in front of thanos is resetting the header. The --web.prefix-header=X-Forwarded-Prefix option can be useful, for example, if Thanos UI is served via Traefik reverse proxy with PathPrefixStrip option enabled, which sends the stripped prefix value in X-Forwarded-Prefix header. This allows thanos UI to be served on a sub-path.").Default("").String()
 
@@ -152,6 +158,8 @@ func registerStore(app *extkingpin.App) {
 			*postingOffsetsInMemSampling,
 			cachingBucketConfig,
 			getFlagsMap(cmd.Flags()),
+			*lazyIndexReaderEnabled,
+			*lazyIndexReaderIdleTimeout,
 		)
 	})
 }
@@ -184,6 +192,8 @@ func runStore(
 	postingOffsetsInMemSampling int,
 	cachingBucketConfig *extflag.PathOrContent,
 	flagsMap map[string]string,
+	lazyIndexReaderEnabled bool,
+	lazyIndexReaderIdleTimeout time.Duration,
 ) error {
 	grpcProbe := prober.NewGRPC()
 	httpProbe := prober.NewHTTP()
@@ -304,6 +314,8 @@ func runStore(
 		enablePostingsCompression,
 		postingOffsetsInMemSampling,
 		false,
+		lazyIndexReaderEnabled,
+		lazyIndexReaderIdleTimeout,
 	)
 	if err != nil {
 		return errors.Wrap(err, "create object storage store")
--- a/cmd/thanos/store.go
+++ b/cmd/thanos/store.go
@@ -107,10 +107,10 @@ func registerStore(app *extkingpin.App) {
 		"Default is 24h, half of the default value for --delete-delay on compactor.").
 		Default("24h"))
 
-	lazyIndexReaderEnabled := cmd.Flag("store.enable-index-header-lazy-loading", "If true, Store Gateway will lazy memory map index-header only once required by a query.").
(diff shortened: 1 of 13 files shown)


pracucci:
OK @bwplotka. I've manually tested it and, so far, I haven't noticed any issue. I think it's ready for a review.

pracucci:
We've tested it on a Cortex cluster where each store-gateway has >65K blocks and we can finally run it without hitting in the max mmap areas kernel limit. It's still a bit slow to startup, but that's another story.

Screenshot 2020-11-12 at 14 35 14

pracucci:
Indexes loaded (green) and indexes unloaded (yellow) with a 15m idle timeout.

Screenshot 2020-11-12 at 14 37 12

cmd/thanos/store.go
```diff
@@ -107,6 +107,12 @@ func registerStore(app *extkingpin.App) {
 		"Default is 24h, half of the default value for --delete-delay on compactor.").
 		Default("24h"))
 
+	lazyIndexReaderEnabled := cmd.Flag("store.enable-index-header-lazy-reader", "If true, Store Gateway will lazy memory map index-header only once required by a query.").
```
bwplotka, line 110: I think it should be default true at some point, so what about unhidding this flag?

```suggestion
	lazyIndexReaderEnabled := cmd.Flag("store.enable-index-header-lazy-reader", "If true, Store Gateway will lazy memory map index-header only once the block is required by a query.").
```

cmd/thanos/store.go
```diff
@@ -107,6 +107,12 @@ func registerStore(app *extkingpin.App) {
...
+		Hidden().Default("false").Bool()
+
+	lazyIndexReaderIdleTimeout := cmd.Flag("store.index-header-lazy-reader-idle-timeout", "If index-header lazy reader is enabled and this idle timeout setting is > 0, memory map-ed index-headers will be automatically released after 'idle timeout' inactivity.").
```
bwplotka, line 113: This can stay hidden, maybe not sure if needed to be tweaked (less flags = simler to use thing is )

pkg/block/indexheader/lazy_binary_reader.go
```diff
@@ -0,0 +1,228 @@
...
+
+type LazyBinaryReaderMetrics struct {
+	loadCount         prometheus.Counter
+	loadFailedCount   prometheus.Counter
+	unloadCount       prometheus.Counter
+	unloadFailedCount prometheus.Counter
+	loadDuration      prometheus.Histogram
+}
+
+func NewLazyBinaryReaderMetrics(reg prometheus.Registerer) *LazyBinaryReaderMetrics {
```
bwplotka, line 33: Technically some one sentence comment would be nice to have (:

pkg/block/indexheader/lazy_binary_reader.go
```diff
@@ -0,0 +1,228 @@
...
+	return r.reader.LabelNames()
+}
+
+// open ensure the underlying binary index-header reader has been successfully opened. Returns
+// an error on failure. This function MUST be called with the read lock already acquired.
+func (r *LazyBinaryReader) open() error {
+	// Nothing to do if we already tried opening it.
+	if r.reader != nil {
+		return nil
+	} else if r.readerErr != nil {
```
bwplotka, line 193: no need for `else` (:

bwplotka: https://thanos.io/tip/contributing/coding-style-guide.md/#control-structure-prefer-early-returns-and-avoid-else

pkg/block/indexheader/lazy_binary_reader.go
```diff
@@ -0,0 +1,228 @@
...
+	// the read lock once done.
+	r.readerMx.RUnlock()
+	r.readerMx.Lock()
+	defer r.readerMx.RLock()
+	defer r.readerMx.Unlock()
+
+	// Ensure none else tried to open it in the meanwhile.
+	if r.reader != nil {
+		return nil
+	} else if r.readerErr != nil {
```
bwplotka, line 207: ditto

pkg/block/indexheader/lazy_binary_reader.go
```diff
@@ -0,0 +1,228 @@
...
+		return r.readerErr
+	}
+
+	level.Debug(r.logger).Log("msg", "lazy loading index-header file", "path", r.filepath)
+	r.metrics.loadCount.Inc()
+	startTime := time.Now()
+
+	reader, err := NewBinaryReader(r.ctx, r.logger, r.bkt, r.dir, r.id, r.postingOffsetsInMemSampling)
+	if err != nil {
+		level.Error(r.logger).Log("msg", "failed to lazy load index-header file", "path", r.filepath, "err", err)
```
bwplotka, line 217: Do we need this log line? It looks to me like we handle this error 3 times - metric, log line AND returning (so probably will be logged anyway), no?

pkg/block/indexheader/lazy_binary_reader.go
```diff
@@ -0,0 +1,228 @@
...
+		r.metrics.loadFailedCount.Inc()
+		r.readerErr = err
+		return err
+	}
+
+	r.reader = reader
+	level.Info(r.logger).Log("msg", "lazy loaded index-header file", "path", r.filepath, "elapsed", time.Since(startTime))
```
bwplotka, line 224: Do we need this log line :thinking: Maybe it should be debug instead?

bwplotka: we have nice metrics for it

pkg/block/indexheader/reader_pool.go
```diff
@@ -0,0 +1,187 @@
...
+	"go.uber.org/atomic"
+
+	"github.com/thanos-io/thanos/pkg/objstore"
+)
+
+type ReaderPool struct {
+	lazyReaderEnabled     bool
+	lazyReaderIdleTimeout time.Duration
+	lazyReaderMetrics     *LazyBinaryReaderMetrics
+	logger                log.Logger
```
bwplotka, line 25: Wonder if it wouldn't be easier to track open / close metrics here instead of lazy BinaryReader? :thinking: Also we might be missing some gauge for currently loaded headers?

pkg/block/indexheader/reader_pool.go
```diff
@@ -0,0 +1,187 @@
...
+	p.readersMx.Lock()
+	defer p.readersMx.Unlock()
+
+	// When this function is called, it means the reader has been closed NOT because was idle
+	// but because the consumer closed it. By contract, a reader closed by the consumer can't
+	// be used anymore, so we can automatically remove it from the pool.
+	delete(p.readers, r)
+}
+
+type readerTracker struct {
```
bwplotka, line 153: Hm, I think `LazyBinaryReader` could do updating time as well, no? (we touch `open` function always).

Also to be nicer we could pass to `LazyBinaryReader` a `func onReaderClosed(*LazyBinaryReader) to not pass the whole `*ReaderPool` WDYT?

bwplotka (approved):
Amazing, generally LGTM, just style/readability things to improve :muscle: 

Thanks!

cmd/thanos/store.go
pracucci: Ok, unhiding it. I wasn't just that optimistic that this could be the default at some point (still need to test performances  on opening a very large block index).

cmd/thanos/store.go
pracucci: I would suggest to keep the hidden flag, to allow all of us to experiment with that and find a better idle timeout. I just put a "random" value so far.

pkg/block/indexheader/lazy_binary_reader.go
pracucci: You're right.

pkg/block/indexheader/lazy_binary_reader.go
pracucci: Moved to debug.

pkg/block/indexheader/reader_pool.go
pracucci: > Wonder if it wouldn't be easier to track open / close metrics here instead of lazy BinaryReader?

That was my initial choice, but unfortunately it's more complicated because the pool doesn't have visibility when the lazy load (`open()`) actually occur. Tracking it in `LazyBinaryReader` turned out to be easier and safer.

pracucci: > Also we might be missing some gauge for currently loaded headers?

Technically, it's `indexheader_lazy_load_total - indexheader_lazy_unload_total`. Do we also need a gauge?

pkg/block/indexheader/reader_pool.go
pracucci: Ok. I removed `readerTracker` as suggested.

pracucci:
@bwplotka Thanks for the review! I've address your comments or commented otherwise. Could you take another look, please?

pkg/block/indexheader/reader_pool.go
bwplotka: > Technically, it's indexheader_lazy_load_total - indexheader_lazy_unload_total. Do we also need a gauge?

As long as you subtract metrics from exactly same process, otherwise: `rate(indexheader_lazy_load_total[xx]) - rate(indexheader_lazy_unload_total[xx]` which is getting funky

Anyway, good for now :+1:

bwplotka (approved):
LGTM! :muscle:

PR #3432, opened 2020-11-11
Upgrade Cortex into Thanos

pracucci:

## Changes

We recently refactored the query-frontend in Cortex, in order to introduce the support for a new component (query-scheduler used by query-frontend v2). The "old" query-frontend logic (v1) has been left untouched, a part from some package refactoring. Also the v1 doesn't handle the downstream, slow query logging and response compression anymore, but they're available as pluggable middlewares.

In this PR I've upgraded Cortex and done the required changes 🤞 

_Note: this upgrade has the nice side effect that Thanos will not vendor anymore Cortex `pkg/storegateway`, thus unblocking the PR #3431._

## Verification

Existing tests.

### DIFF
--- a/cmd/thanos/query_frontend.go
+++ b/cmd/thanos/query_frontend.go
@@ -45,7 +45,6 @@ func registerQueryFrontend(app *extkingpin.App) {
 	cmd := app.Command(comp.String(), "query frontend")
 	cfg := &queryFrontendConfig{
 		Config: queryfrontend.Config{
-			CortexFrontendConfig: &cortexfrontendv1.Config{},
 			// Max body size is 10 MiB.
 			CortexHandlerConfig: &transport.HandlerConfig{
 				MaxBodySize: 10 * 1024 * 1024,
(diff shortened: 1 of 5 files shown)


yeya24 (approved):
LGTM

yeya24:
There is a new supported config in the limits middleware https://github.com/cortexproject/cortex/blob/master/pkg/querier/queryrange/limits.go#L61. 

I think it is useful and we should support it. I will open an issue to track this.

bwplotka (approved):
Amazing, Thanks for the work :muscle:

pracucci:
> There is a new supported config in the limits middleware https://github.com/cortexproject/cortex/blob/master/pkg/querier/queryrange/limits.go#L61.
> 
> I think it is useful and we should support it. I will open a feature request to track this.

@yeya24 Sure. Just pay attention this applies only to range queries. It should work in conjunction with the same limit applied in the querier, and not just in the query-frontend.

Issue #2984, opened 2020-08-05
StoreGateway not picking external labels in querier metadata API endpoint - `/api/v1/label/<label_name>/values`


PavanVasist:
**Components it touches - querier, storegateway, sidecar.**

**Thanos, Prometheus and Golang version used**:

**Thanos**
Version | 0.13.0
-- | --
adf6facb8d6bf44097aae084ec091ac3febd9eb8
HEAD
circleci@b8cd18f8b553
20200622-10:04:50
go1.14.2

**Prometheus **
Version | 2.14.0
-- | --
edeb7a44cbf745f1d8be4ea6f215e79e651bfe19
HEAD
root@df2327081015
20191111-14:27:12
go1.13.4

Thanos - quay.io/thanos/thanos:v0.13.0; Prometheus  - quay.io/prometheus/prometheus:v2.14.0

**Object Storage Provider**:
AWS S3 Bucket

**What happened**:
When external_labels are added to scrape config and tried to filter the label values via metadata API within querier by fetching TSDB from gateway connected to s3 bucket, no values are returned. But when the querier is set directly set to fetch from sidecar, the external_labels are fetched correctly. Some more info - 
https://github.com/grafana/grafana/issues/26764 
https://groups.google.com/forum/#!msg/prometheus-users/c6kaSHiP7Cc/fILFnptyAQAJ

**What you expected to happen**:
The label_values for external_labels should be filtered automatically in the querier API metadata endpoint while fetching it from stoegateway.

**How to reproduce it (as minimally and precisely as possible)**:

**Full logs to relevant components**:

**Anything else we need to know**:

**Environment**:
- OS (e.g. from /etc/os-release): Linux/K8s Platform

PavanVasist:
Any help or work arounds?

bwplotka:
Hey, sorry for delay. What's the use case? And I think it makes total sense to uniform Label values to work similarly between StoreAPI implementations.

PavanVasist:
Thank you for dropping by @bwplotka. 
The idea is - 

- We have a prometheus and thanos-sidecar running in a cluster pushing metrics to AWS S3 bucket.
- In an another cluster am using thanos-store connected to same S3, with compactor and querier(used as datasource to Grafana).

- Now the prometheus running cluster is short-lived as we run some tests and want to capture the metrics every-time the tests run and later we shutdown the cluster. For every run we add different external_labels to perform a comparison of results across these multiple runs in 2nd cluster where store along with querier is run.

[...]

PR #3437, opened 2020-11-12
[store] Add response hints to label APIs

gouthamve:

## Changes

In Cortex, we shard blocks b/w store-gateways, hence we need to send the list of blockids to query and the list of ids that have been queried in the response to make sure all the relevant blocks are queried. This has already been done for the `Series` API.

I have not yet added request hints yet. I will add them in a subsequent PR.

## Verification

tbh, I have not tested it yet. I am only relying on the unit tests. Any tips on how I can test it?


pkg/store/bucket_test.go
```diff
@@ -1897,3 +1897,190 @@ func createBlockWithOneSeriesWithStep(t testutil.TB, dir string, lbls labels.Lab
...
+				},
+			},
+		},
+	}
+
+	for _, tc := range testCases {
+		t.Run(tc.name, func(t *testing.T) {
+			namesResp, err := store.LabelNames(context.Background(), tc.labelNamesReq)
+			testutil.Ok(t, err)
+			testutil.Equals(t, tc.expectedNames, namesResp.Names)
```
gouthamve, line 2055: This fails with:
```
        bucket_test.go:2055: bucket_test.go:2055:
            
                exp: []string{"ext1", "foo", "i"}
            
                got: []string{"foo", "i"}
            
            Diff:
            --- Expected
            +++ Actual
            @@ -1,3 +1,2 @@
            -([]string) (len=3) {
            - (string) (len=4) "ext1",
            +([]string) (len=2) {
              (string) (len=3) "foo",
```


As in, the `extLabels` we passed are not being returned. From what I know, they should be, but I am not even sure what they are. /cc @pracucci @pstibrany

pracucci (approved):
Thanks @gouthamve! LGTM. Could you take a look at the comments, fix tests and rebase `master`, please?

pkg/store/bucket.go
```diff
@@ -1070,13 +1072,27 @@ func (s *BucketStore) LabelNames(ctx context.Context, req *storepb.LabelNamesReq
...
 		indexr := b.indexReader(gctx)
 		g.Go(func() error {
 			defer runutil.CloseWithLogOnErr(s.logger, indexr, "label names")
 
 			// Do it via index reader to have pending reader registered correctly.
 			res := indexr.block.indexHeaderReader.LabelNames()
+
+			// Add  a set for the external labels as well.
+			// We're not adding them directly to res because there could be duplicates.
+			extRes := make([]string, 0, len(b.meta.Thanos.Labels))
```
pracucci, line 1087: You can't safely access `b` inside the goroutine.

pkg/store/bucket.go
```diff
@@ -1119,6 +1150,11 @@ func (s *BucketStore) LabelValues(ctx context.Context, req *storepb.LabelValuesR
 				return errors.Wrap(err, "index header label values")
 			}
 
+			// Add the external label value as well.
+			if extLabelValue, ok := b.meta.Thanos.Labels[req.Label]; ok {
```
pracucci, line 1154: You can't safely access `b` inside the goroutine.

gouthamve:
Fixes https://github.com/thanos-io/thanos/issues/2984

pracucci (approved):
LGTM, thanks!

pkg/store/bucket.go
```diff
@@ -1102,13 +1119,25 @@ func (s *BucketStore) LabelNames(ctx context.Context, req *storepb.LabelNamesReq
...
 		return nil, status.Error(codes.Internal, err.Error())
 	}
+
+	var anyHints *types.Any
+	var err error
+
+	if anyHints, err = types.MarshalAny(resHints); err != nil {
+		err = status.Error(codes.Unknown, errors.Wrap(err, "marshal label names response hints").Error())
+		return nil, err
+	}
```
bwplotka, line 1129: Why declaring `anyHints` and err in separate lines? Why not:

```suggestion
	anyHints, err := types.MarshalAny(resHints); 
	if err != nil {
		return, status.Error(codes.Unknown, errors.Wrap(err, "marshal label names response hints").Error())
	}
```

pkg/store/bucket.go
```diff
@@ -1143,8 +1182,18 @@ func (s *BucketStore) LabelValues(ctx context.Context, req *storepb.LabelValuesR
 	if err := g.Wait(); err != nil {
 		return nil, status.Error(codes.Aborted, err.Error())
 	}
+
+	var anyHints *types.Any
+	var err error
+
+	if anyHints, err = types.MarshalAny(resHints); err != nil {
```
bwplotka, line 1189: ditto

bwplotka (approved):
Nice, work! Minor style nit only, otherwise LGTM :+1:

gouthamve:
@bwplotka Nice catch, not sure what I was thinking there. Fixed it!

bwplotka (approved):
LGTM, we can ignore Markdown check link, it's flaky, otherwise, we can merge on green

PR #3367, opened 2020-10-27
Check bucket series time range when decoding chunks from index

yeya24:
Signed-off-by: Ben Ye <yb532204897@gmail.com>

* [ ] I added CHANGELOG entry for this change.
* [ ] Change is not relevant to the end user.

## Changes

This idea behind this pr is to check the chunk time range when decoding the series from the index. This change can benefit a lot the `skipChunk` queries because it can immediately return if it finds a valid chunk instead of decoding all the chunks in this series. 

Verification

Uses the benchmarks from https://github.com/thanos-io/thanos/pull/3366. Benchmark and compare between this pr and current master.

1. `SkipChunk` query. 

It is expected that this type of query should be much faster than before.

```
➜  thanos git:(improve-index-valid) ✗ benchcmp old new9
benchcmp is deprecated in favor of benchstat: https://pkg.go.dev/golang.org/x/perf/cmd/benchstat
benchmark                                                                           old ns/op      new ns/op      delta
BenchmarkBucketSeriesSkipChunks/1000000SeriesWith1Samples/1of1000000-8              374105769      304828968      -18.52%
BenchmarkBucketSeriesSkipChunks/1000000SeriesWith1Samples/10of1000000-8             366452232      287316200      -21.60%
BenchmarkBucketSeriesSkipChunks/1000000SeriesWith1Samples/1000000of1000000-8        1467578057     1077410930     -26.59%
BenchmarkBucketSeriesSkipChunks/100000SeriesWith100Samples/1of10000000-8            31472869       22883196       -27.29%
BenchmarkBucketSeriesSkipChunks/100000SeriesWith100Samples/100of10000000-8          31160010       23195988       -25.56%
BenchmarkBucketSeriesSkipChunks/100000SeriesWith100Samples/10000000of10000000-8     134128144      108503863      -19.10%
BenchmarkBucketSeriesSkipChunks/1SeriesWith10000000Samples/1of10000000-8            2564966        99784          -96.11%
BenchmarkBucketSeriesSkipChunks/1SeriesWith10000000Samples/100of10000000-8          2799881        125497         -95.52%
BenchmarkBucketSeriesSkipChunks/1SeriesWith10000000Samples/10000000of10000000-8     29997992       373188         -98.76%
```

2. Normal series query

The expected performance of this should be almost the same or a little bit slower than before. However, this pr is about 10% - 20% slower when there are a lot of samples in one series. Not sure how to improve this case. I think there is no extra overhead except we are using more if-else statements than before when decoding the index.

```
[...]

PR #3440, opened 2020-11-12
Fix partial query results

pracucci:
* [x] Change is not relevant to the end user (because not released yet)

## Changes

The PR #3367 introduced a bug causing partial query results. The problem is: in `decodeSeriesWithReq()`, when the condition `if maxt < req.MinTime` is met, the chunk reference `ref0` is not read from the index (and `t0` not updated) and this breaks the decoding of subsequent chunks.

## Verification

Unit and manual tests.


yeya24 (approved):
Thank you for the fix!

bwplotka (approved):
Thank you for spotting this :scream: 

cc @yeya24

PR #3444, opened 2020-11-13
query-frontend: make POST-request to downstream url for labels and series api endpoints

2nick:
Also probably fixes series endpoint proxying as ThanosSeriesResponse.Data used LabelSet which has no UnmarshalJSON (simply changing to ZLabelSet).


bwplotka (approved):
Thanks :+1: 

LGTM

PR #3455, opened 2020-11-17
Add max-query-lookback flag to limit the query start time

yeya24:

## Changes

Upstream Cortex supports `max-query-lookback` to limit the query start time. I think it is useful to make this a configurable flag.

For how it works, please take a look at https://github.com/cortexproject/cortex/blob/master/pkg/querier/queryrange/limits.go#L61-L85

### DIFF
--- a/cmd/thanos/query_frontend.go
+++ b/cmd/thanos/query_frontend.go
@@ -79,6 +79,9 @@ func registerQueryFrontend(app *extkingpin.App) {
 	cmd.Flag("query-range.max-query-parallelism", "Maximum number of query range requests will be scheduled in parallel by the Frontend.").
 		Default("14").IntVar(&cfg.QueryRangeConfig.Limits.MaxQueryParallelism)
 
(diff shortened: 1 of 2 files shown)


GiedriusS (approved):
This seems like a useful feature indeed. Thank you for your work on this!

GiedriusS:
Do we want to add this to CHANGELOG? (:

yeya24:
> Do we want to add this to the CHANGELOG? (:

Sure I will wait for the new release merged to master and then add the changelog.

bwplotka:
Curious, why Querier does not have such flag? I don't get why we have to have query-frontend for this. We already have this kind of flag on sidecar, store GW etc

GiedriusS:
> Curious, why Querier does not have such flag? I don't get why we have to have query-frontend for this. We already have this kind of flag on sidecar, store GW etc

I don't know. We could add similar functionality there too. But these flags in query-frontend and similar flags in query are orthogonal features IMHO. One could want Thanos Query to have "all" of the metrics but only permit users to query _some_ of them via query-frontend.

PR #3458, opened 2020-11-18
Linting changes for changelog

yashrsharma44:
I made some changes for the changelog regarding the linting of it. I think, we don't have any lint checks for markdown, so this might be a good idea to add one.


bwplotka (approved):
Thanks.

What linter you used? (: That's the plan to have one, yes at some point.

Ideally in https://github.com/bwplotka/mdox

PR #3473, opened 2020-11-19
Upgraded Prometheus

pracucci:

## Changes

Today I've introduced a new option to TSDB to allow to configure the head chunks write buffer size. It's something we would need in Cortex. In order to upgrade Prometheus in Cortex, we would need to upgrade it in Thanos first, because of the breaking change to the `NewHead()` signature.

## Verification

Existing tests.


pkg/store/tsdb.go
```diff
@@ -24,6 +23,8 @@ import (
 	"github.com/thanos-io/thanos/pkg/store/storepb"
 )
 
+const RemoteReadFrameLimit = 1048576
```
pracucci, line 26: Moved here from `testutil` package to remove the `testutil` dependency from `pkg/store` (excluding tests), which is the reason why -- this time -- we can't upgrade Prometheus in Cortex without upgrading it in Thanos first.

bwplotka: :+1:

bwplotka (approved):
LGTM, thanks!

PR #3476, opened 2020-11-19
Use bytes.Buffer to detect file size when uploading files

### LINKED ISSUE #3470: compact: warn "unsupported type of io.Reader"
```
thanos, version 0.17.0 (branch: HEAD, revision: 4899d9a3378c7dbd1f635f5271624cda8caec001)
  build user:       root@432c5bd5f94c
  build date:       20201118-17:38:06
  go version:       go1.15
  platform:         linux/amd64
```
**What happened**:
Since upgrade to 0.17, warns are printed to log compact file with 
`level=warn ts=2020-11-19T10:44:50.465036587Z caller=s3.go:399 msg="could not guess file size for multipart upload; upload might be not optimized" name=debug/metas/01EQG1MNBMGVW310107ED36HGP.json err="unsupported type of io.Reader
`

yeya24:

## Changes

Fixes #3470. 

`bytes.Reader` type is not supported when guessing the file size.

### DIFF
--- a/pkg/block/block.go
+++ b/pkg/block/block.go
@@ -103,7 +103,7 @@ func Upload(ctx context.Context, logger log.Logger, bkt objstore.Bucket, bdir st
 		return errors.Wrap(err, "encode meta file")
 	}
 
-	if err := bkt.Upload(ctx, path.Join(DebugMetas, fmt.Sprintf("%s.json", id)), bytes.NewReader(metaEncoded.Bytes())); err != nil {
+	if err := bkt.Upload(ctx, path.Join(DebugMetas, fmt.Sprintf("%s.json", id)), &metaEncoded); err != nil {
 		return cleanUp(logger, bkt, id, errors.Wrap(err, "upload debug meta file"))
 	}
 
--- a/pkg/block/block.go
+++ b/pkg/block/block.go
@@ -98,12 +98,12 @@ func Upload(ctx context.Context, logger log.Logger, bkt objstore.Bucket, bdir st
 		return errors.Wrap(err, "gather meta file stats")
 	}
 
-	metaEncoded := bytes.Buffer{}
+	metaEncoded := strings.Builder{}
 	if err := meta.Write(&metaEncoded); err != nil {
 		return errors.Wrap(err, "encode meta file")
 	}
 
-	if err := bkt.Upload(ctx, path.Join(DebugMetas, fmt.Sprintf("%s.json", id)), &metaEncoded); err != nil {
+	if err := bkt.Upload(ctx, path.Join(DebugMetas, fmt.Sprintf("%s.json", id)), strings.NewReader(metaEncoded.String())); err != nil {
 		return cleanUp(logger, bkt, id, errors.Wrap(err, "upload debug meta file"))
 	}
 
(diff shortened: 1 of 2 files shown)


yeya24:
Screenshot from 2020-11-19 10-45-29

Well, the markdown checker is stupid... We need to exclude the katacoda dir.

yeya24:
There is one more case in `replicate` that doesn't work well with file size https://github.com/thanos-io/thanos/blob/master/pkg/replicate/scheme.go#L302.

Here the underlying type is `Minio.Object` and its size can be detected via a `Stat` call, is it good to support it? The code would be below, though it is ugly...

``` go
func TryToGetSize(r io.Reader) (int64, error) {
	switch f := r.(type) {
	case *os.File:
		fileInfo, err := f.Stat()
		if err != nil {
			return 0, errors.Wrap(err, "os.File.Stat()")
		}
		return fileInfo.Size(), nil
	case *bytes.Buffer:
		return int64(f.Len()), nil
	case *strings.Reader:
		return f.Size(), nil
	case *tracingReadCloser:
		rr, ok := f.r.(*timingReadCloser)
		if ok {
			switch ff := rr.ReadCloser.(type) {
			case *minio.Object:
				stat, err := ff.Stat()
				if err != nil {
					return 0, errors.Wrap(err, "")
				}
				return stat.Size, nil
			}
		}
	}
	return 0, errors.New("unsupported type of io.Reader")
}
```

bwplotka:
I would say yes, @yeya24 

The whole thing of dynamic check is ugly so no worries about this part ;p


I wonder .... is it a time to introduce `ReadSizer` interface finally here? :thinking:

yeya24:
Since we still need to create adapters for different `io.Reader` types, this sounds worse to me because we have to use these adapter types when getting size. It still requires a check step to convert `io.Reader` to our adapter. It is not good and might require some changes in our code, too.
I think the current way is fine.

pracucci (approved):
Thanks @yeya24! LGTM.

pkg/block/block.go
```diff
@@ -103,7 +103,7 @@ func Upload(ctx context.Context, logger log.Logger, bkt objstore.Bucket, bdir st
 		return errors.Wrap(err, "encode meta file")
 	}
 
-	if err := bkt.Upload(ctx, path.Join(DebugMetas, fmt.Sprintf("%s.json", id)), bytes.NewReader(metaEncoded.Bytes())); err != nil {
+	if err := bkt.Upload(ctx, path.Join(DebugMetas, fmt.Sprintf("%s.json", id)), &metaEncoded); err != nil {
```
GiedriusS, line 106: We are reading from `&metaEncoded` before this line hence the offset will be at the end. Or IOW this line will not get the data. The "correct" fix here is probably to use `strings.Builder` and then do `strings.NewReader` (:

GiedriusS (changes requested):
This is a bit more tricky than it seems. The CI failure is related.

pkg/block/block.go
yeya24: Thanks for the investigation! Let me fix this.

yeya24: Hi @GiedriusS, thanks for the help and the CI passed. Please take a look again when you have time.

GiedriusS (approved):
Thank you!

sherifkayad:
@yeya24 for some reason I am still getting warning log messages like:
```
level=warn ts=2021-08-02T13:48:48.7417482Z caller=s3.go:399 msg="could not guess file size for multipart upload; upload might be not optimized" name=debug/metas/01FC3FBJNKZK07BCKY16R25S4C.json err="unsupported type of io.Reader"
```

Is that normal? Can you advise? .. I am using Thanos v0.17.2

iridian-ks:
When using `replicate` the logs are completely littered with:

```
level=warn ts=2023-06-16T23:19:18.121636Z caller=s3.go:487 component=replicate msg="could not guess file size for multipart upload; upload might be not optimized" name=01E79EJ64K891AK9RNYN2E1CEC/chunks/000002 err="unsupported type of io.Reader: *minio.Object"
```

Is this something to worry about? I'm on the latest version Thanos.

PR #3496, opened 2020-11-23
s3: Respect SignatureV2 flag for all credential providers

simonswine:

## Changes

Thanos currently only supports V2 signatures when the credentials are statically specified in its configuration. This change supports using signature V2 on other credential sources, too.

## Verification

The object storage appliance is not fully compatible with S3V4 signatures and we are able to run it statically configured with V2. 

As we need to roll over the used accessKey/secretKey regularly we would like to be able to configure it via the `&credentials.IAM{}` chain.

### DIFF
--- a/pkg/objstore/s3/s3.go
+++ b/pkg/objstore/s3/s3.go
@@ -161,36 +161,54 @@ func NewBucket(logger log.Logger, conf []byte, component string) (*Bucket, error
 	return NewBucketWithConfig(logger, config, component)
 }
 
+type overrideSignerType struct {
+	credentials.Provider
+	signerType credentials.SignatureType
+}
+
+func (s *overrideSignerType) Retrieve() (credentials.Value, error) {
+	v, err := s.Provider.Retrieve()
+	if err != nil {
+		return v, err
+	}
+	if !v.SignerType.IsAnonymous() {
+		v.SignerType = s.signerType
+	}
+	return v, nil
+}
+
 // NewBucketWithConfig returns a new Bucket using the provided s3 config values.
 func NewBucketWithConfig(logger log.Logger, config Config, component string) (*Bucket, error) {
 	var chain []credentials.Provider
+
+	// TODO(bwplotka): Don't do flags as they won't scale, use actual params like v2, v4 instead
+	wrapCredentialsProvider := func(p credentials.Provider) credentials.Provider { return p }
+	if config.SignatureV2 {
+		wrapCredentialsProvider = func(p credentials.Provider) credentials.Provider {
+			return &overrideSignerType{Provider: p, signerType: credentials.SignatureV2}
+		}
+	}
 
 	if err := validate(config); err != nil {
 		return nil, err
 	}
 	if config.AccessKey != "" {
-		signature := credentials.SignatureV4
-		// TODO(bwplotka): Don't do flags, use actual v2, v4 params.
-		if config.SignatureV2 {
-			signature = credentials.SignatureV2
-		}
-
-		chain = []credentials.Provider{&credentials.Static{
+		chain = []credentials.Provider{wrapCredentialsProvider(&credentials.Static{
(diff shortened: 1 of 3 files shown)


pkg/objstore/s3/s3.go
```diff
@@ -161,36 +161,57 @@ func NewBucket(logger log.Logger, conf []byte, component string) (*Bucket, error
...
+		}
+		return p
+	}
+
 	if err := validate(config); err != nil {
 		return nil, err
 	}
 	if config.AccessKey != "" {
-		signature := credentials.SignatureV4
-		// TODO(bwplotka): Don't do flags, use actual v2, v4 params.
```
simonswine, line 173: @bwplotka I am not exactly too sure what exactly this TODO refers to and if it is still relevant.

Should I keep the TODO and if yes what would be a good place for it?

pkg/objstore/s3/s3.go
```diff
@@ -161,36 +161,57 @@ func NewBucket(logger log.Logger, conf []byte, component string) (*Bucket, error
 	return NewBucketWithConfig(logger, config, component)
 }
 
+type overrideSignerType struct {
+	signerType credentials.SignatureType
+	upstream   credentials.Provider
```
bwplotka, line 166:
```suggestion
	wrapped   credentials.Provider
```

It's just opinion, but we use `wrapped` for this kind of wrapped vars and overall wrapper patterns (:

pkg/objstore/s3/s3.go
```diff
@@ -161,36 +161,57 @@ func NewBucket(logger log.Logger, conf []byte, component string) (*Bucket, error
...
+	if err != nil {
+		return v, err
+	}
+	if !v.SignerType.IsAnonymous() {
+		v.SignerType = s.signerType
+	}
+	return v, nil
+}
+
+func (s *overrideSignerType) IsExpired() bool {
```
bwplotka, line 180: No need for this if we would embed `wrapped`/`upstream`

pkg/objstore/s3/s3.go
```diff
@@ -161,36 +161,57 @@ func NewBucket(logger log.Logger, conf []byte, component string) (*Bucket, error
...
 func NewBucketWithConfig(logger log.Logger, config Config, component string) (*Bucket, error) {
 	var chain []credentials.Provider
 
+	wrapCredentialsProvider := func(p credentials.Provider) credentials.Provider {
+		if config.SignatureV2 {
+			return &overrideSignerType{signerType: credentials.SignatureV2, upstream: p}
+		}
+		return p
+	}
+
```
bwplotka, line 194: What about something less dynamic? (: 
```suggestion
        wrapCredentialProvider := func(p credentials.Provider) credentials.Provider { return p }
        if config.SignatureV2 {
	wrapCredentialsProvider = func(p credentials.Provider) credentials.Provider { return &overrideSignerType{signerType: credentials.SignatureV2, upstream: p}
	}
}
```

bwplotka (approved):
Some style nits only. Thanks! :muscle: LGTM otherwise.

pkg/objstore/s3/s3.go
simonswine: Slightly reworded it and added the todo back in

pkg/objstore/s3/s3.go
simonswine: Rebase on embedding makes this unnecessary

pkg/objstore/s3/s3.go
simonswine: Is in the latest push :+1:

pkg/objstore/s3/s3.go
simonswine: Very good point, just done that as part of the last push

bwplotka (approved):
:muscle: Thanks!

LGTM

Issue #3150, opened 2020-09-11
store: Process start times growing linearly with blocks in objstore


gburek-fastly:
**Thanos, Prometheus and Golang version used**:

thanos, version 0.15.0 (branch: HEAD, revision: fbd14b49f95e7543883fedf9381a21127d4dda2b)
  build user:       root@d35b081d3e19
  build date:       20200907-09:29:21
  go version:       go1.14.2

**Object Storage Provider**: GCS 

**What happened**: The time that that Thanos Store takes to become Ready has steadily grown over the last few months, as our block count, including marked-for-deletion blocks, have grown. Currently, it takes about 20 min for the process to start and become ready for grpc query traffic. This makes recovery from OOMs and updates to configs dangerous, as process restarts must be coordinated between HA store nodes so that the service is not unavailable. 

**What you expected to happen**: Thanos Store process restart takes at most a few minutes. 

**How to reproduce it (as minimally and precisely as possible)**: 
1. Create GCS bucket with uploaded thanos blocks of type:
```
{state="loaded"} | 17291
{state="marked-for-deletion"} | 8543
{state="duplicate"} | 2795
{state="time-excluded"} | 2468
{state="no-meta-json"} | 7
{state="label-excluded"} | 0
{state="too-fresh"} | 0
{state="corrupted-meta-json"} | 0
{state="failed"} | 0
```
2. Configure thanos store to use that bucket via objstore.config and start the process
3. Monitor logs for `"duration="` to see the base fetcher duration

**Full logs to relevant components**:

```
level=info ts=2020-09-10T16:09:59.54113385Z caller=http.go:57 service=http/server component=store 
  msg="listening for requests and metrics" address=127.0.0.1:9090
level=info ts=2020-09-10T16:30:39.287137249Z caller=fetcher.go:453 component=block.BaseFetcher 
  msg="successfully synchronized block metadata" duration=20m39.746067576s cached=30429 returned=17226 partial=5
...
level=info ts=2020-09-10T16:31:42.380517614Z caller=store.go:329 
  msg="bucket store ready" init_duration=21m42.83945024s
```

**Anything else we need to know**:

It appears the bulk of the process start time is in https://github.com/thanos-io/thanos/blob/e0b7f7b32e9c1071586cb5fb1b4ed4f134235824/pkg/block/fetcher.go#L400, based on observing `thanos_objstore_bucket_operations_total`.

[...]

PR #3500, opened 2020-11-24
Add concurrency to IgnoreDeletionMarkFilter

pracucci:

## Changes

In Cortex we're currently testing a cluster with 6K tenants. Among the things we've discovered, we've noticed synching metas is slower than we would have expected. Further digging into it, looks like a significant % of time is spent by `IgnoreDeletionMarkFilter.Filter()` and the reason is that there's no parallelisation compared to the metas loading.

In this PR I would like to propose to introduce a concurrency to `IgnoreDeletionMarkFilter` too, set to the same value of the meta syncer concurrency.

I've run a test and below you can see the average meta sync time per tenant: on the left this PR code, on the right `master`. Obviously this is a non issue for a low volume of tenants and/or blocks, but becomes significative when you've to sync hundred of thousands of blocks.

Screenshot 2020-11-24 at 10 48 58

## Verification

Unit tests.

### DIFF
--- a/pkg/block/fetcher.go
+++ b/pkg/block/fetcher.go
@@ -762,16 +762,18 @@ func (f *ConsistencyDelayMetaFilter) Filter(_ context.Context, metas map[ulid.UL
 type IgnoreDeletionMarkFilter struct {
 	logger          log.Logger
 	delay           time.Duration
+	concurrency     int
 	bkt             objstore.InstrumentedBucketReader
 	deletionMarkMap map[ulid.ULID]*metadata.DeletionMark
 }
 
 // NewIgnoreDeletionMarkFilter creates IgnoreDeletionMarkFilter.
-func NewIgnoreDeletionMarkFilter(logger log.Logger, bkt objstore.InstrumentedBucketReader, delay time.Duration) *IgnoreDeletionMarkFilter {
+func NewIgnoreDeletionMarkFilter(logger log.Logger, bkt objstore.InstrumentedBucketReader, delay time.Duration, concurrency int) *IgnoreDeletionMarkFilter {
 	return &IgnoreDeletionMarkFilter{
-		logger: logger,
-		bkt:    bkt,
-		delay:  delay,
+		logger:      logger,
+		bkt:         bkt,
+		delay:       delay,
+		concurrency: concurrency,
 	}
 }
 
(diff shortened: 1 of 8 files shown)


GiedriusS:
This should close https://github.com/thanos-io/thanos/issues/3150, right?

pracucci:
> This should close #3150, right?

I wasn't aware of that issue, but looks it's an optimisation which may help with #3150.

GiedriusS (approved):
LGTM

pkg/block/fetcher.go
```diff
@@ -785,25 +787,69 @@ func (f *IgnoreDeletionMarkFilter) DeletionMarkBlocks() map[ulid.ULID]*metadata.
...
+				m := &metadata.DeletionMark{}
+				if err := metadata.ReadMarker(ctx, f.logger, f.bkt, id.String(), m); err != nil {
+					if errors.Cause(err) == metadata.ErrorMarkerNotFound {
+						continue
+					}
+					if errors.Cause(err) == metadata.ErrorUnmarshalMarker {
+						level.Warn(f.logger).Log("msg", "found partial deletion-mark.json; if we will see it happening often for the same block, consider manually deleting deletion-mark.json from the object storage", "block", id, "err", err)
+						continue
+					}
+					return err
```
zecke, line 815: Do I read it correctly that this stops one of the concurrent workers in case of errors?

pkg/block/fetcher.go
```diff
@@ -785,25 +787,69 @@ func (f *IgnoreDeletionMarkFilter) DeletionMarkBlocks() map[ulid.ULID]*metadata.
...
+		})
+	}
+
+	// Workers scheduled, distribute blocks.
+	eg.Go(func() error {
+		defer close(ch)
+
+		for _, id := range blockIDs {
+			select {
+			case ch <- id:
```
zecke, line 839: Can this block  when  2 x f.Concurrency errors are encountered and not all blocks have been processed yet?

pkg/block/fetcher.go
```diff
@@ -785,25 +787,69 @@ func (f *IgnoreDeletionMarkFilter) DeletionMarkBlocks() map[ulid.ULID]*metadata.
...
 
-		f.deletionMarkMap[id] = m
-		if time.Since(time.Unix(m.DeletionTime, 0)).Seconds() > f.delay.Seconds() {
-			synced.WithLabelValues(markedForDeletionMeta).Inc()
-			delete(metas, id)
-		}
+		return nil
+	})
+
+	if err := eg.Wait(); err != nil {
```
zecke, line 849: And if writing to the channel blocked then ```eg.Wait()``` would not return as it waits for all go routines to finish?

The simplest mitigation would be to change capacity of the channel from f.Concurrency to len(blockIDs)?

pkg/block/fetcher.go
pracucci: I think you're right. Let's move the discussion to https://github.com/thanos-io/thanos/pull/4074.

PR #3503, opened 2020-11-25
query: support basic auth for query API

oiooj:

## Changes

support basic auth to protect data.

image


GiedriusS:
Thank you for your contribution! However, historically we've been against adding such functionality to Thanos Query as it is very easy to place a reverse proxy in front of Thanos Query. As a result, it would be easier to add extra functionality e.g. what if we wanted to support multiple username/password combinations? IMHO we shouldn't merge such a thing. Does anyone else have any other thoughts?

wiardvanrij:
It feels bad writing this, as it is quite some work for this contribution but I have to agree with @GiedriusS .
There are so many solutions for adding auth. Be it basic auth, keycloak, or perhaps even LDAP. This does not have to be added on 'microservice' level. 
Especially in this case there is an added dependency which backfires when using Thanos Frontend Query (which could be your new endpoint from Grafana ;) ). I.e. this implementation should also be made on that component AND there should be a feature that Frontend Query can use basic auth towards Thanos Query. So this is quite some work and overhead which are IMO not worth it.

PR #3506, opened 2020-11-26
Expose metadata.Read()

pracucci:

## Changes

In Cortex we have the need to parse `meta.json` from a `io.Reader`. Since Thanos already has a function for that, can we expose it, instead of duplicating it?

## Verification

N/A


bwplotka (approved):
Ofc (: 

Thanks!

PR #3508, opened 2020-11-26
cmd/thanos/query.go: Timeout DNS resolution with refresh interval

lilic:
While the context is passed to DNS resolver, no timeout is ever set,
which leaves us in an unpredictable state, as we wait for the resolution
indefinitely. Every interval we repeat the DNS resolution, but don't ever
cancel previous attempts even if they have not finished.

Signed-off-by: Lili Cosic <cosiclili@gmail.com>

## Changes
- Adds timeout DNS resolution with refresh interval
While the context is passed to DNS resolver, no timeout is ever set,
which leaves us in an unpredictable state, as we wait for the resolution
indefinitely. Every interval we repeat the DNS resolution, but don't ever
cancel previous attempts even if they have not finished. 

Note as per https://github.com/thanos-io/thanos/pull/3508#issuecomment-734316160 this only applies to the default DNS resolver.

### DIFF
--- a/cmd/thanos/query.go
+++ b/cmd/thanos/query.go
@@ -391,6 +391,8 @@ func runQuery(
 		ctx, cancel := context.WithCancel(context.Background())
 		g.Add(func() error {
 			return runutil.Repeat(dnsSDInterval, ctx.Done(), func() error {
+				ctx, cancel = context.WithTimeout(ctx, dnsSDInterval)
+				defer cancel()
 				if err := dnsStoreProvider.Resolve(ctx, append(fileSDCache.Addresses(), storeAddrs...)); err != nil {
 					level.Error(logger).Log("msg", "failed to resolve addresses for storeAPIs", "err", err)
 				}


lilic:
cc @s-urbaniak @bwplotka we have been hitting some strange DNS resolution problems, so I found that we never actually timeout the DNS resolution. While this may or may not fix our bug, I believe it's the right thing to do, but only dived into the codebase recently so please advise if I am missing something obvious! Thanks!

s-urbaniak:
From my perspective this contribution makes total sense from a higher level perspective, however if I traverse into the underlying call of the miekgdns library I don't see the context being used (unless i miss some implementation detail):

https://github.com/thanos-io/thanos/blob/55cb8ca38b3539381dc6a781e637df15c694e50a/pkg/discovery/dns/miekgdns/resolver.go#L22-L50

lilic:
@s-urbaniak yes you are entirely correct it was the same finding for me, sorry for not mentioning it, the context is not used in the miek library, but it is used in the default go DNS library. I am not sure why they do not make use of it, I was planning on opening an issue for them to potentially handle the context. But this would not affect the current functionality as like you said context is not used, but would solve it for the default resolver.

s-urbaniak:
ahh, that makes perfect sense then, thank you!

bwplotka (approved):
I think this makes sense.

Wonder if `Repeat` should not have this functionality built-in even (: 

Thanks, LGTM :muscle:

lilic:
> Wonder if Repeat should not have this functionality built-in even (:

Yeah was thinking about that as well!

PR #3509, opened 2020-11-27
store add touch series limit

lisuo3389:

## Changes
Add touch series limit


cmd/thanos/store.go
```diff
@@ -67,6 +67,9 @@ func registerStore(app *extkingpin.App) {
 	maxSampleCount := cmd.Flag("store.grpc.series-sample-limit",
 		"Maximum amount of samples returned via a single Series call. The Series call fails if this limit is exceeded. 0 means no limit. NOTE: For efficiency the limit is internally implemented as 'chunks limit' considering each chunk contains 120 samples (it's the max number of samples each chunk can contain), so the actual number of samples might be lower, even though the maximum could be hit.").
 		Default("0").Uint()
+	maxTouchSeriesCount := cmd.Flag("store.grpc.touch-series-limit",
```
bwplotka, line 70: Amazing. What I miss here is the definition of "touched" (: 

Is this applied before fetching Series ID and chunk metas?
Is this applied while fetchink chunk metas and only if there is chunk matching time range?

This will allow understanding this better for users (:  I see from code it's the former and that's ok for now, however, hopefully, we can improve and touch less series here with this work: https://github.com/thanos-io/thanos/issues/3512

pkg/store/bucket.go
```diff
@@ -693,6 +703,11 @@ func blockSeries(
 		return storepb.EmptySeriesSet(), indexr.stats, nil
 	}
 
+	// Reserve seriesLimiter
```
bwplotka, line 706:
```suggestion
	// Reserve series seriesLimiter.
```

bwplotka (approved):
LGTM :hugs: 

Small nit! Thanks, amazing work and good idea :rocket:

bwplotka:
cc @yeya24 as we touch this code

cmd/thanos/store.go
lisuo3389: yes, Is this before fetching Series ID and chunk.

lisuo3389: I changed maxTouchSeriesCount to maxTouchedSeriesCount

cmd/thanos/store.go
```diff
@@ -67,6 +67,9 @@ func registerStore(app *extkingpin.App) {
 	maxSampleCount := cmd.Flag("store.grpc.series-sample-limit",
 		"Maximum amount of samples returned via a single Series call. The Series call fails if this limit is exceeded. 0 means no limit. NOTE: For efficiency the limit is internally implemented as 'chunks limit' considering each chunk contains 120 samples (it's the max number of samples each chunk can contain), so the actual number of samples might be lower, even though the maximum could be hit.").
 		Default("0").Uint()
+	maxTouchedSeriesCount := cmd.Flag("store.grpc.touched-series-limit",
+		"Maximum amount of touch series returned via a single Series call. The Series call fails if this limit is exceeded. 0 means no limit.").
```
yeya24, line 71: Small nit, this should be `touched` as well

pkg/store/bucket.go
```diff
@@ -190,6 +191,10 @@ func newBucketStoreMetrics(reg prometheus.Registerer) *bucketStoreMetrics {
 		Name: "thanos_bucket_store_queries_dropped_total",
 		Help: "Number of queries that were dropped due to the sample limit.",
 	})
+	m.queriesSeriesDropped = promauto.With(reg).NewCounter(prometheus.CounterOpts{
+		Name: "thanos_bucket_store_queries_series_dropped_total",
+		Help: "Number of queries that were dropped due to the series limit.",
```
yeya24, line 196: As I can see that if we want to add more metrics for the limit, why we just reuse `thanos_bucket_store_queries_dropped_total` metric? We can add a label `reason` for the reason of query drop such as `series`, `samples`, `chunks`, etc.

lisuo3389: If modify `thanos_bucket_store_queries_dropped_total`, it will affect `ChunksLimiterFactory`. Cortex depending it.

yeya24: No, the factory interface `ChunksLimiterFactory` won't be changed. The only change would be making `thanos_bucket_store_queries_dropped_total` a `CounterVec` type and adding one more label to it.

When initializing chunksLimiter you just need to do something like below so that's the same.

```
chunksLimiter    = s.chunksLimiterFactory(s.metrics.queriesDropped.WithLabelValues("chunks"))
```

It is also good to keep the current one because chunks and series are two dimensions. We can change it later if needed.

pkg/store/limiter.go
```diff
@@ -18,10 +18,19 @@ type ChunksLimiter interface {
...
+	// Returns an error if the limit has been exceeded. This function must be
+	// goroutine safe.
+	Reserve(num uint64) error
+}
+
 // ChunksLimiterFactory is used to create a new ChunksLimiter. The factory is useful for
 // projects depending on Thanos (eg. Cortex) which have dynamic limits.
 type ChunksLimiterFactory func(failedCounter prometheus.Counter) ChunksLimiter
 
+type SeriesLimiterFactory func(failedCounter prometheus.Counter) SeriesLimiter
```
yeya24, line 32: Would be good to add a comment.

yeya24 (approved):
LGTM

pkg/store/bucket.go
lisuo3389: Yes, you are right

bwplotka: Nice! Is it still todo?

bwplotka: I like this idea (:

bwplotka (approved):
LGTM, just @yeya24 comment still not addressed. 

We can do that in next Pr though

pkg/store/bucket.go
lisuo3389: Okay, let me modify it.

yeya24 (approved):
LGTM

GiedriusS (approved):
Thank you! LGTM, clean code. I have solved the conflicts for `CHANGELOG.md`. Will merge on green :heart:

GiedriusS (changes requested):
It seems like the CI failure is related to the changes. I was able to reproduce the segfault locally as well. Care looking into it?

lisuo3389:
> It seems like the CI failure is related to the changes. I was able to reproduce the segfault locally as well. Care looking into it?

What do i need to do?

GiedriusS (approved):
Thank you for the quick fix!

PR #3513, opened 2020-11-27
Update CircleCI build

SuperQ:
* Update to CircleCI 2.1 config.
* Use new executors feature.
* Use new CirleCI images.
* Update docs.

Signed-off-by: Ben Kochie <superq@gmail.com>


.circleci/config.yml
```diff
@@ -1,17 +1,18 @@
 # NOTE: Current plan gives 1500 build minutes per month.
-version: 2
-# https://circleci.com/blog/circleci-hacks-reuse-yaml-in-your-circleci-config-with-yaml/
-defaults: &defaults
-  docker:
-    # Built by Thanos make docker-ci
-    - image: &default-docker-image quay.io/thanos/thanos-ci:v1.2-go1.15-node
```
bwplotka, line 7: This was actually quite improving our build times we were not needed to prebuilding our deps like many Prometheus versions, alertmanager etc. Are we sure we can get rid of this? :thinking:

bwplotka: I mean we ended up reinstalling all anyway due to not updated thanos-ci image, but that separate story ;p https://app.circleci.com/pipelines/github/thanos-io/thanos/4397/workflows/f06871b1-d847-4332-95c0-bbd23f60c4fc

SuperQ: Looking at recent test pipeline runs, it's not really helping at all. Still about 14min to run the "test" pipeline.

It also looks like the current .promu.yml and other steps are missing `-mod=vendor`, so the vendor dir isn't being used at all.

bwplotka: True, but this is because we did not update this image lately OR somehow we put those tools in wrong dir or simply Makefile mod times are somehow not matching so Makefile reinstall. From looking on current pending it takes at least 5min to reinstall all tools to prebaking makes a lot of sense, WDYT?

SuperQ: IMO, 5 minutes is not worth it, as you need to constantly keep up with Go releases. We have the same problem with Prometheus [golang-builder](https://github.com/prometheus/golang-builder), but we have some cron jobs to bump the Go versions.

bwplotka: Hm.. for 15m worth of build cutting, 1/3 is IMO super worth it, but yes, we need some automation and more work on this. In fact, what would be helpful is to leverage CircleCI build cache instead. Let's merge this for now :+1:

PR #3517, opened 2020-11-30
receive: Clarified debug log message.

bwplotka:

I was scared I see lot's of uploads every minute, but we do that for no uploads as well (:

### DIFF
--- a/pkg/receive/multitsdb.go
+++ b/pkg/receive/multitsdb.go
@@ -18,8 +18,6 @@ import (
 	"github.com/prometheus/prometheus/pkg/labels"
 	"github.com/prometheus/prometheus/storage"
 	"github.com/prometheus/prometheus/tsdb"
-	"golang.org/x/sync/errgroup"
-
 	"github.com/thanos-io/thanos/pkg/block/metadata"
 	"github.com/thanos-io/thanos/pkg/component"
 	"github.com/thanos-io/thanos/pkg/errutil"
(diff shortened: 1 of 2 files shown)


brancz (approved):

kakkoyun (approved):
LGTM. CI is not happy with `sync/atomic` usage.

pkg/receive/multitsdb.go
```diff
@@ -202,16 +203,18 @@ func (t *MultiTSDB) Sync(ctx context.Context) error {
...
 		wg.Add(1)
 		go func() {
-			if uploaded, err := s.Sync(ctx); err != nil {
+			up, err := s.Sync(ctx)
+			if err != nil {
 				errmtx.Lock()
 				merr.Add(errors.Wrapf(err, "upload %d", uploaded))
 				errmtx.Unlock()
 			}
+			atomic.AddInt64(&uploaded, int64(up))
```
squat, line 212: If uploading the block results in an error, then we still call this, is that what we want? If so, then we are simply adding "1" for every single tenant. Rather than using an atomic counter, we could simply return the length of the `tenants`. This would be simpler code and avoid using `atomic`. If we don't want to count errored blocks, then we could still do this without `sync/atomic` by returning the length of the tenants - the length of the multi-error.

bwplotka: We actually:
* can  upload more than one for tenant.
* return error also when we upload one but second failed

So I think we need something like this... WDYT?

squat: ack I see. makes sense :)

Issue #3518, opened 2020-11-30
receive: Cannot use optimized multi part upload.


bwplotka:
```
level=warn ts=2020-11-30T09:00:18.067006266Z caller=s3.go:399 component=receive msg="could not guess file size for multipart upload; upload might be not optimized" name=debug/metas/01ERC63277SJSMAQ2P1W132BH6.json err="unsupported type of io.Reader"
```

Sad outcome of dynamic casting ): 

Help wanted :hugs:

GiedriusS:
https://github.com/thanos-io/thanos/pull/3476

erdong:
I had the same problem. Thanos version is 0.17.0 .

When will this fixed version be released ？

keyhana:
Same here. After upgrading from 0.16.0 to 0.17.1 we are getting this warning from Thanos sidecar and Thanos compactor.
It is also causing high memory usage after the warning is printed which ultimately gets our pods OOMKilled.

ethernoy:
same here, but the compactor seems to be working though, is it just a warning or it actually stops sth from working?

ahurtaud:
Hello, can we get an update on the PR #3476 and so on this issue please?

We cannot load v0.17.x because of this and we would like not to miss it on 0.18 :).

Thank you

mfuezesi:
Hello there
Really like your work.  But now I have the same problem. v16 works, v17 not. With the same issue. Is there another workaround?

GiedriusS:
Fixed in https://github.com/thanos-io/thanos/pull/3476.

PR #3519, opened 2020-11-30
pkg/objstore: make error msg more accurate.

johncming:

## Changes

Fix error output in the log.

Details:

This pr is associated with an issue https://github.com/thanos-io/thanos/issues/3518

## Verification

Review twice😸


bwplotka (approved):
Thanks :+1:

PR #3520, opened 2020-11-30
fix index out of bound bug when comparing ZLabelSets

### LINKED ISSUE #3515: query-frontend 0.17.1: runtime error: index out of range
I faced runtime error: `index out of range [7] with length 7` in query-frontend 0.17.1. There were huge amount of such errors (I suppose on each request), attached full log for one such error.
Unfortunately I do not know how to reproduce it, so far it happened only once.
[query-frontend-0.17.1.err.log](https://github.com/thanos-io/thanos/files/5614135/query-frontend-0.17.1.err.log)

yeya24:

## Changes

Fixes #3515 

The bug is originated from that I got the label pair length using `Size()` method, it is a stupid mistake :-1:  because it doesn't return the number of label pairs.
Added one test case to ensure no panic for comparing the same label sets.
However, we deduplicate the series before sorting them so the case shouldn't happen. Still investigating the root cause and trying to reproduce this from the query frontend side.

### DIFF
--- a/pkg/queryfrontend/labels_codec.go
+++ b/pkg/queryfrontend/labels_codec.go
@@ -87,8 +87,25 @@ func (c labelsCodec) MergeResponse(responses ...queryrange.Response) (queryrange
 		if len(responses) == 1 {
 			return responses[0], nil
 		}
-		seriesData := make(labelpb.ZLabelSets, 0)
 
+		i := 0
+		var resp queryrange.Response
+		for _, response := range responses {
+			if len(response.(*ThanosSeriesResponse).Data) > 0 {
+				i++
+				resp = response
+			}
+			if i > 1 {
+				break
+			}
+		}
+		// Fast path for the case that only one response contains series data and other responses are empty.
+		// We can just return that response to avoid deduplication and sorting.
+		if i == 1 {
+			return resp, nil
+		}
+
+		seriesData := make(labelpb.ZLabelSets, 0)
 		uniqueSeries := make(map[string]struct{})
 		for _, res := range responses {
 			for _, series := range res.(*ThanosSeriesResponse).Data {
@@ -287,7 +304,7 @@ func (c labelsCodec) parseLabelsRequest(r *http.Request, op string) (queryrange.
 		return nil, err
 	}
 
-	result.StoreMatchers, err = parseMatchersParam(r.Form[queryv1.StoreMatcherParam])
+	result.StoreMatchers, err = parseMatchersParam(r.Form[queryv1.StoreMatcherParam], queryv1.StoreMatcherParam)
 	if err != nil {
 		return nil, err
 	}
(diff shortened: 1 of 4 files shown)


metalmatze:
Generally looking good. Some other pair of eyes would be good, as I'm not too familiar with this part.

pkg/queryfrontend/labels_codec.go
```diff
@@ -87,8 +87,25 @@ func (c labelsCodec) MergeResponse(responses ...queryrange.Response) (queryrange
 		if len(responses) == 1 {
 			return responses[0], nil
 		}
-		seriesData := make(labelpb.ZLabelSets, 0)
 
+		i := 0
+		var resp queryrange.Response
+		for _, response := range responses {
+			if len(response.(*ThanosSeriesResponse).Data) > 0 {
```
bwplotka, line 94: I don't get this, why we cannot check this in loop below?

pkg/queryfrontend/queryrange_codec.go
```diff
@@ -111,7 +111,7 @@ func (c queryRangeCodec) DecodeRequest(_ context.Context, r *http.Request) (quer
 		result.ReplicaLabels = r.Form[queryv1.ReplicaLabelsParam]
 	}
 
-	result.StoreMatchers, err = parseMatchersParam(r.Form[queryv1.StoreMatcherParam])
+	result.StoreMatchers, err = parseMatchersParam(r.Form[queryv1.StoreMatcherParam], queryv1.StoreMatcherParam)
```
bwplotka, line 114: This is bad code smell - passing same information twice. Maybe we should pass just `Form` and `matcherKey` to param function?

pkg/store/labelpb/label.go
```diff
@@ -307,7 +307,8 @@ func (z ZLabelSets) Less(i, j int) bool {
 	l := 0
 	r := 0
 	var result int
-	for l < z[i].Size() && r < z[j].Size() {
+	lenI, lenJ := len(z[i].Labels), len(z[j].Labels)
```
bwplotka, line 310: :+1:  (:

bwplotka (approved):
Thanks, but would be amazing to not have unrelated to fix changes in quick bug fix PR. 

Since you have those, there are minor issues/suggestions for unrelated changes, otherwise LGTM

pkg/queryfrontend/queryrange_codec.go
yeya24: Good point.

pkg/queryfrontend/labels_codec.go
yeya24: Makes sense. As you mentioned I ll remove this change since it is unrelated and fix them in next master

pkg/store/labelpb/label.go
yeya24: :(

yeya24:
Hi, please take a look again. @bwplotka  @metalmatze

bwplotka (approved):
Awesome! Thanks :muscle:

bwplotka:
We need changelog but we can add on release PR

iyacontrol:
when release v0.17.2 ? i meet this bug, so had to go back v0.16.0.

PR #3524, opened 2020-12-01
cmd/thanos: fix DNS resolution when ctx is canceled

simonpasquier:

## Changes

c21337e5 introduced a subtle bug: when the DNS resolution context
expires, all subsequent resolutions will fail with 'operation was
canceled' because the function overwrites the global context instead of
creating child context. To make it more obvious, we use different
variable names for the parent and child contexts.

## Verification

Here is a minimal reproducer:
https://play.golang.org/p/Gw9B1cU3mtL

### DIFF
--- a/cmd/thanos/query.go
+++ b/cmd/thanos/query.go
@@ -391,12 +391,12 @@ func runQuery(
 		ctx, cancel := context.WithCancel(context.Background())
 		g.Add(func() error {
 			return runutil.Repeat(dnsSDInterval, ctx.Done(), func() error {
-				ctx, cancel = context.WithTimeout(ctx, dnsSDInterval)
-				defer cancel()
-				if err := dnsStoreProvider.Resolve(ctx, append(fileSDCache.Addresses(), storeAddrs...)); err != nil {
+				resolveCtx, resolveCancel := context.WithTimeout(ctx, dnsSDInterval)
+				defer resolveCancel()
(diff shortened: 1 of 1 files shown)


simonpasquier:
cc @lilic @s-urbaniak

simonpasquier:
Submitted https://github.com/thanos-io/thanos/pull/3524 to address the canceled context issue.

lilic:
Thanks, nice catch! Not sure what you meant by  https://github.com/thanos-io/thanos/pull/3524#issuecomment-736649144 comment here?

cmd/thanos/query.go
```diff
@@ -391,12 +391,12 @@ func runQuery(
 		ctx, cancel := context.WithCancel(context.Background())
 		g.Add(func() error {
 			return runutil.Repeat(dnsSDInterval, ctx.Done(), func() error {
-				ctx, cancel = context.WithTimeout(ctx, dnsSDInterval)
```
bwplotka, line 394: Shadowing variables... the worst... :face_with_head_bandage: 

The fix would be just to add `:` (:  but happy with explicit version as well :+1:

bwplotka (approved):
Nice catch.

Linter for shadowing variables is very much needed.

bwplotka:
Thanks a lot!

simonpasquier:
> Thanks, nice catch! Not sure what you meant by [#3524 (comment)](https://github.com/thanos-io/thanos/pull/3524#issuecomment-736649144) comment here?

Ah this comment was meant to be for your CMO pull request :)

Issue #3526, opened 2020-12-01
query-frontend: results not inclusive when start/end are the same


adamcstephens:
Both the prometheus and query APIs will return the inclusive results for query_range if the start and end times are the same. Query frontend returns no results, thus breaking compatibility with the other two APIs.

**Thanos, Prometheus and Golang version used**:
Thanos - 0.16.0, 0.17.1, master
Prometheus - 2.21.0
Golang - 1.15.5

**Object Storage Provider**:
minio

**What happened**:
No results were returned with query_range start and end set to the same value

**What you expected to happen**:
Return an inclusive set of data, as both Prometheus and Query do

**How to reproduce it (as minimally and precisely as possible)**:
query_range query against query-frontend using the same start/end time.
```
http --follow "http://localhost:19090/api/v1/query_range?query=up&start=$(gdate +%s)&end=$(gdate +%s)&step=120" | jq '.data.result | length'
```

**Anything else we need to know**:

prometheus:
```
❯ http --follow "http://localhost:9001/api/v1/query_range?query=up&start=$(gdate +%s)&end=$(gdate +%s)&step=120" | jq '.data.result | length'
7
```

query:
Doubled since env has two prometheus/sidecar

```
❯ http --follow "http://localhost:10902/api/v1/query_range?query=up&start=$(gdate +%s)&end=$(gdate +%s)&step=120" | jq '.data.result | length'
14
```

query-frontend:
```
❯ http --follow "http://localhost:19090/api/v1/query_range?query=up&start=$(gdate +%s)&end=$(gdate +%s)&step=120" | jq '.data.result | length'
0
```

Same query with a different start time

```
❯ http --follow "http://localhost:19090/api/v1/query_range?query=up&start=$(gdate --date='5 minutes ago' +%s)&end=$(gdate +%s)&step=120" | jq '.data.result | length'
14
```

PR #3527, opened 2020-12-01
query-frontend: inclusive results when start/end are the same

adamcstephens:

## Verification

Verified same way bug report #3526 was tested. When passing start/end with the same value, results are now returned.

```
❯ http --follow "http://localhost:19090/api/v1/query_range?query=up&start=$(gdate +%s)&end=$(gdate +%s)&step=120" | jq '.data.result | length'
14
❯ http --follow "http://localhost:19090/api/v1/query_range?query=up&start=$(gdate --date='5 minutes ago' +%s)&end=$(gdate +%s)&step=120" | jq '.data.result | length'
14
```

### DIFF
--- a/pkg/queryfrontend/split_by_interval.go
+++ b/pkg/queryfrontend/split_by_interval.go
@@ -68,13 +68,17 @@ func (s splitByInterval) Do(ctx context.Context, r queryrange.Request) (queryran
 func splitQuery(r queryrange.Request, interval time.Duration) []queryrange.Request {
 	var reqs []queryrange.Request
 	if _, ok := r.(*ThanosQueryRangeRequest); ok {
-		for start := r.GetStart(); start < r.GetEnd(); start = nextIntervalBoundary(start, r.GetStep(), interval) + r.GetStep() {
-			end := nextIntervalBoundary(start, r.GetStep(), interval)
-			if end+r.GetStep() >= r.GetEnd() {
-				end = r.GetEnd()
-			}
+		if start := r.GetStart(); start == r.GetEnd() {
+			reqs = append(reqs, r.WithStartEnd(start, start))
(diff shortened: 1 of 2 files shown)


pkg/queryfrontend/split_by_interval.go
```diff
@@ -68,13 +68,17 @@ func (s splitByInterval) Do(ctx context.Context, r queryrange.Request) (queryran
...
 	var reqs []queryrange.Request
 	if _, ok := r.(*ThanosQueryRangeRequest); ok {
-		for start := r.GetStart(); start < r.GetEnd(); start = nextIntervalBoundary(start, r.GetStep(), interval) + r.GetStep() {
-			end := nextIntervalBoundary(start, r.GetStep(), interval)
-			if end+r.GetStep() >= r.GetEnd() {
-				end = r.GetEnd()
+		if start := r.GetStart(); start == r.GetEnd() {
+			reqs = append(reqs, r.WithStartEnd(start, start))
+		} else {
+			for start := r.GetStart(); start < r.GetEnd(); start = nextIntervalBoundary(start, r.GetStep(), interval) + r.GetStep() {
```
yeya24, line 74: One small nit, this `start := r.GetStart()` is redundant since `start` is initialized at line 71.

yeya24 (approved):
LGTM, thanks for the contribution!

Maybe @pracucci would be interested in this one since the Cortex frontend has the same issue.

pkg/queryfrontend/split_by_interval.go
adamcstephens: Updated. Thanks :)

bwplotka:
Thank you @adamcstephens and thank you @yeya24 for review. :muscle:

Issue #3456, opened 2020-11-17
query: API does not honor the per request timeout URL query parameter (one slow store makes all of thanos unavailable)


zecke:
**Thanos, Prometheus and Golang version used**:

Thanos v0.16.0
Go 1.15.3

**Object Storage Provider**:

S3

**What happened**:

Executing:

* thanos query --query.timeout=2m (default)
* curl http://127.1:9090/query_range?...&timeout=10s&partial_response=1

Result:
Query timesout after 2m and not 10s.

**What you expected to happen**:

I did expect the query to timeout after 10s (and deliver a partial response)

**How to reproduce it (as minimally and precisely as possible)**:

Make one thanos store really slow. We added a `srv.Context().Done()` into the Series() implementation of the Store. This should force timeouts and hopefully partial responses as other stores have data for the same time range from the same object store.

**Full logs to relevant components**:

The API correctly parses the timeout and creates the matching context:
https://github.com/thanos-io/thanos/blob/v0.16.0/pkg/api/query/v1.go#L232
https://github.com/thanos-io/thanos/blob/v0.16.0/pkg/api/query/v1.go#L336

But the querier prefers to use the global timeout over the per query timeout:

https://github.com/thanos-io/thanos/blob/v0.16.0/pkg/query/querier.go#L210-L211

This was changed in https://github.com/thanos-io/thanos/commit/c9c60c1b8c9e6cb8dd0c0a51513e3894c558b48d and seems a regression. After this change the q.selectTimeout is passed through the QueryableCreator.

The Prometheus Query Engine does seem to call q.Select in a blocking fashion and this means the entire query is blocked. Changing this to use q.ctx does fix the timeout handling but partial_responses seems to remain broken.

bwplotka:
ack, sounds like a bug, thanks for reporting! :muscle:

spaparaju:
Picking up this issue.

zecke:
The issue #2144 is related to this problem for partial_response handling. Let's start with a request like the one below and have one _slow_ store (but we have N+2 stores for the same time range. This means from a user point of view we can tolerate some replicas not providing us with _any_ result).


```curl http://127.1:9090/query_range?...&timeout=10s&partial_response=1```

What will happen with the default thanos configuration (query timeout of 2m, store reply timeout of 0s being disable).

Without this issue fixed: The query will timeout after 2m with no data.
With this issue fixed: The query will timeout after 10s with no data.

[...]

PR #3539, opened 2020-12-04
querier: Queries timeout based on lower value out of global Query timeout and per-request timeout specified as part of the query API

spaparaju:
Instead of using global timeout (at Query component level),  API calls to Query component get timed-out as per-request timeout (if per-request timeout is specified).  This is tracked at this [issue](Bug fix for https://github.com/thanos-io/thanos/issues/3456).

* [x]  I added CHANGELOG entry for this change.

## Changes
Instead of using global timeout (at Query component level),  API calls to Query component get timed-out as per-request timeout (if per-request timeout is specified). 

## Verification

- To simulate slow queries, added time.Sleep() at Series() implementation of [underlying Prometheus store](https://github.com/thanos-io/thanos/blob/a57bbfbe4dc4838093c499b5afa72ed305d4eaae/pkg/store/prometheus.go#L135)
- Observed the time-out behaviour in the cases of specifying / not specifying per-request timeout at the Query API calls. 
    - http://localhost:10904/api/v1/query?query=go_memstats_heap_inuse_bytes&step=10s&timeout=1s
     - http://localhost:10904/api/v1/query_range...


CHANGELOG.md
```diff
@@ -24,6 +24,8 @@ We use _breaking :warning:_ to mark changes that are not backward compatible (re
 ### Changed
 
 - [#3496](https://github.com/thanos-io/thanos/pull/3496) s3: Respect SignatureV2 flag for all credential providers.
+- [#3539](https://github.com/thanos-io/thanos/pull/3539) Query:  API calls to Query component honour per-request timeout (URL parameter)
```
squat, line 27: nit: add period to the end of the sentence

pkg/query/querier.go
```diff
@@ -208,7 +208,19 @@ func (q *querier) Select(_ bool, hints *storage.SelectHints, ms ...*labels.Match
 	// The querier has a context but it gets canceled, as soon as query evaluation is completed, by the engine.
 	// We want to prevent this from happening for the async storea API calls we make while preserving tracing context.
 	ctx := tracing.CopyTraceContext(context.Background(), q.ctx)
-	ctx, cancel := context.WithTimeout(ctx, q.selectTimeout)
+
+	// Check if per-request timeout is specified
```
squat, line 212: tiny nit: comments should be full sentences with `.`s at the end.

squat (approved):
Looks good functionally :))) just small stylistic nits

pkg/query/querier.go
```diff
@@ -208,7 +208,19 @@ func (q *querier) Select(_ bool, hints *storage.SelectHints, ms ...*labels.Match
...
+	perRequestCancelTime, ifSpecified := q.ctx.Deadline()
+
+	var perQueryTimeOutValue time.Duration
+	if ifSpecified {
+		// use the per-request timeout specified
```
bwplotka, line 217: Comments has to be full sentence (: 

(https://thanos.io/tip/contributing/coding-style-guide.md/#comments-should-be-full-sentences)

pkg/query/querier.go
```diff
@@ -208,7 +208,19 @@ func (q *querier) Select(_ bool, hints *storage.SelectHints, ms ...*labels.Match
 	// The querier has a context but it gets canceled, as soon as query evaluation is completed, by the engine.
 	// We want to prevent this from happening for the async storea API calls we make while preserving tracing context.
 	ctx := tracing.CopyTraceContext(context.Background(), q.ctx)
-	ctx, cancel := context.WithTimeout(ctx, q.selectTimeout)
+
+	// Check if per-request timeout is specified
+	perRequestCancelTime, ifSpecified := q.ctx.Deadline()
+
+	var perQueryTimeOutValue time.Duration
+	if ifSpecified {
```
bwplotka, line 216:
```suggestion
	if deadlineSpecified {
```

pkg/query/querier.go
```diff
@@ -208,7 +208,19 @@ func (q *querier) Select(_ bool, hints *storage.SelectHints, ms ...*labels.Match
 	// The querier has a context but it gets canceled, as soon as query evaluation is completed, by the engine.
 	// We want to prevent this from happening for the async storea API calls we make while preserving tracing context.
 	ctx := tracing.CopyTraceContext(context.Background(), q.ctx)
-	ctx, cancel := context.WithTimeout(ctx, q.selectTimeout)
+
+	// Check if per-request timeout is specified
+	perRequestCancelTime, ifSpecified := q.ctx.Deadline()
+
+	var perQueryTimeOutValue time.Duration
```
bwplotka, line 215:
```suggestion
	var perQueryTimeout time.Duration
```
We know it's value, right?

bwplotka (changes requested):
Can we have some unit test showing that this works?

Because I think we don't need this. We depend on `q.ctx` then we create another context with another timeout from that context and we use it. This mean that the whatever deadline was shorter, it will be triggered :hugs: 

So I don't think this is needed, the bug might be different.

renzel24:
I'm scared to release it what is thanos lol

bwplotka:
> I'm scared to release it what is thanos lol

?

CHANGELOG.md
spaparaju: Thanks! This suggested change is taken care.

pkg/query/querier.go
spaparaju: Thanks! This suggested change is taken care.

pkg/query/querier.go
spaparaju: Thanks! This suggested change is taken care.

pkg/query/querier.go
spaparaju: makes sense!

pkg/query/querier.go
spaparaju: placeholder to contain the timeout value (to apply) in both the possibilities (per-request timeout is / is not specified

pkg/query/querier.go
```diff
@@ -208,7 +208,19 @@ func (q *querier) Select(_ bool, hints *storage.SelectHints, ms ...*labels.Match
...
 	ctx := tracing.CopyTraceContext(context.Background(), q.ctx)
-	ctx, cancel := context.WithTimeout(ctx, q.selectTimeout)
+
+	perRequestCancelTime, deadlineSpecified := q.ctx.Deadline()
+
+	var perQueryTimeOut time.Duration
+
+	// If the per-request timeout is specified, the query would timeout per the specified per-request timeout value.
+	// If the per-request timeout is not specified, the query would timeout per the timeout value specified at global query component.
+	if deadlineSpecified {
```
bwplotka, line 218: As I commented before: Not sure if we need this. We depend on q.ctx then we create another context with another timeout from that context and we use it. This mean that the whatever deadline was shorter, it will be triggered

bwplotka: What I am saying is that we can do just

```
ctx, cancel := context.WithTimeout(ctx, perQueryTimeOut)
ctx, cancel := context.WithTimeout(ctx, perRequestCancelTime)
```
Or maybe for minor "efficiency" (one goroutine less)

```
timeout := time.Until(perRequestCancelTime)
if timeout > q.selectTimeout {
    timeout = q.selectTimeout
}
```

spaparaju: This check (if deadlineSpecified) helps two distinct possibilities: (1) User might not specify any per-request timeout value as its optional (2) User might wanna wait longer time than default query component level timeout

bwplotka: Cool, I think I missed (2) totally. The only concern I have is that it's common missed idea and we might have enforcement of some default timeouts in other places as well. Let's add ticket for this / docs.

test/e2e/query_test.go
```diff
@@ -169,6 +170,42 @@ func TestQuery(t *testing.T) {
...
+	prom4, sidecar4, err := e2ethanos.NewPrometheusWithSidecar(s.SharedDir(), "e2e_test_query_for_url_timeout", "ha2", defaultPromConfig("prom-ha", 1, "", filepath.Join(e2e.ContainerSharedDir, "", "*.yaml")), e2ethanos.DefaultPrometheusImage())
+	testutil.Ok(t, err)
+	testutil.Ok(t, s.StartAndWaitReady(prom1, sidecar1, prom2, sidecar2, prom3, sidecar3, prom4, sidecar4))
+
+	// Querier. Both fileSD and directly by flags.
+	q, err := e2ethanos.NewQuerier(s.SharedDir(), "1", []string{sidecar1.GRPCNetworkEndpoint(), sidecar2.GRPCNetworkEndpoint(), receiver.GRPCNetworkEndpoint()}, []string{sidecar3.GRPCNetworkEndpoint(), sidecar4.GRPCNetworkEndpoint()}, nil, "", "")
+	testutil.Ok(t, err)
+	testutil.Ok(t, s.StartAndWaitReady(q))
+
+	ctx, cancel := context.WithTimeout(context.Background(), 1*time.Minute)
```
bwplotka, line 202: 10*time.minute?

test/e2e/query_test.go
```diff
@@ -169,6 +170,42 @@ func TestQuery(t *testing.T) {
...
+	testutil.Ok(t, err)
+	prom3, sidecar3, err := e2ethanos.NewPrometheusWithSidecar(s.SharedDir(), "e2e_test_query_for_url_timeout", "ha1", defaultPromConfig("prom-ha", 0, "", filepath.Join(e2e.ContainerSharedDir, "", "*.yaml")), e2ethanos.DefaultPrometheusImage())
+	testutil.Ok(t, err)
+	prom4, sidecar4, err := e2ethanos.NewPrometheusWithSidecar(s.SharedDir(), "e2e_test_query_for_url_timeout", "ha2", defaultPromConfig("prom-ha", 1, "", filepath.Join(e2e.ContainerSharedDir, "", "*.yaml")), e2ethanos.DefaultPrometheusImage())
+	testutil.Ok(t, err)
+	testutil.Ok(t, s.StartAndWaitReady(prom1, sidecar1, prom2, sidecar2, prom3, sidecar3, prom4, sidecar4))
+
+	// Querier. Both fileSD and directly by flags.
+	q, err := e2ethanos.NewQuerier(s.SharedDir(), "1", []string{sidecar1.GRPCNetworkEndpoint(), sidecar2.GRPCNetworkEndpoint(), receiver.GRPCNetworkEndpoint()}, []string{sidecar3.GRPCNetworkEndpoint(), sidecar4.GRPCNetworkEndpoint()}, nil, "", "")
```
bwplotka, line 198: Let's modify (somehow) adjust select timeout flag to ridiculous small timeout 1s.

test/e2e/query_test.go
```diff
@@ -169,6 +170,42 @@ func TestQuery(t *testing.T) {
 	})
 }
 
+// This test checks if per-request timeout value specified is honored.
+// Pre-requisite for testing this case is to simulate slow response from the underlying TSDB.
+// One option to simulate slow response is to include 'sleep' instruction at Series() implementation at pkg/store/tsdb.go.
+func TestQueryForURLTimeout(t *testing.T) {
```
bwplotka, line 176: Yea, this does not tests much.

bwplotka (dismissed):
Cool, almost done. (: 

LGTM,  otherwise.

spaparaju:
Here are the new findings: the [timeout at the context of a read API call](https://github.com/thanos-io/thanos/blob/a24acb6073a2d0ccd739b9f05be03534a8f192f0/pkg/query/querier.go#L210) already contains the lower value of global query timeout and per-request timeout' .  With this,  we don't get into different scenarios  like lower/higher global query timeout , lower/higher per-request timeout. Instead of picking up global query timeout, timeout at the current context which is lower of global query timeout, per-request timeout need to be used to evaluate queries.

pkg/query/querier.go
```diff
@@ -208,7 +208,10 @@ func (q *querier) Select(_ bool, hints *storage.SelectHints, ms ...*labels.Match
 	// The querier has a context but it gets canceled, as soon as query evaluation is completed, by the engine.
 	// We want to prevent this from happening for the async storea API calls we make while preserving tracing context.
 	ctx := tracing.CopyTraceContext(context.Background(), q.ctx)
```
bwplotka, line 210: Does q.ctx has `ctx, cancel := context.WithTimeout(ctx, q.selectTimeout)`? select Timeout? I don't think we ever set this timeout into q.ctx

bwplotka: Question: Ideally we want request to override timeout - let's see if we can do it.

spaparaju: For q.ctx, 'per-request timeout' is set at  [Thanos Query API](https://github.com/thanos-io/thanos/blob/c634e312e080f531747412285262fea902033215/pkg/api/query/v1.go#L242) and later in call chain 'engine timeout' is set at [Prometheus layer](https://github.com/prometheus/prometheus/blob/41247b69507342b8d26110207af988ad5a1310e5/promql/engine.go#L440) . So the downstream call-chain at package like pkg/query, would not carry per-request timeout any more. 

Added the unit tests to test if lower value out of {global Query timeout, per-request timeout} is used to execute queries.

spaparaju:
for easier reviews, opened a [new PR](https://github.com/thanos-io/thanos/pull/4614) to fix this bug.

PR #3562, opened 2020-12-10
query: Be pessimistic about strict stores

zecke:
We want queries with ?partial_response=0 to be sound and complete.
Unfortunately today many queries can succeed without an error while the
system is heavily degraded.

Make strict stores cover a long time range by default until the right
time range has been discovered. We might need to extend this to labels
too. Also on many consecutive Metadata failures we should revert to a
default minTime and maxTime.

## Changes

StoreSet
* Initialize storeRef maxTime to the maximum possible value
* The first time we set the status, initialize minTime and maxTime to the provided values. This is for initializing with an error case.

## Verification

go test ./pkg/query only so far

### DIFF
--- a/pkg/query/storeset.go
+++ b/pkg/query/storeset.go
@@ -7,6 +7,7 @@ import (
 	"context"
 	"encoding/json"
 	"fmt"
+	"math"
 	"sort"
 	"sync"
 	"time"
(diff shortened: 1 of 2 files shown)


zecke:
In general we prefer availability (partial_responses) but I have uses cases where a _wrong_ non-partial is worse than an error. I have started to use strict stores but this requires the query component to communicate with that store at least once to have the desired impact on ?partial_response=0. The easiest option seems to default initialize maxTime and get the store included in every query.

It's WIP to mostly start a discussion. I have seen two more cases where partial_response=0 will succeed but should give an error.

bwplotka (approved):
This is making sense to me. I love this, sorry for massive lag, next time, feel free to ping us on #thanos-dev if you did not already.

LGTM and thanks for tests!

PR #3566, opened 2020-12-10
Support label matchers in labels API

### LINKED ISSUE #3351: storeAPI: Extend ExternalLabels Selector for LabelsNames and LabelValues
This would be really awesome for further multi-tenancy support with https://github.com/prometheus-community/prom-label-proxy

Since tenants never share the same TSDB block or DB we can easily allow LabelValues and LabelNames API selection support for tenant's data.

The only open question is compatibility. Do we want extra check to ensure if server supports this or not? Without such check we know nothing: The lack of selection might be transparent causing tenant data leak.

cc @pracucci @brancz @yeya24 @kakkoyun

yeya24:

## Changes

Fixes #3351 

Add `match[]` param for LabelNames and LabelValues queries.

Now this is implemented via the `Series` matching. We can do matchers push down to improve this later.

TODO: query frontend also needs `matchers` type for Label names and label values query. Will add them in the next pr.

## Verification

Add unit tests and E2E tests for this feature.

### DIFF
--- a/pkg/api/query/v1.go
+++ b/pkg/api/query/v1.go
@@ -23,6 +23,7 @@ import (
 	"context"
 	"math"
 	"net/http"
+	"sort"
 	"strconv"
 	"strings"
 	"time"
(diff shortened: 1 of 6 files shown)


yeya24:
E2E is flaky #3570

prmsrswt (approved):
Looks good to me!

pkg/api/query/v1.go
```diff
@@ -439,16 +440,39 @@ func (qapi *QueryAPI) labelValues(r *http.Request) (interface{}, []error, *api.A
...
+	var matcherSets [][]*labels.Matcher
+	for _, s := range r.Form[MatcherParam] {
+		matchers, err := parser.ParseMetricSelector(s)
+		if err != nil {
+			return nil, nil, &api.ApiError{Typ: api.ErrorBadData, Err: err}
+		}
+		matcherSets = append(matcherSets, matchers)
+	}
+
+	q, err := qapi.queryableCreate(true, nil, storeDebugMatchers, 0, enablePartialResponse, true).
```
prmsrswt, line 452: Not a nit, just wanted to ask a question. Why did you flip the last argument (`skipChunks`) to `true`, and what it does exactly?

yeya24: A query result contains two parts, the series metadata like label sets, and the sample values. In this case, since we are just querying labels, we can set `skipChunks` to not return samples, only labels are fine.

yeya24: The original issue about this option https://github.com/thanos-io/thanos/issues/1822. 
Also this Prometheus pr https://github.com/prometheus/prometheus/pull/8050 has the same idea to improve the performance.

prmsrswt: Cool, thanks for the answer. :)

bwplotka (approved):
LGTM, thanks. It's very performant though, but we have TODO. Can we add issue as well? 🤗

PR #3574, opened 2020-12-11
bucket: Fix unbalanced chunkPool get/put

zecke:
This is a speculative change that came up when reading the code
while trying to install the pool. We should also move the check
before potentially allocating anything.

## Changes

It seems there is an error path that doesn't return the byte slice into the pool. This could lead to a pool being considered full and all Queries failing.

## Verification

TODO

### DIFF
--- a/pkg/store/bucket.go
+++ b/pkg/store/bucket.go
@@ -1463,13 +1463,13 @@ func (b *bucketBlock) readIndexRange(ctx context.Context, off, length int64) ([]
 }
 
 func (b *bucketBlock) readChunkRange(ctx context.Context, seq int, off, length int64) (*[]byte, error) {
+	if seq < 0 || seq >= len(b.chunkObjs) {
+		return nil, errors.Errorf("unknown segment file for index %d", seq)
+	}
+
 	c, err := b.chunkPool.Get(int(length))
 	if err != nil {
 		return nil, errors.Wrap(err, "allocate chunk bytes")
-	}
-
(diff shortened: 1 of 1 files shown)


pkg/store/bucket.go
```diff
@@ -1469,6 +1469,7 @@ func (b *bucketBlock) readChunkRange(ctx context.Context, seq int, off, length i
 	}
 
 	if seq < 0 || seq >= len(b.chunkObjs) {
+		b.chunkPool.Put(c)
 		return nil, errors.Errorf("unknown segment file for index %d", seq)
 	}
```
yeya24, line 1474: What about moving this if block to the start of the method? Then we don't need to allocate from the chunk pool.

zecke: Sure. It's proposed in the commit message. I wanted to highlight the imbalance first.

zecke: Updated.

GiedriusS (approved):
LGTM!

pracucci (approved):
LGTM, in the sense that code changes are correct. I think that if you hit an "unknown segment file for index" issue you have way bigger issues than allocation a buffer of size `length`, but that's another story.

GiedriusS:
CircleCI for some reason doesn't want to cooperate on this pull request - the cloning of the git repo gets interrupted randomly. I ran the tests locally and they pass. Merging with admin rights (:

PR #3577, opened 2020-12-12
store: Count number of sync success/failures

zecke:
These metrics can be used to detect how stale a store is and the
error rate of syncing.

## Changes

Expose metrics that allow to observe whether or not the block syncing of a store is working or not.

### DIFF
--- a/pkg/store/bucket.go
+++ b/pkg/store/bucket.go
@@ -120,6 +120,10 @@ type bucketStoreMetrics struct {
 
 	seriesFetchDuration   prometheus.Histogram
 	postingsFetchDuration prometheus.Histogram
+
+	blockSyncs              prometheus.Counter
+	blockSyncFailures       prometheus.Counter
+	blockSyncLastSuccessful prometheus.Gauge
 }
 
 func newBucketStoreMetrics(reg prometheus.Registerer) *bucketStoreMetrics {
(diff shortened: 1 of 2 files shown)


yeya24 (approved):
LGTM! I think these metrics make sense. Thanks for your contribution.
Would you mind adding the changelog?

pkg/store/bucket.go
```diff
@@ -237,6 +241,19 @@ func newBucketStoreMetrics(reg prometheus.Registerer) *bucketStoreMetrics {
 		Buckets: []float64{0.001, 0.01, 0.1, 0.3, 0.6, 1, 3, 6, 9, 20, 30, 60, 90, 120},
 	})
 
+	m.blockSyncs = promauto.With(reg).NewCounter(prometheus.CounterOpts{
+		Name: "thanos_bucket_store_block_syncs_successful_total",
+		Help: "Total number of successful block syncs",
```
OGKevin, line 246: All the other help texts finishes with a `.`. Lets be consistent and add it here as well? Same for the other help texts.

zecke: Updated.

yeya24 (commented):
@zecke Would you mind adding a changelog entry for the new metrics? Thank you!

prmsrswt (approved):
Looks fine to me!

pkg/store/bucket.go
```diff
@@ -237,6 +241,19 @@ func newBucketStoreMetrics(reg prometheus.Registerer) *bucketStoreMetrics {
 		Buckets: []float64{0.001, 0.01, 0.1, 0.3, 0.6, 1, 3, 6, 9, 20, 30, 60, 90, 120},
 	})
 
+	m.blockSyncs = promauto.With(reg).NewCounter(prometheus.CounterOpts{
```
pracucci, line 244: [non-blocking] I think it's more handy when we have a count of the total number of attempts (eg. `thanos_bucket_store_block_syncs_total`), regardless they succeeded or not, and then we can failures as you did in `blockSyncFailures`.

This way you can compute the failure rate as `thanos_bucket_store_block_syncs_failures_total  / thanos_bucket_store_block_syncs_total` instead of having to do `thanos_bucket_store_block_syncs_failures_total / (thanos_bucket_store_block_syncs_failures_total + thanos_bucket_store_block_syncs_successful_total)`

pkg/store/bucket.go
```diff
@@ -237,6 +241,19 @@ func newBucketStoreMetrics(reg prometheus.Registerer) *bucketStoreMetrics {
...
+		Name: "thanos_bucket_store_block_syncs_successful_total",
+		Help: "Total number of successful block syncs.",
+	})
+	m.blockSyncFailures = promauto.With(reg).NewCounter(prometheus.CounterOpts{
+		Name: "thanos_bucket_store_block_syncs_failures_total",
+		Help: "Total number of failed block syncs.",
+	})
+	m.blockSyncLastSuccessful = promauto.With(reg).NewGauge(prometheus.GaugeOpts{
+		Name: "thanos_bucket_store_block_last_successful_sync_time",
```
pracucci, line 253: [nit] I think adding `_seconds` at the end help to quickly understand the unit. It's something we did in `thanos_receive_config_last_reload_success_timestamp_seconds` too.

```suggestion
		Name: "thanos_bucket_store_block_last_successful_sync_timestamp_seconds",
```

pracucci (approved):
LGTM. I left a couple of non-blocking comments.

kakkoyun:
@zecke This is actually close to finish line, could you give it another push?

pkg/store/bucket.go
yeya24: I agree with @pracucci. WDYT @zecke? Would you mind making the change? Then we can merge.

zecke:
Please re-open

Issue #3408, opened 2020-11-04
compact: Introduce smart bucket caching for metadata with TTL; replace fetcher cache with it.


bwplotka:
Action Items:

0. Make sure it's programmatically configurable what files / dirs to cache.
1. Currently metadata is cached with two separate mechanisms: Fetcher internal code and cache bucket. Unify this.
2. For compact we don't use obj store level cache only fetcher one. Unify to bucket cache.
3. Add reset cache functionality so we can reset certain files e.g markers when we add them. Also, expose this API through HTTP.
4. Add fuzzing to TTL 
5. Add in-memory bucket caching as well

bwplotka:
4. Add Fuzzing to ttl

bwplotka:
Code to bucket caching: https://github.com/thanos-io/thanos/blob/843307aa13ae724fbafa4abdfb3c1106cc621fb0/pkg/store/cache/caching_bucket.go#L37

Fetcher dir and memory caching https://github.com/thanos-io/thanos/blob/2ad3805de0b208203a75b8511f095a9c00316222/pkg/block/fetcher.go#L149

Sudhar287:
Does it make sense to divide the tasks you mentioned above and send a PR for each one separately?

bwplotka:
Definitely (:

bwplotka:
> I need some clarifications here. Presently when we use baseFetcher, all the meta data is cached to either the run time memory or into a cacheDir Which I understand...
How would I possibly introduce a feature to the users which asks them what directories or files they want to cache?
Do that involves getting a set of block Ids from the user such as 01BKGV7JBM69T2G1BGBGM6KB12   01BKGTZQ1SYQJTR4PB43C8PD98

or both! (: 

No, what I mean rather is what KIND of files to cache where.

Currently, we cache the only meta.json per block in mem and optionally on disk (fetcher). Plus if you do[ bucket caching does its own thing](https://github.com/bwplotka/thanos/blob/843307aa13ae724fbafa4abdfb3c1106cc621fb0/pkg/store/cache/caching_bucket.go#L37). Currently caching bucket implementation is quite flexible and good, but I think we might want to improve configuring YAML and how it's used in our code.

For example:
* impl is in `pkg/store/cache` - Since we want to use it in fetcher for other components, we might want to move it to `pkg/objstore/cache` e.g?
[...]

PR #3579, opened 2020-12-12
in memory cache for caching bucket

Sudhar287:

## Changes

For #5 (Add in-memory bucket caching) in #3408 

## Verification

Wrote a new test case.

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -15,7 +15,6 @@ require (
 	github.com/cortexproject/cortex v1.6.1-0.20210108144208-6c2dab103f20
 	github.com/davecgh/go-spew v1.1.1
 	github.com/facette/natsort v0.0.0-20181210072756-2cd4dd1e2dcb
-	github.com/fatih/color v1.10.0 // indirect
 	github.com/fatih/structtag v1.1.0
 	github.com/felixge/fgprof v0.9.1
 	github.com/fsnotify/fsnotify v1.4.9
@@ -69,7 +68,6 @@ require (
 	gopkg.in/fsnotify.v1 v1.4.7
 	gopkg.in/yaml.v2 v2.3.0
 	gopkg.in/yaml.v3 v3.0.0-20200615113413-eeeca48fe776
-	honnef.co/go/tools v0.0.1-2020.1.6 // indirect
 )
 
 replace (
--- a/go.mod
+++ b/go.mod
@@ -15,6 +15,7 @@ require (
 	github.com/cortexproject/cortex v1.6.1-0.20210108144208-6c2dab103f20
 	github.com/davecgh/go-spew v1.1.1
 	github.com/facette/natsort v0.0.0-20181210072756-2cd4dd1e2dcb
+	github.com/fatih/color v1.10.0 // indirect
 	github.com/fatih/structtag v1.1.0
 	github.com/felixge/fgprof v0.9.1
 	github.com/fsnotify/fsnotify v1.4.9
@@ -68,6 +69,7 @@ require (
 	gopkg.in/fsnotify.v1 v1.4.7
 	gopkg.in/yaml.v2 v2.3.0
 	gopkg.in/yaml.v3 v3.0.0-20200615113413-eeeca48fe776
+	honnef.co/go/tools v0.0.1-2020.1.6 // indirect
 )
 
 replace (
--- a/pkg/cache/inmemory.go
+++ b/pkg/cache/inmemory.go
@@ -7,7 +7,6 @@ import (
 	"context"
 	"sync"
 	"time"
-	"unsafe"
 
 	"github.com/go-kit/kit/log"
 	"github.com/go-kit/kit/log/level"
@@ -28,10 +27,6 @@ var (
 
 const (
 	maxInt = int(^uint(0) >> 1)
-	// Size of a golang slice i.e unsafe.Sizeof([]byte{}) on amd64.
-	sliceHeaderSize = 16
-	// As the cache key is of type String its size is 8 bytes.
-	keyHeaderSize = 8
 )
 
 // InMemoryCacheConfig holds the in-memory cache config.
@@ -43,19 +38,19 @@ type InMemoryCacheConfig struct {
 }
 
 type InMemoryCache struct {
-	mtx sync.Mutex
-
 	logger           log.Logger
-	lru              *lru.LRU
 	maxSizeBytes     uint64
 	maxItemSizeBytes uint64
 
-	curSize uint64
-
-	evicted          prometheus.Counter
-	requests         prometheus.Counter
-	hits             prometheus.Counter
-	hitsExpired      prometheus.Counter
+	mtx         sync.Mutex
+	curSize     uint64
+	lru         *lru.LRU
+	evicted     prometheus.Counter
+	requests    prometheus.Counter
+	hits        prometheus.Counter
+	hitsExpired prometheus.Counter
+	// The input cache value would be copied to an inmemory array
+	// instead of simply using the one sent by the caller.
 	added            prometheus.Counter
 	current          prometheus.Gauge
 	currentSize      prometheus.Gauge
@@ -64,7 +59,12 @@ type InMemoryCache struct {
 }
 
 type cacheDataWithTTLWrapper struct {
-	data       []byte
+	data []byte
+	// The objects that are over the TTL are not destroyed eagerly.
+	// When there is a hit for an item that is over the TTL, the object is removed from the cache
+	// and null is returned.
+	// There is ongoing effort to integrate TTL within the Hashicorp golang cache itself.
+	// This https://github.com/hashicorp/golang-lru/pull/41 can be used here once complete.
 	expiryTime time.Time
 }
 
@@ -78,7 +78,7 @@ func parseInMemoryCacheConfig(conf []byte) (InMemoryCacheConfig, error) {
 	return config, nil
 }
 
-// NewInMemoryCache creates a new thread-safe LRU cache for chunks and metadata and ensures the total cache
+// NewInMemoryCache creates a new thread-safe LRU cache and ensures the total cache
 // size approximately does not exceed maxBytes.
 func NewInMemoryCache(name string, logger log.Logger, reg prometheus.Registerer, conf []byte) (*InMemoryCache, error) {
 	config, err := parseInMemoryCacheConfig(conf)
@@ -89,7 +89,7 @@ func NewInMemoryCache(name string, logger log.Logger, reg prometheus.Registerer,
 	return NewInMemoryCacheWithConfig(name, logger, reg, config)
 }
 
-// NewInMemoryCacheWithConfig creates a new thread-safe LRU cache chunks and metadata ensures the total cache
+// NewInMemoryCacheWithConfig creates a new thread-safe LRU cache and ensures the total cache
 // size approximately does not exceed maxBytes.
 func NewInMemoryCacheWithConfig(name string, logger log.Logger, reg prometheus.Registerer, config InMemoryCacheConfig) (*InMemoryCache, error) {
 	if config.MaxItemSize > config.MaxSize {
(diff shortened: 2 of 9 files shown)


go.mod
```diff
@@ -15,6 +15,7 @@ require (
 	github.com/cortexproject/cortex v1.5.1-0.20201111110551-ba512881b076
 	github.com/davecgh/go-spew v1.1.1
 	github.com/facette/natsort v0.0.0-20181210072756-2cd4dd1e2dcb
+	github.com/fatih/color v1.10.0 // indirect
```
prmsrswt, line 18: This seems unintentional.

go.mod
```diff
@@ -68,6 +69,7 @@ require (
 	gopkg.in/fsnotify.v1 v1.4.7
 	gopkg.in/yaml.v2 v2.3.0
 	gopkg.in/yaml.v3 v3.0.0-20200615113413-eeeca48fe776
+	honnef.co/go/tools v0.0.1-2020.1.6 // indirect
```
prmsrswt, line 72: This as well :)

prmsrswt (approved):
The code looks pretty solid. Maybe let's add a changelog entry for this. :)

Thanks!

pkg/cache/inmemory.go
```diff
@@ -0,0 +1,302 @@
...
+	}
+
+	c := &InMemoryCache{
+		logger:           logger,
+		maxSizeBytes:     uint64(config.MaxSize),
+		maxItemSizeBytes: uint64(config.MaxItemSize),
+	}
+
+	c.evicted = promauto.With(reg).NewCounter(prometheus.CounterOpts{
+		Name:        "thanos_store_cache_items_evicted_total",
```
prmsrswt, line 106: As we are planning to have this unify caching and use this in compactor as well, maybe using `"thanos_cache_items_evicted_total"` will make more sense.

pkg/cache/inmemory.go
```diff
@@ -0,0 +1,302 @@
...
+
+	c.current = promauto.With(reg).NewGauge(prometheus.GaugeOpts{
+		Name:        "thanos_store_cache_items",
+		Help:        "Current number of items in the inmemory cache.",
+		ConstLabels: prometheus.Labels{"name": name},
+	})
+
+	c.currentSize = promauto.With(reg).NewGauge(prometheus.GaugeOpts{
+		Name: "thanos_store_cache_items_size_bytes",
+		Help: "Current byte size of items in the inmemory cache.",
```
prmsrswt, line 149: I think you forgot to add `ConstLabels` here (and in the next ones too).

prmsrswt (commented):
A few more nits I forgot to add in the last review.

pkg/cache/inmemory.go
```diff
@@ -0,0 +1,302 @@
...
+
+	c := &InMemoryCache{
+		logger:           logger,
+		maxSizeBytes:     uint64(config.MaxSize),
+		maxItemSizeBytes: uint64(config.MaxItemSize),
+	}
+
+	c.evicted = promauto.With(reg).NewCounter(prometheus.CounterOpts{
+		Name:        "thanos_store_cache_items_evicted_total",
+		Help:        "Total number of items that were evicted from the inmemory cache.",
```
OGKevin, line 107: This help indicates that this metric is _only_ for inmemory. However, this is not the case for the metric name. If you have a look at the metrics defined here for exmaple: https://github.com/thanos-io/thanos/blob/b5e7b2b493783d89933874293e5c6a97a5b444b5/pkg/cache/memcached.go#L35-L40
You see that the name and help relfect the cache type.

OGKevin: Same for all the other metrics.

GiedriusS:
@Sudhar287 any update on this? :P

Sudhar287:
> @Sudhar287 any update on this? :P 

Waiting for @pstibrany comments before I send the next revision. Is that fine?

pstibrany (commented):
> Waiting for @pstibrany comments before I send the next revision. Is that fine?

Sorry it took me so long to take a look at this. I don't really have any other feedback that others haven't written already.

pkg/cache/inmemory.go
```diff
@@ -0,0 +1,302 @@
...
+		return nil, false
+	}
+	c.hits.Inc()
+	return v.(cacheDataWithTTLWrapper).data, true
+}
+
+func (c *InMemoryCache) set(key string, val []byte, ttl time.Duration) {
+	var size = sliceHeaderSize + uint64(len(val))
+	keySize := keyHeaderSize + uint64(len(key))
+	entrySize := sliceHeaderSize + uint64(len(val)) + uint64(unsafe.Sizeof(time.Now().Add(ttl)))
```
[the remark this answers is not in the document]
Sudhar287: I see.
Then I guess it makes sense to just have `size = len(val)`. The `key` is just a `String` anyway...
This would make it similar to: (without the `*HeaderSize`)
https://github.com/thanos-io/thanos/blob/b5e7b2b493783d89933874293e5c6a97a5b444b5/pkg/store/cache/inmemory.go#L207-L208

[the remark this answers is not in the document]
pstibrany: I wouldn't skip `key`. Eg. when caching attributes via caching bucket, key may be longer than value.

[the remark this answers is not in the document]
Sudhar287: Oh okay. I think the `key` size can be summed up like this:
https://github.com/thanos-io/thanos/blob/b5e7b2b493783d89933874293e5c6a97a5b444b5/pkg/store/cache/inmemory.go#L230

pkg/cache/inmemory.go
Sudhar287: Changed the prefix to `thanos_cache_inmemory_`

pkg/cache/inmemory.go
Sudhar287: nice catch!

pkg/cache/inmemory.go
Sudhar287: Okay. Understood.

pkg/cache/inmemory.go
```diff
@@ -0,0 +1,302 @@
...
+		return
+	}
+
+	if !c.ensureFits(size) {
+		c.overflow.Inc()
+		return
+	}
+
+	// The caller may be passing in a sub-slice of a huge array. Copy the data
+	// to ensure we don't waste huge amounts of space for something small.
```
pstibrany, line 237: Just a thought: Is this a responsibility of cache, or caller? Problem in doing it in cache is that cache must always do a copy, because it doesn't know if it's required or not. Caller can make that distinction however, which is why it may be preferable to not do it here.

Either way, I think this behaviour should be documented at `InMemoryCache` struct.

pkg/cache/inmemory.go
```diff
@@ -0,0 +1,302 @@
...
+	c.currentSize = promauto.With(reg).NewGauge(prometheus.GaugeOpts{
+		Name:        "thanos_cache_inmemory_items_size_bytes",
+		Help:        "Current byte size of items in the inmemory cache.",
+		ConstLabels: prometheus.Labels{"name": name},
+	})
+
+	c.totalCurrentSize = promauto.With(reg).NewGauge(prometheus.GaugeOpts{
+		Name: "thanos_cache_inmemory_total_size_bytes",
+		Help: "Current byte size of items (both value and key) in the inmemory cache.",
+	})
```
pstibrany, line 156: Do we need both?

pkg/cache/inmemory.go
```diff
@@ -0,0 +1,302 @@
...
+			"curSize", c.curSize,
+			"itemSize", size,
+		)
+		return false
+	}
+
+	for c.curSize+size > c.maxSizeBytes {
+		if _, _, ok := c.lru.RemoveOldest(); !ok {
+			level.Error(c.logger).Log(
+				"msg", "LRU has nothing more to evict, but we still cannot allocate the item. Resetting cache.",
```
pstibrany, line 266: If there is nothing more to evict, it means that cache is already empty. Why doing the extra reset in that case?

Once we get to this point, we will never break out of this for-loop.

pstibrany (commented):
Thank you for your work on this PR. There are still some small issues I'd like to understand better (const label, infinite for-loop?)

pkg/cache/inmemory.go
Sudhar287: I disagree that there is a infinite for loop. After `evict`ing all items and `reset`ing the cache, the condition `c.curSize+size > c.maxSizeBytes` will be `false` right? Cause `c.curSize == 0` and `size` cannot be greater than `maxSizeBytes` (line 95, 252). So there will no be an infinite loop right? 

Although I do get your point about the cache `reset` being unnecessary here. I looked into https://github.com/thanos-io/thanos/blob/1fff9a7157d1f3b7d5e0ce36f999b10bb3ae7ec4/pkg/store/cache/inmemory.go#L237 to write this function. Are you suggesting that even the `cache/inmemory.go#L237` is a bit incorrect too?

pstibrany: > I disagree that there is a infinite for loop. After `evict`ing all items and `reset`ing the cache, the condition `c.curSize+size > c.maxSizeBytes` will be `false` right? Cause `c.curSize == 0` and `size` cannot be greater than `maxSizeBytes` (line 95, 252). So there will no be an infinite loop right?

You are right. I missed the check on line 95, and thought it would loop if `size > c.maxSizeBytes`. (Btw, unit-test for this scenario would be worth adding).

pstibrany: > Are you suggesting that even the `cache/inmemory.go#L237` is a bit incorrect too?

Yes, I think `reset` there is not necessary, but harmless. (As in this case)

pkg/cache/inmemory.go
```diff
@@ -0,0 +1,302 @@
...
+
+type InMemoryCache struct {
+	mtx sync.Mutex
+
+	logger           log.Logger
+	lru              *lru.LRU
+	maxSizeBytes     uint64
+	maxItemSizeBytes uint64
+
+	curSize uint64
```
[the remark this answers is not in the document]
Sudhar287: Understood. 
PS: Along with `lru` and `curSize` it seems like `mtx` is also used for other counters too. Your suggestion makes sense.

pkg/cache/inmemory.go
Sudhar287: You are right, I feel this is the responsibility of the caller. But to keep things consistent with https://github.com/thanos-io/thanos/blob/1fff9a7157d1f3b7d5e0ce36f999b10bb3ae7ec4/pkg/store/cache/inmemory.go#L222-L223 I will leave a comment in the `InMemoryCache` struct.

pkg/cache/inmemory.go
Sudhar287: I feel we do. Similar to https://github.com/thanos-io/thanos/blob/1fff9a7157d1f3b7d5e0ce36f999b10bb3ae7ec4/pkg/store/cache/inmemory.go#L48-L49 it not a bad idea to keep track of total size irrespective of how small the `key` is. Please let me know if you have other thoughts.

pkg/cache/inmemory.go
```diff
@@ -0,0 +1,302 @@
...
+		Help: "Maximum number of bytes to be held in the inmemory cache.",
+	}, func() float64 {
+		return float64(c.maxSizeBytes)
+	})
+	_ = promauto.With(reg).NewGaugeFunc(prometheus.GaugeOpts{
+		Name: "thanos_cache_inmemory_max_item_size_bytes",
+		Help: "Maximum number of bytes for single entry to be held in the inmemory cache.",
+	}, func() float64 {
+		return float64(c.maxItemSizeBytes)
+	})
```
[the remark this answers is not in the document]
Sudhar287: Sorry missed it. Nice catch Peter!

pkg/cache/inmemory.go
Sudhar287: >  (Btw, unit-test for this scenario would be worth adding).

Got it. :+1:

pkg/cache/inmemory.go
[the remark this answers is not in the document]
pstibrany: Prometheus counters are safe to use without the mutex. (There is no external mutex involved when *reading* from them anyway.)

kakkoyun:
Hey @Sudhar287, has everything raised by @pstibrany been addressed? We'd appreciate @pstibrany if could have your blessing if it's the case?

Sudhar287:
Yes @kakkoyun, all of @pstibrany comments have been addressed.

kakkoyun (approved):
🥇

pkg/store/cache/inmemory_test.go
```diff
@@ -44,6 +44,20 @@ max_item_size: 2KB
...
+max_item_size: 1MB
+`)
+	cache, err = NewInMemoryIndexCache(log.NewNopLogger(), nil, conf)
+	testutil.NotOk(t, err)
+	testutil.Equals(t, (*InMemoryIndexCache)(nil), cache)
+	// testutil.Equals(t, uint64(1024*1024), cache.maxSizeBytes)
+	// testutil.Equals(t, uint64(2*1024), cache.maxItemSizeBytes)
+
+	// testutil.Equals(t, uint64(1024*1024), cache.maxItemSizeBytes)
+	// testutil.Equals(t, uint64(2*1024), cache.maxSizeBytes)
```
pstibrany, line 60: Commented code should be removed.

pstibrany (commented):
LGTM, thanks, and apologies for super-slow review cycle from my side. :-(

yashrsharma44:
Mission Accomplished 😎 -

pkg/store/cache/inmemory_test.go
kakkoyun: Ahh, I've missed that. @Sudhar287 Would you mind creating another PR to clean this up? If not, I can also handle it.

Issue #3339, opened 2020-10-19
Problem with chunk-pool-size and store panics in pkg/store.(*bucketChunkReader).loadChunks


sevagh:
Hello - since upgrading from Thanos 0.14.0 to 0.16.0-rc0, I started having my Store component fail to load blocks because of https://github.com/thanos-io/thanos/issues/1179

I adjusted chunk-pool-size up from 2GB to 8GB, 16GB, and now 50GB. On 50GB, I get a panic and the store crashes:
```
Oct 19 07:35:45 sv3-thanos4 thanos[47253]: panic: runtime error: slice bounds out of range [23301:14480]
Oct 19 07:35:45 sv3-thanos4 thanos[47253]: goroutine 2092987 [running]:
Oct 19 07:35:45 sv3-thanos4 thanos[47253]: github.com/thanos-io/thanos/pkg/store.(*bucketChunkReader).loadChunks(0xdffd9ddc20, 0x1f7c1c0, 0xe1eaa29040, 0xe1eaa0bca0, 0x5, 0x8, 0x6, 0xde5b01a0de402fe, 0x0, 0x0)
Oct 19 07:35:45 sv3-thanos4 thanos[47253]:         thanos/pkg/store/bucket.go:2099 +0xb87
Oct 19 07:35:45 sv3-thanos4 thanos[47253]: github.com/thanos-io/thanos/pkg/store.(*bucketChunkReader).preload.func3(0x0, 0x1ce13c0)
Oct 19 07:35:45 sv3-thanos4 thanos[47253]:         thanos/pkg/store/bucket.go:2066 +0x9f
Oct 19 07:35:45 sv3-thanos4 thanos[47253]: golang.org/x/sync/errgroup.(*Group).Go.func1(0xe1eaa25980, 0xe1eaa2f800)
Oct 19 07:35:45 sv3-thanos4 thanos[47253]:         go/pkg/mod/golang.org/x/sync@v0.0.0-20200625203802-6e8e738ad208/errgroup/errgroup.go:57 +0x59
Oct 19 07:35:45 sv3-thanos4 thanos[47253]: created by golang.org/x/sync/errgroup.(*Group).Go
Oct 19 07:35:45 sv3-thanos4 thanos[47253]:         go/pkg/mod/golang.org/x/sync@v0.0.0-20200625203802-6e8e738ad208/errgroup/errgroup
```

[...]

PR #3621, opened 2020-12-13
store: Be pessimistic and check how many bytes got copied

zecke:
I have seen a panic in loadChunks and the first thing that came to mind
that the getRange query might not give the result we would like to have.
Check the right number of bytes were copied into the destination.

## Verification

Not yet. Soliciting feedback of how defensive we want to be.

### DIFF
--- a/pkg/store/bucket.go
+++ b/pkg/store/bucket.go
@@ -1472,9 +1472,14 @@ func (b *bucketBlock) readChunkRange(ctx context.Context, seq int, off, length i
 	}
 	defer runutil.CloseWithLogOnErr(b.logger, r, "readChunkRange close range reader")
 
-	if _, err = io.Copy(buf, r); err != nil {
+	copied, err := io.Copy(buf, r)
+	if err != nil {
 		b.chunkPool.Put(c)
 		return nil, errors.Wrap(err, "read range")
+	}
+	if copied != length {
+		b.chunkPool.Put(c)
+		return nil, errors.Errorf("Read less than specified wanted=%v got=%v", length, copied)
 	}
 	internalBuf := buf.Bytes()
 	return &internalBuf, nil
--- a/pkg/store/bucket.go
+++ b/pkg/store/bucket.go
@@ -1479,7 +1479,7 @@ func (b *bucketBlock) readChunkRange(ctx context.Context, seq int, off, length i
 	}
 	if copied != length {
 		b.chunkPool.Put(c)
-		return nil, errors.Errorf("Read less than specified wanted=%v got=%v", length, copied)
+		return nil, errors.Errorf("read %v bytes but got %v", length, copied)
 	}
 	internalBuf := buf.Bytes()
 	return &internalBuf, nil


pracucci:
Do you have a trace of the panic you're experiencing? I would be glad to learn more about the issue you're experiencing.

zecke:
> Do you have a trace of the panic you're experiencing? I would be glad to learn more about the issue you're experiencing.

The panic trace looks like the one here: https://github.com/thanos-io/thanos/issues/3339

I added a defer + recover in there so that a single failure doesn't crash the store and will add enough debug information to find the block. My two cents are I would like this part of the code be defensive:

* Don't crash and take the process down (is it a corrupted block? is it something funny the object store in use)
* Be defensive. E.g. io.CopyN and check that we have copied that many bytes..

pkg/store/bucket.go
```diff
@@ -1472,10 +1472,15 @@ func (b *bucketBlock) readChunkRange(ctx context.Context, seq int, off, length i
...
 
-	if _, err = io.Copy(buf, r); err != nil {
+	copied, err := io.Copy(buf, r)
+	if err != nil {
 		b.chunkPool.Put(c)
 		return nil, errors.Wrap(err, "read range")
 	}
+	if copied != len {
+		b.chunkPool.Put(c)
+		return nil, errors.Errorf("Read less than specified wanted=%v got=%v", len, copied)
```
GiedriusS, line 1482: We should begin it with a lowercase letter. Also, I think this improves the formatting:

```suggestion
		return nil, errors.Errorf("read %v bytes but got %v", len, copied)
```

GiedriusS (commented):
I think this change is reasonable. One minor nit on the error message. Care to update this PR to fix the build errors?

pkg/store/bucket.go
```diff
@@ -1472,10 +1472,15 @@ func (b *bucketBlock) readChunkRange(ctx context.Context, seq int, off, length i
 	}
 	defer runutil.CloseWithLogOnErr(b.logger, r, "readChunkRange close range reader")
 
-	if _, err = io.Copy(buf, r); err != nil {
+	copied, err := io.Copy(buf, r)
+	if err != nil {
 		b.chunkPool.Put(c)
 		return nil, errors.Wrap(err, "read range")
 	}
+	if copied != len {
```
pracucci, line 1480: I guess you meant `if copied != length`

zecke: doh. Thanks linter. ;)

zecke:
The assumption I made about the code and reality do not align yet. My assumption was that we precisely  know where the data we want is located and how large it is. Let me run this locally to understand where the gap is. Any ideas/pointers welcome.

```
level=error ts=2020-12-22T11:20:17.1040842Z msg="function failed. Retrying in next tick" err="unexpected warnings [receive series from Addr: e2e_test_compact-store-gw-1:9091 LabelSets: {case=\"a-partial-overlap-dedup-ready\", replica=\"1\"},{case=\"a-partial-overlap-dedup-ready\", replica=\"2\"},{case=\"a-partial-overlap-dedup-ready\", replica=\"3\"},{case=\"compaction-ready\", replica=\"1\"},{case=\"compaction-ready-after-dedup\", replica=\"1\"},{case=\"compaction-ready-after-dedup\", replica=\"2\"},{case=\"compaction-ready-one-block-marked-for-no-compact\", replica=\"1\"},{case=\"full-replica-overlap-dedup-ready\", replica=\"1\"},{case=\"full-replica-overlap-dedup-ready\", replica=\"2\"},{case=\"no-compaction\", replica=\"1\"},{case=\"partial-multi-replica-overlap-dedup-ready\", replica=\"1\", rule_replica=\"1\"},{case=\"partial-multi-replica-overlap-dedup-ready\", replica=\"1\", rule_replica=\"2\"} Mint: 1585036800000 Maxt: 1585072800000: rpc error: code = Aborted desc = fetch series for block 01ET510R8RSYN56Z5THSV4V2SP: preload chunks: read range for 0: read 17798 bytes but got 1839]"
```

PR #3626, opened 2020-12-15
Blocks upload: Don't do cleanUp when meta.json fails to upload.

pstibrany:
Under certain conditions, when upload of `meta.json` file fails, it will actually appear on the bucket eventually. S3 has been observed to behave like this when it reports 503 "SlowDown" error. Currently `block.Upload` calls `cleanUp` to delete partially uploaded block, but if `meta.json` can still appear in the bucket later, this actually creates invalid block – with `meta.json` file only, but no index or chunks.

## Changes

This PR removes call to `cleanUp` if only `meta.json` upload fails. If `meta.json` has really failed, this will result into partial block (without `meta.json`) and be eventually removed by other components, or upload will be retried by Shipper. But at least it won't generate corrupted blocks.

For more details, see discussion at https://cloud-native.slack.com/archives/CL25937SP/p1608021430416900.

## Verification

Added unit test.

### DIFF
--- a/pkg/block/block.go
+++ b/pkg/block/block.go
@@ -120,6 +120,7 @@ func Upload(ctx context.Context, logger log.Logger, bkt objstore.Bucket, bdir st
 		// Don't call cleanUp here. Despite getting error, meta.json may have been uploaded in certain cases,
 		// and even though cleanUp will not see it yet, meta.json may appear in the bucket later.
 		// (Eg. S3 is known to behave this way when it returns 503 "SlowDown" error).
+		// If meta.json is not uploaded, this will produce partial blocks, but such blocks will be cleaned later.
 		return errors.Wrap(err, "upload meta file")
 	}
 
--- a/pkg/block/block.go
+++ b/pkg/block/block.go
@@ -117,7 +117,10 @@ func Upload(ctx context.Context, logger log.Logger, bkt objstore.Bucket, bdir st
 
 	// Meta.json always need to be uploaded as a last item. This will allow to assume block directories without meta file to be pending uploads.
 	if err := bkt.Upload(ctx, path.Join(id.String(), MetaFilename), &metaEncoded); err != nil {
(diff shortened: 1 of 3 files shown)


pstibrany:
Failed `TestTSDBStore_Series_SplitSamplesIntoChunksWithMaxSizeOf120` test seem to be unrelated to this change.

pkg/block/block.go
```diff
@@ -117,7 +117,10 @@ func Upload(ctx context.Context, logger log.Logger, bkt objstore.Bucket, bdir st
 
 	// Meta.json always need to be uploaded as a last item. This will allow to assume block directories without meta file to be pending uploads.
 	if err := bkt.Upload(ctx, path.Join(id.String(), MetaFilename), &metaEncoded); err != nil {
-		return cleanUp(logger, bkt, id, errors.Wrap(err, "upload meta file"))
+		// Don't call cleanUp here. Despite getting error, meta.json may have been uploaded in certain cases,
+		// and even though cleanUp will not see it yet, meta.json may appear in the bucket later.
+		// (Eg. S3 is known to behave this way when it returns 503 "SlowDown" error).
```
OGKevin, line 122: Does it make sense to also include here that other components will take care of it?

OGKevin (approved):

yeya24 (approved):
Thanks! LGTM

prmsrswt (approved):
Looks good to me as well. :+1:

GiedriusS:
This is a very interesting case. Thank you for your work on this! AIUI pragmatically this means that the meta.json file would have to appear within 30 seconds because we do a `Sync` every 30 seconds. I guess this lines up with AWS S3 restrictions - namely that the limitation is 3500 requests per second. So, we still have 29 seconds left. Is my understanding correct? Either way, merging this (:

GiedriusS (approved):

PR #3632, opened 2020-12-17
Sidecar: Used ExtendedSortedLabels instead of Sort 

Biswajitghosh98:
As of now, I'm just creating a draft PR because a few things are unclear to me : <br>
1. I have assumed `labelpb.ZLabelsFromPromLabels(extLset)...` to return a sorted list, which would definitely need fixing.
2. I believe unless the size of extLset is very large, we won't observe a significant difference with this PR, as asymptotically `O(nlogn + m)` would be very close to `O((n+m)log(n+m))` unless m>>n.

WDYT @bwplotka @yeya24 ? 
 

### DIFF
--- a/pkg/store/prometheus.go
+++ b/pkg/store/prometheus.go
@@ -160,15 +160,17 @@ func (p *PrometheusStore) Series(r *storepb.SeriesRequest, s storepb.Store_Serie
 			return err
 		}
 		for _, lbm := range labelMaps {
-			lset := make([]labelpb.ZLabel, 0, len(lbm)+len(extLset))
-			for k, v := range lbm {
-				lset = append(lset, labelpb.ZLabel{Name: k, Value: v})
+			lset := make([]labelpb.ZLabel, 0, len(lbm))
+			keys := make([]string, 0)
+			for k, _ := range lbm {
+				keys = append(keys,k)
 			}
-			lset = append(lset, labelpb.ZLabelsFromPromLabels(extLset)...)
-			sort.Slice(lset, func(i, j int) bool {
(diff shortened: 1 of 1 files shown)


pkg/store/prometheus.go
```diff
@@ -160,15 +160,17 @@ func (p *PrometheusStore) Series(r *storepb.SeriesRequest, s storepb.Store_Serie
 			return err
 		}
 		for _, lbm := range labelMaps {
-			lset := make([]labelpb.ZLabel, 0, len(lbm)+len(extLset))
-			for k, v := range lbm {
-				lset = append(lset, labelpb.ZLabel{Name: k, Value: v})
+			lset := make([]labelpb.ZLabel, 0, len(lbm))
```
bwplotka, line 163: Instead of trying to optimze map sorting + injection, why not ensuring we don't have unsorted map in the first place?

I think we can change this:

`labelMaps, err := p.client.SeriesInGRPC(s.Context(), p.base, newMatchers, r.MinTime, r.MaxTime)`

To return us a sorted list of labels. And we can do that because this call is making a HTTP request to /api/v1/series on Prometheus which by design, it returns sorted series (!).

the actual work to do is in json parser. That JSON Map sorted, but Golang parse that as unsorted map - so we need enhance parser (Unmarhsaller) to parse it as sorted labels.Labels (:

bwplotka (commented):
Some suggestion in our Thanos Contributor Hours  🤗

kakkoyun:
@Biswajitghosh98 Let's finish this one as well.

PR #3640, opened 2020-12-18
Handle external labels in Store's labels API

### LINKED ISSUE #3639: Inconsistency of the behaviors of handling labelValues requests and external labels
We are using different ways to handle `LabelValues` requests for each store.

For `PrometheusStore` which is the sidecar, it checks whether the label we want exists in external labels, if so just return the value. 

``` Go
func (p *PrometheusStore) LabelValues(ctx context.Context, r *storepb.LabelValuesRequest) (*storepb.LabelValuesResponse, error) {
	externalLset := p.externalLabelsFn()

	// First check for matching external label which has priority.
	if l := externalLset.Get(r.Label); l != "" {
		return &storepb.LabelValuesResponse{Values: []string{l}}, nil
	}

	vals, err := p.client.LabelValuesInGRPC(ctx, p.base, r.Label, nil, r.Start, r.End)
```

For `BucketStore`, we get the label values from the block's index and attach the external label value if it matches.

``` Go
			// Do it via index reader to have pending reader registered correctly.
			res, err := indexr.block.indexHeaderReader.LabelValues(req.Label)
			if err != nil {
				return errors.Wrap(err, "index header label values")
			}

			// Add the external label value as well.
			if extLabelValue, ok := extLabels[req.Label]; ok {
				res = strutil.MergeSlices(res, []string{extLabelValue})
			}

```

For `TSDBStore`, we just ignore the external labels. https://github.com/thanos-io/thanos/blob/master/pkg/store/tsdb.go#L221-L235.

I think we need to define the correct logic of the external labels handling for Labels API to make the whole thing more clear. As we plan to support matchers https://github.com/thanos-io/thanos/issues/3637 so it would be more complex.

yeya24:

Fixes #3639

This pr unifies the behavior of handling external labels on labels API for different stores.

### DIFF
--- a/pkg/store/tsdb.go
+++ b/pkg/store/tsdb.go
@@ -5,6 +5,7 @@ package store
 
 import (
 	"context"
+	"github.com/thanos-io/thanos/pkg/strutil"
 	"io"
 	"math"
 
@@ -214,6 +215,12 @@ func (s *TSDBStore) LabelNames(ctx context.Context, r *storepb.LabelNamesRequest
 	if err != nil {
 		return nil, status.Error(codes.Internal, err.Error())
 	}
+
+	extLabelNames := make([]string, 0, len(s.extLset))
+	for _, lbl := range s.extLset {
+		extLabelNames = append(extLabelNames, lbl.Name)
+	}
+	res = strutil.MergeSlices(res, extLabelNames)
 	return &storepb.LabelNamesResponse{Names: res}, nil
 }
 
@@ -231,5 +238,11 @@ func (s *TSDBStore) LabelValues(ctx context.Context, r *storepb.LabelValuesReque
 	if err != nil {
 		return nil, status.Error(codes.Internal, err.Error())
 	}
+
+	// Add the external label value as well.
+	if extLabelValue := s.extLset.Get(r.Label); extLabelValue != "" {
+		res = strutil.MergeSlices(res, []string{extLabelValue})
+	}
+
 	return &storepb.LabelValuesResponse{Values: res}, nil
 }
--- a/pkg/store/tsdb.go
+++ b/pkg/store/tsdb.go
@@ -5,7 +5,6 @@ package store
 
 import (
 	"context"
-	"github.com/thanos-io/thanos/pkg/strutil"
 	"io"
 	"math"
 
@@ -22,6 +21,7 @@ import (
 	"github.com/thanos-io/thanos/pkg/component"
 	"github.com/thanos-io/thanos/pkg/runutil"
 	"github.com/thanos-io/thanos/pkg/store/storepb"
+	"github.com/thanos-io/thanos/pkg/strutil"
 )
 
 const RemoteReadFrameLimit = 1048576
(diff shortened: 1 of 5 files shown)


yeya24:
For TSDB store and Prometheus store, the workflow is:
1. If the request time range is outside the store's time range, then return an empty array.
2. Else, return the labels result + the external labels names or external label values

For Block store, external labels are per-block so the current logic is fine.

For Proxy store like multitsdb and querier, its external labels will not be added to the results like other stores. This is also an inconsistency but I think it is fine. If anyone has other ideas about this, please let me know.

test/e2e/query_test.go
```diff
@@ -289,12 +289,12 @@ func TestQueryLabelNames(t *testing.T) {
...
 	// Outside time range.
 	labelNames(t, ctx, q.HTTPEndpoint(), nil, timestamp.FromTime(now.Add(-24*time.Hour)), timestamp.FromTime(now.Add(-23*time.Hour)), func(res []string) bool {
-		return len(res) == 0
+		return len(res) == 2
 	})
 
 	labelNames(t, ctx, q.HTTPEndpoint(), []storepb.LabelMatcher{{Type: storepb.LabelMatcher_EQ, Name: "__name__", Value: "up"}},
 		timestamp.FromTime(now.Add(-time.Hour)), timestamp.FromTime(now.Add(time.Hour)), func(res []string) bool {
-			// Expected result: [__name__, instance, job, prometheus, replica]
+			// Expected result: [__name__, instance, job, prometheus, receive, replica, tenant_id]
```
bwplotka, line 297: I like how you updated the commented code to make it work if someone uncomments 😄 😄 
Can we delete this or actually uncomment? 🤗

pkg/store/tsdb.go
```diff
@@ -214,13 +220,24 @@ func (s *TSDBStore) LabelNames(ctx context.Context, r *storepb.LabelNamesRequest
...
+	}
+	res = strutil.MergeSlices(res, extLabelNames)
 	return &storepb.LabelNamesResponse{Names: res}, nil
 }
 
 // LabelValues returns all known label values for a given label name.
 func (s *TSDBStore) LabelValues(ctx context.Context, r *storepb.LabelValuesRequest) (
 	*storepb.LabelValuesResponse, error,
 ) {
+	withinTimeRange, err := s.checkRequestTimeRange(r.End)
```
bwplotka, line 236: This is out of scope for this PR? Since it's small it's ok, let's just add this to PR tittle 🤗

bwplotka (approved):
LGTM 👍🏽 Thanks!

Can you help by adding changelog items for this? 🤗

test/e2e/query_test.go
yeya24: Is it a problem if someone uncomments this 🤔  The E2E test cannot be compiled so we can find the change.

kakkoyun:
@yeya24 Do you want to add anything to this? Or address any comments? We can merge it otherwise.

yeya24:
Sorry for the late reply. Will get back to it this week.

bwplotka:
Friendly ping (:

kakkoyun:
@yeya24 Shall we revive this one?

Issue #3569, opened 2020-12-10
DOC: Website contains some 404


MalloZup:
## Description:

I found two minor issue when navigating the webpage of thanos.

1)  when the user click following link:
![image](https://user-images.githubusercontent.com/10886597/101835535-38166380-3b3c-11eb-8ec3-07fecbbbec35.png)
link is https://github.com/thanos-io/thanos/edit/master/docs/thanos/integrations.md

it redirects actually to a 404. ( Maybe because i'm not maintainer)

2) it seems tthat https://thanos.io/tip/components/ was before https://thanos.io/tip/thanos/components/ 
but some pages are still referencing the  latter

Lot of links are broken in this page: 
https://thanos.io/tip/thanos/integrations.md/#integrations

E.g by https://thanos.io/tip/thanos/components/query.md
https://thanos.io/tip/thanos/components/sidecar.md

hope it helps :sunflower:

DARK-art108:
Can I work on this issue?

MalloZup:
@DARK-art108 woyld be great ! thx for your help

bboreham:
Link to "Querier/Query" from https://thanos.io/tip/thanos/quick-tutorial.md/ also 404s.

Namanl2001:
Is anyone working on this issue or could I take this up?

MalloZup:
@Namanl2001  feel free to take this issue

PR #3645, opened 2020-12-18
website: fixed broken links

Namanl2001:
Fixes [#3569](https://github.com/thanos-io/thanos/issues/3569) 

## Changes

Made changes in some links of quick-tutorial.md and integragions.md file, to point them to the correct files.

## Verification

I tested the website on localhost and the links now direct to the desired pages.

### DIFF
--- a/docs/integrations.md
+++ b/docs/integrations.md
@@ -9,9 +9,9 @@ menu: thanos
 ## StoreAPI
 
 [StoreAPI](https://github.com/thanos-io/thanos/blob/master/pkg/store/storepb/rpc.proto) is a common proto interface for gRPC component
-that can connect to [Querier](components/query.md) in order to fetch the metric series.
-Natively Thanos implements [Sidecar](components/sidecar.md) (local Prometheus data),
-[Ruler](components/rule.md) and [Store gateway](components/store.md).
+that can connect to [Querier](../components/query.md) in order to fetch the metric series.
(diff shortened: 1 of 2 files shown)


prmsrswt:
I guess this works because, on the `thanos.io` website, we have all top-level markdown files under the `/tip/thanos/` prefix.

So the directory structure is like
```
|-getting-started.md
|-integrations.md
|-components/
  |-query.md
  |-store.md
  |-...
|-... 
``` 
but the website has a structure like
```
|-thanos/
  |-getting-started.md
  |-integrations.md
  |-...
|-components/
  |-query.md
  |-store.md
  |-...
|-... 
```

So with your current approach, the links won't work in case we access them from GitHub UI. Also, there is a chance that other contributors may forget this quirk and won't remember to add `../` before the link paths on the top level.

A better approach in my opinion is to automate this during build using Hugo. This way, the links will work in both the generated website and the GitHub UI, plus less chance of human errors.

@thisisobate any thoughts?

thisisobate:
@prmsrswt  You're right. 

We are currently trying to solve the issue of broken links by creating Mdox- a much better tool that automatically fixes markdown link and of course, support versioning of docs out of the box.

You can check out [Mdox here](https://github.com/bwplotka/mdox)

bwplotka (approved):

bwplotka:
Fine as short term workaround. (:

PR #3647, opened 2020-12-18
UI: Handle routing on the client side for the React UI

ajinkya359:
Signed-off-by: ajinkya359 <sajinkya359@gmail.com>
FIxed issue #3216 

## Changes

Added component  NotFound
Removed variable reactAppPaths


pkg/ui/react-app/src/thanos/pages/notFound/NotFound.tsx
```diff
@@ -0,0 +1,9 @@
+import React, { FC } from 'react';
+
+interface NotFoundProps {
+  default: boolean;
```
prmsrswt, line 4: This is not used anywhere.

pkg/ui/react-app/src/thanos/pages/notFound/NotFound.tsx
```diff
@@ -0,0 +1,9 @@
...
+}
+
+const NotFound: FC<NotFoundProps> = () => <h3>404, Page Not Found</h3>;
```
prmsrswt, line 7: It'll be better to make this a bit more clean/visually pleasing. We have some container components in `reactstrap` (like `Contatiner` etc.). Try doing something like
```jsx
<Container fluid>
  <h3>404, Page Not Found</h3>
</Container>
```
to get the `<h3>` centered. Maybe we can also add a link here saying `"Return to homepage"` or simply `"Go back"`.

pkg/ui/ui.go
```diff
@@ -58,20 +39,16 @@ func NewBaseUI(logger log.Logger, menuTmpl string, funcMap template.FuncMap, tmp
 func (bu *BaseUI) serveStaticAsset(w http.ResponseWriter, req *http.Request) {
 	fp := route.Param(req.Context(), "filepath")
 	fp = filepath.Join("pkg/ui/static", fp)
-	bu.serveAsset(fp, w, req)
+	_ = bu.serveAsset(fp, w, req)
```
prmsrswt, line 42: Don't ignore errors. If error happenes here, that means we don't have the file locally, but now we are not returning 404 either, this will result in no response to the client.

```suggestion
	if err := bu.serveAsset(fp, w, req); err != nil {
		level.Warn(bu.logger).Log("msg", "Could not get file", "err", err, "file", fp)
		w.WriteHeader(http.StatusNotFound)
	}
```

pkg/ui/ui.go
```diff
@@ -58,20 +39,16 @@ func NewBaseUI(logger log.Logger, menuTmpl string, funcMap template.FuncMap, tmp
...
 func (bu *BaseUI) serveReactUI(w http.ResponseWriter, req *http.Request) {
 	fp := route.Param(req.Context(), "filepath")
-	for _, rp := range reactAppPaths {
-		if fp != rp {
-			continue
-		}
+	fp = filepath.Join("pkg/ui/static/react/", fp)
+	if err := bu.serveAsset(fp, w, req); err != nil {
 		bu.serveReactIndex("pkg/ui/static/react/index.html", w, req)
 		return
```
prmsrswt, line 50: This early `return` is unnecessary now, as it's going to return anyway.

pkg/ui/ui.go
```diff
@@ -108,14 +85,14 @@ func (bu *BaseUI) getAssetFile(filename string) (os.FileInfo, []byte, error) {
 	return info, file, nil
 }
 
-func (bu *BaseUI) serveAsset(fp string, w http.ResponseWriter, req *http.Request) {
+func (bu *BaseUI) serveAsset(fp string, w http.ResponseWriter, req *http.Request) error {
 	info, file, err := bu.getAssetFile(fp)
 	if err != nil {
 		level.Warn(bu.logger).Log("msg", "Could not get file", "err", err, "file", fp)
-		w.WriteHeader(http.StatusNotFound)
```
prmsrswt, line 115: Move these two lines in the error check after line 42 as mentioned in the previous comment.

prmsrswt (commented):
Good work overall. Just a few comments. Thanks for picking this up.

pkg/ui/ui.go
ajinkya359: Thanks for your review , i made the changes you suggested.

pkg/ui/react-app/src/App.tsx
```diff
@@ -31,7 +31,6 @@ const App: FC<PathPrefixProps & ThanosComponentProps> = ({ pathPrefix, thanosCom
         <QueryParamProvider reachHistory={globalHistory}>
           <Router basepath={`${pathPrefix}/new`}>
             <Redirect from="/" to={`${pathPrefix}/new${defaultRouteConfig[thanosComponent]}`} />
-
             {/*
               NOTE: Any route added here needs to also be added to the list of
```
prmsrswt, line 35: This comment is not true now, let's remove it completely.

pkg/ui/react-app/src/thanos/pages/notFound/NotFound.tsx
```diff
@@ -0,0 +1,27 @@
...
+
+const NotFound: FC<NotFoundProps & PathPrefixProps> = ({ pathPrefix, defaultRoute }) => {
+  return (
+    <div
+      style={{
+        position: 'absolute',
+        left: '50%',
+        top: '50%',
+        transform: 'translate(-50%, -50%)',
+      }}
```
prmsrswt, line 17: `position: absolute` must not be used unless it's actually needed, It can cause a lot of problems. Try using
```css
width: 100%;
min-height: 100vh;
display: flex;
flex-direction: column;
justify-content: center;
align-items: center;
```

Second point, we use CSS modules for styling so don't use inline CSS unless it's necessary.

pkg/ui/ui.go
```diff
@@ -58,20 +39,19 @@ func NewBaseUI(logger log.Logger, menuTmpl string, funcMap template.FuncMap, tmp
 func (bu *BaseUI) serveStaticAsset(w http.ResponseWriter, req *http.Request) {
 	fp := route.Param(req.Context(), "filepath")
 	fp = filepath.Join("pkg/ui/static", fp)
-	bu.serveAsset(fp, w, req)
+	if err := bu.serveAsset(fp, w, req); err != nil {
+		level.Warn(bu.logger).Log("msg", "Could not get file", "err", err, "file", fp)
+		w.WriteHeader(http.StatusNotFound)
+	}
```
prmsrswt, line 45: :+1:

prmsrswt (commented):
Just a few more suggestions.

pkg/ui/ui.go
ajinkya359: I made all the suggested changes, is there anything else i need to change?

prmsrswt (approved):
One small change, everything else looks good to me!

pkg/ui/react-app/.gitignore
```diff
@@ -1,23 +1,116 @@
...
-# testing
-/coverage
+# Runtime data
+pids
+*.pid
+*.seed
+*.pid.lock
 
-# production
-/build
```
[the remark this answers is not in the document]
ajinkya359: Ok , i added that in the last commit.

PR #3648, opened 2020-12-18
comment amendment in dns/provider.go

### LINKED ISSUE #3187: Incorrect comment in DNS provider
This line is the only occurrence of `defaultPort` in this repo:
https://github.com/thanos-io/thanos/blob/00510b5ae7fbcfc42c735532069afead30c3bb01/pkg/discovery/dns/provider.go#L111

I believe what actually happens is it returns an error.

Namanl2001:
Signed-off-by: Namanl2001 <namanlakhwani@gmail.com>

fixes #3187

## Changes

Amended the comment.

## Verification

Verified from the official docs which states that it returns error for non-SRV prefix, when a port is not supplied.
While for SRV prefixes there is no need to specify a port, as the one from the SRV query results will be used.

### DIFF
--- a/pkg/discovery/dns/provider.go
+++ b/pkg/discovery/dns/provider.go
@@ -108,7 +108,7 @@ func GetQTypeName(addr string) (qtype string, name string) {
 
 // Resolve stores a list of provided addresses or their DNS records if requested.
 // Addresses prefixed with `dns+` or `dnssrv+` will be resolved through respective DNS lookup (A/AAAA or SRV).
-// defaultPort is used for non-SRV records when a port is not supplied.
+// For non-SRV records, it will return an error if a port is not supplied.
 func (p *Provider) Resolve(ctx context.Context, addrs []string) error {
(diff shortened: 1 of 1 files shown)


pkg/discovery/dns/provider.go
```diff
@@ -108,7 +108,7 @@ func GetQTypeName(addr string) (qtype string, name string) {
 
 // Resolve stores a list of provided addresses or their DNS records if requested.
 // Addresses prefixed with `dns+` or `dnssrv+` will be resolved through respective DNS lookup (A/AAAA or SRV).
-// defaultPort is used for non-SRV records when a port is not supplied.
+// returns error for non-SRV records when a port is not supplied.
```
prmsrswt, line 111:
```suggestion
// For non-SRV records, it will return an error if a port is not supplied.
```

prmsrswt (approved):
Just a small nit, for better wording. I also verified the code and this amendment seems correct.

PR #3649, opened 2020-12-19
Add a point of suggestion for Mentorship Doc

yashrsharma44:
While I was part of the Community Bridge Mentorship, I found that working with an iterative process is really beneficial. Also having a small but working example helps in building up the project. I thought of adding this in the `mentorship` docs.

Suggestions are welcome :hugs:


yashrsharma44:
large diff due to linting of markdown

bwplotka:
Where is that point? Because all is changed =D

yashrsharma44:
Last two points 😂

bwplotka (approved):
Perfect. Thanks!

PR #3650, opened 2020-12-19
store: Cleaned up API for test/benchmark purposes.

bwplotka:

### DIFF
--- a/pkg/api/query/v1.go
+++ b/pkg/api/query/v1.go
@@ -350,7 +350,7 @@ func (qapi *QueryAPI) queryRange(r *http.Request) (interface{}, []error, *api.Ap
 	}
 
 	step, apiErr := qapi.parseStep(r, qapi.defaultRangeQueryStep, int64(end.Sub(start)/time.Second))
-	if err != nil {
+	if apiErr != nil {
 		return nil, nil, apiErr
 	}
 
--- a/pkg/block/fetcher.go
+++ b/pkg/block/fetcher.go
@@ -186,6 +186,12 @@ func NewBaseFetcher(logger log.Logger, concurrency int, bkt objstore.Instrumente
 	}, nil
 }
 
+// NewRawMetaFetcher returns basic meta fetcher without proper handling for eventual consistent backends or partial uploads.
+// NOTE: Not suitable to use in production.
+func NewRawMetaFetcher(logger log.Logger, bkt objstore.InstrumentedBucketReader) (*MetaFetcher, error) {
+	return NewMetaFetcher(logger, 1, bkt, "", nil, nil, nil)
+}
+
 // NewMetaFetcher returns meta fetcher.
 func NewMetaFetcher(logger log.Logger, concurrency int, bkt objstore.InstrumentedBucketReader, dir string, reg prometheus.Registerer, filters []MetadataFilter, modifiers []MetadataModifier) (*MetaFetcher, error) {
 	b, err := NewBaseFetcher(logger, concurrency, bkt, dir, reg)
--- a/pkg/pool/pool.go
+++ b/pkg/pool/pool.go
@@ -9,15 +9,28 @@ import (
 	"github.com/pkg/errors"
 )
 
-type BytesPool interface {
+// Bytes is a pool of bytes that can be reused.
+type Bytes interface {
+	// Get returns a new byte slices that fits the given size.
 	Get(sz int) (*[]byte, error)
+	// Put returns a byte slice to the right bucket in the pool.
 	Put(b *[]byte)
 }
 
-// BucketedBytesPool is a bucketed pool for variably sized byte slices. It can be configured to not allow
+// NoopBytes is pool that always allocated required slice on heap and ignore puts.
+type NoopBytes struct{}
+
+func (p NoopBytes) Get(sz int) (*[]byte, error) {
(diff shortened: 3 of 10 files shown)


bwplotka:
This is so we can use our structs for reliable go benchmarks from outside package (e.g to integrate multiple things): https://gist.github.com/bwplotka/cbcbbcd1802181b7785da11dcc0f5cfd#

test/e2e/e2ethanos/services.go
```diff
@@ -27,7 +27,7 @@ import (
 	"github.com/thanos-io/thanos/pkg/receive"
 )
 
-const logLevel = "info"
+const infologLevel = "info"
```
prmsrswt, line 30: Just a small nit.

```suggestion
const infoLogLevel = "info"
```

test/e2e/e2ethanos/services.go
```diff
@@ -124,7 +124,7 @@ func NewQuerier(sharedDir, name string, storeAddresses, fileSDStoreAddresses, ru
 		"--http-address":          ":8080",
 		"--query.replica-label":   replicaLabel,
 		"--store.sd-dns-interval": "5s",
-		"--log.level":             logLevel,
+		"--log.level":             "debug", // remove later.
```
prmsrswt, line 127: Need to flip this back to `infoLogLevel`.

pkg/api/query/v1.go
```diff
@@ -307,6 +308,7 @@ func (qapi *QueryAPI) query(r *http.Request) (interface{}, []error, *api.ApiErro
 }
 
 func (qapi *QueryAPI) queryRange(r *http.Request) (interface{}, []error, *api.ApiError) {
+	reqTime := time.Now()
```
prmsrswt, line 311: This was added for debugging I guess.

pkg/store/bucket_test.go
```diff
@@ -2037,7 +2034,7 @@ func setupStoreForHintsTest(t *testing.T) (testutil.TB, *BucketStore, []*storepb
 		tmpDir,
 		indexCache,
 		nil,
-		1000000,
+		nil,
```
prmsrswt, line 2037: Don't we need a `chunkPool` with `maxTotal = 100000` here?

prmsrswt (commented):
Looks good to me, just a few nits.

pkg/block/fetcher.go
```diff
@@ -184,6 +184,12 @@ func NewBaseFetcher(logger log.Logger, concurrency int, bkt objstore.Instrumente
 	}, nil
 }
 
+// NewRawMetaFetcher returns basic meta fetcher without proper handling for eventual consistent backends or partial uploads.
+// NOTE: Not suitable to use in production.
+func NewRawMetaFetcher(logger log.Logger, bkt objstore.InstrumentedBucketReader) (*MetaFetcher, error) {
```
yeya24, line 189: I think it is not a good practice to have a public method that is only for testing purposes? I only find this method used in unit tests, then is it better to move this method to that package?

pkg/store/bucket.go
```diff
@@ -282,14 +283,33 @@ type BucketStore struct {
...
+}
+
+type noopGate struct{}
+
+func (noopGate) Start(context.Context) error { return nil }
+func (noopGate) Done()                       {}
+
 // NewBucketStore creates a new bucket backed store that implements the store API against
 // an object store bucket. It is optimized to work against high latency backends.
+// TODO(bwplotka): Move to config at this point (:
```
yeya24, line 312: +1

bwplotka:
Thanks for reviews, will update!

pkg/block/fetcher.go
bwplotka: Do you have any suggestion, if I want to reuse this across tests?

pkg/pool/pool.go
```diff
@@ -9,15 +9,28 @@ import (
 	"github.com/pkg/errors"
 )
 
-type BytesPool interface {
+// Bytes is a pool of bytes that can be reused.
+type Bytes interface {
```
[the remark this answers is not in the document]
bwplotka: `BytesPool` it's too clear, we already have `pool` as a package name so it's known it's pool. That's kind of idiomatic. I would vote to leverage pool pacakge WDYT?

test/e2e/e2ethanos/services.go
bwplotka: Thanks for catching!

bwplotka:
Rebased!

bwplotka:
PTAL @yeya24 @prmsrswt

pkg/api/query/v1.go
```diff
@@ -350,7 +350,7 @@ func (qapi *QueryAPI) queryRange(r *http.Request) (interface{}, []error, *api.Ap
 	}
 
 	step, apiErr := qapi.parseStep(r, qapi.defaultRangeQueryStep, int64(end.Sub(start)/time.Second))
-	if err != nil {
+	if apiErr != nil {
```
yeya24, line 353: Amazing. How did you catch this

yeya24 (approved):
LGTM

pkg/api/query/v1.go
bwplotka: Laser eyes

PR #3657, opened 2020-12-20
store: Make more S3 http.Transport settings configurable.

zecke:
The http.Transport is not auto-tuning and one size does not seem to fit
all cases. In order to respond to a query a store gateway might need to
fetch a large (thousands) of postings, series and chunks.

While the number of idle connections has been increased recently it can
still be too low or expose bursty (opening, closing) behavior. Allow to
tune most of the http.Transport parameters. I considered embedding the
full Transport but there is already a Transport member.

## Verification

I manually started thanos store and used fmt.Printf to see the options, e.g. that MaxIdleConns's default is taken.

### DIFF
--- a/docs/storage.md
+++ b/docs/storage.md
@@ -84,6 +84,11 @@ config:
     idle_conn_timeout: 1m30s
     response_header_timeout: 2m
     insecure_skip_verify: false
+    tls_handshake_timeout: 10s
+    expect_continue_timeout: 1s
+    max_idle_conns: 100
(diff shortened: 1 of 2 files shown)


OGKevin (approved):

bwplotka (approved):
Totally makes sense. LGTM, but ideally I would move this to base level not per storage provider as it can be used for all storages 🤗 

LGTM anyway, let's do this in other PR. Thanks!

bwplotka:
Sorry for massive lag ):

PR #3660, opened 2020-12-21
Export metadata fetcher metrics

pracucci:

## Changes

In Cortex we're experimenting with a [per-tenant bucket index](https://cortexmetrics.io/docs/proposals/blocks-storage-bucket-index/) which is a single json indexing blocks and deletion marks, so that queriers and store-gateway only need to periodically fetch the bucket index to discover new/deleted blocks instead of running "list objects" operations. If turns out to work well, it's a pattern we may consider backporting to Thanos too.

To implement it in the Cortex store-gateway, we're writing our own custom `MetadataFetcher` (working with the bucket index instead of scanning the bucket). Would be nice to expose the same metrics in the custom implementation too, so I'm wondering what's the sentiment to export `FetcherMetrics`, so that it could be reused by Cortex too.

## Verification

N/A

### DIFF
--- a/pkg/block/fetcher.go
+++ b/pkg/block/fetcher.go
@@ -37,32 +37,35 @@ import (
 
 const FetcherConcurrency = 32
 
-type fetcherMetrics struct {
-	syncs        prometheus.Counter
-	syncFailures prometheus.Counter
-	syncDuration prometheus.Histogram
+// FetcherMetrics holds metrics tracked by the metadata fetcher. This struct and its fields are exported
+// to allow depending projects (eg. Cortex) to implement their own custom metadata fetcher while tracking
+// compatible metrics.
+type FetcherMetrics struct {
+	Syncs        prometheus.Counter
+	SyncFailures prometheus.Counter
+	SyncDuration prometheus.Histogram
 
-	synced   *extprom.TxGaugeVec
-	modified *extprom.TxGaugeVec
+	Synced   *extprom.TxGaugeVec
+	Modified *extprom.TxGaugeVec
 }
 
-func (s *fetcherMetrics) submit() {
-	s.synced.Submit()
-	s.modified.Submit()
+func (s *FetcherMetrics) submit() {
+	s.Synced.Submit()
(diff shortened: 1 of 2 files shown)


GiedriusS (commented):
I'm not ideologically opposed to this even though it feels a bit weird exposing the struct that holds the metrics. Sharing the metrics here for me means that we would be able to seamlessly switch between the metadata fetchers from the PoV of the user if needed in the future which is a nice thing. However, I'm curious to know the opinion of other people.

pracucci:
> I'm not ideologically opposed to this even though it feels a bit weird exposing the struct that holds the metrics. Sharing the metrics here for me means that we would be able to seamlessly switch between the metadata fetchers from the PoV of the user if needed in the future which is a nice thing. However, I'm curious to know the opinion of other people.

Feel weird to me too. I'm wondering what could be a better approach. Currently I've copy-pasted the code into Cortex, but it's very error prone (eg. new label values added in the future), due to the usage of `TxGaugeVec` which requires all possible label values in input (these metrics are individually passed to `MetadataFilter` so on the Cortex side we have no easy way to know which are the possible values, unless we duplicate all filters code too).

_I'm very open to better ideas._

bwplotka (approved):
I don't see any problem with this, let's do it, reuse more!

pkg/block/fetcher.go
```diff
@@ -35,32 +35,35 @@ import (
...
 )
 
-type fetcherMetrics struct {
-	syncs        prometheus.Counter
-	syncFailures prometheus.Counter
-	syncDuration prometheus.Histogram
+// FetcherMetrics holds metrics tracked by the metadata fetcher. This struct and its fields are exported
+// to allow depending projects (eg. Cortex) to implement their own custom metadata fetcher while tracking
+// compatible metrics.
+type FetcherMetrics struct {
```
bwplotka, line 41: What's wrong with this approach? 🤔 

Potentially: Fetcher hasn't been designed to be replaced with a different implementation. 

I think this the evolution of package that needs more implementations than one.

PR #3664, opened 2020-12-22
scripts/quickstart: add external label for Thanos Rule

GiedriusS:
Add an external label for Thanos Rule akin to Thanos Receive:
```
--label 'receive="true"' \
```

In the same fashion, let's add `rule=true`. This enables uploading
blocks for Thanos Rule as Thanos blocks must have at least one external
label.

## Verification

`./scripts/quickstart.sh` works as expected


yeya24 (approved):
Thanks! LGTM.

PR #3669, opened 2020-12-23
website: improve documentation for bucket cache

### LINKED ISSUE #3666: Improve documentation for caching bucket configuration

Relates to my comment here: https://github.com/thanos-io/thanos/issues/1979#issuecomment-749831642

Update docs here: https://thanos.io/tip/components/store.md/#caching-bucket with an example of actually implementing the config for memcached

wiardvanrij:

## Changes
Added an example on the full settings for the Thanos Store bucket cache.

Fixes #3666

### DIFF
--- a/docs/components/store.md
+++ b/docs/components/store.md
@@ -297,14 +297,14 @@ Currently only memcached "backend" is supported:
 type: MEMCACHED # Case-insensitive
 config:
   addresses: []
-  timeout: 0s
-  max_idle_connections: 0
-  max_async_concurrency: 0
-  max_async_buffer_size: 0
-  max_get_multi_concurrency: 0
-  max_item_size: 0
+  timeout: 500ms
+  max_idle_connections: 100
+  max_async_concurrency: 20
+  max_async_buffer_size: 10000
+  max_item_size: 1MiB
+  max_get_multi_concurrency: 100
   max_get_multi_batch_size: 0
-  dns_provider_update_interval: 0s
+  dns_provider_update_interval: 10s
 chunk_subrange_size: 16000
 max_chunks_get_range_requests: 3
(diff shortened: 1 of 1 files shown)


docs/components/store.md
```diff
@@ -296,9 +296,15 @@ Currently only memcached "backend" is supported:
 ```yaml
 type: MEMCACHED # Case-insensitive
 config:
-  addresses:
-    - localhost:11211
-
+  addresses: []
+  timeout: 0s
+  max_idle_connections: 0
+  max_async_concurrency: 0
```
GiedriusS, line 302: What if we would actually write down the default values here and in the other place in this file? I am talking about this: https://github.com/thanos-io/thanos/blob/1fff9a7157d1f3b7d5e0ce36f999b10bb3ae7ec4/pkg/cacheutil/memcached_client.go#L46-L55. I think it would be even more convenient for Thanos users.

GiedriusS (commented):
The changes look good however I have one suggestion.

docs/components/store.md
wiardvanrij: Seems logical to do. Currently I lack some time, but will adjust it soon

docs/components/store.md
```diff
@@ -296,9 +296,15 @@ Currently only memcached "backend" is supported:
 ```yaml
 type: MEMCACHED # Case-insensitive
 config:
-  addresses:
-    - localhost:11211
-
+  addresses: []
```
bwplotka, line 299: Hm... Ideally we could autogenerate this 🤗  but we can update to Giedrius proposed thing first and automate at some point

wiardvanrij:
Sorry for the massive delay :D

bwplotka (approved):
💪🏽 
Thanks!

PR #3670, opened 2020-12-23
Chore: update store cmd description

yeya24:

The description of store is outdated and we support Aliyun OSS already.
Not sure if there is a good way to automate this if we support more Object storage in the future.


prmsrswt (approved):
Thanks, LGTM.

PR #3672, opened 2020-12-25
rule: prevent rule crash from no such host error

damnever:
The rule will crash due to alert managers restart, same with https://github.com/thanos-io/thanos/pull/3257 but I use `dnssrv+`/`dnssrvnoa+`.

Maybe we should make [this line](https://github.com/thanos-io/thanos/blob/3908812e60d1baabf0ba38cd8bf6edf78206d7ab/cmd/thanos/rule.go#L803) fault-tolerant rather than make changes in [pkg/discovery/dns](https://github.com/thanos-io/thanos/tree/master/pkg/discovery/dns).

### DIFF
--- a/pkg/discovery/dns/godns/resolver.go
+++ b/pkg/discovery/dns/godns/resolver.go
@@ -1,0 +1,25 @@
+// Copyright (c) The Thanos Authors.
+// Licensed under the Apache License 2.0.
+
+package godns
+
+import (
+	"net"
+
+	"github.com/pkg/errors"
+)
+
+// Resolver is a wrapper for net.Resolver.
+type Resolver struct {
+	*net.Resolver
+}
+
+// IsNotFound checkout if DNS record is not found.
+func (r *Resolver) IsNotFound(err error) bool {
+	if err == nil {
+		return false
+	}
+	err = errors.Cause(err)
+	dnsErr, ok := err.(*net.DNSError)
+	return ok && dnsErr.IsNotFound
+}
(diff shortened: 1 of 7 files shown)


OGKevin (approved):

damnever:
@bwplotka please take a look.

GiedriusS:
> @bwplotka please take a look.

So no one else is permitted to take a look? :(

GiedriusS (commented):
I think this is the right level of abstraction - it should be up to the end-user (i.e. the caller) to react in a certain way to an error. However, I don't like hardcoding the exact errors into the code at this exact place. What about having a method `IsNotFoundError(err error) bool` in the `ipLookupResolver` interface that we could use here? Then we would have same pattern as in `BucketReader` and I think that works well.

Then, in the case of the `miekg` resolver we could also return a special error if `lookupWithSearchPath` returns NXDOMAIN & no `error` so that the user of that resolver would also know when such an issue occurs.

pkg/discovery/dns/godns/resolver.go
```diff
@@ -0,0 +1,25 @@
...
+)
+
+// Resolver is a wrapper for net.Resolver.
+type Resolver struct {
+	*net.Resolver
+}
+
+// IsNotFound checkout if DNS record is not found.
+func (r *Resolver) IsNotFound(err error) bool {
+	if err != nil {
```
GiedriusS, line 19: Do you mean:
```suggestion
	if err == nil {
```
Here?

GiedriusS (commented):
Just one suggestion but overall LGTM. Thank you!

pkg/discovery/dns/godns/resolver.go
damnever: Yes..

GiedriusS (approved):

GiedriusS:
Thank you!

PR #3681, opened 2020-12-29
Store Gateway Doc Update 

### LINKED ISSUE #3045: docs: Add mention about needed disk space (estimations) for store gateway.
As per: https://cloud-native.slack.com/archives/CK5RSSC10/p1597686261049000

sar99:
Signed-off-by: Saniya Arora <saniya.rishi@gmail.com>

Change is not relevant to the end user.

## Changes

Updated the documentation for the Store gateway component, correcting the estimate disk space used by the precomputed index. Resolves #3045 according to the slack conversation mentioned.

### DIFF
--- a/docs/components/store.md
+++ b/docs/components/store.md
@@ -23,7 +23,7 @@ config:
   bucket: example-bucket
 ```
 
-In general about 1MB of local disk space is required per TSDB block stored in the object storage bucket.
+In general, an average of 6 MB of local disk space is required per TSDB block stored in the object storage bucket. It is for the precomputed index, which includes symbols and postings offsets. The size of this precomputed index can go up to 30 MB per block.
 
 ## Flags
 
--- a/docs/components/store.md
+++ b/docs/components/store.md
@@ -23,7 +23,7 @@ config:
   bucket: example-bucket
 ```
 
-In general, an average of 6 MB of local disk space is required per TSDB block stored in the object storage bucket. It is for the precomputed index, which includes symbols and postings offsets. The size of this precomputed index can go up to 30 MB per block.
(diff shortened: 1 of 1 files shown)


docs/components/store.md
```diff
@@ -23,7 +23,7 @@ config:
   bucket: example-bucket
 ```
 
-In general about 1MB of local disk space is required per TSDB block stored in the object storage bucket.
+In general, an average of 6 MB of local disk space is required per TSDB block stored in the object storage bucket. It is for the precomputed index, which includes symbols and postings offsets. The size of this precomputed index can go up to 30 MB per block.
```
bwplotka, line 26: Thanks, but is it 6MB or 36MB of local disk in the end? (:

sar99: What I understood from the slack conversation was that the space consumption is 6MB in average case but can go up to 30MB in extreme cases. Please do let me know if I got on the wrong track somewhere. :)

bwplotka: Sorry for lag - yes it varies (: Let's mention this in docs then. What about:

```
In general, an average of 6 MB of local disk space is required per TSDB block stored in the object storage bucket, but high cardinality blocks with large label set it can go even to 30MB and more. It is for the pre-computed index, which includes symbols and postings offsets as well as metadata JSON.
```

kakkoyun:
Hey @sar99, could you have another look at it? There are some comments to be addressed then we can merge it.

sar99:
Sincere apologies for the delay. I'll just make the above-recommended changes.

docs/components/store.md
sar99: Thank you for the help! :)
I have made the recommended change.

bwplotka (commented):
Thanks! 🤗

bwplotka (approved):

PR #3685, opened 2020-12-30
Lints: Remove UnixSocket not used and use right comment

MalloZup:
# Description:
Just found this 2 artifacts while doing other things.


yeya24 (approved):
Nice catch. Thanks!

PR #3700, opened 2021-01-06
ui: make old bucket viewer UI work with vanilla blocks

GiedriusS:
Make the old bucket viewer UI work with vanilla Prometheus blocks by
checking whether the `Thanos` part exists before formatting the HTML.

Only changing the old UI here to solicit responses on whether we should
change the API not to return the `thanos` key if it does not exist. IMHO
we should keep it as is and add checks as in this PR.

Screenshot:
ss

## Verification

Built Thanos and ran it with the `FILESYSTEM` bucket on a local Prometheus directory.

### DIFF
--- a/pkg/ui/static/js/bucket.js
+++ b/pkg/ui/static/js/bucket.js
@@ -77,10 +77,10 @@ function draw() {
     }
 }
 
-// isThanosBlock eturns true if it is a block that has been processed by Thanos.
+// isThanosBlock returns true if it is a block that has been processed by Thanos.
 function isThanosBlock(d) {
     return d.thanos.labels !== null &&
-        d.thanos.source != ""
+        d.thanos.source !== ""
 }
 
 function stringify(map) {
--- a/pkg/ui/static/js/bucket.js
+++ b/pkg/ui/static/js/bucket.js
@@ -34,6 +34,12 @@ function draw() {
         dataTable.addRows(thanos.blocks
             .sort((a, b) => a.thanos.downsample.resolution - b.thanos.downsample.resolution)
             .map(function(d) {
+                if (!isThanosBlock(d)) {
+                    title = 'Prometheus'
+                    titles['Prometheus blocks with no Thanos data'] = title
+
+                    return [title, `level: ${d.compaction.level}`, generateTooltip(d), new Date(d.minTime), new Date(d.maxTime)]
+                }
                 // Title is the first column of the timeline.
                 //
                 // A unique Prometheus label that identifies each shard is used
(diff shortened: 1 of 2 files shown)


pkg/ui/static/js/bucket.js
```diff
@@ -71,6 +77,12 @@ function draw() {
     }
 }
 
+// isThanosBlock eturns true if it is a block that has been processed by Thanos.
```
squat, line 80: nit: 
```suggestion
// isThanosBlock returns true if it is a block that has been processed by Thanos.
```

bwplotka:
What about new UI bucket viewer? Will it not work?

bwplotka (approved):
Otherwise, cool idea! 🤗 LGTM ❤️

pkg/ui/static/js/bucket.js
```diff
@@ -71,6 +77,12 @@ function draw() {
...
+function isThanosBlock(d) {
+    return d.thanos.labels !== null &&
```
squat, line 82: Is it guaranteed that non-thanos blocks have a `thanos` field at all? In other words, can we be certain that `d.thanos !== undefined`? Or do we need to do `d.thanos && d.thanos.labels !== null && d.thanos.source !== ""`?

pkg/ui/static/js/bucket.js
```diff
@@ -71,6 +77,12 @@ function draw() {
...
+        d.thanos.source != ""
```
squat, line 83:
```suggestion
        d.thanos.source !== ""
```

squat (commented):
Super cool idea :rocket:

GiedriusS:
> What about new UI bucket viewer? Will it not work?

That's step two that I (we) will do independently. This is a good start. The main question I had is whether we want this:
https://github.com/thanos-io/thanos/blob/master/pkg/block/metadata/meta.go#L59

To become:

```go
Thanos *Thanos `json:"thanos"`
```

If this looks good for everyone then in the future PR for the new UI I will change that struct + these checks. If not then the checks will be similar to the ones currently in this PR. All in all, the current PR looks good to me besides the nit by @squat.

pkg/ui/static/js/bucket.js
GiedriusS: Yep: https://github.com/thanos-io/thanos/blob/master/pkg/block/metadata/meta.go#L59

Since it is not a reference, it means that `d.thanos` will always exist. We might want to change this but then this change will be visible to the end user. I think we should convert it into a reference in the future PR where I (we) will make the bucket viewer UI work with non-Thanos blocks.

Zero value for `labels` is `nil` and for `source` it is an empty string so we don't need to do this. All of the other struct members have `omitempty` so they do not appear at all. It should be enough to check these.

prmsrswt: We can use [optional chaining](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/Optional_chaining) here to handle the case when `d.thanos` is `undefined`.
```suggestion
    return d.thanos?.labels !== null &&
```

prmsrswt (approved):
Looks great!

Issue #2600, opened 2020-05-13
gRPC: Isolate Info to separate gRPC API so we can reuse discovery for different RPCs: Store, Rules and Targets


bwplotka:
Something decided on our public meeting: https://docs.google.com/document/d/165L9DVKHW0sS2PfKYsUCDikwq5XngKkCqTOVjzC6qQE/edit#heading=h.7dcto2wgvh4q

Something to do in later step. This will avoid us to have all those weird flags like separate SD, separate store etc.
We can disable certain APIs if needed on API level (e.g ruler can disable RulesAPI if needed) (: 

cc @s-urbaniak

povilasv:
I think I would vote for new flag and StoreInfo endpoint returning types supported, like previous proposal mentioned.

Plus on the sidecar you could do `--rule-api=disabled` , `--store-api=disabled` or smth similiar.

Thoughts? Maybe we can write a mini proposal for this?

I'm also thinking that potentially we could do TargetsAPI which would show leaf node Targets scraped (I mean Prometheus /targets endpoint) and dedup them? Thoughts?

bwplotka:
> I think I would vote for new flag 

New flag? what flag?

> and StoreInfo endpoint returning types supported, like previous proposal mentioned.

Hm, yea we wanted to different info per type so its kind of equivalent. So we can detect things either by grpc reflection or info... what would be better? :thinking:

bwplotka:
Also for target API there is issue already: https://github.com/thanos-io/thanos/issues/1375

bwplotka:
Ok we discussed offline that it would be amazing to actually split discovery as separate gRPC API.

I think we can do it after Rules API is landed, but overall idea is that `Info` endpoint is shared and Querier will only  wait for `--endpoint` flag or something which will then ask for info API and  THEN discover what APIs are there (: This will ensure things like https://github.com/thanos-io/thanos/pull/2407#discussion_r414743260 and per store TLS will be much easier to implement.

cc @brancz @s-urbaniak @kakkoyun @squat @povilasv @GiedriusS

povilasv:
:+1: I like `--endpoint`

bwplotka:
Changed title.

GiedriusS:
Still valid.

bwplotka:
We need that! (:

yeya24:
Still valid.

bwplotka:
FYI: This assumes Querier is the one that federates "stuff"  (:

[...]

PR #3703, opened 2021-01-07
api: Added Proposal for Discovery Endpoint idea (InfoAPI)

lilic:

## Changes

Add the proposal for unified endpoint discovery as per discussion in https://github.com/thanos-io/thanos/issues/2600.

I highlighted two options, but again up to the maintainers which one you choose, happy to work on any of the ones that we agree! 🎉 

I was not sure what to put under work plan and there was no fixed format for proposals that I could find, might have missed it? I copied from some existing ones if there is anything missing let me know! 🙂

### DIFF
--- a/docs/proposals/210701_endpoint_discovery.md
+++ b/docs/proposals/210701_endpoint_discovery.md
@@ -1,0 +1,62 @@
+---
+title: Unified Endpoint Discovery
+type: proposal
+menu: proposals
+status: XXX
+---
+
+### Related Tickets
+
+* https://github.com/thanos-io/thanos/issues/2600
+
+### Summary
+
+We want to propose a new flag called `--endpoint=<address>` that will be passed to Thanos query and replace the current `--store=<address>` and `--rule=<address>` flags, and add a discovery mechanism which will be used to determine what types of services this endpoint can serve.
+
+### Motivation
+
+Currently, in Thanos Query, the discovery of rules APIs happens via Store API's Info method. This makes it harder if we ever want to not have a coupling to the Store API (which is already planned for [scalable ruler proposal](https://github.com/thanos-io/thanos/blob/master/docs/proposals/202005_scalable-rule-storage.md)).
+
+We also require passing two different flags to the Thanos Query component `--store=<address>` and `--rule=<address>`. If users use both flags, the Query component performs DNS discovery on often the same address multiple times, this can cause occasional DNS issues as this resolving happens so frequently. This is especially confusing when one works and the other doesn't.
+
+Adding new APIs in the future, like
+
+### Goals
+
+Unite discovery of endpoints, which would make code cleaner as well as provide building blocks to provide discover and support for new gRPC services, for example, the Targets API/Exemplar API/Metadata API.
+
+### Solution
+
+Add a new flag called `--endpoint` to Thanos query, and auto-discover what services that endpoint is serving based on metadata each gRPC server exposes.
+
+Each component will expose an Info API, that includes various metadata listed below. Discovery of endpoints will happen via this Info API, there might be a case that the discovery of gRPC will also have to happen via [gRPC reflection](https://github.com/grpc/grpc/blob/master/doc/server-reflection.md), this will be shown when implemention starts.
+
+Info API metadata would include the following fields regardless of the type:
+```
+info:
+  external_labels: blah_1, blah_2, ...
+  store:
+    minTime: 2
+    maxTime: 4
+  rule:
+    blah: ...
+```
+
+Right now this info API would be added to existing components Ruler, Store, Sidecar, Compact, Downsample, Receive. But this would be added in the future for Targets, Examplers, etc.
+
+The upgrade path would be possible in this case, as it's strictly additive for endpoints, and there will be a migration path to fallback to current discovery methods which could be removed in the future.
+
+#### Alternatives
+
+Split Info method out of the Store API into a separate gRPC service called `info`. Each component must implement and expose this. The discovery would happen via this info gRPC service, which each component would need to explicitly state which API it supports. Besides that information, we would also need to include here various info, e.g. LabelSets, Max/Min time, etc. The drawback for this solution is harder upgrades as this would be more complex to implement as we would need to fall back to older APIs.
+
+### Upgrade plan
+
+Regardless of the solution we will be not removing any flags for a given period of time and have fallback code in place to be able to discover in the same way as we do right now. This will ensure smooth upgrade path for the users that make use of the current flags for discovery. We will mark the current flags as depracted and after two releases have passed we can remove the existing store and rule flag from query component and remove any migration code.
+
+### Work plan
+
+1. We would be adding the new info APIs and flags needed for the above solution, so this part would be entirely additive.
+2. Second step would be the integration of this APIs and flags and migration of existing code.
+3. We would not be removing the existing flags (`--store` and `--rule`) for some grace period of time, after that passes we would also have to remove it and any migration code.
+
--- a/docs/proposals/210701_endpoint_discovery.md
+++ b/docs/proposals/210701_endpoint_discovery.md
@@ -42,7 +42,7 @@ info:
     blah: ...
 ```
 
-Right now this info API would be added to existing components Ruler, Store, Sidecar, Compact, Downsample, Receive. But this would be added in the future for Targets, Examplers, etc.
+Right now this info API would be added to existing components Ruler, Store, Sidecar, Compact, Downsample, Receive. But this would be added in the future for Targets, Exemplers, etc.
 
 The upgrade path would be possible in this case, as it's strictly additive for endpoints, and there will be a migration path to fallback to current discovery methods which could be removed in the future.
 
@@ -59,4 +59,3 @@ Regardless of the solution we will be not removing any flags for a given period 
 1. We would be adding the new info APIs and flags needed for the above solution, so this part would be entirely additive.
 2. Second step would be the integration of this APIs and flags and migration of existing code.
 3. We would not be removing the existing flags (`--store` and `--rule`) for some grace period of time, after that passes we would also have to remove it and any migration code.
-


lilic:
cc @bwplotka @brancz

docs/proposals/210701_endpoint_discovery.md
```diff
@@ -0,0 +1,44 @@
...
+
+The upgrade path would be possible in this case, as it's strictly additive for endpoints, and therefore in theory users would need to just upgrade in the correct order, with Thanos Query component being the last to be upgraded.
+
+#### Alternatives
+
+Split Info method out of the Store API into a separate gRPC service called `info`. Each component must implement and expose this. The discovery would happen via this info gRPC service, which each component would need to explicitly state which API it supports. Besides that information, we would also need to include here various info, e.g. LabelSets, Max/Min time, etc. The drawback for this solution is harder upgrades as this would be more complex to implement as we would need to fall back to older APIs.
```
brancz, line 36: I could see this one working as well if it is a strict subset kind of in the form of:

```yaml
info:
  external_labels: ...
  store:
    minTime: ...
    maxTime: ...
  rule:
    xyz: ...
```

What I do like about this is that we would only have to do a single call to discover all metadata, regardless of the type of store, and through the existence of the sub-section we know that the API is available and ready to serve. (also fine with having a top level field that says whether an API is served is fine by me as well, implicit things tend to go wrong or be forgotten)

Although I do agree, the migration path seems kind of painful code complexity wise. If we should go for this, I would actually suggest to implement this 100% parallel and duplicate, so that when done the "old" code can just be deleted in its entirety.

docs/proposals/210701_endpoint_discovery.md
```diff
@@ -0,0 +1,44 @@
...
+
+### Work plan
+
+Start by deprecating the `--store=<address> --rule=<address>` and replace with `--endpoint=<address>` in the Thanos Query component itself and perform discovery in the Thanos Query component.
```
brancz, line 40: I think there could be a bit more detail here. At the very least, the way it can be split up is to do all the 100% additive work (eg. the new APIs/Services), and then the migration/integration of existing code as a second step.

docs/proposals/210701_endpoint_discovery.md
```diff
@@ -0,0 +1,44 @@
...
+
+### Questions to answer
+
+1. What guarantees do we give to the users, when migrating between versions? Can we just remove flags or do we need to fallback to existing ones?
```
brancz, line 44: We prefer not to, and try to at least have some grace period but generally speaking it is possible to remove flags.

brancz (commented):
Very nice! Very much looking forward to having this unified! I do think the alternative with the comments I added are more feasible, but I don't feel too strongly, I could see either working.

Minor side note, regardless of what we choose to do, we should make sure to document a setup once https://github.com/thanos-io/thanos/blob/master/docs/proposals/202005_scalable-rule-storage.md is also implemented, as that will be a bit odd of a setup, requiring potentially multiple `--endpoint` flags as receive and rule component may be within the same Pod (in the case of the user using the fallback solution proposed in that other proposal).

docs/proposals/210701_endpoint_discovery.md
```diff
@@ -0,0 +1,44 @@
...
+
+### Goals
+
+Unite discovery of endpoints, which would make code cleaner as well as provide building blocks to provide discover and support for new gRPC services, for example, the Targets API/Exemplar API/Metadata API.
+
+### Solution
+
+Add a new flag called `--endpoint` to Thanos query, and auto-discover what services that endpoint is serving based on metadata each gRPC server exposes.
+
+The plan would be for the initial discovery of gRPC services available to happen via gRPC reflection. To get further information, we would be adding an Info method to the Rules API. This would contain similar info or metadata information as we get right now from the Store Info method. Potentially in the future, we would implement this same Info method for other services in the, for example, the Targets, Examplers, etc. As mentioned the discovery would happen via reflection initially, the Info method in the Rules API would expose extra metadata about external labels for now, and we can always add more data to it in the future as needed.
```
bwplotka, line 30: Nice! I think that might work and is clean, however, there are few other downsides. Can list more ups & downs between this and alternatives? 🤗  Right now we talked only about upgradability.

1. For example what about overhead? In this case, what if the API server implements both Store and Rules (and maybe Exemplars and Metadata as well someday)? Which Info do we check for healthiness and metadata propagation? Both? That introduces 4x more very often made microservices call. 🤔 

2. What about boilerplate when adding a new API?

Additionally: Important details that might be missing here is `health checking` as well. We use `Info` to check if the underlying API is available or not. Do we change that, or does it stays in?

docs/proposals/210701_endpoint_discovery.md
bwplotka, line 36: >  The drawback for this solution is harder upgrades as this would be more complex to implement as we would need to fall back to older APIs.

I might miss something. I for proposed solution we also would need to fall back to older APIs for components when upgrading both Querier to newer version expecting Rules API to have Info method, and upgrading RulesAPI server to work with old Queriers, no?

docs/proposals/210701_endpoint_discovery.md
```diff
@@ -0,0 +1,44 @@
...
+
+The upgrade path would be possible in this case, as it's strictly additive for endpoints, and therefore in theory users would need to just upgrade in the correct order, with Thanos Query component being the last to be upgraded.
```
bwplotka, line 32: I would argue that this is exactly the same in terms of upgradability than an alternative of having a separate, unified `Info` endpoint. Why it's different?

It's strictly additive sure, but in the same way, adding a unified `Info` endpoint is addictive, no? 🤗

docs/proposals/210701_endpoint_discovery.md
bwplotka: To me, both solutions has the same drawback and we might want to propose some solution in order to mitigate it no matter what we choose here.

docs/proposals/210701_endpoint_discovery.md
```diff
@@ -0,0 +1,44 @@
...
+
+#### Alternatives
+
+Split Info method out of the Store API into a separate gRPC service called `info`. Each component must implement and expose this. The discovery would happen via this info gRPC service, which each component would need to explicitly state which API it supports. Besides that information, we would also need to include here various info, e.g. LabelSets, Max/Min time, etc. The drawback for this solution is harder upgrades as this would be more complex to implement as we would need to fall back to older APIs.
+
```
bwplotka, line 37: There might be a (maybe poor) alternative `#2` that actually might have a better upgradability path.

* Create `SeriesAPI` that will copy `Series` and `Labels...` methods from StoreAPI.
* Deprecate (but not remove) some methods on `StoreAPI`
* Treat `StoreAPI` as the new `Info` only API that all APIs have to implement. Naming is not the best (`InfoAPI` would be better, but it will do the work).
* For compatibility or existing Series-capable servers will register both `StoreAPI` and `SeriesAPI`. 
* NewQuerier will get stop from getting Series from StoreAPI unless Info (e.g) indicates it's older than X Thanos version
* New non-series servers will implement StoreAPI Info methods only and Rules API etc. This will work perfectly with older Queriers as we pass `rules` and `stores` manually (: 

WDYT?

docs/proposals/210701_endpoint_discovery.md
bwplotka, line 44: We can change flags, but requiring strict order of upgrade would be nice to avoid

docs/proposals/210701_endpoint_discovery.md
bwplotka, line 40: We definitely need to write down upgradability plan (or why it's not needed if it's not) for chosen option (:

bwplotka (commented):
Amazing work, super clean design doc 💪🏽 

Let's iterate over it. Feel free to disagree with my points, added few thoughts 🤗 

@yeya24 @pracucci @kakkoyun @GiedriusS @squat feedback wanted - it's major API redesign so let's make it carefully ⚠️

docs/proposals/210701_endpoint_discovery.md
metalmatze, line 30: Could you link to a page that explains gRPC reflection in a bit more detail? I'm not too familiar and think most others might be neither.

docs/proposals/210701_endpoint_discovery.md
```diff
@@ -0,0 +1,44 @@
...
+### Summary
+
+We want to propose a new flag called `--endpoint=<address>` that will be passed to Thanos query and replace the current `--store=<address>` and `--rule=<address>` flags, and add a discovery mechanism which will be used to determine what types of services this endpoint can serve.
+
+### Motivation
+
+Currently, in Thanos Query, the discovery of rules APIs happens via Store API's Info method, this makes it harder if we ever want to not have a coupling to the Store API (which is already planned for [scalable ruler proposal](https://github.com/thanos-io/thanos/blob/master/docs/proposals/202005_scalable-rule-storage.md)).
+
+We also require passing two different flags to the Thanos Query component `--store=<address>` and `--rule=<address>` and perform DNS discovery on often the same address multiple times, this can cause occasional DNS issues. Which is especially confusing when one works and the other doesn't.
```
metalmatze, line 20:
```suggestion
We also require passing two different flags to the Thanos Query component `--store=<address>` and `--rule=<address>` and perform DNS discovery on often the same address multiple times, **this can cause occasional DNS issues**. Which is especially confusing when one works and the other doesn't.
```
Maybe we can make this bold? Am I correctly understanding that this is basically the entire problem with the current approach? A sentence or two on what explicitly were the problems with DNS would be good to know/learn.

lilic: The main problem is that we have to perform calls often twice because in a lot of use cases the endpoints we pass for ruler and store to the querier component, tend to be the same address. This for us at least caused DNS resolving issues, and one got resolved another did not. 

Another motivation is to also unite the discovery to be able to just pass an endpoint and not needing to couple the ruler to the store APIs as it is currently in the thanos querier, that is described in the initial issue and something we found as well when going through the code. So if just ruler is discovered we can at least potentially fetch rules and serve those. I believe this is also one of the uses cases for the work you are doing?

So I would say to sum up, both of those are the main motivation, not sure I would highlight one over the other, wdyt?

docs/proposals/210701_endpoint_discovery.md
lilic: Of course, good point! Would linking to [this](https://github.com/grpc/grpc/blob/master/doc/server-reflection.md) be enough, or also the various functions we would actually call to perform it? I have done the POC locally so can include those as well, not sure if its too much details then?

docs/proposals/210701_endpoint_discovery.md
lilic: Got it, thanks!

bwplotka:
Are we ready for another review round? 🤗

lilic:
No sorry, I was extremely busy with other company internal things and did not have a chance to address any of the comments yet. But will do it today! If this is super urgent to be implemented (e.g. this week) and someone has more time they can take over as well, let me know and I will close this out! Thanks for patience 😅

docs/proposals/210701_endpoint_discovery.md
lilic: So if I understand correctly we would in this case not go via reflection at all to discover the type, as we would always expect each service to serve this metadata?

docs/proposals/210701_endpoint_discovery.md
lilic: Sounds good in any case yes!

docs/proposals/210701_endpoint_discovery.md
lilic: Thanks!

lilic: Thanks, for now it then make more sense to keep the flags and upgrade would not need to be ordered and we can remove later on. I added this part to the work plan, let me know what you think!

docs/proposals/210701_endpoint_discovery.md
lilic: Added, let me know how that looks and if there is anything else I should add there? Thanks!

docs/proposals/210701_endpoint_discovery.md
```diff
@@ -0,0 +1,44 @@
...
+
```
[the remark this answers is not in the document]
lilic: I added this under goals, I think we are saying the same thing? But I guess goals and motivation are similar often :D Should I maybe just remove goals then, what do you think?

docs/proposals/210701_endpoint_discovery.md
lilic: Yes for now I added to the work plan and upgrade plan that we will in thanos query component at least support existing flags and discovery mechanism for two more releases I think this solves this comment?

docs/proposals/210701_endpoint_discovery.md
lilic: Well adding a new unified info would mean removing or moving the current one from the Store, whereas this one proposes keeping that one.

docs/proposals/210701_endpoint_discovery.md
s-urbaniak: +1 on this proposal from my side, having backwards compatible methods is something I believe we shortly discussed as well when we talked about api discovery in the past.

docs/proposals/210701_endpoint_discovery.md
brancz: My bad, I should have been more specific. That is exactly what I meant.

docs/proposals/210701_endpoint_discovery.md
brancz: I think @bwplotka only wanted to enumerate that this option is technically possible. I think the same kind of migration can be done with https://github.com/thanos-io/thanos/pull/3703#discussion_r553971668, by adding the new API, and only having the new `endpoint` flag use that mechanism and the existing flags use the current/old mechanism.

docs/proposals/210701_endpoint_discovery.md
lilic: @bwplotka What do you think of choosing this option, its similar to your initial idea!

After we agree on it, I will document it in details, does that process sound good? 🙂

docs/proposals/210701_endpoint_discovery.md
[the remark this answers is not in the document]
bwplotka: No no, if it's in it goals then fine

Just one place is enough. But yes - motivation defines requirements and drive goals (:

docs/proposals/210701_endpoint_discovery.md
bwplotka: You are right. Unified means removing from Store. However, I think the problem lies that no matter if you add, remove or modify there are upgrade steps that we have to do anyway to make sure there is no strict ordering needed. 

My point is: Both the proposed solution and unified Info needs the same amount of work I would say. That's why the upgrade path is exactly the same as with unified endpoints unless I am missing something. Happy to explain offline 🤗

bwplotka: BUT: @brancz idea which is more clarified  indeed has bit more work on reusing existing Info messages, so ignore me (:

docs/proposals/210701_endpoint_discovery.md
bwplotka: > @bwplotka What do you think of choosing this option, its similar to your initial idea!

It allows Unified `InfoAPI` so yes!

However still I would say `reflection vs checking fields` is something not agreed. With this structure @brancz  proposed we can avoid reflection, but I still think we can (or maybe have to) use reflection for cases where Info will not include any specific component field.

docs/proposals/210701_endpoint_discovery.md
bwplotka: > I think @bwplotka only wanted to enumerate that this option is technically possible.

Well, I kind of like it too. ;p Something to consider - but yes what we really propose here is implementation detail of different migration/upgradability logics

I think we should split upgradability and end state (:

docs/proposals/210701_endpoint_discovery.md
lilic: Sounds great to me, thanks for commenting, I think maybe the discovery might make more sense in the actual PR. We can leave it to be not reflection for now here and in case it makes sense to use reflection initially or for some parts there are no objections?

So to sum up:
- we want to use unifed InfoAPI that includes the above metadata
- discovery via that API or reflection depending
- and try to make upgrades work smoothly regardless

@brancz @bwplotka does that sum it up or did I misunderstand any points here? Thank you!

brancz: That sums it up well, and I agree reflection could come in handy in a migration path, I'd say that would be an implementation detail that can be discussed on the PR though. I think for the scope of this proposal, I'm happy with this.

bwplotka: Perfect summary + 💯

brancz (approved):
Really like how short and precise this turned out. I think it's a testament to how well this was thought out. Nice work!

lilic:
Thanks @brancz, I had to force push as docs was complaining, but nothing changed apart from formating. 

@bwplotka and others please have a look as well, thank you! 🙇‍♀️

docs/proposals/210701_endpoint_discovery.md
```diff
@@ -0,0 +1,62 @@
...
+info:
+  external_labels: blah_1, blah_2, ...
+  store:
+    minTime: 2
+    maxTime: 4
+  rule:
+    blah: ...
+```
+
+Right now this info API would be added to existing components Ruler, Store, Sidecar, Compact, Downsample, Receive. But this would be added in the future for Targets, Examplers, etc.
```
bwplotka, line 45:
```suggestion
Right now this info API would be added to existing components Ruler, Store, Sidecar, Compact, Downsample, Receive. But this would be added in the future for Targets, Exemplers, etc.
```

bwplotka (approved):
Solid. LGTM. Agree with @brancz - good discussion. 

Minor typo only, not a blocker 🤗

PR #3557, opened 2020-12-09
Use direct cache in index reader for symbol values

yeya24:
Signed-off-by: Ben Ye <yb532204897@gmail.com>

* [ ] I added CHANGELOG entry for this change.
* [x] Change is not relevant to the end user.

## Changes

Apply the same improvement as the upstream Prometheus pr https://github.com/prometheus/prometheus/pull/8265.

The benchmark result is great comparing to the current master.

```
name                                                             old time/op    new time/op    delta
TelemeterRealData_Series/alerts2w/01EPXBGA0413QZMTF4XQ0M4Q3E        16.7s ± 0%     11.8s ± 0%  -29.31%
TelemeterRealData_Series/alerts15s/01EPXBGA0413QZMTF4XQ0M4Q3E       3.95s ± 0%     3.93s ± 0%   -0.53%
TelemeterRealData_Series/subssyncs2w/01EPXBGA0413QZMTF4XQ0M4Q3E     2.44s ± 0%     1.18s ± 0%  -51.63%
TelemeterRealData_Series/subs2w/01EPXBGA0413QZMTF4XQ0M4Q3E          14.3s ± 0%      6.3s ± 0%  -56.04%
TelemeterRealData_Series/subs15s/01EPXBGA0413QZMTF4XQ0M4Q3E         2.07s ± 0%     2.11s ± 0%   +2.00%

name                                                             old alloc/op   new alloc/op   delta
TelemeterRealData_Series/alerts2w/01EPXBGA0413QZMTF4XQ0M4Q3E       4.05GB ± 0%    3.50GB ± 0%  -13.54%
TelemeterRealData_Series/alerts15s/01EPXBGA0413QZMTF4XQ0M4Q3E      1.09GB ± 0%    1.08GB ± 0%   -0.37%
TelemeterRealData_Series/subssyncs2w/01EPXBGA0413QZMTF4XQ0M4Q3E     663MB ± 0%     545MB ± 0%  -17.78%
TelemeterRealData_Series/subs2w/01EPXBGA0413QZMTF4XQ0M4Q3E         2.67GB ± 0%    2.22GB ± 0%  -16.99%
TelemeterRealData_Series/subs15s/01EPXBGA0413QZMTF4XQ0M4Q3E         631MB ± 0%     631MB ± 0%   +0.00%

name                                                             old allocs/op  new allocs/op  delta
[...]

PR #3705, opened 2021-01-07
Fix race condition in BinaryReader.LookupSymbol()

pracucci:

## Changes

The symbols cache introduced in https://github.com/thanos-io/thanos/pull/3557 also introduced a subtle race condition, because `LookupSymbol()` may be called concurrently.

In this PR I'm proposing to introduce a lock. With a lock it's obviously slower, but benchmarks show it may not be much significative compared to not having a cache at all (it's late, maybe I'm missing something, so be critic on this please).

### PR compared to `master`

```
benchmark                                                    old ns/op     new ns/op     delta
BenchmarkBinaryReader_LookupSymbol/num_series_=_1024-12      3288          13786         +319.28%
BenchmarkBinaryReader_LookupSymbol/num_series_=_10240-12     1850285       1996616       +7.91%

benchmark                                                    old allocs     new allocs     delta
BenchmarkBinaryReader_LookupSymbol/num_series_=_1024-12      0              0              +0.00%
BenchmarkBinaryReader_LookupSymbol/num_series_=_10240-12     10230          10230          +0.00%

benchmark                                                    old bytes     new bytes     delta
BenchmarkBinaryReader_LookupSymbol/num_series_=_1024-12      0             1             +Inf%
BenchmarkBinaryReader_LookupSymbol/num_series_=_10240-12     40576         40608         +0.08%
```

### Removing the symbols cache at all compared to `master`

```
benchmark                                                    old ns/op     new ns/op     delta
BenchmarkBinaryReader_LookupSymbol/num_series_=_1024-12      3288          179712        +5365.69%
BenchmarkBinaryReader_LookupSymbol/num_series_=_10240-12     1850285       1817189       -1.79%

benchmark                                                    old allocs     new allocs     delta
BenchmarkBinaryReader_LookupSymbol/num_series_=_1024-12      0              1014           +Inf%
BenchmarkBinaryReader_LookupSymbol/num_series_=_10240-12     10230          10230          +0.00%

benchmark                                                    old bytes     new bytes     delta
BenchmarkBinaryReader_LookupSymbol/num_series_=_1024-12      0             3179          +Inf%
BenchmarkBinaryReader_LookupSymbol/num_series_=_10240-12     40576         40572         -0.01%
```

## Verification

Unit tests / benchmark.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -12,6 +12,10 @@ We use _breaking :warning:_ to mark changes that are not backward compatible (re
 
 ## Unreleased
 
+### Changed
+
+- [#3705](https://github.com/thanos-io/thanos/pull/3705) Store: Fix race condition leading to failing queries or possibly incorrect query results.
+
 ## [v0.18.0](https://github.com/thanos-io/thanos/releases) - Release in progress
 
 ### Added
--- a/pkg/block/indexheader/binary_reader.go
+++ b/pkg/block/indexheader/binary_reader.go
@@ -648,12 +648,12 @@ func newBinaryTOCFromByteSlice(bs index.ByteSlice) (*BinaryTOC, error) {
 	}, nil
 }
 
-func (r BinaryReader) IndexVersion() (int, error) {
+func (r *BinaryReader) IndexVersion() (int, error) {
 	return r.indexVersion, nil
 }
 
 // TODO(bwplotka): Get advantage of multi value offset fetch.
-func (r BinaryReader) PostingsOffset(name string, value string) (index.Range, error) {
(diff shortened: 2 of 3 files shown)


bwplotka (approved):
Thanks for the nice benchmark and quick fix. I agree with you, introducing lock is not tragic. LGTM!

BTW you found it while reviewing, not on prod right?

On unrelated note, I am intrigued where we lost that byte: 

image

(((:

bwplotka:
What do you think @yeya24 ? Do you think we can merge? 🤗  Others? (:

pracucci:
> BTW you found it while reviewing

Yes, while reviewing.

PR #3707, opened 2021-01-08
Support bucket rewrite relabel

### LINKED ISSUE #3069: Tool: In-Place Relabelling for older data.
Totally doable request from @SuperQ:

> How possible would it be to write a tool that can basically run a series of metric_relabel_configs against a set of buckets? Basically, stop compaction, run relabel tool, start compaction again?
We're thinking about trying to cleanup some poor labeling choices on historical data. 

I would see this as `thanos tool <cmd>` that takes block ID (?) to recompact.

yeya24:

Fixes #3069 

Add relabel support to bucket rewrite.

## Changes

Add a `RelabelModifier` for blocks relabeling. Unlike the deletion modifier which is used in a streaming way, `RelabelModifier` cannot do this and the reasons are:

1. Relabeling might add new symbols so I have to modify the symbol set and sort it before returning it. Besides, symbols can be deleted because the whole series might be dropped via relabeling.
2. Series are not ordered as relabeling might change the order of series.

These requirements mean that I have to sort almost all the things in the constructor. It is inefficient but currently I don't find a better way to solve it.

## Verification

Unit tests added and manual testing.

### DIFF
--- a/pkg/compactv2/compactor_test.go
+++ b/pkg/compactv2/compactor_test.go
@@ -17,12 +17,15 @@ import (
 	"github.com/go-kit/kit/log"
 	"github.com/oklog/ulid"
 	"github.com/pkg/errors"
+	"github.com/prometheus/common/model"
 	"github.com/prometheus/prometheus/pkg/labels"
+	"github.com/prometheus/prometheus/pkg/relabel"
 	"github.com/prometheus/prometheus/tsdb"
 	"github.com/prometheus/prometheus/tsdb/chunkenc"
 	"github.com/prometheus/prometheus/tsdb/chunks"
 	"github.com/prometheus/prometheus/tsdb/index"
 	"github.com/prometheus/prometheus/tsdb/tombstones"
+
 	"github.com/thanos-io/thanos/pkg/block"
 	"github.com/thanos-io/thanos/pkg/block/metadata"
 	"github.com/thanos-io/thanos/pkg/testutil"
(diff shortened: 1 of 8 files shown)


pkg/compactv2/modifiers.go
```diff
@@ -326,4 +330,103 @@ func (p *delChunkSeriesIterator) Next() bool {
...
-// TODO(bwplotka): Add relabelling.
+type RelabelModifier struct {
+	relabels []*relabel.Config
+}
+
+func WithRelabelModifier(relabels ...*relabel.Config) *RelabelModifier {
+	return &RelabelModifier{relabels: relabels}
+}
+
+func (d *RelabelModifier) Modify(_ index.StringIter, set storage.ChunkSeriesSet, log ChangeLogger, p ProgressLogger) (index.StringIter, storage.ChunkSeriesSet) {
```
yeya24, line 341: I feel there are some bugs in my Modifier implementation. But Idk why.
The weird thing is that after changing the reference in pr https://github.com/thanos-io/thanos/pull/3710, it works. But ideally the existing `deletion modifier` works even without that pr.

pkg/compactv2/modifiers.go
```diff
@@ -328,4 +333,105 @@ func (p *delChunkSeriesIterator) Next() bool {
...
+			}
+
+			var deleted tombstones.Intervals
+			// If minTime is set then there is at least one chunk.
+			if minT != math.MaxInt64 {
+				deleted = deleted.Add(tombstones.Interval{Mint: minT, Maxt: maxT})
+			}
+			log.DeleteSeries(lbls, deleted)
+			p.SeriesProcessed()
+			continue SeriesLoop
```
yashrsharma44, line 384: We don't need to use `continue SeriesLoop` because the else part won't be executed.

yeya24: Oh, good catch!

yeya24:
I will close this pr as I cannot fix the test case without https://github.com/thanos-io/thanos/pull/3710.
If anyone wants to continue this feature, feel free to pick it up.

yeya24:
After some investigation, I found the root cause and I am able to fix the problem. Reopen this pr and I will work on it next week.

yeya24:
This pr should be ready for review. 
Note: Right now it only works for raw blocks. Not for downsampled blocks. I guess that's the same for `bucket rewrite delete` command.

pkg/compactv2/modifiers.go
yeya24: Solved. The root cause is the lazy population of the chunk and buf chunks reuse.

pkg/compactv2/compactor_test.go
```diff
@@ -397,6 +400,192 @@ func TestCompactor_WriteSeries_e2e(t *testing.T) {
...
+			name: "1 block + relabel modifier, series merged after relabeling",
+			input: [][]seriesSamples{
+				{
+					{lset: labels.Labels{{Name: "a", Value: "1"}},
+						chunks: [][]sample{{{1, 1}, {2, 2}, {10, 10}, {20, 20}}}},
+					{lset: labels.Labels{{Name: "a", Value: "2"}},
+						chunks: [][]sample{{{0, 0}, {2, 2}, {3, 3}}, {{4, 4}, {11, 11}, {20, 20}, {25, 25}}}},
+				},
+			},
+			// Drop all label name "a".
```
onprem, line 568: You missed modifying this comment. Maybe you meant something like 
```suggestion
			// Replace values of label name "a" with "0".
```

onprem (dismissed):
LGTM. Great work! Just a small nit.

pkg/compactv2/compactor_test.go
yeya24: Nice. Thanks for catching this!

onprem (approved):
!

onprem:
@yeya24 Maybe let's add a CHANGELOG entry for this before merge. This is too awesome to not have a CHANGELOG :smile:

onprem (approved):
LGTM. Thanks!

PR #3708, opened 2021-01-08
Upgrade Cortex and remove SwiftConfig

pracucci:

## Changes

Today we upgraded Thanos in Cortex, so in this PR I'm upgrading Cortex in Thanos and removing the `swift.SwiftConfig` which is no more required (because we moved to `swift.Config` in Cortex too).

## Verification

Existing tests.

### DIFF
--- a/cmd/thanos/query_frontend.go
+++ b/cmd/thanos/query_frontend.go
@@ -193,7 +193,7 @@ func runQueryFrontend(
 	roundTripper = tripperWare(roundTripper)
 
 	// Create the query frontend transport.
-	handler := transport.NewHandler(*cfg.CortexHandlerConfig, roundTripper, logger)
+	handler := transport.NewHandler(*cfg.CortexHandlerConfig, roundTripper, logger, nil)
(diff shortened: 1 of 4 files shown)


cmd/thanos/query_frontend.go
```diff
@@ -193,7 +193,7 @@ func runQueryFrontend(
 	roundTripper = tripperWare(roundTripper)
 
 	// Create the query frontend transport.
-	handler := transport.NewHandler(*cfg.CortexHandlerConfig, roundTripper, logger)
+	handler := transport.NewHandler(*cfg.CortexHandlerConfig, roundTripper, logger, nil)
```
pracucci, line 196: No need to pass the Prometheus registerer because it's just used for query statistics, which is currently a Cortex specific thing.

yeya24 (approved):
LGTM

GiedriusS (approved):
Thanks! AFAICT this should also improve compaction's performance due to https://github.com/prometheus/prometheus/commit/c8f810083d3f.

PR #3710, opened 2021-01-09
Fix rewrite chunk reuse reference

yeya24:

Fix buf chunks reuse. 

``` go
s.bufChks[i].Chunk = &lazyPopulatableChunk{cr: s.sReader.cr, m: &s.bufChks[i]}
```

Here `m` is a reference to the buffered chunk, we should use a copy instead.
I can reproduce this problem while adding tests for https://github.com/thanos-io/thanos/pull/3707.

### DIFF
--- a/pkg/compactv2/chunk_series_set.go
+++ b/pkg/compactv2/chunk_series_set.go
@@ -47,16 +47,17 @@ func (s *lazyPopulateChunkSeriesSet) Next() bool {
 			continue
 		}
 
+		chks := make([]chunks.Meta, len(s.bufChks))
+		copy(chks, s.bufChks)
 		for i := range s.bufChks {
-			s.bufChks[i].Chunk = &lazyPopulatableChunk{cr: s.sReader.cr, m: &s.bufChks[i]}
+			chks[i].Chunk = &lazyPopulatableChunk{cr: s.sReader.cr, m: &chks[i]}
 		}
 		s.curr = &storage.ChunkSeriesEntry{
 			Lset: make(labels.Labels, len(s.bufLbls)),
 			ChunkIteratorFn: func() chunks.Iterator {
-				return storage.NewListChunkSeriesIterator(s.bufChks...)
+				return storage.NewListChunkSeriesIterator(chks...)
 			},
 		}
-		// TODO: Do we need to copy this?
 		copy(s.curr.Lset, s.bufLbls)
 		return true
 	}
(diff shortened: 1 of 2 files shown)


pkg/compactv2/chunk_series_set.go
```diff
@@ -47,16 +47,17 @@ func (s *lazyPopulateChunkSeriesSet) Next() bool {
 			continue
 		}
 
+		chks := make([]chunks.Meta, len(s.bufChks))
```
bwplotka, line 50: Nice, so this is allowing us to not care about locking, but instead allocate more mem, right? 🤗 

How benchmarks are looking with this?

bwplotka (commented):
Thanks!

pkg/compactv2/chunk_series_set.go
yeya24: Sorry I didn't get the point. How is it related to locking? Did you mean adding a lock to protect `s.bufChks`?

I haven't benchmarked this. For a one-off job, memory seems not the highest priority. Maybe we can call GC manually somewhere if the mem usage is too high?

I am not sure if we can fix this without allocating additional memory. What do you think?

bwplotka: I mean, no lock is needed because we copy the slice?

bwplotka: let's talk offline (:

yeya24: Sorry but I think I missed something. Is `s.bufChks` accessed by multiple goroutines? I thought it is accessed sequentially.

GiedriusS: Have you figured this out offline?

yeya24: Not yet. :rofl:

bwplotka: ;p

bwplotka: Let's restart this. Can you explain what problem are you solving? you mentioned in the description that you are fixing some problem, which one? (: 

Just creating slices everywhere is not the best idea if we care about efficiency here, we never seen any error around this place on production too.

yeya24: I cannot tell tbh. But with this change the relabel pr works. I guess it is the problem of that pr because the rewrite delete works perfect.
But cannot figure out why.

PR #3729, opened 2021-01-18
Block Viewer: Move overlapping blocks to separate rows

### LINKED ISSUE #3709: Block Viewer: Overlapping blocks are not selectable/clickable
In the new Block Viewer, the overlapping blocks with same resolution and compaction level are rendered over each other. This means that you can only select the top-most block to view it's information. This is a major issue because now debugging this becomes significantly harder as you can't see the ULIDs of all of the overlapping blocks.

Screenshot from 2021-01-08 14-02-49

In the old UI, overlapping blocks were put in separate rows, doing this in the new UI would make sense as well.

Screenshot from 2021-01-08 14-01-31

Namanl2001:
Fixes #3709 

# Before changes
## Classic UI
classic
## New UI
before
# After changes
## New UI
Now each block is clickable/selectable. 
after

## Verification
Unit tests added.

### DIFF
--- a/pkg/ui/react-app/src/thanos/pages/blocks/helpers.test.tsx
+++ b/pkg/ui/react-app/src/thanos/pages/blocks/helpers.test.tsx
@@ -1,7 +1,7 @@
 import { sortBlocks, isOverlapping } from './helpers';
 
-// number of blocks in data: 8.
-const data = {
+// Number of blocks in data: 8.
+const overlapCaseData = {
   blocks: [
     {
       compaction: {
@@ -200,7 +200,7 @@ const data = {
   label: 'monitor',
 };
 
-const sorted = sortBlocks(data.blocks, data.label);
+const sorted = sortBlocks(overlapCaseData.blocks, overlapCaseData.label);
 const source = 'prometheus_one';
 
 describe('overlapping blocks', () => {
@@ -239,7 +239,7 @@ describe('overlapping blocks', () => {
 });
 
 describe('isOverlapping helper', () => {
-  const b = data.blocks[0];
+  const b = overlapCaseData.blocks[0];
   it('should return true for perfectly overlapping blocks', () => {
     expect(isOverlapping({ ...b, minTime: 10, maxTime: 20 }, { ...b, minTime: 10, maxTime: 20 })).toBe(true);
   });
--- a/pkg/ui/react-app/src/thanos/pages/blocks/helpers.test.tsx
+++ b/pkg/ui/react-app/src/thanos/pages/blocks/helpers.test.tsx
@@ -1,0 +1,258 @@
+import { sortBlocks, isOverlapping } from './helpers';
+
+// number of blocks in data: 8.
+const data = {
+  blocks: [
+    {
+      compaction: {
+        level: 1,
+        sources: ['01EWZCKPP4K0WYRTZC9RPRM5QK'],
+      },
+      minTime: 1608034200000,
+      maxTime: 1608034500000,
+      stats: {
+        numSamples: 6634538,
+        numSeries: 2334,
+        numChunks: 51057,
+      },
+      thanos: {
+        downsample: {
+          resolution: 0,
+        },
+        labels: {
+          monitor: 'prometheus_one',
+        },
+        source: 'sidecar',
+      },
+      ulid: '01EWZCKPP4K0WYRTZC9RPRM5QK',
+      version: 1,
+    },
+
+    {
+      compaction: {
+        level: 1,
+        sources: ['01ESK5B1WQB6QEZQ4P0YCQXEC4'],
+      },
+      minTime: 1608034200000,
+      maxTime: 1608034500000,
+      stats: {
+        numSamples: 6634538,
+        numSeries: 2334,
+        numChunks: 51057,
+      },
+      thanos: {
+        downsample: {
+          resolution: 0,
+        },
+        labels: {
+          monitor: 'prometheus_one',
+        },
+        source: 'sidecar',
+      },
+      ulid: '01ESK5B1WQB6QEZQ4P0YCQXEC4',
+      version: 1,
+    },
+    {
+      compaction: {
+        level: 1,
+        sources: ['01ET8F8C73GGXH279R6YMTWFHY'],
+      },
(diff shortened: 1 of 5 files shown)


pkg/ui/react-app/src/thanos/pages/blocks/helpers.ts
```diff
@@ -10,14 +10,19 @@ const stringify = (map: LabelSet): string => {
 
 const sortBlocksInRows = (blocks: Block[]): BlocksPool => {
   const pool: BlocksPool = {};
+  const map = new Map(); //map to keep track of similar blocks.
```
bwplotka, line 13:
```suggestion

  // Map to keep track of similar blocks.
  const map = new Map();
```

bwplotka (approved):
Looks good to me 🤗 

@prmsrswt ?

pkg/ui/react-app/src/thanos/pages/blocks/helpers.ts
```diff
@@ -10,14 +10,19 @@ const stringify = (map: LabelSet): string => {
...
-        pool[`${b.compaction.level}-${b.thanos.downsample.resolution}`] = [];
+      if (!map.has(`${b.maxTime}-${b.minTime}`)) map.set(`${b.maxTime}-${b.minTime}`, 0);
+      else map.set(`${b.maxTime}-${b.minTime}`, map.get(`${b.maxTime}-${b.minTime}`) + 1);
+      const count = map.get(`${b.maxTime}-${b.minTime}`);
 
-      pool[`${b.compaction.level}-${b.thanos.downsample.resolution}`].push(b);
+      if (!pool[`${b.compaction.level}-${b.thanos.downsample.resolution}-${count}`])
+        pool[`${b.compaction.level}-${b.thanos.downsample.resolution}-${count}`] = [];
+
+      pool[`${b.compaction.level}-${b.thanos.downsample.resolution}-${count}`].push(b);
```
prmsrswt, line 25: This will only work when we have perfectly overlapping blocks, i.e. when both the `minTime` and `maxTime` is exactly same for these blocks. But we can have partially overlapping blocks as well. The condition we need to check is if a block's `minTime` is less than the `maxTime` of the previous block.

Namanl2001:
Capture

The latest commit checks for every possible case that might occur (as far as I've tested), do let me know if you find any case of overlapping blocks. Thanks.

PS: I shall run the `make assets` after the final approval of the changes.

cc: @prmsrswt

pkg/ui/react-app/src/thanos/pages/blocks/helpers.ts
```diff
@@ -10,14 +10,61 @@ const stringify = (map: LabelSet): string => {
...
+      } else {
+        const preMin = `${blocks[index - 1].minTime}`;
+        const preMax = `${blocks[index - 1].maxTime}`;
+        let nearMax: any;
+
+        maxTimeCount.forEach((v, k) => {
+          if (k <= currMin) nearMax = k;
+        });
+
+        if (currMin < preMax) {
```
bwplotka, line 40: Nice work! 

I wonder if it has to be so many if/elses =D

Good for a start I guess (: Not a blocker - just bit complex, code maybe there is a way simplify

bwplotka (approved):
Thanks!

pkg/ui/react-app/src/thanos/pages/blocks/helpers.ts
Namanl2001: It's true that it is a bit complex or we can say that is the brute force approach for the problem but I tried and didn't find any optimal approach for it. If anyone could suggest any better/optimized approach, I would be happy to implement it otherwise I think we can merge it for now after making a separate function for the if-else ladder as suggested by @prmsrswt. Thanks!

bwplotka: Tip: Recursion of function helpers (:

bwplotka: If we do `isOverlapping` function would be nice (:

pkg/ui/react-app/src/thanos/pages/blocks/helpers.ts
```diff
@@ -10,14 +10,61 @@ const stringify = (map: LabelSet): string => {
 
 const sortBlocksInRows = (blocks: Block[]): BlocksPool => {
   const pool: BlocksPool = {};
+  const minTimeCount = new Map();
+  const maxTimeCount = new Map();
+  const maxToMinTime = new Map();
 
   blocks
```
bwplotka, line 17: Can we move all of this to the function and add unit tests for it? (:

bwplotka: BTW: We care about readability and testability.

bwplotka (changes requested):
LGTM generally! (: 

Would be amazing to have a unit test for it. Then we can work easier with refactoring! 🤗

Namanl2001:
Created a new `isOverlapping` function and added unit tests for the following scenario:

test1

pkg/ui/react-app/src/thanos/pages/blocks/helpers.test.tsx
```diff
@@ -0,0 +1,258 @@
+import { sortBlocks, isOverlapping } from './helpers';
+
+// number of blocks in data: 8.
+const data = {
```
bwplotka, line 4:
```suggestion
const overlapCaseData = {
```

pkg/ui/react-app/src/thanos/pages/blocks/helpers.test.tsx
```diff
@@ -0,0 +1,258 @@
+import { sortBlocks, isOverlapping } from './helpers';
+
+// number of blocks in data: 8.
```
bwplotka, line 3:
```suggestion
// Number of blocks in data: 8.
```

bwplotka (approved):
LGTM! 🤗

prmsrswt (approved):
LGTM, Thanks!

yeya24:
Thanks for the good work! @Namanl2001

PR #3732, opened 2021-01-21
pkg/rules/proxy: fix hotlooping when receiving client errors

### LINKED ISSUE #3717: Thanos rules fanout OOMs occasionally
Thanos 0.15.0
Prometheus 2.22.2
Golang 1.15

**What happened**: When invoking the `/v1/api/rules` endpoint, thanos querier OOMs

**What you expected to happen**: It doesn't OOM

**How to reproduce it (as minimally and precisely as possible)**: So far we have no determinstic reproduction. What we know is that the OOM happens most likely if errors occur during the proxy fanout. This is backed by pprof heap snapshots we gathered during the OOM:
image

We see ~3GB used memory in the callstack collector of pkg/errors: https://github.com/pkg/errors/blob/5dd12d0cfe7f152f80558d591504ce685299311e/stack.go#L163-L169

I don't have a log file at hand but will request one. Previous logs didn't reveal any errors (rules proxy errors are not being logged).

**Environment**:
- OS (e.g. from /etc/os-release): RHOS
- Kernel (e.g. `uname -a`): Unknown
- Others: OpenShift 4.6.8

cc @mbarnes @gsleeman @mrbarge

s-urbaniak:
Currently, if we receive an error from the underlying client stream,
we continue with trying to receive additional data.
This causes a hotloop as we will receive the same error again.
This fixes it by returning in the error case and adds a unit test for the proxy logic.

Fixes #3717

## Verification

Added unit tests, without the fix the unit test will hotloop.

cc @bwplotka @kakkoyun @squat

### DIFF
--- a/pkg/rules/proxy.go
+++ b/pkg/rules/proxy.go
@@ -108,6 +108,9 @@ func (stream *rulesStream) receive(ctx context.Context) error {
 		}
 
 		if err != nil {
+			// An error happened in Recv(), hence the underlying stream is aborted
+			// as per https://github.com/grpc/grpc-go/blob/7f2581f910fc21497091c4109b56d310276fc943/stream.go#L117-L125.
+			// We must not continue receiving additional data from it and must return.
 			err = errors.Wrapf(err, "receiving rules from rules client %v", stream.client)
 
 			if stream.request.PartialResponseStrategy == storepb.PartialResponseStrategy_ABORT {
@@ -118,13 +121,15 @@ func (stream *rulesStream) receive(ctx context.Context) error {
 				return errors.Wrapf(err, "sending rules error to server %v", stream.server)
 			}
 
-			continue
+			// Return no error if response strategy is warning.
+			return nil
 		}
 
 		if w := rule.GetWarning(); w != "" {
 			if err := stream.server.Send(rulespb.NewWarningRulesResponse(errors.New(w))); err != nil {
 				return errors.Wrapf(err, "sending rules warning to server %v", stream.server)
 			}
+			// Client stream is not aborted, it is ok to receive additional data.
 			continue
 		}
 
--- a/pkg/rules/proxy_test.go
+++ b/pkg/rules/proxy_test.go
@@ -1,0 +1,253 @@
+// Copyright (c) The Thanos Authors.
+// Licensed under the Apache License 2.0.
+
+package rules
+
+import (
+	"context"
+	"io"
+	"os"
+	"reflect"
+	"testing"
+
+	"github.com/go-kit/kit/log"
+	"github.com/pkg/errors"
+	"github.com/thanos-io/thanos/pkg/rules/rulespb"
+	"github.com/thanos-io/thanos/pkg/store/storepb"
+	"google.golang.org/grpc"
+)
+
+type testRulesClient struct {
+	grpc.ClientStream
+	rulesErr, recvErr error
+	response          *rulespb.RulesResponse
+	sentResponse      bool
+}
+
+func (t *testRulesClient) String() string {
+	return "test"
+}
+
+func (t *testRulesClient) Recv() (*rulespb.RulesResponse, error) {
+	// A simulation of underlying grpc Recv behavior as per https://github.com/grpc/grpc-go/blob/7f2581f910fc21497091c4109b56d310276fc943/stream.go#L117-L125.
+	if t.recvErr != nil {
+		return nil, t.recvErr
+	}
+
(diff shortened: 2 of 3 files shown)


squat (approved):
looks great @s-urbaniak  :)

prmsrswt (approved):
LGTM!

prmsrswt:
@squat Does it makes sense to get this fix released with `18.0`?

squat:
yes, it very much could make sense. In this case, @s-urbaniak, I would ask that you make the pr against the release-0.18 branch instead. We would later merge this into main.

pkg/rules/proxy.go
```diff
@@ -118,13 +121,15 @@ func (stream *rulesStream) receive(ctx context.Context) error {
...
-			continue
+			// Return no error if response strategy is warning.
+			return nil
 		}
 
 		if w := rule.GetWarning(); w != "" {
 			if err := stream.server.Send(rulespb.NewWarningRulesResponse(errors.New(w))); err != nil {
 				return errors.Wrapf(err, "sending rules warning to server %v", stream.server)
 			}
+			// client stream is not aborted, it is ok to receive additional data.
```
bwplotka, line 132:
```suggestion
			// Client stream is not aborted, it is ok to receive additional data.
```

bwplotka (commented):
Amazing, let's do same on `release-0.18`?

pkg/rules/proxy_test.go
```diff
@@ -0,0 +1,296 @@
...
+	response          *rulespb.RulesResponse
+	sentResponse      bool
+}
+
+func (t *testRulesClient) String() string {
+	return "test"
+}
+
+func (t *testRulesClient) Recv() (*rulespb.RulesResponse, error) {
+	// a naive simulation of underlying grpc Recv behavior as per https://github.com/grpc/grpc-go/blob/7f2581f910fc21497091c4109b56d310276fc943/stream.go#L117-L125.
```
bwplotka, line 32:
```suggestion
	// A simulation of underlying grpc Recv behavior as per https://github.com/grpc/grpc-go/blob/7f2581f910fc21497091c4109b56d310276fc943/stream.go#L117-L125.
```

pkg/rules/proxy_test.go
```diff
@@ -0,0 +1,296 @@
...
+
+	if t.sentResponse {
+		return nil, io.EOF
+	}
+	t.sentResponse = true
+
+	return t.response, nil
+}
+
+func (t *testRulesClient) Header() (metadata.MD, error) {
```
bwplotka, line 45: No need for this. Just implement what you want and embed `RulerClient` grpc interface and don't pas anything into it 😉

pkg/rules/proxy_test.go
```diff
@@ -0,0 +1,296 @@
...
+
+func (t *testRulesServer) Context() context.Context {
+	return context.Background()
+}
+
+func (t *testRulesServer) SendMsg(m interface{}) error {
+	panic("not implemented")
+}
+
+func (t *testRulesServer) RecvMsg(m interface{}) error {
```
bwplotka, line 112: ditto

bwplotka (commented):
Some extra tips

pkg/rules/proxy.go
s-urbaniak: fixed

pkg/rules/proxy_test.go
s-urbaniak: fixed

pkg/rules/proxy_test.go
s-urbaniak: ah, yes, great catch, fixed :+1:

s-urbaniak:
@bwplotka thank you for the review! :green_heart: PTAL for final review :)

s-urbaniak:
Submitted https://github.com/thanos-io/thanos/pull/3734 for the 0.18 backport.

squat:
Mmm rather than doing this in both branches, we should do what we always do which is merge in the release branch and then merge the release branch into the master branch no? If we all agree then I'll close this PR

s-urbaniak:
leaving up to the maintainers, i have both PRs open :-)

prmsrswt:
> Mmm rather than doing this in both branches, we should do what we always do which is merge in the release branch and then merge the release branch into the master branch no? If we all agree then I'll close this PR

@squat Yeah let's close this and merge the PR in release branch. I don't think we will get a conflict but let's avoid polluting the git history with duplicate changes.

PR #3734, opened 2021-01-21
pkg/rules/proxy: fix hotlooping when receiving client errors

### LINKED ISSUE #3717: Thanos rules fanout OOMs occasionally

s-urbaniak:
Currently, if we receive an error from the underlying client stream,
we continue with trying to receive additional data.
This causes a hotloop as we will receive the same error again.
This fixes it by returning in the error case and adds a unit test for the proxy logic.

Fixes #3717

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -12,6 +12,10 @@ We use _breaking :warning:_ to mark changes that are not backward compatible (re
 
 ## Unreleased
 
+### Fixed
+
+- [#3734](https://github.com/thanos-io/thanos/pull/3734) pkg/rules/proxy: fix hotlooping when receiving client errors
+
 ## [v0.18.0](https://github.com/thanos-io/thanos/releases/tag/v0.18.0) - 2021.01.19
 
 ### Added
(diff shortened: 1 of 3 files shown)


prmsrswt (approved):
LGTM (again :))

PR #3738, opened 2021-01-22
Fixed website and added new step for release process (tmp).

bwplotka:


squat:
yeah, saw this problem last week but I couldn't believe the process would so manual. this is definitely something we need to improve and do away with. Documenting it is a first step, but this has to be automated, if nothing else than because we need a more elegant solution.

kakkoyun (approved):
🥇 LGTM

bwplotka:
I know it's part of https://github.com/bwplotka/mdox but it's not easy as you think to automate =D

Issue #3651, opened 2020-12-19
api: Ability to set default step for queries.


bwplotka:
Default step does not make much sense if the scrape interval is common (e.g 15s). Getting too low makes things extremely slow for no good reason.

Flag on querier might be a good idea.

spaparaju:
Will pick up this issue @bwplotka

spaparaju:
I see Thanos Query got two possible ways to customise the downsample / resolution for queries:  

1. specify 'step' duration along with enabling 'enableAutodownsampling' to calculate 'max_source_resolution'
2. explicitly specify 'max_source_resolution'

I am trying to understand the purpose of two options to arrive at seemingly a single outcome: Query time downsampling. 
Is there an advantage of choosing one option over another?

bwplotka:
Sorry for not answering @spaparaju but I am quite confused why we speak about downsampling here. 

Looks like https://github.com/thanos-io/thanos/pull/3740 did most of the work, so let's see if we can merge this 🤗

PR #3740, opened 2021-01-24
Api: Add flag to set default step

hitanshu-mehta:

Fixes : #3651 

This PR Adds a new flag `query.default-step` which allows to set step for range queries. This cli flag will be overwritten when step is provided in request parameter. 

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -14,6 +14,7 @@ We use _breaking :warning:_ to mark changes that are not backward compatible (re
 
 ### Added
 
+- [#3740](https://github.com/thanos-io/thanos/pull/3740) Query: Added `--query.default-set` flag to set default step.
 - [#3700](https://github.com/thanos-io/thanos/pull/3700) ui: make old bucket viewer UI work with vanilla Prometheus blocks
 - [#2641](https://github.com/thanos-io/thanos/issues/2641) Query Frontend: Added `--query-range.request-downsampled` flag enabling additional queries for downsampled data in case of empty or incomplete response to range request.
 - [#3792](https://github.com/thanos-io/thanos/pull/3792) Receiver: Added `--tsdb.allow-overlapping-blocks` flag to allow overlapping tsdb blocks and enable vertical compaction
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -14,7 +14,7 @@ We use _breaking :warning:_ to mark changes that are not backward compatible (re
 
 ### Added
 
-- [#3740](https://github.com/thanos-io/thanos/pull/3740) Query: Added `--query.default-set` flag to set default step.
+- [#3740](https://github.com/thanos-io/thanos/pull/3740) Query: Added `--query.default-step` flag to set default step.
 - [#3700](https://github.com/thanos-io/thanos/pull/3700) ui: make old bucket viewer UI work with vanilla Prometheus blocks
 - [#2641](https://github.com/thanos-io/thanos/issues/2641) Query Frontend: Added `--query-range.request-downsampled` flag enabling additional queries for downsampled data in case of empty or incomplete response to range request.
 - [#3792](https://github.com/thanos-io/thanos/pull/3792) Receiver: Added `--tsdb.allow-overlapping-blocks` flag to allow overlapping tsdb blocks and enable vertical compaction
--- a/cmd/thanos/query.go
+++ b/cmd/thanos/query.go
@@ -125,6 +125,9 @@ func registerQuery(app *extkingpin.App) {
 
 	defaultEvaluationInterval := extkingpin.ModelDuration(cmd.Flag("query.default-evaluation-interval", "Set default evaluation interval for sub queries.").Default("1m"))
 
+	defaultRangeQueryStep := extkingpin.ModelDuration(cmd.Flag("query.default-step", "Set default step for range queries").
+		Default("1s"))
+
 	storeResponseTimeout := extkingpin.ModelDuration(cmd.Flag("store.response-timeout", "If a Store doesn't send any data in this specified duration then a Store will be ignored and partial data will be returned if it's enabled. 0 disables timeout.").Default("0ms"))
 
 	cmd.Setup(func(g *run.Group, logger log.Logger, reg *prometheus.Registry, tracer opentracing.Tracer, _ <-chan struct{}, _ bool) error {
@@ -181,6 +184,7 @@ func registerQuery(app *extkingpin.App) {
 			*webPrefixHeaderName,
 			*maxConcurrentQueries,
 			*maxConcurrentSelects,
+			time.Duration(*defaultRangeQueryStep),
 			time.Duration(*queryTimeout),
 			*lookbackDelta,
 			*dynamicLookbackDelta,
@@ -231,6 +235,7 @@ func runQuery(
 	webPrefixHeaderName string,
 	maxConcurrentQueries int,
 	maxConcurrentSelects int,
+	defaultRangeQueryStep time.Duration,
 	queryTimeout time.Duration,
 	lookbackDelta time.Duration,
 	dynamicLookbackDelta bool,
@@ -454,6 +459,7 @@ func runQuery(
 			enableRulePartialResponse,
 			queryReplicaLabels,
 			flagsMap,
+			defaultRangeQueryStep,
 			instantDefaultMaxSourceResolution,
 			defaultMetadataTimeRange,
 			gate.New(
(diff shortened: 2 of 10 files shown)


pkg/api/query/v1.go
```diff
@@ -224,6 +227,19 @@ func (qapi *QueryAPI) parsePartialResponseParam(r *http.Request, defaultEnablePa
 	return defaultEnablePartialResponse, nil
 }
 
+func (qapi *QueryAPI) parseStep(r *http.Request, defaultRangeQueryStep time.Duration) (step time.Duration, _ *api.ApiError) {
```
bwplotka, line 230: I think this does not fix the issue (but is a very good try!) (: 

The reason is that there is no default set by server. As you could see before your PR server was erroring out when step is not provided. I think that's ok to keep this is way.

To change default I think we need to alter frontend code - as this is where the default 1s is injected. That would be amazing to do! WDYT? (:

bwplotka (changes requested):
The flag looks good, and great work overall, but I have one suggestion 🤗 I think this flag should dicate the actual default in frontend, otherwise this flag has no use, because Grafana or Thanos frontend always passes step.

pkg/api/query/v1.go
hitanshu-mehta: Yes I agree 😊 . This flag is always going to be overwritten. So we need to provide some kind of input field on front-end using which user can set default step ? If yes, Can i work on that as well ?

bwplotka: Of course, feel free to use the same PR

cmd/thanos/query.go
```diff
@@ -125,6 +125,9 @@ func registerQuery(app *extkingpin.App) {
 
 	defaultEvaluationInterval := extkingpin.ModelDuration(cmd.Flag("query.default-evaluation-interval", "Set default evaluation interval for sub queries.").Default("1m"))
 
+	defaultRangeQueryStep := extkingpin.ModelDuration(cmd.Flag("query.default-step", "Set default step for range queries").
```
bwplotka, line 128: I would add more context. For example the fact, that this works well if step is not set and we made Thanos UI doing that. This will not work from Grafana, but Grafana has this `__step` variable which can be used (although not sure how close to this it works).

pkg/api/query/v1.go
```diff
@@ -224,6 +227,19 @@ func (qapi *QueryAPI) parsePartialResponseParam(r *http.Request, defaultEnablePa
...
+	// Overwrite the cli flag when provided as a query parameter.
+	if val := r.FormValue("step"); val != "" {
+		var err error
+		defaultRangeQueryStep, err = parseDuration(val)
+		if err != nil {
+			return 0, &api.ApiError{Typ: api.ErrorBadData, Err: errors.Wrapf(err, "'%s' parameter", defaultRangeQueryStep)}
+		}
+	}
+
+	return defaultRangeQueryStep, nil
```
bwplotka, line 240: Wonder would it makes sense to mimic UI here and do `min(range/250, default step)`, just for consistency with UI? I would love to have same default behaviours 🤗

bwplotka (approved):
LGTM

It's very neat you are using Flags API for it , love it! 🤗 

2 suggestions, otherwise rdy to go 💪🏽

cmd/thanos/query.go
hitanshu-mehta: Understood. I will write proper description of the flag.

pkg/api/query/v1.go
hitanshu-mehta: Sorry missed it 😅 . I'll correct it.

pkg/api/query/v1.go
```diff
@@ -224,6 +228,21 @@ func (qapi *QueryAPI) parsePartialResponseParam(r *http.Request, defaultEnablePa
...
+	// Overwrite the cli flag when provided as a query parameter.
+	if val := r.FormValue(Step); val != "" {
+		var err error
+		defaultRangeQueryStep, err = parseDuration(val)
+		if err != nil {
+			return 0, &api.ApiError{Typ: api.ErrorBadData, Err: errors.Wrapf(err, "'%s' parameter", Step)}
+		}
+	}
+
+	// Default step is used this way to make it consistent with UI
```
bwplotka, line 241:
```suggestion
	// Default step is used this way to make it consistent with UI.
```

pkg/api/query/v1_test.go
```diff
@@ -481,21 +482,34 @@ func TestQueryEndpoints(t *testing.T) {
 				},
 			},
 		},
-		// Missing query params in range queries.
+		// Use default step when missing
```
bwplotka, line 485:
```suggestion
		// Use default step when missing.
```

bwplotka (approved):
LGTM, just small nits (:

bwplotka:
Concern: @prmsrswt that we break Flags API flag name, so we should have this API to be stable

* Test suite for this?
* Another API? 
* Default value to UI!

hitanshu-mehta:
@prmsrswt @bwplotka Just for visibility. As discussed in contributors office hours going with default value in UI.

hitanshu-mehta:
I have done suggested changes. @bwplotka @prmsrswt Can you please have a look ? Thank you :)

bwplotka (approved):
LGTM! 💪🏽

bwplotka:
Let's rebase 🙃

hitanshu-mehta:
Thank you for the review :)

hitanshu-mehta:
Rebased 🚀

pkg/api/query/v1.go
```diff
@@ -225,6 +229,21 @@ func (qapi *QueryAPI) parsePartialResponseParam(r *http.Request, defaultEnablePa
 	return defaultEnablePartialResponse, nil
 }
 
+func (qapi *QueryAPI) parseStep(r *http.Request, defaultRangeQueryStep time.Duration, rangeSeconds int64) (step time.Duration, _ *api.ApiError) {
```
prmsrswt, line 232:
```suggestion
func (qapi *QueryAPI) parseStep(r *http.Request, defaultRangeQueryStep time.Duration, rangeSeconds int64) (time.Duration, *api.ApiError) {
```

Let's not use named returns if we are not going to use the variables.

pkg/ui/react-app/src/pages/graph/PanelList.tsx
```diff
@@ -139,6 +143,10 @@ const PanelList: FC<RouteComponentProps & PathPrefixProps> = ({ pathPrefix = ''
   const { response: storesRes, error: storesErr, isLoading: storesLoading } = useFetch<StoreListProps>(
     `${pathPrefix}/api/v1/stores`
   );
+  const { response: flagsRes, error: flagsErr, isLoading: flagsLoading } = useFetch<FlagMap>(
+    `${pathPrefix}/api/v1/status/flags`
+  );
+  const defaultStep = flagsRes.data && 'query.default-step' in flagsRes.data ? flagsRes.data['query.default-step'] : '1s';
```
prmsrswt, line 149:
```suggestion
  const defaultStep = flagsRes?.data?.['query.default-step'] || '1s';
```

This can be simplified by using Optional Chaining. https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/Optional_chaining

prmsrswt (commented):
Just a few nits, otherwise LGTM. Thanks.

You might need another rebase though (:

pkg/ui/react-app/src/pages/graph/PanelList.tsx
hitanshu-mehta: Much cleaner 😃

prmsrswt (approved):
One last thing. IMO a changelog entry for this makes sense.

hitanshu-mehta:
Added to changelog. Ready to merge 🚀

CHANGELOG.md
```diff
@@ -14,6 +14,7 @@ We use _breaking :warning:_ to mark changes that are not backward compatible (re
 
 ### Added
 
+- [#3740](https://github.com/thanos-io/thanos/pull/3740) Query: Added `--query.default-set` flag to set default step.
```
yeya24, line 17: @hitanshu-mehta A small typo. The flag is `query.default-step` instead of `query.default-set`. After modifying this then we are good to go!

yeya24 (changes requested):

CHANGELOG.md
hitanshu-mehta: Opps. Thank you for pointing out 😄

yeya24 (approved):
LGTM

hitanshu-mehta:
Thank you very much everyone for the review :)

PR #3743, opened 2021-01-25
Create a remote-write Appender for a stateless Ruler

metalmatze:
Still WIP :)
I need to pass down the configuration and reuse a proper HTTP client. But overall it works and I can already remote-write into receivers.

### DIFF
--- a/cmd/thanos/rule.go
+++ b/cmd/thanos/rule.go
@@ -27,17 +27,18 @@ import (
 	"github.com/prometheus/prometheus/pkg/labels"
 	"github.com/prometheus/prometheus/promql"
 	"github.com/prometheus/prometheus/rules"
+	"github.com/prometheus/prometheus/storage"
 	"github.com/prometheus/prometheus/tsdb"
 	"github.com/prometheus/prometheus/util/strutil"
-	"github.com/thanos-io/thanos/pkg/errutil"
-	"github.com/thanos-io/thanos/pkg/extkingpin"
 
 	"github.com/thanos-io/thanos/pkg/alert"
 	v1 "github.com/thanos-io/thanos/pkg/api/rule"
 	"github.com/thanos-io/thanos/pkg/block/metadata"
 	"github.com/thanos-io/thanos/pkg/component"
 	"github.com/thanos-io/thanos/pkg/discovery/dns"
+	"github.com/thanos-io/thanos/pkg/errutil"
 	"github.com/thanos-io/thanos/pkg/extflag"
+	"github.com/thanos-io/thanos/pkg/extkingpin"
 	"github.com/thanos-io/thanos/pkg/extprom"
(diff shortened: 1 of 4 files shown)


metalmatze:
What do people think would be the minimum required to merge this so that the student working on this get move forward with this? @kakkoyun @squat @bwplotka

kakkoyun:
Hey @metalmatze, sorry for the late response. We haven't had enough time to check this out. I'll do my best to have a look at it this week.

metalmatze:
Sure, let me know if you have any questions or what to pair on it.

bwplotka (commented):
Some discussion we have on our 1:2 with @squat and @idoqo.

I think this is a solid start, but I believe we might want to have a more resilient approach where we append to WAL and only then tail WAL and remote write, exactly like Grafana Agent or planned Prometheus agent is doing. 

The problem with the approach in this PR is that every append sends a remote write request. This might be a problem, because:
1. We have one job that evaluates and then send samples synchronously. This will block evaluation if the target is unavailable.
2. We don't persist samples, so if target is unavailable for longer time, we don't have a buffer
3. We cannot use any smarter sharding to combine multiple series into one remote write request.

I wonder if it would be not a better approach to reuse Grafana Cloud or planned Prometheus agent code and decouple evaluation from sending behaviour. So we do those asynchronously with the WAL as a buffer 🤗

bwplotka:
Code we could reuse: https://sourcegraph.com/github.com/grafana/agent@226268aa96a69210fca3dc66036dc4524b4ae1b1/-/blob/pkg/prom/wal/wal.go#L122:6

bwplotka:
Also, we could start with just remote write on top of TSDB, that would be one solution for a start. But we can try directly with no local TSDB approach too.

brancz:
I would start with no persistence for a start. The problematic thing about persistence in the rule evaluation scenario is that you don't know how much data a rule evaluation is going to produce, so by definition the "scalability" of a local storage (even just a WAL) would not be sufficient. I think the better strategy for resiliency is to record when the last evaluation was and "back-fill". After all, even if a ruler is up, but the storage is down, then the data the ruler is going to query is likely stale or non existant anyways.

bwplotka:
Fair point, cc @idoqo

The main problem with this approach is that every rule/alert eval produces a remote write request with a single sample. This is not the best idea overall within our hashing/forwarding approach. We can quickly saturate connections/CPU in larger setups.

brancz:
Without diving into the rule engine in Prometheus I think commit is only called per rule-evaluation, not per sample, right? If so, then the remote-write request would be per rule-evaluation as well. I think that would also be sufficient for a start, after all, this is how Prometheus is going to be issuing remote-write requests soon as well (commit-based).

m-messiah:
Hello! remote-write from a ruler would be a great thing, let's reopen this PR and continue?

PR #3752, opened 2021-01-27
Introduce "block-meta-fetch-concurrency" flag for compact and store

johannesfrey:
Signed-off-by: Johannes Frey <johannes.frey@daimler.com>

## Changes

This PR makes the meta fetcher concurrency configurable for the end user by introducing a corresponding flag for `compact` and `store`. Furthermore it moves the fetcherConcurrency const close to the fetcher code of the block package.

The reason for this PR is to have a way to control how many goroutines are started in parallel to fetch meta data from object storage. Especially with S3 this leads to many parallel HEAD requests which might provoke rate-limiting. This mostly affects `store` as there currently happens an initial sync which does not use memcached.

Furthermore, it seems that the flag `block-sync-concurrency` of `compact` is passed but not used in the `Syncer`. So I'm actually not sure where this value should be used or if this PR would then duplicate the initial intent of that flag :slightly_smiling_face:.

Of course this introduces "yet another flag", so I'm happy for feedback especially if there might be a good reason to not expose this and use a fixed 32.

## Verification

I built the image and deployed `compact` and `store` to a Kubernetes cluster.

<sub>Johannes Frey [johannes.frey@daimler.com](mailto:johannes.frey@daimler.com) Daimler TSS GmbH ([Imprint](https://github.com/Daimler/daimler-foss/blob/master/LEGAL_IMPRINT.md))</sub>

### DIFF
--- a/cmd/thanos/compact.go
+++ b/cmd/thanos/compact.go
@@ -201,10 +201,10 @@ func runCompact(
 	// While fetching blocks, we filter out blocks that were marked for deletion by using IgnoreDeletionMarkFilter.
 	// The delay of deleteDelay/2 is added to ensure we fetch blocks that are meant to be deleted but do not have a replacement yet.
 	// This is to make sure compactor will not accidentally perform compactions with gap instead.
-	ignoreDeletionMarkFilter := block.NewIgnoreDeletionMarkFilter(logger, bkt, deleteDelay/2, fetcherConcurrency)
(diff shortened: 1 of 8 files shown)


bwplotka (approved):
Make sense. Thank you! I wonder if you find more optimal default number 🤗

bwplotka:
This is fine, and the PR is clean so let's go 🚀

bwplotka:
IF we are worried, we could make this flag `Hidden` but not a big deal. We are still in 0.x.y version.

Issue #3744, opened 2021-01-25
sidecar: Allow sidecar to have write (no delete) access: Avoid deleting on failed upload.


bwplotka:
Ideally, we don't want to give delete access to sidecar. And we don't need to delete partial uploads anymore as they are handled well on compactor/store gw sides.

Discussion https://cloud-native.slack.com/archives/CL25937SP/p1611572804008400

AC:
 * We can run a sidecar without Deletion Permission.

SuperQ:
Yes, right now we need to give Object Admin level access to the sidecar, rather than just Create and Read.

mayankshah1607:
Hi @bwplotka 
I'd love to take up this issue and make my first contribution here!

mayankshah1607:
Just a very quick question - how many times should the upload be retried, and for how long?

mayankshah1607:
I have opened a PR - https://github.com/thanos-io/thanos/pull/3756*
Not sure if this would be the right approach, would love to have some feedback :)

bwplotka:
Looks like somehow we have two PRs for the same thing (: Which should we review? Sorry for the lag in reviewing.

Let's maybe first have PR that removes deleting (only for sidecar), that would be the first step, However, both PRs @Biswajitghosh98 and @mayankshah1607 are about retries which are quite controversial.

Sorry, I put retries in the title - I rather meant delete the deletion action for the sidecar and the sidecar will retry automatically on crash. Let's start there.

Improving retries is not as easy as we think, created separate issue for this: https://github.com/thanos-io/thanos/issues/3907

PR #3756, opened 2021-02-01
WIP: sidecar: avoid deleting partial uploads; retry instead

mayankshah1607:
Fixes https://github.com/thanos-io/thanos/issues/3744

- removed deletion of partially uploaded blocks from the sidecar.
- upload is retried upon failure

Signed-off-by: Mayank Shah <mayankshah1614@gmail.com>

### DIFF
--- a/pkg/block/block.go
+++ b/pkg/block/block.go
@@ -103,16 +103,22 @@ func Upload(ctx context.Context, logger log.Logger, bkt objstore.Bucket, bdir st
 		return errors.Wrap(err, "encode meta file")
 	}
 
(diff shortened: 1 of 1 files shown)


bwplotka (changes requested):
Sorry for missing this PR and thank you for your contribution!

Code looks good, however I think I mislead you on the issue https://github.com/thanos-io/thanos/issues/3744#issuecomment-795906349

Added comment to explain, all. This ticket is mostly uploaded removing the need for deletions on the sidecar, not the retry logic. Sidecar will retry in the next cycle so we don't need this. In fact such retries on this level are not recommended, see https://github.com/thanos-io/thanos/issues/3907 for bigger discussion. 🤗 

Hope that makes sense. Let's discuss on issue if something is not clear 🤗

PR #3759, opened 2021-02-01
Fixed panic on concurrent index-header lazy load / unload

pracucci:

## Changes

We've found a panic that may occur when the index-header lazy loading is enabled. The panic happens because of an edge case happening between when the `load()` releases the write lock and re-acquires the read lock. We've been able to sistematically reproduce it with a unit test and this PR should fix it.

## Verification

Unit test.

### DIFF
--- a/pkg/block/indexheader/lazy_binary_reader.go
+++ b/pkg/block/indexheader/lazy_binary_reader.go
@@ -24,7 +24,8 @@ import (
 )
 
 var (
-	errNotIdle = errors.New("the reader is not idle")
+	errNotIdle              = errors.New("the reader is not idle")
+	errUnloadedWhileLoading = errors.New("the index-header has been concurrently unloaded")
 )
 
 // LazyBinaryReaderMetrics holds metrics tracked by LazyBinaryReader.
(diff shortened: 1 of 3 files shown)


pstibrany (approved):
LGTM, good check. Better to fail query with explicit error message than a panic.

pkg/block/indexheader/lazy_binary_reader.go
```diff
@@ -23,6 +23,10 @@ import (
 	"github.com/thanos-io/thanos/pkg/objstore"
 )
 
+var (
+	errUnloadedWhileLoading = errors.New("the index-header has been concurrently unloaded")
```
MalloZup, line 27: `nit` just my pov, do we need here a variable?

Can we just inline the error where it is used? I just checked the codebase it seems we use more the error directly.

But I might be wrong here. :grin:

bwplotka:
Hmmm do we need this PR with your proper fix in separate PR @pracucci ? 🤔 

Or do we still want this to make sure we don't panic on races/bugs we did not yet discover? If that's so let's comment that this is unlikely, but just in case to not create Segfaults, WDYT?

pracucci:
> Hmmm do we need this PR with your proper fix in separate PR @pracucci ? 🤔

I would say yes. This fix a race condition which still exists even if the other PR gets merged.

pracucci:
@bwplotka I've rebased master and CI now passes. Do you still have concerns about this fix?

bwplotka (approved):
I don't have time to review it properly TBH, but don't want to block you. 

Not sure if we really have to  fail whole query (it's enough to just retry kind of 🤔 )

bwplotka:
It's LGTM for a start (:

bwplotka:
Or rather: I reviewed it properly and it makes sense for now, but just don't have time to suggest something better 🙃

PR #3767, opened 2021-02-03
Fixed Planner.Plan() description

pracucci:

## Changes

The planner doesn't work anymore on directories, but just on metas. Because of this, I've updated the comment.

## Verification

N/A


bwplotka (approved):
LGTM! Thanks a lot

PR #3769, opened 2021-02-03
pkg/server/http: Allow passing http.ServeMux with a server option

metalmatze:

## Changes

This allows to pass a specific mux from the outside that the internal server is added on top of.
Nothing really changes.

### DIFF
--- a/pkg/server/http/http.go
+++ b/pkg/server/http/http.go
@@ -39,6 +39,10 @@ func New(logger log.Logger, reg *prometheus.Registry, comp component.Component, 
 	}
 
 	mux := http.NewServeMux()
+	if options.mux != nil {
+		mux = options.mux
+	}
+
 	registerMetrics(mux, reg)
 	registerProbes(mux, prober, logger)
 	registerProfiler(mux)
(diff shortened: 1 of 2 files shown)


pkg/server/http/http.go
```diff
@@ -39,6 +39,10 @@ func New(logger log.Logger, reg *prometheus.Registry, comp component.Component,
 	}
 
 	mux := http.NewServeMux()
```
yeya24, line 41: One small nit. We still allocate memory for a new mux even if we pass one. Is it better to use 
``` go
	var mux *http.ServeMux
	if options.mux != nil {
		mux = options.mux
	} else {
		mux = http.NewServeMux()
	}
```

yeya24 (approved):
LGTM

pkg/server/http/http.go
metalmatze: Right. As this is only called once in the lifecycle ever I'm personally fine with the current way this is implemented. Happy to update nonetheless if you think it's worth it.
The idea being, that the default Server works out of the box and then options can modify the defaults.

bwplotka: Yea, that's not a big deal 👍🏽 Happy to merge.

bwplotka (approved):

metalmatze:
Thanks! :)

Issue #3634, opened 2020-12-17
Upgrade y18n package to v4.0.1


kyoto:
Upgrade y18n package (used by the React UI) from v4.0.0 to v4.0.1, which fixes [CVE-2020-7774](https://nvd.nist.gov/vuln/detail/CVE-2020-7774).

See https://github.com/yargs/y18n/blob/v4/CHANGELOG.md#401-2020-11-30

(Issue is also fixed in y18n v5.0.5)

This is the same issue as https://github.com/prometheus/prometheus/issues/8282

kakkoyun:
@kyoto Thanks for opening this. We should take care of it. Help wanted.
Also, it's curious that we haven't alerted by the dependabot?

kyoto:
This was fixed by https://github.com/thanos-io/thanos/pull/3813

PR #3770, opened 2021-02-04
upgrade  package y18n to version: 4.0.1

spaparaju:
Signed-off-by: spaparaju <paparaju@gmail.com>

upgrade  package y18n to version: 4.0.1. This is a fix for this [bug](https://github.com/thanos-io/thanos/issues/3634).

## Changes

upgrade  package y18n to version to 4.0.1
## Verification

Screenshot 2021-02-04 at 2 55 56 PM

### DIFF
--- a/pkg/ui/react-app/package.json
+++ b/pkg/ui/react-app/package.json
@@ -45,7 +45,8 @@
     "tempusdominus-bootstrap-4": "^5.1.2",
     "tempusdominus-core": "^5.0.3",
     "typescript": "^3.3.3",
-    "use-query-params": "^1.1.6"
+    "use-query-params": "^1.1.6",
+    "y18n": "4.0.1"
   },
   "scripts": {
     "start": "react-scripts start",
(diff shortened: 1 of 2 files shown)


spaparaju:
@kakkoyun appreciate your review

pkg/ui/react-app/package.json
```diff
@@ -45,7 +45,8 @@
     "tempusdominus-bootstrap-4": "^5.1.2",
     "tempusdominus-core": "^5.0.3",
     "typescript": "^3.3.3",
-    "use-query-params": "^1.1.6"
+    "use-query-params": "^1.1.6",
+    "y18n": "4.0.1"
```
prmsrswt, line 49: `y18n` is not a direct dependency so adding it explicitly will create another copy of it with different version. You can see that now we have two entries for `y18n` in `yarn.lock`.

You can see which exact module to update if you run `yarn why y18n`.

```
yarn why y18n
yarn why v1.22.10
[1/4] Why do we have the module "y18n"...?
[2/4] Initialising dependency graph...
[3/4] Finding dependency...
[4/4] Calculating file sizes...
=> Found "y18n@4.0.0"
info Reasons this module exists
   - "yargs" depends on it
   - Hoisted from "yargs#y18n"
   - Hoisted from "react-scripts#webpack-dev-server#yargs#y18n"
   - Hoisted from "react-scripts#webpack#terser-webpack-plugin#cacache#y18n"
info Disk size without dependencies: "24KB"
info Disk size with unique dependencies: "24KB"
info Disk size with transitive dependencies: "24KB"
info Number of shared dependencies: 0
Done in 0.56s.
```
I ran it locally and it looks like we need to upgrade `react-scripts` to the latest minor version.

Also, I think it'd be better to upgrade all dependencies to the latest patch version as well by running `yarn upgrade` or `yarn upgrade-interactive`.

kyoto:
Looks like this was taken care of by https://github.com/thanos-io/thanos/pull/3813

bwplotka:
Nice! I assume we can close this PR then? Thanks for helping anyway!

PR #3771, opened 2021-02-04
Refactoring: merge 2 functions in one compact one.

MalloZup:
# Rationale/Description:

This is part of #2248 .

(/Reworked)

Using structs as refactor

### DIFF
--- a/cmd/thanos/query.go
+++ b/cmd/thanos/query.go
@@ -23,7 +23,6 @@ import (
 	"github.com/prometheus/common/route"
 	"github.com/prometheus/prometheus/discovery/file"
 	"github.com/prometheus/prometheus/discovery/targetgroup"
-	"github.com/prometheus/prometheus/pkg/labels"
 	"github.com/prometheus/prometheus/promql"
 
 	v1 "github.com/thanos-io/thanos/pkg/api/query"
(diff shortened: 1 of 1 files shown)


bwplotka (commented):
Thanks for this, and I overall see benefits, but I am not fan doing this in such order.

Essentially I don't see how this helps with the pattern we establish here: https://github.com/thanos-io/thanos/issues/2248 (introducing config structs), With this there is also downside for one more indent, so it's hard to see if/for etc The main problem though might be inconsistency: We have some function in `run(*)` pattern, some in `config Structs` and now we have 3rd pattern. To me it's might a bit more harm then help.... 🤔 

Any blocker to introduce config structs in this PR? 🤗

MalloZup:
@bwplotka  ok I agree that having structs is useful , I was thinking to add them later on. 

How do you suggest to have such structs? Should I just bruteforce them as the actual big structure args we passed? tia :man_astronaut:

MalloZup:
@bwplotka looking at the previous pr, one thing I don't understand:

a struct was put to 
https://github.com/thanos-io/thanos/pull/2604/files#diff-97b997bb71cef88ab4f8ce34d5bfd5705222b47c7f3c4f91d462fd49a4c4503bR129

the other one https://github.com/thanos-io/thanos/pull/2604/files#diff-6c2257b871fd1196514f664bc7e44cb21681215e0929710d0ad5ceea90b8e122R532

before I will do the refactor,

can we agree that I create a struct with the whole vars here

```
	httpBindAddr, httpGracePeriod := extkingpin.RegisterHTTPFlags(cmd)
	grpcBindAddr, grpcGracePeriod, grpcCert, grpcKey, grpcClientCA := extkingpin.RegisterGRPCFlags(cmd)

	secure := cmd.Flag("grpc-client-tls-secure", "Use TLS when talking to the gRPC server").Default("false").Bool()
	cert := cmd.Flag("grpc-client-tls-cert", "TLS Certificates to use to identify this client to the server").Default("").String()
	key := cmd.Flag("grpc-client-tls-key", "TLS Key for the client's certificate").Default("").String()
	caCert := cmd.Flag("grpc-client-tls-ca", "TLS CA Certificates to use to verify gRPC servers").Default("").String()
	serverName := cmd.Flag("grpc-client-server-name", "Server name to verify the hostname on the returned gRPC certificates. See https://tools.ietf.org/html/rfc4366#section-3.1").Default("").String()

	webRoutePrefix := cmd.Flag("web.route-prefix", "Prefix for API and UI endpoints. This allows thanos UI to be served on a sub-path. Defaults to the value of --web.external-prefix. This option is analogous to --web.route-prefix of Prometheus.").Default("").String()
	webExternalPrefix := cmd.Flag("web.external-prefix", "Static prefix for all HTML links and redirect URLs in the UI query web interface. Actual endpoints are still served on / or the web.route-prefix. This allows thanos UI to be served behind a reverse proxy that strips a URL sub-path.").Default("").String()
	webPrefixHeaderName := cmd.Flag("web.prefix-header", "Name of HTTP request header used for dynamic prefixing of UI links and redirects. This option is ignored if web.external-prefix argument is set. Security risk: enable this option only if a reverse proxy in front of thanos is resetting the header. The --web.prefix-header=X-Forwarded-Prefix option can be useful, for example, if Thanos UI is served via Traefik reverse proxy with PathPrefixStrip option enabled, which sends the stripped prefix value in X-Forwarded-Prefix header. This allows thanos UI to be served on a sub-path.").Default("").String()

	requestLoggingDecision := cmd.Flag("log.request.decision", "Request Logging for logging the start and end of requests. LogFinishCall is enabled by default. LogFinishCall : Logs the finish call of the requests. LogStartAndFinishCall : Logs the start and finish call of the requests. NoLogCall : Disable request logging.").Default("LogFinishCall").Enum("NoLogCall", "LogFinishCall", "LogStartAndFinishCall")

	queryTimeout := extkingpin.ModelDuration(cmd.Flag("query.timeout", "Maximum time to process query by query node.").
		Default("2m"))

		Default("20").Int()

	lookbackDelta := cmd.Flag("query.lookback-delta", "The maximum lookback duration for retrieving metrics during expression evaluations. PromQL always evaluates the query for the certain timestamp (query range timestamps are deduced by step). Since scrape intervals might be different, PromQL looks back for given amount of time to get latest sample. If it exceeds the maximum lookback delta it assumes series is stale and returns none (a gap). This is why lookback delta should be set to at least 2 times of the slowest scrape interval. If unset it will use the promql default of 5m.").Duration()
	dynamicLookbackDelta := cmd.Flag("query.dynamic-lookback-delta", "Allow for larger lookback duration for queries based on resolution.").Hidden().Default("true").Bool()

	maxConcurrentSelects := cmd.Flag("query.max-concurrent-select", "Maximum number of select requests made concurrently per a query.").
		Default("4").Int()

	queryReplicaLabels := cmd.Flag("query.replica-label", "Labels to treat as a replica indicator along which data is deduplicated. Still you will be able to query without deduplication using 'dedup=false' parameter. Data includes time series, recording rules, and alerting rules.").
		Strings()
```
etc

into a struct called `queryConfig` which is in same go file as the others? Let me know and thx for review

MalloZup:
@bwplotka  on my last commit I made some progress.

Just to clarify one point:

I think that having a method with struct, **without** state is an anti-pattern. https://github.com/thanos-io/thanos/pull/2267/files#diff-97b997bb71cef88ab4f8ce34d5bfd5705222b47c7f3c4f91d462fd49a4c4503bR22

Instead if we care about having a struct, we should pass it to the function as implemented it here.

A method in golang with struct should be used when it deals with modifying some struct, but in our case , the `method` is just a place older which is imho confusion.

So I think would be more coherent for us to use just the struct as per commit and then pass it to the functions. Of course we might need to adapt the other parts.

MalloZup:
finally I think is ok to have structs but, I think the `runQuery` function could be also removed.

I could , take care of the other part of the code who introduce the "less" clear pattern so we uniform the code, but I believe that :

- having 1 function
- having struct 
-  removing the run function
- 
would be the way to proceed. I think since this is a technical debt, I think we might need to consider to fix it correctly everywhere in sake of clarity, even if we might take more time on re-changing the pattern. AFAIK we had only 1 PR which introduced that `new` pattern with the method

bwplotka (approved):
I think this looks ready, why it was marked as draft? 🤗 

@MalloZup do you have time to finish it/rebase? LGTM really!

MalloZup:
@bwplotka  yep I forgot to remove I think the Draft.. :grin:  I'm short on time this period, would be great if you could resolve the conflicts if possible otherwise I fear I could remove something wrong since I missed a bit the late development on the main branch.. tia

kakkoyun:
Hey @MalloZup 👋  We're going to cut a release candidate soon, is there any chance for you to push this to the finish line?

kakkoyun:
I think it just needs a rebase?

cmd/thanos/query.go
```diff
@@ -47,114 +47,115 @@ import (
 	"github.com/thanos-io/thanos/pkg/ui"
 )
 
+type queryConfig struct {
```
kakkoyun, line 50: Better to move to the bottom to follow the convention with the other commands.

MalloZup:
@kakkoyun right now unfortunately I don't have time to do this... :( could you as maintainer fix the small things remaining? that would help a lot. I'm sorry for the inconvenient but I am quite busy at moment. Greeting to you and your dog :)

kakkoyun:
@MalloZup 👍 Of course, thanks for letting us know.

Issue #3412, opened 2020-11-05
tests: Flaky e2e: TestQueryExternalPrefix


bwplotka:
E.g: https://github.com/thanos-io/thanos/pull/3409/checks?check_run_id=1354006689#step:5:1480

Also, it would be nice to show logs only for failed tests.

Abhishek357:
I am working on this and will open a PR soon.

PR #3777, opened 2021-02-09
e2e: Resolved Flaky tests

Abhishek357:
Fixes: #3412

## Changes

Tests are retried using the Retry function in runutil which executes the function(which makes it flaky) every interval until timeout or no error is returned.
## Verification

Tested Locally

### DIFF
--- a/test/e2e/query_test.go
+++ b/test/e2e/query_test.go
@@ -371,6 +371,14 @@ func checkNetworkRequests(t *testing.T, addr string) {
 	testutil.Ok(t, runutil.Retry(1*time.Minute, ctx.Done(), func() error {
 		var networkErrors []string
 
+		// Listen for failed network requests and push them to an array.
+		chromedp.ListenTarget(ctx, func(ev interface{}) {
+			switch ev := ev.(type) {
+			case *network.EventLoadingFailed:
+				networkErrors = append(networkErrors, ev.ErrorText)
+			}
+		})
+
 		err := chromedp.Run(ctx,
 			network.Enable(),
 			chromedp.Navigate(addr),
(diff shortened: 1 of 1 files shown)


test/e2e/query_test.go
```diff
@@ -368,31 +368,35 @@ func checkNetworkRequests(t *testing.T, addr string) {
...
-			return fmt.Errorf("some network requests failed: %s", strings.Join(networkErrors, "; "))
+		if err != nil {
+			return err
 		}
-		return nil
-	}()
 
-	testutil.Ok(t, err)
+		// Listen for failed network requests and push them to an array.
+		chromedp.ListenTarget(ctx, func(ev interface{}) {
```
GiedriusS, line 385: Shouldn't this happen before line 371? :thinking: Now we are doing this *after* the requests at the first iteration.

GiedriusS (commented):
Looks good, just one suggestion. Also, could you please share the logs of the error we are fixing? I cannot access the old logs anymore as they have "expired".

Abhishek357:
Error Logs - 
ErrorLog

test/e2e/query_test.go
Abhishek357: Thanks for the review!
Do you mean outside the retry function? 
Well as we can see in the error logs("some network request failed") it fails because the size of networkErrors array is >0 and if we call the ListenTarget outside, will it append errors to this array in all iterations.

GiedriusS: My suggestion was to clean the slice on each iteration. I suppose the current design is OK as you say but don't we need to register this callback (https://pkg.go.dev/github.com/chromedp/chromedp#ListenTarget) before calling `chromedp.Run()`?

test/e2e/query_test.go
```diff
@@ -368,31 +368,35 @@ func checkNetworkRequests(t *testing.T, addr string) {
...
+			switch ev := ev.(type) {
+			case *network.EventLoadingFailed:
+				networkErrors = append(networkErrors, ev.ErrorText)
+			}
+		})
+
+		err = func() error {
```
GiedriusS, line 392: Is there a reason why this is a function? :s Does it give any benefits? I think we can move all of the code inside of the function to the outside.

test/e2e/query_test.go
Abhishek357: You are right we should. Done :+1:

test/e2e/query_test.go
Abhishek357: No, there are no benefits.

prmsrswt (approved):
Looks good to me now. Thanks!

Abhishek357:
Circle CI should pass can we re-run test.

GiedriusS (approved):
LGTM, thank you!

PR #3780, opened 2021-02-10
Added a place holder for development documentation

yashrsharma44:
Let's add a placeholder for adding resources for new contributors and instructions for starting using Thanos.


kakkoyun:
@yashrsharma44 How about merging the document when we actually have content? Why do we put a placeholder?

yashrsharma44:
Okay, that sounds nice, would be fine to add the format in the another PR.

PR #3783, opened 2021-02-11
Allow to customise S3 SSE on a per-request basis

### LINKED ISSUE #3781: Per call objstore Upload/Get options

Cortex uses on Thanos objstore clients. Cortex needs to build per-tenant S3 encryption key support and would like to avoid to create a new S3 client for each tenant: having a single S3 client allows to reuse S3 connections.

The Thanos S3 client allows to config the `SSEConfig` at client creation time. In order to reuse the same client for different tenants, we would need to be able to pass a custom `SSEConfig` to each single `Get()` and `Upload()` call.

A solution would be introducing the ability to pass options to `Get()` and `Upload()`. To not introduce breaking changes in the `objstore.Bucket` signature we could leverage on the `context.Context`.

An idea could be having a `WithOptions(objstore.Bucket) objstore.Bucket` generic function to wrap a `objstore.Bucket` with options which are then injected in the context by the options wrapper. The underlying client implementation (eg. `s3.go`) would be changed to be able to read such options and configure the Get/Upload operation accordingly.

If a bucket doesn't support options (eg. SSE encryption is only supported by S3) such options would be ignored.

- Create an S3 client for each tenant (not an option, we have a large number of tenants managed for each process, we can't afford to re-create S3 connections for each of them)
- Fork S3 client in Cortex (still an option, but not desired)

### Additional context

- [Slack thread](https://cloud-native.slack.com/archives/CL25937SP/p1612947447142700)
- Encryption in Thanos objstore clients is currently supported only by S3

pracucci:

## Changes

In this PR I'm proposing to introduce a context key to allow downstream projects (Cortex in this case) to inject custom SSE config on a per-request basis.

Fixes #3781.

## Verification

Unit test.

### DIFF
--- a/pkg/objstore/s3/s3.go
+++ b/pkg/objstore/s3/s3.go
@@ -47,10 +47,12 @@ const (
 	// SSES3 is the name of the SSE-S3 method for objstore encryption.
 	SSES3 = "SSE-S3"
 
-	// SSEConfigKey is the context key to override SSE config. This feature is used by downstream
+	// sseConfigKey is the context key to override SSE config. This feature is used by downstream
 	// projects (eg. Cortex) to inject custom SSE config on a per-request basis. Future work or
 	// refactoring can introduce breaking changes as far as the functionality is preserved.
-	SSEConfigKey = ctxKey(0)
+	// NOTE: we're using a context value only because it's a very specific S3 option. If SSE will
+	// be available to wider set of backends we should probably add a variadic option to Get() and Upload().
+	sseConfigKey = ctxKey(0)
 )
 
 var DefaultConfig = Config{
@@ -497,7 +499,7 @@ func (b *Bucket) Close() error { return nil }
 
 // getServerSideEncryption returns the SSE to use.
 func (b *Bucket) getServerSideEncryption(ctx context.Context) (encrypt.ServerSide, error) {
-	if value := ctx.Value(SSEConfigKey); value != nil {
+	if value := ctx.Value(sseConfigKey); value != nil {
 		if sse, ok := value.(encrypt.ServerSide); ok {
 			return sse, nil
 		}
@@ -580,4 +582,10 @@ func NewTestBucketFromConfig(t testing.TB, location string, c Config, reuseBucke
 			t.Logf("deleting bucket %s failed: %s", bktToCreate, err)
 		}
 	}, nil
+}
+
+// ContextWithSSEConfig returns a context with a  custom SSE config set. The returned context should be
+// provided to S3 objstore client functions to override the default SSE config.
+func ContextWithSSEConfig(ctx context.Context, value encrypt.ServerSide) context.Context {
+	return context.WithValue(ctx, sseConfigKey, value)
 }
--- a/pkg/objstore/s3/s3.go
+++ b/pkg/objstore/s3/s3.go
@@ -32,6 +32,8 @@ import (
 	"gopkg.in/yaml.v2"
 )
 
+type ctxKey int
+
 const (
 	// DirDelim is the delimiter used to model a directory structure in an object store bucket.
 	DirDelim = "/"
@@ -44,6 +46,11 @@ const (
 
 	// SSES3 is the name of the SSE-S3 method for objstore encryption.
 	SSES3 = "SSE-S3"
+
+	// SSEConfigKey is the context key to override SSE config. This feature is used by downstream
+	// projects (eg. Cortex) to inject custom SSE config on a per-request basis. Future work or
+	// refactoring can introduce breaking changes as far as the functionality is preserved.
+	SSEConfigKey = ctxKey(0)
 )
 
 var DefaultConfig = Config{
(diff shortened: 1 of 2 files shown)


pkg/objstore/s3/s3.go
```diff
@@ -44,6 +46,11 @@ const (
 
 	// SSES3 is the name of the SSE-S3 method for objstore encryption.
 	SSES3 = "SSE-S3"
+
+	// SSEConfigKey is the context key to override SSE config. This feature is used by downstream
+	// projects (eg. Cortex) to inject custom SSE config on a per-request basis. Future work or
+	// refactoring can introduce breaking changes as far as the functionality is preserved.
+	SSEConfigKey = ctxKey(0)
```
pstibrany, line 53: Instead of making this public, I would keep `SSEConfigKey` private, and provide `SetSSEConfigKeyInContext` function, which can then also use a correct `encrypt.ServerSide` type as parameter.

pstibrany (approved):
LGTM, except we can improve type safety by providing setter function.

pkg/objstore/s3/s3.go
pracucci: Yeah. I would like to hear @bwplotka feedback as well, cause this is like the `WithOptions()` idea which was discarded during the call.

pracucci: BTW, I think your suggestion is better. Just wanna check with Bartek to avoid going back and forth with changes.

kakkoyun (commented):
LGTM. 
Let's wait for @bwplotka's blessing and the comment to be addressed https://github.com/thanos-io/thanos/pull/3783/files#r574467906

pkg/objstore/s3/s3.go
```diff
@@ -44,6 +46,11 @@ const (
 
 	// SSES3 is the name of the SSE-S3 method for objstore encryption.
 	SSES3 = "SSE-S3"
+
+	// SSEConfigKey is the context key to override SSE config. This feature is used by downstream
+	// projects (eg. Cortex) to inject custom SSE config on a per-request basis. Future work or
+	// refactoring can introduce breaking changes as far as the functionality is preserved.
```
bwplotka, line 52: Let's add a comment maybe, a NOTE that this is context value only because it's a very specific S3 option. If SSE is available wider we should probably add a variadic option as (https://github.com/thanos-io/thanos/issues/3784#issuecomment-778330460)  WDYT?

pkg/objstore/s3/s3.go
bwplotka: I think `SetSSEConfigKeyInContext` function and `SSEConfigKey = ctxKey(0)` is the same really, no? I am fine with any as this is matching what we agreed.

Unless we want to spend that time on abstracting and putting in some variadic `opts ...UploadOptions` 🤔 on Upload method? (as https://github.com/thanos-io/thanos/issues/3784#issuecomment-778330460 suggsestion)

Something like `Upload(ctx context.Context, name string, r io.Reader, opts ...UploadOptions)` with  and `WithSSEConfig(...)` so someone can call `Upload(ctx, "yolo", b, objstore.WithSSEConfig)`. As long it's not S3 specific only.

bwplotka (approved):
Thank you @pracucci @pstibrany 

LGTM from my side. I think this is a good start 🤗

pkg/objstore/s3/s3.go
pstibrany: > I think `SetSSEConfigKeyInContext` function and `SSEConfigKey = ctxKey(0)` is the same really, no? 

Exposing key relies on client putting the correctly-typed value to the context, and failing in runtime if that’s not the case. Function can have argument of the correct type, and avoid the problem during compilation. Other than that, yes.

pracucci:
Thanks @pstibrany and @bwplotka for your review. I should have addressed your comments. Can you take another look, please?

pstibrany (approved):
LGTM, thanks!

pkg/objstore/s3/s3.go
```diff
@@ -44,6 +46,13 @@ const (
 
 	// SSES3 is the name of the SSE-S3 method for objstore encryption.
 	SSES3 = "SSE-S3"
+
+	// sseConfigKey is the context key to override SSE config. This feature is used by downstream
+	// projects (eg. Cortex) to inject custom SSE config on a per-request basis. Future work or
+	// refactoring can introduce breaking changes as far as the functionality is preserved.
+	// NOTE: we're using a context value only because it's a very specific S3 option. If SSE will
```
bwplotka, line 53:
```suggestion
	// NOTE: We're using a context value only because it's a very specific S3 option. If SSE will
```

bwplotka (approved):
Thanks, LGTM 👍🏽

pkg/objstore/s3/s3.go
bwplotka: not  a biggy, let's merge.

PR #3788, opened 2021-02-12
WIP: Introduce a simple wrapper that wraps a regular bucket with configurable retry backoffs.

johannesfrey:

## Changes

The main motivation for this PR is to better cope with potential rate limits that might be enforced by specific S3 providers. Without the possibility of backoff retries at the level of bucket operations, errors caused by those rate limits (e.g. 503, SlowDown/Service Unavailable) are handled at a "too high" level. More precisely, e.g. the main compaction loop of compact catches this as a retriable error and postpones the whole compaction until the next run interval.
One could also take this one step further and delegate the retries one level lower to the actual S3 client (minio in this case). As a matter of fact minio already performs retries under the hood, but those default values might not suffice and are not really configurable from other packages. There is a [global variable](https://github.com/minio/minio-go/blob/v7.0.2/retry.go#L27) that could be increased, but this might introduce a too tight coupling to this part of the minio package, and this would possibly affect all thanos components that use the client.

This PR is still WIP (tests, comments etc. missing :-) and should rather be considered a design doc with a POC implementation ready for feedback, if this a feasible way to do this?! If not or if the topic is considered "big enough" I could alternatively start with a "proper" design doc?!

Also, credits where credit's due: The actual retry implementations are inspired by this [post](https://blog.gopheracademy.com/advent-2014/backoff/) as well as minio's [retry function](https://github.com/minio/minio-go/blob/master/retry.go#L45).

Unfortunately I didn't completely check other current states of implementations. It looks like https://github.com/thanos-io/thanos/pull/3756 also tries to tackle this in the upload case.

## Verification

Deployed and run in Kubernetes cluster

<sub>Johannes Frey [johannes.frey@daimler.com](mailto:johannes.frey@daimler.com) Daimler TSS GmbH ([Imprint](https://github.com/Daimler/daimler-foss/blob/master/LEGAL_IMPRINT.md))</sub>

### DIFF
--- a/pkg/objstore/backoff.go
+++ b/pkg/objstore/backoff.go
@@ -1,0 +1,248 @@
+package objstore
+
+import (
+	"context"
+	"io"
+	"strings"
+	"time"
+
+	"github.com/go-kit/kit/log"
+	"github.com/go-kit/kit/log/level"
+	"github.com/prometheus/client_golang/prometheus"
+	"github.com/prometheus/client_golang/prometheus/promauto"
+	"github.com/thanos-io/thanos/pkg/runutil"
+)
+
+type backoffBucket struct {
+	Bucket
+	runutil.Backoff
+	logger        log.Logger
+	ignoreObjects []string
+
+	retries *prometheus.CounterVec
+}
+
+func NewBucketWithBackoff(bkt Bucket, backoff runutil.Backoff, logger log.Logger, reg prometheus.Registerer, ignoreObjects ...string) InstrumentedBucket {
+	b := &backoffBucket{
+		Bucket:        bkt,
+		Backoff:       backoff,
+		logger:        logger,
+		ignoreObjects: ignoreObjects,
+		retries: promauto.With(reg).NewCounterVec(prometheus.CounterOpts{
+			Name:        "thanos_objstore_bucket_backoff_retries_total",
+			Help:        "Total number of retried operations against a bucket.",
+			ConstLabels: prometheus.Labels{"bucket": bkt.Name()},
+		}, []string{"operation"}),
+	}
+
+	for _, op := range []string{
+		OpGet,
+		OpGetRange,
+		OpExists,
+		OpUpload,
+		OpDelete,
+	} {
+		b.retries.WithLabelValues(op)
+	}
+	return b
+}
+
+func (b *backoffBucket) Get(ctx context.Context, name string) (io.ReadCloser, error) {
+	const op = OpGet
+
+	r, err := b.Bucket.Get(ctx, name)
+	// Don't retry in case of 404 or if it's an object worth ignoring.
+	if err == nil || b.Bucket.IsObjNotFoundErr(err) || b.isIgnoreObject(name) {
+		return r, err
+	}
+
+	for retry := 1; retry <= b.Backoff.Max(); retry++ {
+		select {
+		case <-ctx.Done():
+			return r, err
+		case <-b.Backoff.Done():
+			return r, err
+		case <-time.After(b.Backoff.Duration(retry)):
+			// Retry again after backoff
+		}
+
+		r, err = b.Bucket.Get(ctx, name)
+		// Don't retry in case of 404 or if it's an object worth ignoring.
+		if err == nil || b.Bucket.IsObjNotFoundErr(err) || b.isIgnoreObject(name) {
+			return r, err
+		}
+
+		level.Warn(b.logger).Log("msg", "get failed", "name", name, "retryAttempt", retry, "error", err)
+		b.retries.WithLabelValues(op).Inc()
+	}
+
+	return r, err
+}
+
+func (b *backoffBucket) GetRange(ctx context.Context, name string, off, length int64) (io.ReadCloser, error) {
+	const op = OpGetRange
+
+	r, err := b.Bucket.GetRange(ctx, name, off, length)
+	// Don't retry in case of 404 or if it's an object worth ignoring.
+	if err == nil || b.Bucket.IsObjNotFoundErr(err) || b.isIgnoreObject(name) {
+		return r, err
+	}
+
+	for retry := 1; retry <= b.Backoff.Max(); retry++ {
+		select {
+		case <-ctx.Done():
+			return r, err
+		case <-b.Backoff.Done():
+			return r, err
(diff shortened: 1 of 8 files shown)


GiedriusS:
Thank you for this! I have attempted to implement something similar back in the day when it was relevant to me: https://github.com/thanos-io/thanos/pull/2785. Is there a reason you haven't added this config under another key in the regular objstore's config: https://thanos.io/tip/thanos/storage.md/#configuration? Perhaps we could have:
```
type: S3
config:
  ...
retry:
  ...
```

pkg/runutil/backoff.go
```diff
@@ -0,0 +1,276 @@
...
+	SEQUENCE    BackoffType = "SEQUENCE"
+)
+
+type Config struct {
+	Type   BackoffType `yaml:"type"`
+	Config interface{} `yaml:"config"`
+}
+
+func NewBackoff(ctx context.Context, logger log.Logger, confContentYaml []byte) (Backoff, error) {
+	level.Info(logger).Log("msg", "loading backoff configuration")
```
GiedriusS, line 29: Any reason we are reimplementing this instead of using https://github.com/cenkalti/backoff or others? Have you investigated them and they do not work for you for some reason?

pkg/runutil/backoff.go
```diff
@@ -0,0 +1,276 @@
...
+
+func (b *exponentialBackoff) Max() int {
+	return b.maxAttempts
+}
+
+func (b *exponentialBackoff) Done() <-chan struct{} {
+	return b.done
+}
+
+func unitToDuration(unit string) (time.Duration, error) {
```
GiedriusS, line 249: I think we can simply use `model.Duration` & remove this function.

GiedriusS (commented):
Another thing that came to my mind: this presumes that all of the remote object storage APIs are idempotent. That might not be true in the case of Delete(), for example. Maybe it's not a problem. Have you investigated this?

pkg/objstore/backoff.go
```diff
@@ -0,0 +1,248 @@
...
+			return r, err
+		case <-b.Backoff.Done():
+			return r, err
+		case <-time.After(b.Backoff.Duration(retry)):
+			// Retry again after backoff
+		}
+
+		r, err = b.Bucket.Get(ctx, name)
+		// Don't retry in case of 404 or if it's an object worth ignoring.
+		if err == nil || b.Bucket.IsObjNotFoundErr(err) || b.isIgnoreObject(name) {
```
pracucci, line 71: I think it's a bit tricky to determine when we should retry. For example, all 4xx errors shouldn't be retried.

pkg/objstore/backoff.go
```diff
@@ -0,0 +1,248 @@
...
+			return r, err
+		}
+
+		level.Warn(b.logger).Log("msg", "get failed", "name", name, "retryAttempt", retry, "error", err)
```
pracucci, line 75: I don't think it should be this wrapper responsability to log. What this wrapper could do is wrapping the returned error with the number of retries attempted, but in my opinion it should be the caller responsability to log it (if required).

pkg/objstore/backoff.go
```diff
@@ -0,0 +1,248 @@
...
+	retries *prometheus.CounterVec
+}
+
+func NewBucketWithBackoff(bkt Bucket, backoff runutil.Backoff, logger log.Logger, reg prometheus.Registerer, ignoreObjects ...string) InstrumentedBucket {
+	b := &backoffBucket{
+		Bucket:        bkt,
+		Backoff:       backoff,
+		logger:        logger,
+		ignoreObjects: ignoreObjects,
+		retries: promauto.With(reg).NewCounterVec(prometheus.CounterOpts{
```
pracucci, line 31: An histogram may be more useful here. If you have a buckets from 0 util the max number of retries, then you could track the number of retries for each API call in order to have a better split.

pkg/objstore/backoff.go
```diff
@@ -0,0 +1,248 @@
...
+
+	return err
+}
+
+func (b *backoffBucket) Iter(ctx context.Context, dir string, f func(string) error) error {
+	return b.Bucket.Iter(ctx, dir, f)
+}
+
+func (b *backoffBucket) Attributes(ctx context.Context, name string) (ObjectAttributes, error) {
+	return b.Bucket.Attributes(ctx, name)
```
pracucci, line 210: Why not retrying this as well?

pkg/objstore/backoff.go
```diff
@@ -0,0 +1,248 @@
...
+
+		level.Warn(b.logger).Log("msg", "delete failed", "name", name, "retryAttempt", retry, "error", err)
+		b.retries.WithLabelValues(op).Inc()
+	}
+
+	return err
+}
+
+func (b *backoffBucket) Iter(ctx context.Context, dir string, f func(string) error) error {
+	return b.Bucket.Iter(ctx, dir, f)
```
pracucci, line 206: A problem of handling retries with a wrapper is that `Iter()` can't be retried. Despite harder to build, I personally believe retries should be handled at a lower level.

pkg/objstore/backoff.go
```diff
@@ -0,0 +1,248 @@
...
+
+	"github.com/go-kit/kit/log"
+	"github.com/go-kit/kit/log/level"
+	"github.com/prometheus/client_golang/prometheus"
+	"github.com/prometheus/client_golang/prometheus/promauto"
+	"github.com/thanos-io/thanos/pkg/runutil"
+)
+
+type backoffBucket struct {
+	Bucket
```
pracucci, line 17: I would make it a named field. The problem is that whenever we'll add more functions to `Bucket` we may want this implementation to break so we can take an informed decision whether to add retries or not to the new function.

pracucci (commented):
Thanks for working on it! I personally have mixed feelings handling retries such an high level because of few reasons:
1. It's hard to do error handling (eg. all 4xx shouldn't be retried)
2. What if a specific bucket client already handles retries? Are we going to retry the retries?

johannesfrey:
Thx for all the feedback and sorry for my late reply. I'm trying to address the mentioned issues, unfortunately I'm a bit burried with some other work, so I hope it's okay that the changes will take a tad longer :pensive:.

PR #3789, opened 2021-02-12
stalebot: Add an exempt label to ignore issue that has a PR

kakkoyun:
This PR adds an exempt label to stalebot to ignore the issue with active PRs. 
To mark the issue that has an active PR, maintainers should add a specific label: [`state: someone-working-on-it`](https://github.com/thanos-io/thanos/labels/state%3A%20someone-working-on-it)

The idea triggered in https://github.com/thanos-io/thanos/issues/3456

What do you think?

### DIFF
--- a/.github/stale.yml
+++ b/.github/stale.yml
@@ -10,6 +10,7 @@ exemptLabels:
   # This label is hardcoded on remind bot (https://probot.github.io/apps/reminders/) and is used by remind bot when
   # issue is being reminded.
   - reminder
+  - 'state: someone-working-on-it' 
 
 # Set to true to ignore issues in a project (defaults to false)
 exemptProjects: false


pracucci (approved):

bwplotka (commented):
Hmm, is there an automated way to tell if PR is open for the issue? I feel it's prone to stagnation, where we will mark issue like this, and then PR got stale and closed / just closed leaving issue statle and infinitely open 🤔

kakkoyun:
Those are all valid points. And I don't have the answers now. I need to spare some time on this and check what are the available options. Maybe we have a smarter stalebot? Or we can try to improve it on the upstream.

yeya24 (approved):
LGTM. Would be good if we can add some comments about this label

yeya24:
Now there is no better plan. But I agree it would be really good if we can automate this. For now, I would say let's try it!

PR #3790, opened 2021-02-12
for downloading object, be pessimistic and check how many bytes got c…

su47flying:
i got the same issue with [#2805]. Compact
i find some download object's length was not same with it's real length(on object storage).
this PR will check the copy lenght, if got the length was not equal to the real length, thanos will do retry.

## Verification

in my env, after we apply this code, it works well.
i can get some retry log for this code.

### DIFF
--- a/pkg/objstore/objstore.go
+++ b/pkg/objstore/objstore.go
@@ -194,8 +194,16 @@ func DownloadFile(ctx context.Context, logger log.Logger, bkt BucketReader, src,
 	}()
 	defer runutil.CloseWithLogOnErr(logger, f, "download block's output file")
 
-	if _, err = io.Copy(f, rc); err != nil {
+	atts, err := bkt.Attributes(ctx, src)
+	if err != nil {
+		return errors.Wrapf(err, "attributes file %s", src)
+	}
+	copied, err := io.Copy(f, rc)
+	if err != nil {
 		return errors.Wrap(err, "copy object to file")
+	}
+	if copied != atts.Size {
+		return errors.Errorf("copy object to file %s. object length %v but got %v", src, atts.Size, copied)
 	}
 	return nil
 }


pkg/objstore/objstore.go
```diff
@@ -194,9 +194,17 @@ func DownloadFile(ctx context.Context, logger log.Logger, bkt BucketReader, src,
 	}()
 	defer runutil.CloseWithLogOnErr(logger, f, "download block's output file")
 
-	if _, err = io.Copy(f, rc); err != nil {
+	atts, err := bkt.Attributes(ctx, src)
```
pracucci, line 197: This will double the number of API calls required to download a file. I don't think we should do it.

su47flying: if don't do it. sometimes return value of io.Copy was less than the legnth of on the object storage.
the file was not complete, but  was not err be return.
so ,the thanos compact will change state to [halt]. halt meas ....
if do it, thanos compact will retry.  
i don't know some graceful approach to fix this issue.

pracucci: > if don't do it. sometimes return value of io.Copy was less than the legnth of on the object storage.

Looks an issue with a specific backend storage that shouldn't be fixed here. Which backend storage are you using?

su47flying: S3

pracucci: I confirm I believe this is not the right place to address it, for a couple of reasons:
1. We shouldn't introduce additional API calls
2. If the reader returned by S3 `Get()` can return partial response without erroring out, then the problem is widespread across the whole codebase (any usage of `Get()`) so fixing it here would just cover a small amount of use cases

_Checking the logic to have a better proposal_

pracucci: See: https://github.com/thanos-io/thanos/issues/2805#issuecomment-778147852

PR #3792, opened 2021-02-12
Introduce "allow overlapping blocks" flag for Thanos receiver

Max-Chandler:

## Changes

This change enables the "allow-overlapping-blocks" flag to be passed down to the TSDB on the Thanos Receive component. 

This overlap can occur where there is a backlog of metrics waiting to be pushed to Thanos and a compaction event happens. The Receiver TSDB will write out the WAL to a TSDB block and start a new block. If the first metric has a time that is earlier than the end of the previous block, these blocks will overlap and cause issues when the next compaction event happens. 

By adding and enabling this flag it allows the receive component to vertically compact these blocks and deal with the overlap. 

This is related to: https://github.com/prometheus/prometheus/issues/8055 

## Verification

I built the image and observed vertical compaction happening on the receivers when the overlap issue occurred by monitoring the `prometheus_tsdb_vertical_compactions_total` metric.


bwplotka (approved):
Thank You!

LGTM 🤗

bwplotka:
Love your PR, clean and focused.

PR #3795, opened 2021-02-12
Truncated S3 "get object" response is reported as error

pracucci:

## Changes

From time to time we hear issues which looks be caused by a truncated response got from S3 or S3-compatible backends. An example is https://github.com/thanos-io/thanos/issues/2805.

I've added a unit test to S3 and GCS to check how the client behaves if the HTTP response body contains less data than the size reported in the `Content-Length`. Surprise surprise... GCS client returns `io.ErrUnexpectedEOF` while S3 client (Minio) returns no error.

I know this is not something we should fix in Thanos (I guess should be fixed in Minio) but this test should prove it.

I've tried to upgrade Minio to `7.0.8` (latest minor available) and the behaviour doesn't change. I've opened a [Minio issue](https://github.com/minio/minio-go/issues/1451) to hear some feedback from maintainers.

## Verification

Unit test.

### DIFF
--- a/pkg/objstore/gcs/gcs_test.go
+++ b/pkg/objstore/gcs/gcs_test.go
@@ -1,0 +1,44 @@
+package gcs
+
+import (
+	"context"
+	"io"
+	"io/ioutil"
+	"net/http"
+	"net/http/httptest"
+	"os"
+	"testing"
+
+	"github.com/go-kit/kit/log"
+
+	"github.com/thanos-io/thanos/pkg/testutil"
+)
+
+func TestBucket_Get_ShouldReturnErrorIfServerTruncateResponse(t *testing.T) {
+	srv := httptest.NewServer(http.HandlerFunc(func(w http.ResponseWriter, r *http.Request) {
+		w.Header().Set("Last-Modified", "Wed, 21 Oct 2015 07:28:00 GMT")
+		w.Header().Set("Content-Length", "100")
+
+		// Write less bytes than the content length.
+		_, err := w.Write([]byte("12345"))
+		testutil.Ok(t, err)
+	}))
+	defer srv.Close()
+
+	os.Setenv("STORAGE_EMULATOR_HOST", srv.Listener.Addr().String())
+
+	cfg := Config{
+		Bucket:         "test-bucket",
+		ServiceAccount: "",
+	}
+
+	bkt, err := NewBucketWithConfig(context.Background(), log.NewNopLogger(), cfg, "test")
+	testutil.Ok(t, err)
+
+	reader, err := bkt.Get(context.Background(), "test")
+	testutil.Ok(t, err)
+
+	// We expect an error when reading back.
+	_, err = ioutil.ReadAll(reader)
+	testutil.Equals(t, io.ErrUnexpectedEOF, err)
+}
--- a/pkg/objstore/gcs/gcs_test.go
+++ b/pkg/objstore/gcs/gcs_test.go
@@ -1,3 +1,6 @@
+// Copyright (c) The Thanos Authors.
+// Licensed under the Apache License 2.0.
+
 package gcs
 
 import (
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -22,6 +22,7 @@ We use _breaking :warning:_ to mark changes that are not backward compatible (re
 ### Fixed
 
 - [#3773](https://github.com/thanos-io/thanos/pull/3773) Compact: Pad compaction planner size check
+- [#3795](https://github.com/thanos-io/thanos/pull/3795) s3: A truncated "get object" response is reported as error.
 - [#3814](https://github.com/thanos-io/thanos/pull/3814) Store: Decreased memory utilisation while fetching block's chunks.
 - [#3815](https://github.com/thanos-io/thanos/pull/3815) Receive: Improve handling of empty time series from clients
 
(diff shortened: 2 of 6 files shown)


pkg/objstore/gcs/gcs_test.go
```diff
@@ -0,0 +1,47 @@
...
+	"os"
+	"testing"
+
+	"github.com/go-kit/kit/log"
+
+	"github.com/thanos-io/thanos/pkg/testutil"
+)
+
+func TestBucket_Get_ShouldReturnErrorIfServerTruncateResponse(t *testing.T) {
+	srv := httptest.NewServer(http.HandlerFunc(func(w http.ResponseWriter, r *http.Request) {
```
bwplotka, line 21: Does it make sense to put to acceptance test? 🤔

bwplotka (approved):
Nice, amazing. I would be happy to merge this with `Skip` on s3 that is waiting on minio fix or any other solution.

This sounds like an annoying bug, we cannot easily mitigate on our side. Let's see if we can have minio maintainers helping us this asap, otherwise, we can try our own investigation ):

pracucci:
> This sounds like an annoying bug, we cannot easily mitigate on our side. Let's see if we can have minio maintainers helping us this asap, otherwise, we can try our own investigation ):

@bwplotka Why did Thanos pick the Minio client instead of the official AWS SDK? I'm sure there's a good reason but I don't know which one.

pracucci:
> This sounds like an annoying bug, we cannot easily mitigate on our side

Not sure about this. I think we can. We could wrap the reader returned by `Get()`, initialise the wrapper with the expected size (the `GetObject()` also include the object size) and count the read byte. If once we hit the `EOF` the number of read bytes != expected, then we return a `io. ErrUnexpectedEOF`.

WDYT?

harshavardhana:
So I tested with master branch and latest release with minio-go the test seems to return appropriate error for truncated body.

```go
package minio

import (
        "context"
        "io"
        "io/ioutil"
        "net/http"
        "net/http/httptest"
        "testing"
)

func TestGetObjectReturnErrorIfServerTruncatesResponse(t *testing.T) {
        srv := httptest.NewServer(http.HandlerFunc(func(w http.ResponseWriter, r *http.Request) {
                w.Header().Set("Last-Modified", "Wed, 21 Oct 2015 07:28:00 GMT")
                w.Header().Set("Content-Length", "100")

                // Write less bytes than the content length.
                w.Write([]byte("12345"))
        }))
        defer srv.Close()

        // New - instantiate minio client with options
        clnt, err := New(srv.Listener.Addr().String(), &Options{})

        obj, err := clnt.GetObject(context.Background(), "bucketName", "objectName", GetObjectOptions{})
        if err != nil {
                t.Fatal(err)
        }

        // We expect an error when reading back.
        if _, err = ioutil.ReadAll(obj); err != io.ErrUnexpectedEOF {
                t.Fatalf("Expected %v, got %v", io.ErrUnexpectedEOF, err)
        }
}
```

```
~  go test -v -run TestGetObjectReturnErrorIfServerTruncatesResponse
=== RUN   TestGetObjectReturnErrorIfServerTruncatesResponse
--- PASS: TestGetObjectReturnErrorIfServerTruncatesResponse (0.00s)
PASS
ok      github.com/minio/minio-go/v7    0.004s
```

harshavardhana:
My bad reproduced it, taking a look what can be done.

pracucci:
> My bad reproduced it, taking a look what can be done.

@harshavardhana I upgraded to Minio 7.0.9, which was supposed to fix it. Unfortunately the unit tests I've added is still failing. I think the reason is because the fix done Minio affects only the first call to `Read()` while in Thanos we call `Read()` multiple times and only the last call has the opportunity to check if the total bytes read match the expected one. Am I correct / wrong?

harshavardhana:
> @harshavardhana I upgraded to Minio 7.0.9, which was supposed to fix it. Unfortunately the unit tests I've added is still failing. I think the reason is because the fix done Minio affects only the first call to `Read()` while in Thanos we call `Read()` multiple times and only the last call has the opportunity to check if the total bytes read match the expected one. Am I correct / wrong?

Yes it is possible to fix this situation as well @pracucci - it may be slightly more involved. I will look into it.

goober:
@harshavardhana do I understand it correctly that this should have been fixed by https://github.com/minio/minio-go/pull/1453 and released in minio-go 7.0.10 ?

pkg/objstore/gcs/gcs_test.go
pracucci: The acceptance test uses the real S3/GCS backend. To simulate this case we need a mocked one. Am I missing anything, or maybe I don't understand what you mean?

pracucci:
> Yes it is possible to fix this situation as well @pracucci - it may be slightly more involved. I will look into it.

 Thanks @harshavardhana for your hard work. Looks like Minio 7.0.10 is solving the issue 🤞

pracucci:
@bwplotka I've re-asked you the review because the purpose of the PR has changed from "add a test" to "fix it".

pkg/objstore/gcs/gcs_test.go
bwplotka: You are right, it's hard to mock, but that what I meant

bwplotka (approved):
Amazing! LGTM, sounds like we are having nice friends on minio side. 

Next step: retries on minio ;p

PR #3796, opened 2021-02-12
Reduce memory allocations in bucketBlock.readChunkRange()

pracucci:

## Changes

We're running the Cortex blocks storage at an increasingly large scale. We're currently investigating how to reduce the memory utilisation in the store-gateway and a low hanging fruit I've noticed is that `bucketBlock.readChunkRange()` waste memory allocations to grow the buffer if the requested length is larger than the largest bucket in the chunk pool (which is 32MB):

Screenshot 2021-02-12 at 18 34 31

The reason is that to ensure that `io.Copy()` doesn't grow the buffer, the buffer needs to be at least `bytes.MinRead` larger than the actual size to write.

This PR fixes it.

I've also added a benchmark, comparing it with `master`:

```
benchmark                                  old ns/op     new ns/op     delta
BenchmarkBucketBlock_readChunkRange-12     89789         70651         -21.31%

benchmark                                  old allocs     new allocs     delta
BenchmarkBucketBlock_readChunkRange-12     12             12             +0.00%

benchmark                                  old bytes     new bytes     delta
BenchmarkBucketBlock_readChunkRange-12     296676        204791        -30.97%
```

## Verification

Bencharmk.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -21,10 +21,7 @@ We use _breaking :warning:_ to mark changes that are not backward compatible (re
 ### Fixed
 
 - [#3773](https://github.com/thanos-io/thanos/pull/3773) Compact: Pad compaction planner size check
-
-### Fixed
-
-- [#3773](https://github.com/thanos-io/thanos/pull/3773) Compact: Pad compaction planner size check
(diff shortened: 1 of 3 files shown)


GiedriusS (approved):
Great work!

PR #3798, opened 2021-02-14
reloader: fix test's flakiness

### LINKED ISSUE #2622: ci/test: Flaky reloader test.
https://circleci.com/gh/thanos-io/thanos/9139?utm_campaign=vcs-integration-link&utm_medium=referral&utm_source=github-build-link

I remember @lilic was touching this test lately, so you mavbe you have idea? I  did not check closer yet (:

GiedriusS:
This is a fix of the `TestReloader_DirectoriesApply` test's
flakiness. I have been able to spot the problem locally after
introducing a few more tests on the metric values.

I have noticed that sometimes I'd get different `inotify` events.

For example here are two different runs:
```
"/tmp/TestReloader_DirectoriesApply249640168/001/rule2.yaml": CREATE
"/tmp/TestReloader_DirectoriesApply249640168/001/rule2.yaml": WRITE
"/tmp/TestReloader_DirectoriesApply249640168/001/rule1.yaml": WRITE
"/tmp/TestReloader_DirectoriesApply249640168/001/rule1.yaml": WRITE
"/tmp/TestReloader_DirectoriesApply249640168/001/rule3.yaml": CREATE
"/tmp/TestReloader_DirectoriesApply249640168/001/rule3.yaml": CREATE
"/tmp/TestReloader_DirectoriesApply249640168/001/rule-dir/rule4.yaml": WRITE
"/tmp/TestReloader_DirectoriesApply249640168/001/rule-dir/rule4.yaml": WRITE
```

```
"/tmp/TestReloader_DirectoriesApply364923838/001/rule2.yaml": CREATE
"/tmp/TestReloader_DirectoriesApply364923838/001/rule2.yaml": WRITE
"/tmp/TestReloader_DirectoriesApply364923838/001/rule1.yaml": WRITE
"/tmp/TestReloader_DirectoriesApply364923838/001/rule1.yaml": WRITE
"/tmp/TestReloader_DirectoriesApply364923838/001/rule3.yaml": CREATE
"/tmp/TestReloader_DirectoriesApply364923838/001/rule3.yaml": CREATE
"/tmp/TestReloader_DirectoriesApply364923838/001/rule-dir/rule4.yaml": WRITE
```

This happens because `inotify` _can_ coalesce different events into one.
In this case, `ioutil.WriteFile()` at first opens the given path with O_TRUNC
and then writes whatever content has been passed. This may or may not result
in two different write events.

Workaround this problem by storing the new content somewhere else locally
at first and then using `os.Rename` to move it into place.

Another issue is that we were reading a variable without almost any protection.
In fact, here is the problematic part:

```
		reloads.Store(reloads.Load().(int) + 1) // The only writer.
```

In fact, when the HTTP's handler lags a bit behind then we might get inconsistent
results. Protect it properly by using `sync.Mutex`.

Closes #2622.

## Changes

Protected the HTTP handler's code in a test with a mutex. Ensures that the writes are "atomic" by using `os.Rename`. Added checks on metrics.

## Verification

The test passes locally consistently. Tested this on CI >=5 times and couldn't reproduce whereas before it was almost consistent.

### DIFF
--- a/pkg/reloader/reloader_test.go
+++ b/pkg/reloader/reloader_test.go
@@ -20,6 +20,8 @@ import (
 	"time"
 
 	"github.com/go-kit/kit/log"
+	"github.com/prometheus/client_golang/prometheus"
+	promtest "github.com/prometheus/client_golang/prometheus/testutil"
 	"go.uber.org/atomic"
 	"go.uber.org/goleak"
 
@@ -168,11 +170,15 @@ func TestReloader_DirectoriesApply(t *testing.T) {
 	l, err := net.Listen("tcp", "localhost:0")
 	testutil.Ok(t, err)
 
-	reloads := &atomic.Value{}
-	reloads.Store(0)
 	i := 0
+	reloads := 0
+	reloadsMtx := sync.Mutex{}
+
 	srv := &http.Server{}
 	srv.Handler = http.HandlerFunc(func(resp http.ResponseWriter, r *http.Request) {
+		reloadsMtx.Lock()
+		defer reloadsMtx.Unlock()
+
 		i++
 		if i%2 == 0 {
 			// Fail every second request to ensure that retry works.
@@ -180,7 +186,7 @@ func TestReloader_DirectoriesApply(t *testing.T) {
 			return
 		}
 
-		reloads.Store(reloads.Load().(int) + 1) // The only writer.
+		reloads++
 		resp.WriteHeader(http.StatusOK)
 	})
 	go func() {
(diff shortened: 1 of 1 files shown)


GiedriusS:
Actually, it's still not helping. Since it's getting late, I will keep this PR open & will keep digging. I am going to make it ready for review once I cannot reproduce it anymore.

GiedriusS:
I think I was able to find the problems:

* `i` and `reloads` were accessed with almost no protection; loading-then-storing without any locks doesn't guarantee that we will get consistent results. The HTTP handler might get invoked more than once at any time if it lags behind;
* `ioutil.WriteFile()` might or might not generate duplicate events - depending on whether `inotify` coalesces them together. I didn't know this before but opening a file with O_TRUNC apparently also counts as a write event: https://man7.org/linux/man-pages/man7/inotify.7.html. To fix this problem I have created a few other temporary files that are later used with `os.Rename`. This way we get consistent write events.

In fact, you can try this PR out _without_ the fixes but with the extra metric values checks. You should be able to reproduce it locally then.

PTAL @thanos-io/thanos-maintainers

prmsrswt (approved):
LGTM! :fire: :fire:

matej-g:
I do not mean to cause additional headache, but I'm putting my observation here - it seems like the test can still flake on occasion, see a run from PR with no code changes - https://app.circleci.com/pipelines/github/thanos-io/thanos/6819/workflows/2e22a9a2-410c-495d-869b-8f9eeeb5ad7e/jobs/15821. Looks like the test skips performing a step and fails subsequently.

GiedriusS:
> I do not mean to cause additional headache, but I'm putting my observation here - it seems like the test can still flake on occasion, see a run from PR with no code changes - https://app.circleci.com/pipelines/github/thanos-io/thanos/6819/workflows/2e22a9a2-410c-495d-869b-8f9eeeb5ad7e/jobs/15821. Looks like the test skips performing a step and fails subsequently.

Maybe there's another place like this that might or might not generate inotify events. Help needed :hugs:

PR #3800, opened 2021-02-15
cleaning comments

Sudhar287:

## Changes

Cleaning comments.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -20,6 +20,7 @@ We use _breaking :warning:_ to mark changes that are not backward compatible (re
 
 ### Fixed
 
+- [#3800](https://github.com/thanos-io/thanos/pull/3800) Cleaning comments.
 - [#3773](https://github.com/thanos-io/thanos/pull/3773) Compact: Pad compaction planner size check
 - [#3796](https://github.com/thanos-io/thanos/pull/3796) Store: Decreased memory allocations while fetching block's chunks.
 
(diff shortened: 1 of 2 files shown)


pkg/store/cache/inmemory_test.go
```diff
@@ -45,19 +45,14 @@ max_item_size: 2KB
 	testutil.Equals(t, uint64(1024*1024), cache.maxSizeBytes)
 	testutil.Equals(t, uint64(2*1024), cache.maxItemSizeBytes)
 
-	// Should instance an in-memory index cache with specified YAML config.s with units.
+	// Should return error when max_size is lesser than max_item_size of cache.
```
squat, line 48: Grammar nit: `less` is used for comparing whereas `lesser` is an adjective.

```suggestion
	// Should return error when max_size is less than max_item_size of cache.
```

CHANGELOG.md
```diff
@@ -20,6 +20,7 @@ We use _breaking :warning:_ to mark changes that are not backward compatible (re
 
 ### Fixed
 
+- [#3800](https://github.com/thanos-io/thanos/pull/3800) Cleaning comments.
```
squat, line 23: We don't need to add this here. We like to keep the change log minimal and user-focussed, meaning it highlights external changes that the user of the software can notice rather than every single internal change and cleanup

yeya24:
Hi @Sudhar287, please let me know if you are still interested in continuing this pr. Thanks for the contribution!

PR #3801, opened 2021-02-15
Refactoring: allow to pass BytesPool to store.NewBucketStore()

pracucci:

## Changes

In Cortex we have 1 bucket store for each tenant. Given the chunks bytes pool is created in the `NewBucketStore()` we're currently inefficiently reusing byte slices across different tenants.

In this PR I'm doing a little refactoring to pass `pool.BytesPool` to `store.NewBucketStore()` so that we can re-use it across different tenants.

## Verification

Existing tests.


pstibrany (approved):

bwplotka (approved):
Makes, sense. LGTM!

PR #3802, opened 2021-02-15
Allow to customise the partitioner used by the BucketStore

pracucci:

## Changes

In Cortex, we would like to experiment around the partitioner. If we can find any better algorithm, we'll then propose it to Thanos.

Can we refactor `store.NewBucketStore()` to accept the `Partitioner` interface in input so we can give a try to alternative implementations?

## Verification

N/A


pstibrany (approved):
LGTM

bwplotka (approved):
Thanks, yes, some help here would be awesome!

PR #3805, opened 2021-02-16
Fix command to find react source files

hitanshu-mehta:

## Changes

Sometimes `make assets` command does not regenerate static assets of react app, so those changes are not reflected during testing the ui.
`shell find` seems to solve this issue.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -58,7 +58,7 @@ PUBLIC_DIR        ?= $(WEB_DIR)/public
 ME                ?= $(shell whoami)
 
 REACT_APP_PATH = pkg/ui/react-app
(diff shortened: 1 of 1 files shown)


yeya24 (commented):
Thanks for the PR. I am not familiar with the frontend part. I am curious about that what's the difference of this change?

@prmsrswt Can you please take a look? Thanks!

prmsrswt (approved):
@yeya24 The problem is that sometimes the `make assets` target doesn't rebuild the React code even though there are changes. We are adapting https://github.com/prometheus/prometheus/blob/master/Makefile#L18 (Prometheus had the same problem and this was the fix)

Issue #3806, opened 2021-02-16
Compact all available plans for a single group concurrently


andrejbranch:
### Is your proposal related to a problem?

We are trying to scale the cortex compactor and we have a really large tenant that continues to grow. One problem I noticed is if the compactor fails for some reason and falls way behind it will still only compact 1 set of overlapping blocks at a time instead of concurrently compacting all known overlapping blocks.

### Describe the solution you'd like

 It would help the compactor keep up if we had a way to concurrently compact all available plans for a single group. This would also help when vertical compaction takes a similar amount of time as a blocks time range, as when a new set of overlapping blocks is uploaded the compactor can start another compaction even though one for the same group is already running.

My proposed solution is to move the planning side of the code out of the compaction side and instead have the grouper return a group for every plan available.

### Describe alternatives you've considered

An alternative I considered was changing the planner so that it returns a list of plans and have a new concurrency in the compact code but this is messier and changes the Planner interface.

andrejbranch:
PR for proposed changes https://github.com/thanos-io/thanos/pull/3807

dswarbrick:
This topic interests me, as I've often watched a thanos compactor plod its way through catching up on compactions / downsampling, and wondered why it seems to do a lot of unnecessary work.

For example, assume that the compactor has fallen behind, and you have five 8h blocks, and four (or more) 2h blocks. Theoretically, the compactor could download all those blocks and spit out a 2d block directly. But instead, I've seen it download the four 2h blocks, compact them into an 8h block, and upload it. The work directory is wiped, needlessly throwing away the just-processed 8h block that it's going to need for the next step. _Then_ it downloads the six 8h blocks and compacts them.

The downsampling is another area where the compactor seems to be a traffic generator for the S3 provider. It appears to download the raw block, downsample it to 5m resolution and upload it. A little bit later, it re-downloads the 5m resolution block, downsamples it to 1h resolution, and uploads that. Wouldn't it make more sense to write out the 5m and 1h resolution blocks _concurrently_ so as to save some S3 traffic, and also a save a lot of read IO?

I hope this behaviour isn't simply due to me configuring it wrong.

bwplotka:
@dswarbrick Yes, that's the naive planning Prometheus had, which we inherited. You are welcome to help to find something more scalable! Worked well so far thanks to being... simple! 🙃 

The issue @andrejbranch mentioned might be different and we seen that too - it's very naive for overlaps too. Let's fix both!

[...]

PR #3807, opened 2021-02-16
Compaction group concurrency

andrejbranch:

## Changes
- Planner is moved out of compaction code and instead the grouper uses the planner to recursively call planner for each group, returning a new set of groups, 1 group for each plan available
- Grouper now depends on planner

## Verification

We have a dev cortex cluster using block storage that was 60 days behind, an ideal situation for testing this code. I was able to catch up 40 times faster as I ran this with group concurrency set to 40. Everything looks good so far.

### DIFF
--- a/pkg/compact/compact.go
+++ b/pkg/compact/compact.go
@@ -207,6 +207,11 @@ type Grouper interface {
 	// Groups returns the compaction groups for all blocks currently known to the syncer.
 	// It creates all groups from the scratch on every call.
 	Groups(blocks map[ulid.ULID]*metadata.Meta) (res []*Group, err error)
+	// Starting a group lets the grouper know that the blocks in the groups are currently
+	// being compacted and to ignore them in future plans.
+	StartGroup(group *Group)
+	// Complete the group lets the grouper know that the blocks in this group can again be planned.
+	CompleteGroup(group *Group)
 }
 
 // DefaultGroupKey returns a unique identifier for the group the block belongs to, based on
@@ -222,18 +227,22 @@ func defaultGroupKey(res int64, lbls labels.Labels) string {
 // DefaultGrouper is the Thanos built-in grouper. It groups blocks based on downsample
 // resolution and block's labels.
 type DefaultGrouper struct {
-	bkt                      objstore.Bucket
-	logger                   log.Logger
-	acceptMalformedIndex     bool
-	enableVerticalCompaction bool
-	compactions              *prometheus.CounterVec
-	compactionRunsStarted    *prometheus.CounterVec
-	compactionRunsCompleted  *prometheus.CounterVec
-	compactionFailures       *prometheus.CounterVec
-	verticalCompactions      *prometheus.CounterVec
-	garbageCollectedBlocks   prometheus.Counter
-	blocksMarkedForDeletion  prometheus.Counter
-	hashFunc                 metadata.HashFunc
+	bkt                             objstore.Bucket
+	logger                          log.Logger
+	acceptMalformedIndex            bool
+	enableVerticalCompaction        bool
+	enableConcurrentGroupCompaction bool
+	compactions                     *prometheus.CounterVec
+	compactionRunsStarted           *prometheus.CounterVec
+	compactionRunsCompleted         *prometheus.CounterVec
+	compactionFailures              *prometheus.CounterVec
+	verticalCompactions             *prometheus.CounterVec
+	garbageCollectedBlocks          prometheus.Counter
+	blocksMarkedForDeletion         prometheus.Counter
+	hashFunc                        metadata.HashFunc
+	planner                         Planner
+	locked                          map[ulid.ULID]bool
+	lockedMtx                       sync.Mutex
 }
 
 // NewDefaultGrouper makes a new DefaultGrouper.
@@ -242,16 +251,19 @@ func NewDefaultGrouper(
 	bkt objstore.Bucket,
 	acceptMalformedIndex bool,
 	enableVerticalCompaction bool,
+	enableConcurrentGroupCompaction bool,
 	reg prometheus.Registerer,
 	blocksMarkedForDeletion prometheus.Counter,
 	garbageCollectedBlocks prometheus.Counter,
 	hashFunc metadata.HashFunc,
+	planner Planner,
 ) *DefaultGrouper {
 	return &DefaultGrouper{
-		bkt:                      bkt,
-		logger:                   logger,
-		acceptMalformedIndex:     acceptMalformedIndex,
-		enableVerticalCompaction: enableVerticalCompaction,
+		bkt:                             bkt,
+		logger:                          logger,
+		acceptMalformedIndex:            acceptMalformedIndex,
+		enableVerticalCompaction:        enableVerticalCompaction,
+		enableConcurrentGroupCompaction: enableConcurrentGroupCompaction,
 		compactions: promauto.With(reg).NewCounterVec(prometheus.CounterOpts{
 			Name: "thanos_compact_group_compactions_total",
 			Help: "Total number of group compaction attempts that resulted in a new block.",
@@ -275,12 +287,39 @@ func NewDefaultGrouper(
 		garbageCollectedBlocks:  garbageCollectedBlocks,
 		blocksMarkedForDeletion: blocksMarkedForDeletion,
 		hashFunc:                hashFunc,
+		planner:                 planner,
+		locked:                  map[ulid.ULID]bool{},
 	}
+}
+
+func (g *DefaultGrouper) StartGroup(group *Group) {
+	if !g.enableConcurrentGroupCompaction {
+		return
+	}
+	g.lockedMtx.Lock()
+	for _, m := range group.metasByMinTime {
+		g.locked[m.ULID] = true
+	}
+	g.lockedMtx.Unlock()
+}
+
+func (g *DefaultGrouper) CompleteGroup(group *Group) {
+	if !g.enableConcurrentGroupCompaction {
+		return
+	}
+	g.lockedMtx.Lock()
+	for _, m := range group.metasByMinTime {
+		delete(g.locked, m.ULID)
+	}
+	g.lockedMtx.Unlock()
 }
 
 // Groups returns the compaction groups for all blocks currently known to the syncer.
 // It creates all groups from the scratch on every call.
 func (g *DefaultGrouper) Groups(blocks map[ulid.ULID]*metadata.Meta) (res []*Group, err error) {
+	if g.enableConcurrentGroupCompaction {
+		return g.concurrentGroups(blocks)
+	}
 	groups := map[string]*Group{}
 	for _, m := range blocks {
 		groupKey := DefaultGroupKey(m.Thanos)
@@ -295,6 +334,7 @@ func (g *DefaultGrouper) Groups(blocks map[ulid.ULID]*metadata.Meta) (res []*Gro
 				m.Thanos.Downsample.Resolution,
 				g.acceptMalformedIndex,
 				g.enableVerticalCompaction,
+				g.enableConcurrentGroupCompaction,
 				g.compactions.WithLabelValues(groupKey),
 				g.compactionRunsStarted.WithLabelValues(groupKey),
 				g.compactionRunsCompleted.WithLabelValues(groupKey),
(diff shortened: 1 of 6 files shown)


pkg/compact/compact_e2e_test.go
```diff
@@ -303,17 +303,11 @@ func TestGroup_Compact_e2e(t *testing.T) {
 		testutil.Equals(t, 0.0, promtest.ToFloat64(grouper.compactions.WithLabelValues(DefaultGroupKey(metas[4].Thanos))))
 		testutil.Equals(t, 0.0, promtest.ToFloat64(grouper.compactions.WithLabelValues(DefaultGroupKey(metas[5].Thanos))))
 		testutil.Equals(t, 4, MetricCount(grouper.compactionRunsStarted))
-		testutil.Equals(t, 2.0, promtest.ToFloat64(grouper.compactionRunsStarted.WithLabelValues(DefaultGroupKey(metas[0].Thanos))))
-		testutil.Equals(t, 2.0, promtest.ToFloat64(grouper.compactionRunsStarted.WithLabelValues(DefaultGroupKey(metas[7].Thanos))))
-		// TODO(bwplotka): Looks like we do some unnecessary loops. Not a major problem but investigate.
```
andrejbranch, line 308: this seems to have been solved by my change

pkg/compact/compact.go
```diff
@@ -206,6 +208,11 @@ type Grouper interface {
 	// Groups returns the compaction groups for all blocks currently known to the syncer.
 	// It creates all groups from the scratch on every call.
 	Groups(blocks map[ulid.ULID]*metadata.Meta) (res []*Group, err error)
+	// Starting a group lets the grouper know that the blocks in the groups are currently
+	// being compacted and to ignore them in future plans
+	StartGroup(group *Group)
```
bwplotka, line 213: Why not have `Group.Lock` method? 🤗

pkg/compact/compact.go
```diff
@@ -271,14 +282,35 @@ func NewDefaultGrouper(
...
 		blocksMarkedForDeletion: blocksMarkedForDeletion,
+		planner:                 planner,
+		planned:                 map[ulid.ULID]bool{},
+	}
+}
+
+func (g *DefaultGrouper) StartGroup(group *Group) {
+	g.plannedMtx.Lock()
+	for _, m := range group.metasByMinTime {
+		g.planned[m.ULID] = true
```
bwplotka, line 293:
```suggestion
		g.locked[m.ULID] = true
```

pkg/compact/compact.go
```diff
@@ -309,18 +342,81 @@ func (g *DefaultGrouper) Groups(blocks map[ulid.ULID]*metadata.Meta) (res []*Gro
 			return nil, errors.Wrap(err, "add compaction group")
 		}
 	}
+
+	if g.planner != nil {
+		newGroups := []*Group{}
+		// split each group by available plans
```
bwplotka, line 348: Comments should be full sentence as per out Style guide 🤗  (Upper case letter and trailing period)

pkg/compact/compact.go
```diff
@@ -309,18 +342,81 @@ func (g *DefaultGrouper) Groups(blocks map[ulid.ULID]*metadata.Meta) (res []*Gro
 			return nil, errors.Wrap(err, "add compaction group")
 		}
 	}
+
+	if g.planner != nil {
```
bwplotka, line 346: Can planner be .. nil?

pkg/compact/compact.go
```diff
@@ -309,18 +342,81 @@ func (g *DefaultGrouper) Groups(blocks map[ulid.ULID]*metadata.Meta) (res []*Gro
...
+	// split each group by available plans
+	var seededRand = rand.New(rand.NewSource(time.Now().UnixNano()))
+	for toCompact, err := g.planner.Plan(context.Background(), group.metasByMinTime); len(toCompact) != 0; toCompact, err = g.planner.Plan(context.Background(), group.metasByMinTime) {
+		if err != nil {
+			return nil, errors.Wrap(err, "split group")
+		}
+		lbls := group.Labels()
+		hash := strconv.Itoa(seededRand.Int())
+		groupKey := group.Key()
+		newGroup, err := NewGroup(
```
bwplotka, line 378: Can we reuse group? This feel a bit wrong that we create group only to then split by group and create again 🤗

pkg/compact/compact.go
```diff
@@ -309,18 +342,81 @@ func (g *DefaultGrouper) Groups(blocks map[ulid.ULID]*metadata.Meta) (res []*Gro
...
 		return res[i].Key() < res[j].Key()
 	})
 	return res, nil
 }
 
+// Recursively plan a groups metas until there are no available plans and return a new group for each plan
+// This allows a single group key to utilize group concurrency when there are multiple plans available to compact
+func (g *DefaultGrouper) splitGroupByPlans(group *Group) ([]*Group, error) {
+	groups := []*Group{}
+	// split each group by available plans
```
bwplotka, line 369: Please make comments full sentence 🤗

pkg/compact/compact.go
```diff
@@ -309,18 +342,81 @@ func (g *DefaultGrouper) Groups(blocks map[ulid.ULID]*metadata.Meta) (res []*Gro
...
+	var seededRand = rand.New(rand.NewSource(time.Now().UnixNano()))
+	for toCompact, err := g.planner.Plan(context.Background(), group.metasByMinTime); len(toCompact) != 0; toCompact, err = g.planner.Plan(context.Background(), group.metasByMinTime) {
```
bwplotka, line 371: I would suggest doing  `for {` - line might be too long, to have it readable.

bwplotka, line 371: It's hard to read in this long line, but not sure if this works as intended: I would expect here some code that would deduce that the next compaction will output block that can be use in next compaction so we could just do the next one instead 🤔

pkg/compact/compact.go
```diff
@@ -309,18 +342,81 @@ func (g *DefaultGrouper) Groups(blocks map[ulid.ULID]*metadata.Meta) (res []*Gro
...
+			return nil, errors.Wrap(err, "create split compaction group")
+		}
+		groups = append(groups, newGroup)
+		for _, m := range toCompact {
+			if err := newGroup.Add(m); err != nil {
+				return nil, errors.Wrap(err, "add to split compaction group")
+			}
+			for i, mm := range group.metasByMinTime {
+				if mm.ULID == m.ULID {
+					group.metasByMinTime = append(group.metasByMinTime[:i], group.metasByMinTime[i+1:]...)
```
bwplotka, line 405: I wonder if this work well. 

1. First of all it does not detect places where output block has to be used for next compaction, that results in unnecessary compaction without output block causing vertical compaction later on.
2.  This still does not detect  case mentioned in issue which is overlapped blocks not being grouped into single compact, no?

pkg/compact/compact_e2e_test.go
```diff
@@ -393,6 +387,122 @@ func TestGroup_Compact_e2e(t *testing.T) {
...
+		comp, err := tsdb.NewLeveledCompactor(ctx, reg, logger, []int64{1000, 3000}, nil)
+		testutil.Ok(t, err)
+
+		planner := NewTSDBBasedPlanner(logger, []int64{1000, 3000})
+
+		grouper := NewDefaultGrouper(logger, bkt, false, true, reg, blocksMarkedForDeletion, garbageCollectedBlocks, planner)
+		bComp, err := NewBucketCompactor(logger, sy, grouper, planner, comp, dir, bkt, 2)
+		testutil.Ok(t, err)
+		extLabels := labels.Labels{{Name: "e1", Value: "1"}}
+		metas := createAndUpload(t, bkt, []blockgenSpec{
```
bwplotka, line 426: This is great, but I would also vote for smaller grouper/planner unit test, it's quite critical path (compactions are irreversible)

pkg/compact/compact.go
```diff
@@ -904,7 +1004,9 @@ func (c *BucketCompactor) Compact(ctx context.Context) (rerr error) {
 			go func() {
 				defer wg.Done()
 				for g := range groupChan {
+					c.grouper.StartGroup(g)
```
bwplotka, line 1007: Why we need that BTW? Do we ever plan/group and compact in the same time? 🤔

bwplotka (commented):
Thanks for this!

This looks quite good, but I wonder if grouper is the right approach. Looks like we are taking planner as blackbox and trying to mitigate things, where actually planner is the thing that should do the most of work (e.g consider bigger group first, not just next one).

Also I don' think this fixes https://github.com/thanos-io/thanos/issues/3806#issuecomment-780017110, this is orthogonal, right?

pkg/compact/compact.go
andrejbranch: this will need to be removed, it was put in place because this test breaks and needs to be updated https://github.com/thanos-io/thanos/pull/3807/files#diff-c2110623e0808d8a764bbe216b3a23fa19bf897238848a977810672b0140fb6fR138

pkg/compact/compact.go
andrejbranch: The reason for this is compacting methods return shouldRerunGroup as true and when i first tested this it seemed that after one group completed the planner would see blocks that are currently being compacted and add groups for them. So start group keeps track of whats currently being compacted and when skip those here https://github.com/thanos-io/thanos/pull/3807/files#diff-be21b0381ac14ee139355d057b390aa69ff338fe1c748ae8a475ce46f3583575R311

pkg/compact/compact.go
andrejbranch: agree hard to read and needs to be refactored. This is to recursively get all available plans for 1 group by removing metas from the parent group on every iteration so eventually the planner has nothing left to plan for that group. It does not deduce future plans as that is a separate issue that needs I wasn't targeting in this feature. Should I?

pkg/compact/compact.go
andrejbranch: Yeah this can be refactored to not use group, all thats really needed is the group key and the metasByMinTime

pkg/compact/compact.go
andrejbranch: 1. Again this feature wasn't meant to deduce what can be used for next compactions. If thats desired, I'll have to put some thought into that. 
2. Im not sure what you mean by that. This line of code is to remove metas from the parent group that have already been planned into a new group. The case mentioned in the issue is that you are behind and have multiple windows of overlapping blocks that can be compacted concurrently.

pkg/compact/compact_e2e_test.go
andrejbranch: agreed

pkg/compact/compact.go
andrejbranch: Will update. Thanks for taking a look at these changes. I agree most of the work should be done by the planner, the reason i didn't go that path is the number of changes required would be substantial. Happy to go down that path if it's preferred

andrejbranch: After a second look Group lock wont work because its the grouper that needs to know about whats currently running.

pkg/compact/compact.go
andrejbranch: After second look reusing group seems like the most efficient way. If you would rather i can make a new type like GroupSplitter or something like that but it will have some redundancy because i will need to duplicate the metasByMinTime sorting and some other fields. Happy to change it if thats preferred

pkg/compact/compact.go
andrejbranch: I will create a separate issue to handle predicting what blocks will be needed next and build on this work.

pkg/compact/compact.go
andrejbranch: After looking into this more, this change I think is still necessary. Say that compaction iterations are set to happen every 30 minutes but a large tenant has compaction runs that take more than 30 minutes. Yes you can increase the compaction frequency but this change would prevent a new iteration trying to compact blocks that are already being compacted.

test/e2e/compact_test.go
```diff
@@ -629,7 +637,13 @@ func TestCompactWithStoreGateway(t *testing.T) {
 		testutil.Ok(t, err)
 		operationMatcher, err := labels.NewMatcher(labels.MatchEqual, "operation", "get")
 		testutil.Ok(t, err)
-		testutil.Ok(t, c.WaitSumMetricsWithOptions(e2e.Equals(478),
+		var expectedOps float64
+		if groupConcurrencyArg == "--no-compact.enable-concurrent-group-compaction" {
+			expectedOps = 478
+		} else {
+			expectedOps = 480
```
andrejbranch, line 644: this is the only metric that the new feature made a couple extra get calls.

pkg/compact/compact.go
```diff
@@ -320,37 +363,161 @@ func (g *DefaultGrouper) Groups(blocks map[ulid.ULID]*metadata.Meta) (res []*Gro
...
+		return res[i].Key() < res[j].Key()
+	})
+	return res, nil
+}
+
+// Recursively plan a groups metas until there are no available plans and return a new group for each plan.
+// This allows a single group key to utilize group concurrency when there are multiple plans available to compact.
+func (g *DefaultGrouper) splitGroupByPlans(group *Group) ([]*Group, error) {
+	groups := []*Group{}
+	var seededRand = rand.New(rand.NewSource(time.Now().UnixNano()))
```
wiardvanrij, line 429: This catches my attention :D  Side-note here; I'm not an expert on this part, so I could be totally wrong. ^^

Should a seedRand only be initialised once? Since this func is called multiple times from other loops. Then we can use https://golang.org/pkg/math/rand/#Intn to get a new pseudo-int. Which is I believe faster/easier. Especially if we do a loop that might (??) be done in nano second, you'll end up with the same seed.

To continue about easier; what about ulid for the hash? I'm not worried about collisions, but it's pretty easy and consistent with the usage of ulid for blocks. 

Not sure if relevant, but the oklog/ulid package has this statement too; 

> Please note that rand.Rand from the math package is not safe for concurrent use. Instantiate one per long living go-routine or use a sync.Pool if you want to avoid the potential contention of a locked rand.Source as its been frequently observed in the package level functions.

wiardvanrij (commented):
This looks really great. I think the real important parts are handled for compaction. Since you already ran this on your dev cluster, do you have any insights in the increased usage of the compactor component?  Hypothetically speaking, if the usage is x2, we should recommend people to adjust resources (and/or concurrency) accordingly.  See this as an extra 'helper' by information.

Also small personal nitpicky question on the rand hash 👍

pkg/compact/compact.go
andrejbranch: good call! I removed the hash completely and instead just used the first ULID in the compaction group to append to the sub directory name which also cleaned up needing hash property on group.

andrejbranch:
@wiardvanrij In our case we saw increased cpu and disk usage because we only have 1 large tenant and it could only use 1 thread at a time. Enabling this feature allowed for utilizing multiple threads. If a user has multiple relatively small tenants I don't think they will notice a difference in usage and might not even need this feature.

pkg/compact/compact.go
```diff
@@ -320,26 +360,143 @@ func (g *DefaultGrouper) Groups(blocks map[ulid.ULID]*metadata.Meta) (res []*Gro
...
+				return nil, errors.Wrap(err, "add to split compaction group")
+			}
+			ids[m.ULID] = true
+		}
+		for _, m := range group.metasByMinTime {
+			if _, ok := ids[m.ULID]; !ok {
+				newMetasByMinTime = append(newMetasByMinTime, m)
+			}
+		}
+		group.metasByMinTime = newMetasByMinTime
```
bwplotka, line 474: Ok, this is something I am the most worried about.

I think this algorithm might be quite ineffective in practice. The reason is that compaction usually depends on each other. I would need to double check but in my experience default compaction planner is pretty naive.

For example, if it sees let's say

```
[2d][8h][8h][2h][2h][2h][2h][2h]
```

First plan would say `[2d][8h][8h]  (let me compact those [2h][2h][2h][2h])   [2h]`

If you take them out, compact will assume gap and will compact concurrently  `[2d][8h][8h] <2d gap>  [2h]` into 2d block:  `[2d] ( let me compact those [8h][8h] <2d gap> ) [2h]`. This is because the compaction algorithm was designed for single group Prometheus one by one blocks.

Consequences for Cortex are lower: You just recompact multiple times unnecessarily. For Thanos for deployments that does not have vertical compaction enabled it will just fail compactor (overlapping blocks).

That's why I believe we have to look at a smarter planner that will give us the planned compactions that we can concurrently attack and will be able to give us feedback anytime (even if other compactions are running) on what to run.

Feel free to change the majority of the interface to accomplish this well with Cortex. We are happy in Thanos to default to the concurrent solution (with simply worker numbers = 1). So I wonder if the current grouper/planner/compaction fits here. Free to rearrange those 3 APIs from ground zero. It's better than just fiddling with something not fitting on existing not well prepared API for concurrency on this level.

bwplotka (changes requested):
Thanks for this! This looks like something we definitely want but I found some potential blockers. I think it's best if we reorganize planner/grouper/compact to something that makes sense more. Essentially our planner has to plan things accordinglity. Unfortunately we cannot assume currently the Planner interface can give you properly planned compaction without full overview of what is currently being compacted. Details in comments. 🤗 

Thanks for this work, I would love to collaborate on something that will work long term for both Cortex and thanos

kakkoyun:
Friendly ping @andrejbranch :relaxed: Could you give us some update? Do you still want to work on this? Shall we revive this?

PR #3808, opened 2021-02-17
Allow downstream projects to customise the Partitioner

pracucci:

## Changes

In https://github.com/thanos-io/thanos/pull/3802 I've exposed `Partitioner` to let Cortex trying to experiment with a different algorithm (in case we find anything interesting we'll upstream it to Thanos then). But `part` is still unexposed, which makes impossible to customise it.

## Verification

N/A


bwplotka (approved):
👍🏽

PR #3810, opened 2021-02-17
Upgraded react dep to resolve security issue.

bwplotka:

As per https://github.com/thanos-io/thanos/security/dependabot/pkg/ui/react-app/yarn.lock/immer/open

### DIFF
--- a/pkg/ui/react-app/package.json
+++ b/pkg/ui/react-app/package.json
@@ -24,6 +24,7 @@
     "enzyme-to-json": "^3.4.3",
     "fuzzy": "^0.1.3",
     "i": "^0.3.6",
+    "immer": "^8.0.1",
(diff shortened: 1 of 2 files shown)


prmsrswt:
Take a look at this comment https://github.com/thanos-io/thanos/pull/3770#discussion_r570243354

TLDR: Updating the dependency directly will not change anything, as it is an indirect dependency, we will still be using the older version in the build. We need to update the **direct** dependency due to which it got installed.

prmsrswt:
I think it might be a good idea to update all dependencies to the latest non-breaking version. WDYT?

kakkoyun:
@prmsrswt Do you want to create a PR for the #3811 ?

bwplotka:
Ack, I am stupid then. thanks for being here Prem (: Let me close my broken attempt (:

Issue #3772, opened 2021-02-05
receive: internal server error with when it receives metadata of metrics in remote write requests (Prometheus >=2.24)


mxmorin:
**Thanos, Prometheus and Golang version used**:

```
thanos, version 0.16.0 (branch: HEAD, revision: dc6a1666bb68cd5c11be54452842e823f57668c5)
  build user:       root@ba806318d94d
  build date:       20201026-13:56:53
  go version:       go1.15
```
**Object Storage Provider**:
S3

**What happened**:
After upgrading Prometheus from version 2.20 to 2.24, Internal server errors suddenly appeared in the logs with no other informations (even in debug mode).
I downgraded the Prometheus version to 2.20 and everything went back to normal.

Is thanos compatible with Prometheus 2.24 ?

**Full logs to relevant components**:
<details>Logs
<p>

```
Jan 28 06:13:22 p1thanos01 thanos[50433]: level=error ts=2021-01-28T05:13:22.69662862Z caller=handler.go:331 component=receive component=receive-handler err= msg="internal server error"
Jan 28 06:13:22 p1thanos01 thanos[50433]: level=error ts=2021-01-28T05:13:22.797967408Z caller=handler.go:331 component=receive component=receive-handler err= msg="internal server error"
Jan 28 06:13:22 p1thanos01 thanos[50433]: level=error ts=2021-01-28T05:13:22.899356197Z caller=handler.go:331 component=receive component=receive-handler err= msg="internal server error"
Jan 28 06:13:23 p1thanos01 thanos[50433]: level=error ts=2021-01-28T05:13:23.000757915Z caller=handler.go:331 component=receive component=receive-handler err= msg="internal server error"
Jan 28 06:13:23 p1thanos01 thanos[50433]: level=error ts=2021-01-28T05:13:23.102497326Z caller=handler.go:331 component=receive component=receive-handler err= msg="internal server error"
Jan 28 06:13:23 p1thanos01 thanos[50433]: level=error ts=2021-01-28T05:13:23.20422861Z caller=handler.go:331 component=receive component=receive-handler err= msg="internal server error"
Jan 28 06:13:23 p1thanos01 thanos[50433]: level=error ts=2021-01-28T05:13:23.305524863Z caller=handler.go:331 component=receive component=receive-handler err= msg="internal server error"
[...]

PR #3815, opened 2021-02-19
Receive: Improve handling of empty time series from clients

m-lawre:

## Changes

If an empty write request is passed into `fanoutForward`, no messages will be sent on the error channel (`ec`) causing the return value to be an empty/default initialised `MultiError`. The error is translated into an "internal server error" at the http handler.

This patch checks that the incoming time series contains data and if not returns immediately.

As to why newer prometheus versions have this behaviour, I'm not currently sure. I believe this will help with #3772.

## Verification

We've been running this in our staging environment for a week or so and we no longer see the internal error message being logged and in its place the info message instead.

### DIFF
--- a/pkg/receive/handler.go
+++ b/pkg/receive/handler.go
@@ -310,6 +310,12 @@ func (h *Handler) receiveHTTP(w http.ResponseWriter, r *http.Request) {
 		tenant = h.options.DefaultTenantID
 	}
 
+	// exit early if the request contained no data
+	if 0 == len(wreq.Timeseries) {
+		level.Info(h.logger).Log("msg", "empty timeseries from client", "tenant", tenant)
+		return
+	}
+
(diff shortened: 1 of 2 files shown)


pkg/receive/handler.go
```diff
@@ -310,6 +310,12 @@ func (h *Handler) receiveHTTP(w http.ResponseWriter, r *http.Request) {
 		tenant = h.options.DefaultTenantID
 	}
 
+	// exit early if the request contained no data
+	if 0 == len(wreq.Timeseries) {
```
squat, line 314: Nit: just to make the intention of the comparison clearer we prefer to put the unknown on the left side
```suggestion
	if len(wreq.Timeseries) == 0 {
```

squat (approved):
lgtm! Thanks!

squat (approved):
Great :) thanks again!

PR #3823, opened 2021-02-22
Add objstore.List() recursive support

### LINKED ISSUE #3784: Object store clients: List() recursively

Cortex internally uses the Thanos objstore clients. We have an use case for which we need to list objects recursively, basically setting `delimiter=""` instead of the default `delimiter="/"`.

I'm proposing to extend `objstore.BucketReader` interface adding the following:

```
type IterOptions struct {
    // When true the iteration ignores the '/' delimiter.
    Recursive bool
}

type BucketReader interface {
    IterWithOptions(ctx context.Context, dir string, options IterOptions, f func(string) error) error
}
```

- Add options directly to `Iter()` but would break signature add extra boilerplate to add to each `Iter()` call
- Add options are variadic at the end of `Iter()`, so `IterWithOptions(ctx context.Context, dir string, f func(string) error, options ...IterOption) error`: could work too, no big preference

### Additional context

Quick analysis of supported backends:
- S3
  - Minio already supports `Recursive` and we're currently hardcoding `Recursive: false`
- GCS
  - `storage.Query` allows to customise the `Delimiter` and we're currently hardcoding it to `/` (setting it to empty string returns the whole tree under the provided prefix, which is what we want with recursive = true)
- Azure
  - To list recursively we need to call `ListBlobsFlatSegment()` instead of `ListBlobsHierarchySegment()`
- Filesystem
  - We could switch from `ioutil.ReadDir()` to `filepath.Walk()` and recursively enter directories as soon as we find them (to keep lexicographically order guarantee)
- COS
  - If my chinese is good enough, according to the [doc](https://cloud.tencent.com/document/product/436/7734) it's just about passing an empty delimiter (supported by the SDK)
- OSS
  - Same here, according to [doc](https://www.alibabacloud.com/help/doc-detail/31965.htm) an empty delimiter will list recursively
- Swift
  - List objects with `swift.ObjectsOpts{ Delimiter: 0 }`

pracucci:

## Changes

As discussed in #3784, in this PR I'm proposing to add `IterOption` support to `objstore.Bucket.List()` supporting an option to enable recursive listing. This feature is not currently used by Thanos but would be used by Cortex (which depends on Thanos objstore clients).

Fixes #3784.

## Verification

Acceptance tests.


go.mod
```diff
@@ -71,6 +71,8 @@ require (
 )
 
 replace (
+	github.com/cortexproject/cortex => github.com/pracucci/cortex v0.0.0-20210222171122-5c74b43721b8
```
pracucci, line 74: To temporarily fix the Cortex compilation issue because of the changed `Iter()` signature. The little change is here:
https://github.com/thanos-io/thanos/compare/pracucci:add-list-recursively

I will open a follow up PR to remove this and upgrade Cortex once this PR will be merged (I will need to upgrade Thanos in Cortex, fix it in Cortex and then remove this override).

cmd/thanos/query_frontend.go
```diff
@@ -175,12 +174,6 @@ func runQueryFrontend(
 		return errors.Wrap(err, "error validating the config")
 	}
 
-	fe, err := cortexfrontendv1.New(cortexfrontendv1.Config{}, nil, logger, reg)
```
pracucci, line 178: Wasn't used by Thanos. Thanos use the tripperware below. Basically this code was doing nothing.

bwplotka: cc @yeya24 🤗

pkg/objstore/azure/azure.go
```diff
@@ -107,33 +107,52 @@ func NewBucket(logger log.Logger, azureConfig []byte, component string) (*Bucket
...
+func (b *Bucket) Iter(ctx context.Context, dir string, f func(string) error, options ...objstore.IterOption) error {
 	prefix := dir
 	if prefix != "" && !strings.HasSuffix(prefix, DirDelim) {
 		prefix += DirDelim
 	}
 
 	marker := blob.Marker{}
+	params := objstore.ApplyIterOptions(options...)
+
+	listOptions := blob.ListBlobsSegmentOptions{
```
bwplotka, line 119: nit: (was that before): we could just put this in single line

pkg/objstore/objstore_test.go
```diff
@@ -19,12 +19,12 @@ func TestMetricBucket_Close(t *testing.T) {
 	testutil.Equals(t, 7, promtest.CollectAndCount(bkt.opsDuration))
 
 	AcceptanceTest(t, bkt.WithExpectedErrs(bkt.IsObjNotFoundErr))
-	testutil.Equals(t, float64(6), promtest.ToFloat64(bkt.ops.WithLabelValues(OpIter)))
+	testutil.Equals(t, float64(9), promtest.ToFloat64(bkt.ops.WithLabelValues(OpIter)))
```
bwplotka, line 22: Why this changed? 🤔

pkg/objstore/testing.go
```diff
@@ -167,21 +169,48 @@ func AcceptanceTest(t *testing.T, bkt Bucket) {
 	sort.Strings(seen)
 	testutil.Equals(t, expected, seen)
 
+	// Can we iter over items from top dir recursively?
```
bwplotka, line 172: 👍🏽

go.mod
```diff
@@ -7,7 +7,9 @@ require (
 	github.com/Azure/azure-storage-blob-go v0.8.0
 	github.com/NYTimes/gziphandler v1.1.1
 	github.com/alecthomas/units v0.0.0-20210208195552-ff826a37aa15
+	github.com/alicebob/gopher-json v0.0.0-20200520072559-a9ecdc9d1d3a // indirect
```
bwplotka, line 10: Were you updating some deps too?

go.mod
```diff
@@ -74,6 +78,7 @@ replace (
 	// Using a 3rd-party branch for custom dialer - see https://github.com/bradfitz/gomemcache/pull/86.
 	// Required by Cortex https://github.com/cortexproject/cortex/pull/3051.
 	github.com/bradfitz/gomemcache => github.com/themihai/gomemcache v0.0.0-20180902122335-24332e2d58ab
+	github.com/cortexproject/cortex => github.com/pracucci/cortex v0.0.0-20210222171122-5c74b43721b8
```
bwplotka, line 81: Is this needed? 🤗

bwplotka (approved):
Super nice, Mostly LGTM, some comments/questions though.

cmd/thanos/query_frontend.go
yeya24: Yes, it is not used so it is safe to remove it. Thanks @pracucci !

pkg/objstore/objstore_test.go
pracucci: Because I've changed the `AcceptanceTest` in `pkg/objstore/testing.go` adding more `Upload()` and `Iter()` calls.  This test just calls the `AcceptanceTest()` (see line above).

go.mod
pracucci: I didn't. These has been picked up from the Cortex upgrade. If I remove the cortex `replace` directive and re-run `go mod vendor && go mod tidy` the `go.mod` rollbacks to the `master` version.

jtlisi (approved):
LGTM

This should allow us to efficiently transition the Cortex Alertmanager/Ruler clients to use the `objstore` implementation.

pstibrany (approved):
Nice work, lgtm.

bwplotka:
The acceptance test @pracucci is actually failing after this PR. See https://github.com/thanos-io/thanos/pull/3841 https://app.circleci.com/pipelines/github/thanos-io/thanos/5124/workflows/e74a9507-75f8-4079-a9a3-bbe730866e3b/jobs/13814

I believe it's not has been run for your PR for some reasons. It might be because is it configured to only run for `write-access` user to avoid spam/malicious use of buckets. You have write access to repo, but maybe you need to log in to Circle CI and "follow" project?  Do you follow it here: https://app.circleci.com/projects/project-dashboard/github/thanos-io/ ?

Something is not working for GCS, so investigating.

go.mod
bwplotka: 👍🏽

pkg/objstore/gcs/gcs.go
```diff
@@ -86,15 +86,22 @@ func (b *Bucket) Name() string {
...
+func (b *Bucket) Iter(ctx context.Context, dir string, f func(string) error, options ...objstore.IterOption) error {
 	// Ensure the object name actually ends with a dir suffix. Otherwise we'll just iterate the
 	// object itself as one prefix item.
 	if dir != "" {
 		dir = strings.TrimSuffix(dir, DirDelim) + DirDelim
 	}
+
+	// If recursive iteration is enabled we should pass an empty delimiter.
+	delimiter := DirDelim
+	if !objstore.ApplyIterOptions(options...).Recursive {
```
bwplotka, line 98: tiny bug, fixed here: https://github.com/thanos-io/thanos/pull/3841

Let's enable acceptance tests for you @pracucci 🤗

PR #3824, opened 2021-02-23
promclient: fix error's message

GiedriusS:
Use the provided `method` in the error messages. I got scared reading
the logs that the requests are still being sent using GET instead of
POST which I had specified.

## Changes

Fixed error messages to use the actual passed HTTP method.

## Verification

Still builds.


kakkoyun (approved):
🥇 LGTM 🚀

PR #3833, opened 2021-02-24
Replaces errutil with tools/pkg/merrors; Fixes very ugly misuse of multi-error

bwplotka:
Previous multi-error implementation could cause very ugly bug of returnig empty multi-error
that should be treated as success not error by API, but if .Err() is not invoked it will be
used as non nil error.

There were 9 places where we had this bug in handler due to MultiError lib allowing to do so.


pkg/receive/handler_test.go
```diff
@@ -54,155 +52,157 @@ func TestDetermineWriteErrorCause(t *testing.T) {
...
 					status.Error(codes.AlreadyExists, "conflict"),
-				}),
+				).Err(),
 				errors.New("foo"),
 				errors.New("bar"),
-			}), "baz"),
+			).Err(), "baz"),
 			threshold: 1,
-			exp:       errors.New("baz: 3 errors: 3 errors: qux; rpc error: code = AlreadyExists desc = conflict; rpc error: code = AlreadyExists desc = conflict; foo; bar"),
+			exp:       errConflict,
```
bwplotka, line 163: Can you double check @kakkoyun ? I think this makes sense more actually 🤔

pkg/receive/handler.go
```diff
@@ -563,7 +563,7 @@ func (h *Handler) fanoutForward(pctx context.Context, tenant string, replicas ma
 			return fctx.Err()
 		case err, more := <-ec:
 			if !more {
-				return errs
+				return errs.Err()
```
bwplotka, line 566: This is the thing that was causing 500 that should be 200

cc @squat @kakkoyun

pkg/receive/handler.go
```diff
@@ -333,7 +333,7 @@ func (h *Handler) receiveHTTP(w http.ResponseWriter, r *http.Request) {
 	case errBadReplica:
 		http.Error(w, err.Error(), http.StatusBadRequest)
 	default:
-		level.Error(h.logger).Log("err", err, "msg", "internal server error")
+		level.Error(h.logger).Log("err", err, "msg", "internal server error while handling receive request")
```
kakkoyun, line 336: 👍

Makefile
```diff
@@ -299,6 +299,7 @@ go-lint: check-git deps $(GOLANGCI_LINT) $(FAILLINT)
 	@# TODO(bwplotka): Add, Printf, DefaultRegisterer, NewGaugeFunc and MustRegister once exception are accepted. Add fmt.{Errorf}=github.com/pkg/errors.{Errorf} once https://github.com/fatih/faillint/issues/10 is addressed.
 	@$(FAILLINT) -paths "errors=github.com/pkg/errors,\
 github.com/prometheus/tsdb=github.com/prometheus/prometheus/tsdb,\
+github.com/prometheus/prometheus/tsdb/errors=github.com/efficientgo/tools/core/pkg/merrors,\
```
kakkoyun, line 302: 👍

pkg/receive/handler.go
kakkoyun: I thought tests were comprehensive enough to catch this :/

pkg/receive/handler_test.go
kakkoyun: AFAIR from previous comments, we didn't want to handle deeply nested errors like this. This was how I implemented while refactoring this. but @squat pointed me out that this is a behaviour change. I'm ok to have this change as long as we are aware of possible implecations.

squat: This is pretty subtle and can easily have big implications on replication. We need to spend a bit of time here analyzing the changes. It would be a lot easier to merge this if the commit maintained the same behavior and only changed the syntax to use the new errors package. My general approach is to keep PRs scoped to one purpose and avoid making other/unrelated changes wherever possible

bwplotka: Yea, sure just was curious if that is not making sense more

bwplotka: But what I gather... we don't know if that makes sense and we don't want to think about it now?

bwplotka: Separare fix in this PR: https://github.com/thanos-io/thanos/pull/3833

bwplotka:
This is ready to merge, but I think I will look `merrors` first and give ability to choose between `Flat` vs `Hierrchical` grouping.

kakkoyun:
Is there any chance to push this to the finish line @bwplotka?

PR #3836, opened 2021-02-25
Reinforcing errutil and fixed ugly bug which caused empty []error to be treated as error

bwplotka:
Previous multi-error implementation could cause very ugly bug of returnig empty multi-error
that should be treated as success not error by API, but if .Err() is not invoked it will be
used as non nil error.

Once we merge this, we can do cleaner solution that slighly change nesting behaviour: https://github.com/thanos-io/thanos/pull/3833

There were 9 places where we had this bug in handler due to MultiError lib allowing to do so.


kakkoyun (approved):
🥇 LGTM

PR #3843, opened 2021-02-26
downsample: ensure consistent order

GiedriusS:
Ensure that we have a consistent order of blocks that we are going to
downsample. `range` over maps doesn't enforce any particular order on
purpose.

This is needed for https://github.com/thanos-io/thanos/pull/3031/files.
ATM in that PR before downsampling we delete all directories which do
not match blocks ULIDs in the remote object storage. Ideally, we should
only keep around the files of a block which we are about to downsample.

It is impossible to do that properly ATM if during another iteration
we'd start from a different block. Thus, let's have a consistent order.

## Changes

Added sorting for block ULIDs before attempting to possibly downsample them.

## Verification

Tests still pass.

### DIFF
--- a/cmd/thanos/downsample.go
+++ b/cmd/thanos/downsample.go
@@ -7,6 +7,7 @@ import (
 	"context"
 	"os"
 	"path/filepath"
+	"sort"
 	"time"
 
 	"github.com/go-kit/kit/log"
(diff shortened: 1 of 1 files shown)


yeya24 (approved):
LGTM

GiedriusS:
Thanks for the quick review!

Issue #3839, opened 2021-02-26
Stop uploading debug meta files to obj store


yeya24:
### Is your proposal related to a problem?

(Write your answer here.)

![Screenshot from 2021-02-25 22-31-03](https://user-images.githubusercontent.com/25150124/109251357-509c2e80-77b9-11eb-95c9-86ce906ff219.png)

As agreed in one previous community meeting, we decided to remove the step of uploading debug meta files because seems it is hardly used.

@thanos-io/thanos-maintainers 

### Describe the solution you'd like

Remove step https://github.com/thanos-io/thanos/blob/12c826a86e0c1de65ea770caf7cecd61acb22f24/pkg/block/block.go#L106-L108 and related tests code.

hitanshu-mehta:
Can i work on this issue?

bwplotka:
Yes please @hitanshu-mehta !

GiedriusS:
Maybe we should also remove whatever already meta files exist in the bucket already? https://github.com/thanos-io/thanos/issues/3397

pstibrany:
I have found these files useful when tracking down history of some blocks during our data conversion. I would weakly prefer to keep them, as the overhead seems tiny.

kakkoyun:
Still valid.

pstibrany:
I think I was the only one for keeping the debug meta files. As I haven't had any use for them in last ~1year and instead suffered from trying to delete lots of them from object store, I am fine with getting rid of them.

(I hope my previous weak preference to keep them isn't what stalled the work here.)

yeya24:
Still valid

r0mdau:
I tried something, kudos to your work @hitanshu-mehta . I was trying to build thanos using this issue as example. I thought maybe it can helps.

PR #3852, opened 2021-03-01
Sidecar/Ruler/Receiver: Add flag to disable upload of debug meta files

hitanshu-mehta:

Fixes: #3839 

## Changes
- Introduced new flag `--shipper.upload-debug-meta-files` in sidecar, ruler and receiver.
- Added test to verify working of flag.

## Verification
- Ran unit and e2e tests locally.


pkg/block/block.go
```diff
@@ -103,10 +102,6 @@ func Upload(ctx context.Context, logger log.Logger, bkt objstore.Bucket, bdir st
 		return errors.Wrap(err, "encode meta file")
 	}
 
-	if err := bkt.Upload(ctx, path.Join(DebugMetas, fmt.Sprintf("%s.json", id)), strings.NewReader(metaEncoded.String())); err != nil {
```
bwplotka, line 106: Hm, decision was to remove this.. optionally? 

TBH I never used, but wonder if this was not used by someone at the end.

bwplotka: One way would be to remove .. let's say by default but enable for opt-in? 🤔

bwplotka (approved):
Nice. Let's decide if we want to remove completely or enable optionally.

pkg/block/block.go
pracucci: I'm fine to fully remove it. I know @pstibrany used it few times, so let me ask it too.

pstibrany: I can only repeat what I wrote here: https://github.com/thanos-io/thanos/issues/3839#issuecomment-786674853

I don’t have strong feelings about it, but found it useful in the past.

bwplotka: Me too, that's why I created this in first place

bwplotka:
Can we move this under some flag for now? 🤗

hitanshu-mehta:
Sure. Flag named `--shipper.upload-debug-meta-files` in sidecar seems correct?

yeya24:
> Sure. Flag named `--shipper.upload-debug-meta-files` in sidecar seems correct?

Receiver and Ruler support uploading blocks as well.

hitanshu-mehta:
Thank you. I will add flags in them too.

cmd/thanos/downsample.go
```diff
@@ -315,7 +315,7 @@ func processDownsampling(ctx context.Context, logger log.Logger, bkt objstore.Bu
 
 	begin = time.Now()
 
-	err = block.Upload(ctx, logger, bkt, resdir, hashFunc)
+	err = block.Upload(ctx, logger, bkt, resdir, hashFunc, true)
 	if err != nil {
```
bwplotka, line 319:
```suggestion
	if err := block.Upload(ctx, logger, bkt, resdir, hashFunc, true); err !=nil
```

pkg/block/block.go
```diff
@@ -98,7 +98,7 @@ func Download(ctx context.Context, logger log.Logger, bucket objstore.Bucket, id
 // It also verifies basic features of Thanos block.
 // TODO(bplotka): Ensure bucket operations have reasonable backoff retries.
 // NOTE: Upload updates `meta.Thanos.File` section.
-func Upload(ctx context.Context, logger log.Logger, bkt objstore.Bucket, bdir string, hf metadata.HashFunc) error {
+func Upload(ctx context.Context, logger log.Logger, bkt objstore.Bucket, bdir string, hf metadata.HashFunc, uploadDebugMetaFiles bool) error {
```
bwplotka, line 101: I think we need uploader struct at this point. Will add this in another PR - too many args TBH (:

bwplotka (commented):
Nice, some suggestion and idea for next step (:

pkg/block/block.go
hitanshu-mehta: Sure 😊

cmd/thanos/tools_bucket.go
```diff
@@ -907,7 +907,7 @@ func registerBucketRewrite(app extkingpin.AppClause, objStoreConfig *extflag.Pat
 				}
 
 				level.Info(logger).Log("msg", "uploading new block", "source", id, "new", newID)
-				if err := block.Upload(ctx, logger, bkt, filepath.Join(*tmpDir, newID.String()), metadata.HashFunc(*hashFunc)); err != nil {
+				if err := block.Upload(ctx, logger, bkt, filepath.Join(*tmpDir, newID.String()), metadata.HashFunc(*hashFunc), true); err != nil {
```
prmsrswt, line 910: The same flag makes sense here too.

pkg/block/block_test.go
```diff
@@ -205,11 +205,29 @@ func TestUpload(t *testing.T) {
...
 		}, 100, 0, 1000, nil, 124, metadata.NoneFunc)
 		testutil.Ok(t, err)
-		err = Upload(ctx, log.NewNopLogger(), bkt, path.Join(tmpDir, b2.String()), metadata.NoneFunc)
+		err = Upload(ctx, log.NewNopLogger(), bkt, path.Join(tmpDir, b2.String()), metadata.NoneFunc, true)
 		testutil.NotOk(t, err)
 		testutil.Equals(t, "empty external labels are not allowed for Thanos block.", err.Error())
 		testutil.Equals(t, 4, len(bkt.Objects()))
 	}
+	{
+		// Upload block without debug meta files.
```
prmsrswt, line 214: Do we have a test that checks the `Upload block with debug meta files.` condition as well?

pkg/verifier/index_issue.go
```diff
@@ -111,7 +111,7 @@ func (IndexKnownIssues) VerifyRepair(ctx Context, idMatcher func(ulid.ULID) bool
 		}
 
 		level.Info(ctx.Logger).Log("msg", "uploading repaired block", "newID", resid)
-		if err = block.Upload(ctx, ctx.Logger, ctx.Bkt, filepath.Join(tmpdir, resid.String()), metadata.NoneFunc); err != nil {
+		if err = block.Upload(ctx, ctx.Logger, ctx.Bkt, filepath.Join(tmpdir, resid.String()), metadata.NoneFunc, true); err != nil {
```
prmsrswt, line 114: This needs to passed through a flag as well I suppose.

pkg/verifier/safe_delete.go
```diff
@@ -136,7 +136,7 @@ func backupDownloaded(ctx context.Context, logger log.Logger, bdir string, backu
 
 	// Upload the on disk TSDB block.
 	level.Info(logger).Log("msg", "Uploading block to backup bucket", "id", id.String())
-	if err := block.Upload(ctx, logger, backupBkt, bdir, metadata.NoneFunc); err != nil {
+	if err := block.Upload(ctx, logger, backupBkt, bdir, metadata.NoneFunc, true); err != nil {
```
prmsrswt, line 139: Ditto, we need a flag here too.

cmd/thanos/config.go
```diff
@@ -133,6 +134,8 @@ func (sc *shipperConfig) registerFlag(cmd extkingpin.FlagClause) *shipperConfig
 	cmd.Flag("shipper.upload-compacted",
 		"If true shipper will try to upload compacted blocks as well. Useful for migration purposes. Works only if compaction is disabled on Prometheus. Do it once and then disable the flag when done.").
 		Default("false").BoolVar(&sc.uploadCompacted)
+	cmd.Flag("shipper.upload-debug-meta-files", "If true shipper will upload debug meta files which can be useful for debugging.").
+		Default("false").BoolVar(&sc.uploadDebubgMetaFiles)
```
prmsrswt, line 138: We are using the value `true` by default in a lot of places (tests etc) so why the default is "false"? IMO it makes more sense to have the default as "true" as it will maintain backward compatibility as well.

pkg/block/block_test.go
hitanshu-mehta: Yes. There is one test case above this. https://github.com/thanos-io/thanos/blob/816628343d4de1a9b8f829178906b3387addd88c/pkg/block/block_test.go#L139

prmsrswt: That test is not verifying the presence of debug meta files, it should. Also, this test that you added, you should check that the debug meta files are NOT uploaded, you are not checking that.

hitanshu-mehta: If we upload with debug meta files then 4 files are uploaded per block. So the test I wrote there are only 7 files in the obj. store (4 files from the test with debug meta files + 3 files from the test without debug metafiles).

prmsrswt: Gotcha :+1:  But I think, like we are verifying ChunkFiles, Meta File etc, we can explicitly verify debug metas too.

hitanshu-mehta: Understood :) I will add that verification. 
I have 2 questions.
1. Should I add that verification in all test cases?
2. Is it good practice that one test case depends on another test case? Because the test case I wrote depends on the previous one. I decided to write this test case this way because otherwise, I have to make a new instance obj. store. WDYT? 🤗

prmsrswt: 1. No need to add it in other tests, it'd be redundant.
2. In this case, I think it's not a good practice. What if someone adds a new test case in between?

hitanshu-mehta: Got it 👍🏻 . I will change the test case accordingly. Thank you :)

hitanshu-mehta:
PR is ready for review 🤗

bwplotka (dismissed):
Looks good! Just rebase. I hope we can improve the `Upload` someday with optional attributes./parameters.

hitanshu-mehta:
Thank you for the review :) I will create a follow-up PR for uploader struct.

bwplotka:
We can see that we touch so many things...

Given you need to rebase any way, maybe it's worth creating uploader struct... in this PR? (:

hitanshu-mehta:
Yes, I agree. To rebase, I have to resolve many merge conflicts.

pstibrany:
If keeping support for use in Cortex was the only reason to keep the flag, I would suggest to simplify the PR, drop the flag and remove support for those debug meta files completely.

We can already add support for debug/meta.json file on our own easily, if we really want to. (eg. by having a special bucket that uploads meta.json to two places).

kakkoyun:
Hey @hitanshu-mehta. We're going to cut a release candidate soon, is there any chance for you to push this to the finish line?

hitanshu-mehta:
Yes, I will be able to do it. I was waiting for the decision on whether we going to completely remove support for debug meta files or not 😅

hitanshu-mehta:
I guess, failed e2e test is not related.

kakkoyun:
> Looks like we decided to have this opt-in by default ([#3852 (comment)](https://github.com/thanos-io/thanos/pull/3852#discussion_r584614558)). Then maybe let's add an entry in the CHANGELOG.

I think the only decision to make is to enable this by default or not? @bwplotka @onprem @pracucci @pstibrany

yeya24:
Let's revisit this. My opinion now is to remove debug metas totally and we don't need a flag.

yeya24:
Let's get back to this finally. @hitanshu-mehta Are you able to continue this? Thank you!

hitanshu-mehta:
@yeya24 Yes. So the final decision is to remove debug meta files completely without any flag. right?

yeya24:
> @yeya24 Yes. So the final decision is to remove debug meta files completely without any flag. right?

Yes! Thanks for the patience.

hitanshu-mehta:
I apologize for the delay from my side. @r0mdau has opened a PR #5110 fixing this issue. Hence, we can close this PR.

PR #3853, opened 2021-03-01
Bump github.com/gogo/protobuf to v1.3.2 in .bingo

simonpasquier:

## Changes

In practice, the recent bump of github.com/gogo/protobuf (introduced by #3804) changes the generated code to fix [CVE-20213121](https://nvd.nist.gov/vuln/detail/CVE-2021-3121).

## Verification

No functional change.


.github/workflows/go.yaml
```diff
@@ -54,6 +54,33 @@ jobs:
...
+          path: ~/go/pkg/mod
+          key: ${{ runner.os }}-go-${{ hashFiles('**/go.sum') }}
+
+      - name: Generate protobuf code
+        run: make proto
+
+      - name: Format
+        run: make format
+
+      - name: Format
```
prmsrswt, line 82:
```suggestion
      - name: Check proto
```

Nit: Something like this?

prmsrswt (approved):
LGTM! I just have a small nit but not a blocker.

bwplotka (approved):
Thanks!

.github/workflows/go.yaml
```diff
@@ -54,6 +54,33 @@ jobs:
...
+        uses: actions/setup-go@v2
+        with:
+          go-version: 1.15.x
+
+      - uses: actions/cache@v1
+        with:
+          path: ~/go/pkg/mod
+          key: ${{ runner.os }}-go-${{ hashFiles('**/go.sum') }}
+
+      - name: Generate protobuf code
```
bwplotka, line 76: Hm, why not just adding proto to `lint` Makefile and... done? 🤔

squat (approved):
v nice :) thanks!

.github/workflows/go.yaml
simonpasquier: whatever works for you :)

simonpasquier: wait it's already done!

.github/workflows/go.yaml
```diff
@@ -3,7 +3,7 @@ name: go
 on:
   push:
     branches:
-      - main
+      - master
```
kakkoyun, line 6: Haven't we just renamed `master` to `main`? I guess this change shouldn't be here.

kakkoyun (changes requested):
Looks great! Only one problem I have seen.

.github/workflows/go.yaml
simonpasquier: good catch! fixed.

kakkoyun (approved):

codefromthecrypt:
FYI: I think you accidentally missed .bingo/Variables.mk and variables.env

simonpasquier:
@codefromthecrypt correct although it doesn't change anything in practice: the code has been generated with gogo/protobuf v1.3.2. I won't have time to submit a follow-up PR so go ahead if you wish :)

codefromthecrypt:
@simonpasquier sure why not. #3902

PR #3854, opened 2021-03-01
mixin: Remove assumed metrics

kakkoyun:
This PR fixes minor inconveniences with mixin. Remove assumed metrics. Use `thanos_info` instead of `kube_pod_info` for dashboard selectors.

## Changes

* Unify file-naming conventions
* Only assume the existence of metrics added by Thanos itself. Remove assumed metrics. Use `thanos_info` instead of `kube_pod_info` for dashboard selectors.
* Remove pod selectors

## Verification

* `make examples`
* Manual import and poke around


examples/alerts/alerts.md
```diff
@@ -117,8 +115,7 @@ rules:
     severity: critical
 - alert: ThanosRuleHighRuleEvaluationWarnings
   annotations:
-    description: Thanos Rule {{$labels.job}} {{$labels.pod}} has high number of evaluation
```
bwplotka, line 120: What's the reason of `labels.pod` removal? It's because it's not always `pod`?

examples/dashboards/compact.json
```diff
@@ -1407,47 +1407,47 @@
...
                      "legendFormat": "inuse stack {{pod}}",
                      "legendLink": null,
                      "step": 10
                   },
                   {
-                     "expr": "go_memstats_heap_inuse_bytes{namespace=\"$namespace\",job=~\"$job\",kubernetes_pod_name=~\"$pod\"}",
+                     "expr": "go_memstats_heap_inuse_bytes{namespace=\"$namespace\",job=~\"$job\"}",
                      "format": "time_series",
                      "intervalFactor": 2,
                      "legendFormat": "inuse heap {{pod}}",
```
bwplotka, line 1453: Interesting that pod is still used here

examples/dashboards/compact.json
```diff
@@ -1407,47 +1407,47 @@
...
+                     "expr": "go_memstats_stack_inuse_bytes{namespace=\"$namespace\",job=~\"$job\"}",
                      "format": "time_series",
                      "intervalFactor": 2,
                      "legendFormat": "inuse stack {{pod}}",
                      "legendLink": null,
                      "step": 10
                   },
                   {
-                     "expr": "go_memstats_heap_inuse_bytes{namespace=\"$namespace\",job=~\"$job\",kubernetes_pod_name=~\"$pod\"}",
+                     "expr": "go_memstats_heap_inuse_bytes{namespace=\"$namespace\",job=~\"$job\"}",
```
bwplotka, line 1450: So no pod filtering? (maybe makes sense)

bwplotka (commented):
> Remove pod selectors

Why?

brancz:
It's probably useful to continue to use the `instance` label instead of `pod`. You can rely on `instance` being there and it's up to the user to relabel something useful to their environment in there.

examples/dashboards/compact.json
kakkoyun: Exactly, it wasn't used much as far as I can see.

examples/dashboards/compact.json
kakkoyun: It's just for presentation, we don't need to filter but it's better to have the ability to distinguish them

examples/alerts/alerts.md
kakkoyun: As suggested I'll use `instance`.

kakkoyun: I removed them because I thought do we actually need to know the pod/instance in description

bwplotka (commented):
Stil I might be stupid, but why we get rid of pod. Having some random IP address is not helpful on dashboards 🤔

brancz:
The best practice is to relabel something sensible into your `instance` label though, so I think it's a reasonable default to use in mixins. FWIW kube-thanos does abide by that and relabels `namespace/pod` into the `instance` label.

kakkoyun:
I just want to make our mixin kubernetes agnostic (this has been asked a couple of times) and I think these changes will do the work.

metalmatze:
I still see a bunch of occasions of the `namespace` label too. So if we want to get Kubernetes agnostic then that needs to go too.

kakkoyun:
IMHO `namespace` is a more generic term. It doesn't necessarily indicate it's a Kubernetes namespace.

kakkoyun:
Discussed this offline and created an issue for it https://github.com/thanos-io/thanos/issues/3867

metalmatze (approved):
Overall this looks really good! :+1: 
Happy to continue with `namespace` in a follow up,

PR #3855, opened 2021-03-01
mixin: Upgrade jsonnet tooling

kakkoyun:


.bingo/variables.env
```diff
@@ -1,4 +1,4 @@
-# Auto generated binary variables helper managed by https://github.com/bwplotka/bingo v0.3.0. DO NOT EDIT.
+# Auto generated binary variables helper managed by https://github.com/bwplotka/bingo v0.2.3. DO NOT EDIT.
```
bwplotka, line 1: Can you use bingo v0.3.0 instead?

bwplotka (approved):
Thanks! 🤗 Can we use latest bingo?

PR #3856, opened 2021-03-01
mixin: Introduce flexible multi-cluster/namespace mode for alerts and dashboards

kakkoyun:

## Changes

* Add namespace to the alerting rules and annotations

## Verification

* `make examples`
* `make example-rules-lint`

Merge after: https://github.com/thanos-io/thanos/pull/3854
Fixes: #3856


bwplotka (commented):
I would love this PR, but unfortunately, it also has those `pod` selector changes, which I don't understand 🤔

kakkoyun:
Friendly ping @metalmatze @bwplotka

brancz:
Didn't we just work on getting rid of kubernetes specific things?

kakkoyun:
Yes, definitely @brancz. I'll be introducing further changes to provide a more abstract solution for this https://github.com/thanos-io/thanos/issues/3867. It'll require more consideration. This would solve our problems in the short-term.

kakkoyun:
Ok, I'm convinced that we need to generalize this. Marking it WIP and jumping to fix these.

kakkoyun:
Friendly ping @bwplotka @metalmatze @brancz. Could you please take one last look at it?

examples/dashboards/bucket_replicate.json
```diff
@@ -47,7 +47,7 @@
                "steppedLine": false,
                "targets": [
                   {
-                     "expr": "sum(rate(thanos_replicate_replication_runs_total{result=\"error\", namespace=\"$namespace\",job=~\"thanos-bucket-replicate.*\"}[$interval])) / sum(rate(thanos_replicate_replication_runs_total{namespace=\"$namespace\",job=~\"thanos-bucket-replicate.*\"}[$interval]))",
+                     "expr": "sum(rate(thanos_replicate_replication_runs_total{result=\"error\", job=\"$job\"}[$interval])) / sum(rate(thanos_replicate_replication_runs_total{job=\"$job\"}[$interval]))",
```
bwplotka, line 50: should we do `by (job, namespace)` here?

examples/dashboards/bucket_replicate.json
```diff
@@ -124,7 +124,7 @@
                "steppedLine": false,
                "targets": [
                   {
-                     "expr": "sum(rate(thanos_replicate_replication_runs_total{result=\"error\", namespace=\"$namespace\",job=~\"thanos-bucket-replicate.*\"}[$interval])) by (result)",
+                     "expr": "sum by (result) (rate(thanos_replicate_replication_runs_total{result=\"error\", job=\"$job\"}[$interval]))",
```
bwplotka, line 127: ditto

examples/dashboards/overview.json
```diff
@@ -149,7 +149,7 @@
                "steppedLine": false,
                "targets": [
                   {
-                     "expr": "sum(rate(http_requests_total{namespace=\"$namespace\",job=~\"thanos-query.*\",handler=\"query\",code=~\"5..\"}[$interval])) / sum(rate(http_requests_total{namespace=\"$namespace\",job=~\"thanos-query.*\",handler=\"query\"}[$interval]))",
+                     "expr": "sum(rate(http_requests_total{job=\"$job\", handler=\"query\",code=~\"5..\"}[$interval])) / sum(rate(http_requests_total{job=\"$job\", handler=\"query\"}[$interval]))",
```
bwplotka, line 152: We have to do by (job, namespace ) otherwise we sum things by all the things, no?

bwplotka (changes requested):
I wonder if we should not have `by (job, namespace)` in most of those places to ensure we are not reducing all to one value. If we have multiple jobs in different namespaces we would count those as one  series with this PR, no? 🤔

kakkoyun:
@bwplotka As stated in the issue(https://github.com/thanos-io/thanos/issues/3867) we try to decouple from any platforms by removing specific notions from the mixin. Thus I removed `namespace` from all the examples. However, we provide a flexible way to add any hierarchy to the alerts and dashboards through configuration. So it's okay to remove `namespace`. I have already added the necessary changelog entry.

examples/dashboards/bucket_replicate.json
bwplotka: Let me comment jsonnet code instead, but this example can mislead people - some of them maybe even copies those example 🤔

mixin/dashboards/compact.libsonnet
```diff
@@ -18,7 +18,7 @@ local g = import '../lib/thanos-grafana-builder/builder.libsonnet';
             'Shows rate of execution for compactions against blocks that are stored in the bucket by compaction group.'
           ) +
           g.queryPanel(
-            'sum(rate(thanos_compact_group_compactions_total{namespace="$namespace",job=~"$job"}[$interval])) by (job, group)',
+            'sum by (job, group) (rate(thanos_compact_group_compactions_total{%s}[$interval]))' % selector,
```
bwplotka, line 21: I think we might want to put `(%(aggregators)s)`. The reason is that by removing namespace we allow multiple namespaces (and clusters - which was before) to be part of this aggregation. So suddenly we are returning sum-med number of group compactions, not one. And it's hidden to the dashboard user, I know it's a tedious work, but it sounds crucial to me 🤔

bwplotka (changes requested):
I understand that, however, the problem is that we are summing namespaces and potentially cluster (whatever users have) into one value.

kakkoyun:
Thanks for pointing out. I'll make sure all the queries are modified with this way.

kakkoyun:
@bwplotka It's ready for another round. I've added missing aggregators.

mixin/README.md
```diff
@@ -59,48 +59,54 @@ This project is intended to be used as a library. You can extend and customize d
 [embedmd]:# (config.libsonnet)
 ```libsonnet
 {
+  local thanos = self,
+  // Hierarcies is a way to help mixin users to add high level hierarcies to their alerts and dashboards.
```
bwplotka, line 63: Thanks for mini docs <3

bwplotka, line 63: WDYT `groupby` or `groups` or `tagets`?

```suggestion
  // Groups <slice?> is a way to help mixin users to add high level target grouping to their alerts and dashboards.
```

mixin/README.md
```diff
@@ -59,48 +59,54 @@ This project is intended to be used as a library. You can extend and customize d
...
   },
-  overview+:: {
-    title: '%(prefix)sOverview' % $.dashboard.prefix,
-  },
   dashboard+:: {
     prefix: 'Thanos / ',
     tags: ['thanos-mixin'],
-    namespaceQuery: 'thanos_status',
+    selector: ['%s="$%s"' % [level, level] for level in std.objectFields(thanos.hierarcies)],
+    aggregator: ['%s' % level for level in std.objectFields(thanos.hierarcies)],
```
bwplotka, line 103: WDYT?

```suggestion
    dimensions: ['%s' % level for level in std.objectFields(thanos.hierarcies)],
```
https://prometheus.io/docs/prometheus/latest/querying/operators/#aggregation-operators

mixin/README.md
bwplotka: Actually, I don't understand why hierarchies are helpful here and why we need them? Can we just show in an example `dimensions/aggregators: job, namespace, cluster`?

mixin/README.md
bwplotka: Reason is.. aggregator is `sum, max, min` etc 🤗

bwplotka (commented):
Looks perfect now, very flexible and consistent. But I would suggest bit different naming 🤗

mixin/dashboards/sidecar.libsonnet
```diff
@@ -1,51 +1,58 @@
...
+          g.latencyPanel('grpc_server_handling_seconds', grpcServerSelector, thanos.sidecar.dashboard.aggregator)
         )
       )
       .addRow(
         g.row('Last Updated')
         .addPanel(
           g.panel('Successful Upload', 'Shows the relative time of last successful upload to the object-store bucket.') +
           g.tablePanel(
-            ['time() - max(thanos_objstore_bucket_last_successful_upload_time{namespace="$namespace",job=~"$job"}) by (job, bucket)'],
+            ['time() - max by (%s) (thanos_objstore_bucket_last_successful_upload_time{%s})' % [utils.joinLabels([thanos.sidecar.dashboard.aggregator, 'bucket']), thanos.sidecar.dashboard.selector]],
```
bwplotka, line 55: Quite understandable (`bucket` label addition)!

bwplotka (approved):
Epic, looks clean and amazing to me! 💪🏽 Let's go! 🚀

Issue #985, opened 2019-03-27
Feature Request: Detect credentials changes without application restarts


Capitrium:
Currently, Thanos reads the bucket configuration file only at startup. This works fine for most cases, but becomes problematic when the credentials are generated and rotated by an external service. In this case, the credentials that Thanos started with will become invalid and interactions with the configured object storage will fail until Thanos is restarted, which is not ideal.

Some potential solutions to this problem include implementing a hot reloading feature, or adding a sidecar container that triggers a restart of Thanos when the config file is updated.

bwplotka:
Thanks!

TBH I would rather go for the latter option. Something that might be helpful is this: https://github.com/jetstack/cert-manager/issues/1168 as it is exactly the same problem.

Particularly https://github.com/pusher/wave 

Essentially on k8s you would have such cases all the time, so this might be the way to go for now. Not sure if it's worth to add complexity to the code as restart might be just enough. We might want to enumerate all components and see if restart is safe. E.g all components, especially Prometheus + sidecar required HA for safe rolling restart etc.

Capitrium:
While Wave seems interesting, to be honest I'm not too enthusiastic about depending on yet another service just to handle the rotation of the config file. It means I need to worry about the integrations with this extra service (does it support StatefulSets and DaemonSets? Or just Deployments, and the others are left out in the cold?), whether the service is highly available, any physical resources required to run this additional service, etc.

It also makes it more difficult to integrate with other services. As an example, the prometheus-operator completely manages the StatefulSets it creates. If it didn't provide a way to set custom annotations on those StatefulSets, how would I opt those pods into using Wave? (Thankfully the prometheus CRD has a `podMetadata` field for this purpose, but hopefully you get my point. :smile:)

K8s already provides updates to ConfigMap and Secret volumes on the kubelet's resync interval, so IMO any applications that expect to run on K8s should also expect any config file mounted from a ConfigMap or Secret to be updated during execution and should handle that accordingly. However, your concerns about maintaining availability during restarts triggered by a config rotation are definitely valid; applications that need to coordinate restarts between highly-available components would need to do so through a ConfigMap or something similar, which definitely adds complexity on the Thanos side.

[...]

PR #3857, opened 2021-03-01
sidecar: Mark sidecar as unhealthy if bucket upload fails

LittleChimera:

## Changes

Health probe flips to unhealthy if sidecar upload to a bucket fails. This would enable restarting the sidecar with liveliness probe in a containerised environment and reload config on restart.

## Verification

TODO

Fixes: https://github.com/thanos-io/thanos/issues/985

### DIFF
--- a/cmd/thanos/sidecar.go
+++ b/cmd/thanos/sidecar.go
@@ -275,6 +275,9 @@ func runSidecar(
 			return runutil.Repeat(30*time.Second, ctx.Done(), func() error {
 				if uploaded, err := s.Sync(ctx); err != nil {
 					level.Warn(logger).Log("err", err, "uploaded", uploaded)
+					statusProber.NotHealthy(err)
+				} else {
+					statusProber.Healthy()
 				}
 
 				minTime, _, err := s.Timestamps()


kakkoyun (changes requested):
Hey @LittleChimera, thanks for the PR.
However, I'm not sure about this change, this will mark sidecar unhealthy as soon as it encounters an error while uploading. The sidecar also responses to the queries. Even though uploading blocks when enabled is pat of the SLA, this shouldn't just mark the process as unhealthy. We should be more clear about this.

LittleChimera:
thanks for the feedback @kakkoyun!
yeah, you have a good point - it's important that sidecar can still be used as a querier even if uploading fails for some time. However, I think this change could still go hand in hand with that. If you had a default Kubernetes probes with readiness pointing to `/-/ready` and liveness pointing to `/-/healthy`, the pod would still be able to receive traffic for some time even if the liveness probe is failing. With this change, only liveliness probe is flipped. I think a good approach in this case would be to set combination `failureThreshold`  and `periodSeconds` on liveness probe to something reasonable to operators so that sidecar doesn't restart too often in case of more permanent error. Consequently, operators can still ensure required availability of the sidecar to respond to queries.

LittleChimera:
> semantically speaking, would it be better if uploader probe was added as a separate probe to `prober.Combine(...)`?

LittleChimera:
@GiedriusS to simplify a bit what I meant - service health being degraded (i.e. marked as unhealthy) doesn't have to trigger immediate restart of the service. Depending on what kind of expectations the operators have, they might configure the service to keep running in degraded state for some time to try to let it recover on its own (no liveness probe or delayed), stop traffic (readiness probe), or restart immediately (liveness probe with no delay).

For thanos-sidecar, I think it would make most sense if liveness probe failed after upload has been failing for some longer time (e.g. 10 minutes), or if it hasn't been responding properly to queries for a shorter time (e.g. 30s). I think that timing decision for both of these should be left to the external system rather than trying to implement this logic into the sidecar itself.

kakkoyun (commented):
I still think it's not a sufficient solution to mark the sidecar as unhealthy and restart when uploads start to fail. We promote alerts for these reasons. Errors could come from different directions. This would prevent the sidecar to run under the degraded state.

We should focus one what we actually try to solve here, `Detect credentials changes without application restarts`. We want to dynamically reload the credentials for the bucket, right? Let's focus on that. If it needs a relatively bigger change, we are happy to accept design proposals 🤗 

Let's go back to design boards and try attacking this issue from another direction. For that, I'm closing this PR.

PR #3866, opened 2021-03-03
Add ReaderWithSize that allows readers to report size of object directly

pstibrany:
This PR makes following changes:
- Add `ReaderWithSize` interface that extends Reader with support to report the size of object directly
- Make `tracingReadCloser` implement `ReaderWithSize`.
- Also added `NopCloserWithSize` that can be used instead of `ioutil.NopCloser`.
- Support for `*bytes.Reader` when getting size.

Especially `ReaderWithSize` should make it simpler to avoid warnings with unknown object size during uploads.

### DIFF
--- a/pkg/objstore/objstore.go
+++ b/pkg/objstore/objstore.go
@@ -145,29 +145,27 @@ func TryToGetSize(r io.Reader) (int64, error) {
 		return int64(f.Len()), nil
 	case *strings.Reader:
 		return f.Size(), nil
-	case Sizer:
-		return f.Size()
+	case ObjectSizer:
+		return f.ObjectSize()
 	}
 	return 0, errors.Errorf("unsupported type of io.Reader: %T", r)
 }
 
-// Sizer can return size of data.
-type Sizer interface {
-	// Size returns length of the data, or error, if it is not available.
-	// It is best to call Size before calling other methods on the type, as some io.Reader implementations may only return
-	// length of unread data.
-	Size() (int64, error)
+// ObjectSizer can return size of object.
+type ObjectSizer interface {
+	// ObjectSize returns the size of the object in bytes, or error if it is not available.
+	ObjectSize() (int64, error)
 }
 
-type nopCloserWithSize struct{ io.Reader }
+type nopCloserWithObjectSize struct{ io.Reader }
 
-func (nopCloserWithSize) Close() error           { return nil }
-func (n nopCloserWithSize) Size() (int64, error) { return TryToGetSize(n.Reader) }
+func (nopCloserWithObjectSize) Close() error                 { return nil }
+func (n nopCloserWithObjectSize) ObjectSize() (int64, error) { return TryToGetSize(n.Reader) }
 
 // NopCloserWithSize returns a ReadCloser with a no-op Close method wrapping
 // the provided Reader r. Returned ReadCloser also implements Size method.
 func NopCloserWithSize(r io.Reader) io.ReadCloser {
-	return nopCloserWithSize{r}
+	return nopCloserWithObjectSize{r}
 }
 
 // UploadDir uploads all files in srcdir to the bucket with into a top-level directory
--- a/pkg/objstore/objstore.go
+++ b/pkg/objstore/objstore.go
@@ -126,6 +126,9 @@ type ObjectAttributes struct {
 }
 
 // TryToGetSize tries to get upfront size from reader.
+// Some implementations may return only size of unread data in the reader, so it's best to call this method before
+// doing any reading.
+//
 // TODO(https://github.com/thanos-io/thanos/issues/678): Remove guessing length when minio provider will support multipart upload without this.
 func TryToGetSize(r io.Reader) (int64, error) {
 	switch f := r.(type) {
@@ -136,6 +139,9 @@ func TryToGetSize(r io.Reader) (int64, error) {
 		}
 		return fileInfo.Size(), nil
 	case *bytes.Buffer:
+		return int64(f.Len()), nil
+	case *bytes.Reader:
+		// Returns length of unread data only.
 		return int64(f.Len()), nil
 	case *strings.Reader:
 		return f.Size(), nil
@@ -150,6 +156,8 @@ type ReaderWithSize interface {
 	io.Reader
 
 	// Size returns size of the object read by this reader, or error, if size is not available.
+	// It is best to call Size before reading the object, as some implementations may only return
+	// length of unread data.
 	Size() (int64, error)
 }
 
--- a/pkg/objstore/objstore.go
+++ b/pkg/objstore/objstore.go
@@ -151,12 +151,12 @@ func TryToGetSize(r io.Reader) (int64, error) {
 	return 0, errors.Errorf("unsupported type of io.Reader: %T", r)
 }
 
-// ReaderWithSize is a Reader that can also return size of the read object.
+// ReaderWithSize is a Reader that can also return size of the read data.
 type ReaderWithSize interface {
 	io.Reader
 
-	// Size returns size of the object read by this reader, or error, if size is not available.
-	// It is best to call Size before reading the object, as some implementations may only return
+	// Size returns length of the data read by this reader, or error, if it is not available.
+	// It is best to call Size before using the Reader, as some implementations may only return
 	// length of unread data.
 	Size() (int64, error)
 }
--- a/pkg/objstore/objstore.go
+++ b/pkg/objstore/objstore.go
@@ -145,18 +145,16 @@ func TryToGetSize(r io.Reader) (int64, error) {
 		return int64(f.Len()), nil
 	case *strings.Reader:
 		return f.Size(), nil
-	case ReaderWithSize:
+	case Sizer:
 		return f.Size()
 	}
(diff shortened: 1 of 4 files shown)


pstibrany:
I believe failed `TestCompactWithStoreGateway` is unrelated to my changes, but happy to fix is proven wrong.

pracucci (approved):
LGTM, thanks!

pkg/objstore/objstore.go
```diff
@@ -137,12 +140,38 @@ func TryToGetSize(r io.Reader) (int64, error) {
...
 	case *strings.Reader:
 		return f.Size(), nil
+	case ReaderWithSize:
+		return f.Size()
 	}
 	return 0, errors.Errorf("unsupported type of io.Reader: %T", r)
 }
 
+// ReaderWithSize is a Reader that can also return size of the read data.
+type ReaderWithSize interface {
```
bwplotka, line 155: I think @pstibrany this makes sense if we would replace the `io.Reader` with this. Otherwise, I wonder if it would be not easier to just have `Sizer` interface and cast to that in `TryToGetSize` (: WDYT?

bwplotka (commented):
I think we could take it.. further 🤗 WDYT?

pkg/objstore/objstore.go
bwplotka: I would be happy I think to accept `ReaderWithSize` explicitly by our Upload - It will being some  complexity to the user, but at least avoid efficiency issues.

pstibrany: > Otherwise, I wonder if it would be not easier to just have `Sizer` interface and cast to that in `TryToGetSize` (: WDYT?

I considered this approach, but it felt very generic. I wanted to express that this is really a Reader which is aware of size of (possibly unread) data. If it was defined only as Sizer, it would be weird to refer to readers’ unread data in the comment. If you however still prefer that approach, I am fine with that.

> I would be happy I think to accept `ReaderWithSize` explicitly by our Upload - It will being some complexity to the user, but at least avoid efficiency issues.

I’m not in favor of this for few reasons:
- [amount of code using Thanos objstore library](https://pkg.go.dev/github.com/thanos-io/thanos/pkg/objstore?tab=importedby), that would need to change. There is also some private code not shown on that page. I’m not sure the benefits outweigh causing so much trouble :)
- Interface is defined such that `Size` can return error. What would Upload do in such case? Fail? Or would it be implementation specific? If it doesn’t fail, why require it in the first place?
- Typical object storage doesn’t need this information in advance, so requiring it by API seems counter-intuitive to me.

pstibrany: I forgot to thank you for your review! :)

bwplotka: This is an amazing discussion to have. Let's go through this, will try to be peeking on this discussion more often (feel free to ping me on slack too). 

> I wanted to express that this is really a Reader who is aware of the size of (possibly unread) data.

Does it matter for purpose of this interface and the caller needs to understand if it's a reader/writer? I think this is actually a greater pattern and consistent with other Go code like e.g `http.Flusher` -> it's not `ReaderFlusher`.  I would vote for Sizer yes.

> Interface is defined such that Size can return error. What would Upload do in such case? Fail? Or would it be implementation specific? If it doesn’t fail, why require it in the first place?

We have exactly the same question now with Sizer or ReaderSizer casting, no?

> amount of code using Thanos objstore library, that would need to change. There is also some private code not shown on that page. I’m not sure the benefits outweigh causing so much trouble :)
> Typical object storage doesn’t need this information in advance, so requiring it by API seems counter-intuitive to me.

Hm, but actually we really really care to provide it otherwise we are doing thins inefficiently. And that amount of private or public Go has to be aware of this and leverage type mechanism instead of run time best effort casting. But you are right, that's a bigger change, happy with sizer for now 🤗

pracucci:
@bwplotka What're your thought after last comment from Peter? I think fixing the tracing wrapper would be good cause it's negatively affecting S3 users.

pstibrany:
> @bwplotka What're your thought after last comment from Peter? I think fixing the tracing wrapper would be good cause it's negatively affecting S3 users.

I believe this PR is now waiting for me to update the interface, but not requiring it in API. I'll send update soon.

pkg/objstore/objstore.go
```diff
@@ -137,12 +140,36 @@ func TryToGetSize(r io.Reader) (int64, error) {
...
+	case Sizer:
+		return f.Size()
 	}
 	return 0, errors.Errorf("unsupported type of io.Reader: %T", r)
 }
 
+// Sizer can return size of data.
+type Sizer interface {
+	// Size returns length of the data, or error, if it is not available.
+	// It is best to call Size before calling other methods on the type, as some io.Reader implementations may only return
```
pracucci, line 157: I get this comment but now that `Sizer` is a dedicated interface this comment is harder to understand.

pstibrany: What's your suggestion? Do you want me to revert back to `ReaderWithSize` or improve comment (suggestions welcome).

pracucci: Sorry for being unclear. I was just suggesting to improve the comment.

pkg/objstore/objstore.go
bwplotka: Good question. Let's iterate slowly (: First optional, but error should fail Upload IMO

pkg/objstore/objstore.go
```diff
@@ -137,12 +140,36 @@ func TryToGetSize(r io.Reader) (int64, error) {
...
+		return int64(f.Len()), nil
 	case *strings.Reader:
 		return f.Size(), nil
+	case Sizer:
+		return f.Size()
 	}
 	return 0, errors.Errorf("unsupported type of io.Reader: %T", r)
 }
 
+// Sizer can return size of data.
```
bwplotka, line 154:
```suggestion
// Sizer can return the size of all data. 
// This interface is meant to deliver optional size gathering capabilities of the other interface like io.Reader. The semantics depends on the main interface.
```

pkg/objstore/objstore.go
```diff
@@ -137,12 +140,36 @@ func TryToGetSize(r io.Reader) (int64, error) {
...
+	// Size returns length of the data, or error, if it is not available.
+	// It is best to call Size before calling other methods on the type, as some io.Reader implementations may only return
+	// length of unread data.
+	Size() (int64, error)
+}
+
+type nopCloserWithSize struct{ io.Reader }
+
+func (nopCloserWithSize) Close() error           { return nil }
+func (n nopCloserWithSize) Size() (int64, error) { return TryToGetSize(n.Reader) }
```
bwplotka, line 165: wowowo, that's not `nop` (partially).

pkg/objstore/objstore.go
bwplotka, line 157: I see how tricky it is now. If we assume `Size` is unread data then all our implementations have to be much more complex for no reason, as Upload only care about the overall initial size before reading 🙃 

Since we are in `objstore.Sizer`, maybe `ObjectSize` method will be easier to understand? I would focus on capabilities we want which is overall size. It will much more obvious. WDYT?

pkg/objstore/objstore.go
bwplotka: I would suggest creating `TrySizer` and just use ioutil.NopCloser WDYT?

bwplotka (dismissed):
This can work, some suggestions though. LGTM after addressing 🤗

pkg/objstore/objstore.go
pstibrany: > I see how tricky it is now. If we assume Size is unread data then all our implementations have to be much more complex for no reason, as Upload only care about the overall initial size before reading 🙃

What I want to express is that *callers* should call `Size()` before doing any other operations because *implementations* may be able to only answer `Size` with **remaining data** (like `(*bytes.Reader).Len()` method)

`Upload` is already doing that – calls `Size` before doing anything else, and requires no change.

bwplotka: Right. That's why I would propose naming this method `ObjectSize` to reflect the initial object size, no matter when you do this. 

But I guess it might be too picky. I would be fine to leave it as it is in your case. Happy with that (:

bwplotka:
Friendly ping 🤗

pkg/objstore/objstore.go
pstibrany: > That's why I would propose naming this method ObjectSize to reflect the initial object size, no matter when you do this.

Sure, we can rename it. We can also specify that Sizer must report full object size. 

What do you say if we rename `Sizer` to `ObjectSizer`, while at it?

pracucci (dismissed):
LGTM, thanks!

bwplotka (approved):

bwplotka:
Thanks! Do you mind rebasing, so we can use @pracucci fix (https://github.com/thanos-io/thanos/pull/4096).

pracucci:
> Thanks! Do you mind rebasing, so we can use @pracucci fix (#4096)?

I just did it. I'm going to upgrade Thanos in Cortex and I would be glad to bring this enhancement too.

kakkoyun (approved):
LGTM 🚀

Issue #3662, opened 2020-12-22
Change objstore Delete() contract to make it idempotent


pracucci:
### Is your proposal related to a problem?

The objstore `Delete()` is expected to return error if the object doesn't exist in the bucket:

https://github.com/thanos-io/thanos/blob/e9260e30f265f60e17d3f61ff361fb7357b3d34a/pkg/objstore/objstore.go#L44-L46

This design decision has some drawbacks:

- S3 "delete object" API is idemponent and Thanos S3 client (based on Minio) is idempotent as well. Basically it violates the contract.
- We have some places where we check for existence before deleting (eg. `block.Delete()`), while if it would be idempotent we could skip the check for existence at all.
- Design wise, isn't more RESTful having deletes idempotent?

However, I may be missing the reason behind this design decision, so I've opened an issue to discuss it.

### Describe the solution you'd like

I would propose to change objstore clients `Delete()` contract to make it idempotent. Object stores where delete is idempotent (eg. S3) would require no changes, while others would require handling the "not found" error status code (typically a 404, like on GCS).

bwplotka:
I am almost certain that we did this ONLY to match with "some" obj storage APIs we were inspired by (GCS) . We are definitely ok to make it idempotent. (: 

I kind of like explicitness. If you delete something that does not exists, I want to know that as a caller. This might mean some eventual consistency or race, so then caller can just ignore it. OR I really want to delete something and I want to know about deletion happening. 

The simple case will be instrumentation. If we delete I want to make that we deleted something. If 20 writes deletes I just have info "they tried to delete and maybe they deleted and maybe it was already deleted". Such instrumentation/metric will NOT tell us how many deletes were in DB (e.g objstore), but rather how many attempts were there.

Apart of instrumentation, I don't see other use cases for non-idempotent Delete, but the instrumentation one might be important, not sure. Ideally, allowing caller to decide makes sense? 

I think I would be happy to make it idempotent, our goal with objstore is indeed to reduce API calls.

pracucci:
> I think I would be happy to make it idempotent, our goal with objstore is indeed to reduce API calls.

I thought the rest of the comment was leading towards not making it idempotent. If it's idempotent, from the consumer perspective, then the `Delete()` would act the same exact way regardless the object existed or not in the storage.

bwplotka:
Yea, because I am torn. 

[...]

PR #3869, opened 2021-03-03
compact: clean up directories properly

GiedriusS:
I couldn't stop thinking about this code for some reason and I have
figured that I had missed one case in #3031. We need to also clean up
the directories in compaction groups. Compaction could fail & it would leave
some new directory with a random ULID on the disk. Before attempting to
do another compaction loop, we need to remove it as well because the
compaction process always produces a new, unique directory.

## Changes

`Compact()` now cleans up the subdirectories in compaction group directories. An error when a directory doesn't exist is ignored because it is a very likely situation with this change so it would be nice to avoid spam.

## Verification

Added e2e test

### DIFF
--- a/pkg/compact/compact.go
+++ b/pkg/compact/compact.go
@@ -957,6 +957,16 @@ func (c *BucketCompactor) Compact(ctx context.Context) (rerr error) {
 
 		ignoreDirs := []string{}
 		for _, gr := range groups {
+			groupIDs := []string{}
+			for _, grID := range gr.IDs() {
+				groupIDs = append(groupIDs, grID.String())
+			}
+
+			groupDir := filepath.Join(c.compactDir, gr.Key())
+			if err := runutil.DeleteAll(groupDir, groupIDs...); err != nil {
+				level.Warn(c.logger).Log("msg", "failed deleting non-compaction group sub-directories/files, some disk space usage might have leaked. Continuing", "err", err, "dir", groupDir)
+			}
+
 			ignoreDirs = append(ignoreDirs, gr.Key())
 		}
 
--- a/pkg/runutil/runutil.go
+++ b/pkg/runutil/runutil.go
@@ -164,6 +164,9 @@ func ExhaustCloseWithErrCapture(err *error, r io.ReadCloser, format string, a ..
 // dir except for the ignoreDirs directories.
 func DeleteAll(dir string, ignoreDirs ...string) error {
 	entries, err := ioutil.ReadDir(dir)
+	if os.IsNotExist(err) {
+		return nil
+	}
 	if err != nil {
 		return errors.Wrap(err, "read dir")
 	}
@@ -171,7 +174,7 @@ func DeleteAll(dir string, ignoreDirs ...string) error {
 
 	for _, d := range entries {
 		if !d.IsDir() {
-			if err := os.RemoveAll(filepath.Join(dir, d.Name())); err != nil {
+			if err := os.RemoveAll(filepath.Join(dir, d.Name())); err != nil && !os.IsNotExist(err) {
 				groupErrs.Add(err)
 			}
 			continue
@@ -186,7 +189,7 @@ func DeleteAll(dir string, ignoreDirs ...string) error {
 		}
 
 		if !found {
-			if err := os.RemoveAll(filepath.Join(dir, d.Name())); err != nil {
+			if err := os.RemoveAll(filepath.Join(dir, d.Name())); err != nil && !os.IsNotExist(err) {
 				groupErrs.Add(err)
 			}
 		}
--- a/test/e2e/compact_test.go
+++ b/test/e2e/compact_test.go
@@ -539,8 +539,26 @@ func TestCompactWithStoreGateway(t *testing.T) {
 
 	// No replica label with overlaps should halt compactor. This test is sequential
 	// because we do not want two Thanos Compact instances deleting the same partially
-	// uploaded blocks and blocks with deletion marks.
+	// uploaded blocks and blocks with deletion marks. We also check that Thanos Compactor
+	// deletes directories inside of a compaction group that do not belong there.
 	{
+		// Precreate a directory. It should be deleted.
+		// In a hypothetical scenario, the directory could be a left-over from
+		// a compaction that had crashed.
+		p := filepath.Join(s.SharedDir(), "data", "compact", "expect-to-halt", "compact")
+
+		testutil.Assert(t, len(blocksWithHashes) > 0)
+
+		m, err := block.DownloadMeta(ctx, logger, bkt, blocksWithHashes[0])
+		testutil.Ok(t, err)
+
+		randBlockDir := filepath.Join(p, compact.DefaultGroupKey(m.Thanos), "ITISAVERYRANDULIDFORTESTS0")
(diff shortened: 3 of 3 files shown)


pkg/runutil/runutil.go
```diff
@@ -164,14 +164,17 @@ func ExhaustCloseWithErrCapture(err *error, r io.ReadCloser, format string, a ..
 // dir except for the ignoreDirs directories.
 func DeleteAll(dir string, ignoreDirs ...string) error {
 	entries, err := ioutil.ReadDir(dir)
+	if os.IsNotExist(err) {
```
bwplotka, line 167: That is an important change. We are making `DeleteAll` idempotent which has some perks but also allows us to spam `DeleteAll` function where we does not need it 🤗 that's why we did not have it first place. Can we avoid this?

pkg/compact/compact.go
```diff
@@ -957,6 +957,16 @@ func (c *BucketCompactor) Compact(ctx context.Context) (rerr error) {
 
 		ignoreDirs := []string{}
 		for _, gr := range groups {
+			groupIDs := []string{}
+			for _, grID := range gr.IDs() {
+				groupIDs = append(groupIDs, grID.String())
+			}
+
+			groupDir := filepath.Join(c.compactDir, gr.Key())
```
bwplotka, line 965: Yea, this is getting bit out of control. That's why we kind of were deleting things upfront on the compactor startup, always in the first place, to not care about every tiny individual case.

For example  now we will still miss case where we crash on left dir behind, but when we start some group is missing (maybe sharding configuration changed) and now we never clean that directory (:

bwplotka (commented):
I understand this fix, but wonder if we can make it more reliable somehow 🤗

pkg/runutil/runutil.go
bwplotka: Actually there is bigger discussion you can join (: 

https://github.com/thanos-io/thanos/issues/3662

bwplotka: It's related to bucket API delete but same problem

pkg/compact/compact.go
pracucci: Should we consider rolling back #3031? :)

GiedriusS: Isn't that a bit of an extreme reaction? Yes, we were just cleaning everything without a second doubt but that's what I think should be improved & it's only natural that sometimes issues occur when doing a new thing. We aren't robots and sometimes mistakes slip through but that's why we have our release process and so on, right? If it's just a matter of cleaning the directories properly then I'm not convinced that the whole thing needs to be rolled back unless this becomes blocking on the Cortex part i.e. if you won't be able to update your dependency on Thanos because of this.

Give me a few days and I think we can fix it just fine.

GiedriusS: @bwplotka 
> For example now we will still miss case where we crash on left dir behind, but when we start some group is missing (maybe sharding configuration changed) and now we never clean that directory (:

How could this happen? The meta fetcher drops IDs that have been removed via the relabeling config and then that gets passed on to the syncer:
```go
		cf := baseMetaFetcher.NewMetaFetcher(
			extprom.WrapRegistererWithPrefix("thanos_", reg), []block.MetadataFilter{
				block.NewLabelShardedMetaFilter(relabelConfig),
				block.NewConsistencyDelayMetaFilter(logger, conf.consistencyDelay, extprom.WrapRegistererWithPrefix("thanos_", reg)),
				ignoreDeletionMarkFilter,
				duplicateBlocksFilter,
				noCompactMarkerFilter,
			}, []block.MetadataModifier{block.NewReplicaLabelRemover(logger, conf.dedupReplicaLabels)},
		)
sy, err = compact.NewSyncer(
			logger,
			reg,
			bkt,
			cf,
			duplicateBlocksFilter,
			ignoreDeletionMarkFilter,
			blocksMarked.WithLabelValues(metadata.DeletionMarkFilename),
			garbageCollectedBlocks,
			conf.blockSyncConcurrency)
```

And then we do this:
```go
		groups, err := c.grouper.Groups(c.sy.Metas())
		if err != nil {
			return errors.Wrap(err, "build compaction groups")
		}
```

So, I'm not sure how there could be compaction groups of blocks that have been filtered out. Could you please elaborate?

pkg/runutil/runutil.go
GiedriusS: We can check in advance whether the directory exists or not. Potentially we could only check the first compaction group but that's not a given because `--compaction.concurrency` could be more than `1`. Do you think this would be a better approach?

pkg/compact/compact.go
pracucci: > Isn't that a bit of an extreme reaction?

Sorry. Was a joke. I should have made it more clear, sorry.

bwplotka: haha, yea let's iterate @GiedriusS  (:

bwplotka: I mean both compaction and downsampling, but I just pointed one problem, there might be more. If there is nothing simple, then we need to just review properly etc, and I might need time for this. Just adding extra deletes "just in case" does not sound like efficient way of doing things. Not because it takes some compute power, is that it surprise reader and just unnecessary increase cognitive load.

Maybe I exaggerate, I will try to pull this code locally and check tomorrow (:

GiedriusS: @bwplotka so, did you have time? What's the verdict? Is there any other case I've missed?

bwplotka: Had zero time, added to https://github.com/thanos-io/thanos/issues/3946

bwplotka:
I propose one solution, wonder would that work @GiedriusS ?

https://github.com/thanos-io/thanos/pull/3869/commits/e49c546012926b99668c9a9bd82b7dff8ebb73d5

pkg/compact/compact.go
bwplotka: Proposed something (:

pkg/runutil/runutil.go
bwplotka: Proposed something.

bwplotka:
Build issues: quay is down.

pkg/runutil/runutil_test.go
```diff
@@ -125,35 +125,43 @@ func TestCloseMoreThanOnce(t *testing.T) {
...
-	_, err = os.Stat(filepath.Join(dir, "01EHBQRN4RF0HSRR1772KW0TN9"))
+	_, err = os.Stat(filepath.Join(dir, "b/"))
 	testutil.Assert(t, os.IsNotExist(err))
-
-	_, err = os.Stat(filepath.Join(dir, "01EHBQRN4RF0HSRR1772KW1TN8/"))
+	_, err = os.Stat(filepath.Join(dir, "file3"))
 	testutil.Assert(t, os.IsNotExist(err))
 
-	_, err = os.Stat(filepath.Join(dir, "01EHBQRN4RF0HSRR1772KW0TN8/"))
+	// Exits.
```
GiedriusS, line 160: Nit: `exists`

GiedriusS (commented):
Just one nit but otherwise LGTM. Thanks for being so proactive & your thoughtful review

Issue #3875, opened 2021-03-04
Add fmt.{Errorf}=github.com/pkg/errors.{Errorf} check in faillint


yashrsharma44:
Since this issue https://github.com/fatih/faillint/issues/10 is resolved, we can add this in the Makefile.

saswatamcode:
@yashrsharma44 Can I take this up? This involves adding the `faillint` check into the Makefile, right?

PR #3881, opened 2021-03-05
Added Faillint check for fmt.Errorf

Biswajitghosh98:
Fixes : #3875

### DIFF
--- a/.bingo/Variables.mk
+++ b/.bingo/Variables.mk
@@ -35,11 +35,11 @@ $(EMBEDMD): $(BINGO_DIR)/embedmd.mod
 	@echo "(re)installing $(GOBIN)/embedmd-v0.0.0-20181127031020-97c13d6e4160"
 	@cd $(BINGO_DIR) && $(GO) build -mod=mod -modfile=embedmd.mod -o=$(GOBIN)/embedmd-v0.0.0-20181127031020-97c13d6e4160 "github.com/campoy/embedmd"
 
-FAILLINT := $(GOBIN)/faillint-v1.5.0
+FAILLINT := $(GOBIN)/faillint-v1.7.0
 $(FAILLINT): $(BINGO_DIR)/faillint.mod
 	@# Install binary/ries using Go 1.14+ build command. This is using bwplotka/bingo-controlled, separate go module with pinned dependencies.
-	@echo "(re)installing $(GOBIN)/faillint-v1.5.0"
-	@cd $(BINGO_DIR) && $(GO) build -mod=mod -modfile=faillint.mod -o=$(GOBIN)/faillint-v1.5.0 "github.com/fatih/faillint"
+	@echo "(re)installing $(GOBIN)/faillint-v1.7.0"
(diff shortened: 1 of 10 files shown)


Biswajitghosh98:
cc @bwplotka , @yashrsharma44

pkg/ui/bindata.go
```diff
@@ -85,7 +85,7 @@ package ui
 import (
 	"bytes"
 	"compress/gzip"
-	"fmt"
+	"fmt" //lint:ignore faillint auto-generated file, hence skipping
```
yashrsharma44, line 88: Let's discuss if we want to modify this.

yashrsharma44 (approved):
Looks good to me, but apart from the point that I highlighted, we can merge this!

pkg/ui/bindata.go
Biswajitghosh98: WDYT @bwplotka ?
Should we edit the file or let it stay as it is ?

bwplotka: I think we cannot ): But maybe we can filter this file from faillint. I think even faillint was having some feature for ignoring generating files 🤔 Let's check in failint repo - We did not update this project recently

bwplotka (approved):
Amazing, one issue only 🤗

pkg/ui/bindata.go
GiedriusS: I think we could improve the `assets` Makefile target to add this line at the top of the file during the run. Then, it would be ignored accordingly. WDYT?

bwplotka: I think failint should ignore generated files by default. I asked @Biswajitghosh98 to check the upstream, let's see. (:

Biswajitghosh98: @bwplotka I was working on another issue, and this completely escaped from my mind. I'll do it first thing tomorrow morning. Sorry for the delay 😅

Biswajitghosh98:
@bwplotka @yashrsharma44 @GiedriusS I updated the faillint version, but for some reason, https://github.com/fatih/faillint/pull/15 isn't working as expected even though the `pkg/ui/bindata.go` file has the suffix "DO NOT EDIT" (ref : https://gotools.org/dmitri.shuralyov.com/go/generated#generated.go-L68)
Any thoughts where to go ahead with it ?

fatih:
@Biswajitghosh98 it needs to start and end with these lines:

```
"// Code generated ... DO NOT EDIT."
```

The file you mentioned in faillint (which can be seen here: https://raw.githubusercontent.com/thanos-io/thanos/main/pkg/ui/bindata.go), however starts with a different keyword:

```
// Package ui Code generated by go-bindata. (@generated) DO NOT EDIT.
```

If you remove the `Package ui ` prefix, it should work.

pkg/ui/bindata.go
yashrsharma44: > I think failint should ignore generated files by default. I asked @Biswajitghosh98 to check the upstream, let's see. (:

I agree with @GiedriusS 's suggestion, improving on the assets Makefile would work for us, since faillint doesn't skip files that do not adhere its comment naming convention - [link](https://github.com/thanos-io/thanos/pull/3881#issuecomment-802343531)

PR #3883, opened 2021-03-05
mixin: Allow excluding individual components through configuration

brancz:

## Changes

Allow excluding individual components alerts/dashboards through mixin configuration.

## Verification

Since by default all should still be included I tested that this is the case by re-generating the examples that are checked into this repo.

@kakkoyun @metalmatze

### DIFF
--- a/mixin/alerts/bucket_replicate.libsonnet
+++ b/mixin/alerts/bucket_replicate.libsonnet
@@ -6,7 +6,7 @@
     p99LatencyThreshold: 20,
   },
   prometheusAlerts+:: {
-    groups+: [
+    groups+: if thanos.bucket_replicate == null then [] else [
       {
         name: 'thanos-bucket-replicate',
         rules: [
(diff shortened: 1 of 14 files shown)


kakkoyun (approved):
LGTM 🚀

andrewgeller:
Some of the dashboard files (eg rule.libsonnet, receive.libsonnet) use the _rule.title_, _receive.title_ to contribute to `__overviewRows__`. I believe this makes using this feature to exclude the components impossible. I suppose `__overviewRows__` could use the same conditional logic?

kakkoyun:
@andrewgeller Thanks for pointing that out. The initial idea was to build up your own set of entry point by including or excluding these component-based files. Now we kind of steered away also an additional path which is fine, as you pointed we should address the issues if we caused any troubles. Could you open an issue so that we don't forget this? Or better a PR with a fix would be awesome 🤗

PR #3886, opened 2021-03-05
Tools: rewrite delete can delete series only if it matches all matchers

### LINKED ISSUE #3885: tools bucket rewrite: Impossible to have multiple matchers?
thanos v0.19.0-rc.0

**Object Storage Provider**:
S3

**What happened**:
I tried to rewrite some blocks in dry-run for now and I am not happy with the output...
I tried to add multiple matchers (either by multiple arrays or multiple label matchers) and for both solution, it is doing an "OR" matcher and not an "AND" matcher like prometheus:

For example if I try:
```
- matchers: \"{job=\\\"job-to-delete\\\",__name__=\\\"up\\\"}\"
```
It will delete ALL timeseries of the job "job-to-delete" AND all metrics "up" from all other jobs as well.

Same if I try something like:
```
- matchers: \"{job=\\\"job-to-delete\\\"}\"
- matchers: \"{__name__=\\\"up\\\"}\"
```

I would like only the "up" metrics of the job "job-to-delete" to be removed

Dry run a deletion with a matcher for a common metric (up) and a job:
```
thanos tools bucket rewrite --id=01EZW9YDX86ZGANKC5VY3E8DHZ --objstore.config-file=bucket-prd.yaml --rewrite.to-delete-config="
- matchers: \"{__name__=~\\\"up\\\",job=~\\\"anyjob\\\"}\"
" --dry-run --log.level=debug
```

`cat` the change.log and see that all "up" metrics are deleted wether or not part of "anyjob", and some "non up" metrics are deleted from the job "anyjob".

I would also expect something like:
```
- matchers: \"{job=\\\"job-to-delete\\\",job=\\\"another-job\\\"}\"
```
to never match and delete anything.

yeya24:

## Changes

Fixes #3885.
For all matchers in a single deletion request, we should use `AND` between matchers instead of `OR`.

## Verification

One test case added.

### DIFF
--- a/pkg/compactv2/compactor_test.go
+++ b/pkg/compactv2/compactor_test.go
@@ -274,6 +274,38 @@ func TestCompactor_WriteSeries_e2e(t *testing.T) {
 				NumChunks:  2,
 			},
 		},
+		{
+			name: "1 blocks + delete modifier, deletion request contains multiple matchers, delete second series",
+			input: [][]seriesSamples{
+				{
+					{lset: labels.Labels{{Name: "a", Value: "1"}, {Name: "b", Value: "1"}},
+						chunks: [][]sample{{{0, 0}, {1, 1}, {2, 2}, {10, 10}, {11, 11}, {20, 20}}}},
+					{lset: labels.Labels{{Name: "a", Value: "1"}, {Name: "b", Value: "2"}},
+						chunks: [][]sample{{{0, 0}, {1, 1}, {2, 2}}, {{10, 11}, {11, 11}, {20, 20}}}},
+					{lset: labels.Labels{{Name: "a", Value: "3"}},
+						chunks: [][]sample{{{0, 0}, {1, 1}, {2, 2}, {10, 12}, {11, 11}, {20, 20}}}},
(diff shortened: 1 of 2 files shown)


pkg/compactv2/compactor_test.go
```diff
@@ -274,6 +274,38 @@ func TestCompactor_WriteSeries_e2e(t *testing.T) {
 				NumChunks:  2,
 			},
 		},
+		{
+			name: "1 blocks + delete modifier, deletion request contains multiple matchers, delete second series",
```
squat, line 278: :+1:

squat (approved):
good find :+1:

bwplotka (approved):
Thanks! 🤗

bwplotka:
Should go on v0.19.0-rc.0?

squat:
Yeah i think so. Maybe this actually should have been made against that branch

bwplotka:
That's fine, let's replicate commit

bwplotka:
https://github.com/thanos-io/thanos/pull/3890

PR #3891, opened 2021-03-06
tools: Fix partial and empty matchers in rewrite

yeya24:

Follow-up of #3886.

## Changes

Require full matchers matching when processing deletion requests. Only partial match or empty label is not allowed.

Let's say a deletion request is `{a="1", b="2"}`, so the series we want to delete must at least matches `{a="1", b="2"}`. It can contain other unrelated labels and it doesn't matter.

We don't want to delete series that have an empty label value of the matchers like `{c="1"}` or series that only has partial matches like `{a="1"}` or `{b="2"}`.

### DIFF
--- a/pkg/compactv2/modifiers.go
+++ b/pkg/compactv2/modifiers.go
@@ -62,12 +62,9 @@ SeriesLoop:
 		for _, deletions := range d.d.deletions {
 			for _, m := range deletions.Matchers {
 				v := lbls.Get(m.Name)
-				if v == "" {
-					continue
-				}
 
 				// Only if all matchers in the deletion request are matched can we proceed to deletion.
-				if !m.Matches(v) {
+				if v == "" || !m.Matches(v) {
 					continue DeletionsLoop
 				}
 			}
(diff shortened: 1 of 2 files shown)


pkg/compactv2/modifiers.go
```diff
@@ -62,12 +62,9 @@ SeriesLoop:
 		for _, deletions := range d.d.deletions {
 			for _, m := range deletions.Matchers {
 				v := lbls.Get(m.Name)
-				if v == "" {
-					continue
-				}
 
 				// Only if all matchers in the deletion request are matched can we proceed to deletion.
-				if !m.Matches(v) {
+				if v == "" || !m.Matches(v) {
```
yeya24, line 67: `lbls.Get(m.Name)` returns an empty string if the label is not present in the label set. What if the label name is present but with an empty string as its value, is this a valid case in Prometheus?

If yes, we need to use `lbls.Has(m.Name)` here.

squat: ack, this is a good point, @yeya24

squat (approved):

pkg/compactv2/modifiers.go
yeya24: Just checked Prometheus and it is not allowed. So checking `v == ""` is fine.

squat: that's right, having empty value is the same as not having the label at all in Prometheus

squat (approved):

PR #3893, opened 2021-03-06
objstore: Provide the correct underlying size in more cases

zecke:
* [x ] Change is not relevant to the end user.

## Changes

Uploads of the shipper (and thanos tool bucket replicate) showed that the filesize could not be guessed. Look into the timing and tracing wrappers and optionally look into the S3 object as well to get the underlying size.

## Verification

Minimal checking with replicate and looking at the uploaded chunks.


GiedriusS:
Have you seen https://github.com/thanos-io/thanos/pull/3866? (:

zecke:
> Have you seen #3866? (:

Nice. This will avoid having to know minio at the objstore package level. :)

PR #3894, opened 2021-03-07
Retry and prevent cleanup of local blocks for upto 5 iterations

Biswajitghosh98:
Checking Proof of concept for #3744

### DIFF
--- a/pkg/block/block.go
+++ b/pkg/block/block.go
@@ -99,6 +99,7 @@ func Download(ctx context.Context, logger log.Logger, bucket objstore.Bucket, id
 // TODO(bplotka): Ensure bucket operations have reasonable backoff retries.
 // NOTE: Upload updates `meta.Thanos.File` section.
 func Upload(ctx context.Context, logger log.Logger, bkt objstore.Bucket, bdir string, hf metadata.HashFunc, retryCounter int) error {
+	var flag bool = false
 	df, err := os.Stat(bdir)
 	if err != nil {
 		return err
@@ -120,7 +121,7 @@ func Upload(ctx context.Context, logger log.Logger, bkt objstore.Bucket, bdir st
 	}
 
 	if meta.Thanos.Labels == nil || len(meta.Thanos.Labels) == 0 {
-		return errors.New("empty external labels are not allowed for Thanos block.")
+		return errors.New("empty external labels are not allowed for Thanos block")
 	}
 
 	meta.Thanos.Files, err = gatherFileStats(bdir, hf, logger)
@@ -135,25 +136,25 @@ func Upload(ctx context.Context, logger log.Logger, bkt objstore.Bucket, bdir st
 	if err := bkt.Upload(ctx, path.Join(DebugMetas, fmt.Sprintf("%s.json", id)), strings.NewReader(metaEncoded.String())); err != nil {
 		if retryCounter == 5 {
 			return cleanUp(logger, bkt, id, errors.Wrap(err, "upload debug meta file"))
-		} else {
-			return Upload(ctx, logger, bkt, bdir, hf, retryCounter+1)
 		}
+		flag = true
 	}
 
 	if err := objstore.UploadDir(ctx, logger, bkt, path.Join(bdir, ChunksDirname), path.Join(id.String(), ChunksDirname)); err != nil {
 		if retryCounter == 5 {
 			return cleanUp(logger, bkt, id, errors.Wrap(err, "upload chunks"))
-		} else {
-			return Upload(ctx, logger, bkt, bdir, hf, retryCounter+1)
 		}
+		flag = true
 	}
 
 	if err := objstore.UploadFile(ctx, logger, bkt, path.Join(bdir, IndexFilename), path.Join(id.String(), IndexFilename)); err != nil {
 		if retryCounter == 5 {
 			return cleanUp(logger, bkt, id, errors.Wrap(err, "upload index"))
-		} else {
-			return Upload(ctx, logger, bkt, bdir, hf, retryCounter+1)
 		}
+		flag = true
+	}
+	if flag && retryCounter < 5 {
+		return Upload(ctx, logger, bkt, bdir, hf, retryCounter+1)
 	}
 
 	// Meta.json always need to be uploaded as a last item. This will allow to assume block directories without meta file to be pending uploads.
--- a/pkg/block/block.go
+++ b/pkg/block/block.go
@@ -104,7 +104,7 @@ func Upload(ctx context.Context, logger log.Logger, bkt objstore.Bucket, bdir st
 
 // UploadWithRetry is a utility function for upload and acts as a workaround for absence of default parameters (which in this case is retryCounter = 0).
 func UploadWithRetry(ctx context.Context, logger log.Logger, bkt objstore.Bucket, bdir string, hf metadata.HashFunc, retryCounter int) error {
-	var flag bool = false
+	flag := false
 	df, err := os.Stat(bdir)
 	if err != nil {
 		return err
--- a/pkg/block/block.go
+++ b/pkg/block/block.go
@@ -98,7 +98,7 @@ func Download(ctx context.Context, logger log.Logger, bucket objstore.Bucket, id
 // It also verifies basic features of Thanos block.
 // TODO(bplotka): Ensure bucket operations have reasonable backoff retries.
 // NOTE: Upload updates `meta.Thanos.File` section.
-func Upload(ctx context.Context, logger log.Logger, bkt objstore.Bucket, bdir string, hf metadata.HashFunc) error {
+func Upload(ctx context.Context, logger log.Logger, bkt objstore.Bucket, bdir string, hf metadata.HashFunc, retryCounter int) error {
 	df, err := os.Stat(bdir)
 	if err != nil {
 		return err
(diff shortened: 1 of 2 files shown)


Biswajitghosh98:
cc @bwplotka @kakkoyun

pkg/block/block.go
```diff
@@ -99,6 +99,12 @@ func Download(ctx context.Context, logger log.Logger, bucket objstore.Bucket, id
 // TODO(bplotka): Ensure bucket operations have reasonable backoff retries.
 // NOTE: Upload updates `meta.Thanos.File` section.
 func Upload(ctx context.Context, logger log.Logger, bkt objstore.Bucket, bdir string, hf metadata.HashFunc) error {
+	return UploadWithRetry(ctx, logger, bkt, bdir, hf, 0)
+}
+
+// UploadWithRetry is a utility function for upload and acts as a workaround for absence of default parameters (which in this case is retryCounter = 0).
+func UploadWithRetry(ctx context.Context, logger log.Logger, bkt objstore.Bucket, bdir string, hf metadata.HashFunc, retryCounter int) error {
+	var flag bool = false
```
yashrsharma44, line 107:
```suggestion
	flag := false
```

yashrsharma44 (commented):
I think we can use a retry package https://pkg.go.dev/github.com/cenkalti/backoff/v4#Retry which abstracts our boilerplate code for writing retry functions. Has different types of retry logic out of the box.

Biswajitghosh98:
> I think we can use a retry package https://pkg.go.dev/github.com/cenkalti/backoff/v4#Retry which abstracts our boilerplate code for writing retry functions. Has different types of retry logic out of the box.

Do we need to import another package as an overload are we good with this logic itself ? WDYT ? @yashrsharma44

pkg/block/block.go
```diff
@@ -120,29 +126,40 @@ func Upload(ctx context.Context, logger log.Logger, bkt objstore.Bucket, bdir st
...
 	if err := meta.Write(&metaEncoded); err != nil {
 		return errors.Wrap(err, "encode meta file")
 	}
 
 	if err := bkt.Upload(ctx, path.Join(DebugMetas, fmt.Sprintf("%s.json", id)), strings.NewReader(metaEncoded.String())); err != nil {
-		return cleanUp(logger, bkt, id, errors.Wrap(err, "upload debug meta file"))
+		if retryCounter == 5 {
+			return cleanUp(logger, bkt, id, errors.Wrap(err, "upload debug meta file"))
+		}
+		flag = true
```
prmsrswt, line 145: Do we really need this flag? We can simply do `return UploadWithRetry(..., retryCounter+1)` here to get the same result.

pkg/block/block.go
```diff
@@ -99,6 +99,12 @@ func Download(ctx context.Context, logger log.Logger, bucket objstore.Bucket, id
 // TODO(bplotka): Ensure bucket operations have reasonable backoff retries.
 // NOTE: Upload updates `meta.Thanos.File` section.
 func Upload(ctx context.Context, logger log.Logger, bkt objstore.Bucket, bdir string, hf metadata.HashFunc) error {
+	return UploadWithRetry(ctx, logger, bkt, bdir, hf, 0)
+}
+
+// UploadWithRetry is a utility function for upload and acts as a workaround for absence of default parameters (which in this case is retryCounter = 0).
+func UploadWithRetry(ctx context.Context, logger log.Logger, bkt objstore.Bucket, bdir string, hf metadata.HashFunc, retryCounter int) error {
```
prmsrswt, line 106: From the name `UploadWithRetry` I will think it runs the `Upload` function with retries, but that is not the case here. It would have been better to have an function `Upload(...)` which handles the upload logic without retry, then a `UploadWithRetries` function that retries this `Upload` fn itself.


But we are only trying to retry when upload fails, so instead of having retries here, doesn't it makes more sense to have a `objstore.UploadDirWithRetries` that will handle this transparently.

pkg/block/block.go
```diff
@@ -120,29 +126,40 @@ func Upload(ctx context.Context, logger log.Logger, bkt objstore.Bucket, bdir st
...
 	}
 
 	if err := objstore.UploadFile(ctx, logger, bkt, path.Join(bdir, IndexFilename), path.Join(id.String(), IndexFilename)); err != nil {
-		return cleanUp(logger, bkt, id, errors.Wrap(err, "upload index"))
+		if retryCounter == 5 {
+			return cleanUp(logger, bkt, id, errors.Wrap(err, "upload index"))
+		}
+		flag = true
+	}
+	if flag && retryCounter < 5 {
```
prmsrswt, line 161: Why 5?

You can answer the question here, but for anyone visiting this code in the future, it's a mystery number. We need to at least create a const for this, something like `const UploadMaxRetries = 5` will do. But ideally, this value should be configurable by the users with a reasonable default. So you need to dynamically pass this value to the function as an argument.

prmsrswt (changes requested):
I have added a few comments on this implementation, but I feel this might not be the right place to solve this problem. Implementing retries for the `objstore.Upload/UploadDir` functions might be a better idea.

Also, the idea of retrying uploads with exponential backoff is worth a look, as upload errors can result from hitting the object storage rate limit.

pkg/block/block.go
Biswajitghosh98: My first commit was exactly that, after which i shifted to this `flag` method. 
Let's say we encounter an error and call `Upload` once again with retry x+1. If that executes successfully we'll be returned back to the scope of function with retry value x.
Actually I feel the flag check should take place in the end, here https://github.com/thanos-io/thanos/blob/fcb62932d5a01c576ea90a33d0efeabc142005ae/pkg/block/block.go#L173 rather than https://github.com/thanos-io/thanos/blob/fcb62932d5a01c576ea90a33d0efeabc142005ae/pkg/block/block.go#L161

pkg/block/block.go
Biswajitghosh98: It is doing exactly that. Isn't it ? @prmsrswt 😕

pkg/block/block.go
Biswajitghosh98: Yes, it was only a first draft to check if the logic itself works at the first place. I'd definitely go for a user defined retryCounter value if the logic recieves a green flag 💯

Biswajitghosh98:
> I have added a few comments on this implementation, but I feel this might not be the right place to solve this problem. Implementing retries for the `objstore.Upload/UploadDir` functions might be a better idea.
> 
> Also, the idea of retrying uploads with exponential backoff is worth a look, as upload errors can result from hitting the object storage rate limit.

I'd surely try to have a look at this as well 🤗

pkg/block/block.go
Biswajitghosh98: The problem with a separate function would be : we'd have to carry over the error which caused a failure, and keeping track of which error caused it to fail would become difficult otherwise. The only reason I made the UploadWithRetries function was to keep the integrity of Upload intact and not add any other parameter to it. 
It would have been really great if Golang supported default params 😁

pkg/block/block.go
yashrsharma44: Rather than using recursive calls for the same, you could create a loop in which you retry for `maxRetryCount` times, in this case, `5`, and if successful, you could break out of the loop. That would be more easier to follow, and you might not need a `flag` variable to do the same.

So something like this:
```go
for i := 0; i < maxRetryCount; i++ {
     if err:= bkt.Upload(...); err == nil {
           break // bucket upload was successful
     }
     log.Debug('retrying again...')
}
```

Biswajitghosh98: Yes, that makes sense, and instead of the for loop, I'll also go for exponential backoff for each upload

bwplotka (changes requested):
Hey, I am really sorry, but I think I mislead you in the https://github.com/thanos-io/thanos/issues/3744#issuecomment-795906349

Added comment to explain, all. This ticket is mostly upload removing need for deletions on sidecar, not the retry logic. Sidecar will retry in next cycle so we don't need this. In fact such retries on this level are not recommended, see https://github.com/thanos-io/thanos/issues/3907 for bigger discussion. 🤗 

hope that makes sense. Also we have another contributor trying to fix the same thing, so ideally we could help with https://github.com/thanos-io/thanos/pull/3756 (review, comment) before creating another PR, what do you think? (:  We can talk about this on tomorrows 1:1

Biswajitghosh98:
> Hey, I am really sorry, but I think I mislead you in the https://github.com/thanos-io/thanos/issues/3744#issuecomment-795906349
> 
> Added comment to explain, all. This ticket is mostly upload removing need for deletions on sidecar, not the retry logic. Sidecar will retry in next cycle so we don't need this. In fact such retries on this level are not recommended, see https://github.com/thanos-io/thanos/issues/3907 for bigger discussion. 🤗 
> 
> hope that makes sense. Also we have another contributor trying to fix the same thing, so ideally we could help with https://github.com/thanos-io/thanos/pull/3756 (review, comment) before creating another PR, what do you think? (:  We can talk about this on tomorrows 1:1 

No worries. We'll see where we go ahead with this tomorrow. I was anyway looking to break the initial jitters of committing to the codebase, and I guess I'm past it by now, so maybe we can also move forward with issues concerning vertical block sharding 😃

Biswajitghosh98:
> Definitely 🤗 
> Should we close this then?

Yes, sure. I'll close this one

Issue #3779, opened 2021-02-09
Improve Bucket LabelValues Call with matchers


yeya24:
### Describe the solution you'd like

Upstream Prometheus merges the improvement about LabelValues with matches today https://github.com/prometheus/prometheus/pull/8400. It would be nice to have this support in Thanos as well.

AC:
1. Add matchers support to TSDB store with the new [LabelValues interface](https://github.com/prometheus/prometheus/pull/8400/files#diff-168e8306e8ccb157b33ea81d4c73f550c64539e3045eca3b06c59f3d9cbadd9dR100).
2. For Prometheus store (sidecar), we can add matchers to the LabelValues query and send it to Prometheus. This is tricky because we also need to check the version of Prometheus. If the version doesn't meet, we fallback to series query.
3.  Implement this new LabelValues interface in Store gateway.

These steps can be done in different PRs.

Namanl2001:
If nobody is working on this then I and @Abhishek357 wants to take this up.

yeya24:
@Namanl2001 That would be great! It would be easier to start with the first and the second point.

bwplotka:
From Contributor Office Hour: Why porting LabelValues matchers, how it's improving performance?

Reason:
1. First: Isolation. Our multitenancy querying pattern is based on injecting label-matcher for `tenant` label which is usually part `external labels`.
2. Performance: You can return less data.

Namanl2001:
I am working on task 1: `Add matchers support to TSDB store`.
As the Prometheus version is updated in Thanos, I am blocked with few questions:

- Are we talking about matchers contained in `storepb.LabelValuesRequest`?
- Passing matchers along with `r.label` [here](https://github.com/thanos-io/thanos/blob/513707d368b3910db8cf4744cf7df48abf176a04/pkg/store/tsdb.go#L224) would give the expected results?

PS: It would be beneficial to get some hints on further debugging/testing.

yeya24:
> Are we talking about matchers contained in storepb.LabelValuesRequest?
Yes, you can use that.

> Passing matchers along with r.label here would give the expected results?
Yes. As the matchers in the `storepb.LabelValuesRequest` are not Prometheus matchers type, you need to first convert them to the correct struct. You can use the snippet below.

``` go
matchers, err := storepb.MatchersToPromMatchers(r.Matchers...)
// TODO: Handle the err above.
res, _, err := q.LabelValues(r.Label, matchers...)
```
[...]

PR #3897, opened 2021-03-08
Added matchers support to TSDB store

Namanl2001:
Fixes Task 1: `Add matchers support to TSDB store` of #3779 
This change improves the performance of the label-values lookup when matchers are specified. With this change, the Queriers won't need to fetch & sort & de-duplicate all the metrics, because they can directly call the `.LabelValues()` method and pass the matchers parameter.
### Verification
Added unit tests.

### DIFF
--- a/pkg/store/tsdb_test.go
+++ b/pkg/store/tsdb_test.go
@@ -348,9 +348,6 @@ func TestTSDBStore_LabelValues(t *testing.T) {
 			end: func() int64 {
 				return timestamp.FromTime(maxTime)
 			},
-			Matchers: []storepb.LabelMatcher{
-				{Type: storepb.LabelMatcher_EQ, Name: "foo", Value: "test"},
-			},
 		},
 		{
 			title:          "add another label value",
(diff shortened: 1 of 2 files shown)


pkg/store/tsdb_test.go
```diff
@@ -347,6 +348,9 @@ func TestTSDBStore_LabelValues(t *testing.T) {
 			end: func() int64 {
 				return timestamp.FromTime(maxTime)
 			},
+			Matchers: []storepb.LabelMatcher{
+				{Type: storepb.LabelMatcher_EQ, Name: "foo", Value: "test"},
+			},
```
yeya24, line 353: I suggest removing this matcher as it doesn't help anything. We can get the same result without matchers.

pkg/store/tsdb_test.go
```diff
@@ -361,6 +365,22 @@ func TestTSDBStore_LabelValues(t *testing.T) {
 				return timestamp.FromTime(maxTime)
 			},
 		},
+		{
+			title:          "add another label value",
```
yeya24, line 369: Let's rename the title. If we only want to test the effects of matchers, actually we don't need to add another label value. We can just use `{Type: storepb.LabelMatcher_EQ, Name: "foo", Value: "test1"}` matcher and the expected value is `test1`.

What about renaming it to "check label value matcher".

yeya24 (approved):
LGTM with some small nits. Btw it would be good to have one more test case that a matcher cannot match any existing series.

yeya24 (approved):
LGTM

PR #3901, opened 2021-03-09
Tools: Rewrite delete should match even if label does not exist

### LINKED ISSUE #3885: tools bucket rewrite: Impossible to have multiple matchers?

ahurtaud:

## Changes

Fixes #3885 - The fix provided by @yeya24 was not enough as it deleted a block matching only 1 matcher when others labelkeys do not exist.
The matcher check is done on every label keys, even if value is empty.

## Verification

Edited the test to "partial" match 1 matcher


squat (approved):

squat (approved):
Thanks again @ahurtaud

ahurtaud:
Hum not sure we want to merge as-is:

- some negative-match ( !="foo" ) was not thougth. We might want to follow this on https://github.com/thanos-io/thanos/pull/3891

PR #3903, opened 2021-03-10
Returning custom grpc code when reaching series/chunk limits

alanprot:

## Changes
This change allow to return a different/custom GRPC code - other than the generic `Aborted` - when the limits are hit via `SeriesLimiter` or `ChunksLimiter`.

The reason for this change is that it is hard for the caller to differentiate when the error returned by [Series](https://github.com/thanos-io/thanos/blob/main/pkg/store/bucket.go#L893) method was generated by a `limiter` or if it was a generic/transient error. Propagating the error code generated by the limiter will create the possibility to return any GRPC code on those cases.

* If the error generated by the `SeriesLimiter` and `ChunksLimiter` is a GRPC error, the [Series](https://github.com/thanos-io/thanos/blob/main/pkg/store/bucket.go#L893) will return a GRPC error with the same errorcode.
## Verification

* Unit tests

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -13,6 +13,7 @@ We use _breaking :warning:_ to mark changes that are not backward compatible (re
 ## Unreleased
 
 ### Added
+- [#3903](https://github.com/thanos-io/thanos/pull/3903) Store: Returning custom grpc code when reaching series/chunk limits.
 
 ### Fixed
 - [#3204](https://github.com/thanos-io/thanos/pull/3204) Mixin: Use sidecar's metric timestamp for healthcheck.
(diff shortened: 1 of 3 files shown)


pracucci (commented):
I'm 👍  with the change but I'm not sure the unit test makes much sense from the Thanos perspective. This change originates from a need we have in Cortex and the test is trying to mimick it, but from Thanos perspective how we're going to use it is irrelevant. Because of this I'm wondering if we should keep the change but revert the unit test changes.

alanprot:
Hi @pracucci, Thanks for your reply.

But i think if we are coding a behaviour here we should also test this behaviour here to make sure it does not change right?

Issue #3888, opened 2021-03-05
website: Create a new page for caching


wiardvanrij:
I would like to propose to add a new page `caching`

# Why?

Because there are 3 types of caches and they all share the same **base** configuration with the exception that:
- Query frontend cache has an extra param `expiration`
- Caching bucket has besides the default config, extra options for metadata and chunk caching

I would like to talk about caching in the first place;
- Why
- When
- How

Followed by going over the base configuration file and what the exceptions/differences are between each 'cache component'

# What would it resolve

I understand that each component has it's own page, but we have a lot of 'cross-reference' information which is relevant for each type of cache. I also believe there could be more/clear information regarding memcached. For instance how discovery via dnssrv works and how one should not use a loadbalancer endpoint. 

Furthermore it should make it easier by explaining that every cache component uses the same type of base configuration. Which IMO makes it easier for people to understand it.

Personally I have really struggled with implementing caching. For example: Wrong memcached setup, wrong cache size, wrong memcached discovery, wrong indentation on the caching bucket configuration. Now this could be that I'm not so bright but I also spend quite some time discovering what type of settings do 'what'. 

Eventually my setup is pretty slick. I want to make it easier for others to do the same :) 

# Resume
- New page on `/tip/thanos/caching.md`
- Gives background and detailed information about both `in-memory` and `memcached cache options in general
- Explains differences between each component
- Components that use/configure caching would link towards `/tip/thanos/caching.md#the-cache-type`

# Example
I've already started a little bit, mainly because I also want to implement some funky bash to retrieve the base variable defaultMemcachedConfig for automatic processing into the docs as default values. 

![image](https://user-images.githubusercontent.com/5786097/110163549-75fbee80-7df0-11eb-8eff-2ce5f44759c1.png)

wiardvanrij:
First init made :)

wiardvanrij:
Note: https://thanos.io/tip/components/store.md/ is missing the flags regarding caching buckets. I.e. `store.caching-bucket`

wiardvanrij:
As discussed in community meeting;

- Keep config/flag information on components pages
- Automate from code/structs into md
- Gather info from various parts/configs/components into that single caching page. Just like storage for example

PR #3908, opened 2021-03-10
[draft] caching website page

wiardvanrij:
Relates to mainly #3888 

and somewhat #3669 

This is an initial PR. I would love more people involved to make the text accurate, solid and useful.


docs/caching.md
```diff
@@ -0,0 +1,238 @@
...
+
+The index cache has no exceptions on the base configuration
+
+## Bucket cache specific for in-memory
+
+`in-memory` bucket cache is not supported. This is only possible with `memcached` as backend.
+
+## Query frontend specific for in-memory
+
+`validity` specifies cache valid time , If set to 0s, so using a default of 24 hours expiration time
```
wiardvanrij, line 149: I'm actually not sure if this is correct, as I could not find information on this :D

docs/caching.md
```diff
@@ -0,0 +1,238 @@
...
+---
+
+# Background
+
+Thanos can implement caching layers on multiple components. In general this should speed-up certain processes. There is no definitive guide on how one should setup an entire Thanos stack as each cluster and use-case is different. Especially what KPI's<sup>1</sup>, SLO's<sup>2</sup> and SLI's<sup>3</sup> matter to you, your customers and/or your stack should have influence on the setup.
+
+However it is possible to give guidance on what each component tries to achieve, what it is used for and how you could configure this.
+
+<sub>
+1. service level agreement<br />
```
wiardvanrij, line 15: I would love to remove the 'brs' but somehow markdown totally ruins the style for new lines by enters.

docs/caching.md
```diff
@@ -0,0 +1,238 @@
...
+  max_item_size: 0
+```
+
+## Index cache specific for in-memory
+
+The index cache has no exceptions on the base configuration
+
+## Bucket cache specific for in-memory
+
+`in-memory` bucket cache is not supported. This is only possible with `memcached` as backend.
```
wiardvanrij, line 145: I'm somewhat confused, as the current docs live on the website state that it is supported here: https://thanos.io/tip/components/store.md/#caching-bucket 

Yet I also saw a place were it states that only memcached is supported.

docs/caching.md
GiedriusS: Weird :thinking: was about to comment about this

docs/caching.md
GiedriusS: Maybe we forgot to update something? https://github.com/thanos-io/thanos/pull/3579

docs/caching.md
wiardvanrij: It adds a 1.5 newline between the 2nd and 3rd line. Not sure how to fix it 🥺

docs/caching.md
```diff
@@ -0,0 +1,238 @@
...
+
+Thanos Store Gateway supports an index cache to speed up postings and series lookups from TSDB blocks indexes
+
+## Bucket cache
+
+Thanos Store Gateway supports a “caching bucket” with chunks and metadata caching to speed up loading of chunks from TSDB blocks.
+
+## Query frontend cache
+
+Query Frontend supports caching query results and reuses them on subsequent queries. If the cached results are incomplete, Query Frontend calculates the required subqueries and executes them in parallel on downstream queriers. Query Frontend can optionally align queries with their step parameter to improve the cacheability of the query results.
```
GiedriusS, line 39: Query frontend also caches label requests :P

docs/caching.md
```diff
@@ -0,0 +1,238 @@
...
+blocks_iter_ttl: 5m
+metafile_exists_ttl: 2h
+metafile_doesnt_exist_ttl: 15m
+metafile_content_ttl: 24h
+metafile_max_size: 1MiB  
+```
+
+## Query frontend specific
+
+`expiration` specifies memcached cache valid time , If set to 0s, so using a default of 24 hours expiration time
```
GiedriusS, line 110: `expiration` means how long the data is stored in Memcached i.e. after that time Memcached evicts those items. Yep, the default is `24h` :hugs: I think we can improve this sentence:
```suggestion
`expiration` specifies how long memcached itself keeps items. After that time, memcached evicts those items. `0s` means the default duration of `24h`.
```

docs/caching.md
```diff
@@ -0,0 +1,238 @@
...
+```yaml
+type: IN-MEMORY
+config:
+  max_size: 0
+  max_item_size: 0
+```
+
+## Index cache specific for in-memory
+
+The index cache has no exceptions on the base configuration
```
GiedriusS, line 141: Full stop missing

GiedriusS (commented):
Maybe also worth mentioning that different services can share the same memcached instance(-s), for example? I see such questions regularly on Slack. Thank you for your work :hugs:

docs/caching.md
```diff
@@ -0,0 +1,238 @@
...
+# Background
+
+Thanos can implement caching layers on multiple components. In general this should speed-up certain processes. There is no definitive guide on how one should setup an entire Thanos stack as each cluster and use-case is different. Especially what KPI's<sup>1</sup>, SLO's<sup>2</sup> and SLI's<sup>3</sup> matter to you, your customers and/or your stack should have influence on the setup.
+
+However it is possible to give guidance on what each component tries to achieve, what it is used for and how you could configure this.
+
+<sub>
+1. service level agreement<br />
+2. service level objective<br />
+3. service level indicator
```
yashrsharma44, line 17:
```suggestion
- service level agreement
- service level objective
- service level indicator
```

docs/caching.md
```diff
@@ -0,0 +1,238 @@
+
+---
+title: Caching
+type: docs
+menu: thanos
+---
+
+# Background
+
+Thanos can implement caching layers on multiple components. In general this should speed-up certain processes. There is no definitive guide on how one should setup an entire Thanos stack as each cluster and use-case is different. Especially what KPI's<sup>1</sup>, SLO's<sup>2</sup> and SLI's<sup>3</sup> matter to you, your customers and/or your stack should have influence on the setup.
```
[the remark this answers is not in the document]
wiardvanrij: Yes, totally agree. I would like more input on this though as I lack the 100% correct knowledge on that subject.

wiardvanrij:
For easier reference a screenshot of the table of contents: 
image

wiardvanrij:
Recent update removed specific configuration information from component pages as I explained in the issue #3888 
IMO this makes each component easier to understand and specific caching configuration can now be found on the 'caching' page. 

I understand this might look weird and not logical but imagine the use-case: A new person that has no affinity with Thanos yet. The component page can be quite complicated. By moving specific information and guiding the user to the correct part, it should become more clear.

Please also checkout this branch and run it locally `make web-serve` to check it out. 

For easier reference i've included a screenshot of the table of contents of the caching page.  I'm 100% open for any feedback on this change in #3888 . It's just merely a suggestion, so don't worry about giving feedback. 


image

wiardvanrij:
> Maybe also worth mentioning that different services can share the same memcached instance(-s), for example? I see such questions regularly on Slack. Thank you for your work 🤗

I've added this, but I could use extra help on wording because I added the following note:

```
It however is not possible to use one memcached instance for multiple store backends as this provides conflicts on index keys.```

I've tested this myself with a Thanos setup that consists of two different setups having a different backend. What will happen if they both use the same memcached instance, is that it will fetch indexes from the index cache (?) and tries to fetch this from the storage backend. Yet it does not exists as it's not part of that setup.

wiardvanrij:
I'm going to close this. I need to refactor the page and include all new changes included. Will make a new PR today!

PR #3916, opened 2021-03-11
pkg/extflag: clarify help for path/content flags

simonpasquier:

## Changes

The help text stated that both flags could be defined at the same time
with higher priority for the path flag while in practice, they are
mutually exclusive.

## Verification

No functional change.


squat (approved):
thanks @simonpasquier :)

PR #3919, opened 2021-03-11
*: Allow to disable automatically setting CORS headers

### LINKED ISSUE #3918: Allow disabling setting CORS headers

We have a proxy in front of Thanos that manages things like authentication etc. which has it's own CORS policy, but since Thanos also sets CORS headers they end up being set multiple times which ends up conflicting.

Add a flag `--disable-cors`. That causes no CORS headers to be set.

Can't really think of another way, except maybe allowing more specific CORS configuration, but that doesn't appear to have been a problem for anyone else so far.

@thanos-io/thanos-maintainers

brancz:

## Changes

Add `--web.disable-cors` flag to every component that exposes an API, allowing to disable automatically setting CORS headers.

## Verification

Ran components with and without the flag using the `scripts/quickstart.sh` script and verified the headers are gone.

Closes #3918 

@thanos-io/thanos-maintainers


prmsrswt (approved):
LGTM!

Issue #3667, opened 2020-12-23
Give better log info regarding which memcached type the logs are from 


wiardvanrij:
When checking log messages from the Store, you could find something like this:

```
level=warn msg="failed to fetch items from memcached" err="read tcp 10.42.8.248:61293->10.42.8.159:11211: i/o timeout"
```

Yet it is not clear which memcached type is being referenced. This could be either the index cache or the bucket cache. 

It should include either `caching-bucket` or `index-cache` (somewhat in line with flags).

wiardvanrij:
Still needed

bwplotka:
Yes! Help wanted.

wiardvanrij:
Done

PR #3929, opened 2021-03-13
Store: Adds the name of the instantiated memcached client to log info

wiardvanrij:

## Changes

Relates to: #3667 

## Verification
Created a build but have not run e2e. Tested runtime of store.

### DIFF
--- a/pkg/cacheutil/memcached_client.go
+++ b/pkg/cacheutil/memcached_client.go
@@ -471,7 +471,7 @@ func (c *memcachedClient) getMultiSingle(ctx context.Context, keys []string) (it
 	// concurrency should be enforced.
 	if c.config.MaxGetMultiConcurrency > 0 {
 		if err := c.getMultiGate.Start(ctx); err != nil {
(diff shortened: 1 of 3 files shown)


pkg/cacheutil/memcached_client.go
```diff
@@ -222,6 +227,7 @@ func newMemcachedClient(
 	)
 
 	c := &memcachedClient{
+		name:        name,
 		logger:      logger,
```
pracucci, line 231: Instead of adding the the name key-value pair to each log entry you can just wrap the logger with that:

```suggestion
		logger:      log.With(logger, "name", name),
```

pracucci (approved):
LGTM (modulo a little suggestion)

pkg/cacheutil/memcached_client.go
wiardvanrij: Yes, that's way more elegant. Thanks and updated!

pracucci: Thanks!

PR #3931, opened 2021-03-15
Create BucketStoreOption to pass as functional options

metalmatze:

## Changes

The NewBucketStore has one of the longest list of parameters I know of :smile: 
Coming across this while reading some code (and seeing @bwplotka's TODO), I though I'd take a bit of time to start a refactoring.
Based on [Dave Cheney's Functional options for friendly APIs](https://dave.cheney.net/2014/10/17/functional-options-for-friendly-apis) I refactored all optional parameters to be passed with functional options.

I'm happy to refactor more, if we deem necessary (there are probably still 5+). 

## Verification

Tests still pass. :relaxed:


pkg/store/bucket.go
```diff
@@ -308,67 +309,102 @@ type noopGate struct{}
...
+func WithDebugLogging() BucketStoreOption {
+	return func(s *BucketStore) {
+		s.debugLogging = true
+	}
+}
+
 // NewBucketStore creates a new bucket backed store that implements the store API against
 // an object store bucket. It is optimized to work against high latency backends.
-// TODO(bwplotka): Move to config at this point.
 func NewBucketStore(
```
pracucci, line 366: What I liked about the previous approach (taking all arguments in input) is that, whenever there's a change here, downstream projects compilation breaks and so it's very loud. With the options approach this is no more guaranteed.

metalmatze: Yeah, that's a fair point. On the other hand passing 10x `nil` to the constructor doesn't seem very nice either. I guess - tradeoffs.

pracucci: If passing 10x nil is due by tests maybe we can have a wrapper function `newTestBucketStore()`? Or am I missing anything?

metalmatze: I guess that would work too, yes.
So far I've never run into problems with the functional options either and it's pretty commonly used.

pracucci: Fair enough. I don't have a strong opinion. My comment is definitely not a blocker.

bwplotka: I see @pracucci point. The sad thing about BucketStore is that it's only used once. Such options are very much more needed for things which are heavily used e.g `block.Upload` function.

I think I am fine with both options here... No strong opinion (:

bwplotka: Plus looks like it's done for optional structs so maybe ok? Less work for you @pracucci too?

pracucci: As already mentioned I don't have a strong opinion. I see why you're heading towards functional options and I'm fine too.

bwplotka:
Any other opinions? =D

bwplotka:
@squat @brancz @GiedriusS ((:

kakkoyun (approved):
LGTM 🚀 

I guess the next thing to discuss when to use which pattern for the functions that have too many parameterers. We can make this a part of our style guide.
e.g if a function has X many parameters, convert it to a configuration struct. If those arguments are optional consider using the functional option pattern.

WDYT?

bwplotka:
Yes, a style guide entry would be amazing (: 

Let's merge this then, anyway. (:

bwplotka:
Thanks @metalmatze

PR #3935, opened 2021-03-15
Use smaller per part size to use lower memory

harshavardhana:

### DIFF
--- a/pkg/objstore/s3/s3.go
+++ b/pkg/objstore/s3/s3.go
@@ -67,8 +67,8 @@ var DefaultConfig = Config{
 		MaxConnsPerHost:       0,
 	},
 	// Minimum file size after which an HTTP multipart request should be used to upload objects to storage.
-	// Set to 128 MiB as in the minio client.
-	PartSize: 1024 * 1024 * 128,
+	// Set to 16 MiB as in the minio client.
+	PartSize: 1024 * 1024 * 16,
 }
 
 // Config stores the configuration for s3 bucket.
@@ -452,7 +452,7 @@ func (b *Bucket) Upload(ctx context.Context, name string, r io.Reader) error {
 	// partSize cannot be larger than object size.
 	partSize := b.partSize
 	if size < int64(partSize) {
-		partSize = 0
+		partSize = size // set partSize to the object size
 	}
 	if _, err := b.client.PutObject(
 		ctx,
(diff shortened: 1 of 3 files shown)


pkg/objstore/s3/s3.go
```diff
@@ -67,8 +67,8 @@ var DefaultConfig = Config{
 		MaxConnsPerHost:       0,
 	},
 	// Minimum file size after which an HTTP multipart request should be used to upload objects to storage.
-	// Set to 128 MiB as in the minio client.
-	PartSize: 1024 * 1024 * 128,
+	// Set to 16 MiB as in the minio client.
+	PartSize: 1024 * 1024 * 16,
```
bwplotka, line 71: Let's be careful here. We don't want such small parts to be uploaded as multipart. We want 128MB multi-parts. But it does not mean that it should hold 128MB or more in memory, so I would fix minio client rather 🤔

pkg/objstore/s3/s3.go
```diff
@@ -452,7 +452,7 @@ func (b *Bucket) Upload(ctx context.Context, name string, r io.Reader) error {
 	// partSize cannot be larger than object size.
 	partSize := b.partSize
 	if size < int64(partSize) {
-		partSize = 0
+		partSize = size // set partSize to the object size
```
bwplotka, line 455: I think we wanted something else here. PartSize = 0 was meaning we don't want multipart.

bwplotka (changes requested):
I think this is changing what was actually intended in the configuration as `PartSize`, let's figure out solution first on issue. Also ideally any of such changes will be benchmarked first, before applying 🤗

bwplotka:
Performed some benchmarks and added benchmark to this PR 🤗 

```
	// Uploading 100MB with:
	// PartSize: 1024 * 1024 * 100 = BenchmarkUpload-12    	      84	 416025119 ns/op	421129826 B/op	     946 allocs/op  (400MB allocated for single op)
	// ParSize:  1024 * 1024 * 25  = BenchmarkUpload-12    	     100	 273648888 ns/op	106639830 B/op	    1576 allocs/op (100MB overallocated).
	// PartSize: 1024 * 1024 * 8   = BenchmarkUpload-12    	      97	 254052232 ns/op	35392395 B/op	    3078 allocs/op (Still 33MB allocated [8x4])
```

As expected @harshavardhana, wonder how it was before this change against race? 🤔

bwplotka:
Thanks for help @harshavardhana 🤗 

Closing this in favor of https://github.com/thanos-io/thanos/pull/3968

PR #3940, opened 2021-03-18
Sidecar: Added matchers support

Namanl2001:
Fixes task 2: `For Prometheus store (sidecar), add matchers to the LabelValues query` of #3779 

### Changes
Implemented API `/api/v1/status/buildinfo` which returns the buildinfo of prometheus. Calling this API once when the sidecar starts and using it in `pkg/store/prometheus.go` `labelValues` method, if the version is `> "2.24"` used labelValues query otherwise fallback to series query.

### Verification
Added unit tests


pkg/store/prometheus.go
```diff
@@ -496,10 +499,31 @@ func (p *PrometheusStore) LabelValues(ctx context.Context, r *storepb.LabelValue
...
+		matchers, err := storepb.MatchersToPromMatchers(r.Matchers...)
+		if err != nil {
+			return nil, status.Error(codes.Internal, err.Error())
+		}
+		sers, err = p.client.SeriesInGRPC(ctx, p.base, matchers, r.Start, r.End)
+		if err != nil {
+			return nil, err
+		}
+		for _, s := range sers {
+			vals = append(vals, s[r.Label])
```
yeya24, line 525: It is possible that the series doesn't contain the required label. You need to check it first.

pkg/store/prometheus.go
```diff
@@ -496,10 +499,31 @@ func (p *PrometheusStore) LabelValues(ctx context.Context, r *storepb.LabelValue
...
+		}
 	}
-	sort.Strings(vals)
```
yeya24, line 503: sorting is still needed for the second case IMO.

pkg/promclient/promclient.go
```diff
@@ -607,6 +607,34 @@ func (c *Client) AlertmanagerAlerts(ctx context.Context, base *url.URL) ([]*mode
...
+		Data struct {
+			Version string `json:"version"`
+		} `json:"data"`
+	}
+
+	if err = json.Unmarshal(body, &b); err != nil {
+		return "", errors.Wrap(err, "unmarshal build info API response")
+	}
+
+	return b.Data.Version, nil
```
yeya24, line 635: Okay so this function just wants to return the version of the Prometheus instead of returning all the buildInfo. So maybe let's rename it to `BuildVersion`? WDYT

pkg/store/prometheus.go
```diff
@@ -496,9 +499,32 @@ func (p *PrometheusStore) LabelValues(ctx context.Context, r *storepb.LabelValue
...
-	if err != nil {
-		return nil, err
+	var (
+		vals []string
+		sers []map[string]string
+		err  error
+	)
+
+	version := p.promVersion()
+	if version > "2.24" {
```
yeya24, line 509: It's better to use a library like https://github.com/blang/semver, but this seems fine.

pkg/store/prometheus.go
```diff
@@ -496,9 +499,32 @@ func (p *PrometheusStore) LabelValues(ctx context.Context, r *storepb.LabelValue
...
+		matchers, err := storepb.MatchersToPromMatchers(r.Matchers...)
+		if err != nil {
+			return nil, status.Error(codes.Internal, err.Error())
+		}
+		sers, err = p.client.SeriesInGRPC(ctx, p.base, matchers, r.Start, r.End)
+		if err != nil {
+			return nil, err
+		}
+		for _, s := range sers {
+			if val, exists := s[r.Label]; exists {
```
yeya24, line 524: @Namanl2001, please take a look at https://github.com/thanos-io/thanos/blob/main/pkg/api/query/v1.go#L870-L886.
Here `s.[r.Label]` might get duplicate values so you need to use a set to deduplicate first and then convert it to a slice later.

pkg/store/prometheus.go
Namanl2001: So no need to use this library. Right?

pkg/store/prometheus.go
Namanl2001: Addressed!

yeya24 (commented):
You still need to fix the failing test cases.

pkg/store/prometheus.go
```diff
@@ -496,9 +499,46 @@ func (p *PrometheusStore) LabelValues(ctx context.Context, r *storepb.LabelValue
...
-		return nil, err
+	var (
+		sers []map[string]string
+		err  error
+	)
+
+	vals := []string{}
+	version := p.promVersion()
+
+	if version > "2.24" {
```
yeya24, line 510: Let's add `if len(r.Matchers) == 0 {` to this line:

```
if len(r.Matchers) == 0 or version > "2.24" {
...
}
```

Namanl2001: Nice Thought! 👍 
Addressed

cmd/thanos/sidecar.go
```diff
@@ -209,14 +209,41 @@ func runSidecar(
...
+		g.Add(func() error {
+			// We retry infinitely until we reach and fetch BuildVersion from our Prometheus.
+			err := runutil.Retry(2*time.Second, ctx.Done(), func() error {
+				if err := m.BuildVersion(ctx); err != nil {
+					level.Warn(logger).Log(
+						"msg", "failed to fetch prometheur version. Is Prometheus running? Retrying",
+						"err", err,
+					)
+					return err
+				}
```
yeya24, line 223: I am still thinking about this logic.
This doesn't work well when Prometheus version is < 2.14.0, where `/api/v1/status/buildinfo` doesn't exist. In this case, we just get 404s but we are still retrying indefinitely. This seems wrong to me.

We can only retry for a short time or we check the return error to stop retrying.

yashrsharma44: Lets retry for lets say `N` times (maybe `N=5`) and exit if retry fails?

cmd/thanos/sidecar.go
```diff
@@ -209,14 +209,41 @@ func runSidecar(
...
 		})
 	}
+	{
+		ctx, cancel := context.WithCancel(context.Background())
+		g.Add(func() error {
+			// We retry infinitely until we reach and fetch BuildVersion from our Prometheus.
+			err := runutil.Retry(2*time.Second, ctx.Done(), func() error {
+				if err := m.BuildVersion(ctx); err != nil {
+					level.Warn(logger).Log(
+						"msg", "failed to fetch prometheur version. Is Prometheus running? Retrying",
```
yashrsharma44, line 219:
```suggestion
						"msg", "failed to fetch prometheus version. Is Prometheus running? Retrying",
```

cmd/thanos/sidecar.go
yashrsharma44: Also for Prometheus `<2.14.0` this function would fail, so we also need to decide how to fetch build version for `<2.14.0`

Namanl2001: As discussed in contributor hours:
Let’s check for 404, if it’s present, set 0 value for promVersion

cmd/thanos/sidecar.go
```diff
@@ -209,14 +209,40 @@ func runSidecar(
...
+
+				level.Info(logger).Log(
+					"msg", "successfully loaded prometheus version",
+				)
+				return nil
+			})
+			if err != nil {
+				return errors.Wrap(err, "failed to get prometheus version")
+			}
+			return nil
```
yeya24, line 233: If you run this in a goroutine and it ends, the sidecar program also stops running because of the run group. Please make sure your code doesn't break the functionality. You can see the tests failed.

I think it is unnecessary to have a separate goroutine for checking the Prometheus version as it is unlikely to change.
We can run it in the main thread once? WDYT? @yashrsharma44 @Namanl2001 @bwplotka

bwplotka: We could put it somewhere here: https://github.com/thanos-io/thanos/pull/3940/files#diff-a89173b10deb4d469ed5524c1addf5b154030c04f1001f9dc91676b7966f1127R157

pkg/store/prometheus.go
```diff
@@ -496,9 +499,39 @@ func (p *PrometheusStore) LabelValues(ctx context.Context, r *storepb.LabelValue
...
+	var (
+		sers []map[string]string
+		err  error
+	)
+
+	vals := []string{}
+	version := p.promVersion()
+
+	if len(r.Matchers) == 0 || version > "2.24" {
+		vals, err = p.client.LabelValuesInGRPC(ctx, p.base, r.Label, r.Matchers, r.Start, r.End)
```
bwplotka, line 511: Are we making gRPC call?

bwplotka: YES

pkg/store/prometheus.go
```diff
@@ -496,9 +499,39 @@ func (p *PrometheusStore) LabelValues(ctx context.Context, r *storepb.LabelValue
...
+		matchers, err := storepb.MatchersToPromMatchers(r.Matchers...)
+		if err != nil {
+			return nil, status.Error(codes.Internal, err.Error())
+		}
+		sers, err = p.client.SeriesInGRPC(ctx, p.base, matchers, r.Start, r.End)
+		if err != nil {
+			return nil, err
+		}
+
+		// using set to handle duplicate values.
```
bwplotka, line 525:
```suggestion
		// Using set to handle duplicate values.
```

bwplotka (commented):
Some comments during Community call

Namanl2001:
As suggested, removed the separate goroutine for fetching the Prometheus version and moved the code block to an existing goroutine. 
The failing circleci test seems unrelated to me.

pkg/store/prometheus.go
yeya24: The naming makes me confused, too. We are making HTTP calls but return GRPC error.

yeya24 (approved):
LGTM. @Namanl2001  Thanks for your patience and great work!

yeya24:
Need another look for this pr to merge. Maybe @yashrsharma44 @bwplotka ?

pkg/store/prometheus.go
```diff
@@ -496,9 +499,39 @@ func (p *PrometheusStore) LabelValues(ctx context.Context, r *storepb.LabelValue
...
+		matchers, err := storepb.MatchersToPromMatchers(r.Matchers...)
+		if err != nil {
+			return nil, status.Error(codes.Internal, err.Error())
+		}
+		sers, err = p.client.SeriesInGRPC(ctx, p.base, matchers, r.Start, r.End)
+		if err != nil {
+			return nil, err
+		}
+
+		// Using set to handle duplicate values.
```
yashrsharma44, line 525: Although we intend a `set` behaviour, we are using `map` right? Lets change it to `map`. WDYT?
```suggestion
		// Using map to handle duplicate values.
```

pkg/store/prometheus_test.go
```diff
@@ -420,7 +420,11 @@ func TestPrometheusStore_LabelValues_e2e(t *testing.T) {
 	u, err := url.Parse(fmt.Sprintf("http://%s", p.Addr()))
 	testutil.Ok(t, err)
 
-	proxy, err := NewPrometheusStore(nil, nil, promclient.NewDefaultClient(), u, component.Sidecar, getExternalLabels, nil)
+	version, err := promclient.NewDefaultClient().BuildVersion(ctx, u)
```
yashrsharma44, line 423: Can you add a test case where the version is **malformed or incorrect**? I am not sure how would that behave, so that might be a good idea to add one and define the behaviour that we might want to do when version in **malformed**. 😄

pkg/store/prometheus.go
```diff
@@ -496,9 +499,39 @@ func (p *PrometheusStore) LabelValues(ctx context.Context, r *storepb.LabelValue
...
-	vals, err := p.client.LabelValuesInGRPC(ctx, p.base, r.Label, nil, r.Start, r.End)
-	if err != nil {
-		return nil, err
+	var (
+		sers []map[string]string
+		err  error
+	)
+
+	vals := []string{}
+	version := p.promVersion()
```
yashrsharma44, line 508: `p.promVersion` might return some garbage value if the `m.BuildVersion` doesn't run correctly, so let's make sure `version` is a correct version string, and not some garbage value.

pkg/store/prometheus_test.go
yashrsharma44, line 423: Let's add a test case where the `version` is a garbage value, and check how your code behaves? However, let's ask @yeya24 @bwplotka if this might occur in the first place or not 🤔

yashrsharma44 (changes requested):
I have added some nits regarding some edge cases when the `version` is not a valid string. If this case sounds unlikely, we can go ahead with the current implementation, looks good to me.

pkg/store/prometheus.go
Namanl2001: > Although we intend a set behaviour, we are using map right?

Yes
So should we change it to `Using map to deduplicate values` ?

pkg/store/prometheus_test.go
Namanl2001: I guess `version` could be garbage value only if `GET /api/v1/status/buildinfo` API returns some garbage value for the Prometheus version. But I think this API would be throwing some err instead of returning garbage value. Not sure though.

pkg/store/prometheus.go
yeya24: We can keep the comment as map[string]struct{} is a common pattern in go for a set

pkg/store/prometheus_test.go
yeya24: Malformed version happens when a non-release prometheus is used.
For this case, I recommend using semver to get the version. If it is malformed, we just use the series way

Namanl2001: > If it is malformed, we just use the series way

What does `malformed version` exactly mean? Does it mean 'incorrect' version or version something like this `1.2.3-beta.1` (pre-release version)?

yeya24: You can take a look at https://semver.org/.
`1.2.3-beta.1` is valid. `Malformed version` here means the version that cannot be parsed by semver. 

For example, if you build Prometheus locally, the version would be your git branch name like `test`. `test` is not of format `x.y.z` so that's the `malformed version`.
You can also check it with https://jubianchi.github.io/semver-check/#/

Namanl2001: Gotta, Thanks. Let's do it.

Namanl2001:
PHAL! @yeya24 @yashrsharma44

pkg/store/prometheus.go
```diff
@@ -496,9 +500,51 @@ func (p *PrometheusStore) LabelValues(ctx context.Context, r *storepb.LabelValue
...
+		baseVer, err := semver.Make("2.24.0")
+		if err != nil {
+			return nil, err
+		}
+		if version.Compare(baseVer) == 1 {
+			lvc = true
+		}
+	}
+
+	if len(r.Matchers) == 0 || lvc {
```
yashrsharma44, line 523:
```suggestion
	// Make sure the prometheus version is a valid semver type.
	version, err := semver.Parse(v)
	if err != nil {
	   return nil, err
	}
	
        baseVer, err := semver.Make("2.24.0")
        if err != nil {
	   return nil, err
        }
	
	if len(r.Matchers) == 0 || version.Compare(baseVer) == 1 {
```

yashrsharma44: Lets simplify things, return early if an error - https://thanos.io/tip/contributing/coding-style-guide.md/#control-structure-prefer-early-returns-and-avoid-else

yashrsharma44 (changes requested):
Suggested some coding style, otherwise LGTM 🔥

pkg/store/prometheus.go
Namanl2001: I think if the version is malformed we can still make a series call instead of returning the error. Please see `If it is malformed, we just use the series way` in https://github.com/thanos-io/thanos/pull/3940#discussion_r621221476

pkg/store/prometheus.go
```diff
@@ -496,9 +500,51 @@ func (p *PrometheusStore) LabelValues(ctx context.Context, r *storepb.LabelValue
...
+	)
+
+	lvc := false // LabelValuesCall
+	vals := []string{}
+	v := p.promVersion()
+
+	version, err := semver.Parse(v)
+	if err == nil {
+		baseVer, err := semver.Make("2.24.0")
+		if err != nil {
```
yashrsharma44, line 515: If this is true(https://github.com/thanos-io/thanos/pull/3940/files#r622405182), then we should not exit early. Then we can remove the error check.

Namanl2001: > Then we can remove the error check.

Sorry, are you saying about the error check at Line 515? As it is for handling error retuned while initializing the `baseVer`.

yashrsharma44: Yup, if it errors out, we can revert to our original behaviour of - 
> we can still make a series call instead of returning the error

Namanl2001: Does this look good?
```
version, err1 := semver.Parse(v)

baseVer, err2 := semver.Make("2.24.0")

if err1 == nil && err2 == nil && version.Compare(baseVer) == 1 {

    lvc = true

}

```

yashrsharma44: Apologies for elongating this, but we can simplify things by using this [package](https://pkg.go.dev/golang.org/x/mod/semver#Compare)

I checked their docs and it stated for their [Compare](https://pkg.go.dev/golang.org/x/mod/semver#Compare) function, we can provide two strings, if the promVersion is invalid, the value would be `0`, so we dont have to worry about that. So lets use this - 

```
defaultVersion := "2.24.0"
if len(r.Matchers) == 0 || semver.Compare(version, defaultVersion) == 1 {
```

Our current implementation is making it unnecessarily verbose and by using the above package, we can simplify things 😄

Namanl2001: Any thoughts on this @yeya24? As you earlier suggested to use a library https://github.com/thanos-io/thanos/pull/3940#discussion_r606446107

yeya24: If we already use the semver library then let's use it here as well.

pkg/store/prometheus.go
```diff
@@ -496,9 +500,47 @@ func (p *PrometheusStore) LabelValues(ctx context.Context, r *storepb.LabelValue
...
+		sers []map[string]string
+		err  error
+	)
+
+	lvc := false // LabelValuesCall
+	vals := []string{}
+	v := p.promVersion()
+
+	version, err1 := semver.Parse(v)
+	baseVer, err2 := semver.Make("2.24.0")
```
yeya24, line 513: You can make it a global variable and we know it doesn't throw an error

pkg/store/prometheus_test.go
yeya24: I am okay to have this in the next pr.

yeya24 (changes requested):

pkg/store/prometheus.go
Namanl2001: Why should we make it a global variable? 🤔  We can still keep `baseVer` a local variable and don't check for an error while declaring and initializing it.

yeya24: For performance. If you call `LabelValues` 1k times, then `semver.Make("2.24.0")` will be called 1k times. If you make it a global variable then it is parsed only once.

yeya24 (approved):
LGTM

yeya24:
cc @yashrsharma44 @bwplotka for a final review. If there is no objection, I can merge it tomorrow.

yashrsharma44 (approved):
LGTM 💪

PR #3947, opened 2021-03-21
Fix broken `service-discovery`  link

LeviHarrison:

## Changes

I fixed a broken link to the `service-discovery` page.

## Verification

Now, when you click on that link it brings you to the correct page.


bwplotka (approved):
Thanks! (: We have long term plan to fix all the links, but this will do for now!

PR #3948, opened 2021-03-22
Make NewInstrumentationMiddleware buckets configurable

metalmatze:

## Changes

Set Thanos Receive specific histogram buckets for http_duration_seconds_bucket so that we can measure Receiver latency properly. Highest bucket being 5s.

It was always 99ms during my benchmarks... 
Which is unsurprising when most requests fall within the buckets `[0.01, 0.1]`

## Verification

I can see proper p99 latencies between 10ms and 100ms now. :)

### DIFF
--- a/pkg/receive/handler.go
+++ b/pkg/receive/handler.go
@@ -150,7 +150,9 @@ func NewHandler(logger log.Logger, o *Options) *Handler {
 
 	ins := extpromhttp.NewNopInstrumentationMiddleware()
 	if o.Registry != nil {
-		ins = extpromhttp.NewInstrumentationMiddleware(o.Registry)
+		ins = extpromhttp.NewInstrumentationMiddleware(o.Registry,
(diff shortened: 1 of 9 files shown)


squat (approved):
lgtm

pkg/receive/handler.go
```diff
@@ -150,7 +150,9 @@ func NewHandler(logger log.Logger, o *Options) *Handler {
 
 	ins := extpromhttp.NewNopInstrumentationMiddleware()
 	if o.Registry != nil {
-		ins = extpromhttp.NewInstrumentationMiddleware(o.Registry)
+		ins = extpromhttp.NewInstrumentationMiddleware(o.Registry,
+			[]float64{0.001, 0.005, 0.01, 0.02, 0.03, 0.04, 0.05, 0.06, 0.07, 0.08, 0.09, 0.1, 0.25, 0.5, 0.75, 1, 2, 3, 4, 5},
```
kakkoyun, line 154: 👍

PR #3950, opened 2021-03-23
mixin: Increase the evaluation period for no rule eval alert for ruler

kakkoyun:

Increase the interval for the alert query to prevent flappy fires.

## Verification

* `make example-rules-lint`


metalmatze (approved):
Cool! LGTM
Could you quickly explain why this changed from 2m to 5m?

kakkoyun:
Sorry for missing out on the description. The intention is to prevent flappy fires of this alert. With the current state, the alert generates lots of false positives in our environments.

PR #3960, opened 2021-03-24
pkg/rules: fix deduplication of equal alerts with different labels

s-urbaniak:
Currently, if an alerting rule having the same name with different
severity labels is being returned from different replicas then they
are being treated as separate alerts.

Given the following alerts a1,a2 with severities s1,s2 returned from
replicas r1,2:

a1[s1,r1]
a1[s2,r1]
a1[s1,r2]
a1[s2,r2]

Then, currently, the algorithm deduplicates to:

a1[s1]
a1[s2]
a1[s1]
a1[s2]

Instead of the intendet result:

a1[s1]
a1[s2]

## Changes

This fixes it by removing replica labels before sorting labels for
deduplication.

## Verification

New unit tests have been added which failed without the fix. Existing tests have been modified to consider the changed sorting order.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -20,6 +20,7 @@ We use _breaking :warning:_ to mark changes that are not backward compatible (re
 
 - [#3204](https://github.com/thanos-io/thanos/pull/3204) Mixin: Use sidecar's metric timestamp for healthcheck.
 - [#3922](https://github.com/thanos-io/thanos/pull/3922) Fix panic in http logging middleware.
+- [#3960](https://github.com/thanos-io/thanos/pull/3960) fix deduplication of equal alerts with different labels
 
 ### Changed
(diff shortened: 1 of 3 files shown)


s-urbaniak:
cc @kakkoyun @bwplotka ptal :pray:

s-urbaniak:
hmm, it seems the e2e test failures are unrelated :thinking: the timing out test `TestQueryExternalPrefixAndRoutePrefix` at least works here on my machine.

kakkoyun (approved):
LGTM. Thanks for the great explanations and tests. 🚀

bwplotka (commented):
Thanks!

s-urbaniak:
@bwplotka (cc @kakkoyun): I am wondering if this can land in 0.19? That way we wouldn't need a cherry-pick downstream :hugs:

Issue #3957, opened 2021-03-23
Refactoring of azure obj store


wiardvanrij:
I wanted to add some extra configuration options, but then kinda hit a wall with the current testing method. As it's quite hard to extend those. So I started refactoring the azure part and I think I can squash a few issues; 

#3933
#2839
#3952 

Also adding; 

- Implementing our own HTTP transport method (and therefore configuration options). Just like S3.
- Adding configuration for `PipelineRetry` method
- Adding configuration for `ReaderRetry` method
- Refactoring tests

As I'm quite far, I wanted to make this issue to check if no one else was on it. Also open for further feedback :) 
If possible add, me as assignee to this issue?

bwplotka:
Thanks!

I think no one is on it, and we saw some issues on GitHub e.g https://github.com/thanos-io/thanos/issues/3952 soi I think it's healthy to do!

wiardvanrij:
Updated the PR. 

Absolutely not tested yet, but the image is: `wiardvanrij/thanos:azure-test`  -> https://hub.docker.com/repository/docker/wiardvanrij/thanos

Going to test this myself later on, but would like more feedback from others..

airkewld:
`
level=error ts=2021-04-29T21:49:30.0974553Z caller=main.go:130 err="yaml: unmarshal errors:\n  line 3: field max_retries not found in type azure.Config\ncreate AZURE client\ngithub.com/thanos-io/thanos/pkg/objstore/client.NewBucket\n\t/go/src/github.com/thanos-io/thanos/pkg/objstore/client/factory.go:77\nmain.runCompact\n\t/go/src/github.com/thanos-io/thanos/cmd/thanos/compact.go:198\nmain.registerCompact.func1\n\t/go/src/github.com/thanos-io/thanos/cmd/thanos/compact.go:89\nmain.main\n\t/go/src/github.com/thanos-io/thanos/cmd/thanos/main.go:128\nruntime.main\n\t/usr/local/go/src/runtime/proc.go:225\nruntime.goexit\n\t/usr/local/go/src/runtime/asm_amd64.s:1371\npreparing compact command failed\nmain.main\n\t/go/src/github.com/thanos-io/thanos/cmd/thanos/main.go:130\nruntime.main\n\t/usr/local/go/src/runtime/proc.go:225\nruntime.goexit\n\t/usr/local/go/src/runtime/asm_amd64.s:1371"
`

Below are the contents of the obj secret
```
type: AZURE
config:
[...]

Issue #3952, opened 2021-03-23
Thanos Compact Crashing due to Azure Timeout


airkewld:
Compact component crashes when it encounters a file it cannot fully download/analyze.
```
level=info ts=2021-03-23T14:01:38.498832601Z caller=http.go:84 service=http/server component=compact msg="internal server is shutdown gracefully" err="syncing metas: incomplete view: meta.json file exists: 01F0DSQC0MQ07MBCFXSWD62NCG/meta.json: cannot get properties for Azure blob, address: 01F0DSQC0MQ07MBCFXSWD62NCG/meta.json: -> github.com/Azure/azure-pipeline-go/pipeline.NewError, /go/pkg/mod/github.com/!azure/azure-pipeline-go@v0.2.2/pipeline/error.go:154\nHTTP request failed\n\nHead \"https://obj_store_name.blob.core.windows.net/thanos/01F0DSQC0MQ07MBCFXSWD62NCG/meta.json?timeout=1\": context deadline exceeded\n"
```

Our obj store config includes a max_retries value of 3, but it does not seem to be passed in correctly.
```
/ # cat /proc/1/cmdline
/bin/thanoscompact--wait--log.level=debug--compact.concurrency=1--log.format=logfmt--objstore.config=type: AZURE
config:
  storage_account: "storage_account"
  storage_account_key: "storage_account_key"
  container: "container_name"
  endpoint: "blob.core.windows.net"
  max_retries: 3
--data-dir=/var/thanos/compact--debug.accept-malformed-index--retention.resolution-raw=30d--retention.resolution-5m=90d--retention.resolution-1h=180d--delete-delay=48h--deduplication.replica-label=prometheus_replica--deduplication.replica-label=rule_replica--tracing.config="config":
  "sampler_param": 2
  "sampler_type": "ratelimiting"
  "service_name": "thanos-compact"
"type": "JAEGER"/ #
```
Any assistance is appreciated.

wiardvanrij:
I'm pretty sure we need to add a timeout value to the retry pipeline part here; https://github.com/thanos-io/thanos/blob/90015469cc81e7532a995527aa903262988d0199/pkg/objstore/azure/helpers.go#L39

See: https://pkg.go.dev/github.com/Azure/azure-storage-blob-go/azblob#RetryOptions

```
// TryTimeout indicates the maximum time allowed for any single try of an HTTP request.
	// A value of zero means that you accept our default timeout. NOTE: When transferring large amounts
[...]

PR #3889, opened 2021-03-05
Cherry-pick from master: Fix parseStep bug introduced in PR #3740 (#3887)

bwplotka:
* Fix bug introduced in PR #3740

Signed-off-by: Hitanshu Mehta <hitanshu99amehta@gmail.com>

* Minor change in test

Signed-off-by: Hitanshu Mehta <hitanshu99amehta@gmail.com>

Cherry picked: https://github.com/thanos-io/thanos/pull/3887

PR #3970, opened 2021-03-24
Refactoring of azure obj store #3957 

wiardvanrij:

## Changes

- Adds configuration options for the Azure Pipeline method
- Adds configuration options for the Azure Reader method
- Adds configuration options for the HTTP transport method
- Adds HTTP Transport override
- Adds MSI authentication (#3889)
- Made parsing of the config strict
- ~~Breaks configuration by moving `max_retries` to it's own `reader_config`~~
- Rewrote tests 

## Verification

**How you tested it?** 
No, other than the tests included

Especially the MSI authentication part is untested. 

**How do you know it works?**
This is tested via other users on https://github.com/thanos-io/thanos/issues/3957 : 

@wiardvanrij  ran your image all night locally with no failures.

_Originally posted by @airkewld in https://github.com/thanos-io/thanos/issues/3957#issuecomment-830087240_

---------
I hope this PR can get tested by users of Azure to confirm the working of the changes AND if this has positive impact for on, for example #3952 - relates to #3957


wiardvanrij:
Need to check why the test fails on the cortex dep. :D

wiardvanrij:
We use this part; https://github.com/cortexproject/cortex/blob/master/pkg/storage/bucket/azure/bucket_client.go#L10 and Cortex uses the Thanos part. 
I think for this change both Cortex and Thanos need to be updated but I'm not sure on how to proceed with this. 

@pracucci  could you weight in here if my statement is even correct? Thank you!

bwplotka:
Cycling dependency problem, so:

* Fork Cortex, fix this on your fork, push remote branch to your fork
* Update go.mod to replace cortex dep with your fork URL and Commit SHA
* Let's review AND merge this PR
* Propose the change in the fork to Cortex 
* Remove replace.

pkg/objstore/azure/azure.go
```diff
@@ -24,13 +27,63 @@ const (
...
+	PipelineConfig: PipelineConfig{
+		MaxTries:      0,
+		TryTimeout:    0,
+		RetryDelay:    0,
+		MaxRetryDelay: 0,
+	},
+	ReaderConfig: ReaderConfig{
+		MaxRetryRequests: 0,
+	},
+	HTTPConfig: HTTPConfig{
```
bwplotka, line 41: Can we reuse this struct?

pkg/objstore/azure/helpers.go
```diff
@@ -64,6 +116,28 @@ func getContainerURL(ctx context.Context, conf Config) (blob.ContainerURL, error
 	return service.NewContainerURL(conf.ContainerName), nil
 }
 
+func DefaultTransport(config Config) *http.Transport {
```
bwplotka, line 119: That sounds like something we can reuse across buckets.

bwplotka (commented):
Some suggestions otherwise look good!

phillebaba:
Is there anything I can do to help move this along? :) Or is this dependent on changes that have to be made in Cortex?

wiardvanrij:
> Is there anything I can do to help move this along? :) Or is this dependent on changes that have to be made in Cortex?

It's currently not blocking but just needs some fixes in packages. Which eventually also requires some changes at the Cortex side. I did however not foresee the complexity at first :p 

But it's good to know that people need this, so I can make it a higher prio!

airkewld:
Glad to see this moving along. My compact pod is dying multiple times a day at this point, hopefully this will resolve the issue for me and others.  
Thanks

pkg/objstore/azure/helpers.go
```diff
@@ -64,6 +116,28 @@ func getContainerURL(ctx context.Context, conf Config) (blob.ContainerURL, error
...
+		MaxIdleConns:          config.HTTPConfig.MaxIdleConns,
+		MaxIdleConnsPerHost:   config.HTTPConfig.MaxIdleConnsPerHost,
+		IdleConnTimeout:       time.Duration(config.HTTPConfig.IdleConnTimeout),
+		MaxConnsPerHost:       config.HTTPConfig.MaxConnsPerHost,
+		TLSHandshakeTimeout:   time.Duration(config.HTTPConfig.TLSHandshakeTimeout),
+		ExpectContinueTimeout: time.Duration(config.HTTPConfig.ExpectContinueTimeout),
+
+		ResponseHeaderTimeout: time.Duration(config.HTTPConfig.ResponseHeaderTimeout),
+		DisableCompression:    config.HTTPConfig.DisableCompression,
+		TLSClientConfig:       &tls.Config{InsecureSkipVerify: config.HTTPConfig.InsecureSkipVerify},
```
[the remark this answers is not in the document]
wiardvanrij: ignore

wiardvanrij:
@airkewld  & @phillebaba see https://github.com/thanos-io/thanos/issues/3957 - I've build an image. But feel free to build it yourself. Wouldn't mind more testers on this feature :)

phillebaba:
I can probably test this as I am already running Thanos in AKS. Hopefully this will work with aad-pod-identity.

wiardvanrij:
> I can probably test this as I am already running Thanos in AKS. Hopefully this will work with aad-pod-identity.

Don't use it on production yet tho.. :D Also can we move potential discussions to the issue? Makes it a bit easier. Thanks so much btw :)

wiardvanrij:
> Cycling dependency problem, so:
> 
> * Fork Cortex, fix this on your fork, push remote branch to your fork
> * Update go.mod to replace cortex dep with your fork URL and Commit SHA
> * Let's review AND merge this PR
> * Propose the change in the fork to Cortex
> * Remove replace.

I changed the configuration in a way that it's not breaking anymore AND it has no dependency issue anymore. 
There is a technical trade-off here by implementing a configuration value that is somewhat obsolete. 

We now use the `max_retries` config value if no specific other 'retry related' configuration is set. This means there is 0 change for those we do not use the extra added configuration options. If one sets specific retry value configurations, the `max_retries` config value is ignored in favor of the more specific config values.  

So this resolved two things:
- No dependency issue anymore
- Backwards compatible

pkg/objstore/azure/helpers.go
wiardvanrij: I would like to propose to not do this in this PR but create a new issue for this. As this would mean a small refactor in various parts in Thanos. This makes the scope of this PR too big IMO. WDYT?

pkg/objstore/azure/azure.go
wiardvanrij: Yes, we should. See my other comments about not doing it in this PR, but create a new issue/PR for this

nicolastakashi:
Hi dudes!
any prevision to release it?

pkg/objstore/azure/azure.go
```diff
@@ -42,34 +95,77 @@ type Bucket struct {
...
 
+// parseConfig unmarshals a buffer into a Config with default values.
+func parseConfig(conf []byte) (Config, error) {
+	config := DefaultConfig
+	if err := yaml.UnmarshalStrict(conf, &config); err != nil {
+		return Config{}, err
+	}
+
+	// If we don't have config specific retry values but we do have the generic MaxRetries.
+	// This is part backwards compatibility but also ease of configuration
```
yashrsharma44, line 150: unless I have misunderstood, do you mean 
```suggestion
	// This part backwards compatibility but also ease of configuration
```
😅

yashrsharma44 (commented):
Small question about the comment 😛

pkg/objstore/azure/azure.go
bwplotka: We could, but why? (:

bwplotka (dismissed):
LGTM, 💪🏽

pkg/objstore/azure/azure.go
wiardvanrij: Yes, thanks, I will make this proper English :D

wiardvanrij:
I will update this PR as soon as this is merged: https://github.com/thanos-io/thanos/pull/4104#issuecomment-831352483
Then I can use that generic http method 💯

phillebaba:
@wiardvanrij had some time now to check this PR. The container is created without an issue but I am seeing authorization issues when uploading blocks from the receiver. I am using aad-pod-identity which adds additional complexities, but unless someone else has verified that this feature works I think you should hold off on merging.

wiardvanrij:
> @wiardvanrij had some time now to try this PR. The container is created without an issue but I am seeing authorization issues when uploading blocks from the receiver. I am using aad-pod-identity which adds additional complexities, but unless someone else has verified that this feature works I think you should hold off on merging.

Thanks! Could you perhaps provide more information in https://github.com/thanos-io/thanos/issues/3957 ?

docs/storage.md
```diff
@@ -332,8 +332,29 @@ config:
...
+    insecure_skip_verify: false
+    tls_handshake_timeout: 0s
+    expect_continue_timeout: 0s
+    max_idle_conns: 0
+    max_idle_conns_per_host: 0
+    max_conns_per_host: 0
+    disable_compression: false
 ```
 
+If `msi_resource` is used, authentication is done via ServicePrincipalToken. The value for Azure should be `https://storage.azure.com`.
```
phillebaba, line 355:
```suggestion
If `msi_resource` is used, authentication is done via ServicePrincipalToken. The value for Azure should be `https://<storage-account-name>.blob.core.windows.net`.
```

pkg/objstore/azure/azure_test.go
```diff
@@ -5,100 +5,178 @@ package azure
...
+pipeline_config:
+  "try_timeout": "10s"`),
+		wantFailParse:    false,
+		wantFailValidate: false,
+	},
+	{
+		name: "msi resource used with storage accounts",
+		config: []byte(`storage_account: "asdfasdf"
+storage_account_key: "asdfsdf"
+msi_resource: "https://storage.azure.com"
```
phillebaba, line 119:
```suggestion
msi_resource: "https://example.blob.core.windows.net"
```

pkg/objstore/azure/azure_test.go
```diff
@@ -5,100 +5,178 @@ package azure
...
+pipeline_config:
+  "try_timeout": "10s"`),
+		wantFailParse:    false,
+		wantFailValidate: true,
+	},
+	{
+		name: "Valid MSI Resource",
+		config: []byte(`storage_account: "myAccount"
+storage_account_key: ""
+msi_resource: "https://storage.azure.com"
```
phillebaba, line 133:
```suggestion
msi_resource: "https://example.blob.core.windows.net"
```

phillebaba (commented):
Just small suggestions to fix the msi resource value.

pkg/objstore/azure/azure.go
bwplotka: Cool let's add in https://github.com/thanos-io/thanos/issues/4104 or so

pkg/objstore/azure/helpers.go
```diff
@@ -30,20 +34,57 @@ func init() {
...
+
+		msiConfig := auth.NewMSIConfig()
+		msiConfig.Resource = conf.MSIResource
+
+		azureServicePrincipalToken, err := msiConfig.ServicePrincipalToken()
+		if err != nil {
+			return nil, err
+		}
+
+		// Get a new token
```
bwplotka, line 49:
```suggestion
		// Get a new token.
```

pkg/objstore/azure/helpers.go
```diff
@@ -30,20 +34,57 @@ func init() {
 	pipeline.SetForceLogEnabled(false)
 }
 
+func getAzureStorageCredentials(conf Config) (blob.Credential, error) {
+
+	if conf.MSIResource != "" {
+
```
bwplotka, line 40:
```suggestion
```

pkg/objstore/azure/helpers.go
```diff
@@ -30,20 +34,57 @@ func init() {
 	pipeline.SetForceLogEnabled(false)
 }
 
+func getAzureStorageCredentials(conf Config) (blob.Credential, error) {
+
```
bwplotka, line 38:
```suggestion
```

pkg/objstore/azure/helpers.go
bwplotka: no need for those new lines

pkg/objstore/azure/helpers.go
```diff
@@ -30,20 +34,57 @@ func init() {
...
+		// Get a new token
+		err = azureServicePrincipalToken.Refresh()
+		if err != nil {
+			return nil, err
+		}
+		token := azureServicePrincipalToken.Token()
+
+		return blob.NewTokenCredential(token.AccessToken, nil), nil
+
```
bwplotka, line 57:
```suggestion
```

pkg/objstore/azure/helpers.go
```diff
@@ -30,20 +34,57 @@ func init() {
...
+	} else {
+		credential, err := blob.NewSharedKeyCredential(conf.StorageAccountName, conf.StorageAccountKey)
+		if err != nil {
+			return nil, err
+		}
+		return credential, nil
+	}
+
```
bwplotka, line 65:
```suggestion
```

pkg/objstore/azure/helpers.go
```diff
@@ -30,20 +34,57 @@ func init() {
...
+		// Get a new token
+		err = azureServicePrincipalToken.Refresh()
+		if err != nil {
+			return nil, err
+		}
+		token := azureServicePrincipalToken.Token()
+
+		return blob.NewTokenCredential(token.AccessToken, nil), nil
+
+	} else {
```
bwplotka, line 58: No need for this else we return above anyway. See https://thanos.io/tip/contributing/coding-style-guide.md/#control-structure-prefer-early-returns-and-avoid-else

bwplotka (approved):
LGTM! There are minor style nits, otherwise I would be happy to merge this.

Great work and sorry for major lag on review!

pkg/objstore/azure/helpers.go
wiardvanrij: Removed the `else` so this newline is now legit.

go.mod
```diff
@@ -5,6 +5,7 @@ require (
 	cloud.google.com/go/storage v1.10.0
 	github.com/Azure/azure-pipeline-go v0.2.2
 	github.com/Azure/azure-storage-blob-go v0.8.0
+	github.com/Azure/go-autorest/autorest/azure/auth v0.5.7
```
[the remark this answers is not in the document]
wiardvanrij: https://github.com/Azure/go-autorest/pull/640

KingTropea999 (approved):

go.mod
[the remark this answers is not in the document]
phillebaba: @wiardvanrij v0.5.8 should now be out with an updated /x/crypto dependency.

docs/storage.md
```diff
@@ -332,8 +332,28 @@ config:
...
+    max_tries: 0
+    try_timeout: 0s
+    retry_delay: 0s
+    max_retry_delay: 0s
+  reader_config:
+    max_retry_requests: 0
+  http_config:
+    idle_conn_timeout: 0s
+    response_header_timeout: 0s
+    insecure_skip_verify: false
```
tomasfreund, line 346: Could [tls_config](https://prometheus.io/docs/prometheus/latest/configuration/configuration/#tls_config) be added instead of just `insecure_skip_verify`?
I know blob storage itself does not support it but we'd like to use a proxy (by changing the `endpoint` in the configuration) with mTLS for blob storage to avoid having to provide the storage account key to every cluster we monitor.
I also think it would be better to make the config as similar to Prometheus as possible for consistency and code resuse (if possible)

wiardvanrij: yes and no. This is in line with all other storage options in Thanos: https://thanos.io/tip/thanos/storage.md/
There is some work being done to make these configs more logical, but for now I try to keep it aligned with how everything in Thanos works (:

go.mod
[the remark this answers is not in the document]
wiardvanrij: I will look at this monday, as I'm on vacation. Feel free to make a PR against my branch tho :)

docs/storage.md
tomasfreund: thanks for the answer, I guess the solution to this problem is to implement a custom proxy and run it as a sidecar, using https with a dummy certificate and `insecure_skip_verify`

wiardvanrij: > thanks for the answer, I guess the solution to this problem is to implement a custom proxy and run it as a sidecar, using https with a dummy certificate and `insecure_skip_verify`

Yea I understand the case, it's just that I don't feel making one specific change here. However, it would make sense to implement a generic tls_config for non-component related configs, but for 'object storage connections'. As this would/should be implemented not only for Azure Blob, but also for all the other options Thanos supports. 
Perhaps you could create a new issue for this?

bwplotka (approved):
Amazing! Let's get this in and get some nice, sweet feedback.

Nice work @wiardvanrij , Mr Golang dev! (:

amrmahdi:
@wiardvanrij Thanks for adding these changes! 
Couple of questions:
1. The `msi_resource` why is it needed? Isn't it just a concat of storage account name and the endpoint?
2. I'm assuming this does not support user assigned managed identity, correct?

amrmahdi:
> @wiardvanrij Thanks for adding these changes!
> Couple of questions:
> 
> 1. The `msi_resource` why is it needed? Isn't it just a concat of storage account name and the endpoint?
> 2. I'm assuming this does not support user assigned managed identity, correct?

I think got my questions answered. I have a follow up PR to support user-assigned managed identity. https://github.com/thanos-io/thanos/pull/4636

wiardvanrij:
> > @wiardvanrij Thanks for adding these changes!
> > Couple of questions:
> > 
> > 1. The `msi_resource` why is it needed? Isn't it just a concat of storage account name and the endpoint?
> > 2. I'm assuming this does not support user assigned managed identity, correct?
> 
> I think got my questions answered. I have a follow up PR to support user-assigned managed identity. #4636

Wow thanks! Really good and thanks for helping out!

PR #3973, opened 2021-03-25
Fix IsObjNotFoundErr() for wrapped errors

pracucci:

## Changes

We have cases in Cortex where we may pass a wrapped error to `IsObjNotFoundErr()`. In this PR I'm proposing to check the error cause in `IsObjNotFoundErr()` so that it can work for wrapped errors too.

I checked all `IsObjNotFoundErr()` implementations and changed only the ones which weren't using `errors.Cause()` or `errors.Is()` yet.

## Verification

N/A


bwplotka (approved):
Perfect. We should never cast without such wrappers, good point. Thanks!

PR #3977, opened 2021-03-26
Add Exemplars for http_request_duration_seconds histogram

### LINKED ISSUE #3944: Instrument http_request_duration_seconds metric with exemplars
From the discussion https://github.com/thanos-io/thanos/discussions/3938, we think it is a good start to instrument https://github.com/thanos-io/thanos/blob/main/pkg/extprom/http/instrument_server.go#L44 with Exemplars.

Exemplar example: 
1. https://github.com/weaveworks/common/pull/202

Help is welcomed!

metalmatze:

## Changes

The Prometheus HTTP handler now enables OpenMetrics as response format upon content negotiation. 
You can test it by running `curl -s -H 'Accept: application/openmetrics-text' localhost:10904/metrics`
This allows us to pass a `traceID` to `ObserveWithExemplar`. 

A lot of the other changes are actually changing the order of calling the middlewares.
Going forward we need to make sure to first call the tracing middleware so that by the time our monitoring/instrumentation middleware gets hit the tracing spans are already populated.

## Verification

Tested locally with Jaeger and Prometheus running `--enable-feature=exemplar-storage`.

Screenshot from 2021-03-26 12-47-25

Closes #3944

### DIFF
--- a/pkg/extprom/http/instrument_server.go
+++ b/pkg/extprom/http/instrument_server.go
@@ -4,11 +4,16 @@
 package http
 
 import (
+	"fmt"
 	"net/http"
+	"strings"
+	"time"
 
+	"github.com/opentracing/opentracing-go"
 	"github.com/prometheus/client_golang/prometheus"
 	"github.com/prometheus/client_golang/prometheus/promauto"
 	"github.com/prometheus/client_golang/prometheus/promhttp"
+	"github.com/uber/jaeger-client-go"
 )
 
 // InstrumentationMiddleware holds necessary metrics to instrument an http.Server
@@ -21,9 +26,9 @@ type InstrumentationMiddleware interface {
 type nopInstrumentationMiddleware struct{}
 
 func (ins nopInstrumentationMiddleware) NewHandler(handlerName string, handler http.Handler) http.HandlerFunc {
-	return http.HandlerFunc(func(w http.ResponseWriter, r *http.Request) {
+	return func(w http.ResponseWriter, r *http.Request) {
 		handler.ServeHTTP(w, r)
-	})
+	}
 }
 
 // NewNopInstrumentationMiddleware provides a InstrumentationMiddleware which does nothing.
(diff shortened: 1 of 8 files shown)


pkg/extprom/http/instrument_server.go
```diff
@@ -90,17 +95,72 @@ func NewInstrumentationMiddleware(reg prometheus.Registerer, buckets []float64)
...
+					observer := ins.requestDuration.WithLabelValues(
+						wd.Status(),
+						handlerName,
+						strings.ToLower(r.Method),
+					)
+					observer.Observe(time.Since(now).Seconds())
+
+					// If we find a tracingID we'll expose it as Exemplar.
+					span := opentracing.SpanFromContext(r.Context())
+					if span != nil {
```
bwplotka, line 119: Good for now, but maybe we can create some way helpers for this (:

bwplotka (approved):
This is sooo good. Thanks @metalmatze !

pkg/extprom/http/instrument_server.go
metalmatze: Yes, that might be worth doing. 
I also looked into bringing parts upstream into client_golang but almost all of these modifications happen in the http.HandlerFunc which is too specific for client_golang it seemed like.

I can quickly try to refactor into a helper. :+1:

metalmatze: Now I have locally:

```go
// within the handler
durationObserver := ins.requestDuration.WithLabelValues(
	wd.Status(),
	handlerName,
	strings.ToLower(r.Method),
)
observeDuration(durationObserver, start, r)

// outside the handler as a helper func

func observeDuration(observer prometheus.Observer, start time.Time, r *http.Request) {
	observer.Observe(time.Since(start).Seconds())

	// If we find a tracingID we'll expose it as Exemplar.
	span := opentracing.SpanFromContext(r.Context())
	if span != nil {
		spanCtx, ok := span.Context().(jaeger.SpanContext)
		if ok {
			observer.(prometheus.ExemplarObserver).ObserveWithExemplar(
				time.Since(start).Seconds(),
				prometheus.Labels{
					"traceID": spanCtx.TraceID().String(),
				},
			)
		}
	}
}
```
WDYT?

bwplotka: Yup, that would work, but I think for one example is not worth refactoring 

Once we have 3+ places to use this then it's worth our work, we will also know more! Thanks!

pkg/extprom/http/instrument_server.go
```diff
@@ -90,17 +95,72 @@ func NewInstrumentationMiddleware(reg prometheus.Registerer, buckets []float64)
...
+						spanCtx, ok := span.Context().(jaeger.SpanContext)
```
yeya24, line 120: Here we just support Jaeger tracer? Is it compatible with `lightstep`, `stackdriver`, etc?

pkg/extprom/http/instrument_server.go
```diff
@@ -90,17 +95,72 @@ func NewInstrumentationMiddleware(reg prometheus.Registerer, buckets []float64)
...
+						if ok {
+							observer.(prometheus.ExemplarObserver).ObserveWithExemplar(
```
yeya24, line 122: Like in the weaveworks example https://github.com/weaveworks/common/pull/202/files#diff-29d7cb6219a01fc80b731ac8004ac0f5edd5a1be507b164939ba4a4d1caf9ac2R70, I think we also need to check for sampled trace ID.

bwplotka:
Let's iterate over it @yeya24 good points. Also I would see `GetTraceID` as some helper function tbh (:

PR #3993, opened 2021-03-30
Refactor flag parsing and runRule arguments for ruler

idoqo:
Signed-off-by: Michael Okoko <okokomichaels@outlook.com>

## Changes

- Refactor `runRule` arguments into a config struct which can then be passed to the function instead.
- Create new config structs for flags that are in the same "namespace" e.g `query.*` and `alertmanagers.*`. (As per #2248 and similar to #2267).

### DIFF
--- a/cmd/thanos/config.go
+++ b/cmd/thanos/config.go
@@ -9,6 +9,8 @@ import (
 	"net/url"
 	"time"
 
+	"github.com/thanos-io/thanos/pkg/extflag"
+
 	"github.com/prometheus/common/model"
 	"github.com/thanos-io/thanos/pkg/extkingpin"
 )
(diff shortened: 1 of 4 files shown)


cmd/thanos/config.go
```diff
@@ -161,3 +167,53 @@ func (wc *webConfig) registerFlag(cmd extkingpin.FlagClause) *webConfig {
 	cmd.Flag("web.disable-cors", "Whether to disable CORS headers to be set by Thanos. By default Thanos sets CORS headers to be allowed by all.").Default("false").BoolVar(&wc.disableCORS)
 	return wc
 }
+
+type queryConfig struct {
```
bwplotka, line 171: Could we maybe put that next to the Querier for visibility? (: 

or is this shared across components?

bwplotka (commented):
It looks generally, thanks! Will review more carefully later.

cmd/thanos/config.go
idoqo: Ohh yeah, this struct actually parses `--query.*` flags (e.g `query.sd-interval`, `query.auto-downsampling`) so It is used by ruler and can later be re-used for querier (and other components that need it).

Thinking of it now, it would be easy to mistake it as a config struct for querier (similar to `ruleConfig` and `sidecarConfig`). Maybe it needs to be renamed? (not sure what a more appropriate name could be). Or it stays this way and querier's config struct could be something like `querierConfig`.

bwplotka: Don't forget to run `make docs` at some point to fix CI issue (:

docs/components/compact.md
```diff
@@ -428,6 +428,10 @@ Flags:
                                 Prometheus relabel-config syntax. See format
                                 details:
                                 https://prometheus.io/docs/prometheus/latest/configuration/configuration/#relabel_config
+      --web.route-prefix=""     Prefix for API and UI endpoints. This allows
```
bwplotka, line 431: This is great, but let's make sure this flag is handled and not registered and not used, by accident (:

docs/components/rule.md
```diff
@@ -269,73 +269,16 @@ Flags:
                                  TLS CA to verify clients against. If no client
                                  CA is specified, there is no client
                                  verification on server side. (tls.NoClientCert)
-      --label=<name>="<value>" ...
```
bwplotka, line 272: Let's review manually here, but @idoqo can you add an issue on Thanos for the task to print flags in sorted order. (Let's make sure to provide the use case).

bwplotka (approved):
LGTM, just let's ensure all flags are used/not missed. (will do that later)

bwplotka (approved):
Nice! Now it's super easy to see what was added 🤗 Thanks to you @idoqo and @wiardvanrij for adding sorting!

PR #3994, opened 2021-03-30
Apply once when WatchInterval == 0 

zsuzhengdu:
## Changes

Fix https://github.com/prometheus-operator/prometheus-operator/issues/3061#issuecomment-619812179

## Verification

Manually tested. Working on unit tests.

### DIFF
--- a/pkg/reloader/reloader.go
+++ b/pkg/reloader/reloader.go
@@ -197,6 +197,11 @@ func (r *Reloader) Watch(ctx context.Context) error {
 		}
 	}
 
+	if r.watchInterval == 0 {
+		// Skip watching the file-system
+		return nil
+	}
+
 	for _, dir := range r.watchedDirs {
 		if err := r.watcher.addDirectory(dir); err != nil {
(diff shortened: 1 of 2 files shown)


bwplotka (commented):
Makes sense to me, thanks.

However, I have a small suggestion. What about doing one-off if `WatchInterval time.Duration is 0`? This makes sense logically and will allow us to have consistent API for reloader. 🤗 WDYT?

zsuzhengdu:
@bwplotka  Would you please help to give another review on this PR? Thanks!

bwplotka (dismissed):
Solid. Thanks!

pkg/reloader/reloader.go
```diff
@@ -197,6 +197,11 @@ func (r *Reloader) Watch(ctx context.Context) error {
 		}
 	}
 
+	if r.watchInterval == 0 {
+		// Skip watching the file-system
```
bwplotka, line 201:
```suggestion
		// Skip watching the file-system.
```

bwplotka (approved):

PR #3999, opened 2021-03-31
Cherry-pick of pkg/rules: fix deduplication of equal alerts with different labels (#3960)

bwplotka:
Currently, if an alerting rule having the same name with different
severity labels is being returned from different replicas then they
are being treated as separate alerts.

Given the following alerts a1,a2 with severities s1,s2 returned from
replicas r1,2:

a1[s1,r1]
a1[s2,r1]
a1[s1,r2]
a1[s2,r2]

Then, currently, the algorithm deduplicates to:

a1[s1]
a1[s2]
a1[s1]
a1[s2]

Instead of the intendet result:

a1[s1]
a1[s2]

This fixes it by removing replica labels before sorting labels for
deduplication.

# Conflicts:
#	CHANGELOG.md


kakkoyun (approved):
lgtm

PR #4007, opened 2021-04-01
Reload credentials from yaml logic for S3 [WIP]

Biswajitghosh98:
cc @kakkoyun @bwplotka 
(This PR is still in WIP and was made to be discussed in the weekly Mentor-Mentee meeting)

Update : 

- [x] Used a goroutine to watch the config file concurrently in  a non blocking way
- [x] Covered Race conditions
- [ ] Code cleanup (The code is ugly in terms of parameter passing. There's a variadic parameter which can be cleaned)

 TODO :
 - [ ] Add tests (Blocker : asserting the access key and secret key)

### DIFF
--- a/pkg/extflag/pathorcontent.go
+++ b/pkg/extflag/pathorcontent.go
@@ -74,10 +74,10 @@ func (p *PathOrContent) Content() ([]byte, error) {
 	return content, nil
 }
 
-func (p *PathOrContent) GetPath() (string, error) {
-	path := ""
+func (p *PathOrContent) GetPath() ([]string, error) {
+	var path []string
 	if len(*p.path) > 0 {
-		path = *p.path
+		path = append(path, *p.path)
 		return path, nil
 	}
 	return path, errors.Errorf("path is empty.")
--- a/pkg/extflag/pathorcontent.go
+++ b/pkg/extflag/pathorcontent.go
@@ -73,3 +73,12 @@ func (p *PathOrContent) Content() ([]byte, error) {
 
 	return content, nil
 }
+
+func (p *PathOrContent) GetPath() (string, error) {
+	path := ""
+	if len(*p.path) > 0 {
+		path = *p.path
+		return path, nil
+	}
+	return path, errors.Errorf("path is empty.")
+}
--- a/pkg/objstore/oss/oss.go
+++ b/pkg/objstore/oss/oss.go
@@ -46,8 +46,6 @@ type Bucket struct {
 	bucket *alioss.Bucket
 }
 
-func (b *Bucket) ReloadCredentials() error { return nil }
-
 func NewTestBucket(t testing.TB) (objstore.Bucket, func(), error) {
 	c := Config{
 		Endpoint:        os.Getenv("ALIYUNOSS_ENDPOINT"),
--- a/pkg/objstore/oss/oss.go
+++ b/pkg/objstore/oss/oss.go
@@ -46,6 +46,8 @@ type Bucket struct {
 	bucket *alioss.Bucket
 }
 
+func (b *Bucket) ReloadCredentials() error { return nil }
+
 func NewTestBucket(t testing.TB) (objstore.Bucket, func(), error) {
 	c := Config{
 		Endpoint:        os.Getenv("ALIYUNOSS_ENDPOINT"),
--- a/pkg/objstore/s3/s3.go
+++ b/pkg/objstore/s3/s3.go
@@ -159,7 +159,7 @@ type Bucket struct {
 	putUserMetadata map[string]string
 	partSize        uint64
 	listObjectsV1   bool
-	confFilepath    string
+	confFilepath    []string
 }
 
 // parseConfig unmarshals a buffer into a Config with default HTTPConfig values.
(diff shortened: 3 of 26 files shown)


pkg/objstore/s3/s3.go
```diff
@@ -307,6 +309,53 @@ func (b *Bucket) Name() string {
...
+			client, err := minio.New(config.Endpoint, &minio.Options{
+				Creds:     credentials.NewChainCredentials(chain),
+				Secure:    !config.Insecure,
+				Region:    config.Region,
+				Transport: rt,
+			})
+			if err != nil {
+				return errors.Wrap(err, "initialize s3 client.")
+			}
+			(*b).client = client
```
bwplotka, line 351: I think there is a race here on both Go and usage level. What if someone is reading a file from S3 and we change this?

pkg/objstore/oss/oss.go
```diff
@@ -46,6 +46,8 @@ type Bucket struct {
 	bucket *alioss.Bucket
 }
 
+func (b *Bucket) ReloadCredentials() error { return nil }
```
bwplotka, line 49: I would add this to S3 for now and use optional interface casting to enable reloading. WDYT? It will allow iterating in easy way (:

pkg/objstore/s3/s3.go
bwplotka: What if one process is calling Upload, and another is calling Iter?

pkg/objstore/s3/s3.go
```diff
@@ -307,6 +309,53 @@ func (b *Bucket) Name() string {
 	return b.name
 }
 
+// Reads the config file and changes the credentials of the instance if yaml is changed.
+func (b *Bucket) ReloadCredentials() error {
+	if len(b.confFilepath) > 0 {
+		c, err := ioutil.ReadFile(b.confFilepath)
```
bwplotka, line 315: It might be to costly to do it on every iteration. Is there any asynchronous way to do this?

* We can setup file watch (you can check reloader code)
* Even something simple like checking every 1m.
* Work channels to pass/change the client (????)

bwplotka (commented):
Nice, amazing start!

Some initial suggestions during our 1:1. 🤗 

+ Let's add test, maybe even let's add unit test, we can check if things does not work, we could check race detector too.

pkg/objstore/s3/s3.go
yashrsharma44, line 351:
```suggestion
			(b).client = client
```
More Info - https://tour.golang.org/moretypes/4

pkg/objstore/s3/s3.go
```diff
@@ -307,6 +309,53 @@ func (b *Bucket) Name() string {
...
+				Transport: rt,
+			})
+			if err != nil {
+				return errors.Wrap(err, "initialize s3 client.")
+			}
+			(*b).client = client
+			return nil
+		}
+	}
+	b.logger.Log("Config file path is empty, credentials are loaded from different source.")
```
yashrsharma44, line 355:
```suggestion
	level.Warn(b.logger).Log("Config file path is empty, credentials are loaded from a different source.")
```
We should use `level.Warn(logger)` other places as well :)

pkg/objstore/s3/s3.go
```diff
@@ -379,6 +429,7 @@ func (b *Bucket) Iter(ctx context.Context, dir string, f func(string) error, opt
 }
 
 func (b *Bucket) getRange(ctx context.Context, name string, off, length int64) (io.ReadCloser, error) {
+	b.ReloadCredentials()
```
yashrsharma44, line 432: Regarding this, are we calling this reloader function on each call? Won't this a little too expensive to have on each function call?

pkg/objstore/oss/oss.go
yashrsharma44: What is optional interface casting 😅 ? @bwplotka

pkg/extflag/pathorcontent.go
```diff
@@ -73,3 +73,12 @@ func (p *PathOrContent) Content() ([]byte, error) {
...
 }
+
+func (p *PathOrContent) GetPath() (string, error) {
+	path := ""
+	if len(*p.path) > 0 {
+		path = *p.path
+		return path, nil
+	}
+	return path, errors.Errorf("path is empty.")
+}
```
yashrsharma44, line 84:
```suggestion
func (p *PathOrContent) GetPath() (string, error) {
	if len(p.path) > 0 {
		return p.path, nil
	}
	return nil, errors.Errorf("path is empty.")
}
```

Biswajitghosh98:
@yashrsharma44 Sorry for the late reply, I was working on the local fork and didn't get a chance to look at the PR further. Thanks for the reviews @bwplotka and @yashrsharma44. I've comitted your suggested changes, and for the multiple function call thing, Bartek suggested to run a thread in parallel to see if the file has changed once every minute, and if it has, only then would I call the reload function, along with a mutex lock, to prevent race conditions

kakkoyun:
@Biswajitghosh98 You made Stalebot unhappy, give it a last push as we discussed :)

PR #4015, opened 2021-04-03
ui: Fix state sharing between windows

### LINKED ISSUE #4004: Query/UI: state is shared between windows
https://github.com/thanos-io/thanos/issues/2578 was only partially fixed, and the same problem still occurs when multiple browser windows are used.

Reproduction sequence:
1. Open two browser windows, which we'll call A and B
2. Type a query into window A and click execute
3. In window B, toggle the deduplication or partial query options
4. Click execute in window A
5. You are now looking at the results of running the query in window A with the options from window B. Nothing in window A will indicate this has occurred.

saswatamcode:

## Changes
Attempts to fix #4004. With this fix, new window graph panels don't start from `id` 0 but from the last recorded index in `localStorage`. This ensures there is no mismatch between query result and the options selected even with multiple windows open.
For e.g, browser tab A is opened and 3 graph panels are added then `localStorage` is,
Screenshot 2021-04-03 at 1 33 32 PM
Screenshot 2021-04-03 at 1 33 43 PM
Then browser tab B is opened and 2 graph panels are added, resulting in the following `localStorage`,
Screenshot 2021-04-03 at 1 34 32 PM
Screenshot 2021-04-03 at 1 34 40 PM

## Verification
Tested locally.

### DIFF
--- a/pkg/ui/static/js/graph.js
+++ b/pkg/ui/static/js/graph.js
@@ -33,17 +33,17 @@ Prometheus.Graph.stepValues = [
 ];
 
 // Get id of last panel and increment.
-let num = 0;
+let num = -1;
 Object.entries(localStorage).map(([key]) => {
-  if (key.includes("enable-dedup") && num < parseInt(key[key.length - 1], 10)) {
-    num = parseInt(key[key.length - 1], 10);
+  if (key.includes("enable-dedup") && num < Number(key[key.length - 1])) {
+    num = Number(key[key.length - 1]);
   }
 });
-Prometheus.Graph.numGraphs = num;
+Prometheus.Graph.numGraphs = num + 1;
 
 Prometheus.Graph.prototype.initialize = function() {
   var self = this;
-  self.id = ++Prometheus.Graph.numGraphs;
+  self.id = Prometheus.Graph.numGraphs++;
 
   // Set default options.
   self.options.id = self.id;
--- a/pkg/ui/static/js/graph.js
+++ b/pkg/ui/static/js/graph.js
@@ -32,11 +32,18 @@ Prometheus.Graph.stepValues = [
   "1w", "2w", "4w", "8w", "1y", "2y"
 ];
 
-Prometheus.Graph.numGraphs = 0;
+// Get id of last panel and increment.
+let num = 0;
+Object.entries(localStorage).map(([key]) => {
+  if (key.includes("enable-dedup") && num < parseInt(key[key.length - 1], 10)) {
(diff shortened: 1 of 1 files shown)


onprem:
We are very close to switching to the React-based UI by default, so it'd be great if you can fix this in the React UI as well. Also, it'd be nice to upstream this fix to the Prometheus UI as well.

saswatamcode:
Sure! Will add that in. 🙂

saswatamcode:
@onprem this issue doesn't seem to affect the React UI since there's no state sharing between the Panel components by default and there's no `localStorage` persistence layer for managing panel options. Will upstream this fix to Prometheus UI!

pkg/ui/static/js/graph.js
```diff
@@ -32,11 +32,18 @@ Prometheus.Graph.stepValues = [
   "1w", "2w", "4w", "8w", "1y", "2y"
 ];
 
-Prometheus.Graph.numGraphs = 0;
+// Get id of last panel and increment.
+let num = 0;
+Object.entries(localStorage).map(([key]) => {
+  if (key.includes("enable-dedup") && num < parseInt(key[key.length - 1], 10)) {
+    num = parseInt(key[key.length - 1], 10);
+  }
```
onprem, line 40:
```suggestion
  if (key.includes("enable-dedup") && num < Number(key[key.length - 1])) {
    num = Number(key[key.length - 1]);
  }
```

parseInt is usually not preferred as it can cause problems in some cases.

pkg/ui/static/js/graph.js
```diff
@@ -32,11 +32,18 @@ Prometheus.Graph.stepValues = [
...
+});
+Prometheus.Graph.numGraphs = num;
 
 Prometheus.Graph.prototype.initialize = function() {
   var self = this;
-  self.id = Prometheus.Graph.numGraphs++;
+  self.id = ++Prometheus.Graph.numGraphs;
```
onprem, line 46:
```suggestion
  self.id = Prometheus.Graph.numGraphs++;
```

I think it's better to keep this line same as before and instead change above implementation to
```js
let num = -1;
// code ....
Prometheus.Graph.numGraphs = num+1;
```

If we don't do this, the first panel will start from `1` instead of `0` which was the case before. IMO it's preferable to keep the old behaviour if possible.

onprem (commented):
Just some minor nits but looks good otherwise. Thanks!

onprem (approved):
LGTM!

PR #4018, opened 2021-04-04
Tracing: change baggage key to be lower kebab-case for Jaeger compatability

### LINKED ISSUE #3951: Jaeger ForceTracing not work
thanos use `X-Thanos-Force-Tracing` as `JaegerDebugHeader` key [https://github.com/thanos-io/thanos/blob/90015469cc81e7532a995527aa903262988d0199/pkg/tracing/jaeger/jaeger.go#L55](https://github.com/thanos-io/thanos/blob/90015469cc81e7532a995527aa903262988d0199/pkg/tracing/jaeger/jaeger.go#L55), but jaeger client always convert request header key to lowercase before compare, [https://github.com/jaegertracing/jaeger-client-go/blob/v2.25.0/propagation.go#L139-L147](https://github.com/jaegertracing/jaeger-client-go/blob/v2.25.0/propagation.go#L139-L147), so the debug header key is never equal to `X-Thanos-Force-Tracing`.

conorevans:
Signed-off-by: Conor Evans <coevans@tcd.ie>

Closes #3951 

## Changes

Make baggage key lowercase for Jaeger support. Per [Jaeger documentation](https://www.jaegertracing.io/docs/1.22/client-libraries/#baggage):

> Limitation: since HTTP headers don’t preserve the case, Jaeger recommends baggage keys to be lowercase-kebab-case, e.g. my-baggage-key-1.

## Verification

Since this is a one-line const change, no tests have been added. I also did not feel terribly inclined to spin up a local Thanos+Jaeger set-up, though I can do so for completeness if desired.

### DIFF
--- a/pkg/tracing/tracing.go
+++ b/pkg/tracing/tracing.go
@@ -10,7 +10,9 @@ import (
 )
 
 // ForceTracingBaggageKey - force sampling header.
-const ForceTracingBaggageKey = "X-Thanos-Force-Tracing"
+// Lower kebab-case for Jaeger support, per documentation:
+// https://www.jaegertracing.io/docs/1.22/client-libraries/#baggage
+const ForceTracingBaggageKey = "x-thanos-force-tracing"
 
 // traceIdResponseHeader - Trace ID response header.
 const traceIDResponseHeader = "X-Thanos-Trace-Id"
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -30,10 +30,10 @@ We use _breaking :warning:_ to mark changes that are not backward compatible (re
 
 ### Changed
 - [#3929](https://github.com/thanos-io/thanos/pull/3929) Store: Adds the name of the instantiated memcached client to log info
-
 - [#3948](https://github.com/thanos-io/thanos/pull/3948) Receiver: Adjust `http_request_duration_seconds` buckets for low latency requests.
 - [#3856](https://github.com/thanos-io/thanos/pull/3856) Mixin: _breaking :warning:_ Introduce flexible multi-cluster/namespace mode for alerts and dashboards. Removes jobPrefix config option. Removes `namespace` by default.
 - [#3937](https://github.com/thanos-io/thanos/pull/3937) Store: Reduce memory usage for range queries.
+- [#4018](https://github.com/thanos-io/thanos/pull/4018) Tracing: Update "force tracing" baggage header to be lower kebab-case for Jaeger support.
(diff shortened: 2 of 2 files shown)


conorevans:
Errors from CI:

```
Error:     rule_test.go:238: TODO: Allow HTTP ports from binaries running on host to be accessible.
--- SKIP: TestRule_AlertmanagerHTTPClient (0.00s)
=== RUN   TestRule
Error:     rule_test.go:324: Flaky test. Fix it. See: https://github.com/thanos-io/thanos/issues/3425.
--- SKIP: TestRule (0.00s)
=== RUN   TestRulePartialResponse
Error:     rule_test.go:561: TODO: Allow HTTP ports from binaries running on host to be accessible.
...
Error:     targets_api_test.go:74: targets_api_test.go:113:
        
         unexpected error: rpc error: code = Internal desc = perform GET request against http://127.0.0.1:49186/api/v1/targets: Get "http://127.0.0.1:49186/api/v1/targets": context deadline exceeded
```

Unrelated

pkg/tracing/tracing.go
```diff
@@ -10,7 +10,9 @@ import (
 )
 
 // ForceTracingBaggageKey - force sampling header.
-const ForceTracingBaggageKey = "X-Thanos-Force-Tracing"
+// Lower kebab-case for Jaeger support, per documentation:
+// https://www.jaegertracing.io/docs/1.22/client-libraries/#baggage
+const ForceTracingBaggageKey = "x-thanos-force-tracing"
```
bwplotka, line 15: I wonder.. can we maybe support both? It will not introduce breaking change for user too. Just let's lower case the input before checking?

conorevans: I considered that, but I wasn't sure I saw the benefit - now I think further, backwards compatibility is preferable, for sure.

I'm not 100% sure I'm envisioning it right, so I'll just check:

```golang
pkg/tracing/http.go
		// If client specified ForceTracingBaggageKey header, ensure span includes it to force tracing.
		span.SetBaggageItem(ForceTracingBaggageKey, r.Header.Get(ForceTracingBaggageKey))
```

There we'd set to `span.SetBaggageItem(strings.ToLower(ForceTracingBaggageKey),...`

Then set the` JaegerDebugHeader` in to `ToLower` as well.

The only other place I can see that is a `forceRecorder` which itself is used for a basic tracer wrapping a gcloud tracer (stackdriver), so I don't see it necessary there?

PR #4019, opened 2021-04-04
Query: range query histogram

thorfour:
Adds a range query histogram that tracks the distribution
of range queries. This can be useful for determining storage
retention requirements.

## Changes

- Adds a range query histogram to the query component.

### DIFF
--- a/pkg/api/query/v1.go
+++ b/pkg/api/query/v1.go
@@ -32,6 +32,8 @@ import (
 	"github.com/go-kit/kit/log"
 	"github.com/opentracing/opentracing-go"
 	"github.com/pkg/errors"
+	"github.com/prometheus/client_golang/prometheus"
+	"github.com/prometheus/client_golang/prometheus/promauto"
 	"github.com/prometheus/common/model"
 	"github.com/prometheus/common/route"
 	"github.com/prometheus/prometheus/pkg/labels"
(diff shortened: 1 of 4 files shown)


pkg/api/query/v1.go
```diff
@@ -430,6 +441,9 @@ func (qapi *QueryAPI) queryRange(r *http.Request) (interface{}, []error, *api.Ap
 
 	qe := qapi.queryEngine(maxSourceResolution)
 
+	// Record the query range requested
```
GiedriusS, line 444: Minor nit: let's end comments with a full stop:
```suggestion
	// Record the query range requested.
```

GiedriusS (approved):

pkg/api/query/v1.go
```diff
@@ -142,6 +147,12 @@ func NewQueryAPI(
 		defaultInstantQueryMaxSourceResolution: defaultInstantQueryMaxSourceResolution,
 		defaultMetadataTimeRange:               defaultMetadataTimeRange,
 		disableCORS:                            disableCORS,
+
+		queryRangeHist: promauto.With(reg).NewHistogram(prometheus.HistogramOpts{
+			Name:    "query_range_duration_seconds",
```
brancz, line 152:
```suggestion
			Name:    "thanos_query_range_requested_timespan_duration_seconds",
```

To be consistent with the rest of the Thanos metrics this should have the `thanos_` prefix, I would also suggest we make this metric slightly more descriptive.

brancz (commented):
Just a small nit on naming, otherwise this is great!

pkg/api/query/v1.go
thorfour: Done.

brancz (approved):

PR #4029, opened 2021-04-07
Mixin: Remove dependency on the rule dashboard from compact dashboard

goober:

## Changes

This change will remove the dependency between the rule dashboard and the compact dashboard mixins to be able to only generate the compact dashboard without defining the rule dashboard.

## Verification

The compact dashboard has been generated and imported to Grafana with and without the included change. No visible differences found.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -12,6 +12,7 @@ We use _breaking :warning:_ to mark changes that are not backward compatible (re
 
 ## Unreleased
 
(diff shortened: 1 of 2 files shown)


wiardvanrij (approved):
Nice catch! 👍

bwplotka (approved):
Thanks 👍🏽 

Thanks @wiardvanrij for review (:

PR #3845, opened 2021-02-26
receiver: Split - Upstream the changes from the PR #3580

yashrsharma44:
Rebased #3580 with the `main` branch.

> Notes:
> * There were few issues with the dynamic type checking for error, I am getting the following error - 
> `errs, ok := unwrappedErr.(errutil.MultiError)`
> `unwrappedErr (variable of type error) cannot have dynamic type errutil.MultiError (missing method Error)`

Fixed it by creating another method for multierror - `Error()`. Let me know if the above makes sense.
Would love some reviews on this.
cc @brancz

PR #3580, opened 2020-12-13
[WIP] receiver: Split

Chans321:
Signed-off-by: Chans321 <tsschand@gmail.com>

* [ ] I added CHANGELOG entry for this change.
* [ ] Change is not relevant to the end user.

PR #4036, opened 2021-04-09
Proposal: Modification of Receive Split proposal.

bwplotka:
👋🏽 @Chans321 @brancz @GiedriusS @kakkoyun @squat @yashrsharma44 @metalmatze 

Sorry for not paying attention to the past proposal, but when we revisited and tried to review  https://github.com/thanos-io/thanos/pull/3845 and https://github.com/thanos-io/thanos/pull/3580  with @kakkoyun we found the potentially easier idea that might be more cleaner for users and maintainers. Feel free to give feedback!

First of all, thanks for all of your hard work on this, and experimental implementations in PR 🤗 We only want to propose different configuration really, but the idea stays exactly the same.

Some learnings for the future:

1. Let's not merge things prematurely. The previous proposal was missing some important parts:
   * Goals were stating implementation goals, not the usage goals. We should focus on usage goals to avoid focusing too much on implementation and not seeing other options.
   * Proposal was missing Alternatives.
   * Motivation was motivating the implementation. I think it should rather motivate GOALS, not implementation yet.
   * Motivation was not clear. Statements like `Right now the receiver handles ingestion, routing and writing, thus leading to too many responsibilities in a single process.` might need some proof.
   * There were some left overs like `-Additionally, we touch possibility for adding a consistent hashing mechanism and buffering the incoming requests.` What that means?
   * The related ticket was for something else - for different database reconfiguration @squat proposed. Given that there was not enough context for this proposal.
   * Lack of migration guide for such breaking change of already stable receive.

2. Implementation PR led to confusions. 
  * It was too large, ideally we can split work into smaller pieces. It was hard to review properly, agree on all items, not miss important piece etc.
  * It was marked as almost ready, which led @yashrsharma44 to spend time on rebasing and cleaning all of it. Instead, it was missing many elements like documentation, migration docs, mixings, dashboards, alerts that should be at least well planned ahead. If we would merge any of those two in this state it could led to even bigger confusion.

### DIFF
--- a/docs/proposals/202012_receive_split.md
+++ b/docs/proposals/202012_receive_split.md
@@ -1,5 +1,5 @@
 ---
-title: Thanos Remote Write
+title: Thanos Routing Receive and Ingesting Receive
 type: proposal
 menu: proposals
 status: accepted
@@ -8,30 +8,74 @@ status: accepted
 ### Related Tickets
 
 * Hashring improvements: https://github.com/thanos-io/thanos/issues/3141
+* Previous proposal implementations: https://github.com/thanos-io/thanos/pull/3845, https://github.com/thanos-io/thanos/pull/3580
+* Distribution idea: https://github.com/thanos-io/thanos/pull/2675
 
 ### Summary
 
-This document describes the motivation and design of splitting the receive component into receive-route and receive.
-Additionally, we touch possibility for adding a consistent hashing mechanism and buffering the incoming requests.
+This document describes the motivation and design of running Receiver in a stateless mode that does not have capabilities to store samples, it only routes remote write
+to further Receivers based on hashring. This allows setting optional deployment model were only Routing Receivers are using hashring files and does the routing and replication. That allows ingesting Receivers to not handle any routing or hashring, only receiving multi tenant writes.
 
 ### Motivation
 
-1. Splitting functionality can be used to make resharding events easier.
-2. Right now the receiver handles ingestion, routing and writing, thus leading to too many responsibilities in a single process.
-This also makes the receiver more difficult to operate and understand for Thanos users.
-Splitting the receiver component into two different components could potentially have the following benefits:
-    1. Resharding events become faster and cause no downtime in ingestion.
-    2. Deployment becomes easier to understand for Thanos users.
-    3. Each component consists of less code.
-    4. The new architecture enables further performance improvements.
+[@squat](https://github.com/squat):
+
+> Currently, any change to the hashring configuration file will trigger all Thanos Receive nodes to flush their multi-TSDBs, causing them to enter an unready state until the flush is complete. This unavailability during a flush allows for a clear state transition, however it can result in downtimes on the order of five minutes for every configuration change. Moreover, during configuration changes, the hashring goes through an even longer period of partial unreadiness, where some nodes begin and finish flushing before and after others. During this partial unreadiness, the hashring can expect high internal request failure rates, which cause clients to retry their requests, resulting in even higher load. Therefore, when the hashring configuration is changed due to automatic horizontal scaling of a set of Thanos Receivers, the system can expect higher than normal resource utilization, which can create a positive feedback loop that continuously scales the hashring.
+
 
 ### Goals
 
-1. Split the receive component into receive-route and receive (and ensure ease of resharding events).
-2. Evaluate any effects on performance by simulating scenarios and collecting and analyzing metrics.
-3. Use consistent hashing to avoid reshuffling time series after resharding events. The exact consistent hashing mechanism to be used needs some further research.
+* Reduce downtime of the ingestion logic in Thanos Receiver
 
-### Drawbacks of the project
-There is no possible way to have a single-process receiver. The user must have a router + a receiver running.
-##### Solution
-We document how the new architecture can be set up to have the same general deployment of the old architecture. (We run router and receiver on the same node).
+### Proposal
+
+We propose allowing to run Thanos Receiver in a mode that only forwards/replicates remote write (distributor mode). You can enable that mode by simply not specifying:
+
+```yaml
+  --receive.local-endpoint=RECEIVE.LOCAL-ENDPOINT
+                                 Endpoint of local receive node. Used to
+                                 identify the local node in the hashring
+                                 configuration.
+```
+
+We can call this mode a "Routing Receiver". Similarly, we can skip specify any hashring to Thanos Receiver (`--receive.hashrings-file=<path>`), explicitly purposing it only for ingesting. We can call this mode "Ingesting Receiver".
+
+User can also mix all of those two modes for various federated hashrings etc. So instead of what we had before:
+
+![Before](https://docs.google.com/drawings/d/e/2PACX-1vTfko27YB_3ab7ZL8ODNG5uCcrpqKxhmqaz3lW-yhGN3_oNxkTrqXmwwlcZjaWf3cGgAJIM4CMwwkEV/pub?w=960&h=720)
+
+We have:
+
+![After](https://docs.google.com/drawings/d/e/2PACX-1vTVrtCGjR4iMbrU7Kj6QAn1a1m4fr-kvoQVDAK4lzQ_wWfXfpLLEE9HB948-WHI5ZG6s1iGWt51R593/pub?w=960&h=720)
+
+This allows us to (optionally) model deployment in a way that avoid expensive re-configuration of the stateful ingesting Receivers after the hashring configuration file has changed.
+
+In comparison to previous proposal (as mentioned in [alternatives](#previous-proposal-separate-receive-route-command) we have big adventages:
+
+1. We can reduce number of components in Thanos system, we can reuse similar component flags and documentation. Users has to learn about one less command and in result Thanos design is much more approachable. Less components mean less maintainance, code and other implicit duties: Separate changelogs, issue confusions, boilerplates, etc.
+2. Allow consistent pattern with Query. We don't have separate StoreAPI component for proxying, we have that baked into Querier. This has been proven to be flexible and understandable, so I would like to propose similar pattern in Receiver.
+3. This is more future proof for potential advanced cases like chain of routers -> receivers -> routers -> receivers for federated writes, so trees with depth n.
+
+### Plan
+
+* Receiver without `--receive.hashrings` does not forward or replicate requests, it routes straight to multi-tsdb.
+* Receiver without ` --Receiver.local-endpoint` will assume that no storage is needed, so will skip creating any resources for multi TSDB.
+* Add changes to the documentation (it's simplistic now). Mention two modes.
+
+### Alternative Solutions
+
+#### Previous Proposal: Separate receive-route command
+
+1. Split the Receiver component into receive-route and Receiver (and ensure ease of resharding events).
+1. Evaluate any effects on performance by simulating scenarios and collecting and analyzing metrics.
+1. Use consistent hashing to avoid reshuffling time series after resharding events. The exact consistent hashing mechanism to be used needs some further research.
+1. Migration: We document how the new architecture can be set up to have the same general deployment of the old architecture. (We run router and Receiver on the same node).
+
+This potentially makes the receiver more difficult to operate and understand for Thanos users. I would argue this is howerver much more harder in overall Thanos deployment. Otherwise this option is exactly the same.
+
+#### Flag for current Receiver --receive-route
+
+Idea would be similar same as in [Proposal](#Proposal), but there will be explicit flag to turn off local storage capabilities.
+
+I think we can have much more understandable logic if we simply not configure hashring for ingesting Receivers and not configure local hashring endpoint to notify that such Receiver instance will never store anything.
+
--- a/docs/proposals/202012_receive_split.md
+++ b/docs/proposals/202012_receive_split.md
@@ -71,11 +71,10 @@ In comparison to previous proposal (as mentioned in [alternatives](#previous-pro
 1. Use consistent hashing to avoid reshuffling time series after resharding events. The exact consistent hashing mechanism to be used needs some further research.
 1. Migration: We document how the new architecture can be set up to have the same general deployment of the old architecture. (We run router and Receiver on the same node).
 
-This potentially makes the receiver more difficult to operate and understand for Thanos users. I would argue this is howerver much more harder in overall Thanos deployment. Otherwise this option is exactly the same.
+This potentially makes the receiver more difficult to operate and understand for Thanos users. I would argue this is however much harder in overall Thanos deployment. Otherwise, this option is exactly the same.
 
 #### Flag for current Receiver --receive-route
 
 Idea would be similar same as in [Proposal](#Proposal), but there will be explicit flag to turn off local storage capabilities.
 
 I think we can have much more understandable logic if we simply not configure hashring for ingesting Receivers and not configure local hashring endpoint to notify that such Receiver instance will never store anything.
-
--- a/docs/proposals/202012_receive_split.md
+++ b/docs/proposals/202012_receive_split.md
@@ -38,7 +38,7 @@ We propose allowing to run Thanos Receiver in a mode that only forwards/replicat
                                  configuration.
 ```
 
-We can call this mode a "Routing Receiver". Similarly, we can skip specify any hashring to Thanos Receiver (`--receive.hashrings-file=<path>`), explicitly purposing it only for ingesting. We can call this mode "Ingesting Receiver".
+We can call this mode a "Routing Receiver". Similarly, we can skip specifying any hashring to Thanos Receiver (`--receive.hashrings-file=<path>`), explicitly purposing it only for ingesting. We can call this mode "Ingesting Receiver".
 
 User can also mix all of those two modes for various federated hashrings etc. So instead of what we had before:
 


docs/proposals/202012_receive_split.md
```diff
@@ -8,30 +8,94 @@ status: accepted
...
+   This also makes the receiver more difficult to operate and understand for Thanos users.
+   Splitting the receiver component into two different components could potentially have the following benefits:
+   1. Resharding events become faster and cause no downtime in ingestion.
+   2. Deployment becomes easier to understand for Thanos users.
+   3. Each component consists of less code.
+
+*Cons:*
 
-### Drawbacks of the project
 There is no possible way to have a single-process receiver. The user must have a router + a receiver running.
```
brancz, line 94: If this is the only con, then there are simple solutions to solve this.

brancz:
I don't see sufficient reason here to change the direction of this project entirely when it's so close to the finish line, and two maintainers having signed off on the previous architecture.

docs/proposals/202012_receive_split.md
```diff
@@ -8,30 +8,94 @@ status: accepted
...
+[@squat](https://github.com/squat):
+
+> Currently, any change to the hashring configuration file will trigger all Thanos Receive nodes to flush their multi-TSDBs, causing them to enter an unready state until the flush is complete. This unavailability during a flush allows for a clear state transition, however it can result in downtimes on the order of five minutes for every configuration change. Moreover, during configuration changes, the hashring goes through an even longer period of partial unreadiness, where some nodes begin and finish flushing before and after others. During this partial unreadiness, the hashring can expect high internal request failure rates, which cause clients to retry their requests, resulting in even higher load. Therefore, when the hashring configuration is changed due to automatic horizontal scaling of a set of Thanos Receivers, the system can expect higher than normal resource utilization, which can create a positive feedback loop that continuously scales the hashring.
+
 
 ### Goals
 
+* Be able to run component that just does remote write forwarding part of Receive and can be part of hashring.
```
bwplotka, line 28: I think this is still suggesting the implementation.

```
* Reduce the downtime of ingestion part of Thanos
```

docs/proposals/202012_receive_split.md
```diff
@@ -8,30 +8,94 @@ status: accepted
...
+
+In comparison to previous proposal (as mentioned in [alternatives](#previous-proposal-separate-receive-route-command) we have big adventages:
+
+1. We can reduce number of components in Thanos system, we can reuse similar component flags and documentation. Users has to learn about one less command and in result Thanos design is much more approachable. Less components mean less maintainance, code and other implicit duties: Separate changelogs, issue confusions, boilerplates, etc.
+2. Allow consistent pattern with Query. We don't have separate StorAPI component for proxying, we have that baked into Querier. This has been proven to be flexible and understandable, so I would like to propose similar pattern in Receive.
+3. This is more future proof for potential advanced cases like chain of routers -> receives -> routers -> receives for federated writes.
+
+### Open Questions
+
+* Buffering the incoming requests.
```
bwplotka, line 63: I just inherited this from the previous proposal. Any ideas what was that about?

docs/proposals/202012_receive_split.md
```diff
@@ -8,30 +8,94 @@ status: accepted
...
+We have:
+
+![After](https://docs.google.com/drawings/d/e/2PACX-1vTVrtCGjR4iMbrU7Kj6QAn1a1m4fr-kvoQVDAK4lzQ_wWfXfpLLEE9HB948-WHI5ZG6s1iGWt51R593/pub?w=960&h=720)
+
+This allows us to (optionally) model deployment in a way that avoid expensive re-configuration of the stateful ingesting Receives after the hashring configuration file has changed.
+
+In comparison to previous proposal (as mentioned in [alternatives](#previous-proposal-separate-receive-route-command) we have big adventages:
+
+1. We can reduce number of components in Thanos system, we can reuse similar component flags and documentation. Users has to learn about one less command and in result Thanos design is much more approachable. Less components mean less maintainance, code and other implicit duties: Separate changelogs, issue confusions, boilerplates, etc.
+2. Allow consistent pattern with Query. We don't have separate StorAPI component for proxying, we have that baked into Querier. This has been proven to be flexible and understandable, so I would like to propose similar pattern in Receive.
```
kakkoyun, line 58: 👍

docs/proposals/202012_receive_split.md
kakkoyun: cc @Chans321 @squat @brancz

kakkoyun (dismissed):
Looks great. With this new proposed implementation, it draws more clear migration path and it's backwards compatible. It would much easier to roll these out.

docs/proposals/202012_receive_split.md
bwplotka: This is left overall, sorry, will be more explicit

bwplotka: I clarified, could you double check?

bwplotka:
> I don't see sufficient reason here to change the direction of this project entirely when it's so close to the finish line, and two maintainers having signed off on the previous architecture. Also there is a docs PR out already.

I understand your @brancz  feelings and it's entirely our bad we did not participate in this discussion, however. It's my fault. I would still love if you could take a look on 3 aspects:

1. > change the direction of this project entirely 

Is it entirely changing? From user there is no much different. Indeed it changing a lot of implementation wise - it avoids tons of work which is good?

2. > when it's so close to the finish line

I would argue with this. I don't think it's close for proper usage by users. Also it's creating breaking changes which violates the stability guarantee we made with receive. Of course we are not in 1.0 but I would expect some migration plan for receive useage. I mentioned reasoning in https://github.com/thanos-io/thanos/pull/4036#issue-612348300 point 2.

This proposal here allows to avoid tricky migration.

3. > , and two maintainers having signed off on the previous architecture

What's the difference if 1, two or hundred of maintainers signed off if the alternative we propose here was not considered (at least in the view of previously written proposal)?

4. I explained the rationales in Proposal. The need of single-process receive and route you commented on is arguable and really low priority issue. The main issue is the user confusion of architectural design and each components purpose, boileplate of another command, inconsistency with Querier pattern. I would love to at least look and consider In my opinion easier solution.

kakkoyun:
@brancz @bwplotka I also want to see this push through the finish line as quickly as possible. If could just tweak the proposal and modify the existing implementation, it'd be faster to merge this. We can even make this happen for v0.20.0 by pushing the cut of RC for a week (we just released a major version recently anyways). I'm happy to help with the effort. 

Current implementation changes the interface of the existing commands drastically, there is no easy migration path using current implementation. Maybe we should acknowledge the proposed design and implementation drift apart. 

FWIW, either we can proposal or implementation, I want to have these changes on the main branch. Initial benchmarks by @metalmatze proved that there are nice gains in here.

docs/proposals/202012_receive_split.md
```diff
@@ -8,30 +8,79 @@ status: accepted
...
+3. This is more future proof for potential advanced cases like chain of routers -> receives -> routers -> receives for federated writes.
```
bwplotka, line 58:
```suggestion
3. This is more future proof for potential advanced cases like chain of routers -> receives -> routers -> receives for federated writes, so trees with depth n.
```

docs/proposals/202012_receive_split.md
```diff
@@ -8,30 +8,79 @@ status: accepted
...
 * Hashring improvements: https://github.com/thanos-io/thanos/issues/3141
+* Previous proposal implementations: https://github.com/thanos-io/thanos/pull/3845, https://github.com/thanos-io/thanos/pull/3580
+* Distribution idea: https://github.com/thanos-io/thanos/pull/2675
 
 ### Summary
 
-This document describes the motivation and design of splitting the receive component into receive-route and receive.
-Additionally, we touch possibility for adding a consistent hashing mechanism and buffering the incoming requests.
+This document describes the motivation and design of running Receive in a stateless mode that does not have capabilities to store samples, it only routes remote write
+to further receives based on hashring. This allows setting optional deployment model were only Routing Receives are using hashring files and does the routing and replication. That allows ingesting Receivers to not handle any routing or hashring, only receiving multi tenant writes.
```
yashrsharma44, line 17: Should it be `receives` or `receiver`? I am getting confused :(
```suggestion
to another receiver based on hashring. This allows setting optional deployment model were only Routing Receives are using hashring files and does the routing and replication. That allows ingesting Receivers to not handle any routing or hashring, only receiving multi tenant writes.
```

yashrsharma44: Because out here in this line - `That allows ingesting Receivers to not handle any routing or hashring, only receiving multi tenant writes.`, we use `receiver` :)

docs/proposals/202012_receive_split.md
```diff
@@ -8,30 +8,79 @@ status: accepted
...
+2. Allow consistent pattern with Query. We don't have separate StorAPI component for proxying, we have that baked into Querier. This has been proven to be flexible and understandable, so I would like to propose similar pattern in Receive.
+3. This is more future proof for potential advanced cases like chain of routers -> receives -> routers -> receives for federated writes.
+
+### Open Questions
+
+* Buffering the incoming requests.
+
+### Plan
+
+* Receive without `--receive.hashrings` does not forward or replicate requests, it routes straight to multi-tsdb.
```
yashrsharma44, line 66: So by **_not using a flag_** we enforce a certain behaviour, shouldn't we use a flag to enforce what we want?

docs/proposals/202012_receive_split.md
```diff
@@ -8,30 +8,79 @@ status: accepted
...
-    3. Each component consists of less code.
-    4. The new architecture enables further performance improvements.
+[@squat](https://github.com/squat):
+
+> Currently, any change to the hashring configuration file will trigger all Thanos Receive nodes to flush their multi-TSDBs, causing them to enter an unready state until the flush is complete. This unavailability during a flush allows for a clear state transition, however it can result in downtimes on the order of five minutes for every configuration change. Moreover, during configuration changes, the hashring goes through an even longer period of partial unreadiness, where some nodes begin and finish flushing before and after others. During this partial unreadiness, the hashring can expect high internal request failure rates, which cause clients to retry their requests, resulting in even higher load. Therefore, when the hashring configuration is changed due to automatic horizontal scaling of a set of Thanos Receivers, the system can expect higher than normal resource utilization, which can create a positive feedback loop that continuously scales the hashring.
+
 
 ### Goals
 
+* Reduce the downtime of ingestion part of Thanos
```
yashrsharma44, line 28:
```suggestion
* Reduce downtime of the ingestion operation of Thanos Receiver
```

docs/proposals/202012_receive_split.md
```diff
@@ -8,30 +8,79 @@ status: accepted
...
+We have:
+
+![After](https://docs.google.com/drawings/d/e/2PACX-1vTVrtCGjR4iMbrU7Kj6QAn1a1m4fr-kvoQVDAK4lzQ_wWfXfpLLEE9HB948-WHI5ZG6s1iGWt51R593/pub?w=960&h=720)
+
+This allows us to (optionally) model deployment in a way that avoid expensive re-configuration of the stateful ingesting Receives after the hashring configuration file has changed.
+
+In comparison to previous proposal (as mentioned in [alternatives](#previous-proposal-separate-receive-route-command) we have big adventages:
+
+1. We can reduce number of components in Thanos system, we can reuse similar component flags and documentation. Users has to learn about one less command and in result Thanos design is much more approachable. Less components mean less maintainance, code and other implicit duties: Separate changelogs, issue confusions, boilerplates, etc.
+2. Allow consistent pattern with Query. We don't have separate StorAPI component for proxying, we have that baked into Querier. This has been proven to be flexible and understandable, so I would like to propose similar pattern in Receive.
```
yashrsharma44, line 57:
```suggestion
2. Allow consistent pattern with Query. We don't have separate StoreAPI component for proxying, we have that baked into Querier. This has been proven to be flexible and understandable, so I would like to propose similar pattern in Receive.
```

yashrsharma44 (commented):
Some nitpick and few questions from my side.
Otherwise LGTM 🥇

brancz:
I'm not necessarily arguing with the technological direction (I do agree this is a good, maybe better direction), but I find it worrisome that we let parts of the community work on something for a long time only to step in and "know it better" and turn over a lot of work someone has put into something. This is being everything else but welcoming to new contributors, and I've noticed this behavior more than once. Part of letting people grow is allowing them to make mistakes and not stepping in and having to be part of every conversation.

Clearly, there are enough maintainers to back this proposal up from a technical perspective, so I'm not going to block this proposal, but for the record, I want to state that I find this behavior damaging for the Thanos project as a whole. Not everyone can always be part of every conversation and we need to accept that, even if it means technological decisions we may not necessarily agree with may be made.

bwplotka:
Thanks, Frederic, this is important to all of us. As we discussed offline, there were lots of offline chats and even meeting between @kakkoyun, @Chans321 @yashrsharma44. The trigger for this proposal was actually done on pre-last Contributor Thanos Hours, which is also recorded and available here: 

```
Topic: Thanos Contributor Office Hours
Date: Apr 1, 2021 11:00 AM Universal Time UTC

Meeting Recording:
https://zoom.us/rec/share/VoV2AVCrZYmvaPk27kSftXnTAtVEb-OQqkEiYVxfXNLFWfdSu_sNiAr7CL92888Z.seIHezcDVw2szx8f
```

Unfortunately, you as the mentor of @Chans321 were not updated on all of those communications, sorry for that!

It's true we should be welcoming, but it's also worth to mention we started this work in Community Bridge Q4 2020. Lot's of new ideas came after that I think are fair to consider.

> Not everyone can always be part of every conversation and we need to accept that, even if it means technological decisions may be made that an individual may not necessarily agree with.

Yea, but I treat this proposal as an improvement over the past work, all reusing the previous work and foundations.

docs/proposals/202012_receive_split.md
bwplotka: You are right: It should be "Receiver": https://thanos.io/tip/contributing/contributing.md/#components-naming

docs/proposals/202012_receive_split.md
bwplotka: Hm, wonder if operation is a right word 🤔

bwplotka: changed to `logic`

docs/proposals/202012_receive_split.md
bwplotka: Looks like no idea anymore. Removing then.

docs/proposals/202012_receive_split.md
bwplotka: Well that's the thing. If you don't use hashring flag, it's by design impossible to do hashring forwarding, because you did not specify hashring, right? (:  That's why we can use it as quite explicit turn off of forwarding logic, no?

bwplotka (commented):
Updated, please PTAL again @yashrsharma44 @kakkoyun 

Thanks for good review @yashrsharma44 !

docs/proposals/202012_receive_split.md
```diff
@@ -8,30 +8,74 @@ status: accepted
...
+### Alternative Solutions
+
+#### Previous Proposal: Separate receive-route command
+
+1. Split the Receiver component into receive-route and Receiver (and ensure ease of resharding events).
+1. Evaluate any effects on performance by simulating scenarios and collecting and analyzing metrics.
+1. Use consistent hashing to avoid reshuffling time series after resharding events. The exact consistent hashing mechanism to be used needs some further research.
+1. Migration: We document how the new architecture can be set up to have the same general deployment of the old architecture. (We run router and Receiver on the same node).
+
+This potentially makes the receiver more difficult to operate and understand for Thanos users. I would argue this is howerver much more harder in overall Thanos deployment. Otherwise this option is exactly the same.
```
yashrsharma44, line 74:
```suggestion
This potentially makes the receiver more difficult to operate and understand for Thanos users. I would argue this is however much harder in overall Thanos deployment. Otherwise, this option is exactly the same.
```

docs/proposals/202012_receive_split.md
```diff
@@ -8,30 +8,74 @@ status: accepted
...
+
+In comparison to previous proposal (as mentioned in [alternatives](#previous-proposal-separate-receive-route-command) we have big adventages:
+
+1. We can reduce number of components in Thanos system, we can reuse similar component flags and documentation. Users has to learn about one less command and in result Thanos design is much more approachable. Less components mean less maintainance, code and other implicit duties: Separate changelogs, issue confusions, boilerplates, etc.
+2. Allow consistent pattern with Query. We don't have separate StoreAPI component for proxying, we have that baked into Querier. This has been proven to be flexible and understandable, so I would like to propose similar pattern in Receiver.
+3. This is more future proof for potential advanced cases like chain of routers -> receivers -> routers -> receivers for federated writes, so trees with depth n.
+
+### Plan
+
+* Receiver without `--receive.hashrings` does not forward or replicate requests, it routes straight to multi-tsdb.
```
yashrsharma44, line 61: Not sure about this suggestion, just that the old statement sounded contrapositive 😆  
```suggestion
* Receiver with `--receive.hashrings` forwards or replicate requests, without this, it would route the request straight to multi-tsdb.
```

docs/proposals/202012_receive_split.md
```diff
@@ -8,30 +8,74 @@ status: accepted
...
+* Receiver without ` --Receiver.local-endpoint` will assume that no storage is needed, so will skip creating any resources for multi TSDB.
```
yashrsharma44, line 62: Similar to above
```suggestion
* Receiver with ` --receiver.local-endpoint` would create storage for multi TSDB, without this flag, we will skip creating any resources for multi TSDB.
```

yashrsharma44 (commented):
LGTM 🔥 , just small questions about wording of the statement 😁

docs/proposals/202012_receive_split.md
```diff
@@ -8,30 +8,74 @@ status: accepted
...
+* Add changes to the documentation (it's simplistic now). Mention two modes.
+
+### Alternative Solutions
+
+#### Previous Proposal: Separate receive-route command
+
+1. Split the Receiver component into receive-route and Receiver (and ensure ease of resharding events).
+1. Evaluate any effects on performance by simulating scenarios and collecting and analyzing metrics.
+1. Use consistent hashing to avoid reshuffling time series after resharding events. The exact consistent hashing mechanism to be used needs some further research.
+1. Migration: We document how the new architecture can be set up to have the same general deployment of the old architecture. (We run router and Receiver on the same node).
```
yashrsharma44, line 72:
```suggestion
1. Split the Receiver component into receive-route and Receiver (and ensure ease of resharding events).
2. Evaluate any effects on performance by simulating scenarios and collecting and analyzing metrics.
3. Use consistent hashing to avoid reshuffling time series after resharding events. The exact consistent hashing mechanism to be used needs some further research.
4. Migration: We document how the new architecture can be set up to have the same general deployment of the old architecture. (We run router and Receiver on the same node).
```

yashrsharma44 (dismissed):
Resolved the issues after Office Hours, looks good to me!

docs/proposals/202012_receive_split.md
yashrsharma44: Resolved during office hours :)

docs/proposals/202012_receive_split.md
yashrsharma44: Resolved during office hours :)

docs/proposals/202012_receive_split.md
```diff
@@ -8,30 +8,73 @@ status: accepted
...
+We propose allowing to run Thanos Receiver in a mode that only forwards/replicates remote write (distributor mode). You can enable that mode by simply not specifying:
+
+```yaml
+  --receive.local-endpoint=RECEIVE.LOCAL-ENDPOINT
+                                 Endpoint of local receive node. Used to
+                                 identify the local node in the hashring
+                                 configuration.
+```
+
+We can call this mode a "Routing Receiver". Similarly, we can skip specify any hashring to Thanos Receiver (`--receive.hashrings-file=<path>`), explicitly purposing it only for ingesting. We can call this mode "Ingesting Receiver".
```
GiedriusS, line 41: Grammar nit:
```suggestion
We can call this mode a "Routing Receiver". Similarly, we can skip specifying any hashring to Thanos Receiver (`--receive.hashrings-file=<path>`), explicitly purposing it only for ingesting. We can call this mode "Ingesting Receiver".
```

GiedriusS (dismissed):
LGTM modulo that one grammar nit. It indeed sucks a lot that we have some initial implementations of the previous version but at the same time I feel that it won't be too much work to update them for this improved version as we are only changing how this new mode will be turned on, right?

bwplotka:
Thank you all for feedback and reviews!

Issue #3081, opened 2020-08-26
Fix links; Move preprocess automation for Thanos website to Go


bwplotka:
Also improve link formating/lniting

AC:
* Version picker stuff (websitepreprocess script) should be moved to Golang so we can test and automate more)
* Should convert all Thanos related links to be local links
* Liche should tests future links not current one - the ones that will be produced, so we need to check against deploy when ready (some race condition is to solve between netlify is up and CI)
* Processing for website has to rearannge links a bit (different dir structure, tip etc)
* Pages from old version should point to the same version pages (so need to be moved to local refs)

Potentially me & @thisisobate will focus on this soon

thisisobate:
Happy to focus on this alongside @bwplotka

daixiang0:
Some fixed by #3106

bwplotka:
We are on it with @thisisobate

bwplotka:
TODO
Kind Regards,
Bartek Płotka (@bwplotka)

kakkoyun:
Still valid and help wanted 🤗

bwplotka:
Status: https://github.com/bwplotka/mdox was created. It's currently blocked on inefficient link checking (rate limiting from GH etc). Proposing this as GSoC project 🤗

gdsoumya:
Hi @bwplotka I would like to work on this during GSoC 2021. I was going through the repo and I found some of the automation scripts here https://github.com/thanos-io/thanos/tree/main/scripts/website, is this all that you want to convert to golang?

RavjotSandhu:
How to start working towards this?

RavjotSandhu:
As far I can understand we need to move current automation scripts one consolidated mdox so that we can use them at other projects. Right @bwplotka ?

RavjotSandhu:
@bwplotka @thisisobate

bwplotka:
There is GSoC project for that already, and we are starting soon!

bwplotka:
Defnitely in progress (: cc @saswatamcode

vaibhavdugar:
Hey! I am a newbie, how can I start contributing to this project? And what are the pre-requisites to begin with this? Can someone help me out please.

bwplotka:
Of course, feel free to help. But this particular issue should be closed (: Thanks for reaching. Feel free to grab any other issue that looks interesting.

PR #4037, opened 2021-04-09
Correcting relative path to linked docs

hsdp-smulford:
There are a number of links in the quick-tutorial.md that use an incorrect relative path. The `components` directory is (currently) at the same level as this doc, so using a relative path of `..` makes the path reference incorrect

* [ *] Change is not relevant to the end user.

## Changes

Updated relative path to linked documents from `..` to `.`, correcting the references.

## Verification

The links now work


bwplotka:
Thank you so much but we aim to fix this on all levels using https://github.com/thanos-io/thanos/issues/3081

Problem is that we aim for those links to work on both GH mardown level as well website ):

PR #4041, opened 2021-04-09
logging: fix the HTTP logger

GiedriusS:
Fix the HTTP logger so that it would take a look at the `url.URL` as
well if `r.Host` is empty. Without this, the logging functionality doesn't quite work
as Cortex uses `url.URL`:

https://github.com/cortexproject/cortex/blob/faf92e8f3dbedfa769fbd6a3ff6e0fbd127694f5/pkg/frontend/downstream_roundtripper.go#L16-L23

Add a test to ensure that this works as expected in both ways.

Without this, the user only sees in the logs:

```
level=error ts=2021-04-09T10:22:05.3319924Z caller=http.go:44 protocol=http http.component=server msg="failed to parse host port for http log decision" err="address foo.bar: missing port in address"
```

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -30,6 +30,7 @@ We use _breaking :warning:_ to mark changes that are not backward compatible (re
 - [#3960](https://github.com/thanos-io/thanos/pull/3960) fix deduplication of equal alerts with different labels
 - [#3937](https://github.com/thanos-io/thanos/pull/3937) Store: Fix race condition in chunk pool.
 - [#4017](https://github.com/thanos-io/thanos/pull/4017) Query Frontend: fix downsampling iterator returning duplicate samples.
+- [#4017](https://github.com/thanos-io/thanos/pull/4041) Logging: fix the HTTP logger
 
 ### Changed
 - [#3929](https://github.com/thanos-io/thanos/pull/3929) Store: Adds the name of the instantiated memcached client to log info
--- a/pkg/logging/http.go
+++ b/pkg/logging/http.go
@@ -39,7 +39,11 @@ func (m *HTTPServerMiddleware) HTTPMiddleware(name string, next http.Handler) ht
 	return func(w http.ResponseWriter, r *http.Request) {
 		wrapped := httputil.WrapResponseWriterWithStatus(w)
 		start := time.Now()
-		_, port, err := net.SplitHostPort(r.Host)
+		hostPort := r.Host
+		if hostPort == "" {
+			hostPort = r.URL.Host
+		}
+		_, port, err := net.SplitHostPort(hostPort)
 		if err != nil {
 			level.Error(m.logger).Log("msg", "failed to parse host port for http log decision", "err", err)
 			next.ServeHTTP(w, r)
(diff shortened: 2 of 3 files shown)


pkg/logging/http.go
```diff
@@ -39,7 +39,11 @@ func (m *HTTPServerMiddleware) HTTPMiddleware(name string, next http.Handler) ht
 	return func(w http.ResponseWriter, r *http.Request) {
 		wrapped := httputil.WrapResponseWriterWithStatus(w)
 		start := time.Now()
-		_, port, err := net.SplitHostPort(r.Host)
+		hostPort := r.Host
```
yashrsharma44, line 42: What should we do, if both `r.Host` and `r.URL.Host` is set, which one should we display?

pkg/logging/http_test.go
```diff
@@ -4,34 +4,64 @@
...
 		_, err := io.WriteString(w, "Test Works")
 		if err != nil {
-			t.Log(err)
+			testutil.Ok(t, err)
 		}
 	}
 	hm := m.HTTPMiddleware("test", http.HandlerFunc(handler))
 
-	req := httptest.NewRequest("GET", "http://example.com/foo", nil)
+	// Cortex way:
```
yashrsharma44, line 32:
```suggestion
	// Regression test for Cortex way - https://github.com/thanos-io/thanos/pull/4041
```

yashrsharma44 (commented):
There are few questions, feel free to skip if it doesn't make sense

LGTM 💪

pkg/logging/http.go
onprem: If `r.Host` is there, it will get priority. We are only using `r.URL.Host` if `r.Host` is an empty string.

onprem (dismissed):
LGTM. Thanks for the fix, we are seeing these error logs in our deployments as well.

pkg/logging/http.go
yashrsharma44: I think we can print anyone of them, as long one of them is non-empty.

GiedriusS: From the docs:

```
    // For client requests, the URL's Host specifies the server to
    // connect to, while the Request's Host field optionally
    // specifies the Host header value to send in the HTTP
    // request.
```

So, we should give priority to `r.Host` over `r.URL.Host` because `r.Host` _might_ be more specific i.e. it's the `Host` header :stuck_out_tongue:

yashrsharma44: Looks solved then 😄 !

yashrsharma44 (approved):

CHANGELOG.md
```diff
@@ -30,6 +30,7 @@ We use _breaking :warning:_ to mark changes that are not backward compatible (re
 - [#3960](https://github.com/thanos-io/thanos/pull/3960) fix deduplication of equal alerts with different labels
 - [#3937](https://github.com/thanos-io/thanos/pull/3937) Store: Fix race condition in chunk pool.
 - [#4017](https://github.com/thanos-io/thanos/pull/4017) Query Frontend: fix downsampling iterator returning duplicate samples.
+- [#4017](https://github.com/thanos-io/thanos/pull/4041) Logging: fix the HTTP logger
```
bwplotka, line 33: Let's mention what is the fix exactly 🤗

pkg/logging/http.go
bwplotka: 👍🏽

bwplotka (approved):
Thanks! LGTM, just let's fix CHANGELOG but we can fix it anytime during release

PR #4042, opened 2021-04-10
UI: Fix conflicts in dependencies

onprem:

## Changes
- Some dependencies were present in both `dependencies` and `devDependencies` section. I kep them in `devDependencies` (since both of them are related to tests) to avoid conflicts.
- I also explicitly added `lezer` as a dependency since it is required as a peer dependency by `codemirror-promql`.

## Verification
`yarn test`


bwplotka (approved):
Thanks!

PR #4013, opened 2021-04-02
Fix the metadata API flaky tests

yeya24:
Signed-off-by: yeya24 <yb532204897@gmail.com>

* [ ] I added CHANGELOG entry for this change.
* [x] Change is not relevant to the end user.

Fixes #4012

## Changes

Refactor the metadata E2E test and use `metadataEqual` function to compare the two responses instead of just using reflect.DeepEqual.

In `metadataEqual` function, each entry in the metadata response is first sorted and then compare. This fixes the problem mentioned in https://github.com/thanos-io/thanos/issues/4012#issuecomment-812542755

Update: I also rewrite the unit tests of metadata API.

PR #4044, opened 2021-04-10
Fix prometheus target tests with correct address

### LINKED ISSUE #4021: TestPrometheus_Targets_e2e doesn't work as expected
The E2E test case for target https://github.com/thanos-io/thanos/blob/main/pkg/targets/prometheus_test.go#L23 doesn't work as expected.

It uses the config file below to start Prometheus. The addresses are hardcoded.

``` yaml
testutil.Ok(t, p.SetConfig(`
global:
  external_labels:
    region: eu-west
scrape_configs:
- job_name: 'myself'
  # Quick scrapes for test purposes.
  scrape_interval: 1s
  scrape_timeout: 1s
  static_configs:
  - targets: ['localhost:9090','localhost:80']
  relabel_configs:
  - source_labels: ['__address__']
    regex: '^.+:80$'
    action: drop
`))
	testutil.Ok(t, p.Start())
```

However, the Prometheus instances are started using `e2eutil.NewPrometheus` which allocates a random port for Prometheus. So actually the scrape config is not working. We need to change the target address to the allocated Prometheus address.

A workaround might be:
1. Create the Prometheus with a random address and use the default configuration with target `localhost:9090`
2. Start Prometheus and wait for it ready. Now we can know the real running port of Prometheus
3. Update the target  address of the config to the real Prometheus address
4. Send a SIGHUP or hit `/-/reload` endpoint for a reload

This is the approach I used to fix the same bug for metadata API https://github.com/thanos-io/thanos/pull/4013.

yeya24:

## Changes

Fix #4021.

Use the same way in pr #4013

### DIFF
--- a/pkg/promclient/promclient.go
+++ b/pkg/promclient/promclient.go
@@ -798,5 +798,5 @@ func (c *Client) TargetsInGRPC(ctx context.Context, base *url.URL, stateTargets 
 	var v struct {
 		Data *targetspb.TargetDiscovery `json:"data"`
 	}
-	return v.Data, c.get2xxResultWithGRPCErrors(ctx, "/targets HTTP[client]", &u, &v)
+	return v.Data, c.get2xxResultWithGRPCErrors(ctx, "/prom_targets HTTP[client]", &u, &v)
 }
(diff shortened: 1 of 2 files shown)


bwplotka (approved):
Perfect, thanks! 

I remember actually reviewing this and missed this 🤦🏽

PR #4045, opened 2021-04-11
UI: Enable Targets page in Querier UI

onprem:
Signed-off-by: Prem Saraswat <prmsrswt@gmail.com>

## Changes
- Added the missing field `globalUrl` in proto, which was used in the UI
- Added `Targets` page to the Querier nav bar

## Verification

The `json` response format of Querier API matches with the response from Prometheus API.

Querier Targets page


pkg/targets/targetspb/rpc.proto
```diff
@@ -67,10 +67,11 @@ message ActiveTarget {
     ZLabelSet labels = 2 [(gogoproto.jsontag) = "labels", (gogoproto.nullable) = false];
     string scrapePool = 3 [(gogoproto.jsontag) = "scrapePool"];
     string scrapeUrl = 4 [(gogoproto.jsontag) = "scrapeUrl"];
-    string lastError = 5 [(gogoproto.jsontag) = "lastError"];
-    google.protobuf.Timestamp lastScrape = 6 [(gogoproto.jsontag) = "lastScrape", (gogoproto.stdtime) = true, (gogoproto.nullable) = false];
-    double lastScrapeDuration = 7 [(gogoproto.jsontag) = "lastScrapeDuration"];
-    TargetHealth health = 8 [(gogoproto.jsontag) = "health"];
+    string globalUrl = 5 [(gogoproto.jsontag) = "globalUrl"];
```
yeya24, line 70: We missed this in the review as this field doesn't exist in Prometheus doc https://prometheus.io/docs/prometheus/latest/querying/api/#targets.
Thanks for catching this!

yeya24 (dismissed):
LGTM!

pkg/targets/targetspb/rpc.proto
onprem: I just caught this because it was required in UI haha. Thanks for the review :)

GiedriusS (commented):
LGTM, just have one question: maybe it's time to make `--target` & other flags not hidden anymore? Because I think with this link in the UI we are going to inevitably get questions about what the link is for if it's nowhere to be found :s

onprem:
> maybe it's time to make --target & other flags not hidden anymore?

Oh, I was not aware that it is hidden. I also agree to make this not hidden.

GiedriusS:
> > maybe it's time to make --target & other flags not hidden anymore?
> 
> Oh, I was not aware that it is hidden. I also agree to make this not hidden.

Should we unhide them in a separate PR before merging this one? :P

squat:
> > > maybe it's time to make --target & other flags not hidden anymore?
> > 
> > 
> > Oh, I was not aware that it is hidden. I also agree to make this not hidden.
> 
> Should we unhide them in a separate PR before merging this one? :P

:+1:

bwplotka (approved):
I would propose to not UNHIDE them until @hitanshu-mehta project of unificating those into single `endpoint` flag as planned in https://thanos.io/tip/proposals/202101_endpoint_discovery.md/ WDYT? 🤗 

Otherwise LGTM!

GiedriusS:
> I would propose to not UNHIDE them until @hitanshu-mehta project of unificating those into single `endpoint` flag as planned in https://thanos.io/tip/proposals/202101_endpoint_discovery.md/ WDYT? 

I think it would lead to lots of questions about what is that page for and why one cannot any instructions about how to use it :thinking:

onprem:
So I guess one solution is to make targets page working but leave it out from the nav bar (similar to a hidden flag, a hidden page). So, whenever we decide to make the flag non-hidden (or we move to the `--endpoint` flag), we can make the page non-hidden too. WDYT?

onprem:
I have rebased this and made the Targets page hidden from Navbar (one can manually go to `/new/targets` though). If we want to unhide the `--target` flag and the Targets page, we can do that in a separate PR.

kakkoyun (approved):
LGTM 🚀

PR #4047, opened 2021-04-12
UI: Upgrade hosted-git-info to address a CVE-2021-23362

kakkoyun:

## Verification

* `yarn test`

### DIFF
--- a/pkg/ui/react-app/package.json
+++ b/pkg/ui/react-app/package.json
@@ -35,6 +35,7 @@
     "downshift": "^6.1.0",
     "enzyme-to-json": "^3.6.1",
     "fuzzy": "^0.1.3",
+    "hosted-git-info": "2.8.9",
     "i": "^0.3.6",
     "jquery": "^3.5.1",
     "jquery.flot.tooltip": "^0.9.0",
--- a/pkg/ui/react-app/yarn.lock
+++ b/pkg/ui/react-app/yarn.lock
@@ -6081,7 +6081,7 @@ hoist-non-react-statics@^3.3.1:
   dependencies:
     react-is "^16.7.0"
 
-hosted-git-info@^2.1.4:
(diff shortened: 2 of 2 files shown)


squat:
sgtm cc @onprem does this actually have the intended effect?

squat (approved):

pkg/ui/react-app/package.json
```diff
@@ -35,6 +35,7 @@
     "downshift": "^6.1.0",
     "enzyme-to-json": "^3.6.1",
     "fuzzy": "^0.1.3",
+    "hosted-git-info": "2.8.9",
```
onprem, line 38: This is not a direct dependency so we should not add it in the `package.json` file

pkg/ui/react-app/yarn.lock
```diff
@@ -6081,7 +6081,7 @@ hoist-non-react-statics@^3.3.1:
   dependencies:
     react-is "^16.7.0"
 
-hosted-git-info@^2.1.4:
+hosted-git-info@2.8.9, hosted-git-info@^2.1.4:
   version "2.8.9"
```
onprem, line 6085: Looks like we are already using version `2.8.9`, maybe we can just bump the minimum allowed version (^2.1.4 to ^2.8.9)

 ```suggestion
hosted-git-info@^2.8.9:
  version "2.8.9"
```

onprem (commented):
> does this actually have the intended effect?

Looks like the actual version we are using is already 2.8.9, hence the hashes etc were not changed.

pkg/ui/react-app/yarn.lock
kakkoyun: I guess we don't need this change. As you said, we already have the correct version as a dependency.

kakkoyun:
https://github.com/thanos-io/thanos/pull/4047#discussion_r612549394

PR #4053, opened 2021-04-13
receiver: fix the calculation logic of minimum number of replicas

GaoJiasheng:

## Changes
I modified the pattern to `int(((h.options.ReplicationFactor - 1) / 2) + 1)`.
The original pattern is `int((h.options.ReplicationFactor / 2) + 1)`. there will be a issue when the `ReplicationFactor` is even numbers.
if the `ReplicationFactor` is `2`. the value of the pattern would be `(2/2+1)=2`. therefore, both replicas must be successfull to return success to prometheus. 
if the `ReplicationFactor` is `4`, the value of the pattern would be `(4/2+1)=3`. therefore, there need to be 3 successful replicas to return success to prometheus.

Recently we alter the replicas from 1 to 2, but all the receivers would be stuck if one receiver was crashed. 
It works after alter this pattern.

## Verification
image

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -29,6 +29,7 @@ We use _breaking :warning:_ to mark changes that are not backward compatible (re
 - [#3937](https://github.com/thanos-io/thanos/pull/3937) Store: Fix race condition in chunk pool.
 - [#4017](https://github.com/thanos-io/thanos/pull/4017) Query Frontend: fix downsampling iterator returning duplicate samples.
 - [#4017](https://github.com/thanos-io/thanos/pull/4041) Logging: fix the HTTP logger
+- [#4053](https://github.com/thanos-io/thanos/pull/4053) receiver: fix the calculation logic of minimum number of successful replicas
 
 ### Changed
 - [#3929](https://github.com/thanos-io/thanos/pull/3929) Store: Adds the name of the instantiated memcached client to log info
--- a/pkg/receive/handler.go
+++ b/pkg/receive/handler.go
@@ -395,7 +395,7 @@ func (h *Handler) forward(ctx context.Context, tenant string, r replica, wreq *p
 
 // writeQuorum returns minimum number of replicas that has to confirm write success before claiming replication success.
 func (h *Handler) writeQuorum() int {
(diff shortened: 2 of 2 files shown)


wiardvanrij:
Thanks so much for the clear enumeration of the change. Really helpful. 

There is also a difference between replication factor which 'dictates' the quorum and the actual replica's you are running for the receive. IMO it's always better to run more replica's than your replication factor is, so you can always have your quorum.

I.e. run one or two replica's for a replica factor of 1. Run a minimum of 3 replica's for a replication factor of 2. 
So your change would work but I'm not sure if it's good to basically run a higher replication factor than actual replica's required. Especially on the 'lower' figure segment. For example your change would require only 1 receive replica to handle a replication factor of 2. 

Personally my 2 cents here is that one should run a higher replica count to meet quorum quota's rather than lowering the quorum quota. What's your thought on this? 

current factor & quorum
```
replication factor: 1 quorum: 1  // but this is fine regardless, no quorum required for 1 replica factor.
replication factor: 2 quorum: 2 
replication factor: 3 quorum: 2 
replication factor: 4 quorum: 3 
replication factor: 5 quorum: 3 
replication factor: 6 quorum: 4 
replication factor: 7 quorum: 4 
replication factor: 8 quorum: 5 
replication factor: 9 quorum: 5 
replication factor: 10 quorum: 6  
```

This change
```
replication factor: 1 quorum: 1 
replication factor: 2 quorum: 1 
replication factor: 3 quorum: 2 
replication factor: 4 quorum: 2 
replication factor: 5 quorum: 3 
replication factor: 6 quorum: 3 
replication factor: 7 quorum: 4 
replication factor: 8 quorum: 4 
replication factor: 9 quorum: 5 
replication factor: 10 quorum: 5
```

squat:
Hi @GaoJiasheng, the current logic is intentional and not actually a bug. We in fact wanted the hashring to require at least more than one half of the nodes to agree I'm order to consider a write successful. If we want to change this then I would suggest we first continue the discussion on one of the existing issues like https://github.com/thanos-io/thanos/issues/3194

PR #4064, opened 2021-04-15
mixin: Downgrade severity of forward failures

brancz:
We had this alert fire during a rollout, however, it is expected that during rollouts some forward requests fail. As long as ingestion as a whole is still succeeding with a quorum, there's no need to alert on anything.

Signed-off-by: Frederic Branczyk <fbranczyk@gmail.com>

cc @kakkoyun @squat @metalmatze


kakkoyun (dismissed):
Looks good to me. This is a cause-based alert and it makes sense to reduce the severity.

CI is unhappy because it misses generated files. I mark it to auto-merge, it'd be great if you can make the CI happy.

kakkoyun (approved):

PR #4070, opened 2021-04-16
Enhance unified endpoint discovery proposal

hitanshu-mehta:

## Changes

1. Added status and owner.
2. Fixed typos.
3. Used Info Service instead of Info API in the main solution.
4. Removed alternative solutions because both main and alternative solutions were similar as discussed [here](https://github.com/thanos-io/thanos/discussions/4003).
4. Added a new alternative solution which proposes to add Info method in all services (eg. Rule, Targets...). Also added possible disadvantages of this approach.

### DIFF
--- a/docs/proposals/202101_endpoint_discovery.md
+++ b/docs/proposals/202101_endpoint_discovery.md
@@ -29,7 +29,7 @@ Unite discovery of endpoints, which would make code cleaner as well as provide b
 
 Add a new flag called `--endpoint` to Thanos query, and auto-discover what services that endpoint is serving based on metadata each gRPC server exposes.
 
-Each component will expose an Info service, that includes various metadata listed below. Discovery of endpoints will happen via this Info service. there might be a case that the discovery of gRPC will also have to happen via [gRPC reflection](https://github.com/grpc/grpc/blob/master/doc/server-reflection.md), this will be clear once implementation starts.
+Each component will expose an Info service, that includes various metadata listed below. Discovery of endpoints will happen via this Info service. There might be a case that the discovery of gRPC will also have to happen via [gRPC reflection](https://github.com/grpc/grpc/blob/master/doc/server-reflection.md), this will be clear once implementation starts.
 
 Info API metadata would include the following fields regardless of the type:
 ```
@@ -42,7 +42,7 @@ info:
     blah: ...
 ```
 
-Right now this info service would be added to existing components Ruler, Store, Sidecar, Compact, Downsample, Receive.
+Right now this Info service would be added to existing components Ruler, Store, Sidecar and Receive.
 
 The upgrade path would be possible in this case, as it's strictly additive for endpoints, and there will be a migration path to fall back to current discovery methods which could be removed in the future.
 
@@ -52,7 +52,7 @@ Initial discovery of gRPC services available will happen via gRPC reflection. To
 
 This alternative solution was also considered, however it was not chosen because of its following drawbacks.
 
-1. For servers implementing multiple API (Eg. Store, Rules, Exemplar), we need to make multiple microservices calls (For given example 3x calls).
+1. For servers implementing multiple API (e.g. Store, Rules, Exemplar), we need to make multiple microservices calls (For given example 3x calls).
 2. There is also ambiguity about which Info method should be used for these kinds of servers.
 
 ### Upgrade plan
@@ -61,6 +61,6 @@ Regardless of the solution, we will be not removing any flags for a given period
 
 ### Work plan
 
-1. We would be adding the new info service and flags needed for the above solution, so this part would be entirely additive.
+1. We would be adding the new Info service and flags needed for the above solution, so this part would be entirely additive.
 2. Second step would be the integration of this service, flags and migration of existing code.
 3. We would not be removing the existing flags (`--store` and `--rule`) for some grace period, after that passes we would also have to remove it and any migration code.
--- a/docs/proposals/202101_endpoint_discovery.md
+++ b/docs/proposals/202101_endpoint_discovery.md
@@ -19,7 +19,7 @@ We want to propose a new flag called `--endpoint=<address>` that will be passed 
 
 Currently, in Thanos Query, the discovery of rules APIs happens via Store API's Info method. This makes it harder if we ever want to not have a coupling to the Store API (which is already planned for [scalable ruler proposal](https://github.com/thanos-io/thanos/blob/main/docs/proposals/202005_scalable-rule-storage.md)).
 
-We also require passing two different flags to the Thanos Query component `--store=<address>` and `--rule=<address>`. If users use both flags, the Query component performs DNS discovery on often the same address multiple times, this can cause occasional DNS issues as this resolving happens so frequently. This is especially confusing when one works and the other doesn't. Adding new APIs in the future would exacerbate this issue with DNS requests.
+We also require passing two different flags to the Thanos Query component `--store=<address>` and `--rule=<address>`. If users use both flags, the Query component performs DNS discovery on often the same address multiple times, which can cause occasional DNS issues as this resolution happens so frequently. This is especially confusing when one DNS lookup works and the other doesn't. Adding new APIs in the future would exacerbate this issue with DNS requests.
 
 ### Goals
 
@@ -44,7 +44,7 @@ info:
 
 Right now this Info service would be added to existing components exposing a gRPC server.
 
-The upgrade path would be possible in this case, as it's strictly additive for endpoints, and there will be a migration path to fall back to current discovery methods which could be removed in the future.
+The upgrade path would be possible in this case, as it's strictly additive for endpoints, and there will be a migration path to fall back to current discovery methods, which could be removed in the future.
 
 ### Alternatives
 
@@ -52,15 +52,15 @@ Initial discovery of gRPC services available will happen via gRPC reflection. To
 
 This alternative solution was also considered, however it was not chosen because of its following drawbacks.
 
-1. For servers implementing multiple API (e.g. Store, Rules, Exemplar), we need to make multiple microservices calls (For given example 3x calls).
+1. For servers implementing multiple APIs (e.g. Store, Rules, Exemplar), we need to make multiple gRPC calls, e.g. three calls to a component implementing Store, Rules and Exemplars APIs.
 2. There is also ambiguity about which Info method should be used for these kinds of servers.
 
 ### Upgrade plan
 
-Regardless of the solution, we will be not removing any flags for a given period and have a fallback code in place to be able to discover in the same way as we do right now. This will ensure a smooth upgrade path for the users that make use of the current flags for discovery. We will mark the current flags as deprecated and after two releases have passed we can remove the existing store and rule flag from the query component and remove any migration code.
+Regardless of the solution, we will be not removing any flags for a given period and have a fallback in place to be able to discover in the same way as we do right now. This will ensure a smooth upgrade path for the users who make use of the current flags for discovery. We will mark the current flags as deprecated and after two releases have passed we can remove the existing component-specific discovery flags from the query component and remove any migration code.
 
 ### Work plan
 
 1. We would be adding the new Info service and flags needed for the above solution, so this part would be entirely additive.
-2. Second step would be the integration of this service, flags and migration of existing code.
+2. Second step would be the integration of this service, the addition of new flags, and migration of existing code.
 3. We would not be removing the existing flags (`--store` and `--rule`) for some grace period, after that passes we would also have to remove it and any migration code.
--- a/docs/proposals/202101_endpoint_discovery.md
+++ b/docs/proposals/202101_endpoint_discovery.md
@@ -19,9 +19,7 @@ We want to propose a new flag called `--endpoint=<address>` that will be passed 
 
 Currently, in Thanos Query, the discovery of rules APIs happens via Store API's Info method. This makes it harder if we ever want to not have a coupling to the Store API (which is already planned for [scalable ruler proposal](https://github.com/thanos-io/thanos/blob/master/docs/proposals/202005_scalable-rule-storage.md)).
 
-We also require passing two different flags to the Thanos Query component `--store=<address>` and `--rule=<address>`. If users use both flags, the Query component performs DNS discovery on often the same address multiple times, this can cause occasional DNS issues as this resolving happens so frequently. This is especially confusing when one works and the other doesn't.
-
-Adding new APIs in the future alike.
+We also require passing two different flags to the Thanos Query component `--store=<address>` and `--rule=<address>`. If users use both flags, the Query component performs DNS discovery on often the same address multiple times, this can cause occasional DNS issues as this resolving happens so frequently. This is especially confusing when one works and the other doesn't. Adding new APIs in the future would exacerbate this issue with DNS requests.
 
 ### Goals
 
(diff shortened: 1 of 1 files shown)


docs/proposals/202101_endpoint_discovery.md
```diff
@@ -19,17 +21,17 @@ Currently, in Thanos Query, the discovery of rules APIs happens via Store API's
 
 We also require passing two different flags to the Thanos Query component `--store=<address>` and `--rule=<address>`. If users use both flags, the Query component performs DNS discovery on often the same address multiple times, this can cause occasional DNS issues as this resolving happens so frequently. This is especially confusing when one works and the other doesn't.
 
-Adding new APIs in the future, like
+Adding new APIs in the future alike.
```
GiedriusS, line 24: This sound a bit weird. `alike` adverb means:

```
in the same or a similar way.
```

So it is missing extra information. Perhaps:

```suggestion
Adding new APIs in the future would exacerbate this issue.
```

docs/proposals/202101_endpoint_discovery.md
```diff
@@ -19,17 +21,17 @@ Currently, in Thanos Query, the discovery of rules APIs happens via Store API's
...
+Unite discovery of endpoints, which would make code cleaner as well as provide building blocks to provide discovery and support for new gRPC services, for example, the Targets API/Exemplar API/Metadata API.
 
 ### Solution
 
 Add a new flag called `--endpoint` to Thanos query, and auto-discover what services that endpoint is serving based on metadata each gRPC server exposes.
 
-Each component will expose an Info API, that includes various metadata listed below. Discovery of endpoints will happen via this Info API, there might be a case that the discovery of gRPC will also have to happen via [gRPC reflection](https://github.com/grpc/grpc/blob/master/doc/server-reflection.md), this will be shown when implemention starts.
+Each component will expose an Info service, that includes various metadata listed below. Discovery of endpoints will happen via this Info service, there might be a case that the discovery of gRPC will also have to happen via [gRPC reflection](https://github.com/grpc/grpc/blob/master/doc/server-reflection.md), this will be shown when implementation starts.
```
GiedriusS, line 34:
```suggestion
Each component will expose an Info service, that includes various metadata listed below. Discovery of endpoints will happen via this Info service, there might be a case that the discovery of gRPC will also have to happen via [gRPC reflection](https://github.com/grpc/grpc/blob/master/doc/server-reflection.md), this will be clear once implementation starts.
```

docs/proposals/202101_endpoint_discovery.md
```diff
@@ -42,20 +44,25 @@ info:
...
+The upgrade path would be possible in this case, as it's strictly additive for endpoints, and there will be a migration path to fall back to current discovery methods which could be removed in the future.
 
-#### Alternatives
+### Alternatives
 
-Split Info method out of the Store API into a separate gRPC service called `info`. Each component must implement and expose this. The discovery would happen via this info gRPC service, which each component would need to explicitly state which API it supports. Besides that information, we would also need to include here various info, e.g. LabelSets, Max/Min time, etc. The drawback for this solution is harder upgrades as this would be more complex to implement as we would need to fall back to older APIs.
+The plan would be for the initial discovery of gRPC services available to happen via gRPC reflection. To get further information, we would be adding an Info method to other services for example, Targets, Exemplars, etc. This would contain similar info or metadata information as we get right now from the Store Info method. We can always add more data to it in the future as needed.
```
GiedriusS, line 53:
```suggestion
The plan would be for the initial discovery of gRPC services available to happen via gRPC reflection. To get further information, we would be adding an Info method to other services, for example, Targets, Exemplars, etc. This would contain similar info or metadata information as we get right now from the Store Info method. We can always add more data to it in the future as needed.
```

GiedriusS: So, will it happen via the Info service or gRPC reflection? I'm confused about this sentence :smile: Maybe I am missing/misunderstanding something. I thought with this proposal we will only use the Info service and we are going to try avoiding reflection

hitanshu-mehta: Yes. We are only planning to use the Info service.
During discussions, It was also proposed that we can add Info methods (similar to the Info method we currently have in store API) to all other services (like rules, targets...) instead of creating a new Info service. Discovery of these methods may be done using gRPC reflection ( still depends on implementation 😅 ). That's why I added this as an alternative solution. 

I am also not sure about should we mention this solution as alternative solution? WDYT :)

docs/proposals/202101_endpoint_discovery.md
hitanshu-mehta: I think we should add this in the previous paragraph because it is also talking about the same thing?

docs/proposals/202101_endpoint_discovery.md
squat, line 53: Just to be perfectly clear this time around, I would suggest including a note along the lines of:
```
The following alternatives were also considered, however they were not chosen because of their respective drawbacks.
1. ...
2. ...
```

docs/proposals/202101_endpoint_discovery.md
```diff
@@ -17,19 +19,17 @@ We want to propose a new flag called `--endpoint=<address>` that will be passed
...
+Unite discovery of endpoints, which would make code cleaner as well as provide building blocks to provide discovery and support for new gRPC services, for example, the Targets API/Exemplar API/Metadata API.
 
 ### Solution
 
 Add a new flag called `--endpoint` to Thanos query, and auto-discover what services that endpoint is serving based on metadata each gRPC server exposes.
 
-Each component will expose an Info API, that includes various metadata listed below. Discovery of endpoints will happen via this Info API, there might be a case that the discovery of gRPC will also have to happen via [gRPC reflection](https://github.com/grpc/grpc/blob/master/doc/server-reflection.md), this will be shown when implemention starts.
+Each component will expose an Info service, that includes various metadata listed below. Discovery of endpoints will happen via this Info service. there might be a case that the discovery of gRPC will also have to happen via [gRPC reflection](https://github.com/grpc/grpc/blob/master/doc/server-reflection.md), this will be clear once implementation starts.
```
GiedriusS, line 32:
```suggestion
Each component will expose an Info service, that includes various metadata listed below. Discovery of endpoints will happen via this Info service. There might be a case that the discovery of gRPC will also have to happen via [gRPC reflection](https://github.com/grpc/grpc/blob/master/doc/server-reflection.md), this will be clear once implementation starts.
```

docs/proposals/202101_endpoint_discovery.md
```diff
@@ -42,20 +42,25 @@ info:
     blah: ...
 ```
 
-Right now this info API would be added to existing components Ruler, Store, Sidecar, Compact, Downsample, Receive. But this would be added in the future for Targets, Exemplers, etc.
+Right now this info service would be added to existing components Ruler, Store, Sidecar, Compact, Downsample, Receive.
```
GiedriusS, line 45: `bucket downsample` doesn't expose anything over gRPC now (and probably will never do that) so I think we shouldn't mention it here: 
```suggestion
Right now this Info service would be added to existing components Ruler, Store, Sidecar, Compact, Receive.
```

docs/proposals/202101_endpoint_discovery.md
```diff
@@ -42,20 +42,25 @@ info:
...
+Initial discovery of gRPC services available will happen via gRPC reflection. To get further information, we would be adding an Info method to other services for example, Targets, Exemplars, etc. This would contain similar info or metadata information as we get right now from the Store Info method. We can always add more data to it in the future as needed.
 
-#### Alternatives
+This alternative solution was also considered, however it was not chosen because of its following drawbacks.
 
-Split Info method out of the Store API into a separate gRPC service called `info`. Each component must implement and expose this. The discovery would happen via this info gRPC service, which each component would need to explicitly state which API it supports. Besides that information, we would also need to include here various info, e.g. LabelSets, Max/Min time, etc. The drawback for this solution is harder upgrades as this would be more complex to implement as we would need to fall back to older APIs.
+1. For servers implementing multiple API (Eg. Store, Rules, Exemplar), we need to make multiple microservices calls (For given example 3x calls).
```
GiedriusS, line 55:
```suggestion
1. For servers implementing multiple API (e.g. Store, Rules, Exemplar), we need to make multiple microservices calls (For given example 3x calls).
```

docs/proposals/202101_endpoint_discovery.md
yeya24: I have the same question. How do we expose gRPC on compact? For what APIs?

onprem: Yeah, I think this was just a slight oversight on our part, there's no plan to have a gRPC API on Compactor yet.

hitanshu-mehta: yes agree with @onprem, It will not affect any component not exposing anything over gRPC. we will remove both compact and downsample.

docs/proposals/202101_endpoint_discovery.md
onprem: > Discovery of endpoints will happen via this Info service.

I think we agreed on using gRPC reflection to discover what services a particular endpoint is exposing? I yes then maybe we should modify the statement to something like 
```suggestion
Each component will expose an Info service, that includes various metadata listed below. Discovery of what services an endpoint is exposing will happen via [gRPC reflection](https://github.com/grpc/grpc/blob/master/doc/server-reflection.md). Once we discover what services that endpoint is exposing (e.g. StoreAPI, TargetsAPI) we can retrieve metadata about them using a single `Info` method call from the Info service.
```

docs/proposals/202101_endpoint_discovery.md
onprem: And we are missing Querier here too, as it also exposes a Store API we should add Info service there as well. Maybe not mentioning the exact component names will be better.
```suggestion
Right now this Info service would be added to existing components exposing a gRPC server.
```

docs/proposals/202101_endpoint_discovery.md
hitanshu-mehta: Yes, I missed it. Thank you for pointing out :)

docs/proposals/202101_endpoint_discovery.md
```diff
@@ -2,34 +2,34 @@
...
+Currently, in Thanos Query, the discovery of rules APIs happens via Store API's Info method. This makes it harder if we ever want to not have a coupling to the Store API (which is already planned for [scalable ruler proposal](https://github.com/thanos-io/thanos/blob/main/docs/proposals/202005_scalable-rule-storage.md)).
 
-We also require passing two different flags to the Thanos Query component `--store=<address>` and `--rule=<address>`. If users use both flags, the Query component performs DNS discovery on often the same address multiple times, this can cause occasional DNS issues as this resolving happens so frequently. This is especially confusing when one works and the other doesn't.
-
-Adding new APIs in the future, like
+We also require passing two different flags to the Thanos Query component `--store=<address>` and `--rule=<address>`. If users use both flags, the Query component performs DNS discovery on often the same address multiple times, this can cause occasional DNS issues as this resolving happens so frequently. This is especially confusing when one works and the other doesn't. Adding new APIs in the future would exacerbate this issue with DNS requests.
```
squat, line 22:
```suggestion
We also require passing two different flags to the Thanos Query component `--store=<address>` and `--rule=<address>`. If users use both flags, the Query component performs DNS discovery on often the same address multiple times, which can cause occasional DNS issues as this resolution happens so frequently. This is especially confusing when one DNS lookup works and the other doesn't. Adding new APIs in the future would exacerbate this issue with DNS requests.
```

docs/proposals/202101_endpoint_discovery.md
```diff
@@ -42,20 +42,25 @@ info:
     blah: ...
 ```
 
-Right now this info API would be added to existing components Ruler, Store, Sidecar, Compact, Downsample, Receive. But this would be added in the future for Targets, Exemplers, etc.
+Right now this Info service would be added to existing components exposing a gRPC server.
+
+The upgrade path would be possible in this case, as it's strictly additive for endpoints, and there will be a migration path to fall back to current discovery methods which could be removed in the future.
```
squat, line 47:
```suggestion
The upgrade path would be possible in this case, as it's strictly additive for endpoints, and there will be a migration path to fall back to current discovery methods, which could be removed in the future.
```

docs/proposals/202101_endpoint_discovery.md
```diff
@@ -42,20 +42,25 @@ info:
...
+Initial discovery of gRPC services available will happen via gRPC reflection. To get further information, we would be adding an Info method to other services for example, Targets, Exemplars, etc. This would contain similar info or metadata information as we get right now from the Store Info method. We can always add more data to it in the future as needed.
 
-#### Alternatives
+This alternative solution was also considered, however it was not chosen because of its following drawbacks.
 
-Split Info method out of the Store API into a separate gRPC service called `info`. Each component must implement and expose this. The discovery would happen via this info gRPC service, which each component would need to explicitly state which API it supports. Besides that information, we would also need to include here various info, e.g. LabelSets, Max/Min time, etc. The drawback for this solution is harder upgrades as this would be more complex to implement as we would need to fall back to older APIs.
+1. For servers implementing multiple API (e.g. Store, Rules, Exemplar), we need to make multiple microservices calls (For given example 3x calls).
```
squat, line 55:
```suggestion
1. For servers implementing multiple APIs (e.g. Store, Rules, Exemplar), we need to make multiple gRPC calls, e.g. three calls for a component implement Store, Rules and Exemplars APIs.
```

docs/proposals/202101_endpoint_discovery.md
```diff
@@ -42,20 +42,25 @@ info:
...
+2. There is also ambiguity about which Info method should be used for these kinds of servers.
 
 ### Upgrade plan
 
-Regardless of the solution we will be not removing any flags for a given period of time and have fallback code in place to be able to discover in the same way as we do right now. This will ensure smooth upgrade path for the users that make use of the current flags for discovery. We will mark the current flags as depracted and after two releases have passed we can remove the existing store and rule flag from query component and remove any migration code.
+Regardless of the solution, we will be not removing any flags for a given period and have a fallback code in place to be able to discover in the same way as we do right now. This will ensure a smooth upgrade path for the users that make use of the current flags for discovery. We will mark the current flags as deprecated and after two releases have passed we can remove the existing store and rule flag from the query component and remove any migration code.
```
squat, line 60:
```suggestion
Regardless of the solution, we will be not removing any flags for a given period and have a fallback in place to be able to discover in the same way as we do right now. This will ensure a smooth upgrade path for the users who make use of the current flags for discovery. We will mark the current flags as deprecated and after two releases have passed we can remove the existing store and rule flag from the query component and remove any migration code.
```

docs/proposals/202101_endpoint_discovery.md
```diff
@@ -42,20 +42,25 @@ info:
...
 
 ### Work plan
 
-1. We would be adding the new info APIs and flags needed for the above solution, so this part would be entirely additive.
-2. Second step would be the integration of this APIs and flags and migration of existing code.
-3. We would not be removing the existing flags (`--store` and `--rule`) for some grace period of time, after that passes we would also have to remove it and any migration code.
+1. We would be adding the new Info service and flags needed for the above solution, so this part would be entirely additive.
+2. Second step would be the integration of this service, flags and migration of existing code.
```
squat, line 65:
```suggestion
2. Second step would be the integration of this service, the addition of new flags flags, and migration of existing code.
```

squat (approved):
The proposal is much clearer now! Thank you :rocket:  Just some small language cleanup and LGTM

docs/proposals/202101_endpoint_discovery.md
onprem: @GiedriusS we did a few changes in the proposal to make this clear. The TLDR is, we are going to use reflection to discover what services (StoreAPI, TargetsAPI, etc.) a particular endpoint is exposing. The alternative is to expose this information as part of the `Info` service itself. We are leaning towards reflection because in this case if some component adds support for a new service (ex. Store GW adds support for Exemplars) it will be automatically discovered by the Querier without extra work in code.

I think it's worth to explain why we are using reflection in the proposal as well.

hitanshu-mehta:
@squat Sorry, I tried to commit directly from Github that dismissed your approved review. Can you please review the pr again? I have committed suggested changes too.

squat (approved):

Issue #4046, opened 2021-04-12
Compactor should ignore dirty data or automatically delete corrupted data


mikechengwei:
**Thanos, Prometheus and Golang version used**:
Thanos version: v0.17.2
Prometheus:  v2.11.0
go version go1.14.4

**Object Storage Provider**:
Ceph S3.
**What happened**:
Thanos compact stuck in continuous restart loop when index file missing.

Error log:
![image](https://user-images.githubusercontent.com/13257127/114332845-93537380-9b79-11eb-8fad-7bf773c4171f.png)
List ceph data chunk folder:
![image](https://user-images.githubusercontent.com/13257127/114332941-c0a02180-9b79-11eb-80c0-bf40c4a84d5c.png)

I think we can actively skip the damaged chunk and continue to merge.

https://github.com/thanos-io/thanos/blob/4bf956deb64b2a12c8c26d28f03a6795d75ada4f/pkg/compact/compact.go#L722-739.

**What you expected to happen**:
Don't keep restarting.
**How to reproduce it (as minimally and precisely as possible)**:
When chunk data loss.

wiardvanrij:
I think we can/should do a better sanity check on these blocks. I.e. check if meta.json, index and chunks are present. Seeing the errors, it just assumes there is an index file. 
Though, I'm not so sure on how to "deal" with it as there can be numerous reasons why the index file is missing. 

I make some assumptions on the logic here but compactor sees the ULID -> downloads the entire folder -> does it's thing. In your case it fails because the index file is not there. Yet what if the index file is in the object store, but it simply failed to download all the contents? What if someone made an 'oopsie' and had the wrong permissions on a file, or what if the disk is full. Even this could happen when for example the sidecar ships its stuff but fails 'somewhere' and mistakenly misses the index. Now the block is not broken at all and could be fixed.

IMO the compactor job is pretty much irreversible. The behaviour of 'dirty data' or even 'corrupted data' should be the exception rather than a standard. Therefore my initial thought here is that we should not make assumptions and unfortunately do manual actions as sanity check before proceeding. 

Do you perhaps have a proposal on when we could define data as corrupted and 100% not fixable? Or WDYT?

mikechengwei:
I get you meaning and  agree with you the behaviour of 'dirty data' or even 'corrupted data' should't be a standard.

I think we can throw exception when meeting corrupted data but not keep container restart loop. We should ensure that the application will not crash due to some exceptions. WDYT? @wiardvanrij

[...]

PR #4072, opened 2021-04-17
When encountering damaged data, continue to compress other data.

mikechengwei:
Optimize: #4046 

## Changes

When encountering damaged data, continue to compress other data.

### DIFF
--- a/pkg/compact/compact.go
+++ b/pkg/compact/compact.go
@@ -728,7 +728,7 @@ func (cg *Group) compact(ctx context.Context, dir string, planner Planner, comp 
 		// Ensure all input blocks are valid.
 		stats, err := block.GatherIndexHealthStats(cg.logger, filepath.Join(bdir, block.IndexFilename), meta.MinTime, meta.MaxTime)
 		if err != nil {
-			level.Error(cg.logger).Log("msg", "gather index issues for block", bdir, err)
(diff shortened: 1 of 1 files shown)


pkg/compact/compact.go
```diff
@@ -728,7 +728,8 @@ func (cg *Group) compact(ctx context.Context, dir string, planner Planner, comp
 		// Ensure all input blocks are valid.
 		stats, err := block.GatherIndexHealthStats(cg.logger, filepath.Join(bdir, block.IndexFilename), meta.MinTime, meta.MaxTime)
 		if err != nil {
-			return false, ulid.ULID{}, errors.Wrapf(err, "gather index issues for block %s", bdir)
+			level.Error(cg.logger).Log("msg", "gather index issues for block ", bdir, err)
```
bwplotka, line 731: In the local scope, this fix is great, but unfortunately, if we don't deal with this and just ignore it, the next compaction will pick the same set of blocks (: 

Let's think about automated solution to this problem. Ideally while changing compactor algorithm: https://github.com/thanos-io/thanos/issues/4046#issuecomment-823373171

bwplotka (changes requested):
Thanks for this attempt, but we need to invest in different planning and mechanisms that allows this as commented as https://github.com/thanos-io/thanos/issues/4046#issuecomment-823373171 

Hope that makes sense 🤗

mikechengwei:
/

PR #4074, opened 2021-04-17
block: Avoid IgnoreDeletionMarkFilter deadlock on many errors

zecke:
The go routine feeding work into the channel could get stuck when
there are more than 2*f.concurrency and more blocks left. Make sure
to completely drain the channel and report any error that occurred.

### DIFF
--- a/pkg/block/fetcher.go
+++ b/pkg/block/fetcher.go
@@ -820,6 +820,7 @@ func (f *IgnoreDeletionMarkFilter) Filter(ctx context.Context, metas map[ulid.UL
 
 	for i := 0; i < f.concurrency; i++ {
 		eg.Go(func() error {
+			var lastErr error
 			for id := range ch {
 				m := &metadata.DeletionMark{}
 				if err := metadata.ReadMarker(ctx, f.logger, f.bkt, id.String(), m); err != nil {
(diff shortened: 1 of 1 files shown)


pracucci (approved):
Simply yet effective, the fix LGTM. Thanks! Have you checked if we can easily write a unit test to cover this regression?

bwplotka (approved):
Nice, thanks!

bwplotka:
Regression test would be amazing, yes!

kakkoyun:
Hey @zecke, friendly ping. We're going to cut a release candidate soon, is there any chance for you to push this to the finish line? I guess we are only missing some tests.

zecke:
The only test I can think of is a change detector that produces more errors than the level of concurrency. Can you think of anything else?

pracucci:
> The only test I can think of is a change detector that produces more errors than the level of concurrency. Can you think of anything else?

You could use a "fake" `objstore.Bucket` implementation which just returns errors and a number of blocks > concurrency. We expect the meta sync to deadlock (and so the test to timeout) in `main` branch while to return error in your PR. Does it make sense?

kakkoyun:
Sorry for the persistent ping @zecke. This seems like an easy win to have in the release. Do you think you can spare some time on this today? No pressure though, I'm just asking.

zecke:
Unfortunately I might not find time before Friday to add a testcase.

kakkoyun:
👍 Thanks, @zecke for letting me know.

bwplotka (approved):
That's alright. Just by looking on it, it makes total sense - good to be merged and we can add test later.

PR #4081, opened 2021-04-20
Make the ReactUI the default one

Nexucis:
This PR is switching the default UI from the old one to the ReactAPP by default.

### DIFF
--- a/pkg/ui/bucket.go
+++ b/pkg/ui/bucket.go
@@ -48,28 +48,33 @@ func NewBucketUI(logger log.Logger, label, externalPrefix, prefixHeader, uiPrefi
 }
 
 // Register registers http routes for bucket UI.
-func (b *Bucket) Register(r *route.Router, registerNewUI bool, ins extpromhttp.InstrumentationMiddleware) {
-	instrf := func(name string, next func(w http.ResponseWriter, r *http.Request)) http.HandlerFunc {
-		return ins.NewHandler(b.externalPrefix+name, http.HandlerFunc(next))
+func (b *Bucket) Register(r *route.Router, ins extpromhttp.InstrumentationMiddleware) {
+	redirectPath := "/blocks"
+	if len(b.uiPrefix) > 0 {
+		redirectPath = b.uiPrefix
 	}
-	r.WithPrefix(b.uiPrefix).Get("/", instrf("root", b.root))
-	r.WithPrefix(b.uiPrefix).Get("/static/*filepath", instrf("static", b.serveStaticAsset))
-	if registerNewUI {
-		// Make sure that "<path-prefix>/new" is redirected to "<path-prefix>/new/" and
-		// not just the naked "/new/", which would be the default behavior of the router
-		// with the "RedirectTrailingSlash" option (https://godoc.org/github.com/julienschmidt/httprouter#Router.RedirectTrailingSlash),
-		// and which breaks users with a --web.route-prefix that deviates from the path derived
-		// from the external URL.
-		r.Get("/new", func(w http.ResponseWriter, r *http.Request) {
-			http.Redirect(w, r, path.Join(GetWebPrefix(b.logger, b.externalPrefix, b.prefixHeader, r), "new")+"/", http.StatusFound)
-		})
-		r.Get("/new/*filepath", instrf("react-static", b.serveReactUI))
-	}
+	r.Get("/", func(w http.ResponseWriter, r *http.Request) {
+		http.Redirect(w, r, path.Join(GetWebPrefix(b.logger, b.externalPrefix, b.prefixHeader, r), redirectPath), http.StatusFound)
+	})
+	// Redirect the original React UI's path (under "/new") to its new path at the root.
+	r.Get("/new/*path", func(w http.ResponseWriter, r *http.Request) {
+		p := route.Param(r.Context(), "path")
+		http.Redirect(w, r, path.Join(GetWebPrefix(b.logger, b.externalPrefix, b.prefixHeader, r), strings.TrimPrefix(p, "/new"))+"?"+r.URL.RawQuery, http.StatusFound)
+	})
+
+	// here we have two routes that serve the same document. It's because it depends where do we come from.
+	// If we are coming from the new UI, it will use the first route.
+	// If we are coming from the old UI, it will use the second route.
+	r.Get(path.Join("/classic", b.uiPrefix), instrf("bucket", ins, b.bucket))
+	r.WithPrefix(b.uiPrefix).Get("/classic", instrf("bucket", ins, b.bucket))
+	r.WithPrefix(b.uiPrefix).Get("/classic/static/*filepath", instrf("static", ins, b.serveStaticAsset))
+	registerReactApp(r, ins, b.BaseUI)
 }
 
 // Handle / of bucket UIs.
-func (b *Bucket) root(w http.ResponseWriter, r *http.Request) {
-	b.executeTemplate(w, "bucket.html", GetWebPrefix(b.logger, path.Join(b.externalPrefix, strings.TrimPrefix(b.uiPrefix, "/")), b.prefixHeader, r), b)
+func (b *Bucket) bucket(w http.ResponseWriter, r *http.Request) {
+	prefix := GetWebPrefix(b.logger, path.Join(b.externalPrefix, strings.TrimPrefix(b.uiPrefix, "/")), b.prefixHeader, r)
+	b.executeTemplate(w, "bucket.html", prefix, b)
 }
 
 func (b *Bucket) Set(blocks []metadata.Meta, err error) {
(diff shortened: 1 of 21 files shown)


pkg/ui/query.go
```diff
@@ -65,38 +64,32 @@ func queryTmplFuncs() template.FuncMap {
...
-	// from the external URL.
-	r.Get("/new", func(w http.ResponseWriter, r *http.Request) {
-		http.Redirect(w, r, path.Join(GetWebPrefix(q.logger, q.externalPrefix, q.prefixHeader, r), "new")+"/graph", http.StatusFound)
+	r.Get("/classic/", func(w http.ResponseWriter, r *http.Request) {
+		http.Redirect(w, r, path.Join(GetWebPrefix(q.logger, q.externalPrefix, q.prefixHeader, r), "/classic/graph"), http.StatusFound)
 	})
-	r.Get("/new/*filepath", instrf("react-static", q.serveReactUI))
+
+	// Redirect the original React UI's path (under "/new") to its new path at the root.
+	r.Get("/new/*path", func(w http.ResponseWriter, r *http.Request) {
```
bwplotka, line 76: Why we need this? There are some hardcoded React paths for /new?

bwplotka (commented):
Nice, thanks for the work! Looks great, just one main question.

pkg/ui/query.go
Nexucis: well nothing related to the code itself it's more because the user would have in the browser history the url prefixed by `/new`.

So it's just to avoid they will have an error by autocompleting the url in their browser

bwplotka: ah. nice!

bwplotka (approved):
Looks good from outside. Have you tested it? (:

Nexucis:
yes I just had some trouble to verify if everything is working fine for the thanos-ruler. (mainly because of my setup), but hopefully everything is fine.

And yeah hopefully for me I tested it, at the first shot, the old UI wasn't working anymore :sweat_smile:

kakkoyun (approved):
LGTM AFAIT. I'd feel safer if we have a green light from @onprem as well?

onprem (commented):
Great work! Everything looks good but when testing locally I noticed that you missed the Block Viewer UI (`pkg/ui/bucket.go`). Let's update the routes there as well otherwise it'll break the block viewer.

Nexucis:
ah yeah indeed, I completely missed the bucket part. Sorry about that, I will update it. Thanks !

Nexucis:
Updated. The only big change is just there was an option to activate or not the new UI. Well like it is becoming the default one, I removed this boolean. Hope you don't mind.

pkg/ui/bucket.go
```diff
@@ -48,28 +48,33 @@ func NewBucketUI(logger log.Logger, label, externalPrefix, prefixHeader, uiPrefi
...
+		p := route.Param(r.Context(), "path")
+		http.Redirect(w, r, path.Join(GetWebPrefix(b.logger, b.externalPrefix, b.prefixHeader, r), strings.TrimPrefix(p, "/new"))+"?"+r.URL.RawQuery, http.StatusFound)
+	})
+
+	// here we have two routes that serve the same document. It's because it depends where do we come from.
+	// If we are coming from the new UI, it will use the first route.
+	// If we are coming from the old UI, it will use the second route.
+	r.Get(path.Join("/classic", b.uiPrefix), instrf("bucket", ins, b.bucket))
+	r.WithPrefix(b.uiPrefix).Get("/classic", instrf("bucket", ins, b.bucket))
+	r.WithPrefix(b.uiPrefix).Get("/classic/static/*filepath", instrf("static", ins, b.serveStaticAsset))
```
Nexucis, line 70: this part is quite ugly, but, I fixed every possible path that you could follow. So we can now move from the old UI to the new one without any issue. 

But definitely dropping the old UI will simplify a lot the way to serve the UI.

I didn't really find the good way to have the same route everywhere. It's mainly becaused the prefix of the UI is not managed in the same way on the old UI and in the UI. (At least that's what I am seing and what I understood)

Nexucis:
looks like I succeeded to break the e2e test :/ no idea how. If someone has a clue on what I broke, that would be cool

kakkoyun:
I'll run e2e tests locally to confirm if it's a flake or a real case.

kakkoyun:
@Nexucis Could you please rebase one last time?

Nexucis:
@kakkoyun I rebased the PR. But I still have the errors. Still strange, I will have to take a deeper look at it to understand what I broke.

kakkoyun:
@Nexucis Let me know if anyways I can help. I want to have this in v0.20.0 :)

Nexucis:
@kakkoyun well if you are able to find more quickly than me why it doesn't work I won't say no :D.

Nexucis:
ah that's interesting, when I'm moving back the code in charge of the UI routing for the bucket, the errors relative to the bucket disappear

pkg/ui/bucket.go
```diff
@@ -48,30 +48,44 @@ func NewBucketUI(logger log.Logger, label, externalPrefix, prefixHeader, uiPrefi
...
+		r.Get("/new/*path", func(w http.ResponseWriter, r *http.Request) {
+			p := route.Param(r.Context(), "path")
+			http.Redirect(w, r, path.Join(GetWebPrefix(b.logger, b.externalPrefix, b.prefixHeader, r), strings.TrimPrefix(p, "/new"))+"?"+r.URL.RawQuery, http.StatusFound)
+		})
+
+		// here we have two routes that serve the same document. It's because it depends where do we come from.
+		// If we are coming from the new UI, it will use the first route.
+		// If we are coming from the old UI, it will use the second route.
+		r.Get(path.Join("/classic", b.uiPrefix), instrf("bucket", ins, b.bucket))
+		r.WithPrefix(b.uiPrefix).Get("/classic", instrf("bucket", ins, b.bucket))
```
onprem, line 69: `panic: a handle is already registered for path '/classic'`

I got this panic when running Block Viewer locally. If the `uiPrefix` is empty string, this will try to register `/classic` again, which causes a panic.

onprem:
> The only big change is just there was an option to activate or not the new UI. Well like it is becoming the default one, I removed this boolean.

@Nexucis At that time I didn't really get what you meant but I got it now. What you removed was a boolean governing if the routes will be registered or not. This was because in compactor the Block Viewer UI is actually registered two times, this is an artifact from the old UI times. So, the old UI needed to be registered twice, but the new UI only needs to be registered only once. That's why that boolean was created and I think we need to keep that (at least until we get rid of the classic UI completely).

pkg/ui/bucket.go
onprem: Ah, I guess the `uiPrefix` is also a block viewer specific artifact as well, stemming from the classic UI.

Nexucis: yeah ... I guess I have to put a condition about the uiPrefix like I did a bit above

Nexucis:
thanks a lot @onprem for the tips ! I will try that

onprem:
@Nexucis I started playing with this and there is a lot of work we need to do because there can be multiple instances of Block Viewer in a single component (compactor has two separate classic UIs running on /loaded and /global).

Take a look at this https://github.com/onprem/thanos/commit/1862dce79fc8a50ce3b14dd6dd6af6d95f1e5edd I made some changes on top of your work (modulo the last revert commit) to make this work.

Nexucis:
> I started playing with this and found out that there is a lot of work we need to do because there can be multiple instances of Block Viewer in a single component (compactor has two separate classic UIs running on /loaded and /global).

By looking about your changes, I'm feeling you fix the issues. Or do you see others problems ? 

Thanks btw for code you shared :)

onprem:
> Or do you see others problems ?

Nope, everything else looked perfectly fine to me.

kakkoyun:
@Nexucis I guess we have solution now. Looking forward to it.

Nexucis:
@kakkoyun yes totally :) thanks to @onprem ! I just have to merge what he did with my code and we will be good.

Nexucis:
alright I fixed the bucket. There is something odd in the bucket somehow if the template {{ uiPrefix }} is used in bucket.html and bucket_menu.html, then it is replaced by an empty string whatever is put in `tmplFuncs`.

But the `{{ pathPrefix }}` is correctly replaced so I injected the path `/classic/...` in the `pathPrefix` instead. Hope it is fine.

So if we forget about this strange things, it should be fine now.

Nexucis:
by looking at the e2e test error, it doesn't look like it is related to what I did, but I'm maybe wrong here :D. So many things are strange when changing the way to serve the UI.

Nexucis:
and by looking at another PR (https://github.com/thanos-io/thanos/pull/4094) , it looks like the e2e is not working well too with the same error than here. Sooooo I would say all good.

kakkoyun:
The Github CI seems like broken, failures are not related, so mering this one.

PR #4086, opened 2021-04-21
Add Protofile for info service

hitanshu-mehta:
This is the protofile for the Info service we are planning to add as part of [unified endpoint discovery](https://thanos.io/tip/proposals/202101_endpoint_discovery.md/) proposal.

### Open questions
1. Can there be disjoint labelsets (i.e. different external labels for different APIs) for the components exposing multiple APIs?
2. Is there any other metadata for any API?

### DIFF
--- a/pkg/info/infopb/rpc.proto
+++ b/pkg/info/infopb/rpc.proto
@@ -1,0 +1,64 @@
+// Copyright (c) The Thanos Authors.
+// Licensed under the Apache License 2.0.
+
+syntax = "proto3";
+package thanos;
+
+import "store/labelpb/types.proto";
+import "store/storepb/rpc.proto";
+import "gogoproto/gogo.proto";
+
+option go_package = "infopb";
+
+option (gogoproto.sizer_all) = true;
+option (gogoproto.marshaler_all) = true;
+option (gogoproto.unmarshaler_all) = true;
+option (gogoproto.goproto_getters_all) = false;
+
+// Do not generate XXX fields to reduce memory footprint and opening a door
+// for zero-copy casts to/from prometheus data types.
+option (gogoproto.goproto_unkeyed_all) = false;
+option (gogoproto.goproto_unrecognized_all) = false;
+option (gogoproto.goproto_sizecache_all) = false;
+
+// Info represents the API that is responsible for gather metadata about the all APIs supported by the component.
+service Info {
+    // Info returns the metadata (Eg. LabelSets, Min/Max time) about all the APIs the component supports.
+    rpc Info(InfoRequest) returns (InfoResponse);
+}
+
+message InfoRequest {}
+
+message InfoResponse {
+    repeated ZLabelSet label_sets = 1 [(gogoproto.nullable) = false];
+    StoreType storeType  = 2; // Maybe rename to ComponentType
+    
+    StoreInfo storeInfo = 3;
+    RulesInfo rulesInfo = 4;
+    MetadataInfo metadataInfo = 5;
+    ExemplarsInfo exemplarsInfo = 6;
+    
+    // Deprecated. Use label_sets instead.
+    repeated Label labels = 7 [(gogoproto.nullable) = false, (gogoproto.customtype) = "github.com/thanos-io/thanos/pkg/store/labelpb.ZLabel"];
+
+}
+
+message StoreInfo {
+    int64 min_time = 1;
+    int64 max_time = 2;
+}
+
+message RulesInfo {
+    int64 min_time = 1;
+    int64 max_time = 2;
+}
+
+message MetadataInfo {
+    int64 min_time = 1;
+    int64 max_time = 2;
+}
+
+message ExemplarsInfo {
+    int64 min_time = 1;
+    int64 max_time = 2;
+}
--- a/pkg/info/infopb/rpc.proto
+++ b/pkg/info/infopb/rpc.proto
@@ -5,7 +5,6 @@ syntax = "proto3";
 package thanos;
 
 import "store/labelpb/types.proto";
-import "store/storepb/rpc.proto";
 import "gogoproto/gogo.proto";
 
 option go_package = "infopb";
@@ -29,18 +28,25 @@ service Info {
 
 message InfoRequest {}
 
+enum ComponentType {
+    UNKNOWN = 0;
+    QUERY = 1;
+    RULE = 2;
+    SIDECAR = 3;
+    STORE = 4;
+    RECEIVE = 5;
+    // DEBUG represents some debug StoreAPI components e.g. thanos tools store-api-serve.
+    DEBUG = 6;
+  }
+
 message InfoResponse {
     repeated ZLabelSet label_sets = 1 [(gogoproto.nullable) = false];
-    StoreType storeType  = 2; // Maybe rename to ComponentType
+    ComponentType ComponentType  = 2;
     
     StoreInfo storeInfo = 3;
     RulesInfo rulesInfo = 4;
     MetadataInfo metadataInfo = 5;
     ExemplarsInfo exemplarsInfo = 6;
-    
-    // Deprecated. Use label_sets instead.
-    repeated Label labels = 7 [(gogoproto.nullable) = false, (gogoproto.customtype) = "github.com/thanos-io/thanos/pkg/store/labelpb.ZLabel"];
-
 }
 
 message StoreInfo {
--- a/pkg/info/infopb/rpc.proto
+++ b/pkg/info/infopb/rpc.proto
@@ -32,14 +32,14 @@ message InfoResponse {
     repeated ZLabelSet label_sets = 1 [(gogoproto.nullable) = false];
     string ComponentType          = 2;
     
-    StoreInfo store               = 3;
+    StoresInfo stores             = 3;
     RulesInfo rules               = 4;
     MetadataInfo metadata         = 5;
     TargetsInfo targets           = 6;
     ExemplarsInfo exemplars       = 7;
 }
 
-message StoreInfo {
+message StoresInfo {
     int64 min_time = 1;
     int64 max_time = 2;
 }
--- a/pkg/info/infopb/rpc.proto
+++ b/pkg/info/infopb/rpc.proto
@@ -30,15 +30,27 @@ message InfoRequest {}
 
 message InfoResponse {
     repeated ZLabelSet label_sets = 1 [(gogoproto.nullable) = false];
-    string ComponentType  = 2;
+    string ComponentType          = 2;
     
-    StoreInfo store = 3;
-    ExemplarsInfo exemplars = 4;
+    StoreInfo store               = 3;
+    RulesInfo rules               = 4;
+    MetadataInfo metadata         = 5;
+    TargetsInfo targets           = 6;
+    ExemplarsInfo exemplars       = 7;
 }
 
 message StoreInfo {
     int64 min_time = 1;
     int64 max_time = 2;
+}
+
+message RulesInfo {
+}
+
+message MetadataInfo {
+}
+
+message TargetsInfo {
 }
 
 message ExemplarsInfo {
--- a/pkg/info/infopb/rpc.proto
+++ b/pkg/info/infopb/rpc.proto
@@ -20,7 +20,7 @@ option (gogoproto.goproto_unkeyed_all) = false;
 option (gogoproto.goproto_unrecognized_all) = false;
 option (gogoproto.goproto_sizecache_all) = false;
 
-// Info represents the API that is responsible for gather metadata about the all APIs supported by the component.
+// Info represents the API that is responsible for gathering metadata about the all APIs supported by the component.
 service Info {
     // Info returns the metadata (Eg. LabelSets, Min/Max time) about all the APIs the component supports.
     rpc Info(InfoRequest) returns (InfoResponse);
--- a/pkg/info/infopb/rpc.proto
+++ b/pkg/info/infopb/rpc.proto
@@ -44,10 +44,7 @@ message InfoResponse {
     ComponentType ComponentType  = 2;
     
     StoreInfo storeInfo = 3;
-    RulesInfo rulesInfo = 4;
-    MetadataInfo metadataInfo = 5;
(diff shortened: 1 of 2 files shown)


pkg/info/infopb/rpc.proto
```diff
@@ -0,0 +1,76 @@
...
+}
+
+message StoreInfo {
+    int64 min_time = 1;
+    int64 max_time = 2;
+}
+
+message RulesInfo {
+    int64 min_time = 1;
+    int64 max_time = 2;
```
yeya24, line 60: What does min max time mean for rules, metadata and targets info?

hitanshu-mehta: Yes. It doesn't make any sense 🤦🏻‍♂️. We always use current rules.

pkg/info/infopb/rpc.proto
```diff
@@ -0,0 +1,70 @@
...
+    RULE = 2;
+    SIDECAR = 3;
+    STORE = 4;
+    RECEIVE = 5;
+    // DEBUG represents some debug StoreAPI components e.g. thanos tools store-api-serve.
+    DEBUG = 6;
+  }
+
+message InfoResponse {
+    repeated ZLabelSet label_sets = 1 [(gogoproto.nullable) = false];
```
yashrsharma44, line 43: Are we using gogoproto? This is not maintained [currently](https://github.com/gogo/protobuf/issues/691), so it would be nice to remove this dependency

bwplotka: Yes, we need to do it, once switching to grpc middleware v2, can you help with that @yashrsharma44 ? (: 

Otherwise, let's focus on the main work here, and remove gogoproto in dedicated PR (:

yashrsharma44: Sure, we can take care of that in another PR 👍

pkg/info/infopb/rpc.proto
```diff
@@ -0,0 +1,70 @@
...
+
+message MetadataInfo {
+}
+
+message ExemplarsInfo {
+    int64 min_time = 1;
+    int64 max_time = 2;
+}
+
+message TargetsInfo {
```
yeya24, line 69: Do we still need `RulesInfo`, `MetadataInfo` and `TargetsInfo`, even though they are just empty structs? Do we want to change them to bools or keep structs for future usage?

hitanshu-mehta: I think we can remove them, because we are going to use gRPC reflection as mentioned in [proposal](https://thanos.io/tip/proposals/202101_endpoint_discovery.md/)  so we will be able to know about these services even if we don't add struct of boolean flags for them. WDYT?

pkg/info/infopb/rpc.proto
```diff
@@ -0,0 +1,58 @@
...
+    RECEIVE = 5;
+    // DEBUG represents some debug StoreAPI components e.g. thanos tools store-api-serve.
+    DEBUG = 6;
+  }
+
+message InfoResponse {
+    repeated ZLabelSet label_sets = 1 [(gogoproto.nullable) = false];
+    ComponentType ComponentType  = 2;
+    
+    StoreInfo storeInfo = 3;
```
bwplotka, line 46:
```suggestion
    StoreInfo store = 3;
```

pkg/info/infopb/rpc.proto
```diff
@@ -0,0 +1,58 @@
...
+    ExemplarsInfo exemplarsInfo = 4;
```
bwplotka, line 47:
```suggestion
    ExemplarsInfo exemplars = 4;
```

bwplotka (approved):
Nice, I think this is a good start TBH, just small suggestion otherwise LGTM.

We can add later info types later on

bwplotka:
I would love @brancz view too and other maintainers before merging though.

squat:
Yes, it would be great to have your inputs 🙇 Prem and I are reviewing in-the-loop during our 1-1s

squat (dismissed):

yeya24 (dismissed):
LGTM

bwplotka:
I pinged @brancz he created reminder for it

hitanshu-mehta:
Thank you everyone for the quick review 😃  !

I have some questions about min and max time for exemplars.

- How are we going to use the min and max time of exemplars?
- Will it be same as the min and max time of store API, if not how can we calculate it?

If we are not going the use them then we can remove them for now. WDYT?

cc: @onprem @squat

yeya24:
Good question.
1. min_time and max_time can be used for filtering. For example, a federated exemplars query against multiple stores that implement Exemplar APIs. The query request contains start and end time so we can filter out the stores by checking their exemplars time range.

2. It is not the same. It is independent I think. And unfortunately, Prometheus doesn't expose any metadata about exemplars. IMO maybe we can add something upstream.

hitanshu-mehta:
Thank you @yeya24. That clears things up! 

> Prometheus doesn't expose any metadata about exemplars.

I guess that should not be a blocker for adding Info service for now 🤔 . At the same time, we can open an issue at Prometheus.

pkg/info/infopb/rpc.proto
```diff
@@ -0,0 +1,58 @@
...
+    SIDECAR = 3;
+    STORE = 4;
+    RECEIVE = 5;
+    // DEBUG represents some debug StoreAPI components e.g. thanos tools store-api-serve.
+    DEBUG = 6;
+  }
+
+message InfoResponse {
+    repeated ZLabelSet label_sets = 1 [(gogoproto.nullable) = false];
+    ComponentType ComponentType  = 2;
```
brancz, line 44: I expected this not to be the component type, but rather the APIs that this server serves, the type can be ambiguous, especially as we already have community extensions that don't strictly fall under any component, eg. the OpenTSDB implementation of the StoreAPI.

brancz:
I think the way I would personally structure it is by having a message type for each type of API that can be served. That way, by just passing a configuration struct for that API we automatically know that API is being served, and it gives us the opportunity to add any kind of configuration/info per API that we want in the future (even if it might be empty today). What was started with StoreInfo and ExemplarInfo goes in the right direction (off the top of my head TargetsAPI and RulesAPI should also be included as part of this work).

pkg/info/infopb/rpc.proto
hitanshu-mehta: We can discover APIs supported by the server using gRPC reflection and retrieve metadata by calling the `Info` method of this service. I have added `ComponentType` to be more consistent with the [Info](https://github.com/thanos-io/thanos/blob/7d118f416be98cf502b9b5e0df4e98b37478f3f3/pkg/store/storepb/rpc.proto#L79) method we already have in StoreAPI.
TBH, not sure about the compatibility of this Info service with community extensions.

hitanshu-mehta:
Yes, initially that was the plan but currently, we don't have any metadata for `Rules`, `Targets`, and `Metadata` APIs, so those structs were empty. Also, we are going to use gRPC reflection as mentioned in [proposal](https://github.com/thanos-io/thanos/blob/main/docs/proposals/202101_endpoint_discovery.md#solution), so the discovery of these services can be done without adding any struct or boolean flags for them.

pkg/info/infopb/rpc.proto
bwplotka: Yea, I think it's a misunderstanding a bit. The component type is really for debugging purposes, we (try) to not rely on this information. Maybe it should be then string in next iteration to indicate that it should not be used for discoverability? We use it for the Store UI page and some compatibility changes around metadata / label advertising in store GW. 🤔

bwplotka (commented):
Yea I think what we have @brancz is what you explain, just it might be confusing with component Type which is really inherited from old Info and used for "naming" (like `User-Agent` for server or something)

pkg/info/infopb/rpc.proto
onprem: Just to clarify a bit more on why we plan to use reflection for discovery. If we use reflection we will have a single source of truth. For example, if a gRPC endpoint exposes `thanos.StoreAPI` only then the Querier will add it to the stores list and otherwise it won't. Another advantage that builds up on this is that if we add some Service to a component, it will be discovered automatically by the Querier without additional changes. For example, if we add support for exemplarsAPI in Receive, we will have to just do that (implement the API) and the Querier will pick this up automatically and add the Receive endpoint to `exemplars` list.

@brancz @bwplotka I am interested in your opinions on this. Do you think we should go with this approach or do you prefer exposing this information as part of the Info Service and not using reflection.

squat:
It sounds like overall we agree but we need to clear up some confusion. Maybe we are just talking past each other. 

This PR proposes that the `InfoResponse` message will have nested message types for every API that we may want to provide information about. It does not currently have embedded messages for APIs that don't provide any info today, e.g. Targets, but as soon as we have a use for some metadata, we can add it to a `TargetsInfo` message type and embed a `TargetsInfo` message in the `InfoResponse` message. 

Frederic, are you saying that we should include a empty `TargetsInfo` message to signal that we implement the Targets API even if the API does not yet have any metadata? In this case we could use gogoproto nullable fields to differentiate between an empty message and no support for that API.

pkg/info/infopb/rpc.proto
brancz: I see, thanks for clarifying everyone. I think what @bwplotka said, to just change the component type to be a string instead of an enum would be best then, that way it can serve the debugging purpose without being constraining on the ecosystem.

brancz:
> Frederic, are you saying that we should include a empty TargetsInfo message to signal that we implement the Targets API even if the API does not yet have any metadata? In this case we could use gogoproto nullable fields to differentiate between an empty message and no support for that API.

Just thinking out loud, but yeah I think this would be nice, as that way we wouldn't have to use gRPC reflection at all, and have all the discovery about the capabilities of a gRPC server in one `Info` call.

bwplotka:
Sounds good, I don't how gRPC reflection works but sounds unsafe/extra call. Btw, we cannot rely on Gogoproto anymore as it's deprecated


We can try to move out of this in another PR, but there should be similar option in normal proto

pkg/info/infopb/rpc.proto
```diff
@@ -0,0 +1,59 @@
...
+    rpc Info(InfoRequest) returns (InfoResponse);
+}
+
+message InfoRequest {}
+
+message InfoResponse {
+    repeated ZLabelSet label_sets = 1 [(gogoproto.nullable) = false];
+    string ComponentType          = 2;
+    
+    StoreInfo store               = 3;
```
bwplotka, line 35: Can we double check how those APIs are called? 

We have singular vs plural names: `StoreInfo` vs `RulesInfo` etc.. can we have consistency? `StoresInfo` or `RuleInfo`? 🤔

bwplotka (commented):
Nice, Looks good, one nit only.

cc @brancz

bwplotka (approved):

pkg/info/infopb/rpc.proto
hitanshu-mehta: I checked other proto files in Thanos. All of them are using plural names. Should we go with plural?

pkg/info/infopb/rpc.proto
```diff
@@ -0,0 +1,59 @@
...
+    rpc Info(InfoRequest) returns (InfoResponse);
+}
+
+message InfoRequest {}
+
+message InfoResponse {
+    repeated ZLabelSet label_sets = 1 [(gogoproto.nullable) = false];
+    string ComponentType          = 2;
+    
+    StoresInfo stores             = 3;
```
bwplotka, line 35: Can we add comments to the new API? All public comments should expose some commentary especially specification like this.

Please note we follows protobuf code style, so field comments should start with `///`

bwplotka, line 35: Also sorry for confusion, we suggested to check other APIs to make sure we the naming makes sense. We definitely have API called `Store`  not `Stores` so that one is wrong (`StoreInfo` makes more sense same with variable, there is single Store implementation there).

For others let's check but I see:

```
/// Rules represents API that is responsible for gathering rules and their statuses.
service Rules {
    /// Rules has info for all rules.
    /// Returned rules are expected to include external labels.
    rpc Rules(RulesRequest) returns (stream RulesResponse);
}
/// Targets represents API that is responsible for gathering targets and their states.
service Targets {
    /// Targets has info for all targets.
    /// Returned targets are expected to include external labels.
    rpc Targets (TargetsRequest) returns (stream TargetsResponse);
}
/// Exemplars represents API that is responsible for gathering exemplars and their states.
service Exemplars {
    /// Exemplars has info for all exemplars.
    /// Returned exemplars are expected to include external labels.
    rpc Exemplars(ExemplarsRequest) returns (stream ExemplarsResponse);
}
service Metadata {
  rpc MetricMetadata(MetricMetadataRequest) returns (stream MetricMetadataResponse);
}
```

I guess others then makes sense, so... can we revert to what we had? Sorry for confusion 🤦🏽

pkg/info/infopb/rpc.proto
```diff
@@ -0,0 +1,59 @@
...
+    RulesInfo rules               = 4;
+    MetadataInfo metadata         = 5;
```
bwplotka, line 37:
```suggestion
    MetricMetadataInfo metric_metadata         = 5;
```

bwplotka (approved):
Sorry for confusion, I elaborated what I meant (: Thanks.

Let's add examples, then LGTM!

pkg/info/infopb/rpc.proto
hitanshu-mehta: No worries! Thank you for checking out APIs. I will revert back to `StoreInfo`

hitanshu-mehta: Yes sure, I will comment the proto file properly.

bwplotka (approved):
Thanks. 

LGTM! I think we can merge this and iterate over this proto if needed, before we migrate all clients/impl to it.

vovanloc0798 (approved):

onprem (approved):
LGTM!

squat (approved):

onprem:
:tada:

PR #4094, opened 2021-04-22
Upgrade Prometheus & Alertmanager

pracucci:

## Changes

We've an use case in Cortex for which we need to upgrade Prometheus and the Alertmanager. Some breaking changes were introduced in the HTTP client (from `prometheus/common`) so we need to upgrade it in Thanos too.

What's the sentiment if we do such upgrade in Thanos as well?

## Verification

_Existing tests_


kakkoyun (dismissed):
LGTM 🚀

kakkoyun:
The Github CI seems like broken, failures are not related, so mering this one.

kakkoyun:
The Github CI seems broken and failures are not related, in any case I'm going to run the tests locally.

yeya24 (dismissed):

kakkoyun (approved):
🎉 🎉 🎉

PR #3937, opened 2021-03-16
store: reduce memory footprint for chunks queries

krya-kryak:
* [x] I added CHANGELOG entry for this change.
* [ ] Change is not relevant to the end user.

## Changes

* Reduce memory footprint for chunks queries by only keeping in memory those chunks and subchunks requested, throwing away all the unnecessary bits; processed chunks are saved into moderately-sized slabs allocated on demand, so no more than one slab is wasted needlessly.
* Fixes for two data races found during work on the main part.

## Verification
A `DownsampledBlockSeries` benchmark was added. Note the reduction in allocated memory amount.
```
$ ~/go/bin/benchstat main_with_new_benchmark.txt slab_benchmark.txt
name                                                                       old time/op    new time/op    delta
BucketSeries/1000000SeriesWith1Samples/1of1000000-8                          90.9ms ± 2%    96.1ms ± 5%   +5.71%  (p=0.008 n=5+5)
BucketSeries/1000000SeriesWith1Samples/10of1000000-8                         90.4ms ± 2%    93.6ms ± 6%   +3.57%  (p=0.032 n=5+5)
BucketSeries/1000000SeriesWith1Samples/1000000of1000000-8                     1.10s ± 7%     1.02s ± 5%   -7.55%  (p=0.032 n=5+5)
BucketSeries/100000SeriesWith100Samples/1of10000000-8                        6.53ms ± 1%    6.50ms ± 1%     ~     (p=1.000 n=5+5)
BucketSeries/100000SeriesWith100Samples/100of10000000-8                      6.47ms ± 0%    6.48ms ± 1%     ~     (p=1.000 n=5+5)
BucketSeries/100000SeriesWith100Samples/10000000of10000000-8                 92.4ms ± 8%    91.9ms ± 8%     ~     (p=1.000 n=5+5)
BucketSeries/1SeriesWith10000000Samples/1of10000000-8                         224µs ± 1%     229µs ± 0%   +1.97%  (p=0.008 n=5+5)
BucketSeries/1SeriesWith10000000Samples/100of10000000-8                       225µs ± 2%     228µs ± 0%     ~     (p=0.151 n=5+5)
BucketSeries/1SeriesWith10000000Samples/10000000of10000000-8                 24.9ms ± 5%    19.5ms ±10%  -21.54%  (p=0.008 n=5+5)
BlockSeries/concurrency:_1-8                                                 11.2ms ± 1%    10.9ms ± 4%   -2.85%  (p=0.016 n=5+5)
[...]

PR #4095, opened 2021-04-22
Fix race condition in pool.BucketedBytes

pracucci:

## Changes

We're reviewing #3937 and we found a race condition introduced in `pool.BucketedBytes.Put()`. If the input buffer is larger than the largest bucket size, then `p.usedTotal` is modified outside the lock.

In this PR I'm rolling back the change done in #3937 which I think was unnecessary. Before the lock only `p.sizes` and `p.buckets[i]` are accessed: `p.sizes` doesn't change and `p.buckets[i]` (which is a `sync.Pool`) is thread safe.

## Verification

N/A

### DIFF
--- a/pkg/pool/pool.go
+++ b/pkg/pool/pool.go
@@ -112,11 +112,12 @@ func (p *BucketedBytes) Put(b *[]byte) {
 			continue
 		}
 		*b = (*b)[:0]
-		p.mtx.Lock()
-		defer p.mtx.Unlock()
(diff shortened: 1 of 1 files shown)


pstibrany (approved):
LGTM. Thanks.

kakkoyun (approved):
LGTM 💯 Thanks for fixing it.

kakkoyun:
The Github CI seems broken and failures are not related, so merging this one.

Issue #3975, opened 2021-03-25
sidecar support https and basic auth access to Prometheus


liangli1213:
Since Prometheus version 2.24, we can configure basic authentication and TLS for Prometheus, in my case, the Prometheus provide service directly to the whole internal network, so we enable login credential and TLS (we don't want everyone access Prometheus and username/password transmitted with plain text), and meanwhile we need configure thanos sidecar in same Prometheus (we have multiple Prometheus).

The requirement is:

1. Provide an option to set up username/password used for Prometheus's login
2. Provide an option like --insecure_skip_verify to skip verify expired certificate

yeya24:
We definitely need this. Help wanted.

Namanl2001:
Can I take this up?

Zophar78:
I'm also really looking forward on such feature! ;)

binhvcom:
+1

yeya24:
@Namanl2001 Yes, please go ahead.

bwplotka:
Acceptance Criteria: 
* User can configure Auth and TLS option for internal Prometheus client that sidecar uses. Ideally we can reuse other Auth and TLS options we have e.g for Alertmanager

bwplotka:
We could use similar configuration to: https://thanos.io/tip/components/rule.md/#configuration

someshkoli:
@Namanl2001 you still working on this ? I have a made a poc for this out of curiosity :thinking: LMK :raised_hands:

Namanl2001:
@someshkoli I was just about to start with this, but if you've already drafted some changes, you can continue, and I'll try to help you with that. Thanks

someshkoli:
> @someshkoli I was just about to start with this, but if you've already drafted some changes, you can continue, and I'll try to help you with that. Thanks

Yeah would definitely need some help. Lets take this to PR

someshkoli:
Still valid :eyes: #4104 requires another round of review

yeya24:
Still valid and we have #4612 now.

yngwiewang:
What's the format of the yaml file? I tried this but it did not work:
```
prometheus:
- http_config:
    basic_auth:
      username: admin
      password: admin
```

PR #4104, opened 2021-04-24
Adds authorization for sidecar-promclient

someshkoli:

## Changes
Added new flag `prometheus.http-client` to sidecar to pass http client config via file or string(similar to [alertmanager](https://thanos.io/tip/components/rule.md/#configuration)). Using this flag an instance of `thanoshttp.Client` is created and is passed to `promclient.NewClient`  method which uses this client as base client instead of whole new client which then makes authorized api calls.  
The same client is passed to reloader config which then makes authorized api calls to prom api.

There are some code duplication which has been duplicated from prometheus which can be easily removed by exposing one of the functions in prometheus `http_config`
:raised_hands: 
Close: #3975
## Verification
Added a e2e test:
Checks if query is able to query when prom uses auth


pkg/promclient/promclient.go
```diff
@@ -800,3 +799,19 @@ func (c *Client) TargetsInGRPC(ctx context.Context, base *url.URL, stateTargets
...
+	MaxIdleConns        int
+	rt                  http.RoundTripper
+}
+
+func (r RoundTripperWithTransportOptions) RoundTrip(req *http.Request) (*http.Response, error) {
+	req2 := new(http.Request)
+	*req2 = *req
+	req2.Header = make(http.Header)
+	for k, s := range req.Header {
+		req2.Header[k] = s
```
someshkoli, line 814: Needs to be removed

cmd/thanos/sidecar.go
```diff
@@ -212,10 +231,13 @@ func runSidecar(
 	}
 
 	{
+		// todo(someshkoli): Not sure how to properly implement RoundTripper for these properties or
```
someshkoli, line 234: @Namanl2001 please have a look at this :thinking: I guess there's no clean way to set these properties. (might need to rewrite auth for promclient or make some changes thanoshttp for custom transporter)

wiardvanrij: I think we need some 'generic' transporter that we can re-use everywhere. I had the same 'issue': https://github.com/thanos-io/thanos/pull/3970/files/e34e60e2848f2233809d67d5398151b670341ea1#diff-95c18d923c79bf32e4f8b3370d9a925aa910643559d351fab6bc307691c8f672R119 

The thanoshttp has one func which is:
```
// NewTransport creates a new http.Transport with default settings.
```
I guess we could add an other function that has httpConfig as argument for re-usability in the entire project.

someshkoli: > I think we need some 'generic' transporter that we can re-use everywhere. I had the same 'issue': https://github.com/thanos-io/thanos/pull/3970/files/e34e60e2848f2233809d67d5398151b670341ea1#diff-95c18d923c79bf32e4f8b3370d9a925aa910643559d351fab6bc307691c8f672R119
> 
> The thanoshttp has one func which is:
> 
> ```
> // NewTransport creates a new http.Transport with default settings.
> ```
> 
> I guess we could add an other function that has httpConfig as argument for re-usability in the entire project.

Agreed, this is being used at 2-3 i guess and making a generic one can improve reusablity. @wiardvanrij I hope the overall idea in implementation is correct ? I can move forward with fixing this transporter issue ?

wiardvanrij: I'm not the right person to make a definitive review. I miss the exact technical knowledge on it. However I was at the meeting and I think you made what was discussed. Also it looks logical and makes sense. 

You could always wait for a review if you think you might invest a lot of time into a feature you are unsure about. However if you have a good feeling about it, go make awesome stuff and continue 👍

someshkoli: Alright thanks

pkg/promclient/promclient.go
```diff
@@ -86,16 +86,16 @@ func NewClient(c HTTPClient, logger log.Logger, userAgent string) *Client {
...
 	return NewWithTracingClient(
 		log.NewNopLogger(),
+		http.Client{},
 		"",
 	)
 }
 
 // NewWithTracingClient returns client with tracing tripperware.
-func NewWithTracingClient(logger log.Logger, userAgent string) *Client {
+func NewWithTracingClient(logger log.Logger, httpClient http.Client, userAgent string) *Client {
```
bwplotka, line 95:
```suggestion
func NewWithTracingClient(logger log.Logger, httpClient *http.Client, userAgent string) *Client {
```

cmd/thanos/sidecar.go
bwplotka: I think this transport and connection flags 

```
conf.connection.maxIdleConnsPerHost
conf.connection.maxIdleConns
```

Are colliding with out `promethues.http-config` flag. The idea would be to unify those? E.g Deprecating connection flags if that makes sense? Or reusing them? 

The goal is to have one way of configuring HTTP configuration.

bwplotka: Let's remove

>   --receive.connection-pool-size=RECEIVE.CONNECTION-POOL-SIZE
                                 Controls the http MaxIdleConns. Default is 0,
                                 which is unlimited
      --receive.connection-pool-size-per-host=100

And use our prometheus flags.

someshkoli:
This implementation helps in providing a generic way to create http clients used across. There are still many instances where directly http library is used to create client which can be replaced with thanos http client instance and can be configured flexibly.
Also solves the issue mentined by @wiardvanrij  is [here](https://github.com/thanos-io/thanos/pull/4104#discussion_r620998741).

pkg/http/http.go
```diff
@@ -63,6 +80,106 @@ func (b BasicAuth) IsZero() bool {
...
+	}
+
+	newRT := func(tlsConfig *tls.Config) (http.RoundTripper, error) {
+		var rt http.RoundTripper = &http.Transport{
+			Proxy:               http.ProxyURL(cfg.ProxyURL.URL),
+			MaxIdleConns:        opts.maxIdleConns,
+			MaxIdleConnsPerHost: opts.maxIdleConnsPerHost,
+			DisableKeepAlives:   true,
+			TLSClientConfig:     tlsConfig,
+			DisableCompression:  true,
```
wiardvanrij, line 125: I think we default to this because there was 'something with it' for Thanos side (_citation needed_). As this is more a generic package we should default to `false` and explicitly set this too `true` on functionalities via arguments that require this to be disabled. 

Tl;dr IMO:  set 'defaults' to something that would be defaults for _all_ packages, and override different cases when needed. Not the other way around. :)

pkg/http/http.go
```diff
@@ -63,6 +80,106 @@ func (b BasicAuth) IsZero() bool {
...
+// NewRoundTripperFromConfig returns a new HTTP RoundTripper configured for the
+// given http.HTTPClientConfig and http.HTTPClientOption.
+func NewRoundTripperFromConfig(cfg config_util.HTTPClientConfig, name string, optFuncs ...HTTPClientOption) (http.RoundTripper, error) {
+	opts := defaultHTTPClientOptions
+	for _, f := range optFuncs {
+		f(&opts)
+	}
+
+	newRT := func(tlsConfig *tls.Config) (http.RoundTripper, error) {
+		var rt http.RoundTripper = &http.Transport{
```
wiardvanrij, line 119: This could be that I'm not looking correctly but aren't we missing certain options? Like `InsecureSkipVerify ` and `ResponseHeaderTimeout` ?

someshkoli: > This could be that I'm not looking correctly but aren't we missing certain options? Like `InsecureSkipVerify ` and `ResponseHeaderTimeout` ?

ResponseHeaderTimeout yes, but AFAIR `InsecureSkipVerify` is handled by tls ?

cmd/thanos/config.go
```diff
@@ -69,24 +70,10 @@ func (pc *prometheusConfig) registerFlag(cmd extkingpin.FlagClause) *prometheusC
...
+	pc.httpClient = extkingpin.RegisterHTTPConfigFlags(cmd)
 	return pc
 }
 
-type connConfig struct {
-	maxIdleConns        int
-	maxIdleConnsPerHost int
-}
-
-func (cc *connConfig) registerFlag(cmd extkingpin.FlagClause) *connConfig {
```
bwplotka, line 80: Let's make sure we add this move to the changelog

pkg/http/http.go
```diff
@@ -5,24 +5,38 @@
...
 	"github.com/prometheus/prometheus/discovery/file"
 	"github.com/prometheus/prometheus/discovery/targetgroup"
+	"golang.org/x/net/http2"
 	"gopkg.in/yaml.v2"
 
 	"github.com/thanos-io/thanos/pkg/discovery/cache"
 )
 
+var defaultHTTPClientOptions = httpClientOptions{
+	maxIdleConns:        0,
```
bwplotka, line 36: 0 feels wrong, see default Go impl:

```
// DefaultTransport is the default implementation of Transport and is
// used by DefaultClient. It establishes network connections as needed
// and caches them for reuse by subsequent calls. It uses HTTP proxies
// as directed by the $HTTP_PROXY and $NO_PROXY (or $http_proxy and
// $no_proxy) environment variables.
var DefaultTransport RoundTripper = &Transport{
	Proxy: ProxyFromEnvironment,
	DialContext: (&net.Dialer{
		Timeout:   30 * time.Second,
		KeepAlive: 30 * time.Second,
		DualStack: true,
	}).DialContext,
	ForceAttemptHTTP2:     true,
	MaxIdleConns:          100,
	IdleConnTimeout:       90 * time.Second,
	TLSHandshakeTimeout:   10 * time.Second,
	ExpectContinueTimeout: 1 * time.Second,
}

```

pkg/http/http.go
```diff
@@ -5,24 +5,38 @@
...
+	maxIdleConnsPerHost: 100, // see https://github.com/golang/go/issues/13801
```
bwplotka, line 37:
```suggestion
	maxIdleConnsPerHost: 100, // See https://github.com/golang/go/issues/13801.
```

pkg/http/http.go
bwplotka: We can, but we need to be aware who is using `exthttp.NewTransport` and how

pkg/http/http.go
```diff
@@ -95,10 +212,23 @@ func NewHTTPClient(cfg ClientConfig, name string) (*http.Client, error) {
 		return nil, err
 	}
 
-	client, err := config_util.NewClientFromConfig(httpClientConfig, name, config_util.WithHTTP2Disabled())
+	// see https://github.com/golang/go/issues/13801
+	if cfg.MaxIdleConns == 0 {
+		cfg.MaxIdleConns = 20000
```
bwplotka, line 217: I don't get it, why we don't use default? Also `0` means no limit, so 0 has meaning too, we should tolerate such option

pkg/http/http.go
```diff
@@ -95,10 +212,23 @@ func NewHTTPClient(cfg ClientConfig, name string) (*http.Client, error) {
 		return nil, err
 	}
 
-	client, err := config_util.NewClientFromConfig(httpClientConfig, name, config_util.WithHTTP2Disabled())
+	// see https://github.com/golang/go/issues/13801
```
bwplotka, line 215: Weird haging comment (and not full sentence), can we remove it?

pkg/http/http.go
```diff
@@ -95,10 +212,23 @@ func NewHTTPClient(cfg ClientConfig, name string) (*http.Client, error) {
...
+	if cfg.MaxIdleConns == 0 {
+		cfg.MaxIdleConns = 20000
+	}
+	if cfg.MaxIdleConnsPerHost == 0 {
+		cfg.MaxIdleConnsPerHost = 1000
```
bwplotka, line 220: Why those numbers?

pkg/http/http.go
```diff
@@ -95,10 +212,23 @@ func NewHTTPClient(cfg ClientConfig, name string) (*http.Client, error) {
...
+	}
+	rt, err := NewRoundTripperFromConfig(
+		httpClientConfig,
+		name,
+		WithMaxIdleConns(cfg.MaxIdleConns),
+		WithMaxIdleConnsPerHost(cfg.MaxIdleConnsPerHost))
 	if err != nil {
 		return nil, err
 	}
+	client := &http.Client{Transport: rt}
```
bwplotka, line 230: Setting transport does not make sense, if we override it later on, why not passing `rt` just below?

pkg/http/http.go
```diff
@@ -262,3 +392,151 @@ func (c *Client) Discover(ctx context.Context) {
 func (c *Client) Resolve(ctx context.Context) error {
 	return c.provider.Resolve(ctx, append(c.fileSDCache.Addresses(), c.staticAddresses...))
 }
+
+// Duplicated from github.com/prometheus/common/config
+// better option would be to expose config_util.newTLSRoundTripper function from prometheus
```
bwplotka, line 397: Yes, can we do that (you can contribute there, we are happy to review and merge it there)

cmd/thanos/config.go
bwplotka: Why it was receive? 🤔

bwplotka (commented):
Don't have time to review fully, but did part of it, looks good generally, some nits.

pkg/http/http.go
someshkoli: > 0 feels wrong, see default Go impl:
> 
> ```
> // DefaultTransport is the default implementation of Transport and is
> // used by DefaultClient. It establishes network connections as needed
> // and caches them for reuse by subsequent calls. It uses HTTP proxies
> // as directed by the $HTTP_PROXY and $NO_PROXY (or $http_proxy and
> // $no_proxy) environment variables.
> var DefaultTransport RoundTripper = &Transport{
> 	Proxy: ProxyFromEnvironment,
> 	DialContext: (&net.Dialer{
> 		Timeout:   30 * time.Second,
> 		KeepAlive: 30 * time.Second,
> 		DualStack: true,
> 	}).DialContext,
> 	ForceAttemptHTTP2:     true,
> 	MaxIdleConns:          100,
> 	IdleConnTimeout:       90 * time.Second,
> 	TLSHandshakeTimeout:   10 * time.Second,
> 	ExpectContinueTimeout: 1 * time.Second,
> }
> ```

Yes right

pkg/http/http.go
someshkoli: Makes sense

pkg/http/http.go
someshkoli: Again from prometheus implementation.

pkg/http/http.go
someshkoli: Great, I'll raise a PR there

pkg/http/http.go
someshkoli: > I don't get it, why we don't use default? Also `0` means no limit, so 0 has meaning too, we should tolerate such option

Actually, this value is picked from `prometheus http` which we were using before, so thought it would be better to keep it the same. Also yeah 0 does for this is acceptable, 'll fix it.

cmd/thanos/config.go
someshkoli: By mistake I assume :thinking: I don't see it being used anywhere

pkg/http/http.go
someshkoli: https://github.com/prometheus/common/pull/301

cmd/thanos/sidecar.go
```diff
@@ -84,7 +81,24 @@ func runSidecar(
...
+		return errors.Wrap(err, "getting http client config")
+	}
+	httpClientConfig, err := thanoshttp.NewClientConfigFromYAML(httpConfContentYaml)
+	if err != nil {
+		return errors.Wrap(err, "parsing http config YAML")
+	}
+
+	httpClient, err := thanoshttp.NewHTTPClient(*httpClientConfig, "thanos-sidecar")
+	if err != nil {
+		return errors.Wrap(err, "Improper http client config")
```
bwplotka, line 95:
```suggestion
		return errors.Wrap(err, "new http client config")
```

pkg/http/http.go
```diff
@@ -35,6 +39,8 @@ type ClientConfig struct {
 	ProxyURL string `yaml:"proxy_url"`
 	// TLSConfig to use to connect to the targets.
 	TLSConfig TLSConfig `yaml:"tls_config"`
+	// TransportConfig for Client transport properties
```
bwplotka, line 42:
```suggestion
	// TransportConfig for Client transport properties.
```

pkg/http/http.go
```diff
@@ -91,15 +195,31 @@ func NewHTTPClient(cfg ClientConfig, name string) (*http.Client, error) {
...
+
+	if cfg.BearerTokenFile != "" {
+		httpClientConfig.BearerTokenFile = cfg.BearerTokenFile
+	}
+
 	if err := httpClientConfig.Validate(); err != nil {
 		return nil, err
 	}
 
-	client, err := config_util.NewClientFromConfig(httpClientConfig, name, config_util.WithHTTP2Disabled())
```
bwplotka, line 98: Can we still use it? and append middleware we want on top?

bwplotka (commented):
LGTM generally, just small nits. Thanks.

bwplotka:
Any update on this?

someshkoli:
> Any update on this?

Its ready for a review, I guess I forgot to put a review request :smile:

yeya24:
@someshkoli Can you rebase on main and fix the conflicts? Thank you!

yeya24:
Hi, @someshkoli, sorry for the late reply. Can you fix the conflict and get CI passing? Your work is really amazing and we would like to have this feature!

someshkoli:
> Hi, @someshkoli, sorry for the late reply. Can you fix the conflict and get CI passing? Your work is really amazing and we really want to have this feature!

will update it soon, have a lot to rebase :")

someshkoli:
Moving this to #4612

someshkoli:
@yeya24 can you please take a look at #4612 ?

yeya24:
Close in favor of #4612

PR #4105, opened 2021-04-24
Adds glob support for filepath in tools command

### LINKED ISSUE #4082: rules-check doesn't accept globs as described in the documentation
The example rules-check command states that rules-check takes a glob of rules files
```
./thanos tools rules-check --rules cmd/thanos/testdata/rules-files/*.yaml
```
https://thanos.io/v0.19/components/tools.md/#rules-check
But this does not work for me.

You either get shell expansion of the glob, resulting in:
```
./thanos tools rules-check --rules config/rules/*.yaml
error parsing commandline arguments: [/bin/thanos tools rules-check --rules config/rules/file1.yaml config/rules/file2.yaml]: unexpected config/rules/file2.yaml
thanos: error: unexpected config/rules/file2.yaml
```

Or without shell expansion:
```
./thanos tools rules-check --rules=config/rules/*.yaml
error parsing commandline arguments: [/bin/thanos tools rules-check --rules=config/rules/*.yaml]: path 'config/rules/*.yaml' does not exist
usage: thanos tools rules-check --rules=RULES
```

The only form that does work is
```
./thanos tools rules-check --rules config/rules/file1.yaml --rules config/rules/file2.yaml
```

This is either a documentation error or the glob handling is not working.

someshkoli:

## Changes
Adds glob check for filepaths provided via cli. Checks for each matching file path and throws `Matching file not found` error if no file matches the pattern.

fixes #4082 

## Verification
Added a basic unit test

### DIFF
--- a/cmd/thanos/tools.go
+++ b/cmd/thanos/tools.go
@@ -4,7 +4,6 @@
 package main
 
 import (
-	"errors"
 	"os"
 	"path/filepath"
 
@@ -12,6 +11,7 @@ import (
 	"github.com/go-kit/kit/log/level"
 	"github.com/oklog/run"
 	"github.com/opentracing/opentracing-go"
+	"github.com/pkg/errors"
 	"github.com/prometheus/client_golang/prometheus"
 	"github.com/thanos-io/thanos/pkg/errutil"
 	"github.com/thanos-io/thanos/pkg/extkingpin"
--- a/cmd/thanos/tools.go
+++ b/cmd/thanos/tools.go
@@ -36,29 +36,23 @@ func registerCheckRules(app extkingpin.AppClause) {
 	})
 }
 
-func checkRulesFiles(logger log.Logger, files *[]string) error {
+func checkRulesFiles(logger log.Logger, patterns *[]string) error {
 	var failed errutil.MultiError
-	errhandle := func(e error) {
-		level.Error(logger).Log("result", "FAILED", "error", e)
-		level.Info(logger).Log()
-		failed.Add(e)
-	}
 
-	for _, fn := range *files {
-		level.Info(logger).Log("msg", "checking", "filename", fn)
-		matches, err := filepath.Glob(fn)
-		if err != nil {
-			errhandle(err)
-			continue
-		}
-		if matches == nil {
+	for _, p := range *patterns {
+		level.Info(logger).Log("msg", "checking", "pattern", p)
+		matches, err := filepath.Glob(p)
+		if err != nil || matches == nil {
 			err = errors.New("matching file not found")
-			errhandle(err)
+			level.Error(logger).Log("result", "FAILED", "error", err)
+			level.Info(logger).Log()
+			failed.Add(err)
 			continue
 		}
-		for _, fn1 := range matches {
+		for _, fn := range matches {
+			level.Info(logger).Log("msg", "checking", "filename", fn)
 			// no need to check path error
-			f, _ := os.Open(fn1)
+			f, _ := os.Open(fn)
 			defer func() { _ = f.Close() }()
 
 			n, errs := rules.ValidateAndCount(f)
--- a/cmd/thanos/tools.go
+++ b/cmd/thanos/tools.go
@@ -38,17 +38,22 @@ func registerCheckRules(app extkingpin.AppClause) {
 
 func checkRulesFiles(logger log.Logger, files *[]string) error {
 	var failed errutil.MultiError
+	errhandle := func(e error) {
+		level.Error(logger).Log("result", "FAILED", "error", e)
+		level.Info(logger).Log()
+		failed.Add(e)
+	}
 
 	for _, fn := range *files {
 		level.Info(logger).Log("msg", "checking", "filename", fn)
 		matches, err := filepath.Glob(fn)
-		if matches == nil {
-			err = errors.New("Matching file not found")
-		}
 		if err != nil {
-			level.Error(logger).Log("result", "FAILED", "error", err)
-			level.Info(logger).Log()
-			failed.Add(err)
+			errhandle(err)
+			continue
+		}
+		if matches == nil {
+			err = errors.New("matching file not found")
+			errhandle(err)
 			continue
 		}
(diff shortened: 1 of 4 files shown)


cmd/thanos/tools.go
```diff
@@ -39,26 +41,33 @@ func checkRulesFiles(logger log.Logger, files *[]string) error {
 
 	for _, fn := range *files {
 		level.Info(logger).Log("msg", "checking", "filename", fn)
-		f, err := os.Open(fn)
+		matches, err := filepath.Glob(fn)
+		if matches == nil {
```
onprem, line 45: We should check the `err` first and then `matches`, because `matches` will most likely be `nil` in case there is some other error and we will be overwriting that error with "Matching file not found".

someshkoli: Right :thinking:

someshkoli:
Also, I guess this change should not affect e2e :thinking:, not sure why its failing :\

cmd/thanos/tools.go
```diff
@@ -36,29 +38,41 @@ func registerCheckRules(app extkingpin.AppClause) {
...
 func checkRulesFiles(logger log.Logger, files *[]string) error {
 	var failed errutil.MultiError
+	errhandle := func(e error) {
+		level.Error(logger).Log("result", "FAILED", "error", e)
+		level.Info(logger).Log()
+		failed.Add(e)
+	}
 
 	for _, fn := range *files {
 		level.Info(logger).Log("msg", "checking", "filename", fn)
```
GiedriusS, line 48: This isn't a filename anymore but rather a pattern:
```suggestion
		level.Info(logger).Log("msg", "checking", "pattern", fn)
```

cmd/thanos/tools.go
```diff
@@ -36,29 +38,41 @@ func registerCheckRules(app extkingpin.AppClause) {
...
 		if err != nil {
-			level.Error(logger).Log("result", "FAILED", "error", err)
-			level.Info(logger).Log()
-			failed.Add(err)
+			errhandle(err)
+			continue
+		}
+		if matches == nil {
+			err = errors.New("matching file not found")
+			errhandle(err)
```
GiedriusS, line 56: We can join these two together

cmd/thanos/tools.go
```diff
@@ -36,29 +38,41 @@ func registerCheckRules(app extkingpin.AppClause) {
...
 			continue
 		}
-		defer func() { _ = f.Close() }()
+		for _, fn1 := range matches {
```
GiedriusS, line 59: Maybe we can output the file's name here like in the past?
```suggestion
		    level.Info(logger).Log("msg", "checking", "filename", fn1)
```

GiedriusS, line 59: Also, can we use a more descriptive variable name here? I suggest using `fn` here.

cmd/thanos/tools.go
```diff
@@ -36,29 +38,41 @@ func registerCheckRules(app extkingpin.AppClause) {
 
 func checkRulesFiles(logger log.Logger, files *[]string) error {
 	var failed errutil.MultiError
+	errhandle := func(e error) {
+		level.Error(logger).Log("result", "FAILED", "error", e)
+		level.Info(logger).Log()
+		failed.Add(e)
+	}
 
 	for _, fn := range *files {
```
GiedriusS, line 47: Let's rename `files` to `patterns` and `fn` to `pattern`

cmd/thanos/tools.go
```diff
@@ -36,29 +38,41 @@ func registerCheckRules(app extkingpin.AppClause) {
...
+			continue
+		}
+		if matches == nil {
+			err = errors.New("matching file not found")
+			errhandle(err)
 			continue
 		}
-		defer func() { _ = f.Close() }()
+		for _, fn1 := range matches {
+			// no need to check path error
```
GiedriusS, line 60: Why not? :thinking:

GiedriusS (commented):
Just some suggestions/questions but overall LGTM! Thank you for your work on this feature (:

cmd/thanos/tools.go
someshkoli: > Why not?

If the path is invalid then it wouldn't be in `matches` at the first place :thinking:

cmd/thanos/tools.go
someshkoli: > We can join these two together

Like single error as `Invalid patter or matching file not found` ?

cmd/thanos/tools.go
GiedriusS: Yes but we still may not be able to open that file for a lot of reasons (:

cmd/thanos/tools.go
GiedriusS: My idea was that we could simply write:
```
errhandle(errors.New("matching file not found"))
```
:smile:

cmd/thanos/tools.go
someshkoli: > Yes but we still may not be able to open that file for a lot of reasons (:

makes sense :thinking:

cmd/thanos/tools.go
```diff
@@ -34,31 +36,42 @@ func registerCheckRules(app extkingpin.AppClause) {
...
+		for _, fn := range matches {
+			level.Info(logger).Log("msg", "checking", "filename", fn)
+			f, er := os.Open(fn)
+			if er != nil {
+				level.Error(logger).Log("result", "FAILED", "error", er)
+				level.Info(logger).Log()
+				failed.Add(err)
+				continue
+			}
+			defer func() { _ = f.Close() }()
```
GiedriusS, line 61: Typically we'd use `rununtil.CloseWithLogOnErr` but since at the end the process closes immediately, this is fine. Just something to keep in mind for future contributions :+1:

GiedriusS (approved):
Awesome work! Thanks a lot

cmd/thanos/tools.go
someshkoli: Got it, Thanks :raised_hands:

Issue #3637, opened 2020-12-17
Support matchers in gRPC LabelNames and LabelValues requests


yeya24:
Initial support for label matchers API is done in https://github.com/thanos-io/thanos/pull/3566.

It just simply sends `Series` requests to stores and deduplicate the results at the API level.

It works but we might improve more by adding matchers to `LabelNames` and `LabelValues` gRPC requests to push down matchers to each store. Each store can get the matched labels and send them back. The network load would be better.

AC:
1. Add `matchers` to `LabelNames` and `LabelValues` gRPC requests
2. For each store, implement the logic to filter the labels based on the given matchers

Biswajitghosh98:
@yeya24 While I'd like to give it a try.

yeya24:
Sorry for being unclear. I am already working on this one. Currently blocked by #3639

Biswajitghosh98:
No worries. I'll pick a different issue 😄

pstibrany:
Hello, if there is no update on this, I would like to implement support for matchers in LabelNames and LabelValues calls in Thanos `store` (gateway) component. (`*BucketStore` type).

My plan is to mimick the implementation in Prometheus TSDB: basically call `blockSeries` with supplied matchers (in case of LabelValues, matchers would be extended with `labelName!=""`) but no chunk-fetching. From returned series set, code would extract unique label names or label values and return them. If no matchers were used, we would reuse current implementation.

My plan does not include implementing support for matchers in other `StoreServer` implementations.

@yeya24 Please let me know if you are still working on this or have other plans, or I can go ahead with my plan.

Thank you!

bwplotka:
Sounds good to me! I think we started some work on sidecar but not for store.

yeya24:
This has been solved. Close.

PR #4107, opened 2021-04-26
Add support for label matchers in LabelNames and LabelValues call to store.

pstibrany:

## Changes

This PR implements support for label matches in `LabelNames` and `LabelValues` API calls to bucket store. It reuses existing `blockSeries` implementation, but without fetching of chunks.

## Verification

Unit tests.

This is a partial fix for issue https://github.com/thanos-io/thanos/issues/3637


pkg/store/bucket.go
```diff
@@ -746,14 +747,17 @@ func (s *bucketSeriesSet) Err() error {
 	return s.err
 }
 
+// blockSeries returns series matching given matchers, that have some data in given time range.
 func blockSeries(
-	extLset labels.Labels,
-	indexr *bucketIndexReader,
-	chunkr *bucketChunkReader,
-	matchers []*labels.Matcher,
-	req *storepb.SeriesRequest,
```
pstibrany, line 754: Instead of `req` we now pass all used parameters explicitly.

pkg/store/bucket.go
```diff
@@ -1154,16 +1160,14 @@ func chunksSize(chks []storepb.AggrChunk) (size int) {
...
 func (s *BucketStore) LabelNames(ctx context.Context, req *storepb.LabelNamesRequest) (*storepb.LabelNamesResponse, error) {
-	resHints := &hintspb.LabelNamesResponseHints{}
-
-	g, gctx := errgroup.WithContext(ctx)
+	reqSeriesMatchers, err := storepb.MatchersToPromMatchers(req.Matchers...)
+	if err != nil {
+		return nil, status.Error(codes.InvalidArgument, errors.Wrap(err, "translate request labels matchers").Error())
+	}
 
-	s.mtx.RLock()
```
pstibrany, line 1161: Note that `s.mtx.RLock()` is now only called after checking hints. Previously wrong hints would return error without unlocking this lock.

pkg/store/bucket.go
```diff
@@ -1188,32 +1201,61 @@ func (s *BucketStore) LabelNames(ctx context.Context, req *storepb.LabelNamesReq
...
+					return errors.Wrapf(err, "label names for block %s", b.meta.ULID)
+				}
 
-			// Add  a set for the external labels as well.
-			// We're not adding them directly to res because there could be duplicates.
-			extRes := make([]string, 0, len(extLabels))
-			for lName := range extLabels {
-				extRes = append(extRes, lName)
-			}
+				sort.Strings(res)
```
yeya24, line 1216: The result of `indexr.block.indexHeaderReader.LabelNames()` is already sorted https://github.com/thanos-io/thanos/blob/main/pkg/block/indexheader/binary_reader.go#L904.
Do you think we can remove this line?

pkg/store/bucket.go
```diff
@@ -1188,32 +1201,61 @@ func (s *BucketStore) LabelNames(ctx context.Context, req *storepb.LabelNamesReq
...
+				seriesSet, _, err := blockSeries(b.extLset, indexr, nil, reqSeriesMatchers, nil, seriesLimiter, true, req.Start, req.End, nil)
+				if err != nil {
+					return errors.Wrapf(err, "fetch series for block %s", b.meta.ULID)
+				}
 
-			sort.Strings(res)
-			sort.Strings(extRes)
+				// Extract label names from all series. Many label names will be the same, so we need to deduplicate them.
+				// Note that label names will already include external labels (passed to blockSeries), so we don't need
+				// to add them again.
```
yeya24, line 1235: :+1:

pkg/store/bucket.go
```diff
@@ -1272,25 +1332,55 @@ func (s *BucketStore) LabelValues(ctx context.Context, req *storepb.LabelValuesR
...
+					return errors.Wrapf(err, "index header label values for block %s", b.meta.ULID)
+				}
+
+				// Add the external label value as well.
+				if extLabelValue := b.extLset.Get(req.Label); extLabelValue != "" {
+					res = strutil.MergeSlices(res, []string{extLabelValue})
+				}
+				result = res
+			} else {
+				seriesSet, _, err := blockSeries(b.extLset, indexr, nil, reqSeriesMatchers, nil, seriesLimiter, true, req.Start, req.End, nil)
```
yeya24, line 1353: This can be optimized with https://github.com/prometheus/prometheus/pull/8400.
We can open a new issue to track it and see if someone would like to contribute to this.

pkg/store/bucket.go
```diff
@@ -1261,7 +1303,25 @@ func (s *BucketStore) LabelValues(ctx context.Context, req *storepb.LabelValuesR
 		}
 	}
 
+	// If we have series matchers, add <labelName> != "" matcher, to only select series that have given label name.
+	if len(reqSeriesMatchers) > 0 {
+		m, err := labels.NewMatcher(labels.MatchNotEqual, req.Label, "")
```
yeya24, line 1308: We can make it a global variable with `labels.MustNewMatcher()`, WDYT? @pstibrany

yeya24 (approved):
LGTM! Thanks for the great work.

pkg/store/bucket.go
pstibrany: Sounds good, but to make sure we don't break it in the future, I will also update `indexheader.Reader` interface to specify the contract about labels being sorted.

pkg/store/bucket.go
pstibrany: I don't quite follow what you mean. This cannot be a global variable, as it uses label name from the request. What am I missing?

pkg/store/bucket.go
pstibrany: **This** PR already uses the same technique, i.e. finding postings for matching series, and only extracting label values from them (see `labelValuesWithMatchers` in https://github.com/prometheus/prometheus/pull/8400). We cannot use https://github.com/prometheus/prometheus/pull/8400 directly, because we don't have all required information (ie. postings) locally in the index header. That's why we use `blockSeries`, which fetches postings for us.

pkg/store/bucket.go
yeya24: Sorry, I miss it. You are right.

pkg/store/bucket.go
yeya24: That makes sense. Thanks for the explanation!

yeya24 (dismissed):
LGTM. Thanks for the great work!

pracucci (dismissed):
Amazing job! 👏  LGTM

The failing `TestTargetsAPI_Fanout` is known to be flaky. Unrelated from this change.

pracucci:
I've just rebased master. Let's see if CI passes now.

pracucci (approved):
Approving again. @yeya24 also approved and only the rebase has been done in the meanwhile.

pkg/store/bucket.go
```diff
@@ -1261,7 +1302,25 @@ func (s *BucketStore) LabelValues(ctx context.Context, req *storepb.LabelValuesR
 		}
 	}
 
+	// If we have series matchers, add <labelName> != "" matcher, to only select series that have given label name.
+	if len(reqSeriesMatchers) > 0 {
```
yeya24, line 1306: What if `reqSeriesMatchers` already contains this unequal matcher? @pstibrany @pracucci

pstibrany: End result will be correct, but there is an optimization we can do (preferably in `blockSeries` to be more general).

yeya24: Make sense. Thank you.

PR #4113, opened 2021-04-27
Fix query dashboards when frontend is deployed

### LINKED ISSUE #4112: Query dashboard error if query-frontend installed

Thanos installed from kube-thanos main branch with jsonnet
Thanos dashboards installed with with jsonnet from [mixin/dashboards.jsonnet](mixin/dashboards.jsonnet)

**Object Storage Provider**: S3

**What happened**:

The current query dashboard is not working when both `thanos-query` and `query-frontend` because the generated dashboard's requests are using `job=\"$job\"` ([here](https://github.com/thanos-io/thanos/blob/main/examples/dashboards/query.json#L148)) that is then interpolated by the grafana variable which is `job=~\".*thanos-query.*\"}` ([here](https://github.com/thanos-io/thanos/blob/main/examples/dashboards/query.json#L1833). So the final resul is `job=\"thanos-query|thanos-query-frontend\"` which is not working because of the strict equality. 
To make it work I updated the dashboard to replace all `job=\"$job\"` by `job=~\"$job\"` and the dashboard began to work. 

PR is coming...

yogeek:
Fixes #4112

### DIFF
--- a/examples/dashboards/query.json
+++ b/examples/dashboards/query.json
@@ -67,7 +67,7 @@
                "steppedLine": false,
                "targets": [
                   {
-                     "expr": "sum by (job, handler, code) (rate(http_requests_total{job=\"$job\", handler=\"query\"}[$interval]))",
+                     "expr": "sum by (job, handler, code) (rate(http_requests_total{job=~\"$job\", handler=\"query\"}[$interval]))",
                      "format": "time_series",
                      "intervalFactor": 2,
                      "legendFormat": "{{job}} {{handler}} {{code}}",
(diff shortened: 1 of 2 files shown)


kakkoyun (changes requested):
Thanks a lot for the contribution. And welcome to Thanos project!

You need to generate examples, you can see the instructions in https://github.com/thanos-io/thanos/tree/main/mixin.

Moreover, do you think other dashboards also could benefit from this improvement?

yogeek:
Hi @kakkoyun, thanks for the help.
I just generated the examples and add the new dashboard to this PR.
About your question, I do not think other dashboards can have the same issue currently, because this seems related to the fact that the query component can be completed by the query-frontend, but this is specific to query as I understood from the doc.

yogeek:
@kakkoyun Apparently the merge is blocked...do I have to do something else on my side please ?

kakkoyun:
Hey @yogeek, you need to sign your comments for DCO https://github.com/thanos-io/thanos/pull/4113/checks?check_run_id=2457531742 
And I'll have another look at it. Thanks 👍

yogeek:
ok @kakkoyun  it's done

kakkoyun (approved):

PR #4123, opened 2021-04-29
match external labels in exemplars API

### LINKED ISSUE #4116: query: query_exemplars does not account for external labels
@goku321

**Thanos, Prometheus and Golang version used**: v0.20.0 docker image

**Object Storage Provider**: gcs

**What happened**:
- configured ```--exemplar=...``` on our local and global queriers
- configure grafana for exemplars

- grafana "time series" panel should show exemplars.
however...
the time series panel includes the full query from the panel, including external labels used by the local prom
instances.
the query\_range queries work as expected.
query\_exemplars did not.

If I send the query\_exemplars a query including the external labels, I get an empty response.
If I remove the external labels and just use labels the local prom instance knows about, I get responses.

- configure a prometheus with external labels, with sidecar
- configure thanos query to see the sidecar
- query thanos-query with the external labels.
(we have a second tier of queriers, but I do not think that is causing the issue.

yeya24:

Fixes #4116 

## Changes

Add logic of matching external labels in the exemplars API proxy.

The matching part is a little bit complex because the input is a query, which might contain multiple vector selectors.
We need to match each vector selector separately and construct the final query after all matches are processed.

## Verification

Tested locally and one unit test.
We don't have an E2E test for exemplars now so no E2E test added.

### DIFF
--- a/pkg/exemplars/proxy.go
+++ b/pkg/exemplars/proxy.go
@@ -10,6 +10,8 @@ import (
 	"github.com/go-kit/kit/log"
 	"github.com/go-kit/kit/log/level"
 	"github.com/pkg/errors"
+	"github.com/prometheus/prometheus/pkg/labels"
+	"github.com/prometheus/prometheus/promql/parser"
 	"github.com/thanos-io/thanos/pkg/exemplars/exemplarspb"
 	"github.com/thanos-io/thanos/pkg/store/storepb"
 	"golang.org/x/sync/errgroup"
@@ -21,8 +23,9 @@ import (
 // Proxy implements exemplarspb.Exemplars gRPC that fanouts requests to
 // given exemplarspb.Exemplars.
 type Proxy struct {
-	logger    log.Logger
-	exemplars func() []exemplarspb.ExemplarsClient
+	logger         log.Logger
+	exemplars      func() []*exemplarspb.ExemplarStore
+	selectorLabels labels.Labels
 }
 
 // RegisterExemplarsServer register exemplars server.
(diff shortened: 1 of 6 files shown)


pkg/exemplars/proxy.go
```diff
@@ -48,16 +52,80 @@ type exemplarsStream struct {
...
+	expr, err := parser.ParseExpr(req.Query)
+	if err != nil {
+		return err
+	}
+
+	selectors := parser.ExtractSelectors(expr)
+
+	newSelectors := make([][]*labels.Matcher, 0, len(selectors))
+	for _, matchers := range selectors {
+		matched, newMatchers:= matchesExternalLabels(matchers, s.selectorLabels)
```
kakkoyun, line 64:
```suggestion
		matched, newMatchers := matchesExternalLabels(matchers, s.selectorLabels)
```

kakkoyun:
Thanks, @yeya24, this looks amazing. I need to muster some brainpower to review this.

yeya24:
@kakkoyun Thanks for the quick response. Yes it is a little bit complex and hard to review it directly from the code. It would be good to deploy it and test the external labels matching feature.

@tcolgate It would be good if you can help test this pr as well.

tcolgate:
I've tested this change locally and we are now seeing exemplars on our staging environment (querying via both layers of thanos querier). LGTM

pkg/exemplars/proxy.go
```diff
@@ -48,16 +52,80 @@ type exemplarsStream struct {
...
+			Start:                   req.Start,
+			End:                     req.End,
+			Query:                   query,
+			PartialResponseStrategy: req.PartialResponseStrategy,
+		}
+
 		es := &exemplarsStream{
-			client:  exemplarsClient,
-			request: req,
+
```
kakkoyun, line 126:
```suggestion
```
nit

kakkoyun (approved):
LGTM

PR #4135, opened 2021-05-01
Update E2E test Prometheus image to 2.26.0

yeya24:

Update E2E Prometheus image to 2.26.0 to enable some tests on new features like Exemplars.

## Changes

This upgrade breaks the rules API test because Prometheus has strict unmarshalling.

This pr updates the original test case:
1. Remove one of the rule files that Prometheus uses in the test case
2. Use a separate rules dir for Thanos Rulers and enable Rulers for rules evaluation
3. Update the response

### DIFF
--- a/test/e2e/rules_api_test.go
+++ b/test/e2e/rules_api_test.go
@@ -16,7 +16,9 @@ import (
 	"github.com/cortexproject/cortex/integration/e2e"
 	"github.com/pkg/errors"
 
+	http_util "github.com/thanos-io/thanos/pkg/http"
 	"github.com/thanos-io/thanos/pkg/promclient"
+	"github.com/thanos-io/thanos/pkg/query"
 	"github.com/thanos-io/thanos/pkg/rules/rulespb"
 	"github.com/thanos-io/thanos/pkg/runutil"
 	"github.com/thanos-io/thanos/pkg/store/labelpb"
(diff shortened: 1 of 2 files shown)


test/e2e/rules_api_test.go
```diff
@@ -33,42 +35,42 @@ func TestRulesAPI_Fanout(t *testing.T) {
...
-	r2, err := e2ethanos.NewRuler(s.SharedDir(), "rule2", rulesSubDir, nil, nil)
-	testutil.Ok(t, err)
-	testutil.Ok(t, s.StartAndWaitReady(r1, r2))
-
 	q, err := e2ethanos.NewQuerier(
 		s.SharedDir(),
 		"query",
-		[]string{sidecar1.GRPCNetworkEndpoint(), sidecar2.GRPCNetworkEndpoint(), r1.GRPCNetworkEndpoint(), r2.GRPCNetworkEndpoint()},
+		// TODO(yeya24): stop hardcoding the GRPC endpoints for rulers.
+		[]string{sidecar1.GRPCNetworkEndpoint(), sidecar2.GRPCNetworkEndpoint(), "rule-rule1:9091", "rule-rule2:9091"},
```
bwplotka, line 71: Why can't we use endpoint method?

test/e2e/rules_api_test.go
```diff
@@ -87,7 +104,7 @@ func TestRulesAPI_Fanout(t *testing.T) {
 	ruleAndAssert(t, ctx, q.HTTPEndpoint(), "", []*rulespb.RuleGroup{
 		{
 			Name: "example_abort",
-			File: "/shared/rules/rules-0.yaml",
+			File: "/shared/rules/rules.yaml",
```
bwplotka, line 107: weird...

bwplotka (commented):
Thanks, just small nits (:

test/e2e/rules_api_test.go
yeya24: If we use endpoints methods of rulers, we need to create and start them first. But to create rulers, we need the address of thanos query first. So either way, we need to hardcode the address for one component.

test/e2e/rules_api_test.go
yeya24: I separate the rule files for Prometheus and Thanos ruler. https://github.com/thanos-io/thanos/pull/4135/files#diff-043bcd9734d5ba38687c9b0199a81e047ff6f744e1b5351e6a3c9c8e96f446a9R42

test/e2e/rules_api_test.go
yeya24: Any comment about this?

bwplotka: Well, I wonder if we can improve the framework for this approach. It sounds quite common. 🤔 Right now we put TODO but there is literally no idea how to solve it 🙈

test/e2e/rules_api_test.go
bwplotka: Ah, makes sense, Thanks

bwplotka (approved):
LGTM, but I would love to think about solution for hardcoding in this PR, otherwise this pattern will spread like a virus (: 

Thanks! Would you have time to help?

test/e2e/rules_api_test.go
yeya24: Fixed. PTAL

yeya24:
Is it good to go?

kakkoyun (approved):
LGTM 🚀

PR #4142, opened 2021-05-03
fix env variables described  in CONTRIBUTING.md

brmurphy:
GOBIN shouldn't be listed twice.


GiedriusS (approved):
:+1:

squat (commented):
Thanks, this is definitely correct, however i wonder what was meant by "set to PATH"?

GiedriusS:
> Thanks, this is definitely correct, however i wonder what was meant by "set to PATH"?

Probably that GOBIN should be in PATH. We can fix this in a follow up PR

PR #4146, opened 2021-05-03
(bugbash) fixes opt.semgrep.err-nil-chec

someshkoli:

## Changes
Removes unwanted err nil checks


someshkoli:
moved to #4147

PR #4149, opened 2021-05-04
Weak random number generator replaced 

nisarg1499:
Signed-off-by: Nisarg Shah nisshah1499@gmail.com

## Changes
Currently in many files, weak random number generator(math/rand) is being used. Replaced it to crypto/rand.

### DIFF
--- a/pkg/block/index.go
+++ b/pkg/block/index.go
@@ -8,7 +8,7 @@ import (
 	"fmt"
 	"hash/crc32"
 	"math"
-	"math/rand"
+	"crypto/rand"
 	"path/filepath"
 	"sort"
 	"strings"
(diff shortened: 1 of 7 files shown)


squat (approved):

squat (commented):
Thanks for writing this @nisarg1499. Using crypto rand sounds great, however we'll need to adjust the code to ensure since this passage has a different interface

nisarg1499:
@squat Yeah agreed. Can you guide me through some structure and changes which I should look to support crypto/rand ?

Namanl2001:
I don't have much context of this, but after a small research, I found that `crypto/rand` costs much higher than `math/rand`, which would affect the overall performance. So do we really need this?

Also there are not much functions in `crypto/rand` [package](https://golang.org/pkg/crypto/rand/). So it might be difficult to use it according to our needs.

PR #4154, opened 2021-05-04
(bugsbash)Package "prometheus/prometheus/pkg/labels" imported twice

asquare14:

* [x ] Change is not relevant to the end user.

## Changes
 Fixes ST1019 : package "github.com/prometheus/prometheus/pkg/labels" is being imported more than once

### DIFF
--- a/pkg/promclient/promclient.go
+++ b/pkg/promclient/promclient.go
@@ -26,7 +26,7 @@ import (
 	"github.com/gogo/status"
 	"github.com/pkg/errors"
 	"github.com/prometheus/common/model"
-	promlabels "github.com/prometheus/prometheus/pkg/labels"
+	"github.com/prometheus/prometheus/pkg/labels"
 	"github.com/prometheus/prometheus/pkg/timestamp"
 	"github.com/prometheus/prometheus/promql"
 	"github.com/prometheus/prometheus/promql/parser"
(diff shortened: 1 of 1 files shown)


squat:
Thank you @asquare14! Could you please sign the DCO (hint git commit --amend --signoff) and push again?

asquare14:
Ahh there's an issue with my laptop and by mistake I removed all the CLI tools. Waiting for some software to get installed. Will signoff in an hour or so. Meanwhile looking at how to signoff using GitHub UI. Thanks for the review !

asquare14:
@squat Done !

pkg/promclient/promclient.go
```diff
@@ -26,7 +26,6 @@ import (
 	"github.com/gogo/status"
 	"github.com/pkg/errors"
 	"github.com/prometheus/common/model"
-	"github.com/prometheus/prometheus/pkg/labels"
```
squat, line 29: Ok, so it looks like we are using both of the names for this package in different places. Some lines, e.g. line 160/159 use the `labels` name for this package whereas other places refer to it as `promlabels`. To make this change work, we need to rename all of the usages to make sure that they all use either `labels` or `promlabels` but not both

squat (commented):
Thanks for following up :) just need to do one more thing to make the code compile

pkg/promclient/promclient.go
asquare14: Done ! Changed promlabels to labels for consistency.

squat (approved):
Amazing, thank you!

PR #4155, opened 2021-05-04
(bugsbash) Error string should not be captilazed in options.go

asquare14:

## Changes
Fixes ST1005 : Error string should not be capitalized and should not end with punctuation or a newline

### DIFF
--- a/pkg/logging/options.go
+++ b/pkg/logging/options.go
@@ -175,7 +175,7 @@ func ParseHTTPOptions(flagDecision string, reqLogConfig *extflag.PathOrContent) 
 
 	// Check if the user enables request logging through flags and YAML.
 	if len(configYAML) != 0 && len(flagDecision) != 0 {
-		return logOpts, fmt.Errorf("both log.request.decision and request.logging has been enabled, please use only one of the flags")
+		return logOpts, fmt.Errorf("both log.request.decision and request.logging have been enabled, please use only one of the flags")
 	}
 	// If old flag is enabled.
 	if len(flagDecision) > 0 {
--- a/pkg/logging/options.go
+++ b/pkg/logging/options.go
@@ -175,7 +175,7 @@ func ParseHTTPOptions(flagDecision string, reqLogConfig *extflag.PathOrContent) 
 
 	// Check if the user enables request logging through flags and YAML.
 	if len(configYAML) != 0 && len(flagDecision) != 0 {
-		return logOpts, fmt.Errorf("both log.request.decision and request.logging has been enabled. Please use one of the flags")
+		return logOpts, fmt.Errorf("both log.request.decision and request.logging has been enabled, please use only one of the flags")
 	}
 	// If old flag is enabled.
 	if len(flagDecision) > 0 {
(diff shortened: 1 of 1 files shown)


squat (dismissed):
this one is tricky because this error is not wrapped any more but is the result of parsing command line arguments.

If other maintainers prefer this output be a full sentence, then we can refrain from merging

pkg/logging/options.go
```diff
@@ -175,7 +175,7 @@ func ParseHTTPOptions(flagDecision string, reqLogConfig *extflag.PathOrContent)
 
 	// Check if the user enables request logging through flags and YAML.
 	if len(configYAML) != 0 && len(flagDecision) != 0 {
-		return logOpts, fmt.Errorf("Both log.request.decision and request.logging has been enabled. Please use one of the flags!")
+		return logOpts, fmt.Errorf("both log.request.decision and request.logging has been enabled. Please use one of the flags")
```
onprem, line 178:
```suggestion
		return logOpts, fmt.Errorf("both log.request.decision and request.logging has been enabled, please use only one of the flags")
```

As @squat said this is a bit tricky, but I'll still prefer no-caps to be consistent. After all this is going to be printed in a structured format `err="both log......"` so it's better to be consistent.

onprem (approved):
Just a small nit. Everything else LGTM.

asquare14:
@onprem and @squat Thanks for the review. Fixed the suggestions :)

pkg/logging/options.go
```diff
@@ -201,7 +201,7 @@ func ParsegRPCOptions(flagDecision string, reqLogConfig *extflag.PathOrContent)
 
 	// Check if the user enables request logging through flags and YAML.
 	if len(configYAML) != 0 && len(flagDecision) != 0 {
-		return []tags.Option{}, logOpts, fmt.Errorf("Both log.request.decision and request.logging-config has been enabled. Please use one of the flags!")
+		return []tags.Option{}, logOpts, fmt.Errorf("both log.request.decision and request.logging-config has been enabled. please use one of the flags")
```
squat, line 204: let's make this one match above:

```suggestion
		return []tags.Option{}, logOpts, fmt.Errorf("both log.request.decision and request.logging-config have been enabled, please use only one of the flags")
```

pkg/logging/options.go
```diff
@@ -175,7 +175,7 @@ func ParseHTTPOptions(flagDecision string, reqLogConfig *extflag.PathOrContent)
 
 	// Check if the user enables request logging through flags and YAML.
 	if len(configYAML) != 0 && len(flagDecision) != 0 {
-		return logOpts, fmt.Errorf("Both log.request.decision and request.logging has been enabled. Please use one of the flags!")
+		return logOpts, fmt.Errorf("both log.request.decision and request.logging has been enabled, please use only one of the flags")
```
squat, line 178:
```suggestion
		return logOpts, fmt.Errorf("both log.request.decision and request.logging have been enabled, please use only one of the flags")
```

squat (commented):
one more iteration plz! let's make sure they are consistent with each other

squat (dismissed):
:shipit:

squat:
ah, need to sign-off the DCO one more time it seems

asquare14:
> ah, need to sign-off the DCO one more time it seems

Done 😛 
Thanks for the review 🚀

squat (approved):

PR #4156, opened 2021-05-04
(bugsbash)error strings should not be capitalized

asquare14:

## Changes

Fixes ST1005 : error strings should not be capitalized

### DIFF
--- a/pkg/logging/http.go
+++ b/pkg/logging/http.go
@@ -88,7 +88,7 @@ func getHTTPLoggingOption(logStart bool, logEnd bool) (Decision, error) {
 	if logStart && logEnd {
 		return LogStartAndFinishCall, nil
 	}
-	return -1, fmt.Errorf("log start call is not supported.")
+	return -1, fmt.Errorf("log start call is not supported")
 }
 
 // getLevel returns the level based logger.


pkg/ui/bindata.go
```diff
@@ -98,15 +98,15 @@ import (
 func bindataRead(data []byte, name string) ([]byte, error) {
 	gz, err := gzip.NewReader(bytes.NewBuffer(data))
 	if err != nil {
-		return nil, fmt.Errorf("Read %q: %v", name, err)
+		return nil, fmt.Errorf("read %q: %v", name, err)
```
squat, line 101: This file is automatically generated using go-bindata, so changing this manually will be overwritten next time the file is generated and will cause our CI to fail :/. Let's leave this out but keep the change from pkg/logging/http.go

squat (commented):
Thank you @asquare14! As mentioned above, let's avoid making manual modifications to the pkg/ui/bindata.go file, which is automatically generated, and keep only the change to pkg/logging/http.go :rocket:

pkg/ui/bindata.go
asquare14: Makes sense 👍🏼  Done !

squat (approved):

PR #4158, opened 2021-05-04
(bugsbash)value of ctx is never used

asquare14:

## Changes
```go
log, ctx := spanlogger.New(ctx, "ParseQueryResponse")
```
to

```go
log, _ := spanlogger.New(ctx, "ParseQueryResponse")
```

Fixes SA4006

### DIFF
--- a/pkg/queryfrontend/labels_codec.go
+++ b/pkg/queryfrontend/labels_codec.go
@@ -197,7 +197,7 @@ func (c labelsCodec) DecodeResponse(ctx context.Context, r *http.Response, req q
 		body, _ := ioutil.ReadAll(r.Body)
 		return nil, httpgrpc.Errorf(r.StatusCode, string(body))
 	}
-	log, ctx := spanlogger.New(ctx, "ParseQueryResponse") //nolint:ineffassign,staticcheck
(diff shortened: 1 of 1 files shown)


squat (approved):
Thanks!

pkg/queryfrontend/labels_codec.go
```diff
@@ -197,7 +197,7 @@ func (c labelsCodec) DecodeResponse(ctx context.Context, r *http.Response, req q
 		body, _ := ioutil.ReadAll(r.Body)
 		return nil, httpgrpc.Errorf(r.StatusCode, string(body))
 	}
-	log, ctx := spanlogger.New(ctx, "ParseQueryResponse") //nolint:ineffassign,staticcheck
```
onprem, line 200: I think this was intentional. Let's say if someone tried to create another span in this function, they need to use the updated `ctx` coming from `spanlogge.New` but they might miss that.

I am not sure though so I'll wait for someone else to review.

kakkoyun: Maybe we can add a proper comment to give a hint to the future developers and remove `nolint` statement and the variable. This is definitely intentional.

PR #4160, opened 2021-05-04
(bugsbash)fixes gosec warning potential file inclusion via variable

asquare14:

## Changes

gosec was showing the following warning "Potential file inclusion via variable". This is because we are trying to open files using dynamic variables. Hence, I've cleaned the bad file paths using filepath.Clean()

Another observation, most of the file paths for which the ```gosec``` was showing a warning, they will never have user-inputted filenames, so I think it is sufficient to remove bad file paths. If it is a user inputted file path, having a basePath joined to the incoming path would be better. 

References
[filePath.Clean()](https://golang.org/pkg/path/filepath/#Clean)
[Stackoverflow](https://stackoverflow.com/questions/52320708/how-to-handle-gosec-linter-warning-potential-file-inclusion-via-variable)

### DIFF
--- a/cmd/thanos/tools.go
+++ b/cmd/thanos/tools.go
@@ -50,7 +50,7 @@ func checkRulesFiles(logger log.Logger, patterns *[]string) error {
 			continue
 		}
 		for _, fn := range matches {
-			level.Info(logger).Log("msg", "checking", "filename", fn)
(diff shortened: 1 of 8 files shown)


squat (approved):
:+1: :+1: :+1:

PR #4163, opened 2021-05-04
remove the old ui

Nexucis:

## Verification

Manual testing by starting the binary locally with the query/ store mode


Nexucis:
alright thanks to @onprem we managed to cleanup a bit more the code of the bucket UI.

I believe a refactoring of the way to serve the UI should be done. I'm feeling we can simplify things but I think that should be done in another PR. This cleanup is already good :D.
And also I think this refactoring should be done in Prometheus first.

So anyway we should be good to merge now :). And let's for the refactoring  later.

bwplotka:
Perfect PR: 
image ((: 

> And also I think this refactoring should be done in Prometheus first. 

Do you suggest it's a requirement for this PR?

Nexucis:
Oh now I mean if a refactoring should be performed to simplify the way to serve the UI, I believe it should be done first on Prometheus.

This PR doesn't imply any refactoring, so we are good to merge.
Removing the old Ui in Prometheus will be done as well, but there are some issues to solve before being able to do that (on Prometheus side)

Nexucis:
@onprem I updated the PR regarding the last changes on the `main` branches.
The only thing that remains is the build of the `assets`. I have some trouble on my current setup (nothing related to this PR), so I can't build it somehow.

But until I found why my laptop doesn't want to run this command, could you (again) take a look at it and tell me if it should be ok to merge it once the `assets` are updated ?

Nexucis:
or if you want, since I allowed edits by maintainer of this PR, you can run the command and push it to the PR.

From my side, excepting the `assets`, it's ok to merge it.

roidelapluie:
Be careful that the new UI has some big flaws:

https://github.com/prometheus/prometheus/issues/8256

https://github.com/prometheus/prometheus/issues/8256#issuecomment-843586555

Nexucis:
ah yeah the color issue can be a bit annoying indeed.
But in the other hand, the benefits of removing the old UI is quite interesting as well. And maybe when having only one UI will increase the contribution on this one only. ( yeah that's a bit an hostage situation hem :D)

If Thanos maintainer think it's a blocking issue and actually we should make some effort to fix it first on Prometheus side first ( the one you linked @roidelapluie and also others like https://github.com/prometheus/prometheus/issues/6881), fine for me.

It can make sense to be able to remove the UI first on Prometheus side and then here. It will just take much more time

onprem:
> But until I found why my laptop doesn't want to run this command, could you (again) take a look at it and tell me if it should be ok to merge it once the assets are updated ?

No worries, I'll take a look at the assets part.

Also, I am in favor of delaying this PR a bit and try to get the issues mentioned by @roidelapluie fixed first. I agree that removing the old UI will remove a lot of friction on the development side, but I don't want to have that at the cost of user inconvenience.

Nexucis:
alright no problem :). I think I can decline this PR. We can reopen it or recreate it once it is the good time ^^. I don't see a good reason to keep it open

Nexucis:
we removed the old ui in Prometheus. So we can restart this one \o/

Nexucis:
but probably you will want to get the perfomance improvement done in the Target, Alert and Service Discovery paged before removing it

PR #4164, opened 2021-05-04
Changed md5 algorithm to sha256

nisarg1499:
Signed-off-by: Nisarg Shah nisshah1499@gmail.com

## Changes
Detected MD5 hash algorithm which is considered insecure. MD5 is not collision resistant and is therefore not suitable as a cryptographic signature. Use SHA256 or SHA3 instead.

md5.Sum() is changed to sha256.Sum256()


pkg/receive/config.go
```diff
@@ -290,7 +290,7 @@ func parseConfig(content []byte) ([]HashringConfig, error) {
 
 // hashAsMetricValue generates metric value from hash of data.
 func hashAsMetricValue(data []byte) float64 {
-	sum := md5.Sum(data)
+	sum := sha256.Sum256(data)
```
yeya24, line 293: I am not sure if this change is necessary. Sha256 is more secure than md5. However, it is not used for the security purpose here. So md5 is fine and Sha256 is even slower.

bwplotka: @yeya24 is right. Is there any way you can ignore Muse to getting us this error if we know about this?

nisarg1499: @bwplotka I don't think so. Other option could be to use SHA1 or SHA2 algorithm. But if we are going to see the speed, than MD5 is the fastest.

nisarg1499: Yes of course. Also xxhash is faster than md5. Thanks @roidelapluie
Should I go with xxhash? @yeya24 @bwplotka

kakkoyun: These tools are supposed to help us. I don't understand why we should just change things to make some tool happy.

nisarg1499: @kakkoyun xxhash is just a suggestion if we want faster algorithm.

kakkoyun: I was questioning the overall purpose of the PR. I'm ok to try xxhash.

nisarg1499: Yaa that I agree with you. :) If we don't need it and if the tool is detecting, we can ignore that.

nisarg1499: So should I move ahead trying xxhash or close this PR?

yeya24:
@nisarg1499 Hi, I would say let's try xxhash. Would you still be interested in continuing this pr?

nisarg1499:
> @nisarg1499 Hi, I would say let's try xxhash. Would you still be interested in continuing this pr?

Hey @yeya24, I would be interested in continuing this PR. I will add the xxhash algorithm. 
One thing to clear out, md5 algorithm guarantees 64 length output. But xxhash has 10 length output. So in the code, is there any dependency on length of the answer?

nisarg1499:
@yeya24 Can you give approval to run workflows?

nisarg1499:
> @nisarg1499 Hi, I would say let's try xxhash. Would you still be interested in continuing this pr?

@yeya24 Can you review the changes?

yeya24:
@kakkoyun @bwplotka  PTAL.

kakkoyun (commented):
I still think we don't actually need this change. I'm inclined to reject, however having this change won't hurt. I'm leaving the decision to @bwplotka

roidelapluie:
For the alertmanager we use md5 for this and we do not plan to change.

kakkoyun:
@nisarg1499 Thanks for this contribution. However we don't think, it worths this change. This is used in neither a performance-critical path nor a security-sensitive path. So it doesn't make any practical sense to have this change.

Please don't get discouraged by this, there are plenty of other opportunities to contribute to Thanos :hugs:

PR #4169, opened 2021-05-04
Query: Add a new Metric (Gauge) to keep track of whether a given store is up or down

### LINKED ISSUE #4124: Add metric around failed Store endpoints from Query node

We have a use case where we'd like to alert/monitor on Store instances that a given Query node can't reach (either because it went down, has a network failure, etc). I see entries in the log for these failures but no metrics that represent that. I noticed there is a metric for DNS lookups though so maybe this could be an easy extension to that?

Example log statement I am referring to

```
 level=warn ts=2021-04-29T21:03:20.75684565Z caller=storeset.go:458 component=storeset msg="update of store node failed" err="getting metadata: fetching store info from test.net:10901: rpc error: code = DeadlineExceeded desc = context deadline exceeded" address=test.net:10901
```

A metric available at `/metrics` that shows a `up` style metric whether this endpoint can be reached or not

### Additional context

Happy to work on this if someone could point me in the right direction. Thanks!

dgrizzanti:
Signed-off-by: David Grizzanti <david_grizzanti@comcast.com>

Fixes #4124 

## Changes

- Added a new gauge Metric to track the up/down status of a given store address as discussed in #4124 
- Primarily intended to track when a given store goes down for an extended period of time and is removed from the active store list

## Verification

I haven't added any unit tests around instrumentation as it seems that is not always done but wasn't sure if there is something I should/could add to e2e tests.

- Validated gauge value is `1` for a given store address when store is shown on the Stores page
- Validated gauge value is `0` for a given store address when store is not accessible/removed from the Stores page

### DIFF
--- a/pkg/query/storeset.go
+++ b/pkg/query/storeset.go
@@ -274,8 +274,8 @@ func NewStoreSet(
 
 		upStatus: promauto.With(reg).NewGaugeVec(prometheus.GaugeOpts{
 			Name: "thanos_store_status",
-			Help: "Status of a given store node address.",
-		}, []string{"addr"}),
+			Help: "Status of a given Store API. A value of 1 means store API available for Querier",
+		}, []string{"external_labels", "store_type"}),
 	}
 	return ss
 }
@@ -630,9 +630,9 @@ func (s *StoreSet) updateStoreStatus(store *storeRef, err error) {
 		status.MinTime = mint
 		status.MaxTime = maxt
 		status.LastError = nil
-		s.upStatus.WithLabelValues(store.addr).Set(1)
+		s.upStatus.WithLabelValues(labelpb.PromLabelSetsToString(store.LabelSets()), store.StoreType().String()).Set(1)
 	} else {
-		s.upStatus.WithLabelValues(store.addr).Set(0)
+		s.upStatus.WithLabelValues(labelpb.PromLabelSetsToString(store.LabelSets()), store.StoreType().String()).Set(0)
 		status.LastError = &stringError{originalErr: err}
 	}
 
--- a/pkg/query/storeset.go
+++ b/pkg/query/storeset.go
@@ -16,6 +16,7 @@ import (
 	"github.com/go-kit/kit/log/level"
 	"github.com/pkg/errors"
 	"github.com/prometheus/client_golang/prometheus"
+	"github.com/prometheus/client_golang/prometheus/promauto"
 	"github.com/prometheus/prometheus/pkg/labels"
 	"github.com/thanos-io/thanos/pkg/exemplars/exemplarspb"
 	"google.golang.org/grpc"
@@ -217,6 +218,8 @@ type StoreSet struct {
 	// Map of statuses used only by UI.
 	storeStatuses         map[string]*StoreStatus
 	unhealthyStoreTimeout time.Duration
+
+	upStatus *prometheus.GaugeVec
 }
 
 // NewStoreSet returns a new set of store APIs and potentially Rules APIs from given specs.
@@ -268,6 +271,11 @@ func NewStoreSet(
 		stores:                make(map[string]*storeRef),
 		storeStatuses:         make(map[string]*StoreStatus),
 		unhealthyStoreTimeout: unhealthyStoreTimeout,
+
+		upStatus: promauto.With(reg).NewGaugeVec(prometheus.GaugeOpts{
+			Name: "thanos_store_status",
+			Help: "Status of a given store node address.",
+		}, []string{"addr"}),
 	}
 	return ss
 }
@@ -622,7 +630,9 @@ func (s *StoreSet) updateStoreStatus(store *storeRef, err error) {
 		status.MinTime = mint
 		status.MaxTime = maxt
 		status.LastError = nil
+		s.upStatus.WithLabelValues(store.addr).Set(1)
 	} else {
+		s.upStatus.WithLabelValues(store.addr).Set(0)
 		status.LastError = &stringError{originalErr: err}
 	}
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -16,6 +16,7 @@ We use _breaking :warning:_ to mark changes that are not backward compatible (re
 
 - [#4107](https://github.com/thanos-io/thanos/pull/4107) Store: `LabelNames` and `LabelValues` now support label matchers.
 - [#4171](https://github.com/thanos-io/thanos/pull/4171) Docker: Busybox image updated to latest (1.33.1)
+- [#4169](https://github.com/thanos-io/thanos/pull/4169) Query: Add Gauge metric to keep track of whether a given store is up or down
 
 ### Fixed
 -
--- a/pkg/query/storeset_test.go
+++ b/pkg/query/storeset_test.go
@@ -1064,6 +1064,10 @@ func TestUpdateStoreStateLastError(t *testing.T) {
 	for _, tc := range tcs {
 		mockStoreSet := &StoreSet{
 			storeStatuses: map[string]*StoreStatus{},
+			upStatus: prometheus.NewGaugeVec(prometheus.GaugeOpts{
+				Name: "thanos_store_status",
+				Help: "Status of a given store node address.",
+			}, []string{"addr"}),
 		}
 		mockStoreRef := &storeRef{
 			addr: "mockedStore",
@@ -1080,6 +1084,10 @@ func TestUpdateStoreStateLastError(t *testing.T) {
 func TestUpdateStoreStateForgetsPreviousErrors(t *testing.T) {
 	mockStoreSet := &StoreSet{
 		storeStatuses: map[string]*StoreStatus{},
+		upStatus: prometheus.NewGaugeVec(prometheus.GaugeOpts{
+			Name: "thanos_store_status",
+			Help: "Status of a given store node address.",
+		}, []string{"addr"}),
 	}
 	mockStoreRef := &storeRef{
 		addr: "mockedStore",
--- a/pkg/query/storeset_test.go
+++ b/pkg/query/storeset_test.go
@@ -13,6 +13,7 @@ import (
 	"time"
 
 	"github.com/pkg/errors"
+	"github.com/prometheus/client_golang/prometheus"
 	"google.golang.org/grpc"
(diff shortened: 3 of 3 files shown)


wiardvanrij:
Looks solid! Have you tested it yourself already?

dgrizzanti:
> Looks solid! Have you tested it yourself already?

@wiardvanrij Yep! I updated the PR description with a few comments and added a `CHANGELOG` entry. Let me know if I need to take a look at anything else.

wiardvanrij:
> > Looks solid! Have you tested it yourself already?
> 
> @wiardvanrij Yep! I updated the PR description with a few comments and added a `CHANGELOG` entry. Let me know if I need to take a look at anything else.

I can't think of anything, looks good. Let's wait for more reviewers 🎉

pkg/query/storeset.go
```diff
@@ -268,6 +271,11 @@ func NewStoreSet(
 		stores:                make(map[string]*storeRef),
 		storeStatuses:         make(map[string]*StoreStatus),
 		unhealthyStoreTimeout: unhealthyStoreTimeout,
+
+		upStatus: promauto.With(reg).NewGaugeVec(prometheus.GaugeOpts{
+			Name: "thanos_store_status",
+			Help: "Status of a given store node address.",
+		}, []string{"addr"}),
```
OGKevin, line 278: Addresses have an insane high cardinality, so this should not be used for a label. e.g. a dynamic look-up that returns a list of k8s pod ips.. Instead, use the external_label set.

OGKevin:
```suggestion
		}, []string{"external_labels"}),
```

pkg/query/storeset.go
```diff
@@ -622,7 +630,9 @@ func (s *StoreSet) updateStoreStatus(store *storeRef, err error) {
 		status.MinTime = mint
 		status.MaxTime = maxt
 		status.LastError = nil
+		s.upStatus.WithLabelValues(store.addr).Set(1)
```
OGKevin, line 633:
```suggestion
		s.upStatus.WithLabelValues(store.LabelSets).Set(1)
```

pkg/query/storeset.go
```diff
@@ -622,7 +630,9 @@ func (s *StoreSet) updateStoreStatus(store *storeRef, err error) {
...
 	} else {
+		s.upStatus.WithLabelValues(store.addr).Set(0)
```
OGKevin, line 635:
```suggestion
		s.upStatus.WithLabelValues(store.LabelSets).Set(0)
```

pkg/query/storeset.go
OGKevin: Also, including the label store_type can also be useful for aggregation purposes. e.g. you have 2 unique external label sets for ruler. You only want to send a critical alert if both are down. To achieve this you can aggregate on the store_type label. Without it, it is not easily achievable.

OGKevin (changes requested):

pkg/query/storeset.go
OGKevin:
```suggestion
		}, []string{"external_labels", "store_type"}),
```

bwplotka (commented):
Hi! 

Thanks for this PR, we have others helping with such work couple of time in the past. My first question is... what is the difference between this and existing metrics `thanos_store_nodes_grpc_connections` 🤔 

also @OGKevin comments are valuable too, thanks! I think `addr` could be in some deployments acceptable, but not on everything, maybe it should be added behind the flag?

OGKevin:
The problem with `thanos_store_nodes_grpc_connections` is mentioned https://github.com/thanos-io/thanos/issues/4177

OGKevin:
> Could you expand on what you meant by "behind the flag"

e.g. `thanos query --include-address-in-metrics` or something. A command-line flag.

OGKevin:
> The labels as strings can be big gnarly so let me know what you think..

This should be fine, `thanos_store_nodes_grpc_connections` also exposes the long string that is `external_labels`. :)

OGKevin (dismissed):
LGTM

dgrizzanti:
Cool - thanks @OGKevin. I'm fine leaving the `addr` out and sticking with the label set and type if that works for others. Just need to make sure the tests pass now 🙂

dgrizzanti:
@OGKevin You're probably more familiar with the e2e tests than me - are the e2e errors a result of the code I added or are they flaky/known issues? It's hard to tell looking at the output and running them locally I don't see the same failures. Thanks!

OGKevin:
@dgrizzanti I'm not unfortunate 😞, @bwplotka || @wiardvanrij maybe?

wiardvanrij:
> @dgrizzanti I'm not unfortunate 😞, @bwplotka || @wiardvanrij maybe?

I've pinged on #thanos-dev slack channel because I'm also not really familiair with the e2e part :)

bwplotka (changes requested):
I feel it's still really similar as `thanos_store_nodes_grpc_connections` (in practice, considering Querier stateleness -ness) and it's not reliable as well for cases you explained. Let me explain below.

```go
func newStoreSetNodeCollector() *storeSetNodeCollector {
	return &storeSetNodeCollector{
		storeNodes: map[component.StoreAPI]map[string]int{},
		connectionsDesc: prometheus.NewDesc(
			"thanos_store_nodes_grpc_connections",
			"Number of gRPC connection to Store APIs. Opened connection means healthy store APIs available for Querier.",
			[]string{"external_labels", "store_type"}, nil,
		),
	}
}
```



### Why it's similar:

The only difference that you want to have is that `thanos_store_status` will stay registered for every spotted external_labels & store_type.

However, this is not how the `up` metric works. The up stays registered for certain `job` and `instance` as long as certain `job` and `instance` stays registered. This is why to ensure certain services are scraped we have to alert on `absent(up{job=<thing_I_expect>})`:

> For each instance scrape, Prometheus stores a sample in the following time series:
> 
> up{job="<job-name>", instance="<instance-id>"}: 1 if the instance is healthy, i.e. reachable, or 0 if the scrape failed.

Of course, this is only related to jobs & instances that are added dynamically e.g through kubernetes pod discovery or DNS discovery and when we change `job` to reflect `app` name. If you e.g add static discovery with static local address, ofc `up` will stay always either 0 or 1.

In the same manner, it's still NOT reliable to alert on `thanos_store_status` (and `thanos_store_nodes_grpc_connections`) in case of dynamic discovery (if you want to ensure the certain connection stays up), because of some reasons as you cannot ensure with just checking 0 or 1 with `up` that the expected pod will get scraped. (Metric will be missing if the pod is no longer discovered because it's not scheduled for scrape, so no `up` series for it).

So with `thanos_store_status` we make it behave opposite to `up`. Yet still it's unreliable because Querier can go up and down frequently (it's stateless), so you can have many cases when Querier restarts and in the similar time expected store API gets down, and your alerting will miss that because Querier service discovery will not see it anymore, so no `thanos_store_status` is registered.

### Solution

To me, I don't think this metric will help anyone, even you, because it's not reliable. I think we need to definitely improve docs, but IMO we need to rely on:

* Strict mode (maybe we need to enhance some parts of it) cc @GiedriusS 
* Expecting a certain number of external_labels and store Types to be present in your alerting (WITH `absent` additionally)

Ideas welcome, I don't want to block you and PR is actually great and high quality, but hope I understand why it can be confusing and unreliable ):

OGKevin:
> However, this is not how the up metric works. The up stays registered for certain job and instance as long as certain job and instance stays discovered. This is why to ensure certain services are scraped we have to alert on absent(up{job=<thing_I_expect>}):

_BUT_ in our case, the store _stays_ and _is_ still discovered! Its just down! 

For example: 

Screenshot 2021-05-12 at 10 19 57

I want to be able to alert on this without absent because 

1. It is not absent, it is down.
2. I will need to create an absent for every single external label pair which can be a lot. e.g. 14 clusters with all 9 unique replicas etc.

I want a metric that just beautifully says 0 in this case like the status page does. The fact that the status page keeps the store but `thanos_store_nodes_grpc_connections` goes absent is/feels like a bug to me. At least its mixed behaviour. Either the metrics is there or the status page needs to drop the store. Which it should not as its DNS discovery.

OGKevin:
So it is literally like `up` the target is discovered but down. There is no way to alert on this case without absent at the moment.

bwplotka:
Fully agree @OGKevin , but this is not what this PR proposes. 🤔 

Can we fix our current metric then? (agree with you!)

bwplotka:
and we have PR for this 🤦🏽

OGKevin:
Yea so we can either fix the `thanos_store_nodes_grpc_connections` to not go absent or introduce this new up-like metric. AFAIK the metric will also go absent when the store is actually dropped if I understand the code correctly.

wiardvanrij:
sorry @dgrizzanti  & @OGKevin  I thought to understand this change better and it did made sense. However what @bwplotka explains is solid and yea, we should figure this out.

dgrizzanti:
Thanks for all the feedback @OGKevin @bwplotka @wiardvanrij. There were a lot of comments and ideas thrown out - let me try and summarize where I was coming from, what I wanted out of this metric (that I don't see provided today), and see where we can go from there 🙂 

In my use case for Querier, we have a statically defined list of store/sidecar endpoints (DNS names to be exact). These endpoints can go down, change IPs, etc but the expectation is that an instance at the given DNS name is up at all times. If for some reason one of these goes down or is unreachable from a Query node, we want to alert on that (unreachable from the perspective of the query node). The current `thanos_store_nodes_grpc_connections` doesn't allow for that, unless we use `absent` which is not really practical as @OGKevin pointed out because we'll have to use possibly hundreds of Alert expressions to capture all of the unique label combinations.

The metric I added, at least my intent, was to have something that behaved like the Prometheus `up` metric in the sense that when target are **discovered** but are down/not-scrapeable, you can alert on that fact with something like `sum(thanos_store_status) < 1` and avoid the complication with `absent`.

I realize this may not work or cause confusing for folks using dynamic service discovery but I still think it would still allow for a helpful way of alerting/knowledge of what the status is of the currently discovery but not-reachable targets.

I'm happy to pivot on this if needed and use `thanos_store_nodes_grpc_connections` as an alternative but not sure that would solve my use case (which I think is the same as @OGKevin) unless we come to an agreement on how discovered but unreachable stores are treated. Thanks!

PR #4172, opened 2021-05-04
(bugbash) Fixes G104 GoSec warnings

rajaSahil:

## Changes

- Fixes unhandled errors( G104 GoSec warnings).

### DIFF
--- a/pkg/verifier/duplicated_compaction.go
+++ b/pkg/verifier/duplicated_compaction.go
@@ -28,7 +28,7 @@ func (DuplicatedCompactionBlocks) VerifyRepair(ctx Context, idMatcher func(ulid.
 		return errors.Errorf("id matching is not supported")
 	}
 
-	level.Info(ctx.Logger).Log("msg", "started verifying issue", "with-repair", repair)
+	_ = level.Info(ctx.Logger).Log("msg", "started verifying issue", "with-repair", repair)
 
 	overlaps, err := fetchOverlaps(ctx, ctx.Fetcher)
 	if err != nil {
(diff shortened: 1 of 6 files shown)


pkg/verifier/duplicated_compaction.go
```diff
@@ -28,7 +28,7 @@ func (DuplicatedCompactionBlocks) VerifyRepair(ctx Context, idMatcher func(ulid.
 		return errors.Errorf("id matching is not supported")
 	}
 
-	level.Info(ctx.Logger).Log("msg", "started verifying issue", "with-repair", repair)
+	_ = level.Info(ctx.Logger).Log("msg", "started verifying issue", "with-repair", repair)
```
asquare14, line 31: I think a better option would be to do this :
```go
err = level.Info(ctx.Logger).Log("msg", "started verifying issue", "with-repair", repair)
if err!=nil {
   return err
}
```
Ignoring the error may cause issues.

asquare14 (changes requested):
Same for other ```_``` statements.

pkg/verifier/duplicated_compaction.go
rajaSahil: Will it be correct to return the error?

asquare14: yes, similar to this https://github.com/thanos-io/thanos/blob/main/pkg/verifier/duplicated_compaction.go#L34

rajaSahil: Okay. Will update it. Thank you!

bwplotka (changes requested):
Thank you for this, but I don't think it makes sense. We have tons of info log lines and we never want to check error for this particular operation (we are fine if this fails). Doing `_` for such heavily used thing is also not desired. This is why we are fine with this particular struct to not have error checked. 

That's why we have it in https://github.com/thanos-io/thanos/blob/b4d233e6fe904c52a22948afea3c3002d894e5df/.errcheck_excludes.txt#L1 (:

yeya24:
Close this as I agree with @bwplotka

PR #4176, opened 2021-05-04
querier: Adds additional stats field in query api

### LINKED ISSUE #4170: Support optional query stats parameter

(Write your answer here.)

Prometheus instant query and range query APIs support an optional stats field to return the timing stats of the executed query.

Thanos should implement the same functionality. For implementation, please check the code in Prometheus:
https://github.com/prometheus/prometheus/blob/main/web/api/v1/api.go#L466-L470
https://github.com/prometheus/prometheus/blob/main/web/api/v1/api.go#L387-L391

(Describe your proposed solution here.)

(Write your answer here.)

### Additional context

(Write your answer here.)

someshkoli:
Signed-off-by: someshkoli <kolisomesh27@gmail.com>

## Changes
Adds additional flag `--query.stats` to query command. Adds option  to pass additional paramater `stats` in form data for query API.

Closes #4170 

## Verification
need to add some endpoint tests

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -16,8 +16,7 @@ We use _breaking :warning:_ to mark changes that are not backward compatible (re
 
 - [#4107](https://github.com/thanos-io/thanos/pull/4107) Store: `LabelNames` and `LabelValues` now support label matchers.
 - [#4171](https://github.com/thanos-io/thanos/pull/4171) Docker: Busybox image updated to latest (1.33.1)
-
-- [#4176](https://github.com/thanos-io/thanos/pull/4176) Query API: Adds optional `Stats param`
+- [#4176](https://github.com/thanos-io/thanos/pull/4176) Query API: Adds optional `Stats param` to return stats for query APIs
 ### Fixed
 -
 ### Changed
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -17,6 +17,7 @@ We use _breaking :warning:_ to mark changes that are not backward compatible (re
 - [#4107](https://github.com/thanos-io/thanos/pull/4107) Store: `LabelNames` and `LabelValues` now support label matchers.
 - [#4171](https://github.com/thanos-io/thanos/pull/4171) Docker: Busybox image updated to latest (1.33.1)
 
+- [#4176](https://github.com/thanos-io/thanos/pull/4176) Query API: Adds optional `Stats param`
 ### Fixed
 -
 ### Changed
--- a/pkg/api/query/v1.go
+++ b/pkg/api/query/v1.go
@@ -42,6 +42,7 @@ import (
 	"github.com/prometheus/prometheus/promql/parser"
 	"github.com/prometheus/prometheus/storage"
 
+	"github.com/prometheus/prometheus/util/stats"
 	"github.com/thanos-io/thanos/pkg/api"
 	"github.com/thanos-io/thanos/pkg/exemplars"
 	"github.com/thanos-io/thanos/pkg/exemplars/exemplarspb"
@@ -68,6 +69,7 @@ const (
 	MatcherParam             = "match[]"
 	StoreMatcherParam        = "storeMatch[]"
 	Step                     = "step"
+	Stat                     = "stat"
 )
 
 // QueryAPI is an API used by Thanos Querier.
@@ -89,6 +91,7 @@ type QueryAPI struct {
 	enableTargetPartialResponse         bool
 	enableMetricMetadataPartialResponse bool
 	enableExemplarPartialResponse       bool
+	enableQueryStats                    bool
 	disableCORS                         bool
 
 	replicaLabels []string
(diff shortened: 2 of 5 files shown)


yeya24 (changes requested):
Thanks for working on this issue.
I think this can be enabled by default and we don't need a flag.

Also please add some unit tests.

pkg/api/query/v1.go
```diff
@@ -68,6 +69,7 @@ const (
 	MatcherParam             = "match[]"
 	StoreMatcherParam        = "storeMatch[]"
 	Step                     = "step"
+	Stat                     = "stat"
```
yeya24, line 72: This should be `stats`.

pkg/api/query/v1.go
```diff
@@ -283,12 +288,22 @@ func (qapi *QueryAPI) parseStep(r *http.Request, defaultRangeQueryStep time.Dura
 		}
 		return defaultRangeQueryStep, nil
 	}
-
 	// Default step is used this way to make it consistent with UI.
 	d := time.Duration(math.Max(float64(rangeSeconds/250), float64(defaultRangeQueryStep/time.Second))) * time.Second
 	return d, nil
 }
 
+func (qapi *QueryAPI) parseStat(r *http.Request, defaultEnableStat bool) (bool, *api.ApiError) {
```
yeya24, line 296: We don't need this function. Like the code in Prometheus, if stats is not an empty string, then we return the query stats.

someshkoli:
> Thanks for working on this issue.
> I think this can be enabled by default and we don't need a flag.
> 
> Also please add some unit tests.

Alright, ill update the changes

yeya24 (changes requested):

pkg/api/query/v1.go
```diff
@@ -332,6 +333,10 @@ func (qapi *QueryAPI) query(r *http.Request) (interface{}, []error, *api.ApiErro
 		return nil, nil, apiErr
 	}
 
+	if apiErr != nil {
+		return nil, nil, apiErr
+	}
```
yeya24, line 338: Remove this?

pkg/api/query/v1.go
```diff
@@ -439,6 +450,10 @@ func (qapi *QueryAPI) queryRange(r *http.Request) (interface{}, []error, *api.Ap
...
```
yeya24, line 455: ditto

pkg/api/query/v1_test.go
```diff
@@ -72,8 +73,21 @@ type endpointTestCase struct {
 	response interface{}
 	errType  baseAPI.ErrorType
 }
+type responeCompareFunction func(interface{}, interface{}) bool
 
-func testEndpoint(t *testing.T, test endpointTestCase, name string) bool {
+// Does a deep compare for two http responses.
+func deepCompare(a interface{}, b interface{}) bool {
+	return reflect.DeepEqual(a, b)
+}
```
yeya24, line 81: Why do we need this function? reflect.DeepEqual has the same signature, right?

yeya24 (commented):
Some small nits. Also it would be good if you can update the changelog

pkg/api/query/v1_test.go
someshkoli: Yeah it does, but I wanted to check if `Stats` is being returned in query result and this could'nt be achieved if we do a deepEqual as Stats gives non consistent values hence added a compare function with same signature and added `lookupStats` function

someshkoli: Also any idea why e2e is failing ? :|

someshkoli: > Why do we need this function? reflect.DeepEqual has the same signature, right?

Oh, I think I got your point. I'll fix it :p

someshkoli:
> Also it would be good if you ca

Also a stupid question, when and when not to update changelog ?

yeya24:
Hi @someshkoli, please rebase on main then we can get CI fixed.

CHANGELOG.md
```diff
@@ -17,6 +17,7 @@ We use _breaking :warning:_ to mark changes that are not backward compatible (re
 - [#4107](https://github.com/thanos-io/thanos/pull/4107) Store: `LabelNames` and `LabelValues` now support label matchers.
 - [#4171](https://github.com/thanos-io/thanos/pull/4171) Docker: Busybox image updated to latest (1.33.1)
 
+- [#4176](https://github.com/thanos-io/thanos/pull/4176) Query API: Adds optional `Stats param`
```
yeya24, line 20: Please remove the blank line above and add one blank line below.
`Adds optional `Stats param` to return stats for query APIs` would be better.

yeya24 (approved):
LGTM. Will merge on green.

someshkoli:
> LGTM. Will merge on green.

yey all greens :tada:

yeya24:
> > Also it would be good if you ca
> 
> Also a stupid question, when and when not to update changelog ?

When your pr changes some known behaviors or are user-facing. IMO this change provides a new param so it is user-facing.

PR #4185, opened 2021-05-05
go.mod: bump Go version

GiedriusS:
Bump the Go version in `go.mod` to make it the same as in other places.

Fixes https://github.com/thanos-io/thanos/commit/67f906e663bf1f93196e7c53df3bf72c9538dd4b.

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -94,4 +94,4 @@ replace (
 	k8s.io/klog/v2 => github.com/simonpasquier/klog-gokit/v2 v2.0.1
 )
 
-go 1.15
+go 1.16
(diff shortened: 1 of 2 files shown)


squat:
AFAIK bumping the go version in the mod file is only needed if we want to declare that we require features from go 1.16, otherwise we can bump the version of go used to build the project by changing our go-build docker container version

GiedriusS:
> AFAIK bumping the go version in the mod file is only needed if we want to declare that we require features from go 1.16, otherwise we can bump the version of go used to build the project by changing our go-build docker container version

I have been chasing some `go.mod` issues and stumbled upon this. (FWIW I ran into this change her: https://github.com/golang/go/issues/43994#issuecomment-770053099). @squat you are correct however then we should update this line here: https://github.com/thanos-io/thanos/commit/67f906e663bf1f93196e7c53df3bf72c9538dd4b#diff-31bcba2ccafa41d46fbbd6d1219f7f1e3b1fb3cad9faa8e4dc521bbb579dd7b3R48. I guess I can do that in another PR?

squat:
Ah, yeah that's too bad, we should probably change that line too

PR #3410, opened 2020-11-04
compact: Added index size limiting planner detecting output index size over 64GB.

bwplotka:
Chained with https://github.com/thanos-io/thanos/pull/3409

Fixes: https://github.com/thanos-io/thanos/issues/1424

Signed-off-by: Bartlomiej Plotka <bwplotka@gmail.com>

PR #4191, opened 2021-05-06
Vertical Block Sharding Proposal

Biswajitghosh98:
This is the first draft of the vertical block sharding proposal. Open for reviews :)


docs/proposals/202011_vertical_block_sharding.md
```diff
@@ -87,17 +88,66 @@ index size).
...
 
 ## Alternatives
 
-* Shard number stored in external label.
+* Keep the system as it is.
+* In the above approach, keep the number of shards dynamic, depending on the use case.
+  * Pros
+    * We can dynamically adjust to the situation. If a certain block stream has only few metrics, we can keep 1 shard, and if during the 2w period, suddenly we have 100 millions series (or some number beyond index size), we can horizontally split to 10 or more. And then later back to 1.
+   * Cons
+     * With dynamically changing metas, it would become difficult to group blocks together for vertical compaction.
```
bwplotka, line 150: But also horizontal compaction

docs/proposals/202011_vertical_block_sharding.md
```diff
@@ -87,17 +88,66 @@ index size).
...
+.
+.
+.
+hash_number_n : 0
+```
+Wherein, n is the number of times we want a blocks index to hit the limit. 
+The current method ignores a block for further compaction once its index reaches a limit. We'd let that stay as it is, with a caveat that whenever a blocks index size hits the limit, we'd find the first 0-valued hash_number (let's say i), then further sub-divide (shard) the existing block with the corresponding `hash_function_i`, and then let the compaction process run as usual.
+Here also, we'd allow grouper and planner to compact (horizontally and vertically) only those blocks together whose `hash_number` external labels match completely. 
+The advantage of this method is, that it would allow us to help grow the size of the block in terms of time difference  (`maxt - mint`) exponentially, so we can make an estimated guess that n would not exceed 3 or 4. 
+The negative aspect of this approach would be, as blocks grow larger, it would become difficult to find a subsequent block to compact it with.
```
bwplotka, line 124: I would also add `hash function` to metada

docs/proposals/202011_vertical_block_sharding.md
```diff
@@ -87,17 +88,66 @@ index size).
...
 ## Proposal
 
-**We propose to vertically shard all blocks which goes beyond the specified index size limit during the compaction process.**
+**We propose to vertically shard all blocks which goes beyond the specified index size limit during the compaction process, else efficiently shard the blocks at compaction level 0.**
 
-* Special metadata in meta.json indicating a shard.
+* We'll be using a special metadata in meta.json indicating a shard.
 
-TBD
+#### Approach 1: Sharding at compaction level 0
```
bwplotka, line 95: Why two approaches? Can we have just one, decoupled to level?

docs/proposals/202011_vertical_block_sharding.md
```diff
@@ -87,17 +88,66 @@ index size).
...
+ * Intra-shard horizontal and vertical compaction would mean less variance in metrics and hence we can expect a larger block in terms of the difference in `maxt` and `mint` when a certain blocks index hits the limit, and then marked with `noCompact`.
+
+#### Approach 2: Sharding a block when its index size hits the limit
+
+This approach is a slight modification to Apporach 1. 
+Instead of creating just one external label `hash_number`, we can create a series of hash_numbers and corresponding `hash_function`s.
+This is also a static method, but an optimization over Approach 1. 
+The way it would work is, intially during block creation, we'll add `n` external labels 
+```
+hash_number_1 : 0
```
bwplotka, line 113: I think we can work on better representation maybe using map as we discussed offline (:

docs/proposals/202011_vertical_block_sharding.md
```diff
@@ -77,6 +77,7 @@ index size).
 
 * Automatically cap the size of index per block to X GB without impacting read performance.
   * Optional: CLI for vertical sharding of old blocks.
```
bwplotka, line 79: Let's remove this, let's not suggest implementation

docs/proposals/202011_vertical_block_sharding.md
```diff
@@ -77,6 +77,7 @@ index size).
 
 * Automatically cap the size of index per block to X GB without impacting read performance.
   * Optional: CLI for vertical sharding of old blocks. 
+* Alternatively, increase the number of compactions before index size hits the current limitations, to improve performance of querying.
```
bwplotka, line 80:
```suggestion
* Allow grouping series within blocks that spans over larger time windows defined in compaction level (e.g up to 2w), to leverage downsampling etc
```

bwplotka (commented):
Looks actually guide good, especially when you elaborated it to me! (: 

Some ideas during our 1:1

* Add  some diagram to make it easier to understand (similar to what you draw me)
* Elaborate test cases we discussed offline
   * How many resources/time would be needed to perform such multi shard compaction? Is there anything blocking to make streaming?
* plus minor suggestions below.

docs/proposals/202011_vertical_block_sharding.md
```diff
@@ -87,17 +88,66 @@ index size).
...
 
+For this approach, we'd begin by adding an external label, `hash_number` during block creation, which would denote the shard number. We can keep it 0, because subsequently the hash function would return values only between 1 and X (inclusive), so there won't be any chance of a possible clash. During the first compaction of the initially created 2h blocks, we'll divide each block into X blocks. The block number i (1<=i<=X) would contain only those series whose hash value of mertic names is i.
+
+Once this step is completed, we can be certain that all the series that belong to a certain block have the same `hash_number`. Now, we'll allow grouper and planner to group only those blocks together which have the same `hash_number`. This would serve us two purposes : 
+
+ * Allow grouper to group blocks in such a way that we can run X concurrent compactor processes.
+ * Intra-shard horizontal and vertical compaction would mean less variance in metrics and hence we can expect a larger block in terms of the difference in `maxt` and `mint` when a certain blocks index hits the limit, and then marked with `noCompact`.
+
+#### Approach 2: Sharding a block when its index size hits the limit
```
bwplotka, line 106: Let's mention binary sharding approach (:

docs/proposals/202011_vertical_block_sharding.md
Biswajitghosh98: Even better, I used a single integer, and its binary representaion would do the job for us 😛

docs/proposals/202011_vertical_block_sharding.md
Biswajitghosh98: Done !

docs/proposals/202011_vertical_block_sharding.md
```diff
@@ -87,17 +87,49 @@ index size).
...
+* If the compacted block is valid, then continue the usual compaction process until it results into an invalid block.
+* If the compacted block is invalid, then add the no compact filter to all the participating blocks.
+
+We can keep growing the binary tree by recursively using the same logic. The reason for calling this adaptive is because at any point of time, the total number of shards is equal to the number of nodes on the tree (say x). If any of the leaves overloads (a block in that particular shards hits the limit), that particular leaf would split into 2, effectively increasing the number of shards to x+2, hence providing just the right amount of shards to get the job done.
```
hitanshu-mehta, line 108: I think the total number of shards should be equal to the number of leaf nodes and every vertical sharding will increase the number of shards to x + 1.

docs/proposals/202011_vertical_block_sharding.md
```diff
@@ -87,17 +87,49 @@ index size).
...
+#### Pros 
+
+* This layered sharding approach would statistically give more evenly distributed shards as compared to using only one hash function and hence sharding the blocks into some x    shards. For layered binary sharding, let's assume that all the 16 hash functions are bad hash functions, and each one of them gives a skewed distribution, even then we'll get a fairly even distribution after sharding through all the layers. This is pictorially explained below ![vbs-image4](../img/vertical_block_sharding_image4.png)
+* The process is adaptive, hence any user can opt to go for default configuration and not worry about setting the number of shards.
+* If for some reasons, the user decides to decrease the number of shards, it can be done by setting a `max_shard_level_limit` to any value between 0 and 16.
+
+#### Cons
+
+* We didn't solve the problem completely, but we delayed it as much as we could to get some benefits out of the delaying process.
+* Cases were large number of blocks participate in a compaction process and result in an invalid block would cause the sharding algorithm to be triggered, even though sharding at that point might not be necessary.
```
hitanshu-mehta, line 126:
```suggestion
* Cases where large number of blocks participate in a compaction process and result in an invalid block would cause the sharding algorithm to be triggered, even though sharding at that point might not be necessary. 
```
small typo

docs/proposals/202011_vertical_block_sharding.md
Biswajitghosh98: @hitanshu-mehta Say we have 10 blocks in shard (0,0), and we're compacting 3 blocks out of those together. Suppose the resultant block hits the limit, then we'll shard the 3 participating blocks, leaving 7 blocks at shard (0,0), 3 at (1,0) and 3 at (1,1), increasing the number of shards from 1 to 3

docs/proposals/202011_vertical_block_sharding.md
```diff
@@ -87,17 +87,49 @@ index size).
...
 
-* Special metadata in meta.json indicating a shard.
+* We'll be using a special metadata in meta.json indicating a shard.
 
-TBD
+### Adaptive Binary Sharding
 
+The idea behind this approach is to leverage on the structural benefits of a binary tree to adaptively shard blocks which hit the maximum limit, either in terms of index size or number of series (set by user). 
 
+For this approach, we'd need to add two extra external label `hash_number`, initialized to 0, `shard_level` initialized to 0, denoting the number of times it has been sharded during the compaction process, and a set of 16 `hash_function`s. The reason for chosing 16 is because 2 raised to the power 16 is 65536, which is a logical upper bound for the number of shards a user might want to have.
```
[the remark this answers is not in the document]
Biswajitghosh98: @hitanshu-mehta Yes. Apologies if it was unclear from my end :)

docs/proposals/202011_vertical_block_sharding.md
Biswajitghosh98: Thanks !

docs/proposals/202011_vertical_block_sharding.md
hitanshu-mehta: Thank you for clearing things up! I was calculating shards the different(maybe wrong) way. To give a rough idea how I calculated, 

block_number(hash_number, hash_level)


	       1(0,0)             2(0,0)            3(0,0)         *4(0,0)    Compaction of blocks 1(0,0), 2(0,0) and 3(0,0) hit the limit
         /        \         /        \         /       \   
      1(0,1)   *1(1,1)   2(0,1)   *2(1,1)     3(0,1)  *3(1,1)             Compaction of blocks 1(0,1), 2(0,1) and 3(0,1) hit the limit
     /      \           /    \              /    \
    *1(0,2) *1(1,2)   *2(0,2) *2(1,2)     *3(0,2) *3(1,2)

The total number of stared(*) blocks is the number of shards.

docs/proposals/202011_vertical_block_sharding.md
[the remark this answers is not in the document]
hitanshu-mehta: No worries :) I think we should use the same term everywhere in the proposal because it can lead up to confusion sometimes. WDYT?

docs/proposals/202011_vertical_block_sharding.md
Biswajitghosh98: @hitanshu-mehta Oh, I see. But this is different from what I had in mind. The proposal would need more pictures and re-wording to make it more clear for everyone, I suppose :)

docs/proposals/202011_vertical_block_sharding.md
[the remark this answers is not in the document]
Biswajitghosh98: @hitanshu-mehta Yes, definitely. I'll change it and add more pictures ASAP

yeya24:
This idea is similar to the mimir split and merge compact. Can we revisit this proposal?

docs/proposals/202011_vertical_block_sharding.md
```diff
@@ -87,17 +87,49 @@ index size).
 
 ## Proposal
 
-**We propose to vertically shard all blocks which goes beyond the specified index size limit during the compaction process.**
+**We propose to vertically shard blocks when it reaches the cap limit in terms of number of series or size of index file, and then adaptively decide the number of shards depending on the users requirement**
 
-* Special metadata in meta.json indicating a shard.
+* We'll be using a special metadata in meta.json indicating a shard.
```
fpetkovski, line 92: Could this be another external label?

docs/proposals/202011_vertical_block_sharding.md
```diff
@@ -87,17 +87,49 @@ index size).
...
 
-TBD
+### Adaptive Binary Sharding
 
+The idea behind this approach is to leverage on the structural benefits of a binary tree to adaptively shard blocks which hit the maximum limit, either in terms of index size or number of series (set by user). 
 
+For this approach, we'd need to add two extra external label `hash_number`, initialized to 0, `shard_level` initialized to 0, denoting the number of times it has been sharded during the compaction process, and a set of 16 `hash_function`s. The reason for chosing 16 is because 2 raised to the power 16 is 65536, which is a logical upper bound for the number of shards a user might want to have.
 
-## Alternatives
+We'll be using `hash_level` and `hash_number` to group and plan blocks to compact together. The way it'd work is, we'll allow grouper to group only those blocks together whose `hash_level`s are same. Also, we'd allow planner to compact only those blocks together that share the same `hash_number`. So, with this, the compactor would run as it is unless a compacted block has hit the limit for the first time.
```
fpetkovski, line 100: Does this mean blocks can get merged even if they're not part of the same stream (have the same external labels)?

docs/proposals/202011_vertical_block_sharding.md
```diff
@@ -87,17 +87,49 @@ index size).
...
+
+We're allowing a compacted block to hit the limit for a maximum of 16 times (if not specified otherwise by the user), and at the 16th level, if/when it further compacts and hits the limit, we're marking it for no more further compactions. The algorithm for sharding would come into play once a compacted block is declared invalid, or its size hits the limit. If the compacted block is still valid, but its index size is too large (or the number of series it holds exceeds the limit decided by the user), we'll look at the current `hash_level`(say `hl`) of the compacted block (if valid), or the blocks participating in the compaction process (if the compacted block is invalid), then we'll either shard the compacted block or the participating blocks (depending on the situation), `using hash_function[hl+1]`, and hence set the `hash_level` of the newly created blocks to `hl+1`, and then upload them for further compaction processes, if `hl` is not equal to `max_shard_level_limit`, which is 16 by default. In case `hl` is equal to `max_shard_level_limit`, we will take one of the 2 decisions depending on whether the compacted block is valid or invalid.
```
fpetkovski, line 104: What does it mean for an index size to be too large? Will this be configurable?

docs/proposals/202011_vertical_block_sharding.md
yeya24: Yes, I think `hash_number` and `shard_level` are external labels as mentioned below.

docs/proposals/202011_vertical_block_sharding.md
yeya24: Yeah, I think it is not mentioned explicitly but all external labels should be the same during the grouping.

docs/proposals/202011_vertical_block_sharding.md
yeya24: Yeah I think it should be configurable. We have supported planning considering index size from https://github.com/thanos-io/thanos/pull/3410 and the index limit is configurable.

yeya24:
Can you describe more about the cuckoo filter part? How can we use it to improve the query performance at store gateway? Is it related to this proposal?

docs/proposals/202011_vertical_block_sharding.md
```diff
@@ -87,17 +87,49 @@ index size).
 
 ## Proposal
 
-**We propose to vertically shard all blocks which goes beyond the specified index size limit during the compaction process.**
+**We propose to vertically shard blocks when it reaches the cap limit in terms of number of series or size of index file, and then adaptively decide the number of shards depending on the users requirement**
```
yeya24, line 90: Do we still want to wait for the block reaching the cap limit to do sharding? With vertical sharding, I think we want to shard the blocks eagerly to get better query performance.

docs/proposals/202011_vertical_block_sharding.md
```diff
@@ -87,17 +87,49 @@ index size).
...
+#### Cons
+
+* We didn't solve the problem completely, but we delayed it as much as we could to get some benefits out of the delaying process.
+* Cases where large number of blocks participate in a compaction process and result in an invalid block would cause the sharding algorithm to be triggered, even though sharding at that point might not be necessary. 
+
+### Design Decisions
+
+#### Decision 1: Using size of index file vs number of metrics in a block as an indicator for sharding
+
+The pros and cons for selecting one design choice over the other is yet to be discovered, and would be more clear after implementation and testing.
```
fpetkovski, line 132: I wonder how this approach compares to the current workaround of setting a maximum index size per block.

PR #4193, opened 2021-05-06
label outer for loop breaks in switch cases

arcolife:
Signed-off-by: arcolife <archit.py@gmail.com>

## Changes

as the title

### DIFF
--- a/pkg/reloader/reloader_test.go
+++ b/pkg/reloader/reloader_test.go
@@ -111,10 +111,11 @@ config:
 
 	reloadsSeen := 0
 	attemptsCnt := 0
+Outer:
 	for {
 		select {
 		case <-ctx.Done():
-			break
+			break Outer
 		case <-time.After(300 * time.Millisecond):
 		}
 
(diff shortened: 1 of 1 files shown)


pkg/reloader/reloader_test.go
```diff
@@ -114,7 +114,6 @@ config:
 	for {
 		select {
 		case <-ctx.Done():
-			break
```
squat, line 117: I think this is a solution to the wrong bug actually! I suspect that this case actually wanted to break out of the `for` loop rather than the `select`. To fix this, we would need to put a label on the loop, e.g `Outer:` on top of the `for` loop and then change this line to `break Outer`

squat (commented):
Thanks for finding this. I think that the but is actually a bit different. More context in-line :)

pkg/reloader/reloader_test.go
arcolife: that makes total sense - accidental bliss :D Thanks!

bwplotka (dismissed):

squat (dismissed):
exactly :) thanks!

arcolife:
@bwplotka e2e CI issues ?

onprem:
There was a issue with the busybox image used in CI. This has been fixed in `main`, can you rebase?

squat (approved):

PR #4197, opened 2021-05-06
Updated the receiver-split proposal

yashrsharma44:
Did some grammar and spelling changes in the receiver proposal, while I was going through it.


yashrsharma44:
Pushed in a wrong branch :(

PR #4200, opened 2021-05-07
(bugsmash) : defer should not be in for loop

asquare14:

Used the tool https://github.com/gostaticanalysis/loopdefer to detect these.


yeya24:
Hi @asquare14, please rebase on main then we can get CI fixed.

PR #4171, opened 2021-05-04
Updates busybox to latest

wiardvanrij:
Signed-off-by: Wiard van Rij <wiard@outlook.com>

* [x] I added CHANGELOG entry for this change.
* [ ] Change is not relevant to the end user.

## Changes

Updates busybox image to latest (via sha hash)

Fixes: https://github.com/thanos-io/thanos/issues/4167

PR #4202, opened 2021-05-07
Updated busybox sha; Added dependabot.

bwplotka:

### DIFF
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -1,0 +1,14 @@
+version: 2
+updates:
+  - package_manager: "gomod"
+    directory: "/"
+    vendor: false
+    schedule:
+      interval: "weekly"
+    labels: ["dependencies"]
+  - package_manager: "docker"
+    directory: "/"
+    schedule:
+      interval: "weekly"
+    labels: ["dependencies"]
+
--- a/Makefile
+++ b/Makefile
@@ -5,18 +5,18 @@ DOCKER_IMAGE_REPO ?= quay.io/thanos/thanos
 DOCKER_IMAGE_TAG  ?= $(subst /,-,$(shell git rev-parse --abbrev-ref HEAD))-$(shell date +%Y-%m-%d)-$(shell git rev-parse --short HEAD)
 DOCKER_CI_TAG     ?= test
 
-SHA=''
+BASE_DOCKER_SHA=''
 arch = $(shell uname -m)
 # Run `DOCKER_CLI_EXPERIMENTAL=enabled docker manifest inspect quay.io/prometheus/busybox:latest` to get SHA or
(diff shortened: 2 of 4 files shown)


.github/dependabot.yml
```diff
@@ -0,0 +1,14 @@
+version: 2
+updates:
+  - package_manager: "gomod"
+    directory: "/"
+    vendor: false
+    schedule:
+      interval: "weekly"
+    labels: ["dependencies"]
+  - package_manager: "docker"
```
daixiang0, line 9: I do not see any rule about busybox SHA update rule, is it a magic?

Makefile
```diff
@@ -5,18 +5,18 @@ DOCKER_IMAGE_REPO ?= quay.io/thanos/thanos
...
 arch = $(shell uname -m)
 # Run `DOCKER_CLI_EXPERIMENTAL=enabled docker manifest inspect quay.io/prometheus/busybox:latest` to get SHA or
 # just visit https://quay.io/repository/prometheus/busybox?tag=latest&tab=tags.
 # TODO(bwplotka): Pinning is important but somehow quay kills the old images, so make sure to update regularly (dependabot?)
-# Update at 2020.2.01
+# Update at 2021.6.07
 ifeq ($(arch), x86_64)
     # amd64
-    SHA="14d68ca3d69fceaa6224250c83d81d935c053fb13594c811038c461194599973"
+    BASE_DOCKER_SHA="de4af55df1f648a334e16437c550a2907e0aed4f0b0edf454b0b215a9349bdbb"
```
wiardvanrij, line 16: This sha is different than the default sha's for amd64 in the dockerfiles? Wouldn't these be the same or do we use some fallback?

wiardvanrij:
Sweet to use dependabot :D

Makefile
daixiang0: Since it is updated, notice the date in comment.

.github/dependabot.yml
bwplotka: It actually works with digest, but not sure how it can detect ARM/AMD tbh, so that will not for for now. Still I want to try this path (:

Makefile
bwplotka: We just always set it, maybe I can remove the one in dockefiles

.github/dependabot.yml
daixiang0: let's try it :)

Makefile
wiardvanrij: Isn't it easier that all 'tools' use the same version no? I made this mistake by not understanding/seeing it :p See this PR: https://github.com/thanos-io/thanos/pull/4171

daixiang0:
Need to re-trigger CI

kakkoyun (approved):

PR #4208, opened 2021-05-09
UI: Fix infinite redirection loop on /

### LINKED ISSUE #4128: Block Viewer: --web.route-prefix flag inconsistent/unexpected redirects

Thanos v0.20.0

**Object Storage Provider**:
Not relevant

**What happened**:
Trying to reach the bucketweb tool from a reverse proxy. Setup as following (which works fine with the query component):
- `--web.external-prefix=thanos/bucketweb/`
- `--web.route-prefix=/`

while the reverse proxy (Ambassador) is using:
```
      prefix: /thanos/bucketweb/
      rewrite: /
```

If I hit:
- `/thanos/bucketweb/blocks` works fine
- `/thanos/bucketweb/api/v1/blocks` works fine
- `/thanos/bucketweb/` redirects to `/thanos/bucketweb` and it breaks

Other combination of those parameters create a whole lot of funky behaviors.

Hitting `/thanos/bucketweb/` redirects to `/thanos/bucketweb/blocks`.

See before.

Looks like the redirect is just missing `/blocks` [here](https://github.com/thanos-io/thanos/blob/main/cmd/thanos/tools_bucket.go#L364), comparing that with [Query](https://github.com/thanos-io/thanos/blob/main/cmd/thanos/query.go#L548)

This behavior was introduced by https://github.com/thanos-io/thanos/pull/3690

onprem:

## Changes
- Server React UI index page on request to `/` instead of figuring out the path to redirect to. This fixes #4128 
- Append `/` prefix to other redirects. Without this the location is interpreted as relative and we end up doing weird redirections like `/prefix/new/prefix/...`.

## Verification
Tested Querier, Ruler, and Block Viewer UI locally with and without external prefixes

### DIFF
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -1,12 +1,12 @@
 version: 2
 updates:
-  - package_manager: "gomod"
+  - package-ecosystem: "gomod"
     directory: "/"
     vendor: false
     schedule:
       interval: "weekly"
     labels: ["dependencies"]
-  - package_manager: "docker"
+  - package-ecosystem: "docker"
     directory: "/"
     schedule:
       interval: "weekly"
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -1,0 +1,14 @@
+version: 2
+updates:
+  - package_manager: "gomod"
+    directory: "/"
+    vendor: false
+    schedule:
+      interval: "weekly"
+    labels: ["dependencies"]
+  - package_manager: "docker"
+    directory: "/"
+    schedule:
+      interval: "weekly"
+    labels: ["dependencies"]
+
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,6 +1,6 @@
 # By default we pin to amd64 sha. Use make docker to automatically adjust for arm64 versions.
-ARG SHA="fca3819d670cdaee0d785499fda202ea01c0640ca0803d26ae6dbf2a1c8c041c"
-FROM quay.io/prometheus/busybox@sha256:${SHA}
+ARG BASE_DOCKER_SHA="14d68ca3d69fceaa6224250c83d81d935c053fb13594c811038c461194599973"
+FROM quay.io/prometheus/busybox@sha256:${BASE_DOCKER_SHA}
 LABEL maintainer="The Thanos Authors"
 
 COPY /thanos_tmp_for_docker /bin/thanos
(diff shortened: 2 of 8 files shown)


onprem:
As this bug was introduced during v0.20.0, does it makes sense to release 0.20.2 with this fix included? cc @kakkoyun

kakkoyun:
Yes, I'd say let's release a minor version with the patch.

onprem:
I think I need to create a PR to the `release-0.20` branch for that, right?

onprem:
Finally managed to change the base correctly :dizzy_face:

onprem:
The e2e tests are failing because of the busybox SHA pinning issue (which is fixed on main).

bwplotka:
Right, however to release the version we need to pull those in too.

onprem:
I see two PRs related to this, do we need both of them here? I was thinking of cherry-picking those two commits from `main`.

https://github.com/thanos-io/thanos/pull/4171 and https://github.com/thanos-io/thanos/pull/4202

GiedriusS:
Are there any tests we could add? I remember how we've had lots of different fixes for the prefix stuff and the issue wasn't completely solved until tests were added that actually programmatically check whether everything works correctly :sweat:

kakkoyun:
As @GiedriusS suggested it'd be great to have tests.

About #4171 and #4202, do we need them for the end artifact or just for CI to pass? I don't have enough context on this.

onprem:
> About #4171 and #4202, do we need them for the end artifact or just for CI to pass? I don't have enough context on this.

I am sorry for the delay on this. I don't have enough context either, but I think we need them for both CI and the end artifact as we won't be able to create docker images to push. I am not even sure if both of them are required or if we can only use one of them.

But since the next release is very soon (first RC in 6 days I think), can we just merge this to main and go ahead.

onprem:
> Are there any tests we could add? I remember how we've had lots of different fixes for the prefix stuff and the issue wasn't completely solved until tests were added that actually programmatically check whether everything works correctly :sweat:

@GiedriusS  Yes, we do have tests for that, but I am not sure how can we add tests for these type of scenarios. There are a lot of places where things can go wrong :sweat_smile: 

The E2E tests for routePrefix etc would have caught this but, but they access the `/graph` page directly and this bug only occurred when you tried to access `/`.

onprem:
Depandabot CI is saying that the config file is wrong and it doesn't have anything like `vendor` in it's schema, but [the documentation](https://docs.github.com/en/code-security/supply-chain-security/keeping-your-dependencies-updated-automatically/configuration-options-for-dependency-updates#vendor) says otherwise :shrug:

kakkoyun (approved):
LGTM 🚀

Issue #4168, opened 2021-05-04
querier: Add TLS to Thanos APIs


bwplotka:
Prometheus have this from January, it was added to Alertmanager this week too. I think it's time to add it to Thanos too (:

roidelapluie:
Feel free to reuse the exporter toolkit if it suits your needs: https://github.com/prometheus/exporter-toolkit

saswatamcode:
Can I look into this?

PR #4211, opened 2021-05-10
Add TLS and basic auth

saswatamcode:

## Changes

This PR adds in TLS and basic auth to Thanos HTTP APIs using [exporter-toolkit](https://github.com/prometheus/exporter-toolkit). Configuration follows a similar pattern to Prometheus, where a file containing [TLS configuration](https://github.com/prometheus/prometheus/blob/main/docs/configuration/https.md) is provided using `http-tls-config` flag.
Addresses issue #4168 

## Verification

Tested locally.

### DIFF
--- a/cmd/thanos/config.go
+++ b/cmd/thanos/config.go
@@ -57,7 +57,7 @@ func (hc *httpConfig) registerFlag(cmd extkingpin.FlagClause) *httpConfig {
 		Default("2m").SetValue(&hc.gracePeriod)
 	cmd.Flag(
 		"http-tls-config",
-		"[EXPERIMENTAL] Path to configuration file that can enable TLS or authentication.",
+		"[EXPERIMENTAL] Path to the configuration file that can enable TLS or authentication for all HTTP endpoints.",
 	).Default("").StringVar(&hc.tlsConfig)
 	return hc
 }
--- a/cmd/thanos/config.go
+++ b/cmd/thanos/config.go
@@ -56,7 +56,7 @@ func (hc *httpConfig) registerFlag(cmd extkingpin.FlagClause) *httpConfig {
 		"Time to wait after an interrupt received for HTTP Server.").
 		Default("2m").SetValue(&hc.gracePeriod)
 	cmd.Flag(
-		"http-tls-config",
+		"http.config",
 		"[EXPERIMENTAL] Path to the configuration file that can enable TLS or authentication for all HTTP endpoints.",
 	).Default("").StringVar(&hc.tlsConfig)
 	return hc
--- a/cmd/thanos/config.go
+++ b/cmd/thanos/config.go
@@ -44,6 +44,7 @@ func (gc *grpcConfig) registerFlag(cmd extkingpin.FlagClause) *grpcConfig {
 
 type httpConfig struct {
 	bindAddress string
+	tlsConfig   string
 	gracePeriod model.Duration
 }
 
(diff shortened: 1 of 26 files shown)


bwplotka (commented):
Thanks! Some thoughts. Also we might want some docs 🙈

cmd/thanos/config.go
```diff
@@ -54,6 +55,10 @@ func (hc *httpConfig) registerFlag(cmd extkingpin.FlagClause) *httpConfig {
 	cmd.Flag("http-grace-period",
 		"Time to wait after an interrupt received for HTTP Server.").
 		Default("2m").SetValue(&hc.gracePeriod)
+	cmd.Flag(
```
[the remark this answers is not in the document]
saswatamcode: Sure. Will add that in!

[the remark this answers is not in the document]
saswatamcode: Currently, exporter-toolkit doesn't take in content(`byte[]`) of the TLS config file as an argument for its server, it only takes in the path(`string`) as highlighted [here](https://github.com/prometheus/exporter-toolkit/blob/master/web/tls_config.go#L178). So using `pathorcontent` isn't currently possible I think, since this flag fetches bytes. @bwplotka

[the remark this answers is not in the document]
bwplotka: ack

pkg/extkingpin/flags.go
```diff
@@ -43,11 +43,14 @@ func RegisterGRPCFlags(cmd FlagClause) (
...
 
 // RegisterCommonObjStoreFlags register flags commonly used to configure http servers with.
-func RegisterHTTPFlags(cmd FlagClause) (httpBindAddr *string, httpGracePeriod *model.Duration) {
+func RegisterHTTPFlags(cmd FlagClause) (httpBindAddr *string, httpGracePeriod *model.Duration, httpTLSConfig *string) {
 	httpBindAddr = cmd.Flag("http-address", "Listen host:port for HTTP endpoints.").Default("0.0.0.0:10902").String()
 	httpGracePeriod = ModelDuration(cmd.Flag("http-grace-period", "Time to wait after an interrupt received for HTTP Server.").Default("2m")) // by default it's the same as query.timeout.
-
-	return httpBindAddr, httpGracePeriod
+	httpTLSConfig = cmd.Flag(
+		"http-tls-config",
```
bwplotka, line 50: One nit, I heard from @roidelapluie that Prometheus flag is called `web.config.file` and not tls-config.file for a good reason: It's not only TLS. I think it's good idea to change it to `http.config` maybe? Also we can expand (contribute) to exporter toolking to support passing parsed configuration too if we want pathorcontent flag (up to us)

bwplotka (approved):
Amazing! LGTM, just one nit.

pkg/extkingpin/flags.go
roidelapluie: From what I understand you do not need changes in the downstream exporter toolkit to achieve what you want.

You could probably to all of this in thanos. --http-config={file}  or --http-config=/path:{file} could be supported without downstram changes.

roidelapluie: Maybe I do not understand the change. Are you trying to use basic auth on certain paths only? Or am I misunderstanding?

roidelapluie: There is a reason why we take a path. The file is read on every request to make sure that the TLS certificate and the users are up to date.

saswatamcode: Oh, okay. Then I guess using `pathorcontent` isn't suitable and the normal `Flag` method would be better. Also, no change is needed in exporter or Thanos then. Thanks a lot for explaining! 🙂

bwplotka (approved):
Epic work! LGTM 💪🏽 


cc @kakkoyun and anyone else

I will merge in few hours, if no objections

kakkoyun (approved):
👍

PR #4217, opened 2021-05-10
Style: add css vendor prefix

Aman-Codes:

## Changes

- Added [CSS Vendor Prefix](https://developer.mozilla.org/en-US/docs/Glossary/Vendor_Prefix) to ensure smooth and almost similar experience in different browsers
- Fixed spacing to ensure uniformity in all CSS file

## Verification

Tested in different browsers

### DIFF
--- a/pkg/ui/react-app/src/App.css
+++ b/pkg/ui/react-app/src/App.css
@@ -32,6 +32,8 @@ input[type='checkbox']:checked + label {
 
 .expression-input .cm-expression-input {
   border: 1px solid #ced4da;
+  -webkit-box-flex: 1;
+  -ms-flex: 1 1 auto;
   flex: 1 1 auto;
   padding: 4px 0 0 8px;
   font-size: 15px;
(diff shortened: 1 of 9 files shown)


onprem:
Hey, for the React UI we are using Create React App, which automatically does vendor-prefixing on build time.

> This project setup minifies your CSS and adds vendor prefixes to it automatically through Autoprefixer so you don’t need to worry about it.

From CRA documentation. More info: https://create-react-app.dev/docs/post-processing-css/

Also the Old UI is kind of deprecated now (there's already a PR to remove it) so I am not sure if it makes sense for us to do changes to the Old UI at this point.

This makes sense for landing page/documentation website though, but I am not sure if this is the right approach. We might be able to use something similar to [Autoprefixer](https://github.com/postcss/autoprefixer) (used in CRA) in Hugo build step to handle this automatically.

onprem:
Apparently Hugo can handle CSS post-processing using PostCSS (https://gohugo.io/hugo-pipes/postcss/) so we can use Autoprefixer.

Though if we want to deal with `postcss-cli` (and it's autoprefixer module) as a requirement for building website is a completely different matter (as it involves dealing with npm now which we wouldn't have needed before). cc @bwplotka

PR #4220, opened 2021-05-11
Pushdown labels values matchers API to store levels

yeya24:

## Changes

Fixes https://github.com/thanos-io/thanos/issues/3779.

Now the labelvalue matchers push down are implemented in the TSDB store, Prometheus store, and the bucket store, thanks to @Namanl2001 and @pstibrany. Now it is ready to replace the old series approach with the new implementation.

Also, this pr adds Prometheus release v2.26.0 to our circle CI test to ensure the compatibility and correctness of the new label values matchers for newer Prometheus versions. This is mentioned in https://github.com/thanos-io/thanos/discussions/4020#discussioncomment-570912.


onprem (approved):
LGTM!

pkg/api/query/v1.go
```diff
@@ -540,21 +540,30 @@ func (qapi *QueryAPI) labelValues(r *http.Request) (interface{}, []error, *api.A
...
 	)
-	// TODO(yeya24): push down matchers to Store level.
 	if len(matcherSets) > 0 {
-		// Get all series which match matchers.
-		var sets []storage.SeriesSet
-		for _, mset := range matcherSets {
-			s := q.Select(false, nil, mset...)
-			sets = append(sets, s)
+		var callWarnings storage.Warnings
+		labelValuesSet := make(map[string]struct{})
```
GiedriusS, line 545: My only question is why are you building a `map[string]struct{}` here instead of adding everything directly to vals:

```go
			retrievedVals, callWarnings, err := q.LabelValues(name, matchers...)
			if err != nil {
				return nil, nil, &api.ApiError{Typ: api.ErrorExec, Err: err}
			}
			warnings = append(warnings, callWarnings...)
			vals = append(vals, retrievedVals...)
```
? :thinking: This should be more efficient.

onprem: I think this is to filter out duplicates.

GiedriusS: Ahh, right! Thanks for the answer.

GiedriusS (approved):

Issue #4180, opened 2021-05-05
bucket inspect: support machine-readable output formats


matthiasr:
### Is your proposal related to a problem?

When I want to do an action on a specific subset of blocks, I need a list of those blocks.
`thanos tools bucket ls` gives me a list of *all* buckets, but does not allow filtering.
`thanos tools bucket inspect` does exactly the filtering it needs, but the output is geared towards human readability.
I need to do `awk` gymnastics to extract the block ID.

### Describe the solution you'd like

Add a `--output` / `-o` flag to `thanos tools bucket inspect`, which allows selecting output formats.
Ideas for formats:

* `table`: current format, default
* `id`: block IDs only, similar to `ls`
* `csv`: quoted comma-separated values
* `tsv`: tab-separated values, great for consumption with Unix tools
* `json`: JSON encoding (exact format tbd), great for consumption with `jq` or scripting languages

### Describe alternatives you've considered

* Currently possible:
```
thanos tools bucket inspect … | awk '$1 == "|" && $2 != "ULID" { print $2 }'
```
* add label filtering to `bucket ls` – solves my concrete issue, but is less generally useful. Making all the information machine readable allows all kinds of integrations. Harder to judge whether the filter is correct.

### Additional context

I chose the proposed flag to be in line with `kubectl`s multiple outputs.
I don't think Thanos needs the full power of that just yet but keeping the flags similar makes them easier to remember.

wiardvanrij:
Well explained and I think this would provide good value. Would you be interested on creating this feature or an initial draft for it?

matthiasr:
Unfortunately I won't have the time anytime soon, but I wanted to put the idea out there 😃

someshkoli:
Created an initial draft for this, PTAL @wiardvanrij  @matthiasr

wiardvanrij:
PR still open, relevant

PR #4228, opened 2021-05-12
bucket inspect: Adds csv, tsv printer for inspect command

someshkoli:
Signed-off-by: someshkoli <kolisomesh27@gmail.com>

## Changes
Adds csv, tsv printer methods to print block metadata table in csv and tsv format respectively.

Further work:
 - add json output (decide format)

Closes: #4180 

## Verification
unit tests needs to be added


wiardvanrij (commented):
Thanks for picking this up, awesome 👍 

This is just my review/thought;
It's actually really smart and clever, but somewhat hard to follow/read

My initial review would be; what about just passing the `*output` into the func, like we do with `*sortBy` and then do a switch compare (default to table). 
Then make a sort of "factory"?  I.e. `printer.Print(data)` where printer is the instantiated type of printer depending on the `output` type? 

LMK if this make sense :p

someshkoli:
> Thanks for picking this up, awesome 
> 
> This is just my review/thought;
> It's actually really smart and clever, but somewhat hard to follow/read
> 
> My initial review would be; what about just passing the `*output` into the func, like we do with `*sortBy` and then do a switch compare (default to table).
> Then make a sort of "factory"? I.e. `printer.Print(data)` where printer is the instantiated type of printer depending on the `output` type?
> 
> LMK if this make sense :p

Yeah it does make sense, I initially thought of making an interface for diff printers along with their configs (TSV could have diff tab lengths etc ) but then doubted if i'm over engineering it....

wiardvanrij:
> > Thanks for picking this up, awesome
> > This is just my review/thought;
> > It's actually really smart and clever, but somewhat hard to follow/read
> > My initial review would be; what about just passing the `*output` into the func, like we do with `*sortBy` and then do a switch compare (default to table).
> > Then make a sort of "factory"? I.e. `printer.Print(data)` where printer is the instantiated type of printer depending on the `output` type?
> > LMK if this make sense :p
> 
> Yeah it does make sense, I initially thought of making an interface for diff printers along with their configs (TSV could have diff tab lengths etc ) but then doubted if i'm over engineering it....

That is actually a fair point! 
I think it would help with readability though, yet you are right that this functionality is between the lines of over-engineering or not. Regardless, it looks good. If you have any doubts, don't hesitate to ask for more input on slack or something 👍

someshkoli:
> > > Thanks for picking this up, awesome
> > > This is just my review/thought;
> > > It's actually really smart and clever, but somewhat hard to follow/read
> > > My initial review would be; what about just passing the `*output` into the func, like we do with `*sortBy` and then do a switch compare (default to table).
> > > Then make a sort of "factory"? I.e. `printer.Print(data)` where printer is the instantiated type of printer depending on the `output` type?
> > > LMK if this make sense :p
> > 
> > 
> > Yeah it does make sense, I initially thought of making an interface for diff printers along with their configs (TSV could have diff tab lengths etc ) but then doubted if i'm over engineering it....
> 
> That is actually a fair point!
> I think it would help with readability though, yet you are right that this functionality is between the lines of over-engineering or not. Regardless, it looks good. If you have any doubts, don't hesitate to ask for more input on slack or something

Yes, would be easier to decide if there are some other use cases tho

cmd/thanos/tools_bucket.go
```diff
@@ -272,6 +282,9 @@ func registerBucketInspect(app extkingpin.AppClause, objStoreConfig *extflag.Pat
 		Default("FROM", "UNTIL").Enums(inspectColumns...)
 	timeout := cmd.Flag("timeout", "Timeout to download metadata from remote storage").Default("5m").Duration()
 
+	output := cmd.Flag("output", "Output format for result. Currently supports table, cvs, and, tsv.").Default("table").String()
```
matthiasr, line 285: the `sortBy` flag seems to use enums to restrict the possible values, is this something that you could do here as well?

cmd/thanos/tools_bucket.go
```diff
@@ -616,7 +636,56 @@ func registerBucketCleanup(app extkingpin.AppClause, objStoreConfig *extflag.Pat
 	})
 }
 
-func printTable(blockMetas []*metadata.Meta, selectorLabels labels.Labels, sortBy []string) error {
+type tablePrinter func(t Table) error
```
matthiasr, line 639: I'm not sure about hardcoding the output to `os.Stdout` in all the functions. What do you think about passing an `io.Writer` into any tablePrinter?

```suggestion
type tablePrinter func(w io.Writer, t Table) error
```

docs/components/tools.md
```diff
@@ -436,6 +436,8 @@ Flags:
                              Path to YAML file that contains object store
                              configuration. See format details:
                              https://thanos.io/tip/thanos/storage.md/#configuration
+      --output="table"       Output format for result. Currently supports table,
+                             cvs, and, tsv.
```
matthiasr, line 440:
```suggestion
                             csv, and tsv.
```

matthiasr: or maybe make it "technical" and leave out the and altogether?

```suggestion
                             csv, tsv.
```

matthiasr (commented):
What do the outputs look like, can you show some examples? Do we need to worry about quoting at all, or does encoding/csv do the right thing?

docs/components/tools.md
someshkoli: > or maybe make it "technical" and leave out the and altogether?

this one

cmd/thanos/tools_bucket.go
someshkoli: Yes, even i was sceptical about using stdout directly here.

cmd/thanos/tools_bucket.go
```diff
@@ -313,8 +326,15 @@ func registerBucketInspect(app extkingpin.AppClause, objStoreConfig *extflag.Pat
...
-
-		return printTable(blockMetas, selectorLabels, *sortBy)
+		if opType == TABLE {
+			return printBlockData(blockMetas, selectorLabels, *sortBy, printTable)
+		} else if opType == TSV {
+			return printBlockData(blockMetas, selectorLabels, *sortBy, printTSV)
+		} else if opType == CSV {
+			return printBlockData(blockMetas, selectorLabels, *sortBy, printCSV)
+		} else {
+			return fmt.Errorf("Invalid output type %s.", *output)
```
GiedriusS, line 336: Typically error messages start with a lower letter because they look properly then in case they get chained e.g. `doing: Creating: linking: ...` vs. `doing: creating: linking: ...`. In fact, this case would be handled for you if you'd use enums as recommended previously

someshkoli: got it thanks. will keep in mind for future contributions

someshkoli:
> What do the outputs look like, can you show some examples? Do we need to worry about quoting at all, or does encoding/csv do the right thing?

Here's the output 


And regarding improper tab space I'm not sure why its like that :thinking:

yeya24:
This looks good and I love it. Quick question, does it support outputting csv/tsv to a file?

matthiasr:
What prevents you from redirecting it to a file like `thanos bucket inspect … --output=csv > bucket.csv`?

someshkoli:
> This looks good and I love it. Quick question, does it support outputting csv/tsv to a file? @someshkoli



> What prevents you from redirecting it to a file like `thanos bucket inspect … --output=csv > bucket.csv`?

++, there isn't much work to output it to a file but might be redundant

GiedriusS:
Is this still a draft?

yeya24:
> What prevents you from redirecting it to a file like `thanos bucket inspect … --output=csv > bucket.csv`?

@matthiasr I am thinking about the logger output. They are also written to stdout and will be mixed with the csv output I guess? An output file flag might be necessary.

someshkoli:
> Is this still a draft?

Not really but tsv output is sort of inconsistent, didn't spend too much time on this, will check soon

someshkoli:
> > Is this still a draft?
> 
> Not really but tsv output is sort of inconsistent, didn't spend too much time on this, will check soon

sorry for delayed update
Not sure what this error is but while reading the output via machine code it works properly so ig we're good to go

cmd/thanos/tools_bucket.go
```diff
@@ -619,7 +642,56 @@ func registerBucketCleanup(app extkingpin.AppClause, objStoreConfig *extflag.Pat
...
+func printCSV(w io.Writer, t Table) error {
+	csv := csv.NewWriter(w)
+	err := csv.Write(t.Header)
+	if err != nil {
+		return err
+	}
+	err = csv.WriteAll(t.Lines)
+	if err != nil {
+		return err
+	}
```
idoqo, line 669: Tiny nit - maybe this is a good place to scope the errors to the if checks?
```suggestion
	if err := csv.Write(t.Header); err != nil {
		return err
	}
	if err := csv.WriteAll(t.Lines); err != nil {
		return err
	}
```

yeya24:
@someshkoli I would like to continue this if you are busy with other work so please let me know 😄

someshkoli:
> @someshkoli I would like to continue this if you are busy with other work so please let me know smile

Will wrap this up, rebasing it

yeya24:
> > @someshkoli I would like to continue this if you are busy with other work so please let me know smile
> 
> Will wrap this up, rebasing it. @yeya24 a review ?

I don't think this works. Can you fix go build?

someshkoli:
> > > @someshkoli I would like to continue this if you are busy with other work so please let me know smile
> > 
> > 
> > Will wrap this up, rebasing it. @yeya24 a review ?
> 
> I don't think this works. Can you fix go build?

Has become too old :""), ton of rebasing required. Moving this to a clean new PR

yeya24:
@someshkoli Can you please also update the changelog?

someshkoli:
> @someshkoli Can you please also update the changelog?

Done👌

yeya24 (approved):
LGTM

PR #4231, opened 2021-05-13
Implement the dual mode for Thanos Receiver

yashrsharma44:
Implementation PR for the Thanos Split Receiver proposal - https://thanos.io/tip/proposals/202012_receive_split.md/

**PR is incomplete, pushed for tracking changes and quick review**
TODO: 
- [ ] Define behaviour when one mode is enabled, and other mode's behaviour is accessed
- [ ] Tests and Docs

### DIFF
--- a/cmd/thanos/receive.go
+++ b/cmd/thanos/receive.go
@@ -6,10 +6,8 @@ package main
 import (
 	"context"
 	"io/ioutil"
-	"net"
 	"os"
 	"path"
-	"strings"
 	"time"
 
 	"github.com/go-kit/kit/log"
@@ -77,18 +75,6 @@ func registerReceive(app *extkingpin.App) {
 		// enable ingestion if endpoint is specified, or both the hashrings config are empty,
 		// otherwise run receiver in distributor mode.
 		enableIngestion := conf.endpoint != "" || (conf.hashringsFileContent == "" && conf.hashringsFilePath == "")
-
-		// If endpoint and hashrings are empty, so try to generate a local endpoint
-		// based on the hostname and the listening port.
-		if conf.endpoint == "" && (conf.hashringsFileContent == "" && conf.hashringsFilePath == "") {
-			hostname, err := os.Hostname()
-			if hostname == "" || err != nil {
-				return errors.New("--receive.local-endpoint is empty and host could not be determined.")
-			}
-			parts := strings.Split(*conf.grpcBindAddr, ":")
-			port := parts[len(parts)-1]
-			conf.endpoint = net.JoinHostPort(hostname, port)
-		}
 
 		return runReceive(
 			g,
@@ -219,7 +205,7 @@ func runReceive(
 	// initial config and mark ourselves as ready after it completed.
 
 	// reloadGRPCServer signals when - (1)TSDB is ready and the Store gRPC server can start.
-	// (2) The Hashring files have changed.
+	// (2) The Hashring files have changed if tsdb ingestion is disabled.
 	reloadGRPCServer := make(chan struct{}, 1)
 	// hashringChangedChan signals when TSDB needs to be flushed and updated due to hashring config change.
 	hashringChangedChan := make(chan struct{}, 1)
--- a/cmd/thanos/receive.go
+++ b/cmd/thanos/receive.go
@@ -214,222 +214,45 @@ func runReceive(
 
 	level.Debug(logger).Log("msg", "setting up tsdb")
 	{
-		log.With(logger, "component", "storage")
-		dbUpdatesStarted := promauto.With(reg).NewCounter(prometheus.CounterOpts{
-			Name: "thanos_receive_multi_db_updates_attempted_total",
-			Help: "Number of Multi DB attempted reloads with flush and potential upload due to hashring changes",
-		})
-		dbUpdatesCompleted := promauto.With(reg).NewCounter(prometheus.CounterOpts{
-			Name: "thanos_receive_multi_db_updates_completed_total",
-			Help: "Number of Multi DB completed reloads with flush and potential upload due to hashring changes",
-		})
-
-		level.Debug(logger).Log("msg", "removing storage lock files if any")
-		if err := dbs.RemoveLockFilesIfAny(); err != nil {
-			return errors.Wrap(err, "remove storage lock files")
+		if err := startTSDB(g, logger, reg, dbs, dbReady, uploadC, hashringChangedChan, upload, uploadDone, statusProber, bkt); err != nil {
+			return err
 		}
-
-		// TSDBs reload logic, listening on hashring changes.
-		cancel := make(chan struct{})
-		g.Add(func() error {
-			defer close(dbReady)
-			defer close(uploadC)
-
-			// Before quitting, ensure the WAL is flushed and the DBs are closed.
-			defer func() {
-				level.Info(logger).Log("msg", "shutting down storage")
-				if err := dbs.Flush(); err != nil {
-					level.Error(logger).Log("err", err, "msg", "failed to flush storage")
-				} else {
-					level.Info(logger).Log("msg", "storage is flushed successfully")
-				}
-				if err := dbs.Close(); err != nil {
-					level.Error(logger).Log("err", err, "msg", "failed to close storage")
-					return
-				}
-				level.Info(logger).Log("msg", "storage is closed")
-			}()
-
-			for {
-				select {
-				case <-cancel:
-					return nil
-				case _, ok := <-hashringChangedChan:
-					if !ok {
-						return nil
-					}
-					dbUpdatesStarted.Inc()
-					level.Info(logger).Log("msg", "updating storage")
-
-					if err := dbs.Flush(); err != nil {
-						return errors.Wrap(err, "flushing storage")
-					}
-					if err := dbs.Open(); err != nil {
-						return errors.Wrap(err, "opening storage")
-					}
-					if upload {
-						uploadC <- struct{}{}
-						<-uploadDone
-					}
-					statusProber.Ready()
-					level.Info(logger).Log("msg", "storage started, and server is ready to receive web requests")
-					dbUpdatesCompleted.Inc()
-					dbReady <- struct{}{}
-				}
-			}
-		}, func(err error) {
-			close(cancel)
-		})
 	}
 
 	level.Debug(logger).Log("msg", "setting up hashring")
 	{
-		// Note: the hashring configuration watcher
-		// is the sender and thus closes the chan.
-		// In the single-node case, which has no configuration
-		// watcher, we close the chan ourselves.
-		updates := make(chan receive.Hashring, 1)
-
-		// The Hashrings config file path is given initializing config watcher.
-		if conf.hashringsFilePath != "" {
-			cw, err := receive.NewConfigWatcher(log.With(logger, "component", "config-watcher"), reg, conf.hashringsFilePath, *conf.refreshInterval)
-			if err != nil {
-				return errors.Wrap(err, "failed to initialize config watcher")
-			}
-
-			// Check the hashring configuration on before running the watcher.
-			if err := cw.ValidateConfig(); err != nil {
-				cw.Stop()
-				close(updates)
-				return errors.Wrap(err, "failed to validate hashring configuration file")
-			}
-
-			ctx, cancel := context.WithCancel(context.Background())
-			g.Add(func() error {
-				level.Info(logger).Log("msg", "the hashring initialized with config watcher.")
-				return receive.HashringFromConfigWatcher(ctx, updates, cw)
-			}, func(error) {
-				cancel()
-			})
-		} else {
-			var ring receive.Hashring
-			// The Hashrings config file content given initialize configuration from content.
-			if len(conf.hashringsFileContent) > 0 {
-				ring, err = receive.HashringFromConfig(conf.hashringsFileContent)
-				if err != nil {
-					close(updates)
-					return errors.Wrap(err, "failed to validate hashring configuration file")
-				}
-				level.Info(logger).Log("msg", "the hashring initialized directly with the given content through the flag.")
-			} else {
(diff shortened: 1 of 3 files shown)


cmd/thanos/receive.go
```diff
@@ -131,16 +129,22 @@ func registerReceive(app *extkingpin.App) {
 			AllowOverlappingBlocks: *tsdbAllowOverlappingBlocks,
 		}
 
-		// Local is empty, so try to generate a local endpoint
-		// based on the hostname and the listening port.
+		// Make sure the both localEndpoint and (hashringsFilePath/hashringsFileContent) are not empty
+		if *localEndpoint == "" && *hashringsFilePath == "" && *hashringsFileContent == "" {
+			return errors.New("both the receive.local-endpoint and (receive.hashrings-file/receive.hashrings) are empty, either of them should be non-empty.")
```
bwplotka, line 134: Can we refer to docs?

cmd/thanos/receive.go
```diff
@@ -226,9 +231,21 @@ func runReceive(
...
 	comp component.SourceStoreAPI,
 	hashFunc metadata.HashFunc,
+	receiverMode receive.ReceiverMode,
 ) error {
+
 	logger = log.With(logger, "component", "receive")
 	level.Warn(logger).Log("msg", "setting up receive")
+
+	switch receiverMode {
+	case receive.INGESTOR_MODE:
```
bwplotka, line 241: I wonder if we can use this time group together distributing work and ingesting mode, so it's selected in one time. (:

bwplotka: let's start with separate functions to register all?

cmd/thanos/receive.go
yashrsharma44: Done

cmd/thanos/receive.go
yashrsharma44: I tried using functions to separate the logic but seems like the functions convoluted the codebase, and passing a large list of parameters was distracting to read.

I have tried collecting the similar sections into one `if` condition, let me know if this makes sense (:

cmd/thanos/receive.go
```diff
@@ -131,16 +129,22 @@ func registerReceive(app *extkingpin.App) {
 			AllowOverlappingBlocks: *tsdbAllowOverlappingBlocks,
 		}
 
-		// Local is empty, so try to generate a local endpoint
-		// based on the hostname and the listening port.
+		// Make sure the both localEndpoint and (hashringsFilePath/hashringsFileContent) are not empty
```
squat, line 132: This seems wrong to me as it makes it impossible for a node to be in "ingestor" mode. It's absolutely fine for a Receive node to have no local endpoint and no hashring configured. This is the default way the receiver runs: all data it receives is deemed to be for itself and is stored in a local TSDB.

cmd/thanos/receive.go
```diff
@@ -131,16 +129,22 @@ func registerReceive(app *extkingpin.App) {
...
+		if *localEndpoint == "" && *hashringsFilePath == "" && *hashringsFileContent == "" {
+			return errors.New("both the receive.local-endpoint and (receive.hashrings-file/receive.hashrings) are empty; see https://thanos.io/tip/components/receive.md for details.")
+		}
+
+		// Default behavior for receiver is both ingestion and distribution
+		receiverMode := receive.DUAL_MODE
```
squat, line 138: I find this abstraction a bit more confusing than it is helpful as it forces us to have lots of checks in the code like "if this but not this" or "if not that but yes this". I think it would be more helpful to have one single flag that says "i run a TSDB". That is really the only thing we care about. Even if there is no hashring file configured, the ingestor mode MUST run a gRPC server because this mode might be at the bottom of a routing tree and know nothing about the distributor nodes that are forwarding traffic to it. This means that there is no difference at all between "dual mode" and "ingestor"

yashrsharma44: Okay, so I think, we can introduce a flag, `disableTSDB`, which would instruct the receiver to **only route data**, and not ingest data. 
So this would be the following behavior of the receiver - 

* **Both Ingestion and Distribution** - Default behavior of receiver, works as it used to, no extra flags.
* **Only Ingestion** - Just don't pass hash ring file path/content, the receiver would just ingest data.
* **Only Distribution** - We can create a flag, `disableTSDB`, which would stop creating a local tsdb for ingesting data.

Thoughts?

cmd/thanos/receive.go
yashrsharma44: I agree with your thoughts, we currently try to fetch the host as our local endpoint, so this behavior has removed that functionality. 

Originally, I thought that we should explicitly pass the `localEndpoint`, but now with the [other approach](https://github.com/thanos-io/thanos/pull/4231#discussion_r632469132), we don't have to change the behavior as such, and I can totally remove this check!

cmd/thanos/receive.go
squat: I don't see a need for this flag. We can unambiguously infer everything we need to know with the presence/absence of the current flags:

1. Hashring info != "" && local endpoint == "": means only in distribution mode, i.e. don't start a TSDB
2. Everything else means start a regular receiver with TSDB, http server, and a gRPC server

This covers all of the possible cases we could want, does not require any new flags, and keeps things simpler IMO

yashrsharma44: Sure, we can configure this decision internally also, Thanks!

yashrsharma44:
Hi @squat !

I have added an e2e test scenario about creating 3 ingestor(hashrings contain only the local endpoint) and 1 distributor. Can you check if the logic for the same is implemented correctly?

I ran the tests, and cross-checked if the behaviour is correct, but would be nice to have your opinion as well 💯

cmd/thanos/receive.go
```diff
@@ -132,15 +130,18 @@ func registerReceive(app *extkingpin.App) {
...
-		if *localEndpoint == "" {
-			hostname, err := os.Hostname()
-			if hostname == "" || err != nil {
-				return errors.New("--receive.local-endpoint is empty and host could not be determined.")
-			}
-			parts := strings.Split(*grpcBindAddr, ":")
-			port := parts[len(parts)-1]
-			*localEndpoint = net.JoinHostPort(hostname, port)
-		}
+		// if *localEndpoint == "" {
```
yashrsharma44, line 133: Do you want me to uncomment this part? However, we might not be able to support the distributor mode, as the `localEndpoint` would always have a non-nil value.

kakkoyun (commented):
This is looking however it gives me an unsettling feeling. It's really hard to follow up on changes and modes.

Could we try to encapsulate logic into functions to increase readability?

I know this was an existing issue but this is a chance to improve the control flow.

cmd/thanos/receive.go
kakkoyun: We should remove this if it's not used.

yashrsharma44:
> I know this was an existing issue but this is a chance to improve the control flow.

Can you link the issue?

cmd/thanos/receive.go
yashrsharma44: In the distribution mode, do we need to start the gRPC server? 

Also, can you list out which parts are enabled in the ingestion / distribution mode
cc @squat

yashrsharma44:
> > I know this was an existing issue but this is a chance to improve the control flow.
> 
> Can you link the issue?

Tracking issue - #2248

cmd/thanos/receive.go
yashrsharma44: Sounds great!

cmd/thanos/receive.go
```diff
@@ -214,235 +214,302 @@ func runReceive(
...
+	dbReady chan struct{},
+	comp component.SourceStoreAPI,
+	dbs *receive.MultiTSDB,
+	webHandler *receive.Handler,
+	grpcLogOpts []grpc_logging.Option,
+	tagOpts []tags.Option,
+	grpcProbe *prober.GRPCProbe,
+
+) error {
+	g.Add(func() error {
```
kakkoyun, line 289: This looks it does the job. I'd prefer keeping the `run.Group` in the top-level function and extract whatever we pass to `g.Add` on a different function.

test/e2e/receive_test.go
```diff
@@ -56,6 +56,77 @@ func ErrorHandler(_ http.ResponseWriter, _ *http.Request, err error) {
 
 func TestReceive(t *testing.T) {
 	t.Parallel()
+	// t.Run("receive_distributor_ingestor_mode", func(t *testing.T) {
```
kakkoyun, line 59: Why these are commented? Should we mark this PR as WIP?

cmd/thanos/receive.go
```diff
@@ -214,235 +214,302 @@ func runReceive(
...
+			startGRPC <- struct{}{}
+		}
+		if s != nil {
+			s.Shutdown(err)
+		}
+		return nil
+	}, func(error) {})
+	// We need to be able to start and stop the gRPC server
+	// whenever the DB changes, thus it needs its own run group.
+	g.Add(func() error {
```
kakkoyun, line 328: The function name says `setupGRPCServer` but this also starts servers. It'd be good to scope of functions small as possible.

kakkoyun (commented):
This is on a good path. Thanks a lot @yashrsharma44.

I have added comments inline.

test/e2e/receive_test.go
yashrsharma44: I wanted to check if the original tests pass after refactoring. I haven't added the logic for split behaviour, so once the tests pass, I will uncomment them.

cmd/thanos/receive.go
yashrsharma44: Shall we split up into two small functions?

cmd/thanos/receive.go
yashrsharma44: Removed. This is a breaking change, so we should mention this in the changelog

cmd/thanos/receive.go
yashrsharma44: Done

test/e2e/receive_test.go
yashrsharma44: Added the tests back

cmd/thanos/receive.go
yashrsharma44: **Yup, this is a breaking change, so lets make a note for this**

cmd/thanos/receive.go
```diff
@@ -203,246 +204,364 @@ func runReceive(
...
+			},
+		)
+	}
+
+	level.Info(logger).Log("msg", "starting receiver")
+	return nil
+}
+
+// runGRPCServer starts the grpc server, once it receives a signal from the startGRPCListening channel.
+func runGRPCServer(g *run.Group,
```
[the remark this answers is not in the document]
yashrsharma44: @muse-dev ignore

cmd/thanos/receive.go
yashrsharma44: For some reason, splitting up tasks are creating a nil pointer for the grpcServer, so I reverted back the change. Let me know if you have any suggestion

vovanloc0798 (approved):

yashrsharma44:
I might set up this branch and play with the benchmarks using this PR - https://github.com/thanos-io/thanosbench/pull/34

So will report the statistics and benchmarks, once I run them up 😎

cmd/thanos/receive.go
```diff
@@ -74,17 +72,9 @@ func registerReceive(app *extkingpin.App) {
...
-		if conf.endpoint == "" {
-			hostname, err := os.Hostname()
-			if hostname == "" || err != nil {
-				return errors.New("--receive.local-endpoint is empty and host could not be determined.")
-			}
-			parts := strings.Split(*conf.grpcBindAddr, ":")
-			port := parts[len(parts)-1]
-			conf.endpoint = net.JoinHostPort(hostname, port)
-		}
+		// enable ingestion if endpoint is specified, or both the hashrings config are empty,
```
squat, line 75: Just a nit: let's break up the sentences to make sure that the logic in the comment is crystal clear:
```suggestion
		// Enable ingestion if endpoint is specified or if both the hashrings configs are empty.
		// Otherwise, run the receiver exclusively as a distributor.
```

cmd/thanos/receive.go
```diff
@@ -203,246 +204,339 @@ func runReceive(
...
+func setupHashring(g *run.Group,
+	logger log.Logger,
+	reg *prometheus.Registry,
+	conf *receiveConfig,
+	hashringChangedChan chan struct{},
+	webHandler *receive.Handler,
+	statusProber prober.Probe,
+	reloadGRPCServer chan struct{},
+	enableIngestion bool,
+
```
squat, line 362: nit
```suggestion
```

test/e2e/e2ethanos/services.go
```diff
@@ -242,6 +243,50 @@ func NewReceiver(sharedDir string, networkName string, name string, replicationF
 	return receiver, nil
 }
 
+func NewReceiverWithDistributorMode(sharedDir string, networkName string, name string, replicationFactor int, hashring ...receive.HashringConfig) (*Service, error) {
```
squat, line 246: Naming nit: there is no concrete thing as "distribution" mode, but we can talk concretely about whether the receiver supports ingesting to a TSDB or not. I would propose calling this `NewReceiverWithoutTSDB`

squat: But not a hard blocker here

squat (dismissed):
Overall looks great to me! I think the PR is a lot simpler now without the extra concepts of the different "modes" :))

test/e2e/e2ethanos/services.go
yashrsharma44: Lets add this to the suggestions, should be consistent with the nomenclature :)

yashrsharma44: Done

kakkoyun (approved):
:1st_place_medal: :rocket:

PR #4237, opened 2021-05-14
UI: fix error in graph panel component

Aman-Codes:

## Changes

- Fixed error in graph panel component. `ul` tag should have only `li` tag as child.

### Before:
image

### After:
image

### DIFF
--- a/pkg/ui/react-app/src/pages/graph/Panel.tsx
+++ b/pkg/ui/react-app/src/pages/graph/Panel.tsx
@@ -360,25 +360,27 @@ class Panel extends Component<PanelProps & PathPrefixProps, PanelState> {
         )}
         <Row>
           <Col>
-            <Nav tabs>
-              <NavItem>
-                <NavLink
-                  className={options.type === 'table' ? 'active' : ''}
-                  onClick={() => this.handleChangeType(PanelType.Table)}
-                >
-                  Table
-                </NavLink>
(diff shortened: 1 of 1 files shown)


ianbillett:
Hey @Aman-Codes - thanks for this contribution :pray: To help us move this along, if this PR fixes an outstanding issue please can you link it? With UI changes it is helpful to provide actual screenshots of the UI rather than just the DOM, that way reviewers can see what the effect will be for end users :+1:

Aman-Codes:
Hi @ianbillett  thanks for the review

Currently there is no open issue for this but if you want then I can create one for this.
In this fix there is no change on the UI which the end user could directly observe but it will improve the screen reader experience. 
HTML syntax errors are not reported in the console but could be check through Lighthouse Report.

**Steps to test this PR:** 
- Go to the Thanos query page
- Run any query and select the graph tag.
- Run the Lighthouse Report from the developer tools of the browser
- In the accessibility section this error is reported
image

Though we have **Eslint** and **Prettier** checks which could report html syntax error 
but **this is a case where the code looks like valid JSX syntax but the generated html from it is not valid.**

ianbillett:
Ok cool - this sounds like a sensible change to make to improve accessibility.

@onprem I think it makes sense to merge this PR, but could you confirm?

ianbillett:
still relevant

PR #4239, opened 2021-05-16
compactor: Add offline deduplication

yeya24:
Signed-off-by: yeya24 <yb532204897@gmail.com>

Fixes #1014 

* [x] I added CHANGELOG entry for this change.
* [ ] Change is not relevant to the end user.

## Changes

Another attempt on offline deduplication using the existing querier's penalty deduplication algorithm.

~~This pr is WIP as I need to do more testing and is currently blocked by https://github.com/prometheus/prometheus/pull/8836, so I use my own fork here.~~

But as mentioned in https://github.com/thanos-io/thanos/issues/1014#issuecomment-746062533, this deduplication algorithm works for 99% of cases but is wrong in 1% case. So it has its own risk.

## Verification

Tested with 2 Prometheus replicas running in k8s. I replicate the minio bucket to a local directory as a filesystem objstore for testing. The local directory is used for running dedup compactor. There are 12h test data and I run 1 thanos store and 1 thanos querier for each bucket.

Undeduped bucket:

Query deduplication has to be enabled.

Screenshot from 2021-05-17 11-09-16

Bucket size: 41m (36m if normal compaction is enabled)

Offline deduped bucket:

Query deduplication disabled.

Screenshot from 2021-05-17 11-10-36

Bucket size: 19m

PR #4242, opened 2021-05-17
Move dedup iterators to dedup package

yeya24:

## Changes

As mentioned in https://github.com/thanos-io/thanos/pull/4239, this pr moves the dedup iterators to its own `dedup` package for future reusability.

Split from https://github.com/thanos-io/thanos/pull/4239.

## Verification

Tests should pass.


pkg/query/querier.go
```diff
@@ -307,7 +308,7 @@ func (q *querier) selectFn(ctx context.Context, hints *storage.SelectHints, ms .
 
 	// The merged series set assembles all potentially-overlapping time ranges of the same series into a single one.
 	// TODO(bwplotka): We could potentially dedup on chunk level, use chunk iterator for that when available.
-	return newDedupSeriesSet(set, q.replicaLabels, len(aggrs) == 1 && aggrs[0] == storepb.Aggr_COUNTER), nil
+	return dedup.NewDedupSeriesSet(set, q.replicaLabels, len(aggrs) == 1 && aggrs[0] == storepb.Aggr_COUNTER), nil
```
kakkoyun, line 311: To avoid stutter, let's rename the method. WDYT?
```suggestion
	return dedup.NewSeriesSet(set, q.replicaLabels, len(aggrs) == 1 && aggrs[0] == storepb.Aggr_COUNTER), nil
```

kakkoyun (approved):
LGTM 🥇 
Only one nit about package and factory method naming.

pkg/query/querier.go
yeya24: SGTM

kakkoyun: 👍  It's ok to keep unexported methods/APIs.

yeya24: Thank you. I have updated this pr. @kakkoyun

PR #4243, opened 2021-05-18
UI: fix loading spinner component

Aman-Codes:

## Changes

Currently, if any request fails the loading spinner still appears, which gives a false notion that the request is still in the processing state. So I have made the loading spinner hidden if the request fails.

## Screenshots
**Before:**
SpinnerError

**After:**
SpinnerFixed

## Verification

Start the React frontend without starting the backend of Thanos
```js
yarn start
```


Namanl2001:
Nice catch! but I think we should not give the option to query if there are internal server errors.

kakkoyun (approved):
LGTM 👍 Thanks for handling this.

GiedriusS (approved):
Thank you! Tested, works well. I think it is much better to have something working than nothing working at all :smile:

PR #4247, opened 2021-05-18
Update prometheus dependency to commit #8836

yeya24:

## Changes

This pr updates Prometheus dependency to https://github.com/prometheus/prometheus/commit/0a8912433a457b54a871df75e72afc3e45a31d5c.

Split from pr https://github.com/thanos-io/thanos/pull/4239.

This update included commit https://github.com/prometheus/prometheus/pull/8723, which breaks the existing TSDB store test. The writability and readability are not deterministic so I removed most of the tests. Let me know if there is a better way to solve this.


yeya24:
This testcase https://github.com/thanos-io/thanos/blob/main/pkg/store/bucket_test.go#L1938 is broken due to the recent 120 chunks split change in Prometheus.

Any idea about how to fix or change it? @metalmatze @bwplotka @onprem

bwplotka:
Hah, well looks like we relied on this ;p

> 
// This method relies on a bug in TSDB Compactor which will just merge overlapping chunks into one big chunk.// If compactor is fixed in the future, we may need a different way of generating the block, or commit// existing block to the repository.

I think, maybe it's time to remove all tests relying on such large chunk situation, since. we hopefully don't have such?

pkg/store/tsdb_test.go
```diff
@@ -684,49 +649,6 @@ func TestTSDBStore_SeriesAccessWithoutDelegateClosing(t *testing.T) {
 					_ = string(c.Raw.Data) // Access bytes by converting them to different type.
 				}))
 			}
-			testutil.NotOk(t, testutil.FaultOrPanicToErr(func() {
```
bwplotka, line 687: Why those are removed?

pkg/store/tsdb_test.go
```diff
@@ -595,26 +580,6 @@ func TestTSDBStore_SeriesAccessWithDelegateClosing(t *testing.T) {
 		// Let's close pending querier!
 		testutil.Equals(t, 1, len(csrv.closers))
 		testutil.Ok(t, csrv.closers[0].Close())
-
-		// Expect flush and close to be unblocked and without errors.
```
bwplotka, line 599: Why those are removed?

bwplotka (commented):
LGTM, just I am worried about other removed tests around Flush.. why?

pkg/store/tsdb_test.go
bwplotka: hmmm prometheus/prometheus#8723 ... Can we skip them instead 🤔

yeya24: Yes, that's a better option.

yeya24:
@bwplotka PR updated, let me know if I need to update anything else.

kakkoyun (approved):
LGTM 🎉 🚀

yeya24:
Issue opened #4266. Will merge it.

PR #4250, opened 2021-05-19
Implement stateless mode for Thanos Ruler

idoqo:
This implements the stateless mode for ruler described in the [Scalable Rule Storage proposal](https://github.com/thanos-io/thanos/blob/main/docs/proposals/202005_scalable-rule-storage.md#proposal)

### DIFF
--- a/cmd/thanos/config.go
+++ b/cmd/thanos/config.go
@@ -226,13 +226,13 @@ func (ac *alertMgrConfig) registerFlag(cmd extflag.FlagClause) *alertMgrConfig {
 }
 
 type ruleRWConfig struct {
-	remoteWrite       bool
-	remoteWriteConfig *extflag.PathOrContent
+	remoteWrite bool
+	configPath  *extflag.PathOrContent
 }
 
 func (rc *ruleRWConfig) registerFlag(cmd extflag.FlagClause) *ruleRWConfig {
 	cmd.Flag("remote-write", "If true, directs ruler to remote-write evaluated samples to the server configured by 'remote-write.config'.").
 		BoolVar(&rc.remoteWrite)
-	rc.remoteWriteConfig = extflag.RegisterPathOrContent(cmd, "remote-write.config", "YAML config for the remote-write server where samples should be sent to. See https://thanos.io/tip/components/rule.md/#query-api", false)
+	rc.configPath = extflag.RegisterPathOrContent(cmd, "remote-write.config", "YAML config for the remote-write server where samples should be sent to. See https://thanos.io/tip/components/rule.md/#query-api", false)
 	return rc
 }
--- a/cmd/thanos/config.go
+++ b/cmd/thanos/config.go
@@ -226,13 +226,10 @@ func (ac *alertMgrConfig) registerFlag(cmd extflag.FlagClause) *alertMgrConfig {
 }
 
 type ruleRWConfig struct {
-	remoteWrite bool
-	configPath  *extflag.PathOrContent
+	configPath *extflag.PathOrContent
 }
 
 func (rc *ruleRWConfig) registerFlag(cmd extflag.FlagClause) *ruleRWConfig {
-	cmd.Flag("remote-write", "If true, directs ruler to remote-write evaluated samples to the server configured by 'remote-write.config'.").
-		BoolVar(&rc.remoteWrite)
-	rc.configPath = extflag.RegisterPathOrContent(cmd, "remote-write.config", "YAML config for the remote-write server where samples should be sent to. See https://thanos.io/tip/components/rule.md/#query-api", false)
+	rc.configPath = extflag.RegisterPathOrContent(cmd, "remote-write.config", "YAML config for the remote-write server where samples should be sent to. This automatically enables stateless mode for ruler and no series will be stored in the ruler's TSDB. See https://thanos.io/tip/components/rule.md/#query-api", false)
 	return rc
 }
--- a/cmd/thanos/config.go
+++ b/cmd/thanos/config.go
@@ -224,3 +224,15 @@ func (ac *alertMgrConfig) registerFlag(cmd extflag.FlagClause) *alertMgrConfig {
 	ac.alertRelabelConfigPath = extflag.RegisterPathOrContent(cmd, "alert.relabel-config", "YAML file that contains alert relabelling configuration.", extflag.WithEnvSubstitution())
 	return ac
 }
+
+type ruleRWConfig struct {
+	remoteWrite       bool
+	remoteWriteConfig *extflag.PathOrContent
+}
+
+func (rc *ruleRWConfig) registerFlag(cmd extflag.FlagClause) *ruleRWConfig {
+	cmd.Flag("remote-write", "If true, directs ruler to remote-write evaluated samples to the server configured by 'remote-write.config'.").
+		BoolVar(&rc.remoteWrite)
+	rc.remoteWriteConfig = extflag.RegisterPathOrContent(cmd, "remote-write.config", "YAML config for the remote-write server where samples should be sent to. See https://thanos.io/tip/components/rule.md/#query-api", false)
+	return rc
+}
--- a/cmd/thanos/config.go
+++ b/cmd/thanos/config.go
@@ -224,12 +224,3 @@ func (ac *alertMgrConfig) registerFlag(cmd extflag.FlagClause) *alertMgrConfig {
 	ac.alertRelabelConfigPath = extflag.RegisterPathOrContent(cmd, "alert.relabel-config", "YAML file that contains alert relabelling configuration.", extflag.WithEnvSubstitution())
 	return ac
 }
-
-type ruleRWConfig struct {
-	configPath *extflag.PathOrContent
-}
-
-func (rc *ruleRWConfig) registerFlag(cmd extflag.FlagClause) *ruleRWConfig {
-	rc.configPath = extflag.RegisterPathOrContent(cmd, "remote-write.config", "YAML config for the remote-write server where samples should be sent to. This automatically enables stateless mode for ruler and no series will be stored in the ruler's TSDB. See https://thanos.io/tip/components/rule.md/#query-api", false)
-	return rc
-}
--- a/cmd/thanos/rule.go
+++ b/cmd/thanos/rule.go
@@ -15,9 +15,6 @@ import (
 	"strings"
 	"time"
 
-	"github.com/prometheus/prometheus/storage"
-	"github.com/thanos-io/thanos/pkg/rules/remotewrite"
-
 	"github.com/go-kit/kit/log"
 	"github.com/go-kit/kit/log/level"
 	grpc_logging "github.com/grpc-ecosystem/go-grpc-middleware/v2/interceptors/logging"
@@ -33,6 +30,7 @@ import (
 	"github.com/prometheus/prometheus/pkg/relabel"
 	"github.com/prometheus/prometheus/promql"
 	"github.com/prometheus/prometheus/rules"
+	"github.com/prometheus/prometheus/storage"
 	"github.com/prometheus/prometheus/tsdb"
 	"github.com/prometheus/prometheus/util/strutil"
 	"github.com/thanos-io/thanos/pkg/errutil"
@@ -53,6 +51,7 @@ import (
 	"github.com/thanos-io/thanos/pkg/promclient"
 	"github.com/thanos-io/thanos/pkg/query"
 	thanosrules "github.com/thanos-io/thanos/pkg/rules"
+	"github.com/thanos-io/thanos/pkg/rules/remotewrite"
 	"github.com/thanos-io/thanos/pkg/runutil"
 	grpcserver "github.com/thanos-io/thanos/pkg/server/grpc"
 	httpserver "github.com/thanos-io/thanos/pkg/server/http"
--- a/cmd/thanos/rule.go
+++ b/cmd/thanos/rule.go
@@ -78,8 +78,7 @@ type ruleConfig struct {
 	alertQueryURL          *url.URL
 	alertRelabelConfigYAML []byte
 
-	rwConfig     ruleRWConfig
-	rwConfigYAML []byte
+	rwConfig *extflag.PathOrContent
 
 	resendDelay    time.Duration
 	evalInterval   time.Duration
@@ -96,7 +95,6 @@ func (rc *ruleConfig) registerFlag(cmd extkingpin.FlagClause) {
 	rc.shipper.registerFlag(cmd)
 	rc.query.registerFlag(cmd)
 	rc.alertmgr.registerFlag(cmd)
-	rc.rwConfig.registerFlag(cmd)
 }
 
 // registerRule registers a rule command.
@@ -123,6 +121,8 @@ func registerRule(app *extkingpin.App) {
 		Default("1m").DurationVar(&conf.resendDelay)
 	cmd.Flag("eval-interval", "The default evaluation interval to use.").
 		Default("30s").DurationVar(&conf.evalInterval)
+
+	conf.rwConfig = extflag.RegisterPathOrContent(cmd, "remote-write.config", "YAML config for the remote-write server where samples should be sent to. This automatically enables stateless mode for ruler and no series will be stored in the ruler's TSDB. If an empty config (or file) is provided, the flag is ignored and ruler is run with its own TSDB.", extflag.WithEnvSubstitution())
 
 	reqLogDecision := cmd.Flag("log.request.decision", "Deprecation Warning - This flag would be soon deprecated, and replaced with `request.logging-config`. Request Logging for logging the start and end of requests. By default this flag is disabled. LogFinishCall: Logs the finish call of the requests. LogStartAndFinishCall: Logs the start and finish call of the requests. NoLogCall: Disable request logging.").Default("").Enum("NoLogCall", "LogFinishCall", "LogStartAndFinishCall", "")
 
@@ -169,14 +169,6 @@ func registerRule(app *extkingpin.App) {
 		}
 		if (len(conf.query.sdFiles) != 0 || len(conf.query.addrs) != 0) && len(conf.queryConfigYAML) != 0 {
 			return errors.New("--query/--query.sd-files and --query.config* parameters cannot be defined at the same time")
-		}
-
-		// Parse and check remote-write config and enable stateless mode for ruler.
-		if conf.rwConfig.configPath != nil {
-			conf.rwConfigYAML, err = conf.rwConfig.configPath.Content()
-			if err != nil {
-				return err
-			}
 		}
 
 		// Parse and check alerting configuration.
@@ -340,16 +332,14 @@ func runRule(
 		db         *tsdb.DB
 	)
 
-	if conf.rwConfig.configPath != nil {
-		conf.rwConfigYAML, err = conf.rwConfig.configPath.Content()
-		if err != nil {
-			return err
-		}
+	rwCfgYAML, err := conf.rwConfig.Content()
+	if err != nil {
+		return err
+	}
+
+	if len(rwCfgYAML) > 0 {
 		var rwCfg remotewrite.Config
-		if len(conf.rwConfigYAML) == 0 {
-			return errors.New("no --remote-write.config was given")
-		}
-		rwCfg, err = remotewrite.LoadRemoteWriteConfig(conf.rwConfigYAML)
+		rwCfg, err = remotewrite.LoadRemoteWriteConfig(rwCfgYAML)
 		if err != nil {
 			return err
 		}
--- a/cmd/thanos/rule.go
+++ b/cmd/thanos/rule.go
@@ -171,8 +171,8 @@ func registerRule(app *extkingpin.App) {
 			return errors.New("--query/--query.sd-files and --query.config* parameters cannot be defined at the same time")
 		}
 
-		// Parse and check remote-write config if it's enabled
-		if conf.rwConfig.remoteWrite {
+		// Parse and check remote-write config and enable stateless mode for ruler.
+		if conf.rwConfig.configPath != nil {
 			conf.rwConfigYAML, err = conf.rwConfig.configPath.Content()
 			if err != nil {
 				return err
@@ -339,10 +339,8 @@ func runRule(
 		queryable  storage.Queryable
 		db         *tsdb.DB
 	)
-	if err != nil {
-		return errors.Wrap(err, "open TSDB")
-	}
-	if conf.rwConfig.remoteWrite {
+
+	if conf.rwConfig.configPath != nil {
 		conf.rwConfigYAML, err = conf.rwConfig.configPath.Content()
 		if err != nil {
 			return err
--- a/cmd/thanos/rule.go
+++ b/cmd/thanos/rule.go
@@ -15,6 +15,9 @@ import (
 	"strings"
 	"time"
 
+	"github.com/prometheus/prometheus/storage"
+	"github.com/thanos-io/thanos/pkg/rules/remotewrite"
+
 	"github.com/go-kit/kit/log"
 	"github.com/go-kit/kit/log/level"
 	grpc_logging "github.com/grpc-ecosystem/go-grpc-middleware/v2/interceptors/logging"
@@ -75,6 +78,9 @@ type ruleConfig struct {
 	alertQueryURL          *url.URL
 	alertRelabelConfigYAML []byte
 
+	rwConfig     ruleRWConfig
+	rwConfigYAML []byte
+
 	resendDelay    time.Duration
 	evalInterval   time.Duration
 	ruleFiles      []string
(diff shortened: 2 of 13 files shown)


test/e2e/rule_test.go
```diff
@@ -556,6 +560,141 @@ func TestRule(t *testing.T) {
...
 }
 
+func TestRule_CanRemoteWriteData(t *testing.T) {
+	t.Parallel()
+
+	testAlertRuleRecordAbsentMetric := `
+groups:
+- name: example_record_rules
+  interval: 100ms
+  rules:
```
bwplotka, line 570: Let's also add Alert so we have both types (:  and this is when variable name makes sense ;p

test/e2e/rule_test.go
```diff
@@ -556,6 +560,141 @@ func TestRule(t *testing.T) {
...
+			Name: "thanos-receiver",
+		},
+	})
+	testutil.Ok(t, err)
+	testutil.Ok(t, s.StartAndWaitReady(r))
+
+	writeTargets(t, filepath.Join(s.SharedDir(), queryTargetsSubDir, "targets.yaml"), querier.NetworkHTTPEndpoint())
+	writeTargets(t, filepath.Join(s.SharedDir(), amTargetsSubDir, "targets.yaml"), am.NetworkHTTPEndpoint())
+
+	t.Run("inject samples into receiver to reset its StoreAPI MinTime", func(t *testing.T) {
```
bwplotka, line 652: This works great, and nice way of generating data. Some thoughts:

* There is no need into creating Prometheus with all extra functionalities. We know that CI GH actions have e limited CPU  time, so we don't want to run extra container if not needed. To generate data, we can quickly craft small remote write request using literally what @metalmatze used in his PR. Just make sure you use the local endpoint, not network endpoint.
* We don't want to reset some min time, we literally want generate, data so can we rename this subtest accordingly (and comment)

OR alternatively we can remove all of this, because our expression does not use any Store API data right now (unless you want to change that). Might simplify our test!

test/e2e/rule_test.go
```diff
@@ -556,6 +560,141 @@ func TestRule(t *testing.T) {
...
+				"job":"myself",
+				"prometheus": "prom",
+				"receive": "1",
+				"replica": "0",
+				"tenant_id": "default-tenant",
+			},
+		})
+	})
+
+	t.Run("query can contact from receiver", func(t *testing.T) {
```
bwplotka, line 672: I think with this query you can always do that. There will debug log messages "No storeAPI match", but that's fine

test/e2e/rule_test.go
```diff
@@ -556,6 +560,141 @@ func TestRule(t *testing.T) {
...
+		queryAndAssertSeries(t, ctx, querier.HTTPEndpoint(), testAbsentQuery, promclient.QueryOptions{
+			Deduplicate: false,
+		}, []model.Metric{
+			{
+				"job": "thanos-receive",
+			},
+		})
+	})
+
+	t.Run("can fetch remote-written samples from receiver", func(t *testing.T) {
```
bwplotka, line 683: I think this should work and let's just without data generation wait for that! (:

test/e2e/rule_test.go
```diff
@@ -556,6 +560,141 @@ func TestRule(t *testing.T) {
...
+			Deduplicate: false,
+		}, []model.Metric{
+			{
+				"__name__": "test_absent_metric",
+				"job":"thanos-receive",
+				"receive": "1",
+				"tenant_id": "default-tenant",
+			},
+		})
+	})
```
bwplotka, line 695: Since we have Alertmanager, maybe we want to create alert and check if the alert was triggered? (But not sure if we want to test that if we already tested this for non stateless ruler)

cmd/thanos/config.go
```diff
@@ -217,3 +217,15 @@ func (ac *alertMgrConfig) registerFlag(cmd extflag.FlagClause) *alertMgrConfig {
...
 	return ac
 }
+
+type ruleRWConfig struct {
+	remoteWrite bool
+	configPath  *extflag.PathOrContent
+}
+
+func (rc *ruleRWConfig) registerFlag(cmd extflag.FlagClause) *ruleRWConfig {
+	cmd.Flag("remote-write", "If true, directs ruler to remote-write evaluated samples to the server configured by 'remote-write.config'.").
```
bwplotka, line 227: Why not check if config is provided and only then enable remote write? ;p

bwplotka: Important! User will expect remote write can be added to the TSDB ruler also.. We need to make sure that is not the case, and if you provide remote write it will make ruler stateless (except the WAL buffer)

cmd/thanos/rule.go
```diff
@@ -163,6 +170,14 @@ func registerRule(app *extkingpin.App) {
 			return errors.New("--query/--query.sd-files and --query.config* parameters cannot be defined at the same time")
 		}
 
+		// Parse and check remote-write config if it's enabled
```
bwplotka, line 173:
```suggestion
		// Parse and check remote-write config if it's enabled.
```

pkg/rules/remotewrite/remotewrite.go
```diff
@@ -0,0 +1,43 @@
...
+		return nil, err
+	}
+	remoteStore := remote.NewStorage(logger, reg, walStore.StartTime, walStore.Directory(), 1*time.Minute, nil)
+	err = remoteStore.ApplyConfig(&config.Config{
+		GlobalConfig:       config.DefaultGlobalConfig,
+		RemoteWriteConfigs: []*config.RemoteWriteConfig{rwConfig.RemoteStore},
+	})
+	if err != nil {
+		return nil, fmt.Errorf("failed applying config to remote storage: %w", err)
+	}
```
bwplotka, line 40:
```suggestion
	if err = remoteStore.ApplyConfig(&config.Config{
		GlobalConfig:       config.DefaultGlobalConfig,
		RemoteWriteConfigs: []*config.RemoteWriteConfig{rwConfig.RemoteStore},
	}); err != nil {
		return nil, errors.Wrapf(err, "applying config to remote storage")
	}
```

pkg/rules/remotewrite/remotewrite.go
```diff
@@ -0,0 +1,43 @@
...
+	fanoutStorage := storage.NewFanout(logger, walStore, remoteStore)
```
bwplotka, line 41: https://thanos.io/tip/contributing/coding-style-guide.md/#avoid-defining-variables-used-only-once

pkg/rules/remotewrite/series.go
```diff
@@ -0,0 +1,256 @@
+package remotewrite
+
```
bwplotka, line 2: Let's comment this file was copied! 

Also Let's add TODO to use Prometheus package when avaialble https://github.com/prometheus/prometheus/pull/8785

test/e2e/rule_test.go
```diff
@@ -9,19 +9,19 @@ import (
...
-	"net/url"
 	"os"
 	"path/filepath"
 	"sync"
 	"testing"
 	"time"
 
+	commoncfg "github.com/prometheus/common/config"
+	"github.com/prometheus/prometheus/config"
+	"github.com/thanos-io/thanos/pkg/rules/remotewrite"
```
squat, line 23: group with other thanos imports

scripts/quickstart.sh
```diff
@@ -245,6 +245,19 @@ QUERIER_JAEGER_CONFIG=$(
...
+REMOTE_WRITE_FLAGS=""
+if [ -n "${STATELESS_RULER_ENABLED}" ]; then
+  cat >/data/rule-remote-write.yaml <<-EOF
+  name: "thanos-receivers"
+  remote_write:
+    url: "http://127.0.0.1:10908/api/v1/receive"
+    name: "receive-0"
+EOF
+
+  REMOTE_WRITE_FLAGS="--remote-write --remote-write.config-file data/rule-remote-write.yaml
```
idoqo, line 257: `--remote-write` flag has been removed and should be reflected here.

cmd/thanos/rule.go
```diff
@@ -75,6 +78,9 @@ type ruleConfig struct {
 	alertQueryURL          *url.URL
 	alertRelabelConfigYAML []byte
 
+	rwConfig     ruleRWConfig
+	rwConfigYAML []byte
```
bwplotka, line 82: Do we need this?

cmd/thanos/rule.go
```diff
@@ -75,6 +78,9 @@ type ruleConfig struct {
 	alertQueryURL          *url.URL
 	alertRelabelConfigYAML []byte
 
+	rwConfig     ruleRWConfig
```
bwplotka, line 81:
```suggestion
	rwConfig  *extflag.PathOrContent
```

cmd/thanos/rule.go
```diff
@@ -165,6 +172,14 @@ func registerRule(app *extkingpin.App) {
 			return errors.New("--query/--query.sd-files and --query.config* parameters cannot be defined at the same time")
 		}
 
+		// Parse and check remote-write config and enable stateless mode for ruler.
+		if conf.rwConfig.configPath != nil {
```
bwplotka, line 176: What if the user passes empty configuration?

cmd/thanos/rule.go
```diff
@@ -165,6 +172,14 @@ func registerRule(app *extkingpin.App) {
...
+			conf.rwConfigYAML, err = conf.rwConfig.configPath.Content()
```
bwplotka, line 177: Let's operate on pathOrContent in stead of extra field

cmd/thanos/rule.go
bwplotka, line 176: I think we can remove this

cmd/thanos/rule.go
```diff
@@ -316,25 +331,54 @@ func runRule(
...
+		queryable  storage.Queryable
+		db         *tsdb.DB
+	)
 
-	db, err := tsdb.Open(conf.dataDir, log.With(logger, "component", "tsdb"), reg, tsdbOpts)
-	if err != nil {
-		return errors.Wrap(err, "open TSDB")
-	}
+	if conf.rwConfig.configPath != nil {
+		conf.rwConfigYAML, err = conf.rwConfig.configPath.Content()
```
bwplotka, line 341: Let's just use local var

cmd/thanos/rule.go
```diff
@@ -316,25 +331,54 @@ func runRule(
...
+		if err != nil {
+			return err
+		}
+		var rwCfg remotewrite.Config
+		if len(conf.rwConfigYAML) == 0 {
+			return errors.New("no --remote-write.config was given")
```
bwplotka, line 347:
```suggestion
			return errors.New("empty --remote-write.config or empty file via --remote-write.config-file was given")
```

pkg/rules/remotewrite/remotewrite.go
```diff
@@ -0,0 +1,41 @@
...
+
+	"github.com/go-kit/kit/log"
+	"github.com/prometheus/client_golang/prometheus"
+	"github.com/prometheus/prometheus/config"
+	"github.com/prometheus/prometheus/storage"
+	"github.com/prometheus/prometheus/storage/remote"
+	"gopkg.in/yaml.v2"
+)
+
+type Config struct {
```
bwplotka, line 15: Can we add public comments

pkg/rules/remotewrite/remotewrite.go
```diff
@@ -0,0 +1,41 @@
...
+	walStore, err := NewStorage(logger, reg, walDir)
+	if err != nil {
+		return nil, err
+	}
+	remoteStore := remote.NewStorage(logger, reg, walStore.StartTime, walStore.Directory(), 1*time.Minute, nil)
+	if err := remoteStore.ApplyConfig(&config.Config{
+		GlobalConfig:       config.DefaultGlobalConfig,
+		RemoteWriteConfigs: []*config.RemoteWriteConfig{rwConfig.RemoteStore},
+	}); err != nil {
+		return nil, fmt.Errorf("failed applying config to remote storage: %w", err)
```
bwplotka, line 38:
```suggestion
		return nil, errros.Wrapf(err, "applying config to remote storage")
```

pkg/rules/remotewrite/wal.go
```diff
@@ -0,0 +1,702 @@
...
+	numDeletedSeries     prometheus.Gauge
+	totalCreatedSeries   prometheus.Counter
+	totalRemovedSeries   prometheus.Counter
+	totalAppendedSamples prometheus.Counter
+}
+
+func newStorageMetrics(r prometheus.Registerer) *storageMetrics {
+	m := storageMetrics{r: r}
+	m.numActiveSeries = prometheus.NewGauge(prometheus.GaugeOpts{
+		Name: "agent_wal_storage_active_series",
```
bwplotka, line 43:
```suggestion
		Name: "wal_storage_active_series",
```

Let's start with `wal...`? Let's remove `agent`

test/e2e/rule_test.go
```diff
@@ -321,7 +325,7 @@ func TestRule_AlertmanagerHTTPClient(t *testing.T) {
 }
 
 func TestRule(t *testing.T) {
-	t.Skip("Flaky test. Fix it. See: https://github.com/thanos-io/thanos/issues/3425.")
+	//t.Skip("Flaky test. Fix it. See: https://github.com/thanos-io/thanos/issues/3425.")
```
bwplotka, line 328: Let's uncomment this?

bwplotka (commented):
Looks good, just minor nits, tests, fixing build errors and metrics (:

pkg/rules/remotewrite/remotewrite.go
metalmatze: Is Thanos still prefering `errors.Wrapf` over `fmt.Errorf` with `%w`?

cmd/thanos/rule.go
```diff
@@ -316,25 +323,52 @@ func runRule(
...
 	if err != nil {
-		return errors.Wrap(err, "open TSDB")
+		return err
 	}
 
-	level.Debug(logger).Log("msg", "removing storage lock file if any")
-	if err := removeLockfileIfAny(logger, conf.dataDir); err != nil {
-		return errors.Wrap(err, "remove storage lock files")
-	}
+	if len(rwCfgYAML) > 0 {
```
idoqo, line 337: @squat @bwplotka 

In this case though, an empty `--remote-write.config` (or `--remote-write.config-file`)will fall back to using the TSDB-backed ruler. This is because `len(rwCfgYAML) == 0` either means:
* The flag was provided but it's an empty file or empty string OR
* The flag was not provided at all.

I couldn't think of a reliable way to distinguish between them. I've also documented this behavior in the flag help, but I don't know if it is something we want. An alternative would be to provide a separate flag for enabling stateless mode (e.g `--remote.write` or `--remote-write.enabled`).

test/e2e/rule_test.go
idoqo: Turns out that adding alerting rules stops the ruler from starting up (causing the test to fail). Here's the relevant log message:
```
19:39:32 rule-1: panic: runtime error: invalid memory address or nil pointer dereference
19:39:32 rule-1: [signal SIGSEGV: segmentation violation code=0x1 addr=0x28 pc=0x8efd92]
19:39:32 rule-1: goroutine 354 [running]:
19:39:32 rule-1: github.com/prometheus/prometheus/storage.(*genericSeriesSetAdapter).Next(0xc000a13b90, 0xc000a42e10)
19:39:32 rule-1: <autogenerated>:1 +0x32
19:39:32 rule-1: github.com/prometheus/prometheus/storage.newGenericMergeSeriesSet(0xc000a849e0, 0x2, 0x2, 0xc000a13b60, 0xc000a13bd0, 0xc000cb1960)
19:39:32 rule-1: /home/michael/go/pkg/mod/github.com/prometheus/prometheus@v1.8.2-0.20210519120135-d95b0972505f/storage/merge.go:319 +0x16c
19:39:32 rule-1: github.com/prometheus/prometheus/storage.(*mergeGenericQuerier).Select.func4(0xc000a849c0, 0x3, 0x4)
19:39:32 rule-1: /home/michael/go/pkg/mod/github.com/prometheus/prometheus@v1.8.2-0.20210519120135-d95b0972505f/storage/merge.go:143 +0x50
19:39:32 rule-1: github.com/prometheus/prometheus/storage.(*lazyGenericSeriesSet).Next(0xc000a42de0, 0xc000a42de0)
19:39:32 rule-1: /home/michael/go/pkg/mod/github.com/prometheus/prometheus@v1.8.2-0.20210519120135-d95b0972505f/storage/lazy.go:28 +0x3b
19:39:32 rule-1: github.com/prometheus/prometheus/rules.(*Group).RestoreForState.func2(0xc0006ce0c0)
19:39:32 rule-1: /home/michael/go/pkg/mod/github.com/prometheus/prometheus@v1.8.2-0.20210519120135-d95b0972505f/rules/manager.go:751 +0x172
19:39:32 rule-1: github.com/prometheus/prometheus/rules.(*AlertingRule).ForEachActiveAlert(0xc000a87ba0, 0xc000cb1d00)
19:39:32 rule-1: /home/michael/go/pkg/mod/github.com/prometheus/prometheus@v1.8.2-0.20210519120135-d95b0972505f/rules/alerting.go:477 +0xed
19:39:32 rule-1: github.com/prometheus/prometheus/rules.(*Group).RestoreForState(0xc0008620e0, 0xc028b4f937e8d55f, 0x790615b7, 0x2f1eb60)
19:39:32 rule-1: /home/michael/go/pkg/mod/github.com/prometheus/prometheus@v1.8.2-0.20210519120135-d95b0972505f/rules/manager.go:736 +0x377
19:39:32 rule-1: github.com/prometheus/prometheus/rules.(*Group).run(0xc0008620e0, 0x2115af0, 0xc0009682d0)
19:39:32 rule-1: /home/michael/go/pkg/mod/github.com/prometheus/prometheus@v1.8.2-0.20210519120135-d95b0972505f/rules/manager.go:400 +0xb98
19:39:32 rule-1: github.com/prometheus/prometheus/rules.(*Manager).Update.func1(0xc00059e200, 0x0, 0xc000041280, 0xc000142780, 0xc0008620e0)
19:39:32 rule-1: /home/michael/go/pkg/mod/github.com/prometheus/prometheus@v1.8.2-0.20210519120135-d95b0972505f/rules/manager.go:988 +0x8d
19:39:32 rule-1: created by github.com/prometheus/prometheus/rules.(*Manager).Update
19:39:32 rule-1: /home/michael/go/pkg/mod/github.com/prometheus/prometheus@v1.8.2-0.20210519120135-d95b0972505f/rules/manager.go:978 +0x5a5
```

Not sure, but I'm guessing it's because of how the WAL storage implements the Querier interface (https://github.com/idoqo/thanos/blob/7d6095cf27dd8e827db174bcb93ccda722c9ea6e/pkg/rules/remotewrite/wal.go#L678)

bwplotka: Thatrs important bug that we have to solve. Looks like alerting was essentially checking WAL for something, which is not possible now, - we need to track what. I think this is the `RestoreForState` function - we probably have to ask querier for it or something - this might need some thoughts.

Great your test caught it!

yeya24 (commented):
Let's fix the conflict and linting issue.

ianbillett:
@idoqo feels like we are very close to being ready to rock & roll :rocket: just a couple of linters & conflicts to fix

GiedriusS:
@idoqo any updates?

pkg/rules/remotewrite/remotewrite.go
```diff
@@ -0,0 +1,49 @@
...
+	"github.com/prometheus/client_golang/prometheus"
+	"github.com/prometheus/prometheus/config"
+	"github.com/prometheus/prometheus/storage"
+	"github.com/prometheus/prometheus/storage/remote"
+	"gopkg.in/yaml.v2"
+)
+
+// Config represents a remote write configuration for Thanos stateless ruler.
+type Config struct {
+	Name        string                    `yaml:"name"`
```
yeya24, line 21: Why do we need `name`? Where is it used?

yeya24: Can we remove this `Config` struct and use the `RemoteWriteConfig` from Prometheus directly?

idoqo: > Can we remove this `Config` struct and use the `RemoteWriteConfig` from Prometheus directly?

Right, and `RemoteWriteConfig` already has the `name` field for it to use.

cmd/thanos/rule.go
```diff
@@ -319,25 +325,52 @@ func runRule(
...
+		}
+		walDir := filepath.Join(conf.dataDir, rwCfg.Name)
+		remoteStore, err := remotewrite.NewFanoutStorage(logger, reg, walDir, rwCfg)
+		if err != nil {
+			return errors.Wrap(err, "set up remote-write store for ruler")
+		}
+		appendable = remoteStore
+		queryable = remoteStore
+	} else {
+		db, err := tsdb.Open(conf.dataDir, log.With(logger, "component", "tsdb"), reg, tsdbOpts, nil)
```
yeya24, line 353: We can use `=` instead of `:=` here

yeya24 (changes requested):
Btw I got panic at line 559. In the case of using remote write storage, `db` will be nil and cause panic later.

```
	tsdbStore := store.NewTSDBStore(logger, db, component.Rule, conf.lset)
```

We need to implement a `TSDBReader` based on the local Wal store and the remote storage as well. However, I don't think the two stores have real `Queryable` implementation so currently it is still not working.

pkg/rules/remotewrite/remotewrite.go
yeya24: Oh, I see. Thanks for the clarification.

cmd/thanos/rule.go
idoqo: `tsdbStore` is used in creating the Store API server for ruler and the way I see it, running ruler in "stateless" mode means it's no longer a Store - maybe it makes sense to only create `tsdbStore` if `db` is not nil?

yeya24: Oh, that makes sense. If we don't want to consider it as store, let's check db and also stop grpc server is remote write is enabled.

yeya24:
@idoqo The failling E2E test is related to your changes. Can you take a look? Also please resolve the conflict.

pkg/rules/remotewrite/remotewrite.go
```diff
@@ -0,0 +1,43 @@
...
+}
+
+// NewFanoutStorage creates a storage that fans-out to both the WAL and a configured remote storage.
+// The remote storage tails the WAL and sends the metrics it reads using Prometheus' remote_write.
+func NewFanoutStorage(logger log.Logger, reg prometheus.Registerer, walDir string, rwConfig *config.RemoteWriteConfig) (storage.Storage, error) {
+	walStore, err := NewStorage(logger, reg, walDir)
+	if err != nil {
+		return nil, err
+	}
+	remoteStore := remote.NewStorage(logger, reg, walStore.StartTime, walStore.Directory(), 1*time.Minute, nil)
```
yeya24, line 35: Let's make `FlushDeadline` configurable as a flag or part of the remote write config file? I feel 1 minute is too long for E2E tests.

cmd/thanos/rule.go
```diff
@@ -522,7 +556,7 @@ func runRule(
 	)
 
 	// Start gRPC server.
-	{
+	if db != nil {
```
yeya24, line 559: I think I missed it last time. Even if it is stateless mode, we still need to start the gRPC server because we need the prober and the `gRPC` rules feature, we can just remove the StoreServer in this case.

Example code:

```go
	// Start gRPC server.
	tlsCfg, err := tls.NewServerConfig(log.With(logger, "protocol", "gRPC"), conf.grpc.tlsSrvCert, conf.grpc.tlsSrvKey, conf.grpc.tlsSrvClientCA)
	if err != nil {
		return errors.Wrap(err, "setup gRPC server")
	}

	options := []grpcserver.Option{
		grpcserver.WithServer(thanosrules.RegisterRulesServer(ruleMgr)),
		grpcserver.WithListen(conf.grpc.bindAddress),
		grpcserver.WithGracePeriod(time.Duration(conf.grpc.gracePeriod)),
		grpcserver.WithTLSConfig(tlsCfg),
	}
	if db != nil {
		tsdbStore := store.NewTSDBStore(logger, db, component.Rule, conf.lset)
		options = append(options, grpcserver.WithServer(store.RegisterStoreServer(tsdbStore)))
	}
	// TODO: Add rules API implementation when ready.
	s := grpcserver.New(logger, reg, tracer, grpcLogOpts, tagOpts, comp, grpcProbe, options...)

	g.Add(func() error {
		statusProber.Ready()
		return s.ListenAndServe()
	}, func(err error) {
		statusProber.NotReady(err)
		s.Shutdown(err)
	})
```

yeya24 (changes requested):

yeya24:
Hi @idoqo, are you still willing to finish this pr? Should be close to merge I think. If you are busy with other work then I am happy to take over.

idoqo:
> Hi @idoqo, are you still willing to finish this pr? Should be close to merge I think. If you are busy with other work then I am happy to take over.

Hi @yeya24, I'm a bit held up by exams atm so, sure

yeya24:
Thanks for the work. Close it in favor of #4731

PR #4253, opened 2021-05-19
Expose downsampleRaw for testing

yeya24:

## Changes

This pr exposes `downsampleRaw` for testing purpose.

### DIFF
--- a/pkg/compact/downsample/downsample.go
+++ b/pkg/compact/downsample/downsample.go
@@ -19,6 +19,8 @@ import (
 	"github.com/prometheus/prometheus/tsdb/chunkenc"
 	"github.com/prometheus/prometheus/tsdb/chunks"
 	"github.com/prometheus/prometheus/tsdb/index"
+	"github.com/prometheus/prometheus/tsdb/tsdbutil"
+
 	"github.com/thanos-io/thanos/pkg/block/metadata"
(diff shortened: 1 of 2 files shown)


pkg/compact/downsample/downsample.go
```diff
@@ -100,7 +102,7 @@ func Downsample(
 
 	var (
 		aggrChunks []*AggrChunk
-		all        []sample
+		all        []tsdbutil.Sample
```
bwplotka, line 105: So in theory this struct was supposed to be used only for testing. It's controversial, but the idea is that we wanted to have reproduced sample struct everywhere as it's easy to copy such small struct and easier to type it (as you could see when refactoring).

Indeed it makes it's difficult to reuse though... I am not sure if we want to use Prometheus struct designed for testing though in our production code 🤔

yeya24: Got it. I see. Then what about exposing `downsample.sample` struct and use it?

yeya24: Updated with the way mentioned in the OH.

kakkoyun (approved):
LGTM 🚀

yeya24:
Will merge in 1h if there is no objection.

bwplotka (approved):
Thanks! LGTM 💪🏽

PR #4258, opened 2021-05-20
Mixin: exclude rule generation if component has been excluded

### LINKED ISSUE #4260: Mixin: Rule generation fails if component is excluded
**Thanos, Prometheus and Golang version used**: x

**Object Storage Provider**: x

**What happened**:

When excluding `receive` I found the rule generation part of mixins no longer worked.

Rule generation doesn't error out.

In your mixin config:

```jsonnet
  query+:: {
    selector: 'job=~".*thanos-query.*"',
    title: '%(prefix)sQuery' % $.dashboard.prefix,
  },
  receive: null
```

When generating the rules, the entire generation process for rules will fail as `receive` no longer has the `dimensions` or `selector` variables. 

jalev:

## Changes

Following on from #3883, if a components is excluded via configuration you can't create rules for it. Resolves #4260

## Verification

Tested it locally by turning off `receive` via our own mixins (using thanos as a mixin library). Since `receive`'s object is now `null` there are no `dimensions` or `selector`s variables and we get an error during generation:

```
RUNTIME ERROR: Not enough values to format: 1, expected more than 1
        std.jsonnet:710:15-103  thunk <val>
        std.jsonnet:715:27-30   thunk <val>
        std.jsonnet:585:22-25   thunk <a>
        std.jsonnet:36:17
        std.jsonnet:36:8-19     thunk <a>
        std.jsonnet:36:8-31     function <anonymous>
        std.jsonnet:36:8-31     function <anonymous>
        std.jsonnet:585:9-26    function <format_code>
        std.jsonnet:715:15-60   thunk <s>
        std.jsonnet:720:24      thunk <str>
        ...
        std.jsonnet:1002:33-72  thunk <array_element>
        std.jsonnet:1006:11-42  function <aux>
        std.jsonnet:1030:59-98  thunk <array_element>
        std.jsonnet:1034:11-42  function <aux>
        std.jsonnet:1002:33-72  thunk <array_element>
        std.jsonnet:1006:11-42  function <aux>
        std.jsonnet:1030:59-98  thunk <array_element>
        std.jsonnet:1034:11-42  function <aux>
        std.jsonnet:1035:5-23   function <anonymous>
        ./service/thanos/prometheus_rules.jsonnet:(1:1)-(6:2)
```

### DIFF
--- a/mixin/rules/bucket_replicate.libsonnet
+++ b/mixin/rules/bucket_replicate.libsonnet
@@ -4,7 +4,7 @@
     selector: error 'must provide selector for Thanos Bucket Replicate dashboard',
   },
   prometheusRules+:: {
-    groups+: [
+    groups+: if thanos.bucket_replicate == null then [] else [
       {
         name: 'thanos-bucket-replicate.rules',
         rules: [],
--- a/mixin/rules/query.libsonnet
+++ b/mixin/rules/query.libsonnet
@@ -5,7 +5,7 @@
     dimensions: std.join(', ', std.objectFields(thanos.targetGroups) + ['job']),
   },
   prometheusRules+:: {
-    groups+: [
+    groups+: if thanos.query == null then [] else [
       {
         name: 'thanos-query.rules',
         rules: [
--- a/mixin/rules/receive.libsonnet
+++ b/mixin/rules/receive.libsonnet
@@ -5,7 +5,7 @@
     dimensions: std.join(', ', std.objectFields(thanos.targetGroups) + ['job']),
   },
   prometheusRules+:: {
-    groups+: [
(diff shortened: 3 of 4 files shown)


kakkoyun (commented):
Thanks for the contribution.
From the looks of it, it seems fine. However, I'm confused by your statement in the Verification step. Is it still broken after your fix?

jalev:
> Thanks for the contribution.
> From the looks of it, it seems fine. However, I'm confused by your statement in the Verification step. Is it still broken after your fix?

Oh, no, it's fine after this fix is applied, sorry for the confusion.

kakkoyun (approved):

PR #4268, opened 2021-05-21
fix: add missing vendor script for offline access

### LINKED ISSUE #4263: running thanos in a K8S cluster without internet access

Currently, it seems that Thanos is requiring getting its JS script from internet:
https://github.com/thanos-io/thanos/blob/main/pkg/ui/templates/bucket.html
This is a blocking point for air-gapped K8S cluster.

Provide all the necessary scripts if internet is not accessible.

Aman-Codes:

## Changes

Added missing vendor script for offline access.
Closes #4263 

## Verification

Access Thanos without internet


EtienneBarbier:
Hi,

It is not so simple, loader.js call other script on internet. Try to search `gstatic.com` in the file.

In addition, it seem not legal to use this api offline.
https://developers.google.com/chart/interactive/faq#using-the-api

onprem:
Heya, as @EtienneBarbier mentioned, we have already done discussion around this and vendoring the Google charts library violates it's terms. The old UI is already deprecated and is replaced by the React based UI which doesn't use Google charts.

Aman-Codes:
> Heya, as @EtienneBarbier mentioned, we have already done discussion around this and vendoring the Google charts library violates it's terms. The old UI is already deprecated and is replaced by the React based UI which doesn't use Google charts.

Okay thank you for these insights. Closing this PR in favour of #4163

PR #4271, opened 2021-05-21
store: send data immediately after getting it

GiedriusS:
Build a min-heap to immediately send back data to the querier after
getting it. Min heap is needed to maintain the invariant of Series()
that the returned data is sorted by the labels. My anecdotal data shows
that this could shave off a few percent of the total duration of queries
because we aren't doing one sorting pass at the end _after_ getting
_all_ of the data and we can free memory quicker.

I don't think any of the current benchmarks can show this difference?
It's probably the most visible when a Series() comes in which matches a
lot of series.

I have only done minimal testing - that I am still able to query data
via `/api/v1/query_range`.

What do you think @bwplotka @kakkoyun @yeya24 ?

### DIFF
--- a/pkg/store/bucket.go
+++ b/pkg/store/bucket.go
@@ -750,7 +750,6 @@ func (s *bucketSeriesSet) Err() error {
 
 // blockSeries returns series matching given matchers, that have some data in given time range.
 func blockSeries(
-	ctx context.Context, // Context for all of the operations, used in all readers as well.
 	extLset labels.Labels, // External labels added to the returned series labels.
 	indexr *bucketIndexReader, // Index reader for block.
 	chunkr *bucketChunkReader, // Chunk reader for block.
@@ -760,7 +759,7 @@ func blockSeries(
 	skipChunks bool, // If true, chunks are not loaded.
 	minTime, maxTime int64, // Series must have data in this time range to be returned.
 	loadAggregates []storepb.Aggr, // List of aggregates to load when loading chunks.
-	h *seriesResponsesHeap,
+	h *seriesResponsesHeap, // Min heap for the series responses.
 ) (storepb.SeriesSet, *queryStats, error) {
 	ps, err := indexr.ExpandedPostings(matchers)
 	if err != nil {
@@ -793,9 +792,6 @@ func blockSeries(
 	)
 
 	var chunkLoader *errgroup.Group
-	if chunkr != nil {
-		chunkLoader, _ = errgroup.WithContext(ctx)
-	}
 	for _, id := range ps {
 		ok, err := indexr.LoadSeriesForTime(id, &symbolizedLset, &chks, skipChunks, minTime, maxTime)
 		if err != nil {
@@ -837,9 +833,8 @@ func blockSeries(
 
 		if h != nil {
 			entry := h.p.Get().(*seriesResponseEntry)
-			entry.lbls = &s.lset
 			entry.series = &s
-			entry.loadingGroup = chunkLoader
+			entry.loadingGroup = &chunkLoader
 			heap.Push(h, entry)
 		}
 		res = append(res, s)
@@ -849,9 +844,8 @@ func blockSeries(
 		return newBucketSeriesSet(res), indexr.stats, nil
 	}
 
-	chunkLoader.Go(func() error {
-		return errors.Wrap(chunkr.load(res, loadAggregates), "load chunks")
-	})
+	g := chunkr.load(res, loadAggregates)
+	chunkLoader = g
 
 	return newBucketSeriesSet(res), indexr.stats.merge(chunkr.stats), nil
 }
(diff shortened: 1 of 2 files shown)


bwplotka:
Thanks, overall there is a lot of room for improvement so great to see some movement here! (: 
 
So essentially we are replacing merge sort that joins data from blocks (series within block are sorted) with min heap?

I will review in June in more detail, but I would be against merging anything on this level without local Go benchmark that shows visible improvement. This is because it might surprising that we actually allocate here more than less etc 🙈

vovanloc0798 (approved):

vovanloc0798 (dismissed):

GiedriusS:
As expected, CPU usage savings are great :scream: however there are some new memory allocations somewhere:
```
benchmark                                   old ns/op     new ns/op     delta
BenchmarkBlockSeries/concurrency:_1-16      9112258       3138514       -65.56%
BenchmarkBlockSeries/concurrency:_2-16      4902091       2117883       -56.80%
BenchmarkBlockSeries/concurrency:_4-16      2963023       1376239       -53.55%
BenchmarkBlockSeries/concurrency:_8-16      2247469       1054152       -53.10%
BenchmarkBlockSeries/concurrency:_16-16     2105249       861687        -59.07%
BenchmarkBlockSeries/concurrency:_32-16     3022499       832255        -72.46%

benchmark                                   old allocs     new allocs     delta
BenchmarkBlockSeries/concurrency:_1-16      31525          31546          +0.07%
BenchmarkBlockSeries/concurrency:_2-16      31352          21138          -32.58%
BenchmarkBlockSeries/concurrency:_4-16      32703          13817          -57.75%
BenchmarkBlockSeries/concurrency:_8-16      33396          10492          -68.58%
BenchmarkBlockSeries/concurrency:_16-16     32935          8976           -72.75%
BenchmarkBlockSeries/concurrency:_32-16     46036          9508           -79.35%

benchmark                                   old bytes     new bytes     delta
BenchmarkBlockSeries/concurrency:_1-16      5140092       16535258      +221.69%
BenchmarkBlockSeries/concurrency:_2-16      5295731       11614433      +119.32%
BenchmarkBlockSeries/concurrency:_4-16      5235786       8172685       +56.09%
BenchmarkBlockSeries/concurrency:_8-16      5442007       6578585       +20.89%
BenchmarkBlockSeries/concurrency:_16-16     6131926       5803753       -5.35%
BenchmarkBlockSeries/concurrency:_32-16     8115555       6078792       -25.10%
```

Trying to dig more into it

pkg/store/bucket.go
```diff
@@ -1097,42 +1191,88 @@ func (s *BucketStore) Series(req *storepb.SeriesRequest, srv storepb.Store_Serie
...
+				shouldSend = labels.Compare(*v.lbls, *lastLbls) != 0
+			}
 
-			var lset labels.Labels
-			if req.SkipChunks {
-				lset, _ = set.At()
-			} else {
-				lset, series.Chunks = set.At()
+			if shouldSend {
+				set := storepb.MergeSeriesSets(newBucketSeriesSet(ss))
```
yeya24, line 1211: Is `storepb.MergeSeriesSets` necessary here? The input is only one series set.

pkg/store/bucket.go
```diff
@@ -1097,42 +1191,88 @@ func (s *BucketStore) Series(req *storepb.SeriesRequest, srv storepb.Store_Serie
...
-			if err = srv.Send(storepb.NewSeriesResponse(&series)); err != nil {
-				err = status.Error(codes.Unknown, errors.Wrap(err, "send series response").Error())
-				return
+
+			ser := v.series
+			ss = append(ss, *ser)
+
+			if h.Len() == 0 && !shouldSend {
+				// Send again! ===
+				set := storepb.MergeSeriesSets(newBucketSeriesSet(ss))
```
yeya24, line 1244: ditto

pkg/store/bucket.go
```diff
@@ -947,6 +963,76 @@ func debugFoundBlockSetOverview(logger log.Logger, mint, maxt, maxResolutionMill
 	level.Debug(logger).Log("msg", "Blocks source resolutions", "blocks", len(bs), "Maximum Resolution", maxResolutionMillis, "mint", mint, "maxt", maxt, "lset", lset.String(), "spans", strings.Join(parts, "\n"))
 }
 
+type seriesResponseEntry struct {
+	lbls         *labels.Labels
```
yeya24, line 967: This `lbls` is the same as `series.lset`? Do we need to keep both?

pkg/store/bucket.go
```diff
@@ -977,8 +1063,14 @@ func (s *BucketStore) Series(req *storepb.SeriesRequest, srv storepb.Store_Serie
 		reqBlockMatchers []*labels.Matcher
 		chunksLimiter    = s.chunksLimiterFactory(s.metrics.queriesDropped.WithLabelValues("chunks"))
 		seriesLimiter    = s.seriesLimiterFactory(s.metrics.queriesDropped.WithLabelValues("series"))
+		h                = &seriesResponsesHeap{
+			entries: make([]*seriesResponseEntry, 0),
+			mtx:     &sync.Mutex{},
```
yeya24, line 1068: sync pool needs to be initialized here.

GiedriusS:
Actually, this probably isn't worth it due to how Go's memory management works - even though we'd "free" the memory quicker, it doesn't reflect on the memory management side before the next GC pass. It would be best to have "full streaming" without a separate buffering & sorting step - then the overall gains from such improvement would be clearer. Thus, closing this (:

PR #4272, opened 2021-05-21
Refactored tracing.md file

iamrajiv:


docs/tracing.md
```diff
@@ -106,18 +110,19 @@ config:
...
 type: LIGHTSTEP
 config:
   access_token: ""
   collector:
-    scheme: ""
-    host: ""
-    port: 0
-    plaintext: false
-    custom_ca_cert_file: ""
+  scheme: ""
```
wiardvanrij, line 122: Is this intended? What changed? Because this should have been autogenerated from the `embedmd` right?

docs/tracing.md
```diff
@@ -6,41 +6,41 @@ menu: thanos
...
-            bucket: <bucket>
-        - --prometheus.url=http://localhost:9090
-        - |
-          --tracing.config=type: STACKDRIVER
-          config:
-            service_name: ""
-            project_id: <project>
-            sample_factor: 16
-        - --tsdb.path=/prometheus-data
+- args:
```
wiardvanrij, line 23: I don't think the indentation is right on this part.

wiardvanrij (commented):
Thanks a lot for the changes. I'm wondering if the `embedmd` is still working with this change? Could you confirm this? Because I'm wondering since the indentation is wrong (at-least, I think it is)

iamrajiv:
@wiardvanrij I am not sure how `embedmd` works like I edited the `.md` file directly.

PR #4273, opened 2021-05-25
*: Added metrics exposed

Abhishek357:

## Changes

Added a new page to reflect the metrics exposed by Thanos to instrument itself using [promlinter](https://github.com/yeya24/promlinter).

## Verification
Using make web-serve, the website now shows the metrics Type, Name, and Help.

### DIFF
--- a/docs/instrument-thanos.md
+++ b/docs/instrument-thanos.md
@@ -1,0 +1,171 @@
+---
+title: Instrumenting Thanos
+type: docs
+menu: thanos
+---
+
+# Instrumenting Thanos
+We use the Prometheus [Go Client](https://github.com/prometheus/client_golang) library to instrument Thanos itself. Following are the metrics exposed by thanos:
+
+```bash
+TYPE                NAME                                                                 HELP
+GAUGE               thanos_delete_delay_seconds                                          Configured delete delay in seconds.
+GAUGE               thanos_compact_halted                                                Set to 1 if the compactor halted due to an unexpected error.
+COUNTER             thanos_compact_retries_total                                         Total number of retries after retriable compactor error.
+COUNTER             thanos_compact_iterations_total                                      Total number of iterations that were executed successfully.
+COUNTER             thanos_compact_block_cleanup_loops_total                             Total number of concurrent cleanup loops of partially uploaded blocks and marked blocks that were executed successfully.
+COUNTER             thanos_compact_aborted_partial_uploads_deletion_attempts_total       Total number of started deletions of blocks that are assumed aborted and only partially uploaded.
+COUNTER             thanos_compact_blocks_cleaned_total                                  Total number of blocks deleted in compactor.
+COUNTER             thanos_compact_block_cleanup_failures_total                          Failures encountered while deleting blocks in compactor.
+COUNTER             thanos_compact_blocks_marked_total                                   Total number of blocks marked in compactor.
+COUNTER             thanos_compact_garbage_collected_blocks_total                        Total number of blocks marked for deletion by compactor.
+COUNTER             thanos_compact_downsample_total                                      Total number of downsampling attempts.
+COUNTER             thanos_compact_downsample_failures_total                             Total number of failed downsampling attempts.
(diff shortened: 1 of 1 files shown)


docs/instrument-thanos.md
```diff
@@ -0,0 +1,171 @@
+---
+title: Instrumenting Thanos
```
yeya24, line 2: `Exposed Metrics` would be better.

docs/instrument-thanos.md
```diff
@@ -0,0 +1,171 @@
...
+type: docs
+menu: thanos
+---
+
+# Instrumenting Thanos
+We use the Prometheus [Go Client](https://github.com/prometheus/client_golang) library to instrument Thanos itself. Following are the metrics exposed by thanos:
+
+```bash
+TYPE                NAME                                                                 HELP
+GAUGE               thanos_delete_delay_seconds                                          Configured delete delay in seconds.
```
yeya24, line 12: Thanks for using `promlinter`. However, the problem here is:
Promlinter is just a static analysis tool, which means it cannot catch some metrics registered in the runtime. So it cannot catch all the metrics instrumented in the codebase. 

For example, it cannot catch the registry prefix.

docs/instrument-thanos.md
```diff
@@ -0,0 +1,171 @@
...
+COUNTER             http_requests_total                                                  Tracks the number of HTTP requests.
+SUMMARY             http_response_size_bytes                                             Tracks the size of HTTP responses.
+COUNTER             thanos_objstore_bucket_operations_total                              Total number of all attempted operations against a bucket.
+COUNTER             thanos_objstore_bucket_operation_failures_total                      Total number of operations against a bucket that failed, but were not expected to fail in certain way from caller perspective. Those errors have to be investigated.
+HISTOGRAM           thanos_objstore_bucket_operation_duration_seconds                    Duration of successful operations against the bucket
+GAUGE               thanos_objstore_bucket_last_successful_upload_time                   Second timestamp of the last successful upload to the bucket.
+GAUGE               status                                                               Represents status (0 indicates failure, 1 indicates success) of the component.
```
yeya24, line 91: This is an example of missing registry prefix. The full metric name would be `thanos_status`, but the linter can only catch part of it.

yeya24 (commented):
T

docs/instrument-thanos.md
Abhishek357: Should we manually add those?

bwplotka: There is, please use the same flow as in other docs like here: https://github.com/thanos-io/thanos/blob/f4a5904b0c3be7565298cf257ad43ac1942021cc/docs/components/compact.md#L318

This points to some output in some file and is controlled by script here https://github.com/thanos-io/thanos/blob/7b45f70759071d8d980925cdd94da5e391bcb6bc/scripts/genflagdocs.sh#L59

In future we will switch to mdox which makes it much simpler https://github.com/bwplotka/mdox#code-generation (we are working on it with @saswatamcode )

Abhishek357: Even if I auto-generate this page some of the names of the metrics will be misleading. So, how should I move forward by using promlinter? 
Maybe we can write a new script that runs Thanos using the quickstart script and use /metrics endpoint to get all the exposed metrics and follow similar flags flow as Bartek mentioned. WDYT?

yeya24: Get metrics from `/metrics` cannot cover all the metrics as well.
I think the problem is caused by the registry prefix pattern. It is not linter friendly and makes things complex. I think the best way is to remove the prefixes and add prefixes as metric namespace.

bwplotka: > Even if I auto-generate this page some of the names of the metrics will be misleading.

Let's discuss this, are we sure we cannot get prefixed metric or metric names there with promlinter? (: cc @yeya24

bwplotka: Related discussion https://github.com/prometheus/client_golang/pull/890

PR #4275, opened 2021-05-26
Use iterators to expand chunks when downsample

### LINKED ISSUE #2542: compact: Ensure downsampled chunks are not larger than 120 samples; stream downsampling more
During bug fixing on https://github.com/thanos-io/thanos/pull/2528 I found that downsampling always encodes whatever is given in the block into huge chunks. This can lead to inefficiency during query time when only a small part of chunk data is needed, but store GW needs to fetch and decode everything. 

See: https://github.com/thanos-io/thanos/blob/55cb8ca38b3539381dc6a781e637df15c694e50a/pkg/compact/downsample/downsample.go#L141

AC:
* Downsampled chunks are not larger than 120 samples
* Chunks are expanded on demand in iterator.

yeya24:

Fix #2542 

## Changes

Use iterators to expand chunks when downsampling is performed.

## Verification

Existing tests should still pass since this is just refactoring.

### DIFF
--- a/pkg/compact/downsample/downsample.go
+++ b/pkg/compact/downsample/downsample.go
@@ -146,7 +146,7 @@ func Downsample(
 				maxt = c.MaxTime
 			}
 
-			numSamples += c.Chunk.NumSamples()
+			numSamples += chks[i].Chunk.NumSamples()
 		}
 
 		// Raw and already downsampled data need different processing.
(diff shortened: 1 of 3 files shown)


Larry5th (dismissed):

yeya24:
Would be great to get some attention on this.

PR #4279, opened 2021-05-27
clearify retention

Domi-cc:


docs/components/sidecar.md
```diff
@@ -30,7 +30,7 @@ If you choose to use the sidecar to also upload data to object storage:
...
-* The retention is recommended to not be lower than three times the min block duration, so 6 hours. This achieves resilience in the face of connectivity issues to the object storage since all local data will remain available within the Thanos cluster. If connectivity gets restored the backlog of blocks gets uploaded to the object storage.
+* The retention of the compactor is recommended to not be lower than three times the min block duration, so 6 hours. This achieves resilience in the face of connectivity issues to the object storage since all local data will remain available within the Thanos cluster. If connectivity gets restored the backlog of blocks gets uploaded to the object storage.
```
onprem, line 33:
```suggestion
* The retention of Prometheus is recommended to not be lower than three times the min block duration, so 6 hours. This achieves resilience in the face of connectivity issues to the object storage since all local data will remain available within the Thanos cluster. If connectivity gets restored the backlog of blocks gets uploaded to the object storage.
```

Well, at least this has proved that this line needed clarification :smile:.

The doc is talking about Prometheus' retention here. The context is, if Prometheus deletes a block before the Sidecar can upload it, you'll loose the data.

wiardvanrij: Maybe my github is bugged but this comment is resolved while the change still talks about the compactor component. This still should be Prometheus :)

Abhishek357 (changes requested):
@Domi-cc can you sign off your commits :)

Domi-cc:
@Abhishek357 done :)

Abhishek357:
@Domi-cc DCO bot still not happy because 2 commits are incorrectly signed. Go to Details and follow along.

onprem (approved):
LGTM, Thanks!

PR #4281, opened 2021-05-27
Nominate `metalmatze` and `onprem` as next release shepherds (for v0.21.0)

onprem:
Signed-off-by: Prem Saraswat <prmsrswt@gmail.com>

cc @metalmatze

### DIFF
--- a/docs/release-process.md
+++ b/docs/release-process.md
@@ -31,8 +31,8 @@ Release shepherd responsibilities:
 
 | Release   | Time of first RC         | Shepherd (GitHub handle)    |
 |-----------|--------------------------|-----------------------------|
-| v0.22.0   | (planned) 2021.06.16     | TBD                         |
-| v0.21.0   | (planned) 2021.05.04     | `@metalmatze` and `@onprem` |
+| v0.22.0   | (planned) 2021.07.06     | TBD                         |
+| v0.21.0   | 2021.05.28               | `@metalmatze` and `@onprem` |
 | v0.20.0   | 2021.04.23               | `@kakkoyun`                 |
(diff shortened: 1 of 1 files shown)


yeya24 (dismissed):

docs/release-process.md
```diff
@@ -29,26 +29,26 @@ Release shepherd responsibilities:
...
-| v0.10.0   | 2020.01.08               | `@GiedriusS`             |
-| v0.9.0    | 2019.11.26               | `@bwplotka`              |
-| v0.8.0    | 2019.10.09               | `@bwplotka`              |
-| v0.7.0    | 2019.08.28               | `@domgreen`              |
-| v0.6.0    | 2019.07.12               | `@GiedriusS`             |
-| v0.5.0    | 2019.06.31               | `@bwplotka`              |
+| Release   | Time of first RC         | Shepherd (GitHub handle)    |
+|-----------|--------------------------|-----------------------------|
+| v0.22.0   | (planned) 2021.06.16     | TBD                         |
+| v0.21.0   | (planned) 2021.05.04     | `@metalmatze` and `@onprem` |
```
squat, line 35: This date is already in the past. Do we mean June?

onprem: I am not sure if we want to modify that or not haha. It was correct for a while, but there were some delays in releases and we have drifted a bit from the planned dates,

hitanshu-mehta:
First release by @onprem 🎉

docs/release-process.md
wiardvanrij: I would suggest to make a new planning. This provides value to users to know when releases are coming so they can act accordingly (i.e. test RC's or upgrade because they are really eager for a feature).

PR #4282, opened 2021-05-27
Querier/Sidecar/StoreGW: Implement Info service and add `--endpoint` flag in Querier

hitanshu-mehta:
This PR implements a Unified Endpoint [Proposal](https://thanos.io/tip/proposals/202101_endpoint_discovery.md/). Open for reviews :)

## Changes

- Implemented Info server.
- Added Info server in Sidecar and StoreGW.
- Added `--endpoint` flag in querier.
- ~Modified storeset tests to verify discovery using `endpoint` flag.~ Completely removed `storeset` and added `endpointset` https://github.com/thanos-io/thanos/pull/4421

## Verification
- Manually tested using grpccurl.
- ~Modified storeset tests to verify discovery using endpoint flag.~

## TODO
- ~E2E tests.~
- Add Info server in remaining components.


cmd/thanos/query.go
```diff
@@ -96,6 +96,9 @@ func registerQuery(app *extkingpin.App) {
 	selectorLabels := cmd.Flag("selector-label", "Query selector labels that will be exposed in info endpoint (repeated).").
 		PlaceHolder("<name>=\"<value>\"").Strings()
 
+	endpoints := cmd.Flag("endpoint", "Addresses of statically configured API servers (repeatable). The scheme may be prefixed with 'dns+' or 'dnssrv+' to detect API servers through respective DNS lookups.").
```
squat, line 99: Maybe
```suggestion
	endpoints := cmd.Flag("endpoint", "Addresses of statically configured Thanos API servers (repeatable). The scheme may be prefixed with 'dns+' or 'dnssrv+' to detect API servers through respective DNS lookups.").
```

pkg/info/info.go
```diff
@@ -0,0 +1,78 @@
...
+	targetAPIPresent   bool
+	metadataAPIPresent bool
+}
+
+func NewInfoServer(
+	component infopb.ComponentType,
+	getLabelSet func() []labelpb.ZLabelSet,
+	getStoreInfo func() *infopb.StoreInfo,
+	getExemplarsInfo func() *infopb.ExemplarsInfo,
+	ruleAPIPresent bool,
```
squat, line 34: Let's make all of these parameters follow the same pattern, so that rather than passing a func for some and a bool for the other, they all accept a func. This will present a more coherent API for users and also ensure that we don't need to change this signature when we add fields, e.g. to ruleAPIInfo, in the future.

pkg/info/info.go
```diff
@@ -0,0 +1,78 @@
...
+	targetAPIPresent bool,
+	metadataAPIPresent bool,
+) *InfoServer {
+	return &InfoServer{
+		component:          component,
+		getLabelSet:        getLabelSet,
+		getStoreInfo:       getStoreInfo,
```
squat, line 41: For convenience, let's allow users to pass a `nil` func. If the func for any of these parameters is `nil`, we replace it with the corresponding `func() *infopb.XInfo { return nil }`

pkg/query/storeset.go
```diff
@@ -68,6 +69,14 @@ type ExemplarSpec interface {
 	Addr() string
 }
 
+type InfoSpec interface {
+	// Addr returns InofAPI Address for the info spec. It is used as its ID.
```
squat, line 73:
```suggestion
	// Addr returns InfoAPI Address for the info spec. It is used as its ID.
```

pkg/query/storeset.go
```diff
@@ -68,6 +69,14 @@ type ExemplarSpec interface {
...
+	Addr() string
+	// Metadata returns current labels, min/max ranges for StoreAPI and ExemplarAPI
```
squat, line 75: Let's start with the smaller abstraction and add this later if we see we need to generalize it to all Info types only if they all need it really. Otherwise, we can keep it in the StoreSpec

squat (commented):
Great work so far; some ideas from our weekly 1-1

bwplotka:
Is it rdy for review? (:

hitanshu-mehta:
I have to add e2e tests. Apart from that, PR is ready for review 🚀

cmd/thanos/store.go
```diff
@@ -379,6 +382,25 @@ func runStore(
...
+			return bs.LabelSet()
+		},
+		func() *infopb.StoreInfo {
+			mint, maxt := bs.TimeRange()
+			return &infopb.StoreInfo{
+				MinTime: mint,
+				MaxTime: maxt,
+			}
+		},
+		func() *infopb.ExemplarsInfo { return nil },
```
yeya24, line 398: Can we use `nil` here? What's the difference between nil and a function returning nil?

test/e2e/info_api_test.go
```diff
@@ -0,0 +1,172 @@
...
+			Insecure:  true,
+		},
+	})
+	testutil.Ok(t, err)
+	testutil.Ok(t, s.StartAndWaitReady(str))
+
+	// Register 1 sidecar using `--store`.
+	// Register 2 sidecars and 1 storeGW using `--endpoint`.
+	// Register `sidecar3` twice to verify it is deduplicated.
+	q, err := e2ethanos.NewQuerierBuilder(s.SharedDir(), "1", []string{sidecar1.GRPCNetworkEndpoint()}).
```
yeya24, line 60: If `endpoint` and specific stores are configured (`metadata`, `target`, etc) together, does it work? Can you add this test case?

cmd/thanos/sidecar.go
```diff
@@ -247,12 +250,46 @@ func runSidecar(
...
+					MaxTime: maxt,
+				}
+			},
+			func() *infopb.ExemplarsInfo {
+				// Currently Exemplars API does not expose metadata such as min/max time,
+				// so we are using default minimum and maximum possible values as min/max time.
+				return &infopb.ExemplarsInfo{
+					MinTime: time.Unix(math.MinInt64/1000+62135596801, 0).UTC().Unix(),
+					MaxTime: time.Unix(math.MaxInt64/1000-62135596801, 999999999).UTC().Unix(),
+				}
```
yeya24, line 273: We can use global variable for the two values to avoid calculating them each time.

cmd/thanos/store.go
hitanshu-mehta: Sorry missed it. Yes, we can pass `nil` as we are doing with other parameters. We are allowing to pass `nil` for convenience. https://github.com/thanos-io/thanos/pull/4282#discussion_r641449154

test/e2e/info_api_test.go
hitanshu-mehta: Yes sure. I will add this test case.

hitanshu-mehta:
I had to rename messages `InfoRequest` and `InfoResponse` because messages with these names are already defined in [proto file](https://github.com/thanos-io/thanos/blob/main/pkg/store/storepb/rpc.proto) of storeAPI. So, a warning for `duplicate proto type registered` was given. This warning message was also being printed in component-specific files when we run `make docs`.

pkg/query/storeset.go
```diff
@@ -312,6 +329,21 @@ func (s *storeRef) Update(labelSets []labels.Labels, minTime int64, maxTime int6
 	s.exemplar = exemplar
 }
 
+func (s *storeRef) UpdateWithStore(labelSets []labels.Labels, minTime int64, maxTime int64, storeType component.StoreAPI, store storepb.StoreClient, rule rulespb.RulesClient, target targetspb.TargetsClient, metadata metadatapb.MetadataClient, exemplar exemplarspb.ExemplarsClient) {
```
GiedriusS, line 332: Nit: maybe this could be folded into `Update` and then we could simply add:

```go
if store != nil {
  s.StoreClient = store
}
```

This would save a bit of code duplication.

pkg/query/storeset_test.go
```diff
@@ -130,6 +131,93 @@ func (s *testStores) CloseOne(addr string) {
...
+	srvs       map[string]*grpc.Server
+	orderAddrs []string
+}
+
+func startInfoSrvs(infoMetas []testInfoMeta) (*testInfoSrvs, error) {
+	info := &testInfoSrvs{
+		srvs: map[string]*grpc.Server{},
+	}
+
+	for _, meta := range infoMetas {
```
GiedriusS, line 162: Need to be careful here and add:
```
meta := meta
```

Otherwise, you'll have problems: https://play.golang.org/p/dCePpqfNcEh

test/e2e/info_api_test.go
```diff
@@ -0,0 +1,178 @@
...
+		testutil.Ok(t, err)
+
+		body, err := ioutil.ReadAll(resp.Body)
+		testutil.Ok(t, err)
+
+		var res struct {
+			Data map[string][]query.StoreStatus `json:"data"`
+		}
+
+		if err = json.Unmarshal(body, &res); err != nil {
```
GiedriusS, line 144: Let's use `testutil.Ok()` :P

test/e2e/info_api_test.go
```diff
@@ -0,0 +1,178 @@
...
+		return nil
+	})
+	testutil.Ok(t, err)
+}
+
+func assertStoreStatus(t *testing.T, component string, res map[string][]query.StoreStatus, expected map[string][]query.StoreStatus) error {
+	t.Helper()
+
+	if len(res[component]) != len(expected[component]) {
+		return fmt.Errorf("Expected %d %s, got: %d", len(expected[component]), component, len(res[component]))
```
GiedriusS, line 165: Errors typically begin with a lower-case letter so that they would look nice when wrapped
```suggestion
		return fmt.Errorf("expected %d %s, got: %d", len(expected[component]), component, len(res[component]))
```

test/e2e/info_api_test.go
```diff
@@ -0,0 +1,178 @@
...
+	}
+
+	url := "http://" + path.Join(q.HTTPEndpoint(), "/api/v1/stores")
+
+	ctx, cancel := context.WithTimeout(context.Background(), time.Second*15)
+	defer cancel()
+
+	err = runutil.Retry(time.Second, ctx.Done(), func() error {
+
+		resp, err := http.Get(url)
```
GiedriusS, line 134: Nit: maybe we could call this with `ctx`? There could be problems if this hangs indefinitely for some reason

GiedriusS (commented):
Just some comments/suggestions but overall LGTM :muscle: nice work!

GiedriusS:
> I had to rename messages `InfoRequest` and `InfoResponse` because messages with these names are already defined in [proto file](https://github.com/thanos-io/thanos/blob/main/pkg/store/storepb/rpc.proto) of storeAPI. So, a warning for `duplicate proto type registered` was given. This warning message was also being printed in component-specific files when we run `make docs`.

Maybe let's call them `InfoAPIRequest` and `InfoAPIResponse` or something to differentiate between those two? We can discuss this in #thanos-dev, maybe someone has even better suggestions

cmd/thanos/sidecar.go
yeya24: @hitanshu-mehta We are still calling `.Unix()` every time. Can't we just initialize `v1.MinTime` and `v1.MaxTime` as int64? Are they used anywhere else?

yeya24 (commented):
I am wondering how would this API work with routing only receivers? IMO routing only receivers shouldn't expose any store API. Can you add a test case for this? cc @yashrsharma44 @hitanshu-mehta

cmd/thanos/sidecar.go
hitanshu-mehta: Currently, they are not being used anywhere else. Agree, I should've used int64. I'll correct it. Thank you!

pkg/query/storeset_test.go
hitanshu-mehta: 😲 Wasn't aware of this behavior! Thank you for pointing out :)

cmd/thanos/query.go
```diff
@@ -96,6 +96,9 @@ func registerQuery(app *extkingpin.App) {
 	selectorLabels := cmd.Flag("selector-label", "Query selector labels that will be exposed in info endpoint (repeated).").
 		PlaceHolder("<name>=\"<value>\"").Strings()
 
+	endpoints := cmd.Flag("endpoint", "Addresses of statically configured Thanos API servers (repeatable). The scheme may be prefixed with 'dns+' or 'dnssrv+' to detect Thanos API servers through respective DNS lookups.").
```
bwplotka, line 99: I would suggest removing `store` at this point. Or are we planning that for future PRs?

bwplotka: Essentially what's the deprecation plan? Can we note the deprecation plan in the store flag comment and changelog?

cmd/thanos/store.go
```diff
@@ -379,6 +382,25 @@ func runStore(
 			cancel()
 		})
 	}
+
+	infoSrv := info.NewInfoServer(
```
bwplotka, line 386: What's the deprecation plan for ths old Info methods? Can we unify implementation somehow at least?

pkg/info/infopb/rpc.proto
```diff
@@ -23,12 +23,12 @@ option (gogoproto.goproto_sizecache_all) = false;
 /// Info represents the API that is responsible for gathering metadata about the all APIs supported by the component.
 service Info {
     /// Info returns the metadata (Eg. LabelSets, Min/Max time) about all the APIs the component supports.
-    rpc Info(InfoRequest) returns (InfoResponse);
+    rpc Info(InfoReq) returns (InfoResp);
```
bwplotka, line 26: First time I see this practice. While it helps with shorter name it violates coding style guide + consistency. Do we think it's worth it?

pkg/query/storeset.go
bwplotka: is it store at this point? sounds really like just `endpointRef`.

Also the amount of arguments sounds like code smell. I wonder if we hide those behind better abstractions

pkg/query/storeset.go
```diff
@@ -597,6 +630,105 @@ func (s *StoreSet) getActiveStores(ctx context.Context, stores map[string]*store
 			level.Warn(s.logger).Log("msg", "ignored rule store", "address", ruleAddr)
 		}
 	}
+
+	// Gather healthy stores map concurrently using info addresses. Build new store if does not exist already.
+	for _, infoSpec := range s.infoSpec() {
```
bwplotka, line 635: All of this is baked in `storeset` but I would see it really part of endpoints

I would really just copy and add a new endpoint flow - it will be easier to main new thing than to try to understand later what to remove from mixed thing.

Then if someone wants to use endpoints store flag should be empty. WDYT?

bwplotka (commented):
Amazing work everyone! 

I wonder only - what's the deprecation plan and how to not add complexity to storeset if it has to be removed at some point and instead focus on `endpointset` that supports old StoreAPI and new separate info

cmd/thanos/query.go
yeya24: If we only need to keep one of `store` and `endpoint`, what about just reusing `store`?

pkg/info/infopb/rpc.proto
hitanshu-mehta: If I don't rename it then a warning message will also be printed in our documentations as I mentioned [here](https://github.com/thanos-io/thanos/pull/4282#issuecomment-860240272). I was not able to find a way to remove them. @GiedriusS has suggested better names `InfoAPIRequest` and `InfoAPIResponse`.

pkg/query/storeset.go
hitanshu-mehta: Agree! I will refactor this code and maybe create a new flow in a different `endpoint.go` file. WDYT ?

cmd/thanos/query.go
hitanshu-mehta: According to [proposal](https://thanos.io/tip/proposals-accepted/202101-endpoint-discovery.md/), we were planning to remove old flags after some grace period (maybe after 1-2 releases).

pkg/query/storeset.go
bwplotka: SGTM

cmd/thanos/query.go
bwplotka: OK

bwplotka: Can we say about `deprecation` for store flag then?

cmd/thanos/sidecar.go
bwplotka: The thing is we can have kind of constants somewhere in global level. Can we do that? It's still to be imp[roved?

cmd/thanos/store.go
bwplotka: ping

pkg/info/infopb/rpc.proto
bwplotka: This is because our pacakge is `thanos`. Do you mind changing this to `thanos.info`?

bwplotka (commented):
Still minor nits, but looks really close to be done!

yeya24:
Quick question about this and our usecase. We have a central thanos query and multiple edge thanos sidecars. If I use `--endpoint` flag to specify sidecars, then is there a way to disable metadata or it is enabled automatically?
I'd like to have a way to disable metadata queries for performance purposes rather than enabling them automatically.

cmd/thanos/store.go
hitanshu-mehta: Really sorry for delay.
Here is the deprecation plan I had in mind.
1. Remove old flags (i.e. `--store`, `--rule`, `--metadata`, `--exemplar`, `--target`) and all the boilerplate code we have for all commands. ( eg. duplication check, dnsProviders ...)
2. Remove `storeset.go` file ( I am working on to move new endpoint flow in new file as discussed [here](https://github.com/thanos-io/thanos/pull/4282#discussion_r653353874)  ) and all dependencies of this file from the code base.
3.  Remove [Info](https://github.com/thanos-io/thanos/blob/bc8421ac181579f1a8b2bf99183667fba965381e/pkg/store/storepb/rpc.proto#L30) rpc from proto file of Store API and all its corresponding implementations. ( to be specific there 5 different implementations we have of [StoreClient](https://github.com/thanos-io/thanos/blob/bc8421ac181579f1a8b2bf99183667fba965381e/pkg/store/storepb/rpc.pb.go#L692) - `BucketStore`, `MultiTSDBStore`, `PrometheusStore`, `ProxyStore` and `TSDBStore`) 

All these steps will be done in single PR. I think this plan should work . 
Please let me know if im missing something :)

hitanshu-mehta: >  Can we unify implementation somehow at least?

Why do we need to unify all info methods?

cmd/thanos/query.go
hitanshu-mehta: Yes sure

hitanshu-mehta:
@yeya24 Currently there isn't any way to disable specific api queries like (metadata, targets etc).

cmd/thanos/store.go
bwplotka: 1. When we want to change flag we usually deprecate flag first (add in help that it will be removed after next 2 releases). Then add new flags that works simultaneously. Fail if both new and old way is used. Map old flags to new one really before starting the program.
2. Agree
3. Generally makes sense, but we need to make sure that e.g old Querier having a client against old StoreAPI will work with new InfoAPI. Probably we will need to leave old Info Method (add deprecated comment) in old APIs. In the same time the new Querier should work with old APIs.
   * Potential flow: 1. Try (reflect)? InfoAPI end endpoint. If gRPC server does not support it we fallback to store, if store does it work then ruler ..... only for first call / error.
   * e2e test
   * IMPORTANT: This does not mean we need to implement separate info method for every server. We can implement one Info method and use the same one for both InfoAPI and e.g StoreAPI.

matej-g:
@hitanshu-mehta is the item from TO-DO - add Info server to remaining components - planned as a part this PR or will this be done separately?

hitanshu-mehta:
@matej-g It is planned to do separately. I'm planning to add info server in Querier, sidecar and storeGW in this PR.

matej-g:
> @matej-g It is planned to do separately. I'm planning to add info server in Querier, sidecar and storeGW in this PR.

Great thanks :+1: If you'd need someone to take that over afterwards, I'll be happy to offload you.

hitanshu-mehta:
> Great thanks +1 If you'd need someone to take that over afterwards, I'll be happy to offload you.

Sure. Thank you for help :)

hitanshu-mehta:
Yes @bwplotka 🚀

pkg/store/prometheus.go
```diff
@@ -607,3 +607,25 @@ func (p *PrometheusStore) LabelValues(ctx context.Context, r *storepb.LabelValue
...
 	return &storepb.LabelValuesResponse{Values: vals}, nil
 }
+
+func (p *PrometheusStore) LabelSet() []labelpb.ZLabelSet {
+	lset := p.externalLabelsFn()
+
+	labels := make([]labelpb.ZLabel, 0, len(lset))
+	labels = append(labels, labelpb.ZLabelsFromPromLabels(lset)...)
+
+	// Until we deprecate the single labels in the reply, we just duplicate
```
GiedriusS, line 617: I don't think this comment makes sense anymore? :thinking: We should remove this

cmd/thanos/query.go
```diff
@@ -101,21 +101,24 @@ func registerQuery(app *extkingpin.App) {
...
-	ruleEndpoints := cmd.Flag("rule", "Experimental: Addresses of statically configured rules API servers (repeatable). The scheme may be prefixed with 'dns+' or 'dnssrv+' to detect rule API servers through respective DNS lookups.").
+	ruleEndpoints := cmd.Flag("rule", "Deprecation Warning - This flag is deprecated and replaced with `endpoints`. Experimental: Addresses of statically configured rules API servers (repeatable). The scheme may be prefixed with 'dns+' or 'dnssrv+' to detect rule API servers through respective DNS lookups.").
 		Hidden().PlaceHolder("<rule>").Strings()
 
-	metadataEndpoints := cmd.Flag("metadata", "Experimental: Addresses of statically configured metadata API servers (repeatable). The scheme may be prefixed with 'dns+' or 'dnssrv+' to detect metadata API servers through respective DNS lookups.").
+	metadataEndpoints := cmd.Flag("metadata", "Deprecation Warning - This flag is deprecated and replaced with `endpoints`. Experimental: Addresses of statically configured metadata API servers (repeatable). The scheme may be prefixed with 'dns+' or 'dnssrv+' to detect metadata API servers through respective DNS lookups.").
```
GiedriusS, line 114:
```suggestion
	metadataEndpoints := cmd.Flag("metadata", "Deprecation Warning - This flag is deprecated and replaced with `endpoint`. Experimental: Addresses of statically configured metadata API servers (repeatable). The scheme may be prefixed with 'dns+' or 'dnssrv+' to detect metadata API servers through respective DNS lookups.").
```

test/e2e/info_api_test.go
```diff
@@ -0,0 +1,194 @@
...
+		resp, err := http.DefaultClient.Do(req)
+		if err != nil {
+			return err
+		}
+
+		body, err := ioutil.ReadAll(resp.Body)
+		if err != nil {
+			return err
+		}
+
```
GiedriusS, line 154: We should also do `defer runutil.CloseWithErrCapture(&err, resp.Body, "response body close")`

GiedriusS (approved):
Last three suggestions before merging :+1: Also, please, please, please avoid force-pushing and overriding previous commits because when this happens the reviewers need to look through the whole code again to see whether it is still the same. That's why there was this big delay before I got to this again :+1:

hitanshu-mehta:
@GiedriusS Really sorry for the trouble. I will avoid force pushing from now on :)

GiedriusS:
> @GiedriusS Really sorry for the trouble. I will avoid force pushing from now on :)

CI failure seems related (:

GiedriusS (approved):
:muscle: next steps are to actually use the new InfoServer & add tests for it? Plus, we should probably also add `--endpoint-strict`? Finally, remove the old flags? (: And that feature to ignore certain service(-s) would be nice too.

Note to others that `--endpoint` with this PR is semantically equivalent to all of the other flags since we prefill the expected APIs (we don't check the InfoServer, yet) so it is completely safe to merge this!

PR #4292, opened 2021-06-01
receive: Enable exemplars ingestion and querying

### LINKED ISSUE #4235: Receiver and ingest exemplars via remote write

Exemplars via remote write are supported in Prometheus v2.27.0. It would be nice to support ingesting exemplars to Receiver and query exemplars from it.

(Write your answer here.)

Similar to https://github.com/cortexproject/cortex/pull/4124.

I would suppose this can be split into multiple small prs.
1. Support ingesting exemplars via remote write. Exemplars are stored in the same TSDB's exemplar ring buffer. They are forwarded based on the hash of the series labels.
2. Add Receiver Exemplars gRPC API.

(Describe your proposed solution here.)

(Write your answer here.)

### Additional context

(Write your answer here.)

onprem:

## Changes
- Enable exemplar ingestion in Receive
- Add exemplars API implementation for TSDB and MultiTSDB
- Expose exemplars gRPC API from Receive

Fixes #4235 

## Verification
- Local testing, modified some existing tests for ingestion. Need to add E2E tests.

### DIFF
--- a/cmd/thanos/receive.go
+++ b/cmd/thanos/receive.go
@@ -24,6 +24,7 @@ import (
 	"github.com/prometheus/prometheus/tsdb"
 
 	"github.com/thanos-io/thanos/pkg/block/metadata"
+	"github.com/thanos-io/thanos/pkg/exemplars"
 	"github.com/thanos-io/thanos/pkg/extkingpin"
 	"github.com/thanos-io/thanos/pkg/logging"
 
@@ -320,6 +321,7 @@ func setupAndRunGRPCServer(g *run.Group,
 			s = grpcserver.New(logger, &receive.UnRegisterer{Registerer: reg}, tracer, grpcLogOpts, tagOpts, comp, grpcProbe,
 				grpcserver.WithServer(store.RegisterStoreServer(rw)),
 				grpcserver.WithServer(store.RegisterWritableStoreServer(rw)),
+				grpcserver.WithServer(exemplars.RegisterExemplarsServer(exemplars.NewMultiTSDB(dbs.TSDBExemplars))),
 				grpcserver.WithListen(*conf.grpcBindAddr),
 				grpcserver.WithGracePeriod(time.Duration(*conf.grpcGracePeriod)),
 				grpcserver.WithTLSConfig(tlsCfg),
--- a/cmd/thanos/receive.go
+++ b/cmd/thanos/receive.go
@@ -764,7 +764,11 @@ func (rc *receiveConfig) registerFlag(cmd extkingpin.FlagClause) {
 
 	cmd.Flag("tsdb.no-lockfile", "Do not create lockfile in TSDB data directory. In any case, the lockfiles will be deleted on next startup.").Default("false").BoolVar(&rc.noLockFile)
 
-	cmd.Flag("tsdb.max-exemplars", "Enables support for ingesting exemplars and set the maximum number that will be stored. 0 (or less) disables exemplars storage.").Default("0").IntVar(&rc.tsdbMaxExemplars)
+	cmd.Flag("tsdb.max-exemplars",
+		"Enables support for ingesting exemplars and sets the maximum number of exemplars that will be stored per tenant."+
+			" In case the exemplar storage becomes full (number of stored exemplars becomes equal to max-exemplars),"+
+			" ingesting a new exemplar will evict the oldest exemplar from storage. 0 (or less) value of this flag disables exemplars storage.").
+		Default("0").IntVar(&rc.tsdbMaxExemplars)
 
 	cmd.Flag("hash-func", "Specify which hash function to use when calculating the hashes of produced files. If no function has been specified, it does not happen. This permits avoiding downloading some files twice albeit at some performance cost. Possible values are: \"\", \"SHA256\".").
 		Default("").EnumVar(&rc.hashFunc, "SHA256", "")
--- a/cmd/thanos/receive.go
+++ b/cmd/thanos/receive.go
@@ -70,6 +70,7 @@ func registerReceive(app *extkingpin.App) {
 			NoLockfile:             conf.noLockFile,
 			WALCompression:         conf.walCompression,
 			AllowOverlappingBlocks: conf.tsdbAllowOverlappingBlocks,
+			MaxExemplars:           conf.tsdbMaxExemplars,
 		}
 
 		// Enable ingestion if endpoint is specified or if both the hashrings configs are empty.
@@ -687,6 +688,7 @@ type receiveConfig struct {
 	tsdbMinBlockDuration       *model.Duration
 	tsdbMaxBlockDuration       *model.Duration
 	tsdbAllowOverlappingBlocks bool
+	tsdbMaxExemplars           int
 
 	walCompression bool
 	noLockFile     bool
@@ -759,6 +761,8 @@ func (rc *receiveConfig) registerFlag(cmd extkingpin.FlagClause) {
 	cmd.Flag("tsdb.wal-compression", "Compress the tsdb WAL.").Default("true").BoolVar(&rc.walCompression)
 
 	cmd.Flag("tsdb.no-lockfile", "Do not create lockfile in TSDB data directory. In any case, the lockfiles will be deleted on next startup.").Default("false").BoolVar(&rc.noLockFile)
+
+	cmd.Flag("tsdb.max-exemplars", "Enables support for ingesting exemplars and set the maximum number that will be stored. 0 (or less) disables exemplars storage.").Default("0").IntVar(&rc.tsdbMaxExemplars)
 
 	cmd.Flag("hash-func", "Specify which hash function to use when calculating the hashes of produced files. If no function has been specified, it does not happen. This permits avoiding downloading some files twice albeit at some performance cost. Possible values are: \"\", \"SHA256\".").
 		Default("").EnumVar(&rc.hashFunc, "SHA256", "")
--- a/pkg/exemplars/multitsdb.go
+++ b/pkg/exemplars/multitsdb.go
@@ -1,0 +1,35 @@
+// Copyright (c) The Thanos Authors.
+// Licensed under the Apache License 2.0.
+
+package exemplars
+
+import (
+	"fmt"
+
+	"github.com/pkg/errors"
+
+	"github.com/thanos-io/thanos/pkg/exemplars/exemplarspb"
+)
+
+// MultiTSDB implements exemplarspb.ExemplarsServer that allows to fetch exemplars a MultiTSDB instance.
+type MultiTSDB struct {
+	tsdbExemplarsServers func() map[string]exemplarspb.ExemplarsServer
+}
+
+// NewMultiTSDB creates new exemplars.MultiTSDB.
+func NewMultiTSDB(tsdbExemplarsServers func() map[string]exemplarspb.ExemplarsServer) *MultiTSDB {
+	return &MultiTSDB{
+		tsdbExemplarsServers: tsdbExemplarsServers,
+	}
+}
+
+// Exemplars returns all specified exemplars from a MultiTSDB instance.
+func (m *MultiTSDB) Exemplars(r *exemplarspb.ExemplarsRequest, s exemplarspb.Exemplars_ExemplarsServer) error {
+	fmt.Println("meer paas request aayi", m.tsdbExemplarsServers())
+	for tenant, es := range m.tsdbExemplarsServers() {
+		if err := es.Exemplars(r, s); err != nil {
+			return errors.Wrapf(err, "get info for tenant %s", tenant)
+		}
+	}
+	return nil
+}
--- a/pkg/exemplars/multitsdb.go
+++ b/pkg/exemplars/multitsdb.go
@@ -5,6 +5,7 @@ package exemplars
 
 import (
 	"github.com/pkg/errors"
+	"github.com/prometheus/prometheus/promql/parser"
 	"google.golang.org/grpc/codes"
 	"google.golang.org/grpc/status"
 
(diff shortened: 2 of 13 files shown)


pkg/exemplars/multitsdb.go
```diff
@@ -0,0 +1,35 @@
...
+// NewMultiTSDB creates new exemplars.MultiTSDB.
+func NewMultiTSDB(tsdbExemplarsServers func() map[string]exemplarspb.ExemplarsServer) *MultiTSDB {
+	return &MultiTSDB{
+		tsdbExemplarsServers: tsdbExemplarsServers,
+	}
+}
+
+// Exemplars returns all specified exemplars from a MultiTSDB instance.
+func (m *MultiTSDB) Exemplars(r *exemplarspb.ExemplarsRequest, s exemplarspb.Exemplars_ExemplarsServer) error {
+	fmt.Println("meer paas request aayi", m.tsdbExemplarsServers())
```
wiardvanrij, line 28: :D

pkg/exemplars/multitsdb.go
```diff
@@ -0,0 +1,35 @@
...
+	for tenant, es := range m.tsdbExemplarsServers() {
```
wiardvanrij, line 29: Why is this a tenant? Isn't it just an instance that supports tsdb/exemplars?

pkg/receive/writer.go
```diff
@@ -104,6 +135,18 @@ func (r *Writer) Write(ctx context.Context, tenantID string, wreq *prompb.WriteR
 		level.Warn(r.logger).Log("msg", "Error on ingesting samples that are too old or are too far into the future", "numDropped", numOutOfBounds)
 		errs.Add(errors.Wrapf(storage.ErrOutOfBounds, "add %d samples", numOutOfBounds))
 	}
+	if numExemplarsOutOfOrder > 0 {
+		level.Warn(r.logger).Log("msg", "Error on ingesting out-of-order exemplars", "numDropped", numExemplarsOutOfOrder)
```
wiardvanrij, line 139: It's a level warn, but the logger msg says it's an error. I would make the message more generic and let log level handle the prefix, like: "warn: exemplars out of order" or something like that.

pkg/receive/writer.go
```diff
@@ -90,6 +94,33 @@ func (r *Writer) Write(ctx context.Context, tenantID string, wreq *prompb.WriteR
...
+					Ts:     ex.Timestamp,
+					HasTs:  true,
+				})
+				switch err {
+				case storage.ErrOutOfOrderExemplar:
+					numExemplarsOutOfOrder++
+					level.Debug(logger).Log("msg", "Out of order exemplar")
+				case storage.ErrDuplicateExemplar:
+					numExemplarsDuplicate++
+					level.Debug(logger).Log("msg", "Out of order exemplar")
```
wiardvanrij, line 117: out of order -> duplicate

pkg/receive/writer.go
```diff
@@ -90,6 +94,33 @@ func (r *Writer) Write(ctx context.Context, tenantID string, wreq *prompb.WriteR
...
+				case storage.ErrExemplarLabelLength:
+					numExemplarsLabelLength++
+					level.Debug(logger).Log("msg", "Label length for exemplar exceeds max limit")
```
wiardvanrij, line 120: Possible to add length and the limit values? Provides value :)

cmd/thanos/receive.go
```diff
@@ -760,6 +764,8 @@ func (rc *receiveConfig) registerFlag(cmd extkingpin.FlagClause) {
 
 	cmd.Flag("tsdb.no-lockfile", "Do not create lockfile in TSDB data directory. In any case, the lockfiles will be deleted on next startup.").Default("false").BoolVar(&rc.noLockFile)
 
+	cmd.Flag("tsdb.max-exemplars", "Enables support for ingesting exemplars and set the maximum number that will be stored. 0 (or less) disables exemplars storage.").Default("0").IntVar(&rc.tsdbMaxExemplars)
```
wiardvanrij, line 767: What does it mean with number?  Could you explain what the number actually is (i.e. the amount of examplars stored?) and what it does (will it drop things?) - IMO would be huge bonus to provide extra documentation to explain users how they should set this value and how they could calculate this.

wiardvanrij (commented):
I've done a review but it's not on the technical part but more 'usability' part :) Really looking forward to this feature 👍

pkg/exemplars/multitsdb.go
onprem: Yep, I know it's here haha. I'll fix this and push XD

cmd/thanos/receive.go
onprem: Yep, it is number of exemplars stored, and yes, it'll drop the oldest exemplars (AFAIR, I have to check on the Prometheus TSDB side).

> IMO would be huge bonus to provide extra documentation to explain users how they should set this value and how they could calculate this.

Yep, I agree. Maybe we can add a page around operating Receive (or whole Thanos stack in general) with exemplars.

pkg/exemplars/multitsdb.go
onprem: Well yeah, but it's an instance for a particular tenant. Maybe a better name is `tenantExemplarServer` or simply `exemplarServer`?

pkg/receive/writer.go
onprem: I was following what we do for samples. I'd like other's opinion here as well to either switch to error, put warn there instead of error, or keep it as it.

pkg/receive/writer.go
onprem: I have added limit but adding the current length is a bit tricky (we have loop over all the label names and values) so I am not sure if the overhead is worth it.

pkg/exemplars/multitsdb.go
yeya24: I have a similar question. How would an exemplar query perform for a single tenant? Looks like each exemplar query request would check all tenants' exemplar storage. It is a little bit expensive.

pkg/exemplars/tsdb.go
```diff
@@ -0,0 +1,62 @@
...
+	if err != nil {
+		return err
+	}
+
+	matchers := parser.ExtractSelectors(expr)
+
+	exemplars, err := eq.Select(r.Start, r.End, matchers...)
+	if err != nil {
+		return err
+	}
```
yeya24, line 46: Line 36 - line 46 is the same for all tenants so maybe we can extract this part out the function?

pkg/receive/handler.go
```diff
@@ -325,7 +325,7 @@ func (h *Handler) receiveHTTP(w http.ResponseWriter, r *http.Request) {
 		tenant = h.options.DefaultTenantID
 	}
 
-	// TODO(yeya24): handle remote write metadata and exemplars.
+	// TODO(yeya24): handle remote write metadata.
 	// exit early if the request contained no data
 	if len(wreq.Timeseries) == 0 {
```
yeya24, line 330: This should be updated as we need to process exemplars only requests.

yeya24 (commented):
I don't see the code of forwarding exemplars. I assume only timeserieses are forwarded. Do you want to include the exemplars part in this pr?

pkg/receive/handler.go
onprem: This is working as is, as exemplars are part of the `Timeseries` array. This is why we don't need any modification on the forwarding part as well.

onprem: You can take a look at the proto definition here: https://github.com/thanos-io/thanos/blob/main/pkg/store/storepb/prompb/types.proto#L66-L72

Since we batch and forward the Timeseries elements, the exemplars get forwarded automatically.

yeya24: Got it. If it is part of the `Timeseries` array then it is good here.

I am not sure about the forwarding part. How does the hashing part work https://github.com/thanos-io/thanos/blob/main/pkg/receive/handler.go#L388?

https://github.com/thanos-io/thanos/blob/main/pkg/receive/hashring.go#L65 Hashring.GetN calculates the hash based on `ts.Labels`. If it is an exemplar series, what will be the value of `ts.Labels`?  Is it the series labels of exemplar?

yeya24: Yes, that makes sense. If it is part of `timeseries` then all good. Thanks.

pkg/exemplars/multitsdb.go
onprem: Ah, there are some differences here. With a series call, you simply get `[]storepb.LabelMatcher` / `[]*labels.Matcher`, while with exemplars, you get `[][]*labels.Matcher`, where you have a union between the outer slices and then similar to Series call, intersection between the inner `[]*labels.Matcher`. WDYT should be the correct approach here?

I was thinking of looping over the outer array, and filter out `[]*labels.Matcher`s that doesn't match with the external labels.

onprem: I have pushed a commit with this approach, LMK if this isn't what you meant.

pkg/exemplars/tsdb.go
onprem: The code changed a bit around this part so resolving this for now.

pkg/exemplars/tsdb.go
```diff
@@ -0,0 +1,89 @@
...
+}
+
+// Exemplars returns all specified exemplars from a TSDB instance.
+func (t *TSDB) Exemplars(r *exemplarspb.ExemplarsRequest, s exemplarspb.Exemplars_ExemplarsServer) error {
+	expr, err := parser.ParseExpr(r.Query)
+	if err != nil {
+		return status.Error(codes.Internal, err.Error())
+	}
+
+	match, selectors := selectorsMatchesExternalLabels(parser.ExtractSelectors(expr), t.extLabels)
```
yeya24, line 39: Again the query parsing is still duplicated. Why not just use the parsed matchers in the method signature?

``` go
func (t *TSDB) Exemplars(matchers [][]*labels.Matcher, s exemplarspb.Exemplars_ExemplarsServer) error { ... }
```

pkg/exemplars/tsdb.go
yeya24: not resolved I think.

cmd/thanos/receive.go
onprem: For now, I tried to make the flag description more descriptive.

pkg/receive/multitsdb.go
```diff
@@ -432,6 +464,14 @@ func (a adapter) Querier(ctx context.Context, mint, maxt int64) (storage.Querier
 	return q, nil
 }
 
+func (a adapter) ExemplarQuerier(ctx context.Context) (storage.ExemplarQuerier, error) {
+	q, err := a.db.ExemplarQuerier(ctx)
```
yeya24, line 468: Can be simplified to `return a.db.ExemplarQuerier(ctx)`

pkg/exemplars/tsdb.go
onprem: Now I got what you meant, we are doing query parsing for each tenant, yeah that's not good.

But we can't change the signature I guess, If we change the signature, `*TSDB` will not implement `exemplarspb.ExemplarsServer` interface.

onprem: One option is to put `[][]*label.Matchers` (or labelpb equivalent) in the `ExemplarsRequest` instead of the raw query. For sidecars, maybe we can reconstruct the query using the matches but I am not sure about this part.

pkg/exemplars/tsdb.go
onprem: Ah, I misunderstood what you were trying to say. I got it now, let's continue at https://github.com/thanos-io/thanos/pull/4292#discussion_r646047501

pkg/exemplars/tsdb.go
```diff
@@ -0,0 +1,89 @@
...
+
+	if !match {
+		return nil
+	}
+
+	if len(selectors) == 0 {
+		return status.Error(codes.InvalidArgument, errors.New("no matchers specified (excluding external labels)").Error())
+	}
```
yeya24, line 47: Line 41 ~ 47 looks very nice! Would you mind also updating https://github.com/thanos-io/thanos/blob/main/pkg/exemplars/proxy.go#L67-L76? 
It is good to return an error if no other matcher specified instead of returning nil.

pkg/exemplars/tsdb.go
yeya24: I am wondering why we need to implement this interface for `TSDB`. It is already implemented in `MultiTSDB`, right? For each tenant's TSDB, as it is part of the multitsdb. Can we access it by a separate function?

onprem: Actually, I don't think we have any hard requirement to implement `exemplarspb.ExemplarsServer` for `exemplars.TSDB`, let's do what you suggested.

yeya24: Yes, that would be better. We can change the part https://github.com/thanos-io/thanos/pull/4292/files#diff-970995d3ac812ca4b94e73daf4773bd1a3f80cd746ef697e2084a29a4802333aR27-R34 and use a new method for TSDB exemplars.

yeya24 (commented):
Really awesome work! Thanks for your patience.

pkg/exemplars/tsdb.go
onprem: Done, pushed a commit with this (https://github.com/thanos-io/thanos/pull/4292/commits/f6a00371c80784d9a479120b56cd8409bddc40f8)

pkg/exemplars/tsdb.go
onprem: Done

yeya24 (dismissed):
Nice! LGTM

pkg/receive/writer.go
bwplotka: I think we are not too strict here, so it's fine to say error if it's only a warning about the not serious error.

kakkoyun: If we handle the returned error on upper levels (this could be logging as well), we don't need to log/handle it twice. I see that you've just followed the existing structure so it seems ok for now.

kakkoyun (approved):
LGTM 🚀

pkg/receive/writer.go
bwplotka: hm.. what I don't like is duplicate error handling. We both add error and warn error... Wonder if we could just handle that on caller side instead. 🤔

bwplotka (approved):
This is epic!!! LGTM 💪🏽 

From when @onprem you are full-pledge backend engineer with so much skill for high quality work like this? 😱

kakkoyun:
Let's rebase and merge this

pkg/receive/writer.go
onprem: Yeah, this is a bit icky but as Kemal noted I was just following the existing structure we use for sample. I think I can try refactoring this in a follow up PR if that's ok.

bwplotka (approved):
Solid work 😱 


LGTM!

Issue #4130, opened 2021-04-30
Add tracing for federated exemplar API


yeya24:
### Describe the solution you'd like

Instrument tracing spans for the federated APIs. This should include the proxy (querier) side as well as the leaf store sides.

Hangzhi:
May I work on this?

yeya24:
Sure, feel free to take it. @Hangzhi

ianbillett:
Hey @Hangzhi :wave: - are you working on this? I can't see any active branches or forks. If not - I'm happy to take this on.

Hangzhi:
@ianbillett Please take this on! Sorry for leaving this inactive.

PR #4238, opened 2021-05-15
added tracing to matadata api

KrishnaSindhur:
Signed-off-by: KrishnaSindhur <krishnasindhur@gmail.com>

* [ ] I added CHANGELOG entry for this change.
* [ ] Change is not relevant to the end user.

## Changes

Add tracing for federated metadata API.

Closes #4132 

## Verification

- make build
- make test-local

PR #4178, opened 2021-05-05
Add tracing spans for rules API

toshi1127:
Signed-off-by: Matsumoto <toshi.matsumoto.3n@gmail.com>

* [ ] I added CHANGELOG entry for this change.
* [x] Change is not relevant to the end user.

## Changes

Add tracing for federated rules API.

Closes #4131 

## Verification

- `make build`
- `make test-local`

PR #4299, opened 2021-06-03
Add tracing to exemplar APIs

ianbillett:

## Changes

Follows https://github.com/thanos-io/thanos/pull/4238 & https://github.com/thanos-io/thanos/pull/4178 - adds tracing instrumentation to the exemplar endpoints.

Closes https://github.com/thanos-io/thanos/issues/4130

## Verification
Standard tests pass:
```
make build
make test-local
```

I have been trying and failing to figure out an additional unit test to verify this behaviour. 

It might be possible to setup a grpc server with a `basictracer-go` and register the exemplar APIs then verify spans are created when the API is called. Previous tracing PRs have not included this so might not be worth the effort :shrug:

### DIFF
--- a/pkg/api/query/v1.go
+++ b/pkg/api/query/v1.go
@@ -795,9 +795,9 @@ func NewExemplarsHandler(client exemplars.UnaryClient, enablePartialResponse boo
 		defer span.Finish()
 
 		var (
-			exemplarsData    []*exemplarspb.ExemplarData
-			exemplarWarnings storage.Warnings
-			exemplarError    error
+			data     []*exemplarspb.ExemplarData
+			warnings storage.Warnings
+			err      error
 		)
 
 		start, err := cortexutil.ParseTime(r.FormValue("start"))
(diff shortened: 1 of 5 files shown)


pkg/api/query/v1.go
```diff
@@ -806,11 +815,15 @@ func NewExemplarsHandler(client exemplars.UnaryClient, enablePartialResponse boo
 			Query:                   r.FormValue("query"),
 			PartialResponseStrategy: ps,
 		}
-		exemplarsData, warnings, err := client.Exemplars(r.Context(), req)
-		if err != nil {
-			return nil, nil, &api.ApiError{Typ: api.ErrorInternal, Err: errors.Wrap(err, "retrieving exemplars")}
+
+		tracing.DoInSpan(ctx, "retrieve_exemplars", func(ctx context.Context) {
+			exemplarsData, exemplarWarnings, exemplarError = client.Exemplars(r.Context(), req)
```
kakkoyun, line 820:
```suggestion
			exemplarsData, exemplarWarnings, exemplarError = client.Exemplars(ctx, req)
```
To propagate correct context, we should use the one in closure.

pkg/api/query/v1.go
```diff
@@ -791,6 +791,15 @@ func NewExemplarsHandler(client exemplars.UnaryClient, enablePartialResponse boo
 	}
 
 	return func(r *http.Request) (interface{}, []error, *api.ApiError) {
+		span, ctx := tracing.StartSpan(r.Context(), "exemplar_query_request")
+		defer span.Finish()
+
+		var (
+			exemplarsData    []*exemplarspb.ExemplarData
```
kakkoyun, line 798: We can be more concise here and just go with data, warnings and err.

kakkoyun (commented):
Looks great :) Only minor issues.

I have added a couple of comments inline.

pkg/api/query/v1.go
ianbillett: good catch :+1:

ianbillett:
Think this puppy is ready to merge :muscle:

kakkoyun (approved):
LGTM :rocket:

squat:
nice :))

PR #4316, opened 2021-06-07
receiver: Added Docs for Receiver Split behaviour

yashrsharma44:
Added some section about receiver's [dual behaviour](https://github.com/thanos-io/thanos/pull/4231), which implements the [Split Proposal](https://github.com/thanos-io/thanos/blob/main/docs/proposals/202012_receive_split.md)

Also added a Changelog for the PR #4231 because it was not added 😬 .


docs/components/receive.md
```diff
@@ -66,6 +74,13 @@ The example content of `hashring.json`:
 With such configuration any receive listens for remote write on `<ip>10908/api/v1/receive` and will forward to correct one in hashring if needed
 for tenancy and replication.
 
+### Enabling Routing/Ingestion Mode for Thanos Receiver
```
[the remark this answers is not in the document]
yashrsharma44: Sure, let me remove this.

docs/components/receive.md
```diff
@@ -15,6 +15,14 @@ Thanos Receive supports multi-tenancy by using labels. See [Multitenancy documen
 For more information please check out [initial design proposal](../proposals/201812_thanos-remote-receive.md).
 For further information on tuning Prometheus Remote Write [see remote write tuning document](https://prometheus.io/docs/practices/remote_write/).
 
+Currently in version [v0.21.2](https://github.com/thanos-io/thanos/blob/main/CHANGELOG.md#v0211---20210604), receiver implements the split behaviour of solo ingestion/routing (see the [proposal](https://github.com/thanos-io/thanos/blob/main/docs/proposals/202012_receive_split.md) for more details). This means, we can enable receiver to ingest data only, or re-route requests only.
```
squat, line 18: I think that the word "currently" will be confusing once v0.22 is released
```suggestion
As of version [v0.21.2](https://github.com/thanos-io/thanos/blob/main/CHANGELOG.md#v0211---20210604), the Thanos receiver implements the [receiver split proposal](https://github.com/thanos-io/thanos/blob/main/docs/proposals/202012_receive_split.md), which allows data ingestion as well as request routing to be independently enabled/disabled.
```

docs/components/receive.md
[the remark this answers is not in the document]
yashrsharma44: done

docs/components/receive.md
```diff
@@ -15,6 +15,14 @@ Thanos Receive supports multi-tenancy by using labels. See [Multitenancy documen
...
+
+This helps receiver relieve of additional responsibility, and can focus on solo behaviour of ingestion/routing. This would enable users to prepare a topology of receivers, containing trees of receiver with depth **N**.
```
squat, line 20: This paragraph starts with `This`, however, it is a new paragraph so it's far away from whatever `This` _is_ and it's not clear what it is referring to. Is `This` the proposal? The Receiver? The implementation of the proposal?

docs/components/receive.md
```diff
@@ -66,6 +74,13 @@ The example content of `hashring.json`:
 With such configuration any receive listens for remote write on `<ip>10908/api/v1/receive` and will forward to correct one in hashring if needed
 for tenancy and replication.
 
+### Enabling Routing/Ingestion for Thanos Receiver
```
squat, line 77: I think that `/` can be confusing because it can imply that the things on each side of the slash are the same, which is not the case here.
```suggestion
### Enabling Routing and Ingestion for the Thanos Receiver
```

docs/components/receive.md
```diff
@@ -66,6 +74,13 @@ The example content of `hashring.json`:
...
+
+We have not added yet another flag for controlling this behaviour. We have relied on the below convention to enable ingestion/routing -
```
squat, line 79: `yet` makes me think that we have plans to do it, which we haven't discussed.
```suggestion
We have not added another flag for controlling this behaviour. Instead, we have relied on the existing flags and the below convention to enable and disable the ingestion and routing functionalities:
```

docs/components/receive.md
```diff
@@ -66,6 +74,13 @@ The example content of `hashring.json`:
...
+* **Ingestion and Distribution** - Make sure `--receive.local-endpoint` and `--receive.hashrings` / `--receive.hashrings-file` are provided.
```
squat, line 80:
```suggestion
* **Ingestion and Distribution** - Make sure that `--receive.local-endpoint` and either `--receive.hashrings` or `--receive.hashrings-file` are provided.
```

docs/components/receive.md
```diff
@@ -66,6 +74,13 @@ The example content of `hashring.json`:
...
+* **Ingestion** - Since routing behaviour is not needed, just provide `--receive.local-endpoint` only.
```
squat, line 81: Let's avoid words like "just", "simply," and "obviously" in general since they can imply that things are easy, which may not be the case for every user. In this case we are using "just" as a synonym for "only" but we we already say "only" at the end, so we don't need to say it twice IMO.
```suggestion
* **Ingestion** - Since routing behaviour is not needed, provide `--receive.local-endpoint` only.
```

docs/components/receive.md
```diff
@@ -66,6 +74,13 @@ The example content of `hashring.json`:
...
+* **Distribution** - For routing functionality, just provide one of the hashring flags - `--receive.hashrings` / `--receive.hashrings-file`. Don't provide local endpoint, as it would enable ingestion also.
```
squat, line 82:
```suggestion
* **Distribution** - For routing functionality, provide one of the hashring configuration flags: either `--receive.hashrings` or `--receive.hashrings-file`. Do not specify a local endpoint, as doing so would enable ingestion.
```

squat: One last thing comes to mind: we sometimes call it "distribution" and other times call it "request routing," or only "routing." For clarity, we should stick to one term and use it consistently.

yashrsharma44: Lets use distribution throughout, sounds good to me.

docs/components/receive.md
yashrsharma44: I have replaced `this` with `implementation of receiver split proposal`. Let me know if this works :D

docs/components/receive.md
```diff
@@ -15,6 +15,14 @@ Thanos Receive supports multi-tenancy by using labels. See [Multitenancy documen
 For more information please check out [initial design proposal](../proposals/201812_thanos-remote-receive.md).
 For further information on tuning Prometheus Remote Write [see remote write tuning document](https://prometheus.io/docs/practices/remote_write/).
 
+As of version [v0.21.2](https://github.com/thanos-io/thanos/blob/main/CHANGELOG.md#v0211---20210604), the Thanos receiver implements the [receiver split proposal](https://github.com/thanos-io/thanos/blob/main/docs/proposals/202012_receive_split.md), which allows data ingestion as well as request routing to be independently enabled/disabled.
```
onprem, line 18: Do you mean `v0.22.0`, as that is going to be the first version in which this work will get released?

docs/components/receive.md
```diff
@@ -15,6 +15,14 @@ Thanos Receive supports multi-tenancy by using labels. See [Multitenancy documen
...
+
+The current implementation of [receiver split proposal](https://github.com/thanos-io/thanos/blob/main/docs/proposals/202012_receive_split.md) helps receiver relieve of additional responsibility, and can focus on solo behaviour of ingestion/routing. This would enable users to prepare a topology of receivers, containing trees of receiver with depth **N**.
+
+![](https://docs.google.com/drawings/d/e/2PACX-1vQaLa9EdF_frGmE6zbK48Zj9a8lIKxdx8NpOCU0eFizGCALRY8uUzZfFJLH8VNvtjyi-YBmVHq6PR8A/pub?w=1442&h=563)
```
onprem, line 22:
```suggestion
![Example Receive architecture diagram](https://docs.google.com/drawings/d/e/2PACX-1vQaLa9EdF_frGmE6zbK48Zj9a8lIKxdx8NpOCU0eFizGCALRY8uUzZfFJLH8VNvtjyi-YBmVHq6PR8A/pub?w=1442&h=563)
```

docs/components/receive.md
```diff
@@ -15,6 +15,14 @@ Thanos Receive supports multi-tenancy by using labels. See [Multitenancy documen
...
+
+The current behaviour for receiver, however has not been modified, and works as expected. So users who are using receiver for both routing and ingestion, should not notice any changes as such. The change is backwards compatible. However, it is suggested that the new behaviour should be used, due to benchmarking, which shows drastic improvement of receiver.
```
onprem, line 24: > which shows drastic improvement of receiver.

Can we talk a bit about what improvements exactly? This statement in it's current state doesn't allow me to understand if the improvements are big enough to opt-into this new way of running receiver.

onprem (commented):
Looking good! Just a few small comments.

docs/components/receive.md
yashrsharma44: Sure, does not make sense to release this feature as part of `v0.21.*`

docs/components/receive.md
yashrsharma44: Lets remove the last line, I would add it, once I have some metrics to show.

docs/components/receive.md
```diff
@@ -15,6 +15,14 @@ Thanos Receive supports multi-tenancy by using labels. See [Multitenancy documen
 For more information please check out [initial design proposal](../proposals/201812_thanos-remote-receive.md).
 For further information on tuning Prometheus Remote Write [see remote write tuning document](https://prometheus.io/docs/practices/remote_write/).
 
+As of version [v0.22.0](https://github.com/thanos-io/thanos/blob/main/CHANGELOG.md#added), the Thanos receiver implements the [receiver split proposal](https://github.com/thanos-io/thanos/blob/main/docs/proposals/202012_receive_split.md), which allows data ingestion as well as request routing to be independently enabled/disabled.
+
```
ianbillett, line 19: > This would enable users to prepare a topology of receivers, containing trees of receiver with depth **N**.

Why would users want to do that? The documentation is the place that users will look for guidance on this topic.

One thing that comes to mind is that this lets the users achieve hard tenancy of receivers - no two customers will share receiver ingest instances.

ianbillett: The receiver split [proposal](https://github.com/thanos-io/thanos/blob/main/docs/proposals/202012_receive_split.md) does a good job of explaining why.

With the changes coming in in `v0.22.0`, only routing-receivers need to watch for configuration changes, and ingesting-receivers do not need to. So when the configuration is updated, the routing-receivers will almost instantaneously re-configure without a lengthy WAL flush (& downtime) like we have in the current setup (routing & ingesting receivers).

yashrsharma44: Shall I add a line at the end -

> If you are curious about why you should setup your receiver topology, please refer the proposal for more details.

yashrsharma44: > With the changes coming in in v0.22.0, only routing-receivers need to watch for configuration changes, and ingesting-receivers do not need to. So when the configuration is updated, the routing-receivers will almost instantaneously re-configure without a lengthy WAL flush (& downtime) like we have in the current setup (routing & ingesting receivers).

Also, I think we can add your description in the doc. The above statement sounds descriptive about the use case. cc @squat WDYT?

bwplotka: I think we need to use full sentences, otherwise sgtm (:

bwplotka (approved):
Thanks 💪🏽 Super useful! Some minor comments only!

docs/components/receive.md
yashrsharma44: > I think we need to use full sentences, otherwise sgtm (:

I didn't get which sentence we should complete, can you please explain 😅

docs/components/receive.md
```diff
@@ -66,6 +74,13 @@ The example content of `hashring.json`:
 With such configuration any receive listens for remote write on `<ip>10908/api/v1/receive` and will forward to correct one in hashring if needed
 for tenancy and replication.
 
+### Enabling Routing and Ingestion for the Thanos Receiver
+
+We have not added another flag for controlling this behaviour. Instead, we have relied on the existing flags and the below convention to enable and disable the ingestion and routing functionalities:
+* **Ingestion and Routing** - Make sure that `--receive.local-endpoint` and either `--receive.hashrings` or `--receive.hashrings-file` are provided.
+* **Ingestion** - Since routing behaviour is not needed, provide `--receive.local-endpoint` only.
```
ianbillett, line 81: Since `--receive.local-endpoint` is only used to figure out if the current instance is in the specified hashring, it really doesn't make sense to the user to specify `--receive.local-endpoint` without also `--receive.hashring` too.

In my opinion it is much clearer for the user if we advise them to enable ingestion mode without specifying any extra parameters.

Also - my preferences would re-arrange these bullets so that number of extra parameters are increasing as the list goes down:
* Ingestion - no parameters.
* Routing - hashring parameters
* Ingestion & Routing - hashrigh parameters & local endpoint parameter.

yashrsharma44: So do you propose to have a default value for `--receive.local-endpoint`, as it was earlier - [link](https://github.com/thanos-io/thanos/blob/release-0.16/cmd/thanos/receive.go#L122)?

Issue #4320, opened 2021-06-09
query-frontend logs failed to parse host port for http log decision


yasharne:
**Thanos, Prometheus and Golang version used**:
```
Version	0.21.1
Revision	3558f4a5adc8d4068f82528570a439d440f98b9d
Branch	HEAD
BuildUser	circleci@eb60c35e294a
BuildDate	20210604-12:24:45
GoVersion	go1.16.4
```

**What happened**:
I have deployed the `query-frontend` to the Kubernetes and there are a lot of logs like this in the container:
```
level=error ts=2021-06-09T04:29:40.506819592Z caller=http.go:48 protocol=http http.component=server msg="failed to parse host port for http log decision" err="address thanos.apps.x.y.z: missing port in address"
```

Here are the command-line flags for the `query-frontend`:
```
            - --query-frontend.compress-responses
            - --query-frontend.downstream-url=http://thanos-query-http.thanos.svc:10902
            - --query-frontend.log-queries-longer-than=5s
            - --query-range.split-interval=24h
            - --query-range.response-cache-max-freshness=1m
            - --query-range.max-retries-per-request=5
            - --query-range.response-cache-config-file=/etc/thanos/frontend.yml
            - --labels.max-retries-per-request=10
            - --labels.split-interval=12h
            - --cache-compression-type=snappy
 ```

**What you expected to happen**:
I haven't defined any address for `query-frontend` in the command-line arguments, if the logger is fetching the address from the user's request why it cannot determine the port address?

**How to reproduce it (as minimally and precisely as possible)**:

**Full logs to relevant components**:
```
level=error ts=2021-06-09T04:29:40.506819592Z caller=http.go:48 protocol=http http.component=server msg="failed to parse host port for http log decision" err="address thanos.apps.x.y.z: missing port in address"
```

**Anything else we need to know**:

yeya24:
I feel this issue is related to the logging middleware. CC @yashrsharma44.

The error happens at https://github.com/thanos-io/thanos/blob/main/pkg/logging/http.go#L43-L51. If there is no port in the address, can we do something to skip this error?

MrYueQ:
@yeya24  Hi. I also encountered this error log, although it does not affect the use of the service. I want to introduce the situation on my side. 
This is an example of my current architecture

![image](https://user-images.githubusercontent.com/14846522/121736376-136e6900-cb2a-11eb-987a-66fb7c12f585.png)

This is my version information and error log
```yaml
thanos --version
[...]

PR #4348, opened 2021-06-16
Fix for parsing port at log middleware

spaparaju:
This PR fixes issue#: [4320](https://github.com/thanos-io/thanos/issues/4320)

## Changes

Added check for `:` prior to [splitting](https://cs.opensource.google/go/go/+/refs/tags/go1.16.5:src/net/ipsock.go;l=169) the port from the hostURL.

## Verification

Tested on standalone, Kube. environments and did not observe parsing errors in the log files.


yeya24 (commented):
Thanks for the fix.
This pr worth a unit test. Would you mind adding one?

pkg/logging/http.go
```diff
@@ -43,14 +44,24 @@ func (m *HTTPServerMiddleware) HTTPMiddleware(name string, next http.Handler) ht
...
+			_, port, err = net.SplitHostPort(hostPort)
+			if err != nil {
+				level.Error(m.logger).Log("msg", "failed to parse host port for http log decision", "err", err)
+				next.ServeHTTP(w, r)
+				return
+			}
 		}
 
-		decision := m.opts.shouldLog(net.JoinHostPort(r.URL.String(), port), nil)
+		var deciderUrl string = r.URL.String()
```
squat, line 60: Nit, we normally use the declare-and-assign operator in go whenever possible
```suggestion
		deciderUrl := r.URL.String()
```

pkg/logging/http.go
```diff
@@ -43,14 +44,24 @@ func (m *HTTPServerMiddleware) HTTPMiddleware(name string, next http.Handler) ht
...
+		if len(port) > 0 {
+			deciderUrl = net.JoinHostPort(r.URL.String(), port)
```
squat, line 62: Nit, we can save this call if we reuse the variable
```suggestion
			deciderUrl = net.JoinHostPort(deciderUrl, port)
```

pkg/logging/http.go
squat: For consistency with the style guide, we normally capitalize all of the letters in abbreviations:
```suggestion
		var deciderURL string = r.URL.String()
```

squat (commented):
Looks good SriKrishna :) i just have small style suggestions

bwplotka:
Nice!

Can we add more clear PR title? It does not tell much what exactly was changed to fix it.

spaparaju:
> Thanks for the fix.
> This pr worth a unit test. Would you mind adding one?

Added a unit test to test with an URL that does not explicitly contain a port number.

GiedriusS:
@spaparaju please fix the DCO:
```
Commit sha: 073971d, Author: Srikrishna Paparaju, Committer: Srikrishna Paparaju; Expected "Srikrishna Paparaju paparaju@gmail.com", but got "Srikrishna Paparaju srikrishna@spaparaj-mac.myfiosgateway.com".

```
https://github.com/thanos-io/thanos/pull/4348/checks?check_run_id=2861409152

yeya24 (approved):
LGTM

ikonyeama:
Still seeing these errors in the logs after upgrading Thanos from 0.21.1 to 0.24.0 
level=error ts=2022-01-08T07:52:08.191747835Z caller=http.go:48 component=rules protocol=http http.component=server msg="failed to parse host port for http log decision" err="address xyz.com: missing port in address"

Issue #4278, opened 2021-05-27
tools bucket replicate: concurrent replication


GiedriusS:
### Is your proposal related to a problem?

Replication is currently slow.

### Describe the solution you'd like

`tools bucket replicate` could upload a few objects at a time instead of one by one

### Describe alternatives you've considered

Copying the objects with some other tool but `tools bucket replicate` is way too convenient because you can use two completely different object storage providers thus there are no alternatives to this, IMHO.

### Additional context

Shouldn't be too hard to implement :thinking:

soniasingla:
@Itaachime feel free to raise PR whenever ready or if you have questions, ask them here/slack!! 🤗

Hangzhi:
Hello @Itaachime, are you working on it. If not, I would love to take this 😃.

Hangzhi:
Hello @GiedriusS. I think I should change some code in this https://github.com/thanos-io/thanos/blob/ebbd68df68902db404d89d6bd8b3f88d6193d851/pkg/replicate/replicator.go. I'm still not very familiar with Thanos. Could you please offer some hints of where to upload a few objects at a time? 😃

vaibhavdugar:
Hey! I am a newbie, how can I contribute in this project? And what are the pre-requisites to begin with this? What are the technologies which I need to learn to contribute to CNCF projects? Can someone help me out please.

peilii:
Hi @GiedriusS,

I would like to work on this issue. Would you please guide me through it?

eshandhawan51:
@Hangzhi Is there any update on this ??

Subhasish-Behera:
hello!i am new here.Can i work on this issue?

yeya24:
@Subhasish-Behera sure go ahead.

Subhasish-Behera:
@yeya24 any design doc that I can refer to build my understanding of codeflow for contributing.

yeya24:
@Subhasish-Behera Unfortuantely we don't have any design doc for this component but the flow should be straightforward.
You could refer to the PRs linked with this issue to see how this could be implemented.

hackeramitkumar:
@yeya24 , can I work on it???

Kartik-Garg:
@GiedriusS @yeya24 I have raised a [PR](https://github.com/thanos-io/thanos/pull/6174) against this issue :tada:

PR #4350, opened 2021-06-17
block replication concurrently

Hangzhi:
Signed-off-by: Hangzhi <1454678938@qq.com>

This PR fixes issue #4278

## Changes

Make block replication run concurrently with goroutine.

### DIFF
--- a/pkg/replicate/scheme.go
+++ b/pkg/replicate/scheme.go
@@ -11,6 +11,7 @@ import (
 	"io/ioutil"
 	"path"
 	"sort"
+	"sync"
 
 	"github.com/go-kit/kit/log"
 	"github.com/go-kit/kit/log/level"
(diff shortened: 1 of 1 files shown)


pkg/replicate/scheme.go
```diff
@@ -203,9 +204,35 @@ func (rs *replicationScheme) execute(ctx context.Context) error {
...
+	errChan := make(chan error)
+	finishChan := make(chan struct{})
+
 	for _, b := range availableBlocks {
-		if err := rs.ensureBlockIsReplicated(ctx, b.BlockMeta.ULID); err != nil {
-			return errors.Wrapf(err, "ensure block %v is replicated", b.BlockMeta.ULID.String())
+		wg.Add(1)
+		go func(b *metadata.Meta) {
+			defer wg.Done()
+			if err := rs.ensureBlockIsReplicated(ctx, b.BlockMeta.ULID); err != nil {
```
bwplotka, line 216: * Set number of workers and allow the user to configure it (bounded concurrency).
* I think we need another mode that group those by external labels and parallelize only between external labels? under flag (full concurrency vs external labels grouping)

pkg/replicate/scheme.go
```diff
@@ -203,9 +204,35 @@ func (rs *replicationScheme) execute(ctx context.Context) error {
 		return availableBlocks[i].BlockMeta.MinTime < availableBlocks[j].BlockMeta.MinTime
 	})
 
+	// Replicate concurrently
```
bwplotka, line 207:
```suggestion
	// Replicate concurrently.
```

https://thanos.io/tip/contributing/coding-style-guide.md/#comments-should-be-full-sentences

bwplotka (commented):
Nice start! Generally good just some comments.

pkg/replicate/scheme.go
Hangzhi: Thanks for the coemmnt. Working on these.

Hangzhi: Hello @bwplotka
What do the external labels mean here?  What level should we group by the blocks ？Should we group the blocks by any distinct label (k-v pairs) or label name?

bwplotka: Some ideas:

A) We could use grouper directly: https://github.com/thanos-io/thanos/blob/8040ff14c4463ae649821abea8b5244699672fc9/pkg/compact/compact.go#L231
to group only by external labels (from meta.json) and downsampling (from met.json as well) as it is safe to run concurrently such groups. Otherwise vertical compaction had to be enabled on target system and even then it's waste of compute someteims

B) We could do full concurrency as we want as long as we don't upload the last important file `meta.json`. With this target system is expected to not touch those blocks until delete period which is by defult 2 days. So we could do expensive/slow upload for blocks without meta.json and use option A for meta.json upload in time order.

Issue #4359, opened 2021-06-18
receive: multiple failure cases in non-trivial receiver topologies


ianbillett:
**Thanos, Prometheus and Golang version used**:

* Thanos: `HEAD`
* Prometheus: `v2.26.0`
* Golang: `go version go1.16.4 linux/amd64`

**Object Storage Provider**: `N/A`

**What happened**:

While working on https://github.com/thanos-io/thanos/issues/4326 - we discovered 4 failure cases when deploying non-trivial [post-split proposal](https://github.com/thanos-io/thanos/pull/4231) `receive` topologies. This results in the Thanos deployment not behaving as expected.

### 1: Request Replica > Ingestor Replication Factor
 <details>

Following the [post-split proposal](https://github.com/thanos-io/thanos/pull/4231), we are able to configure `recveive` instances to operate in `router`, `ingester` or `routing and ingesting` modes.

![arch1](https://user-images.githubusercontent.com/33524269/123135216-72e14880-d449-11eb-904d-39190f339383.png)

In the above example, if `router` is configured with a replication factor of `3`, we would expect the data to be replicated to all `ingestor` instances.

Instead, we observe that `router` fails to replicate the data with `quorum not reached` errors like the following:

<details>
 <summary>Routing logs (formatted for clarity)</summary>

```

09:44:41 receive-d1: level=error name=receive-d1 ts=2021-06-18T08:44:41.943443895Z caller=handler.go:352 component=receive component=receive-handler 
      err="3 errors: 
          replicate write request for endpoint receive_distributor_ingestor_mode-receive-i1:9091: quorum not reached: 
          2 errors: forwarding request to endpoint receive_distributor_ingestor_mode-receive-i2:9091: rpc error: code = InvalidArgument 
          desc = replica count exceeds replication factor; forwarding request to endpoint receive_distributor_ingestor_mode-receive-i1:9091: rpc error: code = AlreadyExists 
          desc = store locally for endpoint receive_distributor_ingestor_mode-receive-i1:9091: conflict; 
           
          replicate write request for endpoint receive_distributor_ingestor_mode-receive-i2:9091: quorum not reached: 
          2 errors: forwarding request to endpoint receive_distributor_ingestor_mode-receive-i3:9091: rpc error: code = InvalidArgument 
          desc = replica count exceeds replication factor; forwarding request to endpoint receive_distributor_ingestor_mode-receive-i2:9091: rpc error: code = AlreadyExists 
[...]

PR #4362, opened 2021-06-22
Add integration tests that excercise Thanos Receive edge cases

ianbillett:

## Changes

This PR adds 3 new integration test scenarios for Thanos Receive.

`router_replication` and `router_tree` both currently fail against `main`.

See https://github.com/thanos-io/thanos/issues/4359 for discussion.

## Verification

Tests are failing :upside_down_face:


GiedriusS:
Thanks for this! I am thinking that maybe let's convert `testutil.Ok` to `testutil.NotOk`, merge, and then fix this in a follow-up PR?

ianbillett:
Nice idea @GiedriusS :+1: 

I've updated the integration tests to fail on the 'correct' value. PTAL.

bwplotka:
`Add failing test` title sounds... scary ;p

test/e2e/receive_test.go
```diff
@@ -101,51 +187,161 @@ func TestReceive(t *testing.T) {
...
+					   └───────┘  │   │  └───────┘
+							   ┌──▼───▼──┐
+							   │         │
+							   │ Router1 │
+						  ┌────┤         ├───────┐
+						  │    └─────────┘       │
+					  ┌───▼─────┐          ┌─────▼─────┐
+					  │         │          │           │
+					  │ Router2 │          │ Ingestor1 │
+				  ┌───┤         ├───┐      │           │
```
bwplotka, line 223: Something looks off here

test/e2e/receive_test.go
```diff
@@ -101,51 +187,161 @@ func TestReceive(t *testing.T) {
...
-				"prometheus": "prom2",
-				"receive":    "i1",
-				"replica":    "0",
-				"tenant_id":  "default-tenant",
+		}, 3)
+
+		// TODO(ianbillett): change this assertion to testutil.Equals when #4359 is fixed.
+		// Currently, we expect this e2e test to fail due to bugs in the reicever split proposal implementation.
+		for _, series := range result {
+			testutil.NotEqual(t, series.Value, 2.0)
```
bwplotka, line 197: Why not equal something else with TODO?

pkg/testutil/testutil.go
```diff
@@ -79,6 +79,21 @@ func Equals(tb testing.TB, exp, act interface{}, v ...interface{}) {
 	tb.Fatal(sprintfWithLimit("\033[31m%s:%d:"+msg+"\n\n\texp: %#v\n\n\tgot: %#v%s\033[39m\n\n", filepath.Base(file), line, exp, act, diff(exp, act)))
 }
 
+// NotEqual fails the test if exp is equal to act.
+func NotEqual(tb testing.TB, exp, act interface{}, v ...interface{}) {
```
bwplotka, line 83: I would be strict and suggest removing it. It would be much better to assert on the expected thing, then assert that thing is not value X ...  WDYT? (:

bwplotka (changes requested):
Nice tests, thanks for those,

But I would not add new helper, see comment 🙈

test/e2e/receive_test.go
bwplotka: Essentially this can be surprising. So easy to match this assertion (so many options that passes). It essentially does not give any value

test/e2e/receive_test.go
ianbillett: Yeah - it renders properly in Goland :thinking: I'll have a play and see if I can make it look right.

pkg/testutil/testutil.go
ianbillett: Yeah fair enough - adding this did feel a bit sketchy :sweat_smile:

bwplotka (approved):
I love it, thanks! 

Sorry for the massive lag on checking those!

PR #4368, opened 2021-06-23
Only propagate write request replica for RouterIngestor receiver mode

ianbillett:

## Changes

Fixes https://github.com/thanos-io/thanos/issues/4359

This PR fixes the data-replication bugs we have seen in receive topologies that use the new functionality introduced in https://github.com/thanos-io/thanos/pull/4231.

The root of these bugs is the mechanism used to detect cycles in receiver clusters. This causes valid write requests to fail when we compose 'tree'-like architectures using `Router` and `Ingestor` modes (for example the ['router replication'](https://github.com/thanos-io/thanos/blob/main/test/e2e/receive_test.go#L119-L142) integration test)

We solve this problem by 'resetting' the write request replica for receivers in `Router` and `Ingestor` mode. The behaviour for `RouterIngestor` (the default before https://github.com/thanos-io/thanos/pull/4231) remains exactly the same.

The downside of this PR is that there are receiver topologies that we **know** will not behave as the user expects. The upside is that these are marginal edge cases for which a preferable architecture will always exist.

Previously we have solved this problem using a flag to enable the 'resetting' behaviour, which required the user to understand a lot about receiver internals. On balance, it is preferable to have better user experiences with a couple of known weird cases (with known fixes) than a confusing user experience (what is a receive replica?).   

## Verification

The recently added failing integration tests (https://github.com/thanos-io/thanos/pull/4362) now pass.


ianbillett:
The failed tests all pass locally - assuming it was a flake - retrying.

bwplotka:
Sorry for lag. It's on me now, I want to make sure there is no other way to detect this case automatically

bwplotka (changes requested):
Ok, I spent 19m on this PR, cloned locally etc, and I still a bit confused why we need an extra flag.

If the receive has NO `local-endpoint` it is router and should always override replica in request, no? 🤔

bwplotka:
OK, sounds like this also does not work well, since it might be an infinite cycle -> single hashring, no local endpoint and single hashring points to itself.

bwplotka (commented):
Keep non surprising behaviour of no nested replications and add flag for `replication-reset`. I wonder if we can do this in  `Hidden` as we still not sure if it's clear to user how to use it (:

ianbillett:
After chatting this through with @bwplotka yesterday - I've decided to change direction with this PR for something that has a better QoL for users.

cmd/thanos/receive.go
```diff
@@ -77,9 +77,11 @@ func registerReceive(app *extkingpin.App) {
 			MaxExemplars:           conf.tsdbMaxExemplars,
 		}
 
-		// Enable ingestion if endpoint is specified or if both the hashrings configs are empty.
-		// Otherwise, run the receiver exclusively as a distributor.
-		enableIngestion := conf.endpoint != "" || (conf.hashringsFileContent == "" && conf.hashringsFilePath == "")
+		// Has the user provided some kind of hashring configuration?
+		hashringSpecified := conf.hashringsFileContent != "" || conf.hashringsFilePath != ""
```
bwplotka, line 81:
```suggestion
		isHashringSpecified := conf.hashringsFileContent != "" || conf.hashringsFilePath != ""
```

cmd/thanos/receive.go
```diff
@@ -77,9 +77,11 @@ func registerReceive(app *extkingpin.App) {
...
+		// Has the user specified the --receive.local-endpoint flag?
+		localEndpointSpecified := conf.endpoint != ""
```
bwplotka, line 83:
```suggestion
		isLocalEndpointSpecified := conf.endpoint != ""
```

cmd/thanos/receive.go
```diff
@@ -77,9 +77,11 @@ func registerReceive(app *extkingpin.App) {
...
+		receiveMode := receive.DetermineMode(hashringSpecified, localEndpointSpecified)
```
bwplotka, line 84: If we do this complex thing of determining, why not just passing `config`? Sounds like related complexity we can hide (reader of code has  to visit this thing anyway)

cmd/thanos/receive.go
```diff
@@ -144,6 +143,9 @@ func runReceive(
 		return err
 	}
 
+	// Has this thanos receive instance been configured to ingest metrics into a local TSDB?
+	enableIngestion := receiveMode == receive.IngestorOnly || receiveMode == receive.RouterIngestor
```
bwplotka, line 147: This is used only once? If yes, let's just inline WDYT?

pkg/receive/config.go
```diff
@@ -29,6 +29,28 @@ var (
...
+	RouterIngestor ReceiverMode = "RouterIngestor"
+)
+
+// DetermineMode returns the ReceiverMode that this receiver is configured to run in.
+// This is used to configure this Receiver's forwarding and ingesting behavior at runtime.
+func DetermineMode(hashringSpecified, localEndpointSpecified bool) ReceiverMode {
+	switch {
+	case hashringSpecified && localEndpointSpecified:
+		return RouterIngestor
+	case hashringSpecified && !localEndpointSpecified:
```
bwplotka, line 46: Let's add note here that IF in this case hashring specifies the same router (aciddently or not) it will cause crash loop (fork bomb) - this is fine, but let's just explicitly write it to not forget

bwplotka (dismissed):
Nice, amazing. Couple of style nits, but otherwise LGTM!

Let's make sure this is released (:

cmd/thanos/receive.go
ianbillett: It is actually used twice [here](https://github.com/thanos-io/thanos/blob/be9b0f5d85a23795403c60aba96e1eb22073a1bc/cmd/thanos/receive.go#L150) & [here](https://github.com/thanos-io/thanos/blob/be9b0f5d85a23795403c60aba96e1eb22073a1bc/cmd/thanos/receive.go#L225)

cmd/thanos/receive.go
ianbillett: yeah that would simplify things - nice idea

bwplotka (approved):
Sweet 👍🏽 

Allowing @GiedriusS to take a look and review before merging

GiedriusS (approved):
Looks great! Well done! I love the ASCII illustrations, makes my life easier :smile:

kakkoyun (commented):
Amazing PR. Modes are much more clear now.

PR #4369, opened 2021-06-24
Build: do not upgrade alpine version

daixiang0:
Signed-off-by: Long Dai <long0dai@foxmail.com>

`apk upgrade` works like `apt upgrade`, it would upgrade OS version which goes against our purpose of ​​using a fixed image version, refer to [doc](https://wiki.alpinelinux.org/wiki/Upgrading_Alpine#Upgrade_Operating_System)


bwplotka (approved):
nice!

PR #4370, opened 2021-06-24
Adds more contact options when creating a new issue on Github

wiardvanrij:

Inspired and copy/pasta'd from jaegertracing :) 

It will look like this (but with our own text)

image

As discussed via slack: https://cloud-native.slack.com/archives/CL25937SP/p1622453498082200


bwplotka (approved):
😱 =D

Issue #4382, opened 2021-06-28
Receivers: Hashring gRPC calls are failing when Remote Write mTLS is enabled 


narenreddycs:
**Thanos, Prometheus and Golang version used**:
v0.20.0

**Object Storage Provider**:
AWS S3

**What happened**:
When mTLS is enabled for client authentication from local prometheus -> Thanos Receivers and hashring is enabled for Thanos receivers, remote write requests are failing with gRPC authentication failures. Only when mTLS is disabled, remote write requests with hashring are succeeding. 

Hashring:
```
  hashrings.json: |
    [
      {
        "hashring": "soft-tenants",
        "endpoints": [
          "thanos-receive-dev-0.thanos-receive-dev-grpc.hashring-dev.svc.cluster.local:10907",
          "thanos-receive-dev-1.thanos-receive-dev-grpc.hashring-dev.svc.cluster.local:10907",
          "thanos-receive-dev-2.thanos-receive-dev-grpc.hashring-dev.svc.cluster.local:10907"
          ]
      }
    ]
 ```
Receiver config:
```
        - receive
        - --tsdb.path=/var/thanos/receive
        - --debug.name=receive       
        - --log.level=debug       
        - --grpc-address=0.0.0.0:10907
        - --http-address=0.0.0.0:10902
        - --remote-write.server-tls-cert=/etc/cert/servercert-config.yaml
        - --remote-write.server-tls-key=/etc/certkey/servercertkey-config.yaml
        - --remote-write.server-tls-client-ca=/etc/certca/servercertca-config.yaml
        - --label=receive_replica="$(NAME)"
        - --label=receive="true"
        - --remote-write.address=0.0.0.0:19291
        - --tsdb.retention=3d  
        - --tsdb.min-block-duration=2h
        - --tsdb.max-block-duration=2h
        - --receive.replication-factor=1
        - --objstore.config-file=/etc/config/object-store-config.yaml
        - --receive.hashrings-file=/var/lib/thanos-receive/hashrings.json
        - --receive.local-endpoint=$(NAME).thanos-receive-dev-grpc.hashring-dev.svc.cluster.local:10907
 ```
local prometheus config:
```
remote_write:
- url: https://<>/api/v1/receive
  tls_config:
    ca_file: ~/client-certs/cacert.crt
    cert_file: ~/client-certs/client.cer
    key_file: ~/client-certs/client.key
```

**What you expected to happen**:
[...]

PR #4388, opened 2021-06-30
cmd/thanos/receive: fix dialopts TLS heuristic

squat:
Currently, the Thanos receive gRPC dialopts are configured to use TLS
when the remote-write HTTP server is configured with a TLS certificate.
This heuristic is not correct: just because the remote-write HTTP server
is exposed over TLS does not mean that the hashring gRPC server is using
TLS. This commit fixes the heuristic for enabling TLS in the gRPC client
for forwarding remote-write requests internally within a hashring.

Fixes: https://github.com/thanos-io/thanos/issues/4382.


ianbillett:
Nice find :muscle: How have you been able to verify this change indeed fixes the problem? I'm not aware of a test that excercises this code path.

bwplotka (approved):

bwplotka:
Thanks!

PR #4377, opened 2021-06-26
proposal: Automated, per-endpoint mTLS

Namanl2001:
Related issue: #977

PR #4389, opened 2021-06-30
querier: per-endpoint configuration

Namanl2001:
The Thanos Querier component supports basic mTLS configuration for internal gRPC communication. This works great for basic use cases but it still requires extra forward proxies to make it work for bigger deployments.

This PR add support for per-endpoint TLS configuration in Thanos Query Component for internal gRPC communication. Here we are introducing a new CLI option `--endpoints.config`, which will accept the content/path to a yaml file which contains a list as follows:
```
- name: ""
  tls_config:
    cert_file: ""
    key_file: ""
    ca_file: ""
    server_name: ""
  endpoints: []
  endpoints_sd_files:
    - files: []
  mode: ""
```

## Changes
Based on proposal #4377 

* Added new `--endpoint.config` cli option
* Added new file `pkg/store/config.go` to load config from cli options (in list form)
* Few changes in `query.go` to iterate over the list items

## Verification

Added e2e tests for `--endpoints.config` with mutual TLS.


pkg/store/config.go
```diff
@@ -0,0 +1,94 @@
...
+			if mp[ep] {
+				return []Config{}, fmt.Errorf("%s endpoint provided more than once", ep)
+			}
+			mp[ep] = true
+		}
+	}
+
+	// Adding --endpoint, --endpoint_sd_files info to []endpointConfig
+	cfg1 := Config{}
+	for _, addr := range endpointAddrs {
```
bwplotka, line 71: What if there are none? (same below)

Essentially it might confusing for everyone that we have empty configs - let's avoid this work and surprises

bwplotka (commented):
Feedback from our 1:1, great job! 

Let's make sure to test it too. Is the plan to NOT to dynamic reload in this PR? (that's great idea BTW)

cmd/thanos/query.go
```diff
@@ -330,58 +338,75 @@ func runQuery(
 		Help: "The number of times a duplicated store addresses is detected from the different configs in query",
 	})
 
-	dialOpts, err := extgrpc.StoreClientGRPCOpts(logger, reg, tracer, secure, skipVerify, cert, key, caCert, serverName)
```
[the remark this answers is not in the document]
Namanl2001: I think we should use `cert, key, caCert, serverName` ONLY for endpoints provided in `--endpoint`, `--endpoint.strict`, `--endpoint.sd-files`.
Added this to the [proposal](https://github.com/thanos-io/thanos/pull/4377/files#diff-53b4dd652e0c2c941d48434c6267578837805bf9efc7508e142f145bd8a65d82R55).

bwplotka (commented):
Nice work!

As mentioned in our 1:1 wonder if we can help @hitanshu-mehta and help ourselves to do this work on https://github.com/thanos-io/thanos/pull/4421 and review his PR 🤔

cmd/thanos/query.go
```diff
@@ -422,11 +453,129 @@ func runQuery(
 			dialOpts,
 			unhealthyStoreTimeout,
 		)
-		proxy            = store.NewProxyStore(logger, reg, stores.Get, component.Query, selectorLset, storeResponseTimeout)
-		rulesProxy       = rules.NewProxy(logger, stores.GetRulesClients)
-		targetsProxy     = targets.NewProxy(logger, stores.GetTargetsClients)
-		metadataProxy    = metadata.NewProxy(logger, stores.GetMetadataClients)
-		exemplarsProxy   = exemplars.NewProxy(logger, stores.GetExemplarsStores, selectorLset)
+		storeSets = append(storeSets, stores)
```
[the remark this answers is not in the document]
Namanl2001: Because we need dialopts for each yaml list item’s endpoints set. 
Storing it in `storeSets` as we need them outside the loop.

cmd/thanos/query.go
```diff
@@ -529,12 +568,72 @@ func runQuery(
...
-	statusProber := prober.Combine(
-		httpProbe,
-		grpcProbe,
-		prober.NewInstrumentation(comp, logger, extprom.WrapRegistererWithPrefix("thanos_", reg)),
+	var (
+		allClients = func() []store.Client {
+			var get []store.Client
+			for _, ss := range storeSets {
+				get = append(get, ss.Get()...)
+			}
```
Namanl2001, line 576: We have this loop in all the 4 variables below, I tried to have this once globally and then just retuning the variables inside these func(), but it's weird to see that it was not working as expected (e2e tests were also failing) and I have to change and make it repetitive in this [commit](https://github.com/thanos-io/thanos/pull/4389/commits/672b9d7810f36e619d34e52f679cbd2670a92a3d) after which all tests were happy.

Namanl2001:
I am stuck with testing mTLS in querier. 
- I've created certs using [this script](https://github.com/thanos-io/thanos/blob/f5be459ab5cd9f1dbabbab8ff5c72783765d3636/test/e2e/certs/create.sh). 

- Without mTLS all e2e-tests are passing in this feature branch.
- I've also tried enabling one way TLS (server side only OR client side only) but got the same error.
 
- I've also tested mTLS on the main branch with same TLS credentials but tests are failing with the same err log as on this branch.

bwplotka:
```
14:00:33 Killing querier-1
14:00:33 querier-1: level=warn name=querier-1 ts=2021-07-28T14:00:33.424080098Z caller=storeset.go:574 component=storeset msg="update of store node failed" err="getting metadata: fetching store info from e2e_test_query_config-sidecar-alone:9091: rpc error: code = DeadlineExceeded desc = latest balancer error: connection error: desc = \"transport: authentication handshake failed: tls: first record does not look like a TLS handshake\"" address=e2e_test_query_config-sidecar-alone:9091
14:00:33 Killing sidecar-ha2
14:00:34 Killing prometheus-ha2
14:00:34 Killing sidecar-ha1
14:00:34 Killing prometheus-ha1
14:00:34 Killing sidecar-remote-and-sidecar
14:00:35 Killing prometheus-remote-and-sidecar
14:00:35 Killing sidecar-alone
=== CONT  TestQueryWithEndpointConfig
Error:     query_test.go:283: query_test.go:283:
        
         unexpected error: unable to find metrics [thanos_store_nodes_grpc_connections] with expected values. Last error: <nil>. Last values: [4]
```

cmd/thanos/query.go
```diff
@@ -330,21 +347,26 @@ func runQuery(
 		Help: "The number of times a duplicated store addresses is detected from the different configs in query",
 	})
 
-	dialOpts, err := extgrpc.StoreClientGRPCOpts(logger, reg, tracer, secure, skipVerify, cert, key, caCert, serverName)
-	if err != nil {
-		return errors.Wrap(err, "building gRPC client")
+	// TLSConfig for endpoints provided in --endpoint, --endpoint.sd-files and --endpoint-strict.
+	var TLSConfig store.TLSConfiguration
+	if secure && len(endpointConfigYAML) == 0 {
```
bwplotka, line 352: It's kind of confusing on 2 levels:

Users: "my flag is overridden without my knowledge" 
Code - secure is global but endpoints TLS and per config

pkg/extgrpc/client.go
```diff
@@ -48,13 +54,14 @@ func StoreClientGRPCOpts(logger log.Logger, reg *prometheus.Registry, tracer ope
 		reg.MustRegister(grpcMets)
 	}
 
-	if !secure {
+	// Insecure if secure is false and no TLS config is supplied.
+	if !secure && (tlsConfig == store.TLSConfiguration{}) {
```
bwplotka, line 58: one way of improving this would be maybe doing `secure  = tlsConfig == store.TLSConfiguration{}`

cmd/thanos/query.go
Namanl2001: Addressed. 👍

Namanl2001:
Finally I was able to test mutual TLS in querier. Earlier I was configuring tls only in querier for both client and server. Which was wrong as in our case sidecar is the server and after configuring the same in sidecar it worked 😄.

This PR needs another round of review.

PS: Failing tests seems unrelated as all tests are passing locally.

pkg/query/storeset.go
```diff
@@ -143,13 +143,13 @@ type storeSetNodeCollector struct {
...
 
-func newStoreSetNodeCollector() *storeSetNodeCollector {
+func newStoreSetNodeCollector(configInstance string) *storeSetNodeCollector {
 	return &storeSetNodeCollector{
 		storeNodes: map[component.StoreAPI]map[string]int{},
 		connectionsDesc: prometheus.NewDesc(
 			"thanos_store_nodes_grpc_connections",
 			"Number of gRPC connection to Store APIs. Opened connection means healthy store APIs available for Querier.",
-			[]string{"external_labels", "store_type"}, nil,
+			[]string{"external_labels", "store_type"}, map[string]string{"config_instance": configInstance},
```
bwplotka, line 152: What is config instance?

test/e2e/certs/testca.crt
```diff
@@ -0,0 +1,19 @@
+-----BEGIN CERTIFICATE-----
```
bwplotka, line 1: Did you set any expiry time?

bwplotka: I wonder if we can just create them in run time instead of committing. The reason is that our `creat` script will be unused for 1y or so , so it will most likely no work when we will need it. It might be better to put it in CI and use it.

test/e2e/e2ethanos/services.go
```diff
@@ -95,25 +94,29 @@ func NewPrometheus(sharedDir, name, config, promImage string, enableFeatures ...
 	return prom, container, nil
 }
 
-func NewPrometheusWithSidecar(sharedDir, netName, name, config, promImage string, enableFeatures ...string) (*e2e.HTTPService, *Service, error) {
+func NewPrometheusWithSidecar(sharedDir, netName, name, config, promImage string, tlsConfig []string, enableFeatures ...string) (*e2e.HTTPService, *Service, error) {
```
bwplotka, line 97:
```suggestion
func NewPrometheusWithSidecar(sharedDir, netName, name, config, promImage string, extraArgs []string, enableFeatures ...string) (*e2e.HTTPService, *Service, error) {
```

bwplotka (commented):
Nice! Looks good, just some nits.

cmd/thanos/query.go
```diff
@@ -190,13 +197,21 @@ func registerQuery(app *extkingpin.App) {
 			return errors.Errorf("Address %s is duplicated for --target flag.", dup)
 		}
 
-		var fileSD *file.Discovery
+		endpointConfigYAML, err := endpointConfig.Content()
+		if err != nil {
+			return err
+		}
+
+		if (len(*fileSDFiles) != 0 || len(*stores) != 0 || *secure) && len(endpointConfigYAML) != 0 {
```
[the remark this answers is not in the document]
Namanl2001: Make sense. I think I misunderstood what you said earlier. Thanks for the clarification 👍

pkg/query/storeset.go
Namanl2001: As we are iterating over the config list, so for more than one list item, it is trying to register a new metrics collector and was throwing the error `panic: duplicate metrics collector registration attempted`, so to make each metrics collector unique, we've added a `name` option in endpoint config.

The other option was to use the config item number to make each metric collector unique. @kakkoyun suggested taking a unique name from the user (for each config) so that in case there is some bad configuration set from the user side, they can debug, and here `name` would be more helpful than a number (e.g., 0, 1) to know where their config is wrong.

The same is the case with the DNS provider, for which it was not giving any specific error but not functioning as expected. For instance, for 2 configs, it was first registering the config[0] endpoints then removing them, saying that "these endpoints are not more active" and registering the config[1] endpoints and then removing these too with the same msg. And now again, adding the config[0] endpoints, and the same process repeats. 
The timely addition and removal is due to the periodical updating of endpoints ["update()"](https://github.com/thanos-io/thanos/pull/4389/files#diff-f9808a0ac67a474a18f6a7ea767593f85c216aeafe322c1ff5d17c81ababff95R476)

test/e2e/certs/testca.crt
Namanl2001: Addressed. It was quite tricky for me but managed to do it after spending hours 😄

cmd/thanos/receive.go
```diff
@@ -123,16 +123,20 @@ func runReceive(
 		return err
 	}
 
+	var TLSConfig store.TLSConfiguration
```
kakkoyun, line 126: nit: We can inline this struct creation. And I think we should.

pkg/query/storeset.go
kakkoyun: Maybe to avoid any confusion call it `provider_name` or `config_provider_name`?

kakkoyun (commented):
Looking great! Only a few wrinkles to iron out!

cmd/thanos/query.go
```diff
@@ -529,12 +573,72 @@ func runQuery(
...
 	}
 
-	grpcProbe := prober.NewGRPC()
-	httpProbe := prober.NewHTTP()
-	statusProber := prober.Combine(
-		httpProbe,
-		grpcProbe,
-		prober.NewInstrumentation(comp, logger, extprom.WrapRegistererWithPrefix("thanos_", reg)),
+	var (
+		allClients = func() []store.Client {
```
kakkoyun, line 577: It would be amazing to add a comment here to explain what we are doing and add a link to the proposal.

cmd/thanos/query.go
```diff
@@ -122,6 +127,8 @@ func registerQuery(app *extkingpin.App) {
 	fileSDInterval := extkingpin.ModelDuration(cmd.Flag("store.sd-interval", "Refresh interval to re-read file SD files. It is used as a resync fallback.").
 		Default("5m"))
 
+	endpointConfig := extflag.RegisterPathOrContent(cmd, "endpoint.config", "YAML file that contains store API servers configuration. Either use this option or separate endpoint options (endpoint, endpoint.sd-files, endpoint.srict).", extflag.WithEnvSubstitution())
```
bwplotka, line 130:
```suggestion
	endpointConfig := extflag.RegisterPathOrContent(cmd, "endpoint.config", "YAML file that contains set of endpoints (e.g Store API) with optional TLS options. Either use this option or separate endpoint options (endpoint, endpoint.sd-files, endpoint.strict).", extflag.WithEnvSubstitution())
```

cmd/thanos/query.go
```diff
@@ -372,16 +393,45 @@ func runQuery(
 		dns.ResolverType(dnsSDResolver),
 	)
 
-	var (
-		stores = query.NewStoreSet(
+	// storeSets contains separate store for each endpointConfig.
+	var storeSets []*query.StoreSet
+	for _, config := range endpointConfig {
+		secure = !(config.TLSConfig == store.TLSConfiguration{})
```
bwplotka, line 399:
```suggestion
		secure = (config.TLSConfig != store.TLSConfiguration{})
```

cmd/thanos/query.go
```diff
@@ -422,94 +472,89 @@ func runQuery(
 			dialOpts,
```
bwplotka, line 472: Am I wrong if I would say that if we somehow pass through spec the dial options per addresses we could have just one storeset right? Do you think it would be worth doing?

cmd/thanos/query.go
```diff
@@ -529,12 +574,74 @@ func runQuery(
...
 	}
 
-	grpcProbe := prober.NewGRPC()
-	httpProbe := prober.NewHTTP()
-	statusProber := prober.Combine(
-		httpProbe,
-		grpcProbe,
-		prober.NewInstrumentation(comp, logger, extprom.WrapRegistererWithPrefix("thanos_", reg)),
+	var (
+		// Adding separate for loop for each client func() below because storeSets is being populated in a go-routine and this code executes before it.
```
bwplotka, line 578: Yea, this is not the best. I mentioned Idea how we can ensure single storeset

bwplotka (dismissed):
Thanks! It looks good to me, except for one part. Added suggestion what we can do to ensure single storeset. Otherwise LGTM! 💪🏽

cmd/thanos/query.go
Namanl2001: Thanks for the review.
> if we somehow pass through spec the dial options per addresses we could have just one storeset right?
 
But dialopts are different for each endpoint config (as it depends on `config.tlsConfig`). So even if we somehow pass it through spec, we would not have one storeset as storeset would be defined inside the loop to have all tlsConfig.

To have only one storeset I thought to take the storeset initialization outside the loop but we can't because it uses dnsProvider which needs to be different for each endpoint config else it gives error.

So we need to have storeset inside the loop (which means more than one storeset) to have different dnsProvider and tlsConfig corresponding to different endpoint sets.

> Do you think it would be worth doing?

As discussed with @kakkoyun (in our 1:1) we expect only few configs (like 4 or 5) from the user so having these much of storesets would not affect the performance much. :)


<!--
Ah so, for one storeset we need to define it once, outside the for loop right?
I thought of your idea as taking the [storeset](https://github.com/thanos-io/thanos/pull/4389/files#diff-f9808a0ac67a474a18f6a7ea767593f85c216aeafe322c1ff5d17c81ababff95R427) outside the loop and then passing dialopts through spec by iterating over endpoint config but found that it is used at many places like store.update() which is inside the loop.
--!>

cmd/thanos/query.go
Namanl2001: > Either use this option or separate endpoint options (endpoint, endpoint.sd-files, endpoint.strict)

This line is wrong as we accept separate endpoint options (endpoint, endpoint.sd-files, endpoint.strict) along with new `--endpoint.config`.

My suggestion: 
`YAML file that contains set of endpoints (e.g Store API) with optional TLS options. To enable TLS either use this option or deprecated ones --grpc-client-tls*` WDYT?

bwplotka: Perfect 👍🏽

cmd/thanos/query.go
bwplotka: > But dialopts are different for each endpoint config (as it depends on config.tlsConfig). So even if we somehow pass it through spec, we would not have one storeset as storeset would be defined inside the loop to have all tlsConfig.

Yea, I think we could do that per addresses.

> As discussed with @kakkoyun (in our 1:1) we expect only few configs (like 4 or 5) from the user so having these much of storesets would not affect the performance much. :)

I know companies running Thanos against 300 of StoreAPIs, so I would be careful here. But also I am not worried about performance too much here, rather maintainability and debuggability when you have different metrics per 100 of those storesets potentially. And different health loops. Not a biggie, maybe I can try propose a PR to yours if we want to compare different solutions? (:

Namanl2001: > I can try propose a PR to yours if we want to compare different solutions

That would be really helpful 🤗

Namanl2001:
rebased successfully to main 😵‍💫 

- [ ] move endpointConfig to `query/config.go`
- [ ] let's try to have one function that loads the config (if non empty) and merge flags into one endpoint config - might be cleaner
- [ ] try to have one endpoint set in `cmd/thanos/query.go`

Namanl2001:
This PR works fine (see [approval](https://github.com/thanos-io/thanos/pull/4389#pullrequestreview-731987027)) but there is a suggestion to ensure a single storeset. The work for which would be carried out on another PR.

This was my Google Summer of Code 2021 project. ❤️ 

Closing

Issue #4305, opened 2021-06-04
receive: Replica count exceeds replication factor although it does not?


bwplotka:
```
level=debug ts=2021-05-27T00:57:03.170460586Z caller=handler.go:641 component=receive component=receive-handler msg="failed to handle request" err="replica count exceeds replication factor"
replicas: 6
receive.replication-factor=2
```

I don't really understand this, replica does not exceed replication 🤔

bwplotka:
This might need more info, I suspect 2 replicas were just up. Worth to document this error/failure mode better

squat:
This is saying that a receiver A got a forwarded message from receiver B.

B said "this is copy number 6 of the time series."

A looks at it's configuration and says "huh, replication factor is only 2, so why did i get copy number 6? This must be a mistake"

ianbillett:
I don't think we need to do anything serious here apart from maybe re-wording the error message slightly. Perhaps this is slightly clearer?
> request replica number exceeds receiver's replication factor

PR #4394, opened 2021-07-01
Add error log to receiver and clarify bad replica error message

ianbillett:
Signed-off-by: Ian Billett <ibillett@redhat.com>

## Changes

Fixes https://github.com/thanos-io/thanos/issues/4305

## Verification

Running the integration tests from https://github.com/thanos-io/thanos/pull/4362 I see the following log:
```
11:25:28 receive-i3: level=error name=receive-i3 ts=2021-07-01T10:25:28.330239422Z caller=handler.go:261 component=receive component=receive-handler err="request replica exceeds receiver replication factor" msg="write request rejected" request_replica=2 replication_factor=1
```

### DIFF
--- a/pkg/receive/handler.go
+++ b/pkg/receive/handler.go
@@ -61,7 +61,7 @@ var (
 	// errConflict is returned whenever an operation fails due to any conflict-type error.
 	errConflict = errors.New("conflict")
 
-	errBadReplica  = errors.New("replica count exceeds replication factor")
+	errBadReplica  = errors.New("request replica exceeds receiver replication factor")
 	errNotReady    = errors.New("target not ready")
 	errUnavailable = errors.New("target not available")
 )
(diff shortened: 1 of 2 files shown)


squat (approved):
lgtm (:

pkg/receive/handler.go
```diff
@@ -258,6 +258,8 @@ type replica struct {
 func (h *Handler) handleRequest(ctx context.Context, rep uint64, tenant string, wreq *prompb.WriteRequest) error {
 	// The replica value in the header is one-indexed, thus we need >.
 	if rep > h.options.ReplicationFactor {
+		level.Error(h.logger).Log("err", errBadReplica, "msg", "write request rejected",
```
bwplotka, line 261: Can we do this in the place where error is handled? Handling error twice (by returning and logging) is a bad smell 🙈

ianbillett: We could :+1:  I chose not to as we would then need to replicate this in two places in a switch statement ([here](https://github.com/thanos-io/thanos/blob/f1597d785b7f33dc3912faa6b993198bd68eae1b/pkg/receive/handler.go#L352) & [here](https://github.com/thanos-io/thanos/blob/f1597d785b7f33dc3912faa6b993198bd68eae1b/pkg/receive/handler.go#L656)).

I felt this was slightly tidier, but happy to be convinced otherwise :man_shrugging:

bwplotka: Hm... there is a risk we will handle things twice, but not a biggie! (:

bwplotka (approved):

PR #4412, opened 2021-07-05
Fixed check for empty string

iamrajiv:
## Description

It is not recommended to use `len` for the empty string test. A string can be tested for its emptiness either by treating it as a slice and calculating the length of the slice or by treating it as a string and directly comparing the value. While both produce identical code when compiled, it makes more sense to treat a string as itself, than a slice, for the sake of comparison of values.


iamrajiv:
@yeya24 @onprem Can you take a look.

CHANGELOG.md
```diff
@@ -19,6 +19,7 @@ We use *breaking :warning:* to mark changes that are not backward compatible (re
 - [#4239](https://github.com/thanos-io/thanos/pull/4239) Add penalty based deduplication mode for compactor.
 - [#4292](https://github.com/thanos-io/thanos/pull/4292) Receive: Enable exemplars ingestion and querying.
 - [#4392](https://github.com/thanos-io/thanos/pull/4392) Tools: Added `--delete-blocks` to bucket rewrite tool to mark the original blocks for deletion after rewriting is done.
+- [#4412](https://github.com/thanos-io/thanos/pull/4412) Fixed check for empty string
```
bwplotka, line 22: I don't think this is user impacting change. Do you mind removing this? (:

bwplotka (dismissed):
Nice, I like it, thanks 💪🏽

squat (approved):
Thanks

PR #4413, opened 2021-07-05
Combined same types of function params

iamrajiv:
- [ ] I added CHANGELOG entry for this change.
- [ ] Change is not relevant to the end user.

## Changes

If parameters of the same type lie consecutively, mention their type once at the end of the last parameter.

Unlike other languages, like C, where all parameters must be specified with types explicitly, it is not required to do so in Go. Combining the types is usually recommended for the sake of brevity.


yeya24 (commented):
Nice, thanks for your contribution. The changes look nice, but we don't need a changelog for it as it is not user-facing. Can you please remove the changelog entry?

yeya24 (approved):
LGTM

PR #4416, opened 2021-07-05
Multiple `append` combined into a single call

iamrajiv:
## Description

Appending calls in a single call of `append` allocates memory just once to accommodate all the elements to be appended. Whereas multiple calls to `append` introduce many overheads, most notably being the possibility of more calls for memory allocation because the total number of elements to be appended over multiple calls of `append` is unknown beforehand, resulting in inaccurate preallocation.


iamrajiv:
@yeya24 @onprem Can you take a look.

yeya24:
For readability reasons, it is better to keep the original code.

PR #4418, opened 2021-07-06
Fixed non-octal os.FileMode

iamrajiv:
## Description

[os.FileMode](https://golang.org/pkg/os/#FileMode) should be an octal number. For example 0700, 0644.


ianbillett (changes requested):
@iamrajiv please take another look at this PR - I think you have borked a commit or something :)

iamrajiv:
@ianbillett Please take a look now.

Namanl2001:
`bindata.go` is an auto generated file and we've to regenerate it after every change in the UI using [make assets](https://github.com/thanos-io/thanos/blob/8d3ea521598cf8d37b6079eabe4e562939a3c66d/Makefile#L98). I guess the changes in this pr are temporary and would be reverted once the file is regenerated again in future. I wonder if we could have some permanent solution to this.

yeya24:
@Namanl2001 is right. Close this pr now.

ianbillett:
@Namanl2001 is right - our [Makefile](https://github.com/thanos-io/thanos/blob/e3b0baf7de9dde1887253b1bb19d78ae71a01bf8/Makefile#L104) generates `pkg/ui/bindata.go` so we don't want to make or maintain any modifications to that.

The changes you propose are valid, but they are resulting from the code generated by the upstream go-bindata project.

bwplotka:
Let's close it and improve go-bindata then. Do you mind contributing there @iamrajiv? Thanks!

iamrajiv:
Yeah sure @bwplotka will try to improve `bindata.go` will make PR soon. I am still going through the code structure of Thanos.

PR #4433, opened 2021-07-09
Added receive mTLS test.

bwplotka:


test/e2e/receive_test.go
```diff
@@ -715,4 +787,145 @@ func TestReceive(t *testing.T) {
...
+				"prometheus": "prom1",
+				"receive":    "ingestor",
+				"replica":    "0",
+				"tenant_id":  "default-tenant",
+			},
+		})
+	})
+}
+
+func cpyDir(scrDir, dest string) error {
```
ianbillett, line 871: What does this function do? A recursive copy from `src` to `dest`? Please add a comment

ianbillett: Is there not a library function to do this for us?

ianbillett (approved):
Generally LGTM - nice to see these parts of the code base being exercised :+1: 

My only though would be that we could generate these certificates as part of the e2e setup rather than committing to the repo :man_shrugging: Not a blocker though.

test/e2e/certs/create.sh
```diff
@@ -0,0 +1,71 @@
...
+  -in e2e_receive_mtls_single_ingestor_client.csr \
+  -CA testca.crt \
+  -CAkey testca.key \
+  -CAcreateserial \
+  -days 99999 \
+  -extfile <(
+    cat <<-EOF
+basicConstraints = CA:FALSE
+nsCertType = client
+nsComment = "OpenSSL Generated Server Certificate"
```
Namanl2001, line 61:
```suggestion
nsComment = "OpenSSL Generated Client Certificate"
```

test/e2e/certs/create.sh
```diff
@@ -0,0 +1,71 @@
...
+  -keyout testca.key \
+  -out testca.crt
+
+openssl genrsa -out testserver.key 2048
+openssl genrsa -out testclient.key 2048
+
+openssl req \
+  -new \
+  -key testserver.key \
+  -subj "/CN=e2e_receive_mtls_single_ingestor-receive-ingestor" \
```
Namanl2001, line 18: I believe we can use the same certs for mTLS testing in querier (and maybe in other components too). So why not have some common name for the global cert files 🤔

Namanl2001:
Work in progress

PR #4435, opened 2021-07-11
Parallelize compactor block downloading and verification step

yeya24:
Signed-off-by: ben.ye <ben.ye@bytedance.com>

Now we are seeing a very slow blocks download and verification step (also due to our large indexes caused by some high cardinality labels).

For example, it takes 4h 23m to download and verify 7 raw blocks.

```
level=info ts=2021-07-10T01:17:45.841189934Z caller=compact.go:762 group="0@{  prometheus_replica=\"prometheus\"}" groupKey=0@17135574873524709607 msg="downloaded and verified blocks; compacting blocks" plan="[/var/thanos/compact/compact/0@17135574873524709607/01FA1KWD5B2DGR069Q96MWHP4M /var/thanos/compact/compact/0@17135574873524709607/01FA1TZEDY8EA5G5BGG0XSG9BG /var/thanos/compact/compact/0@17135574873524709607/01FA2261ZPD22S4E245XH1JZVD /var/thanos/compact/compact/0@17135574873524709607/01FA29V64MYFGG0QG1S9227ZEQ /var/thanos/compact/compact/0@17135574873524709607/01FA2V0GHS1FEVTFKQZ2ZVM79Y /var/thanos/compact/compact/0@17135574873524709607/01FA31XP4FKEE5AC6R6JNX9K98 /var/thanos/compact/compact/0@17135574873524709607/01FA38ZTYQMFR9GW36G717AERS]" duration=4h23m22.457449144s
```

## Changes

Use error group and make the block downloading and index verification concurrent. Ideally, this makes the download and verification step faster.

### DIFF
--- a/pkg/compact/compact.go
+++ b/pkg/compact/compact.go
@@ -727,10 +727,10 @@ func (cg *Group) compact(ctx context.Context, dir string, planner Planner, comp 
 	toCompactDirs := make([]string, len(toCompact))
 
 	var eg errgroup.Group
-	for i, meta := range toCompact {
+	for _, meta := range toCompact {
 		meta := meta
 		bdir := filepath.Join(dir, meta.ULID.String())
-		toCompactDirs[i] = bdir
+		toCompactDirs = append(toCompactDirs, bdir)
 		for _, s := range meta.Compaction.Sources {
 			if _, ok := uniqueSources[s]; ok {
 				return false, ulid.ULID{}, halt(errors.Errorf("overlapping sources detected for plan %v", toCompact))
--- a/pkg/compact/compact.go
+++ b/pkg/compact/compact.go
@@ -724,10 +724,13 @@ func (cg *Group) compact(ctx context.Context, dir string, planner Planner, comp 
 
 	// Once we have a plan we need to download the actual data.
 	begin := time.Now()
+	toCompactDirs := make([]string, len(toCompact))
 
-	toCompactDirs := make([]string, 0, len(toCompact))
-	for _, meta := range toCompact {
+	var eg errgroup.Group
+	for i, meta := range toCompact {
+		meta := meta
 		bdir := filepath.Join(dir, meta.ULID.String())
+		toCompactDirs[i] = bdir
 		for _, s := range meta.Compaction.Sources {
 			if _, ok := uniqueSources[s]; ok {
 				return false, ulid.ULID{}, halt(errors.Errorf("overlapping sources detected for plan %v", toCompact))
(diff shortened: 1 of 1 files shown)


pkg/compact/compact.go
```diff
@@ -724,40 +724,49 @@ func (cg *Group) compact(ctx context.Context, dir string, planner Planner, comp
...
 	begin := time.Now()
+	toCompactDirs := make([]string, len(toCompact))
 
-	toCompactDirs := make([]string, 0, len(toCompact))
-	for _, meta := range toCompact {
+	var eg errgroup.Group
+	for i, meta := range toCompact {
+		meta := meta
 		bdir := filepath.Join(dir, meta.ULID.String())
+		toCompactDirs[i] = bdir
```
squat, line 733: This line of code does not run concurrently, so we can turn this back into the more idiomatic style we tend to use
```suggestion
		toCompactDirs = append(toCompactDirs, bdir)
```

pkg/compact/compact.go
```diff
@@ -724,40 +724,49 @@ func (cg *Group) compact(ctx context.Context, dir string, planner Planner, comp
 
 	// Once we have a plan we need to download the actual data.
 	begin := time.Now()
+	toCompactDirs := make([]string, len(toCompact))
 
-	toCompactDirs := make([]string, 0, len(toCompact))
-	for _, meta := range toCompact {
+	var eg errgroup.Group
+	for i, meta := range toCompact {
```
squat, line 730: If we use the more idiomatic slice generation, then we can revert those change as well
```suggestion
	for _, meta := range toCompact {
```

squat (approved):
Looks good to me 👍 
Any idea if this helps? I wonder if your bottleneck is the network connection to object storage and thus parallelizing the downloads might not bring significant improvements.

yeya24:
> Looks good to me 👍
> Any idea if this helps? I wonder if your bottleneck is the network connection to object storage and thus parallelizing the downloads might not bring significant improvements.

I am trying to deploy this version and see if it helps. I agree that most time is spent on downloading blocks, but the verification step is also slow. See https://share.polarsignals.com/ad54156/

pkg/compact/compact.go
```diff
@@ -724,40 +724,49 @@ func (cg *Group) compact(ctx context.Context, dir string, planner Planner, comp
...
 	// Once we have a plan we need to download the actual data.
 	begin := time.Now()
-
 	toCompactDirs := make([]string, 0, len(toCompact))
+
+	var eg errgroup.Group
 	for _, meta := range toCompact {
+		meta := meta
 		bdir := filepath.Join(dir, meta.ULID.String())
+		toCompactDirs = append(toCompactDirs, bdir)
```
bwplotka, line 733: This is nice, I wonder only, do we always have enough resources (network, mem, CPU) for it to not fail / be slower overall?

bwplotka (approved):
I think it generally makes sense, but it might need more experimentation especially around mem usage. Nice!

yeya24:
Yes, resource might be a concern. Let's hold this pr and I need to do more testing.

pkg/compact/compact.go
yeya24: Now it spawns number of blocks gouroutines. We can use workers to limit the concurrency if the resource is a concern.

yeya24:
This version seems not very stable. The verification is faster but sometimes the pod hangs and I cannot find any error log. 
Not sure if this is caused by resources, but the CPU and memory usage is not high.
Need more testing and experiments.

PR #3686, opened 2020-12-30
Implement federated metric metadata API

yeya24:
Signed-off-by: Ben Ye <yb532204897@gmail.com>

* [ ] I added CHANGELOG entry for this change.
* [ ] Change is not relevant to the end user.

Fixes #3345 

This pr follows the awesome pr https://github.com/thanos-io/thanos/pull/3350 and comment https://github.com/thanos-io/thanos/issues/3345#issuecomment-737172268 to implement the Prometheus metric metadata API https://prometheus.io/docs/prometheus/latest/querying/api/#querying-metric-metadata.

## Verification

Tested manually and E2E tests added.

PR #4438, opened 2021-07-12
Add target metadata API

### LINKED ISSUE #3870: Implement target metadata API
As metric metadata API was merged https://github.com/thanos-io/thanos/pull/3686, it would be good to also support the Prometheus target metadata API (https://prometheus.io/docs/prometheus/latest/querying/api/#querying-target-metadata).

As the API response format is very similar to the previous metric metadata API, it would be not so difficult to support it. We can reuse the existing proto file as well.

PhilipGough:
This PR implements the target metadata API on the query component and fixes #3870.

The implementation largely follows the example set by the metric metadata PR in #3686 and in fact piggybacks on the service defined in the metadata proto that was introduced in that feature, since the two metadata APIs are closely related.

This avoids the need to add an additional flag to define metadata stores. This makes it more convenient but maybe less desirable (we could not take them out of experimental mode independently for example) and I am happy to introduce an additional flag and rpc service if required.

There is an open question on #3870 but I have followed @ianbillett suggestion and went with (Target, Metric) tuple metadata which essentially provides always on deduplication of metadata (a metric that is duplicated across targets will only be returned once from the API).

## Verification
I have added e2e tests

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -30,6 +30,7 @@ We use *breaking :warning:* to mark changes that are not backward compatible (re
 - [#3970](https://github.com/thanos-io/thanos/pull/3970) Azure: Adds more configuration options for Azure blob storage. This allows for pipeline and reader specific configuration. Implements HTTP transport configuration options. These options allows for more fine-grained control on timeouts and retries. Implements MSI authentication as second method of authentication via a service principal token.
 - [#4406](https://github.com/thanos-io/thanos/pull/4406) Tools: Add retention command for applying retention policy on the bucket.
 - [#4430](https://github.com/thanos-io/thanos/pull/4430) Compact: Add flag `downsample.concurrency` to specify the concurrency of downsampling blocks.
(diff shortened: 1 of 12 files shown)


ianbillett (changes requested):
Woof - there are a huge amount of changes in here! 

At a high level scanning the code, I think the changes you are making are the right ones, but this PR is currently too much of a beast for anyone to review thoroughly. 

I will ask that you try and split this PR up into smaller chunks that are easier for reviewers to review - that way you are much more likely to have someone review them thoroughly. 

Perhaps adding the proto definition and the generated code first (that accounts for most of the added lines), but leave them unused - then come back and raise a PR that actually uses this code?

PhilipGough:
@ianbillett thanks for the feedback. I have followed the format of #3686 and felt that the commits are fairly independently reviewable but totally understand the size of the PR is not ideal for reviewing.

Will be happy to split the proto and generated code into an initial PR if that is preferable.

Issue #4432, opened 2021-07-09
[Thanos Rule] Rules Files not reloading after SIGHUP signal


jumakasy:
**Thanos, Prometheus and Golang version used**:

thanos, version 0.21.1 (branch: HEAD, revision: 3558f4a5adc8d4068f82528570a439d440f98b9d)
  build user:       root@744cf7ef4576
  build date:       20210604-12:11:58
  go version:       go1.16.5
  platform:         linux/amd64

**What happened**:
After rule files update, a SIGHUP signal is sent to Thanos in order to reload the rules in run time. The rule files update is no applied by Thanos Rule, keeping applying the same rules as when Thanos was started. 

By restarting  the Thanos  process the rules are updated. 

**What you expected to happen**:
Thanos should reload the rules at runtime after receiving SIGHUP signal, with no need to stop and start the Thanos process.

**How to reproduce it (as minimally and precisely as possible)**:

Run Thanos Rule with minimal configuration and a basic and valid rule_test.yaml file: 

`` thanos rule --log.level debug --log.format logfmt --http-address 0.0.0.0:10902 --http-grace-period 2m --grpc-address 0.0.0.0:10901 --grpc-grace-period 2m  --data-dir ./data --rule-file './*.yml'  --resend-delay 1m --eval-interval 30s --tsdb.block-duration 2h --tsdb.retention 2d  --query thanos-query.domain :20902  ``

Access UI and check that rules defined in rule_test.yamls file are being applied.
Add a new rules file (rule__test_2.yaml) to the same folder.

Get Thanos proccess id: `thanos_pid=$(pgrep thanos)`
Reload Thanos process: `kill -1 $thanos_pid`

Acces the UI, and check the rules.

**Full logs to relevant components**:

After sending the SIGUHP this is the log:

`
level=info ts=2021-07-09T15:43:27.658638469Z caller=main.go:180 msg="caught signal. Reloading." signal=hangup
`
`
level=info ts=2021-07-09T15:41:04.904147954Z caller=main.go:183 msg="reload dispatched."
`
`
level=debug ts=2021-07-09T15:43:49.605131566Z caller=promclient.go:398 component=rules msg="querying instant" url="http://10.103.69.158:30902/api/v1/query?......
`

**Anything else we need to know**:

Tested other Thanos versions, and the last version where it worked properly was  v0.19.0.

Following the same steps with v0.19.0 the rules files are reloaded. The log is different, and its reporting that  rules files are being loaded:
`
[...]

PR #4442, opened 2021-07-13
rule: fix reload signal not working

jmichalek132:
Signed-off-by: Juraj Michalek <juraj.michalek132@gmail.com>

## Changes
Fixed a [bug](https://github.com/thanos-io/thanos/issues/4432) when sending sighup signal to thanos ruler didn't result in loading changes.

## Verification
1. Started thanos locally using `make quickstart`.
2. Modified recording rule in rules.yml created by the quickstart option
3. Sent sighup to thanos ruler
4. Looked at thanos ruler ui to check if changes were successfully loaded (they were)
5. Also added e2e test for it

### DIFF
--- a/test/e2e/rule_test.go
+++ b/test/e2e/rule_test.go
@@ -99,9 +99,6 @@ groups:
 `
 )
 
-
-
-
 type rulesResp struct {
 	Status string
 	Data   *rulespb.RuleGroups
--- a/test/e2e/rule_test.go
+++ b/test/e2e/rule_test.go
@@ -139,10 +139,9 @@ func checkReloadSuccessful(t *testing.T, ctx context.Context, endpoint string, e
 	resp, err := http.DefaultClient.Do(req)
 	testutil.Ok(t, err)
 	testutil.Equals(t, 200, resp.StatusCode)
-	defer resp.Body.Close()
 
 	body, err := ioutil.ReadAll(resp.Body)
-	testutil.Ok(t, err)
+	testutil.Ok(t, resp.Body.Close())
 
 	var data = rulesResp{}
 
--- a/test/e2e/rule_test.go
+++ b/test/e2e/rule_test.go
@@ -140,7 +140,7 @@ func checkReloadSuccessful(t *testing.T, ctx context.Context, endpoint string, e
 	testutil.Ok(t, err)
 	testutil.Equals(t, 200, resp.StatusCode)
 
-	body, err := ioutil.ReadAll(resp.Body)
+	body, _ := ioutil.ReadAll(resp.Body)
 	testutil.Ok(t, resp.Body.Close())
 
 	var data = rulesResp{}
--- a/test/e2e/rule_test.go
+++ b/test/e2e/rule_test.go
@@ -58,7 +58,7 @@ groups:
     annotations:
       summary: "I always complain and allow partial response in query."
 `
-	testAlertRuleAddedLaterWebHandler = `
+	testAlertRuleAddedLater = `
 groups:
 - name: example
   interval: 1s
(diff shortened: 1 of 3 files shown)


GiedriusS:
Thanks for the PR! :muscle: Maybe we could add a test for this? Because this functionality was working, then we broke it, and now this fixes it again. Having a test would make sure that this works. What do you think? :thinking: The seems like a good fit for the e2e tests however you will probably have to improve the e2e framework for Cortex to add a `KillSignal(signal int) error` interface or something like that.

jmichalek132:
Yeah I was thinking adding the test would make sense too. Should I do it as part of this pr or could it be a follow up pr? Since making changes necessary in cortex might take some time.

I was also thinking exposing api to exec commands inside the containers started by the cortex e2e framework might make a bit more sense since it could be useful in other cases not just this one.

jmichalek132:
So it seems the e2e services already provides `Exec` which can be used to send the signal, now I just need to figure test that the reload actually worked.

jmichalek132:
I added the test for reloading using sighup. Currently I am checking that there 3 rule groups (there are 2 before the reload, one is added by the reload). Is the test okay like this?

test/e2e/rule_test.go
```diff
@@ -317,11 +345,16 @@ func TestRule(t *testing.T) {
...
+		// Add a new rule via sending sighup
+		createRuleFile(t, fmt.Sprintf("%s/newrule.yaml", rulesPath), testAlertRuleAddedLater)
+		reloadRulesSignal(t, r)
+		checkReloadSuccessful(t, ctx, r.HTTPEndpoint())
+	})
 
-		createRuleFile(t, fmt.Sprintf("%s/newrule.yaml", rulesPath), testAlertRuleAddedLaterWebHandler)
+	t.Run("http reload works", func(t *testing.T) {
+		// Add a new rule via /-/reload.
+		createRuleFile(t, fmt.Sprintf("%s/newrule.yaml", rulesPath), testAlertRuleAddedLater)
```
GiedriusS, line 357: The test later on here creates a file with the same name & content. Maybe we could have one `testAlertRuleAddedLater` with 3 rule groups & another one with 4 rule groups, let's say? And then use the one with 4 rule groups here. After refactoring `checkReloadSuccessful` to accept the expected rule group count, we could use that function here as well.

test/e2e/rule_test.go
```diff
@@ -97,6 +97,34 @@ func reloadRulesHTTP(t *testing.T, ctx context.Context, endpoint string) {
...
+
+	var data struct {
+		Status string
+		Data   *rulespb.RuleGroups
+	}
+
+	testutil.Ok(t, json.Unmarshal(body, &data))
+	testutil.Equals(t, "success", data.Status)
+
+	testutil.Assert(t, len(data.Data.Groups) == 3, "expected there to be 3 rule groups")
```
GiedriusS, line 125: Then we could put this `3` as a parameter in the function

test/e2e/rule_test.go
```diff
@@ -97,6 +97,34 @@ func reloadRulesHTTP(t *testing.T, ctx context.Context, endpoint string) {
...
+	testutil.Ok(t, err)
+}
+
+func checkReloadSuccessful(t *testing.T, ctx context.Context, endpoint string) {
+	req, err := http.NewRequestWithContext(ctx, "GET", "http://"+endpoint+"/api/v1/rules", ioutil.NopCloser(bytes.NewReader(nil)))
+	testutil.Ok(t, err)
+	resp, err := http.DefaultClient.Do(req)
+	testutil.Ok(t, err)
+	testutil.Equals(t, 200, resp.StatusCode)
+	defer resp.Body.Close()
```
GiedriusS, line 112: Let's just do: `testutil.Ok(t, resp.Body.Close())`

jmichalek132:
Not sure why the unit tests failed but I don't think it's caused by my change. Could we re-run them?

test/e2e/rule_test.go
```diff
@@ -97,6 +97,34 @@ func reloadRulesHTTP(t *testing.T, ctx context.Context, endpoint string) {
...
+
+	body, err := ioutil.ReadAll(resp.Body)
+	testutil.Ok(t, err)
+
+	var data struct {
```
GiedriusS, line 117: We already have the same struct in this file - maybe let's re-use the same struct in two different functions? We could have:
```
type rulesResp struct {
  Status string
  Data *rulespb.RuleGroups
}
```

GiedriusS (commented):
Just a few nits but overall LGTM :muscle: I hope my suggestion is clear here, feel free to ask any questions :smile:

test/e2e/rule_test.go
jmichalek132: addressed

test/e2e/rule_test.go
jmichalek132: Done in the last commit.

test/e2e/rule_test.go
jmichalek132: Could you check the last commit if understood correctly what you suggested please?

test/e2e/rule_test.go
GiedriusS: @jmichalek132 could you please simply do `testutil.Ok(t, resp.Body.Close())` after `ioutil.ReadAll()`? It will be simpler + you'll get a failure if `Close()` fails

GiedriusS (dismissed):
Awesome work! Thanks :heart: could you please do the last small change with regards to the `defer resp.Body.Close()`?

test/e2e/rule_test.go
jmichalek132: Like this?

GiedriusS (approved):

Issue #3112, opened 2020-09-02
Block viewer: New UI extension ideas


bwplotka:
It looks amazing (zoom out view 75%):

![image](https://user-images.githubusercontent.com/6950331/91969717-736bef80-ed0e-11ea-8ffb-b44f11f0a82d.png)

And works well, but some minor nits that come to my mind if someone has spare cycles:

*  [x] It takes to much vertical space, I think I would reduce vertical whitespace to a minimum. 
*  [ ] It would be amazing to just group them by all external labels (no config needed?)
*  [ ] Show different statuses as different colors / details on global view
*  [ ] Vertical lines for some grid like on old UI?
*  [ ] Stats on top: when was the last refresh, how many blocks overall, how many blocks in each group (?)
*  [ ] Search input that will allow to group them on certain labels? 
*  [ ] It would be amazing if we can someday add extra stats like (for loaded page): 
   * What blocks are currently being compacted, downsampled! (: 
* [ ] Using different colors is ok ;p (same as we use for coloring graphs)

cc @prmsrswt

kunal-kushwaha:
I'd like to take this 🙌

daksh-sagar:
@bwplotka are we talking about the vertical whitespace that is present between the blocks like in the first row in the above image?

bwplotka:
Are you talking about first action item? We would like to reduce it to a minimum. First row blocks show some overlapping blocks that's why they are close. We need to make sure it's closer but not too much to show that those are for different external labels (:

daksh-sagar:
I've made some changes, this is what it looks like right now. What do you think @bwplotka ?
![blocks](https://user-images.githubusercontent.com/24277216/94094912-4955ab00-fe3e-11ea-8e64-895ce8ddbb2b.png)
PS - I also texted this to you on Slack yesterday 😁

bwplotka:
It looks better! :muscle: Looking forward for a PR (:

daksh-sagar:
I also asked for Prem for his feedback, he suggested reducing the height of the blocks. I've done that, this is what it looks like. Let me know which one is better and I'll do it accordingly.
![blocks_new](https://user-images.githubusercontent.com/24277216/94127417-85a4fd80-fe76-11ea-8fd8-c1e31c9d7e06.png)

daksh-sagar:
@prmsrswt Need help regarding 5th action item: "display stats on top".  How should it be displayed (visually) ?

daksh-sagar:
[...]

PR #4444, opened 2021-07-14
ui: Add search block UI

adzshaf:

## Changes

This PR is part of LFX Mentorship project (issue: #3112). We want to add Search Block by ID.

## Verification
When we haven't search anything yet:

image

Example search block by ID:

image

Tested on local using `make quickstart`

### DIFF
--- a/pkg/ui/react-app/src/thanos/pages/blocks/BlockInput.tsx
+++ b/pkg/ui/react-app/src/thanos/pages/blocks/BlockInput.tsx
@@ -1,0 +1,30 @@
+import React, { FC, ChangeEvent } from 'react';
+import { Button, InputGroup, InputGroupAddon, InputGroupText, Input } from 'reactstrap';
+import { FontAwesomeIcon } from '@fortawesome/react-fontawesome';
+import { faSearch } from '@fortawesome/free-solid-svg-icons';
+import styles from './blocks.module.css';
+
+interface BlockInputProps {
+  onChange: ({ target }: ChangeEvent<HTMLInputElement>) => void;
+  onClick: () => void;
+}
+
+export const BlockInput: FC<BlockInputProps> = ({ onChange, onClick }) => {
+  return (
+    <>
+      <InputGroup className={styles.blockInput}>
+        <InputGroupAddon addonType="prepend">
+          <InputGroupText>
+            <FontAwesomeIcon icon={faSearch} />
+          </InputGroupText>
+        </InputGroupAddon>
+        <Input placeholder="Search block by ulid" onChange={onChange} />
+        <InputGroupAddon addonType="append">
+          <Button className="execute-btn" color="primary" onClick={onClick}>
+            Search
+          </Button>
+        </InputGroupAddon>
+      </InputGroup>
+    </>
+  );
+};
--- a/pkg/ui/react-app/src/thanos/pages/blocks/Blocks.tsx
+++ b/pkg/ui/react-app/src/thanos/pages/blocks/Blocks.tsx
@@ -1,4 +1,4 @@
-import React, { FC, useMemo, useState } from 'react';
+import React, { ChangeEvent, FC, useMemo, useState } from 'react';
 import { RouteComponentProps } from '@reach/router';
 import { UncontrolledAlert } from 'reactstrap';
 import { useQueryParams, withDefault, NumberParam } from 'use-query-params';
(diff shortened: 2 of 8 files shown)


adzshaf:
Should we add query param so it is easier for user to reference it? e.g `/loaded?id=<BLOCK_ID>` @onprem @squat

yeya24 (commented):
Love it! This is super useful for us. Would be good to support more filter conditions in the future (by external labels, compaction levels, etc)

onprem:
> Should we add query param so it is easier for user to reference it? e.g `/loaded?id=<BLOCK_ID>` @onprem @squat

Yep, that sounds like a nice idea!

pkg/ui/react-app/src/thanos/pages/blocks/BlockInput.tsx
```diff
@@ -0,0 +1,30 @@
...
+import { FontAwesomeIcon } from '@fortawesome/react-fontawesome';
+import { faSearch } from '@fortawesome/free-solid-svg-icons';
+import styles from './blocks.module.css';
+
+interface BlockInputProps {
+  onChange: ({ target }: ChangeEvent<HTMLInputElement>) => void;
+  onClick: () => void;
+}
+
+export const BlockInput: FC<BlockInputProps> = ({ onChange, onClick }) => {
```
onprem, line 12:
```suggestion
export const BlockSearchInput: FC<BlockInputProps> = ({ onChange, onClick }) => {
```

pkg/ui/react-app/src/thanos/pages/blocks/Blocks.tsx
```diff
@@ -58,6 +60,10 @@ export const BlocksContent: FC<{ data: BlockListProps }> = ({ data }) => {
 
   return (
     <>
+      <BlockInput
+        onChange={({ target }: ChangeEvent<HTMLInputElement>): void => setSearchState(target.value)}
+        onClick={() => setBlockSearch(searchState)}
+      />
```
onprem, line 66: This hsould be inside the ternanry on line 67

pkg/ui/react-app/src/thanos/pages/blocks/BlockInput.tsx
```diff
@@ -0,0 +1,30 @@
...
+            <FontAwesomeIcon icon={faSearch} />
+          </InputGroupText>
+        </InputGroupAddon>
+        <Input placeholder="Search block by ulid" onChange={onChange} />
+        <InputGroupAddon addonType="append">
+          <Button className="execute-btn" color="primary" onClick={onClick}>
+            Search
+          </Button>
+        </InputGroupAddon>
+      </InputGroup>
```
onprem, line 27: Let's put these inside a `<form>` tag and use the `onSubmit` handler so pressing enter to search works.

pkg/ui/react-app/src/thanos/pages/blocks/helpers.ts
```diff
@@ -103,3 +104,17 @@ export const download = (blob: Block): string => {
 
   return url;
 };
+
+export const getBlockByUlid = (blocks: Block[], ulid: string): string => {
```
onprem, line 108: Instead of returning the most relevant ulid, let's return a block array, with all the matching blocks.

onprem (commented):
Some nits and a few suggestions. Looking good!

pkg/ui/react-app/src/thanos/pages/blocks/helpers.ts
```diff
@@ -103,3 +104,19 @@ export const download = (blob: Block): string => {
...
+
+  const ulidArray = blocks.map((block) => block.ulid);
+  const fuz = new Fuzzy({ caseSensitive: true });
+
+  const result: FuzzyResult[] = fuz.filter(ulid, ulidArray);
+
+  const resultIndex = result.map((value) => value.index);
+
+  const blockResult = blocks.filter((block, index) => resultIndex.includes(index));
+  return blockResult;
```
squat, line 121: I spoke with @adzshaf and mentioned that I think this could produce confusing experience for the user. If I provide full-length ULID that matches a block name exactly, but fuzzy search returns more than one result, then I would be very confused and asume that all of the blocks have the same name, which should not really be happening. I think it's fair to return multiple blocks if the ULID does not match any block exactly, i.e. it is not full-length or has a typo.

We can do this in a follow up of course, since this UI already provides a ton of value.

squat (approved):
lgtm!

squat (approved):
LGTM-ing again, this force push just rebases on main

squat:
actually! this PR should definitely have a changelog entry! It provides a great new feature for the user

PR #4448, opened 2021-07-16
loadMeta: Reduce the number of queries with object stores

wuming333666:
Signed-off-by: chenjianwei666888@gmail.com

## Changes

loadMeta with cache,Reduce the number of queries with object stores

### DIFF
--- a/pkg/block/fetcher.go
+++ b/pkg/block/fetcher.go
@@ -227,6 +227,9 @@ func (f *BaseFetcher) loadMeta(ctx context.Context, id ulid.ULID) (*metadata.Met
 		metaFile       = path.Join(id.String(), MetaFilename)
 		cachedBlockDir = filepath.Join(f.cacheDir, id.String())
 	)
+	if m, seen := f.cached[id]; seen {
+		return m, nil
+	}
 
 	// TODO(bwplotka): If that causes problems (obj store rate limits), add longer ttl to cached items.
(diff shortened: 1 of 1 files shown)


pkg/block/fetcher.go
```diff
@@ -227,6 +227,10 @@ func (f *BaseFetcher) loadMeta(ctx context.Context, id ulid.ULID) (*metadata.Met
 		metaFile       = path.Join(id.String(), MetaFilename)
 		cachedBlockDir = filepath.Join(f.cacheDir, id.String())
 	)
+	// Reduced query object storage
+	if m, seen := f.cached[id]; seen {
```
yeya24, line 231: Not sure if this is correct. How do you handle consistency and metadata deletion? 
For example, you are running store gateway and compactor. For example, one metadata is cached in memory but the block is deleted later by the compactor. In this case, the result is wrong because it always returns the metadata in memory and never checks the status in the object storage.

wuming333666: I'm not sure I'm understanding this correctly, I see that the delete is done last when the compact operation is performed

yeya24: `BaseFetcher` is used by multiple components, including compactor, store gateway, etc. So this case happens for store gateway.
Even for the compactor, the compactor doesn't update the fetcher's cache after it deletes the block because the underlying baseFetcher object is the same.

PR #4449, opened 2021-07-16
query: Extend metadata service to include target metadata API

PhilipGough:
As requested in the review of #4438, this change extends the metadata service to include target metadata api https://prometheus.io/docs/prometheus/latest/querying/api/#querying-target-metadata.

The implementation largely follows the example set by the metric metadata PR in #3686 and in fact piggybacks on the service defined in the metadata proto that was introduced in that feature, since the two metadata APIs are closely related.

This will avoid the need to add an additional flag to define metadata stores. This makes it more convenient but maybe less desirable (we could not take them out of experimental mode independently for example) and I am happy to introduce an additional flag and rpc service if required.

## Changes
Extend the metadata rpc service

## Verification

None required, updates proto and generated code

### DIFF
--- a/pkg/info/infopb/rpc.proto
+++ b/pkg/info/infopb/rpc.proto
@@ -58,8 +58,12 @@ message StoreInfo {
 message RulesInfo {
 }
 
-/// MetricMetadataInfo holds the metadata related to Metadata API exposed by the component.
+/// MetricMetadataInfo holds the metadata related to Metric Metadata API exposed by the component.
 message MetricMetadataInfo {
+}
+
+/// TargetMetadataInfo holds the metadata related to Target Metadata API exposed by the component.
+message TargetMetadataInfo {
 }
 
 /// TargetsInfo holds the metadata related to Targets API exposed by the component.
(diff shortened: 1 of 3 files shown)


pkg/info/infopb/rpc.proto
```diff
@@ -58,10 +58,14 @@ message StoreInfo {
 message RulesInfo {
 }
 
-/// MetricMetadataInfo holds the metadata related to Metadata API exposed by the component.
+/// MetricMetadataInfo holds the metadata related to Metric Metadata API exposed by the component.
 message MetricMetadataInfo {
 }
 
+/// TargetMetadataInfo holds the metadata related to Target Metadata API exposed by the component.
+message TargetMetadataInfo {
```
yeya24, line 66: I am wondering is it better to combine this and `MetricMetadataInfo` to one `MetadataInfo` field, in the future?
As they are both belonging to `metadatapb`, but different services.

PhilipGough: good point, resolved

PR #4451, opened 2021-07-19
Upgrade automaxprocs to fix debug log format issue when set env GOMAX…

hanjm:

## Desc

when set env GOMAXPROCS (for some container runtime) and debug log level. thanos will output log like
```
$ GOMAXPROCS=1 thanos query --log.level=debug
level=debug ts=2021-07-19T09:09:29.536577Z caller=main.go:103 msg="maxprocs: Honoring GOMAXPROCS=[%!d(string=1)] as set in environment"
```

it is a bug which already fixed by https://github.com/uber-go/automaxprocs/pull/17, we should upgrade it.

## Changes
upgrade automaxprocs

## Verification 

after upgrade automaxprocs
```
$ GOMAXPROCS=1 ./thanos query --log.level=debug
level=debug ts=2021-07-19T11:33:08.827793Z caller=main.go:64 msg="maxprocs: Honoring GOMAXPROCS=[\"1\"] as set in environment"
```


GiedriusS (approved):
Awesome! :+1:

PR #4454, opened 2021-07-19
Add RequestID for Deletion Requests

ilangofman:
Signed-off-by: ilangofman <igofman99@gmail.com>

Hello, I mentioned this briefly in one of the community calls. Currently I am working on the adding the series deletion API in Cortex. To accomplish the task of permanently deleting the series, I would like to leverage the logic of the bucket rewrite work done inside of Thanos. To track which blocks have been rewritten, the proposed idea was to store the deletion request id inside of the `meta.json` file. Since the file is written from within Thanos, this PR will allow us to do this.  

If anyone is intrested, the proposal for the API in Cortex can be found [here](https://cortexmetrics.io/docs/proposals/block-storage-time-series-deletion/l).

Please let me know if there is any questions or concerns with this change.

Thanks

## Changes

- Add a new variable inside the deletion request struct to store the deletion request id (could be empty).


bwplotka (approved):
LGTM, thanks!

Sorry for lag in review.

PR #4458, opened 2021-07-20
Thanos planner component proposal

andrejbranch:


jeromeinsf:
Looks related to https://github.com/cortexproject/cortex/pull/4272 
Might be worth aligning the 2

docs/proposals-accepted/202107-planner-component.md
```diff
@@ -0,0 +1,73 @@
...
+
+Separate compactor's planner into its own component and have compactor workers request available plans from the planner.
+
+## Why
+
+Currently, the Thanos compactor handles planning which blocks to compact as well as vertical/horizontal compaction of the groups. This causes issues with the scalability of compaction for a single tenant. The actual compaction process is relatively simple, as all the process needs to know is what blocks to compact. By contrast, the planning logic is relatively complex and currently can't coordinate with other compactors. To get around the coordination issue, currently the compactor can be sharded by tenant. This will not horizontally scale. When a single tenant becomes very large it will need to distribute the workload across many compaction processes across many instances, regardless of how the workload is divided.
+
+## Goals
+
+* To separate planner into its own component and have compactors communicate with the planner to get the next available plan.
```
bwplotka, line 21: I think the goals are behind this one. Would be nice outline our motivation here. I think it's prerquisite to scale horizontally compactor, especially within same tenant/external labels... no? 🤗 - which is essentially breaking from singleton model

docs/proposals-accepted/202107-planner-component.md
```diff
@@ -0,0 +1,73 @@
...
+message CompactorMessage {
+  CompactorMessageType type = 1
+  CompactionPlan = 2
+}
+
+message PlannerMessage {
+  CompactionPlan = 1
+}
+```
+
```
bwplotka, line 65: Can we mention some alternatives? 

I can think of one: Coordination free gossip-based planning and compaction

docs/proposals-accepted/202107-planner-component.md
```diff
@@ -0,0 +1,73 @@
...
+* To separate planner into its own component and have compactors communicate with the planner to get the next available plan.
+
+## Non-Goals
+
+* Horizontally scaling compaction for a single tenant will be addressed following the completion of this proposal
+
+## Proposal
+![where](../img/thanos_proposal_planner_component.png)
+
+### Separate planner into its own component.
```
bwplotka, line 30:
```suggestion
### Step 1: Separate planner into its own component.
```

docs/proposals-accepted/202107-planner-component.md
```diff
@@ -0,0 +1,73 @@
...
+For the initial implementation a reasonable amount of the current planner code could be reused. One difference is the new planner will need to run plans for all tenants. The planner should maintain a priority queue of available plans with fair ordering between tenants. The planner should also be aware of which plans are currently running and be able to handle the failure of a compaction plan gracefully. After the completion of this proposal, planner can be updated to improve single tenant compaction performance, following the goals of the [compaction group concurrency pull request](https://github.com/thanos-io/thanos/pull/3807).
+
+### GRPC pubsub (bidirectional streaming)
```
bwplotka, line 33:
```suggestion
### Step 2: GRPC pubsub (bidirectional streaming)
```

docs/proposals-accepted/202107-planner-component.md
```diff
@@ -0,0 +1,73 @@
...
+
+## Goals
+
+* To separate planner into its own component and have compactors communicate with the planner to get the next available plan.
+
+## Non-Goals
+
+* Horizontally scaling compaction for a single tenant will be addressed following the completion of this proposal
+
+## Proposal
```
bwplotka, line 27: Can we have some `### Risks` section where we can explain what we do to ensure Planner is not Single point of failure? (HA / Scalability)

bwplotka (commented):
Amazing work! Thanks for this, sorry for late review - was on PTO

@thanos-io/thanos-maintainers PTAL 🤗 

Some initial comments, otherwise it looks good. Let's indeed check Cortex plan - we are touching same part and looks like we want to solve same problem

ianbillett (commented):
Nice proposal :+1: As a community, agreeing upon an approach to refactoring and improving the compactor is long overdue.

Personally, I don't think implementing this as a separate Thanos component is the right approach. Critically, we would break existing users when they upgraded. Preferably, we would follow the pattern set by `receive`.

By default it will run as a single instance, but can also be configured in separate `routing` and `ingesting` modes for advanced users. Scaling up compactor would be an advanced feature, so we could reasonably expect the user to understand the differences and be able to configure it accordingly.

The default behaviour would be to run this work in a configurable number of local worker goroutines so vertically scaling would bring performance benefits, but the advanced feature would be to offload to remote workers so they could horizontally scale workloads with an [HPA](https://kubernetes.io/docs/tasks/run-application/horizontal-pod-autoscale/).

With this in mind, I think there are two sub-problems we should solve in order:
1. Make compactor scale vertically on one node by paralellizing compactions.
2. Make compactor offload compaction tasks to workers.

docs/proposals-accepted/202107-planner-component.md
```diff
@@ -0,0 +1,73 @@
...
+![where](../img/thanos_proposal_planner_component.png)
+
+### Separate planner into its own component.
+For the initial implementation a reasonable amount of the current planner code could be reused. One difference is the new planner will need to run plans for all tenants. The planner should maintain a priority queue of available plans with fair ordering between tenants. The planner should also be aware of which plans are currently running and be able to handle the failure of a compaction plan gracefully. After the completion of this proposal, planner can be updated to improve single tenant compaction performance, following the goals of the [compaction group concurrency pull request](https://github.com/thanos-io/thanos/pull/3807).
```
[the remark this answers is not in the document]
bwplotka: I think we can do that on later stages - there are many ideas we want to do (:

yeya24:
It would be nice to revisit this one someday.
This is very similar to https://rockset.com/blog/remote-compactions-in-rocksdb-cloud/.
We definitely want to make planner a separate component. And there are multiple ways to implement the workers:
1. Have separate compaction workers to listen on some compaction plan RPCs
2. For k8s specific environments, workers can be k8s jobs that download blocks, compact them and upload the final block
3. ...

roystchiang:
Is it possible for me to continue this work?

I would like the planner part to be compatible with Cortex, so that both systems can enjoy the benefit it brings. While we flesh out whether a dedicated planner component is required, I believe both projects can already enjoy some improvement if the compactor is able to spit out parallelizable plans.

bwplotka:
Go for it @roystchiang ! Looks like it went stale. Feel free to take @andrejbranch work and open your own PR, we can close this one 🤗 Thanks! We definitely need help on this!

PR #4462, opened 2021-07-21
ui: Add find overlap block UI

adzshaf:

## Changes

This PR is part of LFX Mentorship project to improve Block Viewer UI (issue: #3112). We want to add feature so user can find overlapping blocks easily.

## Verification
When we haven't enable find overlapping block yet:
image

Example `enable find overlapping block`:

image

Tested on local using `make quickstart`

cc: @onprem @squat

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -9,6 +9,7 @@ NOTE: As semantic versioning states all 0.y.z releases can contain breaking chan
 We use *breaking :warning:* to mark changes that are not backward compatible (relates only to v0.y.z releases.)
 
 ## Unreleased
+- [#4462](https://github.com/thanos-io/thanos/pull/4462) UI: Add find overlap block UI
 
 ### Added
 - [#4453](https://github.com/thanos-io/thanos/pull/4453) Tools: Add flag `--selector.relabel-config-file` / `--selector.relabel-config` / `--max-time` / `--min-time` to filter served blocks.
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -9,7 +9,6 @@ NOTE: As semantic versioning states all 0.y.z releases can contain breaking chan
 We use *breaking :warning:* to mark changes that are not backward compatible (relates only to v0.y.z releases.)
 
 ## Unreleased
-- [#4462](https://github.com/thanos-io/thanos/pull/4462) UI: Add find overlap block UI
 
 ### Added
 - [#4453](https://github.com/thanos-io/thanos/pull/4453) Tools: Add flag `--selector.relabel-config-file` / `--selector.relabel-config` / `--max-time` / `--min-time` to filter served blocks.
@@ -17,6 +16,7 @@ We use *breaking :warning:* to mark changes that are not backward compatible (re
 - [#4487](https://github.com/thanos-io/thanos/pull/4487) Query: Add memcached auto discovery support.
 - [#4444](https://github.com/thanos-io/thanos/pull/4444) UI: Add search block UI.
 - [#4509](https://github.com/thanos-io/thanos/pull/4509) Logging: Adds duration_ms in int64 to the logs.
+- [#4462](https://github.com/thanos-io/thanos/pull/4462) UI: Add find overlap block UI
 
 ### Fixed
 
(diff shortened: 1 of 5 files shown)


pkg/ui/react-app/src/thanos/pages/blocks/helpers.ts
```diff
@@ -103,3 +103,28 @@ export const download = (blob: Block): string => {
 
   return url;
 };
+
+export const getOverlappingBlocks = (blockPools: { [source: string]: BlocksPool }): Set<string> => {
+  let result: Set<string> = new Set();
+
+  Object.keys(blockPools).forEach((pk) => {
```
GiedriusS, line 110: :scream: this seems quite complex. Have you tried seeing if there is a more effective algorithm for this?

pkg/ui/react-app/src/thanos/pages/blocks/helpers.ts
```diff
@@ -103,3 +103,28 @@ export const download = (blob: Block): string => {
...
+    let blocks = Object.values(blockPools[pk]);
+    for (let i = 0; i < blocks.length; i++) {
+      if (blocks[i].length <= 1) {
+        continue;
+      }
+      for (let j = 0; j < blocks[i].length - 1; j++) {
+        for (let k = 0; k < blocks[i][j].length; k++) {
+          for (let l = 0; l < blocks[i][j + 1].length; l++) {
+            if (isOverlapping(blocks[i][j][k], blocks[i][j + 1][l])) {
```
squat, line 119: the computational complexity is not thaaaat bad. The fact that everything is so nested is because the data structure is so complex, but it's not soooo far off from O(N). It's definitely not O(N^2), as that would mean comparing every block in every row in every block pool to every other block. We are (only) comparing every block within one row of a single block pool to all of the other blocks in the same row of the same block pool, which is much less. Maybe we can still simplify this.

pkg/ui/react-app/src/thanos/pages/blocks/helpers.ts
```diff
@@ -103,3 +103,28 @@ export const download = (blob: Block): string => {
...
+
+  Object.keys(blockPools).forEach((pk) => {
+    let blocks = Object.values(blockPools[pk]);
+    for (let i = 0; i < blocks.length; i++) {
+      if (blocks[i].length <= 1) {
+        continue;
+      }
+      for (let j = 0; j < blocks[i].length - 1; j++) {
+        for (let k = 0; k < blocks[i][j].length; k++) {
+          for (let l = 0; l < blocks[i][j + 1].length; l++) {
```
squat, line 118: Can we really assume that we only need to compare with the next row? Or do we need to compare against all blocks in all other rows?

bwplotka:
Nice work @adzshaf , can't wait to see this in action! 💜

pkg/ui/react-app/src/thanos/pages/blocks/SourceView.tsx
```diff
@@ -11,9 +11,11 @@ export const BlocksRow: FC<{
 }> = ({ blocks, gridMinTime, gridMaxTime, selectBlock }) => {
   return (
     <div className={styles.row}>
-      {blocks.map<JSX.Element>((b) => (
-        <BlockSpan selectBlock={selectBlock} block={b} gridMaxTime={gridMaxTime} gridMinTime={gridMinTime} key={b.ulid} />
-      ))}
+      {blocks.map<JSX.Element>((b) => {
+        return (
```
squat, line 15: do we need this change?

adzshaf:
@squat @onprem  I'll add query params like `/loaded?find-overlapping=True` so it will be consistent with other filters, what do you think?

squat:
> @squat @onprem I'll add query params like `/loaded?find-overlapping=True` so it will be consistent with other filters, what do you think?

@adzshaf Yes please! That would be very important for consistency and ease of sharing

CHANGELOG.md
```diff
@@ -9,6 +9,7 @@ NOTE: As semantic versioning states all 0.y.z releases can contain breaking chan
 We use *breaking :warning:* to mark changes that are not backward compatible (relates only to v0.y.z releases.)
 
 ## Unreleased
+- [#4462](https://github.com/thanos-io/thanos/pull/4462) UI: Add find overlap block UI
```
squat, line 12: Plz move this down to the `### Added` section

squat (commented):
Overall looks great! Two small suggestions from my side and it's good :)

adzshaf:
@squat ready to be merged 👍

squat (approved):
🎉 great work!

yeya24 (approved):
LGTM

PR #4465, opened 2021-07-21
Fix typo and add a useful alternative GOPROXY

Hangzhi:
Signed-off-by: Hangzhi <1454678938@qq.com>

Fix some minor typos and add an alternative GOPROXY. This GOPROXY is for developers in mainland China who cannot connect the official ` https://proxy.golang.org `.

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -91,7 +91,7 @@ It's key to get familiarized with the style guide and mechanics of Thanos, espec
 
     export GOPATH="~/Repos/thanosgopath" # Use if you want to have an isolated directory for deps, otherwise, the directory where you have installed golang will be used.
 
-    export GOPROXY="https://proxy.golang.org"  
+    export GOPROXY="https://proxy.golang.org"
     export PATH="$GOBIN:$PATH"
     ```
 
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -1,6 +1,6 @@
 # Contributing
 
-This document explain the process of contributing to the Thanos project.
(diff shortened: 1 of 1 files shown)


CONTRIBUTING.md
```diff
@@ -84,14 +84,14 @@ It's key to get familiarized with the style guide and mechanics of Thanos, espec
 
 * Find any directory in your system your want Thanos repo in. e.g `~/Repos` -
   * `cd ~/Repos`
-  * Make sure that the GOBIN, GOPATH and GOPROXY (useful) environment variables are set and that GOBIN is included in your PATH. For example -
+  * Make sure that the GOBIN, GOPATH and GOPROXY (useful) environment variables are set and that GOBIN is included in your PATH. You may use `GOPROXY=https://goproxy.cn,direct` as an alternative if you cannot visit `https://proxy.golang.org`. For example -
```
Namanl2001, line 87: I'm able to access `https://goproxy.cn/` in India but not `https://goproxy.cn,direct` is there a typo? 🤔  If not then it would be nice to mention it explicitly that this alternative is only for China.

Namanl2001 (approved):
Just a small suggestion else LGTM 👍 . Thanks!

CONTRIBUTING.md
Hangzhi: `GOPROXY=direct` is to configure the go command to download all modules directly from source repositories, set GOPROXY to direct. In this case, direct will take effect when we cannot access the proxy. This is optional for developers. I'll remove the direct to avoid misunderstanding.

GiedriusS (approved):
Thanks!

PR #4468, opened 2021-07-22
Rule: Fix temporary rule filename composition issue

### LINKED ISSUE #4232: Ruler: skips rule files based on their names without any error/warning log.
**Thanos version used**:

thanos image: quay.io/thanos/thanos:v0.17.2

**What happened**:

We run thanos-rule with the following flag:
```
--rule-file=/var/thanos/rules/*.yaml
```
to load rules from all the yaml files we put under the above location.
We use `kustomize` `configMapGenerator` to create a configMap with the needed files:
```
configMapGenerator:
  - name: alerts
    files:
      - accounting.yaml=resources/alerts/accounting.yaml
      - billing.yaml=resources/alerts/billing.yaml
      - cbc.yaml=resources/alerts/cbc.yaml
      - partner-data.yaml=resources/alerts/partner-data.yaml
      - partner.yaml=resources/alerts/partner.yaml
      - payments.yaml=resources/alerts/payments.yaml
      - prometheus.yaml=resources/alerts/prometheus.yaml
      - unicom.yaml=resources/alerts/unicom.yaml
```
and mount it under the passed location:
```
        volumeMounts:
        - mountPath: /var/thanos/rules
          name: rules
      volumes:
      - configMap:
          defaultMode: 420
          name: alerts-845c2ttgdg
        name: rules
```
As a result, we can see all the rule files mounted in the pods:
```
$ kubectl --context=dev-merit --namespace=sys-mon exec -ti thanos-rule-56f565b58b-7lvnw -- ls /var/thanos/rules
accounting.yaml    partner-data.yaml  prometheus.yaml
billing.yaml       partner.yaml       unicom.yaml
cbc.yaml           payments.yaml

```

Moreover, thanos ruler is able to see all 8 files on startup:
```
level=info ts=2021-05-11T16:15:45.776401996Z caller=rule.go:836 component=rules msg="reload rule files" numFiles=8
```

Thanos only loads rules from 6/8 files. In particular, the following:
```
/var/thanos/rules/accounting.yaml
/var/thanos/rules/billing.yaml
/var/thanos/rules/cbc.yaml
/var/thanos/rules/payments.yaml
/var/thanos/rules/prometheus.yaml
/var/thanos/rules/unicom.yaml
```

The end result is that we are missing rules defined in 2 yaml files.

Load all the rules, or give an error on why it failed to do so.

Following our example should be sufficient here.

No useful logs from thanos regarding this one. We were profiling the issue via trying different sets of our rule files.

A workaround is to prefix all mounted files with `_`, which seems to do the trick and allows the ruler to load all our rules fine.

In particular, we believe that this behaviour is a result of: https://github.com/thanos-io/thanos/blob/f4a5904b0c3be7565298cf257ad43ac1942021cc/pkg/rules/manager.go#L358 and the way file names are trimmed to be written into tmp dir.

**Environment**:
Kube version: v1.21.0 
OS: Flatcar Container Linux by Kinvolk 2765.2.3 (Oklo)
kernel: 5.10.32-flatcar
container runtime: containerd://1.4.4

matej-g:

## Changes

Provides a fix for current issue with rule filenames, which are being written into a `.tmp-rules` directory for further processing. As was discovered in the linked issue, the current implementation could lead to temporary rule filenames being composed in a way, where trimming of the filename leads to unexpected result, i.e. the `TrimLeft` function can in some cases trim the filename in a way which results in duplication of temporary rule filenames. This in turn manifests itself as some rules being 'ignored' by the rule component.

Since trimming the filename does not seem to have the desired effect, the fix proposed here is to simply leave the full filename as is, which practically leads to same result but without possible name duplication issue.

Fixes #4232

## Verification
Related tests have been updated to assess the fix is working; also tested manually with local run of `thanos rule`


wiardvanrij (approved):
If you could rebase this from main, that would be great. LGTM, thank you

GiedriusS (approved):

GiedriusS:
LGTM, I have fixed the conflicts!

GiedriusS:
Seems Sonatype Lift is having problems, merging

GiedriusS (approved):

GiedriusS (approved):

Issue #3442, opened 2020-11-13
How to fix Compactor out-of-order chunks?


lorenzo-biava:
**Thanos, Prometheus and Golang version used**:

thanos, version 0.15.0 (branch: HEAD, revision: fbd14b49f95e7543883fedf9381a21127d4dda2b)
  build user:       circleci@b18149728583
  build date:       20200907-09:47:14
  go version:       go1.14.2

**Object Storage Provider**:

Azure Blob

**What happened**:

The compactor is halted during compaction due to an "out-of-order chunks" error.

**What you expected to happen**:

Find a way to fix the block in some way.
I run the `tools bucket verify --repair` command, hoping to get something repaired but nothing happened.
Looking at the doc, there is no indication of how fix this error; even if it has to be done manually, I see no link to any procedure or tool to help with that.
[Only overlaps are mentioned in the doc](https://github.com/thanos-io/thanos/blob/master/docs/operating/troubleshooting.md#overlaps)
Looking at previous issues, the only possibly relevant one I've found is https://github.com/thanos-io/thanos/issues/267 but does not contain any guidance on how to fix it.

**How to reproduce it (as minimally and precisely as possible)**:

Not sure what went wrong there; I can provide a link to the faulty block if needed.

**Full logs to relevant components**:

<details>Logs
<p>

```
level=info ts=2020-11-12T14:58:41.326584314Z caller=http.go:57 service=http/server component=compact msg="listening for requests and metrics" address=0.0.0.0:10902
level=info ts=2020-11-12T14:58:41.429273647Z caller=compact.go:944 msg="start sync of metas"
level=info ts=2020-11-12T14:58:41.869377033Z caller=fetcher.go:453 component=block.BaseFetcher msg="successfully synchronized block metadata" duration=542.828221ms cached=96 returned=96 partial=0
level=info ts=2020-11-12T14:58:41.910604952Z caller=fetcher.go:453 component=block.BaseFetcher msg="successfully synchronized block metadata" duration=37.305665ms cached=96 returned=96 partial=0
level=info ts=2020-11-12T14:58:42.547706875Z caller=fetcher.go:453 component=block.BaseFetcher msg="successfully synchronized block metadata" duration=1.118392926s cached=96 returned=66 partial=0
level=info ts=2020-11-12T14:58:42.548690014Z caller=compact.go:949 msg="start of GC"
level=info ts=2020-11-12T14:58:42.548906222Z caller=compact.go:961 msg="start of compactions"
[...]

PR #4469, opened 2021-07-22
skip blocks with out-of-order chunk during compaction

huyan0:
Signed-off-by: Yang Hu <yhuz@amazon.com>

## Changes
- add a check for OOO chunks in block index, and mark such block for no compaction, and continue on with the compaction
- add testutil to generate an index with OOO chunks

## Verification
- verified the exepected behavior with `compact_e2e_test.go`

Addresses: #3442 
cc: @yeya24 @alvinlin123 @roystchiang

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -17,6 +17,7 @@ We use *breaking :warning:* to mark changes that are not backward compatible (re
 - [#4444](https://github.com/thanos-io/thanos/pull/4444) UI: Add search block UI.
 - [#4509](https://github.com/thanos-io/thanos/pull/4509) Logging: Adds duration_ms in int64 to the logs.
 - [#4462](https://github.com/thanos-io/thanos/pull/4462) UI: Add find overlap block UI
+- [#4469](https://github.com/thanos-io/thanos/pull/4469) Compact: Add flag `compact.skip-block-with-out-of-order-chunks` to skip blocks with out-of-order chunks during compaction instead of halting
 
 ### Fixed
 
--- a/cmd/thanos/compact.go
+++ b/cmd/thanos/compact.go
@@ -147,9 +147,10 @@ func newCompactMetrics(reg *prometheus.Registry, deleteDelay time.Duration) *com
 	m.blocksMarked = promauto.With(reg).NewCounterVec(prometheus.CounterOpts{
 		Name: "thanos_compact_blocks_marked_total",
 		Help: "Total number of blocks marked in compactor.",
-	}, []string{"marker"})
-	m.blocksMarked.WithLabelValues(metadata.NoCompactMarkFilename)
-	m.blocksMarked.WithLabelValues(metadata.DeletionMarkFilename)
+	}, []string{"marker", "reason"})
+	m.blocksMarked.WithLabelValues(metadata.NoCompactMarkFilename, metadata.OutOfOrderChunksNoCompactReason)
+	m.blocksMarked.WithLabelValues(metadata.NoCompactMarkFilename, metadata.IndexSizeExceedingNoCompactReason)
+	m.blocksMarked.WithLabelValues(metadata.DeletionMarkFilename, "")
 
 	m.garbageCollectedBlocks = promauto.With(reg).NewCounter(prometheus.CounterOpts{
 		Name: "thanos_compact_garbage_collected_blocks_total",
@@ -281,7 +282,7 @@ func runCompact(
 			cf,
 			duplicateBlocksFilter,
 			ignoreDeletionMarkFilter,
-			compactMetrics.blocksMarked.WithLabelValues(metadata.DeletionMarkFilename),
+			compactMetrics.blocksMarked.WithLabelValues(metadata.DeletionMarkFilename, ""),
 			compactMetrics.garbageCollectedBlocks,
 			conf.blockSyncConcurrency)
 		if err != nil {
@@ -347,15 +348,17 @@ func runCompact(
 		conf.acceptMalformedIndex,
 		enableVerticalCompaction,
 		reg,
-		compactMetrics.blocksMarked.WithLabelValues(metadata.DeletionMarkFilename),
+		compactMetrics.blocksMarked.WithLabelValues(metadata.DeletionMarkFilename, ""),
 		compactMetrics.garbageCollectedBlocks,
+		compactMetrics.blocksMarked.WithLabelValues(metadata.NoCompactMarkFilename, metadata.OutOfOrderChunksNoCompactReason),
 		metadata.HashFunc(conf.hashFunc),
+		conf.skipBlockWithOutOfOrderChunks,
 	)
 	planner := compact.WithLargeTotalIndexSizeFilter(
 		compact.NewPlanner(logger, levels, noCompactMarkerFilter),
 		bkt,
 		int64(conf.maxBlockIndexSize),
-		compactMetrics.blocksMarked.WithLabelValues(metadata.DeletionMarkFilename),
+		compactMetrics.blocksMarked.WithLabelValues(metadata.NoCompactMarkFilename, metadata.IndexSizeExceedingNoCompactReason),
 	)
 	blocksCleaner := compact.NewBlocksCleaner(logger, bkt, ignoreDeletionMarkFilter, deleteDelay, compactMetrics.blocksCleaned, compactMetrics.blockCleanupFailures)
 	compactor, err := compact.NewBucketCompactor(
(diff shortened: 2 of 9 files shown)


cmd/thanos/compact.go
```diff
@@ -349,6 +349,7 @@ func runCompact(
 		reg,
 		compactMetrics.blocksMarked.WithLabelValues(metadata.DeletionMarkFilename),
 		compactMetrics.garbageCollectedBlocks,
+		compactMetrics.blocksMarked.WithLabelValues(metadata.NoCompactMarkFilename),
```
yeya24, line 352: I am wondering if we need a new `reason` label for this mark counter. Now we cannot differentiate any compact blocks due to large index or out of order chunks.

yeya24 (commented):
Love it! Super high-quality code. Thanks for the contribution @huyan0

pkg/block/index.go
```diff
@@ -111,12 +111,9 @@ func (i HealthStats) Issue347OutsideChunksErr() error {
 	return nil
 }
 
-// CriticalErr returns error if stats indicates critical block issue, that might solved only by manual repair procedure.
-func (i HealthStats) CriticalErr() error {
-	var errMsg []string
-
-	if i.OutOfOrderSeries > 0 {
-		errMsg = append(errMsg, fmt.Sprintf(
+func (i HealthStats) OutOfOrderChunksErr() error {
```
alvinlin123, line 114: I am wondering, do you think we can just skip compaction for all the CriticalErr; instead of just OutOfOrderChunksErr?

huyan0: I think it's okay on the compact side given we cannot generate more blocks than not compacting...I don't think there's much implication on the retrieval side either but would like to hear what @yeya24 thinks : )
Could you also provide some context on why the project decided to halt compaction in the first place?

cmd/thanos/compact.go
huyan0: sounds reasonable, especially if we want to skip all blocks with critical error as @alvinlin123 suggested

huyan0:
I have updated the metric counter, will add to changelog if there's no further comments for now

yeya24 (dismissed):
This pr LGTM. But it changes the default behavior and might be critical so it would be good to hear from other maintainers.

wiardvanrij:
The code looks solid and you referenced an issue. If I understand correctly, we now mark it as corrupted and skip it. That still leaves a block corrupted. Are there any follow-up steps that a user could do? Basically I would really love to have some documentation on:
- How could one find corrupted blocks?
- What is the impact for the user if that happens?
- Could one still fix a corrupted block and let it get compacted after the fact?
- anything else that provides context to users

This would IMO be the cherry on top :)

bwplotka:
TODO: Can we check if those blocks are *readable* by Thanos Store/Cortex Store

bwplotka:
TODO: Review & check for consistency with other issues and how we handle them (:

huyan0:
@bwplotka  regarding your two comments:

1. the blocks are still readable because the mark that we add to the block is for compact only, and the store API has no requirement on whether the chunks are sorted: [link](https://github.com/thanos-io/thanos/blob/aa148f8fdb281a9ea4c1e5a7ee2e1336d8459c4f/pkg/store/storepb/rpc.pb.go#L703)

2. We handle this error differently than the other errors encountered during compaction; for those we halt the compaction. Other than OOO chunks, the two other specific errors compact check for [are these](https://github.com/thanos-io/thanos/blob/aa148f8fdb281a9ea4c1e5a7ee2e1336d8459c4f/pkg/compact/compact.go#L752) , for which we attempt to repair or compacts if a feature flag is set, so I don't see an overarching consistent approach to adhere to, other than logging the occurrence to provide context for user. This is understandable since these errors are unexpected and specific. I agree a deep dive into the root cause is needed to address this in long term to come up with a more coherent fix.

pkg/block/index.go
huyan0: @bwplotka  do you have some context on the discussion? :)

huyan0:
PR still pending review :) Any updates needed?

cmd/thanos/compact.go
```diff
@@ -668,6 +671,9 @@ func (cc *compactConfig) registerFlag(cmd extkingpin.FlagClause) {
 		"Default is due to https://github.com/thanos-io/thanos/issues/1424, but it's overall recommended to keeps block size to some reasonable size.").
 		Hidden().Default("64GB").BytesVar(&cc.maxBlockIndexSize)
 
+	cmd.Flag("compact.skip-block-with-out-of-order-index", "When set to true, mark blocks containing index with out-of-order chunks for no compact instead of halting the compaction").
```
yeya24, line 674: What about `compact.skip-block-with-out-of-order-chunks`?

yeya24:
Btw let's update the changelog and we are ready to go I think

yeya24:
@huyan0 Sorry for the late review. I will merge the pr after you resolve the conflict.

yeya24 (approved):
LGTM

cmd/thanos/compact.go
```diff
@@ -147,8 +147,8 @@ func newCompactMetrics(reg *prometheus.Registry, deleteDelay time.Duration) *com
 	m.blocksMarked = promauto.With(reg).NewCounterVec(prometheus.CounterOpts{
 		Name: "thanos_compact_blocks_marked_total",
 		Help: "Total number of blocks marked in compactor.",
-	}, []string{"marker"})
-	m.blocksMarked.WithLabelValues(metadata.NoCompactMarkFilename)
+	}, []string{"marker", "reason"})
+	m.blocksMarked.WithLabelValues(metadata.NoCompactMarkFilename, metadata.OutOfOrderChunksNoCompactReason)
```
yeya24, line 151: Let's add other reasons for no compact marker.

cmd/thanos/compact.go
```diff
@@ -147,8 +147,8 @@ func newCompactMetrics(reg *prometheus.Registry, deleteDelay time.Duration) *com
...
 	m.blocksMarked.WithLabelValues(metadata.DeletionMarkFilename)
```
yeya24, line 152: Panic here. We need a reason for deletion marker. Empty string is fine here.

cmd/thanos/compact.go
```diff
@@ -349,13 +349,15 @@ func runCompact(
 		reg,
 		compactMetrics.blocksMarked.WithLabelValues(metadata.DeletionMarkFilename),
```
yeya24, line 350: Panic again here.

pkg/compact/compact.go
```diff
@@ -939,6 +978,21 @@ func (c *BucketCompactor) Compact(ctx context.Context) (rerr error) {
 							continue
 						}
 					}
+					// if block has out of order chunk, mark the block for no copmaction and continue
```
yeya24, line 981: typo `copmaction` and please add a period here

yeya24 (changes requested):
There might be other panics caused by that deletion marker metric. Please check and fix them. Thanks for the patience.

yeya24:
Can you do a rebase on latest main rather than including changes from another pr.

CHANGELOG.md
```diff
@@ -16,6 +16,8 @@ We use *breaking :warning:* to mark changes that are not backward compatible (re
 - [#4487](https://github.com/thanos-io/thanos/pull/4487) Query: Add memcached auto discovery support.
 - [#4444](https://github.com/thanos-io/thanos/pull/4444) UI: Add search block UI.
 - [#4509](https://github.com/thanos-io/thanos/pull/4509) Logging: Adds duration_ms in int64 to the logs.
+- [#4462](https://github.com/thanos-io/thanos/pull/4462) UI: Add find overlap block UI
+- [#4469](https://github.com/thanos-io/thanos/pull/4482) Compact: Add flag `compact.skip-block-with-out-of-order-chunks` to skip blocks with out-of-order chunks during compaction instead of halting
```
yeya24, line 20: The pr number seems incorrect in the link.

pkg/compact/compact.go
```diff
@@ -541,6 +555,27 @@ func IsIssue347Error(err error) bool {
 	return ok
 }
 
+// OutOfOrderChunkError is a type wrapper for OOO chunk error from validating block index.
+type OutOfOrderChunksError struct {
+	err error
+
```
yeya24, line 561: Let's remove this line

pkg/testutil/testutil.go
```diff
@@ -203,3 +211,106 @@ func FaultOrPanicToErr(f func()) (err error) {
...
+var indexFilename = "index"
+
+type indexWriterSeries struct {
+	labels labels.Labels
+	chunks []chunks.Meta // series file offset of chunks
+}
+
+type indexWriterSeriesSlice []*indexWriterSeries
+
+// putOutOfOrderIndex updates the index in blockDir with an index containing an out-of-order chunk
```
yeya24, line 224: uppercase

pkg/testutil/testutil.go
```diff
@@ -203,3 +211,106 @@ func FaultOrPanicToErr(f func()) (err error) {
...
+				valset = map[string]struct{}{}
+				values[l.Name] = valset
+			}
+			valset[l.Value] = struct{}{}
+		}
+		postings.Add(uint64(i), s.labels)
+	}
+
+	err = iw.Close()
+	return err
```
yeya24, line 315: return iw.Close()

yeya24 (approved):
LGTM

huyan0:
looking... Sorry for the back and forth. Will try to address it

yeya24 (approved):
LGTM! Thanks for the contribution.

PR #4486, opened 2021-07-26
Add Joblift to adopter

galan:
Signed-off-by: Daniel Galán y Martins <git@galan.de>

## Changes
We are using Thanos for nearly a year now, also in use on production, and would be proud to show this in your adopters list.


GiedriusS (approved):

GiedriusS:
Thanks! :heart:

Issue #3879, opened 2021-03-05
Give guidance that memcached should not be configured on loadbalancer level


wiardvanrij:
The 
```
config:
  addresses: []
```
Can be used in various ways. It should be clear that each memcached instance should be defined, rather than a 'main' instance E.g. "loadbalancer" adres. This is because the client does not do autodiscovery (Correct me if I'm wrong): https://docs.aws.amazon.com/AmazonElastiCache/latest/mem-ug/AutoDiscovery.HowAutoDiscoveryWorks.html

# Plan

I will update the documentation to make this clear

kakkoyun:
AFAIK you can use DNS service discovery or you can specify all. i.e https://github.com/thanos-io/thanos/blob/befb0257e8a3d49cb80b00794a4892abfb7ccb72/pkg/cacheutil/memcached_client.go#L81
Kubernetes e.g https://github.com/observatorium/deployments/blob/258c17c7889351cc7366a5c3b6162ad216d666d0/components/thanos.libsonnet#L241

@pstibrany @pracucci would know better.

But yes, let's improve documentation on this.

wiardvanrij:
Right, thanks @kakkoyun . I will include that as well.  It's important to either use such DNS discovery or define the single targets yourself (in case of for example AWS). If you only define the k8s svc, you are going to have a bad time :)

PR #4487, opened 2021-07-26
query: add memcached autodiscovery support

roystchiang:
## Changes
* Add a flag to enable auto-discovery for memcached
* Add a new provider/resolver for auto-discovery

## Verification
* Integrated into cortex, and verified that auto-discovery is able to find scaled up nodes without having to restart pods
* Tested with [AWS ElastiCache Auto Discovery](https://docs.aws.amazon.com/AmazonElastiCache/latest/mem-ug/AutoDiscovery.HowAutoDiscoveryWorks.html), but should work with [Google Cloud ](https://cloud.google.com/memorystore/docs/memcached/auto-discovery-overview)too, since the protocol seem the same

I also updated the memcached config documentation so that we can close https://github.com/thanos-io/thanos/issues/3879

### DIFF
--- a/docs/components/store.md
+++ b/docs/components/store.md
@@ -295,7 +295,7 @@ config:
 
 The **required** settings are:
 
-- `addresses`: list of memcached addresses, that will get resolved with the [DNS service discovery](../service-discovery.md/#dns-service-discovery) provider. If your cluster supports auto-discovery, you should use the flag `auto_discovery` instead.
+- `addresses`: list of memcached addresses, that will get resolved with the [DNS service discovery](../service-discovery.md/#dns-service-discovery) provider. If your cluster supports auto-discovery, you should use the flag `auto_discovery` instead and only point to *one of* the memcached servers. This typically means that there should be only one address specified that resolves to any of the alive memcached servers. Use this for Amazon ElastiCache and other similar services.
 
 While the remaining settings are **optional**:
 
--- a/docs/components/store.md
+++ b/docs/components/store.md
@@ -290,11 +290,12 @@ config:
   max_item_size: 0
   max_get_multi_batch_size: 0
   dns_provider_update_interval: 0s
+  auto_discovery:false
 ```
 
 The **required** settings are:
 
-- `addresses`: list of memcached addresses, that will get resolved with the [DNS service discovery](../service-discovery.md/#dns-service-discovery) provider.
+- `addresses`: list of memcached addresses, that will get resolved with the [DNS service discovery](../service-discovery.md/#dns-service-discovery) provider. If your cluster supports auto-discovery, you should use the flag `auto_discovery` instead.
 
 While the remaining settings are **optional**:
 
@@ -306,6 +307,7 @@ While the remaining settings are **optional**:
 - `max_get_multi_batch_size`: maximum number of keys a single underlying operation should fetch. If more keys are specified, internally keys are splitted into multiple batches and fetched concurrently, honoring `max_get_multi_concurrency`. If set to `0`, the batch size is unlimited.
 - `max_item_size`: maximum size of an item to be stored in memcached. This option should be set to the same value of memcached `-I` flag (defaults to 1MB) in order to avoid wasting network round trips to store items larger than the max item size allowed in memcached. If set to `0`, the item size is unlimited.
 - `dns_provider_update_interval`: the DNS discovery update interval.
+- `auto_discovery`: whether to use the auto-discovery mechanism for memcached.
 
 ## Caching Bucket
 
--- a/docs/components/store.md
+++ b/docs/components/store.md
@@ -290,7 +290,7 @@ config:
   max_item_size: 0
   max_get_multi_batch_size: 0
   dns_provider_update_interval: 0s
-  auto_discovery:false
+  auto_discovery: false
 ```
 
 The **required** settings are:
--- a/pkg/cacheutil/memcached_client.go
+++ b/pkg/cacheutil/memcached_client.go
@@ -21,6 +21,7 @@ import (
 	"gopkg.in/yaml.v2"
 
 	"github.com/thanos-io/thanos/pkg/discovery/dns"
+	memcacheDiscovery "github.com/thanos-io/thanos/pkg/discovery/memcache"
 	"github.com/thanos-io/thanos/pkg/extprom"
 	"github.com/thanos-io/thanos/pkg/gate"
 	"github.com/thanos-io/thanos/pkg/model"
@@ -53,6 +54,7 @@ var (
 		MaxGetMultiConcurrency:    100,
 		MaxGetMultiBatchSize:      0,
 		DNSProviderUpdateInterval: 10 * time.Second,
+		AutoDiscovery:             false,
 	}
 )
 
@@ -114,6 +116,9 @@ type MemcachedClientConfig struct {
 
 	// DNSProviderUpdateInterval specifies the DNS discovery update interval.
 	DNSProviderUpdateInterval time.Duration `yaml:"dns_provider_update_interval"`
+
+	// AutoDiscovery configures memached client to perform auto-discovery instead of DNS resolution
+	AutoDiscovery bool `yaml:"auto_discovery"`
 }
 
 func (c *MemcachedClientConfig) validate() error {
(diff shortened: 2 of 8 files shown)


GiedriusS (commented):
Please rebase on latest `main`, that should fix CircleCI

GiedriusS:
@roystchiang could you please fix the linter messages? :hugs:

roystchiang:
oops. I've fixed the linter messages now.

pkg/discovery/memcache/provider_test.go
```diff
@@ -0,0 +1,88 @@
...
+	sort.Strings(addresses)
+
+	testutil.Ok(t, err)
+	testutil.Equals(t, []string{"dns-1:11211", "dns-2:8080"}, addresses)
+
+	resolver.configs = map[string]*ClusterConfig{
+		"memcached-cluster-1": {nodes: []Node{{dns: "dns-1", ip: "ip-1", port: 11211}, {dns: "dns-3", ip: "ip-3", port: 11211}}},
+		"memcached-cluster-2": {nodes: []Node{{dns: "dns-2", ip: "ip-2", port: 8080}}},
+	}
+	err = provider.Resolve(ctx, clusters)
```
GiedriusS, line 41: Nit: I suggest just doing `testutil.Ok(t, provider.Resolve(ctx, clusters))`

pkg/discovery/memcache/provider_test.go
```diff
@@ -0,0 +1,88 @@
...
+	provider := NewProvider(log.NewNopLogger(), nil, 5*time.Second)
+	resolver := mockResolver{
+		configs: map[string]*ClusterConfig{
+			"memcached-cluster-1": {nodes: []Node{{dns: "dns-1", ip: "ip-1", port: 11211}}},
+			"memcached-cluster-2": {nodes: []Node{{dns: "dns-2", ip: "ip-2", port: 8080}}},
+		},
+	}
+	provider.resolver = &resolver
+
+	err := provider.Resolve(ctx, clusters)
```
GiedriusS, line 61: Same here

pkg/discovery/memcache/provider_test.go
```diff
@@ -0,0 +1,88 @@
...
+	addresses := provider.Addresses()
+	sort.Strings(addresses)
+
+	testutil.Ok(t, err)
+	testutil.Equals(t, []string{"dns-1:11211", "dns-2:8080"}, addresses)
+
+	resolver.configs = nil
+	resolver.err = errors.New("oops")
+	err = provider.Resolve(ctx, clusters)
```
GiedriusS, line 70: And so on :smile:

pkg/discovery/memcache/resolver.go
```diff
@@ -0,0 +1,111 @@
...
+	dialTimeout time.Duration
+}
+
+func (s *memcachedAutoDiscovery) Resolve(ctx context.Context, address string) (config *ClusterConfig, err error) {
+	conn, err := net.DialTimeout("tcp", address, s.dialTimeout)
+	if err != nil {
+		return nil, err
+	}
+	defer func() {
+		err = conn.Close()
```
GiedriusS, line 41: If `err` is not nil then this nonetheless overrides the value of it, right? I suggest doing this here:
```
defer runutil.CloseWithErrCapture(&err, conn, "closing connection")
```

pkg/cacheutil/memcached_client.go
```diff
@@ -177,6 +182,11 @@ type memcachedClient struct {
 	dataSize   *prometheus.HistogramVec
 }
 
+type AddressProvider interface {
```
GiedriusS, line 185: Could you please provide a comment for this interface:
```
// AddressProvider is ...
```

Also, `dns.Provider` should implement it, right? We could write: `var _ AddressProvider = (*Provider)(nil)`

pkg/discovery/memcache/provider.go
```diff
@@ -0,0 +1,107 @@
...
+
+	"github.com/go-kit/kit/log"
+	"github.com/go-kit/kit/log/level"
+	"github.com/prometheus/client_golang/prometheus"
+	"github.com/prometheus/client_golang/prometheus/promauto"
+	"github.com/thanos-io/thanos/pkg/errutil"
+	"github.com/thanos-io/thanos/pkg/extprom"
+)
+
+type Provider struct {
```
GiedriusS, line 20: Please add a comment for this.

pkg/discovery/memcache/resolver.go
```diff
@@ -0,0 +1,111 @@
...
+	"bufio"
+	"context"
+	"fmt"
+	"net"
+	"strconv"
+	"strings"
+	"time"
+)
+
+type ClusterConfig struct {
```
GiedriusS, line 16: Do these two types need to be exported if they have only private members?

pkg/cacheutil/memcached_client.go
```diff
@@ -220,20 +230,30 @@ func newMemcachedClient(
...
-		dns.GolangResolverType,
-	)
+	promRegisterer := extprom.WrapRegistererWithPrefix("thanos_memcached_", reg)
+
+	var addressProvider AddressProvider
+	if config.AutoDiscovery {
+		addressProvider = memcacheDiscovery.NewProvider(
+			logger,
+			promRegisterer,
+			2*time.Second)
```
GiedriusS, line 240: I suppose we can go ahead with this for now

GiedriusS (commented):
A few nits but overall good work :+1:

pkg/cacheutil/memcached_client.go
roystchiang: ahh. I meant to set it with the current configurable timeout. thanks :)

pkg/discovery/memcache/resolver.go
roystchiang: thanks for the pro-tip

docs/components/store.md
```diff
@@ -290,11 +290,12 @@ config:
   max_item_size: 0
   max_get_multi_batch_size: 0
   dns_provider_update_interval: 0s
+  auto_discovery: false
 ```
 
 The **required** settings are:
 
-- `addresses`: list of memcached addresses, that will get resolved with the [DNS service discovery](../service-discovery.md/#dns-service-discovery) provider.
+- `addresses`: list of memcached addresses, that will get resolved with the [DNS service discovery](../service-discovery.md/#dns-service-discovery) provider. If your cluster supports auto-discovery, you should use the flag `auto_discovery` instead.
```
GiedriusS, line 298: Could we please expand this section a bit:
```suggestion
- `addresses`: list of memcached addresses, that will get resolved with the [DNS service discovery](../service-discovery.md/#dns-service-discovery) provider. If your cluster supports auto-discovery, you should use the flag `auto_discovery` instead and only point to _one of_ the memcached servers. This typically means that there should be only one address specified that resolves to any of the alive memcached servers. Use this for Amazon ElastiCache and other similar services.
```
? Does this make sense?

GiedriusS (approved):

GiedriusS:
@roystchiang please add the DCO and let's merge 🤗

GiedriusS:
@roystchiang make docs now fails :/

roystchiang:
hahahahahaha. yeah I'm on it. sorry about that

pkg/discovery/memcache/provider.go
```diff
@@ -0,0 +1,111 @@
...
+
+	p.clusterConfigs = clusterConfigs
+
+	return errs.Err()
+}
+
+// Addresses returns the latest addresses present in the Provider.
+func (p *Provider) Addresses() []string {
+	var result []string
+	for _, config := range p.clusterConfigs {
```
GiedriusS, line 105: Upon my final reading of this code I've noticed that this is missing a lock:
```
p.RLock()
defer p.RUnlock()
```

In practice it doesn't matter whether this exists because we call one method after the other but could you please add this lock/unlock for consistency? I swear it's my last comment 😂

GiedriusS (dismissed):

pkg/discovery/memcache/provider.go
roystchiang: thanks for taking a careful look :)

really appreciate it.

GiedriusS (approved):

PR #4493, opened 2021-07-28
Automatic certificate rotation without restart

Namanl2001:
This PR implements the second part of this [proposal](https://github.com/thanos-io/thanos/blob/main/docs/proposals-accepted/202106-automated-per-endpoint-mTLS.md), i.e. 

> Adding a new `--grpc-server-max-connection-age` CLI option which takes time as input and controls how often to re-do the tls handshake and re-read the certificates. This in reality is controlled by the keepalive’s MaxConnectionAge option (which will close the connection after every (e.g. 15m) inputted time duration) so when the connection is closed it requires a new tls handshake.
While reading the cert_file on each handshake we can improve the performance by keeping track of it’s modification time. We should re-load the certificates only if they are actually rotated and there is some change in their modification time otherwise returning the previously stored one. Also if the file is not modified recently then it could be found on the cache memory of the system running thanos and it would be faster to get the file details (i.e. modification time).

### Changes
Added new `--grpc-server-max-connection-age` flag which takes "time" as input and controls how often to re-read the tls certificates (BOTH client and server certs) and redo the TLS handshake.

### Verification

Added e2e tests.

### DIFF
--- a/cmd/thanos/query.go
+++ b/cmd/thanos/query.go
@@ -58,7 +58,7 @@ func registerQuery(app *extkingpin.App) {
 	cmd := app.Command(comp.String(), "Query node exposing PromQL enabled Query API with data retrieved from multiple store nodes.")
 
 	httpBindAddr, httpGracePeriod, httpTLSConfig := extkingpin.RegisterHTTPFlags(cmd)
-	grpcBindAddr, grpcGracePeriod, grpcCert, grpcKey, grpcClientCA := extkingpin.RegisterGRPCFlags(cmd)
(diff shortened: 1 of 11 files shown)


kakkoyun (approved):
This looks good to me. Simple and easy to reason about. Thanks a lot for the tests as well.
One thing we could consider is to utilize the reloader that we already have which is used to reload Prometheus instances when config changes. It requires minor refactoring to be more generic.

Let's wait for @bwplotka review as well and we can merge it.

bwplotka (approved):
I love it, good job, thanks! 💪🏽

PR #4508, opened 2021-08-02
Adjust and rename `ThanosSidecarUnhealthy` to `ThanosSidecarNoConnectionToStartedPrometheus`; Remove `ThanosSidecarPrometheusDown` alert; Remove unused `thanos_sidecar_last_heartbeat_success_time_seconds` metrics

arajkumar:
Prior to this fix, `ThanosSidecarUnhealthy` would fire even when
Prometheus is busy with WAL replay. This would trigger a false positive alert.

This PR considers `prometheus_tsdb_data_replay_duration_seconds` metric from
Prometheus for `ThanosSidecarUnhealthy` alert. In order to correlate
Thanos and Prometheus metrics we need to specify common label(s) which
can be configured through `thanosPrometheusCommonDimensions` jsonnet
variable.

This PR also removes `ThanosSidecarPrometheusDown` as it would fire at the same as `ThanosSidecarUnhealthy`.

Rename `ThanosSidecarUnhealthy` to `ThanosSidecarNoConnectionToStartedPrometheus`.

Remove unused `thanos_sidecar_last_heartbeat_success_time_seconds` metric.

Fixes https://github.com/thanos-io/thanos/issues/3915.

Signed-off-by: Arunprasad Rajkumar <arajkuma@redhat.com>

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -36,6 +36,7 @@ We use *breaking :warning:* to mark changes that are not backward compatible (re
 - [#4476](https://github.com/thanos-io/thanos/pull/4476) UI: fix incorrect html escape sequence used for '>' symbol.
 - [#4532](https://github.com/thanos-io/thanos/pull/4532) Mixin: Fixed "all jobs" selector in thanos mixin dashboards.
 - [#4607](https://github.com/thanos-io/thanos/pull/4607) Azure: Fix Azure MSI Rate Limit
+- [#4508](https://github.com/thanos-io/thanos/pull/4508) Adjust and rename `ThanosSidecarUnhealthy` to `ThanosSidecarNoConnectionToStartedPrometheus`; Remove `ThanosSidecarPrometheusDown`.
 
 ### Changed
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -13,6 +13,10 @@ We use *breaking :warning:* to mark changes that are not backward compatible (re
 ### Added
 - [#4680](https://github.com/thanos-io/thanos/pull/4680) Query: add `exemplar.partial-response` flag to control partial response.
 
+### Fixed
+
+- [#4508](https://github.com/thanos-io/thanos/pull/4508) Adjust and rename `ThanosSidecarUnhealthy` to `ThanosSidecarNoConnectionToStartedPrometheus`; Remove `ThanosSidecarPrometheusDown` alert; Remove unused `thanos_sidecar_last_heartbeat_success_time_seconds` metrics.
+
 ## v0.23.0 - In Progress
 
 ### Added
@@ -36,7 +40,6 @@ We use *breaking :warning:* to mark changes that are not backward compatible (re
 - [#4476](https://github.com/thanos-io/thanos/pull/4476) UI: fix incorrect html escape sequence used for '>' symbol.
 - [#4532](https://github.com/thanos-io/thanos/pull/4532) Mixin: Fixed "all jobs" selector in thanos mixin dashboards.
 - [#4607](https://github.com/thanos-io/thanos/pull/4607) Azure: Fix Azure MSI Rate Limit
-- [#4508](https://github.com/thanos-io/thanos/pull/4508) Adjust and rename `ThanosSidecarUnhealthy` to `ThanosSidecarNoConnectionToStartedPrometheus`; Remove `ThanosSidecarPrometheusDown`.
 
 ### Changed
 
--- a/cmd/thanos/sidecar.go
+++ b/cmd/thanos/sidecar.go
@@ -138,10 +138,6 @@ func runSidecar(
 			Name: "thanos_sidecar_prometheus_up",
 			Help: "Boolean indicator whether the sidecar can reach its Prometheus peer.",
 		})
-		lastHeartbeat := promauto.With(reg).NewGauge(prometheus.GaugeOpts{
-			Name: "thanos_sidecar_last_heartbeat_success_time_seconds",
-			Help: "Timestamp of the last successful heartbeat in seconds.",
-		})
 
 		ctx, cancel := context.WithCancel(context.Background())
 		g.Add(func() error {
@@ -191,7 +187,6 @@ func runSidecar(
 				)
 				promUp.Set(1)
 				statusProber.Ready()
-				lastHeartbeat.SetToCurrentTime()
 				return nil
 			})
 			if err != nil {
@@ -213,7 +208,6 @@ func runSidecar(
 					promUp.Set(0)
 				} else {
 					promUp.Set(1)
-					lastHeartbeat.SetToCurrentTime()
 				}
 
 				return nil
--- a/examples/alerts/alerts.yaml
+++ b/examples/alerts/alerts.yaml
@@ -301,16 +301,6 @@ groups:
       severity: warning
 - name: thanos-sidecar
   rules:
-  - alert: ThanosSidecarPrometheusDown
-    annotations:
-      description: Thanos Sidecar {{$labels.instance}} cannot connect to Prometheus.
-      runbook_url: https://github.com/thanos-io/thanos/tree/main/mixin/runbook.md#alert-name-thanossidecarprometheusdown
-      summary: Thanos Sidecar cannot connect to Prometheus
-    expr: |
-      thanos_sidecar_prometheus_up{job=~".*thanos-sidecar.*"} == 0
-    for: 5m
-    labels:
-      severity: critical
   - alert: ThanosSidecarBucketOperationsFailed
     annotations:
       description: Thanos Sidecar {{$labels.instance}} bucket operations are failing
@@ -326,9 +316,13 @@ groups:
       description: Thanos Sidecar {{$labels.instance}} is unhealthy for more than
         {{$value}} seconds.
       runbook_url: https://github.com/thanos-io/thanos/tree/main/mixin/runbook.md#alert-name-thanossidecarunhealthy
-      summary: Thanos Sidecar is unhealthy.
+      summary: Thanos Sidecar cannot access Prometheus, even though Prometheus seems
+        healthy and has reloaded WAL.
     expr: |
-      time() - max by (job, instance) (thanos_sidecar_last_heartbeat_success_time_seconds{job=~".*thanos-sidecar.*"}) >= 240
+      time() - max by (pod, job, instance) (thanos_sidecar_last_heartbeat_success_time_seconds{job=~".*thanos-sidecar.*"}) >= 240
+      AND on (pod) (
+      min by (pod) (prometheus_tsdb_data_replay_duration_seconds) != 0
+      )
     for: 5m
     labels:
       severity: critical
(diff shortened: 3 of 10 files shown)


examples/alerts/alerts.yaml
```diff
@@ -328,7 +318,10 @@ groups:
       runbook_url: https://github.com/thanos-io/thanos/tree/main/mixin/runbook.md#alert-name-thanossidecarunhealthy
       summary: Thanos Sidecar is unhealthy.
     expr: |
-      time() - max by (job, instance) (thanos_sidecar_last_heartbeat_success_time_seconds{job=~".*thanos-sidecar.*"}) >= 240
+      time() - max by (pod, job, instance) (thanos_sidecar_last_heartbeat_success_time_seconds{job=~".*thanos-sidecar.*"}) >= 240
+      AND on (pod) (
+        min by (pod) (prometheus_tsdb_data_replay_duration_seconds) != 0
```
bwplotka, line 323: This is quite neat actually! We know the duration is done after finish so we expect Prometheus to be healthy.

But instead of doing this long term, can we see if Prometheus expose any health metric and use that instead? 🤔 

Anyway, for now this will do 👍🏽

examples/alerts/alerts.yaml
```diff
@@ -328,7 +318,10 @@ groups:
       runbook_url: https://github.com/thanos-io/thanos/tree/main/mixin/runbook.md#alert-name-thanossidecarunhealthy
       summary: Thanos Sidecar is unhealthy.
```
bwplotka, line 319:
```suggestion
      summary: Thanos Sidecar cannot access Prometheus, even though Prometheus seems healthy and has reloaded WAL.
```

bwplotka (approved):
I am not 100% sure if this is right approach. Sounds like we would like to set some dependency and alert only if Prometheus is healthy but sidecar is not.

I think in this case that might work, but let's make sure to be explicit. I would even vote for alert name to be explit, to something `SidecarNoConnectionToStartedPrometheus`, WDYT?

arajkumar:
Thanks a lot @bwplotka for this quick review.

> I am not 100% sure if this is right approach. Sounds like we would like to set some dependency and alert only if Prometheus is healthy but sidecar is not.

Do you mean to express alert dependencies through[ inhibition rules](https://prometheus.io/docs/alerting/latest/alertmanager/#inhibition)? If so, I thought of that and @simonpasquier suggested that it might over complicate.

> I think in this case that might work, but let's make sure to be explicit. I would even vote for alert name to be explit, to something `SidecarNoConnectionToStartedPrometheus`, WDYT?

+1, How about `ThanosSidecarNotConnectedToPrometheus`?

examples/alerts/alerts.yaml
arajkumar: AFAIK, I couldn't find a reasonable one. May be @dgrisonnet @paulfantom know something?

bwplotka:
Yes it might be hard, especially when one might not alert when Prometheus is reloading (which makes sense!)

I would prefer what I propose as from title you expect this failing when it's not connected, but it will NOT fail if not connected AND if Prometheus is reloading which might surprise op. (:

examples/alerts/alerts.yaml
bwplotka: There is none, but we can add one if we want. Question, do we want (:

examples/alerts/alerts.yaml
```diff
@@ -301,16 +301,6 @@ groups:
       severity: warning
 - name: thanos-sidecar
   rules:
-  - alert: ThanosSidecarPrometheusDown
```
bwplotka, line 304: Is this intentional?

bwplotka: Description or Title of this PR is not mentioning this change 🤔

arajkumar: @bwplotka Yes it was intensional, but excluding it in the title is not :) 

I've removed it as it would fire at the same time as ThanosSidecarPrometheusDown., so found it to be redundant. Let me amend the title and description to add info above removal of ThanosSidecarPrometheusDown.

More details: https://github.com/thanos-io/thanos/issues/3915#issuecomment-885475143

examples/alerts/alerts.yaml
arajkumar: I'm not really sure. As you mentioned, I will go ahead with the current solution and improve it in future.

CHANGELOG.md
```diff
@@ -18,6 +18,7 @@ We use *breaking :warning:* to mark changes that are not backward compatible (re
 
 - [#4468](https://github.com/thanos-io/thanos/pull/4468) Rule: Fix temporary rule filename composition issue.
 - [#4476](https://github.com/thanos-io/thanos/pull/4476) UI: fix incorrect html escape sequence used for '>' symbol.
+- [#4508](https://github.com/thanos-io/thanos/pull/4508) Fix ThanosSidecarUnhealthy alert during prometheus WAL replay and remove ThanosSidecarUnhealthy.
```
bwplotka, line 21: This entry does not make sense (fix `ThanosSidecarUnhealthy` and remove `ThanosSidecarUnhealthy`) - what we try to say here?

examples/alerts/alerts.yaml
bwplotka: `ThanosSidecarPrometheusDown` will be the same as `ThanosSidecarPrometheusDown`? 🤔 Typo?

bwplotka:
What about alert name?

What about `ThanosSidecarNoConnectionToStartedPrometheus`?

arajkumar:
> ThanosSidecarNoConnectionToStartedPrometheus

@bwplotka looks good to me.

CHANGELOG.md
```diff
@@ -18,6 +18,7 @@ We use *breaking :warning:* to mark changes that are not backward compatible (re
 
 - [#4468](https://github.com/thanos-io/thanos/pull/4468) Rule: Fix temporary rule filename composition issue.
 - [#4476](https://github.com/thanos-io/thanos/pull/4476) UI: fix incorrect html escape sequence used for '>' symbol.
+- [#4508](https://github.com/thanos-io/thanos/pull/4508) Refactor sidecar alerts.
```
bwplotka, line 21: This is not useful changelog item, is it? (:

examples/alerts/alerts.yaml
bwplotka: ping 
image

bwplotka (approved):
Thanks! Let's fix changelog, otherwise LGTM 💪🏽

examples/alerts/alerts.yaml
arajkumar: @bwplotka PR description already has info about removal of `ThanosSidecarPrometheusDown`.

CHANGELOG.md
arajkumar: Good to have, how about "Refactor sidecar alerts(ThanosSidecarUnhealthy, ThanosSidecarPrometheusDown)"?

examples/alerts/alerts.yaml
```diff
@@ -321,14 +311,18 @@ groups:
...
       description: Thanos Sidecar {{$labels.instance}} is unhealthy for more than
         {{$value}} seconds.
-      runbook_url: https://github.com/thanos-io/thanos/tree/main/mixin/runbook.md#alert-name-thanossidecarunhealthy
-      summary: Thanos Sidecar is unhealthy.
+      runbook_url: https://github.com/thanos-io/thanos/tree/main/mixin/runbook.md#alert-name-thanossidecarnoconnectiontostartedprometheus
+      summary: Thanos Sidecar cannot access Prometheus, even though Prometheus seems
+        healthy and has reloaded WAL.
     expr: |
-      time() - max by (job, instance) (thanos_sidecar_last_heartbeat_success_time_seconds{job=~".*thanos-sidecar.*"}) >= 240
+      time() - max by (pod, job, instance) (thanos_sidecar_last_heartbeat_success_time_seconds{job=~".*thanos-sidecar.*"}) >= 240
```
bwplotka, line 322: Sorry for challenging this last minute, but isn't expression with `thanos_sidecar_prometheus_up` simpler?

 ```suggestion
      thanos_sidecar_prometheus_up{job=~".*thanos-sidecar.*"}) == 0
```

File: pkg/rules/rules_test.go [hunk 7]
```diff
@@ -82,7 +82,7 @@ func testRulesAgainstExamples(t *testing.T, dir string, server rulespb.RulesServ
 		{
 			Name:                    "thanos-sidecar",
 			File:                    filepath.Join(dir, "alerts.yaml"),
-			Rules:                   []*rulespb.Rule{someAlert, someAlert, someAlert},
```
bwplotka, line 85: Do we  need this change?

bwplotka (commented):
2 more suggestions. Also please run `make docs` again - tests are failing.

Thanks!

pkg/rules/rules_test.go
arajkumar: Since we have removed the `ThanosSidecarPrometheusDown` alert, now sidecar can emit only 2 alerts, hence the change :)

examples/alerts/alerts.yaml
arajkumar: Great question :) AFAIK, thanos_sidecar_last_heartbeat_success_time_seconds and thanos_sidecar_prometheus_up carries the same information encoded in different ways. Let me evaluate and switch to  thanos_sidecar_prometheus_up if it is the case.

arajkumar: @bwplotka If we switch to thanos_sidecar_last_heartbeat_success_time_seconds, We may not able to show the duration at which sidecar is not connected in the alert description. Hope that is fine.
e.g.
> Thanos Sidecar thanos-sidecar-1 is unhealthy for more than 720 seconds.

arajkumar:
@bwplotka I don't know, for some reason `make check-docs` never fails on machine but always fails on CI. Looks like something related to line ending on OSX vs Linux.

CHANGELOG.md
```diff
@@ -28,6 +28,7 @@ We use *breaking :warning:* to mark changes that are not backward compatible (re
 - [#4476](https://github.com/thanos-io/thanos/pull/4476) UI: fix incorrect html escape sequence used for '>' symbol.
 - [#4532](https://github.com/thanos-io/thanos/pull/4532) Mixin: Fixed "all jobs" selector in thanos mixin dashboards.
 - [#4607](https://github.com/thanos-io/thanos/pull/4607) Azure: Fix Azure MSI Rate Limit
+- [#4508](https://github.com/thanos-io/thanos/pull/4508) Adjust and rename `ThanosSidecarUnhealthy` to `ThanosSidecarNoConnectionToStartedPrometheus`; Remove `ThanosSidecarPrometheusDown`.
```
bwplotka, line 31: This has to be moved now to the top of this change log (different release)

cmd/thanos/sidecar.go
```diff
@@ -138,10 +138,6 @@ func runSidecar(
 			Name: "thanos_sidecar_prometheus_up",
 			Help: "Boolean indicator whether the sidecar can reach its Prometheus peer.",
 		})
```
bwplotka, line 140: We need to reflrect this change in CHANGELOG too

bwplotka (approved):
Thanks, LGTM, just we need to move the changelog item now and add one more info.

bwplotka (approved):
👍🏽 Thanks!

PR #4509, opened 2021-08-02
adds duration_ms in int64 to the logs

rhassanein:

## Motivation

Currently the logged lines contain a `duration` field that is encoded as a string with varying units. When ingesting these structured logs into a logging system (eg. elasticsearch), the field is parsed as a string which doesn't allow for range queries.

Example from v0.19.0:
```json
{
      "ts": "2021-08-02T11:38:26.928570654Z",
      "cached": 23731,
      "duration": "2.09562852s",
      "tier": "sv",
      "stage": "main",
      "returned": 1567,
      "type": "monitoring",
      "component": "block.BaseFetcher",
      "msg": "successfully synchronized block metadata",
      "caller": "fetcher.go:476",
      "level": "info",
      "partial": 0
}
```
Adding a dedicated `duration_ms` integer field solves this problem.

Example:
```json
{
      "duration": "2.09562852s",
      "duration_ms": "2095",
}
```
In order to avoid schema conflicts and introducing a breaking change, we're adding a new field rather than changing the existing one.

## Changes

Adds an integer `duration_ms` to the logged lines with the `duration` field.

## Verification

@igorwwwwwwwwwwwwwwwwwwww thanks for co-authoring this PR.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -15,6 +15,7 @@ We use *breaking :warning:* to mark changes that are not backward compatible (re
 - [#4482](https://github.com/thanos-io/thanos/pull/4482) COS: Add http_config for cos object store client.
 - [#4487](https://github.com/thanos-io/thanos/pull/4487) Query: Add memcached auto discovery support.
 - [#4444](https://github.com/thanos-io/thanos/pull/4444) UI: Add search block UI.
+- [#4509](https://github.com/thanos-io/thanos/pull/4509) Logging: Adds duration_ms in int64 to the logs.
 
 ### Fixed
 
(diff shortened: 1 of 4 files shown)


GiedriusS (dismissed):
This is interesting, I would've just changed `duration` to always use a fixed unit but this works too and makes sense :hugs: thx

GiedriusS:
Please rebase your branch on the latest `main`, the tests should be fixed there.

rhassanein:
@GiedriusS thanks for reviewing this! I just added a CHANGELOG entry and attempted a main rebase, hopefully that fixes the tests.

bwplotka (approved):
Thank you!

vanugrah:
Sweet! I was just thinking about this :D

PR #4516, opened 2021-08-03
reduce the length of scenario name

clyang82:
Some linux systems support less than 64 characters for a host-name. the host name is more than 64 in this case `TestToolsBucketWebExternalPrefixAndRoutePrefix` so that the case cannot be passed:
```
06:22:50 toolsBucketWeb-1: level=error name=toolsBucketWeb-1 ts=2021-08-03T06:22:50.42873554Z caller=runutil.go:101 msg="function failed. Retrying in next tick" err="BaseFetcher: iter bucket: Get \"http://e2e_test_tools_bucket_web_external_prefix_and_route_prefix-minio-8080:8080/toolsBucketWeb_test/?location=\": dial tcp: lookup e2e_test_tools_bucket_web_external_prefix_and_route_prefix-minio-8080: no such host"

```

## Changes

reduce the length of scenario name

### DIFF
--- a/test/e2e/tools_bucket_web_test.go
+++ b/test/e2e/tools_bucket_web_test.go
@@ -112,7 +112,7 @@ func TestToolsBucketWebExternalPrefix(t *testing.T) {
 func TestToolsBucketWebExternalPrefixAndRoutePrefix(t *testing.T) {
 	t.Parallel()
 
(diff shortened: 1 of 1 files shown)


GiedriusS (approved):
I suppose `e2e.NewScenario` could return an error in such a case. Care improving that function? (:

clyang82:
Yes. Thanks @GiedriusS 

BTW, how does it work in github action?

PR #4535, opened 2021-08-09
[v2] store: reuse buffers for serializing Series() responses

GiedriusS:
v1 can be found here: https://github.com/thanos-io/thanos/pull/4512

Compared with v1:
* Uses less CPU because Size() is called only once just as before;

* Uses `protoc-go-inject-field` to inject extra fields into SeriesResponse
which are used for marshaling the responses;

* Uses []byte instead of bytes.Buffer because bytes.Buffer allocates 2x
amount of memory and in our cases the responses are more or less of the
same size so less memory is wasted;

I think this version is more maintainable since the details of the
allocation are hidden from the Series() function itself, it all happens
in the Marshal() function.

Benchmarks:
```
name                                                           old time/op    new time/op    delta
BucketSeries/1000000SeriesWith1Samples/1of1000000-16             97.8ms ± 6%    89.4ms ± 8%   -8.63%  (p=0.000 n=49+45)
BucketSeries/1000000SeriesWith1Samples/10of1000000-16            92.7ms ±12%    88.6ms ± 7%   -4.42%  (p=0.000 n=48+47)
BucketSeries/1000000SeriesWith1Samples/1000000of1000000-16        1.14s ± 4%     1.11s ± 3%   -2.89%  (p=0.000 n=50+49)
BucketSeries/100000SeriesWith100Samples/1of10000000-16           6.58ms ± 2%    6.70ms ± 9%     ~     (p=0.125 n=46+50)
BucketSeries/100000SeriesWith100Samples/100of10000000-16         6.73ms ± 5%    6.82ms ± 9%     ~     (p=0.575 n=49+50)
BucketSeries/100000SeriesWith100Samples/10000000of10000000-16     123ms ± 5%     119ms ± 7%   -2.92%  (p=0.000 n=48+50)
BucketSeries/1SeriesWith10000000Samples/1of10000000-16            131µs ± 6%     129µs ± 6%   -0.85%  (p=0.027 n=47+47)
BucketSeries/1SeriesWith10000000Samples/100of10000000-16          129µs ± 2%     129µs ± 8%     ~     (p=0.358 n=44+49)
BucketSeries/1SeriesWith10000000Samples/10000000of10000000-16    38.6ms ± 9%    34.7ms ±10%  -10.11%  (p=0.000 n=49+50)

name                                                           old alloc/op   new alloc/op   delta
BucketSeries/1000000SeriesWith1Samples/1of1000000-16             62.0MB ± 0%    62.0MB ± 0%     ~     (p=0.529 n=47+50)
BucketSeries/1000000SeriesWith1Samples/10of1000000-16            62.1MB ± 0%    62.1MB ± 0%     ~     (p=0.334 n=49+50)
BucketSeries/1000000SeriesWith1Samples/1000000of1000000-16       1.37GB ± 0%    1.27GB ± 0%   -7.03%  (p=0.000 n=49+50)
BucketSeries/100000SeriesWith100Samples/1of10000000-16           4.85MB ± 0%    4.85MB ± 0%     ~     (p=0.365 n=50+50)
BucketSeries/100000SeriesWith100Samples/100of10000000-16         4.86MB ± 0%    4.86MB ± 0%     ~     (p=0.579 n=50+50)
BucketSeries/100000SeriesWith100Samples/10000000of10000000-16     157MB ± 4%     130MB ± 5%  -16.99%  (p=0.000 n=50+50)
BucketSeries/1SeriesWith10000000Samples/1of10000000-16            213kB ± 0%     213kB ± 0%   +0.14%  (p=0.000 n=50+48)
BucketSeries/1SeriesWith10000000Samples/100of10000000-16          213kB ± 0%     213kB ± 0%   +0.14%  (p=0.000 n=50+50)
BucketSeries/1SeriesWith10000000Samples/10000000of10000000-16     115MB ± 0%      62MB ± 8%  -45.98%  (p=0.000 n=49+50)

name                                                           old allocs/op  new allocs/op  delta
BucketSeries/1000000SeriesWith1Samples/1of1000000-16              9.69k ± 0%     9.70k ± 1%     ~     (p=0.143 n=49+50)
BucketSeries/1000000SeriesWith1Samples/10of1000000-16             9.79k ± 0%     9.79k ± 0%     ~     (p=0.845 n=49+49)
BucketSeries/1000000SeriesWith1Samples/1000000of1000000-16        11.0M ± 0%     10.0M ± 0%   -9.06%  (p=0.000 n=49+50)
BucketSeries/100000SeriesWith100Samples/1of10000000-16            1.10k ± 0%     1.10k ± 0%   +0.27%  (p=0.000 n=50+50)
BucketSeries/100000SeriesWith100Samples/100of10000000-16          1.14k ± 0%     1.14k ± 0%     ~     (p=0.622 n=50+50)
BucketSeries/100000SeriesWith100Samples/10000000of10000000-16     1.10M ± 0%     1.00M ± 0%   -9.04%  (p=0.000 n=49+50)
BucketSeries/1SeriesWith10000000Samples/1of10000000-16              200 ± 0%       202 ± 0%   +1.00%  (p=0.000 n=48+50)
BucketSeries/1SeriesWith10000000Samples/100of10000000-16            200 ± 0%       202 ± 0%   +1.00%  (p=0.000 n=49+50)
BucketSeries/1SeriesWith10000000Samples/10000000of10000000-16      167k ± 0%      167k ± 0%   +0.00%  (p=0.000 n=50+50)
```

CPU usage is +/- the same when you take the variance into account, the
allocated memory is the real difference and, as expected, the benchmarks
show the most improvement on the cases with lots of callers.

I have added a `$(MAKE) format` step after `$(MAKE) proto` so that the `*.pb.go` would have
a consistent style and CI would pass.


pkg/store/storepb/custom.go
```diff
@@ -456,3 +461,119 @@ func CompareLabels(a, b []Label) int {
...
 	return labelpb.ZLabelsToPromLabels(lset)
 }
+
+// Type alias because protoc-go-inject-field does not support
+// managing imports.
+type syncPool = sync.Pool
+
+// The following were copied/pasted from gogoprotobuf generated code with changes
+// to make it work with sync.Pool / []byte slice.
+func (m *SeriesResponse) Marshal() (dAtA []byte, err error) {
```
kakkoyun, line 471: I know it's a copy-pasta from generated code but this makes my eyes bleed 
```suggestion
func (m *SeriesResponse) Marshal() (data []byte, err error) {
```

kakkoyun (commented):
Looks amazing! Awesome work!

I wonder how hard it's going to be to migrate to https://github.com/planetscale/vtprotobuf. It already supports pooling for responses, right? This is more or less the same thing we achieve with gogo?

GiedriusS:
> Looks amazing! Awesome work!
> 
> I wonder how hard it's going to be to migrate to https://github.com/planetscale/vtprotobuf. It already supports pooling for responses, right? This is more or less the same thing we achieve with gogo?

Yep, but only "on the other side". It generates more or less the same methods for marshaling that we could reuse here once we move to it:

https://github.com/planetscale/vtprotobuf/blob/27a3a9c2e0ac2c1a52d57d409bcea997ec1bf76b/features/marshal/marshalto.go#L554-L583

So, the "glue" code in this PR will still remain similar once we move to it AFAICT. For unmarshaling it is much nicer: https://github.com/planetscale/vtprotobuf/blob/5a02622d1e2a726cf08df370eccdca6844f11e70/features/pool/pool.go#L86-L95. It's going to be lots of work to move to it IMHO as evidenced by the Vitess PR https://github.com/vitessio/vitess/pull/8075 :smile:

pkg/store/storepb/custom.go
bwplotka: Thanks for this, I like the attempt of hiding complexity but there are a couple of worrying parts in terms of readability and maintainability of this code:
* `**[]byte` is something we can definitely avoid. I guess we do that to have extra info if this was pooled or not but this goes away if we fix next problem:
* We have mixed responsibility on who is suppose to get bytes from pool. 1 -> user of `SeriesResponse` has to get before and put after send. 2 ->  SeriesResponse itself has to use pool to resize the object.

I can see the reason behind it. But wonder if we can simplify this, by:
1. Making sure pool is only used by SeriesResponse. We "get" buf on Marshal and resize if needed
2. Add "Close" method to series response which puts off the buf when we say "it is closed"?

This way the whole thing is encapsulated: Pool is used only here and we never pass any outside bytes and have this weird double pointer

pkg/store/storepb/custom.go
```diff
@@ -456,3 +461,114 @@ func CompareLabels(a, b []Label) int {
...
+	size := m.Size()
+
+	var respBuf []byte
+	if m.respBuf == nil {
+		respBuf = make([]byte, size)
+	} else {
+		if cap(**m.respBuf) < size {
+			if m.respPool != nil {
+				m.respPool.Put(*m.respBuf)
```
bwplotka, line 480: Does it mean that pooled buffers will have only 32 bytes only? I think this might limit usability of this pooling. I wonder if we don't want to resize for bigger ones and put resized ones with some maximum length on it.

pkg/store/storepb/custom.go
```diff
@@ -456,3 +461,114 @@ func CompareLabels(a, b []Label) int {
...
+	i--
+	data[i] = 0x12
+	return len(data) - i, nil
+}
+func (m *SeriesResponse_Hints) MarshalTo(data []byte) (int, error) {
+	size := m.Size()
+	return m.MarshalToSizedBuffer(data[:size])
+}
+
+func (m *SeriesResponse_Hints) MarshalToSizedBuffer(data []byte) (int, error) {
```
bwplotka, line 559: Those are changed or just copied?

pkg/store/storepb/rpc.proto
```diff
@@ -118,7 +118,11 @@ enum Aggr {
   COUNTER = 5;
 }
 
+// @inject_field: respBuf **[]byte
+// @inject_field: respPool *syncPool
```
bwplotka, line 122: Nice trick!

bwplotka (dismissed):
Thanks for this! I have two bigger suggestions, otherwise LGTM!

I need to agree on top of @kakkoyun - https://github.com/planetscale/vtprotobuf can do pooling too - even in exactly similar way as my suggestion 🙈 

Still, the path for gogo proto removal is far, so we could merge this in. But this code will removed when we switch to new vtprotobuf yes

pkg/store/storepb/custom.go
GiedriusS: It means that it is the initial size so they would be useful for responses with minimal labels such as `a="b"` and one sample, for example. If there are more samples and/or more labels then they would be resized to be big just enough. I'm not sure I get the suggestion about the maximum length - we don't make them any bigger than they are right now in this PR.

pkg/store/storepb/custom.go
GiedriusS: These are copied and minimally cleaned up as per @kakkoyun's suggestion.

pkg/store/storepb/custom.go
GiedriusS: I made it the way it is right now is because ideally the same buffer would be shared between marshaling different responses:

* Get a buffer from sync.Pool
* Pass it to Marshal() via SeriesResponse which can make it larger if necessary
* The original caller i.e. Series() function puts it back - it doesn't know if the underlying slice has been changed

This seems completely natural to me because multiple SeriesResponse structs are for different responses, only the buffer & pool is shared for the actual serialization.

If I understood you correctly, you are proposing moving all of the logic into SeriesResponse itself. Then, we would have something like:
```go
var resp storepb.SeriesResponse
..
resp = storepb.NewSeriesResponse(newSeries, pool)
```

And then in a loop, we would have to do:
```go
resp.Result = newSeries
```

As a result, a connection between the SeriesResponse members and a higher-level function, Series(), would appear. We could add a function to "hide" it but it would be very shallow. I'm not sure if this 2nd option is really better :thinking: 

Given that we will remove this code in a few months hopefully it's probably not worth spending so much energy on minute details such as this :smile: 

If you really think this should be the way then I can make this change but because you've approved this I assume that such a minor possible change isn't a blocker to merge this PR

GiedriusS:
@bwplotka I have made changes according to your suggestions:

- 32 bytes constant has been removed and now we do not allocate more than is necessary;
- I moved all of the `sync.Pool` logic into the Marshal() function and the new Close() function;
- Created a new constructor function that accepts a `*sync.Pool` and used it in Series().

Benchmarks are even more impressive now:
```
name                                                           old time/op    new time/op    delta
BucketSeries/1000000SeriesWith1Samples/1of1000000-16             95.0ms ± 8%    96.4ms ± 5%     ~     (p=0.221 n=25+24)
BucketSeries/1000000SeriesWith1Samples/10of1000000-16            92.9ms ±11%    93.5ms ±11%     ~     (p=0.714 n=24+25)
BucketSeries/1000000SeriesWith1Samples/1000000of1000000-16        1.12s ± 2%     1.13s ± 5%     ~     (p=0.091 n=23+25)
BucketSeries/100000SeriesWith100Samples/1of10000000-16           6.49ms ± 2%    6.63ms ± 5%   +2.23%  (p=0.001 n=21+25)
BucketSeries/100000SeriesWith100Samples/100of10000000-16         6.70ms ± 3%    6.51ms ± 2%   -2.78%  (p=0.000 n=24+22)
BucketSeries/100000SeriesWith100Samples/10000000of10000000-16     122ms ± 3%     114ms ± 2%   -6.69%  (p=0.000 n=25+25)
BucketSeries/1SeriesWith10000000Samples/1of10000000-16            128µs ± 2%     120µs ± 1%   -5.94%  (p=0.000 n=24+24)
BucketSeries/1SeriesWith10000000Samples/100of10000000-16          126µs ± 5%     120µs ± 1%   -4.87%  (p=0.000 n=24+23)
BucketSeries/1SeriesWith10000000Samples/10000000of10000000-16    39.3ms ± 5%    33.1ms ± 5%  -15.87%  (p=0.000 n=24+25)

name                                                           old alloc/op   new alloc/op   delta
BucketSeries/1000000SeriesWith1Samples/1of1000000-16             62.0MB ± 0%    62.0MB ± 0%     ~     (p=0.314 n=21+25)
BucketSeries/1000000SeriesWith1Samples/10of1000000-16            62.1MB ± 0%    62.1MB ± 0%     ~     (p=0.052 n=25+25)
BucketSeries/1000000SeriesWith1Samples/1000000of1000000-16       1.37GB ± 0%    1.26GB ± 0%   -7.61%  (p=0.000 n=25+25)
BucketSeries/100000SeriesWith100Samples/1of10000000-16           4.85MB ± 0%    4.85MB ± 0%     ~     (p=0.788 n=25+25)
BucketSeries/100000SeriesWith100Samples/100of10000000-16         4.86MB ± 0%    4.86MB ± 0%   -0.03%  (p=0.008 n=25+25)
BucketSeries/100000SeriesWith100Samples/10000000of10000000-16     158MB ± 2%     130MB ± 4%  -17.83%  (p=0.000 n=25+25)
BucketSeries/1SeriesWith10000000Samples/1of10000000-16            213kB ± 0%     212kB ± 0%   -0.47%  (p=0.000 n=25+24)
BucketSeries/1SeriesWith10000000Samples/100of10000000-16          213kB ± 0%     212kB ± 0%   -0.53%  (p=0.000 n=25+25)
BucketSeries/1SeriesWith10000000Samples/10000000of10000000-16     115MB ± 0%      57MB ± 6%  -50.53%  (p=0.000 n=24+25)

name                                                           old allocs/op  new allocs/op  delta
BucketSeries/1000000SeriesWith1Samples/1of1000000-16              9.68k ± 0%     9.69k ± 0%     ~     (p=0.242 n=24+25)
BucketSeries/1000000SeriesWith1Samples/10of1000000-16             9.79k ± 0%     9.80k ± 0%     ~     (p=0.762 n=25+25)
BucketSeries/1000000SeriesWith1Samples/1000000of1000000-16        11.0M ± 0%     10.0M ± 0%   -9.06%  (p=0.000 n=25+25)
BucketSeries/100000SeriesWith100Samples/1of10000000-16            1.10k ± 0%     1.10k ± 0%     ~     (p=0.435 n=25+25)
BucketSeries/100000SeriesWith100Samples/100of10000000-16          1.14k ± 0%     1.13k ± 0%   -0.13%  (p=0.018 n=25+25)
BucketSeries/100000SeriesWith100Samples/10000000of10000000-16     1.10M ± 0%     1.00M ± 0%   -9.04%  (p=0.000 n=22+25)
BucketSeries/1SeriesWith10000000Samples/1of10000000-16              200 ± 0%       200 ± 0%     ~     (all equal)
BucketSeries/1SeriesWith10000000Samples/100of10000000-16            200 ± 0%       200 ± 0%     ~     (all equal)
BucketSeries/1SeriesWith10000000Samples/10000000of10000000-16      167k ± 0%      167k ± 0%   -0.00%  (p=0.025 n=25+23)
```

Please take a look :)

pkg/store/bucket.go
```diff
@@ -1092,8 +1096,17 @@ func (s *BucketStore) Series(req *storepb.SeriesRequest, srv storepb.Store_Serie
 		s.metrics.seriesGetAllDuration.Observe(stats.getAllDuration.Seconds())
 		s.metrics.seriesBlocksQueried.Observe(float64(stats.blocksQueried))
 	}
+
+	var resp *storepb.SeriesResponse
+	defer func(r **storepb.SeriesResponse) {
```
bwplotka, line 1101: 👁️ 🩸

pkg/store/storepb/custom.go
```diff
@@ -456,3 +475,138 @@ func CompareLabels(a, b []Label) int {
...
+			if poolBuf == nil {
+				respBuf = make([]byte, size)
+				m.respBuf = &respBuf
+			} else {
+				m.respBuf = poolBuf.(*[]byte)
+				respBuf = *m.respBuf
+			}
+		} else {
+			if cap(*m.respBuf) < size {
+				if m.respPool != nil {
```
bwplotka, line 515: Not sure if we have to put too small one back -if it was too small, most likely it will be too small for other responses too.

Not a blocker, it should work both ways, just guessing here a bit.

bwplotka (approved):
Let's go!!! 💪🏽 

Thanks for this!

PR #4357, opened 2021-06-18
docs: Add `mdox` remote/local link checking and move web preprocess

saswatamcode:
This PR adds `mdox` local and remote link checking with smart validation.

 cc: @bwplotka 

Note: This fixes all links in docs. So while all links will work from GitHub now, website links may break. This will be fixed in subsequent PRs where current preprocess will be replaced by `mdox tranform`.

PR #4536, opened 2021-08-10
docs-0.22: Add mdox link checking and formatting

saswatamcode:
Adds mdox to `release-0.22` branch. Related to #4357.


bwplotka (approved):
Amazing, LGTM! Let's make sure docs fail is passing before merging (:

bwplotka:
This is surprising: 
image
 
I think we should never check against tip, since we care about local link here no? This link should be localised, but it was not 🤔

PR #4540, opened 2021-08-10
docs-0.18: Add mdox link checking and formatting

saswatamcode:
Adds mdox to `release-0.18` branch. Related to #4357.


bwplotka (approved):
Let's go!

Test failures are worrying but those are because somehow prometheus bingo .mod file does not have the right replace statements (they might work before but due to how the go module works, Go mod picked newer dep that now is incompatible for this Prom version). I would say, this breaks Prom installation on this branch even without @saswatamcode change, so not a blocker.

PR #4544, opened 2021-08-10
docs-0.14: Add mdox link checking and formatting

saswatamcode:
Adds mdox to `release-0.14` branch. Related to #4357.


bwplotka (approved):
Let's go!

Test are worrying but those are because somehow prometheus bingo .mod file does not have right replace statements (they might work before but due to how go module works, Go mod picked newer dep that now is incompatible for this Prom version). I would say, it will break this branch even without @saswatamcode change, so not a blocker.

docs/tracing.md
```diff
@@ -41,8 +40,8 @@ In Kubernetes it is as easy as (on Thanos sidecar example):
 
 1. Create new directory under `pkg/tracing/<provider>`
 2. Implement `opentracing.Tracer` interface
-3. Add client implementation to the factory in [factory](/pkg/tracing/client/factory.go) code. (Using as small amount of flags as possible in every command)
-4. Add client struct config to [cfggen](/scripts/cfggen/main.go) to allow config auto generation.
+3. Add client implementation to the factory in [factory](../pkg/tracing/client/factory.go) code. (Using as small amount of flags as possible in every command)
```
bwplotka, line 43: Let's add feature to mdox transform that will glue github prefix (with correct version to out of docs links)

PR #4552, opened 2021-08-11
compactor: adds downsample duration histogram

vanugrah:
Signed-off-by: Anugrah Vijay <anugrah.vijay@gmail.com>

## Changes
- Registers and implements downsampleDuration histogram. 
- New var for downsample duration to avoid recomputing time.Since across logs/metric. 

## Verification
- Local validation of compactor with local minIO bucket

### DIFF
--- a/cmd/thanos/downsample.go
+++ b/cmd/thanos/downsample.go
@@ -55,7 +55,7 @@ func newDownsampleMetrics(reg *prometheus.Registry) *DownsampleMetrics {
 	m.downsampleDuration = promauto.With(reg).NewHistogramVec(prometheus.HistogramOpts{
 		Name:    "thanos_compact_downsample_duration_seconds",
 		Help:    "Duration of downsample runs",
-		Buckets: []float64{64, 128, 256, 512, 1024, 2048, 4096, 8192}, // or 1m, 5m, 15m, 30m, 60m, 120m, 240m
+		Buckets: []float64{60, 300, 900, 1800, 3600, 7200, 14400}, // 1m, 5m, 15m, 30m, 60m, 120m, 240m
 	}, []string{"group"})
 
 	return m
--- a/cmd/thanos/downsample.go
+++ b/cmd/thanos/downsample.go
@@ -38,6 +38,7 @@ import (
 type DownsampleMetrics struct {
 	downsamples        *prometheus.CounterVec
 	downsampleFailures *prometheus.CounterVec
+	downsampleDuration prometheus.Histogram
 }
 
 func newDownsampleMetrics(reg *prometheus.Registry) *DownsampleMetrics {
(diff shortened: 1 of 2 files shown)


vanugrah:
Based on @yeya24 's feedback on the [previous PR](https://github.com/thanos-io/thanos/pull/4551) I will be: 
- Adding the downsample resolution as label to the histogram. 
- Increasing the starting bucket size

vanugrah:
With respect to the histogram buckets, we could get rid of the initial few buckets and make it `64, 128, 256, 512, 1024, 2048, 4096, 8192` or we could define more user friendly intervals such as 1m, 5m, 15m, 30m, 1h, 2h, 4h. WDYT?

cmd/thanos/downsample.go
```diff
@@ -51,6 +53,11 @@ func newDownsampleMetrics(reg *prometheus.Registry) *DownsampleMetrics {
 		Name: "thanos_compact_downsample_failures_total",
 		Help: "Total number of failed downsampling attempts.",
 	}, []string{"group"})
+	m.downsampleDuration = promauto.With(reg).NewHistogramVec(prometheus.HistogramOpts{
+		Name:    "thanos_compact_downsample_duration_seconds",
+		Help:    "Duration of downsample runs",
+		Buckets: []float64{1, 2, 4, 8, 16, 32, 64, 128, 256, 512, 1024, 2048, 4096, 8192},
+	}, []string{"resolution"})
```
yeya24, line 60: We can just reuse the label `group` to be aligned with other downsample metrics. This reflect the block resolution as well

yeya24 (changes requested):

cmd/thanos/downsample.go
vanugrah: Makes sense. Could also help you identify the biggest offending group as well.

yeya24:
Anything you still want to update? @vanugrah If not please mark this pr ready for review. 😄

vanugrah:
Bucket ranges?  Currently it mimics the compactors geometric  series: `64, 128, 256, 512, 1024, 2048, 4096, 8192`. 
But changing that to recognizable intervals may be more user friendly: `1m, 5m, 15m, 30m, 60m, 120m, 240m`.

yeya24:
> Bucket ranges? Currently it mimics the compactors geometric series: `64, 128, 256, 512, 1024, 2048, 4096, 8192`.
> But changing that to recognizable intervals may be more user friendly: `1m, 5m, 15m, 30m, 60m, 120m, 240m`.

Latter sounds better.

yeya24 (approved):
LGTM. Thanks for the contribution.

vanugrah:
😄 Thanks for the speedy feedback!

vanugrah:
So for flakey e2e tests do we just rerun?

yeya24:
> So for flakey e2e tests do we just rerun?

I restarted it and it is passing now

vanugrah:
Wooot 🎉  Thanks for your help @yeya24 . More PR's soon to come.

GiedriusS (commented):
With `--wait` doesn't this mean that we will have infinitely different groups? :thinking: Perhaps a histogram over all durations would be more suitable i.e. drop the `group` label?

vanugrah:
@GiedriusS Any thoughts?

vanugrah:
@yeya24  Your review was marked stale because I had to resolve a merge conflict on the Changelog. Care to re-review?

yeya24 (approved):
LGTM. Let's merge it.

PR #4566, opened 2021-08-13
Fix -compact.skip-block-with-out-of-order-chunks

pracucci:

## Changes

The PR https://github.com/thanos-io/thanos/pull/4469 was just merged but I've the feeling it's not correct. The problem is that when the skip is **disabled** it should still be a critical error while right now the error is ignored at all (see comment annotation in the PR).

Also, from a design perspective, I don't think it's the responsibility of the grouper to know if skipping is enabled or not, so I've moved that config to the `BucketCompactor`.

## Verification

Existing tests.

### DIFF
--- a/pkg/compact/compact.go
+++ b/pkg/compact/compact.go
@@ -229,20 +229,19 @@ func defaultGroupKey(res int64, lbls labels.Labels) string {
 // DefaultGrouper is the Thanos built-in grouper. It groups blocks based on downsample
 // resolution and block's labels.
 type DefaultGrouper struct {
-	bkt                            objstore.Bucket
-	logger                         log.Logger
-	acceptMalformedIndex           bool
-	enableVerticalCompaction       bool
-	compactions                    *prometheus.CounterVec
-	compactionRunsStarted          *prometheus.CounterVec
-	compactionRunsCompleted        *prometheus.CounterVec
-	compactionFailures             *prometheus.CounterVec
-	verticalCompactions            *prometheus.CounterVec
-	garbageCollectedBlocks         prometheus.Counter
-	blocksMarkedForDeletion        prometheus.Counter
-	blocksMarkedForNoCompact       prometheus.Counter
-	hashFunc                       metadata.HashFunc
-	skipChunksWithOutOfOrderBlocks bool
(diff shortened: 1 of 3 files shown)


pkg/compact/compact.go
```diff
@@ -783,7 +776,7 @@ func (cg *Group) compact(ctx context.Context, dir string, planner Planner, comp
 			return false, ulid.ULID{}, halt(errors.Wrapf(err, "block with not healthy index found %s; Compaction level %v; Labels: %v", bdir, meta.Compaction.Level, meta.Thanos.Labels))
 		}
 
-		if err := stats.OutOfOrderChunksErr(); cg.skipChunksWithOutofOrderBlocks && err != nil {
+		if err := stats.OutOfOrderChunksErr(); err != nil {
```
pracucci, line 779: With the code in `main`, if `cg.skipChunksWithOutofOrderBlocks` is `false` then we don't return error but we do `toCompactDirs = append(toCompactDirs, bdir)` below, which is wrong.

pkg/compact/compact.go
```diff
@@ -977,8 +973,10 @@ func (c *BucketCompactor) Compact(ctx context.Context) (rerr error) {
 							continue
 						}
 					}
-					// if block has out of order chunk, mark the block for no compaction and continue.
-					if IsOutOfOrderChunkError(err) {
+					// If block has out of order chunk and it has been configured to skip it,
+					// then we can mark the block for no compaction so that the next compaction run
+					// will skip it.
+					if IsOutOfOrderChunkError(err) && c.skipBlocksWithOutOfOrderChunks {
```
pracucci, line 979: I believe this is the right place where we should check for `skipBlocksWithOutOfOrderChunks`. If enabled (which means we skip it), then we do mark the block for no-compaction and we continue processing the next job from the queue. If disabled (which means we don't skip the block) then we should error out (done below).

GiedriusS:
@huyan0 @yeya24 could you please take a look?

huyan0:
LGTM, thanks for the fix.

GiedriusS (approved):
:+1: thx

PR #4567, opened 2021-08-13
mixin: fix unable to generate dashboard when component is null

thanatchakromsang:

## Changes

- Make dashboard overview conditional to behave correctly when disable some components
- Fix `bucket_replication` wrong dimension reference cause error when `rule: null` 

**before**
image

## Verification

by using following command to test without any change to `mixin/config.libsonnet` and update `mixin/config.libsonnet` components to null

```bash
make examples
```

Following `mixin/config.libsonnet` should generate dashboards, rules, and alerts but not error 

```jsonnet
{
  local thanos = self,
  targetGroups+:: {},
  query: null,
}


yeya24 (approved):
LGTM.
I will let @kakkoyun or @metalmatze merge this pr because they are jsonnet experts.

PR #4568, opened 2021-08-14
docs(ruler): improve ruler documentations

weyert:
Adds a call out that you need to add `--store` argument to make the time series created by the recording rules accessible in Thanos Query / UI

## Changes

Small improvement to the documentation for Thanos Rule to callout that you  need add it as a store Thanos Query been struggling with days.

## Verification

Documentation change


docs/components/rule.md
```diff
@@ -96,6 +96,9 @@ labels:
   [ <labelname>: <labelvalue> ]
 ```
 
+Note: If you make use of recording rules, make sure that you expose your Ruler instance as a store in the Thanos Querier so that the new time series can be queried
+as part of Thanos Query. You can do this by adding a new `--store <thanos-ruler-ip>` command-line argument to the Thanos Query command.
```
GiedriusS, line 100: Maybe worth mentioning that it is not the only way of doing this:
```suggestion
as part of Thanos Query. One of the ways you can do this is by adding a new `--store <thanos-ruler-ip>` command-line argument to the Thanos Query command.
```

CHANGELOG.md
```diff
@@ -18,6 +18,8 @@ We use *breaking :warning:* to mark changes that are not backward compatible (re
 - [#4509](https://github.com/thanos-io/thanos/pull/4509) Logging: Adds duration_ms in int64 to the logs.
 - [#4462](https://github.com/thanos-io/thanos/pull/4462) UI: Add find overlap block UI
 - [#4469](https://github.com/thanos-io/thanos/pull/4469) Compact: Add flag `compact.skip-block-with-out-of-order-chunks` to skip blocks with out-of-order chunks during compaction instead of halting
+- [#4568](https://github.com/thanos-io/thanos/pull/4568) Thanos Rule Add note about Rule needs to be part of store list of Query
```
GiedriusS, line 21: We don't typically add documentation changes here :+1:

GiedriusS (commented):
Thx for your PR, just some suggestions :beer:

CHANGELOG.md
weyert: @GiedriusS Does that mean you want me to remove it?

docs/components/rule.md
weyert: Thanks, that makes sense. As I have been banging my head for days without this 'issue'. I am wondering what other ways would be? As you referring to using a service discovery url or something totally different?

CHANGELOG.md
GiedriusS: Yep

weyert: Removed it :)

GiedriusS:
Please add the DCO :smile:

weyert:
> Please add the DCO 😄

Haha, done. And I thought I could quickly make the change via Github :)

ianbillett:
@weyert docs linter is still not happy I'm afraid - `make check-docs` should repro the failure locally for you.

bwplotka (approved):
👍🏽 Thanks, LGTM!

bwplotka:
Note that soon we might be able to skip this part as we want ruler to write to receiver (:

weyert:
> Note that soon we might be able to skip this part as we want ruler to write to receiver (:

Interesting! Do you have any tickets I could follow on this?

PR #4575, opened 2021-08-17
pkg/query: Embed new StoreAddrSpec in other interfaces for consistency

metalmatze:

## Changes

It might be nicer to have an embedded interface where it's the same for all of them.
I started deleting the interfaces like TargetSpec, MetadataSpec etc but I do think they improve readability in other places, so I decided to still keep them around.

## Verification

Tests


yeya24 (approved):
LGTM!

PR #4586, opened 2021-08-20
*: update Cortex (and Prometheus) dependency

GiedriusS:
Update Cortex dependency (and Prometheus together) to include
https://github.com/cortexproject/cortex/commit/70dddb6b70b87f899ab115e79426a3ba522ae6ad.

storage.Querier finally has support for label.Matchers so now we can
fully pass them down. This is covered by tests. Plus, I've played around a bit
with these changes locally.


yeya24:
Nice! Is this pr ready for review now?

GiedriusS:
> Nice! Is this pr ready for review now?

No, still need to fix the tests hence it is a draft

pkg/store/tsdb.go
```diff
@@ -204,10 +205,23 @@ func (s *TSDBStore) LabelNames(ctx context.Context, r *storepb.LabelNamesRequest
...
+		return nil, status.Error(codes.Internal, err.Error())
+	}
+
+	if len(res) > 0 {
+		for _, lbl := range s.extLset {
+			res = append(res, lbl.Name)
+		}
+		sort.Strings(res)
+	}
+
```
GiedriusS, line 224: This is probably the biggest change - in this function :smile:

GiedriusS:
@yeya24 fixed the tests, PTAL :beers:

pkg/api/query/v1.go
```diff
@@ -657,13 +657,13 @@ func (qapi *QueryAPI) labelNames(r *http.Request) (interface{}, []error, *api.Ap
 		return nil, nil, apiErr
 	}
 
-	var matcherSets [][]*labels.Matcher
+	var matcherSets []*labels.Matcher
```
yeya24, line 660: The behavior changes here. We need to support multiple matchers, same as the label values API.

GiedriusS:
@yeya24 thanks for the very quick review! :heart:

bwplotka (approved):
Sweet, thanks! 💪🏽

pkg/promclient/promclient.go
```diff
@@ -702,7 +702,7 @@ func (c *Client) SeriesInGRPC(ctx context.Context, base *url.URL, matchers []*la
 	return m.Data, c.get2xxResultWithGRPCErrors(ctx, "/prom_series HTTP[client]", &u, &m)
 }
 
-// LabelNames returns all known label names. It uses gRPC errors.
+// LabelNames returns all known label names constrained by the given matchers. It uses gRPC errors.
 // NOTE: This method is tested in pkg/store/prometheus_test.go against Prometheus.
 func (c *Client) LabelNamesInGRPC(ctx context.Context, base *url.URL, matchers []storepb.LabelMatcher, startTime, endTime int64) ([]string, error) {
```
yeya24, line 707: This is still not sufficient I think. If the Prometheus server doesn't support label name matchers, then all label names will be returned. What about doing the same as https://github.com/thanos-io/thanos/blob/main/pkg/store/prometheus.go#L533-L567? Checking Prometheus version seems safer here.
@GiedriusS @bwplotka

bwplotka: Yea, some compatibility code has to be there, thanks for being careful. I am on PTO until Monday, will catch on this unless someone else does 🤗

GiedriusS: PTAL https://github.com/thanos-io/thanos/pull/4628.

PR #4587, opened 2021-08-21
docs: Remove embedmd in `examples` and `mixin`

saswatamcode:

## Changes

This PR removes `embedmd` and shifts to `mdox` for markdown content present in `examples` and `mixin`. Also removes `genflagdocs.sh` as its functionality has been replaced in #4357.

Open question: Should Examples and Mixin docs be added to the website sidebar, as currently they can only be referenced through GitHub? 🙂

## Verification

Tested locally.


Namanl2001:
> Should Examples and Mixin docs be added to the website sidebar?

IMHO yes, because there are some pages on the website that reference examples and Mixin docs so instead of hardcoding the GitHub links and redirecting them to Github we can have separate sections for them on the website itself. And I guess it would be a better experience to have everything in one place.

Let's wait for some maintainers' review on this. thanks

bwplotka (approved):
Sweet, thanks!

bwplotka:
Do we want to add them in this PR or next?

saswatamcode:
Will handle in next PR! 🙂

Issue #4591, opened 2021-08-23
Store The request returns inconsistent data and reports an error


mengxifl:
store error 

**Thanos, Prometheus and Golang version used**:  prometheus 2.17 

**Object Storage Provider**: ceph S3

**What happened**: query give data error

this is same grafana panel !

![image](https://user-images.githubusercontent.com/48743664/130465438-a5f94647-9e00-4da8-b761-c1407b97fd04.png)

I set the address of the query on grafana 

![image](https://user-images.githubusercontent.com/48743664/130465087-43e9c75a-d658-41cd-9d1f-30834e889920.png)

when i click this button like this , it not same

![image](https://user-images.githubusercontent.com/48743664/130465612-fd170467-c7fb-4886-9bf0-6125d5941d0f.png)

I think store have some error

**What you expected to happen**: 

I want to get all 

**How to reproduce it (as minimally and precisely as possible)**:

I set 3 receives then I set  

```
apiVersion: apps/v1
kind: StatefulSet
metadata:
  name: thanos-receive
spec:
  serviceName: thanos-receive
  replicas: 3
  selector:
    matchLabels:
      app: thanos-receive
  template:
    metadata:
      labels:
        app: thanos-receive
    spec:
      containers:
      - name: thanos-receive
        image:  thanosio/thanos:main-2021-08-18-5cfa3ff6
        imagePullPolicy: IfNotPresent
        command:
        - /bin/sh
        - "-c"
        - |
          PRE_HOSTNAME_NAME=`hostname | grep -E -o "[a-z-]+[a-z]"`
          ALL_ENDPOINTS_PEER_ADDRESS=""

          i=0
          while [ "$i" -le $(($NODE_NUMBER-1)) ]; do
            ALL_ENDPOINTS_PEER_ADDRESS=$ALL_ENDPOINTS_PEER_ADDRESS\"${PRE_HOSTNAME_NAME}-$i.${STS_SERVICENAME}:10901\",
            i=$(( i + 1 ))
          done

          ALL_ENDPOINTS_PEER_ADDRESS=${ALL_ENDPOINTS_PEER_ADDRESS:0:-1}

          echo ${ALL_ENDPOINTS_PEER_ADDRESS}


          mkdir /runningConfig/
          cp /baseConfig/storageConfig.yaml /runningConfig/storageConfig.yaml
          cp /baseConfig/hashrings-file /runningConfig/hashrings-file


[...]

PR #4593, opened 2021-08-23
store: fix marshaling with sync.Pool

GiedriusS:
Do not forget to check the length of a slice returned by sync.Pool.
Annotate the whole function with comments to aid understanding of it.

Cover the case with a test that fails on `main`.

Fixes https://github.com/thanos-io/thanos/issues/4591.

cc @yeya24 as you've noticed this too

### DIFF
--- a/pkg/store/storepb/custom.go
+++ b/pkg/store/storepb/custom.go
@@ -497,32 +497,45 @@ func (m *SeriesResponse) Marshal() (dAtA []byte, err error) {
 
 	var respBuf []byte
 
-	// No pool defined, allocate directly.
+	// Slow path with no sync.Pool.
 	if m.respPool == nil {
 		respBuf = make([]byte, size)
-	} else {
-		if m.respBuf == nil {
-			poolBuf := m.respPool.Get()
-			if poolBuf == nil {
-				respBuf = make([]byte, size)
-				m.respBuf = &respBuf
-			} else {
-				m.respBuf = poolBuf.(*[]byte)
-				respBuf = *m.respBuf
-			}
-		} else {
-			if cap(*m.respBuf) < size {
-				if m.respPool != nil {
-					m.respPool.Put(m.respBuf)
-				}
-				respBuf = make([]byte, size)
-				m.respBuf = &respBuf
-			} else {
-				respBuf = *m.respBuf
-			}
+		m.respBuf = nil
+
+		n, err := m.MarshalToSizedBuffer(respBuf)
+		if err != nil {
+			return nil, err
 		}
+		return respBuf[len(respBuf)-n:], nil
 	}
 
+	// Fast path with sync.Pool.
+	// m.respBuf must not be nil so that it would be returned to the pool.
+
+	// If no pre-allocated buffer has been passed then try to get a new one.
+	if m.respBuf == nil {
+		poolBuf := m.respPool.Get()
+		// No previous buffer found in the pool, try to allocate.
+		if poolBuf == nil {
+			respBuf = make([]byte, size)
+		} else {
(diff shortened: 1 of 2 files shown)


yeya24 (approved):
LGTM. Thanks for the quick fix.

pkg/store/storepb/custom.go
```diff
@@ -497,32 +497,45 @@ func (m *SeriesResponse) Marshal() (dAtA []byte, err error) {
...
-		if m.respBuf == nil {
-			poolBuf := m.respPool.Get()
-			if poolBuf == nil {
-				respBuf = make([]byte, size)
-				m.respBuf = &respBuf
-			} else {
-				m.respBuf = poolBuf.(*[]byte)
-				respBuf = *m.respBuf
-			}
+		m.respBuf = nil
```
yeya24, line 503: Is this line necessary?

bwplotka: Yea, I don't think so

pkg/store/storepb/custom.go
```diff
@@ -497,32 +497,45 @@ func (m *SeriesResponse) Marshal() (dAtA []byte, err error) {
...
+
+		n, err := m.MarshalToSizedBuffer(respBuf)
+		if err != nil {
+			return nil, err
+		}
+		return respBuf[len(respBuf)-n:], nil
+	}
+
+	// Fast path with sync.Pool.
+	// m.respBuf must not be nil so that it would be returned to the pool.
```
bwplotka, line 513: Does it need to non nil if we do the work below? Can we always do that and remove any outside call pool Get?

pkg/store/storepb/custom.go
```diff
@@ -497,32 +497,45 @@ func (m *SeriesResponse) Marshal() (dAtA []byte, err error) {
...
+	// Last sanity check of the size before the marshaling.
+	if cap(respBuf) < size {
+		if m.respPool != nil {
+			m.respPool.Put(&respBuf)
+		}
+		respBuf = make([]byte, size)
 	}
+	m.respBuf = &respBuf
 
+	// Possibly trim it so that there wouldn't be left-over "garbage" in the slice.
```
bwplotka, line 538: Let's add TODO, but I think we should not trim on 100% calls as between small and large response we would have similar allocs and latency and if we would not pool

pkg/store/storepb/custom_test.go
```diff
@@ -525,3 +526,47 @@ func TestMatchersToString_Translate(t *testing.T) {
 
 	}
 }
+
+// Tests whether the Marshal() function properly checks the size
+// of a slice returned from a sync.Pool.
+// Reference: https://github.com/thanos-io/thanos/issues/4591.
```
bwplotka, line 532:
```suggestion
// Regression test against https://github.com/thanos-io/thanos/issues/4591.
```

bwplotka (approved):
Amazing, LGTM, just one nit.

pkg/store/storepb/custom.go
GiedriusS: I'm not sure I understood this comment. The callers using the old function pass `nil` thus we need to handle this case - it is not guaranteed that the pool exists

bwplotka: I know, i thought we can simplify, and allow this method to `Get`, never pass it from outside - not a biggie

bwplotka (commented):
Looks like we still have this data race we discussed here here https://cloud-native.slack.com/archives/CL25937SP/p1629797444067700

bwplotka (approved):
It does not fix everything, but fix some of things. LGTM and let's iterate.

PR #4594, opened 2021-08-23
reloader: Expose metrics to give info about last operation result/time

PhilipGough:

## Changes

In order to facilitate alerting on errors in the config reloader, this PR exposes four gauges which provide information on the last operation for both reloads and config applies.

These could be used similarly to `prometheus_config_last_reload_success_timestamp_seconds` and  `prometheus_config_last_reload_successful ` exposed by Prometheus.


CHANGELOG.md
```diff
@@ -18,6 +18,7 @@ We use *breaking :warning:* to mark changes that are not backward compatible (re
 - [#4509](https://github.com/thanos-io/thanos/pull/4509) Logging: Adds duration_ms in int64 to the logs.
 - [#4462](https://github.com/thanos-io/thanos/pull/4462) UI: Add find overlap block UI
 - [#4469](https://github.com/thanos-io/thanos/pull/4469) Compact: Add flag `compact.skip-block-with-out-of-order-chunks` to skip blocks with out-of-order chunks during compaction instead of halting
+- [#4594](https://github.com/thanos-io/thanos/pull/4469) reloader: Expose metrics in config reloader to give info on the last operation.
```
yeya24, line 21: The link is wrong.

pkg/reloader/reloader.go
```diff
@@ -243,7 +271,11 @@ func (r *Reloader) Watch(ctx context.Context) error {
 		if err := r.apply(applyCtx); err != nil {
 			r.configApplyErrors.Inc()
 			level.Error(r.logger).Log("msg", "apply error", "err", err)
+			r.lastConfigApplySuccess.Set(0)
+			continue
 		}
+		r.lastConfigApplySuccess.Set(1)
+		r.lastConfigApplySuccessTimestamp.SetToCurrentTime()
```
yeya24, line 278: Do we need to have two metrics for `config apply` and `reload` here? As I can see, if the reload fails, then the config apply fails as well. WDYT?

PhilipGough: I saw we were maintaining counters for both so just followed the format. But happy to remove the `apply` metrics if they add little value?

pkg/reloader/reloader.go
```diff
@@ -154,6 +158,18 @@ func New(logger log.Logger, reg prometheus.Registerer, o *Options) *Reloader {
 				Help: "Total number of reload requests that failed.",
 			},
 		),
+		lastReloadSuccess: promauto.With(reg).NewGauge(
+			prometheus.GaugeOpts{
+				Name: "reloader_last_reload_successful",
```
bwplotka, line 163: Why querying counter of all reloads does not give us that? 🤔

pkg/reloader/reloader.go
```diff
@@ -154,6 +158,18 @@ func New(logger log.Logger, reg prometheus.Registerer, o *Options) *Reloader {
...
+				Help: "Whether the last reload attempt was successful",
+			},
+		),
+		lastReloadSuccessTimestamp: promauto.With(reg).NewGauge(
+			prometheus.GaugeOpts{
+				Name: "reloader_last_reload_success_timestamp_seconds",
```
bwplotka, line 169: This, plus counter of reloads should give us enough info, so no need for `reloader_last_reload_successful` maybe? (Let's avoid too many ways of doing the same thing)

pkg/reloader/reloader.go
```diff
@@ -166,6 +182,18 @@ func New(logger log.Logger, reg prometheus.Registerer, o *Options) *Reloader {
 				Help: "Total number of config apply operations that failed.",
 			},
 		),
+		lastConfigApplySuccess: promauto.With(reg).NewGauge(
+			prometheus.GaugeOpts{
+				Name: "reloader_last_config_apply_successful",
```
bwplotka, line 187: ditto

pkg/reloader/reloader.go
bwplotka: I think we can have both if we want - or add label for all metrics like `op=reload` or `op=apply`. But I would argue that if you know reload fail, you can dive later on why

bwplotka: No need to have root cause metric here

bwplotka (commented):
Nice, some suggestions first (:

pkg/reloader/reloader.go
PhilipGough: @bwplotka - I don't understand this sorry. We maintain a counter for total reloads and failed reloads. How would I determine from those if the latest retry was a success?

pkg/reloader/reloader.go
PhilipGough: Removed these, on reflection I don't think they are important from an alerting persepctive.

pkg/reloader/reloader.go
PhilipGough: Just looking at the mixin for bad config alert for Prometheus it achieves it using similar https://github.com/prometheus/prometheus/blob/main/documentation/prometheus-mixin/alerts.libsonnet#L8

Still unsure how we would guard against false alarm by determining without determining if they very last attempt was a success.

For example say I have a 90% failure rate on reloads, I might alert on that, although I wouldn't care if I knew the last reload attempt actually forced the reload. How would we achieve the same guard with the timestamp?

bwplotka: Right - we can do that with ` increase(total[5m])  - increase(fails[5m])  > 0` no? 

But happy to stick to standard if we are familiar with it

bwplotka (approved):
Not a fan of duplicated info here, but happy to stick to standard if we want.

PR #4611, opened 2021-08-30
.*: Fix Azure MSI token refresh

phillebaba:

## Changes

The changes made in #4607  was missing a token refresh as I expected that to be done by the client. This was not the case and after leaving things running over the weekend the token eventually expired and everything stopped working. This change adds a token refresh function which renews the token before it expires.

## Verification

Have tested a build of this branch in an AKS cluster, and it works. I will leave it running for another day just to make sure that the token refresh actually works.

### DIFF
--- a/pkg/objstore/azure/helpers.go
+++ b/pkg/objstore/azure/helpers.go
@@ -16,6 +16,8 @@ import (
 	"github.com/Azure/azure-pipeline-go/pipeline"
 	blob "github.com/Azure/azure-storage-blob-go/azblob"
 	"github.com/Azure/go-autorest/autorest/azure/auth"
+	"github.com/go-kit/kit/log"
+	"github.com/go-kit/kit/log/level"
 )
 
 // DirDelim is the delimiter used to model a directory structure in an object store bucket.
@@ -34,24 +36,29 @@ func init() {
 	pipeline.SetForceLogEnabled(false)
 }
 
-func getAzureStorageCredentials(conf Config) (blob.Credential, error) {
+func getAzureStorageCredentials(logger log.Logger, conf Config) (blob.Credential, error) {
 	if conf.MSIResource != "" {
 		msiConfig := auth.NewMSIConfig()
 		msiConfig.Resource = conf.MSIResource
 
-		azureServicePrincipalToken, err := msiConfig.ServicePrincipalToken()
+		spt, err := msiConfig.ServicePrincipalToken()
 		if err != nil {
 			return nil, err
 		}
-
-		// Get a new token.
-		err = azureServicePrincipalToken.Refresh()
-		if err != nil {
+		if err := spt.Refresh(); err != nil {
 			return nil, err
 		}
-		token := azureServicePrincipalToken.Token()
 
-		return blob.NewTokenCredential(token.AccessToken, nil), nil
+		return blob.NewTokenCredential(spt.Token().AccessToken, func(tc blob.TokenCredential) time.Duration {
+			err := spt.Refresh()
+			if err != nil {
+				level.Error(logger).Log("msg", "could not refresh MSI token", "err", err)
+				// Retry later as the error can be related to API throttling
+				return 30 * time.Second
+			}
+			tc.SetToken(spt.Token().AccessToken)
+			return spt.Token().Expires().Sub(time.Now().Add(2 * time.Minute))
+		}), nil
 	}
 
 	credential, err := blob.NewSharedKeyCredential(conf.StorageAccountName, conf.StorageAccountKey)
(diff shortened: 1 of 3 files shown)


pkg/objstore/azure/helpers.go
```diff
@@ -38,20 +38,31 @@ func getAzureStorageCredentials(conf Config) (blob.Credential, error) {
...
 		if err != nil {
 			return nil, err
 		}
 
-		// Get a new token.
+		var tokenRefresher blob.TokenRefresher = func(credential blob.TokenCredential) time.Duration {
+			err := azureServicePrincipalToken.Refresh()
+			if err != nil {
+				// Retry later as the error can be related to API throttling
+				return 30 * time.Second
```
phillebaba, line 50: Its annoying that there is no good error handling in this library. It might be good to print the error to make it visible. Exiting on an error is not a good solution as the error good be temporary of related to an API rate limit, so its a lot better to retry. Any thoughts?

pkg/objstore/azure/helpers.go
```diff
@@ -38,20 +38,31 @@ func getAzureStorageCredentials(conf Config) (blob.Credential, error) {
...
-
 		azureServicePrincipalToken, err := msiConfig.ServicePrincipalToken()
 		if err != nil {
 			return nil, err
 		}
 
-		// Get a new token.
+		var tokenRefresher blob.TokenRefresher = func(credential blob.TokenCredential) time.Duration {
+			err := azureServicePrincipalToken.Refresh()
+			if err != nil {
```
bwplotka, line 48:
```suggestion
			if err := azureServicePrincipalToken.Refresh(); err != nil {
```

pkg/objstore/azure/helpers.go
bwplotka: Yes, We should report that in the library GH issues ASAP.

Maybe there are different error codes, but not documented? 🤗 

Do you think this will work for now? Happy to merge, for now.

pkg/objstore/azure/helpers.go
```diff
@@ -38,20 +38,31 @@ func getAzureStorageCredentials(conf Config) (blob.Credential, error) {
...
+				// Retry later as the error can be related to API throttling
```
bwplotka, line 49: I would 100% add metric or error log to indicate that we hit this and why (err)

pkg/objstore/azure/helpers.go
```diff
@@ -38,20 +38,31 @@ func getAzureStorageCredentials(conf Config) (blob.Credential, error) {
...
+				return 30 * time.Second
+			}
+			token := azureServicePrincipalToken.Token()
+			credential.SetToken(token.AccessToken)
+			// Refresh token two minutes before actual expiration
+			exp := token.Expires().Sub(time.Now().Add(2 * time.Minute))
+			return exp
```
bwplotka, line 56:
```suggestion
			return token.Expires().Sub(time.Now().Add(2 * time.Minute))
```

pkg/objstore/azure/helpers.go
```diff
@@ -38,20 +38,31 @@ func getAzureStorageCredentials(conf Config) (blob.Credential, error) {
...
+		}
+
 		err = azureServicePrincipalToken.Refresh()
```
bwplotka, line 59: ditto

pkg/objstore/azure/helpers.go
```diff
@@ -38,20 +38,31 @@ func getAzureStorageCredentials(conf Config) (blob.Credential, error) {
...
 		msiConfig := auth.NewMSIConfig()
 		msiConfig.Resource = conf.MSIResource
-
 		azureServicePrincipalToken, err := msiConfig.ServicePrincipalToken()
 		if err != nil {
 			return nil, err
 		}
 
-		// Get a new token.
+		var tokenRefresher blob.TokenRefresher = func(credential blob.TokenCredential) time.Duration {
```
bwplotka, line 46: what about defining this function outside? (normally) ;p

bwplotka (dismissed):
Nice, love it. Left some suggestions. Thanks! 💪🏽

pkg/objstore/azure/helpers.go
phillebaba: I added a log to make it visible, hopefully it is enough for now.

pkg/objstore/azure/helpers.go
phillebaba: Cant really do that as it need access to the spt.

phillebaba:
@bwplotka hopefully this solves all comments and issues. I will leave a build of this running for at least 12 hours to make sure that the token refresh works. The only thing I am not super happy about is the token refresh failure but it might be something that we need to re visit in the future instead.

pkg/objstore/azure/helpers.go
bwplotka: yup

bwplotka (approved):
Thanks, LGTM

bwplotka:
Nice, thanks. Can we add issue to the library we use, so they are aware?

phillebaba:
Created an issue in the library, linking it here to make it easier to track.

https://github.com/Azure/azure-storage-blob-go/issues/303

PR #4613, opened 2021-08-30
store: add retry block for bucket initial sync

### LINKED ISSUE #4489: bucket store initial sync: sync block: BaseFetcher: iter bucket: Server not initialized, please try again.
Store initial sync fails on temporary bucket error without retrying.

Bucket: S3.

Repro: I started minio and immediately then Thanos Store and it failed because minio was still starting (despite readiness probe being OK BTW :/ )

someshkoli:

## Changes
Pushes bucket initial sync into a retry block to avoid server shutdown due to objstore unavailability. 
Closes #4489

### DIFF
--- a/cmd/thanos/store.go
+++ b/cmd/thanos/store.go
@@ -361,11 +361,15 @@ func runStore(
 			defer runutil.CloseWithLogOnErr(logger, bkt, "bucket client")
 
 			level.Info(logger).Log("msg", "initializing bucket store")
+			// note(someshkoli): flag required to input interval duration ?
+			runutil.RetryWithLog(logger, time.Second*2, ctx.Done(), func() error {
+				if err := bs.InitialSync(ctx); err != nil {
+					return errors.Wrap(err, "bucket store initial sync")
+				}
+				return nil
+			})
 			begin := time.Now()
-			if err := bs.InitialSync(ctx); err != nil {
-				close(bucketStoreReady)
-				return errors.Wrap(err, "bucket store initial sync")
-			}
+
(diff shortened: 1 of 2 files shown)


cmd/thanos/store.go
```diff
@@ -361,11 +361,19 @@ func runStore(
 			defer runutil.CloseWithLogOnErr(logger, bkt, "bucket client")
 
 			level.Info(logger).Log("msg", "initializing bucket store")
-			begin := time.Now()
-			if err := bs.InitialSync(ctx); err != nil {
-				close(bucketStoreReady)
-				return errors.Wrap(err, "bucket store initial sync")
+			// note(someshkoli): flag required to input interval duration ?
+			if err := runutil.RetryWithLog(logger, time.Second*2, ctx.Done(), func() error {
```
bwplotka, line 365: I would still fail, after some time. Can we set timeout for this? e.g 5m?

cmd/thanos/store.go
```diff
@@ -361,11 +361,19 @@ func runStore(
...
+				if err := bs.InitialSync(ctx); err != nil {
+					return errors.Wrap(err, "bucket store initial sync")
+				}
+				return nil
+			}); err != nil {
+				runutil.CloseWithLogOnErr(logger, bs, "bucket store")
+				return err
 			}
+
+			begin := time.Now()
```
bwplotka, line 375: This does not make sense (:

bwplotka: let's add it above

bwplotka (commented):
Nice, thanks! Some suggestions

someshkoli:
Still valid

PR #4614, opened 2021-08-30
query: fix to respect per-request timeout if specified

spaparaju:
fix for #3456

## Changes

Query component times out the queries under per-request-timeout if specified.
## Verification

- To simulate slow queries, added time.Sleep() at Series() implementation of [underlying Prometheus store](https://github.com/thanos-io/thanos/blob/a57bbfbe4dc4838093c499b5afa72ed305d4eaae/pkg/store/prometheus.go#L135)
- Observed the time-out behaviour in the cases of specifying / not specifying per-request timeout at the Query API calls. 
    - http://localhost:10904/api/v1/query?query=go_memstats_heap_inuse_bytes&step=10s&timeout=1s
     - http://localhost:10904/api/v1/query_range...
     - added a test to check if queries timeout as per per-request timeout if per-request timeout is specified.

### DIFF
--- a/pkg/query/querier.go
+++ b/pkg/query/querier.go
@@ -210,7 +210,7 @@ func (q *querier) Select(_ bool, hints *storage.SelectHints, ms ...*labels.Match
 	// We want to prevent this from happening for the async store API calls we make while preserving tracing context.
 	ctx := tracing.CopyTraceContext(context.Background(), q.ctx)
 	// The timeout value at q.ctx would contain least of {global Query timeout, per-request timeout}.
-	// For q.ctx, 'per-request timeout' is set pkg/api/query/v1.go (query(),query_range()) and later in call chain 'global Query timeout' is set at prometheus/promql/engine.go (exec()).
+	// For q.ctx, 'per-request timeout' is set by Thanos Query component and later replaced by 'global Query timeout' at downstream promql package of upstream Prometheus.
 	requestCancelTime, _ := q.ctx.Deadline()
 	requestTimeOut := time.Until(requestCancelTime)
 	ctx, cancel := context.WithTimeout(ctx, requestTimeOut)
--- a/pkg/query/querier.go
+++ b/pkg/query/querier.go
@@ -209,7 +209,11 @@ func (q *querier) Select(_ bool, hints *storage.SelectHints, ms ...*labels.Match
 	// The querier has a context but it gets canceled, as soon as query evaluation is completed, by the engine.
 	// We want to prevent this from happening for the async store API calls we make while preserving tracing context.
 	ctx := tracing.CopyTraceContext(context.Background(), q.ctx)
-	ctx, cancel := context.WithTimeout(ctx, q.selectTimeout)
+	// The timeout value at q.ctx would contain least of {global Query timeout, per-request timeout}.
+	// For q.ctx, 'per-request timeout' is set pkg/api/query/v1.go (query(),query_range()) and later in call chain 'global Query timeout' is set at prometheus/promql/engine.go (exec()).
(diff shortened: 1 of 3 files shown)


pkg/query/querier.go
```diff
@@ -209,7 +209,11 @@ func (q *querier) Select(_ bool, hints *storage.SelectHints, ms ...*labels.Match
 	// The querier has a context but it gets canceled, as soon as query evaluation is completed, by the engine.
 	// We want to prevent this from happening for the async store API calls we make while preserving tracing context.
 	ctx := tracing.CopyTraceContext(context.Background(), q.ctx)
-	ctx, cancel := context.WithTimeout(ctx, q.selectTimeout)
+	// The timeout value at q.ctx would contain least of {global Query timeout, per-request timeout}.
+	// For q.ctx, 'per-request timeout' is set pkg/api/query/v1.go (query(),query_range()) and later in call chain 'global Query timeout' is set at prometheus/promql/engine.go (exec()).
```
bwplotka, line 213: That comment will obsolete very quickly, I would remove it.

pkg/query/querier.go
```diff
@@ -209,7 +209,11 @@ func (q *querier) Select(_ bool, hints *storage.SelectHints, ms ...*labels.Match
...
+	requestCancelTime, _ := q.ctx.Deadline()
+	requestTimeOut := time.Until(requestCancelTime)
+	ctx, cancel := context.WithTimeout(ctx, requestTimeOut)
```
bwplotka, line 216: I think we don't need this. context we take timeout on is from `ctx := tracing.CopyTraceContext(context.Background(), q.ctx)` which already should have deadline. 

If we remove this code, your test fails?

bwplotka (changes requested):
Thanks! It's clean and readable code, but curious if we really need it. 🤔

pkg/query/querier.go
spaparaju: Updated the comment and the new comment explains the reason for this PR.

pkg/query/querier.go
spaparaju: For q.ctx, 'per-request timeout' is set by [Thanos Query component ](https://github.com/thanos-io/thanos/blob/1205a222cb1e1be4862ef39b4c3a9c538056cf83/pkg/api/query/v1.go#L302)and later replaced by 'global Query timeout' at [downstream 'promql' package](https://github.com/prometheus/prometheus/blob/ad642a85c061443631404a72e7f1ecb4ab350b8e/promql/engine.go#L477) of upstream Prometheus.

pkg/query/querier.go
```diff
@@ -209,7 +209,11 @@ func (q *querier) Select(_ bool, hints *storage.SelectHints, ms ...*labels.Match
 	// The querier has a context but it gets canceled, as soon as query evaluation is completed, by the engine.
 	// We want to prevent this from happening for the async store API calls we make while preserving tracing context.
 	ctx := tracing.CopyTraceContext(context.Background(), q.ctx)
```
bwplotka, line 211: Ok, so I know where is confusion from (👍🏽 
This exact line *on purpose* removes cancellation (plus timeout). Adding it below "maybe" makes sense, but we need to also think if 

> // We want to prevent this from happening for the async store API calls we make while preserving tracing context.

Is still what we want.... 🤔

bwplotka (commented):
Ok, two things left.

One thing in the comment.

Second thing is that I pulled this code and commented out

> //requestCancelTime, _ := q.ctx.Deadline()
> //requestTimeOut := time.Until(requestCancelTime)

And `TestQueryEndpointsForQueryTimeout` still passes 🤔  so either test is wrong or we miss something.

Thanks!

PR #4616, opened 2021-08-31
UI: Fix error only appearing once for failing queries

### LINKED ISSUE #4301: query: Query errors don't get replayed and queries silently fail on second try
**What happened**:
An error in a query displays an error. If that is "clicked away" by clicking the little x, the same error is not displayed again, when the query is run again, even if it is at a later time 

Show the error again

enter falsey query, see error, click little x, run the same query again. 

Version:

<h2></h2>
version | 0.20.0-rc.0
-- | --
0dfd10788c5ffbe5fca5547336ee868319a5119a
HEAD
circleci@c193b4737e52
20210423-12:06:20
go1.16.3

nishidhaSri:
Signed-off-by: Nishidha Sri <nishidhasri@gmail.com>

## Changes

Fixes #4301

Changed state change to null after discarding the error in src/pages/graph/Panel.tsx

## Verification

Tested it by manually firing a wrong query and then discarding it and firing the wrong query again. Did it several times.

### DIFF
--- a/pkg/ui/react-app/src/pages/graph/Panel.tsx
+++ b/pkg/ui/react-app/src/pages/graph/Panel.tsx
@@ -282,6 +282,10 @@ class Panel extends Component<PanelProps & PathPrefixProps, PanelState> {
     this.setOptions({ storeMatches: selectedStores || [] });
   };
 
+  handleToggleAlert = (): void => {
+    this.setState({ error: null });
+  };
+
   render() {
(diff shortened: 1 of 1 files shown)


bwplotka (dismissed):
LGtm, thanks

@onprem do you mind checking this?

onprem (commented):
Nice work! The code looks good, but you have to update the assets file `bindata.go` as well. Just run `make assets` and commit the result.

PR #4620, opened 2021-09-01
block (UI and API): Add mark deletion and no compaction UI

adzshaf:
This is part of LFX mentorship project (issue: #3112). We want to add mark deletion and no compaction button in block details UI.

## Changes
- Add `/block/mark` endpoint for trigger marking block API
- Add button for mark deletion and no compaction in UI, with modal for asking the detail/reason

## Verification

Tested in local using `make quickstart`

Example of mark deletion:
Example Deletion

Result of mark deletion:
image

Example of mark no compaction:
Example No Compaction

Result of mark no compaction:
image

cc: @squat @onprem

### DIFF
--- a/pkg/api/blocks/v1.go
+++ b/pkg/api/blocks/v1.go
@@ -4,7 +4,6 @@
 package v1
 
 import (
-	"context"
 	"net/http"
 	"time"
 
@@ -64,7 +63,7 @@ func (bapi *BlocksAPI) Register(r *route.Router, tracer opentracing.Tracer, logg
 	instr := api.GetInstr(tracer, logger, ins, logMiddleware, bapi.disableCORS)
 
 	r.Get("/blocks", instr("blocks", bapi.blocks))
-	r.Post("/blocks/mark", instr("blocks", bapi.markBlock))
+	r.Post("/blocks/mark", instr("blocks_mark", bapi.markBlock))
 }
 
 func (bapi *BlocksAPI) markBlock(r *http.Request) (interface{}, []error, *api.ApiError) {
--- a/pkg/api/blocks/v1.go
+++ b/pkg/api/blocks/v1.go
@@ -4,16 +4,23 @@
 package v1
 
 import (
+	"context"
 	"net/http"
 	"time"
 
 	"github.com/go-kit/kit/log"
+	"github.com/oklog/ulid"
 	"github.com/opentracing/opentracing-go"
+	"github.com/pkg/errors"
+	"github.com/prometheus/client_golang/prometheus"
+	"github.com/prometheus/client_golang/prometheus/promauto"
 	"github.com/prometheus/common/route"
 	"github.com/thanos-io/thanos/pkg/api"
+	"github.com/thanos-io/thanos/pkg/block"
 	"github.com/thanos-io/thanos/pkg/block/metadata"
 	extpromhttp "github.com/thanos-io/thanos/pkg/extprom/http"
 	"github.com/thanos-io/thanos/pkg/logging"
+	"github.com/thanos-io/thanos/pkg/objstore"
 )
 
 // BlocksAPI is a very simple API used by Thanos Block Viewer.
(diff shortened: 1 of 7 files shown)


pkg/api/blocks/v1.go
```diff
@@ -55,6 +63,42 @@ func (bapi *BlocksAPI) Register(r *route.Router, tracer opentracing.Tracer, logg
...
+		return nil, nil, &api.ApiError{Typ: api.ErrorBadData, Err: errors.New("Action cannot be empty")}
+	}
+
+	id, err := ulid.Parse(idParam)
+	if err != nil {
+		return nil, nil, &api.ApiError{Typ: api.ErrorBadData, Err: errors.Errorf("ULID %q is not valid: %v", idParam, err)}
+	}
+
+	switch actionParam {
+	case "DELETION":
```
bwplotka, line 88: Can move this to Go content so we don't make typo/mistake here?

bwplotka (approved):
It looks pretty sweet, thank your!

I would be happy to merge it TBH, LGTM!

bwplotka:
Do you mind rebasing to fix bindata?

pkg/api/blocks/v1.go
```diff
@@ -55,6 +63,42 @@ func (bapi *BlocksAPI) Register(r *route.Router, tracer opentracing.Tracer, logg
...
+		err := block.MarkForDeletion(r.Context(), bapi.logger, bapi.bkt, id, detailParam, promauto.With(nil).NewCounter(prometheus.CounterOpts{}))
```
yeya24, line 89: We don't need to use a stub counter here. I think it makes sense to use a separate counter to track how many blocks are deleted or marked manually.

yeya24:
What about having adding this to the latest release? @bwplotka

pkg/api/blocks/v1.go
yeya24: cc @adzshaf

yeya24 (commented):
Let's fix the comment, resolve the conflicts and add a changelog entry? Then we are ready to merge it I think.

yeya24:
Just added a commit to resolve conflict. I feel this is ready to go! cc @onprem @adzshaf

pkg/ui/react-app/src/thanos/pages/blocks/BlockDetails.tsx
```diff
@@ -11,6 +11,36 @@ export interface BlockDetailsProps {
...
+        body,
+      });
+
+      if (!response.ok) {
+        throw new Error(response.statusText);
+      }
+      setModalAction('');
+    } catch (error) {
+      setModalAction('');
+    }
```
onprem, line 41:
```suggestion
    } finally {
      setModalAction('');
    }
```

pkg/api/blocks/v1.go
```diff
@@ -32,8 +39,27 @@ type BlocksInfo struct {
...
+	Deletion ActionType = iota
+	NoCompaction
+	Unknown
+)
+
+func parse(s string) ActionType {
+	switch s {
+	case "DELETION":
+		return Deletion
+	case "NO COMPACTION":
```
onprem, line 54:
```suggestion
	case "NO_COMPACTION":
```

Just a suggestion. Whitespace doesn't always play nice in bash scripts etc. so maybe it's better to keep these constants a single word.

onprem (approved):

yeya24:
Sorry for my poor git skill. I rebased the code and all previous commits are also signed off by me, too. Hope this is okay @adzshaf.

I fixed the bindata and tested it working locally. Overall this looks good, but there is some work to fix:

In the dark mode, I expect the button to have a lighter color.
image

Also after I click `Mark Deletion` or `Mark No Compaction`, it works and I can see that the marker created. But in the UI I didn't get any status response to tell me whether the work succeeds or not. This is something important to have.

yeya24 (approved):
LGTM. Let's fix nits in next pr.

PR #4421, opened 2021-07-06
query: add endpointset flow

hitanshu-mehta:
* [ ] I added CHANGELOG entry for this change.
* [ ] Change is not relevant to the end user.

## Changes

- This PR is part of implementing the [Unified Endpoint Proposal.](https://thanos.io/tip/proposals-accepted/202101-endpoint-discovery.md/)
- I have implemented Info gRPC Service in #4282 and also added `--endpoint` flag. But new endpoint flow is baked into the existing `storeset` flow. This can make the migration process difficult as discussed [here](https://github.com/thanos-io/thanos/pull/4282#discussion_r653353874). So I have added new endpoint flow in a separate file.
- Most of the code is copied from `storeset.go` but I have tried to make it more generic.

## Verification

- Added unit tests. ( Again similar to unit tests of storeset but made more generic and covered some additional scenarios relevant to endpoint )

PR #4625, opened 2021-09-01
Cut release 0.23.0-rc.0

bwplotka:

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -10,33 +10,35 @@ We use *breaking :warning:* to mark changes that are not backward compatible (re
 
 ## Unreleased
 
-## v0.23.0 - In Progress
+## [v0.23.0-rc.0](https://github.com/thanos-io/thanos/tree/release-0.23) - 2021.09.02
 
 ### Added
 
-- [#4453](https://github.com/thanos-io/thanos/pull/4453) Tools: Add flag `--selector.relabel-config-file` / `--selector.relabel-config` / `--max-time` / `--min-time` to filter served blocks.
-- [#4482](https://github.com/thanos-io/thanos/pull/4482) COS: Add http_config for cos object store client.
-- [#4487](https://github.com/thanos-io/thanos/pull/4487) Query: Add memcached auto discovery support.
-- [#4444](https://github.com/thanos-io/thanos/pull/4444) UI: Add search block UI.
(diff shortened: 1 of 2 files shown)


wiardvanrij (approved):

GiedriusS:
https://github.com/thanos-io/thanos/pull/4586#discussion_r700735487 I suppose this needs fixing before the release :(

bwplotka:
Yup, that's a small blocker: https://github.com/thanos-io/thanos/pull/4586#discussion_r701162085

bwplotka:
rebased release branch on fixed main 💪🏽 Let's go!

GiedriusS:
Do we need to mention https://github.com/thanos-io/thanos/pull/4421? With that PR we now pick up all of the APIs that a node exposes automatically, right?

bwplotka:
You are absolutely right I missed this. Just checked the code it does not matter indeed 💪🏽

bwplotka:
Double checking now if this will work will old Rule/Metadata etc APIs?

bwplotka:
cc @hitanshu-mehta

bwplotka:
I think there is regression here - if someone points to old Ruler API (e.g Thanos Ruler) in `--rule` - we will assume it's store API

hitanshu-mehta:
@bwplotka Yes, currently we are assuming it exposes both rule and store APIs (see [here](https://github.com/thanos-io/thanos/blob/bd134d7a823708fa135e7a6931e76f581be5f879/pkg/query/endpointset.go#L107)), because previously we were keeping `rule`, `metadata`, `target` and `exemplar` APIs only if we have store API at those endpoints. (see [here](https://github.com/thanos-io/thanos/blob/ff363498fc95cfe17de894d7237bcf38bdd0bc36/pkg/query/storeset.go#L541)).

bwplotka:
I understand, but I  think we can't assume that. One can roll querier without rolling other components. Adding test case and potential fix here:  https://github.com/thanos-io/thanos/pull/4645

bwplotka:
I might wrong, all might be good thanks to `fillExpectedAPIs` 🤔

bwplotka:
Addressed comment, @GiedriusS @hitanshu-mehta and confirmed all is good, no regression is there ([spent ~2h on assuring that = worth it XD!]( https://github.com/thanos-io/thanos/pull/4645)). Thanks for amazing work (:

GiedriusS (approved):

yeya24 (approved):

PR #4626, opened 2021-09-02
Call out the importance of ordering of the hashring configuration

pepov:
See: https://github.com/thanos-io/thanos/discussions/4505

I'm not sure if this proposal should be updated in-line, or should be called out somewhere else, but since this seems to be the authoritative documentation about hashrings I'm opening it here first.

### DIFF
--- a/docs/proposals-done/201812-thanos-remote-receive.md
+++ b/docs/proposals-done/201812-thanos-remote-receive.md
@@ -124,7 +124,7 @@ A sample of the configuration of tenants and their respective infrastructure:
 ]
 ```
 
-To start, exact matches of tenant IDs will be used to distribute requests to receive endpoints. Should it be necessary, more sophisticated mechanisms can be added later. When a request is received, the tenant specified in the request is tested against the configured allowed tenants for each hashring until an exact match is found. If a hashring specifies no explicit tenants, then any tenant is considered a valid match; this allows for a cluster to provide soft-tenancy. Requests whose tenant ID matches no other hashring explicitly, will automatically land in this soft tenancy hashring. If no matching hashring is found and no soft tenancy is configured, the receiver responds with an error.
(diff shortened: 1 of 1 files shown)


docs/proposals-done/201812-thanos-remote-receive.md
```diff
@@ -124,7 +124,7 @@ A sample of the configuration of tenants and their respective infrastructure:
...
+To start hashrings are processed in the order they are defined in the configuration. Exact matches of tenant IDs will be used to distribute requests to receive endpoints in the first matching hashring. Should it be necessary, more sophisticated mechanisms can be added later. When a request is received, the tenant specified in the request is tested against the configured allowed tenants for each hashring until an exact match or until a soft tenant is found. If a hashring specifies no explicit tenants, then any tenant is considered a valid match; this allows for a cluster to provide soft-tenancy. Requests will automatically land in the first soft tenant hashring if their tenant ID matches no other hashring explicitly before the soft tenant hashring. If no matching hashring is found and no soft tenancy is configured, the receiver responds with an error.
```
squat, line 127: `soft` and `hard` tenancy are kind of ill-defined concepts and the concept of soft-tenancy is first introduced *after* you already use it. I suggest we simplify this a bit so it makes more sense to the reader and so we have less repetition:
```suggestion
 Hashrings are processed in the order they are defined in the configuration. Exact matches of tenant IDs will be used to distribute requests to receive endpoints in the first matching hashring. If a hashring specifies no explicit tenants, then any tenant is considered a valid match; this allows for a cluster to provide soft tenancy. Should it be necessary, more sophisticated mechanisms can be added later. When a request is received, the tenant specified in the request is tested against the configured allowed tenants for each hashring until an exact match or until a soft tenant is found.  If no matching hashring is found and no soft tenancy is configured, the receiver responds with an error.
```

PR #4636, opened 2021-09-03
Azure: Support authentication using user-assigned managed identity

amrmahdi:
Adds support for [user-assigned managed identity](https://docs.microsoft.com/en-us/azure/active-directory/managed-identities-azure-resources/overview#managed-identity-types).

When using the `user_assigned_id` the change does not require setting `msi_resource` it defaults to `https://<storage account name>.<endpoint>` unless the config explicitly sets the `msi_resource`.

Tested against Azure.

### DIFF
--- a/pkg/objstore/azure/helpers.go
+++ b/pkg/objstore/azure/helpers.go
@@ -15,6 +15,7 @@ import (
 
 	"github.com/Azure/azure-pipeline-go/pipeline"
 	blob "github.com/Azure/azure-storage-blob-go/azblob"
+	"github.com/Azure/go-autorest/autorest/adal"
 	"github.com/Azure/go-autorest/autorest/azure/auth"
 	"github.com/go-kit/kit/log"
 	"github.com/go-kit/kit/log/level"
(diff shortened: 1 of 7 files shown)


pkg/objstore/azure/helpers.go
```diff
@@ -37,11 +38,27 @@ func init() {
...
+		if resource == "" {
+			resource = fmt.Sprintf("https://%s.%s", conf.StorageAccountName, conf.Endpoint)
+		}
+
+		var (
+			spt *adal.ServicePrincipalToken
+			err error
+		)
+
+		if conf.UserAssignedID != "" {
```
wiardvanrij, line 52: What if there is a function like `getServicePrincipalToken` which just returns the `*adal.ServicePrincipalToken`? That removes those two vars defines and allows for nice returns.

wiardvanrij (approved):
Ah this is nice. Explains the different auth methods much better than I could figure out from the MS docs lol. Readme is spot on and makes a lot of sense. Just a little nitpick/suggestion on the code logic.

pkg/objstore/azure/helpers.go
amrmahdi: Nice suggestion

wiardvanrij (dismissed):

amrmahdi:
@bwplotka Hoping this can get on the 0.23 release treain :)

pkg/objstore/azure/helpers.go
```diff
@@ -68,6 +66,23 @@ func getAzureStorageCredentials(logger log.Logger, conf Config) (blob.Credential
...
 
+func getServicePrincipalToken(logger log.Logger, conf Config) (*adal.ServicePrincipalToken, error) {
+	resource := conf.MSIResource
+	if resource == "" {
+		resource = fmt.Sprintf("https://%s.%s", conf.StorageAccountName, conf.Endpoint)
+	}
+
+	if conf.UserAssignedID != "" {
+		level.Debug(logger).Log("msg", "using user assigned identity:", conf.UserAssignedID)
+		return adal.NewServicePrincipalTokenFromMSIWithUserAssignedID("", resource, conf.UserAssignedID)
```
phillebaba, line 77: Nit: This function is deprecated, instead `NewServicePrincipalTokenFromManagedIdentity` should be used.
https://github.com/Azure/go-autorest/blob/cf74520d3483ff497328aa9f0f0e2ed5e55a70ad/autorest/adal/token.go#L749-L752

amrmahdi: Thanks for the note. Updated to not use the deprecated method. I ended up using the same constructs as the system-assigned identity existing code, and only set the client id if the user assigned identity is specified, which further simplified the code.

GiedriusS (approved):
:+1:

PR #4653, opened 2021-09-10
Tests: Query Frontend - Fix incorrect check in E2E tests

matej-g:

## Changes

Small fix for what seems to be an incorrect condition - in the check func, we should be returning error if there are no results, not when there are results available. This seems to currently always make the E2E test fail.

## Verification

Run tests locally - :heavy_check_mark:


hanjm:
cool.

yeya24 (approved):
LGTM!

PR #4662, opened 2021-09-16
lint: Do not check `*.md` files for white noise

matej-g:

## Changes

`make lint` currently clashes with `make check-docs` due to difference in formatting. To resolve this, I suggest to drop `*.md` from being checked against white noise in the `lint` target, as this should be taken care of in `check-docs` target.

## Verification

Ran `make lint` and `make check-docs` successfully.


matej-g:
Hm OK, so `lint` seems to be conflicting with `check-docs`

matej-g:
I think the solution here is to drop `*.md` files, see the updated description.

wiardvanrij:
Thanks for this. Though I'm somewhat curious as my recent PR did have white noise in the docs. Like the change I made actually looks good and it should be done like that?
I still agree that there is something weird that needs to be fixed tho. WDYT?

matej-g:
@wiardvanrij The issue I tried to solve here is we have two 'competing' targets - both try to format `*.md` files, but in fact it seems we want the formatting done by `check-docs`.

IMHO we should:
1) Ignore `*.md` files as part of the `lint` target (only lint code, not the docs, which makes sense to me)
2) Do whichever formatting on `*.md` we want in the `check-docs` target, including removing whitespace noise (for that, `check-docs` needs to be adjusted, if it currently does not remove whitespaces properly)

wiardvanrij:
> 

yea totally agree. Remove the check on `.md` files here, add it on the `check-docs` and we are good!

matej-g:
It looks like ultimately, it's `mdox fmt` that is adding the extra white spaces. I opened https://github.com/bwplotka/mdox/issues/85, perhaps afterwards there won't be need to run the `.sh` script on top of it.

matej-g:
ping @wiardvanrij does this look good to you now?

wiardvanrij:
> ping @wiardvanrij does this look good to you now?

Yup I think this makes sense! Looks good 👍

Makefile
```diff
@@ -227,7 +227,9 @@ docs: build examples $(MDOX)
 check-docs: ## checks docs against discrepancy with flags, links, white noise.
 check-docs: build examples $(MDOX)
 	@echo ">> checking formatting and local/remote links"
-	PATH=${PATH}:$(GOBIN) $(MDOX) fmt --check -l --links.localize.address-regex="https://thanos.io/.*" --links.validate.config-file=$(MDOX_VALIDATE_CONFIG) $(MD_FILES_TO_FORMAT)
+	PATH=${PATH}:$(GOBIN) $(MDOX) fmt $(MD_FILES_TO_FORMAT)
```
GiedriusS, line 230: Why did we drop `--check -l --links.localize.address-regex="https://thanos.io/.*" --links.validate.config-file=$(MDOX_VALIDATE_CONFIG)` from here? :thinking:

matej-g: :facepalm: Thanks for catching this, I guess I removed it accidentally during some testing. Fixing it right now. It looks like we'll need to drop the `--check` flag still, since otherwise the `mdox` command fails (due to diff in white space handling). However, that should be fine, since below we are checking for clean work tree, which would catch any formatting issues.

matej-g: ping @GiedriusS - hopefully now all is good, the PR might need one more re-run since doc check failed on a link check...

squat: Restarting 👍

squat (approved):

Makefile
matej-g: I found the link it was moved to, made the change, hopefully now it succeeds.

GiedriusS (dismissed):

Makefile
GiedriusS:
```
>> detecting white noise
cannot 'run make docs and commit changes': you have unstaged changes.
M	docs/proposals-accepted/202108-more-granular-query-performance-metrics.md
```

mhm ((:

matej-g: It never ends, does it :smirk:. Fixed now, the formatting also insisted on replacing `_` with `*` in `CHANGELOG.md` so I added that as well.

GiedriusS (approved):

Issue #4579, opened 2021-08-18
e2e tests: Move from Cortex packages to efficientgo/e2e


matej-g:
This relates to issue https://github.com/thanos-io/thanos/issues/3832#issuecomment-785823600, separating it so this change can be followed more clearly.

There is a (nearly) drop-in replacement for Cortex integration test packages  (`github.com/cortexproject/cortex/integration/e2e`) which lives at `github.com/efficientgo/e2e`. In order to achieve one of the steps in aiming to not be dependent on Cortex, the e2e tests should be moved to this new package.

matej-g:
also feel free to assign :point_up: to me

matej-g:
Full resolution is blocked until https://github.com/efficientgo/e2e/issues/7 is fixed

matej-g:
So I'm starting to suspect there might be data race in the compactor, but I haven't been able to chase it down completely. It's not always reproducible, but on a couple of runs, I can make the compactor panic with:

<details><summary>fatal error: concurrent map iteration and map write (full output)</summary>

```
fatal error: concurrent map iteration and map write
goroutine 295 [running]:
runtime.throw({0x1ad3fc9, 0x4627b9})
/usr/local/go/src/runtime/panic.go:1198 +0x71 fp=0xc000ac51b0 sp=0xc000ac5180 pc=0x435651
runtime.mapiternext(0x18)
/usr/local/go/src/runtime/map.go:858 +0x4eb fp=0xc000ac5220 sp=0xc000ac51b0 pc=0x40f82b
reflect.mapiternext(0x195)
/usr/local/go/src/runtime/map.go:1346 +0x19 fp=0xc000ac5238 sp=0xc000ac5220 pc=0x4626f9
reflect.(*MapIter).Next(0xc0007a8960)
/usr/local/go/src/reflect/value.go:1618 +0x99 fp=0xc000ac5258 sp=0xc000ac5238 pc=0x4bcd59
encoding/json.mapEncoder.encode({0x0}, 0xc000128f80, {0x180df80, 0xc000705798, 0x1}, {0x9, 0x0})
/usr/local/go/src/encoding/json/encode.go:799 +0x305 fp=0xc000ac5418 sp=0xc000ac5258 pc=0x5e6885
encoding/json.mapEncoder.encode-fm(0x19b6860, {0x180df80, 0xc000705798, 0xc000ac54b0}, {0x17, 0x0})
[...]

PR #4663, opened 2021-09-16
Fetcher: Fix data races

matej-g:

## Changes
In the process of refactoring E2E tests, it was discovered that some compactor tests are occasionally failing due to a data race (https://github.com/thanos-io/thanos/issues/4579#issuecomment-917716313).

After building the Docker image with race detector enabled and running the compactor E2E tests, it was discovered that there is a number of data races occurring, due to concurrent map write / reading.

This PR attempts to address all the discovered races.

## Verification

Ran compactor tests with race detector enabled on the Thanos binary - :heavy_check_mark:


pkg/block/fetcher.go
```diff
@@ -723,7 +739,10 @@ func (r *ReplicaLabelRemover) Modify(_ context.Context, metas map[ulid.ULID]*met
 			level.Warn(r.logger).Log("msg", "block has no labels left, creating one", r.replicaLabels[0], "deduped")
 			l[r.replicaLabels[0]] = "deduped"
 		}
-		metas[u].Thanos.Labels = l
+
+		nm := *meta
```
bwplotka, line 743: deep copy! 🤗

pkg/block/fetcher.go
```diff
@@ -797,13 +818,23 @@ func NewIgnoreDeletionMarkFilter(logger log.Logger, bkt objstore.InstrumentedBuc
...
+	}
+
+	return deletionMarkMap
 }
 
 // Filter filters out blocks that are marked for deletion after a given delay.
 // It also returns the blocks that can be deleted since they were uploaded delay duration before current time.
 func (f *IgnoreDeletionMarkFilter) Filter(ctx context.Context, metas map[ulid.ULID]*metadata.Meta, synced *extprom.TxGaugeVec) error {
+	f.mtx.Lock()
 	f.deletionMarkMap = make(map[ulid.ULID]*metadata.DeletionMark)
```
bwplotka, line 836: Why is that needed, do we access this map somewhere else?

If that's the case I think we have to build this map locally and only then swap in locked fashion, right?

bwplotka: Otherwise we will have "partial" results

matej-g: It's being accessed in [`DeletionMarkBlocks`](DeletionMarkBlocks). I'm wondering what are the ramifications of having only 'partial' results in the map, shouldn't we pick up any leftovers during the next run of the cleaner func? (Isn't that how it works now, as current version is also potentially allowing to access map with 'partial' results?).

Otherwise you're right, we can also build it separately and swap afterwards while using lock.

bwplotka: I am just assuming there is a reader somewhere else accessing the deletion map concurrently. For a short period instead of 20000 deleted items it would have zero elements and the back to 200 000. I would need to check what exactly wrong can happen with this, but it is just asking for problems here, no? (: Maybe now, things would eventually heal, but tomorrow no. No one would assume that internal map is flaky unless it's called `flakyDeleionMarkMap` across instances with 20 lines of comments on why and still that might be surprising for next writer/reader (:

pkg/block/fetcher.go
```diff
@@ -797,13 +818,23 @@ func NewIgnoreDeletionMarkFilter(logger log.Logger, bkt objstore.InstrumentedBuc
 
 // DeletionMarkBlocks returns block ids that were marked for deletion.
 func (f *IgnoreDeletionMarkFilter) DeletionMarkBlocks() map[ulid.ULID]*metadata.DeletionMark {
-	return f.deletionMarkMap
+	f.mtx.Lock()
+	defer f.mtx.Unlock()
+
+	deletionMarkMap := make(map[ulid.ULID]*metadata.DeletionMark)
```
bwplotka, line 824: 👍🏽

bwplotka:
```suggestion
	deletionMarkMap := make(map[ulid.ULID]*metadata.DeletionMark, len(f.deletionMarkMap))
```

pkg/block/fetcher.go
bwplotka: I would fix this before moving on

bwplotka (commented):
Responded (:

pkg/block/fetcher.go
matej-g: Thanks for clarifying, I see the point now :+1: I adjusted it, please let me know if it looks good now!

pkg/block/fetcher.go
```diff
@@ -871,6 +900,10 @@ func (f *IgnoreDeletionMarkFilter) Filter(ctx context.Context, metas map[ulid.UL
 		return errors.Wrap(err, "filter blocks marked for deletion")
 	}
 
+	f.mtx.Lock()
+	f.deletionMarkMap = deletionMarkMap
```
bwplotka, line 904: 👍🏽

bwplotka (approved):
Thank You! LGTM! 💪🏽

PR #4667, opened 2021-09-16
Add a pure aws-sdk-go auth

sylr:

## Changes

Added an authentication method for S3 which uses the AWS SDK for go.

The AWS SDK handles more authentication methods than the minio-go lib, e.g:

`/etc/thanos/thanos-storage.yml`
```
type: S3
config:
  bucket: "thanos"
  endpoint: "s3.eu-central-1.amazonaws.com"
  aws_sdk_auth: true
  region: "eu-central-1"
  access_key: ""
  insecure: false
  signature_version2: false
  secret_key: ""
  sse_config:
    type: "SSE-KMS"
    kms_key_id: "arn:aws:kms:eu-central-1:99999999999:key/<uuid>"
    kms_encryption_context: {}
    encryption_key: ""
```

`/etc/thanos/aws-config.ini`
```
[profile prometheus]
web_identity_token_file = /var/run/secrets/eks.amazonaws.com/serviceaccount/token
role_arn = arn:aws:iam::111111111111111:role/EKSPrometheusRole

[profile thanos]
source_profile = prometheus
role_arn = arn:aws:iam::99999999999:role/EKSThanosRole
```

Env vars:
```
AWS_CONFIG_FILE=/etc/thanos/aws-config.ini
AWS_PROFILE=thanos
AWS_ARN_ROLE=arn:aws:iam::99999999999:role/EKSThanosRole
```

Here the thanos sidecar running alongside the prometheus container has an AWS session provided via a web identity token file (EKS IRSA feature) and the config file make it so that the `thanos` profile is configured to use the session from the prometheus profile to assume a role in a different account.

This allows to have cross AWS accounts connectivity without the need for access and secret keys (which are baaaaaad).


sylr:
@bwplotka could you please review this ?

pkg/objstore/s3/s3_aws_sdk_auth.go
```diff
@@ -0,0 +1,54 @@
...
+	return credentials.New(&AWSSDKAuth{
+		Region: region,
+	})
+}
+
+// Retrieve retrieves the keys from the environment.
+func (a *AWSSDKAuth) Retrieve() (credentials.Value, error) {
+	cfg, err := awsconfig.LoadDefaultConfig(context.TODO(), awsconfig.WithRegion(a.Region))
+	if err != nil {
+		return credentials.Value{}, fmt.Errorf("unable to load AWS SDK config, %v", err)
```
bwplotka, line 33:
```suggestion
		return credentials.Value{}, errors.Wrap(err, "load AWS SDK config")
```

pkg/objstore/s3/s3_aws_sdk_auth.go
```diff
@@ -0,0 +1,54 @@
...
+	}
+
+	creds, err := cfg.Credentials.Retrieve(context.TODO())
+	if err != nil {
+		return credentials.Value{}, fmt.Errorf("unable to retrieve AWS SDK credentials, %v", err)
```
bwplotka, line 38:
```suggestion
		return credentials.Value{}, errors.Wrap(err, "retrieve AWS SDK credentials")
```

pkg/objstore/s3/s3_aws_sdk_auth.go
```diff
@@ -0,0 +1,54 @@
...
+// AWSSDKAuth retrieves credentials from the aws-sdk-go.
+type AWSSDKAuth struct {
+	Region string
+	creds  aws.Credentials
+}
+
+// NewAWSSDKAuth returns a pointer to a new Credentials object
+// wrapping the environment variable provider.
+func NewAWSSDKAuth(region string) *credentials.Credentials {
+	return credentials.New(&AWSSDKAuth{
```
bwplotka, line 24: Is implementation that complex that is worth to import the aws-sdk-go?

Essentially we need to compare the maintenance work. 

* Is it better for us to copy (potentially tiny) code and maintain it? As far as I understand this, it's just taking correct environment variables, that's it?
* Is it better to contribute such logic to `minio-go`? 
* Is it better to import whole `aws-sdk-go` (dependency hell)

pkg/objstore/s3/s3.go
```diff
@@ -213,7 +214,12 @@ func NewBucketWithConfig(logger log.Logger, config Config, component string) (*B
 	if err := validate(config); err != nil {
 		return nil, err
 	}
-	if config.AccessKey != "" {
+
+	if config.AWSSDKAuth {
+		chain = []credentials.Provider{
+			wrapCredentialsProvider(&AWSSDKAuth{Region: config.Region}),
+		}
+	} else if config.AccessKey != "" {
```
bwplotka, line 222: Can we return error if both accessKey and AWSSDKAuth is set?

bwplotka (commented):
Nice, thanks! Well implemented in, but I am really curious if it's complex enough to import whole new complex module 🤔 I would add issue on `minio-go` too, so they are aware of this gap.

pkg/objstore/s3/s3.go
sylr: Sure

pkg/objstore/s3/s3_aws_sdk_auth.go
sylr: I used `aws-sdk-go-v2` which contrary to the first `aws-sdk-go` has split modules for every services so we should import only what is required for the features we need.

It does much more than taking env vars.

It also leverage the well known AWS config files allowing to automatically assume roles based on other roles via the STS API. All that logic is well and consistently implemented across all AWS SDKs so I see no reason we should have our own implementation.

sylr:
@bwplotka I'm about to fix the conflict, could we merge this before cutting 0.24.0-rc0 ?

sylr:
@bwplotka ?

docs/storage.md
```diff
@@ -93,6 +94,8 @@ config:
 
 At a minimum, you will need to provide a value for the `bucket`, `endpoint`, `access_key`, and `secret_key` keys. The rest of the keys are optional.
 
+However if you set `aws_sdk_auth: true` thanos will use the default authentication methods of the AWS SDK for go based on [known environment variables](https://docs.aws.amazon.com/cli/latest/userguide/cli-configure-envvars.html) (`AWS_PROFILE`, `AWS_WEB_IDENTITY_TOKEN_FILE` ... etc) and known AWS config files (~/.aws/config). If you turn this on then only `bucket` & `endpoint` are required.
```
wiardvanrij, line 97:
```suggestion
However if you set `aws_sdk_auth: true` Thanos will use the default authentication methods of the AWS SDK for go based on [known environment variables](https://docs.aws.amazon.com/cli/latest/userguide/cli-configure-envvars.html) (`AWS_PROFILE`, `AWS_WEB_IDENTITY_TOKEN_FILE` ... etc) and known AWS config files (~/.aws/config). If you turn this on, then the `bucket` and `endpoint` are the required config keys.
```

wiardvanrij (commented):
I'm sort of confident that if you _don't_ use an access_key / secret_key and leave everything empty, you can just use it 'as is'. For example we use KIAM roles that places the AWS env vars and this just works out of the box. Without the need for the AWS SDK. 

It feels this is an implementation for something that already works? I could however be completely missing a feature or point :)

sylr:
The purpose of this is indeed to allow Thanos to work with assumed roles based on all the mechanisms implemented in the AWS SDK.

sylr:
Could we please merge this?

V3ckt0r:
+1 for this. I have this exact usecase as @sylr
> This allows to have cross AWS accounts connectivity without the need for access and secret keys (which are baaaaaad).

pkg/objstore/s3/s3_aws_sdk_auth.go
V3ckt0r: [here](https://aws.github.io/aws-sdk-go-v2/docs/configuring-sdk/#specifying-credentials) is the `LoadDefaultConfig` credential chain we get by importing this.

>>> Environment variables.
Static Credentials (AWS_ACCESS_KEY_ID, AWS_SECRET_ACCESS_KEY, AWS_SESSION_TOKEN)
Web Identity Token (AWS_WEB_IDENTITY_TOKEN_FILE)
Shared configuration files.
SDK defaults to credentials file under .aws folder that is placed in the home folder on your computer.
SDK defaults to config file under .aws folder that is placed in the home folder on your computer.
If your application uses an ECS task definition or RunTask API operation, IAM role for tasks.
If your application is running on an Amazon EC2 instance, IAM role for Amazon EC2.

GiedriusS (dismissed):
Looks good :+1: the logic sounds good to me in https://github.com/thanos-io/thanos/pull/4667/files#r749424917 :thinking:

GiedriusS:
Could you please rebase on `main` and add the DCO?

docs/storage.md
```diff
@@ -99,6 +100,8 @@ config:
 
 At a minimum, you will need to provide a value for the `bucket`, `endpoint`, `access_key`, and `secret_key` keys. The rest of the keys are optional.
 
+However if you set `aws_sdk_auth: true` Thanos will use the default authentication methods of the AWS SDK for go based on [known environment variables](https://docs.aws.amazon.com/cli/latest/userguide/cli-configure-envvars.html) (`AWS_PROFILE`, `AWS_WEB_IDENTITY_TOKEN_FILE` ... etc) and known AWS config files (~/.aws/config). If you turn this on, then the `bucket` and `endpoint` are the required config keys.
```
GiedriusS, line 103: Last comment. This says:
> If you turn this on, then the `bucket` and `endpoint` are the required config keys.

We have this function: https://github.com/thanos-io/thanos/blob/6b6a39f2a6073c44a434728dc9e958b1da333dd7/pkg/objstore/s3/s3.go#L343-L366

Do we need to make any adjustments to it? Because it seems like this sentence doesn't match up with that code :thinking:

sylr: I guess we can move https://github.com/thanos-io/thanos/pull/4667/files#diff-03a36882ad7e01567144dad0fb3742a57a0c84a9d2dbc3c3077e941d2e7ff67cR229-R231 in validate ?

sylr: Done.

sylr:
Is everything good for you @GiedriusS ?

GiedriusS (approved):
LGTM

PR #4672, opened 2021-09-17
scripts: Install grpcurl@latest when querying a store API

fpetkovski:
Running `scripts/insecure_grpcurl_series.sh` yields the following error:
```
no required module provides package github.com/fullstorydev/grpcurl/cmd/grpcurl; to add it:
	go get github.com/fullstorydev/grpcurl/cmd/grpcurl
```

This PR resolves it by adding the `@latest` prefix to the go install argument

## Changes

* Modify `scripts/insecure_grpcurl_series.sh` to install `github.com/fullstorydev/grpcurl/cmd/grpcurl@latest` instead of `github.com/fullstorydev/grpcurl/cmd/grpcurl`
## Verification

I ran `scripts/insecure_grpcurl_series.sh` locally.

### DIFF
--- a/scripts/insecure_grpcurl_series.sh
+++ b/scripts/insecure_grpcurl_series.sh
@@ -38,7 +38,7 @@ if [ -z "${REQUESTED_MAX_TIME}" ]; then
   exit 1
 fi
 
-go install github.com/fullstorydev/grpcurl/cmd/grpcurl
+go install github.com/fullstorydev/grpcurl/cmd/grpcurl@v1.8.2
 
 SERIES_REQUEST='{
   "min_time": '${REQUESTED_MIN_TIME}',


scripts/insecure_grpcurl_series.sh
```diff
@@ -38,7 +38,7 @@ if [ -z "${REQUESTED_MAX_TIME}" ]; then
   exit 1
 fi
 
-go install github.com/fullstorydev/grpcurl/cmd/grpcurl
```
bwplotka, line 41: We don't like taking latest as there is high chances it won't work the same and did not break compatibility 🙃 Can we pin to latest version instead? (or some version)

fpetkovski: Sure thing, done :+1:

GiedriusS (dismissed):
:+1: LGTM, could you please fix the DCO?

fpetkovski:
Thanks for the review, done :+1:

GiedriusS (approved):

PR #4680, opened 2021-09-19
Query: add exemplar.partial-response flag to control partial response (#4676)

hanjm:
Fix #4676.
Signed-off-by: hanjm <hanjinming@outlook.com>

* [ ] I added CHANGELOG entry for this change.
* [x] Change is not relevant to the end user.

## Verification

test #4676.

PR #4681, opened 2021-09-19
Query: Fix (*Stream).receive infinite loop when target response Unimplemented error (#4676)

### LINKED ISSUE #4676: Exemplar: Can not set partial response.

**Object Storage Provider**:
COS
**What happened**:
query_exemplar response error: 
error: "retrieving exemplars: proxy Exemplars: receiving exemplars from exemplars client &{0xc000b2a000}: rpc error: code = Unimplemented desc = unknown service thanos.Exemplars"

partial response with warning.
Thanos Query 0.23 beta + a old version sidecar not support exemplar

seems missing flag to control exemplar partial reponse

hanjm:
Fix #4676.
This PR base on #4680.

grpc doc say https://github.com/grpc/grpc-go/blob/v1.40.0/stream.go#L123
```
	// RecvMsg blocks until it receives a message into m or the stream is
	// done. It returns io.EOF when the stream completes successfully. On
	// any other error, the stream is aborted and the error contains the RPC
	// status.
```

## Changes

if gRPC bid-stream response error, `return`.

## Verification

test in #4676

### DIFF
--- a/pkg/exemplars/proxy.go
+++ b/pkg/exemplars/proxy.go
@@ -191,8 +191,8 @@ func (stream *exemplarsStream) receive(ctx context.Context) error {
 			if err := stream.server.Send(exemplarspb.NewWarningExemplarsResponse(err)); err != nil {
 				return errors.Wrapf(err, "sending exemplars error to server %v", stream.server)
 			}
-
-			continue
+			// Not an error if response strategy is warning.
+			return nil
 		}
 
 		if w := exemplar.GetWarning(); w != "" {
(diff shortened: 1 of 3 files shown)


hanjm:
Hi @yeya24 @bwplotka , This is a import fix to compatible with old store/sidecar, could you like to help review? thank you.

yeya24 (approved):
LGTM! Thanks for fixing this. @hanjm

bwplotka:
Thanks a lot!

PR #4687, opened 2021-09-21
Implement objectSizer for minio object and timingReadCloser

yeya24:

Running thanos bucket replicate to replicate data between two S3 buckets. I got:

```
level=warn ts=2021-09-21T07:25:43.073055Z caller=s3.go:447 component=replicate msg="could not guess file size for multipart upload; upload might be not optimized" name=01FFPW8PXW9TRWC0PGKFHCNFYM/chunks/000001 err="unsupported type of io.Reader: *objstore.timingReadCloser"
```

To fix this, I implemented ObjectSizer interface for `*objstore.timingReadCloser` support `*minio.Object` when checking file size.

### DIFF
--- a/pkg/objstore/objstore.go
+++ b/pkg/objstore/objstore.go
@@ -6,15 +6,15 @@ package objstore
 import (
 	"bytes"
 	"context"
-	"github.com/minio/minio-go/v7"
 	"io"
 	"os"
 	"path/filepath"
 	"strings"
 	"time"
 
-	"github.com/go-kit/log"
-	"github.com/go-kit/log/level"
+	"github.com/go-kit/kit/log"
+	"github.com/go-kit/kit/log/level"
+	"github.com/minio/minio-go/v7"
 	"github.com/pkg/errors"
 	"github.com/prometheus/client_golang/prometheus"
 	"github.com/prometheus/client_golang/prometheus/promauto"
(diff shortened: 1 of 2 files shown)


pkg/objstore/objstore.go
```diff
@@ -145,6 +146,12 @@ func TryToGetSize(r io.Reader) (int64, error) {
 		return int64(f.Len()), nil
 	case *strings.Reader:
 		return f.Size(), nil
+	case *minio.Object:
```
hanjm, line 149: 💪🏻   Could add a unit test?

hanjm: pkg/objstore is a abstract package for all of objectstore, i think it should not assert to a certain implement. another way to solve the replicate issue is implement `objstore.ObjectSizer` for this . like the PR #4687.
"[pkg/objstore](https://github.com/thanos-io/thanos/pull/4687/files/026a4b7276cbb4fbb5d41bc454c799f388ee8210#diff-41e5eeb71bb7097b6a59cdc9eb0002efd60df6b42394773d6d444eb8fec80676)"

```
	// Add size info into reader for pass it to Upload function.
	r := objectSizerReadCloser{ReadCloser: resp.Body, size: resp.ContentLength}
	return r, nil
}

type objectSizerReadCloser struct {
	io.ReadCloser
	size int64
}

// ObjectSize implement objstore.ObjectSizer.
func (o objectSizerReadCloser) ObjectSize() (int64, error) {
	return o.size, nil
}
```

yeya24: Makes sense. Thank you!

PR #4690, opened 2021-09-22
sidecar, receive: Fixed label matching using matchers + external labels; unify tests and behaviour.

bwplotka:
It was super inconsitent. 

There might be still inconsistencies across other store APIs but fixing sidecar and receive first 🤗

### DIFF
--- a/pkg/store/acceptance_test.go
+++ b/pkg/store/acceptance_test.go
@@ -1,0 +1,239 @@
+// Copyright (c) The Thanos Authors.
+// Licensed under the Apache License 2.0.
+
+package store
+
+import (
+	"context"
+	"testing"
+	"time"
+
+	"github.com/pkg/errors"
+	"github.com/prometheus/prometheus/pkg/labels"
+	"github.com/prometheus/prometheus/pkg/timestamp"
+	"github.com/prometheus/prometheus/storage"
+	"github.com/thanos-io/thanos/pkg/store/storepb"
+	"github.com/thanos-io/thanos/pkg/testutil"
+)
+
+type labelNameCallCase struct {
+	matchers []storepb.LabelMatcher
+	start    int64
+	end      int64
+
+	expectedNames []string
+	expectErr     error
+}
+
+type labelValuesCallCase struct {
+	label string
+
+	matchers []storepb.LabelMatcher
+	start    int64
+	end      int64
+
+	expectedValues []string
+	expectErr      error
+}
+
+// testLabelAPIs tests labels methods from StoreAPI from closed box perspective.
+func testLabelAPIs(t *testing.T, startStore func(extLset labels.Labels, append func(app storage.Appender)) storepb.StoreServer) {
+	t.Helper()
+
+	now := time.Now()
+	extLset := labels.FromStrings("region", "eu-west")
+	for _, tc := range []struct {
+		desc             string
+		appendFn         func(app storage.Appender)
+		labelNameCalls   []labelNameCallCase
+		labelValuesCalls []labelValuesCallCase
+	}{
+		{
+			desc: "no label in tsdb, empty results",
+			labelNameCalls: []labelNameCallCase{
+				{start: timestamp.FromTime(minTime), end: timestamp.FromTime(maxTime)},
+			},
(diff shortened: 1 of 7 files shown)


ianbillett:
How was this broken / inconsistent before? I'm afraid I'm struggling to review as I don't know what the impact of the changes you are making are :see_no_evil:

bwplotka:
TL:DR: Tests I create were failing for different reasons. Mainly stuff like:

`Give me all labels from Comonent A with external labels A =B when I match A=B` - should give me all labels, but in past for sidecar and receiver it would return nothing, because external labels are not part of stored data

pkg/store/acceptance_test.go
```diff
@@ -0,0 +1,236 @@
...
+					label:    "bar",
+					matchers: []storepb.LabelMatcher{{Type: storepb.LabelMatcher_EQ, Name: "region", Value: "different"}},
+				},
+			},
+		},
+	} {
+		t.Run(tc.desc, func(t *testing.T) {
+			appendFn := tc.appendFn
+			if appendFn == nil {
+				appendFn = func(storage.Appender) { return }
```
ianbillett, line 190: This is the line causing the linter to [fail](https://github.com/thanos-io/thanos/pull/4690/checks?check_run_id=3674773050#step:6:16).

ianbillett (dismissed):
LGTM - the extensive set of tests that were failing and are now passing makes me confident in this change.

The only final thing I would ask is to create an issue to make this change to the other Store API implementations.

bwplotka:
Created https://github.com/thanos-io/thanos/issues/4693 🤗

pkg/store/prometheus.go
```diff
@@ -545,42 +552,54 @@ func (p *PrometheusStore) LabelNames(ctx context.Context, r *storepb.LabelNamesR
 		}
 	}
 
+	if len(lbls) > 0 {
+		for _, lbl := range extLset {
+			lbls = append(lbls, lbl.Name)
```
matej-g, line 557:
```suggestion
		for _, extLbl := range extLset {
			lbls = append(lbls, extLbl.Name)
```
Nit: To make it obvious what is being appended to what

pkg/store/prometheus.go
```diff
@@ -545,42 +552,54 @@ func (p *PrometheusStore) LabelNames(ctx context.Context, r *storepb.LabelNamesR
...
+		return nil, status.Error(codes.InvalidArgument, err.Error())
+	}
+	if !match {
+		return &storepb.LabelValuesResponse{Values: nil}, nil
+	}
+
 	var (
 		sers []map[string]string
-		err  error
+		lvc  = false // LabelValuesCall
```
matej-g, line 588: Nit: Would be cool to add same explanation above to `lnc` - looking first time at the code I needed a minute to connect the dots. Ideally even having full name `isLabelValueCall` instead of a comment would be nicer :nail_care:.

pkg/store/prometheus_test.go
```diff
@@ -595,16 +416,15 @@ func TestPrometheusStore_Series_MatchExternalLabel_e2e(t *testing.T) {
 	}, srv.SeriesSet[0].Labels)
 
 	srv = newStoreSeriesServer(ctx)
-	// However it should not match wrong external label.
-	err = proxy.Series(&storepb.SeriesRequest{
+	// Howeve,r it should not match wrong external label.
```
matej-g, line 419:
```suggestion
	// However it should not match wrong external label.
```

matej-g (approved):
Changes are looking good! :star: Optional nits.

pkg/store/prometheus.go
bwplotka: Yea, I don't like those cryptic names

bwplotka: Will fix

bwplotka:
Fixed PTAL (:

matej-g (dismissed):

matej-g (approved):

ianbillett (approved):

GiedriusS (commented):
Thx, looks good. You merged just before I finished reviewing :+1:

PR #4696, opened 2021-09-23
Query: Put cache name on tracing spans

bboreham:
With multiple caches, this change lets you see clearly which spans relate to which cache.

## Changes

Added a name member to caches, add this as a tag when tracing.

## Verification

Left as an exercise.


pracucci (commented):
Makes sense to me!

Could you fix the cache mock in tests, please?
```
pkg/store/cache/caching_bucket_test.go:227:22: cannot use cache (type *mockCache) as type cache.Cache in argument to cfg.CacheGetRange:
	*mockCache does not implement cache.Cache (missing Name method)
```

metalmatze (approved):

PR #4702, opened 2021-09-25
Store: Fix (*model.TimeOrDurationValue).String() when value is negative

hanjm:
Signed-off-by: Jimmiehan <hanjinming@outlook.com>

before this fix, set --min-time or --max-time negative value will be empty at page. beacuse prometheus/common model.Duration does not support negative. https://github.com/prometheus/common/blob/main/model/time.go#L236

image

fixed:
image

## Changes

`func (tdv *TimeOrDurationValue) String()` support negative.
## Verification

Unit test.


yeya24 (approved):
LGTM! Thanks for catching this.

GiedriusS (approved):
:muscle:

PR #4710, opened 2021-09-27
[Store] Metric to capture timestamp of the last loaded block

spaparaju:
Signed-off-by: SriKrishna Paparaju <paparaju@gmail.com>

This PR adds metric to capture timestamp of the last loaded block.

## Verification

Added unit test logic, tested thru 'TestBucketStore_e2e() of github.com/thanos-io/thanos/pkg/store'


pkg/store/bucket_e2e_test.go
```diff
@@ -207,10 +210,37 @@ func prepareStoreWithTestBlocks(t testing.TB, dir string, bkt objstore.Bucket, m
 	ctx, cancel := context.WithTimeout(context.Background(), 30*time.Second)
 	defer cancel()
 
+	// Check if the blocks are being loaded.
+	start := time.Now()
+	time.Sleep(time.Second * 1)
```
matej-g, line 215: Do we need a test for this? Do we test other metrics somewhere? To me personally it just feels like we're introducing more complexity to the test while maybe not gaining that much value.

matej-g (commented):
I think it looks fine :+1: But maybe could you give us a bit more context why this is required? What is the use case for it etc.?

pkg/store/bucket.go
```diff
@@ -145,6 +146,11 @@ func newBucketStoreMetrics(reg prometheus.Registerer) *bucketStoreMetrics {
 		Name: "thanos_bucket_store_blocks_loaded",
 		Help: "Number of currently loaded blocks.",
 	})
+	m.lastLoadedBlock = promauto.With(reg).NewGauge(prometheus.GaugeOpts{
+		Name: "thanos_bucket_store_blocks_last_loaded",
+		Help: "Timestamp when last block got loaded.",
+	})
+	m.lastLoadedBlock.Set(0)
```
ianbillett, line 153: I think people will get confused if we set this to `0` - on start up we will report that the last loaded block was at the unix epoch.

spaparaju: I was in two minds :- start the clock from 1970 (or) start with lowest possible Unix timestamp. 

👍  Updated the code to start the initial timestamp to reflect Jan. 1st of 1970.

pkg/store/bucket_e2e_test.go
spaparaju: Well, this is one way to test if the newly added metric `thanos_bucket_store_blocks_last_loaded` is working as expected.

spaparaju:
> I think it looks fine 👍 But maybe could you give us a bit more context why this is required? What is the use case for it etc.?

Good point! Updated the PR description.

pkg/store/bucket_e2e_test.go
matej-g: Yes I understand the intention. I'm just not sure whether this needs testing (hence my question if we're testing other metrics). I'll leave the decision to you if other reviewers do not share this feeling.

ianbillett:
In the Thanos UI stores tab, each store already reports the `min_time` which (_I think_) comes from Store's [InfoResponse RPC](https://github.com/thanos-io/thanos/blob/bdf7e219d0b7495d5e1cf781c9b82426be2d92ea/pkg/store/storepb/rpc.proto#L81). That's the exact piece of information we want to expose here. Is that what your change is exposing here?

In an ideal world, each component that exposes the Store API would expose this information. Users deploy Thanos in all kinds of configuration, it would be useful to have this information available even if they are deploying components other than Store.

Issue #305, opened 2018-04-20
Query Pushdown


jacksontj:
Hello thanos people! I just ran into your project the other day and wanted to chat with someone about some potential performance improvements for the `Querier` nodes. Having read a little through the docs it sounds like the querier nodes basically just aggregate data from N prometheus servers into a single endpoint. I have a project https://github.com/jacksontj/promxy which does exactly that. I started back in October of last year, and since then have a lot of perf learnings (specifically mechanisms to reduce the query-load on the prom system).

I'll start by stating some context, to help center the rest of the feedback. The first version of Promxy that I wrote was a simple implementation of the Storage interface for prom. This was then attached to an engine and the API and *tada* remote querier. The only real magic in that first version was how to merge datasets between hosts etc.

From my reading of the code and docs it seems that Thanos' Query service does the same thing.

There are a variety of optimizations I've done since that first version but the main 2 improvements I've done since are:

~**1. Marshaling Performance**
In the case of a remote querier the amount of data being marshaled and shipped across the wire is a major concern. Since you are starting with the RemoteRead API I expect the marshaling latency isn't going to be a huge issue (its fast and code-gen). The main concern here is the memory pressure this puts on prometheus itself. On the JSON side this was [**really** bad](https://github.com/prometheus/prometheus/issues/3601) -- but the issue is similar on the proto side of things-- specifically that when you ask for N bytes of data that consumes at minimum 2*N bytes of memory on the prometheus server (in the JSON case it was 14x, so proto is in a much better place). To solve this for the proto side would require implementing streaming to an io.Writer in the proto libraries. In my experience this was the most important feature not *so* much because of the performance, but because of the stability-- without this feature you can easily exhaust memory on a remote prometheus host with a seemily "safe" query.~

EDIT(@bwplotka): This was done already.

**2. NodeReplacer API**
[...]

PR #4712, opened 2021-09-29
query: add initial time-based query pushdown

GiedriusS:
Add time-based pushdown mechanism. It works by directly querying a
Thanos Store node if it is the only one partially or fully matching the
time range of a query. I believe that this will be the case most of the
times because typically Sidecar/Ruler/Receive cover a few days of time
range and horizontally balanced Thanos Store instances cover the
historical data from remote object storage. With query-frontend in front
splitting the queries into smaller parts means that with higher time ranges, most of
the queries can be simply pushed down. For example, with Sidecars
covering 2 days of data, with a 7d query, and a split interval of 1d
this means that 5/7 * 100 = ~71% of processing can be pushed down to the
leaf nodes.

Small ad-hoc tests which touch a lot of timeseries show a 2x reduction
in query duration.

Hide this under an "experimental feature" flag
because this still doesn't have everything implemented intentionally to have a smaller change:

* `lazySeriesSet` for concurrent `Select()`s in Thanos Store
* Missing some of the Prometheus engine flags
* Support the selection of StoreAPI nodes in the UI (use only them, not
all endpoints)
* No chunks/series limiters

Also, this API could be potentially implemented in other components.

If the approach looks good then we can merge this and implement these in follow-up PRs.

The main parts of this change is in:

* New QueryAPI service
* New `pkg/pushdown/`
* API request handling code


yeya24 (commented):
Love this work! Can you add another E2E test case between query and store gateway with and without query pushdown? Is the returned result the same for range query?

yeya24:
We are filtering stores using the given time range in the API. But when promql engine performs the Select call, will it use the same time range given from the API?

https://github.com/prometheus/prometheus/blob/f103acd5135b8bbe885b17a73dafc7bbb586319c/promql/engine.go#L762
From this function, IIUC the time range will be adjusted based on the promql queries. WDYT?

GiedriusS:
> We are filtering stores using the given time range in the API. But when promql engine performs the Select call, will it use the same time range given from the API?
> 
> https://github.com/prometheus/prometheus/blob/f103acd5135b8bbe885b17a73dafc7bbb586319c/promql/engine.go#L762 From this function, IIUC the time range will be adjusted based on the promql query itself. WDYT?
> 
> If we can extract this logic from promql, then maybe it is possible to pushdown instant queries as well.

That's true but I'd like to start simple and cover only one case at the beginning as in this PR. Maybe you could add this idea to https://github.com/thanos-io/thanos/issues/305?

pkg/query/endpointset.go
```diff
@@ -146,6 +147,7 @@ func (es *grpcEndpointSpec) fillExpectedAPIs(componentType component.Component,
 				MinTime: mintime,
 				MaxTime: maxTime,
 			},
+			Query: &infopb.QueryInfo{},
```
GiedriusS, line 150: Note that without `--endpoint` it means that we assume that all Stores implement QueryAPI :s

pkg/pushdown/querypb/rpc.proto
```diff
@@ -0,0 +1,35 @@
...
+option (gogoproto.marshaler_all) = true;
+option (gogoproto.unmarshaler_all) = true;
+option (gogoproto.goproto_getters_all) = false;
+// Do not generate XXX fields to reduce memory footprint and opening a door
+// for zero-copy casts to/from prometheus data types.
+option (gogoproto.goproto_unkeyed_all) = false;
+option (gogoproto.goproto_unrecognized_all) = false;
+option (gogoproto.goproto_sizecache_all) = false;
+
+service Query {
```
bwplotka, line 19: OK, I think the main question we have to ask ourselves - do we need another API, let me explain on main comment.

bwplotka (commented):
OK, I think the main question we have to ask ourselves - do we need another API.

The thing is that StoreAPI is well equipped for the work - the pipeline is ready for doing global eval, local eval or mixed. Let's go through different executions for query `count(metric1{target=ext1})`. This makes PromQL engine selects in Querier  `metric1{target=ext1}` (and with `count` function in hints). Then Querier proxy code (StoreAPI querier implementation) fanouts to all matching stores. And let's say for simple case. There is only ONE matching store. And no duplicated metrics. We can then:

1.  Ask this matching store for non-pushdown StoreAPI call.  This will return all selected `metric1` series from remote store. We pass that as it is to PromQL engine in StoreSet. Then PromQL counts them up and returns result.

2. But in the same time, we can do a special StoreAPI.Series call  with e.g pushdown=true field that will tell the underlying store  to give us count of all metric1 series. So it will return us single series 
 for `count(metric1{target=ext1})` as long as we know there are no duplicates across fanouts, we can put that as a result to PromQL engine. PromQL engine will take that as only series count it (but it's only one, so fine) and return to response.

We can enhance 2 and even pass step so only necessary samples are returned that would be use by PromQL. 

I know it might be complex, so maybe a second method to service makes sense, but I would really really like to avoid another API - too much complexity IMO 🤔

bwplotka:
If we would just  expose QuerAPI through gRPC - this might work, but only for super limited cases, where is only one Store matched. We block (or need separate path) for cases where we have `count, sum` etc function and multiple stores.. I think this is bit of missed opportunity - and we sacrifice a huge complexity if we would have now: `QueryAPI, StoreAPI and StorePushdownAPI`... etc

yeya24:
> But in the same time, we can do a special StoreAPI.Series call with e.g pushdown=true field that will tell the underlying store to give us count of all metric1 series. So it will return us single series
for count(metric1{target=ext1}) as long as we know there are no duplicates across fanouts, we can put that as a result to PromQL engine. PromQL engine will take that as only series count it (but it's only one, so fine) and return to response.

PromQL engine only gets series set as result and it will try to apply `count` again. How to bypass this and tell it the result was already calculated? I feel if we don't have another API other than `Select` we cannot achieve query pushdown.

bwplotka:
What else do you want than `series` result as input to PromQL? (: If global PromQL can take raw series but also pre-calculated series from another PromQL engine, in same way, no? (:

GiedriusS:
> OK, I think the main question we have to ask ourselves - do we need another API.
> 
> ...
> I know it might be complex, so maybe a second method to service makes sense, but I would really really like to avoid another API - too much complexity IMO thinking

I understand you and remember this idea from your design document but are these two things exclusive to each other? I really think that we can have both. Just like most of the databases do optimizations at many layers, I think we should do both (and much more). The optimization in this PR is dead simple and quite trivial - the only complex part IMO is implementing the PromQL engine.

To be honest, there is probably no clear-cut answer. But this pull request would solve already many problems in at least two big deployments of Thanos that I know of. Implementing pushdown has been the longest time on our backlog but it has never been properly implemented - perhaps that is because the idea in its entirety is quite complex and we must divide it into smaller parts as in this PR?

Also, I understand that we should strive to implement as generic solutions as possible to ultimately reduce the complexity so that it would be easier to understand the code and maintain it but also on the other hand IMHO sometimes optimization come in smaller pieces, and two extra function calls in `queryRange()` doesn't seem like a huge overhead at all. 

@yeya24 what's your opinion on this topic?

yeya24:
> Also, I understand that we should strive to implement as generic solutions as possible to ultimately reduce the complexity so that it would be easier to understand the code and maintain it but also on the other hand IMHO sometimes optimization come in smaller pieces, and two extra function calls in queryRange() doesn't seem like a huge overhead at all.
@yeya24 what's your opinion on this topic?

They can coexist and I think it is okay to have a separate API for pushdown for now as it is under an experimental feature flag. In the short term, it is hard to hack the promql engine to achieve pushdown directly.

What we need to do is to ensure the results consistency between pushdown and normal query. Continuous benchmarking or tests would help so that we can compare the performance.

yeya24:
> Both are just series with some samples, nothing else.

Right, I understand the final results are just some samples. But the promql engine will apply the promql functions, aggregators again even if some series are pre-aggregated. 
I think that's why we need a new API to bypass the series post-processing from the engine. If it is possible to have a hook at the engine side then that would be great.

GiedriusS:
Could we split this PR then into smaller ones? I could start with a pull request that adds the query engine to Thanos Store with tests. Maybe that would be easier?

bwplotka:
Anyway, I think we are still not sure if we need a different API or not. That is what blocks us from developing on main IMO. I would keep it on the feature branch still. We can try different approaches and compare.  I would move this PR as it is into one Thanos pushdown branch (not your forks's PR), then I can try reusing Series and we can experiment. You see I would love to avoid extra API because of complexities like https://github.com/thanos-io/thanos/issues/4780

We would need to implement all twice and it's not trivial code to fan out all. The SIngle pipeline would be nice, if possible.

GiedriusS:
> > But the promql engine will apply the promql functions, aggregators again even if some series are pre-aggregated.
> 
> That is totally fine, no? We need to do that in order to aggregate pre-aggregated results (:

It's not fine to do that with functions. For example, if you do `sum_over_time`, where step < range, and then apply `sum_over_time` on top then the result will be the double of `sum_over_time`. :P

PR #4714, opened 2021-09-30
Endpointset: Do not use info client to obtain metadata (for now)

### LINKED ISSUE #4699: Endpoint set on query is causing increase in gRPC error rate
We have recently switched to Thanos `v0.23.0-rc.0` for all our components on all our environments, which includes introduction of [endpoint set flow for query](https://github.com/thanos-io/thanos/pull/4421). In our `query`, we are specifying the `--store` flags by leveraging the `dnssrv+` directive. It points to a number of different components - `store`, `rule` and `receive`.

Not long after, we have noticed an increase in alerts firing for [ThanosQueryGrpcClientErrorRate](https://github.com/thanos-io/thanos/blob/main/examples/alerts/alerts.md#query). After an investigation, we've noticed that most of these errors have error code `Unimplemented`. After looking at the code, we are suspecting this is coming from the [`Info` gRPC call](https://github.com/thanos-io/thanos/blob/0d524a91ff19320ce183d404254efcd03eb8643f/pkg/query/endpointset.go#L79) which is part of the attempt to obtain metadata. However, since we also have a some components specified in the stores, which yet do not implement the `Info` gRPC service, these are returning the `Unimplemented` error back after attempting the gRPC call (and only subsequently proceeding to use store API to obtain the data as fallback).

For illustration here's the graph for `Unimplemented` error rate we've been seeing since deploying `v0.23.0-rc.0` on 9th September:
image

All of this is causing 'false positives' due to the `Unimplemented` errors, causing the raise in error rate and triggering the alerts. I guess after Info API is implemented for all components in https://github.com/thanos-io/thanos/pull/4282#issuecomment-925778837 this should no longer be a problem, but I'm wondering whether there is any interim measure. For example, if we could do `Info` call only on components which are known to implement it at this time, otherwise go straight to back-up solution in form of Store API call.

matej-g:

## Changes

In #4699 it was discovered that users might be receiving 'false' reports due to increase in number of `Unimplemented` error, caused by attempting to obtain metadata via gRPC service, which is not yet implemented.

The fix proposed here is to remove info client for now and instead always fall back to using store API to obtain metadata. After https://github.com/thanos-io/thanos/pull/4282 we should be able to switch back to using info client.

The PR also adjusts the tests accordingly and fixes some discrepancies (e.g. the mocked rule component was not exposing store API)

Resolves #4699

## Verification

Updated tests.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -10,6 +10,10 @@ We use *breaking :warning:* to mark changes that are not backward compatible (re
 
 ## Unreleased
 
+### Fixed
+
+- [#4714](https://github.com/thanos-io/thanos/pull/4714) Endpointset: Do not use info client to obtain metadata.
+
 ## [v0.23.0](https://github.com/thanos-io/thanos/tree/release-0.23) - 2021.09.23
 
 ### Added
--- a/pkg/query/endpointset.go
+++ b/pkg/query/endpointset.go
@@ -77,14 +77,16 @@ func (es *grpcEndpointSpec) Addr() string {
 // Metadata method for gRPC endpoint tries to call InfoAPI exposed by Thanos components until context timeout. If we are unable to get metadata after
 // that time, we assume that the host is unhealthy and return error.
 func (es *grpcEndpointSpec) Metadata(ctx context.Context, client *endpointClients) (*endpointMetadata, error) {
-	if client.info != nil {
-		resp, err := client.info.Info(ctx, &infopb.InfoRequest{}, grpc.WaitForReady(true))
-		if err != nil {
-			return nil, errors.Wrapf(err, "fetching info from %s", es.addr)
-		}
+	// TODO(@matej-g): Info client should not be used due to https://github.com/thanos-io/thanos/issues/4699
+	// Uncomment this after it is implemented in https://github.com/thanos-io/thanos/pull/4282.
+	// if client.info != nil {
+	// 	resp, err := client.info.Info(ctx, &infopb.InfoRequest{}, grpc.WaitForReady(true))
+	// 	if err != nil {
+	// 		return nil, errors.Wrapf(err, "fetching info from %s", es.addr)
+	// 	}
 
-		return &endpointMetadata{resp}, nil
-	}
+	// 	return &endpointMetadata{resp}, nil
+	// }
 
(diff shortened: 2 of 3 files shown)


CHANGELOG.md
```diff
@@ -10,6 +10,10 @@ We use *breaking :warning:* to mark changes that are not backward compatible (re
 
 ## Unreleased
 
+### Fixed
```
bwplotka, line 13: 👍🏽

pkg/query/endpointset.go
```diff
@@ -76,17 +77,25 @@ func (es *grpcEndpointSpec) Addr() string {
 // Metadata method for gRPC endpoint tries to call InfoAPI exposed by Thanos components until context timeout. If we are unable to get metadata after
 // that time, we assume that the host is unhealthy and return error.
 func (es *grpcEndpointSpec) Metadata(ctx context.Context, client *endpointClients) (*endpointMetadata, error) {
-	resp, err := client.info.Info(ctx, &infopb.InfoRequest{}, grpc.WaitForReady(true))
-	if err != nil {
-		// Call Info method of StoreAPI, this way querier will be able to discovery old components not exposing InfoAPI.
-		metadata, merr := es.getMetadataUsingStoreAPI(ctx, client.store)
-		if merr != nil {
-			return nil, errors.Wrapf(merr, "fallback fetching info from %s after err: %v", es.addr, err)
+	if client.info != nil {
```
bwplotka, line 80: I would add the same comment here otherwise readers will be surprised. OR just remove the whole `client.info`

bwplotka: (We can get it back from git)

bwplotka (approved):
👍🏽 Thanks! One nit only.

pkg/query/endpointset.go
matej-g: You're right, without having the context it's surprising. I'll add the same comment to keep it consistent.

matej-g:
If anyone could re-run just the failed flow on for CircleCI test, I believe it's a flaky test (see https://github.com/thanos-io/thanos/pull/3798#issuecomment-879845585) and not a real failure. Thanks.

bwplotka (approved):

bwplotka:
Giving up.

sherifkayad:
@matej-g @bwplotka I guess this fix hasn't made it to v0.24, has it? .. I am still seeing the issue there

PR #4715, opened 2021-09-30
Moved http configurations packages around [Refactor]

bwplotka:
Rationatles:
* Query HTTP is not for query, it's literally HTTP generic config used by ruler.


yashrsharma44:
Is there any convention that we would like to follow while naming packages? Because I have seen some of the packages having underscore, while some don't.
But looks like a good change 💯

bwplotka:
Definitely underscore is an anti-pattern: https://golang.org/doc/effective_go#package-names 

> packages are given lower case, single-word names; there should be no need for underscores or mixedCaps

yeya24 (approved):
LGTM!

Issue #4713, opened 2021-09-30
Tools: bucket rewrite tool doesn't work with dry run and multiple block IDs


yeya24:
**Thanos, Prometheus and Golang version used**: main

When using bucket rewrite tool with dry run mode and specifying multiple blocks, it only processes the first block and the process exits.

From the code https://github.com/thanos-io/thanos/blob/main/cmd/thanos/tools_bucket.go#L1137, it returns nil when dry run mode is enabled. It should continue execution in this case.

yeya24:
Just to double check. For each specified block, rewrite tool will rewrite it separately. Is this intended or should it rewrite all specified blocks into a larger one?
cc @bwplotka

yeya24:
Close this for now.

PR #4716, opened 2021-09-30
Fix bucket tools to dry run with multiple block ID's

4molybdenum2:
Signed-off-by: 4molybdenum2 <tathagatapaul7@gmail.com>

Fixed Bucket Tools such that it supports Dry Run with multiple block ID's (#4713)

## Changes

- Changed the code such that when dry run mode is enabled for multiple id's the function doesn't return nil after first iteration of the loop and instead continues to the next iteration.
- Updated logger to log the block ID for each such block in dry run.


yeya24:
Can u remove the unrelated change of bindata?

4molybdenum2:
Yea, I just noticed...I removed them

yeya24 (approved):
Thanks. LGTM

PR #4727, opened 2021-10-02
Add series label for unhealthy index debug log

### LINKED ISSUE #4470: Compactor: Provide more info when unhealthy index found

Compactor halted when out-of-order chunks were found.

The error log is:

```
level=error ts=2021-07-22T06:50:47.485784592Z caller=compact.go:477 msg="critical error detected; halting" 
err="compaction: group 0@8630763776085252337: block with not healthy index found /var/thanos/compact/compact/0@8630763776085252337/01FB5VKXMFT7HZ2R1WKTNNKG00; Compaction level 1; Labels: map[cluster:lhr1 prometheus:gitops-platform-metrics/k8s prometheus_replica:prometheus-k8s-0]: 
1/1041404 series have an average of 1.000 out-of-order chunks: 0.000 of these are exact duplicates (in terms of data and time range)"
```

It only mentions 1 series has out-of-order chunks but doesn't say which series causes the problem. So the only way to fix this is to delete the whole block, which is not ideal.

Provide the series information in the logs so that I can use the bucket rewrite tool to delete the bad series, instead of deleting the whole block.

The only concern here is what if there are too many bad series? Do we need to display them all? But maybe that's okay.

### Additional context

(Write your answer here.)

metonymic-smokey:
Signed-off-by: metonymic-smokey <ahuja.aditi@gmail.com>

Fixes #4470 

## Changes
The debug log on compaction when an unhealthy index is found now mentions the series in the out of order index too. Earlier, only the proportion of such series was logged.

### DIFF
--- a/pkg/block/index.go
+++ b/pkg/block/index.go
@@ -43,8 +43,6 @@ type HealthStats struct {
 	TotalSeries int64
 	// OutOfOrderSeries represents number of series that have out of order chunks.
 	OutOfOrderSeries int
-	// OutOfOrderSeries represents the labels of series that are out of order
-	OutOfOrderSeriesLabels []labels.Labels
 
 	// OutOfOrderChunks represents number of chunks that are out of order (older time range is after younger one).
 	OutOfOrderChunks int
@@ -115,20 +113,12 @@ func (i HealthStats) Issue347OutsideChunksErr() error {
 
 func (i HealthStats) OutOfOrderChunksErr() error {
 	if i.OutOfOrderChunks > 0 {
-		formatedSeriesLabels := ""
-		for _, value := range i.OutOfOrderSeriesLabels {
-			for _, label := range value {
-				formatedSeriesLabels += label.Name + ":" + label.Value + ","
-			}
-			formatedSeriesLabels += "\n"
-		}
 		return errors.New(fmt.Sprintf(
 			"%d/%d series have an average of %.3f out-of-order chunks: "+
-				"The out of order series are: %s.%.3f of these are exact duplicates (in terms of data and time range)",
+				"%.3f of these are exact duplicates (in terms of data and time range)",
 			i.OutOfOrderSeries,
 			i.TotalSeries,
 			float64(i.OutOfOrderChunks)/float64(i.OutOfOrderSeries),
-			formatedSeriesLabels,
 			float64(i.DuplicatedChunks)/float64(i.OutOfOrderChunks),
 		))
 	}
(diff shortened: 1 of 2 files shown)


pkg/block/index.go
```diff
@@ -328,6 +339,7 @@ func GatherIndexHealthStats(logger log.Logger, fn string, minTime, maxTime int64
 		if ooo > 0 {
 			stats.OutOfOrderSeries++
 			stats.OutOfOrderChunks += ooo
+			stats.OutOfOrderSeriesLabels = append(stats.OutOfOrderSeriesLabels, lset)
```
yeya24, line 342: We can just add a debug log here to print out the labels of every out-of-order series.
WDYT? @metonymic-smokey 
This seems simpler and we don't even need to save the slice.

pkg/compact/compact.go
```diff
@@ -773,7 +773,7 @@ func (cg *Group) compact(ctx context.Context, dir string, planner Planner, comp
 		}
 
 		if err := stats.CriticalErr(); err != nil {
-			return false, ulid.ULID{}, halt(errors.Wrapf(err, "block with not healthy index found %s; Compaction level %v; Labels: %v", bdir, meta.Compaction.Level, meta.Thanos.Labels))
+			return false, ulid.ULID{}, halt(errors.Wrapf(err, "block with not healthy index found %s; Compaction level %v; Labels: %v;", bdir, meta.Compaction.Level, meta.Thanos.Labels))
```
yeya24, line 776: Is this needed?

metonymic-smokey: My bad, missed my eye.

pkg/block/index.go
metonymic-smokey: good point @yeya24 
I'll print it out here and remove this field from the struct. 
thanks!

metonymic-smokey:
@yeya24 after these latest reviews, I've retained only the most essential code i.e. printing the labelset when an out of order series occurs.

pkg/block/index.go
```diff
@@ -328,6 +328,13 @@ func GatherIndexHealthStats(logger log.Logger, fn string, minTime, maxTime int64
 		if ooo > 0 {
 			stats.OutOfOrderSeries++
 			stats.OutOfOrderChunks += ooo
+			OutOfOrderSeriesLabels := ""
+			for _, label := range lset[:len(lset)-1] {
+				OutOfOrderSeriesLabels += label.Name + ":" + label.Value + ","
+			}
+			OutOfOrderSeriesLabels += lset[len(lset)-1].Name + ":" + lset[len(lset)-1].Value
+			debugMessage := fmt.Sprintf("The out of order series labels are %s", OutOfOrderSeriesLabels)
+			level.Debug(logger).Log("msg", debugMessage)
```
yeya24, line 337: You can just use 
``` go
level.Debug(logger).Log("msg", "found out of order series", "labels", lset)
```

yeya24 (approved):
LGTM

PR #4739, opened 2021-10-05
Tests: Fix flaky store gateway memcached test

matej-g:

## Changes

The recently introduced test for store gateway with memacached cache seems to be causing a lot of failures (see e.g. https://github.com/thanos-io/thanos/runs/3800344854).

I suspect the metric used for testing - `thanos_store_bucket_cache_operation_hits_total` - is not reflecting the actual operations (not sure why though?). Using `thanos_store_index_cache_hits_total` seems to make the test work as intended.

## Verification

Run tests locally + expecting CI pass

### DIFF
--- a/test/e2e/store_gateway_test.go
+++ b/test/e2e/store_gateway_test.go
@@ -380,7 +380,7 @@ func TestStoreGatewayMemcachedCache(t *testing.T) {
 			},
 		)
 
-		testutil.Ok(t, s1.WaitSumMetrics(e2e.Equals(0), "thanos_store_bucket_cache_operation_hits_total"))
+		testutil.Ok(t, s1.WaitSumMetrics(e2e.Equals(0), "thanos_store_index_cache_hits_total"))
 	})
 
 	t.Run("query with cache hit", func(t *testing.T) {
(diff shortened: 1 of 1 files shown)


onprem:
That store has two caches, the index cache for storing index files and the bucket cache for caching bucket operations. Using `thanos_store_index_cache_hits_total` means we are testing the index cache, but that is not what this test is intended for. We want to test the Bucket cache that we configured with Memcached.

I think I have a hunch about the race condition though. We are scarping the store using a Prometheus instance, and if we test the metric before Prometheus is able to scrape it, the test will fail.

onprem:
Also, index cache is working fine, because we are using in memory (by default) for it.

matej-g:
> Thanos store has two caches, the index cache for storing index files and the bucket cache for caching bucket operations. Using `thanos_store_index_cache_hits_total` means we are testing the index cache, but that is not what this test is intended for. We want to test the Bucket cache that we configured with Memcached.

Ah, I see now, so that is not even the memcached as we're using the default in-memory index cache, it does not make sense, you're right (but yes, it works...)

akanshat:
Yes, got it!!! Working on a fix. Maybe I will manually pass the Memcached config.

PR #4742, opened 2021-10-06
store: batch postings requests to cache

GiedriusS:
Batch postings requests to the cache to improve the efficiency and so that we could
leverage the already existing batching mechanism in the memcached-client layer.

So, instead of a bunch of requests to get one or two keys, there are fewer roundtrips but
with longer GET commands.

If the approach looks good then I could add this to the series requests as well?

In my tests, this greatly reduces the number of `i/o timeout` errors.

## Changes

Changed the IndexCache interface that it would be possible to retrieve multiple `labels.Label` from specified `ulid.ULID`s. Added a `CorkedIndexCache` interface which builds on top - it actually performs the postings fetch only after the counter becomes zero. `DonePostings()` has been added so that the possible `FetchPostings` callers could signal to `CorkedIndexCache` that they won't call `FetchPostings`. It's not harmful to call it more, even if FetchPostings was called.

## Verification

Improved testing suite + existing tests that execute this path. Adhoc tests in pre-production.

### DIFF
--- a/pkg/store/bucket.go
+++ b/pkg/store/bucket.go
@@ -1030,7 +1030,7 @@ func (s *BucketStore) Series(req *storepb.SeriesRequest, srv storepb.Store_Serie
 		}
 	}
 
-	matchingBlocks := []matchingBlock{}
+	matchingBlocks := []*matchingBlock{}
 
 	s.mtx.RLock()
 
@@ -1047,7 +1047,7 @@ func (s *BucketStore) Series(req *storepb.SeriesRequest, srv storepb.Store_Serie
 			debugFoundBlockSetOverview(s.logger, req.MinTime, req.MaxTime, req.MaxResolutionWindow, bs.labels, blocks)
 		}
 
-		matchingBlocks = append(matchingBlocks, matchingBlock{
+		matchingBlocks = append(matchingBlocks, &matchingBlock{
 			matchers: blockMatchers,
 			blocks:   blocks,
 		})
@@ -1058,7 +1058,7 @@ func (s *BucketStore) Series(req *storepb.SeriesRequest, srv storepb.Store_Serie
 	c.CorkAt(corkAt)
 
 	for _, matchingBlock := range matchingBlocks {
-		blockMatchers := matchingBlock.matchers
+		matchingBlock := matchingBlock
 
 		for _, b := range matchingBlock.blocks {
 			b := b
@@ -1094,7 +1094,7 @@ func (s *BucketStore) Series(req *storepb.SeriesRequest, srv storepb.Store_Serie
 					b.extLset,
 					indexr,
 					chunkr,
-					blockMatchers,
+					matchingBlock.matchers,
 					chunksLimiter,
 					seriesLimiter,
 					req.SkipChunks,
--- a/pkg/store/bucket.go
+++ b/pkg/store/bucket.go
@@ -320,9 +320,16 @@ func (b *BucketStore) validate() error {
 
 type noopCache struct{}
 
+func (noopCache) CorkAt(n uint) {}
+func (noopCache) DonePostings() {}
+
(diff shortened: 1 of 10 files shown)


GiedriusS:
cc @pracucci as you've worked on the memcached backend originally and this PR tries to address issues with i/o timeouts while using memcached

akanshat (dismissed):
Looks good to me, just a small comment.

GiedriusS:
I was also thinking that maybe separating FetchMultiPostings() into separate methods will result in a cleaner `CorkedIndexCache` interface as now `FetchMultiPostings()` does the queuing + the actual fetching together (only once, though). So, we would have:

* `QueueFetchMultiPostings()` for queuing what needs to be retrieved;
* `DonePostings()` that will always get called for reducing the internal counter. It's fine to call it more than once;
* `DoFetchMultiPostings()` for actually performing FetchMultiPostings() only once with the queued data

Then, we could remove the extra tracker struct. @akanshat what do you think about this idea?

Antiarchitect:
Any updates on this?

Antiarchitect:
Any update on this?

GiedriusS:
I have rebased this on top of `main`. The tests still pass. @akanshat could you please take a look once you have some time?

GiedriusS:
> @GiedriusS It looks good but are we not proceeding with this change?
> 
> > I was also thinking that maybe separating FetchMultiPostings() into separate methods will result in a cleaner `CorkedIndexCache` interface as now `FetchMultiPostings()` does the queuing + the actual fetching together (only once, though). So, we would have:
> > 
> > * `QueueFetchMultiPostings()` for queuing what needs to be retrieved;
> > * `DonePostings()` that will always get called for reducing the internal counter. It's fine to call it more than once;
> > * `DoFetchMultiPostings()` for actually performing FetchMultiPostings() only once with the queued data

I don't know :shrug: 

The methods here should perform these functions:
* https://github.com/thanos-io/thanos/pull/4742/files#diff-3e2896fafa6ff73509c77df2c4389b68828e02575bb4fb78b6c34bcfb922a7ceR1838 -> calls `DonePostings` if it has not been called yet (https://github.com/thanos-io/thanos/pull/4742/files#diff-7106db7bf336e2dcf943625b5421eb32434c70a12089e46e688fcbd69062cd4eR93-R97)
* `DoFetchMultiPostings()` has been converted into an internal variable: https://github.com/thanos-io/thanos/pull/4742/files#diff-7106db7bf336e2dcf943625b5421eb32434c70a12089e46e688fcbd69062cd4eR103-R106

Maybe both are valid patterns and it's ok to leave the code as-is? Explicit queue+fetch vs. `defer someResource.Close()`. The problem with the 1st option is that we would still have to have something like `defer someResource.Close()` to reduce the counter in the cork in our code. So, maybe let's keep the code as-is i.e. everything happens "magically" behind the interface?

akanshat:
Yep, makes sense!

GiedriusS:
Could you please approve this pull request if it looks good? Perhaps maybe you want to become a Thanos triager/maintainer, at least for the caching layer? If you have some time and will, of course (:

akanshat (approved):

akanshat:
I would love to!

matej-g (commented):
Hey @GiedriusS, I finally had some time to look at this in depth and I quiet like the solution :+1: 

I have few minor suggestions. Once this lands in `main`, I'd be happy to take this to our test environment and report on how number of timeouts changes on our side (something which we still, after numerous tweaks, struggle with from time to time)

GiedriusS:
Thanks for the review @matej-g, I will take a look at this again soon. I wonder if we could rewrite the whole Series() operation to be more streaming. My concern is that with this PR we are blocking further actions on each block before all index operations are done so this could limit us if there's lag on any operation on some block i.e. one laggy index operation blocks further progress on all blocks :disappointed:

PR #4726, opened 2021-10-01
Interactive Example: Do not use monitoring endpoint

matej-g:
Signed-off-by: Matej Gera <matejgera@gmail.com>

* [ ] I added CHANGELOG entry for this change.
* [x] Change is not relevant to the end user.

## Changes

See https://github.com/thanos-io/thanos/issues/4642 - the interactive test cannot run on MacOS due to upstream issue (usage of package available only for Linux platform). This is only due to usage of `e2emonitoring` package, which is however not necessary for running the interactive example. In order to make the example accessible to more users, I'd suggest to remove the monitoring endpoint for now.

## Verification

Waiting on final verification from an Apple user :sweat_smile:

PR #4779, opened 2021-10-14
Examples: Fix the interactive test for MacOS users

matej-g:

## Changes

This PR includes a couple of fixes, targeted towards MacOS users, namely:
1. Bumping to latest version of `efficientgo/e2e` includes a fix which will allow running the interactive test on MacOS
2. Adjusting some `cp -t` usage in the code which was not compatible with MacOS version of `cp`.
3. Made the block generation profile variable, with smaller default, in order to make this run with less RAM.
4. Updated documentation to include more details.

## Verification

Perhaps an Apple user could give the final confirmation? :green_apple:


matej-g:
> I encountered this error (after running the interactive test previously i.e, `interactive/data` dir exists) but not sure if this related to `efficentgo/e2e`,

Oof, I think I know this one :face_with_head_bandage:. Since we're running the Prometheus instance in Docker, but we want to scrape metrics from the [host machine](https://github.com/efficientgo/e2e/blob/b388c346332bfcbfc8a8a81dffae7883802973c8/monitoring/monitoring.go#L154), there needs to be connection from the container to the host. The framework assumes that the host machine will be reachable on the [network's gateway IP](https://github.com/efficientgo/e2e/blob/main/env_docker.go#L116), but this does not seem to be work in all cases. Since the host metrics never get scraped, the operation times out after 2 minutes.

saswatamcode:
Oh! I think this is due to [docker networking](https://docs.docker.com/desktop/mac/networking/#use-cases-and-workarounds) being different on macOS. 

An easy fix seems to be just replacing [this line](https://github.com/efficientgo/e2e/blob/main/env_docker.go#L116) with `d.hostAddr = "gateway.docker.internal"` (would need to detect OS), which I tried locally and it works (thanosbench issue remains though). 🙂

Maybe I can raise PR?

saswatamcode:
For the `thanosbench` issue, I don't think it's related to the platform as `createData()` uses containers for it. 

Seems like [`createData()`](https://github.com/thanos-io/thanos/blob/main/examples/interactive/interactive_test.go#L45) should be generating 9 blocks for each data-dir i.e, store1, store2, prom1, prom2 but it only generates five before erroring out with status code 137. I tested this on both `Darwin x86_64` and `Linux x86_64`(wsl2) and got the same results i.e, [same logs](https://github.com/thanos-io/thanos/pull/4726#issuecomment-943137067) and 5 blocks generated before the error.

Changing block plan profile from [`continuous-1w-small`](https://github.com/thanos-io/thanosbench/blob/master/pkg/blockgen/profiles.go#L83) to [`continuous-30d-tiny`](https://github.com/thanos-io/thanosbench/blob/master/pkg/blockgen/profiles.go#L96) seems to work here i.e, it actually generates 9 blocks and doesn't error out. But I think it only has 5 series per block. 

Maybe this is an issue with thanosbench? 🤕

matej-g:
> Edit: Also the status code 137 means probably means it's getting OOM killed in some way, even though docker inspect shows it wasn't,
>
> Maybe this is an issue with thanosbench or I'm doing something wrong? face_with_head_bandage

Works fine on my machine :tm:. I'm wondering if you're right about the memory, especially if you're running on MacOS or in other virtualized environment with a memory limit (on my Linux machine, I seem to be limited only by my host machine's available memory). What does `docker stats` show you when you run the test? What is the `MEM USAGE / LIMIT` for `thanosbench` container(s)?

saswatamcode:
Yes, I considered that. [Docker for macOS](https://docs.docker.com/desktop/mac/#resources) limits containers to 2GB memory and half the number of host CPUs by default. So I bumped memory to 8GB and max CPUs. But still the same result. `MEM USAGE / LIMIT` is at `7.005GiB / 7.774GiB` by the end.

```
        level=info ts=2021-10-18T10:04:51.148618306Z caller=block.go:94 msg="generating block" spec="[1626667200005 - 1626696000004](7h59m59.999s) "
        level=info ts=2021-10-18T10:04:51.185266496Z caller=head.go:644 msg="Replaying on-disk memory mappable chunks if any"
        level=error ts=2021-10-18T10:04:51.21856407Z caller=head.go:649 msg="Loading on-disk chunks failed" err="iterate on on-disk chunks: out of sequence m-mapped chunk for series ref 17"
        level=info ts=2021-10-18T10:04:51.218680786Z caller=head.go:760 msg="Deleting mmapped chunk files"
        level=info ts=2021-10-18T10:04:51.218699391Z caller=head.go:763 msg="Deletion of mmap chunk files failed, discarding chunk files completely" err="cannot handle error: iterate on on-disk chunks: out of sequence m-mapped chunk for series ref 17"
        level=info ts=2021-10-18T10:04:51.218709409Z caller=head.go:658 msg="On-disk memory mappable chunks replay completed" duration=33.361098ms
        level=info ts=2021-10-18T10:04:51.218716063Z caller=head.go:660 msg="WAL not found"
        level=info ts=2021-10-18T10:05:04.207970966Z caller=writer.go:123 msg=flushing series_count=10000 mint=2021-07-19T04:00:15.005Z maxt=2021-07-19T12:00:00.005Z
        level=info ts=2021-10-18T10:05:07.100724541Z caller=compact.go:494 msg="write block" mint=1626667215005 maxt=1626696000006 ulid=01FJ9DS7AGE6YSRBHXEVHF1RC0 duration=2.892528562s
        level=info ts=2021-10-18T10:05:07.226623846Z caller=block.go:100 msg="generated block" path=/shared/01FJ9DS7AGE6YSRBHXEVHF1RC0 count=5
        level=info ts=2021-10-18T10:05:08.7544727Z caller=block.go:94 msg="generating block" spec="[1626494400006 - 1626667200005](47h59m59.999s) "
        level=info ts=2021-10-18T10:05:08.910671352Z caller=head.go:644 msg="Replaying on-disk memory mappable chunks if any"
        level=error ts=2021-10-18T10:05:08.971713339Z caller=head.go:649 msg="Loading on-disk chunks failed" err="iterate on on-disk chunks: out of sequence m-mapped chunk for series ref 17"
        level=info ts=2021-10-18T10:05:08.972019642Z caller=head.go:760 msg="Deleting mmapped chunk files"
        level=info ts=2021-10-18T10:05:08.972106229Z caller=head.go:763 msg="Deletion of mmap chunk files failed, discarding chunk files completely" err="cannot handle error: iterate on on-disk chunks: out of sequence m-mapped chunk for series ref 17"
        level=info ts=2021-10-18T10:05:08.972974055Z caller=head.go:658 msg="On-disk memory mappable chunks replay completed" duration=62.193655ms
        level=info ts=2021-10-18T10:05:08.973001678Z caller=head.go:660 msg="WAL not found"
        : exit status 137
```

`docker stats` shows memory increasing right after the last log line until finally container is stopped with 137 exit code. It seems to generate the sixth block but is killed before flushing series and writing the block.

@yeya24 had the same logs so maybe he can replicate this?

matej-g:
@yeya24 when you get a moment, may we get a review / merge here if it looks OK to you? Thanks :blush:

PR #4781, opened 2021-10-14
Added PromQL compatibility automatic e2e test suite.

bwplotka:
👋🏽 Adding e2e test so it's easier to run PromQL compliance tests.

It's skipped by default because:

* It's scrape based - data set is build up by scraping Promlab servers
* You need to build your own promql-compliance-tester image locally (https://github.com/prometheus/compliance/pull/46).

If we/https://github.com/prometheus/compliance project will automate above (provide static dataset as we do in https://github.com/thanos-io/thanos/blob/b894fd6537c8b445aa9d907867e492bb26cebcd5/examples/interactive/interactive_test.go#L45) and push image on CI then we might consider running it on every commit.

### DIFF
--- a/test/e2e/compatibility_test.go
+++ b/test/e2e/compatibility_test.go
@@ -71,7 +71,7 @@ scrape_configs:
 
 	t.Run("receive", func(t *testing.T) {
 		testutil.Ok(t, ioutil.WriteFile(filepath.Join(compliance.Dir(), "receive.yaml"),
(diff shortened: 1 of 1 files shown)


bwplotka:
cc @juliusv

GiedriusS:
We should also probably document these optional tests :smile: Finally, did you know that you can just ask for a review from `thanos-io/maintainers` and then everyone will get added automatically?

PR #4786, opened 2021-10-16
Enabling negative offset and @ modifier

### LINKED ISSUE #4022: Negative offset with flag --enable-feature=promql-negative-offset

metonymic-smokey:
Fixes #4022

## Changes
Added enableNegativeOffset and enableAtModifier to engineFactory.
Added unit tests for negative offset and @ modifier.  

## Verification

Run the tests using `go test --run=TestQueryEndpoints`.  The tests passed.

### DIFF
--- a/test/e2e/query_frontend_test.go
+++ b/test/e2e/query_frontend_test.go
@@ -59,7 +59,7 @@ func TestQueryFrontend(t *testing.T) {
 
 	// Ensure we can get the result from Querier first so that it
 	// doesn't need to retry when we send queries to the frontend later.
-	queryAndAssertSeries(t, ctx, q.Endpoint("http"), queryUpWithoutInstance, time.Now(), promclient.QueryOptions{
+	queryAndAssertSeries(t, ctx, q.Endpoint("http"), queryUpWithoutInstance, time.Now, promclient.QueryOptions{
 		Deduplicate: false,
 	}, []model.Metric{
 		{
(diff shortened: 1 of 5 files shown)


metonymic-smokey:
The unit test and testdata files for negative offset is from [this](https://github.com/prometheus/prometheus/blob/main/cmd/promtool/unittest_test.go) file. I have added a new file for it in the interests of brevity. 
To run the test, run `go test query_test.go --run=TestNegativeOffset`

yeya24:
> The unit test and testdata files for negative offset is from [this](https://github.com/prometheus/prometheus/blob/main/cmd/promtool/unittest_test.go) file. I have added a new file for it in the interests of brevity. To run the test, run `go test query_test.go --run=TestNegativeOffset`

I don't think we need to copy this from Prometheus. In the unit test, we just need to add two more test cases to test queries with @ and negative offset can return the expected results, right?

Take a look at https://github.com/thanos-io/thanos/blob/main/pkg/api/query/v1_test.go#L208-L220 and you just need to modify the queries.

PR #4789, opened 2021-10-17
Ignore compaction group with only 1 block

yeya24:

If the compaction group only contains 1 block metadata, then there is nothing to plan and compact. In this case, we can safely ignore the group.


GiedriusS:
Oops, sorry, brain fart. I meant that I thought the planners were taking care of this - they check whether there is any work to perform. So, this won't bring any visible change to users, right? Except for those metrics and not creating/removing dirs unnecessarily

yeya24:
> So, this won't bring any visible change to users, right? Except for those metrics and not creating/removing dirs unnecessarily

Yeah, that's correct. There is no difference except the metrics I would say. Maybe some CPU time can be saved because we omit the planning phase.
But overall it is a tiny thing so this pr is mainly a cleanup.

GiedriusS:
> > So, this won't bring any visible change to users, right? Except for those metrics and not creating/removing dirs unnecessarily
> 
> Yeah, that's correct. There is no difference except the metrics I would say. Maybe some CPU time can be saved because we skip the planning phase. But overall it is a tiny thing so this pr is mainly a cleanup.

:+1: let's remove those old comments in the test and merge this

PR #4790, opened 2021-10-18
Add tombstone cli and mask store series using tombstones

yeya24:

## Changes

This pr continues the previous tombstones related work. It includes two prs:

- https://github.com/thanos-io/thanos/pull/3006
- https://github.com/thanos-io/thanos/pull/3075 

## Verification

TODO:
Unit tests and E2E tests TBD.


yeya24:
> I think we are still missing https://github.com/thanos-io/thanos/pull/3075/files#r477281247 to avoid unnecessary postings lookups? ;P

I am not sure but I guess what I am doing now is what Bartek meant in https://github.com/thanos-io/thanos/pull/3075/files#r477281247?

Now for tombstones matchers, we don't expand postings. Instead we match tombstone matchers with each series labels we get.

yeya24:
I am still thinking about per block tombstones.
For global tombstones, when should we clean up the data in each block and delete the tombstones?

yeya24:
Think about the tombstone approach a little bit more.
Initially, I think it would be super expensive for the compactor to check series for every blocks using global tombstone files. As the compactor doesn't download blocks indexes at start, it has to download all the block index locally to check if the block has series to be rewritten, even if the block doesn't match the series marchers at all. That's why I think per block tombstone makes more sense cuz it is clear what blocks to rewrite.

But per block tombstone has its issues as well like UX, scalability, etc. To overcome the issue from the global tombstone approach, I'm wondering if we can have another gRPC API on store gateway to get the block IDs that matches specific matchers. Then the compactor can leverage this API to get the index information and perform series deletion.

pkg/store/bucket.go
```diff
@@ -1443,6 +1475,13 @@ func (s *BucketStore) LabelValues(ctx context.Context, req *storepb.LabelValuesR
 	}, nil
 }
 
+func (s *BucketStore) SyncTombstones(ctx context.Context) (err error) {
+	s.tombstonesMtx.Lock()
+	s.tombstones, err = tombstone.ReadTombstones(ctx, s.bkt, s.logger)
```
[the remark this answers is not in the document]
yeya24: I guess I misunderstood your suggestion when I first commented on this 🤣 I think I just applied what you suggested. Thanks

pkg/tombstone/tombstone.go
```diff
@@ -0,0 +1,100 @@
...
+
+// Tombstone represents a tombstone.
+type Tombstone struct {
+	Matchers     metadata.Matchers `json:"matchers"`
+	MinTime      int64             `json:"minTime"`
+	MaxTime      int64             `json:"maxTime"`
+	CreationTime int64             `json:"creationTime"`
+	Author       string            `json:"author"`
+	Reason       string            `json:"reason"`
+}
```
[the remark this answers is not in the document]
yeya24: If we use `labels` to represent the external labels (tenants) for each tombstone then it doesn't support matching two tenants. We can either match 1 tenant or all tenants (specify no labels).

cmd/thanos/tools_bucket.go
```diff
@@ -153,6 +155,13 @@ type bucketMarkBlockConfig struct {
 	blockIDs []string
 }
 
+type bucketDeleteConfig struct {
```
yeya24, line 158: I am wondering if we need to rename it because `delete` command might be ambiguous, whether it deletes series or deletes the whole block.
Maybe we can rename it to `delete-series` or `create-tombstone`, as we may need `undelete-series` or `remove-tombstone` in the future...

cmd/thanos/store.go
```diff
@@ -123,6 +124,12 @@ func (sc *storeConfig) registerFlag(cmd extkingpin.FlagClause) {
 	cmd.Flag("block-meta-fetch-concurrency", "Number of goroutines to use when fetching block metadata from object storage.").
 		Default("32").IntVar(&sc.blockMetaFetchConcurrency)
 
+	cmd.Flag("sync-tombstone-duration", "Repeat interval for syncing the tombstones between local and remote view.").
```
yeya24, line 127: Wondering if we need another flag to enable/disable tombstone masking.

pkg/store/bucket.go
```diff
@@ -824,11 +834,36 @@ func blockSeries(
...
+					if !matcher.Matches(val) {
+						continue
+					}
+					if skipChunks {
+						continue PostingsLoop
+					}
+					tombstoneIntervals = tombstoneIntervals.Add(promtombstones.Interval{Mint: ts.MinTime, Maxt: ts.MaxTime})
+				}
+			}
+		}
```
yeya24, line 853: I am wondering which way is better. Now we are matching tombstones at the query time, which is online processing. Another way is to build an in-memory Prometheus tombstone for each matched block when we load tombstones each time. Since block contents are immutable, we can cache the in-mem tombstone results each time based on the tombstone ids.
Ideally, the second option is better in performance. But happy to hear more ideas.

PR #4795, opened 2021-10-21
query: fix deadlock in endpointset

### LINKED ISSUE #4766: query: Query API is unresponsive but health/ready APIs report healthy/ready
**Thanos, Prometheus and Golang version used**: Thanos v0.23.0-rc.1

**Object Storage Provider**: Azure, accessed via a managed identity.

**What happened**: I have deployed 2 Thanos query nodes, Prometheus, Thanos query frontend, Ruler, Store, Grafana in multiple Azure k8s service clusters using thanos, kube-prometheus-stack and grafana helm charts. The Query nodes are configured in a hierarchical fashion, such that one query node collects metrics solely from the cluster and the other queries the query node in every deployed cluster. Query Frontend sits in front of the top level query node.

The query API of the top level query node is completely unresponsive and does not respond or timeout API calls. I've tried using the different API endpoints e.g. query, labels but they are all unresponsive. However, both the healthy and ready APIs are reporting success and I'm not seeing any significant logs in the query node other than some SRV records failing to be resolved (which is expected as those clusters are currently down).

The query API in the low level query node is also fine and I can query metrics from the cluster, I just am unable to query metrics covering multiple clusters.

**What you expected to happen**: Query API to respond to queries when pod is reported healthy or timeout.

**How to reproduce it (as minimally and precisely as possible)**: I don't have exact reproduction steps. I have deployed multiple clusters and in 6, the top level query API has come up successfully while in 2, query API was unresponsive. Restarting the Query pod has resolved the issue in one cluster but since the healthy API is reporting success, this is not happening automatically.

GiedriusS:
Avoid RLock()ing twice as described here:
https://github.com/thanos-io/thanos/issues/4766#issuecomment-948743455
(due to
https://stackoverflow.com/questions/30547916/goroutine-blocks-when-calling-rwmutex-rlock-twice-after-an-rwmutex-unlock/30549188).
Fix it by removing HasClients() and simply changing it with `er.clients != nil`.

Closes #4766.

### DIFF
--- a/pkg/query/endpointset.go
+++ b/pkg/query/endpointset.go
@@ -687,46 +687,39 @@ func (er *endpointRef) ComponentType() component.Component {
 	return component.FromString(er.metadata.ComponentType)
 }
 
-func (er *endpointRef) HasClients() bool {
-	er.mtx.RLock()
-	defer er.mtx.RUnlock()
-
-	return er.clients != nil
-}
-
 func (er *endpointRef) HasStoreAPI() bool {
 	er.mtx.RLock()
 	defer er.mtx.RUnlock()
 
-	return er.HasClients() && er.clients.store != nil
+	return er.clients != nil && er.clients.store != nil
 }
 
 func (er *endpointRef) HasRulesAPI() bool {
 	er.mtx.RLock()
 	defer er.mtx.RUnlock()
 
-	return er.HasClients() && er.clients.rule != nil
+	return er.clients != nil && er.clients.rule != nil
 }
 
 func (er *endpointRef) HasTargetsAPI() bool {
 	er.mtx.RLock()
 	defer er.mtx.RUnlock()
 
-	return er.HasClients() && er.clients.target != nil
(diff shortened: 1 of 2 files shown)


hanjm:
Hi @GiedriusS @squat ,  I found at least 4 people report this bug, it critical because query will halt . Could we released a miro version like 0.23.2, just cherry pick this import pr. Thanks very much.

squat:
This PR is in the release candidate for 0.24.0. i think that's enough. Or do you think it's so severe that we need to back port the fix?

hanjm:
@squat  I think it's severe to do that.  Someone use 0.23.x always face this deadlock issue, the query api will halt and process will not auto reboot, It will be nice to have 0.23.2 for 0.23.x series. 😄 

I am also face this issue on production upgrade, other case like  https://cloud-native.slack.com/archives/CK5RSSC10/p1637585247351700 https://cloud-native.slack.com/archives/CK5RSSC10/p1637265253335900 .

PR #4801, opened 2021-10-26
Metrics for compaction and downsampling process

### LINKED ISSUE #3985: Have easier to way to tell the progress of compactor (and downsampling)
Currently, you need to check Thanos Compact UI and check if all older blocks are bigger. There should be only up to 5 of 2h, 8h and 2d blocks. Rest should be compacted to 2w. Similar for downsampled blocks. 

Or you can check the number of compactions per day to see if the number stabilizes. 

Both ways are pretty manual. I would propose adding metric suggesting the backlog of compaction to make. This requires potentially changing our compaction planner logic, which is already planned for https://github.com/thanos-io/thanos/issues/3405

metonymic-smokey:

## Changes
This PR adds metrics to tracks the progress of compaction and downsampling such as number of iterations, blocks, etc.
Fixes #3985 

## Verification
Metrics shown on the `/metrics` endpoint: 
Compaction:
image

Downsampling: 
image


pkg/compact/compact.go
```diff
@@ -470,6 +483,177 @@ func (cg *Group) Resolution() int64 {
...
+
+	for key, iters := range groupCompactions {
+		ps.ProgressMetrics.NumberOfCompactionRuns.WithLabelValues(key).Add(float64(iters))
+		ps.ProgressMetrics.NumberOfCompactionBlocks.WithLabelValues(key).Add(float64(groupBlocks[key]))
+	}
+
+	return nil
+}
+
+type DownsampleMetrics struct {
```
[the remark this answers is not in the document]
yeya24: 👍 Make sene.

pkg/compact/compact_test.go
```diff
@@ -174,3 +177,240 @@ func BenchmarkGatherNoCompactionMarkFilter_Filter(b *testing.B) {
 	}
 
 }
+
+func TestPlanSimulate(t *testing.T) {
+	logger := log.NewNopLogger()
```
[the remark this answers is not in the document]
yeya24: I thought the logger was used by our simulator struct before. If it is also needed by the tsdb planner then all good.

pkg/compact/compact.go
```diff
@@ -470,6 +483,177 @@ func (cg *Group) Resolution() int64 {
 	return cg.resolution
 }
 
+type ProgressMetrics struct {
+	NumberOfCompactionRuns   *prometheus.GaugeVec
+	NumberOfCompactionBlocks *prometheus.GaugeVec
+}
+
+type ProgressCalculator interface {
```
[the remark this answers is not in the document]
yeya24: Let's only focus on the interface itself. `ProgressCalculator calculates progress from given metadata groups`. Something like this is already good enough.

`ProgressCalculator updates the results/metrics from the compaction planning and downsampling simulations.`

Here you are talking about:
1. update `results/metrics` from simulations. This is the implementation details and shouldn't be part of the interface comment. We can even have a struct which implements this interface but doesn't update any metrics

2. `from the compaction planning and downsampling simulations`. Same as the last one. This is the detailed interface implementer information. Let's remove them here as well.

pkg/compact/compact.go
```diff
@@ -470,6 +483,177 @@ func (cg *Group) Resolution() int64 {
...
+	ProgressCalculate(ctx context.Context, groups []*Group) error
+}
+
+type CompactionSimulator struct {
```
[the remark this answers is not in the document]
yeya24: Let's be more consistent on the naming. If the interface is called `ProgressCalculator` then let's call it also `CompactionProgressCalculator` instead of `xxxSimulator` here.

pkg/compact/compact_test.go
```diff
@@ -174,3 +177,241 @@ func BenchmarkGatherNoCompactionMarkFilter_Filter(b *testing.B) {
...
+	planner := NewTSDBBasedPlanner(log.NewNopLogger(), []int64{
+		int64(1 * time.Hour / time.Millisecond),
+		int64(2 * time.Hour / time.Millisecond),
+		int64(8 * time.Hour / time.Millisecond),
+		int64(2 * 24 * time.Hour / time.Millisecond),
+	})
+	reg := prometheus.NewRegistry()
+	ps := NewCompactionSimulator(reg, planner)
+
+	metas := []*metadata.Meta{
```
yeya24, line 191: Can you add more test cases and make it a table like you do in the downsample tests?

pkg/compact/compact.go
```diff
@@ -470,6 +483,182 @@ func (cg *Group) Resolution() int64 {
...
+		ps.ProgressMetrics.NumberOfCompactionRuns.WithLabelValues(key).Add(float64(iters))
+		ps.ProgressMetrics.NumberOfCompactionBlocks.WithLabelValues(key).Add(float64(groupBlocks[key]))
+	}
+
+	return nil
+}
+
+// DownsampleMetrics contains Prometheus metrics related to downsampling progress.
+type DownsampleMetrics struct {
+	BlocksDownsampled *prometheus.GaugeVec
```
yeya24, line 572: Let's call it `NumberOfDownsampleBlocks` here? As we already have`NumberOfCompactionBlocks` in the compaction simulator metrics

pkg/compact/compact.go
```diff
@@ -470,6 +483,177 @@ func (cg *Group) Resolution() int64 {
...
+				return errors.Wrapf(err, "append meta")
+			}
+			tmpGroups = append(tmpGroups, g)
+		}
+
+		groups = tmpGroups
+	}
+
+	for key, iters := range groupCompactions {
+		ps.ProgressMetrics.NumberOfCompactionRuns.WithLabelValues(key).Add(float64(iters))
```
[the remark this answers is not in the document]
metonymic-smokey: @yeya24 thanks, this is a better alternative!
I've pushed this in a recent commit. Could you let me know if that is right?
Thanks!

pkg/compact/compact.go
metonymic-smokey: Yes, that makes it more consistent. Thanks!

metonymic-smokey: Thanks for the detailed review of the comments! Certainly learnt a lot about concise comments without going into implementation details. I've fixed the names and comments in a recent commit.

pkg/compact/compact_test.go
metonymic-smokey: Yes, I'll organise and add more test cases for both the unit tests once everything else is done. Since I know how to do table testing now, it shouldn't take long.

pkg/compact/compact.go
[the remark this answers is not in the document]
yeya24: You can just use the unregisterer to initialize your simulator structs. You don't need to explicitly call `MustRegister` method.

Something like this:

```go
unRegisterer := &receive.Unregisterer{reg: reg}
ps := compact.NewCompactionProgressCalculator(unRegisterer, tsdbPlanner)
```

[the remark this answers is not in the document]
metonymic-smokey: Thanks! I've fixed this in a recent commit.

pkg/compact/compact_test.go
metonymic-smokey: I've added some more cases and organised TestPlanSimulate using table tests. I'll continue adding more cases. PTAL and let me know if that is okay.
Thanks!

pkg/compact/compact_test.go
```diff
@@ -174,3 +178,358 @@ func BenchmarkGatherNoCompactionMarkFilter_Filter(b *testing.B) {
...
+
+func TestPlanSimulate(t *testing.T) {
+	reg := prometheus.NewRegistry()
+	unRegisterer := &receive.UnRegisterer{Registerer: reg}
+
+	extLabels := labels.FromMap(map[string]string{"a": "1"})
+
+	for _, tcase := range []struct {
+		testName    string
+		testPlanner *tsdbBasedPlanner
```
yeya24, line 194: I see we are just using the same TSDB planner and creating the planner every time. If you just plan to use the same tsdb planner you can just remove it from the testcase and create a single TSDB planner at the beginning.

yeya24: I think it would be good if we can also create a mock planner for testing

pkg/compact/compact_test.go
```diff
@@ -174,3 +178,358 @@ func BenchmarkGatherNoCompactionMarkFilter_Filter(b *testing.B) {
...
+						Downsample: metadata.ThanosDownsample{
+							Resolution: downsample.ResLevel0,
+						},
+					},
+				},
+			},
+			expected: 1.0,
+		},
+		{
+			// this test case returns 0 blocks to be downsampled since the resolution is resLevel2, which is skipped when grouping blocks for downsampling.
```
yeya24, line 446: ditto

pkg/compact/compact_test.go
```diff
@@ -174,3 +178,358 @@ func BenchmarkGatherNoCompactionMarkFilter_Filter(b *testing.B) {
...
+	reg := prometheus.NewRegistry()
+	unRegisterer := &receive.UnRegisterer{Registerer: reg}
+
+	for _, tcase := range []struct {
+		testName string
+		input    []*metadata.Meta
+		expected float64
+	}{
+		{
+			// this test case has 1 block to be downsampled out of 2 since for the second block, the difference between MinTime and MaxTime is less than the acceptable threshold, DownsampleRange0
```
yeya24, line 405: 👍 Can you capitalize the sentence?

pkg/compact/compact_test.go
metonymic-smokey: Thanks! Re-used a single planner in the test cases now.

cmd/thanos/compact.go
```diff
@@ -458,6 +460,56 @@ func runCompact(
...
+					}
+				}
+
+				for _, group := range groups {
+					ps.CompactProgressMetrics.NumberOfCompactionRuns.WithLabelValues(group.Key())
+					ps.CompactProgressMetrics.NumberOfCompactionBlocks.WithLabelValues(group.Key())
+				}
+
+				if err = ps.ProgressCalculate(ctx, groups); err != nil {
+					return errors.Wrapf(err, "could not simulate planning")
```
yeya24, line 503: Can you update the comment to be more specific? `could not calculate compaction progress`

cmd/thanos/compact.go
```diff
@@ -458,6 +460,56 @@ func runCompact(
...
+				if !conf.disableDownsampling {
+					downGroups, err := grouper.Groups(originalMetas)
+					if err != nil {
+						return errors.Wrapf(err, "could not group original metadata into downsample groups")
+					}
+					for _, group := range downGroups {
+						ds.DownsampleProgressMetrics.NumberOfBlocksDownsampled.WithLabelValues(group.Key())
+					}
+					if err := ds.ProgressCalculate(ctx, downGroups); err != nil {
+						return errors.Wrapf(err, "could not simulate downsampling")
```
yeya24, line 493: Let's not call it as `simulate`

cmd/thanos/compact.go
```diff
@@ -458,6 +460,56 @@ func runCompact(
...
+				}
+
+				originalMetas := sy.Metas()
+				groups, err := grouper.Groups(originalMetas)
+				if err != nil {
+					return errors.Wrapf(err, "could not group original metadata")
+				}
+
+				if !conf.disableDownsampling {
+					downGroups, err := grouper.Groups(originalMetas)
```
yeya24, line 485: Why not just call it `groups`?

cmd/thanos/compact.go
```diff
@@ -458,6 +460,56 @@ func runCompact(
...
+			return runutil.Repeat(5*time.Minute, ctx.Done(), func() error {
+
+				if err := sy.SyncMetas(ctx); err != nil {
+					return errors.Wrapf(err, "could not sync metas")
+				}
+
+				originalMetas := sy.Metas()
+				groups, err := grouper.Groups(originalMetas)
+				if err != nil {
+					return errors.Wrapf(err, "could not group original metadata")
```
yeya24, line 481: We don't need to call it `orgininal`

cmd/thanos/compact.go
```diff
@@ -458,6 +460,56 @@ func runCompact(
...
+				}
+
+				if !conf.disableDownsampling {
```
yeya24, line 484: Can you put this code block after compaction progress is done?

pkg/compact/compact_test.go
```diff
@@ -174,3 +178,360 @@ func BenchmarkGatherNoCompactionMarkFilter_Filter(b *testing.B) {
...
+		testName          string
+		testPlannerRanges []int64
+		input             []*metadata.Meta
+		expected          planResult
+	}{
+		// In this test case, the first four blocks are planned for compaction in the first run. These are then removed from the group and then the next two blocks from the original group are planned for compaction in the second run.
+		// Hence, a total of 6 blocks are planned for compaction over 2 runs.
+		{
+			testName: "two_runs",
+			testPlannerRanges: []int64{
```
yeya24, line 203: You don't have to make ranges a variable. Just passing them to the planner at line 189 is fine.

pkg/compact/compact_test.go
yeya24: Can you also add a mock planner then?

pkg/compact/compact_test.go
```diff
@@ -174,3 +178,360 @@ func BenchmarkGatherNoCompactionMarkFilter_Filter(b *testing.B) {
 	}
 
 }
+
+type planResult struct {
```
yeya24, line 182: 👍

yeya24, line 182: Can we make this struct as an inner struct in the `testPlannerRanges` test case?

pkg/compact/compact_test.go
metonymic-smokey: The two ranges have different values. How can I pass them to planner in the beginning?

yeya24: Ok I see. But why do you need to use two different ranges here? Using one range up to 2d is the same for the two, right?

metonymic-smokey: Alright, addressed this in a recent commit.

pkg/compact/compact_test.go
metonymic-smokey: What exactly is a mock planner? I'm not fully clear.
Thanks!

pkg/compact/compact_test.go
yeya24: This is not fixed

pkg/compact/compact_test.go
```diff
@@ -174,3 +178,483 @@ func BenchmarkGatherNoCompactionMarkFilter_Filter(b *testing.B) {
...
+	extLabels := labels.FromMap(map[string]string{"a": "1", "b": "2"})
+
+	for _, tcase := range []struct {
+		testName string
+		input    []*metadata.Meta
+		expected float64
+	}{
+		{
+			// This test case has 1 block to be downsampled out of 2 since for the second block, the difference between MinTime and MaxTime is less than the acceptable threshold, DownsampleRange0
+			testName: "min_max_time_diff_test",
```
yeya24, line 489: So test name is confusing. Can you please rename it?

pkg/compact/compact_test.go
```diff
@@ -174,3 +178,483 @@ func BenchmarkGatherNoCompactionMarkFilter_Filter(b *testing.B) {
...
+					Thanos: metadata.Thanos{
+						Version: 1,
+						Labels:  map[string]string{"b": "2"},
+					},
+				},
+			},
+			expected: planResult{
+				compactionRuns:   1.0,
+				compactionBlocks: 4.0,
+			},
```
yeya24, line 453: Can you also check the result metrics for each group as well?

pkg/compact/compact.go
```diff
@@ -470,6 +483,186 @@ func (cg *Group) Resolution() int64 {
...
+	return &CompactionProgressCalculator{
+		planner: planner,
+		CompactProgressMetrics: &CompactProgressMetrics{
+			NumberOfCompactionRuns: promauto.With(reg).NewGaugeVec(prometheus.GaugeOpts{
+				Name: "thanos_compact_todo_compactions",
+				Help: "number of iterations to be done",
+			}, []string{"group"}),
+			NumberOfCompactionBlocks: promauto.With(reg).NewGaugeVec(prometheus.GaugeOpts{
+				Name: "thanos_compact_todo_compaction_blocks",
+				Help: "number of blocks planned to be merged",
```
yeya24, line 514: `planned to be compacted` would be better

pkg/compact/compact.go
```diff
@@ -470,6 +483,186 @@ func (cg *Group) Resolution() int64 {
 	return cg.resolution
 }
 
+// CompactProgressMetrics contains Prometheus metrics related to planning and compaction progress.
```
yeya24, line 486: `related to compaction progress` is better. No need to mention planning here

cmd/thanos/compact.go
```diff
@@ -590,9 +642,12 @@ type compactConfig struct {
 	enableVerticalCompaction                       bool
 	dedupFunc                                      string
 	skipBlockWithOutOfOrderChunks                  bool
+	compactionProgressMetrics                      bool
 }
 
 func (cc *compactConfig) registerFlag(cmd extkingpin.FlagClause) {
+	cmd.Flag("progress-metrics", "Enables the progress metrics, indicating the progress of the compaction and downsampling processes").Default("false").BoolVar(&cc.compactionProgressMetrics)
```
yeya24, line 649: we can remove the trailing `processes`

yeya24 (commented):
LGTM with only small nits. I think we are ready to merge this after fixing the nits.
Great work @metonymic-smokey!

pkg/compact/compact_test.go
metonymic-smokey: I'm doing that specifically for the two_groups_test in my latest commit.

cmd/thanos/compact.go
```diff
@@ -590,9 +642,12 @@ type compactConfig struct {
 	enableVerticalCompaction                       bool
 	dedupFunc                                      string
 	skipBlockWithOutOfOrderChunks                  bool
+	compactionProgressMetrics                      bool
 }
 
 func (cc *compactConfig) registerFlag(cmd extkingpin.FlagClause) {
+	cmd.Flag("progress-metrics", "Enables the progress metrics, indicating the progress of compaction and downsampling").Default("false").BoolVar(&cc.compactionProgressMetrics)
```
GiedriusS, line 649: I mean why even have a flag for this? We get so many questions about how to tell whether Thanos Compact is progressing or not so I think this would be useful for everyone

cmd/thanos/compact.go
GiedriusS: And we can reduce indentation by:
```go
if conf.disableDownsampling {
  return nil
  }  
```

cmd/thanos/compact.go
yeya24: What about still having a flag but default to true? I want users to be able to disable it as some clusters setup doesn't need to check compact backlog

cmd/thanos/compact.go
yeya24: I mean in the log unnecessary to call it original. "Could not group metadata" is good enough

pkg/compact/compact_test.go
```diff
@@ -174,3 +179,705 @@ func BenchmarkGatherNoCompactionMarkFilter_Filter(b *testing.B) {
...
+						},
+					},
+				},
+			},
+			expected: 2.0,
+		},
+	} {
+		groups := []*Group{
+			{
+				key:            "a",
```
yeya24, line 812: Why hardcore the group key here? This way you are still testing only one group

yeya24 (changes requested):

pkg/compact/compact_test.go
```diff
@@ -174,3 +179,705 @@ func BenchmarkGatherNoCompactionMarkFilter_Filter(b *testing.B) {
...
+			expected: planResult{
+				compactionRuns:   1.0,
+				compactionBlocks: 4.0,
+			},
+		},
+	} {
+		if ok := t.Run(tcase.testName, func(t *testing.T) {
+			groups := []*Group{
+				{
+					key:            "b",
```
yeya24, line 538: Please not hard code the group key.
You need to use grouper to get groups from the input metadata

cmd/thanos/compact.go
GiedriusS: Sounds good 👍

pkg/compact/compact_test.go
metonymic-smokey: Understood, I've tried to fix this in a recent commit. PTAL and let me know if that is okay.
Thanks!

pkg/compact/compact_test.go
metonymic-smokey: Attempted to fix in a recent commit. Thanks.

pkg/compact/compact_test.go
```diff
@@ -522,29 +521,44 @@ func TestCompactProgressCalculate(t *testing.T) {
...
 	for _, tcase := range []struct {
 		testName string
 		input    []*metadata.Meta
-		expected float64
+		expected groupedResult
 	}{
 		{
-			// This test case has 1 block to be downsampled out of 2 since for the second block, the difference between MinTime and MaxTime is less than the acceptable threshold, DownsampleRange0
-			testName: "minTime_maxTime_diff_threshold_test",
+			testName: "first_test",
```
yeya24, line 556: The test case name doesn't mean anything. Can you update all the names?

pkg/compact/compact_test.go
```diff
@@ -687,78 +763,33 @@ func TestDownsampleProgressCalculate(t *testing.T) {
...
-						},
-					},
-				},
-			})
-		}
+			blocks := make(map[ulid.ULID]*metadata.Meta, len(tcase.input))
+			for _, meta := range tcase.input {
+				blocks[meta.ULID] = meta
+			}
+			groups, _ = grouper.Groups(blocks)
```
yeya24, line 781: Check error here

pkg/compact/compact_test.go
```diff
@@ -687,78 +763,33 @@ func TestDownsampleProgressCalculate(t *testing.T) {
...
-		groups := []*Group{
-			{
-				key:            "a",
-				labels:         extLabels,
-				resolution:     downsample.ResLevel1,
-				metasByMinTime: tcase.input,
-			},
-		}
+		if ok := t.Run(tcase.testName, func(t *testing.T) {
+			groups := make([]*Group, 3)
```
yeya24, line 775: This is not necessary

pkg/compact/compact_test.go
metonymic-smokey: This is already fixed in a previous commit.

pkg/compact/compact_test.go
metonymic-smokey: thanks, fixed it for both tests.

pkg/compact/compact_test.go
metonymic-smokey: Thanks, fixed it for both tests.

pkg/compact/compact_test.go
```diff
@@ -174,3 +179,622 @@ func BenchmarkGatherNoCompactionMarkFilter_Filter(b *testing.B) {
...
+	} {
+		if ok := t.Run(tcase.testName, func(t *testing.T) {
+			blocks := make(map[ulid.ULID]*metadata.Meta, len(tcase.input))
+			for _, meta := range tcase.input {
+				blocks[meta.ULID] = meta
+			}
+			groups, err := grouper.Groups(blocks)
+			if err != nil {
+				level.Warn(logger).Log("msg", "unable to form groups")
+			}
```
yeya24, line 783: No need to log error. Please use `testutil.Ok` here

pkg/compact/compact_test.go
yeya24: I don't think these are fixed.

pkg/compact/compact_test.go
```diff
@@ -174,3 +179,622 @@ func BenchmarkGatherNoCompactionMarkFilter_Filter(b *testing.B) {
...
+			// form groups from the input metadata - do not hardcode groups. hence, grouper.Groups should stay
+			groups, err := grouper.Groups(blocks)
+			if err != nil {
+				level.Warn(logger).Log("msg, unable to form groups")
+			}
+			ps := NewCompactionProgressCalculator(unRegisterer, planner)
+			err = ps.ProgressCalculate(context.Background(), groups)
+			metrics := ps.CompactProgressMetrics
+			testutil.Ok(t, err)
+			for _, key := range keys {
```
yeya24, line 509: As we discussed on Slack, it is unnecessary to iterate over all keys and you don't need all 3 groups in all test cases. Use expected map to check the results.

pkg/compact/compact_test.go
```diff
@@ -174,3 +179,622 @@ func BenchmarkGatherNoCompactionMarkFilter_Filter(b *testing.B) {
...
+		if ok := t.Run(tcase.testName, func(t *testing.T) {
+			blocks := make(map[ulid.ULID]*metadata.Meta, len(tcase.input))
+			for _, meta := range tcase.input {
+				blocks[meta.ULID] = meta
+			}
+			// form groups from the input metadata - do not hardcode groups. hence, grouper.Groups should stay
+			groups, err := grouper.Groups(blocks)
+			if err != nil {
+				level.Warn(logger).Log("msg, unable to form groups")
+			}
```
yeya24, line 504: No need to use logger.

metonymic-smokey: thanks, fixed this for both tests now.

pkg/compact/compact_test.go
metonymic-smokey: Thanks, I'm using `testutil.Ok` now for error handling.

pkg/compact/compact_test.go
```diff
@@ -174,3 +178,275 @@ func BenchmarkGatherNoCompactionMarkFilter_Filter(b *testing.B) {
...
+	m[1].Thanos.Labels = map[string]string{"b": "2"}
+	m[1].Thanos.Downsample.Resolution = downsample.ResLevel1
+	m[2].Thanos.Labels = map[string]string{"a": "1", "b": "2"}
+	m[2].Thanos.Downsample.Resolution = downsample.ResLevel2
+	for ind, meta := range m {
+		keys[ind] = DefaultGroupKey(meta.Thanos)
+	}
+
+	var bkt objstore.Bucket
+	temp := promauto.NewCounter(prometheus.CounterOpts{Name: "test_metric_for_group", Help: "this is a test metric for downsample progress tests"})
```
yeya24, line 368: https://app.circleci.com/pipelines/github/thanos-io/thanos/7943/workflows/3f1a0011-dd6c-473e-acc0-bffafbd09f3e/jobs/17125.

When you use `promauto` library to register the metrics, please add `promauto.With(reg).NewCounter(...)` to specify the registry. Otherwise it will use the default global registry.

pkg/compact/compact_test.go
metonymic-smokey: Thanks, fixed this for downsampling in a recent commit. Will fix for compact progress next.

pkg/compact/compact_test.go
metonymic-smokey: I have done this in a recent commit but the CircleCI test is still failing.

pkg/compact/compact.go
```diff
@@ -470,6 +483,186 @@ func (cg *Group) Resolution() int64 {
...
+}
+
+// NewCompactProgressCalculator creates a new CompactionProgressCalculator.
+func NewCompactionProgressCalculator(reg prometheus.Registerer, planner *tsdbBasedPlanner) *CompactionProgressCalculator {
+	return &CompactionProgressCalculator{
+		planner: planner,
+		CompactProgressMetrics: &CompactProgressMetrics{
+			NumberOfCompactionRuns: promauto.With(reg).NewGaugeVec(prometheus.GaugeOpts{
+				Name: "thanos_compact_todo_compactions",
+				Help: "number of iterations to be done",
```
yeya24, line 510: Can you update the message to `number of compactions to be done`?

cmd/thanos/compact.go
yeya24: This is still not solved @metonymic-smokey

cmd/thanos/compact.go
```diff
@@ -458,6 +459,51 @@ func runCompact(
...
+				}
+
+				metas := sy.Metas()
+				groups, err := grouper.Groups(metas)
+				if err != nil {
+					return errors.Wrapf(err, "could not group original metadata")
+				}
+
+				ps.CompactProgressMetrics.NumberOfCompactionRuns.Reset()
+				ps.CompactProgressMetrics.NumberOfCompactionBlocks.Reset()
```
yeya24, line 483: Can you put the reset method inside `ProgressCalculate`? IMO it is best to put it above https://github.com/thanos-io/thanos/pull/4801/files#diff-be21b0381ac14ee139355d057b390aa69ff338fe1c748ae8a475ce46f3583575R564.

cmd/thanos/compact.go
```diff
@@ -458,6 +459,51 @@ func runCompact(
...
+				if err = ps.ProgressCalculate(ctx, groups); err != nil {
+					return errors.Wrapf(err, "could not calculate compaction progress")
+				}
+
+				if !conf.disableDownsampling {
+					groups, err = grouper.Groups(metas)
+					if err != nil {
+						return errors.Wrapf(err, "could not group original metadata into downsample groups")
+					}
+					ds.DownsampleProgressMetrics.NumberOfBlocksDownsampled.Reset()
```
yeya24, line 494: ditto. This should be the implementation detail of the `ProgresssCalculate` method and let's put reset inside that method.

pkg/compact/compact_test.go
metonymic-smokey: Fixed for both tests now.

yeya24 (approved):
LGTM. Thanks for the patience. Let's try it!

PR #4808, opened 2021-10-28
Fix store filter select menu color in dark mode

saswatamcode:

## Changes

This PR fixes the Store Filter select menu text color to be more legible in dark mode.

Before fix:

Screenshot 2021-10-28 at 7 41 57 PM

After fix:

Screenshot 2021-10-28 at 7 53 13 PM

Note: Had to type annotate `catch` blocks due to errors when running `make assets`, but is unrelated to PR.
## Verification

Tested locally.


yeya24 (approved):
LGTM

yeya24:
I will merge this pr first to unblock other UI prs. If there are things to improve in this one, then let's fix them later. 😄 
cc @onprem

PR #4816, opened 2021-10-29
storepb: expose step/range on storepb

GiedriusS:
Expose step/range on storepb so that Sidecar could use this data to
only retrieve data which is actually needed: instead of sending
everything from min time to max time, let's iterate over using the step
and only send the `range` time in the future from any given step.

Remote read supports multiple queries in one request so we could
leverage that for this feature
(https://github.com/prometheus/prometheus/blob/main/prompb/remote.proto#L32).
Prometheus remote read sends XOR encoded chunks so there shouldn't be
more load on Prometheus server, and we should be able to successfully
reduce the necessary bandwidth between Sidecar and Query.


yeya24:
> Remote read supports multiple queries in one request so we could
leverage that for this feature
(https://github.com/prometheus/prometheus/blob/main/prompb/remote.proto#L32)

For the multiple queries here, how would Prometheus return data in one response? Is it actually doing a `union` for response serieses of all queries?

yeya24 (approved):

GiedriusS:
@yeya24 I was thinking about https://github.com/GiedriusS/thanos/commit/a186b5be390dfe2425aed71975cab2323533e2d3 - that commit actually works, just need to do a small benchmark whether it actually reduces the amount of network traffic

yeya24:
I am good with the change but let's get more reviews from other maintainers about the proto change.
It would be better to include GiedriusS@a186b5b this commit to this pr and make sure this way has improvement. WDYT?

GiedriusS:
> I am good with the change but let's get more reviews from other maintainers about the proto change. It would be better to include [GiedriusS/thanos@a186b5b](https://github.com/GiedriusS/thanos/commit/a186b5b) this commit to this pr and make sure this way has improvement. WDYT?

I want to start small by including these two extra hints in `SeriesRequest`. I think it will be cleaner to use those hints downstream i.e. in Sidecar to generate those time ranges instead of putting everything in the request. I think we are the most active maintainers :smile: so, of course, if anyone wants to chime in then feel free. Otherwise, I will come back with a simple bench to see whether it helps. And if it does then I'll merge this and work on the 2nd part - using this data in Sidecar

yeya24 (approved):
Let's try it!

PR #4821, opened 2021-11-01
Make hyperlinks work

clyang82:

### DIFF
--- a/.mdox.yaml
+++ b/.mdox.yaml
@@ -14,6 +14,8 @@ localLinksStyle:
   hugo:
     indexFileName: "_index.md"
 
+linkPrefixForNonMarkdownResources: "https://github.com/thanos-io/thanos/blob/main"
+
 transformations:
 
   - glob: "$(EXTERNAL_GLOB_REL)CHANGELOG.md"


bwplotka (commented):
This is by design, like this and it's done by our automation. IF some links does not work, it should be fixed within automation project. What is not working? From GH or from website?

cc @saswatamcode

saswatamcode:
I think these links (plus any non-markdown resource links) are broken on the website but fine on GH. An automatic [solution](https://github.com/bwplotka/mdox/pull/76) for this was already implemented in mdox but the config wasn't updated here. 

It just involves adding the line below to [`.mdox.yaml`](https://github.com/thanos-io/thanos/blob/main/.mdox.yaml),
```yaml
linkPrefixForNonMarkdownResources: "https://github.com/thanos-io/thanos/blob/main"
```

Maybe it can be updated in this PR?  🙂  
@bwplotka @clyang82

clyang82:
Thanks @bwplotka @saswatamcode for your comments.  Good to know we have automatic solution for this. Let me use this PR to update `.mdox.yaml`.

saswatamcode (approved):
Thank you! All the links that were broken seem to be fixed now! 💫

yeya24 (approved):
Thanks for the contribution!

PR #4827, opened 2021-11-01
adds alert if a block is not uploaded during last 6 hours

spaparaju:
Signed-off-by: SriKrishna Paparaju <paparaju@gmail.com>

## Changes
- Added new alert `ThanosStoreObjstoreOperationBlocksMissing` to flag not even a single block has been uploaded during last 6 hours.

## Verification

- make example-rules-lint
- make examples


CHANGELOG.md
```diff
@@ -18,6 +18,7 @@ We use *breaking :warning:* to mark changes that are not backward compatible (re
 - [#4710](https://github.com/thanos-io/thanos/pull/4710) Store: add metric to capture timestamp of the last loaded block.
 - [#4736](https://github.com/thanos-io/thanos/pull/4736) S3: Add capability to use custom AWS STS Endpoint.
 - [#4764](https://github.com/thanos-io/thanos/pull/4764) Compactor: add `block-viewer.global.sync-block-timeout` flag to set the timeout of synchronization block metas.
+- [#4827](https://github.com/thanos-io/thanos/pull/4827) Compactor: add a new alert `ThanosStoreObjstoreOperationBlocksMissing` to flag not even a single block has been uploaded during last 6 hours.
```
yeya24, line 21: Is this alert related to the compactor?

spaparaju: It is not to Compactor.. I have fixed this.

CHANGELOG.md
```diff
@@ -18,6 +18,7 @@ We use *breaking :warning:* to mark changes that are not backward compatible (re
 - [#4710](https://github.com/thanos-io/thanos/pull/4710) Store: add metric to capture timestamp of the last loaded block.
 - [#4736](https://github.com/thanos-io/thanos/pull/4736) S3: Add capability to use custom AWS STS Endpoint.
 - [#4764](https://github.com/thanos-io/thanos/pull/4764) Compactor: add `block-viewer.global.sync-block-timeout` flag to set the timeout of synchronization block metas.
+- [#4827](https://github.com/thanos-io/thanos/pull/4827) Store: add a new alert `ThanosStoreObjstoreOperationBlocksMissing` to flag not even a single block has been uploaded during last 6 hours.
```
yeya24, line 21: Uploaded is confusing. Can we just use `loaded`?

examples/alerts/alerts.yaml
```diff
@@ -388,6 +388,19 @@ groups:
...
       severity: warning
+  - alert: ThanosStoreObjstoreOperationBlocksMissing
+    annotations:
+      description: Thanos Store {{$labels.job}}%s Bucket could not upload a single
+        block during the last {{$value}} seconds
+      runbook_url: https://github.com/thanos-io/thanos/tree/main/mixin/runbook.md#alert-name-thanosstoreobjstoreoperationblocksmissing
+      summary: During the last 6 hours, Thanos Store could not upload a single block.
+    expr: |
+      (
+        absent(thanos_bucket_store_blocks_last_loaded_timestamp_seconds != 1) and (time() - thanos_bucket_store_blocks_last_loaded_timestamp_seconds > 6 * 60 * 60)
```
yeya24, line 399: Is `absent(thanos_bucket_store_blocks_last_loaded_timestamp_seconds != 1)` valid? I guess we need to use `absent(thanos_bucket_store_blocks_last_loaded_timestamp_seconds) != 1`

yeya24 (commented):
What's the usecase of this alert? In the case of stores time partition, it is possible that the store gateway which serves old data doesn't get any uploaded data for a long period.

spaparaju:
> What's the usecase of this alert? In the case of stores time partition, it is possible that the store gateway which serves old data doesn't get any uploaded data for a long period.

Right, for serving old data this alert might be noisy / need to silenced. For platforms which continuously ingest data, this alert can lookout for any data loss.

examples/alerts/alerts.yaml
spaparaju: Good catch! 👍  fixed the query.

spaparaju:
@yeya24 the review comments been taken care.

yeya24:
Personally, I prefer to add alerts that can be adapted to all the scenarios.
This alert doesn't work in the case I mentioned so it might be better to be configured somewhere else as an additional alerting rule.

That's just my personal opinion so maybe let's get more feedback first.
WDYT? @spaparaju

matej-g:
> That's just my personal opinion so maybe let's get more feedback first. WDYT? @spaparaju

My 2 cents, but to me it also seems like this alert would potentially be too noise / producing false positives. I'd say if users do have the need to alert (in mentioned scenario where continuous ingestion is expected), they can introduce the alert on their own, based on the existing metric. To have this included in the mixins, it might cause more headaches to users than benefit.

spaparaju:
@yeya24 @matej-g thanks for the feedback. I was trying to get this alert closer to Thanos upstream as the underlying metric for tracking `Thanos Store block load time` is specific to Thanos. Yes, I agree with you both that alert might make sense for the upstream hosted services than at the underlying tech. layer. Closing out this issue. Thanks!

Issue #1972, opened 2020-01-09
Move tracing to OpenTelemetry


bwplotka:
It would be nice to reuse ready libraries then invent and maintain our own (: 

AC:
* Our tracing library is replaced by https://opentelemetry.io/
* Support for existing providers is preserved: Jaeger, Elastic, Lightstep, Stackdriver

squat:
Would love to see this. And we’ll be in line with other projects moving in this direction eg prom

kakkoyun:
I'm happy to work on this.

daixiang0:
@bwplotka i am happy to work on this.

bwplotka:
Awesome! Thanks for taking a look. My thoughts:

* What about different backends (e.g lightstep, ELK APM) that OpenTelemetry does not support?
* Definitely IF we are doing the change, let's start small, so option 1.

> BTW, OpenTelemetry is still under active development, some features in OpenTracing may be not supported i think.

* Yes, that's my worry. That's why we should be careful. Let's focus now on what is missing that we are using? I guess small try does not hurt, if you want to give it a go - but you need to make sure you test it against e.g some open source provider like Jaeger.

This is kind of worrying state of Go right now. Maybe it's worth to put a friendly issue on https://github.com/open-telemetry/opentelemetry-go github with a question for current dates/blockers?

![image](https://user-images.githubusercontent.com/6950331/73429195-f4a37400-4332-11ea-81f1-bd9d9c0152e7.png)

daixiang0:
I file a [issue](https://github.com/open-telemetry/opentelemetry-go/issues/453) about Elastic/Lightstep support, i would try the bridge.

lawrencejones:
Given opencencus and opentracing are both mature libraries with LTS guarantees, would it make more sense to pick one of those for now? I've held off using opentelemetry while it's in alpha to avoid needless changes, and under the expectation that opencencus/opentracing -> opentelemetry will eventually become a well worn migration path.

bwplotka:
We use opentracing right now, but it does not define actual RPC protocol, only client API. So we are forced to maintain client code against different tracing backends ourselves. Something that we would like to avoid

kakkoyun:
As far as I know, OpenTelemetry has a fairly stable API for tracing. 
However, they are still in alpha, maybe we should wait for a little bit more. https://opentelemetry.io/project-status/

daixiang0:
[...]

PR #4838, opened 2021-11-07
Tracing: Migrate Google Cloud (Stackdriver) client to OpenTelemetry

matej-g:

## Changes

This is first in the series of PRs to start migrating our tracing from the OpenTracing packages to OpenTelemetry. This PR in particular moves Google Cloud provider to OpenTelemetry. In details:

- Migrate Stackdriver to OpenTelemetry, rename to Google Cloud, while keeping the configuration backwards compatible
- Introduce utilities (bridge tracer and sampler), which will be required to build the OpenTelemetry tracer(s) while keeping the current behavior and features (e.g. forcing tracing by HTTP header)
- Adjust HTTP middleware
- Update documentation

Partially addresses https://github.com/thanos-io/thanos/issues/1972.

## Verification

Tests have been updated; the change has been also verified manually by sending traces to Google Cloud.


pkg/tracing/google_cloud/google_cloud.go
```diff
@@ -0,0 +1,84 @@
...
+}
+
+func collectAttributes(serviceName string) []attribute.KeyValue {
+	attr := []attribute.KeyValue{
+		semconv.ServiceNameKey.String(serviceName),
+		attribute.String("binary_revision", version.Revision),
+	}
+
+	if len(os.Args) > 1 {
+		attr = append(attr, attribute.String("binary_cmd", os.Args[1]))
```
yeya24, line 80: Why do we need to have binary Infos to attributes? Maybe remove this or at least make this configurable?

hanjm: Opentelemetry specification says lots of semantic_conventions , we should use it as possible. like `process.executable.name`. https://github.com/open-telemetry/opentelemetry-specification/blob/main/specification/resource/semantic_conventions/process.md#process

matej-g: Honestly, this was purely to keep the behavior, attribute name etc. the same as it is now, as we wanted to make the change as 'backwards compatible' as possible. But yes, OTEL has more detailed guidance on semantic conventions and I'm definitely planning to follow as we move the rest of the instrumentation to OTEL. However, if the feeling is that this can be changed already here, I'm happy to follow through.

matej-g:
ping @yeya24 @hanjm if you have interest and capacity to review by any chance, I'd be grateful :pray:

hanjm:
💪 LGTM

hanjm:
ping @yeya24

pkg/tracing/google_cloud/google_cloud.go
```diff
@@ -0,0 +1,84 @@
...
+		return nil, err
+	}
+
+	return newTracerProvider(ctx, logger, tracesdk.NewBatchSpanProcessor(exporter),
+		config.SampleFactor, config.ServiceName), nil
+}
+
+func newTracerProvider(ctx context.Context, logger log.Logger, processor tracesdk.SpanProcessor, sampleFactor uint64, serviceName string) *tracesdk.TracerProvider {
+	fraction := 1 / float64(sampleFactor)
+	if sampleFactor == 0 {
```
yeya24, line 50: I guess this check should be before line 59? @matej-g

pkg/tracing/google_cloud/google_cloud.go
```diff
@@ -0,0 +1,84 @@
...
+		fraction = 0
+	}
+
+	resource, err := resource.New(ctx, resource.WithAttributes(collectAttributes(serviceName)...))
+	if err != nil {
+		level.Warn(logger).Log("msg", "detecting resources for tracing provider failed", "err", err)
```
yeya24, line 56: If there is an error, should we return it or do we need to deal with it in line 67 as well?

yeya24:
Would like to move this forward 😄

pkg/tracing/google_cloud/google_cloud.go
matej-g: If there's an error, we should still receive valid (although possibly empty) `resource`. I think best we can do is log warning and move on (or we can possibly change level to `error`). I'll add a comment as well.

matej-g:
Hey @yeya24 thanks for the ping. I updated the PR and also now updated OTEL to version `1.3.0`.

However, after further testing, I'm noticing the traces get 'broken' between components :disappointed:. For example when executing query -> receiver, it shows up as _two_ distinct traces for each component - it gets 'split' on the store gRPC operation between client and server; I'm not sure exactly why, whether some changes are required to the middleware as well. I'll need more time to sort it out.

hanjm:
@matej-g are you do inject span context to header in grpc/http client? like this https://github.com/open-telemetry/opentelemetry-go-contrib/blob/main/instrumentation/google.golang.org/grpc/otelgrpc/interceptor.go#L94

matej-g:
Apologies for the long delay on this one, but I needed to find a bit of time to pin this down.

@hanjm I could not at first understand why the data was not being propagated, as you point out, but in the end I discovered that the OpenTelemetry's `BridgeTracer` [supports only HTTP headers](https://github.com/open-telemetry/opentelemetry-go/blob/main/bridge/opentracing/bridge.go#L626) as a carrier when propagating. Since in the gRPC middleware we are using the [gRPC metadata](https://github.com/grpc-ecosystem/go-grpc-middleware/blob/v2.0.0-rc.2/interceptors/tracing/client.go#L95) as a carrier, the trace injection was always failing due to unsupported carrier. I created a wrapper for the bridge tracer to work around this. I have tested is as well on Google Cloud and now the traces are looking as expected:
trace_example

cc @yeya24

matej-g:
Gentle ping @hanjm @yeya24, if you'd get a chance to re-review, thank you :pray:

pkg/tracing/client/factory.go
```diff
@@ -16,18 +16,20 @@ import (
...
 
 type TracingProvider string
 
 const (
-	STACKDRIVER TracingProvider = "STACKDRIVER"
-	JAEGER      TracingProvider = "JAEGER"
-	ELASTIC_APM TracingProvider = "ELASTIC_APM"
-	LIGHTSTEP   TracingProvider = "LIGHTSTEP"
+	STACKDRIVER  TracingProvider = "STACKDRIVER"
+	GOOGLE_CLOUD TracingProvider = "GOOGLE_CLOUD"
```
hanjm, line 29: golang const should be camelCase https://github.com/golang/go/wiki/CodeReviewComments#mixed-caps

hanjm:
Overall LGTM, Thanks you.

hanjm (approved):

yeya24:
We should be good to go after fixing @hanjm 's comments 👍

pkg/tracing/client/factory.go
matej-g: Looks like we have capitalized const names in other places handling configs (e.g. also in cache factory code), but I cannot find any reason for this. I'll fix it :+1:

matej-g:
Thanks @hanjm @yeya24, hopefully now this is good to go!

yeya24 (approved):
LGTM

PR #4839, opened 2021-11-08
docs: update receiver docs

jmichalek132:
Signed-off-by: Juraj Michalek <juraj.michalek132@gmail.com>

## Changes
Adding documentation for new thanos receive modes. (Resurrecting https://github.com/thanos-io/thanos/pull/4316)

## Verification
Just docs changes


jmichalek132:
Should be also the thanos receive split [proposal ](https://thanos.io/tip/proposals-accepted/202012-receive-split.md/)moved into the done section?

yeya24:
> Should be also the thanos receive split [proposal ](https://thanos.io/tip/proposals-accepted/202012-receive-split.md/)moved into the done section?

Yeah. SGTM

jmichalek132:
Hey, since you had comments on the previous PR any thoughts on this one? @ianbillett @bwplotka

docs/components/receive.md
```diff
@@ -12,6 +12,12 @@ For more information please check out [initial design proposal](../proposals-don
...
 
+As of version [v0.22.0](https://github.com/thanos-io/thanos/blob/main/CHANGELOG.md#added), the Thanos receiver implements the [receiver split proposal](../proposals-done/202012-receive-split.md), which allows data ingestion as well as request routing to be independently enabled/disabled. Thanks to these changes only routing-receivers need to watch for configuration changes, and ingesting-receivers do not need to. So when the configuration is updated, the routing-receivers will almost instantaneously re-configure without a lengthy WAL flush (& downtime) like we have in the current setup (routing & ingesting receivers). This would also enable users to prepare a topology of receivers, containing trees of receiver with depth **N**.
+
+![Example Receive architecture diagram](https://docs.google.com/drawings/d/e/2PACX-1vQaLa9EdF_frGmE6zbK48Zj9a8lIKxdx8NpOCU0eFizGCALRY8uUzZfFJLH8VNvtjyi-YBmVHq6PR8A/pub?w=1442&h=563)
+
+The current behaviour for receiver, however has not been modified, and works as expected. So users who are using receiver for both routing and ingestion, should not notice any changes as such. The change is backwards compatible.
```
squat, line 19: Again as mentioned earlier, the meaning of `current` differs between here and the previous sentence. Instead of describing how things _used_ to be and how they are now, I think it would provide considerably more value to the user to describe example configurations for the receiver that are possible _today_ and to discuss their trade-offs.

docs/components/receive.md
```diff
@@ -12,6 +12,12 @@ For more information please check out [initial design proposal](../proposals-don
 
 > NOTE: As the block producer it's important to set correct "external labels" that will identify data block across Thanos clusters. See [external labels](../storage.md#external-labels) docs for details.
 
+As of version [v0.22.0](https://github.com/thanos-io/thanos/blob/main/CHANGELOG.md#added), the Thanos receiver implements the [receiver split proposal](../proposals-done/202012-receive-split.md), which allows data ingestion as well as request routing to be independently enabled/disabled. Thanks to these changes only routing-receivers need to watch for configuration changes, and ingesting-receivers do not need to. So when the configuration is updated, the routing-receivers will almost instantaneously re-configure without a lengthy WAL flush (& downtime) like we have in the current setup (routing & ingesting receivers). This would also enable users to prepare a topology of receivers, containing trees of receiver with depth **N**.
```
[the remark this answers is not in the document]
jmichalek132: How about this?
```suggestion
Since the [receiver split proposal](../proposals-done/202012-receive-split.md) has been implemented in thanos receive, this allows data ingestion as well as request routing to be independently enabled/disabled. Thanks to these changes only routing-receivers need to watch for configuration changes, and ingesting-receivers do not need to. So when the configuration is updated, the routing-receivers will almost instantaneously re-configure without a lengthy WAL flush (& downtime) like we have in the current setup (routing & ingesting receivers). This would also enable users to prepare a topology of receivers, containing trees of receiver with depth **N**.
```

docs/components/receive.md
jmichalek132: I think it's still needs to be called out in some manner that without changing the provided flags the behavior won't change with upgrading to new version,

docs/components/receive.md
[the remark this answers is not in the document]
jmichalek132: @squat could you please take a second look at this pr :) ?

docs/components/receive.md
squat: Fair to call it out, but in that case the differentiator should not be time, e.g. "since x", but rather configuration, e.g. "when running the receiver like this, X happens; when running it like that, Y happens". Making it temporal is less direct and as such is not as useful to users IMO and quickly becomes out of date

PR #4848, opened 2021-11-10
Compactor: Add metrics for retention progress

### LINKED ISSUE #4835: Compact: Retention progress metrics

Because retention happens after compaction and downsampling, so it might get huge retention backlog to do in the object storage. We also got the same problem mentioned in https://github.com/thanos-io/thanos/pull/4406.

It would be great to have a way to tell how many blocks are beyond their retention period but don't get deleted.

We have introduced the progress calculator to get compaction and downsampling progress as metrics in https://github.com/thanos-io/thanos/pull/4801.

We can implement another `RetentionProgressCalculator` to do this work.

metonymic-smokey:
Fixes #4835 

## Changes
Added a retention progress calculator, similar to the ones for compaction and downsampling.


yeya24:
Change is relevant to users. Can you update the changelog?

pkg/compact/compact.go
```diff
@@ -667,6 +667,55 @@ func (ds *DownsampleProgressCalculator) ProgressCalculate(ctx context.Context, g
 	return nil
 }
 
+// RetentionProgressMetrics contains Prometheus metrics related to retention progress.
+type RetentionProgressMetrics struct {
+	NumberOfBlocksDeleted *prometheus.GaugeVec
```
yeya24, line 672: `NumberOfBlocksDeleted` uses past tense but the progress is future tense. Let's rename it to `NumberOfBlocksToDelete` or `NumberOfBlocksRetention`?

pkg/compact/compact.go
```diff
@@ -667,6 +667,55 @@ func (ds *DownsampleProgressCalculator) ProgressCalculate(ctx context.Context, g
...
+	retentionByResolution map[ResolutionLevel]time.Duration
+}
+
+// NewRetentionProgressCalculator creates a new RetentionProgressCalculator.
+func NewRetentionProgressCalculator(reg prometheus.Registerer, retentionByResolution map[ResolutionLevel]time.Duration) *RetentionProgressCalculator {
+	return &RetentionProgressCalculator{
+		retentionByResolution: retentionByResolution,
+		RetentionProgressMetrics: &RetentionProgressMetrics{
+			NumberOfBlocksDeleted: promauto.With(reg).NewGaugeVec(prometheus.GaugeOpts{
+				Name: "thanos_compact_todo_deleted_blocks",
```
yeya24, line 687: Same here. What about `thanos_compact_todo_retention_blocks`?

yeya24 (dismissed):
LGTM. Only some small nits about naming. Let's mark this pr for review.

pkg/compact/compact.go
metonymic-smokey: @yeya24 I think `deletion_blocks` is more suitable than `retention_blocks`.

pkg/compact/compact_test.go
```diff
@@ -205,6 +205,148 @@ func createBlockMeta(id uint64, minTime, maxTime int64, labels map[string]string
 	return m
 }
 
+func TestRetentionProgressCalculate(t *testing.T) {
+	logger := log.NewNopLogger()
+	reg := prometheus.NewRegistry()
+	unRegisterer := &receive.UnRegisterer{Registerer: reg}
```
yeya24, line 211: I believe we don't need to use unregisterer. Can you also remove it in the compact and downsample tests?

metonymic-smokey: Thanks!
Could you PTAL at my latest commit and let me know if that's what's expected?

yeya24 (approved):
LGTM

PR #4850, opened 2021-11-10
bump codemirror-promql and codemirror to the last version

Nexucis:


Nexucis:
as a side note, this bump is introducing the support of the trigonometric functions in the UI (like rad, cos ...etc.).
I don't know if you guys already did the bump of the tsdb. So just in case, this PR should be merged once the tsdb has been bumped.

Nexucis:
I think this one is for you @onprem

pkg/testutil/e2eutil/sysprocattr.go
```diff
@@ -1,6 +1,7 @@
 // Copyright (c) The Thanos Authors.
 // Licensed under the Apache License 2.0.
 
+//go:build !linux
```
onprem, line 4: This seems unrelated to the current changes.

onprem (approved):
LGTM, great work as always. Thanks!

I have verified on the current main that the tsdb version we are pulling supports these functions.

pkg/testutil/e2eutil/sysprocattr.go
Nexucis: yeah that happens when I ran the command `make assets` it called `make format` and this one is doing this.
Maybe it's a missed when moving to golang v1.17?

PR #4856, opened 2021-11-11
Mixin: Add Query Frontend grafana dashboard

jessicalins:

## Changes

Hey! :wave: 
I've added a grafana dashboard for the Query Frontend component. I used the Query/Store dashboards as base.
Here is a screenshot of how it currently looks like:

query_frontend_dashboard

I'm not sure if all metrics that are interesting are represented in the dashboard, but I've tried to add some basic ones (RED metrics/caching/resources)

## Verification

I've ran `make examples` to generate the `query_frontend.json` file and I've imported it to Grafana to get an idea how it looks like.

### DIFF
--- a/mixin/config.libsonnet
+++ b/mixin/config.libsonnet
@@ -28,7 +28,7 @@
     selector: 'job=~".*thanos-query.*"',
     title: '%(prefix)sQuery' % $.dashboard.prefix,
   },
-  queryFrontend+:: {
+  query_frontend+:: {
     selector: 'job=~".*thanos-query-frontend.*"',
     title: '%(prefix)sQueryFrontend' % $.dashboard.prefix,
   },
--- a/mixin/config.libsonnet
+++ b/mixin/config.libsonnet
@@ -28,7 +28,7 @@
     selector: 'job=~".*thanos-query.*"',
     title: '%(prefix)sQuery' % $.dashboard.prefix,
   },
-  query_frontend+:: {
+  queryFrontend+:: {
     selector: 'job=~".*thanos-query-frontend.*"',
     title: '%(prefix)sQuery Frontend' % $.dashboard.prefix,
   },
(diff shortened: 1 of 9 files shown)


mixin/config.libsonnet
```diff
@@ -28,6 +28,10 @@
     selector: 'job=~".*thanos-query.*"',
     title: '%(prefix)sQuery' % $.dashboard.prefix,
   },
+  query_frontend+:: {
```
jessicalins, line 31: I saw that we would like to fix the naming convention to e.g. `queryFrontend` (this TODO: https://github.com/thanos-io/thanos/blob/main/mixin/config.libsonnet#L52)
I've tried at first to use it like that, but when running `make examples` and when it tries to generate the json file, I get:
```bash
/home/jessica/go/bin/jsonnet-v0.17.0 -J mixin/vendor -m examples/dashboards mixin/dashboards.jsonnet
RUNTIME ERROR: Field does not exist: query_frontend
```
Does someone have an idea of what it could be?

saswatamcode: I was able to replicate this error by changing everything to `queryFrontend` except the `query_frontend.json` in [this line](https://github.com/thanos-io/thanos/pull/4856/files#diff-1ac563a8576dac71b781fbc9a9e8d301eaeb833d1de582673845f176e68d2c74R15). Once changed to `queryFrontend.json`, `make examples` runs successfully!

I think this is due to how examples are generated by [`dashboards.jsonnet`](https://github.com/thanos-io/thanos/blob/main/mixin/dashboards.jsonnet).

jessicalins: thanks, that solved the problem!
since I'm already touching those files, should I also update https://github.com/thanos-io/thanos/blob/main/mixin/config.libsonnet#L52 to this convention in this PR? or better in another one?

bwplotka (approved):
Amazing work!

Let's merge and iterate, but looks good from generally! Thanks 💪🏽

mixin/config.libsonnet
saswatamcode: Yes, it would be better to do the update in a separate PR! 🙂

jessicalins: thanks, added here: https://github.com/thanos-io/thanos/pull/4859

PR #4859, opened 2021-11-12
Mixin: Follow naming convention for bucketReplicate

jessicalins:
Signed-off-by: Jéssica Lins <jlins@redhat.com>

## Changes

As discussed [here](https://github.com/thanos-io/thanos/pull/4856#discussion_r748283282) and to follow up on this TODO: https://github.com/thanos-io/thanos/blob/b1c7483ea9be11a81e2c7422f19b811e81b98859/mixin/config.libsonnet#L56-L57
I just renamed `bucket_replicate` -> `bucketReplicate` to keep the naming convention consistent.

## Verification

After the changes I've ran `make examples` to regenerate the files.

### DIFF
--- a/examples/alerts/alerts.md
+++ b/examples/alerts/alerts.md
@@ -578,7 +578,7 @@ rules:
     description: Thanos Replicate is failing to run, {{$value | humanize}}% of attempts
       failed.
     runbook_url: https://github.com/thanos-io/thanos/tree/main/mixin/runbook.md#alert-name-thanosbucketreplicateerrorrate
-    summary: Thanose Replicate is failing to run.
+    summary: Thanos Replicate is failing to run.
   expr: |
     (
       sum by (job) (rate(thanos_replicate_replication_runs_total{result="error", job=~".*thanos-bucket-replicate.*"}[5m]))
@@ -612,17 +612,6 @@ rules:
 ```yaml mdox-exec="cat examples/tmp/thanos-component-absent.yaml"
 name: thanos-component-absent
 rules:
-- alert: ThanosBucketReplicateIsDown
-  annotations:
-    description: ThanosBucketReplicate has disappeared. Prometheus target for the
-      component cannot be discovered.
-    runbook_url: https://github.com/thanos-io/thanos/tree/main/mixin/runbook.md#alert-name-thanosbucketreplicateisdown
-    summary: Thanos component has disappeared.
-  expr: |
-    absent(up{job=~".*thanos-bucket-replicate.*"} == 1)
-  for: 5m
-  labels:
-    severity: critical
 - alert: ThanosCompactIsDown
   annotations:
     description: ThanosCompact has disappeared. Prometheus target for the component
--- a/examples/alerts/alerts.yaml
+++ b/examples/alerts/alerts.yaml
@@ -542,7 +542,7 @@ groups:
       description: Thanos Replicate is failing to run, {{$value | humanize}}% of attempts
         failed.
       runbook_url: https://github.com/thanos-io/thanos/tree/main/mixin/runbook.md#alert-name-thanosbucketreplicateerrorrate
-      summary: Thanose Replicate is failing to run.
+      summary: Thanos Replicate is failing to run.
     expr: |
       (
         sum by (job) (rate(thanos_replicate_replication_runs_total{result="error", job=~".*thanos-bucket-replicate.*"}[5m]))
@@ -569,17 +569,6 @@ groups:
       severity: critical
 - name: thanos-component-absent
   rules:
-  - alert: ThanosBucketReplicateIsDown
-    annotations:
-      description: ThanosBucketReplicate has disappeared. Prometheus target for the
-        component cannot be discovered.
-      runbook_url: https://github.com/thanos-io/thanos/tree/main/mixin/runbook.md#alert-name-thanosbucketreplicateisdown
-      summary: Thanos component has disappeared.
-    expr: |
-      absent(up{job=~".*thanos-bucket-replicate.*"} == 1)
-    for: 5m
-    labels:
-      severity: critical
   - alert: ThanosCompactIsDown
     annotations:
       description: ThanosCompact has disappeared. Prometheus target for the component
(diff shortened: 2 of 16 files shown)


examples/alerts/alerts.md
```diff
@@ -612,17 +612,6 @@ rules:
...
-  annotations:
-    description: ThanosBucketReplicate has disappeared. Prometheus target for the
-      component cannot be discovered.
-    runbook_url: https://github.com/thanos-io/thanos/tree/main/mixin/runbook.md#alert-name-thanosbucketreplicateisdown
-    summary: Thanos component has disappeared.
-  expr: |
-    absent(up{job=~".*thanos-bucket-replicate.*"} == 1)
-  for: 5m
-  labels:
-    severity: critical
```
yeya24, line 625: Looks like this alert was removed at https://github.com/thanos-io/thanos/pull/3856. Is this an intended change? @kakkoyun

saswatamcode (dismissed):
Thanks for this! 💫
Docs check seems to fail due to `examples/alerts/alerts.yaml` not being updated. I think running `make examples-clean` and then `make docs` will fix this.

Maybe this is a Makefile issue due to `examples/alerts/alerts.yaml` not being [phony](https://www.gnu.org/software/make/manual/html_node/Phony-Targets.html). 🤔

examples/alerts/alerts.yaml
```diff
@@ -569,17 +569,6 @@ groups:
       severity: critical
 - name: thanos-component-absent
   rules:
-  - alert: ThanosBucketReplicateIsDown
```
saswatamcode, line 572: I think the circleCI test failure is related now, as `alerts.yaml` was changed, so the `thanos-component-absent` group now has 6 alerts instead of 7. I think removing one `someAlert` in [`rules_test.go`](https://github.com/thanos-io/thanos/blob/main/pkg/rules/rules_test.go#L53) should do the trick! 🙂

examples/alerts/alerts.md
kakkoyun: Yes, there is another generator that handles the generation of these alerts. This was redundant.

examples/alerts/alerts.yaml
kakkoyun: Exactly this. Maybe using these alerts in ruler tests was not a really good idea. These sort of changes shouldn't break something in go tests. IMHO we should clone some of these file on a `testdata` folder in ruler tests and use those instead of alerts.yaml

examples/dashboards/dashboards.md
```diff
@@ -10,7 +10,7 @@ There exists Grafana dashboards for each component (not all of them complete) ta
 - [Thanos Receiver](receive.json)
 - [Thanos Sidecar](sidecar.json)
 - [Thanos Ruler](rule.json)
-- [Thanos Replicate](bucket_replicate.json)
+- [Thanos Replicate](bucketReplicate.json)
```
kakkoyun, line 13:
```suggestion
- [Thanos Replicate](bucket_replicate.json)
```

I think we should keep this convention and change the `queryFrontend.json` above for filenames.

mixin/README.md
```diff
@@ -113,8 +113,7 @@ This project is intended to be used as a library. You can extend and customize d
     thanosPrometheusCommonDimensions: 'namespace, pod',
     title: '%(prefix)sSidecar' % $.dashboard.prefix,
   },
-  // TODO(kakkoyun): Fix naming convention: bucketReplicate
-  bucket_replicate+:: {
+  bucketReplicate+:: {
```
kakkoyun, line 116: 👍

kakkoyun (changes requested):
Looks good. Thanks for taking care of this.

I have already commented inline, it would be amazing to keep snake_case for filenames but use camelCase for variable names in code.

examples/dashboards/dashboards.md
jessicalins: I tried to do this [here](https://github.com/thanos-io/thanos/pull/4856#discussion_r747791376) but it was failing, because we were using the first part of the filename to set the `component` variable: https://github.com/thanos-io/thanos/blob/main/mixin/dashboards/defaults.libsonnet#L16

So for this case, we would get `component = bucket_replicate` from `bucket_replicate.json`, instead of `bucketReplicate`.

I thought then we could have a jsonnet function that handles this case - so in case we can split the filename by `_`, we update the component name to use `camelCase` instead of `snake_case`.

I tried to do this in this commit: https://github.com/thanos-io/thanos/pull/4859/commits/aa46b6b5fc9bfba6c2f7b34a145dc8a06b78af5d so now it should hopefully work, either if we use filenames as `bucket_replicate.json`, `query_frontend.json`  or if we use 'simple filenames' - e.g. `query.json`

I'm new to jsonnet so maybe the code is not that readable/there are better ways to do this, but the idea was to split the component name by `_` and, in case we have more then 1 word, then we make the first letter of the other words uppercase, then we join all of them in one word again to have the component name in camelCase format.

Let me know what you think.. I can also move this commit/the related changes to other PR :)

examples/alerts/alerts.yaml
jessicalins: makes sense, changed it!

mixin/dashboards/defaults.libsonnet
```diff
@@ -13,7 +14,11 @@
   // Automatically add a uid to each dashboard based on the base64 encoding
   // of the file name and set the timezone to be 'default'.
   grafanaDashboards:: {
-    local component = std.split(filename, '.')[0],
+    local componentName = std.split(filename, '.')[0],
+    local componentParts = std.split(componentName, '_'),
```
kakkoyun, line 18: just a nit: We can have a `utils.sanitizeComponentName` to hide all the implementation details behind a function.

examples/dashboards/dashboards.md
kakkoyun: This looks good to me. I think you did an amazing job :+1:

kakkoyun (dismissed):
LGTM :rocket:

yeya24 (approved):
Not sure why CodeQL job is queued. Merge it manually...

PR #4865, opened 2021-11-13
Make progress metrics interval configurable

yeya24:

## Changes

1. Only enable progress metrics if `wait` mode is enabled
2. Make the interval configurable instead of hardcoded 5m
3. Remove the original `progress-metrics` bool flag. Now to disable progress metrics, just set interval to 0.

### DIFF
--- a/cmd/thanos/compact.go
+++ b/cmd/thanos/compact.go
@@ -693,7 +693,7 @@ func (cc *compactConfig) registerFlag(cmd extkingpin.FlagClause) {
 		Default("5m").DurationVar(&cc.blockViewerSyncBlockTimeout)
 	cmd.Flag("compact.cleanup-interval", "How often we should clean up partially uploaded blocks and blocks with deletion mark in the background when --wait has been enabled. Setting it to \"0s\" disables it - the cleaning will only happen at the end of an iteration.").
 		Default("5m").DurationVar(&cc.cleanupBlocksInterval)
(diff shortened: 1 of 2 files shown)


cmd/thanos/compact.go
```diff
@@ -694,6 +693,8 @@ func (cc *compactConfig) registerFlag(cmd extkingpin.FlagClause) {
 		Default("5m").DurationVar(&cc.blockViewerSyncBlockTimeout)
 	cmd.Flag("compact.cleanup-interval", "How often we should clean up partially uploaded blocks and blocks with deletion mark in the background when --wait has been enabled. Setting it to \"0s\" disables it - the cleaning will only happen at the end of an iteration.").
 		Default("5m").DurationVar(&cc.cleanupBlocksInterval)
+	cmd.Flag("compact.progress-interval", "How often we should calculate the compaction progress in the background when --wait has been enabled. Setting it to \"0s\" disables it. Now compaction, downsampling and retention progress are supported.").
```
metonymic-smokey, line 696: Just a small nit. I think phrasing it this way might be better: "frequency of calculating the compaction progress in the background when --wait has been enabled". WDYT?

yeya24: Done.

yeya24:
I'll merge it now as it is a small change.

GiedriusS (commented):
:+1:

PR #4874, opened 2021-11-16
query: add `--endpoint-strict` flag to statically configure endpoints

hitanshu-mehta:
This PR completes 3rd task from [this](https://github.com/thanos-io/thanos/issues/2600#issuecomment-967241095) list.

## Changes
- Added `--endpoint-strict` flag (similar to `--store-strict`) which can be used to add statically configured endpoints.

### DIFF
--- a/cmd/thanos/query.go
+++ b/cmd/thanos/query.go
@@ -121,8 +121,11 @@ func registerQuery(app *extkingpin.App) {
 	targetEndpoints := cmd.Flag("target", "Deprecation Warning - This flag is deprecated and replaced with `endpoint`. Experimental: Addresses of statically configured target API servers (repeatable). The scheme may be prefixed with 'dns+' or 'dnssrv+' to detect target API servers through respective DNS lookups.").
 		Hidden().PlaceHolder("<target>").Strings()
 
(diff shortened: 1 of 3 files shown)


GiedriusS (dismissed):

cmd/thanos/query.go
```diff
@@ -398,6 +403,12 @@ func runQuery(
 		dns.ResolverType(dnsSDResolver),
 	)
 
+	for _, endpoint := range strictEndpoints {
+		if dns.IsDynamicNode(endpoint) {
+			return errors.Errorf("%s is a dynamically specified endpoint i.e. it uses SD and that is not permitted under strict mode. Use --endpoint for this", endpoint)
+		}
+	}
+
```
matej-g, line 411: Nit: Having this right after doing the same for `strictStores` would read tiny bit better :blush:

matej-g (commented):
Let's go :rocket: Let's just not forget to update the CHANGELOG, otherwise good to go :raised_hands:.

matej-g (approved):
:heart:

GiedriusS (approved):

PR #4882, opened 2021-11-19
Update Prometheus to latest main + patch to bring back renamed packages.

pstibrany:
This PR updates Prometheus to latest master + https://github.com/prometheus/prometheus/pull/9824. This extra PR is necessary because Thanos depends on Cortex, which in turns still depends on removed packages.

We cannot update Prometheus in Cortex first, because Cortex also depends on Thanos, which also breaks without this update.


go.mod
```diff
@@ -94,9 +95,11 @@ replace (
 	// TODO: Remove this: https://github.com/thanos-io/thanos/issues/3967.
 	github.com/minio/minio-go/v7 => github.com/bwplotka/minio-go/v7 v7.0.11-0.20210324165441-f9927e5255a6
 	// Make sure Prometheus version is pinned as Prometheus semver does not include Go APIs.
-	github.com/prometheus/prometheus => github.com/prometheus/prometheus v1.8.2-0.20211101135822-b862218389fc
+	// github.com/prometheus/prometheus => github.com/prometheus/prometheus v1.8.2-0.20211119115433-692a54649ed7
+	// Custom fork with missing packages (testing, before upstreaming).
+	github.com/prometheus/prometheus => github.com/pstibrany/prometheus v1.8.2-0.20211119134801-de51b7b3200a
 	github.com/sercand/kuberesolver => github.com/sercand/kuberesolver v2.4.0+incompatible
-	google.golang.org/grpc => google.golang.org/grpc v1.29.1
+	// google.golang.org/grpc => google.golang.org/grpc v1.29.1
```
yeya24, line 102: Is this not necessary anymore?

pkg/block/index.go
```diff
@@ -14,17 +14,20 @@ import (
 	"strings"
 	"time"
 
+	"github.com/prometheus/prometheus/storage"
+
 	"github.com/thanos-io/thanos/pkg/block/metadata"
```
yeya24, line 19: Can you please help group these imports?

go.mod
pstibrany: With pin to v1.29.1 the code no longer compiles. I can move the pin to v1.40.0 if that's what you want.

pkg/block/index.go
pstibrany: I've fixed grouping of imports in all modified files, using common grouping order: stdlib, 3rd-party imports, and project imports. It may be good to add linter for this (eg. Cortex has one), but that would be a separate PR imho.

go.mod
yeya24: Yes please.

yeya24 (dismissed):
LGTM. Thanks for the great work!

go.mod
pstibrany: OK, done.

yeya24 (approved):

pstibrany:
Thank you for super-fast PR turnaround!

PR #4886, opened 2021-11-21
*: Implement info server for missing components

matej-g:

## Changes

This PR is the continuation of the series of PRs to resolve the issue https://github.com/thanos-io/thanos/issues/2600 implementing the endpoints proposal.

The changes include:
- Implementing the info server for remaining components - querier, receiver and ruler. This includes also adjustments to store implementation to make it possible to obtain necessary store information (time range, label set)
- Some improvements to info package and info server options (mainly simplifies providing default non-empty method implementations)
- Simplifies some parts of endpoint package (removes unnecessary interface and min/max timestamps)
- Re-enables getting endpoint metadata via store info method, since now all components in latest version implement the info server.

## Verification

Tests have been adjusted and passed locally. Manual test has been executed as well.


Makefile
```diff
@@ -237,7 +237,7 @@ test-e2e: docker
 	@echo ">> cleaning docker environment."
 	@docker system prune -f --volumes
 	@echo ">> cleaning e2e test garbage."
-	@rm -rf ./test/e2e/e2e_integration_test*
+	@rm -rf ./test/e2e/e2e_*
```
onprem, line 240: :smile:

pkg/info/info.go
```diff
@@ -26,53 +26,99 @@ type InfoServer struct {
...
+
+func (fn serverOptionFunc) applyServerOption(cfg *InfoServer) {
+	fn(cfg)
+}
+
+func WithLabelSetFunc(getLabelSet ...func() []labelpb.ZLabelSet) ServerOption {
+	if len(getLabelSet) == 0 {
+		return serverOptionFunc(func(s *InfoServer) {
+			s.getLabelSet = func() []labelpb.ZLabelSet { return []labelpb.ZLabelSet{} }
+		})
```
onprem, line 56: I am wondering if converting it to an `interface` is required? We can probably make it simpler by just using a type alias.

```suggestion
type ServerOptionFunc func(*InfoServer)

func WithLabelSetFunc(getLabelSet ...func() []labelpb.ZLabelSet) ServerOptionFunc {
	if len(getLabelSet) == 0 {
		return func(s *InfoServer) {
			s.getLabelSet = func() []labelpb.ZLabelSet { return []labelpb.ZLabelSet{} }
		}
```

And modifying the `NewInfoServer` just a bit.
```go
func NewInfoServer(
	component string,
	options ...ServerOptionFunc,
) *InfoServer {
	srv := &InfoServer{
		component: component,
	}

	for _, o := range options {
		o(srv)
	}

	return srv
}
```

WDYT?

pkg/info/info.go
```diff
@@ -26,53 +26,99 @@ type InfoServer struct {
...
+func WithStoreInfoFunc(getStoreInfo ...func() *infopb.StoreInfo) ServerOption {
+	if len(getStoreInfo) == 0 {
+		return serverOptionFunc(func(s *InfoServer) {
+			s.getStoreInfo = func() *infopb.StoreInfo { return &infopb.StoreInfo{} }
+		})
 	}
+
+	return serverOptionFunc(func(s *InfoServer) {
+		s.getStoreInfo = getStoreInfo[0]
+	})
```
onprem, line 73: I like this pattern :100: 

The only thing that might be a con is the function signature allows more than 1 arguments, and it is not clear to the users that anything other than the first argument will be ignored. Maybe we can add a comment on this function saying that.

onprem (approved):
Nice work @matej-g! This is looking good to go, just some small non-blocking suggestions.

Makefile
GiedriusS: :+1: this is good, I have a bunch of leftovers `e2e_12345` and so on :smile:

matej-g: Yeah, unfortunately this was forgotten during the e2e migration, should be good from now on :blush:

pkg/info/info.go
matej-g: You're right, the intent was to make this so that it cannot be implemented by others, but it does not really have any value in this context.

pkg/info/info.go
matej-g: That's a good suggestion, I wanted to add more comments overall in the info package, I'll do that.

onprem (approved):
LGTM, Thanks for this!

GiedriusS (approved):
:muscle: awesome work!

GiedriusS:
Ran tests locally and seems like everything works as expected

PR #4887, opened 2021-11-22
Add doc for thanos rule stateless mode

yeya24:

This pr adds the missing document for Thanos Ruler stateless mode.

### DIFF
--- a/docs/components/rule.md
+++ b/docs/components/rule.md
@@ -230,7 +230,7 @@ thanos rule \
 The remote write config file is exactly the same as the [Prometheus remote write config format](https://prometheus.io/docs/prometheus/latest/configuration/configuration/#remote_write).
 
 **NOTE:**
-1. Metadata config is not supported in this mode and will be ignored if provided in the remote write configuration.
+1. `metadata_config` is not supported in this mode and will be ignored if provided in the remote write configuration.
 2. Ruler won't expose Store API for querying data if stateless mode is enabled. If the remote storage is thanos receiver then you can use that to query rule evaluation results.
 
 ## Flags
--- a/docs/components/rule.md
+++ b/docs/components/rule.md
@@ -229,7 +229,7 @@ thanos rule \
 
 The remote write config file is exactly the same as the [Prometheus remote write config format](https://prometheus.io/docs/prometheus/latest/configuration/configuration/#remote_write).
 
-**NOTE:** 
+**NOTE:**
 1. Remote write metadata config is not supported in this mode and will be ignored if provided in the remote write configuration.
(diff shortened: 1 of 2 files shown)


docs/components/rule.md
```diff
@@ -202,6 +204,35 @@ In case of Ruler in HA you need to make sure you have the following labelling se
...
+    --objstore.config-file      "bucket.yml" \
+    --label                     'monitor_cluster="cluster1"' \
+    --label                     'replica="A"' \
+    --remote-write.config-file  'rw-config.yaml'
+```
+
+The remote write config file is exactly the same as the [Prometheus remote write config format](https://prometheus.io/docs/prometheus/latest/configuration/configuration/#remote_write).
+
+**NOTE:**
+1. Remote write metadata config is not supported in this mode and will be ignored if provided in the remote write configuration.
```
GiedriusS, line 233: What's "remote write metadata config"? :thinking:

docs/components/rule.md
```diff
@@ -6,7 +6,9 @@
 
 The `thanos rule` command evaluates Prometheus recording and alerting rules against chosen query API via repeated `--query` (or FileSD via `--query.sd`). If more than one query is passed, round robin balancing is performed.
 
-Rule results are written back to disk in the Prometheus 2.0 storage format. Rule nodes at the same time participate in the system as source store nodes, which means that they expose StoreAPI and upload their generated TSDB blocks to an object store.
+By default, rule evaluation results are written back to disk in the Prometheus 2.0 storage format. Rule nodes at the same time participate in the system as source store nodes, which means that they expose StoreAPI and upload their generated TSDB blocks to an object store.
+
+Rule also has a stateless mode which sends rule evaluation results to some remote storages via remote write for better scalability. This way, rule nodes only work as a data producer and the remote receive nodes work as source store nodes.
```
GiedriusS, line 11: Maybe let's clarify the difference even more:

```suggestion
Rule also has a stateless mode which sends rule evaluation results to some remote storages via remote write for better scalability. This way, rule nodes only work as a data producer and the remote receive nodes work as source store nodes. It means that Thanos Rule in this mode does _not_ expose the StoreAPI.
```

docs/components/rule.md
yeya24: image

Metadata config depends on scrape manager so it is not used.

docs/components/rule.md
yeya24: Done

hanjm:
Awsome, LGTM

docs/components/rule.md
GiedriusS: Ahh, I see. Maybe we can then replace "remote write metadata config" with "remote write `metadata_config`" to emphasize that it is a key's name?

yeya24: Done

GiedriusS (approved):

PR #4891, opened 2021-11-22
Add no lock option for stateless rule agent DB

yeya24:

Thanks to the version update in https://github.com/thanos-io/thanos/pull/4882, we can bring https://github.com/prometheus/prometheus/pull/9623 in to support the no lock option in the agent DB.


GiedriusS (approved):
:tada:

PR #4897, opened 2021-11-24
querier: Add validation for querier address flags

### LINKED ISSUE #4894: Query: Endpoint (store) flag should ignore empty or obviously incorrect addresses
**Thanos, Prometheus and Golang version used**: Thanos - latest `main` build

**What happened**:
When I provide empty or obviously wrong address to the `store` or `endpoint` flag, query still tries to contact the store to obtain metadata, which can never succeed, resulting in logs being spammed with:
```
level=warn ts=2021-11-23T13:21:35.047200231Z caller=endpointset.go:500 component=endpointset msg="update of node failed" err="getting metadata: fallback fetching info from : rpc error: code = Canceled desc = context canceled" address=
```

Thanos should ignore empty and obviously incorrect `store` / `endpoint` address; I would expect to be warned it the logs about the fact.

Build Thanos from latest `main` (or just use latest release) and run `query` with flags `--endpoint=` (empty) and `--endpoint=abc` (obviously wrong)

At minimum, we could check if the address is in valid format `<host>:<port>`; otherwise refuse it and give warning.

saswatamcode:

## Changes

This PR adds validation for querier address flags. 

Adds the `validateAddrs()` function which checks for duplicates, empty or invalid addresses provided using `--endpoint`, `--store`, `--rule`, `--metadata`, `--exemplar` or `--target` flags.

Fixes #4894.

## Verification

Tested locally.


cmd/thanos/query.go
```diff
@@ -733,20 +737,29 @@ func removeDuplicateEndpointSpecs(logger log.Logger, duplicatedStores prometheus
...
+			return errors.New("Address is empty.")
+		}
+
+		qtypeAndName := strings.SplitN(addr, "+", 2)
+		hostAndPort := strings.SplitN(addr, ":", 2)
+		if len(qtypeAndName) != 2 && len(hostAndPort) != 2 {
+			return errors.Errorf("Address %s is not of <host>:<port> format or a valid DNS query.", addr)
 		}
 
-		set[s] = struct{}{}
```
GiedriusS, line 746: Maybe we can also call https://pkg.go.dev/net#ParseIP here to check `hostAndPort[0]`?

GiedriusS (commented):
I was thinking that maybe instead of using `Strings()` on these flags, we could define a custom parser (https://github.com/alecthomas/kingpin#custom-parsers) that automatically calls `validateAddrs`? That way we wouldn't miss calling these functions on the provided values. What do you think? :thinking:

saswatamcode:
This would be better! Let me try it out and push a commit. 🙂

saswatamcode:
I've changed this to custom kingpin parser which is defined in `extkingpin/flags.go` now. 🙂

GiedriusS (approved):
:+1:

PR #4901, opened 2021-11-25
store+querier: QueryPushdown PoC using StoreAPI and nested PromQL.

bwplotka:
This yolo query pushdown implementation is semi-hardcorded for instant query `sum(count_over_time(<>[2w])) by (cluster, replica)`, but can be extended to any associative query. It allows Querier to detect detailed query hints (function, range step etc in nested fashion)  and pushdown them to proxy and underlying leafs. Leaf (in this case store) gets hints in request, fetches matched series from each block concurrently AND before giving to merge sort it aggregates using `AggregatedSeriesSet` wrapped with promQL engine. This needs a bit of work ofc to optimize PromQL engine for such usage. Hacked it for demo purposes for now.   

Pros of this approach:
* Ultra-simple
* Should be quicker (benchmarks in progress)

Cons:
* Does not work (YET!) (: 
* Still if the block is too large (too many series) we cannot shard query by series - this can be solved by series ID sharding on top of it. 


bwplotka:
Tests will not work because I hijacked dedup option to control QP

bwplotka:
I have some nice panic to fix, and it should work then ;p 

```
11:33:57 store1: pushed query on block
11:33:57 store1: : sum(count_over_time(name{}[1209600000ms])) by ( cluster, replica )
11:33:57 store1: panic: runtime error: invalid memory address or nil pointer dereference
11:33:57 store1: [signal SIGSEGV: segmentation violation code=0x1 addr=0x18 pc=0x1193449]
11:33:57 store1: goroutine 1655 [running]:
11:33:57 store1: github.com/thanos-io/thanos/pkg/store/promqlaggr.(*seriesQueryable).Close(0xc020b90160)
11:33:57 store1: <autogenerated>:1 +0x29
11:33:57 store1: github.com/prometheus/prometheus/promql.(*Engine).execEvalStmt(0xc020b422a0, {0x1ff7c30, 0xc020b48ab0}, 0xc000899020, 0xc01b3ba2d0)
11:33:57 store1: /home/bwplotka/Repos/thanosgopath/pkg/mod/github.com/prometheus/prometheus@v1.8.2-0.20211119115433-692a54649ed7/promql/engine.go:647 +0xe4d
11:33:57 store1: github.com/prometheus/prometheus/promql.(*Engine).exec(0xc020b422a0, {0x1ff7c30, 0xc020b48ab0}, 0xc000899020)
11:33:57 store1: /home/bwplotka/Repos/thanosgopath/pkg/mod/github.com/prometheus/prometheus@v1.8.2-0.20211119115433-692a54649ed7/promql/engine.go:548 +0x7eb
11:33:57 store1: github.com/prometheus/prometheus/promql.(*query).Exec(0xc000899020, {0x1ff7c30, 0xc01b1eefc0})
11:33:57 store1: /home/bwplotka/Repos/thanosgopath/pkg/mod/github.com/prometheus/prometheus@v1.8.2-0.20211119115433-692a54649ed7/promql/engine.go:186 +0xb4
11:33:57 store1: github.com/thanos-io/thanos/pkg/store/promqlaggr.AggregatedSeriesSet({0x1ff7c30, 0xc01b1eefc0}, {0x1fdea00, 0xc01a4fbef0}, 0xc01b1eeba0, 0x0, {0xc00b97d900, 0x2, 0xc01b123fd0})
11:33:57 store1: /home/bwplotka/Repos/thanos/pkg/store/promqlaggr/aggregator.go:77 +0x342
11:33:57 store1: github.com/thanos-io/thanos/pkg/store.(*BucketStore).Series.func2()
11:33:57 store1: /home/bwplotka/Repos/thanos/pkg/store/bucket.go:1082 +0x4fa
11:33:57 store1: golang.org/x/sync/errgroup.(*Group).Go.func1()
11:33:57 store1: /home/bwplotka/Repos/thanosgopath/pkg/mod/golang.org/x/sync@v0.0.0-20210220032951-036812b2e83c/errgroup/errgroup.go:57 +0x67
11:33:57 store1: created by golang.org/x/sync/errgroup.(*Group).Go
11:33:57 store1: /home/bwplotka/Repos/thanosgopath/pkg/mod/golang.org/x/sync@v0.0.0-20210220032951-036812b2e83c/errgroup/errgroup.go:54 +0x92
11:33:57 store-prom2: pushed query on block
11:33:57 store-prom2: : sum(count_over_time(name{}[1209600000ms])) by ( cluster, replica )
11:33:57 store-prom2: panic:
11:33:57 store-prom2: runtime error: invalid memory address or nil pointer dereference
11:33:57 store-prom2: [signal SIGSEGV: segmentation violation code=
11:33:57 store-prom2: 0x1
11:33:57 store-prom2: addr=0x18 pc=0x1193449]
11:33:57 store-prom2: goroutine 4570 [running]:
11:33:57 store-prom2: github.com/thanos-io/thanos/pkg/store/promqlaggr.(*seriesQueryable).Close
11:33:57 store-prom2: (0xc0456f6dc0)

```

bwplotka:
Current status:

Without pushdown the query `sum(count_over_time(continuous_app_metric0[2w])) by (cluster, replica)` on my data (20k series full of samples for 2w = 1.6 billions of samples) .. [after 2.7m timeout] (no dedup).

With QP enabled, same query on same setup takes only ~20s! (no dedup). Previously my runs of no pushdown were returning within 1.7m (my laptop doing more work now?), but in this case the QP case would be probably faster too. We can assume more than `~6x` time improvement with this very naive implementation.

image

Resource usage is better too. Total CPU time is only slightly bigger, within 2 CPU (more concurrency), memory is enormously better. Not much of CPU increase which indicates more concurreny is possible. Total memory 2x less, BUT distributed across nodes (QP, nonQP, nonQP, QP runs):

image

image

image

image

GiedriusS (commented):
Do you want to do this in a feature branch? Is it ready for a review? :p 

I am convinced that wrapping everything into the current `SeriesResponse` is the way to go because with pushdown, we will probably never have "100% correct responses" either way. For example, when executing `rate(foo[5m])` you won't know whether there was any missing data in `foo[5m]` because `rate` extrapolates. So, we will lose some accuracy either way. But, I think that's a reasonable trade-off to make. I would vote to make this feature opt-out, not opt-in.

bwplotka:
> Do you want to do this in a feature branch? Is it ready for a review? :p

I think not as it is now. It's a demo what can be done. If we want this even in this naitve form we need to explore:

* Adding button for turning this off and on
* Passing more hints to PromQL
* Creating extensive unit tests
* Making sure this WON'T be used with duplicated or overlapped data

So - lots of work still.

>  I am convinced that wrapping everything into the current SeriesResponse is the way to go because with pushdown, we will probably never have "100% correct responses" either way. For example, when executing rate(foo[5m]) you won't know whether there was any missing data in foo[5m] because rate extrapolates. So, we will lose some accuracy either way. But, I think that's a reasonable trade-off to make. I would vote to make this feature opt-out, not opt-in.

I disagree with that. I would aim with 100% correctness. We can totally work with `rate` boundaries by returning raw samples near the boundaries of shared time range, plus we can shard and query pushdown to disjoint series (!) where no rate can extrapolate outside of matched time range (:

pracucci:
Good job! The PoC makes sense to me. I have few very high level comments:
- At the current stage, it requires a guarantee that a series is only within a single block for the given time range, so we can't enable it if compactor is lagging behind (or enable it for series queried from receiver).
- I think you may have problems (in terms of correctness) with NaN and StaleNaN and the PromQL lookback period when they are around the block range period boundaries.
- Design wise, I'm not super convinced about running PromQL engine in the store. My concern is just a architectural thing.
- Have you considered pushing down the partial query instead of hints?

bwplotka:
Thanks for the comments @pracucci! 

> At the current stage, it requires a guarantee that a series is only within a single block for the given time range, so we can't enable it if compactor is lagging behind (or enable it for series queried from receiver). Looks quite a strong guarantee to me. Most queries hit most recent data (and can be slow because querying hundred thousands of series).

Correct, see an initial version that in theory does not need dedup/non-overlapped series: https://github.com/thanos-io/thanos/pull/4917 This is for trivial stuff like `min/max` etc.

In terms of correctness around edges and duplication for other functions, we have to find some solutions indeed. In Thanos world, it might be safe to pushdown to Prometheus and make sure data not overlaps e.g time-wise with stores. For HA this get's tricker, but maybe there is a way to do deduplication in some dynamic fashion - instead of doing this per series, assume full time range from one Prometheus is not queryable during downtime period.

For Receivers and Cortex ingesters, it's might be tricker indeed because we shard and replicate by individual series. We could still try to group receivers/ingesters in some way or pushdown the calculation while keeping `replica` label. Something to think about.
 
> I think you may have problems (in terms of correctness) with NaN and StaleNaN and the PromQL lookback period when they are around the block range period boundaries.

True, this has to be handled correctly, similar to `rate` algorithm discussed in https://github.com/thanos-io/thanos/pull/4901#issuecomment-981835385 (: 
 
> Design wise, I'm not super convinced about running PromQL engine in the store. My concern is just a architectural thing.

Could you elaborate? Do you think it would need to be too generic? I like the idea of reusing same, tricky piece, but I might be wrong.

> Have you considered pushing down the partial query instead of hints?

Yes and I don't see a difference. You can construct query from hints and vice versa. So is it worth to keep preparing? hints does not require that parsing 🤔 But I don't think it makes any difference if we choose partial PromQL string vs hints. Do you have any benefit I might have missed?

bwplotka:
NOTE: https://github.com/thanos-io/thanos/pull/4917 Filips started version which could be merged.

This PR is only for overview and will be closed as soon as discussion die out and PoC will be finished (:

PR #4903, opened 2021-11-25
Compactor: Add tracing support

### LINKED ISSUE #4832: Add tracing support for compaction

We have some metrics in the compactor of its overall status. But right now it is hard to track the status and time taken for each compaction run.

Add tracing support for compactor and instrument different steps in compaction. For example:
1. Block planning
2. Block download
3. Block verification
4. Block compaction
5. Block upload

metonymic-smokey:
Fixes #4832

## Changes
Added tracing support for the various compaction steps(similar to Querier). 

## Verification
Tested locally using Jaeger.


pkg/compact/compact.go
```diff
@@ -1008,12 +1016,19 @@ func (cg *Group) compact(ctx context.Context, dir string, planner Planner, comp
...
+		}, opentracing.Tags{"block.id": meta.ULID})
+
+		if err != nil {
 			return false, ulid.ULID{}, retry(errors.Wrapf(err, "download block %s", meta.ULID))
 		}
 
 		// Ensure all input blocks are valid.
-		stats, err := block.GatherIndexHealthStats(cg.logger, filepath.Join(bdir, block.IndexFilename), meta.MinTime, meta.MaxTime)
+		var stats block.HealthStats
+		tracing.DoInSpan(ctx, "compaction_block_healthcheck", func(ctx context.Context) {
```
matej-g, line 1029: Nit: Maybe `_healthstats` since the method does not check, only collects stats?
```suggestion
		tracing.DoInSpan(ctx, "compaction_block_healthstats", func(ctx context.Context) {
```

matej-g (commented):
This is looking cool! I was wondering about one more thing, which I think could be particularly useful in the context of compaction and that is to also record the error in the span (we would have the error in the logs, but I think having them directly on span would be more handy). 

We are using our convenience method `DoInSpan` here, but that does not capture an error if it occurs. Maybe we could have another method (e.g. `DoInSpanWithErr`) which could also set the error on span. Does that sound good?

pkg/compact/compact.go
metonymic-smokey: Logging errors sounds good! I had tried it one of the earlier commits but reverted considering errors were already being logged. 
Instead of a custom function, there is an existing function, `span.LogFields(opentracing_log.Error(err))`. Maybe I could use that?

matej-g: I think the issue is that if we use our "custom" method `DoInSpan`, we do not have access to the actual span. Hence why I suggested a kind of enhanced version of the same method. Other option would be to not use `DoInSpan` and instrument the code "manually".

metonymic-smokey: Thanks.
There is a way to get the span from the current context, which I have used in my recent commit. That way, I don't have to instrument manually and can log errors with minimal changes.

metonymic-smokey: I have pushed a commit with an implementation of the former. Please let me know what you think.
thanks!

pkg/tracing/tracing.go
```diff
@@ -72,6 +73,16 @@ func StartSpan(ctx context.Context, operationName string, opts ...opentracing.St
...
 }
 
+// DoInSpanWtihErr executes function doFn inside new span with `operationName` name and hooking as child to a span found within given context if any.
+// It uses opentracing.Tracer propagated in context. If no found, it uses noop tracer notification.
+// It logs the error inside the new span created, which differentiates it from DoInSpan and DoWithSpan.
+func DoInSpanWithErr(ctx context.Context, operationName string, doFn func(context.Context) error, opts ...opentracing.StartSpanOption) {
+	span, newCtx := StartSpan(ctx, operationName, opts...)
+	defer span.Finish()
+	err := doFn(newCtx)
+	span.LogFields(opentracing_log.Error(err))
```
matej-g, line 83: 1) I'm not sure about the behavior if `err` is `nil` (whether it still logs error), but maybe we could check `err != nil` before we log error?
2) I think we can replace this with method from the `ext` package, see https://pkg.go.dev/github.com/opentracing/opentracing-go@v1.2.0/ext#LogError. This will set all necessary things on the span for us

matej-g (commented):
Great job @metonymic-smokey! Thanks for going above and beyond to address things. I'm wondering about one last thing, afterwards I believe this is ready.

Let's try to get a review from maintainer as well, perhaps @yeya24 as original issue reporter.

pkg/tracing/tracing.go
metonymic-smokey: Good point @matej-g. Thanks!
I've made the changes in a recent commit. Let me know if they're okay.

pkg/tracing/tracing.go
```diff
@@ -72,6 +74,18 @@ func StartSpan(ctx context.Context, operationName string, opts ...opentracing.St
...
 
+// DoInSpanWtihErr executes function doFn inside new span with `operationName` name and hooking as child to a span found within given context if any.
+// It uses opentracing.Tracer propagated in context. If no found, it uses noop tracer notification.
+// It logs the error inside the new span created, which differentiates it from DoInSpan and DoWithSpan.
+func DoInSpanWithErr(ctx context.Context, operationName string, doFn func(context.Context) error, opts ...opentracing.StartSpanOption) {
+	span, newCtx := StartSpan(ctx, operationName, opts...)
+	defer span.Finish()
+	err := doFn(newCtx)
+	if err != nil {
+		ext.LogError(span, err, opentracing_log.Error(err))
```
matej-g, line 85: I believe we do not need to pass the last value, since the method will already include the field for error.

matej-g (approved):
:100:

pkg/compact/compact.go
```diff
@@ -1008,12 +1018,21 @@ func (cg *Group) compact(ctx context.Context, dir string, planner Planner, comp
 			uniqueSources[s] = struct{}{}
 		}
 
-		if err := block.Download(ctx, cg.logger, cg.bkt, meta.ULID, bdir); err != nil {
+		tracing.DoInSpanWithErr(ctx, "compaction_block_download", func(ctx context.Context) error {
+			err = block.Download(ctx, cg.logger, cg.bkt, meta.ULID, bdir)
+			return err
+		}, opentracing.Tags{"block.id": meta.ULID})
+
```
squat, line 1025: tiny nit
```suggestion
```

pkg/compact/compact.go
```diff
@@ -1008,12 +1018,21 @@ func (cg *Group) compact(ctx context.Context, dir string, planner Planner, comp
...
+		if err != nil {
 			return false, ulid.ULID{}, retry(errors.Wrapf(err, "download block %s", meta.ULID))
 		}
 
 		// Ensure all input blocks are valid.
-		stats, err := block.GatherIndexHealthStats(cg.logger, filepath.Join(bdir, block.IndexFilename), meta.MinTime, meta.MaxTime)
+		var stats block.HealthStats
+		tracing.DoInSpanWithErr(ctx, "compaction_block_healthstats", func(ctx context.Context) error {
```
squat, line 1032: Since the Golang name is `...HealthStats`, `health` and `stats` should be treated as separate words
```suggestion
		tracing.DoInSpanWithErr(ctx, "compaction_block_health_stats", func(ctx context.Context) error {
```

pkg/compact/compact.go
```diff
@@ -764,7 +766,11 @@ func (cg *Group) Compact(ctx context.Context, dir string, planner Planner, comp
 		return false, ulid.ULID{}, errors.Wrap(err, "create compaction group dir")
 	}
 
-	shouldRerun, compID, err := cg.compact(ctx, subDir, planner, comp)
+	var err error
+	tracing.DoInSpanWithErr(ctx, "group_compaction", func(ctx context.Context) error {
```
squat, line 770: All other spans are named `compaction_<x>`, should we make this follow the same convention?

squat (approved):
Looks good overall! Some very small nits

squat (approved):

squat:
lgtm! restarting failed jobs and we can merge :))

metonymic-smokey:
Thanks for the reviews @matej-g @squat !

PR #4908, opened 2021-11-26
UI: Unify behavior for time ranges when min / max time is not available / applicable

matej-g:

## Changes
Components which expose store API behave somewhat differently when it comes to returning store time range information, where for example:
- If no stores are available, receiver will return [`math.MinInt64`](https://github.com/thanos-io/thanos/blob/main/pkg/store/multitsdb.go#L121) or [empty response, i.e. `0`](https://github.com/thanos-io/thanos/blob/main/pkg/store/multitsdb.go#L68)
- By default the [sidecar returns `0000-01-01T00:00:00Z`](https://github.com/thanos-io/thanos/blob/main/cmd/thanos/sidecar.go#L491)

This means that on the actual UI, we can get various results, ranging from `Invalid date` to `0000-01-01 00:00:00` - which all signalize the same thing: the information is not available or not known.

My suggestion here is to:
1) For UI purposes, consider valid dates to be in range `1` - `(math.MaxInt64 - 1)`
2) Use the 'infinity' icon for min time instead of the invalid or misleading value (this is already used for max time)

If people find the icon confusing, possible alternative would be to use a different icon or some default, minimum time, like 0 UNIX timestamp.

## Verification

UI tests have been adjusted. I also did few manual checks with the interactive tests.


GiedriusS:
IMO this is a step in the right direction but we also get a lot of questions about what the infinity symbols mean so I'd like to fix it in one go :/ perhaps it would be easy to add a tooltip for those icons saying something like "This node either has no information or is unable to synchronize its state"? Maybe not even on the icons but on the address itself? Something akin to https://stackoverflow.com/questions/7117073/add-a-tooltip-to-a-div. Could we add this information in the `title` of addresses?

matej-g:
> Maybe we could add this not even on the icons but on the address itself? Something akin to https://stackoverflow.com/questions/7117073/add-a-tooltip-to-a-div. Could we add this information in the `title` of addresses?

I like this idea a lot, less confusing than just having icons. I think we could distinguish all cases, i.e. if no min and max time, we can provide information that the time range is unknown; if only one of the ranges is available, we can just say that the store has no max or min time limit. I'm not much of a front-end developer :sweat_smile:, but I'll try to come back with a solution.

matej-g:
@GiedriusS I made few more changes, as suggested:
- I changed the 'infinity' icon into a simple 'minus' icon
- Added tooltip on store's address and min / max time cell (see screenshot)
- Specified the times are in UTC
tooltip

GiedriusS (commented):
Could you please rebase on the newest main?

pkg/ui/react-app/src/thanos/pages/stores/StorePoolPanel.tsx
```diff
@@ -16,13 +16,25 @@ export const columns = [
...
-  'Max Time',
+  'Min Time (UTC)',
+  'Max Time (UTC)',
   'Last Successful Health Check',
   'Last Message',
 ];
 
-export const MAX_TIME = 9223372036854775807;
+export const storeTimeRangeMsg = (validMin: boolean, validMax: boolean): string => {
+  if (!validMin && !validMax) {
```
GiedriusS, line 26: I think it is important to distinguish between the cases:

* Time range is `MIN_TIME; MAX_TIME` i.e. when Prometheus still hasn't uploaded any blocks. In this case, Query will always send all queries to that StoreAPI; 
* Time range is `MAX_TIME; MIN_TIME` i.e. when Thanos Store has no blocks. In this case, Query will not send any queries to it.

As is right now, both stores would show up with minuses in the Store page and with the same tooltip. I suggest at the minimum to change this function to distinguish between these two states so that it would be clear what's happening. Perhaps something like:
"This store's time range data is not available but we will send all queries to it"? Or maybe we could make the minuses different colors depending on the state? For example, the minuses could be green if we will still send all queries to it. Or perhaps the node's address could have a different color depending on the state e.g. green = we will send all queries to it, yellow = we will send certain queries to it depending on the labels & min/max time, red = no queries will be sent to it?

WDYT @matej-g ?

GiedriusS (commented):
Looks good, just one suggestion.

pkg/ui/react-app/src/thanos/pages/stores/StorePoolPanel.tsx
matej-g: Hm, I'm wondering whether min / max time ranges and the information whether we are querying the store are two separate things. For the min / max, I would expect simply to have the information stated in the dashboard if it is available - if not, regardless of the reason (i.e. no block uploaded or no block in store), we just signal that this information cannot be showed because it is unknown.

But I think you're onto something with color coding the nodes (perhaps we could even add another column, besides `Status` and add details on querying?), although it feels like a separate feature.

matej-g: ping @GiedriusS, how do you suggest to proceed? Do you agree with the suggestion and if so, should we do this in the current PR?

GiedriusS: Mhm, I guess you are correct. Whether or not we are sending queries to a node is different information altogether and we could work on this in future pull requests. :+1:

GiedriusS (approved):

GiedriusS:
Could you please merge main into your branch?

matej-g:
Done, thanks for the review! I also opened a separate issue for that color-coding idea.

GiedriusS (approved):

PR #4917, opened 2021-11-30
Implement query pushdown for a subset of aggregations

fpetkovski:
Certain aggregations can be executed safely on leaf nodes without
worrying about data duplication or overlap. One such example is the max
function which can be computed on local data by the leaves before it is
computed globally by the querier.

This commit implements local aggregation in the Prometheus sidecar for
all functions which are safe to execute locally. The feature can be enabled by 
passing the `--enable-feature evaluate-queries` option to the sidecar.


pkg/store/prometheus.go
```diff
@@ -179,6 +179,11 @@ func (p *PrometheusStore) Series(r *storepb.SeriesRequest, s storepb.Store_Serie
 		return nil
 	}
 
+	if r.QueryHints.IsFuncDistributive() {
```
bwplotka, line 182: 👍🏽

pkg/store/prometheus.go
```diff
@@ -220,6 +225,55 @@ func (p *PrometheusStore) Series(r *storepb.SeriesRequest, s storepb.Store_Serie
...
 
+func (p *PrometheusStore) queryPrometheus(s storepb.Store_SeriesServer, r *storepb.SeriesRequest) error {
+	ctx := context.Background()
+	queryOpts := promclient.QueryOptions{
+		Deduplicate:             false,
+		PartialResponseStrategy: storepb.PartialResponseStrategy_WARN,
+		Method:                  "GET",
+	}
+
+	matrix, _, err := p.client.QueryRange(ctx, p.base, r.ToPromQL(), r.MinTime, r.MaxTime, r.QueryHints.Step/1000, queryOpts)
```
bwplotka, line 236: Is it always range query?

pkg/store/storepb/custom.go
```diff
@@ -517,3 +517,34 @@ func (c *SeriesStatsCounter) Count(series *Series) {
...
+	return fmt.Sprintf("%s %s (%s) (%s)", m.QueryHints.Func, m.QueryHints.ByOrWithout(), grouping, matchers)
+}
+
+func (m *QueryHints) IsFuncDistributive() bool {
+	distributiveFuncs := []string{
+		"max",
+		"min",
+		"group",
+		"topk",
+		"bottomk",
```
bwplotka, line 533: Are we sure, those are distributive?

pkg/store/storepb/custom.go
```diff
@@ -517,3 +517,34 @@ func (c *SeriesStatsCounter) Count(series *Series) {
...
 	}
 }
+
+func (m *SeriesRequest) ToPromQL() string {
+	grouping := strings.Join(m.QueryHints.Grouping, ",")
+	matchers := MatchersToString(m.Matchers...)
+	return fmt.Sprintf("%s %s (%s) (%s)", m.QueryHints.Func, m.QueryHints.ByOrWithout(), grouping, matchers)
+}
+
+func (m *QueryHints) IsFuncDistributive() bool {
```
bwplotka, line 527: Can we make sure to add comment what do you mean distributive as it might be non trivial

pkg/store/storepb/rpc.proto
```diff
@@ -108,10 +108,33 @@ message SeriesRequest {
...
+
+  // query_hints are the hints coming from the PromQL engine when
+  // requesting a storage.SeriesSet for a given expression.
+  QueryHints query_hints = 12 [(gogoproto.nullable) = false];
+}
+
+// Analogous to storage.SelectHints.
+message QueryHints {
+  // Query step size in milliseconds.
+  int64 step = 1;
```
bwplotka, line 126:
```suggestion
  int64 step_milis = 1;
```

bwplotka (commented):
Nice, good start! Looks generally good, but the topk/bottomk I don't think can be pushdown-ed easily 🤔 Essentially the top from node X might be not the top from part Y.

bwplotka:
cc @GiedriusS what if we start with this slowly into main?

pkg/store/storepb/custom.go
fpetkovski: I think they are because the querier will still execute a bottomk on the resultset. In other words
```
bottomk(everything) = bottomk( bottomk(node a), bottomk(node b) )
```
I just don't know if deduplication can affect the result in some way. If the querier sees a union of all nodes, then it will do the right thing.

fpetkovski:
We could also add a flag to the sidecar to toggle this feature on and off. Something like a feature flag.

pkg/store/prometheus.go
fpetkovski: I think it needs to be because of lookback, but I can double check.

pkg/store/prometheus.go
```diff
@@ -220,6 +225,55 @@ func (p *PrometheusStore) Series(r *storepb.SeriesRequest, s storepb.Store_Serie
 	return p.handleStreamedPrometheusResponse(s, httpResp, queryPrometheusSpan, extLset)
 }
 
+func (p *PrometheusStore) queryPrometheus(s storepb.Store_SeriesServer, r *storepb.SeriesRequest) error {
+	ctx := context.Background()
```
GiedriusS, line 229: Let's use `s.Context()` here?

pkg/store/prometheus.go
```diff
@@ -220,6 +225,55 @@ func (p *PrometheusStore) Series(r *storepb.SeriesRequest, s storepb.Store_Serie
...
+	queryOpts := promclient.QueryOptions{
+		Deduplicate:             false,
+		PartialResponseStrategy: storepb.PartialResponseStrategy_WARN,
+		Method:                  "GET",
```
GiedriusS, line 233: I think this doesn't do anything - AddTo() does nothing and we always use GET:
```
	body, _, err := c.req2xx(ctx, &u, http.MethodGet)
	if err != nil {
		return nil, nil, errors.Wrap(err, "read query range response")
	}
```

pkg/store/prometheus.go
```diff
@@ -220,6 +225,55 @@ func (p *PrometheusStore) Series(r *storepb.SeriesRequest, s storepb.Store_Serie
 	return p.handleStreamedPrometheusResponse(s, httpResp, queryPrometheusSpan, extLset)
 }
 
+func (p *PrometheusStore) queryPrometheus(s storepb.Store_SeriesServer, r *storepb.SeriesRequest) error {
+	ctx := context.Background()
+	queryOpts := promclient.QueryOptions{
```
GiedriusS, line 230: I think in general these options don't do anything because here we are querying Prometheus and these options are only understandable by Thanos Query :smile:

pkg/store/prometheus.go
GiedriusS, line 236: Wouldn't it be possible to pass down the same hints via remote read and read streamed responses that way: https://github.com/prometheus/prometheus/blob/39d79c3cfb86c47d6bc06a9e9317af582f1833bb/prompb/remote.proto#L30-L70? :thinking:

fpetkovski: Interesting, I didn't think about this. Would remote read be able to apply the aggregations from the hints?

GiedriusS: Prometheus passes down the hints to Select() so I think it should work: https://github.com/prometheus/prometheus/blob/main/storage/remote/read_handler.go#L141-L155. Perhaps worth adding a test for testing at least one of these functions that it returns the same result as in: https://github.com/thanos-io/thanos/pull/4712/files#diff-3a68fda5485fac918834a9a7d2c3522f859de6e21f0b41ff2031eebb2579eed0? @bwplotka perhaps I am wrong, feel free to correct me

bwplotka: So hints do not do anything right now if you use them with remote read (: 

We have multiple options:

1. Add PromQL pass for remote read hints
2. Optimize Query API to stream series to us

All need to be done on the Prom side to leverage that. I think in mean time we could do it naively using Query API - that's the closest to what we want (: 

I think remote read + hints that pre-aggregates would be nice if community will accept it. I can try to start putting issues about that. WDYT?

pkg/store/prometheus.go
fpetkovski: Good catch, updated

fpetkovski:
I think we might want to add some tests as well to make sure we don't introduce a regression. I'll try to do that by EOW

pkg/store/prometheus.go
GiedriusS: :+1: sounds good, so let's start with a JSON-based implementation. (1) seems better, it should be relatively easy to take the provided hints into account while executing? :thinking:  :smile: anyway, but that's for the future. But let's add a feature flag, right?

pkg/store/storepb/custom.go
bwplotka: I think that's not true. I wish it was ;p 

All even without `@` modified - there is also this modifier which makes things even more tricky.
Imagine this case:

topk(1, sum(metric1) by (label1)))

```
Node A:

metric1{label1="1", ...} 10
metric1{label1="2", ...} 9

Node B

metric1{label1="1", ...} 1
metric1{label1="2", ...} 9
```

Overall answer is `metric1{label1="2"} 18`, but if we push down we have `metric1{label1="1"} 10` from node A and  `metric1{label1="2"} 9` from node B. when we do aggregated topk 1 it will give `metric1{label1="1"} 10` which is overall incorrect answer.

bwplotka: so I think we can pushdown topk and bottomk but only in certain cases

bwplotka: As @fpetkovski mentioned offline - it's fine now as we push only first layer of hints

GiedriusS (commented):
Could we reuse the same `--enable-feature` pattern for this instead of a separate flag?

fpetkovski:
@GiedriusS @bwplotka this is now ready for review, could you ptal :)

CHANGELOG.md
```diff
@@ -9,6 +9,7 @@ NOTE: As semantic versioning states all 0.y.z releases can contain breaking chan
 We use *breaking :warning:* to mark changes that are not backward compatible (relates only to v0.y.z releases.)
 
 ## Unreleased
+- [#4917](https://github.com/thanos-io/thanos/pull/4917) Sidecar: Add flag `--evaluate-queries` to Thanos sidecar to enable local execution of certain queries.
```
GiedriusS, line 12: You forgot to change this part after changing to `--enable-feature`? (:

pkg/store/prometheus.go
```diff
@@ -220,18 +230,66 @@ func (p *PrometheusStore) Series(r *storepb.SeriesRequest, s storepb.Store_Serie
 	return p.handleStreamedPrometheusResponse(s, httpResp, queryPrometheusSpan, extLset)
 }
 
-func (p *PrometheusStore) handleSampledPrometheusResponse(s storepb.Store_SeriesServer, httpResp *http.Response, querySpan tracing.Span, extLset labels.Labels) error {
-	ctx := s.Context()
+func (p *PrometheusStore) queryPrometheus(s storepb.Store_SeriesServer, r *storepb.SeriesRequest) error {
+	step := math.Max(float64(r.QueryHints.StepMillis/1000), 30)
+	minTime := math.Max(float64(r.MinTime), float64(r.MaxTime-5*time.Minute.Milliseconds()))
```
GiedriusS, line 235: I feel like this part needs an explanation in the comments, it's unclear from reading the code what is happening.

I assume 5 minutes come from the default PromQL lookback interval? Do we need to adjust this interval depending on the original PromQL engine's lookback interval? 

Also, shouldn't this be `math.Min` because this right now effectively limits each query to `[MaxTime-5m; MaxTime]` if `MaxTime-MinTime > 5m`? :thinking:

pkg/store/prometheus.go
```diff
@@ -220,18 +230,66 @@ func (p *PrometheusStore) Series(r *storepb.SeriesRequest, s storepb.Store_Serie
...
+	opts := promclient.QueryOptions{}
+	matrix, _, err := p.client.QueryRange(s.Context(), p.base, r.ToPromQL(), int64(minTime), r.MaxTime, int64(step), opts)
+	if err != nil {
+		return err
+	}
+
+	for _, vector := range matrix {
+		var lbls []labels.Label
```
GiedriusS, line 243: You could potentially make this a bit faster by preallocating:

```go
externalLbls := p.externalLabelsFn()
lbls := make([]labels.Label, 0, len(externalLbls)+len(vector.Metric))
...
```

GiedriusS (commented):
Good work, just some comments on the lookback interval :thinking:

pkg/store/prometheus.go
fpetkovski: Thanks, updated

CHANGELOG.md
fpetkovski: Ah yes, updated now, thanks :)

pkg/store/prometheus.go
fpetkovski: It looks like we can always use a query range, we just have to make sure the step is set to a positive value when the query  coming from the Querier is an instant. This simplifies the logic a lot and we don't have to do obscure modifications to the query parameters.

pkg/store/prometheus.go
fpetkovski: Looks like we can always use a query range, we just need to use a positive step for instant queries.

pkg/store/prometheus.go
fpetkovski: Looks like we have this report now: https://github.com/thanos-io/thanos/pull/4917/files#annotation_2368016155

Not sure if that's important thought.

wiardvanrij:
I think this would benefit having some technical documentation. For example a small explanation of 'what is the fanout', maybe a table of what functions would work and which don't. I noticed a lot knowledge present that sometimes "you just have to know" - it would be nice to have this somewhere written down.

pkg/api/query/v1.go
```diff
@@ -92,6 +92,7 @@ type QueryAPI struct {
 	enableTargetPartialResponse         bool
 	enableMetricMetadataPartialResponse bool
 	enableExemplarPartialResponse       bool
+	enableQueryPushdown                 bool
```
bwplotka, line 95: You did it, nice!

pkg/store/prometheus.go
```diff
@@ -220,18 +226,78 @@ func (p *PrometheusStore) Series(r *storepb.SeriesRequest, s storepb.Store_Serie
...
+			return err
+		}
+		matrix = result
+	} else {
+		vector, _, err := p.client.QueryInstant(s.Context(), p.base, r.ToPromQL(), time.Unix(r.MaxTime/1000, 0), opts)
+		if err != nil {
+			return err
+		}
+
+		matrix = make(model.Matrix, 0, len(vector))
```
bwplotka, line 246: TODO to myself: Add issue about remote read understanding aggr or streamed PromQL

pkg/store/prometheus.go
```diff
@@ -220,18 +226,78 @@ func (p *PrometheusStore) Series(r *storepb.SeriesRequest, s storepb.Store_Serie
...
+	}
+
+	externalLbls := p.externalLabelsFn()
+	for _, vector := range matrix {
+		lbls := make([]labels.Label, 0, len(externalLbls)+len(vector.Metric))
+		// Attach labels from samples
+		for k, v := range vector.Metric {
+			lbls = append(lbls, labels.FromStrings(string(k), string(v))...)
+		}
+		// Attach external labels for compatibility with remote read
```
bwplotka, line 267:
```suggestion
		// Attach external labels for compatibility with remote read.
```

pkg/store/storepb/query_hints.go
```diff
@@ -0,0 +1,48 @@
...
+	"strings"
+)
+
+func (m *QueryHints) toPromQL(labelMatchers []LabelMatcher) string {
+	grouping := m.Grouping.toPromQL()
+	matchers := MatchersToString(labelMatchers...)
+	queryRange := m.Range.toPromQL()
+
+	query := fmt.Sprintf("%s %s (%s%s)", m.Func.Name, grouping, matchers, queryRange)
+	// Remove double spaces if some expressions are missing
```
bwplotka, line 17:
```suggestion
	// Remove double spaces if some expressions are missing.
```

pkg/store/storepb/rpc.proto
```diff
@@ -108,18 +108,60 @@ message SeriesRequest {
...
+  // Deprecated: Use query_hints instead.
   int64 step = 10;
 
   // Range vector selector range in milliseconds.
+  // Deprecated: Use query_hints instead.
   int64 range = 11;
+
+  // Indicates whether the store should evaluate queries
+  // before sending back samples.
+  bool enableQueryPushdown = 12;
```
bwplotka, line 120: Maybe stupid idea, but why not controling if query pushdown should be send or not by either sending query hints or not sending them?

bwplotka: Less params == simpler

test/e2e/e2ethanos/services.go
```diff
@@ -144,6 +144,9 @@ func NewPrometheusWithSidecarCustomImage(e e2e.Environment, name, promConfig, we
 		"--tsdb.path":         dataDir,
 		"--log.level":         infoLogLevel,
 	}
+	if sidecarEvaluateQueries {
```
bwplotka, line 147: No need for this anymore?

bwplotka (dismissed):
I love this work - solid PR. LGTM, just small nits. Also we are missing, docs but we can add them in separate PR!

cc @GiedriusS @squat @brancz

GiedriusS (dismissed):
Played around with this locally - looks good modulo @bwplotka's comments. I'm not sure how I dismissed Bartek's review with a commit that I didn't even author or push :joy:  Let's play with it behind a feature flag and see how we can further improve upon this base :+1:

pkg/store/prometheus.go
```diff
@@ -220,18 +226,78 @@ func (p *PrometheusStore) Series(r *storepb.SeriesRequest, s storepb.Store_Serie
...
+	opts := promclient.QueryOptions{}
+	step := r.QueryHints.StepMillis / 1000
+	if step != 0 {
+		result, _, err := p.client.QueryRange(s.Context(), p.base, r.ToPromQL(), r.MinTime, r.MaxTime, step, opts)
+		if err != nil {
+			return err
+		}
+		matrix = result
+	} else {
+		vector, _, err := p.client.QueryInstant(s.Context(), p.base, r.ToPromQL(), time.Unix(r.MaxTime/1000, 0), opts)
```
GiedriusS, line 241:
```suggestion
		vector, _, err := p.client.QueryInstant(s.Context(), p.base, r.ToPromQL(), timestamp.Time(r.MaxTime), opts)
```

GiedriusS:
@fpetkovski any update on this? I'd like to merge this as a good base so that we could continue further push down related work.

fpetkovski:
Ah sorry I saw it was approved and didn't see the comments that followed up. I'll address them today so that we can merge it :+1:

fpetkovski:
@GiedriusS all comments should be addressed

pkg/query/querier.go
```diff
@@ -268,12 +286,17 @@ func (q *querier) selectFn(ctx context.Context, hints *storage.SelectHints, ms .
...
+	if q.enableQueryPushdown {
+		queryHints = storeHintsFromPromHints(hints)
+	}
 	if err := q.proxy.Series(&storepb.SeriesRequest{
 		MinTime:                 hints.Start,
 		MaxTime:                 hints.End,
 		Matchers:                sms,
 		MaxResolutionWindow:     q.maxResolutionMillis,
 		Aggregates:              aggrs,
+		QueryHints:              &queryHints,
```
GiedriusS, line 299: But now `QueryHints` will always be not nil, right? I think there's a bug now here. It should be something like:

```go
var queryHints *storepb.QueryHints
if q.enableQueryPushdown {
  generatedHints := storeHintsFromPromHints(hints)
  queryHints = generatedHints
}
```

@fpetkovski :thinking:

fpetkovski: Yes, you're right, good catch. I applied your suggestion

GiedriusS (approved):

GiedriusS (approved):

GiedriusS (approved):

bwplotka:
💪🏽

pkg/store/prometheus.go
bwplotka: https://github.com/prometheus/prometheus/issues/10040

pkg/store/storepb/custom.go
```diff
@@ -517,3 +517,27 @@ func (c *SeriesStatsCounter) Count(series *Series) {
...
+// IsSafeToExecute returns true if the function or aggregation from the query hint
+// can be safely executed by the underlying Prometheus instance without affecting the
+// result of the query.
+func (m *QueryHints) IsSafeToExecute() bool {
+	distributiveOperations := []string{
+		"max",
+		"max_over_time",
+		"min",
+		"min_over_time",
+		"group",
```
juliusv, line 534: Why not `count()` btw.? I guess some others would also work, but I guess you're just starting with a few...

pkg/store/storepb/custom.go
juliusv: @bwplotka Wouldn't the same be true for `max()` as well, since it's mostly equivalent to `topk(1, ...)`? Both would only work if you either don't aggregate away the external labels, or if you first push down the inner `sum()` and feed its merged result to a `topk()` / `max()`. And then the same is true for `bottomk()` / `min()`...

pkg/store/storepb/custom.go
GiedriusS: Because then PromQL engine goes ahead and calculates count() again on the already correct results i.e. pushed down data thus corrupting it. I have suggested [something here](https://github.com/prometheus/prometheus/issues/10101) but it got no attention so I haven't pursued it further.

juliusv: Ah right, I had assumed there was already some facility for changing the outer operation (in the `count` case, to `sum`), but that's not required for min/max/group.

The comment on https://github.com/prometheus/prometheus/issues/10101 is interesting though. Indeed, if the same series can exist on two leafs (I think this is only the case for HA deduplication in Thanos, right?), then pushing `rate()` or even `count()` down in a correct way would be impossible (if you do the count across two HA replicas and you get 10 on each, you don't know if you really counted the same 10 series both times or counted different ones, or some overlapping sets).

pkg/store/storepb/custom.go
```diff
@@ -517,3 +517,27 @@ func (c *SeriesStatsCounter) Count(series *Series) {
...
+
+// IsSafeToExecute returns true if the function or aggregation from the query hint
+// can be safely executed by the underlying Prometheus instance without affecting the
+// result of the query.
+func (m *QueryHints) IsSafeToExecute() bool {
+	distributiveOperations := []string{
+		"max",
+		"max_over_time",
+		"min",
+		"min_over_time",
```
yeya24, line 533: May I ask why `avg_over_time` is not safe? Is it because in Thanos we do deduplication at Query time?
If we do write time deduplication like Cortex then this seems fine

fpetkovski: Yes, that's the reason. I believe it should be safe if you have unique data.

PR #4922, opened 2021-12-03
Sidecar: Use flag to ignore overlap check, add test for it

### LINKED ISSUE #4787: Shipper: allow uploading compacted overlapping blocks?

If vertical compaction is enabled at the compactor side, then it is okay to upload out of order blocks to the remote bucket and the overlapping check is not needed.

Have a new shipper flag to ignore the overlap checker. This could be a hidden flag as the vertical compaction flag in the compactor is also hidden.

metonymic-smokey:
Fixes #4787 

## Changes
Add a flag to ignore overlap check if vertical compaction is enabled on the compactor side.


metonymic-smokey:
Todo:
1. Clarify doubt about adding to `receive.MultiTSDB`
2. Add extra parameters to shipper tests.

metonymic-smokey:
@yeya24 @hanjm could you let me know if I'm proceeding on the right track here?
Should I add a similar flag for `receive` since that invokes a shipper too?

hanjm:
I think is right. @yeya24

pkg/receive/multitsdb.go
```diff
@@ -320,6 +320,7 @@ func (t *MultiTSDB) startTSDB(logger log.Logger, tenantID string, tenant *tenant
 			metadata.ReceiveSource,
 			false,
 			t.allowOutOfOrderUpload,
+			true, // Just set this to true as a temporary default until query is clarified.
```
yeya24, line 323: Let's set it to false as it is the default bahavior.

cmd/thanos/config.go
```diff
@@ -140,6 +141,9 @@ func (sc *shipperConfig) registerFlag(cmd extkingpin.FlagClause) *shipperConfig
 			"This can trigger compaction without those blocks and as a result will create an overlap situation. Set it to true if you have vertical compaction enabled and wish to upload blocks as soon as possible without caring"+
 			"about order.").
 		Default("false").Hidden().BoolVar(&sc.allowOutOfOrderUpload)
+	cmd.Flag("shipper.ignore-overlap",
+		"If true, shipper will skip checking for overlaps and allow uploading out of order blocks. Uploading out of order blocks is acceptable when vertical compaction is enabled at the compactor side.").
```
yeya24, line 145: Looks like this flag represents the same thing as the `allow-out-of-order-uploads` flag from the description. Does it make more sense to combine merge the two? Just skip the overlap checking when `allow-out-of-order-uploads` is enabled.

pkg/shipper/shipper_e2e_test.go
```diff
@@ -224,7 +224,7 @@ func TestShipper_SyncBlocksWithMigrating_e2e(t *testing.T) {
 		defer upcancel2()
 		testutil.Ok(t, p.WaitPrometheusUp(upctx2, logger))
 
-		shipper := New(log.NewLogfmtLogger(os.Stderr), nil, dir, bkt, func() labels.Labels { return extLset }, metadata.TestSource, true, false, metadata.NoneFunc)
+		shipper := New(log.NewLogfmtLogger(os.Stderr), nil, dir, bkt, func() labels.Labels { return extLset }, metadata.TestSource, true, false, false, metadata.NoneFunc)
```
yeya24, line 227: I expect an E2E test to verify that there is no overlap check anymore.

cmd/thanos/config.go
metonymic-smokey: Yes, I think combining them would be better.

metonymic-smokey: I think instead of skipping the check, it would be better to log the error and continue as usual if `allow-out-of-order-uploads` is enabled?

metonymic-smokey: I have used the existing flag for skipping the overlap check in a recent commit.

pkg/shipper/shipper_e2e_test.go
metonymic-smokey: Yes, this will take me some time to work through.
Thanks!

cmd/thanos/config.go
metonymic-smokey: Since I'm now using the existing flag, I'm working on creating a unit test for it, similar to the one for `uploadCompacted` flag.

bwplotka:
Can we remove WIP from title and use Draft feature - also can we remove draft/WIP if you want this to be merged / reviewed by others? Thanks! (:

pkg/shipper/shipper_e2e_test.go
```diff
@@ -351,3 +351,154 @@ func TestShipper_SyncBlocksWithMigrating_e2e(t *testing.T) {
 		testutil.Assert(t, ok == false, "fifth block was reuploaded")
 	})
 }
+
+// TestShipper_SyncOverlapBlocks_e2e is a unit test for the functionality by allowOutOfOrderUploads flag. This allows compacted(compaction level greater than 1) blocks to be uploaded despite overlapping time ranges.
+func TestShipper_SyncOverlapBlocks_e2e(t *testing.T) {
+	e2eutil.ForeachPrometheus(t, func(t testing.TB, p *e2eutil.Prometheus) {
```
bwplotka, line 357: Do we need this to be done for each Prometheus version? Sounds quite unnecessary 🤔 I might vote for some cheaper test instead, it all cost in waiting time =D

bwplotka (dismissed):
LGTM, just tests seems a little .. large, fur such small flag.

Wonder if a smaller unit test is possible or maybe adding a case to existing e2e?

pkg/shipper/shipper_e2e_test.go
metonymic-smokey: I get your point @bwplotka . Addressed it in a recent commit. 
I added a separate unit test since I thought the other tests in the shipper E2E tests were for similar features. 
To keep the test small, I've only added one case with two blocks.

metonymic-smokey: Another reason I added the test in `pkg/shipper/shipper_e2e_tests` is since I couldn't find the shipper tests to add it to in `test/e2ethanos`, hence decided to add it to the existing e2e shipper test.

metonymic-smokey: If you feel having the test is extraneous, I can reduce the PR to just the flag condition check.

pkg/shipper/shipper_e2e_test.go
```diff
@@ -351,3 +351,155 @@ func TestShipper_SyncBlocksWithMigrating_e2e(t *testing.T) {
 		testutil.Assert(t, ok == false, "fifth block was reuploaded")
 	})
 }
+
+// TestShipper_SyncOverlapBlocks_e2e is a unit test for the functionality by allowOutOfOrderUploads flag. This allows compacted(compaction level greater than 1) blocks to be uploaded despite overlapping time ranges.
+func TestShipper_SyncOverlapBlocks_e2e(t *testing.T) {
+	var p *e2eutil.Prometheus
```
yeya24, line 357: This line seems unnecessary

yeya24 (approved):
LGTM. Just one small nit.

pkg/shipper/shipper_e2e_test.go
metonymic-smokey: Thanks! Fixed.

yeya24 (approved):
LGTM

PR #4926, opened 2021-12-06
backport (query: fix deadlock in endpointset (#4795))

aymericDD:
Avoid RLock()ing twice as described here:
https://github.com/thanos-io/thanos/issues/4766#issuecomment-948743455
(due to
https://stackoverflow.com/questions/30547916/goroutine-blocks-when-calling-rwmutex-rlock-twice-after-an-rwmutex-unlock/30549188).
Fix it by removing HasClients() and simply changing it with `er.clients != nil`.

### DIFF
--- a/pkg/query/endpointset.go
+++ b/pkg/query/endpointset.go
@@ -688,46 +688,39 @@ func (er *endpointRef) ComponentType() component.Component {
 	return component.FromString(er.metadata.ComponentType)
 }
 
-func (er *endpointRef) HasClients() bool {
-	er.mtx.RLock()
-	defer er.mtx.RUnlock()
-
-	return er.clients != nil
-}
-
 func (er *endpointRef) HasStoreAPI() bool {
 	er.mtx.RLock()
 	defer er.mtx.RUnlock()
 
-	return er.HasClients() && er.clients.store != nil
+	return er.clients != nil && er.clients.store != nil
 }
 
 func (er *endpointRef) HasRulesAPI() bool {
 	er.mtx.RLock()
 	defer er.mtx.RUnlock()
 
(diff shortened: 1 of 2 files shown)


GiedriusS:
Thx, let's add the DCO :+1: cc @squat

squat:
Sounds good to me. I can cut the next patch release for 0.23 and add this to the change log

aymericDD:
@GiedriusS I can pass the DCO step because it is yours commit.

GiedriusS:
> @GiedriusS I cannot pass the DCO step because it is your commit.

Ofc you can, just add your DCO on top with the suggested commands ;P

aymericDD:
My bad :innocent:  My DCO is added to the commit

squat:
Thanks @aymericDD (:
The docs check is failing because of this recent PR: https://github.com/thanos-io/thanos/pull/4896
It seems that when we build docs, we are using the directory structure of the current commit but then trying to access the docs from `main`, which is not guaranteed to work, and indeed fails in this case because the `main` branch has moved this doc. If we want to get this PR to build, we need to cherry-pick this commit, which seems silly. I would say we merge this and then consider making a PR to the release-0.23 branch to fix CI.

WDYT @GiedriusS?

bwplotka (approved):
Makes sense, thanks!

aymericDD:
@squat do you know when the fix release will be published? It is very painful for us to maintain Thanos in production. Our queries are stuck or restart after a long time and we have high latency....

squat:
Today 👍

aymericDD:
Thanks @squat  :+1:

squat:
Here's the new PR to cut 0.23.2: https://github.com/thanos-io/thanos/pull/4936

PR #4927, opened 2021-12-06
rule: Added ability to specify multiple remote write targets.

bwplotka:
This is mainly for non surprising remote write YAML format, but also for ability
to add more targets, similar to agent supported model.

### DIFF
--- a/test/e2e/rule_test.go
+++ b/test/e2e/rule_test.go
@@ -492,7 +492,12 @@ func TestRule_CanRemoteWriteData(t *testing.T) {
 	testutil.Ok(t, e2e.StartAndWaitReady(receiver))
 	rwURL := mustURLParse(t, e2ethanos.RemoteWriteEndpoint(receiver.InternalEndpoint("remote-write")))
 
-	q, err := e2ethanos.NewQuerierBuilder(e, "1", receiver.InternalEndpoint("grpc")).Build()
+	receiver2, err := e2ethanos.NewIngestingReceiver(e, "2")
+	testutil.Ok(t, err)
+	testutil.Ok(t, e2e.StartAndWaitReady(receiver2))
+	rwURL2 := mustURLParse(t, e2ethanos.RemoteWriteEndpoint(receiver2.InternalEndpoint("remote-write")))
+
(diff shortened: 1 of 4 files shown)


test/e2e/rule_test.go
```diff
@@ -515,10 +515,10 @@ func TestRule_CanRemoteWriteData(t *testing.T) {
 				Scheme: "http",
 			},
 		},
-	}, &config.RemoteWriteConfig{
+	}, []*config.RemoteWriteConfig{{
 		URL:  &common_cfg.URL{URL: rwURL},
 		Name: "thanos-receiver",
-	})
+	}})
```
yeya24, line 521: Worth a test case for multiple remote write targets?

bwplotka: Hm. maybe not, since it's the Agent logic? We don't want to test Agent really

bwplotka: whatever, let's do it ;p

bwplotka: done

bwplotka:
PTAL @yeya24

bwplotka:
@hanjm wanna help in review/approve? Your reviews counts! (:

yeya24 (approved):
Awesome work!
LGTM

ahurtaud:
Hello,

Wasn't this change a `BREAKING` change instead of `ADDED` feature ?

I will confirm after fixing tomorrow my rule component on my side, but after upgrading from v0.24 to v0.25:
Because my remote_write config was not an array, it was not taken into account, and I had no error neither. At least my k8s pod was not crashing, so liveness probe was ok, but not remote writing anything :)

I have downgraded to v0.24 before switching to array, but I think changelog can be improved if that's the case.

ahurtaud:
EDIT2 ^^': I was right, and here is the diff:
on 0.24 a valid remote write config was:
```yaml
url: http://e2e_test_rule_remote_write-receive-1:8081/api/v1/receive
name: thanos-receiver
follow_redirects: false
```


on 0.25, a valid one is:
```yaml
remote_write:
- url: http://e2e_test_rule_remote_write-receive-1:8081/api/v1/receive
  name: thanos-receiver
  follow_redirects: false
```

Which means that version 0.25 with a config for 0.24 is not making the ruler crash but not taking the remote_write config,
because the yaml key remote_write did not exist

PR #3862, opened 2021-03-02
Upstream changes for PR #3361 - Added request logging for gRPC and HTTP in the server side

yashrsharma44:
I had deleted my thanos fork for some reason, and so I was not able to push the changes in the same branch.
Created another PR to reflect the changes for #3361

PR #4934, opened 2021-12-08
Document request logging configuration

### LINKED ISSUE #4861: Docs: Document the query logging feature better
Query logging is a useful feature, but right now the flags `--request.logging-config` and `--request.logging-config-file` do not seem to be sufficiently documented. We have only [a link](https://github.com/thanos-io/thanos/blob/b1c7483ea9be11a81e2c7422f19b811e81b98859/docs/components/query.md?plain=1#L368) to a personal GitHub gist page without much detail right now.

It would be nice if we had:
- All the documentation in this repository
- The documentation explains all possible configuration parameters
- There is a useful example of a request logging config

PhilipGough:
Fixes #4861 

### DIFF
--- a/docs/logging.md
+++ b/docs/logging.md
@@ -127,7 +127,7 @@ The following example shows how the logging config can be supplied to the `sidec
               level: ERROR
               decision:
                 log_start: false
-                log_end: true          
+                log_end: true
         - --tsdb.path=/prometheus-data
 ```
 
--- a/docs/logging.md
+++ b/docs/logging.md
@@ -1,0 +1,134 @@
+# Logging
+
+Thanos supports request logging via flags.
+
+Components are either configured using `--request.logging-config-file` to reference to the configuration file or `--request.logging-config` to provide configuration as YAML directly.
+
+## Configuration
+
+Configuration can be supplied globally, which applies to both `grpc` and `http`. Alternatively `http` and/or `grpc` can be configured independently.
+
+### Options
+
+Valid `level` for `options` should be one of `INFO`, `WARNING`, `ERROR` or `DEBUG`.
+
+```yaml
+options:
+  level: ERROR
+```
+
+The request logging is configurable per phase via the `decision`.
+
+Valid decision combinations are as follows:
+
+* Log details only when the request completes - Log level is determined by the status code from the response.
+
+```yaml
+options:
+  decision:
+    log_start: false
+    log_end: true
+```
+
+* Log details prior to the request and when the request completes - Pre request log is made at `debug` level.
+
+```yaml
+options:
+  decision:
+    log_start: true
+    log_end: true
+```
+
+### HTTP
+
+Specifying an `http` block enables request logging for each endpoint.
+
+```yaml
+http:
+  options:
+    level: DEBUG
+    decision:
+      log_start: true
+      log_end: true
+```
+
+The log level of the "post request" log is determined by the status code in the response.
+
+The following mappings apply:
+* `200 <= status code < 500` = `debug`
+* `status code > 500` = `error`
+
+Optionally, additional configuration can be supplied via the `config` block to create an allowlist of endpoint and port combinations. Note that this feature requires an exact match if enabled.
+
+```yaml
+http:
+  config:
+    - path: /api/v1/query
+      port: 10904
+    - path: /api/v1/app_range/metrics
+      port: 3456
+```
+
+### gRPC
+
+Specifying a `grpc` block enables request logging for each service and method name combination.
+
+```yaml
+grpc:
+  options:
+    level: DEBUG
+    decision:
+      log_start: true
+      log_end: true
+```
+
+The log level of the "post request" log is determined by the [code](https://grpc.github.io/grpc/core/md_doc_statuscodes.html) in the response.
+
+The following mappings apply:
+* Codes `2, 12, 13 ,15` = `error`
+* All other codes = `debug`
+
+Optionally, additional configuration can be supplied via the `config` block to create an allowlist of service and method combinations. Note that this feature requires an exact match if enabled.
+
+```yaml
+grpc:
+  config:
+    - service: thanos.Store
+      method: Info
+```
+
+## How to use `config` flags?
+
+The following example shows how the logging config can be supplied to the `sidecar` component:
+
+```yaml
+      - args:
+        - sidecar
+        - |
+          --objstore.config=type: GCS
+          config:
+            bucket: <bucket>
+        - --prometheus.url=http://localhost:9090
+        - |
+          --request.logging-config=http:
+            config:
+              - path: /api/v1/query
+                port: 10904
+            options:
+              level: DEBUG
+              decision:
+                log_start: true
+                log_end: true
+          grpc:
+            config:
+              - service: thanos.Store
+                method: Info
+            options:
+              level: ERROR
+              decision:
+                log_start: false
+                log_end: true          
+        - --tsdb.path=/prometheus-data
+```
+
+Note that in the above example, logs will be emitted at `debug` level. These logs will be filtered unless the flag `--log.level=debug` is set.
--- a/docs/components/query-frontend.md
+++ b/docs/components/query-frontend.md
@@ -291,11 +291,11 @@ Flags:
                                  flag (mutually exclusive). Content of YAML file
                                  with request logging configuration. See format
                                  details:
-                                 https://gist.github.com/yashrsharma44/02f5765c5710dd09ce5d14e854f22825
+                                 https://thanos.io/tip/thanos/logging.md/#configuration
       --request.logging-config-file=<file-path>
                                  Path to YAML file with request logging
                                  configuration. See format details:
-                                 https://gist.github.com/yashrsharma44/02f5765c5710dd09ce5d14e854f22825
+                                 https://thanos.io/tip/thanos/logging.md/#configuration
       --tracing.config=<content>
                                  Alternative to 'tracing.config-file' flag
                                  (mutually exclusive). Content of YAML file with
--- a/docs/components/query.md
+++ b/docs/components/query.md
@@ -376,11 +376,11 @@ Flags:
                                  flag (mutually exclusive). Content of YAML file
                                  with request logging configuration. See format
                                  details:
-                                 https://gist.github.com/yashrsharma44/02f5765c5710dd09ce5d14e854f22825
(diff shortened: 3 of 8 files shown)


docs/logging.md
```diff
@@ -0,0 +1,75 @@
...
+
+Configuration can be supplied for either gRPC, HTTP or both. Options can be supplied globally, which is applied to both gRPC and HTTP, or individually to either or both protocols.
+
+### HTTP
+
+```yaml
+http:
+  config:
+    - path: /api/v1/query
+      port: 10904
```
yeya24, line 17: What's this port config for?

PhilipGough: That would be coming from https://github.com/thanos-io/thanos/blob/main/pkg/logging/yaml_parser.go#L34

However I did not look too deeply into the implementation. Just double checked the original example https://gist.github.com/yashrsharma44/02f5765c5710dd09ce5d14e854f22825 against what was exposed by those types.

yeya24: It would be nice to document this field in more detail. Maybe @yashrsharma44 can help answer this. I feel confused when I saw this config TBH.

PhilipGough: Agree, I just don't have enough context to provide more info. I had a look at https://github.com/thanos-io/thanos/pull/3862 but it didn't tell me much.

docs/logging.md
```diff
@@ -0,0 +1,75 @@
+# Logging
+
+Thanos supports query logging via flags.
```
matej-g, line 3: I think we can make this statement even more generic, since this mechanism supports _request_ logging in general, which means it works for all components.

docs/logging.md
matej-g: I believe this is to specify which port to log the requests on - this might make sense in the context of components exposing multiple HTTP endpoints (e.g. receiver). But I'm also thinking if we could make the user experience here more intuitive and somehow figure out which HTTP server it is based on paths? :thinking:

matej-g (commented):
So glad we'll have this important feature documented better :heart: Adding few comments.

docs/logging.md
PhilipGough: I am not sure why it has to be so complicated personally, but maybe missing some context.

At https://github.com/thanos-io/thanos/blob/main/pkg/logging/http.go#L63 we purposely check the port and append it to the decider url. So a request for example to `http:localhost:9090/api/v1/query` comes in initially as `/api/v1/query` before we bolt on the port and pass it to the "decider" function. Could it not just be excluded entirely?

yeya24: Yeah, I agree. Exact match is inflexible.

PhilipGough: we can avoid the exact match and allow essentially a wildcard match by not configuring anything. Same applies for gRPC. I've documented as such

docs/logging.md
```diff
@@ -0,0 +1,134 @@
+# Logging
+
+Thanos supports request logging via flags.
+
+Components are configured using `--request.logging-config-file` to reference to the configuration file or `--request.logging-config` as YAML directly.
+
+## Configuration
+
+Configuration can be supplied for either gRPC, HTTP or both. Options can be supplied globally, which is applied to both gRPC and HTTP, or individually to either or both protocols.
```
matej-g, line 9:
```suggestion
Configuration can be supplied for either gRPC, HTTP or both. Options can be supplied globally, which is applied to both gRPC and HTTP, or individually to either of both protocols.
```

docs/logging.md
```diff
@@ -0,0 +1,134 @@
+# Logging
+
+Thanos supports request logging via flags.
+
+Components are configured using `--request.logging-config-file` to reference to the configuration file or `--request.logging-config` as YAML directly.
```
matej-g, line 5:
```suggestion
Components are either configured using `--request.logging-config-file` to reference to the configuration file or `--request.logging-config` to provide configuration as YAML directly.
```

Although I think might already be clear from e.g. `--help`?

docs/logging.md
```diff
@@ -0,0 +1,134 @@
...
+
+## Configuration
+
+Configuration can be supplied for either gRPC, HTTP or both. Options can be supplied globally, which is applied to both gRPC and HTTP, or individually to either or both protocols.
+
+### Options
+
+Valid `level` for `options` should be one of `INFO`, `WARNING`, `ERROR` or `DEBUG`.
```
matej-g, line 13: So what does `level` set? :thinking:

matej-g (commented):
This is great, thanks for digging in and documenting the behavior! Just few smallish suggestions.

docs/logging.md
PhilipGough: So `level`, from what I can see provides an additional filter. Take the following example

I have enabled `--log.level=debug` globally and request logging is enabled with `ERROR` - I will not see any of the `debug` logs emitted from the http middleware.

This could be seen as a nuisance or as a feature :)

docs/logging.md
PhilipGough: I think my wording is correct here actually but I see it can be confusing. Does this work for you:

```
Configuration can be supplied globally which applies to both `grpc` and `http`. Alternatively both `http` and/or `grpc` can be configured independently.
```

matej-g: Sounds good!

docs/logging.md
matej-g: Understood, then let's keep it!

matej-g (approved):
I think this is looking good now, great improvement! Good job @PhilipGough!

docs/logging.md
```diff
@@ -0,0 +1,134 @@
...
+options:
+  decision:
+    log_start: false
+    log_end: true
+```
+
+* Log details prior to the request and when the request completes - Pre request log is made at `debug` level.
+
+```yaml
+options:
```
dwilliams782, line 36: I think it would be worth adding what the default behaviour is here. From my testing, I think the default behaviour is:

```
options:
  level: DEBUG
  decision:
    log_start: true
    log_end: true
```

As the behaviour seems to be the same with or without this if `--log-level=debug` is specified?

dwilliams782:
I've found this PR when trying to work out how the logging config works. Thanks a lot for your effort!

docs/logging.md
PhilipGough: IIUC - the default is `ERROR` - https://github.com/thanos-io/thanos/blob/main/pkg/logging/grpc.go#L45

docs/logging.md
dwilliams782: I'm not able to get this config section to work. Without config, I can see three HTTP log entries:

GET /api/v1/query_exemplars?query=<query>
GET /api/v1/query?query=<query>
GET /api/v1/query_range?query=<query>

but these are removed when I add:

```
config:
    - path: /api/v1/query
      port: 10902
    - path: /api/v1/query_exemplars
      port: 10902
    - path: /api/v1/query_range
      port: 10902
```

10902 being the http port. Have any of you got this working, if so, what did you do differently? Maybe we can use this to improve the docs and make this a bit more clear?

PhilipGough: @dwilliams782 it is looking for an exact match. regex matching etc is not supported - I've documented the behaviour on https://github.com/thanos-io/thanos/pull/4934/files#diff-1a5e57d66b08bec5a71860d67f06912a004f03368c36910099d4ffae12970c7dR61

Omitting the config will match all as you have seen.

dwilliams782: @PhilipGough how do we know what an exact match is? AFAIK my config above should be an "exact" match, in that, they are the three endpoints being emitted (when running without config) and that is my HTTP port. How else can I diagnose this?

PhilipGough: @dwilliams782 - the exact match would include any query params etc - I think it would be beneficial to be able to add `\/api\/v1\/query\?.+` for example, but that is for another issues.

And I am sorry I don't have all the answers since I put this together by stumbling through the same pain points as you are right now :)

My diagnosis was done by running locally and outputting logs and stepping through the debugger - not aware of a better way.

dwilliams782: Ah I see! Perhaps your example of using config should specify that any parameters need to match too? That wasn't obvious to me.

Yeah, I really appreciate your efforts here. I'm not expecting you to know the answers, but hopefully asking the questions can help improve the documentation when we do stumble on the answers. Cheers!

PhilipGough: Yeah no worries, good point, I can add a couple of examples with query params to the example conf and it might make it more obvious

By the way @matej-g has created https://github.com/thanos-io/thanos/issues/4961 to track some of the issues and figure out improvements to this so might be worth keeping an eye on that going forward

wiardvanrij:
Hi thanks. Let's get this merged. Could you rebase from main as we've changed some tests and these are now missing?

PhilipGough:
@wiardvanrij - rebased - docs check seems to be failing but not sure that it is related if you wouldn't mind running the check again?

wiardvanrij:
> @wiardvanrij - rebased - docs check seems to be failing but not sure that it is related if you wouldn't mind running the check again?

Re-ran the job but it really has a diff in it. It _should_ make a change on your doc when you run it locally, that needs to be committed. I can take a look later, because I lack a bit of time at the moment to help you out right now.

PhilipGough:
> Re-ran the job but it really has a diff in it. It should make a change on your doc when you run it locally, that needs to be committed.

Noted, I ran `make docs` and it did indeed produce a diff but it feels like it is incorrect regardless..I've committed the results so lets see what the check shows

wiardvanrij:
@PhilipGough , are you sure you have rebased correctly? We made some changes in our docs/formatting. Perhaps manually diff your Makefile with that of ours in main?

PhilipGough:
@matej-g - will try that, although the output of the github action is `run make docs and commit changes` 🙃

matej-g (approved):
Everything is passing now 🎉 May we requests a merge @wiardvanrij? 🙂

wiardvanrij:
Yes, this is better than what we have now. Thanks so much @PhilipGough and sorry for the bad experience on getting all the tests done :( 

P.s. If people find things that are not correct or require extra input, I encourage those to iterate on this document and create a PR to improve it. For now, let's move forward.

wiardvanrij (approved):

PhilipGough:
> P.s. If people find things that are not correct or require extra input, I encourage those to iterate on this document and create a PR to improve it. For now, let's move forward.

My thoughts exactly. There were some suggestions for changes but I think getting this over the line firs tis a good step. Thanks @wiardvanrij, @matej-g

PR #4938, opened 2021-12-09
resolver: Improve logs when A record does not resolve from SRV record set

AdmiralGT:

## Changes

- I've improved the log when an A record of part of a SRV record fails to resolve to include which A record did not resolve.

## Verification

- None, this feels fairly low risk.

## Other

I'm hoping this is fairly uncontroversial, but happy to take advice on naming. I didn't think this was worth raising an issue about, but we've had problems where A records didn't resolve as part of a SRV record and only outputting the SRV record in the log.

### DIFF
--- a/pkg/discovery/dns/resolver.go
+++ b/pkg/discovery/dns/resolver.go
@@ -118,7 +118,7 @@ func (s *dnsSD) Resolve(ctx context.Context, name string, qtype QType) ([]string
 					return nil, errors.Wrapf(err, "lookup IP addresses %q", host)
 				}
 				if len(resIPs) == 0 {
-					level.Error(s.logger).Log("msg", "failed to lookup IP addresses", "host", host, "err", err)
+					level.Error(s.logger).Log("msg", "failed to lookup IP addresses", "srv", host, "a", rec.Target, "err", err)
 				}
 			}
 			for _, resIP := range resIPs {


pkg/discovery/dns/resolver.go
```diff
@@ -118,7 +118,7 @@ func (s *dnsSD) Resolve(ctx context.Context, name string, qtype QType) ([]string
 					return nil, errors.Wrapf(err, "lookup IP addresses %q", host)
 				}
 				if len(resIPs) == 0 {
-					level.Error(s.logger).Log("msg", "failed to lookup IP addresses", "host", host, "err", err)
+					level.Error(s.logger).Log("msg", "failed to lookup IP addresses", "srv", host, "a", rec.Target, "err", err)
```
GiedriusS, line 121: If I am not wrong the convention is to use upper-case letters for DNS record names? (:

GiedriusS: I forgot that :+1: makes sense, let's keep them lowercase then for consistency

GiedriusS (approved):
Let's add the DCO & we can merge :+1:

AdmiralGT:
Sorry, I squashed to remove the 2 unwanted commits and add the DCO. No actual changes have occured.

GiedriusS (approved):
:+1: thank you :hugs:

PR #4939, opened 2021-12-09
sidecar: not ready when Prometheus is unavailable

GiedriusS:
## Changes

Set Sidecar to not be ready when Prometheus is unavailable. This avoids
problems when Prometheus is replaying WAL and Sidecar is "up" but then
Query hangs while waiting for a response from Prometheus/Sidecar. Also,
it gets shown as UP in the Stores page with no ext. labels during this
time that is confusing.

## Verification

Improved e2e tests which pass.

### DIFF
--- a/test/e2e/query_test.go
+++ b/test/e2e/query_test.go
@@ -6,6 +6,7 @@ package e2e_test
 import (
 	"context"
 	"fmt"
+	"net/http"
 	"net/http/httptest"
 	"net/url"
 	"os"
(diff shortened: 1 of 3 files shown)


test/e2e/query_test.go
```diff
@@ -103,6 +104,40 @@ func sortResults(res model.Vector) {
...
+			return err
+		}
+		resp, err := http.DefaultClient.Do(req)
+		if err != nil {
+			return err
+		}
+		defer runutil.CloseWithErrCapture(&rerr, resp.Body, "closing resp body")
+
+		if resp.StatusCode == 200 {
+			return fmt.Errorf("got status code %d", resp.StatusCode)
```
Namanl2001, line 135:
```suggestion
			return fmt.Errorf("sidecar should not be ready, got HTTP status code %d OK", resp.StatusCode)
```

Namanl2001 (commented):
just a comment suggestion, otherwise LGTM. Thanks

test/e2e/query_test.go
yeya24: Why `OK`?

GiedriusS: Because that's what usually goes after `200` in the HTTP response https://developer.mozilla.org/en-US/docs/Web/HTTP/Status/200? (:

Namanl2001: > Because that's what usually goes after `200` in the HTTP response https://developer.mozilla.org/en-US/docs/Web/HTTP/Status/200? (:

yes, in case someone doesn't know what status code `200` signifies.

yeya24: Printing the status code and OK the same time seems redundant.

GiedriusS: So, can we merge this then @yeya24 ?

yeya24 (approved):
LGTM

PR #4956, opened 2021-12-15
logging: Allow request logging config to determine log level

PhilipGough:
Perhaps I am misunderstanding the original intentions, but the request logging configuration seems overly complicated.

I started working on documenting the config in https://github.com/thanos-io/thanos/pull/4934 and there are some comments there, but overall I find the config is unclear.

This PR changes the implementation slightly, since my take is that if I have gone to the trouble of configuring request logging, all I want to do is specify at what level I believe that should be output and have that determined by the ` --log.level`

This change allows the 'level' config in request logging 'options'
to determine the output level for those logs.

Prior to this, without setting --log.level flag on the component
to debug, no logs were emitted.

With this implementation, I think it would make sense to restrict the logging options to one of `debug`, `info`

If this approach seems reasonable to the maintainers we would need to tackle gRPC in similar fashion.

### DIFF
--- a/pkg/logging/http.go
+++ b/pkg/logging/http.go
@@ -24,17 +24,17 @@ type HTTPServerMiddleware struct {
 }
 
 func (m *HTTPServerMiddleware) preCall(name string, start time.Time, r *http.Request) {
-	logger := m.opts.filterLog(m.logger)
-	level.Debug(logger).Log("http.start_time", start.String(), "http.method", fmt.Sprintf("%s %s", r.Method, r.URL), "http.request_id", r.Header.Get("X-Request-ID"), "thanos.method_name", name, "msg", "started call")
+	m.opts.levelledLogger(m.opts.filterLog(m.logger)).Log(
+		"http.start_time", start.String(), "http.method", fmt.Sprintf("%s %s", r.Method, r.URL),
+		"http.request_id", r.Header.Get("X-Request-ID"), "thanos.method_name", name, "msg", "started call")
 }
 
 func (m *HTTPServerMiddleware) postCall(name string, start time.Time, wrapped *httputil.ResponseWriterWithStatus, r *http.Request) {
-	status := wrapped.Status()
-	logger := log.With(m.logger, "http.method", fmt.Sprintf("%s %s", r.Method, r.URL), "http.request_id", r.Header.Get("X-Request-ID"), "http.status_code", fmt.Sprintf("%d", status),
-		"http.time_ms", fmt.Sprintf("%v", durationToMilliseconds(time.Since(start))), "http.remote_addr", r.RemoteAddr, "thanos.method_name", name)
-
-	logger = m.opts.filterLog(logger)
-	m.opts.levelFunc(logger, status).Log("msg", "finished call")
+	m.opts.levelledLogger(m.opts.filterLog(m.logger)).Log(
+		"http.method", fmt.Sprintf("%s %s", r.Method, r.URL), "http.request_id", r.Header.Get("X-Request-ID"),
+		"http.status_code", wrapped.Status(), "http.time_ms", fmt.Sprintf("%v", durationToMilliseconds(time.Since(start))),
+		"http.remote_addr", r.RemoteAddr, "thanos.method_name", name,
+		"msg", "finished call")
 }
 
 func (m *HTTPServerMiddleware) HTTPMiddleware(name string, next http.Handler) http.HandlerFunc {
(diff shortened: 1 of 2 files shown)


pkg/logging/http.go
```diff
@@ -24,17 +24,17 @@ type HTTPServerMiddleware struct {
 }
 
 func (m *HTTPServerMiddleware) preCall(name string, start time.Time, r *http.Request) {
-	logger := m.opts.filterLog(m.logger)
-	level.Debug(logger).Log("http.start_time", start.String(), "http.method", fmt.Sprintf("%s %s", r.Method, r.URL), "http.request_id", r.Header.Get("X-Request-ID"), "thanos.method_name", name, "msg", "started call")
+	m.opts.levelledLogger(m.opts.filterLog(m.logger)).Log(
```
PhilipGough, line 27: this logger could probably be configured once in the constructor?

pkg/logging/http.go
```diff
@@ -24,17 +24,17 @@ type HTTPServerMiddleware struct {
...
 }
 
 func (m *HTTPServerMiddleware) postCall(name string, start time.Time, wrapped *httputil.ResponseWriterWithStatus, r *http.Request) {
-	status := wrapped.Status()
-	logger := log.With(m.logger, "http.method", fmt.Sprintf("%s %s", r.Method, r.URL), "http.request_id", r.Header.Get("X-Request-ID"), "http.status_code", fmt.Sprintf("%d", status),
-		"http.time_ms", fmt.Sprintf("%v", durationToMilliseconds(time.Since(start))), "http.remote_addr", r.RemoteAddr, "thanos.method_name", name)
-
-	logger = m.opts.filterLog(logger)
-	m.opts.levelFunc(logger, status).Log("msg", "finished call")
+	m.opts.levelledLogger(m.opts.filterLog(m.logger)).Log(
```
PhilipGough, line 33: as above

pkg/logging/http.go
PhilipGough: thinking about this some more I am also not sure that the filter is even needed considering we have setup a logger via `cmd/main.go` that is filtered based on `--log.level` and passed in to this middleware

pkg/logging/options.go
```diff
@@ -69,6 +70,12 @@ func WithFilter(f FilterLogging) Option {
 	}
 }
 
+func WithLevelledLogger(f LevelledLogging) Option {
```
matej-g, line 73: I'm wondering what is actually the reason to have `WithLeveledLogger` and `WithFilter` to be functional options and exported? It's not like we support users to provide their own leveled logger or filter, or do we?

pkg/logging/options.go
```diff
@@ -125,6 +136,21 @@ func DefaultCodeToLevel(logger log.Logger, code int) log.Logger {
 	return level.Error(logger)
 }
 
+func DefaultLevelledLogger(logger log.Logger, lvl string) log.Logger {
```
matej-g, line 139: Should we allow only `Info` and `Debug` in here as you mentioned in the PR description? Or perhaps we could validate level beforehand and only accept those two levels?

pkg/logging/http.go
matej-g: I think if we use the `leveledLogger` here to achieve the intended effect (i.e. log the requests on the configured level), `filterLog` should not be necessary as you said - we should have that governed by the global flag.

pkg/logging/options.go
PhilipGough: I agree, I am not sure that the functional options are required here. However I am not sure if there were other plans for the `logging` package to be used elsewhere or what the initial driver for that code was. It does seem excessive for the way the package is currently used.

To answer your question in regards to this `LevelledLogger` - the input in the config determines the level to log at, which is why I am using this currently. I believe it could all be done in the constructor as I had commented previously

pkg/logging/options.go
PhilipGough: See https://github.com/thanos-io/thanos/pull/4956#discussion_r770723183 but yes, or we could restrict the request logging config to only those two options while keeping the package more generic.

I didn't want to overhaul to much here without feedback from maintainers first :)

pkg/logging/http.go
PhilipGough: For sure, totally agree - maybe @bwplotka could chime in and let us know if there were "bigger plans" for the logging package or do we really need to expose these functional options at all?

PhilipGough:
closing based on the docs improvements in #4934

we can track improvements in #4961

Issue #4960, opened 2021-12-16
Deadlock if error while downsampling


GiedriusS:
**Object Storage Provider**:

S3

**What happened**:

Error occurred while downsampling but Thanos Compactor got stuck

**What you expected to happen**:

For Thanos Compactor to get into a "halt" state

**How to reproduce it (as minimally and precisely as possible)**:

Seems like it should be enough to have multiple blocks "to be downsampled" and for an error to occur while downsampling one of them.

**Full logs to relevant components**:

Goroutine dump:
<details>Dump
<p>

```
goroutine profile: total 24
4 @ 0x4380f6 0x4309fe 0x463049 0x4d5532 0x4d689a 0x4d6888 0x56c9c9 0x57e605 0x76fd2e 0x5519c3 0x551b1d 0x770aec 0x468981
#	0x463048	internal/poll.runtime_pollWait+0x88		/usr/lib/go-1.17/src/runtime/netpoll.go:229
#	0x4d5531	internal/poll.(*pollDesc).wait+0x31		/usr/lib/go-1.17/src/internal/poll/fd_poll_runtime.go:84
#	0x4d6899	internal/poll.(*pollDesc).waitRead+0x259	/usr/lib/go-1.17/src/internal/poll/fd_poll_runtime.go:89
#	0x4d6887	internal/poll.(*FD).Read+0x247			/usr/lib/go-1.17/src/internal/poll/fd_unix.go:167
#	0x56c9c8	net.(*netFD).Read+0x28				/usr/lib/go-1.17/src/net/fd_posix.go:56
#	0x57e604	net.(*conn).Read+0x44				/usr/lib/go-1.17/src/net/net.go:183
#	0x76fd2d	net/http.(*persistConn).Read+0x4d		/usr/lib/go-1.17/src/net/http/transport.go:1926
#	0x5519c2	bufio.(*Reader).fill+0x102			/usr/lib/go-1.17/src/bufio/bufio.go:101
#	0x551b1c	bufio.(*Reader).Peek+0x5c			/usr/lib/go-1.17/src/bufio/bufio.go:139
#	0x770aeb	net/http.(*persistConn).readLoop+0x1ab		/usr/lib/go-1.17/src/net/http/transport.go:2087

4 @ 0x4380f6 0x447ed2 0x7727bb 0x468981
#	0x7727ba	net/http.(*persistConn).writeLoop+0xfa	/usr/lib/go-1.17/src/net/http/transport.go:2386

[...]

PR #4962, opened 2021-12-16
downsample: fix deadlock if error occurs

GiedriusS:
Fix deadlock that occurs if there is some backlog of blocks to be
deduplicated, and an error occurs. For this to trigger, there needs to
be at least `downsampleConcurrency + 2` blocks in the backlog.

Add test to cover this regression. Affected versions 0.22.0 - 0.23.1.

Fixes https://github.com/thanos-io/thanos/issues/4960.


wiardvanrij (approved):
I've done a review, though it's a bit passed my experience level; it looks really good and it's clear what it fixes. So, thats the scope of my review :)

bwplotka (approved):
Thanks! Great resolution 👍🏽 

I don't thing this happens to default setups though as we don't have concurrency by default, right?

GiedriusS:
> Thanks! Great resolution 👍🏽
> 
> I don't thing this happens to default setups though as we don't have concurrency by default, right?

It does, the test shows that. :D It's enough to have an error occur while downsampling and to have some blocks in the downsampling backlog for this to occur. Most of the time it is enough to have two extra in the backlog but if you are lucky (depending on the scheduler) the deadlock might not happen

bwplotka:
👍🏽

PR #4963, opened 2021-12-16
compact+store+tools: Filter for duplicate block data only within a compaction group and in parallel

nberkley:

## Changes

Changes DeduplicateFilter to only look for duplicate block data within a single compaction group. This both provides a speedup by a factor of the number of group and allows the deduplication to be parallelized over groups. Concurrency level is taken from block-meta-fetch-concurrency as it is for other filters. In our deployment this dropped meta fetch time in the compactor from 10 hours to 6 minutes.

This does mean that, if you're using a replica label in the compactor, stores may load unnecessary blocks since they won't spot label change that will put the duplicate data in two different streams. This resolves after ignore-deletion-marks-delay when the store acknowledges that the old blocks were deleted. This seems like a good tradeoff since it only impacts a small section of data and provides a massive speedup for large buckets. It could be resolved by making the store aware of replica labels, but this diff is already getting rather large.

Also moves group key computation from DefaultGroupKey in the compact package to GroupKey on the Thanos object in the metadata package as now both the compact and block packages need to compute these.

## Verification

local tests pass, and we're actively using a version of the compactor with this change (though based off the 0.22 release, before the rebase)


nberkley:
As far as I can tell the e2e test failures here are not related to my changes. Those tests pass consistently when they're run in isolation locally (full runs time out, but so does a clean checkout of main).

hanjm:
Awsome performance improvment. 🚀

hanjm:
ping @yeya24

nberkley:
@bwplotka You might care about this one (you did the review for the PR that introduced this logic)

cmd/thanos/compact.go
```diff
@@ -271,9 +271,11 @@ func runCompact(
 				labelShardedMetaFilter,
 				consistencyDelayMetaFilter,
 				ignoreDeletionMarkFilter,
+				block.NewReplicaLabelRemover(logger, conf.dedupReplicaLabels),
 				duplicateBlocksFilter,
 				noCompactMarkerFilter,
-			}, []block.MetadataModifier{block.NewReplicaLabelRemover(logger, conf.dedupReplicaLabels)},
+				timePartitionMetaFilter,
```
GiedriusS, line 277: `timePartitionMetaFilter` has been listed twice in this list? :thinking:

cmd/thanos/tools_bucket.go
```diff
@@ -803,7 +803,7 @@ func registerBucketCleanup(app extkingpin.AppClause, objStoreConfig *extflag.Pat
 					block.NewConsistencyDelayMetaFilter(logger, tbc.consistencyDelay, extprom.WrapRegistererWithPrefix(extpromPrefix, reg)),
 					ignoreDeletionMarkFilter,
 					duplicateBlocksFilter,
-				}, []block.MetadataModifier{block.NewReplicaLabelRemover(logger, make([]string, 0))},
```
GiedriusS, line 806: Shouldn't we add this to the filters list?

cmd/thanos/tools_bucket.go
```diff
@@ -1327,7 +1327,7 @@ func registerBucketRetention(app extkingpin.AppClause, objStoreConfig *extflag.P
 					block.NewConsistencyDelayMetaFilter(logger, tbc.consistencyDelay, extprom.WrapRegistererWithPrefix(extpromPrefix, reg)),
 					duplicateBlocksFilter,
 					ignoreDeletionMarkFilter,
-				}, []block.MetadataModifier{block.NewReplicaLabelRemover(logger, make([]string, 0))},
```
GiedriusS, line 1330: Same here - does this belong to the filters list?

pkg/block/fetcher.go
```diff
@@ -716,8 +694,8 @@ func NewReplicaLabelRemover(logger log.Logger, replicaLabels []string) *ReplicaL
 	return &ReplicaLabelRemover{logger: logger, replicaLabels: replicaLabels}
 }
 
-// Modify modifies external labels of existing blocks, it removes given replica labels from the metadata of blocks that have it.
-func (r *ReplicaLabelRemover) Modify(_ context.Context, metas map[ulid.ULID]*metadata.Meta, modified *extprom.TxGaugeVec) error {
+// Filter modifies external labels of existing blocks, it removes given replica labels from the metadata of blocks that have it.
+func (r *ReplicaLabelRemover) Filter(_ context.Context, metas map[ulid.ULID]*metadata.Meta, synced *extprom.TxGaugeVec, modified *extprom.TxGaugeVec) error {
```
GiedriusS, line 698: This is the most interesting part of this PR :thinking: trying to understand this. Why can't we leave the old `Modify` interface and apply it to each group individually?

cmd/thanos/compact.go
nberkley: Oh, yeah, I think I messed that up in resolving merge conflicts. Fixed.

cmd/thanos/tools_bucket.go
nberkley: Doesn't matter here. It removes no labels as declared, it's just dead code.

cmd/thanos/tools_bucket.go
nberkley: Same issue, it's a duplicate label remover that removes no labels.

pkg/block/fetcher.go
nberkley: The concrete problem is sequencing. The replica label modifier changes what groups things are in so it needs to happen before deduplication or you actually do the wrong thing in the presence of replica labels. And you can't just move modifiers to be first because relabeling needs to see the original labels. So it has to fit into the middle of the filter list between those steps.

The two interfaces were identical other than the name of the metric to update so it seemed both cleanest and easiest to just pass both metrics to all of them and merge them into one pipeline of processing stages.

GiedriusS:
Seems like CI failure is related.

nberkley:
Yeah, someone added some new calls to NewStoreGW on main since last run. I'll have a fixed version up in a few minutes.

nberkley:
Should be set now if someone can approve running tests.

yeya24 (approved):
LGTM

yeya24 (approved):
LGTM

wiardvanrij:
I only did a basic check, I would recommend already adding the Changelog so this can move forward asap :)

nberkley:
> I only did a basic check, I would recommend already adding the Changelog so this can move forward asap :)

I thought the guideline was changelog only for user impacting changes? Happy to add one if that's not correct.

wiardvanrij:
> I thought the guideline was changelog only for user impacting changes? Happy to add one if that's not correct.

Hm I can imagine that is listed somewhere (if you know where, please give me a link :) ) 
However the top of the readme states

> All notable changes to this project will be documented in this file

So, just from my perspective, I think it adds value to add this change to the readme. Let's say I have an edge-case when upgrading my Thanos version. I will search for changes related to my issue/component. It's nice to see this in the changelog. 

Basically any change in components can have user impact, if something isn't correct ;p (I know, maybe a lame answer but it has nothing to do with the quality here, just in general!)

nberkley:
> Hm I can imagine that is listed somewhere (if you know where, please give me a link :) )

From the Changelog and Review Procedure in CONTRIBUTING.md:
> If your change affects users (adds or removes feature) consider adding the item to the CHANGELOG.

Also the PR template.

But point taken. I'll add a changelog entry. Better to over communicate than under communicate.

test/e2e/compact_test.go
```diff
@@ -455,7 +454,9 @@ func testCompactWithStoreGateway(t *testing.T, penaltyDedup bool) {
 		Type:   client.S3,
 		Config: e2ethanos.NewS3Config(bucket, m.InternalEndpoint("https"), e2ethanos.ContainerSharedDir),
 	}
-	str, err := e2ethanos.NewStoreGW(e, "1", svcConfig, "")
+
+	// Crank down the deletion mark delay since deduplication can miss blocks in the presence of replica labels it doesn't know about.
+	str, err := e2ethanos.NewStoreGW(e, "1", svcConfig, "", []string{"--ignore-deletion-marks-delay=2s"})
```
bwplotka, line 459: Do we now depend on this test to run faster than 2s?

pkg/block/fetcher.go
bwplotka: LGTM to me

pkg/block/fetcher.go
```diff
@@ -587,45 +574,50 @@ var _ MetadataFilter = &DeduplicateFilter{}
...
+	var groupChan = make(chan []*metadata.Meta)
 
-	metasByResolution := make(map[int64][]*metadata.Meta)
-	for _, meta := range metas {
-		res := meta.Thanos.Downsample.Resolution
-		metasByResolution[res] = append(metasByResolution[res], meta)
-	}
-
-	for res := range metasByResolution {
+	//Start up workers to deduplicate work groups when they're ready
```
bwplotka, line 594:
```suggestion
	// Start up workers to deduplicate workgroups when they're ready.
```

pkg/block/fetcher.go
```diff
@@ -637,55 +629,41 @@ func (f *DeduplicateFilter) filterForResolution(root *Node, metaSlice []*metadat
...
+			parentSources := parent.Compaction.Sources
+
+			// child's sources are present in parent's sources, filter it out.
+			if contains(parentSources, childSources) {
+				duplicates = append(duplicates, child.ULID)
+				continue childLoop
+			}
+		}
+
+		// Child's sources not covered by any member of coveringSet, add it to coveringSet
```
bwplotka, line 647:
```suggestion
		// Child's sources not covered by any member of coveringSet, add it to coveringSet.
```

bwplotka (dismissed):
Great job! Just minor nits, otherwise LGTM!

Thanks and sorry for delay. 🤗

test/e2e/compact_test.go
nberkley: No, it now _cannot_ run faster than 2s.

nberkley: To be more clear, this is working around the bit mentioned in the description where store gateways load duplicates for a bit in the presence of replica labels. It cranks the window during which we ignore deletion marks down so that the test can pass once we recognize the deletion marks.

nberkley:
@bwplotka Can you take another look? Think all the nits are addressed.

GiedriusS (approved):
Solid work! Thank you :muscle:

Issue #4949, opened 2021-12-14
Automatic update of Busybox sha.


bwplotka:
We use ` quay.io/prometheus/busybox@sha256:${BASE_DOCKER_SHA}` as base image. We pin SHA so we now exactly what SHA is running for each docker tag of Thanos image.

Unfortunately, quay kills old SHAs very often and apparently Prometheus busybox is upgraded often too.

Let's automate this process so PR with new sha is proposed once some interval. 

AC:
* We no longer need to manually hardcode and change base image SHAs for Prometheus busy box as explained [here](
https://github.com/thanos-io/thanos/blob/c7a44e23febe94f7f1b290c18fe778bf72cd01fd/Makefile#L15)

matej-g:
We had a small discussion on the Slack channel (https://cloud-native.slack.com/archives/CL25937SP/p1639582762123800) and it came up to couple of suggestions:
- Since it looks like we're currently always forces to use the `latest` image (since there's only that image tag in the https://quay.io/repository/prometheus/busybox?tab=tags repository), one idea was to - 1) _not_ pin the SHA, but get it on-the-fly from the `latest` image for both architectures; 2) include this information as a Docker label on the image
- Move to another base image which would have multiple version tags available (not only `latest`) and truly pin to the SHA which we want to use

wiardvanrij:
At the moment working on this

matej-g:
@wiardvanrij  I think it finally clicked for me why we cannot do it without pinning in the repository, since then whichever version / commit we're building the Thanos image for, we're always on `latest` BusyBox image - but it seems if the old BusyBox images do not exist after `latest` is overwritten, realistically we're unable to build with those non-`latest` base images since they are not available in the repo.

Anyways, long story short, having automatic PRs that would periodically sink the SHA sounds simpler and more straightforward way to go :+1:

wiardvanrij:
> 

Awesome! :D Yea I can explain this rationale later on, I got to go now, will update my PR later with exactly that. Also cheers for working together on this 👍

GiedriusS:
Can we close this now? (:

GiedriusS:
https://github.com/thanos-io/thanos/pull/4965 fixed this.

wiardvanrij:
Oh didn't got your previous reply.. but yeah 👍

PR #4964, opened 2021-12-16
Adds GitHub workflow to automate busybox container SHAs

wiardvanrij:
Signed-off-by: Rij, W. van (Wiard) <Wiard.vanRij@devolksbank.nl>

Relates to #4949 

## Changes

It uses a GitHub workflow with a cron, to check against quay for the prometheus/busybox image. It then fetches the SHA's per platform and places this in a `.busybox-versions` 
If there is a change, then a PR is made against `main`

The . file is loaded into the Makefile, so the vars are accessible in our build args. 

## Verification

On my fork, I have tested the workflow. 

## Other
I can add more rationale on other solutions and why I picked this later in the issue. Though regardless I see this as a somewhat "easier" fix and to be honest we still should consider a different approach. For example asking Prometheus to create more meaningful tags perhaps? :)


saswatamcode:
Thanks for this! 💫 
Can you resolve the conflicts so that checks start?

wiardvanrij:
Will remake this, as I made a complete mess :D

PR #4965, opened 2021-12-16
Adds GitHub flow to automate busybox container SHAs

wiardvanrij:
Signed-off-by: Rij, W. van (Wiard) <wiard@outlook.com>

Relates to #4949 

## Changes

The .github workflow does a cronjob task. It checks Quay for the latest Prometheus/busybox and fetches the SHA's for all platforms. These values are placed as bash variables in `.busybox-versions`. That file is included in the Makefile so the vars can be used as build arguments.

The cronjob will make a PR against `main`, if there is a change. 

Consider changing the cronjob to do it every 2-3 days, and allow manual updates if needed (?).

The reason we pin the version is that we want to know which version we are using AND that builds are consistent. Not just for the main branch but as well for local development, etc.

## Verification

I have tested this in my own fork. Which did work. I assume (ye ye) it should work here as well out of the box. I can't test if we have the proper rights, but it _should_ be fine as I did not use any special things here.


.github/workflows/container-version.yaml
```diff
@@ -0,0 +1,25 @@
+on:
+  schedule:
+    - cron: '37 13 * * *'
```
saswatamcode, line 3: Why at 13:37 daily?

.github/workflows/container-version.yaml
```diff
@@ -0,0 +1,25 @@
...
+      - uses: actions/checkout@v2
+      - name: Run busybox updater
+        run: |
+          chmod +x ./scripts/busybox-updater.sh
+          ./scripts/busybox-updater.sh
+        shell: bash
+      - name: Create Pull Request
+        uses: peter-evans/create-pull-request@v3
+        with:
+          signoff: true
```
saswatamcode, line 18: Maybe we can add `branch: update-sha-action` and `delete-branch: true` here so that the branches are auto-deleted after merge and we know what branches are made by this action?

.github/workflows/container-version.yaml
```diff
@@ -0,0 +1,25 @@
...
+  checkVersionAndCreatePR:
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v2
+      - name: Run busybox updater
+        run: |
+          chmod +x ./scripts/busybox-updater.sh
+          ./scripts/busybox-updater.sh
+        shell: bash
+      - name: Create Pull Request
```
saswatamcode, line 15: How about adding a step after this to log the PR link using this: https://github.com/peter-evans/create-pull-request#action-outputs? It can then be referencable from [Action run](https://github.com/thanos-io/thanos/actions). 🙂

saswatamcode (approved):
This looks awesome! 💫
Just some questions/suggestions.

.github/workflows/container-version.yaml
wiardvanrij: It's just a daily check. 1337 is just a random time :p 
We should see how it works/runs and adjust cron to our likings.

.github/workflows/container-version.yaml
wiardvanrij: It creates a patch, so I believe there is less of a mess in PR style. We can further check things by the PR's it's making as reference if needed. If this behaviour needs to be changed we can iterate on it. 
Basically it's now in the style of dependabot.

.github/workflows/container-version.yaml
wiardvanrij: Yea I'm not sure what you mean. 
What I do think is that we should move this forward, and you make an iteration with your magic on it. You have some good ideas and know how to make this better. No need to be limited by my work :) WDYT?

.github/workflows/container-version.yaml
matej-g: Yeah, random :wink:

.github/workflows/container-version.yaml
```diff
@@ -0,0 +1,25 @@
...
+        uses: peter-evans/create-pull-request@v3
+        with:
+          signoff: true
+          token: ${{ secrets.GITHUB_TOKEN }}
+          commit-message: 'Updates busybox SHA'
+          title: '[busybox] Updates busybox SHAs'
+          body: >
+            This PR is auto-generated by 
+            [create-pull-request](https://github.com/peter-evans/create-pull-request).
+          labels: busybox, automated pr
```
matej-g, line 25: Looks like we can add `team-reviewers` as well, maybe it would be worth to add Thanos maintainers or so, so we make sure these PRs do not slip through?

scripts/busybox-updater.sh
```diff
@@ -0,0 +1,28 @@
+#!/usr/bin/env bash
+
+# This script is called via .github/workflows/container-version.yaml
+# No need to manually run this (unless you want to force an update NOW)
+
+# Get the tags from the registry, so we can get the base manifest_digest ID
+echo "Doing first CURL request"
```
matej-g, line 7: Nit: maybe specify what each call is doing e.g.:
```suggestion
echo "Getting latest tags"
```

.github/workflows/container-version.yaml
```diff
@@ -0,0 +1,25 @@
...
+          chmod +x ./scripts/busybox-updater.sh
+          ./scripts/busybox-updater.sh
+        shell: bash
+      - name: Create Pull Request
+        uses: peter-evans/create-pull-request@v3
+        with:
+          signoff: true
+          token: ${{ secrets.GITHUB_TOKEN }}
+          commit-message: 'Updates busybox SHA'
+          title: '[busybox] Updates busybox SHAs'
```
matej-g, line 21: Nit: PR titles usually follow convention `<component>: <title>`, maybe
```suggestion
          title: '*: Updates Prometheus BusyBox image SHAs'
```

matej-g (approved):
Nicely done :raised_hands: Few optional suggestion, but it's already good to go :+1:

.github/workflows/container-version.yaml
saswatamcode: Sure, we can totally iterate! Would be nice to also get community input as well once this is in action! 🤗

.github/workflows/container-version.yaml
wiardvanrij: We will need a private access token to enable this.

.github/workflows/container-version.yaml
wiardvanrij: implemented

Makefile
```diff
@@ -441,4 +437,4 @@ $(PROTOC):
 	@PROTOC_VERSION="$(PROTOC_VERSION)" TMP_GOPATH="$(TMP_GOPATH)" scripts/installprotoc.sh
 	@echo ">> installing protoc@${PROTOC_VERSION}"
 	@mv -- "$(TMP_GOPATH)/bin/protoc" "$(GOBIN)/protoc-$(PROTOC_VERSION)"
-	@echo ">> produced $(GOBIN)/protoc-$(PROTOC_VERSION)"
+	@echo ">> produced $(GOBIN)/protoc-$(PROTOC_VERSION)"
```
yeya24, line 440: Please add a new line

.github/workflows/container-version.yaml
yeya24: Please add a new line here.

yeya24 (dismissed):
LGTM and some nits only. Thanks!

wiardvanrij:
I currently don't have a workstation anymore, figured to manually edit those things via GitHub online. Seems like it won't sign my commits then.. :( 
This might take a bit until I have set up a workplace with proper ssh/gpg etc again. Sorry about that.

yeya24 (approved):
LGTM

PR #4970, opened 2021-12-17
Exclude blocks marked for deletion

### LINKED ISSUE #4929: Add flag to "thanos tools bucket ls" to exclude marked blocks

I was trying to prepare a for loop to do rewrites and accidentally did rewrites on blocks that were marked for deletion. This ended up creating overlaps which prevented compactor from starting up.

It'd be nice if `ls` had an option to exclude blocks marked for deletion.

Alternatively maybe `rewrite` should refuse to rewrite a block marked for deletion?

metonymic-smokey:
Fixes #4929 

## Changes
Adding a flag to `tools bucket ls` to exclude blocks marked from deletion in the `ls` output. 

## Verification
Marked a block using `tools bucket mark` and verified that it did not appear in subsequent `ls` output.


metonymic-smokey:
Todo:
1. Is an extra `deleteDelay` flag required to filter or should I set it to some default duration?
2. Update docs and CHANGELOG.

cmd/thanos/tools_bucket.go
```diff
@@ -440,6 +443,18 @@ func registerBucketLs(app extkingpin.AppClause, objStoreConfig *extflag.PathOrCo
 			return err
 		}
 
+		if tbc.excludeDelete {
+			// Temporary placeholder values - should I add deleteDelay as a parameter for Ls too?
+			f := block.NewIgnoreDeletionMarkFilter(logger, bkt, 0, 1)
+
+			// Created fetcherMetrics since Filter required an *extprom.TxGaugeVec.
+			fetcherMetrics := block.NewFetcherMetrics(reg, nil, nil)
```
yeya24, line 451: The filtering step can be added to line 383

metonymic-smokey: Thanks @yeya24 ! This approach is definitely better. Fixed in a recent commit.

cmd/thanos/tools_bucket.go
```diff
@@ -377,9 +380,16 @@ func registerBucketLs(app extkingpin.AppClause, objStoreConfig *extflag.PathOrCo
...
 		}
 
-		fetcher, err := block.NewMetaFetcher(logger, block.FetcherConcurrency, bkt, "", extprom.WrapRegistererWithPrefix(extpromPrefix, reg), nil, nil)
-		if err != nil {
-			return err
+		var fetcher *block.MetaFetcher
+
+		if tbc.excludeDelete {
+			ignoreDeletionMarkFilter := block.NewIgnoreDeletionMarkFilter(logger, bkt, 0, block.FetcherConcurrency)
+			fetcher, err = block.NewMetaFetcher(logger, block.FetcherConcurrency, bkt, "", extprom.WrapRegistererWithPrefix(extpromPrefix, reg), []block.MetadataFilter{ignoreDeletionMarkFilter}, nil)
```
yeya24, line 387: This can be extracted out of the if else block.
You can just use something like:
``` go
var filters []block.MetadataFilter
if tbc.excludeDelete {
    // add ignore deletion filter to filters
}
fetcher, err = block.NewMetaFetcher(xxx)
```

yeya24 (changes requested):

cmd/thanos/tools_bucket.go
metonymic-smokey: Thanks! Fixed in a recent commit.

yeya24:
@metonymic-smokey LGTM. Can you add a changelog entry?

yeya24 (approved):
LGTM

Issue #4820, opened 2021-11-01
Add custom certificate CA support for object store


clyang82:
### Is your proposal related to a problem?

Currently, the implementation of connecting with object store is short of tls configuration.
https://github.com/thanos-io/thanos/blob/2695ecd69c4464914ad4b4480b5e924ebfcd919d/pkg/objstore/s3/s3.go#L150

for my case, I would need the ability to set a custom CA somewhere so that I can connect the external object store. for a workaround, i have to disable tls verification by
```
type: S3
config:
  bucket: ""
  endpoint: ""
  access_key: ""
  insecure: false
  signature_version2: false
  encrypt_sse: false
  secret_key: ""
  http_config:
    insecure_skip_verify: true
```

(Write your answer here.)

### Describe the solution you'd like

Support configuring certificate CA in `http_config`

(Describe your proposed solution here.)

### Describe alternatives you've considered

(Write your answer here.)

### Additional context

(Write your answer here.)

clyang82:
List the tasks for this issue:
- [ ] Support custom CA for S3
- [ ] Support custom CA for Azure
- [ ] Support custom CA for GCS
- [ ] Update e2e test to enable secure access to MinIO with TLS
- [ ] Update kube-thanos to support mount CA certs by secret

quchangl-github:
@clyang82  I got the update, Azure storage account doesn't support client key access - https://docs.microsoft.com/en-us/answers/questions/807958/where-to-find-the-client-ca-and-key.html.
So I don't think Azure storage can be accessed by customized CA. I saw your update [here](https://github.com/thanos-io/thanos/issues/4820#issuecomment-997771023) , GCS is using an inline a Service Account, I think Azure storage has a similar situation, it's using a storage account. Thanks.

PR #4974, opened 2021-12-20
 Add custom CA support for azure

clyang82:

## Changes

use tls_config to connect with azure object store so that we can pass the certificate in.

part of #4820


wiardvanrij (approved):
LGTM :) 

I also did not do it, but we really should make a generic wrapper for all the backends to use :D

clyang82:
Yes. Agree with your point. I have checked again. besides S3 and Azure, cos also need support tls_config. I will have a `tls_config.go` under objstore package so that it can be shared by other stores.

wiardvanrij (approved):
Really nice!

GiedriusS (approved):

Issue #3570, opened 2020-12-11
query frontend: Flaky E2E test


yeya24:
**What happened**:

https://github.com/thanos-io/thanos/pull/3566/checks?check_run_id=1534028614

```
=== RUN   TestQueryFrontend/query_frontend_works_for_instant_query
queryAndAssert: Waiting for 1 results for query sum(up) without (instance)
=== RUN   TestQueryFrontend/query_frontend_works_for_range_query_and_it_can_cache_results
=== CONT  TestQueryFrontendMemcachedCache
Error:     query_frontend_test.go:467: query_frontend_test.go:467:
        
         unexpected error: unable to find metrics [thanos_frontend_split_queries_total] with expected values. Last error: <nil>. Last values: [2]
        
23:20:43 Killing query-frontend-1
```

The query splitting test case is flaky because we are using the current timestamp to send queries, which makes the splitting behavior non-deterministic.

wiardvanrij:
Still relevant

wiardvanrij:
If the start and end time falls between a new day (passed 23:59), it will see this as '24h' (default --query-range.split-interval). My guess is that certain behavior has an underlying bug or that I'm a lost in the complexity. It will need to double the testing values if the hour is "1" for `thanos_frontend_split_queries_total` and `cortex_cache_*` However, only `thanos_frontend_split_queries_total` will need to double its value on hour 0. `cortex_cache_*`'s do not. This makes no sense to me and I also can't figure out why..

PR #4976, opened 2021-12-21
Remove flaky behavior in Thanos Query Frontend with a temporary fix

wiardvanrij:

Relates to https://github.com/thanos-io/thanos/issues/3570

I was killing myself on why this was happening and it hit me as it became night. Since it was already a long standing issue and I had no clue on how to properly fix it: tada. Here we are.

## Changes

Well, apparently (correct me if I'm wrong), queries are split by default of 24h. However if you do a query with a 'now' timestamp this 24h can already become 'true' if it passes the 00:00 (or 24:00, depending on how you look at it) mark. 

## Verification

Ran these tests locally and manually changing my time from 23:xx to 01:xx to 02:00 and it works on all cases.

### DIFF
--- a/test/e2e/query_frontend_test.go
+++ b/test/e2e/query_frontend_test.go
@@ -387,16 +387,19 @@ func TestQueryFrontend(t *testing.T) {
 	})
 }
 
-// https://github.com/thanos-io/thanos/issues/3570
-// "The query splitting test case is flaky because we are using the current timestamp to send queries, which makes the splitting behavior non-deterministic"
-// (Wiard)
-// This should be resolved in a much better way, but for now it removes the flakyness.
-func didItSplitValue(n float64, nightSwitch bool) float64 {
+// (Wiard) Query splitting behaviour is somewhat non-deterministric, due to the fact that we use a literal timestamp.FromTime(now.Add(-time.Hour).
+// If the start and end time falls between a new day (passed 23:59), it will see this as '24h' (default --query-range.split-interval).
+// My guess is that certain behavior has an underlying bug or that I'm a lost in the complexity.
+// It will need to double the testing values if the hour is "1" for thanos_frontend_split_queries_total and cortex_cache_*
+// However, only thanos_frontend_split_queries_total will need to double its value on hour 0. cortex_cache_*'s do not.
+// This test needs to be fixed properly and possibly requires more debugging, especially on the non-logical behaviour (or more documentation is required).
+// https://github.com/thanos-io/thanos/issues/3570 - however, for now it won't fail tests if ran between hour 0 and 1.
+func doubleValueIfSplit(n float64, whenPartialOvernight bool) float64 {
 
 	currentTime := time.Now()
 
 	// Some metrics DO NOT increase if the query range is from 'now -1hour' if it the hour is "0". Then just return the expected value.
-	if nightSwitch && currentTime.Hour() == 0 {
+	if whenPartialOvernight && currentTime.Hour() == 0 {
 		return n
 	}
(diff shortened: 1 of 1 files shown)


test/e2e/query_frontend_test.go
```diff
@@ -387,6 +387,27 @@ func TestQueryFrontend(t *testing.T) {
 	})
 }
 
+// https://github.com/thanos-io/thanos/issues/3570
+// "The query splitting test case is flaky because we are using the current timestamp to send queries, which makes the splitting behavior non-deterministic"
+// (Wiard)
+// This should be resolved in a much better way, but for now it removes the flakyness.
+func DidItSplitValue(n float64, nightSwitch bool) float64 {
```
yeya24, line 394: Let's make it lower case as it is only used in tests.

yeya24 (commented):
Nice. This test is flaky due to the usage of time.Now() which makes the splitting behavior varies when the timestamp changes. 
Like you mentioned it would be better to use absolute timestamp directly as Cortex and Loki query frontend tests do. But the current approach is good as well.

test/e2e/query_frontend_test.go
```diff
@@ -387,6 +387,27 @@ func TestQueryFrontend(t *testing.T) {
 	})
 }
 
+// https://github.com/thanos-io/thanos/issues/3570
+// "The query splitting test case is flaky because we are using the current timestamp to send queries, which makes the splitting behavior non-deterministic"
+// (Wiard)
+// This should be resolved in a much better way, but for now it removes the flakyness.
+func didItSplitValue(n float64, nightSwitch bool) float64 {
```
GiedriusS, line 394: I don't get the meaning of the variable `nightSwitch`, I feel like it's not descriptive :thinking: or maybe I am stupid :joy: Why for `cortex_cache_*` metrics we don't apply the doubling logic after the first if branch?

test/e2e/query_frontend_test.go
```diff
@@ -387,6 +387,27 @@ func TestQueryFrontend(t *testing.T) {
...
+
+	currentTime := time.Now()
+
+	// Some metrics DO NOT increase if the query range is from 'now -1hour' if it the hour is "0". Then just return the expected value.
+	if nightSwitch && currentTime.Hour() == 0 {
+		return n
+	}
+
+	// For the rest, if it is hour "0" or "1", the query is split, and we should expect double it's values.
+	if currentTime.Hour() == 1 || currentTime.Hour() == 0 {
```
GiedriusS, line 404: Is this dependent on the actual query? :thinking:

test/e2e/query_frontend_test.go
wiardvanrij: To be honest I thought I knew it yesterday but as I was trying to describe it here, it did not make any sense. I agree that if it would split, you should expect double the values on every related metric - at every timestamp. Now we only have that behavior when the hour is '1', and it's different on hour '0'. 

What I'm now trying to achieve is to actually remove the flakiness of this test. The internals have not changed, we merely hotfixed a test not fail if you try to run the test at a specific time of day. By no means should this be THE fix, it's a temporary thing that requires more effort. 

I've re-opened https://github.com/thanos-io/thanos/issues/3570 and placed my comment there as well. Like, this requires some investigation. If it does make sense, we should create better tests and possibly update the documentation.

I understand if we don't merge this, to create further confusion perhaps.

test/e2e/query_frontend_test.go
wiardvanrij: See my comment above.

PR #4985, opened 2021-12-22
store: fix label extending

GiedriusS:
Use the appropriate function `labelpb.ExtendSortedLabels` to extend the
labelset in `queryPrometheus` with external labels. It maintains the
invariant that labelset is a sorted set of labels. Otherwise, random
errors might occur with pushdown enabled:
`Error executing query: vector cannot contain metrics with the same labelset`

I have triggered this with a weird query `max_over_time(up[5m])` & with
HA pairs.


GiedriusS:
cc @fpetkovski

fpetkovski (approved):
Nice, thanks for catching this

PR #4987, opened 2021-12-22
Fix pushdown with deduplication on

GiedriusS:
Add some cases for pushdown with deduplication to test how it works.

There are some bugs here :'/ 

Fix them by creating a new series iterator that performs the given function over all replicas.


test/e2e/query_test.go
```diff
@@ -1006,3 +1004,132 @@ func synthesizeSamples(ctx context.Context, prometheus *e2e.InstrumentedRunnable
...
+		result       model.Vector
+	}{
+		{
+			query:        "max (my_fake_metric)",
+			prom1Samples: []fakeMetricSample{{"i1", 1}, {"i2", 5}, {"i3", 9}},
+			prom2Samples: []fakeMetricSample{{"i1", 3}, {"i2", 4}, {"i3", 10}},
+			result: []*model.Sample{
+				{
+					Metric: map[model.LabelName]model.LabelValue{},
+					Value:  9, // Why? Very weird.
```
GiedriusS, line 1024: I suppose dedup picks up the first seriesSet instead of the second one and then just goes on with it even though without pushdown we would compute max() over the whole data set?

GiedriusS (commented):
Here are some test cases to show some bugs in pushdown, unfortunately :/  cc @fpetkovski

test/e2e/query_test.go
```diff
@@ -761,9 +761,7 @@ func TestSidecarQueryEvaluation(t *testing.T) {
 			testutil.Ok(t, synthesizeSamples(ctx, prom2, tc.prom2Samples))
 
 			testQuery := func() string { return tc.query }
-			queryAndAssert(t, ctx, q.Endpoint("http"), testQuery, time.Now, promclient.QueryOptions{
-				Deduplicate: true,
```
GiedriusS, line 765: This wasn't actually testing deduplication since both Prometheus nodes were called differently

GiedriusS (commented):
Actually, I think what happens is that query_range aligns the timestamps for us then the deduplication doesn't work properly since they are all the same :/

fpetkovski:
Hm, I wonder if we could inject a synthetic label when a query is evaluated in the sidecar.

GiedriusS:
I think we could leverage series response hints for that. I'd prefer not to add metadata to labels to avoid allocations. In my opinion, we could solve the problem shown in these tests with a custom iterator that performs these functions (max/min) on top of all series from each replica to replicate the old behavior. What do you think, @fpetkovski ?

GiedriusS:
I have pushed a new pushdown series iterator that performs the given function over all replicas before returning the result. What do you think about this approach @bwplotka @fpetkovski ? Just need to decide how we'd like to indicate in SeriesResponse that results have been precomputed - via labels, via hints, or some other mechanism. I'd like to know your thoughts :hugs:

fpetkovski:
Can we add a test for the case where Querier is pushing a query to 2 different sets of replicated sidecars. For example, s1 and s2 being replicas of each other, and s3 and s4 also being replicas of each other. I am wondering what would happen if s1 and s3 have overlapping series, would the new iterator deduplicate them?

pkg/query/querier.go
```diff
@@ -331,6 +348,10 @@ func (q *querier) selectFn(ctx context.Context, hints *storage.SelectHints, ms .
 		warns: warns,
 	}
 
+	if allPushedDown && (strings.Contains(hints.Func, "max") || strings.Contains(hints.Func, "min")) {
```
GiedriusS, line 351: Note from the meeting: if the proxied data come here from different sources i.e. pushed down vs. not pushed down, we fall back to the old dedup iterator and in the results, you will see the same problem on time ranges where data comes from mixed sources. Ideally, we should have one iterator that handles this.

pkg/pushdown/iter.go
```diff
@@ -0,0 +1,212 @@
...
+
+	replicas []storage.Series
+	lset     labels.Labels
+	peek     storage.Series
+	ok       bool
+
+	function string
+}
+
+func NewSeriesSet(set storage.SeriesSet, replicaLabels map[string]struct{}, function string) storage.SeriesSet {
```
GiedriusS, line 27: This could be potentially be the same as in dedup.NewSeriesSet.

GiedriusS:
> Can we add a test for the case where Querier is pushing a query to 2 different sets of replicated sidecars. For example, s1 and s2 being replicas of each other, and s3 and s4 also being replicas of each other. I am wondering what would happen if s1 and s3 have overlapping series, would the new iterator deduplicate them?

Could you please elaborate on this with some examples? Because the deduplication should work the same.

I have pushed v2 of this change:

- Now the single dedup iterator handles interleaved cases no matter from where the data came from;
- All pushed-down series are passed through `dedupSeriesSet`. All pushed down series are shoved into `pushdownSeriesIterator` and then it becomes a regular series in `dedupSeries`. 

I think this approach is much cleaner and handles all cases. PTAL @bwplotka @fpetkovski :pray:

GiedriusS:
@yeya24 maybe this would be interesting to you as well

GiedriusS:
Untitled-2022-01-17-1805

Made a small picture showing how it works. @fpetkovski @bwplotka friendly ping :hugs:

niaurys (approved):

GiedriusS:
Given the upcoming release and no other comments, I'm likely to merge this to avoid shipping a broken feature. Next up on the list: https://github.com/thanos-io/thanos/issues/5069.

PR #4991, opened 2021-12-23
start minio server with tls

clyang82:

## Changes

Enable e2e connect MinIO with TLS enabled：
- Generate certs for MinIO TLS
- Update S3 Configure to have `tls_config` field

part of #4820

### DIFF
--- a/test/e2e/e2ethanos/services.go
+++ b/test/e2e/e2ethanos/services.go
@@ -31,6 +31,7 @@ import (
 
 	"github.com/thanos-io/thanos/pkg/alert"
 	"github.com/thanos-io/thanos/pkg/httpconfig"
+	"github.com/thanos-io/thanos/pkg/objstore"
 	"github.com/thanos-io/thanos/pkg/objstore/client"
 	"github.com/thanos-io/thanos/pkg/objstore/s3"
 	"github.com/thanos-io/thanos/pkg/queryfrontend"
@@ -1013,7 +1014,7 @@ func NewS3Config(bucket, endpoint, basePath string) s3.Config {
 		Endpoint:  endpoint,
 		Insecure:  false,
 		HTTPConfig: s3.HTTPConfig{
-			TLSConfig: s3.TLSConfig{
+			TLSConfig: objstore.TLSConfig{
 				CAFile:   filepath.Join(basePath, "data", "certs", "CAs", "ca.crt"),
 				CertFile: filepath.Join(basePath, "data", "certs", "public.crt"),
 				KeyFile:  filepath.Join(basePath, "data", "certs", "private.key"),
--- a/test/e2e/e2ethanos/services.go
+++ b/test/e2e/e2ethanos/services.go
@@ -957,12 +957,12 @@ func genCerts(certPath, privkeyPath, caPath, serverName string) error {
 		KeyUsage:     x509.KeyUsageDigitalSignature,
 	}
 
-	caPrivKey, err := rsa.GenerateKey(rand.Reader, 1024)
+	caPrivKey, err := rsa.GenerateKey(rand.Reader, 2048)
 	if err != nil {
 		return err
 	}
 
-	certPrivKey, err := rsa.GenerateKey(rand.Reader, 1024)
+	certPrivKey, err := rsa.GenerateKey(rand.Reader, 2048)
 	if err != nil {
 		return err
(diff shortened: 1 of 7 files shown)


GiedriusS (dismissed):

test/e2e/e2ethanos/services.go
```diff
@@ -814,37 +822,50 @@ http {
...
+		map[string]int{"https": 8090},
+		"https").Init(
 		e2e.StartOptions{
 			Image: image,
 			// Create the required bucket before starting minio.
-			Command:   e2e.NewCommandWithoutEntrypoint("sh", "-c", fmt.Sprintf(strings.Join(commands, " && "), minioKESGithubContent, minioKESGithubContent, bktName, 8090)),
-			Readiness: e2e.NewHTTPReadinessProbe("http", "/minio/health/ready", 200, 200),
+			Command: e2e.NewCommandWithoutEntrypoint("sh", "-c", fmt.Sprintf(strings.Join(commands, " && "), minioKESGithubContent, minioKESGithubContent, bktName, 8090)),
+			//TODO(@clyang82): This is a temporary workaround for https://github.com/efficientgo/e2e/issues/9
+			//Readiness: e2e.NewHTTPReadinessProbe("http", "/minio/health/ready", 200, 200),
```
matej-g, line 855: I'm wondering if this could not increase flakiness somewhere down the road, since we'll be assuming `minio` is ready straight away :thinking:. Could we try to do the readiness probe with the workaround I described in that commented issue?

clyang82: Do you mean try to do the readiness probe with the workaround here? I think we should support this feature in `efficientgo/e2e` and then update here.

matej-g: I'm not sure how soon that will be fixed in the upstream, but the flakiness could be immediate. I would agree with either solution (workaround or fix it upstream) as long as it can be done ASAP.

clyang82: @matej-g please look at the fix in upstream - https://github.com/efficientgo/e2e/pull/19
I also have a workaround for the readiness probe so that we can merge it this PR firstly. I can open another PR since the upstream fix is acceptable.

matej-g: Awesome! Thanks a ton for addressing my concern, this is looking good now 👍. We can replace it with proper option once it lands in the upstream.

matej-g (commented):
This now looks ready to me, I'm just not sure about the test failure, I cannot tell if it's related or not. Otherwise looking good to go 😊.

clyang82:
> This now looks ready to me, I'm just not sure about the test failure, I cannot tell if it's related or not. Otherwise looking good to go 😊.

Thanks @matej-g . I did test in my local environment. it works well. so it seems it is not related with my fix.

matej-g:
> Thanks @matej-g . I did test in my local environment. it works well. so it seems it is not related with my fix.

From logs it seems the issue is with `minio` server not being ready, see:
```
2022-01-10T09:50:19.7465904Z 09:50:19 store-gw-1: level=info name=store-gw-1 ts=2022-01-10T09:50:19.741901135Z caller=grpc.go:131 service=gRPC/server component=store msg="listening for serving gRPC" address=:9091
2022-01-10T09:50:19.7468549Z 09:50:19 store-gw-1: level=info name=store-gw-1 ts=2022-01-10T09:50:19.742995743Z caller=http.go:93 service=http/server component=store msg="internal server is shutdown gracefully" err="bucket store initial sync: sync block: BaseFetcher: iter bucket: Server not initialized, please try again."
```

Since we're touching the readiness probe, I'm wondering if this is having any effect or the `minio` readiness is still flaky even in current `main` with the workaround :thinking:. Perhaps rerunning the tests on this PR and seeing if they succeed would help us tell.

clyang82:
It passes the thanos end-to-end tests by adding `sleep 1` to fix the flakiness of minio. I will update once the [upstream PR](https://github.com/efficientgo/e2e/pull/19) is merged

matej-g (approved):
:trophy:

clyang82:
@GiedriusS Can you help to review again? Thanks.

yeya24 (approved):
LGTM

clyang82:
Thank you @yeya24

PR #5005, opened 2021-12-27
Increase indexheader_lazy_load_duration_seconds buckets

pracucci:

## Changes

In Cortex we've noticed that the 99th percentile computed on`indexheader_lazy_load_duration_seconds` is always capped to 5s even if the average can be way higher. Reason is that the largest histogram bucket is 5s but lazy loading can take way longer than that (mostly depending on index size and the disk speed).

In this PR I proposes to add more buckets to the metric, to have a better estimation of 99th percentile on high durations.

> Change is not relevant to the end user.

Is it? Do you track such changes in CHANGELOG?

## Verification

N/A

### DIFF
--- a/pkg/block/indexheader/lazy_binary_reader.go
+++ b/pkg/block/indexheader/lazy_binary_reader.go
@@ -59,7 +59,7 @@ func NewLazyBinaryReaderMetrics(reg prometheus.Registerer) *LazyBinaryReaderMetr
 		loadDuration: promauto.With(reg).NewHistogram(prometheus.HistogramOpts{
 			Name:    "indexheader_lazy_load_duration_seconds",
 			Help:    "Duration of the index-header lazy loading in seconds.",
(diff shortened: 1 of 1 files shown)


squat (dismissed):
Thanks @pracucci

yeya24 (approved):
Thanks @pracucci. I think 300s is little bit too large but overall LGTM.

pracucci:
> Thanks @pracucci. I think 300s is little bit too large but overall LGTM.

We've actually seen several minutes too, on a store-gateway under heavy load with indexes each of few tens GB each.

pracucci:
If there are no concerns, can we get this merged, please?

squat (approved):

squat:
thanks @pracucci :taco:

pracucci:
Thanks to you @squat ! 🌮

PR #5014, opened 2021-12-30
query: set default times for query_exemplars API

### LINKED ISSUE #5011: Thanos Query’s query_exemplars API behaviour differs from Prometheus’s
**Thanos, Prometheus and Golang version used**:  main-2021-12-23-cb447781, v2.31.1, ??

**Object Storage Provider**:

Backblaze B2 using S3 API

**What happened**:

I have a Grafana 8.3.3 instance configured to with a Prometheus datasource that points to Thanos Query. I noticed that Grafana would not display exemplars when the datasource was Thanos Query, but functioned as expected when Prometheus itself was the datasource.

I expected behaviour between Prometheus and Thanos to be consistent so that exemplars would be displayed in Grafana.

1. Install Grafana 
2. Install Thanos
3. Configure a datasource in Grafana to point to Thanos Query
4. Navigate to Explore and execute any query e.g., `up{app="prometheus"}`

Mouse-over the Exemplars icon and note that the tooltip displays `Exemplars for this query are not available`
Open your browser’s debug console and note all the failed requests to: `https://<grafanahostname>/api/datasources/proxy/1/api/v1/query_exemplars`

Grafana makes an API call to determine if exemplars are available. If the API call is successful then exemplars are enabled, otherwise they are disabled.

This is the line of code in the Grafana code base that performs the check: https://github.com/grafana/grafana/blob/43c81ddd23f574ab2bba8c23e008a6cc003028f3/public/app/plugins/datasource/prometheus/datasource.ts#L872

I can emulate the API check using curl: `curl -v -d query=test http://localhost:9090/api/v1/query_exemplars`

The response to the above API call differs between Prometheus and Thanos.

**Prometheus**

```
< HTTP/1.1 200 OK
< Content-Type: application/json
< Date: Wed, 29 Dec 2021 12:22:38 GMT
< Content-Length: 30
< 
* Connection #0 to host localhost left intact
{"status":"success","data":[]}
```

**Thanos**

```
< HTTP/1.1 400 Bad Request
< Access-Control-Allow-Headers: Accept, Accept-Encoding, Authorization, Content-Type, Origin
< Access-Control-Allow-Methods: GET, OPTIONS
< Access-Control-Allow-Origin: *
< Access-Control-Expose-Headers: Date
< Cache-Control: no-store
< Content-Type: application/json
< Vary: Accept-Encoding
< X-Thanos-Trace-Id: 2c8274dc001f4912
< Date: Wed, 29 Dec 2021 12:25:34 GMT
< Content-Length: 126
< 
{"status":"error","errorType":"bad_data","error":"rpc error: code = Code(400) desc = cannot parse \"\" to a valid timestamp"}
```

N/A

paulgrav:
Signed-off-by: Paul Grave <paul@stomer.net>

## Changes

Fixes #5011 

I replaced calls to `cortexutil.ParseTime` with `parseTimeParam` in `NewExemplarsHandler` since the later allows default times to be set.
I removed the `cortexutil` since the above change made it redundant.

The impact is that this behaviour mirrors Prometheus’ behaviour and allows exemplars to be displayed in Grafana.

## Verification

I effected some functional tests. I ran my own container (paulgrave/thanos:0.24.1) on my cluster and validated that exemplars are now displayed in Grafana.
Executed `curl -v -d query=test http://localhost:10902/api/v1/query_exemplars` to validate that a) API doesn’t error b) returns a 200 instead of a 400.
Executed `curl -v -d query='my_requests_total{app="prom-py"}' -d start=1640770731.689 -d end=(date +%s)   http://localhost:10902/api/v1/query_exemplars | jq` to validate that exemplars are indeed returned like before.
Thanos is running as expected on my local cluster.

Unfortunately `make lint` and `make test` are broken on my M1 Mac.

### DIFF
--- a/pkg/api/query/v1.go
+++ b/pkg/api/query/v1.go
@@ -28,7 +28,6 @@ import (
 	"strings"
 	"time"
 
-	cortexutil "github.com/cortexproject/cortex/pkg/util"
 	"github.com/go-kit/log"
 	"github.com/opentracing/opentracing-go"
 	"github.com/pkg/errors"
@@ -821,18 +820,18 @@ func NewExemplarsHandler(client exemplars.UnaryClient, enablePartialResponse boo
 			err      error
 		)
 
-		start, err := cortexutil.ParseTime(r.FormValue("start"))
+		start, err := parseTimeParam(r, "start", infMinTime)
 		if err != nil {
 			return nil, nil, &api.ApiError{Typ: api.ErrorBadData, Err: err}
 		}
-		end, err := cortexutil.ParseTime(r.FormValue("end"))
+		end, err := parseTimeParam(r, "end", infMaxTime)
 		if err != nil {
 			return nil, nil, &api.ApiError{Typ: api.ErrorBadData, Err: err}
 		}
 
 		req := &exemplarspb.ExemplarsRequest{
-			Start:                   start,
-			End:                     end,
+			Start:                   timestamp.FromTime(start),
+			End:                     timestamp.FromTime(end),
 			Query:                   r.FormValue("query"),
 			PartialResponseStrategy: ps,
 		}


yeya24 (approved):
LGTM. Thanks for your contribution!

PR #5015, opened 2021-12-30
[Website] fixed overflow company images and hero section logo 

NikhilSharma03:

### Overflow company images in `Used By` section

Screenshot from 2021-12-30 16-54-26
Screenshot from 2021-12-30 16-54-34

### Extremely Small logo image in `Hero` section (from viewport 576px - 750px)

Screenshot from 2021-12-30 17-02-54

### DIFF
--- a/website/static/main.css
+++ b/website/static/main.css
@@ -62,7 +62,7 @@ pre {
     height: 100px;
 }
 
-.img-adopter a{
+.img-adopter a {
     width: 100%;
     height: 100%;
 }
--- a/website/static/main.css
+++ b/website/static/main.css
@@ -50,7 +50,7 @@ pre {
 }
 
 .img-adopter {
-    padding: 25px;
+    padding: 20px;
     background-color: #fff;
     box-shadow: 0 2px 4px rgba(0,0,0,.1);
     border: 1px solid #dee2e6;
@@ -62,9 +62,15 @@ pre {
     height: 100px;
 }
 
-.img-adopter img {
-    max-width: 100%;
-    max-height: 100%;
+.img-adopter a{
+    width: 100%;
+    height: 100%;
+}
+
+.img-adopter a img {
+    object-fit: contain;
+    width: 100%;
+    height: 100%;
 }
 
 .header-anchor { font-size: 100%; visibility: hidden;}
--- a/website/static/main.css
+++ b/website/static/main.css
@@ -50,7 +50,7 @@ pre {
 }
 
 .img-adopter {
-    padding: 20px;
+    padding: 15px;
     background-color: #fff;
     box-shadow: 0 2px 4px rgba(0,0,0,.1);
     border: 1px solid #dee2e6;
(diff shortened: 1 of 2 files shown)


website/static/main.css
```diff
@@ -62,9 +62,15 @@ pre {
     height: 100px;
 }
 
-.img-adopter img {
-    max-width: 100%;
-    max-height: 100%;
+.img-adopter a{
```
squat, line 65: Nit: 
```suggestion
.img-adopter a {
```

website/static/main.css
```diff
@@ -166,3 +172,10 @@ input[type="search"]::placeholder { /* Chrome, Firefox, Opera, Safari 10.1+ */
         width: 1em;
     }
 }
+
+@media (max-width: 768px){
+    .hero-logo-container {
+        flex-grow: 1;
+        max-width: 100%;
+    }
+}
```
wiardvanrij, line 181: Please end with a newline

wiardvanrij (commented):
Thanks for this! One nitpick :)

squat (commented):
Looks pretty good @NikhilSharma03 
Can you share "after" screenshots too?

website/static/main.css
NikhilSharma03: Ok

NikhilSharma03:
## After Changes

### Fixed Overflow company images in Used By section
Screenshot from 2021-12-30 17-27-20
Screenshot from 2021-12-30 17-27-24

### Extremely Small logo image in Hero section (from viewport 576px - 750px)
Screenshot from 2021-12-30 17-27-32

wiardvanrij:
Looks really clean! Do you think it's possible to make them all equal in size? Now it sort of looks like "premium" users due to the size difference :)

NikhilSharma03:
@wiardvanrij I see :thinking: . Actually, the containers are of the same sizes. It is just that the image logo styles are different.

NikhilSharma03:
Tried using `object-fit: fill` results in,

squat:
> Now it sort of looks like "premium" users due to the size difference :)

I think we already had this problem before, so I wouldn't make it a blocker for this fix. Nevertheless, I think that the fundamental issue there is that our containers have a specific width:height ratio so logos that are closer to that ratio will appear larger. On top of that, some logos have more padding in their images than others.

I don't think there's anything we can do trivially to solve these issues, other than cropping logos to be the same size. An additional option would be to fix one dimension, e.g. height, and allow the other dimension to grow. However, this would result in a page that is not a fixed grid and would actually give non-square logos an advantage. Logos that are very wide would have tons of area while square or narrow logos would have much smaller areas.

squat:
Ah, @NikhilSharma03 I think what @wiardvanrij is referring to is actually the last row. Because this row has fewer logos, the containers are all larger. Can we make it so that containers are _always_ the same width?

NikhilSharma03:
@squat Actually All of the containers are of the same width. In the Screenshot (Last Row), I just cropped out the `SoundCloud` logo only since it was the one going out of the container.

Preview:
https://deploy-preview-5015--thanos-io.netlify.app/

wiardvanrij:
Right, awesome job 😎

squat:
Ah ok, I see, the were two different images

squat (approved):
Looks good to me :)
I just restarted the docs check; it had a flake trying to connect to prometheus.io

GiedriusS:
> Looks pretty good @NikhilSharma03 Can you share "after" screenshots too?

You can actually open the Netlify preview to see the difference :sunglasses:

PR #5017, opened 2021-12-31
Merge release 0.23 to main

squat:
This commit follows our release process and merges the latest commits/tags from the release-0.23 branch back into main. Next will be merging the release-0.24 branch back into main.

cc @GiedriusS


GiedriusS (approved):
DCO missing? Also, don't forget to use a "merge" commit so that tags would end up on the `main` branch

squat:
It's weird.. DCO is complaining about a lot of the previous commits in there, including some of yours :( https://github.com/thanos-io/thanos/pull/5017/checks?check_run_id=4673473289

It seems to be complaining that people changed their email addresses, e.g. @vinted.com vs @gmail.com

GiedriusS:
Weird :/ since all commits have `Signed-off-by`, maybe we can just click "Set DCO to pass"? What's interesting is that I have my old emails on my GitHub profile but for some reason the check expects the other e-mail

GiedriusS:
Clicked "Set DCO to pass".

PR #5018, opened 2021-12-31
Merge release 0.24 to main

squat:
This commit follows our release process and merges the latest commits/tags from the release-0.24 branch back into main. Let's merge this after https://github.com/thanos-io/thanos/pull/5017. This will likely need another rebase.

cc @GiedriusS


squat:
ping @GiedriusS
same issue as before. do you mind if I set DCO to `pass` for your commits?

VERSION
```diff
@@ -1 +1 @@
-0.24.0-dev
+0.24.0
```
GiedriusS, line 1: I think this now needs to be `0.25.0-dev`?

GiedriusS (dismissed):
:+1: yes, feel free to do that. Thx for this!

VERSION
squat: Ach yes that's right

squat: :tada: done

GiedriusS (approved):

PR #5020, opened 2022-01-03
store: close bkt properly

GiedriusS:
Just a small fix: add bkt closing to the run group instead of closing it
at the of the function that only sets things up. In practice, this
didn't cause a bug because Close() is a noop. But as good Go
programmers, let's Close() the bucket properly at the end of Thanos
Store's lifetime.

### DIFF
--- a/cmd/thanos/store.go
+++ b/cmd/thanos/store.go
@@ -279,11 +279,15 @@ func runStore(
 	}
 
 	// Ensure we close up everything properly.
-	defer func() {
-		if err != nil {
-			runutil.CloseWithLogOnErr(logger, bkt, "bucket client")
-		}
-	}()
+	{
+		done := make(chan struct{})
+		g.Add(func() error {
+			<-done
+			return bkt.Close()
+		}, func(error) {
+			close(done)
+		})
+	}
 
 	// Create the index cache loading its config from config file, while keeping
 	// backward compatibility with the pre-config file era.
--- a/cmd/thanos/store.go
+++ b/cmd/thanos/store.go
@@ -278,17 +278,6 @@ func runStore(
 		return errors.Wrap(err, "get content of index cache configuration")
 	}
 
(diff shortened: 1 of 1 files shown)


cmd/thanos/store.go
```diff
@@ -279,11 +279,15 @@ func runStore(
 	}
 
 	// Ensure we close up everything properly.
-	defer func() {
-		if err != nil {
-			runutil.CloseWithLogOnErr(logger, bkt, "bucket client")
-		}
-	}()
+	{
```
matej-g, line 282: I'm trying to follow the code - it looks like we're also closing `bkt` in another run group below at [365](https://github.com/thanos-io/thanos/blob/main/cmd/thanos/store.go#L365). Do we need both?

Also, are other components using bucket clients adhering to this or are they using simply `defer` instead as well?

cmd/thanos/store.go
```diff
@@ -279,11 +279,15 @@ func runStore(
...
+		done := make(chan struct{})
+		g.Add(func() error {
+			<-done
+			return bkt.Close()
```
matej-g, line 286: Should we retain `CloseWithLogOnErr` here so that we can log the potential close error?

matej-g (commented):
Always great to see best practices being turned into action :raised_hands: I was wondering about one thing, see comment.

cmd/thanos/store.go
GiedriusS: You are correct, I have missed that one. We were talking about this particular code snippet that I've edited in our meetings with Prem & Akansha so I haven't even bothered to check other places. With this, we can simply remove this code block. I don't know about all clients but I think it's customary to call Close() at the end of using some resource.

matej-g (approved):

squat (approved):
nice :taco: so, tl;dr we are already closing the bkt properly further on down

PR #5028, opened 2022-01-04
Feature/quick docs - only check docs that are changed for quick dev work

wiardvanrij:

## Changes

Added an extra Makefile option that only checks for .md files that are actually changed. Removing the need to iterate over the entire project (takes minutes!).

## Verification

Locally tested.


kakkoyun (approved):
This is a great change. How about using it for CI as well?

GiedriusS:
> This is a great change. How about using it for CI as well?

I don't think we can because we care about _all_ links working. There is an unfortunate consequence that any PR may fail due to unrelated changes but I think it's better than to have essentially basically close to no link checking at all :/

PR #5034, opened 2022-01-06
cli: print usage before error when argument parsing causes an error

jan--f:
This intends to improve error reporting in k8s. Before this change
thanos would print the error, followed by the usage. The latter part
would fill up a container status message (in say `kubectl describe`)
with the usage, hiding the actual error.

Signed-off-by: Jan Fajerski <jfajersk@redhat.com>


jan--f:
I'm sure there is a better solution for this, I went with the smallest change to see what the maintainers think.

GiedriusS (approved):

GiedriusS:
I think this makes sense and it's a nice usability improvement. :+1:

squat (approved):
Nice

GiedriusS:
Documentation check seems broken :(

Issue #5033, opened 2022-01-06
Pushdown: fix querying when proxied metrics have a mixture of pushed down and not pushed down data


GiedriusS:
PromQL engine removes `__name__` from results with `min_over_time` and `max_over_time`. When a query touches both pushed down results and not-pushed down results i.e. Thanos Store/Sidecar, this leads to an error during query time:
```
Error executing query: vector cannot contain metrics with the same labelset
```

Let's attach `__name__` forcefully from the Sidecar's perspective if those functions come in?

GiedriusS:
Actually, it's not so straightforward because one could execute a query such as:
```
max_over_time({__name__=~"prometheus_build_info|prometheus_api_remote_read_queries"}[5m])
```

And then from the result, it will be impossible to understand which labels belong to which metric :( perhaps then `__name__` needs to be stripped inside of the PromQL engine in Thanos Query when `min_over_time/max_over_time` come in.

PR #5035, opened 2022-01-06
query: fix querying with interleaved data

GiedriusS:
Fix querying when some data has been pushed down and some - hasn't.
Since `max_over_time` and `min_over_time` remove `__name__` from the
results either way, let's do that inside Select() to have a unified form
of data.

Add test to cover this case.

Without this code, the test fails:

```
level=error ts=2022-01-06T16:28:37.959956685Z msg="function failed. Retrying in next tick" err="read query instant response: expected 2xx response, got 422. Body: {\"status\":\"error\",\"errorType\":\"execution\",\"error\":\"vector cannot contain metrics with the same labelset\"}\n"
```

Closes https://github.com/thanos-io/thanos/issues/5033.

### DIFF
--- a/pkg/query/querier.go
+++ b/pkg/query/querier.go
@@ -14,6 +14,7 @@ import (
 	"github.com/pkg/errors"
 	"github.com/prometheus/client_golang/prometheus"
 	"github.com/prometheus/client_golang/prometheus/promauto"
+	"github.com/prometheus/common/model"
 	"github.com/prometheus/prometheus/model/labels"
 	"github.com/prometheus/prometheus/storage"
 	promgate "github.com/prometheus/prometheus/util/gate"
@@ -308,6 +309,22 @@ func (q *querier) selectFn(ctx context.Context, hints *storage.SelectHints, ms .
 	var warns storage.Warnings
 	for _, w := range resp.warnings {
 		warns = append(warns, errors.New(w))
+	}
+
+	// Delete the metric's name from the result because that's what the
+	// PromQL does either way and we want our iterator to work with data
+	// that was either pushed down or not.
+	if q.enableQueryPushdown && (hints.Func == "max_over_time" || hints.Func == "min_over_time") {
+		for i := range resp.seriesSet {
+			lbls := resp.seriesSet[i].Labels
+			for j, lbl := range lbls {
+				if lbl.Name != model.MetricNameLabel {
+					continue
+				}
+				resp.seriesSet[i].Labels = append(resp.seriesSet[i].Labels[:j], resp.seriesSet[i].Labels[j+1:]...)
+				break
+			}
+		}
 	}
 
 	if !q.isDedupEnabled() {
--- a/test/e2e/query_test.go
+++ b/test/e2e/query_test.go
@@ -10,6 +10,7 @@ import (
 	"net/http/httptest"
 	"net/url"
 	"os"
+	"path"
 	"path/filepath"
 	"sort"
 	"strings"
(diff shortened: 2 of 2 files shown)


GiedriusS:
cc @fpetkovski

fpetkovski (dismissed):

test/e2e/query_test.go
```diff
@@ -631,6 +638,93 @@ func newSample(s fakeMetricSample) model.Sample {
...
+// min_over_time.
+func TestSidecarStorePushdown(t *testing.T) {
+	t.Parallel()
+
+	// Build up.
+	e, err := e2e.NewDockerEnvironment("e2e_sidecar_store_pushdown")
+	testutil.Ok(t, err)
+	t.Cleanup(e2ethanos.CleanScenario(t, e))
+
+	prom1, sidecar1, err := e2ethanos.NewPrometheusWithSidecar(e, "p1", defaultPromConfig("p1", 0, "", ""), "", e2ethanos.DefaultPrometheusImage(), "remote-write-receiver")
```
yeya24, line 652: Is it intended to enable remote write receiver?

test/e2e/query_test.go
```diff
@@ -631,6 +638,93 @@ func newSample(s fakeMetricSample) model.Sample {
...
+	queryAndAssertSeries(t, ctx, q.Endpoint("http"), func() string {
+		return "max_over_time(my_fake_metric[2h])"
+	}, time.Now, promclient.QueryOptions{
+		Deduplicate: true,
+	}, []model.Metric{
+		{
+			"instance":   "foo",
+			"prometheus": "p1",
+		},
+	})
```
yeya24, line 725: Let's add another query for `min_over_time` as well?

test/e2e/query_test.go
fpetkovski: The test synthesizes samples by remote writing them, which is why the receiver feature is needed.

test/e2e/query_test.go
GiedriusS: Added tests for `min_over_time` & some other tricky cases: https://github.com/thanos-io/thanos/pull/5035/commits/203f449843681d05157030bec3cfa094f52a9fe7.

test/e2e/query_test.go
yeya24: Thanks @fpetkovski. I thought it was using data from the block only. Didn't notice the remote write part.

yeya24 (approved):
LGTM

pkg/query/querier.go
```diff
@@ -310,6 +311,22 @@ func (q *querier) selectFn(ctx context.Context, hints *storage.SelectHints, ms .
 		warns = append(warns, errors.New(w))
 	}
 
+	// Delete the metric's name from the result because that's what the
+	// PromQL does either way and we want our iterator to work with data
+	// that was either pushed down or not.
+	if q.enableQueryPushdown && (hints.Func == "max_over_time" || hints.Func == "min_over_time") {
```
yeya24, line 317: Just curious why promql does this. Isn't it a bug?

GiedriusS: Nope, it is a feature. Think of these functions like they are calculating the maximum over _all_ input, not just for each unique labelset. I think this is where it happens: https://github.com/prometheus/prometheus/blob/d677aa4b29a8a0cf9f61af04bbf5bfdce893cf23/promql/engine.go#L1304-L1310.

GiedriusS: Once more functions are being pushed down, we will need to improve the logic here. Probably we will need to refactor this into a separate struct? I don't know yet

Issue #5036, opened 2022-01-06
Remove constant spam of "changing probe status"


GiedriusS:
With Sidecar now changing its readiness status depending on whether it can establish a connection with Prometheus, there is spam of:

```
{"caller":"intrumentation.go:48","level":"info","msg":"changing probe status","status":"ready","ts":"2022-01-06T16:37:27.595400008Z"}
```

We shouldn't print this message if the status is set to the same value as the previous one.

AvineshTripathi:
Hey @GiedriusS very new here is this something I can go for without knowing the codebase

GiedriusS:
Yep. https://github.com/thanos-io/thanos/blob/e03452e54ce951d6eaa7a91b616b0aef5e0aafdd/pkg/prober/intrumentation.go#L48 here we need to introduce a new variable in `InstrumentationProbe` called `ready` or something. Like here: https://github.com/thanos-io/thanos/blob/19dcc7902d2431265154cefff82426fbc91448a3/pkg/prober/http.go#L19-L20. And then check in Ready() if we weren't ready before. If true then print. Same with NotReady().

AvineshTripathi:
okay i ll go for this please assign me

AvineshTripathi:
Hey @GiedriusS I have done some changes but not sure if that is working can you tell how do I check it

GiedriusS:
Run the quickstart script or in some other way, or manually a pair of Thanos Query+Sidecar (& Prometheus). If everything is ok then you won't see a constant spam as in the original description.

JuozasVainauskas:
@AvineshTripathi are You still working on this and planning to finish it or can I open a PR with my fix for this issue?

AvineshTripathi:
Hey @JuozasVainauskas I have done the changes but not able to set up in my local device so I think you should go for the PR i ll unaaisgn myself 
P.S. sorry for the trouble

GiedriusS:
Fixed in https://github.com/thanos-io/thanos/pull/5051

PR #5051, opened 2022-01-10
prober: Remove spam of changing probe status

JuozasVainauskas:
Issue: https://github.com/thanos-io/thanos/issues/5036

## Changes

Remove constant spam of "changing probe status" if status of the probe was set to the same value as before

## Verification

I ran `make quickstart` without my changes and saw that `"changing probe status"` messages are spammed even though probe status was set to the same value as before. Then I added my changes, ran `make quickstart` again and got certain that these kind of logs do not repeat if probe status does not change to different value than before.

### DIFF
--- a/pkg/prober/intrumentation.go
+++ b/pkg/prober/intrumentation.go
@@ -8,6 +8,7 @@ import (
 	"github.com/go-kit/log/level"
 	"github.com/prometheus/client_golang/prometheus"
 	"github.com/prometheus/client_golang/prometheus/promauto"
+	"sync"
 
 	"github.com/thanos-io/thanos/pkg/component"
 )
@@ -24,8 +25,9 @@ type InstrumentationProbe struct {
 	component component.Component
 	logger    log.Logger
 
-	status       *prometheus.GaugeVec
-	formerStatus string
+	statusMetric *prometheus.GaugeVec
+	mu           sync.Mutex
+	statusString string
 }
 
 // NewInstrumentation returns InstrumentationProbe records readiness and healthiness for given component.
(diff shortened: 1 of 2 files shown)


pkg/prober/intrumentation.go
```diff
@@ -13,8 +13,9 @@ import (
 )
 
 const (
-	ready   = "ready"
-	healthy = "healthy"
+	ready    = "ready"
+	notReady = "not-ready"
```
yeya24, line 17: Do we need notReady status? !ready just means notReady, right?

pkg/prober/intrumentation.go
```diff
@@ -23,7 +24,8 @@ type InstrumentationProbe struct {
 	component component.Component
 	logger    log.Logger
 
-	status *prometheus.GaugeVec
+	status       *prometheus.GaugeVec
+	formerStatus string
```
GiedriusS, line 28: We need a lock here since Ready/NotReady could be called from multiple goroutines.

pkg/prober/intrumentation.go
GiedriusS: Maybe we can leave `notReady` and use it in the log calls? So, we'd have:
```go
		level.Warn(p.logger).Log("msg", "changing probe status", "status", notReady, "reason", err)
```

pkg/prober/intrumentation.go
```diff
@@ -44,24 +48,34 @@ func NewInstrumentation(component component.Component, logger log.Logger, reg pr
 
 // Ready records the component status when Ready is called, if combined with other Probes.
 func (p *InstrumentationProbe) Ready() {
-	p.status.WithLabelValues(ready).Set(1)
-	level.Info(p.logger).Log("msg", "changing probe status", "status", "ready")
+	p.statusMetric.WithLabelValues(ready).Set(1)
+	p.mu.Lock()
+	defer p.mu.Unlock()
+	if p.statusString != ready {
+		level.Info(p.logger).Log("msg", "changing probe status", "status", "ready")
```
GiedriusS, line 55:
```suggestion
		level.Info(p.logger).Log("msg", "changing probe status", "status", ready)
```

GiedriusS (commented):
Please run `make go-format` to fix the linter issues :+1:

GiedriusS (approved):
:+1: thx

PR #5053, opened 2022-01-10
Add hyperia as an adopter 

mohavee:

Hello, first of all thanks for your work. At Hyperia we use Thanos in production and we would be glad to be an adopter.


squat (approved):
Thanks @mohavee

Issue #4664, opened 2021-09-16
E2E Tests: Consider enabling race detector for test binary


matej-g:
Enabling race detector for E2E tests could help us in catching concurrency problems early on (see context here: https://github.com/thanos-io/thanos/issues/4579#issuecomment-917716313).

We could introduce a separate Dockerfile to build image specifically for testing, with separate `promu` config (we'll need to enable CGO and use a non-Alpine based image).

bwplotka:
Yes 👍🏽

tend2infinity:
Hey, I wanted to work on this issue, can anyone give me some guidance regarding this, thank you!

matej-g:
Hey @tend2infinity, I'd say go for it!

As for guidance, it is written in the description in a kind of condensed form, but I believe we need to:
- Adjust (or probably create a new one really) Dockerfile which will take as a base image something else other than the Alpine image - the problem is the [race detector](https://go.dev/blog/race-detector) depends on having CGO enabled, which is not part of the Alpine image
- Add a new configuration for `promu` (utility which Thanos uses to build binaries) to ensure the binary for testing is actually build with CGO enabled. Have a look at current config in `.promu.yml` to get the picture.
- Adjust the `make` targets to use the new Dockerfile / build configuration for E2E tests.

If I still haven't cleared out something, feel free to ask!

tend2infinity:
Hey thanks a lot @matej-g , I was actually kinda confused but now it seems clear, I'll ask if I get stuck into something

matej-g:
This is actually in-progress, we just need a bit more :heart: to resolve races that have already been detected, see my comment in https://github.com/thanos-io/thanos/pull/5066#issuecomment-1056935086. These should be addressed before the race detection is merged, otherwise we'll (logically) end up with failing CI tests. Hopefully I'll (or someone else?) will be able to find the time to take a look soon :disappointed:.

cc @tend2infinity

PR #5066, opened 2022-01-15
Made new config for e2e tests having a non alpine image with cgo enabled

tend2infinity:

## Changes

- Added a new Dockerfile specifically for performing e2e tests using a non alpine image
- Added a new promu configuration for enabling cgo and race flags
- Specified targets in the makefile

## Verification
- Tested locally
### Part of
Issue #4664


.promue2e.yml
```diff
@@ -0,0 +1,24 @@
+go:
```
matej-g, line 1: Nit: Maybe we should name this file something like `.promu.e2e.yml`? It looks a bit more readable to me :nerd_face:.

Dockerfile.e2e-tests
```diff
@@ -0,0 +1,23 @@
+# By default we pin to amd64 sha. Use make docker to automatically adjust for arm64 versions.
+ARG BASE_DOCKER_SHA="14d68ca3d69fceaa6224250c83d81d935c053fb13594c811038c461194599973"
+
+# Taking a non-alpine image for e2e tests so that cgo can be enabled
+FROM golang:1.17 as builder
+
+WORKDIR $GOPATH/src/github.com/thanos-io/thanos
+# Change in the docker context invalidates the cache so to leverage docker
+# layer caching, moving update and installing apk packages above COPY cmd
+# More info https://docs.docker.com/develop/develop-images/dockerfile_best-practices/#leverage-build-cache
```
matej-g, line 10: Looks like this was just copy pasted from `Dockerfile.multi-stage` and we don't need it here?

Dockerfile.e2e-tests
```diff
@@ -0,0 +1,23 @@
...
+
+COPY . $GOPATH/src/github.com/thanos-io/thanos
+
+RUN git update-index --refresh; make build
```
matej-g, line 14: Should this be `build-e2e`? Otherwise we're building binary without the race detector, right?

```suggestion
RUN git update-index --refresh; make build-e2e
```

Makefile
```diff
@@ -176,6 +183,12 @@ else
 docker: docker-multi-stage
 endif
 
+.PHONY: docker-e2e
+docker-e2e: ## Builds 'thanos' docker for e2e tests
+docker-e2e: build-e2e
```
matej-g, line 188: Do we need to do `build-e2e` if it's part of the Dockerfile build?

matej-g (commented):
This looking nicely :clap: I added few comments + just please make sure you signed your commits (see section 'Signing your work' in https://github.com/thanos-io/thanos/blob/main/CONTRIBUTING.md#pull-request-process), so that we can kick start the CI as well!

.promue2e.yml
tend2infinity: Okay, will make this change

Dockerfile.e2e-tests
tend2infinity: Yup, actually I wanted to keep the next set of comments here, which were related to the "COPY" command, I guess I've mistakenly removed that and kept this comment, no problems! I'll make the change

Dockerfile.e2e-tests
tend2infinity: Tested locally and this works alright!

Makefile
tend2infinity: Yes, now this needs to be removed

tend2infinity:
@matej-g made the required changes

matej-g (commented):
Looking good, thanks @tend2infinity! Let's ping maintainers to make CI run for this PR.

Dockerfile.e2e-tests
```diff
@@ -0,0 +1,20 @@
+# By default we pin to amd64 sha. Use make docker to automatically adjust for arm64 versions.
```
squat, line 1: We need to update this comment, i think:
```suggestion
# By default we pin to amd64 sha. Use make docker-e2e to automatically adjust for arm64 versions.
```

.promu.e2e.yml
```diff
@@ -0,0 +1,24 @@
...
+  ldflags: |
+    -X github.com/prometheus/common/version.Version={{.Version}}
+    -X github.com/prometheus/common/version.Revision={{.Revision}}
+    -X github.com/prometheus/common/version.Branch={{.Branch}}
+    -X github.com/prometheus/common/version.BuildUser={{user}}@{{host}}
+    -X github.com/prometheus/common/version.BuildDate={{date "20060102-15:04:05"}}
+crossbuild:
+  platforms:
+    - linux/amd64
+    - darwin/amd64
```
squat, line 21: Why do we need anything other than Linux for the e2e tests? All of the tests run inside of a container, which means that the will run on Linux, and thus need a Linux binary, right?

Makefile
```diff
@@ -143,6 +143,13 @@ build: check-git deps $(PROMU)
 	@echo ">> building Thanos binary in $(PREFIX)"
 	@$(PROMU) build --prefix $(PREFIX)
 
+.PHONY: build-e2e
+build-e2e: ## Build Thanos binary for e2e tests using `promu` with cgo enabled
+build-e2e: check-git deps $(PROMU)
+	@echo ">> building Thanos binary in $(PREFIX)"
+	@$(PROMU) --config=".promu.e2e.yml" build --prefix $(PREFIX)
+
+
```
squat, line 152: Nit
```suggestion
```

Makefile
```diff
@@ -143,6 +143,13 @@ build: check-git deps $(PROMU)
 	@echo ">> building Thanos binary in $(PREFIX)"
 	@$(PROMU) build --prefix $(PREFIX)
 
+.PHONY: build-e2e
+build-e2e: ## Build Thanos binary for e2e tests using `promu` with cgo enabled
+build-e2e: check-git deps $(PROMU)
+	@echo ">> building Thanos binary in $(PREFIX)"
+	@$(PROMU) --config=".promu.e2e.yml" build --prefix $(PREFIX)
```
squat, line 150: Why do we need promu at all? No matter what platform i am running the tests on, the way the tests are configured, only a single image will be used, and that image will always have the same architecture as the machine on which `make test-e2e` was invoked.

It seems to me that promu is overkill/overengineered here. Instead, the Dockerfile.e2e-tests should directly embed the `go build` command with all of the necessary flags. cc @matej-g

Dockerfile.e2e-tests
tend2infinity: Gotcha! will update this

.promu.e2e.yml
tend2infinity: Okay, makes sense , we can remove the other platforms for sure

Makefile
tend2infinity: Actually I was following the convention, since the normal build command in the makefile is using promu and just to maintain homogeneity I used the promu utility here as well, maybe we can have a discussion around this in office hours, whatever is the conclusion, I can make the required changes.

matej-g: My thinking was the same, regardless of whether we need multiplatform build or not - since we're using Promu in the normal build flow, I didn't find it to be too much of an overhead, while being more uniform with the 'main' build flow. If the feeling is this is overengineerd, I don't have a problem with switching to simple `go build`.

tend2infinity: I guess it's postponed and maybe we'll have it on feb 3, I'll discuss this for sure and get this unblocked

matej-g: I'm not sure if we'll be able to get a clear guidance, maybe we can try to continue with the solution that seem to be agreeable for each of us, i.e. go with the `go build` suggestion? That would help us to unblock ASAP and we can always iterate later on an alternative.

Either way it looks like unblocking the failing E2E tests would need to be addressed before that.

tend2infinity: @matej-g Okay I'll integrate everything in the docker file itself

tend2infinity: Hey @matej-g you can check now, all the e2e tests passed locally for me

tend2infinity:
Hey @matej-g can you review it once ;)

matej-g (dismissed):
I think this is looking good :+1: One more cleanup suggestion :broom:

tend2infinity:
Changes done, works alright!

matej-g:
Thanks @tend2infinity and sorry for the delayed answer! Looks like everything is working fine, we are actually already catching some data races. Did you hit these during local runs? I think we should fix them before we merge this (otherwise E2E  tests will be broken). What I found based on test logs on my local machine:
- Data race in label set method on bucket store (I added fix, see PR https://github.com/thanos-io/thanos/pull/5204)
- Data race in receive in remote write handler (somewhere in https://github.com/thanos-io/thanos/blob/main/pkg/receive/handler.go#L538 - needs investigating / fixing)
- Data race in blocks API in set method (https://github.com/thanos-io/thanos/blob/main/pkg/api/blocks/v1.go#L133 - needs investigating / fixing)

matej-g:
Still relevant but not finished, on my to-do list...

tend2infinity:
Hey @matej-g sorry for the delay, I will look into these issues that you've mentioned and will let you know the progress

Issue #5037, opened 2022-01-06
Next steps with groupcache (galaxycache)


GiedriusS:
After https://github.com/thanos-io/thanos/pull/4818 has been merged, I am writing down a list of TODO:

- [ ] Make pull request that pushes changes (Thanos update) back into Cortex
- [x] Write documentation
- [ ] Switch back to the original galaxycache version by implementing the epoch strategy on keys suggested here (https://github.com/vimeo/galaxycache/pull/25, https://medium.com/vimeo-engineering-blog/video-metadata-caching-at-vimeo-a54b25f0b304)
- [x] Add microbenchmarks for fetching keys via HTTP. 
- [x] Enable H2C on our HTTP server to enable multiplexing even without TLS
- [ ] Add e2e tests with TLS on Thanos Store w/ galaxycache. Currently galaxycache's HTTP client doesn't use TLS even though the server can do it.
- [ ] Implement fetching multiple keys over one HTTP request. This isn't so trivial since we directly copy data from memory to the response's body :/ needs some kind of encoding scheme here that has minimal overhead?

For implementing the last two, perhaps we could even switch to gRPC?

Maybe I have missed something?

yeya24:
Looking forward to the documentation. Can't wait to try it 👍

GiedriusS:
https://github.com/thanos-io/thanos/pull/5068 fixes [4] and [5].

GiedriusS:
From our experience, groupcache is still slow due to the last point - we need to fetch multiple keys in one request.

pulkitent:
Hey @GiedriusS, I would like to pick a sub-issue of this open issue which is **Make pull request that pushes changes (Thanos update) back into Cortex**

As I am a newbie to open source contribution & Thanos codebase, can someone share some docs to help me understand the requirement better?

Any acceptance criteria for this feature? Or which flow/component in the Thanos codebase should I focus on? :)

bwplotka:
Amazing, thanks! So `Make pull request that pushes changes (Thanos update) back into Cortex` : I wonder if this is still relevant, given we vendor Cortex code. Could you double-check @GiedriusS ? Also I think it less important, given we are not sure if groupcache can be a better solution than Redis. We need to first figure out if groupcache is worth pursuing, so perhaps your (any new ideas) around making it faster (and benchmark) should be the next step here.

cc @yeya24 @GiedriusS does it sounds right?

GiedriusS:
[...]

PR #5068, opened 2022-01-17
groupcache: enable H2C + add benchmarks

GiedriusS:
Add h2c benchmarks. Somehow h2c is slowed than regular HTTP locally for
me. Probably because these aren't real life benchmarks i.e. establishing
TCP connections locally has no cost. And if there is only one operation
happening at any time then h2 has to do more work hence more CPU usage.

With 500 parallel requests going on at the same time, h2 becomes faster
than (or is at the same level as) h1 while using a minimal number of TCP
connections. Thus, it will work much faster in real life situations. So,
let's enable it.

```
name                                                               time/op
GroupcacheRetrieval/h2c/seq-16                                      227µs ± 1%
GroupcacheRetrieval/h2c/parallel=500-16                            54.8µs ±20%
GroupcacheRetrieval/h1,_max_one_TCP_connection/seq-16               150µs ± 2%
GroupcacheRetrieval/h1,_max_one_TCP_connection/parallel=500-16      146µs ± 1%
GroupcacheRetrieval/h1,_unlimited_TCP_connections/seq-16            145µs ± 3%
GroupcacheRetrieval/h1,_unlimited_TCP_connections/parallel=500-16  52.8µs ± 9%

name                                                               alloc/op
GroupcacheRetrieval/h2c/seq-16                                      183kB ± 0%
GroupcacheRetrieval/h2c/parallel=500-16                             143kB ± 1%
GroupcacheRetrieval/h1,_max_one_TCP_connection/seq-16               161kB ± 0%
GroupcacheRetrieval/h1,_max_one_TCP_connection/parallel=500-16      162kB ± 0%
GroupcacheRetrieval/h1,_unlimited_TCP_connections/seq-16            161kB ± 0%
GroupcacheRetrieval/h1,_unlimited_TCP_connections/parallel=500-16   116kB ± 2%

name                                                               allocs/op
GroupcacheRetrieval/h2c/seq-16                                        283 ± 0%
GroupcacheRetrieval/h2c/parallel=500-16                               256 ± 1%
GroupcacheRetrieval/h1,_max_one_TCP_connection/seq-16                 260 ± 0%
GroupcacheRetrieval/h1,_max_one_TCP_connection/parallel=500-16        262 ± 0%
GroupcacheRetrieval/h1,_unlimited_TCP_connections/seq-16              260 ± 0%
GroupcacheRetrieval/h1,_unlimited_TCP_connections/parallel=500-16     279 ± 1%
```


GiedriusS:
cc @akanshat @onprem

pkg/cache/groupcache.go
```diff
@@ -93,6 +96,12 @@ func NewGroupcacheWithConfig(logger log.Logger, reg prometheus.Registerer, conf
 	cfg *CachingBucketConfig) (*Groupcache, error) {
 	httpProto := galaxyhttp.NewHTTPFetchProtocol(&galaxyhttp.HTTPOptions{
 		BasePath: basepath,
+		Transport: &http2.Transport{
+			AllowHTTP: true,
+			DialTLS: func(network, addr string, cfg *tls.Config) (net.Conn, error) {
```
GiedriusS, line 101: For now, let's disable TLS but in the future, we can come back to this.

marijus-ravickas (approved):

pkg/cache/groupcache_test.go
```diff
@@ -16,32 +17,43 @@ import (
...
+const selfURLH1 = "http://localhost:12345"
+const selfURLH2C = "http://localhost:12346"
+
+func TestMain(m *testing.M) {
 	reg := prometheus.NewRegistry()
 	router := route.New()
 
 	bkt := objstore.NewInMemBucket()
-	b.Cleanup(func() { bkt.Close() })
+	defer bkt.Close()
```
kakkoyun, line 42:
```suggestion
	b.Cleanup(func() { bkt.Close() })
```
Why do we go back to defer here? AFAIK there's no functional difference and `Cleanup` better to avoid errors.

pkg/cache/groupcache_test.go
```diff
@@ -16,32 +17,43 @@ import (
...
 
-	testutil.Ok(b, bkt.Upload(context.Background(), "test", strings.NewReader(payload)))
-
-	listener, err := net.Listen("tcp4", "0.0.0.0:0")
-	testutil.Ok(b, err)
-
-	listenerAddr := listener.Addr().String()
-	port := strings.Split(listenerAddr, ":")[1]
+	if err := bkt.Upload(context.Background(), "test", strings.NewReader(payload)); err != nil {
+		fmt.Printf("failed to upload: %s\n", err.Error())
```
kakkoyun, line 47: Why not `Fatalf`?

pkg/cache/groupcache_test.go
```diff
@@ -52,30 +64,170 @@ func BenchmarkGroupcacheRetrieval(b *testing.B) {
...
 		},
-		"/_groupcache/",
+		basePath,
 		router,
 		bkt,
 		cachingBucketConfig,
 	)
-	testutil.Ok(b, err)
+	if err != nil {
+		fmt.Printf("failed creating group cache: %s\n", err.Error())
```
kakkoyun, line 80: Ditto

pkg/cache/groupcache_test.go
```diff
@@ -52,30 +64,170 @@ func BenchmarkGroupcacheRetrieval(b *testing.B) {
...
-	go func() { _ = http.Serve(listener, router) }()
-	b.Cleanup(func() { listener.Close() })
 	cachingBucketConfig.SetCacheImplementation(groupCache)
 
-	b.ResetTimer()
-	b.ReportAllocs()
+	exitVal := m.Run()
+
+	httpServer.Shutdown(nil)
+	os.Exit(exitVal)
```
kakkoyun, line 100:
```suggestion
	defer httpServer.Shutdown(nil)
	os.Exit(m.Run())
```
Maybe more idiomatic?

kakkoyun (dismissed):
Overall LGTM.

Just a couple of testing nits.

pkg/cache/groupcache_test.go
GiedriusS: It was a benchmark function before, now it is `TestMain(m *testing.M)`. The `M` type doesn't have those methods: https://pkg.go.dev/testing#M. Thus, `defer` is used because there's no better option. I hope that explains this change :smile:

pkg/cache/groupcache_test.go
GiedriusS: It was a benchmark function before, now it is `TestMain(m *testing.M)`. The `M` type doesn't have those methods: https://pkg.go.dev/testing#M. Thus, `fmt.Printf` is used because there's no better option. I hope that explains this change :smile:

pkg/cache/groupcache.go
GiedriusS: I will add it to https://github.com/thanos-io/thanos/issues/5037

pkg/cache/groupcache_test.go
```diff
@@ -52,30 +64,172 @@ func BenchmarkGroupcacheRetrieval(b *testing.B) {
...
 	)
-	testutil.Ok(b, err)
+	if err != nil {
+		fmt.Printf("failed creating group cache: %s\n", err.Error())
+		os.Exit(1)
+	}
+
+	go func() {
+		if err = httpServer.ListenAndServe(); err != nil {
+			fmt.Printf("failed to listen: %s\n", err.Error())
```
onprem, line 86: We should do an `os.Exit(1)` here as well. If we fail to start the http server, we should fail the test here instead of continuing with it.

pkg/cache/groupcache_test.go
```diff
@@ -52,30 +64,172 @@ func BenchmarkGroupcacheRetrieval(b *testing.B) {
...
+		}
+	}()
+	go func() {
+		if err = httpServerH2C.ListenAndServe(); err != nil {
+			fmt.Printf("failed to listen: %s\n", err.Error())
```
onprem, line 91: ditto, `os.Exit(1)` after this.

onprem (commented):
Thanks, this looks good. I have a minor suggestion which we can take care of in a follow-up. 

Sorry for the late review.

PR #5076, opened 2022-01-19
add NewBucketWithConfig for cos objstore

Nicholaswang:
Change is not relevant to the end-user.

## Changes
add NewBucketWithConfig func for cos.

## Note
I am importing objstore/cos package from thanos in other projects. Hoping to use NewBucketWithConfig as in s3 package rather than NewBucket() which only support byte slice parameter.


GiedriusS (approved):
:+1: thx

GiedriusS:
Add the DCO please and we can merge this.

Nicholaswang:
> Add the DCO please and we can merge this.

DCO added

GiedriusS (approved):

PR #5091, opened 2022-01-22
Add tenant tag to receive multitsdb_tenant_series

### LINKED ISSUE #5002: Receiver: instrument multitsdb tenant series tracing tag correctly
At https://github.com/thanos-io/thanos/blob/main/pkg/store/multitsdb.go#L274, a `tenant` tag is added. But it uses `ClientAddContextTags` method and that method is only used in gRPC method calls. In the case of tenant's local TSDB, there is no gRPC call anymore so the instrumentation doesn't work.

I expect to see https://github.com/thanos-io/thanos/blob/main/pkg/store/multitsdb.go#L172 get properly instrumented with the tenant tag.

Screenshot of the traces for tenant series call:
image

metonymic-smokey:
fixes #5002 

## Changes
Adding a tag with the tenant ID when tracing queries which use `receive`.


pkg/store/multitsdb.go
```diff
@@ -122,9 +122,14 @@ func newTenantSeriesSetServer(
...
 
+type clientSpanTagKey struct {
+	tenant string
+}
+
 func (s *tenantSeriesSetServer) Series(store storepb.StoreServer, r *storepb.SeriesRequest) {
 	var err error
-	tracing.DoInSpan(s.ctx, "multitsdb_tenant_series", func(_ context.Context) {
+	tracing.DoInSpan(s.ctx, "multitsdb_tenant_series", func(ctx context.Context) {
+		ctx = context.WithValue(ctx, clientSpanTagKey{tenant: s.tenant}, s.tenant)
```
yeya24, line 132: It might work but you are making it too complex. Can you just add a tag like https://github.com/thanos-io/thanos/blob/main/pkg/compact/compact.go#L773?

yeya24:
Can you mark it ready for review?

metonymic-smokey:
Done @yeya24. Adding a CHANGELOG entry too.
Thanks!

pkg/store/multitsdb.go
metonymic-smokey: Thanks, I had initially looked at the grpc-opentracing codebase and used the method of instrumentation referring to that.

yeya24:
> Done @yeya24. Adding a CHANGELOG entry too.
> 
> Thanks!

Thank you! Have you verified that this fix works?

yeya24:
Can you please also rebase master? Let's see if we can fix the CI error

metonymic-smokey:
I'm in the process of verifying it once more, since my laptop had a few issues the past couple of days. Sorry for the delay.   
Rebase origin/main results in quite a few merge conflicts so I'm working on fixing those too.

PR #5093, opened 2022-01-22
Support triggering update sha workflow manually

yeya24:

## Changes

Now the workflow is triggered by a cron job every day. But if the busybox image is updated after running the cron job that day, then the CI will be broken until the next day.

To support better control on this, I added support to trigger it manually. See doc https://docs.github.com/en/actions/using-workflows/events-that-trigger-workflows#workflow_dispatch.

## Verification

I am not sure if the current syntax is correct or not so would be great to have more eyes for review.

### DIFF
--- a/.github/workflows/container-version.yaml
+++ b/.github/workflows/container-version.yaml
@@ -1,4 +1,5 @@
 on:
+  workflow_dispatch: {}
   schedule:
     - cron: '37 13 * * *'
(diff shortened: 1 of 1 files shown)


GiedriusS (approved):
Looks good :+1:

GiedriusS:
CI failure seems unrelated?

PR #5096, opened 2022-01-24
Use Milliseconds instead of Second since ResLevel1DownsampleRange is …

clyang82:
Use Milliseconds instead of Second since ResLevel1DownsampleRange is in Milliseconds


GiedriusS (approved):
Thank you for the quick fix!

PR #5099, opened 2022-01-25
Fixed typo in sidecar docs

PabloL007:

## Changes

- Changed `...on order to use...` to `...in order to use...` in the object storage upload instructions for the sidecar component
- Changed `...which is needed to avoid the uploaded data corruption...` to `...which is needed to avoid the corruption of uploaded data...` in the object storage upload instructions for the sidecar component

## Verification

N/A


PabloL007:
@squat updated, thanks!

squat (approved):

PR #5103, opened 2022-01-26
Fix race condition in filesystem bucket client Delete()

pracucci:

## Changes

I found a race condition in the filesystem bucket client `Delete()`. The issue happen if 2 concurrent deletes occurs on different files in a subfolder containing only these 2 files. Both `Delete()` will try to delete the empty dir, but one of the two will fail with:

```
unexpected error: fdopendir /path/to/subfolder: no such file or directory
```

In this PR I propose to handle the `no such file or directory` in the `isDirEmpty()`. `isDirEmpty()` is used by Iter() and Delete(). In `Iter()` it's just used to skip empty dirs, so I think this change is fine there too.

## Verification

New unit test.

### DIFF
--- a/pkg/objstore/filesystem/filesystem.go
+++ b/pkg/objstore/filesystem/filesystem.go
@@ -199,7 +199,8 @@ func isDirEmpty(name string) (ok bool, err error) {
 	if os.IsNotExist(err) {
 		// The directory doesn't exist. We don't consider it an error and we treat it like empty.
 		return true, nil
-	} else if err != nil {
+	}
+	if err != nil {
 		return false, err
 	}
 	defer runutil.CloseWithErrCapture(&err, f, "dir open")
--- a/pkg/objstore/filesystem/filesystem.go
+++ b/pkg/objstore/filesystem/filesystem.go
@@ -11,12 +11,11 @@ import (
 	"os"
 	"path/filepath"
 
-	"github.com/thanos-io/thanos/pkg/objstore"
+	"github.com/pkg/errors"
 	"gopkg.in/yaml.v2"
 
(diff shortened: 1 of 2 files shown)


pkg/objstore/filesystem/filesystem_test.go
```diff
@@ -0,0 +1,43 @@
...
+import (
+	"context"
+	"strings"
+	"sync"
+	"testing"
+
+	"github.com/thanos-io/thanos/pkg/testutil"
+)
+
+func TestDelete_EmptyDirDeletionRaceCondition(t *testing.T) {
```
pracucci, line 12: This test fails in `main`.

GiedriusS (approved):
LGTM :+1:

pkg/objstore/filesystem/filesystem.go
```diff
@@ -197,12 +196,15 @@ func (b *Bucket) Upload(_ context.Context, name string, r io.Reader) (err error)
 
 func isDirEmpty(name string) (ok bool, err error) {
 	f, err := os.Open(filepath.Clean(name))
-	if err != nil {
+	if os.IsNotExist(err) {
+		// The directory doesn't exist. We don't consider it an error and we treat it like empty.
+		return true, nil
+	} else if err != nil {
```
squat, line 202: Nit: since the first if returns, we normally style the second if without an else
```suggestion
	}
	if err != nil {
```

squat (approved):
Lgtm, one small style nit

pkg/objstore/filesystem/filesystem.go
pracucci: Sure, addressed it!

squat (approved):
thank you @pracucci!!

PR #5105, opened 2022-01-27
fix: changing description to not include rule groups

prashbnair:
Signed-off-by: Prashant Balachandran <pnair@redhat.com>

Changing the description of the ThanosRuleNoEvaluationFor10Intervals alert rule to not include the number of rule groups as the expression was not evaluating the number of rule groups


matej-g (approved):
Looking good, thanks @prashbnair!

GiedriusS (approved):

Issue #4812, opened 2021-10-29
Rules: support external-label matcher. [Feature Request]


hanjm:
### Is your proposal related to a problem?

In sidecar mode, fanout rules api may response lots of rules. Could we support external-label based matcher, a bit like grafana? 
![image](https://user-images.githubusercontent.com/7621241/139385395-25f8db6e-32a2-452d-a516-be5af52f9388.png)

yeya24:
It would be good to support this in Prometheus first. I think this makes sense for example in multi tenant mode

hanjm:
> It would be good to support this in Prometheus first. I think this makes sense for example in multi tenant mode

I make a issue in Prometheus just now. 
Yeah, in multi tenant mode, prom-label-proxy can be auto inject tenant_id matcher.

PR #5111, opened 2022-01-29
Add matcher support to Query Rules endpoint

saswatamcode:
This PR adds support for label matchers as URL query params to the `api/v1/rules` endpoint of Thanos Query API. This will not break existing functionality.

It filters both recording and alerting rules via their non-templated labels and doesn't take the rendered labels of active alerts into account.

Currently, this PR only implements this functionality in Thanos, but this can be changed to a Prometheus-based implementation in the future (relevant PR: [prometheus#10194](https://github.com/prometheus/prometheus/pull/10194))! 

Addresses #4812.

Open question: Templated labels are identified by parsing the label values as templates and checking the number and type of node. So if it is parsed as a single node of type `parser.NodeText` it is treated as a non-templated label. Would a regex-based approach be better suited here?

## Changes
- Adds support for `match[]` URL param in `api/v1/rules` of Query API.
- Adds `MatcherString` to  `RulesRequest` in rules proto

## Verification

Tested locally.


pkg/rules/rules.go
```diff
@@ -67,6 +80,52 @@ func (rr *GRPCClient) Rules(ctx context.Context, req *rulespb.RulesRequest) (*ru
...
 }
 
+// filterRules filters rules in a group according to given matcherSets.
+func filterRules(ruleGroups []*rulespb.RuleGroup, matcherSets [][]*labels.Matcher) []*rulespb.RuleGroup {
+	if len(matcherSets) == 0 || len(ruleGroups) == 0 {
+		return ruleGroups
+	}
+
+	for _, g := range ruleGroups {
+		filteredRules := []*rulespb.Rule{}
```
hanjm, line 90: A small advice: use `filteredRules := g.Rules[:0]` avoid allocate. 
https://github.com/golang/go/wiki/SliceTricks#filtering-without-allocating

CHANGELOG.md
```diff
@@ -24,6 +24,7 @@ We use *breaking :warning:* to mark changes that are not backward compatible (re
 - [#4974](https://github.com/thanos-io/thanos/pull/4974) Store: Support tls_config configuration for connecting with Azure storage.
 - [#4999](https://github.com/thanos-io/thanos/pull/4999) COS: Support `endpoint` configuration for vpc internal endpoint.
 - [#5059](https://github.com/thanos-io/thanos/pull/5059) Compactor: Adding minimum retention flag validation for downsampling retention.
+- [#5111](https://github.com/thanos-io/thanos/pull/5111) Add matcher support to Query Rules endpoint
```
hanjm, line 27: A small nit: add dot at the line end.

hanjm:
It will be better to add some unit tests  (:

pkg/rules/rules.go
saswatamcode: Done!

CHANGELOG.md
saswatamcode: Done! Also, have added tests! 🙂

saswatamcode:
Added tests pass [here](https://app.circleci.com/pipelines/github/thanos-io/thanos/8710/workflows/60867648-52fe-4974-b159-27eae2ae45ef/jobs/18052?invite=true#step-107-531)!

pkg/rules/rulespb/rpc.proto
```diff
@@ -40,6 +40,7 @@ message RulesRequest {
     }
     Type type = 1;
     PartialResponseStrategy partial_response_strategy = 2;
+    repeated string MatcherString = 3;
```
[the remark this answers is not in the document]
saswatamcode: Sure, let me change!

pkg/rules/rules.go
```diff
@@ -58,6 +61,16 @@ func (rr *GRPCClient) Rules(ctx context.Context, req *rulespb.RulesRequest) (*ru
 		return nil, nil, errors.Wrap(err, "proxy Rules")
 	}
 
+	var matcherSets [][]*labels.Matcher
+	for _, s := range req.MatcherString {
+		matchers, err := parser.ParseMetricSelector(s)
+		if err != nil {
+			return nil, nil, errors.Wrap(err, "proxy Rules")
```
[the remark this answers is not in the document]
saswatamcode: Ah! Nice catch, will change this!

pkg/rules/rules.go
```diff
@@ -58,6 +61,16 @@ func (rr *GRPCClient) Rules(ctx context.Context, req *rulespb.RulesRequest) (*ru
 		return nil, nil, errors.Wrap(err, "proxy Rules")
 	}
 
+	var matcherSets [][]*labels.Matcher
```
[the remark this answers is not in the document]
saswatamcode: Yes, will pre-allocate!

pkg/rules/rulespb/rpc.proto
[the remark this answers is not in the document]
saswatamcode: Done!

pkg/rules/rules.go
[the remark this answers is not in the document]
saswatamcode: Done!

pkg/rules/rules.go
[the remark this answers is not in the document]
saswatamcode: Done!

matej-g (dismissed):
Nice job @saswatamcode!

Re regex approach, to me doing this node check seems more straightforward (and possibly also better performance-wise).

I'm wondering how long before this is supported in the downstream in Prometheus, since the linked PR seem to be well in progress, but I guess this is a good solution for the meantime :+1:.

pkg/rules/rules.go
```diff
@@ -58,6 +61,16 @@ func (rr *GRPCClient) Rules(ctx context.Context, req *rulespb.RulesRequest) (*ru
 		return nil, nil, errors.Wrap(err, "proxy Rules")
 	}
 
+	var err error
+	matcherSets := make([][]*labels.Matcher, len(req.MatcherString))
```
jessicalins, line 65: should we check if `len(req.MatcherString) > 0` or this is already covered from [here](https://github.com/thanos-io/thanos/pull/5111/files#diff-f14ed7ae9f9be7d76dea77aa31304b22de00af7e5f08720a7b6e75f721caf24aR790)?

jessicalins (commented):
cool! @saswatamcode :tada:

pkg/rules/rules.go
saswatamcode: If the length is zero, `matcherSets` would be an empty array of type `[][]*labels.Matcher` and the loop below this won't run. `filterRules` gets called right after and it checks the length of `matcherSets` [here](https://github.com/thanos-io/thanos/pull/5111/files#diff-a46b946d0da6144f5d8f5515bf381041db57211dff8fec08daf39fc46ab2b0a1R85) and if it is 0, returns all the rules. So we kind of do check for the length of `MatcherString`, just not directly. 🙂

jessicalins: sounds good!

pkg/api/query/v1.go
```diff
@@ -787,10 +787,15 @@ func NewRulesHandler(client rules.UnaryClient, enablePartialResponse bool) func(
 			typ = int32(rulespb.RulesRequest_ALL)
 		}
 
+		if err := r.ParseForm(); err != nil {
+			return nil, nil, &api.ApiError{Typ: api.ErrorInternal, Err: errors.Errorf("error parsing request form='%v'", typeParam)}
```
onprem, line 791: We are parsing the form for `MatcherParam` right? Then I think this should be 
```suggestion
			return nil, nil, &api.ApiError{Typ: api.ErrorInternal, Err: errors.Errorf("error parsing request form='%v'", MatcherParam)}
```

pkg/rules/rules.go
```diff
@@ -67,6 +80,52 @@ func (rr *GRPCClient) Rules(ctx context.Context, req *rulespb.RulesRequest) (*ru
 	return &rulespb.RuleGroups{Groups: resp.groups}, resp.warnings, nil
 }
 
+// filterRules filters rules in a group according to given matcherSets.
+func filterRules(ruleGroups []*rulespb.RuleGroup, matcherSets [][]*labels.Matcher) []*rulespb.RuleGroup {
+	if len(matcherSets) == 0 || len(ruleGroups) == 0 {
```
onprem, line 85: Just a nit: Although there is no harm in having `len(ruleGroups) == 0` here, it is redundant, because the following loop wont run and it'll just return.

```suggestion
	if len(matcherSets) == 0 {
```

pkg/rules/rules_test.go
```diff
@@ -954,3 +955,361 @@ func TestDedupGroups(t *testing.T) {
 		})
 	}
 }
+
+func TestFilterRules(t *testing.T) {
+	for _, tc := range []struct {
+		name         string
+		matcherSets  [][]*labels.Matcher
+		groups, want []*rulespb.RuleGroup
+	}{
```
onprem, line 964: Just to be more resilient, it'd be great if you can add a couple more cases.
- Multiple rules with only templated labels, and non-templated matcherSets
- A basic case but with other type of matcher than just `MatchEqual`

onprem (dismissed):
Neat work! I have a couple of non-blocking comments, but this looks ready to merge. :rocket:

pkg/api/query/v1.go
saswatamcode: Done!

pkg/rules/rules.go
saswatamcode: Sure, removed it.

pkg/rules/rules_test.go
saswatamcode: Yes, have added a test case with templated labels and non templated matcherSet as well as a regexp matcher test case!

onprem (approved):
LGTM. Great work!

PR #5114, opened 2022-02-02
Use time.RFC3339 in bucket inspect command for consistency

kfdm:

## Changes

Most places in thanos use RFC3339 for formatting, but I noticed that `thanos tools bucket inspect` had the format hardcoded.

## Verification

Built the thanos binary and tried the inspect command to verify the output

### DIFF
--- a/cmd/thanos/tools_bucket.go
+++ b/cmd/thanos/tools_bucket.go
@@ -417,7 +417,7 @@ func registerBucketLs(app extkingpin.AppClause, objStoreConfig *extflag.PathOrCo
 				maxTime := time.Unix(m.MaxTime/1000, 0)
 
 				if _, err = fmt.Fprintf(os.Stdout, "%s -- %s - %s Diff: %s, Compaction: %d, Downsample: %d, Source: %s\n",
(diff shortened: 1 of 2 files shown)


GiedriusS (dismissed):
:+1: thx

cmd/thanos/tools_bucket.go
```diff
@@ -997,8 +997,8 @@ func (t Table) Less(i, j int) bool {
 
 func compare(s1, s2 string) bool {
 	// Values can be either Time, Duration, comma-delimited integers or strings.
-	s1Time, s1Err := time.Parse("02-01-2006 15:04:05", s1)
-	s2Time, s2Err := time.Parse("02-01-2006 15:04:05", s2)
+	s1Time, s1Err := time.Parse(time.RFC1123, s1)
```
GiedriusS, line 1000: Does this need to be `RFC3339`?

kfdm: Good catch 😆  I'm not sure how I mixed up my copy/paste there but fixed now 🙇

squat (approved):
Thanks :)))

PR #5117, opened 2022-02-03
feat bucket replicate: added flag to ignore blocks marked for deletion

FUSAKLA:
Hi, we bumped into some race conditions with `tools bucket replicate`  from bucket with a compactor running, such as replication of a block that just gets deleted by compactor. 
Also, this can save some bandwidth, since it's mostly useless to sync blocks that will be deleted eventually.

Would you be ok accepting this? I'm open to any suggestions., thanks!

## Changes

 - added a new flag `--ignore-marked-for-deletion` to the `thanos bucket replicate` command 

## Verification

 - added tests
 - tested manually


pkg/replicate/scheme.go
```diff
@@ -192,7 +198,8 @@ func (rs *replicationScheme) execute(ctx context.Context) error {
 	}
 
 	for id, meta := range metas {
-		if rs.blockFilter(meta) {
+		_, err := rs.fromBkt.ReaderWithExpectedErrs(rs.fromBkt.IsObjNotFoundErr).Get(ctx, path.Join(meta.ULID.String(), metadata.DeletionMarkFilename))
```
GiedriusS, line 201: Do we need to have `--ignore-deletion-marks-delay` instead? I wonder if it makes sense :thinking: Mhm, probably not, because we are only copying the blocks here, and they aren't being replaced with one another.

Also, do we need to log `err` if we fail (not due to 404) to check whether the deletion mark exists? Preferably with `debug` log level.

pkg/replicate/scheme_test.go
```diff
@@ -116,6 +125,28 @@ func TestReplicationSchemeAll(t *testing.T) {
...
+				testutil.Ok(t, err)
+				d, err := json.Marshal(deletionMark)
+				testutil.Ok(t, err)
+				_ = originBucket.Upload(ctx, path.Join(ulid.String(), "meta.json"), bytes.NewReader(b))
+				_ = originBucket.Upload(ctx, path.Join(ulid.String(), "deletion-mark.json"), bytes.NewReader(d))
+				_ = originBucket.Upload(ctx, path.Join(ulid.String(), "chunks", "000001"), bytes.NewReader(nil))
+				_ = originBucket.Upload(ctx, path.Join(ulid.String(), "index"), bytes.NewReader(nil))
+			},
+			assert: func(ctx context.Context, t *testing.T, originBucket, targetBucket *objstore.InMemBucket) {
+				if len(targetBucket.Objects()) != 0 {
```
GiedriusS, line 145: Nit: let's do: `testutil.Equals(t, ...)`.

cmd/thanos/tools_bucket.go
```diff
@@ -695,6 +695,7 @@ func registerBucketReplicate(app extkingpin.AppClause, objStoreConfig *extflag.P
 	maxTime := model.TimeOrDuration(cmd.Flag("max-time", "End of time range limit to replicate. Thanos Replicate will replicate only metrics, which happened earlier than this value. Option can be a constant time in RFC3339 format or time duration relative to current time, such as -1d or 2h45m. Valid duration units are ms, s, m, h, d, w, y.").
 		Default("9999-12-31T23:59:59Z"))
 	ids := cmd.Flag("id", "Block to be replicated to the destination bucket. IDs will be used to match blocks and other matchers will be ignored. When specified, this command will be run only once after successful replication. Repeated field").Strings()
+	ignoreMarkedForDeletion := cmd.Flag("ignore-marked-for-deletion", "Do not replicate blocks that have deletion mark.").Bool()
```
GiedriusS, line 698: I think this should even be the default TBH, surprised that it is not :smile:

GiedriusS (commented):
Please rebase on top of newest `main`, those whitespace changes in docs/... are not needed.

cmd/thanos/tools_bucket.go
FUSAKLA: Well I was also a bit surprised  :D 
But didn't want to change the default behavior, but if we agree, I'd be happy to add `--replicated-marked-for-deletion` flag instead. Not sure if anyone could have a use case to replicated blocks that should be deleted.

Another thing is that it does not replicate the deletion marks (see https://github.com/thanos-io/thanos/pull/5118)

pkg/replicate/scheme_test.go
FUSAKLA: fixed here https://github.com/thanos-io/thanos/pull/5117/commits/53bab986162fa219d1e0f7761e7b5a8e61545076

pkg/replicate/scheme.go
FUSAKLA: Yep, I thought it's enough for replication just the fact that it's marked, didn't think of any use case for the delay here.

Debug log added here https://github.com/thanos-io/thanos/pull/5117/commits/69db471417ea9b6b9d6ef5baf8094da0f3d768cb

FUSAKLA:
Thanks, @GiedriusS for the swift review!

The white space changes were caused by the `make docs`, not sure why. Rolled it back.

FUSAKLA:
Passing through the docs test is a task on it's own I see :smile:

GiedriusS (approved):
LGTM, let's see for a few days if anyone has any other suggestions

cmd/thanos/tools_bucket.go
GiedriusS: Maybe let's keep the old behaviour for now :+1:

pkg/replicate/scheme.go
```diff
@@ -192,7 +198,11 @@ func (rs *replicationScheme) execute(ctx context.Context) error {
...
+		if err != nil && !rs.fromBkt.IsObjNotFoundErr(err) {
+			level.Debug(rs.logger).Log("msg", "failed to read deletion mark from origin bucket", "error", err)
```
wiardvanrij, line 203: Is this an error, error? Or is this more like 'it could be there, which is an acceptable use-case'? - I'm questioning if this then needs to be Debug log level, info or perhaps even being a fatal error then?

wiardvanrij:
Only got one question, other than that, LGTM! :)

pkg/replicate/scheme.go
FUSAKLA: Right, the level here is disputable. The semantics is "I was not able to tell if the block is marked as deleted" and the question is what should be the behavior in such case.

This way could happen that we replicate a block marked for deletion even if the flag was enabled.  
I don't have a strong opinion here, maybe it would be safer to bail out completely with an error instead of possibly doing something that configuration tends to forbid.

FUSAKLA: @GiedriusS WDYT?

GiedriusS: Mhm, yeah, maybe let's return an error instead of logging it? I remember there being a discussion about the same exact point but can't find anything now. Something along the lines of "crash early" :thinking: maybe we should add this to our code style document

FUSAKLA: Ok, agreed changed in https://github.com/thanos-io/thanos/commit/b96326a0c66943a712bbd8189b59dbe8ad264e04 to fail with an error

FUSAKLA:
Not sure why but the PR diff broke GitHub somehow :sweat_smile: 
Getting 500 on the changes page..

wiardvanrij (approved):
LGTM

PR #5118, opened 2022-02-03
tools bucket replicate: deletion marks

FUSAKLA:
If running replicate from an origin bucket with running compactor, old removed blocks get replicated but never marked as a deleted and once a new compacted block is synced this leads to overlapping blocks.

This PR adds synchronization of the deletion marks even in case the block is already synchronized (metadata JSON already there) because deletion mark is something that can change.

## Changes

 - `tools bucket replicate` now replicates event deletion marks

## Verification

 - Tests added

### DIFF
--- a/pkg/replicate/scheme.go
+++ b/pkg/replicate/scheme.go
@@ -230,6 +230,7 @@ func (rs *replicationScheme) ensureBlockIsReplicated(ctx context.Context, id uli
 	chunksDir := path.Join(blockID, thanosblock.ChunksDirname)
 	indexFile := path.Join(blockID, thanosblock.IndexFilename)
 	metaFile := path.Join(blockID, thanosblock.MetaFilename)
+	deletionMarkFile := path.Join(blockID, metadata.DeletionMarkFilename)
 
 	level.Debug(rs.logger).Log("msg", "ensuring block is replicated", "block_uuid", blockID)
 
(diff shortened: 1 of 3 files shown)


FUSAKLA:
Based on https://github.com/thanos-io/thanos/pull/5117, marking as a draft until solved

pkg/replicate/scheme.go
```diff
@@ -192,7 +198,11 @@ func (rs *replicationScheme) execute(ctx context.Context) error {
 	}
 
 	for id, meta := range metas {
-		if rs.blockFilter(meta) {
+		_, err := rs.fromBkt.ReaderWithExpectedErrs(rs.fromBkt.IsObjNotFoundErr).Get(ctx, path.Join(meta.ULID.String(), metadata.DeletionMarkFilename))
+		if err != nil && !rs.fromBkt.IsObjNotFoundErr(err) {
+			level.Debug(rs.logger).Log("msg", "failed to read deletion mark from origin bucket", "error", err)
```
wiardvanrij, line 203: Is this an error, error? Or is this more like 'it could be there, which is an acceptable use-case'? - I'm questioning if this then needs to be Debug log level, info or perhaps even being a fatal error then?

FUSAKLA:
So the ignoring of blocks marked for deletion is now merged, so this is also GTG if desired at all.

Issue #4831, opened 2021-11-03
Thanos Receive: "Out of bounds metric" 


Kampe:
Hello!

We run a handful of "edge" sites in a hub and spoke topology and feed metrics from remote promethei to a centralized receiver hash-ring.  We see issues pop up quite often relating to "Out Of Bounds" metrics popping up in receiver.  The prometheus remote write will begin to return 500s to prometheus and Reciever latency climbs, impacting other tenants. We're unsure if the metrics are too far in the future, or in the past, however we assume the past as it's likely we could have experienced network or latency issues with shipment of remote metrics to cloud. It seems this is related to TSDB and Prometheus WAL limitations that were amended in recent prometheus releases?  https://github.com/prometheus/prometheus/issues/8534

The shipping prometheus itself is setup with the prometheus-operator, and has been configured with `disableCompaction: true` to ensure it doesn't ship old blocks to Thanos, too (even though we'd LOVE (read: need) to have them...)

Worse?  When prometheus fails to remote write, it just keeps retrying, getting into a loop with the receiver until the block is discarded.  This is 100% not intended and there probably needs to be a feature aded on one side of the system here to prevent this (as data caps are a real thing and this gets into 100s of GBs quick if unchecked).

Related and similar: https://github.com/thanos-io/thanos/issues/4767 - However we certainly have unique labels on our items and rarely are we getting 409 conflicts, but constantly 500s.

Thanos Version: 0.22.0
Prometheus Version: 0.30.1

**Object Storage Provider**:
GCP

**What happened**:
Thanos Receive will on occasion get into a fit with the samples a prometheus sends up to it for storage throwing Out of Bounds errors, causing cascading issues for the whole Thanos ingestion of metrics globally within system.

**What you expected to happen**:
Thanos Receive accepts the samples and stores them in the bucket without issue.

**How to reproduce it (as minimally and precisely as possible)**:
We see this most with shipments of kube-state-metrics, however you can get any prometheus into this issue with a receiver if you wait long enough.

**Full logs to relevant components**:
```
component=receive component=receive-writer msg="Error on ingesting out-of-order samples" numDropped=47
[...]

PR #5131, opened 2022-02-06
[RECEIVE] Make out of bounds a non-recoverable error

wiardvanrij:
Signed-off-by: Wiard van Rij <wvanrij@roku.com>

Relates to: https://github.com/thanos-io/thanos/issues/4831

Prometheus' remote write receiver treats `ErrOutOfOrder` as non-recoverable.
https://github.com/prometheus/prometheus/issues/5649#issuecomment-970730016_

## Changes

Out of bounds would result in a http 409, which causes Prometheus to retry, eventually causing an infinite loop which can only be solved by flushing the entire Prometheus data dir. 

This change removes OutOfBounds err from ConflictErr to it's own entity and it will allow Prometheus to deal with this as non-recoverable (i.e. dropping it, instead of retry).

Also made a small adjustment in the Debug.Log, as it could not print the sample, which causes a confusing debug message like: `sample="unsupported value type"` - Now it will print the sample value and timestamp.

## Verification

I mean, i'll be honest that it was a pain in the a. to test and debug this. What I have checked is that Prometheus now handles this as non-recoverable and nothing will build up anymore as pending. 

I have adjusted the test cases and added the OutOfBounds for it.

### DIFF
--- a/pkg/receive/handler.go
+++ b/pkg/receive/handler.go
@@ -360,7 +360,7 @@ func (h *Handler) receiveHTTP(w http.ResponseWriter, r *http.Request) {
 		http.Error(w, err.Error(), http.StatusServiceUnavailable)
 	case errConflict:
 		http.Error(w, err.Error(), http.StatusConflict)
-	case storage.ErrOutOfBounds:
+	case errOutOfBounds:
 		http.Error(w, err.Error(), http.StatusBadRequest)
 	case errBadReplica:
 		http.Error(w, err.Error(), http.StatusBadRequest)
--- a/pkg/receive/handler.go
+++ b/pkg/receive/handler.go
@@ -59,8 +59,8 @@ const (
 
 var (
 	// errConflict is returned whenever an operation fails due to any conflict-type error.
-	errConflict = errors.New("conflict")
-
+	errConflict    = errors.New("conflict")
+	errOutOfBounds = storage.ErrOutOfBounds
 	errBadReplica  = errors.New("request replica exceeds receiver replication factor")
 	errNotReady    = errors.New("target not ready")
 	errUnavailable = errors.New("target not available")
@@ -360,6 +360,8 @@ func (h *Handler) receiveHTTP(w http.ResponseWriter, r *http.Request) {
 		http.Error(w, err.Error(), http.StatusServiceUnavailable)
 	case errConflict:
 		http.Error(w, err.Error(), http.StatusConflict)
+	case storage.ErrOutOfBounds:
+		http.Error(w, err.Error(), http.StatusBadRequest)
 	case errBadReplica:
 		http.Error(w, err.Error(), http.StatusBadRequest)
 	default:
@@ -679,7 +681,6 @@ func isConflict(err error) bool {
 	return err == errConflict ||
 		err == storage.ErrDuplicateSampleForTimestamp ||
 		err == storage.ErrOutOfOrderSample ||
-		err == storage.ErrOutOfBounds ||
 		status.Code(err) == codes.AlreadyExists
 }
 
@@ -694,6 +695,11 @@ func isNotReady(err error) bool {
 func isUnavailable(err error) bool {
 	return err == errUnavailable ||
 		status.Code(err) == codes.Unavailable
+}
+
+func isOutOfBounds(err error) bool {
+	return err == storage.ErrOutOfBounds ||
+		status.Code(err) == codes.OutOfRange
 }
 
 // retryState encapsulates the number of request attempt made against a peer and,
@@ -738,6 +744,7 @@ func determineWriteErrorCause(err error, threshold int) error {
 		{err: errConflict, cause: isConflict},
 		{err: errNotReady, cause: isNotReady},
 		{err: errUnavailable, cause: isUnavailable},
+		{err: errOutOfBounds, cause: isOutOfBounds},
 	}
 	for _, exp := range expErrs {
 		exp.count = 0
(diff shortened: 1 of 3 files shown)


Kampe:
I appreciate the time you took to find this case!

pkg/receive/handler.go
```diff
@@ -360,6 +360,8 @@ func (h *Handler) receiveHTTP(w http.ResponseWriter, r *http.Request) {
 		http.Error(w, err.Error(), http.StatusServiceUnavailable)
 	case errConflict:
 		http.Error(w, err.Error(), http.StatusConflict)
+	case errOutOfBounds:
+		http.Error(w, err.Error(), http.StatusBadRequest)
```
matej-g, line 364: Maybe I'm missing something here, but my understanding was that Prometheus treats as _recoverable_ only network errors, `5xx` responses and, if configured to, `429` responses (see [this line](https://github.com/prometheus/prometheus/blob/579331446a04503b2b1515767e7ceaceab2c6342/storage/remote/client.go#L213) and below).

It therefore should not make difference if we're responding with bad request (`400`) or  conflict (`409`), since conflict is already unrecoverable, right?

wiardvanrij: For Prometheus, 400 = unrecoverable. If that's your question? Also, if that's untrue, than I'm happy to get pointed to some resource. However, I've tested this and the response now becomes unrecoverable for Prometheus.

matej-g: As far as I understood, _only_ `5xx` codes are recoverable for Prometheus (see the link in my comment). Meaning if until now we were returning `409` for out of bounds error, it should have already been working as expected, i.e. there should not be a real difference between the codes. As long as we return any `4xx` code, it should be treated as unrecoverable (with the possible exception of `429`).

wiardvanrij: Right, I understand you and the code (missed the link) now. Funny thing is though, that I did _not_ experience this behaviour. Only after my change I actually see the non-recoverable errors in the Prometheus logs. 🤔

Kampe: More on this: https://github.com/prometheus/prometheus/pull/5649#issuecomment-970730016

kungfuchicken: looking forward to seeing this fixed. it's causing us pain in production.

matej-g: @wiardvanrij I did a small test today: I spun up a VM where I skewed the clock on purpose (made it run in the 'past' compared to my host machine), I brought up an instance of Prometheus, configured to remote write to Thanos instance running on my host. 

I first run it against Thanos build from latest `main`. The result - Prometheus informs me this is a non-recoverable error:
```
ts=2022-02-12T12:26:39.089Z caller=dedupe.go:112 component=remote level=error remote_name=6c3063 url=http://192.168.56.1:19291/api/v1/receive msg="non-recoverable error" count=9 exemplarCount=0 err="server returned HTTP status 409 Conflict: store locally for endpoint 127.0.0.1:10901: conflict"
```
Built from your branch:
```
ts=2022-02-12T12:34:32.266Z caller=dedupe.go:112 component=remote level=error remote_name=6c3063 url=http://192.168.56.1:19291/api/v1/receive msg="non-recoverable error" count=63 exemplarCount=0 err="server returned HTTP status 400 Bad Request: store locally for endpoint 127.0.0.1:10901: out of bounds"
```

I haven't noticed any unexpected behavior on either side (Thanos or Prometheus). Prometheus treated it (in both cases) simply as unrecoverable error.

wiardvanrij: thanks @matej-g  really appreciate the testing - Then I need more user input. @Kampe did you have any chance to test it or could you provide more info? Because if we now look at the code, it already should be correct. I would agree that we sometimes see odd behaviour but then we really need to know why that is happening. 

If I have some more time, i'll try to figure things out as well.

Kampe: I responded in the issue ticket https://github.com/thanos-io/thanos/issues/4831, but there definitely seems to be two issues happening with 500 level errors and 409's.  Unsure if they are related, but it seems we can on atleast a monthly cadence (sometimes weekly) can produce them.

We have not had the chance to roll the changes to any of our environments, we don't have much in the way of tooling setup to be able to roll non registry hosted third party tools to even developer environments so testing without a prebuilt container for us is pretty intensive, however when we have a container we can reference I can get taht rolled in a matter of minutes.

phillebaba:
@wiardvanrij do you still need some testing for this feature? I think I experience this issue with a specific tenant which causes the receiver to consume a lot more memory than usual during certain scenarios, to the point where it will eventually OOM kill. Looking at the logs it seems like the issue is consistent with the description as the Prometheus agents will continue to retry the same failed out of order requests over and over again.

wiardvanrij:
> @wiardvanrij do you still need some testing for this feature? I think I experience this issue with a specific tenant which causes the receiver to consume a lot more memory than usual during certain scenarios, to the point where it will eventually OOM kill. Looking at the logs it seems like the issue is consistent with the description as the Prometheus agents will continue to retry the same failed out of order requests over and over again.

sorry for the lag. 
The problem with this issue is that it's very hard to replicate and for some reason, I thought I had fixed it (as my tests were successful) but it does not make any sense given the input I've got here. 
I think we might just have to close this MR and follow up in the issue?

gabrielSoudry:
Can you merge this please ? The message "unsupported value type" is very confusing and we take a lot of time to figure out what's going on ... (And we still have a lot of 409).

Our solution : send less metrics...
Thanks you !

PR #5138, opened 2022-02-08
cache: fix quickstart script, improve checking

GiedriusS:
Check that SelfURL also doesn't end with a slash since the galaxycache's
code checks whether the strings are equivalent when picking peers, and
peers cannot have slashes at the end so we need the same, exact format
of the URLs.

Fix quickstart script so that it wouldn't have slashes at the end.

Idea for the future: let's add smoke tests for `scripts/quickstart.sh`
to ensure smooth developer experience?

### DIFF
--- a/pkg/cache/groupcache.go
+++ b/pkg/cache/groupcache.go
@@ -88,6 +88,9 @@ func parseGroupcacheConfig(conf []byte) (GroupcacheConfig, error) {
 			return GroupcacheConfig{}, fmt.Errorf("peer %d must not have a trailing slash (%s)", i, peer)
 		}
 	}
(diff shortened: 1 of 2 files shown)


matej-g (approved):
Looks good to me :+1: One alternative I could imagine is taking care of this for the user (removing trailing slash if one is detected)

GiedriusS:
> Looks good to me +1 One alternative I could imagine is taking care of this for the user (removing trailing slash if one is detected)

What if there are two or more trailing slashes? Maybe someone will have some weird configuration with reverse proxies in front of Thanos Store :thinking: :shrug:

niaurys (approved):

PR #5144, opened 2022-02-11
UI: improve graph color

Nexucis:
I'm reporting the improvement made by @suyashtava in Prometheus through: https://github.com/prometheus/prometheus/pull/10179

Note: After this PR we should be able to remove the old UI like we did in Prometheus \o/.


GiedriusS (approved):
Great stuff, thx! Thanks for helping us keep the UI in Thanos up to date :beers:

PR #5148, opened 2022-02-12
Receive: Added tenant ID tag for span

### LINKED ISSUE #5002: Receiver: instrument multitsdb tenant series tracing tag correctly

metonymic-smokey:
Fixes #5002 

## Changes
Added tenant tag for Receive spans.


metonymic-smokey:
@yeya24 please review it when you're free. 
Many thanks!

yeya24 (approved):
LGTM

metonymic-smokey:
Could you re-trigger the docs test? It's timing out.

PR #5153, opened 2022-02-14
Receive: option to extract tenant from client certificate

4xoc:

## Changes

This PR adds the option to extract the tenant label value from a supplied TLS client certificate. It takes priority over any header value sent. As such, it ensures that incoming remote_write requests are always put to the correct tenant. The attribute value from the cert can be one of organization, organizationalUnit or commonName. Running a private PKI for remote_write is highly recommended to control the values of the certificate. However, any client certificate will suffice as long as it is valid. If extracting the tentant from the certificate fails, the write request is denied with a 400 status code.
This feature is disabled by default.

## Verification

Running this change for some time now in a test environment without any issues.

### DIFF
--- a/cmd/thanos/receive.go
+++ b/cmd/thanos/receive.go
@@ -191,7 +191,7 @@ func runReceive(
 		Registry:          reg,
 		Endpoint:          conf.endpoint,
 		TenantHeader:      conf.tenantHeader,
-		TenantAttribute:   conf.tenantAttribute,
+		TenantField:       conf.tenantField,
 		DefaultTenantID:   conf.defaultTenantID,
 		ReplicaHeader:     conf.replicaHeader,
 		ReplicationFactor: conf.replicationFactor,
@@ -709,7 +709,7 @@ type receiveConfig struct {
 	refreshInterval   *model.Duration
 	endpoint          string
 	tenantHeader      string
-	tenantAttribute   string
+	tenantField       string
 	tenantLabelName   string
 	defaultTenantID   string
 	replicaHeader     string
@@ -773,7 +773,7 @@ func (rc *receiveConfig) registerFlag(cmd extkingpin.FlagClause) {
 
 	cmd.Flag("receive.tenant-header", "HTTP header to determine tenant for write requests.").Default(receive.DefaultTenantHeader).StringVar(&rc.tenantHeader)
 
-	cmd.Flag("receive.tenant-certificate-attribute", "Use TLS client certificate's attribute to determine tenant for write requests. Must be one of organization, organizationalUnit or commonName.").Default("").StringVar(&rc.tenantAttribute)
+	cmd.Flag("receive.tenant-certificate-field", "Use TLS client's certificate field to determine tenant for write requests. Must be one of "+receive.CertificateFieldOrganization+", "+receive.CertificateFieldOrganizationalUnit+" or "+receive.CertificateFieldCommonName+". This setting will cause the receive.tenant-header flag value to be ignored.").Default("").EnumVar(&rc.tenantField, "", receive.CertificateFieldOrganization, receive.CertificateFieldOrganizationalUnit, receive.CertificateFieldCommonName)
 
 	cmd.Flag("receive.default-tenant-id", "Default tenant ID to use when none is provided via a header.").Default(receive.DefaultTenant).StringVar(&rc.defaultTenantID)
 
--- a/cmd/thanos/receive.go
+++ b/cmd/thanos/receive.go
@@ -191,6 +191,7 @@ func runReceive(
 		Registry:          reg,
 		Endpoint:          conf.endpoint,
 		TenantHeader:      conf.tenantHeader,
+		TenantAttribute:   conf.tenantAttribute,
 		DefaultTenantID:   conf.defaultTenantID,
 		ReplicaHeader:     conf.replicaHeader,
 		ReplicationFactor: conf.replicationFactor,
@@ -708,6 +709,7 @@ type receiveConfig struct {
 	refreshInterval   *model.Duration
 	endpoint          string
 	tenantHeader      string
+	tenantAttribute   string
 	tenantLabelName   string
 	defaultTenantID   string
 	replicaHeader     string
@@ -770,6 +772,8 @@ func (rc *receiveConfig) registerFlag(cmd extkingpin.FlagClause) {
 	cmd.Flag("receive.local-endpoint", "Endpoint of local receive node. Used to identify the local node in the hashring configuration.").StringVar(&rc.endpoint)
 
 	cmd.Flag("receive.tenant-header", "HTTP header to determine tenant for write requests.").Default(receive.DefaultTenantHeader).StringVar(&rc.tenantHeader)
+
+	cmd.Flag("receive.tenant-certificate-attribute", "Use TLS client certificate's attribute to determine tenant for write requests. Must be one of organization, organizationalUnit or commonName.").Default("").StringVar(&rc.tenantAttribute)
 
 	cmd.Flag("receive.default-tenant-id", "Default tenant ID to use when none is provided via a header.").Default(receive.DefaultTenant).StringVar(&rc.defaultTenantID)
 
--- a/pkg/receive/handler.go
+++ b/pkg/receive/handler.go
@@ -848,25 +848,22 @@ func (h *Handler) getTenantFromCertificate(r *http.Request) (string, error) {
 	switch h.options.TenantField {
 
 	case CertificateFieldOrganization:
-		if len(cert.Subject.Organization) > 0 {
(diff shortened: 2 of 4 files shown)


cmd/thanos/receive.go
```diff
@@ -767,6 +769,8 @@ func (rc *receiveConfig) registerFlag(cmd extkingpin.FlagClause) {
 
 	cmd.Flag("receive.tenant-header", "HTTP header to determine tenant for write requests.").Default(receive.DefaultTenantHeader).StringVar(&rc.tenantHeader)
 
+	cmd.Flag("receive.tenant-certificate-attribute", "Use TLS client certificate's attribute to determine tenant for write requests. Must be one of organization, organizationalUnit or commonName.").Default("").StringVar(&rc.tenantAttribute)
+
```
matej-g, line 773: I think we could use `EnumVar` instead of `StringVar` here, since there are predetermined options possible only.

matej-g (commented):
Thanks for the contribution @4xoc, I'd say this is looking good! :+1: One suggestion.

cmd/thanos/receive.go
4xoc: Good point, updated the flag type.

matej-g (dismissed):

pkg/receive/handler.go
```diff
@@ -339,6 +340,51 @@ func (h *Handler) receiveHTTP(w http.ResponseWriter, r *http.Request) {
 		tenant = h.options.DefaultTenantID
 	}
 
+	// Checking certs for possible value.
```
squat, line 343: I'd say this whole block and the 5 lines above should all be extracted into their own function like
`func (h *Handler) findTenant(r *http.Request) str`

pkg/receive/handler.go
```diff
@@ -339,6 +340,51 @@ func (h *Handler) receiveHTTP(w http.ResponseWriter, r *http.Request) {
...
+				}
+
+			default:
+				// Unknown/unsupported attribute requested, can't continue.
+				level.Error(h.logger).Log("err", err, "msg", "tls client cert attribute requested is not supported")
+				http.Error(w, err.Error(), http.StatusInternalServerError)
+				return
+			}
+
+		} else {
```
squat, line 382: let's invert this and make an early return instead, for readability:
```go
if len(r.TLS.PeerCertificates) == 0 {
	http.Error(w, "could not get required certificate attribute from client cert", http.StatusBadRequest)
	return
}
```

cmd/thanos/receive.go
```diff
@@ -767,6 +769,8 @@ func (rc *receiveConfig) registerFlag(cmd extkingpin.FlagClause) {
 
 	cmd.Flag("receive.tenant-header", "HTTP header to determine tenant for write requests.").Default(receive.DefaultTenantHeader).StringVar(&rc.tenantHeader)
 
+	cmd.Flag("receive.tenant-certificate-attribute", "Use TLS client certificate's attribute to determine tenant for write requests. Must be one of organization, organizationalUnit or commonName.").Default("").EnumVar(&rc.tenantAttribute, "", "organization", "organizationalUnit", "commonName")
```
squat, line 772: In the x509 world these are called `field`s, so let's use this naming convention
```suggestion
	cmd.Flag("receive.tenant-certificate-field", "Use TLS client certificate's attribute to determine tenant for write requests. Must be one of organization, organizationalUnit or commonName.").Default("").EnumVar(&rc.tenantAttribute, "", "organization", "organizationalUnit", "commonName")
```

squat, line 772:
```suggestion
	cmd.Flag("receive.tenant-certificate-attribute", "The field of the client certificate to use to determine the tenant ID. Must be one of organization, organizationalUnit or commonName.").Default("").EnumVar(&rc.tenantField, "", "organization", "organizationalUnit", "commonName")
```

squat, line 772: We really should document that this takes precedence over the tenant header flags if both are provided.

pkg/receive/handler.go
```diff
@@ -339,6 +340,51 @@ func (h *Handler) receiveHTTP(w http.ResponseWriter, r *http.Request) {
...
+			// First cert is the leaf authenticated against.
+			cert := r.TLS.PeerCertificates[0]
+
+			switch h.options.TenantAttribute {
+
+			case "organization":
+				if len(cert.Subject.Organization) > 0 {
+					tenant = cert.Subject.Organization[0]
+				} else {
+					http.Error(w, "could not get organization attribute from client cert", http.StatusBadRequest)
```
squat, line 355: All of these `else`s break the soft tenancy provided by allowing un-specified tenants to be ingested by a "default" hashring. This should be correctly handled to provide feature parity with the HTTP header approach IMO

pkg/receive/handler.go
```diff
@@ -339,6 +340,51 @@ func (h *Handler) receiveHTTP(w http.ResponseWriter, r *http.Request) {
...
 
+	// Checking certs for possible value.
+	if h.options.TenantAttribute != "" {
+		if len(r.TLS.PeerCertificates) > 0 {
+			// First cert is the leaf authenticated against.
+			cert := r.TLS.PeerCertificates[0]
+
+			switch h.options.TenantAttribute {
+
+			case "organization":
```
squat, line 351: Let's turn these into an enum to make them:
1. easier to discover the options for code using the package and
2. harder to mess-up with a typo.

squat (changes requested):
Hi @4xoc 
Thanks for this proposal. Generally I think this could be a nice idea. Please take a look at my concerns so we can get the PR into mergable shape.

pkg/receive/handler.go
4xoc: The whole idea is to limit tenant values to a trusted source and as such this needs to hard-fail. It's only so when enabling the feature, which I think is a reasonable thing to do. It ensures that no tenant could send data for a different tenant due to misconfiguration or intentional attack on a receive cluster. This can certainly be documented better which I will take care of. Same goes for it beating the header value in priority.

4xoc:
Hey @squat, would you mind having a look at the changes I've done to address your suggestions?

pkg/receive/handler.go
```diff
@@ -799,3 +816,45 @@ func (p *peerGroup) get(ctx context.Context, addr string) (storepb.WriteableStor
...
+		return "", errors.New("could not get required certificate field from client cert")
+	}
+
+	// First cert is the leaf authenticated against.
+	cert := r.TLS.PeerCertificates[0]
+
+	switch h.options.TenantField {
+
+	case CertificateFieldOrganization:
+		if len(cert.Subject.Organization) > 0 {
```
squat, line 835: Mit: consider inverting all of these `if`s to be `if len(...) == 0` so that we return early and this avoid the `else`

4xoc:
> Also, as mentioned here [#5153 (comment)](https://github.com/thanos-io/thanos/pull/5153#discussion_r811229465) do you want to document the change in soft-tenancy behavior in this PR or in a follow-up?

I working already on another feature for whitelisting tenant values. In combination with this PR, it would complete a hard tenancy enforcement I'm aiming for. With that new feature I'll update the docs about the changes in (soft) tenancy as it requires more documentation anyway.

squat:
Sounds fine, in that case, please clean up the if-elses as mentioned and we can merge :)

4xoc:
Hi @squat, fixed the merge conflict and rebased, anything else missing to get this PR merged?

matej-g (commented):
I'm happy to approve this as well, just the docs check failure seems related (yes, a bit annoying after all that time). Once we have this fixed, I'm sure @squat will be happy to approve as well. Thanks!

pkg/receive/handler.go
```diff
@@ -810,3 +828,42 @@ func (p *peerGroup) get(ctx context.Context, addr string) (storepb.WriteableStor
 	p.cache[addr] = client
 	return client, nil
 }
+
+// GetTenantFromCertificate extracts the tenant value from a client's presented certificate. The x509 field to use as
```
matej-g, line 832: nit: To match method name and comment
```suggestion
// getTenantFromCertificate extracts the tenant value from a client's presented certificate. The x509 field to use as
```

matej-g (approved):

4xoc:
Ah, that check didn't work for a long time at all. Here's the update for it, as previously mentioned, I'll update the docs more about the tenant enforcing feature this PR is part of.

wiardvanrij (approved):
LGTM

matej-g (approved):
:tada:

matej-g:
@squat Could we ask to get approval for this?

squat (approved):
Sorry for the delay. Looks fine 🎉
I would like to see some unit and/or e2e tests for this feature since it affects the core of receive but i won't block.

PR #5162, opened 2022-02-15
fix tools bucket replicate: fix ignoring of deleted blocks

FUSAKLA:
Even though ignoring of deleted blocks by the `tools bucket replicate`  added in https://github.com/thanos-io/thanos/pull/5117 works

It does not use the code base for filtering already prepared and that brings one additional drawback and that is it fails on reading block metadata if in process of deletion (which was also the case we were hitting)

The former implementation first read all meta files (which fails for the block that is being deleted) and only after checks for the deletion mark. The deletion mark has to be checked first, as the filtered meta fetcher does.

Sorry about that

Please, @GiedriusS and @wiardvanrij would you take a look one more time? :pray:

### DIFF
--- a/pkg/replicate/replicator.go
+++ b/pkg/replicate/replicator.go
@@ -251,7 +251,7 @@ func newMetaFetcher(
 	if ignoreMarkedForDeletion {
 		filters = append(filters, thanosblock.NewIgnoreDeletionMarkFilter(logger, fromBkt, 0, concurrency))
 	}
-	fetcher, err := thanosblock.NewMetaFetcher(
+	return thanosblock.NewMetaFetcher(
 		logger,
 		concurrency,
 		fromBkt,
@@ -259,8 +259,4 @@ func newMetaFetcher(
 		reg,
 		filters,
 	)
-	if err != nil {
-		return nil, err
-	}
-	return fetcher, nil
 }
--- a/pkg/replicate/replicator.go
+++ b/pkg/replicate/replicator.go
@@ -10,8 +10,6 @@ import (
 	"strings"
 	"time"
 
-	"github.com/thanos-io/thanos/pkg/objstore"
-
 	extflag "github.com/efficientgo/tools/extkingpin"
 	"github.com/go-kit/log"
 	"github.com/go-kit/log/level"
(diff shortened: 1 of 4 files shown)


pkg/replicate/replicator.go
```diff
@@ -243,3 +235,24 @@ func RunReplicate(
 
 	return nil
 }
+
+func newMetaFetcher(logger log.Logger, fromBkt objstore.InstrumentedBucket, reg prometheus.Registerer, minTime, maxTime thanosmodel.TimeOrDurationValue, concurrency int, ignoreMarkedForDeletion bool) (*thanosblock.MetaFetcher, error) {
```
FUSAKLA, line 239: Abstracted to separate func so it can be used in tests

metalmatze: Thanos uses multiple lines for that many parameters, I think? Maybe you can add some line breaks. :relaxed:

pkg/replicate/scheme_test.go
```diff
@@ -14,19 +14,27 @@ import (
 	"testing"
 	"time"
 
+	"github.com/thanos-io/thanos/pkg/model"
```
metalmatze, line 17: Needs to be sorted correctly.

pkg/replicate/replicator.go
FUSAKLA: done!

yeya24: I don't get why this helps in testing. Can you please elaborate more?

pkg/replicate/replicator.go
```diff
@@ -243,3 +235,32 @@ func RunReplicate(
...
+	concurrency int,
+	ignoreMarkedForDeletion bool,
+) (*thanosblock.MetaFetcher, error) {
+	filters := []thanosblock.MetadataFilter{
+		thanosblock.NewTimePartitionMetaFilter(minTime, maxTime),
+	}
+	if ignoreMarkedForDeletion {
+		filters = append(filters, thanosblock.NewIgnoreDeletionMarkFilter(logger, fromBkt, 0, concurrency))
+	}
+	fetcher, err := thanosblock.NewMetaFetcher(
```
yeya24, line 254: We can just return `thanosblock.NewMetaFetcher(...)`

pkg/replicate/replicator.go
FUSAKLA: Well I don't say it's, much but at least it abstracts the creation of the fetcher based on the input flags config and the initialization and configuration of the fetcher filters is the same in the tests and real code path.

pkg/replicate/replicator.go
FUSAKLA: Right, the error was previously wrapped which was the reason it was not returned right a way, but I eventually wrapped it where the func is called and forget about this. Thanks

pkg/replicate/replicator.go
yeya24: I see. That sounds good.

PR #5167, opened 2022-02-17
Docs: Tidy up docs-related make targets

matej-g:

## Changes

Make targets to check and / or generate docs can be confusing since:
1) We have target called `check-docs` which however also generates and formats docs
2) We have `docs` target, which generates docs but this target is unused at the moment

The suggested way to improve this is:
1) Make `docs` target generate and format docs correctly, including removing white noise.
2) Make `check-docs` target check generated docs against discrepancies in formatting, links and white noise.

Also fixes CHANGELOG formatting issue that is currently in `main`.

## Verification

Ran the targets locally.


hanjm:
LGTM

saswatamcode (approved):
This looks good to me! 👍

.github/workflows/docs.yaml
```diff
@@ -28,6 +28,6 @@ jobs:
           key: ${{ runner.os }}-go-${{ hashFiles('**/go.sum') }}
 
       - name: Check docs
-        run: make check-docs
+        run: make docs
```
squat, line 31: I'm a bit worried about this name change because it seems inconsistent. I think that `check-*` is more in line with our other make targets, for example the one below, which are designed to exit non-zero when something is wrong

matej-g: You are right, this does not fit nicely with other targets. I changed this to:
- Have `docs` target correctly generate and format the docs, including white noise handling
- Keep `check-docs` as a target to check against discrepancies and exiting with error if some are found

GiedriusS (approved):
:muscle:

PR #5180, opened 2022-02-23
store: Avoid 404 errors on deletion-mark.json

garc33:

## Changes

Check deletion-mark.json meta file exists before to retrieve it in order to avoid http 404 issues.

### DIFF
--- a/pkg/block/metadata/markers.go
+++ b/pkg/block/metadata/markers.go
@@ -91,6 +91,13 @@ func (n *NoCompactMark) markerFilename() string { return NoCompactMarkFilename }
 // ReadMarker reads the given mark file from <dir>/<marker filename>.json in bucket.
 func ReadMarker(ctx context.Context, logger log.Logger, bkt objstore.InstrumentedBucketReader, dir string, marker Marker) error {
 	markerFile := path.Join(dir, marker.markerFilename())
+	ok, err := bkt.Exists(ctx, markerFile)
+	if err != nil {
+		return errors.Wrapf(err, "file exists: %s", metaFile)
+	}
+	if (!ok) {
+		return ErrorMarkerNotFound
+	}
 	r, err := bkt.ReaderWithExpectedErrs(bkt.IsObjNotFoundErr).Get(ctx, markerFile)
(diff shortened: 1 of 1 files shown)


GiedriusS (commented):
Could you please elaborate on what do you mean by
```
it in order to avoid http 404 issues.
```
? What are "HTTP 404 issues"?

pkg/block/metadata/markers.go
```diff
@@ -91,6 +91,13 @@ func (n *NoCompactMark) markerFilename() string { return NoCompactMarkFilename }
...
 func ReadMarker(ctx context.Context, logger log.Logger, bkt objstore.InstrumentedBucketReader, dir string, marker Marker) error {
 	markerFile := path.Join(dir, marker.markerFilename())
+	ok, err := bkt.Exists(ctx, markerFile)
+	if err != nil {
+		return errors.Wrapf(err, "file exists: %s", metaFile)
+	}
+	if (!ok) {
+		return ErrorMarkerNotFound
+	}
 	r, err := bkt.ReaderWithExpectedErrs(bkt.IsObjNotFoundErr).Get(ctx, markerFile)
```
GiedriusS, line 101: The code here should handle this, no? :thinking:

pkg/block/metadata/markers.go
```diff
@@ -91,6 +91,13 @@ func (n *NoCompactMark) markerFilename() string { return NoCompactMarkFilename }
 // ReadMarker reads the given mark file from <dir>/<marker filename>.json in bucket.
 func ReadMarker(ctx context.Context, logger log.Logger, bkt objstore.InstrumentedBucketReader, dir string, marker Marker) error {
 	markerFile := path.Join(dir, marker.markerFilename())
+	ok, err := bkt.Exists(ctx, markerFile)
+	if err != nil {
+		return errors.Wrapf(err, "file exists: %s", metaFile)
```
akanshat, line 96: I think you need to update the name from `metaFile` to `markerFile` to avoid the errors here.

akanshat (changes requested):

PR #5181, opened 2022-02-24
E2E Tests: Improve Prometheus configurability and make query tests more resilient

matej-g:

## Changes

I have noticed that some of the query E2E tests are prone to flakiness, I've seen a couple CI failures (e.g. https://github.com/thanos-io/thanos/runs/5262364980?check_suite_focus=true#step:5:383) as well as failed runs locally.

The tests fail due to an out of bounds error upon trying to remote write, excerpt:
```
=== CONT  TestSidecarStorePushdown
Error:     query_test.go:702: query_test.go:702:
        
         unexpected error: server returned HTTP status 400 Bad Request: out of bounds
        
03:21:07 prometheus-p1: level=error ts=2022-02-20T03:21:07.379Z caller=write_handler.go:53 component=web msg="Out of order sample from remote write" err="out of bounds"
```

I suspect this is due to the fact that in some test cases, we're remote writing time series with timestamps in the past. However, since with the current default Prometheus configuration, Prometheus is configured to scrape itself as well. I suspect that on occasion, if Prometheus is scraped before remote write, this makes the minimum valid time to be set to 'now', which causes the remote write requests with past timestamps to return out of bounds error.

The change to mitigate here is to make the self-scraping of the test Prometheus instance optional, as for these mentioned test cases this might interfere with remote write requests. Additionally, I moved the Prometheus config method to `e2ethanos` package from the query test file, since it's used across multiple E2E tests and fits better there. Lastly, I bumped slightly the wait time for `minio` to be ready, since I'm still seeing on occasion error when `minio` is not ready to accept requests.

## Verification

Ran E2E locally multiple times with success (as opposed to hitting out of bounds error before).


test/e2e/e2ethanos/services.go
```diff
@@ -856,7 +856,7 @@ func NewMinio(env e2e.Environment, name, bktName string) (*e2e.InstrumentedRunna
 			Command: e2e.NewCommandWithoutEntrypoint("sh", "-c", fmt.Sprintf(strings.Join(commands, " && "), minioKESGithubContent, minioKESGithubContent, bktName, 8090)),
 			//TODO(@clyang82): This is a temporary workaround for https://github.com/efficientgo/e2e/issues/9
 			//Readiness: e2e.NewHTTPReadinessProbe("http", "/minio/health/ready", 200, 200),
-			Readiness: e2e.NewCmdReadinessProbe(e2e.NewCommand("sh", "-c", "sleep 1 && curl -k https://127.0.0.1:8090/minio/health/ready")),
+			Readiness: e2e.NewCmdReadinessProbe(e2e.NewCommand("sh", "-c", "sleep 2 && curl -k https://127.0.0.1:8090/minio/health/ready")),
```
hanjm, line 859: Good catch. 
I think we can make this readiness probe use a loop retry with timeout. :)
```
curl --retry 5 --retry-delay 1
```

test/e2e/e2ethanos/services.go
```diff
@@ -1026,3 +1026,66 @@ func NewS3Config(bucket, endpoint, basePath string) s3.Config {
...
 }
+
+// NOTE: by using aggregation all results are now unsorted.
+var QueryUpWithoutInstance = func() string { return "sum(up) without (instance)" }
+
+// e2ethanos.DefaultPromConfig returns Prometheus config that sets Prometheus to:
+// * expose 2 external labels, source and replica.
+// * optionallly scrape self. This will produce up == 0 metric which we can assert on.
+// * optionally remote write endpoint to write into.
+func DefaultPromConfig(name string, replica int, remoteWriteEndpoint, ruleFile string, scrapeSelf bool, scrapeTargets ...string) string {
```
hanjm, line 1037: golang document shoud be follow this format:
// DefaultPromConfig is xxx.
https://go.dev/doc/effective_go#commentary

matej-g: Good catch, I missed this during find-and-replace, let me fix it

saswatamcode (dismissed):
Thanks for tackling test flakiness! 🙂

test/e2e/e2ethanos/services.go
```diff
@@ -1026,3 +1026,66 @@ func NewS3Config(bucket, endpoint, basePath string) s3.Config {
...
 }
+
+// NOTE: by using aggregation all results are now unsorted.
+var QueryUpWithoutInstance = func() string { return "sum(up) without (instance)" }
+
+// DefaultPromConfig returns Prometheus config that sets Prometheus to:
+// * expose 2 external labels, source and replica.
+// * optionallly scrape self. This will produce up == 0 metric which we can assert on.
+// * optionally remote write endpoint to write into.
+func DefaultPromConfig(name string, replica int, remoteWriteEndpoint, ruleFile string, scrapeSelf bool, scrapeTargets ...string) string {
```
bwplotka, line 1037: Maybe we could play with something like 

`DefaultPromConfig(name string, replica int, remoteWriteEndpoint, ruleFile string, scrapeTargets ...string)` and have constant for `const LocalPrometheusScrape string = "localhost:9090"`? Not a biggie (:

bwplotka (approved):
👍🏽  thanks! just minor optional nit - up to you

test/e2e/e2ethanos/services.go
matej-g: I like this idea much more :+1: Better than adding new parameter. Changed.

GiedriusS (dismissed):
:+1: let's fix the conflicts

matej-g:
Done, should be good to go now!

GiedriusS (approved):

Issue #5045, opened 2022-01-08
Bucket UI: finding overlapping blocks doesn't work as expected


yeya24:
**Thanos, Prometheus and Golang version used**: v0.24.0

**Object Storage Provider**: S3

**What happened**:

The thanos compactor halted and it found some overlapping blocks.

![image](https://user-images.githubusercontent.com/25150124/148662527-b7584de1-9043-4958-8147-9cc877f9b025.png)

Then I checked the bucket UI, indeed the two blocks are overlapping.

![image](https://user-images.githubusercontent.com/25150124/148662535-8b5f53de-04ed-48bd-99eb-3c258c486c9d.png)

![image](https://user-images.githubusercontent.com/25150124/148662543-c675ee47-8054-4d51-b16e-fe0fc0e29301.png)

![image](https://user-images.githubusercontent.com/25150124/148662549-742566b1-397a-45bd-bc2b-acc759818d8b.png)

If I enable `finding overlapping blocks` option, it doesn't work and I cannot find anything in the UI.

**What you expected to happen**:

If I click `finding overlapping blocks`, I can see all the overlapping blocks. If there is no overlapping block, the UI can display a message that there is no overlapping now.

tend2infinity:
Hey @yeya24 I'll work on this

matej-g:
Go for it @tend2infinity :rocket:

tend2infinity:
Hey @yeya24, I would really like to how I can recreate your condition because for me the overlapping blocks checkbox works absolutely alright. I tried uploading duplicate blocks manually by changing the ULID and playing a little bit with min-time and max-time. Here are some screenshots of what I was testing -

![Screenshot from 2022-02-15 16-21-41](https://user-images.githubusercontent.com/61948033/154047878-d9ba65e5-6220-4103-ad25-0e4c38e14233.png)

![Screenshot from 2022-02-15 16-22-00](https://user-images.githubusercontent.com/61948033/154047942-db9ce397-0df7-4d63-ae5c-71e5cad504ce.png)

yeya24:
@tend2infinity Thanks for the work. Sorry I should provide a reproducible example for you to help you work on this at the beginning. Right now I don't think I have cases to reproduce.

> If there is no overlapping block, the UI can display a message that there is no overlapping now.
Can you please try to implement this feature instead? This one seems more straightforward. We can see if the bug itself is reproducible or not later.

yeya24:
Still valid

[...]

PR #5182, opened 2022-02-24
added edge case for no overlapping blocks

tend2infinity:
Signed-off-by: soumya <somu12.ss@gmail.com>

### Closes 
Issue #5045

## Changes

- Added edge case for showing no overlapping blocks

- Removed crashing bug in the search bar when we tick enable finding overlapping blocks

## Verification

Tested locally

### Screenshots
Screenshot from 2022-02-24 07-27-33
Screenshot from 2022-02-24 07-37-10


GiedriusS (commented):
Could you please add a unit test for this buggy case so that we would avoid running into this in the future?

tend2infinity:
Working on it

tend2infinity:
relevant

Issue #5026, opened 2022-01-04
Do not set Thanos Query as ready before first service discovery


GiedriusS:
### Is your proposal related to a problem?

After a restart, Thanos Query starts up with no store nodes connected to it before the first iteration of service discovery succeeds. During this time, Thanos Query is "ready" which means that traffic flows to it even though it knows about zero store API nodes. Thus, "no data" responses get cached.

### Describe the solution you'd like

The query component should only report its status as "ready" once it has done at least one iteration of service discovery (successfully or not).

### Describe alternatives you've considered

Writing automation around this i.e. some external program should wait for one service discovery iteration to happen before putting Thanos Query back into the list of nodes ready to receive traffic e.g. into a load-balancer. But, that's cumbersome.

### Additional context

Affected >v0.24.0.

tend2infinity:
I would like to work on this issue

matej-g:
Hey @tend2infinity go for it :star2:

tend2infinity:
Hey @GiedriusS @matej-g , So I tried to get around this problem using following two approaches, unfortunately neither of them gave me the desired results, My explanation maybe a bit vague, I would really love to explain it in depth in some community hours meet, but tilll then can you suggest me some iterations in my approach or maybe some different approach!

- Declaring a channel in query.go and using it to write some value in it when we successfully run our first iteration of service discovery . Since writing some value to a channel is a blocking action and it will block the further execution of that goRoutine until we read the value present inside the channel in some other goRoutine, I was able to use this in my favour by receiving the value  in that particular goRoutine where we are making out statusProber READY. 

- In this way I was able to detect when the very first iteration of service discovery gets completed and my scheduler waits until I get the iteration and after that, I can easily turn my querier as "ready".

- The problem with this approach, I end up in a deadlock because I'm not able to read all the values I write to the channel, that's why I feel my approach till now isn't apt.

- Another approach in which I tried to run the update endpoint and service discovery functions only once, instead of running it continuously in fixed intervals as it is written in query.go. 

- I wanted to run it once and then use g.Run() to run all the goRoutines completely so that my first iteration should complete, then I would let all the goRoutines run simultaneously in regular intervals, as its written in the codebase.

- The problem with this approach is, my context gets destroyed if I don't use runUtil.repeat(), and once my context is destroyed, the whole querier exits. I'm attaching a screenshot below.
[...]

PR #5183, opened 2022-02-24
Making the querier ready after first iteration of service discovery

tend2infinity:
Signed-off-by: soumya <somu12.ss@gmail.com>

### Closes
Issue #5026

## Changes

- Running a single iteration of service discovery before making the querier ready

## Verification
Tested locally

### Screenshots
I am attaching a screenshot of my command line logs -
Screenshot from 2022-02-25 01-39-09

### DIFF
--- a/cmd/thanos/query.go
+++ b/cmd/thanos/query.go
@@ -690,7 +690,15 @@ func runQuery(
 			grpcserver.WithMaxConnAge(grpcMaxConnAge),
 		)
 
+		firstServiceDiscovery := true
+
 		g.Add(func() error {
+			if firstServiceDiscovery {
+				ctx, cancel := context.WithCancel(context.Background())
+				endpoints.Update(ctx)
+				cancel()
+				firstServiceDiscovery = false
+			}
 			statusProber.Ready()
 			return s.ListenAndServe()
 		}, func(error) {


cmd/thanos/query.go
```diff
@@ -690,7 +690,15 @@ func runQuery(
 			grpcserver.WithMaxConnAge(grpcMaxConnAge),
 		)
 
+		firstServiceDiscovery := true
+
 		g.Add(func() error {
+			if firstServiceDiscovery {
+				ctx, cancel := context.WithCancel(context.Background())
+				endpoints.Update(ctx)
```
GiedriusS, line 698: The problem is that the endpoints will be updated twice on boot. :( Why not use channels for this? Something like:

```go
initialDiscovery := make(chan int)

// Add this where the update happens.
g.Add(func() error {
   ...
   endpoints.Update(ctx)
   initialDiscovery <- 1
   close(initialDiscovery)
   ...
}, ...)

// Then add this just before statusProber.Ready()
<-initialDiscovery
```

That should do it, right?

Here's an example: https://go.dev/play/p/-vKlpjXKpCk.

GiedriusS (changes requested):

cmd/thanos/query.go
tend2infinity: I thought about doing this only, the problem is we are updating our endpoints every 5 seconds, take a look at the function where we are updating in query.go and my comments,

```
// Periodically update the store set with the addresses we see in our cluster.
	{
		g.Add(func() error {
			return runutil.Repeat(5*time.Second, ctx.Done(), func() error { 
// if we write a value to a channel inside this function and close it afterwards, means we actually do it every 5 seconds, and obviously a channel cannot be closed more than once, will lead to a deadlock
				endpoints.Update(ctx)
				return nil
			})
		}, func(error) {
			cancel()
			endpoints.Close()
		})
	}
```
Moreover, if we write something to the channel it will block that goroutine until some other goroutine reads it from the channel, Here we will read the value from the channel only once, since we do statusProber.Ready() only once, while we write in it after every 5 seconds, which again leads to a deadlock

GiedriusS: So, maybe we could use a variable there to avoid trying to close the channel twice? :thinking: We would still avoid the problem of calling Update() twice on startup.

matej-g: Or how about 1) do and wait on the first service discovery; 2) once the first service discovery is run, only then we kick off the periodical updates logic and unblock server from starting to listen. Does that make sense?

tend2infinity:
Working on it

Issue #5189, opened 2022-02-28
query: ExemplarQueryResult error


Reamer:
**Thanos, Prometheus and Golang version used**:

Thanos: quay.io/thanos/thanos:v0.24.0 - go1.16.12
Prometheus: 2.29.2 - go1.16.6
Grafana: grafana/grafana:8.4.2

**What happened**:
I see the following error in the Grafana log
```
{"err":"[]v1.ExemplarQueryResult: decode slice: expect [ or n, but found \u0000, error found in #0 byte of ...||..., bigger context ...||...","logger":"tsdb.prometheus","lvl":"eror","msg":"Exemplar query failed","query":"process_cpu_usage{namespace=\"spark-test\"}","t":"2022-02-28T09:07:46.63+0000"}
{"err":"[]v1.ExemplarQueryResult: decode slice: expect [ or n, but found \u0000, error found in #0 byte of ...||..., bigger context ...||...","logger":"tsdb.prometheus","lvl":"eror","msg":"Exemplar query failed","query":"sum(zeppelin_note_cache_hit_total{namespace=\"spark-test\"}) / (sum(zeppelin_note_cache_hit_total{namespace=\"spark-test\"}) + sum(zeppelin_note_cache_miss_total{namespace=\"spark-test\"}))","t":"2022-02-28T09:07:46.73+0000"}
{"err":"[]v1.ExemplarQueryResult: decode slice: expect [ or n, but found \u0000, error found in #0 byte of ...||..., bigger context ...||...","logger":"tsdb.prometheus","lvl":"eror","msg":"Exemplar query failed","query":"process_files_open_files{namespace=\"spark-test\"}","t":"2022-02-28T09:07:46.73+0000"}

```

**What you expected to happen**:

No erros

**How to reproduce it (as minimally and precisely as possible)**:

It just happens when I open a Grafana dashboard.

**Full logs to relevant components**:

<details>Logs
<p>
level=info ts=2022-02-28T09:04:26.161364912Z caller=client.go:55 msg="enabling client to server TLS"
level=info ts=2022-02-28T09:04:26.161683494Z caller=options.go:115 msg="TLS client using provided certificate pool"
level=info ts=2022-02-28T09:04:26.161707098Z caller=options.go:148 msg="TLS client authentication enabled"
level=info ts=2022-02-28T09:04:26.166055848Z caller=options.go:27 protocol=gRPC msg="disabled TLS, key and cert must be set to enable"
[...]

PR #5202, opened 2022-03-02
fix: add null check to exemplar data

tomas-mota:
Fixes: #5189

## Changes

Added a nil check for exemplar data

Did not add unit tests, since this is a rather simple method that already has no test


GiedriusS (approved):
LGTM, please add the DCO (:

GiedriusS (approved):

PR #5205, opened 2022-03-03
Append ruler labels as external labels when using stateless ruler mode

yeya24:

Fixes https://cloud-native.slack.com/archives/CK5RSSC10/p1646228262242149

## Changes

For the stateless ruler mode, we still need to add the `labels` as external labels written to the remote storage.

## Verification

Tested it locally and updated the E2E test.


matej-g (approved):
Looking good :+1:

PR #5206, opened 2022-03-03
cache: add timeout for groupcache's fetch operation

GiedriusS:
Add a timeout for groupcache's fetch operation. It is useful when there
are network errors - if loading from a peer fails then we still might
have a chance to be able to load data from remote object storage
ourselves.

## Changes

Added creation of a new context with the specified deadline if it is not zero.

## Verification

N/A

### DIFF
--- a/docs/components/store.md
+++ b/docs/components/store.md
@@ -442,7 +442,7 @@ In the `peers` section it is possible to use the prefix form to automatically lo
 
 Note that there must be no trailing slash in the `peers` configuration i.e. one of the strings must be identical to `self_url` and others should have the same form. Without this, loading data from peers may fail.
 
-If timeout is set to zero then there is no timeout for fetching and fetching's lifetime is equal to the lifetime to the original request's lifetime. It is recommended to keep it more than zero. It is generally preferred to keep this value higher because the fetching operation potentially includes loading that data from remote object storage.
+If timeout is set to zero then there is no timeout for fetching and fetching's lifetime is equal to the lifetime to the original request's lifetime. It is recommended to keep it higher than zero. It is generally preferred to keep this value higher because the fetching operation potentially includes loading of data from remote object storage.
 
 ## Index Header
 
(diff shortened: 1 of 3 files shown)


pkg/cache/groupcache.go
```diff
@@ -59,13 +60,17 @@ type GroupcacheConfig struct {
 
 	// How often we should resolve the addresses.
 	DNSInterval time.Duration `yaml:"dns_interval"`
+
+	// Timeout specifies the read/write timeout.
+	Timeout time.Duration
```
matej-g, line 65: Should we: 1) prefer `model.Duration` here?; 2) add `yaml` tag?

GiedriusS: All of the other providers use `time.Duration` so I used the same here. We should probably stick to `model.Duration` in code which is somehow related to PromQL? :shrug:  I'll add a yaml tag to be explicit, good point :+1:

matej-g: Seems you're right, we use `time.Duration` for caching YAML configs, although in other places in the code (e.g. object store configs) we are using `model.Duration`, which is a bit confusing. I thought we leverage `model.Duration` to keep the duration parsing logic consistent across all configs, but it's not true.

docs/components/store.md
```diff
@@ -441,6 +442,8 @@ In the `peers` section it is possible to use the prefix form to automatically lo
 
 Note that there must be no trailing slash in the `peers` configuration i.e. one of the strings must be identical to `self_url` and others should have the same form. Without this, loading data from peers may fail.
 
+If timeout is set to zero then there is no timeout for fetching and fetching's lifetime is equal to the lifetime to the original request's lifetime. It is recommended to keep it more than zero. It is generally preferred to keep this value higher because the fetching operation potentially includes loading that data from remote object storage.
```
matej-g, line 445: Nits:
```suggestion
If timeout is set to zero then there is no timeout for fetching and fetching's lifetime is equal to the lifetime to the original request's lifetime. It is recommended to keep it higher than zero. It is generally preferred to keep this value higher because the fetching operation potentially includes loading of data from remote object storage.
```

matej-g (approved):

matej-g (approved):

PR #5224, opened 2022-03-09
receive: remove sort on label hashing

pablo-ruth:

## Changes

I removed the sort on labels before hashinf them when using an hashring. The Prometheus [remote write specification](https://docs.google.com/document/d/1LPhVRSFkGNSuU1fBd81ulhsCPR4hkSZyyBj1SZ8fWOM/edit#heading=h.3he8womg60lm) stand that:

> MUST have label names sorted in lexicographical order

So I'm wondering if it's necessary to keep the sort on the receive side tor Thanos infrastructure with two "router only" receivers. I've done a pprof and I see that the sort code is using around 8% of CPU time, so it could be a significant optimization if we can skip it:

Screenshot from 2022-03-09 15-10-49

## Verification

We are using the patched version for some time on our receivers and confirmed a reduction in resource consumption. We didn't see any side-effect.


squat (dismissed):
This makes a lot of sense to me. For large requests this can really be a heavy step

pablo-ruth:
Great! I added the CHANGELOG entry :)

squat (approved):

PR #5226, opened 2022-03-09
Create v0.25.1 built with Go 1.17.8

metalmatze:
The binaries published with this release are built with Go1.17.8 to avoid [CVE-2022-24921](https://cve.mitre.org/cgi-bin/cvename.cgi?name=CVE-2022-24921).

Since it's the same Go minor version I wasn't able to find any other instance of changing Go from 1.17.7 to 1.17.8 and the patch version bump should happen when building with CI.


squat (approved):
😅Hard to tell if this actually has an effect because we don't pin patch versions; they float in the GitHub actions. But i guess we "know" because of documentation on GitHub actions side. Still, there could be weird, unlikely distributed system effects that cause our builds to use old, lower patch versions of 1.17.x

metalmatze:
Agreed. I checked the GitHub Actions logs and it seems that newer images for building were pulled. So I'm hopefully that simply building a newer image will fix the issue at hand.

bwplotka:
Thanks!

PR #5229, opened 2022-03-11
Change error cleanup in `objstore.DownloadDir` to delete files not destination dir

dimitarvdimitrov:

## Changes

Change error cleanup in `objstore.DownloadDir` to delete files not directories: 

`dst` is always a directory. If any file after the first fails to download, the cleanup will fail because the destination already contains at least one file. This PR changes the cleanup logic to clean up successfully downloaded files one by one instead of attempting to clean up the whole dst directory.

This PR does not account for the case where there are nested directories. While we can do `os.RemoveAll` on any child directories, it's unsafe. We don't know for sure that all files in child directories were downloaded as a result of _this_ `DownloadDir` call and didn't exist before it.

### DIFF
--- a/pkg/objstore/objstore.go
+++ b/pkg/objstore/objstore.go
@@ -276,6 +276,7 @@ func DownloadDir(ctx context.Context, logger log.Logger, bkt BucketReader, origi
 		downloadedFiles = append(downloadedFiles, dst)
 		return nil
 	}); err != nil {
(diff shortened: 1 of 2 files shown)


GiedriusS (commented):
Could you please add a unit test for this?

kakkoyun:
Even though, replacement hasn't done in main repo, better to target https://github.com/thanos-io/objstore from now on.

bwplotka (approved):
Thanks!

PR #5239, opened 2022-03-16
Write custom `errors` package with stack trace functionality

### LINKED ISSUE #5176: Find replacement for https://github.com/pkg/errors
https://github.com/pkg/errors is now archived since `errors` package has a way to wrap errors AND the main maintainer is stepping down from the maintainer team. Here are details https://github.com/pkg/errors/issues/245

Unfortunately, default library `errors` wrapping is not the best. Personally, I don't like three facts: 

* It requires doing `fmt.Errors("something: %w", err)` with `%w` which is ultra-easy to forgot and do %v instead. There are linters for that, but add extra lint when our language is compiled... feels weird. Plus it is prone to inconsistent wrapping.
* Doing `errors.New` sometimes then relying on `fmt` to wrap errors does not look clean. Swapping between `fmt` and `errors` in return looks like unnecessary complexity for the reader.
* There is no stack trace functionality which we use heavily https://github.com/thanos-io/thanos/blob/main/cmd/thanos/main.go#L131

Options:

A. We could write our own wrapper over `errors` to accomplish this, borrowing some code from pkg/errors, but it is (especially stack tracing) a bit of complexity to the maintainer.
B. We could use other drop-in replacements https://github.com/emperror/errors - it was not touched lately but seems like not many issues are on it. The API is a bit large though.
C. We could live without stack trace functionality. This is what the majority of projects are doing lately.

I am not a fan of standard library flow as mentioned before, and I don't think we mind setting a new standards in the ecosystem (e.g with https://github.com/efficientgo/tools/tree/main/core). But for time being, maybe it makes sense to start with B and check how it goes first? 🤔 

Thoughts?

bisakhmondal:
To test
```go
package errors_test

import (
	"testing"

	"github.com/thanos-io/thanos/pkg/errors"
)

func TestErrors(t *testing.T) {
	err := errors.New("an error occurred")
	err = errors.Wrap(err, "wraper1")
	t.Logf("error: %+v", err)
}
```

```shell
/home/bisakh/go/go1.17.8/bin/go tool test2json -t /tmp/GoLand/___TestErrors_in_github_com_thanos_io_thanos_pkg_errors.test -test.v -test.paniconexit0 -test.run ^\QTestErrors\E$
=== RUN   TestErrors
    errors_test.go:11: error: wraper1
        > github.com/thanos-io/thanos/pkg/errors_test.TestErrors	/home/bisakh/Desktop/OSS/thanos/pkg/errors/errors_test.go:10
        > testing.tRunner	/home/bisakh/go/go1.17.8/src/testing/testing.go:1259
        > runtime.goexit	/home/bisakh/go/go1.17.8/src/runtime/asm_amd64.s:1581
        an error occurred
        > github.com/thanos-io/thanos/pkg/errors_test.TestErrors	/home/bisakh/Desktop/OSS/thanos/pkg/errors/errors_test.go:9
        > testing.tRunner	/home/bisakh/go/go1.17.8/src/testing/testing.go:1259
        > runtime.goexit	/home/bisakh/go/go1.17.8/src/runtime/asm_amd64.s:1581
--- PASS: TestErrors (0.00s)
PASS
```

closes #5176 

## Verification
TODO:
- [ ] write tests
- [ ] full migration of existing error packages


bisakhmondal:
[WIP] How does the output look, @matej-g ?

pkg/errors/errors.go
```diff
@@ -0,0 +1,124 @@
+// Copyright (c) The Thanos Authors.
+// Licensed under the Apache License 2.0.
+
+// The idea of writing errors package in thanos is highly motivated from the Tast project of Chromium OS Authors. However, instead of
+// copying the package, we end up writing our own simplified logic borrowing some ideas from the errors and github.com/pkg/errors.
+// A big thanks to all of them.
```
saswatamcode, line 6: I wonder if we need to mention Chromium's license here too. I believe we do it for some Prometheus code that we use.

pkg/errors/errors.go
```diff
@@ -0,0 +1,124 @@
...
+func (b *base) Format(s fmt.State, verb rune) {
+	if verb == 'v' && s.Flag('+') {
+		s.Write([]byte(formatErrorChain(b)))
+		return
+	}
+
+	s.Write([]byte(b.Error()))
+}
+
+// Public functions that can be directly used through the package
```
saswatamcode, line 48: This is implied based on the casing. 🙂

pkg/errors/stacktrace.go
```diff
@@ -0,0 +1,46 @@
...
+
+package errors
+
+import (
+	"fmt"
+	"runtime"
+	"strings"
+)
+
+// stack holds the address of
```
saswatamcode, line 12: I think this would be ` holds a snapshot of program counters.`

pkg/errors/stacktrace.go
```diff
@@ -0,0 +1,46 @@
...
+type stacktrace []uintptr
+
+// New captures a stack trace. skip specifies the number of frames to skip from
+// a stack trace. skip=0 records stack.New call as the innermost frame.
```
saswatamcode, line 16: `skip` isn't passed as an argument here.

pkg/errors/stacktrace.go
```diff
@@ -0,0 +1,46 @@
...
+func newStackTrace() stacktrace {
+	const stackDepth = 16 // record maximum 16 frames (if available)
+
+	pc := make([]uintptr, stackDepth+1)
+	// using skip=3 for not to count the program counter address of
```
saswatamcode, line 21: 👍👍

pkg/errors/stacktrace.go
```diff
@@ -0,0 +1,46 @@
...
+func (s stacktrace) String() string {
+	var buf strings.Builder
+
+	// CallersFrames takes the slice of Program Counter addresses returned by Callers to
+	// retrieve function/file/line information.
+	cf := runtime.CallersFrames(s)
+	for {
+		// more indicates if the next call will be successful or not.
+		frame, more := cf.Next()
+		// used formatting scheme <`>`space><function name><newline><tab><filepath><:><line><newline>
```
saswatamcode, line 39: Perhaps an example would make it more understandable here.

saswatamcode (commented):
This looks good overall! Awesome work! 💫
Would love to see some more examples of this in action.

Just some suggestions.

pkg/errors/errors.go
bisakhmondal: Hi, thanks for the review. That's what I am also wondering too. To be honest, the current program is not an exact replication (here and there I have got rid of some unnecessary codes, interfaces, used string Builder for efficient formatting, changed the return signature of a few functions) of the Chromium OS Package.
Here's their license https://source.chromium.org/chromiumos/chromiumos/codesearch/+/main:src/platform/tast/LICENSE

What do you think we should do here?

pkg/errors/errors.go
bisakhmondal: Haha true. Some redundant comment. Removing

pkg/errors/stacktrace.go
bisakhmondal: Yep. Missed

pkg/errors/stacktrace.go
bisakhmondal: We won't need it. Removing the comments. The `pkg/errors` uses hardcoded skip=3. I think that's sufficient here too.

pkg/errors/stacktrace.go
bisakhmondal: Yes sure.

pkg/errors/errors.go
saswatamcode: I think it's fine if we just add a line pointing to this license here, as we don't copy but rather adapt code from Chromium. But I don't feel too strongly about this either way. 🙂
WDYT @matej-g?

Makefile
```diff
@@ -360,7 +360,8 @@ go-lint: check-git deps $(GOLANGCI_LINT) $(FAILLINT)
 	$(call require_clean_work_tree,'detected not clean work tree before running lint, previous job changed something?')
 	@echo ">> verifying modules being imported"
 	@# TODO(bwplotka): Add, Printf, DefaultRegisterer, NewGaugeFunc and MustRegister once exception are accepted. Add fmt.{Errorf}=github.com/pkg/errors.{Errorf} once https://github.com/fatih/faillint/issues/10 is addressed.
-	@$(FAILLINT) -paths "errors=github.com/pkg/errors,\
+	@$(FAILLINT) -paths "errors=github.com/thanos-io/thanos/pkg/errors,\
+github.com/pkg/errors=github.com/thanos-io/thanos/pkg/errors,\
```
bisakhmondal, line 364: I think we should replace all the instances of /pkg/errors in a separate PR instead of unnecessarily bloating this one. 
```shell
💥➜ thanos git:(pkg/errors) ✗ make lint 2>&1 | wc -l
144
```
Your views @matej-g?

metalmatze:
Do I understand this implementation correctly, that it's mostly a shim for the stdlib errors packages that supports stack traces as part of errors now?

bisakhmondal:
> Do I understand this implementation correctly, that it's mostly a shim for the stdlib errors packages that supports stack traces as part of errors now?

Hi @metalmatze, yes you are right. It extends the functionality errors package by combining
- a stacktrace with error chaining capabilities
- a `Wrap` and `Wrapf` functionality
- an `Errorf` functionality (closest alternative `fmt.Errorf`)

In simple words, a very minimalistic, simple and readable replacement of `github.com/pkg/errors` package : )
Thank you!

bisakhmondal:
> I wonder if we should not start by adding this error lib in https://github.com/efficientgo/tools/tree/main/core to keep util things as ultra small module?

Hi, thanks for the suggestion @bwplotka and I am sorry for the late reply. I just took a look at the package and especially the `merrors` pkg. Here are my few concerns.

- One question that's repeatedly coming to my mind is what's the utility of multierrors. Go has been evolved in such a language where errors are handled right after where they have been generated. Errors are errors - no matter how small/big the impact might be in the runtime. It's definitely a bad idea to not to act where err1 has been generated and add the err2 or others. Also the program/function should return the first error it has encountered not the aggregation of all errors (as this is implied, if step1 fails, it is highly likely that step2 and step... will fail).
```go
//  merr := merrors.New(err1)
//  merr.Add(err2, errOrNil3)
//  for _, err := range errs {
//    merr.Add(err)
//  }
//  return merr.Err()
```

- As per the actual issue, if I am not wrong we are looking for an alternative to `github.com/pkg/errors`. I am afraid, this package doesn't suffices so. There is no `wrap` or `unwrap` method implemented in the multierrors. To use this package we have to again fall back to `errors.New` of `fmt.Errorf` package
 (with something like `merrors.New(errors.New("the extra context"), err)`. 

  Even if we do so, the `Is` and `As` method will be inefficient in this case, 
https://github.com/efficientgo/tools/blob/fe763185946be83b20da626605319733bd7f97cb/core/pkg/merrors/errors.go#L116
It will iterate through the extra context which we have to pass as an error(due to the design of the package).

- Taking from the issue itself, 
> There is no stack trace functionality which we use heavily https://github.com/thanos-io/thanos/blob/main/cmd/thanos/main.go#L131 

- And finally, the methods exposed from the package will completely destroy backward compatibility. It's not just a simple change in the import path anymore (from 'pkg/errors` to 'github.com/thanos-io/thanos/pkg/errors`). We have to change each and every instance of the package where `err` has been used. In the initial days after the migration, there will be certain development toil among the community to adapt to something new.


Hey, I am new to the community and I don't have much context about the discussions and decisions : ) Here I just tried to do some research and share my findings if the `efficientgo/tools` meets the requirements. I think no for the aforementioned reasons.  WDYT? 

Thanks : )

saswatamcode:
Hey @bisakhmondal , thanks for the detailed analysis! :)

To address some of your concerns,

> One question that's repeatedly coming to my mind is what's the utility of multierrors. Go has been evolved in such a language where errors are handled right after where they have been generated. Errors are errors - no matter how small/big the impact might be in the runtime. It's definitely a bad idea to not to act where err1 has been generated and add the err2 or others. Also the program/function should return the first error it has encountered not the aggregation of all errors (as this is implied, if step1 fails, it is highly likely that step2 and step... will fail).
//  merr := merrors.New(err1)
//  merr.Add(err2, errOrNil3)
//  for _, err := range errs {
//    merr.Add(err)
//  }
//  return merr.Err()

Hmm, I'd say there is definitely utility in having `merrors` package. :)
Even in Go, we might not always want to handle an error as soon as it occurs! Sometimes, we may just want to return all errors in one go, and this package is definitely useful for that.

For example, say in our docs CI, we have a tool running called [mdox](https://github.com/bwplotka/mdox), which checks each markdown file and ensures that they are correctly formatted and have correct links. You run it once via `make docs` and it outputs all the errors that might exist in all the files.

If we were to just return the very first error encountered, users would have to run it again and again until they have fixed each error, which is definitely not optimal and would be a painful process.

There are use cases for this in Thanos too, for example, the `tools rule-check` command which checks if all the rule files passed are valid or not. See 
https://github.com/thanos-io/thanos/blob/19dcc7902d2431265154cefff82426fbc91448a3/cmd/thanos/tools.go#L49

You'd also find such cases scattered throughout the codebase.

So there are definitely use cases for this! I think there are also other implementations of this like https://github.com/hashicorp/go-multierror.

But a question I have for your point is, why does the existence of `merrors` matter here? This PR is just a replacement for the errors package that we currently use, not related to the way we handle errors. So the suggestion is to add it to our utils repo `efficientgo/tools` and then use it from there. We don't really need to change existing `merrors`, but add a new errors package, right?

> As per the actual issue, if I am not wrong we are looking for an alternative to [github.com/pkg/errors](http://github.com/pkg/errors). I am afraid, this package doesn't suffices so. There is no wrap or unwrap method implemented in the multierrors. To use this package we have to again fall back to errors.New of fmt.Errorf package
(with something like merrors.New(errors.New("the extra context"), err).
Even if we do so, the Is and As method will be inefficient in this case,
https://github.com/efficientgo/tools/blob/fe763185946be83b20da626605319733bd7f97cb/core/pkg/merrors/errors.go#L116
It will iterate through the extra context which we have to pass as an error(due to the design of the package).

Not sure what you mean here. I think @bwplotka's point was to add this as a separate "errors" package in the tools repo. Not as an extension or modification to the existing `merrors` one.

> And finally, the methods exposed from the package will completely destroy backward compatibility. It's not just a simple change in the import path anymore (from 'pkg/errors to '[github.com/thanos-io/thanos/pkg/errors](http://github.com/thanos-io/thanos/pkg/errors)). We have to change each and every instance of the package where err has been used. In the initial days after the migration, there will be certain development toil among the community to adapt to something new.

Hmm. Why would it destroy backward compatibility? We aren't changing/removing any of the existing functionality, right?

If this package is added to `efficientgo/tools` repo, iiuc, it would still be a change in import path from `pkg/errors` to `[github.com/efficientgo/tools/pkg/errors](http://github.com/efficientgo/tools/pkg/errors)` + a version bump in `go.mod`. There wouldn't be any difference in dev toil when compared with adding a custom errors package that this PR is proposing.

I might be understanding something differently about your points here, so correct me if I'm wrong :).

bisakhmondal:
> But a question I have for your point is, why does the existence of `merrors` matter here? This PR is just a replacement for the errors package that we currently use, not related to the way we handle errors. So the suggestion is to add it to our utils repo `efficientgo/tools` and then use it from there. We don't really need to change existing `merrors`, but add a new errors package, right?
> 

Ahh, Thanks for the clarification @saswatamcode. I misinterpreted @bwplotka's comment thinking we are going to use the error package (`merrors`) that is currently present in the tools package. 

Definitely, there are certain use cases where combining multiple errors makes sense. As I said, I misinterpreted that we are going to make this a default throughout the thanos project which actually doesn't make sense.

> > As per the actual issue, if I am not wrong we are looking for an alternative to [github.com/pkg/errors](https://github.com/pkg/errors). I am afraid, this package doesn't suffices so. There is no wrap or unwrap method implemented in the multierrors. To use this package we have to again fall back to errors.New of fmt.Errorf package
> > (with something like merrors.New(errors.New("the extra context"), err).
> > Even if we do so, the Is and As method will be inefficient in this case,
> > https://github.com/efficientgo/tools/blob/fe763185946be83b20da626605319733bd7f97cb/core/pkg/merrors/errors.go#L116
> > It will iterate through the extra context which we have to pass as an error(due to the design of the package).
> 
> Not sure what you mean here. I think @bwplotka's point was to add this as a separate "errors" package in the tools repo. Not as an extension or modification to the existing `merrors` one.

Yep, I got it now : ) I am totally fine with keeping the error package as a part of thanos project or inside efficientgo/tools. I have no strong opinion here : )

bisakhmondal:
> But we can start small with just implementing it here for now

Great! Would you mind giving it a quick look and sharing some reviews? I'll write the test suite then : )
Thank you!

pkg/errors/errors.go
```diff
@@ -0,0 +1,143 @@
...
+	return &base{
+		info:  fmt.Sprintf(format, args...),
+		stack: newStackTrace(),
+		err:   nil,
+	}
+}
+
+// Wrap creates a new error with the given message, wrapping another error and a stacktrace in details.
+// If cause is nil, this is the same as New.
+func Wrap(cause error, msg string) error {
```
bwplotka, line 70: Since we innovate here, I wonder if we have to have both `Wrapf` and `Wrap` at the end, why not having only `Wrapf` and just do `format` without `%v` or something? 🤔

pkg/errors/errors.go
```diff
@@ -0,0 +1,143 @@
...
+	return &base{
+		info:  msg,
+		stack: newStackTrace(),
+		err:   nil,
+	}
+}
+
+// Errorf creates a new error with the given message and a stacktrace in details.
+// An alternative to fmt.Errorf function.
+func Errorf(format string, args ...interface{}) error {
```
bwplotka, line 60: Same here - again - it's our package, we can innovate. 

What about only 

`errors.Newf` and `errors.Wrapf`?

Or even naming it without `f` but supporting fmt? No harm here, but simper API for everyone. So:

`errors.New(format string, args ...interface{})` and `errors.Wrap(err, format string, args ...interface{})` would work too

pkg/errors/errors.go
```diff
@@ -0,0 +1,143 @@
...
+
+import (
+	"errors"
+	"fmt"
+	"strings"
+)
+
+// base is the fundamental struct that implements the error interface and the acts as the backbone of this errors package.
+type base struct {
+	info  string     // error message passed through calls like errors.Wrap, errors.Errorf
```
bwplotka, line 20:
```suggestion
	info  string     // Error message passed through calls like errors.Wrap, errors.Errorf.
```

Let's use full sentence for all comments, including those short ones, please 🙈

pkg/errors/errors.go
```diff
@@ -0,0 +1,143 @@
...
+	}
+	return u.Unwrap()
+}
+
+// Is is a wrapper of built-in errors.Is. It reports whether any error in err's
+// chain matches target. The chain consists of err itself followed by the sequence
+// of errors obtained by repeatedly calling Unwrap.
+func Is(err, target error) bool {
+	return errors.Is(err, target)
+}
```
bwplotka, line 121: wonder if alias `var Is = errors.Is` would not work? (and same for `As`).

bwplotka (commented):
Great job. I have a few proposals - feel free to challenge non sensible former ideas ppl used to do. Especially with `Wrap` vs `Wrapf` is something we could improve 🤔 

Let's make sure it's clear and ready.

We also need some at least basic tests. Thanks!

.golangci.yml
```diff
@@ -16,6 +16,9 @@ run:
   #     vendor$, third_party$, testdata$, examples$, Godeps$, builtin$
   skip-dirs: vendor
 
+  skip-files:
+    - pkg/errors/errors.go # due to the errors import
```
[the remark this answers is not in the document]
bisakhmondal: Indeed, I want to keep the files for linting but suppress the error as well due to the import of native errors package
image

pkg/errors/errors.go
bisakhmondal: It would work, updating

pkg/errors/errors.go
bisakhmondal: Haha, sure.

.golangci.yml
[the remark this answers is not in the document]
saswatamcode: The lint error is from [faillint](https://github.com/fatih/faillint) based on paths defined [here](https://github.com/thanos-io/thanos/blob/main/Makefile#L369). I think it'd be best to just add a nolint directive here like so, 
```go
import (
	//lint:ignore faillint Custom errors package needs to import standard library errors.
	"errors"
	"fmt"
	"strings"
)
```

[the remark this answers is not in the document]
bisakhmondal: Thanks, @saswatamcode for the references. It works. Actually, by mistake, I was trying with `//nolint` directive and for the obvious reason, it wasn't doing any good.

pkg/errors/stacktrace.go
```diff
@@ -0,0 +1,48 @@
...
+// stacktrace holds a snapshot of program counters.
+type stacktrace []uintptr
+
+// newStackTrace captures a stack trace. It skips first 3 frames to record the
+// snapshot of the stack trace at the origin of a particular error. It tries to
+// record maximum 16 frames (if available).
+func newStackTrace() stacktrace {
+	const stackDepth = 16 // record maximum 16 frames (if available)
+
+	pc := make([]uintptr, stackDepth)
```
sthaha, line 21: This implementation has the same issue as `pkg/errors` i.e. the `pc` escapes to heap thus the entirety of the 16 uintptr is kept in memory even if only the stack is only 1 call frame deep.  I have explained the issue in details [here](https://github.com/emperror/errors/issues/27) based on a real-world scenario. 



Please try `go build -gcflags='-m -m' ./pkg/errors/stacktrace.go` and you would find 

```
pkg/errors/stacktrace.go:21:12: make([]uintptr, stackDepth) escapes to heap:
pkg/errors/stacktrace.go:21:12:   flow: pc = &{storage for make([]uintptr, stackDepth)}:
pkg/errors/stacktrace.go:21:12:     from make([]uintptr, stackDepth) (spill) at pkg/errors/stacktrace.go:21:12
pkg/errors/stacktrace.go:21:12:     from pc := make([]uintptr, stackDepth) (assign) at 
```
The solution is to allocate the buffer `pc` on stack like the original implementation and then copy whats needed off it to prevent escaping to heap. 

```suggestion
	const maxDepth = 16
	var pcs [maxDepth]uintptr.  // allocate on stack

	n := runtime.Callers(3, pcs[:])
	st := make(stack, n)
	copy(st, pcs[:n])
	return st
```

Another thought is if we really need to hold on to the entire CallFrame until `String()` is called? At the expense of some compute, we could calculate Stacktrace string immediately.

bisakhmondal: Nice nitpick and great catch @sthaha.
>  the pc escapes to heap thus the entirety of the 16 uintptr is kept in memory even if only the stack is only 1 call frame deep.

Yes, you are right - there is no point in holding an extra chunk of unused heap memory, And I think you are copying to another temporary slice because even though we are returning `stacktrace(pc[:n])` the capacity of PC still remains 16 (allocated memory), right?

But this additional copy adds some extra complexity : ) (though it's very minor and negligible as we are dealing with length of 16) From escape analysis the function complexity gets increased from 88 -> 96
```
./stacktrace.go:20:6: cannot inline newStackTrace: function too complex: cost 96 exceeds budget 80
.
.
./stacktrace.go:20:6: cannot inline newStackTrace: function too complex: cost 88 exceeds budget 80

```

Since go1.2 we have 3 index slice capability where the third param can define the capacity of the newly created slice.

So just updating the return statement to `return stacktrace(pc[:n:n])` would yield the same effect of the suggestion you proposed.

Thanks a lot. That was really awesome.

> Another thought is if we really need to hold on to the entire CallFrame until String() is called? At the expense of some compute, we could calculate Stacktrace string immediately.

I think we shouldn't do it for the following reasons
- Its compute expensive - runtime needs to retrieve caller information
- More memory usage - the `String` method is meant for human consumption so naturally, this adds a lot of extra information, text etc compared to the 16 element slice of type unitptr. During error changing it would be worse.
- Not every error needs the stacktrace. For specific format verbs like `%+v` stacktrace gets dumped recursively (only then String gets called).

sthaha: > Since go1.2 we have 3 index slice capability where the third param can define the capacity of the newly created slice.
💯  .. Thank you, I wasn't aware :) 

> I think we shouldn't do it for the following reasons ... 

> Its compute expensive - runtime needs to retrieve caller information 
Have you tried a benchmark? 

> More memory usage - the String method is meant for human consumption so naturally, this adds a lot of extra information, text etc compared to the 16 element slice of type unitptr. During error changing it would be worse.

What I meant was to only hold on to information that you need than have `uintptr` which now points to the callframe and IIUC, holding onto the callframe  will now prevent GC from cleaning the callframe info until the error is garbage collected as well. 

It may be worth seeing the inuse allocation if we don't hold on to the callframes.

bisakhmondal: > > I think we shouldn't do it for the following reasons ...
> 
> > Its compute expensive - runtime needs to retrieve caller information
> 
> Have you tried a benchmark?
> 

IIUC, in the new approach that you are suggesting we should call the `String` immediately at the expense of an extra call which might not be required if the error is not logged with an `"%+v"` format verb. So we are adhering to extra computation which might not be of any use. So obviously it will be expensive in time against the current approach  as the current one only calls string when it's required.

> > More memory usage - the String method is meant for human consumption so naturally, this adds a lot of extra information, text etc compared to the 16 element slice of type unitptr. During error chaining it would be worse.
> 
> What I meant was to only hold on to information that you need than have `uintptr` which now points to the callframe and IIUC, holding onto the callframe will now prevent GC from cleaning the callframe info until the error is garbage collected as well.
> 

Now coming to the memory footprint, IMHO storing unstructured data (the stack trace output string) is generally a bad idea. If we change the output string format from `fmt.Sprintf("> %s\t%s:%d\n", frame.Func.Name(), frame.File, frame.Line)` this to something else memory footprint gets changed. So I think this optimization won't be valid in longer run.

After giving it a thought, I think it's a tradeoff, why?
Memory usage:
- current approach: 16*4 bytes uintptr + unreleased memory from gc (not sure what it stores on that PC address, might be some function related metadata to populate the callFrames. definitely releases ass the resources used inside the function)
- proposed approach: large unstructured string + still some unreleased memory (not sure when will the next cycle of go GC will kick off and it's dependent on the config). 

I am not sure if we can benchmark this anyhow or not.

>the error is garbage collected as well.

If you see, in Go world err has the comparatively shortest lifespan than anything else. In a well written program they gets handled immediately by being returned to the caller function or logged into the sink. So I think we are good here. After all it's an internal package and if we found some untoward runtime behaviour we can always do different optimization. 

Thank you : )

bisakhmondal:
It looks like the Lift and DCO CIs are pending forever. Does this take this long, especially the code analysis by lift? Is there a way we can optimize it? I'd love to. 

Hi guys, could you please take another look at this PR when you have some time : )

cc @bwplotka

pkg/errors/errors.go
bwplotka: Nice.

After some thinking, let's keep `f` word suffix so we are consistent with some lints/tooling that tries to verify formatters. Just removing non f and being able to produce `format` string without variables is fine. WDYT?

bwplotka: Also next time, just in case use 

```
func benchWrapf(errorMsg *string, b *testing.B) {
       b.ResetAllocs()
	for i := 0; i < b.N; i++ {
		_ = errors.Wrap(errors.New("random error"), *errorMsg)
	}
}
```

pkg/errors/stacktrace.go
bwplotka: > If you see, in Go world err has the comparatively shortest lifespan than anything else. 

Not always. Check our fixed issue: https://github.com/thanos-io/thanos/issues/5257

Anyway, I asked @sthaha to remind us about the optimization emperor did - to make sure we are aware. As usually we can iterate over it, we won't do it perfectly over time - the main thing is to get APIs as best as possible - it's hard to change them later.

pkg/errors/errors.go
```diff
@@ -0,0 +1,128 @@
+// Copyright (c) The Thanos Authors.
+// Licensed under the Apache License 2.0.
+
+//nolint
```
bwplotka, line 4: Again, why nolint?

pkg/errors/errors.go
```diff
@@ -0,0 +1,128 @@
...
+		stack: newStackTrace(),
+		err:   nil,
+	}
+}
+
+// Wrap creates a new error with the given message, wrapping another error and a stacktrace in details.
+// If cause is nil, this is the same as New. This Wrap method combines both Wrap and Wrapf functionality.
+// If wrapped is used with a format string with format verbs and the args, it will be same as Wrapf (a
+// convention used in error packages).
+func Wrap(cause error, format string, args ...interface{}) error {
```
bwplotka, line 70: I wonder - there might be use cases where one wants to use stack traces always - thus doing something like `errors.Wrap(err)` always 🙃  - we could play with `Wrap(cause error, args ...interface{})` for this.

I think however this is the case is not worth optimization for - quite rare pattern and if we want you can always do another method for this.

But I think text is still useful - if you

bwplotka (dismissed):
This is an amazing job. LGTM. Before merge my only question is really about doing: `Newf` and `Wrapf` (A) vs `New` and `Wrap` (B).

I like the fact that with `no-f` (B) version it's cleaner and we show that we don't ever support Wrapper/New without sprintf like formatting. On other hand people got used to things, plus there might be nice tooling which work only if method is `f` suffix (don't have example - I saw linters for Printf formatting errors). 

I think I am leaning towards (A) actually... Thoughts?

bwplotka:
Let's try to push (or repush) another commit for DCO to unstuck. You can ignore `lift` - it's not required and rarely working 🙃

pkg/errors/errors.go
bisakhmondal: > Tast project

spellchecker thinks the project name should be `Taste` instead of the keyword `Tast` :upside_down_face:

pkg/errors/errors.go
bisakhmondal: Great! Thanks

bisakhmondal:
> This is an amazing job. LGTM. Before merge my only question is really about doing: `Newf` and `Wrapf` (A) vs `New` and `Wrap` (B).
> 
> I like the fact that with `no-f` (B) version it's cleaner and we show that we don't ever support Wrapper/New without sprintf like formatting. On other hand people got used to things, plus there might be nice tooling which work only if method is `f` suffix (don't have example - I saw linters for Printf formatting errors).
> 
> I think I am leaning towards (A) actually... Thoughts?


Hi, thanks for the nice feedback and the reviews. 
I am fine with both approaches as through benchmarks we have seen that they are of equivalent performance. 

Yes, to support approach A, what you have said related to developers' experience and the linters definitely make sense. People are accustomed to having both `f` and non `f` APIs and having only non `f` APIs (approach B) will definitely incur some adaptation cost and multiple visit to the package codes to understand what/how actually it does things.

Okay then, do let me know your final thoughts - we are going with approach A, right? I'll make the necessary changes ASAP : )
Thank you!

pkg/errors/errors.go
bisakhmondal: > But I think text is still useful 

Yes, absolutely. I am not sure if allowing the `errors.Wrap(err)` without a "reason string" would be a good idea or not. I think it would cause bad development practice as most of the time inside wrap the reason string would be omitted (just because it's optional :smile: ).

bwplotka:
Yes, I think A is safer.

bisakhmondal:
> Yes, I think A is safer.

Sure, updating : )

pkg/errors/errors.go
bisakhmondal: Hey, sorry, I have reverted back to the old one.
Reason: Though using `var Is = errors.Is` would definitely make things work but users will have a slightly poor IDE experience as the ide will treat those functions as variables (as actually, they are in the modified pkg). So the auto-completion works a bit weird way
see the screenshot
image

I dug in a little bit to find an alternative approach to tackle this, but it seems popular packages do use the wrapping while exposing their client-side APIs from the internal packages.
ref: https://github.com/temporalio/sdk-go/blob/fd0d1eb548eb0621a5395581cfe2c418704b007c/client/client.go#L435-L476

I hope you are okay with it. @bwplotka?

bwplotka: Makes sense, but worth adding comment why we did this this way 🙃

pkg/errors/errors.go
```diff
@@ -0,0 +1,155 @@
...
+		stack: newStackTrace(),
+		err:   nil,
+	}
+}
+
+// Wrap returns a new error with the supplied message as an argument and wrapping another error
+// with a stacktrace containing recent call frames.
+//
+// If cause is nil, this is the same as New.
+func Wrap(cause error, msg string) error {
```
bwplotka, line 79: I see we left our both non `f` and `f` methods at the end? 🤔

bwplotka (dismissed):
Nice, LGTM, just minor nits. We have opportunity to simplify this package, but we ended up having both `New, Errorf, Wrap and Wrapf`. Again, there is an opportunity to simplify this, so reviewers do not need to be a pain and saying "Why you used `Errrof("non formattable error")` and not `New` (and similar for Wrap)

pkg/errors/errors.go
bisakhmondal: Hey, I thought we are focussing on end-users experience that's why ended up keeping both.
I have changed my mind, let's simplify things and stick to only the `f` versions :smile:

pkg/errors/errors.go
bisakhmondal: Sure : )

pkg/errors/errors.go
bisakhmondal: Mission accomplished : )
image

matej-g (dismissed):
Thanks for putting in all the great work @bisakhmondal :muscle: This has been through a lot of review cycles and it looks like we have settled on a nice, minimalist API design! Looks good to me!

pkg/errors/errors.go
```diff
@@ -0,0 +1,142 @@
...
+// The idea of writing errors package in thanos is highly motivated from the Tast project of Chromium OS Authors. However, instead of
+// copying the package, we end up writing our own simplified logic borrowing some ideas from the errors and github.com/pkg/errors.
+// A big thanks to all of them.
+
+// Package errors provides basic utilities to manipulate errors with a useful stacktrace. It combines the
+// benefits of errors.New and fmt.Errorf world into a single package.
+package errors
+
+import (
+	//lint:ignore faillint Custom errors package needs to import standard library errors.
```
bwplotka, line 14: 👍🏽

pkg/errors/errors.go
bwplotka: BTW I know why it did not work for you. We can do `type Is = errors.Is` I think. Let's merge and iterate (:

bwplotka (dismissed):
LGTM! 

This is amazing work! And easy to use 💪🏽 

I still would love to have this as standalone errors module or part of core repo, but let's prove it works for us well 💪🏽  This deserves some blog post (:

matej-g:
I think this needs rebasing, the docs check will keep failing due to an outdated link in the docs :disappointed: cc @bisakhmondal

pkg/errors/errors.go
bisakhmondal: > I think this needs rebasing, the docs check will keep failing due to an outdated link in the docs disappointed cc @bisakhmondal

Done. Thanks to you both for the awesome review : )

bisakhmondal:
> I think this needs rebasing, the docs check will keep failing due to an outdated link in the docs disappointed cc @bisakhmondal

Done. Thanks to you both for the awesome review : )

bisakhmondal:
Hi Guys, all the CIs have passed. Please approve and merge this. Then I'll proceed with the rest of the refactoring : )

bisakhmondal:
Ping @bwplotka @matej-g 
Let's merge this guys : )

bwplotka (approved):
Sorry for the massive lag! KubeCon toil. LGTM!

Let's try to use it and improve on the way. Ideally at some point move to [efficientgo/core](https://github.com/efficientgo/tools) for others to import it.

bwplotka:
Thank you! 🎉 Amazing work!

bisakhmondal:
Thank you everyone for those awesome reviews :)

PR #5242, opened 2022-03-17
Ruler: Fix WAL directory in stateless mode

matej-g:

## Changes
When running the Thanos ruler in stateless mode, the WAL directory that is passed to the agent is not constructed correctly. See this output when starting ruler:
```
level=info ts=2022-03-17T12:40:21.537774846Z caller=db.go:306 msg="replaying WAL, this may take a while" dir=/var/thanos/rule/wal/wal
level=info ts=2022-03-17T12:40:21.537966759Z caller=db.go:357 msg="WAL segment loaded" segment=0 maxSegment=0
```
Notice the `dir` key in the log line, which contains extra sub-directory, ending in `wal/wal`. 

The suggested fix is to provide the 'base' data directory instead of WAL directory, since the full WAL path is built by the agent.

We are also presuming this might have cause us seeing some gaps in data, since stateless ruler could not longer read data from the previous WAL directory, since it was reading from the `wal/wal` "sub-subdirectory".

Also credit goes to @saswatamcode for pairing with me on this issue!

## Verification

Stateless ruler tests are passing.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -12,6 +12,8 @@ We use *breaking :warning:* to mark changes that are not backward compatible (re
 
 ### Fixed
 
+- [#5242](https://github.com/thanos-io/thanos/pull/5242) Ruler: Make ruler use the correct WAL directory.
+
 ### Added
 
 - [#5220](https://github.com/thanos-io/thanos/pull/5220) Query Frontend: Add `--query-frontend.forward-header` flag, forward headers to downstream querier.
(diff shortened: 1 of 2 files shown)


ianbillett (approved):

yeya24 (approved):
LGTM!

bwplotka:
Are we really sure those WALs are compatible? I am not 100% convinced they are

bwplotka:
I think it is indeed https://github.com/prometheus/prometheus/blob/main/tsdb/agent/db.go#L42 - ignore me then, LGTM (:

PR #5250, opened 2022-03-22
Implement GRPC query API

fpetkovski:
With the current GRPC APIs, layering Thanos Queriers results in
the root querier getting all of the samples and executing the query
in memory. As a result, the intermediary Queriers do not do any
intensive work and merely transport samples from the Stores to the
root Querier.

When data is perfectly sharded, users can implement a pattern where
the root Querier instructs the intermediary ones to execute the queries
from their stores and return back results. The results can then be
concatenated by the root querier and returned to the user.

In order to support this use case, this commit implements a GRPC API
in the Querier which is analogous to the HTTP Query API exposed
by Prometheus.

Signed-off-by: fpetkovski <filip.petkovsky@gmail.com>

## Changes

Implement a query API through GRPC, analogous to the one exposed on HTTP. 

## Verification

Through e2e tests


squat:
I wonder if we should have a proposal to cover this new API and cement it in the Thanos project officially, just like the Store API. WDYT?

fpetkovski:
I think a proposal would be useful to document the API and describe the rationale behind it. Should I add it to this PR?

squat:
🤝 great yes please!

GiedriusS (approved):
I think this is the next, natural step. I have tried implementing a similar thing here: https://github.com/thanos-io/thanos/pull/4712. Perhaps we could get this smaller PR through first before implementing this API in Thanos Store. Yeah, a small document/proposal would be nice with the rationale :thinking:

docs/proposals-done/202203-grpc-query-api.md
```diff
@@ -0,0 +1,108 @@
...
+
+This is useful for a few reasons: 
+
+* When Queriers register disjoint Store targets, they should be able to deduplicate series and then execute the query without concerns of duplicate data from other queriers. This new API would enable users to effectively partition by Querier, and avoid shipping raw series back from each disjointed Querier to the root Querier.
+* If Queriers register conjoint Store targets, users would be able to express a query sharding strategy between Queriers to more effectively distribute query load amongst a fleet of homogenous Queriers. 
+* The proposed Query API utilizes gRPC instead of HTTP, which would enable gRPC streaming from root Querier all the way to the underlying Store targets (Query API -> Store API) and unlock the performance benefits of gRPC over HTTP.
```
GiedriusS, line 16: Looks good. I'd suggest adding another point here according to my work here https://github.com/thanos-io/thanos/pull/4712. Something like:

* If there only one StoreAPI connected to Thanos Query completely covers the requested range then it's more optimal to execute the query directly using this API instead of sending raw samples & only then to execute the query

Makefile
```diff
@@ -25,6 +25,9 @@ else ifeq ($(arch), armv8)
 else ifeq ($(arch), arm64)
 	# arm64
 	BASE_DOCKER_SHA=${arm64}
+else ifeq ($(arch), aarch64)
```
bwplotka, line 28: M1 Pro? (:

docs/proposals-done/202203-grpc-query-api.md
```diff
@@ -0,0 +1,108 @@
+## Introduce a Query gRPC API
+
+* **Owners:**
+    * `@fpetkovski` 
+    * `@mzardab` 
+
+> TL;DR: Introducing a new gRPC API for `/query` and `/query_range`
+
+## Why
+We want to be able to distinguish between gRPC Store APIs and other Queriers in the query path. Currently, Thanos Query implements the gRPC Store API and the root Querier does not distinguish between Store targets and other Queriers that are capable of processing a PromQL expression before returning the result. The new gRPC Query API will allow a querier to fan out query execution, in addition to Store API selects.
```
bwplotka, line 10: > We want to be able to distinguish between gRPC Store APIs and other Queriers in the query path.

I don't get that - you can distinguish it easily with configuration, no need for a new API for this particular point?

docs/proposals-done/202203-grpc-query-api.md
bwplotka: 👍🏽

docs/proposals-done/202203-grpc-query-api.md
```diff
@@ -0,0 +1,108 @@
...
+
+This is useful for a few reasons: 
+
+* When Queriers register disjoint Store targets, they should be able to deduplicate series and then execute the query without concerns of duplicate data from other queriers. This new API would enable users to effectively partition by Querier, and avoid shipping raw series back from each disjointed Querier to the root Querier.
+* If Queriers register conjoint Store targets, users would be able to express a query sharding strategy between Queriers to more effectively distribute query load amongst a fleet of homogenous Queriers.
```
bwplotka, line 15: What `conjoint` might mean in this context?

docs/proposals-done/202203-grpc-query-api.md
```diff
@@ -0,0 +1,108 @@
...
+### Audience
+* Thanos Maintainers
+* Thanos Users
+
+## How
+
+We propose defining the following gRPC API:
+```protobuf
+service Query {
+  rpc Query(QueryRequest) returns (QueryResponse);
```
bwplotka, line 40: > he two main reasons are the streaming capabilities that come out of the box with gRPC, statically typed API spec, as well as the lower bandwidth utilization which protobuf enables.

so..

```suggestion
  rpc Query(QueryRequest) returns (stream QueryResponse);
```

?

docs/proposals-done/202203-grpc-query-api.md
```diff
@@ -0,0 +1,108 @@
...
+  
+  rpc QueryRange(QueryRangeRequest) returns (QueryRangeResponse);
```
bwplotka, line 42:
```suggestion
  rpc QueryRange(QueryRangeRequest) returns (stream QueryRangeResponse);
```

docs/proposals-done/202203-grpc-query-api.md
```diff
@@ -0,0 +1,108 @@
...
+  repeated StoreMatchers storeMatchers = 6 [(gogoproto.nullable) = false];
+
+  bool enableDedup = 7;
+  bool enablePartialResponse = 8;
+  bool enableQueryPushdown = 9;
+  bool skipChunks = 10;
+}
+
+message QueryResponse {
+  repeated prometheus_copy.TimeSeries timeseries = 1 [(gogoproto.nullable) = false];
```
bwplotka, line 67: oneof stream message with stats and warnings as well as we have in StoreAPI?

pkg/api/query/grpc.go
```diff
@@ -0,0 +1,161 @@
...
+
+	"github.com/prometheus/prometheus/promql"
+	"github.com/thanos-io/thanos/pkg/api/query/querypb"
+	"github.com/thanos-io/thanos/pkg/query"
+	"github.com/thanos-io/thanos/pkg/store/labelpb"
+	"github.com/thanos-io/thanos/pkg/store/storepb/prompb"
+	"google.golang.org/grpc"
+)
+
+type GrpcAPI struct {
```
bwplotka, line 18:
```suggestion
type GRPCAPI struct {
```

Same as we don't have `HttpAPI`?

bwplotka (commented):
This is great, but I wonder if we can make it right from start and allow streaming? 🤔

Makefile
fpetkovski: Yeah, the good stuff :)

docs/proposals-done/202203-grpc-query-api.md
fpetkovski: Sure, we can implement streaming in the initial version

docs/proposals-done/202203-grpc-query-api.md
fpetkovski: It means overlapping/interleaved series. Updated the sentence

docs/proposals-done/202203-grpc-query-api.md
bwplotka: <3

bwplotka:
https://github.com/thanos-io/thanos/runs/5694925196?check_suite_focus=true is failing -> please run `make docs` and commit changes (auto formatting).

pkg/api/query/grpc.go
```diff
@@ -0,0 +1,170 @@
...
+	}
+}
+
+func RegisterQueryServer(queryServer querypb.QueryServer) func(*grpc.Server) {
+	return func(s *grpc.Server) {
+		querypb.RegisterQueryServer(s, queryServer)
+	}
+}
+
+func (grpcAPI *GRPCAPI) Query(request *querypb.QueryRequest, server querypb.Query_QueryServer) error {
```
bwplotka, line 40:
```suggestion
func (g *GRPCAPI) Query(request *querypb.QueryRequest, server querypb.Query_QueryServer) error {
```

nit: We can make it shorter if we want. (not a blocker)

pkg/api/query/querypb/query.proto
```diff
@@ -0,0 +1,89 @@
...
+}
+
+message StoreMatchers {
+  repeated LabelMatcher labelMatchers = 1 [(gogoproto.nullable) = false];
+}
+
+message QueryResponse {
+  oneof result {
+    /// warnings are additional messages coming from the PromQL engine.
+    string warnings = 1;
```
bwplotka, line 48: +1

bwplotka (approved):
Amazing work, thanks! Do you mind auto-formatting docs and removing extra dependency `gotest`?

Otherwise, LGTM, thanks for streaming approach from the start 💪🏽

go.mod
```diff
@@ -96,6 +96,7 @@ require (
 	gopkg.in/fsnotify.v1 v1.4.7
 	gopkg.in/yaml.v2 v2.4.0
 	gopkg.in/yaml.v3 v3.0.0-20210107192922-496545a6307b
+	gotest.tools v2.2.0+incompatible
```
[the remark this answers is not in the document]
fpetkovski: Do we have something for polling in that package?

bwplotka (approved):
Thanks!

bwplotka (approved):
Thanks! Amazing job 💪🏽

docs/proposals-done/202203-grpc-query-api.md
moadz: Sure, but we can't distinguish between them via gRPC, only via HTTP.

PR #5255, opened 2022-03-25
info: Return store info only when the service is ready

yeya24:

## Changes

Store only returns store info when its status is ready.
This is helpful in our scenario:
When a prometheus itself is down (due to storage error), the pod readiness check failed and the endpoint got removed from the service. However, because of our CNI solution, the IPVS service backend will remain for 10m(same as the pod terminationGraceSeconds) so that the Thanos Query can still connect to the sidecar for 10m, producing a lot of Query partial errors.

Once the sidecar detects that Prometheus is down, it should let the Query know that it is not available for serving any requests because it is not ready. Same for other stores.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -26,6 +26,7 @@ We use *breaking :warning:* to mark changes that are not backward compatible (re
 - [#5224](https://github.com/thanos-io/thanos/pull/5224) Receive: Remove sort on label hashing
 - [#5231](https://github.com/thanos-io/thanos/pull/5231) Tools: Bucket verify tool ignores blocks with deletion markers.
 - [#5244](https://github.com/thanos-io/thanos/pull/5244) Query: Promote negative offset and `@` modifier to stable features as per Prometheus [#10121](https://github.com/prometheus/prometheus/pull/10121).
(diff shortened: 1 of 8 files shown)


matej-g (approved):
Makes sense, looks good @yeya24 :+1:

squat (approved):
I think this is fine, however, I'm a bit worried about hypothetical meltdown-type issues where we have two layers of statefulness on top of one another.

The typical example is where you tunnel a TCP connection over SSH, which itself is built on TCP. Both connections are statefull resulting in unnecessary amount of retries that can actually overload and break the underlying connection or saturate the remote server with too many retried packets.

I can't think of an exact issue but I think there could be some similar subtleties here with two layers of readiness awareness

yeya24:
Yeah at the beginning I was wondering if I can solve it at the k8s side and I opened https://github.com/prometheus-operator/prometheus-operator/pull/4681. However, there are two cases of prometheus pod shutdown: one is the normal rollout and another one is readiness/liveness failed and being killed. I think the first case still have the graceful termination but the second case to remove that store from querier as soon as possible.

yeya24:
Thanks for the quick review!

hanjm:
Useful feature, I am facing same issue.

PR #5258, opened 2022-03-30
Build with Go 1.18

sylr:
Signed-off-by: Sylvain Rabot <sylvain@abstraction.fr>

## Changes

- Upgrade to Go 1.18
- Upgrade golangci-lint to v1.45.2
- Upgrade go orb version in circleci config
- Upgrade docker version in circleci config


GiedriusS:
CI failures seem related? :thinking:

sylr:
@GiedriusS could you relaunch the CI to see if errors persist?

GiedriusS:
Still the same after 2 more runs. This is some weird message :thinking: needs investigation.

sylr:
If anyone wants to take over, be my guest.

sylr:
@bwplotka I believe the problem in the tests is that the prometheus server used in the test use json-iterator/reflect2 versions that do not work with go1.18.

I tried to upgrade them with bingo but go nowhere. Could you take a look ?

sylr:
All right I got it, everything is green 🎉

GiedriusS (approved):

PR #5262, opened 2022-03-31
Add the required base64 encryption step

achede22:
This is a required step that just worked for me, I saw it in a youtube video but there is no reference into the Thanos official documents.

Please Check.

### DIFF
--- a/docs/storage.md
+++ b/docs/storage.md
@@ -257,6 +257,7 @@ config:
   service_account: ""
 ```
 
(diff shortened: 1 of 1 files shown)


metalmatze:
There's also `stringData` which does the base64 encoding on the fly and you don't need to do it manually:
https://kubernetes.io/docs/concepts/configuration/secret/#restriction-names-data

PR #5269, opened 2022-04-07
receive: Better error handling of remote write requests.

bwplotka:

### DIFF
--- a/pkg/receive/handler.go
+++ b/pkg/receive/handler.go
@@ -259,28 +259,27 @@ type replica struct {
 }
 
 func (h *Handler) handleRequest(ctx context.Context, rep uint64, tenant string, wreq *prompb.WriteRequest) error {
+	tLogger := log.With(h.logger, "tenant", tenant)
+
 	// This replica value is used to detect cycles in cyclic topologies.
 	// A non-zero value indicates that the request has already been replicated by a previous receive instance.
 	// For almost all users, this is only used in fully connected topologies of IngestorRouter instances.
 	// For acyclic topologies that use RouterOnly and IngestorOnly instances, this causes issues when replicating data.
-	// See discussion in: https://github.com/thanos-io/thanos/issues/4359
+	// See discussion in: https://github.com/thanos-io/thanos/issues/4359.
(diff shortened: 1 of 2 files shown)


squat (dismissed):

bwplotka:
Ok, rdy to review (:

pkg/receive/handler.go
```diff
@@ -334,21 +340,22 @@ func (h *Handler) receiveHTTP(w http.ResponseWriter, r *http.Request) {
...
+	// this would mean lack of forward compatibility for remote write proto.
 	if len(wreq.Timeseries) == 0 {
-		level.Debug(h.logger).Log("msg", "empty timeseries from client", "tenant", tenant)
+		// TODO(yeya24): Handle remote write metadata.
+		if len(wreq.Metadata) > 0 {
+			// TODO(bwplotka): Do we need this error message?
+			level.Debug(tLogger).Log("msg", "only metadata from client; metadata ingestion not supported; skipping")
+			return
+		}
+		level.Debug(tLogger).Log("msg", "empty remote write request; client bug or newer remote write protocol used?; skipping")
```
PhilipGough, line 352: Does this warrant being logged at a level other than debug? Perhaps info|warn since we would perhaps want to know if this occurs in production where it would likely get buried by the logger?

Or perhaps we wouldn't care, in which case we might as well remove the log entirely?

bwplotka: You can ask the same for all debug logs 🙃

Sometimes we care: When we want to check why we have some many request but not many samples, or maybe we want to let tenant know that he is using to new protocol.
Mosty we don't: Sending newer protocol should not page us during night, this is a user error, than can be buried with logs... 

>

PhilipGough: > You can ask the same for all debug logs

Indeed, this just stood out to me as something that might be useful as an info log but for sure we wouldnt not want to be paged or have a user error spam the logs. Explanation makes sense!

PhilipGough (approved):
lgtm

moadz (commented):
LGTM

PR #5277, opened 2022-04-12
[docs] Update Helm Chart example

ilpianista:

## Changes

Fix url to Helm Chart and remove reference to the deprecated prometheus Helm Chart.

## Verification

I deployed it this way.


ilpianista:
Pinging @onprem and @daixiang0. Free to close this if you think it isn't worth it.

PR #5281, opened 2022-04-14
Use correct filesystem/network path separators when uploading blocks

aknuds1:

## Changes
* `block.upload`: Use correct separators for filesystem paths and object storage paths respectively
* `objstore.UploadDir`: Use correct separators for filesystem paths and object storage paths respectively
* `objstore.UploadDir`: Use `filepath.WalkDir` instead of `filepath.Walk` since the latter is [preferred](https://pkg.go.dev/path/filepath#Walk) since Go 1.16.

## Verification

I've run the test suite.

### DIFF
--- a/pkg/block/block.go
+++ b/pkg/block/block.go
@@ -145,11 +145,11 @@ func upload(ctx context.Context, logger log.Logger, bkt objstore.Bucket, bdir st
 		return errors.Wrap(err, "encode meta file")
 	}
 
-	if err := objstore.UploadDir(ctx, logger, bkt, path.Join(bdir, ChunksDirname), path.Join(id.String(), ChunksDirname)); err != nil {
+	if err := objstore.UploadDir(ctx, logger, bkt, filepath.Join(bdir, ChunksDirname), path.Join(id.String(), ChunksDirname)); err != nil {
 		return cleanUp(logger, bkt, id, errors.Wrap(err, "upload chunks"))
 	}
 
-	if err := objstore.UploadFile(ctx, logger, bkt, path.Join(bdir, IndexFilename), path.Join(id.String(), IndexFilename)); err != nil {
+	if err := objstore.UploadFile(ctx, logger, bkt, filepath.Join(bdir, IndexFilename), path.Join(id.String(), IndexFilename)); err != nil {
 		return cleanUp(logger, bkt, id, errors.Wrap(err, "upload index"))
 	}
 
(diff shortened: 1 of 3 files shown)


squat (approved):
Thanks @aknuds1 this looks good. We should be strict about this everywhere

aknuds1:
@squat do you know why .bingo/prometheus.mod changed? I didn't even notice until checking the PR diff.

aknuds1:
@squat do you know how come the docs check [fails](https://github.com/thanos-io/thanos/runs/6021420600?check_suite_focus=true)? Doesn't look like something I can fix.

aknuds1:
`make test` fails consistently for me BTW, even if I disabled e2e tests, also on the main branch.

squat:
re ran the docs test. this seems like a network flake

squat:
hmm i wonder if this is a flake. seems like it's failing on the reloader, which should be totally unrelated to block uploading code, which is the only big change.

pkg/block/block.go
```diff
@@ -145,11 +145,11 @@ func upload(ctx context.Context, logger log.Logger, bkt objstore.Bucket, bdir st
 		return errors.Wrap(err, "encode meta file")
 	}
 
-	if err := objstore.UploadDir(ctx, logger, bkt, path.Join(bdir, ChunksDirname), path.Join(id.String(), ChunksDirname)); err != nil {
+	if err := objstore.UploadDir(ctx, logger, bkt, filepath.Join(bdir, ChunksDirname), path.Join(id.String(), ChunksDirname)); err != nil {
```
pracucci, line 148: `path.Join()` is correct in this file because on object storages it's always `/`. If I remember correctly, the objstore client API contract is to always use `/`.

_Same comment applies below._

aknuds1: @pracucci but the `srcdir` argument to `objstore.UploadDir` refers to a filesystem directory, not an object storage one?

pracucci: Oh sorry, you're definitely right!

pracucci (approved):
👍  The change makes sense to me.

squat:
doc generation is currently blocked in our CI. PR #5285 should fix this.

GiedriusS:
@aknuds1 please merge the newest `main` into your branch or rebase your branch

aknuds1:
@GiedriusS thanks, done.

squat (approved):
❤️

PR #5285, opened 2022-04-16
docs: fix URLs preventing generation and unblock CI

squat:
This PR fixes the URLs for several links in the docs that are currently breaking CI.

I noticed that CI was failing [0] for PR
https://github.com/thanos-io/thanos/pull/5284 because Ian had changed
his GitHub handle from @ianbillett to @bill3tt. This PR  fixes this.

Currently, documentation generation is failing because mdox can't fetch
some GitHub documentation pages since the URLs for the help content has
changed. This commit updates the links to use the correct URLs.

[0] https://github.com/thanos-io/thanos/runs/6050355497?check_suite_focus=true#step:5:135

### DIFF
--- a/.mdox.validate.yaml
+++ b/.mdox.validate.yaml
@@ -20,3 +20,6 @@ validators:
   # couldn't reach even when curl-ed.
   - regex: 'cloud\.baidu\.com'
     type: 'ignore'
+  # 403 when curl-ed from GitHub actions, though not from a developer machine. Likely due to secondary rate limits.
+  - regex: 'docs\.github\.com'
+    type: 'ignore'
--- a/MAINTAINERS.md
+++ b/MAINTAINERS.md
@@ -31,7 +31,7 @@ Full list of triage persons is displayed below:
 | Name          | Slack         | GitHub                                       | Company |
 |---------------|---------------|----------------------------------------------|---------|
 | Adrien Fillon | `@Adrien F`   | [@adrien-f](https://github.com/adrien-f)     |         |
-| Ian Billett   | `@billett`    | [@ianbillett](https://github.com/ianbillett) | Red Hat |
+| Ian Billett   | `@billett`    | [@bill3tt](https://github.com/bill3tt)       | Red Hat |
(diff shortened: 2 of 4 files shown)


squat:
This PR is failing because the busybox SHAs are incorrect. This is fixed by: https://github.com/thanos-io/thanos/pull/5283.
That PR will likely also fail because of docs, which is fixed by this PR. We will need to override CI for one of them.

squat:
Two of the pages are failing to fetch because GitHub doesn't like that we don't send a user-agent. This causes the GitHub documentation site to respond with a 403. If we send a very basic header like `User-Agent: Chrome`, the request works.

saswatamcode:
Hmm, we do [send a `User-Agent: mdox` request header](https://github.com/bwplotka/mdox/blob/2dc35528ad58696de32ecb9ed1bddde644c5afd1/pkg/mdformatter/linktransformer/config.go#L134) in mdox as the GitHub API has [rules against](https://docs.github.com/en/rest/overview/resources-in-the-rest-api#user-agent-required) that. I think this might be due to some sort of rate limit like https://docs.github.com/en/rest/overview/resources-in-the-rest-api#secondary-rate-limits. A solution could be to just ignore docs.github.com links for now.

bill3tt (approved):
Yikes - I hadn't appreciated that this change would cause CI to break. Thanks @squat 🙌

squat:
Hmm I'm not totally convinced about rate limits. If it were that, then GitHub should return a 429 Too Many Requests. In any case, retrying girl CI job several hours later results in the same errors, making me believe it's really a permissions error and not just a rate limit. We can definitely ignore failed checks for GitHub kinks, even though that's a shame :/

saswatamcode:
I'll try to add a flag to `mdox` to allow showing the actual responses from failed requests. In the meantime maybe docs.github.com links can be ignored?

squat:
Ack, I'll update this PR to ignore those URLs for now

squat:
Urgent to unblock CI
cc @saswatamcode

saswatamcode (approved):
Thanks! 🚀

PR #5295, opened 2022-04-22
v0.25.2: QFE: Don't cache when dedup is false.

bwplotka:
Fixes https://issues.redhat.com/browse/MON-2437

We can discuss later if dedup should be in the cache key, but right now, dedup false and true use exactly the same caching result which is wrong. 

TODO After this is merged: Merge to maint.


yeya24:
I agree it makes sense to add dedup option as part of the cache key. We should maybe add some other options as well.

bwplotka:
PTAL @yeya24 I don't think so it makes really sense. Dedup off is really used as debug option, no?

bwplotka:
But we can consider that at some point, maybe for now safer to skip?

bwplotka:
Let's discuss on Monday/Tuesday, I think I would be ok cache all dedups (imagine rare case of someone avoiding deduplication by default).

cc @moadz @jessicalins @bill3tt

yeya24:
> PTAL @yeya24 I don't think so it makes really sense. Dedup off is really used as debug option, no?

I think there are cases where users want to enable it not only for debugging. But anyway the thing is that we need to differentiate this from the cache and not use the cached deduped result when dedup is disabled

pkg/queryfrontend/queryrange_codec.go
```diff
@@ -180,7 +180,7 @@ func parseDurationMillis(s string) (int64, error) {
 }
 
 func parseEnableDedupParam(s string) (bool, error) {
-	enableDeduplication := true
+	enableDeduplication := true // Deduplication is enabled by default.
```
jessicalins, line 183: I think it would be nice to also update docs here https://thanos.io/tip/components/query.md/#deduplication saying that this option is enabled by default

pkg/queryfrontend/request.go
```diff
@@ -13,12 +13,17 @@ import (
 	"github.com/prometheus/prometheus/model/timestamp"
 )
 
-// TODO(yeya24): add partial result when needed.
-// ThanosRequest is a common interface defined for specific thanos requests.
-type ThanosRequest interface {
+// ThanosRequestStoreMatcherGetter is a an interface for store matching that all request share.
```
jessicalins, line 16:
```suggestion
// ThanosRequestStoreMatcherGetter is a an interface for store matching that all requests share.
```

pkg/queryfrontend/roundtrip.go
```diff
@@ -272,10 +272,20 @@ func newLabelsTripperware(
 	}, nil
 }
 
-// Don't go to response cache if StoreMatchers are set.
+// shouldCache controls what kind of Thanos request should be cached.
+// Don't cache:
+// * if StoreMatchers are set (since it is debug option).
+// * if Dedup is disabled (this is done only for debugging).
+// * Caching option disables cache.
```
jessicalins, line 279: I'm not sure I get what this comment means?

jessicalins (commented):
left a small suggestion, mostly to also update docs if that makes sense :)
I think @bill3tt github's link needs to be updated (so that `check-docs` [check](https://github.com/thanos-io/thanos/runs/6133937791?check_suite_focus=true) passes) and fix `TestQueryFrontendMemcachedCache` [test](https://github.com/thanos-io/thanos/runs/6133937878?check_suite_focus=true)

moadz:
>  It is something that can be configured in the Query globally so you cannot assume it is debugging only
@yeya24 I feel like if a user specifies `--query.replica-label` in the config, they are enforcing dedup across Thanos. If no label is present, even with dedup enabled by default it won't impact results and will still be cached. The query HTTP API does not enforce the dedup query param, so semantically we should always enforce dedup if possible or make the dedup flag compulsory. 

IMHO enforcing dedup so that the `--query.replica-label` is enforced transparently, and amending the docs to reflect this is smallest effective change that will fix this problem permanently.

pkg/queryfrontend/roundtrip.go
moadz: Yup this is confusing, are you referring to `Cache-Control` headers on the request?

moadz (commented):
Missing changelog + doc updates

pkg/queryfrontend/roundtrip.go
bwplotka: Yup

bwplotka:
Douglas will follow up with another PR (:

Issue #4780, opened 2021-10-14
Querier: Deterministic buffered queries


bwplotka:
I plan to give a formal proposal + PoC, but starting an issue to cover the idea to solve the problem we have on the Querier fanout side which creates confusion, annoyance and overall extra latency.

I explained this problem to the team and on the latest Community Hours, but for initial info:

## Problem

Current Querier algorithm looks following:

![image](https://user-images.githubusercontent.com/6950331/137357489-62ddfd24-80dd-48d6-abc6-1721705b3e4c.png)
 
While it's not easy to spot, but what is happening in practice is that `MergeSort` is choosing which storeAPIs to fetch message from one-by-one. On average it's round robin, but that depends on order of series. In practice this looks as follows:

If we do query (`start_timestamp_ms:1626134100000 end_timestamp_ms:1627344000000 matchers:<type:RE name:"__name__" value:"continuous_app_metric9.*" > `) for 2w from 5 different StoreAPIs (I literally used [ `interactive_test`](https://github.com/thanos-io/thanos/blob/e75ce708234384fbf7dfcab13294aaf17aaa18fd/examples/interactive/interactive_test.go#L91)), it takes 6+ seconds on my machine:

![image](https://user-images.githubusercontent.com/6950331/137358036-c24fda0e-ca59-4c9f-accd-e8ecdb1d114f.png)

This test has Jaeger so I added lower granularity spans that tells us exactly wait-times for (https://github.com/bwplotka/thanos/commit/e75ce708234384fbf7dfcab13294aaf17aaa18fd):

* fetching single series message from store: `querier.streamSeriesSet.waitForStoreResult`
* merge sort actually receiving it from this  goroutine: `querier.streamSeriesSet.waitForMergeSort`

What we see for this query is that those go routines are starved for most of the times. E.g this sidecar fetch was waiting 500ms just to be able to pass it's data to merge sort: 
![image](https://user-images.githubusercontent.com/6950331/137358865-5ebca09a-26ab-4d30-a0ab-29d8d87e3619.png)

This means sidecar itself gRPC service was waiting to send its messages too.

This happens for all stores. Essentially computation and slowest Stores causes additive latency which adds up.

And ONLY when we merge all, PromQL can start doing it's job due to https://github.com/prometheus/prometheus/issues/7326

### Consequences:

 * Latency is more or less cumulative, despite fetching concurrently. We essentially could do more concurrency here, since we buffer all messages anyway.
[...]

PR #5296, opened 2022-04-24
store: use k-way merging for proxying logic

GiedriusS:
Use k-way merging for merging multiple responses coming from multiple
StoreAPIs. This avoids having a bottleneck of a Go channel that can only
hold 10 responses at once. So, now we will Recv() messages as fast as
possible. Users should be able to see much quicker queries because
network will be used as much as possible.

Benchmarks:

```
name                                                                                     old time/op    new time/op    delta
ProxySeries/1000000SeriesWith1Samples/4_client_with_1_samples,_250000_series_each-16        1.01s ± 7%     0.24s ± 2%  -76.62%  (p=0.008 n=5+5)
ProxySeries/1000000SeriesWith1Samples/single_client_with_1_samples,_1000000_series-16       926ms ± 3%     454ms ± 2%  -50.99%  (p=0.008 n=5+5)
ProxySeries/100000SeriesWith100Samples/4_client_with_25_samples,_25000_series_each-16      95.6ms ± 1%    23.4ms ± 3%  -75.52%  (p=0.016 n=4+5)
ProxySeries/100000SeriesWith100Samples/single_client_with_100_samples,_100000_series-16    93.1ms ± 2%    45.1ms ± 1%  -51.63%  (p=0.008 n=5+5)
ProxySeries/1SeriesWith10000000Samples/4_client_with_2500000_samples,_1_series_each-16     39.5µs ± 1%    25.2µs ± 1%  -36.22%  (p=0.008 n=5+5)
ProxySeries/1SeriesWith10000000Samples/single_client_with_10000000_samples,_1_series-16    22.6µs ± 0%    15.2µs ± 0%  -32.71%  (p=0.008 n=5+5)

name                                                                                     old alloc/op   new alloc/op   delta
ProxySeries/1000000SeriesWith1Samples/4_client_with_1_samples,_250000_series_each-16        186MB ± 0%     108MB ± 0%  -41.87%  (p=0.008 n=5+5)
ProxySeries/1000000SeriesWith1Samples/single_client_with_1_samples,_1000000_series-16       186MB ± 0%     107MB ± 0%  -42.18%  (p=0.008 n=5+5)
ProxySeries/100000SeriesWith100Samples/4_client_with_25_samples,_25000_series_each-16      18.5MB ± 0%    10.0MB ± 0%  -45.79%  (p=0.008 n=5+5)
ProxySeries/100000SeriesWith100Samples/single_client_with_100_samples,_100000_series-16    18.5MB ± 0%    10.6MB ± 0%  -42.69%  (p=0.008 n=5+5)
ProxySeries/1SeriesWith10000000Samples/4_client_with_2500000_samples,_1_series_each-16     10.9kB ± 0%     6.1kB ± 0%  -44.06%  (p=0.008 n=5+5)
ProxySeries/1SeriesWith10000000Samples/single_client_with_10000000_samples,_1_series-16    5.09kB ± 0%    3.06kB ± 0%  -39.83%  (p=0.008 n=5+5)

name                                                                                     old allocs/op  new allocs/op  delta
ProxySeries/1000000SeriesWith1Samples/4_client_with_1_samples,_250000_series_each-16        5.00M ± 0%     1.00M ± 0%  -80.00%  (p=0.008 n=5+5)
ProxySeries/1000000SeriesWith1Samples/single_client_with_1_samples,_1000000_series-16       5.00M ± 0%     1.00M ± 0%  -80.00%  (p=0.008 n=5+5)
ProxySeries/100000SeriesWith100Samples/4_client_with_25_samples,_25000_series_each-16        500k ± 0%      100k ± 0%  -79.96%  (p=0.008 n=5+5)
ProxySeries/100000SeriesWith100Samples/single_client_with_100_samples,_100000_series-16      500k ± 0%      100k ± 0%  -79.98%  (p=0.016 n=5+4)
ProxySeries/1SeriesWith10000000Samples/4_client_with_2500000_samples,_1_series_each-16        180 ± 0%       125 ± 0%  -30.56%  (p=0.008 n=5+5)
ProxySeries/1SeriesWith10000000Samples/single_client_with_10000000_samples,_1_series-16       109 ± 0%        80 ± 0%  -26.61%  (p=0.008 n=5+5)
```

`TestProxyStore_SeriesSlowStores` also passes which tests how everything works with slow stores.

Rewritten chunk deduplication logic to use hashing.

In the future, we could get some more marginal gains by using a selection/tournament tree.


pkg/store/storepb/custom.go
```diff
@@ -257,6 +258,28 @@ func (s *uniqueSeriesSet) Next() bool {
 	return true
 }
 
+func (m AggrChunk) Hash() string {
```
yeya24, line 261: Seems the hash is a little bit expensive to calculate...

pkg/store/proxy_heap.go
```diff
@@ -0,0 +1,261 @@
...
+		return d.responses[0]
+	}
+
+	chunkDedupMap := map[string]*storepb.AggrChunk{}
+
+	for _, resp := range d.responses {
+		for _, chk := range resp.GetSeries().Chunks {
+			h := chk.Hash()
+
+			if _, ok := chunkDedupMap[h]; !ok {
```
yeya24, line 51: Curious why do we need to do deduplication here? Will exactly the same chunks be returned from the target stores in a single `Series` grpc call?

GiedriusS: To avoid duplicating work when multiple StoreAPIs return the same chunks. This can happen due to misconfiguration. For example, multiple Thanos Store instances connected that use the same bucket.

pkg/store/storepb/custom.go
GiedriusS: Previously we were [comparing each chunk byte-by-byte using recursion and deduplicating this way.](https://github.com/thanos-io/thanos/blob/79e70da702228ac0282fc7639f9f160d922b6dcb/pkg/store/storepb/custom.go#L160-L203) So, in this regard the complexity is the same AIUI - we are _still_ reading each and every byte. Just that instead of comparison, we are now calculating hashes and then using them in a map.

I was thinking that maybe we could use md5 instead here as we don't care about collisions here really? In the very unlikely case of collisions, the upper layer will take care of deduplicating. MD5 is much faster. [We could use this](https://github.com/minio/md5-simd).

bwplotka:
This is pretty sweet 💪🏽 Sounds like this might fix https://github.com/thanos-io/thanos/issues/4780 - will take a look this week or next (:

pkg/store/proxy_heap.go
```diff
@@ -0,0 +1,261 @@
...
+		return nil
+	} else if len(d.responses) == 1 {
+		return d.responses[0]
+	}
+
+	chunkDedupMap := map[string]*storepb.AggrChunk{}
+
+	for _, resp := range d.responses {
+		for _, chk := range resp.GetSeries().Chunks {
+			h := chk.Hash()
```
fpetkovski, line 49: Since this is done serially, does it make sense to reuse one buffer for hashing in order to reduce allocations?

GiedriusS:
Tried this pull request in prod with a query that returns ~21k series - it reduced the query duration by about 30% :exploding_head:

pkg/store/proxy_heap.go
```diff
@@ -0,0 +1,264 @@
...
+	}()
+	if len(d.responses) == 0 {
+		return nil
+	} else if len(d.responses) == 1 {
+		return d.responses[0]
+	}
+
+	chunkDedupMap := map[string]*storepb.AggrChunk{}
+
+	md5Hash := md5.New()
```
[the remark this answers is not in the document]
GiedriusS: ignore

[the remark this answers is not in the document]
GiedriusS: ignore

pkg/store/proxy_heap.go
GiedriusS: Applied your suggestion :+1:

fpetkovski:
I tested this in our infra and can confirm improved performance for queries across lots of series. Dashboards also seem to be loading more smoothly. 

Looking forward to seeing this merged 🚀

pkg/store/proxy.go
```diff
@@ -255,305 +270,209 @@ func (s *ProxyStore) Series(r *storepb.SeriesRequest, srv storepb.Store_SeriesSe
...
-			// We might be able to skip the store if its meta information indicates it cannot have series matching our query.
-			if ok, reason := storeMatches(gctx, st, r.MinTime, r.MaxTime, matchers...); !ok {
-				storeDebugMsgs = append(storeDebugMsgs, fmt.Sprintf("store %s filtered out: %v", st, reason))
-				continue
-			}
+	wg := &sync.WaitGroup{}
 
-			storeDebugMsgs = append(storeDebugMsgs, fmt.Sprintf("Store %s queried", st))
+	storeDebugMsgs := []string{}
+	actualRequest := &storepb.SeriesRequest{
```
bwplotka, line 276: `actualRequest` does not mean much - maybe simply `r` and rename old `r` as `originalRequest`?
```suggestion
	// We can't modify `r`, so let's copy request with new matchers.
	r := &storepb.SeriesRequest{
```

pkg/store/proxy.go
```diff
@@ -236,8 +238,21 @@ func (s cancelableRespSender) send(r *storepb.SeriesResponse) {
...
+	frameTimeoutCtx := context.Background()
+	var cancel context.CancelFunc
+	if responseTimeout != 0 {
+		frameTimeoutCtx, cancel = context.WithTimeout(frameTimeoutCtx, responseTimeout)
+		return frameTimeoutCtx, cancel
+	}
+	return frameTimeoutCtx, func() {}
+}
+
 func (s *ProxyStore) Series(r *storepb.SeriesRequest, srv storepb.Store_SeriesServer) error {
```
bwplotka, line 256:
```suggestion
func (s *ProxyStore) Series(originalRequest *storepb.SeriesRequest, srv storepb.Store_SeriesServer) error {
```

pkg/store/proxy.go
```diff
@@ -255,305 +270,209 @@ func (s *ProxyStore) Series(r *storepb.SeriesRequest, srv storepb.Store_SeriesSe
...
+	}
 
-			// This is used to cancel this stream when one operation takes too long.
-			seriesCtx, closeSeries := context.WithCancel(gctx)
-			seriesCtx = grpc_opentracing.ClientAddContextTags(seriesCtx, opentracing.Tags{
-				"target": st.Addr(),
-			})
-			defer closeSeries()
+	stores := []Client{}
+	for _, st := range s.stores() {
```
bwplotka, line 289: What's wrong with scheduling all here?

pkg/store/proxy.go
```diff
@@ -255,305 +270,209 @@ func (s *ProxyStore) Series(r *storepb.SeriesRequest, srv storepb.Store_SeriesSe
...
-	ctx    context.Context
-	logger log.Logger
-
-	stream storepb.Store_SeriesClient
-	warnCh directSender
+			rCh := make(chan *recvResponse)
+			done := make(chan struct{})
+			go func() {
+				for {
+					// TODO: we can buffer this in files to have infinitely
```
bwplotka, line 346:
```suggestion
					// TODO(GiedriusS): we can buffer this in files to have infinitely
```

Please add your username if you authored TODO, otherwise we don't know how might have more context about.

pkg/store/proxy.go
```diff
@@ -255,305 +270,209 @@ func (s *ProxyStore) Series(r *storepb.SeriesRequest, srv storepb.Store_SeriesSe
...
+					// scalable read path.
+					// TODO: this sorely needs some limit on the number of series
+					// per each request.
+					r, err := cl.Recv()
```
bwplotka, line 350: So you actually got rid of `startStreamSeriesSet`? Nice!

pkg/store/proxy.go
```diff
@@ -255,305 +270,209 @@ func (s *ProxyStore) Series(r *storepb.SeriesRequest, srv storepb.Store_SeriesSe
...
-// streamSeriesSet iterates over incoming stream of series.
-// All errors are sent out of band via warning channel.
-type streamSeriesSet struct {
-	ctx    context.Context
-	logger log.Logger
-
-	stream storepb.Store_SeriesClient
-	warnCh directSender
+			rCh := make(chan *recvResponse)
+			done := make(chan struct{})
```
bwplotka, line 343: To me, we can remove this channel and just use `seriesContext` no?

Less complexity with same functionality

pkg/store/proxy.go
```diff
@@ -255,305 +270,209 @@ func (s *ProxyStore) Series(r *storepb.SeriesRequest, srv storepb.Store_SeriesSe
...
+						close(rCh)
+						return
+					case rCh <- &recvResponse{r: r, err: err}:
+					}
+				}
+			}()
+
+			// The `defer` only executed when function return, we do `defer cancel` in for loop,
+			// so make the loop body as a function, release timers created by context as early.
+			handleRecvResponse := func() (next bool) {
```
bwplotka, line 362: I feel like separate function (normal one) might be more readable. Especially if we get rid of `done` channel. Are you sure this is better? 🤔

pkg/store/proxy.go
```diff
@@ -255,305 +270,209 @@ func (s *ProxyStore) Series(r *storepb.SeriesRequest, srv storepb.Store_SeriesSe
...
-			span.Finish()
-			close(s.recvCh)
-			wg.Done()
-		}()
-
-		numResponses := 0
-		defer func() {
-			if numResponses == 0 {
-				emptyStreamResponses.Inc()
+				storeResponses[i].responses = append(storeResponses[i].responses, rr.r)
```
bwplotka, line 410: Those slices will kill us, we might want to buffer / pool them at some point

pkg/store/proxy.go
```diff
@@ -255,305 +270,209 @@ func (s *ProxyStore) Series(r *storepb.SeriesRequest, srv storepb.Store_SeriesSe
...
-			}
+	level.Debug(reqLogger).Log("msg", "Series: started fanout streams", "status", strings.Join(storeDebugMsgs, ";"))
 
-			if rr.err != nil {
-				s.handleErr(errors.Wrapf(rr.err, "receive series from %s", s.name), done)
-				return false
-			}
-			numResponses++
-			bytesProcessed += rr.r.Size()
+	wg.Wait()
```
bwplotka, line 424: I think this can work for a start, but we can actually start k-merging much sooner - as soon as we have first frames from each store. 🤔 

We can start with this simple solution (which is already better since we buffer ASAP), but let's ensure this code does not need to be refactored full to start merging asynchronously, WDYT?

pkg/store/proxy_heap.go
```diff
@@ -0,0 +1,264 @@
...
+	// If no chunks were requested.
+	if len(chunkDedupMap) == 0 {
+		return storepb.NewSeriesResponse(&storepb.Series{
+			Labels: d.responses[0].GetSeries().Labels,
+			Chunks: d.responses[0].GetSeries().Chunks,
+		})
+	}
+
+	finalChunks := make([]storepb.AggrChunk, 0, len(chunkDedupMap))
+
```
bwplotka, line 71:
```suggestion
```

pkg/store/proxy_heap.go
```diff
@@ -0,0 +1,264 @@
...
+	return len(d.responses) > 0 || d.previousNext
+}
+
+// ProxyResponseHeap is a heap for storepb.SeriesSets.
+// It performs k-way merge between all of those sets.
+// TODO(GiedriusS): can be improved with a tournament tree.
+// This is O(n*logk) but can be Theta(n*logk). However,
+// tournament trees need n-1 auxiliary nodes so there
+// might not be much of a difference.
+type ProxyResponseHeap []ProxyResponseHeapNode
```
bwplotka, line 159: Integrated with go Heap! Amazing 🤩

pkg/store/proxy_test.go
```diff
@@ -1025,6 +1025,10 @@ func TestProxyStore_SeriesSlowStores(t *testing.T) {
 			return
 		}
 	}
+
+	// Wait until the last goroutine exits which is stuck on time.Sleep().
+	// Otherwise, goleak complains.
+	time.Sleep(5 * time.Second)
```
bwplotka, line 1031: Yea, I think that might be actually something we do wrong. Why we have to wait? What is not cleaned  5 seconds after the request. I don't think we can afford those eventual leaks in processes that will make dozens or hundreds QPS 🤔 

I think this is connected to the comment I made around `frameCtx`?

bwplotka (approved):
Thanks @GiedriusS for leading this work, amazing stuff!

I think we could switch to this, but there are two points to remember: 

1.  I think the efficiency boost (e.g latency) is caused not by a different merge algorithm, but actually by buffering all messages ASAP (which we [agreed makes sense](https://github.com/thanos-io/thanos/issues/4780))

We could trigger an infinite (kind of) buffer by just changing this channel length to something large (e.g million).. which is obv bad, because we would always allocate that many.

image

My point is: In this PR with buffering all messages change (which is amazing), we ALSO changed to k-merge which might be actually slower/worse than the previous MergeSort. We don't know as we changed two things  in the same time here. 

2. We still have a bottleneck on waiting for ALL messages until we start streaming merged responses. I think we can do k-merge in more asynchronous way. My worry is that if we want to implement it we will have to refactor this code 3rd time 🙃 

Maybe those 2 things are fine to agree to. I would be keen on merging this as it's simple and have immediate value, so approving. I added some readability suggestions (nits).

bwplotka (changes requested):
Well, I think we also have to figure out this: https://github.com/thanos-io/thanos/pull/5296#discussion_r865966619

This feels like a future problem if we ignore it 🤔

pkg/store/proxy.go
GiedriusS: Because later on, we do:
```
		storeResponses = make([]struct {
			responses []*storepb.SeriesResponse
		}, len(stores))
```

To avoid reallocating slices. And, IMHO it is more readable.

pkg/store/proxy.go
```diff
@@ -236,8 +238,21 @@ func (s cancelableRespSender) send(r *storepb.SeriesResponse) {
...
 
-// Series returns all series for a requested time range and label matcher. Requested series are taken from other
-// stores and proxied to RPC client. NOTE: Resulted data are not trimmed exactly to min and max time range.
+type recvResponse struct {
+	r   *storepb.SeriesResponse
+	err error
+}
+
+func frameCtx(responseTimeout time.Duration) (context.Context, context.CancelFunc) {
+	frameTimeoutCtx := context.Background()
```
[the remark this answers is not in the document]
GiedriusS: Good point, will do.

pkg/store/proxy.go
GiedriusS: Yes but it's not trivial, we've tried to migrate away from gogoprotobuf but that work is still not finished/stuck. With gogoprotobuf, it's impossible to do pooling without writing some kind of custom plugin / or forking it. Because it's unmaintained, it's not worth investing more time into gogoprotobuf.

pkg/store/proxy.go
GiedriusS: The reason this is a function is that the scope of `defer` is the outer function. So, if it would be embedded in the function itself then we would have a bunch of timers created via context.WithTimeout. Since this code is closely related to the outer code, I'd rather leave it as is.

bwplotka (approved):
Nice! Amazing. Thanks for making lazy response buffering 👍🏽 

Wonder what's the benchmarks are now. Happy to merge!

GiedriusS:
Please don't merge this yet - even though the code works and the benchmarks look even better (will post tomorrow) but I think there's some lock contention going. So, let me come back with some final tweaks and a bigger comment.

GiedriusS:
Mhm, it seems like it is impossible to implement this properly because gRPC Recv() doesn't take any context variable, it uses the stream's context. Thus, to implement a per-request timeout, you need to spawn a separate goroutine. This implies that you will need a channel in the middle to send responses back to the original goroutine or you'll need to synchronize somehow differently. Because our goal is to Recv() as fast as possible, it means that we must avoid having a channel in the middle. We could do a simple loop recv -> append in the main goroutine but if Recv() blocks indefinitely then it would block everything going after it so two separate receiver/sender goroutines are necessary. https://github.com/grpc/grpc-go/issues/1229#issuecomment-302755717 this is a real pity that this is the accepted solution, to spawn a new goroutine for each Recv() operation. I'll try to think about this more.

bwplotka:
Any progress?

bwplotka:
Anything we can do to help @GiedriusS ? (:

GiedriusS:
IMHO we should merge the lazy approach as the most potential gains in the eager one could come from https://github.com/grpc/grpc-go/issues/5401 i.e. not using a separate goroutine. However, that means we would lose per-RecvMsg() timeout so it's not feasible. What do you think about simply merging the completely lazy approach and then working on making Prometheus PromQL engine stream the samples?

squat:
Who knows when grpc/grpc-go#5401 could happen. I would agree on merging to take advantage now.

fpetkovski:
I am also happy to see this shipped in its current incarnation.

GiedriusS:
Updated the PR, I think it's ready to go:

* Added back the series statistics that I had removed accidentally;
* Added some unit tests for the deduplicating heap;
* Implemented the timeout for each Recv() call using `time.Timer`

The only caveat of these changes is that in case of a timeout, the user will see "context canceled" instead of "failed to receive any data in $DURATION" because the timer cancels the context after the timeout.

Benchmarks:

```go
name                                                                                     old time/op    new time/op    delta
ProxySeries/1000000SeriesWith1Samples/4_client_with_1_samples,_250000_series_each-16        990ms ± 5%     183ms ± 7%   -81.49%  (p=0.000 n=14+15)
ProxySeries/1000000SeriesWith1Samples/single_client_with_1_samples,_1000000_series-16       958ms ± 5%     255ms ± 2%   -73.39%  (p=0.000 n=15+13)
ProxySeries/100000SeriesWith100Samples/4_client_with_25_samples,_25000_series_each-16       101ms ± 5%      19ms ± 3%   -81.43%  (p=0.000 n=15+14)
ProxySeries/100000SeriesWith100Samples/single_client_with_100_samples,_100000_series-16    96.4ms ± 2%    25.6ms ± 1%   -73.41%  (p=0.000 n=12+13)
ProxySeries/1SeriesWith10000000Samples/4_client_with_2500000_samples,_1_series_each-16     39.8µs ± 3%    28.5µs ± 5%   -28.55%  (p=0.000 n=14+15)
ProxySeries/1SeriesWith10000000Samples/single_client_with_10000000_samples,_1_series-16    23.2µs ± 2%    17.2µs ± 4%   -25.99%  (p=0.000 n=15+15)

name                                                                                     old alloc/op   new alloc/op   delta
ProxySeries/1000000SeriesWith1Samples/4_client_with_1_samples,_250000_series_each-16        186MB ± 0%      84MB ± 0%   -54.79%  (p=0.000 n=15+15)
ProxySeries/1000000SeriesWith1Samples/single_client_with_1_samples,_1000000_series-16       186MB ± 0%      83MB ± 0%   -55.11%  (p=0.000 n=14+15)
ProxySeries/100000SeriesWith100Samples/4_client_with_25_samples,_25000_series_each-16      18.5MB ± 0%     7.6MB ± 0%   -58.75%  (p=0.000 n=15+14)
ProxySeries/100000SeriesWith100Samples/single_client_with_100_samples,_100000_series-16    18.5MB ± 0%     8.2MB ± 0%   -55.66%  (p=0.000 n=15+15)
ProxySeries/1SeriesWith10000000Samples/4_client_with_2500000_samples,_1_series_each-16     11.3kB ± 0%     8.3kB ± 0%   -26.39%  (p=0.000 n=11+15)
ProxySeries/1SeriesWith10000000Samples/single_client_with_10000000_samples,_1_series-16    5.20kB ± 0%    3.52kB ± 0%   -32.23%  (p=0.000 n=15+15)

name                                                                                     old allocs/op  new allocs/op  delta
ProxySeries/1000000SeriesWith1Samples/4_client_with_1_samples,_250000_series_each-16        5.00M ± 0%     0.00M ± 0%   -99.99%  (p=0.000 n=15+15)
ProxySeries/1000000SeriesWith1Samples/single_client_with_1_samples,_1000000_series-16       5.00M ± 0%     0.00M ± 0%  -100.00%  (p=0.000 n=13+15)
ProxySeries/100000SeriesWith100Samples/4_client_with_25_samples,_25000_series_each-16        500k ± 0%        0k ± 0%   -99.95%  (p=0.000 n=15+15)
ProxySeries/100000SeriesWith100Samples/single_client_with_100_samples,_100000_series-16      500k ± 0%        0k ± 0%   -99.97%  (p=0.000 n=15+14)
ProxySeries/1SeriesWith10000000Samples/4_client_with_2500000_samples,_1_series_each-16        185 ± 0%       144 ± 0%   -22.16%  (p=0.000 n=15+15)
ProxySeries/1SeriesWith10000000Samples/single_client_with_10000000_samples,_1_series-16       111 ± 0%        81 ± 0%   -27.03%  (p=0.000 n=15+15)
```

With k6 I have tested on a real-life Thanos with Query+Sidecar. A query that touches ~25M series:

```
     http_req_duration..............: avg=9.53s   min=8.93s   med=9.19s   max=11.3s    p(90)=10.99s  p(95)=11.15s  
```

vs. vanilla Thanos

```
     http_req_duration..............: avg=9.72s    min=8.97s   med=9.29s   max=12.54s  p(90)=10.64s   p(95)=11.59s 
```

So, the improvement is around 5%. Still not bad :smile:

pkg/store/proxy.go
GiedriusS: Refactored so that it would be a one-line change to switch to lazy evaluation.

fpetkovski:
I deployed this in one of our environments, and I saw a drop in the p90 for gRPC stream duration.
image

However, I also noticed a drop in series selects per second.
image

Would the latter be a concern, or an artefact of the new implementation?

GiedriusS:
@fpetkovski that's a big drop :scream: previously we were doing _everything_ concurrently but with the eager version we wait for all of the results at first and then merge everything. Perhaps one of the Series() calls during proxying fails and thus the other ones don't happen? Is there anything in the logs? Also, I wonder if it would help to switch to lazy evaluation because that would be much closer to the previous behavior.

fpetkovski:
Yes, I can try this again next week, currently I am OOO. 

The drop seems to be exactly 50%, so I was wondering if we were either doing something more efficiently :)

fpetkovski:
I tried the lazy approach and the `grpc_client_handled_total` metric was almost the same as in the eager approach. However, I did feel like the individual queries were faster with the lazy strategy. 

I don't have an explanation for why the metric would drop in half, but I also don't see an actual issue. There was no increase in the error rate and the `thanos_query_concurrent_selects_gate_duration_seconds_count` metric, which measures how many series selects we are doing, stayed the same as on the main branch.

So I'm happy with seeing this merged whenever others feel it's good to go.

fpetkovski:
Sorry actually the Lazy implementation does have some issue. In the logs I get see:
```
Internal desc = runtime error: invalid memory address or nil pointer dereference" msg="returning partial response"
```

This is when the querier is used as a store for another (root level) querier.

pkg/store/proxy_heap.go
```diff
@@ -0,0 +1,780 @@
...
+	// warning at a time that gets handled in the beginning.
+	lbls := d.responses[0].GetSeries().Labels
+
+	return storepb.NewSeriesResponse(&storepb.Series{
+		Labels: lbls,
+		Chunks: finalChunks,
+	})
+}
+
+func (d *dedupResponseHeap) Next() bool {
```
bwplotka, line 111: Small nit, but can we have `Next` implementation before `At`? Usually we want to see Next impl first

bwplotka (approved):
The benchmark results 😱 😱 😱 



> The only caveat of these changes is that in case of a timeout, the user will see "context canceled" instead of "failed to receive any data in $DURATION" because the timer cancels the context after the timeout.

Yea, this might be a bit confusing, so we could fix this message in later PRs (plus mention it in changelog?)

Looks amazing. Thanks for implementing two choices.

Can we add a quick flag to enable lazy one via Querier flag?

pkg/store/proxy.go
```diff
@@ -263,346 +258,88 @@ func (s *ProxyStore) Series(r *storepb.SeriesRequest, srv storepb.Store_SeriesSe
...
-	s.currSeries, ok = <-s.recvCh
-	return ok
-}
 
-func (s *streamSeriesSet) At() (labels.Labels, []storepb.AggrChunk) {
-	if s.currSeries == nil {
-		return nil, nil
-	}
-	return s.currSeries.PromLabels(), s.currSeries.Chunks
+	return nil
```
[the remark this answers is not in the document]
bwplotka: I think that's important one!

bwplotka:
We don't check one error, so this might some intermediate issue we accidentally pass as success?

pkg/store/proxy_heap.go
```diff
@@ -0,0 +1,780 @@
...
+	ret.wg.Add(1)
+
+	// Start a goroutine and immediately buffer everything.
+	go func(st Client, l *eagerRespSet) {
+		seriesStats := &storepb.SeriesStatsCounter{}
+		bytesProcessed := 0
+
+		defer func() {
+			if shardInfo != nil {
+				level.Info(logger).Log("msg", "Done fetching series",
```
fpetkovski, line 653: Just found this here, best to remove it on time 😄

pkg/store/proxy_heap.go
```diff
@@ -0,0 +1,780 @@
...
+		shardMatcher:         shardMatcher,
+	}
+
+	go func(st Client, l *lazyRespSet) {
+		bytesProcessed := 0
+		seriesStats := &storepb.SeriesStatsCounter{}
+
+		defer func() {
+			if shardInfo != nil {
+				level.Info(logger).Log("msg", "Done fetching series",
```
fpetkovski, line 391: Same with this log

pkg/store/proxy_heap.go
```diff
@@ -0,0 +1,780 @@
...
+		previousNext: h.Next(),
+	}
+}
+
+func (d *dedupResponseHeap) Err() error {
+	return d.h.Error()
+}
+
+func (d *dedupResponseHeap) At() *storepb.SeriesResponse {
+	defer func() {
```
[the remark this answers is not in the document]
GiedriusS: Good point, removed it. No need for it.

pkg/store/proxy.go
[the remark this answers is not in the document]
GiedriusS: Actually, it's not necessary because we convert *all* errors into warning responses and then the Series() function takes a decision whether to convert it into an error or not :upside_down_face:

GiedriusS:
Updated PR:

* Fixed everything according to comments;
* Remove `Error()` method to not mislead readers - we actually convert all errors into warnings now and then the Series() function itself decides whether those warnings are errors or not depending on what partial response strategy has been set;
* Added a hidden parameter for setting the proxying strategy and used it in e2e tests to test how it works;
* Tried to repro @fpetkovski's case in e2e tests but cannot reproduce it;
* Brought back `could not receive data from X StoreAPI in Y seconds` - we now check whether the timer has fired or not if the error is `context.Canceled`. If that's true then it means that the timer has fired and that the per-Recv() timeout has been reached.

Eager strategy benchmarks:
```
name                                                                                     old time/op    new time/op    delta
ProxySeries/1000000SeriesWith1Samples/4_client_with_1_samples,_250000_series_each-16        1.03s ± 5%     0.18s ± 5%   -82.34%  (p=0.000 n=10+10)
ProxySeries/1000000SeriesWith1Samples/single_client_with_1_samples,_1000000_series-16       955ms ± 4%     260ms ±14%   -72.76%  (p=0.000 n=10+10)
ProxySeries/100000SeriesWith100Samples/4_client_with_25_samples,_25000_series_each-16      96.2ms ± 3%    18.7ms ± 5%   -80.54%  (p=0.000 n=8+9)
ProxySeries/100000SeriesWith100Samples/single_client_with_100_samples,_100000_series-16     117ms ±22%      25ms ± 2%   -79.03%  (p=0.000 n=10+8)
ProxySeries/1SeriesWith10000000Samples/4_client_with_2500000_samples,_1_series_each-16     42.4µs ± 9%    30.8µs ± 8%   -27.42%  (p=0.000 n=10+10)
ProxySeries/1SeriesWith10000000Samples/single_client_with_10000000_samples,_1_series-16    23.6µs ± 4%    17.6µs ± 2%   -25.42%  (p=0.000 n=10+10)

name                                                                                     old alloc/op   new alloc/op   delta
ProxySeries/1000000SeriesWith1Samples/4_client_with_1_samples,_250000_series_each-16        186MB ± 0%      84MB ± 0%   -54.79%  (p=0.000 n=9+10)
ProxySeries/1000000SeriesWith1Samples/single_client_with_1_samples,_1000000_series-16       186MB ± 0%      83MB ± 0%   -55.11%  (p=0.000 n=9+10)
ProxySeries/100000SeriesWith100Samples/4_client_with_25_samples,_25000_series_each-16      18.5MB ± 0%     7.6MB ± 0%   -58.75%  (p=0.000 n=10+10)
ProxySeries/100000SeriesWith100Samples/single_client_with_100_samples,_100000_series-16    18.5MB ± 0%     8.2MB ± 0%   -55.66%  (p=0.000 n=9+10)
ProxySeries/1SeriesWith10000000Samples/4_client_with_2500000_samples,_1_series_each-16     11.3kB ± 0%     8.1kB ± 0%   -28.10%  (p=0.000 n=9+9)
ProxySeries/1SeriesWith10000000Samples/single_client_with_10000000_samples,_1_series-16    5.20kB ± 0%    3.47kB ± 0%   -33.15%  (p=0.000 n=10+10)

name                                                                                     old allocs/op  new allocs/op  delta
ProxySeries/1000000SeriesWith1Samples/4_client_with_1_samples,_250000_series_each-16        5.00M ± 0%     0.00M ± 1%   -99.99%  (p=0.000 n=9+10)
ProxySeries/1000000SeriesWith1Samples/single_client_with_1_samples,_1000000_series-16       5.00M ± 0%     0.00M ± 1%  -100.00%  (p=0.000 n=10+10)
ProxySeries/100000SeriesWith100Samples/4_client_with_25_samples,_25000_series_each-16        500k ± 0%        0k ± 0%   -99.95%  (p=0.000 n=10+8)
ProxySeries/100000SeriesWith100Samples/single_client_with_100_samples,_100000_series-16      500k ± 0%        0k ± 0%   -99.97%  (p=0.000 n=9+9)
ProxySeries/1SeriesWith10000000Samples/4_client_with_2500000_samples,_1_series_each-16        185 ± 0%       144 ± 0%   -22.16%  (p=0.000 n=10+10)
ProxySeries/1SeriesWith10000000Samples/single_client_with_10000000_samples,_1_series-16       111 ± 0%        81 ± 0%   -27.03%  (p=0.000 n=10+10)

```

Lazy strategy benchmarks:
```
name                                                                                     old time/op    new time/op    delta
ProxySeries/1000000SeriesWith1Samples/4_client_with_1_samples,_250000_series_each-16        1.03s ± 5%     0.19s ± 3%  -81.92%  (p=0.000 n=10+10)
ProxySeries/1000000SeriesWith1Samples/single_client_with_1_samples,_1000000_series-16       955ms ± 4%     274ms ± 2%  -71.29%  (p=0.000 n=10+10)
ProxySeries/100000SeriesWith100Samples/4_client_with_25_samples,_25000_series_each-16      96.2ms ± 3%    19.3ms ± 5%  -79.89%  (p=0.000 n=8+10)
ProxySeries/100000SeriesWith100Samples/single_client_with_100_samples,_100000_series-16     117ms ±22%      28ms ± 2%  -76.16%  (p=0.000 n=10+9)
ProxySeries/1SeriesWith10000000Samples/4_client_with_2500000_samples,_1_series_each-16     42.4µs ± 9%    30.4µs ± 1%  -28.33%  (p=0.000 n=10+9)
ProxySeries/1SeriesWith10000000Samples/single_client_with_10000000_samples,_1_series-16    23.6µs ± 4%    17.3µs ± 3%  -26.50%  (p=0.000 n=10+10)

name                                                                                     old alloc/op   new alloc/op   delta
ProxySeries/1000000SeriesWith1Samples/4_client_with_1_samples,_250000_series_each-16        186MB ± 0%      81MB ± 1%  -56.17%  (p=0.000 n=9+10)
ProxySeries/1000000SeriesWith1Samples/single_client_with_1_samples,_1000000_series-16       186MB ± 0%      77MB ± 5%  -58.43%  (p=0.000 n=9+10)
ProxySeries/100000SeriesWith100Samples/4_client_with_25_samples,_25000_series_each-16      18.5MB ± 0%     7.4MB ± 1%  -60.27%  (p=0.000 n=10+10)
ProxySeries/100000SeriesWith100Samples/single_client_with_100_samples,_100000_series-16    18.5MB ± 0%     6.9MB ± 4%  -62.51%  (p=0.000 n=9+10)
ProxySeries/1SeriesWith10000000Samples/4_client_with_2500000_samples,_1_series_each-16     11.3kB ± 0%     8.4kB ± 0%  -25.54%  (p=0.000 n=9+9)
ProxySeries/1SeriesWith10000000Samples/single_client_with_10000000_samples,_1_series-16    5.20kB ± 0%    3.55kB ± 0%  -31.77%  (p=0.000 n=10+10)

name                                                                                     old allocs/op  new allocs/op  delta
ProxySeries/1000000SeriesWith1Samples/4_client_with_1_samples,_250000_series_each-16        5.00M ± 0%     0.00M ±52%  -99.99%  (p=0.000 n=9+9)
ProxySeries/1000000SeriesWith1Samples/single_client_with_1_samples,_1000000_series-16       5.00M ± 0%    0.00M ±107%  -99.98%  (p=0.000 n=10+9)
ProxySeries/100000SeriesWith100Samples/4_client_with_25_samples,_25000_series_each-16        500k ± 0%        0k ±13%  -99.93%  (p=0.000 n=10+10)
ProxySeries/100000SeriesWith100Samples/single_client_with_100_samples,_100000_series-16      500k ± 0%        0k ±72%  -99.91%  (p=0.000 n=9+10)
ProxySeries/1SeriesWith10000000Samples/4_client_with_2500000_samples,_1_series_each-16        185 ± 0%       148 ± 0%  -20.00%  (p=0.000 n=10+10)
ProxySeries/1SeriesWith10000000Samples/single_client_with_10000000_samples,_1_series-16       111 ± 0%        82 ± 0%  -26.13%  (p=0.000 n=10+10)
```

```

**Review by fpetkovski — APPROVED** [review]
I am happy to see this merged in the current state. The issue I noticed might be specific to our environment, so if I can troubleshoot it myself if it happens again.

**Comment by fpetkovski** [comment]
Ha, the e2e test failure seems to be exactly what I saw :)

**Comment by GiedriusS** [comment]
> Ha, the e2e test failure seems to be exactly what I saw :)

Yeah 👍 trying to fix it and then we can merge this

**Comment by GiedriusS** [comment]
I think I have fixed all of the bugs. PTAL @bwplotka @fpetkovski @yeya24 :pray:

File: pkg/store/proxy_heap.go [hunk 23]
```diff
@@ -0,0 +1,743 @@
...
+			continue
+		}
+		for _, chk := range resp.GetSeries().Chunks {
+			for _, field := range []*storepb.Chunk{
+				chk.Raw, chk.Count, chk.Max, chk.Min, chk.Sum, chk.Counter,
+			} {
+				if field == nil {
+					continue
+				}
+				h := xxhash.Sum64(field.Data)
```
fpetkovski, line 133: As a follow up improvement, would it make sense to do this hashing in the stores so that we offload some work from the querier proxy? We might also think about other operations that can be pre-computed at the store level.

bwplotka (approved):
I love it! @fpetkovski also had amazing improvements already with eager option 💪🏽 

Thanks for addressing all. Let's go! 👍🏽

PR #5301, opened 2022-04-26
Avoid starting goroutines for memcached batch requests before gate

### LINKED ISSUE #4967: Memcached client can start high numbers of goroutines
[this behaviour was observed with the code inside Cortex, but I believe it's called the same way inside Thanos]

This code: https://github.com/thanos-io/thanos/blob/397fcb19d93b2a279b57ad10ff1aea811f62345c/pkg/cacheutil/memcached_client.go#L460-L482

starts a goroutine for each batch requested, then if the admin has set a bound on concurrency they will mostly block here:
https://github.com/thanos-io/thanos/blob/397fcb19d93b2a279b57ad10ff1aea811f62345c/pkg/cacheutil/memcached_client.go#L503-L510

For large queries there can be millions of keys here: https://github.com/thanos-io/thanos/blob/8f278cdb5793684c71ab67dd4d75c7cabec887bc/pkg/store/bucket.go#L2198-L2200

We run with `max-get-multi-concurrency=100` and `max-get-multi-batch-size=100`
so we can have hundreds of thousands of goroutines doing nothing.

I think it would be better to avoid starting the goroutines until the gate will allow them to proceed, or perhaps run a pool of goroutines to do the work.

56quarters:
Signed-off-by: Nick Pillitteri <nick.pillitteri@grafana.com>

## Changes

Use the doWithBatch function to avoid starting goroutines to fetch batched
results from memcached before they are allowed to run via the concurrency
Gate. This avoids starting many goroutines which cannot make any progress
due to a concurrency limit.

Fixes #4967

## Verification

Unit tests run. 

### DIFF
--- a/pkg/cacheutil/cacheutil.go
+++ b/pkg/cacheutil/cacheutil.go
@@ -1,0 +1,42 @@
+// Copyright (c) The Thanos Authors.
+// Licensed under the Apache License 2.0.
+
+package cacheutil
+
+import (
+	"context"
+
+	"golang.org/x/sync/errgroup"
+
+	"github.com/thanos-io/thanos/pkg/gate"
+)
+
+// doWithBatch do func with batch and gate. batchSize==0 means one batch. gate==nil means no gate.
+func doWithBatch(ctx context.Context, totalSize int, batchSize int, ga gate.Gate, f func(startIndex, endIndex int) error) error {
+	if totalSize == 0 {
+		return nil
+	}
+	if batchSize <= 0 {
+		return f(0, totalSize)
+	}
+	g, ctx := errgroup.WithContext(ctx)
+	for i := 0; i < totalSize; i += batchSize {
+		j := i + batchSize
+		if j > totalSize {
+			j = totalSize
+		}
+		if ga != nil {
+			if err := ga.Start(ctx); err != nil {
+				return nil
+			}
+		}
+		startIndex, endIndex := i, j
+		g.Go(func() error {
+			if ga != nil {
+				defer ga.Done()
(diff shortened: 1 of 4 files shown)


56quarters:
The failed end-to-end test `TestMetadataAPI_Fanout` passes locally and so might be a flake?

wiardvanrij:
> The failed end-to-end test `TestMetadataAPI_Fanout` passes locally and so might be a flake?

Yea it was (:  Re-ran them.

wiardvanrij (approved):
I noticed I said something in that issue as well a while ago so I understand what we are solving here. So, LGTM but let's wait for a second pair of eyes. - And thanks for doing this!

GiedriusS (approved):
:+1: thank you

56quarters:
I have identified an issue introduced by this change. I will follow up with another PR later today. If the context provided to `getMultiBatched` is cancelled, not all batches will run and therefore the later reads from the `results` channel will block. The solution is using `select` to wait for `results` or `ctx.Done()`

56quarters:
Fix is available here: https://github.com/thanos-io/thanos/pull/5314

Thank you for your patience!

wiardvanrij:
@56quarters 
I thought I also pulled https://github.com/thanos-io/thanos/pull/5301 into the 0.26-rc0, but I actually did not. I noticed this when trying to cherry pick the fix for the issue introduced from the PR in rc1. Therefore, we just ignore this issue for this release.. it will be part of 0.27 now. 
It will be a small delay but then we can release it properly :)

PR #5324, opened 2022-05-02
reloader: Force trigger reload when config rollbacked

heylongdacoder:
Signed-off-by: heylongdacoder <heylongdacoder@gmail.com>

## Changes

Force reload when previous triggerReload failed

Related issue: https://github.com/prometheus-operator/prometheus-operator/issues/4708

## Verification

Added TestReloader_ConfigRollback test case

### DIFF
--- a/pkg/reloader/reloader.go
+++ b/pkg/reloader/reloader.go
@@ -96,6 +96,7 @@ type Reloader struct {
 
 	lastCfgHash         []byte
 	lastWatchedDirsHash []byte
+	forceReload         bool
 
 	reloads                    prometheus.Counter
 	reloadErrors               prometheus.Counter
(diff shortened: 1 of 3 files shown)


pkg/reloader/reloader_test.go
```diff
@@ -167,6 +167,119 @@ config:
...
+		case <-ctx.Done():
+			t.Fatalf("Timeout with faulty = %t, reloadsSeen = %d", faulty, reloadsSeen)
+		case <-time.After(300 * time.Millisecond):
+		}
+
+		rel := reloads.Load().(int)
+		reloadsSeen = rel
+
+		if reloadsSeen == 1 && !faulty {
+			// Initial apply seen (without doing nothing).
```
yeya24, line 254: Did you mean `without doing anything`?

pkg/reloader/reloader.go
```diff
@@ -363,11 +364,13 @@ func (r *Reloader) apply(ctx context.Context) error {
 		}
 		r.reloads.Inc()
 		if err := r.triggerReload(ctx); err != nil {
+			r.forceReload = true
```
yeya24, line 367: Even if we set it to true, this condition is not used during the retry process. Why we don't set it on line 385?

pkg/reloader/reloader_test.go
heylongdacoder: yup, you are right, I have changed it :)

pkg/reloader/reloader.go
heylongdacoder: you are right, I have moved it to line 385 :)

yeya24 (commented):
I am good with this. Maybe @simonpasquier can you please help review this?

simonpasquier:
lgtm

yeya24 (approved):
Thanks for the contribution!

PR #5333, opened 2022-05-05
receive: Implement head series limits per tenant

onprem:

## Changes

Add the ability to limit the total number of series in head per tenant per receive replica. The limit can be specified as an additional HTTP header (`X-Thanos-Series-Limit` by default) on the remote write request.

As mentioned above, this limit is per tenant, per replica, which means that with the current implementation, the tenant can actually write more series in total than specified in the limit. For example, in a hashring with 3 instances of Receive, with a series limit of `100`, the tenant can actually write at most `300` active series (assuming equal distribution of series between all 3 nodes). The actual limit can be less than `300` in case the load is not equally distributed and one node hits the limit earlier, but it will never be less than `100`.

### Alternative

An alternative approach is to calculate an effective `limit` by using the `replication factor` and number of nodes in the hashring. The limit actually used for local tsdb writes can be ` defined_limit * replication_factor / num_nodes`.

So essentially if the over all limit we want is `150` and we have `3` nodes with replication factor `1`, we will have a per node local limit of `50` (`150 * 1 / 3`).

But this assumes that the data is equally distributed among all nodes, but in practice, it can happen that one node hits the `50` series limit earlier than other nodes, effectively denying the service before the user actually hit the `150` series limit.

## Verification

Tested locally with multiple configurations of Receive including split Router and Ingester mode.


douglascamata:
Please, do not forget to add some documentation about this. 🙏 

Also, I couldn't find a good definition in Prometheus or Thanos docs of what is considered an "active series". This hides key information from the reader and I believe we should add a link to it in the docs of this feature.

I found this one below in [Prometheus 1.18 storage documentation](https://prometheus.io/docs/prometheus/1.8/storage/), but it's not easy to understand (when is a chunk closed?).

> A series with an open head chunk is called an active series

[This blog post from Grafana](https://grafana.com/docs/grafana-cloud/fundamentals/active-series-and-dpm/#:~:text=For%20Prometheus,%20a%20time%20series,within%20the%20last%20eight%20hours) seems to have an easier to understand definition:

> For Prometheus, a time series is considered active if new data points have been generated within the last 15 to 30 minutes.

Maybe we should open an issue for Prometheus to define what's considered an "active series"?

onprem:
@wiardvanrij I think what you said makes sense, and we might not even need to make some shared state, if we can somehow utilise the replication step to share this info. I'll explore that a bit and report here if it doesn't work out.

@douglascamata Yes! we need good documentation around it, especially with the tradeoffs that we are making. The definition of "active" series in this PR's context is simple. It's the number of series currently in the head block of the tenant's `tsdb` instance. So basically the number of unique series ingested since the last block was cut (i.e 2 hours window with default settings). I think the grafana definition might be same as well, it might just be that they cut new block every 15 to 30 minutes?

pkg/receive/writer.go
```diff
@@ -81,6 +82,14 @@ func (r *Writer) Write(ctx context.Context, tenantID string, wreq *prompb.WriteR
 			lset = labelpb.ZLabelsToPromLabels(t.Labels)
 		}
 
+		// If the ref is 0, it indicates that inserting the samples will create a new time series.
+		// We do the seriesLimit check before actually creating the series in head because even after
+		// a rollback, the series will stay in head, defeating the whole purpose of series limit.
+		if ref == 0 && seriesLimit > 0 && s.NumSeries() >= seriesLimit {
```
yeya24, line 88: Can we move this to the for loop on line 78? I think we don't need to allocate memory for labels in this case.

pkg/receive/writer.go
```diff
@@ -81,6 +82,14 @@ func (r *Writer) Write(ctx context.Context, tenantID string, wreq *prompb.WriteR
...
+			_ = app.Rollback()
```
yeya24, line 89: What are we trying to rollback in this case? We only check whether the series is new in TSDB head and we don't append anything. Do we still need this line?

onprem: Because we don't want to do partial writes, so I am explicitly rolling back the previous appends. I think due to the early return `app.Commit()` won't be called, so no partial writes anyways, but still calling `app.Rollback()` is looks like a nice idea to me as it will close the `Appender`.

pkg/receive/writer_test.go
```diff
@@ -160,7 +160,7 @@ func TestWriter(t *testing.T) {
 			w := NewWriter(logger, m)
 
 			for idx, req := range testData.reqs {
-				err = w.Write(context.Background(), DefaultTenant, req)
+				err = w.Write(context.Background(), 0, DefaultTenant, req)
```
fpetkovski, line 163: Would be nice to have a test with a non-zero limit.

pkg/receive/handler.go
```diff
@@ -258,7 +262,7 @@ type replica struct {
 	replicated bool
 }
 
-func (h *Handler) handleRequest(ctx context.Context, rep uint64, tenant string, wreq *prompb.WriteRequest) error {
+func (h *Handler) handleRequest(ctx context.Context, rep, seriesLimit uint64, tenant string, wreq *prompb.WriteRequest) error {
```
fpetkovski, line 265: The `seriesLimit` field seems to be added to the protobuf, so can we add it once to the request to avoid propagating it to every function?

fpetkovski (commented):
This would be an amazing feature for any multi-tenant setup. I wonder what is the status of the PR and if any help is needed to push it forward :)

PR #5337, opened 2022-05-06
sidecar/compact/store/receiver - Add the prefix option to buckets

jademcosta:

## Changes

This PR is another try to do the proposed on #1318
Some decisions had to be made in order for making it work, and those will be highlighted with comments in the code.
Although we used PR https://github.com/thanos-io/thanos/pull/3289 to understand where could we make this implementation, we rewrote everything from scratch.

Many thanks for the author of PR 3289 :)

## Verification

We've added unit tests, and added the prefix test to all e2e tests.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -10,6 +10,8 @@ We use *breaking :warning:* to mark changes that are not backward compatible (re
 
 ## Unreleased
 
+- [#5337](https://github.com/thanos-io/thanos/pull/5337) Thanos Object Store: Add the `prefix` option to buckets
+
 ### Fixed
 - [#5339](https://github.com/thanos-io/thanos/pull/5339) Receive: Fix deadlock on interrupt in routerOnly mode
 - [#5357](https://github.com/thanos-io/thanos/pull/5357) Store: fix groupcache handling of slashes
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -10,14 +10,13 @@ We use *breaking :warning:* to mark changes that are not backward compatible (re
 
 ## Unreleased
 
-- [#5337](https://github.com/thanos-io/thanos/pull/5337) Thanos Object Store: Add the `prefix` option to buckets
-
 ### Fixed
 - [#5339](https://github.com/thanos-io/thanos/pull/5339) Receive: Fix deadlock on interrupt in routerOnly mode
 - [#5357](https://github.com/thanos-io/thanos/pull/5357) Store: fix groupcache handling of slashes
 
 ### Added
 
+- [#5337](https://github.com/thanos-io/thanos/pull/5337) Thanos Object Store: Add the `prefix` option to buckets
 - [#5352](https://github.com/thanos-io/thanos/pull/5352) Cache: Add cache metrics to groupcache.
 - [#5391](https://github.com/thanos-io/thanos/pull/5391) Receive: Add relabeling support.
 
--- a/docs/components/tools.md
+++ b/docs/components/tools.md
@@ -101,6 +101,7 @@ type: GCS
 config:
   bucket: ""
   service_account: ""
+prefix: ""
 ```
 
 Bucket can be extended to add more subcommands that will be helpful when working with object storage buckets by adding a new command within [`/cmd/thanos/tools_bucket.go`](../../cmd/thanos/tools_bucket.go)  .
@@ -601,6 +602,7 @@ type: GCS
 config:
   bucket: ""
   service_account: ""
+prefix: ""
 ```
 
 ```$ mdox-exec="thanos tools bucket downsample --help"
@@ -675,6 +677,7 @@ type: GCS
 config:
   bucket: ""
   service_account: ""
+prefix: ""
 ```
 
 ```$ mdox-exec="thanos tools bucket mark --help"
--- a/docs/storage.md
+++ b/docs/storage.md
@@ -64,7 +64,6 @@ type: S3
 config:
   bucket: ""
   endpoint: ""
-  prefix: ""
   region: ""
   aws_sdk_auth: false
   access_key: ""
@@ -333,7 +332,6 @@ config:
   storage_account_key: ""
   container: ""
   endpoint: ""
-  prefix: ""
   max_retries: 0
   msi_resource: ""
   user_assigned_id: ""
@@ -388,7 +386,6 @@ config:
   password: ""
   domain_id: ""
   domain_name: ""
-  prefix: ""
   project_id: ""
   project_name: ""
   project_domain_id: ""
@@ -416,7 +413,6 @@ config:
   region: ""
   app_id: ""
   endpoint: ""
-  prefix: ""
   secret_key: ""
   secret_id: ""
   http_config:
@@ -448,7 +444,6 @@ config:
   bucket: ""
   access_key_id: ""
   access_key_secret: ""
-  prefix: ""
 ```
 
 Use --objstore.config-file to reference to this configuration file.
@@ -464,7 +459,6 @@ config:
   endpoint: ""
   access_key: ""
   secret_key: ""
-  prefix: ""
 ```
 
 #### Filesystem
@@ -477,7 +471,6 @@ NOTE: This storage type is experimental and might be inefficient. It is NOT advi
 type: FILESYSTEM
 config:
   directory: ""
-  prefix: ""
 ```
 
 ### How to add a new client to Thanos?
--- a/docs/storage.md
+++ b/docs/storage.md
@@ -63,6 +63,7 @@ NOTE: Minio client was mainly for AWS S3, but it can be configured against other
 type: S3
 config:
   bucket: ""
+  prefix: ""
   endpoint: ""
   region: ""
   aws_sdk_auth: false
--- a/docs/storage.md
+++ b/docs/storage.md
@@ -102,7 +102,7 @@ At a minimum, you will need to provide a value for the `bucket`, `endpoint`, `ac
 
 However if you set `aws_sdk_auth: true` Thanos will use the default authentication methods of the AWS SDK for go based on [known environment variables](https://docs.aws.amazon.com/cli/latest/userguide/cli-configure-envvars.html) (`AWS_PROFILE`, `AWS_WEB_IDENTITY_TOKEN_FILE` ... etc) and known AWS config files (~/.aws/config). If you turn this on, then the `bucket` and `endpoint` are the required config keys.
 
-The field `prefix` can be used to transparently use prefixes in your S3 bucket. That way, you may point distinct Thanos instances to the same bucket, while avoiding one instance messing with the data from another instance. This allows multiple Thanos deployments to use the same bucket without having to list all the files in it when you point Thanos store to the bucket (even when you are using tenants feature).
+The field `prefix` can be used to transparently use prefixes in your S3 bucket. This allows you to separate blocks coming from different sources into paths with different prefixes, making it easier to understand what's going on (i.e. you don't have to use Thanos tooling to know from where which blocks came).
 
 The AWS region to endpoint mapping can be found in this [link](https://docs.aws.amazon.com/general/latest/gr/s3.html).
 
--- a/docs/storage.md
+++ b/docs/storage.md
@@ -96,6 +96,7 @@ config:
     kms_encryption_context: {}
     encryption_key: ""
   sts_endpoint: ""
+prefix: ""
 ```
 
 At a minimum, you will need to provide a value for the `bucket`, `endpoint`, `access_key`, and `secret_key` keys. The rest of the keys are optional.
@@ -257,6 +258,7 @@ type: GCS
 config:
   bucket: ""
   service_account: ""
+prefix: ""
 ```
 
 ##### Using GOOGLE_APPLICATION_CREDENTIALS
@@ -358,6 +360,7 @@ config:
       key_file: ""
       server_name: ""
       insecure_skip_verify: false
+prefix: ""
 ```
 
 If `msi_resource` is used, authentication is done via system-assigned managed identity. The value for Azure should be `https://<storage-account-name>.blob.core.windows.net`.
@@ -398,6 +401,7 @@ config:
   connect_timeout: 10s
   timeout: 5m
   use_dynamic_large_objects: false
+prefix: ""
 ```
 
 #### Tencent COS
@@ -423,6 +427,7 @@ config:
     max_idle_conns: 100
     max_idle_conns_per_host: 100
     max_conns_per_host: 0
+prefix: ""
 ```
 
 The `secret_key` and `secret_id` field is required. The `http_config` field is optional for optimize HTTP transport settings. There are two ways to configure the required bucket information:
@@ -444,6 +449,7 @@ config:
   bucket: ""
   access_key_id: ""
   access_key_secret: ""
+prefix: ""
 ```
 
 Use --objstore.config-file to reference to this configuration file.
@@ -459,6 +465,7 @@ config:
   endpoint: ""
   access_key: ""
   secret_key: ""
+prefix: ""
 ```
 
 #### Filesystem
@@ -471,6 +478,7 @@ NOTE: This storage type is experimental and might be inefficient. It is NOT advi
 type: FILESYSTEM
 config:
   directory: ""
+prefix: ""
 ```
 
 ### How to add a new client to Thanos?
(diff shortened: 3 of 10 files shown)


pkg/objstore/prefixed_bucket.go
```diff
@@ -0,0 +1,88 @@
...
+	"strings"
+)
+
+type PrefixedBucket struct {
+	bkt    Bucket
+	prefix string
+}
+
+func NewPrefixedBucket(bkt Bucket, prefix string) Bucket {
+	pbkt := &PrefixedBucket{bkt: bkt, prefix: strings.Trim(prefix, DirDelim)}
```
jademcosta, line 15: The usage of `DirDelim` here is something that tricked me.
This is because we're stating that directories are going to be separated by `/`. I believe this would make filesystem bucket type to not work on Windows OS. But then I saw that metrics bucket [uses slashes](https://github.com/jademcosta/thanos/blob/787a97aafa2fe1dca7ae1e18dc75588991cc2f9d/pkg/objstore/objstore.go#L273), and as nobody complained, I believe Windows users doesn't run Thanos very often.
What is your opinion on that?

pkg/objstore/client/factory.go
```diff
@@ -81,5 +85,19 @@ func NewBucket(logger log.Logger, confContentYaml []byte, reg prometheus.Registe
 	if err != nil {
 		return nil, errors.Wrap(err, fmt.Sprintf("create %s client", bucketConf.Type))
 	}
-	return objstore.NewTracingBucket(objstore.BucketWithMetrics(bucket.Name(), bucket, reg)), nil
+
+	var prefixedBucket objstore.Bucket
+	if validPrefix(bucketConf.Config.Prefix) {
```
jademcosta, line 90: Although this seems a micro perf improvement, I guess it has the advantage of reducing risks of bugs for those who don't want to use prefixes (while also increasing perf for those).

jademcosta:
Aaaaand I just notice we broke some e2e tests due to our new config field. We will work on that :) Feedback is welcome anyway!

pkg/objstore/client/factory.go
yeya24: If prefix is configured but it is invalid, should we catch that error case and print the error log?

pkg/objstore/client/factory.go
```diff
@@ -81,5 +85,19 @@ func NewBucket(logger log.Logger, confContentYaml []byte, reg prometheus.Registe
...
+		prefixedBucket = objstore.NewPrefixedBucket(bucket, bucketConf.Config.Prefix)
+		level.Debug(logger).Log("msg", "using prefix on bucket access", "prefix", bucketConf.Config.Prefix)
+	} else {
+		prefixedBucket = bucket
```
yeya24, line 94: I think this else block is unnecessary. What about just using the code below on line 91.
``` go
bucket = objstore.NewPrefixedBucket(bucket, bucketConf.Config.Prefix)
```

yeya24 (commented):
Thanks for the contribution. I want to see prefix support in the objstore library so it is nice to see someone working on it again.
Can you add some E2E tests for this?
Btw I think we have a separate repo for the objstore library now https://github.com/thanos-io/objstore. But Thanos code is still using the old lib. I am wondering if we need to make changes on that repo.

pkg/objstore/client/factory.go
jademcosta: The thing is that we wanted to not use the prefixedBucket decorator in cases it is not needed. That would reduce the blast radius in case we introduce a bug. What do you think?

pkg/objstore/client/factory.go
jademcosta: I liked your idea :)
But I don't have an idea of how I would differentiate between a not filled key and a non-existent key. would love to have more ideas on this one.

pkg/objstore/prefixed_bucket_test.go
```diff
@@ -0,0 +1,102 @@
...
+	pBkt := NewPrefixedBucket(bkt, prefix)
+	rc1, err := pBkt.Get(context.Background(), "file1.jpg")
+
+	testutil.Ok(t, err)
+	defer func() { testutil.Ok(t, rc1.Close()) }()
+	content, err := ioutil.ReadAll(rc1)
+	testutil.Ok(t, err)
+	testutil.Equals(t, "test-data1", string(content))
+
+	pBkt.Upload(context.Background(), "file2.jpg", strings.NewReader("test-data2"))
```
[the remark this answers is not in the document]
jademcosta: @sonatype-lift ignore

pkg/objstore/prefixed_bucket_test.go
```diff
@@ -0,0 +1,102 @@
...
+
+	ok, err := pBkt.Exists(context.Background(), "file1.jpg")
+	testutil.Ok(t, err)
+	testutil.Assert(t, ok, "expected exits")
+
+	attrs, err := pBkt.Attributes(context.Background(), "file1.jpg")
+	testutil.Ok(t, err)
+	testutil.Assert(t, attrs.Size == 10, "expected size to be equal to 10")
+
+	bkt.Upload(context.Background(), strings.Trim(prefix, "/")+"/dir/file1.jpg", strings.NewReader("test-data1"))
```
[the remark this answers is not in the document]
jademcosta: @sonatype-lift ignore

pkg/objstore/prefixed_bucket_test.go
```diff
@@ -0,0 +1,102 @@
...
+	UsesPrefixTest(t, bkt, prefix)
+
+	prefix = "someprefix"
+	bkt = NewPrefixedBucket(NewInMemBucket(), prefix)
+	AcceptanceTest(t, bkt)
+	UsesPrefixTest(t, bkt, prefix)
+}
+
+func UsesPrefixTest(t *testing.T, bkt Bucket, prefix string) {
+	bkt.Upload(context.Background(), strings.Trim(prefix, "/")+"/file1.jpg", strings.NewReader("test-data1"))
```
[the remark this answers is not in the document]
jademcosta: @sonatype-lift ignore

pkg/objstore/prefixed_bucket_test.go
```diff
@@ -0,0 +1,102 @@
...
+	pBkt.Upload(context.Background(), "file2.jpg", strings.NewReader("test-data2"))
+	rc2, err := bkt.Get(context.Background(), strings.Trim(prefix, "/")+"/file2.jpg")
+
+	testutil.Ok(t, err)
+	defer func() { testutil.Ok(t, rc2.Close()) }()
+	contentUpload, err := ioutil.ReadAll(rc2)
+	testutil.Ok(t, err)
+	testutil.Equals(t, "test-data2", string(contentUpload))
+
+	pBkt.Delete(context.Background(), "file2.jpg")
```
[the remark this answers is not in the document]
jademcosta: @sonatype-lift ignore

jademcosta:
We've added a bunch of e2e tests. On all current tested flows to be more precise. Feedback is more than welcome :D

pkg/objstore/prefixed_bucket.go
GiedriusS: Yeah, I think your observation is correct however I think we could still fix this. Could you open up an issue for this? I think we could mark it as a "good first issue" because it should be relatively straightforward? :thinking:

GiedriusS (commented):
Thanks for your awesome work! :beers:  Haven't looked through the whole code yet but added some comments from my side.

docs/storage.md
```diff
@@ -102,6 +102,8 @@ At a minimum, you will need to provide a value for the `bucket`, `endpoint`, `ac
 
 However if you set `aws_sdk_auth: true` Thanos will use the default authentication methods of the AWS SDK for go based on [known environment variables](https://docs.aws.amazon.com/cli/latest/userguide/cli-configure-envvars.html) (`AWS_PROFILE`, `AWS_WEB_IDENTITY_TOKEN_FILE` ... etc) and known AWS config files (~/.aws/config). If you turn this on, then the `bucket` and `endpoint` are the required config keys.
 
+The field `prefix` can be used to transparently use prefixes in your S3 bucket. That way, you may point distinct Thanos instances to the same bucket, while avoiding one instance messing with the data from another instance. This allows multiple Thanos deployments to use the same bucket without having to list all the files in it when you point Thanos store to the bucket (even when you are using tenants feature).
```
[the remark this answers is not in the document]
jademcosta: Hi!
I don't know if I understood all the advantages you describe, but the part I did, makes sense.
I replaced the second part with the explanation you gave here. But, please, feel free to ask for changes in it, as you certainly answer more questions related to Thanos than me, so you probably know more how people use it :)

pkg/objstore/prefixed_bucket.go
jademcosta: Done: https://github.com/thanos-io/thanos/issues/5380

pkg/objstore/client/factory.go
dudaduarte: nice idea @GiedriusS, we moved it to NewPrefixedBucket

pkg/objstore/prefixed_bucket.go
```diff
@@ -0,0 +1,99 @@
...
+}
+
+func validPrefix(prefix string) bool {
+	prefix = strings.Replace(prefix, "/", "", -1)
+	return len(prefix) > 0
+}
+
+func conditionalPrefix(prefix, name string) string {
+	if len(name) > 0 && len(prefix) > 0 {
+		return prefix + DirDelim + name
```
GiedriusS, line 32: nit: let's call `withPrefix` here? actually I think we only need to check for `len(name) > 0` here because `len(prefix) > 0` is already given our initial checks?

pkg/objstore/prefixed_bucket.go
```diff
@@ -0,0 +1,99 @@
...
+
+func (p *PrefixedBucket) Close() error {
+	return p.bkt.Close()
+}
+
+// Iter calls f for each entry in the given directory (not recursive.). The argument to f is the full
+// object name including the prefix of the inspected directory.
+// Entries are passed to function in sorted order.
+func (p *PrefixedBucket) Iter(ctx context.Context, dir string, f func(string) error, options ...IterOption) error {
+	if len(p.prefix) > 0 {
```
GiedriusS, line 49: Nit: we now check whether the prefix is valid at the beginning (includes a check that it is not empty) so this is not needed

pkg/objstore/prefixed_bucket_test.go
```diff
@@ -0,0 +1,101 @@
...
+	"io/ioutil"
+	"sort"
+	"strings"
+	"testing"
+
+	"github.com/thanos-io/thanos/pkg/testutil"
+)
+
+func TestPrefixedBucket_Acceptance(t *testing.T) {
+	prefix := "/someprefix/anotherprefix/"
```
GiedriusS, line 17: Since I believe that you will clean up the code a bit either way, could we put these strings into a string slice and iterate over the elements, and call those functions? I think it would make this way neater.

pkg/objstore/prefixed_bucket_test.go
```diff
@@ -0,0 +1,101 @@
...
+
+	testutil.Ok(t, err)
+	defer func() { testutil.Ok(t, rc1.Close()) }()
+	content, err := ioutil.ReadAll(rc1)
+	testutil.Ok(t, err)
+	testutil.Equals(t, "test-data1", string(content))
+
+	testutil.Ok(t, pBkt.Upload(context.Background(), "file2.jpg", strings.NewReader("test-data2")))
+	rc2, err := bkt.Get(context.Background(), strings.Trim(prefix, "/")+"/file2.jpg")
+	testutil.Ok(t, err)
```
GiedriusS, line 53: Double `err` test, we can remove one of these

pkg/objstore/prefixed_bucket_test.go
```diff
@@ -0,0 +1,101 @@
...
+	rc3, err := pBkt.GetRange(context.Background(), "file1.jpg", 1, 3)
+	testutil.Ok(t, err)
+	defer func() { testutil.Ok(t, rc3.Close()) }()
+	content, err = ioutil.ReadAll(rc3)
+	testutil.Ok(t, err)
+	testutil.Equals(t, "est", string(content))
+
+	ok, err := pBkt.Exists(context.Background(), "file1.jpg")
+	testutil.Ok(t, err)
+	testutil.Assert(t, ok, "expected exits")
```
GiedriusS, line 75:
```suggestion
	testutil.Assert(t, ok, "expected to exist")
```

pkg/objstore/prefixed_bucket.go
jademcosta: Makes sense. Done!

pkg/objstore/prefixed_bucket.go
jademcosta: Done!

jademcosta:
One more thing I don't know how to do: how can I update the docs with the `prefix` field?

GiedriusS:
> One more thing I don't know how to do: how can I update the docs with the `prefix` field? When I add it manually it breaks the tests.

Yeah, we generate those docs automatically from the structs. Since this option is "hidden" now in the types, the check cannot pass. What about adding a new field to the [BucketConfig type](https://github.com/thanos-io/thanos/blob/7ba274c69cb0a848c7e8e1c593221faa80d8b6e6/pkg/objstore/client/factory.go#L41)? Then we wouldn't have to have hacks like [this](https://github.com/thanos-io/thanos/pull/5337/files#diff-aba15f20dff65b99f7320cb411ad588f2ee051714d3bda4d986b5a05d85bb105R91).

pkg/objstore/client/factory.go
```diff
@@ -41,6 +41,7 @@ const (
 type BucketConfig struct {
 	Type   ObjProvider `yaml:"type"`
 	Config interface{} `yaml:"config"`
+	Prefix string      `yaml:"prefix" default:""`
```
dudaduarte, line 44: > Yeah, we generate those docs automatically from the structs. Since this option is "hidden" now in the types, the check cannot pass. What about adding a new field to the [BucketConfig type](https://github.com/thanos-io/thanos/blob/7ba274c69cb0a848c7e8e1c593221faa80d8b6e6/pkg/objstore/client/factory.go#L41)? Then we wouldn't have to have hacks like [this](https://github.com/thanos-io/thanos/pull/5337/files#diff-aba15f20dff65b99f7320cb411ad588f2ee051714d3bda4d986b5a05d85bb105R91).

Hey @GiedriusS , I followed your suggestion here :) The only thing I don't love about bringing the prefix to the BucketConfig is that all the bucket options are inside Config 🤔
But with this change, we also keep the prefix layer with its validations and don't  take the risk to break something by changing all the Config types that BucketConf handles, so I think it's the best alternative

GiedriusS: :+1: yeah, I think this is the best solution out of the alternatives

docs/components/tools.md
```diff
@@ -811,7 +814,7 @@ Flags:
       --rewrite.to-relabel-config-file=<file-path>
                                 Path to YAML file that contains relabel configs
                                 that will be applied to blocks
-      --tmp.dir="/tmp/thanos-rewrite"
+      --tmp.dir="/var/folders/pr/h0912y1s3zj507j30fzj87gh0000gn/T/thanos-rewrite"
```
GiedriusS, line 817: What's this random directory? :thinking:

dudaduarte: I'm not sure actually, this change was generated when I ran `make check-docs`, but the docs check kept breaking. I removed this change and the docs check passed, I didn't understand why since it also rans `mack check-docs` 🤔

GiedriusS (approved):
:tada: let's see if there any other comments/suggestions

CHANGELOG.md
```diff
@@ -10,6 +10,8 @@ We use *breaking :warning:* to mark changes that are not backward compatible (re
 
 ## Unreleased
 
+- [#5337](https://github.com/thanos-io/thanos/pull/5337) Thanos Object Store: Add the `prefix` option to buckets
```
yeya24, line 13: Can we move this line under `Added`?

docs/components/tools.md
yeya24: This thing happens to me. I feel it is an issue specific to MacOS. The temp dir is generated incorrectly.

dudaduarte:
> Thanks for your awesome work! 🍻 Haven't looked through the whole code yet but added some comments from my side.

Thanks a lot @GiedriusS !
I just lost your approve by commiting a fix in the changelog 😞  I'm sorry
I waited a few days for new comments or suggestions, I guess we can merge it now :D

CHANGELOG.md
dudaduarte: Done! Thanks @yeya24

yeya24 (approved):
LGTM. Thanks for the contribution!

PR #5342, opened 2022-05-09
Reference implementation: vertical query sharding

fpetkovski:
This PR is a reference implementation for vertical query sharding.


pkg/store/prometheus.go
```diff
@@ -195,11 +196,13 @@ func (p *PrometheusStore) Series(r *storepb.SeriesRequest, s storepb.Store_Serie
...
 	}
 
-	if r.QueryHints != nil && r.QueryHints.IsSafeToExecute() {
+	shardMatcher := r.ShardInfo.Matcher()
+	if r.QueryHints != nil && r.QueryHints.IsSafeToExecute() && !shardMatcher.IsSharded() {
 		return p.queryPrometheus(s, r)
 	}
 
 	q := &prompb.Query{StartTimestampMs: r.MinTime, EndTimestampMs: r.MaxTime}
+	level.Info(p.logger).Log("min_time", r.MinTime, "max_time", r.MaxTime)
```
yeya24, line 205: This log will be printed a lot of times. Can we add more fields and make it debug level?

pkg/queryfrontend/roundtrip.go
```diff
@@ -50,7 +52,11 @@ func NewTripperware(config Config, reg prometheus.Registerer, logger log.Logger)
 	queryRangeCodec := NewThanosQueryRangeCodec(config.QueryRangeConfig.PartialResponseStrategy)
 	labelsCodec := NewThanosLabelsCodec(config.LabelsConfig.PartialResponseStrategy, config.DefaultTimeRange)
 
-	queryRangeTripperware, err := newQueryRangeTripperware(config.QueryRangeConfig, queryRangeLimits, queryRangeCodec,
+	queryRangeTripperware, err := newQueryRangeTripperware(
```
yeya24, line 55: Is query sharding working with Instant query? I see no reason why it doesn't...

yeya24 (commented):
Thanks for the great work! Overall this is amazing and I can see unit tests and E2E tests are passing.
I am trying to understand this feature a bit more:
1. Is there already a design proposal for this feature? It is a little bit big so maybe need some discussions. If not, maybe we can add some information to the PR description section first.
2. The sharding query seems triggered by the query frontend, are range query and instant query both supported? Our use case focuses on more instant query performance.
3. It would be good to have a test case in the E2E promql compliance test to make sure it is still compatible.

pkg/querysharding/analysis.go
```diff
@@ -0,0 +1,137 @@
...
+func newShardableByLabels(labels []string, by bool) QueryAnalysis {
+	labels = without(labels, excludedLabels)
+
+	return QueryAnalysis{
+		shardBy:        by,
+		shardingLabels: labels,
+	}
+}
+
+func (q QueryAnalysis) scopeToLabels(labels []string, by bool) QueryAnalysis {
```
yeya24, line 29: Can we use pointer receiver instead? This way we can avoid copying the object when returning the value.

pkg/queryfrontend/shard_query.go
```diff
@@ -0,0 +1,85 @@
...
+		return nil, err
+	}
+
+	if !analysis.IsShardable() {
+		return s.next.Do(ctx, r)
+	}
+
+	reqs := s.shardQuery(r, analysis)
+
+	reqResps, err := queryrange.DoRequests(ctx, s.next, reqs, s.limits)
```
yeya24, line 51: Why the sharding feature is implemented on the query frontend? It is also possible to do the same thing on the querier side directly. WDYT?

pkg/querysharding/analyzer.go
```diff
@@ -0,0 +1,106 @@
...
+				isShardable = false
+				return fmt.Errorf("expressions with %s are not shardable", n.Func.Name)
+			}
+		case *parser.BinaryExpr:
+			analysis = analysis.scopeToLabels(n.VectorMatching.MatchingLabels, n.VectorMatching.On)
+		case *parser.AggregateExpr:
+			labels := make([]string, 0)
+			if len(n.Grouping) > 0 {
+				labels = n.Grouping
+			}
```
yeya24, line 59: I think we can simplified line 56 ~ 59 to just `labels := n.Grouping`, or simply pass `n.Grouping` to scopeToLabels

pkg/querysharding/analyzer.go
```diff
@@ -0,0 +1,106 @@
...
+
+// NewQueryAnalyzer creates a new QueryAnalyzer
+func NewQueryAnalyzer() *QueryAnalyzer {
+	return &QueryAnalyzer{}
+}
+
+// Analyze analyzes a query and returns a QueryAnalysis.
+
+// Analyze uses the following algorithm:
+// * if a query has subqueries, such as label_join or label_replace,
```
yeya24, line 26: I guess you want to mention 2 cases here: if a query has subqueries or a query contains non shardable functions then it is not shardable

pkg/store/prometheus.go
```diff
@@ -366,27 +396,33 @@ func (p *PrometheusStore) handleStreamedPrometheusResponse(s storepb.Store_Serie
...
+		timePrometheusRead += time.Now().Sub(startPromRead).Seconds()
 
 		if len(res.ChunkedSeries) != 1 {
 			level.Warn(p.logger).Log("msg", "Prometheus ReadRequest_STREAMED_XOR_CHUNKS returned non 1 series in frame", "series", len(res.ChunkedSeries))
 		}
 
 		framesNum++
 		for _, series := range res.ChunkedSeries {
+			startSharding := time.Now()
+			if !shardMatcher.MatchesZLabels(series.Labels) {
```
yeya24, line 421: Just curious that is the current implementation has better performance than non-sharding queries?
I can see the sidecar will send more remote read requests to Prometheus. Prometheus will respond the same series results back to the sidecar and the filtering is done on the sidecar side. 

This causes more load on the Prometheus side. I am wondering if we can filter on the Prometheus side. Maybe combined with https://github.com/prometheus/prometheus/issues/10420.

fpetkovski:
Thanks for the extensive review @yeya24 ! 

You're right, we got a bit ahead of ourselves. We wanted to submit the proposal yesterday and link this PR, but it just needed a bit more work. However, it's now ready for review: https://github.com/thanos-io/thanos/pull/5350

Feel free to go over that and leave any comments you might have, and I'll try to address some of your concerns here to make it easier for other reviewers.

yeya24:
What's the plan for this pr? When is it ready for review?

fpetkovski:
Hello, I posted in the thanos-dev channel yesterday on this topic. I intend to start cleaning up the PR in the next one or two weeks. Would you prefer it to be split into two smaller ones, or is one large PR okay?

cmd/thanos/query_frontend.go
```diff
@@ -136,6 +136,8 @@ func registerQueryFrontend(app *extkingpin.App) {
 
 	cmd.Flag("query-frontend.forward-header", "List of headers forwarded by the query-frontend to downstream queriers, default is empty").PlaceHolder("<http-header-name>").StringsVar(&cfg.ForwardHeaders)
 
+	cmd.Flag("query-frontend.num-shards", "Number of queriers to use when sharding PromQL queries").IntVar(&cfg.NumShards)
```
fpetkovski, line 139: We might use a more descriptive flag here

pkg/queryfrontend/queryrange_codec.go
```diff
@@ -262,3 +294,16 @@ func matchersToStringSlice(storeMatchers [][]*labels.Matcher) []string {
 	}
 	return res
 }
+
+func encodeShardInfo(info *storepb.ShardInfo) (string, error) {
```
fpetkovski, line 298: This might not be ideal because it will be hard to set the parameter manually in the URL. We can start like this and optimize later, or we can come up with something more user friendly.

pkg/queryfrontend/roundtrip.go
fpetkovski: Yes, this will be possible, but I think it's better to do it as a follow up.

pkg/queryfrontend/shard_query.go
fpetkovski: The reason for this is because we need to split the query and unify the result, and this is already possible with query frontend. If we added it to the querier, then that querier has to talk to other queriers which is additional complexity that we need to introduce.

pkg/store/prometheus.go
fpetkovski: Yeah this was only for debugging purposes :) Should be gone now.

pkg/store/prometheus.go
fpetkovski: This is a very good observation, it is hard to shard directly in Prometheus. I did some benchmarking and it seems that the improvements are still there even for Prometheus. I guess people will need to test themselves and we can call this out in the docs, but I believe there is value in sharding in the sidecar.

pkg/store/proxy.go
```diff
@@ -503,6 +531,10 @@ func startStreamSeriesSet(
 			}
 
 			if series := rr.r.GetSeries(); series != nil {
+				if applySharding && !shardMatcher.MatchesZLabels(series.Labels) {
```
fpetkovski, line 534: Is this placed at the correct line to not interfere with cancellations?

pkg/api/query/v1.go
```diff
@@ -295,6 +297,25 @@ func (qapi *QueryAPI) parseStep(r *http.Request, defaultRangeQueryStep time.Dura
...
+	if data == "" {
+		return nil, nil
+	}
+
+	if len(data) == 0 {
+		return nil, nil
+	}
+
+	var info storepb.ShardInfo
+	err := json.Unmarshal([]byte(data), &info)
```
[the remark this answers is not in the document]
fpetkovski: Yes, done 👍

pkg/querysharding/analysis.go
fpetkovski: Will change to pointer receiver.

fpetkovski:
@bwplotka @yeya24 @GiedriusS this should now be cleaned up and ready for review. It's a bigger PR so please don't hesitate to ask questions :)

pkg/store/proxy.go
```diff
@@ -470,6 +490,14 @@ func startStreamSeriesSet(
 				}
 			}
 		}()
+
+		shardMatcher := shardInfo.Matcher()
+		applySharding := shardInfo != nil && !storeSupportsSharding
+		if applySharding {
```
fpetkovski, line 496: This makes sure we select the correct subset of series in the querier in case the underlying store does not support series sharding.

yeya24: Let's keep it for safeguard as during the migration some old stores might still there.

test/e2e/query_frontend_test.go
```diff
@@ -9,6 +9,9 @@ import (
 	"testing"
 	"time"
 
+	"github.com/thanos-io/thanos/pkg/block/metadata"
+	"github.com/thanos-io/thanos/pkg/testutil/e2eutil"
```
yeya24, line 13: Let's merge the imports

cmd/thanos/query_frontend.go
yeya24: Is it possible to specify the number of shards to use for each request? `query-frontend.num-shards` this flag can be a default fallback value if no shard num is specified from the query.

pkg/store/proxy.go
GiedriusS: I think so, this should make it so that not matching series are skipped

pkg/store/storepb/shard_info.go
```diff
@@ -0,0 +1,104 @@
...
+func (m *ShardInfo) Matcher() *ShardMatcher {
+	if m == nil || m.TotalShards < 1 {
+		return &ShardMatcher{
+			isSharded: false,
+		}
+	}
+
+	return &ShardMatcher{
+		isSharded:        true,
+		buf:              make([]byte, 10*units.Kilobyte),
```
GiedriusS, line 82: I think we should introduce a private `sync.Pool` for this and make a function Close() that returns this buffer to the pool. I think everywhere we have code like:
```go
	shardMatcher := r.ShardInfo.Matcher()
	...
	do something with shardMatcher
```

So, as far as I can tell it should be trivial to use a sync.Pool here:

```go
	shardMatcher := r.ShardInfo.Matcher()
	defer shardMatcher.Close()
```

pkg/store/prometheus.go
GiedriusS: It's still here? :thinking:

pkg/store/storepb/shard_info.go
fpetkovski: Sounds good, will do that.

test/e2e/query_frontend_test.go
fpetkovski: Should be together now

pkg/querysharding/analyzer.go
fpetkovski: Yes good idea, I amended the comment.

pkg/store/prometheus.go
fpetkovski: Should be gone for good :)

cmd/thanos/query_frontend.go
yeya24: Yeah, do we have any plan to dynamically specify the number of shards to use for each query? If not specified we can default to this flag.

fpetkovski: The number of shards will be predefined and static, we don't have the option of tweaking the sharding number dynamically.

fpetkovski:
@yeya24 yes, I've tried this both on my laptop and in an actual environment. The PR should be ready to run, but I intend to test once more to make sure code review changes don't introduce a bug.

pkg/store/storepb/shard_info.go
fpetkovski: I made this change and pushed a new commit. Could you take a look if this is what you had in mind?

yeya24 (approved):
This is very solid work. LGTM

cmd/thanos/query_frontend.go
fpetkovski: I modified the description a bit to make things a bit more clear

fpetkovski:
I pushed a new commit to also include external labels during sharding.

GiedriusS (approved):
Looks solid! :muscle: let's wait a few days and if there are no other comments/suggestions then let's merge this

pkg/queryfrontend/shard_query.go
bwplotka: 👍🏽

bwplotka: Can be added later.

pkg/querysharding/analysis.go
```diff
@@ -0,0 +1,140 @@
+// Copyright (c) The Thanos Authors.
+// Licensed under the Apache License 2.0.
+
+package querysharding
+
+var excludedLabels = []string{"le"}
+
+type QueryAnalysis struct {
+	// Labels to shard on
```
bwplotka, line 9:
```suggestion
	// shardingLabels represent labels to shard on.
```

pkg/querysharding/analysis.go
```diff
@@ -0,0 +1,140 @@
...
+	shardingLabels []string
+
+	// When set to true, sharding is `by` shardingLabels,
+	// otherwise it is `without` shardingLabels.
```
bwplotka, line 13:
```suggestion
	// shardBy is specifiying if we shard by the `shardingLabels` (if true),
	// or without `shardingLabels`.
```

pkg/store/storepb/rpc.proto
```diff
@@ -118,8 +118,13 @@ message SeriesRequest {
   // query_hints are the hints coming from the PromQL engine when
   // requesting a storage.SeriesSet for a given expression.
   QueryHints query_hints = 12;
+
+  // shard_info is used by the querier to request a specific
+  // shard of blocks instead of entire blocks.
```
bwplotka, line 123: That is not entirely true - we shard by series.

bwplotka (approved):
Good work 💪🏽 So much good stuff coming, just comment nits that we can fix in different PRs. 👍🏽

PR #5347, opened 2022-05-09
Adds more info about the formatting part.

### LINKED ISSUE #5282: Make it easier to pass documentation checks, especially for new users
**What happened**:
A lot of PR's failing on something lame such as a whitespace. We also have multiple ways to format docs in our Makefile.

That if ANY user makes a change, it should be VERY easy to contribute without having to figure special things out, install anything or get stuck in a PR

Make a doc change without applying any formatting and make a MR :)

I will be picking this up.

wiardvanrij:
Signed-off-by: Wiard van Rij <wvanrij@roku.com>

Closes #5282

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -156,6 +156,15 @@ $ git push origin <your_branch_for_new_pr>
 
 **Tests your changes**
 
+**Formatting**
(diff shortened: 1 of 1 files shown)


bwplotka (approved):
<3

PR #5350, opened 2022-05-10
Submit a proposal for vertical query sharding

fpetkovski:
This commit submits a proposal for vertical query sharding in Thanos.
Looking for feedback and comments!

## Changes
N/A

## Verification
N/A


docs/proposals-accepted/202205-vertical-query-sharding.md
```diff
@@ -0,0 +1,152 @@
...
+
+### Drawbacks & Exceptions
+
+*Not all queries are easily shardeable*
+There are certain aggregations for which sharding cannot be performed safely. For these cases the sharder can simply fall back to completely executing the expression as a single query. These cases include the use of functions such as `label_replace` and `label_join` which create new labels inside PromQL, when the query is being executed. Since such labels can be arbitrarily created while executing queries, Stores will be unaware of them and cannot take them into account when sharding the matching series set. 
+
+#TODO Mention promql parsing/transformation to increase queries that can be sharded
+
+*Impact of sharded queries on block retrieval* 
+Given that queries are sharded based on series, and not time, for a given leaf most metrics for each independent shard will likely reside in the same block(s). This can multiply the cost of retrieving series from a given block since the block has to be traversed N number of times, where N is the number of distinct shards.
```
GiedriusS, line 105: `groupcache` should alleviate these problems ;P

docs/proposals-accepted/202205-vertical-query-sharding.md
```diff
@@ -0,0 +1,152 @@
...
+The following diagram illustrates how the sharding algorithm would work end to end:
+
+![Vertical sharding](../img/vertical-sharding.png "Vertical query sharding")
+
+### Drawbacks & Exceptions
+
+*Not all queries are easily shardeable*
+There are certain aggregations for which sharding cannot be performed safely. For these cases the sharder can simply fall back to completely executing the expression as a single query. These cases include the use of functions such as `label_replace` and `label_join` which create new labels inside PromQL, when the query is being executed. Since such labels can be arbitrarily created while executing queries, Stores will be unaware of them and cannot take them into account when sharding the matching series set. 
+
+#TODO Mention promql parsing/transformation to increase queries that can be sharded
```
GiedriusS, line 102: Something missing here?

docs/proposals-accepted/202205-vertical-query-sharding.md
```diff
@@ -0,0 +1,152 @@
...
+* Utilizing the existing query range caching implementation 
+* Vertical-query sharding a query that has already been time-sliced reduces the cardinality of each respective query shard
+The following diagram illustrates how the sharding algorithm would work end to end:
+
+![Vertical sharding](../img/vertical-sharding.png "Vertical query sharding")
+
+### Drawbacks & Exceptions
+
+*Not all queries are easily shardeable*
+There are certain aggregations for which sharding cannot be performed safely. For these cases the sharder can simply fall back to completely executing the expression as a single query. These cases include the use of functions such as `label_replace` and `label_join` which create new labels inside PromQL, when the query is being executed. Since such labels can be arbitrarily created while executing queries, Stores will be unaware of them and cannot take them into account when sharding the matching series set.
```
GiedriusS, line 100: Only these two cases? :thinking:

GiedriusS (commented):
:+1:

docs/proposals-accepted/202205-vertical-query-sharding.md
moadz: With block-sharded store-gw, will store-gw in the same groupcache ever need to talk to each other?

docs/proposals-accepted/202205-vertical-query-sharding.md
moadz: Many many more, these are more limitations with the currently proposed implementation for aggregation operations.

docs/proposals-accepted/202205-vertical-query-sharding.md
moadz:
```suggestion
*Support for more of the PromQL standard library*

Sharding an aggregation operation is relatively simple as a grouping/excluding set of label matchers are always specified. This allows us to shard the StoreAPI series calls by guaranteeing that all unique values for a given label end up on the same shard. Furthermore,  groupbys in PromQL are associative, and queries can be concatenated into a single result or re-evaluated without compromising correctness.

For the broader PromQL standard library, this is not always the case. For examples arithmetic operators on two subqueries or computing an average across an entire label set (this would actually require each shard to return a `count` and `sum` to correctly compute the average in a final evaluation in the root query). Whilst this proposal focusses on aggregation operations, an AST mapper has been implemented to determine if a PromQL expression contains non-shardeable attributes or operations. This will be extended to transform queries into shardeable subqueries in a future proposal. 
```

docs/proposals-accepted/202205-vertical-query-sharding.md
```diff
@@ -0,0 +1,152 @@
...
+
+The reason why partitioning the series set on the grouping labels works is because a grouping aggregation can be executed independently on each combination of values for the labels found in the grouping clause. In our example, as long as series with the same `pod` label end up in the same query shard, we can safely perform the shard and merge strategy. 
+
+### Initiating sharded queries
+
+The Query Frontend component already has useful splitting and merging capabilities which are currently used for sharding queries horizontally. In order to provide the best possible user experience and not burden users with running additional component(s), we propose adding a new middleware to Query Frontend which will implement the vertical sharding algorithm proposed here after the step-alignment and horizontal time-slicing steps. This would allow users to restrict the maximum complexity of a given query based on vertical sharding of already time-sliced queries. The only new user-specified parameter would be the number of shards in which to split PromQL aggregations.
```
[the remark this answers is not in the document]
fpetkovski: Yes, this is correct and we can do it. The only reason is that QFE has some logic for splitting and merging range queries, and we used that in the PoC. I'll update the doc to indicate that the same strategy can be applied to instant queries.

docs/proposals-accepted/202205-vertical-query-sharding.md
```diff
@@ -0,0 +1,152 @@
...
+
+#TODO Mention promql parsing/transformation to increase queries that can be sharded
+
+*Impact of sharded queries on block retrieval* 
+Given that queries are sharded based on series, and not time, for a given leaf most metrics for each independent shard will likely reside in the same block(s). This can multiply the cost of retrieving series from a given block since the block has to be traversed N number of times, where N is the number of distinct shards.
+
+In our sharded query benchmarks, we found that the increased volume of Series calls to Store Gateways and Receivers did not lead to a significant latency increase. However, we did notice a higher remote-read latency in Prometheus caused by multiple remote-read calls for the same Series call. This is due to the fact that the sharding has to be done in the sidecar, after retrieving all matching series for a query. A similar concern was raised in Prometheus already, with a proposal to support sharding series natively in the Prometheus TSDB (https://github.com/prometheus/prometheus/issues/10420).
```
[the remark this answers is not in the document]
fpetkovski: I think we'll have to take care of respecting `max-query-parallelism`, but other than that, we should be able to use both vertical and horizontal sharding independently of each other.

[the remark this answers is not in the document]
moadz: The proposed implementation does vertical sharding after time slicing: 
`step-align -> timeslice -> vertically shard query` 

Both techniques are complimentary as the maximum query cardinality becomes `active series per configured QFE time slice (e.g. 12 hours) / number of shards configured (e.g. 10 vertical shards)`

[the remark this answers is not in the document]
yeya24: As mentioned in https://cloud-native.slack.com/archives/CL25937SP/p1648741512490379, I think we can reuse the Mimir Prometheus fork.
I am wondering if we can try to integrate with that first. It should be more performant in the case of the sidecar and an additional label matcher is easy to add at the query time.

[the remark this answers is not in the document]
fpetkovski: I think we should clarify in the proposal that a high cardinality query would still execute faster overall, just that Prometheus will take a bit longer to retrieve requested series over remote-read. 

Could you elaborate a bit on how we would reuse the Mimir Prometheus fork? Will people have to run that instead of upstream Prometheus?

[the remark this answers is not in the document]
yeya24: That's a good point. We cannot force users to run that version of Prometheus. But I believe that the upstream TSDB will have that native sharding support later. I hope our solution can be at least compatible with it and later we can switch to it once it is upstreamed.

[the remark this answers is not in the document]
fpetkovski: I just saw that we can also request only series without chunks from remote read: https://github.com/prometheus/prometheus/blob/3947238ce0e3ff7e7c6e9502f73275cdbc233054/tsdb/querier.go#L140

Something I would also like to try out in the sidecar is:
* request only series from remote read
* calculate only the series which match a shard
* request series and chunks for a shard

I'll report findings once I've run some benchmarks.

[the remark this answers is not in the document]
yeya24: This one works but it sends more remote read requests to Prometheus (one request per series). I am curious what's the benchmark results...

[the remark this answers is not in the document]
fpetkovski: I was thinking we could fetch batches of series in fewer requests, but now that I'm looking at the code, I don't think that is easy to do. So this approach probably won't work as I expected.

docs/proposals-accepted/202205-vertical-query-sharding.md
```diff
@@ -0,0 +1,152 @@
...
+and 
+```
+sum by (pod) (memory_usage_bytes{pod="prometheus-2"})
+```
+
+We can therefore execute each one of the two queries in a separate query shard and concatenate the result before returning it to the user.
+
+### Dynamic series partitioning
+
+Since Queriers have no information about which timeseries are available in Stores, they cannot easily rewrite a single aggregation into two disjoint ones. They can, however, propagate information to Stores that instruct them to only return one disjoint shard of the series that match particular selectors. To achieve this, each query shard would propagate the total number of shards, its own shard index, and the grouping labels discovered in the PromQL expression. Stores would then perform a `hashmod` on the grouping labels against each series and return only those series whose `hashmod` equals the requested shard index.
```
yeya24, line 65: Mimir's split and merge compactor shards a large block to multiple small blocks. Each small-block contains all series for a specific shard.
If the store gateway is shard-aware or each small block contains a special label indicating the shard id, then we don't have to do `hashmod` during query times and performance can be improved since less blocks need to be queried.

I am not sure if we can reuse that split and merge compactor, but we can do something similar for compaction time sharding.

fpetkovski: This is indeed something we considered, but in the end we were concerned with the complexity of the approach. The way I understand this, we either have to implement some sort of sharding at write, which would likely not work for the Prometheus sidecar, or make users configure disjoint shards themselves, which could break backwards compatibility. 

It also turns out that calculating `hashmod` is quite cheap and can be done on the fly. The only overhead is the multiple requests that go to the Stores, but for receivers and store gateways this is not a problem because sharding is done before chunks are even retrieved. With the Prometheus sidecar it could be a concern because sharding is done once series with chunks are retrieved, but for the time being there isn't really a good way to solve this. 

If sharding is added to the TSDB, we can modify our approach to take advantage of that, but we thought it was more important to start with something that already delivers value.

fpetkovski: I just tried a high cardinality query against Prometheus with and without sharding. So with sharding, retrieving series took 1.7 seconds (hashing was just 0.01s) and the overall query took 3.5 seconds. Without sharding, retrieval took 1.3 seconds but the entire query took 6.5 seconds.

yeya24: Thanks. I think the current one is good as a start and we can iterate.
As the tsdb shard index has not been upstreamed yet, we can think about it later.

GiedriusS: My 2 cents: I think sharding could happen in many places simultaneously, e.g. just like we shard requested cache keys into batches and so on. The idea, in general, is the same - to perform more work at the same time i.e. use resources more efficiently. It could be trickier to implement fair resource sharing between tenants in the future with sharding in many places. Still, I think it's a non-problem currently since usually multi-tenancy is achieved by each tenant having their own read/write paths with something like shuffle sharding i.e. separate processes in general.

yeya24:
Very nice proposal! This is super good as a start and I cannot wait to try it.
Can you please sign off the dco and fix other CI related issues?

docs/proposals-accepted/202205-vertical-query-sharding.md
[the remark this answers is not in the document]
yeya24: Can you please update this accordingly?

docs/proposals-accepted/202205-vertical-query-sharding.md
```diff
@@ -0,0 +1,152 @@
...
+
+### Pitfalls of the current solution
+
+When executing a PromQL query, a Querier will pull series from all of its downstream Stores in memory and feed them to the Prometheus query engine. In cases where a query causes a large number of series to be fetched, executing the query can lead to high memory usage in Queriers. As a result, even when more than one Querier is available, work is concentrated in a single Querier and cannot be easily distributed.
+
+## Goals
+* **Must** allow the sharding of aggregation expressions across N queries, where N denotes a user-provided sharding factor  
+* **Must** fall back to single query0 when a non-shardable query or querier is encountered in the query path
```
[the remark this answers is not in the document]
fpetkovski: Yeah typo :) Should be fixed.

fpetkovski:
I pushed an update with linted docs, but the CI seems to be a bit flaky with URLs

docs/proposals-accepted/202205-vertical-query-sharding.md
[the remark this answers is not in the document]
fpetkovski: Updated. Let me know if you think if there's something else to be added.

docs/proposals-accepted/202205-vertical-query-sharding.md
yeya24: Can we also mention the compaction time sharding for store gateway in this proposal? We don't have to implement this for phase 1 but this is something nice to have.

fpetkovski: Should we mention this under alternatives?

yeya24: It is a good complement to the current proposal. I think it is not really an alternative and we can have both.

fpetkovski: I added it under a `Future improvements` section

yeya24 (approved):
LGTM

docs/proposals-accepted/202205-vertical-query-sharding.md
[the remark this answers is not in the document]
GiedriusS: :+1: it's clear now, thanks!

docs/proposals-accepted/202205-vertical-query-sharding.md
```diff
@@ -0,0 +1,155 @@
...
+
+*Extended horizontal sharding* Thanos Query Frontend already does horizontal (based on time-range) sharding by splitting up queries into smaller time-ranges. As an alternative to vertical sharding, more granular horizontal sharding could be implemented to split a query up between Queriers at smaller increments. This does not have the same deduplication correctness problem as query pushdown, as overlapping time-ranges can be effectively deduplicated. This implementation, however, presents other challenges around aligning queries in such a way that guarantees a sample on the threshold is not double counted in distinct shards. Generally, time-ranges are also more complex to handle compared to a simple hashmod of label value pairs as the scrape interval can be different for each metric. Horizontal sharding also does not address cardinality, instead sharding based on number of samples over a time range, which already highly compressible.
```
GiedriusS, line 123: Small suggestion regarding grammar:

```suggestion
*Extended horizontal sharding* Thanos Query Frontend already does horizontal (based on time-range) sharding by splitting up queries into smaller time-ranges. As an alternative to vertical sharding, more granular horizontal sharding could be implemented to split a query up between Queriers at smaller increments. This does not have the same deduplication correctness problem as query pushdown, as overlapping time-ranges can be effectively deduplicated. This implementation, however, presents other challenges around aligning queries in such a way that guarantees a sample on the threshold is not double counted in distinct shards. Generally, time-ranges are also more complex to handle compared to a simple hashmod of label value pairs as the scrape interval can be different for each metric. Horizontal sharding also does not address cardinality, instead, sharding is based on number of samples over a time range, which are already highly compressible.
```

GiedriusS:
@fpetkovski the GitHub UI doesn't add a DCO :joy:

GiedriusS (dismissed):
:+1: LGTM, solid work

GiedriusS (approved):

PR #5356, opened 2022-05-11
Group memcached keys based on server when performing batch gets

### LINKED ISSUE #5353: Proposal: change the way memached keys are batched by the client

Hello Thanos team!

Currently, keys given to the [`GetMulti` method](https://github.com/thanos-io/thanos/blob/8d1179e2cc76b6eb0d04acb4d02e4ae546ae38a7/pkg/cacheutil/memcached_client.go) are split [into batches](https://github.com/thanos-io/thanos/blob/8d1179e2cc76b6eb0d04acb4d02e4ae546ae38a7/pkg/cacheutil/memcached_client.go#L478) and requests are made to memcached servers [in parallel](https://github.com/thanos-io/thanos/blob/8d1179e2cc76b6eb0d04acb4d02e4ae546ae38a7/pkg/cacheutil/cacheutil.go#L34-L39).

The problem is that because the keys aren't in any particular order, each batch will likely need to make a connection to every server in the memcached cluster.

For example, if we have 1000 keys, a batch size of 100 keys, and 10 servers in our memcached cluster:

* Each of the 1000 keys is split into a batch of 100 keys:
* Each batch of 100 keys will likely need to talk to all 10 servers
* `10 batches * 10 servers per batch` means making 100 connections for 1000 keys  

Instead of splitting keys for batching not using any knowledge of the keys, we could instead order the 1000 keys based on which server they will hash to. Then, when splitting the keys into batches each batch will likely only need to make a connection to 1 server.

* Each of the 1000 keys is ordered by which server the key lives on before being split into a batch of 100 keys
* Each batch of 100 keys will likely only need to talk to a single server (hash based so it might be more)
* `10 batches * 1 server per batch` means making ~10 connections for 1000 keys.

I've created a [proof-of-concept](https://github.com/56quarters/thanos/commit/ca1e541e87b6c113ed79d8e31a9ca9f70bcafd47) of ordering keys based on the server they hash to and am currently testing in a Mimir dev cluster.

N/A

### Additional context

This was originally noticed and written up by @bboreham here: https://github.com/grafana/mimir/issues/423

Initial results are promising. I'm not sure if Thanos deployments use the same caches as Mimir but I'll describe what I've seen:

* Number of connections to the chunks cache (TSDB data, large objects) is roughly the same
* Number of timeouts reading from the chunks cache has dropped
* Number of connections to the index cache (TSDB index, small objects) has dropped
* Number of timeouts is unchanged (very few previously, very few now)

Note that this change was deployed around **20:15** in the following graphs:

Chunks cache connections:
Screenshot 2022-05-09 at 16-36-33 Explore - cortex-dev-01-dev-us-central-0 - Grafana

Index cache connections:
Screenshot 2022-05-09 at 16-34-31 Explore - cortex-dev-01-dev-us-central-0 - Grafana

Timeouts:
Screenshot 2022-05-09 at 18-04-34 Explore - cortex-dev-01-dev-us-central-0 - Grafana

56quarters:
Signed-off-by: Nick Pillitteri <nick.pillitteri@grafana.com>

## Changes

Order and group keys during batch get operations based on the memcached
server they will be sharded to. This reduces the number of connections
that must be made within each batch of get operations.

## Verification

* Unit tests added
* Ran with changes in Mimir dev cluster

Fixes #5353

### DIFF
--- a/pkg/cacheutil/memcached_client.go
+++ b/pkg/cacheutil/memcached_client.go
@@ -472,10 +472,10 @@ func (c *memcachedClient) getMultiBatched(ctx context.Context, keys []string) ([
 		getMultiGate = c.getMultiGate
 	}
 
-	// Group keys based on which memcached server they will be sharded to. Grouping keys that
+	// Sort keys based on which memcached server they will be sharded to. Sorting keys that
 	// are on the same server together before splitting into batches reduces the number of
 	// connections required and increases the number of "gets" per connection.
-	groupedKeys := c.groupKeysByServer(keys)
+	sortedKeys := c.sortKeysByServer(keys)
 
 	// Allocate a channel to store results for each batch request. The max concurrency will be
 	// enforced by doWithBatch.
@@ -489,7 +489,7 @@ func (c *memcachedClient) getMultiBatched(ctx context.Context, keys []string) ([
 	// from `results` below. The wrapped `getMultiSingle` method will still check our context
 	// and short-circuit if it has been canceled.
 	_ = doWithBatch(context.Background(), len(keys), c.config.MaxGetMultiBatchSize, getMultiGate, func(startIndex, endIndex int) error {
-		batchKeys := groupedKeys[startIndex:endIndex]
+		batchKeys := sortedKeys[startIndex:endIndex]
 
 		res := &memcachedGetMultiResult{}
 		res.items, res.err = c.getMultiSingle(ctx, batchKeys)
@@ -544,14 +544,14 @@ func (c *memcachedClient) getMultiSingle(ctx context.Context, keys []string) (it
 	return items, err
 }
 
-// groupKeysByServer groups cache keys within a slice based on which server they are
+// sortKeysByServer sorts cache keys within a slice based on which server they are
 // sharded to using a memcache.ServerSelector instance. The keys are ordered so keys
 // on the same server are next to each other. Any errors encountered determining which
-// server a key should be on will result in returning keys ungrouped, in the same order
-// they were supplied in. Note that output is not guaranteed to be any particular order
-// *except* that keys will be grouped by server. The order of keys returned may change
-// from call to call.
-func (c *memcachedClient) groupKeysByServer(keys []string) []string {
+// server a key should be on will result in returning keys unsorted (in the same order
+// they were supplied in). Note that output is not guaranteed to be any particular order
+// *except* that keys sharded to the same server will be together. The order of keys
+// returned may change from call to call.
+func (c *memcachedClient) sortKeysByServer(keys []string) []string {
 	bucketed := make(map[string][]string)
 
 	for _, key := range keys {
--- a/pkg/cacheutil/memcached_client.go
+++ b/pkg/cacheutil/memcached_client.go
@@ -88,6 +88,14 @@ type memcachedClientBackend interface {
 	Set(item *memcache.Item) error
 }
 
+// updatableServerSelector extends the interface used for picking a memcached server
+// for a key to allow servers to be updated at runtime.
+type updatableServerSelector interface {
+	memcache.ServerSelector
+
+	SetServers(servers ...string) error
+}
+
 // MemcachedClientConfig is the config accepted by RemoteCacheClient.
 type MemcachedClientConfig struct {
 	// Addresses specifies the list of memcached addresses. The addresses get
@@ -162,7 +170,7 @@ type memcachedClient struct {
 	logger   log.Logger
 	config   MemcachedClientConfig
 	client   memcachedClientBackend
-	selector *MemcachedJumpHashSelector
+	selector updatableServerSelector
 
 	// Name provides an identifier for the instantiated Client
 	name string
@@ -238,7 +246,7 @@ func NewMemcachedClientWithConfig(logger log.Logger, name string, config Memcach
 func newMemcachedClient(
 	logger log.Logger,
 	client memcachedClientBackend,
-	selector *MemcachedJumpHashSelector,
+	selector updatableServerSelector,
 	config MemcachedClientConfig,
 	reg prometheus.Registerer,
 	name string,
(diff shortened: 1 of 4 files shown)


pkg/cacheutil/memcached_client.go
```diff
@@ -531,6 +544,35 @@ func (c *memcachedClient) getMultiSingle(ctx context.Context, keys []string) (it
...
 }
 
+// groupKeysByServer groups cache keys within a slice based on which server they are
+// sharded to using a memcache.ServerSelector instance. The keys are ordered so keys
+// on the same server are next to each other. Any errors encountered determining which
+// server a key should be on will result in returning keys ungrouped, in the same order
+// they were supplied in. Note that output is not guaranteed to be any particular order
+// *except* that keys will be grouped by server. The order of keys returned may change
+// from call to call.
+func (c *memcachedClient) groupKeysByServer(keys []string) []string {
```
bwplotka, line 554:
```suggestion
func (c *memcachedClient) sortKeysByServer(keys []string) []string {
```

bwplotka: Technically we are not returning groups

pkg/testutil/testutil.go
```diff
@@ -87,6 +87,46 @@ func Equals(tb testing.TB, exp, act interface{}, v ...interface{}) {
 	tb.Fatal(sprintfWithLimit("\033[31m%s:%d:"+msg+"\n\n\texp: %#v\n\n\tgot: %#v%s\033[39m\n\n", filepath.Base(file), line, exp, act, diff(exp, act)))
 }
 
+// Contains fails the test if needle is not contained within haystack, if haystack or needle is
+// an empty slice, or if needle is longer than haystack.
+func Contains(tb testing.TB, haystack, needle []string) {
```
bwplotka, line 92: 👍🏽

pkg/cacheutil/memcached_client.go
```diff
@@ -464,6 +472,11 @@ func (c *memcachedClient) getMultiBatched(ctx context.Context, keys []string) ([
 		getMultiGate = c.getMultiGate
 	}
 
+	// Group keys based on which memcached server they will be sharded to. Grouping keys that
```
bwplotka, line 475:
```suggestion
	// Sort keys based on which memcached server they will be sharded to. Grouping keys that
```

bwplotka (dismissed):
I wonder if it makes sense to use `sort` vs `group` but it's a minor nit - otherwise, this is pretty solid. Thanks!

pkg/cacheutil/memcached_client.go
56quarters: Good point! Will update to use "sort" nomenclature everywhere.

pkg/cacheutil/memcached_client.go
```diff
@@ -88,6 +88,14 @@ type memcachedClientBackend interface {
 	Set(item *memcache.Item) error
 }
 
+// updatableServerSelector extends the interface used for picking a memcached server
+// for a key to allow servers to be updated at runtime.
+type updatableServerSelector interface {
+	memcache.ServerSelector
+
+	SetServers(servers ...string) error
```
GiedriusS, line 96: This is only for Cortex/Mimir, right? What's the use case behind this function?

56quarters: The use case was to allow me to swap out the `MemcachedJumpHashSelector` with a mock so that I could test the `sortKeysByServer` method. I will add a comment to this method based on the existing comment on `MemcachedJumpHashSelector.SetServers`

pkg/cacheutil/memcached_client.go
```diff
@@ -531,6 +544,35 @@ func (c *memcachedClient) getMultiSingle(ctx context.Context, keys []string) (it
...
+		}
+
+		addrString := addr.String()
+		bucketed[addrString] = append(bucketed[addrString], key)
+	}
+
+	var out []string
+	for srv := range bucketed {
+		out = append(out, bucketed[srv]...)
+	}
```
bboreham, line 571: Did you consider returning `[][]string` here, to save copying every entry?
(I've seen millions of keys here, when called from Mimir).

56quarters: I hadn't. The batching method needs `[]string` to split up. I'm not sure how I could return `[][]string` here without needing to copy it into a `[]string` eventually.

56quarters: Yes! I thought about that initially and decided to go with the bucketing approach since it's `O(n)` compared to sorting which is `O(n log n)`. I can run some benchmarks if you think we should investigate that?

56quarters: Pushed some benchmarks here: https://github.com/56quarters/thanos/commit/5fc84926a63a7c7d026ac1866b14f1061fe96424

Sorting appears to be slower (slightly slower for a small number of keys, much slower for a large number) and perform more allocations in all but the smallest number of keys.

```
$ go test -cpu=1 -run=XXX -benchmem -bench=BenchmarkMemcachedClient_sortKeysByServer ./pkg/cacheutil/
goos: linux
goarch: amd64
pkg: github.com/thanos-io/thanos/pkg/cacheutil
cpu: Intel(R) Core(TM) i9-10885H CPU @ 2.40GHz
BenchmarkMemcachedClient_sortKeysByServer/10_keys                 514545              2259 ns/op             960 B/op         20 allocs/op
BenchmarkMemcachedClient_sortKeysByServer/100_keys                 76731             15666 ns/op            9152 B/op        116 allocs/op
BenchmarkMemcachedClient_sortKeysByServer/1000_keys                 8308            146631 ns/op           81315 B/op       1022 allocs/op
BenchmarkMemcachedClient_sortKeysByServer/10000_keys                 656           2213951 ns/op         1187478 B/op      10036 allocs/op
BenchmarkMemcachedClient_sortKeysByServer2/10_keys                424206              3002 ns/op             680 B/op         28 allocs/op
BenchmarkMemcachedClient_sortKeysByServer2/100_keys                21447             54334 ns/op           12536 B/op        522 allocs/op
BenchmarkMemcachedClient_sortKeysByServer2/1000_keys                2263            538877 ns/op          120551 B/op       5022 allocs/op
BenchmarkMemcachedClient_sortKeysByServer2/10000_keys                205           5624709 ns/op         1199250 B/op      49967 allocs/op
PASS
ok      github.com/thanos-io/thanos/pkg/cacheutil       11.488s
```

56quarters: Another option could be introducing an iterator object here that knows how to iterate over an `[][]string` (or `map[string][]string`) but that seems like a lot to save a few allocations to me.

GiedriusS (approved):
LGTM, thanks for the benchmarks! This is a neat optimization. Let's see if @bboreham has more comments before merging.

GiedriusS:
:+1: let's merge this and try to improve it if perf looks bad

PR #5357, opened 2022-05-11
groupcache: fix handling of slashes

GiedriusS:
Use https://github.com/julienschmidt/httprouter#catch-all-parameters for
the groupcache route otherwise slashes in the cache's key gets
interpreted by the router and thus groupcache's function never gets
invoked, and all clients get 404.

Remove test regarding cache hit because now Thanos Store during test
constantly generates cache hits due to 1s delay between block
information refreshes.


matej-g (approved):
Nicely done :beers:

bwplotka:
Well found! Thanks 👍🙏👍

PR #5368, opened 2022-05-16
ruler: add metric thanos alert sent by alert name

fgouteroux:
Hello thanos team,

It will be great to have the total number of alerts sent by alertmanager and by alert name.  It will be useful to make some stats about how often an alert is triggered.

I'm not sure that if it's relevant to add another metric, and about the name too.

## Changes

Adding metric: Total number of alerts sent by alertmanager and alert name.

## Verification

Trigger an alert
Check the metric thanos_alert_sender_alerts_sent_by_alertname_total in  thanos ruler metrics endpoint

### DIFF
--- a/pkg/alert/alert.go
+++ b/pkg/alert/alert.go
@@ -229,10 +229,11 @@ type Sender struct {
 	alertmanagers []*Alertmanager
 	versions      []APIVersion
 
-	sent    *prometheus.CounterVec
-	errs    *prometheus.CounterVec
-	dropped prometheus.Counter
-	latency *prometheus.HistogramVec
+	sent            *prometheus.CounterVec
+	sentByAlertName *prometheus.CounterVec
+	errs            *prometheus.CounterVec
+	dropped         prometheus.Counter
+	latency         *prometheus.HistogramVec
 }
 
 // NewSender returns a new sender. On each call to Send the entire alert batch is sent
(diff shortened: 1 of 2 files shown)


yeya24 (commented):
I love this idea and I think this could be a nice addition to have.
@fgouteroux Can you please rebase your PR on main? This should fix the CI issue.

GiedriusS (commented):
Thanks for your work! However, what's the use-case here? If it's about tracing where specific alerts came from then I believe Alertmanager itself should be a better place to solve this. There were lots of discussions about the identity of an alert during Prometheus Day @ KubeCon last week. I don't think the `alertname` is enough to identify an alert. And I believe that if we'd give users such a metric, the next question would be: what about other labels? Hence, I don't think that this is a good idea

fgouteroux:
> Thanks for your work! However, what's the use-case here? If it's about tracing where specific alerts came from then I believe Alertmanager itself should be a better place to solve this. There were lots of discussions about the identity of an alert during Prometheus Day @ KubeCon last week. I don't think the `alertname` is enough to identify an alert. And I believe that if we'd give users such a metric, the next question would be: what about other labels? Hence, I don't think that this is a good idea

It's not about tracing, it could be usefull to see which alertname is mostly triggered. In my case I'm using one ruler per tenant with a shared alertmanager, so each tenant manage his own alerts rules but have no access to the alertmanager. We only expose the alertmanager api through a custom proxy with authentification. Based on that metrics provided by the alertmanager are not very helpful as we only see the total of alert sent to an integration like PagerDuty. This is why ruler metrics fit to our design and let us provide a grafana dashboard with all ruler metrics related to the tenant.

yeya24:
> I don't think the alertname is enough to identify an alert.

Good point. I think alertname is only unique in one rule group. But can be duplicated across groups.

Issue #4810, opened 2021-10-28
Store: Make initial sync more robust


matej-g:
I stumbled upon this during some tests - if the [initial sync](https://github.com/thanos-io/thanos/blob/main/cmd/thanos/store.go#L365) fails, this stops the whole initialization of a store and the process quits. The connection error might, however, be just ephemeral (e.g. the store is not ready yet or there was a temporary issue with the connection). It might be worthwhile to retry the initial sync in such cases, in order not to cause the process to exit directly after one failed attempt.

matej-g:
I think this is still valid

shaoyi1997:
Hi! I can take this issue :)

PlytonRexus:
Hi! Is this issue still valid?

Kartik-Garg:
I have raised a working PR against this issue :)

Kartik-Garg:
@matej-g if the [PR](https://github.com/thanos-io/thanos/pull/6050) fixes this issue, can we close this issue now with respect to the PR? :smile:

matej-g:
@Kartik-Garg yes please, thank you 🙇

PR #5373, opened 2022-05-19
store: retry bucket store initial sync upon failure

shaoyi1997:
Signed-off-by: shaoyi1997 <e0310084@u.nus.edu>

This PR fixes [#4810](https://github.com/thanos-io/thanos/issues/4810).

## Changes

Currently, initial sync of the bucket store fails without retrying upon a temporary bucket error.
This PR wraps `bs.initialSync` in `runutil.RetryWithLog` and introduces 2 flags, `--initial-sync-retry-interval` and `--initial-sync-retry-timeout`, to allow configuration of the interval and timeout durations for retrying the initial sync.

## Verification

Immediately after starting the minio server, start store ([gist](https://gist.github.com/shaoyi1997/7d7d0eeffff64a88339267ad96a94d11), adapted from `quickstart.sh`)
Store should retry sync and succeed with initial sync

### DIFF
--- a/cmd/thanos/store.go
+++ b/cmd/thanos/store.go
@@ -61,6 +61,8 @@ type storeConfig struct {
 	maxConcurrency              int
 	component                   component.StoreAPI
 	debugLogging                bool
+	initialSyncRetryInterval    time.Duration
+	initialSyncRetryTimeout     time.Duration
 	syncInterval                time.Duration
 	blockSyncConcurrency        int
 	blockMetaFetchConcurrency   int
@@ -113,6 +115,12 @@ func (sc *storeConfig) registerFlag(cmd extkingpin.FlagClause) {
 	sc.component = component.Store
 
 	sc.objStoreConfig = *extkingpin.RegisterCommonObjStoreFlags(cmd, "", true)
+
+	cmd.Flag("initial-sync-retry-interval", "Interval between retries for the initial sync of the bucket store.").
+		Default("2s").DurationVar(&sc.initialSyncRetryInterval)
+
+	cmd.Flag("initial-sync-retry-timeout", "Timeout for the initial sync of the bucket store.").
+		Default("3m").DurationVar(&sc.initialSyncRetryTimeout)
 
 	cmd.Flag("sync-block-duration", "Repeat interval for syncing the blocks between local and remote view.").
 		Default("3m").DurationVar(&sc.syncInterval)
@@ -364,7 +372,10 @@ func runStore(
 			level.Info(logger).Log("msg", "initializing bucket store")
 			begin := time.Now()
 
-			err := runutil.RetryWithLog(logger, 2*time.Second, ctx.Done(), func() error {
+			initialSyncCtx, cancel := context.WithTimeout(ctx, conf.initialSyncRetryTimeout)
+			defer cancel()
+
+			err := runutil.RetryWithLog(logger, conf.initialSyncRetryInterval, initialSyncCtx.Done(), func() error {
 				return bs.InitialSync(ctx)
 			})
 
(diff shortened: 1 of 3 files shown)


cmd/thanos/store.go
```diff
@@ -363,14 +363,21 @@ func runStore(
 
 			level.Info(logger).Log("msg", "initializing bucket store")
 			begin := time.Now()
-			if err := bs.InitialSync(ctx); err != nil {
+
+			err := runutil.RetryWithLog(logger, 2*time.Second, ctx.Done(), func() error {
```
yeya24, line 367: Can we make the interval configurable? Btw using `ctx.Done` here doesn't have any timeout. Let use another context with a configurable timeout.

cmd/thanos/store.go
```diff
@@ -363,14 +363,21 @@ func runStore(
...
+				syncErr := bs.InitialSync(ctx)
+				return errors.Wrap(syncErr, "bucket store initial sync")
```
yeya24, line 369: We don't need to wrap the error here. Otherwise in line 374 we will wrap the error twice.

cmd/thanos/store.go
shaoyi1997: Yep that makes a lot of sense! Have added 2 flags, `--initial-sync-retry-timeout` and `--initial-sync-retry-interval` for configuring the retry behavior.

matej-g: I'm a bit wary of adding two new flags just for this feature, my initial idea was to have a simple retry mechanism :thinking:. I think having configurable interval makes sense, but how about setting some sensible timeout for all cases?

shaoyi1997: I did a brief perusal of the codebase and found a `timeout` flag for downloading metadata from a bucket in [`tools_bucket.go`](https://github.com/thanos-io/thanos/blob/7ba274c69cb0a848c7e8e1c593221faa80d8b6e6/cmd/thanos/tools_bucket.go#L199). Perhaps we could still retain the `--initial-sync-retry-timeout` flag to allow user configuration of bucket timeout across components (if that makes sense, cmiiw), else we can set the timeout to 5 minutes for all cases following the default for `timeout` in `tools_bucket.go`

matej-g: I'm not strictly against, just a bit cautious about adding new flags, just to keep things simple for users (I guess initial sync is not such a defining feature that many users would need to tweak this know, but might be just my feeling). Thoughts @yeya24?

Either way let's try to move this forward, it looks good to me otherwise!

matej-g: ping @yeya24, looks like there's not much missing to carry this over the finish line.

cmd/thanos/store.go
```diff
@@ -114,6 +116,12 @@ func (sc *storeConfig) registerFlag(cmd extkingpin.FlagClause) {
 
 	sc.objStoreConfig = *extkingpin.RegisterCommonObjStoreFlags(cmd, "", true)
 
+	cmd.Flag("initial-sync-retry-interval", "Interval between retries for the initial sync of the bucket store.").
+		Default("2s").DurationVar(&sc.initialSyncRetryInterval)
+
+	cmd.Flag("initial-sync-retry-timeout", "Timeout for the initial sync of the bucket store.").
+		Default("3m").DurationVar(&sc.initialSyncRetryTimeout)
+
```
bwplotka, line 124: I wonder if this will not break users - e.g. if initial sync is failing on config error and you rely pod to crashloop for that.. you just make this process much longer 🤔 


Potentially keeping it smaller like 30s for default and then you can play with lager value makes sense

bwplotka (changes requested):
Hmmm, I am torn on this. Initial sync is meant to test your configuration and not set your prod ready until you cannot connect to the bucket (e.g to catch permission denied). I think we need to resign this path if we want to retry - or perhaps retry on special errors only. 

For safety, we could for now make safe default (not breaking existing setups). WDYT?

PR #5381, opened 2022-05-24
update_issue_template

Akshit42-hue:
Signed-off-by: Axit Patel <patelakshit2025@gmail.com>

## Changes

As discuss in the call. I have made some changes in issue template

### DIFF
--- a/.github/ISSUE_TEMPLATE/bug-report.md
+++ b/.github/ISSUE_TEMPLATE/bug-report.md
@@ -50,7 +50,6 @@ Uncomment if you would like to post collapsible logs:
 -->
 
 **Anything else we need to know**:
-
 <!--
 Uncomment and fill if you use not casual environment or if it might be relevant.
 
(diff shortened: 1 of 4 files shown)


.github/ISSUE_TEMPLATE/bug-report.md
```diff
@@ -60,3 +59,16 @@ Uncomment and fill if you use not casual environment or if it might be relevant.
 - Others:
 
 -->
+
+**Screenshots**
```
bwplotka, line 63: hm, all of this is already present in the comment. Do you mean we should uncomment what we have above? (:

Akshit42-hue: The contribution guide should be present on the issue template the main goal behind this is new contributor can easily set-up on their local system. Let's assume the issue is related to website so in the website template their should be a guidelines on how to setup website on their local system :)

PR #5382, opened 2022-05-24
port codebase to adapt to new thanos errors pkg replacing github.com/pkg/errors

### LINKED ISSUE #5176: Find replacement for https://github.com/pkg/errors

bisakhmondal:
In continuation of #5239 

Related Issue: closes #5176 

## Changes

- Replace old github.com/pkg/errors import
- Update New,Errorf -> `Newf` and Wrap -> `Wrapf`
- Pr #5239, depends on `github.com/thanos-io/thanos/pkg/testutil` internal pkg leading to give rise of cyclic dependency - Fixed

Used the following snippet to automate the replacement and then a `make format` with manual imports ordering
```bash
for file in $(git grep -l "github.com/pkg/errors" | grep ".go"); do
  sed -i 's/errors\.Wrap(/errors\.Wrapf(/g' $file
  sed -i 's/errors\.Errorf(/errors\.Newf(/g' $file
  sed -i 's/errors\.New(/errors\.Newf(/g' $file
  sed -i 's/github\.com\/pkg\/errors/github\.com\/thanos-io\/thanos\/pkg\/errors/g'  $file
  echo "$file processing done"
done
```

## Verification
`make lint && make build` works successfully


saswatamcode:
Thanks a lot for this effort! I wonder if we should do this port in one go, or split it component-wise somehow. 🙂 
Also, test failures seem related!

matej-g (commented):
Looks good! I'm impartial as to whether we do this in one go or split by components or so. Since it's mechanical change, I'm fine with having it all here. However I'm also wondering if those failures are related :thinking:.

saswatamcode:
Would you like to push this forward @bisakhmondal? 🙂

matej-g:
What's the status of this?

bisakhmondal:
the code has been ported to https://github.com/efficientgo/core
I think once @bwplotka finalizes the APIs there (had some double thoughts), I can clean it up with the new imports.
That's the plan, right? @bwplotka

matej-g:
Still needed

Issue #5349, opened 2022-05-09
QueryFrontend: compile regex once when matching strings


yeya24:
From code https://github.com/thanos-io/thanos/blob/main/pkg/queryfrontend/roundtrip.go#L123, we are compiling the regex every time to match strings. This can be optimized by only compiling the regex once and checking if the string matches.

```go
			matched, err := regexp.MatchString("/api/v1/label/.+/values$", r.URL.Path)
			if err == nil && matched {
				return labelValuesOp
			}
```

sandipanpanda:
Hi, I am new to Thanos and would like to help in fixing this.

yeya24:
Please go ahead. @sandipanpanda

sandipanpanda:
Yes, I am working on it.

Akshit42-hue:
Hey @jgbernalp, I would like to give it a try

djdongjin:
I think this can be closed now. @yeya24

yeya24:
Thanks @djdongjin.

PR #5383, opened 2022-05-24
QueryFrontend: pre-compile regexp

djdongjin:
Signed-off-by: Jin Dong <djdongjin95@gmail.com>

## Changes

Pre-compile regexp. Solve #5349.

`regexp.MustCompile` returns `regexp` directly or, if failed, panic.

Another way is:

```go
var opPattern, _ = regexp.Compile("/api/v1/label/.+/values$")
...
if opPattern && opPattern.MatchString(r.URL.Path) {
    return labelValuesOp
}
```

## Verification

Unit tests and `make test` succeeded.

### DIFF
--- a/pkg/queryfrontend/roundtrip.go
+++ b/pkg/queryfrontend/roundtrip.go
@@ -27,7 +27,7 @@ const (
 	seriesOp       = "series"
 )
 
-var opPattern = regexp.MustCompile("/api/v1/label/.+/values$")
+var labelValuesPattern = regexp.MustCompile("/api/v1/label/.+/values$")
 
 // NewTripperware returns a Tripperware which sends requests to different sub tripperwares based on the query type.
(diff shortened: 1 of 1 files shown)


onprem (approved):
Looks good. Thanks!

pkg/queryfrontend/roundtrip.go
```diff
@@ -27,6 +27,8 @@ const (
 	seriesOp       = "series"
 )
 
+var opPattern = regexp.MustCompile("/api/v1/label/.+/values$")
```
yeya24, line 30: What about renaming `opPattern` to `labelValuesPattern`? In case we have more patterns to match in the future.

djdongjin: renamed.

yeya24 (approved):
LGTM. Thanks for the contribution!

PR #5394, opened 2022-05-29
rule: fix empty $externalLabels when templating labels in rule.

roastiek:
Signed-off-by: Rostislav Benes <r.dee.b.b@gmail.com>

## Changes

External labels set up by `--label` option of the rule are not passed down when templating labels of alerts or recording rules, so `$externalLabels` is always an empty map.

The `externalLabels` parameter of `Manager.Update` method is used only for templates and should have no additional effect.


yeya24 (approved):
Thanks! LGTM.

PR #5400, opened 2022-05-31
Expose block.GatherFileStats.

pstibrany:
* [na] I added CHANGELOG entry for this change.

## Changes

This PR exposes `block.GatherFileStats` function so that it can be used outside of `Upload` function. I've also removed TODOs from it, as I don't think they should be implemented.

## Verification

I haven't changed the function itself, only renamed it.


yeya24 (approved):
LGTM

Issue #5395, opened 2022-05-30
Federate the api/v1/status/tsdb endpoint from Prometheus


fpetkovski:
### Is your proposal related to a problem?

It would be great if Thanos Query could federate the `api/v1/status/tsdb` endpoint from Prometheus instances. This will allow users to analyze cardinality on the fly.

### Describe the solution you'd like

Thanos Query could fanout to Prometheus sidecars and return a federated result from all endpoints. 

### Describe alternatives you've considered

We currently run https://github.com/thought-machine/prometheus-cardinality-exporter, but this requires ingesting metrics which further increases cardinality :)

bwplotka:
Yes, I would see that as a potential gRPC API for our endpoints that Querier could merge. 

> We currently run https://github.com/thought-machine/prometheus-cardinality-exporter, but this requires ingesting metrics which further increases cardinality :)

Not if you use separate meta-monitoring which is a best practice (: 

I think the main questions is to nail down exact API - what we want to people use it for and if a nice metric exposed by the receiver to meta-monitoring cannot fulfil the same purpose. 

### Why not

* Metrics exposed by receiver could hold the same information?
* Is this data for admin / operator of Thanos or for Thanos user?

yeya24:
@fpetkovski Any next step for this issue?

fpetkovski:
Unfortunately I don't have capacity to work on this right now, I am primarily focused on the new engine. If no one has stepped up to tackle this, maybe we can close it for now?

PR #4818, opened 2021-10-30
store: Add Groupcache as a cache backend

akanshat:
* [x] I added CHANGELOG entry for this change.
* [ ] Change is not relevant to the end user.

## Changes
- Added Groupcache.
- Implemented Cache interface to groupcache.
- Added a new package cachekey.

## Verification
Existing tests passed.

Issue #4651, opened 2021-09-09
Docs Bug -- Out-of-order Samples Error  troubleshooting has an unclear remedy


robertjsullivan:
**Thanos, Prometheus and Golang version used**: N/A

**Object Storage Provider**:

**What happened**:
On this docs page https://thanos.io/tip/operating/troubleshooting.md/#out-of-order-samples-error, the resolution for "Remote  Prometheus is running in high availability mode" references setting replica_external_label_name.  I searched the web and couldn't find out where that could be set in Thanos or Prometheus. Is setting this value required to send HA pair Prometheus data to the same Thanos Receiver? Or would unique external_labels in Prometheus be enough? I think it would be good if the install docs here also mentioned how to handle HA Prometheus pairs: https://thanos.io/tip/components/receive.md/ , if any considerations are necessary.

For reference we are running into the "Error on ingesting out-of-order samples" error when running two Prometheus servers scraping the same data, but with unique external_labels.  Scaling down to a single Prometheus server solves the problem. I'm hoping to verify we have the correct configuration for this scenario before troubleshooting further.  

**What you expected to happen**:
Able to find docs on how to set up and troubleshoot the Thanos Receiver with HA Prometheus pairs.

**How to reproduce it (as minimally and precisely as possible)**:
N/A

**Full logs to relevant components**:
N/A

**Anything else we need to know**:
Hopefully this is the right place to raise a docs bug.  If not, please let me know and I'll post elsewhere. Thanks!

yeya24:
`replica_external_label_name` represents the external label configured in your Prometheus HA pairs that is used for identifying Prometheus replicas.

> Is setting this value required to send HA pair Prometheus data to the same Thanos Receiver? Or would unique external_labels in Prometheus be enough?

Anyway, different Prometheus replicas need unique external labels.

robertjsullivan:
Thanks for the clarification.  So if I set my Prometheis up with:

> global:
>   external_labels:
>     some-label: some-unique-value

Do I need to tell the Thanos receiver what label identifies the replica as unique (i.e. some-label)?  Or is the fact that the time series that are coming out of those Prometheus pairs being unique is enough?  Does the receiver attempt to do any sort of deduplication?

yeya24:
> Thanks for the clarification. So if I set my Prometheis up with:
> 
> > global:
> > external_labels:
> > some-label: some-unique-value
> 
[...]

PR #5402, opened 2022-05-31
Expose tsdb status in receiver

fpetkovski:
This commit implements the `api/v1/status/tsdb` API in the Receiver.

## Changes

Implemented the `api/v1/status/tsdb` endpoint in the Receiver.

## Verification

Manual testing

### DIFF
--- a/pkg/api/status/v1.go
+++ b/pkg/api/status/v1.go
@@ -1,0 +1,125 @@
+// Copyright (c) The Thanos Authors.
+// Licensed under the Apache License 2.0.
+
+// Copyright 2016 The Prometheus Authors
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package status
+
+import (
+	"fmt"
+	"math"
+	"net/http"
+
+	"github.com/go-kit/log"
+	"github.com/opentracing/opentracing-go"
+	"github.com/prometheus/client_golang/prometheus"
+	"github.com/prometheus/common/route"
+	"github.com/prometheus/prometheus/model/labels"
+	"github.com/prometheus/prometheus/tsdb"
+	"github.com/prometheus/prometheus/tsdb/index"
+	v1 "github.com/prometheus/prometheus/web/api/v1"
+	"github.com/thanos-io/thanos/pkg/api"
+	extpromhttp "github.com/thanos-io/thanos/pkg/extprom/http"
+	"github.com/thanos-io/thanos/pkg/logging"
+)
+
+// Stat holds the information about individual cardinality.
+type Stat struct {
+	Name  string `json:"name"`
+	Value uint64 `json:"value"`
+}
+
+func convertStats(stats []index.Stat) []Stat {
+	result := make([]Stat, 0, len(stats))
+	for _, item := range stats {
+		item := Stat{Name: item.Name, Value: item.Count}
+		result = append(result, item)
+	}
+	return result
+}
+
+// TSDBStatus has information of cardinality statistics from postings.
+type TSDBStatus struct {
+	HeadStats                   v1.HeadStats `json:"headStats"`
+	SeriesCountByMetricName     []Stat       `json:"seriesCountByMetricName"`
+	LabelValueCountByLabelName  []Stat       `json:"labelValueCountByLabelName"`
+	MemoryInBytesByLabelName    []Stat       `json:"memoryInBytesByLabelName"`
+	SeriesCountByLabelValuePair []Stat       `json:"seriesCountByLabelValuePair"`
+}
+
+type GetStatsFunc func(r *http.Request, statsByLabelName string) (*tsdb.Stats, error)
+
+type Options struct {
+	GetStats GetStatsFunc
+	Registry *prometheus.Registry
+}
+
+type StatusAPI struct {
+	getTSDBStats GetStatsFunc
+	options      Options
+}
+
+func New(opts Options) *StatusAPI {
+	return &StatusAPI{
+		getTSDBStats: opts.GetStats,
+		options:      opts,
+	}
+}
+
+func (sapi *StatusAPI) Register(r *route.Router, tracer opentracing.Tracer, logger log.Logger, ins extpromhttp.InstrumentationMiddleware, logMiddleware *logging.HTTPServerMiddleware) {
+	instr := api.GetInstr(tracer, logger, ins, logMiddleware, false)
+	r.Get("/api/v1/status/tsdb", instr("tsdb_status", sapi.httpServeStats))
+}
+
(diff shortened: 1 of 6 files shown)


pkg/api/status/v1.go
```diff
@@ -0,0 +1,110 @@
...
+	result := make([]Stat, 0, len(stats))
+	for _, item := range stats {
+		item := Stat{Name: item.Name, Value: item.Count}
+		result = append(result, item)
+	}
+	return result
+}
+
+// TSDBStatus has information of cardinality statistics from postings.
+type TSDBStatus struct {
```
fpetkovski, line 38: We could remove these structs if https://github.com/prometheus/prometheus/pull/10783 gets merged.

saswatamcode (commented):
Thanks! 🚀

Just a question! 🙂

pkg/receive/multitsdb.go
```diff
@@ -289,6 +291,18 @@ func (t *MultiTSDB) TSDBExemplars() map[string]*exemplars.TSDB {
...
+func (t *MultiTSDB) Stats(tenantID, statsByLabelName string) *tsdb.Stats {
+	t.mtx.RLock()
+	defer t.mtx.RUnlock()
+
+	tenant, ok := t.tenants[tenantID]
+	if !ok {
+		return nil
+	}
+
+	return tenant.readyS.get().db.Head().Stats(statsByLabelName)
```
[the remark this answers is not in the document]
saswatamcode: Maybe documenting this in some way would be great!

pkg/api/status/v1.go
```diff
@@ -0,0 +1,126 @@
...
+// limitations under the License.
+
+package status
+
+import (
+	"fmt"
+	"math"
+	"net/http"
+
+	"github.com/thanos-io/thanos/pkg/api"
```
yeya24, line 24: Let's group the imports

pkg/api/status/v1.go
```diff
@@ -0,0 +1,126 @@
...
+	}
+}
+
+func (sapi *StatusAPI) Register(r *route.Router, tracer opentracing.Tracer, logger log.Logger, ins extpromhttp.InstrumentationMiddleware, logMiddleware *logging.HTTPServerMiddleware) {
+	instr := api.GetInstr(tracer, logger, ins, logMiddleware, false)
+	r.Get("/api/v1/status/tsdb", instr("tsdb_status", sapi.httpServeStats))
+}
+
+func (sapi *StatusAPI) httpServeStats(r *http.Request) (interface{}, []error, *api.ApiError) {
+	s, err := sapi.getTSDBStats(r, labels.MetricName)
```
yeya24, line 87: So this API can only fetch the TSDB stats for a specific tenant. Admins cannot use this API to to see the global TSDB stats (cross tenant).

fpetkovski: Yes, that's correct. We could extend the endpoint to provide a global view by traversing all tenants and merging the results.

pkg/api/status/v1.go
GiedriusS: It was merged, we can re-use the struct now :+1:

fpetkovski: Do I need to go get the latest prometheus version from master/main, or do we need to wait for a new point release? I'm not sure how Thanos manages dependencies.

yeya24: We can use latest main directly.

pkg/api/status/v1.go
yeya24: Sounds good. Let's have an issue to track it after this pr is merged.

pkg/api/status/v1.go
```diff
@@ -0,0 +1,125 @@
...
+func New(opts Options) *StatusAPI {
+	return &StatusAPI{
+		getTSDBStats: opts.GetStats,
+		options:      opts,
+	}
+}
+
+func (sapi *StatusAPI) Register(r *route.Router, tracer opentracing.Tracer, logger log.Logger, ins extpromhttp.InstrumentationMiddleware, logMiddleware *logging.HTTPServerMiddleware) {
+	instr := api.GetInstr(tracer, logger, ins, logMiddleware, false)
+	r.Get("/api/v1/status/tsdb", instr("tsdb_status", sapi.httpServeStats))
```
yeya24, line 82: Other federated APIs like rules, targets, exemplars are exposed as gRPC APIs at the store, and federations are done at the Querier level.
For this API do we have any plan to implement the gRPC version of it and then aggregated at the Query level?

fpetkovski: So we have this issue at the moment: https://github.com/thanos-io/thanos/issues/5395

Maybe we can discuss federation there, but I think it can be a good idea to do it. We just have to see how merging is going to work, and whether we want to merge the results in a single json object in the result, or have one object per store target. I think this is a question for getting tsdb stats for all tenants as well.

saswatamcode: Regarding merging, I'm wondering how data replication for Receive will affect it. AFAIU, we use quorum-based logic for replication, so maybe merging this would lead to some inaccuracies in TSDB statistics for a tenant?

fpetkovski: Yes agreed. The stats can get very inaccurate if we merge across replicas.

pkg/api/status/v1.go
yeya24: Yes, we need to update Cortex or Cortex folks will update Thanos when they need to

fpetkovski: One more thing, what is the reason for using github.com/akanshat/cortex instead github.com/cortexproject/cortex? Which one should actually be updated?

saswatamcode: I think we use it for groupcache changes https://github.com/thanos-io/thanos/pull/4818. Maybe Thanos version in fork can be updated cc: @akanshat?

akanshat: Yes, so while adding groupcache we had to make some changes in Thanos, which affected cortex imports just a little. Then we decided to update my fork of cortex and point Thanos to it, and once groupcache changes were merged we were going to create a PR to Cortex, to have those changes merged and point Thanos back to the original cortex.

Now [#4651](https://github.com/cortexproject/cortex/pull/4651) is open, it's been reviewed by @GiedriusS and @bwplotka but not by any cortex maintainer yet. I am not sure what the next step is. @GiedriusS would you like to help me out with this?

fpetkovski: While waiting for that to get merged, should we maybe rebase the fork against the latest master in Cortex in order to get this PR unblocked? Otherwise, we will blocked whenever we try to update the prometheus/prometheus dependency.

pkg/receive/multitsdb.go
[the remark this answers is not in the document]
fpetkovski: Docs added 👍

pkg/api/status/v1.go
bwplotka: No need for relase

bwplotka: We could rebase cortex fork.

bwplotka (approved):
I think this is solid. It is a good start, and we can go from here to federated merged view if needed. Also federated does not mean "merged". We could make an API that does not merge or merge partially or something. Many options. 

We can reuse struct later on too.

bwplotka:
Unrelated but important issue? 
image

Or main got broken from last consistent hashing?

fpetkovski:
Looks like the classic non-nil interface pointing to a nil object. Should be fixed now.

bwplotka (approved):

bwplotka (approved):
Thanks, LGTM! 👍🏽

Issue #5186, opened 2022-02-25
Rule: Alertmanager config creation does not error out when provided with an URL with an empty scheme


jessicalins:
**Thanos version used**: 0.24.0

**Object Storage Provider**: S3

**What happened**:
When configuring [alertmanagers.url](https://github.com/thanos-io/thanos/blob/main/cmd/thanos/config.go#L207) to for example:
```bash
dnssrv+_http._tcp.example.com
```
The Alertmanager configuration still gets created and Thanos Rule errors out with:

```bash
unsupported protocol scheme
```
The error is returned only when Thanos tries to [post](https://github.com/thanos-io/thanos/blob/main/pkg/alert/alert.go#L457) to the specified alert URL.

This seems to happen because the URL gets parsed [here](https://github.com/thanos-io/thanos/blob/main/pkg/alert/config.go#L92) by `url.Parse` and it does not create an error, even if the scheme is not in the format as expected from the [tests](https://github.com/thanos-io/thanos/blob/main/pkg/alert/config_test.go#L85)

Since `Scheme` is empty this [logic](https://github.com/thanos-io/thanos/blob/main/pkg/alert/config.go#L99) does not happen and `Scheme` remains empty, causing it to return an error just when trying to post to the configured URL.

**What you expected to happen**:

It would be nice to error out earlier, maybe during the config creation:
https://github.com/thanos-io/thanos/blob/main/pkg/alert/config.go#L92-L95

**How to reproduce it (as minimally and precisely as possible)**:
* Change `alertmanagers.url` config to a URL that has this format: `dnssrv+_http._tcp.example.com`
* It is also possible to change one of the test cases in [config_test.go](https://github.com/thanos-io/thanos/blob/main/pkg/alert/config_test.go#L47) and verify that the resulting struct does have an empty `Scheme` field.

cc @matej-g @saswatamcode

Harshitavkb26:
Hi, @matej-g. I am interested in working on this issue.

matej-g:
Hey @Harshitavkb26 go for it :rocket:

jessicalins:
hi, what is the status of this issue? In case no one is working on it I can pick it up :)

matej-g:
@jessicalins go for it!

matej-g:
I'm wondering if there's space to unify the formats between all configs / components :thinking:, since that DNS directive is being used with other components. Maybe there is way to reuse or adapt this to use the `discovery/dns` package?

[...]

PR #5405, opened 2022-06-02
Rule: Error out earlier when building alertmanager config

jessicalins:
Signed-off-by: Jéssica Lins <jessicaalins@gmail.com>

^ I haven't added this change to CHANGELOG - not sure if this is relevant - let me know if it is!

## Changes

This tries to solve issue https://github.com/thanos-io/thanos/issues/5186
In case an `alertmanagers.url` has a empty scheme or host, we now error out earlier. This helps preventing the case described in the issue and avoids Thanos Rule erroring out with `unsupported protocol scheme` when trying to post to an URL with empty scheme, for example.

## Verification

* I've added test cases for both empty scheme/empty host.
* Ran `make test-local`

### DIFF
--- a/pkg/alert/config.go
+++ b/pkg/alert/config.go
@@ -95,7 +95,15 @@ func BuildAlertmanagerConfig(address string, timeout time.Duration) (Alertmanage
 	}
 
 	scheme := parsed.Scheme
+	if scheme == "" {
+		return AlertmanagerConfig{}, errors.New("alertmanagers.url contains empty scheme")
+	}
+
 	host := parsed.Host
+	if host == "" {
+		return AlertmanagerConfig{}, errors.New("alertmanagers.url contains empty host")
+	}
+
 	for _, qType := range []dns.QType{dns.A, dns.SRV, dns.SRVNoA} {
 		prefix := string(qType) + "+"
 		if strings.HasPrefix(strings.ToLower(scheme), prefix) {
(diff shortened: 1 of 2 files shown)


saswatamcode (approved):
Thanks!

pkg/alert/config.go
```diff
@@ -96,6 +96,10 @@ func BuildAlertmanagerConfig(address string, timeout time.Duration) (Alertmanage
 
 	scheme := parsed.Scheme
 	host := parsed.Host
+	if scheme == "" || host == "" {
+		return AlertmanagerConfig{}, errors.New("alertmanagers.url contains empty scheme or host")
+	}
+
```
matej-g, line 102: Optional: We could separate this into separate conditions so that we can tell user exactly which part is missing.

matej-g (approved):
Nice! :+1:

pkg/alert/config_test.go
```diff
@@ -132,6 +132,10 @@ func TestBuildAlertmanagerConfiguration(t *testing.T) {
 			address: "://user:pass@localhost:9093",
 			err:     true,
 		},
+		{
+			address: "dnssrv+_http._tcp.example.com",
+			err:     true,
+		},
```
douglascamata, line 138: This test only cover the scheme check, right?

Shouldn't there be one that covers the host error too?

saswatamcode (approved):
LGTM!

matej-g (approved):
:medal_sports:

Issue #4972, opened 2021-12-18
Consider using true consistent hashring impl in Receive


bwplotka:
I just noticed that our receive hashring was never... hashring (: 

Our mod hash is quite naive and has problems to shard our data when changes occur. Every time new replica is added or removed series shards are jumping across replicas. IF we want more dynamic system we need to fix this.

cc @spaparaju @squat @jmichalek132  @onprem 

AC: 

* Use some hashring implementation (e.g https://github.com/serialx/hashring) to fix this.

Read more about the problem here: https://www.toptal.com/big-data/consistent-hashing

The key challenge is how to distribute hashring without key value store. I assume we can do it using our https://github.com/observatorium/thanos-receive-controller

PS: Once we fix this we could play with different thanos-receive-controller semantics: We could reshape hashring on every pod restart with little impact!

squat:
Just to for correctness: our hashring _is_ a hashring, however it's not a consistent hashring! I totally support this; it's one of the main improvements I suggested during last year's Thanos Receiver kubecon talk. This would be a big help to help us horizontally scale receiver sets without having metrics bounce around everywhere.

bwplotka:
Hm, I don't think module hash can be called hashing, but nevertheless, I think we are on the same page! (((:

bisakhmondal:
> PS: Once we fix this we could play with different thanos-receive-controller semantics: We could reshape hashring on every pod restart with little impact!

Hi guys! I can improve this module. I have recently developed an efficient infrastructure with hash rings.  Even there are several research papers on the same.

My basic approach would be 
1. Use `xxhash` (extremely fast) to encode the entities (server-id/name) for buckets (analogy of servers in a cluster) by pivoting them in a circle with the necessary replication factor and routing the balls (entities that we want to route ). The hashing implementation that we can leverage is https://ai.googleblog.com/2017/04/consistent-hashing-with-bounded-loads.html
> Our mod hash is quite naive and has problems to shard our data when changes occur. Every time new replica is added or removed series shards are jumping across replicas.

2. This type of load rebalancing takes a lot of memory. I know the right tool to mitigate the issue. introducing `jump consistent hashing`. https://arxiv.org/pdf/1406.2294.pdf

I am looking forward to working on this project for GSoC 2022. I am new in thanos : ) With the right guidance from you guys, I am sure that I can pull it off. 
What do you think about the small fragment of the proposal that I just proposed? Please do let me know. Thank you!

[...]

PR #5408, opened 2022-06-08
Consistent hashing

fpetkovski:
This commit adds support for distributing series in Receivers using
consistent hashing based on the libketama algorithm.

The following article provides a great summary of all available options: https://dgryski.medium.com/consistent-hashing-algorithmic-tradeoffs-ef6b8e2fcae8.

Fixes https://github.com/thanos-io/thanos/issues/4972

## Changes

Add support for consistent hashrings in Thanos Receiver.

## Verification

Unit tests.


fpetkovski:
For the sake of testing, I disabled flushing the TSDB when the hashring changes. Here's how head series look like now when a new receiver is added.

The receiver-3 instance came up around 11:16, but receivers 1 and 2 had no bump in head series.
One thing I noticed is that the sharding algorithm is not as balanced as hashmod, but this might not be a big problem in practice.

image

pkg/receive/hashring_test.go
```diff
@@ -148,3 +151,210 @@ func TestHashringGet(t *testing.T) {
...
+	}
+}
+
+func TestConsistentHashringConsistency(t *testing.T) {
+	series := makeSeries(10000)
+
+	ringA := []string{"node-1", "node-2", "node-3"}
+	a1, err := assignSeries(series, ringA)
+	if err != nil {
+		t.Fatal(err)
```
GiedriusS, line 246: Let's use `testutil.Ok()`

GiedriusS (approved):
LGTM, let's just use `testutil.*` helpers instead of `t.Fatalf` and so on

pkg/receive/hashring_test.go
```diff
@@ -148,3 +153,188 @@ func TestHashringGet(t *testing.T) {
...
+			expectedNode: "node-3",
+		},
+	}
+
+	for _, test := range tests {
+		t.Run(test.name, func(t *testing.T) {
+			hashRing := newConsistentHashring(test.nodes, 10)
+			result, err := hashRing.GetN("tenant", test.ts, test.n)
+			if err != nil {
+				t.Error(err)
```
GiedriusS, line 232: Still not using any helpers here ;P

bwplotka:
Amazing work, will check after my PTO (tomorrow+)!

Kind Regards,
Bartek Płotka ***@***.***)

pkg/receive/hashring_test.go
fpetkovski: Ah missed it :) Should be gone now

pkg/receive/hashring.go
```diff
@@ -6,14 +6,31 @@ package receive
...
 
 	"github.com/thanos-io/thanos/pkg/store/storepb/prompb"
 )
 
+// HashringAlgorithm is the algorithm used to distribute series in the ring.
+type HashringAlgorithm string
+
+const (
+	AlgorithmHashmod    HashringAlgorithm = "hashmod"
+	AlgorithmConsistent HashringAlgorithm = "consistent"
```
douglascamata, line 26: Should this be called "ketama" instead of "consistent"? This that other consistent hashing algorithms can be added in the future without breaking backwards compatibility.

fpetkovski: I agree, we should give it a more specific name. We can change it to either `ring-hash`, `ketama`, `dynamo` or something along those lines. I'm open to suggestions :)

douglascamata: `ketama` sounds like a good one to me. It's one algorithm that uses a ring-hash. There could be other algorithms using ring-hashes and then we would have a name conflict.

docs/components/receive.md
```diff
@@ -126,6 +126,9 @@ Flags:
                                  Alternative to 'receive.hashrings-file' flag
                                  (lower priority). Content of file that contains
                                  the hashring configuration.
+      --receive.hashrings-algorithm=hashmod
```
bwplotka, line 129: Kind of sad our tooling does not put those values in help. Do you mean explicitly adding those options for people to know how to use it?

pkg/receive/hashring.go
bwplotka, line 26: I need to agree with @douglascamata I think it would quite good to specify exactly what implementation we use. If it's based on libketama from https://dgryski.medium.com/consistent-hashing-algorithmic-tradeoffs-ef6b8e2fcae8 - let's mentioned that in comment, name of this `ketama`, and rename structs.

bwplotka (approved):
Looks amazing! Just readability comment, otherwise let's merge, and check the prod!

pkg/receive/hashring.go
fpetkovski: Sounds good 👍 Renamed to `ketama`

fpetkovski: 👍

fpetkovski:
Renamed `consistent` to `ketama`. 

I also wonder if what @bill3tt mentioned makes sense, and whether we want to even expose different options to users. One advantage of `hashmod` is that series distribution is still a bit more even, so it might take a bit more time for the consistent hashing algorithm to get there.

bwplotka:
I think that's a good point to think about new default in 1.0 or before. For now let's treat this option as "feature flag" we can experiment with (:

hack/config.yml
```diff
@@ -0,0 +1,35 @@
+global:
```
bwplotka, line 1: Should we remove `hack` dir? you can put `.gitignore` to not accidently commit those (:

bwplotka (commented):
Amazing, let's remove `hack` dir though (mistake?)

hack/config.yml
fpetkovski: Ah yes, removed :)

PR #5409, opened 2022-06-08
S3: Add config option to enforce the minio DNS lookup

Jakob3xD:
This MR allows to force DNS style lookup, as by default the [minio lib only does this for specific providers](https://github.com/minio/minio-go/blob/78350132c2c9a1d9a9a6815d80d57d42e3c40604/pkg/s3utils/utils.go#L71-L82). In my use case we have S3 running on Ceph with Proxys in front which only allows DNS style requests.

* [x] I added CHANGELOG entry for this change
* [ ] Change is not relevant to the end user

## Changes
1. Add S3 config option to force the DNS lookup
2. Updated docs with the new option

## Verification
I tested those changes on my Thanos Cluster with a custom S3 Ceph which only allows access via DNS Style requests.

### DIFF
--- a/docs/storage.md
+++ b/docs/storage.md
@@ -89,7 +89,7 @@ config:
   trace:
     enable: false
   list_objects_version: ""
-  dns_style: false
+  bucket_lookup_type: auto
   part_size: 67108864
   sse_config:
     type: ""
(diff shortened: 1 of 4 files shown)


docs/storage.md
```diff
@@ -89,6 +89,7 @@ config:
   trace:
     enable: false
   list_objects_version: ""
+  dns_style: false
```
bwplotka, line 92: This makes sense, just can we:

* Find a more specific name?
* Use enums instead of boolean - it will "scale" better if there will be a 3rd option.

For example:

```suggestion
  dns_style: "virtual-hosted" # Or `"path" read more about it here https://docs.aws.amazon.com/AmazonS3/latest/userguide/VirtualHosting.html
```

bwplotka (changes requested):
Small nit, otherwise LGTM (: Thanks!

docs/storage.md
Jakob3xD: As requested we changed the name and are using enums.
If

Jakob3xD: The CI/test are fixed now.

bwplotka (approved):
Thanks!

PR #5410, opened 2022-06-09
query: Close() after using query

GiedriusS:
This should reduce memory usage because Close() returns points back to a
sync.Pool.


Titasp (dismissed):

min-mik (dismissed):

niaurys (approved):

squat (approved):
very nice

PR #5411, opened 2022-06-09
Migrating Jaeger tracing from OpenTracing to OpenTelemetry; Add OTLP exporter

metonymic-smokey:
part of fixing #1972 

## Changes

This PR provides two changes that brings us closer to closing the gap between migrating from OpenTracing to OpenTelemetry, in particular:
a) This moves Jaeger client from OpenTracing to OpenTelemetry, _without breaking changes_. This means users can keep using their existing tracing configuration (some options will now not have any effect, but these are the ones that are obsolete after move to OTEL; read more in the updated documentation in the PR)
b) This adds the OTLP exporter, meaning now users will be able to leverage this protocol as well, with support of both gRPC and HTTP client.

## Verification

- Added E2E tests to validate Jaeger tracing
- Manual tests


metonymic-smokey:
@matej-g could you please review my current progress, when you're free.
1. I've created a function which returns an OTLP exporter but I'm not sure how to use the exporter.
2. Working on E2E testing.

Thanks!

pkg/tracing/jaeger/jaeger.go
```diff
@@ -33,6 +43,94 @@ func (t *Tracer) GetTraceIDFromSpanContext(ctx opentracing.SpanContext) (string,
...
+		}
+	}
+
+	processor := tracesdk.NewBatchSpanProcessor(exporter)
+	tp := newTraceProvider(ctx, logger, processor, config.SamplerParam, config.ServiceName)
+
+	return tp, nil
+}
+
+func newTraceProvider(ctx context.Context, logger log.Logger, processor tracesdk.SpanProcessor,
```
matej-g, line 92: This is what I guess will be a bit more complicate, but I think we will have to go over _all_ configuration options from the existing `jaeger.Config` struct and correlate them with the OTEL exporter configuration, to ensure a compatibility between old and new configs (to the max extent we can). So this will be quite a bit more complex than the Google Cloud tracer.

Some options map clearly (e.g. agent host-port, collector endpoint etc.), but some options need a bit more work. The current  version of the code, for example, assumes we always use ratio-based sampler (and so we're always creating `tracesdk.TraceIDRatioBased`), but we should first inspect `SamplerType` and choose the appropriate sampler type based on that.

Then again, we might find not all options map to the new OTEL config (for example, IIRC, the `RPCMetrics` does not have an equivalent in the OTEL config).

I believe most of the core options can be mapped to the new config, but even with the remaining ones we should be careful and if we cannot find an option equivalent in the OTEL config, we need to state this in the changelog and warn users.

But for now I'd just focus on mapping the ones we can (looks like `Tags`, `Sampler*` and `Reporter*` _could_ be possible).

pkg/tracing/client/factory.go
```diff
@@ -30,13 +39,38 @@ const (
 	Jaeger      TracingProvider = "JAEGER"
 	ElasticAPM  TracingProvider = "ELASTIC_APM"
 	Lightstep   TracingProvider = "LIGHTSTEP"
+	name                        = "thanos"
```
matej-g, line 42: Leftover?

pkg/tracing/jaeger/jaeger.go
```diff
@@ -33,6 +43,94 @@ func (t *Tracer) GetTraceIDFromSpanContext(ctx opentracing.SpanContext) (string,
...
+				migration.ForceTracingAttributeKey,
+			),
+		),
+		tracesdk.WithResource(resource),
+	)
+
+	return tp
+}
+
+func collectAttributes(serviceName string) []attribute.KeyValue {
```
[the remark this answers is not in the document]
metonymic-smokey: Thanks, moved to the `tracing` package.

pkg/tracing/jaeger/jaeger.go
```diff
@@ -33,6 +43,94 @@ func (t *Tracer) GetTraceIDFromSpanContext(ctx opentracing.SpanContext) (string,
...
+
+		exporter, err = otel_jaeger.New(otel_jaeger.WithAgentEndpoint(jaegerAgentEndpointOptions...))
+		if err != nil {
+			return nil, err
+		}
+	} else {
+		exporter, err = otel_jaeger.New(nil)
+		if err != nil {
+			return nil, err
+		}
```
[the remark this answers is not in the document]
metonymic-smokey: I added it since if neither endpoint nor agent is configured, it will rely on the default options for [collector endpoint](https://pkg.go.dev/go.opentelemetry.io/otel/exporters/jaeger#WithCollectorEndpoint) and [agent](https://pkg.go.dev/go.opentelemetry.io/otel/exporters/jaeger#WithAgentEndpoint).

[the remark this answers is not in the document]
matej-g: I see now, maybe we could simplify this into a separate method which would looked at endpoint, agent host / port and returned one `EndpointOption` to use with `New` method?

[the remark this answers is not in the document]
metonymic-smokey: good idea, I'll do that for packages jaeger and OTLP. 
Thanks!

pkg/tracing/jaeger/jaeger.go
metonymic-smokey: Thanks, I have attempted to factor in some more options from `config_yaml.go`. I _think_ some options such as `SamplerManagerHostPort` and `ReporterFlushInterval` don't have an equivalent in the OTel Go SDK. I'll keep looking, just in case there is an equivalent for these options.

pkg/tracing/client/factory.go
```diff
@@ -30,13 +39,38 @@ const (
...
 )
 
 type TracingConfig struct {
 	Type   TracingProvider `yaml:"type"`
 	Config interface{}     `yaml:"config"`
 }
 
+// NewOTELTracer returns an OTLP exporter based tracer.
+func NewOTELTracer(ctx context.Context, logger log.Logger) trace.Tracer {
```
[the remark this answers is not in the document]
metonymic-smokey: I have created a new package `otlp` in a recent commit with some options for configuration.  I will try to add some more config options if possible. 
Please let me know what you think.

test/e2e/tracing_test.go
```diff
@@ -0,0 +1,119 @@
...
+	"github.com/thanos-io/thanos/test/e2e/e2ethanos"
+
+	"github.com/efficientgo/e2e"
+	"github.com/efficientgo/tools/core/pkg/backoff"
+	"github.com/efficientgo/tools/core/pkg/testutil"
+	"gopkg.in/yaml.v2"
+)
+
+// Test to check if the trace provider works as expected.
+func TestJaegerTracing(t *testing.T) {
```
metonymic-smokey, line 27: @matej-g this is a draft of the E2E test up till now, still a WIP. 
Could you please review it up till here and let me know?
thanks!

test/e2e/tracing_test.go
```diff
@@ -0,0 +1,119 @@
...
+		if response.StatusCode != http.StatusOK {
+			t.Logf("\n\n WAITING")
+			b.Wait()
+			continue
+		}
+
+		// We got a 200 response.
+		defer response.Body.Close()
+
+		body, err := ioutil.ReadAll(response.Body)
```
matej-g, line 113: We could potentially also assert on if response has expected content (e.g. is trace ID present)

metonymic-smokey: That's where the issue comes in.
For some reason, I keep getting empty response when I run this test locally.

pkg/tracing/jaeger/jaeger.go
metonymic-smokey: Regarding the config options in `jaeger/config_yaml.go`, the config struct has become rather large and unwieldy with all the new options added. I was considering creating new structs for `AgentOptions` and `CollectorOptions` and referencing those in the `Config` struct for a more structured config YAML and readable code. But since existing users might have their configs structured the current way, it might involve some re-work for them.
What do you think?

test/e2e/tracing_test.go
matej-g: I'm seeing some panic in the CI logs, it's probably related:
```
12:30:43 querier-1: panic: runtime error: invalid memory address or nil pointer dereference
12:30:43 querier-1: [signal SIGSEGV: segmentation violation code=0x1 addr=0x18 pc=0x1647a96]
12:30:43 querier-1: goroutine 1 [running]:
12:30:43 querier-1: go.opentelemetry.io/otel/exporters/jaeger.New({0x0, 0x0})
12:30:43 querier-1: /home/runner/go/pkg/mod/go.opentelemetry.io/otel/exporters/jaeger@v1.7.0/jaeger.go:47 +0x36
12:30:43 querier-1: github.com/thanos-io/thanos/pkg/tracing/jaeger.NewTracerProvider({0x21ea6c0, 0xc000054018}, {0x21ac500, 0xc000098190}, {0xc0000c9000, 0x3d, 0x40})
12:30:43 querier-1: /home/runner/work/thanos/thanos/pkg/tracing/jaeger/jaeger.go:62 +0x125
12:30:43 querier-1: github.com/thanos-io/thanos/pkg/tracing/client.NewTracer({0x21ea6c0, 0xc000054018}, {0x21ac500, 0xc000098190}, 0x7ffc7fbedd6f, {0xc0000da9c0, 0x57, 0x60})
12:30:43 querier-1: /home/runner/work/thanos/thanos/pkg/tracing/client/factory.go:68 +0x2e6
12:30:43 querier-1: main.main()
12:30:43 querier-1: /home/runner/work/thanos/thanos/cmd/thanos/main.go:102 +0xd71
```

metonymic-smokey: Thanks!
The CI test passes now but my E2E tests are failing.

metonymic-smokey:
@matej-g I have added unit tests for Jaeger in a recent commit, which are mostly similar to the ones for GCloud, other than the test for `parseTags`. 
PTAL and let me know if they're okay, I'll add similar ones for OTLP too then.
thanks!

pkg/tracing/jaeger/jaeger.go
```diff
@@ -33,6 +41,126 @@ func (t *Tracer) GetTraceIDFromSpanContext(ctx opentracing.SpanContext) (string,
...
+	resource, err := resource.New(ctx, resource.WithAttributes(attributes...))
+	if err != nil {
+		level.Warn(logger).Log("msg", "jaeger: detecting resources for tracing provider failed", "err", err)
+	}
+
+	tp := tracesdk.NewTracerProvider(
+		tracesdk.WithSpanProcessor(processor),
+		tracesdk.WithSampler(
+			migration.SamplerWithOverride(
+				tracesdk.ParentBased(tracesdk.TraceIDRatioBased(samplingFraction)),
```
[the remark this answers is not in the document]
metonymic-smokey: Thanks!
Added some sampler types in a recent commit.

pkg/tracing/jaeger/config_yaml.go
```diff
@@ -4,165 +4,140 @@
...
-		u, err := url.ParseRequestURI(cfg.Endpoint)
-		if err != nil {
-			return nil, errors.Wrapf(err, "cannot parse endpoint=%s", cfg.Endpoint)
-		}
-		rc.CollectorEndpoint = u.String()
-		user := cfg.User
-		pswd := cfg.Password
-		if user != "" && pswd == "" || user == "" && pswd != "" {
-			return nil, errors.Errorf("you must set %s and %s parameters together", cfg.User, cfg.Password)
+func getSampler(samplerType string, samplingFraction float64, parentConfig ParentBasedSamplerConfig) tracesdk.Sampler {
```
metonymic-smokey, line 105: @matej-g I have added the considerations for sampler type here.
I'm not sure how to add rate limiting samplers yet.
PTAL and let me know if this is okay, thanks!

test/e2e/e2ethanos/services.go
```diff
@@ -166,6 +166,36 @@ func NewPrometheusWithSidecarCustomImage(e e2e.Environment, name, promConfig, we
 	return prom, sidecar
 }
 
+func NewPrometheusWithJaegerTracingSidecarCustomImage(e e2e.Environment, name, promConfig, webConfig,
```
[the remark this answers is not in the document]
metonymic-smokey: I have changed this to the original but the E2E tests still fail and I don't see a panic/crash in the logs, so I _think_ it could be something unrelated to my changes?

metonymic-smokey:
@matej-g quick update:
1. Added remote sampler for Jaeger. 
2. Since I haven't found an alternative to the rate limiting strategy workaround we discussed earlier, I'll work on adding a remote sampler with rate limiting strategy.
3. Made some small changes to the unit tests and sampler selection in Jaeger.
4. Re E2E tests, the automated tests on this PR still fail despite not modifying the failing test. The error faced is `No such network: e2e_test_query`, which I _think_ seems to be unrelated to my changes. Could it be due to a flaky E2E test?

PTAL and let me know how I could work on these, thanks!

pkg/tracing/jaeger/config_yaml.go
```diff
@@ -4,165 +4,167 @@
...
-	ReporterLogSpans       bool          `yaml:"reporter_log_spans"`
-	Endpoint               string        `yaml:"endpoint"`
-	User                   string        `yaml:"user"`
-	Password               string        `yaml:"password"`
-	AgentHost              string        `yaml:"agent_host"`
-	AgentPort              int           `yaml:"agent_port"`
-	Gen128Bit              bool          `yaml:"traceid_128bit"`
+	ServiceName                        string                   `yaml:"service_name"`
+	Disabled                           bool                     `yaml:"disabled"`
+	RPCMetrics                         bool                     `yaml:"rpc_metrics"`
```
matej-g, line 30: As we spoke, do you think it would make sense to print a deprecation warning for options that are not applicable to new tracer?

pkg/tracing/otlp/otlp.go
```diff
@@ -0,0 +1,70 @@
...
+
+	tp := newTraceProvider(exporter)
+
+	return tp, nil
+}
+
+func newTraceProvider(exporter *otlptrace.Exporter) *tracesdk.TracerProvider {
+	tp := tracesdk.NewTracerProvider(
+		tracesdk.WithBatcher(exporter),
+		tracesdk.WithResource(resource.NewWithAttributes(
```
matej-g, line 65: Should we also use `resource.New` here to obtain the resource to detect some resources automatically? Same as it is done in other exporters. For reference: https://github.com/open-telemetry/opentelemetry-collector-contrib/blob/main/processor/resourcedetectionprocessor/README.md

pkg/tracing/jaeger/jaeger.go
```diff
@@ -6,18 +6,19 @@ package jaeger
...
-	"github.com/uber/jaeger-client-go/config"
-	jaeger_prometheus "github.com/uber/jaeger-lib/metrics/prometheus"
+	"go.opentelemetry.io/otel/attribute"
+	otel_jaeger "go.opentelemetry.io/otel/exporters/jaeger"
+	"go.opentelemetry.io/otel/sdk/resource"
+	tracesdk "go.opentelemetry.io/otel/sdk/trace"
+	"gopkg.in/yaml.v2"
 )
 
 // Tracer extends opentracing.Tracer.
```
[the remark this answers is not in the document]
metonymic-smokey: So, a quick look tells me that the `GetTraceIDFromSpanContext` method is used [here](https://github.com/thanos-io/thanos/blob/45943b84ee77f9f9471caf4c4bb284477f8a4589/pkg/tracing/http.go#L52). So I don't think I can remove it yet.

pkg/tracing/jaeger/config_yaml.go
```diff
@@ -164,6 +164,30 @@ func getRemoteOptions(config Config) []jaegerremote.Option {
 	return remoteOptions
 }
 
+func getRateLimitingOptions(config Config) []jaegerremote.Option {
```
metonymic-smokey, line 167: @matej-g this is the first draft of rate limiting sampler options.
This assumes the strategy will be parsed from the endpoint, which is polled regularly.

Regarding the maximum number of traces per second, I was considering adding another option to the config or reading a JSON file, and then posting that in JSON format to the URL. This way, the user can add that too to one config, instead of having a separate config on the URL.
Please let me know what you think, not entirely sure when it comes to remote and rate limited samplers 😅 

Final leg of the PR, we got this 💪🏻 

Thanks!

docs/tracing.md
```diff
@@ -86,14 +86,22 @@ config:
   sampler_manager_host_port: ""
   sampler_max_operations: 0
   sampler_refresh_interval: 0s
+  sampler_parent_config:
```
matej-g, line 89: I think it would be nice to preface these options with some introductory texts, e.g. what are the defaults used when user does not specify certain options? Maybe we could provide link to OTEL documentation.

pkg/tracing/jaeger/config_yaml.go
```diff
@@ -4,165 +4,204 @@
...
-			host = cfg.AgentHost
-		}
-
-		port := jaeger.DefaultUDPSpanServerPort
-		if cfg.AgentPort != 0 {
-			port = cfg.AgentPort
-		}
-		rc.LocalAgentHostPort = net.JoinHostPort(host, strconv.Itoa(port))
+	// InitialSamplingRate is the sampling probability when the backend is unreachable.
+	if config.InitialSamplingRate != 0.0 {
```
matej-g, line 185: Aren't these options relevant for remote sampler as well?

pkg/tracing/jaeger/config_yaml.go
matej-g: Thanks Aditi! Looking good with remote sampler, but I have a feeling we had misunderstood each other regarding rate limiting sampler.

rate limiting sampler =/= remote sampler

Rate limiting sampler should not need to contact a remote endpoint, it should work based on a configuration provided to Thanos. Example from existing version Thanos:
```
  "sampler_param": 2
  "sampler_type": "ratelimiting"
  "service_name": "thanos-receive"
```

This will allow only 2 traces per second. So this sampler is "static" and is not controlled by remote endpoint. I probably did not explain this correctly during last F2F - that's why we need to find a way to introduce this sampler, because I believe this sampler type _by itself_ might not yet exist in OTEL. Basically we would need this sampler https://github.com/open-telemetry/opentelemetry-go-contrib/blob/main/samplers/jaegerremote/sampler.go#L115 as a stand-alone sampler.

metonymic-smokey: @matej-g thanks!
I have added an implementation of rate limiting sampler, based on related OTel-go code. I had to introduce quite a few of the functions required since rate-limiting samplers related utilities are `internal` to OTel-go.
I _think_ this is the only way that strikes me now to introduce rate limiting samplers. It might be a temporary, stop-gap arrangement until the required utilities are exported in future by OTel-go. 
Please let me know what you think.
thanks!

docs/tracing.md
```diff
@@ -114,6 +114,8 @@ You will also need to ensure that the authentication with the API is working, fo
 *Note:* The `type` in the configuration below can have either value `GOOGLE_CLOUD` or `STACKDRIVER` - this is to ensure backwards compatibility.
 
 ```yaml mdox-exec="go run scripts/cfggen/main.go --name=google_cloud.Config"
+# github.com/shirou/gopsutil/v3/cpu
```
metonymic-smokey, line 117: Accidentally committed; will fix.

ahayworth:
👋 I wanted to ask if context propagation has been considered in this PR at all? Based on my reading of this PR in conjunction with the opentracing bridge already in the codebase, it seems like we might be implicitly forcing a switch to opentracing "basic" propagation? Notably, that's not the behavior of the current tracing functions, so that means there's a risk of breaking tracing setups as-is today. Specifically, I can assert that the jaeger tracing functionality already present in Thanos today does _not_ use opentracing "basic" propagation (as implemented by [otel-go's opentracing propagator](https://github.com/open-telemetry/opentelemetry-go-contrib/blob/main/propagators/ot/ot_propagator.go#L32)); rather, it seems to use jaeger's native propagation.

(Notably: OpenTracing never really forced a specific standard at all for context propagation - the "opentracing" propagation wasn't widely adopted among "opentracing" tracers. See context [here](https://github.com/open-telemetry/opentelemetry-specification/blob/1ef7871c02170f07ce6e8bf3ec16c1d88cdb65a4/specification/context/api-propagators.md#propagators-distribution) in the OTel spec)

Related: if we used the new [`autoprop` package](https://github.com/open-telemetry/opentelemetry-go-contrib/pull/2258) as part of this PR, it would make it easy to extend support for multiple context propagation methods simultaneously - and even give that control to the user via environmental variables that are part of the OpenTelemetry spec. That may be an easy way to handle the different formats that are needed to maintain compatibility.

(Unless we desire a breaking change in this PR, of course! That's definitely valid!)

metonymic-smokey:
thanks @ahayworth, I shall take a look at the package.
cc @matej-g what do you think?

matej-g:
Thanks for the detailed explanation @ahayworth, yes to me it totally makes sense to use the `autoprop` package instead to support multiple formats :+1:

pkg/tracing/jaeger/jaeger.go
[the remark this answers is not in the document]
matej-g: But unless I miss something, this tracer won't be used anymore. We should actually get into the `else` branch of the code you linked in here: https://github.com/thanos-io/thanos/blob/45943b84ee77f9f9471caf4c4bb284477f8a4589/pkg/tracing/http.go#L57

matej-g (commented):
Awesome job on the remote sampler! :raised_hands: 
I added comments for few outstanding things + we could use the `autoprop` package as suggested in the comment by @ahayworth , afterwards this is looking good to go :heavy_check_mark:

docs/tracing.md
```diff
@@ -72,7 +72,7 @@ Currently supported tracing backends:
 
 ### Jaeger
 
-Client for https://github.com/jaegertracing/jaeger tracing.
+Client for https://github.com/jaegertracing/jaeger tracing. Parent config options are used with ParentBased samplers. This is a [link](https://github.com/open-telemetry/opentelemetry-specification/blob/main/specification/trace/sdk.md#parentbased) to the OTel specification which details the defaults.
```
matej-g, line 75: We should also add docs for OTLP tracing

metonymic-smokey:
Thank you so much @matej-g for pulling this to the finish line!
The recently added changes look good to me!

pkg/tracing/otlp/config_yaml.go
```diff
@@ -0,0 +1,143 @@
...
+	}
+
+	if config.RetryConfig.RetryEnabled {
+		options = append(options, otlptracehttp.WithRetry(createHTTPRetryConfig(config)))
+	}
+
+	if config.Headers != nil {
+		options = append(options, otlptracehttp.WithHeaders(config.Headers))
+	}
+	// how to specify JSON/binary format here?
```
GiedriusS, line 104: We can remove this? What does this mean?

pkg/tracing/otlp/otlp.go
```diff
@@ -0,0 +1,83 @@
...
+// NewOTELTracer returns an OTLP exporter based tracer.
+func NewTracerProvider(ctx context.Context, logger log.Logger, conf []byte) (*tracesdk.TracerProvider, error) {
+	config := Config{}
+	if err := yaml.Unmarshal(conf, &config); err != nil {
+		return nil, err
+	}
+
+	var exporter *otlptrace.Exporter
+	var err error
+	switch config.ClientType {
```
GiedriusS, line 37: Always lower-case this or upper-case this to make it indifferent to the case of the letters?

test/e2e/tracing_test.go
```diff
@@ -0,0 +1,119 @@
...
+		})
+	testutil.Ok(t, e2e.StartAndWaitReady(newJaeger))
+
+	jaegerConfig, err := yaml.Marshal(client.TracingConfig{
+		Type: client.Jaeger,
+		Config: jaeger.Config{
+			ServiceName:  "thanos-sidecar",
+			SamplerType:  "const",
+			SamplerParam: 1,
+			Endpoint:     "http://" + newJaeger.InternalEndpoint("jaeger.thrift") + "/api/traces", // make this a var
```
GiedriusS, line 54: Unfinished? Or either remove this comment :smile:

GiedriusS (commented):
A bunch of nits that aren't blockers. Is my understanding correct that OTLP doesn't support client-side sampling? Or is it a limitation of the proposed implementation?

pkg/tracing/otlp/config_yaml.go
ahayworth: OTLP can be carried over GRPC/Proto, HTTP/Proto, or HTTP/JSON (with JSON still being "experimental").

ahayworth:
> Is my understanding correct that OTLP doesn't support client-side sampling? Or is it a limitation of the proposed implementation?

The OpenTelemetry SDK spec [talks about sampling support for client SDKs.]
https://github.com/open-telemetry/opentelemetry-specification/blob/main/specification/trace/sdk.md#sampling) The [go SDK seems to support it as well](https://github.com/open-telemetry/opentelemetry-go/blob/main/sdk/trace/sampler_env.go).

It looks like the PR is [forcing 100% sampling](https://github.com/thanos-io/thanos/pull/5411/files?show-deleted-files=true&show-viewed-files=true&file-filters%5B%5D=#diff-544c3b568c49d72e6103c46a8d667ffba44f8a3edd831dbc1cb9e9c4c945175cR71) ? However, by default the go SDK will (eventually) [pull sampler configs from the env](https://github.com/open-telemetry/opentelemetry-go/blob/8c6e6c4bac681079ef8a17e9ff4a775cb1eea5bb/sdk/trace/provider.go#L104) if not otherwise dissuaded.

GiedriusS (dismissed):
:muscle: thank you! Just needs to be updated to work with the new e2e test framework version

metonymic-smokey:
@GiedriusS @matej-g i've updated the E2E tests and looks like they all pass :)
Thank you so much for working with me on this one!

GiedriusS (approved):
shipping time

PR #5415, opened 2022-06-12
Submit a proposal for Active Series Limiting for Hashring Topology

saswatamcode:
The PR submits a proposal for Active Series Limiting for Hashring Topology in Thanos.
 
Looking for feedback and comments! 🙂 

Deploy preview here: https://deploy-preview-5415--thanos-io.netlify.app/tip/proposals-accepted/202206-active-series-limiting-hashring.md/

## Changes

N/A

## Verification

N/A


docs/proposals-accepted/202206-active-series-limiting-hashring.md
```diff
@@ -0,0 +1,206 @@
...
+* Non-trivial increase in latency
+* Difficult to scale up/down
+* Adds more complexity to how Receivers in hashring communicate with each other
+
+## Alternatives
+
+There are a few alternative to what is proposed above,
+
+* An alternative could be just not to limit active series globally and make do with local limits only.
+* [Consistent hashing](https://github.com/thanos-io/thanos/issues/4972) might be implemented and problems with sharding can be sorted out, which would make adding Receive replicas to hashring a non-disruptive operation, so that solutions like HPA can be used and make scale up/down operations much easier to the point where limits are not needed.
```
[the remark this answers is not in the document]
saswatamcode: Yes! Let me add! 🙂

matej-g:
My 2 cents :coin::
We talked about this F2F with some folks including @saswatamcode, so putting it briefly here as well. From conceptual standpoint, I'm wondering if doing this directly in Thanos is the right choice. Thanos, as I look at it, has no 'real' concept of tenancy and the receive component knows only enough about tenancy to make storing / retrieving metrics possible. We could then start asking questions about other tenancy features not being within scope of Thanos (authN, authZ for example), when in fact Thanos is (for me) philosophically geared towards being simple and primarily single-tenanted solution (maybe that's up for discussion as well?). I don't have exact answers on how technically this can be done without internal knowledge of receiver data, but I can imagine it could be based around digging into remote write requests before they hit the receive endpoint (e.g. some middle-layer / proxy that would keep count of series based on what's in the request). 

I'm not strictly against doing this in Thanos, if we come to the conclusion this is the most suitable place to do this due to limiting factors (i.e. maybe we really do need internal, 'unexported' knowledge of what's in Receiver to make per-tenant limiting feasible), but I'd still like to see this alternative mentioned and at least explored in the proposal.

saswatamcode:
Yes @matej-g! I was a bit confused as to whether this could be listed as an alternative within the proposal, so let me add it here!

douglascamata:
@matej-g @saswatamcode the so called "Receive Validator" mode could be developed as a new alpha/experimental component in the Thanos project for a while. This way if the community decides Thanos is not a good place for the component to live, we could easily move it out of tree and into another umbrella. What do you think?

saswatamcode:
@douglascamata, just my 2c here but wouldn't that be going against the [proposal process](https://thanos.io/tip/contributing/proposal-process.md/#proposals-process) in a way?

This idea is made into a proposal that the community can provide feedback and validate before anything is changed or implemented, so that we don't have to go through the process of removing it, later on, AFAIU. 

Also, if it is decided that this would not be implemented in Thanos, then this proposal would be a record of those decisions and can explain why when a future discussion comes up related to this. 🙂

douglascamata:
My idea to make it a separate experimental component gives the Thanos' community time to decide on embracing (or not) the feature-set in this proposal without blocking its development. Seeing how this could work and its implications in the system can provide valuable insights to the decision

Of course it could also be started out of Thanos and later moved into it if the community accepts so and you are interested in this approach.

douglascamata:
Overall, I don't see anything wrong with proposing "an experiment". But I'll let maintainers decide on it. 😄

matej-g:
IMHO an experiment (a proof of concept) would make sense here, but I don't know if it makes to develop something beyond that _in the context of this proposal_. I would see it meaningful to decide based on such proof of concept whether to take that route within Thanos project or choose a different alternative. Of course, nothing from this prevents creating such experiment (and potentially develop it further) outside of Thanos.

saswatamcode:
Yup, we can totally have draft PRs with e2e tests to showcase some of the approaches!

docs/proposals-accepted/202206-active-series-limiting-hashring.md
```diff
@@ -0,0 +1,207 @@
...
+
+We could leverage a meta-monitoring solution which scrapes metrics from all Receivers as and then programmatically query for metrics like `prometheus_tsdb_head_series` and sum across tenants and instances periodically and limit based on that value (`latestCurrentSeries`).
+
+This approach would add another mode to Receive apart from Router, Ingestor & RouterIngestor, called **“Validator”**.
+
+Within Validator, we do not need to calculate an increase based on requests, as this will be handled by Receive instrumentation and meta-monitoring solution. We only need to query the latest HEAD series value for a tenant summed across all receives and limit remote write requests if the result of the instant query is greater than the configured limit.
+
+This value can also be cached, and the query for latest value can be executed periodically.
+
+So if a user configures a *per-tenant* limit, say `globalSeriesLimit`, the resultant limiting equation here would simply be `globalSeriesLimit >= latestCurrentSeries` which is checked on request.
```
bill3tt, line 80: If the tenant exceeds the `globalSeriesLimit`, and then sends a remote write request that does not create any new series - would we accept or reject this request?

docs/proposals-accepted/202206-active-series-limiting-hashring.md
```diff
@@ -0,0 +1,207 @@
...
+* In case, external meta-monitoring solution is down, can fall back to per-replica-tenant limits
+
+#### Cons:
+* Not very accurate
+  * We do not know exact state of each TSDB, only know view of meta-monitoring solution, which gets updated on every scrape
+  * We do not account for how much a remote write request will increase the number of active series, only infer that from query result after the fact
+* Data replication (quorum-based) will likely cause inaccuracies in HEAD stat metrics
+* Dependence on external system, so best effort availability
+
+### Receive Validator
```
bill3tt, line 100: Rather than a brand new mode within Thanos Receive (which would take us to 4 and it is confusing enough already!), another option would be to make this functionality only present within the 'router' mode.

docs/proposals-accepted/202206-active-series-limiting-hashring.md
```diff
@@ -0,0 +1,207 @@
...
+
+Furthermore, we also have each tenant’s [Appendable](https://pkg.go.dev/github.com/thanos-io/thanos/pkg/receive#Appendable) in multitsdb, which returns a Prometheus [storage.Appender](https://pkg.go.dev/github.com/prometheus/prometheus/storage#Appender), which can in turn give us a [storage.GetRef](https://pkg.go.dev/github.com/prometheus/prometheus/storage#GetRef.GetRef) interface. This helps us know if a TSDB has a cached reference for a particular set of labels in its HEAD.
+
+This GetRef interface returns a [SeriesRef](https://pkg.go.dev/github.com/prometheus/prometheus/storage#SeriesRef) when a set of labels is passed to it. If the SeriesRef is 0, it means that that set of labels is not cached, and any sample with that set of labels will generate a new active series. This data can also be fetched from a new endpoint like `api/v1/getrefmap` and merged across replicas.
+
+This approach would also add another mode to Receive apart from Router, Ingestor & RouterIngestor, called **“Validator”** which could actually be a Receive instance in front of Receive hashring. This is where we can get data from hashring Receivers and merge them to limit remote write requests.
```
bill3tt, line 112: The receive router split proposal makes it possible to compose arbitrary trees of receivers - so we could leverage that instead of creating a new mode entirely.

bill3tt:
This is a super well-written proposal @saswatamcode on a particularly tricky topic :hugs: 

Reading through your lists of pros & cons, I think we can refine our requirements slightly, which may help us land on a solution:
* **We do not need to be 100% accurate with this series limiting.** The goal here is to prevent unbounded data writing, but being 10% over a limit probably isn't a _massive_ problem.
* **We can tolerate a increase in write latency.** The key attribute is the availability of the write path, rather than the latency of the write path. Obviously a 500% increase in write latency is not ok, but a ~20% or so increase in order to implement solid limiting capability is likely a perfectly acceptable trade-off.

Solutions that require network responses from all replicas before making a yes/no decision will struggle when part of the hash-ring is unavailable. Currently, remote write requests are deemed accepted when a quorum of their receivers return 200.

docs/proposals-accepted/202206-active-series-limiting-hashring.md
saswatamcode: In the meta-monitoring approach, we don't check how many new series a remote write request creates. We only check the number of currently active series. So if a tenant has exceeded the limit, all of their write requests would be rejected. :)

docs/proposals-accepted/202206-active-series-limiting-hashring.md
saswatamcode: When I mention Receive Validator in this proposal, it is essentially Receive Router but with limiting logic. Would it still be confusing in that case?

douglascamata: In general, I think the situation is getting complex with Thanos Receiver: ingester, router, ingester+router, router+validator. 🌀

docs/proposals-accepted/202206-active-series-limiting-hashring.md
```diff
@@ -0,0 +1,207 @@
...
+## Related links/tickets
+
+* https://github.com/thanos-io/thanos/pull/5333
+* https://github.com/thanos-io/thanos/pull/5402
+* https://github.com/thanos-io/thanos/issues/5404
+* https://github.com/thanos-io/thanos/issues/4972
+
+## Why
+
+Thanos is built to be a set of components that can be composed into a highly available metrics system with unlimited storage capacity. But for achieving true HA we need to ensure that tenants in our write path, cannot push too much data and cause issues. There need to be limits in place so that our ingestion systems maintain the level of QoS and only block the tenant that exceeds limits.
```
bwplotka, line 18:
```suggestion
Thanos is built to be a set of components that can be composed into a highly available metrics system with unlimited storage capacity. But for achieving true HA we need to ensure that tenants in our write path, cannot push too much data and cause issues. There need to be limits in place so that our ingestion systems maintain the level of  Quality of Service (QoS) and only block the tenant that exceeds limits.
```

Sorry, my book writing intuition comes in 🙈

docs/proposals-accepted/202206-active-series-limiting-hashring.md
```diff
@@ -0,0 +1,207 @@
...
+
+Thus, any remote write request can be failed completely, with a 429 status code and appropriate error message, if it increases the number of active series above the configured limit for a tenant. Partially accepting write requests might lead to confusing results and error semantics, so we propose to avoid this and focus on retries from client-side.
+
+There are however a few challenges to this, as tenant metric data is distributed and replicated across multiple Thanos Receive replicas. Also, with a multi-replica setup, we have the concept of per-replica-tenant and per-tenant limits that can be defined as,
+
+* **Per-replica-tenant limit**: The limit imposed for active series per tenant, per replica of Thanos Receive. An initial implementation of this is already WIP in this [PR](https://github.com/thanos-io/thanos/pull/5333). This can also be treated as the active series limit for non-hashring topology or single replica Receive setups.
+
+* **Per-tenant limit**: The overall limit imposed for active series per tenant across all replicas of Thanos Receive. This is essentially what this proposal is for.
+
+## Proposal
```
bwplotka, line 60: Can we have single proposal (just choose one you think makes sense the most) and put rest in Alternatives please. We can then copy over stuff to swap chosen idea if needed

docs/proposals-accepted/202206-active-series-limiting-hashring.md
```diff
@@ -0,0 +1,207 @@
...
+Within Validator, we do not need to calculate an increase based on requests, as this will be handled by Receive instrumentation and meta-monitoring solution. We only need to query the latest HEAD series value for a tenant summed across all receives and limit remote write requests if the result of the instant query is greater than the configured limit.
+
+This value can also be cached, and the query for latest value can be executed periodically.
+
+So if a user configures a *per-tenant* limit, say `globalSeriesLimit`, the resultant limiting equation here would simply be `globalSeriesLimit >= latestCurrentSeries` which is checked on request.
+
+<img src="../img/meta-monitoring-validator.png" alt="Meta-monitoring-based Validator" width="800"/>
+
+#### Pros:
+* Simpler as compared to other solutions and easier to implement
```
bwplotka, line 85:
```suggestion
#### Pros:

* Simpler as compared to other solutions and easier to implement
```

docs/proposals-accepted/202206-active-series-limiting-hashring.md
```diff
@@ -0,0 +1,207 @@
...
+* Lesser endpoint calls, so improved latency
+* Relies on "external to Thanos" system, and doesn’t increase load on Receive
+* Does not add much tenancy-based complexity to Thanos
+* No need to merge statistics across replicas, handled by meta-monitoring
+* Additional request-based rate limiting can be done within same component
+* In case, external meta-monitoring solution is down, can fall back to per-replica-tenant limits
```
bwplotka, line 91:
```suggestion
* In case, external meta-monitoring solution is down, can fall back to per-replica-tenant limits
* Growing our instrumentation to improve validator, improves our observability as well
```

docs/proposals-accepted/202206-active-series-limiting-hashring.md
bwplotka: I think we could mention it's not Receive Validator but rather "Adding validation" to receive routers?

bwplotka (commented):
Great job! Added some comments 💪🏽 Let's goooo, we need this!

docs/proposals-accepted/202206-active-series-limiting-hashring.md
saswatamcode: Yes, was planning to do just that! 🙂

docs/proposals-accepted/202206-active-series-limiting-hashring.md
```diff
@@ -0,0 +1,226 @@
...
+
+There are a few alternative to what is proposed above,
+
+### Receive Router Validation
+
+We can implement some new endpoints on Thanos Receive.
+
+Firstly, we can take advantage of the `api/v1/status/tsdb` endpoint that is exposed by [Prometheus TSDB](https://prometheus.io/docs/prometheus/latest/querying/api/#tsdb-stats) but is yet to be implemented in Thanos Receive ([in-review PR](https://github.com/thanos-io/thanos/pull/5402) which utilizes tenant headers to get local tenant TSDB stats in Receive).
+
+In its current WIP implementation, it can provide us stats for each local TSDB of a tenant which contains a measure of HEAD series (so active series). We can merge this to get the total number of HEAD or active series a tenant has.
```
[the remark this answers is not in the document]
saswatamcode: Yes! Will update!

saswatamcode:
I've created #5520 for the meta-monitoring approach, will update this proposal shortly. 🙂

docs/proposals-accepted/202206-active-series-limiting-hashring.md
```diff
@@ -0,0 +1,226 @@
...
+
+* The current count of active series for a tenant (across all replicas if it is a *per-tenant* limit)
+* The user configured limit (*per-tenant* and *per-replica-tenant* can be different). We can assume this would be available as a user flag and would be same for all tenants (initially)
+* The increase in the number of active series, when a new tenant [remote write request](https://github.com/prometheus/prometheus/blob/v2.36.1/prompb/remote.proto#L22) would be ingested (this can be optional as seen in [meta-monitoring approach](#meta-monitoring-based-validator)).
+
+There are a few ways in which we can achieve the outlined goals of this proposal and get the above measurements to impose a limit. The order of approaches is based on preference.
+
+### Meta-monitoring-based Receive Router Validation
+
+We could leverage a meta-monitoring solution which scrapes metrics from all Receivers as and then programmatically query for metrics like `prometheus_tsdb_head_series` and sum across tenants and instances periodically and limit based on that value (`latestCurrentSeries`).
```
douglascamata, line 72: I am confused about what is the "meta-monitoring solution". Is it a single Receive, called "Validator", that will be querying all the other Receive instances to aggregates tenants information? 

What happens in hashring architectures of multiple Router/Ingestor instances where a request could arrive at any instance? Is each Router/Ingestor also a Validator? Or one of them is picked to be a validator?

saswatamcode: Meta-monitoring is an external system that monitors Thanos components that you are running. Monitoring for your monitoring. 🙂 

In the context of this proposal, this would mean any solution which is capable of scraping metrics from all your Receive instances. That could be a single Prometheus that scrapes metrics from the cluster you are running Thanos in, or something more elaborate. 

This would usually vary from use case to use case, which is why we only mention "meta-monitoring" instead of accurately describing some set up.

douglascamata: > This would usually vary from use case to use case, which is why we only mention "meta-monitoring" instead of accurately describing some set up.

We don't need an accurate description, but a simple example. Like what you said below: 

> In the context of this proposal, this would mean any solution which is capable of scraping metrics from all your Receive instances. That could be a single Prometheus that scrapes metrics from the cluster you are running Thanos in, or something more elaborate.

This is so well written. IMHO, it should be part of the proposal so that every reader has the same understanding of what this "meta monitoring" system is.

saswatamcode: Will add!

docs/proposals-accepted/202206-active-series-limiting-hashring.md
```diff
@@ -0,0 +1,229 @@
...
+
+This is the simplest solution that can be implemented within Thanos Receive that can help us achieve best-effort limiting and stability. The fact that it does not rely on inter-Receive communication, which is very complex to implement, makes it a pragmatic solution.
+
+A full-fledged reference implementation of this can be found here: https://github.com/thanos-io/thanos/pull/5520.
+
+## Alternatives
+
+There are a few alternatives to what is proposed above,
+
+### Receive Router Validation
```
bwplotka, line 114: Perhaps we could try to put number prefixes for easier reference when discussing this proposal

docs/proposals-accepted/202206-active-series-limiting-hashring.md
```diff
@@ -0,0 +1,229 @@
...
+* In case, external meta-monitoring solution is down, can fall back to per-replica-tenant limits
+* Growing our instrumentation to improve validator, improves our observability as well
+
+#### Cons:
+
+* Not very accurate
+  * We do not know exact state of each TSDB, only know view of meta-monitoring solution, which gets updated on every scrape
+  * We do not account for how much a remote write request will increase the number of active series, only infer that from query result after the fact
+* Data replication (quorum-based) will likely cause inaccuracies in HEAD stat metrics
+* Dependence on external system, so best effort availability
```
bwplotka, line 102: It's not best effort availability - this depends on the external system.

I would rather form it as: 
```suggestion
* Dependence on an external meta monitoring system that is Prometheus API compatible. It's fairly easy and reliable with local Prometheus setup, but if the user uses 3rd party system like central remote monitoring, this might be less trivial to setup. 
```

docs/proposals-accepted/202206-active-series-limiting-hashring.md
```diff
@@ -0,0 +1,229 @@
...
+#### Pros:
+
+* Simpler as compared to other solutions and easier to implement
+* Lesser endpoint calls, so improved latency
+* Relies on "external to Thanos" system, and doesn’t increase load on Receive
+* Does not add much tenancy-based complexity to Thanos
+* No need to merge statistics across replicas, handled by meta-monitoring
+* Additional request-based rate limiting can be done within same component
+* In case, external meta-monitoring solution is down, can fall back to per-replica-tenant limits
+* Growing our instrumentation to improve validator, improves our observability as well
```
bwplotka, line 94:
```suggestion
* Growing our instrumentation to improve validator, improves our observability as well
* Easy to iterate. We can move out of meta-monitoring to different standalone solution if needed later on.
```

bwplotka (approved):
Great job, I love it! Really proud of your work here @saswatamcode  - very concise proposal 💪🏽 

LGTM, just small nits.

docs/proposals-accepted/202206-active-series-limiting-hashring.md
saswatamcode: Have added section numbers 🙂

saswatamcode:
Thanks, @bwplotka! 🙂

fpetkovski (approved):

bwplotka (approved):

bwplotka:
The fair point we missed by @matej-g - https://github.com/thanos-io/thanos/pull/5415#issuecomment-1153584790 let's start GH discussion for it, I think it's important, but we already leaned towards embedded tenancy functionality. (Multi-tenant receiveR).

PR #5417, opened 2022-06-13
Ruler: Change default evaluation interval to 1m

matej-g:

## Changes

As we're in the process of making our alerting compliant with the [Prometheus alerting compliance specification](https://github.com/prometheus/compliance/blob/main/alert_generator/specification.md#executing-an-alerting-rule), I have noticed we are using the default rule evaluation interval of `30s` whereas the spec requires this to be `1m` (referring to evaluation):
```
This MUST be done at regular intervals and the interval MUST be the interval from the parent <rule_group> of this alerting rule, which MUST default to 1 minute if not specified in the config.
```

I could not find any historical reasons for the difference between Thanos and Prometheus, I assume there is no practical reason for the difference and possibly this difference has been there since the initial implementation of ruler.

## Verification

/

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -23,6 +23,7 @@ We use *breaking :warning:* to mark changes that are not backward compatible (re
 ### Changed
 
 - [#5410](https://github.com/thanos-io/thanos/pull/5410) Query: Close() after using query. This should reduce bumps in memory allocations.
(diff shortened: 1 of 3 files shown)


squat (approved):
Nice find

GiedriusS (approved):

PR #5421, opened 2022-06-14
*: move to protobuf v2 api + vtprotobuf

GiedriusS:
Builds on top of our previous work. I think we could merge this on green.

Pros:
* We are getting rid of a huge unmaintained dependency;
* It unblocks enabling pooling for SeriesResponses.

Implementation notes:
* We are now downloading protoc compiler with needed includes from the Vitess project; in the future let's do this package ourselves? This is needed because IIRC upstream doesn't provide a trimmed zip with only the compiler + protobuf includes
* All generated types became pointers because upstream protobuf v2 API doesn't support the previously used extension 
* ~~I think this will force all of our downstream users i.e. Cortex/Mimir to upgrade to the protobuf V2 API~~ now that Cortex is in included in our repo, I have made the respective changes + regenerated protobuf files using the new compiler

TODO in future PRs:
* Bring back ZLabels - it's a bit trickier now because we will have to use `sed` to delete methods;


pkg/metadata/metadatapb/rpc.pb.go
```diff
@@ -1,1509 +1,516 @@
...
-	return m.Size()
-}
-func (m *MetricMetadata) XXX_DiscardUnknown() {
-	xxx_messageInfo_MetricMetadata.DiscardUnknown(m)
+	return storepb.PartialResponseStrategy(0)
 }
 
-var xxx_messageInfo_MetricMetadata proto.InternalMessageInfo
+type MetricMetadataResponse struct {
+	state         protoimpl.MessageState
```
GiedriusS, line 92: Generated structs now contain types that have locks inside hence we cannot copy them. What do you think should be done with them?

Issue #5404, opened 2022-06-02
Add write limits for tenants in Thanos Receiver 


douglascamata:
### Is your proposal related to a problem?

Tenants can overload Thanos Receivers with remote write requests and bring the whole system down.

### Describe the solution you'd like

I would like put maximum limits on the size of remote write requests, so that a single tenant is less likely to negatively affect others.

These are the proposed "knobs" for limiting (please feel free to propose more and give your opinion) the remote write endpoint usage:

- Amount of timeseries per request. Labels: **tenant** and **receive instance**.
- Amount of samples per request. Labels: **tenant** and **receive instance**.
- Size in bytes of the incoming request body. Labels: **tenant** and **receive instance**.
- Amount of concurrent HTTP requests. Labels: **receive instance**.

Hitting the two first limits should trigger an HTTP response with status code 413 (entity too large) and the last one should triggers a 429 (too many requests). In the future, the 413 error can be identified by the remote write clients and the data split into smaller requests under the limits. 

I would also like to expose the current values and limits of each "knob" as metrics of Thanos Receive. This would allow easy tracking of the limit system.

**To ensure backwards compatibility**, limiting should be optional and disabled by default.

For the sake of simplicity and iteration, starting with a global value for each knob (all tenants have the same upper bound limit) and later adding the possibility of configuring different values per tenant.

### Describe alternatives you've considered

- Ask tenants to be mindful of the amount of metrics they are sending
- Ask tenants to be careful when writing their remote write configurations
- Adding smarter/stateful rate limit knobs, that could track limits across a ring of Receive. But initial work for this is being started at #5333. Meanwhile, simpler limits can be added and be helpful.

### Additional context

- Similar abandoned proposal: #3963 (PR at #3964)
- Lots of inspiration being taken from [Cortex's limits](https://cortexmetrics.io/docs/configuration/configuration-file/#limits_config)

### TODO

- [x] Export the criteria for limiting the remote write requests as metrics of Thanos Receive (#5424).
- [x] Add charts with these metrics in the Receive dashboard. (#5472)
- [x] Add support for configuring & applying limits (single limit for all tenants, #5527). 
- [x] Add support for per-tenant limits. (#5565)
- [x] Add support for hot-reloading tenant limit configuration (#5673)
- [x] Add charts with these metrics in the Receive dashboard.

hanjm:
I think we can use bloom filter to track per tenant cardinality.

douglascamata:
[...]

PR #5424, opened 2022-06-15
[receive] Export metrics about remote write requests per tenant

douglascamata:

## Changes

Done as the 1st small step towards #5404: measuring information before we are able to start to apply limits.

* Export 4 new metrics from Thanos Receive that describe the shape of remote write requests **per tenant**:
  * total of timeseries per request
  * total of samples per request
  * total of bytes per request
  * ~~amount of in-flight HTTP requests~~ (cut out for a separate PR)

## Verification

* Ran it locally.


pkg/receive/handler.go
```diff
@@ -142,6 +148,38 @@ func NewHandler(logger log.Logger, o *Options) *Handler {
...
+			}, []string{"code", "method", "tenant"},
+		),
+		writeInflightHTTPRequests: promauto.With(o.Registry).NewGaugeVec(
+			prometheus.GaugeOpts{
+				Namespace: "thanos",
+				Subsystem: "receive",
+				Name:      "write_requests_inflight",
+				Help:      "The number of inflight write HTTP requests being handled at the same time.",
+			}, []string{"tenant"},
+		),
```
douglascamata, line 182: Interestingly, these two last (total of bytes per request and amount of inflight HTTP requests), are already exported by the `pkg/extprom/http.defaultInstrumentationMiddleware`. Although such middleware is generic and doesn't include the tenant label, which is important for the Thanos Receive context.

I am not sure what to do with the default instrumentation middleware in this case. I think it should be adapted for Thanos Receive, so that we don't duplicate work to keep track of these metrics. I am looking for feedback/help on this regard.

pkg/receive/handler.go
```diff
@@ -142,6 +148,38 @@ func NewHandler(logger log.Logger, o *Options) *Handler {
...
+			}, []string{"code", "method", "tenant"},
+		),
+		writeSamplesTotal: promauto.With(o.Registry).NewCounterVec(
+			prometheus.CounterOpts{
+				Namespace: "thanos",
+				Subsystem: "receive",
+				Name:      "write_samples_total",
+				Help:      "The number of sampled received in the incoming write requests.",
+			}, []string{"code", "method", "tenant"},
+		),
```
douglascamata, line 166: I would like to propose to transform these two metrics into histograms or summaries. What are the recommendations? 

In case histogram is a good choice, I think it would be good to let the user configure the buckets and we provide some sensible defaults. Maybe a `prometheus.ExponentialBucket` would be a good starter?

CHANGELOG.md
```diff
@@ -21,6 +21,8 @@ We use *breaking :warning:* to mark changes that are not backward compatible (re
 - [#5352](https://github.com/thanos-io/thanos/pull/5352) Cache: Add cache metrics to groupcache.
 - [#5391](https://github.com/thanos-io/thanos/pull/5391) Receive: Add relabeling support.
 - [#5408](https://github.com/thanos-io/thanos/pull/5391) Receive: Add support for consistent hashrings.
+- [#5424](https://github.com/thanos-io/thanos/pull/5424) Receive: export metrics regarding size of remote write requests
+- [#5424](https://github.com/thanos-io/thanos/pull/5424) HTTP metrics: export number of in-flight HTTP requests
```
bwplotka, line 25: Hmm, it really looks like we should probably split this PR ;p

pkg/extprom/http/instrument_server.go
```diff
@@ -37,97 +36,63 @@ func NewNopInstrumentationMiddleware() InstrumentationMiddleware {
...
-			},
-			[]string{"code", "handler", "method"},
-		),
-	}
-	return &ins
 }
 
 // NewHandler wraps the given HTTP handler for instrumentation. It
-// registers four metric collectors (if not already done) and reports HTTP
+// registers five metric collectors (if not already done) and reports HTTP
```
bwplotka, line 51: What's the reason behind in-flight? It's usually misleading metrics given 15s interval and requests finishing sooner.. but maybe we can try having it to double check

pkg/receive/handler.go
bwplotka: Histograms with hardcoded (initally) sane buckets could do the work.  👍🏽 

I would create manually exponential buckets - to manually check what we might find useful. Also probably we could define (if not yet) some sane limits for shape of request which can configure max bucket (:

pkg/receive/handler.go
```diff
@@ -363,25 +395,32 @@ func (h *Handler) receiveHTTP(w http.ResponseWriter, r *http.Request) {
 		return
 	}
 
+	responseStatusCode := http.StatusOK
 	err = h.handleRequest(ctx, rep, tenant, &wreq)
 	if err != nil {
```
bwplotka, line 400:
```suggestion
	if err = h.handleRequest(ctx, rep, tenant, &wreq); err != nil {
```

pkg/receive/handler.go
```diff
@@ -363,25 +395,32 @@ func (h *Handler) receiveHTTP(w http.ResponseWriter, r *http.Request) {
...
+		case errConflict:
+			responseStatusCode = http.StatusConflict
+			http.Error(w, err.Error(), responseStatusCode)
+		case errBadReplica:
+			responseStatusCode = http.StatusBadRequest
+			http.Error(w, err.Error(), responseStatusCode)
+		default:
+			level.Error(tLogger).Log("err", err, "msg", "internal server error")
+			responseStatusCode = http.StatusInternalServerError
+			http.Error(w, err.Error(), responseStatusCode)
```
bwplotka, line 418: Should we put this line after switch, instead of repeating it everywhere?

pkg/receive/handler.go
```diff
@@ -142,6 +145,22 @@ func NewHandler(logger log.Logger, o *Options) *Handler {
 				Help: "The number of times to replicate incoming write requests.",
 			},
 		),
+		writeTimeseriesTotal: promauto.With(o.Registry).NewCounterVec(
+			prometheus.CounterOpts{
+				Namespace: "thanos",
+				Subsystem: "receive",
+				Name:      "write_timeseries_total",
+				Help:      "The number of timeseries received in the incoming write requests.",
+			}, []string{"code", "method", "tenant"},
```
bwplotka, line 154: Let's kill the "method" - not interesting for us.

pkg/receive/handler.go
```diff
@@ -363,25 +395,32 @@ func (h *Handler) receiveHTTP(w http.ResponseWriter, r *http.Request) {
...
-	case errConflict:
-		http.Error(w, err.Error(), http.StatusConflict)
-	case errBadReplica:
-		http.Error(w, err.Error(), http.StatusBadRequest)
-	default:
-		level.Error(tLogger).Log("err", err, "msg", "internal server error")
-		http.Error(w, err.Error(), http.StatusInternalServerError)
+	h.writeTimeseriesTotal.WithLabelValues(strconv.Itoa(responseStatusCode), r.Method, tenant).Add(float64(len(wreq.Timeseries)))
+	for _, timeseries := range wreq.Timeseries {
+		h.writeSamplesTotal.WithLabelValues(strconv.Itoa(responseStatusCode), r.Method, tenant).Add(float64(len(timeseries.Samples)))
```
bwplotka, line 423: This is interesting... so with histogram we will store number samples in each series in bucketed view.... But this will be unrelated to requests really. Which is  fine I guess, just a specific behaviour of this metric. But then metrics should be called `write_samples_per_series_total`.

Alternative is to  have `write_samples_total`  which will observe "sum" of samples in all timeseries - might be more reasonable to  have. cc @moadz as it's similar to T-shirt sizes request discussion - What happened to that initiative? Did we add anything at the end?

bwplotka (commented):
Amazing work! Some comments (:

pkg/extprom/http/instrument_server.go
douglascamata: @bwplotka my attention adding the in-flight is to allow users of Thanos Receive to get better insights in situations of high load, where the number of concurrent requests will be relevant. Also, later on there will be a "metered gate" to limit the amount of in-flight requests.

pkg/receive/handler.go
douglascamata: It was supposed to be the sum of samples in all timeseries since the start. I'll fix it and keep an eye on @moadz answer 👀 👍

douglascamata:
> I would create manually exponential buckets - to manually check what we might find useful. Also probably we could define (if not yet) some sane limits for shape of request which can configure max bucket (:

@bwplotka: it's difficult to suggest a lax limit for the buckets that makes me happy, as it depends on the writer's configuration and the size of the Receive instance. So I searched for some inspiration in [the default Prometheus remote write configuration](https://prometheus.io/docs/practices/remote_write/#memory-usage), which sends 500 samples per second. Let me know if you have more specific suggestions for the buckets.

pkg/receive/handler.go
moadz: The PR for query metrics has not been raised yet.

pkg/extprom/http/instrument_server.go
moadz: This is actually quite useful to know given atm there is no notion of enqueued requests in receiver. This is particularly useful for troubleshooting issues with load/receiver getting requests that it is not completing.

douglascamata:
PTAL whenever you can, @bwplotka 👍

pkg/receive/handler.go
bwplotka: ack. Started some yolo discussions about limits though (semi-related https://github.com/thanos-io/thanos/issues/5438)

bwplotka (approved):
👍🏽 Amazing, LGTM

pkg/extprom/http/instrument_server.go
douglascamata: Split PR with in-flight requests metrics: #5424.

PR #5426, opened 2022-06-20
remove unused block-sync-concurrency flag

yeya24:

## Changes

The flag `--block-sync-concurrency` in compactor is not used anywhere so remove it.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -28,6 +28,8 @@ We use *breaking :warning:* to mark changes that are not backward compatible (re
 
 ### Removed
 
+- [#5426](https://github.com/thanos-io/thanos/pull/5426) Compactor: Remove flag `--block-sync-concurrency`.
+
 ## [v0.26.0](https://github.com/thanos-io/thanos/tree/release-0.26) - 2022.05.05
 
 ### Fixed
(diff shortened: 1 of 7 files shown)


CHANGELOG.md
```diff
@@ -28,6 +28,8 @@ We use *breaking :warning:* to mark changes that are not backward compatible (re
 
 ### Removed
 
+- [#5426](https://github.com/thanos-io/thanos/pull/5426) Compactor: Remove flag `--block-sync-concurrency`.
```
matej-g, line 31: Maybe let's specify this was not being used, so it does not give impression this is a breaking change?

matej-g (commented):
Looks good, probably just E2E need small adjustment.

CHANGELOG.md
yeya24: Updated.

matej-g (approved):

squat:
What do we you mean by `is not used anywhere`? It seems that the flag value _was_ passed to the NewMetaSyncer func, no?

yeya24:
> What do you mean by `is not used anywhere`? It seems that the flag value _was_ passed to the NewMetaSyncer func, no?

It is passed into the syncer but syncer doesn't use this value anywhere.

PR #5315, opened 2022-04-28
Added alert compliance test for Thanos

bwplotka:
Fixes https://github.com/thanos-io/thanos/issues/5203

Chained on top of https://github.com/thanos-io/thanos/pull/5348, let's review that one first.

PR #5433, opened 2022-06-22
Alert Compliance: Fix wrong ruler configuration

matej-g:

## Changes

This is a follow-up for https://github.com/thanos-io/thanos/pull/5315. There is one more thing that got mixed up during rebasing, which is that we should not be using replica label for ruler (this causes the test to fail). This change adds configuration to force not using label.

## Verification
Ran the alert compliance test suite locally on top fresh `main`, all tests are passing.

### DIFF
--- a/test/e2e/e2ethanos/services.go
+++ b/test/e2e/e2ethanos/services.go
@@ -494,12 +494,6 @@ func NewRulerBuilder(e e2e.Environment, name string) *RulerBuilder {
 	}
 }
 
-// Force no replica label; by default we use replica label in tests.
-func (r *RulerBuilder) WithoutReplicaLabel() *RulerBuilder {
-	r.replicaLabel = ""
-	return r
-}
-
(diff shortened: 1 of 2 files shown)


test/e2e/e2ethanos/services.go
```diff
@@ -494,6 +494,12 @@ func NewRulerBuilder(e e2e.Environment, name string) *RulerBuilder {
 	}
 }
 
+// Force no replica label; by default we use replica label in tests.
+func (r *RulerBuilder) WithoutReplicaLabel() *RulerBuilder {
+	r.replicaLabel = ""
+	return r
+}
```
yeya24, line 501: Can we use `WithReplicaLabel("")` directly?

matej-g: You're right, that just as readable and simpler, I removed that option.

yeya24 (approved):

Issue #4941, opened 2021-12-09
Allow relabel while compacting blocks


vibgy:
### Is your proposal related to a problem?

Some labels have super high cardinality but are relevant for a short time. 

For example - `pod_name` is useful compared to IP address (`instance` label) if you are looking at relatively recent metrics, but after these pods get deleted, `pod_name` is not relevant. 

We would like to be able to drop labels while compacting blocks. 

### Describe the solution you'd like

I would like to be able to drop some labels while compacting blocks (in Thanos compactor)
(e.g. Drop `pod_name` label while creating a 2d block out of multiple 8h blocks)

### Describe alternatives you've considered

We have tried to do it manually using Thanos bucket rewrite, but it is error prone if the compactor is running at the same time. Moreover, it’s not efficient because a new block has to be created and the old one has to be deleted. 

### Additional context

There is a precedence for a client provided `mergeFunc` ([here](https://github.com/prometheus/prometheus/blob/main/tsdb/compact.go#L85)) to the `tsdb.LeveledCompactor` that manipulates SeriesChunk sets. Along the same lines, we could extend the tsdb compactor to allow for providing a `relabelFunc`. This would allow a client to modify the `Symbols` and `Labels` before writing the compacted block. For example - This would allow thanos compactor to pass in a `relableFunc` while compacting the appropriate block with user provided relabelConfig. 

If this proposal makes sense to the Thanos community, then I can create a feature request in Prometheus repo and can also create PRs. 

If you made it this far, thank you for reading. Would love you inputs and ideas.

yeya24:
> e.g. Drop pod_name label while creating a 2d block out of multiple 8h blocks

This makes more sense to do some aggregation like recording rule instead of dropping labels directly.

 > There is a precedence for a client provided mergeFunc (here) to the tsdb.LeveledCompactor that manipulates SeriesChunk sets. Along the same lines, we could extend the tsdb compactor to allow for providing a relabelFunc. This would allow a client to modify the Symbols and Labels before writing the compacted block. For example - This would allow thanos compactor to pass in a relableFunc while compacting the appropriate block with user provided relabelConfig.

Maybe you can raise this to the Prometheus repo and see how we can extend the compactor to fullfil your requirements.

bwplotka:
[...]

Issue #903, opened 2019-03-11
Per rule aggregation retention


bwplotka:
In ideal world retention is not necessary on downsampling/raw leve, but on aggregation level. 

We need to have the way to bring it through in LTS system like Thanos.

AC:
* User can specify certain aggregation to be retained longer than others
* Design doc (proposal is in place)

It cames to the fact that you want per metric retention ideally on compactor. This is bit related to [delete_series](https://github.com/improbable-eng/thanos/issues/415) as it might involve block rewrite in edge case... We need to design this. 

Thoughts @improbable-ludwik @brancz @devnev @domgreen

brancz:
There have been discussions on per time-series retention on upstream Prometheus before, I think at least having a discussion with the team is worth it, just to see if there are any insights from back then.

> proposal is in place

What do you mean by this? As far as I can tell there is no design written up anywhere, but may very well have missed it.

Off the top of my head, this could be a configuration which is a combination of a Prometheus style label selector, as well a respective rule of which resolution to keep for how long.

As a whole this is definitely not trivial, but I agree much needed.

bwplotka:
Sorry, no proposal has to be in place, that's what I meant.

(: Relabel like config makes sense but essentially we are talking about rewrite in compactor for this right?

brancz:
Configuration is a technicality, I'm not entirely sure relabelling would work exactly but something close to that probably yes. I agree the compactor is the component that needs to take care of this by re-writing blocks.

ipstatic:
With the federation system we have in place now, we have trained users that want metrics to be preserved (federated) have to use a specific recording rule style name so that the metrics would get federated. We would love to continue this practice with Thanos and only retain metrics that meet a specific format for an extended period of time.

bwplotka:
Some offline discussions revealed that users still do it with Thanos, but using federation + Thanos on top.

```
The point was to shard ingestion and execute recording rules against each sharded ingestion gateway (which runs with minimal retention).  We have some very high-cardinality metrics and centralized ingestion alone was prohibitively expensive.  That prometheus federation layer allows us to compute/ingest aggregates (without retaining the raw metrics).
```

I think we should aim to allow users to avoid this, but cannot see immdiately blocker for that, otherwise than more complex system and query being able to fetch data with some lag (rule eval lag + federated scrape)

smalldirector:
Just saw these discussion threads on potential feature requirement `per metric retention on compactor`, do we still look for this feature on compactor side?

[...]

PR #5436, opened 2022-06-23
Add series deletion with compactor proposal

yeya24:

## Changes

Add a proposal about series deletion. Continue work of https://thanos.io/tip/proposals-accepted/202012-deletions-object-storage.md/

For the reference implementation, please take a look at the two below:
1. [Tombstone CLI and Store Gateway Series Masking](https://github.com/thanos-io/thanos/pull/4790)
2. [Tombstone Compactor](https://github.com/yeya24/thanos/pull/375)

### DIFF
--- a/docs/proposals-accepted/202206-block-series-deletion-compactor.md
+++ b/docs/proposals-accepted/202206-block-series-deletion-compactor.md
@@ -33,7 +33,7 @@ The main workflow of the series deletion will look like this:
 
 ### Tombstone File
 
-The design of the tombstone file mainly follows the previous proposal. Tombstone file is immutable. It is using a custom JSON format, single file per request and global, which means it applies to all the blocks. For the detailed considerations and tradeoffs, please refer to the original proposal.
+The design of the tombstone file mainly follows the previous proposal. Tombstone file is immutable. It is using a custom JSON format, single file per request and global, which means it applies to all the blocks if no external label or block ID matchers are specified. For the detailed considerations and tradeoffs, please refer to the original proposal.
 
 The file format is as follows:
 
@@ -53,97 +53,54 @@ cat 01G61GC0Q957TVZ2FC8S6Y2ZK7.json
 }
 ```
 
-Similar to the block `meta.json` file, the tombstone file is named using ULID which is basically its creation time. For other fields we have:
-* **label matchers**: label matchers to match the block and series. Like this example, blocks can be matched using external labels and blocks can be matched using a special `__block_id` label. All other labels are applied to series. This design makes it easy to support multi-tenant environment and match specific blocks.
+Similar to the block `meta.json` file, the tombstone file is named by ULID from its creation time. For other fields we have:
+* **label matchers**: label matchers to match the block and series. Like this example, blocks can be matched using external labels and blocks can be matched using a special `__block_id` label. All other labels are applied to series. If no external label matcher and block ID matcher specified, tombstones apply to all blocks. This design makes it easy to support multi-tenant environment and match specific blocks.
 * **start timestamp**
-* **end timestamp**: end timestamp should be the tombstone creation time if not specified, otherwise the compactor needs to wait for a long time to make sure all matched blocks are processed. Considering the tombstone pending period, the end timestamp can be set to `(now + pending period)` as well. We don't consider history blocks backfilling now.
+* **end timestamp**: If end timestamp can be a future time, then the compactor might need to wait for a long time to make sure all matched blocks are processed. To avoid this, end timestamp is the tombstone creation time if not specified and cannot be later than it. Considering the tombstone pending period, the end timestamp can be set to `(now + pending period)` as well. We don't consider history blocks backfilling for now.
 * **author**
 * **reason**
-
-### Store Gateway Masks Series based on Tombstones
-
-Like block metadata, the store gateway also fetches tombstones in a configurable interval, which means some delay is expected: user creates a tombstone but she is still able to query the data to be deleted.
-
-There are different ways to mask series based on tombstones. We will discuss two options here:
-
-#### Option 1: Mask Series At Query Time
-
-During query time, the store gateway decides what blocks to query based on the request min, max time and label matchers. Meanwhile, it can also decide what tombstones are going to be applied for the selected blocks based on tombstone time range and label matchers.
-
-Then during the query time, it checks the series label sets against the tombstone matchers, as well as the chunks and tombstone time range to mask the series data.
-
-Pros:
-* Easy to implement and very straightforward
-* Less memory usage.
-
-Cons:
-* Slower at query time
-
-#### Option 2: Mask Series At Tombstone Sync Time
-
-During the tombstone file sync, store gateway can also eagerly apply the tombstones to all blocks, calculate the matched postings and time range for each block. For each block, each tombstone's processed result will be stored in memory using the Prometheus `MemTombstone` [format](https://github.com/prometheus/prometheus/blob/main/tsdb/tombstones/tombstones.go#L230).
-
-```go
-// Result for the processed result for each tombstone, each block.
-type MemTombstones struct {
-    intvlGroups map[storage.SeriesRef]Intervals
-    mtx         sync.RWMutex
-}
-
-// Result for the processed result for each block.
-// Key: global tombstone ID, Value: processed result for each tombstone.
-type BlockTombstones map[ulid.ULID]*MemTombstones
-```
-
-Since all the blocks and tombstones are immutable, this result can be easily cached in memory or in filesystem. Ideally for each tombstone, each block we just need to calculate once. Since `BlockTombstones` is a map, it is easy to add or delete one key if users want to add more tombstones or remove one tombstone.
-
-During the query time, the cached tombstone will be used. As it knows which `SeriesRef`, which intervals to apply, it is very straightforward to mask series.
-
-Pros:
-* More efficient during query time, less processing required.
-* As data can be cached there is not much computation required.
-
-Cons:
-* More memory required.
-* If read is much less than write, it is inefficient as the pre-calculated result won't be used much.
-
-#### Considerations
-
-Considering the infrequent change of tombstones, option 2 is preferred as all data can be cached and it is faster during query time. 
 
 ### Tombstone Deletion with Compactor
 
 Compactor performs the job of deleting the series from object storage and cleans up tombstone files.
 
 The main challenge is about scalability. The reasons are:
-* Global tombstone files makes the compactor difficult to know which blocks match the tombstones. It needs to scan all the block indexes to find the matched blocks.
-* If one block has too many matched series to mask, the query performance can be hugely impacted. In this case, compactor planner should be able to know the ratio of the masked series / total series. If the ratio is high, the block needs to be compacted and masked series should be removed from the block.
+* Global tombstone files make the compactor difficult to know which blocks match the tombstones. It needs to scan all the blocks to find the blocks matched.
+* If one block has too many matched series to mask, the query performance can be hugely impacted. In this case, compactor planner should be able to tell the ratio of the masked series / total series. If the ratio is high, the block needs to be compacted and masked series should be removed from the block.
 
 There are other questions about how we should perform the deletion along with compaction:
 * Series are deleted at normal compaction. This might need to wait a long time and hard to apply for blocks at the max compaction level.
 * Rewrite blocks separately. This needs more coordination with compaction to avoid duplicate/unnecessary work.
 
-To overcome those challenges, we propose three options.
+To overcome those challenges, I'd like to propose one solution with blocks scanning from the compactor.
 
-#### Option 1: Compactor Scans Blocks
+#### Proposed Option: Compactor Scans Blocks and Cache Tombstone Results
 
-Like store gateway, compactor can also scan block indexes to find the blocks matching the tombstones. Similar to the `Mask Series At Tombstone Sync Time` of store gateway, compactor can calculate the processed tombstones for each block, each tombstone and cache the results locally. As blocks and tombstones are immutable, the result needs to be calculated only once and then reused easily. As the Prometheus format tombstones are calculated, during the compaction time the compactor just needs to generate the tombstone file in the filesystem and the TSDB compactor will apply the deletion.
+Compactor can scan block indexes to find the blocks and series matching the tombstones. In Prometheus, an in-memory `MemTombstone` [format](https://github.com/prometheus/prometheus/blob/main/tsdb/tombstones/tombstones.go#L230) is used to efficiently mask and delete series during query and compaction time.
 
-There are two approaches about how to scan block indexes:
-1. Download all the blocks locally and check against the tombstones.
+With block index, compactor can just calculate the prometheus `MemTombstone` for each block, each tombstone and cache the results locally. We can use the code below.
+
+As blocks and tombstones are immutable, the result needs to be calculated only once and then reused easily. As the Prometheus format tombstones are calculated, during the compaction time the compactor just needs to generate the tombstone file in the filesystem and the TSDB compactor will apply the deletion.
+
+There are two approaches about how to scan blocks:
+1. Download blocks locally and check against the tombstones. This needs a large amount of disk space and we might download and check each block one by one.
 2. Use part of the store gateway code to scan the block index and chunks on the object storage to generate `MemTombstones` result.
 
-Approach 2 should be better as only block indexes need to be downloaded, chunks data are reading from the object storage on the fly so disk space can be saved. As compactor is an offline component, it is acceptable to read from the object storage rather than downloading the whole data.
+Approach 1 is simple and easy to implement. But it doesn't scale because it needs to download the block locally and then check the tombstone matchers. If there is no matching series then it is a waste of time and disk space to download the whole block.
 
-The pitfall is that we need to expose part of the store gateway code like `BucketIndexReader`, `BucketChunkReader` as well as other internal code to be used by the compactor.
+So we propose our solution with option 2. With this option, only block indexes need to be downloaded, chunks data are reading from the object storage on the fly so disk space can be saved.  As compactor is an offline component, it is acceptable to read from the object storage rather than downloading the whole data.
 
-#### Option 2: New API to Match Tombstones
+Another change is that we need to expose part of the store gateway code like `BucketIndexReader`, `BucketChunkReader` as well as other internal code to be used by the compactor. Some refactorings are needed to make those code reusable.
 
-Since the block indexes are already downloaded by the store gateway, it can already scan the blocks and calculate the tombstones. So a way here is to expose some APIs from store gateway, and then compactor can query the APIs to know what blocks can be deleted.
+#### Alternatives
 
-The gRPC API is called `Tombstone` API. The compactor can either query from the store gateway directly or query from queriers, and queriers do the API federation from multiple store gateway instances.
+##### New API to Match Tombstones
 
-The example gRPC format looks like below. The response is almost the same as the `MemTombstone` result.
+Since the block indexes are already downloaded by the store gateway, it can already scan the blocks and calculate the tombstones. So a way here is to expose some APIs from store gateway, and then compactor can query the APIs to get the tombstones or know which blocks to download for deletion.
+
+We propose a new `Tombstone` gRPC API at store gateway and querier is able to federate `Tombstone` API from multiple store gateways instances. The compactor can query from queriers to get required tombstones data. Querier federation is required because each store gateway only takes care of some blocks when sharding is enabled.
+
+The example gRPC request and response look like below. The response is almost the same as the `MemTombstone` result.
 
 ```protobuf
 service Tombstone {
@@ -171,21 +128,72 @@ message Interval {
 }
 ```
 
-After getting the query result, the compactor does almost the same thing as the previous option: generate the tombstone file during compaction time.
+After getting the query result, the compactor does almost the same thing as the proposed option: generate the tombstone file during compaction time.
+
+This option can be considered as a variation for the proposed solution. Compactors avoid scanning blocks by querying store gateway, thus improving scalability.
 
 The drawback for this approach is that we are inventing a new API for this purpose only, which might be a bit of pain in the future.
 
-#### Option 3: Delete Series using `thanos tools block rewrite` CLI
+##### Delete Series using `thanos tools block rewrite` CLI
 
-The previous two options do the series deletion at compaction time, this option does actual series deletion using block rewrite. For the index scanning, it is still required to choose either option 1 or 2 to avoid unncessary rewrite even if the block doesn't have any matched series.
+To perform series deletion, there is a `block rewrite tool`. But to integrate it with the compactor, the CLI won't be used directly. Instead, we will reuse the code to implement the rewrite logic in the compactor.
 
-This is not efficient as rewrite is a different process from compaction. We need to add this step and make sure rewrite can work together with compaction and no redundant work is done.
+To perform block rewrite, we still need to know what blocks to rewrite/delete so block scanning is still required to avoid unnecessary operations. For the blocks scanning, either scanning blocks directly by the compactor or querying `Tombstone` APIs can be used.
 
-#### Considerations
+Overall, this is not efficient as rewrite is a different step from compaction so blocks might need to be compacted multiple times, causing [write amplification](https://en.wikipedia.org/wiki/Write_amplification) issues. We also need to add this step to the planner and make sure it works together with existing compaction.
 
-Overall, among the 3 options, option 1 is better as it is easier to implement and straightforward. No additional API required or no additional logic change on the compaction side.
+### Additional Design Considerations
 
-### Other Design Considerations
+#### Store Gateway Masks Series based on Tombstones
+
+Like block metadata, the store gateway also fetches tombstones in a configurable interval, which means some delay is expected: user creates a tombstone but she is still able to query the data to be deleted.
+
+There are different ways to mask series based on tombstones. In the previous proposal, we didn't mention the detailed implementation. We will discuss two options here:
+
+##### Option 1: Mask Series At Query Time
+
+During query time, the store gateway decides what blocks to query based on the request min, max time and label matchers. Meanwhile, it can also decide what tombstones are going to be applied for the selected blocks based on tombstone time range and label matchers.
+
+Then during the query time, it checks the series label sets against the tombstone matchers, as well as the chunks and tombstone time range to mask the series data.
+
+Pros:
+* Easy to implement and very straightforward
+* Less memory usage.
+
+Cons:
+* Slower at query time
+
+##### Option 2: Mask Series At Tombstone Sync Time
+
+During the tombstone file sync, store gateway can also eagerly apply the tombstones to all blocks, calculate the matched postings and time range for each block. The result is the same as `MemTombstone` we mentioned before.
+
+```go
+// Result for the processed result for each tombstone, each block.
+type MemTombstones struct {
+    intvlGroups map[storage.SeriesRef]Intervals
+    mtx         sync.RWMutex
+}
+
+// Result for the processed result for each block.
+// Key: global tombstone ID, Value: processed result for each tombstone.
+type BlockTombstones map[ulid.ULID]*MemTombstones
+```
+
+Since all the blocks and tombstones are immutable, this result can be easily cached in memory or in filesystem. Ideally for each tombstone, each block we just need to calculate once. Because `BlockTombstones` is a map, it is easy to add or delete one key if users want to add more tombstones or remove one tombstone.
+
+During the query time, the cached tombstone will be used. As it knows which `SeriesRef`, which intervals to apply, it is very straightforward to mask series.
+
+Pros:
+* More efficient during query time, less processing required.
+* As data can be cached there is not much computation required.
+
+Cons:
+* More memory required.
+* If read is much less than write, it is inefficient as the pre-calculated result won't be used much.
+
+##### Considerations
+
+Considering the infrequent change of tombstones, option 2 is preferred as all data can be easily cached, and it is faster during query time.
 
 #### Compaction Planner for Tombstone
 
@@ -201,11 +209,11 @@ For any new blocks that are compacted by blocks along with some tombstone reques
 
 Tombstones will skip blocks that have already been applied with its ID.
 
-### Tombstones sharding
+#### Tombstones sharding
 
 Unlike blocks, we cannot shard tombstones simply using its ULIDs because each tombstone is able to match all blocks from the object storage. When compactor sharding is enabled, it still needs to check all the tombstones in the shard to find the matched blocks.
 
 ## Conclusions
 
 * For store gateway, it masks series at tombstone sync time.
-* For compactor, it scans block indexes directly and caches the tombstones results. During compaction time, it generates the tombstone file and series deletion is done by the TSDB compactor](#option 1:-compactor-scans-blocks.
+* For compactor, it scans block indexes directly and caches the tombstones results. During compaction time, it generates the tombstone file and series deletion is done by the TSDB compactor.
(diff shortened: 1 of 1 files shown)


docs/proposals-accepted/202206-block-series-deletion-compactor.md
```diff
@@ -0,0 +1,206 @@
...
+
+After one compaction iteration is done, metadata will be synced again and compactor starts doing garbage collection. This time the compactor can sync tombstones again and check is there any matched block for the tombstone file. If no, then the tombstone file can be deleted.
+
+For any new blocks that are compacted by blocks along with some tombstone requests, we want to avoid calculating tombstones on it again as those series are deleted from it already. We propose a new field in the `meta.json` file called `tombstone_applied` to indicate what tombstones have been applied to the block.
+
+Tombstones will skip blocks that have already been applied with its ID.
+
+### Tombstones sharding
+
+Unlike blocks, we cannot shard tombstones simply using its ULIDs because each tombstone is able to match all blocks from the object storage. When compactor sharding is enabled, it still needs to check all the tombstones in the shard to find the matched blocks.
```
yeya24, line 201: The tricky thing of only sharding blocks, not tombstones is that we cannot know whether all matched blocks are processed for the given tombstone as each compactor only has information about its own blocks

docs/proposals-accepted/202206-block-series-deletion-compactor.md
```diff
@@ -0,0 +1,206 @@
...
+
+#### Option 1: Compactor Scans Blocks
+
+Like store gateway, compactor can also scan block indexes to find the blocks matching the tombstones. Similar to the `Mask Series At Tombstone Sync Time` of store gateway, compactor can calculate the processed tombstones for each block, each tombstone and cache the results locally. As blocks and tombstones are immutable, the result needs to be calculated only once and then reused easily. As the Prometheus format tombstones are calculated, during the compaction time the compactor just needs to generate the tombstone file in the filesystem and the TSDB compactor will apply the deletion.
+
+There are two approaches about how to scan block indexes:
+1. Download all the blocks locally and check against the tombstones.
+2. Use part of the store gateway code to scan the block index and chunks on the object storage to generate `MemTombstones` result.
+
+Approaches 2 should be better as only block indexes need to be downloaded, chunks data are reading from the object storage on the fly so disk space can be saved. As compactor is an offline component, it is acceptable to read from the object storage rather than downloading the whole data.
```
douglascamata, line 133: Typo here in `approaches`, should be `approach`.

docs/proposals-accepted/202206-block-series-deletion-compactor.md
```diff
@@ -0,0 +1,206 @@
...
+
+Pros:
+* More efficient during query time, less processing required.
+* As data can be cached there is not much computation required.
+
+Cons:
+* More memory required.
+* If read is much less than write, it is inefficient as the pre-calculated result won't be used much.
+
+From my perspective, option 2 is better because data can be cached.
```
douglascamata, line 109: As I reader, I wonder what is `option 2` here. You mentioned "options" before, but you didn't number them 1 or 2. I am confused whether this is a reference to `Option 2: New API to Match Tombstones` (inside the tombstone deletion heading). 

I recommend numbering the options `Mask Series At Query Time` and `Mask Series At Tombstone Sync Time`, like done for the discussions about deletion with Compactor.

I believe that putting your recommendation under a separate heading would be great to give it evidence.

docs/proposals-accepted/202206-block-series-deletion-compactor.md
```diff
@@ -0,0 +1,206 @@
...
+
+The drawback for this approach is that we are inventing a new API for this purpose only, which might be a bit of pain in the future.
+
+#### Option 3: Delete Series using Rewrite Logic
+
+The previous two options do the series deletion at compaction time, this option does actual series deletion using block rewrite. For the index scanning, it is still required to choose either option 1 or 2 to avoid unncessary rewrite even if the block doesn't have any matched series.
+
+This is not so efficient as rewrite is a different process than compaction. We need to add this step and make sure rewrite can work together with compaction and no redundant work is done.
+
+Overall, among the 3 options, option 1 is better as it is easier to implement and straightforward.
```
douglascamata, line 181: Same comment here about giving more evidence to your recommendation by putting it under a `My recommendation` (or something like this) heading. This is buried inside the description of an option and it's not part of the option itself.

docs/proposals-accepted/202206-block-series-deletion-compactor.md
```diff
@@ -0,0 +1,206 @@
...
+Tombstones will skip blocks that have already been applied with its ID.
+
+### Tombstones sharding
+
+Unlike blocks, we cannot shard tombstones simply using its ULIDs because each tombstone is able to match all blocks from the object storage. When compactor sharding is enabled, it still needs to check all the tombstones in the shard to find the matched blocks.
+
+## Conclusions
+
+* For store gateway, it masks series at tombstone sync time.
+* For compactor, it scans block indexes directly and caches the tombstones results. During compaction time, it generates the tombstone file and series deletion is done by the TSDB compactor.
```
douglascamata, line 206: I think it's valuable to link back to the headings where these options where described, so that in the future we can quickly jump to this section and then back to the options that were implemented.

docs/proposals-accepted/202206-block-series-deletion-compactor.md
yeya24: Thanks! That makes total sense and I should really do it from the beginning. 💯

douglascamata (approved):
LGTM. Awesome job, @yeya24 👏

docs/proposals-accepted/202206-block-series-deletion-compactor.md
```diff
@@ -0,0 +1,210 @@
...
+
+The main challenge is about scalability. The reasons are:
+* Global tombstone files makes the compactor difficult to know which blocks match the tombstones. It needs to scan all the block indexes to find the matched blocks.
+* If one block has too many matched series to mask, the query performance can be hugely impacted. In this case, compactor planner should be able to know the ratio of the masked series / total series. If the ratio is high, the block needs to be compacted and masked series should be removed from the block.
+
+There are other questions about how we should perform the deletion along with compaction:
+* Series are deleted at normal compaction. This might need to wait a long time and hard to apply for blocks at the max compaction level.
+* Rewrite blocks separately. This needs more coordination with compaction to avoid duplicate/unnecessary work.
+
+To overcome those challenges, we propose two options.
```
bwplotka, line 125: Can we propose one option (pick the one you like the most) and put rest to alternatives?

docs/proposals-accepted/202206-block-series-deletion-compactor.md
```diff
@@ -0,0 +1,210 @@
...
+
+Cons:
+* More memory required.
+* If read is much less than write, it is inefficient as the pre-calculated result won't be used much.
+
+#### Considerations
+
+Considering the infrequent change of tombstones, option 2 is preferred as all data can be cached and it is faster during query time. 
+
+### Tombstone Deletion with Compactor
```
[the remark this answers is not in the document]
bwplotka: Reading further that store series masking seems like a good "additonal context", not a proposal. Should we put it in "Additional Context" section then?

docs/proposals-accepted/202206-block-series-deletion-compactor.md
```diff
@@ -0,0 +1,210 @@
...
+
+There are two approaches about how to scan block indexes:
+1. Download all the blocks locally and check against the tombstones.
+2. Use part of the store gateway code to scan the block index and chunks on the object storage to generate `MemTombstones` result.
+
+Approach 2 should be better as only block indexes need to be downloaded, chunks data are reading from the object storage on the fly so disk space can be saved. As compactor is an offline component, it is acceptable to read from the object storage rather than downloading the whole data.
+
+The pitfall is that we need to expose part of the store gateway code like `BucketIndexReader`, `BucketChunkReader` as well as other internal code to be used by the compactor.
```
bwplotka, line 137: That's fine - it's part of the same repo.

docs/proposals-accepted/202206-block-series-deletion-compactor.md
```diff
@@ -0,0 +1,210 @@
...
+
+#### Option 2: New API to Match Tombstones
+
+Since the block indexes are already downloaded by the store gateway, it can already scan the blocks and calculate the tombstones. So a way here is to expose some APIs from store gateway, and then compactor can query the APIs to know what blocks can be deleted.
```
bwplotka, line 141: It's worth to mention that this only works if Store overreach all blocks you want to work with.

docs/proposals-accepted/202206-block-series-deletion-compactor.md
```diff
@@ -0,0 +1,210 @@
...
+
+There are two approaches about how to scan block indexes:
+1. Download all the blocks locally and check against the tombstones.
+2. Use part of the store gateway code to scan the block index and chunks on the object storage to generate `MemTombstones` result.
+
+Approach 2 should be better as only block indexes need to be downloaded, chunks data are reading from the object storage on the fly so disk space can be saved. As compactor is an offline component, it is acceptable to read from the object storage rather than downloading the whole data.
+
+The pitfall is that we need to expose part of the store gateway code like `BucketIndexReader`, `BucketChunkReader` as well as other internal code to be used by the compactor.
+
+#### Option 2: New API to Match Tombstones
```
bwplotka, line 139: I like it - it feels like low-effort clear API to me 👍🏽

docs/proposals-accepted/202206-block-series-deletion-compactor.md
```diff
@@ -0,0 +1,210 @@
...
+
+```protobuf
+service Tombstone {
+  rpc Tombstone(TombstoneRequest) returns (TombstoneResponse);
+}
+
+message TombstoneRequest {
+  int64 min_time = 1;
+  int64 max_time = 2;
+  repeated LabelMatcher matchers = 3 [(gogoproto.nullable) = false];
```
bwplotka, line 155: In some way, compactor is responsible only for some blocks (some streams). Should we perhaps give that info in requests? (maybe I am overcomplicating)?

docs/proposals-accepted/202206-block-series-deletion-compactor.md
```diff
@@ -0,0 +1,210 @@
...
+  int64 start = 1;
+  int64 end = 2;
+}
+```
+
+After getting the query result, the compactor does almost the same thing as the previous option: generate the tombstone file during compaction time.
+
+The drawback for this approach is that we are inventing a new API for this purpose only, which might be a bit of pain in the future.
+
+#### Option 3: Delete Series using Rewrite Logic
```
bwplotka, line 177:
```suggestion
#### Option 3: Delete Series using `thanos tools block rewrite` CLI
```

docs/proposals-accepted/202206-block-series-deletion-compactor.md
```diff
@@ -0,0 +1,210 @@
...
+* Rewrite blocks separately. This needs more coordination with compaction to avoid duplicate/unnecessary work.
+
+To overcome those challenges, we propose two options.
+
+#### Option 1: Compactor Scans Blocks
+
+Like store gateway, compactor can also scan block indexes to find the blocks matching the tombstones. Similar to the `Mask Series At Tombstone Sync Time` of store gateway, compactor can calculate the processed tombstones for each block, each tombstone and cache the results locally. As blocks and tombstones are immutable, the result needs to be calculated only once and then reused easily. As the Prometheus format tombstones are calculated, during the compaction time the compactor just needs to generate the tombstone file in the filesystem and the TSDB compactor will apply the deletion.
+
+There are two approaches about how to scan block indexes:
+1. Download all the blocks locally and check against the tombstones.
```
bwplotka, line 132: We cannot download all blocks locally - we might not have enough disk for this. Should we instead scan one by one periodically? 

Would be nice to just think now how often we sync and what's the pseudoalgorithm for this. This would give us some idea about cost and delays - and tell us if we can start with option 1 or we should move to option 2 straight away 🤗

docs/proposals-accepted/202206-block-series-deletion-compactor.md
bwplotka: I think we could propose it as an optimization if the option 1 will opt 1 will proven too be slow / too expensive in terms of CPU/mem/objstorage access

bwplotka (dismissed):
Great job! Some comments @yeya24 

Thanks! Great ideas.

@GiedriusS @kakkoyun @matej-g @wiardvanrij @brancz @fpetkovski  any preferences on options here?

docs/proposals-accepted/202206-block-series-deletion-compactor.md
```diff
@@ -0,0 +1,210 @@
...
+
+The file format is as follows:
+``` bash
+cat 01G61GC0Q957TVZ2FC8S6Y2ZK7.json
+{
+        "ulid": "01G61GC0Q957TVZ2FC8S6Y2ZK7",
+        "matchers": [
+                "__name__=\"k8s_app_metric39\"",
+                "cluster=\"us\"",
+                "__block_id=\"01G61D0Q4BHTRC7HW08YRD0KGS\""
```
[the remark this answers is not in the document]
yeya24: Yes. This is something I added. But block ID is an optional field, which means users can choose to not add it and then the matchers can match all blocks. If they choose to add it then they can delete the specific block they want to. (Even multiple blocks using =~)

docs/proposals-accepted/202206-block-series-deletion-compactor.md
```diff
@@ -0,0 +1,210 @@
...
+        ],
+        "minTime": -62167219200000,
+        "maxTime": 253402300799000,
+        "author": "Ben Ye",
+        "reason": "not specified"
+}
+```
+
+Similar to the block `meta.json` file, the tombstone file is named using ULID which is basically its creation time. For other fields we have:
+* **label matchers**: label matchers to match the block and series. Like this example, blocks can be matched using external labels and blocks can be matched using a special `__block_id` label. All other labels are applied to series. This design makes it easy to support multi-tenant environment and match specific blocks.
```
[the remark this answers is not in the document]
yeya24: My thought is to use `tombstone CLI` and has an optional `block id` flag to do equal match. If `block id` is specified multiple times then we make it `__block_id=~(block1|block2|block3)`

docs/proposals-accepted/202206-block-series-deletion-compactor.md
```diff
@@ -0,0 +1,210 @@
...
+* **start timestamp**
+* **end timestamp**: end timestamp should be the tombstone creation time if not specified, otherwise the compactor needs to wait for a long time to make sure all matched blocks are processed. Considering the tombstone pending period, the end timestamp can be set to `(now + pending period)` as well. We don't consider history blocks backfilling now.
+* **author**
+* **reason**
+
+### Store Gateway Masks Series based on Tombstones
```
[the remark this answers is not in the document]
yeya24: Yes, it was implemented in https://github.com/thanos-io/thanos/pull/3075 but hasn't been finished. AFAIU, we agree on the high level idea of the initial proposal but for the implementation details, since the pr is not merged I won't consider it as implemented.

docs/proposals-accepted/202206-block-series-deletion-compactor.md
```diff
@@ -0,0 +1,210 @@
...
+Then during the query time, it checks the series label sets against the tombstone matchers, as well as the chunks and tombstone time range to mask the series data.
+
+Pros:
+* Easy to implement and very straightforward
+* Less memory usage.
+
+Cons:
+* Slower at query time
+
+#### Option 2: Mask Series At Tombstone Sync Time
```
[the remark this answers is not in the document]
yeya24: You are right. 🤣 Maybe I shouldn't talk about the series masking thing as the two options are almost the same, just the implementation details.

docs/proposals-accepted/202206-block-series-deletion-compactor.md
yeya24: Yeah you are totally right, I don't have any idea about this so far but if we go with the API direction we definitely need to design this.
Maybe an optional block ID array can be added to only generate tombstones for those blocks if matched? Or we can add block IDs to matchers.

docs/proposals-accepted/202206-block-series-deletion-compactor.md
yeya24: As mentioned below, compactors talk to queriers and queriers federate responses from stores. This way should work but a little bit complicated...

docs/proposals-accepted/202206-block-series-deletion-compactor.md
```diff
@@ -0,0 +1,210 @@
...
+
+## Non-Goals
+
+* Allow setting different retentions per series. This feature can be built on top of this work ideally but it is out of scope of this proposal.
+
+## How
+
+The main workflow of the series deletion will look like this:
+
+1. User create a tombstone file via a CLI tool and the tombstone file is uploaded to the object storage.
```
fpetkovski, line 29: Will the CLI do this, or should it be automated through a CI pipeline?

docs/proposals-accepted/202206-block-series-deletion-compactor.md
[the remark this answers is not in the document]
fpetkovski: Can we use something like the prometheus relabel configs, or is that too complicated for now?

docs/proposals-accepted/202206-block-series-deletion-compactor.md
```diff
@@ -0,0 +1,210 @@
...
+2. Store Gateway syncs tombstone files periodically and makes the block series based on the tombstones.
+3. Tombstone files have a pending period, which can be either specified by each tombstone or configured globally. Before the pending period is over, the tombstone file will remain in the object storage and user can remove a tombstone file using the CLI tool. Once the pending period is over, the compactor starts processing the tombstones and performs deletion.
```
fpetkovski, line 31: Is the `pending period` similar to the consistency delay in compactor/store-gw?

docs/proposals-accepted/202206-block-series-deletion-compactor.md
```diff
@@ -0,0 +1,210 @@
...
+Since all the blocks and tombstones are immutable, this result can be easily cached in memory or in filesystem. Ideally for each tombstone, each block we just need to calculate once. Since `BlockTombstones` is a map, it is easy to add or delete one key if users want to add more tombstones or remove one tombstone.
+
+During the query time, the cached tombstone will be used. As it knows which `SeriesRef`, which intervals to apply, it is very straightforward to mask series.
+
+Pros:
+* More efficient during query time, less processing required.
+* As data can be cached there is not much computation required.
+
+Cons:
+* More memory required.
```
fpetkovski, line 106: Would it be possible to use the filesystem instead, similar to how store-gw downloads the index file?

docs/proposals-accepted/202206-block-series-deletion-compactor.md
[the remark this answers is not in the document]
fpetkovski: Implementing removal in compactor is already going providing value, so I agree that masking can be done as an iterative improvement.

docs/proposals-accepted/202206-block-series-deletion-compactor.md
```diff
@@ -0,0 +1,210 @@
...
+* **Owners:**
+    * @yeya24
+
+## Summary
+
+This design document continues the feature of series deletion for object storage discussed in [proposal](https://thanos.io/tip/proposals-accepted/202012-deletions-object-storage.md/). This proposal is based on the previous work and focuses more on the actual implementation plan and the unaddressed questions from the original proposal like performing the actual deletion in the object storage using Thanos compactor.
+
+## Goals
+
+* Make it more clear about the main workflow of the series deletion process.
```
matej-g, line 18: nit:

```suggestion
* Make the main workflow of the series deletion process clearer.
```

docs/proposals-accepted/202206-block-series-deletion-compactor.md
```diff
@@ -0,0 +1,210 @@
...
+
+After getting the query result, the compactor does almost the same thing as the previous option: generate the tombstone file during compaction time.
+
+The drawback for this approach is that we are inventing a new API for this purpose only, which might be a bit of pain in the future.
+
+#### Option 3: Delete Series using Rewrite Logic
+
+The previous two options do the series deletion at compaction time, this option does actual series deletion using block rewrite. For the index scanning, it is still required to choose either option 1 or 2 to avoid unncessary rewrite even if the block doesn't have any matched series.
+
+This is not so efficient as rewrite is a different process than compaction. We need to add this step and make sure rewrite can work together with compaction and no redundant work is done.
```
matej-g, line 181: nit:
```suggestion
This is not efficient as rewrite is a different process from compaction. We need to add this step and make sure rewrite can work together with compaction and no redundant work is done.
```

docs/proposals-accepted/202206-block-series-deletion-compactor.md
```diff
@@ -0,0 +1,210 @@
...
+
+#### Considerations
+
+Overall, among the 3 options, option 1 is better as it is easier to implement and straightforward. No additional API required or no additional logic change on the compaction side.
+
+### Other Design Considerations
+
+#### Compaction Planner for Tombstone
+
+If series deletion is done at compaction time, then no much thing to change at the planner as we just need to delete series at compaction time. For the ratio calculation, as the tombstone results are pre-calculated it is easy to know the ratio as well. The threshold can be made configurable too.
```
matej-g, line 191: nit:
```suggestion
If series deletion is done at compaction time, then there's not much to change about the planner, as we just need to delete series at compaction time. For the ratio calculation, as the tombstone results are pre-calculated, it is easy to know the ratio as well. The threshold can be made configurable too.
```

docs/proposals-accepted/202206-block-series-deletion-compactor.md
matej-g, line 31: What is the rationale for having a pending period?

yeya24: Users can undo the tombstone if they change their mind.

docs/proposals-accepted/202206-block-series-deletion-compactor.md
yeya24: Yes, since latency is not really bottleneck for compactor we can definitely store those data in the filesystem. But those files shouldn't be too large as it contains series ids and intervals only. To optimize the space further we can use bitmap or bloom filter.

docs/proposals-accepted/202206-block-series-deletion-compactor.md
yeya24: Yes the cli will upload the tombstone file to the bucket directly. Similar to other bucket tools

docs/proposals-accepted/202206-block-series-deletion-compactor.md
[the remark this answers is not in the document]
yeya24: Hi Filip, if I understand this correctly, the tombstone file is global in the bucket and each tombstone file can match all the blocks (if no external label marcher and block id marchers are specified).
So if we are talking about sharding, store gateway and compactor need to gather all tombstones to match their blocks. Their blocks are sharded already, but tombstones are not.

That might be an issue when sharded compactors are running. Because each compactor only has view for its own blocks, we don't know whether a tombstone file has been applied to all blocks or not. So some kind of compactor coordinations are needed. WDYT about this issue?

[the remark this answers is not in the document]
yeya24: If we want to shard on tombstones then we might need per block tombstone file…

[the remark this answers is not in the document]
fpetkovski: What I had in mind was to use relabel configs instead of explicit matchers to keep or drop individual series. But I guess that will increase the scope of the proposal to allow complete series rewriting in compactor, not just deletion. So something to keep in mind for the future, but too complicated for implementing in the first version.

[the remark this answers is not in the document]
yeya24: Oh I see what you meant. Sorry I was talking about different things... 
Similar idea has been mentioned in https://github.com/thanos-io/thanos/issues/4941. We supported this in `bucket rewrite tool` and I proposed similar things to Prometheus https://github.com/prometheus/prometheus/pull/9413 before.

But indeed this has a different scope. Maybe it is better to think about it when we are working on the per-series retention period feature on compactor https://github.com/thanos-io/thanos/issues/903.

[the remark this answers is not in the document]
bwplotka: Yes, we discussed this - we could add new field for tombstones to support rewrites at some point.

I like block ID thingy @yeya24  - we could add that. Just let's fix the sentence that it applies to all blocks, unless block ID is specified

docs/proposals-accepted/202206-block-series-deletion-compactor.md
```diff
@@ -0,0 +1,210 @@
...
+
+There are other questions about how we should perform the deletion along with compaction:
+* Series are deleted at normal compaction. This might need to wait a long time and hard to apply for blocks at the max compaction level.
+* Rewrite blocks separately. This needs more coordination with compaction to avoid duplicate/unnecessary work.
+
+To overcome those challenges, we propose two options.
+
+#### Option 1: Compactor Scans Blocks
+
+Like store gateway, compactor can also scan block indexes to find the blocks matching the tombstones. Similar to the `Mask Series At Tombstone Sync Time` of store gateway, compactor can calculate the processed tombstones for each block, each tombstone and cache the results locally. As blocks and tombstones are immutable, the result needs to be calculated only once and then reused easily. As the Prometheus format tombstones are calculated, during the compaction time the compactor just needs to generate the tombstone file in the filesystem and the TSDB compactor will apply the deletion.
```
ilangofman, line 129: If there is a block that has series that needs to be deleted, the block is going to be compacted again? But this time its only one block being sent to the compactor with an additional tombstone file? Am I understanding correctly?

docs/proposals-accepted/202206-block-series-deletion-compactor.md
```diff
@@ -0,0 +1,210 @@
...
+
+Overall, among the 3 options, option 1 is better as it is easier to implement and straightforward. No additional API required or no additional logic change on the compaction side.
+
+### Other Design Considerations
+
+#### Compaction Planner for Tombstone
+
+If series deletion is done at compaction time, then no much thing to change at the planner as we just need to delete series at compaction time. For the ratio calculation, as the tombstone results are pre-calculated it is easy to know the ratio as well. The threshold can be made configurable too.
+
+The exception is for blocks that at max compaction level, no compaction will happen to it and the tombstones will remain forever if the ratio is lower than the deletion threshold. In this case the planner can check the block compaction level first, if it is max compaction level and the tombstone exists for some period of time, then do compaction for that block.
```
ilangofman, line 193: Is the ratio going to be calculated per each tombstone or based on the total of all tombstones that can act on a particular block? Asking because even if one tombstone doesn't reach above the threshold, the sum of multiple smaller tombstones might.

docs/proposals-accepted/202206-block-series-deletion-compactor.md
yeya24: What do you mean by compacted again?
This only happens once and tsdb compaction by default takes care of series deletion if there is any tombstones.

But this only applies to blocks that don't reach compaction max level. For max level block we need to schedule separate compaction for them in the planner.

docs/proposals-accepted/202206-block-series-deletion-compactor.md
yeya24: It is for all tombstones. We calculate the number of series to be deleted for all tombstones that match this block.

docs/proposals-accepted/202206-block-series-deletion-compactor.md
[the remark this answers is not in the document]
yeya24: In my experience, sometimes the compactor halts due to some out-of-order chunks from some blocks. Then I can just delete those series using the given block IDs.
Other than that, the use case seems rare.

docs/proposals-accepted/202206-block-series-deletion-compactor.md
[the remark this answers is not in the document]
yeya24: Thanks!

docs/proposals-accepted/202206-block-series-deletion-compactor.md
yeya24: Thanks! Updated.

docs/proposals-accepted/202206-block-series-deletion-compactor.md
yeya24: Thanks! Updated.

docs/proposals-accepted/202206-block-series-deletion-compactor.md
```diff
@@ -0,0 +1,210 @@
...
+
+### Store Gateway Masks Series based on Tombstones
+
+Like block metadata, the store gateway also fetches tombstones in a configurable interval, which means some delay is expected: user creates a tombstone but she is still able to query the data to be deleted.
+
+There are different ways to mask series based on tombstones. We will discuss two options here:
+
+#### Option 1: Mask Series At Query Time
+
+During query time, the store gateway decides what blocks to query based on the request min, max time and label matchers. Meanwhile, it can also decide what tombstones are going to apply for the selected blocks based on tombstone time range and label matchers.
```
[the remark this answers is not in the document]
yeya24: Thanks! Updated.

docs/proposals-accepted/202206-block-series-deletion-compactor.md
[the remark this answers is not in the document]
yeya24: > ust let's fix the sentence that it applies to all blocks, unless block ID is specified

The matchers support matching external labels, too for multi-tenancy purposes. I think we can still call it global and it applies to all blocks. Just we can tell that the matchers won't match some blocks so we just filter them out. Like how we do requests filtering on store gateway.

docs/proposals-accepted/202206-block-series-deletion-compactor.md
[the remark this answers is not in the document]
yeya24: Masking was proposed in the previous proposal. I think it would be better to go back and discuss masking there in the old proposal... Instead of talking about it here. I am not sure if it is okay to modify the old one.

docs/proposals-accepted/202206-block-series-deletion-compactor.md
```diff
@@ -0,0 +1,219 @@
...
+
+Pros:
+* More efficient during query time, less processing required.
+* As data can be cached there is not much computation required.
+
+Cons:
+* More memory required.
+* If read is much less than write, it is inefficient as the pre-calculated result won't be used much.
+
+##### Considerations
```
bwplotka, line 194: What this means? We already added `### Proposed Option"`.

Is this section still relevant?

matej-g:
Do we have any updates here @yeya24?

yeya24:
> Do we have any updates here @yeya24?

I am still looking for feedbacks from the community. I've already mentioned the approaches we could take so once we agree the proposal itself is good, we can merge and I'll open a pr.
As listed in the PR description I already implemented one approach so feel free to take a look.

paprickar:
Hey @yeya24,

I was looking for a solution to delete only specific time series (similar to what we can do with the thanos tool rewrite command "matcher").

I saw settings retention via compactor but this is globally only.

Your proposal would allow deletion of specific time series but only on demand if I understand it correct.

Therefore could we maybe extend the JSON to allow the "retention" use case?

maxAge: time in seconds how long you want to keep a metrics from Time.now()
keep: if compactor should delete the tombstone file or not

`{
        "ulid": "01G61GC0Q957TVZ2FC8S6Y2ZK7",
        "matchers": [
                "__name__=\"k8s_app_metric39\"",
                "cluster=\"us\"",
                "__block_id=\"01G61D0Q4BHTRC7HW08YRD0KGS\""
        ],
        "maxAge": 86400 ,
        "keep": true,
        "author": "Ben Ye",
        "reason": "not specified"
}`

paprickar:
@yeya24 @matej-g does one of you maybe have a status on this topic in general maybe?

PR #5440, opened 2022-06-23
Add HTTP metrics for in-flight requests

douglascamata:

## Changes

This was split off of #5424.

* Add a new metric called `http_inflight_requests` in the default handler instrumentation to measure the amount of inflight HTTP requests.

## Verification

Manual testing.


CHANGELOG.md
```diff
@@ -31,9 +31,10 @@ We use *breaking :warning:* to mark changes that are not backward compatible (re
 - [#5409](https://github.com/thanos-io/thanos/pull/5409) S3: Add option to force DNS style lookup.
 - [#5352](https://github.com/thanos-io/thanos/pull/5352) Cache: Add cache metrics to groupcache.
 - [#5391](https://github.com/thanos-io/thanos/pull/5391) Receive: Add relabeling support.
-- [#5408](https://github.com/thanos-io/thanos/pull/5391) Receive: Add support for consistent hashrings.
+- [#5408](https://github.com/thanos-io/thanos/pull/5408) Receive: Add support for consistent hashrings.
 - [#5391](https://github.com/thanos-io/thanos/pull/5391) Receive: Implement api/v1/status/tsdb.
 - [#5424](https://github.com/thanos-io/thanos/pull/5424) Receive: export metrics regarding size of remote write requests
+- [#5440](https://github.com/thanos-io/thanos/pull/5440) HTTP metrics: export number of in-flight HTTP requests
```
jessicalins, line 37: super nit but I think the pattern seems to be to end the sentence with `.` as in the other changelog entries?

pkg/extprom/http/metrics.go
```diff
@@ -29,26 +30,37 @@ func newDefaultMetrics(reg prometheus.Registerer, buckets []float64, extraLabels
...
 				Name: "http_response_size_bytes",
 				Help: "Tracks the size of HTTP responses.",
 			},
 			append([]string{"code", "handler", "method"}, extraLabels...),
 		),
+
+		inflightHTTPRequests: promauto.With(reg).NewGaugeVec(
+			prometheus.GaugeOpts{
+				Name: "http_inflight_requests",
+				Help: "Current number of active requests.",
```
jessicalins, line 61: maybe we could add a more specific description, something like:
```suggestion
				Help: "Current gauge of in-flight requests.",
```
wdyt?

jessicalins (commented):
cool! :star:

pkg/extprom/http/metrics.go
douglascamata: I think it's redundant to describe it as a "gauge" in the help, because counters should have the postfix `_total` as per [Prometheus' naming conventions](https://prometheus.io/docs/practices/naming/). If a metric's name doesn't end with `_total`, it should be or behave like a gauge.

To follow up, I don't think the help should say exactly the same as the metric name. Otherwise the help is not helpful. So I thought, "maybe someone will not know exactly what we mean with 'in-flight' and 'active' is a more commonly used word".

CHANGELOG.md
douglascamata: Fixed this and other changelog entries lacking the period in [7db4419](https://github.com/thanos-io/thanos/pull/5440/commits/7db44191909a20624bf98acf1f2f679455944ac7).

pkg/extprom/http/metrics.go
jessicalins: I think when the metric ends with `_total` indeed it should indicate a counter, but if it does not end with `_total` not necessarily means that it is a gauge, as far as I understand (for [example](https://github.com/prometheus/prometheus/blob/01c55822169a9123367e333dc407edd8a263f75d/discovery/kubernetes/client_metrics.go))
I've seen other examples where also the name of the metric repeats its type in the help message (e.g. [here](https://promlabs.com/blog/2020/09/25/metric-types-in-prometheus-and-promql#histogram-metrics)) but not necessarily means it should be like that.

I think would be ok keeping `number` - though regarding `active` requests for me reads a bit more unclear, since I wouldn't know what the definition of an `active` request would be.

saswatamcode: Maybe it would be a good idea to define "in-flight" within the help message? 🙂

douglascamata: @jessicalins the metrics in the reference you sent that aren't explicitly a gauge are summaries/histograms. They will though generate gauges/counters with the `_sum` and `_total` (respectively) postfixes under the hood.

I improved the help text for the in-flight HTTP requests metrics in 
7f0adc4. 👍

douglascamata:
@jessicalins @saswatamcode: please take another look at the PR whenever you can.

saswatamcode (approved):

jessicalins (dismissed):
:rocket:

CHANGELOG.md
```diff
@@ -21,20 +21,21 @@ We use *breaking :warning:* to mark changes that are not backward compatible (re
...
 - [#5409](https://github.com/thanos-io/thanos/pull/5409) S3: Add option to force DNS style lookup.
 - [#5352](https://github.com/thanos-io/thanos/pull/5352) Cache: Add cache metrics to groupcache.
 - [#5391](https://github.com/thanos-io/thanos/pull/5391) Receive: Add relabeling support.
-- [#5408](https://github.com/thanos-io/thanos/pull/5391) Receive: Add support for consistent hashrings.
+- [#5408](https://github.com/thanos-io/thanos/pull/5408) Receive: Add support for consistent hashrings.
 - [#5391](https://github.com/thanos-io/thanos/pull/5391) Receive: Implement api/v1/status/tsdb.
-- [#5424](https://github.com/thanos-io/thanos/pull/5424) Receive: export metrics regarding size of remote write requests
+- [#5424](https://github.com/thanos-io/thanos/pull/5424) Receive: export metrics regarding size of remote write requests.
 - [#5420](https://github.com/thanos-io/thanos/pull/5420) Receive: Automatically remove stale tenants.
+- [#5440](https://github.com/thanos-io/thanos/pull/5440) HTTP metrics: export number of in-flight HTTP requests.
```
matej-g, line 38: This new entry should be above in the `Unreleased` section, since the RC is already out.

douglascamata: Thanks for noticing. Probably I messed up fixing a merge conflict with `main`. 🤦 

Fixed in 67edcf2.

matej-g (dismissed):
Looking good! :star: And thanks for the changelog fixes!

douglascamata:
@matej-g can you merge this, please? 🙏 

Or we are waiting for other official members of the team to review?

pkg/extprom/http/metrics.go
```diff
@@ -29,26 +30,37 @@ func newDefaultMetrics(reg prometheus.Registerer, buckets []float64, extraLabels
...
 			},
 			append([]string{"code", "handler", "method"}, extraLabels...),
 		),
+
+		inflightHTTPRequests: promauto.With(reg).NewGaugeVec(
+			prometheus.GaugeOpts{
+				Name: "http_inflight_requests",
+				Help: "Current number of HTTP requests the handler is responding to.",
+			},
+			append([]string{"handler"}, extraLabels...),
```
squat, line 63: All of the other metrics also consider a `method` label. Is there any reason not to collect that label here as well?

bwplotka (approved):
LGTM thanks for the work and reviews everyone! 💪🏽

bwplotka:
Hm, indeed lint error is weird. Something easy we can check is rebase to latest?

PR #5441, opened 2022-06-24
Scripts: Add quickstart script for alert generator compliance test

matej-g:

## Changes

We have recently introduced the alert generator compliance test via E2E (see https://github.com/thanos-io/thanos/pull/5315). While this is useful for being able to run the test as part of our E2E suite, I want to propose also adding a separate script to spin up all components for the test run. Couple of reasons why:
- We want to provide example configuration in the [alert generator compliance repo](https://github.com/prometheus/compliance/tree/main/alert_generator) to signal to users that Thanos is compliant. But because the E2E is running inside Docker, the configuration we have in the test is transient and cannot be easily made into a 'runnable' example config
- Users do not need to look within our test files (and having to uncomment the skip call etc.), all is spun up with one script
- Users do not even have to have Go installed (as opposed to E2E test), they can directly run this test with virtually no extra hurdles
- Test is more transparent in a sense that users can observe the behavior (look at the web UIs) of the components, since these are running on the host machine (as opposed to containers without exposed ports).

## Verification
Tested the alert generator compliance suite with the script on my local machine.

### DIFF
--- a/scripts/alert-compliance-quickstart.sh
+++ b/scripts/alert-compliance-quickstart.sh
@@ -10,7 +10,7 @@
 # For the test you will need to run Thanos binary built with the latest 'main'
 # branch or with version >= 0.27.0.
 #
-# After all comopnents are running, you can start the alert generator compliance tester
+# After all components are running, you can start the alert generator compliance tester
 # with `thanos-example.yaml`` configuration provided in here:
 # https://github.com/prometheus/compliance/blob/main/alert_generator/test-prometheus.yaml
 set -euo pipefail
--- a/scripts/alert-compliance-quickstart.sh
+++ b/scripts/alert-compliance-quickstart.sh
@@ -25,16 +25,16 @@ export ALERT_COMPLIANCE_RULES
 curl -sNL -o "${ALERT_COMPLIANCE_RULES}" "https://raw.githubusercontent.com/prometheus/compliance/main/alert_generator/rules.yaml"
 
 ${THANOS_EXECUTABLE} receive \
-  --label 'receive_replica="0"' \
+  --label='receive_replica="0"' \
   --tsdb.path="${TMP_DATA}" &
 
 # We make sure to filter out the 'receive_replica' and 'tenant_id' labels,
 # which are added by the receiver (they cannot be present during the test).
 ${THANOS_EXECUTABLE} query \
-  --http-address 0.0.0.0:19192 \
-  --store 0.0.0.0:10901 \
-  --rule 0.0.0.0:20901 \
-  --grpc-address 0.0.0.0:19099 \
+  --http-address=0.0.0.0:19192 \
+  --store=0.0.0.0:10901 \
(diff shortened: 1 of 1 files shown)


saswatamcode (approved):
Thanks! Awesome work! 🌟

scripts/alert-compliance-quickstart.sh
```diff
@@ -0,0 +1,58 @@
+#!/bin/bash
+#
+# This is a script for running the Prometheus Alert Generator Compliance
+# test suite. Read more at: https://github.com/prometheus/compliance/tree/main/alert_generator
+#
+# It is the most minimal setup you can use for testing on a local machine.
+# The script will start all necessary components (receive, ruler, querier)
+# with appropriate confguration.
+#
+# For the test you will need to run Thanos binary built with the latest 'main'
```
jessicalins, line 10: :+1:

jessicalins (commented):
cool! :sparkle: 
I tested the script locally and it seems to work fine :)
left a small suggestion

jessicalins (dismissed):
LGTM, awesome work :tada: 
I think last thing would be to fix the commit [sign off](https://github.com/thanos-io/thanos/pull/5441/checks?check_run_id=7113930198)

matej-g:
@bwplotka those are good counterarguments (I also do not want to push this change at all costs, as it is extra stuff to maintain, but still thought it would be a nice suggestion). To give more context how I was thinking about this:
- For a developer, the E2E tests are useful: I make some changes and I want to make sure we are still compliant = great, I can run this E2E test
- For any other user (e.g. anyone who just wants to run / play with Thanos) and they want to see for themselves that we are complian (before they decide to use us vs. other solution, for example): They don't need to clone the repo (need just one script), they don't need Go or build Thanos (using latest release is fine), they should simply be able to run the script + alert compliance tester and be able to immediately and transparently see we are compliant.

I think I look at the compliance to be more for users to be able to transparently verify we comply with the official specification, more than for developers (but who also need it if they make relevant change).

> Perhaps adding this as example script to https://github.com/prometheus/compliance/blob/main/alert_generator and our file as a comment makes sense?
I was thinking about that as well, and we could try if the compliance suite maintainers are fine with it, it's just quiet a few lines so it seemed liked it would be better if it's separate script. But we could try this alternative as well.

> Adding steps or script that runs compliance go test. Remove Skip and put a special build tag that triggers the test?
That would be fine as well, although users would still need to clone the repo, have Go etc. (point I addressed above, but maybe it's not that strong of a hurdle).

bwplotka (commented):
Don't want to block this, it's little harm in having the extra script, just it might get obsolete pretty soon.

matej-g:
I don't think we have a strong enough agreement on this, I don't want to introduce something that might not be really needed / might become obsolete fast. I'd rather close this and if we decide to introduce the script, whether here or in the compliance repo, we can reuse this PR :+1:.

Issue #5272, opened 2022-04-08
Thanos compactor failing to start with first pass of downsampling failed error


infa-mlagad:
**Thanos, Prometheus and Golang version used**:
Thanos image version - v.0.25.1

**Object Storage Provider**:
AWS

**What happened**:
Thanos compactor going in crashlooping state. Below error in the logs.
`level=warn ts=2022-04-08T06:41:18.018546402Z caller=intrumentation.go:67 msg="changing probe status" status=not-ready reason="error executing compaction: first pass of downsampling failed: downsampling to 60 min: downsample block xxxx to window 3600000: expected downsampled chunk (*downsample.AggrChunk) got *chunkenc.XORChunk instead for series: 6566"
level=info ts=2022-04-08T06:41:18.018588894Z caller=http.go:84 service=http/server component=compact msg="internal server is shutting down" err="error executing compaction: first pass of downsampling failed: downsampling to 60 min: downsample block xxxx to window 3600000: expected downsampled chunk (*downsample.AggrChunk) got *chunkenc.XORChunk instead for series: 6566"
level=info ts=2022-04-08T06:41:18.019725719Z caller=http.go:103 service=http/server component=compact msg="internal server is shutdown gracefully" err="error executing compaction: first pass of downsampling failed: downsampling to 60 min: downsample block xxxx to window 3600000: expected downsampled chunk (*downsample.AggrChunk) got *chunkenc.XORChunk instead for series: 6566"
level=info ts=2022-04-08T06:41:18.019767702Z caller=intrumentation.go:81 msg="changing probe status" status=not-healthy reason="error executing compaction: first pass of downsampling failed: downsampling to 60 min: downsample block xxxx to window 3600000: expected downsampled chunk (*downsample.AggrChunk) got *chunkenc.XORChunk instead for series: 6566"
[...]

PR #5453, opened 2022-06-30
During 1h downsampling skip XOR chunks that may erroneously be present in 5m resolution blocks

eharcevs:
Fixes https://github.com/thanos-io/thanos/issues/5272

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -21,7 +21,7 @@ We use *breaking :warning:* to mark changes that are not backward compatible (re
 ## [v0.27.0-rc0](https://github.com/thanos-io/thanos/tree/release-0.27) - Release in progress
 
 ### Fixed
-- [#5453](https://github.com/thanos-io/thanos/pull/5453) Compact: Fix crash during 1h downsampling of 5m resolution blocks
+- [#5453](https://github.com/thanos-io/thanos/pull/5453) Compact: Skip erroneous empty non `*AggrChunk` chunks during 1h downsampling of 5m resolution blocks.
 - [#5339](https://github.com/thanos-io/thanos/pull/5339) Receive: Fix deadlock on interrupt in routerOnly mode
 - [#5357](https://github.com/thanos-io/thanos/pull/5357) Store: fix groupcache handling of slashes
 - [#5427](https://github.com/thanos-io/thanos/pull/5427) Receive: Fix Ketama hashring replication consistency
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -21,6 +21,7 @@ We use *breaking :warning:* to mark changes that are not backward compatible (re
 ## [v0.27.0-rc0](https://github.com/thanos-io/thanos/tree/release-0.27) - Release in progress
 
 ### Fixed
+- [#5453](https://github.com/thanos-io/thanos/pull/5453) Compact: Fix crash during 1h downsampling of 5m resolution blocks
 - [#5339](https://github.com/thanos-io/thanos/pull/5339) Receive: Fix deadlock on interrupt in routerOnly mode
 - [#5357](https://github.com/thanos-io/thanos/pull/5357) Store: fix groupcache handling of slashes
 - [#5427](https://github.com/thanos-io/thanos/pull/5427) Receive: Fix Ketama hashring replication consistency
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -9,6 +9,7 @@ NOTE: As semantic versioning states all 0.y.z releases can contain breaking chan
 We use *breaking :warning:* to mark changes that are not backward compatible (relates only to v0.y.z releases.)
 
 ## Unreleased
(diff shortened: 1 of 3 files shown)


pkg/compact/downsample/downsample.go
```diff
@@ -153,9 +155,12 @@ func Downsample(
 			for _, c := range chks {
 				ac, ok := c.Chunk.(*AggrChunk)
 				if !ok {
-					return id, errors.Errorf("expected downsampled chunk (*downsample.AggrChunk) got %T instead for series: %d", c.Chunk, postings.At())
+					// Downsampled block can erroneously contain XOR chunks, skip those
```
bwplotka, line 158: Do you mind checking if it's empty XOR? This will allow us to be sure! Thanks 💪🏽

eharcevs: Thanks, adjusted as suggested, please review!

pkg/compact/downsample/downsample.go
```diff
@@ -153,9 +155,16 @@ func Downsample(
 			for _, c := range chks {
 				ac, ok := c.Chunk.(*AggrChunk)
 				if !ok {
-					return id, errors.Errorf("expected downsampled chunk (*downsample.AggrChunk) got %T instead for series: %d", c.Chunk, postings.At())
+					if c.Chunk.NumSamples() == 0 {
+						// Downsampled block can erroneously contain empty XOR chunks, skip those
+						// https://github.com/thanos-io/thanos/issues/5272
+						level.Warn(logger).Log("msg", fmt.Sprintf("expected downsampled chunk (*downsample.AggrChunk) got an empty %T instead for series: %d", c.Chunk, postings.At()))
+					} else {
+						return id, errors.Errorf("expected downsampled chunk (*downsample.AggrChunk) got a non-empty %T instead for series: %d", c.Chunk, postings.At())
```
yeya24, line 163: Just wondering if we encounter a non-empty XOR chunk then an error is returned. But in this case, what can we do to repair this block currently? If there is no way to fix it can we have a flag to skip this chunk as well?

bwplotka: Yea, that's kind of by design - we want to ensure we catch problems instead of ignoring them. But yea, I would not oppose extra flag for it.

CHANGELOG.md
```diff
@@ -21,6 +21,7 @@ We use *breaking :warning:* to mark changes that are not backward compatible (re
 ## [v0.27.0-rc0](https://github.com/thanos-io/thanos/tree/release-0.27) - Release in progress
 
 ### Fixed
+- [#5453](https://github.com/thanos-io/thanos/pull/5453) Compact: Fix crash during 1h downsampling of 5m resolution blocks
```
bwplotka, line 24: * This should go to the "Unrelased" section
* Let's be obvious what we are actually adding, so:

```suggestion
- [#5453](https://github.com/thanos-io/thanos/pull/5453) Compact: Skip erroneous empty non `*AggrChunk` chunks during 1h downsampling of 5m resolution blocks.
```

pkg/compact/downsample/downsample.go
bwplotka: we could do that in next PR

pkg/compact/downsample/downsample.go
```diff
@@ -153,9 +155,16 @@ func Downsample(
 			for _, c := range chks {
 				ac, ok := c.Chunk.(*AggrChunk)
 				if !ok {
-					return id, errors.Errorf("expected downsampled chunk (*downsample.AggrChunk) got %T instead for series: %d", c.Chunk, postings.At())
+					if c.Chunk.NumSamples() == 0 {
+						// Downsampled block can erroneously contain empty XOR chunks, skip those
+						// https://github.com/thanos-io/thanos/issues/5272
+						level.Warn(logger).Log("msg", fmt.Sprintf("expected downsampled chunk (*downsample.AggrChunk) got an empty %T instead for series: %d", c.Chunk, postings.At()))
+					} else {
```
bwplotka, line 162: Let's avoid `else` and do `continue` here instead?

bwplotka (approved):
Small nits only, thanks!

pkg/compact/downsample/downsample.go
eharcevs: done, please review

pkg/compact/downsample/downsample.go
eharcevs: agree, thanks

yeya24 (commented):
Can you please sign off your commits? DCO check is falling. Others LGTM

yeya24 (approved):
LGTM

pkg/compact/downsample/downsample.go
yeya24: Yeah, let me have another issue to track this. I still want to have a way to fix it somehow in the repair tool because we cannot do anything to fix the error chunk now.

PR #5460, opened 2022-07-01
receive: Fix race condition when pruning tenants

fpetkovski:
Pruning Receiver tenants has a race condition caused by concurrently
removing items from the tenants map.

This commit addresses the issue by using a mutex to guard the tenants map.

Note: it might be a good idea to run unit tests with the `-race` flag in the CI.

## Changes

Address race condition when pruning stale tenants.

## Verification

```
go test ./pkg/receive/... -race -run TestMultiTSDBPrune -v
```

### DIFF
--- a/pkg/receive/multitsdb.go
+++ b/pkg/receive/multitsdb.go
@@ -215,6 +215,9 @@ func (t *MultiTSDB) Prune(ctx context.Context) error {
 	var (
 		wg   sync.WaitGroup
 		merr errutil.SyncMultiError
+
+		prunedTenants []string
+		pmtx          sync.Mutex
 	)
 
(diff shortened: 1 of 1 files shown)


fpetkovski:
I also noticed that we use a global lock on all tenants [here](https://github.com/thanos-io/thanos/pull/5460/files#diff-61eda56d07cd51eb9fde89c6bd9f1fe7e260edb05eb8fb4c1c81aa258f034331R212).

Do we want to optimize this path given that pruning is done relatively infrequently (each 2 hours)?

yeya24 (approved):
LGTM. Thanks for the quick fix!

fpetkovski:
Agreed, we run this in prod and doesn't seem to be an issue

PR #5468, opened 2022-07-04
Endpoint: Refactor get clients methods

### LINKED ISSUE #5458: Race condition in proxy store
Thanos version: thanosio/thanos:main-2022-06-23-f60b6598

**What happened**:
A Querier crashed several times with the following message

> Message:   east1-2\", replica=\"thanos-receiver-17\"},{k8s_cluster="[redacted]", replica=\"thanos-receiver-17\"},{k8s_cluster="[redacted]", replica=\"thanos-receiver-17\"},{k8s_cluster="[redacted]", replica=\"thanos-receiver-17\"},{k8s_cluster="[redacted]", replica=\"thanos-receiver-17\"},{k8s_cluster="[redacted]", replica=\"thanos-receiver-17\"},{k8s_cluster="[redacted]", replica=\"thanos-receiver-17\"},{k8s_cluster="[redacted]", replica=\"thanos-receiver-17\"},{k8s_cluster="[redacted]", replica=\"thanos-receiver-17\"},{k8s_cluster="[redacted]", replica=\"thanos-receiver-17\"},{k8s_cluster="[redacted]", replica=\"thanos-receiver-17\"},{k8s_cluster="[redacted]", replica=\"thanos-receiver-17\"},{k8s_cluster="[redacted]", replica=\"thanos-receiver-17\"},{k8s_cluster="[redacted]", replica=\"thanos-receiver-17\"},{k8s_cluster="[redacted]", replica=\"thanos-receiver-17\"},{k8s_cluster="[redacted]", replica=\"thanos-receiver-17\"},{k8s_cluster="[redacted]", replica=\"thanos-receiver-17\"} Mint: -9223372036854775808 Maxt: 9223372036854775807: rpc error: code = Canceled desc = grpc: the client connection is closing"
panic: runtime error: invalid memory address or nil pointer dereference
[signal SIGSEGV: segmentation violation code=0x1 addr=0x30 pc=0x158f678]

> goroutine 631657 [running]:
github.com/thanos-io/thanos/pkg/query.(*endpointRef).Series(0xc000cb9a40, {0x21fe8d0, 0xc00ba115f0}, 0xc, {0x0, 0x1, 0x1})
  <autogenerated>:1 +0x38
github.com/thanos-io/thanos/pkg/store.(*ProxyStore).Series.func1()
  /app/pkg/store/proxy.go:314 +0xf5f
golang.org/x/sync/errgroup.(*Group).Go.func1()
  /go/pkg/mod/golang.org/x/sync@v0.0.0-20210220032951-036812b2e83c/errgroup/errgroup.go:57 +0x67
created by golang.org/x/sync/errgroup.(*Group).Go
  /go/pkg/mod/golang.org/x/sync@v0.0.0-20210220032951-036812b2e83c/errgroup/errgroup.go:54 +0x92

I would expect Querier to handle the removal of a store gracefully while executing a query.

The issue is hard to reproduce locally since it seems to be caused by a race condition when a query is executed and a store is removed from the endpoints at the same time.

This is where the panic originates from
https://github.com/thanos-io/thanos/blob/f644d75586e60d86e2861cb5a9bceb7bee98188b/pkg/store/proxy.go#L314

matej-g:

## Changes
Upon reports from #5458 I looked at our endpoint set logic and I believe there is potential for buggy behavior with respect to clients.

How the code works now is that an `endpointRef` includes a `clients` struct, where we build the clients 'ahead of time', depending on components metadata and on which APIs it exposes. However, the metadata can change in some cases (i.e. for some components, the storage might or might not [be ready](https://github.com/thanos-io/thanos/blob/main/cmd/thanos/receive.go#L364) - depending on this probe, we're either returning store metadata or not). This in turn affects if the client is [created or removed.](https://github.com/thanos-io/thanos/blob/5f6ee25d7a8059632e09240068fe1a019cb47245/pkg/query/endpointset.go#L642)

I believe there's a race hidden. When a proxy store then calls [`Series()` method on the store clients](https://github.com/thanos-io/thanos/blob/main/pkg/store/proxy.go#L314), some clients might have been removed mid-query (i.e. set to `nil`), resulting in a panic (as I presume is happening in #5458). This is due to the fact that the store clients refer to the pre-built `clients` in `endpointRef`. But because there's no guarantee the client is there (since an update might have occurred while a proxy query was in progress), this might result in a panic.

The proposal here is to rather just keep the metadata as part of the `endpointRef` and build clients on-the-fly, depending on available metadata. This way, the client is always available for the duration of a whole proxy store operation (but if it's e.g. not active anymore, will not be included for the next proxy query).

Fixes (potentially) #5458.

## Verification

I wasn't not able to verify this is indeed problem with #5458, since it's not easy to reproduce, but it seems plausible to me. I'm also wondering if anyone has performance concerns doing it this way vs. 'pre-building' the clients on update.

### DIFF
--- a/pkg/query/endpointset.go
+++ b/pkg/query/endpointset.go
@@ -61,17 +61,17 @@ func (es *GRPCEndpointSpec) Addr() string {
 
 // Metadata method for gRPC endpoint tries to call InfoAPI exposed by Thanos components until context timeout. If we are unable to get metadata after
 // that time, we assume that the host is unhealthy and return error.
-func (es *GRPCEndpointSpec) Metadata(ctx context.Context, client *endpointClients) (*endpointMetadata, error) {
-	if client.info != nil {
-		resp, err := client.info.Info(ctx, &infopb.InfoRequest{}, grpc.WaitForReady(true))
+func (es *GRPCEndpointSpec) Metadata(ctx context.Context, infoClient infopb.InfoClient, storeClient storepb.StoreClient) (*endpointMetadata, error) {
+	if infoClient != nil {
+		resp, err := infoClient.Info(ctx, &infopb.InfoRequest{}, grpc.WaitForReady(true))
 		if err == nil {
 			return &endpointMetadata{resp}, nil
 		}
 	}
 
 	// Call Info method of StoreAPI, this way querier will be able to discovery old components not exposing InfoAPI.
-	if client.store != nil {
-		metadata, err := es.getMetadataUsingStoreAPI(ctx, client.store)
+	if storeClient != nil {
+		metadata, err := es.getMetadataUsingStoreAPI(ctx, storeClient)
 		if err != nil {
 			return nil, errors.Wrapf(err, "fallback fetching info from %s", es.addr)
 		}
@@ -368,7 +368,12 @@ func (e *EndpointSet) GetStoreClients() []store.Client {
 	stores := make([]store.Client, 0, len(e.endpoints))
 	for _, er := range e.endpoints {
 		if er.HasStoreAPI() {
-			stores = append(stores, er)
+			// Make a new endpointRef with store client.
+			stores = append(stores, &endpointRef{
+				StoreClient: storepb.NewStoreClient(er.cc),
+				addr:        er.addr,
+				metadata:    er.metadata,
+			})
 		}
 	}
 	return stores
@@ -382,7 +387,7 @@ func (e *EndpointSet) GetQueryAPIClients() []querypb.QueryClient {
 	stores := make([]querypb.QueryClient, 0, len(e.endpoints))
 	for _, er := range e.endpoints {
 		if er.HasQueryAPI() {
-			stores = append(stores, er.clients.query)
+			stores = append(stores, querypb.NewQueryClient(er.cc))
 		}
 	}
 	return stores
@@ -396,7 +401,7 @@ func (e *EndpointSet) GetRulesClients() []rulespb.RulesClient {
 	rules := make([]rulespb.RulesClient, 0, len(e.endpoints))
 	for _, er := range e.endpoints {
 		if er.HasRulesAPI() {
-			rules = append(rules, er.clients.rule)
+			rules = append(rules, rulespb.NewRulesClient(er.cc))
 		}
 	}
 	return rules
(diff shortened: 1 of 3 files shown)


fpetkovski:
Having immutable clients sounds good to me 👍

fpetkovski:
Hm I think there's some issue with the implementation though. After a while, rulers start failing queries with `No StoreAPIs matched for this query`

pkg/query/endpointset.go
```diff
@@ -61,17 +61,17 @@ func (es *GRPCEndpointSpec) Addr() string {
 
 // Metadata method for gRPC endpoint tries to call InfoAPI exposed by Thanos components until context timeout. If we are unable to get metadata after
 // that time, we assume that the host is unhealthy and return error.
-func (es *GRPCEndpointSpec) Metadata(ctx context.Context, client *endpointClients) (*endpointMetadata, error) {
-	if client.info != nil {
-		resp, err := client.info.Info(ctx, &infopb.InfoRequest{}, grpc.WaitForReady(true))
+func (es *GRPCEndpointSpec) Metadata(ctx context.Context, infoClient infopb.InfoClient, storeClient storepb.StoreClient) (*endpointMetadata, error) {
```
bwplotka, line 64:
```suggestion
func (es *GRPCEndpointSpec) Metadata(ctx context.Context, infoClient infopb.InfoClient, fallbackStoreClient storepb.StoreClient) (*endpointMetadata, error) {
```

bwplotka (approved):
cc @SrushtiSapkale and @saswatamcode as we plan to touch on this code in the near future.

LGTM, looks perfect.

Clients are cheap to create - single alloc only. Should not be a big deal and if it is we can optimize later.


image

bwplotka:
Hm 

> Hm I think there's some issue with the implementation though. After a while, rulers start failing queries with No StoreAPIs matched for this query

Wonder if that's related problem to this PR, worth do more debugging. I don't see relation from code point of view 🤔

fpetkovski:
Yes, I can test again tomorrow and try to do some troubleshooting/debugging.

matej-g:
I'm also wondering what would be the case. Tests are passing fine and I even ran the long-running alerting compliance test to see if querying / rule evaluation starts to fail after a while, but everything ran fine. Still, I'm wondering if there is something present in your setup that might be triggering the issue.

fpetkovski:
What I am noticing is that rulers start to emit `ThanosRulerRuleEvaluationWithWarning` and log `No StoreAPIs matched for this query`. I checked the alerting rules and indeed it seems that there's no store that matches the query, but with the latest main image (main-2022-07-07-5f31fc5a), this is not reported as an error.

fpetkovski:
@matej-g Looking at [the code](https://github.com/thanos-io/thanos/blob/main/pkg/store/proxy.go#L335), I think the warning here is expected. So maybe this PR actually fixes a problem that existed before where the warning was not reported.

PR #5470, opened 2022-07-05
receiver: Expose stats for all tenants

fpetkovski:
Thanos Receiver supports the Prometheus tsdb status API and can expose
TSDB stats for a single tenant.

This commit extends that functionality and allows users to request
TSDB stats for all tenants using the all_tenants=true query parameter.

@yeya24 this is a follow up on what we discussed here: https://github.com/thanos-io/thanos/pull/5402#discussion_r887562883

## Changes

Allow exposing TSDB stats for all tenants in Receiver.

## Verification

Manual testing on local machine.


pkg/api/status/v1.go
```diff
@@ -84,43 +86,30 @@ func (sapi *StatusAPI) Register(r *route.Router, tracer opentracing.Tracer, logg
...
+	if stats == nil {
 		return nil, nil, &api.ApiError{Typ: api.ErrorBadData, Err: fmt.Errorf("unknown tenant")}
 	}
 
-	metrics, err := sapi.options.Registry.Gather()
-	if err != nil {
-		return nil, []error{err}, nil
-	}
-
-	chunkCount := int64(math.NaN())
```
fpetkovski, line 101: This is an extremely hacky way to get number of head chunks and I suggest we remove the field from the response. Admins can still get this data since it is exposed as a metric with separate series per tenant.

I'm not sure why Prometheus does it this way tbh.

bwplotka: For consistency I guess. So you propose to skip this in our API?

bwplotka: I would still do it IMO

bwplotka (approved):
Nice thanks!

I think it's worth getting that one thing from metric, but I agree it's hacky (kind of anti-pattern).

pkg/api/status/v1.go
fpetkovski: So we will split the values per tenant using the tenant label?

fpetkovski:
@bwplotka mind taking another look?

bwplotka (dismissed):
Amazing, looks good! 💪🏽

pkg/api/status/v1.go
```diff
@@ -18,15 +18,16 @@ package status
 
 import (
 	"fmt"
-	"math"
 	"net/http"
 
+	"github.com/prometheus/client_golang/prometheus"
+
+	"github.com/prometheus/prometheus/tsdb"
```
yeya24, line 25: Can we merge the imports?

pkg/receive/handler.go
```diff
@@ -17,6 +17,8 @@ import (
 	"sync"
 	"time"
 
+	"github.com/thanos-io/thanos/pkg/api"
+
 	statusapi "github.com/thanos-io/thanos/pkg/api/status"
```
yeya24, line 22: Let's merge this.

pkg/receive/multitsdb.go
```diff
@@ -12,6 +12,8 @@ import (
 	"sync"
 	"time"
 
+	"github.com/thanos-io/thanos/pkg/api/status"
+
```
yeya24, line 16: Let's merge import

pkg/receive/multitsdb.go
```diff
@@ -393,7 +398,39 @@ func (t *MultiTSDB) Stats(tenantID, statsByLabelName string) *tsdb.Stats {
...
+			defer mu.Unlock()
+			result = append(result, status.TenantStats{
+				Tenant: tenantID,
+				Stats:  stats,
+			})
+		}(tenantID, tenantInstance)
+	}
+	wg.Wait()
+
+	return result
```
yeya24, line 433: I prefer to sort the results by tenant name.

pkg/receive/multitsdb.go
```diff
@@ -32,7 +34,10 @@ import (
 	"github.com/thanos-io/thanos/pkg/store/labelpb"
 )
 
-type GetStatsFunc func(tenantID, statsByLabelName string) *tsdb.Stats
+type TSDBStats interface {
+	SingleTenantStats(tenantID, statsByLabelName string) *status.TenantStats
+	AllTenantStats(statsByLabelName string) []status.TenantStats
```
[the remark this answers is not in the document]
fpetkovski: Yes, the idea was to use nil to signify a missing tenant. If we merge the two methods into one, how would we request stats for only one tenant? Would we set tenantID to an empty string?

pkg/api/status/v1.go
fpetkovski: Prometheus imports should be merged now

pkg/receive/multitsdb.go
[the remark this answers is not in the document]
yeya24: If the reason is only to check whether the tenant exists or not we can definitely have another return value.
If later we want to implement `TSDBStats` for other purposes then we will find ourselves implementing two methods, but for the same purpose.

```go
type TSDBStats interface {
  TenantStats(statsByLabelName string, tenantIDs ...string) ([]status.TenantStats, error)
}
```

Since we might need to get only some tenant's stats, not all tenants. Using an array is better for me.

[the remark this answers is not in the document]
fpetkovski: Got it, I think it's a very good point. I unified both methods under one.

CHANGELOG.md
```diff
@@ -18,6 +18,7 @@ We use *breaking :warning:* to mark changes that are not backward compatible (re
 - [#5440](https://github.com/thanos-io/thanos/pull/5440) HTTP metrics: export number of in-flight HTTP requests.
 - [#5424](https://github.com/thanos-io/thanos/pull/5424) Receive: Export metrics regarding size of remote write requests.
 - [#5420](https://github.com/thanos-io/thanos/pull/5420) Receive: Automatically remove stale tenants.
+- [#5402](https://github.com/thanos-io/thanos/pull/5402) Receive: Implement exposing TSDB stats for all tenants
```
yeya24, line 21: This needs a period. CI failure seems due to this.

pkg/receive/handler.go
```diff
@@ -271,17 +274,27 @@ func (h *Handler) testReady(f http.HandlerFunc) http.HandlerFunc {
...
+	if getAllTenantStats {
+		return h.options.TSDBStats.TenantStats(statsByLabelName), nil
+	}
+
 	if tenantID == "" {
 		tenantID = h.options.DefaultTenantID
 	}
 
-	return h.options.GetTSDBStats(tenantID, statsByLabelName), nil
+	return h.options.TSDBStats.TenantStats(statsByLabelName, tenantID), nil
```
yeya24, line 297: There is still no way to query multiple tenants (not all) from the API. Do we want to provide this capability to users?

CHANGELOG.md
fpetkovski: Thanks, good catch.

pkg/receive/handler.go
fpetkovski: My thought was that getting stats for all tenants should be good enough for most users that want to identify where high cardinality series come from. But we can add it if you think the use case is going to be common.

yeya24: It would be great if we can make it more extensible to support those use cases at this stage. If you think the current implementation is good to be extended for those cases in the future then I am good.

yeya24 (dismissed):
LGTM

pkg/receive/handler.go
fpetkovski: Yes, I think we should be good to extend functionality in the future. We can pass multiple tenant IDs [here](https://github.com/thanos-io/thanos/pull/5470/files#diff-61eda56d07cd51eb9fde89c6bd9f1fe7e260edb05eb8fb4c1c81aa258f034331R393
). So the main question will be how the users will provide the tenant IDs, whether as comma-separated values in the `THANOS-TENANT` header or something else.

fpetkovski:
Added an update for the docs in the TSDB stats receiver section.

CHANGELOG.md
```diff
@@ -20,6 +20,7 @@ We use *breaking :warning:* to mark changes that are not backward compatible (re
 - [#5420](https://github.com/thanos-io/thanos/pull/5420) Receive: Automatically remove stale tenants.
 - [#5472](https://github.com/thanos-io/thanos/pull/5472) Receive: add new tenant metrics to example dashboard.
 - [#5475](https://github.com/thanos-io/thanos/pull/5475) Compact/Store: Added `--block-files-concurrency` allowing to configure number of go routines for download/upload block files during compaction.
+- [#5402](https://github.com/thanos-io/thanos/pull/5402) Receive: Implement exposing TSDB stats for all tenants.
```
yeya24, line 23: One last thing... Let's fix the PR number to #5470 then we can merge.

yeya24 (approved):
LGTM

Issue #5130, opened 2022-02-05
Move back to Minio upstream


wiardvanrij:
Since https://github.com/thanos-io/thanos/issues/3967 we are using a fork. Which was absolutely necessary, however it will now sort of shoot us in the foot for future features. 

For example: https://github.com/thanos-io/thanos/issues/5101

but also I'm running on a M1 laptop and local dev is not working due to:

```
# github.com/shirou/gopsutil/cpu
../../../go/pkg/mod/github.com/shirou/gopsutil@v2.20.3-0.20200314133625-53cec6b37e6a+incompatible/cpu/cpu_darwin_cgo.go:13:5: warning: 'TARGET_OS_MAC' is not defined, evaluates to 0 [-Wundef-prefix=TARGET_OS_]
# github.com/shirou/gopsutil/disk
disk_darwin.c:28:2: warning: 'IOMasterPort' is deprecated: first deprecated in macOS 12.0 [-Wdeprecated-declarations]
/Library/Developer/CommandLineTools/SDKs/MacOSX.sdk/System/Library/Frameworks/IOKit.framework/Headers/IOKitLib.h:132:1: note: 'IOMasterPort' has been explicitly marked deprecated here
# github.com/shirou/gopsutil/process
../../../go/pkg/mod/github.com/shirou/gopsutil@v2.20.3-0.20200314133625-53cec6b37e6a+incompatible/process/process_darwin.go:576:34: undefined: KinfoProc
../../../go/pkg/mod/github.com/shirou/gopsutil@v2.20.3-0.20200314133625-53cec6b37e6a+incompatible/process/process_darwin.go:577:8: undefined: KinfoProc
../../../go/pkg/mod/github.com/shirou/gopsutil@v2.20.3-0.20200314133625-53cec6b37e6a+incompatible/process/process_darwin.go:590:32: undefined: KinfoProc
../../../go/pkg/mod/github.com/shirou/gopsutil@v2.20.3-0.20200314133625-53cec6b37e6a+incompatible/process/process_darwin.go:594:62: undefined: KinfoProc
```

Main problem that needs to be resolved is the race that caused the memory increase upstream. 

Also, two things here that we need to split. One is that we use minio in the quickstart and e2e, but also as client. I _think_ we could update the one via bingo for e2e and solve local dev issues on M1 machines?

wiardvanrij:
Still relevant

B0go:
[...]

Issue #5101, opened 2022-01-25
IMDSv2 support requires minio-go/v7 v7.0.13


chrisHMRC:
### Is your proposal related to a problem?

Current Thanos does not support AWS IMDSv2 

### Describe the solution you'd like

Upgrade the minio-go library to IMDSv2 supported version i.e v7.0.11+

### Describe alternatives you've considered

Only way.

### Additional context

Would be fantastic if it can be added asap.

chrisHMRC:
Believe this is more complicated that just incrementing the minio version; since currently thanos is running a fork; from another issue.

krishicks:
👋 I'd love to see IMDSv2 support added to Thanos. I cloned the upstream and found it merged cleanly into the [fork](https://github.com/bwplotka/minio-go/tree/optimize).

Is it possible to bump the fork as a quick way to get IMDSv2 support?

vxe:
Hi do we have an ETA for this?

bwplotka:
Sounds like more people are impacted with this. Help wanted! (:

wiardvanrij:
We all want it fast, but we will hit new issues if we don't move back to upstream and provide upstream fixes when needed. 

The fork was basically a hotfix for a serious issue. This is, with all due respect, a feature request. 

The clean path is to move back to upstream if possible. We can't start to maintain a fork of minio (: As we are also an OSS project, this is maybe a nice incentive to get help on this issue.

P.s. I'm not strictly saying that this is the path we must walk, but from a projects point of view the most healthy path IMO.

bwplotka:
Blocked on: https://github.com/thanos-io/thanos/issues/5393

bwplotka:
I updated fork, so we can start using IMDSv2 unless there is anything to do on our side to enable it.

See: https://github.com/thanos-io/thanos/pull/5474

markmsmith:
@bwplotka Did you figure out if this just works now, or if there's remaining work to be done in Thanos to enable it?

oded-dd:
Hey. What is the current status?

matej-g:
Hey @oded-dd, we're now on a [newer `minio-go`](https://github.com/thanos-io/objstore/blob/main/go.mod) version, so unless there are more extra steps needs to enable IMDSv2, this should be considered resolved.

oded-dd:
@matej-g Thank you, we tested it and it seems to work for version 0.28

matej-g:
[...]

PR #5474, opened 2022-07-05
Updated minio-go fork to latest.

bwplotka:
NOTE: Optimization is propopsed to upstream to avoid fork in future.

Relates to https://github.com/thanos-io/thanos/issues/5101 and https://github.com/thanos-io/thanos/issues/5130

Signed-off-by: bwplotka <bwplotka@gmail.com>


yeya24:
As the upstream pr has been accepted, let's remove the minio fork?

bwplotka:
Yes, will double check and update this, thanks!

pkg/objstore/s3/s3.go
```diff
@@ -485,6 +485,7 @@ func (b *Bucket) Upload(ctx context.Context, name string, r io.Reader) error {
 			PartSize:             partSize,
 			ServerSideEncryption: sse,
 			UserMetadata:         b.putUserMetadata,
+			NumThreads:           4,
```
matej-g, line 488: Any reason to set this to 4? Is there a default if this is not set? Should we add a comment with explanation?

matej-g (commented):
Feels great to remove things :broom:, nice job also on the upstream work!

pkg/objstore/s3/s3.go
bwplotka: Happy to add ocmment. This is to make sure we are not prone to internal updates to that number. we can play with adjusting this as well in furure.

GiedriusS (approved):
:muscle:

Issue #5414, opened 2022-06-09
Add configurable parallelism to objstore.DownloadDir and objstore.UploadDir


gburek-fastly:
Our environment has large blocks that are comprised of many chunks: 
```
512 MiB      gs://-/01G53KK11WRQTSM7M2W1217H5H/chunks/002371
512 MiB      gs://-/01G53KK11WRQTSM7M2W1217H5H/chunks/002372
512 MiB      gs://-/01G53KK11WRQTSM7M2W1217H5H/chunks/002373
187.95 MiB   gs://-/01G53KK11WRQTSM7M2W1217H5H/chunks/002374
1.16 TiB     gs://-/01G53KK11WRQTSM7M2W1217H5H/chunks/
1.18 TiB     gs://-/01G53KK11WRQTSM7M2W1217H5H/
```

They are slow to upload and download:
```
"duration":"5h40m44.011902037s","msg":"downloaded and verified blocks; compacting blocks"
...
"duration":"8h9m20.598265105s","msg":"uploaded block"
```

It appears that thanos [iterates through a bucket](https://github.com/thanos-io/thanos/blob/main/pkg/objstore/objstore.go#L263) and downloads these files sequentially and it does [the same for uploading them](https://github.com/thanos-io/thanos/blob/main/pkg/objstore/objstore.go#L183). 

It would be a huge performance boost for compact nodes to be able to concurrently upload and download the individual objects that a block is comprised of.

kakkoyun:
This is fixed by https://github.com/thanos-io/thanos/pull/5475

PR #5475, opened 2022-07-05
objstore: Download and Upload block files in parallel

alanprot:
Hi...

We were doing some tests on the thanos compactor on cortex and we noticed that for very big tenants the compactor was mostly waiting the blocks to be downloaded but at the same time it was not using all the network or disk available capacity.

Looking at the thanos code we noticed that the blocks and chunks are being downloaded one by one so we decided to test 2 changes (`Download chunks in parallel` and `Download blocks in parallel`) that seems to have improved a lot the compaction time.

On or test case we were compacting 6 blocks ~ 200GB each as as we can see below we were taking ~45 minutes to download each block at ~80 MiB/s (almost 7 hours to download all the blocks)

Screen Shot 2022-07-05 at 3 11 31 PM
 
* Download chunks in parallel

This change already improved the download time quite a bit from 45min per block to ~10 min per block at ~ 580 MiB/s.

Screen Shot 2022-07-05 at 3 13 55 PM

We can still see the 10 min gaps between blocks where thanos is reading the index (GatherIndexHealthStats).

* `Download blocks in parallel`

This change tries to take advantage of the time thanos is reading the index to continue to download blocks.

Screen Shot 2022-07-05 at 3 20 53 PM

With both change the time required to download and read the index of all blocks decreased from almost 7 hours to ~1 hour.

PS: We may wanna limit the download concurrency.

## Changes
* Download chunks in parallel
* * `Download blocks in parallel`

## Verification

draft

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -19,6 +19,7 @@ We use *breaking :warning:* to mark changes that are not backward compatible (re
 - [#5424](https://github.com/thanos-io/thanos/pull/5424) Receive: Export metrics regarding size of remote write requests.
 - [#5420](https://github.com/thanos-io/thanos/pull/5420) Receive: Automatically remove stale tenants.
 - [#5472](https://github.com/thanos-io/thanos/pull/5472) Receive: add new tenant metrics to example dashboard.
+- [#5475](https://github.com/thanos-io/thanos/pull/5475) Compact/Store: Added `--block-fetch-concurrency` allowing to configure number of go routines for download block files during compaction.
 
 ### Changed
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -20,6 +20,7 @@ We use *breaking :warning:* to mark changes that are not backward compatible (re
 - [#5420](https://github.com/thanos-io/thanos/pull/5420) Receive: Automatically remove stale tenants.
 - [#5472](https://github.com/thanos-io/thanos/pull/5472) Receive: add new tenant metrics to example dashboard.
 - [#5475](https://github.com/thanos-io/thanos/pull/5475) Compact/Store: Added `--block-fetch-concurrency` allowing to configure number of go routines for download block files during compaction.
+- [#5475](https://github.com/thanos-io/thanos/pull/5475) Compact/Store: Added `--block-files-concurrency` allowing to configure number of go routines for download block files during compaction.
 
 ### Changed
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -21,6 +21,7 @@ We use *breaking :warning:* to mark changes that are not backward compatible (re
 - [#5472](https://github.com/thanos-io/thanos/pull/5472) Receive: add new tenant metrics to example dashboard.
 - [#5475](https://github.com/thanos-io/thanos/pull/5475) Compact/Store: Added `--block-fetch-concurrency` allowing to configure number of go routines for download block files during compaction.
 - [#5475](https://github.com/thanos-io/thanos/pull/5475) Compact/Store: Added `--block-files-concurrency` allowing to configure number of go routines for download block files during compaction.
+- [#5475](https://github.com/thanos-io/thanos/pull/5475) Compact/Store: Added `--block-files-concurrency` allowing to configure number of go routines for download/upload block files during compaction.
 
 ### Changed
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -18,7 +18,8 @@ We use *breaking :warning:* to mark changes that are not backward compatible (re
 - [#5440](https://github.com/thanos-io/thanos/pull/5440) HTTP metrics: export number of in-flight HTTP requests.
 - [#5424](https://github.com/thanos-io/thanos/pull/5424) Receive: Export metrics regarding size of remote write requests.
 - [#5420](https://github.com/thanos-io/thanos/pull/5420) Receive: Automatically remove stale tenants.
-- [#5475](https://github.com/thanos-io/thanos/pull/5475) Compact/Store: Added `--block-files-concurrency` allowing to configure number of go routines for upload/download block files during compaction.
+- [#5472](https://github.com/thanos-io/thanos/pull/5472) Receive: add new tenant metrics to example dashboard.
+- [#5475](https://github.com/thanos-io/thanos/pull/5475) Compact/Store: Added `--block-files-concurrency` allowing to configure number of go routines for download/upload block files during compaction.
 
 ### Changed
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -18,10 +18,7 @@ We use *breaking :warning:* to mark changes that are not backward compatible (re
 - [#5440](https://github.com/thanos-io/thanos/pull/5440) HTTP metrics: export number of in-flight HTTP requests.
 - [#5424](https://github.com/thanos-io/thanos/pull/5424) Receive: Export metrics regarding size of remote write requests.
 - [#5420](https://github.com/thanos-io/thanos/pull/5420) Receive: Automatically remove stale tenants.
-- [#5472](https://github.com/thanos-io/thanos/pull/5472) Receive: add new tenant metrics to example dashboard.
-- [#5475](https://github.com/thanos-io/thanos/pull/5475) Compact/Store: Added `--block-fetch-concurrency` allowing to configure number of go routines for download block files during compaction.
-- [#5475](https://github.com/thanos-io/thanos/pull/5475) Compact/Store: Added `--block-files-concurrency` allowing to configure number of go routines for download block files during compaction.
-- [#5475](https://github.com/thanos-io/thanos/pull/5475) Compact/Store: Added `--block-files-concurrency` allowing to configure number of go routines for download/upload block files during compaction.
+- [#5475](https://github.com/thanos-io/thanos/pull/5475) Compact/Store: Added `--block-files-concurrency` allowing to configure number of go routines for upload/download block files during compaction.
 
 ### Changed
 
--- a/pkg/compact/compact.go
+++ b/pkg/compact/compact.go
@@ -913,7 +913,7 @@ func (cg *Group) areBlocksOverlapping(include *metadata.Meta, exclude ...*metada
 }
 
 // RepairIssue347 repairs the https://github.com/prometheus/tsdb/issues/347 issue when having issue347Error.
-func RepairIssue347(ctx context.Context, logger log.Logger, bkt objstore.Bucket, blocksMarkedForDeletion prometheus.Counter, g *Group, issue347Err error) error {
+func RepairIssue347(ctx context.Context, logger log.Logger, bkt objstore.Bucket, blocksMarkedForDeletion prometheus.Counter, blockFilesConcurrency int, issue347Err error) error {
 	ie, ok := errors.Cause(issue347Err).(Issue347Error)
 	if !ok {
 		return errors.Errorf("Given error is not an issue347 error: %v", issue347Err)
@@ -953,7 +953,7 @@ func RepairIssue347(ctx context.Context, logger log.Logger, bkt objstore.Bucket,
 	}
 
 	level.Info(logger).Log("msg", "uploading repaired block", "newID", resid)
-	if err = block.Upload(ctx, logger, bkt, filepath.Join(tmpdir, resid.String()), metadata.NoneFunc, objstore.WithUploadConcurrency(g.blockFilesConcurrency)); err != nil {
+	if err = block.Upload(ctx, logger, bkt, filepath.Join(tmpdir, resid.String()), metadata.NoneFunc, objstore.WithUploadConcurrency(blockFilesConcurrency)); err != nil {
 		return retry(errors.Wrapf(err, "upload of %s failed", resid))
 	}
 
@@ -1245,7 +1245,7 @@ func (c *BucketCompactor) Compact(ctx context.Context) (rerr error) {
 					}
 
 					if IsIssue347Error(err) {
-						if err := RepairIssue347(workCtx, c.logger, c.bkt, c.sy.metrics.blocksMarkedForDeletion, g, err); err == nil {
+						if err := RepairIssue347(workCtx, c.logger, c.bkt, c.sy.metrics.blocksMarkedForDeletion, g.blockFilesConcurrency, err); err == nil {
 							mtx.Lock()
 							finishedAllGroups = false
 							mtx.Unlock()
--- a/pkg/compact/compact.go
+++ b/pkg/compact/compact.go
@@ -232,6 +232,7 @@ type DefaultGrouper struct {
 	blocksMarkedForDeletion  prometheus.Counter
 	blocksMarkedForNoCompact prometheus.Counter
 	hashFunc                 metadata.HashFunc
+	blockFetchConcurrency    int
 }
 
 // NewDefaultGrouper makes a new DefaultGrouper.
@@ -245,6 +246,7 @@ func NewDefaultGrouper(
 	garbageCollectedBlocks prometheus.Counter,
 	blocksMarkedForNoCompact prometheus.Counter,
 	hashFunc metadata.HashFunc,
+	blockFetchConcurrency int,
 ) *DefaultGrouper {
 	return &DefaultGrouper{
 		bkt:                      bkt,
@@ -275,6 +277,7 @@ func NewDefaultGrouper(
 		garbageCollectedBlocks:   garbageCollectedBlocks,
 		blocksMarkedForDeletion:  blocksMarkedForDeletion,
 		hashFunc:                 hashFunc,
+		blockFetchConcurrency:    blockFetchConcurrency,
 	}
 }
 
@@ -304,6 +307,7 @@ func (g *DefaultGrouper) Groups(blocks map[ulid.ULID]*metadata.Meta) (res []*Gro
 				g.blocksMarkedForDeletion,
 				g.blocksMarkedForNoCompact,
 				g.hashFunc,
+				g.blockFetchConcurrency,
 			)
 			if err != nil {
 				return nil, errors.Wrap(err, "create compaction group")
@@ -342,6 +346,7 @@ type Group struct {
 	blocksMarkedForDeletion     prometheus.Counter
 	blocksMarkedForNoCompact    prometheus.Counter
 	hashFunc                    metadata.HashFunc
+	blockFetchConcurrency       int
 }
 
 // NewGroup returns a new compaction group.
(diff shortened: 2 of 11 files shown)


yeya24:
I have a previous PR for this https://github.com/thanos-io/thanos/pull/4435/files and I did some tests on this. If there are multiple blocks to compact it is easy to occupy all the disk spaces and bandwidth.
What about supporting a number of concurrency workers to download/verify the block?

alanprot:
Yeah... I think we can limit the concurrency.

Also doing just the first change (download chunks in parallel) already improves a lot this case and it may be safer (as at this stage we are just piping the bytes from the remote storage to the disk)

alanprot:
@bwplotka any thoughts about this?

yeya24:
@alanprot Docs failling is not related to this pr. But it would be good if you can try to rebase on main and try again.

Btw please sign off your commits.

yeya24:
> Yeah... I think we can limit the concurrency.
> 
> Also doing just the first change (download chunks in parallel) already improves a lot this case and it may be safer (as at this stage we are just piping the bytes from the remote storage to the disk)

Let's do this in this pr. For most users, no concurrency (worker = 1) is good enough. Making this configurable is better for them to choose.

alanprot:
Thanks for the feedback @yeya24 I will try to do this today

alanprot:
@yeya24 You think i should create an "DownloadOptions" like that [IterOption](https://github.com/thanos-io/thanos/blob/18b70896553f350d904d993b6fb9ac640249890b/pkg/objstore/objstore.go#L101) that already exists?

alanprot:
HI @yeya24 

I changed the PR to just download chunks in parallel and also have a configuration (default 1) to control the concurrency.

Thanks

yeya24 (commented):
If it is just downloading chunks, I feel we don't need the concurrency limit. The worker limit thing I meant is mainly on another change you did. 
For this change, overall I think is good. Any thoughts from other maintainers?

pkg/objstore/objstore.go
```diff
@@ -119,6 +121,39 @@ func ApplyIterOptions(options ...IterOption) IterParams {
...
+}
+
+// WithFetchConcurrency is an option to set the concurrency of the download operation.
+func WithFetchConcurrency(concurrency int) DownloadDirOption {
+	return func(params *DownloadDirParams) {
+		params.concurrency = concurrency
+	}
+}
+
+func ApplyDownloadOptions(options ...DownloadDirOption) DownloadDirParams {
```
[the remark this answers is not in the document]
alanprot: > If it is just downloading chunks, I feel we don't need the concurrency limit. The worker limit thing I meant is mainly on another change you did. For this change, overall I think is good. Any thoughts from other maintainers?

Oh.. but downloading chunks in parallel can have the same effect of downloading blocks  in parallel  with regards to use  lots of bandwidth - that 's why added the concurrency limit.


> Let's make it private?

Yeah.. lemme change that.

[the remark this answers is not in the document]
yeya24: Yes, I agree. My initial thought was to limit on that code only so that the function signature remains unchanged. This one is also good and even more granular, only signature changed a little bit. Let's see what other maintainers think.

yeya24:
https://github.com/thanos-io/thanos/issues/5414 is also related so we might need to cover upload part, too (maybe someone else can do it in another pr)

alanprot:
> #5414 is also related so we might need to cover upload part, too (maybe someone else can do it in another pr)

Oh.. I can definitely do the upload part as well.. if we think make sense.

yeya24:
Yes, that would be great. Maybe we can use the same flag for both download and upload concurrency then we might need to rename it somehow as it is called 'fetch-concurrency' only.

WDYT?

alanprot:
The compact configuration is already [block-fetch-concurrency](block-fetch-concurrency) but i think in this case we may wanna change to something like `block-objstore-concurrency`?

The would keep the [DownloadDirOption](https://github.com/thanos-io/thanos/pull/5475/files#diff-41e5eeb71bb7097b6a59cdc9eb0002efd60df6b42394773d6d444eb8fec80676R125) as it and create a new one for upload (UploadDirOption) as those calls can have different options in the future (even now it already have as UploadDir does not have `ignoredPaths`), make sense?

yeya24:
> The compact configuration is already [block-fetch-concurrency](https://github.com/thanos-io/thanos/pull/5475/files#diff-6c2257b871fd1196514f664bc7e44cb21681215e0929710d0ad5ceea90b8e122R693) but i think in this case we may wanna change to something like block-objstore-concurrency?

Yeah, we need to change the flag name to cover upload, too. I think it is better to make it more specific. `block-xxx-concurrency` is a little bit ambiguous, what about `block-files-objstore-concurrency` or simply `objstore-files-concurrency`? 

Naming is so hard 😄

alanprot:
Ok @yeya24  :D 

I added the Upload Concurrency  as well.. 

Feel free to ask to change the parameter name! hahha =p

pkg/compact/compact.go
```diff
@@ -901,7 +913,7 @@ func (cg *Group) areBlocksOverlapping(include *metadata.Meta, exclude ...*metada
 }
 
 // RepairIssue347 repairs the https://github.com/prometheus/tsdb/issues/347 issue when having issue347Error.
-func RepairIssue347(ctx context.Context, logger log.Logger, bkt objstore.Bucket, blocksMarkedForDeletion prometheus.Counter, issue347Err error) error {
+func RepairIssue347(ctx context.Context, logger log.Logger, bkt objstore.Bucket, blocksMarkedForDeletion prometheus.Counter, g *Group, issue347Err error) error {
```
yeya24, line 916: I prefer to pass `g.blockFilesConcurrency` or `UploadDirOption` in the function signature rather than passing the whole `Group` struct, WDYT?

yeya24 (approved):
Amazing work. Only one nit left and rebase for changelog conflict.

LGTM.

pkg/compact/compact.go
alanprot: I think make sense...

I was just thinking "we may need more thing from the group in the future and if so, we will not need to change the signature".. but passing the `blockFilesConcurrency ` is fine as well.

alanprot:
Rebased! :D

CHANGELOG.md
```diff
@@ -18,6 +18,9 @@ We use *breaking :warning:* to mark changes that are not backward compatible (re
 - [#5440](https://github.com/thanos-io/thanos/pull/5440) HTTP metrics: export number of in-flight HTTP requests.
 - [#5424](https://github.com/thanos-io/thanos/pull/5424) Receive: Export metrics regarding size of remote write requests.
 - [#5420](https://github.com/thanos-io/thanos/pull/5420) Receive: Automatically remove stale tenants.
+- [#5475](https://github.com/thanos-io/thanos/pull/5475) Compact/Store: Added `--block-fetch-concurrency` allowing to configure number of go routines for download block files during compaction.
+- [#5475](https://github.com/thanos-io/thanos/pull/5475) Compact/Store: Added `--block-files-concurrency` allowing to configure number of go routines for download block files during compaction.
+- [#5475](https://github.com/thanos-io/thanos/pull/5475) Compact/Store: Added `--block-files-concurrency` allowing to configure number of go routines for download/upload block files during compaction.
```
yeya24, line 23: Can you please remove the duplicates here?

yeya24 (commented):
It would be good to have other people's feedback for this pr. @kakkoyun @bwplotka @GiedriusS, also @pracucci maybe?

CHANGELOG.md
alanprot: Gosh.. That was the rebase.. removing!!

pkg/objstore/objstore.go
```diff
@@ -172,29 +232,45 @@ func NopCloserWithSize(r io.Reader) io.ReadCloser {
 
 // UploadDir uploads all files in srcdir to the bucket with into a top-level directory
 // named dstdir. It is a caller responsibility to clean partial upload in case of failure.
-func UploadDir(ctx context.Context, logger log.Logger, bkt Bucket, srcdir, dstdir string) error {
+func UploadDir(ctx context.Context, logger log.Logger, bkt Bucket, srcdir, dstdir string, options ...UploadDirOption) error {
 	df, err := os.Stat(srcdir)
+	opts := applyUploadDirOptions(options...)
+	g, ctx := errgroup.WithContext(ctx)
+	guard := make(chan struct{}, opts.concurrency)
```
kakkoyun, line 239: `errgroup.SetLimit` exactly for limiting the number of active goroutines. I think we should use it rather than implement our own mechanism.
Were there any reasons why you preferred it this way?

pkg/objstore/objstore.go
```diff
@@ -172,29 +232,45 @@ func NopCloserWithSize(r io.Reader) io.ReadCloser {
...
+			return UploadFile(ctx, logger, bkt, src, dst)
+		})
 
-		dst := path.Join(dstdir, filepath.ToSlash(srcRel))
-		return UploadFile(ctx, logger, bkt, src, dst)
+		return nil
 	})
+
+	if err == nil {
+		err = g.Wait()
```
kakkoyun, line 270: ~~In any case, we should wait for all the other running goroutine or call them. We might leak goroutines here, no?~~

kakkoyun: NVM. `The first call to return a non-nil error cancels the group; its error will be returned by Wait.`

kakkoyun: Better to add a comment that error handling and context cancelling is handles by the `errgroup`

pkg/objstore/objstore.go
```diff
@@ -172,29 +232,45 @@ func NopCloserWithSize(r io.Reader) io.ReadCloser {
 
 // UploadDir uploads all files in srcdir to the bucket with into a top-level directory
 // named dstdir. It is a caller responsibility to clean partial upload in case of failure.
-func UploadDir(ctx context.Context, logger log.Logger, bkt Bucket, srcdir, dstdir string) error {
+func UploadDir(ctx context.Context, logger log.Logger, bkt Bucket, srcdir, dstdir string, options ...UploadDirOption) error {
 	df, err := os.Stat(srcdir)
+	opts := applyUploadDirOptions(options...)
+	g, ctx := errgroup.WithContext(ctx)
```
kakkoyun, line 238: `The derived Context is canceled the first time a function passed to Go returns a non-nil error or the first time Wait returns, whichever occurs first.`

pkg/objstore/objstore.go
```diff
@@ -254,34 +330,55 @@ func DownloadFile(ctx context.Context, logger log.Logger, bkt BucketReader, src,
...
 // DownloadDir downloads all object found in the directory into the local directory.
-func DownloadDir(ctx context.Context, logger log.Logger, bkt BucketReader, originalSrc, src, dst string, ignoredPaths ...string) error {
+func DownloadDir(ctx context.Context, logger log.Logger, bkt BucketReader, originalSrc, src, dst string, options ...DownloadDirOption) error {
 	if err := os.MkdirAll(dst, 0750); err != nil {
 		return errors.Wrap(err, "create dir")
 	}
+	opts := applyDownloadDirOptions(options...)
+
+	g, ctx := errgroup.WithContext(ctx)
+	guard := make(chan struct{}, opts.concurrency)
```
kakkoyun, line 340: ditto.

pkg/objstore/objstore.go
```diff
@@ -254,34 +330,55 @@ func DownloadFile(ctx context.Context, logger log.Logger, bkt BucketReader, src,
...
+					level.Debug(logger).Log("msg", "not downloading again because a provided path matches this one", "file", name)
+					return nil
+				}
+			}
+			if err := DownloadFile(ctx, logger, bkt, name, dst); err != nil {
 				return err
 			}
+
+			m.Lock()
+			defer m.Unlock()
```
kakkoyun, line 370: ditto.

pkg/objstore/objstore.go
```diff
@@ -254,34 +330,55 @@ func DownloadFile(ctx context.Context, logger log.Logger, bkt BucketReader, src,
...
+		guard <- struct{}{}
+		g.Go(func() error {
+			defer func() { <-guard }()
+			dst := filepath.Join(dst, filepath.Base(name))
+			if strings.HasSuffix(name, DirDelim) {
+				if err := DownloadDir(ctx, logger, bkt, originalSrc, name, dst, options...); err != nil {
+					return err
+				}
+				m.Lock()
+				defer m.Unlock()
```
kakkoyun, line 355: This lock is just to guard `downloadedFiles`, no? I would rather we minimize the critical section rather than use `defer`.

kakkoyun (changes requested):
Good job! Thanks for doing this.

I have requested a couple of little changes.

pkg/objstore/objstore.go
yeya24: We will return at line 357 and the next defer will return at line 372. I think this should be fine, but also okay to change to make it clearer.

pkg/objstore/objstore.go
```diff
@@ -119,6 +121,64 @@ func ApplyIterOptions(options ...IterOption) IterParams {
 	return out
 }
 
+// DownloadDirOption configures the provided params.
+type DownloadDirOption func(params *DownloadDirParams)
+
+// DownloadDirParams holds the DownloadDir() parameters and is used by objstore clients implementations.
+type DownloadDirParams struct {
```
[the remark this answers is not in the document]
yeya24: This is a good point. I think it is better to make it cleaner if possible. 
To remove those we might need to use global variables to store the concurrency value passed from flags. As I think it is weird to have concurrency set on the grouper.

pkg/compact/compact.go
```diff
@@ -901,7 +913,7 @@ func (cg *Group) areBlocksOverlapping(include *metadata.Meta, exclude ...*metada
 }
 
 // RepairIssue347 repairs the https://github.com/prometheus/tsdb/issues/347 issue when having issue347Error.
-func RepairIssue347(ctx context.Context, logger log.Logger, bkt objstore.Bucket, blocksMarkedForDeletion prometheus.Counter, issue347Err error) error {
+func RepairIssue347(ctx context.Context, logger log.Logger, bkt objstore.Bucket, blocksMarkedForDeletion prometheus.Counter, blockFilesConcurrency int, issue347Err error) error {
```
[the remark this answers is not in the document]
alanprot: Make sense. I will change!

pkg/objstore/objstore.go
```diff
@@ -119,6 +121,64 @@ func ApplyIterOptions(options ...IterOption) IterParams {
 	return out
 }
 
+// DownloadDirOption configures the provided params.
+type DownloadDirOption func(params *DownloadDirParams)
```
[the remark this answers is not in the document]
alanprot: 👍

pkg/objstore/objstore.go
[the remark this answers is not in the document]
alanprot: > nit: Could we get rid of the additional layer of abstractions around params? Do we really need download/upload params structs?

I think i did this as i am just following the same pattern we have on the [IterOption](https://github.com/thanos-io/thanos/blob/b5d0a6167d97c0e179dc9d637fe708091da28f3c/pkg/objstore/objstore.go#L73). I particularly think this is more extensible but im ok to change as well.

> Now it is weird to have concurrency set on the grouper.

We already have other config set in the grouper like the [enableVerticalCompaction](https://github.com/thanos-io/thanos/blob/f60b65988c062eb4c3f68ebb4b1b9ea33b11ca19/cmd/thanos/compact.go#L346).. so i think should be fine, no?

pkg/objstore/objstore.go
alanprot: Yeah.. i can make more clear.

pkg/objstore/objstore.go
alanprot: 👍

pkg/objstore/objstore.go
```diff
@@ -119,6 +121,64 @@ func ApplyIterOptions(options ...IterOption) IterParams {
...
+		opt(&out)
+	}
+	return out
+}
+
+// UploadDirOption configures the provided params.
+type UploadDirOption func(params *UploadDirParams)
+
+// UploadDirParams holds the UploadDir() parameters and is used by objstore clients implementations.
+type UploadDirParams struct {
```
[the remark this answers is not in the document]
alanprot: 👍

pkg/objstore/objstore.go
alanprot: 👍

pkg/objstore/objstore.go
alanprot: good catch! i Will change

alanprot: Good catch! i Will change

pkg/objstore/objstore.go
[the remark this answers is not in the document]
alanprot: WDYT?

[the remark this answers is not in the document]
yeya24: > I was just following the same pattern we have on the [IterOption](https://github.com/thanos-io/thanos/blob/b5d0a6167d97c0e179dc9d637fe708091da28f3c/pkg/objstore/objstore.go#L73). I particularly think this is more extensible but i'm ok to change as well.

👍 

> We already have other config set in the grouper like the [enableVerticalCompaction](https://github.com/thanos-io/thanos/blob/f60b65988c062eb4c3f68ebb4b1b9ea33b11ca19/cmd/thanos/compact.go#L346).. so i think should be fine, no?

`DownloadDir` belongs to `objstore` and it is not a method of `Grouper` so it is a little bit weird to have it. This is different from `enableVerticalCompaction` as `compact` is a method of `Grouper` so it is nature to keep it. But I also see that the `Grouper` also has `hashFunc` which is exactly the same case we are discussing about. We should really improve this to have a dedicated struct to keep those settings. But for this pr, let's keep it in the grouper and we can refactor it later.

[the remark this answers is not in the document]
alanprot: Oh iIsee what you mean...

pkg/objstore/objstore.go
```diff
@@ -172,29 +232,43 @@ func NopCloserWithSize(r io.Reader) io.ReadCloser {
 
 // UploadDir uploads all files in srcdir to the bucket with into a top-level directory
 // named dstdir. It is a caller responsibility to clean partial upload in case of failure.
-func UploadDir(ctx context.Context, logger log.Logger, bkt Bucket, srcdir, dstdir string) error {
+func UploadDir(ctx context.Context, logger log.Logger, bkt Bucket, srcdir, dstdir string, options ...UploadOption) error {
 	df, err := os.Stat(srcdir)
+	opts := applyUploadOptions(options...)
+	g, ctx := errgroup.WithContext(ctx)
+	g.SetLimit(opts.concurrency)
```
alanprot, line 239: SetLimit was introduced fairly recently: https://go-review.googlesource.com/c/sync/+/405174

So needed to update the `golang.org/x/sync` package.

alanprot:
Hi @yeya24 @kakkoyun ...

I think i made all the changes requested :D

yeya24 (approved):
LGTM. cc @kakkoyun

kakkoyun (approved):
LGTM! :100: Awesome work!

yeya24:
@alanprot So I guess we will have a follow-up pr to parallelize downloading blocks during compaction? That was the change you made at the original commit

alanprot:
Should we also close https://github.com/thanos-io/thanos/issues/5414?

PR #5477, opened 2022-07-06
promclient: Ignore 405 errors for prometheus buildVersion API requests

olasd:
Older versions of prometheus (such as 2.7 which is shipped by Debian
buster) return a 405 error for non-existent API endpoints instead of the
404 returned by more recent versions.

Signed-off-by: Nicolas Dandrimont <olasd@softwareheritage.org>

## Changes

Added a case for 405 (Method Not Allowed) responses in the promclient build version API requests.

## Verification

This change was tested against the prometheus version shipped in Debian buster
```
||/ Name           Version        Architecture Description
+++-==============-==============-============-==========================================
ii  prometheus     2.7.1+ds-3+b11 amd64        Monitoring system and time series database
```


bwplotka (approved):
Thanks!

Issue #5450, opened 2022-06-27
Thanos Querier/store gateway is failing to return just 2 months of data


anoop2503:
**Thanos, version used**: v0.23.1

**Object Storage Provider**: S3

**What happened**:  
Deployed only thanos-querier and thanos-store-gateway to use an S3 bucket to display historical data. However, it is failing to return just 2 months of data. 

**What you expected to happen**: Thanos should return data without failure

**How to reproduce it (as minimally and precisely as possible)**:

Deploy thanos-store-gateway to point to S3 bucket. Deploy thanos querier to use only thanos-store-gateway. Try to pull historical data for 2 months.

**Full logs to relevant components**:

Getting following error from thanos querier log:
`level=warn ts=2022-06-27T15:25:49.79866573Z caller=endpointset.go:525 component=endpointset msg="update of node failed" err="getting metadata: fallback fetching info from xx.xx.xx.xxx:10901: rpc error: code = DeadlineExceeded desc = context deadline exceeded" address= xx.xx.xx.xxx:10901`

Getting following error from Store-Gateway logs:
`level=debug ts=2022-06-27T15:26:32.614591931Z caller=bucket.go:1087 msg="stats query processed" stats="&{blocksQueried:0 postingsTouched:0 postingsTouchedSizeSum:0 postingsToFetch:0 postingsFetched:0 postingsFetchedSizeSum:0 postingsFetchCount:0 postingsFetchDurationSum:0 cachedPostingsCompressions:0 cachedPostingsCompressionErrors:0 cachedPostingsOriginalSizeSum:0 cachedPostingsCompressedSizeSum:0 cachedPostingsCompressionTimeSum:0 cachedPostingsDecompressions:0 cachedPostingsDecompressionErrors:0 cachedPostingsDecompressionTimeSum:0 seriesTouched:0 seriesTouchedSizeSum:0 seriesFetched:0 seriesFetchedSizeSum:0 seriesFetchCount:0 seriesFetchDurationSum:0 chunksTouched:0 chunksTouchedSizeSum:0 chunksFetched:0 chunksFetchedSizeSum:0 chunksFetchCount:0 chunksFetchDurationSum:0 getAllDuration:0 mergedSeriesCount:0 mergedChunksCount:0 mergeDuration:0}" err="rpc error: code = Aborted desc = fetch series for block 01FQEABSWZBJVT8Z0E0F4Q8KY6: expanded matching posting: get postings: read postings range: read range: context canceled"`

[...]

Issue #5340, opened 2022-05-09
implement Unified Endpoint Discovery proposal


saswatamcode:
This issue currently tracks the work that needs to be done to complete the implementation of the [Unified Endpoint Discovery](https://thanos.io/tip/proposals-accepted/202101-endpoint-discovery.md/), so that the usage of endpoint flag can be recommended by default instead of the current state of flags in Thanos, which can cause [confusion](https://github.com/thanos-io/thanos/issues/5238).

A/C:

- This PR has to be completed: https://github.com/thanos-io/thanos/pull/4785
- This issue needs to be addressed: https://github.com/thanos-io/thanos/issues/4683
- Proposal needs to be implemented: https://thanos.io/tip/proposals-accepted/202101-endpoint-discovery.md/

This is planned to be proposed as a mentorship topic with @bwplotka! 🙂

sandipanpanda:
Hi mentors, I am Sandipan, an undergrad and open source enthusiast. I am interested to take this opportunity of mentorship and to contribute to Thanos by helping in fixing this issue. I am going through the linked issues, pull requests, proposal, and docs.

edith007:
Hi @saswatamcode! 👋🏻 
I am really interested in working on this project. And have even started exploring thanos project and trying to fix bugs. 🙇🏻

daxsis:
Hello @saswatamcode, I applied through LFX and was curious is there anything else except sending resume and cover letter to the application? Maybe some challenge that I missed? Otherwise I'm pretty stocked about the project and looking forward to contributing to these proposal.

Akshit42-hue:
Excited for working on this issue :)

PR #5480, opened 2022-07-08
Expose Endpoint Info timeout parameter

yeya24:

## Changes

Fixes https://github.com/thanos-io/thanos/issues/5450

Expose info timeout param so that users can configure it based on their use case use.

For our use case, we have Thanos Query at cloud control plane and Thanos Sidecars at edge across the world (like Asia, Africa, etc). Due to the physical location distance and network issues, 5 seconds timeout is not sufficient for all cases.


fpetkovski (approved):

matej-g (commented):
It makes sense to introduce this parameter, but I'm a bit wary of introducing new separate flags in face of work being done in https://github.com/thanos-io/thanos/issues/5340. I assume one of the results will be that we will have a comprehensive endpoint configuration, instead of having individual flags. I'm just wondering if this would not mean we introduce a new flag only to deprecate it couple of releases later, when the `endpoint.config` will potentially supersede it. Maybe having this as a hidden flag first would make sense and then move it to the endpoint config. But maybe @bwplotka, @saswatamcode or @SrushtiSapkale can give us more qualified answer!

yeya24:
Good to know we have WIP work on that. If that's the case yeah definitely better than adding this new flag separately.

GiedriusS (commented):
Yeah... maybe let's make it hidden for now? :thinking:

GiedriusS (approved):

matej-g (approved):

matej-g (commented):
Looks like one more conflict, should be good to go afterwards!

GiedriusS (dismissed):

GiedriusS (approved):

PR #5483, opened 2022-07-08
E2E Tests: Use default HTTP config for minio tests

matej-g:

## Changes

Use the same default HTTP config as we use for S3 in tests. Although tests are working, we've been seeing some output indicating `minio` might be struggling (#5474) and I assume this might be due to improper HTTP client config. After the change, I've seen decrease in client-related error messages from `minio` during a test run.

## Verification

Tests ran locally :ok_hand:


kakkoyun (approved):
LGTM :100:

PR #5490, opened 2022-07-12
receive: Tenant whitelist

4xoc:

## Changes

This PR adds a new flag for the receive component that enforces specific tenant values. The flag can be used more than once and if used, any tenant value not matching any white listed entry is disregarded. The validation is done after processing other tenant value sources (like certificate field names or headers). As a result, this provides strong enforcement of tenants within receive.

## Verification

Change runs for some time in combination with certificate field names.


fpetkovski:
Thank you for the contribution. I wonder how this feature is different from using the `tenants` field in the hashrings config to only allow a specific list of tenants.

4xoc:
Right, in that case this feature is indeed already existing. I'm having trouble finding this behavior explained in the docs, would you know where to find it and add it here. That would probably make it easier for someone in the future to find out about this.

fpetkovski:
Yeah I agree it's a bit unclear on how to achieve the behavior. If you're up for it, you could add it here: https://thanos.io/tip/components/receive.md/

PR #5491, opened 2022-07-12
bump codemirror and promql editor to the last version

Nexucis:
This PR is following changes initially made in prometheus : https://github.com/prometheus/prometheus/pull/10841

After this PR, hopefully dependabot should be able to handle the update of the dependencies around `codemirror` and `@prometheus-io/codemirror` :)


yeya24 (approved):
Thanks!

PR #5493, opened 2022-07-13
Parallel download blocks - Follow up of #5475

alanprot:
This is the follow up with the second change described on https://github.com/thanos-io/thanos/pull/5475 (Download Blocks in Parallel).

Currently we download and read the blocks index sequentially (on a single go routine) - the download phase for each block is very IO bound while reading the index is very CPU bound.  This change tries to take advantage of the time thanos is reading the index to continue to download blocks.

The graphs below shows the volume utilization before and after this change:

Before:

Screen Shot 2022-07-05 at 3 13 55 PM

After:

Screen Shot 2022-07-05 at 3 20 53 PM


pkg/compact/compact.go
```diff
@@ -1007,53 +1014,66 @@ func (cg *Group) compact(ctx context.Context, dir string, planner Planner, comp
...
-		tracing.DoInSpanWithErr(ctx, "compaction_block_download", func(ctx context.Context) error {
-			err = block.Download(ctx, cg.logger, cg.bkt, meta.ULID, bdir, objstore.WithFetchConcurrency(cg.blockFilesConcurrency))
-			return err
-		}, opentracing.Tags{"block.id": meta.ULID})
-		if err != nil {
-			return false, ulid.ULID{}, retry(errors.Wrapf(err, "download block %s", meta.ULID))
-		}
+		func(ctx context.Context, meta *metadata.Meta) {
+			g.Go(func() error {
+				tracing.DoInSpanWithErr(ctx, "compaction_block_download", func(ctx context.Context) error {
```
yeya24, line 1032: Hi @alanprot, if I understand correctly, it is still first IO bound -> CPU bound because for each block the process is still sequential. What's the difference of this between your original commit of the previous pr?

alanprot: It is still "Io Bound" -> "CPU Bound".. The advantage here are:
 *  We now can have some blocks downloading while others are reading the index
 * We can use more than 1 CPU to read the indexes now.
 
This change is pretty similar to my original one: https://github.com/thanos-io/thanos/pull/5476/commits/b83620b0308ed37ca96bef11e59775d241155b18

CHANGELOG.md
```diff
@@ -21,6 +21,7 @@ We use *breaking :warning:* to mark changes that are not backward compatible (re
 - [#5472](https://github.com/thanos-io/thanos/pull/5472) Receive: add new tenant metrics to example dashboard.
 - [#5475](https://github.com/thanos-io/thanos/pull/5475) Compact/Store: Added `--block-files-concurrency` allowing to configure number of go routines for download/upload block files during compaction.
 - [#5470](https://github.com/thanos-io/thanos/pull/5470) Receive: Implement exposing TSDB stats for all tenants
+- [#5493](https://github.com/thanos-io/thanos/pull/5493) Compact/Store: Added `--compact.blocks-concurrency` allowing to configure number of go routines for download blocks during compactions.
```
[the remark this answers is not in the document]
alanprot: make sense!

pkg/compact/compact.go
yeya24: The change is almost the same as the original commit but I thought you would change some logic. Maybe I misunderstood it yesterday. Thanks for the clarification! 💯

yeya24: Yes, I agree. Compared to the index verification operation which is CPU bound, the IO part is taking a much longer time to download the block which is a major factor in this case. I think 1 and 2 are pretty similar.

yeya24:
Let's fix the two nits and then we can merge. Thanks @alanprot

alanprot:
> Let's fix the two nits and then we can merge. Thanks @alanprot

done! :D

cmd/thanos/compact.go
```diff
@@ -703,11 +705,13 @@ func (cc *compactConfig) registerFlag(cmd extkingpin.FlagClause) {
...
 	cmd.Flag("compact.concurrency", "Number of goroutines to use when compacting groups.").
 		Default("1").IntVar(&cc.compactionConcurrency)
+	cmd.Flag("compact.blocks-fetch-concurrency", "Number of goroutines to use when download block during compaction.").
+		Default("1").IntVar(&cc.compactBlocksFetchConcurrency)
 	cmd.Flag("downsample.concurrency", "Number of goroutines to use when downsampling blocks.").
 		Default("1").IntVar(&cc.downsampleConcurrency)
 
 	cmd.Flag("delete-delay", "Time before a block marked for deletion is deleted from bucket. "+
-		"If delete-delay is non zero, blocks will be marked for deletion and compactor component will delete blocks marked for deletion from the bucket. "+
+		"If delete-delay is non zero, blocks 0will be marked for deletion and compactor component will delete blocks marked for deletion from the bucket. "+
```
yeya24, line 714: Typo?

docs/components/compact.md
```diff
@@ -343,7 +346,7 @@ Flags:
                                 it via --deduplication.func.
       --delete-delay=48h        Time before a block marked for deletion is
                                 deleted from bucket. If delete-delay is non
-                                zero, blocks will be marked for deletion and
+                                zero, blocks 0will be marked for deletion and
```
yeya24, line 349: Ditto

yeya24 (changes requested):

yeya24 (approved):
LGTM

kakkoyun (approved):

PR #5494, opened 2022-07-13
receive: Fix stats nil pointer panic

fpetkovski:
When fetching TSDB stats from receivers, certain TSDBs might not be
initialized yet. This can lead to a nil pointer access when the
status endpoint is accessed before all TSDBs are initialized.

This commit adds an explicit check for each tenant's TSDB when
exporting TSDB stats.


kakkoyun (dismissed):

squat (dismissed):

squat:
xref: https://github.com/thanos-io/thanos/pull/5495

squat:
@fpetkovski I think you will need to rebase on main to get the check to pass

PR #5498, opened 2022-07-13
Switch from Reach Router to React Router

Nexucis:
To be able to update react to the v17, we need first to replace the router used.

You can take a look at this old PR for more details : https://github.com/prometheus/prometheus/pull/9273

Note: this is still in progress as for unknown reason, the page doesn't load with no error which is super weird :(. I am probably missing something.
Maybe @onprem you have an idea why the page doesn't load ?


Nexucis:
mmm I didn't find the time to fix it. But I don't think I am so far to finish it.
I will close it and hoping I will find the time to finish it

PR #5500, opened 2022-07-13
Upgrade mdox with cache and some http settings to reduce CI failures

douglascamata:

## Changes

- `mdox` as upgraded to the most recent commit in the `master` branch: https://github.com/bwplotka/mdox/commit/25b9abcf90a064d41422f45900b389f50c1b8ec2. This head includes:
  - Cache of the link checks (added in https://github.com/bwplotka/mdox/commit/25b9abcf90a064d41422f45900b389f50c1b8ec2).
  - Important configurations for link checks (added in https://github.com/bwplotka/mdox/commit/d39d57f434dd89a279ff267c96535a8f7eedb715, see the [docs](https://github.com/bwplotka/mdox#link-validation-configuration)). For the moment we bump the max connections per host and set the timeout of each request to 1 minute. Other changes can be done later as we see how the CI behaves.
- Added a Github Actions cache for the `mdox` cache file based on the hash of the markdown files it works with.

## Verification

* Ran `make check-docs` locally.
* Verified that cache builds and restores worked fine.


.mdox.validate.yaml
```diff
@@ -1,4 +1,8 @@
 version: 1
+timeout: '1m'
+host_max_conns: 10
+cache:
+  type: 'SQLite'
```
saswatamcode, line 5: Maybe adding a `jitter` param with a duration of a few days would be good too? 🙂

.github/workflows/docs.yaml
```diff
@@ -22,10 +22,19 @@ jobs:
...
           key: ${{ runner.os }}-go-${{ hashFiles('**/go.sum') }}
+          restore-keys: |
+            ${{ runner.os }}-go-
+
+      - uses: actions/cache@v3
+        with:
+          path: .mdoxcache
+          key: ${{ runner.os }}-mdox-${{ hashFiles('docs/**/*.md', 'examples/**/*.md', 'mixin/**/*.md', '*.md') }}
+          restore-keys: |
+            ${{ runner.os }}-mdox-
```
saswatamcode, line 37: I wonder if a maintainer could check if this is actually in cache now with https://docs.github.com/en/rest/actions/cache#list-github-actions-caches-for-a-repository. Would be neat to check beforehand!

saswatamcode (dismissed):
Thanks a lot! This should make docs check way faster and less flaky! 💪🏻 

Some tiny suggestions!

.mdox.validate.yaml
douglascamata: Oh, my bad. I thought there was a default jitter. Shall we use `validity/2` as jitter? What do you think?

saswatamcode: SGTM!

douglascamata: Done in https://github.com/thanos-io/thanos/pull/5500/commits/7ec0faecec4827973a88db63ea31c549a25d8833

PR #5502, opened 2022-07-14
Receiver: Handle storage exemplar multi-error

matej-g:

## Changes

Currently, we are not considering exemplar storage errors to be conflict errors within our multi-error logic. This means that in cases when the majority of the errors comes from exemplar storage errors, these will 'fall through' and it will be determined that the error cause is internal server error (which would also cause Prometheus to retry remote write, which is not what we want in case of conflict-like storage error).

This PR proposes to handle exemplar storage errors the same way we handle sample storage errors - these should be considered conflict errors and handled appropriately in the multi-error logic as well.

## Verification

Added unit tests.


squat (approved):
Looks good to me @matej-g!

Issue #3832, opened 2021-02-24
Remove module dependency cycle between Thanos and Cortex


abursavich:
Following up here from a Slack conversation with @bwplotka (cc: @pracucci)...

## Summary of the issue

An old version of Prometheus pulled in `k8s.io/client-go@v12.0.0+incompatible`.
An old version of Alertmanager pulled in that version of Prometheus.
An old version of Cortex pulled in that version of Alertmanager.
An old version of Thanos pulled in that version of Cortex
A newer version of Cortex pulled in that version of Thanos
A newer version of Thanos pulled in that version of Cortex
...
A newer version of Cortex pulled in that version of Thanos
A newer version of Thanos pulled in that version of Cortex
A recent version of Cortex pulled in that version of Thanos
A recent version of Thanos pulled in that version of Cortex

Newer versions of Prometheus have moved to `k8s.io/client-go@v0.20.2`.
Newer versions of Alertmanager have removed the dependency on Prometheus.
Newer versions of Cortex have moved to newer versions of Alertmanager.
Newer versions of Thanos have moved to newer versions of Cortex.
Newer versions of Cortex have moved to newer versions of Thanos.

Kubernetes releases predate go modules and historical tags do not conform to go module standards.
Go treats `k8s.io/client-go@v12.0.0+incompatible` as newer than `k8s.io/client-go@v0.20.2`.
The cycle between Thanos and Cortex keeps their oldest history together alive in the dependency graph.

If you depend on Thanos or Cortex, then you get stuck with `k8s.io/client-go@v12.0.0+incompatible` and must use replace directives if you need to use a newer version such as `k8s.io/client-go@v0.20.2`.

## Solution A

Go doesn't allow dependency cycles in packages, so there must not be any. Thanos and Cortex could work together to break the module cycle my extracting shared packages into separate modules.

This is theoretically the better option, but I don't know how much work would be involved and it may add a significant maintenance burden.

## Solution B

Thanos and Cortex could work together to reset the cycle in their dependency graph. The cycle would continue to exist, but the historical dependencies would be dropped. This is probably the quicker fix and the other solution may still be undertaken in the future if desired.

Here's how this could work, which I've tested with dummy modules:

#### Step 1
- Thanos adds commit X that removes Cortex from its `go.mod` file. At this commit, Thanos won't build properly.
- Cortex adds commit Y that removes Thanos from its `go.mod` file. At this commit, Cortex won't build properly.

#### Step 2
[...]

PR #5254, opened 2022-03-24
*: Move objstore out of repo

brancz:
* [ ] I added CHANGELOG entry for this change.
* [x] Change is not relevant to the end user.

## Changes

Closes https://github.com/thanos-io/thanos/issues/3701

For this to go green we first need Cortex to migrate to the new dep.

## Verification

Code and tests are unchanged, it's just moving code.

cc @onprem @kakkoyun @bwplotka @squat @pracucci @yeya24

PR #5504, opened 2022-07-15
chore: Vendor Cortex dependency as an internal package

kakkoyun:

Introducing a cyclic dependency between Cortex and Thanos was not a good idea. It makes it really hard to upgrade dependencies for Thanos and Cortex (i.e. Prometheus). This solution is not the ideal one. It would be great to extract out `Query Frontend` packages from Cortex but as far as I can tell, it won't be easy. So ideally we should have our own.

Vendored version is the tip of the main as of 15.07.2022

Fixes https://github.com/thanos-io/thanos/issues/3832
Blocked by https://github.com/thanos-io/thanos/pull/5484
Related https://github.com/thanos-io/thanos/pull/5254

## Changes

* Vendor Cortex dependency as an internal package

## To Do

* [x] Rebase after Prometheus upgrade
* [x] Run formatter
* [x] Make sure linter is happy
* [x] Add licence headers to the vendored packages

## Verification

* `make build`
* `make test`


bwplotka (dismissed):
Thanks a lot. It's ugly solution, but the best we have now! Will unblock us in short term.

Would be keen on merging. Lgtm!

pracucci (approved):
I agree this cycle dependency has always been a pain.

squat:
haha wow according to sonatype we could contribute a lot of small enhancements to the cortex code :)

kakkoyun:
@sonatype-lift ignoreall

kakkoyun:
The CodeQL errors from the vendored packages will be ignored. I have added the necessary configuration. Yet the failure is there from previous runs.

douglascamata (approved):
LGTM :)

alvinlin123:
> I agree this cycle dependency has always been a pain.

me too :)

Issue #5497, opened 2022-07-13
Compact: crashes when finding a sample with out of order labels


douglascamata:
**Thanos**: v0.27.0-rc.0

**What happened**:

Thanos Compact found a posting with out of order labels and started crashing infinitely. 

**What you expected to happen**:

Thanos Compact finds a posting with out of order labels and then:

- Fixes it before compacting if possible.
- Otherwise, ignores it.
- Log at warning/critical level that this happened.
- Exports a metric to track this error so that alerts can be created to trigger when this happens a lot.

Most importantly, I would love Compact to not completely crash and stop doing its job.

**How to reproduce it (as minimally and precisely as possible)**:

Have a block with posting containing out of order labels and try to compact them.

**Full logs to relevant components**:

<details>Logs
<p>

```
level=warn ts=2022-07-13T13:09:33.180528476Z caller=index.go:267 msg="out-of-order label set: known bug in Prometheus 2.8.0 and below" labelset="{_id=\"test\", __name__=\"rhobs_e2e\"}" series=244977
level=warn ts=2022-07-13T13:09:33.180773659Z caller=intrumentation.go:67 msg="changing probe status" status=not-ready reason="error executing compaction: first pass of downsampling failed: downsampling to 5 min: input block index not valid: index contains 1 postings with out of order labels"
level=info ts=2022-07-13T13:09:33.180794468Z caller=http.go:84 service=http/server component=compact msg="internal server is shutting down" err="error executing compaction: first pass of downsampling failed: downsampling to 5 min: input block index not valid: index contains 1 postings with out of order labels"
level=info ts=2022-07-13T13:09:33.182943478Z caller=http.go:103 service=http/server component=compact msg="internal server is shutdown gracefully" err="error executing compaction: first pass of downsampling failed: downsampling to 5 min: input block index not valid: index contains 1 postings with out of order labels"
level=info ts=2022-07-13T13:09:33.18296634Z caller=intrumentation.go:81 msg="changing probe status" status=not-healthy reason="error executing compaction: first pass of downsampling failed: downsampling to 5 min: input block index not valid: index contains 1 postings with out of order labels"
[...]

PR #5508, opened 2022-07-15
Receiver: Validate labels in write requests

### LINKED ISSUE #5499: Receiver: Metrics with out-of-order labels get accepted
Latest Thanos (`0.27.0`)

**What happened**:
Receiver happily accepts metrics with out-of-order labels. By itself, this seemed to be acceptable (i.e. TSDB will append them without complaining). But this is causing issues in compactor, which expects labels to be in order and this causes it to get into a crash loop.

This is similar to situation that happened previously with https://github.com/prometheus/prometheus/pull/5372. Though we're in remote write land and we may assume in most cases we are getting metrics from 'official' clients that should adhere to the labels ordering, this cannot be guaranteed with all clients (as is evidenced by our related issue https://github.com/thanos-io/thanos/issues/5497, which we caused by running some custom tests with an 'unofficial' client from https://github.com/observatorium/up).

I would expect the receiver to handle this, at least in a minimal fashion, to prevent potential issues with compactor down the line. Possible options:
- Silently fix label ordering
- Warn about wrong label order
- Both of the above
- Refuse request as bad request
- Make this configurable

matej-g:

## Changes

This PR proposes to introduce a label verification in the receive writer. The justification is that, even though the remote write specification has not been yet fully finalized (see [doc](https://docs.google.com/document/d/1LPhVRSFkGNSuU1fBd81ulhsCPR4hkSZyyBj1SZ8fWOM/edit#)), there are already cases where we are accepting write requests, without checking labels, which is causing issues down the line, e.g. due to invalid index (see related issues #5497 and #5499). Although there is not an official specification for remote write as of yet, due to these reasons I believe we should not be condoning 'bad behavior' of clients and we should not be accepting write requests with problematic labels.

At minimum, I suggest to check for cases of empty, duplicate and out-of-order labels, which is consistent with current remote write spec draft (excerpt from the link above):
```
- MUST NOT contain repeated label names.
- MUST have label names sorted in lexicographical order.
- MUST NOT contain any empty label names or values.
```
Similarly how samples are handled, series with offending labels shall be dropped from the request (while samples for series with valid labels will be appended). Any of the above mentioned labels errors shall be considered to be conflict errors (similarly to sample and exemplar storage error), resulting in same handling as other conflict errors (i.e. we return `409` indicating that the request should not be retried).

On a related note, this PR also renames the confusing method `PrometheusIssue5372Err` to `OutOfOrderLabelsErr` error.

Resolves #5499

## Verification
Added unit tests for all related parts of code (label validation, remote writer, conflict error handling)


pkg/store/labelpb/label.go
```diff
@@ -346,6 +352,44 @@ func HashWithPrefix(prefix string, lbls []ZLabel) uint64 {
 	return xxhash.Sum64(b)
 }
 
+// ValidateLabels validates label names and values (checks for empty
+// names and values, out of order labels and duplicate label names)
+// Returns appropriate error if validation fails on a label.
+func ValidateLabels(lbls []ZLabel) error {
```
fpetkovski, line 358: Any concerns with increasing write latency due to this added validation?

squat: We could benchmark it so that we can make an informed decision about how much the latency changes. Maybe the validation is negligible compared to the storing or the metrics in the TSDB, maybe not.

fpetkovski: Benchmarking if possible would be great. If that's too complicated, I can keep an eye on this after deploying.

Now that you mentioned TSDB, I wonder why it even accepts writing samples with OOO labels. Should we also open an issue in prometheus/prometheus?

matej-g: I think that is a valid concern, I'd see how complicated it would be to add a benchmark for this.

As for TSDB, I'm not sure whether this was ever a requirement or a problem, I've seen various reference to ordering sprinkled throughout the repo / codebase (https://github.com/prometheus/prometheus/issues/8861#issuecomment-855885421, https://github.com/prometheus/prometheus/pull/10532, https://github.com/prometheus/prometheus/pull/5372), but I cannot find any authoritative doc saying how labels must be ordered at time of ingestion.

pkg/store/labelpb/label.go
```diff
@@ -346,6 +352,44 @@ func HashWithPrefix(prefix string, lbls []ZLabel) uint64 {
...
+
+	// Check first label.
+	l0 := lbls[0]
+	if l0.Name == "" || l0.Value == "" {
+		return ErrEmptyLabels
+	}
+	labelNames[l0.Name] = struct{}{}
+
+	// Iterate over the rest, check each for empty / duplicates and
+	// check lexographical ordering.
```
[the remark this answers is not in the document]
matej-g: Good catch :+1: I'll fix the typo. I'd like to retain the `lexicographical` term, as it seems to be a rather standard term and has a specific meaning for the ordering (e.g. in lexicographical order, uppercase letters come before lowercase letters, which might be ambiguous if we just say it should be ordered alphabetically). But perhaps adding alphabetical in parenthesis would make sense to make it more understandable. Sounds good?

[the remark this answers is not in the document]
douglascamata: Sounds good. 👍

pkg/store/labelpb/label.go
matej-g: I actually realized that we can simplify the validation, since we are expecting the labels to be ordered, we can first check order and then duplication, meaning we don't have to use a map to check duplicates (saving us allocations).

I ran benchmarks from https://github.com/thanos-io/thanos/pull/5533 on old and new code and there's no discernible difference:
```
name                                     old time/op    new time/op    delta
WriterTimeSeriesWithSingleLabel_10-12      4.08µs ± 0%    4.02µs ± 0%   ~     (p=1.000 n=1+1)
WriterTimeSeriesWithSingleLabel_100-12     22.6µs ± 0%    21.3µs ± 0%   ~     (p=1.000 n=1+1)
WriterTimeSeriesWithSingleLabel_1000-12     238µs ± 0%     197µs ± 0%   ~     (p=1.000 n=1+1)
WriterTimeSeriesWith10Labels_10-12         6.18µs ± 0%    5.57µs ± 0%   ~     (p=1.000 n=1+1)
WriterTimeSeriesWith10Labels_100-12        37.9µs ± 0%    41.1µs ± 0%   ~     (p=1.000 n=1+1)
WriterTimeSeriesWith10Labels_1000-12        368µs ± 0%     430µs ± 0%   ~     (p=1.000 n=1+1)

name                                     old alloc/op   new alloc/op   delta
WriterTimeSeriesWithSingleLabel_10-12        411B ± 0%      411B ± 0%   ~     (all equal)
WriterTimeSeriesWithSingleLabel_100-12       413B ± 0%      411B ± 0%   ~     (p=1.000 n=1+1)
WriterTimeSeriesWithSingleLabel_1000-12      454B ± 0%      453B ± 0%   ~     (p=1.000 n=1+1)
WriterTimeSeriesWith10Labels_10-12           410B ± 0%      410B ± 0%   ~     (all equal)
WriterTimeSeriesWith10Labels_100-12          411B ± 0%      412B ± 0%   ~     (p=1.000 n=1+1)
WriterTimeSeriesWith10Labels_1000-12         460B ± 0%      471B ± 0%   ~     (p=1.000 n=1+1)

name                                     old allocs/op  new allocs/op  delta
WriterTimeSeriesWithSingleLabel_10-12        9.00 ± 0%      9.00 ± 0%   ~     (all equal)
WriterTimeSeriesWithSingleLabel_100-12       9.00 ± 0%      9.00 ± 0%   ~     (all equal)
WriterTimeSeriesWithSingleLabel_1000-12      9.00 ± 0%      9.00 ± 0%   ~     (all equal)
WriterTimeSeriesWith10Labels_10-12           9.00 ± 0%      9.00 ± 0%   ~     (all equal)
WriterTimeSeriesWith10Labels_100-12          9.00 ± 0%      9.00 ± 0%   ~     (all equal)
WriterTimeSeriesWith10Labels_1000-12         9.00 ± 0%      9.00 ± 0%   ~     (all equal)
```

pkg/receive/writer.go
```diff
@@ -70,6 +75,25 @@ func (r *Writer) Write(ctx context.Context, tenantID string, wreq *prompb.WriteR
...
+			case labelpb.ErrDuplicateLabels:
+				numLabelsDuplicates++
+				level.Debug(tLogger).Log("msg", "Duplicate labels in the label set", "lset", t.Labels)
+			case labelpb.ErrEmptyLabels:
+				numLabelsEmpty++
+				level.Debug(tLogger).Log("msg", "Labels with empty name in the label set", "lset", t.Labels)
+			}
+
+			continue
+		}
```
bwplotka, line 95: Let's hide err var if we can

```suggestion
		if err := labelpb.ValidateLabels(t.Labels); err != nil {
			switch err {
			case labelpb.ErrOutOfOrderLabels:
				numLabelsOutOfOrder++
				level.Debug(tLogger).Log("msg", "Out of order labels in the label set", "lset", t.Labels)
			case labelpb.ErrDuplicateLabels:
				numLabelsDuplicates++
				level.Debug(tLogger).Log("msg", "Duplicate labels in the label set", "lset", t.Labels)
			case labelpb.ErrEmptyLabels:
				numLabelsEmpty++
				level.Debug(tLogger).Log("msg", "Labels with empty name in the label set", "lset", t.Labels)
			}

			continue
		}
```

pkg/receive/writer.go
```diff
@@ -70,6 +75,25 @@ func (r *Writer) Write(ctx context.Context, tenantID string, wreq *prompb.WriteR
...
+				level.Debug(tLogger).Log("msg", "Out of order labels in the label set", "lset", t.Labels)
+			case labelpb.ErrDuplicateLabels:
+				numLabelsDuplicates++
+				level.Debug(tLogger).Log("msg", "Duplicate labels in the label set", "lset", t.Labels)
+			case labelpb.ErrEmptyLabels:
+				numLabelsEmpty++
+				level.Debug(tLogger).Log("msg", "Labels with empty name in the label set", "lset", t.Labels)
+			}
+
+			continue
```
bwplotka, line 94: Should we add any default case just in case someone modify verifier and forgets to update here? (:

bwplotka (approved):
Amazing work and thanks for optimizations 💪🏽

pkg/receive/writer.go
matej-g: Good idea :+1:

bwplotka (approved):
Thanks 💪🏽

douglascamata (approved):
:shipit: 🚀 🌔

PR #5510, opened 2022-07-15
*: Move objstore out of repo

### LINKED ISSUE #3701: Put `objstore` outside of Thanos main repo?
Rationales:

* Even with Go 1.7 importing whole Thanos module is... hard. Lot's of `replace` statements and overall many dependencies.
* Release model is not within semver semantics.

We can actually have objstore as separate module. That might work as long as it's hermetic and does not use further Thanos imports. This will help importability but release model would be equally bad as now.

This module is so well defined and establish that it would be worth to main in separe repo.

Wonder @pracucci if it's not something Cortex and other projects would be interested more. Most likely it would improve chances for external maintainers just of this focused module. I definitely would love to use for other smaller OSS project, but don't want to due to Thanos deps. 🤔 

@brancz @kakkoyun @squat

kakkoyun:

## Changes

Fixes #3701 
Supercedes https://github.com/thanos-io/thanos/pull/5254

For this to go green we first need Cortex to migrate to the new dep.

## Verification

Code and tests are unchanged, it's just moving code.

cc  @bwplotka @brancz @yeya24 @GiedriusS


yeya24 (approved):
LGTM. Let's merge this on green.

PR #5516, opened 2022-07-18
chore: Update Go modules

SuperQ:

## Changes

* Update weaveworks/common to remove node_exporter indirect dep.
* Update simonpasquier/klog-gokit/v2.
* Update google.golang.org/grpc lock to v1.45.0.
* Cleanup replacements that are now handled by indirect requirements.
* Fixup grpc.WithInsecure() use.


squat (approved):
Thanks @SuperQ ❤️

PR #5520, opened 2022-07-19
Meta-monitoring based active series limiting

saswatamcode:
This PR adds optional meta-monitoring based active series limiting for tenants using Receive Router. As discussed in proposal #5415. 🙂 

## Changes

- Added four new flags to Receive,
  - `receive.tenant-limits.max-head-series`: Active series limit to set for all tenants.
  - `receive.tenant-limits.meta-monitoring-url`: Prometheus Query API compatible meta-monitoring solution.
  - `receive.tenant-limits.meta-monitoring-query`: Query to execute against meta-monitoring to get active series data. Optional.
  - `receive.tenant-limits.meta-monitoring-client`: HTTP Client config file for connecting to meta-monitoring 
- Added `QueryMetamonitoring` method to Receive handler which makes Prometheus Query to meta-monitoring endpoint every 15s and caches result
- Added `isUnderLimit` method to Receive handler, refers to cache and limits remote write requests in Router/RouterIngestor mode
- Five new receive e2e tests to try out such limiting with Avalanche 

## Verification

Locally + e2e tests.


cmd/thanos/receive.go
```diff
@@ -821,6 +831,18 @@ func (rc *receiveConfig) registerFlag(cmd extkingpin.FlagClause) {
...
+	cmd.Flag("receive.limit-meta-monitoring.url", "Meta-monitoring URL which is compatible with Prometheus Query API for active series limiting.").Default("http://localhost:9090").URLVar(&rc.metaMonitoringUrl)
+
+	cmd.Flag("receive.limit-meta-monitoring.query", "PromQL Query to execute against meta-monitoring, to get the current number of active series for each tenant, across Receive replicas.").Default("sum(prometheus_tsdb_head_series) by (tenant)").StringVar(&rc.metaMonitoringLimitQuery)
+
+	rc.metaMonitoringHttpClient = extflag.RegisterPathOrContent(
+		cmd,
+		"receive.limit-meta-monitoring.http-client",
+		"YAML file or string with http client configs for meta-monitoring.",
+	)
+
```
fpetkovski, line 845: I would suggest grouping all tags under the same namespace. Something like
```
receive.tenant-limits.series-limit
receive.tenant-limits.meta-monitoring-url
receive.tenant-limits.meta-monitoring-query
receive.tenant-limits.meta-monitoring-client
```

pkg/receive/handler.go
```diff
@@ -165,6 +176,24 @@ func NewHandler(logger log.Logger, o *Options) *Handler {
 				Help: "The number of times to replicate incoming write requests.",
 			},
 		),
+		configuredTenantLimit: promauto.With(registerer).NewGauge(
+			prometheus.GaugeOpts{
+				Name: "thanos_receive_tenant_active_series_limit",
```
fpetkovski, line 181: I think using head series is better for consistency with other metrics.

pkg/receive/handler.go
```diff
@@ -456,6 +487,21 @@ func (h *Handler) receiveHTTP(w http.ResponseWriter, r *http.Request) {
 		return
 	}
 
+	// Impose limits only if Receive is in Router or RouterIngestor mode.
+	if h.receiverMode == RouterOnly || h.receiverMode == RouterIngestor {
```
fpetkovski, line 491: Curious, why can't we have this feature in the dual mode as well?

pkg/receive/handler.go
```diff
@@ -482,6 +528,63 @@ func (h *Handler) receiveHTTP(w http.ResponseWriter, r *http.Request) {
...
+	if err != nil {
+		return true, errors.Wrap(err, "getting http client config")
+	}
+
+	httpClientConfig, err := httpconfig.NewClientConfigFromYAML(httpConfContentYaml)
+	if err != nil {
+		return true, errors.Wrap(err, "parsing http config YAML")
+	}
+
+	httpClient, err := httpconfig.NewHTTPClient(*httpClientConfig, "thanos-receive")
```
fpetkovski, line 550: I'm not sure if we should be creating a new client per remote-write request. This can be quite wasteful in terms of memory and connection use. Can we create the client when the receiver starts?

fpetkovski (commented):
Amazing stuff

pkg/receive/handler.go
```diff
@@ -165,6 +176,24 @@ func NewHandler(logger log.Logger, o *Options) *Handler {
...
 		),
+		configuredTenantLimit: promauto.With(registerer).NewGauge(
+			prometheus.GaugeOpts{
+				Name: "thanos_receive_tenant_active_series_limit",
+				Help: "The configured limit for active series of tenants.",
+			},
+		),
+		aboveLimit: promauto.With(registerer).NewGaugeVec(
+			prometheus.GaugeOpts{
+				Name: "thanos_receive_series_above_limit",
```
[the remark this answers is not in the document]
saswatamcode: So this metric exposes the difference between the current head series and the limit per tenant. Do you mean that we can derive this via some other PromQL expression? 🙂

cmd/thanos/receive.go
douglascamata: > receive.tenant-limits.series-limit

I would suggest to name this something like `receive.tenant-limits.max-active-series`. Reading only `series-limit` doesn't tell me they are active series.

pkg/receive/handler.go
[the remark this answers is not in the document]
fpetkovski: Yes, that's what I was thinking. It seems like a redundant metric which can be derived in PromQL by subtracting head_series from series_limit or vice versa, depending on what the user wants to see.

cmd/thanos/receive.go
fpetkovski: Agreed, sounds much better.

pkg/receive/handler.go
fpetkovski: Absolutely. Querying periodically and caching the result is the way to go 👍

pkg/receive/handler.go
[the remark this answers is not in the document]
saswatamcode: Will update! 🙂

pkg/receive/handler.go
saswatamcode: By dual mode, you mean why not also in IngestorOnlyr?

fpetkovski: Ah yes, dual mode seems to be `RouterIngestor`. I guess we don't need limiting for `IngestorOnly` because the router will tackle that?

saswatamcode: Yes exactly. 🙂

fpetkovski: Got it, thanks 👍

docs/components/receive.md
```diff
@@ -77,6 +77,20 @@ The example content of `hashring.json`:
...
+
+This can used by specifying the following flags,
+- `--receive.per-tenant-limit`: Specifies the total number of active series for any tenant, across all replicas (including data replication), allowed by Thanos Receive.
+- `--receive.limit-meta-monitoring.url`: Specifies Prometheus Query API compatible meta-monitoring endpoint.
+- `--receive.limit-meta-monitoring.query`: Optional flag to specify PromQL query to execute against meta-monitoring.
+- `receive.limit-meta-monitoring.http-client`: Optional YAML file/string specifying HTTP client config for meta-monitoring.
+
+NOTE:
+- It is possible that Receive ingests more active series than the specified limit, as it relies on meta-monitoring, which may not have the latest data for current number of active series of a tenant at all times.
+- Thanos Receive performs best-effort limting. In case meta-monitoring is down/unreachable, Thanos Receive will not impose limits.
```
pedro-stanaka, line 92: typo: 
```suggestion
- Thanos Receive performs best-effort limiting. In case meta-monitoring is down/unreachable, Thanos Receive will not impose limits.
```

docs/components/receive.md
```diff
@@ -77,6 +77,20 @@ The example content of `hashring.json`:
 
 With such configuration any receive listens for remote write on `<ip>10908/api/v1/receive` and will forward to correct one in hashring if needed for tenancy and replication.
 
+## Limiting
+
+Thanos Receive, in Router mode, supports limiting tenant active series, to maintain stability of the system. It uses any Prometheus Query API compatible meta-monitoring solution to get the current number of active series, and compares that with a configured limit, before ingesting any tenant's remote write request. In case a tenant has gone above the limit, their remote write requests are failed fully.
+
+This can used by specifying the following flags,
```
pedro-stanaka, line 84: nit (passive voice): 

```suggestion
To use the feature, one should specify the following flags:
```

pkg/receive/handler.go
```diff
@@ -482,6 +528,63 @@ func (h *Handler) receiveHTTP(w http.ResponseWriter, r *http.Request) {
...
+	// series. As such metric is updated in intervals, it is possible
+	// that Receive ingests more series than the limit, before detecting that
+	// a tenant has exceeded the set limits.
+	for _, e := range vectorRes {
+		for k, v := range e.Metric {
+			// Search for metric which has tenant label for a particular tenant.
+			if k == "tenant" && string(v) == tenant {
+				h.aboveLimit.WithLabelValues(tenant).Set(float64(e.Value) - float64(h.options.MaxPerTenantLimit))
+				if float64(e.Value) >= float64(h.options.MaxPerTenantLimit) {
+					level.Error(logger).Log("msg", "tenant above limit", "currentSeries", float64(e.Value))
```
pedro-stanaka, line 575: Would be nice to have the limit on this log line, so we don't need to check elsewhere for the limit.

pkg/receive/handler.go
```diff
@@ -456,6 +487,21 @@ func (h *Handler) receiveHTTP(w http.ResponseWriter, r *http.Request) {
...
+	if h.receiverMode == RouterOnly || h.receiverMode == RouterIngestor {
+		under, err := h.isUnderLimit(ctx, tenant, tLogger)
+		if err != nil {
+			level.Info(tLogger).Log("msg", "error while limiting", "err", err.Error())
+
+		}
+
+		// Fail request fully if tenant has exceeded set limit.
+		if !under {
+			http.Error(w, "tenant is above active series limit", http.StatusTooManyRequests)
```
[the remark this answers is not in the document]
saswatamcode: Yes, that is essentially what a client like Prometheus does, it retries up til 2 hours: https://prometheus.io/docs/practices/remote_write/#remote-write-characteristics. So the client would need to retry until their head series metric goes below set limit. And Thanos fails the request fully, so there is no confusion of partial writes.

To answer your comment question, yes this is the approach chosen and the proposal has been updated to reflect that. The other approaches are listed as alternatives. 🙂

pkg/receive/handler.go
```diff
@@ -482,6 +528,63 @@ func (h *Handler) receiveHTTP(w http.ResponseWriter, r *http.Request) {
...
+
+	httpClient, err := httpconfig.NewHTTPClient(*httpClientConfig, "thanos-receive")
+	if err != nil {
+		return true, errors.Wrap(err, "improper http client config")
+	}
+
+	c := promclient.NewWithTracingClient(logger, httpClient, httpconfig.ThanosUserAgent)
+
+	vectorRes, _, err := c.QueryInstant(ctx, h.options.MetaMonitoringUrl, h.options.MetaMonitoringLimitQuery, time.Now(), promclient.QueryOptions{})
+	if err != nil {
```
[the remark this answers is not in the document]
saswatamcode: Good idea! Will update! 🙂

pkg/receive/handler.go
saswatamcode: Yes, will add! 🙂

docs/components/receive.md
```diff
@@ -77,6 +77,20 @@ The example content of `hashring.json`:
...
+- `--receive.per-tenant-limit`: Specifies the total number of active series for any tenant, across all replicas (including data replication), allowed by Thanos Receive.
+- `--receive.limit-meta-monitoring.url`: Specifies Prometheus Query API compatible meta-monitoring endpoint.
+- `--receive.limit-meta-monitoring.query`: Optional flag to specify PromQL query to execute against meta-monitoring.
+- `receive.limit-meta-monitoring.http-client`: Optional YAML file/string specifying HTTP client config for meta-monitoring.
```
douglascamata, line 88:
```suggestion
- `--receive.limit-meta-monitoring.http-client`: Optional YAML file/string specifying HTTP client config for meta-monitoring.
```

docs/components/receive.md
```diff
@@ -77,6 +77,20 @@ The example content of `hashring.json`:
 
 With such configuration any receive listens for remote write on `<ip>10908/api/v1/receive` and will forward to correct one in hashring if needed for tenancy and replication.
 
+## Limiting
+
+Thanos Receive, in Router mode, supports limiting tenant active series, to maintain stability of the system. It uses any Prometheus Query API compatible meta-monitoring solution to get the current number of active series, and compares that with a configured limit, before ingesting any tenant's remote write request. In case a tenant has gone above the limit, their remote write requests are failed fully.
```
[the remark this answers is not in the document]
saswatamcode: Do you mean we should mention something like "meta-monitoring should scrape all Thanos Receive instances"? 🙂 
Will add that!

But as for how to set up meta-monitoring solution in general, I think that would vary by use case.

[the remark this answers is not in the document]
saswatamcode: "Meta-monitoring" under the Thanos docs [Operating](https://thanos.io/tip/operating/) section would be a good idea, and it could specify how to set up meta-monitoring for different Thanos architectures. This would probably need a dedicated different PR. 🙂

pkg/receive/handler.go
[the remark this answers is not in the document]
saswatamcode: Removed in https://github.com/thanos-io/thanos/pull/5520/commits/b647b541e4202fd6c91967fad20f35f2a94a8363

pkg/receive/handler.go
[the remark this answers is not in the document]
saswatamcode: Added in https://github.com/thanos-io/thanos/pull/5520/commits/b647b541e4202fd6c91967fad20f35f2a94a8363

pkg/receive/handler.go
saswatamcode: Added in https://github.com/thanos-io/thanos/pull/5520/commits/b647b541e4202fd6c91967fad20f35f2a94a8363

docs/components/receive.md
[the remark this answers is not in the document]
saswatamcode: Updated in https://github.com/thanos-io/thanos/pull/5520/commits/b647b541e4202fd6c91967fad20f35f2a94a8363

pkg/receive/handler.go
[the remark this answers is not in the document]
saswatamcode: Yes, I was also confused about what status code would be suitable here. I saw that an explicit config option exists, as you mentioned, which is why I used 429. But maybe there is some better default here? 🤔 

So once a tenant hits their limit, all their remote write requests would fail. So we would not ingest any data, but would still have to partially process the retry requests (essentially till `isUnderLimit`). Once the metric decreases in value to under the limit, we can start ingesting again. I kind of wanted to replicate such an error state in a dedicated e2e test, but haven't been able to figure out how to do so via containers. Will still try to see if there's a clever but accurate way of doing this!  🙂

pkg/receive/handler.go
saswatamcode: Still need to do this!  🙂

pkg/receive/handler.go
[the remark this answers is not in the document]
douglascamata: The proposal states that:

> Thus, any remote write request can be failed completely, with a 429 status code and appropriate error message, **if it increases the number of active series above the configured limit for a tenant**

So not all their remote write requests would fail. Those requests that won't create any net-new head series should still be accepted.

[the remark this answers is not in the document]
douglascamata: I don't think there is a good HTTP code for this. 

HTTP code 413 for Entity Too Large doesn't make sense as the "entity" in the request can probably be accepted at a later time. A request that gets a 413 should never be retried.

We have 422 for Unprocessable Entity, but I never seen it being used in cases of limits. 

I think we can stick to the 429, which is the closest to the meaning of the situation here. We can include a `Retry-After` header to help clients stop retrying non-stop and give the servers some rest. 😴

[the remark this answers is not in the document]
saswatamcode: > So not all their remote write requests would fail. Those requests that won't create any net-new head series should still be accepted.

That isn't possible in this approach as we don't check how much a remote write request increases the number of head series. But you're right, there's an error in the proposal. Let me update! 🙂

pkg/receive/handler.go
saswatamcode: So in https://github.com/thanos-io/thanos/pull/5520/commits/96a821eccd0f5a25b549a492237910d37edbc988, I added the query logic inside of a `runutil.Repeat` which executes the query every 15s, and caches the result in a map in tenant_id:currentSeries format. This map is used in `isUnderLimit` method.

I was a bit worried, that in a hashring setup this might have sync issues, with a late-started Receive having outdated cache. So I added a test for this too. 🙂 

Let me know if this makes sense, or if the previous approach is preferred here!

pkg/receive/handler.go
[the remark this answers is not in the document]
douglascamata: > I'm also not sure if returning retry after makes sense because we don't know when in the future the limit will be lifted, given this is not a time-based rate limiting (and besides Prometheus has internal retry mechanism with backoff).

The point of the retry after header in this case is to avoid badly configured Prometheus remote write clients spamming retries. It would work as a self-defense mechanism for Receive, as long as the client respects it (optional maybe?). 

> 429 - this means we won't be able to predict whether client retries on failed requests (Do we want the client to retry in all cases? Are we fine with having this optional? Can we ensure clients are aware of this behavior in receiver?)

Prometheus remote write client already retries on 429 if they are configured to do so, this stays the same. Clients are aware that they might be limited and decide if they want to retry as their data might be accepted later on -- it's a temporary limit. 

I don't see any reason for a different behavior.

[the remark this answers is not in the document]
matej-g: > The point of the retry after header in this case is to avoid badly configured Prometheus remote write clients spamming retries. It would work as a self-defense mechanism for Receive, as long as the client respects it (optional maybe?).

That feels a bit to me like a misuse of the header. As a client, I'd have legitimate expectation that my request (if it's not a bad request) will go through after specified time limit. But we cannot guarantee that in this case, because we don't know when this will happen. If we want to prevent abuse by 'bad' clients in general, perhaps there are other mechanisms worth exploring, outside of the scope of this PR.

> Prometheus remote write client already retries on 429 if they are configured to do so, this stays the same. Clients are aware that they might be limited and decide if they want to retry as their data might be accepted later on -- it's a temporary limit.

I think I'm just wondering how temporary is the limit. This basically refers back to what we discussed with @saswatamcode above in https://github.com/thanos-io/thanos/pull/5520#discussion_r925146840./ How long until an active series becomes 'free' for a tenant? (Maybe it would be good to document this so users understand better what constitutes an active series and how long it is active). Could it be that in an extreme case, client _A_ might never get their request in, because once the limit drops, it gets filled by another client _X_ and thus  client _A_ will always see `429`? It feels like the clients would be constantly 'fighting' for the few spot. These are still some of the aspects I'm trying to grasp my head around and wonder whether they could pose issues, maybe I just a need a clarification on my understanding of the limits.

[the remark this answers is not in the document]
douglascamata: >  How long until an active series becomes 'free' for a tenant? 

100% we have to document this. I found in [Grafana's docs](https://grafana.com/docs/grafana-cloud/fundamentals/active-series-and-dpm/#active-series) that `For Prometheus, a time series is considered active if new data points have been generated within the last 15 to 30 minutes.`. I couldn't find any official information about in Prometheus' it docs though. 

So the temporary aspect of the limit could range from almost 0 to almost 30 minutes. 

> Could it be that in an extreme case, client A might never get their request in, because once the limit drops, it gets filled by another client X and thus client A will always see 429? It feels like the clients would be constantly 'fighting' for the few spot.

This could happen. It's a common problem of multiple clients under the same rate limiting group being very active.

In a non-distributed, one way to manage this is to have a central function/class/whatever that will receive requests from different sources and weigh in how many from each will be sent to the server. This way it could try to be fair and give each source 50% of the available limit.

In a distributed system it's more difficult to coordinate this. Even more difficult in the remote write context, as clients often are completely isolate from each other.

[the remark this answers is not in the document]
saswatamcode: > Could it be that in an extreme case, client A might never get their request in, because once the limit drops, it gets filled by another client X and thus client A will always see 429? It feels like the clients would be constantly 'fighting' for the few spot.

So yes, this can happen with client X and client A both sending remote write requests with tenant headers for the same tenant and one of them essentially getting "starved" due to limiting and having to fight for some HEAD series spots. But that is kind of the point of limiting I think, where we protect Thanos ingestion system health, by blocking requests. 🙂 

If a non-admin persona Thanos user notices that a client of theirs cannot seem to send metrics due to limits being hit continually, they would have a few solutions,

- Request admin persona Thanos user to increase limits so their starved client actually gets a chance to send metrics
- Re-evaluate what metrics are being sent by their clients and maybe apply metric keep/drop configs accordingly
- Load balance remote write requests to give "starved" clients a fair chance

> This could happen. It's a common problem of multiple clients under the same rate limiting group being very active.
In a non-distributed system, one way to manage this is to have a central function/class/whatever that will receive requests from different sources and weigh in how many from each will be sent to the server. This way it could try to be fair and give each source 50% of the available limit.

Thanos Receive is client-agnostic, and only tracks tenants as separate TSDBs need to be created for each of them. How a particular "quota" of HEAD series is being used by a tenant should perhaps be a client-side issue, in this kind of system? 🙂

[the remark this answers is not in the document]
douglascamata: > How a particular "quota" of HEAD series is being used by a tenant should perhaps be a client-side issue, in this kind of system?

It is totally the client's responsibility. What I mean is that if the client is a distributed system, it's non-trivial to ensure fairness in this scenario. There's nothing we can do.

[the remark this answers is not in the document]
matej-g: I think we have agreement on that it's client's problem, but I'm still focusing on the server side. Imagine we have 10 Prometheus instances writing to us periodically, but we know 9 out those will not be able to get their requests in (in the imagined "race for the free spot" scenario, as I described above) because of the limit, but we'll still be like "yes please, keep spamming us with requests although we will refuse them with very high probability".

And that's where I'm wondering, are we fine with having clients continuously sending us pointless requests? Does this pose a performance issue? Is there a better way to communicate to clients how long to wait / when a series limit is lifted? For these I don't have good answers :upside_down_face:.

pkg/receive/handler.go
```diff
@@ -482,6 +548,62 @@ func (h *Handler) receiveHTTP(w http.ResponseWriter, r *http.Request) {
...
+	if h.options.MaxPerTenantLimit == 0 || h.options.MetaMonitoringUrl.Host == "" {
+		return true, nil
+	}
+
+	// In such limiting flow, we ingest the first remote write request
+	// and then check meta-monitoring metric to ascertain current active
+	// series. As such metric is updated in intervals, it is possible
+	// that Receive ingests more series than the limit, before detecting that
+	// a tenant has exceeded the set limits.
+	v, ok := h.tenantCurrentSeriesMap[tenant]
```
matej-g, line 591: It looks like that after switching to periodical querying, here we're reading from the map and we write to it `QueryMetaMonitoring` in a separate goroutine. Should ensure this is concurrency safe by syncing with mutex?

pkg/receive/handler.go
```diff
@@ -482,6 +548,62 @@ func (h *Handler) receiveHTTP(w http.ResponseWriter, r *http.Request) {
...
+	if !ok {
```
matej-g, line 592: In this case, are we taking into consideration this could be the first request from a new tenant? This should not be considered an error I think.

pkg/receive/handler.go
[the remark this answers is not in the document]
saswatamcode: > And that's where I'm wondering, are we fine with having clients continuously sending us pointless requests? Does this pose a performance issue? Is there a better way to communicate to clients how long to wait / when a series limit is lifted? For these I don't have good answers

Let me look into this a bit more! While we fail requests early when limit is hit, there might still be some performance problems. Maybe we cannot mitigate via server-side alone, so would need to suggest using some client config options like backoff/retries on client-side.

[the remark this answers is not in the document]
matej-g: Even strictly outside of Thanos, I'm wondering about impact just in the terms of all the overhead of making wasteful HTTP requests from many clients for extended period of time (maybe I'm being overly worried :shrug:).

With client side, that could work, but if we consider that we are moving into territory where we do not control the clients, it's harder to enforce such configs (otherwise we could just force relabel configuration on the client and we would not need series limits after all :slightly_smiling_face:).

Anyways, I don't want this to lead to a never ending cycle of conversation. Let's see what will be @saswatamcode's findings :bow:.

pkg/receive/handler.go
```diff
@@ -482,6 +548,62 @@ func (h *Handler) receiveHTTP(w http.ResponseWriter, r *http.Request) {
...
+
+	vectorRes, _, err := c.QueryInstant(ctx, h.options.MetaMonitoringUrl, h.options.MetaMonitoringLimitQuery, time.Now(), promclient.QueryOptions{})
+	if err != nil {
+		h.metaMonitoringErr.Inc()
+		return err
+	}
+
+	level.Debug(h.logger).Log("msg", "successfully queried meta-monitoring", "vectors", len(vectorRes))
+
+	// Construct map of tenant name and current HEAD series.
```
bwplotka, line 565:
```suggestion
	// Construct map of tenant name and current active series.
```

Uppercase HEAD looks like HTTP HEAD method - let's avoid this confusion.

pkg/receive/handler.go
```diff
@@ -196,6 +227,27 @@ func NewHandler(logger log.Logger, o *Options) *Handler {
...
+		if err != nil {
+			level.Error(h.logger).Log("msg", "getting http client config", "err", err.Error())
+		}
+
+		httpClientConfig, err := httpconfig.NewClientConfigFromYAML(httpConfContentYaml)
+		if err != nil {
+			level.Error(h.logger).Log("msg", "parsing http config YAML", "err", err.Error())
+		}
+
+		h.metaMonitoringClient, err = httpconfig.NewHTTPClient(*httpClientConfig, "thanos-receive")
```
bwplotka, line 245:
```suggestion
		h.metaMonitoringClient, err = httpconfig.NewHTTPClient(*httpClientConfig, "meta-mon-for-limit")
```

I believe this is a client name, so let's make it explicit. The fact we scrape it from thanos-receive will be obvious with `job`

pkg/receive/handler.go
```diff
@@ -165,6 +178,24 @@ func NewHandler(logger log.Logger, o *Options) *Handler {
 				Help: "The number of times to replicate incoming write requests.",
 			},
 		),
+		configuredTenantLimit: promauto.With(registerer).NewGauge(
```
bwplotka, line 181: I think it's time to perhaps move it to it's own struct, especially given races in map we have to stop and background querying.

pkg/receive/handler.go
```diff
@@ -482,6 +548,62 @@ func (h *Handler) receiveHTTP(w http.ResponseWriter, r *http.Request) {
...
+		return true, errors.New("tenant not in current series map")
+	}
+
+	if v >= float64(h.options.MaxPerTenantLimit) {
+		level.Error(logger).Log("msg", "tenant above limit", "currentSeries", v, "limit", h.options.MaxPerTenantLimit)
+		h.limitedRequests.WithLabelValues(tenant).Inc()
+		return false, nil
+	}
+
+	level.Debug(logger).Log("msg", "tenant is under limit", "currentSeries", v)
```
bwplotka, line 602: Let's remove it, not very useful and it's critical path.

test/e2e/receive_test.go
```diff
@@ -602,4 +605,702 @@ func TestReceive(t *testing.T) {
 			},
 		})
 	})
+
+	t.Run("single_active_series_limiting", func(t *testing.T) {
```
bwplotka, line 609: This adds a lot of new tests to maintain. 

Adding  limit check to our previous test is probably overkill given eventuallity, but maybe? I would be also fine with just one most complex arch test instead of five.

cmd/thanos/receive.go
```diff
@@ -293,6 +298,23 @@ func runReceive(
 		)
 	}
 
+	if (receiveMode == receive.RouterOnly || receiveMode == receive.RouterIngestor) && conf.maxPerTenantLimit != 0 {
+		level.Debug(logger).Log("msg", "setting up periodic meta-monitoring query for limiting cache")
```
bwplotka, line 302: This can be info level. Let's also mention the threshold.

bwplotka (commented):
Amazing work! Some comments, but otherwise, nice iteration!

cmd/thanos/receive.go
```diff
@@ -821,6 +848,18 @@ func (rc *receiveConfig) registerFlag(cmd extkingpin.FlagClause) {
 
 	cmd.Flag("receive.replication-factor", "How many times to replicate incoming write requests.").Default("1").Uint64Var(&rc.replicationFactor)
 
+	cmd.Flag("receive.tenant-limits.max-head-series", "The total number of active or HEAD series that a tenant is allowed to have within a Receive topology.").Uint64Var(&rc.maxPerTenantLimit)
```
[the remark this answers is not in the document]
douglascamata: Considering we will also move to per-tenant limits on #5527, I will hide the flags over there too. 

Me and @saswatamcode plan to work together to improve the configuration of limits by moving them to a yaml file.

pkg/receive/handler.go
saswatamcode: Have implemented this in: https://github.com/thanos-io/thanos/pull/5520/commits/a8449564cda0fd8e139a9e7443090dfcdb443fdb

cmd/thanos/receive.go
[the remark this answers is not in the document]
saswatamcode: Have linked to docs section in https://github.com/thanos-io/thanos/pull/5520/commits/a8449564cda0fd8e139a9e7443090dfcdb443fdb

docs/components/receive.md
```diff
@@ -77,6 +77,22 @@ The example content of `hashring.json`:
 
 With such configuration any receive listens for remote write on `<ip>10908/api/v1/receive` and will forward to correct one in hashring if needed for tenancy and replication.
 
+## Limiting
+
+Thanos Receive, in Router or RouterIngestor mode, supports limiting tenant active or HEAD series, to maintain stability of the system. It uses any Prometheus Query API compatible meta-monitoring solution to get the current number of active series, and compares that with a configured limit, before ingesting any tenant's remote write request. In case a tenant has gone above the limit, their remote write requests are failed fully.
+
+Meta-monitoring in this context refers to an external monitoring system scraping all Thanos Receive instances and exposing them in an API compatible with the Prometheus Query API.
```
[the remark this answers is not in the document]
saswatamcode: Yes, have mentioned this in: https://github.com/thanos-io/thanos/pull/5520/commits/a8449564cda0fd8e139a9e7443090dfcdb443fdb

pkg/receive/handler.go
```diff
@@ -165,6 +178,24 @@ func NewHandler(logger log.Logger, o *Options) *Handler {
 				Help: "The number of times to replicate incoming write requests.",
 			},
 		),
+		configuredTenantLimit: promauto.With(registerer).NewGauge(
+			prometheus.GaugeOpts{
+				Name: "thanos_receive_tenant_head_series_limit",
+				Help: "The configured limit for active or HEAD series of tenants.",
```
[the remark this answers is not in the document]
saswatamcode: Yes TODO was already there in https://github.com/thanos-io/thanos/pull/5520/files#diff-42c21b7b04cc61ab0cda17794cc1efff14802e0e89a85503d28601e721c1dd31R598 🙂

cmd/thanos/receive.go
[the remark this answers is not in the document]
saswatamcode: Yes I have hidden the flags too! Also, needed to raise https://github.com/efficientgo/tools/pull/14 for this! 🙂

pkg/receive/handler.go
```diff
@@ -397,6 +449,20 @@ func (h *Handler) receiveHTTP(w http.ResponseWriter, r *http.Request) {
 
 	tLogger := log.With(h.logger, "tenant", tenant)
 
+	// Impose limits only if Receive is in Router or RouterIngestor mode.
+	if h.receiverMode == RouterOnly || h.receiverMode == RouterIngestor {
+		under, err := h.isUnderLimit(tenant, tLogger)
+		if err != nil {
+			level.Error(tLogger).Log("msg", "error while limiting", "err", err.Error())
```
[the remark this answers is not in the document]
saswatamcode: Here, under is already set to true as per this [line](https://github.com/thanos-io/thanos/pull/5520/files#diff-42c21b7b04cc61ab0cda17794cc1efff14802e0e89a85503d28601e721c1dd31R613). Having an explicit error here, would mean that we are no longer limiting in best-effort way. WDYT? 🙂

pkg/receive/handler.go
saswatamcode: Changed this in https://github.com/thanos-io/thanos/pull/5520/commits/a8449564cda0fd8e139a9e7443090dfcdb443fdb

pkg/receive/handler.go
saswatamcode: Have moved all the limiting logic to a separate struct in https://github.com/thanos-io/thanos/pull/5520/commits/a8449564cda0fd8e139a9e7443090dfcdb443fdb 🙂

test/e2e/receive_test.go
saswatamcode: So I've removed all except the multitenant hashring limiting test! 🙂

pkg/receive/handler.go
saswatamcode: So this error is not explicit (due to best-effort limiting), we only log this error. There doesn't seem to be a neat way of detecting that this is the first request of a tenant via Handler. Also, in case a tenant sends multiple request before we have had a chance to query meta-monitoring, all those requests would log errors, even if we detect first one.

WDYT would be a good way of handling this? 🙂

pkg/receive/handler.go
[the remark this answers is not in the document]
saswatamcode: So I think this is how this limit and its recovery will behave under different scenarios (without manual client config changes),

- New tenant on Receive sending requests from clients and remaining under the total limit: Will not have any issues with sending remote write requests
- New tenant on Receive that started going above the limit right from the first remote write request: The first request gets ingested and `prometheus_tsdb_head_series` gauge is incremented, which causes any subsequent remote write requests to be limited. The Receive `prometheus_tsdb_head_series` metric value (can) decrease when Receive tenant TSDB compacts the head block into a new persistent block, which happens every 2 hours. The decrease can also be measured via `prometheus_tsdb_head_series_removed_total`. However, if the clients are only pushing samples of the same series, then the metric may not decrease at all. If the clients have `retry_on_http_429` enabled, they will keep retrying their remote write requests during those 2 hours otherwise they will fail all their remote write requests. How many times they will retry (if enabled) each request is dependent on the `min/max_backoff` options which by default are 30ms to 5s. As this tenant has started going above the limit right from the beginning, it would have to wait for a full 2 hours before getting its limit "quota" back. On the client side you can see how this would limiting would fail or retry remote write requests with metrics like `prometheus_remote_storage_retried_samples_total` or `prometheus_remote_storage_failed_samples_total`.
- Old tenant on Receive that has started going above the limit: If a tenant which was under the set limit for some time, started going above their limit (maybe due to a new client being added/an existing client pushing more data), their remote write requests would start to be limited. The client retry semantics & frequency would be the same as mentioned in the point above. How Receive `prometheus_tsdb_head_series` increases and decreases would also be the same. However, the duration after which this tenant can get its quota back is variable. If the Receive tenant TSDB head block was about to reach block length (`chunkRange * 3/2`), then the tenant can get its quota back in maybe a minute. However, if the head block was just compacted, it might take 2 hours.

I hope this makes it a bit clearer. Feel free to point out if I'm wrong somewhere, or if something you thought could be done differently here. 🙂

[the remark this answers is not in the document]
matej-g: Thanks for going above and beyond to test this, I appreciate it @saswatamcode :bow:. I think it roughly confirms how I imagined this would work. Do you think it is worth it to document this and to explain behavior this behavior of clients? Then users can also make informed decision whether they want to enable the retry on `429` option on Prometheus side or not.

pkg/receive/handler.go
```diff
@@ -109,12 +132,13 @@ type Handler struct {
...
-	expBackoff   backoff.Backoff
-	peerStates   map[string]*retryState
-	receiverMode ReceiverMode
+	mtx               sync.RWMutex
+	hashring          Hashring
+	peers             *peerGroup
+	expBackoff        backoff.Backoff
+	peerStates        map[string]*retryState
+	receiverMode      ReceiverMode
+	ActiveSeriesLimit *activeSeriesLimit
```
matej-g, line 141: I guess we need only this field exported only because of the meta-monitoring query in `cmd` package, right? How about implementing that method on the `Handler` directly to make it fit with the rest of handler code organization?

pkg/receive/handler.go
```diff
@@ -482,6 +564,64 @@ func (h *Handler) receiveHTTP(w http.ResponseWriter, r *http.Request) {
 	h.writeSamplesTotal.WithLabelValues(strconv.Itoa(responseStatusCode), tenant).Observe(float64(totalSamples))
 }
 
+// QueryMetaMonitoring queries any Prometheus Query API compatible meta-monitoring
+// solution with the configured query for getting current active (head) series of all tenants.
+// It then populates tenantCurrentSeries map with result.
+func (a *activeSeriesLimit) QueryMetaMonitoring(ctx context.Context, logger log.Logger) error {
+	a.mtx.Lock()
+	defer a.mtx.Unlock()
```
matej-g, line 572: Perhaps we don't need this whole section locked and we could just lock / unlock when writing to map? (same with reading with the other method)

docs/components/receive.md
```diff
@@ -77,6 +77,23 @@ The example content of `hashring.json`:
 
 With such configuration any receive listens for remote write on `<ip>10908/api/v1/receive` and will forward to correct one in hashring if needed for tenancy and replication.
 
+## Limiting
```
matej-g, line 80: Shall we mention this is an experimental feature? I assume since we're hiding the flags for now.

pkg/receive/handler.go
matej-g: Hm that is true that we could receive multiple requests before the 15 second interval from a new tenant, feels like that's simply shortcoming of the meta-monitoring approach. Either way, are there other cases where a tenant would be missing from the map, apart from the case of new tenant? In that case I'd not consider it an error but simply return `true` without error and let the tenant have it "for free" before we query meta-monitoring.

 One thing that comes to my mind would be using a channel to signal to the querying goroutine to fire off, once we detect a request from new tenant has been ingested? Might be easier said than done.

matej-g (commented):
Nice progress :raised_hands: I put in one more round of comments, but it feels the PR is in good shape overall.

pkg/receive/handler.go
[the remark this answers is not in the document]
saswatamcode: Will add a section to docs!

pkg/receive/handler.go
saswatamcode: Will try this out. But another reason why a tenant may not be in the map is that meta-monitoring does not have metrics for one instance of a Receive, thereby completely missing a few tenants. In which case such an error log could be a clue. WDYT?

pkg/receive/handler.go
saswatamcode: I did this as I was extracting this logic to another struct! But yes, I could add the `QueryMetamonitoring` method back to `Handler`! 🙂

docs/components/receive.md
saswatamcode: Sure!

pkg/receive/handler.go
```diff
@@ -196,6 +243,27 @@ func NewHandler(logger log.Logger, o *Options) *Handler {
 		h.replicationFactor.Set(1)
 	}
 
+	h.ActiveSeriesLimit.configuredTenantLimit.Set(float64(o.MaxPerTenantLimit))
+	h.ActiveSeriesLimit.tenantCurrentSeriesMap = map[string]float64{}
```
fpetkovski, line 247: Do we need a constructor function for `ActiveSeriesLimit`?

Also, have you considered having an interface and a `NopSeriesLimiter` that is injected in the handler when no limits are applied? This should avoid a lot of the partial state and explicit checks throughout the codebase.

cmd/thanos/receive.go
```diff
@@ -293,6 +298,23 @@ func runReceive(
 		)
 	}
 
+	if (receiveMode == receive.RouterOnly || receiveMode == receive.RouterIngestor) && conf.maxPerTenantLimit != 0 {
```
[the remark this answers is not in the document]
saswatamcode: Yup, this is better! 🙂

pkg/receive/handler.go
saswatamcode: That's a good idea! Let me update the PR with something like this!

pkg/receive/handler.go
matej-g: Hm right I have not thought about hard tenancy :thinking:. Could we detect within meta-monitoring if we have metrics for expected number of receiver instances and if not warn about it?

pkg/receive/handler.go
```diff
@@ -482,6 +564,64 @@ func (h *Handler) receiveHTTP(w http.ResponseWriter, r *http.Request) {
...
+	// series. As such metric is updated in intervals, it is possible
+	// that Receive ingests more series than the limit, before detecting that
+	// a tenant has exceeded the set limits.
+	v, ok := a.tenantCurrentSeriesMap[tenant]
+	if !ok {
+		return true, errors.New("tenant not in current series map")
+	}
+
+	if v >= float64(a.limit) {
+		level.Error(logger).Log("msg", "tenant above limit", "currentSeries", v, "limit", a.limit)
```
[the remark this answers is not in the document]
saswatamcode: So the reason why I was logging here, was that we can log the data from the active series map in this function, as it would still be under read lock. This might help debug issues, where the cache map of a particular Receive node is out of sync for some reason, which can cause it to impose wrong limits. WDYT? 🙂

pkg/receive/handler.go
```diff
@@ -397,6 +417,20 @@ func (h *Handler) receiveHTTP(w http.ResponseWriter, r *http.Request) {
 
 	tLogger := log.With(h.logger, "tenant", tenant)
 
+	// Impose limits only if Receive is in Router or RouterIngestor mode.
+	if h.receiverMode == RouterOnly || h.receiverMode == RouterIngestor {
```
fpetkovski, line 421: Should we inject the series limiter as a dependency into the handler? This way in the handler we always call `isUnderLimit` and the `nopHandler` will always return false. 

Ideally I would like to have this `if` decision made in one place, and we already do it here: https://github.com/thanos-io/thanos/pull/5520/files#diff-3d45f9beb5b670e94fe2715e21477a3fb8f5db7bf97f01cddf6e108bb98761a6R301-R302

cmd/thanos/receive.go
```diff
@@ -821,6 +849,19 @@ func (rc *receiveConfig) registerFlag(cmd extkingpin.FlagClause) {
 
 	cmd.Flag("receive.replication-factor", "How many times to replicate incoming write requests.").Default("1").Uint64Var(&rc.replicationFactor)
 
+	cmd.Flag("receive.tenant-limits.max-head-series", "The total number of active (head) series that a tenant is allowed to have within a Receive topology. For more details refer: https://thanos.io/tip/components/receive.md/#limiting").Hidden().Uint64Var(&rc.maxPerTenantLimit)
+
+	cmd.Flag("receive.tenant-limits.meta-monitoring-url", "Meta-monitoring URL which is compatible with Prometheus Query API for active series limiting.").Hidden().URLVar(&rc.metaMonitoringUrl)
+
+	cmd.Flag("receive.tenant-limits.meta-monitoring-query", "PromQL Query to execute against meta-monitoring, to get the current number of active series for each tenant, across Receive replicas.").Default("sum(prometheus_tsdb_head_series) by (tenant)").Hidden().StringVar(&rc.metaMonitoringLimitQuery)
+
+	rc.metaMonitoringHttpClient = extflag.RegisterPathOrContent(
```
fpetkovski, line 858: Breaking down long lines is definitely a good idea, but to b e consistent with the rest of the flags, maybe we should just keep everything in the same line?

pkg/receive/handler.go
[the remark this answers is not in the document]
fpetkovski: I see. Well we can return the number of series as part of the result, but for simplicity we can also keep things as they are 👍

fpetkovski (commented):
This has shaped up really well! Just a few minor comments from my side and we should be good to go.

pkg/receive/handler.go
saswatamcode: Have implemented this here: https://github.com/thanos-io/thanos/pull/5520/commits/d101a6160acfb6059703d605ff4b472f25059a45

fpetkovski (dismissed):
Great stuff 💪

go.mod
```diff
@@ -246,6 +246,9 @@ replace (
 	// Required by Cortex https://github.com/cortexproject/cortex/pull/3051.
 	github.com/bradfitz/gomemcache => github.com/themihai/gomemcache v0.0.0-20180902122335-24332e2d58ab
 
+	// TODO(saswatamcode): Remove when https://github.com/efficientgo/tools/pull/14 is merged.
```
bwplotka, line 249: Merged (:

bwplotka (approved):
Thanks! LGTM.

One nit, but we can do that in separate PR

go.mod
saswatamcode: Removed in https://github.com/thanos-io/thanos/pull/5520/commits/4780bf73adc7548693af76af1a058cfd00f6986e 🙂

PR #5525, opened 2022-07-20
Update all bingo dependencies

fpetkovski:
This commit updates all bingo dependencies to their latest versions.

It pins golang.org/x/sys to v0.0.0-20220715151400-c0bba94af5f8 for
the github.com/google/go-jsonnet dependency in order to prevent
failures when running make docs on Mac OS.

## Changes

Updates all bingo dependencies to their latest versions.

## Verification

Relying on CI.


kakkoyun (approved):

kakkoyun:
Even though it looks good to me CI doesn't agree with me :sweat_smile:

fpetkovski:
I think it's a legit failure, let me look into it

.bingo/Variables.mk
```diff
@@ -41,93 +41,89 @@ $(GO_BINDATA): $(BINGO_DIR)/go-bindata.mod
...
 
 PROMDOC := $(GOBIN)/promdoc-v0.8.0
 $(PROMDOC): $(BINGO_DIR)/promdoc.mod
 	@# Install binary/ries using Go 1.14+ build command. This is using bwplotka/bingo-controlled, separate go module with pinned dependencies.
 	@echo "(re)installing $(GOBIN)/promdoc-v0.8.0"
 	@cd $(BINGO_DIR) && $(GO) build -mod=mod -modfile=promdoc.mod -o=$(GOBIN)/promdoc-v0.8.0 "github.com/plexsystems/promdoc"
 
-PROMETHEUS_ARRAY := $(GOBIN)/prometheus-v2.4.3+incompatible $(GOBIN)/prometheus-v1.8.2-0.20200724121523-657ba532e42f $(GOBIN)/prometheus-v1.8.2-0.20210331101223-3cafc58827d1
-$(PROMETHEUS_ARRAY): $(BINGO_DIR)/prometheus.mod $(BINGO_DIR)/prometheus.1.mod $(BINGO_DIR)/prometheus.2.mod
+PROMETHEUS := $(GOBIN)/prometheus-v0.37.0
```
fpetkovski, line 116: @bwplotka any reason why we needed multiple versions of Prometheus? Can we update to using the latest one only?

kakkoyun: AFAIK these are for testing against different versions of Prometheus.

fpetkovski:
Seems like Prometheus made some changes in error reporting and service discovery, so I had to adjust some tests. CI should be passing now.

kakkoyun (approved):
LGTM.

Just a question: Did we decide not to test against multiple versions? Or was it already obsolete?

fpetkovski:
The way I understand it, bingo used to install multiple versions of Prometheus, but only one version was used in tests. This is why I could just change the default Prometheus version here and the tests seemed to pass: https://github.com/thanos-io/thanos/pull/5525/files#diff-3e24274a6f32aa2c4ff63cf6b23579aa2e5d2929fcd8397b161b1f08fd1322d8R41

Maybe there was a test for a previous version which was removed? Unfortunately don't have enough context on this topic.

kakkoyun:
I'm happy to move forward. We can bring back stuff if it's needed.

@bwplotka, maybe you have the context?

PR #5527, opened 2022-07-20
Receive: add per request limits for remote write

douglascamata:

## Changes

Implement write requests size limits in Thanos Receive. 

These limits are applied per request and can be configured with the following
command line arguments:

- `--receive.write-request-limits.max-size-bytes`: the maximum body size.
- `--receive.write-request-limits.max-series`: the maximum amount of series in a
  single remote write request.
- `--receive.write-request-limits.max-samples`: the maximum amount of samples in a
  single remote write request (summed from all series).

Any request above these limits will cause an 413 HTTP response (_Entity Too Large_)
and should not be retried without modifications. It's up to remote write clients to
split up the data and retry or completely drop it.

Another limit that was added to control concurrency on the remote write requests:

- `--receive.write-request-limits.max-concurrency`: the maximum amount remote write
  requests that will be concurrently processed while others wait.

**By default all these limits are disabled.** 

The flags are also hidden until per-tenant configuration in a file is implemented. A mention that this feature is WIP and might change whenever the per-tenant configuration file lands is present in the Receive docs.

This is a part of the work planned in #5404.

## Verification

- [x] Added some cases to the Receive's handler tests and they pass.
- [x] Running it locally.

## Future work

- Add support for per-tenant limit configuration

### DIFF
--- a/cmd/thanos/receive.go
+++ b/cmd/thanos/receive.go
@@ -863,6 +863,8 @@ func (rc *receiveConfig) registerFlag(cmd extkingpin.FlagClause) {
 
 	rc.reqLogConfig = extkingpin.RegisterRequestLoggingFlags(cmd)
 
+	// TODO(douglascamata): allow all these limits to be configured per tenant
+	// and move the configuration to a file.
 	cmd.Flag("receive.write-request-limits.max-series",
 		"The maximum amount of series accepted in remote write requests.").
 		Default("0").Int64Var(&rc.writeSeriesLimit)
--- a/cmd/thanos/receive.go
+++ b/cmd/thanos/receive.go
@@ -864,7 +864,8 @@ func (rc *receiveConfig) registerFlag(cmd extkingpin.FlagClause) {
 	rc.reqLogConfig = extkingpin.RegisterRequestLoggingFlags(cmd)
 
 	// TODO(douglascamata): allow all these limits to be configured per tenant
-	// and move the configuration to a file.
+	// and move the configuration to a file. Then this is done, remove the
+	// "hidden" modifier on all these flags.
 	cmd.Flag("receive.write-request-limits.max-series",
 		"The maximum amount of series accepted in remote write requests."+
 			"The default is no limit, represented by 0.").
--- a/cmd/thanos/receive.go
+++ b/cmd/thanos/receive.go
@@ -198,25 +198,26 @@ func runReceive(
 	)
 	writer := receive.NewWriter(log.With(logger, "component", "receive-writer"), dbs)
 	webHandler := receive.NewHandler(log.With(logger, "component", "receive-handler"), &receive.Options{
-		Writer:                writer,
-		ListenAddress:         conf.rwAddress,
-		Registry:              reg,
-		Endpoint:              conf.endpoint,
-		TenantHeader:          conf.tenantHeader,
-		TenantField:           conf.tenantField,
-		DefaultTenantID:       conf.defaultTenantID,
-		ReplicaHeader:         conf.replicaHeader,
-		ReplicationFactor:     conf.replicationFactor,
-		RelabelConfigs:        relabelConfig,
-		ReceiverMode:          receiveMode,
-		Tracer:                tracer,
-		TLSConfig:             rwTLSConfig,
-		DialOpts:              dialOpts,
-		ForwardTimeout:        time.Duration(*conf.forwardTimeout),
-		TSDBStats:             dbs,
-		WriteSeriesLimit:      conf.writeSeriesLimit,
-		WriteSamplesLimit:     conf.writeSamplesLimit,
-		WriteRequestSizeLimit: conf.writeRequestSizeLimit,
+		Writer:                       writer,
+		ListenAddress:                conf.rwAddress,
+		Registry:                     reg,
+		Endpoint:                     conf.endpoint,
+		TenantHeader:                 conf.tenantHeader,
+		TenantField:                  conf.tenantField,
+		DefaultTenantID:              conf.defaultTenantID,
+		ReplicaHeader:                conf.replicaHeader,
+		ReplicationFactor:            conf.replicationFactor,
+		RelabelConfigs:               relabelConfig,
+		ReceiverMode:                 receiveMode,
+		Tracer:                       tracer,
+		TLSConfig:                    rwTLSConfig,
+		DialOpts:                     dialOpts,
+		ForwardTimeout:               time.Duration(*conf.forwardTimeout),
+		TSDBStats:                    dbs,
+		WriteSeriesLimit:             conf.writeSeriesLimit,
+		WriteSamplesLimit:            conf.writeSamplesLimit,
+		WriteRequestSizeLimit:        conf.writeRequestSizeLimit,
+		WriteRequestConcurrencyLimit: conf.writeRequestConcurrencyLimit,
 	})
 
 	grpcProbe := prober.NewGRPC()
@@ -767,9 +768,10 @@ type receiveConfig struct {
 	reqLogConfig      *extflag.PathOrContent
 	relabelConfigPath *extflag.PathOrContent
 
-	writeSeriesLimit      int
-	writeSamplesLimit     int
-	writeRequestSizeLimit int
+	writeSeriesLimit             int
+	writeSamplesLimit            int
+	writeRequestSizeLimit        int
+	writeRequestConcurrencyLimit int
 }
 
 func (rc *receiveConfig) registerFlag(cmd extkingpin.FlagClause) {
@@ -872,6 +874,10 @@ func (rc *receiveConfig) registerFlag(cmd extkingpin.FlagClause) {
 	cmd.Flag("receive.request-limits.max-size-bytes",
 		"The maximum size (in bytes) of remote write requests.").
 		Default("0").IntVar(&rc.writeRequestSizeLimit)
+
+	cmd.Flag("receive.request-limits.max-concurrency",
+		"The maximum size (in bytes) of remote write requests.").
+		Default("0").IntVar(&rc.writeRequestConcurrencyLimit)
 }
 
 // determineMode returns the ReceiverMode that this receiver is configured to run in.
--- a/cmd/thanos/receive.go
+++ b/cmd/thanos/receive.go
@@ -198,22 +198,25 @@ func runReceive(
 	)
 	writer := receive.NewWriter(log.With(logger, "component", "receive-writer"), dbs)
 	webHandler := receive.NewHandler(log.With(logger, "component", "receive-handler"), &receive.Options{
-		Writer:            writer,
-		ListenAddress:     conf.rwAddress,
-		Registry:          reg,
-		Endpoint:          conf.endpoint,
-		TenantHeader:      conf.tenantHeader,
-		TenantField:       conf.tenantField,
-		DefaultTenantID:   conf.defaultTenantID,
-		ReplicaHeader:     conf.replicaHeader,
-		ReplicationFactor: conf.replicationFactor,
-		RelabelConfigs:    relabelConfig,
-		ReceiverMode:      receiveMode,
-		Tracer:            tracer,
-		TLSConfig:         rwTLSConfig,
-		DialOpts:          dialOpts,
-		ForwardTimeout:    time.Duration(*conf.forwardTimeout),
-		TSDBStats:         dbs,
+		Writer:                writer,
+		ListenAddress:         conf.rwAddress,
+		Registry:              reg,
+		Endpoint:              conf.endpoint,
+		TenantHeader:          conf.tenantHeader,
+		TenantField:           conf.tenantField,
+		DefaultTenantID:       conf.defaultTenantID,
+		ReplicaHeader:         conf.replicaHeader,
+		ReplicationFactor:     conf.replicationFactor,
+		RelabelConfigs:        relabelConfig,
+		ReceiverMode:          receiveMode,
+		Tracer:                tracer,
+		TLSConfig:             rwTLSConfig,
+		DialOpts:              dialOpts,
+		ForwardTimeout:        time.Duration(*conf.forwardTimeout),
+		TSDBStats:             dbs,
+		WriteSeriesLimit:      conf.writeSeriesLimit,
+		WriteSamplesLimit:     conf.writeSamplesLimit,
+		WriteRequestSizeLimit: conf.writeRequestSizeLimit,
 	})
 
 	grpcProbe := prober.NewGRPC()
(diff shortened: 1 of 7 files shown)


cmd/thanos/receive.go
```diff
@@ -853,6 +860,18 @@ func (rc *receiveConfig) registerFlag(cmd extkingpin.FlagClause) {
...
+		"The maximum amount of series accepted in remote write requests.").
+		Default("0").IntVar(&rc.writeSeriesLimit)
+
+	cmd.Flag("receive.request-limits.max-samples",
+		"The maximum amount of samples accepted in remote write requests.").
+		Default("0").IntVar(&rc.writeSamplesLimit)
+
+	cmd.Flag("receive.request-limits.max-size-bytes",
+		"The maximum size (in bytes) of remote write requests.").
+		Default("0").IntVar(&rc.writeRequestSizeLimit)
```
yeya24, line 874: Do we want to support limits on the tenant level?

douglascamata: Yup, this is something that I plan to add as a follow up. I am also talking with @saswatamcode, because his active series work will also need different per-tenant limits. We will propose that all the limits should be configured using a yaml file to make it simpler.

pkg/receive/handler.go
```diff
@@ -397,10 +439,31 @@ func (h *Handler) receiveHTTP(w http.ResponseWriter, r *http.Request) {
...
+		defer h.writeGate.Done()
+	}
+
 	// ioutil.ReadAll dynamically adjust the byte slice for read data, starting from 512B.
 	// Since this is receive hot path, grow upfront saving allocations and CPU time.
 	compressed := bytes.Buffer{}
 	if r.ContentLength >= 0 {
+		// If the content length is known, we can block the request based on
+		// max size limit here already and avoid growing the buffer.
+		sizeLimit := int64(h.options.WriteRequestSizeLimit)
```
saswatamcode, line 461: I wonder if we can isolate limiting logic into a dedicated function (apart from `h.writeGate`) and use that in `receiveHTTP`. Maybe to a function like `requestUnderLimit` which takes in request as argument and returns parsed request and bool.

As these are a lot of `ifs` which would happen every time a remote write request is received, I think benchmarking this logic would be nice! Separate `requestUnderLimit` function would allow this! 🙂

pkg/receive/handler_test.go
```diff
@@ -690,6 +690,87 @@ func TestReceiveQuorum(t *testing.T) {
 	}
 }
 
+func TestReceiveWriteRequestLimits(t *testing.T) {
```
saswatamcode, line 693: Maybe an e2e test could help in understanding some real-world client behavior eg Prometheus, avalanche. WDYT?

saswatamcode (commented):
Thanks for amazing work! 🌟 

Some questions/suggestions!

pkg/receive/handler.go
douglascamata: Let me see what I can do to isolate the logic. 👍

I have no worries regarding the ifs, they should add such a tiny amount of time to the handler's execution that the benchmark would show mostly noise. Most of the time spent in limiting logic will be due to the metrics increment and writing the error response back. 

If others also worry about this I can try to work out on a benchmark and see whether it's useful.

pkg/receive/handler.go
```diff
@@ -449,6 +518,23 @@ func (h *Handler) receiveHTTP(w http.ResponseWriter, r *http.Request) {
...
+		http.Error(w, "too many timeseries", http.StatusRequestEntityTooLarge)
+		return
+	}
+
+	totalSamples := 0
+	for _, timeseries := range wreq.Timeseries {
+		totalSamples += len(timeseries.Samples)
+	}
+	samplesLimit := h.options.WriteSamplesLimit
+	if samplesLimit > 0 && totalSamples > samplesLimit {
```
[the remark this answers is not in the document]
douglascamata: I can make the condition inside the `if` much shorter and easier to read no matter your screen's resolution. 😄

pkg/receive/handler_test.go
douglascamata: Will add 👍

douglascamata: @saswatamcode actually, what do you mean with `understanding some real-world client behavior eg Prometheus, avalanche.`? 

Do you mean, for example, testing that a well configured Prometheus or Avalanche will retry on a 429? I don't believe we need an e2e test to cover this.

I started working on the e2e tests and the further I progress on them the more I think they have no real value beyond what the handler test provides. The only additional value the e2e adds that I can see is to ensure that the CLI args are properly passed down to the request limiting code. Besides this, the request limiting code can be tested without having to add an e2e test.

cmd/thanos/receive.go
yeya24: So I am curious if we are going the per-tenant limit way, do we still want to keep those flags in this PR? Is it working as the default value if no limit is explicitly configured in the config file?

pkg/receive/handler_test.go
saswatamcode: I meant if seeing how a hashring would behave, with such request limits and multiple tenants, would be useful?
But yes let's see what maintainers have to say! Unit tests might be sufficient here. 🙂

cmd/thanos/receive.go
douglascamata: Good question, @yeya24. I would prefer to have everything in the same place, the file. But defining the default value quickly through CLI args seem convenient too. I have no strong opinion on this. 

We can maybe support CLI and file for default limit and document that one of them has priority over the other.

docs/components/receive.md
```diff
@@ -77,6 +77,25 @@ The example content of `hashring.json`:
...
+## Limiting
+
+### Request limits (work in progress)
+
+Thanos Receive supports setting limits on the incoming remote write request sizes. These limits should help you to prevent a single tenant from being able to send big requests and possibly crash the Receive.
+
+These limits are applied per request and can be configured with the following command line arguments:
+
+- `--receive.write-request-limits.max-size-bytes`: the maximum body size.
+- `--receive.write-request-limits.max-concurrency`: the maximum amount remote write requests that will be concurrently processed while others wait.
```
fpetkovski, line 89:
```suggestion
- `--receive.write-request-limits.max-concurrency`: the maximum amount of remote write requests that will be concurrently processed while others wait.
```

pkg/receive/handler.go
```diff
@@ -147,6 +156,13 @@ func NewHandler(logger log.Logger, o *Options) *Handler {
 			Max:    30 * time.Second,
 			Jitter: true,
 		},
+		writeGate: gate.NewNoop(),
+		requestLimiter: newRequestLimiter(
```
fpetkovski, line 160: Instead of extending the existing handler, is it possible to make the limiter a middleware that wraps the handler? This could be cleaner to maintain long term.

pkg/receive/handler.go
```diff
@@ -397,10 +420,27 @@ func (h *Handler) receiveHTTP(w http.ResponseWriter, r *http.Request) {
 
 	tLogger := log.With(h.logger, "tenant", tenant)
 
+	if h.writeGate != nil {
```
fpetkovski, line 423: Do we need to check for `nil` since the default gate is a no-op gate?

pkg/receive/handler_test.go
```diff
@@ -690,6 +690,85 @@ func TestReceiveQuorum(t *testing.T) {
...
+			}
+
+			appendables := []*fakeAppendable{
+				{
+					appender: newFakeAppender(nil, nil, nil),
+				},
+			}
+			handlers, _ := newTestHandlerHashring(appendables, 1)
+			handler := handlers[0]
+			handler.requestLimiter = newRequestLimiter(1*1024*1024, 20, 200, nil)
```
fpetkovski, line 741: You can use `units.Megabytes` from `github.com/alecthomas/units` for readability.

pkg/receive/handler_test.go
```diff
@@ -690,6 +690,85 @@ func TestReceiveQuorum(t *testing.T) {
...
+		{
+			name:          "Request above body size limit",
+			status:        http.StatusRequestEntityTooLarge,
+			amountSeries:  300,
+			amountSamples: 150,
+		},
+	} {
+		t.Run(tc.name, func(t *testing.T) {
+			if tc.amountSamples == 0 {
+				tc.amountSamples = 1
```
fpetkovski, line 731: What's the reason for this mutation?

pkg/receive/handler.go
```diff
@@ -380,8 +403,8 @@ func (h *Handler) receiveHTTP(w http.ResponseWriter, r *http.Request) {
 	var err error
 	span, ctx := tracing.StartSpan(r.Context(), "receive_http")
 	defer span.Finish()
-
```
fpetkovski, line 383: This blank line was actually nice :)

pkg/receive/handler.go
```diff
@@ -380,8 +403,8 @@ func (h *Handler) receiveHTTP(w http.ResponseWriter, r *http.Request) {
...
 	tenant := r.Header.Get(h.options.TenantHeader)
+
```
fpetkovski, line 407:
```suggestion
```

fpetkovski (commented):
Nice, this looks quite solid.

pkg/receive/handler_test.go
douglascamata: This is to kind of have 1 as default value for the test scenarios that do not explicitly set `amountSamples`.

pkg/receive/handler.go
douglascamata: Unfortunately I don't think that's a good idea, because then the limiter middleware will have to do work that is done again in the main handler: 

- Checking content length header is easy, but when it's not there we have to read the request body to check its size, which involves trying a snappy decompression too.
- Unmarshalling the request body and looping over the data to count timeseries and samples.

I wanted to avoid doing extra work at all costs to keep the "hot path" of Receive fast.

douglascamata:
Thx for the review, @fpetkovski. Some comments were taken care off already and I replied to others. 🙇

pkg/receive/handler.go
fpetkovski: I see, that's a good point. Well initially I thought the middleware would be on [this](https://github.com/douglascamata/thanos/blob/7efee5e7d59527c7464a84b1b2d9436691ed74bf/pkg/receive/handler.go#L370) handler which already has the decoded protobuf request, but here we don't have the raw body correct?

Also if I may ask, what is the practical use of setting limits on the request body? As an admin, I am not sure I would know what to set this limit to. I can understand the impact of limiting series or samples, but I would find it hard to know how to configure the body size limit.

fpetkovski (approved):
Thanks, this lgtm!

cmd/thanos/receive.go
```diff
@@ -853,6 +862,29 @@ func (rc *receiveConfig) registerFlag(cmd extkingpin.FlagClause) {
 		Default("false").Hidden().BoolVar(&rc.allowOutOfOrderUpload)
 
 	rc.reqLogConfig = extkingpin.RegisterRequestLoggingFlags(cmd)
+
+	// TODO(douglascamata): allow all these limits to be configured per tenant
```
bwplotka, line 866:
```suggestion
	// TODO(douglascamata): Allow all these limits to be configured per tenant
```

bwplotka (approved):
Amazing, looks good generally, great work! Leaving for @matej-g  to check as well.

docs/components/receive.md
```diff
@@ -77,6 +77,25 @@ The example content of `hashring.json`:
...
+## Limiting
+
+### Request limits (work in progress)
+
+Thanos Receive supports setting limits on the incoming remote write request sizes. These limits should help you to prevent a single tenant from being able to send big requests and possibly crash the Receive.
+
+These limits are applied per request and can be configured with the following command line arguments:
+
+- `--receive.write-request-limits.max-size-bytes`: the maximum body size.
+- `--receive.write-request-limits.max-concurrency`: the maximum amount of remote write requests that will be concurrently processed while others wait.
```
matej-g, line 89: I'm assuming the max concurrency limit won't result in `413` response, maybe it would be worth to put this into a separate section, since it's not really a size limit but rather a gating mechanism. What do you think?

pkg/receive/limiter_test.go
```diff
@@ -0,0 +1,152 @@
+// Copyright (c) The Thanos Authors.
+// Licensed under the Apache License 2.0.
+
+package receive
+
+import (
+	"testing"
+)
```
matej-g, line 8: nit (curious that `fmt` does not pick this up :thinking: )
```suggestion
import "testing"
```

matej-g (commented):
Solid implementation, nice tests! :muscle: I'm have only few notes about docs, otherwise I'm happy with this :blush:

docs/components/receive.md
```diff
@@ -77,6 +77,25 @@ The example content of `hashring.json`:
...
+- `--receive.write-request-limits.max-series`: the maximum amount of series in a single remote write request.
+- `--receive.write-request-limits.max-samples`: the maximum amount of samples in a single remote write request (summed from all series).
+
+Any request above these limits will cause an 413 HTTP response (*Entity Too Large*) and should not be retried without modifications. It's up to remote write clients to split up the data and retry or completely drop it.
```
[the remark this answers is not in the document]
douglascamata: Agreed. I'll write more about this. 👍 

> I'm wondering however whether there'd be a will to do something like that in the upstream 🤔, to process 413 in some way other than just dropping it.

That's something I've considered. I think we would have to add error codes to the response, so that a client (Prometheus or something else) could know whether the 413 is because of too many series, too many samples, or simply request body size, and react accordingly so that however it splits the request it'll be accepted next time (or likely accepted).

docs/components/receive.md
douglascamata: Suuure, it makes sense. I'll move it.

matej-g (approved):
Awesome addition and great docs! :heart:

saswatamcode (approved):
Awesome! 💪🏻

douglascamata:
@saswatamcode I just pushed an update on [58d362b](https://github.com/thanos-io/thanos/pull/5527/commits/58d362b346a5877b8617d6a789f102e3d7569ff1) that runs the request limit tests with a 3 node hashring. This should prevent us from changing the batching logic and cause an incoming request that was accepted to hit limits when batched for fanout. 

Seems like the fanout requests are now sent via gRPC:

- From this part in the handler: https://github.com/thanos-io/thanos/blob/main/pkg/receive/handler.go#L656
- It sends this request: https://github.com/thanos-io/thanos/blob/fd275f8ac3d906c4ebb3c3f8be9db8701a71eabd/pkg/store/storepb/rpc.pb.go#L1122
- Which ends up calling `h.handleRequest`: https://github.com/thanos-io/thanos/blob/main/pkg/receive/handler.go#L768

At this point, the limits logic is completely skipped.

pkg/receive/handler_test.go
bwplotka: Less + simpler tests while having the same verifications level = better, so I think a unit tests will suffice here. 

> It means we're testing Prometheus/Avalanche and not Thanos.

In some way yes, but also... are we sure those projects have those tests? 🙃 

Anyway, let's investigate/create focus tests in separate stream of work

bwplotka (approved):
Thanks 💪🏽 LGTM

PR #5530, opened 2022-07-21
Fix proposal on website

saswatamcode:
Currently, the Automated Per Endpoint TLS proposal does not show on the website menu due to lack of frontmatter.
We noticed during discussion with @SrushtiSapkale.
 
## Changes

- Add frontmatter to per endpoint TLS proposal

## Verification

Tested locally


squat (approved):
Nice catch!

PR #5533, opened 2022-07-22
Receiver: Add benchmark for receive writer

matej-g:

## Changes

Off the back of the discussion on https://github.com/thanos-io/thanos/pull/5508#discussion_r923067652, I thought it would be good add some benchmarks for receive writer, to help us reason about changes in the writer (given this is the crucial piece of remote write path).

## Verification

Results:
```
goos: linux
goarch: amd64
pkg: github.com/thanos-io/thanos/pkg/receive
cpu: Intel(R) Core(TM) i7-9850H CPU @ 2.60GHz
BenchmarkWriterTimeSeriesWithSingleLabel_10-12            297813              3568 ns/op             411 B/op          9 allocs/op
BenchmarkWriterTimeSeriesWithSingleLabel_100-12            47799             21589 ns/op             412 B/op          9 allocs/op
BenchmarkWriterTimeSeriesWithSingleLabel_1000-12            5859            193874 ns/op             453 B/op          9 allocs/op
BenchmarkWriterTimeSeriesWith10Labels_10-12               219638              5028 ns/op             410 B/op          9 allocs/op
BenchmarkWriterTimeSeriesWith10Labels_100-12               34004             33783 ns/op             412 B/op          9 allocs/op
BenchmarkWriterTimeSeriesWith10Labels_1000-12               3147            344160 ns/op             458 B/op          9 allocs/op
PASS
ok      github.com/thanos-io/thanos/pkg/receive 31.432s
```

### DIFF
--- a/pkg/receive/writer_test.go
+++ b/pkg/receive/writer_test.go
@@ -5,6 +5,7 @@ package receive
 
 import (
 	"context"
+	"fmt"
 	"io/ioutil"
 	"os"
 	"strings"
@@ -173,4 +174,81 @@ func TestWriter(t *testing.T) {
 			}
 		})
 	}
+}
+
+func BenchmarkWriterTimeSeriesWithSingleLabel_10(b *testing.B)   { benchmarkWriter(b, 1, 10) }
+func BenchmarkWriterTimeSeriesWithSingleLabel_100(b *testing.B)  { benchmarkWriter(b, 1, 100) }
+func BenchmarkWriterTimeSeriesWithSingleLabel_1000(b *testing.B) { benchmarkWriter(b, 1, 1000) }
+
+func BenchmarkWriterTimeSeriesWith10Labels_10(b *testing.B)   { benchmarkWriter(b, 10, 10) }
+func BenchmarkWriterTimeSeriesWith10Labels_100(b *testing.B)  { benchmarkWriter(b, 10, 100) }
+func BenchmarkWriterTimeSeriesWith10Labels_1000(b *testing.B) { benchmarkWriter(b, 10, 1000) }
+
+func benchmarkWriter(b *testing.B, labelsNum int, seriesNum int) {
+	dir, err := ioutil.TempDir("", "test")
+	testutil.Ok(b, err)
+	defer func() { testutil.Ok(b, os.RemoveAll(dir)) }()
+
+	logger := log.NewNopLogger()
+
+	m := NewMultiTSDB(dir, logger, prometheus.NewRegistry(), &tsdb.Options{
+		MinBlockDuration:      (2 * time.Hour).Milliseconds(),
+		MaxBlockDuration:      (2 * time.Hour).Milliseconds(),
+		RetentionDuration:     (6 * time.Hour).Milliseconds(),
+		NoLockfile:            true,
+		MaxExemplars:          0,
+		EnableExemplarStorage: true,
+	},
+		labels.FromStrings("replica", "01"),
+		"tenant_id",
+		nil,
+		false,
+		metadata.NoneFunc,
+	)
+	defer func() { testutil.Ok(b, m.Close()) }()
+
+	testutil.Ok(b, m.Flush())
+	testutil.Ok(b, m.Open())
+
+	app, err := m.TenantAppendable("foo")
+	testutil.Ok(b, err)
+
+	w := NewWriter(logger, m)
+
+	ctx, cancel := context.WithTimeout(context.Background(), 10*time.Second)
+	defer cancel()
+
+	testutil.Ok(b, runutil.Retry(1*time.Second, ctx.Done(), func() error {
+		_, err = app.Appender(context.Background())
+		return err
+	}))
+
+	timeSeries := generateLabelsAndSeries(labelsNum, seriesNum)
+
+	wreq := &prompb.WriteRequest{
+		Timeseries: timeSeries,
+	}
+
+	b.ReportAllocs()
+	b.ResetTimer()
+
+	for i := 0; i < b.N; i++ {
+		_ = w.Write(ctx, "foo", wreq)
+	}
+}
+
+func generateLabelsAndSeries(labelsNum int, seriesNum int) []prompb.TimeSeries {
+	// Generate some labels first.
+	l := make([]labelpb.ZLabel, 0, labelsNum)
+	l = append(l, labelpb.ZLabel{Name: "__name__", Value: "test"})
+	for i := 0; i < labelsNum; i++ {
+		l = append(l, labelpb.ZLabel{Name: fmt.Sprintf("label_%q", rune('a'-1+i)), Value: fmt.Sprintf("%d", i)})
+	}
+
+	ts := make([]prompb.TimeSeries, 0, seriesNum)
+	for j := 0; j < seriesNum; j++ {
+		ts = append(ts, prompb.TimeSeries{Labels: l, Samples: []prompb.Sample{{Value: 1, Timestamp: 10}}})
+	}
+
+	return ts
 }
--- a/pkg/receive/writer_test.go
+++ b/pkg/receive/writer_test.go
@@ -185,7 +185,7 @@ func BenchmarkWriterTimeSeriesWith10Labels_100(b *testing.B)  { benchmarkWriter(
 func BenchmarkWriterTimeSeriesWith10Labels_1000(b *testing.B) { benchmarkWriter(b, 10, 1000) }
 
 func benchmarkWriter(b *testing.B, labelsNum int, seriesNum int) {
-	dir, err := ioutil.TempDir("", "test")
+	dir, err := ioutil.TempDir("", "bench-writer")
 	testutil.Ok(b, err)
 	defer func() { testutil.Ok(b, os.RemoveAll(dir)) }()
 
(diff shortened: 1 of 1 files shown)


pkg/receive/writer_test.go
```diff
@@ -174,3 +175,80 @@ func TestWriter(t *testing.T) {
...
+
+	ctx, cancel := context.WithTimeout(context.Background(), 10*time.Second)
+	defer cancel()
+
+	testutil.Ok(b, runutil.Retry(1*time.Second, ctx.Done(), func() error {
+		_, err = app.Appender(context.Background())
+		return err
+	}))
+
+	timeSeries := generateLabelsAndSeries(labelsNum, seriesNum)
```
fpetkovski, line 226:
```suggestion
	timeSeries := generateLabelsAndSeries(numLabels, numSeries)
```

pkg/receive/writer_test.go
```diff
@@ -174,3 +175,80 @@ func TestWriter(t *testing.T) {
...
+func BenchmarkWriterTimeSeriesWithSingleLabel_10(b *testing.B)   { benchmarkWriter(b, 1, 10) }
+func BenchmarkWriterTimeSeriesWithSingleLabel_100(b *testing.B)  { benchmarkWriter(b, 1, 100) }
+func BenchmarkWriterTimeSeriesWithSingleLabel_1000(b *testing.B) { benchmarkWriter(b, 1, 1000) }
+
+func BenchmarkWriterTimeSeriesWith10Labels_10(b *testing.B)   { benchmarkWriter(b, 10, 10) }
+func BenchmarkWriterTimeSeriesWith10Labels_100(b *testing.B)  { benchmarkWriter(b, 10, 100) }
+func BenchmarkWriterTimeSeriesWith10Labels_1000(b *testing.B) { benchmarkWriter(b, 10, 1000) }
+
+func benchmarkWriter(b *testing.B, labelsNum int, seriesNum int) {
+	dir, err := ioutil.TempDir("", "test")
```
fpetkovski, line 188: To allow concurrent execution, maybe we should give the temp dir a unique name?
```suggestion
	dir, err := ioutil.TempDir("", "bench-writer")
```

fpetkovski (commented):
Amazing stuff, just some small suggestions from my side

fpetkovski (approved):

douglascamata (changes requested):

pkg/receive/writer_test.go
```diff
@@ -174,3 +175,80 @@ func TestWriter(t *testing.T) {
...
+func BenchmarkWriterTimeSeriesWithSingleLabel_1000(b *testing.B) { benchmarkWriter(b, 1, 1000) }
+
+func BenchmarkWriterTimeSeriesWith10Labels_10(b *testing.B)   { benchmarkWriter(b, 10, 10) }
+func BenchmarkWriterTimeSeriesWith10Labels_100(b *testing.B)  { benchmarkWriter(b, 10, 100) }
+func BenchmarkWriterTimeSeriesWith10Labels_1000(b *testing.B) { benchmarkWriter(b, 10, 1000) }
+
+func benchmarkWriter(b *testing.B, labelsNum int, seriesNum int) {
+	dir, err := ioutil.TempDir("", "bench-writer")
+	testutil.Ok(b, err)
+	defer func() { testutil.Ok(b, os.RemoveAll(dir)) }()
```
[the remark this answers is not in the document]
matej-g: 100% honesty, I copy-pasted this from the other test and haven't thought about it much. I guess it should not be failure in and of itself, since tests pass even if we fail to cleanup :shrug:. I guess in this case it's more of a check that test ran OK rather that validating test results.

douglascamata (approved):
LGTM :shipit:

pkg/receive/writer_test.go
```diff
@@ -174,3 +175,80 @@ func TestWriter(t *testing.T) {
...
+func BenchmarkWriterTimeSeriesWithSingleLabel_10(b *testing.B)   { benchmarkWriter(b, 1, 10) }
+func BenchmarkWriterTimeSeriesWithSingleLabel_100(b *testing.B)  { benchmarkWriter(b, 1, 100) }
+func BenchmarkWriterTimeSeriesWithSingleLabel_1000(b *testing.B) { benchmarkWriter(b, 1, 1000) }
+
+func BenchmarkWriterTimeSeriesWith10Labels_10(b *testing.B)   { benchmarkWriter(b, 10, 10) }
+func BenchmarkWriterTimeSeriesWith10Labels_100(b *testing.B)  { benchmarkWriter(b, 10, 100) }
+func BenchmarkWriterTimeSeriesWith10Labels_1000(b *testing.B) { benchmarkWriter(b, 10, 1000) }
+
+func benchmarkWriter(b *testing.B, labelsNum int, seriesNum int) {
+	dir, err := ioutil.TempDir("", "bench-writer")
```
bwplotka, line 188: There is `b.TempDir()` which creates dir that will be automatically removed we could use (:

pkg/receive/writer_test.go
```diff
@@ -174,3 +175,80 @@ func TestWriter(t *testing.T) {
...
+		MaxExemplars:          0,
+		EnableExemplarStorage: true,
+	},
+		labels.FromStrings("replica", "01"),
+		"tenant_id",
+		nil,
+		false,
+		metadata.NoneFunc,
+	)
+	defer func() { testutil.Ok(b, m.Close()) }()
```
bwplotka, line 208:
```suggestion
	b.Cleanup(func() { testutil.Ok(b, m.Close()) })
```

pkg/receive/writer_test.go
```diff
@@ -174,3 +175,80 @@ func TestWriter(t *testing.T) {
...
+		_ = w.Write(ctx, "foo", wreq)
+	}
+}
+
+func generateLabelsAndSeries(numLabels int, numSeries int) []prompb.TimeSeries {
+	// Generate some labels first.
+	l := make([]labelpb.ZLabel, 0, numLabels)
+	l = append(l, labelpb.ZLabel{Name: "__name__", Value: "test"})
+	for i := 0; i < numLabels; i++ {
+		l = append(l, labelpb.ZLabel{Name: fmt.Sprintf("label_%q", rune('a'-1+i)), Value: fmt.Sprintf("%d", i)})
```
bwplotka, line 245: %q feels like it's adding quotes, no? This might mean those labels are not valid label names.

GiedriusS (approved):

pkg/receive/writer_test.go
```diff
@@ -174,3 +175,80 @@ func TestWriter(t *testing.T) {
...
+	}
+
+	ts := make([]prompb.TimeSeries, 0, numSeries)
+	for j := 0; j < numSeries; j++ {
+		ts = append(ts, prompb.TimeSeries{Labels: l, Samples: []prompb.Sample{{Value: 1, Timestamp: 10}}})
```
bwplotka, line 250: isn't the same timestamp a problem? I thought we should be getting conflict in append.

pkg/receive/writer_test.go
```diff
@@ -174,3 +175,80 @@ func TestWriter(t *testing.T) {
...
+
+	wreq := &prompb.WriteRequest{
+		Timeseries: timeSeries,
+	}
+
+	b.ReportAllocs()
+	b.ResetTimer()
+
+	for i := 0; i < b.N; i++ {
+		_ = w.Write(ctx, "foo", wreq)
```
bwplotka, line 236: So I wrote this in my book just recently (should be out soon!), but it's worth to check error here always.

Yes - it adds some overhead to the result, but it's almost nothing (one if - if we add `testutil.Ok(b, w.Write...)`) but it ensures we write correct thing.

In this particular case I *think* you measure wrong thing - we should get conflicts error from second iteration 🙈 Even if I am wrong, it's worth to ensure error of test always for future users of this tests.

pkg/receive/writer_test.go
```diff
@@ -174,3 +175,80 @@ func TestWriter(t *testing.T) {
 		})
 	}
 }
+
+func BenchmarkWriterTimeSeriesWithSingleLabel_10(b *testing.B)   { benchmarkWriter(b, 1, 10) }
```
bwplotka, line 179: Do you mind adding comment with what options you recommend running this test e.g what `benchtime`? (:

pkg/receive/writer_test.go
bwplotka: Unfortunately either we have to mock the check for this in TSDB (hard) or create / update ts on every write.

bwplotka (changes requested):
Some nits, plus I think there is a minor blocker, potentially easy to fix here otherwise, great job! (:

pkg/receive/writer_test.go
matej-g: It does single quote the characters but it still makes it valid and ordered set of labels. I could make it into a string if that's more cleaner looking.

pkg/receive/writer_test.go
matej-g: So I'm not 100% sure, but it looks it has to do something how Prometheus [detects duplicates](https://github.com/prometheus/prometheus/blob/main/tsdb/head_append.go#L342). Even with an error check, the current version of benchmark does not error out. If I manipulated the timestamps to be out of order, I was seeing the conflict errors in benchmark. I'll adjust the samples to be for different timestamps, but technically it seems the behavior of the benchmark is correct (error free).

pkg/receive/writer_test.go
matej-g: This is a good point :+1: I usually followed the rule to leave error check out to remove the overhead (I would usually first run the benchmarks locally with error check to ensure it's working and then remove the check). But I agree, from the perspective of usability and future changes, it does make sense to keep the error check in.

pkg/receive/writer_test.go
matej-g: I never used non-default benchtime, are there reasons to choose non-default value here? Do you know a book that explain such things by any chance? :upside_down_face:

pkg/receive/writer_test.go
bwplotka: Great! I would say then let's say comment clearly somewhere that we DON'T include potential overhead (CPU, mem) of TSDB append as we don't append anything.

pkg/receive/writer_test.go
bwplotka: Yes , it explains (: somrething in ch7, something in ch8

pkg/receive/writer_test.go
matej-g: Will do :+1:

PR #5534, opened 2022-07-22
query: Set struct return by query api alerts same as prometheus api

audig:
Query alerts api  (/api/v1/alerts) return a list of items in object Alerts
There is a difference between api object return by Prometheus `alerts` and Thanos `Alerts`: 
A uppercase in Thanos, a lowercase in prometheus

Example of return json object: 
Thanos -> Alerts:
`json 
{"status":"success","data":{"Alerts":[..]}}
`
Prometheus -> alerts:
`json 
{"status":"success","data":{"alerts":[..]}}
`

It can be usefull to have the same struc of object. Our usecase is to use nagstamon to scrape alerts. Nagstamon can scrape prometheus, so we want to use type prometheus to get information on Thanos. This functionnality is broken due to uppercase on 'Alerts'

This PR is linked to this one which implements alerts api on thanos -> #5315

## Changes

Override representation of Alerts Object in json by using `json:"alerts"

## Verification

I build thanos with my commit, and try on my kubernetes cluster with the new version (Docker image: `audig/thanos:pr-5534`), i've checked the response of `/api/v1/alerts` and i've now an json with `alerts`
tempsnip

My use case with nagstamon is ok, i can get information about alerts in thanos with a prometheus connection type

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -13,6 +13,7 @@ We use *breaking :warning:* to mark changes that are not backward compatible (re
 
 ### Fixed
 - [#5502](https://github.com/thanos-io/thanos/pull/5502) Receive: Handle exemplar storage errors as conflict error.
+- [#5534](https://github.com/thanos-io/thanos/pull/5534) Query: Set struct return by query api alerts same as prometheus api
 
 ### Added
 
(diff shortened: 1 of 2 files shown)


fpetkovski (approved):
👍

GiedriusS (approved):

PR #5537, opened 2022-07-25
Remove empty RuleGroups in api/v1/rules when using matchers

saswatamcode:
This PR ensures that RuleGroups without any Rules are not returned in api/v1/rules response, when filtering Rules via label matchers. 

A use case for this is when implementing tenancy on top of this endpoint, and ensuring tenants don't see another tenant's Rules or RuleGroups. 

When not using label matchers all RuleGroups are returned, irrespective of the number of Rules.

## Changes

- Update `filterRules` method to remove empty Rule groups after filtering
- Update tests

## Verification

Tested locally + unit tests.

### DIFF
--- a/pkg/rules/rules.go
+++ b/pkg/rules/rules.go
@@ -86,26 +86,26 @@ func filterRules(ruleGroups []*rulespb.RuleGroup, matcherSets [][]*labels.Matche
 		return ruleGroups
 	}
 
-	gc := 0 // Group count.
+	groupCount := 0
 	for _, g := range ruleGroups {
-		rc := 0 // Rule count.
+		ruleCount := 0
 		for _, r := range g.Rules {
 			// Filter rules based on matcher.
 			rl := r.GetLabels()
 			if matches(matcherSets, rl) {
-				g.Rules[rc] = r
-				rc++
+				g.Rules[ruleCount] = r
+				ruleCount++
 			}
 		}
-		g.Rules = g.Rules[:rc]
+		g.Rules = g.Rules[:ruleCount]
 
 		// Filter groups based on number of rules.
 		if len(g.Rules) != 0 {
-			ruleGroups[gc] = g
-			gc++
+			ruleGroups[groupCount] = g
+			groupCount++
 		}
 	}
-	ruleGroups = ruleGroups[:gc]
+	ruleGroups = ruleGroups[:groupCount]
 
 	return ruleGroups
 }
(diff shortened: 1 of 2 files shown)


matej-g (approved):
Nice find and quick fix! :+1: LGTM

pkg/rules/rules.go
```diff
@@ -86,15 +86,21 @@ func filterRules(ruleGroups []*rulespb.RuleGroup, matcherSets [][]*labels.Matche
...
 			}
 		}
-		g.Rules = filteredRules
+
+		ruleGroups[i].Rules = filteredRules
+		// Filter Groups which do not have any rules.
+		if len(ruleGroups[i].Rules) == 0 {
+			ruleGroups = append(ruleGroups[:i], ruleGroups[i+1:]...)
+			i-- // -1 as the slice just got shorter
+		}
```
douglascamata, line 103: This double `for` loop is becoming complicated (modifying the loop variable `i` inside the loop itself is a bad sign) because it wants to modify `ruleGroups` in-place. Why not to create a kind of "result" slice and you only append to it if there are still rules present in a group after the filtering?

saswatamcode: Yeah, it's a bit confusing, so using https://github.com/golang/go/wiki/SliceTricks#filter-in-place instead. That way it's not violating best practice + not allocating new variables, here:  https://github.com/thanos-io/thanos/pull/5537/commits/18a69bb837fdb19c3b102d2ca389cef3122894f6 🙂

pkg/rules/rules.go
```diff
@@ -86,16 +86,26 @@ func filterRules(ruleGroups []*rulespb.RuleGroup, matcherSets [][]*labels.Matche
 		return ruleGroups
 	}
 
+	gc := 0 // Group count.
```
douglascamata, line 89: This comment is a sign that the variable should be named `groupCount` instead of `gc`.

pkg/rules/rules.go
```diff
@@ -86,16 +86,26 @@ func filterRules(ruleGroups []*rulespb.RuleGroup, matcherSets [][]*labels.Matche
...
 	for _, g := range ruleGroups {
-		filteredRules := g.Rules[:0]
+		rc := 0 // Rule count.
```
douglascamata, line 91: Same remark as the one above on line 89.

matej-g (approved):

douglascamata (approved):

PR #5538, opened 2022-07-25
Refactor EndpointSet

fpetkovski:
This commit refactors the EndpointSet struct in order to make it easier
to understand and work with.

## Changes

* Refactor endpointset.go.
* Add additional tests to help understand specific behavior of EndpointSet.

## Verification

Unit tests and manual testing.

### DIFF
--- a/pkg/query/endpointset.go
+++ b/pkg/query/endpointset.go
@@ -258,6 +258,10 @@ type EndpointSet struct {
 	endpointsMetric *endpointSetNodeCollector
 }
 
+// nowFunc is a function that returns time.Time.
+// Test code can inject a function through which
+// time can be modified before updating the EndpointSet.
+// Production code can use time.Time.
 type nowFunc func() time.Time
 
 // NewEndpointSet returns a new set of Thanos APIs.
@@ -365,7 +369,7 @@ func (e *EndpointSet) updateEndpoint(ctx context.Context, spec *GRPCEndpointSpec
 
 	if !exists {
 		var err error
-		er, err = e.newEndpointRef(ctx, e.logger, spec, e.dialOpts...)
+		er, err = e.newEndpointRef(ctx, spec)
 		if err != nil {
 			level.Warn(e.logger).Log("msg", "update of node failed", "err", err, "address", spec.Addr())
 			return
@@ -396,7 +400,7 @@ func (e *EndpointSet) pruneEndpoints(nextEndpoints map[string]*GRPCEndpointSpec)
 }
 
 // pruneTimedOutEndpoints removes unhealthy endpoints for which the last
-// successful health chech is older than the unhealthyEndpointTimeout.
+// successful health check is older than the unhealthyEndpointTimeout.
 // Strict endpoints are not removed.
 func (e *EndpointSet) pruneTimedOutEndpoints() {
 	endpoints := e.endpointRefs()
@@ -594,14 +598,14 @@ type endpointRef struct {
 
 // newEndpointRef creates a new endpointRef with a gRPC channel to the given the IP address.
 // The call to newEndpointRef will return an error if establishing the channel fails.
-func (e *EndpointSet) newEndpointRef(ctx context.Context, logger log.Logger, spec *GRPCEndpointSpec, dialOpts ...grpc.DialOption) (*endpointRef, error) {
-	conn, err := grpc.DialContext(ctx, spec.Addr(), dialOpts...)
+func (e *EndpointSet) newEndpointRef(ctx context.Context, spec *GRPCEndpointSpec) (*endpointRef, error) {
+	conn, err := grpc.DialContext(ctx, spec.Addr(), e.dialOpts...)
 	if err != nil {
 		return nil, errors.Wrap(err, "dialing connection")
 	}
 
 	return &endpointRef{
-		logger:   logger,
+		logger:   e.logger,
 		created:  e.now(),
 		addr:     spec.Addr(),
 		isStrict: spec.IsStrictStatic(),
--- a/pkg/query/endpointset.go
+++ b/pkg/query/endpointset.go
@@ -61,7 +61,7 @@ func (es *GRPCEndpointSpec) Addr() string {
 
 // Metadata method for gRPC endpoint tries to call InfoAPI exposed by Thanos components until context timeout. If we are unable to get metadata after
 // that time, we assume that the host is unhealthy and return error.
-func (es *GRPCEndpointSpec) Metadata(ctx context.Context, infoClient infopb.InfoClient, storeClient storepb.StoreClient) (*endpointMetadata, error) {
+func (es *endpointRef) Metadata(ctx context.Context, infoClient infopb.InfoClient, storeClient storepb.StoreClient) (*endpointMetadata, error) {
 	if infoClient != nil {
 		resp, err := infoClient.Info(ctx, &infopb.InfoRequest{}, grpc.WaitForReady(true))
 		if err == nil {
@@ -81,13 +81,13 @@ func (es *GRPCEndpointSpec) Metadata(ctx context.Context, infoClient infopb.Info
 	return nil, errors.New(noMetadataEndpointMessage)
 }
 
-func (es *GRPCEndpointSpec) getMetadataUsingStoreAPI(ctx context.Context, client storepb.StoreClient) (*endpointMetadata, error) {
+func (es *endpointRef) getMetadataUsingStoreAPI(ctx context.Context, client storepb.StoreClient) (*endpointMetadata, error) {
 	resp, err := client.Info(ctx, &storepb.InfoRequest{})
 	if err != nil {
 		return nil, err
 	}
 
-	infoResp := es.fillExpectedAPIs(component.FromProto(resp.StoreType), resp.MinTime, resp.MaxTime)
+	infoResp := fillExpectedAPIs(component.FromProto(resp.StoreType), resp.MinTime, resp.MaxTime)
 	infoResp.LabelSets = resp.LabelSets
 	infoResp.ComponentType = component.FromProto(resp.StoreType).String()
 
@@ -96,7 +96,7 @@ func (es *GRPCEndpointSpec) getMetadataUsingStoreAPI(ctx context.Context, client
 	}, nil
 }
 
-func (es *GRPCEndpointSpec) fillExpectedAPIs(componentType component.Component, mintime, maxTime int64) infopb.InfoResponse {
+func fillExpectedAPIs(componentType component.Component, mintime, maxTime int64) infopb.InfoResponse {
 	switch componentType {
 	case component.Sidecar:
 		return infopb.InfoResponse{
@@ -246,7 +246,7 @@ type EndpointSet struct {
 
 	// Endpoint specifications can change dynamically. If some component is missing from the list, we assume it is no longer
 	// accessible and we close gRPC client for it, unless it is strict.
-	endpointSpec             func() []*GRPCEndpointSpec
+	endpointSpec             func() map[string]*GRPCEndpointSpec
 	dialOpts                 []grpc.DialOption
 	gRPCInfoCallTimeout      time.Duration
 	unhealthyEndpointTimeout time.Duration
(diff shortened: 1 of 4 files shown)


pkg/query/endpointset.go
```diff
@@ -600,22 +550,6 @@ func (e *EndpointSet) GetEndpointStatus() []EndpointStatus {
 	return statuses
 }
 
-func (e *EndpointSet) cleanUpEndpointStatuses(endpoints map[string]*endpointRef) {
```
fpetkovski, line 603: Any idea why this was here? From what I understand, strict endpoints are always used, and non-strict endpoints are removed when info calls fail.

fpetkovski: Ok figured it out. We actually want to keep unhealthy endpoints around for a while for the purpose of showing them as `DOWN` in the UI. 

Update the PR and added some more tests for this use case.

fpetkovski:
cc @saswatamcode

pkg/query/endpointset.go
```diff
@@ -345,28 +350,125 @@ func (e *EndpointSet) Update(ctx context.Context) {
...
+		}
+	}
+
+	for _, er := range obsoleteEndpoints {
+		e.removeEndpoint(er)
+	}
+}
+
+// pruneTimedOutEndpoints removes unhealthy endpoints for which the last
+// successful health chech is older than the unhealthyEndpointTimeout.
```
saswatamcode, line 399:
```suggestion
// successful health check is older than the unhealthyEndpointTimeout.
```

pkg/query/endpointset.go
```diff
@@ -241,29 +241,28 @@ func (c *endpointSetNodeCollector) Collect(ch chan<- prometheus.Metric) {
...
+	endpointsMtx    sync.RWMutex
 	endpoints       map[string]*endpointRef
 	endpointsMetric *endpointSetNodeCollector
-
-	// Map of statuses used only by UI.
-	endpointStatuses         map[string]*EndpointStatus
-	unhealthyEndpointTimeout time.Duration
 }
 
+type nowFunc func() time.Time
```
saswatamcode, line 261: Maybe a comment as to why this is a function instead of just `time.Time` would be great? Afaiu, this is so that we can get the exact current time while pruning endpoints and setting endpoint status. 🙂

saswatamcode (dismissed):
Thanks a ton for this! 🙂 

It makes EndpointSet much easier to understand, especially with EndpointStatus being a part of endpointRef. And functions like `isQueryable` make it way more organized!

Some tiny comments!

pkg/query/endpointset.go
fpetkovski: Yes good idea, added 👍

pkg/query/endpointset.go
bwplotka: 👍🏽

pkg/query/endpointset.go
```diff
@@ -302,39 +306,44 @@ func (e *EndpointSet) Update(ctx context.Context) {
...
+		e.pruneEndpoints(endpointSpecs)
+	}()
+	wg.Wait()
+
+	e.pruneTimedOutEndpoints()
+
+	newEndpoints := e.endpointRefs()
+	level.Debug(e.logger).Log("msg", "updated endpoints", "activeEndpoints", len(newEndpoints))
 
+	// Update stats
```
bwplotka, line 344:
```suggestion
	// Update stats.
```

pkg/query/endpointset.go
```diff
@@ -302,39 +306,44 @@ func (e *EndpointSet) Update(ctx context.Context) {
...
-	for addr, er := range e.endpoints {
-		endpoints[addr] = er
-	}
-	e.endpointsMtx.RUnlock()
-
-	level.Debug(e.logger).Log("msg", "starting to update API endpoints", "cachedEndpoints", len(endpoints))
-
-	activeEndpoints := e.getActiveEndpoints(ctx, endpoints)
-	level.Debug(e.logger).Log("msg", "checked requested endpoints", "activeEndpoints", len(activeEndpoints), "cachedEndpoints", len(endpoints))
+	currentEndpoints := e.endpointRefs()
```
bwplotka, line 309:
```suggestion
	currentEndpoints := e.cachedEndpoints()
```
I am missing what kind of refs and endpoints we are returning here

pkg/query/endpointset.go
```diff
@@ -302,39 +306,44 @@ func (e *EndpointSet) Update(ctx context.Context) {
...
 
-	stats := newEndpointAPIStats()
-
-	// Close endpoints which are not active this time (are not in active endpoints map).
-	for addr, er := range endpoints {
-		if _, ok := activeEndpoints[addr]; ok {
-			stats[er.ComponentType()][labelpb.PromLabelSetsToString(er.LabelSets())]++
+	endpointSpecs := make(map[string]*GRPCEndpointSpec)
+	for _, spec := range e.endpointSpec() {
+		if _, duplicate := endpointSpecs[spec.Addr()]; duplicate {
```
bwplotka, line 314: Do we need this if we have https://github.com/thanos-io/thanos/blob/6d426d2b300ee192e617d7f5be1041aef308c798/cmd/thanos/query.go#L448 dedup ?

pkg/query/endpointset.go
```diff
@@ -345,28 +354,125 @@ func (e *EndpointSet) Update(ctx context.Context) {
...
+func (e *EndpointSet) updateEndpoint(ctx context.Context, spec *GRPCEndpointSpec) {
+	e.endpointsMtx.RLock()
+	er, exists := e.endpoints[spec.Addr()]
+	e.endpointsMtx.RUnlock()
+
+	if !exists {
+		var err error
+		er, err = e.newEndpointRef(ctx, spec)
+		if err != nil {
+			level.Warn(e.logger).Log("msg", "update of node failed", "err", err, "address", spec.Addr())
```
bwplotka, line 374:
```suggestion
			level.Warn(e.logger).Log("msg", "new endpoint creation failed", "err", err, "address", spec.Addr())
```

pkg/query/endpointset.go
```diff
@@ -345,28 +354,125 @@ func (e *EndpointSet) Update(ctx context.Context) {
...
+			return
+		}
 	}
 
-	e.endpointsMetric.Update(stats)
+	metadata, err := spec.Metadata(ctx, infopb.NewInfoClient(er.cc), storepb.NewStoreClient(er.cc))
+	if err != nil {
+		level.Warn(e.logger).Log("msg", "update of node failed", "err", errors.Wrap(err, "getting metadata"), "address", spec.Addr())
```
bwplotka, line 381:
```suggestion
		level.Warn(e.logger).Log("msg", "update of endpoint failed", "err", errors.Wrap(err, "getting metadata"), "address", spec.Addr())
```

pkg/query/endpointset.go
```diff
@@ -345,28 +354,125 @@ func (e *EndpointSet) Update(ctx context.Context) {
 				"address", addr, "extLset", extLset, "duplicates", fmt.Sprintf("%v", stats[component.Sidecar][extLset]+stats[component.Rule][extLset]+1))
 		}
 		stats[er.ComponentType()][extLset]++
+		if _, ok := currentEndpoints[er.addr]; !ok {
+			level.Info(e.logger).Log("msg", fmt.Sprintf("adding new %v with %+v", er.ComponentType(), er.apisPresent()), "address", addr, "extLset", extLset)
```
bwplotka, line 358: Why adding this here and not when we actually added it?

pkg/query/endpointset.go
```diff
@@ -345,28 +354,125 @@ func (e *EndpointSet) Update(ctx context.Context) {
...
 }
 
-// GetStoreClients returns a list of all active stores.
-func (e *EndpointSet) GetStoreClients() []store.Client {
+func (e *EndpointSet) removeEndpoint(er *endpointRef) {
+	e.endpointsMtx.Lock()
+	defer e.endpointsMtx.Unlock()
+
+	if _, ok := e.endpoints[er.Addr()]; ok {
+		level.Info(er.logger).Log("msg", unhealthyEndpointMessage, "address", er.Addr(), "extLset", labelpb.PromLabelSetsToString(er.LabelSets()))
```
bwplotka, line 437: to be consistent I would add addition log to `addEndpoint` as well.

pkg/query/endpointset.go
```diff
@@ -345,28 +354,125 @@ func (e *EndpointSet) Update(ctx context.Context) {
...
+		delete(e.endpoints, er.Addr())
+	}
+	er.Close()
+}
+
+// endpointsMap returns a map from endpoint address to *endpointRef.
```
bwplotka, line 443: obsolete comment

bwplotka:
```suggestion
// cachedEndpoints returns a map from endpoint address to *endpointRef.
```

pkg/query/endpointset.go
```diff
@@ -345,28 +354,125 @@ func (e *EndpointSet) Update(ctx context.Context) {
...
+	for _, er := range obsoleteEndpoints {
+		e.removeEndpoint(er)
+	}
+}
+
+// pruneTimedOutEndpoints removes unhealthy endpoints for which the last
+// successful health check is older than the unhealthyEndpointTimeout.
+// Strict endpoints are not removed.
+func (e *EndpointSet) pruneTimedOutEndpoints() {
+	endpoints := e.endpointRefs()
```
bwplotka, line 406: Hm, wouldn't it be easier if we would operate during full Update duration on local copy of endpoints like before?

Pros:
* Avoiding constant lock/unlock of `endpointRefs` which might slow down a bit readers of map (every query/API use on Querier I assume). Not a significant slowness in practice probably, though (to measure potentially).
* Simpler code and debugging flow - Update is atomic. No partial read upon update, e.g Fetch of endpoint in the middle of configuration changed we apply to endpoint file or DNS. I can't see a major problem, other than surprise when someone updates configuration/DNS with current code, but it's some unknown piece that we have test or think about  (partial reads). Plus it's not easy to test against, prone to semantic race.

Cons:
* Insignificantly less interactive updates since we have to wait for full update, but I don't see this being a huge problem.

bwplotka (commented):
Great job, thanks!

The main question is around partial reads. I personally would vote for simpler code and simpler flow but not a blocker (:

bwplotka:
cc @SrushtiSapkale for review as @fpetkovski is refactoring code we will be using in your project (:

pkg/query/endpointset.go
fpetkovski: I thought about this as well, but we also have endpoints that are update in-place, so we can't easily have full atomicity (and we probably don't need to). 

However, I think we can definitely make additions and removals of endpoints atomic. Now that we have more precise tests, it should be safe and easy to make the change. I'll give it a try.

pkg/query/endpointset.go
fpetkovski: Hm good question. I think EndpointSet should be self-contained and not rely on external assumptions. We might want to use it in other parts of the project (query frontend comes to mind since querier now has a gRPC API). So maybe it's best to stay defensive?

pkg/query/endpointset.go
fpetkovski: I think I just took this over from the previous implementation. Will remove.

fpetkovski: What I've also noticed is that these logs can be very verbose when the labelset is present. So we might want to log the IP and supported APIs, at least with info level?

pkg/query/endpointset.go
bwplotka: Thanks, up to you - I just thought it will simplify things

fpetkovski: Additions and removals should be atomic now. It also made the code much more concise and simple.

pkg/query/endpointset.go
fpetkovski: I moved this into the EndpointSet constructor. We now make sure whatever is returned from specs() is deduplicated already.

bwplotka (approved):
Thanks, amazing! 💪🏽

Issue #5335, opened 2022-05-06
Load balancing for Store Gateways


fpetkovski:
### Is your proposal related to a problem?

It currently makes little sense to run store gateways in HA mode since Queriers do a fanout across all stores. 
In order to enable seamless updates and allow spreading spreading load more evenly, we should consider implementing load balancing capabilities in Queriers.

### Describe the solution you'd like
Ideally, load-balancing should be done on the client-side to mitigate the stickiness of gRPC connections. 
We can identify replicated stores by their labels when selecting series and instead of fanning out across all stores, we can use round-robin or pick replicas at random.

One potential solution would be to introduce a new object that implements the `store.Client` interface and holds references to stores with the same labels. This object would be returned by and will randomly proxy requests to one of the downstream stores.

https://github.com/thanos-io/thanos/blob/3de8ee7245967c6660beab1e0ebf5015b02137b3/pkg/query/endpointset.go#L364-L375

### Describe alternatives you've considered

This was already discussed in https://github.com/thanos-io/thanos/issues/199 and was closed with a suggestion to use a load balancer or rely on kube-proxy in Kubernetes. The suggested approach will not work due to the sticky nature of gRPC connections.

wiardvanrij:
We do have to consider the fact that we don't want this for sidecars. As data can be different per Prometheus instance. 
Currently, the most efficient way (imo) for load spread is to implement sharding. I would be curious what 'overhead' we would have with load balancing over multiple stores. Anyhow +1 from me. 
Might even be a cool LFX project as well.

fpetkovski:
Still relevant.

PR #5505, opened 2022-07-15
Proposal: Granular Endpoint Configuration

SrushtiSapkale:
Signed-off-by: Srushti Sapkale <srushtiisapkale@gmail.com>

* [ ] I added CHANGELOG entry for this change.
* [x] Change is not relevant to the end user.

PR #5548, opened 2022-07-28
Implement endpoint groups

fpetkovski:
This draft PR illustrates how an HA group of endpoints can be configured in Thanos Query.

Looking for feedback on the approach!

Fixes https://github.com/thanos-io/thanos/issues/5335

## Changes

Add the flags `endpoint-group` and `endpoint-group-strict` to Thanos Query for load balancing instead of fanout.

### DIFF
--- a/cmd/thanos/query.go
+++ b/cmd/thanos/query.go
@@ -11,6 +11,8 @@ import (
 	"strings"
 	"time"
 
+	"google.golang.org/grpc"
+
 	"github.com/go-kit/log"
 	"github.com/go-kit/log/level"
 	grpc_logging "github.com/grpc-ecosystem/go-grpc-middleware/v2/interceptors/logging"
@@ -128,6 +130,9 @@ func registerQuery(app *extkingpin.App) {
 	endpoints := extkingpin.Addrs(cmd.Flag("endpoint", "Addresses of statically configured Thanos API servers (repeatable). The scheme may be prefixed with 'dns+' or 'dnssrv+' to detect Thanos API servers through respective DNS lookups.").
 		PlaceHolder("<endpoint>"))
 
+	endpointGroups := extkingpin.Addrs(cmd.Flag("endpoint-group", "DNS name of statically configured Thanos API server groups (repeatable). Targets resolved from the DNS name will be queried in a round-robin instead of a fanout manner. This flag should be used when connecting a Thanos Query to HA groups of Thanos components.").
+		PlaceHolder("<endpoint-group>"))
+
 	stores := extkingpin.Addrs(cmd.Flag("store", "Deprecation Warning - This flag is deprecated and replaced with `endpoint`. Addresses of statically configured store API servers (repeatable). The scheme may be prefixed with 'dns+' or 'dnssrv+' to detect store API servers through respective DNS lookups.").
 		PlaceHolder("<store>"))
 
@@ -150,6 +155,9 @@ func registerQuery(app *extkingpin.App) {
 
 	strictEndpoints := cmd.Flag("endpoint-strict", "Addresses of only statically configured Thanos API servers that are always used, even if the health check fails. Useful if you have a caching layer on top.").
 		PlaceHolder("<staticendpoint>").Strings()
+
+	strictEndpointGroups := extkingpin.Addrs(cmd.Flag("endpoint-group-strict", "DNS name of statically configured Thanos API server groups (repeatable) that are always used, even if the health check fails.").
+		PlaceHolder("<endpoint-group-strict>"))
 
 	fileSDFiles := cmd.Flag("store.sd-files", "Path to files that contain addresses of store API servers. The path can be a glob pattern (repeatable).").
 		PlaceHolder("<path>").Strings()
@@ -291,6 +299,7 @@ func registerQuery(app *extkingpin.App) {
 			selectorLset,
 			getFlagsMap(cmd.Flags()),
 			*endpoints,
+			*endpointGroups,
 			*stores,
 			*ruleEndpoints,
 			*targetEndpoints,
@@ -312,6 +321,7 @@ func registerQuery(app *extkingpin.App) {
 			*defaultMetadataTimeRange,
 			*strictStores,
 			*strictEndpoints,
+			*strictEndpointGroups,
 			*webDisableCORS,
 			enableQueryPushdown,
 			*alertQueryURL,
@@ -367,6 +377,7 @@ func runQuery(
 	selectorLset labels.Labels,
 	flagsMap map[string]string,
 	endpointAddrs []string,
+	endpointGroupAddrs []string,
 	storeAddrs []string,
 	ruleAddrs []string,
 	targetAddrs []string,
@@ -388,6 +399,7 @@ func runQuery(
 	defaultMetadataTimeRange time.Duration,
 	strictStores []string,
 	strictEndpoints []string,
+	strictEndpointGroups []string,
 	disableCORS bool,
 	enableQueryPushdown bool,
 	alertQueryURL string,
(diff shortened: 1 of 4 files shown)


fpetkovski:
One downside of using load balancing from gRPC is that dns resolution is done once during startup and is cached for a very long time, potentially forever. Addresses will be re-resolved when the downstream target goes away. 

This can be problematic if the endpoing group is scaled out and new targets are added. The recommended workaround here is to set `max_connection_age` to something like `5m` which will cause periodic dns resolution, but also a complete recreation of all connections to the endpoint group.

fpetkovski:
cc @saswatamcode @SrushtiSapkale

pkg/extgrpc/client.go
```diff
@@ -20,6 +20,27 @@ import (
 	"github.com/thanos-io/thanos/pkg/tracing"
 )
 
+// EndpointGroupGRPCOpts creates gRPC dial options for connecting to endpoint groups.
+// For details on retry capabilities, see https://github.com/grpc/proposal/blob/master/A6-client-retries.md#retry-policy-capabilities
+func EndpointGroupGRPCOpts() []grpc.DialOption {
+	serviceConfig := `
+{
+  "loadBalancingPolicy":"round_robin",
+  "retryPolicy": {
```
saswatamcode, line 29: Perhaps this should be configurable via flags?

pkg/query/endpointset.go
```diff
@@ -590,7 +595,8 @@ type endpointRef struct {
 // newEndpointRef creates a new endpointRef with a gRPC channel to the given the IP address.
 // The call to newEndpointRef will return an error if establishing the channel fails.
 func (e *EndpointSet) newEndpointRef(ctx context.Context, spec *GRPCEndpointSpec) (*endpointRef, error) {
-	conn, err := grpc.DialContext(ctx, spec.Addr(), e.dialOpts...)
+	dialOpts := append(e.dialOpts, spec.dialOpts...)
+	conn, err := grpc.DialContext(ctx, spec.Addr(), dialOpts...)
```
saswatamcode, line 599: Afaiu, this would mean that any group of Thanos endpoints can be configured to be load-balanced. But we don't check if those components are actually HA via same labels or so. 

So in case of faulty configuration, debugging might be confusing here.

saswatamcode (dismissed):
Thanks a lot for this! ✨

Some comments!

pkg/extgrpc/client.go
fpetkovski: Yes, this could be a good idea. The new endpoint config would also help here.

pkg/query/endpointset.go
fpetkovski: That is correct, and I think this is not something we can easily solve. It seems to be one of the drawbacks of using the built-in load balancing in gRPC alongside the max connection age workaround.

fpetkovski:
I am not sure what the status of the LFX project is, but this could be a short term solution that gets us unblocked until we have a better implementation.

trevorriles:
Currently very interested in this work. What are the blockers for this being merged?

fpetkovski:
There was an LFX project that was a superset of this functionality: https://github.com/thanos-io/thanos/pull/5505

Maybe @saswatamcode or @matej-g have some information on how far that got and whether it makes sense to continue working on this PR.

trevorriles:
> There was an LFX project that was a superset of this functionality: #5505
> 
> Maybe @saswatamcode or @matej-g have some information on how far that got and whether it makes sense to continue working on this PR.

Thanks for the link!

saswatamcode:
Yes, we made some progress on that end, but I think there were some more items left as well. This is also related to https://github.com/thanos-io/thanos/issues/2600

I'll try to organize them a bit, and move this forward! 🙂

fpetkovski:
I see that HA endpoints are out of scope of that proposal: https://github.com/thanos-io/thanos/pull/5505/files#diff-5dad1d444b473dcd0b72f4770b3ba03089499cfaf027205c83914f63124644e5R38. Are you looking to cover them in your work?

fpetkovski:
@SuperQ brought this feature up yesterday during contributor hours. Do we want to proceed with something like this until we have more extensive endpoint configuration? It would be great to not have to run envoy for load balancing gRPC connections since it's another component in the query path that can break. cc @saswatamcode @bwplotka

pkg/extgrpc/client.go
saswatamcode: Can we add this flag in this PR, with this particular config as the default? 🙂

cmd/thanos/query.go
```diff
@@ -128,6 +128,9 @@ func registerQuery(app *extkingpin.App) {
 	endpoints := extkingpin.Addrs(cmd.Flag("endpoint", "Addresses of statically configured Thanos API servers (repeatable). The scheme may be prefixed with 'dns+' or 'dnssrv+' to detect Thanos API servers through respective DNS lookups.").
 		PlaceHolder("<endpoint>"))
 
+	endpointGroups := extkingpin.Addrs(cmd.Flag("endpoint-group", "DNS name of statically configured Thanos API server groups (repeatable). Targets resolved from the DNS name will be queried in a round-robin instead of a fanout manner. This flag should be used when connecting a Thanos Query to HA groups of Thanos components.").
```
saswatamcode, line 131:
```suggestion
	endpointGroups := extkingpin.Addrs(cmd.Flag("endpoint-group", "DNS name of statically configured Thanos API server groups (repeatable). Targets resolved from the DNS name will be queried in a round-robin manner by default instead of a fanout manner. This flag should be used when connecting a Thanos Query to HA groups of Thanos components.").
```

cmd/thanos/query.go
```diff
@@ -500,6 +510,20 @@ func runQuery(
...
+					addr := fmt.Sprintf("dns:///%s", eg)
+					spec := query.NewGRPCEndpointSpec(addr, false, extgrpc.EndpointGroupGRPCOpts()...)
+					specs = append(specs, spec)
+				}
+
+				for _, eg := range strictEndpointGroups {
+					grpc.WithDisableRetry()
+
+					addr := fmt.Sprintf("dns:///%s", eg)
+					spec := query.NewGRPCEndpointSpec(addr, true, extgrpc.EndpointGroupGRPCOpts()...)
```
saswatamcode, line 523: Should `grpc.WithDisableRetry` be passed along side the other grpc options?
```suggestion
					addr := fmt.Sprintf("dns:///%s", eg)
					spec := query.NewGRPCEndpointSpec(addr, true, append(extgrpc.EndpointGroupGRPCOpts(), grpc.WithDisableRetry())...)
```

saswatamcode (approved):
Thanks! I think it would be great to proceed with this until we have a full-fledged endpoint configuration! 💪🏻 

Just some minor comments on this!

pkg/extgrpc/client.go
fpetkovski: This policy is really an implementation detail, I don't think it makes sense to expose it to the end user. The retry always has to be there so that the client will re-resolve dns endpoints when the one it's connected to goes away. This is how HA is implemented in gRPC.

cmd/thanos/query.go
fpetkovski: Ah, I think this like should be completely removed. The retry should always be there as explained in the comment below.

pkg/extgrpc/client.go
saswatamcode: I see!

saswatamcode (dismissed):
LGTM!

fpetkovski:
I marked these flags as experimental, there could be some hidden dragons that we'll uncover over time.

docs/components/tools.md
```diff
@@ -813,7 +813,7 @@ Flags:
       --rewrite.to-relabel-config-file=<file-path>
                                 Path to YAML file that contains relabel configs
                                 that will be applied to blocks
-      --tmp.dir="/tmp/thanos-rewrite"
+      --tmp.dir="/var/folders/0b/r42s6hv96hs2w84ycr2yf7qh0000gn/T/thanos-rewrite"
```
saswatamcode, line 816: Seems like this causes docs CI to fail

saswatamcode (approved):

docs/components/tools.md
fpetkovski: Thanks, looks like this happens on my machine regularly.

saswatamcode: Yeah I think it happens due to os.MkdirTemp, which behaves differently on mac

cmd/thanos/query.go
```diff
@@ -128,6 +128,9 @@ func registerQuery(app *extkingpin.App) {
 	endpoints := extkingpin.Addrs(cmd.Flag("endpoint", "Addresses of statically configured Thanos API servers (repeatable). The scheme may be prefixed with 'dns+' or 'dnssrv+' to detect Thanos API servers through respective DNS lookups.").
 		PlaceHolder("<endpoint>"))
 
+	endpointGroups := extkingpin.Addrs(cmd.Flag("endpoint-group", "Experimental: DNS name of statically configured Thanos API server groups (repeatable). Targets resolved from the DNS name will be queried in a round-robin, instead of a fanout manner. This flag should be used when connecting a Thanos Query to HA groups of Thanos components.").
```
benjaminhuo, line 131: `HA groups of Thanos components.` means store only?

fpetkovski: It can be any component, querier, store, exemplar etc.. as long as they implement the same APIs. So a group should not have a mix of components.

PR #5553, opened 2022-07-29
Tracing: Add Jaeger B3 header support

ahayworth:

## Changes

Jaeger tracing uses its own HTTP/GRPC header to perform context
propagation between services - "uber-trace-id". However, Jaeger
tracing clients also support context propagation via the B3 header
format (as popularized by Zipkin tracing).

This commit adds support for configuring the B3 header format within
thanos tracing. The main motivation for doing this is interoperability
with other systems that may call into or out of Thanos components - for
example, Grafana or Envoy - both of which support Jaeger tracing, but
use B3 headers instead of the Jaeger-specific "uber-trace-id" format headers.

Unfortunately, we cannot directly toggle this as an option or configure
it via the environment; the jaeger-go library requires us to pass it in
as an option when we call `NewTracer`. In order to facilitate this, we
extend the `ParseConfigFromYaml` file to not only return a _Jaeger_
configuration object, but the actual YAML configuration object as well.
This allows us to check and see if the user requested this kind of
header propagation, and we can then construct the necessary additional
options to `NewTracer`.

Of course, this is all legacy configuration in many ways. The tracing
world is rapidly moving towards OpenTelemetry and W3C headers for
context propagation. However, there is still utility in supporting B3
headers until everyone (including Thanos!) finishes their OpenTelemetry
migration projects.

## Verification

Programmatically testing this change is tricky, because we need to actually
inspect the headers of the HTTP and GRPC requests that flow through the
system with tracing enabled. I was able to verify that this worked through
manual testing via the query frontend:

1. I started a local ngrok tunnel, which allows me to intercept requests
   between the query frontend and a downstream querier (I configured it
   statically for port 10904, which is one of the querier ports).
2. I launched a local OpenTelemetry Collector and Jaeger all-in-one image
   to receive and display traces.
3. I modified `scripts/quickstart.sh` to:
  - Add a query frontend, with the `query-frontend.downstream-url` pointed
    to the intercepting `ngrok` tunnel.
  - Add appropriate tracing configurations for all components.

Once those modifications were in place and launched with `make quickstart`,
I was able to write queries in the query frontend, see that the intercepted
API calls to the downstream querier contained tracing headers in the
expected format, and verify that the traces were ingested into the Jaeger
backend via the OpenTelemetry collector (the OpenTelemetry collector was not
strictly necessary, but I prefer to use it while debugging).

Screenshot showing the intercepted API request, with correct B3 headers:
Screen Shot 2022-07-29 at 11 16 35 AM

Screenshot of the complete trace ingested to the Jaeger backend, with
correct context-propagation and parent/child span relationships preserved:
Screen Shot 2022-07-29 at 11 16 44 AM

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -10,6 +10,7 @@ We use *breaking :warning:* to mark changes that are not backward compatible (re
 
 ## Unreleased
 - [#5453](https://github.com/thanos-io/thanos/pull/5453) Compact: Skip erroneous empty non `*AggrChunk` chunks during 1h downsampling of 5m resolution blocks.
+- [#5553](https://github.com/thanos-io/thanos/pull/5553) Tracing: Add Jaeger B3 header support.
 
 ### Fixed
 - [#5502](https://github.com/thanos-io/thanos/pull/5502) Receive: Handle exemplar storage errors as conflict error.
--- a/docs/tracing.md
+++ b/docs/tracing.md
@@ -95,6 +95,7 @@ config:
   agent_host: ""
   agent_port: 0
   traceid_128bit: false
+  use_b3_headers: false
 ```
 
 ### Google Cloud (formerly Stackdriver)
--- a/pkg/tracing/jaeger/config_yaml.go
+++ b/pkg/tracing/jaeger/config_yaml.go
@@ -38,14 +38,21 @@ type Config struct {
 	AgentHost              string        `yaml:"agent_host"`
 	AgentPort              int           `yaml:"agent_port"`
 	Gen128Bit              bool          `yaml:"traceid_128bit"`
+
+	// In order to use B3 headers, we must construct B3 propagators
+	// and then directly pass them as an option to NewTracer (which
+	// happens in `jaeger.go`). There is no appropriate configuration
+	// toggle to declaratively use B3 headers, and also no environment
+	// variable support.
+	UseB3Headers bool `yaml:"use_b3_headers"`
 }
 
 // ParseConfigFromYaml uses config YAML to set the tracer's Configuration.
-func ParseConfigFromYaml(cfg []byte) (*config.Configuration, error) {
+func ParseConfigFromYaml(cfg []byte) (*config.Configuration, *Config, error) {
 	conf := &Config{}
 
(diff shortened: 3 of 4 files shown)


ahayworth:
cc @fpetkovski, since I think you may be interested in this.

GiedriusS (commented):
Is there a reason not to add these by default?

yeya24:
Since we have another WIP pr for jaeger Opentelemetry migration, maybe it is better to have this after https://github.com/thanos-io/thanos/pull/5411 is done?

ahayworth:
> Is there a reason not to add these by default?

@GiedriusS That depends on exactly what you mean with the question.
- If you mean "make the default `true`", then I would say that it's a breaking change for anyone that's relying on the current behavior today (and I don't know how many people that might be).
- If you mean "why not always add those headers _in addition_ to the normal `uber-trace-id` header", that's certainly possible. In _OpenTelemetry_ terms, that would be a ["composite propagator"](https://opentelemetry.io/docs/reference/specification/context/api-propagators/#composite-propagator), and it isn't extraordinarily difficult to implement. However, the OpenTracing standard never defined such a concept and so the `jaeger-client-go` library didn't implement one (as far as I can see, anyways). We'd just need to write that here. That said: I don't know if enough people would want it to justify the effort.

> Since we have another WIP pr for jaeger Opentelemetry migration, maybe it is better to have this after https://github.com/thanos-io/thanos/pull/5411 is done?

@yeya24 If that's going to merge soon, then perhaps it is better to wait. My impression was that the PR may not be close to merging given the length of time it has been open, but that impression may have been mistaken! 😄 

That said - it looks like #5411 would configure W3C context propagation for all supported tracing backends, which may not be what is desired (or intended). There's a new [`autoprop` package](https://github.com/open-telemetry/opentelemetry-go-contrib/pull/2258) for the otel-go SDK which may be an attractive option, since it would allow us to easily extend support for additional propagation types other than W3C and also allows for the otel-standard `OTEL_PROPAGATORS` env var support. If included in #5411, that could make this PR obsolete entirely.

fpetkovski:
Looks like this PR is not ready for review. Does it make sense to focus efforts there @ahayworth ?

Issue #5444, opened 2022-06-24
Thanos Receive: High query latency for exemplars with zero exemplars in storage


vanugrah:
**Thanos, Prometheus and Golang version used**:
- Thanos: 0.26.0
- Prometheus: 2.35.0

**Object Storage Provider**:
Pure Storage 

**What happened**:
Hey folks, 

We've encountered an issue with our production thanos deployment that we suspect may be a bug. Long story short we are using thanos in both the sidecar approach alongside Prometheus as well as with Thanos receive and remote write. 

While the majority of our thanos infrastructure was already on version 0.26, our thanos receive deployment and the querier sitting in front of it was still running version 0.23. Last week we upgraded thanos receive to version 0.26 and began noticing something quite surprising: our query latency from Grafana degraded by over 100x! This resulted in Grafana taking an additional 1-2 minutes to load data into panels, even for queries spanning just a few hours. 

Tracing revealed that the query_range methods were still blazing fast (subsecond) in most cases. However exemplar queries to Thanos Receive were very slow (30-50s) and in some cases timing out. 

![Screen Shot 2022-06-24 at 6 49 42 PM](https://user-images.githubusercontent.com/11529857/175528676-cc370473-1da7-45a6-9d70-083749f0fe87.png)

This came as an even greater shock because we do not use exemplars. We have never enabled exemplar collection in Prometheus nor have we enabled exemplar storage in thanos receive (max exemplar storage is set to zero). I've also confirmed that we in fact have absolutely no exemplars in storage via the following metrics: 

![Screen Shot 2022-06-24 at 6 54 02 PM](https://user-images.githubusercontent.com/11529857/175529194-47ffbfca-2dfc-46d6-9755-7728706aa8b3.png)

To make things even more interesting, we are only seeing this behavior with Thanos receive. This behavior is not observed between thanos query -> thanos sidecar -> prometheus. In most cases exemplar queries to thanos sidecar respond in less than 150ms which can be seen in the above trace.

For the time being we've disabled exemplar querying from Grafana to restore our query performance, but I think it is worth investigating what is causing this query latency specifically with thanos receive. 

**What you expected to happen**:
Thanos receive exemplar queries perform comparably to thanos sidecar exemplar queries given that neither thanos receive nor Prometheus have exemplar storage enabled. 

**How to reproduce it (as minimally and precisely as possible)**:
Query thanos receive with exemplars enabled. 

**Anything else we need to know**:
[...]

PR #5554, opened 2022-07-30
Fix multi-tenant exemplar matchers

fpetkovski:
The exemplar proxy synthesizes a query based on PromQL expression matchers
and individual store's label sets. When a store has multiple label sets
with same label names but different values (e.g. multitenant Receivers),
each exemplar matcher will be repeated once for each label set. Because of this,
a receiver hosting 200 tenants can get the same exemplar matcher 200 times. This leads
to the underlying stores slowing down and timing out when asked for exemplars.

This commit modifies the exemplar proxy to deduplicate matchers and only send
a matcher once to an underlying store.

Fixes https://github.com/thanos-io/thanos/issues/5444

## Changes

Fix querying exemplars from multi-tenant stores.

## Verification

Unit tests and testing in our environment.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -13,7 +13,8 @@ We use *breaking :warning:* to mark changes that are not backward compatible (re
 
 ### Fixed
 - [#5502](https://github.com/thanos-io/thanos/pull/5502) Receive: Handle exemplar storage errors as conflict error.
-- [#5534](https://github.com/thanos-io/thanos/pull/5534) Query: Set struct return by query api alerts same as prometheus api
+- [#5534](https://github.com/thanos-io/thanos/pull/5534) Query: Set struct return by query api alerts same as prometheus api.
+- [#5554](https://github.com/thanos-io/thanos/pull/5534) Query/Receiver: Fix querying exemplars from multi-tenant receivers.
 
 ### Added
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -14,7 +14,7 @@ We use *breaking :warning:* to mark changes that are not backward compatible (re
 ### Fixed
 - [#5502](https://github.com/thanos-io/thanos/pull/5502) Receive: Handle exemplar storage errors as conflict error.
 - [#5534](https://github.com/thanos-io/thanos/pull/5534) Query: Set struct return by query api alerts same as prometheus api.
-- [#5554](https://github.com/thanos-io/thanos/pull/5534) Query/Receiver: Fix querying exemplars from multi-tenant receivers.
+- [#5554](https://github.com/thanos-io/thanos/pull/5554) Query/Receiver: Fix querying exemplars from multi-tenant receivers.
 
 ### Added
 
--- a/pkg/exemplars/proxy.go
+++ b/pkg/exemplars/proxy.go
@@ -80,8 +80,10 @@ func (s *Proxy) Exemplars(req *exemplarspb.ExemplarsRequest, srv exemplarspb.Exe
 		exemplars []*exemplarspb.ExemplarData
 	)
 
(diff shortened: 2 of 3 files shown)


CHANGELOG.md
```diff
@@ -13,7 +13,8 @@ We use *breaking :warning:* to mark changes that are not backward compatible (re
 
 ### Fixed
 - [#5502](https://github.com/thanos-io/thanos/pull/5502) Receive: Handle exemplar storage errors as conflict error.
-- [#5534](https://github.com/thanos-io/thanos/pull/5534) Query: Set struct return by query api alerts same as prometheus api
+- [#5534](https://github.com/thanos-io/thanos/pull/5534) Query: Set struct return by query api alerts same as prometheus api.
+- [#5554](https://github.com/thanos-io/thanos/pull/5534) Query/Receiver: Fix querying exemplars from multi-tenant receivers.
```
yeya24, line 17: Let's update the PR number?

pkg/exemplars/proxy_test.go
```diff
@@ -5,12 +5,17 @@ package exemplars
...
 	"io"
 	"os"
 	"reflect"
 	"sync"
 	"testing"
 
+	"github.com/prometheus/prometheus/promql/parser"
+	"google.golang.org/grpc/codes"
+	"google.golang.org/grpc/status"
+
```
yeya24, line 18: Can we group the imports?

pkg/exemplars/proxy.go
```diff
@@ -96,27 +98,27 @@ func (s *Proxy) Exemplars(req *exemplarspb.ExemplarsRequest, srv exemplarspb.Exe
...
 			}
-			// Construct the query by concatenating metric selectors with '+'.
-			// We cannot preserve the original query info, but the returned
-			// results are the same.
-			if query == "" {
-				query += "{" + metricsSelector + "}"
-			} else {
-				query += " + {" + metricsSelector + "}"
+
+			matchers := make([]string, 0, len(matcherSet))
```
yeya24, line 105: Can we name the variable differently? Since we use the var name in line 87 already.
Also I think we can reuse matchers slice in the for loop here

pkg/exemplars/proxy.go
```diff
@@ -80,10 +81,11 @@ func (s *Proxy) Exemplars(req *exemplarspb.ExemplarsRequest, srv exemplarspb.Exe
 	)
 
 	for _, st := range s.exemplars() {
-		query := ""
+		var queryParts []string
```
yeya24, line 84: We can probably reuse `queryParts` slice instead of initializing it to nil every time.

yeya24 (commented):
Nice. I think the change makes sense. Just some nits only.

pkg/exemplars/proxy_test.go
fpetkovski: I keep getting ungrouped imports. Do you know why `goimports` doesn't sort them>

yeya24: Probably try something else like `gci`. `goimports` doesn't handle some cases well if I remember correctly

fpetkovski: Thanks! Should be sorted and grouped now.

yeya24: I guess it still needs some changes. We need to group thanos in the last section.

fpetkovski: I see, `goimports  -l -w -local github.com/thanos-io/thanos pkg/` seems to have done the trick.

yeya24: That's good to know!

yeya24 (approved):
LGTM

